PR #2385, opened 2019-09-22
Change spatial model default frame to icrs

cdeil:
This PR changes the spatial model default frame from "galactic" to "icrs".

Using ICRS is much more common in astronomy, Galactic frame is more exotic, that was introduced just because the Gammapy devs do Galactic science.

Putting "icrs" gives consistency with SkyCoord and Map.

@adonath - Could you please fix this reference file?
https://gist.github.com/cdeil/53057a588522e1fe8ec521214d67606f
It's pretty annoying to have this non-stable reference data file while we iterate on models. E.g. it will need another update once we change any parameter or model name. I guess it's hard to avoid that reference file.

### DIFF
--- a/gammapy/cube/tests/test_fit.py
+++ b/gammapy/cube/tests/test_fit.py
@@ -72,7 +72,7 @@ def get_psf(geom_etrue):
 
 @pytest.fixture
 def sky_model():
-    spatial_model = SkyGaussian(lon_0="0.2 deg", lat_0="0.1 deg", sigma="0.2 deg")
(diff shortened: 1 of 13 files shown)


adonath:
I appended one commit adding the `frame` attribute to the model serialisation and updated the test files in https://github.com/gammapy/gammapy-extra/commit/417b97cfffd8a8c3246fb1c7a0edb12862530dc6. Travis should go green now...

adonath:
Thanks @cdeil! Please go ahead with you cleanup of the spatial models...

PR #2390, opened 2019-09-23
High-level interface: 3D analysis use case

Bultako:
This PR adds features for 3d analysis in high-level interface as well as a 3d config template.

It also enables access for reduced datasets and the `fit` object as attributes of the `Analysis` class and replaces existing method names `reduce()` and `fit()` by `get_datasets()` and `run_fit()` 

Documentation has been updated accordingly.

```
>>> from gammapy.scripts import Analysis 
>>> analysis = Analysis(template="3d") 
INFO:gammapy.scripts.analysis:Setting logging config: {'level': 'INFO'}
>>> analysis.get_observations() 
INFO:gammapy.scripts.analysis:Fetching observations.
INFO:gammapy.scripts.analysis:Info for OBS_ID = 23592
- Start time: 53347.91
- Pointing pos: RA 82.01 deg / Dec 22.01 deg
- Observation duration: 1686.0 s
- Dead-time fraction: 6.212 %

INFO:gammapy.scripts.analysis:Info for OBS_ID = 23523
- Start time: 53343.92
- Pointing pos: RA 83.63 deg / Dec 21.51 deg
- Observation duration: 1687.0 s
- Dead-time fraction: 6.240 %

INFO:gammapy.scripts.analysis:Info for OBS_ID = 23526
- Start time: 53343.95
- Pointing pos: RA 83.63 deg / Dec 22.51 deg
- Observation duration: 1683.0 s
- Dead-time fraction: 6.555 %

INFO:gammapy.scripts.analysis:Info for OBS_ID = 23559
- Start time: 53345.96
- Pointing pos: RA 85.25 deg / Dec 22.01 deg
- Observation duration: 1686.0 s
- Dead-time fraction: 6.398 %
>>> analysis.get_datasets()                                                                                                                                                       
INFO:gammapy.cube.make:Processing observation: OBS_ID = 23592
INFO:gammapy.cube.make:Processing observation: OBS_ID = 23523
INFO:gammapy.cube.make:Processing observation: OBS_ID = 23526
INFO:gammapy.cube.make:Processing observation: OBS_ID = 23559
INFO:gammapy.scripts.analysis:Reading model.
INFO:gammapy.scripts.analysis:SkyModel

Parameters: 

	   name     value   error      unit         min       max    frozen
	--------- --------- ----- -------------- --------- --------- ------
	    lon_0 8.363e+01   nan            deg 8.000e+01 8.500e+01  False
	    lat_0 2.214e+01   nan            deg 2.000e+01 2.400e+01  False
	    index 2.600e+00   nan                      nan       nan  False
	amplitude 5.000e-11   nan cm-2 s-1 TeV-1       nan       nan  False
	reference 1.000e+00   nan            TeV       nan       nan   True
>>> analysis.run_fit()                                                                                                                                                            
INFO:gammapy.scripts.analysis:Fitting with joint likelihood.
INFO:gammapy.scripts.analysis:OptimizeResult

	backend    : minuit
	method     : minuit
	success    : True
	message    : Optimization terminated successfully.
	nfev       : 241
	total stat : 61231.07
>>> analysis.get_residuals()                                                                                                                                                      
INFO:gammapy.scripts.analysis:Calculating residuals.
INFO:gammapy.scripts.analysis:WcsNDMap

	geom  : WcsGeom 
 	axes  : ['lon', 'lat', 'energy']
	shape : (250, 250, 4)
	ndim  : 3
	unit  : 
	dtype : float64
```


gammapy/scripts/analysis.py
```diff
@@ -63,13 +64,16 @@ def __init__(self, config=None, template="basic"):
         self.observations = None
         self.geom = None
```
adonath: The information such as `geom`, `psf_kernel`, `maps`, `images`  is available on the `Dataset` objects, so users can write:
```
analysis.datasets.datasets[0].psf
analysis.datasets.datasets[0].counts.geom
```
I realise, that this is not as convenient as having it directly on the `Analysis` object, but the problem with these single arguments is, that in the end there will be many datasets. I.e. not only a single `PSFKernel` but 20 of those. The same goes for the the `geom` in the end there are many (different) geoms like for the counts map, the exposure map and IRF maps.

My proposal would be to remove `self.geom`, `self.psf_kernel`, `self.maps` and `self.images`. Instead I could  open a follow-up PR to simplify the dataset access so that users can access datasets by name like so:
```
analysis.datasets["obs-1234"].psf
analysis.datasets["obs-1234"].counts.geom
```
What do you think?

gammapy/scripts/analysis.py
```diff
@@ -81,15 +85,21 @@ def settings(self):
...
-            log.info("Fitting available only for joint likelihood with 1D spectrum.")
+        if self._validate_fitting_settings():
+            for obs in self.datasets.datasets:
+                # TODO: fit_range handled in jsonschema validation class
+                if "fit_range" in self.settings["fit"]:
+                    e_min = u.Quantity(self.settings["fit"]["fit_range"]["min"])
+                    e_max = u.Quantity(self.settings["fit"]["fit_range"]["max"])
+                    obs.mask_fit = obs.counts.energy_mask(e_min, e_max)
+                obs.model = self.model
+            log.info("Fitting with joint likelihood.")
```
adonath: Nitpicky comment: currently for 3D it's not a joint-likelihood fit, but we stack the data before.

gammapy/scripts/analysis.py
```diff
@@ -179,31 +202,12 @@ def get_observations(self):
...
-                table_psf, self.geom, **psf_params
-            )
-
-        else:
-            # TODO raise error?
-            log.info("Data reduction method not available.")
+    def get_residuals(self):
+        log.info("Calculating residuals.")
+        npred = self.datasets.datasets[0].npred()
+        self.residuals = Map.from_geom(self.geom)
```
adonath: Note that all of our dataset classes already have `.residuals()` method, the you can just re-use. See:
https://docs.gammapy.org/dev/api/gammapy.cube.MapDataset.html#gammapy.cube.MapDataset.residuals
https://docs.gammapy.org/dev/api/gammapy.spectrum.SpectrumDatasetOnOff.html#gammapy.spectrum.SpectrumDatasetOnOff.residuals

PR #2392, opened 2019-09-24
High-level interface: refactor config settings templates

Bultako:
This PR addresses some/most of the points discussed in last developers call concerning refactoring and handling of templates for default configuration settings. The main idea is to use several YAML files as templates for the default values of config settings, instead of a single centralized `schema.yaml` file, hoping that maintenance of several simple files will be easier than for a single complex centralized file. The `schema.yaml` will now only be used to validate the values of the settings provided by the user via the interface or custom config file. 

The dump of the settings to a file with `analysis.config.dump()` writes only the YAML formatted param/values of the settings, not commented helping lines are provided in the dump. 

It remains the question of how to provide clear documentation for the format and structure of the parameter/values in the configuration settings. The commented helping lines in the internal `schema.yaml` file have been kept, these are used to generate the different sections of the docs on config settings and also in the `analysis.config.print_help()` method.

Another issue that have to be solved is how to link to YAML files describing the models, in the `schema.yaml` file used for validation. Do we already use a mechanism to validate the param/values of YAML serialized models?

### DIFF
--- a/docs/scripts/index.rst
+++ b/docs/scripts/index.rst
@@ -40,6 +40,7 @@ them into a file that you can edit to start a new analysis from the modified con
 
     >>> print(analysis.config)
     >>> analysis.config.dump("config.yaml")
+    INFO:gammapy.scripts.analysis:Configuration settings saved into config.yaml
     >>> analysis = Analysis.from_file("config.yaml")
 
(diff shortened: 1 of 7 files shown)


adonath:
Thanks a lot @Bultako!

PR #2393, opened 2019-09-24
Minor cleanup of the Analysis class

adonath:
This PR cleans-up and simplifies the `Analysis` class. It includes the following changes:

- Implement `Analysis.from_template()` and remove the template handling from `__init__`. I also implemented `Config.read()`. This way config files can be read and validated independently.
- Remove the model specification from the config file and replace it with a file reference. This will simplify the jsonschema of the analysis config.
- Rename the `data_reducer` parameter to `dataset-type`. Mainly because I wanted to learn and try to edit the JSON schema.


gammapy/scripts/analysis.py
```diff
@@ -44,24 +44,25 @@ class Analysis:
...
     >>> from gammapy.scripts import Analysis
-    >>> analysis = Analysis(template="1d")
+    >>> analysis = Analysis.from_template(template="1d")
 
     TODO: show a working example of running an analysis.
     Probably not here, but in high-level docs, linked to from class docstring.
     """
 
-    def __init__(self, config=None, template="basic"):
-        self._config = Config(config, template)
```
Bultako: This is a relevant point.
I think it's good to have a default settings by default when instantiating `Analysis`, if not I think we risk having validation errors during the analysis process.

adonath: I see your point. I guess in the end it's just a matter of the API. As far as I understood these two things would lead to the same result :
```
# old
analysis = Analysis(config, template="basic")

# new
analysis = Analysis.from_template("basic")
analysis.config.update_settings(config)
```
Is this correct?

If so I find the latter solution a little bit simpler and more explicit, because it's clear, that you start from a basic configuration and then overwrite and extend it by user settings.
In the first case it was unclear to me at first how the `template` and the `config` are combined internally. It also lead to much more complicated code, because you had to pass the `template` info to the `Config` object, where it was read from file and then the user and template setting were stored and handled separately.

adonath:
Thanks for the review @Bultako! I'll go ahead and update the docs page now...

gammapy/scripts/analysis.py
adonath: Ah, I guess the difference is that in the first case the `config` is validated on init, while in the second case `config` is not validated?

I guess it's probably a good idea if the validation always happens on `.update_settings()` anyway?

Bultako: I agree with you.
But what I meant was just having basic default settings when doing `analysis = Analysis()` which I guess it is not forbidden.

adonath: I could also restore the behavior, that `Analysis.from_template("basic")` is equivalent to `Analysis(config=None)` if you prefer. Alternatively we can make `config` a required
 argument...

Bultako: > I could also restore the behavior, that Analysis.from_template("basic") is equivalent to Analysis(config=None)

Yep
If the user really wants an empty config it's just `Analysis(config={})` should do it, no?

adonath: OK, I'll change the init default.

adonath: Done

gammapy/scripts/analysis.py
```diff
@@ -44,24 +44,25 @@ class Analysis:
...
-    template : str
-        Template for configuration settings
 
     Examples
     --------
     Example how to create an Analysis object:
 
     >>> from gammapy.scripts import Analysis
-    >>> analysis = Analysis(template="1d")
+    >>> analysis = Analysis.from_template(template="1d")
```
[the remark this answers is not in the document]
adonath: In this case it probably does not matter, but there is a huge advantage in calling methods by keyword arguments: you can change order and add keyword arguments later. E.g. if we add `Analysis.from_template(base_template="1d", template="1d-extended")`. Also users and developers see explicitly which argument is set, if there are more arguments. E.g. compare:
```python3
geom = WcsGeom.create((0, 0), "5 deg", "CEL", "TAN")

# vs

geom = WcsGeom.create(
    skydir=(0, 0),
    width="5 deg",
    coordsys="CEL",
    proj="TAN"
)
```
The latter version is more code, but much better readable I think.

adonath:
Thanks @Bultako!

Issue #2353, opened 2019-09-09
Clean up model names


cdeil:
This is a reminder issue for myself to clean up the model class names, once #2319 is done.
Moving the models broke all user scripts & notebooks anyways, so now is a good time to do renames of models.

Plan for names is to add "Source", "Spatial", "Spectral" and "Time" in a uniform way: https://github.com/gammapy/gammapy/blob/master/docs/development/pigs/pig-016.rst#introduce-gammapymodeling

cdeil:
@adonath @registerrier and I discussed and agreed on model names yesterday:
https://github.com/gammapy/gammapy/wiki/Model-class-names

I will do the renames today.

adonath:
The remaining action item here is to find a goo scheme and rename the `SkyModel` and `SkyDiffuseCube` models...

adonath:
My proposed solution here is to allow for energy dependent spatial models. This means the functionality of the `SkyDiffuseCube` is taken over by the `TemplateSpatialModel`.

If we stick with the current `SkyModel` then we should probably rename the `BackgroundModel` to `DataModel`. Or the other way around, rename `SkyModel` to `SourceModel`.

adonath:
After #3051 and #3089, the background handling is taken over by the `FoVBackgroundModel` class. The current `BackgroundModel` ca still be used to handle template in reconstructed energy. The Plan here should be to replace this with `TemplateSpatialModel` and `apply_irf={"exposure": False, "psf": Falsse, "edisp": False}` and document it. Then the `BackgroundModel` class could go...

adonath:
Alternatively we could just rename `BackgroundModel` to `NpredTemplateSpatialModel` or `DataTemplateSpatialModel`...

QRemy:
+1 to keep it and rename as`NpredTemplateSpatialModel`

registerrier:
+1 to keep as well.
How about `CountsTemplateSpatialModel`?

QRemy:
Counts would implicitly mean `int` no ? which is not the case

registerrier:
> `Counts` would implicitly mean `int` no ? which is not the case

Right, indeed. If the terminology `npred`/`counts` is clear enough it should be fine to use `NpredTemplateSpatialModel`.
BTW does it really need to be `Spatial`?

QRemy:
maybe `NpredTemplateSkyModel` as it takes a spectral_model as argument

adonath:
Here is my overview:
[...]

PR #2394, opened 2019-09-25
Add normalize attribute to `TemplateSpatialModel` serialization

QRemy:


adonath:
Thanks @QRemy! Please provide a short description (one sentence...) of the content of the PR next time...

 I was wondering whether we should remove the `normalize` argument completely. I think the Fermi XML spec does not have such an argument either. What do you think?

QRemy:
I think it's useful to keep this option.  For example you want to have the normalize False when using  a gas map, but True when fitting a morphological template obtained form residuals.

adonath:
@QRemy I agree. I just checked with @cdeil, that the Fermi Science Tools as well as gammalib have similar options:
- `"map_based_integral"` (https://fermi.gsfc.nasa.gov/ssc/data/analysis/scitools/xml_model_defs.html#SpatialMap)
- `"normalize"` (http://cta.irap.omp.eu/gammalib-devel/doxygen/classGModelSpatialDiffuseMap.html#afb99a59c33796fd7c4b9573c6a25a084)

So let's keep it. I have two minor comments then:
- Please rebase after #2353
- Remove the in place `.normalize()` method and move the code to `__init__()` in the `if normalize:` block . Then rename the parameter consistently (in the serialisation as well) to `normalize`.

cdeil:
@qremy or @adonath when you merge this in: could you please change to consistent `normalize` with `z` also in `gammapy/irf/psf_check.py` and the docstrings?
See  https://gist.github.com/cdeil/b4b97a5f029ae7123f1ad2b69dab52aa#file-gistfile1-txt-L15

(we already decided and changed to consistent `optimize`, `serialize`, ...)

adonath:
Thanks @QRemy!

PR #2396, opened 2019-09-25
Fix fgst ccube format unit handling

adonath:
This PR fixes a bug reported by @QRemy in private communication. The error was introduced in #2379, because of confusing meta information in the Fermi counts cube files. While the primary header specifies `MeV` as a unit for the third axis, the `EBOUNDS` table actually uses `keV`.
A regression test was also added.


cdeil:
@QRemy @adonath - Nice catch!

I think generally the primary header sometimes is lying, also it has WCS info for the energy axis that generally should be ignored - only the energy table should be used.

PR #2397, opened 2019-09-25
Minor cleanup of Analysis class #2

adonath:
This PR is a follow up to #2393. It implements the following changes:
- Move the `geometry` section to `reduction.geom`, I think the definition of the data geom is clearly part of the data reduction
- Allow for a different true geom in 3d analysis config
- Move `offset_max` to `reduction` section. Again the offset max is part of the data reduction and to the geometry.
- I removed the `max_radius_psf` from `geometry`. Again it's rather part of the data reduction and not the geometry.


gammapy/scripts/config/schema.yaml
```diff
@@ -186,6 +186,57 @@ properties:
...
+                        type: array
+                        items: {type: number}
+                        minItems: 1
+                        maxItems: 2
+                    axes:
+                        type: array
+                        additionalProperties: false
+                        items:
+                            "$ref": "#/definitions/map_axis"
+                required: [axes]
```
Bultako: I think we can extract the two duplicated geom sections and reference to them as definitions, in the same way it is done with i.e. models or map-axis

gammapy/scripts/config/schema.yaml
```diff
@@ -229,36 +279,34 @@ properties:
 #                unit: TeV
 #                interp: log
 #
```
Bultako: as `geom` is not a main section of the settings anymore, shouldn't we integrate this helping comments block into the `reduction` block somehow?

gammapy/scripts/config/schema.yaml
```diff
@@ -229,36 +279,34 @@ properties:
...
+#                maxItems: 2
+#            axes:
+#                type: object
+#                additionalProperties: false
+#                properties:
+#                    e_reco:
+#                        "$ref": "#/definitions/energy_axis"
+#                    e_true:
+#                        "$ref": "#/definitions/energy_axis"
+#        required: [axes]
```
Bultako: better remove? it's starting to be a long file..

gammapy/scripts/config/schema.yaml
```diff
@@ -186,6 +186,57 @@ properties:
...
+                        type: array
+                        items: {type: number}
+                        minItems: 1
+                        maxItems: 2
+                    axes:
+                        type: array
+                        additionalProperties: false
+                        items:
+                            "$ref": "#/definitions/map_axis"
+                required: [axes]
```
Bultako: Do we need to add more required properties other than `axes` or remove the requirement on this one? Note that as it is defined here `axes` will only be required if `geom` is not empty. We also define below conditional requirements for `geom` when `dataset-type` is a `MapDataset`

gammapy/scripts/config/schema.yaml
```diff
@@ -342,7 +390,7 @@ if:
                     const: MapDataset
 then:
     properties:
-        geometry:
+        geom:
```
Bultako: Since `geom` is now nested into `reduction` this should also be reflected in the conditionals.

```
if:
    properties:
        reduction:
            properties:
                dataset-type:
                     const: MapDataset
then:
    properties:
        reduction:
            properties:
            required: [geom]
                geom:
                    required: [binsz, width, skydir, coordsys, proj, axes]
                    properties:
                        axes:
                        required: [e_reco]
```

Note that you may add also a conditional validation for `geom-irf` if you need it.

gammapy/scripts/config/schema.yaml
```diff
@@ -186,6 +186,57 @@ properties:
             dataset-type:
                 enum: [not assigned, SpectrumDatasetOnOff, MapDataset]
             containment_correction: {type: boolean}
+            offset-max: {type: string}
```
Bultako: This is actually a number, because astrropy quantities are considered as numbers in the validation as in the case for `lon` and `lat` for the observations filtering above.

adonath: Done

gammapy/scripts/config/schema.yaml
adonath: Done

gammapy/scripts/config/schema.yaml
adonath: Done

gammapy/scripts/config/schema.yaml
adonath: Done

gammapy/scripts/config/schema.yaml
adonath: Done

Issue #2374, opened 2019-09-18
Use float32 dtype for maps


cdeil:
Most of the memory and CPU time in Gammapy is related to maps.

Currently we have a mix of 32 and 64 bit float dtype for maps.

I suggest we chant to use 32 bit float as much as possible

I expect that this will significantly reduce memory use of Gammapy, and improve speed, with only ~ 10 lines of code change.

A related question that we could discuss here or elsewhere is whether we should change maps with `int` to `float32` also. My guess is that we use int64 currently for counts data in some places, so float32 would reduce memory use. Probably speed doesn't change, depending on what operations we do for these counts maps. Overflow shouldn't be an issue. Another advantage of using float maps everywhere is that possibly we can remove the complex logic to handle pixels that aren't "on the sky" like for AIT WCS, because then the float `nan` can be used everywhere, not a special int value like -1 and special code is needed for this.

It's non-trivial but not really hard to find the places where we create maps with undesired dtype. E.g. with this one line change
```
diff --git a/gammapy/maps/base.py b/gammapy/maps/base.py
index bec1eec0a..65dd9fcbb 100644
--- a/gammapy/maps/base.py
+++ b/gammapy/maps/base.py
@@ -70,6 +70,8 @@ class Map(abc.ABC):
 
     @data.setter
     def data(self, val):
+        print(f"\nCreating map with dtype {val.dtype} and shape {val.shape}")
+
         if val.shape != self.geom.data_shape:
             raise ValueError(
                 "Shape {val.shape!r} does not match map data shape {self.geom.data_shape!r}"
```
and running the gammapy.cube tests I get this:
https://gist.github.com/cdeil/fa92fd4c35bac6858fbbf687922dd5f2

We could also log the tracebacks showing the callers - probably it's just a few places in the code where we need to edit.

I'm not sure yet if / how it's possible to avoid re-introducing float64 or int dtypes in the future.
Probably we want to have a dtype check in the `map.data` setter, and e.g. log warnings or fail tests when float64 appears (but allow it generally for users of Map outside Gammapy)?

Yesterday, I was thinking something could be done at the Numpy array level, but now I don't think that's the case any more. Related: https://stackoverflow.com/questions/57982230/how-to-avoid-numpy-type-conversions

This is a reminder issue for me at the v0.15 milestone - but comments & feedback or help are welcome any time of course!

registerrier:
Is this issue still relevant? All `Dataset` members are `float32` or `bool`. 
[...]

PR #2405, opened 2019-09-26
Change value clipping in `LogScale` class

QRemy:
Change interpolation clipping to float64 precision
The clipping to float32 precision caused a bug when evaluating a table model containing the gamma-ray emissivity spectrum of the gas (that is <<10**-38/MeV/s/sr above 500GeV)

### DIFF
--- a/gammapy/utils/interpolation.py
+++ b/gammapy/utils/interpolation.py
@@ -139,7 +139,7 @@ class InterpolationScale:
 class LogScale(InterpolationScale):
     """Logarithmic scaling"""
 
-    tiny = np.finfo(np.float32).tiny
+    tiny = np.finfo(np.float64).tiny
 
     def _scale(self, values):
         values = np.clip(values, self.tiny, np.inf)
--- a/gammapy/utils/interpolation.py
+++ b/gammapy/utils/interpolation.py
@@ -151,7 +151,10 @@ class LogScale(InterpolationScale):
         output = np.exp(values)
         is_tiny = abs(output) - self.tiny <= self.tiny
         if np.any(is_tiny):
-            output[is_tiny] = 0.0
+            try:
+                output[is_tiny] = 0.0
+            except(TypeError):
+                output = 0.0
             warnings.warn(
                 "Interpolated values reached float32 precision limit", Warning
             )
--- a/gammapy/utils/interpolation.py
+++ b/gammapy/utils/interpolation.py
@@ -95,10 +95,7 @@ class ScaledRegularGridInterpolator:
             values = self.scale.inverse(values)
 
         tiny = np.finfo(np.float32).tiny
-        try:
-            mask = abs(values.value) - tiny <= tiny
-        except (AttributeError):
-            mask = abs(values) - tiny <= tiny
+        mask = abs(values.value) - tiny <= tiny
         if np.any(mask):
             values[mask] = 0.0
             warnings.warn(
--- a/gammapy/utils/interpolation.py
+++ b/gammapy/utils/interpolation.py
@@ -94,19 +94,6 @@ class ScaledRegularGridInterpolator:
             values = self._interpolate(points[0])
             values = self.scale.inverse(values)
 
-        tiny = np.finfo(np.float32).tiny
-        try:
-            mask = abs(values.value) - tiny <= tiny
-        except (AttributeError):
-            mask = values - tiny <= tiny
-        if np.any(mask):
-            values[mask] = 0.0
-            warnings.warn(
-                "Interpolated values reached float32 precision limit", Warning
-            )
-            # for example TableModel used to define a diffuse model
-            # could require large precision so users may want to redefine unit scaling.
-
         if clip:
             values = np.clip(values, 0, np.inf)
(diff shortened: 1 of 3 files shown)


QRemy:
Here is an example:
```python
energy=np.array([1.00000000e+05, 1.25892541e+05, 1.58489319e+05, 1.99526231e+05,
       2.51188643e+05, 3.16227766e+05, 3.98107171e+05, 5.01187234e+05,
       6.30957344e+05, 7.94328235e+05, 1.00000000e+06,1.25892541e+06,
       1.58489319e+06, 1.99526231e+06])

values=np.array([7.26683943e-35, 3.95773531e-35, 2.15549951e-35, 1.17394868e-35,
       5.44841625e-36, 2.43964992e-36, 1.09240767e-36, 4.89149905e-37,
       2.19027782e-37, 9.80745754e-38, 4.39150790e-38,1.96639562e-38,
       8.80497507e-39, 3.94262401e-39])

model=TableModel(energy=energy,values=values*u.Unit("MeV-1 s-1 sr-1"))
result=model.evaluate(energy,norm=1.,tilt=0.,reference=1*u.TeV)
print(result)
print(np.finfo(np.float32).tiny)

[7.26683943e-35 3.95773531e-35 2.15549951e-35 1.17394868e-35
 5.44841625e-36 2.43964992e-36 1.09240767e-36 4.89149905e-37
 2.19027782e-37 9.80745754e-38 4.39150790e-38 1.96639562e-38
 1.17549435e-38 1.17549435e-38] 1 / (MeV s sr)
1.1754944e-38
```
The two last values are below float32 precision so they are clipped to a higher value than expected.
Changing this to float64 here does not prevent to have the flux or npred of the final model (spatial+spectral) returned in float32 if you want.

I added this test:
```python
np.testing.assert_allclose(values/values.max(),result.value/values.max())

  File "<ipython-input-532-185cc20d386d>", line 15, in <module>
    np.testing.assert_allclose(values/values.max(),result.value/values.max())

  File "/anaconda3/envs/gpy-dev/lib/python3.7/site-packages/numpy/testing/_private/utils.py", line 1501, in assert_allclose
    verbose=verbose, header=header, equal_nan=equal_nan)

  File "/anaconda3/envs/gpy-dev/lib/python3.7/site-packages/numpy/testing/_private/utils.py", line 827, in assert_array_compare
    raise AssertionError(msg)

AssertionError: 
Not equal to tolerance rtol=1e-07, atol=0

Mismatch: 14.3%
Max absolute difference: 0.00010751
Max relative difference: 0.66459864
 x: array([1.000000e+00, 5.446295e-01, 2.966213e-01, 1.615487e-01,
       7.497642e-02, 3.357237e-02, 1.503278e-02, 6.731261e-03,
       3.014072e-03, 1.349618e-03, 6.043216e-04, 2.705985e-04,
       1.211665e-04, 5.425500e-05])
 y: array([1.000000e+00, 5.446295e-01, 2.966213e-01, 1.615487e-01,
       7.497642e-02, 3.357237e-02, 1.503278e-02, 6.731261e-03,
       3.014072e-03, 1.349618e-03, 6.043216e-04, 2.705985e-04,
       1.617614e-04, 1.617614e-04])
```

adonath:
Thanks @QRemy! I agree the clipping of the tiny values might be an issue, but I think the fix needs some more thought, because it re-introduces the problem for `np.float32` values, that we tried to solve with the clipping:
```python3
from gammapy.utils.interpolation import LogScale
import numpy as np

LogScale.tiny = np.finfo(np.float64).tiny

values = np.array([1e-12, 0], dtype=np.float32)

scale = LogScale()
scale(values)
```
Gives:
```
array([-27.631021,       -inf], dtype=float32)
```
Where the `-inf` value breaks the linear interpolation, that is then done on the scaled values (I guess these are the test fails we see on Travis...). I guess there are two solutions:
- If we really need high precision fluxes at high energy, we could change the default units for the Fermi isotropic diffuse model.
- Somehow check the dtype of `values` (e.g. using `np.dtype()`) and set the clip values based on the dtype. I can't remember if there was a good motivation to set it to a fixed value, probably not. But then the issue well appear again if we maybe address #2374.

gammapy/utils/interpolation.py
```diff
@@ -92,9 +94,21 @@ def __call__(self, points, method="linear", clip=True, **kwargs):
             values = self._interpolate(points[0])
             values = self.scale.inverse(values)
 
+        tiny = np.finfo(np.float32).tiny
```
adonath: As this is a specific problem of the `LogScale`, I think we should move the code for the clipping there. So my proposal would be to define `self.is_tiny < (values < self.tiny)` in `LogScale._scale` and the re-apply the mask in `LogScale._inverse` with `values[self._is_tiny] = 0`. This will also get rid of the `try` and `except` block for quantities, because the interpolation scale is called with the units stripped-of.

QRemy: I moved the clipping code to LogScale.inverse, and I defined the mask only here as it need to be applied only to the output values. I added an other try and except block for the case where value is not an array.

PR #2407, opened 2019-09-26
Minor cleanup of Analysis class #3

Bultako:
This PR is a follow up to #2397. It implements the following changes: 

- Provide a shorter and more explicit error message on validation of settings with `json-schema`.
- Always validate settings when updating with `config.update_settings` method.
- Print only single-# comments in `config.print_help()` to avoid printing undesired other comments as help (yes, we need to handle help display in a different way in the future..)
- Add validation for non-empty model in `run_fit()`
- Improve documentation and consistency in docstrings.
- Add validation for non-empty `fp_binning` in `get_flux_points`method.
- Add a default parameter `source` in `get_flux_points`method.

In the meantime I've noticed that the flux points binning used for `MapDatasets` comes from the actual energy axis binning of the dataset, and does not use the `e_edges` param  of `FluxPointEstimator` method. I don't know if there's a reason for that.

### DIFF
--- a/docs/scripts/index.rst
+++ b/docs/scripts/index.rst
@@ -33,6 +33,8 @@ an IPython console or a notebook.
     >>> analysis = Analysis()
         INFO:gammapy.scripts.analysis:Setting logging config: {'level': 'INFO'}
 
+Configuration and methods
+=========================
 You can have a look at the configuration settings provided by default, and also dump
 them into a file that you can edit to start a new analysis from the modified config file.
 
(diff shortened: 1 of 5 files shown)


adonath:
Just a quick comment on:
>In the meantime I've noticed that the flux points binning used for MapDatasets comes from the >
actual energy axis binning of the dataset, and does not use the e_edges param of FluxPointEstimator method. I don't know if there's a reason for that.

I think there is misunderstanding. For the computation of flux points the energy bins of the data are "grouped". That means in general the energy binning of the data should be "finer" than the one used for the flux points. If this is not the case (if the flux point being is finer or equivalent to the binning of the data), the resulting binning of the flux points will be equivalent to the data. In you specific case I would propose to reduce the data binning back to 4 bins and further reduce the flux point binning to 2 bins (also to reduce the run-time for the test).

Bultako:
> In your specific case I would propose to reduce the data binning back to 4 bins and further reduce the flux point binning to 2 bins (also to reduce the run-time for the test). I hope this makes sense...

Ok.
Done.

adonath:
Thanks @Bultako!

PR #2408, opened 2019-09-27
Rewrite hess.ipynb tutorial

adonath:
This PR rewrites the `hess.ipynb` tutorial to use the new `Analysis` class. 

ToDo:
- [x] Add more explanations
- [x] Add computing flux points
- [ ] Add exercises

### DIFF
--- a/tutorials/hess.ipynb
+++ b/tutorials/hess.ipynb
@@ -4,15 +4,27 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "# H.E.S.S. with Gammapy\n",
+    "# Analysis of H.E.S.S. DL3 data with Gammapy\n",
     "\n",
-    "This tutorial explains how to analyse [H.E.S.S.](https://www.mpi-hd.mpg.de/hfm/HESS) data with Gammapy.\n",
+    "In September 2018 the [H.E.S.S.](https://www.mpi-hd.mpg.de/hfm/HESS) collaboration released a small subset of archival data in FITS format. This tutorial explains how to analyse this data with Gammapy.\n",
     "\n",
-    "We will analyse four observation runs of the Crab nebula, which are part of the [H.E.S.S. first public test data release](https://www.mpi-hd.mpg.de/hfm/HESS/pages/dl3-dr1/). In this tutorial we will make an image and a spectrum. The [light_curve.ipynb](light_curve.ipynb) notbook contains an example how to make a light curve.\n",
+    "We will analyse four observation runs of the Crab nebula, which are part of the [H.E.S.S. first public test data release](https://www.mpi-hd.mpg.de/hfm/HESS/pages/dl3-dr1/). The data was release without correspoding background models. In [background_model.ipynb](background_model.ipynb) we show how to make a simple background model, which is also used in this tutorial. The background model is not perfect; it assumes radial symmetry and is in general derived from only a few observations.\n",
     "\n",
-    "To do a 3D analysis, one needs to do a 3D background estimate. In [background_model.ipynb](background_model.ipynb) we have started to make a background model, and in this notebook we have a first look at a 3D analysis. But the results aren't OK yet, the background model needs to be improved. In this analysis, we also don't use the energy dispersion IRF yet, and we only analyse the data in the 1 TeV to 10 TeV range. The H.E.S.S. data was only released very recently, and 3D analysis in Gammapy is new. This tutorial will be improved soon.\n",
+    "**Note:** The high level `Analysis` class is a new feature added in Gammapy v0.14. In the curret state it supports the standard analysis cases of a joint or stacked 3D and 1D analysis. It provides only limited access to analaysi parameters via the config file. It is expected that the format of the YAML config will be extended and change in future Gammapy versions.\n",
     "\n",
-    "This tutorial also shows how to do a classical image analysis using the ring bakground. "
+    "We will first show how to configure and run a stacked 3D analysis and then address the classical spectral analysis using reflected regions later. The structure of the tutorial follows a typical analysis\n",
+    "\n",
+    "- Analysis configuration\n",
+    "- Observation slection\n",
+    "- Data reduction\n",
+    "- Model fitting"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Setup"
    ]
   },
   {
@@ -22,7 +34,8 @@
    "outputs": [],
    "source": [
     "%matplotlib inline\n",
-    "import matplotlib.pyplot as plt"
+    "import matplotlib.pyplot as plt\n",
+    "import numpy as np"
    ]
   },
   {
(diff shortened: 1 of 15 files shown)


tutorials/hess.ipynb
```diff
@@ -72,12 +69,64 @@
...
+    "model: model.yaml\n",
+    "\n",
+    "fit:\n",
+    "    fit_range:\n",
+    "        max: 100 TeV\n",
+    "        min: 1 TeV\n",
+    "\n",
+    "flux: {}\n",
+    "\"\"\"\n",
+    "filename = Path(\"config.yaml\")\n",
```
cdeil: Usually if one writes the config in Python scripts or Jupyter notebooks, one would just use it, not write it to file, no?
Wouldn't it be better to just pass it to some Config classmethod that takes a YAML config string and uses it?
Or if you want to write to file, suggest to use [%%writefile](https://ipython.readthedocs.io/en/stable/interactive/magics.html#cellmagic-writefile) and remove the Path import and code.

tutorials/hess.ipynb
```diff
@@ -72,12 +69,64 @@
...
+    "        proj: TAN\n",
+    "        axes:\n",
+    "          - name: energy\n",
+    "            hi_bnd: 30\n",
+    "            lo_bnd: 1\n",
+    "            nbin: 5\n",
+    "            interp: log\n",
+    "            node_type: edges\n",
+    "            unit: TeV\n",
+    "    geom-irf:\n",
```
cdeil: Suggest to remove `geom-irf` for now, keep it simple.

tutorials/hess.ipynb
```diff
@@ -72,12 +69,64 @@
...
+    "    filters:\n",
+    "        - filter_type: par_value\n",
+    "          value_param: Crab\n",
+    "          variable: TARGET_NAME\n",
+    "\n",
+    "reduction:\n",
+    "    dataset-type: MapDataset\n",
+    "    stack-datasets: true\n",
+    "    offset-max: 2.5 deg\n",
+    "    psf-kernel-radius: 0.3 deg\n",
```
cdeil: I see e.g. `psf-kernel-radius` doesn't appear here: https://docs.gammapy.org/dev/scripts/index.html
Can we get a dump of the fixed part of the schema so that this at least shows up in one place in the docs?
Also: this is a very low-level parameter, should it even exist? Shouldn't this be chosen by Gammapy from the PSF size to have good containment?
If you expose it or put any Angle as default, you have created a simple way to silently get bad analysis results and publish wrong science results, e.g. pair halos for AGN or Crab extension ...

tutorials/hess.ipynb
```diff
@@ -72,12 +69,64 @@
...
+    "        axes:\n",
+    "            - name: energy\n",
+    "              hi_bnd: 30\n",
+    "              lo_bnd: 1\n",
+    "              nbin: 11\n",
+    "              interp: log\n",
+    "              node_type: center\n",
+    "              unit: TeV\n",
+    "\n",
+    "model: model.yaml\n",
```
cdeil: Do we need this reference to the other file? Can't we just remove this from the config schema and just take model config separately in the API?
If you put it here, it makes it harder or impossibly to do the use case where one writes config from Python & Jupyter notebooks and doesn't write to files at all.

tutorials/hess.ipynb
```diff
@@ -72,12 +69,64 @@
...
    "outputs": [],
    "source": [
-    "data_store = DataStore.from_file(\n",
-    "    \"$GAMMAPY_DATA/hess-dl3-dr1/hess-dl3-dr3-with-background.fits.gz\"\n",
-    ")\n",
-    "mask = data_store.obs_table[\"TARGET_NAME\"] == \"Crab\"\n",
-    "obs_table = data_store.obs_table[mask]\n",
-    "observations = data_store.get_observations(obs_table[\"OBS_ID\"])"
+    "config = \"\"\"\n",
+    "general:\n",
```
cdeil: Default logging should be fine. Remove from this first example people see? There config should be as simple as possible, no?

tutorials/hess.ipynb
```diff
@@ -86,19 +135,45 @@
...
-    "pos_crab = SkyCoord(83.633, 22.014, unit=\"deg\")"
+    "model_config = \"\"\"\n",
+    "components:\n",
+    "- name: crab\n",
+    "  type: SkyModel\n",
+    "  spatial:\n",
+    "    type: PointSpatialModel\n",
+    "    frame: icrs\n",
+    "    parameters:\n",
+    "    - name: lon_0\n",
```
cdeil: Can you make an alternative parameter parser work?
```
- lon_0: 83.63 deg
- lat_0: ...
```
This would be shorter and more pleasant to write?
Should be possible to put this in a single place in a generic way for the code that creates `Parameters` from config?

tutorials/hess.ipynb
registerrier: Yes. In particular, the value given here is too small to get correct results. We should not provide a tutorial suggesting wrong parameters to users.

tutorials/hess.ipynb
registerrier: An issue similar as the one below, we have to ensure that definition of IRF is correct given the IRF stored values and the choice of the user reco energy bins.

tutorials/hess.ipynb
adonath: I agree we should remove this parameter for now, because it's really low level and instead choose a reasonable default.

tutorials/hess.ipynb
adonath: @registerrier What do you exactly mean by this? The binning for the `PSFKernel` and `EnergyDispersion` are deferred from the true and reconstructed geoms internally. I think those are chose correctly...

My much preferred solution for this would be to go to a simpler parametrisation for now, as I've proposed a few times:

`MapDataset.create(geom, energy_true_axis, rad_axis, migra_axis, bins_irf)` and later add a `margin` argument to account for spill over.

registerrier: I simply meant that `energy_true_axis` has to be chosen in a way that is consistent with what is in the IRF and with the energy reco binning. e.g. if `e_reco` starts at 300 GeV, `e_true` should start sufficiently below, typically 100 GeV.

adonath: Thanks for the explanation, that makes sense...I guess it's somehow a more general problem, that we don't handle in the code yet. I think we don't check e.g. `e_true` and `e_reco` in the `SpectrumExtraction` either. But I agree it would be simpler to remove the `geom-irf` in the config file for this analysis...

cdeil: Note that Fermi ST and ctools don't expose it like this to end-users.
One can choose an energy binning for counts (`gtbin`), and for exposure (`gtexpcube2`), but otherwise things are always chosen and handled internally.
I think the requirement on the user is to (a) choose counts and exposure with energy binnings that aren't too large (so that PL approximations for interpolation and integration are fine) and such that exposure covers the counts range.
If you open up more options and combinations in the high-level interface you will get more usage error.
I don't know how Fermi ST chooses geoms e.g. for PSF and EDISP internally.

adonath: I agree with you @cdeil. I think we should expose `geom` and `energy_true_axis` for now, which is sufficient for 98% of the users. Later we can maybe add  `rad_axis`, `migra_axis` and `binsz_irf` for consistency and "power-users". The alignment of the different maps and things like margin for the exposure and IRF maps should be handled internally.

Bultako:
In last PR merged on high-level interface I have modified the `hess.ipynb` tutorial to be compliant and expose the last modifs, addressing some of the comments of this thread. Maybe it may be helpful for inspiration in the tutorial rewrite.

https://github.com/gammapy/gammapy/blob/master/tutorials/hess.ipynb

tutorials/hess.ipynb
adonath: Done

tutorials/hess.ipynb
adonath: I removed `geom-irf` for now...

tutorials/hess.ipynb
adonath: Done

tutorials/hess.ipynb
adonath: Done

Issue #2387, opened 2019-09-23
Model registry pattern


cdeil:
We currently have this coding pattern for spatial, spectral and time models:

https://github.com/gammapy/gammapy/blob/3e6a03386d5280d24ab64cc520cc2eddc023d0a5/gammapy/modeling/models/__init__.py#L10-L40

Now we introduced a model `tag` attribute like this:

https://github.com/gammapy/gammapy/blob/3e6a03386d5280d24ab64cc520cc2eddc023d0a5/gammapy/modeling/models/spectral.py#L437

I'd like to discuss a bit if we should keep as-is or change the model registry a bit.

- In the current approach, it's easy to forget to add something to the registry. E.g. `SkyDiffuseConstant` is currently missing.
- A dict mapping class name string to class object isn't super useful (invites typos), one could just have a list of classes, and then `cls.__name__` gives the name and `cls.tag` the tag.
- Since Python 3.6 it's very simple to auto-generate a list of subclasses, see [here](https://docs.python.org/3/whatsnew/3.6.html#pep-487-simpler-customization-of-class-creation).

I think my suggestion would be that we remove the manual registry dicts, and instead do the auto-registry of subclasses.

That always happens on import, so it does make it a bit harder to control for the unlikely problematic cases where different downstream packages (e.g. CTA GPS pipeline and HESS GPS pipeline or CTA EGAL people or something) define a Gammapy model sub-class with the same class name, and someone imports their packages together - then the later import clobbers the earlier. But overall this problem to have a global shared namespace for models in the registry remains, probably other packages and pipelines would still modify the current registry on init and just clobber.

We could maybe avoid this issue by keeping the dict instead of list, and use tag names as keys and try to make them unique, like here:

https://github.com/astropy/astropy/blob/a1e72dd7b8eca5fc95c9f40557e8ceaae30d4d81/astropy/io/misc/asdf/data/schemas/astropy.org/astropy/coordinates/angle-1.0.0.yaml#L5

But that seems like overkill at the moment - I think I would start by using auto-registry on subclassing, and just have a list like [here](https://docs.python.org/3/whatsnew/3.6.html#pep-487-simpler-customization-of-class-creation), and then only if needed later refactor again to something more general (keeping the auto-registry, just changing the registry data structure and mechanism).

@adonath @registerrier @QRemy or anyone interested - Thoughts?

cdeil:
[...]

PR #2409, opened 2019-09-27
Improve model registries and __all__ lists

cdeil:
This PR improves the model registries and __all__ lists a bit. See #2387.

I introduce a custom list class with currently 5 lines of code. I think this will grow a little bit and we will use it to generate very nice docs, or to offer the option to print a `Table` listing models and descriptions at runtime. It also offers another place where we can have serialisation logic / validation if deemed useful.

The other change here is to define a good `__all__`, so that the API docs appear in better order (currently was pretty random), and also I think it's easier to maintain explicit lists of models in a single file (before was duplicated in the `models/spatial.py` etc and `models.__init__`, and indeed there were inconsistencies, i.e. things we forgot to add to the registry.

### DIFF
--- a/gammapy/catalog/registry.py
+++ b/gammapy/catalog/registry.py
@@ -1,5 +1,12 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 from astropy.table import Table
+from .hess import SourceCatalogHGPS
+from .gammacat import SourceCatalogGammaCat
+from .fermi import SourceCatalog3FGL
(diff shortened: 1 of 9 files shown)


cdeil:
This is what it looks like now:

Screenshot 2019-09-27 at 09 54 54

adonath:
Thanks @cdeil! I agree it's a better sorting. I also think it's good idea to introduce the `ModelRegistry`, even if it's only a tiny wrapper now..

PR #2410, opened 2019-09-27
Implement `ConstantTemporalModel`

fabiopintore:
This PR aims at including a method to create constant lightcurve models into `~gammapy.modeling.models.temporal`.
The scheme is quite simple, as it takes as input a time array and a normalization. The output is an `~astropy.table` object which can be given to `LightCurveTemplateTemporalModel`.


adonath:
Thanks @fabiopintore! After #2400 I think we should rather implement a `ConstantTemporalModel` instead of `.from_constant()`. Maybe we can discuss the details offline...

cdeil:
Yesterday when renaming the models and introducing an (empty for now) `TemporalModel` base class I was also thinking about this.

I'm still not sure what we want here, all options seem to have small pros and cons:

1. use `temporal_model = None` (default, most of the time) to mean constant temporal model, and when sampling just sample uniformly within the GTIs.
2. add `ConstantTemporalModel` class.
3. add `LightCurveTemporalModel.from_constant`

I think solution 1 is fine?

Solution 2 is fine as well. It does pose questions concerning whether that model should have a norm or not (we intentionally didn't put a `ConstantSpatialModel` so far), and if the unit should be empty (a factor, no unit added), or a flux unit.

Solution 3 is OK in principle, but has some issues, e.g. whether the model should take a `time` array and whether that should be Time objects or floats (see current implementation). Overall I think we could just not add it, and instead show a docs example how to make a `Table` with 3 lines of code and then pass this to `LightCurveTemporalModel` to turn it into a model? The classmethod doesn't offer much?

There is the precedent of ctools having added a constant temporal model, and corresponding serialisation: http://cta.irap.omp.eu/ctools/users/user_manual/models_temporal.html
So from that POV, the easiest might be to do the same, and then follow their lead also wrt serialisation. Presumably it's done such that if no temp model is in the serialised spec, then it's assumed constant?

@adonath @QRemy @registerrier - Thoughts?

- Do we want `ConstantTemporalModel`?
- Do we want `ConstantSpatialModel`? (to handle e.g. the isotropic diffuse case)
- Should we use `None` to handle the "no temporal model" case, or should we always create and use `ConstantTemporalModel` objects?

Would be nice to sort this out quickly ...

My suggestion is to default to what Gammalib / ctools does and then to only put something different if we find that it can be done in a better way. I remember these things were discussed and considerations of use cases such as diffuse emission made etc. years ago, and I partly was involved, but don't remember all the arguments. Can someone look and comment how they use the constant temporal model and what units they have for the LC and constant temp model norms, i.e. whether it's always a flux or always only a multiplier, or a mix?

registerrier:
I think that `ConstantSpatialModel` and `ConstantTemporalModel` are fine. 
I don't think a `SkyModel` should contain a `TemporalModel` a priori. This is useful mostly for event sampling, but more or less never for analysis.

gammapy/modeling/models/temporal.py
```diff
@@ -11,14 +11,94 @@
...
+        time : array_like
+            Time since the ``reference`` time.
+
+        Returns
+        -------
+        norm : float
+            Mean norm
+        """
+
+        return np.ones(len(time), float) * self.norm.value
```
adonath: You can use `np.ones_like(time) * self.norm.value`, should be a bit simpler...

gammapy/modeling/models/temporal.py
```diff
@@ -11,14 +11,94 @@
...
+
+    def sample_time(self, n_events, t_min, t_max, t_delta="1 s", random_state=0):
```
adonath: Note the `t_delta` is not needed anymore, please remove...

gammapy/modeling/models/temporal.py
```diff
@@ -11,14 +11,94 @@
...
+            
+        Parameters
+        ----------
+        n_events : int
+            Number of events to sample.
+        t_min : `~astropy.time.Time`
+            Start time of the sampling.
+        t_max : `~astropy.time.Time`
+            Stop time of the sampling.
+        t_delta : `~astropy.units.Quantity`
```
adonath: Also remove `t_delta` form the docstring...

gammapy/modeling/models/temporal.py
```diff
@@ -11,14 +11,94 @@
...
+        -------
+        time : `~astropy.units.Quantity`
+            Array with times of the sampled events.
+        """
+
+        time_unit = u.second
+
+        t_min = Time(t_min)
+        t_max = Time(t_max)
+        t_delta = u.Quantity(t_delta)
```
adonath: Just remove, it's unused...

gammapy/modeling/models/tests/test_temporal.py
```diff
@@ -107,6 +111,24 @@ def test_time_sampling():
...
+def test_ConstantTemporalModel():
+    norm = 10.0
+    temporal_model = ConstantTemporalModel(norm)
+
+    t_ref = "2010-01-01T00:00:00"
+    t_min = "2010-01-01T00:00:00"
+    t_max = "2010-01-01T08:00:00"
+
+    sampler = temporal_model.sample_time(
+        n_events=2, t_min=t_min, t_max=t_max, random_state=0, t_delta="10 min"
```
adonath: Again just remove `t_delta`...

gammapy/modeling/models/tests/test_temporal.py
```diff
@@ -107,6 +111,24 @@ def test_time_sampling():
...
+    )
+
+    sampler = u.Quantity((sampler - Time(t_ref)).sec, "s")
+
+    assert len(sampler) == 2
+    assert_allclose(sampler.value, [15805.82891311, 20597.45375153], rtol=1e-5)
+
```
adonath: Please add a short test for `.evaluate_norm_at_time()`

gammapy/modeling/models/temporal.py
fabiopintore: OK.

gammapy/modeling/models/temporal.py
fabiopintore: OK.

gammapy/modeling/models/temporal.py
fabiopintore: OK.

gammapy/modeling/models/tests/test_temporal.py
fabiopintore: OK.

gammapy/modeling/models/tests/test_temporal.py
fabiopintore: Ok.

adonath:
Thanks @fabiopintore!

cdeil:
Added the `ConstantTemporalModel` to the registry: https://github.com/gammapy/gammapy/commit/7fbde7d77445f2ad4c231fe9f5b1b7f0b3d83638

PR #2411, opened 2019-09-27
Modified spectrum_simulation.ipynb to take background from the IRF

registerrier:
This PR removes the usage of a power law background model in `spectrum_simulation.ipynb` and replaces it with the background extracted from the IRF. This removes the usage of `SpectrumEvaluator` in the notebook.

I have not added a PSF correction to limit the complexity of the notebook.


adonath:
Thanks @registerrier!

PR #2412, opened 2019-09-27
Remove model XML serialization

QRemy:
remove current xml serialization, it will be re-introduced latter 
updating from_dict/to_dict on models to support ctools/scitools conventions


adonath:
Yes, +1 to remove. If there is anything useful to keep we can get it back from the git history anytime...

Issue #2283, opened 2019-07-15
Use Algolia DocSearch for Gammapy docs


joleroi:
I suggest to use [DocSearch](https://community.algolia.com/docsearch/) instead of the build-in sphinx search functionality for the gammapy online docs. 

You can test how the search window behaves with DocSearch [here](https://reactjs.org/docs/getting-started.html)

I've recently done this for another website so I'd be willing to take over the task. It's not a big deal, you basically have to register on the DocSearch web page and they scrape the website for you and send you a piece of JS code that you have to inject to the website. I already contacted them and they'd be willing to support gammapy.

Oh I forgot to mention - it' free ...

cdeil:
@joleroi - Thanks for the suggestion!

Do you have an example of a Python Sphinx docs project, how the results look there?
I guess just trying it out and seeing what it looks like in a PR and how much better it is to what we have now is quite some work for you? However, I trust that what they deliver is good, probably we can just assume that. That's the advantage.

---

My main concern with putting this for the Gammapy docs is user privacy - probably what they do is to track users. That's a discussion we should have. E.g. it's very common to 
 have https://docs.readthedocs.io/en/stable/guides/google-analytics.html on documentation pages, e.g. Astropy does have this, and they use the analytics sometimes to see how many users they have and what users look at. That's extremely valuable. But also it's bad that Google tracks so much and depending on how Gammapy is used in the future by projects it might even be illegal to track users.

There's also the concern that this introduces another cloud service that could go down or break in the future. But presumably they have super-high uptime, and one could fall back to the Sphinx search we have at the moment? So that's probably fine.

---

An alternative to this DocSearch is to have a custom Google search to have better search results. That's e.g. what https://scikit-learn.org/stable/ does.

Either way - the main question is: take better search results and analytics, at the cost of user tracking?

cdeil:
@joleroi - we discussed it. Others don’t seem concerned about user tracking at all. Let’s do it!

joleroi:
@cdeil thanks for your thoughts.

I was going to suggest that I open a PR for you to test the new functionality and then talk about user tracking again (after you've seen the shiny feature). But if you already agreed on this - fine. I'll proceed.

cdeil:
Yes, let's try it out.

If we get a bit of user tracking anyways, it might be nice to just get all the benefits we can and to have analytics about users, i.e. how many people visit gammapy.org and docs.gammapy.org and which docs pages they look at?

[...]

PR #2413, opened 2019-09-27
Revert introduction of docsearch

joleroi:
This adresses #2283 

It was a very nice git revert exercise!


cdeil:
@joleroi - Thank you!

The `html_css_files` was a good edit I think. If you put that back it would be useful, but it's not required.

For your ML package you do or plan to use Algolia?

Issue #2286, opened 2019-07-16
Maps API uses unintuive and unnecessary abbreviations


maxnoe:
E.g. binsz for the pixel size.

When I read it, I assumed this would be the z-values for the bins, e.g. the data.
I think gammapy should have a clear api that avoids unnecessary abbreviations like this.

Also it does not accept simple stuff like setting data with units.
It complains that one should set `unit` and data without units. Why not simply do that, if data is a quantity?

cdeil:
Initially gammapy.maps was written using mostly Numpy arrays with very little Quantity support. Then MapCoords was added, and some quantity support was added.

@adonath - What do you suggest for v0.14 / v1.0? Does it make sense to make a list of things to change / improve for the maps API, or do you just want to do some cleanup over the next months without up-front plan/discussion?

adonath:
I'm in the process of cleaning up the Map API. @MaxNoe and @registerrier, any proposal for a better name instead of `binsz`?

registerrier:
`binsize` or `binwidth` is more explicit. Probably the first one, `binsize`, is more natural and less confusing since `width` is already used for the `Map` size.

registerrier:
We have not progressed a lot here. Now that we are supposed to have a stable API, I suppose we should close the issue or defer it to the second LTS.

maxnoe:
Is there already a decision how to handle deprecations? E.g. we could add a `binsize` argument now that supports units and add the unit support, keeping the unitless `binsz` but raising a deprecation warning if that is given pointing to the new improved API.

```python
def __init__(self, binsz=None, ..., binsize=0.5 * u.deg):
    if binsz is not None:
        warnings.warn(GammapyDeprecationWarning, "Using binsz is deprecated and will be removed in gammapy 2.0, use binsize instead")
        binsize = u.Quantity(binsz, u.deg)
```

adonath:
I think we can make refactoring of `gammapy.maps` part of the v2.0 roadmap.

Issue #1788, opened 2018-09-13
CrabSpectrum as factory function


cdeil:
I was using `gammapy.spectrum.CrabSpectrum` again today, and was puzzled by the API. It's a class with only one method `__init__` which is a factory function, that stores the return value on the instance since `__init__` can't return things:
http://docs.gammapy.org/dev/_modules/gammapy/spectrum/crab.html#CrabSpectrum
So callers always do `CrabSpectrum('hess').model`, never use the `CrabSpectrum` object for anything else.

Looking over #777 it seems there might be a use case to have the class and extend it, or there might not.

This is a reminder issue for v0.9 to discuss this point again and to consider changing this to a factory function `crab_spectrum` that does what `CrabSpectrum.__init__` does now.

We could also consider moving this to the `gammapy.spectrum.models` namespace.

adonath:
I think introducing a factory method is just fine. The weak motivation for having the `CrabSpectrum` class was to keep a reference for the model in the `.reference` attribute. But it's probably sufficient to keep the reference e.g. by the variable name...

cdeil:
Since Python is dynamic we could just do this?
```
model = ...
model.meta = {"reference": reference}
return model
```
I think `reference` is sometimes a (bad) model parameter name for the reference energy, and we now expose it on the model class. So we have to be careful concerning namespace collisions?

I'll wait for @registerrier to comment, then send a PR for this soon.

adonath:
@deil My point was even that we probably don't need to define this meta data for now. Users can take care of it themselves, by choosing good variable names: `hess_crab_pl = crab_spectrum("hess_pl")`

cdeil:
OK, so the task here is to refactor `CrabSpectrum` into a function,  and to not attache the reference to the returned model as is currently done [here](https://github.com/gammapy/gammapy/blob/50b2b8ae43456011af0575d45dba0c5a9aeae1c8/gammapy/spectrum/crab.py#L137).

Remaining decision needed IMO is whether the function name should be `crab_spectrum` or `make_crab_spectrum`, and whether to expose it in the `gammapy.spectrum` namespace or in the `gammapy.spectrum.models` namespace.

My recommendation would be to call it `make_crab_spectrum`, to expose it in `gammapy.spetrum.models`, and to keep the code location and filename as-is: `gammapy/spectrum/crab.py`.

(similarly for #2213 I would suggest to call it `make_cosmic_ray_spectrum`)

@adonath @registerrier - What's your preference?

[...]

PR #2414, opened 2019-09-27
Add Model.create and tests for all models

cdeil:
This PR adds a first, super rough version of a `make_model` factory function.

I mainly introduced it because I wanted to start adding tests for all models, to see if they serialise properly or not. But I imagine apart from being used in tests, this is also the function that will evolve to the common end-user convenience API to make models.

Clearly there's many options: I exposed `make_model` in the `gammapy.modeling` namespace where `Model` and `Models` is also (could have been `gammapy.modeling.models`). I made it a function, whereas in gammapy.maps we have `Map.create`, i.e. this pattern to have the factory functions attached to base classes. I would prefer to go uniform and put `gammapy.maps.make_map`, and also break the API a bit to support quantities better (see #2286) in v0.15. One can argue it's better to break in a clear way, also changing function name, instead of only changing arguments and break some callers, but really most users will have to update.

So if there's strong opposition here, I will change to `Model.create` today and bring this up again later. Otherwise I'd suggest we merge this in now. I have to leave shortly, but tonight I could continue a bit here. E.g. one change I would like to make is from
```
spectral_model = make_model("PowerLaw2SpectralModel", amplitude="1e-10 cm-2 s-1", index=3)
```
to
```
spectral_model = make_model("spectral", "powerlaw2", amplitude="1e-10 cm-2 s-1", index=3)
```
and to change to `PowerLaw2SpectralModel.tag = "powerlaw2"` (and simliarly for all other tags).
I didn't manage to write down a proposal for tag names and there's many options (snake case, camel case, underscores, keep full classname). Preferences?

@adonath @registerrier @QRemy - thoughts?

### DIFF
--- a/gammapy/modeling/model.py
+++ b/gammapy/modeling/model.py
@@ -3,7 +3,7 @@ import copy
 import astropy.units as u
 from .parameter import Parameters
 
-__all__ = ["Model", "make_model"]
+__all__ = ["Model"]
 
 
 class Model:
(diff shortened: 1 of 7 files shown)


gammapy/modeling/model.py
```diff
@@ -59,3 +59,27 @@ def from_dict(cls, data):
         for parameter in init.parameters.parameters:
             setattr(init, parameter.name, parameter)
         return init
+
+
+def make_model(tag, *args, **kwargs):
```
adonath: As I already said, I care more about consistency, than the actual name. So I would prefer to introduce it as `create_model()` for now, to keep the consistency with the existing `create_crab_spectral_model` and `create_cosmic_ray_spectral_model` (admittedly those were decided without extensive discussion by @JouvinLea , but following one of your comments https://github.com/gammapy/gammapy/issues/1788#issuecomment-499476024). It would be weird to have  `make_model()` showing up in the docs here: https://docs.gammapy.org/dev/modeling/index.html#id1. It would just introduce further diversity in the naming scheme in the model namespace for no good reason. It would be even more weird, in the light of the extensive model re-naming we did for v0.14. If you feel strongly about the `make` in the name, then please also rename the other `create_` factory functions at the same time. Even if we don't have it consistent everywhere in Gammapy let us try to keep at least the newly introduced models namespace consistent.

adonath:
Concerning the naming scheme for tags I have a tiny preference for "kebab-case" or dash separation of words. But this is just my personal taste. I like when strings look different than variable or class names in the code. I have the same preference for string arguments in functions (e.g. `f(method="constant-width")` vs. `f(method="constant_width")`, where the latter looks like a variable name). Or dictionary keys such as `data["SpectralModel"]` vs. `data["spectral-model"]`, where the first looks like a class name and could be confused with `data[SpectralModel]` if you don't look closely. In practice it does not matter as much, because you have code highlighting etc.  YAML keywords are like dictionary keys to me, hence my preference for the dashed separation of words.

That's what it would look like:
`power-law`, `power-law-2`, `super-exp-cutoff-power-law-3fgl`, `phase-curve-template`

cdeil:
I changed to `Model.create` in c0f9714eeee5f0bf93c9da5e41ba0a3eb45e879a for consistency with `Map.create`. I do plan to continue to improve it and to maybe do the tag names as you propose in the next days. But for now I'm merging this in.

PR #2415, opened 2019-09-27
Fix gammapy/modeling/tests/data/make.py

QRemy:
remove selection="simple" from datasets.to_yaml() call


adonath:
Following up https://github.com/gammapy/gammapy/commit/65c113f503f159bfaa079cfcdf5fd7a55fd2f745. Thanks @QRemy!

PR #2416, opened 2019-09-27
Minor cleanup of Analysis class #4

Bultako:
This PR is a follow up to #2407. It implements the following changes agreed in the developers call this morning:

- Generation of settings sections YAML code in the documentation is generated from anew `docs.yaml` file and no more from comments in the `schema.yaml`. The `docs.yaml` file is validated in tests and include some comments to guide the user. So we have now five YAML files to maintain including `schema.yaml`, I think there's room for improving this in the future..
- Rename `config.print_help()` method into `config.help()` and display the whole content of the `docs.yaml` file or a specific section of the settings in that file.
- Remove example code in the docstring of the `Analysis` class and put a link to the documentation instead.
- Remove parameter model from the config settings. Fetch the model with the `get_model()`method providing the filename of its YAML serialization and attach it to the datasets.
- Update of the documentation, improvement of internal validation methods and code refactoring.

### DIFF
--- a/docs/scripts/index.rst
+++ b/docs/scripts/index.rst
@@ -29,8 +29,9 @@ an IPython console or a notebook.
 
 .. code-block:: python
 
-    >>> from gammapy.scripts import Analysis
-    >>> analysis = Analysis()
+    >>> from gammapy.scripts import Analysis, AnalysisConfig
+    >>> config = AnalysisConfig()
+    >>> analysis = Analysis(config)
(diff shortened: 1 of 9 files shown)


adonath:
It's not exactly related to this PR, but when I looked at the `docs.yaml` in your code, I was wondering whether the comments couldn't be parsed using `pyyaml`. After a quick search I found this is actually not supported (see https://github.com/yaml/pyyaml/issues/90). It's a minor thing but I guess we should keep this in mind and maybe document for users, that comments in their config files are not round-tripped but dropped.

Issue #2402, opened 2019-09-26
Coordinate frame `fk5` not handled properly in `SpatialModel.evaluate_geom()`


ctrichard:
Dear all,

I have trouble using SkyDiffuseMap from a 2D fits map

```

mod = Map.read("FakeModelMap_CrabGauss_XY83.63_22.01_SigmaXY0.00_0.00.fits")
print(sum(sum(mod.data)))

```
output  : 
0.9999999588496848

-----------------------------------------------------
```

spatial_model = SkyDiffuseMap(mod,norm=1)
print(spatial_model.evaluate(22.5 * u.deg,83 * u.deg,1 * u.Unit('cm-2 s-1')))

```

output  : 
[0.] 1 / sr
---------------------------------------------------

If I use this model as imput of a skymodel for a MapDataset then the npred is always zero from this model.
The BackgroundModel gives good npred values.

Did I do something wrong ?

Cheers,
Cyril.

adonath:
Can you please execute the following:
```python3
mod = Map.read("FakeModelMap_CrabGauss_XY83.63_22.01_SigmaXY0.00_0.00.fits")
print(mod.geom)
```
...and post the output here? This will be helpful for the discussion.

ctrichard:
Thx for looking into this !

Here is the output :  
```
Invalid value BUNIT='None' in FITS header. Setting empty unit.

WcsGeom

	axes       : lon, lat
	shape      : (200, 200)
	ndim       : 2
	coordsys   : CEL
	projection : SFL
	center     : 83.6 deg, 22.0 deg
	width      : 4.0 deg x 4.0 deg


```

adonath:
In the specific example you posted above the order of the coordinates is not correct. What does the following code give:

```python3
spatial_model = SkyDiffuseMap(mod,norm=1)
print(spatial_model(lon=83 * u.deg, lat=22.5 * u.deg))
```
?

ctrichard:
Here is the ou tput :
`[0.] 1 / sr`

adonath:
Ok, this makes it hard to debug remotely. Can you provide me with the specific FITS file somehow?

ctrichard:
May I answer  this way ?

The files are the fits map and a more detailed notebook

Thanks

Le jeu. 26 sept. 2019 à 14:53, Axel Donath <notifications@github.com> a
écrit :

adonath:
[...]

PR #2417, opened 2019-09-28
Remove FK5 option in spatial models

cdeil:
This is a follow-up PR to  #2406 which addressed #2402.

I revert the option and handling of "fk5" as a possible frame for spatial models.

Trying to handle "fk5" ourselves isn't a good strategy, because then we'd have to handle "fk4" as well and basically all the frames astronomers use and that Astropy already has built-in with 1000s of lines of code.

So either we change spatial models to use SkyCoord and Astropy Frame (like we did in `regions`), or we keep the limit to only "icrs" and "galactic" as the two valid options (like e.g. Fermi tools or Gammalib does). Either way seems acceptable to me, with pros and cons. Especially if we keep using Quantity for most models, and Time for time models, then using SkyCoord for spatial models makes sense, we'd just have to find a way to map it to Parameter objects.

This PR is unfinished, there's two fails:
https://gist.github.com/cdeil/7b222ec808fa860af8b9c7ba6fd0a22b

I'm not sure why - is it because there's an example spatial template with FK5 frame involved? or do we have to call some frame -- coordsys tranlation function from gammapy.maps or adjust an API there?

@adonath - Could you please take over?

### DIFF
--- a/gammapy/maps/geom.py
+++ b/gammapy/maps/geom.py
@@ -168,9 +168,9 @@ def get_shape(param):
 
 
 def coordsys_to_frame(coordsys):
-    if coordsys in ["CEL", "C"]:
+    if coordsys in ["CEL", "C", "fk5", "fk4", "icrs"]:
         return "icrs"
-    elif coordsys in ["GAL", "G"]:
+    elif coordsys in ["GAL", "G", "galactic"]:
         return "galactic"
     else:
         raise ValueError(f"Unrecognized coordinate system: {coordsys!r}")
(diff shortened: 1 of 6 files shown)


adonath:
Thanks @cdeil!

I attached one commit, where I moved all the coordinate frame handling to `coordsys_to_frame()` and I also simplified `skyoord_to_lonlat()`. This is still not a  nice solution but at least the coordinate frame handling is in a single place now.

I think in general we should adapt the `Map` API to get rid of the `GAL` and `CEL` keywords and just use `astropy.wcs.utils.celestial_frame_to_wcs` to build the `WCS` objects in first place. I think Astropy should be self-consistent with coordinate transformations like `SkyCoord.to_pix(wcs)`. The limitation to `CEL` and `GAL` in `Map.create()` actually complicates the code, because we cannot use `astropy.coordinates.SkyCoord` and `WCS` consistently.  

The actual problem with #2402 arose when maps from "outside" (not created by Gammapy) enter the analysis. In `Map.create()` we only support two coordinate systems, while `Map.read()` supports whatever coordinate system is defined in the wcs. As long as we stick with Astropy frames in our API this should just work out of the box and our code will support more coordinate systems and becomes even simpler.

PR #2420, opened 2019-09-30
Fix spatial model frame in source catalogs

QRemy:
add default frame="galactic" in hess et gammacat source catalogs

### DIFF
--- a/gammapy/catalog/tests/test_fermi.py
+++ b/gammapy/catalog/tests/test_fermi.py
@@ -171,7 +171,8 @@ class TestFermi3FGLObject:
 
     @pytest.mark.parametrize("ref", SOURCES_3FGL, ids=lambda _: _["name"])
     def test_spatial_model(self, ref):
-        self.cat[ref["idx"]].spatial_model
+        model = self.cat[ref["idx"]].spatial_model
+        model.frame = "galactic"
 
     @pytest.mark.parametrize("ref", SOURCES_3FGL, ids=lambda _: _["name"])
     def test_sky_model(self, ref):
@@ -353,7 +354,8 @@ class TestFermi3FHLObject:
 
     @pytest.mark.parametrize("ref", SOURCES_3FHL, ids=lambda _: _["name"])
     def test_spatial_model(self, ref):
-        self.cat[ref["idx"]].spatial_model
+        model = self.cat[ref["idx"]].spatial_model
+        model.frame = "galactic"
 
     @pytest.mark.parametrize("ref", SOURCES_3FHL, ids=lambda _: _["name"])
     def test_sky_model(self, ref):
--- a/gammapy/catalog/tests/test_fermi.py
+++ b/gammapy/catalog/tests/test_fermi.py
@@ -169,10 +169,35 @@ class TestFermi3FGLObject:
         assert_quantity_allclose(dnde, ref["dnde"])
         assert_quantity_allclose(dnde_err, ref["dnde_err"])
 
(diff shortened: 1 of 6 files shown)


cdeil:
@QRemy - Thanks!

For gamma-cat, probably we should put RA / DEC and icrs frame models, as it's more common?
HGPS is explicitly a Galactic catalog, so there Galactic frame makes sense.

Fix or improve tests?

@QRemy - You decide what to put for now. For sure this is an important fix and should go into v0.14, i.e. before 4pm today.

QRemy:
I prefer to let this in galactic as most of the source are in the Galactic plane.
I have added a test to check the frame.

cdeil:
@QRemy - Thanks! I'll have a look shortly and merge this in.

cdeil:
@QRemy - There was another bug which I fixed earlier for the Fermi spatial models: ec0843fe05e83d078f3cdc8ae33929cf342e8f89

Now I see that there we also fill GLON / GLAT but don't set the frame correctly.

Could you please add a test case for those also, and fix them?

gammapy/catalog/tests/test_fermi.py
```diff
@@ -171,7 +171,8 @@ def test_spectral_model(self, ref):
 
     @pytest.mark.parametrize("ref", SOURCES_3FGL, ids=lambda _: _["name"])
     def test_spatial_model(self, ref):
-        self.cat[ref["idx"]].spatial_model
+        model = self.cat[ref["idx"]].spatial_model
+        model.frame = "galactic"
```
cdeil: Put assert and ==

gammapy/catalog/tests/test_fermi.py
```diff
@@ -353,7 +354,8 @@ def test_spectral_model(self, ref):
 
     @pytest.mark.parametrize("ref", SOURCES_3FHL, ids=lambda _: _["name"])
     def test_spatial_model(self, ref):
-        self.cat[ref["idx"]].spatial_model
+        model = self.cat[ref["idx"]].spatial_model
+        model.frame = "galactic"
```
cdeil: Put assert and ==

cdeil:
@QRemy - Ready for final review / merging? Did you check all `SpatialModel` in `gammapy/catalog`?

QRemy:
Yes they are not defined for the others catalogs

cdeil:
I added tests to cover the 3FGL spatial models in ebbcaaf3d .

There's plenty of simple improvements coming next for catalog spatial models, e.g. filling elongation parameters, and parameter errors, and better tests. But I think we should do that after the v0.14 release.

@adonath - There's one more issue here. The 3FGL template map models come with fk5 frame. Is that OK for evaluation? Or should that frame be changed to ICRS on read? (If not easy to resolved, can be done later). Assigning this issue to you know. (Suggest to wait for CI to pass, and if tests fail, reduce assert precisions)

adonath:
Thanks @cdeil! It's not documented yet, but after #2417 the evaluation of models in the `fk5` frame,
should work correctly.

cdeil:
Does this mean we do frame transformations during model evaluation? That should be avoided for performance reasons. No need to discuss now ... let's talk and improve spatial models in the next weeks when you are back.

adonath:
@cdeil Yes we do, but the transformed coordinates are cached, so there is no significant effect on the performance now...

cdeil:
Doesn't need a changelog entry ... issue was introduced and fixed during v0.14 dev cycle.

PR #2424, opened 2019-09-30
High-level interface: improve test coverage

Bultako:
This PR is a follow-up of #2421
In general, it adds tests for missing parameters and non-valid options that `schema.yaml` validation cannot detect.


adonath:
Thanks @Bultako!

PR #2435, opened 2019-10-03
Add modeling notebook with model plot examples

cdeil:
This PR adds a `tutorials/modeling.ipynb` tutorial notebook. The idea is that it will be similar to the maps intro tutorial -- showing how to work with modeling & model classes, but not execute analyses.

It's not written here - scope, content will have to be discussed a bit first and then someone has to actually write it. But I wanted to move the existing plot examples from the spatial and spectral models here, because I agree it's better to have one "model gallery" as @adonath mentioned a few days ago, and to keep the auto-generated API docs small and simple. Especially with models there is value in having different models in the same document and to be able to compare them.

@adonath - OK to go in this direction? And to limit the scope of this PR to just moving the existing examples like I did, i.e. not actually writing the notebook content now?


docs/tutorials.rst
```diff
@@ -24,6 +24,7 @@ Notebooks
 
    notebooks/first_steps.ipynb
    notebooks/intro_maps.ipynb
+   notebooks/modeling.ipynb
```
adonath: If the idea of the notebook is similar to the maps intro, then maybe rename it to `intro_models.ipynb`? I think "modeling" maybe suggests, that there is fitting involved...

cdeil:
> In spectrum_simulation we have an example how to define a UserModel. Maybe move this to the models notebook as well?

Prefer to leave this to a later PR.

Am considering writing a PIG or wiki page to outline the docs and tutorial organisation we want for v1.0 and to have a good discussion in Oct, and then a clear plan to implement it in Nov.

> Add the notebook to the tutorials/notebooks.yaml?

Was already the case

> If the idea of the notebook is similar to the maps intro, then maybe rename it to intro_models.ipynb? I think "modeling" maybe suggests, that there is fitting involved...

I changed to `models.ipynb`, and I also changed `intro_maps.ipynbs` to `maps.ipynb`.
IMO there's no clear separation which notebooks are "intro" and which aren't, most tutorials are and will be intro. So I think not putting "intro" on some of them is more consistent.
Concerning whether it should be "modeling" or "models" I also was 50:50 when starting this PR. Which name is better will depend on how far we go in this tutorial concerning modeling framework explanation, or if we do add some fit to show how models and results come out after the fit. But OK to put "models" for now, and probably we'll limit the scope of this notebook to be the "built-in model gallery", and possibly to show how to create user-defined models.

@adonath - OK to put like this for now?
(if you want something changed, please just go ahead with a follow-up commit when merging this in)

cdeil:
@adonath ( or maybe @Bultako ?) - do you understand or can you reproduce this error?
```
NotImplementedError: Plotting the grid for the minor ticks is not supported.
```
https://travis-ci.org/gammapy/gammapy/jobs/593438153#L1344

(I can't reproduce this locally, and don't really understand it.)

For some reason, we pick up the `astropy.visualization.wcsaxes` axis, and there `plt.grid` fails?

I can see how this might happen ... in [SpectralModel.plot](https://docs.gammapy.org/0.14/_modules/gammapy/modeling/models/spectral.html#SpectralModel.plot) we use the pattern we use everywhere to get the axis via `plt.gca()`.
```
ax = plt.gca() if ax is None else ax
```
So if the sky image plotting from above in the notebook creates some non-standard axis, we will keep re-using that?
On the other hand, different cells in a Jupyter notebook seem to be decoupled from previous cells, i.e. they never overplot the content that was plotted there, no?
(e.g. the other spectral plots in that notebook and everywhere else in Gammapy work fine).

Thoughts?

@adonath - could you please also look at https://github.com/astropy/regions/issues/296 and comment? Maybe we should clean up our MPL usage pattern in Gammapy also a bit, especially to avoid this accidental mixing of old wcs axis plots with later normal plots?

PR #2441, opened 2019-10-05
Add GTI export in datasets

registerrier:
GTIs have been added to datasets but were not exported in the `SpectrumDatasetOnOff.to_ogip()` and `MapDataset.write()` methods. They were also not read in `SpectrumDatasetOnOff.from_ogip()` and `MapDataset.read()` methods.

This PR solves this issue by introducing  `GTI.to_hdulist()` and `GTI.from_hdulist()` methods.

### DIFF
--- a/gammapy/cube/fit.py
+++ b/gammapy/cube/fit.py
@@ -589,6 +589,9 @@ class MapDataset(Dataset):
             )
             hdulist += mask_fit_map.to_hdulist(hdu="mask_fit")[exclude_primary]
 
+        if self.gti is not None:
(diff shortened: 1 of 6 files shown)


registerrier:
Thanks.  I introduced the `to_hdulist` and `from_hdulist` for what I saw as API consistency. It is a bit weird to create an `HDUList` for a single table indeed and the `.table` API should be sufficient. 
Yet the object being not a `Table`  but rather containing one I found that a specific API was better. 

We will see what to do with the API consistency globally as proposed in #2442 .

Merging now.

PR #2444, opened 2019-10-07
Remove `MapMaker` class

adonath:
This PR removes the `MapMaker` class and replaces it with the following pattern:
```
stacked = MapDataset.create(geom=geom)

for obs in observations:
    maker = MapMakerObs(obervation=obs, geom=geom)
    dataset = maker.run()
    stacked.stack(dataset)
```
As we now have the high-level analysis interface I think it's acceptable to write a few more lines of code in the 2nd level API, but gaining the possibility to fit e.g. background model norms before stacking. Or first doing the data reduction for a joint-analysis and only later stack the reduced data for a stacked analysis.

As the `MapMakerObs` (we should rename it soon) now actually produces a `MapDataset` this reduces a lot of code where create `MapDataset` object by hand in the tutorials.


cdeil:
From the description, that does seem like a good idea to me, having both MapMaker and Analysis that do that kind of data reduction seems like unnecessary duplication.
I guess a similar change would come in 1D spectral analysis?
Is a PIG still planned for data reduction, or is this just the clear way how data reduction should be done, and no PIG and design discussion is needed any more?

adonath:
@cdeil I think these are a few first clean-up steps. We might still need a PIG to plan and describe a general `Maker` and `MakerChain` system for Gammapy (which also includes configuration).

@registerrier I agree once we have the `MapDatasetOnOff`, we can proceed with the clean-up of the `MapMakerRing`. Let's discuss the detail in the call later.

adonath:
Thanks @cdeil and @registerrier for the quick approval.

PR #2446, opened 2019-10-08
Implement `WcsNDMap.stack()` method

adonath:
This PR implements a `WcsNDMap.stack()` method. The `.stack()` methode is meant to be used together with `WcsNDMap.cutout()` to handle the common use-case of processing a cutout of a map and the pasting the result into a larger map. To achieve this in a consistent manner a `WcsGeom.cutout_info` is added, which contains a reference to the parent geom as well as the parent and cutout slices. Later the info could be extended e.g. to also contain views e.g. into larger coordinate arrays.

### DIFF
--- a/gammapy/cube/fit.py
+++ b/gammapy/cube/fit.py
@@ -276,7 +276,7 @@ class MapDataset(Dataset):
                     evaluator.update(self.exposure, self.psf, self.edisp)
 
                 npred = evaluator.compute_npred()
-                npred_total.stack(npred, check=False)
+                npred_total.stack(npred)
 
         return npred_total
 
@@ -357,24 +357,24 @@ class MapDataset(Dataset):
         """
         if self.counts and other.counts:
             self.counts.data[~self.mask_safe] = 0
-            self.counts.stack(other.counts, weights=other.mask_safe, check=False)
+            self.counts.stack(other.counts, weights=other.mask_safe)
 
         if self.exposure and other.exposure:
-            self.exposure.stack(other.exposure, check=False)
+            self.exposure.stack(other.exposure)
 
         if self.background_model and other.background_model:
             bkg = self.background_model.evaluate()
             bkg.data[~self.mask_safe] = 0
             other_bkg = other.background_model.evaluate()
             other_bkg.data[~other.mask_safe] = 0
-            bkg.stack(other_bkg, check=False)
+            bkg.stack(other_bkg)
             self.background_model = BackgroundModel(bkg, name=self.background_model.name)
 
         if self.mask_safe is not None and other.mask_safe is not None:
             # TODO: make mask_safe a Map object
             mask_safe = Map.from_geom(self.counts.geom, data=self.mask_safe)
             mask_safe_other = Map.from_geom(other.counts.geom, data=other.mask_safe)
-            mask_safe.stack(mask_safe_other, check=False)
+            mask_safe.stack(mask_safe_other)
             self.mask_safe = mask_safe.data
 
         if self.psf and other.psf:
--- a/gammapy/cube/fit.py
+++ b/gammapy/cube/fit.py
@@ -273,7 +273,7 @@ class MapDataset(Dataset):
                 # if the model component drifts out of its support the evaluator has
                 # has to be updated
                 if evaluator.needs_update:
-                    evaluator.update(self.exposure, self.psf, self.edisp, self._geom)
+                    evaluator.update(self.exposure, self.psf, self.edisp)
 
                 npred = evaluator.compute_npred()
                 npred_total.stack(npred, check=False)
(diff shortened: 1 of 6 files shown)


gammapy/maps/wcsnd.py
```diff
@@ -669,24 +669,48 @@ def cutout(self, position, width, mode="trim"):
...
+            Other map to stack
+        weights : `WcsNDMap`
+            Other map to be used as weights.
+        check : bool
+            Check whether geoms are aligned.
+
+        """
+        if check:
+            if other.geom != self.geom:
+                if other.geom.cutout_info is None:
```
adonath: Note to myself: add check for the parent geometry.

gammapy/maps/wcsnd.py
```diff
@@ -669,24 +669,47 @@ def cutout(self, position, width, mode="trim"):
...
+    def stack(self, other, weights=None, check=True):
+        """Stack cutout into larger map.
+
+        Parameters
+        ----------
+        other : `WcsNDMap`
+            Other map to stack
+        weights : `~numpy.ndarray`
+            Other map to be used as weights.
+        check : bool
```
registerrier: Is the `check` argument really needed? Why not always check if the géométrie is really a cutout from the other?

gammapy/maps/wcsnd.py
```diff
@@ -669,24 +669,47 @@ def cutout(self, position, width, mode="trim"):
...
+        slices = geom_cutout.cutout_info["parent-slices"]
+        cutout_slices = Ellipsis, slices[0], slices[1]
 
-        geom = WcsGeom(c2d.wcs, c2d.shape[::-1], axes=self.geom.axes)
         data = self.data[cutout_slices]
 
-        return self._init_copy(geom=geom, data=data)
+        return self._init_copy(geom=geom_cutout, data=data)
+
+    def stack(self, other, weights=None, check=True):
```
registerrier: Since this is doing the same thing as `coadd` except that it works for a cutout why not name this `coadd_cutout`? `stack` is already used for datasets. This might be confusing since this method could be used for other usages than stacking, no?

gammapy/cube/fit.py
```diff
@@ -360,23 +357,24 @@ def stack(self, other):
         """
         if self.counts and other.counts:
             self.counts.data[~self.mask_safe] = 0
-            self.counts.coadd(other.counts, weights=other.mask_safe)
+            self.counts.stack(other.counts, weights=other.mask_safe, check=False)
```
registerrier: Why `check = False`?  Could there be cases where someone tries to stack datasets which are not cutouts?

gammapy/maps/wcsnd.py
registerrier: Another option is to add a test in `coadd` so that `if self.is_cutout_from(other)` the simple slice approach is used. Then you don't need to check in each `stack` method what to use.

gammapy/maps/wcs.py
```diff
@@ -179,13 +179,15 @@ class WcsGeom(Geom):
...
+    cutout_info : dict
+        Dict with cutout info, if the `WcsGeom` was created by `WcsGeom.cutout()`
     """
 
     _slice_spatial_axes = slice(0, 2)
     _slice_non_spatial_axes = slice(2, None)
     is_hpx = False
 
-    def __init__(self, wcs, npix, cdelt=None, crpix=None, axes=None):
+    def __init__(self, wcs, npix, cdelt=None, crpix=None, axes=None, cutout_info=None):
```
registerrier: Should `cutout_info` be an argument in the constructor? It could be added dynamically after the cutout is performed

gammapy/maps/wcsnd.py
adonath: Currently there is the problem, that we don't create necessarily all the cutouts from the same  parent geometry. The situation will improve a bit after #2450. Once #2450 is merged I'll rebase this PR and re-address you comment.

gammapy/cube/fit.py
adonath: I removed the `check` option...

gammapy/maps/wcsnd.py
adonath: I removed the `check` argument again...

gammapy/maps/wcsnd.py
adonath: I agree there is an overlap in functionality of `.stack()` and `.coadd()`. So `.coadd()` is the completely general, but slow method and `.stack()` is the method to be used in the  single use-case in combination with `.cutout()`. I have a minimal preference to not merge both methods into one for two minor reasons:
1. I like the mirrored API between `MapDataset` and `Map`, so `MapDataset.stack()`, calls into `Map.stack()` and `PSFMap.stack()` etc.
2. Adding the fast track to `.coadd()` complicates the code in `.coadd()` to a lot of cases, with checking for aligned geometries, checking if `.cutout_info` is set etc. I think it's easier to maintain the `.cutout()` and `.stack()` use-case in a separate place.

PR #2447, opened 2019-10-08
Add an `obs_info` table on datasets

registerrier:
This PR is a small addition to datasets classes (`SpectrumDataset`, `SpectrumDatasetOnOff`, `MapDataset`) to add the `obs_info` from the `DatastoreObservation` in the form of a `Table`. This allows to keep track of the parameters of the parent observation(s).

The `Table` offers easier usage than e.g. a `dict` in a `meta`. In particular, stacking is simply obtained with a `vstack` and allows for the set of `obs_info` of the observations that were reduced to create the dataset. A property is added on `Datasets` to stack the `obs_info` of all datasets in the list. 

This addition is useful to obtain the distribution of typical parameters .e.g. zenith angle, efficiency etc. It can be useful for dataset grouping and stacking for similar observation conditions.

### DIFF
--- a/gammapy/cube/fit.py
+++ b/gammapy/cube/fit.py
@@ -68,6 +68,9 @@ class MapDataset(Dataset):
         Mask defining the safe data range.
     gti : '~gammapy.data.GTI'
         GTI of the observation or union of GTI if it is a stacked observation
+    obs_info : `~astropy.table.Table`
+        Table listing informations on observations used to create the dataset
+        One line per observation for stacked datasets.
     """
 
     def __init__(
@@ -84,6 +87,7 @@ class MapDataset(Dataset):
         evaluation_mode="local",
         mask_safe=None,
         gti=None,
+        obs_info=None,
     ):
         if mask_fit is not None and mask_fit.dtype != np.dtype("bool"):
             raise ValueError("mask data must have dtype bool")
@@ -100,6 +104,7 @@ class MapDataset(Dataset):
         self.name = name
         self.mask_safe = mask_safe
         self.gti = gti
+        self.obs_info = obs_info
         if likelihood == "cash":
             self._stat = cash
             self._stat_sum = cash_sum_cython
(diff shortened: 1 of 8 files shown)


adonath:
Thanks @registerrier! I think in general it needs some more thought how we want handle this kind of information. I'm adding @Bultako to the discussion, as he might have an opinion of the provenance aspect of this.

While I completely agree this information is useful to have on the `MapDatatset`, I currently don't like that we introduce an attribute for specific information for pointed observations. E.g. it's unclear how to fill the `.obs_info` for a Dataset containing Fermi-LAT data. It's also not clear to me either, how this information is handled, when we split an observation into multiple time bins. Do you have a specific use-case in mind, where the `.obs_info` is needed? It might help to clarify how the information is handled. One obvious use-case might be event sampling, where the pointing info is required to write the info to the event file meta data...

registerrier:
The typical uses cases I was considering for keeping track of observation conditions of each dataset:
- stacking or grouping in consistent observations conditions (e.g. for spectral analysis) after dataset production. Typically you extract run by run spectra and you stack them later according to their zenith angle, efficiency etc.
- plotting the distributions of the observations conditions of the reduced dataset.
- more generally any type of post data reduction selection of datasets. 

This could be done in a different manner by keeping the `EventList.meta`  for instance on the `count` member. Yet, using a table offers a nice way to stack the various observation info.

I understand that a Fermi dataset would not contain such a table, but I don't see this as a problem. And if we are to introduce a `MapDataset` maker for Fermi data, we could probably build such a table.

Bultako:
I understand it's useful to have provenance info glued into the datasets serialization, historically it has been done in this way with `HISTORY` FITS headers and additional tables like this one for `obs_info`. But I think we have to be conscious that we cannot keep all provenance info in this way.

I'm ok with this as long as we consider this as a *good to have* (kind of helper functionality or shortcut) to access very specific info that we have clearly identified as a real need for pretty common use cases. I don't know if it's really useful to access the list of observations without knowing the parameters used in the dataset reduction process. If you think it is the case, I'd say go.  But please, do not extend too much this strategy for storing all provenance in the FITS (i.e. IRFs used, parameters used in the reduction process, energy bins, geometry, model, best fit params, etc.) or we will end up with a complex API for i/o and, worst of all, provenance stored in an unstructured way with no provenance model.

FYI
There's a PIG on provenance coming soon :)

registerrier:
We discussed this in the gammapy telecon today. 
We agreed that such a  feature is important but how this information is stored and loaded is still unclear. 
Closing the PR for now and opening issue #2452   to discuss more generally on the issue.

PR #2448, opened 2019-10-10
Improve 4FGL catalog support

QRemy:
-add 4FGL sources infos
-add spatial_model, sky_model, and lightcurve for 4FGL sources 
-fix lightcurve infos display for 3FGL and 4FGL sources


cdeil:
@QRemy - Thanks!

Comments based on
```
pytest -v -s gammapy/catalog/tests/test_fermi.py --cov gammapy.catalog --cov-report html
```
and a quick view at the code with PyCharm.

Please add the lines to generate the reference txt data files to `make.py`, and change the filenames to match the style of 3FGL (lower-case and with underscore).

Then copy & paste the `TestFermi3FGLObject.test_str` to `TestFermi4FGLObject.test_str`, change the sources to 4FGL, and get it to pass (fails for me locally, didn't look why).

For the spatial model, I see in the coverage that the `GaussianSpatialModel` case isn't tested - please add a test.

I think the `_info_position` method is 100% identical. Can you please pull this (and any other method if possible) into a base class `SourceCatalogObjectFermiBase` and inherit from it to avoid the copy & paste. We should try to reduce the code duplication, but keep it simple, i.e. it's a case-by-case decision what to put in the base class and not. But I think there are some quick wins, like `_info_position`, where we can avoid some code duplication without introducing much complexity.

Please push and leave another comment here, and I'll review again (or directly merge in).

QRemy:
I have defined a SourceCatalogObjectFermiBase used with 3FGL, 3FHL, and 4FGL 
-moved to base class:
     - __str__()
     - _info_basic()
     - _info_position()
     -_info_morphology()
     -_info_spectral_points()
     -sky_model()
     -is_pointlike()
- methods empty by default
     - _info_more()
     - _info_spectral_fit()
     - _info_lightcurve()

gammapy/catalog/tests/data/make.py
```diff
@@ -24,3 +30,5 @@
 open("gammacat_hess_j1813-178.txt", "w").write(str(cat["HESS J1813-178"]))
 open("gammacat_hess_j1848-018.txt", "w").write(str(cat["HESS J1848-018"]))
 open("gammacat_vela_x.txt", "w").write(str(cat["Vela X"]))
+
+files=["4FGL J0000.3-7355","4FGL J0001.5+2113","4FGL J0002.8+6217","4FGL J1409.1-6121e"]
```
cdeil: This `files` is unused, no? Remove.

gammapy/catalog/tests/data/4fgl_J0001.5+2113.txt
```diff
@@ -0,0 +1,69 @@
...
+10000.000  30000.000   7.739e-15         nan   8.847e-12     1.594e-16           nan     1.822e-13  True   1.770e-11     3.646e-13       0.0
+30000.000 300000.000   4.368e-16         nan   7.089e-12     1.312e-17           nan     2.130e-13  True   1.418e-11     4.260e-13       0.0
+
+*** Lightcurve info ***
+
+Lightcurve measured in the energy band: 100 MeV - 100 GeV
+
+Variability index : 920.677
+Significance peak (100 MeV - 100 GeV)    : 36.358
+Integral flux peak (100 MeV - 100 GeV)   : 2.05e-07 1 / (cm2 s) +- 6.98e-09 1 / (cm2 s) cm^-2 s^-1
```
cdeil: Please fix units here in the printout.

gammapy/catalog/tests/data/4fgl_J0001.5+2113.txt
```diff
@@ -0,0 +1,69 @@
...
+Time peak                                : 4.76e+08 s s (Mission elapsed time)
```
cdeil: Please fix units here in the printout.

gammapy/catalog/fermi.py
```diff
@@ -47,52 +48,469 @@ def compute_flux_points_ul(quantity, quantity_errp):
...
+        elif tevcat_flag == "E":
+            tevcat_message = "Extended TeV source (diameter > 40 arcmins)"
+        else:
+            tevcat_message = "N/A"
+        ss += "{:<16s} : {}\n".format("TeVCat flag", tevcat_message)
+        return ss
+
+    @abc.abstractmethod
+    def _info_more(self):
+        return "\n"
```
cdeil: Please put `pass` for abstract methods, not some return value like "\n" that's never used.

PR #2449, opened 2019-10-10
Fix and improve 2HWC catalog source models

QRemy:
add spatial_models and sky_models for 2HAWC catalogue

### DIFF
--- a/gammapy/catalog/hawc.py
+++ b/gammapy/catalog/hawc.py
@@ -2,7 +2,12 @@
 """HAWC catalogs (https://www.hawc-observatory.org)."""
 import numpy as np
 from astropy.table import Table
-from gammapy.modeling.models import PowerLawSpectralModel
+from gammapy.modeling.models import (
+    PowerLawSpectralModel,
+    SkyModel,
+    DiskSpatialModel,
+    PointSpatialModel,
+)
 from gammapy.utils.scripts import make_path
 from .core import SourceCatalog, SourceCatalogObject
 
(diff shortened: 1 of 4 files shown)


QRemy:
They describe the source model as a point source or a disk of fixed radius of 0.5, 1, 2 degrees by default, but they try to re-define this radius based on residual map:

> After the search, a residual map is generated and halo-like structures are visible around several sources modeled as point sources. This halo is used to define a tentative source radius for the secondary source model when fitting the energy spectrum (results presented in Table 3
of the next section). This radius should not be regarded as a definite measurement of the source extent but can nonetheless provide useful information on how much the
spectrum measurement depends on the source region definition.

I have added this comment in the sky_model docstring : The radius of the secondary source model is a rough estimate based on the residual excess above the point source model.  This radius should not be regarded as a definite measurement of the source extent.

adonath:
@QRemy and @cdeil For the hgps catalog we also have different spectral models and there we have chosen the pattern, that users can choose which model to return, see here: https://github.com/gammapy/gammapy/blob/master/gammapy/catalog/hess.py#L436

Maybe we just do the same here and add a `which` argument to `spectral_model()` and `sky_model()`? I guess the option would be maybe `point-like` and `extended`. If `extended` is not available, then an error is thrown...

@cdeil I would prefer not to hide `.spectral_model()` and `.spatial_model()`, because we don't do it for the other source catalogs either. Let's just go for consistency now and later re-think and change our pattern homogeneously, if we decide it's better...

cdeil:
@adonath - OK.
@QRemy - I'll finish this PR up tomorrow.

I noticed that the spectral index sign is filled incorrectly, I'll fix that in this PR also, changed the title to "Fix and improve" just now.

gammapy/catalog/hawc.py
```diff
@@ -127,14 +132,61 @@ def spectral_models(self):
...
-    # TODO: add spatial models
+    def _get_spatial_model(self, idx):
+        d = self.data
+        label = f"spec{idx}_"
+
+        r_0 = d[label + "radius"]
+        if r_0 != 0.0:
+            model = DiskSpatialModel(d["glon"], d["glat"], r_0, frame="galactic")
+        else:
+            model = PointSpatialModel(d["glon"], d["glon"], frame="galactic")
```
cdeil: Typo: you filled "glon" for "lat". I noticed by adding a test case checking the parameter value in the model.

cdeil:
@QRemy - Thanks!

I'm merging this in now, and open a reminder issue that we should add some API to get and set position errors on spatial models. (At the moment the position error information isn't filled yet)

PR #2450, opened 2019-10-10
Rename and refactor `MapMakerObs`

adonath:
This PR renames the`MapMakerObs` to `MapDatasetMaker` and refactors it to be used in the following way:

```
maker = MapDatasetMaker(geom=geom, offset_max=4 * u.deg)

for obs in observations:
    dataset = maker.run(obs)
```
This API change separates the data from the configuration of the maker, and allows to re-use the same object for multiple observations.

The new API makes it also possible to create e.g. a counts map only like so:

```
counts = Map.from_geom()
maker = MapDatasetMaker(geom=geom, offset_max=4 *u.deg)

for obs in observations:
    counts_obs = maker.make_counts(obs)
    counts.coadd(counts_obs)
```


gammapy/cube/make.py
```diff
@@ -33,196 +33,313 @@ class MapMakerObs:
...
+        Radial axis for psf map.
+    cutout : bool
+        Whether to cutout the observation.
     """
 
     def __init__(
         self,
-        observation,
         geom,
         offset_max,
```
registerrier: If we move the max offset cut to a later step, it becomes useless to have it as an argument here. 
Then `cutout_size=5*u.deg` could be the default and setting it to `None` would result in no cutout performed.

Or is it for a later change?

gammapy/cube/make.py
```diff
@@ -33,196 +33,313 @@ class MapMakerObs:
...
-    def _fov_mask(self, coords):
-        pointing = self.observation.pointing_radec
-        offset = coords.skycoord.separation(pointing)
-        return offset >= self.offset_max
+    def _cutout_geom(self, geom, observation):
+        if self.cutout:
+            cutout_kwargs = {
+                "position": observation.pointing_radec,
+                "width": 2 * self.offset_max,
+                "mode": "trim",
```
registerrier: Shouldn't the cutout mode be an argument?
If we are to perform ring background later on, we cannot use `trim`.

gammapy/cube/make.py
```diff
@@ -33,196 +33,313 @@ class MapMakerObs:
...
+            geom=geom,
             max_offset=self.offset_max,
-            exposure_map=self.maps["exposure_irf"],
+            exposure_map=exposure,
         )
-        self.maps["psf"] = psf_map
+        return psf
+
+    @lru_cache(maxsize=1)
+    def make_mask_safe(self, observation):
```
registerrier: OK. See the comment above regarding when to make the safe range.

gammapy/cube/make.py
adonath: Sure, I can make it an argument. What was the reason again, that we cannot use partially contained observations for the ring background?

registerrier: yes, if the observation is partially contained you're likely to get an unexpected behavior.

gammapy/cube/make.py
adonath: Maybe we wait for some more comments to #2451 and I make a follow up PR? But it would be definitely my preference to remove the mask safe computation...

gammapy/cube/make.py
adonath: Is there a more specific explanation, that I could put in the docstring? I guess technically it still works, but it comes with increased systematics at the edge of the image?

registerrier: Yes, it also increases the probability of having empty OFF pixels

adonath: I've added the `cutout_mode` option. I think general we should maybe re-discuss the default choice,  so that we compute the whole observation by default. I can make a separate issue for this....

gammapy/cube/make.py
adonath: Will be addressed in a follow-up PR...

adonath:
Thanks for the review @registerrier!

PR #2454, opened 2019-10-11
Add a norm to the EBL model

JouvinLea:
This PR adds a norm to the EBL model to give the possibility to the user to add a norm for the AbsorbedSpectralModel class and also to let it free in the fits.

It is also adding a small argument for the peek() method of FLuxPointDataset because for now the uncertainties are not handle properly and it crashed if you try to plot the error of this model.

### DIFF
--- a/gammapy/modeling/models/spectral.py
+++ b/gammapy/modeling/models/spectral.py
@@ -1422,13 +1422,23 @@ class AbsorbedSpectralModel(SpectralModel):
         parameter value for absorption model
     parameter_name : str, optional
         parameter name
+    alpha_norm: float
+        Norm of the EBL model, by default 1.
+    alpha_norm_frozen: bool
+        False if you want to have it as a free parameter
     """
 
     __slots__ = ["spectral_model", "absorption", "parameter", "parameter_name"]
     tag = "AbsorbedSpectralModel"
 
     def __init__(
-        self, spectral_model, absorption, parameter, parameter_name="redshift"
+        self,
+        spectral_model,
+        absorption,
+        parameter,
+        parameter_name="redshift",
+        alpha_norm=1.0,
+        alpha_norm_frozen=True,
     ):
         self.spectral_model = spectral_model
         self.absorption = absorption
@@ -1440,6 +1450,8 @@ class AbsorbedSpectralModel(SpectralModel):
 
         parameters = spectral_model.parameters.parameters.copy()
         parameters.append(par)
+        self.alpha_norm = Parameter("alpha_norm", alpha_norm, frozen=alpha_norm_frozen)
+        parameters.append(self.alpha_norm)
 
         super().__init__(parameters)
 
@@ -1449,9 +1461,11 @@ class AbsorbedSpectralModel(SpectralModel):
         # since it does not belong to the spectral model
         parameter = kwargs[self.parameter_name]
         del kwargs[self.parameter_name]
+        del kwargs["alpha_norm"]
 
         flux = self.spectral_model.evaluate(energy=energy, **kwargs)
         absorption = self.absorption.evaluate(energy=energy, parameter=parameter)
+        absorption = np.power(absorption, self.alpha_norm.value)
         return flux * absorption
 
     def to_dict(self):
--- a/gammapy/modeling/models/spectral.py
+++ b/gammapy/modeling/models/spectral.py
@@ -1410,7 +1410,22 @@ class Absorption:
 
 
 class AbsorbedSpectralModel(SpectralModel):
-    """Spectral model with EBL absorption.
+    r"""Spectral model with EBL absorption.
+       The absorption of the EBL is described as:
+
+    .. math::
+        \exp{ \left ( -\alpha \times \tau(E,z) \right )}
+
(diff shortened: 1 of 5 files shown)


gammapy/modeling/models/spectral.py
```diff
@@ -1422,13 +1422,23 @@ class AbsorbedSpectralModel(SpectralModel):
         parameter value for absorption model
     parameter_name : str, optional
         parameter name
+    alpha_norm: float
+        Norm of the EBL model, by default 1.
+    alpha_norm_frozen: bool
```
cdeil: @JouvinLea - Please remove this `alpha_norm_frozen` option.
Just put `frozen=True` in the `Parameter` call in `__init__` below.
If someone wants to change this, they can change to `model.alpha_norm.frozen = False` after `__init__`.
This is what we do for all other cases where parameters are usually frozen, e.g. the reference energies in spectral models.
We will improve the model init methods, but manually adding options like `alpha_norm_frozen` is not the way to go.

gammapy/modeling/models/spectral.py
```diff
@@ -1440,6 +1450,8 @@ def __init__(
 
         parameters = spectral_model.parameters.parameters.copy()
         parameters.append(par)
+        self.alpha_norm = Parameter("alpha_norm", alpha_norm, frozen=alpha_norm_frozen)
+        parameters.append(self.alpha_norm)
```
cdeil: Use this instead of two append calls:
```
parameters.extend([par, self.alpha_norm])
```
I think this works too and does the same thing, you can use that if you prefer:
```
parameters += [par, self.alpha_norm]
```

gammapy/modeling/tests/test_serialize_yaml.py
```diff
@@ -130,8 +130,8 @@ def test_datasets_to_io(tmpdir):
     assert dataset0.model.skymodels[1].name == "gll_iem_v06_cutout"
 
     assert (
-        dataset0.model.skymodels[0].parameters["reference"]
-        is dataset1.model.skymodels[1].parameters["reference"]
+            dataset0.model.skymodels[0].parameters["reference"]
+            is dataset1.model.skymodels[1].parameters["reference"]
```
cdeil: Please revert this extra indentation. We use 4 space indentation, you changed to 8 space indentation here for some reason.

gammapy/spectrum/flux_point.py
```diff
@@ -1289,21 +1289,23 @@ def residuals(self, method="diff"):
...
-    def peek(self, method="diff/model", **kwargs):
+    def peek(self, method="diff/model", plot_error=True, **kwargs):
         """Plot flux points, best fit model and residuals.
 
         Parameters
         ----------
         method : {"diff", "diff/model", "diff/sqrt(model)"}
             Method used to compute the residuals, see `MapDataset.residuals()`
+        plot_error : bool
+            False to not plot the errors of the model since for some of them it is not handle yet properly in Gammapy
```
cdeil: Please shorten description to:
```
Plot error band?
```

gammapy/spectrum/flux_point.py
```diff
@@ -1393,6 +1397,8 @@ def plot_spectrum(self, ax=None, fp_kwargs=None, model_kwargs=None):
             Keyword arguments passed to `FluxPoints.plot`.
         model_kwargs : dict
             Keywords passed to `SpectralModel.plot` and `SpectralModel.plot_error`
+        plot_error : bool
+            False to not plot the errors of the model since for some of them it is not handle yet properly in Gammapy
```
cdeil: Please shorten description (see comment above)

JouvinLea:
@cdeil 
I fixed the small docs and space issues.
I added a description for the Adsorbed model, hope it is clearer!

Personally I would prefer to stay with this plot_error parameter. I put a TODO to say to remove it when the issue is fixed. I know exactly which line to comment to have my script running but this is a useful plot routine so I think for non developper it is better to have this option. But as you wish, I can comments those lines by myself each time I run my script.

gammapy/spectrum/flux_point.py
```diff
@@ -1426,7 +1432,8 @@ def plot_spectrum(self, ax=None, fp_kwargs=None, model_kwargs=None):
         plot_kwargs.setdefault("color", ax.lines[-1].get_color())
         del plot_kwargs["label"]
 
-        if self.model.parameters.covariance is not None:
+        # TODO: remove this plot error option when there is a better manipulation of uncertainties
+        if self.model.parameters.covariance is not None and plot_error:
```
adonath: @JouvinLea Maybe just go with a try except statement here and remove the `plot_error` option? Like so:

```
#TODO: remove try except....
try:
    self.model.plot_error()
except WhatEverError:
    log.debug("Model does not support evaluation of errors")
    pass

```

cdeil:
@JouvinLea - Please fix:
https://dev.azure.com/gammapy/gammapy/_build/results?buildId=2035&view=logs&jobId=d7d66435-5ae4-561a-5876-2cd4e19b3ce2&taskId=922dfe32-5b90-53bc-18db-0fa0ecbbd97b&lineStart=187&lineEnd=188&colStart=1&colEnd=1

cdeil:
@JouvinLea - Thanks!

PR #2455, opened 2019-10-14
Improve datasets serialization to support different dataset classes

QRemy:
-add dataset types registry
-simplify dict_to_datasets
-move update_dataset to MapDataset.update
then specific read/write/update methods should be added to the other dataset classes


gammapy/cube/fit.py
```diff
@@ -675,10 +677,31 @@ def read(cls, filename, name=""):
         hdulist = fits.open(str(filename))
         return cls.from_hdulist(hdulist, name=name)
 
+    @classmethod
+    def from_dict(cls, data, components, models):
```
adonath: Please add a docstring here...

PR #2458, opened 2019-10-15
PIG 17 - Provenance capture

Bultako:
This PIG proposes to develop a solution for automatic capturing provenance when using the high-level interface. For the moment **only provenance capture and storage** and it only covers the high-level interface. The prov information would be logged in a *structured way* following the IVOA Prov model (which is based on W3C prov model). The logging settings declared in the config file of the high-level interface would activate prov logging and where to log this info.  There is no need for external dependencies. 

There has been been some [prototyping work](https://github.com/Bultako/gammapy/tree/prov/gammapy/utils/provenance) already, and standard provenance [files](https://openprovenance.org/store/documents/1241) and [graphs](https://app.box.com/s/yp8wjo73yvhy7nyh7b8rac4rtrk4dqi0) may be easily built from the log file content with what is identified as *prov info*.

### DIFF
--- a/docs/development/pigs/pig-017.rst
+++ b/docs/development/pigs/pig-017.rst
@@ -3,7 +3,7 @@
 .. _pig-017:
 
 ****************************
-PIG 16 - Provenance tracking
+PIG 17 - Provenance tracking
 ****************************
 
 * Author: José Enrique Ruiz, Mathieu Servillat
@@ -21,42 +21,64 @@ features will be addressed in the coming releases. Having this high-level interf
 allow us to think on procuring a way to automatically record a structured provenance of
 the data analysis processes undertaken in working sessions or in Python scripts.
 
-A structured provenance capture opens the door to provide the means to perform comparisons
-and forensic studies of different analysis processes, extracting relevant information,
-which in turn should improve reproducibility and reuse by the community. The proposed
-contribution described in this PIG is a necessary first step to accomplish this ultimate
-goal, which will only be achieved in another PIG with tools for provenance inspection and
-filtering.
+A structured provenance capture using a standard model opens the door to provide the
+means to perform comparisons and forensic studies of different analysis processes, extracting
+relevant information, which in turn should improve reproducibility and reuse by the community.
+The proposed contribution described in this PIG is a necessary first step to accomplish this
+ultimate goal, which will only be achieved with tools for provenance inspection and filtering.
+The latter could be found as solutions developed outside Gammapy or, in the worst of the cases,
+as future custom Gammapy development proposed in a future PIG.
 
 Status
 ======
 
-Logging
+- Logging issues
+- W3C / IVOA Model
+- ctapipe Provenance API
+- prov Python API
 
 Proposal
 ========
 
 - Interoperability with other provenance logs (ctapipe)
     - Use Prov IVOA model
-- Reuse and extend ctapipe Provenance API
+
+- Reuse and extend ctapipe Provenance API (execution environment)
+    - Execution environment
+    - Data processing steps
+    - Parameter values used and datasets involved
+    - Detailed relationships among processes and datasets
+
 - Extensibility and modularity
     - Simple and non-intrusive in Gammapy code-base
+    - Re-use description YAML file
+
 - Serialization
     - Use Prov IVOA model
-    - Files in prov model format
-    - Parsed/converted into prov model format
-        - files in python objects format
-        - relational database
-        - structlogs
-
+    - Log files with specific PROV logs syntax
+    - Log files parsed/converted into prov model format
+    - Use a .prov folder for prov serialisation and intermediate files -> sharing
+    - Graph generation
 
 
 Outlook
 =======
 
-- OPUS
-- database M.Sanguillon SQL Alchemy  into SQLite and Posgres
-- Provenance extraction
+- Relational database serialisation (M. Sanguillon SQL Alchemy into SQLite and PosgreSQL)
+- Re-use json-schema file to generate YAML
+- Provenance extraction / filtering features
+- Independent Python API
+- Distributed infrastructure deployment (OPUS)
+
+
+Alternatives
+============
+
+- ctapipe tools
+- Eliot
+- structlog
+- autologging
+- noWorkflow
 
 Task list
 =========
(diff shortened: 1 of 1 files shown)


docs/development/pigs/pig-017.rst
```diff
@@ -0,0 +1,217 @@
...
+logs as a structured sequence of nested actions and subactions, as well as the parameters values involved.
+These APIs could be used to develop the custom-model provenance capture layer proposed in this PIG, but they
+add an external dependency that for our simple model we find oversized and could be easily avoided. It remains
+that if technical blockers are found during the development of this PIG, we could always explore the solutions
+offered by these APIs.
+
+Task list
+=========
+
+- General cleaning of Gammapy code-base to replace ``print()`` functions with standard Python ``logging``.
```
cdeil: You can remove this, we already use `logging` consistently instead of `print`. The places where we use `print` is for cases where we explicitly write information to the user screen, e.g. when they print out the information for a catalog source, i.e. someone explicitly calls `print(source_catalogs["hgps"]["crab"])`. Is it clear that those are cases where we in the library should call `print` instead of `log`, or does this need discussion?

Bultako:
> Is this something you want to implement within the next ~ month, i.e. for v1.0?
> Or is the plan what to implement next year?

I agree we are short on time for v1.0
Moreover, what is proposed in the PIG is only half of the whole. The other half is user-friendly prov **inspection** features. If I had to chose I'd prefer no to ship a v1.0 with only half of the whole. 

Nevertheless, there are some open issues related with provenance, and I'd like those discussions to be aware of this PIG.

cdeil:
Discussed in Gammapy CC call just now. Agreed to table this PIG and work for now to allow us to focus on other things for v1.0, and then to discuss and work on this in early 2020.

mservillat:
I discussed this with @cboisson, it would be good to keep an eye on this implementation regularly as it is already advanced prototyping work, and it is not too intrusive or conflicting with other gammapy features. Though there is **capture** and **inspection** of provenance, I think gammapy is mainly concerned by the **capture** following a standard (IVOA/W3C). Having the provenance information compatible with a standard, it will be possible for it to be explored/used/disseminated with external tools, or more importantly combined with external provenance information (what happened before or after a gammapy analysis).

docs/development/pigs/pig-017.rst
```diff
@@ -0,0 +1,217 @@
+.. include:: ../../references.txt
+
+.. _pig-017:
+
+***************************
+PIG 17 - Provenance capture
+***************************
+
+* Author: José Enrique Ruiz, Mathieu Servillat, lead developers welcome...
```
cdeil: Suggest to remove "lead developers welcome" now and then to add anyone that makes significant contributions to this PIG (usually via comments, not edits) as co-authors.

docs/development/pigs/pig-017.rst
```diff
@@ -0,0 +1,217 @@
...
+* Created: October 15, 2019
+* Accepted:
+* Status: Draft
+* Discussion:
```
cdeil: Add link to the PR.

docs/development/pigs/pig-017.rst
```diff
@@ -0,0 +1,217 @@
...
+- Develop a small set of tools which translate provenance logs into W3C syntax formatted files.
+- Develop a small set of tools which produce graphs of the workflow process using provenance logs.
+
+Decision
+========
+TBD
+
+
+.. _High-level interface: https://docs.gammapy.org/0.14/scripts/index.html
+.. _PIG12: https://docs.gammapy.org/0.14/development/pigs/pig-012.html
```
cdeil: PIG 12 is merged, so you can remove the URL here, and in the text use a Sphinx reference, i.e.
```
:ref:`pig-012`
```

docs/development/pigs/pig-017.rst
```diff
@@ -0,0 +1,217 @@
...
+as with other astronomical archive services or software providing provenance information in the Virtual
+Observatory framework and with the W3C provenance models. In this optic, the development will be done on
+top of `ctapipe.core.provenance`_, re-using its present capabilities when possible and extending it. The
+information captured will relate to the execution environment, the overall high-level interface configuration,
+the workflow execution process (i.e the relationships among datasets and processes and the parameter values
+used in each process step), and the serialisations of the intermediate and final datasets or objects produced.
+
+A simple and non-intrusive implementation strategy should be considered, with minor modifications in the
+code-base of Gammapy. Metaclass inheritance, class decoration and subclassing mechanisms applied to the
+``Analysys`` class of the high-level interface will be explored. The code will be placed in one or a few
```
cdeil: Analysys -> Analysis

docs/development/pigs/pig-017.rst
```diff
@@ -0,0 +1,217 @@
...
+PROV-XML and PROV-JSON), as well as to provenance graphs (i.e. PDF, PNG, SVG). The use of a custom
+structured log format allows us to record additional information other than the strictly permitted by
+W3C standards.
+
+All the configuration settings and parameter values may be accessed via the ``Analysis`` class at the very
+moment of the execution of each task. Captured tasks, datasets and hashable objects in general need a unique
+identifier that could be generated with MD5 hashing. The intermediate and final datasets and objects produced
+will be saved into the local disk if this option is enabled as a configuration parameter, and links to the
+serialisations of these objects will be recorded accordingly in the provenance log file. Finally, it is
+advisable to proceed with a general cleaning of the Gammapy code-base so to replace ``print()`` functions
```
cdeil: Remove this statement. We don't have any `print` in Gammapy.
It is the case that we have too few and not well-thought-out `log` statements, so you could make a comment saying that in Gammapy we should improve what it logged, so that users can from log files in most cases understand what analysis was executed and what happened.

QRemy:
@bkhelifi, @mservillat, @cboisson 
What's the status of provenance capture plan ?
How do we  continue this PIG ?

mservillat:
This is a work I prepared with Jose Enrique, this also started the logprov project (https://github.com/mservillat/logprov).
However, this is besed on Python inspection and can slow down the data processing... we have to discuss if this is stil the way to go, or if we should do something more similar to ctapipe (https://github.com/cta-observatory/ctapipe/blob/main/src/ctapipe/core/provenance.py).

registerrier:
There is little point in keeping the PIG open as is. I would propose to close this PIG, mark it rejected and merge the PR so that it appears it the list of discussed PIGs.

Does it sound fine @mservillat @bkhelifi ?

mservillat:
I think it should indeed be closed, and maybe it will be reopened as another PIG after some discussions, maybe in connection with the ACME project.

Issue #2349, opened 2019-09-09
Modernise codebase to Python 3.6


cdeil:
This is a reminder issue to update the Gammapy codebase to use Python 3.6 features, especially f-strings and dict instead of OrderedDict.

See #2218 and [here](https://github.com/gammapy/gammapy/blob/master/docs/development/pigs/pig-013.rst#task-list).

cdeil:
I'm moving this to v0.15. The only concrete remaining action item I'm aware of is to use f-strings instead of `str.format` in `gammapy.catalog`. Those lines can be found by searching for `.format`. In #2363 many were auto-formatted, but apparently that tool is limited to simple expressions only.

cdeil:
There's still `gammapy/catalog/fermi.py` that has a lot of `str.format` where f-strings could be used, but really `fermi.py` needs to be improved first to have less code duplication.

Apart from that there's other Python 3.6 (or 3.7 if we bump) features we could use, e.g. `*` to have keyword-only arguments, or more use of `pydantic` or `dataclasses` to get automatic type validation beyond `gammapy/analysis/config.py`, and there's some other things, e.g. I think we should use generators more `yield` instead of accumulating lists in methods and returning those.

But most of the work here is done, and judging from experience generic long-term reminder Github issues aren't very useful, so I'm closing this one now.

PR #2218, opened 2019-06-06
PIG 13 - Gammapy dependencies and distribution

cdeil:
This PIG is to make a clear plan and gather feedback concerning Gammapy dependencies and distribution.

Please give me a day to work on this, don't comment yet.

PR #2350, opened 2019-10-16
Modernise Gammapy codebase

cdeil:
This is a reminder issue to update the Gammapy codebase to use recent features from Numpy, Astropy, ..., i.e. workarounds for old versions of those dependencies can be removed. Partially this can be done by reviewing `TODO` comments or grepping for `version`, partially this has to be done manually by browsing our codebase.

See #2218 and [PIG 13](https://github.com/gammapy/gammapy/blob/master/docs/development/pigs/pig-013.rst) and the related #2349.

### DIFF
--- a/docs/development/howto.rst
+++ b/docs/development/howto.rst
@@ -464,25 +464,6 @@ File and directory path handling
 
 In Gammapy use `pathlib.Path` objects to handle file and directory paths.
 
-.. code-block:: python
-
-    from pathlib import Path
-
-    dir = Path('folder/subfolder')
-    filename = dir / 'filename.fits'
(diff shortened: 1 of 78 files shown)


cdeil:
@adonath - Is b8d1bb07b79edf7b73483e70e2de570625b53fe6 OK? (tests pass)

Are you aware of any other code that we can modernise by using modern Numpy & Astropy features?

cdeil:
I'll update from `tmpdir` to `tmp_path`, to use `pathlib.Path` objects in tests like we do in the normal code. Not sure when this was introduced in pytest, I think already a few years ago. See http://doc.pytest.org/en/latest/tmpdir.html

PR #2468, opened 2019-10-17
Improve map and spectrum events fill methods

cdeil:
This PR makes the methods to fill events into `Map` or `CountsSpectrum` a bit better.

At least IMO, it's a bit more consistent this way.

Before, we had `fill_map_counts` as a function, and `CountsSpectrum.fill` as a method. Now, it's `container.fill_events` in both cases with the same API.

Before we had `events. _get_coord_from_geom` as private, but were accessing it from the outside in `fill_map_counts`. Now, it's `events.map_coord` and part of the public API, simiilar to other properties `events` already has.


gammapy/data/event_list.py
```diff
@@ -481,19 +481,20 @@ def check(self, checks="all"):
...
         Parameters
         ----------
         geom : `~gammapy.maps.Geom`
-            the geom used to define the MapCoord
+            Geometry
 
         Returns
         -------
-        mapcoord : `~gammapy.Map.MapCoord`
+        coord : `~gammapy.maps.MapCoord`
```
adonath: The method currently returns a `dict` and not a `MapCoord` object. I guess there are two options:
- Rename method to just `coords()` an keep returning a dict
- Change to create a `MapCoord` object inside

gammapy/data/event_list.py
```diff
@@ -481,19 +481,20 @@ def check(self, checks="all"):
         checker = EventListChecker(self)
         return checker.run(checks=checks)
 
-    def _get_coord_from_geom(self, geom):
-        """Return MapCoord of the EventList relative to an input geometry.
+    def map_coord(self, geom):
```
adonath: I think the `EventList` API would be more consistent, if we don't pass `geom` and make `map_coord` a property just as `.time`, `.offset`, `.skycoord`. Just bundle all relevant coordinates and avoid copies if we are concerned about efficiency...

gammapy/maps/base.py
```diff
@@ -771,6 +771,10 @@ def interp_by_pix(self, pix, interp=None, fill_value=None):
         """
         pass
 
+    def fill_events(self, events):
```
adonath: I'm still 48:52 to not add `.fill_events()` and just use `Map.fill_by_coord(events.map_coord)` otherwise one could argue why we don't implement `.fill_bkg()`, `fill_psf()`, etc. as well...

gammapy/data/event_list.py
cdeil: I'll change to return a `MapCoord` object

gammapy/maps/base.py
cdeil: We do have this for PSF:
https://github.com/gammapy/gammapy/blob/de8e1180184d3403484902cd612d655c0b593f5b/gammapy/maps/wcsnd.py#L615

For BKG - there's no simple object such that `map.fill_bkg(bkg)` would make sense.

gammapy/data/event_list.py
cdeil: Did you see this?

https://github.com/gammapy/gammapy/pull/2468/files#diff-0e905bc71b734b5e047ddf9dce8d2ef6R59-R64

@registerrier already implemented logic to match events table to arbitrary map axis, e.g. for `EVENT_ID`, if the names match in a case insensitive way.

Removing the `geom` argument here would remove that feature, or reduce it to a hard-coded preset list of coords.

So I'll leave as-is for now, I still think this PR is a step in the right direction.
But yes, possibly this is something that could be changed in the future, probably we should only decide once we have real-world use cases (i.e. a data reduction and fittting working for maps with a TIME or  EVENT_ID axis).

IF the conclusion is that `geom` shouldn't be passed to `events.map_coord`, then I agree that `map.fill_events` isn't useful any more - it'll be easy to replace, because you can just right-click in PyCharm and say "list callers" or use grep.

cdeil:
@adonath - I changed to return a MapCoord in 64cc992 .

It currently doesn't work. I think there's a bug in `MapCoord.create` - it doesn't take the frame information when `skycoord` is passed? @adonath - Can you please have a look and fix this issue either here or in a separate PR if you want a clean changelog for fixes?
https://gist.github.com/cdeil/7d8f1e26444d2479e1989f837a09d0ae

PR #2475, opened 2019-10-18
Implement `ReflectedRegionsBackgroundMaker`

adonath:
This PR implements a `ReflectedRegionsBackgroundMaker` and a small first test in `test_reflected.py`. Other existing tests will be moved over step by step in following PRs.

### DIFF
--- a/gammapy/spectrum/reflected.py
+++ b/gammapy/spectrum/reflected.py
@@ -49,7 +49,7 @@ class ReflectedRegionsFinder:
     exclusion_mask : `~gammapy.maps.WcsNDMap`, optional
         Exclusion mask
     binsz : `~astropy.coordinates.Angle`
-        Bin size of the reference map used for region finding. Default : 0.01 deg
+        Bin size of the reference map used for region finding.
 
     Examples
     --------
@@ -137,9 +137,9 @@ class ReflectedRegionsFinder:
         center : `~astropy.coordinates.SkyCoord`
             Rotation point
         binsz : `~astropy.coordinates.Angle`
-            Reference map bin size. Default : 0.01 deg
+            Reference map bin size.
         min_width : `~astropy.coordinates.Angle`
-            Minimal map width. Default : 0.3 deg
+            Minimal map width.
 
         Returns
         -------
@@ -270,7 +270,7 @@ class ReflectedRegionsBackgroundEstimator:
     observations : `~gammapy.data.Observations`
         Observations to process
     binsz : `~astropy.coordinates.Angle`
-        Optional, bin size of the maps used to compute the regions, Default '0.01 deg'
+        Optional, bin size of the maps used to compute the regions.
     kwargs : dict
         Forwarded to `~gammapy.spectrum.ReflectedRegionsFinder`
     """
@@ -350,9 +350,9 @@ class ReflectedRegionsBackgroundEstimator:
         cmap : `~matplotlib.colors.ListedColormap`, optional
             Color map to use
         idx : int, optional
-            Observations to include in the plot, default: all
+            Observations to include in the plot.
         add_legend : boolean, optional
-            Enable/disable legend in the plot, default: False
+            Enable/disable legend in the plot.
         """
         import matplotlib.pyplot as plt
 
@@ -439,7 +439,7 @@ class ReflectedRegionsBackgroundMaker:
     exclusion_mask : `~gammapy.maps.WcsNDMap`, optional
         Exclusion mask
     binsz : `~astropy.coordinates.Angle`
-        Bin size of the reference map used for region finding. Default : 0.01 deg
+        Bin size of the reference map used for region finding.
     """
 
     def __init__(
@@ -461,7 +461,7 @@ class ReflectedRegionsBackgroundMaker:
         self.max_region_number = max_region_number
 
     def _get_finder(self, observation):
-        finder = ReflectedRegionsFinder(
+        return ReflectedRegionsFinder(
             binsz=self.binsz,
             exclusion_mask=self.exclusion_mask,
             center=observation.pointing_radec,
@@ -471,7 +471,6 @@ class ReflectedRegionsBackgroundMaker:
             max_region_number=self.max_region_number,
             angle_increment=self.angle_increment,
         )
-        return finder
 
     def make_counts_off(self, dataset, observation):
         """Make off counts.
@@ -524,7 +523,6 @@ class ReflectedRegionsBackgroundMaker:
         -------
         dataset_on_off : `SpectrumDatasetOnOff`
             On off dataset.
-
         """
         counts_off = self.make_counts_off(dataset, observation)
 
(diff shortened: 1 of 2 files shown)


gammapy/spectrum/reflected.py
```diff
@@ -414,3 +416,122 @@ def plot(self, fig=None, ax=None, cmap=None, idx=None, add_legend=False):
...
+        observation : `DatastoreObservation`
+            Data store observation.
+
+        Returns
+        -------
+        dataset_on_off : `SpectrumDatasetOnOff`
+            On off dataset.
+
+        """
+        kwargs = {}
```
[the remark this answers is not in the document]
adonath: Done

gammapy/spectrum/reflected.py
```diff
@@ -414,3 +419,124 @@ def plot(self, fig=None, ax=None, cmap=None, idx=None, add_legend=False):
...
+        self.region = region
+        self.binsz = binsz
+        self.exclusion_mask = exclusion_mask
+        self.angle_increment = Angle(angle_increment)
+        self.min_distance = Angle(min_distance)
+        self.min_distance_input = Angle(min_distance_input)
+        self.max_region_number = max_region_number
+
+    def _get_finder(self, observation):
+        finder = ReflectedRegionsFinder(
```
cdeil: Another coding style suggestion. If a temp variable is just created and returned, but never used, I suggest to put `return` directly and avoid the temp variable. I've changed to that coding pattern already mostly throughout Gammapy

gammapy/spectrum/reflected.py
```diff
@@ -414,3 +419,124 @@ def plot(self, fig=None, ax=None, cmap=None, idx=None, add_legend=False):
...
+    def make_counts_off(self, dataset, observation):
+        """Make off counts.
+
+        Parameters
+        ----------
+        dataset : `SpectrumDataset`
+            Spectrum dataset.
+        observation : `DatastoreObservation`
+            Data store observation.
+
```
cdeil: Remove one empty line

gammapy/spectrum/tests/test_reflected.py
```diff
@@ -161,16 +164,17 @@ def bad_on_region(exclusion_mask, on_region):
...
-        assert len(bkg_estimator.result[1].off_region) == 11
-        assert "Reflected" in str(bkg_estimator.result[1])
-
-    @requires_dependency("matplotlib")
-    def test_plot(self, bkg_estimator):
-        with mpl_plot_check():
-            bkg_estimator.plot(idx=1, add_legend=True)
-            bkg_estimator.plot(idx=[0, 1])
+def test_reflected_bkg_maker(spectrum_dataset_maker, reflected_bkg_maker, observations):
+
```
cdeil: Remove empty line at start of functions (consistency within Gammapy)

gammapy/spectrum/reflected.py
```diff
@@ -414,3 +419,124 @@ def plot(self, fig=None, ax=None, cmap=None, idx=None, add_legend=False):
...
+    min_distance : `~astropy.coordinates.Angle`, optional
+        Minimal distance between two consecutive reflected regions
+    min_distance_input : `~astropy.coordinates.Angle`, optional
+        Minimal distance from input region
+    max_region_number : int, optional
+        Maximum number of regions to use
+    exclusion_mask : `~gammapy.maps.WcsNDMap`, optional
+        Exclusion mask
+    binsz : `~astropy.coordinates.Angle`
+        Bin size of the reference map used for region finding. Default : 0.01 deg
```
cdeil: Suggest to remove the "Default: 0.01 deg" here and almost everywhere in Gammapy.
It's very hard to maintain, we have maybe ~ 10,000 parameters in function calls with defaults, and they often change and people forget to edit the docstring.

gammapy/spectrum/reflected.py
adonath: Done

gammapy/spectrum/reflected.py
adonath: Done

gammapy/spectrum/tests/test_reflected.py
adonath: Done

gammapy/spectrum/reflected.py
adonath: Agreed. This was a copy and paste error from `ReflectedRegionsFinder`, I modified the other docstrings in `reflected.py` as well.

gammapy/spectrum/reflected.py
```diff
@@ -414,3 +419,122 @@ def plot(self, fig=None, ax=None, cmap=None, idx=None, add_legend=False):
...
+
+        region_union = regions[0]
+
+        for region in regions[1:]:
+            region_union = region_union.union(region)
+
+        wcs = finder.reference_map.geom.wcs
+        events_off = observation.events.select_region(region_union, wcs)
+
+        counts_off = dataset.counts.copy()
```
registerrier: This should copy the contents as well. It should be created from scratch to have an empty vector

gammapy/spectrum/reflected.py
```diff
@@ -414,3 +419,122 @@ def plot(self, fig=None, ax=None, cmap=None, idx=None, add_legend=False):
...
+
+    def __init__(
+        self,
+        region,
+        angle_increment="0.1 rad",
+        min_distance="0 rad",
+        min_distance_input="0.1 rad",
+        max_region_number=10000,
+        exclusion_mask=None,
+        binsz="0.01 deg",
```
registerrier: If there is a default value, it is better to state it in the docstring no?

gammapy/spectrum/reflected.py
adonath: Ouch...that's indeed a serious bug. I guess I'd have noticed, when moving the existing tests over, but it's good to fix it now. Thanks!

gammapy/spectrum/reflected.py
adonath: I think at some point we decided (not sure if it's documented somewhere), that it's easier to maintain, if the default values are not repeated in the docstring again. Users always see the defaults together with the docstring when they use e.g. `help(ReflectedRegionsBackgroundMaker)` or `ReflectedRegionsBackgroundMaker?` in interactive environments. The same goes for the online documentation see e.g. [here](https://docs.gammapy.org/0.14/api/gammapy.spectrum.ReflectedRegionsBackgroundEstimator.html#gammapy.spectrum.ReflectedRegionsBackgroundEstimator). So the information would be basically repeated...

gammapy/spectrum/reflected.py
adonath: No, actually I realised `CountsSpectrum.fill_events()` overwrites the existing data (see [here](https://github.com/gammapy/gammapy/blob/master/gammapy/spectrum/core.py#L131)). This makes the behaviour different from `Map.fill_by_coord()` which adds the data. I guess we should make it consistent...

adonath: I changed to not copy and will do the consistency change in a follow up PR (it might affect other parts in `gammapy.spectrum`)...

PR #2477, opened 2019-10-18
Add SafeMaskMaker

adonath:
This PR implements a `SafeMaskMaker`. Currently it only support spectral datasets, it can be later adapted / extended to support `MapDataset` as well. This PR also adds a simple test. Again existing tests will be moved in subsequent PRs.

### DIFF
--- a/gammapy/spectrum/make.py
+++ b/gammapy/spectrum/make.py
@@ -217,9 +217,9 @@ class SafeMaskMaker:
     """
 
     def __init__(
-        self, methods=("aeff-default",), aeff_percent=10, bias_percent=10
+        self, methods="aeff-default", aeff_percent=10, bias_percent=10
     ):
-        self.methods = methods
+        self.methods = list(methods)
         self.aeff_percent = aeff_percent
         self.bias_percent = bias_percent
 
--- a/gammapy/spectrum/make.py
+++ b/gammapy/spectrum/make.py
@@ -217,7 +217,7 @@ class SafeMaskMaker:
     """
 
     def __init__(
-        self, methods=["aeff-default"], aeff_percent=10, bias_percent=10
+        self, methods=("aeff-default",), aeff_percent=10, bias_percent=10
     ):
         self.methods = methods
         self.aeff_percent = aeff_percent
@@ -246,8 +246,7 @@ class SafeMaskMaker:
             log.warning(f"No thresholds defined for obs {observation}")
             e_min, e_max = None, None
 
-        mask_safe = dataset.counts.energy_mask(emin=e_min, emax=e_max)
-        return mask_safe
+        return dataset.counts.energy_mask(emin=e_min, emax=e_max)
 
     def make_mask_energy_aeff_max(self, dataset):
         """Make safe energy mask from aeff max.
@@ -264,8 +263,7 @@ class SafeMaskMaker:
         """
         aeff_thres = self.aeff_percent / 100 * dataset.aeff.max_area
         e_min = dataset.aeff.find_energy(aeff_thres)
-        mask_safe = dataset.counts.energy_mask(emin=e_min)
-        return mask_safe
+        return dataset.counts.energy_mask(emin=e_min)
 
     def make_mask_energy_edisp_bias(self, dataset):
         """Make safe energy mask from aeff max.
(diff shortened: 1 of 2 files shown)


cdeil:
Introducing the SafeMaskMaker(TM)

Screenshot 2019-10-18 at 13 56 46

adonath:
@cdeil I guess a `SafeMaskMaker` is what every gamma-ray scientist needs for healthy skin as well as healthy analysis results (hope we don't get a lawsuit for using the name, because we don't have funding to pay lawyers, but wait a second...what about renaming it to `SafeMaskEstimator`? :-)

gammapy/spectrum/make.py
```diff
@@ -195,3 +195,120 @@ def run(self, observation, selection=None):
...
+    aeff_percent : float
+        Percentage of the maximal effective area to be used
+        as lower energy threshold for method "aeff-max".
+    bias_percent : float
+        Percentage of the energy bias to be used as lower
+        energy threshold for method "edisp-bias"
+    """
+
+    def __init__(
+        self, methods=["aeff-default"], aeff_percent=10, bias_percent=10
```
cdeil: Don't put mutable contains like lists or dicts as default arguments. If needed, put `None` and put the default in the method. https://docs.python-guide.org/writing/gotchas/

gammapy/spectrum/make.py
```diff
@@ -195,3 +195,120 @@ def run(self, observation, selection=None):
...
+            Safe data range mask.
+        """
+        try:
+            e_max = observation.aeff.high_threshold
+            e_min = observation.aeff.low_threshold
+        except KeyError:
+            log.warning(f"No thresholds defined for obs {observation}")
+            e_min, e_max = None, None
+
+        mask_safe = dataset.counts.energy_mask(emin=e_min, emax=e_max)
```
cdeil: Just return `mask_safe`, no point in creating a temp variable.

gammapy/spectrum/make.py
```diff
@@ -195,3 +195,120 @@ def run(self, observation, selection=None):
...
+            Dataset to compute mask for.
+
+        Returns
+        -------
+        mask_safe : `~numpy.ndarray`
+            Safe data range mask.
+        """
+        aeff_thres = self.aeff_percent / 100 * dataset.aeff.max_area
+        e_min = dataset.aeff.find_energy(aeff_thres)
+        mask_safe = dataset.counts.energy_mask(emin=e_min)
```
cdeil: Just return `mask_safe`, no point to create a temp variable.

gammapy/spectrum/tests/test_make.py
```diff
@@ -49,6 +49,11 @@ def spectrum_dataset_maker_crab():
     return SpectrumDatasetMaker(region=region, e_reco=e_reco, e_true=e_true)
 
 
+@pytest.fixture()
+def safe_mask_maker():
+    return SafeMaskMaker()
```
cdeil: Remove fixture, just put `SafeMaskMaker()` below.
Fixtures are only useful for resources that take time to produce (load from disk, compute) or lots of code.
Otherwise they are bad, because one has to scroll up and down from the tests to see what it is.

gammapy/spectrum/tests/test_make.py
```diff
@@ -87,3 +92,20 @@ def test_spectrum_dataset_maker_hess_cta(
 
     assert_allclose(datasets[0].background.data.sum(), 2.238345, rtol=1e-5)
     assert_allclose(datasets[1].background.data.sum(), 2.164593, rtol=1e-5)
+
+
+def test_safe_mask_maker(spectrum_dataset_maker_crab, safe_mask_maker, observations_hess_dl3):
+
```
cdeil: Never put empty lines at the start of functions (consistency coding style).

cdeil:
One more thing: I'd suggest to never put ` in GH titles. That's markdown, but it's not rendered in titles.
And we copy & paste the titles into RST where you'd have to write double ``.
Easiest to just not put the `. It's mostly what we have in the past 2476 issues and PRs.

gammapy/spectrum/make.py
adonath: Done

gammapy/spectrum/make.py
adonath: Done

gammapy/spectrum/tests/test_make.py
adonath: Done

gammapy/spectrum/tests/test_make.py
adonath: Done

PR #2478, opened 2019-10-18
Improve 2FHL catalog support

QRemy:
- add more infos
- add spatial_model and sky_model
- have to check the extended template for RX J1713, for now I guess it's the same than for 3FGL and added a copy to read with 2FHL in gammapy-extra

### DIFF
--- a/gammapy/catalog/core.py
+++ b/gammapy/catalog/core.py
@@ -201,11 +201,16 @@ class SourceCatalog:
         data = table_row_to_dict(self.table[index])
         data[self._source_index_key] = index
 
-        try:
+        if "Extended_Source_Name" in data:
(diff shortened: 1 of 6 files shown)


cdeil:
I see https://github.com/gammapy/gammapy-extra/commit/fef1d9b50509319b79ac521783910a74eb98233f

@adonath - could you please update the dataset index to fix this?
https://dev.azure.com/gammapy/gammapy/_build/results?buildId=2137

Otherwise 👍 to merge this in.

(the string formatting stuff can be shortened, but that's the case for all catalogs, I can do that in a follow-up PR)

adonath:
I updated the dataset index file in https://github.com/gammapy/gammapy/commit/2b2b475c5f7efc4f5fe90ffeb7725937331c6085

Issue #2484, opened 2019-10-24
Noisy edisp at 0° offset in simulate_dataset


facero:
`dataset = simulate_dataset(skymodel, geom, pointing, irfs, livetime=50 * u.h)`
While playing with `simulate_dataset` I realized that the edisp generated is very noisy.
This is due to the fact that the default value for `offset` is 0° and the edisp in the IRF is very noisy for that offset:
![edisp_offset0](https://user-images.githubusercontent.com/8652898/67485574-9e0b0780-f66a-11e9-907f-05e3f5111394.png)
When trying offset=1° then it looks ok. So this is not a bug of gammapy in itself but 

In fact we should remove the offset keyword in simulate_dataset, this could be estimated by geom and pointing objects ?

I chatted with @adonath who suggested another way. Could you comment on that ?

On a longer term, we could imagine having pointing being a list of SkyCoord and livetime being an array to simulate a more realistic observation with a pointing pattern. Then as when reading run, we could have a bakground map, weighted PSF and edisp.

cdeil:
@facero - Which IRFs?

Please add a link to the file, or even a code snippet or link to a notebook in a gist to reproduce the issue.

My guess is that we shouldn't spend any time one this, and just encourage CTA and HESS to produce good IRFs in the future. Trying to fix bad IRFs in the science tools has the potential to be a huge time sink and probably fundamentally not the right approach.

facero:
Good point, this info helps indeed ;-)
```
irfs = load_cta_irfs("$GAMMAPY_DATA/cta-1dc/caldb/data/cta/1dc/bcf/South_z20_50h/irf_file.fits")
irfs['edisp'].to_energy_dispersion(offset=0*u.deg).peek()
```

cdeil:
The CTA prod3 IRF issues at small offsets, especially for EDISP, are well-known. See e.g.:

- https://indico.cta-observatory.org/event/1575/contributions/16494/attachments/13676/16554/2018-05-15_CTA_Gammapy.pdf
- https://indico.cta-observatory.org/event/1686/contributions/16076/attachments/13413/16140/2018-04-energy-dispersion-knoedlseder.pdf

As far as I know, Jürgen tried to implement workarounds in ctools, which in turn caused problems with the HESS IRFs, and fundamentally now everyone is of the opinion that IRF producers have to smooth and science tools shouldn't do that.

[...]

PR #2486, opened 2019-10-24
update default value in simulate_dataset (0°->1°)

facero:
to avoid irfs issue at offset 0° mentioned in #2484


adonath:
Thanks @facero!

PR #2487, opened 2019-10-24
Add PSFMap.sample_coord method

fabiopintore:
This PR aims at adding the method `sample_coord` into `~gammapy.cube.psf_map`. The method is developed to take as input a psf map and a `~gammapy.maps.MapCoord` objects, where the latter represents the map of a set of simulated events (with lon, lat and energies). The method gives as output a `~gammapy.maps.MapCoord` object where the new coordinates of the events have been corrected for the given PSF.
Tests have been added.

### DIFF
--- a/docs/development/pigs/pig-011.rst
+++ b/docs/development/pigs/pig-011.rst
@@ -106,7 +106,7 @@ While we could provide helper functions to prepare the time bins. ``astropy.time
 is relatively easy to use, so that a user would execute code similar to the following
 example:
 
-.. code ::
+::
 
     from astropy.time import Time
     time_start = Time("2006-07-29 20:00:00.000")
@@ -168,7 +168,7 @@ After creating the ``Datasets``, the user would first define the model to be use
 and freeze its parameters. Then we would apply it to the various ``Dataset`` objects
 before calling the ``LightCurveEstimator``:
 
-.. code ::
+::
 
     sky_model = SkyModel(spatial_model=spatial_model, spectral_model=spectral_model)
     sky_model.parameters["index"].frozen = True
@@ -199,7 +199,7 @@ chi-square test, fractional variance estimation.
 
 Example usage:
 
-.. code ::
+::
 
     lightcurve.plot('amplitude')
 
@@ -210,7 +210,7 @@ The ``LightCurve`` object should be able to rpovide some rebinning functionaliti
 In particular, it stores the fit statistic scan, it should be possible to perform
 minimal significance rebinning:
 
-.. code ::
+::
 
     simple_rebin_lc = lightcurve.rebin(factor=2)
     # or
@@ -226,7 +226,7 @@ Time bins
 To work with time bins, we could also rely on ``astropy.timeseries`` if we force the dependency to astropy v>3.2.
 This would look like:
 
-.. code ::
+::
 
     from astropy.timeseries import BinnedTimeSeries
     times = BinnedTimeSeries(time_bin_start="2006-07-29 20:00:00.000",


gammapy/cube/psf_map.py
```diff
@@ -331,3 +333,46 @@ def stack(self, other):
     def copy(self):
         """Copy PSFMap"""
         return deepcopy(self)
+
+    def sample_coord(psf_map, map_coord, random_state=0):
+        """ Apply PSF corrections on the coordinates of a set of simulated events.
+        psf_map : `~gammapy.cube.psf_map.PSFMap`.
```
adonath: Please move the descriptions of the parameters in a separate `Parameters` section like so:
```
Parameters
----------
map_coord : `~gammapy.maps.MapCoord`
    Sequence of coordinates and energies of sampled events
....
```


Please also remove the documentation of `psf_map` once it's renamed to `self`.

gammapy/cube/psf_map.py
```diff
@@ -331,3 +333,46 @@ def stack(self, other):
...
+
+        Returns
+        -------
+        map_coord : `~gammapy.maps.MapCoord` object.
+            Sequence of PSF-corrected coordinates of the input map_coord map.
+        """
+
+        random_state = get_random_state(random_state)
+
+        map_coord["energy"].sort()
```
adonath: Is there a specific reason why you have to sort by energy here?

gammapy/cube/psf_map.py
```diff
@@ -331,3 +333,46 @@ def stack(self, other):
...
+        pix_coord = sample_pdf.sample_axis()
+        offset = psf_map.psf_map.geom.axes[0].pix_to_coord(pix_coord)
+
+        phi_angle = random_state.uniform(360, size=len(map_coord.lon)) * u.deg
+        event_positions = SkyCoord(
+            map_coord.lon, map_coord.lat, frame="galactic", unit="deg"
+        )
+        event_positions = event_positions.directional_offset_by(phi_angle, offset)
+
+        map_coord.lon[:] = event_positions.l
```
adonath: I think we should not overwrite the input coordinates, but just create a new `MapCoord` object, including the reconstructed positions.

gammapy/cube/tests/test_psf_map.py
```diff
@@ -182,3 +193,52 @@ def test_psfmap_stacking():
...
+    npred = evaluator.compute_npred()
+    return evaluator, npred
+
+
+def test_sample_coord():
+    psf_map = make_test_psfmap(0.1 * u.deg, shape="gauss")
+    eval, npred = get_npred_map()
+
+    nmap = WcsNDMap(geom=eval.geom, data=npred.data)
+    coords = nmap.sample_coord(n_events=2, random_state=0)
```
adonath: For the unit test there is not need to sample coordinates from an npred map. You could just create a `MapCoord` object "by hand" like so: 
```
coords = MapCoord({"lon": [0, 0] * u.deg, "lat": [0, 0.5] * u.deg, "energy": [1, 3] * u.TeV})
```

Then you can get rid of the `get_npred_map()` function as well.

gammapy/cube/tests/test_psf_map.py
```diff
@@ -182,3 +193,52 @@ def test_psfmap_stacking():
...
+
+    coords_corrected = PSFMap.sample_coord(psf_map=psf_map, map_coord=coords)
```
adonath: Note that the way `.sample_coord()` is used is actually by directly calling `psf_map.sample_coord()`, instead of calling `PSFMap.sample_coord()` and passing the `psf_map`.

gammapy/maps/tests/test_wcsnd.py
```diff
@@ -652,3 +652,7 @@ def test_map_sampling():
     assert_allclose(events["RA_TRUE"].data, [266.307081, 266.442255], rtol=1e-5)
     assert_allclose(events["DEC_TRUE"].data, [-28.753408, -28.742696], rtol=1e-5)
     assert_allclose(events["ENERGY_TRUE"].data, [2.755397, 1.72316], rtol=1e-5)
+
+    assert coords["lon"].unit == "deg"
+    assert coords["lat"].unit == "deg"
```
adonath: Once you do a rebase against master these changes should disappear...

gammapy/cube/psf_map.py
```diff
@@ -331,3 +333,46 @@ def stack(self, other):
     def copy(self):
         """Copy PSFMap"""
         return deepcopy(self)
+
+    def sample_coord(psf_map, map_coord, random_state=0):
```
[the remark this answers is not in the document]
fabiopintore: Done.

cdeil:
I see a hard-coded `frame="galactic"`. Probably the code doesn't handle both "icrs" and "galactic" correctly?

I think one has to use SkyCoord like in [EventListBase.map_coord](https://docs.gammapy.org/dev/_modules/gammapy/data/event_list.html#EventListBase.map_coord) to get the frame handling right?

gammapy/cube/psf_map.py
```diff
@@ -331,3 +333,51 @@ def stack(self, other):
...
+        }
+
+        pdf = self.psf_map.interp_by_coord(coord)
+
+        sample_pdf = InverseCDFSampler(pdf, axis=1)
+        pix_coord = sample_pdf.sample_axis()
+        offset = self.psf_map.geom.axes[0].pix_to_coord(pix_coord)
+
+        phi_angle = random_state.uniform(360, size=len(map_coord.lon)) * u.deg
+        event_positions = SkyCoord(
```
adonath: You could just use `map_coord.skycoord` to get a `SkyCoord` object, this will get the frame handling right.

gammapy/cube/psf_map.py
```diff
@@ -331,3 +333,51 @@ def stack(self, other):
...
+            map_coord.lon, map_coord.lat, frame="galactic", unit="deg"
+        )
+        event_positions = event_positions.directional_offset_by(phi_angle, offset)
+
+        corr_coord = MapCoord(
```
adonath: And similarly here: you can just use `MapCoord.create({"skycoord": event_positions, "energy": map_coord["energy"]})`

fabiopintore: Done

gammapy/cube/psf_map.py
```diff
@@ -331,3 +332,45 @@ def stack(self, other):
...
+        random_state = get_random_state(random_state)
+
+        coord = {
+            "lon": map_coord.lon.reshape(-1, 1),
+            "lat": map_coord.lat.reshape(-1, 1),
+            "energy": map_coord["energy"].reshape(-1, 1),
+            "theta": (self.psf_map.geom.axes[0].center),
+        }
+
+        pdf = self.psf_map.interp_by_coord(coord)
```
cdeil: You're still creating a `coord` dict dropping frame information, and then passing it to `self.psf_map.interp_by_coord(coord)`, which can't work if `self.psf_map` and `map_coord` have a different frame.

Please:

1. Change your test case to be like that, i.e. fail because of the frame handling bug
2. Fix the code. Probably this works? `coord = {"skycoord": mapcoord.skycoord,...}`, or something similar.

gammapy/cube/tests/test_psf_map.py
```diff
@@ -182,3 +187,26 @@ def test_psfmap_stacking():
...
+    )
+
+    coords_corrected = psf_map.sample_coord(map_coord=coords)
+
+    events = Table()
+    events["LON_TRUE"] = coords_corrected.lon
+    events["LAT_TRUE"] = coords_corrected.lat
+    events["ENERGY_TRUE"] = coords["energy"]
+
+    try:
```
cdeil: You want to establish which frame it is.
So remove try-except, and only keep the assert statement on the frame that you actually get.

cdeil:
@adonath @fabiopintore - Did you consider using `EventList` instead of `MapCoord` objects here?

I think that's the only use case for this method, and using it has a few small advantages:
- already has https://docs.gammapy.org/dev/api/gammapy.data.EventList.html#gammapy.data.EventList.map_coord
- Creating sampled `EventList` objects is the only use case, so if you make the API here `MapCoord`, you'll just make callers transform EventList -> MapCoord -> call sample -> MapCoord -> EventList?

Probably I'm missing something and there's a reason for your choice?

adonath:
@cdeil I think there is no clear advantage between using `MapCoord`, `EventList` or even a `Table`. The choice of the `MapCoord` comes from the Map API. For now we decided to call the method `.sample_coord()` (just because it's exactly what it does). But I agree it's maybe an option to rename it e.g. to `.sample_events()` and work with `EventList` directly. The one could adapt `PSFMap.sample()` accordingly. I think we possibly revisit our choice, when we work on the higher-level API (`MapDatasetEventSampler`) and see how it works all together...

cdeil:
I think `MapCoord` doesn't support `Time` properly yet, whereas `EventList` does. But yes, either a `MapCoord` or `EventList` based interface for event sampling is reasonable.

fabiopintore:
This PR continued in #2496

Issue #2460, opened 2019-10-15
Add position error info for spatial models


cdeil:
In #2449 I noticed that we don't have an API to get and set position error information for spatial models, and left TODO items here specifically for the 2HWC catalog to fill this information:

https://github.com/gammapy/gammapy/blob/6821570a87bb6693394c8a4ebe535e1770b82c38/gammapy/catalog/hawc.py#L143

https://github.com/gammapy/gammapy/blob/6821570a87bb6693394c8a4ebe535e1770b82c38/gammapy/catalog/tests/test_hawc.py#L84-L85

This is something that is very commonly needed, most users want to write this into their papers and plot it, but it's not so easy to compute, and it's not obvious how to add this to Gammapy.

Just as background to what I mean, see here:
http://www.star.bris.ac.uk/~mbt/topcat/sun253/sun253.html#form-skycorr
The assumption is to have a Gaussian PDF (sometimes symmetric, sometimes elongated) on the Sky, and to have the position error represented by an ellipse or circle. So we could use [regions.EllipseSkyRegion](https://astropy-regions.readthedocs.io/en/latest/api/regions.EllipseSkyRegion.html) or [DiskSpatialModel](https://docs.gammapy.org/0.14/api/gammapy.modeling.models.DiskSpatialModel.html) to represent the position error information, or just have a dict with major, minor, angle, frame information.

Probably we should add get and set methods to the spatial model base class, that transform these numbers to / from the model covariance matrix? How about taking `EllipseSkyRegion` in the get and set? For the "get" this would be then consistent with getting a region that represents the source orientation and extent (see #2422).

And then fill position error information via this API when making spatial models from catalog information (instead of doing `lon_err / cos(lat)` manually)?

@adonath @QRemy @registerrier - Thoughts?

Anyone interested to make this addition?

QRemy:
Added in #2491 and #2501

Issue #2422, opened 2019-09-30
Add SpatialModel to_region method


cdeil:
I think it would be nice to have a `to_region` method on the spatial models, that creates a matching `regions.SkyRegion` object.

This is very nice for debugging, e.g. it's common to plot a residual map, and on top of it one often wants to illustrate the model components with regions.

It's pretty straightforward to add and test.

For Gaussians there's the question what radius to put. I think there we could add an option, and maybe default to 2 sigma or some containment fraction, since 1 sigma only contains 40% of the emission and is very small compared to e.g. disk where we would put `radius` and it's at 100% containment.

For maps we could put a polygon with the four corners, or even have a footprint with more points (e.g. 10 per side) for better accuracy of the outline. This has been implemented before, in [WCS.calc_footprint](https://docs.astropy.org/en/stable/api/astropy.wcs.WCS.html#astropy.wcs.WCS.calc_footprint) and I think also somewhere in Gammapy.

I think the Fermi catalogs in DS9 format like [8yr_catalog/gll_psc_v20.reg](https://fermi.gsfc.nasa.gov/ssc/data/access/lat/8yr_catalog/gll_psc_v20.reg) do contain ellipses for the extended sources:
```
fk5;point( 359.9816,-37.6160)# text={4FGL J2359.9-3736}
fk5;ellipse( 345.4940, 58.9200,  0.2488,  0.2488, 90.0000)# text={4FGL J2301.9+5855e}
```

@luca-giunti or @QRemy or anyone - interested to work on this?

cdeil:
Implemented by @QRemy in #2491.

PR #2491, opened 2019-10-27
Add SpatialModel.position_error and SpatialModel.to_region

QRemy:
This PR implements #2460 and #2422 .

- add SpatialModel.position_error attribute as DiskSpatialModel (or None)
- set position_error for supported catalogs
- add SpatialModel.getcontour() method, return a matplotlib path object and a ds9 .reg string
- add models_to_reg() in serialize in order to export a list of SpatialModel to ds9 .reg file,
so is can be used to generate position error and extended sources contours in .reg,
Use polygons to export contours as in ds9, except for DiskSpatialModel where an ellipse is used (whatever the contour threshold selected).

For example:
[4fgl_extended_20percentmax.reg.txt](https://github.com/gammapy/gammapy/files/3776213/4fgl_extended_20percentmax.reg.txt)
[4fgl_extended_50percentmax.reg.txt](https://github.com/gammapy/gammapy/files/3776214/4fgl_extended_50percentmax.reg.txt)
[4fgl_position_error.reg.txt](https://github.com/gammapy/gammapy/files/3776215/4fgl_position_error.reg.txt)


cdeil:
@QRemy - Please use sky region objects in the API (see #2460 and #2422  and https://astropy-regions.readthedocs.io), both for the position error, and for the model extent.

I think `spatial_model.to_region` should just return the appropriate region object, i.e. often point or circle or ellipse. E.g. rasterising and contouring a point or small source won't yield good results, users will have to fiddle with binsz and other paramters too much. My suggestion would be to delete the contouring code completely. We could add it in the future, either on [TemplateSpatialModel](https://docs.gammapy.org/0.14/api/gammapy.modeling.models.TemplateSpatialModel.html), but probably better as a method on `gammapy.maps.WCSNdMap` or in astropy-regions.
Note that there is contouring code already available in http://docs.astropy.org/en/stable/visualization/wcsaxes/images_contours.html - that could be used for now, and we could document or add an API to transform back and forth between MPL contour points and polygon region objects.

Concerning the code how to transform back and forth between covariance matrix and ellipse parameters, see https://docs.astropy.org/en/stable/_modules/astropy/modeling/functional_models.html#Gaussian2D or https://multinorm.readthedocs.io/en/latest/_modules/multinorm.html#MultiNorm.error_ellipse
If you figure out the other direction (ellipse to covariance), let me know and I'll add it as a helper classmethod also on MultiNorm (and then soon move that functionality to Gammapy to the Parameters class).

gammapy/modeling/serialize.py
```diff
@@ -3,11 +3,35 @@
...
+__all__ = [
+    "regions_to_reg",
+    "models_to_dict",
+    "dict_to_models",
+    "dict_to_datasets",
+    "datasets_to_dict",
+]
+
+
+def regions_to_reg(regions, filename, **kwargs):
```
cdeil: Please remove this helper function.
Ideally there shouldn't be `None` sometimes, it's better to fill `NaN` and always return the same object.

Even if you keep like this, then filtering out the `None` is just one line:
```
regions = [_ for _ in regions if _ is not None]
```
Let's not introduce a function for that, just put that 1 line in the 1-2 places where you need it.

gammapy/modeling/tests/test_serialize_yaml.py
```diff
@@ -1,18 +1,38 @@
...
+@requires_data()
+def test_models_to_reg(tmpdir):
+    model1 = DiskSpatialModel(
+        0 * u.deg, 0 * u.deg, 1 * u.deg, 0.8, 30 * u.deg, frame="galactic"
+    )
+    model2 = GaussianSpatialModel(5 * u.deg, 1 * u.deg, 0.3 * u.deg, frame="galactic")
+    filename = tmpdir / "contours.reg"
+    fileref = get_pkg_data_filename("data/contours.reg")
+    regions_to_reg([model1.to_region, model2.to_region], filename)
+    assert cmp(filename, fileref)
```
cdeil: You just have 2 regions. You can assert against those either by using the Python object, or if you want str compare, put the ref output directly in this test code file.

gammapy/modeling/models/spatial.py
```diff
@@ -521,3 +598,13 @@ def to_dict(self):
...
+
+    @property
+    def to_region(self):
+        """Return model as a `~regions.RectangleSkyRegion`."""
+        return RectangleSkyRegion(
+            center=self.position,
+            width=self.map.geom.width[0][0],
+            height=self.map.geom.width[1][0],
+            angle=0 * u.deg,
+        )
```
cdeil: Does this work? I think it might sometimes, but sometimes FITS images have a rotation in the WCS and then this comes out incorrectly?

Probably the right solution is to compute a footprint from the corners and to return it as a SkyPolygonRegion (or later to sub-sample the sides for better resolution). So call [WCS.calc_footprint](https://docs.astropy.org/en/stable/api/astropy.wcs.WCS.html#astropy.wcs.WCS.calc_footprint) and fill the points into a SkyPolygonRegion.

Suggest to leave at least a TODO comment to check that and add  a test for this later - or file a reminder follow-up issue referencing this comment.

gammapy/modeling/models/spatial.py
```diff
@@ -57,6 +92,11 @@ def to_dict(self):
         data["parameters"] = data.pop("parameters")
         return data
 
+    @abc.abstractmethod
+    def to_region(self):
+        """Return model as a `~regions.SkyRegion`."""
```
cdeil: "model" -> "model outline"

gammapy/modeling/models/spatial.py
```diff
@@ -45,6 +52,34 @@ def position(self):
...
+            sub_covar[1, 0] *= np.cos(self.lat_0.quantity).value
+            eig_vals, eig_vecs = np.linalg.eig(sub_covar)
+            lon_err, lat_err = np.sqrt(eig_vals)
+            y_vec = eig_vecs[:, 0]
+            phi = (np.arctan2(y_vec[1], y_vec[0]) * u.rad).to("deg")
+            err = np.sort([lon_err, lat_err])
+            if err[1] == lon_err:
+                phi += 90 * u.deg
+            return EllipseSkyRegion(
+            center=self.position,
```
cdeil: Use `black` to format code here.

gammapy/modeling/models/spatial.py
```diff
@@ -45,6 +52,34 @@ def position(self):
...
             raise ValueError("Model does not have a defined center position")
 
+    _position_error = None
+
+    @property
+    def position_error(self):
+        """Set position error from covariance matrix"""
+        if self.parameters.covariance is None:
+            return self._position_error
+        else:
```
cdeil: Suggest to remove the `else:` and de-dent this.
I.e. handle the abnormal case first and return early.
And then have the main code at the left-most indentation level.

gammapy/catalog/hess.py
```diff
@@ -516,6 +517,10 @@ def spatial_model(self):
             model = GaussianSpatialModel(
                 lon_0=glon, lat_0=glat, sigma=d["Size"], frame="galactic"
             )
+            if not np.isnan(d["GLON_Err"]):
```
cdeil: You have this `if np.isnan` code  many times.
Can this be avoided?

I think just passing `NaN` along is fine, just remove the ifs?

QRemy:
If the errors are given in the catalog, like in the HGPS, I go through covar (but only the diagonal terms are set), otherwise I set directly the position error given in the catalog through this _position_error.

gammapy/modeling/models/spatial.py
```diff
@@ -57,6 +94,11 @@ def to_dict(self):
         data["parameters"] = data.pop("parameters")
         return data
 
+    @abc.abstractmethod
```
cdeil: Suggest to delete this abstractmethod in the base class. The method `to_region` isn't there e.g. for the SpatialConstant model, and then linters show warnings when sub-classes don't implement the ABC interface.
The normal solution in Python is to just not use ABC, i.e. delete this abstract method in the base class.

gammapy/modeling/models/spatial.py
```diff
@@ -107,6 +149,11 @@ def evaluate_geom(self, geom):
         w = self._grid_weights(x, y, x0, y0)
         return w / geom.solid_angle()
 
+    @property
+    def to_region(self):
```
cdeil: It's weird to have `model.to_region` as a property. It should be either `model.to_region()` (method call), or `model.region` (property).

My suggestion would be to change everywhere to this and remove the `@property`
```
def to_region(self, **kwargs):
    return SomeRegion(..., ***kwargs)
```
Passing along kwargs to the sky region init means that callers can e.g. set colors or styles to make it pretty how they like.
Methods are also the more flexible way, in case we want to add arguments to the API later, we can without breaking all callers. With properties you have to be sure that you'll never want arguments.

gammapy/modeling/models/spatial.py
```diff
@@ -45,6 +52,36 @@ def position(self):
...
+
+    @property
+    def position_error(self):
+        """Set position error from covariance matrix"""
+        if self.parameters.covariance is None:
+            return self._position_error
+        pars = self.parameters
+        idx = [pars._get_idx("lon_0"), pars._get_idx("lat_0")]
+        sub_covar = pars.covariance[idx, :][:, idx]
+        cos_lat = self.lat_0.quantity.to_value("rad")
```
cdeil: Put the `np.cos` call on this line, to make the variable name `cos_lat` true, and then remove the `np.cos` call on the three next lines.

gammapy/catalog/gammacat.py
```diff
@@ -357,6 +358,10 @@ def spatial_model(self):
             raise NoDataAvailableError(f"No spatial model available: {self.name}")
         else:
             raise NotImplementedError(f"Unknown spatial model: {morph_type!r}")
+        model._position_error = CircleSkyRegion(
```
cdeil: It would be better if you change everywhere to set like this:
```
pos_err = CircleSkyRegion(...)
model.set_position_error(position_error)
```
For now, you can still store away in `model._position_error`.
But soon, we will change to store away in covar instead.
And if you put that setter helper method now, you don't have to update all callers later.

Generally, when you get or set a private property like `model._position_error` from outside the class, it's not a good and clean API and needs to be changes.

In this case, the best final API I can imagine is this error regions-based one, where we can get and set on the model object.

gammapy/catalog/fermi.py
```diff
@@ -185,6 +186,22 @@ def _info_lightcurve(self):
     def is_pointlike(self):
         return self.data["Extended_Source_Name"].strip() == ""
 
+    def _set_position_error(self, model):
+        d = self.data
+        if "Conf_68_PosAng" in d:
+            percent="68"
```
cdeil: As discussed on Slack, this is not correct.
You should implement the inverse of the formula here:
https://github.com/gammapy/gammapy/blob/35d52370f283ad250d17bfc6af0803247afa3b1f/gammapy/utils/gauss.py#L88
and then call it like this:
```
sigma = Gauss2D.from_containment(0.68).sigma
```

It's the catalog source class responsibility to go to a uniform format when creating the model object.
So if 3FHL requires completely different code or the if / else becomes too complex, then you have to put a method in the 3FHL class, even if it means some code duplication.
It's always possible to have a helper method on the base class or anywhere and call that from multiple places.

gammapy/modeling/serialize.py
```diff
@@ -7,7 +7,12 @@
 # TODO: move this elsewhere ?
 DATASETS = Registry([MapDataset, SpectrumDataset, FluxPointsDataset])
 
-__all__ = ["models_to_dict", "dict_to_models", "dict_to_datasets", "datasets_to_dict"]
+__all__ = [
```
cdeil: I think there's a difference in line length in the black formatter?
Can you please try this?
```
pip install -U black
black gammapy/modeling/serialize.py gammapy/catalog/fermi.py
```
It should revert this unrelated linebreak  change which is not according to our coding format.

PR #2493, opened 2019-10-28
Add region info to `CountsSpectrum` and adapt tutorials

adonath:
This PR add region information to the `CountsSpectrum` object and adapts the tutorials to use this information for plotting. It concludes the following changes:

- Add `CountsSpectrum.region` attribute
- Implement `compound_region_to_list()` and `list_to_compound_region()` helper functions
- Implement `CountsSpectrum.plot_region()` method
- Implement `plot_spectrum_datasets_off_regions` helper function
- Adapt tutorials to the new `SpectrumDatasetMaker` reduction scheme

### DIFF
--- a/gammapy/spectrum/core.py
+++ b/gammapy/spectrum/core.py
@@ -9,6 +9,7 @@ from gammapy.maps import MapAxis
 from gammapy.maps.utils import edges_from_lo_hi
 from gammapy.utils.fits import ebounds_to_energy_axis, energy_axis_to_ebounds
 from gammapy.utils.scripts import make_path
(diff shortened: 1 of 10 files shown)


adonath:
Thanks @registerrier! I agree we have to think about how to handle the region information when stacking. I guess up to a certain point it makes sense to keep it. On the other hand if you stack O(1e3) observations and each of the observations has >10 off regions, you end up with a lot of regions info, that you might not want to keep. The same question goes for the serialisation I guess. I think ideally the region info would be serialised along with the data, but for a stacked observation with O(1e3) individual regions it just becomes complex...

PR #2498, opened 2019-10-30
Add mask_safe handling in MapDataset.to_image

luca-giunti:


gammapy/cube/fit.py
```diff
@@ -867,11 +868,20 @@ def to_image(self, spectrum=None, keepdims=True):
         dataset : `MapDataset`
             Map dataset containing images.
         """
+        self.counts.data[~self.mask_safe.data] = 0
```
adonath: Note that this modifies the data in place and leaves the parent `MapDataset` in a modified state. This we should avoid and make a copy of the data instead. This can be achieved e.g. by just multiplying the mask as well:

```python3
from gammapy.maps import WcsNDMap
import numpy as np

m = WcsNDMap.create(width=(8, 5))
m.data += 1

mask = np.ones(m.geom.data_shape)
mask[10:20] = 0

m_masked = m * mask

m_masked.plot()
```

We could also add a `weights` argument to `.sum_over_axis()`.

gammapy/cube/fit.py
```diff
@@ -867,11 +868,20 @@ def to_image(self, spectrum=None, keepdims=True):
...
+        self.background_model.evaluate().data[~self.mask_safe.data] = 0
+
         counts = self.counts.sum_over_axes(keepdims=keepdims)
         exposure = _map_spectrum_weight(self.exposure, spectrum)
         exposure = exposure.sum_over_axes(keepdims=keepdims)
         background = self.background_model.evaluate().sum_over_axes(keepdims=keepdims)
 
+        slices_num = self.mask_safe.geom.get_axis_by_name("ENERGY").nbin
```
adonath: Note that the `+` operator corresponds to the logical or operation:
```python3
import numpy as np
mask_1 = np.array([[1, 0]], dtype=bool)
mask_2 = np.array([[1], [0]], dtype=bool)

print(mask_1 + mask_2)
print()
print(np.logical_or(mask_1, mask_2))
```

So you can basically use `.sum_over_axis()` for the mask as well (as long as the `dtype` is `bool`)

gammapy/cube/fit.py
```diff
@@ -867,11 +868,20 @@ def to_image(self, spectrum=None, keepdims=True):
...
+        mask_safe_image = self.mask_safe.slice_by_idx({"energy": 0})
+        for idx in range(1, slices_num):
+            mask_slice = self.mask_safe.slice_by_idx({"energy": idx})
+            mask_safe_image = mask_safe_image or mask_slice
```
cdeil: The `or` operator doesn't work for masks.

You should use `|` or `&` operators always:

```
In [6]: np.array([0, 0, 1, 1], dtype=bool) & np.array([1, 0, 1, 0], dtype=bool)                                                                                                                   
Out[6]: array([False, False,  True, False])

In [7]: np.array([0, 0, 1, 1], dtype=bool) or np.array([1, 0, 1, 0], dtype=bool)                                                                                                                  
ValueError                                Traceback (most recent call last)
<ipython-input-7-8bce39258b03> in <module>
----> 1 np.array([0, 0, 1, 1], dtype=bool) or np.array([1, 0, 1, 0], dtype=bool)

ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

Don't use `+`, it's not appropriate. Always use `&` (logical and) or `|` (logical or)`.

adonath: I agree, I think the simplest and clearest expression to reduce the 3D mask would be:

```python3
idx = mask_safe.geom.get_axis_index_by_name("energy")
mask_image = np.logical_or(mask_safe, axis=idx)
```

adonath:
Thanks @luca-giunti!

PR #2505, opened 2019-11-01
Add a tutorial for joint 1d/3d analysis

QRemy:
This tutorial illustrates how to run a joint analysis with different datasets. We look at the gamma-ray emission from the Crab nebula between 10 GeV and 100 TeV. The spectral parameters are optimized by combining a 3D analysis of Fermi-LAT data, a ON/OFF spectral analysis of HESS data, and flux points from HAWC.

### DIFF
--- a/tutorials/joint_1d_3d_analysis.ipynb
+++ b/tutorials/joint_1d_3d_analysis.ipynb
@@ -1,0 +1,405 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Joint Crab analysis with 1D and 3D datasets"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "This tutorial illustrates how to run a joint analysis with different datasets.\n",
+    "We look at the gamma-ray emission from the Crab nebula between 10 GeV and 100 TeV.\n",
+    "The spectral parameters are optimized by combining a 3D analysis of Fermi-LAT data, a ON/OFF spectral analysis of HESS data, and flux points from HAWC.  \n",
+    "\n",
+    "\n"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "#%load_ext nb_black\n",
+    "import os\n",
+    "import numpy as np\n",
+    "from astropy import units as u\n",
+    "import matplotlib.pyplot as plt\n",
+    "from gammapy.modeling import Fit, Datasets\n",
+    "from gammapy.spectrum import (\n",
+    "    FluxPoints,\n",
+    "    FluxPointsEstimator,\n",
+    "    FluxPointsDataset,\n",
+    "    SpectrumDatasetOnOff,\n",
+    ")\n",
+    "from pathlib import Path"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Data and models files\n",
+    "\n",
+    "We are going to use pre-made datasets. Links toward other tutorials detailing on how to prepare datasets are given at the end.\n",
+    "The datasets serialization produce YAML files listing the datasets and models. In the following cells we show an example containning only the Fermi-LAT dataset and the Crab model.\n",
+    "\n",
+    "Fermi-LAT-3FHL_datasets.yaml:"
+   ]
+  },
+  {
+   "cell_type": "raw",
+   "metadata": {},
+   "source": [
+    "datasets:\n",
+    "- name: Fermi-LAT\n",
+    "  type: MapDataset\n",
+    "  likelihood: cash\n",
+    "  models:\n",
+    "- Crab Nebula\n",
+    "  background: background\n",
+    "  filename: $GAMMAPY_DATA/fermi-3fhl-crab/Fermi-LAT-3FHL_data_Fermi-LAT.fits"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Fermi-LAT-3FHL_models.yaml:"
+   ]
+  },
+  {
+   "cell_type": "raw",
+   "metadata": {},
+   "source": [
+    "components:\n",
+    "- name: Crab Nebula\n",
+    "  type: SkyModel\n",
+    "  spatial:\n",
+    "    type: PointSpatialModel\n",
+    "    frame: icrs\n",
+    "    parameters:\n",
+    "    - name: lon_0\n",
+    "      value: 83.63310241699219\n",
+    "      unit: deg\n",
+    "      min: .nan\n",
+    "      max: .nan\n",
+    "      frozen: true\n",
+    "    - name: lat_0\n",
+    "      value: 22.019899368286133\n",
+    "      unit: deg\n",
+    "      min: -90.0\n",
+    "      max: 90.0\n",
+    "      frozen: true\n",
+    "  spectral:\n",
+    "    type: LogParabolaSpectralModel\n",
+    "    parameters:\n",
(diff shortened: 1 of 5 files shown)


tutorials/joint_1d_3d_analysis.ipynb
```diff
@@ -0,0 +1,1314 @@
...
+      ]
+     },
+     "metadata": {},
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "FGL4 = SourceCatalog4FGL()\n",
+    "crab = FGL4[\"Crab Nebula\"]  #\n",
+    "crab_mod = crab.sky_model()\n",
```
registerrier: maybe `crab_model` is a more explicit name

tutorials/joint_1d_3d_analysis.ipynb
```diff
@@ -0,0 +1,1314 @@
...
+    "    reference=crab.data[\"Pivot_Energy\"],\n",
+    ")\n",
+    "print(crab_mod)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "We use the events associated to 3FHL catalog in order to a produce counts cube. We select a 10-by-8 degree region centered on the Crab in Galactic coordinates, and photon energies between 10 GeV and 2 TeV.  "
```
registerrier: I would explain that this is not performing a real data reduction of Fermi data but simply taking reduced products from the 3FHL catalog to create our own dataset.

Also a 10x8 degree map seems large for data above 10 GeV no?

tutorials/joint_1d_3d_analysis.ipynb
```diff
@@ -0,0 +1,1314 @@
...
+    "e_reco = counts.geom.axes[0].edges\n",
+    "edisp = EnergyDispersion.from_diagonal_response(e_true=e_true, e_reco=e_reco)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Given the PSF of Fermi-LAT, sources outside the analysis region can be reconstructed inside. So we have to a define mask that excluded pixels close to our region boundary. The size of this margin is set as the PSF 99% containment radius."
+   ]
```
registerrier: a mask that excludes

tutorials/joint_1d_3d_analysis.ipynb
```diff
@@ -0,0 +1,1314 @@
...
+    "bkg_iso = eval_iso.compute_npred()\n",
+    "\n",
+    "# merge iem and iso  backgrounds for simplicity\n",
+    "background_total = bkg_iso + bkg_iem\n",
+    "background_model = BackgroundModel(background_total)\n",
+    "\n",
+    "background_total.sum_over_axes().smooth(2).plot(\n",
+    "    stretch=\"log\", vmax=50, cmap=\"nipy_spectral\"\n",
+    ");"
+
```
registerrier: I think that this cell shows that we need a helper function to create a `BackgroundModel` from a `SkyDiffuseCube`. I don't think `MapEvaluator` should be something a typical user has to work with.

registerrier: For this part, using the high level interface might be nice.

registerrier: Also this part should not appear here. We should ship an exclusion mask for the crab in the data and provide an howto to explain how to create one from a catalog.

registerrier: I would explain here that we need to create a dataset object by adding a model to the flux points. This might not be obvious to a user.

registerrier: Is this part useful here? You don't use the covariance matrix here right?

registerrier: I would add "HESS DL3 DR1" to make it clear this is a specific set of data

registerrier: Maybe "Fermi-LAT - 3FHL"

cdeil:
@QRemy - Thanks!

Please list the notebook in [tutorials/notebooks.yaml](https://github.com/gammapy/gammapy/blob/master/tutorials/notebooks.yaml), so that it gets executed and tested as part of continuous integration.

Also, please add the notebook in the toctree and as a bullet point in [docs/tutorials.rst](https://github.com/gammapy/gammapy/blob/master/docs/tutorials.rst) so that it shows up in the Gammapy docs. Test locally with `make docs-all` and `make docs-show` that the tutorial notebook does show up, and that it looks OK.

Please try to add more level 2 section headings, and keep them very short. Example: https://docs.gammapy.org/dev/notebooks/maps.html
The reason is that the section headings are part of the navigation sidebar, so very long names don't fit there. And they are anchors one can link to when sharing URLs to show colleagues how something works, or also we can link to sections from other parts of the Gammapy docs (e.g. the HOWTO we're starting). So if you have very long sections, it's hard to link to or tell people where something is.

I didn't read through the tutorial. @registerrier already made some suggestions, and probably @adonath wants to review this as well. So I'll just leave some very general suggestions:

1. Suggest to only polish this for ~ 1 day for now and then go back to improving datasets code. (and then come back to this tutorial and other docs in the coming weeks)
2. Try to shorten this. Overall the notebook is very long and I think ~ 70% of this tutorial does data reduction and duplicates other tutorials? The main focus of this tutorial should be on datasets serialisation / loading, and multi-dataset fitting.
3. Does it run fast? Aim for 10 - 20 sec to execute the tutorial. If it takes much longer, do something to make it run faster (e.g. smaller FOV or less data, or better fit start parameters). Please add a `%%timeit` to the one or few cells that take the longest to execute - IMO it's nice for readers to see where usually the heavy computations will happen.
4. Please never commit Jupyter notebook output cells. They usually contain images and are a few MB per notebook, which if we put this in our version history, with re-execution and many notebooks quickly  snowballs to a very large Gammapy code repository which we want to avoid. So please use `make polish` to clean the output cells and amend the commit here, to actually not have the images in the version history. If you want to share a rendered notebook to make review easier (which is very useful) - post it in a https://gist.github.com/ paste the Gist URL in https://nbviewer.jupyter.org/ and share that URL here in the PR in a comment, when asking for review.

QRemy:
I splitted the datasets preparation from the analysis part. So now the analysis tutorial only read pre-made files. It takes about 15s to run.

Can we keep the preparation tutorials somewhere without listing it directly in the doc ?
I would like to give only a link at this end of this tutorial for users that would like to see how the files have been generated without having to go through 3 different tutorials.

tutorials/joint_1d_3d_analysis.ipynb
QRemy: It's used by crab_spec.plot_error().

tutorials/joint_1d_3d_analysis.ipynb
QRemy: yes, reduced to 5x4

tutorials/joint_1d_3d_analysis.ipynb
QRemy: Maybe, for now I just moved all the datasets preparation steps in a separate script and only read pre-made files in the analysis script.

QRemy: Same comment that above.

cdeil:
@QRemy - You still have the commit with the output PNG in the notebook in this branch:

https://github.com/gammapy/gammapy/pull/2505/commits/8fc7542bf98e5af48a0755a934e0bd656cc5cf09#diff-cff32e9881017cdb9cc0e619eaa2b17dR197

Could you please squash the commits here to avoid adding this to our git history?

cdeil:
Looks like this is failing because you use an absolute path that only works on your machine:
https://travis-ci.org/gammapy/gammapy/jobs/607673321#L1835
You have to use a relative path to make this work on other machines.

adonath:
Thanks @QRemy!

PR #2506, opened 2019-11-04
Create unique parameters list in Datasets

cdeil:
This PR moves the creation of unique parameters from `Parameters.__init__` to `Datasets.parameters`.

I think that's much better to have that as an explicit operation, not something that silently occurs on `Parameters.__init__`. Some arguments were given in #2494 .

Unfortunately we don't have any unit tests for `Datasets` yet, but there were three failing tests if I simply removed the unique operation, one for serialsation, one for 1D and one for 3D analysis. With this change they all pass.

I would suggest to merge this as-is, as a very small, but important change, and then to continue with covariance and parameters stacking related changes in a follow-up PR.

### DIFF
--- a/gammapy/modeling/datasets.py
+++ b/gammapy/modeling/datasets.py
@@ -101,7 +101,13 @@ class Datasets:
 
     @property
     def parameters(self):
-        return Parameters.from_stack(_.parameters for _ in self.datasets)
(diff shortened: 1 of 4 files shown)


adonath:
Thanks @cdeil!

PR #2508, opened 2019-11-04
Update to use new hess DL3 background models

adonath:
This PR updates the data download index file to contain the new DL3 background models it also updates the tutorials and existing test to make uses of it.


cdeil:
There's one fail: https://travis-ci.org/gammapy/gammapy/jobs/607089083#L3099
I think this part of the `Background3D` code was always broken and just not used so far.
@adonath - Please apply a minimal fix here, or xfail this and merge this in, and I could send a follow-up PR within the next 1-2 days.

PR #2520, opened 2019-11-06
Refactor ring background maker

luca-giunti:
In this PR, I refactor the `RingBackgroundEstimator` into `RingBackgroundMaker` and `AdaptiveRingBackgroundEstimator` into `AdaptiveRingBackgroundMaker`. For the moment, those are still two independent makers, but could eventually be merged into a single maker later on.

I compared the performances of the new "maker" pattern to the old scheme. You can see this comparison in two notebooks:
- Using v0.13: https://github.com/luca-giunti/share/blob/master/ring_13.ipynb
-  Using the new class: https://github.com/luca-giunti/share/blob/master/ring_dev.ipynb

Using 4 runs, the computation time for the non-adaptive ring algorithm drops from `5.46 s ± 469 ms` to `2.61 s ± 125 ms`, and for the adaptive ring from `7.52 s ± 622 ms` to `3.08 s ± 334 ms`.

Also,the results look very similar. However, as you can see in the notebooks, there is a small shift in `alpha`: taking a random bin, I find a shift from `0.00077920605` to `0.000740094`. I think this might be related to the fact that in v0.13 we were stacking the off_exposure using a simple `fill_by_coord()` (see [here](https://github.com/gammapy/gammapy/blob/7d06f74d407f42a164477da8d8922bd70c93ae76/gammapy/cube/make.py#L553)), i.e. no weighting by the obs time or off counts was performed... Is that possible?

### DIFF
--- a/gammapy/cube/fit.py
+++ b/gammapy/cube/fit.py
@@ -1011,10 +1011,9 @@ class MapDatasetOnOff(MapDataset):
     @property
     def alpha(self):
         """Exposure ratio between signal and background regions"""
-        tmp = self.acceptance / self.acceptance_off
-        mask = np.where(np.isnan(tmp.data))
-        tmp.data[mask] = 0
-        return tmp
+        alpha = self.acceptance / self.acceptance_off
+        alpha.data = np.nan_to_num(alpha.data)
+        return alpha
 
     @property
     def background(self):
--- a/gammapy/cube/fit.py
+++ b/gammapy/cube/fit.py
@@ -1011,7 +1011,10 @@ class MapDatasetOnOff(MapDataset):
     @property
     def alpha(self):
         """Exposure ratio between signal and background regions"""
-        return self.acceptance / self.acceptance_off
+        tmp = self.acceptance / self.acceptance_off
+        mask = np.where(np.isnan(tmp.data))
+        tmp.data[mask] = 0
+        return tmp
 
     @property
     def background(self):
--- a/gammapy/cube/fit.py
+++ b/gammapy/cube/fit.py
@@ -877,8 +877,8 @@ class MapDataset(Dataset):
         background = background.sum_over_axes(keepdims=keepdims)
 
         idx = self.mask_safe.geom.get_axis_index_by_name("ENERGY")
-        data = np.logical_or.reduce(self.mask_safe.data, axis=idx)
-        mask_image = WcsNDMap(geom=self.mask_safe.geom.to_image(), data=data)
+        data = np.logical_or.reduce(self.mask_safe.data, axis=idx, keepdims=keepdims)
+        mask_image = WcsNDMap(geom=counts.geom, data=data)
 
         # TODO: add edisp and psf
         edisp = None
--- a/gammapy/cube/ring.py
+++ b/gammapy/cube/ring.py
@@ -4,7 +4,7 @@ import itertools
 import numpy as np
 from astropy.convolution import Ring2DKernel, Tophat2DKernel
 from astropy.coordinates import Angle
-from gammapy.maps import scale_cube
+from gammapy.maps import scale_cube, Map
 from gammapy.cube.fit import MapDatasetOnOff
 
 __all__ = ["AdaptiveRingBackgroundMaker", "RingBackgroundMaker"]
@@ -33,6 +33,9 @@ class AdaptiveRingBackgroundMaker:
         Integration radius used for alpha computation.
     method : {'fixed_width', 'fixed_r_in'}
         Adaptive ring method.
+    exclusion : `~gammapy.maps.WcsNDMap`
+        Exclusion mask for regions with known gamma-ray emission.
+
 
     See Also
     --------
@@ -152,8 +155,6 @@ class AdaptiveRingBackgroundMaker:
         ----------
         dataset : `~gammapy.cube.fit.MapDataset`
             Input map dataset.
-        exclusion : `~gammapy.maps.WcsNDMap`
-            Exclusion mask for regions with known gamma-ray emission.
 
         Returns
         -------
@@ -164,19 +165,11 @@ class AdaptiveRingBackgroundMaker:
         counts = dataset.counts
         background = dataset.background_model.map
         kernels = self.kernels(counts)
-        exclusion = self.exclusion_mask
 
-        if (
-            exclusion is not None
-            and exclusion.geom.data_shape != counts.geom.data_shape
-        ):
-            # Reproject the exclusion mask to the cutout geom
-            cutout_coord = counts.geom.get_coord()
-            reproj_exclusion = counts.copy(data=np.zeros(cutout_coord.shape))
-            reproj_exclusion.fill_by_coord(
-                cutout_coord, exclusion.get_by_coord(cutout_coord)
-            )
-            exclusion = reproj_exclusion
+        # reproject exclusion mask
+        coords = counts.geom.get_coord()
+        data = self.exclusion_mask.get_by_coord(coords)
+        exclusion = Map.from_geom(geom=counts.geom, data=data)
 
         cubes = {}
         cubes["counts_off"] = scale_cube(
@@ -296,8 +289,6 @@ class RingBackgroundMaker:
         ----------
         dataset : `~gammapy.cube.fit.MapDataset`
             Input map dataset.
-        exclusion : `~gammapy.maps.WcsNDMap`
-            Exclusion mask for regions with known gamma-ray emission.
 
         Returns
         -------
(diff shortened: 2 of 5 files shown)


gammapy/cube/ring.py
```diff
@@ -313,51 +291,81 @@ def kernel(self, image):
...
-            result["alpha"].data[not_has_exposure] = 0
+    def run(self, dataset, exclusion):
+        """Run ring background maker
 
-        result["background_ring"] = result["alpha"] * result["off"]
+        Parameters
+        ----------
+        dataset : `~gammapy.cube.fit.MapDataset`
+            Input map dataset.
+        exclusion : `~gammapy.maps.WcsNDMap`
```
adonath: In the `ReflectedRegionsBackgroundMaker` the `exclusion_mask` is taken on `__init__`. I think for consistency it would be good to also move it to `RingBackgroundMaker.__init__()`. Alternatively we could also attach it to the dataset, but this we should discuss as separate issue.

gammapy/cube/ring.py
```diff
@@ -131,122 +110,139 @@ def kernels(self, image):
...
+            On off dataset.
+        """
+        cubes = self.make_cubes(dataset, exclusion)
+        acceptance, acceptance_off, counts_off = self._reduce_cubes(cubes, dataset)
 
+        not_has_off_acceptance = acceptance_off.data <= 0
+        acceptance_off.data[not_has_off_acceptance] = np.nan
 
-class RingBackgroundEstimator:
+        fft_noise_threshold = 1e-6
```
adonath: This hard-coded threshold seem problematic to me. I guess in general the `acceptance_off` is zero, where the ring completely falls into an exclusion region, no? What if we ring-convolve the `exclusion_mask` as well and derive the off mask from this? I guess we can also use `scripy.ndimage.morphology`. Here is a minimal example:
```python3
from astropy.convolution import Ring2DKernel
from scipy.ndimage.morphology import binary_erosion, binary_dilation
from gammapy.maps import Map

binsz= 0.02
center = SkyCoord("0 deg", "0 deg")
circle = CircleSkyRegion(center=center, radius=0.4 *u.deg)

exclusion_mask = Map.create(width=2, binsz=binsz)
exclusion_mask.data = exclusion_mask.geom.region_mask([circle], inside=True)

ring = Ring2DKernel(10, width=5)
ring.normalize("peak")

data = binary_erosion(exclusion_mask.data, structure=ring.array)
mask = exclusion_mask.copy(data=data)
```

gammapy/cube/ring.py
```diff
@@ -131,122 +110,139 @@ def kernels(self, image):
...
+        not_has_acceptance = acceptance.data <= fft_noise_threshold
+        counts_off.data[not_has_acceptance] = 0
+        acceptance_off.data[not_has_acceptance] = 0
+        acceptance.data[not_has_acceptance] = 0
+
+        mask_safe = dataset.counts.copy(
```
adonath: I think we should definitely keep and use the `mask_safe`, probably instead of modifying the data, as you did above. My naive proposal would be, that the `mask_safe` should be taken into account when convolving the acceptance (instead of relying on it to be zero) and then the `mask_safe` should be modified to exclude regions with no `acceptance_off` (computed as I proposed above). 

I have to think about it once more, to get a clearer picture of how to handle the `mask_safe` here, but I think in general we should leave the data untouched (not setting anything to zero or `NaN`) but keep track of the information in the `mask_safe` instead. When the dataset is stacked the data is modified anyway (set to zero). And this gives a single, well defined place, where the data is touched.

As I said: I'll think about it once more and make a proposal. It should not stop us form merging this PR.

gammapy/cube/ring.py
```diff
@@ -131,122 +110,139 @@ def kernels(self, image):
...
+        return MapDatasetOnOff(
+            counts=dataset.counts,
+            counts_off=counts_off,
+            acceptance=acceptance,
+            acceptance_off=acceptance_off,
+            exposure=dataset.exposure,
+            mask_fit=dataset.mask_fit,
+            psf=dataset.psf,
+            edisp=dataset.edisp,
+            background_model=None,
```
adonath: No need to redefine the default value.

gammapy/cube/ring.py
```diff
@@ -131,122 +110,139 @@ def kernels(self, image):
...
+            name=dataset.name,
+            evaluation_mode="local",
```
adonath: Again no need to redefine the default.

gammapy/cube/tests/test_ring.py
```diff
@@ -2,94 +2,109 @@
...
+
+    assert_allclose(datasets[0].alpha.data[0][100][100], 0.000668635)
+    assert_allclose(datasets[0].exposure.data[0][100][100], 639038346.7895743)
+
+
+@requires_data()
+def test_adaptive_ring_bkg_maker(map_dataset_maker, observations, exclusion_region):
+    datasets = {}
+
+    for method in ["fixed_width", "fixed_r_in"]:
```
adonath: The way we typically test different parametrisation is by using `pytest.mark.parametrise()` see e.g.  https://github.com/gammapy/gammapy/blob/master/gammapy/cube/tests/test_make.py#L35
This has the advantage, that parametrised test fail independently. However in the current implementation it's not too bad, because you have separate asserts. On the other hand `pytest.mark.parametrise()` would reduce the code duplication even more. So I'd prefer to change it...

gammapy/cube/tests/test_ring.py
```diff
@@ -2,94 +2,109 @@
...
+    """Example mask for testing."""
+    pos = SkyCoord(83.633, 22.014, unit="deg", frame="icrs")
+    return CircleSkyRegion(pos, Angle(0.15, "deg"))
+
+
+@pytest.fixture(scope="session")
+def observations():
+    """Example observation list for testing."""
+    datastore = DataStore.from_file(
+        "$GAMMAPY_DATA/hess-dl3-dr1/hess-dl3-dr3-with-background.fits.gz"
```
adonath: One more comment: this currently fails on Travis, because we removed the old background models. Just change this to `DataStore.from_dir("$GAMMAPY_DATA/hess-dl3-dr1")` and it should work again. Locally you have to update you gammapy data folder by deleting it and running `make dataset-download`.

gammapy/cube/fit.py
```diff
@@ -1011,7 +1011,10 @@ def parameters(self):
     @property
     def alpha(self):
         """Exposure ratio between signal and background regions"""
-        return self.acceptance / self.acceptance_off
+        tmp = self.acceptance / self.acceptance_off
+        mask = np.where(np.isnan(tmp.data))
```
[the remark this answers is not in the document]
luca-giunti: Yes, it has an effect. If I remove these two lines, the `acceptance_off` of the stacked map (after the ring background estimation) looks like thistest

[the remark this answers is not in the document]
adonath: Sorry, I didn't notice, that you set `np.nan` explicitly as a placeholder here: https://github.com/gammapy/gammapy/pull/2520/files#diff-5c957f91195d9d25604ec77a67ba3ceeR323 
OK, to keep for now, but I think in general we should try to simplify the data masking as much as possible, basically leaving the data untouched and take into account and modify the `safe_mask` instead where needed (can be done in a follow up PR...).

gammapy/cube/ring.py
luca-giunti: Done. In the new implementation, I had to add a `fill_by_coord()` to reproject internally the `exclusion_mask` to the cutout geom. With 4 runs, I checked that this increases the computation time of ~4%, which is not too bad I guess

adonath: Yes, this is completely acceptable.

luca-giunti:
Thanks @adonath for the review! I implemented what you requested, with the exception of the following comments:
- https://github.com/gammapy/gammapy/pull/2520#discussion_r343655882: As you said, we need a proposal for the `mask_safe` handling
- https://github.com/gammapy/gammapy/pull/2520#discussion_r343651270: I completely agree that the hard-coded value is bad. As you suggested, I tried to implement a solution based on `scripy.ndimage.morphology` but I can't get it to work...maybe, if you have it clear in mind could you add this yourself before merging?

luca-giunti:
Thanks @adonath for the review! I implemented what you requested, with the exception of the following comments:
- https://github.com/gammapy/gammapy/pull/2520#discussion_r343655882: As you said, we need a proposal for the `mask_safe` handling
- https://github.com/gammapy/gammapy/pull/2520#discussion_r343651270: I completely agree that the hard-coded value is bad. As you suggested, I tried to implement a solution based on `scripy.ndimage.morphology` but I can't get it to work...maybe, if you have it clear in mind could you add this yourself before merging?

adonath:
Thanks a lot @luca-giunti! I added a few clean-up commits addressing the following things:
- I removed the `.parameters` attribute from the makers and exposed the parameters directly on the class (mainly for consistency with the other Maker classes)
- I simplified the exclusion mask reprojection a bit to avoid the `.fill_by_coord()` call
- I also simplified the data masking to not modify the data, but just the `mask_safe`, the mask safe is applied consistently on `.stack()` setting the data off `counts_off` and `acceptance_off` to zero.
- I already adapted the image analysis tutorial, because it was failing on Travis-CI

I'll wait for Travis to pass and the go ahead and merge.

PR #2522, opened 2019-11-07
Add sub-covariance matrix management

QRemy:
- parameters.get_subcovar(names): get a sub-covariance matrix from a list of parameter names.
- parameters.set_covar_from_parameters(parameters): set a sub-covariance matrix from a parent or a sub-array of a covariance matrix from a child
- set the sub-covariance matrix of all models after Fit.run()

### DIFF
--- a/gammapy/modeling/fit.py
+++ b/gammapy/modeling/fit.py
@@ -95,6 +95,8 @@ class Fit:
         fit_result : `FitResult`
             Results
         """
+        from gammapy.modeling.models import SkyModels
+
         if optimize_opts is None:
             optimize_opts = {}
         optimize_result = self.optimize(**optimize_opts)
@@ -112,6 +114,17 @@ class Fit:
         # TODO: not sure how best to report the results
         # back or how to form the FitResult object.
         optimize_result._success = optimize_result.success and covariance_result.success
+
+        # set sub-covariance matrix for each model
+        for dataset in self.datasets.datasets:
+            if isinstance(dataset.model, SkyModels):
+                for model in dataset.model.skymodels:
+                    pars = model.parameters
+                    pars.set_subcovar_from_parameters(self._parameters)
+                    model.spatial_model.parameters.set_subcovar_from_parameters(pars)
+                    model.spectral_model.parameters.set_subcovar_from_parameters(pars)
+            else:
+                dataset.model.parameters.set_subcovar_from_parameters(self._parameters)
 
         return optimize_result
 
--- a/gammapy/modeling/fit.py
+++ b/gammapy/modeling/fit.py
@@ -77,7 +77,7 @@ class Fit:
 
     @lazyproperty
     def _parameters(self):
-        return self.datasets.parameters.unique_parameters
+        return self.datasets.parameters
 
     def run(self, optimize_opts=None, covariance_opts=None):
         """
--- a/gammapy/modeling/fit.py
+++ b/gammapy/modeling/fit.py
@@ -95,8 +95,6 @@ class Fit:
         fit_result : `FitResult`
             Results
         """
-        from gammapy.modeling.models import SkyModels
-
         if optimize_opts is None:
             optimize_opts = {}
         optimize_result = self.optimize(**optimize_opts)
@@ -116,15 +114,8 @@ class Fit:
         optimize_result._success = optimize_result.success and covariance_result.success
 
         # set sub-covariance matrix for each model
-        for dataset in self.datasets.datasets:
-            if isinstance(dataset.model, SkyModels):
-                for model in dataset.model.skymodels:
-                    pars = model.parameters
-                    pars.set_subcovar_from_parameters(self._parameters)
-                    model.spatial_model.parameters.set_subcovar_from_parameters(pars)
-                    model.spectral_model.parameters.set_subcovar_from_parameters(pars)
-            else:
-                dataset.model.parameters.set_subcovar_from_parameters(self._parameters)
+        for model in self.datasets._model_list:
+            model.parameters.set_subcovariance(self._parameters)
 
         return optimize_result
 
--- a/gammapy/modeling/fit.py
+++ b/gammapy/modeling/fit.py
@@ -77,7 +77,7 @@ class Fit:
 
     @lazyproperty
     def _parameters(self):
-        return self.datasets.parameters
+        return self.datasets.parameters.unique_parameters
 
     def run(self, optimize_opts=None, covariance_opts=None):
         """
@@ -112,10 +112,6 @@ class Fit:
         # TODO: not sure how best to report the results
         # back or how to form the FitResult object.
         optimize_result._success = optimize_result.success and covariance_result.success
-
-        # set sub-covariance matrix for each model
-        for model in self.datasets._model_list:
-            model.parameters.set_subcovariance(self._parameters)
 
         return optimize_result
 
--- a/gammapy/modeling/models/cube.py
+++ b/gammapy/modeling/models/cube.py
@@ -57,11 +57,8 @@ class SkyModels:
 
     @property
     def parameters(self):
-        parameters = []
-        for skymodel in self.skymodels:
-            for p in skymodel.parameters:
-                parameters.append(p)
-        return Parameters(parameters)
+        parameters = Parameters.from_stack([_.parameters for _ in self.skymodels])
+        return parameters
 
     @property
     def names(self):
--- a/gammapy/modeling/models/cube.py
+++ b/gammapy/modeling/models/cube.py
@@ -57,8 +57,7 @@ class SkyModels:
 
     @property
     def parameters(self):
-        parameters = Parameters.from_stack([_.parameters for _ in self.skymodels])
-        return parameters
+        return Parameters.from_stack([_.parameters for _ in self.skymodels])
 
     @property
     def names(self):
--- a/gammapy/modeling/models/spatial.py
+++ b/gammapy/modeling/models/spatial.py
@@ -77,7 +77,7 @@ class SpatialModel(Model):
                 angle=np.nan * u.deg,
             )
         pars = self.parameters
-        sub_covar = pars.get_covariance(["lon_0", "lat_0"])
+        sub_covar = pars.get_subcovariance(["lon_0", "lat_0"])[0]
         cos_lat = np.cos(self.lat_0.quantity.to_value("rad"))
         sub_covar[0, 0] *= cos_lat ** 2.0
         sub_covar[0, 1] *= cos_lat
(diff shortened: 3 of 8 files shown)


gammapy/modeling/fit.py
```diff
@@ -113,6 +115,17 @@ def run(self, optimize_opts=None, covariance_opts=None):
         # back or how to form the FitResult object.
         optimize_result._success = optimize_result.success and covariance_result.success
 
+        # set sub-covariance matrix for each model
+        for dataset in self.datasets.datasets:
+            if isinstance(dataset.model, SkyModels):
+                for model in dataset.model.skymodels:
+                    pars = model.parameters
+                    pars.set_subcovar_from_parameters(self._parameters)
+                    model.spatial_model.parameters.set_subcovar_from_parameters(pars)
```
adonath: I think this is incomplete, because the model can be a `SkyDiffuseCube` as well. Also setting the covariance for the `background_model` is missing. Instead I would propose a different approach:

- Add a `Datasets._model_list` read only property, which walks all datasets and collects all model components from the individual datasets in a list.

- Then the code above reduces to:
```
for model in datasets._model_list:
    model.parameters.covariance = self._parameters.get_subcovariance(model.parameters)
```

adonath: This would also be a first step towards a global models objects we've discussed in the past week...

gammapy/modeling/parameter.py
```diff
@@ -588,6 +588,34 @@ def freeze_all(self):
...
+        ----------
+        names : list
+           List of string
+        """
+        idx = [self._get_idx(name) for name in names]
+        return self.covariance[idx, :][:, idx]
+
+    def set_subcovar_from_parameters(self, parameters):
+        """Set sub-covariance matrix from a parent or child `~gammapy.modeling.Parameters`."""
+        if len(self.unique_parameters) < len(parameters.unique_parameters):
```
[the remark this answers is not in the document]
QRemy: - If we set model.spatial_model.parameter from results.parameters then len(self)<len(parameters)
- if we set model.parameters from model.spatial_model and model.spectral_model separately then len(self)>len(parameters)

gammapy/modeling/parameter.py
```diff
@@ -588,6 +588,34 @@ def freeze_all(self):
         for par in self._parameters:
             par.frozen = True
 
+    def get_subcovar(self, names):
```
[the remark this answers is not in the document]
QRemy: This is used in spatial_model.position_error as
`sub_covar = pars.get_subcovar(["lon_0", "lat_0"])
`
In such case you have to look for the parameter by name anyway and I think it is more convenient to have this method rather than filtering the objects to define a new parameters instance and then applying set_subcovar_from_parameters.

gammapy/modeling/parameter.py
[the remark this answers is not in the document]
adonath: Shouldn't the latter case be handled in `Parameters.from_stack()`?

gammapy/modeling/parameter.py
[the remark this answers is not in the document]
adonath: As this method is part of the public API I would strongly recommend to only support the safe case (i.e. a `Parameters` object). Note that this is mainly a matter of documentation, because the case you've described above still (silently) works. I would just not expose it in the docstring to prevent incorrect results.

gammapy/modeling/parameter.py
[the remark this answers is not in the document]
QRemy: yes I added a call to this method in from_stack()

gammapy/modeling/parameter.py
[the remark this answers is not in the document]
QRemy: ok removed

gammapy/modeling/parameter.py
```diff
@@ -306,15 +306,19 @@ def scipy_mvn(self):
...
         """
         pars = itertools.chain(*parameters_list)
-
-        # TODO: Fix covariance stacking!
-        # covariances = [Parameters(_)._any_covariance for _ in parameters_list]
-        # covariance = scipy.linalg.block_diag(*covariances)
-        covariance = None
-        return cls(pars, covariance)
+        parameters = cls(pars)
+        npars = len(parameters)
```
adonath: I noticed that there is currently an inconsistency: here the covariance is set to `len(parameters)`, while in the function below, the indexing into the covariance matrix relies on `unique_parameters`. What is the convention currently used? Is the covariance adapted to the full or the unique parameter list. Can you please comment @cdeil?

cdeil: Always use the full parameters. Creating the subset of unique parameters should be an explicit operation, only used in the optimiser interface, never in internal methods.

cdeil:
@QRemy - Thanks for working on this!

If you remove the "TODO" for the stacking, then you should also remove the "FIXME" here and make sure stacking works properly:
https://github.com/gammapy/gammapy/blob/955a692a6ee78ddf766b06ab4c61a39729521c60/gammapy/modeling/tests/test_parameter.py#L130
I'm not sure why this wasn't working before, maybe it was because of the call to unique in `__init__` that we had before, which I removed, so now it might just work.

Otherwise: the code you have (duplicated 2x) to make the mask to filter the sub-covariance matrix with the for loop and enumerate isn't nice. I suggest to instead add this:
```
@property
def parameters_array(self):
    return np.array(self._parameters)
```
and from then on only use Numpy array expressions to do the sub-set?
That's a temp solution, I think we should change `_parameters` from Python list to Numpy array internally as data member. If you want, you can try that directly here, but my guess is that it will require code changes in ~ 5 other places and might be done better as a separate PR.

As in the inline discussion above: the processing should be done on all parameters, don't call `unique_parameters`. Suggest you change your test case to have duplicate parameters, and fail, and only then to fix the implementation.

---

I haven't looked at this in detail. My guess is that doing lots of subclass-specific processing as in `_model_list` in the base class isn't a great solution, and also that the covar code can be improved. But I'd still suggest you merge this in tomorrow, however far you get, and then we start fresh on Monday.

gammapy/modeling/models/cube.py
```diff
@@ -57,11 +57,8 @@ def __init__(self, skymodels):
 
     @property
     def parameters(self):
-        parameters = []
-        for skymodel in self.skymodels:
-            for p in skymodel.parameters:
-                parameters.append(p)
-        return Parameters(parameters)
+        parameters = Parameters.from_stack([_.parameters for _ in self.skymodels])
```
cdeil: Just `return`, remove temp local variable `parameters` and save one line.

gammapy/modeling/models/spatial.py
```diff
@@ -77,7 +77,7 @@ def position_error(self):
                 angle=np.nan * u.deg,
             )
         pars = self.parameters
-        idx = [pars._get_idx("lon_0"), pars._get_idx("lat_0")]
+        idx = [pars._get_idx(name) for name in ["lon_0", "lat_0"]]
```
cdeil: Shouldn't the idea here be that you don't need to work with `idx` externally at all?
It should be possible to call a method to get the sub-covar directly?
Doesn't this work already or should work?
```
sub_covar = pars.get_covariance(["lon_0", "lat_0"])
```

gammapy/modeling/tests/test_parameter.py
```diff
@@ -186,3 +184,16 @@ def test_parameters_autoscale():
     pars.autoscale()
     assert_allclose(pars[0].factor, 2)
     assert_allclose(pars[0].scale, 10)
+
+
+def test_subcovar():
+    pars_0 = Parameters([Parameter("a", 10), Parameter("b", 20), Parameter("c", 30)])
+    pars_1 = Parameters(list(pars_0.unique_parameters)[:2])
```
cdeil: What are you doing here?
`pars_0` doesn't have the same `Parameter` object twice.
Why are you calling `unique_parameters` on it?
Can you make the test case clearer?

gammapy/modeling/tests/test_parameter.py
```diff
@@ -186,3 +184,16 @@ def test_parameters_autoscale():
...
+    pars_0.covariance = np.array([[3, 4, 5], [7, 8, 9], [10, 11, 12]])
+
+    pars_1.set_subcovariance(pars_0)
+    assert np.all(pars_1.covariance == np.array([[3, 4], [7, 8]]))
```
cdeil: Never call `assert np.all(a==b)`
It will not print the two arrays if there's a different in the test report.
Always call `assert_allclose` or `assert_equal` from `numpy.testing` instead.

gammapy/modeling/models/spatial.py
QRemy: If you don't want pars._get_idx("lon_0") here, we need a method to get sub-covariance by nam. Axel said it's not safe because of parameters duplication so I removed it.

adonath: Maybe that was a misunderstanding, my [comment](https://github.com/gammapy/gammapy/pull/2522#discussion_r343736895) was that `parameters.get_subcovariance()` should not document the possibility to call it with names, as it is un-safe. Internally (as you know, that there are no parameter duplications) it still work and you can use it...

gammapy/modeling/parameter.py
adonath: I'm just understanding the full consequences of the change to keep the full parameter list internally and I'm not very happy with it. Imaging a joint-fit of ~20 datasets and a model with ~20 parameter. So you get a 400x400 covariance matrix of which 95% of the information is duplicated? This does not makes sense to me. You also need to extend the API to things like `.unique_covariance` and `.set_covariance_factors()` to also extend the matrix for duplicated parameters. Currently the covariance is also not handled correctly in e.g. `_expand_factor_matrix()`, which still relies on the parameter list to be unique. It silently works now, because `Datasets.parameters` returns the unique parameter list. But the problem also appears here https://github.com/gammapy/gammapy/pull/2522/files#diff-c17632aab77905a57af61541e9dfb52bR118, where `.set_subcovariance()` relies on the unique parameter list to be passed in. My strong recommendation would be to undo the change to the full parameter list and filter out duplicated parameters in `.from_stack()` instead.

gammapy/modeling/fit.py
```diff
@@ -113,6 +113,10 @@ def run(self, optimize_opts=None, covariance_opts=None):
         # back or how to form the FitResult object.
         optimize_result._success = optimize_result.success and covariance_result.success
 
+        # set sub-covariance matrix for each model
+        for model in self.datasets._model_list:
```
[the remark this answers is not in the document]
QRemy: added a TODO about reworking the _model_list

cdeil:
@QRemy - I talked with @adonath - he will finish up this PR and merge it in.

The scope was too big: the covar copy part doesn't work as is for the case of models on multiple datasets, and needs more thought and work on the optimiser interface.

Let's discuss and continue early next week ...

PR #2524, opened 2019-11-08
Add maxsize limit to lru_cache()

adonath:
This PR defines the `maxsize` parameter in the `lru_cache()` we use in `WcsGeom` to fix a memory leak reported by Aion Viana in private and confirmed by the first benchmarks of e.g. [maps_3d](https://github.com/gammapy/gammapy-benchmarks/blob/master/benchmarks/results/maps_3d/0.15.dev373%2Bg74763d09f/2019-10-31/results.png).

The memory leak was due to a misunderstanding from my side of the `maxsize` parameter and the `lru_cache()` in general. The `lru_cache()` acts on the class level and not the instance level, this means the `maxsize` parameter refers to the last recent calls on the class level (i.e. across all instances). This also means that even if an instance of `WcsGeom` goes out of scope, the cache is still kept if `maxsize=None`. I now changed `maxsize` to the typical number of `WcsGeom` objects involved in the data reduction, so the cache is cleared after 5 independent calls. I already re-run the benchmarks and now the results look as expected:
- [maps_3d](https://github.com/gammapy/gammapy-benchmarks/blob/master/benchmarks/results/maps_3d/0.15.dev491%2Bg955a692a6.d20191108/2019-11-08/results.png)
- [spectrum_1d](https://github.com/gammapy/gammapy-benchmarks/blob/master/benchmarks/results/spectrum_1d/0.15.dev491%2Bg955a692a6.d20191108/2019-11-08/results.png)

In general it's not ideal that `lru_cache()` acts on the class level, because you have to limit the `maxsize` to prevent memory leaks, and this also means, that the caching behaviour depends on the number of instances of `WcsGeom` and previous calls to cached methods the user has done in the same process. On the other hand right now the `lru_cache` is not critical for performance of the data reduction and could be also removed. Pragmatically I would say as a users it's still better to get the faster performance of `.get_coord()` in most cases then never.

### DIFF
--- a/gammapy/cube/make.py
+++ b/gammapy/cube/make.py
@@ -79,6 +79,9 @@ class MapDatasetMaker:
         self.cutout_width = 2 * self.offset_max
         self.cutout = cutout
 
+        # define cached methods
+        self.make_exposure_irf = lru_cache(maxsize=1)(self.make_exposure_irf)
+
     def _cutout_geom(self, geom, observation):
         if self.cutout:
             return geom.cutout(
@@ -164,7 +167,6 @@ class MapDatasetMaker:
             geom=geom,
         )
 
-    @lru_cache(maxsize=1)
     def make_exposure_irf(self, observation):
         """Make exposure map with irf geometry.
 
@@ -274,7 +276,6 @@ class MapDatasetMaker:
             exposure_map=exposure,
         )
 
-    @lru_cache(maxsize=1)
     def make_mask_safe(self, observation):
         """Make offset mask.
 
@@ -293,7 +294,6 @@ class MapDatasetMaker:
         data = offset >= self.offset_max
         return Map.from_geom(geom, data=data)
 
-    @lru_cache(maxsize=1)
     def make_mask_safe_irf(self, observation):
         """Make offset mask with irf geometry.
 
--- a/gammapy/maps/tests/test_wcs.py
+++ b/gammapy/maps/tests/test_wcs.py
@@ -251,9 +251,26 @@ def test_wcsgeom_get_coord():
     assert_allclose(coord.lat[0, 0].value, -1.5)
     assert coord.lat[0, 0].unit == "deg"
 
-    coord_cached = geom.get_coord(mode="edges")
-    # test coord caching
-    assert id(coord) == id(coord_cached)
+
+def test_wcsgeom_instance_cache():
+    geom_1 = WcsGeom.create(npix=(3, 3))
+    geom_2 = WcsGeom.create(npix=(3, 3))
+
+    coord_1, coord_2 = geom_1.get_coord(), geom_2.get_coord()
+
+    assert geom_1.get_coord.cache_info().misses == 1
+    assert geom_2.get_coord.cache_info().misses == 1
+
+    coord_1_cached, coord_2_cached = geom_1.get_coord(), geom_2.get_coord()
+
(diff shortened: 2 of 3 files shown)


cdeil:
I didn't know about this behaviour of `lru_cache` either, I handn't used it on methods, only functions.

A 5-fold caching, or any other number, at the class level is really bad. The only reason to cache is to improve performance. If we don't need it, delete it. Introducing something that sometimes works on small test cases, but then gives bad performance suddenly when having 6 instead of 5 datasets is horrible.

I think overall we need caching, mostly in the likelihood eveluation setup phase to have fast fitting, but maybe also in data reduction. I think e.g. solid angle is already used in map evaluation and likelihood fitting?

Probably the best solution here is to remove the use of `lru_cache` and solely to use `astropy.utils.lazyproperty`, which we already do in Gammapy:
https://github.com/gammapy/gammapy/search?q=lazyproperty&unscoped_q=lazyproperty
That means to only cache properties, not methods with arguments.
I don't think there is a real need to have methods cached, at least for now?
In the future, we might want to have it depending on parameter value changes.
You might have to add a few properties which are the common case we currently use, that doesn't requires arguments.

Alternatively we could implement a per-instance cached method. But it's more complex:
https://stackoverflow.com/a/33672499/498873

adonath:
@cdeil It seems there is a surprisingly simple solution https://stackoverflow.com/a/14946506.

I can confirm that it works:

```python3
from functools import lru_cache

class TestInstanceCache:   
    def __init__(self, x):
        self.x = x
        self.add = lru_cache()(self.add)

    def add(self, y):
        return self.x + y


class TestClassCache:
    def __init__(self, x):
        self.x = x
    
    @lru_cache()
    def add(self, y):
         return self.x + y
        
t_cls_1 = TestClassCache(x=1)
t_cls_1.add(3)
t_cls_1.add(3)
t_cls_1.add(3)

t_cls_2 = TestClassCache(x=2)
t_cls_2.add(3)
t_cls_2.add(3)
t_cls_2.add(3)

print(t_cls_1.add.cache_info())
print(t_cls_2.add.cache_info())

t_inst_1 = TestInstanceCache(x=1)
t_inst_1.add(3)
t_inst_1.add(3)
t_inst_1.add(3)


t_inst_2 = TestInstanceCache(x=2)
t_inst_2.add(3)
t_inst_2.add(3)
t_inst_2.add(3)

print(t_inst_1.add.cache_info())
print(t_inst_2.add.cache_info())
```

Which prints:
```
CacheInfo(hits=4, misses=2, maxsize=128, currsize=2)
CacheInfo(hits=4, misses=2, maxsize=128, currsize=2)

CacheInfo(hits=2, misses=1, maxsize=128, currsize=1)
CacheInfo(hits=2, misses=1, maxsize=128, currsize=1)
```
In the first case creates a global cache in the class, in the second case it creates the cache per instance, which is exactly what we want.

cdeil:
@adonath - Cool. Then let's do that, no?

(I didn't think about what happens and if there could be any gotchas with this bound method replacement with a cached version in `__init__`.)

Extra points if there's a unit test proving that it works (not obvious, could work in some Python versions but not others), i.e. create ~ 2 instances, call something, check return object IDs or that cache info.
(But OK to just apply the fix if you don't want to write a test.)

adonath:
@cdeil I agree we should do this. I can also add a small test...

PR #2526, opened 2019-11-09
Remove SNRCat

### LINKED ISSUE #2162: Change SNRCat to gammapy-data
`test_load_catalog_snrcat` is currently failing because it fetches from http://www.physics.umanitoba.ca/snr/SNRcat/ which is currently down.

https://github.com/gammapy/gammapy/blob/6c1c8f48dde5cc47394e21b2985b9f297689f973/gammapy/catalog/tests/test_snrcat.py#L6-L7

There's a TODO in `SNRCat` already to consider putting a copy in `gammapy-extra`:

https://github.com/gammapy/gammapy/blob/6c1c8f48dde5cc47394e21b2985b9f297689f973/gammapy/catalog/snrcat.py#L43

For the second "live" catalog we have that can change any time, gamma-cat`, we currently have a copy in gammapy-extra and access it via gammapy-data:

https://github.com/gammapy/gammapy/blob/6c1c8f48dde5cc47394e21b2985b9f297689f973/gammapy/catalog/gammacat.py#L456

My suggestion would be to also put a fixed stable version of SNRCat in gammapy-extra, to include that in gammapy-data, and to access that from the test.

We could also add a `SNRCat.from_url` to allow both load from gammapy-data or from URL, and in that case would have to decide whether to leave that method untested, or whether to mark the test as "allowed to fail" to not disturb devs and CI when the web service goes down.

@Bultako @adonath - Thoughts?

cdeil:
This PR removes http://snrcat.physics.umanitoba.ca/ from `gammapy.catalog.

The code in Gammapy didn't do much, mainly cleanup to be able to get the catalog as a nice `astropy.table.Table`.

I think most Gammapy people don't consider this in scope for Gammapy, and I now kind of agree, especially since we already removed (years ago) the ATNF pulsar catalog, and didn't add other catalogs that are relevant for gamma-ray astronomers (e.g. AGN).

Fixes #2162


cdeil:
@adonath - How does this get into the dataset index?
```
$ ack snrcat
dev/datasets/gammapy-data-index.json
127:    "path": "catalogs/make_snrcat.py",
128:    "url": "https://github.com/gammapy/gammapy-extra/raw/master/datasets/catalogs/make_snrcat.py",
175:    "path": "catalogs/snrcat.fits.gz",
176:    "url": "https://github.com/gammapy/gammapy-extra/raw/master/datasets/catalogs/snrcat.fits.gz",
193:    "path": "catalogs/green_snrcat_check.ipynb",
194:    "url": "https://github.com/gammapy/gammapy-extra/raw/master/datasets/catalogs/green_snrcat_check.ipynb",
```
Can you please remove it?

PR #2528, opened 2019-11-11
Move Analysis to new gammapy.analysis

Bultako:
This PR renames `scripts` module into `analysis` everywhere in the gammapy code base, checks for regression tests and documentation building have been made.


cdeil:
@adonath @registerrier - Do you think we should make this name change or not?

I'm 50:50, OK either way.

adonath:
@cdeil @Bultako I remember, that we talked about this during one of the dev meetings, but I think we did not take a decision. 

I slightly prefer `gammapy.analysis` as a namespace for the `Analysis` class, but what I don't like is that the cli scripts are in the same sub-package. So I would even propose to introduce a separate `gammapy.analysis` sub-package, copy over the `Analysis` and `AnalysisConfig` class  and keep the `cli` in `gammapy.scripts`.

Bultako:
Yes, it's true no decission has been taken on renaming the package so far, though it has been evocated several times. But I think it's better to take a decission while it's still not very hard to do it (not much code in the high-level interface), users are still not used to work with it, and the big work on documentation has not yet started. With this PR I kind of force to take the decission now, sorry :)

> I slightly prefer gammapy.analysis as a namespace for the Analysis class, but what I don't like is that the cli scripts are in the same sub-package. So I would even propose to introduce a separate gammapy.analysis sub-package, copy over the Analysis and AnalysisConfig class and keep the cli in gammapy.scripts.

I'm +1 to do that.

registerrier:
> > I slightly prefer gammapy.analysis as a namespace for the Analysis class, but what I don't like is that the cli scripts are in the same sub-package. So I would even propose to introduce a separate gammapy.analysis sub-package, copy over the Analysis and AnalysisConfig class and keep the cli in gammapy.scripts.
> 
> I'm +1 to do that.

I am +1 too.

cdeil:
I'm also OK to just create `gammapy.analysis`, and to keep `gammapy.scripts` for the scripts.
Please either do that from scratch in a new PR, or if you do it here, please squash commits to avoid moving things back and forth and losing version history connect.

> With this PR I kind of force to take the decission now, sorry :)

That's perfectly fine. I think usually an issue with a suggestion and waiting for comments for a day is more effective, but if the work to make a PR isn't very high, just brining the PR is a good way, as long as you're willing to take the risk to have it rejected or having to change it significantly like in this case.

Bultako:
>  or if you do it here, please squash commits to avoid moving things back and forth and losing version history connect.

ok, it's done.

docs/analysis/index.rst
```diff
@@ -0,0 +1,266 @@
+.. include:: ../references.txt
+
+.. _HLI:
```
cdeil: `HLI` -> `analysis` (and same for links below), to follow our label style in Gammapy.

Issue #2517, opened 2019-11-05
Consistent value for name attribute in datasets


Bultako:
We have a `name` attribute for 1D and 3D datasets. The values for these attributes are set in:

https://github.com/gammapy/gammapy/blob/fe6f23d8b7e3fc1d2f7f00a2dbe368769bea0d0d/gammapy/cube/make.py#L355

https://github.com/gammapy/gammapy/blob/fe6f23d8b7e3fc1d2f7f00a2dbe368769bea0d0d/gammapy/spectrum/make.py#L186

Is it intentional e.g. for fitting maps and spectra together? then maybe the naming should be a bit clearer...

adonath:
> Is it intentional e.g. for fitting maps and spectra together?

No, I think this just happened by accident. Are there any preferences for the default name we should choose? I guess we could also just just enumerate the datasets and put e.g. `"map-dataset-1"` or `"spectrum-dataset-1"`. I think in general it's probably not a good idea to keep track of meta information such as observation id in the name, so maybe we shouldn't even try at all to do it in the name and introduce `.meta` later...

adonath:
Note: a simple enumeration of the datasets (instead of the obs-id) also avoids this issue: https://github.com/gammapy/gammapy/issues/2515 on the dataset level.

cdeil:
The question of names for models and datasets has come up many times in the past year. It's a difficult one, but it is solvable once and for all by making a few decisions and writing a few lines of documentation, so we should do that to avoid going in circles.

My 2 cents: generating good names for models and datasets is hard, because of different use cases. For the case of run-wise datasets, using the OBS_ID to make a name is the best solution. If you make a lightcurve and either split one run into many datasets, or combine different runs into one stacked dataset, then you're in trouble with the OBS_ID.

Possible solutions here are:
- Delete `name` attribute from dataset and model objects, because it's not needed at all, it's just "annotation" which is nice to have in some cases, but causing problems in others.
- If you keep the names, make a decision and state it in the docs whether Gammapy internally requires names to be unique. I'm strongly -1 to require unique names, because as far as i can see it's just not needed. (e.g. for compound models I removed the re-exposition of the sub-model parameters on the compound model, to avoid the issue of name clashes if e.g. multiple `norm` parameters are around).
[...]

PR #2529, opened 2019-11-11
Improve SkyModels and Datasets

cdeil:
This PR improves the `SkyModels` and `Datasets` containers by making their internal list representation private and only offering a minimal API. This is required to achieve internal consistency between the models/datasets and parameter/covariance attributes, and possibly name uniqueness checks (see #2517). If the Python list `sky_models.sky_models` is accessed from outside, users can do whatever they like, and we have no chance of keeping a `sky_models.parameters` in sync.

I've removed the `names` check in `SkyModels.__init__` for now. We can add it back with tests (there wasn't any), and a better implementation that prints the object IDs and name of a duplicate entry. We could do it in `__init__`, or on `models.get("crab")`, i.e. only if someone used name-based access. Should be done in a consistent way for models and datasets.

The scope here is limited to cleanup - I'd prefer to merge this in soon, and then to send other PRs to implement #2517 and polish these classes.

### DIFF
--- a/gammapy/analysis/core.py
+++ b/gammapy/analysis/core.py
@@ -174,11 +174,9 @@ class Analysis:
             if isinstance(dataset, MapDataset):
                 dataset.model = self.model
             else:
-                if len(self.model.skymodels) > 1:
-                    raise ValueError(
-                        "Can only fit a single spectral model at one time."
-                    )
-                dataset.model = self.model.skymodels[0].spectral_model
+                if len(self.model) > 1:
+                    raise ValueError("Cannot fit multiple spectral models")
+                dataset.model = self.model[0].spectral_model
         log.info(self.model)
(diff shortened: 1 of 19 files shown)


cdeil:
@adonath - Please review.

The diff is already quite large so I'd prefer to leave the change to move `Parameters` object creation to `__init__` and to add consistency checks between models and parameters to a follow-up PR. That will likely be more tricky, and benefit from adding unit tests, whereas this is mainly a cleanup of `Datasets` and `SkyModels` callers.

cdeil:
Access by name is still there (in `__getitem__`). Concerning uniqueness checks: #2517 is still open, we can address it in the next days.

cdeil:
Follow-up commit in master: 35c46155d

Issue #2102, opened 2019-03-29
Single object vs container confusion


cdeil:
I find it confusing if one is allowed to pass single objects or containers.

Several such cases have been introduced recently: see cases below where one can pass a single model or a list, or a background model or a list, or a dataset or list of datasets.

Suggest to simplify those and to only accept lists.

The coding pattern to use is usually:
```
def __init__(self, things):
    self.things = Things(things)
```
and to make it behave like a no-op if `things` is already a `Things` object, and store away the list if a list is passed.

The one small drawback is that sometimes one has to write `Fit(datasets=[my_dataset])`, i.e. an extra `[]` if one only has a single object where the code can handle several. IMO the convenience of being able to write `Fit(datasets=my_dataset)` isn't worth the cost:

- Requires more code (isinstance checks)
- Requires more tests (two cases)
- Requires more complex docstrings
- Two ways to do the same thing - confusion by users and devs where which is accepted and needs to be handled.

---

A bit related is the issue that `SkyModel` is not a base class, but the concrete factorised sky model where spectrum and spatial part are factorised. I think the `gammapy.utils.fitting` code would become clearer if we rename and make `SkyModel` the base class (just how `Map` and `Dataset` are the base classes). Also from just reading around today I found the code with source & background models quite complex - I wonder if a single models list and flags that declare what to do with the model wouldn't be simpler.

@adonath - up to you if you want to finish the datasets and models PIGs first and have everything clear, or just go one pull request / discussion at a time in the coming weeks.

---

Concretely, I saw these places in the current code that I would suggest we simplify:

https://github.com/gammapy/gammapy/blob/b2e78e72521929513a26c59bd1bc343282133b7b/gammapy/cube/fit.py#L24-L25

https://github.com/gammapy/gammapy/blob/e699c3658f6aacff52873508706a962f275021b6/gammapy/cube/fit.py#L36-L37

https://github.com/gammapy/gammapy/blob/b2e78e72521929513a26c59bd1bc343282133b7b/gammapy/cube/fit.py#L80-L98

https://github.com/gammapy/gammapy/blob/b2e78e72521929513a26c59bd1bc343282133b7b/gammapy/utils/fitting/datasets.py#L37-L51

adonath:
I'm completely fine with this suggestion. I think I can take an afternoon before the 0.12 release and implement this pattern everywhere in Gammapy. Alternatively we leave it up to the big API change before the 1.0 release...

PR #2530, opened 2019-11-12
Fix Datasets vs Datasets confusion

cdeil:
This PR is one step to address #2102, it fixes some confusion of `Datasets` vs `Dataset`.

The general idea is to be more strict, to simplify code within Gammapy, but also to make it easier for users to learn the one right way to do it, and not rely on `Datasets` <-> `Dataset` auto-conversions for the common case of one dataset. I think this is clearer, and it will even be required to move `Parameters` to `Datasets.__init__` and to let users access the results from `Datasets`. (Same arguments for `SkyModels`. Being more strinct in `Datasets.__init__` caught one usage bug [here](https://nbviewer.jupyter.org/github/gammapy/gammapy/blob/master/tutorials/joint_1d_3d_analysis.ipynb#HESS-DL3:-1D-ON/OFF-dataset-for-spectral-fitting) where accidentally a single `Dataset` instead of a list of datasets was created.

There's still plenty of more cleanup needed, especially many datasets have a `model` attribute which really is a `models` object. That's for a follow-up PR. One change I'm doing here already is to remove `SkyModels.evaluate`, with the idea that `SkyModels` is the dumb container that does book-keeping and I/O, but evaluation is only at the `SkyModel` level, and typically done on cutouts, not how the current `SkyModels.evaluate` was written. Looks like it wasn't used.

@adonath - Please review.

### DIFF
--- a/gammapy/cube/tests/test_fit.py
+++ b/gammapy/cube/tests/test_fit.py
@@ -375,7 +375,7 @@ def test_map_fit_one_energy_bin(sky_model, geom_image):
     sky_model.parameters["sigma"].value = 0.21
     dataset.background_model.parameters["norm"].frozen = True
 
-    fit = Fit(dataset)
+    fit = Fit([dataset])
     result = fit.run()
 
     assert result.success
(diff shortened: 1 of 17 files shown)


cdeil:
I want to continue, so I'm merging this in now.
Comments still welcome though ... this cleanup is work in progress.

Issue #2181, opened 2019-05-28
Improve PSF3DChecker


cdeil:
The `PSF3DChecker` should be improved.

* https://github.com/gammapy/gammapy/blob/master/gammapy/irf/psf_check.py
* https://github.com/gammapy/gammapy/blob/master/gammapy/irf/tests/test_psf_check.py

It's currently the slowest automated test in Gammapy, taking 6.4 seconds on my machine (see [here](https://gist.github.com/cdeil/16f259226bf5e35c36d504f187b48a26)).

It looks like the check methods are implemented as Python for loops, where most or all of them could be written with a few lines of Numpy array expressions instead. That should drastically reduce the code size and runtime.

There's also this code which basically re-implements a PSF method to compute an integral in Gammapy in the checker - that should be removed or rewritten:
https://github.com/gammapy/gammapy/blob/2fab5df0ed27491cf8c3eb4836db98da41aeb794/gammapy/irf/psf_check.py#L200-L203

If this is not possible for some check, we could change to a simpler test - the main point of the automated tests is to make it easy to check for obviously bad IRFs for batch processing (i.e. the people producing IRFs in CTA, HESS, .... or as a quick check when consuming or doing data quality for new FITS data).

I have a slight preference for improving the checker code in Gammapy, as opposed to just deleting it again, as we make a push for code quality and v1.0 in Gammapy. But removing is also an option, with the assumption that such checker code would re-appear in ctapipe or Gammapy in future years.

This code was added in #763 and #767 

cc @lmohrmann or @dltiziani

cdeil:
See #2532

adonath:
The code has been deleted in #2532. So there is nothing to improve anymore...closing the issue now.

PR #2532, opened 2019-11-12
Remove PSF3DChecker

cdeil:
This PR removes the `PSF3DChecker`, a class that scripts some PSF3D IRF checks.

The problem with this code, and a few 1000 other lines of code in Gammapy, is:
a) it's only needed by very few people, not clear if anyone is using it, or if it's in scope for Gammapy
b) the code quality isn't good and no-one has time to improve it (see #2181)

I think the right thing to do is to remove it from Gammapy, and to close many old issues (bring it down to < 100 open issues at least), with a comment that such functionality could be re-added to Gammapy in the future, but we'll only allow it in and maintain it if the quality is good. More likely, you'll spend another 1-2 years to improve the core functionality of Gammapy, and then it will remain to be seen where IRF code for CTA is developed and maintained. In any case, IMO a small and polished Gammapy will be better, and the code will forever be in existing releases and the git version history, it's not lost.

@lmohrmann - Are you importing and running this as part of HESS DL3 production?
@registerrier - Do you use it in Paris?

@adonath and @registerrier as future Gammapy lead maintainers - do you want to remove or keep this?


adonath:
I agree 100% that we should aim for a small and cleaned-up Gammapy code base and not keep (or maintain...) old code, that is not necessarily in focus of Gammapy. 

`gammapy/irf/tests/test_psf_check.py` is also one of our longest running tests, so in general I'd prefer to remove the `PSF3DChecker` as well, but let's wait for @lmohrmann and @registerrier to comment. If it's code that is regularly used for IRF production, we should find a compromise.

cdeil:
Cool. I'll continue with cleanup and more PRs then.

Somewhat related to the IRF checker is to have FITS prod checks. In 7ce9ebc24 I've removed `gammapy/irf/tests/test_fits_prod.py`. There we previously had several HESS FITS prods (HD HAP, HAP-FR, ParisAnalysis) and made sure that Gammapy could read the IRFs from those. That was many years ago, before there was a spec for formats. Having them isn't as useful any more, IRF and DL3 tests should rather be structured by spec version, not HESS FITS production.

PR #2533, opened 2019-11-13
Improve Parameters.set_error

cdeil:
This PR improves `Parameters.set_error`, to work like `Parameters.__init__` via keyword arguments.
See the docstring and example.

This replaces the previous `set_error` and `set_parameter_errors` methods and unifies them into one method (avoiding the code duplication, and resolving the TODO comments that this needs cleanup).

### DIFF
--- a/gammapy/catalog/fermi.py
+++ b/gammapy/catalog/fermi.py
@@ -206,7 +206,7 @@ class SourceCatalogObjectFermiBase(SourceCatalogObject):
         scale_1sigma = Gauss2DPDF().containment_radius(percent)
         lat_err = semi_major.to("deg") / scale_1sigma
         lon_err = semi_minor.to("deg") / scale_1sigma / np.cos(d["DEJ2000"].to("rad"))
-        model.parameters.set_parameter_errors(dict(lon_0=lon_err, lat_0=lat_err))
+        model.parameters.set_error(lon_0=lon_err, lat_0=lat_err)
         model.phi_0 = phi_0
 
     def sky_model(self):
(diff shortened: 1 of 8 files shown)


cdeil:
There's issues with the Fermi spatial models related to `_set_spatial_errors`:
https://gist.github.com/cdeil/807833ea46924a5a2b12c6fdb91b51f4

Setting position errors now raises an error for the template models that don't have position parameters.

It's this code:

https://github.com/gammapy/gammapy/blob/b6f38f5f44abed338c8062a7769259085ed86aa2/gammapy/catalog/fermi.py#L189-L210

and callers for each catalog:

https://github.com/gammapy/gammapy/blob/b6f38f5f44abed338c8062a7769259085ed86aa2/gammapy/catalog/fermi.py#L378-L412

I think overall this code has to be restructured. It's buggy, e.g. `phi_0` is set as the position error position angle, but it is for disk and Gaussian the orientation of the spatial model. Also only some errors are filled, namely the ones for position, but not other parameters like source extensions.

I'll `xfail` those two tests now and merge this in now, leaving the larger work to fix `fermi.py` spatial models to a follow-up PR.

cdeil:
Hmm, so the kwargs API doesn't support using parameter objects as keys like here: d4c7c9224c0e03023d7f2281ed20c55ab8872862

For now it's not needed, and it's not clear to me if this is the API we want to offer to users to set initial step sizes for fitting. If yes, we might want to switch back to a single dict API or two methods.
(but the code duplication / behaviour should be uniform for sure, and for me, offering the same API to set values and errors is a big plus)

Issue #2403, opened 2019-09-26
Possible improper usage of `compute_lima_image`


registerrier:
With the current implementation `compute_lima_image` takes as argument any possible `Kernel2D`. This can yield incorrect results. One can apply Li & Ma only by summing counts, so that applying convolution is valid only if the kernel is made of `1`and `0`. 
The only valid kernels should therefore be `Tophat2DKernel`, `Ring2DKernel`  and possibly `Box2DKernel`.

I am not sure there is a strong use case for other kernels than `Tophat2DKernel`. So that the only parameters should be `compute_lima_image(n_on, n_off, a_on, a_off, correlation_radius="0.1 deg")`.
Besides avoiding incorrect results if someone was to use a `Gaussian2DKernel`, this allows to avoid requiring to define a kernel in pixel coordinates.

There is one class that uses `compute_lima_image`, with any possible kernel, the `KernelBackgroundEstimator`.  The latter can still function with only a tophat kernel, so I propose to adapt it as well to use only a correlation radius. If not the significance it computes is incorrect.
 
Opinion @adonath ?

cdeil:
I think it's also possible to have a PSF or PSF-convolved source template "matched filter" weights kernel and to use that, and to get a very fast and close approximation to what likelihood amplitude fitting / Li&Ma gives?

- https://www.aanda.org/articles/aa/full/2006/30/aa3498-05/aa3498-05.html
- https://www.aanda.org/articles/aa/full_html/2009/09/aa11311-08/aa11311-08.html

I see that currently we call `kernel.normalize("peak")` (see [here](https://docs.gammapy.org/0.13/_modules/gammapy/detect/lima.html#compute_lima_image)), which normalises to a peak value of 1 (see [here](http://docs.astropy.org/en/stable/api/astropy.convolution.Kernel.html#astropy.convolution.Kernel.normalize)). That's probably not useful for any application, we should remove that and let the caller decide the kernel normalisation?

My guess is that changing Li & Ma to only a "correlation radius" argument might be going in the wrong direction, and instead we should aim to create a more consistent API with TSMapEstimator and also keep the option to have any kernel shape for the "matched filter" method in Steward?

(I'm not sure, other options are to do what you propose, or to just delete those two functions and show an example, since the building blocks are map correlate and `significance` after correlation, i.e. both present in Gammapy)

registerrier:
Well I think that the function should be doing what it says it is doing. You would not apply Li & Ma with any kernel.
[...]

Issue #752, opened 2016-11-02
Add ASmooth 2FHL example


adonath:
Add an example notebook, that runs the `ASmooth` algorithm on the Fermi 2FHL dataset. Or even reproduce Fig. 1 and Fig 5. from the [2FHL paper](https://arxiv.org/abs/1508.04449).

QRemy:
Added to detect notebook in #2681

Issue #751, opened 2016-11-02
Add background estimation to ASmooth


adonath:
`ASmooth` uses an internal background estimation (ring of 3-4 sigma for Gaussian, and 1-4/3 sigma for Tophat), which should be added as an option to the `ASmooth` class.

registerrier:
Do we expect to do something here, or should we close the issue @adonath ?

PR #2535, opened 2019-11-13
Remove gammapy.detect.CWT

cdeil:
This PR removes `gammapy.detect.CWT`.

It will forever be available in Gammapy v0.14, here, if someone wants to use this, or improve the code and re-add it to Gammapy in the future:
- Code (794 lines, 672 sloc): https://github.com/gammapy/gammapy/blob/v0.14/gammapy/detect/cwt.py
- Tests (258 lines, 212 sloc): https://github.com/gammapy/gammapy/blob/v0.14/gammapy/detect/tests/test_cwt.py
- Tutorial: https://docs.gammapy.org/0.14/notebooks/cwt.html

IMO this functionality is in scope for Gammapy, but the quality of the code, tests and docs is too low to keep it and to offer it to users, and we should instead focus our efforts on Li & MA and TS code, tests and docs (see e.g. #2403).

I could enumerate the issues and work that should be done, but maybe this removal isn't controversial, so I'll not make the case for now. I think it would be a week of full-time work for an expert like @adonath or @registerrier to get this into shape, and to make sure the results are solid and the documentation is good. This was added 5+ years ago without tests, and then we had a GSoC student add tests and write the tutorial, but we didn't review the results at the time, and didn't understand the results we got for different parameters, I suspect it's not giving good results. This is also something that extremely few Gammapy users need, so it's not high priority for us, which explains why we haven't worked on this in the past years, even though it was clear that it needs work.

Of course there are many other things in Gammapy that need work, and in each case we have to make a decision to invest time to improve, or keep and improve later, or remove for now, with a comment that PRs in the future are welcome, with good quality implementation, tests and docs. In this case I think removing is best, if I were a potential contributor to Gammapy with wavelet skills, I'd probably be interested to add something, but not to clean up these 1000 lines of code and tests.

@registerrier @adonath - What do you think?

If you want to keep this, please make a proposal what to do with `cwt.ipynb`, as part of our upcoming documentation effort.


cdeil:
We should also discuss [gammapy.detect.KernelBackgroundEstimator](https://docs.gammapy.org/0.14/api/gammapy.detect.KernelBackgroundEstimator.html), [gammapy.detect.ASmooth](https://docs.gammapy.org/0.14/api/gammapy.detect.ASmooth.html), the two ring background estimators, and the TS and Li&Ma code.

Those are in better shape than CWT, where it's 1000 lines of code, and it's most different from our overall API style (not using `Map` objects internally, two separate algorithm-specific CWD container classes, for maps and kernels). That's why I started with CWT. But IMO we should consider more removals, or improve and unify what we keep.

adonath:
@cdeil I agree we have to talk about the other classes such as `ASmooth`, `TSMapEstimator` etc., but there I'm clearly against removing those (not sure you actually proposed it...). I think all of theses can be adapted with minimal effort to feature a uniform API (if this is not the case already),  which probably means to adapt the API to use `MapDataset` and `MapDatasetOnOff` on input.

cdeil:
`KernelBackgroundEstimator` is similar to `CWT` in the sense that it's an iterative 2D image algorithm that produces a background model, exclusion mask and significance map. So we could remove it.

On the other hand, the one-step method of `KernelBackgroundEstimator` is a generalisation of ring background estimation, so ideally we would offer that general version and the ring methods without code duplication. Throwing away the general solution and only offering a ring kernel is weird. And even the iterative method is not uncommon - the iterated ring background estimation was the method used to make exclusion masks and background estimates for HGPS and other Galactic papers.

For `ASmooth` I have a similar concern as for `CWT` - it was never validated. A quick way to validate it could be to try to reproduce the 3FHL map - email the corresponding author if the asmooth map from 3FHL that's in the paper as a Figure isn't publicly available as a FITS file.  But for ASmooth we can discuss in #751 and  #752 - @adonath if you want to improve those this year, I suggest you leave a comment with action items, if you want to just do nothing this year, move back to v1.1 milestone.

PR #2541, opened 2019-11-14
Add source detection to Analysis class

Bultako:
**Description**
This PR adds source detection features in the high-lelve interface.

For this a new section `detection` has been added in the parameter schema, with parameters `threshold` and `map`(possible values are `sqrt_ts`, `flux` and `niter`. A `template-2d.yaml` file has been added. The source detection is performed using `gammapy.detect.TSMapEstimator` class. 

The methods for these new features are:
-  `analysis.detect()` 
-  `analysis.plot_detections()`

There are three new properties that can be accessed with `analaysis` instance.
- `analysis.TSmaps`
- `analysis.detections`
- `analysis.detection_map`

Documentation has been updated and tests have been added.

**Example use**
```
>>> analysis.detect()
INFO:gammapy.analysis.core:Proceeding to source detection.

>>> print(analysis.detections)
value   x   y      ra       dec
                  deg       deg
------ --- --- --------- ---------
36.086 252 197 266.42400 -29.00490
16.726 206 204 266.83194 -28.14603

>>> analysis.plot_detections()
```
detections

### DIFF
--- a/docs/analysis/index.rst
+++ b/docs/analysis/index.rst
@@ -251,6 +251,30 @@ for the source and/or the background.
             unit  :
             dtype : float64
 
+Source detection
+----------------
+
+You can also run source detection algorithms in wide field stacked 2D images. For this you need to
+assign a value for the `threshold` parameter in the ``detection`` section and a map where to perform
+the detection (i.e. `sqrt_ts`, `flux` or `niter`.
+
+.. gp-howto-hli:: detection
+
+.. code-block:: python
+
+    >>> analysis.detect()
+    INFO:gammapy.analysis.core:Proceeding to source detection.
+    >>> print(analysis.detections)
+    value   x   y      ra       dec
+                      deg       deg
+    ------ --- --- --------- ---------
+    36.086 252 197 266.42400 -29.00490
+    16.726 206 204 266.83194 -28.14603
+    >>> analysis.plot_detections()
+
+.. gp-image:: hli/detections.png
+    :width: 100%
+
 Using the high-level interface
 ------------------------------
 
--- a/gammapy/analysis/config/docs.yaml
+++ b/gammapy/analysis/config/docs.yaml
@@ -104,3 +104,8 @@ flux-points:
         unit: TeV
         interp: log
         node_type: edges
+
+# Section: detection
+# Threshold for source detection on a TS map
+detection:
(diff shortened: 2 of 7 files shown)


cdeil:
I have to say: I'm -1 to keep adding such things to `Analysis`.

Is the goal of `Analysis` to expose all or most functionality from Gammapy via this single class and a single config file?

I think the extra value you're adding with this addition for users is minimal.

Why expose this functionality? Why not TS maps or Li & Ma maps or ASmooth maps? Or KernelBackgroundEstimator?
https://docs.gammapy.org/dev/detect/index.html

We didn't discuss about scope of Analysis much, and to what extent it supports modeling & fitting, but re-exposing `gammapy.detect` on it wasn't even mentioned:
https://docs.gammapy.org/dev/development/pigs/pig-012.html

I think if we want to polish `Analysis` in the coming weeks and deliver a good v1.0, we should improve the core features (e.g. make observation selection or background modeling and stacking options better, expose documentation for all available options better, not add more stuff to it. This is horribly verbose (https://docs.gammapy.org/dev/analysis/index.html#data-reduction-and-datasets), and there are inconsistencies: for some angles, Quantities are allowed, others must be floats hard-coded in degrees, ...

Exposing more functionality via Analysis is something we can discuss also, but I'd suggest to discuss first and make a decision before coding. @Bultako - You prepare a slide or two listing the remaining high-priority work for Analysis for Monday?
https://github.com/gammapy/gammapy-meetings/tree/master/coding-sprints/2019-11_Granada#monday
If you want feedback from me or the lead devs before, send around a draft or show it briefly in the Gammapy dev call tomorrow.

Bultako:
@cdeil 
I agree there are higher priorities for the high-level interface, all the ones you have exposed and more. And as I see it, except doing better documentation, all are intrinsically related with improving code outside the `analysis/core.py` script. It's true that source detection is not mentioned in the PIG, but I thought this was part of the 2D analysis process and moreover it was one of few the things I could do it fast and easy just by my own before the coding sprint. There's no problem if we decide not to add it :)

Yes, there's a slot for the high-level interface in the presentations of monday afternoon, where I will expose status, priorities and next steps.

cdeil:
Discussed in the Gammapy call just now. Decided to not add this for now.

Issue #2304, opened 2019-07-22
Improve MCMC sampling code


cdeil:
As mentioned in #2289, we should improve the MCMC sampling code:

- Add automated tests using a very simple example that runs in 1 second
- Try to use only parameter values, not factors in MCMC code -- establish results with a test that has parameters of roughly `1, 1e-10, 1e10`.

Once that is done, we can try to refactor the code in sampling.py to have a well-defined interface like what we have in the `Fit` class, and to have `emcee` only as one supported backend, as suggested by @QRemy in https://github.com/gammapy/gammapy/pull/2255#issuecomment-509158205 .

adonath:
@facero Do you maybe gave time to work on this in the coming weeks?

facero:
Yes, good idea. This is long time overdue.

adonath:
@facero Any chance the MCMC code will get some attention soon?

facero:
I'm really bad at following up this topic. Sorry. Not sure I'll be able to work on it this week as we're preparing the CTA consortium meeting + I'm attending another collaboration meeting this week. If you find some other volunteers that would probably work faster

adonath:
Preliminary solved with #3429, further discussion are in #3430.

PR #2546, opened 2019-11-14
Rename likelihood to stat

### LINKED ISSUE #2149: Introduce a uniform term for -2 * log(likelihood)?
In the past the `.likelihood()` method of the `Dataset` object has caused confusion, because it does not return the likelihood, but rather  `- 2 * log(likelihood)`. Apart from improving the documentation, there is probably a good reason to rename the functions and use the convention everywhere in Gammapy. 

What we have now:
- `Dataset.likelihood()` and `Dataset.likelihood_per_bin()`
- `OptimizeResult.total_stat`
- `gammapy.stats.fit_statistic`
- `loglike` and `dloglike` for flux points (see https://gamma-astro-data-formats.readthedocs.io/en/latest/spectra/binned_likelihoods/index.html#required-columns)
-  `total_stat`, `best_fit_stat`, etc. as variable names

What we have now is technically consistent in the sense that we only work with `- 2 * log(L)` and never with the likelihood directly or without the factor of 2. Still it would nice to reflect this in the method and variable names, so it becomes easier for users as well as developers to understand which quantity we refer to. 

My proposal would be to change to `fit_statistic` everywhere in Gammapy but leave the flux-points spec untouched for now. Any other proposals or opinions, @registerrier, @AtreyeeS @facero, @Bultako?

cdeil:
This pull request fixes #2149, it renames "likelihood", "loglike" terms to "stat".
The idea is that we always use the fit statistic consistently, and for Poisson likelihood it is given by `stat = -2 * log(L)`.
I went with the variant that I prefer, where I use an explicit `stat_sum` and `stat_array` for cases where both sum and array are involved. In output results where only the sum is stored, I use `stat`.

I noticed that the flux point and light curve code is a lot of duplicated copy & paste, I had to do the same edits twice (possibly introducing bugs along the way). Can we do better?

Tests pass, but @adonath - there's probably places I missed?

(I didn't do the `dloglike` cases yet. Are those TS values? -> have to check)

### DIFF
--- a/gammapy/cube/fit.py
+++ b/gammapy/cube/fit.py
@@ -464,7 +464,7 @@ class MapDataset(Dataset):
         if self.gti and other.gti:
             self.gti = self.gti.stack(other.gti).union()
 
-    def likelihood_per_bin(self):
+    def stat_array(self):
         """Likelihood per bin given the current model parameters"""
         return self._stat(n_on=self.counts.data, mu_on=self.npred().data)
 
@@ -1049,7 +1049,7 @@ class MapDatasetOnOff(MapDataset):
         """Excess (counts - alpha * counts_off)"""
         return self.counts.data - self.background.data
 
-    def likelihood_per_bin(self):
+    def stat_array(self):
         """Likelihood per bin given the current model parameters"""
         mu_sig = self.npred().data
         on_stat_ = wstat(
--- a/gammapy/cube/fit.py
+++ b/gammapy/cube/fit.py
@@ -165,7 +165,7 @@ class MapDataset(Dataset):
 
         stat = np.nan
         if self.model is not None or self.background_model is not None:
-            stat = self.likelihood()
+            stat = self.stat_sum()
         str_ += "\t{:32}: {:.2f}\n\n".format("Fit statistic value (-2 log(L))", stat)
 
         # model section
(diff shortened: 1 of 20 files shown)


cdeil:
The remaining task here is to deal with the SED profiles.

From the description here, it sounds to me like `loglike`, `loglike_null` and `dloglike_scan` should not include the factor 2, whereas `ts` does include it:
https://gamma-astro-data-formats.readthedocs.io/en/latest/spectra/flux_points/index.html#likelihood-columns

I think there was a bug then, because here we put `-2 * dloglike` in `dloglike_scan` column?
https://github.com/gammapy/gammapy/blob/ef80dc1ebc5b8160ee120da40af8a2f7c1246f9c/gammapy/spectrum/flux_point.py#L1088-L1089
So the factor 2 is incorrect, and for the sign I'm not sure.

@adonath - ?

cdeil:
@adonath - Ready for final review. Locally `make test` and `make docs-all` passes.
Can we merge this in quickly? It touches 20 files and leaving it open blocks other work.

I didn't touch `sampling.py`. Would prefer to leave this to a separate PR, see #2304 .
It's better to add a test first, and it's ~ 1 day of work.

I checked the plot in `tutorials/spectrum_analysis.ipynb` - looks ok.

cdeil:
@adonath - Thanks for the quick review.

PR #2552, opened 2019-11-16
Remove unused functions from `gammapy.irf`

adonath:
**Description**
This PR removes some unused functions from `gammapy.irf` and `gammapy.cube`:
- Remove untested and unused `_fov_background_norm`. The plan is to introduce this as a `FoVBackgroundMaker` or similar
- Remove `apply_containment_fraction()` and implement the 2 lines of code, where they are used
- Remove `compute_energy_thresholds()`, which is now superseded by the `SafeMaskMaker`
- Remove `make_mean_edisp()`, the same can be achieved with `SpectrumDataset.stack()` now. 

**Dear reviewer**
No real review required, once the CI builds are green I'll go ahead an merge.


cdeil:
@adonath - What about `make_mean_psf`? (see [here](https://github.com/gammapy/gammapy/blob/ef014b21c7c82ba72c2f4d895102161999f4c70b/gammapy/irf/irf_reduce.py#L52-L81))

Do you want to keep it or replace it with something else?

I noticed it in static analysis, because PyCharm complains about the coding pattern that can lead to `UnboundLocalError`:
```
>>> from gammapy.irf import make_mean_psf
>>> make_mean_psf([], 42)
  File "<stdin>", line 1, in <module>
  File "/Users/deil/work/code/gammapy/gammapy/irf/irf_reduce.py", line 81, in make_mean_psf
    return stacked_psf
UnboundLocalError: local variable 'stacked_psf' referenced before assignment
```
I guess if the code to create one object isn't a lot, then we should put one line code duplication before and in the loop, and rewrite it like this?
```python
def make_mean_psf(observations, position, energy=None, rad=None):
    psf = make_psf(observations[0], position, energy, rad)
    for observation in observations[1:]:
        psf_obs = make_psf(observation, position, energy, rad)
        psf = psf.stack(psf_obs)

    return psf
```
This will lead to an `IndexError`, which seems appropriate for that case, and is much nicer for users than an `UnboundLocalError`.

Another coding pattern we could use:
```
def make_mean_psf(observations, position, energy=None, rad=None):
    return EnergyDependentTablePSF.from_stack(
        make_psf(obs, position, energy, rad)
        for obs in observations
    )
```

This might be a general point for functions where we accumulate or stack, i.e. we should probably clean up in other places as well and always use the same coding pattern.

Issue #2512, opened 2019-11-04
Gammapy validation: joint Crab analysis


cdeil:
As part of the Gammapy validation effort while preparing v1.0, we should try to reproduce the joint likelihood spectral analysis for the Crab nebula that we did with Gammapy v0.9 a wile ago.

This is a good use case for the high-level joint dataset and fitting API we are currently developing in Gammapy (see e.g. #2505), and to test parameter error and 2D error contour computations. This would also be a good check for the safe energy mask handling and energy dispersion apply for 1D spectral analysis, topics that were checked in detail when doing the joint Crab analysis.

A task description and references are here:

- https://github.com/gammapy/gammapy-benchmarks
- https://github.com/gammapy/gammapy-benchmarks/blob/master/validation
https://github.com/open-gamma-ray-astro/joint-crab

This is on a best-effort basis, the starting point should be to try to load the reduced OGIP data and reproduce the joint log parabola fit. Then one could continue and either reproduce the analysis including the modified likelihood that includes systematics (see if that is possible or even easy with the current dataset API), and / or look at the parameter errors and correlations in detail.

For this, you don't have to be a Gammapy developer, what's required is some Python & Gammapy & gamma-ray experience, and a few days of free time in Nov 2019.

We're looking for help! If you're interested in this task, please leave a comment here, or contact me on Slack.

cc @cosimoNigro @kbruegge @Bultako @TarekHC  @QRemy

registerrier:
joint-crab validation is under way. See:
https://github.com/gammapy/gammapy-benchmarks/blob/master/validation/joint-crab/README.md

For now agreement is good, except for Fermi. The `Dataset` building approach in this case is significantly different from what it used to be. Here we go from a `MapDataset` to a `SpectrumDataset`. The resulting `aeff` seems to be off. I am currently checking for this.

registerrier:
The issue dates from the joint-crab paper. The coordinate used is passed in galactic frame to a `Map` defined in celestial coordinates. This results in an improper exposure estimate at the Crab position.
I changed the `joint-crab` repo code to extract the exposure at the right position and we now get consistent results.

adonath:
@registerrier Anything left to be done here? Is the joint fit already available?

registerrier:
Yes the joint fit is done see: 
https://github.com/gammapy/gammapy-benchmarks/blob/master/validation/joint-crab/results/joint_summary.md

[...]

PR #2553, opened 2019-11-17
Remove MapDataset cstat likelihood option

cdeil:
This is a follow-up to #2546, where we renamed "likelihood" to "stat".

While implementing that PR, I noticed that `MapDataset` had an option `likelihood="cash"` which was a string, but then it was set to `self.likelihood_type=likelihood` in init, and in the class `likelihood` actually was the function, not the string. I found this confusing.

Now that the function `likelihood` was renamed to `stat` it's better, but still, I would prefer to be more explicit and to either have `likelihood_type` or `stat_name` or somthing like that as argument for `__init__`, and then to have the same name for the attribute, or to remove the option completely.

This PR removes the option, which is my preference, because then `MapDataset` and `SpectrumDataset` are the same, in `SpectrumDataset` it was already hard-coded to the `cash` statistic. I've actually never seen any gamma-ray paper use `cstat`, I think everyone uses `cash`. `cstat` is something that was only used in a few stats papers where the relation to `chi2` was explored, showing that they match in the high-stats limit - which we never have in gamma-ray astronomy, there's always low stats at high energies and in small spatial bins, so really IMO offering `cstat` isn't doing users a favour, and it's not worth the extra effort (documentation, testing). Instead we should offer more powerful ways to customise the likelihood than by a string, e.g. via subclassing and a datasets registry (see #2512 and #2387), for the few % of users that need it (they likely won't want `cash`, but something else).

@adonath - Thoughts?

### DIFF
--- a/gammapy/cube/fit.py
+++ b/gammapy/cube/fit.py
@@ -16,7 +16,7 @@ from gammapy.maps import Map, MapAxis, WcsGeom
 from gammapy.modeling import Dataset, Parameters
 from gammapy.modeling.models import BackgroundModel, SkyModel, SkyModels
 from gammapy.spectrum import SpectrumDataset
-from gammapy.stats import cash, cash_sum_cython, cstat, cstat_sum_cython, wstat
+from gammapy.stats import cash, cash_sum_cython, wstat
 from gammapy.utils.random import get_random_state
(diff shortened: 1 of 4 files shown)


cdeil:
Done in 61fda765c. Merging.

PR #2569, opened 2019-11-20
Clean up of gammapy.time

registerrier:
**Description**

This pull request cleans a bit `gammapy.time`.

- It removes variability tests methods from the `LightCurve` object and make them stand alone functions.
- It removes the `gammapy.time.exptest` function which works only on events. Keeping it is confusing because it is not compatible with the binned datasets we produce for now.
- Rapid update of the rst documentation

**Dear reviewer**

the PR is ready for review.


cdeil:
> removes the gammapy.time.exptest function which works only on events. Keeping it is confusing because it is not compatible with the binned datasets we produce for now.

Suggest to put `exptest` back, IMO this is going in the wrong direction IMO.

Since we have different views what to keep and remove, can we please have a short lead dev discussion before removing things, or wait for lead dev approval on PRs deleting stuff?

There are binned and unbinned variability tests. In CTA and HESS (see F. Brun upcoming paper) we need both. `exptest` is a published method people use, the code and test and docstring quality was good, v1.0 quality. Gammapy has methods to work with events, those have times, it's completely usable now, nothing to do. Simulating events should come ASAP, but that's not needed to have methods that work with events, we have EventList now.

Maybe one sentence was missing in the docs "Gammapy currently has one method to check variability using event lists and event arrival times: exptest."?

PR #2570, opened 2019-11-20
Adapt simulation notebooks

AtreyeeS:
Adapt simulation notebooks for binned simulations!

We have working notebooks now using the in-memory `Observation` class and the `Dataset.fake()` method. The scope of the 1D and 3D tutorials remain the same as before, but we might adapt them in future


adonath:
This PR was accidentally opened from an incorrect branch. Closing now.

PR #2574, opened 2019-11-20
Generalisation of the power law with an exponential cut-off

bkhelifi:
**Description**
This PR contains the generalisation of the power law with an exponential cut-off, adding a power inside the exponential term.

**Dear reviewer**
The model has been improved by adding an 'alpha' parameter, being egual to 1 by default. Its peak function has been adapted accordingly, as an analytical function exists (its formula has been re-checked using plots). 
In parallel, the corresponding test function has been updated to test the computation with an alpha different from 1.


gammapy/modeling/models/spectral.py
```diff
@@ -771,17 +773,18 @@ class ExpCutoffPowerLawSpectralModel(SpectralModel):
     amplitude = Parameter("amplitude", "1e-12 cm-2 s-1 TeV-1")
     reference = Parameter("reference", "1 TeV", frozen=True)
     lambda_ = Parameter("lambda_", "0.1 TeV-1")
+    alpha_ = Parameter("alpha_", "1.0", frozen=True)
```
registerrier: why `alpha_`?

gammapy/modeling/models/spectral.py
```diff
@@ -771,17 +773,18 @@ class ExpCutoffPowerLawSpectralModel(SpectralModel):
...
         """Evaluate the model (static function)."""
         pwl = amplitude * (energy / reference) ** (-index)
         try:
-            cutoff = np.exp(-energy * lambda_)
+            cutoff = np.exp(- np.power(energy * lambda_, alpha_))
         except (AttributeError, TypeError):
             from uncertainties.unumpy import exp
 
-            cutoff = exp(-energy * lambda_)
+            cutoff = exp(- pow(energy * lambda_, alplha_))
```
registerrier: misspelling

registerrier: also use `np.power`

gammapy/modeling/models/spectral.py
```diff
@@ -791,16 +794,17 @@ def e_peak(self):
...
         reference = p["reference"].quantity
         index = p["index"].quantity
         lambda_ = p["lambda_"].quantity
-        if index >= 2:
+        alpha_ = p["alpha_"].quantity
+        if index >= 2 or lambda_ == 0. or alpha_ == 0.:
             return np.nan * reference.unit
         else:
-            return (2 - index) / lambda_
+            return pow((2 - index)/alpha_, 1/alpha_) / lambda_
```
registerrier: `np.power`

registerrier:
There is a remaining issue in the `analysis_3d.ipynb` notebook. This is due to the size of the covariance matrix which has changed. 
I think the operation required is ugly and error prone. I don't know if we can avoid it.

PR #2587, opened 2019-11-21
Activate EDispMap in MapEvaluator

adonath:
**Description**

This work was done pair-coding with @AtreyeeS. This pull request changes to allow using an `EDispMap` in the `MapEvaluator`. To simplify the adaption we allow for both an `EnergyDispersion` and `EDispMap` object for now. The following tests we already adapted to use the `EDispMap`:
- `gammapy/cube/tests/test_fit.py`
- `gammapy/spectrum/tests/test_flux_points_estimator.py`
- `gammapy/time/tests/test_lightcurve.py`

The change in the flux point estimator and lightcurve tests are due to to fact of using `simulate_dataset()` to generate the test data. The IRFs were taken at an offset of `1 deg`
independent of the position of the model component. After this PR the "correct" edisp (defined
at the position of the model component) is taken.

This PR also adds the following helper functions:
- `PSFMap.from_energy_dependent_table_psf()`
- `EDsipMap.from_diagonal_response()`

In follow-up PRs we will adapt to use `PSFMap` as well and probably disallow to an `EnergyDispersion` object for the `MapDataset`.


gammapy/cube/tests/test_psf_map.py
```diff
@@ -361,3 +361,25 @@ def test_make_mean_psf(data_store):
 
     assert not np.isnan(psf.psf_value.value).any()
     assert_allclose(psf.psf_value.value[22, 22], 12206.1665, rtol=1e-3)
+
+
+@pytest.mark.parametrize(
+    "position", [
+        SkyCoord("0 deg", "0 deg"),
```
cdeil: You can just put `["0d 0d", "180d 0d", ...]` here, and then put `SkyCoord(position)` once in the code below, might on one line then. (same in the test above).
And I think you have to add `@requires_data`.

PR #2436, opened 2019-11-21
LogGaussianSpectralModel mean position

cdeil:
While doing #2435 I noticed that the LogGaussianSpectralModel mean has unit energy, but doesn't seem to be the parabola peak location when plotting the spectral model in log flux vs log energy. Looks like a bug either in the formula or from not handling Quantity properly?

- https://gist.github.com/cdeil/ee98cb276e06621ca361085319a755d7
- https://docs.gammapy.org/0.14/api/gammapy.modeling.models.LogGaussianSpectralModel.html

Fermi ST or ctools only has the normal Gaussian spectral model? Maybe we could check against those at least? Or check for approximate consistency between our two Gaussian distributions (in lin and log energy)? For the same norm, mean and sigma, with very small sigma, the two distributions should be almost the same, no?

@JouvinLea or @QRemy or anyone - interested to work on this?
(fix if it's a bug, and in any case improve docstring & tests a bit, i.e. send a PR).

### DIFF
--- a/gammapy/modeling/models/__init__.py
+++ b/gammapy/modeling/models/__init__.py
@@ -43,7 +43,6 @@ SPECTRAL_MODELS = Registry(
         LogParabolaSpectralModel,
         TemplateSpectralModel,
         GaussianSpectralModel,
-        LogGaussianSpectralModel,
         AbsorbedSpectralModel,
         NaimaSpectralModel,
         ScaleSpectralModel,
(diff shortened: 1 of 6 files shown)


cdeil:
I changed the weirdly parameterised LogGaussianSpectralModel to _LogGaussianSpectralModel and put it in the cosmic ray spectral model file where it's used. Changed the parameter names to no longer suggest it's the `mean` and `width`. And removed this from the Gammapy model registry and docs tutorial.

So we keep this model for the one place it's used, but we don't expose and maintain a `LogGaussianSpectralModel` with this parametrisation. Whether or not we add `LogGaussianSpectralModel` in the future is TBD.

PR #2596, opened 2019-11-22
Remove error when ReflectedRegionsBackgroundMaker finds no off region

registerrier:
**Description**

This pull request corrects the `ReflectedRegionsBackgroundMaker` if no OFF region is found. 
Now a `SpectrumDatasetOnOff` is returned with `count_off` set to None and `acceptance_off` set to 0. 
In order to avoid issues for users, the loop in `Analysis.get_observations()` contains a `log.info` to warn that an observation has no OFF region. The dataset is discarded from the final list. 

A test has been added on the `ReflectedRegionsBackgroundMaker`. But not on the `Analysis` class. Maybe there should be one too.

**Dear reviewer**

This is ready for review, with the caveat concerning the test on `Analysis`


registerrier:
All tests pass. Merging.

PR #2597, opened 2019-11-22
Simplify `MapDataset` simulation

AtreyeeS:
**Description**
!-- What is the motivation for this pull request? --
This PR simplifies MapDataset simulation.
Instead of having to write 6 lines of code to create a fake observation and an empty dataset (which can be quite non-intuitive), the user just needs to call `simulate_map` and pass the simulation parameters.

Adapts the old `simulate_dataset.py` and uses it in `simulate_3d` notebook

**Dear reviewer**

I do not know if it might be useful to do the same for spectral simulation as well.
On one hand, it will make the API uniform, and simulation really easy.
But for spectral simulation, users would probably like to play to different sorts of background shapes etc, so I don't know how we can support that easily.

Also, I propose that we club the simulation tutorials into one in a future PR instead of having separate ones for Map and spectrum.

### DIFF
--- a/gammapy/cube/simulate.py
+++ b/gammapy/cube/simulate.py
@@ -1,12 +1,8 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 """Simulate observations"""
 import astropy.units as u
-from gammapy.cube import (
-    MapDataset,
-    PSFKernel,
-    make_map_background_irf,
-    make_map_exposure_true_energy,
-)
+from gammapy.data import Observation
+from gammapy.cube import MapDatasetMaker, MapDataset
 from gammapy.maps import WcsNDMap
 from gammapy.modeling.models import BackgroundModel
 from gammapy.utils.random import get_random_state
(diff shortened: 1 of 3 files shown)


AtreyeeS:
Probably this can be a class method on `MapDataset`? and users do a `MapDataset.simulate()` (and a `SpectralDataset.simulate()` )

adonath:
At the coding sprint on Friday we agreed to re-use the `Analysis` class for the data reduction part of the simulation. @AtreyeeS Do you still think this is feasible?

AtreyeeS:
The analysis class needs to be adapted to be useful in this case. Specifically, there needs to an option to choose the types of maps to be created, as passed as the `selection` argument in `MapDatasetMaker.run()`.
Once that I adapted, we can review the simulation statues. I suppose we should close this PR.

cdeil:
@AtreyeeS - There' already quite a few merge conflicts here. Sorry! Can you rebase and fix tests so that we can merge this in? Or would it be easier to close this PR and re-start from master?

AtreyeeS:
It is easier to close this PR and start afresh. I close this now

Issue #2431, opened 2019-10-02
Avoid interpolate float precision warning 


cdeil:
Since Gammapy v0.14 we show this warning to end users for most analysis tutorials:
```
/Users/adonath/github/adonath/gammapy/gammapy/utils/interpolation.py:159: Warning: Interpolated values reached float32 precision limit
  "Interpolated values reached float32 precision limit", Warning
```

See https://docs.gammapy.org/0.14/search.html?q=float32+precision+limit&check_keywords=yes&area=default

We have to handle such issues internally. If we emit warnings to users for things that are normal, it reduces trust and teaches them to ignore warnings from Gammapy.

I didn't look at #2405 in detail, I realise it's a difficult issue, but I think we should revisit it again. Is simply removing the warning a possibility? Is no special handling a possibility, would the code use `-np.inf` and we get an (acceptable?) value of 0 for those values?

Was added here:

https://github.com/gammapy/gammapy/pull/2405/files#diff-64254d0bf18d1bc87ecece0b70ca61caR158-R159

@adonath - Assigning to you.
cc @QRemy

cdeil:
Fixed by @QRemy in #2598 .

PR #2598, opened 2019-11-22
Remove interpolation warning

QRemy:
Remove interpolation warning

### DIFF
--- a/gammapy/utils/interpolation.py
+++ b/gammapy/utils/interpolation.py
@@ -145,6 +145,7 @@ class LogScale(InterpolationScale):
         values = np.clip(values, self.tiny, np.inf)
         return np.log(values)
 
+    @staticmethod
     def _inverse(self, values):
         output = np.exp(values)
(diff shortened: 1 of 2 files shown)


cdeil:
@QRemy - Thanks!

Please always x-ref the relevant issues (and assign those to yourself before you start working on them).
In this case: This PR fixes #2431 .

It should be possible to write this in a simpler way. Is this equivalent?
```
    @staticmethod
    def _inverse(self, values):
        output = np.exp(values)
        return np.where(abs(output) - self.tiny <= self.tiny, 0, output)
```

There's no test for this?
@QRemy - Could you please add a file `gammapy/utils/tests/test_interpolate.py` and add a unit test for this?

Especially here, this is something that has a high potential to work on one machine / OS / Numpy version, but not another. So it's very valuable to have a test for CI and maintainers that shows what behaviour & output we expect. It shouldn't be a lot of tests, overall <~ 10 lines of code in that file.

cdeil:
@QRemy - Thank you!

Issue #2434, opened 2019-10-03
Remove parameter scaling?


adonath:
In preparation of the H.E.S.S. DL3 data tutorial, I found issues with convergence of the model fits both using the `iminuit` and `sherpa` backend. In both cases the optimiser would try positions outside the map already on the first iteration, which currently leads to an error (whether this behaviour is a good one, is a different discussion...). The reason for this weird behaviour is, that we currently scale **all** parameters before the fit and only pass the `factors` to the optimiser.  E.g. if you do a crab analysis in `icrs` coordinates, the optimiser will try a step size in the order of ~10 deg in longitude, which does not make sense. There are now different solutions to fix this behavior:

- Some optimisers can take an initial estimate of the covariance matrix. We could assume plausible default errors for all parameters, but we need a way to build a global covariance matrix form these sub-covariances, e.g. by adding a `Parameters.concatenate()` method.
- We could add a `autoscale` flag to our `Parameter` class, to switch of the scaling for certain parameters (especially position...)
- We could remove the parameter scaling completely. This would actually be my favoured solution. Currently the parameter scaling is only really required (if at all) for the spectral `amplitude` parameter, where the diagonal entries on the covariance matrix can get very small. As we support units for parameters, the scaling can be achieved natively, by moving the scale to the unit of the parameter like `amplitude = u.Quantity("1.34 1e-12 TeV-1 cm-2 s-1")`. This puts a little bit of responsibility on the side of the user, but this can be handled with documentation if needed. Removing the scaling would simplify our `Parameter` class a lot. 

Any opinions @registerrier and @cdeil?

registerrier:
I completely agree that in many cases the scaling is meaningless, in particular for coordinates which are defined only to a closed interval in the real set. 
Technically, it is probably only useful for `amplitude` parameter which can be distributed over several decades. We could alternatively use a log-like representation of the amplitude in the minimizer.  

Often other tools (e.g. sherpa, XSpec) have a `guess` or `norm` function that try to find the good `amplitude` of the model w.r.t. to the data. We could get rid of the amplitude scaling by adding such a functionality, namely `Fit.guess()`. This would require that we have a way to determine 'guessable' parameters, so this is likely not very different from adding an option `autoscale`.

So we might explore your second solution where the only `autoscale=True` parameters would be the amplitudes. If we have all the code there already for scaling, it might be bad to undo everything no?

cdeil:
[...]

PR #2602, opened 2019-11-24
Test error propagation for NaimaModel

luca-giunti:
**Description**

I implemented a minimal test for error propagation in case of NaimaModel. Notice that the test is failing by now: fixes elsewhere are needed before merging.

Here is a code snippet to reproduce the problem:
```
from gammapy.modeling.models import PowerLawSpectralModel, NaimaSpectralModel
import astropy.units as u
import naima

#Test error propagation for a simple powerlaw
model=PowerLawSpectralModel()
model.parameters.set_error(amplitude=0.1 * model.amplitude.value)
dnde, dnde_err = model.evaluate_error(1 * u.TeV)
print(dnde_err / dnde)

#Test error propagation for a naimamodel
particle_distribution = naima.models.PowerLaw(
    amplitude=2e33 / u.eV, e_0=10 * u.TeV, alpha=2.5
)
radiative_model = naima.radiative.PionDecay(
    particle_distribution, nh=1 * u.cm ** -3
)
model = NaimaSpectralModel(radiative_model)
model.parameters.set_error(amplitude=0.1 * model.amplitude.value)
dnde, dnde_err = model.evaluate_error(1 * u.TeV)
print(dnde_err / dnde)
```
The second print statement returns `0.0` instead of the expected value `0.1`.  The problem seems to be due to the large amplitude value 2e33 , if I change that to a typical value like 1e-11 then I do get non-zero output.

This might be related to #2434, or to the introduction of an absolute `eps` in [`_evaluate_gradient`](https://github.com/gammapy/gammapy/blob/a2adc298d1635948c69acc4807b4416dbaba0b16/gammapy/modeling/models/spectral.py#L50).

**Dear reviewer**
Please, commit some fixes before merging... Thanks!

### DIFF
--- a/gammapy/modeling/models/tests/test_spectral.py
+++ b/gammapy/modeling/models/tests/test_spectral.py
@@ -447,6 +447,12 @@ class TestNaimaModel:
         val = model(self.e_array)
         assert val.shape == self.e_array.shape
 
+        # Test error propagation
+        # TODO: This test is failing. it has to be fixed before merging
+        model.parameters.set_error(amplitude=0.1 * model.amplitude.value)
+        dnde, dnde_err = model.evaluate_error(1 * u.TeV)
+        assert_allclose(dnde_err / dnde, 0.1)
+
     def test_ic(self):
         import naima
 
@@ -581,12 +587,6 @@ class TestSpectralModelErrorPropagation:
 
         out = model.evaluate_error(0.1 * u.TeV)
         assert_allclose(out.data, [1.548176e-10, 1.933612e-11], rtol=1e-3)
-
-    def test_naima_model_error_proprgation(self):
-        # Regression test for Naima model
-        # https://github.com/gammapy/gammapy/issues/2190
-        # TODO: implement test case. Move to Naima model tests!
-        pass
 
     def test_absorption_model_error_propagation(self):
         # Regression test for absorption model


cdeil:
@luca-giunti - Thanks!

I think this happens because the step is too small in this case, not representable by float:
```
>>> 2e33 + 1e-12 == 2e33
True
```

@adonath - Should we change to take a step that's a fraction of the parameter value?
(as in the prototype [here](https://nbviewer.jupyter.org/github/gammapy/gammapy-extra/blob/master/experiments/uncertainty_estimation_prototype.ipynb#Differentials))
Or is there a better way?

Another point: the [approx_jacobian](https://github.com/scipy/scipy/blob/41763e3ea532fc0e9305dde87a48ab3b4d14eac8/scipy/optimize/slsqp.py#L31) doesn't seem to be part of the public API, and also we don't use the vector-valued feature, no? So probably we should change to [scipy.optimize.approx_fprime](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.approx_fprime.html)?
According to the docstring it does take an array-valued `epsilon`, so should be OK for our use case where we need to have a sensible stepsize for each parameter.

adonath:
I'll look into this now...

adonath:
Thanks @luca-giunti! The test and fix is now included in #2623 .

PR #2603, opened 2019-11-24
Test error propagation for AbsorbedSpectralModel

luca-giunti:
**Description**
I implemented a minimal test for error propagation in case of `AbsorbedSpectralModel`. Related issue: #1046


cdeil:
@luca-giunti - Thanks!

PR #2604, opened 2019-11-25
Implement additional methods on the safe mask maker

luca-giunti:
In this PR:
- I implemented the `SafeMaskMaker.make_mask_edisp_bias(...)` method for `MapDataset` and `MapDatasetOnOff`, that was previously raising `NotImplementedError`
- I implemented a new `SafeMaskMaker.make_mask_bkg_peak(...)` method, motivated by the HESS DL3 analysis validation paper ([page 11, section 5.5.1](https://arxiv.org/pdf/1910.08088.pdf))

As soon as this gets merged, I will resolve corresponding the [TODO](https://github.com/gammapy/gammapy-benchmarks/blob/3097a6a73c13b16f5b170002b7c0ef7461725100/validation/hess-dl3-dr1/make.py#L207) in the validation script. This will make the code simpler and shorter there.

### DIFF
--- a/gammapy/cube/make.py
+++ b/gammapy/cube/make.py
@@ -295,7 +295,8 @@ class SafeMaskMaker:
         Percentage of the energy bias to be used as lower
         energy threshold for method "edisp-bias"
     position : `~astropy.coordinates.SkyCoord`
-        Position at which the `aeff_percent` or `bias_percent` are computed.
+        Position at which the `aeff_percent` or `bias_percent` are computed. By default,
+        it uses the position of the center of the map.
     offset_max : str or `~astropy.units.Quantity`
         Maximum offset cut.
     """
@@ -416,12 +417,11 @@ class SafeMaskMaker:
         edisp = dataset.edisp
 
         if isinstance(dataset, (MapDataset, MapDatasetOnOff)):
-            if self.position is None:
-                raise AttributeError(
-                    "To compute the energy bias, a position has to be set"
-                )
+            position = self.position
+            if position is None:
+                position = dataset.counts.geom.center_skydir
             e_reco = dataset.counts.geom.get_axis_by_name("energy").edges
-            edisp = edisp.get_energy_dispersion(self.position, e_reco)
+            edisp = edisp.get_energy_dispersion(position, e_reco)
             counts = dataset.counts.geom
         else:
             counts = dataset.counts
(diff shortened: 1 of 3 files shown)


gammapy/cube/make.py
```diff
@@ -395,21 +405,59 @@ def make_mask_energy_edisp_bias(self, dataset):
...
         mask_safe : `~numpy.ndarray`
             Safe data range mask.
         """
+        edisp = dataset.edisp
+
         if isinstance(dataset, (MapDataset, MapDatasetOnOff)):
-            raise NotImplementedError(
-                "'edisp-bias' method currently only supported for spectral datasets"
-            )
+            if self.position is None:
```
adonath: Maybe we compute the threshold at the center of the map as a default? However I'm not sure this is a safe assumption. Do you know @registerrier what is the typical quality of IRFs at `offset=0 deg`?

gammapy/cube/make.py
```diff
@@ -395,21 +405,59 @@ def make_mask_energy_edisp_bias(self, dataset):
...
+        if isinstance(dataset, (MapDataset, MapDatasetOnOff)):
+            background_spectrum = dataset.background_model.map.get_spectrum()
+            counts = dataset.counts.geom
+        else:
+            background_spectrum = dataset.background
+            counts = dataset.counts
+
+        peak = background_spectrum.data.max()
+        idx = list(background_spectrum.data).index(peak)
+        e_min = background_spectrum.energy.center[idx]
```
adonath: Note the implementation differs from the description "upper edge of the energy bin with the highest predicted background rate". Please unify to whatever is described in the paper...

gammapy/cube/make.py
```diff
@@ -395,21 +405,59 @@ def make_mask_energy_edisp_bias(self, dataset):
...
+            counts = dataset.counts.geom
+        else:
+            counts = dataset.counts
 
-        e_min = dataset.edisp.get_bias_energy(self.bias_percent / 100)
-        return dataset.counts.energy_mask(emin=e_min)
+        e_min = edisp.get_bias_energy(self.bias_percent / 100)
+        return counts.energy_mask(emin=e_min)
+
+    def make_mask_bkg_peak_energy(self, dataset):
```
adonath: Minor: please rename to `make_mask_energy_bkg_peak()`, so that the all energy mask methods start with `make_mask_energy_xxx`

gammapy/cube/make.py
```diff
@@ -395,21 +405,59 @@ def make_mask_energy_edisp_bias(self, dataset):
...
+        """
+
+        if isinstance(dataset, (MapDataset, MapDatasetOnOff)):
+            background_spectrum = dataset.background_model.map.get_spectrum()
+            counts = dataset.counts.geom
+        else:
+            background_spectrum = dataset.background
+            counts = dataset.counts
+
+        peak = background_spectrum.data.max()
```
adonath: This can be simplified using `np.argmax()` instead.

PR #2605, opened 2019-11-25
Add analysis config using pydantic

Bultako:
This PR implements a functional model based on `pydantic` framework for handling and validation settings/configuration parameters of the high-level interace. 

As exposed in last condig sprint, this technical solution solves [all the needs for settings handling](https://github.com/cdeil/gammapy-config#requirements-and-wishes) using an agreed [simpler syntax](https://github.com/cdeil/gammapy-config/blob/master/example-config.yaml) for the parameters names, structure and cofig file.

This PR is a first step towards replacement of the existing solution based on `jsonschema`, it just adds the model and features, but still does not remove what we have. Final implementation should come in a future PR once this is reviewed and accepted.

### DIFF
--- a/gammapy/analysis/config.py
+++ b/gammapy/analysis/config.py
@@ -29,7 +29,7 @@ class EnergyType(Quantity):
 
     @classmethod
     def validate(cls, v):
-        assert isinstance(Quantity(v).to("erg"), Quantity)
+        assert v.unit.physical_type == "energy"
         return EnergyType(v)
 
 
--- a/gammapy/analysis/config.py
+++ b/gammapy/analysis/config.py
@@ -1,8 +1,8 @@
 from astropy.coordinates import Angle
 from astropy.time import Time
 from astropy.units import Quantity
-from gammapy.utils.scripts import make_path, read_yaml
-from pydantic import BaseModel, FilePath, validator
+from gammapy.utils.scripts import read_yaml
+from pydantic import BaseModel, FilePath
 from pydantic.utils import deep_update
 from pathlib import Path
 from typing import List
@@ -161,10 +161,6 @@ class Data(GammapyBaseModel):
     obs_file: FilePath = None
     obs_cone: SpatialCircleRange = SpatialCircleRange()
     obs_time: TimeRange = TimeRange()
-
-    @validator("datastore")
-    def datastore_exists(cls, v):
-        return make_path(v)
 
 
 class Log(GammapyBaseModel):
--- a/gammapy/analysis/config.py
+++ b/gammapy/analysis/config.py
@@ -19,7 +19,7 @@ class AngleType(Angle):
 
     @classmethod
     def validate(cls, v):
-        return Angle(v)
+        return AngleType(v)
 
 
 class EnergyType(Quantity):
@@ -30,7 +30,7 @@ class EnergyType(Quantity):
     @classmethod
     def validate(cls, v):
         assert isinstance(Quantity(v).to("erg"), Quantity)
-        return Quantity(v)
+        return EnergyType(v)
 
 
 class TimeType(Time):
@@ -40,7 +40,7 @@ class TimeType(Time):
 
     @classmethod
     def validate(cls, v):
-        return Time(v)
+        return TimeType(v)
 
 
 class FrameEnum(str, Enum):
@@ -58,11 +58,8 @@ class GammapyBaseModel(BaseModel):
         extra = 'forbid'
         json_encoders = {
             AngleType: lambda v: f"{v.value} {v.unit}",
-            Angle: lambda v: f"{v.value} {v.unit}",
             EnergyType: lambda v: f"{v.value} {v.unit}",
-            Quantity: lambda v: f"{v.value} {v.unit}",
             TimeType: lambda v: f"{v.value}",
-            Time: lambda v: f"{v.value}",
         }
 
     @classmethod
--- a/gammapy/analysis/config.py
+++ b/gammapy/analysis/config.py
@@ -1,13 +1,13 @@
+from enum import Enum
+from pathlib import Path
+from typing import List
 from astropy.coordinates import Angle
 from astropy.time import Time
 from astropy.units import Quantity
-from gammapy.utils.scripts import read_yaml
+import yaml
 from pydantic import BaseModel, FilePath
 from pydantic.utils import deep_update
-from pathlib import Path
-from typing import List
-from enum import Enum
-import yaml
+from gammapy.utils.scripts import read_yaml
 
 __all__ = ["AnalysisConfig"]
 
--- a/gammapy/analysis/config.py
+++ b/gammapy/analysis/config.py
@@ -1,9 +1,12 @@
-from typing import List
-import yaml
-from astropy.time import Time
 from astropy.coordinates import Angle
+from astropy.time import Time
 from astropy.units import Quantity
-from pydantic import BaseModel
+from gammapy.utils.scripts import make_path, read_yaml
+from pydantic import BaseModel, FilePath, validator
+from pathlib import Path
+from typing import List
+from enum import Enum
+import yaml
 
 __all__ = ["Config", "General"]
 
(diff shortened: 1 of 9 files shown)


gammapy/analysis/config.py
```diff
@@ -0,0 +1,189 @@
...
+
+
+class AngleType(Angle):
+    @classmethod
+    def __get_validators__(cls):
+        yield cls.validate
+
+    @classmethod
+    def validate(cls, v):
+        return AngleType(v)
```
cdeil: Can you please add some of the basic tests on what types the config attributes are that we used in the prototype, like here?
https://github.com/cdeil/gammapy-config/blob/422398b002545804bfed91d69c95c96b9cca2d15/test_config_pydantic.py#L12
(one for each custom type you define, use an `isinstance` in the test to check what the config attribute is)

I think here we should `return Angle(v)` to get an angle attribute, not the `AngleType` subclass, no?

Looking at https://pydantic-docs.helpmanual.io/usage/types/#custom-data-types and  https://github.com/samuelcolvin/pydantic/blob/master/pydantic/types.py I don't see them doing type casts, only type checks, so I wonder if we're doing it correctly. For sure `Angle(v)` does what we want, see the tests we had.

gammapy/analysis/config.py
```diff
@@ -0,0 +1,189 @@
...
+
+
+class EnergyType(Quantity):
+    @classmethod
+    def __get_validators__(cls):
+        yield cls.validate
+
+    @classmethod
+    def validate(cls, v):
+        assert isinstance(Quantity(v).to("erg"), Quantity)
```
cdeil: Suggest to write this:
```
v = Quantity(v)
assert v.unit.physical_type == "energy"
return v
```
The main advantage is that this converts to a Quantity once, whereas the current code does it twice.

gammapy/analysis/config.py
```diff
@@ -0,0 +1,189 @@
...
+        return yaml.dump(yaml.safe_load(self.json()))
+
+    def update_from_dict(self, other):
+        data = deep_update(self.dict(exclude_defaults=True), other.dict(exclude_defaults=True))
+        return Config(**data)
+
+
+class Skydir(GammapyBaseModel):
+    frame: FrameEnum = FrameEnum.icrs
+    lon: AngleType = AngleType("83.633 deg")
```
cdeil: Please rename to `SkyCoordType`, for naming consistency with the other types above.

The Crab as default position is no good. Last week I made the mistake myself to use a config template and get a source position which of course wasn't my source position. Can you just remove the default here for now?
If it is needed, then the default should be put where `SkyCoord` is defined, with a `#TODO: try to remove the default`.

gammapy/analysis/config.py
```diff
@@ -0,0 +1,189 @@
...
+    lat: AngleType = AngleType("22.014 deg")
+
+
+class EnergyAxis(GammapyBaseModel):
+    min: EnergyType = EnergyType("0.1 TeV")
```
cdeil: Suggest to put `min: EnergyType = "0.1 TeV"`
The type casting should work on assignment also here for the defaults, no?

gammapy/analysis/config.py
```diff
@@ -0,0 +1,189 @@
...
+    max: EnergyType = EnergyType("10 TeV")
+    nbins: int = 30
+
+
+class SpatialCircleRange(GammapyBaseModel):
+    frame: FrameEnum = FrameEnum.icrs
+    lon: AngleType = AngleType("83.633 deg")
```
cdeil: Same here and everywhere. Remove the `AngleType` and other manual type casts, and let the framework do the type cast on assignment.

Maybe use a `SkyCoord` here instead of separate lon / lat / frame?

gammapy/analysis/config.py
```diff
@@ -0,0 +1,189 @@
...
+class Data(GammapyBaseModel):
+    datastore: Path = "$GAMMAPY_DATA/hess-dl3-dr1/"
+    obs_ids: List[int] = []
+    obs_file: FilePath = None
+    obs_cone: SpatialCircleRange = SpatialCircleRange()
+    obs_time: TimeRange = TimeRange()
+
+    @validator("datastore")
+    def datastore_exists(cls, v):
+        return make_path(v)
```
cdeil: Remove this validator?
It doesn't check if the path is really present on disk.
Probably we shouldn't until that path is accessed, because someone might write a script where they generate that path only after the loading the config file?
Generally I don't think creating a config should hit the disk and check stuff there.

gammapy/analysis/config.py
```diff
@@ -0,0 +1,189 @@
...
+from astropy.units import Quantity
+from gammapy.utils.scripts import make_path, read_yaml
+from pydantic import BaseModel, FilePath, validator
+from pydantic.utils import deep_update
+from pathlib import Path
+from typing import List
+from enum import Enum
+import yaml
+
+__all__ = ["Config", "General"]
```
cdeil: Remove "General" here? why expose it?

Maybe put "AnalysisConfig" instead of "Config"?
We might have other configs in Gammapy some day (e.g. for simulation)?
I'm 50:50 on this.

Bultako: Done.

gammapy/analysis/config.py
Bultako: Yes.
Note we also have some `FilePath` types that checks the existence of the file. 
https://pydantic-docs.helpmanual.io/usage/types/#pydantic-types

It is not the case for the datastore folder. Anyway, the functions should not have that name since it does not check existence of datastore folder, it just expands the env vars on setting (not reading), which I guess it should not be the task of the model.

gammapy/analysis/config.py
Bultako: It seems that the type casting does not work on assignment, just on setting.
I think we need to keep this forced castings.

gammapy/analysis/config.py
Bultako: > Maybe use a SkyCoord here instead of separate lon / lat / frame?


I do not see how to use `SkyCoord` if we want to keep the syntax below.
```
obs_cone: {frame: galactic, lon: 83.633 deg, lat: 22.014 deg, radius: 3 deg}
```

gammapy/analysis/config.py
Bultako: Ok.
This means that ckecking that `skydir` has been defined should be done elsewhere, no?

gammapy/analysis/config.py
Bultako: Done.

gammapy/analysis/config.py
Bultako: Ok for the tests.

Re: `return Angle(v)`

When setting the values of an angle (i.e. `config.datasets.geom.wcs.binsize="0.1 deg"`), the type set may be `Angle` or `AngleType` depending on what we decide to return in the validation function. 

Since it seems that type casting does not work on assignment (i.e. default values declared), and it only works on setting, we need to do force casting on the definition of the model. If we do force casting to `AngleType` I would also leave `AngleType` for settings.

Bultako:
@cdeil 
Thanks for the review.
I've addressed most of the comments and leave some replies for those not fixed.

cdeil:
@Bultako - I've done some updates in 90ab0c417 .

I set `validate_all=True`, which means the defaults are now validated and type cast as well. (See https://pydantic-docs.helpmanual.io/usage/model_config/). I think that's what we want?

I also changed the types so that native Astropy types like Quantity and Angle are stored as attributes. I think we want users to see those, and not be confused by pydantic config validation subclasses, no?

@Bultako - OK?

gammapy/analysis/config.py
cdeil: How about this?
```
class SpatialCircleRange(GammapyBaseModel):
    center: SkyCoordType = None
    radius: AngleType = None
```
which would match https://astropy-regions.readthedocs.io/en/latest/api/regions.CircleSkyRegion.html

cdeil: How about this?
```
class SpatialCircleRange(GammapyBaseModel):
    center: SkyCoordType = None
    radius: AngleType = None
```
which would match https://astropy-regions.readthedocs.io/en/latest/api/regions.CircleSkyRegion.html

Bultako:
Thanks !
It seems fine 👍

Bultako:
I agree, it is a way to always have an up-to-date sample config file that exposes the present syntax and options for the config parameters. 

I can add it tomorrow soon in this PR.

Bultako:
@cdeil 
I think we already make a check/validation of the model against a reference config file: `config.yaml` When we test with `test_config_create_from_yaml` function, the validation of the model against the file is done at that moment, hence we can detect any possible problem in the structure, names and values of the parameters in the config file. Is this what you wanted or I misunderstood the idea?

cdeil:
I think reading that config file tests a lot, but not the defaults used in `config.py`. So making a config from the defaults and comparing it against a reference as an additional test would test more. But yes, it's not a must-have. OK to merge as-is and move on to switch the new for the old.

Issue #2504, opened 2019-11-01
Problem of defining a SkyModel for simple 1D analysis from the high level interface.


JouvinLea:
Hi all,
As I mentioned to Regis, I tried the high level interface for all of my Gammapy analysis.
I am for the moment only doing 1D spectral analysis.
When I try to adapt the notebook of doing a 3D analysis with the high level interface to perform a very basic 1D analysis, It was really strange for me to have to define a spatial component for the model.... I had to go in the test to understand that including for 1D analysis, we have to define a spatial component since in the high level interface it is using a SkyModel for the analysis.model.

Honestly, I don't think it is a good idea and I guess that a lot of standard user would find it also strange to define a spatial component for 1D analysis. Furthermore, when we have our fit performed and we want to inspect a bit the model from the analysis object, we have to write that to access the spectral part `analysis.model.skymodels[0].spectral_model` whereas for 1D fitting I am only interested in this component.

What do you think?

cdeil:
There's pros and cons, but overall I think having a 3D model for 1D spectral analysis is a good thing.
For Galactic sources it's very common that there's overlap with emission from other components, and that the source of interest is an extended source.
As far as I can see, having a 3D model is the only sane way to keep track of contamination and containment fractions in the 1D region-based spectral analysis.

So basically, I think the power for analyses and the simplicity to have one standard way is the way to go, instead of to invent a special limited second way where only a spectral model is needed. Note that very soon we'll add time-dependent model component, which is then needed for AGN or binaries or pulsars. Again: the general model class will have that optional component anyways, and re-implementing time-dependence on just the spectral model objects will be painful for us, and possibly also more complex for users to learn a second way to compose models. And yet another thing: we'll want to account for exposure and background variations across an on region eventually, and for that a spatial model component is needed.

I think there's some things we can do to make this easier for users: Number 1 is documentation, just explain how it works and how to get the results for 1D analysis. Then maybe add some conveniences, such as e.g. we could consider to auto-add a point source model at the center of the region, if a users starts a 1D spectral analysis without a spatial model set.

Adding "high-priority" label, v0.15, and @adonath - assigning to you.

cdeil:
[...]

PR #2606, opened 2019-11-25
Use SkyModel in spectral datasets follow up

adonath:
**Description**

This pull request is a follow up PR to #2595 and addresses #2504. It implements the following changes:
- Change default of `spatial_model` in the `SkyModel` class to a `ConstantFluxSpatialModel`, that is introduced for this purpose. I considered this solution the "smaller evil", because the alternative solution would be to introduce `if spatial_model is None` in ~10 places (incl. `MapEvaluator` and the model evaluation functions). Also properties like `SkyModel.frame` and `SkyModel.evaluation_radius` would fail on access. The `ConstantFluxSpatialModel` does not have any parameters and is serialised as following:
```
- name: source
  type: SkyModel
  spatial:
    type: ConstantFluxSpatialModel
  spectral:
    type: PowerLawSpectralModel
```
I realise this not as convenient for users, on the other hand it is an accurate description of what happens during the model evaluation, so I actually prefer it as a more explicit solution. It also opens up the way to use spatial models in future for containment correction...

- Unify the model setters in `SpectrumDataset` and `MapDataset`.
- I decided to move the  the loop over model components out of the `SpectrumEvaluator` this opens up the way to re-use the `MapEvaluator` for evaluation of spectral models as well in future.
- Add unit test to evaluate multiple model components in `SpectrumDataset`

**Dear reviewer**
Please give your opinion on the solution with the `ConstantFluxSpatialModel`.

### DIFF
--- a/gammapy/cube/fit.py
+++ b/gammapy/cube/fit.py
@@ -87,8 +87,6 @@ class MapDataset(Dataset):
         mask_safe=None,
         gti=None,
     ):
-        if model is None:
-            model = SkyModels([])
         if mask_fit is not None and mask_fit.data.dtype != np.dtype("bool"):
             raise ValueError("mask data must have dtype bool")
         if mask_safe is not None and mask_safe.data.dtype != np.dtype("bool"):
@@ -213,22 +211,24 @@ class MapDataset(Dataset):
 
     @model.setter
     def model(self, model):
-        if isinstance(model, SkyModel):
+        if model is None:
+            model = SkyModels([])
+
+        elif isinstance(model, SkyModel):
             model = SkyModels([model])
 
         self._model = model
 
-        if model is not None:
-            evaluators = []
+        evaluators = []
 
-            for component in model:
-                evaluator = MapEvaluator(
-                    component, evaluation_mode=self.evaluation_mode
-                )
-                evaluator.update(self.exposure, self.psf, self.edisp, self._geom)
-                evaluators.append(evaluator)
+        for component in model:
+            evaluator = MapEvaluator(
+                component, evaluation_mode=self.evaluation_mode
+            )
+            evaluator.update(self.exposure, self.psf, self.edisp, self._geom)
+            evaluators.append(evaluator)
 
-            self._evaluators = evaluators
+        self._evaluators = evaluators
 
     @property
     def parameters(self):
@@ -268,16 +268,15 @@ class MapDataset(Dataset):
         if self.background_model:
             npred_total += self.background_model.evaluate()
 
-        if self.model:
-            for evaluator in self._evaluators:
-                # if the model component drifts out of its support the evaluator has
-                # has to be updated
-                if evaluator.needs_update:
-                    evaluator.update(self.exposure, self.psf, self.edisp, self._geom)
+        for evaluator in self._evaluators:
+            # if the model component drifts out of its support the evaluator has
+            # has to be updated
+            if evaluator.needs_update:
+                evaluator.update(self.exposure, self.psf, self.edisp, self._geom)
 
-                if evaluator.contributes:
-                    npred = evaluator.compute_npred()
-                    npred_total.stack(npred)
+            if evaluator.contributes:
+                npred = evaluator.compute_npred()
+                npred_total.stack(npred)
 
         return npred_total
 
--- a/gammapy/cube/tests/test_fit.py
+++ b/gammapy/cube/tests/test_fit.py
@@ -542,7 +542,7 @@ def test_map_dataset_onoff_fits_io(images, tmp_path):
     dataset.write(tmp_path / "test.fits")
 
     dataset_new = MapDatasetOnOff.read(tmp_path / "test.fits")
-    assert dataset_new.model is None
+    assert len(dataset_new.model) == 0
     assert dataset_new.mask.dtype == bool
 
     assert_allclose(dataset.counts.data, dataset_new.counts.data)
--- a/gammapy/modeling/models/__init__.py
+++ b/gammapy/modeling/models/__init__.py
@@ -21,6 +21,7 @@ class Registry(list):
 SPATIAL_MODELS = Registry(
     [
         ConstantSpatialModel,
+        ConstantFluxSpatialModel,
         TemplateSpatialModel,
         DiskSpatialModel,
         GaussianSpatialModel,
--- a/gammapy/modeling/models/cube.py
+++ b/gammapy/modeling/models/cube.py
@@ -7,6 +7,8 @@ import astropy.units as u
 from gammapy.maps import Map
 from gammapy.modeling import Model, Parameter, Parameters
 from gammapy.utils.scripts import make_path, read_yaml, write_yaml
+from .spatial import ConstantFluxSpatialModel
+from .spectral import PowerLawSpectralModel
 
 
 class SkyModelBase(Model):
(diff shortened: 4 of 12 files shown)


cdeil:
I'm -1 on introducing a "dummy" spatial model (and spectral and temporal). Suggest to use `None` if there is none. There shouldn't be 10 places where the check is needed if there is a spatial model, only basically two: when writing or evaluating a list of models, the one without a spatial model should be skipped. Re-exposing `frame` and `evaluation_radius` on `SkyModel` probably isn't a good idea in the first place, suggest to remove it and access `SpatialModel.frame` directly.

This is an advice. If you're convinced that this dummy model is useful, OK. For uniformity it should have unit `sr-1` like all other spatial models, no?

registerrier:
I have the same opinion as @cdeil . If there is no `SpatialModel`, the content should be set to `None`. A few `if ... else` statement seems acceptable for that.

adonath:
Thanks @cdeil and @registerrier, I'll change to what you propose...

adonath:
@cdeil and @registerrier I've thought about it once more and I'm still convinced my proposed solution is a good one. The `ConstantFluxSpatialModel` is not a dummy model, but it actually behaves exactly how I would expect it: especially it integrates to unity in the evaluation region, like all other spatial models we have:

```python
from gammapy.modeling.models import ConstantFluxSpatialModel, GaussianSpatialModel, ConstantSpatialModel

m = Map.create(npix=100)

gauss = GaussianSpatialModel()
const_flux = ConstantFluxSpatialModel()
const = ConstantSpatialModel()

print((gauss.evaluate_geom(m.geom) * m.geom.solid_angle()).sum())
print((const_flux.evaluate_geom(m.geom) * m.geom.solid_angle()).sum())
print((const.evaluate_geom(m.geom) * m.geom.solid_angle()).sum())
```
Which returns:
```
0.99999847462073
1.0000000000000002
0.030423085697401923
```
When used with a spectral model in a `SkyModel` the amplitude of the spectral model represents the spatially integrated flux, like for all other spatial models.

Our current `ConstantSpatialModel` deviates from this. It represents a constant surface brightness, which is of course not the same thing and does not integrate to unity in the evaluation region. So I'd even propose to change the definition of our current `ConstantSpatialModel` to be in line with our other spatial models and rename the current `ConstantModelSpatialModel` to `ConstantSurfaceBrightnessSpatialModel`, to make clear, that it is different. Depending on the use-case and quantity users are interested in both models are useful. Note that ctools also offers this option http://cta.irap.omp.eu/ctools/users/user_manual/getting_started/models.html#diffuse-source, where the options `DiffuseIsotropic` and `ConstantValue` are available.

Alternatively we can handle the two different cases with a `normalize` flag, as we do for the `TemplateSpatialModel`.

cdeil:
@adonath - I think you're overestimating the value, and underestimating the cost of introducing this second and special model. I think `None` and point source serve the same use cases nicely.

The cost is that you have to explain this special model to Gammapy developers and users. I don't think you can keep it an implementation detail, it will show up in printout etc. Probably you'll need the `if` anyways in printout and serialisation. You're already adding a dummy method here with NotImplementedError for the new model, a sign that it doesn't fit the current scheme nicely.

> ctools also offers this option

Are you sure? I think they just have an alias or second model for legacy compatibility reasons with the Fermi ST?

> It also opens up the way to use spatial models in future for containment correction

Are you sure? Does it really offer benefits over `None` and the models we already have (point and constant)?

> Alternatively we can handle the two different cases with a normalize flag

If really needed, then yes, that would be preferable over a new second constant model.

Can't we use the existing point source model for 1D spectral analysis? That's already pre-integrated like the new model you want to add, no?

adonath:
Thanks @cdeil! There are mainly two things, that I don't like about the current status of the code, that I have tried to solve with my proposal. As usually I don't care as much about the actual implementation, then I care about consistency:
 
- We currently handle a 2D analysis as a 3D analysis with one spectral bin. So for a 2D analysis users have to define a spectral model as well. I think we should do exactly the same for a 1D analysis: handle it as a 3D analysis with one spatial bin and make users define a spatial model as well. 
- There is an inconsistency in the definition of the `ConstantSpatialModel` compared to our other spatial models. First it is the only spatial model, that has an additional `norm` parameter, while all the other spatial models carry the norm in the spectral model. Second it is normalised to the "peak" value instead of the integral. 

Maybe I was trying to solve to many problems at once, so in order to proceed I would propose the following:

- As we already got the feedback (e.g. #2504) that it's weird to have to define a spatial model for the 1D analysis as well, I would propose to make only the `spatial_model` optional on the `SkyModel`. The convention we get is that the flux parameter is always carried by the spectral model, independent of the kind of analysis users do. So they always have to define at least the spectral model for 1D, 2D as well as 3D analyses.
- In a follow up PR I'll remove the `value` parameter from the `ConstantSpatialModel` model and change the normalisation that it integrates to unity on the sphere (i.e. return 1 / (4 pi)). This makes it consistent with all other spatial models.

Does this compromise sound reasonable to you @cdeil and @registerrier?

cdeil:
@adonath - OK as direction, but there's still some key details that need to be agreed. Could you send a PR with a docs section and we finalise the design and discuss there, before doing extensive code changes like in this PR?

I mean something like [here](http://cta.irap.omp.eu/ctools/users/user_manual/getting_started/models.html#implementation) and [here](http://cta.irap.omp.eu/ctools/users/user_manual/getting_started/models.html#spectral-model-components) that explains the general concept, and the exceptions needed.

Can / should we avoid that these are special cases?
- [ConstantSpatialModel](https://docs.gammapy.org/0.14/api/gammapy.modeling.models.ConstantSpatialModel.html)
- [SkyDiffuseCube](https://docs.gammapy.org/0.14/api/gammapy.modeling.models.SkyDiffuseCube.html)
- [PointSpatialModel](https://docs.gammapy.org/0.14/api/gammapy.modeling.models.PointSpatialModel.html)

Partly previously, and now in this PR, we've tried hard to avoid the `if` statement in model evaluation for these cases. But I think even if for point model we have achieved that for now, we already know that for efficient analysis of point sources we will want to special-case their evaluation? And if we have the `if` already in the model evalution, then does it matter much any more if there's a second or third model that's special?

I don't have a full proposal concerning special-casing of these models, or whether they should have a norm or not, and whether or not their unit-handling and evaluation should be special. But we can't keep this vague if we want to finish up and get something good, we need to have the details written down - like in the example [here](https://github.com/gammapy/gammapy/issues/2517#issuecomment-556993333), but preferably directly in the docs to go fast.

adonath:
After private discussion with @cdeil, I decided to close this PR and break down the changes into smaller steps:
- Remove the current default for the spectral model from `SkyModel` and require users to always define at least the spectral model
- Allow the spatial model to be `None` in the `SkyModel`
- Add an optional temporal model to `SkyModel`, which is needed for event sampling, that @fabiopintore is working on
- Add a short paragraph  in `docs/modeling/index.rst`

cdeil:
@adonath - Sounds good. Thank you!

For me, the key is to remove the defaults here:

https://github.com/gammapy/gammapy/blob/360f5237a11c6e73ad5dcbf04111cbecd83d68ee/gammapy/modeling/models/cube.py#L122-L127

and to adapt callers to pass in sensible models if / as needed.

> Allow the spatial model to be None in the SkyModel

I guess this could mean an hour or a day of work.
Note that currently in 1D spec analysis we basically don't support spatial models and the assumption is point source or full containment, with a flag offered whether to apply PSF correction.
So there the solution might be to create a `PointSpatialModel` as default for now, and then full support for spatial models in 1D region-based analysis comes next year.

PR #2608, opened 2019-11-26
Add gammapy download datasets -- tests flag

Bultako:
This PR adds an optional `--tests` flag to `gammapy download datasets` command line tool, so to include datasets needed for development tests in the download only when the flag is set.

With this PR `gammapy download datasets` will download 433 files/ 172MB whilst `gammapy download datasets --tests` will download 503 files/ 177MB


Bultako:
Oops.. Sure !
What about also adding the `tests`flag here?

https://github.com/gammapy/gammapy/blob/f03e5820778d3dd3d8c5b24574090e7aeb44aee7/.travis.yml#L146

https://github.com/gammapy/gammapy/blob/afa87b4f01fb1d0900c28d1421e845f5f5dead43/azure-pipelines.yml#L43

https://github.com/gammapy/gammapy/blob/afa87b4f01fb1d0900c28d1421e845f5f5dead43/azure-pipelines.yml#L118

cdeil:
Yes, I think you have to add the tests flag to the CI builds, or they will fail.

PR #2609, opened 2019-11-26
Improve CTA with Gammapy tutorial 

cdeil:
This PR is an update (almost rewrite) of the "CTA with Gammapy" tutorial.

- [new](https://nbviewer.jupyter.org/github/cdeil/gammapy/blob/019880500212f8d82385d8a487f2cdf9b8a10d8a/tutorials/cta.ipynb)
- [old](https://nbviewer.jupyter.org/github/gammapy/gammapy/blob/7a04bb567858042be2955d05847a5816e3db540e/tutorials/cta.ipynb)

As discussed previously, this is the entry point for CTA users for Gammapy, and it's linked to from here: https://docs.gammapy.org/dev/tutorials/index.html#notebooks

It's not in v1.0 quality yet, but IMO a big step in that direction. I would prefer to merge now, and then to work on other things, and then to come back to this. Generally I think merging docs PRs in very quickly, and then doing short presentations & feedback for each heavily updated notebook based on an updated https://docs.gammapy.org/dev would be a good way to go fast and to deliver as good as possible docs quickly.

I would suggest to delete https://gammapy.org/cta.html now, since that doesn't contain up to date or useful information any more, that information is now at the start of this notebook.

@registerrier - Thoughts?


cdeil:
Merging now. We can discuss in the next Gammapy docs call how to improve further.

registerrier:
Thanks @cdeil.  I did not have time to read it fully before you merged.
- I would remove the get in touch part. This should appear in the tutorials.html page but not on each notebook.
- I think the setup section could be moved after files (since all the other commands are shell commands and the gammapy stuff starts later).
- I am not sure we need commented sections in the notebooks. This can be moved elsewhere, no? (e.g. how-to etc)
   - Is it the best place to discuss XML files? We could have a specific how-to for this.
   - The CTA performance files section could be better placed in simulation or sensitivity, no?

There remain a few typos.

cdeil:
> I would remove the get in touch part.

OK to remove the one sentence linking to https://gammapy.org/contact.html or moving it to the tutorials page if you prefer.

More important question: Keep & update, or remove https://gammapy.org/cta.html ? (what about HESS, Fermi, MAGIC, ...?)

> I think the setup section could be moved after files

Not sure if this will work. We should post updated dev docs and look at the navbar outline and then discuss. In the Friday docs call?

> discuss XML files?

I think we should show an example. Agreed, since Fermi-LAT also has XML files, it could go somewhere else. But this is not 1-2, but 10-20 lines of code to show how to do this. Too long for a HOWTO entry? Better in some tutorial notebook?

> The CTA performance files section could be better placed in simulation or sensitivity, no?

I think giving the info where to get FITS CTA responses is the most important part of this "Gammapy for CTA" users.
If we move it to some simulation or sensitivity notebooks, discoverability will suffer, and then how would you avoid duplicating this information a few times?

> There remain a few typos.

Can you please simply fix typos as you read here and everywhere and commit directly to master? This is more efficient than review comments.

registerrier:
> More important question: Keep & update, or remove https://gammapy.org/cta.html ? (what about HESS, Fermi, MAGIC, ...?)

Well, it is not up to date for sure.
There is a bit of uncertainty on what belongs to general page and what belongs to the documentation. There is a similar concern regarding the overview page  for instance. 
The page is there to connect to CTA for political reason, so maybe keep ti here but update the content.

cdeil:
I removed the contact link, fixed a heading level, and resolved a TODO concerning the issues at FOV offset 0 deg by adding some text in a42d5bb28

Issue #2613, opened 2019-11-27
Selecting run with a time range


bkhelifi:
**Gammapy version**
The dev version

**Bug description**
Using the HESS data release, the selection of observations using the table selection function via the criteria 'time_box' crashes. This is linked to the management of the absolute times or relative ones (as defined in the gadf). Both should be accepted somehow.

**Expected behavior**
Whatever the stored time format (relative or absolute), our time selection should work. Two ways are proposed, either via the data_store (using by the LC) or via the obs_table (for any purpose). I propose to keep for the moment these ways (knowing also that their behaviours are different), taking care that both are working

**To Reproduce**
data_store = DataStore.from_dir("$GAMMAPY_DATA/hess-dl3-dr1")
time_selection = dict(type='time_box', time_range=Time(['2006-07-29T00:00:01', '2006-07-30T13:00:00']))
selected_obs_table = data_store.obs_table.select_observations(time_selection)

cdeil:
Complete reproducer:
```
from astropy.time import Time
from gammapy.data import DataStore

data_store = DataStore.from_dir("$GAMMAPY_DATA/hess-dl3-dr1")
time_selection = dict(type='time_box', time_range=Time(['2006-07-29T00:00:01', '2006-07-30T13:00:00']))
selected_obs_table = data_store.obs_table.select_observations(time_selection)
```


Output:
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/deil/work/code/gammapy/gammapy/data/obs_table.py", line 274, in select_observations
    "TSTART", selection["time_range"], selection["inverted"]
  File "/Users/deil/work/code/gammapy/gammapy/data/obs_table.py", line 163, in select_time_range
    if self.meta["TIME_FORMAT"] == "absolute":
KeyError: 'TIME_FORMAT'
```

Somewhat related: #1172

Since many years the spec says that the time ref header keys are required in obs table:
https://gamma-astro-data-formats.readthedocs.io/en/latest/data_storage/obs_index/index.html#mandatory-header-keywords

It doesn't mention a `TIME_FORMAT` key, so `if self.meta["TIME_FORMAT"] == "absolute"` is no good.

https://github.com/gammapy/gammapy/blob/360f5237a11c6e73ad5dcbf04111cbecd83d68ee/gammapy/data/obs_table.py#L163

[...]

PR #2618, opened 2019-11-27
Fix ObservationTable select time

registerrier:
**Description**

This pull request corrects an improper behavior in `ObservationTable.select_time` reported in #2613.
This PR modifies the behavior of `ObservationTable.select_time` introducing `time_start` and `time_stop` properties. This creates a behavior closer to `EventList.select_time`.
Here we check for overlap, since an observation can be fully or partly enclosed in the time bin and we leave the option to the user to select one case or the other. 

The test no longer use a table with absolute times stored. `ObservationTable` now assumes only a table with time reference and elapse time as all other time tables in gammapy. 

**Dear reviewer**

The PR is ready for review. The new code is tested.

### DIFF
--- a/gammapy/data/tests/test_obs_table.py
+++ b/gammapy/data/tests/test_obs_table.py
@@ -259,7 +259,7 @@ def test_select_time_box():
     obs_table_time = make_test_observation_table(
         n_obs=10,
         date_range=(datestart, dateend),
-        use_abs_time=True,
+        use_abs_time=False,
         random_state=random_state,
     )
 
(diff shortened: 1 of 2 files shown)


gammapy/data/tests/test_obs_table.py
```diff
@@ -259,18 +259,26 @@ def test_select_time_box():
...
     value_range = Time(["2012-01-01T01:00:00", "2012-01-01T02:00:00"])
     selection = dict(type="time_box", time_range=value_range)
     selected_obs_table = obs_table_time.select_observations(selection)
-    time_start = selected_obs_table["TSTART"]
-    assert (value_range[0] < time_start).all()
-    assert (time_start < value_range[1]).all()
+    time_start = selected_obs_table.time_start
+    time_stop = selected_obs_table.time_stop
 
+    assert (value_range[0] < time_start).all() & (time_stop < value_range[1]).all()
```
cdeil: Typical case of re-implementing the method in the test. Should be changed to a simple small test case and direct assert on output. But OK, that's old baggage that was already here.

gammapy/data/tests/test_obs_table.py
```diff
@@ -259,18 +259,26 @@ def test_select_time_box():
...
+
+    # Test selection with partial overlap of observations and value_range
+    selection_partial = dict(type="time_box", time_range=value_range, partial_overlap=True)
+    selected_obs_table = obs_table_time.select_observations(selection_partial)
+    time_start = selected_obs_table.time_start
+    time_stop = selected_obs_table.time_stop
+
+    assert (value_range[1] > time_start).all() & (time_stop > value_range[0]).all()
```
cdeil: Another copy of the time select implementation code in the test. Sigh.

cdeil:
I changed the title to "Fix ObservationTable select time". IMO it's nice to have most PRs start with "Fix", "Add", "Remove" or "Improve" - then PR titles in the changelog can be parsed quickly, and users can see where bugs were fixed.

gammapy/data/tests/test_obs_table.py
registerrier: Yes that is not very good. But with random values it is difficult to do differently. Indeed the tests should be completely refactored.

registerrier:
Having a hard-coded selection on `TSTART` as before is not very nice. So the fully contained observation seems a more expected behavior.
I have introduced `partial_overlap` also because the `inverted` option is meaningful if you have this option.  

I propose to merge for now, at least to solve #2613. All tests should be refactored at some point and this code clarified. I open an issue for that. To be sorted out hopefully for 1.0.

PR #2619, opened 2019-11-27
Unify model setters of all datasets

adonath:
**Description**

This pull request is another follow-up to #2505. It unifies the model setters of `SpectrumDataset`, `FluxPointDataset` and `MapDataset` and handles the case the model is None consistently on I/O.


cdeil:
Rename to dataset.models to make it clear that it’s a Models container object?
Use empty Models to represent the case of no models?
I think then you can remove the if None check in front of some for loops.

adonath:
I'd prefer to make any additional changes in follow-up PRs and merge this one...

PR #2621, opened 2019-11-28
Rewrite AnalysisConfig and update Analysis

Bultako:
**Description**

This pull request is a follow-up of https://github.com/gammapy/gammapy/pull/2605. It provides a complete reimplementation of the high-level interface using the recently added framework for settings handling based on `pydantic` 

- `Analysis` and `AnalysisConfig` classes are now in separated files.
- The previous `jsonschema` framework dependencies have been removed
- The several template YAML files have been modified following syntax agreed.
- The code in `Analysis` class and tests has been highly modified refactored.
- Instantiation of `AnalysisConfig`  with settings in a YAML-formatted string is now done with  `AnalysisConfig.from_yaml(str)`
- Instantiation of `AnalysisConfig`  with settings in a YAML file is now done with  `AnalysisConfig.from_yaml(filename=file.yaml)`
- No existing methods have been removed, no new methods added.
- Observation filtering using `obbs_ids` in a file and time filtering is not implemented.
- Tutorials have been fixed and documentation slightly modified.


cdeil:
@Bultako - This looks great!

I'll have a closer look and try it out in the next hour.

The `test_config_create_from_yaml` if failing. Can you reproduce / fix? Or should I have a look?

https://dev.azure.com/gammapy/gammapy/_build/results?buildId=2641&view=logs&jobId=d7d66435-5ae4-561a-5876-2cd4e19b3ce2&taskId=922dfe32-5b90-53bc-18db-0fa0ecbbd97b&lineStart=142&lineEnd=143&colStart=1&colEnd=1

gammapy/analysis/core.py
```diff
@@ -293,262 +285,110 @@ def _extract_irf_kernels(self, dataset):
...
+    def _make_energy_axis(axis):
+        """Helper method to build an energy axis."""
+        axis_params = dict(name="energy",
+                           lo_bnd=axis.min.value,
+                           hi_bnd=axis.max.value,
+                           nbin=axis.nbins,
+                           unit=axis.min.unit,
+                           interp="log",
+                           node_type="edges")
+        energy_axis = MapAxis.from_bounds(**axis_params)
```
cdeil: Suggest to rewrite like this, since the local `axis_params` and `energy_axis` variables provide no extra value:
```
return MapAxis.from_bounds(
    name="energy",
    ....
)
```

gammapy/analysis/tests/test_analysis.py
```diff
@@ -175,8 +134,8 @@ def test_analysis_1d_stacked():
     assert_allclose(analysis.datasets["stacked"].counts.data.sum(), 404)
     pars = analysis.fit_result.parameters
 
-    assert_allclose(pars["index"].value, 2.689559, rtol=1e-3)
-    assert_allclose(pars["amplitude"].value, 2.81629e-11, rtol=1e-3)
+    assert_allclose(pars["index"].value, 2.676283, rtol=1e-3)
+    assert_allclose(pars["amplitude"].value, 4.554215e-11, rtol=1e-3)
```
cdeil: Why did the flux amplitude increase by a factor of 2 here?
Ideally this PR would just change to the new code, but keep the existing analysis tests the same.

gammapy/analysis/core.py
```diff
@@ -204,55 +165,88 @@ def get_flux_points(self, source="source"):
...
         fp.table["is_ul"] = fp.table["ts"] < 4
-        model = self.model[source].copy()
-        self.flux_points = FluxPointsDataset(data=fp, model=model)
+        self.flux_points = FluxPointsDataset(data=fp, model=self.model[source])
         cols = ["e_ref", "ref_flux", "dnde", "dnde_ul", "dnde_err", "is_ul"]
         log.info("\n{}".format(self.flux_points.data.table[cols]))
 
-    @staticmethod
-    def _create_geometry(params):
+    def update_config(self, config=None, filename=""):
```
cdeil: Can we move `analysis.update_config` to `analysis.config.update`?

If you look at the current `analysis.update_config` you'll see that it only accesses `self.config`, nothing else from `self`, i.e. it only interacts with the config.

Or is there a technical problem with this suggestion, e.g. because it's not clear how to do an in-place update of the values? Then maybe the method could still be on the config class, and the analysis class only has a wrapper like this?
```
class Analysis:
    def update_config(self, config):
        self.config = self.config.update(config)
```

gammapy/analysis/core.py
Bultako: Done

cdeil:
@Bultako - This is almost ready to be merged.

Could you please run this
```
pytest -v gammapy/analysis --cov gammapy.analysis --cov-report html
open htmlcov/index.html
```
and add 1-2 tests, or delete code that isn't used?

Specifically add 1-2 lines in some test to check the energy validation, and change an existing test to have an exclusion mask and check that this works.

The preference should always be to be as simple as possible and have as little code as possible, so if you're struggling to come up with a test case for something, consider just deleting that option and code. (I think we can still simplify, but probably we should just merge this in in ~ 1 hour, and then continue in future PRs).

cdeil:
@Bultako - Can you please run this?
```
make polish
git add gammapy/analysis
git commit -m 'make polish in gammapy/analysis
git checkout -- .
```
This runs `isort` and `black` formatting, just on the code you changed, and then discards other formatting changes, that would possibly conflict with other PRs.

@adonath @registerrier @Bultako - I think generally this should be the procedure: we try to keep Gammapy master clean so that people can just run `make polish` before committing. This requires maintainers to run it before merging in PRs for contributors that don't do it. I think that's fine, usually there's anyways other small edits one has to do, e.g. to fix docstring formatting, clean up variable names, ... And if that's not the case, like in the past weeks most of the time, then at least we can do it for code we're touching using the commands above.

docs/analysis/index.rst
```diff
@@ -138,35 +134,23 @@ The observations are stored as a list of `~gammapy.data.DataStoreObservation` ob
...
 
-The data reduction process needs a choice of a dataset type, declared as the class name
-(`~gammapy.cube.MapDataset`, `~gammapy.spectrum.SpectrumDatasetOnOff`) in the ``reduction`` section of the settings. For the
-estimation of the background with a dataset type `~gammapy.spectrum.SpectrumDatasetOnOff`, a ``background_estimator``
-is needed, other parameters related with the ``on_region`` and ``exclusion_mask`` FITS file
-may be also present. Parameters for geometry are also needed and declared in this section,
-as well as a boolean flag ``stack-datasets``.
+The data reduction process needs a choice of a dataset type, declared as ``1d`` or ``3d``
+in the ``datasets`` section of the settings. For the estimation of the background in a ``1d``
+use case, a background ``method`` is needed, other parameters related like the ``onregion``
```
cdeil: Suggest to rename `onregion` -> `on_region` in the config file.
We already have `analysis.background_estimator.on_region`, and we use `snake_case` almost consistently in the config file (except `binsize`, but that word we'll probably re-discuss anyways).

gammapy/analysis/config/config.yaml
```diff
@@ -36,7 +36,7 @@ datasets:
         method: reflected
         exclusion:    # fits file for exclusion mask
     # mandatory for 1d
```
cdeil: I suggest to remove all these "mandatory for 1d" and "mandatory for 3d" comments.
Two reasons:
- It's not clear what they mean. Currently we have defaults, so it's not mandatory to set them in user config files, saying those are "mandatory" isn't true. I guess you mean "used for 1d" and "used for 3d"?
- We will very likely completely restructure this 1D / 3D config options in the next days, and very quickly any comments saying what is and isn't used for 1D / 3D will turn into lies. We might add back these comments once the dust has settled, but more likely we'll try to come up with a schema that is clear by itself, so that we don't have to rely on comments to communicate this information to the users.

gammapy/analysis/config/template-1d.yaml
```diff
@@ -1,46 +1,22 @@
 ---
```
cdeil: @Bultako - Can you please remove this `---` at the top of each YAML file.

I think it's optional, if omitted the result is the same, it works.

People will forget to put this all the time, I don't see the point of putting it or teaching users about it, so would prefer to just not use it in our YAML files and examples.

docs/analysis/index.rst
Bultako: Done.

gammapy/analysis/config/config.yaml
Bultako: Done.

gammapy/analysis/config/template-1d.yaml
Bultako: Done.

gammapy/analysis/core.py
```diff
@@ -56,7 +44,7 @@ class Analysis:
 
     def __init__(self, config=None):
         if isinstance(config, dict):
-            self._config = AnalysisConfig(config)
+            self._config = AnalysisConfig(**config)
         elif isinstance(config, AnalysisConfig):
             self._config = config
         else:
```
cdeil: Change:
```
ValueError: Dict or `AnalysiConfig` object required.
```
To:
```
TypeError: config must be dict or AnalysisConfig
```
lower-case "dict" and fix typo in "AnalysisConfig", and change to a `TypeError`.

Bultako: Done.

gammapy/analysis/config/template-1d.yaml
Bultako: Done.

cdeil:
Not introduced in this PR, but in the last PR that added pyvalidate, but still I think here we could clean it up a bit:

SkyCoordType is not really a type like `AngleType` or `EnergyType`, it's a `pydantic.BaseModel` subclass, whereas the others are not. If we had chosen to represent skycoord by a string with special parsing in the config file, it would have been a type.

So suggest to drop the `Type` from `SkyCoordType`. We can't call it `SkyCoord`, that would be too confusing because that's the Astropy class name. Maybe `SkyCoordModel` or `SkyCoordConfig`? Probalbly we should then also change to `EnergyAxisModel` or `EnergyAxisConfig` and similarly add a postfix to all pydantic model subclasses?

For the tests, could you please change `test_config_basics` to remove the two setters on `config.datasets.geom.wcs.skydir` and `config.data.obs_time.start` and just assert that the default is None?
The setting of those and check what then comes out should go in a second separate test below the test for defaults.
To be more easy to maintain, the default checks should be grouped like in the model, i.e. this together:
```
    assert isinstance(config.fit.fit_range, EnergyRange)
    assert isinstance(config.fit.fit_range.min, Quantity)
```

What I don't like is this:
```
from gammapy.analysis.config import (
    AnalysisConfig,
    Axes,
    Background,
    BackgroundMethodEnum,
    Data,
    Datasets,
    EnergyAxis,
    EnergyRange,
    Fit,
    FluxPoints,
    Fov,
    FrameEnum,
    General,
    Geom,
    Log,
    Selection,
    SpatialCircleRange,
    TimeRange,
    Wcs,
)
```
It's so much stuff, and a mix of type and model objects, without a consistent naming scheme that makes it obvious what is a custom data type (https://pydantic-docs.helpmanual.io/usage/types/#custom-data-types) and what is a model (https://pydantic-docs.helpmanual.io/usage/models/).

So sorry for the long comment, I was typing while I was thinking. Bottom like, my suggestion is:
```
- Add "Model" postfix for all models.
- Specifically SkyCoordType should become SkyCoordModel
- Simplify the default config test. I don't think we have to be complete, and test the types of all the models. Just checking that some things come out with the correct type, especially the items where we use a custom type, is enough.
- Move the assignment test to config to a separate test.
```

cdeil:
Suggest to split `from_yaml` and `to_yaml` into two methods.
Add `read` and `write` methods that take a filename.
Keep `from_yaml` and `to_yaml`, but simplify them to only take a YAML `str`.
The caller always knows that they have.

This simplifies the implementation and makes it consistent with the rest of Gammapy.
Mostly we use this pattern, see e.g. https://docs.gammapy.org/0.14/_modules/gammapy/maps/base.html#Map.write (in that case using FITS instead of YAML, but the principle is the same).

gammapy/analysis/tests/test_analysis.py
```diff
@@ -14,127 +14,86 @@
...
 def test_config():
     config = AnalysisConfig()
-    assert config.settings["general"]["logging"]["level"] == "INFO"
-    cfg = {"general": {"outdir": "test"}}
-    config.update_settings(cfg)
-    assert config.settings["general"]["logging"]["level"] == "INFO"
-    assert config.settings["general"]["outdir"] == "test"
-
-    with pytest.raises(ValueError):
+    assert config.general.log.level == "info"
```
cdeil: Here you just make a config and assert on the default log level.
That doesn't belong in `test_analysis.py`, but in `test_config.py`.
Add the assert on log level default there.

I'm not sure if this `test_config` then is still useful.
Probably you should remove the `config=None` default in `Analysis.__init__` since it's never allowed.
So `config` becomes a required argument.
And then you can just delete this test, or pass some invalid type like `Analysis(config="spam")` and check that a TypeError is raised.

cdeil:
Can we now just delete this mini validation framework from `Analysis`?

If you want to keep it: what value does it provide?

Does it somehow lead to nicer or earlier errors?
Can you give an example / use case where this matters?

```
    def _validate_observations_settings(self):
        """Validate settings before proceeding to observations selection."""
        if not self.config.data.datastore:
            log.info("No datastore defined")
            log.info("Observation selection cannot be done.")
            return False
        datastore_path = make_path(self.config.data.datastore)
        if datastore_path.is_file():
            self.datastore = DataStore().from_file(datastore_path)
        elif datastore_path.is_dir():
            self.datastore = DataStore().from_dir(datastore_path)
        else:
            raise FileNotFoundError(f"Datastore {datastore_path} not found.")
        return True

    def _validate_reduction_settings(self):
        """Validate settings before proceeding to data reduction."""
        if not self.observations or len(self.observations) == 0:
            log.info("No observations selected.")
            log.info("Data reduction cannot be done.")
            return False
        return True

    def _validate_set_model_settings(self):
        """Validate settings before proceeding to model attach."""
        if not self.datasets or len(self.datasets) == 0:
            log.info("No datasets reduced.")
            return False
        return True

    def _validate_fitting_settings(self):
        """Validate settings before proceeding to fit."""
        if not self.model:
            log.info("No model attached to datasets.")
            log.info("Fit cannot be done.")
            return False
        return True

    def _validate_fp_settings(self):
        """Validate settings before proceeding to flux points estimation."""
        if not self.fit:
            log.info("No results available from fit.")
            log.info("Flux points calculation cannot be done.")
            return False
        return True
```

cdeil:
Move `Analysis._set_logging()` to `AnalysisConfig.configure_logging()`?
Then logging config is on one class instead of spread across two.

I'm not sure calling `self.config.configure_logging()` in `Analysis.__init__` is a good idea, since it's a global setting, and `logging.basicConfig` can only be called once per Python interpreter session, so we can't have different logging config if we have multiple analyses per Python process.
But that's a tricky issue that we shouldn't try to solve here.
So for now, I suggest you keep the `self.config.configure_logging()` in `Analysis.__init__`, just more the method that interacts with config fields over to `AnalysisConfig` to make it easier to maintain.

gammapy/analysis/tests/test_analysis.py
cdeil: There are other tests like `test_docs_file` or `test_config_to_yaml` that only relate to `AnalysisConfig`, `Analysis` isn't needed. Move those over from `test_config.py` to `test_analysis.py`.

gammapy/analysis/tests/test_analysis.py
```diff
@@ -14,127 +14,86 @@
...
     analysis.get_observations()
-    assert len(analysis.observations) == config_obs["result"]
+    assert len(analysis.observations) == 4
+    # TODO
+    # obs_file
+    # obs_time
 
 
 @pytest.fixture(scope="session")
 def config_analysis_data():
```
cdeil: This `config_analysis_data` is only used in a single test.
Remove the fixture and move it to that test.

Bultako: Done.

gammapy/analysis/tests/test_analysis.py
Bultako: removed `config=None` in `Analysis.__init__`
and left the test as this:

```
def test_basics():
    config = AnalysisConfig()
    assert "AnalysisConfig" in str(config)
        Analysis("spam")
```

gammapy/analysis/tests/test_analysis.py
Bultako: Done.

gammapy/analysis/tests/test_analysis.py
Bultako: moved `test_docs_file` and `test_config_to_yaml` to `test_config.py`

docs/analysis/index.rst
```diff
@@ -235,22 +210,6 @@ is stored in the ``flux_points`` property as a `~gammapy.spectrum.FluxPoints` ob
...
 
-Residuals
----------
-
-For 3D analysis we can compute a residual image to check how good are the models
-for the source and/or the background.
-
-.. code-block:: python
-
-    >>> analysis.datasets.datasets[0].residuals()
```
cdeil: Why remove this example how to get residuals?
OK to remove if you don't think it's good to have at this place.
If it's just because it was broken, I think this would work:
```
analysis.datasets[0].residuals()
analysis.datasets[0].plot_residuals()
```

Bultako: Yes, it was broken when I tested. It works when rebasing.
I've put it back.

gammapy/analysis/config.py
```diff
@@ -66,127 +79,210 @@ class Config:
...
-class Background(GammapyBaseModel):
+class BackgroundConfig(GammapyBaseConfig):
     method: BackgroundMethodEnum = BackgroundMethodEnum.reflected
     exclusion: FilePath = None
 
 
-class Axes(GammapyBaseModel):
-    energy: EnergyAxis = EnergyAxis()
-    energy_true: EnergyAxis = EnergyAxis()
+class AxesConfig(GammapyBaseConfig):
```
cdeil: Suggest to put the more explicit EnergyAxisConfig or EnergyAxesConfig

gammapy/analysis/config.py
```diff
@@ -66,127 +79,210 @@ class Config:
...
-class EnergyAxis(GammapyBaseModel):
+class EnergyAxisConfig(GammapyBaseConfig):
     min: EnergyType = "0.1 TeV"
     max: EnergyType = "10 TeV"
     nbins: int = 30
 
 
-class SpatialCircleRange(GammapyBaseModel):
-    frame: FrameEnum = FrameEnum.icrs
+class SpatialCircleRangeConfig(GammapyBaseConfig):
```
cdeil: Suggest to rename to `SpatialCircleConfig`, IMO a circle isn't a "range".

gammapy/analysis/config.py
```diff
@@ -66,127 +79,212 @@ class Config:
...
+    def write(self, filename, overwrite=False):
+        fname = Path(filename).name
+        fpath = Path(self.general.outdir) / fname
+        if fpath.exists() and not overwrite:
+            raise IOError(f"File {filename} already exists.")
+        fpath.write_text(self.to_yaml())
+        log.info(f"Configuration settings saved into {fpath}")
+
+    def to_yaml(self):
+        return yaml.dump(yaml.safe_load(self.json()), sort_keys=False, indent=4)
```
cdeil: Put `yaml.dump(self.dict(),...)`. Add a docstring saying that YAML `str` is returned.

gammapy/analysis/config.py
```diff
@@ -66,127 +79,212 @@ class Config:
...
+        return info.expandtabs(tabsize=4)
+
+    @classmethod
+    def read(cls, filename):
+        config = read_yaml(filename)
+        return cls.from_yaml(config)
+
+    @classmethod
+    def from_yaml(cls, settings):
+        if isinstance(settings, str):
```
cdeil: Suggest to remove this isinstance check, and to only accept `str`.
Docstring needs to be added explaining what `from_yaml` does.

There's no point in allowing a settings dict here.
If someone has a dict, they should do `config = AnalysisConfig(**settings)`, not `config = AnalysisConfig.from_yaml(settings)`.
So only accept YAML in `from_yaml`.

gammapy/analysis/config.py
```diff
@@ -66,127 +79,212 @@ class Config:
...
+            settings = yaml.safe_load(settings)
+        return AnalysisConfig(**settings)
+
+    def write(self, filename, overwrite=False):
+        fname = Path(filename).name
+        fpath = Path(self.general.outdir) / fname
```
cdeil: Suggest to just write to `filename`, not `outdir/filename`. Keep it simple. outdir is something that Analysis uses to put files.

I'm 60:40 on this choice, if you prefer, OK to keep `outdir/filename`.

But in any case, you have to add a docstring saying what this does (especially if it's `outdir/filename`, as a user I wouldn't expect that).

gammapy/analysis/config.py
```diff
@@ -66,127 +79,212 @@ class Config:
...
+        """Updates config with provided settings.
+
+         Parameters
+         ----------
+         config : string dict or `AnalysisConfig` object
+             Configuration settings provided in dict() syntax.
+         filename : string
+             Filename in YAML format.
+         """
+        try:
```
cdeil: Remove the complex try - except and I/O here.
A user can just write `config.update(yaml.load("config.yaml"))` if they want to read config from YAML and update an existing config, no?

gammapy/analysis/config.py
Bultako: well, `yaml.dump(self.dict(),...)` does not provide a standard dict representation since the types are exposed in `<brackets>` with the values, so `yaml.dump` breaks..

cdeil: You're right. Interesting. I've asked about it here: https://github.com/samuelcolvin/pydantic/issues/1043
For now, suggest to change to `yaml.dump(json.load(self.json())`, and to leave a comment why that is needed, possibly with a link to https://github.com/samuelcolvin/pydantic/issues/1043

gammapy/analysis/tests/test_analysis.py
```diff
@@ -109,6 +112,25 @@ def test_analysis_1d():
     assert_allclose(dnde[-1].value, 5.382879e-21, rtol=1e-2)
 
 
+@requires_data()
+def test_exclusion_region(tmp_path):
+    config = AnalysisConfig.from_template("1d")
+    analysis = Analysis(config)
+
+    skydir = SkyCoord(83, 22, unit="deg", frame="icrs")
+    exclusion_region = CircleSkyRegion(center=SkyCoord(85, 23, unit="deg", frame="icrs"), radius=1 * u.deg)
```
cdeil: Write `SkyCoord("85d 23d")` to be shorter and try to fit on one line (black will break this line otherwise). `frame="icrs"` is the default, no need to specify it.

gammapy/analysis/tests/test_analysis.py
```diff
@@ -109,6 +112,25 @@ def test_analysis_1d():
...
+    exclusion_mask = Map.create(npix=(150, 150), binsz=0.05, skydir=skydir, proj="TAN", coordsys="CEL")
```
cdeil: Rely on defaults and keep it as short as possible. You can omit `coordsys="CEL"`. Probably the default `proj="CAR"` would also be fine?

gammapy/analysis/tests/test_analysis.py
```diff
@@ -109,6 +112,25 @@ def test_analysis_1d():
...
+    mask = exclusion_mask.geom.region_mask([exclusion_region], inside=False)
+    exclusion_mask.data = mask.astype(int)
+    filename = tmp_path / "exclusion.fits"
+    exclusion_mask.write(filename)
+    config.datasets.background.exclusion = str(filename)
```
cdeil: Do you need to put `str(filename)` here. It would be nice if assigning `Path` objects works transparently, converting to `str` on assignment. Does it already?

Bultako: Done

gammapy/analysis/tests/test_analysis.py
Bultako: Done

gammapy/analysis/tests/test_analysis.py
Bultako: Done

gammapy/analysis/config.py
Bultako: Done

gammapy/analysis/config.py
Bultako: Done

gammapy/analysis/config.py
Bultako: Done

gammapy/analysis/config.py
Bultako: Done

gammapy/analysis/config.py
Bultako: Done

Bultako:
@cdeil 
Thanks very much for this review !
I think I've addressed all your comments.
You can make a final polish if you wish or merge :)

cdeil:
@Bultako - I'll fix https://travis-ci.org/gammapy/gammapy/jobs/618348141#L3061 and do some polishing and then will merge this in shortly.

cdeil:
@Bultako - Some fixes and cleanup in a3431c8bf50f52ef1e12eb7cf7623618bfd297ce and fb08628a1f39185a5dd10d38dbaecb0c7b659d90

I'll merge now.

We should discuss & improve at least the following two points tomorrow:
- `AnalysisConfig.update` is broken, for some of the inputs no state is changed and nothing is returned. You didn't notice in the test because you don't check that something changed. We need to discuss what options to offer (fewer, please), and then fix for those inputs.
- Can we simplify `Analysis.set_model`, to have fewer cases? It looks like you return False on invalid inputs, so silently nothing happens. I think we should raise a ValueError instead? Maybe we can remove the I/O and only offer one required input to this method?
- Calling `config.help("spam") does nothing. I think it should raise an error when passing some section that doesn't exist? And generally it's not clear how a user would figure out what sections are available, it's not mentioned in the docstring. Overall I'd like to get rid of help and `_get_doc_sections`, but I guess it's used in the docs build, so not easily possible. Maybe investigate switching to https://sphinx-pydantic.readthedocs.io/ ?

And then there's bigger tasks, like improving the config schema, or reviewing and polishing the code in the Analysis class. But this PR leaves it in a much better state than before, improving that is for future PRs.

@Bultako - Overall, I have to say, I'm super happy with this PR. I think it's a big step forward for usability of Gammapy (config tab completion, validation), and also for maintainability (much clearer separation of responsibilities between Analysis and AnalysisConfig, and I find `config.py` much more readable than the JSON schema, partly because we simplified the schema, partly because I find Python more readable. Muchas gracias, señor!

PR #2601, opened 2019-11-24
Implemented minimal error propagation test for NaimaSpectralModel

luca-giunti:
I implemented a minimal test for error propagation in case of NaimaModel. #2190 might be closed now

PR #2623, opened 2019-11-28
Fix SpectralModel.evaluate_error gradient computation

adonath:
**Description**

This pull request addresses #2601. I've change to the alternative method proposed in the [prototype notebook](https://github.com/gammapy/gammapy-extra/blob/master/experiments/uncertainty_estimation_prototype.ipynb), which allowed to use a separate eps per parameter.

### DIFF
--- a/gammapy/modeling/models/spectral.py
+++ b/gammapy/modeling/models/spectral.py
@@ -4,7 +4,6 @@ import operator
 import numpy as np
 import scipy.optimize
 import scipy.special
-from scipy.optimize.slsqp import approx_jacobian
 import astropy.units as u
 from astropy.table import Table
 from gammapy.maps import MapAxis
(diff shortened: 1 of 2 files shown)


gammapy/modeling/models/spectral.py
```diff
@@ -47,34 +46,45 @@ def __sub__(self, model):
...
+
+    def evaluate_error(self, energy, precision=1e-4):
         """Evaluate spectral model with error propagation.
 
         Parameters
         ----------
         energy : `~astropy.units.Quantity`
             Energy at which to evaluate
+        precision : float
+            Numerical precision of the gradient evaluation.
```
cdeil: Suggest to rename the parameter and description. It's not the precision of the gradient. It's the step size for the parameter.

gammapy/modeling/models/tests/test_spectral.py
```diff
@@ -589,8 +589,21 @@ def test_ecpl_model(self):
...
         assert_allclose(out.data, [1.548176e-10, 1.933612e-11], rtol=1e-3)
 
-    def test_naima_model_error_proprgation(self):
-        # Regression test for Naima model
-        # https://github.com/gammapy/gammapy/issues/2190
-        # TODO: implement test case. Move to Naima model tests!
-        pass
+
+@requires_dependency("naima")
+def test_naima_model_error_propagation():
```
cdeil: Suggest to move the `evaluate_error` call the the Naima example we already have, as here:
https://github.com/gammapy/gammapy/pull/2602/files
Advantage: less test code.

gammapy/modeling/models/spectral.py
adonath: I renamed the parameter to `epsilon` (that's what we have for `SpectralModel.spectral_index()` as well) and changed the description accordingly.

gammapy/modeling/models/tests/test_spectral.py
adonath: Done

PR #2625, opened 2019-11-28
Add sample_sources to MapDatasetEventSampler

fabiopintore:
This PR aims at adding the method `sample_sources` to the `MapDatasetEventSampler` class. The method will sample source events once a `Mapdataset` object is given in input.

### DIFF
--- a/gammapy/cube/simulate.py
+++ b/gammapy/cube/simulate.py
@@ -109,7 +109,7 @@ class MapDatasetEventSampler:
     def __init__(self, random_state="random-seed"):
         self.random_state = get_random_state(random_state)
 
-    def _sample_coord_time(self, npred, temporal_model, gti, MC_ID=0):
+    def _sample_coord_time(self, npred, temporal_model, gti):
         """Sample source model components.
 
         Parameters
@@ -120,8 +120,6 @@ class MapDatasetEventSampler:
             Temporal model.
         gti : `MapDataset` object
             Good time intervals of the given MapDataset object.
-        MC_ID : int
-            Monte Carlo identifier of the sampled source.
 
         Returns
         -------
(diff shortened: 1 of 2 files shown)


gammapy/cube/simulate.py
```diff
@@ -109,6 +109,49 @@ class MapDatasetEventSampler:
...
+        for idx, evaluator in enumerate(dataset._evaluators):
+            table = Table()
+            evaluator.edisp = None
+            evaluator.psf = None
+            npred = evaluator.compute_npred()
+            n_events = self.random_state.poisson(np.sum(npred.data))
+
+            # sample position
+            coords = npred.sample_coord(n_events, self.random_state)
+            table["ENERGY"] = coords["energy"]
```
adonath: Note that this column should be `"ENERGY_TRUE"` and the same for `"RA_TRUE"` and `"DEC_TRUE"`

fabiopintore:
> Thanks @fabiopintore! I've left one inline-comment, but I have one more suggestion:
> 
> To reduce code duplication between the two sampling methods, I would propose to introduce a `._sample_coord_time(npred, temporal_model, gti)` method, that can be re-used. This method contains the part:
> 
> ```python
>        table = Table()
>        n_events = self.random_state.poisson(np.sum(npred.data))
> 
>         # sample position
>         coords = npred.sample_coord(n_events, self.random_state)
>         table["ENERGY"] = coords["energy"]
>         table["RA"] = coords.skycoord.icrs.ra.deg
>         table["DEC"] = coords.skycoord.icrs.dec.deg
>         table["MC_ID"] = 0
> 
>         # sample time
>         time_start, time_stop, time_ref = (
>             gti.time_start,
>             gti.time_stop,
>             gti.time_ref,
>         )
>         time = temporal_model.sample_time(n_events, time_start, time_stop, self.random_state)
>         table["TIME"] = u.Quantity(((time.mjd - time_ref.mjd) * u.day).to(u.s)).value
> ```
> 
> Which is common to the `.sample_background()` and `.sample_sources()`methods. I hope this makes sense...

Yes, it makes sense. So what are the steps now? Merge this PR and then open a new one where I introduce the `_sample_coord_time` method, modifying in parallel the `.sample_background()` and `.sample_sources()` methods?

adonath:
@fabiopintore Please just add the modifications in this PR...

adonath:
Thanks a lot @fabiopintore!

PR #2629, opened 2019-11-28
Improve gammapy.data.Observations

cdeil:
This PR improves `gammapy.data.Observations`. It changes from a public data member `list` to a private one `_observations`. This matches what I recently did for other containers in Gammapy: Parameters, Datasets, SkyModels.

It doesn't add much value, but I would suggest we sub-class `collections.abc.Sequence` here and also for `Parameters`, `Datasets` and `SkyModels`. This checks that the required `__getitem__` and `__len__` have been added (which is the case), and offers mixin methods `__contains__`, `__iter__`, `__reversed__`, `index`, and `count`, so that overall those containers then satisfy the so-called Python `Sequence` protocol. See https://docs.python.org/3/library/collections.abc.html

@adonath - OK?

Should I add `collections.abc.Sequence` now in this PR?

### DIFF
--- a/gammapy/analysis/core.py
+++ b/gammapy/analysis/core.py
@@ -84,7 +84,7 @@ class Analysis:
         obs_list = self.datastore.get_observations()
         if len(self.config.observations.obs_ids):
             obs_list = self.datastore.get_observations(data_settings.obs_ids)
-        selected_obs["OBS_ID"] = [obs.obs_id for obs in obs_list.list]
+        selected_obs["OBS_ID"] = [obs.obs_id for obs in obs_list]
         ids = selected_obs["OBS_ID"].tolist()
         # TODO
         # if self.config.observations.obs_file:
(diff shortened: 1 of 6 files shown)


adonath:
+1 to change to `collections.abc.Sequence`, I think it's good way to define the minimal API for these container classes...

cdeil:
Done in fa03086418a9e74c5c247d8879a2fdcd57b5eeaa .

I'll merge as soon as CI passes, locally it's fine.

Do we define any other list-like classes in Gammapy?

The `__init__` of the four cases is always different. I could send another PR to make it uniform. But @adonath - Maybe you touched the inits already for datasets to do the name cleanup?

adonath:
@cdeil No, so far I didn't touch the `__init__` methods, please go ahead with the clean-up...

cdeil:
CI fails are unrelated, just `gammapy download` failed. Merging now.

PR #2631, opened 2019-11-29
Add Datasets read and write methods

QRemy:
Change names of datasets serialisation methods
datasets.to_yaml() -> datasets.write()
datasets.from_yaml() -> datasets.read()

### DIFF
--- a/gammapy/cube/tests/test_fit.py
+++ b/gammapy/cube/tests/test_fit.py
@@ -672,7 +672,7 @@ def test_datasets_io_no_model(tmpdir):
 
     datasets = Datasets([dataset_1, dataset_2])
 
-    datasets.to_yaml(path=tmpdir, prefix="test")
+    datasets.write(path=tmpdir, prefix="test")
 
     filename_1 = tmpdir / "test_data_1.fits"
     assert filename_1.exists()
--- a/gammapy/modeling/datasets.py
+++ b/gammapy/modeling/datasets.py
@@ -160,8 +160,8 @@ class Datasets(collections.abc.Sequence):
         """
         from .serialize import dict_to_datasets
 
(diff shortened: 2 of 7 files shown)


cdeil:
@QRemy - Please when opening a PR, always set a project, milestone and assignee. Assignee is the guy you want to review and merge your PR, in this case @adonath or me is fine.

If you don't have permission to set those things
 let me know and I'll add you to the Gammapy maintainers team.

cdeil:
@QRemy - Thanks.

Please adapt this caller:

https://dev.azure.com/gammapy/gammapy/_build/results?buildId=2699&view=logs&jobId=d7d66435-5ae4-561a-5876-2cd4e19b3ce2&taskId=922dfe32-5b90-53bc-18db-0fa0ecbbd97b&lineStart=142&lineEnd=143&colStart=1&colEnd=1

cdeil:
Probably for consistency, we should call `make_path` in `read`, just like you do in write in https://github.com/gammapy/gammapy/pull/2631/files#diff-8b37a3d39becd8c4acbc3951517b883fR182 and like we do in all our read methods.

What this will achieve is that variable names are expanded, i.e. we can use `Datasets.read("$GAMMAPY_DATA/some/file.yaml")` in tutorials and the env var will be expanded.
I'm not sure it's a good pattern, but we have it in basically all our read methods, so `Datasets.read` should work the same way.

cdeil:
Changed title to "Add Datasets read and write methods" which I think will be more useful for users to see in the changelog.

I'm merging this now. Obviously we don't use this enough in the docs. @AtreyeeS - Please try to use it in the first and second analysis tutorial and your benchmark scripts, and report if you find any issues.

PR #2635, opened 2019-11-29
Change Datasets model to models

### LINKED ISSUE #2102: Single object vs container confusion

cdeil:
This PR is an attempt to fix #2102 by changing `dataset.model` to `dataset.model` on all our five dataset classes. If we do this change, we should do it now, before v0.15.

There's still two failing tests (see [here](https://gist.github.com/cdeil/3cb466921b87b79acd91dffb930916be)) and I didn't update the tutorials or docs yet. There's merge conflicts already. Probably I'll have to re-start, but this PR shows that the change can be done within an hour.

As far as I understand it, our codebase already has migrated to this design that `dataset.model` should always be a `SkyModels` (plural) object. So IMO this change is the logical consequence, easier to understand for users. The logic is to accept for convenience different objects in init and in the setter (usually `SkyModel`, `SkyModels` and `list` of `SkyModel`), but then in the `@models.setter` to have checks and converters that guarantee that `_models` will always be a `SkyModels`, so everywhere where we access it (~ 1000 places) we can rely on that.

There's still the concern that we allow `None` and empty `SkyModels` list, which can cause issues (see [here](https://gist.github.com/cdeil/3cb466921b87b79acd91dffb930916be)). We could either allow both and getters have to check on access. Or we only allow one in the setter.

@adonath @registerrier @QRemy - Thoughts?

If we go this way: should I rebase here and we try to push this in? Or should I re-start at a better time (e.g. Sunday evening)?

### DIFF
--- a/gammapy/cube/fit.py
+++ b/gammapy/cube/fit.py
@@ -229,9 +229,7 @@ class MapDataset(Dataset):
 
         evaluators = []
         for model in self.models:
-            evaluator = MapEvaluator(
-                model, evaluation_mode=self.evaluation_mode
-            )
+            evaluator = MapEvaluator(model, evaluation_mode=self.evaluation_mode)
             evaluator.update(self.exposure, self.psf, self.edisp, self._geom)
             evaluators.append(evaluator)
 
--- a/gammapy/cube/fit.py
+++ b/gammapy/cube/fit.py
@@ -998,7 +998,7 @@ class MapDatasetOnOff(MapDataset):
 
     Parameters
     ----------
-    model : `~gammapy.modeling.models.SkyModel` or `~gammapy.modeling.models.SkyModels`
+    models : `~gammapy.modeling.models.SkyModels`
         Source sky models.
     counts : `~gammapy.maps.WcsNDMap`
         Counts cube
@@ -1035,7 +1035,7 @@ class MapDatasetOnOff(MapDataset):
 
     def __init__(
         self,
-        model=None,
+        models=None,
         counts=None,
         counts_off=None,
         acceptance=None,
@@ -1070,7 +1070,7 @@ class MapDatasetOnOff(MapDataset):
         self.mask_fit = mask_fit
         self.psf = psf
         self.edisp = edisp
-        self.model = model
+        self.models = models
         self.name = name
         self.mask_safe = mask_safe
         self.gti = gti
@@ -1100,8 +1100,8 @@ class MapDatasetOnOff(MapDataset):
         """List of parameters (`~gammapy.modeling.Parameters`)"""
         parameters = []
 
-        if self.model:
-            parameters += self.model.parameters
+        if self.models:
+            parameters += self.models.parameters
 
         return Parameters(parameters)
 
--- a/gammapy/cube/fit.py
+++ b/gammapy/cube/fit.py
@@ -807,11 +807,16 @@ class MapDataset(Dataset):
 
     def to_dict(self, filename=""):
         """Convert to dict for YAML serialization."""
+        if self.models is None:
+            models = []
+        else:
+            models = [_.name for _ in self.models]
+
         return {
             "name": self.name,
             "type": self.tag,
             "likelihood": self.likelihood_type,
-            "models": [_.name for _ in self.models],
+            "models": models,
             "background": self.background_model.name,
             "filename": str(filename),
         }
--- a/gammapy/cube/fit.py
+++ b/gammapy/cube/fit.py
@@ -44,7 +44,7 @@ class MapDataset(Dataset):
 
     Parameters
     ----------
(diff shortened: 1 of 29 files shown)


gammapy/cube/fit.py
```diff
@@ -804,7 +811,7 @@ def to_dict(self, filename=""):
             "name": self.name,
             "type": self.tag,
             "likelihood": self.likelihood_type,
-            "models": [_.name for _ in self.model],
+            "models": [_.name for _ in self.models],
```
QRemy: This fail for the None case no ?

cdeil: Yes.
How should we fix?
Adapt this and other callers and add an if/else?
Or create `self._models = SkyModels([])` in `@models.setter`?

QRemy: we have this in the spectrumDatasetOnOff
"""
        if self.model is not None:
            models = [_.name for _ in self.model]
        else:
            models = []

        return {
            "name": self.name,
            "type": self.tag,
            "models": models,
            "likelihood": self.likelihood_type,
            "filename": filename,
        }
"""

cdeil: So I put the `if` and we allow `None`?
I'm not sure, but I think that would also be my preference.

adonath:
Thanks @cdeil, I'm +1 to rebase. In #2632 I removed the `e_true` argument from the `SpectrumEvaluator`, so you have to adapt the changes here.

Concerning `None` vs. an empty `SkyModels`. I don't really care. Both works, I guess the empty models container leads to less code, but as long as we don't have something like `SkyModels.append()` it's barely useful for users and becomes and implementation detail...do whatever you prefer.

gammapy/cube/fit.py
cdeil: I've put the same if in both places. @QRemy - Thanks for the advice!

cdeil:
@adonath - I've rebased and fixed all tests. I did not update the tutorials or docs yet.

I have to go now. If you can finish this up and merge it in, please do.
Otherwise, please leave a review and I'll try to do it tonight or tomorrow.

cdeil:
Just FYI: I would suggest to rename `SkyModels` -> `Models` tomorrow, and use that as a generic container, i.e. for special datasets like dark matter, they can add a model on the dataset anywhere they like with whatever parameters they like, and everything will work as long as they stick the models in `dataset.models`, it can be background or IRF or any other models.
Whether we manage to refactor BG models or not before v0.15 or v1.0, this would be my suggestion where to go.

adonath:
@cdeil I have some time now to work on this, I'll fix the remaining test fails and merge this PR after...

adonath:
Thanks @cdeil!

cdeil:
@adonath - Thank you for finishing this up!

PR #2636, opened 2019-11-29
Add map_selection parameter in Analysis 

Bultako:
**Description**

This pull request adds `map_selection` parameter to the Analysis class and YAML files of the High-lelve interface, as an optional param for `MapDatasetMaker`. 

**Dear reviewer**

I have the following warning when `map_selection=['background']`

```
/Users/jer/git/gammapy/gammapy/gammapy/cube/psf_kernel.py:109: RuntimeWarning: invalid value encountered in true_divide
  img += vals.value / vals.sum().value
```

Tests have not been implemented, but test_coverage has not decreased.

### DIFF
--- a/gammapy/analysis/config.py
+++ b/gammapy/analysis/config.py
@@ -12,6 +12,7 @@ import yaml
 from pydantic import BaseModel, FilePath
 from pydantic.utils import deep_update
 from gammapy.utils.scripts import make_path, read_yaml
+from gammapy.cube import MapDatasetMaker
 
 __all__ = ["AnalysisConfig"]
 
@@ -67,6 +68,14 @@ class FrameEnum(str, Enum):
 
 class BackgroundMethodEnum(str, Enum):
     reflected = "reflected"
+
+
+class MapSelectionEnum(str, Enum):
+    counts = "counts"
+    exposure = "exposure"
+    background = "background"
+    psf = "psf"
+    edisp = "edisp"
 
 
 class GammapyBaseConfig(BaseModel):
@@ -155,7 +164,7 @@ class DatasetsConfig(GammapyBaseConfig):
     type: ReductionTypeEnum = ReductionTypeEnum.spectrum
     stack: bool = True
     geom: GeomConfig = GeomConfig()
-    map_selection: List[str] = None
+    map_selection: List[MapSelectionEnum] = MapDatasetMaker.available_selection
     background: BackgroundConfig = BackgroundConfig()
     on_region: SpatialCircleConfig = SpatialCircleConfig()
     containment_correction: bool = True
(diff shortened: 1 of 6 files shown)


adonath:
@Bultako I can take a look at the warning in master later...

cdeil:
@Bultako - Thanks!

Maybe change the default to `['counts', 'exposure', 'background']` to be consistent with the code?

https://docs.gammapy.org/0.14/api/gammapy.cube.MapMaker.html#gammapy.cube.MapMaker.run

And maybe adapt one of the existing tests to run with `selection = ['exposure', 'background']` and assert after that `counts` isn't set? I'm not sure if that is a good idea / easy. If not, just don't test this new setting, it should work OK.

cdeil:
If you want excellent error messages, I think you'd have to introduce either an `Enum`, or write a validate method. You know, if someone puts `selection = ["spam"]` in the config. Also something that's nice to have, not must have, like the test.

adonath:
@cdeil and @Bultako  The new default is to compute all maps  `["counts", "background", "exposure", "edisp", "psf"]`, see here: https://docs.gammapy.org/dev/api/gammapy.cube.MapDatasetMaker.html#gammapy.cube.MapDatasetMaker

cdeil:
@Bultako - You can import and put `MapDatasetMaker.available_selection` as default.
```
In [1]: from gammapy.cube import MapDatasetMaker                                                                                                             

In [2]: MapDatasetMaker.available_selection                                                                                                                  
Out[2]: ['counts', 'exposure', 'background', 'psf', 'edisp']
```
We'll have to talk about how to keep config and options consistent between AnalysisConfig and the rest of Gammapy, but for now, importing things and setting them as default in `gammapy/analysis/config.py` is the way to go, no?

adonath:
@cdeil I agree maintaining the defaults in `config.py` is the way to go in future...

cdeil:
> I agree maintaining the defaults in config.py is the way to go in future...

Here I was suggesting the opposite. To import the available options from `MapDatasetMaker` and set them also in `config.py`. So `gammapy.analysis.config` is high-level, importing defaults from lower-level places.
Maintaining configs in `config.py` would mean to import `gammapy.analysis.config` in `gammapy.cube` and other places.
It's not clear to me which way to go in Gammapy, how to integrate low- and high-level better.

I think for now for v0.15, we should go the way to import in `config.py` from other places in Gammapy.

@adonath @Bultako  - Agreed? Or go the other way?

adonath:
> I think for now for v0.15, we should go the way to import in config.py from other places in Gammapy.

I agree. Importing from `config.py` will create a lot of circular imports now. So let' avoid it and import from the lower level...

PR #2627, opened 2019-11-28
Activate PSFMap for fitting

AtreyeeS:
**Description**

This pull request activates `PSFMap` on the MapEvaluator. Similar to what is done in #2587 , the option to keep PSFKernel is still on. Once this PR is merged, I will deactivate the kernels and adapt the notebooks.

**Dear reviewer**

One specific doubt I have is what should be the `max_radius` in 
`self.psf = psf.get_psf_kernel(self.model.position, geom=exposure.geom, max_radius=0.8 * u.deg)`
Having it as a parameter will be extremely confusing for the users. Currently I am passing a large enough value which should hopefully suffice for most cases.

I did not adapt the tests in `analysis` to avoid conflicts with the changes @Bultako might be making.

PR #2639, opened 2019-11-29
Use PSFMap in Analysis class

adonath:
**Description**

This pull request activates the `PSFMap` in the `Analysis` class. I also removed the `psf_kernel_radius` parameter, which is not needed anymore, as we changed to full containment in #2627.


adonath:
@AtreyeeS Locally not a single number changed. But we will wait for Ci anyway...

PR #2642, opened 2019-11-29
Improve Analysis and AnalysisConfig

Bultako:
**Description**

This pull request is a follow-up of https://github.com/gammapy/gammapy/pull/2621
It addresses the final suggestions made in https://github.com/gammapy/gammapy/pull/2621#issuecomment-559614542

- code refactoring for `AnalysisConfig.update()` method and `Analysis.set_model()`
- adds classmethod for `Analysis.read_model(filename)` 
- removes the built-in `AnalysisConfig.help()` method

It also removes a not needed `config.yaml` file.


gammapy/analysis/tests/test_analysis.py
```diff
@@ -23,25 +23,21 @@ def test_init():
...
     general:
         outdir: test
     """
     analysis.update_config(data)
     assert analysis.config.general.outdir == "test"
-    analysis = Analysis(cfg)
-    data = "spam"
-    with pytest.raises(ValueError):
+    analysis = Analysis(AnalysisConfig())
+    data = 0
```
cdeil: What is this `data = 0`? Not used? Remove?

gammapy/analysis/core.py
```diff
@@ -122,36 +122,35 @@ def get_datasets(self):
...
         for dataset in self.datasets:
             dataset.model = self.model
-
         log.info(self.model)
 
+    @classmethod
+    def read_model(cls, filename):
+        """Reads model rom YAML file."""
+        filepath = make_path(filename)
+        return SkyModels.from_yaml(filepath)
```
cdeil: I think `Analysis.read_model` should read and set the model, i.e. call `set_model` from inside `read_model`, no?
That's more conveninet, and if someone just wants to read a model they call `SkyModels.from_yaml`, no need to bother the Analysis class.

gammapy/analysis/core.py
```diff
@@ -122,36 +122,35 @@ def get_datasets(self):
...
+    def set_model(self, model):
+        """Set model to datasets.
 
         Parameters
         ----------
-        model: dict or string
-            Dictionary or string in YAML format with the serialized model.
-        filename : string
-            Name of the model YAML file describing the model.
+        model: `SkyModel` or string
```
cdeil: Put space after colon. Rename `model` -> `models`, no?

Bultako:
@cdeil @registerrier @adonath 
I've fixed the `light_curve.ipynb` tutorial to work with the modified `set_model` method. Though it should pass the tests now, somehow it is producing bad results in the cells of the fitting.

cdeil:
I merged #2640 just now, so you'll have to revert or rebase the LC tutorial edits here. I can have a look in 1-2 hours, if the issue is still there, let me know.

Bultako:
@cdeil 
thanks, now it works nicely

PR #2644, opened 2019-12-01
Simplify SpectrumDatasetMaker parameters

adonath:
**Description**

This pull request removes the `e_true`, `e_reco` and `region` parameters from the `SpectrumDatasetMaker` and instead takes an empty dataset on `.run()`. This change 
has the advantage that the configuration of a dataset is all in a single place: in the `SpectrumDataset.create()` method. Before this PR it was repeated in the `SpectrumDatasetMaker`.


gammapy/analysis/core.py
```diff
@@ -18,6 +18,7 @@
     FluxPointsEstimator,
     ReflectedRegionsBackgroundMaker,
     SpectrumDatasetMaker,
+    SpectrumDataset
```
cdeil: Run `make polish` before committing (or before merging PRs), to keep formatting clean in gammapy master, and `make polish` available and useful for everyone, to not have to worry about committing unrelated changes.

cdeil:
I'll merge this now, since @Bultako in #2646 and also I want to touch these files tonight.
@adonath - Thanks!

PR #2645, opened 2019-12-01
Remove FluxPointsDataset chi2assym option

cdeil:
This PR removes the `chi2assym` from `FluxPointsDataset`.

It wasn't  covered by a test, and not well documented (with an example and in the [stats](https://docs.gammapy.org/0.14/stats/index.html) docs, explaining it's relation to normal chi2 and likelihood profiles).

My main argument why we shouldn't offer is is that is isn't standard or useful really. At https://cxc.harvard.edu/sherpa/statistics/ there's many chi2 variants, but the asymmetric one isn't one of them. I don't recall having seen the asymmetric chi2 used in papers at all. And it's not a good technique, if the likelihood is asymmetric, then the likelihood profile method which we should add soon is much better.

This is also a follow-up to #2546, removing the last case of confusing of "likelihood" vs "stat" vs "likelihood_type" names in the Gammapy codebase.

### DIFF
--- a/gammapy/spectrum/flux_point.py
+++ b/gammapy/spectrum/flux_point.py
@@ -1155,8 +1155,6 @@ class FluxPointsDataset(Dataset):
         Flux points.
     mask_fit : `numpy.ndarray`
         Mask to apply for fitting
-    likelihood : {"chi2", "chi2assym"}
-        Likelihood function to use for the fit.
     mask_safe : `numpy.ndarray`
         Mask defining the safe data range.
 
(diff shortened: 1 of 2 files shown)


cdeil:
It was used here in the SED fitting tutorial: https://travis-ci.org/gammapy/gammapy/jobs/619286087#L1804

I noticed that `drop_ul` here wasn't workig here:
https://docs.gammapy.org/0.14/notebooks/sed_fitting_gammacat_fermi.html#Load-spectral-points
It kept the UL columns, because for 3FHL and 3GL the `nan` values were in `dnde_errn`, but the `drop_ul` method only checked the `dnde` column.
This is the only place where we call `drop_ul`, and it was buggy.

@adonath  - Given that it's difficult to define exactly how `drop_ul` should work, i.e. what columns it should check, how about we simply delete that method?
For now, I'll just fix this case manually, without calling `drop_ul`.

cdeil:
Fixed in 1ca828396 .

@adonath - Are you OK with this removal?
(not sure if I added this asym feature or you or if you care about it)

PR #2646, opened 2019-12-01
Add obs_file and obs_time features for observation selection in Analysis

Bultako:
**Description**

This pull request adds still missing features to improve observation filtering with the Analysis class of the high-level interface. 

- adds `obs_ids` read from a file declared in `obs_file` parameter
- filters `obs_ids` with start nad stop time declared in `obs_time` parameters
- makes `border=0` when filtering `of obs_ids ` with `obs_cone` parameter
- removes old comments not needed inducing to confusion in `light_curve.ipynb`

### DIFF
--- a/gammapy/analysis/core.py
+++ b/gammapy/analysis/core.py
@@ -4,6 +4,7 @@ import logging
 import numpy as np
 from astropy import units as u
 from astropy.coordinates import SkyCoord
+from astropy.table import Table
 from regions import CircleSkyRegion
 import yaml
 from gammapy.analysis.config import AnalysisConfig
@@ -79,9 +80,17 @@ class Analysis:
 
         log.info("Fetching observations.")
         observations_settings = self.config.observations
-        if len(observations_settings.obs_ids) and observations_settings.obs_file is not None:
-            raise ValueError("Values for both parameters obs_ids and obs_file are not accepted.")
-        elif not len(observations_settings.obs_ids) and observations_settings.obs_file is None:
+        if (
+            len(observations_settings.obs_ids)
+            and observations_settings.obs_file is not None
+        ):
+            raise ValueError(
+                "Values for both parameters obs_ids and obs_file are not accepted."
+            )
+        elif (
+            not len(observations_settings.obs_ids)
+            and observations_settings.obs_file is None
+        ):
             obs_list = self.datastore.get_observations()
             ids = [obs.obs_id for obs in obs_list]
         elif len(observations_settings.obs_ids):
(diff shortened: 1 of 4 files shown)


cdeil:
I think only allowing one or the other would be best. Letting one take precedence (e.g. `obs_config.obs_ids` over `obs_config.obs_file`) is also "teachable". Taking the union or intersection of the two I think is hard to teach. Basically: try to choose the behaviour that users would guess without reading the docs. And if that doesn't exist because it's intrisically unclear, don't guess.
```
>>> import this
Errors should never pass silently.
In the face of ambiguity, refuse the temptation to guess.
```

Bultako:
@cdeil 
I've made the modif you suggest to raise an error when both `obs_ids` and `obs_file` have declared values. But I think it is good to keep the return all observations from datastore behaviour when no values for none of `obs_ids` and `obs_file` are declared. If these values are taken as filters, this is what I would expect if I chose not to apply a filter.

cdeil:
@Bultako - I think the test you have for `obs_file` is still not really testing that the obs list in that file has any effect? You write a list of all `OBS_ID` from the datastore, so whether or not this obs file exists and has a subset of runs has no effect. A better test IMO would be like the one I suggested above, where I put two explicit OBS_ID and then check that those are the ones used. You could add another test where the user lists an OBS_ID in the file that doesn't exist (e.g. because of a typo, in the test you could put "42"), and you check that an error is raised in that case.

Overall the tests you have are too long, it would be better to split those out into multiple tests. Usually tests should have the structure "arrange - act - assert", but these tests are "arrange - act -assert - arrange - act -assert - arrange - act - assert - ..." and it's not always obvious if the previous "arrange" or "act" has any effect on later tests. I think it's OK now, but splitting into multiple tests would make the tests simpler. (just listed one after the other, not using parametrize).

gammapy/analysis/core.py
```diff
@@ -79,33 +78,36 @@ def get_observations(self):
...
+        elif not len(observations_settings.obs_ids) and observations_settings.obs_file is None:
+            obs_list = self.datastore.get_observations()
+            ids = [obs.obs_id for obs in obs_list]
+        elif len(observations_settings.obs_ids):
+            obs_list = self.datastore.get_observations(observations_settings.obs_ids)
+            ids = [obs.obs_id for obs in obs_list]
+        else:
+            path = make_path(self.config.observations.obs_file)
+            with open(path, 'r') as f:
+                ids = [int(obs_id) for obs_id in f.readlines()]
```
cdeil: Even if it's a very simple format here, I would suggest to use `Table.read(path, format="ascii.csv")` or some other ascii variant format here, just for consistency with the rest of Gammapy where we always used `Table.read` to read any table.

PR #2651, opened 2019-12-02
Add SkyModels read and write methods

cdeil:
In #2621,  #2631 and #2642 we change `AnalysisConfig` and `Datasets` to have a `read` and `write` method, which calls `from_yaml` and `to_yaml` and does the I/O.

This two-step way to do I/O, and to have `read` and `write` as the methods that hit the disk is a good pattern in general, it's also done e.g. in Astropy with `Table.read` and `Table.write`.

Suggest to do the same in `SkyModels`, to rename the existing `from_yaml` and `to_yaml` there to `read` and `write` (and to keep the `from_yaml` and `to_yaml` and return YAML strings, as interface for users that want to work with YAML in-memory (e.g. to print in a notebook, or to send to a database or whatever they like).

Example how it should look:

https://github.com/gammapy/gammapy/blob/e93e605084e1dffa0e47eeaff64c3be851717238/gammapy/analysis/config.py#L228-L232

@QRemy @adonath - Thoughts?

### DIFF
--- a/gammapy/analysis/core.py
+++ b/gammapy/analysis/core.py
@@ -162,7 +162,7 @@ class Analysis:
     def read_models(self, path):
         """Read models from YAML file."""
         path = make_path(path)
-        models = SkyModels.from_yaml(path)
+        models = SkyModels.read(path)
         self.set_models(models)
 
     def run_fit(self, optimize_opts=None):
--- a/gammapy/modeling/models/cube.py
+++ b/gammapy/modeling/models/cube.py
@@ -8,7 +8,7 @@ import numpy as np
 import astropy.units as u
 from gammapy.maps import Map
 from gammapy.modeling import Model, Parameter, Parameters
-from gammapy.utils.scripts import make_path, write_yaml
+from gammapy.utils.scripts import make_path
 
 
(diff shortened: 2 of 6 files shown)


cdeil:
@adonath - See commits attached here. Thoughts?

I'll run `make docs-all` now locally to see if I missed any tutorial notebooks to update.

cdeil:
@AtreyeeS @registerrier - You can now go ahead and put examples of `SkyModels.read`, `SkyModels.write`, `SkyModels.from_yaml` and `SkyModels.to_yaml` in the docs.

@registerrier - Specifically here I would suggest to show `SkyModels.from_yaml`:
https://docs.gammapy.org/dev/notebooks/analysis_1.html#Model-fitting

It's very typical that users have 2 or more model components. In that case they should use `SkyModels` I/O, not `SkyModel` I/O. For ctools, I'm not even sure there is an I/O on the individual model object. I guess there is, but it's never advertised in the docs, in Fermi ST and ctools there's always one standard way shown, to have a "source library" with one or more sources. I suggest we do the same, teach it that way, or users will be confused.

PR #2652, opened 2019-12-02
Adapt analysis3d with HLI

AtreyeeS:
**Description**
I have adapted the `analysis3d` tutorial  to use the `Analysis` class for the data reduction and then do the modelling and and fitting by hand. I also removed the `analysis_3d_joint` and appended this to the end of the present tutorial.

**Dear reviewer**

I have not stripped the output cells yet because it might be a bit easier to go through the tutorial with the output visible. Once the PR is merge ready, I will clean it up


adonath:
Thanks @AtreyeeS! Minor comment first: please always strip the output, because committing the output and removing it again in a follow up commit will create a diff which is stored in the git history. If we do this many times the size of our `.git` folder will grow for no good reason. For review I'm always checking out the branch of the PR and execute the notebooks locally anyway.

To avoid the diff in the git history for this PR you can use `git reset --soft HEAD~`, strip the output, re-commit the changes and force push (`git push -f AtreyeeS analysis3d`) to this PR (maybe you wanted to do this anyway, then please just ignore this comment...)

AtreyeeS:
@adonath ! Sure! Will keep in mind from next time. Stripped the output

registerrier:
Apparently Travis failed for this notebook. I have not tried it locally yet. Need to check.

AtreyeeS:
> Apparently Travis failed for this notebook. I have not tried it locally yet. Need to check.

The travis fail seems to be related to `test_ring_bkg_maker()` which is not affected by this PR!

cdeil:
@AtreyeeS - You have to be careful with heading levels in the notebook, since they are used for the navigation bar on the left. Currently it looks like this:

Screenshot 2019-12-02 at 19 55 47

I'll apply some fixes now in master, but likely you'll have to take another look and polish.
The goal should be that each notebook starts with a level 1 heading (single `#`), and then has ~ 5 parts that are given by a short level 2 heading (two `##`). You can have level 3 headings (three `###`), which can be useful to have as "anchors" that one can link to, but those don't show up in the navigation bar.

Issue #1952, opened 2018-11-26
Creating axis with one bin with node_type center


AtreyeeS:
Hello! I am not sure if it is a bug or a problem with my understanding. It seems it is impossible to create an axis with only one energy bin and `node_type=center`
eg:
`import numpy as np`
`from gammapy.maps import MapAxis`
`MapAxis(nodes=np.linspace(0, 10, 1), unit='TeV', name='test', node_type='center', )`

fails. Incidentally, `axis = MapAxis(nodes=np.linspace(0, 10, 1),  node_type='center')` pretends to work, but then, accessing the `axis` object fails.

AtreyeeS:
Well, as @registerrier just explained to me, with only one `node`, it obviously fails to understand what the `min` and `max` values are. So, it should fail.

But I still think its a bit weird. How does a a user create a differential map with only one bin? At the very least, I think the axis creation should fail at the first step instead of pretending to work and then giving strange problems later.

comments @cdeil , @adonath ?

Bultako:
> But I still think its a bit weird. How does a a user create a differential map with only one bin? 

I guess a quick patch in the code of MapAxis to create an axis of one single bin with `node_type='center'` could be to create the axis with `node_type='edges'`

@AtreyeeS 
Does the command below solve your needs?

```python
MapAxis(nodes=(0,10), unit='TeV', name='test', node_type='edges', )

MapAxis

	name       : test      
	unit       : 'TeV'     
	nbins      : 1         
	node type  : edges     
	edges min  : 0.0e+00 TeV
	edges max  : 1.0e+01 TeV
	interp     : lin   
```

AtreyeeS:
@Bultako : Well, this should work for most use cases, but this then, `node_type` is `edges`. And in some special cases, eg: https://github.com/gammapy/gammapy/blob/93b4238ae6a3c36e605f77eaa16b9def0692b34d/gammapy/cube/models.py#L301

we specifically want it to be center.

adonath:
Thanks @AtreyeeS, for bringing this issue up. Conceptionally, I think, it makes totally sense to have a `MapAxis` with just one bin and `node_type=center`. Technically there is the issue, that we use linear interpolation to transform pixel to physical coordinates for the axis, which requires at least two nodes.
So I'm not sure what to do here. Maybe implement a special case and change to zero order interpolation for one-binned axes?

cdeil:
This is the error:
```
>>> from gammapy.maps import MapAxis
[...]

PR #2653, opened 2019-12-02
Disallow one energy bin for node-type 'center'

adonath:
**Description**

This pull request addresses #1952. While looking at the issue again I noticed that in the modelling and fitting we require the bin-width to be defined, because we integrate the spectral model over a given energy range. This requires the `node_type="edges"` even if there is only a single energy bin defined. This PR just disallows the cases `nbin=1` and `node_type="center"`, so it fails early with a good error message.

### DIFF
--- a/gammapy/maps/geom.py
+++ b/gammapy/maps/geom.py
@@ -322,6 +322,9 @@ class MapAxis:
         if ~(np.all(nodes == np.sort(nodes)) or np.all(nodes[::-1] == np.sort(nodes))):
             raise ValueError("MapAxis: node values must be sorted")
 
+        if len(nodes) == 1 and node_type == "center":
+            raise ValueError("Single bins can only be used with node-type 'edges'")
+
         if isinstance(nodes, u.Quantity):
             unit = nodes.unit if nodes.unit is not None else ""
             nodes = nodes.value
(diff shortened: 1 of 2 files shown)


AtreyeeS:
What happens when we want to create an exposure map with only one energy bin, which might be expected, say,  for an image analysis...?

adonath:
@AtreyeeS This was exactly my point: currently we support image analysis with a 3D map and a single energy bin. In the model evaluation (see [here](https://github.com/gammapy/gammapy/blob/master/gammapy/cube/fit.py#L1475)) we currently require the bin-width of the energy bin to be defined. The same holds if we change to `SpectralModel.integrate(emin=, emax=)` at some point. So right now we need the information on the energy range in the exposure (even for an image analysis) to be able to evaluate the model...

AtreyeeS:
Yes, I completely see your point. My only concern is image_analysis should not be affected. I am not sure if we have tests to cover it now. What happens when I do a `squash()` on a differential map now?

adonath:
@AtreyeeS By calling `.squash()` the node-type of the corresponding axis is changed to `edges`, which is a well defined behaviour I think. The same happens for the call to `.sum_over_axes()`.

AtreyeeS:
Okay! That makes sense

PR #2657, opened 2019-12-02
Adapt simulate3d to work with the HLI

AtreyeeS:
**Description**

This pull request adapts simulate3d to work with the new HLI.

**Dear reviewer**

I do not know if this is easier to use than the previous implementation. It seems to me that without using the HLI, it is fewer lines of code. Or perhaps I am just not very familiar with the Analysis class yet... Please take a look and let me know if you prefer this to the previous one

### DIFF
--- a/gammapy/cube/fit.py
+++ b/gammapy/cube/fit.py
@@ -254,8 +254,10 @@ class MapDataset(Dataset):
             return self.counts.geom
         elif self.background_model is not None:
             return self.background_model.map.geom
-        else:
+        elif self.exposure:
             return self.exposure.geom
+        else:
+            raise ValueError("No map available to extract shape")
 
     @property
     def _energy_axis(self):
@@ -264,14 +266,7 @@ class MapDataset(Dataset):
     @property
     def data_shape(self):
         """Shape of the counts or background data (tuple)"""
-        if self.counts:
-            return self.counts.data.shape
-        elif self.background_model:
-            return self.background_model.map.data.shape
-        elif self.exposure:
-            return self.exposure.data.shape
-        else:
-            raise ValueError("No map available to extract shape")
+        return self._geom.data_shape
 
     def npred(self):
         """Predicted source and background counts (`~gammapy.maps.Map`)."""
--- a/gammapy/cube/fit.py
+++ b/gammapy/cube/fit.py
@@ -263,8 +263,15 @@ class MapDataset(Dataset):
 
     @property
     def data_shape(self):
-        """Shape of the counts data (tuple)"""
-        return self.counts.data.shape
+        """Shape of the counts or background data (tuple)"""
+        if self.counts:
+            return self.counts.data.shape
+        elif self.background_model:
+            return self.background_model.map.data.shape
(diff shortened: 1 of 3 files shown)


gammapy/cube/fit.py
```diff
@@ -263,8 +263,15 @@ def _energy_axis(self):
 
     @property
     def data_shape(self):
-        """Shape of the counts data (tuple)"""
-        return self.counts.data.shape
```
adonath: I just noticed we have hidden `MapDataset._geom` property, which already does exactly this. The data shape is also available as `geom.data_shape`. So my proposal would be to make the `._geom` public as and maybe call it `.geom_ref`. Then you can also use it in the `MapDatasetMaker`. The `ValueError` you can also move to `geom_ref`.

gammapy/cube/fit.py
```diff
@@ -263,8 +263,15 @@ def _energy_axis(self):
...
+        """Shape of the counts or background data (tuple)"""
+        if self.counts:
+            return self.counts.data.shape
+        elif self.background_model:
+            return self.background_model.map.data.shape
+        elif self.exposure:
```
registerrier: Normally to produce the simulation you should have a background map, no? 
One should avoid using `exposure` since it is a different data space.

gammapy/cube/make.py
```diff
@@ -312,7 +312,13 @@ def make_mask_offset_max(self, dataset, observation):
             Maximum offset mask.
 
         """
-        separation = dataset.counts.geom.separation(observation.pointing_radec)
+        if dataset.counts:
+            geom = dataset.counts.geom
+        elif dataset.background_model:
+            geom = dataset.background_model.map.geom
+        elif dataset.exposure:
```
registerrier: same here.

gammapy/cube/fit.py
AtreyeeS: Yes I completely agree. It was my confusion as well.
I do not know if it is any use case to simulate counts without a background (this is how the spectral simulation tutorial starts, so I thought there might possibly be some use case that I can't foresee). I kept the exposure only for that reason.

I am happy to remove it!

AtreyeeS: Also, the `Mapdataset._geom()` follows this

gammapy/cube/fit.py
AtreyeeS: Any reason why you want it to be called `geom_ref` instead of `geom`?

AtreyeeS:
My personal feeling is that doing simulations using the HLI is a bit tedious. I would love all of your opinions, but perhaps we should not merge this tutorial without some thought. The changes to the other files need to be merged though, so I don't know the elegant solution regarding this

cdeil:
@adonath (or anyone) - can you find a way to finalise this PR?
Either merge somehow, or close and start over later.

AtreyeeS:
We decided not to use HLI for the simulations. I have reverted to the old tutorial, and only added a `SafeMaskMaker`

adonath:
Thanks @AtreyeeS!

PR #2659, opened 2019-12-02
Add MapDatasetEventSampler sample_psf method

fabiopintore:
This PR aims at adding a `sample_psf` method in the `MapDatasetEventSampler` class. The method applies PSF correction to the coordinates of the simulated events.

### DIFF
--- a/gammapy/cube/simulate.py
+++ b/gammapy/cube/simulate.py
@@ -10,7 +10,7 @@ from gammapy.cube import (
     make_map_exposure_true_energy,
 )
 from gammapy.data import EventList
-from gammapy.maps import WcsNDMap
+from gammapy.maps import WcsNDMap, MapCoord
 from gammapy.modeling.models import BackgroundModel, ConstantTemporalModel
 from gammapy.utils.random import get_random_state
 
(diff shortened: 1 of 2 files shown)


gammapy/cube/simulate.py
```diff
@@ -176,3 +176,37 @@ def sample_background(self, dataset):
...
+        events : `EventList`
+            Event list.
+
+        Returns
+        -------
+        events : `EventList`
+            Event list with reconstructed position columns.
+        """
+
+        energy_unit = events.table["RA_TRUE"].unit
```
cdeil: This is an angular unit, not an energy unit, no?
Suggest to remove that line, and just will "deg" below.
That's what the spec requires: https://gamma-astro-data-formats.readthedocs.io/en/latest/events/events.html

gammapy/cube/simulate.py
```diff
@@ -176,3 +176,37 @@ def sample_background(self, dataset):
...
+            PSF map.
+        events : `EventList`
+            Event list.
+
+        Returns
+        -------
+        events : `EventList`
+            Event list with reconstructed position columns.
+        """
+
```
cdeil: Remove empty line.

gammapy/cube/simulate.py
```diff
@@ -176,3 +176,37 @@ def sample_background(self, dataset):
         table.rename_column("DEC_TRUE", "DEC")
 
         return EventList(table)
+
+    def sample_psf(self, psf_map, events):
+        """Sample psf map.
+
+        Parameters
+        ----------
+        psf_map : `PSFMap`
```
cdeil: You need to give full links to resolve, in this case:
```
~gammapy.cube.PSFMap
```
and below
```
~gammapy.data.EventList
```

gammapy/cube/tests/test_simulate.py
```diff
@@ -116,3 +117,19 @@ def test_MDE_sample_background():
...
     assert_allclose(bkg_evt.table["MC_ID"][0], 0, rtol=1e-5)
+
+
+@requires_data()
+def test_MDE_sample_psf():
+    psf_map = make_test_psfmap(0.1 * u.deg, shape="gauss")
+
+    dataset = dataset_maker()
+    sampler = MapDatasetEventSampler(random_state=0)
+    src_evt = sampler.sample_sources(dataset=dataset)
```
cdeil: Suggest to call `src_evt` and `src_evt_corr` both `events, like this:
```
events = sampler.sample_sources(dataset=dataset)
events = sampler.sample_psf(psf_map, events)
```
and then below you can have the asserts on `events.table`.

fabiopintore: Done

gammapy/cube/simulate.py
fabiopintore: Done

gammapy/cube/simulate.py
fabiopintore: Done

gammapy/cube/simulate.py
fabiopintore: Yes, sorry. A mistake in bad copy and paste.
I've considered only 'deg'.

adonath:
Thanks @fabiopintore!

PR #2649, opened 2019-12-02
Remove 1D and 3D config templates

cdeil:
I suggest to remove this:

- [AnalysisConfig.from_template](https://github.com/gammapy/gammapy/blob/3ddea70ee39fadfa71c985feeb6f11312b483a7b/gammapy/analysis/config.py#L203-L218)
- [gammapy/analysis/config/template-1d.yaml](https://github.com/gammapy/gammapy/blob/f4f62ad2324c81cef7870ce97d14c9ccf67bcc83/gammapy/analysis/config/template-1d.yaml)
- [gammapy/analysis/config/template-3d.yaml](https://github.com/gammapy/gammapy/blob/f4f62ad2324c81cef7870ce97d14c9ccf67bcc83/gammapy/analysis/config/template-3d.yaml)

The problem if you keep it is that users will use it (as already the case in `gammapy-benchmarks` and personal analyses by several people), and the confusion will be very large: what are the defaults from which template? when to use which one?

There's really no reason for these two templates, I think they are the remnant of an early version of the analysis config code? In the current schema there's default values for 1D and 3D analysis, and it's not clear when one would just use `config = AnalysisConfig()` (the main default, as given by `gammapy/analysis/config.py`) or when to use one of the two templates. Documenting the differences is not the way to go, it will be horribly complex for users for no benefit, and it will slow developments down a lot - really the current config schema isn't great, it should be adapted to have something more clearly structured, for that probably a proposal is needed in the form of a PR showing a new config file scheme, before touching the code.

In the Gammapy dev call last Friday this seemed to be very controversial to remove these two templates now, I'm not quite sure why, whether it was really a difference of opinion, or just a misunderstanding of what we currently have.

@adonath @registerrier - up to you.


cdeil:
I've attached a commit here that removes `AnalysisConfig.from_template` (4f4049bda7f1878f95c5a028825168484ec1e31e) and another one that renames the test example files to clarify that they are examples, not templatest (9ec7d5ef12985ecaf834c9dffc18297fbac28532).

I've also reduced from 4 to 2 runs, which improves test execution time a bit.

And I've edited the `gammapy.analysis` docs page slightly, to remove the mention of the no longer existing `AnalysisConfig.help` and `AnalysisConfig.from_template`. Overall, both `Analysis` code and docs will need further polish in the coming weeks, but I think that's as good as it gets for v0.15.

@adonath  - OK?

registerrier:
That's fine with me.

cdeil:
Should be OK (tests pass locally). Merging now.

PR #2663, opened 2019-12-03
Extend changelog for v0.15

adonath:
**Description**

This PR extends the changelog for Gammapy v0.15. I've added the most important bug fixes, API changes and new features since the beginning of the coding sprint.

**Dear reviewer**
Please check the "what's new" section for completeness and make suggestions for improving the
wording.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,38 +7,43 @@ Summary
 +++++++
 
 - Released Dec 2, 2019
-- 11 contributors
-- 97 pull requests (not all listed below)
+- 12 contributors
+- 186 pull requests (not all listed below)
 
 **What's new**
 
 A major focus of the work for Gammapy v0.15 was the clean-up and unification of
-the spectrum and cube data reduction. Gammapy now features a
-``MapDatasetMaker``, and ``SpectrumDatasetMaker`` which directly produce a
-``MapDataset`` or ``SpectrumDataset`` from DL3 data. The existing background
-estimation classes were adapted by introducing a
-``ReflectedRegionsBackgroundMaker``, ``RingBackgroundMaker`` and
-``AdaptiveRingbackgroundMaker``. Those makers also be chained to create custom
-data reduction workflows.
+the spectrum and map data reduction. Gammapy now features a ``MapDatasetMaker``,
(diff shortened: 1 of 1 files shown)


cdeil:
@adonath - Thanks!

My suggestions:

- Start by mentioning the new `gammapy.analysis`. Link to it's docs page and the "first analysis" notebook.
- Then as second I would mention the improvements to data reduction and datasets like you do, and link to the "second analysis" notebook.
- Then I would mention significant progress on documentation, linking to "overview", "howto" and "tutorials", maybe mention the PIG, and say that it's work in progress and ask for user feedback.
- I wouldn't specifically say "support for HAWC data". People expect that to mean events, maps, analysis support. Really it's just spectral points used in that tutorial. Suggest to change that tutorial to avoid the confusion, but for now, to just not claim "support for HAWC data".
- For catalog, suggest to link to the docs page
- Not sure if the position error and spatial model region feature needs a mention
- Definitely do mention that we now have support and tutorials for binned simulation, via the new observation class. That's super important for CTA users, but also existing IACTs.
- Mention that there are many breaking changes and removals, since we improve consistency as we move to v1.0? E.g. the change `dataset.model` to `dataset.models` will trip up all existing users, their scripts will break. But there's other such changes, like `read` / `write`, or `likelihood` -> `stat`, map reproject. I don't sugegst to list all the breaking changes, they are already in the PR list, in the summary I would just acknowledge that this is a rather big breaking release. "This release contains several breaking changes and removal of non-essential parts of Gammapy (see PR list below), which we made to improve consistency and quality throughout Gammapy, as we develop towards Gammapy v1.0". Something like that.
- Wording could be improved in parts, but that can either be skipped, or done in a second go.

I think this will make the "what's new" more useful for users, if they get links to the docs and examples what's new.

adonath:
Thanks @cdeil and @registerrier, I'll address your comments now...

adonath:
Thanks @cdeil  and @registerrier for the comments.

cdeil:
@adonath - Thank you for taking the time to write a great "what's new" entry. I think it's very important to point users to the right places and get good feedback. This is a good release for a 2 month dev cycle.

PR #2665, opened 2019-12-03
Polish tutorial notebooks

cdeil:
This pull request does some last-minute polish for the tutorial notebooks before v0.15:

- Rename `tutorials/sed_fitting_gammacat_fermi.ipynb` -> `tutorials/sed_fitting.ipynb` (names show up in URLs, we want short names, we said analysis notebooks shouldn't be instrument specific)
- Rename `tutorials/joint_1d_3d_analysis.ipynb` -> `tutorials/analysis_joint.ipynb` (more consistent with title and other analysis notebook names. Now appears after `analysis_3d` consistently in navbar and on page)
- Remove unused imports in tutorial notebooks
- Remove broken link to `hgps.ipynb` that was removed yesterday.


cdeil:
Suggest `analysis_mwl` or another short title.

For v1.0 I'd also suggest to change from HAWC spectral points to radio and X-ray data and use Naima, and to have this as the MWL tutorial, showing basics how to work with datasets also elsewhere.

But put what you prefer, you're right, just "joint" isn't descriptive enough.

adonath:
I agree `analysis_mwl.ipynb` is good choice, will use this.

Issue #2593, opened 2019-11-21
Clean-up detect.ipynb


registerrier:
The detect.ipynb still contains many TODOs. It has to be cleaned up.

- The comparison with 3FHL is not really important here. I suggest to remove it.
- Remove exercices

registerrier:
This is resolved thanks to PR #2681 . Closing issue.

PR #2681, opened 2019-12-10
Update detection notebook

QRemy:
Extend the notebook to show more of gammapy.detect,  detail how to :
- perform adaptive smoothing on counts image
- produce 2-dimensional test-statistics (TS)
- run a peak finder to detect point-source candidates
- compute Li & Ma singificance images
- estimate source candidates radius and excess counts

Address issues #2593 and #752

### DIFF
--- a/tutorials/detect.ipynb
+++ b/tutorials/detect.ipynb
@@ -6,16 +6,22 @@
    "source": [
     "# Source detection with Gammapy\n",
     "\n",
-    "## Introduction\n",
+    "## Context\n",
+    "\n",
+    "The first task in a source cataloque production is to identify significant excess in the data that can be associated to unknown sources and provide a preliminary parametrization in term of position, extent, and flux. In this notebook we will use Fermi-LAT data to illustrate how to detect candidate sources in counts images with known background.\n",
+    "\n",
+    "**Objective: build a list of significant excesses in a Fermi-LAT map**\n",
+    "\n",
+    "\n",
+    "## Proposed approach \n",
     "\n",
     "This notebook show how to do source detection with Gammapy using the methods available in `~gammapy.detect`.\n",
-    "\n",
     "We will use images from a Fermi-LAT 3FHL high-energy Galactic center dataset to do this:\n",
     "\n",
     "* perform adaptive smoothing on counts image\n",
     "* produce 2-dimensional test-statistics (TS)\n",
     "* run a peak finder to detect point-source candidates\n",
-    "* compute Li & Ma singificance images\n",
+    "* compute Li & Ma significance images\n",
     "* estimate source candidates radius and excess counts\n",
     "\n",
     "Note that what we do here is a quick-look analysis, the production of real source catalogs use more elaborate procedures.\n",
@@ -47,27 +53,6 @@
     "%matplotlib inline\n",
     "import matplotlib.pyplot as plt\n",
     "import matplotlib.cm as cm\n",
-    "\n",
-    "# defalut matplotlib colors without grey\n",
-    "colors = [\n",
-    "    u\"#1f77b4\",\n",
-    "    u\"#ff7f0e\",\n",
-    "    u\"#2ca02c\",\n",
-    "    u\"#d62728\",\n",
-    "    u\"#9467bd\",\n",
-    "    u\"#8c564b\",\n",
-    "    u\"#e377c2\",\n",
-    "    u\"#bcbd22\",\n",
-    "    u\"#17becf\",\n",
-    "]"
-   ]
-  },
-  {
-   "cell_type": "code",
-   "execution_count": null,
-   "metadata": {},
-   "outputs": [],
-   "source": [
     "from gammapy.maps import Map\n",
     "from gammapy.detect import (\n",
     "    ASmooth,\n",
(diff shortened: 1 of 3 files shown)


tutorials/detect.ipynb
```diff
@@ -8,22 +8,25 @@
...
+    "We will use images from a Fermi-LAT 3FHL high-energy Galactic center dataset to do this:\n",
     "\n",
-    "* produce 2-dimensional test-statistics (TS) images using Fermi-LAT 3FHL high-energy Galactic center dataset\n",
-    "* run a peak finder to make a source catalog\n",
-    "* do some simple measurements on each source\n",
-    "* compare to the 3FHL catalog\n",
+    "* perform adaptive smoothing on counts image\n",
+    "* produce 2-dimensional test-statistics (TS)\n",
+    "* run a peak finder to detect point-source candidates\n",
+    "* compute Li & Ma singificance images\n",
```
registerrier: significance

tutorials/detect.ipynb
```diff
@@ -42,7 +45,21 @@
    "outputs": [],
    "source": [
     "%matplotlib inline\n",
-    "import matplotlib.pyplot as plt"
+    "import matplotlib.pyplot as plt\n",
+    "import matplotlib.cm as cm\n",
+    "\n",
+    "# defalut matplotlib colors without grey\n",
+    "colors = [\n",
```
registerrier: this might be a bit confusing here no?

tutorials/detect.ipynb
```diff
@@ -87,6 +114,40 @@
     ")"
    ]
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Adaptive smoothing\n",
```
registerrier: Maybe this specific question deserves a small notebook on its own. Something like how to produce an adaptively smoothed map? 
We probably need to re-discuss how to present how-to's that do not explicitly belong to a specific tutorial.

tutorials/detect.ipynb
```diff
@@ -139,9 +200,10 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "## Source catalog\n",
+    "## Source candidates\n",
     "\n",
-    "Let's run a peak finder on the `sqrt_ts` image to get a list of sources (positions and peak `sqrt_ts` values)."
+    "Let's run a peak finder on the `sqrt_ts` image to get a list of point-sources candidates (positions and peak `sqrt_ts` values).\n",
+    "The `find_peaks` function perform a local maximun search in a sliding window, the argument `min_distance` is the minimum pixel distance between peaks (smallest possible value and default is 1 pixel)."
```
registerrier: performs a local...

tutorials/detect.ipynb
```diff
@@ -87,6 +114,40 @@
...
+    "images[\"counts\"].plot(add_cbar=True, vmax=10);"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## TS map estimation\n",
+    "\n",
+    "The TS map is computed fitting by a single parameter amplitude fit as described in Appendix A of [Stewart (2009)](https://ui.adsabs.harvard.edu/abs/2009A%26A...495..989S/abstract). The fit is simplified by finding roots of the derivative of the fit statistics (default settings use [Brent's method](https://en.wikipedia.org/wiki/Brent%27s_method))."
```
registerrier: You might expand a bit here to have a more pedagogical tutorial.

tutorials/detect.ipynb
```diff
@@ -181,9 +244,10 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "## Measurements\n",
+    "## Li & Ma significance maps\n",
```
registerrier: Maybe here, the interesting point is rather to go beyond the point source assumption that is made by the `TSMapEstimator`

tutorials/detect.ipynb
```diff
@@ -181,9 +244,10 @@
...
     "\n",
-    "* TODO: show cutout for a few sources and some aperture photometry measurements (e.g. energy distribution, significance, flux)"
+    "Additionaly we can compute significance for an observed number of counts and known background using equation (17) from the [Li & Ma (1983)](https://ui.adsabs.harvard.edu/abs/1983ApJ...272..317L/abstract). We can perform this calculation intergating the counts within different radius. To do so we use an astropy Tophat kernel with the `compute_lima_image` function.\n",
```
registerrier: Here you should say that this is an extension of the Li & Ma formula for known background.

tutorials/detect.ipynb
QRemy: I added a transition comment at the end of the TS map part about extended sources:

> Note that we used the instrument point-spread-function (PSF) as kernel, so the hypothesis we test is the presence of a point source. In order to test for extended sources we would have to use as kernel an extended template convolved by the PSF. Alternatively, we can compute the significance of an extended excess using the Li & Ma formalism, which is faster as no fitting is involve.

tutorials/detect.ipynb
QRemy: Added TS definition and reference

tutorials/detect.ipynb
registerrier: Perfect

Issue #2576, opened 2019-11-20
Separate image_analysis.ipynb in two notebooks


registerrier:
`image_analysis.ipynb` is doing too much at a time. One thing is background normalization techniques when performing an image, the other is source fitting in 2D.

I propose to separate both and to create:
- one notebook explaining background normalization, in particular ring background, possibly with and without HLI.
- one notebook performing source fitting

registerrier:
This is now resolved by PR #2684. Closing.

PR #2684, opened 2019-12-11
Update image analysis tutorials

AtreyeeS:
This pull request cleans up the image analysis tutorials as described in #2576 
I have tried to add some context à la `light_curve_flare` notebook


tutorials/ring_background.ipynb
```diff
@@ -0,0 +1,415 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Ring Background Estimation\n",
+    "\n",
+    "## Context:\n",
+    "Classically, image analysis in IACT have often relied on the ring background model to get a background estimate. This method typically involves extracting a ring around a trial source position to estimate the background strength. To correct for the varying camera acceptance within the field of view, an acceptance correction function is used to compute the normalisation for each position of the ring. To read more about this method, see [here.](https://arxiv.org/abs/astro-ph/0610959)\n",
```
registerrier: You could explain more precisely that when you want to produce a map you have to account for the large residual hadronic background to create an excess map, assumed to be a map go gamma-ray events. In the absence of a solid template bkg model it is not possible to obtain reliable background model a priori. It was often found necessary in classical cherenkov astronomy to perform a local renormalization of the existing templates, usually with a ring kernel. This assumes that most of the events are background and requires to have an exclusion mask to remove regions with bright signal from the estimation.

registerrier: Then you can give an objective:
Create an excess (gamma-ray events) map of MSH 15-52 as well as a significance map to determine how solid the signal is.

tutorials/ring_background.ipynb
```diff
@@ -0,0 +1,415 @@
...
+    "\n",
+    "## Proposed approach:\n",
+    "\n",
+    "The analysis workflow is roughly\n",
+    " - Compute the sky maps keeping each observation separately using the `Analysis` class\n",
```
registerrier: Since you have to perform a loop over images anyway, maybe using the HLI is not needed here.

tutorials/ring_background.ipynb
```diff
@@ -0,0 +1,415 @@
...
+    " - Estimate the background using the `RingBackgroundMaker`\n",
+    " - Compute significance and excess maps using `compute_lima_on_off_image`\n",
```
registerrier: here it is a 'correlated excess map', but that might be too jargon

tutorials/ring_background.ipynb
```diff
@@ -0,0 +1,415 @@
...
+    " \n",
+    "The Li&Ma images thus computed can be used for normal modelling and fitting."
```
registerrier: Not really. Only the normalized background can.

tutorials/ring_background.ipynb
```diff
@@ -0,0 +1,415 @@
...
+    "    r_in=\"0.5 deg\", width=\"0.3 deg\", exclusion_mask=exclusion_mask\n",
+    ")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Create a stacked dataset\n",
+    "Now, we extract the background for each dataset and then stack the maps together to create a sinngle stacked map for further analysis"
```
registerrier: single

tutorials/ring_background.ipynb
```diff
@@ -0,0 +1,415 @@
...
+    "    )  # Ring extracting makes sense only for 2D analysis\n",
+    "    dataset_on_off = ring_maker.run(dataset_image)\n",
+    "    stacked_on_off.stack(dataset_on_off)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "This `stacked_on_off` has `on` and `off` counts and acceptance maps which we will use in all further analysis"
```
registerrier: you should explain what `acceptance_on` and `acceptance_off` stand for.

tutorials/ring_background.ipynb
```diff
@@ -0,0 +1,415 @@
...
+   "source": [
+    "print(stacked_on_off)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Compute Li&Ma maps\n",
+    "We need to convolve our maps with an apprpriate smoothing kernel. Since astropy convolution kernels only accept integers, we first convert our required size in degrees to int depending on our pixel size."
```
registerrier: Those are not Li&Ma maps but correlated excess and significance map.

The former is the integrated excess in circles of a given size, while the latter gives the significance of this excess. This significance is computed according to the Li & Ma expression for ON and OFF Poisson measurements and refer to the stat doc.

tutorials/ring_background.ipynb
```diff
@@ -0,0 +1,415 @@
...
+    "\n",
+    "ax2.set_title(\"Excess map\")\n",
+    "excess_map.get_image_by_idx((0,)).plot(ax=ax2, add_cbar=True)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "It is often important to look at the signficance distribution outside the exclusion region to understand if the exclusion regions were properly estimated.\n",
```
registerrier: to check that the background estimation is not contaminated by gamma-ray events. This can be the case when exclusion regions are not large enough.

tutorials/modeling_2D.ipynb
```diff
@@ -0,0 +1,320 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# To see the best fit values along with the errors\n",
+    "analysis.fit_result.parameters.to_table()"
```
registerrier: How about showing more here. For instance computing and plotting position error contours for the source position? Does it slow down the notebook a lot?

tutorials/modeling_2D.ipynb
```diff
@@ -0,0 +1,320 @@
...
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Modeling and fitting 2D images using Gammapy\n",
+    "\n",
+    "## Prerequisites:\n",
+    " - To understand how a generel modelling and fiiting works in gammapy, please refer to the [analysis_3d tutorial](analysis_3d.ipynb)\n",
+    "\n",
+    "## Context:\n",
+    "Classically, images analysis in IACTs has been done using a ring background analysis. To understand how this works in gammapy, please refer to the [ring_background tutorial](ring_background.ipynb). Recently, with the availabilty of background IRFs with DL3 data, we can directly do spatial modelling in 2D. Of course, you can also extract maps using the ring background method and then use the same for modelling as described here.\n",
```
registerrier: I don't think you need to mention ring analysis here.

registerrier: The context here is that you want the determine the position and morphology of an object. To do so, you don't necessarily have to resort to a full 3D fitting but can perform a simple image fitting, in particular in an energy range where the PSF does not vary strongly or if you want to explore a possible energy dependance of the morphology.

registerrier: I would give the objective. Here localize a source and/or constrain its morphology.

tutorials/ring_background.ipynb
AtreyeeS: I thought so initially as well, but without the HLI, it means exposing the `DataStore`, the `SafeMaskMaker` etc, which seems to distract the main focus of the notebook. So, it seemed clearer to stick with the HLI, and add an extra loop. 

Of course, I can change this if you prefer.

tutorials/modeling_2D.ipynb
AtreyeeS: On my machine, adding a call to `analysis.fit.minos_contour` increases the execution time form 20 secs to 33 secs. Perhaps this should better reside on the models notebook?

AtreyeeS:
Thanks @registerrier for your comments. I have also made a small change in `MapDatasetOnOff`. Currently, `MapDatasetOnOff.name` if forced to `acceptance_off` which is quite confusing.

tutorials/ring_background.ipynb
registerrier: OK. I guess we will have to update to a new solution as soon as the HLI is adapted to the ring background.

tutorials/modeling_2D.ipynb
registerrier: For how many points?

AtreyeeS: The default, 10 points

registerrier: OK. Then leave as is. Just mention the modeling/fitting notebook for now.

PR #2690, opened 2019-12-15
Update models gallery notebook

QRemy:
- cleanup, restructure, and add more comments in the spatial and spectral models sections
- illustrate the to_region() and position_error() methods for spatial models and show how to export these as ds9 regions files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -66,7 +66,8 @@ tutorials/model-best-fit.yaml
 tutorials/crab-3datasets
 tutorials/analysis_1
 tutorials/analysis_3d
-tutorials/config.yaml
+tutorials/*.yaml
+Tutorials/*.reg
 gammapy-datasets
 temp
 
--- a/gammapy/catalog/fermi.py
+++ b/gammapy/catalog/fermi.py
@@ -69,7 +69,7 @@ class SourceCatalogObjectFermiBase(SourceCatalogObject, abc.ABC):
             ss += self._info_more()
         if "position" in ops:
             ss += self._info_position()
-            if not self._is_pointlike:
+            if not self.is_pointlike:
                 ss += self._info_morphology()
         if "spectral" in ops:
             ss += self._info_spectral_fit()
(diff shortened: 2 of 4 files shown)


QRemy:
If I remove the 3D models, the serialisation, the to_region(), and the custom models sections to other notebook or howto, there is almost nothing left in this notebook and it could be simply moved to https://docs.gammapy.org/0.15/modeling/index.html that already contains some details about models, except if the goal is to have a 100% coverage of the notebooks, is it ?

Generally I would prefers to have most of the information on one topic in one place rather than splitted accros several notebooks and howto (these could just be added latter to point toward notebooks or API reference pages,  not a priority).
In that case, my idea was to expose all models (including 1D, 2D, 3D)  in this notebook and show their "optionnal" functionality , as custom models, to_regions, plotting, models serialization...
While the modeling notebook would be more focused on using the models with datasets in a fit, explain datasets serialisation, and play with results: attach covariance to models; get errors; plot butterfly, confidence contours, likelihood profile, residuals... actually there is a lot of overlap with analysis 2 notebook.

@adonath The bug in Travis seems to appear for astropy version 3.2.2/3.2.3, it works for version 3.2.1 should we go back to this one ?

registerrier:
While I see the advantage of keeping most of the things in one place, we must keep in mind that tutorials are there to help users understand how gammapy works. It is also important that information is easily accessible. I think a bit of duplication is not an issue but is rather expected.

For models, maybe it is feasible to do a general overview of how models work, an exhaustive list of models (presenting their parameters, a plot and their YAML definition) as well as a presentation of utility features all at once.  But I am afraid that this becomes way too long.

For the model gallery itself, indeed it is not clear that we need it in the form of a notebook. It could go in a dedicated rst page since in itself it is not really a tutorial. This is then another option.

adonath:
@QRemy I attached a commit to fix the astropy error, it turned out out there was a simple work around. I'll wait for Travis-CI and then merge...

registerrier:
Thanks @QRemy . Fail is unrelated. Merging now.

PR #2697, opened 2019-12-24
Add FoVBackgroundMaker class

registerrier:
**Description**

This pull request introduces a `FoVBackgroundMaker` class. It is very simple. It takes an exclusion mask on `__init__`. `FoVBackgroundMaker.run(dataset)` simply performs a fit of the model. The assumption here is that it only consists of the `background_model`.

Tests have been included, in particular one checking that in case of non-convergence the parameters are the initial ones. In this case a simple `log.warning` is raised. (this should be tested properly). It is not clear that this is the best behavior. Because ideally the dataset should be excluded.

**Dear reviewer**

please comment on the choice of for non-convergence.

### DIFF
--- a/gammapy/cube/fov.py
+++ b/gammapy/cube/fov.py
@@ -1,10 +1,12 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 """FoV background estimation."""
+import logging
 from gammapy.maps import Map
 from gammapy.modeling import Fit
 
 __all__ = ["FoVBackgroundMaker"]
 
+log = logging.getLogger(__name__)
 
 class FoVBackgroundMaker:
     """Normalize template background on the whole field-of-view.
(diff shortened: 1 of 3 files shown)


gammapy/cube/fov.py
```diff
@@ -0,0 +1,97 @@
...
+            vals = self.exclusion_mask.get_by_coord(coords)
+            mask_map.data += vals
+
+        return mask_map.data.astype("bool")
+
+    def _fit_bkg(self, dataset):
+        """Fit the FoV background model on the dataset counts data"""
+
+        # Here we assume that the model is only the background model
+        # TODO : freeze all model components not related to background model?
```
adonath: I think freezing all components instead of the background is probably the better (safer) default behaviour, especially if the default method is `fit` as well.

registerrier: That's right. 
I changed the default to `scale` which is probably more natural.
I added the parameters freezing in case a `SkyModel` is set on the dataset. Parameters are then unfrozen. A test is added to check for this.

registerrier:
Modifications to the tutorials will be made in a further pull request. 
All tests pass. Merging.

PR #2702, opened 2020-01-07
Fix typo in analysis_1.ipynb

samcarter:
typo


adonath:
Thanks a lot @samcarter!

PR #2705, opened 2020-01-07
Rename EnergyDispersion to EDispKernel

adonath:
**Description**

This pull request renames the `EnergyDispersion` class to `EDispKernel`. The main motivation is consistency with `PSFKernel` and clarify that `EDispKernel` is and already integrated redistribution matrix. We could also consider to include RMF in the name somehow, but I slightly prefer to introduce consistency with the `PSFKernel` in the naming.


adonath:
Thanks @registerrier! I agree this needs an entry in the changelog. Typically I have done this a few days before the release in a uniform way. There is the option that we add changelog entries "on the fly" in each PR, but even in this case I would probably need to check all the entries before the release. So I would keep the current way, that I add those things a few days before the release and ask for feedback.

PR #2707, opened 2020-01-09
Implement MapDatasetOnOff.to_spectrum_dataset() and .cutout()

registerrier:
**Description**

This pull request introduces the `MapDatasetOnOff.to_spectrum_dataset()` to properly compute the `SpectrumDatasetOnOff` with the `counts_off`, `acceptance`, and `acceptance_off` computed. They are evaluated such that excess number is conserved. This is tested using the hgps map.
The typical usage would be to produce profiles from a list of regions or possibly to build a spectrum from a true ON OFF dataset.

While doing the PR I noticed that `.cutout()` is not properly overloaded. Currently it does not return a `.acceptance` and `.counts_off` etc. This PR corrects for this by adding an overloaded `cutout` for `MapDatasetOnOff` with an associated test. (Note that `MapDataset.cutout` is not explicitly tested by tested through `MapEvaluator`.

**Dear reviewer**

this should be ready to merge.

### DIFF
--- a/gammapy/cube/fit.py
+++ b/gammapy/cube/fit.py
@@ -16,7 +16,7 @@ from gammapy.maps import Map, MapAxis
 from gammapy.modeling import Dataset, Parameters
 from gammapy.modeling.models import BackgroundModel, SkyModel, SkyModels
 from gammapy.modeling.parameter import _get_parameters_str
-from gammapy.spectrum import SpectrumDataset
+from gammapy.spectrum import SpectrumDataset, SpectrumDatasetOnOff
 from gammapy.stats import cash, cash_sum_cython, wstat
 from gammapy.utils.random import get_random_state
 from gammapy.utils.scripts import make_path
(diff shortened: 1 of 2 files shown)


gammapy/cube/fit.py
```diff
@@ -1351,6 +1351,96 @@ def from_hdulist(cls, hdulist, name=""):
...
+            If only one value is passed, a square region is extracted.
+        mode : {'trim', 'partial', 'strict'}
+            Mode option for Cutout2D, for details see `~astropy.nddata.utils.Cutout2D`.
+
+        Returns
+        -------
+        cutout : `MapDatasetOnOff`
+            Cutout map dataset.
+        """
+        kwargs = {"gti": self.gti}
```
adonath: I think `kwargs` I unused here, just remove...

registerrier: Done

PR #2710, opened 2020-01-11
Introduce gammapy.datasets sub-module?

adonath:
**Description**

This pull request explores the possibility of introducing a `gammapy.datasets` sub-module, which contains all `Dataset` class. The idea for this came up in [PIG 16](https://docs.gammapy.org/0.15/development/pigs/pig-016.html). My impression is that it clarifies the package structure a lot. It is another step into direction of structuring the package by concept, rather the 1D, 3D analyses etc. The `docs/datasets` folder gives a natural place to explain about the concept of datasets in a more detailed way then the overview page. This was missing before.

To avoid circular imports, a few delayed imports were introduced (while other could be removed). The delayed imports appear for now in the `spectrum/make.py` and `cube/make.py` files, where the dataset classes are need to create the final class to be returned.

Any opinions @registerrier, @Bultako, @luca-giunti,  @AtreyeeS, @QRemy?


adonath:
Discussed with @registerrier offline: I'll write this proposal up in a short PIG together with a few other proposed changes to the package structure.

adonath:
This PR already has to many merge conflicts, that I want to resolve them. I opened #2720 to discuss the package structure, but not before v0.16.

adonath:
Thanks @registerrier, for commenting I'll consider the comments in the new PR I will open once PIG 19 has settled.

PR #2716, opened 2020-01-15
Move parameter link reference to a separate key 

QRemy:
-Move parameter link reference to a separate key for serialisation instead of being  encapsulated in the name
-keep it separate from parameter instance id so it is more readable and remains the same than user define linkage in yaml file


adonath:
Thanks @QRemy, looks good to me now. I've merged https://github.com/gammapy/gammapy-extra/pull/131 as well, so we should get a green Travis-CI on master...

PR #2718, opened 2020-01-16
Adapt TSMapEstimator to take a MapDataset as input

registerrier:
**Description**

This pull request adapts the `TSMapEstimator` to take a `MapDataset` as argument. This requires an adaptation since 2D maps with an energy axis are needed. Typically one would call it after `dataset.to_image()`.
An improvement over the previous version is that it now uses `npred` rather than the simple background, so that one can compute the TSMap of residuals from a model.

The detect.ipynb has been adapted. Since the `ASmooth ` and Li & Ma significance parts are not adapted yet, they have been temporarily removed and will be put back once these parts are adapted as well.

**Dear reviewer**

should be ready for review.

### DIFF
--- a/gammapy/detect/test_statistics.py
+++ b/gammapy/detect/test_statistics.py
@@ -235,6 +235,9 @@ class TSMapEstimator:
             sqrt_ts = np.where(ts > 0, np.sqrt(ts), -np.sqrt(-ts))
         return map_ts.copy(data=sqrt_ts)
 
+    @staticmethod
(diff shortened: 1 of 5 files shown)


registerrier:
A notebook was failing because of a `Wcsgeom.__eq__` comparison called for `Map` arithmetics.
The issue was due to the fact that `astropy.wcs.WCS.wcs.compare` is by default enforcing strict comparison of ancillary keywords. I had to find the correct (but not really documented) keyword that removes this strict comparison.

Issue #1304, opened 2018-02-13
Location of I/O code in Gammapy


cdeil:
@MaxNoe in https://github.com/gammapy/gammapy/issues/1265#issuecomment-365260031 made this comment that I think we should discuss separately (i.e. in this new issue):

> I think that core functionality must be completely agnostic of file formats.

> Having an IO layer around the core functionality is nice to have and supporting the data structures of multiple existing telescopes and also new, better formats will be easy as soon as all library functionality does not depend on files but a common in-memory data representation.

> So everything outside gammapy.io or something similar should not have anything to do with file formats. Functions should only get astropy Quantity or Tables or plain numpy arrays.

> Writing high level science tools then becomes easy. But mangling IO and computation just makes it much more complex, harder to use, harder to maintain and reduces the number of people and experiments that can effectively work with gammapy.

---

Just to summarise what we have in Gammapy, and to add my 2 cents.

In Gammapy, we almost always have the pattern that `__init__` takes simple objects, mostly Numpy arrays, or Astropy Quantity, SkyCoord, Time, Table, WCS objects, and those are used in analysis. I think we're in agreement that this is good.

I/O is currently often located in classmethods called `read` and `write` and then for the existing FITS formats in classmethods `from_hdulist` and `to_hdulist` that are located on the classes that represent the in-memory data structure. This is the point we should discuss, @MaxNoe is proposing to split all that code out into `gammapy.io`. Presumably that's how it's done in [ctapipe.io](https://cta-observatory.github.io/ctapipe/io/index.html)?

Good examples for what we have in Gammapy are the [maps](http://docs.gammapy.org/dev/maps/index.html), or to take a very simple example for discussion, see how the [gammapy.data.PointingInfo](http://docs.gammapy.org/dev/api/gammapy.data.PointingInfo.html) is implemented and does I/O via classmethods attached to the container class.

(I've already admitted that the `DataStoreObservation` is not OK and should be rewritten as you propose; for that, please comment on #1303 if you have any suggestions how it should be done in detail.)

---

@mackaiver @MaxNoe @woodmd @registerrier @kosack @adonath @joleroi  - do you think we should split off all I/O code into `gammapy.io`, or stick with the I/O coding pattern we have now, or some mix or something different?

Note that the suggestion by @mackaiver in #1285 is very similar - there the question is whether the existing `plot` methods in Gammapy should be split out into separate plot functions / classes.

---

Personally I'm a bit in favour of having I/O in classmethods on container classes.

[...]

PR #2720, opened 2020-01-17
PIG 19 - Package structure follow up

adonath:
**Description**

This pull request contains a draft for a PIG with a few follow-up proposal to restructure the sub-packages in Gammapy. Mainly introducing a `gammapy.datasets` and `gammapy.makers` sub-package. 

**Dear reviewer**
Please take a look at the first draft and comment. Especially the introduction of ``gammapy.estimators`` and ``gammapy.visualisation`` is something I would like to have feedback on.

### DIFF
--- a/docs/development/pigs/index.rst
+++ b/docs/development/pigs/index.rst
@@ -33,5 +33,6 @@ label`_ .
     pig-014
     pig-016
     pig-018
+    pig-019
 
 .. _pull requests with the "pig" label: https://github.com/gammapy/gammapy/issues?q=label%3Apig
--- a/docs/development/pigs/pig-018.rst
+++ b/docs/development/pigs/pig-018.rst
@@ -291,7 +291,7 @@ Decision
 The PIG was extensively discussed and received a lot of feedback on `GH 2463`_.
 The main suggestions received were incorporated. There was some controversy e.g.
 whether we should have more shorter pages and notebooks, or fewer longer ones.
-This PIG was accepted on Nov 6, 2016, although we'd like to note that the
+This PIG was accepted on Nov 6, 2019, although we'd like to note that the
 outline and changes described above aren't set in stone, we expect the
 documentation to evolve and improve in an interative fashion over the coming
 weeks, but also in 2020 and after.
--- a/docs/development/pigs/pig-019.rst
+++ b/docs/development/pigs/pig-019.rst
@@ -3,7 +3,7 @@
 .. _pig-019:
 
 ********************************************
-PIG 19 - Gammapy package structure Follow Up
+PIG 19 - Gammapy package structure follow up
 ********************************************
 
 * Author: Axel Donath, Régis Terrier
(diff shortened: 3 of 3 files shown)


registerrier:
Maybe issue #1304 could also be to consider for this specific proposal. There could be some `gammapy.io` sub package.

facero:
Hi,
I think this is the good timing to do all those changes (before 1.0 rc).
I like `gammapy.estimator` (or `evaluator` ?)  and `visualisation`.
I agree with most changes and would suggest that PSFKernel and EDispKernel, EDispMap and PSFMap go to `irf` instead of `maps`.
Concerning the `gammapy.core` I'm not sure this is a good idea because there would be discussion about what's core and what's not. In addition, this definition could evolve in the future ...
When you say sub-package can be "resolved" I imagine you mean deleted (not a bad wording just me trying to be sure).

AtreyeeS:
Hi! Most of the proposals are totally non-controversial, and doing it for 0.17 would be nice.

- Having a common `gammapy.datasets` and `gammapy.makers` is definitely a good idea - it makes things cleaner from a user perspective as well.

- Regarding moving `EDispMap` and `PSFMap`, I agree with @facero that they should better reside on `gammapy.irf` than `gammapy.maps` -  I tend to think of`maps` as generalised containers than can be adapted according to need, whereas the `EDispMap` and `PSFMap` have a specific implementation with hard coded axes, etc.

- While I agree having a common visualisation sub-package would be nice in future, I think the present `Map.plot()` or `IRF.peek()` is very convenient. I suppose in future it would be along the lines of `gammapy.visualisations.peek(aeff)` or `gammapy.visualisations.plot(aeff)`, which is also OK.
But perhaps these will move later, and for now some plotting will still reside on the objects? Having two different APIs in 1.0 might be a bit unpleasant, no?

- I have no strong opinion regarding `estimators`. I will indeed simplify our code base a bit, but a user probably thinks in terms of doing a spectral analysis and getting flux points, or a temporal analysis and getting lightcurves or searching for new sources - as per a science interest. So our present `gammapy.time` or `gammapy.spectrum` or `gammapy.detect` is possibly more intuitive. But I agree, at present we do call `FluxPointEstimator` even for a 3D analysis, so the new proposed approach might help. We should document it nicely, though!

Bultako:
> Maybe issue #1304 could also be to consider for this specific proposal. There could be some `gammapy.io` sub package.

With the addition of a `datasets` sub package, I wonder how intuitive could be for new users to separate `data` from `datasets. I think many of the features in `data` sub package are related with IO (only DL3 format now). If we add a new `io` sub package, maybe we could rename `data` into `io` and  then add the other i/o features there. (i.e. datasets, models, HLI config, etc.)

adonath:
Thanks @facero and @AtreyeeS for the feedback. I agree now that `gammapy.irf` is probably the better place for the `EDispMap` and `PSFMap` and corresponding kernels. I'll adapt the section in the PIG accordingly.

@facero Yes, longterm "resolve" means "delete". What I wanted to express with "resolve", was that there is maybe little functionality that must be moved somewhere else before the sub-package can be deleted. 

@AtreyeeS I agree 100% that methods like `Map.plot()` and `IRF.peek() `are convenient and useful and I don't propose to remove them from the API. The proposal here is to just maintain the functionality longterm in a different place, because it's expected, that plotting functionality will grow. I think even for a simple `MapDataset` there are probably ~10-20 interesting plots to make (npred spectra by source component, residuals, residual significance distribution, npred images, counts profiles, excluded significance images, ts maps etc.). What I want to avoid is that we get all these `MapDataset.plot_something()` functions, which add up to a few hundred lines of code. Instead I think it's nicer if we have something like a `MapDatasetPlotter()` in future, where we can bundle all the plotting related functionality. 

@Bultako I agree that `gammapy.data` is not well defined as a subpackage. It mainly exposes DL3 data access, with part of the I/O functionality living in `gammapy.irf`. I guess it's certainly an option to try to improve this as well. A simple solution could be ti just rename it to `gammapy.dl3` or `gammapy.dl3data`, possibly `gammapy.observation` as well. I'll think about it once more...

PR #2723, opened 2020-01-20
Add time scale info in GTI.__repr__

registerrier:
**Description**

This pull request adds the time scale (time standard) information to the `GTI.__repr__` to avoid issues when comparing GTI times to `Time` objects defined with the `utc` time scale.

**Dear reviewer**


adonath:
Thanks @registerrier!

PR #2727, opened 2020-01-21
Rename SkyModels to Models

adonath:
**Description**

This pull request renames the current `SkyModels` class to `Models`. The motivation is a step in preparation to make `Models` also include `BackgroundModel` objects. It also then follows the general scheme of having a `Model` base class and a `Models` container class.

**Dear reviewer**
Please let me know if you see any problems with this backwards incompatible change.


adonath:
@registerrier This is the start of a few changes we discussed with @QRemy and Christoph before Christmas, the idea is to unify the `SkyModel` and `BackgroundModel` handling into a single global `Models` objects.

PR #2728, opened 2020-01-22
Add auto-generated unique names

QRemy:
Implement some features discussed in #2517 in particular from this [comment](https://github.com/gammapy/gammapy/issues/2517#issuecomment-556993333)
- add auto-generated unique names for model and dataset on init and copy
- force unique names on datasets and skymodels


adonath:
One more thing: 

Can you please add a `Datasets.names` and `SkyModels.names` method, that just returns a list of names of the objects contained? Might be helpful in some cases...

adonath:
Travis-CI fails are unrelated...

PR #2730, opened 2020-01-23
Add an extended source spectral analysis tutorial

registerrier:
**Description**

This pull request adds a tutorial explaining how to perform extended source analysis in gammapy. It uses RX J1713 data. It runs in 9sec on my machine. It follows the description introduced some time ago in PR #2672. 

Maybe some discussion could be added on the question of handling of observations with no OFF regions because it is likely to be more frequent for extended objects.

**Dear reviewer**

This should be ready for review.

### DIFF
--- a/tutorials/extended_source_spectral_analysis.ipynb
+++ b/tutorials/extended_source_spectral_analysis.ipynb
@@ -279,7 +279,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "stacked.counts.sum_over_axes().plot()#smooth(width=\"0.05 deg\").plot()\n",
+    "stacked.counts.sum_over_axes().smooth(width=\"0.05 deg\").plot()\n",
     "on_region.to_pixel(stacked.counts.geom.wcs).plot()"
    ]
   },
--- a/tutorials/extended_source_spectral_analysis.ipynb
+++ b/tutorials/extended_source_spectral_analysis.ipynb
@@ -1,0 +1,461 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Spectral analysis for extended sources with gammapy\n",
+    "\n",
+    "## Prerequisites:\n",
+    "\n",
+    "- Understanding of spectral analysis techniques in classical Cherenkov astronomy.\n",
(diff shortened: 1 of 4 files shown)


adonath:
Thanks @registerrier, I agree this is a very useful tutorial (I already know that a few users wanted to do such an analysis...).

I'll do a more detailed review this afternoon, but for now I have one early comment:

I guess in practice it might be hard to find any reflected regions for extended sources in the Galactic Plane (this was exactly the problem we had for the HGPS)? So one could show a slightly different approach in addition:
- Extract a 3D MapDataset
- Use FoVBackgroundEstimator 
- Create a `SpectrumDataset` with fixed background
- Do the fit and compare

What do you think?

registerrier:
It is easy indeed to modify the notebook to perform a fit with the `SpectrumDataset`. If you simply add `"background"` to the `MapDatsetMaker` init and  comment the line calling the `bkg_maker.run(dataset,obs)` you have a fit with a fixed background model. And it works. The result is rather good, but this is not really surprising given the brightness of this source.

We could indeed advertise other ways to perform fits that are made possible with gammapy but we probably should test them a bit or ask people to test these ideas.

tutorials/extended_source_spectral_analysis.ipynb
```diff
@@ -0,0 +1,461 @@
...
+    "We first set the datastore and retrieve a few observations from our source."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "datastore_path = \"/Users/terrier/Code/gammapy-dev/gammapy-datasets\"\n",
```
bkhelifi: The path should be modified to maybe $GAMMAPY_DATA

registerrier: Indeed. Thanks.

adonath:
@bkhelifi With the current implementation of the `Analysis` class we will only have a limited selection of analysis workflows available. So we cannot attach the corresponding YAML files to the notebooks, because those (special) analysis workflows are not possible using the HLI (and as long as we have the hard-coded workflows, they will never be IMHO...the options of combining makers and create custom data reduction chains are just endless...sow we stick to the 2nd level API for that).

registerrier:
Travis CI fail is unrelated. Merging.

I propose to move the discussion about what to have as tutorials in HLI vs mid-level API to an issue.

Issue #2736, opened 2020-01-27
Crash of print(dataset) when there is no mask_safe


bkhelifi:
**Gammapy version**
V0.15

**Bug description**
print(dataset)
  File "/home/khelifi/MesProgrammes/anaconda3/envs/gammapy-0.15/lib/python3.7/site-packages/gammapy/cube/fit.py", line 129, in __str__
    mask = self.mask_safe.reduce_over_axes(np.logical_or).data
AttributeError: 'NoneType' object has no attribute 'reduce_over_axes'

Which is expected if self.mask_safe is None!

**Expected behavior**
No crash

**Proposal**
Add a test if self.safe_mask exists before making any operation on it!

registerrier:
The issue is not only with `__str__`. Looking at the code, we see that `MapDataset.stack` and `MapDatasetOnOff.stack` also call `mask_safe` without checking properly. 

We can either:
- Create a `safe_mask` property that returns a `Map` with the correct geom full of 1. But this implies that a geom can be found.
- Add a `if .. else` statement in the missing cases. But it seems that we miss a general set of tests for the various combinations of existing or non-existing properties.

adonath:
This specific issue is resolved, but let's keep it open as reminder for systematic tests...

Issue #2734, opened 2020-01-24
MapDatasetOnOff.to_image() does not work properly


registerrier:
`MapDatasetOnOff.to_image()` is not overloaded. Therefore its members are not processed accordingly. This should be implemented.

registerrier:
This is solved by PR #2743. Closing.

PR #2743, opened 2020-01-28
Add a MapDatasetOnOff.to_image() method

registerrier:
**Description**

This pull request adds a `MapDatasetOnOff.to_image()` to `MapDatasetOnOff`. Previously it was only using the parent `to_image` method. Now it is properly overloaded to sum `counts_off`, `acceptance`, `acceptance_off`.
I also added `name` in many docstrings where it was missing. 
I also added a change to `MapDataset.__str__` to test if `mask_safe` is set or not and added a test for that.

**Dear reviewer**

This should solve issue #2734 . Part of issue #2736 as well. The `stack` methods should be adapted to deal with possible absence of `mask_safe`

### DIFF
--- a/gammapy/cube/fit.py
+++ b/gammapy/cube/fit.py
@@ -1483,7 +1483,29 @@ class MapDatasetOnOff(MapDataset):
         """
         dataset = super().to_image(spectrum, name)
 
-        return dataset
+        kwargs = {}
+        if self.counts_off is not None:
+            kwargs["counts_off"] = self.counts_off.sum_over_axes(keepdims=True)
+
+        if self.acceptance is not None:
+            kwargs["acceptance"] = self.acceptance.reduce_over_axes(func=np.mean, keepdims=True)
+            background = self.background.sum_over_axes(keepdims=True)
+            kwargs["acceptance_off"] = (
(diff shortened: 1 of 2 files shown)


gammapy/cube/fit.py
```diff
@@ -1431,6 +1464,63 @@ def cutout(self, position, width, mode="trim"):
...
+
+        if self.acceptance is not None:
+            acceptance = self.acceptance * mask_safe
+            kwargs["acceptance"] = acceptance.sum_over_axes(keepdims=True)
+            background = self.background * mask_safe
+            background = background.sum_over_axes(keepdims=True)
+            kwargs["acceptance_off"] = (
+                kwargs["acceptance"] * kwargs["counts_off"] / background
+            )
+        kwargs["counts"] = dataset.counts
```
adonath: For the `SpectrumDatasetOnOff` I have introduced `.from_spectrum_dataset(dataset, acceptance, acceptance_off)` at some point. This might be helpful here as (and probably in the ring background estimators as well...)

registerrier: Yes absolutely. I propose to add this in a future PR and adapt there.

Issue #2678, opened 2019-12-10
Add `append` and `extend` methods to `data.Observations` (and other container classes)


lmohrmann:
It would be very helpful if the `data.Observations` container class had methods to add additional observations, like `append` and `extend`. It was previously possible to achieve this by accessing the `list` attribute of this class, but that seems to have been removed. It's still possible by accessing the `_observations` attribute, but that is more of a workaround really.

I know the question of how container classes should be implemented in Gammapy has already been discussed in the past, although I didn't manage to find the corresponding issue. It seems the general idea is to use `collections.abc.Sequence`, but that doesn't seem to provide methods to add to the collection (or I simply haven't discovered them?)

adonath:
Thanks for opening a reminder issue again @lmohrmann! I think the old issue was actually the following PR https://github.com/gammapy/gammapy/pull/2021. What I didn't notice was, that the PR was not fully addressed by #2102. 

For Gammapy v0.15 (in https://github.com/gammapy/gammapy/commit/fa03086418a9e74c5c247d8879a2fdcd57b5eeaa) we introduced the inheritance from `collections.abc.Sequence` for the following classes: 
- `Datasets`
- `SkyModels`
- `Observations`
- `Parameters`

All of these container classes are now instantiated with a list of objects and just offer a minimal API to access the data in it. For now all of these containers are not mutable, i.e. cannot be modified in place. I agree that for all of these classes (except the `Parameters`), it would be useful to offer a minimal API to extend, append or delete items in the container. I'd propose the following:

For `Datasets`, `SkyModels` and `Observations` we change the inheritance to a [MutableSequence](https://docs.python.org/3/library/collections.abc.html). This requires implementing an `.insert()`, `__delitem__()` and `__setitem__()` methods and then offers `.append()`, `.extend()`, `.pop()`, `.reverse()`, `.remove()` as mixing methods. What do you think @registerrier?

adonath:
Addressed in #2744.

Issue #2638, opened 2019-11-29
Add Observation base class


cdeil:
I noticed in  https://docs.gammapy.org/dev/api/gammapy.cube.MapDatasetMaker.html that all the docstrings say that they support `DataStoreObservation`.

Really now we have two classes that are (or should be) supported here and in other places where an observation is used:

- https://docs.gammapy.org/dev/api/gammapy.data.Observation.html
- https://docs.gammapy.org/dev/api/gammapy.data.DataStoreObservation.html

I suggest to resolve this issue by introducing a base class, like this:
```
class Observation:
    """Observation base class."""
    # empty, nothing for now
    # will grow methods, might become an abc.ABC

class MemoryObservation(Observation):
    # exactly what we have now

class DataStoreObservation(Observation):
    # exactly what we have now
```
and to adapt docstrings like in `MapDatasetMaker` (and isinstance checks if we have any).

That's something that can be done in ~ 1 hour, and IMO is the right, achievable solution for v1.0.
The `Observation` and `Observations` parallels the design of `Dataset` and `Datasets`, where `Dataset` is a small base class: https://docs.gammapy.org/dev/api/gammapy.modeling.Dataset.html

@AtreyeeS @adonath @registerrier  - Thoughts?

Anyone willing to send a PR? (otherwise I'd be happy to do it, if it's agreed that it's an improvement)

adonath:
Solved in #2753.

PR #2744, opened 2020-01-28
Change base class for observation(s)

QRemy:
- add Observation base class for DataStoreObservation and MemoryObservation as suggested in #2638
- change Observations inheritance to MutableSequence to gain .append(), .extend(), .pop(), .remove() methods as discussed in #2678

### DIFF
--- a/gammapy/data/observations.py
+++ b/gammapy/data/observations.py
@@ -21,11 +21,11 @@ __all__ = ["Observation", "DataStoreObservation", "MemoryObservation", "Observat
 log = logging.getLogger(__name__)
 
 
-class Observation:
+class ObservationBase:
     """Observation base class."""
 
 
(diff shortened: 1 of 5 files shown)


gammapy/data/observations.py
```diff
@@ -306,10 +310,16 @@ def __init__(self, observations=None):
...
-        if isinstance(key, str):
-            key = self.ids.index(key)
+        return self._observations[self._get_idx(key)]
+
+    def __delitem__(self, key):
+        del self._observations[self._get_idx(key)]
 
-        return self._observations[key]
+    def __setitem__(self, key, obs):
+        self._observations[self._get_idx(key)] = obs
```
adonath: Maybe add an `isinstance(obs, (Observation, DataStoreObservation))` check here.

PR #2745, opened 2020-01-28
Change SkyModels and Datasets inheritance to MutableSequence

QRemy:
- change SkyModels and Datasets inheritance to MutableSequence in order to gain .append(), .extend(), .pop(), .remove() methods as discussed in #2678

### DIFF
--- a/gammapy/modeling/datasets.py
+++ b/gammapy/modeling/datasets.py
@@ -86,7 +86,7 @@ class Dataset(abc.ABC):
         return residuals
 
 
-class Datasets(collections.abc.Sequence):
+class Datasets(collections.abc.MutableSequence):
     """Dataset collection.
 
     Parameters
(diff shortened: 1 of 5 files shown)


gammapy/modeling/models/core.py
```diff
@@ -182,22 +183,43 @@ def __str__(self):
 
     def __add__(self, other):
```
adonath: I think the `__add__` method could call `self.extend()` and `self.append()` now.

gammapy/modeling/models/core.py
```diff
@@ -182,22 +183,43 @@ def __str__(self):
...
-            return self._models[val]
-        elif isinstance(val, str):
+    def __getitem__(self, key):
+        return self._models[self._get_idx(key)]
+
+    def __delitem__(self, key):
+        del self._models[self._get_idx(key)]
+
+    def __setitem__(self, key, model):
+        if model.name in self.names:
```
adonath: I think we check the mode instance here like `if isinstance(model, (SkyModel, SkyDiffuseCube))` and throw an error else.

gammapy/modeling/models/core.py
```diff
@@ -182,22 +183,43 @@ def __str__(self):
...
+            raise (ValueError("Model names must be unique"))
+        self._models[self._get_idx(key)] = model
+
+    def insert(self, idx, model):
+        if model.name in self.names:
```
adonath: Same here, please add an `isinstance()` check.

gammapy/modeling/datasets.py
```diff
@@ -257,16 +257,32 @@ def info_table(self, cumulative=False):
...
-            return self._datasets[val]
-        elif isinstance(val, str):
+    def __getitem__(self, key):
+        return self._datasets[self._get_idx(key)]
+
+    def __delitem__(self, key):
+        del self._datasets[self._get_idx(key)]
+
+    def __setitem__(self, key, dataset):
+        if dataset.name in self.names:
```
adonath: Maybe add an `isinstance(dataset, Dataset)` check here?

gammapy/modeling/datasets.py
```diff
@@ -257,16 +257,32 @@ def info_table(self, cumulative=False):
...
+            raise (ValueError("Dataset names must be unique"))
+        self._datasets[self._get_idx(key)] = dataset
+
+    def insert(self, idx, dataset):
+        if dataset.name in self.names:
```
adonath: Also add an `isinstance()` check here?

gammapy/modeling/models/core.py
QRemy: Not changed, as we want to return a new object rather than updating the first one

Issue #2688, opened 2019-12-13
Default value for SafeMaskMaker and etrue in the HLI


JouvinLea:
Hi all,
Last week, I encountered some issues running 1D analysis from the HLI for two reasons:
- 1) The binning in etrue was not propagated from the config file to the spectrum extraction where you adopted some hardcoded value. This is now fixed by @registerrier (https://github.com/gammapy/gammapy/pull/2673)

- 2) In the HLI, you are also adopting a default method to compute the safemaskmaker for the 1D extraction. I personally don't use this functionality for MAGIC data in my scripts not using the HLI. So I need a way to not compute it from the HLI as well.
I started coded something to get rid of the issue but I know it is not the right way to go. Maybe we should have also a larger discussion on what should be use as default value for etrue or for this safeMakerMask.
Personally for the safeMakerMask I will add a method in the class that if it is equal to None it doesn’t do anything and I will use that as default for the HLI. We can add a member, methods_safe_mask_maker, in the class DatasetsConfig of the HLI.
Then I will make it very clear in the notebook of the HLI for 1D analysis how to modify this safemaskmethod. Also we add to add a notebook for the HLI for 1D analysis.

What do you think? Any opinion on that?

adonath:
Thanks @JouvinLea, you proposal goes in the same direction that @luca-giunti suggest.

The solution is just to expose all configuration from the `SafeMaskMaker` and if no config is present, no safe mask is computed at all.

In the meantime you can always "delete" the safe mask from the dataset after data reduction by setting it to `None` like so:
```python3
for dataset in analysis.datasets:
    dataset.mask_safe = None
```
The safe mask is only applied to the data during stacking. For joint analyses you can always modify it any time later. If needed you can also re-stack.

adonath:
Fixed in #2747 and #2673.

Issue #2704, opened 2020-01-07
Allow to pass fixed offset to SafeMaskMaker?


lmohrmann:
I often need to compute energy thresholds (say, based on the energy bias) for a given specific offset angle (e.g. 2 degrees). Currently, it is only possible to specify a certain position for the evaluation of the energy bias, but in particular if one is dealing with extended sources I think it's a valid use case to evaluate the bias at a fixed offset.

It's of course possible to do this "by hand", but it would be a lot more convenient if the `SafeMaskMaker` could do this.

I realize that adding this possibility adds quite a bit of complexity, since the user would need to specify by which method the bias should be evaluated (position of fixed offset) in addition to passing the position or offset, respectively. So I'd like to hear other opinions on this matter.

adonath:
I'm fine to add more options to the `SafeMaskMaker` as needed. I think we can just add an additional `fixed_offset` attribute, which is exclusive with the existing `position` attribute. The new position can be computed from the pointing position using `SkyCoord.directional_offset_by()`. I think it doesn't  add to much complexity...PRs are welcome any time!

lmohrmann:
OK, I will assign this one to me and see if I can find the time to propose a PR.

adonath:
I think this is rather simple to implement (~2 hrs including tests). @lmohrmann Are you still interested to give this a try?

lmohrmann:
Interested: yes
Time to do it: no, unfortunately

So please go ahead implement this if you think it's useful (I think it would be). But it's not something that's absolutely necessary I guess.

adonath:
Addressed in #3338.

PR #2747, opened 2020-01-29
Modify Config and Analysis to support SafeMaskMaker

registerrier:
**Description**

This pull request is a proposal to try to solve issue #2688. It adds a `SafeMaskMethodEnum` to list possible methods and a `SafeMaskConfig` containing methods and the parameters required. This means that the parameters are always passed even if the actual method they refer to is not used. 

**Dear reviewer**

this is preliminary and needs test. It the right approach @Bultako @adonath @JouvinLea ?


gammapy/analysis/config.py
```diff
@@ -130,16 +143,19 @@ class BackgroundConfig(GammapyBaseConfig):
...
 
 class EnergyAxesConfig(GammapyBaseConfig):
     energy: EnergyAxisConfig = EnergyAxisConfig()
     energy_true: EnergyAxisConfig = EnergyAxisConfig()
 
-
-class SelectionConfig(GammapyBaseConfig):
-    offset_max: AngleType = "2.5 deg"
-
-
```
Bultako: The `offset_max` is also used elsewhere in `core.py` to produce cutouts of datasets.  Its value is fetched from `datasets.geom.selection` settings. 

Does it make sense to have two different values for the offset? One used to produce cutouts of datasets in the analysis process and another one related with the safe mask?  If we keep two different params for the offset, they could have the same names since they would at different levels in the settings.

For example:
-  `datasets.geom.selection.offse_max`
-  `datasets.safe_mask.settings.offset_max`

gammapy/analysis/config.py
```diff
@@ -154,7 +170,6 @@ class WcsConfig(GammapyBaseConfig):
 
 class GeomConfig(GammapyBaseConfig):
     wcs: WcsConfig = WcsConfig()
-    selection: SelectionConfig = SelectionConfig()
```
Bultako: See comment above on keeping this parameter in the settings for cutouts and not use the one for the safe mask.

gammapy/analysis/config.py
```diff
@@ -130,16 +143,19 @@ class BackgroundConfig(GammapyBaseConfig):
     method: BackgroundMethodEnum = BackgroundMethodEnum.reflected
     exclusion: FilePath = None
 
+class SafeMaskSettingsConfig(GammapyBaseConfig):
+    offset_max: AngleType = "2.5 deg"
+    aeff_percent: float = 10
+    bias_percent: float = 10
```
Bultako: I guess these are not the whole collection of parameters needed for all the possible safe mask methods. Also note that you could have `None` values by default for all these other params missing.

gammapy/analysis/core.py
```diff
@@ -246,11 +246,14 @@ def _map_making(self):
...
         geom_irf["binsz_irf"] = geom_settings.wcs.binsize_irf.to("deg").value
-        offset_max = geom_settings.selection.offset_max
         log.info("Creating datasets.")
 
         maker = MapDatasetMaker(selection=self.config.datasets.map_selection)
-        maker_safe_mask = SafeMaskMaker(methods=["offset-max"], offset_max=offset_max)
+
+        safe_mask_selection = self.config.datasets.safe_mask.methods
+        safe_mask_settings = self.config.datasets.safe_mask.settings.dict()
+        offset_max = safe_mask_settings['offset_max']
```
Bultako: See comment above on keeping this parameter in the settings for cutouts and not use the one for the safe mask.

gammapy/analysis/core.py
```diff
@@ -296,7 +299,11 @@ def _spectrum_extraction(self):
             exclusion_region = Map.read(datasets_settings.background.exclusion)
             bkg_maker_config["exclusion_mask"] = exclusion_region
         bkg_maker = ReflectedRegionsBackgroundMaker(**bkg_maker_config)
-        safe_mask_maker = SafeMaskMaker(methods=["aeff-default", "aeff-max"])
+
+
+        safe_mask_selection = self.config.datasets.safe_mask.methods
+        safe_mask_settings = self.config.datasets.safe_mask.settings.dict()
+        safe_mask_maker = SafeMaskMaker(methods=safe_mask_selection, **safe_mask_settings)
```
Bultako: Maybe we could have the `safe_mask_maker` as a property of `analysis` and avoid  code duplication or at least pass it as an argument to `_spectrum_extraction` and `_map_making` internal in the `get_datasets` method.

gammapy/analysis/core.py
```diff
@@ -246,11 +246,14 @@ def _map_making(self):
...
-        offset_max = geom_settings.selection.offset_max
         log.info("Creating datasets.")
 
         maker = MapDatasetMaker(selection=self.config.datasets.map_selection)
-        maker_safe_mask = SafeMaskMaker(methods=["offset-max"], offset_max=offset_max)
+
+        safe_mask_selection = self.config.datasets.safe_mask.methods
+        safe_mask_settings = self.config.datasets.safe_mask.settings.dict()
+        offset_max = safe_mask_settings['offset_max']
+        maker_safe_mask = SafeMaskMaker(methods=safe_mask_selection, **safe_mask_settings)
```
Bultako: Does `maker_safe_mask` fails when is run without a method (i.e. an empty list)? 
In that case we should prevent it from running it.

gammapy/analysis/config.py
registerrier: Normally I removed the `datasets.geom.selection.offset_max`to keep only one. And now the `MapDataset.cutout` use the value in `datasets.safe_mask.setting.offset_max` in `Analysis`.
The code should be changed in such a way that if `SafeMaskMaker` does not use the `offset-max` method, it shouldn't do a cutout.

 I understand the logic of the `cutout` to be directly connected to the `offset_max` mask. 
If we have no `offset-max`, we probably shouldn't be doing any cutout no?

gammapy/analysis/config.py
registerrier: Normally the `offset_max` used by `cutout` here should be directly connected to the `SafeMaskMaker`. The parameter config is probably more logically found on the `SafeMaskConfig`.

gammapy/analysis/config.py
registerrier: Yes I think we miss a `position` parameter that would take a `SkyCoord` in the parameter list (I am not sure if we would really use that option, but it exists at the moment). And this could also be adapted, as requested in #2704.

gammapy/analysis/core.py
registerrier: Normally it does not. I assume that the empty list would mean no safe mask applied.

gammapy/analysis/config.py
adonath: In general I think the  `offset_max` and cutout width should be two different parameters. The first argument is handling of PSF leakage, so the cutout should be larger then the `offset_max` to account for PSF leakage in joint analyses. Secondly the offset_max can be changed after data reduction (before stacking) when the cutout is larger.

However the cutout-width doesn't have to be configurable by the user from the HLI. So I'm fine to have only one `offset-max` parameter in the HLI and derive internally the cutout-width in a reasonable way...

Bultako:
As discussed in last developers call, there's a general agreement in solving very specific needs like this one adding a generic `params` dict in a configuration section. So to modify the default values of optional parameters of a specific method of the Gammapy API (in this case [SafeMaskMaker](https://docs.gammapy.org/0.15/api/gammapy.cube.SafeMaskMaker.html)), in the form as it has been done in #2756. It will be the responsibility of the user to set the values of this generic `params` dict, since the HLI does not provide any validation on the values neither the names of the params, and also does not expose these internal optional parameters with dot completion.

If everybody agrees, I guess we could close this PR and start a new one?

gammapy/analysis/config.py
registerrier: The `SelectionConfig` is back in.

tutorials/analysis_3d.ipynb
```diff
@@ -592,7 +592,699 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.7.3"
+   "version": "3.7.0"
```
adonath: Is it possible to avoid this extra output somehow? Maybe just edit the notebook json directly for now, without re-executing it? And then we ask @Bultako for help next week?

Issue #2754, opened 2020-01-30
MapDataset.cutout fails with mode=partial option


registerrier:
`MapDataset.cutout` is used in the observation loop to produce a `MapDataset`. By default it is set with option `mode='trim'`. This option simply cut a fraction of the target dataset irrespective of the requested map size. 

This option is an issue if one is to use `FoVBavkground` or `RingBackground`. Especially in the latter case, one needs to have the full field of view to perform proper ring background estimation.
It is therefore necessary to use option `mode='partial'` for the observation loop.

With the current dev version (and former ones as well), the `Map.cutout` does not support this option, because the data of the new `Map` must be created first and only the parent slice should be set to the cutout values. 
Also `WcsNDMap.stack` fails for such maps because the `weight` argument does not use slices (and its geometry is not compared to that of the input `Map`).

adonath:
Solved in #2755.

PR #2755, opened 2020-01-30
Fix WcsNDMap and MapDataset cutout to support mode='partial'

registerrier:
**Description**

This pull request solves issue #2754. First, an array of the correct size filled with 0 is created with the correct `dtype`. Then we add the content of the `WcsNDMap` using the the parenting cutout slices.
The `cutout-slices` are also used for the `weight` map that is used in `WcsNDMap`.

Note that I had put a test to ensure that `other` and `weights` have the same `WcsGeom`. This breaks many tests as it seems not to be the case in many situation. I found that puzzling. Maybe this should be clarified.

**Dear reviewer**

This is ready for review. We need to see whether `mode='partial'` should be the default for most observation loops. I tend to believe it should be since at least `RingBackgroundMaker` and `FoVBrackgroundMaker` need for field of view datasets to properly work.
Any opinion @adonath @AtreyeeS ?

I put v0.16 as milestone, and we'll see if we can really merge this in on time.

### DIFF
--- a/gammapy/cube/fit.py
+++ b/gammapy/cube/fit.py
@@ -1255,11 +1255,11 @@ class MapDatasetOnOff(MapDataset):
         # Factor containing: self.alpha * self.counts_off + other.alpha * other.counts_off
         tmp_factor = (self.alpha * self.counts_off).copy()
         tmp_factor.data[~self.mask_safe.data] = 0
-        tmp_factor.stack(other.alpha * other.counts_off, weights=other.mask_safe.data)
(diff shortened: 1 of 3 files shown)


adonath:
The "issue" is that the weights argument is sometimes provided as spatial map only (without energy axis). Internally we then rely on broadcasting to apply the weights. I think this behaviour is fine. In https://github.com/gammapy/gammapy/commit/81cdd1b42da5a7f1b27b784c4131b814f9db8c3f I added at least a test for the spatial geometries...

PR #2760, opened 2020-02-03
Li&Ma statistics

bkhelifi:
**Description**
This pull request aims to add functions based on Li&Ma methods to compute excess errors. Following recommendations of @registerrier, a new class has been created that contains all functions to compute the Li&Ma errors. For logical reasons, the Li&Ma significance computation that exists has been moved from poisson.py to this new class.

This PR contains also the test functions, and also the improvement of the documentation in order to describe this new class.

**Possible improvements**
In this current implementation, few points might be improved/cleaned:
- As by default the 'significance' functions can take as arguments either an int or an array, the current codes support this feature. The 'tricks' to make it could be improved however: see lines 100-113 and 185-198 of li_ma.py
- The reference of the Li&Ma paper in the docstrings and the rst files is not coded identically. What is the good way?
- The content of the Li&Ma general documentation could be developed as a statistical lecture. But I think that the reading of the paper is the best. Feel free to adapt/change the documentation.

### DIFF
--- a/gammapy/stats/li_ma.py
+++ b/gammapy/stats/li_ma.py
@@ -1,0 +1,347 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+"""Li and Ma algorithm to compute parameter confidence limits and significance."""
+import logging
+import numpy as np
+
+__all__ = [
+    "lm_loglikelihood",
+    "lm_dexcess_down",
+    "lm_dexcess_up",
+    "lm_dexcess",
+    "lm_significance_on_off",
+    "lm_significance"
+]
+
+log = logging.getLogger(__name__)
+
+
+def lm_loglikelihood(n_on, n_off, alpha, signal):
+    r"""Compute Li&Ma Log-likelihood.
+
+    Note: This is a little trick to avoid problems for negative signal
+    (note that this method is often called with signal=excess).
+    In that case the Poisson signal + b0 can become negative
+    and log(signal + b0) isn't defined, i.e. the likelihood cannot be computed.
+    To avoid this problem, the on and off regions are switched
+
+    For more information see :ref:`documentation <li_ma>`.
+
+    Parameters
+    ----------
+    n_on : float
+        Observed number of counts in the on region
+    n_off : float
+        Observed number of counts in the off region
+    alpha : float
+        On / off region exposure ratio for background events
+    signal : float
+        Number of excess in the on region
+
+    Returns
+    -------
+    log_likelihood : `~numpy.ndarray`
+
+    """
+    if signal > 0:
+        n1 = n_on
+        n2 = n_off
+        aa = alpha
+    else:
+        n2 = n_on
+        n1 = n_off
+        aa = 1.0 / alpha
+        signal = -aa * signal
+
+    # Maximum likelihood background estimate b0
+    tt = (1+aa) * signal - aa * (n1+n2)
+    delta = tt*tt + 4 * aa * (1+aa) * signal * n2
+    b0 = (np.sqrt(delta) - tt) / (2 + 2*aa)
+    # b0 should always be >= 0.
+    # Just to avoid possible problems with rounding errors:
+    if b0 < 0:
+        b0 = 0.
+
+    # Compute Poisson log likelihood r for two measurements: on / off
+    ll = n1 * np.log(signal + b0) + n2 * np.log(b0 / aa) - (signal + b0) - b0 / aa
+    return ll
+
+
+def lm_dexcess_down(n_on, n_off, alpha, n_sig=1.):
+    r"""Compute downward excess uncertainties of an on-off observation based on Li&Ma [1]. [LiMa1983]_
+
+    Searches the average value of the signal for which the log-likelihood ratio is n_sig away
+
+    Note: The required step in log likelihood for "nsig sigma errors" is offset = 0.5*n_sig^2
+            http://wwwasdoc.web.cern.ch/wwwasdoc/minuit/node30.html
+
+    For more information see :ref:`documentation <li_ma>`.
+
+    Parameters
+    ----------
+    n_on : array_like
+        Observed number of counts in the on region
+    n_off : array_like
+        Observed number of counts in the off region
+    alpha : array_like
+        On / off region exposure ratio for background events
+    n_sig : float
+        Number of sigma
+
+    Returns
+    -------
+    significance : `~numpy.ndarray`
+        Error estimate
+
+     References
+    ----------
+    * Li and Ma, "Analysis methods for results in gamma-ray astronomy"
+      <https://ui.adsabs.harvard.edu/abs/1983ApJ...272..317L>`_
+    """
+    scalar = False
+    if np.isscalar(n_on):
+        n_on = np.array([n_on])
+        scalar = True
+    if np.isscalar(n_off):
+        n_off = np.array([n_off])
+        scalar = True
+    if np.isscalar(alpha):
+        alpha = np.array([alpha])
+        scalar = True
+    n_on = np.asanyarray(n_on, dtype=np.float64)
+    n_off = np.asanyarray(n_off, dtype=np.float64)
+    alpha = np.asanyarray(alpha, dtype=np.float64)
+    n_sig = np.full(n_on.shape, n_sig)
+
+    # Optimal value of signal
+    excess0 = n_on - n_off * alpha
+    l0 = np.zeros(n_on.shape)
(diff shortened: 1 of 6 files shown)


gammapy/stats/li_ma.py
```diff
@@ -0,0 +1,347 @@
...
+        On / off region exposure ratio for background events
+    signal : float
+        Number of excess in the on region
+
+    Returns
+    -------
+    log_likelihood : `~numpy.ndarray`
+
+    """
+    if signal > 0:
```
adonath: Please rewrite this expression to be compatible with `~numpy.ndarray`.

gammapy/stats/li_ma.py
```diff
@@ -0,0 +1,347 @@
...
+    -------
+    significance : `~numpy.ndarray`
+        Error estimate
+
+     References
+    ----------
+    * Li and Ma, "Analysis methods for results in gamma-ray astronomy"
+      <https://ui.adsabs.harvard.edu/abs/1983ApJ...272..317L>`_
+    """
+    scalar = False
```
adonath: In general there is not need for this type of check. Once the expression are compatible with `~numpy.ndarray` the scalar handling come for free. See e.g.
```python3
>>> import numpy as np
>>> a = np.log(5)
>>> print(a)
1.6094379124341003
```

vs.
```python3
>>> import numpy as np
>>> a = np.log([5])
>>> print(a)
[1.60943791]
```
So just write numpy expressions...

gammapy/stats/li_ma.py
```diff
@@ -0,0 +1,347 @@
...
+    n_on = np.asanyarray(n_on, dtype=np.float64)
+    n_off = np.asanyarray(n_off, dtype=np.float64)
+    alpha = np.asanyarray(alpha, dtype=np.float64)
+    n_sig = np.full(n_on.shape, n_sig)
+
+    # Optimal value of signal
+    excess0 = n_on - n_off * alpha
+    l0 = np.zeros(n_on.shape)
+    for ii in range(len(n_on)):
+        l0[ii] = lm_loglikelihood(n_on[ii], n_off[ii], alpha[ii], excess0[ii])
```
adonath: Once you allow for arrays in `lm_loglikelihood` this loop can go...

gammapy/stats/li_ma.py
```diff
@@ -0,0 +1,347 @@
...
+
+    # Search for signal such as -2 * log H > 1 * sig
+    # i.e. find a signal > dExcess_Up so that we can
+    # use an interval bisection algorithm to find dExcess_Up
+    offset = np.full(n_on.shape, n_sig*n_sig*0.5)
+    signal = excess0.copy() - 1
+    ll = np.zeros(n_on.shape)
+    for ii in range(len(n_on)):
```
adonath: This double Python loop is very inefficient and should be replaced by numpy expressions. I see two possibilities:
- Either use Scipy root finding e.g. https://docs.scipy.org/doc/scipy/reference/optimize.html (Brent is fast and stable...)
- Avoid the python loop by extending the numpy arrays to another dimension and reducing along this dimension again

gammapy/stats/li_ma.py
```diff
@@ -0,0 +1,347 @@
...
+            count += 1
+
+    # Dichotomic search (i.e. interval bisection algorithm)
+    up = excess0.copy()
+    low = signal.copy()
+    signal = (up+low)/2.
+    for ii in range(len(n_on)):
+        ll[ii] = lm_loglikelihood(n_on[ii], n_off[ii], alpha[ii], signal[ii])
+        count = 0.
+        while np.abs(l0[ii]-ll[ii]-offset[ii]) > 1e-4 and np.isfinite(ll[ii]) and count < 100:
```
adonath: See my comments above and maybe try to avoid hard-coded numbers such as `1e-4` and `100`. If they are of relevance for the algorithm just expose them as function arguments.

gammapy/stats/li_ma.py
```diff
@@ -0,0 +1,347 @@
...
+    -------
+    significance : `~numpy.ndarray`
+        Error estimate
+
+    References
+    ----------
+    .. [1] Li and Ma, "Analysis methods for results in gamma-ray astronomy",
+       `Link <https://ui.adsabs.harvard.edu/abs/1983ApJ...272..317L>`_
+    """
+    scalar = False
```
adonath: See comment above...

gammapy/stats/li_ma.py
```diff
@@ -0,0 +1,347 @@
...
+        scalar = True
+    n_on = np.asarray(n_on, dtype=np.float64)
+    n_off = np.asanyarray(n_off, dtype=np.float64)
+    alpha = np.asanyarray(alpha, dtype=np.float64)
+    n_sig = np.full(n_on.shape, n_sig)
+
+    # Optimal value of signal
+    excess0 = n_on - n_off * alpha
+    l0 = np.zeros(n_on.shape)
+    for ii in range(len(n_on)):
```
adonath: See comment above...

gammapy/stats/li_ma.py
```diff
@@ -0,0 +1,347 @@
...
+        l0[ii] = lm_loglikelihood(n_on[ii], n_off[ii], alpha[ii], excess0[ii])
+
+    # Search for signal such as -2 * log H > 1 * sig
+    # i.e. find a signal > dExcess_Up so that we can
+    # use an interval bisection algorithm to find dExcess_Up
+    offset = np.full(n_on.shape, n_sig*n_sig*0.5)
+    signal = excess0.copy() + 1
+    ll = np.zeros(n_on.shape)
+    for ii in range(len(ll)):
```
adonath: Try to avoid the python loop e.g. using scipy root finding.

gammapy/stats/li_ma.py
```diff
@@ -0,0 +1,347 @@
...
+            signal[ii] = (up[ii]+low[ii])/2.
+            ll[ii] = lm_loglikelihood(n_on[ii], n_off[ii], alpha[ii], signal[ii])
+            count += 1
+
+    if scalar:
+        return np.abs(signal[0] - excess0[0])
+    return np.abs(signal - excess0)
+
+
+def lm_dexcess(n_on, n_off, alpha, n_sig=1.):
```
adonath: My recommendation would be to just remove this function. I think it's totally fine if users write the 3 lines to compute the mean themselves. I think there is not need to introduce an extra function with a lot of documentation. for such a simple operation...

gammapy/stats/li_ma.py
```diff
@@ -0,0 +1,347 @@
...
+    -------
+    significance : `~numpy.ndarray`
+        Significance estimate
+
+    References
+    ----------
+    .. [1] Li and Ma, "Analysis methods for results in gamma-ray astronomy",
+       `Link <https://ui.adsabs.harvard.edu/abs/1983ApJ...272..317L>`_
+    """
+    n_on = np.asanyarray(n_on, dtype=np.float64)
```
adonath: I think in general there is not need to parse the values as arrays. Numpy takes care of this...

gammapy/stats/li_ma.py
```diff
@@ -0,0 +1,347 @@
...
+    r"""Compute upward excess uncertainties of an on-off observation based on Li&Ma [1].
+
+    Searches the average value of the signal for which the log-likelihood ratio is n_sig away
+
+    Note: The required step in log likelihood for "nsig sigma errors" is offset = 0.5*n_sig^2
+            http://wwwasdoc.web.cern.ch/wwwasdoc/minuit/node30.html
+
+    Parameters
+    ----------
+    n_on : array_like
```
adonath: Fo the doctoring jus link to `~numpy.ndarray`, instead of `array_like`.

bkhelifi: Ok. For information, in poisson.py, everything is in array_like. Should we clean also poisson.py?

gammapy/stats/li_ma.py
bkhelifi: Agree. For the functions in poisson.py, this code supports this 2 ways of doing, using float and numpy expressions.
Should we change also poisson.py (and all the documentation) such that the functions take only numpy expression?

gammapy/stats/li_ma.py
adonath: Yes, please. I guess at some point we will do this systematically, but any place where it's fixed   already is one place less...:-)

adonath:
@bkhelifi Can you confirm this PR is superseded by #2775?

adonath:
Superseded by #2775.

PR #2762, opened 2020-02-03
Li Ma based significance map estimator

registerrier:
**Description**

This pull request creates a `SignificanceMapEstimator` that works with datasets. It simply wraps the `compute_lima_significance_image` functions.
Note that the on-off version was modified to convolve the `n_off`. If we don't do so,  no convolution is applied for off counts which can be absurd for real OFF measurements (i.e. not ring based). The change for the hgps test are less than 0.15 in absolute value. 

**Dear reviewer**


gammapy/detect/lima.py
```diff
@@ -2,13 +2,68 @@
...
+__all__ = [
+    "SignificanceMapEstimator",
+    "compute_lima_image",
+    "compute_lima_on_off_image",
+]
 
 log = logging.getLogger(__name__)
 
 
+class SignificanceMapEstimator:
```
adonath: I have a slight preference for naming this `LiMaMapEstimator` or similar, because it refers to a very specific (even if it's standard...) method of estimating significance. In general there are other methods. In many ways I see this `MapEstimator` on the same level as a `TSMapEstimator` (in fact I think one can show that the Li&Ma solution corresponds to the TS map case of constant exposure, background and flat source kernel...).

In this regard I think it's also an option to extend the number of  maps computed by the `LiMaMapEstimator` to more quantities, such as `flux`, `flux_err` and `flux_ul` later.

gammapy/detect/lima.py
```diff
@@ -2,13 +2,68 @@
...
+
+    Parameters
+    ----------
+    correlation_radius : ~astropy.coordinate.Angle
+        correlation radius to use
+    """
+
+    def __init__(self, correlation_radius):
+        self.radius = Angle(correlation_radius)
+
```
adonath: Maybe attach the existing `compute_lima_image` methods as `staticmethod` on the `LiMaMapEstimator`? Then all the functionality is bundled in a single name-space? Like `LiMaMapEstimator.estimate_maps(counts, background, kernel)` and `LiMaMapEstimator.estimate_maps_on_off(counts, counts_off, acceptance, acceptance_off, kernel)`

gammapy/detect/lima.py
```diff
@@ -2,13 +2,68 @@
...
+        This requires datasets with only 1 energy bins (image-like).
+        Usually you can obtain one with `dataset.to_image()`
+
+
+        Parameters
+        ----------
+        dataset : `~gammapy.cube.MapDataset` or `~gammapy.cube.MapDataset`
+            input dataset
+        """
```
adonath: Add a `Returns` section here?

gammapy/detect/lima.py
```diff
@@ -2,13 +2,68 @@
...
+        if not isinstance(dataset, MapDataset) and not isinstance(
```
adonath: `isinstance()` can be used with a tuple like so: `isinstance(dataset, (MapDataset, MapDatasetOnOff))`. It also takes inheritance into account, this means `isinstance(dataset_on_off, MapDataset)` will also return `True`. I would propose to simplify here a bit.

gammapy/detect/lima.py
registerrier: Done

gammapy/detect/lima.py
registerrier: Done. 
But I guess in the long run this might not be the most relevant name for this type of `Estimator`.

gammapy/detect/lima.py
registerrier: Right. Done.

gammapy/detect/lima.py
registerrier: Oops, done.

Issue #2682, opened 2019-12-11
Add a tutorial demonstrating confidence contours creation and plotting


registerrier:
We currently miss a tutorial notebook showing how to perform the computation of confidence contours for parameters after a fit. We should add one.

It should focus only on fit and contours, I suggest to re-use reduced spectral data from the joint-crab paper in $GAMMAPY_DATA/joint-crab/spectra. 
It should show the loading of datasets, the model creation (here e.g. LogParabola), the Fit itself the call to `minos_contours` and then plot the result.

registerrier:
This is done in PR #2764. Closing.

PR #2764, opened 2020-02-05
Add modeling and fitting  tutorial notebook

QRemy:
- add  a tutorial notebook showing how to perform the computation of confidence contours for parameters after a fit as suggested in #2682
 
Show the loading of datasets, the model creation (LogParabola), the Fit itself, the call to minos_contours, and then plot the result.

The implementation largely re-use the code from the joint-crab validation:
 https://github.com/gammapy/gammapy-benchmarks/blob/master/validation/joint-crab/make.py

- add extra functionality to plot multiple contours levels
- add outliers filter for aberrant contour points given by Minuit
- change the interpolation scheme


QRemy:
> Thanks @QRemy . This looks very nice.
> As we discussed, at some point some of the contour plotting functions should go in the API itself.
> There is one thing that puzzles me. The best beta value is 0.26 but the 1 sigma contours does not seem to include it. Is there a problem, or did I miss something?
> 
> Also you might use alpha instead of gamma for the index to be consistent with the parameter naming scheme.


The difference in beta seems to comes from  the np.log(10) factor here:
`contours["contour_alpha_beta"] = {
        "alpha": contour["x"].tolist(),
        "beta": (contour["y"]*np.log(10)).tolist(),
    } `

This is copied from the validation code, was it something introduce to match the parametrization of the joint Crab paper ?

registerrier:
Ah yes indeed. The joint crab paper used a log10 expression to match the sherpa model which is expressed in log10 hence the factor. This should be removed here.

PR #2768, opened 2020-02-07
Refactor map dataset background model

adonath:
**Description**

This pull request refactors the background model handling in `MapDataset`. Instead of a separate `background_model` attribute the `BackgroundModel` is now part of the general `Models` object. This unifies the treatment of model components. To simplify the transition I have kept the `MapDataset.background_model` for now, which is defined from the `BackgroundModel` contained in `Models`.


adonath:
@registerrier and @QRemy to avoid future merge conflicts I'll merge this PR now, but we will re-address the question of background model handling in follow-up PRs.

PR #2772, opened 2020-02-11
Add region serialization on CountsSpectrum and SpectrumDataset

registerrier:
**Description**

This pull request introduces a region serialization mechanism on the `CountsSpectrum` and the `SpectrumDataset`.
It follows the scheme defined here https://fits.gsfc.nasa.gov/registry/region/region.pdf, which is implemented in `regions.fits_region_objects_to_table` and  `regions.FITSRegionParser`. A `Table` storing the `PixelRegion` is build and a `WCS` object is added to be able to convert to and from `SkyRegion`.

Tests have been added to check that the regions after serialization match the input ones.

**Dear reviewer**

This should be ready for review. The stacking process still remains to be implemented for `CountsSpectrum.region`. It is not totally clear how to include this in the arithmetic implemented there.

### DIFF
--- a/gammapy/spectrum/core.py
+++ b/gammapy/spectrum/core.py
@@ -5,7 +5,8 @@ import numpy as np
 import astropy.units as u
 from astropy.io import fits
 from astropy.table import Table
-from regions import fits_region_objects_to_table
+from astropy.wcs import WCS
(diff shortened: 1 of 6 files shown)


adonath:
@registerrier Sorry I just realised this is already the case. So no further comments from my side...

PR #2775, opened 2020-02-12
Introduce fit statistics classes to estimate signal counts errors and upper limits

registerrier:
**Description**

This pull request introduces new classes to perform excess counts estimates, namely significance, error, asymmetric errors, upper limits. They both rely of the fit statistics functions `cash` and `wstat`.
 They cover the cases of Poisson signal with known background (`cash`) and unknown background  (`wstat`). They can replace the Li & Ma significance methods, because they compute the same thing without using the explicit expression from the Li&Ma paper.

This PR provides a more general approach to error estimation that PR #2760 .

**Dear reviewer**

The choice has been made to work only with float entries and to leave the iteration over arrays to the user. This is because the asymmetric errors and the UL requires root finding per entry. 
This might be changed by introducing an iterator over the arrays to compute the errors.

Another thing that could be added but I am not sure it is useful is a scan of the TS value.


registerrier:
I have added and then removed direct significance calculations. 
Indeed there is a inconsistency between the implicit definition of significance in delta TS tests e.g. a la Li & Ma where we call significance the square root of delta TS which is not a real 1 sided p-value.
Globally we should clarify this a bit and probably stay with delta TS as much as possible.

adonath:
@registerrier I agree we should use `delta_ts` consistently and show in the "How To" how to translate delta_ts into (1 sided / 2 sided) sigma and the other way around.

gammapy/stats/fit_stat_estimator.py
```diff
@@ -0,0 +1,178 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+import abc
+import numpy as np
+from scipy.optimize import brentq
+from gammapy.stats import wstat, cash
+
+__all__ = ["WStatEvaluator", "CashEvaluator"]
+
+
+class FitStatisticEvaluator(abc.ABC):
```
adonath: I have thought again once about the name and noticed that we should maybe put the word "counts" in there, because that's how you basically explained the functionality to us in the past. We could maybe distinguish between a `FitStatistic`, which is basically just the statistic function used for (binned) likelihood fits and a `CountsStatistic`, which is what you introduce here. So what about renaming `CashEvaluator` to `CashCountsStatistic` and `WStatEvaluator` to `WStatCountsStatistic`? I think the `Evaluator` suffix might be a bit mis-leading as we already have a `MapEvaluator`, which has nothing to do with these classes here...

registerrier: Done

gammapy/stats/fit_stat_estimator.py
```diff
@@ -0,0 +1,156 @@
...
+
+        Searches the signal value for which the test statistics is n_sigma away from the maximum.
+
+        Parameters
+        ----------
+        n_sigma : float
+            Confidence level of the uncertainty expressed in number of sigma. Default is 1.
+        """
+        min_range = self.excess - 2 * n_sigma * self.std
+
```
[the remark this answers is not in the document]
registerrier: Done

PR #2776, opened 2020-02-13
Add support for FoVBackground on the HLI

registerrier:
**Description**

This pull request adds the `FoVBackgroundMaker` to the `Analysis` class.

A `fov_background` method is added to the `BackgroundMethodEnum` as well as a `none`  method. (This is meant to avoid asking users to select `reflected` if they want no background estimate on a 3d method.)

A `parameters` dict is added to the `BackgroundConfig` and is empty by default.

In the `Analysis` class itself, a test is added for the background method to create the correct `BackgroundMaker`. If the `maker` exists it is called in the dataset production loop.
The analysis_1 notebook has been adapted to include the FoVBackground.

**Dear reviewer**

is this is the right approach?

### DIFF
--- a/gammapy/analysis/config.py
+++ b/gammapy/analysis/config.py
@@ -67,8 +67,8 @@ class FrameEnum(str, Enum):
 
 class BackgroundMethodEnum(str, Enum):
     reflected = "reflected"
-    fov_background = "fov_background"
-
+    fov = "fov_background"
+    no_background = "None"
 
 class SafeMaskMethodsEnum(str, Enum):
     aeff_default = "aeff-default"
--- a/gammapy/analysis/config.py
+++ b/gammapy/analysis/config.py
@@ -68,7 +68,6 @@ class FrameEnum(str, Enum):
 class BackgroundMethodEnum(str, Enum):
     reflected = "reflected"
     fov = "fov_background"
-    no_background = ""
 
 
 class SafeMaskMethodsEnum(str, Enum):
@@ -139,7 +138,7 @@ class FitConfig(GammapyBaseConfig):
 
 
 class BackgroundConfig(GammapyBaseConfig):
-    method: BackgroundMethodEnum = BackgroundMethodEnum.reflected
+    method: BackgroundMethodEnum = None
     exclusion: FilePath = None
     parameters: dict = {}
 
--- a/gammapy/analysis/config.py
+++ b/gammapy/analysis/config.py
@@ -130,7 +130,7 @@ class TimeRangeConfig(GammapyBaseConfig):
 class FluxPointsConfig(GammapyBaseConfig):
     energy: EnergyAxisConfig = EnergyAxisConfig()
     source: str = "source"
-    params: dict = {}
+    parameters: dict = {}
 
 
 class FitConfig(GammapyBaseConfig):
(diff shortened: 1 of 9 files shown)


registerrier:
Is the option "None" in `BackgroundMethodEnum` good to have no background maker applied @Bultako @adonath ?

gammapy/analysis/core.py
```diff
@@ -252,6 +252,20 @@ def _map_making(self):
...
+        bkg_maker_config.update(self.config.datasets.background.parameters)
+
+        bkg_method = self.config.datasets.background.method
+        if bkg_method == "fov_background":
+            log.debug(f"Creating FoVBackgroundMaker with arguments {bkg_maker_config}")
+            bkg_maker = FoVBackgroundMaker(**bkg_maker_config)
+        else:
+            bkg_maker = None
+
         stacked = MapDataset.create(geom=geom, name="stacked", **geom_irf)
```
Bultako: Does it make sense to fetch the exclusion_region when no background method is provided?
I would suggest to re-use the same coding pattern found in `_spectrum_extraction` function below.

```
        bkg_method = self.config.datasets.background.method
        if bkg_method == "fov_background":
            bkg_maker_config = {}
            if datasets_settings.background.exclusion:
                exclusion_region = Map.read(datasets_settings.background.exclusion)
                bkg_maker_config["exclusion_mask"] = exclusion_region
            
            bkg_maker_config.update(self.config.datasets.background.parameters)
            bkg_maker = FoVBackgroundMaker(**bkg_maker_config)
            log.debug(
                f"Creating FoVBackgroundMaker with arguments {bkg_maker_config}"
            )
        else:
            bkg_maker = None
            log.warning(
                f"No background maker set for 3d analysis. Check configuration."
            )
```

gammapy/analysis/core.py
```diff
@@ -292,11 +310,22 @@ def _spectrum_extraction(self):
...
-            exclusion_region = Map.read(datasets_settings.background.exclusion)
-            bkg_maker_config["exclusion_mask"] = exclusion_region
-        bkg_maker = ReflectedRegionsBackgroundMaker(**bkg_maker_config)
+
+        bkg_method = self.config.datasets.background.method
+        if bkg_method == "reflected":
+            bkg_maker_config = {}
+            if datasets_settings.background.exclusion:
+                exclusion_region = Map.read(datasets_settings.background.exclusion)
+                bkg_maker_config["exclusion_mask"] = exclusion_region
```
Bultako: Could we add the following line here, as it is done in the `map_making()' function ?
```
bkg_maker_config.update(self.config.datasets.background.parameters)
```

gammapy/analysis/config.py
```diff
@@ -67,6 +67,8 @@ class FrameEnum(str, Enum):
 
 class BackgroundMethodEnum(str, Enum):
     reflected = "reflected"
+    fov = "fov_background"
+    no_background = "None"
```
Bultako: Since all fields should be `str` types, I'd suggest to use `no_background = ""` instead

Bultako:
> Is the option "None" in `BackgroundMethodEnum` good to have no background maker applied @Bultako @adonath ?

@registerrier 
please, see https://github.com/gammapy/gammapy/pull/2776#discussion_r380553464 above

Bultako:
Thanks @registerrier for this PR.
I've addressed some review comments and improved test coverage.
All tests passed, merging now.

PR #2777, opened 2020-02-13
Add processing options for SkyModel and SkyDiffuseCube 

QRemy:
Add options to apply or not psf and edisp to SkyModel and SkyDiffuseCube, so ones can use pre-processed templates


adonath:
Thanks @QRemy, as discussed in the dev call today please also add the `exposure=True` option and maybe modify the Fermi-LAT tutorial (or is there a better one?) to make use of this.

PR #2779, opened 2020-02-13
Exclusion mask notebook

registerrier:
**Description**

This pull request adds a notebook to the tutorials to demonstrate how to create an exclusion mask.

**Dear reviewer**

This should be ready for review. Please comment on missing elements. etc.

### DIFF
--- a/tutorials/exclusion_mask.ipynb
+++ b/tutorials/exclusion_mask.ipynb
@@ -11,7 +11,7 @@
     "### Prerequisites\n",
     "\n",
     "- Understanding of basic analyses in 1D or 3D.\n",
-    "- Usage of `regions` and catalogs. \n",
+    "- Usage of `~regions` and catalogs, see the [catalog notebook](catalog.ipynb). \n",
     "\n",
     "### Context\n",
     "\n",
@@ -28,8 +28,7 @@
     "We can rely on known sources positions and properties to build a list of regions (here `~regions.SkyRegions`) enclosing most of the signal that our detector would see from these objects. We show below how to build this list manually or from an existing catalog. \n",
     "\n",
     "Finally, we show how to build the mask from a `MapDataset`, finding pixels which contain statistically significant signal. To do so, we use the `LiMaMapEstimator`\n",
-    "\n",
-    "Building a reliable exclusion mask is often an iterative process. "
+    " "
    ]
   },
   {
@@ -228,7 +227,7 @@
    "outputs": [],
    "source": [
     "mask_data = geom.region_mask(regions, inside=False)\n",
-    "mask_map = Map.from_geom(geom, data=mask_data)"
+    "mask_map_catalog = Map.from_geom(geom, data=mask_data)"
    ]
   },
   {
@@ -237,6 +236,25 @@
    "metadata": {},
    "outputs": [],
    "source": [
+    "mask_map_catalog.plot()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Combining masks\n",
+    "\n",
+    "If two masks share the same geometry it is easy to combine them with `Map` arithmetics."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "mask_map *= mask_map_catalog\n",
     "mask_map.plot()"
    ]
   },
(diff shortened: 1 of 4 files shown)


QRemy:
Thanks Régis
About the mask from significant pixels, there is a nice feature of skimage we could illustrate: the hysteresis thresholding. It allows to define two thresholds and mask only the pixels between the low and high thresholds if they are not continuously connected to a pixel above the high threshold. This allows to better preserve the structure of the excesses than a strong cut at 5sigma and is more reliable than just lowering the threshold.

The code is quite short:
```python
from skimage import filters
sig_low = 2
sig_high = 4
hys_mask=filters.apply_hysteresis_threshold(result["significance"].data,sig_low,sig_high) 
mask_map = result["significance"].copy()
mask_map.data = ~hys_mask
```
Here is the HGPS significance map with 0.2deg correlation radius filtered using such hysteresis mask
image

QRemy:
> Thanks Régis
> About the mask from significant pixels, there is a nice feature of skimage we could illustrate: the hysteresis thresholding. It allows to define two thresholds and mask only the pixels between the low and high thresholds if they are not continuously connected to a pixel above the high threshold. This allows to better preserve the structure of the excesses than a strong cut at 5sigma and is more reliable than just lowering the threshold, or adding a fixed margin.
> 
> The code is quite short:
> 
> ```python
> from skimage import filters
> sig_low = 2
> sig_high = 4
> hys_mask=filters.apply_hysteresis_threshold(result["significance"].data,sig_low,sig_high) 
> mask_map = result["significance"].copy()
> mask_map.data = ~hys_mask
> ```
> 
> Here is the HGPS significance map with 0.2deg correlation radius filtered using such hysteresis mask
> image

tutorials/exclusion_mask.ipynb
```diff
@@ -0,0 +1,348 @@
...
+    "\n",
+    "**Objective: Build an exclusion mask around the Crab nebula excluding gamma-ray sources in the region.**\n",
+    "\n",
+    "### Proposed approach\n",
+    "\n",
+    "Here we have to build a `Map` object, we must first define its geometry and then we can determine which pixels to exclude.\n",
+    "\n",
+    "We can rely on known sources positions and properties to build a list of regions (here `~regions.SkyRegions`) enclosing most of the signal that our detector would see from these objects. We show below how to build this list manually or from an existing catalog. \n",
+    "\n",
+    "Finally, we show how to build the mask from a `MapDataset`, finding pixels which contain statistically significant signal. To do so, we use the `LiMaMapEstimator`\n",
```
AtreyeeS: Maybe you can add a line saying that the different sections show 3 different, independent methods of building masks.
Also, if users want, they can club different methods using simple map arithmetics, eg: `exclusion = exclusion1 * exclusion2`

tutorials/exclusion_mask.ipynb
```diff
@@ -0,0 +1,348 @@
...
+    "## Create the mask from a list of regions\n",
+    "\n",
+    "One can build an exclusion mask from regions. We show here how to proceed."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Define the geometry\n",
```
AtreyeeS: Maybe the definition of geometry should come before "Create the mask from a list of regions", as it is common to all methods

tutorials/exclusion_mask.ipynb
```diff
@@ -0,0 +1,348 @@
...
+    "We want to compute the significance per pixels for the data integrated over energy. We reduce the dataset to a simple image. "
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "datasets[0].psf = None\n",
```
AtreyeeS: Why do you remove the `psf`? To improve computation time? Then maybe this can be mentioned because it can get a bit distracting otherwise...

tutorials/exclusion_mask.ipynb
registerrier: Good point. Added to the notebook.

tutorials/exclusion_mask.ipynb
registerrier: At the moment the `MapDataset.to_image()` fails if the `psf` is a `PSFKernel`. I just added a test that do not perform the `PSFKernel.to_image` to avoid the error.

The `PSFKernel.to_image()` needs to be implemented and a special handling has to be performed in `MapDataset.to_image()`

registerrier:
I added the reference to the `skimage.apply_hysteresis_threshold` filter as a note. scikit-image is not a required dependency of gammapy. So we cannot run it without installing `skimage` first.

PR #2782, opened 2020-02-18
Add datasets_names attribute to cube models

QRemy:
Add datasets_names attribute to each 3d model.
Adapt serialization so the models.yaml file contains datasets list for each model and remove the models and backgrounds lists from the datasets.yaml file.
If model.datasets_names is  not set or set as "all" the model is applied to all datasets.

### DIFF
--- a/gammapy/cube/fit.py
+++ b/gammapy/cube/fit.py
@@ -740,14 +740,19 @@ class MapDataset(Dataset):
     def from_dict(cls, data, components, models):
         """Create from dicts and models list generated from YAML serialization."""
         dataset = cls.read(data["filename"], name=data["name"])
-        bkg_name = data["background"]
-        model_names = data["models"]
-        models_list = [model for model in models if model.name in model_names]
+        models_list = [
+            model
+            for model in models
+            if data["name"] in model.datasets_names or model.datasets_names == "all"
+        ]
         models = Models(models_list)
 
         for component in components["components"]:
             if component["type"] == "BackgroundModel":
-                if component["name"] == bkg_name:
+                if (
+                    dataset.name in component.get("datasets_names", [])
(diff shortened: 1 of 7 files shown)


gammapy/modeling/models/cube.py
```diff
@@ -73,6 +74,10 @@ def __init__(
         self.apply_irf = {"exposure": True, "psf": True, "edisp": True}
         if apply_irf is not None:
             self.apply_irf.update(apply_irf)
+        if datasets_names is not None:
```
adonath: As the default is an immutable str (`"all"`), you can just use it directly as a default, like `__init__(..., datasets_names="all")`.

gammapy/modeling/models/cube.py
```diff
@@ -348,7 +359,10 @@ def __init__(
         self.apply_irf = {"exposure": True, "psf": True, "edisp": True}
         if apply_irf is not None:
             self.apply_irf.update(apply_irf)
-
+        if datasets_names is not None:
```
adonath: See my comment above...

gammapy/cube/fit.py
```diff
@@ -740,14 +740,19 @@ def read(cls, filename, name=None):
     def from_dict(cls, data, components, models):
         """Create from dicts and models list generated from YAML serialization."""
         dataset = cls.read(data["filename"], name=data["name"])
-        bkg_name = data["background"]
-        model_names = data["models"]
-        models_list = [model for model in models if model.name in model_names]
+        models_list = [
```
adonath: I think this logic of selecting the model component should rather be moved to the the models setter of each dataset (search for `@models.setter`). This ensures a consistent behaviour between all datasets.

PR #2783, opened 2020-02-18
Add __call__ method to TemporalModel

AtreyeeS:
**Description**

This is a very simple PR to add a call method on the TemplateModel and renames `evaluate_norm_at_time()` to simply `evaluate`.

**Dear reviewer**

### DIFF
--- a/gammapy/modeling/models/temporal.py
+++ b/gammapy/modeling/models/temporal.py
@@ -22,6 +22,11 @@ class TemporalModel(Model):
 class TemplateTemporalModel(TemporalModel):
     """Template temporal model base class."""
 
+    def __call__(self, time):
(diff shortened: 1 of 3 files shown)


AtreyeeS:
> Thanks @AtreyeeS, looks good to me. Could you still add a test for the `__call__` method? E.g. just change the tests to call the model directly instead of calling `.evaluate()`.

Hi @adonath ! I did not want to do it because adding a `.integral()`, which the `__call__ ` will finally call, in the next PR will change the values...

adonath:
@AtreyeeS Just to clarify: the `__call__` interface does not call into `.integral()` but only `.evaluate()`. It is just for convenience and uniformity with the spectral and spatial models we have.

PR #2787, opened 2020-02-23
Add TemporalModel integral method

AtreyeeS:
**Description**

This pull request adds a `.integral()` on the temporal models.

**Dear reviewer**

There are a few issues I am not sure about.

1. The integral returned should physically have units of time. Should this method then return a `Quantity`?  At present, I am implicitly assuming that it returns in units of seconds, but am returning a float.
2. I have made some changes to the existing code and tests to ensure that the time handling is done properly between the times on the template model and the gtis passed - please do take a look!


AtreyeeS:
Thanks @adonath ! I have addressed the changes

Issue #2430, opened 2019-10-02
Add Windows CI build for notebooks and docs


cdeil:
In https://docs.gammapy.org/0.14/notebooks/hess.html#Model-fitting I see that we're using `!cat model-best-fit.yaml` to show the file contents. This will not work on Windows, it probably causes a "command not found" error. I'm not sure if there is a good cross-platform solution, I just now asked here: https://discourse.jupyter.org/t/why-does-cat-work-on-unix-but-not-windows/2094/4?u=cdeil

This issue is a reminder for myself to add a Windows CI build that executes and tests the notebooks, or to extend one of the existing Windows CI builds in [azure-pipelines.yml](https://github.com/gammapy/gammapy/blob/master/azure-pipelines.yml). I expect that this will reveal a few small issues like `!cat` not working on Windows -- those should be fixed in the same pull request where the CI build is added.

adonath:
In #2793 I added the notebook build for Azure.

PR #2793, opened 2020-02-28
Add notebooks test to azure pipelines

adonath:
**Description**

This pull request adds testing notebooks to our azure build. See #2430.


adonath:
@Bultako The `healpy` and `sherpa` dependencies are missing on Windows. Is there a way to skip the notebook test if a dependency is missing?

adonath:
Thanks @Bultako!

PR #2796, opened 2020-03-02
Change beta sign in SmoothBrokenPowerLawSpectralModel

QRemy:
add beta*=np.sign(index2 - index1) in SmoothBrokenPowerLawSpectralModel so  the hardening case is supported without changing beta sign.


QRemy:
By default beta is frozen so this problem won't show up.
I guess If ones unfreeze beta, they should also set the min to something positive. 
This could be done be done in the default, but I don't think that fitting beta is really needed or even relevant.

PR #2798, opened 2020-03-03
Add aeff-max safe energy threshold for MapDataset

luca-giunti:
**Description**
I implemented the missing method `SafeMaskMaker.make_mask_energy_aeff_max(...)` for the the case of `MapDataset`.

In the code , I work with the exposure instead of the effective area itself, which is equivalent in this case. I extract the exposure values at the given position, and fill up an `EffectiveAreaTable` on which I can call the `.find_energy()` method.

**Dear reviewer**
Should we pass two additional arguments `emin` and `emax` (as in `make_mask_energy_aeff_max(self, dataset, emin=None, emax=None)`) to narrow down the energy range range in case of non-unique solutions? These would be naturally forwarded to `aeff.find_energy(aeff_thres, emin, emax)`. This question is also valid for the `make_mask_energy_edisp_bias(...)` method.

### DIFF
--- a/gammapy/makers/safe.py
+++ b/gammapy/makers/safe.py
@@ -3,7 +3,8 @@ import logging
 import numpy as np
 from astropy.coordinates import Angle
 from gammapy.datasets import MapDataset
-from gammapy.maps import Map
+from gammapy.maps import Map, MapCoord
+from gammapy.irf import EffectiveAreaTable
 
 __all__ = ["SafeMaskMaker"]
 
@@ -113,11 +114,11 @@ class SafeMaskMaker:
         return mask
 
     def make_mask_energy_aeff_max(self, dataset):
-        """Make safe energy mask from aeff max.
+        """Make safe energy mask from effective area maximum value.
 
         Parameters
         ----------
-        dataset : `~gammapy.spectrum.SpectrumDataset` or `~gammapy.spectrum.SpectrumDatasetOnOff`
+        dataset : `~gammapy.modeling.Dataset`
             Dataset to compute mask for.
 
         Returns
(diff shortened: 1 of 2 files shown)


luca-giunti:
> Thanks @luca-giunti, this looks good to me!
> 
> Just an idea: we could extend `WcsNDMap.get_spectrum()` to implement the case for `PointSkyRegion` and then call `dataset.to_spectrum_dataset()`. This would simplify the code a lot...

Hi! True, but `dataset.to_spectrum_dataset()` would compute a lot of things we don't need, so for performances reasons I chose this (more verbose) approach...

gammapy/makers/safe.py
```diff
@@ -113,29 +114,46 @@ def make_mask_energy_aeff_default(dataset, observation):
...
         if isinstance(dataset, MapDataset):
-            raise NotImplementedError(
-                "'aeff-max' method currently only supported for spectral datasets"
+            position = self.position
+            if position is None:
+                position = dataset.counts.geom.center_skydir
+            exposure = dataset.exposure
+            position = position.transform_to("icrs")
+            energy = exposure.geom.get_axis_by_name("energy_true")
+            coord = MapCoord.create(
```
adonath: You can use `MapCoord.create({"skycoord": position, "energy": energy.center})` directly and avoid the coordinate transform before...

adonath:
@luca-giunti That's why I proposed the short-cut for the `PointSkyRegion` (to avoid computing a mask...) and we could still call `exposure.get_spectrum()` directly. But I see your concerng, computing e.g. the `EDispKernel` could take some time...I'm fine to leave as is.

I've left one more inline comment, otherwise looks good to me.

gammapy/makers/safe.py
luca-giunti: Done! Thanks

adonath:
Thanks @luca-giunti, no further comments from my side...

PR #2799, opened 2020-03-03
Adding the functionality to sample the background only in MapDatasetEventSampler

fabiopintore:
Currently the `MapDatasetEventSampler` class allows the users to sample sources and background events or, alternatively, source events only. However, the possibility to sample background events only is not taken into account yet.
This PR aims at offering this functionality to the users.

### DIFF
--- a/gammapy/cube/simulate.py
+++ b/gammapy/cube/simulate.py
@@ -264,8 +264,7 @@ class MapDatasetEventSampler:
         events : `~gammapy.data.EventList`
             Event list.
         """
-        try:
-          if dataset.models[1]:
+        if len(dataset.models) > 1:
             events_src = self.sample_sources(dataset)
 
             if dataset.psf:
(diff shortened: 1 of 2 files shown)


gammapy/cube/simulate.py
```diff
@@ -263,24 +264,31 @@ def run(self, dataset, observation=None):
...
-            events_src = self.sample_edisp(dataset.edisp, events_src)
-        else:
-            events_src.table["ENERGY"] = events_src.table["ENERGY_TRUE"]
-
-        if dataset.background_model:
-            events_bkg = self.sample_background(dataset)
-            events = EventList.stack([events_bkg, events_src])
-        else:
-            events = events_src
+        try:
```
adonath: When using the `try` `except` statement it is important to keep the code in the block as short as possible, otherwise all errors will be silently caught. If possible we should avoid it completely...

Can you change the implementation check e.g.
```
if len(dataset.models) > 1 and dataset.background_model is not None:
``` 
This should be simpler...

gammapy/cube/tests/test_simulate.py
```diff
@@ -50,6 +52,49 @@ def dataset():
     return dataset
 
 
+@pytest.fixture(scope="session")
```
adonath: Please avoid the copy and pasted code here and just add the case with the background model only to one of the existing tests...

gammapy/cube/simulate.py
fabiopintore: Okay, thanks. I've removed it. 
However, I suggest that the condition should be `if len(dataset.models) > 1:` because, nested to that, there is already the condition `if dataset.background_model:`.

gammapy/cube/tests/test_simulate.py
fabiopintore: Ok.

PR #2807, opened 2020-03-04
Add filename argument in BackgroundModel().from_dict()

QRemy:
Fix minor bug in BackgroundModel serialization: filename argument was not set when reading background using .from_dict()


adonath:
@QRemy There is a strange `EOFError` on Travis and Azure. I first thought it was a fluke, but I think it's an actual error. Right now I have no idea what this could be...

PR #2809, opened 2020-03-04
Add sampling of temporal models to event sampler

fabiopintore:
This PR aims at modifying the `MapDatasetEventSamplers.sample_sources()` function, in order to allow the users to sample the time of the source events from a given `dataset.temporal_model`. Currently, the class just adopt a `ConstantTemporalModel`.

### DIFF
--- a/gammapy/datasets/simulate.py
+++ b/gammapy/datasets/simulate.py
@@ -77,11 +77,10 @@ class MapDatasetEventSampler:
             evaluator.model.apply_irf["edisp"] = False
             npred = evaluator.compute_npred()
 
-            # temporal_model = ConstantTemporalModel()
-            if model.temporal_model is None:
-                temporal_model = ConstantTemporalModel()
-            else:
+            if hasattr(model, "temporal_model"):
                 temporal_model = model.temporal_model
+            else:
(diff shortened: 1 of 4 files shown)


fabiopintore:
> Thanks @fabiopintore! This definitely needs a test. Please add a temporal model component to the test model [here](https://github.com/gammapy/gammapy/blob/master/gammapy/datasets/tests/test_simulate.py#L31) and add an assert for the time values of the events...

I've included the temporal model to the dataset and added a test on the times. However, it's not clear to me why some of the other tests changed the values of the sampled events.

adonath:
@fabiopintore The Travis build fails with a `MemoryError` could maybe investigate why? If there is no bug, you could maybe make the test dataset smaller...

fabiopintore:
> @fabiopintore The Travis build fails with a `MemoryError` could maybe investigate why? If there is no bug, you could maybe make the test dataset smaller...

The dataset is quite small (it's around 400 KB). Not clear to me why Travis says `MemoryError: Unable to allocate 28.2 GiB ...`. There is nothing so big. And the tests are not failing locally. The Travis error is not clear at all.

fabiopintore:
Sorry, I closed the PR by mistake.

PR #2810, opened 2020-03-04
Flux estimator

registerrier:
**Description**

This pull request introduces the `FluxEstimator`.
So far result return is in the form of a `dict`. Do we prefer to have a `Table` row instead?
**Dear reviewer**

Some work still needed. 
The early PR should help test with the new scipy 1.4.1.

### DIFF
--- a/gammapy/estimators/flux_estimator.py
+++ b/gammapy/estimators/flux_estimator.py
@@ -156,6 +156,6 @@ class FluxEstimator:
             "ref_e2dnde": self.ref_model(e_ref) * e_ref ** 2,
         }
 
-        result.update(self.estimator.run(self.model.parameters['norm'], steps, self.norm_values))
+        result.update(self.estimator.run(self.model.parameters['norm'], steps, null_value=0, scan_values=self.norm_values))
         return result
 
--- a/gammapy/estimators/flux_estimator.py
+++ b/gammapy/estimators/flux_estimator.py
@@ -10,7 +10,7 @@ __all__ = ["FluxEstimator"]
 
 log = logging.getLogger(__name__)
 
-class FluxEstimator:
+class FluxEstimator(ParameterEstimator):
     """Flux estimator.
 
     Estimates flux for a given list of datasets with their model in a given energy range.
@@ -64,9 +64,9 @@ class FluxEstimator:
                 " of the same type and data shape."
             )
 
-        self.datasets = datasets.copy()
+        datasets = datasets.copy()
 
-        dataset = self.datasets[0]
+        dataset = datasets[0]
 
         model = dataset.models[source].spectral_model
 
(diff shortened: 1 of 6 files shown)


gammapy/estimators/flux_estimator.py
```diff
@@ -0,0 +1,161 @@
...
+    sigma_ul : int
+        Sigma to use for upper limit computation.
+    reoptimize : bool
+        Re-optimize other free model parameters.
+    """
+
+    def __init__(
+            self,
+            datasets,
+            source=0,
```
adonath: Maybe we should make the `source` argument mandatory?

gammapy/estimators/flux_estimator.py
```diff
@@ -0,0 +1,161 @@
...
+    def __str__(self):
+        s = f"{self.__class__.__name__}:\n"
+        s += str(self.datasets) + "\n"
+        s += str(self.model) + "\n"
+        return s
+
+    def _set_scale_model(self, datasets):
+        # set the model on all datasets
+        for dataset in datasets:
+            if len(dataset.models) > 1:
```
adonath: I think this `if...else` block can go just indexing into `models` should be fine.

gammapy/estimators/flux_estimator.py
```diff
@@ -0,0 +1,161 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+import logging
+import numpy as np
+from astropy import units as u
+from gammapy.datasets import Datasets
+from gammapy.modeling.models import ScaleSpectralModel
+from gammapy.estimators import ParameterEstimator
```
adonath: Use a relative import here like `from .parameter_estimator import ParameterEstimator`...

gammapy/estimators/flux_estimator.py
```diff
@@ -0,0 +1,161 @@
...
+
+__all__ = ["FluxEstimator"]
```
adonath: For now I would suggest to make the `FluxEstimator` not public and keep it as an implementation detail of the `LightCurveEstimator` and `FluxPointsEstimator`, which expose the end-user API.

gammapy/estimators/flux_estimator.py
registerrier: Done

gammapy/estimators/flux_estimator.py
registerrier: Done, but kept absolute import

gammapy/estimators/flux_estimator.py
registerrier: Removed `FluxEstimator` and `ParameterEstimator` from exposed API

gammapy/estimators/flux_estimator.py
registerrier: It's gone.

PR #2812, opened 2020-03-05
Add ring background estimation in high-level interface.

Bultako:
**Description**

This pull request adds the possibility to choose the "ring background estimation" for 3D analysis in the high-level interface.

### DIFF
--- a/gammapy/analysis/config/docs.yaml
+++ b/gammapy/analysis/config/docs.yaml
@@ -39,11 +39,11 @@ datasets:
             energy_true: {min: 0.1 TeV, max: 10 TeV, nbins: 30}
     map_selection: ['counts', 'exposure', 'background', 'psf', 'edisp']
     background:
-        method: reflected   # also fov_background
+        method: ring        # also fov_background, reflected for 1d
         exclusion:          # fits file for exclusion mask
-        parameters: {}
+        parameters: {r_in: 0.3 deg, width: 0.3 deg} # ring
     safe_mask:
-            methods: ['aeff-default','offset-max' ]
+            methods: ['aeff-default', 'offset-max']
             parameters: {offset_max: 2.5 deg}
     on_region: {frame: icrs, lon: 83.633 deg, lat: 22.014 deg, radius: 3 deg}
     containment_correction: true
--- a/gammapy/analysis/config/docs.yaml
+++ b/gammapy/analysis/config/docs.yaml
@@ -41,7 +41,7 @@ datasets:
     background:
         method: ring        # also fov_background, reflected for 1d
         exclusion:          # fits file for exclusion mask
-        parameters: {r_in: 0.3 deg, width: 0.3 deg} # ring
+        parameters: {r_in: 0.7 deg, width: 0.7 deg} # ring
     safe_mask:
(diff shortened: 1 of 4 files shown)


Bultako:
Thanks @registerrier 

If I understand correctly, maybe having different values for the `stack` keyword could solve this. For example, for the use case you expose `stack: first` would stack the datasets and then compute the background on them,  `stack: last` would compute the background as we append and stack the datasets (this is what we do now with `stack: True`), no value for `stack` or no keyword `stack` will behave in the same way as `stack: False` now.

Does this make sense? 
Do you identify other uses cases for `stack: first`? 
with other background estimators?

adonath:
@registerrier and @Bultako For now I would only support the case of image analysis in the HLI i.e. sum the images before applying the ring. Applying the ring in energy bands is really a non-standard use-case and I would not support it in the HLI.

AtreyeeS:
I tend to agree with @adonath that we only support image analysis of the ring in the HLI for now. We have not really tested how the ring behaves for 3D maps, so perhaps it is a bit early to support `stack: last` options at the moment...

registerrier:
Then the question is whether we can really on making maps with a geometry carrying only one energy bin or if we have to go through `map_dataset.to_image()` before applying `RingBackgroundMaker`.

adonath:
@registerrier You are right. My proposal would be add an oversampling option for the exposure to the `MapDatasetMaker`. Or rather rename `background_oversampling` to just `oversampling` and also apply to the exposure...

gammapy/analysis/config/docs.yaml
```diff
@@ -39,11 +39,11 @@ datasets:
             energy_true: {min: 0.1 TeV, max: 10 TeV, nbins: 30}
     map_selection: ['counts', 'exposure', 'background', 'psf', 'edisp']
     background:
-        method: reflected   # also fov_background
+        method: ring        # also fov_background, reflected for 1d
         exclusion:          # fits file for exclusion mask
-        parameters: {}
+        parameters: {r_in: 0.3 deg, width: 0.3 deg} # ring
```
registerrier: Maybe the default `r_in` is too small. How about 0.7 deg for instance?

adonath:
As discussed in the dev meeting today we will aim for the solution to treat image analysis as a cube analysis with one reconstructed energy bin. This means the implementation in this PR can basically stay as is, but we need to add more documentation. 

I noticed that the current implementation gives incorrect results, because the `RingBackgroundMaker` returns a `MapDatasetOnOff`, which is stacked into `MapDataset`. Technically this is well defined, because one can just stack the counts, background, exposure PSF and EDISP, but right now the result is not correct, because the `MapDatasetOnOff` does not define a `BackgroundModel`. There are now two possible fixes:
- After running the `RingBackground` one can define the background model like `dataset.models = Models([BackgroundModel(dataset.background)])`
- Or we change `MapDataset.stack()` to handle this correctly...

We have to refactor it anyway, once we change the background model handling.

Bultako:
@registerrier @adonath 
I've made some modifs in the line proposed in https://github.com/gammapy/gammapy/pull/2812#issuecomment-598736329
You may continue the review.

PR #2814, opened 2020-03-06
Add a tutorial for the event sampler

fabiopintore:
I request to include a tutorial for the event sampler. The tutorial should show how to sample events from a given Sky model, how to save the events into a file and compare the sampled events with the simulated models.


tutorials/event_sampler_tutorial.ipynb
```diff
@@ -0,0 +1,582 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Create the dataset\n",
+    "In this section we show how to create the `Dataset` object needed for the event sampling (for more info about to create `Dataset` objects, please visit this [link](https://docs.gammapy.org/dev/notebooks/analysis_2.html#Preparing-reduced-datasets-geometry)). \n",
+    "\n",
+    "Hereafter, we select the IRF from the South configuration of the CTA DC1 to simulate one observation. We set the pointing position of the simulated field at the Galactic Center and we fix the exposure time to 8 hr. "
+   ]
```
AtreyeeS: These details might reside on the Proposed Approach

tutorials/event_sampler_tutorial.ipynb
```diff
@@ -0,0 +1,582 @@
...
+    "result.parameters.to_table()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The results are excellent as all the observed parameters are consistent withing 1 sigma with the simulated ones."
+   ]
+  },
```
AtreyeeS: Perhaps remove this section on extended source simulation and add it as suggested exercises? The methodology doesn't really change...

tutorials/event_sampler_tutorial.ipynb
```diff
@@ -0,0 +1,582 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# dataset config\n",
+    "ENERGY_AXIS = MapAxis.from_energy_bounds(\"0.1 TeV\", \"100 TeV\", nbin=10, per_decade=True)\n",
```
AtreyeeS: Is the `e_reco` binning critical for the event sampling?

tutorials/event_sampler_tutorial.ipynb
```diff
@@ -0,0 +1,582 @@
...
+    "counts = Map.create(frame=\"galactic\", skydir=(0, 0.), binsz=0.02, npix=(150, 150))\n",
+    "counts.fill_events(events)\n",
+    "counts.plot(add_cbar=True)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Fit the simulated data\n",
```
AtreyeeS: To keep the tutorial short, perhaps you can just show how to fill `counts.dataset` with the simulated map, and then write the dataset to disc. Explain that at this stage, we have standard reduced datasets which we can fit using standard techniques, eg: `simulate_3d` tutorial.

tutorials/event_sampler_tutorial.ipynb
```diff
@@ -0,0 +1,582 @@
...
+   "outputs": [],
+   "source": [
+    "events.peek()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Let's write the photon event list to a FITS file:"
```
AtreyeeS: At this stage, we essentially have a simulated observation.
Perhaps link to `analysis_2` tutorial, showing how to use `Datastore.from_events_files()` to use these simulated files?

tutorials/event_sampler_tutorial.ipynb
fabiopintore: I know this lengthen the tutorial. I was thinking to present also the fitting process in order to show that the observed parameters are compatible with the simulated ones, as an on the fly cross-check. 
However, I can also add the reference to the `simulate_3d` tutorial.

tutorials/event_sampler_tutorial.ipynb
fabiopintore: Yes, it should be very well sampled.

tutorials/event_sampler_tutorial.ipynb
fabiopintore: Right! I leave only the point-like case and the template case.

tutorials/event_sampler_tutorial.ipynb
fabiopintore: Okay, I move it.

tutorials/event_sampler_tutorial.ipynb
fabiopintore: I agree. I've added it.

> Thanks @fabiopintore for this tutorial. If I may suggest a few changes, the new format of having Prerequisites, Context, Objective and Proposed Approach seems to be useful to users. See, eg: https://github.com/gammapy/gammapy/blob/master/tutorials/modeling_2D.ipynb or https://github.com/gammapy/gammapy/blob/master/tutorials/light_curve_flare.ipynb
> 
> It would be nice to reformat the Introduction into this format.
> 
> Also, add this tutorial in https://github.com/gammapy/gammapy/blob/master/tutorials/notebooks.yaml

Thanks @AtreyeeS for the comments. I updated the tutorial following them.

adonath:
@fabiopintore I checked once again and the total run-time of the tutorial is currently ~70s. Unfortunately this is too long. Here are a few suggestions for the reduction of the runtime:
- Remove the fitting step and just replace it with a link to one of the other tutorials
- Remove reading  and writing the dataset, and just use the one you have in memory already, if you want to keep the I/O, just write to a plain FITS file (and not `fits.gz`). The zip compression and de-compresssion takes time...
- Possibly reduce the dataset size to 3 deg x 3 deg

adonath:
@fabiopintore There is something wrong with the commit history in this PR. The simplest solution would be to just restart from the current master and copy over the event sampling tutorial...

fabiopintore:
> @fabiopintore There is something wrong with the commit history in this PR. The simplest solution would be to just restart from the current master and copy over the event sampling tutorial...

@adonath Are you suggesting to create a new branch and then open a new PR linked to this one?

adonath:
@fabiopintore Or force push to this one...as you prefer...

adonath:
Superseded by #2829.

PR #2815, opened 2020-03-06
Remove CountsSpectrum class

adonath:
**Description**

This pull request removes the `CountsSpectrum` class and replaces its usage with the `RegionNDMap`. There are minimal changes in the test values probably related to the refactoring
of the `SpectrumEvaluator` 

**Dear reviewer**


adonath:
@registerrier I adapted the `FluxPointsEstimator`, which might cause merge conflicts with your work, but the changes are small. Please let me know if I can help to resolve those. I x-failed the `TestFluxPointEstimator.test_run_pwl` test, where I observed a weird behaviour. After the changes introduced here, the first flux point resulted in `np.nan`, but checking the likelihood and best fit value everything looked fine, but for some reason `iminuit` reports the fit as "failed". Maybe you can re-active the test...

To avoid further merge conflicts, I'll merge this PR now.

Bultako:
@adonath 

The `fake` methods in `SpectrumDatasetOnOff` and `MapDataSetOnOff` still need a `background_model` parameter, now as type `RegionNDMap` 

https://github.com/gammapy/gammapy/blob/4987620e98242280541c087bc6a79853041adc3d/gammapy/datasets/spectrum.py#L768

https://github.com/gammapy/gammapy/blob/4987620e98242280541c087bc6a79853041adc3d/gammapy/datasets/map.py#L1258

Shouldn't we remove this parameter and use `self.counts` instead?

adonath:
@Bultako I think the plan is rather that we implement a better background model handling for `SpectrumDataset`. For `MapDataset` the background model is already part of the `models` declaration. So the logic will become the same:

```
bkg = BackgroundModel()
dataset = SpectrumDataset(models=[bkg])
dataset.fake()
```
And
```
bkg = BackgroundModel()
dataset = SpectrumDatasetOnOff(models=[bkg])
dataset.fake()
```
In this case there is no need to pass any background model to `.fake()` anymore.

PR #2817, opened 2020-03-07
Remove spectrum evaluator

adonath:
**Description**

This pull request removes the `SpectrumEvaluator` and unifies the model evaluation with the `MapEvaluator`. I includes the following changes:
- I added  a `SkyModel.integrate(geom)` and `SpatialModel.integrate(geom)` method
- I changed `SkyModel.integrate()` to call into `SpectralModel.integral()` as well. This causes a lot of small (and some major) changes in the test values, because for many test we have use large energy bins.
- I added a `SpectrumDataset.exposure` property

**Dear reviewer**
I think its important, that we re-run the validation on this branch to see, what changes it causes to the result before we merge. @luca-giunti @AtreyeeS @registerrier @QRemy

### DIFF
--- a/gammapy/datasets/tests/test_map.py
+++ b/gammapy/datasets/tests/test_map.py
@@ -154,8 +154,8 @@ def test_fake(sky_model, geom, geom_etrue):
     dataset.fake(314)
 
     assert real_dataset.counts.data.shape == dataset.counts.data.shape
-    assert_allclose(real_dataset.counts.data.sum(), 8220.399727)
-    assert_allclose(dataset.counts.data.sum(), 8375)
+    assert_allclose(real_dataset.counts.data.sum(), 9525.755807)
+    assert_allclose(dataset.counts.data.sum(), 9723)
 
 
 @pytest.mark.xfail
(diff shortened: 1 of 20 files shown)


gammapy/datasets/tests/test_map.py
```diff
@@ -426,19 +426,19 @@ def test_map_fit_one_energy_bin(sky_model, geom_image):
     assert result.success
 
     npred = dataset.npred().data.sum()
-    assert_allclose(npred, 4076.779039, rtol=1e-3)
-    assert_allclose(result.total_stat, 5722.439112, rtol=1e-3)
+    assert_allclose(npred, 16538.124036, rtol=1e-3)
+    assert_allclose(result.total_stat, -34844.125047, rtol=1e-3)
```
AtreyeeS: Is this change expected? Going to negative TS?

gammapy/modeling/models/cube.py
```diff
@@ -196,15 +196,34 @@ def evaluate_geom(self, geom, gti=None):
...
+        ).reshape((-1, 1, 1))
+
+        if self.spatial_model and not isinstance(geom, RegionGeom):
+            # TODO: integrate spatial model over region to correct for
+            #  containment
+            geom_image = geom.to_image()
+            value = value * self.spatial_model.integrate(geom_image).quantity
+
+        if self.temporal_model:
+            integral = self.temporal_model.integral(gti.time_start, gti.time_stop)
```
AtreyeeS: Perhaps we should uniform naming between `.integrate` and `.integral` at some point?

adonath: The API is different, because one takes `Geom` object and returns a `Map`, the other just works with numpy arrays. I could rename `.integrate()` to `.integrate_geom()` to avoid confusion?

adonath: I've done the rename for now...

PR #2820, opened 2020-03-10
Replace LiMaMapEstimator with ExcessMapEstimator

registerrier:
**Description**

This pull request replaces the `LiMaMapEstimator` with a more general `CorrelatedExcessMapEstimator`. The later takes into account the relevant `CountsStatistic` of the `MapDataset` or `MapDatasetOnOff`. 
As discussed with @adonath , the choice is made that arguments for estimators are all passed on `__init__`, while the `run` method only accepts a `steps` argument. This should allow for serialization of estimators in the future. 
For convenience, a property setter was added for the correlation radius and dataset argument.
It is now possible to compute error maps, asymmetric error maps and upper limit maps.

The `compute_significance_map_on_off` and `compute_significance_map` functions are removed. It is therefore necessary to use a `MapDataset` for significance map estimation.

**Dear reviewer**

The PR should be ready.


AtreyeeS:
Thanks @registerrier ! This is indeed what was asked for during the user feedback session! Users will not have to worry about kernel shapes etc. 
But how does this work for custom likelihoods, eg: the `DMDatasets` being prepared by @Bultako ?

Bultako:
@AtreyeeS 
In DM simulation analysis the custom likelihoods are now built on `wstat` function implemented in on/off datasets.

https://github.com/gammapy/gammapy/blob/b4b548c0ff24ce7246e2acac0de524510d15e99f/gammapy/stats/fit_statistics.py#L106

I understand this  `wstat` function  will still remain.

registerrier:
Yes indeed, `wstat` is there to remain of course. 
The future new `Dataset` classes can indeed be a problem in the current implementation. 
I guess that the solution is to include the relevant `CountsStatistic` class to the `Dataset`. In practice the `CountsStatistic` contain the `wstat` or `cash` function. Everything could go through this class.

PR #2826, opened 2020-03-13
Add notebook on statistics in gammapy

registerrier:
**Description**

This pull request adds a notebook to demonstrate `CountsStatistic` functions and how we compute significances from delta TS as well as confidence errors on excess counts.
We can show some plots of cash and wstat statistic profiles to help a user with little statistic knowledge how calculations are performed and what our TS, sqrt TS, significance and errors mean. 
This is complementary to the modeling notebook which shows how this works in the context of model fitting.

**Dear reviewer**

### DIFF
--- a/docs/tutorials/index.rst
+++ b/docs/tutorials/index.rst
@@ -110,6 +110,7 @@ Advanced tutorials
    ../notebooks/exclusion_mask.ipynb
    ../notebooks/overview.ipynb
    ../notebooks/maps.ipynb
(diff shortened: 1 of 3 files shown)


registerrier:
Thanks @adonath . I agree that most of the notebook is either text of the plotting code. I hesitated initially and decided to give the notebook a try. But the result does indeed suggest that a documentation page is more efficient here. 
I will close the PR and open a new one with a rst page instead.

registerrier:
Superseded by #2839

PR #2833, opened 2020-03-19
Add covariance serialization

QRemy:
- add Covariance.read() and modify Covariance.write() to work with models serialization
- remove .to_table() which is broken if parameters have duplicate names.
  For now unique names are generated on .write() using make_name() but once 
  models won't have duplicate parameters names, we could use param_name +
 model_name that will make the covariance matrix files more readable.

### DIFF
--- a/gammapy/modeling/models/core.py
+++ b/gammapy/modeling/models/core.py
@@ -239,7 +239,7 @@ class Models(collections.abc.MutableSequence):
         path = make_path(path)
         if path.exists() and not overwrite:
             raise IOError(f"File exists already: {path}")
-        if self._covariance is not None:
+        if self._covariance is not None and len(self.parameters) != 0:
             filename = splitext(str(path))[0] + "_covariance.dat"
             self._covariance.write(self._models, filename)
         path.write_text(self.to_yaml())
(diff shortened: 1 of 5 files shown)


gammapy/modeling/models/core.py
```diff
@@ -274,7 +282,13 @@ def to_dict(self):
         for model in self._models:
             model_data = model.to_dict()
             models_data.append(model_data)
-        return {"components": models_data}
+        if self._covariance.filename is not None:
+            return {
+                "components": models_data,
+                "covariance": str(self._covariance.filename),
```
adonath: I think here we can just rely on a default filename create from the models filename like `models.yaml` and `models_covariance.ecsv` or `my_model.yaml` and `my_model_covariance.ecsv`. More flexibility is then provided with the `Model.write_covariance()` and `Model.read_covariance()` API.

adonath:
Just to add: the `Covariance.read()` and `Covariance.write()` can be removed then...

gammapy/modeling/models/core.py
QRemy: I define the name on models.write() using the same prefix than for the models.yaml file, now it is passed  through self._covar_file

adonath: OK, sounds good...

PR #2841, opened 2020-03-27
Clean up of model parameter handling

adonath:
**Description**

This pull request does a bit of (random) cleanup to the model parameter handling:
- Use direct attribute access of models internally (`self.norm` instead of `self.parameters["norm"]`)
- Allow to initialise model parameters with a `Parameter` object (so far only for internal use...)
- Add `Model.from_parameters()` for convenience...


QRemy:
The attributes are defined on init and not updated if the parameters object is modified, by set_link() for example, while to self.parameters["..."] always works no ?

adonath:
Thanks @QRemy, you are right. This is exactly the remaining problem (and actually was the initial motivation for this PR...). So currently there can still be a duplicated (and out of sync) parameter state, because parameters are stored independently on the model as attribute and in the `model.parameters` list. I remember that you have seen exactly this problem in one of your analyses. Do you maybe have a minimal example that I can use as a unit test to fix this behaviour?

QRemy:
Here is an example for the link problem
```python
from gammapy.modeling.models import DiskSpatialModel, GaussianSpatialModel
g1 = GaussianSpatialModel(sigma=1*u.deg)
g2 = GaussianSpatialModel(sigma=2*u.deg)
g1.parameters.link(g1.sigma,g2.sigma)
print(g1.parameters["sigma"])
print(g1.sigma)
```
Moving the link function to the model level and we could update the attributes and solve this. Is there other functions that modify the parameters object ?

adonath:
Thanks @QRemy! I have decided to change to the following design:

- Model parameters are attached to the model instance and `.parameters()` just collects those
- Remove `Parmeters.link()`. Parameter linking is now achieved by directly setting the parameter objects like
```

pwl1 = PowerLawSpectralModel(index=2)
pwl2 = PowerLawSpectralModel(index=1)
pwl1.index = pwl2.index

# or 

pwl1 = PowerLawSpectralModel(index=2)
pwl2 = PowerLawSpectralModel(index=pwl1.index)

``` 
Let me know if you think there are other problem involved, otherwise I'd like to go ahead an merge the PR soon.

QRemy:
Thanks it looks fine, no further comments.

adonath:
Thanks @QRemy!

PR #2843, opened 2020-03-28
Adapt ExcessMapEstimator to estimators init scheme

registerrier:
**Description**

This pull request changes the `__init__` method of `ExcessMapEstimator` and moves the `dataset` argument to the `run()` function. 

**Dear reviewer**

### DIFF
--- a/gammapy/estimators/excess_map.py
+++ b/gammapy/estimators/excess_map.py
@@ -44,8 +44,6 @@ class ExcessMapEstimator:
 
     Parameters
     ----------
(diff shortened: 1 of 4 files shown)


gammapy/estimators/excess_map.py
```diff
@@ -81,11 +68,13 @@ def correlation_radius(self, correlation_radius):
         """Sets radius"""
         self._correlation_radius = Angle(correlation_radius)
 
-    def run(self, steps="all"):
+    def run(self, dataset, steps="all"):
```
adonath: Please move the `steps` argument to init as well then we have full consistency with what we have for our `Maker` classes..

PR #2838, opened 2020-03-26
Fix fails in Azure pipelines CI 

Bultako:
**Description**

This pull request tries to fix fails in Azure pipelines CI because of *files not found* errors due to *429 too many requests* errors when downloading GAMMAPY_DATA.

It just adds a cache for `GAMMAPY_DATA` folder.

**Dear reviewer**

This is a work-in progress PR.
When azure CI tests turn into green we could merge it.

PR #2847, opened 2020-03-30
Download bundle compressed datasets file

Bultako:
**Description**

This pull request significantly modifies the way how the datasets are downloaded internally with `gammapy download datasets` and `gammapy download tutorials`. It solves latest issues found with parallelized download of files due to *"[429] too many requests"* errors that Github throws (see also https://github.com/gammapy/gammapy/pull/2838)

The datasets are now downloaded as a single compressed file hosted in the new [gammapy-data Github repository](https://github.com/gammapy/gammapy-data), and then extracted into the destination folder. This modification is completely transparent to the user since  `gammapy download` keeps the same options, parameters and behaves in the same way in the front-end. 

On the other hand, it changes the development workflow for PRs where datasets are added or modified. In these rare cases, a previous sync  with [gammapy-data Github repository](https://github.com/gammapy/gammapy-data) has to be done, as well as re-tagging the `master` tag.

### DIFF
--- a/dev/datasets/gammapy-data-index.json
+++ b/dev/datasets/gammapy-data-index.json
@@ -7,64 +7,55 @@
     "path": "cta-1dc/README.md",
     "url": "https://github.com/gammapy/gammapy-extra/raw/master/datasets/cta-1dc/README.md",
     "filesize": 497,
-    "hashmd5": "64cde583f9a4f2c655d008a0c6a69da3",
-    "itempath": "/Users/jer/git/gammapy/gammapy-extra/datasets/cta-1dc/README.md"
+    "hashmd5": "64cde583f9a4f2c655d008a0c6a69da3"
    },
    {
     "path": "cta-1dc/make.py",
     "url": "https://github.com/gammapy/gammapy-extra/raw/master/datasets/cta-1dc/make.py",
     "filesize": 1940,
-    "hashmd5": "3d5004b41f25d0fbc98d8307c30932cb",
-    "itempath": "/Users/jer/git/gammapy/gammapy-extra/datasets/cta-1dc/make.py"
+    "hashmd5": "3d5004b41f25d0fbc98d8307c30932cb"
    },
    {
     "path": "cta-1dc/caldb/data/cta/1dc/bcf/South_z20_50h/irf_file.fits",
     "url": "https://github.com/gammapy/gammapy-extra/raw/master/datasets/cta-1dc/caldb/data/cta/1dc/bcf/South_z20_50h/irf_file.fits",
     "filesize": 3755520,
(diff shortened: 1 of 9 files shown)


docs/development/howto.rst
```diff
@@ -74,6 +71,22 @@ Skip unit tests for some Astropy versions
...
+`gammapy-data Github repository <https://github.com/gammapy/gammapy-data>`__. It is recommended that developers
+have `$GAMMAPY_DATA` environment variable pointing to the local folder where they have fetched the `gammapy-data` Github
+repository, so they can push and pull eventual modification of its content. In the cases they want to make a pull
+request with new or modified datasets, these are the steps to follow:
+
+#. Update the content of the remote specific dataset repository (i.e. `gammapy-extra`, `gammapy-fermi-lat-data`,
+   `gamma-cat`).
+#. Update the content of your local `gammapy-data` repository and JSON index file by runinng
+   `python make_datasets_index.py` in the `dev/datasets/` folder.
+#. Push eventual changes in `gammapy-data` to remote repository and re-tag the `master` tag.
```
adonath: Why is the master tag needed here? Can't we just download from the master branch?

dev/datasets/gammapy-data-index.json
```diff
@@ -7,55 +7,64 @@
     "path": "cta-1dc/README.md",
     "url": "https://github.com/gammapy/gammapy-extra/raw/master/datasets/cta-1dc/README.md",
     "filesize": 497,
-    "hashmd5": "64cde583f9a4f2c655d008a0c6a69da3"
+    "hashmd5": "64cde583f9a4f2c655d008a0c6a69da3",
+    "itempath": "/Users/jer/git/gammapy/gammapy-extra/datasets/cta-1dc/README.md"
```
adonath: Why is this local path needed? I think we should clearly avoid writing user specific path information for the index file...do you have any idea how to avoid this?

docs/development/howto.rst
Bultako: I don't know how to download the bundle tar.gz file without making a *tag* in the repository.

dev/datasets/gammapy-data-index.json
Bultako: You're right.
I've made a commit with the modif that removes this local path from the index file.

docs/development/howto.rst
adonath: I think we can just use the url `https://github.com/gammapy/gammapy-data/tarball/master`...

Bultako: Yes!!
I've pushed a commit with the modifs :)

adonath: Great thanks!

PR #2848, opened 2020-03-30
Adapt FluxEstimator, FluxPointsEstimator and ParameterEstimator

registerrier:
**Description**

This pull request adapts the `FluxEstimator`, `FluxPointsEstimator` and `ParameterEstimator` to the `__init__` and `run` schemes of the `estimators` sub package.
The inheritance scheme is maintained, and dataset copy is limited to `FluxPointsEstimator` only. 

Notebooks and high level interface have been modified accordingly.

**Dear reviewer**

We should check with the benchmarking that the performances are not degraded.


registerrier:
After checking with `gammapy-benchmark`, this does not really change the CPU time and the memory load.

Issue #2859, opened 2020-04-08
Linking parameters of two models


tunbehaun273:
**Gammapy version 0.17**

**Bug description**
When fitting two datasets with two different models, but linking the parameters of the models `(model1.amplitude = model2.amplitude)` and starting a fit I get the following error: 

```
ValueError                                Traceback (most recent call last)
<timed exec> in <module>

/home/woody/caph/mppi086h/software/anaconda3/envs/gammapy-0.17/lib/python3.7/site-packages/gammapy/modeling/fit.py in run(self, backend, optimize_opts, covariance_opts)
    112             return optimize_result
    113 
--> 114         covariance_result = self.covariance(backend, **covariance_opts)
    115         # TODO: not sure how best to report the results
    116         # back or how to form the FitResult object.

/home/woody/caph/mppi086h/software/anaconda3/envs/gammapy-0.17/lib/python3.7/site-packages/gammapy/modeling/fit.py in covariance(self, backend, **kwargs)
    216 
    217             covariance = Covariance.from_factor_matrix(
--> 218                 parameters=self._models.parameters, matrix=factor_matrix
    219             )
    220             self._models.covariance = covariance

/home/woody/caph/mppi086h/software/anaconda3/envs/gammapy-0.17/lib/python3.7/site-packages/gammapy/modeling/covariance.py in from_factor_matrix(cls, parameters, matrix)
     70 
     71         if not matrix.shape == (npars, npars):
---> 72             matrix = cls._expand_factor_matrix(matrix, parameters)
     73 
     74         scales = [par.scale for par in parameters]

/home/woody/caph/mppi086h/software/anaconda3/envs/gammapy-0.17/lib/python3.7/site-packages/gammapy/modeling/covariance.py in _expand_factor_matrix(matrix, parameters)
     58         mask = np.array([par.frozen for par in parameters])
     59         free_parameters = ~(mask | mask[:, np.newaxis])
---> 60         matrix_expanded[free_parameters] = matrix.ravel()
     61         return matrix_expanded
     62 

ValueError: NumPy boolean array indexing assignment cannot assign 49 input values to the 144 output values where the mask is true
```

In my case there are 5 free model parameters and two free background_model parameters (one norm for each dataset), so 7 free parameters are fitted and the covariance matrix is of size 7x7. But the fit counts the 5 free model parameters twice so it expects a 12x12 matrix.

**To Reproduce**
create two datasets
create two models
[...]

PR #2861, opened 2020-04-09
Fix covariance matrix expansion for duplicate parameters

QRemy:
Fix covariance matrix expansion for duplicate parameters
Resolve issue #2859

### DIFF
--- a/gammapy/datasets/tests/test_io.py
+++ b/gammapy/datasets/tests/test_io.py
@@ -3,7 +3,7 @@ from numpy.testing import assert_allclose
 from gammapy.datasets import Datasets
 from gammapy.modeling.models import Models
 from gammapy.utils.testing import requires_data
-
+from gammapy.modeling import Fit
 
 @requires_data()
 def test_datasets_to_io(tmp_path):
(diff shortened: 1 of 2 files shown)


QRemy:
The datasets IO test fails because it do not read the new version of the models.yaml file updated in gammapy-extra.

gammapy/datasets/tests/test_io.py
```diff
@@ -59,3 +64,9 @@ def test_datasets_to_io(tmp_path):
     assert dataset0.psf is not None
     assert dataset0.edisp is not None
     assert_allclose(dataset0.background_model.evaluate().data.sum(), 4094.2, atol=0.1)
+
+    Fit(datasets).run()
```
registerrier, line 68: I think this now requires a `@requires_minuit`.

registerrier:
I have updated the gammapy-data repo with the correct yaml file. Tests are now running.

gammapy/datasets/tests/test_io.py
QRemy: added @requires_dependency("iminuit")

PR #2866, opened 2020-04-18
TemplateSpatialModel in sr-1

JouvinLea:
This PR fix a small bug for the TemplateSpatialModel. In the evaluate model, it is returning a Quantity in Map.unit instead of sr-1. This create unit error for example in the compute_npred() of the cube if you didn't define the unit of the model as sr-1.

It is now forced to give the unit as sr-1.

- harmonise the return of the API for the `ShellSpatialModel` and `ConstantSpatialModel` for the return of evaluate

### DIFF
--- a/gammapy/modeling/models/spatial.py
+++ b/gammapy/modeling/models/spatial.py
@@ -466,7 +466,7 @@ class ShellSpatialModel(SpatialModel):
             value[mask] = (value - np.sqrt(radius ** 2 - sep ** 2))[mask]
             value[sep > radius_out] = 0
 
-        return u.Quantity(norm * value, "sr-1", copy=False)
+        return norm * value
 
     def to_region(self, **kwargs):
         """Model outline (`~regions.CircleAnnulusSkyRegion`)."""
@@ -507,7 +507,7 @@ class ConstantSpatialModel(SpatialModel):
     @staticmethod
     def evaluate(lon, lat, value):
         """Evaluate model."""
-        return u.Quantity(value, "sr-1", copy=False)
+        return value
 
     @staticmethod
     def to_region(**kwargs):
@@ -524,7 +524,7 @@ class ConstantSpatialModel(SpatialModel):
 class TemplateSpatialModel(SpatialModel):
     """Spatial sky map template model (2D).
 
-    This is for a 2D image. The unit of the map has to be in ``sr-1``.
+    This is for a 2D image. The unit of the map has to be equivalent to ``sr-1``.
 
     Use `~gammapy.modeling.models.SkyDiffuseCube` for 3D cubes with
     an energy axis.
@@ -534,7 +534,7 @@ class TemplateSpatialModel(SpatialModel):
     Parameters
     ----------
     map : `~gammapy.maps.Map`
-        Map template. The unit has to be in ``sr-1``.
+        Map template. The unit has to be equivalent to ``sr-1``.
     norm : float
         Norm parameter (multiplied with map values)
     meta : dict, optional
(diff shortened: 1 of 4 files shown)


gammapy/modeling/models/spatial.py
```diff
@@ -507,7 +507,7 @@ def to_dict(self):
     @staticmethod
     def evaluate(lon, lat, value):
         """Evaluate model."""
-        return value
+        return u.Quantity(value, "sr-1", copy=False)
```
adonath, line 510: Same here, I think there is no gain in forcing the output unit here. The output unit depends on the parameter unit of the model, which is just fine. I users want to have `sr-1` they can always do `model(lon, lat).to("sr-1")`. Instead it comes with a minimal performance (not really relevant) drawback, because we transform the units and make a copy if it is different.

gammapy/modeling/models/spatial.py
```diff
@@ -563,6 +565,8 @@ def __init__(
             filename = str(make_path(filename))
 
         self.map = map
+        if self.map.unit != "sr-1":
```
adonath, line 568: I would just keep this check here. It's much better to introduce a convention and let the code fail early (on init of the `TemplateSpatialModel` ), I would modify the code to also allow equivalent units such as `deg2`. You can use e.g. `unit.is_equivalent()`. Also adapt the error message...

gammapy/modeling/models/spatial.py
```diff
@@ -604,13 +608,15 @@ def read(cls, filename, normalize=True, **kwargs):
...
+        elif m.unit != "sr-1":
+            raise ValueError("The map unit should be in sr-1")
         return cls(m, normalize=normalize, filename=filename)
 
     def evaluate(self, lon, lat, norm):
         """Evaluate model."""
         coord = {"lon": lon.to_value("deg"), "lat": lat.to_value("deg")}
         val = self.map.interp_by_coord(coord, **self._interp_kwargs)
-        return u.Quantity(norm.value * val, self.map.unit, copy=False)
+        return u.Quantity(norm.value * val, "sr-1", copy=False)
```
adonath, line 619: This some is basically redundant. Once you have checked the map unit on init you can just use it here and it's guaranteed to be compatible...

gammapy/modeling/models/spatial.py
```diff
@@ -602,8 +606,10 @@ def read(cls, filename, normalize=True, **kwargs):
             Keyword arguments passed to `Map.read()`.
         """
         m = Map.read(filename, **kwargs)
+
```
JouvinLea, line 609: @adonath
sorry I add the black lines to have the diff here. I don't know about that in the read and from_dict() method. I don't like too much to force a default unit if there is none? Should we remove those in read and from_dict() ?

gammapy/modeling/models/spatial.py
```diff
@@ -466,7 +466,7 @@ def evaluate(lon, lat, lon_0, lat_0, radius, width):
             value[mask] = (value - np.sqrt(radius ** 2 - sep ** 2))[mask]
             value[sep > radius_out] = 0
 
-        return norm * value
+        return u.Quantity(norm * value, "sr-1", copy=False)
```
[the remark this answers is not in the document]
JouvinLea: @adonath I agree but it is done for all the others spatial models to force sr-1 unit so I was trying to have it harmonize !

gammapy/modeling/models/spatial.py
registerrier: How about checking that the unit is equivalent to `sr-1` and if not add a `log.warning(Spatial template unit is not equivalent to sr^-1. Changing unit to sr^-1)`?

gammapy/modeling/models/tests/test_io.py
```diff
@@ -157,7 +157,9 @@ def test_absorption_io(tmp_path):
 def make_all_models():
     """Make an instance of each model, for testing."""
     yield Model.create("ConstantSpatialModel")
-    yield Model.create("TemplateSpatialModel", map=Map.create(npix=(10, 20)))
+    map = Map.create(npix=(10, 20))
```
registerrier, line 160: avoid using `map` as a variable name.

registerrier: And use `Map.create(npix=(10, 20), unit='sr-1')`

PR #2867, opened 2020-04-20
Fix bug in MapDatasetEventSampler

fabiopintore:
This PR fixs a minor bug of the `~gammapy.dataset.MapDatasetEventSampler`. Currently, when the sampled source is very weak and no source events are generated, the code breaks. This is due to a bug in the creation of the `~astropy.Table` of the events. 
The PR corrects the issue by introducing a simple if condition based on the number of simulated source events.

### DIFF
--- a/gammapy/datasets/simulate.py
+++ b/gammapy/datasets/simulate.py
@@ -86,7 +86,11 @@ class MapDatasetEventSampler:
                 temporal_model = ConstantTemporalModel()
 
             table = self._sample_coord_time(npred, temporal_model, dataset.gti)
-            table["MC_ID"] = idx + 1
+            if len(table) > 0:
+                table["MC_ID"] = idx + 1
+            else:
+                mcid = table.Column(name="MC_ID", length=len(table), dtype=int)
+                table.add_column(mcid)
             events_all.append(EventList(table))
 
         return EventList.stack(events_all)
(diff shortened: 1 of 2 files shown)


gammapy/datasets/simulate.py
```diff
@@ -86,7 +86,11 @@ def sample_sources(self, dataset):
                 temporal_model = ConstantTemporalModel()
 
             table = self._sample_coord_time(npred, temporal_model, dataset.gti)
-            table["MC_ID"] = idx + 1
+            if len(table) > 0:
+                table["MC_ID"] = idx + 1
+            else:
+                mcid = table.Column(name="MC_ID", length=len(table), dtype=int)
```
registerrier, line 92: Maybe put explicitly `len(table)` to `0` here?

gammapy/datasets/simulate.py
```diff
@@ -316,13 +320,13 @@ def run(self, dataset, observation=None):
         if len(dataset.models) > 1:
             events_src = self.sample_sources(dataset)
 
-            if dataset.psf:
+            if dataset.psf and (len(events_src.table) > 0):
```
registerrier, line 323: Is the condition `(len(events_src.table) > 0)` useful here or should it apply to both:
```
if len(events_src.table) > 0:
    if dataset.edisp:
...
    else:
...
    if dataset.psf:
...
```

gammapy/datasets/simulate.py
fabiopintore: You're right. I changed to `0`.

I also added a new unit test called `test_mde_sample_weak_src`. This verifies that, when the source is too faint, the simulated events are related only to the background.

gammapy/datasets/simulate.py
fabiopintore: Yes, it can be applied to both. I modified it.

Issue #2865, opened 2020-04-17
Adapt MapDataset.to_image() to produce IRFs in reco energy


registerrier:
`MapDataset.to_image()` is used to reduce 3D datasets to 2D ones. The resulting `MapDataset` can then be used for morphological studies.

One difficulty with this approach is that the resulting IRFs are summed over true energies and might not properly reflect the masked energy bins and will not cover the same energy range if the true energy range is different from the reco energy range (which is usually preferable). 

Two behaviors are possible:
- true energy IRFs are left untouched, `to_image` is applied only to counts and background model. `edisp` must be adapted to have only one reco energy bin.
- true energy exposure and PSF are reprojected on reco energy (with the edisp) and then summed over the correct energy range.  e.g:
`expo(x,y,E_i) = \int dE_0 dOmega Aeff(E_0,x_0,y_0) Phi(E_0) PSF(x,y|x_0,y_0,E_0) Edisp(i|E_0)`
`PSF(delta_x,delta_y, E_i) =  \int dE_0 Aeff(E_0,x_0,y_0) Phi(E_0) psf(x,y|x_0,y_0) Edisp(i|E_0)/expo(x,y,E_i)`
`\phi(E_0)` being the spectral model used for averaging over energies.

I suggest we offer the second option. The first one is not very useful in itself. It is probably as simple to perform 3D fits, possibly limiting to a precise energy range.

This could be implemented first by adding methods to compute the average exposure and PSF in the reco geometry. Then `MapDataset.to_image()` can call them.

Opinions @AtreyeeS @adonath ?

AtreyeeS:
Thanks @registerrier for bringing this issue up. Indeed the `to_image` is totally useless in its present form.

I am a bit unsure if the first option is equivalent to performing 3D fits in a given range, say for energy dependent morphologies where I do not want to take such complications into account but fit a GaussianSpatialModel above a threshold.

I am also confused about how to implement the second option. If the method to compute the `reco_exposure` resides on the `MapDataset`, it does not have access to the `aeff`.

As a technicality, the exposure axis must be named `energy_true` for model evaluation. It be confusing to have an exposure in reco energy and can it etrue...

A side note, currently `to_image` discards the models on the imaged dataset. I would prefer that they remain.

adonath:
Thanks @registerrier for bringing up this topic again. 

[...]

PR #2870, opened 2020-04-26
Modify to_image() to keep true energy axis

AtreyeeS:
**Description**

This pull request modifies `MapDataset.to_image()` to keep the true energy axis as discussed in #2865 . 

**Dear reviewer**

I have kept the models on the imaged dataset. This makes it easier to make the residual images.
Is this ok?


gammapy/datasets/map.py
```diff
@@ -909,27 +907,64 @@ def to_image(self, spectrum=None, name=None):
...
             background = self.background_model.evaluate() * mask_safe
             background = background.sum_over_axes(keepdims=True)
-            kwargs["models"] = Models(
-                [BackgroundModel(background, datasets_names=[name])]
-            )
+        else:
+            background = None
+
+        if isinstance(self.edisp, EDispMap):
+            kwargs["edisp"] = self.edisp
```
registerrier, line 917: `edisp` is unchanged?

gammapy/datasets/map.py
```diff
@@ -909,27 +907,64 @@ def to_image(self, spectrum=None, name=None):
...
+        else:
+            kwargs["edisp"] = None
+            if self.exposure is not None:
+                exposure = _map_spectrum_weight(self.exposure, spectrum)
+                kwargs["exposure"] = exposure.sum_over_axes(keepdims=True)
 
-        if self.psf is not None:
             # TODO: implement PSFKernel.to_image()
-            if not isinstance(self.psf, PSFKernel):
+            if isinstance(self.psf, PSFMap):
```
registerrier, line 927: Normally this should not be applied right?

gammapy/datasets/map.py
```diff
@@ -909,27 +907,64 @@ def to_image(self, spectrum=None, name=None):
...
             else:
-                # assume exposure at center position
                 kwargs["psf"] = None
 
+        kwargs["models"] = self._copy_models(background=background, name=name)
+
         return self.__class__(**kwargs)
 
+    def _copy_models(self, background=None, name=None):
+        """
```
registerrier, line 937: Here I suppose the problem is more consistency between different `Dataset` transforming methods. 
I probably wouldn't export more than the transformed background model as is the case now. 

We need to define the correct strategy in the long run.

AtreyeeS:
Thanks @registerrier for your comments. I have removed the copy on the models for the present, but this is something I would like to put back in near future.

Perhaps related, I now keep the parameters of the norm and tilt of the background model of the 3D dataset on imaged dataset. Perhaps this is related to model handling as well, but without this, I do not see much use to the `to_image`.

gammapy/datasets/map.py
```diff
@@ -869,14 +868,13 @@ def to_spectrum_dataset(self, on_region, containment_correction=False, name=None
         return SpectrumDataset(**kwargs)
 
     def to_image(self, spectrum=None, name=None):
-        """Create images by summing over the energy axis.
+        """Create images by summing over the reco energy axis.
+        True energy axis is left unchanged to have proper model evaluation.
 
-        Exposure is weighted with an assumed spectrum,
+        In case an EdispMap is not present, true energy axis is also summed;
```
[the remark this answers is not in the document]
AtreyeeS: It seems a default `edisp` is not necessary. I am not totally sure why, but it seems to be created by default at some point.

AtreyeeS:
Closing this for now, opening a new one.

PR #2872, opened 2020-04-28
Mcmc init

facero:
**Description**
In testing the MCMC for different sources, the evaluation of the likelihood sometime returned an error with a message like `cannot evaluate PSF at given location`. Turns out that the walker was initialized out of the ROI.
So far the walker are initialized with a `spread = p0 * 0.5%` (copied from naima).
This is giving a small spread for Norm and index parameters (too small?) but for spatial coordinates (lon, lat) this doesn't make sense.
For lon = 0° the spread is 0° and for lon = 300° the sigma is 1.5° which can initiate some walkers several degrees from the ROI producing the error described above.

This pull request initiate the spread of the spectral parameters and spatial parameters differently. The spread for lon, lat is no longer a fraction but fixed to 0.1°.

**Dear reviewer**
This is a pretty simple change.
Are the names `lon_0`, `lat_0` the same across all spatial models ? (I think so but double checking)

### DIFF
--- a/gammapy/modeling/sampling.py
+++ b/gammapy/modeling/sampling.py
@@ -82,15 +82,26 @@ def run_mcmc(dataset, nwalkers=8, nrun=1000, threads=1):
     import emcee
 
     dataset.models.parameters.autoscale()  # Autoscale parameters
-    pars = [par.factor for par in dataset.models.parameters.free_parameters]
-    ndim = len(pars)
+    #pars = [par.factor for par in dataset.models.parameters.free_parameters]
+    #ndim = len(pars)
 
     # Initialize walkers in a ball of relative size 0.5% in all dimensions if the
     # parameters have been fit, or to 10% otherwise
+    # Handle source position spread differently with a spread of 0.1°
     # TODO: the spread of 0.5% below is valid if a pre-fit of the model has been obtained.
     # currently the run_mcmc() doesn't know the status of previous fit.
-    spread = 0.5 / 100
-    p0var = np.array([spread * pp for pp in pars])
+    p0var = []
+    pars = []
+    spread = 0.5/100
+    spread_pos = 0.1 # in degrees
+    for par in dataset.models.parameters.free_parameters:
+        pars.append(par.factor)
+        if par.name in ["lon_0", "lat_0"]:
+            p0var.append(spread_pos/par.scale)
+        else:
+            p0var.append(spread*par.factor)
+
+    ndim = len(pars)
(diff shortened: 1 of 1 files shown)


facero:
Another question is the fraction to initiate the other parameters. 0.5% is a small spread and it is what is hard coded in naima after a valid fit is obtained. For a spectral index of 2.0 this gives a spread of 0.01. Same for a norm of 2 flux units. This seems pretty small. 5% would seem more reasonable.
On the other hand I'm not sure it will change anything in the end because the walkers will evolve to probe the likelihood profile. So after a few 100 runs, the walker distribution could be independent of the initial conditions.
This is something that I could test.

gammapy/modeling/sampling.py
```diff
@@ -82,15 +82,24 @@ def run_mcmc(dataset, nwalkers=8, nrun=1000, threads=1):
...
     # parameters have been fit, or to 10% otherwise
+    # Handle source position spread differently with a spread of 0.1°
     # TODO: the spread of 0.5% below is valid if a pre-fit of the model has been obtained.
     # currently the run_mcmc() doesn't know the status of previous fit.
-    spread = 0.5 / 100
-    p0var = np.array([spread * pp for pp in pars])
+    p0var = []
+    pars = []
+    spread = 0.5/100
+    spread_pos = 0.1 # in degrees
```
registerrier, line 94: Is the 0.1° valid in all situations? 
Is there a possibility to guess a plausible range from the input `dataset`?

facero: This would be better indeed but not sure how.
The uncertainty of the source position (hence initial spread) depends on pt src/ext src, its brightness, etc
Even if 0.1° is not optimal, the walkers distribution "adapts" to the shape of the local likelihood after some run.
What is better to avoid is to have a spread too large where the init position is outside of the box or possibly taking the position of another source.

registerrier:
OK. Thanks @facero. We can merge and we will see how to improve on this later.

facero:
Before the next release, I would like to implement some unit tests for the MCMC but I don't know exactly which parts to test. The notebook test it but they are not part of the automated tests. Are they ?
It could be a mini run of `run_mcmc()` but this might take ~20 sec a bit long for a unit test?

registerrier:
Indeed, having some unit tests for `sampling.py` would be safer. In theory, every line of code should be executed by at least one test. You would also test on the output of each function.
You might run `run_mcmc` but it could be for a basic an extremely simplified case with very few walkers. This does not need to be physically meaningful, unit tests are just there to ensure code pieces are running as expected.

PR #2883, opened 2020-05-04
Improve path handling in serialization

### LINKED ISSUE #2875: Datasets.read() fails unless called in the same folder where Datasets.write() was called
**Gammapy version**
v.17

**Bug description**
Hello! I have stored a Datasets  object after fitting, and later I tried to reload it using Datasets.read(filedata=..., filemodel=...). I encounter the following issue: I can reload it only from the same folder in which I had called Datasets.write(...). If I try to load it from another folder, I get the message:

FileNotFoundError: [Errno 2] No such file or directory: 'datasets_best_fit_0.02/10_models_covariance.dat'

I think the issue comes from here :  https://github.com/gammapy/gammapy/blob/f373ad6e7439c515e7f2f85938adb2afc73a7bc8/gammapy/modeling/models/core.py#L286 . We should probably substitute the relative path with the absolute path in that line.

The issue can be quicly solved, from a user perspective, by passing an absolute path to `Datasets.write(path=...)`. But maybe we could convert internally to an absolute path also in case a relative one is passed.

Noticed that also the data files, and not only the covariance, are affected by this: https://github.com/gammapy/gammapy/blob/f373ad6e7439c515e7f2f85938adb2afc73a7bc8/gammapy/datasets/core.py#L228

QRemy:
- change make_path() to return Path(os.path.expandvars(path)).resolve()
- add make_path(filename) in covariance-io
- datasets-io already use make_path(filename) so it should also work for data files.

Resolve #2875


QRemy:
If there is any path in the yaml file then all files have to be in the same folder. That's not very convenient for the template files that ones have probably already store somewhere else (for example diffuse models and extended sources templates).

Alternatively the issue could be solved by using relative path for covariance and data files.

gammapy/datasets/core.py
```diff
@@ -200,12 +202,23 @@ def read(cls, filedata, filemodel):
...
+        else:
+            filemodel = make_path(filemodel)
+
+        models = Models.read(filemodel)
+        data_list = read_yaml(filedata)
 
         datasets = []
         for data in data_list["datasets"]:
-            dataset = DATASETS.get_cls(data["type"]).from_dict(data, models)
+            dataset = DATASETS.get_cls(data["type"]).from_dict(data, models, path=path)
```
registerrier, line 221: To avoid code duplication in the various `from_dict` methods, isn't it simpler to format the full path inside the loop here. We can safely assume that `Datasets.read` should be the main way to read datasets from disk no?

gammapy/datasets/map.py
```diff
@@ -766,9 +766,16 @@ def read(cls, filename, name=None):
             return cls.from_hdulist(hdulist, name=name)
 
     @classmethod
-    def from_dict(cls, data, models):
+    def from_dict(cls, data, models, path=""):
```
registerrier, line 769: See above

gammapy/datasets/flux_points.py
```diff
@@ -107,7 +107,7 @@ def write(self, filename, overwrite=True, **kwargs):
         table.write(filename, overwrite=overwrite, **kwargs)
 
     @classmethod
-    def from_dict(cls, data, models):
+    def from_dict(cls, data, models, path=""):
```
registerrier, line 110: see above

gammapy/datasets/spectrum.py
```diff
@@ -1225,7 +1225,7 @@ def write(self, filename, overwrite):
         self.to_ogip_files(outdir=outdir, overwrite=overwrite)
 
     @classmethod
-    def from_dict(cls, data, models):
+    def from_dict(cls, data, models, path=""):
```
registerrier, line 1228: See above

gammapy/utils/scripts.py
```diff
@@ -63,7 +63,7 @@ def make_name(name=None):
         return name
 
 
-def make_path(path):
+def make_path(path, abs_path=False):
```
registerrier, line 66: Maybe the change is not required if one simply add `.resolve()` to the path returned by `make_path` when it is needed.

gammapy/datasets/core.py
```diff
@@ -200,12 +202,23 @@ def read(cls, filedata, filemodel):
         """
         from . import DATASETS
 
-        models = Models.read(make_path(filemodel))
-        data_list = read_yaml(make_path(filedata))
+        path = make_path(path)
+
+        if (path / filedata).exists():
```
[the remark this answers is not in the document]
QRemy: If filedata is an absolute path it would fail and then filedata is used directly.

PR #2888, opened 2020-05-06
Adding unit test for the mcmc sampling

facero:
**Description**

This pull request adds unit tests for the sampling MCMC.
With few runs for the MCMC, the full tests take only 3sec

Minor question:
I tried to use pytest fixture to avoid having to load the datasets in each test but this didn't work as I expected. Can I use fixture here to load the dataset for all functions here ?

### DIFF
--- a/gammapy/modeling/tests/test_sampling.py
+++ b/gammapy/modeling/tests/test_sampling.py
@@ -1,9 +1,9 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
-import pytest
 from gammapy.datasets import Datasets
 from gammapy.modeling.sampling import run_mcmc, ln_uniform_prior
 import numpy as np
-#@pytest.fixture
+from gammapy.utils.testing import requires_dependency
+
 def get_datasets():
     dataset = Datasets.read("$GAMMAPY_DATA/fermi-3fhl-crab/Fermi-LAT-3FHL_datasets.yaml",
                             "$GAMMAPY_DATA/fermi-3fhl-crab/Fermi-LAT-3FHL_models.yaml")
@@ -35,9 +35,10 @@ def test_lnprob():
     parameters["amplitude"].value = 1000
     assert ln_uniform_prior(dataset) == -np.inf
 
+@requires_dependency("emcee")
 def test_runmcmc():
     #Running a small MCMC on pregenerated datasets
     dataset = get_datasets()
-    sampler = run_mcmc(dataset, nwalkers=6, nrun=10)  # to speedup the test
+    run_mcmc(dataset, nwalkers=6, nrun=10)  # to speedup the test
 
 
--- a/gammapy/modeling/tests/test_sampling.py
+++ b/gammapy/modeling/tests/test_sampling.py
@@ -45,6 +45,5 @@ def test_runmcmc(dataset):
     import emcee
 
     sampler = run_mcmc(dataset, nwalkers=6, nrun=10)  # to speedup the test
-    assert type(sampler) is emcee.ensemble.EnsembleSampler
-
+    assert isinstance(sampler, emcee.ensemble.EnsembleSampler)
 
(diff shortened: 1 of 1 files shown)


facero:
The tests passed on my machine but apparently not in here due to `ModuleNotFoundError: No module named 'emcee'`.
Is it possible to add it to the pipeline configuration ?

registerrier:
I think you need to specify that the dependency is required:
`from gammapy.utils.testing import requires_dependency`

and before the function:
`@requires_dependency("emcee")`

gammapy/modeling/tests/test_sampling.py
```diff
@@ -0,0 +1,44 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+from gammapy.datasets import Datasets
+from gammapy.modeling.sampling import run_mcmc, ln_uniform_prior
+import numpy as np
+from gammapy.utils.testing import requires_dependency
```
registerrier, line 5: Here you also need to import `requires_data`

gammapy/modeling/tests/test_sampling.py
```diff
@@ -0,0 +1,44 @@
...
+    parameters["lat_0"].min = -90
+    parameters["lat_0"].max = 90
+    parameters["lon_0"].min = 0
+    parameters["lon_0"].max = 360
+    parameters["amplitude"].min = 0.01 * parameters["amplitude"].value
+    parameters["amplitude"].max = 100 * parameters["amplitude"].value
+
+    return dataset
+
+def test_lnprob():
```
registerrier, line 27: Here add the decorator: `@requires_data()`

gammapy/modeling/tests/test_sampling.py
```diff
@@ -0,0 +1,44 @@
...
+    parameters = dataset.models.parameters
+
+    #paramater is within min, max boundaries
+    assert ln_uniform_prior(dataset) == 0.0
+    # Setting amplitude outside min, max values
+    parameters["amplitude"].value = 1000
+    assert ln_uniform_prior(dataset) == -np.inf
+
+@requires_dependency("emcee")
+def test_runmcmc():
```
registerrier, line 39: Here you also need the decorator: `@requires_data()`

gammapy/modeling/tests/test_sampling.py
```diff
@@ -0,0 +1,44 @@
...
+    #Running a small MCMC on pregenerated datasets
+    dataset = get_datasets()
+    run_mcmc(dataset, nwalkers=6, nrun=10)  # to speedup the test
+
```
registerrier, line 43: There is nothing to assert here?

gammapy/modeling/tests/test_sampling.py
```diff
@@ -0,0 +1,44 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+from gammapy.datasets import Datasets
+from gammapy.modeling.sampling import run_mcmc, ln_uniform_prior
+import numpy as np
```
registerrier, line 4: Import non gammapy dependencies first

gammapy/modeling/tests/test_sampling.py
```diff
@@ -0,0 +1,44 @@
...
+from gammapy.utils.testing import requires_dependency
+
+def get_datasets():
```
registerrier, line 7: You might want to use a purest fixture here.
First you need to `import pytest`

then rewrite the definition of the function:
```
def dataset():
```

then add as a argument for the test functions : `dataset`
i.e. `def test_lnprob(dataset)`
and 
`def test_runmcmc(dataset) `

gammapy/modeling/tests/test_sampling.py
facero: I added assert that the sampler returned is an emcee.ensemble.EnsembleSampler  object.
As the walker are random hard to test any specific value.

gammapy/modeling/tests/test_sampling.py
facero: thanks, now I was able to use pytest fixtures !

PR #2889, opened 2020-05-08
MNT: Ignore import error in pytest collection

pllim:
This is an automated update made by the `batchpr` tool :robot: to future-proof test collection behavior `pytest-doctestplus` - feel free to close if it doesn't look good or isn't relevant! You can report issues to @pllim.


adonath:
I think so far this is not relevant to us. We have to activate doc testing first see #101.

PR #2890, opened 2020-05-09
Improve Models management on Dataset and Datasets

### LINKED ISSUE #2880: Issue with model setting for joint dataset
**Gammapy version**
0.17

**Bug description**
As explained in the [3D analysis tutorial](https://docs.gammapy.org/0.17/notebooks/analysis_3d.html) (see cell [23]), I assign a model to each `MapDataset` inside a joint `Datasets` using:
```
for dataset in datasets:
    dataset.models.append(model_joint)
```
This works fine However, if I store the `Datasets` and later reload to add another model component, it fails with 
```
ValueError: Model names must be unique
```
Here is a working example: https://github.com/luca-giunti/share/blob/master/Bug_report.ipynb

QRemy:
This PR introduces a new class of Models to be used as the models property of Datasets and Dataset.
- It returns only the models that effectively contributes to the datseset, i.e. if dataset.name in model.datasets_names or model.datasets_names is None (this test was previously done on npred computation).
- It allows to have Models management operations as append/remove... that work at both Dataset and Datasets levels.
- Using append on a single dataset should be more intuitive now as
dataset.models.append(model) also append  dataset.name to model.datasets_names
(only if datasets_names is not None).
- Force consistency between models listed in each dataset of a datasets and each model.datasets_names. For example if a model with datasets_names = None appears in only one dataset, it will be automatically added to the others.

Resolve #2880


registerrier:
Merging now.

PR #2896, opened 2020-05-12
Add pyproject.toml, fixes 2895

### LINKED ISSUE #2895: Install fails when Cython is not already installed
**Gammapy version**

Every I tested

**Bug description**

The setup.py hard imports Cython, so setuptools has no chance to install it.

**Expected behavior**

Install cython if needed before attempting to import it

**To Reproduce**

```
$ python -m venv test_gammapy
$ . test_gammapy/bin/activate
$ pip install gammapy
```

**Other information**
There is a very simple fix: provide a pyproject.toml that requires cython for the build system.

MaxNoe:
**Description**
Add `pyproject.toml` to specify build system requiredments (setuptools, wheel, Cython).

Fixes #2895 

For more info on the pyproject.toml and why every python project should include one, read:

* https://www.python.org/dev/peps/pep-0517/
* https://www.python.org/dev/peps/pep-0518/
* https://snarky.ca/what-the-heck-is-pyproject-toml/

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,3 @@
+[build-system]
+requires = ["setuptools >= 40.6.0", "wheel", "Cython"]
+build-backend = "setuptools.build_meta"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,9 @@
 [build-system]
-requires = ["setuptools >= 40.6.0", "wheel", "Cython"]
+requires = [
+	"setuptools >= 40.6.0",
+	"wheel",
+	"Cython",
+	'numpy >= 1.16',
+	'setuptools_scm[toml]'
+]
 build-backend = "setuptools.build_meta"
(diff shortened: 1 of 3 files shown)


MaxNoe:
The errors look unrelated to my changes (reduction over zero size array in one of the notebooks)

adonath:
Thanks a lot @MaxNoe, this looks good to me!

I quickly checked and it seems most projects in the astropy ecosystem depending on cython use exactly this approach (see e.g. [astropy](https://github.com/astropy/astropy/blob/master/pyproject.toml) & [reproject](https://github.com/astropy/reproject/blob/master/pyproject.toml)). I'll check this setup later in a clean environment locally and once I can confirm it works I'll merge. We maybe want to define a fixed cython version later, but this decision I'll postpone to the cleanup phase before the next release...

adonath:
The local install worked perfectly. I'll go ahead and merge now. Thanks again @MaxNoe!

PR #2905, opened 2020-05-23
Fix evaluators cache

### LINKED ISSUE #2904: Adding and removing models
I found some strange behaviour while adding and removing models in the present dev version of gammapy. I try to reproduce a very basic example here.

https://github.com/AtreyeeS/test_notebooks/blob/master/simulate_3d.ipynb

Making the datasets is just using the part from simulate_3d notebook.
In the next section, I (i) add a model, (ii) remove it and (ii) add it back (with some different initial parameters). The `npred` calculation after (iii) goes back to what it was after (i), which is unexpected.

This, of course, leads to the minimiser royally failing later on...

@QRemy @registerrier @luca-giunti - have you faced this before? Any idea what is causing this?

QRemy:
- Evaluators dict key are now model.hex_id instead of .name as suggested in #2904 
- Automatically clear the evaluators of the removed models
- Move evaluators set up from .npred() to .evaluators() property
 Resolve #2904

### DIFF
--- a/gammapy/datasets/map.py
+++ b/gammapy/datasets/map.py
@@ -246,13 +246,14 @@ class MapDataset(Dataset):
 
         if self.models:
             for model in self.models:
-                evaluator = self._evaluators.get(model.name)
+                model_id = hex(id(model))
+                evaluator = self._evaluators.get(model_id)
 
                 if evaluator is None:
                     evaluator = MapEvaluator(
                         model=model, evaluation_mode=self.evaluation_mode, gti=self.gti
                     )
-                    self._evaluators[model.name] = evaluator
+                    self._evaluators[model_id] = evaluator
 
                 # if the model component drifts out of its support the evaluator has
                 # has to be updated
--- a/gammapy/datasets/map.py
+++ b/gammapy/datasets/map.py
@@ -212,9 +212,23 @@ class MapDataset(Dataset):
     @property
     def evaluators(self):
         """Model evaluators"""
-        # this call is needed to trigger the setup of the evaluators
-        if not self._evaluators:
-            self.npred()
+
+        if self.models:
+            for model in self.models:
+                model_id = hex(id(model))
+                evaluator = self._evaluators.get(model_id)
+
+                if evaluator is None:
+                    evaluator = MapEvaluator(
+                        model=model, evaluation_mode=self.evaluation_mode, gti=self.gti
+                    )
+                    self._evaluators[model_id] = evaluator
+
+                # if the model component drifts out of its support the evaluator has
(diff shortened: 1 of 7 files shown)


gammapy/datasets/map.py
```diff
@@ -244,25 +262,10 @@ def npred(self):
...
-                # if the model component drifts out of its support the evaluator has
-                # has to be updated
-
-                if evaluator.needs_update:
-                    evaluator.update(self.exposure, self.psf, self.edisp, self._geom)
-
-                if evaluator.contributes:
-                    npred = evaluator.compute_npred()
-                    npred_total.stack(npred)
+        for _, evaluator in self.evaluators.items():
```
adonath, line 265: As the `_` is unused, just simplify to `for evaluator in self.evaluators`.

gammapy/datasets/map.py
```diff
@@ -212,9 +212,27 @@ def models(self, models):
...
+        if self.models:
+            hex_ids = []
+            for model in self.models:
+                evaluator = self._evaluators.get(model.hex_id)
+                hex_ids.append(model.hex_id)
+                if evaluator is None:
+                    evaluator = MapEvaluator(
+                        model=model, evaluation_mode=self.evaluation_mode, gti=self.gti
+                    )
+                    self._evaluators[model.hex_id] = evaluator
```
adonath, line 225: We can actually use the model directly as a key here, like `self._evaluators[model]`, this results in the same behaviour but will be simpler. It also avoids to introduce the `model.hex_id` property. Maybe it's also an option then use set differences likes:

```python3
evaluators = set(["a", "b", "c"])
models = set(["b", "d", "f"])

# evaluators to delete    
for key in evaluators.difference(models):
    print(key)
```

QRemy:
Thanks @adonath 
About the parallelization last time I tried to distribute fitting of pre-made datasets with multuprocessing.pool  it was no longer possible because of the introduction of @lru_cache that prevent to pickle the object.

Here I don't see a problem with parallelization of because even if the models have different ids   we are still be able to compute their npred in parallel correctly as the evaluators are generated and removed dynamically  from self.models that contains their new ids.

So you can do data reduction multi-core -> add models one-core -> compute npred multi-core -> compute likelihood and update model one core...

QRemy:
I created another branch with parallel evaluation and caching of comput_npred() in MapDataset:
https://github.com/QRemy/gammapy/blob/test_parallel_npred/gammapy/datasets/map.py
For the test datasets it's not improving computation time but for bigger datasets the psf convolution will take longer so it should be worth.
I'll test this with a low energy Fermi dataset.

PR #2906, opened 2020-05-23
Add read/write example for exclusion masks

AtreyeeS:
This PR adds an example for reading and writing of exclusion masks in the `exclusion_mask.ipynb` tutorial.


Bultako:
<p>test message here..</p>

PR #2907, opened 2020-05-25
Add info_dict to MapDataset

### LINKED ISSUE #2749: info_table() fails for MapDatasets
**Gammapy version**

0.16.dev

**Bug description**

When working with MapDatasets calling the `info_table` method fails with the following error.
```
AttributeError: 'MapDataset' object has no attribute 'info_dict'
```

**Expected behavior**

I would expect at least not to expose the `info_table` method for MapDatasets

AtreyeeS:
This PR fixes #2749


AtreyeeS:
> Thanks @AtreyeeS . This looks good to me. I suggest a few missing things below:
> 
> Would it be possible to create a large region enclosing the full `Map` for the case where `region=None` to limit code duplication?

Adapted! 

> I also think this needs a change in `Datasets` to have region as a possible argument to `info_table` no?
Sorry! I forgot to commit the file. Done!

> Then this can be added as an example to check the signal on the Crab in e.g. analysis_2.ipynb.
analysis_2 does a stacked analysis, so I am not sure if the info_table is helpful here. Rather, I adapted analysis_3d

PR #2909, opened 2020-05-27
fix MapDatasetOnOff.mask_safe on read

### LINKED ISSUE #2908: Reading of MapDatasetOnOff from disk yields incorrect mask_safe
**Gammapy version**
0.17

**Bug description**
When a `MapDatasetOnOff` is first written to and then read from disk, the `mask_safe` attribute is not handled correctly. Before writing to disk this attribute is a `WcsNDMap`, but after reading from disk it is a plain `numpy.ndarray`.

**Expected behavior**
The data set read from disk should have a `mask_safe` that is of type `WcsNDMap`.

**To Reproduce**
Create a `MapDatasetOnOff`, then write to and read from disk (using `write` and `read`).

QRemy:
MapDatasetOnOff.mask_safe is defined as a Map on read 
Resolve #2908

### DIFF
--- a/gammapy/datasets/map.py
+++ b/gammapy/datasets/map.py
@@ -86,8 +86,6 @@ class MapDataset(Dataset):
         if mask_safe is not None and mask_safe.data.dtype != np.dtype("bool"):
             raise ValueError("mask data must have dtype bool")
 
-
-
         self._name = make_name(name)
         self._background_model = None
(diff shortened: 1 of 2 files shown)


QRemy:
I added a test on mask_safe.data, in addition to the test on mask_safe.geom.

PR #2910, opened 2020-05-27
Add more template spectral model corrections

QRemy:
Add optional log-parabola and exponential cut-off corrections  to TemplateSpectralModel, SkyDiffuseCube and background model.
Allows more flexibility in the models re-optimisation for example it could be useful if they are extrapolated in the fitting range.

Maybe adding a similar all-in one spectral model could be useful for classification and automatise decisions on the optimal shape, no ?

I will add more test and update the docstring in additional commits if there is no objection on the basic idea.

### DIFF
--- a/gammapy/analysis/tests/test_analysis.py
+++ b/gammapy/analysis/tests/test_analysis.py
@@ -317,7 +317,7 @@ def test_analysis_3d():
     analysis.get_flux_points()
 
     assert len(analysis.datasets) == 1
-    assert len(analysis.fit_result.parameters) == 8
+    assert len(analysis.fit_result.parameters) == 11
     res = analysis.fit_result.parameters
     assert res["amplitude"].unit == "cm-2 s-1 TeV-1"
     assert len(analysis.flux_points.data.table) == 2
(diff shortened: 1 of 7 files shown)


adonath:
Can’t this be handled more generally by multiplying spectral models? And fixing the norm of one? I think this would be a more flexible solution...

QRemy:
Yes maybe we could remove the norm and tilt arguments from TemplateSpectralModel, SkyDiffuseCube and BackgroundModel and have a model_correction argument as an other spectral model eventually coumpound, but I'm not sure that it would be simpler to define. And for now do we really need corrections more complex than exponential cut-off or log-parabola that would justify a more generic solution ?

adonath:
I have thought a bit about this and I think it should get a dedicated section in a PIG about model handling in Gammapy. I think these "spectral correction" models are needed for `SkyDiffuseCube`, `BackgroundModel` and `SpectralTemplate` and I think we should develop a solution that works for all in the same way. One possibility is exactly what you suggested to add an additional argument to these models, or maybe allow to use them with `SkyModel`, by using the template as a spatial model and putting the correction as spectral model or similar...

I'll start drafting a PIG this week, would you be interested in contributing? This would be very helpful!

adonath:
The plan for this is now outlined in #2944.

PR #2912, opened 2020-06-01
Cache npred in MapEvaluator

QRemy:
- add caching of parameters values and npred on evaluators such as npred is not recomputed if all parameters values do not change  (used by default). 
This speed up fitting and flux points estimation in 3FHL validation by a factor of 2.

### DIFF
--- a/gammapy/datasets/map.py
+++ b/gammapy/datasets/map.py
@@ -100,6 +100,7 @@ class MapDataset(Dataset):
         self.mask_safe = mask_safe
         self.models = models
         self.gti = gti
+        self.use_cache = use_cache
 
         # check whether a reference geom is defined
         _ = self._geom
@@ -1823,9 +1824,9 @@ class MapEvaluator:
             Predicted counts on the map (in reco energy bins)
         """
 
-        pars = self.model.parameters.values
+        pars = list(self.model.parameters.values)
         npred = self._npred_cached
-        if pars != self._pars_cached:
+        if self._pars_cached != pars:
             self._pars_cached = pars
             if isinstance(self.model, BackgroundModel):
                 npred = self.model.evaluate()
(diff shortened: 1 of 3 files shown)


gammapy/datasets/map.py
```diff
@@ -256,15 +256,16 @@ def data_shape(self):
         """Shape of the counts or background data (tuple)"""
         return self._geom.data_shape
 
-    def npred(self):
+    def npred(self, use_cache=True):
```
adonath, line 259: Maybe move `use_cache` to `MapDataset.__init__()` so that it becomes really configurable by the user.

gammapy/datasets/map.py
```diff
@@ -1814,15 +1818,24 @@ def compute_npred(self):
         npred : `~gammapy.maps.Map`
             Predicted counts on the map (in reco energy bins)
         """
-        flux = self.compute_flux()
 
-        if self.model.apply_irf["exposure"]:
-            npred = self.apply_exposure(flux)
+        pars = [p.value for p in self.model.parameters]
```
adonath, line 1822: There is a `Parameters.values` property you can use here.

gammapy/datasets/map.py
```diff
@@ -1671,21 +1672,22 @@ def __init__(
...
         evaluation_mode="local",
     ):
+
         self.model = model
         self.exposure = exposure
         self.psf = psf
         self.edisp = edisp
         self.gti = gti
         self.contributes = True
+        self._npred = None
```
adonath, line 1682: Maybe rename to `self._npred_cached` and `self._pars_cached` to be a bit more explicit?

gammapy/datasets/map.py
```diff
@@ -1814,15 +1818,24 @@ def compute_npred(self):
         npred : `~gammapy.maps.Map`
             Predicted counts on the map (in reco energy bins)
         """
-        flux = self.compute_flux()
 
-        if self.model.apply_irf["exposure"]:
-            npred = self.apply_exposure(flux)
+        pars = [p.value for p in self.model.parameters]
+        npred = self._npred
+        if pars != self._pars:
```
adonath, line 1824: This needs a numerically stable comparison such as `np.allclose()`.

PR #2923, opened 2020-06-15
Implement ExcessProfileEstimator

bkhelifi:
**Description**

This pull request is about the new estimator allowing to make image profiles according to any direction and projections.
It takes in arguments a list of regions into which the estimation of the stats is made. An helper function is added to compute automatically a list of regions allowing to make a projection orthogonally to a line.
This new class allows now to derive the profile in counts, excess, flux and radiance, with all associated errors and TS.

**Dear reviewer**
This PR is associated with the one containing the storage class of the profile. The PR id is #2922. As consequence, the tests of this PR will fail as they require the storage class. Both PR should be temporary merged to pass the tests


bkhelifi:
I have moved the storage class from ~gammapy.visualization` to this sub-package.

gammapy/estimators/map_profile.py
```diff
@@ -0,0 +1,313 @@
...
+from gammapy.maps import MapAxis
+from gammapy.utils.table import table_from_row_data
+from gammapy.estimators import ImageProfile
+from gammapy.stats import WStatCountsStatistic, CashCountsStatistic
+from gammapy.datasets import SpectrumDatasetOnOff
+
+__all__ = ["MapProfileEstimator", "make_orthogonal_boxes"]
+
+
+class MapProfileEstimator:
```
adonath, line 16: To keep consistency within our estimator naming I would actually call this thing `ExcessProfileEstimator`, (or `ExcessMapProfileEstimator`, if the result is a `MapProfile`) because it's equivalent in (algorithmic) functionality to the `ExcessMapEstimator` we already have.

gammapy/estimators/map_profile.py
```diff
@@ -0,0 +1,313 @@
...
+    """Estimate profile from a DataSet.
+
+    Parameters
+    ----------
+    regions : list of `regions`
+        regions to use
+    axis : `~gammapy.maps.MapAxis`
```
adonath, line 23: I think this additional `axis` parameter is confusing somehow, as it duplicates the information, that is already contained in the regions. I think the only solution would be to store the information on the regions in the result table as well....but I'll think about it once more.

gammapy/estimators/map_profile.py
```diff
@@ -0,0 +1,313 @@
...
+
+        Returns
+        --------
+        imageprofile : `~gammapy.estimators.ImageProfile`
+            Return an image profile class containing the result
+        """
+        results = self.make_prof(dataset, steps)
+        table = table_from_row_data(results)
+        if isinstance(self.regions[0], RectangleSkyRegion):
+            table.meta["PROFILE_TYPE"] = "orthogonal"
```
adonath, line 181: If we introduce this kind of meta information it needs to be documented somewhere, e.g. what are the other profile types and how are they defined?

gammapy/estimators/map_profile.py
```diff
@@ -0,0 +1,313 @@
...
+            if isinstance(spds, SpectrumDatasetOnOff):
+                result["alpha"] = stats.alpha
+            result.update({
+                "counts": stats.n_on,
+                "background": stats.background,
+                "excess": stats.excess,
+            })
+
+            if "ts" in steps:
+                if isinstance(stats.delta_ts, list) or isinstance(stats.delta_ts, np.ndarray):
```
adonath, line 131: This type-casting and instance checks (all below) should not be needed...please remove.

gammapy/estimators/map_profile.py
```diff
@@ -0,0 +1,313 @@
...
+                    result["errp"] = np.asanyarray(_errp[0], dtype=np.float32)
+                else:
+                    result["errp"] = np.asanyarray(_errp, dtype=np.float32)
+ 
+            if "ul" in steps:
+                result["ul"] = stats.compute_upper_limit(self.n_sigma_ul)
+ 
+            # What exposure to take? And waht about the mask?
+            result.update(
+                {"exposure": spds.aeff.data.data.max() * spds.livetime,
```
adonath, line 157: I agree the exposure is not well defined here, I would actually remove it...

bkhelifi:
1/ We can store without problem into a variable the list of datasets, easy... What I do not understand is the need to create a function that at then end contains only this loop:
    for index, reg in enumerate(self.regions):
            spds = dataset.to_spectrum_dataset(reg)
Is it not an over-design? Or maybe I have not catched what you meant.

2/ renaming the function is not a big deal (done). But I have not understand which different possibilities the function offers. It takes in argument only SkyCoord now... Maybe, you were just confused the function name.

adonath:
@bkhelifi It's mostly a matter of taste. I generally prefer a pattern where the `Estimator` and `Maker` classer are stateless. This means the state of the classes do not dependent on the input, so there are no special variables on the `Estimator` or `Maker` classes, which are modified by the algorithm, but results are passed on to functions instead. In addition it is good practice to split the algorithm in smaller steps, which allows users to access intermediate results. 

```
class Estimator:
    def __init__(self, config):
        self.config = config

    def step_1(self, dataset):
        return result

    def step_2(self, datasets):
        return result

    def run(self, dataset):
        result_1 = self.step_1(dataset)
        result = self.step_2(result_1)
        return result
    
```

This has the advantage, that intermediate results can be re-produced by users for debugging or writing to disk in between. This is the pattern that we used for `gammapy.makers` and I think it's a nice and clean pattern. So for consistency we should probably adapt the same for `Estimator` classes.

bkhelifi:
I understand your point about design... This is then implemented

gammapy/estimators/map_profile.py
bkhelifi: Well, naming is always subjective...
This estimator does far more than making a profile of excess: counts, background, excess, flux, brightness.
The name used here includes Map as it takes a MapDataset object in input....
And the name of the file is coherent with the name of the class.

However, one can change the name of the class, but to be coherent the name of the file also (leading to the close of this PR maybe, and the opening of an other one...).

Any advice?

gammapy/estimators/map_profile.py
bkhelifi: In a table such that any astropy.regions can be well represented whatever the wcs... damned! If this solution is chosen, I need really some code.

gammapy/estimators/map_profile.py
bkhelifi: I put this metadata to anticipate the development of other type of projection (e.g. profile within wedges, profile within annulus, etc). This metadata can then be used in the used to identify which type of projection is used (using indeed the document of this class).
For the moment, I think that this metadata is not super useful as only one projection method is used. But, I propose to keep it for the future...

gammapy/estimators/map_profile.py
```diff
@@ -0,0 +1,313 @@
...
+    Example
+    --------
+    This example shows how to compute a counts profile for the Fermi galactic
+    center region::
+
+        import matplotlib.pyplot as plt
+        from astropy import units as u
+        from gammapy.estimators import MapProfileEstimator, make_orthogonal_boxes, image_profile
+        from gammapy.datasets import Datasets
+
```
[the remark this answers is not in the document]
bkhelifi: Oups! sorry for this oversight

gammapy/estimators/map_profile.py
bkhelifi: There is a problem of difference of type in the return values between WStatCountsStatistic and CashCountsStatistic. This was my trick to unify the outputs and to make work all other functions...

gammapy/estimators/map_profile.py
adonath: @bkhelifi I mostly agree that the choice of naming schemes is subjective, however once a scheme is chosen I think it should be consistent within the software. Otherwise maintenance as well as usability is affected. In this case the functionality is very close to the `ExcessMapEstimator`, which also does much more (like computing counts, background, etc. ). 

That's why I said in one of my other comments, whatever we do, it should just be consistent between the `ExcessMapEstimator` and `ExcessProfileEstimator`. That's all I care about. If you have a better suggestion for a name I'd be open. But you have to convince @registerrier on the renaming of the `ExcessMapEstimator` then...:-)

gammapy/estimators/map_profile.py
adonath: Then maybe put one sentence in the docstring to explain what `orthogonal` means? I found it not obvious. I guess other profile types would be `radial` or similar? I think with the concept of using regions (which I like a lot!) we somehow gave up on simple definitions of profiles such as "along x", "along y" or "radial". In that sense it's probably again the best solution to just keep the region information in the result table. Then the information would be complete...

gammapy/estimators/map_profile.py
bkhelifi: Really, I would like not! The whole purpose of using a SpectrumDataset is to be able to have flux profile. And I rewrote the old class with this objective.
When in the future, the function to_spectrum_dataset will offer options to compute the IRFs in a different manner (e.g. mean over a region, center of a region, etc), then we will fold these options in this estimator.

However, your comment reminds me that I forget to mention an important point about the standard use case of usage of this estimator. We run it after the Ring (ie 1 energy bin) and we make a profile. In this case, the computation of the exposure is correct as the method to_spectrum_dataset compute correctly the effective area (the average exposure divided by time).
In case of usage after the FoVMaker, then we can have several energy bins. And in this case, one can have issues!

So, to conclude, I propose to precise in the documentation to run this estimator with maps produced by the ring. Does this solution acceptable?

gammapy/estimators/map_profile.py
adonath: Typically all functions in gammapy should just preserve the input type, unless there is a very good reason (e.g. performance, precision or memory consumption). That is what numpy does by default as well.

What is the exact problem that you observed in the `CashCountsStatistics` class? We should definitely try to fix it there, otherwise we will repeat the same code in all places where `CashCountsStatistics` is used...

gammapy/estimators/map_profile.py
```diff
@@ -0,0 +1,313 @@
...
+        fermi_prof.peek()
+        plt.show()
+
+        ax = plt.gca()
+        ax.set_yscale('log')
+        ax = fermi_prof.plot("flux", ax=ax)
+
+    """
+
+    def __init__(self, regions, axis, n_sigma = 1, n_sigma_ul=3):
```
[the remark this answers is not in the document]
bkhelifi: It should be... I agree.
However, the function to_spectrum_dataset does not take in argument a mask. So I thinks that this inclusion will create algorithm errors in the profile estimator.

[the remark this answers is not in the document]
bkhelifi: @registerrier : by the way, for this reason, should be not remove the line containing `mask = spds.mask if spds.mask is not None else slice(None)` and all the masks because of this reason?
The mask included in the SpectrumDataset should be empty as the function to_spectrum_dataset does not manage the mask of the associated MapDataset (a possible feature missing in this function, by the way)...

gammapy/estimators/map_profile.py
adonath: Thanks @bkhelifi, for the additional explanation. I think using the `ProfileEstimator` with the ring background only, would be a huge limitation. I'm pretty sure users want to use it with data cubes as well.

I fully agree the `ProfileEstimator` should compute a flux profile as well, but we should do it correctly and take into account the energy dependence of the exposure. So compute the flux internally per energy bin and the sum up. As the energy dependence of the exposure is lost in the final result table, I was in favour removing it, but the result should definitely contain a flux column!

gammapy/estimators/map_profile.py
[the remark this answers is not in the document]
adonath: I just checked and you are right, we do not handle any mask in `MapDataset.to_spectrum_dataset()`. I think this should be changed, I'll make a reminder issue for this. Handling of an exclusion mask for the profile estimator can be implemented in follow up PR then...

gammapy/estimators/map_profile.py
bkhelifi: I fully agree that this metadata deserves an explanation...

We have thought with @registerrier to store the regions in the final results of the profiles. This can be the solution... the class `ImageProfile` having 2 arguments, the table and the list of regions, by removing then this metadata.

Just tell me your preference and I implement it

gammapy/estimators/map_profile.py
bkhelifi: For one case, one returns an array, the other a float. This leads afterwards issue when creating the final table, propagating this issue up to the plotting function for the ULs... (ie the line 295 and 302 of the last version of image_profile.py)

gammapy/estimators/map_profile.py
bkhelifi: Done

gammapy/estimators/map_profile.py
bkhelifi: Fixed and thus simply, thanks to a PR of Régis

gammapy/estimators/map_profile.py
bkhelifi: The last version contains a major upgrade, introducing the management of energy bins and also the flux computation by the dataset. I 'resolve' thus this comment as it does not stand anymore.

gammapy/estimators/map_profile.py
bkhelifi: The dictionnary metadata and their keyword are described within the documentation

gammapy/estimators/map_profile.py
```diff
@@ -0,0 +1,367 @@
...
+        self.regions = regions
+        self.axis = axis
+        self.n_sigma = n_sigma
+        self.n_sigma_ul = n_sigma_ul
+        self.spectrum_datasets = None
```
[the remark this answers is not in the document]
bkhelifi: axis, n_sigma and n_sigma_ul are used in the function `make_prof'. I remove the last one

gammapy/estimators/map_profile.py
```diff
@@ -0,0 +1,367 @@
...
+        datasets : `~gammapy.datasts.Datasets`
+            the list of `~gammapy.datasets.SpectrumDataset` computed in each box
+        """
+        sp_datasets = []
+        for reg in self.regions:
+            spds = dataset.to_spectrum_dataset(reg)
+            sp_datasets.append(spds)
+        return sp_datasets
+
+    def make_prof(self, sp_datasets, steps, spectrum=None):
```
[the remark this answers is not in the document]
bkhelifi: done

gammapy/estimators/map_profile.py
```diff
@@ -0,0 +1,367 @@
...
+        if isinstance(self.regions[0], RectangleSkyRegion):
+            table.meta["PROFILE_TYPE"] = "orthogonal_rectangle"
+        if spectrum is None:
+            spectrum = PowerLawSpectralModel(index=2.0)
+        table.meta["SPECTRAL_MODEL"] = spectrum.to_dict()
+
+        return ImageProfile(table)
+
+
+def make_orthogonal_boxes_new(start_pos, end_pos, wcs, fullwidth, nbins=1):
```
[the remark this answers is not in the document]
bkhelifi: Sorry, I forget to make some cleaning... This function is the correct one: I will rename it and move it in utils.region

gammapy/estimators/map_profile.py
```diff
@@ -0,0 +1,367 @@
...
+                                 angle=rot_angle)
+        regions.append(reg)
+    
+    axis = MapAxis.from_nodes(coords[0].separation(coords))
+    axis.name = 'projected distance'
+
+    return regions, axis
+
+
+def make_orthogonal_rectangle_sky_regions(start_pos, end_pos, wcs, fullwidth, nbins=1):
```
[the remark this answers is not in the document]
bkhelifi: done

gammapy/estimators/map_profile.py
```diff
@@ -0,0 +1,367 @@
...
+    regions = []
+    for i in range(nbins):
+        reg = RectangleSkyRegion(center=coords[i],
+                                 width=box_width,
+                                 height=u.Quantity(fullwidth),
+                                 angle=rot_angle)
+        regions.append(reg)
+    
+    axis = MapAxis.from_nodes(coords[0].separation(coords))
+    axis.name = 'projected distance'
```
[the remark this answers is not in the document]
bkhelifi: done

gammapy/estimators/map_profile.py
```diff
@@ -0,0 +1,367 @@
...
+        Returns
+        --------
+        datasets : `~gammapy.datasts.Datasets`
+            the list of `~gammapy.datasets.SpectrumDataset` computed in each box
+        """
+        sp_datasets = []
+        for reg in self.regions:
+            spds = dataset.to_spectrum_dataset(reg)
+            sp_datasets.append(spds)
+        return sp_datasets
```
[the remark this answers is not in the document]
bkhelifi: done

adonath:
Superseded by #2948.

PR #2925, opened 2020-06-16
Add an option to avoid saving Covariance to file

QRemy:
This PR introduces a "save_covariance" option (True by default) in Models.write() and Datasets.write() so ones can decide to save or not  the covariance file.

### DIFF
--- a/gammapy/datasets/core.py
+++ b/gammapy/datasets/core.py
@@ -222,7 +222,7 @@ class Datasets(collections.abc.MutableSequence):
             datasets.append(dataset)
         return cls(datasets)
 
-    def write(self, path, prefix="", overwrite=False, save_covariance=True):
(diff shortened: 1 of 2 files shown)


gammapy/datasets/core.py
```diff
@@ -222,7 +222,7 @@ def read(cls, path, filedata="_datasets.yaml", filemodel="_models.yaml"):
             datasets.append(dataset)
         return cls(datasets)
 
-    def write(self, path, prefix="", overwrite=False):
+    def write(self, path, prefix="", overwrite=False, save_covariance=True):
```
adonath, line 225: Nitpicky: maybe rename the option to `write_covariance` so that it is in line with the name of the corresponding method on the `Models` class.

Issue #2919, opened 2020-06-10
`plot_spectrum_datasets_off_regions` shows up to 10 sets of off regions


luca-giunti:
**Gammapy version**
v0.17

**Bug description**
This is very much low priority.

The function [plot_spectrum_datasets_off_regions](https://github.com/gammapy/gammapy/blob/9428f86b55850776dfdc95b098106dc6a8a94081/gammapy/visualization/utils.py#L6) don't show all the sets of reflected regions in case the passed `Datasets` contains more than 10 datasets.

**Other information**
The reason is that there are not enough colors to plot all the sets of reflected regions. The colors are defined with `color_cycle = plt.rcParams["axes.prop_cycle"]` which leads to 10 colors (hence the 10 sets of reflected regions that are displayed).

This helper function is not really adapted for the use with many observations, also because:
- we display the run ID in the legend for each observation, and for tens of runs this can become messy
- there is no `alpha` option set for the regions, so if two runs have the same reflected regions only one of them will appear in the plot 

Since this function lives now in `gammapy.visualization`, it might be worth cleaning it up a bit to generalize its use

adonath:
I thinking's general it is tricky to make a plot where all off-regions for many observations  are plotted on the the same sky map. To avoid cluttering of the plot one would need to work with alpha values, line styles and many more colors, as you already suggested. 

I think the plotting functionality we provide will always be limited to standard versions of plots and not be adaptable to every users needs. Especially choice of colors, alpha and fonts etc. is also a matter of taste in many cases. Offering all options to customise a plot puts a lot additional implementation and maintenance effort on our side and I would clearly try to avoid it.

In my view the solution to this problem is documentation. Looking at the code of [plot_spectrum_datasets_off_regions](https://github.com/gammapy/gammapy/blob/master/gammapy/visualization/utils.py#L6) it is really limited and we should also encourage users to just copy and paste from it, when they need a custom version of the plot. 

Back to the issue:
Because of the loop over datasets in the plotting method, it will be hard to implement a clean keyword forwarding and we should probably just rely on the matplotlib property cycle. This means if users want more colors they need to modify the standard property cycle of matplotlib using `plt.rcParams`. But this definitely requires a docs example...

adonath:
I think the action item here would be:
- Modify plot_spectrum_datasets_off_regions to take all properties from the property cycle (i.e. include at least linestyle and alpha)
- Show how to modify the property cycle from "outside" and but this as an example in the docstring

[...]

PR #2932, opened 2020-06-27
Fix plot_spectrum_datasets_off_regions and make customization more flexible

alexiscoutinho:
**Description**
This pull request aims to solve issue #2919. It cycles through `prop_cycle` exposing the dynamic properties. These have priority over the static properties passed to `plot_kwargs` before the loop. This implementation aims to allow the user to freely customize the plot. The only exception is `facecolor` which cannot be read from `rcParams`. The legend can be toggled and has an appropriate default. I also inverted the order of the `ax` attribution as it was raising compatibility errors for me in the master branch and because it makes more sense for the _override_ to come first.

**Dear reviewer**
I would like to know if `Axes` other than WCS are/should be supported here and, if yes, what would be the most generic/appropriate `Axes` object name to put on lines 13-14. Is the rest of the documentaion perfect?

### DIFF
--- a/gammapy/visualization/utils.py
+++ b/gammapy/visualization/utils.py
@@ -3,32 +3,67 @@ import numpy as np
 __all__ = ["plot_spectrum_datasets_off_regions", "plot_contour_line"]
 
 
-def plot_spectrum_datasets_off_regions(datasets, ax=None):
-    """Plot spectrum datasets of regions.
+def plot_spectrum_datasets_off_regions(datasets, ax=None, legend=None, **kwargs):
+    """Plot spectrum datasets' off regions.
 
     Parameters
     ----------
     datasets : list of `SpectrumDatasetOnOff`
-        List of spectrum on-off datasets
+        List of spectrum on-off datasets.
+    ax : `~`
+        .
+    legend : bool
+        Whether to display the legend. By default True if `len(datasets) <= 10`.
+    kwargs : dict
+        Keyword arguments used in `~gammapy.maps.RegionNDMap.plot_region`.
+        Can contain a `cycler.Cycler` in a `prop_cycle` item.
+
+    Notes
+    -----
+    Properties from the `prop_cycle` have maximum priority except `edgecolor`.
+    `edgecolor` is selected from the sources below in this order:
+        `kwargs["edgecolor"]`
+        `kwargs["prop_cycle"]`
+        `~matplotlib.RcParams["axes.prop_cycle"]`
+        `~matplotlib.RcParams["patch.edgecolor"]`
+    `~matplotlib.RcParams["patch.facecolor"]` is never used.
+
+    Examples
+    --------
+    >>> plot_spectrum_datasets_off_regions(datasets, ax, legend=False, lw=2.5)
+    >>> plot_spectrum_datasets_off_regions(datasets, ax, alpha=.3, facecolor='k')
+    >>> plot_spectrum_datasets_off_regions(
+            datasets, ax, ls='--', prop_cycle=plt.cycler('color', list('rgb'))
+        )
+    >>> plt.rc('legend', fontsize=9)
+    >>> plt.rc('patch', edgecolor='blue')
+    >>> plot_spectrum_datasets_off_regions(
+            datasets, ax, legend=True, prop_cycle=plt.cycler('ls', ['-', '-.'])
+        )
     """
     import matplotlib.pyplot as plt
     import matplotlib.patches as mpatches
 
-    ax = plt.gca(projection=datasets[0].counts_off.geom.wcs) or ax
-
-    color_cycle = plt.rcParams["axes.prop_cycle"]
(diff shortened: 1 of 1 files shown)


gammapy/visualization/utils.py
```diff
@@ -3,32 +3,67 @@
...
+    --------
+    >>> plot_spectrum_datasets_off_regions(datasets, ax, legend=False, lw=2.5)
+    >>> plot_spectrum_datasets_off_regions(datasets, ax, alpha=0.3, facecolor='k')
+    >>> plot_spectrum_datasets_off_regions(
+            datasets, ax, ls='--', prop_cycle=plt.cycler('color', list('rgb'))
+        )
+    >>> plt.rc('legend', fontsize=9)
+    >>> plt.rc('patch', edgecolor='blue')
+    >>> plot_spectrum_datasets_off_regions(
+            datasets, ax, legend=True, prop_cycle=plt.cycler('ls', ['-', '-.'])
```
adonath, line 41: Maybe modify the example here to cycle both through the color and lines style? This might be even more helpful...

alexiscoutinho: In this last example I wanted to show that it is possible to not cycle edgecolors (if the user wants to have a specific color for different `Datasets` in the same `ax`, for example). However, it would be inappropriate to set each static `edgecolor` through rcParams. Therefore, I will improve it and make my intentions clearer by grouping this example with a second function call. I'll also boost the previous example with a `color` + `ls` cycle.

alexiscoutinho:
What about the `ax` object name(s)? Does Hpx play a role in it? It seems it's always converted to a `WCSAxes`, so should I write `~astropy.visualization.wcsaxes.WCSAxes`?

adonath:
@alexiscoutinho Yes, exactly HPX maps are converted to WCS for plotting. So no special mention of HPX needed...

alexiscoutinho:
I addressed the documentation and example issues. In the process, I noticed that the function wasn't behaving as wanted when dealing with multiple datasets in the same `Axes`. I fixed that and also took the time to allow a lot more customization on the legend. The legend icons now show two joined circles instead of a rectangle. The maximum width of the code is now 96 columns (at the signature), should I crop more?

adonath:
@alexiscoutinho Concerning the codacy issue: in Python mutable type should basically never used as default arguments, because they are created at function definition. Consider the following example:
```python3
def f(a=[]):
    a.append(len(a))
    return a
    
print(f())
print(f())
print(f())
print(f())
``` 

Which prints:
```
[0]
[0, 1]
[0, 1, 2]
[0, 1, 2, 3]
```
Even if the function is called with the default value (without any arguments), the default value changes each time the function is called. This can lead to unpredictable behaviours. If a mutable default is needed we use the following pattern in Gammapy:
```python3
def f(a=None):
    if a is None:
        a = []
    a.append(len(a))
    return a

```

alexiscoutinho:
@adonath I am currently working on supporting ellipse icons (in the legend) with proper rotations and relative sizes. However, to know the rotation of the ellipses in the pair (which simulates the rotation of the off regions), I need the observation pointing position which isn't explicitly stored in `SpectrumDatasetOnOff`. As a workaround, I was getting it from the `WCS` object: `dataset.counts_off.geom.wcs.wcs.crval`. Unfortunately, this is an array and I couldn't find any explicit information about the frame to turn it to a `SkyCoord`. The closest thing I found was `dataset.counts_off.geom.wcs.wcs.ctype` which gives the coordinate names and projection. Unless there is a function which turns `CTYPEia` to `SkyCoord` frame names, I would need to make a bunch of string comparisons (switch) which wouldn't be neat. Therefore, what's the best way to get the observation pointing position from a `SpectrumDatasetOnOff`?

adonath:
@alexiscoutinho What is the status here? I think once you removed the default dict for the legend keywords this PR is ready to be merged...

alexiscoutinho:
@adonath I am currently working on supporting ellipse icons/pairs with proper rotations and relative sizes. The bulk is done, but I still need to code some fallback cases if the region is neither a circle nor an ellipse. These unsupported patches will default to rectangles or something else. However, I can still do that dict fix just to make it ready.

adonath:
Thanks @alexiscoutinho, but please don't continue with the ellipses. I honestly think this is too much of a detail, which will lead to a lot of code, we have to maintain in future. The point is that the work wouldn't be done with the ellipses, because the regions can really have arbitrary shapes (such as ellipses, polygons, rectangles, annuli, etc.). It will be way too complicated to support all these shapes in the legend. I think your time is better spend elsewhere (just in case you would further like to contribute to Gammapy...). However I appreciate the PR in the current state and once the issue with the dict as default is fixed it can be merged...

alexiscoutinho:
@adonath Indeed there is a lot more code. I think the whole thing more than tripled in lines just by adding ellipse support and "exception" handling. And this is without those optimizations that tend to duplicate code. If so, I'm gonna bench that extra work for now.

Issue #2917, opened 2020-06-05
Fit converges with success=True when parameters are stuck at the borders


luca-giunti:
**Gammapy version**
v0.17

**Bug description**
3D fitting with MINUIT appears to be succesfull (`success=True`) even if the optimizer got stuck at the boundaries of the allowed range for some fitted parameter.

**Expected behavior**
The fit should not report `success=True` if one or more parameters have assumed the max/min value of their allowed range, or at least some error message should be printed.

**To Reproduce**
See this working [example](https://github.com/luca-giunti/share/blob/master/analysis_1.ipynb).
It is a simple adaptation of the `analysis_1.ipynb` notebook. The relevant cells are:
- [19] here, I changed the first-guess value of `lon_0`: instead of starting from nearby the crab position (`lon_0=83.633`), I start from an offset position (`lon_0=83.4`), so that the model will try to shift to `lon_0=83.633`. But I also impose a max value of `83.5` for `lon_0`, so that I know that the optimizer will get stuck at `lon_0=83.5`. To understand why MINUIT gets stuck there, check section 5.3.1 in the [MINUIT manual](https://root.cern.ch/download/minuit.pdf)
- [22] MINUIT wrongly tells us that `success=True`
- [24] I check that indeed the model got stuck at `lon_0=83.5`. Also, we can notice that the error on `lon_0` has a crazy value: `1.643e-07` (which is how I found out this issue in the first place)

**Other information**
This behavior is completely expected in MINUIT fitting, so it's not really a gammapy problem in this sense. However, what should be done by gammapy is printing some error if a fit converges with parameters at the boundaries (and maybe in that case force `success=False`).

AtreyeeS:
Is this something that we want gammapy to do or it is the users responsibility to check that nothing hits the borders?

luca-giunti:
In my experience, it is way too easy to overlook this kind of issue. I really think we should raise a warning. Imagine someone has a large number of models stored in `MapDataset.models`: he should check one by one that there aren't parameters at the edge?

adonath:
[...]

PR #2937, opened 2020-07-07
Send warning in case parameter value is set outside boundaries

luca-giunti:
**Description**
This PR solves #2917 . The value setter for the `Parameter` class is now responsible to check that:

- the value is contained in the min/max range;
- the value is different from the min/max bounds by at least 1%. This level of tolerance is a somewhat arbitrary choice.

In case either of these condition is False, a `RuntimeWarning` is sent.

I checked the sane behavior here: https://github.com/luca-giunti/share/blob/master/Parameters_ouside_bounds.ipynb
During a fit, in case MINUIT hits the min/max borders for some parameters, the user will be notified. So even if the outcome of the fit is `success=True`, it will be clear that something went wrong. This can be seen in cell [22] of https://github.com/luca-giunti/share/blob/master/analysis_1.ipynb .

**Dear reviewer**
#2917 can be closed as soon as this is merged.

### DIFF
--- a/gammapy/modeling/parameter.py
+++ b/gammapy/modeling/parameter.py
@@ -3,6 +3,7 @@
 import collections.abc
 import copy
 import itertools
+import warnings
 import numpy as np
 from astropy import units as u
 from gammapy.utils.table import table_from_row_data
@@ -78,6 +79,8 @@ class Parameter:
         self.name = name
         self._link_label_io = None
         self.scale = scale
+        self.min = min
+        self.max = max
 
         # TODO: move this to a setter method that can be called from `__set__` also!
         # Having it here is bad: behaviour not clear if Quantity and `unit` is passed.
@@ -89,8 +92,6 @@ class Parameter:
             self.factor = value
             self.unit = unit
 
-        self.min = min
-        self.max = max
         self.frozen = frozen
         self._error = error
 
@@ -208,6 +209,7 @@ class Parameter:
     @value.setter
     def value(self, val):
         self._factor = float(val) / self._scale
+        self._check_value_within_bounds()
 
     @property
     def quantity(self):
@@ -219,6 +221,23 @@ class Parameter:
         val = u.Quantity(val, unit=self.unit)
         self.value = val.value
         self.unit = val.unit
+
+    def _check_value_within_bounds(self):
+        """Send a warning if value is outside the min/max range (or at the border within 1%)"""
+        min_nan = np.isnan(self.min)
+        max_nan = np.isnan(self.max)
+        min_ok = self.value > self.min and ~np.isclose(
+            self.value - self.min, 0, atol=1e-2
+        )
+        max_ok = self.value < self.max and ~np.isclose(
+            self.value - self.max, 0, atol=1e-2
+        )
+        if (~min_nan and ~min_ok) or (~max_nan and ~max_ok):
+            warnings.warn(
+                f"Setting parameter '{self.name}' value {self.value} near or outside bounds [{self.min}, {self.max}]",
+                RuntimeWarning,
+                stacklevel=2,
+            )
 
     def __repr__(self):
         return (
--- a/gammapy/modeling/parameter.py
+++ b/gammapy/modeling/parameter.py
@@ -234,7 +234,9 @@ class Parameter:
                         f"Value {self.value} is outside bounds [{self.min}, {self.max}] for parameter '{self.name}'"
                     )
                 else:
-                    raise ValueError(f"Value {self.value} is outside bounds [{self.min}, {self.max}] for parameter '{self.name}'")
+                    raise ValueError(
+                        f"Value {self.value} is outside bounds [{self.min}, {self.max}] for parameter '{self.name}'"
+                    )
 
     def __repr__(self):
         return (
--- a/gammapy/modeling/parameter.py
+++ b/gammapy/modeling/parameter.py
@@ -138,8 +138,7 @@ class Parameter:
     @factor.setter
     def factor(self, val):
         self._factor = float(val)
-        if self.frozen is False:
-            self.check_limits()
+        self.check_limits()
 
     @property
     def scale(self):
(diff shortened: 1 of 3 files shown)


gammapy/modeling/parameter.py
```diff
@@ -220,6 +222,23 @@ def quantity(self, val):
...
+        min_nan = np.isnan(self.min)
+        max_nan = np.isnan(self.max)
+        min_ok = self.value > self.min and ~np.isclose(
+            self.value - self.min, 0, atol=1e-2
+        )
+        max_ok = self.value < self.max and ~np.isclose(
+            self.value - self.max, 0, atol=1e-2
+        )
+        if (~min_nan and ~min_ok) or (~max_nan and ~max_ok):
+            warnings.warn(
```
adonath, line 236: I think we should use `log.warn()` here, so that users get access to the logging configuration...

gammapy/modeling/parameter.py
```diff
@@ -220,6 +222,23 @@ def quantity(self, val):
         self.value = val.value
         self.unit = val.unit
 
+    def _check_value_within_bounds(self):
+        """Send a warning if value is outside the min/max range (or at the border within 1%)"""
+        min_nan = np.isnan(self.min)
+        max_nan = np.isnan(self.max)
+        min_ok = self.value > self.min and ~np.isclose(
+            self.value - self.min, 0, atol=1e-2
```
adonath, line 230: The docstring states it would be a relative tolerance but `np.isclose()` uses `atol` here. In any case the 1% seems a like an arbitrary choice to me. I checked and e.g. [sherpa emits a warning](https://github.com/sherpa/sherpa/blob/358ab71a0e1f1e639a33e71bcbd3b4fa0fe7a979/sherpa/fit.py#L420) only in case of equality of the value to the boundary. Have you tested what happens without the tolerance?

gammapy/modeling/parameter.py
```diff
@@ -208,6 +209,7 @@ def value(self):
     @value.setter
     def value(self, val):
         self._factor = float(val) / self._scale
+        self._check_value_within_bounds()
```
adonath, line 212: During fitting we actually go through the `factor` setter, so probably the warning is never emitted. To avoid the code duplication I would propose to change move the check to the `factor` setter and change the `value` setter to call `self.factor = float(val) / self._scale`...

gammapy/modeling/parameter.py
adonath: Addendum: for `iminuit` it might be that without the tolerance this warning is maybe never emitted, because the parameter values only approach the boundary (because of the `atanh` parameter scaling...). Does `iminuit ` emits a warning on its own?

gammapy/modeling/parameter.py
luca-giunti: OK, I didn't know that. However, it seems that the warning is already emitted (check cell [22] of https://github.com/luca-giunti/share/blob/master/analysis_1.ipynb) from the value setter somehow. But if you prefer, I will do as you propose

adonath: Probably the warning is emitted on I/O already, but I think not during fitting...

gammapy/modeling/parameter.py
luca-giunti: > The docstring states it would be a relative tolerance but np.isclose() uses atol here

Ups! I fix it sorry

> for iminuit it might be that without the tolerance this warning is maybe never emitted, because the parameter values only approach the boundary (because of the atanh parameter scaling...)

Exactly, I think a standard equality will never be satisfied for a MINUIT fit. That's why I introduced this tolerance value.

> Does iminuit emit a warning on its own?

I don't think so, otherwise we wouldn't have this issue at all, no?

luca-giunti:
@adonath What to do with [this](https://github.com/gammapy/gammapy/blob/4a0bf2a66dd10a7bb603eec29516e27d40586616/gammapy/modeling/models/spatial.py#L289)? Actually, I don't think it's a good habit to initialize parameters at their minimum by default. On one hand, `e=0` seems to be the best default, so we should remove the min value. But on the other, removing the min/max here is a shame because we know that mathematically those are hard bounds, and for e<0 the model don't make any sense...

adonath:
@luca-giunti If the parameter is frozen, it's maybe fine if the value is at the boundary and we just don't emit a warning?

luca-giunti:
Thanks @adonath . Another question: indeed, I realized that raising a warning with an arbitrary `rtol`/`atol` choice is bad. For example, `rtol=1e-2` would cause the unwanted warning
```
>>> lon_0 = Parameter("lon_0", "303 deg", min=300)
Setting parameter 'lon_0' value 303.0 near or outside bounds [300.0, nan]
```
Seeing how `numpy.isclose()` is implemented, it seems that `atol` and `rtol` should be somehow adjusted case by case, or false positives/negatives could arise. Any simple solution for this?

adonath:
@luca-giunti I doubt that there is a simple solution to this. Let's discuss again in tomorrows meeting...

QRemy:
I think we should avoid calling `_check_value_within_bounds ` in the setter, but call it at the beginning of fit, returning a strict error there, and at the end of the fit, returning only a warning.

adonath:
@QRemy In general I think it's better to raise a warning as early as possible (i.e. already at the stage of setting the parameter value), but I agree that in the case e.g. where iminuit tries values outside the limits (for whatever reason...) this would lead to possibly "annoying" warnings during fitting. 

If @luca-giunti also prefers the check pre and post fit, I would propose to add `Parameter.check_limits(strict=False)` and `Parameters.check_limits(strict=False)` to implement this behaviour...

luca-giunti:
@adonath @QRemy , I agree that, if it's true that iminuit routinely tries values outside the bounds , these warnings will become annoying. 

However I think by doing what you suggest we lose the possibility to forbid (or at least discourage) things like
```
p = Parameter(value=0, min=1)
``` 
unless we call `Parameter.check_limits(strict=False)` also in the setter. 

So, wrapping it up:
1) I will move the check code from the function `Parameter._check_value_within_bounds()` to `Parameter.check_limits(strict=False)` and `Parameter.check_limits(strict=False)`. `strict=True` will raise a `ValueError`, `strict=False` will log a simple warning. These functions will be called in the parameter setter (`strict=False`), at the beginning of a fit (`strict=True`), at the end of a fit (`strict=False`)
2) `Parameter.check_limits()` will verify that the parameter value is >= min and <= max. In other words, I will drop the checks `np.isclose(value , min)` and `np.isclose(value , max)`, because we ran into the issue of defining a proper default tolerance value
3) I will open a PR on the `iminuit` page, asking to add a flag or emit a warning when some fitted parameter has converged "too close to" min/max. If this will be implemented in `iminuit` at some point, we will possibly remove the call of `Parameter.check_limits` after the fit...

Do you agree?

adonath:
@luca-giunti Sounds good to me!

luca-giunti:
Related iminuit issue: https://github.com/scikit-hep/iminuit/issues/449

PR #2938, opened 2020-07-08
Add shorter tags for models

QRemy:
Redefine most of models tags with a list including shorter aliases to be used with model.create(tag) and in YAML serialization. By default the tag returned in model.to_dict() is the 0th tag which is also the class name.


adonath:
Thanks a lot @QRemy! As discussed in the meeting this morning, please adapt the YAML tags to the one proposed in PIG 21 here: https://github.com/gammapy/gammapy/blob/0ab86da3d9fca7046e67b40f23332442d1e715b9/docs/development/pigs/pig-21.rst#simplify-yaml-representation

I'll continue with the review this afternoon...

PR #2941, opened 2020-07-09
Add PiecewiseBrokenPowerLawSpectralModel

QRemy:
Add PiecewiseBrokenPowerLawSpectralModel, a generalised broken power law  defined by a set of energy and intensity values (nodes) that are piecewise connected by power laws.


adonath:
The plan for this is now outlined in #2944.

PR #2944, opened 2020-07-09
PIG 21 - Model framework improvements

adonath:
**Description**

This pull request adds a draft for PIG proposing further improvements and simplifications to the Gammapy model framework. 

**Dear reviewer**
The PIG is still in a draft state it will be improved step by step by discussion and comments in this PR.


docs/development/pigs/pig-21.rst
```diff
@@ -0,0 +1,348 @@
...
+    bpwl_norm = PiecewiseBrokenPowerlawNormSpectralModel(norms=, energies=)
+    const_norm = ConstantNormSpectralModel(norm=)
+
+    # typically used like
+	template = TemplateSpectralModel()
+    model = template * pwl_norm
+
+    model = BackgroundModel(
+		map=map,
+		spectral_model=pwl_norm
```
registerrier, line 53: Isn't it simpler and more intuitive that `BackgroundModel` takes a `spectral_norm` parameter rather than a `SpectralModel` which are assumed to have a flux unit.

docs/development/pigs/pig-21.rst
```diff
@@ -0,0 +1,348 @@
...
+
+We propose to add energy dependence to the ``TemplateSpatialModel`` and
+replace the current ``SkyDiffuseCube`` by:
+
+
+.. code::
+
+    spatial_model = TemplateSpatialModel.read("my_cube.fits")
+
+    model = SkyModel(
```
registerrier, line 101: I think this is really confusing. 
If a spectral model is supposed to return a flux it should be clearly doing this. So passing a `NormModel` for a `SpectralModel` is confusing.
I'd rather have the possibility to do :
```
pwl_norm = PowerLawNormModel()   (or MultModel)
template = TemplateSpatialModel.read("my_cube.fits")
template = pwl_norm * template
```
or if implementing the `CoumpoundModel` for the template is too complex having a property to set the spectral normalization (or simply passing it on init).

BTW, with this scheme (`TemplateSpatialModel` for `SkyDiffuseCube`) we have to find a way to make it clear which component actually carries the proper flux unit.

docs/development/pigs/pig-21.rst
adonath: The `BackgroundModel` should take a `SpectralNormModel` of course...will correct this...

docs/development/pigs/pig-21.rst
adonath: That's the main point of this proposal: giving up the strong assumption that a `SpectralModel` should return a flux. This means we don't have a single convention, but rely on users combining the models correctly instead, if the flux is contained in the spatial model component instead.

I can see how the solution is simple and clean:
```
pwl_norm = PowerLawNormModel()   (or MultModel)
template = TemplateSpatialModel.read("my_cube.fits")
template = pwl_norm * template
```
But right now the model that combines the spectral, spatial and temporal is the `SkyModel`. So the last line in the code example `template = pwl_norm * template` should actually return a `SkyModel`...the only alternative solution would be that it returns an (energy-dependent) `SpatialModel`, but I think this would be even more confusing and require that we allow to use `SkyModel` without a spectral model, like:

```
pwl_norm = PowerLawNormModel()   (or MultModel)
template = TemplateSpatialModel.read("my_cube.fits")
template = pwl_norm * template

source = SkyModel(spatial_model=template)
```
That's why I proposed to combine it on the `SkyModel` level...

docs/development/pigs/pig-21.rst
```diff
@@ -0,0 +1,348 @@
...
+
+Proposal
+========
+
+Spectral Norm Models
+--------------------
+
+For the purpose of adjusting template based models we propose to introduce a new
+class of spectral models. Those spectral models feature a norm parameter instead
+of amplitude and are named using the ``NormSpectralModel`` suffix:
```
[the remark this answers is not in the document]
adonath: Thanks @registerrier for the comment here. Here are a few remarks from my side:

- I fully agree with the spatial multiplicative models, however there is technically no need to introduce a special class for it (except for a `CompoundSpatialModel` or similar), because we can just re-use our exiting spatial models. They don't have an amplitude parameter and always return a "norm" already. In a way this is an inconsistency, because they don't have a "norm" suffix in the name.
- I think the core of the problem here is that the "amplitude" parameter is bound to the `SpectralModel` class. Which prevents a "clean use" as a model to scale other existing models (which was the only motivation to introduce the `NormSpectralModel`...). Maybe we re-evaluate the solution to attach the "amplitude" to the `SkyModel`?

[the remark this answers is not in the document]
adonath: Possibly we can also check whether we can simplify the way to re-use our normal `SpectralModel`. As I have shown before, this works:

```
amplitude = Parameter("amplitude", unit="")
pwl = PowerLawSpectralModel(amplitude=amplitude, index=2)
```
Do you think it's an improvement if we could get this to work:
```
pwl = PowerLawSpectralModel(amplitude=1, index=0)
```
?

PR #2945, opened 2020-07-09
PSF-convolved spatial model caching in MapEvaluator

QRemy:
Implement caching of PSF-convolved spatial model in MapEvaluator.
(speed up 3FHL validation fit and flux points computation by a factor of about two)

### DIFF
--- a/gammapy/datasets/map.py
+++ b/gammapy/datasets/map.py
@@ -1894,13 +1894,10 @@ class MapEvaluator:
         if self.model.spatial_model and not isinstance(self.geom, RegionGeom):
             spatial_pars = list(self.model.spatial_model.parameters.values)
             spatial_conv = self._spatial_conv_cached
-            if self._spatial_pars_cached != spatial_pars or self._spatial_conv_cached is None:
+            if self._spatial_pars_cached != spatial_pars or spatial_conv is None:
                 self._spatial_pars_cached = spatial_pars
-                spatial_conv = Map.from_geom(geom=self.geom, data=1)
                 geom_image = self.geom.to_image()
-                integral = self.model.spatial_model.integrate_geom(geom_image)
-                spatial_conv.data = np.ones(value.shape) * integral.data
-                spatial_conv.unit = integral.unit
+                spatial_conv = self.model.spatial_model.integrate_geom(geom_image)
                 if self.psf and self.model.apply_irf["psf"]:
                     spatial_conv = self.apply_psf(spatial_conv)
                 self._spatial_conv_cached = spatial_conv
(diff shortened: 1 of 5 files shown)


QRemy:
I was just thinking to add a hidden "out-sync" or "has_changed" property on parameter, parameters, model and models with decorators on every method that modify them so we can track their changes and also make the datasets.models/.parameters/.evaluators trigger update only when necessary.

adonath:
@QRemy I agree this might be useful for models as well (and possibly solve part of the problems we discussed this morning in the dev call). However I would propose to start with the `Parameter.changed_value` and `Parameters.changed_value` in this PR and then extend the concept after some discussion to models next week.

adonath:
@QRemy I think this status flag can be part of the public API, there is no need to hide it...

QRemy:
@adonath 
I tried to cache the parameters values on the parameters, or to not cache them at all but update a status flag each time a setter is called. The problem with these two approaches is that you cannot track the status of each evaluator independently. Indeed you can reset the status flag (or update cached parameters values) only at the higher level in datasets.stat_sum when all evaluators are updated otherwise it results in inconsistencies, but then caching is available only for the fit and not at lower level for example when calling directly dataset.npred.
So I think we should keep the current solution where the evaluators are fully responsable of cachin,g so their status can be tracked individually (anyway all computation heavy operation go through them, so it make sense).

QRemy:
For the 2D map convolution with a 3D kernel,  what if the kernel is an array ? We don't know the target geom exactly, so we have to add a fake MapAxis as shown below, do you see a cleaner way to do this ?

```python
       geom = self.geom.copy()
        if isinstance(kernel, PSFKernel):
            kmap = kernel.psf_kernel_map
            if not np.allclose(
                self.geom.pixel_scales.deg, kmap.geom.pixel_scales.deg, rtol=1e-5
            ):
                raise ValueError("Pixel size of kernel and map not compatible.")
            kernel = kmap.data.astype(np.float32)
            geom.axes.append(kmap.geom.axes[0])

        if len(self.data.shape) < len(kernel.shape):
            convolved_data = np.empty(
                ((kernel.shape[0]),) + self.data.shape, dtype=np.float32
            )
            nodes = range(kernel.shape[0] + 1)
            if geom.data_shape != convolved_data.shape:
                geom.axes.append(MapAxis(nodes))
                print("append",geom.data_shape)

            for idx in nodes[:-1]:
                convolved_data[idx] = conv_function(
                    self.data.astype(np.float32), kernel[idx], **kwargs
                )
            return Map.from_geom(geom=geom, data=convolved_data)
        else:
            convolved_data = np.empty(self.data.shape, dtype=np.float32)
            for img, idx in self.iter_by_image():
                ikern = Ellipsis if kernel.ndim == 2 else idx
                convolved_data[idx] = conv_function(
                    img.astype(np.float32), kernel[ikern], **kwargs
                )
            return self._init_copy(data=convolved_data)
```

adonath:
@QRemy For the case, where the kernel is a 3D array, I think it's fine to throw an error. For there internal use-case we always use a `PSFKernel` object anyway...

PR #2947, opened 2020-07-09
Improve spectral residuals plot

luca-giunti:
**Description**
Introduced vertical error bars for spectral plotting.

For residuals, vertical errorbars are computed as:
- sqrt(residual), if `method=="diff"`
- 1,  if `method=="diff/model"` or `method=="diff/sqrt(model)"`

**Dear reviewer**
Here is what the plots look like
Screenshot from 2020-07-09 18-12-45


gammapy/datasets/spectrum.py
```diff
@@ -370,9 +367,14 @@ def plot_counts(self, ax=None):
         import matplotlib.pyplot as plt
 
         ax = plt.gca() if ax is None else ax
-
-        self.npred_sig().plot(ax=ax, label="mu_src")
-        self.excess.plot(ax=ax, label="Excess")
+        self.excess.plot(
+            ax=ax, label="Measured excess", yerr=np.sqrt(self.excess.data.flatten())
+        )
+        self.npred_sig().plot(
```
registerrier, line 373: I would not put errors on the predicted signal. You could either use points or a continuous line (e.g. an histogram line).

gammapy/datasets/spectrum.py
```diff
@@ -422,7 +424,11 @@ def plot_residuals(self, method="diff", ax=None, **kwargs):
         residuals = self.residuals(method=method)
         label = self._residuals_labels[method]
 
-        residuals.plot(ax=ax, color="black", **kwargs)
+        if method == "diff":
+            yerr = np.sqrt(np.abs(residuals.data.flatten()))
```
registerrier, line 428: The error is not on the residuals but rather on the total data. The issue is what to do when you have no signal.
You could use `CashCountsStatistic` to compute the errors and residuals.

registerrier: Or `WStatCountsStatistic` for `SpectrumDatasetOnOff`

luca-giunti: @registerrier You suggest we actually remove the three options "diff", "diff/model" and "diff/sqrt(model)" and we always compute residuals using `CashCountsStatistic`/`WStatCountsStatistic`?

adonath: I would definitely keep the simple methods, they are still useful. In general there are two options:
- Add another counts statistic based residual method as option and do the equivalent on `MapDataset` for consistency. Ideally in a follow-up PR.
- Keep the simple residual and plotting methods on the dataset as is and implement more advanced methods in a `MapDatasetPlotter` (API left to be defined...), where we then have access to all `Estimator` classes and can also compute correlated residual maps etc.

I think it adds confusion to our API if we offer e.g. a TS based or `ExcessMapEstimator` based residual method in `MapDataset.residuals`. This would require importing  the Estimator in the dataset and create circular dependencies etc. I find it also weird if a call to `dataset.plot_residuals()` takes O(~10-100) seconds to compute a residuals map. In general I'm trying to avoid to make the `MapDataset` a "god-object". Even if it's probably the "way of least resistance" . I think in terms of code maintenance, API design and configuration it will not do us any good. Even adding the counts-based method requires adding and duplicating  further parameters to `datasets.residuals()` such as `n_sigma` etc. The only API design that work long-term is some kind of hierarchical thing like:

```python3
class MapDatasetPlotter(Plotter):
    def __init__(self, output_path, residual_estimator, residual_plot_config):
        self.residual_estimator = residual_estimator
        self.output_path = output_path
        self.residual_plot_config = residual_plot_config
   
    def plot_residuals(self, dataset):
        """Plot residual map"""
        result = self.residual_estimator.run(dataset)
        self.result["sqrt_ts"].plot(self.residual_plot_config)
        plt.savefig(self.output_path)
    
```
Introducing the `Plotter` API will require a PIG and some design work...

luca-giunti: I agree, @adonath . I actually started coding a more elaborate solution to add new `methods` for residuals computation to take into account the counts statistics properly, but it quickly became a duplication of the `ExcessMapEstimator`. In this PR, I would rather stick with a simple adaptation of the spectral plotting functions, keeping the `dataset.plot_residuals()` methods quick and handy. A proper statistical treatment will be done with a `Plotter` API as you suggest...

luca-giunti:
Example plot:
example

adonath:
Thanks @luca-giunti, the resulting plot looks really nice!

PR #2948, opened 2020-07-10
Profile bis

bkhelifi:
**Description**
This pull request replaces the PR #2923

**Dear reviewer**
I have kept the old file with the old maker and old class. Can I remove it now?

PR #2949, opened 2020-07-10
Excess profile class

bkhelifi:
**Description**
This pull request contains only the storage class of the excess profile estimator. 

**Dear reviewer**
The PR #2948 has been splitted and contains only the estimator.

### DIFF
--- a/gammapy/estimators/__init__.py
+++ b/gammapy/estimators/__init__.py
@@ -9,6 +9,7 @@ from .lightcurve import *
 from .profile import *
 from .sensitivity import *
 from .ts_map import *
+from .image_profile import *
 
 ESTIMATOR_REGISTRY = Registry(
     [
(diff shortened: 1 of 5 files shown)


adonath:
This PR is linked to #3075. I'll relabel it for v1.0

adonath:
After #3491 and clarification of `npred_excess` in #3581 , `FluxPoints` can be used to represent an excess profile as well. We will add a new format such that `FluxPoints.to_table(sed_type="flux", format="profile")` returns a corresponding table object, that users can interact with. I'll close this PR now.

adonath:
Thanks a lot @bkhelifi for this prototype! I think it has inspired many discussions and also helped to arrive at the final solution of the an ND `FluxPoints` object, which handles all cases.

PR #2950, opened 2020-07-13
Remove IRFStacker class

adonath:
**Description**

This pull request removes the `IRFStacker` class and replaces its use with `EDispKernelMap.stack()`. I also decided to remove the `make_mean_psf()` and `make_psf()` helper methods. I moved over the existing tests and they passed without changing any reference value (except for adjusting a 4th digit being the dot...).

I would like to replace the `make_mean_psf()` with a little docs example. Do you have any suggestion for a good place for this @registerrier?


adonath:
For reference here is the code snippet for the mean PSF:
```python3
from gammapy.maps import MapAxis, RegionGeom
from gammapy.makers import MapDatasetMaker
from gammapy.irf import PSFMap

data_store = DataStore.from_dir("$GAMMAPY_DATA/hess-dl3-dr1")

energy_axis = MapAxis.from_energy_bounds(0.1, 30, 100, unit="TeV", name="energy_true")
rad_axis = MapAxis.from_bounds(0, 1, 100, unit="deg", name="theta")

geom = RegionGeom.create("icrs;point(83.63, 22.01)", axes=[rad_axis, energy_axis])

maker = MapDatasetMaker()

stacked_psf = PSFMap.from_geom(geom=geom)

for obs in data_store.get_observations([23523, 23526, 23559, 23592]):
    psf = maker.make_psf(geom=geom, observation=obs)
    stacked_psf.stack(psf)

psf = stacked_psf.get_energy_dependent_table_psf()
psf.plot_containment_vs_energy()
```

It is not exactly short, but completely compatible in API with the `MapDataset` reduction, which I think is a strong advantage...

PR #2957, opened 2020-07-27
Adapt MapDataset.stack

AtreyeeS:
**Description**

This pull request adapts MapDataset.stack() to properly account for the safe mask

**Dear reviewer**

The tests are not well adapted for this change. I am sure how to devise a good one.
Also, I do not know if it is useful to have stacking of `EDispMap`, considering it will only be used for event sampling.

### DIFF
--- a/gammapy/datasets/map.py
+++ b/gammapy/datasets/map.py
@@ -446,12 +446,11 @@ class MapDataset(Dataset):
             self.counts.stack(other.counts, weights=other_mask_safe)
 
         if self.exposure and other.exposure:
-            mask_image = self.mask_safe.reduce_over_axes(func=np.logical_or)
-            self.exposure *= mask_image.data
-            # TODO: apply energy dependent mask to exposure. Does this require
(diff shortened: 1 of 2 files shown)


adonath:
Thanks a lot @AtreyeeS! I think as a minimal test we should implement something like I did here:

https://github.com/adonath/notebooks-public/blob/master/energy_threshold_stack.ipynb

This compares the npred of two stacked datasets against the stacked npred of the same datasets. If the energy threshold is applied correctly this should give consistent results. I think two small test datasets, filled with ones in counts and background but two different energy thresholds and a diagonal `EDispKernelMap` should be sufficient...

I think currently there are still test fails on Travis-CI, please fix.

AtreyeeS:
Thanks @adonath !
I updated a test.
I do not know if the logic is a bit convoluted in the entire process... Please do take a look.
Also, I think we might adapt `EdispKernelMap.from_diagonal()` - it returns an empty exposure map which can give confusing results later.

PR #2958, opened 2020-07-30
Adapt `ExcessMapEstimator` to optionally apply mask_fit 

AtreyeeS:
**Description**

This pull request adapts the `ExcessMapEstimator` to optionally apply a mask_fit and directly return excess 2D images from a 3D dataset.

**Dear reviewer**

Here, I apply the masks before doing the convolution. @luca-giunti once mentioned that this can sometimes lead to artefacts along the edge of the mask, but we have to test under which circumstances this can happen.

### DIFF
--- a/gammapy/estimators/excess_map.py
+++ b/gammapy/estimators/excess_map.py
@@ -92,11 +92,12 @@ class ExcessMapEstimator(Estimator):
     tag = "ExcessMapEstimator"
     available_selection = ["errn-errp", "ul"]
 
-    def __init__(self,
+    def __init__(
+        self,
         correlation_radius="0.1 deg",
         n_sigma=1,
         n_sigma_ul=3,
-        selection='all',
+        selection="all",
         apply_mask_fit=False,
         return_image=False,
     ):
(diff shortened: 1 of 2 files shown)


AtreyeeS:
Thanks @registerrier ! I made some of the changes.
I did not understand the part of "compute the sum over axis in a range to be specified on init" - This mask can be provided through the mask fit, no? Is passing it on the init necessary ?

adonath:
@AtreyeeS The idea that came up during discussion between @registerrier and myself, was that all `Estimator` classes will take a new energy binning as parameters to specify in which energy range the resulting maps should be computed (very analog to what we do in the `FluxPointsEstimator`). But it's completely fine to not implement anything in this PR. We will add this uniformly at a later stage...

gammapy/estimators/excess_map.py
```diff
@@ -16,29 +16,48 @@
...
+    dataset, kernel, apply_mask_fit=False, return_image=False
+):
     """Return CountsDataset objects containing smoothed maps from the MapDataset"""
     # Kernel is modified later make a copy here
     kernel = copy.deepcopy(kernel)
     kernel.normalize("peak")
 
+    mask = np.ones(dataset.data_shape, dtype=bool)
+    if dataset.mask_safe:
+        mask = dataset.mask_safe
```
registerrier, line 29: Maybe `mask *= dataset.mask_safe` 
to avoid changing values of `mask_safe`?

PR #2959, opened 2020-07-31
Lazy dataset loading prototype

adonath:
**Description**

This pull request is a near production ready proof of concept to extend the lazy loading we have for `Observation` objects to `MapDataset`. The use-case is clearly to be able to load a large number of observations O(1e3) in to the Python session without loading the actual data. I think we should offer two options:

- Cache the data in memory, which is needed for fitting
- Don't cache the data, which is needed for stacking
- Possibly a combination e.g. for the estimation of light curves , where the data needs to be cached for the fit, but once the fit is done the data could go out of memory. So we possibly need an explicit `dataset.clear_data()` or similar.

The question is what other options do we offer? Is lazy loading always enabled? Any other potential issue you see?

### DIFF
--- a/gammapy/data/observations.py
+++ b/gammapy/data/observations.py
@@ -80,8 +80,8 @@ class Observation:
 
     @property
     def gti(self):
-        events = self.obs_filter.filter_gti(self._gti)
-        return events
+        gti = self.obs_filter.filter_gti(self._gti)
+        return gti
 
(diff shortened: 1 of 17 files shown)


adonath:
This misses some documentation, which will be added in a follow-up PR. It is feature for "power-users" anyway...

PR #2961, opened 2020-07-31
Add possiblity to plot Map with non-spatial axes of size 1

registerrier:
**Description**

This pull request changes the behavior of `WcsNDMap.plot()` to allow for plotting of maps with 1 energy axis and only one bin. Since this is the default format we impose on datasets, we shouldn't force the user to do `datasets.counts.sum_over_axis().plot()` for datasets with only one energy bin.

**Dear reviewer**


registerrier:
The two tutorial notebooks still using unnecessary `get_image_by_idx` were modified.
Tests pass. Merging now.

PR #2964, opened 2020-08-03
Modify MapDataset.to_image()

AtreyeeS:
**Description**

This pull request fixes `MapDataset.to_image()` and adds `EdispKernel.to_image()` and `EdispKernelMap.to_image()`

**Dear reviewer**

The test results in `test_asmooth_map` have changed significantly because earlier the models were being dropped from the imaged dataset


AtreyeeS:
One issue is here: https://github.com/gammapy/gammapy/blob/f55930ae2a05c7c2d643183318ca8ce179c9c704/gammapy/estimators/asmooth_map.py#L129

A sum_over_axes() on the exposure is not quite correct anymore - the exposure should be weighed by the spectrum.
We can have an optional `spectrum` argument, but I am not sure that this is the most intuitive place to for the user.

adonath:
Thanks @AtreyeeS, I think the assumption with `.sum_over_axes()` here was the the exposure contains only one energy bin anyway...so `.to_image()` was called before. I have to think a bit how we handle this and comment again.

adonath:
@AtreyeeS @registerrier I think we still need to compute a weighted exposure map for the map estimators. Probably it's best if we just compute it locally in the `TSMapEstimator` and `ASmoothMapEstimator` for now.

registerrier:
We might need it as well in `ExcessMapEstimator` if we want to compute fluxes there as well.

What we need is not really a `sum_over_axes` though, but we need an exposure map in reco energy.
We can get it by estimating the `npred` of the following model 
`model = SkyModel(spatial=ConstantSpatialModel, spectral=PowerLawSpectralModel)`

adonath:
Thanks @registerrier! Good point, this also accounts for the energy dispersion if available...

AtreyeeS:
> We might need it as well in `ExcessMapEstimator` if we want to compute fluxes there as well.
> 
> What we need is not really a `sum_over_axes` though, but we need an exposure map in reco energy.
> We can get it by estimating the `npred` of the following model
> `model = SkyModel(spatial=ConstantSpatialModel, spectral=PowerLawSpectralModel)`

Is it OK to keep an index of 2 for the `PowerLawSpectralModel` for all cases?

registerrier:
Maybe we can add the utility function `compute_reco_energy_exposure(dataset, spectral_model)`  somewhere. 
It would create a `MapEvaluator` with a `SkyModel` defined as above and the `dataset` geometry,` exposure` and `edisp`.
Then you compute the `npred=map_evaluator.compute_npred()`
After that you only need to determine for each energy reco bin `(emin, emax)`:
`ref_flux = spectral_model.integral(emin,emax)`
and `reco_exposure = npred/ref_flux`

If you divide an `ExcessMap` by `reco_exposure`, you should get an (integrated) flux.

adonath:
@AtreyeeS For now I just called `_map_spectrum_weight` locally in the `TSMapEstimator` and `ASMoothMapEstimator` this should fix the test fails...in a later PR we can then introduce the exposure map in repo energy...

PR #2967, opened 2020-08-04
Allow for varying energy range between datasets in FluxPointEstimator

adonath:
**Description**

This pull request allows a varying energy range of datasets in the `FluxPointsEstimator`.

### DIFF
--- a/gammapy/maps/geom.py
+++ b/gammapy/maps/geom.py
@@ -308,6 +308,29 @@ class MapAxis:
 
         self._nbin = nbin
 
+    def is_aligned(self, other, atol=1e-2):
+        """Check if other map axis is aligned.
+
+        Two axes are aligned if their center coordinate values map to integers
+        on the other axes as well.
+
+        Parameters
+        ----------
+        other : `MapAxis`
+            Other map axis.
+        atol : float
+            Absolute numerical tolerance for the comparison measured in bins.
+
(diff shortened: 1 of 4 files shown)


gammapy/maps/geom.py
```diff
@@ -308,6 +308,29 @@ def __init__(self, nodes, interp="lin", name="", node_type="edges", unit=""):
 
         self._nbin = nbin
 
+    def is_aligned(self, other, atol=2e-2):
+        """Check if other map axis is aligned.
+
+        Two axes are aligned if their center coordinate values map to integers
```
registerrier, line 314: Testing only on centers does not guarantee that the test will work. Imagine the following situation:
```
ax1=MapAxis([1,2,3], interp='lin', node_type='edges')                                                                                    
ax2=MapAxis([1.5,2.5], interp='log', node_type='center')   
```
These two axes would be considered aligned although their edges are definitely not. One should test that `interp_mode` is the same. I suppose  this is sufficient.

adonath: Thanks, I'll address this in a follow up PR...

adonath: Fixed in master: https://github.com/gammapy/gammapy/commit/c0e58f48fdb990256d3b1e683200f7649b300f0e

PR #2968, opened 2020-08-05
Clean up estimators

registerrier:
**Description**

This pull request is a further clean up of estimators.

`steps` names are unified between the various estimators ('ts', 'err', 'earn-errp','ul'). 
The TS map estimator is simplified (removed newton and least square methods) and unified (added asymmetric errors). 
Also the `n_sigma` factor is applied also to covariance error estimates to make sure asymmetric and symmetric errors are consistent.

**Dear reviewer**

### DIFF
--- a/gammapy/estimators/ts_map.py
+++ b/gammapy/estimators/ts_map.py
@@ -100,18 +100,6 @@ class TSMapEstimator(Estimator):
         Sample down the input maps to speed up the computation. Only integer
         values that are a multiple of 2 are allowed. Note that the kernel is
         not sampled down, but must be provided with the downsampled bin size.
-    method : str ('root')
-        The following options are available:
-
-        * ``'root brentq'`` (default)
-            Fit amplitude by finding the roots of the the derivative of the fit
-            statistics using the brentq method.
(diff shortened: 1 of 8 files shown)


gammapy/estimators/ts_map.py
```diff
@@ -146,20 +132,12 @@ def __init__(
         model=None,
         kernel_width="0.2 deg",
         downsampling_factor=None,
-        method="root brentq",
-        error_method="covar",
         n_sigma=1,
         ul_method="covar",
```
adonath, line 136: Probably we should remove the `ul_method` choice as well? I think computing the upper limits on the covariance errors is not really a good idea, so probably we should just rely on likelihood profiles...

gammapy/estimators/ts_map.py
```diff
@@ -146,20 +132,12 @@ def __init__(
         model=None,
         kernel_width="0.2 deg",
         downsampling_factor=None,
```
adonath, line 134: I think we should also remove the downsampling option here soon and provide a more general solution that also works for other map estimators. However this can be done in a different PR...

PR #2970, opened 2020-08-05
Clean up estimators - 3

registerrier:
**Description**

This pull request removes the `ts` and `err` steps. `steps` are now set on `init` and no longer in `run`.  An `available_selection` member is added similarly to `Makers`. A general method is added on the base class to set the selection properly.

A set of getter and setter properties have been added as well to allow for change on-the-fly.

**Dear reviewer**

### DIFF
--- a/docs/tutorials/analysis_3d.ipynb
+++ b/docs/tutorials/analysis_3d.ipynb
@@ -418,9 +418,9 @@
    "outputs": [],
    "source": [
     "# TODO: clean this up\n",
-    "estimator = ExcessMapEstimator(correlation_radius=\"0.1 deg\")\n",
+    "estimator = ExcessMapEstimator(correlation_radius=\"0.1 deg\", selection=[])\n",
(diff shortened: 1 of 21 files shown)


registerrier:
I corrected the remaining errors. 
One unsatisfactory aspect is the necessity to explicitly pass `selection=[]` or `selection=None` to avoid computing asymmetric errors and ul. In particular, this looks a bit weird in the tutorials.

adonath:
@registerrier I guess it's the question of what we choose as default: either compute everything or compute the minimal quantities. I'll think about it once more and comment again tomorrow...

adonath:
@registerrier What about the following:
- An `"all"` shortcut which applies all steps defined in `Estimator.selection_available` or `Estimator.selection_all`
- A `"min"` shortcut which applies a minimal set of steps defined in `Estimator.selection_min`
- And then just take any arbitrary list defined by the user like `selection=["min", "errn-errp"]`

PR #2973, opened 2020-08-07
Modified required columns for likelihood SED type

registerrier:
**Description**

This pull request modifies the required columns on the `Likelihood` SED type. The `norm-scan` and `stat-scan` columns are now optional. This allows the estimator to run even without asking for the likelihood scan. In this case the SED type return is `dnde`. The code has been modified to return a `likelihood` type in the general case.

**Dear reviewer**

### DIFF
--- a/gammapy/estimators/flux_point.py
+++ b/gammapy/estimators/flux_point.py
@@ -870,12 +870,9 @@ class FluxPointsEstimator(FluxEstimator):
 
         table = table_from_row_data(rows=rows, meta={"SED_TYPE": "likelihood"})
 
-        flux_points = FluxPoints(table)
-        if "norm-scan" not in self.selection:
-            return flux_points.to_sed_type("dnde")
-        else:
-            return flux_points
-        
+        #TODO: this should be changed once likelihood is fully supported
+        return FluxPoints(table).to_sed_type("dnde")
+
     @staticmethod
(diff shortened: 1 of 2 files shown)


gammapy/estimators/flux_point.py
```diff
@@ -870,8 +869,13 @@ def run(self, datasets, steps="all"):
             rows.append(row)
 
         table = table_from_row_data(rows=rows, meta={"SED_TYPE": "likelihood"})
-        return FluxPoints(table).to_sed_type("dnde")
 
+        flux_points = FluxPoints(table)
+        if "norm-scan" not in self.selection:
```
adonath, line 874: Why do we need this extra case? Isn't moving the `"norm-scan"` to the optional columns sufficient?

registerrier: For some reason, the current version of the `FluxPointsEstimator` transforms the `FluxPoints` object to `dnde`.  I kept this option only for the case without the `norm-scan`. But indeed it can be removed.
Do you know why it was returning a `dnde` `FluxPoints`?

adonath: I think for "historical" reasons, some people said it's maybe weird if the returned object does not contain any flux or dnde column. I think it's fine like this and users can just convert to any type they want with a single call to `.to_sed_type()` after...

PR #2976, opened 2020-08-11
Fix filename handling in read/write methods

alexiscoutinho:
Wrapped `filename` in `make_path()`, as in `Map.read`, to allow for environment variables.

### DIFF
--- a/gammapy/datasets/flux_points.py
+++ b/gammapy/datasets/flux_points.py
@@ -108,7 +108,7 @@ class FluxPointsDataset(Dataset):
 
         table["mask_fit"] = mask_fit
         table["mask_safe"] = self.mask_safe
-        table.write(filename, overwrite=overwrite, **kwargs)
+        table.write(make_path(filename), overwrite=overwrite, **kwargs)
 
     @classmethod
     def from_dict(cls, data, models, **kwargs):
--- a/gammapy/datasets/map.py
+++ b/gammapy/datasets/map.py
@@ -863,7 +863,7 @@ class MapDataset(Dataset):
         overwrite : bool
             Overwrite file if it exists.
         """
-        self.to_hdulist().writeto(make_path(filename), overwrite=overwrite)
+        self.to_hdulist().writeto(str(make_path(filename)), overwrite=overwrite)
 
     @classmethod
     def _read_lazy(cls, name, filename, cache):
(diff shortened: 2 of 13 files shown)


alexiscoutinho:
I see... I searched all python scripts for the strings `read`, `write`, `open` and `file` and found that the same problem occurs in: `DataStoreMaker.read_events_info`, `FluxPointsDataset.write`, `IRFMap.read`, `IRFMap.write`, `PSF3D.write`, `HESSMultiGaussPSF._read_ascii`, `EnergyDependentMultiGaussPSF.write`, `PSFKing.write` and `EnergyDependentTablePSF.write`. Should I rename this PR and push these fixes or wait for the _automation_?

On the other hand, I'm not sure if this fix should apply to:
- Is `MapDataset._read_lazy` intended to not treat/expand environment variables?
- Is it appropriate to treat environment variables in _download.py_, _downloadclasses.py_, _jupyter.py_, _notebooks_links.py_ and _notebooks_process.py_?

adonath:
Thanks @alexiscoutinho for looking into this. I agree expanding environment variables could be useful for writing as well. 

A few comments:
- `MapDataset._read_lazy`, always goes through `Map.read()` no change needed there...
- `DataStoreMaker.read_events_info` is not part of our public API, no change needs there...
- Again `HESSMultiGaussPSF._read_ascii` is not part of our public API, no change needs there...
- I remember in older version of Astropy `fits.open()` and `hdulist.writeto()`, didn't support a `Path` object, just to be sure please convert to a string before, like `hdulist.writeto(str(filename))`

I think all the other places are appropriate, so please go ahead and add the change to this PR...

alexiscoutinho:
It seems like `IRFMap` is not part of the public API, so should I keep it as it is? The issue with `MapDataset._read_lazy` I tried to point out is on line 872: `kwargs["gti"] = GTI(Table.read(filename, hdu="GTI"))`. `Table.read` would raise a `FileNotFoundError` and this method is accessible through the public `MapDataset.read`.

adonath:
Thanks @alexiscoutinho, I see what you mean with `._read_lazy`, simply changing the line to `GTI.read(filename)` should fix the issue...

adonath (approved):
Thanks @alexiscoutinho! I have no further comments...

PR #2977, opened 2020-08-11
Add Npred caching for SkyDiffuseCube

QRemy:
Modify MapEvaluator.compute_npred() to support caching of SkyDiffuseCube npred values.


QRemy:
This will be resolved by #3012 by removing SkyDiffuseCube and replacing it with a TemplateSpatialModel in a SkyModel for which the caching is set.

PR #2978, opened 2020-08-11
Add support for energy dependent spatial models

QRemy:
- TemplateSpatial model support for cubes
- remove TemplateSpatial norm parameter
- Add optional energy argument in spatial model evaluation to support 3d case
- simplify units checks in TemplateSpatial and add _check_unit() method in SkyModel init

### DIFF
--- a/gammapy/modeling/models/cube.py
+++ b/gammapy/modeling/models/cube.py
@@ -78,6 +78,7 @@ class SkyModel(SkyModelBase):
 
         self.apply_irf = apply_irf
         self.datasets_names = datasets_names
+        self._check_unit()
         super().__init__()
 
     @property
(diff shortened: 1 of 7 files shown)


gammapy/modeling/models/spatial.py
```diff
@@ -109,9 +121,12 @@ def position_error(self):
         )
 
     def evaluate_geom(self, geom):
-        """Evaluate model on `~gammapy.maps.Geom`."""
         coords = geom.get_coord(frame=self.frame)
-        return self(coords.lon, coords.lat)
+        energy_name = _get_energy_axis(geom)
+        if energy_name is not None:
+            return self(coords.lon, coords.lat, coords[energy_name])
```
adonath, line 127: I think we should rely on numpy broadcasting here, just as we do in `SkyModel.evaluate_geom()`, so evaluate the model not using the full 3D coordinate arrays, but relying on the outer product between the spatial and energy coordinates...

gammapy/modeling/models/spatial.py
```diff
@@ -156,8 +171,11 @@ def plot(self, ax=None, geom=None, **kwargs):
 
         data = self.evaluate_geom(geom)
         m = Map.from_geom(geom, data=data.value, unit=data.unit)
-        _, ax, _ = m.plot(ax=ax, **kwargs)
 
+        if len(m.data.shape) > 2:
```
adonath, line 175: Maybe for consistency we just copy the `Map` pattern here? So expose the interactive plotting using a separate `.plot_interactive()` method? I can see that merging results in a more compact API, but I think consistency is more important here...

gammapy/modeling/models/cube.py
```diff
@@ -91,6 +92,26 @@ def _check_covariance(self):
                 [model.covariance for model in self._models],
             )
 
+    def _check_unit(self):
+        from astropy.time import Time
+        from gammapy.data.gti import GTI
+
+        # evaluate over a test geom to check output unit
+        # TODO simpler way to test this ?
```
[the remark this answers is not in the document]
QRemy: I tried but the evaluate method is missing for PointSpatialModel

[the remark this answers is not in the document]
adonath: Indeed you are right, so just keep as is...

gammapy/modeling/models/spatial.py
QRemy: Is this working for a full 3d template model ? SkyDiffuseCube use the SkyModelBase.evaluate_geom that look like this one, no ?

adonath: Yes, it should work. In `Map.interp_by_coord()`, we call `MapCoord.create()`, which does the broadcasting of the arrays on init:

```python3
import numpy as np

axis = MapAxis.from_energy_bounds("1 TeV", "10 TeV", nbin=10)

m = Map.create(width=10, axes=[axis])

m.interp_by_coord(
{
    "skycoord": m.geom.to_image().get_coord().skycoord,
    "energy": axis.center.reshape((-1, 1, 1))
}
)
```

PR #2981, opened 2020-08-13
Add basic NormSpectralModels

QRemy:
- add NormPowerLawSpectralModel
- add NormLogparabolaSpectralModel
- add NormExpCutoffPowerLawSpectralModel
- left a TODO about where and how to document the difference with their corresponding amplitude-based model.
- left a TODO about plot method redefinition
- more complex models will be added in separate PR


adonath:
Thanks @QRemy, for consistency with the PIG I add one commit to rename to `...NormSpectralModel`, if we find good reasons to use the `Norm` as a prefix, I'm open to change it consistently again...we can re-discuss in one of the next dev meetings.

PR #2988, opened 2020-08-20
Add test for on-off fitting

AtreyeeS:
**Description**

This pull request adds a test and extends the ring tutorial for `MapDatasetOnOff` modelling and fitting.

**Dear reviewer**

### DIFF
--- a/docs/tutorials/ring_background.ipynb
+++ b/docs/tutorials/ring_background.ipynb
@@ -66,7 +66,13 @@
     "from gammapy.makers import RingBackgroundMaker\n",
     "from gammapy.estimators import ExcessMapEstimator\n",
     "from gammapy.maps import Map\n",
-    "from gammapy.datasets import MapDatasetOnOff"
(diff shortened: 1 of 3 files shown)


adonath:
@AtreyeeS I just checked the residual map of the on-off fit and it doesn't look right. I think we have to overwrite the `npred()` of the `MapDatasetOnOff` to include the background...

AtreyeeS:
Yes indeed @adonath ! The residual maps do not change because the source models are not taken into account. I filed #2989  for this exact reason - the estimator does not consider the `npred` for `MapDatasetOnOff` while making the residual maps

adonath:
@AtreyeeS I meant something in addition: even the `dataset_on_off.plot_residuals()` does not look right, because the `MapDatasetOnOff.background` is not included in the npred. However in `stat_array()` I think it's handled correctly...

AtreyeeS:
Closing due to merge conflicts

PR #2990, opened 2020-08-24
Optional exposure map for the EdispMap and PSF in the MapDataset

LauraOlivera:
**Description**

This pull request makes the presence of an exposure map for the EdispKernel and PSF optional, when loading a MapDataset from a hdulist.

**Dear reviewer**


adonath (approved):
Thanks @LauraOlivera! I don't have any additional comments...

We could think about creating the exposure map for the IRF maps from the "normal" exposure map, but for now this is just fine.

adonath:
Thanks again, @LauraOlivera!

PR #2993, opened 2020-08-26
Reuse FluxPointsEstimator in LightCurveEstimator

adonath:
**Description**

This pull request is a proof of concept to re-use the `FluxPointsEstimator` in the `LightCurveEstimator`. Therein arguments for this are the following:

- Consistency in configuration parameters
- Code-reuse
- Possibility to compute light-curves in energy bins

However this change modifies the current method, because by introducing the `FluxPointsEstimator` the fit range is limited to the specified energy range. Previously the fit was always done on the full energy range and then flux was derived from the best fit model in the specified energy range.


registerrier (approved):
Thanks @adonath ! This looks good to me.

adonath:
Thanks @registerrier! I'll go ahead and merge now...

PR #2997, opened 2020-09-02
Adapt SpectrumDataset to use background_model instead of background

AtreyeeS:
**Description**

This pull request cleans up `SpectrumDataset` to use a `background_model` instead of a fixed `background`.
I think having a `background_model` also makes comparisons with `wstat` fitting easier.

**Dear reviewer**

I have tried to match the `MapDataset` API, so removed `SpectrumDataset.background`.  `SpectrumDatasetOnOff.background` still exists. I do not know if it might be good to have a `background` property which simply returns `self.background_model.evaluate()`.


adonath (approved):
Thanks a lot @AtreyeeS! I don't have any further comments...

PR #2998, opened 2020-09-03
Add e_edges to ExcessMapEstimator

registerrier:
**Description**

This pull request introduces the `e_edges` parameter on the `ExcessMapEstimator` to harmonize with other `Estimators`. There is a already `make_image` option that already exists. We might adapt it to follow the general structure.

**Dear reviewer**

### DIFF
--- a/gammapy/datasets/map.py
+++ b/gammapy/datasets/map.py
@@ -1409,9 +1409,6 @@ class MapDatasetOnOff(MapDataset):
         gti=None,
         meta_table=None,
     ):
-        if mask_fit is not None and mask_fit.dtype != np.dtype("bool"):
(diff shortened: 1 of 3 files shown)


adonath (commented):
Thanks @registerrier! 

Just a first quick comment: I think the `return_image` option becomes obsolete at this point? My proposal would be, that we always return the energy axes as well (even if it is just one bin) and users can drop the axis, if they want after...

registerrier:
Yes, the `return_image` option is obsolete in practice. It can be replaced by proper usage of `e_edges`. We can remove it.
I will update the PR accordingly.

adonath (approved):

Issue #2989, opened 2020-08-20
Behaviour of `ExcessMapEstimator`


AtreyeeS:
Currently, the `ExcessMapEstimator` behaves differently for `MapDataset` and `MapDatasetOnOff`.
In the first case, the `npred` is taken as the background - ie, the background and the source models are both considered. In the second case, only the `counts_off` and the `background` are considered, and the counts from the source ignored.

This is quite confusing. We should enforce similar behaviour, and I think that the `npred` should be considered in case of  `MapDatasetOnOff` as well. What do you think @registerrier @adonath ?

adonath:
@AtreyeeS I guess, including sources in the `MapDatasetOnOff` background does not make it and on-off measurement anymore? I need to think about this some more...

adonath:
@AtreyeeS Can you check: isn't it that the background is missing from `MapDatastOnOff.npred()`?

AtreyeeS:
Yes, `MapDatasetOnOff.npred()` does not have the background added. But I am not sure if it is OK to just add the background in this case to the model predicted counts? Does it still behave properly with `wstat` in that case?

registerrier:
With wstat the actual background is a function of `npred`. The background method we use only give the estimated background in case no emission in present (i.e. `alpha * counts_off`).

You cannot use the background in the `npred` otherwise the `wstat` does not mean anything. 

For `ExcessMapEstimator` to behave consistently with `MapDataset`  and `MapDatasetOnOff`, you have to modify the `WStatCountStatistic` so that you can deal with WStat with an additional Poisson component with known intensity. It should fairly straightforward to do.

For the moment, there is a `mu_bkg` parameter in `wstat` but it does not do what we need (it simply replaces the proper profile estimate by a fixed value (I don't know why).

adonath:
Thanks @registerrier, for the explanation! However in my comment above I meant something slightly different: even for `wstat` one can get an expected background given the current npred from the the signal (see https://docs.gammapy.org/0.17/api/gammapy.stats.get_wstat_mu_bkg.html#gammapy.stats.get_wstat_mu_bkg). Can't we modify `MapDatasetOnOff.background` to return this background estimate and then include it the `MapDatasetOnOff.npred()` as well? The `stat_array()` remains unchanged...

adonath:
I have thought a bit more about this:
[...]

PR #3000, opened 2020-09-04
Adapt Wstat handling to take signal counts into account

AtreyeeS:
**Description**

This pull request modifies `MapDatasetOnOff/SpectrumDatasetOnOff/WStatCountsStatistic.background`
to take into account the `n_on` and the `mu_sig` as discussed in #2988 #2989 

**Dear reviewer**

This leads to some significant changes in the tests, for some of which I do not really understand the expected result. Mentioning the same inline.


gammapy/estimators/excess_map.py
```diff
@@ -34,30 +34,33 @@ def convolved_map_dataset_counts_statistics(
...
+            a_on = a_on.sum_over_axes(keepdims=True)
+            a_off = a_off.sum_over_axes(keepdims=True)
 
-        background_conv = background.convolve(kernel.array)
         n_off_conv = n_off.convolve(kernel.array)
+        mu_sig_conv = npred.convolve(kernel.array)
 
         with np.errstate(invalid="ignore", divide="ignore"):
-            alpha_conv = background_conv / n_off_conv
+            alpha_conv = a_on.convolve(kernel.array) / a_off.convolve(kernel.array)
```
AtreyeeS, line 58: Is this convolved alpha correct?

gammapy/estimators/sensitivity.py
```diff
@@ -61,13 +63,12 @@ def estimate_min_excess(self, dataset):
         n_off = dataset.counts_off.data
 
         stat = WStatCountsStatistic(
-            n_on=np.ones_like(n_off), n_off=n_off, alpha=dataset.alpha.data
```
AtreyeeS, line 64: Why was n_on chose as `1` previously? In the absence of a signal estimate, `n_on` should be as expected from the off counts, no?

gammapy/estimators/tests/test_excess_profile.py
```diff
@@ -72,9 +74,9 @@ def test_radial_profile():
...
     assert_allclose(imp_prof[7]["x_ref"], 0.15, atol=1e-4)
     assert_allclose(imp_prof[7]["counts"], [980.0, 980.0], atol=1e-2)
-    assert_allclose(imp_prof[7]["excess"], [784.0, 784.0], atol=1e-2)
-    assert_allclose(imp_prof[7]["sqrt_ts"], [23.886444, 23.886444], atol=1e-5)
-    assert_allclose(imp_prof[7]["errn"], [-34.075141, -34.075141], atol=1e-5)
-    assert_allclose(imp_prof[0]["ul"], [75.834983, 75.834983], atol=1e-5)
-    assert_allclose(imp_prof[0]["flux"], [7.99999987e-06, 8.00000010e-06], atol=1e-3)
-    assert_allclose(imp_prof[0]["solid_angle"], [6.853891e-07, 6.853891e-07], atol=1e-5)
\ No newline at end of file
+    assert_allclose(imp_prof[7]["excess"], [98.0, 98.0], atol=1e-2)
```
AtreyeeS, line 77: I have not used the `ExcessProfileEstimator`. Looking at the test results, I do not know if the expected behaviour is still correct or totally messed up!

Also, looking at the doc strings I am a bit confused if this works with both `Spectrum` and `Map` Datasets and OnOff Datasets for the two cases. It is tested only for `SpectrumDatasetOnOff`

gammapy/estimators/tests/test_sensitivity.py
```diff
@@ -47,12 +47,12 @@ def test_cta_sensitivity_estimator(spectrum_dataset):
     assert_allclose(row["energy"], 1.33352, rtol=1e-3)
     assert_allclose(row["e2dnde"], 3.40101e-11, rtol=1e-3)
     assert_allclose(row["excess"], 334.454, rtol=1e-3)
-    assert_allclose(row["background"], 3600, rtol=1e-3)
+    assert_allclose(row["background"], 3000, rtol=1e-3)
```
AtreyeeS, line 50: Again, is the marginalised background expected here or `alpha*n_off`

gammapy/estimators/sensitivity.py
adonath: The information `n_on` is not used in this case. So I guess it's just some place-holder value. One could maybe use `NaN` to make it more clear...

AtreyeeS: I see.
But should this not call - `WStatCountsStatistic._excess_matching_significance` ?

AtreyeeS: Also, `SensitivityEstimator.estimate_min_excess` supposedly takes a `SpectrumDataset` as input but uses `WStatCountsStatistic`.
Should it be adapted to use either `cash` or `wstat` as per the case, or does this work only for `SpectrumDatasetOnOff`?

registerrier:
Note there are merge conflicts with the `ExcessMapEstimator`. Can you resolve them?

gammapy/datasets/spectrum.py
```diff
@@ -607,7 +616,7 @@ def peek(self, figsize=(16, 4)):
         ax1.set_title("Counts")
 
         if isinstance(self, SpectrumDatasetOnOff) and self.counts_off is not None:
-            self.background.plot_hist(ax=ax1, label="alpha * N_off")
+            self.background.plot_hist(ax=ax1, label="background")
```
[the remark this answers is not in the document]
registerrier: I guess that `peek` might return `alpha * n_off`. In particular, since no source model is actually displayed.

But I have no strong opinion there. Of course one needs to document properly what is actually plotted.

gammapy/datasets/spectrum.py
```diff
@@ -836,12 +845,22 @@ def __str__(self):
 
     @property
     def background(self):
-        """"""
-        return self.alpha * self.counts_off.data
+        """from wstat formula"""
+        if self.counts:
+            n_on = self.counts.data
+        else:
+            n_on = self.alpha.data * self.counts_off.data
```
registerrier, line 852: I am not sure what it means to compute the background this way. 
It would seem logical that the method would fail if no `counts` is present. Otherwise we have to assume something concerning what they should be, no?

gammapy/datasets/tests/test_map.py
```diff
@@ -1052,15 +1052,15 @@ def test_stack_dataset_dataset_on_off():
     dataset_on_off.counts_off += 1
     dataset.stack(dataset_on_off)
 
-    assert_allclose(dataset.background_model.map.data, 0.2)
+    assert_allclose(dataset.background_model.map.data, 0.166667, rtol=1e-3)
```
registerrier, line 1055: I think this should not have changed (there is no model defined on the `MapDatasetOnOff`). I suppose the issue is connected with the issue in `stack`.

gammapy/datasets/tests/test_map.py
```diff
@@ -1052,15 +1052,15 @@ def test_stack_dataset_dataset_on_off():
...
 
 
 @requires_data()
 def test_info_dict_on_off(images):
     dataset = get_map_dataset_onoff(images)
     info_dict = dataset.info_dict()
     assert_allclose(info_dict["counts"], 4299, rtol=1e-3)
-    assert_allclose(info_dict["excess"], -22.52, rtol=1e-3)
+    assert_allclose(info_dict["excess"], -22.5135, rtol=1e-3)
```
registerrier, line 1063: Again this should not have changed if no source model is defined on the `MapDatasetOnOff`.

gammapy/datasets/tests/test_map.py
```diff
@@ -1076,7 +1076,7 @@ def test_info_dict_on_off(images):
     assert_allclose(info_dict["a_on"], 0.068363324, rtol=1e-4)
     assert_allclose(info_dict["a_off"], 322.83185, rtol=1e-4)
     assert_allclose(info_dict["alpha"], 0.0002117614, rtol=1e-4)
-    assert_allclose(info_dict["excess"], -22.52295, rtol=1e-4)
+    assert_allclose(info_dict["excess"], -22.513524, rtol=1e-4)
```
registerrier, line 1079: Same remark here. It shouldn't change.

registerrier (commented):
Thanks @AtreyeeS .

There is an issue with `stack` because it needs `alpha*n_off`.

I realize looking at the test value changes that we change a bit the philosophy here. So we might think a bit about how to present things.

Basically when `mu_sig=0` the background is not `alpha*count_off` but `alpha/(alpha+1)*(counts+counts_off`, because the background is deduced from both measurements. 
So we are not exactly in the Li&Ma presentation where the so-called excess `counts - alpha * counts_off` is in fact the estimate of `mu_sig`. 

So we have an issue of definition because `excess` are two different things here.

gammapy/datasets/spectrum.py
AtreyeeS: This I can remove. I added this because `SensitivityEstimator` was using `dataset.background` without `dataset.counts`. I later realised it was just calling it to get the `geom`, so this if condition can go.

gammapy/datasets/tests/test_map.py
AtreyeeS: This I was a bit confused about - Does it make sense to stack a `MapDatasetOnOff` on a `MapDataset`?

gammapy/datasets/tests/test_map.py
AtreyeeS: But `excess = counts - background`. `background` changed from `alpha*n_off` to `alpha/(alpha+1)*(counts+counts_off)`

gammapy/stats/tests/test_counts_statistic.py
```diff
@@ -83,17 +83,12 @@ def test_cash_excess_matching_significance(mu_bkg, significance, result):
...
-    (10, 5, 0.3, [8.5, 3.5853812867949024, 3.365860865528742e-4]),
-    (10, 23, 0.1, [7.7, 3.443415522820395, 5.74416016688779e-4]),
-    (1, 20, 1.0, [-19, -4.590373638528086, 4.424532535784618e-06]),
-    (
-        5 * ref_array,
-        1 * ref_array,
-        1 * ref_array,
-        [4.0, 1.7061745691234782, 0.087975582112],
-    ),
+    (1, 2, 1, [-0.5, -0.5044155295, 0.613969407]),
```
registerrier, line 86: Why does it change here?

registerrier (changes requested):
Thanks @AtreyeeS ! 

I have left some comments inline.

The main one would be that we discussed using `counts_off_normalized` for the alpha*n_off property. 
Also some changes in the CountsStatistics are not totally clear. Is it possible to split into 2 PRs? 
I don't understand why `wstat` tests should change.

gammapy/stats/counts_statistic.py
```diff
@@ -74,12 +74,15 @@ def compute_errp(self, n_sigma=1):
...
-            )
+            try:
+                errp[it.multi_index] = brentq(
+                    self._stat_fcn,
+                    self.excess[it.multi_index],
+                    max_range[it.multi_index],
+                    args=(self.TS_max[it.multi_index] + n_sigma ** 2, it.multi_index),
+                )
+            except ValueError:
+                errp[it.multi_index] = -self.n_on[it.multi_index]
```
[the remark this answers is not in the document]
AtreyeeS: I mirrored what was done for `errn`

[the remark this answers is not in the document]
AtreyeeS: I see now - it just needs to be prepared for a `ValueError`. Changed it now

AtreyeeS:
Thanks @registerrier for the review! The changes in the test values were because of confusion between `excess` and `background`. Made the changes!

registerrier (approved):
Thanks @AtreyeeS ! Everything looks good!

PR #3001, opened 2020-09-07
Improve the plot/peek interface, documentation and figures

alexiscoutinho:
**Description**
This PR aims to make the **datasets** sub-package's `plot` and `peek` user interface more flexible, robust and standardized. It clarifies the `plot_counts` and `plot_excess` methods of `SpectrumDataset`. It also fixes some minor visual issues, improves the presentation of some figures and improves some already written documentation. If it's accepted, it could be extended to other sub-packages.


alexiscoutinho:
I immediately notice that a couple of notebooks would have to have `vmin` and `vmax` wrapped in `wcsnd_kwargs=dict()` when calling `MapDataset.plot_residuals`. Should I update the notebooks or break the code symmetry and turn `wcsnd_kwargs` into `**kwargs` after `regnd_kwargs`?

alexiscoutinho:
Not only the previous signature is much more convenient, it also isn't less symmetrical than the attempted one, since all `plot_residuals` methods have `**kwargs`. Therefore, I reverted signature. However, @adonath, would you prefer `regionND_kwargs` to be all lowercase?

alexiscoutinho:
When fixing notebooks, I tried to be the least invasive possible on subjective matters. Thus, I kept most `figsize=(8, 6)` during the conversion. However, @adonath, noting that the new default is `(8, 7)`, should I use it instead and clean the tutorials' code?

adonath:
Thanks a lot @alexiscoutinho, indeed the `.peek()` methods and plotting methods which plot multiple things need unification!

However I think we need a bit of discussion to come up with a uniform scheme, that also fits what we had done before. In general we agreed in Gammapy, that we only offer limited configuration for plotting in Gammapy, just because plotting is mostly a matter of taste and there are way too many possibilities to support all. So e.g. for styling we just rely on Matplotlib's RC params and keyword forwarding to Matplotlib methods.

I think in general we have three different cases to solve:

### Plotting  one thing

`.plot_something()` methods which only plot a single thing take an optional axes argument and optional keywords which are just forwarded to the corresponding Matplotlib function, such as `imshow()`, `plot()`, `errrorbar()`, `pcolormesh()` etc. If no `ax` keyword is given, internally we just call `plt.gca()`.  In this case we decided to not take a `fig` argument, because the figure can always be create outside of the function call like:

```python3
fig = plt.figure(**fig_kwargs)
ax = fig.add_subplot()
data.plot_something(ax=ax)
```
This is nice because multiple plots can be easily combined, by passing in different axes like:

```python3
fig = plt.figure(**fig_kwargs)
ax1 = fig.add_subplot()
ax2 = fig.add_subplot()
data.plot_something(ax=ax1)
data.plot_something_else(ax=ax2)
```
So I think this we should definitely keep (I haven't check whether this is changed in your PR...). We could maybe check once we have this consistently, but by now I think this is mostly the case. Those methods should always return the axes object (in case it's created inside).

 ### Plotting multiple things
There are some methods, which call multiple `.plot_something()` methods internally (e.g. `SpectrumDataset.plot_fit()`). In this case, I think we should introduce multiple `ax` arguments and corresponding `kwargs` like `ax_spectrum` and `kwargs_spectrum`. 

In this case I  think we should still create the figure outside of the plotting function and if needed just pass it on like.

```python3
fig = plt.figure(**fig_kwargs)
ax_something = fig.add_subplot()
ax_2 = fig.add_subplot()

axes = data.plot_multiple_things(ax_something=ax_something, ax_2=ax=2, kwargs_something={}, kwargs_2={})
```
In this case the keyword forwarding goes to the other plotting function like `.plot_something()` and the method could return a dictionary with the axes.

### Peek methods
I think the `.peek()` methods should not take any arguments at all (except maybe a figure argument or a `GridSpec`). The problem there is that the `.peek()` methods possibly call up to 5 othee plotting functions (e.g. `Observation.peek()`) and any keyword forwarding or configuration will be complicated and confusing as well. Instead we should make sure that the `.peek()` methods are super simple and just create a GridSpec (or subplots) internally and call the other `.plot_something(ax=ax)` and `.plot_more(ax_something=ax_something, ax_2=ax_2)` in order. 

What do you think? BTW: Have you already joined the Gammapy Slack? Then we could have a call next week and discuss this in more depth?

alexiscoutinho:
@adonath, I agree with you in general, but from what I understood there could be some inconsistencies:
- "There are way too many possibilities to support all", does this mean `fig_kwargs` and `ax_kwargs` don't effectively cover all possibilities? If so, then I agree with dropping them. My idea was to make them as generic as possible to satisfy all users.
- "Keyword forwarding to Matplotlib methods", that's pretty much what I did, so what is the issue? Is it because, once again, I tried to make something that covers all practical cases, but since it doesn't, it is better to just not try to do it and settle with making the signature minimalistic?
- "Optional keywords which are just forwarded to the corresponding Matplotlib function, such as `imshow()`, `plot()`, `errrorbar()`, `pcolormesh()` etc", I believe you meant forwarding to the applicable highest level method, like `FluxPoints.plot` and `SpectralModel.plot` in `FluxPointsDataset.plot_residuals`, right?
- "In this case, I think we should introduce multiple `ax` arguments and corresponding `kwargs`", I'm not completely sure this is the best idea as it could unnecessarily "bloat" the signature. The user would still have the possibility to change the `ax` geometry by picking up the retures Axes and could also call the inner `.plot_something()` methods which would definetely support `ax`. I think the solution for this would be to compare all relevant signatures with and without these extra `ax`s.
- I agree that `.peek` methods should not allow customization of individual plots, but I definitely think it should accept at least one Object argument, like `fig`, to satisfy the more object-oriented users (like me). Maybe even the other generic `kwargs` of the latest commit depending on your answers to the first two points.

And finally, I don't have an email with a proper domain for Slack. My other email is just @usp.br, not @ifsc.usp.br. :P

adonath:
@alexiscoutinho I just invited you to the Gammapy Slack, 

Let me try to address your comments:
1. I think we should not use `fig_kwargs` and `ax_kwargs` at all, but just rely on users creating the axes objects and figures outside and passing them on to the Gammapy plot functions. I think this works just as good and gives users the full freedom of configuring the Matplotlib axes.
2. Sorry this was not precise from my side: I think we should use the keyword forwarding everywhere, except for the axes and figures creation, just because there are many ways to create axes, like `subplot()`, `add_axes()`, `gridspec()`, `add_subpplot()` etc. (see also my point above). And I think we should just always rely on users creating the axes outside, which is simpler and gives even more freedom to the users...
3. I proposed two levels of keyword forwarding:
    - The keyword forwarding is simple for the `plot_one_thing()` methods, where we just pass on the keywords to the internally called Matplotlib functions, like `imshow()`, `errrobar()` etc. and document it.
    - In the case of `plot_mutiple_things()` methods, we just pass on the keywords to the different `plot_one_thing()` methods So e.g. in the case of `SpectrumDataset.plot_fit()` we introduce `.plot_fit(ax_counts, ax_residuals, kwargs_counts={}, kwargs_residuals={})`. Do you think this makes sense? (see my point below..)

4. I agree here. This was just my first proposed solution I think most `plot_mutiple_things()` methods internally call at most 3 other `plot_one_thing()` methods, so I think a accepting multiple axes and keyword argument could be an acceptable solution. 
5. I'm totally fine if we add a `fig` argument to our `peek()` methods. This will give at least minimal possibilities for configuration.

alexiscoutinho:
@adonath I updated `spectrum.py` according to our discussion, what do you think? Furthermore, do you prefer if `MapDataset.plot_residuals` was split (following the form of `SpectrumDataset.plot_fit`)? As of now, it is kind of a hybrid between a `.plot_multiple_things` and two `.plot_something`. Its signature would be like the below one if kept as is:
```python
def plot_residuals(
    self,
    ax_spatial=None,
    ax_spectral=None,
    method="diff",
    smooth_kernel="gauss",
    smooth_radius="0.1 deg",
    kwargs_spatial=None,
    region=None,
    kwargs_spectral=None,
):
    """
    Plot spatial and spectral residuals.

    The spectral residuals are extracted from the provided region, and the
    normalization used for the residuals computation can be controlled using
    the method parameter. If no region is passed, only the spatial
    residuals are shown.

    Parameters
    ----------
    ax_spatial : `~astropy.visualization.wcsaxes.WCSAxes`
        Axes to plot spatial residuals on.
    ax_spectral : `~matplotlib.axes.Axes`
        Axes to plot spectral residuals on.
    method : {"diff", "diff/model", "diff/sqrt(model)"}
        Normalization used to compute the residuals, see `MapDataset.residuals`.
    smooth_kernel : {"gauss", "box"}
        Kernel shape.
    smooth_radius: `~astropy.units.Quantity`, str or float
        Smoothing width given as quantity or float. If a float is given, it
        is interpreted as smoothing width in pixels.
    kwargs_spatial : dict
        Keyword arguments passed to `~matplotlib.axes.Axes.imshow`.
    region: `~regions.SkyRegion`
        Target sky region.
    kwargs_spectral : dict
        Keyword arguments passed to `~matplotlib.axes.Axes.errorbar`.

    Returns
    -------
    ax_spatial, ax_spectral : `~astropy.visualization.wcsaxes.WCSAxes`, `~matplotlib.axes.Axes`
        Spatial and spectral residuals plots.
    """
```

gammapy/datasets/spectrum.py
```diff
@@ -336,23 +336,60 @@ def energy_range(self):
...
+        Returns
+        -------
+        ax_spectrum, ax_residuals : `~matplotlib.axes.Axes`
+            Spectrum and residuals plots.
         """
-        from matplotlib.gridspec import GridSpec
         import matplotlib.pyplot as plt
+        from matplotlib.gridspec import GridSpec
+
+        if not ax_spectrum or not ax_residuals:
```
adonath, line 363: I think we could possibly even simplify and require that `ax_spectrum` and `ax_residuals` must be both provided, and just raise an error if this is not the case...

adonath:
Thanks @alexiscoutinho! I think this goes exactly in the right direction. I have left one comment to simplify even more, but otherwise feel free to go ahead!

gammapy/datasets/flux_points.py
```diff
@@ -275,107 +296,102 @@ def plot_residuals(self, ax=None, method="diff", **kwargs):
 
         Parameters
         ----------
-        ax : `~matplotlib.pyplot.Axes`
-            Axes object.
-        method : {"diff", "diff/model", "diff/sqrt(model)"}
-            Method used to compute the residuals, see `MapDataset.residuals()`
+        ax : `~matplotlib.axes.Axes`
+            Axes to plot on.
+        method : {"diff", "diff/model"}
```
bkhelifi, line 301: Why the method "diff/sqrt(model)" disappears in this function? I might have not catched the global logic...

gammapy/datasets/flux_points.py
```diff
@@ -275,107 +296,102 @@ def plot_residuals(self, ax=None, method="diff", **kwargs):
...
-            self.data.e_ref.value, residuals.value, xerr=xerr, yerr=yerr, **kwargs
-        )
+        ax.errorbar(fp.e_ref.value, residuals.value, xerr=xerr, yerr=yerr, **kwargs)
+        ax.axhline(0, color=kwargs["color"], lw=0.5)
 
         # format axes
-        ax.axhline(0, color="black", lw=0.5)
-        ax.set_ylabel("Residuals {}".format(unit.__str__()))
-        ax.set_xlabel(f"Energy ({self._e_unit})")
+        ax.set_xlabel(f"Energy [{self._e_unit}]")
```
bkhelifi, line 341: Which energy is it? reconstructed or true? It might be useful that this is always precised in the plot function, for non-TeV experts...

gammapy/datasets/flux_points.py
```diff
@@ -275,107 +296,102 @@ def plot_residuals(self, ax=None, method="diff", **kwargs):
...
         plot_kwargs.update(fp_kwargs)
         plot_kwargs.setdefault("label", "Flux points")
-        ax = self.data.plot(ax=ax, **plot_kwargs)
+        ax = self.data.plot(ax, **plot_kwargs)
 
         plot_kwargs = kwargs.copy()
-        plot_kwargs.setdefault("energy_range", self._e_range)
-        plot_kwargs.setdefault("zorder", 10)
         plot_kwargs.update(model_kwargs)
+        plot_kwargs.setdefault("energy_range", self._e_range)
```
bkhelifi, line 386: the reco or true energy range? @adonath, as we have discussed in the past, is it not useful to precise it always in the variable names?

gammapy/datasets/flux_points.py
adonath: Of course it's not the case in general, but for the flux points computed with Gammapy it is reconstructed energy. At some point we introduced the convention, that "energy" refers to reconstructed energy, while we make it explicit for true energy (https://docs.gammapy.org/dev/irf/index.html#irf-axis-naming). So the axis label is fine here I think...

gammapy/datasets/flux_points.py
alexiscoutinho: Because an error is raised on line 312/331 for that `yerr` option.

adonath:
Thanks a lot @alexiscoutinho for keeping this PR up to date. To get this in for v0.18 I'll fix the remaining merge conflicts, do a little more clean-up and merge. So please don't add further commits for now...

adonath:
@alexiscoutinho I'm sorry to say I wasn't able to resolve the remaining merge conflicts. The development tempo is so high in Gammapy, that it's hard to keep track. I'll re-label the milestone now and ideally your a available for a short call or are maybe able to join the co-working week https://github.com/gammapy/gammapy-meetings/tree/master/coding-sprints/2020-11-Co-Working-Week, because we are very interested to get the cleanup in...

adonath (approved):
Thanks a lot @alexiscoutinho, excellent work!

PR #3002, opened 2020-09-08
PSFMap from Gaussian 

LauraOlivera:
**Description**

This pull request creates a function `PSFMap.from_gaussian()` to create an all-sky `PSFMap` with Gaussian profiles, which is useful for tests. The shape (sigma) can depend on energy but it does not depend on position. I used existing functions to create a `TablePSF` from a Gaussian profile, extended it to an `EnergyDependentTablePSF` and then finally a `PSFMap`. Note that it would also be straightforward to have a disk profile, as that is already an option in `TablePSF.from_shape()`.

**Dear reviewer**

I also created a test, `test_psfmap_from_gauss()`, to make sure the behavior is the expected one.

### DIFF
--- a/gammapy/irf/psf_map.py
+++ b/gammapy/irf/psf_map.py
@@ -300,7 +300,7 @@ class PSFMap(IRFMap):
         return cls(psf_map=psf_map, exposure_map=exposure_map)
 
     @classmethod
-    def from_gauss(cls,energy, rad, sigma):
+    def from_gauss(cls, energy, rad, sigma):
         """Create all -sky PSF map from Gaussian width.
 
         This is used for testing and examples.
@@ -334,12 +334,12 @@ class PSFMap(IRFMap):
             # one width per energy
             energytable_temp = np.zeros(tableshape)*u.sr**-1
             for i in np.arange(tableshape[0]):
-                energytable_temp[i,:] = TablePSF.from_shape(shape='gauss', width=sigma[i], rad=rad).psf_value
+                energytable_temp[i, :] = TablePSF.from_shape(shape='gauss', width=sigma[i], rad=rad).psf_value
         else:
             raise AssertionError('There need to be the same number of sigma values as energies')
         energytable = EnergyDependentTablePSF(energy, rad, exposure=None, psf_value=energytable_temp)
-        psfmap = cls.from_energy_dependent_table_psf(energytable)
-        return psfmap
+        psf_map = cls.from_energy_dependent_table_psf(energytable)
+        return psf_map
 
     def to_image(self, spectrum=None, keepdims=True):
         """Reduce to a 2-D map after weighing
--- a/gammapy/irf/psf_map.py
+++ b/gammapy/irf/psf_map.py
@@ -301,7 +301,7 @@ class PSFMap(IRFMap):
 
     @classmethod
     def from_gauss(cls,energy, rad, sigma):
-        """Create PSF map from Gaussian width.
+        """Create all -sky PSF map from Gaussian width.
 
         This is used for testing and examples.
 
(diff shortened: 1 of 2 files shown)


gammapy/irf/psf_map.py
```diff
@@ -299,6 +299,48 @@ def from_energy_dependent_table_psf(cls, table_psf):
         exposure_map.quantity += data
         return cls(psf_map=psf_map, exposure_map=exposure_map)
 
+    @classmethod
+    def from_gauss(cls, energy, rad, sigma):
```
adonath, line 303: We could directly use a `MapAxis` object here for `energy` and `rad` (in this case I would rename it to `energy_axis_true` and `rad_axis` to be compatible with `MapDataset.create()`. For "historical" reasons the `MapAxis` has not really been used for IRFs, but as the `.from_gauss()` method is part of the `PSFMap`API (where we rely on `Map` and `MapAxis` objects) this seems more natural...

gammapy/irf/psf_map.py
```diff
@@ -299,6 +299,48 @@ def from_energy_dependent_table_psf(cls, table_psf):
...
+
+        Parameters
+        ----------
+        energy : `~astropy.units.Quantity`
+            Energy (1-dim)
+        rad : `~astropy.units.Quantity` with angle units
+            Offset angle wrt source position (1-dim)
+        sigma : `~astropy.coordinates.Angle`
+            Gaussian width.
+        Returns
```
adonath, line 320: Missing empty line before...

gammapy/irf/psf_map.py
```diff
@@ -299,6 +299,48 @@ def from_energy_dependent_table_psf(cls, table_psf):
...
+        -------
+        psf_map : `PSFMap`
+            Point spread function map.
+        """  
+
+        # note: it would be straightforward to also have disk shape instead
+        # of gauss
+        tableshape = (energy.shape[0], rad.shape[0])
+        if np.size(sigma) == 1:
```
adonath, line 329: Minor: for 1d arrays we typically use the standard Python `len()` expression...

gammapy/irf/psf_map.py
```diff
@@ -299,6 +299,48 @@ def from_energy_dependent_table_psf(cls, table_psf):
...
+            # same width for all energies
+            tablepsf = TablePSF.from_shape(shape='gauss', width=sigma, rad=rad)
+            energytable_temp = np.tile(tablepsf.psf_value, (tableshape[0], 1))
+        elif np.size(sigma) == np.size(energy):
+            # one width per energy
+            energytable_temp = np.zeros(tableshape)*u.sr**-1
+            for i in np.arange(tableshape[0]):
```
adonath, line 336: Again minor: often in Python using `idx` is preferred over a single character like `I`...

gammapy/irf/tests/test_psf_map.py
```diff
@@ -382,3 +382,51 @@ def test_to_image():
...
+    psfmap = PSFMap.from_gauss(energy, rad, sigma)
+    assert psfmap.psf_map.geom.axes[0] == rad_axis
+    assert psfmap.psf_map.geom.axes[1] == energy_axis
+    assert psfmap.psf_map.unit == Unit("sr-1")
+    assert psfmap.psf_map.data.shape == (energy.shape[0], rad.shape[0], 1, 2)
+    assert_allclose(
+        psfmap.get_energy_dependent_table_psf().containment_radius(1*u.TeV)[0], 
+        psfmap.containment_radius_map(1*u.TeV).data[0][0]*u.deg
+    )
+    assert_almost_equal(
```
adonath, line 405: I think using `assert_allclose` here just works as well?

gammapy/irf/tests/test_psf_map.py
```diff
@@ -382,3 +382,51 @@ def test_to_image():
...
+        psfmap.containment_radius_map(energy[3],0.68).data[0][0]/sigma[3].value, 
+         1.5095921854516636, decimal=2
```
adonath, line 407: If you only care about a certain precision for the reference value, you could just give the relevant decimal places...

adonath (changes requested):
Thanks a lot @LauraOlivera! I have left a few inline comments, once those are addressed the PR is ready to merge...

gammapy/irf/psf_map.py
LauraOlivera: I think you're right. I did it this way because the input to the `EnergyDependentTablePSF` were not `MapAxis` but just 1-d arrays. I changed it now.

gammapy/irf/psf_map.py
LauraOlivera: If `sigma` is just a float or an `astropy.Quantity`, `len(sigma)` throws an error. Also  `np.size("0.1deg") = 1`, while` len("0.1deg")=6`.

adonath: Ah I see, makes sense to keep...!

adonath (approved):
Thanks @LauraOlivera, I have no further comments...

PR #3005, opened 2020-09-10
Allow custom spectral models corrections for BackgroundModel

QRemy:
Add spectral_model argument to BackgroundModel instead of norm and tilt arguments in order to allows more refine parametrisation. Default spectral_model is PowerLawNormSpectralModel which is equivalent to previous parametrisation.

### DIFF
--- a/docs/tutorials/analysis_2.ipynb
+++ b/docs/tutorials/analysis_2.ipynb
@@ -261,7 +261,7 @@
     "    # fit background model\n",
     "    dataset = maker_fov.run(dataset)\n",
     "    print(\n",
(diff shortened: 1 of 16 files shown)


QRemy:
The test failures in gammapy/estimators should be solved by updating gammapy-data
https://github.com/gammapy/gammapy-data/pull/2

adonath (dismissed):
Thanks @QRemy, I think you need to adapt the docstring of `BackgroundModel` class. Otherwise looks good to me...

adonath (approved):
Thanks @QRemy, I have no further comments...

PR #3007, opened 2020-09-10
Interpolation in makers map

LauraOlivera:
**Description**

This pull request makes it possible to have a `Map` object as the `aeff` and `bkg` inputs to the data reduction performed by the `MapDatasetMaker`, as well as an `EdispKernelMap` and `PSFMap` for the `edisp` and `psf`. The the input maps are interpolated to the reference geometry before being bundled into the resulting `MapDataset`.

**Dear reviewer**
I included a test that creates a dummy `Map` effective area and background, a dummy `PSFMap`, and `EdispKernelMap` and dummy `GTI` and `Events`. They are bundled into an `Observation` and run through the `MapDatasetMaker` with a given target `geom`.

### DIFF
--- a/gammapy/makers/map.py
+++ b/gammapy/makers/map.py
@@ -13,7 +13,6 @@ from .utils import (
     make_map_background_irf,
     make_map_exposure_true_energy,
     make_psf_map,
-    interpolate_map_IRF,
     interpolate_edisp_kernel_map,
     interpolate_psf_map,
 )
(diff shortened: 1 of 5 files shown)


adonath:
Thanks @LauraOlivera, the test fails on Azure seemed unrelated. I restarted the builds and will do a review later today...

LauraOlivera:
@adonath I moved the map interpolation to the `Map` class and replaced it for each IRFs. This way no extra methods are necessary in `makers/utils.py` . I also added a test for the `Map.interp_to_geom()` method.

adonath (approved):
Thanks a lot @LauraOlivera, I have no further comments..

PR #3008, opened 2020-09-11
Make 'selection' first kwarg in Maker classes

alexiscoutinho:
**Description**
Since `selection` is the only argument that makes sense passing positionally (an average reader could understand its meaning even without the keyword), it would be very convenient to make it the first keyword argument in the signatures. `SafeMaskMaker` already does something similar.

### DIFF
--- a/gammapy/makers/map.py
+++ b/gammapy/makers/map.py
@@ -25,17 +25,17 @@ class MapDatasetMaker(Maker):
 
     Parameters
     ----------
-    background_oversampling : int
-        Background evaluation oversampling factor in energy.
     selection : list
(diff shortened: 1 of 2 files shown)


AtreyeeS:
Yes indeed! I agree.. most users will specify `selection` rather than play around with `background_oversampling`

registerrier:
Yes a more uniform calling scheme between `makers` is definitely a good thing. Thanks!

alexiscoutinho:
Also, should we stick with `selection` or `methods`?

adonath (approved):
Thanks @alexiscoutinho, no further comments...

PR #3009, opened 2020-09-16
Add Dataset.resample_energy_axis method

registerrier:
**Description**

This pull request introduces new functionalities on `MapDataset`, `SpectrumDataset` and their ONOff variants. 
The `group_over_energy` method allows to rebin a `MapDataset` according to a set of reco energy edges. It relies on `Map.resample_axis`. The latter has been updated to support other `ufunc` than `np.add` (e.g. `np.logical_or` for masks). It has also been modified to consider only `normal` energy group and not `underflows` and `overflows` from `MapAxis.group_table`.
The `EDispKernelMap` has also been modified to support the `resample_axis` method.

This should be allow for simpler code for the edges support in the `Estimators`. 

**Dear reviewer**

### DIFF
--- a/gammapy/datasets/map.py
+++ b/gammapy/datasets/map.py
@@ -1317,15 +1317,15 @@ class MapDataset(Dataset):
             if self.__dict__.pop(name, False):
                 log.info(f"Clearing {name} cache for dataset {self.name}")
 
-    def resample_energy_axis(self, e_edges=None, name=None):
-        """Resample MapDataset over reco energy edges.
+    def resample_energy_axis(self, axis=None, name=None):
+        """Resample MapDataset over new reco energy axis.
 
         Counts are summed taking into account safe mask.
 
         Parameters
         ----------
-        e_edges : `~astropy.units.Quantity`
-            the energy edges.
+        axis : `~gammapy.maps.MapAxis`
+            the new reco energy axis.
         name: str
             Name of the new dataset.
 
@@ -1334,11 +1334,10 @@ class MapDataset(Dataset):
         dataset: `MapDataset`
             Resampled dataset .
         """
-        if e_edges is None:
+        if axis is None:
             e_axis = self._geom.get_axis_by_name("energy")
             e_edges = u.Quantity([e_axis.edges[0], e_axis.edges[-1]])
-
-        axis = MapAxis.from_edges(e_edges, name="energy", interp=self._geom.axes[0].interp)
+            axis = MapAxis.from_edges(e_edges, name="energy", interp=self._geom.axes[0].interp)
 
         name = make_name(name)
         kwargs = {}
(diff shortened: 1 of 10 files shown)


gammapy/datasets/map.py
```diff
@@ -1360,6 +1311,79 @@ def reset_data_cache(self):
...
+        Returns
+        -------
+        dataset: `MapDataset`
+            Grouped dataset .
+        """
+        if e_edges is None:
+            e_axis = self._geom.get_axis_by_name("energy")
+            e_edges = u.Quantity([e_axis.edges[0], e_axis.edges[-1]])
+
+        axis = MapAxis.from_edges(e_edges, name="energy")
```
adonath, line 1335: I know that we have the `e_edges` argument elsewhere in Gammapy, but by now I think it would be the better option to just directly take a `MapAxis` object, as the first thing we do in many places is to create a `MapAxis`...

gammapy/datasets/map.py
```diff
@@ -1360,6 +1311,79 @@ def reset_data_cache(self):
             if self.__dict__.pop(name, False):
                 log.info(f"Clearing {name} cache for dataset {self.name}")
 
+    def group_over_energy(self, e_edges=None, name=None):
```
adonath, line 1314: Minor: for consistency with the Map API I think I would prefer the name `resample_energy_axis()`...

gammapy/datasets/map.py
```diff
@@ -1360,6 +1311,79 @@ def reset_data_cache(self):
...
+
+        if self.mask_safe is not None:
+            weights = self.mask_safe
+            kwargs["mask_safe"] = self.mask_safe.resample_axis(axis=axis, ufunc=np.logical_or)
+        else:
+            weights = None
+
+        if self.counts is not None:
+            kwargs["counts"] = self.counts.resample_axis(axis=axis, weights=weights)
+            print(kwargs["counts"].geom.axes[0].edges)
```
adonath, line 1352: Remove print statement...

adonath (changes requested):
Thanks @registerrier, I have left a few minor inline comments. Otherwise looks good to me. There are still some Travis-CI fails you need to address..

gammapy/datasets/map.py
registerrier: Done

gammapy/datasets/map.py
registerrier: Done

gammapy/datasets/map.py
registerrier: Indeed passing a `MapAxis` directly might be more relevant. 
Note that there might be some confusion with `downsample(axis=None)` which expects an axis name rather than an axis per-se.

adonath: Yes, good point! Maybe we should re-name those argument to `axis_name` uniformly...maybe open a reminder issue?

PR #3011, opened 2020-09-16
Remove parameters in TemplateSpectralModel

QRemy:
Remove parameters in TemplateSpectralModel. Spectral corrections should be applied via a compound model with one of the NormSpectralModel. Adding example on how to do this in a tutorial could be done in a separate PR.


adonath (changes requested):
Thanks @QRemy, this looks basically good to me. However I think we should at least add one test, that corresponds to the previous behaviour. I.e. implement at least one test case, where the `TemplateSpectralModel` is multiplied with a `PowerLawNormSpectralModel` and evaluated...

adonath:
@QRemy There are remaining test fails related to the plotting of the `TemplateSpectralModel`, maybe you can check: https://travis-ci.org/github/gammapy/gammapy/jobs/727952927#L7033

adonath (approved):
Thanks @QRemy, this looks good to me now. For some reason there are unrelated files in this PR in the `pip-wheel-meta` folder. Could you remove those and add the meta folder to `.gitignore` instead?

adonath (approved):
Thanks @QRemy, I have no further comments...

PR #3012, opened 2020-09-17
Replace SkyDiffuseCube by TemplateSpatialModel

QRemy:
This PR implement changes to replace SkyDiffuseCube by SkyModel with 3D TemplateSpatialModel:
- remove SkyDiffuseCube
- fix the SkyModel evaluation with 3D TemplateSpatialModel
- add optional argument unit in TemplateSpatialModel.read() for convenience
- adapt test data (PR in gammapy-data to be merged)
- adapt notebooks

### DIFF
--- a/docs/tutorials/event_sampling.ipynb
+++ b/docs/tutorials/event_sampling.ipynb
@@ -460,7 +460,7 @@
    "outputs": [],
    "source": [
     "diffuse_cube = TemplateSpatialModel.read(\n",
-    "    \"$GAMMAPY_DATA/fermi-3fhl-gc/gll_iem_v06_gc.fits.gz\", normalize=False, unit = \"cm-2 s-1 MeV-1 sr-1\"\n",
+    "    \"$GAMMAPY_DATA/fermi-3fhl-gc/gll_iem_v06_gc.fits.gz\", normalize=False\n",
     ")\n",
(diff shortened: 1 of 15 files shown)


adonath:
Already one comment: somewhere we need to put a big red box (either docs/modeling/models) or the model gallery, to really make clear that the `TemplateSpatialModel` is the only case, where the flux amplitude can be stored in the spatial part of the factorised model.

adonath:
I checked and the data unit is not correctly set when using `format="fgst-template"`. The reason is that the format is used for the exposure cube as well, which has a different unit. It is even worse than that, because just from the header information alone once cannot distinguish, whether the FITS file contains the exposure or the sky diffuse template (such a bad idea...meta data is so important)! By default the exposure cube written by the Fermi science tools contains a GTI hdu as well. This information we could use...

For the unit handling my proposal would be the following: we just stick with the current scheme, that if no unit is defined we assume it's `sr-1` if a unit is defined with just take this one. 

In addition we modify the `Map.read()` code to identify the `fgst-template` format correctly and set the unit of the data accordingly. Any file written with Gammapy should contain the unit of the data.

adonath:
@QRemy In #3019 I fixed the issue with the unit handling of the Fermi FGST FITS files. Can you check whether this resolves the need for the special unit handling for the `TemplateSpatialModel`?

adonath (approved):
Thanks, @QRemy! Looks good to me...

PR #3020, opened 2020-09-22
Add region parameter to MapDataset's energy_range

alexiscoutinho:
**Description**
This PR simply adapts the `_energy_axis` property and `energy_range` method from `SpectrumDataset` for `MapDataset`s. `energy_range`, therefore, supports regions for safe energy mask extraction.

**Dear reviewer**
I'm really not sure how to extend these to the `MapDatasetOnOff` class, any help?


adonath (changes requested):
Thanks a lot @alexiscoutinho! I left two minor inline comments. The current implementation is already useful, so I think we can merge the PR, once you added a test for this functionality.

 In general I think we even need a bit more. I added you to the #dev channel in the Gammapy Slack, where we started to discuss a bit which methods we want for the energy range estimation. I'll also talk to @registerrier about it again...

gammapy/datasets/map.py
```diff
@@ -570,6 +570,49 @@ def stat_array(self):
         """Likelihood per bin given the current model parameters"""
         return cash(n_on=self.counts.data, mu_on=self.npred().data)
 
+    @property
+    def _energy_axis(self):
```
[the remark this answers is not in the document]
alexiscoutinho: Does that mean `SpectrumDataset._energy_axis` should be dropped as well?

alexiscoutinho:
@adonath Since `.energy_range`, regardless of method, revolves around `RegionNDMap`, I thought it would be better to move the bulk code of these methods there. Probably in an "internal" (`_`) namespace because the method would only make sense if the dtype was `bool`.

gammapy/datasets/map.py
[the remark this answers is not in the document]
adonath: No, for the `SpectrumDataset` the information on the `_energy_axis` is used, while for the `MapDataset` the equivalent thing would be `._geom`. One could possibly change to `._geom` for the `SpectrumDataset` as well, but that's mostly a detail...

adonath:
Thanks @alexiscoutinho, we haven't forgot this...but I think we cannot get it in for v0.18. I'll re-label now and the we get back to it after v0.18...

adonath:
@alexiscoutinho Are you still interested in working on this, or OK to close?

alexiscoutinho:
@adonath Have the methods been decided though? I have lost track of the discussions. :|

alexiscoutinho:
In the meantime, I think it's nice to put this feature to use as is.

gammapy/datasets/spectrum.py
```diff
@@ -69,6 +69,11 @@ def plot_fit(
     def _energy_unit(self):
         return self._geom.axes[0].unit
 
+    @property
+    def energy_range(self):
+        """Energy range defined by the safe mask"""
+        return super().energy_range()
```
alexiscoutinho, line 75: I'm not 100% sure calling super() like this is ideal.

gammapy/datasets/map.py
```diff
@@ -337,20 +337,35 @@ def data_shape(self):
         """Shape of the counts or background data (tuple)"""
         return self._geom.data_shape
 
-    def _energy_range(self, mask = None):
-        """Compute the energy range with or without the mask fit."""
+    # TODO: make this support different methods?
```
bkhelifi, line 340: I do not understand the meaning of this ToDo? Which kind of different methods?

adonath: I think the idea here was to maybe support different methods like `.energy_range(which="safe")`. This is bit more flexible than introducing multiple attributes like `.energy_range_safe` and `.energy_range_fit`...

alexiscoutinho: Since I had to adapt this comment a long time ago, it remains as a change and is carried in diffs to this date even though it was already removed from master.

adonath:
@alexiscoutinho Are you interested to implement the proposal in https://github.com/gammapy/gammapy/issues/3427#issuecomment-894755805 instead?

alexiscoutinho:
@adonath I'll try.

PR #3022, opened 2020-09-23
Modify MapDataset.downsample options and add MapDatasetOnOff.downsample

registerrier:
**Description**

This pull request modifies the `downsample` API for the `MapDataset`. For clarity, the `axis` argument becomes `axis_name` to make clear a string is expected. 
The `MapDatasetOnOff.downsample` method is also implemented (it used to `raise NotImplementedError`). 
Tests are added for both methods.

Technically, it is obvious that `downsample` is to be applied to the `energy` axis. So that a boolean flag, might be enough (e.g. `energy_downsample = False` by default). 

**Dear reviewer**

Some possible further clean-up:
- Modify `Map.downsample` to take an `axis_name` as argument as well.
- Add a `downsample` method on `SpectrumDataset`.

### DIFF
--- a/gammapy/datasets/map.py
+++ b/gammapy/datasets/map.py
@@ -1179,7 +1179,7 @@ class MapDataset(Dataset):
 
         if self.counts is not None:
             kwargs["counts"] = self.counts.downsample(
-                factor=factor, preserve_counts=True, axis=axis_name, weights=self.mask_safe
+                factor=factor, preserve_counts=True, axis_name=axis_name, weights=self.mask_safe
             )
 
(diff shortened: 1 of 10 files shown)


adonath (approved):
Thanks @registerrier, I don't have any comments, except that we should also do the rename for the following functions as well:
- `Geom.upsample()` and `Geom.downsample()`
- `Map.upsample()`
But this could be done in a follow up PR as well.

adonath (approved):
Thanks @registerrier! I have no further comments...

PR #3023, opened 2020-09-25
Clean up MapEvaluator 

adonath:
**Description**

This pull request cleans up the `MapEvaluator` a bit, especially tries to simplify the code for caching, so that it becomes easier to maintain. The idea was to use `lru_cache()` as well, so that we have a unified handle on caching in Gammapy,

### DIFF
--- a/gammapy/datasets/map.py
+++ b/gammapy/datasets/map.py
@@ -2142,6 +2142,24 @@ class MapEvaluator:
         self._cached_parameter_values = None
         self._cached_parameter_values_spatial = None
 
+    # workaround for the lru_cache pickle issue
+    # see e.g. https://github.com/cloudpipe/cloudpickle/issues/178
+    def __getstate__(self):
+        state = self.__dict__.copy()
+        for key, value in state.items():
+            func = getattr(value, "__wrapped__", None)
+            if func is not None:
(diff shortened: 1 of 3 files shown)


adonath:
@QRemy Do you need the possibility to pickle a `MapEvaluator` object?

QRemy:
The possibility to pickle a MapEvaluator object is needed to use multiprocessing and by extension eemc with more than one core.  For example in notebook mcmc_sampling.ipynb if ones change the default for `threads` as:
`sampler = run_mcmc(dataset, nwalkers=6, nrun=150, threads=4)`
you get the following error 
`PicklingError: Can't pickle <functools._lru_cache_wrapper object at 0x1c198b4d58>: it's not the same object as gammapy.datasets.map.MapEvaluator._compute_npred`
while this do not append in master. 
Did you change something about the cache in Map ? because I would have expected to fail also because of it.

QRemy (dismissed):
Thanks @adonath, I have no further comment besides the potential problem of pickle with lru_cache(). I checked it with 3FHL validation (--selection debug), and in the simple region tested computation time is comparable with master if USE_NPRED_CACHE = True and increase by a factor of about 4 if False, as previously.

Issue #2253, opened 2019-06-21
Implement an interactive parameter editing


facero:
Following today's discussion at the gammapy dev. I managed to get a proof of concept of editing the model parameters interactively in the notebook with [ipysheet](https://ipysheet.readthedocs.io/). It can be very useful when analysing a region with multiple sources. The nice thing is that you can sort by frozen/free param which is very useful.

A proof of concept minimal working example is here:

```
from gammapy.spectrum.models import PowerLaw
from gammapy.image.models import SkyPointSource
from gammapy.cube.models import SkyModel
from gammapy.utils.fitting import Parameters

def edit_interactive(sky_model):
    
    import pandas as pd
    from ipysheet import from_dataframe, to_dataframe
    
    model_dict = sky_model.parameters.to_dict()
    df_model=pd.DataFrame(model_dict['parameters'])

    sheet = from_dataframe(df_model)
    display(sheet)
    
    return sheet
    
def sheet_to_model(sheet):
    dict_sheet = to_dataframe(sheet).to_dict()
    model_dict_=[]
    for i in range(sheet.rows):
        dict_={}
        for k in dict_sheet.keys():
            dict_[k]=dict_sheet[k][str(i)]
        model_dict_.append(dict_)

    model_dict={'parameters':model_dict_,'covariance': None}

    sky_model.parameters = Parameters.from_dict(model_dict)
    
    
    return sky_model


spatial_model  = SkyPointSource(lon_0="50 deg", lat_0="40 deg", frame="icrs") 
spectral_model = PowerLaw(index=3, amplitude="1e-11 cm-2 s-1 TeV-1", reference="1 TeV")
sky_model = SkyModel(spatial_model=spatial_model, spectral_model=spectral_model)

#You can sort frozen/free param by clicking the header of any column
#Edit the value param, not the scale or factor
sheet = edit_interactive(sky_model)

#then in a separate cell
sky_model = sheet_to_model(sheet)
print(sky_model.parameters['lon_0'].value)

```

It's still quite some work to have a fully functional implementation but I leave this here for futur reference if we want to implement it one day and for you to try.
Due to the formatting of small numbers (e.g. 1e-11) shows 0 for amplitude but I've seen that this can be changed. Also cells can be read-only, colored, etc

adonath:
[...]

PR #3029, opened 2020-09-28
Helper functions for the interactive parameter editing

facero:
**Description**
This PR is a follow-up of #2253, to implement an interactive editing GUI for the list of all parameters in a Model. This is done using the package [qgrid](https://github.com/quantopian/qgrid) and it works in Jupyter Notebook and Jupyter Lab (not tested). It allows to visualize, sort, filter parameters and update the Model. 

This is useful for the analysis of a given source with complex models.
This PR implements only helper functions, the dependencies (qgrid, pandas) are not in the core gammapy and are needed only when the user calls qgrid.
The idea is to have this as a How-To in the documentation.

**Dear reviewer**
Is there need for tests for this or more documentation ?
We can change the name of the functions if needed.


gammapy/modeling/models/core.py
```diff
@@ -361,6 +362,46 @@ def to_dict(self):
         else:
             return {"components": models_data}
 
+    def to_parameters_table(self):
+        """Convert Models parameters to an astropy Table."""
+        pars = self.parameters.to_table()
+        rows = []
+        for name in self.parameters_unique_names:
+            row = {}
+            #Warning: splitting of parameters will break is source name has a "." in its name.
```
adonath, line 371: Even for `BackgroundModel` we now have the `spectral` prefix in the parameter name. So you can actually rely on the order of the string components in the unique name like `model.type.name`. However we should add a check and disallow parameter names with a `.` in it...this is a change that I can do.

gammapy/modeling/models/core.py
```diff
@@ -361,6 +362,46 @@ def to_dict(self):
...
+        return  table
+
+    def update_from_table(self, t):
+        """Update Models from an astropy Table."""
+        self._check_non_editable_columns(t, self._table_cached)
+        parameters_dict = [dict(zip( t.colnames, row)) for row in t]
+        for k, data in enumerate(parameters_dict):
+            self._update_from_dict(self.parameters[k], data)
+
+    def _update_from_dict(self, parameter, data):
```
adonath, line 390: As this is a private function anyway you could directly rely on `Parameter.__dict__.update()` and remove this function...

gammapy/modeling/models/core.py
```diff
@@ -361,6 +362,46 @@ def to_dict(self):
...
+                row["model"], row["type"], _ = parts
+            elif len(parts) == 2:
+                row["model"], row["type"] = parts[0], ""
+            rows.append(row)
+        table = table_from_row_data(rows)
+        table = hstack([table, pars])
+        self._table_cached = table
+        return  table
+
+    def update_from_table(self, t):
```
adonath, line 383: As you rely on the order of the parameters here, maybe we could just move this this to `Parameters.set_parameters_from_table()` or similar? Because we already have the API to update parameters like `Parameters.set_parameters_factors()`

adonath (changes requested):
Thanks @fabiopintore, I have left a a few comments. Please also add a unit test somewhere calling this functionality.

fabiopintore:
I suppose I've been cited by mistake :)

adonath:
@fabiopintore Ooops, indeed. Sorry for the noise...:-)

gammapy/modeling/models/core.py
facero: I'm back having a look at this unfinished PR. I tried to update the parameters using  `__dict__.update` but it doesn't seem to update the parameters in the end. 
```
    for k, data in enumerate(parameters_dict):
        if k ==0 : print(self.parameters[k], data)
#        self._update_from_dict(self.parameters[k], data) # previous way
        self.parameters[k].__dict__.update(data) # new. is this ok? doesn't crash bu doesn't update the values
```

facero: Fixed adding `update_from_dict` to parameters in PR #3137

gammapy/modeling/models/core.py
facero: Changed name to `set_parameters_from_table` but still in `models/core.py`

gammapy/modeling/models/core.py
facero: there was an issue before of length 2 or 3 when `BackgroundModel` or other param. This is now simplified.
Avoiding a `.` in the name of the source should be added to prevent this from braking.

gammapy/modeling/models/core.py
facero: added tests. 
Passing on y local machine

gammapy/modeling/models/core.py
facero: I thing this is fixed (except the fact to add a check to prevent `.` in names.

gammapy/modeling/models/core.py
```diff
@@ -372,6 +373,35 @@ def to_dict(self, full_output=False):
         else:
             return {"components": models_data}
 
+    def to_parameters_table(self):
+        """Convert Models parameters to an astropy Table."""
+        pars = self.parameters.to_table()
+        rows = []
+        #Warning: splitting of parameters will break is source name has a "." in its name.
+        for name in self.parameters_unique_names:
```
adonath, line 381: By now the `Parameter` object actually has a `.type` attribute which returns `spectral`, `spatial` or `temporal`. So you could actually modify `Parameters.to_table()` to include this information (not `Parameter.to_dict()`, because we use it for serialisation as well...). Then in the `.to_parameters_table()` you could just add the model name information...

gammapy/modeling/models/core.py
```diff
@@ -372,6 +373,35 @@ def to_dict(self, full_output=False):
...
+        table = hstack([table, pars])
+        self._table_cached = table
+        return  table
+
+    def set_parameters_from_table(self, t):
+        """Update Models from an astropy Table."""
+#        self._check_non_editable_columns(t, self._table_cached)
+        parameters_dict = [dict(zip( t.colnames, row)) for row in t]
+        #Astropy Table has masked values for NaN. Replacing with np.nan in dict.
+        for k, data in enumerate(parameters_dict):
```
adonath, line 396: Can you check, whether this part could be moved to the attribute setters of the `Parameter` object? E.g cast to `bool` in https://github.com/gammapy/gammapy/blob/master/gammapy/modeling/parameter.py#L204...

adonath (commented):
Thanks @facero I have left two small comment with suggestion to simplify the implementation...

gammapy/modeling/models/core.py
facero: I managed to move the test for `frozen` parameter being a bool and the replacement of np masked values in the setter.

gammapy/modeling/models/core.py
facero: Done. 
`to_table()` now includes the model type and `to_parameters_table()` is much shorter and simpler to read.

facero:
I think I manage to implement all the feedback. The two main functions `to_parameters_table()` and `set_parameters_from_table()` are now short and concise. The checks have been distributed to the setters().
The code is now much easier to read than at the beggining of the PR. Thanks @adonath  !

gammapy/modeling/models/core.py
```diff
@@ -428,6 +428,21 @@ def to_dict(self, full_output=False):
...
+    def to_parameters_table(self):
+        """Convert Models parameters to an astropy Table."""
+        table = self.parameters.to_table()
+        #Warning: splitting of parameters will break is source name has a "." in its name.
+        model_name = [name.split(".")[0] for name in self.parameters_unique_names]
+        table.add_column(model_name, name='model', index=0)
+        self._table_cached = table
+        return  table
+
+    def set_parameters_from_table(self, t):
```
adonath, line 440: One last minor thing: maybe rename here to `.update_parameters_from_table()`, the we have a uniform `.update_from_something()` pattern...

adonath (changes requested):
Thanks @facero, I have fixed the merge conflict and have one last minor suggestion. Otherwise this looks good to me!

gammapy/modeling/models/core.py
facero: done

adonath (approved):
Thanks @facero, remaining CI fails are unrelated...I'll merge now.

registerrier:
Thanks @facero .
Should we close #2253 ? Or should we wait to implement some documentation on how to use the new feature?

AtreyeeS:
Hi @facero : Do we (or you personally somewhere) have a tutorial on how to use this? This is not linked in the HowTo at present...

adonath:
@AtreyeeS There is an open PR for Gammapy recipes: https://github.com/gammapy/gammapy-recipes/pull/3

AtreyeeS:
thanks @adonath

facero:
Yes this was moved to the recipes section. At the time it was working only with the dev version. It should work fine now with the released version. @AtreyeeS do you want to link the HowTo to the recipes ?
(I should try again to finish the PR now with the latest version).
@Bultako any recent changes in the recipes website ?

AtreyeeS:
I think it is something very useful, specially when dealing with a large number of models, so we should advertise it.
I do not know if `HowTo` is the ideal place (due to difficulties in maintaining version compatibility with the recipes). We can discuss this @registerrier @adonath @bkhelifi @QRemy .

PR #3030, opened 2020-09-29
Fix inconsistency between EventList.stack(), GTI.stack() and rest of stack methods.

LauraOlivera:
**Description**

This pull request fixes the existing inconsistency with the behavior of `EventList.stack()`, renaming it to `EventList.from_stack()`, which takes as an input a list of `EventList` objects. It also creates a new method` eventlist.stack(other)` to append a single `EventList` to an existing one.

### DIFF
--- a/gammapy/data/event_list.py
+++ b/gammapy/data/event_list.py
@@ -93,6 +93,23 @@ class EventListBase:
         stacked_table = vstack_tables(tables, **kwargs)
         return cls(stacked_table)
 
+    def stack(self, other):
+        """Stack with another EventList.
+        
+        Calls `~astropy.table.vstack`.
+
+        Parameters
+        ----------
+        other : `~gammapy.data.EventList`
+            Event list to stack to self
+
+        Returns
+        -------
+        new_event_list : `~gammapy.data.EventList`
+            New event list
+        """
+        return self.__class__(vstack([self.table, other.table]))
+
     def __str__(self):
         ss = (
             "EventList info:\n"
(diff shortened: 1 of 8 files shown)


gammapy/data/event_list.py
```diff
@@ -93,6 +93,23 @@ def stack(cls, event_lists, **kwargs):
...
+        ----------
+        other : `~gammapy.data.EventList`
+            Event list to stack to self
+
+        Returns
+        -------
+        new_event_list : `~gammapy.data.EventList`
+            New event list
+        """
+        return self.__class__(vstack_tables([self.table, other.table]))
```
adonath, line 111: On detail here: for the other classes we have such as `Map` or `Datasets`  the `.stack()` method work "in-place", this means it does not return anything but just modifies the existing data member. In this case you could just overwrite `self.table` and not return anything...

adonath (changes requested):
Thanks @LauraOlivera, I have left one inline comment otherwise looks good to me!

gammapy/data/event_list.py
LauraOlivera: done. Initially I did it not in place because that is the way it is in `GTI.stack()`...

adonath: Good to know! I'll check once again, but likely we are then going to change it for `GTI` as well...

LauraOlivera: I can do that too, if you want

adonath: Yes, please go ahead. Thanks!

adonath (approved):
Thanks a lot @LauraOlivera, I have no further comments...

PR #3037, opened 2020-10-02
Add generalized gaussian model

### LINKED ISSUE #2878: 2D Sersic implementation and custom models
As discussed at the previous developer meeting, people that used to work with 2D Sersic profile in Sherpa would like to have this spatial model in gammapy.
I do have a working solution for this using astropy models and a custom gammapy model. However in this implementation the normalization over the sky is computed over a grid which is not standard for gammapy models (I use this because it works for any complex composite models for which the normalisation is not obvious).

I don't think it's a priority to have all requested models in gammapy, but the custom models class may help to simplify the definition of custom models (compatible with serialization).

Here is the code that was working on v0.16,:

```python
class CustomSpatialModel(SpatialModel):
    r"""Generate spatial model from custom function

    Parameters
    ----------
    lon_0 : `~astropy.coordinates.Longitude`
        :math:`\text{lon}_0`
    lat_0 : `~astropy.coordinates.Latitude`
        :math:`\text{lat}_0`
    """
    tag="CustomSpatialModel"
    def __init__(self, fct, args, frame="galactic", normalize=True, reval=5.0, dr=0.02):
        for arg in args:
            name = str(arg)
            if name == "lon_0":
                lon_0 = Longitude(args[name]).wrap_at("180d")
                setattr(self, name, Parameter(name, lon_0, min=-180, max=180))
            elif name == "lat_0":
                lat_0 = Latitude(args[name])
                setattr(self, name, Parameter(name, lat_0, min=-90, max=90))
            else:
                setattr(self, name, Parameter(name=name, factor=args[name]))
        self.fct = fct
        self.normalize = normalize
        self.reval = reval
        self.dr = dr
        self.frame = frame
        super()._init_from_parameters([getattr(self, str(arg)) for arg in args])
        
    def to_dict(self):
        return {
                "type":self.tag,
                "function": self.fct.__name__,
                "options":{"normalize": bool(self.normalize),"reval": float(self.reval),"dr": float(self.dr)},
                "parameters": self.parameters.to_dict()["parameters"]
                }

    @classmethod
    def from_dict(cls, data):
        fct=globals().get(data["function"])
        normalize = data["options"].get("normalize", True)
        reval = data["options"].get("reval", 5.0)
        dr = data["options"].get("dr", 0.02)
        args = {x["name"].split("@")[0]: x["value"] * u.Unit(x["unit"]) for x in data["parameters"]}
        init = cls(fct, args, normalize=normalize, reval=reval, dr=dr)
        init.parameters.update_from_dict(data)
        for parameter in init.parameters:
            setattr(init, parameter.name, parameter)
        return init

    def _compute_norm(self, **kwargs):
        theta = np.arange(-self.reval, self.reval, self.dr)
        lon_grid, lat_grid = np.meshgrid(theta, theta)
        grid = self.fct(lon_grid*u.deg, lat_grid*u.deg, **kwargs)
        return (180.0 / np.pi / self.dr) ** 2 / grid.sum()

    @property
    def evaluation_radius(self):
        r"""Evaluation radius (`~astropy.coordinates.Angle`).

        Set as :math:`5\sigma`.
        """
        return Angle(self.reval * u.deg)

    def evaluate(self, lon, lat, **kwargs):
        """Evaluate the model"""
        grid = self.fct(lon, lat, **kwargs)
        if self.normalize == True:
            grid *= self._compute_norm(**kwargs)
            return u.Quantity(grid.value, u.Unit("sr-1"), copy=False)
        else:
            return  grid

def SersicSpatialModel(lon, lat, lon_0, lat_0, r_eff,n, ellip, theta):
    from astropy.modeling.functional_models import Sersic2D
    from astropy.units.quantity import Quantity
    amplitude=1.
    if lon_0>180*u.deg:
        lon_0-=360*u.deg
    if lon_0<-180*u.deg:
        lon_0+=360*u.deg
    lon[lon>180*u.deg]-=360*u.deg
    lon[lon<-180*u.deg]+=360*u.deg
    if isinstance(n,Quantity):
        n=n.value
    return Sersic2D.evaluate(lon,lat,amplitude,r_eff,n,lon_0,lat_0,ellip,theta)

from gammapy.modeling.models import SPATIAL_MODELS
SPATIAL_MODELS.append(CustomSpatialModel)

params= {"function":SersicSpatialModel.__name__,
         "options":
         {"normalize":True,
         "reval":5.*r0,
         "dr":0.04},
         "parameters":[
         {'name':"lon_0",'value':l0,'min':l0-r0,'max':l0+r0,"unit":u.deg},
         {'name':"lat_0",'value':b0,'min':b0-r0,'max':b0+r0,"unit":u.deg},
         {'name':"r_eff",'value':r0+0.01,'min':0.01,'max':2*r0,"unit":u.deg},
         {'name':"n",'value':0.5,'min':0.01,'max':10.,"unit":"","frozen":False},
         {'name':"ellip",'value':0.,'min':0.,'max':1.,"unit":"","frozen":True},
         {'name':"theta",'value':0.,'min':0, "max":np.pi,"unit":u.rad,"frozen":True}]}
spatial_model=CustomSpatialModel.from_dict(params)
```

Similarly for spectral models ones can use something like that (this was compatible with an earlier version before the SmoothBrokenPowerLaw was added to gammapy):
```python
class CustomSpectralModel(SpectralModel):
    r"""Generate spectral model from custom function

    Parameters
    ----------
    """
    tag="CustomSpectralModel"
    def __init__(self, fct, args):
        for arg in args:
            name = str(arg)
            setattr(self, name, Parameter(name=name, factor=args[name]))
        self.fct = fct
        super().__init__([getattr(self, str(arg)) for arg in args])

    def to_dict(self):
        return {
                "type": self.tag,
                "function": self.fct.__name__,
                "parameters": self.parameters.to_dict()["parameters"]
                }
            
    @classmethod
    def from_dict(cls,data):
        fct=globals().get(data["function"])
        args = {x["name"].split("@")[0]: x["value"] * u.Unit(x["unit"]) for x in data["parameters"]}
        init = cls(fct, args)
        init.parameters = Parameters.from_dict(data)
        for parameter in init.parameters.parameters:
            setattr(init, parameter.name, parameter)
        return init

    def evaluate(self, energy, **kwargs):
        """Evaluate the model"""
        return self.fct(energy,**kwargs)

from gammapy.modeling.models import SPECTRAL_MODELS
SPECTRAL_MODELS.append(CustomSpectralModel)

def SBPL(energy, amplitude, E_break, alpha_1, alpha_2, delta=0.1):
    from astropy.modeling.powerlaws import SmoothlyBrokenPowerLaw1D as fct
    return fct.evaluate(energy, amplitude, E_break, alpha_1, alpha_2, delta)

params= {"function":SBPL.__name__,
         "parameters":
        [{'name':"amplitude",'value':amp,'min':0,"unit":u.Unit("cm-2 s-1 TeV-1")},
         {'name':"E_break",'value':1.,'min':0.,"unit":u.TeV},
         {'name':"alpha_1",'value':2.,"unit":""},
         {'name':"alpha_2",'value':2.,"unit":""}]}
spectral_model = CustomSpectralModel.from_dict(params)
```

QRemy:
Add generalized gaussian to spatial models.
Resolve #2878 using similar code than in astropy, but with a different parametrisation and an approximated analytical norm.

### DIFF
--- a/examples/models/spatial/plot_gen_gauss.py
+++ b/examples/models/spatial/plot_gen_gauss.py
@@ -16,8 +16,8 @@ the normalization is expressed as:
 .. math::
     N = \frac{1}{ 2 \pi (1-e) r_{\rm eff}^2 \eta \Gamma(2\eta)}\,,
 where :math:`\Gamma` is the gamma function.
-This analytical norm is approximated so it may not integrate to unity in extremal cases 
-if ellipticity tend to one and radius is large  or :math:`\eta` much larger than one (outside the default range).
+This analytical norm is approximated so it may not integrate to unity in extremal cases
+if ellipticity tend to one and radius is large or :math:`\eta` much larger than one (outside the default range).
 """
 
 # %%
--- a/examples/models/spatial/plot_gen_gauss.py
+++ b/examples/models/spatial/plot_gen_gauss.py
@@ -1,0 +1,84 @@
+r"""
+.. _generalized-gaussian-spatial-model:
+
+Generalized Gaussian Spatial Model
+==================================
+
+This is a spatial model parametrising a generalized Gaussian function.
+
+By default, the Generalized Gaussian is defined as :
+
+.. math::
+    \phi(\text{lon}, \text{lat})  = \phi(\text{r}) = N \times \exp \left[ - \left( \frac{r}{r_{\rm eff}} \right)^ \left( 1/\eta \right) \right] \,,
+
+the normalization is expressed as:
+
+.. math::
+    N = \frac{1}{ \pi (1-e) r_{\rm eff}^2 2^\eta \left[ \eta \Gamma(\eta) \right]^3}\,,
+where :math:`\Gamma` is the gamma function.
+This analytical norm is approximated so it may not integrate to unity in extremal cases,
+for example if ellipticity tend to one, or if :math:`\eta` is defined outside the default range.
+
+"""
+
+# %%
+# Example plot
+# ------------
+# Here is an example plot of the model for different shape parameter:
+
+from gammapy.maps import WcsGeom, Map
+from gammapy.modeling.models import (
+    GeneralizedGaussianSpatialModel,
+    Models,
+    PowerLawSpectralModel,
+    SkyModel,
+)
+from astropy import units as u
+import matplotlib.pyplot as plt
+
+lon_0 = 20
+lat_0 = 0
+reval = 3
+dr = 0.02
+geom = WcsGeom.create(
+    skydir=(lon_0, lat_0), binsz=dr, width=(2 * reval, 2 * reval), frame="galactic",
+)
+
+tags = [r"Disk, $\eta=0.01$", r"Gaussian, $\eta=0.5$", r"Laplacian, $\eta=1$"]
+eta_range = [0.01, 0.5, 1]
+r_eff = 1
+e = 0.5
+phi = 45 * u.deg
+fig, axes = plt.subplots(1, 3, figsize=(9, 6))
+for k, eta in enumerate(eta_range):
+    model = GeneralizedGaussianSpatialModel(
+        lon_0=lon_0 * u.deg,
+        lat_0=lat_0 * u.deg,
+        eta=eta,
+        r_eff=r_eff * u.deg,
+        e=e,
+        phi=phi,
+        frame="galactic",
+    )
+    meval = model.evaluate_geom(geom)
+    ax = axes[k]
+    Map.from_geom(geom=geom, data=meval.value, unit=meval.unit).plot(ax=ax)
+    pixreg = model.to_region().to_pixel(geom.wcs)
+    pixreg.plot(ax=ax, edgecolor="g", facecolor="none", lw=2)
+    ax.set_title(tags[k])
+    ax.set_xticks([])
+    ax.set_yticks([])
+plt.tight_layout()
+
+# %%
+# YAML representation
+# -------------------
+# Here is an example YAML file using the model:
+
+pwl = PowerLawSpectralModel()
+gengauss = GeneralizedGaussianSpatialModel()
+
+model = SkyModel(spectral_model=pwl, spatial_model=gengauss, name="pwl-gengauss-model")
+models = Models([model])
+
+print(models.to_yaml())
--- a/examples/models/spatial/plot_gen_gauss.py
+++ b/examples/models/spatial/plot_gen_gauss.py
@@ -17,8 +17,7 @@ the normalization is expressed as:
     N = \frac{1}{ 2 \pi (1-e) r_{\rm eff}^2 \eta \Gamma(2\eta)}\,,
 where :math:`\Gamma` is the gamma function.
 This analytical norm is approximated so it may not integrate to unity in extremal cases 
-if ellipticity tend to one with large radius or :math:`\eta` much larger than one (outside the default range).
-
+if ellipticity tend to one and radius is large  or :math:`\eta` much larger than one (outside the default range).
 """
 
 # %%
--- a/examples/models/spatial/plot_gen_gauss.py
+++ b/examples/models/spatial/plot_gen_gauss.py
@@ -45,26 +45,25 @@ geom = WcsGeom.create(
 
 tags = [r"Disk, $\eta=0.01$", r"Gaussian, $\eta=0.5$", r"Laplacian, $\eta=1$"]
 eta_range = [0.01, 0.5, 1]
-r_eff = 1
+r_0 = 1
 e = 0.5
 phi = 45 * u.deg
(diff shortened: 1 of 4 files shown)


examples/models/spatial/plot_gen_gauss.py
```diff
@@ -0,0 +1,83 @@
...
+    skydir=(lon_0, lat_0), binsz=dr, width=(2 * reval, 2 * reval), frame="galactic",
+)
+
+tags = [r"Disk, $\eta=0.01$", r"Gaussian, $\eta=0.5$", r"Laplacian, $\eta=1$"]
+eta_range = [0.01, 0.5, 1]
+r_eff = 1
+e = 0.5
+phi = 45 * u.deg
+fig, axes = plt.subplots(1, 3, figsize=(9, 6))
+for k, eta in enumerate(eta_range):
```
adonath, line 52: You can directly iterate over the axes here as well like `for ax, eta in zip(axes, eta_range):`

examples/models/spatial/plot_gen_gauss.py
```diff
@@ -0,0 +1,83 @@
...
+    model = GeneralizedGaussianSpatialModel(
+        lon_0=lon_0 * u.deg,
```
adonath, line 54: The units could be moved to where thew variable is defined. `WcsGeom.create()` handled units correctly...

gammapy/modeling/models/tests/test_spatial.py
```diff
@@ -103,6 +105,30 @@ def test_sky_gaussian():
...
 
 
+def test_generalized_gaussian():
+    reval = 6
+    dr = 0.01
+    geom = WcsGeom.create(
+        skydir=(0, 0), binsz=dr, width=(2 * reval, 2 * reval), frame="galactic",
+    )
+    # check normalization is robut for a large set of values
+    for eta in np.arange(0.1, 1.01, 0.3):
```
adonath, line 115: This kind of testing is usually implement using `purest.mark.parametrize()`, see e.g. https://github.com/gammapy/gammapy/blob/master/gammapy/maps/tests/test_base.py#L29. The big advantage is that the "parameterised" tests will be collected and executed by `pytest` as individual tests. So developers will get a much more meaningful test output.

gammapy/modeling/models/spatial.py
```diff
@@ -371,6 +371,73 @@ def to_region(self, **kwargs):
...
+    lon_0 = Parameter("lon_0", "0 deg")
+    lat_0 = Parameter("lat_0", "0 deg", min=-90, max=90)
+    r_eff = Parameter("r_eff", "1 deg")
+    eta = Parameter("eta", 0.5, min=0.01, max=1.0)
+    e = Parameter("e", 0.0, min=0.0, max=1.0, frozen=True)
+    phi = Parameter("phi", "0 deg", frozen=True)
+
+    @staticmethod
+    def evaluate(lon, lat, lon_0, lat_0, r_eff, eta, e, phi):
+        lon = Longitude(lon).wrap_at(f"{np.min(lon)+180*u.deg}")
```
adonath, line 406: Can this be unified / simplified using the `compute_sigma_eff` function we use in the other elliptical models?

examples/models/spatial/plot_gen_gauss.py
```diff
@@ -0,0 +1,83 @@
...
+lon_0 = 20
+lat_0 = 0
+reval = 3
+dr = 0.02
+geom = WcsGeom.create(
+    skydir=(lon_0, lat_0), binsz=dr, width=(2 * reval, 2 * reval), frame="galactic",
+)
+
+tags = [r"Disk, $\eta=0.01$", r"Gaussian, $\eta=0.5$", r"Laplacian, $\eta=1$"]
+eta_range = [0.01, 0.5, 1]
```
adonath, line 47: Totally minor: maybe you can use 4 values, so that we get a full 2x2 grid in the model gallery?

adonath (changes requested):
Thanks @QRemy, I have left a few comments. The most important one would be the refactoring of the tests and check whether the `compute_sigma_eff` function can be re-used...

examples/models/spatial/plot_gen_gauss.py
QRemy: not if skydir is given as a tuple

adonath: The we should make this work...:-) But I'm fine to leave it as is here and change later...

examples/models/spatial/plot_gen_gauss.py
QRemy: I prefer to keep only 3 because these values correspond to "named" nested model.

adonath (approved):
Thanks a lot @QRemy, I have no further comments!

PR #3038, opened 2020-10-02
Implement Fit.stat_surface() method

luca-giunti:
**Description**
Implement a `Fit` method that computes the 2-dimensional fit statistic profile (or rather surface) by fitting the model on a 2D grid of parameters values. It is a simple extension of the `Fit.stat_profile()` method.

**Dear reviewer**
I have a few comments:
1) Is the name OK? Better ideas?
2) I propose, in a further PR, to improve both the `Fit.stat_profile()` and  `Fit.stat_profile_2D()` methods by:
- returning an additional dictionary element (`"results"`) containing the fit results (i.e. the output of `Fit.optimize()`) at each grid node. Extremely useful for debugging 
- using the `tqdm` library (already shipped with gammapy) to provide a nice progress bar. 
3) Regarding the plotting: I think it doesn't really reduce to a simple call of `plt.contour()`, because in the end what is nice to inspect is not the 2D profile of `-2loglike` values retuned by `Fit.stat_profile_2D`, but rather the profile of "how many Gaussian sigma deviations away from the minimum each point of the grid is". This is easy to achieve for the method `Fit.stat_profile`, because there is just 1 floating degree of freedom and the `sqrt(TS)` gives directly the profile in number of Gaussian sigmas, but in the 2D case people would have to compute the corresponding p-values of a Chi2 function with 2 df and convert that to standard deviations... it's actually 3 simple lines of `scipy.stats` code, but I would like to embed it into a a function. Whether this function goes into `gammapy.visualization` or lives in a tutorial, is an open question

### DIFF
--- a/gammapy/modeling/fit.py
+++ b/gammapy/modeling/fit.py
@@ -380,8 +380,6 @@ class Fit:
             Parameter values to evaluate the fit statistic for.
         reoptimize : bool
             Re-optimize other parameters, when computing the fit statistic profile.
-        backend : str
-            Which backend to use (see ``gammapy.modeling.registry``)
         **optimize_opts : dict
             Keyword arguments passed to the optimizer. See `Fit.optimize` for further details.
 
--- a/gammapy/modeling/fit.py
+++ b/gammapy/modeling/fit.py
@@ -367,8 +367,7 @@ class Fit:
         The method used is to vary two parameters, keeping all others fixed.
         So this is taking a "slice" or "scan" of the fit statistic.
 
-        Caveat: This method can be very computationally intensive and slow. In future releases, multi-core computation
-        will make it more efficient.
+        Caveat: This method can be very computationally intensive and slow
 
         See also: `Fit.minos_contour`
 
--- a/gammapy/modeling/fit.py
+++ b/gammapy/modeling/fit.py
@@ -359,7 +359,7 @@ class Fit:
 
         return {"values": values, "stat": np.array(stats)}
 
-    def stat_profile_2D(
+    def stat_surface(
         self, x, y, x_values, y_values, reoptimize=False, **optimize_opts
     ):
         """Compute fit statistic contour.
--- a/gammapy/modeling/fit.py
+++ b/gammapy/modeling/fit.py
@@ -396,6 +396,8 @@ class Fit:
         stats = []
         with parameters.restore_values:
             for x_value, y_value in itertools.product(x_values, y_values):
+                # TODO: Remove log.info() and provide a nice progress bar
+                log.info(f"Processing: x={x_value}, y={y_value}")
                 x.value = x_value
                 y.value = y_value
                 if reoptimize:
--- a/gammapy/modeling/fit.py
+++ b/gammapy/modeling/fit.py
@@ -1,6 +1,7 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 import logging
 import numpy as np
+import itertools
 from astropy.utils import lazyproperty
 from .covariance import Covariance
 from .iminuit import confidence_iminuit, covariance_iminuit, mncontour, optimize_iminuit
@@ -100,7 +101,6 @@ class Fit:
         fit_result : `FitResult`
             Results
         """
-
 
         if optimize_opts is None:
             optimize_opts = {}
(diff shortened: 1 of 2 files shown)


gammapy/modeling/fit.py
```diff
@@ -359,23 +359,63 @@ def stat_profile(
 
         return {"values": values, "stat": np.array(stats)}
 
-    def stat_contour(self):
+    def stat_profile_2D(
```
adonath, line 362: To me a "2d profile" sounds a bit strange, maybe we call it e.g. `stat_surface()` instead?

adonath (changes requested):
Thanks @luca-giunti, I have left two minor comments. 

We should probably extend the modelling tutorial: https://docs.gammapy.org/0.17/notebooks/modeling.html to show this as well, but this can be done in a separate PR.

luca-giunti:
Hi @adonath ! Thanks for the review. Do you approve either of the two proposals that I made in my 2nd point (i.e. the idea of returning the fit results for debug purpose and/or using `tqdm` to provide some insight on the state of the method execution)? Since the profiling strategy (both in 1D and 2D) can be very time consuming, I think these modifications might be worthwhile... 

Anyway, I will go ahead and adapt the modeling tutorial, as soon as this PR is merged.

adonath:
Thanks @luca-giunti! I agree returning the fit result objects is good idea (actually we should possibly extend it to estimators).

I'm not opposed to adding a progress bar, but again I think we should introduce it uniformly. I could imagine 10 more place were this would be useful. So as a compromise maybe just add ` log.info()` statement for now in the loop?

luca-giunti:
OK @adonath , I agree 100%. So if you agree I will open a reminder issue on this, i.e.: returning fit results and providing progress bars in a consistent way.

gammapy/modeling/fit.py
```diff
@@ -359,23 +359,63 @@ def stat_profile(
...
 
+        Caveat: This method can be very computationally intensive and slow. In future releases, parallel computation
+        will make it more efficient.
+
         See also: `Fit.minos_contour`
 
         Parameters
         ----------
-        TODO
+        x, y : `~gammapy.modeling.Parameter`
```
registerrier, line 377: There is confusion here. Should the method take `Parameter` directly or parameter names in the form of `str`.
How does it work when two parameters have the  same name?

gammapy/modeling/fit.py
```diff
@@ -359,23 +359,63 @@ def stat_profile(
...
+                    stat = self.datasets.stat_sum()
+
+                stats.append(stat)
+
+        stats = np.array(stats)
+        stats = stats.reshape(
+            (np.asarray(x_values).shape[0], np.asarray(y_values).shape[0])
+        )
+
+        return {"x_values": x_values, "y_values": y_values, "stat": stats}
```
registerrier, line 418: For consistency with `Estimators`, I would suggest to replace `stat` by `stat_scan` in the keys of the returned dict and possibly `x_values` by `x_scan`.
Could we use the actual parameter name rather than `x`?

registerrier (approved):
Thanks @luca-giunti. This looks good. 

My main question concerns the parameters of the function: does it take it `Parameter` objects or names? I assume the question can also be raised about `minos_contour` though.

I suppose a second function to properly extract the required contours might be needed.

gammapy/modeling/fit.py
```diff
@@ -359,23 +359,63 @@ def stat_profile(
...
-    def stat_contour(self):
+    def stat_surface(
+        self, x, y, x_values, y_values, reoptimize=False, **optimize_opts
+    ):
         """Compute fit statistic contour.
 
         The method used is to vary two parameters, keeping all others fixed.
         So this is taking a "slice" or "scan" of the fit statistic.
 
+        Caveat: This method can be very computationally intensive and slow. In future releases, parallel computation
```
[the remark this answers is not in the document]
adonath: I would even say just remove the hint to the parallel computation, I think the docstring is not the right place for the a "feature announcement". But keep the caveat...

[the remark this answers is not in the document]
luca-giunti: Agreed

gammapy/modeling/fit.py
luca-giunti: We have to be consistent between `stat_profile`, `stat_surface` and `minos_contour`. Is it OK if we clarify this in a further PR?

adonath: I'd be fine with this. I think the only way to solve this to access parameters via the model itself, e.g. by introducing something like `Model.get_parameter_by_unique_name("model_name.spectral.parameter_name")`

gammapy/modeling/fit.py
luca-giunti: Same comment as above. I used `stat` and `x/y` because it is consistent with `Fit.stat_profile` and`Fit.minos_contour()`. I would suggest to merge this, and later clean up the code to ensure consistency between `Fit` methods and `Estimator` classes

adonath (approved):
Thanks @luca-giunti! No further comment from my side, but please address the comments by @registerrier in a follow up PR.

PR #3041, opened 2020-10-05
Add integral_error for models.spectral

fabiopintore:
This PR adds the `integral_error` function into ~gammapy.modeling.models.spectral, which will calculate the error of the integral flux for a given spectral model in a given energy range.
A test for a `PowerLawSpectralModel` has also been added.

### DIFF
--- a/gammapy/modeling/models/spectral.py
+++ b/gammapy/modeling/models/spectral.py
@@ -169,18 +169,17 @@ class SpectralModel(Model):
 
         Parameters
         ----------
-        emin, emax : `~astropy.units.Quantity`
-            Lower and upper bound of integration range.
+        emin, emax : `~numpy.array`
+            Arrays of minimum and maximum energies.
 
         Returns
         -------
         flux, flux_err : tuple of `~astropy.units.Quantity`
             Integral flux and flux error betwen emin and emax.
         """
-        E = np.linspace(emin,emax,1000)
-        dE = E[1] - E[0]
-        flux = np.sum(self.integral(E[:-1],E[1:]))
-        flux_err = np.sum(self.evaluate_error(E[:-1], epsilon=1e-4)[1]) * dE
+        dE = emax - emin
+        flux = np.sum(self.integral(emin,emax))
+        flux_err = np.sum(self.evaluate_error(emin, epsilon=1e-4)[1] * dE)
 
         return u.Quantity([flux.value, flux_err.value], unit=flux.unit)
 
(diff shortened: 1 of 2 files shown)


adonath (changes requested):
Thanks @fabiopintore, one important comment already: 
`.integral_error()` hast work with arrays of emin and emax, please take a look a the prototype implementation again and adapt accordingly.

gammapy/modeling/models/spectral.py
```diff
@@ -163,6 +163,26 @@ def integral(self, emin, emax, **kwargs):
...
+        emin, emax : `~numpy.array`
+            Arrays of minimum and maximum energies.
+
+        Returns
+        -------
+        flux, flux_err : tuple of `~astropy.units.Quantity`
+            Integral flux and flux error betwen emin and emax.
+        """
+        dE = emax - emin
+        flux = np.sum(self.integral(emin,emax))
```
adonath, line 181: The idea is, that `flux` as well as `flux_err` can be arrays as well. You can either achieve this by using relative error propagation like `df = f * dg ./ df` or using the approach outlined in the "Flux" section here: https://github.com/gammapy/gammapy-extra/blob/master/experiments/uncertainty_estimation_prototype.ipynb

gammapy/modeling/models/spectral.py
```diff
@@ -163,6 +163,24 @@ def integral(self, emin, emax, **kwargs):
...
+        emin, emax : `~numpy.array`
+            Arrays of minimum and maximum energies.
+
+        Returns
+        -------
+        flux, flux_err : tuple of `~astropy.units.Quantity`
+            Integral flux and flux error betwen emin and emax.
+        """
+        flux = self.integral(emin,emax)
+        flux_err = flux * self.evaluate_error(emin, epsilon=1e-4)[1] / self(emin)
```
adonath, line 181: Minor: as `evaluate_error()` returns both dude as well as dude_err, there is no need to recompute it. You can just use  `dnde, dnde_err = self.evaluate_error()` and maybe use the center energy like `np.sqrt(e_min * e_max)`

adonath (changes requested):
Thanks @fabiopintore, looks mostly good to me now. Can you maybe add one more test case. E.g. using an exponential cutoff model?

gammapy/modeling/models/spectral.py
fabiopintore: Done.

adonath (approved):
Thanks a lot @fabiopintore, no further comments from my side!

PR #3043, opened 2020-10-05
Harmonise npred between OnOff and normal datasets

AtreyeeS:
**Description**

This pull request ensures:
 - `npred_sig(model=None)`  : returns the model predicated counts
- `npred`  : returns the total (signal + background) counts

This behaviour is now consistent between OnOff and Normal datasets.
I also adapt the docs to sort the inconsistancy between `mu_sig` as the expected signal counts in the signal or the background region

**Dear reviewer**

### DIFF
--- a/gammapy/datasets/map.py
+++ b/gammapy/datasets/map.py
@@ -86,9 +86,7 @@ def create_map_dataset_geoms(
     if migra_axis:
         geom_edisp = geom_irf.to_cube([migra_axis, energy_axis_true])
     else:
-        geom_edisp = geom_irf.to_cube(
-            [geom.axes["energy"], energy_axis_true]
-        )
+        geom_edisp = geom_irf.to_cube([geom.axes["energy"], energy_axis_true])
 
     return {
         "geom": geom,
@@ -338,6 +336,32 @@ class MapDataset(Dataset):
                 npred = evaluator.compute_npred()
                 npred_total.stack(npred)
         return npred_total
+
+    def npred_sig(self, model=None):
+        """"Model predicted signal counts. If a model is passed, predicted counts from that component is returned.
+        Else, the total signal counts are returned.
+
+        Parameters
+        -------------
+        model: `~gammapy.modeling.models.Models`, optional
+            The model to computed the npred for. If none, the sum of all components (minus the background model)
+            is returned
+
+        Returns
+        ----------
+        npred_sig: `gammapy.maps.WcsNDMap`
+        """
+        if model is None:
+            if self.background_model is None:
+                return self.npred()
+            return self.npred() - self.background_model.evaluate()
+        else:
+            return MapEvaluator(
+                model=model,
+                exposure=self.exposure,
+                edisp=self._edisp_kernel,
+                gti=self.gti,
+            ).compute_npred()
 
     @classmethod
     def from_geoms(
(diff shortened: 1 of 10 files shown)


gammapy/datasets/map.py
```diff
@@ -339,6 +337,32 @@ def npred(self):
                 npred_total.stack(npred)
         return npred_total
 
+    def npred_sig(self, model=None):
+        """"Model predicted signal counts. If a model is passed, predicted counts from that component is returned.
+        Else, the total signal counts are returned.
+
+        Parameters
+        -------------
+        model: `~gammapy.modeling.models.Models`, optional
```
adonath, line 346: This should be a `SkyModel` object...

gammapy/datasets/map.py
```diff
@@ -339,6 +337,32 @@ def npred(self):
...
+            The model to computed the npred for. If none, the sum of all components (minus the background model)
+            is returned
+
+        Returns
+        ----------
+        npred_sig: `gammapy.maps.WcsNDMap`
+        """
```
adonath, line 353: Missing second line here...

gammapy/datasets/map.py
```diff
@@ -339,6 +337,32 @@ def npred(self):
...
+        if model is None:
+            if self.background_model is None:
+                return self.npred()
+            return self.npred() - self.background_model.evaluate()
+        else:
+            return MapEvaluator(
```
adonath, line 359: I would just rely on `.evaluators[model]` here and not create a new `MapEvaluator`, also the current implementation does not work, because `MapDataset._edisp_kernel` does not exist.

gammapy/datasets/map.py
```diff
@@ -1525,9 +1556,31 @@ def excess(self):
...
+            Parameters
+            -------------
+            model: `~gammapy.modeling.models.Models`, optional
+               The model to computed the npred for.
+
+            Returns
+            ----------
+            npred_sig: `gammapy.maps.WcsNDMap`
+        """
+        if model is None:
```
adonath, line 1572: I think this can be simplified, as the `if model is None` handling is already implemented in the `MapDataset` class.

gammapy/datasets/spectrum.py
```diff
@@ -293,6 +293,32 @@ def npred(self):
...
+        Returns
+        ----------
+        npred_sig: `gammapy.maps.RegionNDMap`
+        """
+        if model is None:
+            if self.background_model is None:
+                return self.npred()
+            return self.npred() - self.background_model.evaluate()
+        else:
+            return MapEvaluator(
```
adonath, line 315: Again I would just rely on `.evaluators[model]`

gammapy/datasets/spectrum.py
```diff
@@ -933,6 +961,28 @@ def alpha(self):
         """Exposure ratio between signal and background regions"""
         return self.acceptance / self.acceptance_off
 
+    def npred_sig(self, model=None):
```
adonath, line 964: Why is there a 2nd implementation of `.npred_sig()`? Copy paster error?

adonath (changes requested):
Thanks a lot @AtreyeeS , I have left a few comments...

gammapy/datasets/map.py
AtreyeeS: The behaviour is a bit different in the 2 cases; for `model=None`, `MapDatasetOnOff.npred_sig` is the sum of all model components, whereas for `MapDataset.npred_sig` is `npred - background`.
Do you have a specific suggestion for simplification?

AtreyeeS:
Thanks @adonath ! It works much better using the `evaluators` directly! Addressed the comments

gammapy/datasets/map.py
adonath: Isn't this the same?

adonath: Ah, I see result-wise this is the same, but implementation-wise it is not. Ok, to keep. Maybe we can simplify this later....

adonath (approved):
Thanks @AtreyeeS, I have no further comments...

PR #3047, opened 2020-10-07
Prototype for datasets parallel evaluation with ray

QRemy:
This PR add support for parallel evaluation of datasets using ray
https://github.com/ray-project/ray.

This is possible without changes of the current datasets and models API in gammapy, it requires only an few internal changes. It relies mainly on the switch of the datasets to ray-actors during the fit and the copy of the fitted parameters from datasets to actors.

The gain in time increase with problem complexity: large number of joint datasets, and large number of fit call required. The drawback of this prototype is that the copy of datasets to actors increases memory cost.

For now I keep this as a preliminary draft to be discussed as post 1.0 improvement.
(the test script could be moved to a dedicated benchmark in final version)


adonath:
Thanks a lot @QRemy, this looks very promising! I'm pretty sure we will take some time and develop this further in 2021.

For now what would be interesting to us, is if the model management works. Do you have a minimal example, that shows, that the `ProperModels` approach works? So create multiple remote datasets add a new model component to one of them and synchronise the others remotely?

QRemy:
In the case where the DatasetsActors are copied from existing datasets user can still fully interact with the DatasetsActor as for a Datasets.

I added another examples where the DatasetsActors is created directly from actors setup after the data reduction, in that case I introduced a list of "ghost" datasets that contain only models (but not data that are on the remote) so user can modify global model as they want and then the model are copied and parameters updated on the remote only during the fit.

Jaleleddine:
The last time me and Regis have executed this Prototype for datasets parallel evaluation, we have not noted a difference in the time execution with and withou ray, unless we were tricked. Currently, i am exploring a multi-core processing (Parallel Python) for the same goal, it works well for simple examples and gives remarkable time saving.

adonath:
Thanks @Jaleleddine for looking into this, but my recommendation would be to not spend too much tome in this now. I'm pretty sure to find a good solution for distributed computing, we will need to rely on external consulting from dask / ray experts as well...

adonath:
When using this prototype I saw a factor of ~2 in performance. What I already learned from this prototype is that, we need a design where one remote process (or actor) can handle multiple dataset objects. Introducing one process per dataset introduce too much overhead in communication...

gammapy/datasets/core.py
```diff
@@ -490,3 +500,76 @@ def index(self, key):
...
+                    res = res(**kwargs) #no longer parallel but works with plots
+            return results
+        return wrapper
+
+    def _copy_cavariance_data(self):
+        # TODO: this avoid  ValueError: assignment destination is read-only in set_subcovariance
+        # self._data[np.ix_(idx, idx)] = covar.data.copy()
+        # better way/place to do that ?
+        for m in self.models:
+            m._covariance.data = m._covariance._data.copy()
```
QRemy, line 560: should no longer be necessary after #4521

gammapy/datasets/tests/test_map.py
```diff
@@ -707,6 +707,62 @@ def test_map_fit(sky_model, geom, geom_etrue):
...
+    models["test-model"].spatial_model.sigma.frozen = True
+
+    datasets.models = models
+    dataset_2.counts = dataset_2.npred()
+    dataset_1.counts = dataset_1.npred()
+
+    models["test-1-bkg"].spectral_model.norm.value = 0.49
+    models["test-2-bkg"].spectral_model.norm.value = 0.99
+
+    actors = DatasetsActor(datasets)
```
bkhelifi, line 732: In term of API, it is not possible to hide this line in the Fit class? A use of a 'parallel boolean' will allow to replace the list of datasets by the actor.... and then will lighten the user interface by avoiding to use very technical classes.
Do you get my point?

QRemy: Yes I did that in the begining but I removed it because : 
-if the datasets remain available in the main process and the  actors are recreated on each fit.run() call that's not very efficient
-if the datasets are replaced by actors in place (as now) it is better if the user is responsible of it because the API change to get and set attributes on the remote actors. First I don't think the setter should be implemented because attributes on the remote dataset should be immutable except for `models` (and maybe `mask_fit`), for example for now `datasets[0].counts = 1` run but does not apply to the remote dataset.
Secondly you gain the possibility to call (in parallel) any dataset attributes from the datasets like `datasets.counts()`,  `datasets.npred()`, `datasets.plot_residuals()`..., but the same could be done with the current datasets for consistency.

Ideally the `MapDatasetActor` should be have a method `from_maker` to fill them direclty in their respective processes so data are not copied. I'm not sure yet if this should be done through the `DatasetsMaker` or if the `DatasetsActor` should have a `from_maker` method that behave differently.

QRemy: I improved the get and set mecanism.
Now the remote class is automatically updated, and trying to change anything except models and mask_fit raise an error.

gammapy/datasets/actors.py
```diff
@@ -0,0 +1,188 @@
...
+    Support only datasets composed of MapDataset.
+
+    Parameters
+    ----------
+    datasets : `Datasets`
+        Datasets
+    """
+
+    def __init__(self, datasets=None):
+        from ray import get
```
bkhelifi, line 27: I have a naive question: to my knowledge ray is not yet a mandatory dependency. Can you confirm?
If ray is not installed, this line will create a dirty error if a user has a code that resembles to the test function. Is it not better to raise a clean error in ray does not exist?
The `DatasetsActor`  is the main API interface, now....

gammapy/datasets/actors.py
```diff
@@ -0,0 +1,188 @@
...
+
+    Parameters
+    ----------
+    dataset : `MapDataset`
+        MapDataset
+    """
+
+    def __init__(self, dataset):
+        self.__dict__.update(dataset.__dict__)
+        if self.models is None:
```
bkhelifi, line 181: Can you add a test for that?

gammapy/datasets/tests/test_map.py
```diff
@@ -707,6 +707,61 @@ def test_map_fit(sky_model, geom, geom_etrue):
...
+    fit = Fit()
+    result = fit.run(datasets=actors)
```
bkhelifi, line 734: An other naive question: how this test check that a multiprocessing is really done? Is it possible to check that? Do we have a return message containing the number of processes that has been used? 
And should we add this information in the FitResult class?

QRemy:
> The `DatasetsActor` seems to be now the main API interface In this context, the docstring might be improved for the non-experts... saying that this is the main class to make multi-processing computation, adding an exemple in the docstring to explain how to use it (in addition to a small tutorial). In this context, is it useful to expose the `MapDatasetActor` class? If yes, its docstring could be also improved.

I wouldn't say that The `DatasetsActor` is the main API interface, even for multi-processing computation the other methods we use for embarrassingly parallel problems (FP,LC,TS) are probably faster as you don't have to exchange informations between processes.

Now the DatasetsActor has the same methods/properties as the Datasets except that most of them are immutable and it add the possibility to access list of methods results and properties like `datasets.counts` but this can also be added to the datasets. So I would just mention that and add a see also line tin the docstring that to refers to the Datasets.

gammapy/datasets/tests/test_map.py
QRemy: ray send messages about the number of core in use, I dont' think we need to add this info in the FitResult.

bkhelifi:
As a note for the next dev call: what could be the strategy for the estimators? ...

gammapy/datasets/actors.py
QRemy: The message is `ModuleNotFoundError: No module named 'ray'` same as for the other optional modules (for example healpy), we don't have a special warning message.

gammapy/datasets/core.py
```diff
@@ -3,6 +3,7 @@
 import collections.abc
 import copy
 import html
+import inspect
```
QRemy, line 6:
```suggestion
```

registerrier (approved):

I must say the implementation of the frontend and backend ray actors is a bit complex to follow. 

I am fine with introducing this as a non-public API for now.
Opinion @adonath ?

gammapy/modeling/parameter.py
```diff
@@ -713,6 +713,24 @@ def set_parameter_factors(self, factors):
...
+    def set_parameter_values(self, values):
+        """Set values of all parameters.
+
+        Used in the optimizer interface.
+        """
+        idx = 0
+        for parameter in self._parameters:
+            if not parameter.frozen:
+                parameter.value = values[idx]
+                idx += 1
```
QRemy, line 725:
```suggestion
```

gammapy/datasets/actors.py
```diff
@@ -0,0 +1,192 @@
...
+            self.models = dataset.models
+        self.mask_fit = dataset.mask_fit
+        self._to_update = {}  # models and mask_fit are already ok from actor init
+
+    @property
+    def name(self):
+        return self._name
+
+    def _update_stat_sum(self):
+        values = self.models.parameters.get_parameter_values()
```
QRemy, line 160:
```suggestion
        values = self.models.parameters.free_parameters.value
```

gammapy/modeling/parameter.py
```diff
@@ -713,6 +713,14 @@ def set_parameter_factors(self, factors):
                 parameter.factor = factors[idx]
                 idx += 1
 
+
+    def get_parameter_values(self):
+        """Get values of all parameters.
+
+        Used in the optimizer interface.
+        """
+        return [p.value for p in self._parameters if not p.frozen]
```
QRemy, line 722:
```suggestion
```

QRemy:
With the latest changes it works for flux points (lightcurve not tested) but from the test I can tell that it is very inefficient.
I think the overhead come from  the methods that loop over datasets, in particular here a lot of time is wasted in getting `datasets.energy_ranges` that load the counts map for each remote to get this information.

 Theses methods should rewritten case by case such as the loop is parallelised or to get smaller variables from the remote (for example here we could introduce energy_range_full on dataset and get that instead of the counts map).

Screenshot 2023-11-20 at 01 08 07

gammapy/datasets/actors.py
```diff
@@ -0,0 +1,225 @@
...
+                    raise TypeError(
+                        f"For now datasets parallel evaluation is only supported for MapDataset, got {d0.tag} instead"
+                    )
+                datasets_list.append(MapDatasetActor(d0))
+                datasets.remove(d0)  # moved to remote so removed from main process
+            self._datasets = datasets_list
+            self._ray_get = get
+
+        # trigger actors auto_init_wrapper (so overhead so appears on init)
+        self.name
```
QRemy, line 47: added this to ensure that the actors setup is done on init so the time is not counted in the first method used
Screenshot 2023-11-20 at 13 43 48
so the method I mentioned don't add that much time, it's just the creation of new remote actors for each energy bins (that could be solved using masking instead of slicing).

QRemy:
added test for lightcurve and it works

gammapy/datasets/actors.py
QRemy: yes slicing will create new datasets/actors so the data will be copied and ray will have to start a new kernel to perform the computations on this dataset.

gammapy/datasets/actors.py
```diff
@@ -0,0 +1,192 @@
...
+            while datasets:
+                d0 = datasets[0]
+                datasets_list.append(MapDatasetActor(d0))
+                datasets.remove(d0)  # moved to remote so removed from main process
+            self._datasets = datasets_list
+            self._actors = actors
+            self._ray_get = get
+
+    def insert(self, idx, dataset):
+        if isinstance(dataset, Dataset):
```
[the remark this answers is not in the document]
QRemy: I could do it for the other datasets in separate PR because each of them require to create new actor classes.

gammapy/datasets/__init__.py
```diff
@@ -24,6 +25,7 @@
     "Dataset",
     "DATASET_REGISTRY",
     "Datasets",
+    "DatasetsActor",
```
QRemy, line 28:
```suggestion
```

QRemy:
I implemented the suggestions from the code reviews, so now it is private and added the tests for the estimators.
I'm going to merge this now and we could do more cleanup in other PRs because this one getting too long.

Issue #2953, opened 2020-07-17
Apply `mask_safe` in `MapDataset.to_spectrum_dataset()`


adonath:
This is a reminder issue to also apply the `dataset.mask_safe` in the `.to_spectrum_dataset()` method. Given the case, that the extraction region overlaps with the cut in offset, this will lead to different results, if the mask is applied or not. For consistency with `.stack()` and `.to_image()` I propose to apply the mask to the data...

adonath:
Additional comments on the same issue are in #3018

PR #3049, opened 2020-10-07
Mask to spectrum

registerrier:
**Description**

This pull request adds `mask_safe` support in `MapDataset.to_spectrum_dataset`.

The `WcsNDMap.get_spectrum()` is modified to ensure dtype is conserved.

**Dear reviewer**

This should be ready. A test for `MapDatasetOnOff.to_spectrum_dataset()` might be added to make sure things are done consistently there as well.

### DIFF
--- a/gammapy/datasets/spectrum.py
+++ b/gammapy/datasets/spectrum.py
@@ -956,7 +956,9 @@ class SpectrumDatasetOnOff(SpectrumDataset):
     @property
     def alpha(self):
         """Exposure ratio between signal and background regions"""
-        return self.acceptance / self.acceptance_off
+        alpha = self.acceptance / self.acceptance_off
+        np.nan_to_num(alpha.data, copy=False)
+        return alpha
 
     def npred_sig(self, model=None):
         """"Model predicted signal counts. If a model is passed, predicted counts from that component is returned.
--- a/gammapy/datasets/spectrum.py
+++ b/gammapy/datasets/spectrum.py
@@ -667,7 +667,6 @@ class SpectrumDataset(Dataset):
         """
         info = dict()
         mask = self.mask_safe.data if in_safe_energy_range else slice(None)
-        print(self.mask_safe.data)
 
         info["name"] = self.name
         info["livetime"] = self.livetime.copy()
--- a/gammapy/datasets/spectrum.py
+++ b/gammapy/datasets/spectrum.py
@@ -667,6 +667,7 @@ class SpectrumDataset(Dataset):
         """
         info = dict()
         mask = self.mask_safe.data if in_safe_energy_range else slice(None)
+        print(self.mask_safe.data)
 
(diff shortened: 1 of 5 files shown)


gammapy/datasets/spectrum.py
```diff
@@ -667,6 +667,7 @@ def info_dict(self, in_safe_energy_range=True):
         """
         info = dict()
         mask = self.mask_safe.data if in_safe_energy_range else slice(None)
+        print(self.mask_safe.data)
```
adonath, line 670: Remove print statement...

gammapy/datasets/tests/test_map.py
```diff
@@ -201,6 +201,11 @@ def test_to_spectrum_dataset(sky_model, geom, geom_etrue, edisp_mode):
     spectrum_dataset_corrected = dataset_ref.to_spectrum_dataset(
         on_region, containment_correction=True
     )
+    mask = np.ones_like(dataset_ref.counts, dtype='bool')
+    mask[1,:,:]=0
```
adonath, line 205: Maybe define a spatial part for the mask as well?

gammapy/maps/wcsnd.py
```diff
@@ -26,7 +26,7 @@
 
 
 class WcsNDMap(WcsMap):
-    """HEALPix map with any number of non-spatial dimensions.
+    """WCS map with any number of non-spatial dimensions.
```
adonath, line 29: Funny that this hasn't been noticed earlier...does it mean that users never look at docstrings? Or maybe they just don't care...:-)

adonath (changes requested):
Thanks @registerrier, functionality wise this looks good to me!

However I think there is an efficiency concern for large images. E.g. in the `ExcessProfileEstimator` we call this methods O(1e2) time or more. Always creating a `mask_safe` if there is none and always multiplying the full array is maybe problematic. Instead I would propose the following solution, that we have e.g. in `.stack()` as well:
- Add a `weights` argument to `.get_spectrum()` and `.to_region_nd_map()`
- Check if weights is not None and take it into account on the cutout data if present. Either by making a second cutout for the weights or re-using the slices defined in `cutout.geom.cutout_info`
- The implementation in `.to_spectrum_dataset()` should then simplify to `.get_spectrum(..., weights=self.mask_safe)`

registerrier:
A `weights` argument has been added to `WcsNDMap.get_spectrum` and `WcsNDMap.to_on_region`. It is assumed that the geometry is identical to the `Map` (contrary to `WcsNDMap.stack`.
The `dtype` issue in now dealt with using `np.any` instead of `np.sum` for the mask.

adonath (approved):
Thanks a lot @registerrier, no further comments from my side.

adonath:
Fixes #2953.

PR #2942, opened 2020-07-09
PIG 20 - Global model API

adonath:
**Description**
This pull request includes a draft for the PIG for a global model API in Gammapy. It outlines handling and distribution of model components from and to datasets as well as dataset interaction with models.
 

**Dear reviewer**
The PIG is still in a draft state it will be improved step by step by discussion and comments in this PR.

PR #3051, opened 2020-10-08
Introduce dataset specific FoVBackgroundModel

adonath:
**Description**

This pull request introduce the dataset specific `BackgroundIRFModel`, as proposed in the #2942. technically there is no real need for this `BackgroundIRFModel`, but it simplifies the model and dataset bookkeeping, as it's de facto treated as an observation specific IRF. Including stacking, mask_safe handling etc. Technically it could be handled as a "normal" model component as well, but then the bookkeeping become tricky as a separate data reduction process for the models would be required. The logic is now as following:

- The MapDataset now takes a `background` argument, which represents just the plain background map.
- Optionally there is the `BackgroundIRFModel`, which holds the background parameters and is multiplied with the background map for npred evaluation
- The `MapDataset.background` property returns the model corrected background, if the `BackgroundIRFModel` is present in the model list

In general my impression is that introducing this special case for the instrumental background of IACTs cleans up our dataset / model concept and simplifies the implementation in many places. This way we get a struct separation between `MapDataset` and `Models` which we didn't have before. However there are a few questions to decide:

- Should `MapDataset` objects come with a pre-defined `BackgroundIRFModel`? Currently this is the case, but only because that's how it used to be. On one hand it's convenient for users, but it also leads to a non-transparent model handling, as implicitly the background is always fitted, without an explicit model declaration.
- Maybe the model is only added to the dataset, if the `FoVBackgroundMaker` is called explicit. Does the `FoVBackgroundMaker` add the model with the predefined parameters or does it directly apply the correction and resets the model?
- Which names are appropriate for the model? I feel like one could also introduce `MapDataset.acceptance` and an `AcceptanceBackgroundModel` or `FoVBackgroundModel`. This way one would get a cleared distinction between the map / acceptance model and the predicted background from it.
- The `BackgroundIRFModel` can only be assigned to a single `MapDataset` and the name cannot be chosen freely. It is hard-coded to `dataset-name-bkg`. I think we should keep it this way. Are there any objections?
- What to do about the current `BackgroundModel`? My proposal would be to rename it to `DataModel` and basically keep as is...

Once this PR is in we can decide on a uniform model handling in dataset methods. But I'll make a separate issue for this. in general.


adonath:
The actual implementation of the model is here: https://github.com/gammapy/gammapy/pull/3051/files?file-filters%5B%5D=.py#diff-d1acd71a310271432246338319845e19

adonath:
The corresponding changes in `gammapy-data` are here: https://github.com/gammapy/gammapy-data/pull/4

adonath:
I would like to summarise my proposal for some of the questions raise above:
- I think we should have the background-irf as an implicit model component on the `MapDataset`, which is always applied, even if there is no `BackgroundIRFModel` defined. The `BackgroundIRFModel` only holds the model parameters and only comes into play, when the background is supposed to be fitted. If users don't want it to be part of the npred (which is almost never the case I suppose), they either define a `BackgroundIRFmodel` with `norm=0`, or even clearer just set `dataset.background = None`. With the implicit model component we also get consistent behaviour with the `MapDatasetOnOff`.
- If possible I would like to avoid the double bookkeeping of the `Models` and `Dataset`. In fact it is one of the main motivations for this PR. Once the background data is attached to the `Dataset`, it can be always handled correctly in dataset operations like `.to_spectrum_dataset()`, `.to_image()` etc. The actual model definition would be reset in this case, but the background is again kept as an implicit model component.

adonath:
@registerrier As a compromise to make the model definition more explicit one could maybe allow for `BackgroundIRFModel` without any parameters at all, but it would appear e.g. in the printout and would be created by default....

AtreyeeS:
Thinking a bit about this, I think this PR makes good sense if we document it well

- `Dataset.npred()` always takes the background into account. This distinguishes it from `npred_sig`
- The `FoVBackgroundMaker` adds the `BackgroundIRFModel` - this makes it clear to the user where the background model is really created.

Otherwise I used to find it a bit disconcerting to have a background_model norm as a free parameter without having asked for it.

adonath:
Thanks @AtreyeeS for the feedback! As I wrote in one of my comments before, we could maybe allow the `BackgroundIRFModel` to have `spectral_model = None` and thus no parameters. This has the advantage, that the model would still appear as an explicit model component when e.g. printing the dataset.

AtreyeeS:
So it will read something like:

```
Number of model: 1
Number of parameters: 0  
Number of free parameters: 0
Component 0: BackgroundIRFModel
  Name : dataset-name-bkg
  Datasets names: [dataset-name]
  Parameters:
      None
```

I suppose this should be fine, but I do not know if it is intuitive what a parameter less model means (model with all frozen parameters is obviously fine) - is it something users are familiar with?

adonath:
@AtreyeeS Yes, something like this. I think in general a model component without parameters is fine, it is the case for the `ConstantTemporalModel` and the `ConstantFluxSpatialModel` as well...

AtreyeeS:
True! But `ConstantTemporalModel` and the `ConstantFluxSpatialModel`are used with a `SpectralModel`, so it never looked parameter-less.

Anyways, this is not a problem.
So your proposal is
- `MapDataset` implicitly creates a  `BackgroundIRFModel` with `spectral_model = None`
- `FoVBackgroundMaker` adds a `spectral_model ` to the `BackgroundIRFModel`, `spectral_model = PowerLawSpectralModel` by default.

registerrier:
After some test, I still find it confusing that by default a `BackgroundModel` is created and when applying `dataset.models = None` you still have something applied. While we can document that, this is unintuitive and will lead to errors.

How about splitting `dataset._models` and `dataset._background_model`?
- You have then a more intuitive and self explanatory approach. e.g. you don't need to do `dataset.models.append(model)` to define a source model.   
- And indeed a regular user would assume that dataset.models = models should work. Actually that’s how we are exposing spectral analysis with OnOff datasets
- Then you only need to replace the `dataset.background` setter by a `dataset.background_model` setter
    * If you want to to put a background model to 0, you can set its norm to 0 or simply `dataset.background_model = None`
- You just need to handle the special case of `BackgroundModel` in the `ProperModels` loop.
- Then would would store the IRF map itself e.g. with `dataset._irf_background_map`, and by default the `BackgroundModel` is defined on this map.

adonath:
>After some test, I still find it confusing that by default a BackgroundModel is created and when applying dataset.models = None you still have something applied. While we can document that, this is unintuitive and will lead to errors.

Here I don't agree, because the standard use case (really like 98% of the time) is to take the background model into account in the npred. I even think it's the more safe default behaviour, because if users don't define the background model, it will be never lead to converging fit results. Another argument for me is that it is done implicitly for the `MapDatasetOnOff` as well. To me the difference is just whether you derive the model from a single off observation and take the error into account, or multiple off observations and ignore the error on the background model. So I think the consistency we get by accessing `dataset.background` (which returns the optimised / corrected background model) in both cases is a good thing.

The point is that with the implicit background handling we effectively get the splitting between the source model and the background model, you proposed. And in addition we get the same default behaviour for both dataset types: 
```python3
models = Models([source_1, source_2]) # no background model declared here
dataset.models = models
dataset_on_off.models = models
```
Would work in both cases and implicitly add the background map to the npred prediction. Only in case the background model is supposed to be fitted one adds another model component using the `BackgroundIRFModel. If the background is not supposed to be taken into account (for whatever reason, the remaining 2%...) one can just set the `background` to `None` like:
```python3
datasets.background = None
```
I think this is well defined and reasonable default behaviour.

registerrier:
> Here I don't agree, because the standard use case (really like 98% of the time) is to take the background model into account in the npred. I even think it's the more safe default behaviour, because if users don't define the background model, it will be never lead to converging fit results. Another argument for me is that it is done implicitly for the `MapDatasetOnOff` as well. To me the difference is just whether you derive the model from a single off observation and take the error into account, or multiple off observations and ignore the error on the background model. So I think the consistency we get by accessing `dataset.background` (which returns the optimised / corrected background model) in both cases is a good thing.

 I think we agree here. The dealt behavior should always be that the IRF background is taken into account and having `dataset.background` yielding a normalized background in all situations is the good behavior.

My point was not that you might want to remove the background model by doing `dataset.models = model` but rather that doing it (as is the typical assumption for assigning a source model) you will end up having   `print(dataset.models)` mentioning no background model, while in practice it still applies one (even if only a map).

```python
# First do data reduction
print(dataset.models)
```
yields
```
Component 0: BackgroundIRFModel
  
    Name                      : 62-b9AFG-bkg
    Datasets names            : ['62-b9AFG']
    Spectral model type       : PowerLawNormSpectralModel
    Parameters:
      norm                    :   1.000              
      tilt         (frozen)   :   0.000              
      reference    (frozen)   :   1.000  TeV  
```
 and then if you do:
```python
dataset.models = model
print(dataset.models)
```
yields:
```
Component 0: SkyModel
  
    Name                      : gc-source2
    Datasets names            : None
    Spectral model type       : ExpCutoffPowerLawSpectralModel
    Spatial  model type       : PointSpatialModel
    Temporal model type       : 
    Parameters:
      index                   :   2.268              
      amplitude               :   2.81e-12  1 / (cm2 s TeV)
      reference    (frozen)   :   1.000  TeV         
      lambda_                 :   0.039  1 / TeV     
      alpha        (frozen)   :   1.000              
      lon_0                   :  -0.052  deg         
      lat_0                   :  -0.053  deg  
```
Seeing the background model occurence in one case and not the other is confusing. So that if the IRF background are always applied, the `BackgroundIRFModel` should always exist, unless the user explictly does `dataset.background_model = None`.

AtreyeeS:
> Seeing the background model occurence in one case and not the other is confusing. So that if the IRF background are always applied, the `BackgroundIRFModel` should always exist, unless the user explictly does `dataset.background_model = None`.

Is it necessary to allow a special handling for `dataset.background_model = None` ?
Since this is quite a non-standard use case, the user can implement it by setting `BackgroundIRFModel` norm to `0` or somehow passing a background template with zero filled data. So maybe `MapDataset` can always implicitly apply a `BackgroundIRFModel` ?

adonath:
Thanks @registerrier, now I see what you mean. A little clearer behaviour in this regard was if `MapDataset` objects never come with pre-defined background irf models. So after the data reduction no model is set at all. If the `FoVBackgroundMaker` is run during the data reduction, one just get's the corrected map. Optionally the `FoVBackgroundMaker` could return the `BackgroundIRFModel`, so that users can collect it for joint fits or at least write results to file, if they are interested, like:
```python3

datasets = Datasets()
models = Models()

for obs in observations:
    dataset = dataset_maker.run(reference, obs)
    dataset, bkg_model = fov_bkg_maker.run(dataset, obs)
    datasets.append(datasets)
    models.append(bkg_model) # this is optional if the information is relevant

# And finally
datasets.write("datasets.yaml")
models.write("models.yaml")
```
This is slightly more inconvenient then having the pre-defined background models on the datasets, however it is much more explicit and maybe even simpler to explain.

Another solution is maybe, that we implicitly always attach the `BackgroundIRFModel` in the `dataset.models` setter in case it is not present. However this is maybe problematic with regard to the global model representation again.

adonath:
BTW: if the `FoVBackgroundMaker` creates this model we could maybe re-name it to `FoVBackgroundModel` instead of `BackgroundIRFModel`

QRemy:
- `BackgroundModel` is part of the models so if ones want to  add a source they use :
`dataset.models.append(model)`, or `dataset.models = Models([model, dataset.background_model))`,
or `dataset.models.extent(models)` if multiples.
If they do `dataset.models = model` then  the background model is not taken into account and they will learn of that mistake understanding that `BackgroundModel` is part of the models.
(so I'm fine with the current behavior).
- I don't think the default behaviour should be no correction applied to background_model i.e no BackgroundIRFModel defined, because it is never perfect and in the fit you want a maximum of freedom to kill the sources.
- I would like to still have the current `BackgroundModel` (eventually renamed) in addition of the `BackgroundIRFModel` because it's convenient to store and filter other components that are also background. For example `[m for m in models if isinstance(m, SkyModel)]` will give me only my sources of interest, while `[m for m in models if isinstance(m, BackgroundModel) or isinstance(m, BackgroundIRFModel) ]` will give the backgrounds like irf-bkg, diffuse-bkg(s), far-sources-bkg...
- Can we write a MapDataset with background_model = None ? because it was broken previously

adonath:
> I don't think the default behaviour should be no correction applied to background_model i.e no BackgroundIRFModel defined, because it is never perfect and in the fit you want a maximum of freedom to kill the sources. (and it should be listed with the models).

If I understood correctly you would keep a default with a spectral correction to the background applied, basically keeping the behaviour we have now?

QRemy:
> If I understood correctly you would keep a default with a spectral correction to the background applied, basically keeping the behaviour we have now

yes except that I agree that .background_model should return only a Map, the original one, and the BackgroundIRFModel defined in addition of it as you did in this PR.

adonath:
Yesterday I talked to @registerrier and I have a few more additional thought on this: 

- The solution for the `BackgroundIRFModel` is closely related to the question of the global model API (nothing new...)
- Given that we want to be able to compute and store the full model covariance, there **has to be** one single "global" model object, that knows about all parameters. This object can only be either a global `Models` or `Datasets` object
- For this global model object there are only two possible approaches:
 
1. **"bottom - up"**: all model objects live on the dataset level and are "collected" by the "global" object. This means
users interact with the model like so:
```python3

datasets["dataset-1"].models.append(model)
print(datasets.models)

# something like this should rather be forbidden...I think synchronising in both
# directions is just to complicated for no good reason
datasets.models.append(SkyModel(datasets_names=[]))

```
This solution means that in a distributed computing environment we have to copy model objects into the server process.

2.  **"top - bottom":** all model objects live in the global model objects and are distributed to the individual datasets. In a distributed computing environment this global models object would live in the server process as well. This solution means that users will interact with the model like so:
```python3
datasets.models.append(SkyModel(datasets_names=[]))

# and something like this should rather be forbidden...again two avoid 2 way synchronisation
datasets["dataset-1"].models.append(model)
```
The difference is that in the `"bottom up"` approach the model has to be copied in the server process, every time the user interacts with it. In the "top - bottom" approach all models are present in the server process and are only copied when e.g. a fit is setup.

To me it seems the `"bottom-up"` is the simpler and clearer approach. The consequence for this PR would be that model objects are never created on a `Dataset` level, but just passed on to the dataset either using the `.models` setter or something like `dataset.set_model()`.

QRemy:
It is much simpler if the models exit both on the main process and the remotes because then the global model is always updated in the main process by users interactively or by the optimizer during the fit and then copied to remotes.  So the difference between top-bottom and bottom-top affect the direction from where we copy the models only on the first init of the datasets object.

adonath:
@QRemy Maybe I need to understand better how the synchronisation works, but how is something like this supposed to work in a distributed environment:
```python 3
models = SkyModel(..., datasets_names=["dataset-2"])
datasets["dataset-1"].models.append(models)
```
So the `Datasets` object lives in the main process, along with the new model component. On `Datasets.__getitem__` I would assume the remote dataset is copied into the main process and the models attribute is updated. Then the dataset goes out of scope and is garbage collected. On the next access of `datasets.models`, the dataset is copied again and the model component is added to the global model? Once it is added to the global model it is distributed to the other datasets, which could happen in a top-down manner I guess. I think this is way too complicated to implement / maintain and document. 

By now I'm leaning more and more to the solution, where there is only a single `Models()` object in the session, which is the only interface to all models for the users. It could be coupled to the `Datasets` object but it doesn't have to be. On the dataset level there will then be a `.set_models()` method, which sets up the evaluators. The parameter update happens via something like `dataset.set_models_parameters()` and the `datasets.models` attributed will go away or has to be strictly read only (which is hard to achieve in Python...). The information which model is part of which dataset in stored in the global model object, via the `dataset_names` attribute, or can be checked by printing the dataset (which would rely on the model copies stored in the `.evaluators`).

Looking at your prototype it seems the datasets [remain in the main process]( https://github.com/gammapy/gammapy/pull/3047/files#diff-5e2196c9f989466f5f778cb926bc19c62d5cfb8fc89f5ad640a9d7c8ad6d4a85R518). Which doesn't make sense to me. Of course one still gets the improved performance of the parallel processing, but there might be applications where the datasets don't fit into the memory of the server process. Also on a single machine the memory is used inefficiently, because a copy of the data remains in the main process. I guess one can imagine to not copy the full dataset, but  then we need to have special operations to access for example `.counts`, `.exposure` etc., which seems to add a lot of code overhead for no good reason.

QRemy:
> Looking at your prototype it seems the datasets [remain in the main process](https://github.com/gammapy/gammapy/pull/3047/files#diff-5e2196c9f989466f5f778cb926bc19c62d5cfb8fc89f5ad640a9d7c8ad6d4a85R518)

In the alternative init below, you can see that `.from_actors() `create a fake list of empty datasets that contains only the models copied from the remote and not the data. So after init users can interact with models without accessing the remote. Then the models are updated only in [Fit.run()](https://github.com/gammapy/gammapy/pull/3047/files#diff-e14ed82149eb23f131e4f809b4eb6648e5d134d9176bf82d27b6950d4fd9b103R113)

adonath:
In general this requires some more documentation, that I'm going to add in a follow-up PR....

adonath:
I now implemented the following:
- Rename `BackgroundIRFModel` to `FoVBackgroundModel`
- Store the background map in `MapDataset.background`
- Compute the model corrected background in `.npred_background()`
- Change `MapDatasetOnOff.background` to return `alpha * counts_off`
- Change `MapDatasetOnOff.npred_background()` to return the model dependent predicted background
- For now I kept the behaviour of having pre-defined background models, this could be revisited in context of the global model handling

AtreyeeS:
Hi! I am slightly uncomfortable with the name `MapDatasetOnOff.npred_background()` - it is rather a normalised background. Rest, I think is quite nice

adonath:
Thanks @AtreyeeS. Note that `MapDatasetOnOff.npred_background()` returns the marginalised background, given the source model. So in both cases `.npred_background()` is dependent on the model. One could even imagine that in future `MapDataset.npred_background()` internally marginalises over the background parameters as well...in this case the behaviour would be 100% consistent between the datasets...

adonath:
To follow up on the naming convention of `background` / `npred_background`:

- `MapDataset.background` contains the off counts from many stacked off observations normalised to the observation time
- `MapDatasetOnOff.background` contains the off counts from a single off observations normalised to the observation time
- `MapDataset.npred_background()` returns the predicted background counts depending on the source model (currently there is no implicit marginalisation, but OK...)
- `MapDatasetOnOff.npred_background()` returns the predicted background counts depending on the source model, by marginalising

Together with the notations introduced in https://docs.gammapy.org/dev/stats/index.html#notations, I think we can achieve a 100% consistent behaviour and naming scheme for our dataset and counts statistics classes.

Please think about it once again @registerrier and @AtreyeeS. I am rather confident, that you will agree after...

AtreyeeS:
> * `MapDataset.background` contains the off counts from many stacked off observations normalised to the observation time

I would not call it `off counts`, but I see what you mean

> * `MapDatasetOnOff.background` contains the off counts from a single off observations normalised to the observation time

`MapDatasetOnOff.background` should also be modified accordingly after stacking, no?  
Maybe the doc string can specify `normalised off counts = alpha * n_off` (or counts_off, whatever it is called now - sorry I am getting a bit confused with the large diff)

> * `MapDataset.npred_background()` returns the predicted background counts depending on the source model (currently there is no implicit marginalisation, but OK...)

I agree with you that it is consistent now API wise, but it is still a bit confusing unless well explained. For now, it is fine, but I  think it needs better documentation before 1.0

adonath:
For some reason the Travis build is stuck, but the Azure builds and local builds are fine. I'll go ahead an merge this PR now and prepare a follow up PR to clarify and clean up the stats / dataset naming of quantities.

PR #3052, opened 2020-10-09
Adapt LightCurveEstimator to take energy edges

registerrier:
**Description**

This pull request modifies the `LightCurveEstimator` to take several energy edges at once. It basically return a `FluxPoints` for each time bin. And they are stored as array in the table.
The `LightCurve` object has to be modified to take this into account.

**Dear reviewer**

The PR is not finished. It is open for discussion.
Once possible different approach is to return several distinct light curves. But this is a different behavior than the `ExcessProfileEstimator`.


adonath (approved):
Thanks @registerrier, there is one remaining test fail, otherwise looks good to me..

adonath (approved):

PR #3053, opened 2020-10-09
Add bkg systematics condition for the sensitivity computation

bkhelifi:
**Description**
For the existing class, a new condition is added to compute the sensitivity, the excess being larger than a fraction of the number of background. This condition deals in the CTA way the background systematics.

**Dear reviewer**
An extra-argument is passed to the init function, the fraction of background counts. As a consequence, in addition to the criterion 'significance' and 'gamma', the new criteria 'bkg' is inserted in the final table.
The associated test function is updated to test this new criteria.

### DIFF
--- a/docs/tutorials/cta_sensitivity.ipynb
+++ b/docs/tutorials/cta_sensitivity.ipynb
@@ -182,7 +182,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "sensitivity_estimator = SensitivityEstimator(gamma_min=5, n_sigma=3)\n",
+    "sensitivity_estimator = SensitivityEstimator(gamma_min=5, n_sigma=3, bkg_syst_fraction=0.10)\n",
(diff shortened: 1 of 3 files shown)


docs/tutorials/cta_sensitivity.ipynb
```diff
@@ -182,7 +182,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "sensitivity_estimator = SensitivityEstimator(gamma_min=5, n_sigma=3)\n",
+    "sensitivity_estimator = SensitivityEstimator(gamma_min=5, n_sigma=3, bkg_syst_fraction=0.10)\n",
```
AtreyeeS, line 185: Perhaps it is better to add a line about the `bkg_syst_fraction` in the previous markup cell.

adonath (approved):
Thanks @bkhelifi, no further comments from my side...

PR #3056, opened 2020-10-10
Return fit results dict in Fit methods

luca-giunti:
**Description**
As additional output, I appended all fit results to the dictionary returned by `Fit.stat_profile()` and `Fit.stat_surface()`.

**Dear reviewer**
Can we do the same for Estimators? It's a bit more problematic there because they don't return dicts...

### DIFF
--- a/gammapy/modeling/fit.py
+++ b/gammapy/modeling/fit.py
@@ -326,7 +326,8 @@ class Fit:
         Returns
         -------
         results : dict
-            Dictionary with keys "values" and "stat".
+            Dictionary with keys "values", "stat" and "results". The latter contains an
+            empty list, if `reoptimize` is set to False
         """
         parameters = self._parameters
         parameter = parameters[parameter]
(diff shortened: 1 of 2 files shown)


gammapy/modeling/fit.py
```diff
@@ -326,7 +326,8 @@ def stat_profile(
         Returns
         -------
         results : dict
-            Dictionary with keys "values" and "stat".
+            Dictionary with keys "values", "stat" and "results". The latter contains an
```
AtreyeeS, line 329: Hi @luca-giunti ! I find it a bit uncomfortable to have a `results` key inside a `results` dict. Probably it is better to remake it and explain what it means?

luca-giunti: Ah right. I could rename the dict key to `"fit_results"`. Would it be Ok then?

adonath (changes requested):
Thanks @luca-giunti, I just have the comment to add at least one assert on the returned fit results in our tests somewhere...

adonath (approved):
Thanks @luca-giunti, no further comments from my side.

Issue #3006, opened 2020-09-10
Divide by zero in signal estimate in dataset when using a tabulated flux


tstolarczyk:
gammapy 0.17 official

hello!

I am creating a ```SpectrumDataset``` following the notebook "Spectrum simulation".
I use the same IRF, extract ```e_true``` as proposed (from ```aeff```).
If I use the default example (PowerLaw) given, I have no problem.

When I use my own flux, ```print(dataset)``` crashes (division by zero - see log below), when computing the expected signal.

My flux is created from a tabular spectrum. 
It is initially absorbed with Dominguez
It has 40 flux points for 40 energies.
The spectrum goes brutally to zero around 10 TeV, and the last flux point is at zero strictly.
It is created with (```grb``` is a class with the data)
```
tab = TemplateSpectralModel(energy = grb.Eval,
                                    values               = grb.fluxval[idx],
                                    norm                 = 1.,
                                    interp_kwargs={"values_scale": "linear"})

spec = AbsorbedSpectralModel(tab,grb.eblabs,z)
```

I create a ```e_reco``` from 30 GeV to 500 GeV, in particular, to avoid reaching very low fluxes at high energies.

The data set is created as follows:
```
model = SkyModel(spectral_model=spec,name="GRB")
dataset_empty = SpectrumDataset.create(e_reco = e_reco.edges, 
                                       e_true = e_true.edges, 
                                       region = region)
maker = SpectrumDatasetMaker(selection=["aeff", "edisp", "background"])
dataset = maker.run(dataset_empty, obs)
print (dataset) # This works

dataset.models = model
#dataset.fake(random_state=42)
print(dataset) # This crashes with my flux
```

I have made the following tests:
1. Remove the last point (at zero in the flux table)
2. Remove the EBL absorption
3. Multiply the flux by 1e7 to be at the level of the PowerLaw example

In all cases, it crashes.

Any idea of what happens?
And whatever happens, I guess the division by zero should be prevented....

Thanks, 
Thierry S.

```

---------------------------------------------------------------------------
RuntimeWarning                            Traceback (most recent call last)
<ipython-input-15-3e95f779dd45> in <module>
      2 dataset.models = model
      3 #dataset.fake(random_state=42)
----> 4 print(dataset)

~\AppData\Local\Continuum\anaconda3\gpy0.17\lib\site-packages\gammapy\datasets\spectrum.py in __str__(self)
    111         npred = np.nan
    112         if self.models is not None:
[...]

Issue #139, opened 2014-07-14
Numpy warning handling


cdeil:
There's a bunch of cases in `gammapy` where `np.where` or `np.select` is used e.g. when evaluating `log` or `sqrt` or dividing.

Apparently some of those cases throw warnings that appear for the end-user and in the docs build:

```
<string>:26: RuntimeWarning: invalid value encountered in log                  
/private/tmp/gammapy/build/lib.macosx-10.9-x86_64-2.7/gammapy/spectrum/powerlaw.py:65: RuntimeWarning: divide by zero encountered in true_divide
  term1 = e / (-g + 1)
/private/tmp/gammapy/build/lib.macosx-10.9-x86_64-2.7/gammapy/spectrum/powerlaw.py:67: RuntimeWarning: invalid value encountered in multiply
  return term1 * term2
/opt/local/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/numpy/ma/core.py:6094: RuntimeWarning: overflow encountered in power
  result = np.where(m, fa, umath.power(fa, fb)).view(basetype)
No labeled objects found. Use label='...' kwarg on individual plots.
<string>:46: RuntimeWarning: divide by zero encountered in divide
<string>:46: RuntimeWarning: divide by zero encountered in log10
<string>:46: RuntimeWarning: invalid value encountered in log10
```

The [`np.errstate`](http://docs.scipy.org/doc/numpy/reference/generated/numpy.errstate.html) context manager should be used to avoid these warnings for the end-user (if the code can't be re-written in a better way.

cc @adonath

cdeil:
I think I found a way to identify the scripts / lines of code where the warnings occur.
Putting this in the `gammapy/__init__.py` file:
```python
import numpy as np
np.seterr(all='raise')
```
I get this output:
https://gist.github.com/cdeil/608306fd50c27a454e7d

I'll try to clean up the warnings and get a clean sphinx build today or tomorrow ...

cdeil:
Here's another trick to debug the origin of warnings:
https://github.com/astropy/astropy/pull/4486#issuecomment-172541301

Apart from cleaning up the Gammapy docs build, those two tricks should be mentioned in the developer docs.

cdeil:
This works: http://stackoverflow.com/a/22376126/498873

adonath:
Part of this is addressed in #1701. When building the docs I don't see any warning at all...does sphinx suppress those?

cdeil:
[...]

PR #3057, opened 2020-10-11
Display progress bars in Analysis, Fit methods and Estimator classes

luca-giunti:
**Description**
I used `tqdm` to provide a progress bar in Analysis, Fit methods and Estimator classes.

In some places, there are warnings that break the progress bar, namely:
```
In: analysis_1d.get_datasets()
Out:
  0%|          | 0/4 [00:00<?, ?it/s]No background model defined for dataset cMBPsezR


 25%|██▌       | 1/4 [00:00<00:00,  3.24it/s]No background model defined for dataset gdLztbDM


 50%|█████     | 2/4 [00:00<00:00,  3.30it/s]No background model defined for dataset SQLYi5zW


 75%|███████▌  | 3/4 [00:00<00:00,  3.36it/s]No background model defined for dataset IeRiDiuG


100%|██████████| 4/4 [00:01<00:00,  3.41it/s]
```
In these cases, I guess we should probably display all the warnings after the progress bar is closed, but I wouldn't know how... 

**Dear reviewer**
In my local machine, I get a weird error when I test `pytest gammapy/analysis/test_analysis.py`: a very long traceback ending in `ValueError: I/O operation on closed file`. Hopefully someone knows how to fix this...

### DIFF
--- a/gammapy/analysis/core.py
+++ b/gammapy/analysis/core.py
@@ -20,6 +20,7 @@ from gammapy.maps import Map, MapAxis, WcsGeom
 from gammapy.modeling import Fit
 from gammapy.modeling.models import BackgroundModel, Models
 from gammapy.utils.scripts import make_path
+from gammapy.utils.pbar import pbar
 
 __all__ = ["Analysis"]
 
@@ -270,30 +271,31 @@ class Analysis:
 
         stacked = MapDataset.create(geom=geom, name="stacked", **geom_irf)
 
-        if datasets_settings.stack:
-            for obs in self.observations:
-                log.info(f"Processing observation {obs.obs_id}")
-                cutout = stacked.cutout(obs.pointing_radec, width=2 * offset_max)
-                dataset = maker.run(cutout, obs)
-                dataset = maker_safe_mask.run(dataset, obs)
-                if bkg_maker is not None:
-                    dataset = bkg_maker.run(dataset)
-                if bkg_method == "ring":
-                    dataset.models = Models([BackgroundModel(dataset.background)])
-                log.debug(dataset)
-                stacked.stack(dataset)
-            datasets = [stacked]
-        else:
-            datasets = []
-            for obs in self.observations:
-                log.info(f"Processing observation {obs.obs_id}")
-                cutout = stacked.cutout(obs.pointing_radec, width=2 * offset_max)
-                dataset = maker.run(cutout, obs)
-                dataset = maker_safe_mask.run(dataset, obs)
-                if bkg_maker is not None:
-                    dataset = bkg_maker.run(dataset)
-                log.debug(dataset)
-                datasets.append(dataset)
+        with pbar(total=len(self.observations), show_pbar=True) as pb:
+            if datasets_settings.stack:
+                for obs in self.observations:
+                    cutout = stacked.cutout(obs.pointing_radec, width=2 * offset_max)
+                    dataset = maker.run(cutout, obs)
+                    dataset = maker_safe_mask.run(dataset, obs)
+                    if bkg_maker is not None:
+                        dataset = bkg_maker.run(dataset)
+                    if bkg_method == "ring":
+                        dataset.models = Models([BackgroundModel(dataset.background)])
+                    log.debug(dataset)
+                    stacked.stack(dataset)
+                    pb.update(1)
+                datasets = [stacked]
+            else:
+                datasets = []
+                for obs in self.observations:
+                    cutout = stacked.cutout(obs.pointing_radec, width=2 * offset_max)
+                    dataset = maker.run(cutout, obs)
+                    dataset = maker_safe_mask.run(dataset, obs)
+                    if bkg_maker is not None:
+                        dataset = bkg_maker.run(dataset)
+                    log.debug(dataset)
+                    datasets.append(dataset)
+                    pb.update(1)
         self.datasets = Datasets(datasets)
 
     def _spectrum_extraction(self):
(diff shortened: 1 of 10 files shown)


adonath:
Thanks @luca-giunti! I already have a few general comments:

Please move the `show_pbar` to the `__init__` methods and make it a class argument, so that it becomes configurable in future. Maybe also choose a more explicit name such as `show_progress`, or `show_progress_bar`. I would also prefer the default to be `show_progress=False`, because in many cases (and most places in our tutorials) the execution is fast and one does not really need to know the progress, if something takes just a few seconds.

In general we need to think a bit about what is useful to show as progress (energy bins, pixels, parameters, etc.) and maybe document it with a prefix / suffix to the progress bar. We also have to see what we do with nested loops.

It seems `tqdm` has an `auto` submodule, which chooses a nicer visualisation when working in notebooks:
```python
from tqdm.auto import trange
from time import sleep

for i in trange(4, desc='1st loop'):
    for j in trange(5, desc='2nd loop'):
        for k in trange(50, desc='3rd loop', leave=False):
            sleep(0.01)
```
In general I'm not sure we should introduce `tqdm` as a required dependency, maybe we just treat as optional, so when its installed, users get a progress bar if not, they don't. But Ideally we can handle this in one single place. I'll think about it once more...

luca-giunti:
> Thanks @luca-giunti! I already have a few general comments:
> 
> Please move the `show_pbar` to the `__init__` methods and make it a class argument, so that it becomes configurable in future. Maybe also choose a more explicit name such as `show_progress`, or `show_progress_bar`. I would also prefer the default to be `show_progress=False`, because in many cases (and most places in our tutorials) the execution is fast and one does not really need to know the progress, if something takes just a few seconds.
> 
Agreed!
> In general we need to think a bit about what is useful to show as progress (energy bins, pixels, parameters, etc.) and maybe document it with a prefix / suffix to the progress bar. We also have to see what we do with nested loops.

Ideed. I agree we should document the "units" of the progress (i.e. energy bin, dataset, ...) on the bar itself. It's actually easy with `tqdm(desc="")`.  I will add this.

Regarding the nested loops, in this PR I just assumed that people would always be interested on the most external layer. For example, calling `LightCurveEstimator.run()` would display the progess over time bins, but not the progress of the `stat_profile()` that is called within each single bin. I think this is a reasonable guess, but I am open to different ideas.



> It seems `tqdm` has an `auto` submodule, which chooses a nicer visualisation when working in notebooks:
> 
> ```python
> from tqdm.auto import trange
> from time import sleep
> 
> for i in trange(4, desc='1st loop'):
>     for j in trange(5, desc='2nd loop'):
>         for k in trange(50, desc='3rd loop', leave=False):
>             sleep(0.01)
> ```
I tried this, but than I tested in a jupyter lab nb and the bar was not rendered. Apparently, one needs the ipywidgets jlab extension to be installed (read the FAQ at the bottom of this [page](https://ipywidgets.readthedocs.io/en/stable/user_install.html#installing-the-jupyterlab-extension)). Is it still OK to use it?

> In general I'm not sure we should introduce `tqdm` as a required dependency, maybe we just treat as optional, so when its installed, users get a progress bar if not, they don't. But Ideally we can handle this in one single place. I'll think about it once more...

Fine, let me know about that!

In general, we have to agree on a coherent list of functions and methods that can (optionally) be called with a `show_progress_bar=True` option. For the moment, in my PR I considered:
- `Fit.stat_profile()`, `Fit.stat_contour()` methods
- the following estimators: `ASmoothMapEstimator`, `ExcessMapEstimator`, `ExcessProfileEstimator`, `FluxPointsEstimator`, `LightCurveEstimator`, `ParameterEstimator`
- the following `Analysis` methods: `._map_making()`, `_spectrum_extraction`

These are basically the places where I could spot potentially time consuming `for` loops. If you have others in mind, please comment. I also thought about adding `Datastore.get_observations()` to the list, but finally I didn't, as in general it should be fast.

Bultako:
I have found [Enlighten](https://python-enlighten.readthedocs.io) which solves the logging breaks. 

The pros:
- it is an out of the box solution that does not require touching lines of code across multiple scripts using logging,  just replacing `tdqm` by `enlighten`
- actually it is the only solution that I've found so far..

The cons:
- it does not work in notebooks :(

adonath:
@luca-giunti I have spend some time researching as well and I couldn't find a solution either. The best reference I found so far was https://github.com/tqdm/tqdm/issues/313

Bultako:
my 2 cents about this..

- logging breaks
I do not see an easy solution, apart from Enlighten the only solution that I see is unifying the direction output of all loggers under a unique handler that would stream out the logs with `tdqm.write()` method. This would require some work because we create a logger at the beginning of each script that logs with `log = logging.getLogger(__name__)`. There's more, I think it is better to provide the possibility to keep different handlers, as we do now, which would allow us to send gammapy and dependency libraries to different streams, as an example. If there are logs that continuously break the progress bar I would just modify the level of these logs and send them as a lower level or modify my user log level to filter them out and receive only info logs (we have done a similar thing when we disable astropy or numpy warnings in the past). If it is an error log I think it is normal that it breaks the bar, if it is a warning and it is really noisy we can filter them out, if it is an error log should be a warning, just modif the level it has been set in the gammapy code.

- behavior of progress bar
`tqdm` provides a `disable` flag to not display the bar, so the `dummy` class in the context manager is not needed.

- pbar.py script
as a cause of the previous item, I think this `pbar.py` script  is not really needed and we could call `tqdm` directly instead of the `pbar` context manager.

adonath:
I fully agree with you here @Bultako. I think redirecting the output is not a good solution, we'd like to keep full functionality for the logging. I think the best thing we can do for now is reduce the number of warnings (see e.g #139, #3006) and allow in some cases that they break the progress bar...

adonath:
@luca-giunti Do you see any chance of getting this into Gammapy soon?

luca-giunti:
@adonath I'm on it. I'll open a new PR cause this is full of merge conflicts at this point

adonath:
Superseded by #3343.

PR #3060, opened 2020-10-12
Remove code duplication between MapDataset and SpectrumDataset

adonath:
**Description**

This pull request makes the `SpectrumDataset` inherit from the `MapDataset` and therefore removes ~340 lines of duplicated  code. There are some little caveats:
- A `RegionNDSpectrum.get_spectrum()` method, which does nothing and returns `self`
- A `RegionNDSpectrum.cutout()` method, which does nothing and returns `self`
- `SpectrumDataset.from_hdulist()` and `SpectrumDataset.to_hdulist()` does not work  (has never worked...)
- `SpectrumDataset.read()` and `SpectrumDataset.write()` does not work (has never worked...)
- There is now a `SpectrumDataset.psf` attributed, which is fixed to `None`

For unknown reasons the error estimates for some fit parameters in the test changed. The change is in order of a few precent, which I think is acceptable and does not require further investigation.


adonath:
@Bultako There is one remaining problem here: with making `SpectrumDataset` inherit from `MapDataset` the docstring is inherited as well. Is there any possibility to insert the name of the class in the doctoring at runtime or via a Sphinx reference?

adonath:
@Bultako The basic issue is the following:
```python3
class Parent:
    """Parent class"""
    def __init__(self, x):
        self.x = x
    
    def add(self, y):
        """Add a number
        
        Returns
        -------
        parent : `Parent`
            Parent class
        """
        return self.__class__(self.x + y)
    

    
class Child(Parent):
    """Child class"""
    def multiply(self, y):
        """Multiply a number
        
        Returns
        -------
        child : `Child`
            Child class
        """
        return self.__class__(self.x + y)
      
child = Child(x=42)
print(child.add.__doc__)
print(type(child.add(4)))
```
In the unmodified inherited `Child.add` method, there still appears the `Parent` class in the docstring, even if a `Child` object is returned. Ideally we adapt the dosctring, but I guess that's not really possible, without re-implementing it. Of course we could also adapt the parent docstring to mention all possible sub-types to be returned...

Bultako:
@adonath 
I do not get the point
Which behavior you would like to have?
The `add` method and docstring **hard-coded text** are only defined in the `parent`
The output below seems legit to me.

```
In [1]: print(child.add.__doc__)
Add a number
        
        Returns
        -------
        parent : `Parent`
            Parent class

In [2]: print(child.__doc__)
Child class
```

adonath:
@Bultako Indeed you are right, the docstring is hard-coded to the class / method definition. There is probably no (reasonable) way to change it afterwards for the sub-classes. I guess a it's acceptable if we just mention the sub-class in the docstring of the parent. I'm going to implement this and then merge...

PR #3062, opened 2020-10-12
Add option to remove default output in models yaml file

QRemy:
Add option full_output (True by default) to model.to_yaml() and .to_dict(). Switching to False remove the entries for min, max, frozen and error if they are the same than the class defaults or nan.

### DIFF
--- a/gammapy/modeling/models/core.py
+++ b/gammapy/modeling/models/core.py
@@ -115,7 +115,7 @@ class Model:
             names = self.parameters.names
             for k, name in enumerate(names):
                 init = base.__dict__[name].to_dict()
-                for item in ["min", "max", "error"]:
+                for item in ["min", "max", "frozen", "error"]:
                     if params[k][item] == init[item] or np.isnan(init[item]):
                         del params[k][item]
         return {"type": tag, "parameters": params}
--- a/gammapy/modeling/models/core.py
+++ b/gammapy/modeling/models/core.py
@@ -105,10 +105,20 @@ class Model:
         """A deep copy."""
         return copy.deepcopy(self)
 
-    def to_dict(self):
+    def to_dict(self, full_output=True):
         """Create dict for YAML serialisation"""
         tag = self.tag[0] if isinstance(self.tag, list) else self.tag
-        return {"type": tag, "parameters": self.parameters.to_dict()}
+        params = self.parameters.to_dict()
+
+        if full_output is False:
+            base = self.__class__
+            names = self.parameters.names
+            for k, name in enumerate(names):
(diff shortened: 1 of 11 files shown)


gammapy/modeling/models/core.py
```diff
@@ -105,10 +105,20 @@ def copy(self):
...
         return copy.deepcopy(self)
 
-    def to_dict(self):
+    def to_dict(self, full_output=True):
         """Create dict for YAML serialisation"""
         tag = self.tag[0] if isinstance(self.tag, list) else self.tag
-        return {"type": tag, "parameters": self.parameters.to_dict()}
+        params = self.parameters.to_dict()
+
+        if full_output is False:
```
adonath, line 113: Please just use `if not full_output:` here. The `is` pattern kind of works accidentally here...

gammapy/modeling/models/core.py
```diff
@@ -105,10 +105,20 @@ def copy(self):
         """A deep copy."""
         return copy.deepcopy(self)
 
-    def to_dict(self):
+    def to_dict(self, full_output=True):
```
adonath, line 108: I think for writing we should change the default behaviour to `full_output= False`...

adonath (dismissed):
Thanks @QRemy, I have left two minor inline comments...

gammapy/modeling/models/core.py
QRemy: If I change for full_output = False as default then I would not apply simplified output to frozen as it too important to be hidden by default.

adonath: I think assuming a parameter is not frozen by default is fine...that's what we do in the `Parameter` class as well.

QRemy: ok

adonath:
Thanks @QRemy, the output looks much cleaner now:
```
components:
-   name: pwl-gauss-model
    type: SkyModel
    spectral:
        type: PowerLawSpectralModel
        parameters:
        -   name: index
            value: 2.0
            unit: ''
        -   name: amplitude
            value: 1.0e-12
            unit: cm-2 s-1 TeV-1
        -   name: reference
            value: 1.0
            unit: TeV
    spatial:
        type: GaussianSpatialModel
        frame: icrs
        parameters:
        -   name: lon_0
            value: 0.0
            unit: deg
        -   name: lat_0
            value: 0.0
            unit: deg
        -   name: sigma
            value: 1.0
            unit: deg
        -   name: e
            value: 0.0
            unit: ''
        -   name: phi
            value: 0.0
            unit: deg

```

adonath (approved):
No further comment from my side...

PR #3063, opened 2020-10-12
Add PiecewiseNormSpectralModel

QRemy:
Add a norm spectral model that parametrises a piecewise broken power law 
with a free norm parameter at each fixed energy node.


gammapy/modeling/models/spectral.py
```diff
@@ -857,6 +858,89 @@ def evaluate(energy, index1, index2, amplitude, ebreak, reference, beta):
...
+
+    def to_dict(self):
+        return {
+            "type": self.tag[0],
+            "parameters": self.parameters.to_dict(),
+            "energy": {
+                "data": self.energy.data.tolist(),
+                "unit": str(self.energy.unit),
+            },
+            "norms": {"data": self.norms.data.tolist(), "unit": str(self.norms.unit),},
```
adonath, line 930: The duplication of the norm values, here doesn't make sense to me. I think it's rather confusing then it helps (I can see how it might be shorter to define a list of norms, but it misses the extra information...). I would remove this....

gammapy/modeling/models/spectral.py
```diff
@@ -857,6 +858,89 @@ def evaluate(energy, index1, index2, amplitude, ebreak, reference, beta):
...
+        if norms is None:
+            norms = np.ones(len(energy))
+        if len(norms) != len(energy):
+            raise ValueError("dimension mismatch")
+        if len(norms) < 2:
+            raise ValueError("Input arrays must contians at least 2 elements")
+        parameters = Parameters(
+            [Parameter(f"norm{k}", norm) for k, norm in enumerate(norms)]
+        )
+        self._set_default_parameters(parameters)
```
adonath, line 890: Passing the parameters to `super().__init__()` should work and remove the need for the `_set_default_parameters`

gammapy/modeling/models/spectral.py
```diff
@@ -857,6 +858,89 @@ def evaluate(energy, index1, index2, amplitude, ebreak, reference, beta):
...
+    """ Piecewise broken power-law spectral model
+       with renormalization at fixed energy nodes.
+       
+       For more information see :ref:`piecewise-broken-powerlaw-norm-spectral`.
+
+    Parameters
+    ----------
+    energy : `~astropy.units.Quantity`
+        Array of energies at which the model values are given (nodes).
+    norms : array
```
adonath, line 871: I think it should be possible to pass a list of `Parameter` objects as well...

gammapy/modeling/models/spectral.py
```diff
@@ -857,6 +858,89 @@ def evaluate(energy, index1, index2, amplitude, ebreak, reference, beta):
...
+            raise ValueError("dimension mismatch")
+        if len(norms) < 2:
+            raise ValueError("Input arrays must contians at least 2 elements")
+        parameters = Parameters(
+            [Parameter(f"norm{k}", norm) for k, norm in enumerate(norms)]
+        )
+        self._set_default_parameters(parameters)
+
+    @classmethod
+    def from_parameters(cls, parameters, **kwargs):
```
adonath, line 893: Once you allow to take `Parameter` objects on init as well, this should be obsolete...

gammapy/modeling/models/spectral.py
```diff
@@ -857,6 +858,89 @@ def evaluate(energy, index1, index2, amplitude, ebreak, reference, beta):
...
+    def __init__(self, energy, norms=None):
+        self._energy = energy
+        if norms is None:
+            norms = np.ones(len(energy))
+        if len(norms) != len(energy):
+            raise ValueError("dimension mismatch")
+        if len(norms) < 2:
+            raise ValueError("Input arrays must contians at least 2 elements")
+        parameters = Parameters(
+            [Parameter(f"norm{k}", norm) for k, norm in enumerate(norms)]
```
adonath, line 888: Minor: add and underscore before the index like `norm_0`, `norm_1`, etc.

adonath (changes requested):
Thanks @QRemy, I have already left some comments. I'll now check out the branch and play around with it a bit...

gammapy/modeling/models/spectral.py
QRemy: _set_default_parameters can be removed but from_parameters still need to be redefined

gammapy/modeling/models/spectral.py
```diff
@@ -857,6 +858,87 @@ def evaluate(energy, index1, index2, amplitude, ebreak, reference, beta):
...
+
+    def __init__(self, energy, norms=None, parameters=None):
+        self._energy = energy
+        if norms is None:
+            norms = np.ones(len(energy))
+        else:
+            if len(norms) != len(energy):
+                raise ValueError("dimension mismatch")
+            if len(norms) < 2:
+                raise ValueError("Input arrays must contians at least 2 elements")
```
registerrier, line 890: contain

registerrier (commented):

I have a few general questions/suggestions:
- Since what is defined are really power laws (i.e. no spectral indices given) but really a piecewise function in loglog space, couldn't it be simpler and easier to understand to call it a `PiecewiseNormSpectralModel`?  Possibly one could add options to have log-log or lin-lin interpolations.
- I don't understand the approach passing both `norm` and `parameters`. I would assume the default `__init__` to take a list of parameters. Then a class method `form_norms` could be introduced.
- Is it possible to have energies as parameters as well?

QRemy:
Thanks @registerrier 
- I agree with the name change for `PiecewiseNormSpectralModel`. Is it really a use case for lin-lin interp ?
- My initial idea was to have energy and norms as input because it simpler, and passing parameters would be the alternative init as anyway the from_parameters method has to be redefine to work with the serialization. @adonath suggested to have also parameters as input to simply from_parameters method.
- Energies could be parameters (as in the ctools) but it will be very hard for the fit to converge, it's likely more stable to twice the number of nodes to have more precision with the same number of parameters. For the implementation we could set internally the min/max to avoid energies inversion/overlapping but that may be in conflict with the min/max settings by the users. I would prefer keep fixed nodes.

adonath:
@QRemy For the interpolation you could rely on https://github.com/gammapy/gammapy/blob/master/gammapy/utils/interpolation.py#L118, the users have any option they like...

adonath (approved):
Thanks @QRemy, I'll go ahead and merge and fix remaining test fails in master...

PR #3065, opened 2020-10-13
Document stat_contour in modeling tutorial

luca-giunti:
**Description**
I document the use of `Fit.stat_surface()` in the modeling and fitting tutoria

**Dear reviewer**
I personally feel this tutorial is one of the most relevant in the gammapy docs, as in the end it's all about modeling and fitting and inspecting results. Yet it's almost hidden within the advanced tutorial sections...


adonath:
@luca-giunti There is still one test fails on Travis-Ci please fix...

docs/tutorials/modeling.ipynb
```diff
@@ -55,6 +55,7 @@
     "import numpy as np\n",
```
registerrier, line 55: <p>I would only mention reoptimize=False in the bullet list, and really conclude the cell with a strong statement on the correct approach for confidence contours.</p><p>It is important to remind that the contours are approximative.  In particular, when the parameter range boundaries are close to the contours. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/gammapy/gammapy/pull/3065/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='44'/>

registerrier, line 55: <p>Is this really needed? Can't we simply explain that given the normalization of our TS (-2 log L, i.e. chi square like) we expect the 1 sigma contour to be at +1, 2 sigma at +4 and 3 sigma +9?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/gammapy/gammapy/pull/3065/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='47'/>

luca-giunti: @registerrier There are 2 floating degrees of freedom in the `stat_surface` method, so one has to look at the percentiles of a chi square with df=2. The sqrt rule of thumb to connect TS and sigma levels doesn't apply in that case. No?

luca-giunti: <p>I actually see a bug here in my code, I should change <code>df=1</code>  to <code>df=2</code>  ...</p>

luca-giunti: test

I tested that the contour computed by `Fit.minos_contour` and `Fit.stat_surface` (with `reoptimize=True`!!) are consistent. In the figure, the contours are slightly off but this is due to the interpolation made by `plot_contour_line`: the crosses (a bit hard to see, sorry), representing the actual points computed by MINOS, follow exactly the contours computed by `fit.stat_profile`. Notice that, in the tutorial nb, since we have `reoptimize=False` there is a slight difference in the contours... but that's understood. 

This means that using a chi2 with df=2 to compute the significance from `Fit.stat_surface` is coherent with `Fit.minos_contour`...

registerrier: OK. Good!

Then my main suggestion would rather be to keep delta TS values in the array but simply show how to compute the corresponding levels. That seems easier since it does not require array manipulation.

registerrier: It is probably the nicest solution, but how to proceed in general should be documented somewhere

luca-giunti: OK. Maybe @adonath you have a preference?

luca-giunti: Apologies @registerrier , I messed up here. As I commented above, in the tutorial I was using `df=1`, and since I forgot to change it the contour comparison was still obtained with `df=1`. So apparently it is sufficient to take the sqrt(TS) to be consistent with `minos_contour`... this will simplify the tutorial cell without the need for a dedicated function in `gammapy.visualization`. 

I will add a short explanation in the tutorial...

adonath (approved):
Thanks @luca-giunti, I have no further comments....I'll go ahead and merge now. If there are remains things, please open a new PR...

PR #3066, opened 2020-10-14
Add option to preserve counts in Map.interp_to_geom()

LauraOlivera:
**Description**

This pull request adds an option `preserve_counts` to the method `Map.interp_to_geom()`. It is then used in the data reduction step of a `MapDataset` for the background, in case the input is a `Map`.

It also makes it optional to have an exposure map for the `edisp` and `psf` in a dataset when saving it to a file (this was missing from PR [#2990](https://github.com/gammapy/gammapy/pull/2990))

### DIFF
--- a/gammapy/maps/core.py
+++ b/gammapy/maps/core.py
@@ -774,6 +774,7 @@ class Map(abc.ABC):
         interp_map : `Map`
             Interpolated Map
         """
+
         coords = geom.get_coord()
 
         # set nearest neighbour interpolation for mask as default
@@ -781,6 +782,8 @@ class Map(abc.ABC):
             kwargs.setdefault("interp", 0)
 
         if preserve_counts:
+            if geom.ndim > 2:
+                assert self.geom.axes[0] == geom.axes[0] # Energy axis has to match
             old_map_copy = self.copy()
             old_map_copy.data /= self.geom.solid_angle().to_value("deg2")
             data = old_map_copy.interp_by_coord(coords, **kwargs)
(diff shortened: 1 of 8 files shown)


gammapy/maps/core.py
```diff
@@ -776,11 +780,16 @@ def interp_to_geom(self, geom, **kwargs):
         if self.data.dtype == bool:
             kwargs.setdefault("interp", 0)
 
-        data = self.interp_by_coord(coords, **kwargs)
+        if preserve_counts:
+            old_map_copy = self.copy()
+            old_map_copy.data /= self.geom.solid_angle().to_value("deg2")
```
adonath, line 785: I just realised, that the bins size in energy could change as well. So we probably need a more general solution like:
```python3
bin_volume = self.geom.bin_volume.value
old_map_copy.data /= bin_volume

data *= geom.bin_volume.to_value(bin_volume.unit)

```

adonath (changes requested):
Thanks @LauraOlivera, I have left one inline comment...otherwise looks good to me.

gammapy/maps/core.py
adonath: Conclusion from the dev meeting today: keep the implementation, but add a check on the equality of the energy axis. This simply works like `energy_axis_1 == energy_axis_2`...

LauraOlivera: done! :+1:

adonath (approved):
Thanks a lot @LauraOlivera! No further comments from my side...

PR #3067, opened 2020-10-14
Fix documentation for gammapy.irf

cosimoNigro:
I have been using the `gammapy.irf` classes a lot lately and noticed the documentation was incomplete and 
contained references to classes that do not exist anymore.

All the `.rst` files containing the irf components description have been updated. 
I linked every IRF components to the corresponding description in the OGADF format and provided a snippet of code
opening and visualising the component of a run of the HESS DL3 DR1 
(I think more detailed functionalities are anyhow covered by the API and the tutorial notebooks).

Let me know if the new irf documentation structure sits well with you, if it is too little or to much detailed.
I am also not familiar with the PSF classes so if something has to be improved there let me know. 

Thank you,

Cosimo

### DIFF
--- a/docs/irf/bkg.rst
+++ b/docs/irf/bkg.rst
@@ -10,7 +10,7 @@ of detector coordinates and true energy.
 
 Its format specifications are available in :ref:`gadf:bkg_3d`.
 
-This is the format in which IACT DL3 energy dispersions are usually provided, as an example:
+This is the format in which IACT DL3 background rates are usually provided, as an example:
 
 .. plot:: irf/plot_bkg_3d.py
     :include-source:
--- a/docs/irf/bkg.rst
+++ b/docs/irf/bkg.rst
@@ -26,5 +26,3 @@ Its format specifications are available in :ref:`gadf:bkg_2d`.
 You can produce a radially-symmetric background from a list of DL3 observations 
 devoid of gamma-ray signal using the tutorial in 
 `background_model.html <../tutorials/backgorund_model.html>`__ 
-
-
--- a/docs/irf/bkg.rst
+++ b/docs/irf/bkg.rst
@@ -1,0 +1,30 @@
+.. _irf-bkg:
+
+Background
+==========
+
+as a function of of true energy and detector coordinates (:ref:`gadf:bkg_3d`) 
+-----------------------------------------------------------------------------
+The `~gammapy.irf.Background3D` class represents a background rate as a function
+of detector coordinates and true energy. 
+
+Its format specifications are available in :ref:`gadf:bkg_3d`.
+
+This is the format in which IACT DL3 energy dispersions are usually provided, as an example:
+
+.. plot:: irf/plot_bkg_3d.py
+    :include-source:
+
+
+as a function of of true energy and offset angle, radially symmetric (:ref:`gadf:bkg_2d`)
+-----------------------------------------------------------------------------------------
+The `~gammapy.irf.Background2D` class represents a background rate as a function 
+of offset angle from the field of view center and true energy.
+
+Its format specifications are available in :ref:`gadf:bkg_2d`.
+
+You can produce a radially-symmetric background from a list of DL3 observations 
+devoid of gamma-ray signal using the tutorial in 
+`background_model.html <../tutorials/backgorund_model.html>`__ 
+
+
--- a/docs/irf/bkg.rst
+++ b/docs/irf/bkg.rst
@@ -3,10 +3,10 @@
 Background
 ==========
 
-as a function of of true energy and detector coordinates (:ref:`gadf:bkg_3d`) 
------------------------------------------------------------------------------
-The `~gammapy.irf.Background3D` class represents a background rate as a function
-of detector coordinates and true energy. 
+as a function of reconstructed energy and detector coordinates (:ref:`gadf:bkg_3d`) 
+--------------------------------------------------------------------------------------
+The `~gammapy.irf.Background3D` class represents a background rate per solid 
+angle as a function of detector coordinates and reconstructed energy. 
 
 Its format specifications are available in :ref:`gadf:bkg_3d`.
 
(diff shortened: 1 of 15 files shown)


docs/irf/psf.rst
```diff
@@ -1,10 +1,42 @@
 .. _irf-psf:
 
-Point Spread Function
```
adonath, line 3: Minor: maybe keep the explicit title? It's maybe easier to read...

docs/irf/bkg.rst
```diff
@@ -0,0 +1,30 @@
...
+
+Its format specifications are available in :ref:`gadf:bkg_3d`.
+
+This is the format in which IACT DL3 energy dispersions are usually provided, as an example:
+
+.. plot:: irf/plot_bkg_3d.py
+    :include-source:
+
+
+as a function of of true energy and offset angle, radially symmetric (:ref:`gadf:bkg_2d`)
```
adonath, line 19: See above...

docs/irf/bkg.rst
```diff
@@ -0,0 +1,30 @@
...
+==========
+
+as a function of of true energy and detector coordinates (:ref:`gadf:bkg_3d`) 
+-----------------------------------------------------------------------------
+The `~gammapy.irf.Background3D` class represents a background rate as a function
+of detector coordinates and true energy. 
+
+Its format specifications are available in :ref:`gadf:bkg_3d`.
+
+This is the format in which IACT DL3 energy dispersions are usually provided, as an example:
```
adonath: I think you mean background here...

docs/irf/edisp.rst
```diff
@@ -1,11 +1,36 @@
...
+Its format specifications are available in :ref:`gadf:edisp_2d`
+
+This is the format in which IACT DL3 energy dispersions are usually provided, as an example:
 
 .. plot:: irf/plot_edisp.py
+    :include-source:
+
+as a function of true energy (:ref:`gadf:ogip-rmf`)
+---------------------------------------------------
+`~gammapy.irf.EDispKernel` instead represents an energy dispersion as a function of true energy only
```
adonath, line 21: There is a bit more to say here: the `EDispKernel` is also integrated along the migration axis, to get the RMF matrix with a given "reco" energy binning....

docs/irf/psf.rst
```diff
@@ -1,10 +1,42 @@
...
-* `~gammapy.irf.PSF3D`
+This is the format in which IACT DL3 PSFs are usually provided, as an example:
 
+.. plot:: irf/plot_psf.py
+    :include-source:
+
+
+as a function of true energy (:ref:`gadf:psf_gtpsf`)
+----------------------------------------------------
+`~gammapy.irf.EnergyDependentTablePSF` instead represents an energy dispersion as a function of true energy only
```
adonath, line 21: Typo: I guess you mean PSF here...

docs/irf/psf.rst
```diff
@@ -1,10 +1,42 @@
...
+
+Its format specifications are available in :ref:`gadf:psf_gtpsf`.
+
+Such a PSF can be obtained by evaluating the `~gammapy.irf.PSF3D` at a given offset (using `~astropy.coordinates.Angle`)
+
+.. plot:: irf/plot_psf_table.py
+    :include-source:
+
+**Note:** the offset angle on the x axis is not an offset angle from the field of view center but the distance between 
+the source position and the reconstructed gamma ray position.
```
adonath, line 32: For this we have introduced the convention to name it "rad", could be mentioned here...

adonath (changes requested):
Thanks @cosimoNigro, I have a few first minor comments already. It would be good if @registerrier could do another review...

registerrier (changes requested):
Thanks a lot @cosimoNigro !

This looks good. I have proposed inline a few changes to clarify some parts.

cosimoNigro:
Hello @adonath and @registerrier,

I have implemented your comments.

As for the script for the 2D background generation from a list of OFF, shall I explicitly mention that it will be moved to the contributed notebooks or I just leave it like this and then we modify the link to it when we actually move it?

cosimoNigro:
Another thing, the notation is a bit mixed: 
- in the IRF implementation description we use (E_true, E_reco), (p_true, p_reco);
- in the IRF theory we mix subscripted and non-subscripted (E, E_reco), (p, p_reco);
which one should we use?
I think if the reconstructed quantities are subscripted also the true quantities should be, for uniformity, so it should be
`A_{\rm eff}(E_{\rm true}, p_{\rm true})`
`PSF(P_{\rm recdo} | E_{\rm true}, p_{\rm true})`


what do you suggest? If we go for (E, E_reco), (p, p_reco) then the migration reads `\mu = E_{\rm reco} / E` and the rad reads `\delta p = p_{\rm true} - p`. But the integrals will have ugly differentials `d p_{\rm true}`.

Another option is priming the estimated quantities, so (E, E'), (p, p'). Though this does not relate and recall the axis naming convention.

Which one then?
- [ ] subscripted (E_true, E_reco), (p_true, p_reco);
- [ ] primed (E, E'), (p, p');
- [ ] mixed (E, E_reco), (p, p_reco) - as is now.

cosimoNigro:
For the moment I reformatted everything to follow the third option: `(E, E_reco), (p, p_reco)`.
I removed all the `_{\rm true}` subscripts in the IRF implementation description.

adonath:
Thanks @cosimoNigro! In the axis naming convention we use `energy` and `energy_true`, so maybe use `E` and `E_true` (or shorter like `E_t`) for the formulae? I'm not saying this is the clearest option, but the consistency with axis naming might be a plus. But I don't feel strongly there...

You can keep the reference to the background notebook, we will change it once the notebook actually is moved over....

cosimoNigro:
Hi @adonath,

I have subscrtipted the true quantity `(E_{\rm true}, p_{\rm true})` and left unsubscripted the estimated one.
I added in the AXIS naming convention a reference to the notation used in theory.

I am good to merge if you do not have further requests.

adonath:
@cosimoNigro Perfect, thanks a lot! Let's wait for the CI builds and then merge...

adonath (approved):
Thanks again @cosimoNigro, no further comments from my side...

adonath:
Here are the rendered docs: https://docs.gammapy.org/dev/irf/index.html

PR #3073, opened 2020-10-21
Print parameter values to 4 significant digits

AtreyeeS:
**Description**

Currently, `Parameters.to_table()` prints the values of the parameters to 3 significant places.
This is sometimes annoying for the longitude values, eg: `3.324e+02` where we do not know what the second value place is.

This is a small PR to fix the same.

**Dear reviewer**


adonath (approved):
Thanks @AtreyeeS, makes sense...

PR #3075, opened 2020-10-23
Proposal for addition of a FluxMap container

registerrier:
**Description**

This pull request presents a possible `FluxMap` container to work with products of flux `Estimators`. It can be used for flux points or maps thanks to the `Map` framework. It can be used to convert to regular `FluxPoints`.
It contains maps in the 'likelihood' SED type (i.e. reference flux, and norm) and a reference model to allow for sed type conversions. The latter are provided by a set of properties e.g. `fluxmap.eflux_err`.

The class also offers the option to serialize the product to disk in fits, through an export to hdulist. A specific keyword is added in the primary header to specify the SED type. The export is based on `to_dict` and `from_dict` methods that allow to perform exports/imports of the maps to/from any sed type.

The major issue here is of course that we rely on a model but we have not yet a proper way to export it along with the fits maps. Unless we also export a yaml file containing the link to the the fits file.

**Dear reviewer**

Please comment. I can also modify `Estimators` to have them use the `FluxMap` to clarify its usage.


gammapy/estimators/flux_map.py
```diff
@@ -0,0 +1,480 @@
...
+        norm_errp=None,
+        norm_ul=None,
+        norm_scan=None,
+        stat_scan=None,
+        ts=None,
+        counts=None,
+        ref_model=None,
+    ):
+        self.geom = norm.geom
+        self.dnde_ref = dnde_ref
```
adonath, line 99: Isn't `dnde_ref` duplicated information with the `ref_model` here? If we need to represent an array of dnde values, I guess we can so as well just with the `TemplateSpectralModel`

gammapy/estimators/flux_map.py
```diff
@@ -0,0 +1,480 @@
...
+
+OPTIONAL_MAPS = {
+    "dnde": ["dnde_err", "dnde_errp", "dnde_errn", "dnde_ul", "is_ul"],
+    "e2dnde": ["e2dnde_err", "e2dnde_errp", "e2dnde_errn", "e2dnde_ul", "is_ul"],
+    "flux": ["flux_err", "flux_errp", "flux_errn", "flux_ul", "is_ul"],
+    "eflux": ["eflux_err", "eflux_errp", "eflux_errn", "eflux_ul", "is_ul"],
+    "likelihood": ["norm_err", "norm_errn", "norm_errp","norm_ul", "norm_scan", "stat_scan"],
+}
+
+class FluxMap:
```
adonath, line 42: In general If fell like  the plural `FluxMaps` is maybe more appropriate here, but I can see e.g. in the `IRFMap` we use singular as well, even if it contains in fact two maps....

gammapy/estimators/flux_map.py
```diff
@@ -0,0 +1,480 @@
...
+        self.dnde_ref = dnde_ref
+        self.norm = norm
+        self.norm_err = norm_err
+        self.norm_errn = norm_errn
+        self.norm_errp = norm_errp
+        self.norm_ul = norm_ul
+        self.norm_scan = norm_scan
+        self.stat_scan = stat_scan
+        self.ts = ts
+        self.counts = counts
```
adonath, line 108: Here I'm not a 100% whether we should include this information on the raw data, if we do so, I guess it means we have to include things like `excess` and `npred` as well. However these are then typically duplicated information from the `MapDataset` (with adjusted energy binning...)

gammapy/estimators/flux_map.py
```diff
@@ -0,0 +1,480 @@
...
+        -------
+        fluxpoints : `~gammapy.estimators.FluxPoints`
+            the flux points object
+        """
+        if coord is None:
+            coord = self.geom.center_skydir
+        energies = self.e_ref
+        coords = MapCoord.create(dict(skycoord=coord, energy=energies))
+
+        ref = self.dnde_ref.get_by_coord(coords) * self.dnde_ref.unit
```
adonath, line 257: I guess one could call `interp_by_coord()` as well here and let users pass the interpolation order...

gammapy/estimators/flux_map.py
```diff
@@ -0,0 +1,480 @@
...
+        Parameters
+        ----------
+        filename : str
+            Filename to write to.
+        overwrite : bool
+            Overwrite file if it exists.
+        sed_type : str
+            sed type to convert to. Default is `Lielihood`
+        """
+        self.to_hdulist().writeto(str(make_path(filename)), overwrite=overwrite)
```
adonath, line 336: To reduce the file size, one could remove the `BANDS` HDU here...

adonath (commented):
Thanks @registerrier, I have left a few preliminary comments here. All in all I think this a really nice feature and we should invest the time to arrive a nice consistent design with `FluxPoints`, where we are missing still missing the convenience properties such as `.dnde` etc.

gammapy/estimators/flux_map.py
registerrier: Yes it is removed

gammapy/estimators/flux_map.py
registerrier: For now all non flux maps are kept within the dictionary but are not accessible via a specific property.
They are serialized along the others as well.

gammapy/estimators/flux_map.py
registerrier: Well I think it is a bit dangerous to interpolate over flux values, errors and upper limits. A priori I would rather enforce nearest neighbor .

gammapy/estimators/flux_map.py
registerrier: Indeed, but there is currently a bug in `Maps` preventing that I think.

gammapy/estimators/flux_map.py
```diff
@@ -0,0 +1,399 @@
...
+        self.gti = gti
+
+        super().__init__(data, spectral_model=reference_model.spectral_model)
+
+    @staticmethod
+    def _default_model():
+        return SkyModel(spatial_model=PointSpatialModel(), spectral_model=PowerLawSpectralModel(index=2))
+
+    @property
+    def e_ref(self):
```
adonath, line 93: For `FluxPoints` we have introduced the more explicit `energy_ref`...make it consistent here?

gammapy/estimators/flux_map.py
```diff
@@ -0,0 +1,399 @@
...
+        coords = MapCoord.create(dict(skycoord=coord, energy=energies))
+
+        ref = self.dnde_ref
+
+        fp = dict()
+        fp["norm"] = self.norm.get_by_coord(coords) * self.norm.unit
+
+        for quantity in self._available_quantities:
+            norm_quantity = f"norm_{quantity}"
+            res = self.__getattribute__(norm_quantity).get_by_coord(coords)
```
adonath, line 135: I think you can directly use `geattr(self, norm_quantity)` to avoid the access to the internal method...

gammapy/estimators/flux_map.py
```diff
@@ -0,0 +1,399 @@
...
+        -------
+        map_dict : dict
+            dictionary containing the requested maps.
+        """
+        if sed_type == "likelihood":
+            map_dict = self.data
+        else:
+            map_dict = {}
+            for entry in REQUIRED_MAPS[sed_type]:
+                map_dict[entry] = self.__getattribute__(entry)
```
adonath, line 174: See above...

gammapy/estimators/flux_map.py
```diff
@@ -0,0 +1,399 @@
...
+        ----------
+        filename : str
+            Filename to write to.
+        filename_model : str
+            Filename of the model (yaml format).
+            If None, keep string before '.' and add '_model.yaml' suffix
+        overwrite : bool
+            Overwrite file if it exists.
+        sed_type : str
+            sed type to convert to. Default is `Lielihood`
```
adonath, line 200: "Likelihood"

gammapy/estimators/flux_map.py
```diff
@@ -0,0 +1,399 @@
...
+
+        hdulist = self.to_hdulist(sed_type)
+
+        models = Models(self.reference_model)
+        models.write(filename_model, overwrite=overwrite)
+        hdulist[0].header['MODEL'] = filename_model.as_posix()
+
+        hdulist.writeto(str(make_path(filename)), overwrite=overwrite)
+
+    def to_hdulist(self, sed_type="likelihood", hdu_bands=None):
```
adonath, line 219: I would much prefer convention over configuration here and not allow to define the name of the `BANDS` hdu, once it's introduce people will use it and modify the names as they like. Let's be strict on write but forgiving on read...

I guess the option is to allow for a single band HDU only? Maybe we just offer this behaviour, I'm not sure whether we need the case with all individual band HDUs defined. I'll think about once again...

adonath (approved):
Thanks @registerrier! I have left a few preliminary minor comments, I'll go ahead now and start to try it out...

adonath:
I started to try this out in the detect notebook and here are a few little issues I found:
- This is not related to this PR, however when I tried to do `flux_maps.get_flux_points().to_sed_type("dnde")`, I noticed the unit in the dude column was missing, I think computations with `Column` object don't preserve units, so we have to access `.quantity` here: https://github.com/gammapy/gammapy/blob/master/gammapy/estimators/flux_point.py#L393
- I think by default we should convert the `norm` quantity to unites, it came out as `keV/TeV`...
- I think you could already expose the `ts` and `sqrt_ts` maps here. If if present they are not considered in the `.get_flux_points()` method...
- When calling `.get_flux_points()`, the derived quantities come with a 2D array shaped columns in the table, which currently makes the plotting of the `FluxPoints` fail... 
- Maybe add a minimal `__str__` method, that plots the geometry, the contained quantities and the reference model?

gammapy/estimators/flux_map.py
registerrier: Changed to FluxMaps

gammapy/estimators/flux_map.py
registerrier: Done

gammapy/estimators/flux_map.py
registerrier: Done

gammapy/estimators/flux_map.py
registerrier: Done

registerrier:
Thanks @adonath .

> * This is not related to this PR, however when I tried to do `flux_maps.get_flux_points().to_sed_type("dnde")`, I noticed the unit in the dnde column was missing, I think computations with `Column` object don't preserve units, so we have to access `.quantity` here: https://github.com/gammapy/gammapy/blob/master/gammapy/estimators/flux_point.py#L393
> * I think by default we should convert the `norm` quantity to unity, it came out as `keV/TeV`...

This is done in `FluxMaps.from_dict` and a test is added.

> * I think you could already expose the `ts` and `sqrt_ts` maps here. If if present they are not considered in the `.get_flux_points()` method...
Done. They might even be added on the base class `FluxEstimate` actually.

> * When calling `.get_flux_points()`, the derived quantities come with a 2D array shaped columns in the table, which currently makes the plotting of the `FluxPoints` fail...


> * Maybe add a minimal `__str__` method, that plots the geometry, the contained quantities and the reference model?

Done.

adonath:
Thanks a lot @registerrier, everything works fine now. Is this ready to merge from your side?

registerrier:
The only remaining question is the serialization of the bands HDU. I have tried to store it as a single band but I have an issue for now. Otherwise everything is fine.
We might merge and solve this in a follow up PR.

adonath (approved):
Thanks @registerrier, no further comment from my side.

PR #3081, opened 2020-10-25
Unify axis ordering in gammapy.irf

AtreyeeS:
**Description**

This pull request fixes the axes order in gammapy.irf

**Dear reviewer**


adonath:
Thanks a lot @AtreyeeS! In #3067 @cosimoNigro added some documentation to `gammapy.irf`. Maybe we take the chance and document the internal axis order, that you now implemented for the classes as well? So just adding one sentence per class to the following pages:
- https://docs.gammapy.org/dev/irf/aeff.html#irf-aeff
- https://docs.gammapy.org/dev/irf/edisp.html#irf-edisp
- https://docs.gammapy.org/dev/irf/psf.html#irf-psf
- https://docs.gammapy.org/dev/irf/bkg.html#irf-bkg

AtreyeeS:
Hi! The doc-strings already mention (some) axis oder - I have put them in order now. 
This part of the docs read quite well :)

adonath:
Thanks a lot @AtreyeeS! I used a little script to check the axis order. Here are the results:

Before:
```
Effective area: 

	Axes names:  ['energy_true', 'offset']
	Axes shape:  (96, 6)
	Data shape:  (96, 6)

Energy dispersion: 

	Axes names:  ['energy_true', 'migra', 'offset']
	Axes shape:  (96, 160, 6)
	Data shape:  (96, 160, 6)

Point spread function: 

	Axes names:  ['rad', 'offset', 'energy_true']
	Axes shape:  (144, 6, 32)
	Data shape:  [144, 6, 32]

Background: 

	Axes names:  ['energy', 'fov_lon', 'fov_lat']
	Axes shape:  (20, 50, 50)
	Data shape:  (20, 50, 50)

```

After:
```
Effective area: 

	Axes names:  ['energy_true', 'offset']
	Axes shape:  (96, 6)
	Data shape:  (96, 6)

Energy dispersion: 

	Axes names:  ['energy_true', 'migra', 'offset']
	Axes shape:  (96, 160, 6)
	Data shape:  (96, 160, 6)

Point spread function: 

	Axes names:  ['energy_true', 'offset', 'rad']
	Axes shape:  (32, 6, 144)
	Data shape:  [32, 6, 144]

Background: 

	Axes names:  ['energy', 'fov_lon', 'fov_lat']
	Axes shape:  (20, 50, 50)
	Data shape:  (20, 50, 50)

```

The recommended axis orders in GADF are:
- [Effective area 2D](https://gamma-astro-data-formats.readthedocs.io/en/latest/irfs/full_enclosure/aeff/index.html#aeff-2d) ENERGY, THETA
- [EDISP 2D](https://gamma-astro-data-formats.readthedocs.io/en/latest/irfs/full_enclosure/edisp/index.html#edisp-2d): ENERGY, MIGRA, THETA
- [PSF 3D](https://gamma-astro-data-formats.readthedocs.io/en/latest/irfs/full_enclosure/psf/psf_table/index.html#psf-table): ENERGY, THETA, RAD.
- [Background 3D](https://gamma-astro-data-formats.readthedocs.io/en/latest/irfs/full_enclosure/bkg/index.html#bkg-3d): ENERGY, DETX, DETY

So it seems only the PSF 3D was transposed, but is fixed now. The numpy axis order is now compliant with the recommended order in GADF. From what I can tell it's also equivalent now to what is used in pyirf. Can you please confirm @maxnoe?

adonath:
Looking a bit closer there are a few weird things:
- The in memory order is now compliant with GADF (as shown in my post before)
- But to arrive at this order we need to transpose the data on read for: [`EnergyDispersion2D`](https://github.com/AtreyeeS/gammapy/blob/axes/gammapy/irf/energy_dispersion.py#L154), [`EffectiveArea2D`](https://github.com/AtreyeeS/gammapy/blob/axes/gammapy/irf/effective_area.py#L401) and [`PSF3D`](https://github.com/AtreyeeS/gammapy/blob/axes/gammapy/irf/psf_3d.py#L121)
- But not for [`Background3D`](https://github.com/AtreyeeS/gammapy/blob/axes/gammapy/irf/background.py#L112)

I guess this suggest that the axis order is written inconsistently in the background files? 

I think we should systematically check once the reference files provided in the GADF repo...

maxnoe:
Data should always need to be transposed as the FITS axis order convention is FORTRAN, vs. numpy's is C like.

adonath:
Thanks @maxnoe, yes that's what I thought too. However it doesn't seem to be the case for the background models, we currently use for tutorials and testing (HESS DL3 DR1 and CTA 1DC)...

AtreyeeS:
Hi @adonath ! I was worried about `Background3D` as well.

The example file given in GADF has what is written (energy, det_x, deg_y)

However, the HESS-DL3-DR1 fits release and the CTA 1DC IRFs have the axis order reversed in the fits files (deg_x, deg_y, energy).

adonath:
Thanks @AtreyeeS for confirming. We should definitely add unit tests (but in an another PR...) for the example files given in the GADF repo (not sure why we haven't already...). Then we have to see how we handle both cases in our `Background3D.from_table()` methods...

maxnoe:
I have seen some files having a `CREF` header attribute specifying axis order like so:

```
CREF    = (ENERG,THETA,FOO)
```

But: This cref header attribute is mentioned nowhere in the GADF docs, nor can I find any FITS documentation mentioning this.

The purpose is quite obvious, but I cannot find where this is ever defined.

adonath:
@maxnoe Just to confirm: `pyirf` uses the convention, that the recommended axis order corresponds to the Numpy order. Is this correct? Of course one could argue, that the recommended axis order of the GADF FITS specification should independent of the implementation and correspond to the one used in FITS itself, but in practice, I guess we don't really care...

maxnoe:
@adonath Yes, we decided that it is least confusing for users if the axis order matches the convention of the current environment, which for python / numpy means row-major / c and for writing the GADF files using FITS / Fortran / column major order.

lmohrmann:
@adonath It's possible the background model files for HESS data that I produced have an order that is not consistent with the other IRF files. When I read the array in python I get the order (energy, det_x, det_y), without having to transpose - I guess that is what you saw as well?

Is it possible to support both? Guess this will be difficult at least for the (presumably rare) case that the number of energy bins matches the number of spatial bins... 

I'm happy to produce the files with the order reversed in the future, but it would of course be great if it were possible to still read the current files also with future versions of Gammapy.

adonath:
Thanks @lmohrmann for checking! For some reason the 1DC files use a different axis order as well, but I completely agree that we have to support both, so that we can continue to use the already existing files. However we should still change our default behaviour to be compliant with GADF.

Just to confirm:
```python3

path = make_path("$GAMMAPY_DATA/hess-dl3-dr1/data/hess_dl3_dr1_obs_id_020136.fits.gz")

bkg_3d = Table.read(path, hdu="BKG")
print("Bkg  :", bkg_3d["BKG"].quantity[0].shape)

aeff_2d = Table.read(path, hdu="AEFF")
print("Aeff :", aeff_2d["EFFAREA"].quantity[0].shape)

psf_3d = Table.read(path, hdu="PSF")
print("Psf  :", psf_3d["RPSF"].quantity[0].shape)

edisp_2d = Table.read(path, hdu="EDISP")
print("Edisp:", edisp_2d["MATRIX"].quantity[0].shape)
```
Results in:
```
Bkg  : (20, 50, 50)
Aeff : (6, 96)
Psf  : (144, 6, 32)
Edisp: (6, 160, 96)
```


@AtreyeeS My proposal here is the following: as the test pass, we will merge this PR and then we open a second PR implementing a unit test for the `Background2D` and `Background3D` with axis order compliant with GADF. Internally we will handle both cases until we have new reference files...

maxnoe:
@adonath Did you see my comment about CREF? Do these files have this header attribute? 

Incidentally, one of the only goodle hits about `"CREF" fits header` was a comment by @cdeil in this repo concerning the same stuff.

maxnoe:
Ok, I did not find it until now, here is the definition in the GADF docs:

https://open-gamma-ray-astro.github.io/gamma-astro-data-formats/irfs/irf_axes/index.html

So you should check for a `CREF*` and `CUNIT*` header attribute for axis order / unit.

adonath:
Thanks @maxnoe! It seems the `CREF` keyword is not present in the DL3 DR1 files:
```python3
from astropy.io import fits

path = make_path("$GAMMAPY_DATA/hess-dl3-dr1/data/hess_dl3_dr1_obs_id_020136.fits.gz")
hdu = fits.open(path)["BKG"]
print(hdu.header)
``` 
```
XTENSION= 'BINTABLE'           / binary table extension                         
BITPIX  =                    8 / array data type                                
NAXIS   =                    2 / number of array dimensions                     
NAXIS1  =               200960 / length of dimension 1                          
NAXIS2  =                    1 / length of dimension 2                          
PCOUNT  =                    0 / number of group parameters                     
GCOUNT  =                    1 / number of groups                               
TFIELDS =                    7 / number of table fields                         
EXTNAME = 'BKG     '                                                            
FOVALIGN= 'RADEC   '                                                            
HDUDOC  = 'https://github.com/open-gamma-ray-astro/gamma-astro-data-formats'    
HDUVERS = '0.2     '                                                            
HDUCLASS= 'GADF    '                                                            
HDUCLAS1= 'RESPONSE'                                                            
HDUCLAS2= 'BKG     '                                                            
HDUCLAS3= 'FULL-ENCLOSURE'                                                      
HDUCLAS4= 'BKG_3D  '                                                            
TTYPE1  = 'DETX_LO '                                                            
TFORM1  = '50E     '                                                            
TUNIT1  = 'deg     '                                                            
TTYPE2  = 'DETX_HI '                                                            
TFORM2  = '50E     '                                                            
TUNIT2  = 'deg     '                                                            
TTYPE3  = 'DETY_LO '                                                            
TFORM3  = '50E     '                                                            
TUNIT3  = 'deg     '                                                            
TTYPE4  = 'DETY_HI '                                                            
TFORM4  = '50E     '                                                            
TUNIT4  = 'deg     '                                                            
TTYPE5  = 'ENERG_LO'                                                            
TFORM5  = '20E     '                                                            
TUNIT5  = 'TeV     '                                                            
TTYPE6  = 'ENERG_HI'                                                            
TFORM6  = '20E     '                                                            
TUNIT6  = 'TeV     '                                                            
TTYPE7  = 'BKG     '                                                            
TFORM7  = '50000E  '                                                            
TUNIT7  = 'MeV-1 s-1 sr-1'                                                      
TDIM7   = '(50,50,20)'                                                          
```
And it's not present in the reference files provided by GADF either:
```
XTENSION= 'BINTABLE'           / binary table extension                         
BITPIX  =                    8 / array data type                                
NAXIS   =                    2 / number of array dimensions                     
NAXIS1  =                 1200 / length of dimension 1                          
NAXIS2  =                    1 / length of dimension 2                          
PCOUNT  =                    0 / number of group parameters                     
GCOUNT  =                    1 / number of groups                               
TFIELDS =                    5 / number of table fields                         
OBS_ID  =                31415 / Observation ID                                 
LO_THRES=                  0.1 / Low energy threshold [TeV]                     
HI_THRES=                   50 / High energy threshold [TeV]                    
HDUDOC  = 'https://github.com/open-gamma-ray-astro/gamma-astro-data-formats'    
HDUVERS = '0.2     '                                                            
HDUCLASS= 'GADF    '                                                            
HDUCLAS1= 'RESPONSE'                                                            
HDUCLAS2= 'BKG     '                                                            
HDUCLAS3= 'FULL-ENCLOSURE'                                                      
HDUCLAS4= 'BKG_2D  '                                                            
EXTNAME = 'BACKGROUND'         / extension name                                 
TTYPE1  = 'ENERG_LO'                                                            
TFORM1  = '20D     '                                                            
TTYPE2  = 'ENERG_HI'                                                            
TFORM2  = '20D     '                                                            
TTYPE3  = 'THETA_LO'                                                            
TFORM3  = '5D      '                                                            
TTYPE4  = 'THETA_HI'                                                            
TFORM4  = '5D      '                                                            
TTYPE5  = 'BKG     '                                                            
TFORM5  = '100D    '                                                            
TDIM5   = '(5,20)  '                                                            
```
So for now we cannot really on it and have to implement a workaround...

adonath (approved):
Thanks @AtreyeeS, for this PR I have no further comments...

maxnoe:
@adonath I don't think implementing a workaround here makes sense since:

a) The file does not implement the recommended axis order
b) Fails to communicate the actual axis order

This should just fail!

adonath:
@maxnoe Yes, I agree...but no :-)  Of course I would prefer a clean and strict handling from the Gammapy side as well, but in practice it means that CTA 1DC and the current HESS FITS productions couldn't be used with Gammapy v0.18, which I think is just not an option. I doubt that the e.g. the 1DC will be recreated and I'm not sure when the next HESS FITS production will be released...

So far in Gammapy we followed the premise to be "kind on read" and "strict on write", which I think is the best compromise between user experience / convenience and code simplicity and maintenance. I need to think about what the best solution here is. I think the minimal action is the following:

- We add the recommended axis order to the format spec in https://github.com/gammapy/gammapy/blob/master/gammapy/irf/io.py#L10, with a reference to GADF. There we handle the translation of the names and add missing meta data...
- We add unit tests for the correct recommended axis order
- We add a check of the axis order on read, raise a warning if it's not compliant and guess the order from the shape
- We list this somewhere as a known issue...

Once the FITS exporters are adapted and new releases are out for some time, we can remove the support for the old files.

AtreyeeS:
Yes, I think we must support both for the time. I will do that in a follow-up PR.
But as @lmohrmann pointed out, I do not know how to guess the order from the shape in case of equal bins in `energy`, `det_x`, `det_y`

maxnoe:
> So far in Gammapy we followed the premise to be "kind on read" and "strict on write",

Yes, sorry, this is in general a good approach for implementing file specs!

But it should error in the (improbable case) that the axis order is not determined uniquely, e.g. same number of bins for two axis.

adonath:
@maxnoe I agree we should only allow the case where ` axes_shape == data_shape[::-1]` is uniquely true (so require in addition that `data_shape != data_shape[::-1]`) and we should only allow it for the `Background` classes. 

@AtreyeeS In case of equal bins, we assume the correct order. As far as I know, none of the existing data files has this case. If new ones are created, they have to be compliant with the correct axis order.

maxnoe:
I don't think it is a good idea to implent this guessing only for one special case for one IRF.

Either you want to be able to guess the axis order, than you should implement the guessing for all IRFs, or you don't, then this should error.

Only supporting it for one special case of one IRF will probably result in some pretty confusing behaviour for users.

maxnoe:
Ah, but I see you cannot guess here, since DETX and DETY have the same number of bins...

bummer, so a special case is needed.

lmohrmann:
Concerning whether to implement the "guessing" only for the background or all IRFs: given that a wrong ordering occurs only for the background IRF in all known FITS productions, I don't think one needs to be very strict here, especially since this would only be meant as a temporary solution (until these data are "outdated enough" to be able to drop support for them).

For the guessing, I think it's fair to assume that DET_X,DET_Y will always be in that order, so in case the third axis has a different number of bins (which will usually be the case), this determines the order.

PR #3084, opened 2020-10-27
Enable different axis orders in Background fits files 

AtreyeeS:
**Description**

This pull request .enables `Background2D` and `Background3D` to work with both the GADF recommenced axis order and the current HESS/CTA release as discussed in #3081 .
Strangely enough, our tests in `test_irf_write` had the special case of same number of bins in (energy, lat, lon)! I changed the energy binning to get sane results. 

**Dear reviewer**

I added  GADF files in Gammapy-data in https://github.com/gammapy/gammapy-data/pull/5
Those are required for the tests.

### DIFF
--- a/gammapy/irf/background.py
+++ b/gammapy/irf/background.py
@@ -111,9 +111,14 @@ class Background3D:
         #  have a reverse order (lon, lat, E) than recommened in GADF(E, lat, lon)
         #  For now, we suport both.
 
-        data = table[bkg_name].data[0] * data_unit
+        data = table[bkg_name].data[0].T * data_unit
         shape = (energy_axis.nbin, fov_lon_axis.nbin, fov_lat_axis.nbin)
+
+        if shape == shape[::-1]:
+            log.error("Ambiguous axes order in Background fits files!")
+
         if np.shape(data) != shape:
+            log.debug("Transposing background table on read")
             data = data.transpose()
 
         return cls(
@@ -289,9 +294,14 @@ class Background2D:
         # have a reverse order (theta, E) than recommened in GADF(E, theta)
         # For now, we suport both.
 
-        data = table[bkg_name].data[0] * data_unit
+        data = table[bkg_name].data[0].T * data_unit
         shape = (energy_axis.nbin, offset_axis.nbin)
+
+        if shape == shape[::-1]:
+            log.error("Ambiguous axes order in Background fits files!")
+
         if np.shape(data) != shape:
+            log.debug("Transposing background table on read")
             data = data.transpose()
 
         return cls(
--- a/gammapy/irf/background.py
+++ b/gammapy/irf/background.py
@@ -88,8 +88,10 @@ class Background3D:
         else:
             raise ValueError('Invalid column names. Need "BKG" or "BGD".')
 
-        data_unit = u.Unit(table[bkg_name].unit, parse_strict="silent")
-        if isinstance(data_unit, u.UnrecognizedUnit):
+        data_unit = table[bkg_name].unit
+        if data_unit is not None:
+            data_unit = u.Unit(table[bkg_name].unit, parse_strict="silent")
+        if isinstance(data_unit, u.UnrecognizedUnit) or (data_unit is None):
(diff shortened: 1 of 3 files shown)


gammapy/irf/background.py
```diff
@@ -105,11 +107,20 @@ def from_table(cls, table):
             table, column_prefix="DETY", format="gadf-dl3"
         )
 
+        # TODO: The present HESS and CTA backgroundfits files
+        #  have a reverse order (lon, lat, E) than recommened in GADF(E, lat, lon)
+        #  For now, we suport both.
+
+        data = table[bkg_name].data[0] * data_unit
+        shape = (energy_axis.nbin, fov_lon_axis.nbin, fov_lat_axis.nbin)
+        if np.shape(data) != shape:
```
adonath, line 116: I think we should be even more strict and use `if data.shape == shape[::-1] and not shape == shape[::-1]` or a similar condition. This catches the case, where the transpose of the shape is ambiguous

gammapy/irf/tests/test_background.py
```diff
@@ -144,6 +144,15 @@ def test_background_3D_read():
     assert data.unit == "s-1 MeV-1 sr-1"
 
 
+@requires_data()
+def test_background_3D_read_gadf():
+    filename = "$GAMMAPY_DATA/tests/bkg_3d_full_example.fits"
+    bkg = Background3D.read(filename)
+    data = bkg.data.data
+    assert data.shape == (20, 15, 15)
```
adonath, line 152: Just to be more explicit you could add an assert one the axes names as well...

gammapy/irf/tests/test_irf_write.py
```diff
@@ -9,18 +9,17 @@
 
 class TestIRFWrite:
     def setup(self):
-        self.energy_lo = np.logspace(0, 1, 11)[:-1] * u.TeV
```
adonath, line 12: Is there a good reason why the number of bins changed here? If not please undo...

gammapy/irf/background.py
adonath: And could emit a warning if this case occurs? Like `log.warning("Assuming transposed axis order for background...")`

adonath (changes requested):
Thanks @AtreyeeS, I have left a few minor comments...

gammapy/irf/tests/test_irf_write.py
AtreyeeS: We had 10 bins each in energy, lon, lat - the one ambiguous case we decided not to support.
So, I changed it to rather have 9 bins in energy.

gammapy/irf/background.py
AtreyeeS: Yes, I thought of adding a warning - but practically, since this warning will come up in *ALL* HESS and CTA analysis, it might give the user an idea that "gammapy warns a lot, ignore when it cries".

Of course, this is a question of broader philosophy of gammapy - so I have no strong opinion here. Let me know what you decide

registerrier: This could be considered an `info` message than a real `warning`. 
This would alleviate this issue.

gammapy/irf/tests/test_irf_write.py
adonath: Indeed, thanks!

gammapy/irf/background.py
adonath: I would then rather propose the other direction and make it a `log.debug()` statement. The point is just to have at least some possibility of knowing that the data was transposed during reading...

AtreyeeS: Ok! I added `log.debug`

gammapy/irf/tests/test_background.py
AtreyeeS: Done

gammapy/irf/background.py
AtreyeeS: Also, an error for shape == shape[::-1

adonath (approved):
Thanks a lot @AtreyeeS, no further comments from my side..

PR #3085, opened 2020-10-28
Remove feldman cousins method

adonath:
**Description**

This pull request removes the Feldman-Cousins method from `gammapy.stats`. It’s a widely used method, but right now it doesn’t integrate in our API and it’s completely unclear how to use it e.g. with model fitting and / or the counts statistic classes. Apparently there are also issues with the implementation https://groups.google.com/g/gammapy/c/IHMO6ToZM3Y, that no-one was able to resolve (including myself…it would require spending a few days reading and understanding the paper). The current implementation is not very efficient either and still uses Python loops over arrays etc.  For model fitting as well as counts statistics we now use the same uniform method for ULs based on the stat profiles.

The code is still in the git history and can be brought back as soon as we know, how it better integrates in our API.


dlennarz:
What kind of integration into the API are you looking for?

There are no issues with the implementation. In that post, user was looking at "gaussian distribution around the origin", but used x_bins=np.arange(0, 20), which is appropriate only for a Poisson process. For the plot of the Gaussian process, x_bins = np.linspace(-n_sigma * sigma, n_sigma * sigma, n_bins_x, endpoint=True) was used with n_sigma = 10 and n_bins_x = 1000.

I added the analytical solutions (fc_find_acceptance_interval_poisson and fc_find_acceptance_interval_gauss) which are not optimised for speed (straight usage of a for loop). They could be improved to reduce the execution time, but the whole point was that I added a numerical solution that is optimised for speed and it should be used in lieu of these functions. They are merely a cross-check.

adonath:
Thanks @dlennarz for the clarification! There are currently two main places where we expose UL computation in the API:
- [CountsStatistics.compute_upper_limit](https://github.com/gammapy/gammapy/blob/master/gammapy/stats/counts_statistic.py#L83) for counts based, model-free analysis á la Li&Ma
- [ParameterEstimator.estimate_ul](https://github.com/gammapy/gammapy/blob/master/gammapy/estimators/parameter.py#L203) for model based forward folding á la Fermi

I think it's an option to re-exposure the Feldman-Cousins as an alternative method there, however even in this case it's still a good idea to have the implementation in an independent package. There are a few other places in Gammapy where we rely on 3rd party packages for additional (optional) functionality.

PR #3086, opened 2020-10-28
Stats background convention 

adonath:
**Description**

This pull request is a follow up to #3051. It clarifies and unifies the notation between the datasets and stats classes. The convention is now the following:

- `n_bkg` or `background` is the "known" background level, computed as `counts_off * alpha`, neglecting errors on the background and ignoring the signal 
- `mu_bkg` or `npred_background` is the background level estimated from minimising the stat, possibly taking a signal into account

In addition this PR renames `npred_sig` to `npred_signal` and adapts the table in the `gammapy.stats` docs page.

### DIFF
--- a/gammapy/datasets/spectrum.py
+++ b/gammapy/datasets/spectrum.py
@@ -219,7 +219,7 @@ class SpectrumDataset(MapDataset):
             label="Measured excess",
             yerr=np.sqrt(np.abs(self.excess.data.flatten())),
         )
-        self.npred_sig().plot_hist(ax=ax, label="Predicted signal counts")
+        self.npred_signal().plot_hist(ax=ax, label="Predicted signal counts")
 
         ax.legend(numpoints=1)
         ax.set_title("")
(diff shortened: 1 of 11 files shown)


adonath:
A better description is maybe:

- `n_bkg` or `background` is the estimated background level, based on  "off" measurements, without taking the "on" counts  into account. 
- `mu_bkg` or `npred_background` is a refined "fitted" background level estimated by taking "on" counts and a signal prediction into account

gammapy/datasets/spectrum.py
```diff
@@ -477,10 +477,20 @@ def npred_background(self):
             n_on=self.counts.data,
             n_off=self.counts_off.data,
             alpha=self.alpha.data,
-            mu_sig=self.npred_sig().data,
+            mu_sig=self.npred_signal().data,
         )
         return RegionNDMap.from_geom(geom=self._geom, data=mu_bkg)
 
+    def npred_off(self):
+        """Predicted counts in the off region
```
AtreyeeS, line 485: Maybe add this is `alpha*npred_background` and link to the Wstat formula?

adonath: Thanks, indeed. There are probably a few more places where such a link would be useful, I'll check it consistently...

adonath: Done

PR #3087, opened 2020-10-29
Remove line about reprojection

AtreyeeS:
The markdown says there is an example for reproduction onto different WCS, but it is not there is the code.
So, I remove it for now. We can add an example for 1.0


adonath (approved):
Thanks @AtreyeeS!

PR #3088, opened 2020-10-29
Allowing Estimators norm parameter to be negative

registerrier:
**Description**

This pull request modifies the behavior of `Estimators` to have a consistent treatment of the `norm` parameter and of the `ts` value.

- `norm` was forced to be positive in all `Estimators` based on `FluxEstimator`. This follows a common approach in Fermi analysis, but this introduces systematic biases when looking at non-existent sources. Preventing the flux to become negative makes the mean value always positive. This might lead to incorrect results if the `Estimator` results are treated incorrectly: e.g. taking the average flux from `LightCurveEstimator` for a non-existent source. This is now changed. The `norm.min` is non longer 0. Only a single test value has changed (and only a very little). 

- The `ts` handling in `Estimator` is also confusing. In some estimators the sign of `norm` is applied to `ts`. Having a negative `ts` might be really confusing. This is now removed. 

- The `sqrt_ts` still carries the sign of the `norm`. We might change its name to `significance` to avoid false comparisons with Fermi tools results which are always positive.

**Dear reviewer**

should we leave an option to force `norm>=0` on all `Estimator`?

Opinions @adonath , @QRemy @AtreyeeS @bkhelifi ?

### DIFF
--- a/gammapy/estimators/core.py
+++ b/gammapy/estimators/core.py
@@ -44,7 +44,7 @@ class Estimator(abc.ABC):
                 raise ValueError(f"{difference} is not a valid method.")
 
     @staticmethod
-    def get_sqrt_ts(ts):
+    def get_sqrt_ts(ts, norm):
         r"""Compute sqrt(TS) value.
 
         Compute sqrt(TS) as defined by:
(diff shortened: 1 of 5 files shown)


adonath:
Thanks @registerrier! I don't think we should introduce an option for the estimators to force the norm in the positive range. I think consistency is the much better choice here, because otherwise users will choose different options and then we need to explain the implications of the choice again...

bkhelifi:
> Thanks @registerrier! I don't think we should introduce an option for the estimators to force the norm in the positive range. I think consistency is the much better choice here, because otherwise users will choose different options and then we need to explain the implications of the choice again...

I fully agree with you

adonath:
@registerrier I guess this is ready to merge? I don't have any further comments...

registerrier:
Yes I think it is ready.

adonath (approved):
Thanks @registerrier, Travis-CI fails are unrelated...

PR #3089, opened 2020-10-29
Introduce DatasetModels class and global model

adonath:
**Description**

This pull request introduces an immutable `DatasetModels` class and makes the `Models` object de facto a "global model" object. The logic is now the following:

- The `DatasetModels` class is immutable, so thing like `.append()`, `__setitem__` etc. do not work. This is important because we do caching of per model IRFs etc. in the `MapDataset` object. By making the `DatasetModels` immutable we can avoid to implement complex checking whether model states changed...
- The `Datasets.models` getter now collects all unique models from the sub-datasets and builds an immutable global model object.
- The `Datasets.models` setter takes a `Models` object and then distributes the sub-components to the individual datasets.
- All the creation of the `MapEvaluators` now happens in the `MapDataset.models` setter

Interaction now happens typically with the global `Models` object like:

```python3

datasets = Datasets([dataset_1, dataset_2])

fov_bkg_model_1 = FoVBackgroundModel(dataset_name=dataset_1.name)
fov_bkg_model_2 = FoVBackgroundModel(dataset_name=dataset_2.name)
model = SkyModel()

datasets.models = [model, fov_bkg_model_1, fov_bkg_model_2]
```

The work in #3086 and #3051 led to a unification between the background handling of the `MapDataset` and `MapDatasetOnOff`, by introducing the `.background` and `.npred_background()` attributes and method. In the `MapDatasetOnOff`, there is an implicit application of the background estimated from off counts and for the sake of consistency I'd prefer to introduce the same for the `MapDataset`. 

In addition this PR makes the models filename in `Datasets.read()` optional.


QRemy:
- All the evaluators are reset each time a model is added, so we loose the caching for the other models https://github.com/gammapy/gammapy/blob/7422519f4d9cc9dd9ca24ac4b570dc41256fc656/gammapy/datasets/map.py#L285
- could we cache datasets models so the datasets.models do not rebuild the list each time ? I guess the following could be moved to the setter and add an private attribute to cache DatasetModels 
https://github.com/gammapy/gammapy/blob/7422519f4d9cc9dd9ca24ac4b570dc41256fc656/gammapy/datasets/core.py#L150

adonath:
Just to underline: I think one important aspect here is the separation of the serialisation of the `Models` and `Datasets`. Because while the `Models` definition changes often during an analysis, the `Datasets` don't.

So you can do:
```python3

datasets = Datasets.read("datasets.yaml")

models = Models.read("my-model.yaml")
datasets.models = models

fit = Fit(datasets)
fit.optimize()

models.write("models-best-fit-1.yaml")

models.append(SkyModel())
datasets.models = models

fit = Fit(datasets)
fit.optimize()

models.write("models-best-fit-2.yaml")
```
I also added minimal "model-management" on the `Models` class. So e.g. one an serialise background and sky models independently:

```python3
bkg_models = models.select(tag="fov-bkg")
bkg_models.write("my-bkg-models.yaml")

sky_models = models.select(tag="sky-model")
sky_models.write("my-sky-models.yaml")

# and merge those again
models = Models(sky_models + bkg_models)
```

The fact that the model typically changes often and the data doesn't is also the main argument for me against serialising model parameters with the dataset. In future versions of Gammapy, we could maybe think about serialising a `MapDataset` to yaml, which then contains the `FoVBackgroundModel`  as well...

When thinking about this I also realised that serialising the `mask_fit` together with the dataset, was maybe not the best idea, because again it might change, with different passes of the same fit, while the data and IRFs and `mask_safe` don't. But this is a minor issue...

adonath:
@QRemy The `Fit` class already does the caching of the model internally: https://github.com/gammapy/gammapy/blob/master/gammapy/modeling/fit.py#L85, so for fitting it shouldn't be an issue...but I'll check once again.

Re-using cached `MapEvaluators` I'm not a 100%, I think setting up the evaluators is rather fast (compared to fitting). I just checked and in the setup phase of the evaluators we spent ~90% of the time in computing the `PSFKernel` from the `PSFMap`. I think the clear solution to speed this up in future is to introduce a `PSFKernelMap`, which we need anyway to handle non-radially symmetric PSFs...

registerrier:
I still think the design is confusing. There is some distinction between background and sky models at some level and there is none at others. 

On the dataset level, things are split between signal and background, each having their own `npred`. But for the model definition, the responsibility is left to the user to consider all models at once. 

The call `bkg_model = FoVBackgroundModel(dataset_name=dataset.name)` does not bring much. It requires the user to know about `FoVBackgroundModel` and dataset names. I would consider simpler and more obvious that datasets have all by default a background model. The `setter` on the dataset could then simply be:
```
def define_background_model(self, spectral_model=None):
        self._background_model = FoVBackgroundModel(spectral_model=spectral_model, dataset_name=self.name)
``` 

On `Datasets` you can then browse all models and background models:
```
def models(self):
     models = {}

     for dataset in self:
         if dataset.models is not None:
             for model in dataset.models:
                 models[model] = model
         if dataset.background_model is not None:
             models[dataset.background_model] = dataset.background_model
     return DatasetModels(list(models.keys()))
```

For clarity, one could even introduce a `sky_models` attribute rather than `models` on the `Dataset` level.

Note also that `cutout`, `downsample`, `pad`, `resample_energy_axis` and `to_spectrumdataset`  loose all background model information and simply call `npred_background` to recreate a `background` `Map`.
This is also rather unintuitive if somehow the background models are connected to the `Dataset`. Why not create the new background model with the same map and with the spectral norm model chosen?

adonath:
@registerrier I have also thought about it once again, but I don't agree with you here. The main argument for a "all in" handling of background models on the dataset (that we had ~ 1 year ago) was serialisation and the desire to avoid the "double" bookkeeping of models on the dataset objects. Especially on the "global" level.

> The call bkg_model = FoVBackgroundModel(dataset_name=dataset.name) does not bring much. It requires the user to know about FoVBackgroundModel and dataset names.

It does the bookkeeping of the background component and parameters, that's it. 

However I would still propose mainly two follow up changes to this PR (maybe as a compromise):
- Remove the default creation of the `FoVBackgroundModel`, so a `MapDataset` (and all other datasets...) never come with pre-defined background model parameters
- **But** implicitly apply the `.background` in the npred in `MapDataset`, to get a save default behaviour.
- And make clear in the print output of the `MapDataset`, that the `.background` is part of the npred.

This gives a stable default behaviour and it would lead to a fully consistent behaviour for all `MapDataset`, `MapDatasetOnOff`, `SpectrumDataset`! and `SpectrumDatasetOnOff`. I think it's very likely that the `SpectrumDataset` will be merged with the `MapDataset` at some point, I think it's harder to explain to users that we have a different behaviour (pre-defined background models or not) depending on whether there is a region defined or not. 

However I fully agree, that we need to think about once more on the behaviour of `.to_image()`, `to_spectrum_dataset()`, `.resample_energy_axis()` etc. Of course one could pre-define the `FoVBackgroundModel`, but I think it many cases it's not even desired. E.g. in `.to_spectrum_dataset()` or even in `.to_image()` one would need to freeze spectral background parameters etc. Or choose a different background model (e.g if one uses a `PiecewiseNormSpectralModel`, aligned with the energy binning of the background and one calls `.resample_energy_axis()`  ) I think this leads to a lot of implicit model bookkeeping and need for documentation of special cases. Let's leave the bookkeeping to the user for now. 

My proposal here would be again, that `Dataset` objects never come with predefined background models, so the model information is reset in any of the `.to_spectrum_dataset()`, `.to_image()` etc. methods. In general this leaves us with two possibilities:

- Always implicitly apply the spectral model correction if it's defined and use `.npred_background()` to define  the new `.background`. 
- Always use the un-corrected background map and rely on `.background`

In the first case one could remove the `FoVBackgroundModel` component to prevent the correction
```python3
dataset.models = None
new_dataset = dataset.to_image()
```
 In the second case one could explicitly set the background to `npred_background()` like:
```python3
dataset.background = dataset.npred_background()
new_dataset = dataset.to_image()
```
I don't really have a strong opinion there, I think one or the other works. os we just decided on one behaviour, implement and document it consistently...

AtreyeeS:
Hello! I agree with @adonath that the double booking was quite painful.
However, `.cutout`, `.to_image` etc losing the background_model information is not nice.
One of the reasons behind using using these methods is to fit the background on a different grid (larger/ finer/coarser - as the case may be) and then concentrate on a specific region. The parameters may probably be frozen in this case, but having to set by hand the background model is super inconvenient.

adonath:
@AtreyeeS Yes, I see this point as well. However as I have tried to illustrate in my previous comment, I think it's not always clear how to treat the background parameters in these methods either:
- In `.cutout()` one could copy over the model parameters, but taking it strictly they are not valid anymore, because the data has changed and re-fitting would result in different parameters. This is true for all other methods as well...
- In `.to_image()` one would need to freeze spectral parameters
- In `.to_spectrum_dataset()` one would need to freeze all parameters
- In `.resample_energy_axis()` it's unclear what to do with a `PiecewiseNormSpectralModel`
- In `.slice_by_energy()` one would need to freeze parameters of the `PiecewiseNormSpectralModel` outside the energy range
- etc.

So we would need to decide on the behaviour and document it in all cases. And for all methods it will be slightly different, depending on the spectral model chosen for the `FoVBackgroundModel`. I think all in all it is more confusing, then letting the user re-define the parameterisation themselves.

adonath:
The first proposal for this PR was outlined in #2942 a few months ago. Since then we had many discussion and we didn't manage to find a 100% convincing solution. This PR achieve at least the following:
- It introduces a global model API such that setting and assigning models on a global level `Datasets.models = Models.read()` works
- It improvers the separation between `Models` and `Datasets` such that they can be serialised independently. In the previous design, e.g. reading a new global model required reading all datasets again.
- Once the follow-up changes are implemented it achieves a consistent behaviour between all dataset objects, where models are never pre-defined and the background npred is applied implicitly, if the map is available.
- It achieves a consistent behaviour for all dataset methods such as `.to_image()`, `.cutout()` and even `.read()` and `.write()`. Where the model parameter information is always handled separately. 
- The behaviour can be explained and documented in a consistent way.
- This PR puts the burden of model bookkeeping on the should if the user, however it can be further simplified in future.
- This PR removes the upward reference to `Datasets` from the `ProperModels` level, which prevented a simple solution for distributed computing in future.

All in all I think this PR improves the existing design, even it if it fails to improve the user convenience. The previous discussion didn't lead to a better overall design for this. So because of the lack of a fully worked out alternative solution, I'll go ahead and merge this PR. I'll add documentation to explain the concept to users in a follow up PR.

AtreyeeS:
For the model bookkeeping after dataset operations, I do not think it is problematic if the behaviour on the different methods is different - as long as they do something that is expected and it is well documented.
For example, it indeed makes sense that the model values on `.cutout` are not the best fit parameters after doing the cutout - that is the reason for doing the analysis on a larger Geom in the first place!

Also, regarding freezing of parameters, it is ok to put that burden on the user - but having to reset the models every time is painful and error-prone.

Since this PR is already merged, I can try to update my scripts on this and come up with some book-keeping proposals for 1.0

adonath:
Thanks for the additional input @AtreyeeS! I fully agree we can add further functionality for bookkeeping on models. In this case I think we should also consider adding this functionality directly to the `Models` class.  In this PR I already added `Models.select(
dataset_name=, tag=, name_substring=, )` which allows users to organise and select models. I can easily imaging many other useful functions such as:
-  `Model.to_spatial_template(name="")`, which creates a spatial template from a list of `SkyModel` components
- `Models.parameters.freeze(type="spectral")` or `Models.parameters.freeze(type="spatial")`
- `Models.select_region(region=)`, which selects all models in a given region.
- `SourceCatalog.models` to create a model from an existing catalog
- etc. 

I think the advantage of the parallel bookkeeping the `FoVBackgroundModel` together with the `MapDataset` is overestimated. The typical use-case is that the `MapDataset` objects are created first and then a model is defined to be fit on those. Then the model is updated iteratively and fit to the same data again, without changing the dataset. Let's consider e.g. the  use case of an extended source spectral analysis, with 3D background correction, where the dataset changes in between:

```python3

dataset = MapDataset()
dataset.models = [SkyModel(, name="my-source"), FoVBackgroundModel(dataset_name=dataset.name)]

fov_maker = FoVBackgroundMaker()
dataset = fov_maker.run(dataset)

spectrum_dataset = dataset.to_spectrum_dataset(region=region)

# to prepare spectral fitting only this additional step is required...
spectrum_dataset.models = dataset.models["my-source"]
```

From my perspective it's totally fine to ask users to re-define the model if the associated data has changed. In the simplest case it is just re-setting the same model on the dataset again. That's it...
```python3
dataset_new = dataset.resample_energy_axis(name=dataset.name)
datasets_new.models = dataset.models
```
So typical one get's this one additional line of code for now.

registerrier:
OK. Fine. I understand most of the improvements of this PR. 
I think the issue is more at the level of models. I am not sure how things will go once one add background profiling though. Background modeling will become a purely internal feature to `Dataset`.

Regarding the example you give here:
```
dataset_new = dataset.resample_energy_axis(name=dataset.name)
dataset_new.models = dataset.models
```
I think we should forbid having two datasets with the same name in memory. Then if I create two datasets with the same name, how can I know which background map will be applied?

PR #3100, opened 2020-11-02
Simplify EBL absorption spectral model

QRemy:
This PR simplify EBL absorption spectral model as proposed in PIG 21: #2944


gammapy/modeling/models/spectral.py
```diff
@@ -1451,18 +1450,22 @@ class Absorption:
...
+    alpha_norm: float
+        Norm of the EBL model
     interp_kwargs : dict
         Interpolation option passed to `ScaledRegularGridInterpolator`.
         By default the models are extrapolated outside the range. To prevent
         this and raise an error instead use interp_kwargs = {"extrapolate": False}
     """
 
-    tag = "Absorption"
+    tag = ["EBLAbsorptionNormSpectralModel", "ebl-absorption"]
```
adonath, line 1463: Include "norm" in the short tag here? Like `ebl-norm` or `ebl-absorbtion-norm`?

gammapy/modeling/models/spectral.py
```diff
@@ -1451,18 +1450,22 @@ class Absorption:
...
+    alpha_norm = Parameter("alpha_norm", 1.0, frozen=True)
+    redshift = Parameter("redshift", 0.1, frozen=True)
 
-    def __init__(self, energy, param, data, filename=None, interp_kwargs=None):
-        self.filename = filename
+    def __init__(self, energy, param, data, redshift, alpha_norm, interp_kwargs=None):
+        self.filename = None
         # set values log centers
         self.param = param
```
adonath, line 1470: Is the `.param` still used? I think it's fixed to `redshift` now...

gammapy/modeling/models/spectral.py
```diff
@@ -1572,137 +1596,20 @@ def read_builtin(cls, name, interp_kwargs=None):
...
-
         """
         models = dict()
         models["franceschini"] = "$GAMMAPY_DATA/ebl/ebl_franceschini.fits.gz"
         models["dominguez"] = "$GAMMAPY_DATA/ebl/ebl_dominguez11.fits.gz"
         models["finke"] = "$GAMMAPY_DATA/ebl/frd_abs.fits.gz"
 
-        return cls.read(models[name], interp_kwargs=interp_kwargs)
-
-    def table_model(self, parameter):
```
adonath, line 1588: I guess one could convert it to a `PiecewiseNormSpectralModel`, but I'm not sure if there is any real use-case for this, so removing is fine I guess...

adonath (approved):
Thanks a lot @QRemy, I have left two minor inline comments...

gammapy/modeling/models/spectral.py
QRemy: I think we still need it to setup the ScaledRegularGridInterpolator and interpolate the table at the redshift value.

gammapy/modeling/models/spectral.py
QRemy: It was used only in the test_absorbed_extrapolate, I just replace it using evaluate
https://github.com/gammapy/gammapy/blob/ba65d5aa7be4bd77f4778c48420db499dd8d28d8/gammapy/modeling/models/tests/test_spectral.py#L476

gammapy/modeling/models/spectral.py
adonath: Indeed you are right...

adonath:
Thanks @QRemy!

PR #3101, opened 2020-11-02
Adapt energy naming convention

fabiopintore:
These PR extends the energy naming convention started in #3092 .


adonath (changes requested):
Thanks @fabiopintore, I have left two inline comments...

gammapy/catalog/fermi.py
```diff
@@ -217,7 +217,7 @@ class SourceCatalogObject4FGL(SourceCatalogObjectFermiBase):
         "ASSOC_GAM2",
         "ASSOC_GAM3",
     ]
-    _ebounds = u.Quantity([50, 100, 300, 1000, 3000, 10000, 30000, 300000], "MeV")
+    energy_edges = u.Quantity([50, 100, 300, 1000, 3000, 10000, 30000, 300000], "MeV")
```
adonath, line 220: Name is OK now, but please leave it as a hidden attribute with a trailing `_`...like `_energy_edges`

gammapy/data/event_list.py
```diff
@@ -307,11 +307,11 @@ def select_parameter(self, parameter, band):
         mask &= self.table[parameter].quantity < band[1]
         return self.select_row_subset(mask)
 
-    def _default_plot_ebounds(self):
+    def _default_plot_energybounds(self):
```
adonath, line 310: Could rename this to `_default_plot_energy_edges`

adonath (approved):
Thanks @fabiopintore, I have left two more comments otherwise looks good...

adonath:
Thanks @fabiopintore, I'll go ahead and merge this now, to avoid that we run into merge conflicts...

PR #3113, opened 2020-11-05
Modified TSMapEstimator to keep model term

registerrier:
**Description**

This pull request changes the `TSMapEstimator` to avoid losing the `model` when downsampling the dataset, or when slicing it.
A test is added to ensure inclusion of the model is properly taken into account.

**Dear reviewer**


adonath (approved):
Thanks @registerrier!

PR #3114, opened 2020-11-05
Corrected exclusion mask notebook

registerrier:
**Description**

This pull request because of the safe mask of the Fermi crab dataset, the `ExcessMapEstimator` is returning some nan values over the map. This small change avoids having them in the exclusion mask.

**Dear reviewer**


adonath (approved):
Thanks @registerrier!

PR #3116, opened 2020-11-06
Fix model handling in FluxEstimator

adonath:
**Description**

This pull request fixes a bug in the model setting of the `FluxEstimator`: as the `slice_by_energy()` method relies on `.npred_background()` internally, the `FoVBackgroundModel` has to be reset to it's default values and not applying the old values....

### DIFF
--- a/gammapy/analysis/tests/test_analysis.py
+++ b/gammapy/analysis/tests/test_analysis.py
@@ -325,8 +325,8 @@ def test_analysis_3d():
     assert len(analysis.flux_points.data.table) == 2
     dnde = analysis.flux_points.data.table["dnde"].quantity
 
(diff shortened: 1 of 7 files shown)


registerrier (commented):
Thanks @adonath . 
Are we sure that changing the background model parameters will behave correctly in case the `reoptimize` argument  is set to `True`?
In particular, all the user defined parameter settings on the background model are gone. If we are to leave the `reoptimize` option possible, we should leave the initial background model on the slice dataset no?
In practice, I suppose that only the background normalization should be free (as well as other model components).

adonath:
Thanks @registerrier ! I guess right now there is no other option. Since we currently rely on `npred_background()` in all the `.slice_by_energy()`  etc. methods we have to reset the background model before re-optimising.

Issue #3119, opened 2020-11-09
Optional key words required for dl3 data 


SimoneMender:
**Gammapy version**
Gammapy version: 0.18

**Bug description**
`DataStore.from_events_files(paths)` fails with `KeyError: "Keyword 'ALT_PNT' not found."` even tough 'ALT_PNT' is an optional column (see https://gamma-astro-data-formats.readthedocs.io/en/latest/events/pointing.html).

**Expected behavior**
`DataStore.from_events_files(paths)` should work without information about 'ALT_PNT'.

maxnoe:
* `DEADC` is optional in the `OBS_INDEX` HDUs, but gammapy assumes it is always there.

Gammapy should read `DEADC` from the `EVENTS` tables themselves, where it is a required keyword.

adonath:
The original issue is resolved by #3613.

PR #3120, opened 2020-11-09
Make reading DL3 data more standard conform

maxnoe:
**Description**

### `TELESCOP`
Reading in DL3 data required that `TELESCOP` is set in the `AEFF` table. 

* This is a strange place to look for it
* It is not required by the GADF standard that `TELESCOP` is even present.

So I:

* changed it to get `TELESCOP` from the event list, where it is more commonly stored and make it optional (setting it to `UNKNOWN` if not available.

### `DEADC`

`DEADC` is an optional column in the `OBS_INDEX` HDU but a required header keyword in the `EVENTS` table.

Gammapy required that it was available in the `OBS_INDEX` HDU. Changed to read it from the `EVENTS` table, where
it is guaranteed to be there for a standard compliant file.

### DIFF
--- a/gammapy/analysis/core.py
+++ b/gammapy/analysis/core.py
@@ -117,7 +117,7 @@ class Analysis:
             self.observations = self.observations.select_time([(start, stop)])
         log.info(f"Number of selected observations: {len(self.observations)}")
         for obs in self.observations:
-            log.debug(obs)
+            log.debug(str(obs))
 
     def get_datasets(self):
         """Produce reduced datasets."""
@@ -143,7 +143,7 @@ class Analysis:
         if not self.datasets or len(self.datasets) == 0:
             raise RuntimeError("Missing datasets")
 
-        log.info(f"Reading model.")
+        log.info("Reading model.")
         if isinstance(models, str):
             self.models = Models.from_yaml(models)
         elif isinstance(models, Models):
@@ -295,7 +295,7 @@ class Analysis:
                 if bkg_method == "ring":
                     dataset = dataset.to_map_dataset()
 
-                log.debug(dataset)
+                log.debug(str(dataset))
                 stacked.stack(dataset)
             datasets = [stacked]
         else:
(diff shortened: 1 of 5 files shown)


adonath:
Thanks a lot @maxnoe for the PR! I think the reason why the meta information is partly taken from a different place is simulation. So there is the use-case of defining an `Observation` object without events and the using it to compute a reduced `MapDataset`, which then can be used to simulate events from the given IRFs and source model. In this case the meta info is partly taken from the IRFs and partly provided by the user (such as pointing, observation time, dead time correction) This is e.g. done in  https://docs.gammapy.org/0.18.1/tutorials/event_sampling.html.

However in general I agree we need to check the conformity of required and optional meta information (see also https://github.com/gammapy/gammapy/issues/3119). I'll comment again when I'm officially back at work next week...

maxnoe:
The issues here prevented valid files from being read in by gammapy.

I understand you want to enable things going further than the current standard (e.g. mc studies), but not being able to reading in standard compliant files is a bug.

adonath:
> I understand you want to enable things going further than the current standard (e.g. mc studies), but not being able to read in standard compliant files is a bug.

I completely agree, I just wanted to provide the context for the fix, because I think there are still test fails related to simulation...I'll take a look now and do a review...

adonath:
Here is some more info / context:
- The `Observation` class does lazy-loading via the `LazyFitsData` descriptor and does not cache the data, this is needed to be able to process a large amount of data and not run out of memory.
- So accessing `.events.meta` will in fact read the events file again every time the meta info is accessed. 
- That's why there is the additional `obs_info` attribute which holds the meta independently of the data. 

So I think we should definitely keep the implementation, where meta data is accessed via the `obs_info` attribute. In the `Observation` object the meta data is then translated into Python object via properties, such as `Time` and `SkyCoord` objects. In some cases we handle the missing keywords by putting in `NaN` as place-holder. In other case I think the call to the property just fails, if the meta data is not present. It think even this is acceptable behaviour, however I fully agree we should nowhere rely on the presence of those optional keywords, except if they are explicitly needed to run the code. So we should systematically check the places where we access the info...

PR #3122, opened 2020-11-12
Correct excess_matching_significance behavior

registerrier:
**Description**

This pull request corrects `WStatCountsStatictics.n_sig_matching_significance`. 
After this commit : https://github.com/gammapy/gammapy/commit/5c962385f5ba016a68f43893cedcb27d9e9a4796  a bug was introduced in the method after replacing `background` or `n_bkg` (alpha*n_off)  by `mu_bkg` (the actual estimated background).

It seems that the estimated background as computed here is not a really relevant quantity. It has been removed, to avoid further confusion.

The error was reported by @maxnoe on slack.

**Dear reviewer**

### DIFF
--- a/gammapy/stats/counts_statistic.py
+++ b/gammapy/stats/counts_statistic.py
@@ -218,14 +218,6 @@ class WStatCountsStatistic(CountsStatistic):
             self.mu_sig = np.asanyarray(mu_sig)
 
     @property
-    def mu_bkg(self):
-        """Expected background"""
-        mu_bkg = self.alpha * get_wstat_mu_bkg(
-            n_on=self.n_on, n_off=self.n_off, alpha=self.alpha, mu_sig=self.mu_sig,
-        )
-        return np.nan_to_num(mu_bkg)
-
-    @property
     def n_bkg(self):
         """Known background computed alpha * n_off"""
         return self.alpha * self.n_off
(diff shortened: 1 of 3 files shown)


maxnoe:
Why is this milestoned for 1.0? 0.18.1 was just released and this bug was immediately identified.  Iit should be fixed in a fast 0.18.2.

gammapy/stats/counts_statistic.py
```diff
@@ -217,14 +217,6 @@ def __init__(self, n_on, n_off, alpha, mu_sig=None):
         else:
             self.mu_sig = np.asanyarray(mu_sig)
 
-    @property
-    def mu_bkg(self):
```
adonath, line 221: I agree this can be removed, in the case of "counts statistics", we expect `n_bkg = mu_bkg`, `n_sig = mu_sig`, `n_on = mu_on` etc. anyway. However ever I noticed that` n_bkg = mu_bkg` is not fulfilled (which could be considered as the actual bug...). This is because we deliberately set `mu_sig = 0` in `WStatCountsStatistic.__init__`, however in the case of no known signal component it should be rather set to `n_on - alpha * n_off`. Maybe we should fix this as well? Of possibly rename `mu_sig` here completely, as it does not correspond to the what is defined [here](https://docs.gammapy.org/0.18.1/stats/index.html#notations)...

adonath:
@maxnoe We use the v1.0 milestone sometimes as placeholder. We'll tag another v0.18.2 release this week.

adonath (approved):

PR #3126, opened 2020-11-17
Fix #3123

### LINKED ISSUE #3123: `LightCurveEstimator` gives bizarre results with `LogParabolaSpectralModel` or `ExpCutoffPowerLawSpectralModel`
### version

present dev / 0.18.1

### Issue

Using the `LightCurveEstimator` with a `LogParabolaSpectralModel` or a `ExpCutoffPowerLawSpectralModel` is giving a bizarre error:

- The unit on the `flux` column vanishes
- Has a shape of `(n,1)` and length `n` (n-number of points in the light curve) instead of the expected shape of `(1)` and length `n`

### To reproduce:
Change to a `LogParabolaSpectralModel` in the `light curve_simulation` notebook

adonath:
**Description**

This pull request fixes #3123. In #3070, the behaviour for scalars for `integrate_spectrum` was changed unintentionally and went unnoticed because of missing test coverage. This PR adds a test for ECPL flux points, simplifies `integrate_spectrum()` and adds a regression test for the scalar case.

@AtreyeeS Could you please check once if this fixes the behaviour you noticed in #3123?

### DIFF
--- a/gammapy/astro/darkmatter/tests/test_spectra.py
+++ b/gammapy/astro/darkmatter/tests/test_spectra.py
@@ -32,5 +32,5 @@ def test_DMAnnihilation():
     )
     differential_flux = model.evaluate(energy=1 * u.TeV, scale=1).to("cm-2 s-1 TeV-1")
 
-    assert_quantity_allclose(integral_flux.value, 6.19575457e-14, rtol=1e-5)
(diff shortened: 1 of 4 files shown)


AtreyeeS:
Thanks @adonath ! Just checked - it indeed solves #3123

AtreyeeS (approved):

adonath:
Thanks @AtreyeeS!

PR #3129, opened 2020-11-18
Fix `spectral._propagate_error`

fabiopintore:
This PR fixes a small bug in the `spectral._propagate_error` function.


adonath (commented):
Thanks @fabiopintore!

PR #3133, opened 2020-11-20
Change find_peaks to work with flat maps rather than images

registerrier:
**Description**

This pull request is a simple modification of `gammapy.estimators.utils.find_peaks` to work with flat `Map` and not only image like `Map`. The `map.get_image_by_idx((0,))` that was necessary in the detect.ipynb tutorial is unclear and unnecessary.

**Dear reviewer**

### DIFF
--- a/docs/tutorials/cta_data_analysis.ipynb
+++ b/docs/tutorials/cta_data_analysis.ipynb
@@ -325,7 +325,7 @@
    "outputs": [],
    "source": [
     "sources = find_peaks(\n",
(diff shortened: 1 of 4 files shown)


adonath (approved):

adonath:
Sorry, the CTA tutorial could be adapted as well: https://docs.gammapy.org/0.18.2/tutorials/cta_data_analysis.html#Source-Detection

adonath (approved):

PR #3135, opened 2020-11-24
Fix API links in notebooks

Bultako:
This small PR fixes broken links found in notebooks pointing to the autogenerated API doc.


adonath (approved):
Thanks @Bultako, no further comments from my side...

PR #3136, opened 2020-11-24
Evaluate spatial model in RegionGeom

LauraOlivera:
**Description**

This pull request was done together with @fabiopintore with the goal to be able to evaluate a `SpatialModel` in a `RegionGeom`. To do this, we created a method "`RegionGeom.to_wcs_geom()`" that returns the smallest geometry that contains a region. This is used by "`RegionGeom.get_wcs_coords()`", which returns a `MapCoord` object with the coordinates that are inside of this region. This can then be used to evaluate a spatial model as `model(coord.lon, coord.lat)`.

A small, working example:
```

from gammapy.maps import RegionGeom, MapAxis
from gammapy.modeling.models import GaussianSpatialModel
center = SkyCoord("0d", "0d")
radius = 1 * u.deg
circle = CircleSkyRegion(center, radius)
axis = MapAxis.from_energy_bounds("1 TeV", "10 TeV", nbin=10)
geom = RegionGeom(region=circle, axes=[axis])
positions = geom.get_wcs_coord()
model = GaussianSpatialModel(lon_0=0 * u.deg, lat_0=0 * u.deg, sigma=3*u.deg, e=0.8, phi=30 * u.deg)
v = model(positions.lon, positions.lat)
model_counts = v.sum()*geom.solid_angle()
```

**Dear reviewer**

There are tests missing at the moment, we will be working on them now.

### DIFF
--- a/gammapy/maps/region.py
+++ b/gammapy/maps/region.py
@@ -192,21 +192,9 @@ class RegionGeom(Geom):
         wcs_geom : `~WcsGeom`
             A WCS geometry object.
         """
-        wcs_new = self.wcs.copy()
-        width = self.width.value
-        binsz = np.abs(self.wcs.wcs.cdelt)
-        npix = (
-            np.rint(width[0] / binsz[0]).astype(int),
-            np.rint(width[1] / binsz[1]).astype(int),
-        )
-        nxpix = int(npix[0].flat[0])
-        nypix = int(npix[1].flat[0])
-        refpix = ((nxpix + 1) / 2.0, (nypix + 1) / 2.0)
-        wcs_new.wcs.crpix = refpix
-        wcs_new.array_shape = npix[0].flat[0], npix[1].flat[0]
-        wcs_geom = WcsGeom(wcs_new,
-                        npix=npix,
-                        )
+        wcs_geom_region = WcsGeom(wcs=self.wcs, npix=self.wcs.array_shape)
+        wcs_geom = wcs_geom_region.cutout(position=self.center_skydir, width=self.width)
+        wcs_geom = wcs_geom.to_cube(self.axes)
         return wcs_geom
 
     def get_wcs_coord(self):
@@ -215,15 +203,14 @@ class RegionGeom(Geom):
 
         Returns
         -------
-        region_coords : `~MapCoord`
+        region_coord : `~MapCoord`
             MapCoord object with the coordinates inside
             the region.
         """
         wcs_geom = self.to_wcs_geom()
         common_coord = self.contains(wcs_geom.get_coord())
-        region_coords_skycoord = wcs_geom.get_coord().skycoord[common_coord]
-        region_coords = MapCoord.create(region_coords_skycoord)
-        return region_coords
+        region_coord = wcs_geom.get_coord().apply_mask(common_coord)
+        return region_coord
 
     def to_cube(self, axes):
         axes = copy.deepcopy(self.axes) + axes
--- a/gammapy/maps/region.py
+++ b/gammapy/maps/region.py
@@ -215,13 +215,14 @@ class RegionGeom(Geom):
 
         Returns
         -------
-        region_coords : `~SkyCoord`
-            SkyCoord object with the coordinates inside
+        region_coords : `~MapCoord`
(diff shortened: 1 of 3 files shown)


gammapy/maps/region.py
```diff
@@ -183,6 +183,48 @@ def bin_volume(self):
...
+    def to_wcs_geom(self):
+        """Get the minimal equivalent geometry that
+            contains the region.
+
+        Returns
+        -------
+        wcs_geom : `~WcsGeom`
+            A WCS geometry object.
+        """
+        wcs_new = self.wcs.copy()
```
adonath, line 195: Now looking at the implementation, I realise that the simple solution that you proposed could work with a little modification as well. Let me try to explain:

I think for the `RegionGeom` we have two `wcs` definitions one that is passed on `__init__`, which is not "spatially confined" and then the one created here which is equivalent, but "spatially confined". You described this nicely as "minimal equivalent geometry that contains the region". The point is now that if we just go through `WcsGeom.create()` to create this, the reference point for the projection will be shifted (forced to the center) and thus the original `wcs` is possibly modified. I noticed that the same now happens in the current implementation as well.

However I think there is good alternative solution, if we just go through `.cutout()` here is a minimal example:
```python3
from gammapy.maps import WcsGeom, RegionGeom
geom = RegionGeom.create("icrs;circle(0, 0, 1)")

wcs_geom = WcsGeom(wcs=geom.wcs, npix=geom.wcs.array_shape)
print(wcs_geom)

cutout = wcs_geom.cutout(position=geom.center_skydir, width=geom.width)
print(cutout)
```
This will leave the original reference point untouched and still create a "confined" wcs.

adonath (dismissed):
Thanks a lot @LauraOlivera and @fabiopintore! I have left two inline comments.

In general it might be a good idea to cache the coordinates returned by `.get_wcs_coord()`, but this is something I can add in follow up PR at some point, as it's very technical...

gammapy/maps/region.py
```diff
@@ -183,6 +183,48 @@ def bin_volume(self):
...
+        nypix = int(npix[1].flat[0])
+        refpix = ((nxpix + 1) / 2.0, (nypix + 1) / 2.0)
+        wcs_new.wcs.crpix = refpix
+        wcs_new.array_shape = npix[0].flat[0], npix[1].flat[0]
+        wcs_geom = WcsGeom(wcs_new,
+                        npix=npix,
+                        )
+        return wcs_geom
+
+    def get_wcs_coords(self):
```
adonath, line 212: Minor: maybe call this `.get_wcs_coord()` (without the "s" at the end), just for consistency with the other `.get_coord()` method...

adonath:
BTW: The code in the example you posted could be directly implemented in `SpatialModel.evaluate_geom()` and used as a test as well...

gammapy/maps/region.py
```diff
@@ -183,6 +183,48 @@ def bin_volume(self):
...
+        npix = (
+            np.rint(width[0] / binsz[0]).astype(int),
+            np.rint(width[1] / binsz[1]).astype(int),
+        )
+        nxpix = int(npix[0].flat[0])
+        nypix = int(npix[1].flat[0])
+        refpix = ((nxpix + 1) / 2.0, (nypix + 1) / 2.0)
+        wcs_new.wcs.crpix = refpix
+        wcs_new.array_shape = npix[0].flat[0], npix[1].flat[0]
+        wcs_geom = WcsGeom(wcs_new,
```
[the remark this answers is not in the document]
LauraOlivera: Very good point, now if the `RegionGeom` has extra axis, the output geometry has them too. Thanks!

gammapy/maps/region.py
LauraOlivera: Thank you so much Axel, that is so much nicer - I didn't know this method existed. The implementation now is much simpler and also more correct I think.

registerrier (approved):
Thanks @LauraOlivera  and @fabiopintore .

This looks good. I have no special comment.

adonath (approved):
Thanks a lot @LauraOlivera and @fabiopintore, no further comments from my side...

PR #3137, opened 2020-11-25
Adding a function to update parameters from dict

facero:
This PR adds the ability to update a Parameter from a dictionary.
This introduces a public API and avoids the need to access the hidden functions and hidden parameter name in the dict.
A simple test has been added to test that parameters are indeed updated.
This will be helpful for another PR #3029

### DIFF
--- a/gammapy/modeling/parameter.py
+++ b/gammapy/modeling/parameter.py
@@ -248,10 +248,11 @@ class Parameter:
         return copy.deepcopy(self)
 
     def update_from_dict(self, data):
-        """Update parameters from a dict. Do not modify the name of parameters.
-            Key name is accessed as a hidden parameters with _keyname. """
+        """Update parameters from a dict.
+           Protection against changing parameter name."""
         data.pop("name")
-        self.__dict__.update({ '_'+k: v for k, v in data.items() })
+        for k in data.keys(): setattr(self, k, data[k])
+
 
     def to_dict(self):
         """Convert to dict."""
--- a/gammapy/modeling/parameter.py
+++ b/gammapy/modeling/parameter.py
@@ -247,6 +247,11 @@ class Parameter:
         """A deep copy"""
         return copy.deepcopy(self)
 
+    def update_from_dict(self, data):
+        """Update parameters from a dict.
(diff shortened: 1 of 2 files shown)


gammapy/modeling/parameter.py
```diff
@@ -247,6 +247,11 @@ def copy(self):
         """A deep copy"""
         return copy.deepcopy(self)
 
+    def update_from_dict(self, data):
+        """Update parameters from a dict.
+            Key name is accessed as a hidden parameters with _keyname. """
+        self.__dict__.update({ '_'+k: v for k, v in data.items() })
```
adonath, line 253: Maybe we should forbid (or ignore) to modify the `.name` attribute, because I think it will break model evaluation...

adonath (commented):
Thanks @facero, I just left one little comment, otherwise looks good to me!

gammapy/modeling/parameter.py
facero: indeed. It could be ignored. There is also an issue with min, max if set to 'nan' this is a string instead of a float. in dict or table, it shows as nan but it is np.nan in reality.

adonath: I think this is handled in the min and max setters: https://github.com/gammapy/gammapy/blob/master/gammapy/modeling/parameter.py#L169

I now realise going through `__dict__` was maybe a bad idea..because we do some type casting / checking in the setters. However using `setattr()` would probably work better:
```python3
from gammapy.modeling import Parameter
p = Parameter("test", value=4)
setattr(p, "min", "42")
print(p.min)
```

facero: indeed, this can be problematic with value for example with factor and scaling

adonath: Maybe change the implementation to rely on on `setattr(self, k, v)`, then it's safe?

facero: changed to `setattr()` and it indeed solves the issues when handling values and units, etc
Also simpler to read as there is no need to access hidden param.

adonath (approved):
Thanks a lot @facero, no further comment from my side...

PR #3139, opened 2020-11-26
Fix spatial.integrate_geom

fabiopintore:
@LauraOlivera and me open this PR to correct a bug in the current `modeling.models.spatial.integrate_geom()` function, which prevents to estimate the correct integral value of the spatial model for a given `RegionGeom` object. 
The new implementations fix it and it also allows for numpy broadcasting when the `RegionGeom` has an extra energy axis.


adonath (approved):
Thanks @fabiopintore and @LauraOlivera, no further comments from my side...

Issue #2915, opened 2020-06-03
Add option on ExcessMapEstimator not to correlate OFF data


registerrier:
By default OFF data is correlated with the ON data in the `ExcessMapEstimator`. This allows a standard behavior for real OFF observation and ring-like OFF datasets. 

Yet, as @lmohrmann noted, a number of classical analysis do not correlate OFF deduced from ring background. It might therefore be worth to keep the option available not to convolve the OFF map for a `MapDatasetOnOff`.

adonath:
Hm, I would presume there is only one correct way of doing this?

PR #3140, opened 2020-11-26
Add an option to ExcessMapEstimator to choose to correlate off events

mdebony:
**Description**
This pull request is linked to the issue #2915
An option to ExcessMapEstimator has been added to be able to choose to correlate OFF events. The default behavior is changed to not correlate off events in the case of a MapDatasetOnOff.
This option doesn't affect the behavior in the case of a MapDataset.

**Dear reviewer**
Wouldn't it better to calculate alpha from acceptance and acceptance_off field present in MapDatasetOnOff, than on the ratio between off events and predicted events by the background model ?

### DIFF
--- a/gammapy/estimators/excess_map.py
+++ b/gammapy/estimators/excess_map.py
@@ -59,6 +59,10 @@ class ExcessMapEstimator(Estimator):
     If a model is set on the dataset the excess map estimator will compute the excess taking into account
     the predicted counts of the model.
 
+    Some background estimation techniques like ring background or adaptive ring background will provide already
+    correlated data for OFF. In the case of already correlated OFF data, the OFF data should not be correlated again,
+    and so the option correlate_off should set to False (default).
+
     Parameters
     ----------
     correlation_radius : ~astropy.coordinate.Angle
--- a/gammapy/estimators/excess_map.py
+++ b/gammapy/estimators/excess_map.py
@@ -30,18 +30,19 @@ def convolved_map_dataset_counts_statistics(dataset, kernel, mask, correlate_off
     n_on_conv = np.rint(n_on.convolve(kernel.array).data)
 
     if isinstance(dataset, MapDatasetOnOff):
-        background = dataset.background * mask
-        background.data[dataset.acceptance_off.data == 0] = 0.0
         n_off = dataset.counts_off * mask
         npred_sig = dataset.npred_signal() * mask
+        acceptance_on = dataset.acceptance * mask
+        acceptance_off = dataset.acceptance_off * mask
 
-        background_convolve = background.convolve(kernel.array)
         npred_sig_convolve = npred_sig.convolve(kernel.array)
+        acceptance_on_convolve = acceptance_on.convolve(kernel.array)
         if correlate_off:
             n_off = n_off.convolve(kernel.array)
+            acceptance_off = acceptance_off.convolve(kernel.array)
 
         with np.errstate(invalid="ignore", divide="ignore"):
-            alpha = background_convolve / n_off
+            alpha = acceptance_on_convolve / acceptance_off
 
         return WStatCountsStatistic(
             n_on_conv.data, n_off.data, alpha.data, npred_sig_convolve.data
--- a/gammapy/estimators/excess_map.py
+++ b/gammapy/estimators/excess_map.py
@@ -18,7 +18,7 @@ __all__ = [
 log = logging.getLogger(__name__)
 
 
-def convolved_map_dataset_counts_statistics(dataset, kernel, mask):
+def convolved_map_dataset_counts_statistics(dataset, kernel, mask, correlate_off):
     """Return CountsDataset objects containing smoothed maps from the MapDataset"""
     # Kernel is modified later make a copy here
     kernel = copy.deepcopy(kernel)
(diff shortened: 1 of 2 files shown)


mdebony:
Hi,
I've update the test default parameter to False (just a small mistake). 
I see that known the test doesn't work anymore due to the quite high difference in significance in this specific case.
What is the protocol to update test ?

gammapy/estimators/excess_map.py
```diff
@@ -95,13 +97,15 @@ def __init__(
         selection_optional="all",
         energy_edges=None,
         apply_mask_fit=False,
+        correlate_off=False
```
registerrier, line 100: I would put the default at `True` since this is also the behavior for regular datasets.

gammapy/estimators/excess_map.py
```diff
@@ -82,6 +82,8 @@ class ExcessMapEstimator(Estimator):
     apply_mask_fit : Bool
         Apply a mask for the computation.
         A `~gammapy.datasets.MapDataset.mask_fit` must be present on the input dataset
+    correlate_off : Bool
+        Correlate OFF events in the case of a MapDatasetOnOff
```
registerrier, line 86: You could add a line in the docstring before the parameters explaining the role of this parameter.
Something like, "Some OFF estimation techniques, such as the ring background and adaptive ring background, will yield already correlated `counts_off` maps. They should therefore not be correlated again and the `correlated_off` parameter must be set to False.

gammapy/estimators/excess_map.py
```diff
@@ -33,18 +33,18 @@ def convolved_map_dataset_counts_statistics(dataset, kernel, mask):
...
         background.data[dataset.acceptance_off.data == 0] = 0.0
         n_off = dataset.counts_off * mask
-
-        background_conv = background.convolve(kernel.array)
-        n_off_conv = n_off.convolve(kernel.array)
-
         npred_sig = dataset.npred_signal() * mask
-        mu_sig = npred_sig.convolve(kernel.array)
+
+        background_convolve = background.convolve(kernel.array)
```
registerrier, line 38: I think you would convolve the background (alpha*counts_off) only if correlate_off is set to True no?

gammapy/estimators/excess_map.py
```diff
@@ -33,18 +33,18 @@ def convolved_map_dataset_counts_statistics(dataset, kernel, mask):
...
+        npred_sig_convolve = npred_sig.convolve(kernel.array)
+        if correlate_off:
+            n_off = n_off.convolve(kernel.array)
 
         with np.errstate(invalid="ignore", divide="ignore"):
```
registerrier, line 43: The `alpha` you want to use is either :
- `alpha_conv = background_conv / n_off_conv` if you want to apply correlation (i.e. it the off weighted means of the alpha).
- `alpha = acceptance_conv/acceptance_off` if you don't correlate the OFFs, since in that case the `acceptance_off` is left untouched.

registerrier (changes requested):
Thanks @mdebony . I think you need to change a bit to have the correct `alpha` when no correlation is applied. See inline comments.

Regarding tests. The existing ones should be unchanged. You should add one for the non correlated case in the tests/test_excess_map.py file.
You could create a `test_excess_map_onoff_dataset_no_correlation` function similar to `test_significance_map_estimator_map_dataset_on_off(simple_dataset_on_off)` but simply doing one test with the correlate_off parameter set to True

gammapy/estimators/excess_map.py
mdebony: On the slack discussions, we discuss to by default not correlate off. As it seems to me that this function, in the case of a MapDatasetOnOff will mainly be called after a ring background it's seem to me better to not correlate off by default.

gammapy/estimators/excess_map.py
mdebony: Ok sure I will do that.

gammapy/estimators/excess_map.py
mdebony: If I'm right in this case the background is used as an estimation of the acceptance for On events, as we correlate On events, we should also correlate the acceptance.

gammapy/estimators/excess_map.py
mdebony: Ok. For the off correlation case, will it not be better to use also the correlated acceptance On and correlated acceptance Off ?

mdebony:
Ok I will update the test this way.

gammapy/estimators/excess_map.py
registerrier: A priori, we don't return the correlated `acceptance` and the correlated `acceptance_off`. So it might not be necessary to compute them, the point is to get the correct `alpha`.

gammapy/estimators/excess_map.py
registerrier: The ON acceptance is given by the `dataset.acceptance` member. 
`dataset.background = dataset.alpha * dataset.counts_off` in the case of a `MapDatasetOnOff`.

gammapy/estimators/excess_map.py
registerrier: OK. I suppose the important point is that it is well documented.

adonath:
@mdebony and @registerrier What is the status here?

mdebony:
I've been very busy the last two weeks, so not a lot of change actually, but I'm planning to working on it the coming week.

mdebony:
Hi,
I've update the alpha calculation, add some sentence in the documentation in order to explain better the parameter and add test for the case without off correlation.

registerrier (approved):
Thanks @mdebony ! This looks good.

PR #3144, opened 2020-11-30
Add the option to impose a minimal width in RegionGeom.to_wcs_geom()

LauraOlivera:
**Description**

This pull request adds the option `width_min` to the existing method `RegionGeom.to_wcs_geom()` to force a minimum width for the geometry. The input can be both a single value or 2, one for each spatial dimension

### DIFF
--- a/gammapy/maps/region.py
+++ b/gammapy/maps/region.py
@@ -11,6 +11,7 @@ from gammapy.utils.regions import (
     list_to_compound_region,
     make_region,
 )
+from gammapy.maps.wcs import _check_width
 from .core import MapCoord
 from .geom import Geom, MapAxes, MapAxis, pix_tuple_to_idx
 from .wcs import WcsGeom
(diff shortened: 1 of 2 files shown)


gammapy/maps/region.py
```diff
@@ -185,17 +185,33 @@ def bin_volume(self):
...
+            Minimal width for the resulting geometry.
+            Can be a single number or two, for 
+            different minimum widths in each spatial dimension.
+
         Returns
         -------
         wcs_geom : `~WcsGeom`
             A WCS geometry object.
         """
+        if width_min is not None:
```
adonath, line 204: Does this work?
```python3
width = np.min([self.width.to_value("deg"), _check_width(width_min)], axis=0)
```
Where `_check_width()` is imported from `gammapy.maps.wcs`...

adonath (commented):
Thanks @LauraOlivera, I have left two inline comments...

gammapy/maps/region.py
```diff
@@ -185,17 +185,33 @@ def bin_volume(self):
...
         return bin_volume
 
-    def to_wcs_geom(self):
+    def to_wcs_geom(self, width_min=None):
         """Get the minimal equivalent geometry that
             contains the region.
 
+        Parameters
+         ----------
+            width_min : `~astropy.quantity.Quantity`
```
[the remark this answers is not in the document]
LauraOlivera: ups

gammapy/maps/region.py
LauraOlivera: oh yes, that is fantastic! once again, had no idea this existed, very useful function!

adonath (approved):

PR #3145, opened 2020-12-01
Change the `_compute_flux_spatial` method on `MapEvaluator`

fabiopintore:
This PR includes some changes in the `_compute_flux_spatial` function in `~gammapy.datasets.map.MapEvaluator`. 
We now allows the function to calculate the integral flux over either a `WcsGeom` or a `RegionNDMap`. The integrated flux is also corrected for the PSF.

To do:
1) We need to add tests!
2) possibility to enlarge the tangential geometry so that the PSF is fully contained.

@LauraOlivera

### DIFF
--- a/gammapy/datasets/map.py
+++ b/gammapy/datasets/map.py
@@ -2572,12 +2572,12 @@ class MapEvaluator:
         value = self.compute_flux_spectral()
 
         if self.geom.is_region:
-            evaluate_spatial_model = self.geom.region is not None
+            evaluate_spatial_model = self.geom.region is not None and self.psf
         else:
             evaluate_spatial_model = True
 
         if self.model.spatial_model and evaluate_spatial_model:
-            value = value * self.compute_flux_spatial().quantity
+            value = value * self.compute_flux_spatial()
 
         if self.model.temporal_model:
             value *= self.compute_temporal_norm()
@@ -2599,7 +2599,7 @@ class MapEvaluator:
 
             if self.psf and self.model.apply_irf["psf"]:
                 values = self.apply_psf(values)
-            value = (values.data * mask).sum(axis=(1,2))
+            value = (values.quantity * mask).sum(axis=(1, 2))
 
         else:
             value = self.model.spatial_model.integrate_geom(self.geom)
--- a/gammapy/datasets/map.py
+++ b/gammapy/datasets/map.py
@@ -2573,10 +2573,8 @@ class MapEvaluator:
         """Compute psf convolved and temporal model corrected flux."""
         value = self.compute_flux_spectral()
 
-        if self.model.spatial_model and not isinstance(self.geom, RegionGeom):
-            value = value * self.compute_flux_spatial().quantity
-        else:
-            value = value * self.compute_flux_spatial().quantity
+        if self.model.spatial_model:
+            value = value * self.compute_flux_spatial()#.quantity
 
         if self.model.temporal_model:
             value *= self.compute_temporal_norm()
@@ -2593,13 +2591,12 @@ class MapEvaluator:
         """
         if isinstance(self.geom, RegionGeom):
             wcs_geom = self.geom.to_wcs_geom()
+            mask = self.geom.contains(wcs_geom.get_coord())
             values = self.model.spatial_model.integrate_geom(wcs_geom)
-            tmp_map = Map.from_geom(geom=wcs_geom, data=values.data,
-                                    unit=values.unit)
 
             if self.psf and self.model.apply_irf["psf"]:
-                values = self.apply_psf(tmp_map)
-            value = ((values.data * wcs_geom.solid_angle())).sum()
+                values = self.apply_psf(values)
+            value = (values.data[mask]).sum()
 
         else:
             values = self.model.spatial_model.integrate_geom(self.geom)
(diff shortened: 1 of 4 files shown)


gammapy/datasets/map.py
```diff
@@ -2571,17 +2575,37 @@ def compute_flux_psf_convolved(self):
 
         if self.model.spatial_model and not isinstance(self.geom, RegionGeom):
             value = value * self.compute_flux_spatial().quantity
+        else:
```
adonath, line 2578: Just remove the `not isinstance(self.geom, RegionGeom)` before...

gammapy/datasets/map.py
```diff
@@ -2571,17 +2575,37 @@ def compute_flux_psf_convolved(self):
...
+
+        Returns
+        ----------
+        value: `~astropy.units.Quantity`
+            Psf-corrected, integrated flux over a given region.
+        """
+        if isinstance(self.geom, RegionGeom):
+            wcs_geom = self.geom.to_wcs_geom()
+            values = self.model.spatial_model.integrate_geom(wcs_geom)
+            tmp_map = Map.from_geom(geom=wcs_geom, data=values.data,
```
adonath, line 2597: `.interate_geom()` already returns a `Map` object...no need to create it here...

gammapy/datasets/map.py
```diff
@@ -2571,17 +2575,37 @@ def compute_flux_psf_convolved(self):
...
+                                    unit=values.unit)
+
+            if self.psf and self.model.apply_irf["psf"]:
+                values = self.apply_psf(tmp_map)
+            value = ((values.data * wcs_geom.solid_angle())).sum()
```
adonath, line 2602: This misses the application of the region mask, that you had in your previous implementation. As you have called `.integrate_geom()` already, there is no need to multiply again with the solid angle...

adonath (changes requested):
Thanks @fabiopintore, I have left a few comments. Please add a unit test now to check the implementation...

gammapy/datasets/tests/test_map.py
```diff
@@ -1361,3 +1361,46 @@ def test_downsample_onoff():
...
+    background_region = RegionNDMap.create(region, axes=[energy_axis_true])
+    background_region.data += 2
+
+    dataset_region = MapDataset(psf=psf, exposure=exposure_region, background=background_region)
+    dataset_region.models = [model]
+
+    npred_region = dataset_region.npred()
+
+    # for a WCSmap
+    exposure_wcs = Map.create(axes=[energy_axis_true], region=region, map_type="region")
```
adonath, line 1395: Note this does not create a `WCSMap`, but again a `RegionMap`, for a WCS map you need to choose `map_type="wcs"`...please adapt the test.

adonath (changes requested):
Thanks @fabiopintore, I would suggest to add a second test using a circular region of the size  of 1 sigma of the Gaussian PSF and check the spread corresponds to ~39%.

fabiopintore:
I added a further test to check if a circular region of the 1sigma value of the PSF contains the correct amount of flux. 
@adonath, @LauraOlivera  please may you check if I made the right things? The tests pass, but it seems to me that the predicted counts, when using a `RegionGeom`, are too much than expected for the given exposure...

gammapy/datasets/map.py
```diff
@@ -2569,19 +2573,36 @@ def compute_flux_psf_convolved(self):
...
+            Psf-corrected, integrated flux over a given region.
+        """
+        if isinstance(self.geom, RegionGeom):
+            wcs_geom = self.geom.to_wcs_geom()
+            mask = self.geom.contains(wcs_geom.get_coord())
+            values = self.model.spatial_model.integrate_geom(wcs_geom)
+
+            if self.psf and self.model.apply_irf["psf"]:
+                values = self.apply_psf(values)
+            value = (values.data[mask]).sum()
```
adonath, line 2599: I started to test implementation and I noticed a bug in this line: we have to sum only over the spatial axes here and not all axes (and not including the energy axis). So the statement here should be something like `.sum(axis=(1, 2))`... I need to check whether the implementation in `SpatialModel.integrate_geom()` is correct.

gammapy/datasets/tests/test_map.py
```diff
@@ -1361,3 +1361,67 @@ def test_downsample_onoff():
     assert downsampled.counts.data.sum() == dataset_onoff.counts.data.sum()
     assert downsampled.counts_off.data.sum() == dataset_onoff.counts_off.data.sum()
     assert_allclose(downsampled.alpha.data, 0.5)
+
+
+def test_compute_flux_spatial():
```
adonath, line 1366: I realised this test case is maybe a bit to high level. What about if we start on the `MapEvaluator` level?
So define:
- Constant exposure (filled with ones...) with 2 energy bins
- Point source models with a `ConstSpectralModel`
- Gaussian PSF kernel, created with `PSFKernel.from_gauss(geom=region_geom.to_wc_geom())`
- Create a `MapEvaluator()`
- `MapEvaluator.compute_flux_spatial()` should then return exactly the PSF containment per energy as an array like [0.39, 0.39]...

adonath (changes requested):
Thanks @fabiopintore, I noticed a bug in the implementation and made a proposal for a better unit test...

adonath (approved):
Thanks @fabiopintore!

PR #3146, opened 2020-12-01
Move dataset I/O code

adonath:
**Description**

This pull request moves the code from `.to_ogip_files()` and `.from_ogip_files()` to `datasets/io.py`.  As `.to_ogip_files()` is one of the longest functions we had in Gammapy I wanted to refactor and split it. For this I introduced a new **non-public** API: `OGIPDatasetReader` and `OGIPDatasetWriter`. The motivation mostly was to simplify the code, but I can easily imaging that at some point the `Reader` and `Writer` API will become public and can be used in pipeline configurations...for now it remains an implementation detail. I think it's also likely that we will support further formats in future so I think splitting out the code and allow for a registry in future is a good first step.

@registerrier I realised we currently don't support `SpectrumDatasetOnOff.read()`. I think we should just change it to something like `SpectrumDatasetOnOff.read(filename, format="ogip")` for now.

### DIFF
--- a/gammapy/datasets/io.py
+++ b/gammapy/datasets/io.py
@@ -12,8 +12,12 @@ from gammapy.irf import EDispKernelMap, EDispKernel
 from .spectrum import SpectrumDatasetOnOff
 
 
+__all__ = ["DatasetWriter", "DatasetReader", "OGIPDatasetReader", "OGIPDatasetWriter"]
+
+
 class DatasetReader(abc.ABC):
     """Dataset reader base class"""
+
     @property
     @abc.abstractmethod
     def tag(self):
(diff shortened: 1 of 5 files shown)


registerrier (approved):

Note that there is no specific reason why OGIP format can't be used to store a regular `SpectrumDataset`. It is only the way one will use the content of the background file. 
There might be keywords set in the header to help determine how to use this.  Alternatively one could leave an option here to create one dataset type or the other?

This might be a short term solution for the problem of serializing `SpectrumDataset`

adonath:
Thanks @registerrier, indeed I have never thought about this, but the data stored in `SpectrumDataset` is a subset of what is stored in the `SpectrumDatasetOnOff`. I guess the `background` could be stored in the background file and the back scale remains undefined...

adonath:
Thanks @registerrier!

PR #3148, opened 2020-12-02
Add various models and parameters management options

QRemy:
This PR add various models and parameters management options:

- modify Models.select() to take only a boolean mask as input
- add Models.mask() that return a boolean mask combining the different conditions with AND operator. The separation between mask and select allows to build more complex selections using other operators.
- add conditions for Models.mask() based on spatial model tag, spectral model tag, and frozen status

- rename parameters.restore_value as parameters.restore_status
- add option restore_values (True by default, if False only the frozen status is restored).
 TODO: also restore the covariance (reset to zero if frozen is set to True) ?

- add Models.set_parameters_bounds() in order to set min/max/value for all the selected models types and parameters names.

- add Models.freeze()/unfreeze() methods
-- freeze: freeze all parameters or only spatial or only spectral 
-- unfreeze: restore frozen status to default for all parameters or only spatial or only spectral

- add all new tests in gammapy.modeling.models/test/test_management.py

Further models management related to Datasets operations, like .cutout(), will be added in a separate PR.

### DIFF
--- a/gammapy/modeling/models/core.py
+++ b/gammapy/modeling/models/core.py
@@ -196,23 +196,37 @@ class Model:
         
         Parameters
         ----------
-        model_type : {None, "spatial", "spectral"}
+        model_type : {None, "spatial", "spectral", "temporal"}
            freeze all parameters or only spatial or only spectral 
         """
 
         if model_type is None or self.type == model_type:
             self.parameters.freeze_all()
-        elif model_type == "spatial" and hasattr(self, "spatial_model"):
+        elif (
+            model_type == "spatial"
+            and hasattr(self, "spatial_model")
+            and self.spatial_model
+        ):
             self.spatial_model.parameters.freeze_all()
-        elif model_type == "spectral" and hasattr(self, "_spectral_model"):
+        elif (
+            model_type == "spectral"
+            and hasattr(self, "_spectral_model")
+            and self._spectral_model
+        ):
             self._spectral_model.parameters.freeze_all()
+        elif (
+            model_type == "temporal"
+            and hasattr(self, "_spectral_model")
+            and self.temporal_model
+        ):
+            self.temporal_model.parameters.freeze_all()
 
     def unfreeze(self, model_type=None):
         """Restore parameters frozen status to default depending on model type
         
         Parameters
         ----------
-        model_type : {None, "spatial", "spectral"}
+        model_type : {None, "spatial", "spectral", "temporal"}
            restore frozen status to default for all parameters or only spatial or only spectral
         """
 
@@ -220,11 +234,14 @@ class Model:
             if self.type == model_type or hasattr(self, "spatial_model"):
                 self._unfree_to_default()
         if model_type is None or model_type == "spatial":
-            if hasattr(self, "spatial_model"):
+            if hasattr(self, "spatial_model") and self.spatial_model:
                 self.spatial_model._unfree_to_default()
         if model_type is None or model_type == "spectral":
-            if hasattr(self, "_spectral_model"):
+            if hasattr(self, "_spectral_model") and self._spectral_model:
                 self._spectral_model._unfree_to_default()
+        if model_type is None or model_type == "temporal":
+            if hasattr(self, "temporal_model") and self.temporal_model:
+                self.temporal_model._unfree_to_default()
 
     def _unfree_to_default(self):
         try:
@@ -593,6 +610,8 @@ class DatasetModels(collections.abc.Sequence):
                     sub_model = getattr(model, "spatial_model", None)
                 elif model_type == "spectral":
                     sub_model = getattr(model, "_spectral_model", None)
+                elif model_type == "temporal":
+                    sub_model = getattr(model, "temporal_model", None)
                 if sub_model:
                     inlist = np.any([t in sub_model.tag for t in tag])
                     selection[km] &= tag == sub_model.tag or inlist
@@ -641,13 +660,16 @@ class DatasetModels(collections.abc.Sequence):
         """
 
         models = self.select(tag=tag, model_type=model_type)
-        parameters = models.parameters.select(name=parameters_names)
+        parameters = models.parameters.select(
+            name=parameters_names, model_type=model_type
+        )
+        n = len(parameters)
         if min is not None:
-            parameters.mins = min
+            parameters.minima = np.ones(n) * min
         if max is not None:
-            parameters.maxs = min
+            parameters.maxima = np.ones(n) * max
         if value is not None:
-            parameters.values = value
+            parameters.values = np.ones(n) * value
 
     def freeze(self, model_type=None):
         """Freeze parameters depending on model type
--- a/gammapy/modeling/models/core.py
+++ b/gammapy/modeling/models/core.py
@@ -191,65 +191,14 @@ class Model:
         """Frozen status of a model, True if all parameters are frozen """
         return np.all([p.frozen for p in self.parameters])
 
-    def freeze(self, model_type=None):
-        """Freeze parameters depending on model type
-        
-        Parameters
-        ----------
-        model_type : {None, "spatial", "spectral", "temporal"}
-           freeze all parameters or only spatial or only spectral 
-        """
-
-        if model_type is None or self.type == model_type:
-            self.parameters.freeze_all()
-        elif (
-            model_type == "spatial"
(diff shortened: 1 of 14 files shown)


gammapy/modeling/models/core.py
```diff
@@ -455,43 +455,186 @@ def copy(self):
...
+                    selection[km] &= False
+            if frozen is not None:
+                all_frozen = np.all([p.frozen for p in model.parameters])
+                if frozen == True:
+                    selection[km] &= all_frozen
+                else:
+                    selection[km] &= ~all_frozen
+        return selection
+
+    def set_parameters_bounds(
```
adonath, line 550: It's not documented, but in general users are free to overwrite defaults of Gammapy models on the class level e.g.:

```python3
from gammay.modeling.models import PowerLawSpectralModel, Parameter

PowerLawSpectralModel.index = Parameter("index", value=3, min=0, max=5)
```
However I'm not sure it's pattern we should advertise, need to think about it once more....

adonath (changes requested):
Thanks a lot @QRemy, functionality wise this is really nice! I have a few comments on the code location and API. In general would you mind to split this PR and bring in the global functionality across multiple PRs? This makes it easier to review and understand.

My proposal would be to start on the `Parameters` level:
- Introduce `.restore_status()`
- Introduce `Parameters.unfreeze_all()`
- Introduce support for index arrays in `Parameters.__getitem__`
- Introduce `Parameters.select(type=["spatial", "spectral"], name=, name_substring=)`
- Introduce `Parameters.mins` and `Parameters.maxs`, with the corresponding setters, so that something like this works `parameters.mins = 5` 

Continue on the `Model` level:
- Introduce `Model.freeze()`
- Introduce `Model.unfreeze()` (here one can rely on the default parameters)
- Introduce `Model.frozen` property

Continue on the `SkyModel` level:
- Introduce `SkyModel.unfreeze(...)` (need to call `.unfreeze()` per model component)

An finally on the `Models` level:
- Introduce support for index arrays in `Models.__getitem__`
- Introduce `Models.select(...)`
- Introduce `Models.freeze(...)`
- Introduce `Models.unfreeze(...)`
- Introduce `Model.frozen` property

If you prefer you can still implement everything in this PR, but it will lead to a large diff...

gammapy/modeling/models/core.py
QRemy: but set_parameters_bounds() can change the settings a posteriori

QRemy:
Thanks for the suggestions.

I don't see how models.parameters.select(type="spectral", name=, name_substring=) can work as we cannot tell if a parameter is spatial or spectral, especially as we do not guaranty the uniqueness of the parameters names across all models.

gammapy/modeling/models/core.py
```diff
@@ -455,43 +455,186 @@ def copy(self):
...
 
-        for model in self:
-            selection = True
+        return self.__class__(models)
 
-            if dataset_name:
-                selection &= (
-                    model.datasets_names is None or dataset_name in model.datasets_names
-                )
+    def mask(
```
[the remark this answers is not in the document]
QRemy: I think we need a method that return the mask,  so .select() return a mask and then self[self.select()] return the new list, ok ?

gammapy/modeling/models/core.py
adonath: Indeed, my first comment was not meant to replace a functionality, but just as an idea that came to my mind...

adonath:
@QRemy Note the type of parameter is already accessible on the `Parameter` object with some "descriptor magic":
```python3
pwl = PowerLawSpectralModel()
print(pwl.amplitude.type)
```
I think this pattern is stable, as parameters are never shared between different model types. But it cannot easily be extended to e.g access the model name on the parameter or similar...

adonath:
Some more comments:
- Indeed `.restore_status()` can stay as is.
- Yes, `Parameters.unfreeze_all()` would unfreeze parameters which are frozen by default. So we do not rely on it in e.g. `Model.unfreeze()`, however I think it can still be useful e.g. like `models.parameters.select(name="norm").unfreeze_all()`

gammapy/modeling/models/core.py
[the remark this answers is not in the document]
adonath: Right now I don't see the use-case for exposing the mask to users, however internally it's of course useful. I think we have the same pattern in `EventList.select_foobar()`

QRemy:
- OK for the parameter type.
- For the problem of same names with different models, I guess ones have to apply models.select() and then parameters.select() otherwise that's quite dangerous and error prone. So do you really want this parameters.select() ?

gammapy/modeling/models/core.py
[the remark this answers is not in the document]
QRemy: It is necessary to combine conditions with other operators like or, not...

adonath:
@QRemy Yes, I think `Parameters.select()` is still useful, even if it's used mostly internally in combination with `models.select()`.

gammapy/modeling/models/core.py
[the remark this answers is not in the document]
adonath: I would still prefer if `.select()`, remains a one line method for users, but we could simply add something like `.get_selection_mask()` and `.select()` just calls into it...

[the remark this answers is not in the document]
QRemy: ok then I will add a boolean mask in the possible parameters for .select()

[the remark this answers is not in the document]
adonath: My proposal would be to allow for numpy like indexing such that `models[mask]` directly works...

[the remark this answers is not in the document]
adonath: ...however I'm not sure if this is easy to a achieve, but maybe you can give it a try.

gammapy/modeling/parameter.py
```diff
@@ -332,6 +332,34 @@ def types(self):
         """Parameter types"""
         return [par.type for par in self]
 
+    @property
+    def minima(self):
```
adonath, line 336: Here I'm leaning 60% towards calling this `mins` , I agree "minima" sounds better, but the connection between `Parameter.min` and `Parameters.mins` seems stronger. Any opinion @registerrier, @AtreyeeS?

adonath (changes requested):
Thanks a lot @QRemy, I have  left a few final comments. Once those are addressed (let's wait for @registerrier or @AtreyeeS input on the names), the PR can go in...

gammapy/modeling/parameter.py
QRemy: `maxs` sound very odd, no ?

adonath: Yes, I agree it sounds odd, but the connection to `Parameter.max` is stronger...maybe @registerrier or @AtreyeeS have a preference, I don't insist in any way...

registerrier: Well, I am not fond of `mins` and `maxs` which do not sound very clear to me. But it is likely that `minima` and `maxima` are also ambiguous w.r.t. optimal parameter values (e.g. the the value which maximizes likelihood or minimize TS).
maybe a more explicit `range_min` and `range_max` or simply keep the `min` and `max` on the `Parameters` as well.

adonath: I think the logic of `mins` and `maxs` is just like `value` and `values`...what about if we just ignore the plural and mirror what we have for `Parameter`? So we introduce:

- `Parameters.min` and `Parameters.max`
- Rename  to `Parameters.value` (there is e.g. `Quantity.value`, which returns an array as well)

QRemy: ok done

gammapy/modeling/models/core.py
```diff
@@ -455,43 +477,179 @@ def copy(self):
...
+                    selection[km] &= False
+
+            if frozen is not None:
+                if frozen == True:
+                    selection[km] &= model.frozen
+                else:
+                    selection[km] &= ~model.frozen
+        return selection
+
+    def restore_status(self, restore_values=True):
```
adonath, line 577: This is addition is an excellent idea!

adonath (approved):
Thanks a lot @QRemy! This might need some final polishing, but I think we should just go ahead and merge now. Then @AtreyeeS can start to use it and work on the model management tutorial. Any feedback and suggestions of improvement she has, can be implemented in follow up PRs...

PR #3151, opened 2020-12-03
PIG 22: flux estimates API and estimators unification

registerrier:
**Description**

This pull request adds PIG 22 which proposes a cleaner flux estimates API as well as a unification and clarification of existing estimators.
 
**Dear reviewer**


adonath:
@registerrier I think formally we can withdraw the PIG, however there are still a few interesting ideas we should not forget about. Especially the general dataset grouping.

PR #3152, opened 2020-12-03
Unify dataset I/O interface

adonath:
**Description**

This pull request unifies the dataset read interface:
- Remove `SpectrumDatasetOnOff.to_ogip_files(use_sherpa)` and expose it via `.write(path, format="ogip")` and `.write(path, format="ogip-sherpa")`
- Remove `SpectrumDatasetOnOff.from_ogip_files()` and expose it via `.read()`
- Remove the model setting in `Dataset.copy()` to be uniform with the other dataset methods

### DIFF
--- a/gammapy/datasets/core.py
+++ b/gammapy/datasets/core.py
@@ -41,6 +41,16 @@ class Dataset(abc.ABC):
         pass
 
     @property
+    def name(self):
+        return self._name
+
+    def to_dict(self):
+        """Convert to dict for YAML serialization."""
+        name = self.name.replace(" ", "_")
+        filename = f"{name}.fits"
+        return {"name": self.name, "type": self.tag, "filename": filename}
+
+    @property
     def mask(self):
         """Combined fit and safe mask"""
         if self.mask_safe is not None and self.mask_fit is not None:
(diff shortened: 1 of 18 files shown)


gammapy/datasets/core.py
```diff
@@ -64,18 +74,23 @@ def stat_array(self):
...
-
-        # propagate new dataset name
-        if new._models is not None:
-            for m in new._models:
-                if m.datasets_names is not None:
-                    for k, d in enumerate(m.datasets_names):
-                        if d == self.name:
-                            m.datasets_names[k] = name
+        # TODO: check the model behaviour?
+        new.models = None
```
QRemy, line 93: I don't see a problem in keeping the models for .copy().

adonath: I agree it's not necessarily a problem, however the way it was used in our code, was basically to copy and modify the model, which I think is non-ideal, because it makes the copy of the data for no good reason. I think in general there are not many use-cases, where a copy of the data is needed, so I think we should advertise `Models.copy()` instead. In the rare cases, where the data is copied, one can easily re-set the model and get's a uniform behaviour with all the other dataset methods, so I'd prefer to keep this change.

QRemy: If users have to re-set the models by themselfs, they will have also to propagate the new datasets names which is not very convenient (using append rather than replace as there is no copy).
Concerning uniformity with the other methods, maybe you can add an option keep_models : bool, that adapt or reset the models, and I will introduce that in a separate PR for the other methods (already started for .cutout())

adonath: I'd be open to include something like this, but maybe we do things in order and start at the `Models` level first? Like:
- `Models.reassign(dataset_name=, dataset_name_new=)`
- `Models.select_region(region, margin="0.1 deg")`
- `Models.select_mask(mask, margin="0.1 deg")`
- Add `WcsGeom.footprint()`, which returns a region...

Then my proposal would be that we fix the `npred_background()` behaviour and only finally introduce the (optional) model handling. Does this sound reasonable?

QRemy: ok let's discuss that later

PR #3153, opened 2020-12-04
Refactor gammapy download

Bultako:
This PR is an important refactoring of `gammapy download` code to have a more simple and more maintainable code base. It also simplifies the release process. The most relevant modifs are the following:

- removal of downloading scripts feature, no collections of .py scripts provided neither tested
- removal of needed versioned index files for datasets and notebooks for future releases in `gammapy-webpage` repo
- removal of downloading specific datasets or notebooks, a notebooks tar file is needed in n `gammapy-webpage` repo
- removal of downloading tutorials feature, instead two commands needed to download notebooks and datasets
- addition of feature `gammapy jupyter tar` to create a tar file of notebooks present in the docs folder structure
- addition of function `gammapy.utils.script.get_notebooks_path()` to retrieve all notebooks present in docs 
- removal of `parfive` dependency

With these new modifs, static images used in notebooks should live in an `images` folder , updated versions of the notebooks will come in another PR. Also PRs needing new datasets not present now in `gammapy-data` repo would need the addition of these datasets to that repository to pass CI.


adonath (commented):
Thanks a lot @Bultako, this looks like a much simplified process now! I think it might be a good idea to integrate the creation of the notebook tar into our docs build process and upload the tar file to the docs repo instead. 

In addition we could created one box or button at the top of the page here: https://docs.gammapy.org/0.18.2/tutorials/index.html, which allows to download all notebooks as a tar file. However if we change to sphinx gallery for the notebooks, later the its probably not worth the effort now.

Bultako:
@adonath 

> I think it might be a good idea to integrate the creation of the notebook tar into our docs build process and upload the tar file to the docs repo instead.

The creation of the tar file is only needed in the release making process, but not in the docs build. In other words, we do not want to create a tar every time we update the dev docs, right? Maybe better to add a `make release` option that created the tar and automatize every possible step? I could try doing this in another PR.

> In addition we could created one box or button at the top of the page here: https://docs.gammapy.org/0.18.2/tutorials/index.html, which allows to download all notebooks as a tar file. However if we change to sphinx gallery for the notebooks, later the its probably not worth the effort now.
 
I hope we will move quite a good fraction of the tutorials to notebooks in the API docs, and in that case the tar file will also contain those notebooks (not only those present in the tutorials) Moreover, yes, there's a PR on gallery view for tutorials coming soon :)

adonath:
> The creation of the tar file is only needed in the release making process, but not in the docs build. In other words, we do not want to create a tar every time we update the dev docs, right? 

Why not? I think the notebook tar should be small of size and it could be directly integrated in the docs build, without a separate release step. In addition we then provide the download link in a box on the tutorials page at the top. Basically what we do with the other tutorials as well, but instead of the downloading a single `.ipynb` users can download the whole tar with all notebooks on the page...

Bultako:
Ok, so this would mean that the notebooks tar files will live in the different version folders of the `gammapy-docs` repo and that we will use the `gammapy-webpage` repo only to store the release environment yaml files. I think this can be done, the notebooks tar file is less than ~1MB in size. I think in that case we should also move the release environment yaml files to `gammapy-docs`, and even try to create them automatically.

adonath:
@Bultako Yes, this would be my proposed behaviour. No separate upload of the notebooks during a release, but just provid the download button in the docs. We could possibly keep a minimal index file on the webpage for `gammapy download` command, which just points to the place in the docs. Just in case e.g. we need to move our docs to a different server. But this index file would be minimal, like:
```yaml
v0.18.2: "https//:url/notebooks.tar.gz"
v1.0: "https//:other-url/notebooks.tar.gz"
```

Bultako:
@adonath 
I've made the functional modifs requested.

The versioned `notebooks.tar` bundle are published within the versioned documentations, together with the `gammapy-environemnt.yml` file in the `docs/_static/` folder. So there's no need to place versioned index files  for environment, notebooks, scripts, datasets or whatsoever anymore !! The `make docs-all` command does all the work. Moreover, I've added a link to the each published tutorial upper-box so to download the `notebooks.tar` bundle.

Bultako:
Some info related with the release process and where the versioned files that may be downloaded with `gammapy download` are hosted.

- Only the pair `gammapy-environment.yml` and `notebooks.tar` may be downloaded as versioned files.
- They are published in [gammapy-docs](https://github.com/gammapy/gammapy-docs) repo in `docs/_static/` folder for each release.
- The `docs/_static/` folder in this `gammapy` repo uses this pair to build the docs, only those in `gammapy-docs` repo  may be downloaded.
- The `make docs-all` command replaces the `notebooks.tar``docs/_static/` folder with thew last versions of all notebooks present in the docs.
- There's a Github action in `gammapy-docs` repo that automatically creates and publish a `dev` version of the docs.

The actions to make during the release (i.e. v0.19) process in the `gammapy-docs` repo has the following items.
- Build an updated version of the dev docs using Github action.
- Copy the `docs/dev` folder as a new `docs/0.19` folder.
- Edit `0.19/docs/_static/gammapy-environment.yml` file and adapt conda env name and dependency versions.
- Edit `stable/index.html` to point to 0.19.

.github/workflows/ci.yml
```diff
@@ -47,7 +47,7 @@ jobs:
         run: |
           python -m pip install --upgrade pip setuptools wheel
           pip install astropy click cython numpy pydantic pyyaml regions scipy
-          pip install pytest pytest-astropy
+          pip install pytest pytest-astropy requests tqdm
```
maxnoe, line 50: If you do not install your dependencies explicitly, you directly test if your setup.py lists the correct dependencies.

Instead, use `extras` and do something like `pip install '.[all]'`

adonath:
Thanks @Bultako after some more thought, let's keep the environment files in the webpage repo for now, it's likely we have to change them based on user feedback and then having the git tag one them is meaningless.

docs/_download/environment_dev.yml
```diff
@@ -0,0 +1,37 @@
+# Conda environment for Gammapy version dev
```
adonath, line 1: Let's keep the env files inbox the webpage repo for now...

docs/install/index.rst
```diff
@@ -16,20 +16,20 @@ Gammapy and it's dependencies by executing these commands in a terminal:
 
 .. code-block:: bash
 
-    curl -O https://gammapy.org/download/install/gammapy-0.18.2-environment.yml
-    conda env create -f gammapy-0.18.2-environment.yml
+    curl -O https://docs.gammapy.org/0.19/_download/environment_0.19.yml
```
adonath, line 19: This doesn't make sense to me, I think in the dev version we should still point to the last stable version for install, in this case v0.18.2

docs/install/index.rst
```diff
@@ -16,20 +16,20 @@ Gammapy and it's dependencies by executing these commands in a terminal:
...
     lines with ``sherpa`` and ``healpy``. Those are optional dependencies that
     currently aren't available on Windows.
 
-This creates a conda environment called ``gammapy-0.18.2`` which you can activate via:
+This creates a conda environment called ``gammapy-0.19`` which you can activate via:
 
 .. code-block:: bash
 
-    conda activate gammapy-0.18.2
+    conda activate gammapy-0.19
```
adonath, line 32: See comments above...

adonath (changes requested):
One more minor thing: can we use `notebooks-dev.tar.gz` and `notebooks-0.18.2.tar.gz`, this would keep consistency with our current folder names...

Bultako:
> Thanks @Bultako after some more thought, let's keep the environment files in the webpage repo for now, it's likely we have to change them based on user feedback and then having the git tag one them is meaningless.

@adonath 
I already copied the env files and the notebooks.tar file in the docs repo.
i.e. see the [0.18.2/_downloads](https://github.com/gammapy/gammapy-docs/tree/master/docs/0.18.2/_downloads) folder at the bottom, and all the other `downloads` folder in the their respective versioned doc folders. So, if we need to change them they are there. Is that what you mean?

docs/install/index.rst
Bultako: You're right I'll fix this.

Bultako:
> One more minor thing: can we use `notebooks-dev.tar.gz` and e.g. `notebooks-0.18.2.tar.gz`, this would keep consistency with our current folder names...?

@adonath 
You mean using a dash instead of an underscore?

adonath (approved):
Thanks a lot @Bultako, I finally understood most of the setup and I think this PR nicely simplifies the download of the notebooks! I'll go ahead and merge now.

PR #3155, opened 2020-12-07
Add a tutorials notebooks gallery layout

Bultako:
This PR adds a gallery view feature for the tutorials index page.  

For every notebook, the thumbnail that will be displayed in the gallery has been selected adding `"tags": ["nbsphinx-thumbnail"]` in the metadata of the code cell that creates the image selected.

The `modeling_2d.ipynb` notebook does not have any image to display.

In order to create the gallery the empty notebooks have to be filled/executed, which is done only when we run the command `make docs-all`. Hence the command `make docs-sphinx`, though it will not fail, it will create a gallery with *broken* thumbnails. 

You can check how the tutorials gallery looks like [here](https://www.iaa.csic.es/~jer/gammapy/tutorials/index.html).


adonath:
Please provide feedback @registerrier, @AtreyeeS, @fabiopintore, @LauraOlivera, @luca-giunti, @facero, @bkhelifi, @Jaleleddine, @QRemy

.gitignore
```diff
@@ -87,4 +87,7 @@ dev/crab
 #!tutorials/images/*
 *.fits
 *.fits.gz
+*.fits_arf.gz
```
adonath, line 90: Oops this is a bug that I introduced in #3152. It seems `pathlib` does not recognise `.fits.gz` as a suffix but just `.gz`. I'll fix this in a separate PR and the you can remove this again...

adonath: Fixed in master: https://github.com/gammapy/gammapy/commit/1165717638b54712292d46b501e712e39f9f63eb

adonath:
Looking at the preview, I think I like this. It improves the visual structure of the tutorials page and with some further an even better choice of thumbnails (I think it is already good!) and better one sentence descriptions this is very going to be nice. Some thoughts on the general structure (independent of the Gallery):
- I think the "overview" should go in the Getting Started section
- I think the spectral analysis of extended sources should got in the spectral analysis section
- Maybe the 1D and 3D binned simulation should go in the 1D and 3D section, for the lightcurve simulation should go into the simulation section as well-
- I think the Fermi-LAT tutorial could go in the 3D section as well
- I think the scripts section should be at the bottom, after the extra topics

registerrier:
I agree this looks really nice and is much more accessible that the current page.
Maybe we don't need the high granularity in the left column menu for tutorials then. It is not very nice to have all the API menu going out of the screen when selecting one subcategory of tutorials.

facero:
This is very nice and helps to browse the tutorials. 
Maybe this does not exist or is too complicated but is it possible to add a short description when the cursor hovers the image ? 
For the low level analysis the description could the objective:
`Objective: Create a 3D dataset of the Crab using the H.E.S.S. DL3 data release 1 and perform a simple model fitting of the Crab nebula using the lower level gammapy API.`
But not all the notebooks have an objective apparently.

Bultako:
@facero 

> Maybe this does not exist or is too complicated but is it possible to add a short description when the cursor hovers the image ?

It is not complicated.
See the *Using Cell Metadata to Select a Thumbnail* thumbnail [here](https://nbsphinx.readthedocs.io/en/0.8.0/subdir/gallery.html) at the bottom of the page for a preview.
It would be more complex to modify the tooltip style..

facero:
The default tooltip style is great. The current issue is that there is no **Objective** for each notebook right ? 
In general this could be helpful to know in 1 or 2 sentences what the notebook is about. This cell would then be marked for the descriptor.

registerrier:
Indeed, we should have a more uniform structure for tutorials and some miss objective section. This is something we need to add for the documentation issues.

adonath:
@registerrier Any specific proposal for this PR already?

Concerning the tooltips: @Bultako maybe you add a first proposal here and then we improve in follow up PRs? Or would you prefer to merge this PR sooner and do improvements later...?

Bultako:
I can make a first proposal for tooltips and grouping structure in this PR, and then improve in a later PRs (I know editing notebooks metadata can be cumbersome, but this is what we have..). We will also need to move notebooks to API docs, for this I will make a WIP/draft PR and we could make the improvements there.

AtreyeeS:
This is really nice! 
I like @facero suggestion of showing the **Objectives** on mouse-hover.
If @Bultako can implement the feature now, we can improve upon them

Bultako:
You may find a first proposal for tooltips and general grouping structure for the tutorials gallery in the [preview](https://www.iaa.csic.es/~jer/gammapy/tutorials/index.html).

adonath:
Thanks @Bultako! As I mentioned during the dev meeting, I think with the additional info such as the thumbnail and tooltip, we don't need this fine grained categories for the tutorials anymore. Here is my proposal for the re-structuring:

- Move extended source analysis to 1D section
- Move 1D simulation to 1D section
- Move 3D simulation to 3D section
- Move dark matter to API tutorials
- Move pulsar tutorial to time section
- Move source detection to 2D image section
- Move sensitivity to 1D section
- MCMC shampooing could go in 3D analysis section

I think we should split the Fermi-LAT tutorial and keep one which shows the data access and one with the data reduction and the analysis part, but this will be a separate PR

Bultako:
@adonath 

I've made the requested modifs and updates the preview webpage.
There's still a big group **Package** that could remove and place the notebooks as part of the API docs in another PR.

adonath (approved):
Thanks a lot @Bultako! It think it's time to merge this and improve in follow up PRs instead...

PR #3157, opened 2020-12-08
Add .reassign_dataset() method to models

QRemy:
This PR add a .reassign_dataset(dataset_name, new_dataset_name) method to Model and Models 
in order to update the datasets_names replacing one name by another. This is used for example in the flux estimator, it could be also  used after the copy of a dataset.

### DIFF
--- a/gammapy/modeling/models/core.py
+++ b/gammapy/modeling/models/core.py
@@ -205,18 +205,18 @@ class Model:
         
         Parameters
         ----------
-        dataset_name : str or list
+        datasets_names : str or list
             Name of the datasets where the model is currently defined
-        new_dataset_name : str or list
+        new_datasets_names : str or list
             Name of the datasets where the model should be defined instead.
             If multiple names are given the two list must have the save lenght,
             as the reassignment is element-wise.
         """
-        model = self.copy()
+        model = self.copy(name=self.name)
         if not isinstance(datasets_names, list):
             datasets_names = [datasets_names]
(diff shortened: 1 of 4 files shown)


gammapy/modeling/models/core.py
```diff
@@ -651,6 +670,19 @@ def frozen(self):
...
+        Parameters
+        ----------
+        dataset_name : str
+            Name of a dataset where the model is currently defined
+        new_dataset_name : str
+            Name of a dataset where the model should be defined instead
+        """
+        for m in self:
+            m.reassign_dataset(dataset_name, new_dataset_name)
+
```
adonath, line 685: I think we should make a copy of the model and return it here...

adonath (changes requested):
Thanks @QRemy, I have two main points:
- I think the `Model.reassign()` method should make a copy of the model and return it. This way the model assignment becomes a "one liner"
- I think `Model.reassign()` should be able to handle multiple dataset names or possibly take a mapping between the old and new names like:
```python3
slices = datasets.slice_by_energy()
names = dict(zip(datasets.names, slices.names))
slices.models = datasets.models.reassign(names)
```

adonath (approved):
Thanks @QRemy, I'll merge this now and continue with the `npred_background()` handling. I'm sure we will adapt the handling of the models (like when two make copies, renames etc.) some more in follow up PRs. But I'd like to get the gain some experience with this first...

PR #3158, opened 2020-12-09
Add binary_erosion/dilation to WcsNDMap

QRemy:
This PR add  binary_erosion and binary_dilation methods to WcsNDMap.
These methods are applied only if the map is a boolean mask.
The mask erosion can be used for example to reduce mask_fit by the width of the PSF after a dataset.cutout() operation.

### DIFF
--- a/gammapy/maps/wcsnd.py
+++ b/gammapy/maps/wcsnd.py
@@ -176,9 +176,7 @@ class WcsNDMap(WcsMap):
                 for x in pix
             ]
         )
-        return ndi.map_coordinates(
-            self.data.T, pix, order=order, mode="nearest"
-        )
+        return ndi.map_coordinates(self.data.T, pix, order=order, mode="nearest")
 
     def _interp_by_coord_griddata(self, coords, interp=None):
         order = interp_to_order(interp)
@@ -312,9 +310,7 @@ class WcsNDMap(WcsMap):
         else:
             data = self.data
 
-        data = ndi.map_coordinates(
-            data.T, tuple(pix), order=order, mode="nearest"
-        )
+        data = ndi.map_coordinates(data.T, tuple(pix), order=order, mode="nearest")
 
         if preserve_counts:
             data *= geom.bin_volume().value
@@ -600,22 +596,28 @@ class WcsNDMap(WcsMap):
             If True apply the margin only from edges (Default is False).
         """
         if self.data.dtype != bool:
-            raise(TypeError, "Binary erosion should be applied only on boolean mask")
+            raise (TypeError, "Binary erosion should be applied only on boolean mask")
         if from_edges:
             mask_margin = np.ones(self.data.shape[-2:], dtype=bool)
         else:
-            #TODO: should we apply the erosin independently to each other dim
+            # TODO: should we apply the erosin independently to each other dim
             # for now just combine non-spatial dim with OR condition
             axis = tuple(np.arange(len(self.geom.axes)))
-            mask_margin = np.sum(self.data, axis=axis).astype(bool) 
+            mask_margin = np.sum(self.data, axis=axis).astype(bool)
         if not isinstance(margin, tuple):
             margin = (margin, margin)
-        shape = tuple([int(np.ceil(x / scale).value * 2 + 1) for x, scale in zip(margin, self.geom.pixel_scales)])
-        mask_margin =  ndi.binary_erosion(mask_margin[..., :, :], structure=np.ones(shape))
+        shape = tuple(
+            [
+                int(np.ceil(x / scale).value * 2 + 1)
+                for x, scale in zip(margin, self.geom.pixel_scales)
+            ]
+        )
+        mask_margin = ndi.binary_erosion(
+            mask_margin[..., :, :], structure=np.ones(shape)
+        )
         idx_y, idx_x = np.where(~mask_margin)
         self.data[..., idx_y, idx_x] = False
 
-            
     def binary_dilation(self, margin):
         """Binary dilation of boolean mask addding a given margin
         Parameters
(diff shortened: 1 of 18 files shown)


gammapy/maps/wcsnd.py
```diff
@@ -589,6 +585,63 @@ def get_spectrum(self, region=None, func=np.nansum, weights=None):
 
         return self.to_region_nd_map(region=region, func=func, weights=weights)
 
+    def binary_erosion(self, margin, from_edges=False):
+        """Binary erosion of boolean mask removing a given margin
+        Parameters
```
adonath, line 590: Missing empty line...

gammapy/maps/wcsnd.py
```diff
@@ -589,6 +585,63 @@ def get_spectrum(self, region=None, func=np.nansum, weights=None):
 
         return self.to_region_nd_map(region=region, func=func, weights=weights)
 
+    def binary_erosion(self, margin, from_edges=False):
```
adonath, line 588: My proposal would be to make this more like `WcsNDMap.smooth()`, so provide a `width` and `kernel` or `structure` argument and offer the options `box` and `disk`. Also allow for additional keyword forwarding to `binary_erosion()` for the boundary handling.

gammapy/maps/wcsnd.py
```diff
@@ -589,6 +585,63 @@ def get_spectrum(self, region=None, func=np.nansum, weights=None):
...
+            If only one value is passed, the same margin is applied in (lon, lat).
+        from_edge : bool
+            If True apply the margin only from edges (Default is False).
+        """
+        if self.data.dtype != bool:
+            raise (TypeError, "Binary erosion should be applied only on boolean mask")
+        if from_edges:
+            mask_margin = np.ones(self.data.shape[-2:], dtype=bool)
+        else:
+            # TODO: should we apply the erosin independently to each other dim
```
adonath, line 603: Again I would make it more similar to `.convolve()` and `.smooth()`: so for now apply the erosion, per image plane. This offers the possibility to handle energy dependent "kernels" in future (which seems reasonable, as e.g. the PSF size is energy dependent as well.).

gammapy/maps/wcsnd.py
```diff
@@ -589,6 +585,63 @@ def get_spectrum(self, region=None, func=np.nansum, weights=None):
...
+        ----------
+        margin : tuple of `~astropy.units.Quantity`
+            Angular sizes of the margin in (lon, lat) in that specific order.
+            If only one value is passed, the same margin is applied in (lon, lat).
+        from_edge : bool
+            If True apply the margin only from edges (Default is False).
+        """
+        if self.data.dtype != bool:
+            raise (TypeError, "Binary erosion should be applied only on boolean mask")
+        if from_edges:
```
adonath, line 600: This option I don't understand, I would expect `.binary_erosion()` to just act on `self.data` and not create any new mask. But we could e.g. offer something equivalent to `WcsGeom.region_mask()`, like `WcsGeom.boundary_mask(width=(lon, lat))` or similar...

gammapy/maps/wcsnd.py
```diff
@@ -589,6 +585,63 @@ def get_spectrum(self, region=None, func=np.nansum, weights=None):
...
+            [
+                int(np.ceil(x / scale).value * 2 + 1)
+                for x, scale in zip(margin, self.geom.pixel_scales)
+            ]
+        )
+        mask_margin = ndi.binary_erosion(
+            mask_margin[..., :, :], structure=np.ones(shape)
+        )
+        idx_y, idx_x = np.where(~mask_margin)
+        self.data[..., idx_y, idx_x] = False
```
adonath, line 619: I think we should return a new map here just as we do for `.smooth()` and `.convolve()`...

gammapy/maps/wcsnd.py
adonath, line 588: Minor: we typically name the methods after the verb, like `.convolve()` (and not `.convolution()`). So my proposed name here would either be `.erode()` or `.binary_erode()`.

adonath (changes requested):
Thanks @QRemy I have left some comments, mostly to make the new methods more uniform with to e.g. `.convolve()` or `.smooth()`.

gammapy/maps/wcsnd.py
QRemy: `from_edge` erode only the edges. The use case is the following :
 you have a dataset with mask_fit that mask some source already taking into account the psf, then you do a cutout in that case you want to apply the mask erosion only from the edges of the geom not based on the mask values.

QRemy: ok if you prefer to have it as a separate method

gammapy/maps/wcsnd.py
QRemy: - I renamed margin to width
- about the choice of option between `structure` or `kernel`, and if all `binary_erosion()` keywords are relevant for forwarding, I need to think about it, so maybe I could do that latter in a separate PR.

gammapy/maps/wcsnd.py
```diff
@@ -589,6 +604,42 @@ def get_spectrum(self, region=None, func=np.nansum, weights=None):
...
+        Parameters
+        ----------
+        width : tuple of `~astropy.units.Quantity`
+            Angular sizes of the margin in (lon, lat) in that specific order.
+            If only one value is passed, the same margin is applied in (lon, lat).
+        """
+        mask_data = _apply_binary_operations(self, width, ndi.binary_erosion)
+        return self._init_copy(data=mask_data)
+
+    def boundary_mask(self, width):
```
adonath, line 618: For consistency please move this to `WcsGeom.boundary_mask(width, inside=True)`. We already have `WcsGeom.region_mask()` which does a very similar thing. Internally you can still rely on the `.binary_erode()`, however there might be more efficient implementations...

adonath (changes requested):
Thanks @QRemy, I have left one comment regarding the code location `.boundary_mask()`, otherwise looks good to me. Once this is addressed the PR can go in...

gammapy/maps/wcsnd.py
QRemy: -moved .boundary_mask() from WcsNDMap to WcsGeom
-modified WcsGeom.region_mask() such as output is a WcsNDMap

adonath: Perfect thanks! In general it's maybe not a bad idea to have the mask creation on the `Map` level, because we have the pattern of "factory methods", such as `.create()`, `.from_geom()`, etc., but for now the consistency is more important...

adonath (approved):
Thanks a lot @QRemy, no further comments from my side...

PR #3161, opened 2020-12-10
Improve irf axis name checks

adonath:
**Description**

This pull request improves the check for IRF axis names and order. It mostly implements a `MapAxes.assert_names(required_names=)` method, which checks on the required axis names and order. In cases it does not match the following error is thrown:
```
ValueError: Incorrect axis order or names. Expected axis order: ['energy_true', 'rad'], got: ['rad', 'energy_true'].
```

I also fixes a bug in reading `Background3D` files, because for the creating of the energy axis we relied on the `EXTNAME` keyword which is optional in the GADF. This was changed to rely on `HDUCLAS2`, which is a mandatory keyword. During the implementation I noticed that, the IRF write methods are not compliant with the gadf format , as they miss required meta info. What do we do about this?

### DIFF
--- a/gammapy/irf/background.py
+++ b/gammapy/irf/background.py
@@ -69,6 +69,7 @@ class Background3D:
             data=data,
             interp_kwargs=interp_kwargs,
         )
+        self.data.axes.assert_names(["energy", "fov_lon", "fov_lat"])
         self.meta = meta or {}
 
     def __str__(self):
@@ -146,6 +147,7 @@ class Background3D:
         axes = MapAxes(self.data.axes[::-1])
         table = axes.to_table(format="gadf-dl3")
         table.meta = self.meta.copy()
+        table.meta["HDUCLAS2"] = "BKG"
         table["BKG"] = self.data.data.T[np.newaxis]
         return table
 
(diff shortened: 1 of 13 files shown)


registerrier (approved):
Thanks @adonath . The `MapAxes.assert_names()` is a really nice way to handle the issue of testing.

Fine with the change in `MapAxis.from_table` to handle correctly the `HDUCLASS2` requirement. Yet I think that format handling/checking should be done by the corresponding API level, here `irf`.  It is a bit inconsistent to have IRF formats at the `maps` level.  So we should try to have the mechanism outside of `MapAxis`.

adonath:
> It is a bit inconsistent to have IRF formats at the maps level. So we should try to have the mechanism outside of MapAxis.

In the current state I agree, but I think it's a valid assumption that IRF classes will be based on `Map` or similar as well at some point in future. Using `MapAxis` for IRFs as well was technically a nice improvement, because it allowed to introduce the axis scaling and handling `center` as well as `edges` values along with removing a lot of code duplication. So I'm almost 99.99% sure we will never have a separate `IRFAxis` class, but will stick with the model that a `MapAxis` can be serialised to multiple formats. However the code for the serialisation could live somewhere else, e.g. in `gammapy/irf/io.py`

PR #3162, opened 2020-12-11
Correct Models.from_dict to check input parameters names

registerrier:
**Description**

This pull request corrects the behavior of `Models.from_dict()`. Each model is set according to the input parameter name rather than relying on the order of default parameters.
If a parameter is missing a warning is sent to `log` and the default value is set.

Maybe one could check also that all parameters on input are actually used to avoid silent errors? 

**Dear reviewer**


adonath (approved):
Thanks a lot @registerrier, no further comments from my side...

PR #3165, opened 2020-12-14
Improve documentation of RegionGeom and RegionNDMap

LauraOlivera:
This pull request aims to expand the existing documentation for the `RegionGeom` and `RegionNDMap` classes.

This is done by adding missing docstrings and improving existing ones in the code, as well as with the creation of a `regionmap.rst` file.

Note that this is still a work in progress at the moment, as the RST file is missing the `RegionNDMap` section, and the rest is still in a draft state. I will keep working on it in the next days. However, general feedback at this stage is very welcome, since this is the first time I've ever contributed to gammapy documentation and I might be doing things in a less-than-optimal way.

So far I've just sort of listed what you can do with a `RegionGeom`, but I realize it might be maybe too much detail, and some of the code snippets have too much repetition. I intend to spend some more expanding more "conceptual" explanations, perhaps with examples in context of how you would normally use each of these objects.

### DIFF
--- a/docs/maps/regionmap.rst
+++ b/docs/maps/regionmap.rst
@@ -11,6 +11,7 @@ The RegionGeom and RegionNDMap
 This page provides examples and documentation specific to the Region
 classes. 
 
+
 RegionGeom
 ==========
 A `~RegionGeom` is analogous to a  map geometry `~Geom`, but instead of a fine grid on a rectangular region, 
@@ -301,17 +302,16 @@ created from an existing `~RegionGeom`, this can be done in the same step:
     region_map = RegionNDMap.create("icrs;circle(83.63, 22.01, 0.5)", axes=[energy_axis])
 
     # Fill the region map
-    #with the same value for each of the 12 energy bins
-    region_map.data = 1 
-    # or with an entry for each of the 12 energy bins
-    region_map.data = np.linspace(1,50,12) 
+    # with an entry for each of the 12 energy bins
+    region_map.data = np.logspace(-2,3,12) 
 
-    # Create another region map with the same RegionGeom but different data
+    # Create another region map with the same RegionGeom but different data,
+    # with the same value for each of the 12 energy bins
     geom = region_map.geom
-    region_map_2 = RegionNDMap.from_geom(geom, data = np.linspace(50,100,12))
+    region_map_2 = RegionNDMap.from_geom(geom, data = 1)
 
     # Access the data
-    print(region_map_2.data)
+    print(region_map_1.data)
 
 The data contained in a region map is a `~numpy.ndarray` with shape defined by the underlying
 `~RegionGeom.data_shape`. In the case of only spatial dimensions, the shape is just (1,1), one single
@@ -336,4 +336,72 @@ or a function of the non-spatial axis bins. This is done by `~RegionNDMap.plot()
         region_map.plot_region()
 
 + Plotting the map contents:
-    T
+    This is only possible if the region map has a non-spatial axis.
+
+    .. code-block:: python
+
+        from gammapy.maps import MapAxis, RegionNDMap
+        import numpy as np  
+
+        # Create a RegionNDMap with 12 energy bins
+        energy_axis = MapAxis.from_bounds(100., 1e5, 12, interp='log', name='energy', unit='GeV')
+        region_map = RegionNDMap.create("icrs;circle(83.63, 22.01, 0.5)", axes=[energy_axis])
+
+        # Fill the data along the energy axis and plot
+        region_map.data = np.logspace(-2,3,12) 
+        region_map.plot()
+
+    .. plot:: 
+
+        from gammapy.maps import MapAxis, RegionNDMap
+        import numpy as np  
+
+        # Create a RegionNDMap with 12 energy bins
+        energy_axis = MapAxis.from_bounds(100., 1e5, 12, interp='log', name='energy', unit='GeV')
+        region_map = RegionNDMap.create("icrs;circle(83.63, 22.01, 0.5)", axes=[energy_axis])
+
+        # Fill the data along the energy axis and plot
+        region_map.data = np.logspace(-2,3,12) 
+        region_map.plot()
+
+    Similarly, the map contents can also be plotted as a histogram:
+
+    .. code-block:: python
+
+        from gammapy.maps import MapAxis, RegionNDMap
+        import numpy as np  
+
+        # Create a RegionNDMap with 12 energy bins
+        energy_axis = MapAxis.from_bounds(100., 1e5, 12, interp='log', name='energy', unit='GeV')
+        region_map = RegionNDMap.create("icrs;circle(83.63, 22.01, 0.5)", axes=[energy_axis])
+
+        # Fill the data along the energy axis and plot
+        region_map.data = np.logspace(-2,3,12) 
+        region_map.plot_hist()
+
+    .. plot:: 
+
+        from gammapy.maps import MapAxis, RegionNDMap
+        import numpy as np  
+
+        # Create a RegionNDMap with 12 energy bins
+        energy_axis = MapAxis.from_bounds(100., 1e5, 12, interp='log', name='energy', unit='GeV')
+        region_map = RegionNDMap.create("icrs;circle(83.63, 22.01, 0.5)", axes=[energy_axis])
+
+        # Fill the data along the energy axis and plot
+        region_map.data = np.logspace(-2,3,12) 
+        region_map.plot_hist()
+
+
+Writing and reading a RegionNDMap to/from a FITS file
+-----------------------------------------------------
+Region maps can be written to and read from a FITS file with the
+`~RegionNDMap.write()` and `~RegionNDMap.read()` methods:
+
+.. code-block:: python
+
+    from gammapy.maps import RegionNDMap,MapAxis
+    energy_axis = MapAxis.from_bounds(100., 1e5, 12, interp='log', name='energy', unit='GeV')
+    m = RegionNDMap.create("icrs;circle(83.63, 22.01, 0.5)", axes=[energy_axis])
+    m.write('file.fits', overwrite=True)
+    m = RegionNDMap.read('file.fits')
--- a/docs/maps/regionmap.rst
+++ b/docs/maps/regionmap.rst
@@ -7,4 +7,17 @@ This page provides examples and documentation specific to the Region
 classes. 
 
 RegionGeom
+-----------
+Like a geometry but instead of a small grid on a rectangular region, 
+this is just one large pixel with an arbitrary shape that can have any 
+number of non-spatial dimensions.
+
+
+
+RegionNDMap
+-----------
+It's essentially a Map with one spatial bin that can have an arbitrary 
+shape and whatever non-spatial axis you want.
+It is to a RegionGeom what a Map is to a WcsGeom: it contains the data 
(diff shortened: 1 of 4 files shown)


docs/maps/regionmap.rst
```diff
@@ -0,0 +1,180 @@
...
+        center = SkyCoord("0 deg", "0 deg", frame="galactic")
+        region =  CircleSkyRegion(center=center, radius=1*u.deg)
+        geom = RegionGeom(region)
+
+        # Get coordinates and indexes defined by the region geometry
+        coordinates = geom.get_coord()
+        indexes = geom.get_idx()
+
+        # Check if a coordinate is contained in the region
+        >>> geom.contains(center)
```
adonath, line 119: This example could be extended to use an array of coordinate as well...

docs/maps/regionmap.rst
```diff
@@ -0,0 +1,180 @@
...
+        # Add an energy axis to the region geometry
+        energy_axis = MapAxis.from_bounds(100., 1e5, 12, interp='log', name='energy', unit='GeV')
+        geom_energy = geom.to_cube[energy_axis]
+
+        # Get the image region geometry without energy axis. 
+        # Note that geom_image == geom
+        geom_image = geom_energy.to_image()
+
+        # Get the minimal wcs geometry that contains the region
+        wcs_geom = geom.to_wcs_geom()
```
adonath, line 173: Maybe print the geometry here at the end....

gammapy/maps/region.py
```diff
@@ -170,6 +218,14 @@ def crop(self):
         raise NotImplementedError("Cropping of `RegionGeom` not supported")
 
     def solid_angle(self):
+        """Get solid angle of the region.
```
adonath, line 221: Here is some Gammapy developing weirdness: for `Map` objects the docstrings are inherited from the base class. If you check https://github.com/gammapy/gammapy/blob/master/gammapy/maps/geom.py#L2199, you can see, that the doctoring is already implemented. The problem is how to find whether it's missing? I think the best way for now is just to check the docstring interactively using `help(geom.solid_angle)`...if you think the docstring can be  improved, you can either edit th one in the base class, or re-implement it for the `RegionGeom` (where sometimes return types might differ...)

gammapy/maps/regionnd.py
```diff
@@ -189,6 +189,27 @@ def create(cls, region, axes=None, dtype="float32", meta=None, unit="", wcs=None
     def downsample(
         self, factor, preserve_counts=True, axis_name="energy", weights=None
     ):
+        """Downsample the non-spatial dimension by a given factor.
```
adonath, line 192: Those docstrings are defined in the base class as well, but I think here the behaviour slightly differs in the way that the default axis to downsample is "energy", while for a `WcsNDMap` it's the spatial axes. So definitely keep the docstring here!

adonath (commented):
Thanks a lot @LauraOlivera! I have left some preliminary comments, otherwise this already looks very good to me! Of course there is a number of things to add (plotting, plotting of regions), reading and writing etc.

We have some example files you could use for reading:
```python3
from gammapy.maps import RegionNDMap 

path = "$GAMMAPY_DATA/joint-crab/spectra/hess/"
counts = RegionNDMap.read(path + "pha_obs23523.fits")
print(counts.geom)
```
Unfortunately it does not define the region information. But might still be useful...the region info is currently optional, which might be worth to mention as well (but only for backwards compatibility...and reading existing file, which don't have the info).

In general just continue as you have done so far. Currently there is almost no documentation of the `RegionGeom` and `RegionNDMap`. So any documentation will be helpful and we don't have to deliver a 100% version in this PR, but rather get in the minimal content first!

docs/maps/regionmap.rst
```diff
@@ -0,0 +1,180 @@
...
+
+    from astropy.coordinates import SkyCoord
+    from regions import CircleSkyRegion
+    from gammapy.maps import MapAxis, RegionGeom
+    import astropy.units as u
+    center = SkyCoord("0 deg", "0 deg", frame="galactic")
+    region =  CircleSkyRegion(center=center, radius=1*u.deg)
+    energy_axis = MapAxis.from_bounds(100., 1e5, 12, interp='log', name='energy', unit='GeV')
+    geom = RegionGeom(region, axes=[energy_axis])
+
```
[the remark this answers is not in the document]
LauraOlivera: oh true, I had forgotten about this - I will add it, thanks!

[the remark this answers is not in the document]
adonath: It's good to show both ways anyway...

gammapy/maps/region.py
LauraOlivera: Oh right, I had noticed this "weirdness" too, but somehow missed that `solid_angle` already had a docstring. I already modified some that, even though they had the base class docstring, the text didn't quite describe the `RegionGeom` implementation of the method (like in the case in your comment below). I will be more careful and consistent with this though, thanks!

adonath: As long as the docstring you have now is more complete than the one in the base class, just keep it. For developers it can be also nice to have the docstring close to the implementation of the method.

LauraOlivera:
- I have more or less written about most things I wanted to write about. Please let me know if you think there is anything I should expand or change. The next step I think would be to link to this page from an existing page, and in that regard I think it would be nice if maybe there was a dedicated `Map` page, instead of it being the main one, and have an intermediate introductory page from which we link to this page, the `Map` and `HpxMap` pages... What do you think @AtreyeeS, @adonath ?

- I've also fixed a small bug from my PR #3167 because if the region had an energy axis, one of the cases failed. I fixed it adding `.to_image()` to the wcs geom for plotting.

- Another thing is that right now, the snippet for saving and writing `RegionNDMaps` to file doesn't work, and I think something is weird:
  
  ```
    from gammapy.maps import RegionNDMap,MapAxis
  
    energy_axis = MapAxis.from_bounds(100., 1e5, 12, interp='log', name='energy', unit='GeV')
    m = RegionNDMap.create("icrs;circle(83.63, 22.01, 0.5)", axes=[energy_axis])
    m.write('file.fits', overwrite=True)
    m = RegionNDMap.read('file.fits')
  ```
  results in:
  ```
    File "<stdin>", line 1, in <module>
    File "/nfs/d22/hin/nieto/gammapy/gammapy/maps/regionnd.py", line 308, in read
      return cls.from_hdulist(hdulist, format=format, ogip_column=ogip_column)
    File "/nfs/d22/hin/nieto/gammapy/gammapy/maps/regionnd.py", line 362, in from_hdulist
      table = Table.read(hdulist[hdu])
    File "/home/hin/nieto/d1/anaconda3/envs/gammapy-dev/lib/python3.7/site-packages/astropy/io/fits/hdu/hdulist.py", line 314, in __getitem__
      self._positive_index_of(key))
    File "/home/hin/nieto/d1/anaconda3/envs/gammapy-dev/lib/python3.7/site-packages/astropy/io/fits/hdu/hdulist.py", line 769, in _positive_index_of
      index = self.index_of(key)
    File "/home/hin/nieto/d1/anaconda3/envs/gammapy-dev/lib/python3.7/site-packages/astropy/io/fits/hdu/hdulist.py", line 750, in index_of
      raise KeyError(f'Extension {key!r} not found.')
  KeyError: "Extension 'SPECTRUM' not found."
  ```
  and as far as I can tell by looking at the `.read()` and `.write()` function, it should look for the hdu `'COUNTS'` and not `'SPECTRUM'`... what do you think?

adonath:
Thanks a lot @LauraOlivera, I'll investigate the issue with reading and writing now...

adonath:
@LauraOlivera The fix for the read / writ issue is contained here: #3171. I think we will have merge conflicts, so either we merge #3171 and you fix the conflicts and add the documentation on I/O here, or we merge this PR now and you add the documentation on I/O in a second PR. What do you prefer?

docs/maps/regionmap.rst
```diff
@@ -0,0 +1,419 @@
...
+    # using the regions package
+    region =  RectangleSkyRegion(center=center, width=1*u.deg, height=2*u.deg, angle=45*u.deg)
+    geom = RegionGeom(region)
+
+    # Equivalent factory method call
+    geom = RegionGeom.create(region) 
+
+
+Higher dimensional region geometries (cubes and hypercubes) can be constructed by passing a list of `~MapAxis` objects for non-spatial dimensions with the axes parameter:
```
AtreyeeS, line 61: Maybe mention "can be constructed **in exactly the same way as a WCSGeom**..."?

docs/maps/regionmap.rst
```diff
@@ -0,0 +1,419 @@
...
+
+            axes       : ['lon', 'lat']
+            shape      : (202, 202)
+            ndim       : 2
+            frame      : galactic
+            projection : TAN
+            center     : 0.0 deg, 0.0 deg
+            width      : 2.0 deg x 2.0 deg
+
+Plotting a RegionGeom
```
AtreyeeS, line 209: This is a very useful example. Linking it from the HowTo will be nice

LauraOlivera:
Thank you for fixing the write/read issue, @adonath! I think the best is that you merge #3171 and then I fix the conflicts and add the documentation in this same PR

AtreyeeS (dismissed):
Thanks @LauraOlivera ! Very nice documentation.

> The next step I think would be to link to this page from an existing page, and in that regard I think it would be nice if maybe there was a dedicated Map page, instead of it being the main one, and have an intermediate introductory page from which we link to this page, the Map and HpxMap pages

I totally agree! At present, I think you can link to it from the main maps page, https://docs.gammapy.org/dev/maps/index.html#using-gammapy-maps just after `HEALPix Based Maps`. This page is going to be modified in the near future I am sure, with all the code examples replaced by the maps notebook, but having a link to each of the supports `geom` should be useful

LauraOlivera:
I think it is done for now, @adonath @AtreyeeS. I linked to the page from the `Map` page and added some caveats about large regions and the WCS projection.

For the write/fix issue, I am confused. What I wrote is what works right now and almost verbatim from PR #3171, but my understanding was actually that saving and writing files was not going to be supported. Did I get it wrong? If so, I can change the snippet. I thought `RegionNDMaps` could only be saved while inside a `SpectrumDataset` but those methods are not defined... so I guess I just got confused :sweat_smile:

adonath:
@LauraOlivera Sorry this was mis-communicated from my side: the best mapping of the OGIP serialisation format (https://heasarc.gsfc.nasa.gov/docs/heasarc/ofwg/docs/spectra/ogip_92_007/node5.html) is indeed to a `SpectrumDatasetOnOff` (and not `SpectrumDataset`, however we could probably support this...).  The OGIP format defines a specification for additional quantities such as "QUALITY" (corresponds to "mask_safe") or "BACKSCALE" (corresponds to `acceptance_off`). 

However I checked once again the OGIP specification and noticed we can write a valid "standalone" PHA file at least for counts and exposure data. According to the specification one can write information such as "BACKSCALE" or "QUALITY" in the header of a FITS file as well, if it does not vary with energy. I decided to support this like so: https://github.com/gammapy/gammapy/blob/master/gammapy/maps/regionnd.py#L419. If a single counts spectrum is written to the format the required meta data is set to reasonable default. I think this is an acceptable compromise for now. However I think we should not advertise this too much and try to introduce our own "gadf" format instead. 


As this PR is already in a very good shape I will go ahead and merge it now. You could maybe add another little section with the serialisation in a follow up PR....details to be discussed next year...:-)

PR #3166, opened 2020-12-14
Unify npred background handling in dataset methods

adonath:
**Description**

This pull request unifies the `npred_background` handling in the `MapDataset` methods. The idea is that in methods such as `.to_image()` and `.to_spectrum()` we always rely on `background` and not `npred_background()`. I'm still not 100% sure whether this is the best default behaviour however it simplifies the model definition for the returned dataset a lot. Let's take it as a prototype and discuss some more...

### DIFF
--- a/gammapy/datasets/core.py
+++ b/gammapy/datasets/core.py
@@ -238,6 +238,9 @@ class Datasets(collections.abc.MutableSequence):
     def slice_by_energy(self, energy_min, energy_max):
         """Select and slice datasets in energy range
 
+        The method keeps the current dataset names. Datasets, that do not
+        contribute to the selected energy range are dismissed.
+
         Parameters
         ----------
         energy_min, energy_max : `~astropy.units.Quantity`
@@ -256,7 +259,7 @@ class Datasets(collections.abc.MutableSequence):
                 dataset_sliced = dataset.slice_by_energy(
                     energy_min=energy_min,
                     energy_max=energy_max,
-                    name=dataset.name + "-slice",
+                    name=dataset.name,
                 )
             except ValueError:
                 log.info(
--- a/gammapy/datasets/map.py
+++ b/gammapy/datasets/map.py
@@ -1415,7 +1415,7 @@ class MapDataset(Dataset):
             kwargs["exposure"] = self.exposure.pad(pad_width=pad_width, mode=mode)
 
         if self.background is not None:
-            kwargs["background"] = self.npred_background().pad(
+            kwargs["background"] = self.background.pad(
                 pad_width=pad_width, mode=mode
             )
 
--- a/gammapy/datasets/map.py
+++ b/gammapy/datasets/map.py
@@ -1530,7 +1530,7 @@ class MapDataset(Dataset):
         Returns
         -------
         dataset: `MapDataset` or `SpectrumDataset`
-            Resampled dataset .
+            Resampled dataset.
         """
         name = make_name(name)
         kwargs = {"gti": self.gti, "name": name, "meta_table": self.meta_table}
@@ -1557,7 +1557,7 @@ class MapDataset(Dataset):
             )
 
         if self.background is not None and self.stat_type == "cash":
-            kwargs["background"] = self.npred_background().resample_axis(
+            kwargs["background"] = self.background.resample_axis(
                 axis=energy_axis, weights=self.mask_safe
             )
 
--- a/gammapy/datasets/map.py
+++ b/gammapy/datasets/map.py
@@ -1229,7 +1229,7 @@ class MapDataset(Dataset):
             )
 
         if self.stat_type == "cash" and self.background is not None:
-            kwargs["background"] = self.npred_background().get_spectrum(
+            kwargs["background"] = self.background.get_spectrum(
                 on_region, func=np.sum, weights=self.mask_safe
             )
 
@@ -1301,7 +1301,7 @@ class MapDataset(Dataset):
             kwargs["exposure"] = self.exposure.cutout(**cutout_kwargs)
 
         if self.background is not None and self.stat_type == "cash":
-            kwargs["background"] = self.npred_background().cutout(**cutout_kwargs)
+            kwargs["background"] = self.background.cutout(**cutout_kwargs)
 
         if self.edisp is not None:
             kwargs["edisp"] = self.edisp.cutout(**cutout_kwargs)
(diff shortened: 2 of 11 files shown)


adonath:
I have started to think a bit about the ideal behaviour here. In general my impression from the past discussions, was that it would be desirable to keep the model definition in methods like `.to_image()`, `.to_spectrum_dataset()` etc. If we try to achieve this I think we should implement the following behaviour:

```python3
dataset = MapDataset()
dataset.models = Models.read()
npred = dataset.npred()

region = CircleSkyRegion()
npred_spec = npred.get_spectrum(region=region)

spec = dataset.to_spectrum_dataset(region=region)
spec_npred = spec.npred()

# of course not 100% the same, but as close as possible...
assert_allclose(npred_spec.data, spec_npred.data)
```
The same goes for the methods such as `.to_image()` or `.resample_energy_axis()`. So keep the model definition the way, that the npred calculation is consistent. Implementing this behaviour raises a few points:
- Should we create a new dataset name by default in methods, like `.to_image()` etc. or just use the existing one? I think the uniqueness of the name is less important here, as the modified datasets, will only rarely be used together with original dataset in a `Datasets` object. Keeping the name provides some "provenance" from which the dataset the other dataset was derived...
- Providing a new dataset name could be optional like `.to_image(name="image-dataset")`, and if it's provided the model is "re-assigned"
- For the `Models.reassign` I proposed to make a copy, however I realised this is problematic, as we still rely on the model id in `Datasets.models`. So consider the following example:
```python3

model = SkyModel(, name="my-source")
dataset_1.models = [model]
dataset_2.models = [model]

region = CircleSkyRegion()
spec_1 = dataset_1.to_spectrum_dataset(region=region)
spec_2 = dataset_2.to_spectrum_dataset(region=region)

print(Datasets([spec_1, spec)2]).models)
```
Would fail with a duplicated model name error. So I think the behaviour should be to pass on the references and "re-assign" models "in place".

- As implemented in this PR the background model should not be "applied", but the model definition is passed on.
- Deleting or selecting model components in `.to_spectrum_dataset()` or `.cutout()` is a weird behaviour I think.  So e.g. a call to `.cutout()`, dismisses model components outside. I think an optional, simple default behaviour could be just to freeze the corresponding model parameters. E.g. `.to_spectrum_dataset()` would freeze spatial parameters, `.cutout()` freezes not contained model components, `to_image()` freezes spectral model parameters, except for the norms and amplitudes etc.

I could go ahead and implement the proposed behaviour in this PR and the we can try it out a bit. What do you think @registerrier, @QRemy, @AtreyeeS, @luca-giunti?

adonath:
One more addendum: for consistency `Dataset.copy(name=)` should only copy the data, but keep the model reference. Again if a new name is provided, the model "re-assigned".

QRemy:
> So I think the behaviour should be to pass on the references and "re-assign" models "in place".

I would just add an option to make the copy optional in those methods
so we can  keep the behavior of "re-assign" with copy which can be useful in some cases
 and introduce another behavior without copy using append to assign the models in place more easily, like
`[m.datasets_names.append(new_name) for m in models if m.datasets_names is not None]`

> simple default behaviour could be just to freeze the corresponding model parameters

I agree with that, but I would still dismiss the components that do not contribute as we have already this information from the
evaluators. This make the models loading from catalogues much more convenient.

>  I think the uniqueness of the name is less important here, as the modified datasets, will only rarely be used together with the original dataset in a Datasets object. Keeping the name by default provides some "provenance" from which dataset the other dataset was derived...

yes and so by default there is no need for models copy and "re-assign", models remain the same than those of the parent dataset which is very convenient if we want to continue working with both parent and child datasets.

registerrier:
I am bit perplexed here. 
It seems to me that the variety of uses cases can be very large. While we might facilitate some of them, we also risk confusing users (and ourselves).

Should models be re-assigned or copied, I think is use case dependent, same for the background transform. 

If the approach is to leave the model handling to the user, I suppose we have to stick to that. Operations on a `Dataset` should not touch the models, and yield a `Dataset` devoid of any model. 

The more things evolve, the it seems to me that we shouldn't be doing `dataset.models = models` but `model.datasets = datasets`...

adonath:
Thanks @registerrier! As I said in my previous comment my impression from the last discussions was, that there was a preference for minimal model handling. But personally I still prefer to leave the model management to the users so I mostly agree with you here.

> The more things evolve, the more it seems to me that we shouldn't be doing dataset.models = models but model.datasets = datasets...

Note that by introducing the `datasets_names` attribute on the `SkyModel` and `FoVBackgroundModel` class we exactly went into this direction with bookkeeping. I think directly attaching the `Dataset` objects is not a good idea as it will prevent the distribution across multiple cores. It's not a new observation, but the fact that model objects are shared between multiple `Dataset` objects, fundamentally prevents to modify model objects with dataset operations in a clean way...there is no way around this, even more if the `Dataset` objects live in different processes. 

However there are still  a few little thing we can to improve the model handling:
- Not rely on `npred_background`
- Keep the dataset name in `.to_image()` etc. by default
This means users can e.g. do:
```python3
dataset_spec = dataset.to_spectrum_datasets(region=)
models = dataset.models.copy()
models.freeze("spatial")
dataset_spec.model = models
``` 
- Keeping the dataset name in `.cutout()` has mild consequences on joint analyses:  assigning a new name to the dataset becomes mandatory, because at the end all datasets are collected in a `Datasets` object. Alternatively we keep the current behaviour on dataset names and users have to explicitly keep the name of the old dataset like:
```python3
dataset_spec = dataset.to_spectrum_datasets(region=, name=dataset.name)
models = dataset.models.copy()
models.freeze("spatial")
dataset_spec.model = models
``` 
Which is not a bad option either, I think...

adonath:
One more compromise could be to keep the model in the output dataset, if the name of the dataset stays the same like:
```python3
spec = dataset.to_spectrum_dataset(region, name=dataset.name)
assert spec.models is not None

spec = dataset.to_spectrum_dataset(region, name="new-name")
assert spec.models is None
```

registerrier:
Note that the expected behaviors are different for different `Dataset` transformation methods.
- For `stack` one expect that the `npred_background` is used by default.
- For `to_image` I would assume this is the same, because the spectral norm is impossible to constrain once you have only a 2D datasets. But some use cases might require to keep it.

A common API could then be to add an argument `apply_background_model` to the transformation methods. It is relatively simple to document and covers all use cases.

adonath:
@registerrier I basically agree, however I think it  can be solved with simple documentation as well, instead of  introducing a new API. I think something along the lines of:

```python3
spec = dataset.to_spectrum(region=region)
spec.background = dataset.npred_background().get_spectrum(region=region)
```
Is very explicit and almost self-explaining...

adonath:
@registerrier Associated to the point on stacking: I guess the question is whether the `FoVBackgroundMaker` should directly apply the correction or just attach the `FoVBackgroundModel`. In the current behaviour it just attaches the model and the in `.stack()` it realties on `.npred_background()`.

adonath:
I'll go ahead and merge this now, I still think it leads to simpler code and a more consistent and better documentable behaviour in general.

PR #3167, opened 2020-12-15
Improve the plot_region method for RegionNDMap and RegionGeom

LauraOlivera:
This pull request fixes the default behavior of `RegionNDMap.plot_region()`, which used to fail if no axis was provided. The method has also been moved to the RegionGeom class, so a RegionGeom can be plotted too. 

If no axis is provided, the default behavior now is to plot on the minimal WCS geometry given by `RegionGeom.to_wcs_geom()`
Some examples:
```
from gammapy.maps import RegionNDMap
region_map = RegionNDMap.create("icrs;circle(83.63, 22.01, 0.5)")
region_map.plot_region()
```

and
```
from gammapy.maps import RegionGeom
geom = RegionGeom.create("icrs;circle(83.63, 22.01, 0.5)")
geom.plot_region()
```
now both result in the same image:

image

### DIFF
--- a/gammapy/maps/region.py
+++ b/gammapy/maps/region.py
@@ -4,6 +4,8 @@ from astropy import units as u
 from astropy.coordinates import SkyCoord
 from astropy.table import Table
 from astropy.wcs import WCS
+from astropy.visualization.wcsaxes import WCSAxes
+
 from astropy.wcs.utils import proj_plane_pixel_area, wcs_to_celestial_frame
 from regions import FITSRegionParser, fits_region_objects_to_table
 from gammapy.utils.regions import (
@@ -432,12 +434,16 @@ class RegionGeom(Geom):
         from matplotlib.collections import PatchCollection
 
         if ax is None:
-            wcs_geom = self.to_wcs_geom()
-            m = Map.from_geom(wcs_geom)
-            fig, ax, cbar = m.plot(add_cbar=False)
+            ax = plt.gca()
+            if not isinstance(ax, WCSAxes):
+                ax.remove()
+                wcs_geom = self.to_wcs_geom()
+                m = Map.from_geom(wcs_geom)
+                fig, ax, cbar = m.plot(add_cbar=False)
         regions = compound_region_to_list(self.region)
         artists = [region.to_pixel(wcs=ax.wcs).as_artist() for region in regions]
-
+        kwargs.setdefault("fc", "None")
+        kwargs.setdefault("ec", "b")
         patches = PatchCollection(artists, **kwargs)
         ax.add_collection(patches)
         return ax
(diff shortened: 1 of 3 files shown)


gammapy/maps/region.py
```diff
@@ -415,3 +415,29 @@ def union(self, other):
...
+            Axes to plot on. If no axes are given,
+            the region is shown using the minimal
+            equivalent WCS geometry.
+        **kwargs : dict
+            Keyword arguments forwarded to `~regions.PixelRegion.as_artist`
+        """
+        import matplotlib.pyplot as plt
+        from matplotlib.collections import PatchCollection
+
+        if ax is None:
```
adonath, line 434: We could still keep the previous behaviour, with `ax = plt.gca()`, which works if the "current" axis is a `WCSAxes` and just catch the case, where the current axis is not a was one...

gammapy/maps/region.py
```diff
@@ -415,3 +415,29 @@ def union(self, other):
...
+            wcs_geom = self.to_wcs_geom()
+            m = Map.from_geom(wcs_geom)
+            fig, ax, cbar = m.plot(add_cbar=False)
+        regions = compound_region_to_list(self.region)
+        artists = [region.to_pixel(wcs=ax.wcs).as_artist() for region in regions]
+
```
adonath, line 440: Maybe set `kwargs.setdefault("fc", "None")` here, so that the region is not filled? Might be the better default, if users plot on sky images...

gammapy/maps/tests/test_region.py
```diff
@@ -288,4 +289,13 @@ def test_get_wcs_coord(region):
     region_coords = geom.get_wcs_coord()
     coord = SkyCoord("100d", "30d")
     assert geom.contains(region_coords.skycoord[0])
-    assert geom.contains(coord) is not True
\ No newline at end of file
+    assert geom.contains(coord) is not True
```
adonath, line 292: I know it's unrelated but this should rather read `assert not gom.contains(coord)`

adonath (changes requested):
Thanks a lot @LauraOlivera, I have left a few minor inline comments, otherwise looks good to me!

gammapy/maps/region.py
LauraOlivera: you're right, I changed it, thanks!

gammapy/maps/region.py
LauraOlivera: I also added a default edge color because else the region was not visible when plotted alone

LauraOlivera:
The output of the snippets above now looks like this:

image

adonath (approved):
Thanks @LauraOlivera, Travis fails are unrelated...

PR #3168, opened 2020-12-16
Remove evaluator update from models setter

QRemy:
Remove evaluator update from models setter for speed up.
Setting many models or reading large datasets will be considerably faster.
Then evaluators update will be done only  once npred is computed.

### DIFF
--- a/gammapy/datasets/map.py
+++ b/gammapy/datasets/map.py
@@ -297,8 +297,6 @@ class MapDataset(Dataset):
                         gti=self.gti,
                         use_cache=USE_NPRED_CACHE,
                     )
-                    # TODO: do we need the update here?
-                    evaluator.update(self.exposure, self.psf, self.edisp, self._geom)
                     self._evaluators[model.name] = evaluator
 
         self._models = models
--- a/gammapy/datasets/simulate.py
+++ b/gammapy/datasets/simulate.py
@@ -64,8 +64,14 @@ class MapDatasetEventSampler:
         events : `~gammapy.data.EventList`
             Event list
         """
+
         events_all = []
         for idx, evaluator in enumerate(dataset.evaluators.values()):
+            if evaluator.needs_update:
+                evaluator.update(
+                    dataset.exposure, dataset.psf, dataset.edisp, dataset._geom
+                )
+
             flux = evaluator.compute_flux()
             npred = evaluator.apply_exposure(flux)
 


adonath:
Thanks @QRemy! Right now I don't see a clear advantage in the sense, that the time needed for the setup is still the same, it just moves it from the model setting to the fitting. I guess the underlying question is rather whether we can speed up the model setup. I have done some profiling and it seems we spend all the time in computing the `PSFKernel`:

Screenshot 2020-12-16 at 15 35 21

This is ~0.5s per model component and dataset using ~30 true energy bins and a spatial bin size of 0.02 deg. I can easily see how this accumulates for many model components. Looking at the code we are currently using oversampling (see https://github.com/gammapy/gammapy/blob/master/gammapy/irf/psf_map.py#L138). If we rely on users providing a reasonable pixels size matching the PSF, we can probably remove the oversampling and speed up the model setup by a factor of ~4 ^ 2. 

@registerrier I think you implemented the oversampling. Do you think its needed?

A few other ideas to optimise would be to reduce the kernel size and exploit the radial symmetry of the kernel, so it would be sufficient to interpolate one "quarter" of the kernel and the copy mirror the rest.

adonath:
@QRemy Apart from that I don't feel strongly how to distribute the setup time. I guess a model is set only if the npred computation is done at some point as well...

adonath:
The current size of the kernel is computed as either twice the size of the rad max of the `PSFMap` or the size of the exposure cutout, depending on which one is smaller...

QRemy:
I made this change because when I read the the global CTA-PGS datasest it reduces reading time from ~1000s to 50s.  Then if ones want to read this global dataset make a cutout and continue their analysis only on a dedicated region this modification is usefull.

adonath:
@QRemy In v0.18.2 one can already read the datasets and models independently, like:

```python3
# should be rather fast as no model is defined 
datasets = Datasets.read("datasets.yaml")

# reading models should be fast as well
models = Models.read("models.yaml")

# now for the setup, which takes some time
datasets.models = models
```

QRemy:
Well the problem for me is that, `datasets.models = models` take longer than `datasets = Datasets.read("datasets.yaml")`

adonath:
@QRemy Yes, but why would set the model just for doing the cutouts?

QRemy:
Yes for that case the models selection can be done first and set after.
 I think there is still a point a making the setter faster and delay the update of the evaluators so for users working interactively they have to wait ones instead of twice, so only the evaluation will take time which is already the case anyway.

adonath:
@QRemy OK, the argument for interactive analysis I get...

PR #3169, opened 2020-12-16
Add .select_region() and .select_mask() methods to Models

QRemy:
This PR add models management methods to Models  :
- .select_region(regions) :  Select skymodels with center position contained within a given region
- .contribute(mask, psf): check if a model contribute within a mask map

### DIFF
--- a/gammapy/modeling/models/core.py
+++ b/gammapy/modeling/models/core.py
@@ -641,7 +641,7 @@ class DatasetModels(collections.abc.Sequence):
 
         return np.array(selection, dtype=bool)
 
-    def select_mask(self, mask, margin=None, include_evaluation_radius=True):
+    def select_mask(self, mask, margin=None, use_evaluation_region=True):
         """Check if skymodels contribute within a mask map.
     
         Parameters
@@ -653,7 +653,7 @@ class DatasetModels(collections.abc.Sequence):
             Add a margin in degree to the source evaluation radius.
             The default is None. Used to take into account PSF width.
 
-        include_evaluation_radius : bool
+        use_evaluation_region : bool
             Account for the extension of the model or not. The default is True.   
 
         Returns
@@ -668,7 +668,7 @@ class DatasetModels(collections.abc.Sequence):
 
         models = self.select(tag="SkyModel")
         contribute = np.array(
-            [m.contributes(mask, margin, include_evaluation_radius) for m in models]
+            [m.contributes(mask, margin, use_evaluation_region) for m in models]
         )
         return models[contribute]
 
--- a/gammapy/modeling/models/core.py
+++ b/gammapy/modeling/models/core.py
@@ -245,8 +245,8 @@ class Model:
 
         return model
 
-    def contribute(self, mask, psf=None):
-        """Check if a model contribute within a mask map
+    def contribute(self, mask, margin=None, inside=True):
+        """Check if a model contributes within a mask map
            Note that BackgroundModel are assumed to always contribute. 
 
         Parameters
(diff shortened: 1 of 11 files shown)


gammapy/modeling/models/core.py
```diff
@@ -612,6 +669,61 @@ def selection_mask(
 
         return np.array(selection, dtype=bool)
 
+    def contribute(self, mask, psf=None):
```
adonath, line 672: I would argue taking a `PSFKernel` object here is non-ideal, as it's rather complex to create. A simple number like `margin="0.6 deg"` is in general sufficient. So having a signature like:

```python3
models.select_mask(mask, margin="0.6 deg", inside=True)
```
Is easier to use...the connection to the `PSFKernel` could be done using an example in the doctoring instead.

gammapy/modeling/models/core.py
```diff
@@ -237,6 +242,58 @@ def reassign(self, datasets_names, new_datasets_names):
                         model.datasets_names[k] = new_dataset_name
         return model
 
+    def contribute(self, mask, psf=None):
```
adonath, line 245: Again I would change the signature to `contributes(mask, margin="0.6 deg", inside=True)`. It's more flexible and easier to use. Again the connection to the `PSFKernel` could be done just with a little docs example. Internally to apply the margin I would rely on the `WcsNDMap.erode()`.

gammapy/modeling/models/core.py
```diff
@@ -612,6 +669,61 @@ def selection_mask(
...
+            True  where models contribute within the mask geom
+            but its center is located ouside the region where mask==True
+        """
+
+        nmod = len(self)
+        inside = np.zeros(nmod, dtype=bool)
+        outside = np.zeros(nmod, dtype=bool)
+        for k, m in enumerate(self):
+            inside[k], outside[k] = m.contribute(mask, psf)
+        return inside, outside
```
adonath, line 698: Here I would just return one of them and add the `inside=True/False` parameter. I can see how it can be advantageous to return both, however for API consistency with `.region_mask()` and `.boundary_mask()` I would prefer the `inside=` option.

adonath (changes requested):
Thanks @QRemy! Sorry for the delay here, I have left a few comment to address.

gammapy/modeling/models/core.py
QRemy: For now I would prefer to not apply the margin using WcsNDMap.erode() so it match the behaviour of the MapEvaluator. Then I will check if the MapEvaluator can be adapted to use .contribute() and make an other PR if so.

gammapy/modeling/models/core.py
```diff
@@ -612,6 +669,61 @@ def selection_mask(
...
+
+        Returns
+        -------
+        models : `DatasetModels`
+            Selected models 
+        """
+
+        models = self.select(tag="SkyModel")
+
+        if not isinstance(regions, list):
```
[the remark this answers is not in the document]
QRemy: I modified `WcsGeom.region_mask()`  to accept a single region

gammapy/modeling/models/core.py
```diff
@@ -612,6 +669,61 @@ def selection_mask(
...
+            regions = [regions]
+
```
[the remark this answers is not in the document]
QRemy: The init of `RegionGeom` fails in this case because:
 "AttributeError: 'CompoundSkyRegion' object has no attribute 'center'"

adonath:
Sorry @QRemy, for the delay on this, it took me a while to find the time to think about the code structure again. My goal is always to break the functionality down into as many useful sub-functions as possible, that we can re-use in other places too. So here comes my final proposal:
- Let's introduce `SpatialModel.evaluation_region`, this can probably just rely on `.to_region()` or the other way around. This allows to handle non-circular regions as well later. The difference is just the size of the region...maybe adding `Models.evaluation_regions` is useful as well.
- In addition we introduce `Map.contains_region(regions=regions)`, which checks whether a Region is contained in a mask. Ideally this can handle `PixRegions` as well. The main point here, is that the method can be reused in the `ReflectedRegionsFinder`, where exactly the same functionality is needed (I can do the refactoring later...). A list of regions should be supported as well and return a boolean array.
- Using these methods the `SkyModel.contributes()` becomes very simple, possibly it could be replaced by `mask.dilate(margin=).contains_region(model.evaluation_region)` completely in the `MapEvaluator` as well
- The `Models.select_region()` is fine as is
- `Models.select_mask()`, also relies on `model.evaluation_region` and `.contains_region()`

I think this way we have consistency between the selection in `MapEvaluator` and `Models.select_mask()`, as well as additional functionality, that can bis used in the other places as well. Does this sound reasonable to you?

QRemy:
Sounds good, I just have some doubts about `mask.dilate(margin=)` it may have extend the geom size by the margin but that will cause problems elsewhere, and Map.contains_region() should take into account the values of the mask but that make not much sense for non-boolean map.

adonath:
Thanks @QRemy, those are good points:
- I think this case can be handled with a call to `Map.pad()` beforehand or having different modes in `Map.binary_dilate()`
- I agree the `.contains_region()` is only well defined for a boolean mask, we could name it `.mask_contians_region()` for clarity and raise an error if the map doesn't contain boolean values, just as we do for `.binary_dilate()`, at some point I added a flag called `Map.is_mask`...

adonath (approved):
Thanks a  lot @QRemy some of the methods might need a unit test, but this can be done later...

PR #3180, opened 2021-01-11
Remove code duplication from models.spectral

fabiopintore:
There is a code duplication in `models.spectral`, related to the functions that treat the error propagation. The PR aims at removing them and merge them into one new function called `_propagate_error()`.


adonath (approved):
Thanks @fabiopintore, no further comments from my side...

PR #3182, opened 2021-01-12
Introduce FluxEstimate class

registerrier:
**Description**

This pull request adds the `FluxEstimate` class that will serve as a basis for flux estimators results.  Following the proposal in PIG 22, it is designed to take a reference `SpectralModel` and a `data` element containing `norm` based quantities.  `data` can be a `Table` or a `dict` of `Map` to support flux points or flux maps.

Properties are defined to allow direct access to flux quantities in the various SED types defined in the current version of gadf (namely dnde, e2dnde, flux and eflux).  

Tests have been added to check both `Table` and `Map` inputs.

**Dear reviewer**

### DIFF
--- a/gammapy/estimators/flux_estimate.py
+++ b/gammapy/estimators/flux_estimate.py
@@ -1,7 +1,5 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
-import astropy.units as u
-from gammapy.maps import Geom, MapAxis
-from gammapy.modeling.models import PowerLawSpectralModel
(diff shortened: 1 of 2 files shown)


adonath:
Thanks a lot @registerrier, so far I don't have any comments. I think the best way to proceed here is merge and the continue the work on #3075?

registerrier:
OK. Yes once this is in we can adapt the `FluxMap` PR.

registerrier:
Merging now and moving to #3075

PR #1288, opened 2018-02-07
Generalise instrument response classes

kbruegge:
This replaces all the code in the irf subpackage except for the PSF related stuff. 
Use like this.
```
from gammapy.irf import InstrumentResponse
irf = InstrumentResponse.from_fits(path, extension='EFFECTIVE AREA')
point = {'THETA': 3.5*u.deg, 'ENERG': 1*u.TeV}
interpolated_result = irf.evaluate(point)
```

PR #3185, opened 2021-01-13
Introduce IRF base class and remove code duplication

adonath:
**Description**

This pull request introduce a `IRF` base class, that serves as a base for all DL3 type IRFs in `gammapy.irf`. The motivation is as following:
- Remove code duplication
- Create a more uniform API with our `Map` class  

It comes with the following changes:
- The IRF class takes `data`, `unit`, `meta` and `axes` (instead of `geom`) on init
- It defines `IRF.axes`, `IRF.data` and `IRF.quantity`
- It works in ND for interpolation and evaluation
- The subclasses declare the "required_axes" for their specific type
- All `read()`, `write()`, `from_table()`, `.to_table()` etc. was moved to the base class
- The GADF DL3 specification is declared independently in `io.py`
-  I decided to remove the direct access to the axes attributes via e.g. `.energy_axis`, `.energy_true_axis` etc. Axes can be just accessed via `.axes["energy"]`
- All functionality from the `NDDataArray` was moved to the `IRF` class for now and the `NDDataArray` was removed. 

TODO:
- [x] Add unit test for `Background2D` data reduction. It does not seem to exist

**Dear reviewer**
This PR includes way to many changes for a detailed review, please give feedback on the overall concept. Smaller improvements will be implemented in follow up PRs.


adonath:
This is basically a continuation of the work proposed in #1288.

gammapy/irf/core.py
```diff
@@ -1,17 +1,284 @@
...
+
+        for key, value in kwargs.items():
+            coord = kwargs.get(key, value)
+            if coord is not None:
+                coords_default[key] = u.Quantity(coord, copy=False)
+
+        return self._interpolate(coords_default.values(), method=method)
+
+    # TODO: define a proper integration method
+    def integrate_energy(self, method="linear", **kwargs):
```
registerrier, line 146: Can integrate methods be general enough that they belong to the more abstract base class?
Here the integration is on reco energy, hence the method is only applicable to background. (And background is the only IRF defined as a differential quantity over energy).
I would suggest to put an energy integration method only on the `Background` base class.

gammapy/irf/core.py
```diff
@@ -1,17 +1,284 @@
...
+        array : `~astropy.units.Quantity`
+            Returns 2D array with axes offset
+        """
+        axis = self.axes.index("energy")
+        data = self.evaluate(**kwargs, method=method)
+        energy = kwargs["energy"]
+        return trapz_loglog(data, energy, axis=axis)
+
+    # TODO: define a proper integration method
+    @lazyproperty
```
registerrier, line 169: Same comment here. This is applicable only to PSF and would rather fit on a general PSF base class.

gammapy/irf/core.py
```diff
@@ -1,17 +1,284 @@
...
+
+        Returns
+        -------
+        irf : `IRF`
+            IRF class
+        """
+        with fits.open(str(make_path(filename)), memmap=False) as hdulist:
+            return cls.from_hdulist(hdulist, hdu=hdu)
+
+    @classmethod
```
registerrier, line 225: I think to prepare the likely multiplicity of formats we need to provide the `format` as an argument to read/write and `from_table` methods.

gammapy/irf/core.py
```diff
@@ -1,17 +1,284 @@
...
+        data = table[column_name].quantity[0].transpose()
+
+        return cls(
+            axes=axes,
+            data=data.value,
+            meta=table.meta,
+            unit=data.unit
+        )
+
+    def to_table(self):
```
registerrier, line 250: same comment about formats

gammapy/irf/core.py
```diff
@@ -1,17 +1,284 @@
...
+        """Convert to `~astropy.table.Table`."""
+        table = self.axes.to_table(format="gadf-dl3")
```
registerrier, line 252: I think that in the long run having format specification inside `MapAxis` will lead to complexity in code duplication. 
Ideally we should have an `IRFReader` that knows about formats.

gammapy/irf/io.py
```diff
@@ -19,6 +13,45 @@
 }
 
 
+# The key is the class tag. The hdu name is not strictly declared in the GADF
+# format
+# TODO: extend the info here
+IRF_DL3_HDU_SPECIFICATION = {
+    "bkg_3d": {
+        "hdu": "BACKGROUND",
```
registerrier, line 21: Here it is `extname`  rather than `hdu` no?

gammapy/irf/io.py
```diff
@@ -19,6 +13,45 @@
 }
 
 
+# The key is the class tag. The hdu name is not strictly declared in the GADF
```
registerrier, line 16: Technically the EXTNAME is provided by the HDU index no?
If there is no specification it might be unsafe to use one.

registerrier (commented):
Thanks @adonath . That's an important step.

I only had a brief overview for now. I have left general comments in line.
I think we should avoid having methods specific to certain sub-category of IRF on the base `IRF` class.
Also the current scheme opens the possibility to have  in-memory objects more decoupled from the format, but this would need an explicit format argument in I/O methods as well as a format dependent HDU keywords specification

gammapy/irf/core.py
adonath: Good point. I'll check whether I can make this maybe a more general integration method. Otherwise I'll move it to `Background` again...

gammapy/irf/core.py
adonath: Right now there is only the `gadf-dl3` format and we will not get any other format soon, but I agree we could prepare the API for this...

gammapy/irf/io.py
adonath: Ah, indeed. I forgot about this. So this question only applies to writing IRFs and to provide a default for reading. I guess there we can use any internal Gammapy convention we want. So I'll just keep what we have and remove the comment...

gammapy/irf/core.py
adonath: I renamed this method to `integrate_log_log()` and added an `axis_name` argument. Maybe it's useful in another context as well.

gammapy/irf/io.py
adonath: Done

gammapy/irf/core.py
adonath: I agree this will be the long term solution...

adonath:
Thanks @registerrier, I addressed your comments. I'll proceed with the merge now. I'm happy to address more comments in follow up PRs.

Issue #3118, opened 2020-11-06
Participate in iminuit 2.0 beta?


HDembinski:
Dear gammapy team,

I am about to finish a major rewrite of iminuit, version 2.0, that replaces Cython as the tool to wrap C++ Minuit2 with pybind11, which is going to solve several issues that the legacy code had. All the good things that this will bring are listed on top of this PR:
scikit-hep/iminuit#502

Switching to the new version of iminuit should be completely transparent to you, since the new version passes the comprehensive suite of unit tests of iminuit-v1.x. However, I would like to use this opportunity to finally remove interface that has been successively marked as deprecated in versions 1.3 to 1.5.

Therefore my two question to you:

    Did you take note of the deprecation warnings in iminuit and did you keep up with the interface changes?
    Are you interested in trying out a Beta release of v2.0 to work out any possible bugs in the new version before the release?

Best regards,
Hans, iminuit maintainer

adonath:
Thanks @HDembinski,  in general I can't remember that we had some deprecation warnings from `iminuit`. So it seems we only rely on features and the API that you continue to support. In general we are happy to test the beta version and give feedback if necessary...

HDembinski:
Thank you! You can install install the beta from test.pypi:
`pip install -i https://test.pypi.org/simple/ iminuit`

HDembinski:
iminuit v2.0 is released, the breaking changes in the interface are documented in the [changelog](https://iminuit.readthedocs.io/en/stable/changelog.html) which can serve as an upgrade guide.

adonath:
Thanks a lot @HDembinski ! I was able to implement the support for iminuit v2.0 in #3409. However there was one remaining change I could not resolve. Previously `.mncontour()` also returned additional status info, now it only returns the contour. Is the status info still available somewhere else? Then I could adapt the implementation [here]( https://github.com/gammapy/gammapy/pull/3409/files#diff-865154da598a214a473468b661ce8d47a68cb34d8ddd7eeb0792fab157c44cfaR163).

HDembinski:
The old code implemented the MnContour scan "manually". The code now uses the class MnContours from C++ MINUIT. There is no status code anymore, because MnContours does not return a status flag.

adonath:
Thanks for the clarification @HDembinski!

adonath:
The PR updating Gammapy to iminuit>2.0 has been merged (#3409). Thanks @HDembinski for the help. Will close the issue now..

PR #3188, opened 2021-01-14
Fix deprecation warnings

adonath:
**Description**

This pull request fixes deprecation warnings and raise them as exceptions to fix them early in the CI. Supersedes #2315.


adonath:
I looked into this and we get deprecation warnings from `iminuit`, which I can not suppress. So we have to adapt to the new `iminuit` API. Related #3118.

PR #3191, opened 2021-01-15
Remove code duplication between SpectrumDatasetMaker and MapDatasetMaker

adonath:
**Description**

This pull request .remove the code duplication between `SpectrumDatasetMaker` and `MapDatasetMaker`, by making the `SpectrumDatasetMaker` inherit from the `MapDatasetMaker`. This has one little change in analysis results, if the background template is used: in the `SpectrumDataset` the background was extracted from the give offset forced at `fov_lon=0`. In the `MapDatasetMaker` the correct position is used, which is relevant for non-radially symmetric background models, which we have for HESS. That's why some test values changed. Otherwise it looks good.

### DIFF
--- a/gammapy/makers/tests/test_spectrum.py
+++ b/gammapy/makers/tests/test_spectrum.py
@@ -83,8 +83,8 @@ def test_spectrum_dataset_maker_hess_dl3(spectrum_dataset_crab, observations_hes
     assert_allclose(datasets[0].exposure.meta["livetime"].value, 1581.736758)
     assert_allclose(datasets[1].exposure.meta["livetime"].value, 1572.686724)
 
-    assert_allclose(datasets[0].npred_background().data.sum(), 7.74732, rtol=1e-5)
-    assert_allclose(datasets[1].npred_background().data.sum(), 6.118879, rtol=1e-5)
+    assert_allclose(datasets[0].npred_background().data.sum(), 7.747881, rtol=1e-5)
+    assert_allclose(datasets[1].npred_background().data.sum(), 5.731624, rtol=1e-5)
 
 
 @requires_data()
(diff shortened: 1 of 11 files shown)


adonath:
Minor: shall we name the `make_` functions in `makers/utils.py` uniformly?
-   "make_map_background_irf" -> "make_background_map"
-    "make_edisp_map"
-    "make_edisp_kernel_map"
-    "make_psf_map"
-    "make_map_exposure_true_energy" -> "make_exposure_map"
-    "make_theta_squared_table"

gammapy/makers/tests/test_spectrum.py
```diff
@@ -83,8 +83,8 @@ def test_spectrum_dataset_maker_hess_dl3(spectrum_dataset_crab, observations_hes
     assert_allclose(datasets[0].exposure.meta["livetime"].value, 1581.736758)
     assert_allclose(datasets[1].exposure.meta["livetime"].value, 1572.686724)
 
-    assert_allclose(datasets[0].npred_background().data.sum(), 7.74732, rtol=1e-5)
-    assert_allclose(datasets[1].npred_background().data.sum(), 6.118879, rtol=1e-5)
+    assert_allclose(datasets[0].npred_background().data.sum(), 7.747881, rtol=1e-5)
+    assert_allclose(datasets[1].npred_background().data.sum(), 5.731624, rtol=1e-5)
```
registerrier, line 87: Here the change on the background counts seems quite large (~8%). Any idea why?

gammapy/maps/wcsmap.py
```diff
@@ -203,6 +203,8 @@ def to_hdulist(self, hdu=None, hdu_bands=None, sparse=False, format="gadf"):
 
         hdu_out = self.to_hdu(hdu=hdu, hdu_bands=hdu_bands, sparse=sparse)
 
+        # TODO: make this serialisable
+        self.meta.pop("livetime", None)
```
registerrier, line 207: Why is this needed? If this is required because of the defined format, it might be safer to copy the meta `dict` and remove the livetime entry in the copy.

registerrier (approved):
Thanks @adonath . This looks good. On background value in a test seems to have changed significantly. Any idea why?
Is it possible now to directly include in the maker the treatment of extended regions?

gammapy/makers/tests/test_spectrum.py
adonath: Yes, in the previous implementation the background evaluation was forced to fov-lat=0, fov-lon = offset, assuming the background was radially symmetric. Now the correct position in fov coordinates is used.

gammapy/maps/wcsmap.py
adonath: Yes, indeed modifying the meta dict on write is a bad idea, I‘ll fix this...

adonath: I added a custom json serialiser, that handles `Quantity` objects. This is maybe useful in another context as well...

PR #3197, opened 2021-01-21
Introduce PSF and ParametricPSF base classes

adonath:
**Description**

This pull request introduce a `PSF` and `ParametricPSF` base class, with the goal to reduce code duplication. The most important changes are:

- Move every generic methods such as `.containment()` and `.containment_radius()` to a common `PSF` base class
- Move and rename `.plot_containment_radius()` and `.plot_containment_radius_vs_energy()` to the base class as well
- Implement `PSFKing.evaluate_containment()`, so that containment computation works for the King PSF as well
- Remove further code duplication, with `.to_psf3d()` and `.to_energy_dependent_table_psf()`
- All methods such as `.containment()`, `.containment_radius()` and `.evaluate()` support numpy broadcasting now, no more Python loops for evaluation of those.
- Use linear interpolation of the parameters in the `KingPSF`

The `ParametericPSF` is a bit special:
- It contains multiple arrays for the different parameters, which I decided to store in a `np.recarray()` for now
- Correspondingly the `.unit` and `.quantity` attributes return a dict of the parameter value units and quantities

All in all this removes ~400 lines of duplicated code, while making the API more consistent and (in principle) generalises the PSF classes to ND.


adonath:
I would like to continue with the clean up, so I'm merging this now. @registerrier Please try it out in master and let me know if you have further comments, I'll continue to work on `gammapy.irf` next week anyway...

PR #3198, opened 2021-01-22
Fix bug in HpxMap.plot_grid()

LauraOlivera:
This pull request fixes the `plot_grid()` method when used with Healpix maps. It also adds a test for that use case.


adonath:
Thanks @LauraOlivera, the CI fail is unrelated...

adonath (approved):

PR #3200, opened 2021-01-23
Small modifs in documentation structure

Bultako:
This PR adds to all the pages of the dev version of docs the red alert box stating we are not reading the stable version of the docs. It also modifies the left menu bar, adding a link to the list of releases, as well as in the introduction description text of Gammapy in the homepage of the docs.

You may see how it look likes in the [published dev version of the Gammapy docs](https://docs.gammapy.org/dev/).


adonath (approved):
Thanks a lot @Bultako, this looks good!

PR #3202, opened 2021-01-25
Add an option to the SpectrumDatasetMaker to average IRFs over a Region

LauraOlivera:
This pull request adds a flag `average_over_region` to the `SpectrumDatasetMaker` to, instead of just taking the value at the region center, average (or add) the IRF values along the entire region.

The current status is very preliminary, as I've only modified the exposure calculation as a first step. Following steps are:

- Sum background over region :heavy_check_mark: 
- ~Average PSF~ 
- Average energy dispersion :heavy_check_mark: 
- Add the tests :heavy_check_mark:  

I will add these features in the following days.

A further step, ~cmaybe in a separate (or parallel) PR would be~  to add a method `RegionGeom.get_wcs_weight()` to compute spatial weights to achieve a better compromise between speed and accuracy in the integration steps. :heavy_check_mark:

### DIFF
--- a/gammapy/makers/utils.py
+++ b/gammapy/makers/utils.py
@@ -56,7 +56,6 @@ def make_map_exposure_true_energy(pointing, livetime, aeff, geom, average_over_r
 
     energy_true = wcs_geom.axes["energy_true"].center
 
-
     exposure = aeff.evaluate(
         offset=offset, energy_true=energy_true[:, np.newaxis, np.newaxis]
     )
--- a/gammapy/makers/utils.py
+++ b/gammapy/makers/utils.py
@@ -46,7 +46,8 @@ def make_map_exposure_true_energy(pointing, livetime, aeff, geom, use_region_cen
         Exposure map
     """
     if not use_region_center:
-        wcs_geom = geom.to_wcs_geom().upsample(2)
+        wcs_geom = geom.to_wcs_geom()
+        weights = geom.get_wcs_weights().data
         mask = geom.contains(wcs_geom.get_coord())
     else:
         wcs_geom = geom
@@ -66,7 +67,7 @@ def make_map_exposure_true_energy(pointing, livetime, aeff, geom, use_region_cen
     if not use_region_center:
         # TO DO: add weights
         exposure[~mask] = np.nan
-        data = np.nanmean(exposure.value, axis=(1,2))
+        data = np.nanmean(weights*exposure.value, axis=(1,2))
     else:
         data = exposure.value
 
@@ -204,7 +205,7 @@ def make_map_background_irf(pointing, ontime, bkg, geom, oversampling=None, use_
     return bkg_map
 
 
-def make_psf_map(psf, pointing, geom, exposure_map=None, use_region_center=True):
+def make_psf_map(psf, pointing, geom, exposure_map=None):
     """Make a psf map for a single observation
 
     Expected axes : rad and true energy in this specific order
--- a/gammapy/makers/utils.py
+++ b/gammapy/makers/utils.py
@@ -66,6 +66,7 @@ def make_map_exposure_true_energy(pointing, livetime, aeff, geom, average_over_r
         data = np.mean(exposure.value, axis=(1,2))
     else:
         data = exposure.value
+        
     return Map.from_geom(
         geom=geom, data=data, unit=exposure.unit, meta=meta
     )
--- a/gammapy/makers/utils.py
+++ b/gammapy/makers/utils.py
@@ -64,7 +64,8 @@ def make_map_exposure_true_energy(pointing, livetime, aeff, geom, average_over_r
 
     if average_over_region:
         # TO DO: add weights
-        data = np.mean(exposure.value, axis=(1,2))
+        exposure[~mask] = np.nan
+        data = np.nanmean(exposure.value, axis=(1,2))
     else:
         data = exposure.value
 
--- a/gammapy/makers/utils.py
+++ b/gammapy/makers/utils.py
@@ -248,7 +248,7 @@ def make_psf_map(psf, pointing, geom, exposure_map=None):
     return PSFMap(psfmap, exposure_map)
 
 
-def make_edisp_map(edisp, pointing, geom, exposure_map=None):
+def make_edisp_map(edisp, pointing, geom, exposure_map=None, average_over_region=False):
     """Make a edisp map for a single observation
 
     Expected axes : migra and true energy in this specific order
@@ -266,7 +266,10 @@ def make_edisp_map(edisp, pointing, geom, exposure_map=None):
     exposure_map : `~gammapy.maps.Map`, optional
         the associated exposure map.
         default is None
-
+    average_over_region: Bool
+        If geom is a RegionGeom, whether to just
+        consider the values at the region center
+        or the average over the whole region
     Returns
     -------
     edispmap : `~gammapy.irf.EDispMap`
@@ -276,17 +279,29 @@ def make_edisp_map(edisp, pointing, geom, exposure_map=None):
     migra = geom.axes["migra"].center
 
     # Compute separations with pointing position
-    offset = geom.separation(pointing)
+    if average_over_region:
+        wcs_geom = geom.to_wcs_geom().upsample(2)
+        mask = geom.contains(wcs_geom.get_coord())
+    else:
+        wcs_geom = geom
+
+    offset = wcs_geom.separation(pointing)
 
     # Compute EDisp values
     edisp_values = edisp.evaluate(
         offset=offset,
         energy_true=energy_true[:, np.newaxis, np.newaxis, np.newaxis],
         migra=migra[:, np.newaxis, np.newaxis],
-    )
+    ).to_value("")
+
+    if average_over_region:
+        # TO DO: add weights
+        edisp_values[~mask] = np.nan
+        data = np.nanmean(edisp_values, axis=(2,3))
+    else:
+        data = edisp_values
 
     # Create Map and fill relevant entries
-    data = edisp_values.to_value("")
     edispmap = Map.from_geom(geom, data=data, unit="")
     return EDispMap(edispmap, exposure_map)
 
--- a/gammapy/makers/utils.py
+++ b/gammapy/makers/utils.py
@@ -48,12 +48,10 @@ def make_map_exposure_true_energy(pointing, livetime, aeff, geom, use_region_cen
     if not use_region_center:
         wcs_geom = geom.to_wcs_geom()
         weights = geom.get_wcs_weights().data
-        mask = geom.contains(wcs_geom.get_coord())
     else:
         wcs_geom = geom
-        mask = 1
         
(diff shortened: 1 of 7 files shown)


gammapy/makers/utils.py
```diff
@@ -110,6 +130,10 @@ def make_map_background_irf(pointing, ontime, bkg, geom, oversampling=None):
         Reference geometry
     oversampling: int
         Oversampling factor in energy, used for the background model evaluation.
+    average_over_region: Bool
+        If geom is a RegionGeom, whether to just
+        consider the values at the region center
+        or the average over the whole region
```
registerrier, line 136: Here it is not an average but a sum. Maybe the argument could reflect that.

gammapy/makers/utils.py
```diff
@@ -35,23 +35,43 @@ def make_map_exposure_true_energy(pointing, livetime, aeff, geom):
...
 
     Returns
     -------
     map : `~gammapy.maps.WcsNDMap`
         Exposure map
     """
-    offset = geom.separation(pointing)
-    energy_true = geom.axes["energy_true"].center
+    if average_over_region:
+        wcs_geom = geom.to_wcs_geom().upsample(2)
```
registerrier, line 49: Why the factor of 2 here? Is there a criterion to define it a priori?

registerrier (commented):
Thanks @LauraOlivera ! This is good.

My main concern is to know whether the right place to add the relevant piece of code is `make_map_exposure`, `make_background_exposure`. Adding the option to average on the functions is a bit confusing. Is it possible to convert to `wcs_geom` inside the `SpectrumDatasetMaker`? Or is it an inefficient solution?

gammapy/makers/utils.py
adonath: I agree here I think the oversampling by a factor of 2 is an arbitrary choice, so one could either make it an argument (possibly `region_oversampling`, like `background_oversampling`), or what I already proposed to @LauraOlivera, was to use a set of weights to represent the region, like:

image

This set of weights can be pre-computed an cached on the `RegionGeom` and should help to improve the the accuracy of the integration a lot...

adonath:
@registerrier I think the choice of having this inside the `make_` utils functions is to not pass the tangential WCS geom to the `make_` function but work with a flattened list of coordinates inside for efficiency (I think currently this is not the case...). I think for a circular region, that is well contained inside the tangential WCS it doesn't matter but for weird region shapes, where most of the projected area is empty it probably does...

gammapy/makers/utils.py
```diff
@@ -132,11 +156,17 @@ def make_map_background_irf(pointing, ontime, bkg, geom, oversampling=None):
     coords = {
         "energy": geom.axes["energy"].edges.reshape((-1, 1, 1))
     }
+    if average_over_region:
+        wcs_geom = geom.to_wcs_geom().upsample(2)
```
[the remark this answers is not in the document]
LauraOlivera: totally arbitrary, I had it as a placeholder of sorts until I implement the weights that Axel described below - sorry, should have specified that in a comment.

gammapy/makers/utils.py
LauraOlivera: hm good point. Do you mean changing the name to `sum_over_region` for this case or just changing the description to "sum over the whole region" and keep average in the name? I could see how changing the name could be confusing since the flag that gets passed would still be called `average_over_region`....

gammapy/makers/utils.py
LauraOlivera: Completely agree. I just had not decided yet if I wanted to implement this here on in a separate PR. First I wanted to make sure I was doing things right without weights, and I plan to introduce it next

gammapy/makers/utils.py
adonath: This part I'm not sure about either...I think it's definitely good to adapt the description. In general we will have of course only one argument on the `Maker` and I agree for consistency we should choose the same name everywhere. However maybe there is a name that matches both cases approximately? Possibly `sum_over_region` is slightly better? In the keyword description of the `Maker` we mention the different cases anyway...

adonath: Another possibility is to name the other case: in a sense summing / averaging over the region I the correct thing to do and just taking the value at the center of the region is an approximation. So one could call the option e.g. `use_region_center=True` as well...and check for `if not use_region_center:`

LauraOlivera: I think I like `use_region_center` better actually - it is more accurate and I think that should be the case which choice is made explicit, not the most correct one

registerrier: Yes indeed that is a better choice.

gammapy/makers/utils.py
registerrier: OK indeed. 
I suppose one question is the default `binsz` that is defined on the `RegionGeom`. Is there a way to determine it a priori? e.g. `radius` or `width` /10?

adonath: Right now it's hard coded to `binsz=0.01` (https://github.com/gammapy/gammapy/blob/master/gammapy/maps/region.py#L48), this isn't ideal but in practice a choice that works for many cases I guess. We could add a parameter `RegionGeom(..., binsz_wcs=)`, which allows advanced users to modify the bin size to either improve accuracy or performance for large regions. By default we just work with `0.02` or `0.01` and the oversampling weights, I proposed below.

adonath: I'm not sure we can come up with a generic solution like `width/10`, because regions can have all sorts of shapes...

LauraOlivera: I agree that there should be an option to specify the bin size of the underlying `WcsGeom`. I tried to think of ways to get that info from the region size but I ran into the problem Axel describes below, because it might be tricky with weird region shapes. I think a reasonable default (0.01) + the weights is the way to go.

adonath: The only way I would see is to add some kind of system, where we declare for each shape, which bin size to use:

```python3
def get_region_binsz(region):
    if isinstance(region, CircleSkyRegion):
        return 0.1 * region.radius
    elif isinstance(region , ...):
        return 0.2 * region.width
    elif isinstance(region, CompoundRegion):
        regions = compound_region_to_list(region)
        return np.min([get_region_binsz(region) for region in regions])
    else:
        raise ValueError(f"Region {region} not supported")
```
I'm not sure whether it's worth it...

gammapy/makers/utils.py
LauraOlivera: should the default behavior then be `use_region_center=True` (like it is right now) or should we change it?

adonath: I think we should definitely keep the `use_region_center=True` in this PR. However just to get a feel on the effect of this option, I'd be interested to see how our analysis results / validation results change when the averaging / integration is applied somehow. I don't expect large changes, but who knows...

LauraOlivera: From what I've seen of how much the test values changes, it is quite a small effect. You can see it if you compare the values for the background from the test I implemented with those of the test above, which is the same dataset but without averaging...

adonath: Ah, indeed in this case the effect is smaller than 1%...

gammapy/makers/utils.py
```diff
@@ -35,23 +35,40 @@ def make_map_exposure_true_energy(pointing, livetime, aeff, geom):
...
 
     Returns
     -------
     map : `~gammapy.maps.WcsNDMap`
         Exposure map
     """
-    offset = geom.separation(pointing)
-    energy_true = geom.axes["energy_true"].center
+    if not use_region_center:
+        wcs_geom = geom.to_wcs_geom()
```
adonath, line 49: Here is one last suggestion to make it "perfect" in terms of performance. Creating a full WCS geometry here results in using a full 2d array of coordinates to be used for evaluation in `aeff.evaluate()`, in practice many of pixels of the 2d array are empty anyway (the ones outside the region). So the proposal here is to only use the flat coordinate list returned by `RegionGeom.get_wcs_coord()`. Look at the following minimal example:
```python3
import numpy as np

energy = np.array([1, 3, 10, 30])

a = np.ones((3, 2))
b = np.ones((6))
b_weights = 0.5 * np.ones((6,))

# a is a 2d array of coordinates
data_wcs = a * energy[:, np.newaxis, np.newaxis]
print(data_wcs.shape)

# b is a 1d array of coordinates
values = (b * energy[:, np.newaxis, np.newaxis])

# divide by sum of weight for the mean
data_region = np.sum(values * b_weights, axis=-1, keepdims=True) / b_weights.sum() 
print(data_region.shape) 
```
I now realise I was maybe a bit fast with the introduction of the `.get_wcs_weights()` method. Maybe we should modify `.get_wcs_coords()` to return both: the valid coordinates and the corresponding weights for each coordinate to represent the region. So the typical usage would be:
```python3
coords, weights = region_geom.get_wcs_coords()

values = np.sum(something.evaluate(coords) * weights)
```
Does this make sense?

gammapy/maps/region.py
```diff
@@ -291,6 +295,35 @@ def get_wcs_coord(self):
...
+
+        # if the region has non-spatial axis, this way is faster
+        # create the spatial weights only
+        wcs_geom_upsampled = wcs_geom.to_image().upsample(factor=factor)
+        data = self.contains(wcs_geom_upsampled.get_coord()).astype(float)
+        weights_no_axis = Map.from_geom(wcs_geom_upsampled, data=data)
+        weights_no_axis = weights_no_axis.downsample(factor=factor)
+        weights_no_axis.data /= weights_no_axis.data.max()
+
+        # And then make the map with the right dimensions
```
adonath, line 318: With the proposal I made above it is sufficient to create a 1d array of weights, so no need to broadcast to the length of the other axes here and below....

gammapy/maps/region.py
```diff
@@ -291,6 +295,35 @@ def get_wcs_coord(self):
...
+        data = weights_no_axis.data
+        for axis in self.axes:
+            data = np.stack(axis.nbin*[data])
+
+        weights = Map.from_geom(wcs_geom, data=data)
```
adonath, line 323: I think typically there is no need to store the weights in a map, I did in my example just to visualise it...

gammapy/makers/utils.py
LauraOlivera: that is a good idea. I agree that it is better. Indeed the "unnecessary" pixels were bothering me too and I think the what you propose of merging `get_wcs_coords` and `get_wcs_weights` is nice too. I'll implement it tomorrow.

gammapy/maps/region.py
LauraOlivera: good point :+1:

adonath (approved):
Thanks a lot @LauraOlivera , I added one commit, where I moved part of the code to compute the weights to `WcsGeom.region_weights()`, which I newly introduced. I thought this functionality might be useful on it's own, in some cases...

Good work! I'll go ahead and merge, once the CI is green.

Issue #2216, opened 2019-06-06
How to test logging in Gammapy


cdeil:
This is a reminder issue for myself to add a dev docs entry how to test logging in Gammapy.

It's triggered by https://github.com/gammapy/gammapy/pull/2214#discussion_r291045047 where @registerrier introduced the `capsys` pytest fixture (see https://docs.pytest.org/en/latest/capture.html#accessing-captured-output-from-a-test-function) in Gammapy, to check that some code generated a given log message.

That's one method we could use, probably there are others (that e.g. work via `logging` and not `stdout` and `stderr`). Should look around a bit and see what other projects use.

adonath:
We now have some recommendation how to test warning from the log output. https://docs.gammapy.org/dev/development/howto.html#caplog-fixture I think any more detailed testing of log output is not required. Will close the issue now.

PR #3203, opened 2021-01-25
Use caplog in FoVBackgroundMaker tests

Jaleleddine:
**Description**

This pull request adds caplog in test_fov

**Dear reviewer**

### DIFF
--- a/gammapy/makers/background/tests/test_fov.py
+++ b/gammapy/makers/background/tests/test_fov.py
@@ -126,7 +126,7 @@ def test_fov_bkg_maker_fit_nocountsnobackground(obs_dataset, exclusion_mask, cap
 
     assert caplog.records[-1].levelname == "WARNING"
     assert "Fit did not converge for test-fov" in caplog.records[-1].message
-    #assert "Background model parameters might be unphysical for test-fov" in caplog.records[-1].message
+    
 
 
 @requires_data()
--- a/gammapy/makers/background/tests/test_fov.py
+++ b/gammapy/makers/background/tests/test_fov.py
@@ -1,5 +1,9 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
+from typing import List, Dict
+
 import pytest
+import logging
+
 from numpy.testing import assert_allclose
 from astropy.coordinates import Angle, SkyCoord
 from regions import CircleSkyRegion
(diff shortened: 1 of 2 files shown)


gammapy/makers/background/tests/test_fov.py
```diff
@@ -1,5 +1,9 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
+from typing import List, Dict
```
adonath, line 2: It seems the import is unused...please remove.

adonath:
Thanks a lot @Jaleleddine, it seems there is still a test fail: https://github.com/gammapy/gammapy/pull/3203/checks?check_run_id=1763067685#step:9:1876, please fix. 

If I understood correctly `caplog` is a way to test logging output. In general we should probably add an entry in the developer documentation for this: https://docs.gammapy.org/0.18.2/development/howto.html, so that people know how to use it for other tests.

Otherwise the PR looks good to me!

registerrier (commented):
Thanks @Jaleleddine .

I think the CI fail comes from the absence of warning raised in the `FoVBackgroundMaker` with the `"fit"` option. This was recently removed and I think is a good indication that such tests are needed.

One connected issue is #2216  . At the time, the decision on the technical solution was delayed before deciding on a good choice for the logging itself. 
It seems that log testing is nevertheless needed and that `logging` is probably the solution we are going to keep at least for some time. 

Is the practical solution here satisfactory? I think that we should systematically test that "WARNING" are present in the 'log'. Should we test the log message as well?

registerrier:
BTW, any idea why test fails on linux and not on the MacOS CI?

adonath:
@registerrier I fully agree, that we will just keep in using `logging` for now and I agree it's good to have more systematic testing for this. However in general I would say this is already very detailed work and I think for now we do the testing on a best effort basis, where we think important warnings are emitted...

@registerrier Is the CI fail because the failing fit is emitted as info and not a warning? https://github.com/gammapy/gammapy/blob/master/gammapy/makers/background/fov.py#L145? I have no idea why this just appears on Linux...maybe @Bultako knows?

adonath:
@Jaleleddine I think you accidentally pushed to the wrong branch: https://github.com/gammapy/gammapy/tree/caplog, can you maybe check and push to this PR instead? Let me know (Slack) if you need help with that...

gammapy/makers/background/tests/test_fov.py
```diff
@@ -131,7 +126,7 @@ def test_fov_bkg_maker_fit_nocountsnobackground(obs_dataset, exclusion_mask, cap
 
     assert caplog.records[-1].levelname == "WARNING"
     assert "Fit did not converge for test-fov" in caplog.records[-1].message
-    assert "Background model parameters might be unphysical for test-fov" in caplog.records[-1].message
+    #assert "Background model parameters might be unphysical for test-fov" in caplog.records[-1].message
```
registerrier, line 129: Please remove commented line

registerrier (approved):
Thanks @Jaleleddine . See inline comment

adonath (approved):
Thanks a lot @Jaleleddine, I'll proceed and merge this PR now, please add an entry in the developer documentation in a follow up PR.

PR #3204, opened 2021-01-25
Select nearest well-defined irf for sources outside geom

QRemy:
This PR forces the MapEvaluator to use the nearest  well-defined IRF inside geom to evaluate sources outside geom that contributes inside. The goal is to avoid extrapolation that may be unstable depending on the IRF files and could cause npred to be NaN. Among two non-optimal solutions better choose the stable one, right ?

### DIFF
--- a/gammapy/datasets/map.py
+++ b/gammapy/datasets/map.py
@@ -5,11 +5,12 @@ import numpy as np
 import astropy.units as u
 from astropy.io import fits
 from astropy.nddata.utils import NoOverlapError
+from astropy.coordinates import SkyCoord
 from astropy.table import Table
 from astropy.utils import lazyproperty
 from regions import CircleSkyRegion
 from gammapy.data import GTI
-from gammapy.irf import EDispKernelMap, EDispMap, PSFKernel,  PSFMap
+from gammapy.irf import EDispKernelMap, EDispMap, PSFKernel, PSFMap
 from gammapy.maps import Map, MapAxis, RegionGeom
 from gammapy.modeling.models import BackgroundModel, DatasetModels, FoVBackgroundModel
 from gammapy.stats import (
@@ -1418,9 +1419,7 @@ class MapDataset(Dataset):
             kwargs["exposure"] = self.exposure.pad(pad_width=pad_width, mode=mode)
 
         if self.background is not None:
-            kwargs["background"] = self.background.pad(
-                pad_width=pad_width, mode=mode
-            )
+            kwargs["background"] = self.background.pad(pad_width=pad_width, mode=mode)
 
         if self.edisp is not None:
             kwargs["edisp"] = self.edisp.copy()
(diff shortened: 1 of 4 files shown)


gammapy/datasets/map.py
```diff
@@ -2510,13 +2509,31 @@ def update(self, exposure, psf, edisp, geom):
         # TODO: simplify and clean up
         log.debug("Updating model evaluator")
         # cache current position of the model component
+        geom2d = geom.to_image()
```
adonath, line 2512: Maybe we should rely on the IRF geom here and not the counts geom? Maybe just pass directly `MapDataset.mask_safe_psf` or `MapDataset.mask_safe_edisp` to `.update()`? The coarser bin size will make the update faster and accessing the IRF mask ensures consistency...

gammapy/datasets/map.py
```diff
@@ -2510,13 +2509,31 @@ def update(self, exposure, psf, edisp, geom):
...
+        if (
+            self.model.position is not None
+            and isinstance(geom2d, WcsGeom)
+            and not geom2d.contains(self.model.position)[0]
+        ):
+            coords = SkyCoord(
```
adonath, line 2518: You can directly use `geom.to_image().get_coord().skycoord`. As this is a very common thing to do, we could maybe also introduce `Geom.get_skycoord()` as a short-cut...

gammapy/datasets/map.py
```diff
@@ -2510,13 +2509,31 @@ def update(self, exposure, psf, edisp, geom):
...
+                geom2d.get_coord()["lon"].squeeze(),
+                geom2d.get_coord()["lat"].squeeze(),
+                frame=geom2d.frame,
+            )
+            separation = coords.separation(self.model.position)
+            separation[exposure.sum_over_axes().data.squeeze() == 0] = np.inf
+            ilon, ilat = np.where(separation == np.min(separation))
```
adonath, line 2525: Minor: you could use a flattened coordinate array and `np.argmin()`

gammapy/datasets/map.py
```diff
@@ -2510,13 +2509,31 @@ def update(self, exposure, psf, edisp, geom):
...
+            irf_position = coords[ilon[0], ilat[0]]
+            log.warning(
+                f"Center position for {self.model.name} model is outside dataset geom, using nearest IRF defined within geom"
+            )
+        else:
+            irf_position = self.model.position
```
adonath, line 2531: Maybe also introduce `MapEvaluator.irf_position`? Might be useful for debugging in some cases...

adonath (changes requested):
Thanks a lot @QRemy, this finally fixes a problem, we have known for some time now! I have left a few comments for simplification. Most importantly I think we should use `mask_safe_psf` or `mask_safe_edisp` to check the valid IRF positions, for consistency.

gammapy/datasets/map.py
QRemy: ok added mask_safe_psf to MapEvaluator argument, and some extra code to deal with the cases where mask_safe is None or psf is None.

gammapy/datasets/map.py
```diff
@@ -2506,16 +2508,41 @@ def update(self, exposure, psf, edisp, geom):
...
+        mask_safe_psf : `~gammapy.maps.Map`
+            Mask safe map of boolean type.
         """
         # TODO: simplify and clean up
         log.debug("Updating model evaluator")
         # cache current position of the model component
 
+        if mask_safe_psf is None:
+            mask = exposure.copy()
+            mask.data = (exposure.data != 0) & np.isfinite(exposure.data)
```
adonath, line 2520: I think we can safely assume that the `mask_safe` is "in sync" between the exposure and the IRF map (if this wasn't the case it would be a big problem anyway...). So creating the mask for the exposure should not yield any new information...

gammapy/datasets/map.py
```diff
@@ -2506,16 +2508,41 @@ def update(self, exposure, psf, edisp, geom):
...
+        else:
+            mask = mask_safe_psf
+        mask = mask.reduce_over_axes(func=np.logical_or)
+
+        if (
+            self.model.position is not None
+            and isinstance(mask.geom, WcsGeom)
+            and not mask.geom.contains(self.model.position)[0]
```
adonath, line 2528: I don't think this condition is correct, because the model can be at a position inside the geometry, but where no IRFs are defined because of the FoV cut. I think in this case we still want to use the nearest valid IRF...

adonath (changes requested):
Thanks @QRemy, I have left two more comments, once those are addressed, the PR is ready to merge from my side...

adonath (approved):
Thanks a lot @QRemy, no further comments from my side...

Issue #3184, opened 2021-01-12
Add option to not include runs with missing IRFs?


AtreyeeS:
Both the HESS fits productions have some runs in the index tables with some missing IRFs.

Currently, `datastore.get_observations()` raises a warning for these observations, but removing these observations is left to the user. The analysis breaks at the Map making stage.
This is quite unpleasant.

Maybe we can add an option `datastore.get_observations(required_IRF=all)  with all = ['bkg', 'psf', 'aeff', 'edisp']`? 

I think we decided at some point that checking the datastore is not the responsibility of Gammapy, but this use case seems to necessary.

Comments @registerrier @adonath ?

registerrier:
This sounds a reasonable feature. After all, many data production are in the making and we should not expect that fully clean and complete databases to be the norm any time soon.

adonath:
Solved in #3205.

PR #3205, opened 2021-01-27
Adapt Datastore to not include runs with missing IRFs

AtreyeeS:
**Description**

This pull request addresses #3184 
I added some tests, but since we do not ship imperfect datastores, it is difficult to test thoroughly. I tested locally on my machine for some HESS productions, it  works as expected.


adonath:
Thanks @AtreyeeS, I guess adding the v0.18.2 env file happened by accident? Please remove, I'll do a review later today...

AtreyeeS:
@adonath ! Sorry, they were indeed added by accident. removed them

gammapy/data/data_store.py
```diff
@@ -189,14 +189,25 @@ def info(self, show=True):
         else:
             return s
 
-    def obs(self, obs_id):
+    def obs(self, obs_id, required_IRF="all"):
```
adonath, line 192: Is it possible to expose this option just just in `.get_observations()`? Just as we do for `skip_missing`? I would argue that in `.obs()`, when a specific `obs_id` is given we should just return it...

gammapy/data/data_store.py
```diff
@@ -208,28 +219,30 @@ def obs(self, obs_id):
...
+            hdu = self.hdu_table.hdu_location(obs_id=obs_id, hdu_type=irf)
+            if hdu is None:
+                log.warning(f"Skipping run with missing IRF obs_id: {obs_id!r}")
+                return None
+            kwargs[irf] = hdu
 
-    def get_observations(self, obs_id=None, skip_missing=False):
+        return Observation(**kwargs)
+
+    def get_observations(self, obs_id=None, skip_missing=False, required_IRF="all"):
```
adonath, line 245: Rename to `required_irfs` (all lower-case and plural...). Or for consistency one could also choose `skip_missing_irfs=[]`...

In the loop you check the wether the corresponding IRF is `None`.

adonath (dismissed):
Thanks @AtreyeeS, I have left one comment regarding the name of the option and the location of the code...

gammapy/data/tests/test_data_store.py
```diff
@@ -88,6 +88,12 @@ def test_datastore_get_observations(data_store):
     observations = data_store.get_observations([11111, 23523], skip_missing=True)
     assert observations[0].obs_id == 23523
 
+    # Test that only desired IRFs are loaded
+    observations = data_store.get_observations([23523], required_IRF=["aeff"])
```
adonath, line 92: This behaviour I find a bit strange: so specifying `required_IRF=["aeff"]` prevents users from loading the background even if it is available? I think the behaviour should be such, that one just skips the run if a required IRFs is not present instead...

AtreyeeS: I tried to think about an optimal behaviour - the way I see it is this is the list of IRFs that will be used in the analysis. Since the other IRFs will not be used, they are not loaded.
Would you prefer to load the non-essential IRFs as well? Will adapt accordingly then.

Currently, one does skip the run if a required IRF is not present.

adonath: Yes, I think I prefer to allow to load the non-required IRFs as well. Forbidding this I think is too restrictive for no good reason. My preferable behaviour would be just to have the option in `.get_observations(..., skip_missing_irfs=[])` or `get_observations(..., required_irfs=[])` .

adonath:
As proposed by @registerrier this morning in the dev meeting: please add a unit test for this, by reading an existing HDU table and removing the corresponding entry you would like to test for.

AtreyeeS:
Thanks! Made the changes.
The tests took me some time because `data_store.hdu_table._hdu_type_stripped` is a `lazyproperty` and does not reflect the changes after changing the `hdu_table`. I do not know if there is a way to re-compute lazyproperties, so redefined it in the test file...

gammapy/data/data_store.py
AtreyeeS: Done!

gammapy/data/tests/test_data_store.py
```diff
@@ -88,6 +89,24 @@ def test_datastore_get_observations(data_store):
     observations = data_store.get_observations([11111, 23523], skip_missing=True)
     assert observations[0].obs_id == 23523
 
+    # Test that desired IRFs are loaded
+    hdu_table = data_store.hdu_table
```
adonath, line 93: Please make this a separate test and remove the scope definition of the `data_store` fixture: then a new data store will are created and you can safely modify the hdu table.

gammapy/data/data_store.py
```diff
@@ -258,7 +266,15 @@ def get_observations(self, obs_id=None, skip_missing=False):
                 else:
                     raise err
             else:
-                obs_list.append(obs)
+                flag = True
+                for irf in required_IRF:
+                    if getattr(obs, irf) is None:
```
adonath, line 271: I realised this implementation is problematic, because it will load existing IRFs from file, which makes it very inefficient. The problem is that once you call `getattr()` the `LazyFitsData` descriptor will load the data from disk. To avoid this you directly have to access hidden attributes. If you print the `Observation.__dict__` object:
```python3
{'obs_id': 23526,
 'aeff': None,
 '_edisp_hdu': <gammapy.utils.fits.HDULocation at 0x10601aee0>,
 '_psf_hdu': <gammapy.utils.fits.HDULocation at 0x1060b78e0>,
 '_bkg_hdu': <gammapy.utils.fits.HDULocation at 0x1060ab1c0>,
 '__gti_hdu': <gammapy.utils.fits.HDULocation at 0x1060b7070>,
 '__events_hdu': <gammapy.utils.fits.HDULocation at 0x1060b7a60>,
 'obs_filter': <gammapy.data.filters.ObservationFilter at 0x138f100d0>}
```
You can see, that missing irfs, are set to `None`, while the existing ones are defined as hidden data members with the `HDULocation` defining the location of the data. So you need work with e.g. `obs.__dict__.get(irf_name, False) is None` instead.

gammapy/data/data_store.py
```diff
@@ -238,12 +227,31 @@ def get_observations(self, obs_id=None, skip_missing=False):
...
+                * `edisp`: Energy dispersion
+                * `psf` : Point Spread Function
+            By default, all the IRFs are required.
 
         Returns
         -------
         observations : `~gammapy.data.Observations`
             Container holding a list of `~gammapy.data.Observation`
         """
+        available_IRF = ["aeff", "edisp", "psf", "bkg"]
```
adonath, line 245: Please stick with the Python convention and make IRF all lower-case. like `required_irf` and `available_irf`

adonath (changes requested):
Thanks @AtreyeeS, I have one more substantial comment on the implementation, which need too addressed. Let me know if you need further explanation on the descriptor pattern...

adonath:
Thanks a lot @AtreyeeS , I'll check out the behaviour one more time in a notebook myself, but so far it looks good!

adonath (approved):
Thanks @AtreyeeS , this works nicely...no further comments from my side.

PR #3207, opened 2021-01-27
Remove EnergyDependentTablePSF

adonath:
**Description**

This pull request remove the `EnergyDependentTablePSF`. The motivation is to reduce code duplication as well as number of abstractions to represent IRFs in Gammapy. It includes the following changes:

- Implement support for "gtpsf" on the `PSFMap`, which is represented using `RegionGeom`, unfortunately the meta information on the positing of the PSF is missing in the "gtpsf" format, so the region information is None.
- Implement `PSFMap.containment()` and `PSFMap.containment_radius`
- Implement `PSFMap.plot_containment_radius_vs_energy()` and `PSFMap.plot_psf_rad()`
- Implement `PSFMap.to_region_nd_map()`, which allows to extract a `PSFMap` with `RegionGeom` for given region / position

**Dear reviewer**
- Right now the `.plot_something()` methods always plot the PSF at the center of the map, however one could add a `position` argument. I'd prefer to implement such changes in follow up PRs...


gammapy/irf/psf/core.py
```diff
@@ -250,6 +216,42 @@ def plot_containment_radius(self, ax=None, fraction=0.68,  add_cbar=True, **kwar
 
         return ax
 
+    def plot_psf_vs_rad(self, ax=None, offset=[0] * u.deg, energy_true=[0.1, 1, 10] * u.TeV, **kwargs):
```
registerrier, line 219: Could it be possible to add a `offset = Angle(offset)` to allow strings?

gammapy/irf/psf/kernel.py
```diff
@@ -23,31 +23,30 @@ class PSFKernel:
     --------
     ::
 
-        import numpy as np
         from gammapy.maps import Map, WcsGeom, MapAxis
-        from gammapy.irf import EnergyDependentMultiGaussPSF, PSFKernel
+        from gammapy.irf import PSFMap
```
registerrier, line 27: Generally, the status of code examples in docstrings is not clear. I suppose for this type of lower level API this might still be useful since we probably don't explain much about it in the docs.

gammapy/irf/psf/psf_map.py
```diff
@@ -80,138 +75,216 @@ def psf_map(self):
...
+        exposure_psf = Map.from_geom(geom_exposure, unit="m2 s")
+        psf_map = Map.from_geom(geom, unit="sr-1")
+        return cls(psf_map, exposure_psf)
 
-        if position.size != 1:
-            raise ValueError(
-                "EnergyDependentTablePSF can be extracted at one single position only."
-            )
+    def to_region_nd_map(self, region):
+        """Convert to region ND PSF map
```
registerrier, line 98: Maybe  "Returns a PSFMap extracted in a region."

Maybe explain that this should work only with a `WcsGeom` (or `HpxGeom`?)

gammapy/irf/psf/psf_map.py
```diff
@@ -80,138 +75,216 @@ def psf_map(self):
...
-                "rad": 0 * u.deg,
-            }
-            data = self.exposure_map.interp_by_coord(coords).squeeze()
-            exposure = data * self.exposure_map.unit
+        Returns
+        -------
+        psf : `PSFMap`
+            PSF map with region geometry.
+        """
+        if region is None:
```
registerrier, line 112: Maybe add a check on the current geometry of the error and raise something if the type is already region?

registerrier (approved):
Thanks @adonath .
This looks good. Of course I could not look into details. I have put some comments inline.

gammapy/datasets/map.py
```diff
@@ -2524,12 +2515,11 @@ def update(self, exposure, psf, edisp, geom, mask_safe_psf):
             and np.any(mask_safe_psf.data)
             and isinstance(mask_safe_psf.geom, WcsGeom)
         ):
-            mask = mask_safe_psf.reduce_over_axes(func=np.logical_or)
-            coords = mask.geom.get_coord().skycoord
+            coords = mask_safe_psf.geom.get_coord().skycoord
```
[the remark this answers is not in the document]
adonath: This is an unrelated change, that happened when I rebased after #3204. I'll check again if it is justified...

docs/tutorials/fermi_lat.ipynb
```diff
@@ -497,30 +483,17 @@
...
+    "psf.plot_psf_vs_rad(energy_true=energy)\n",
     "\n",
-    "energy_range = [50, 2000] * u.GeV\n",
     "spectrum = PowerLawSpectralModel(index=2.3)\n",
-    "psf_mean = psf_table.table_psf_in_energy_range(\n",
-    "    energy_range=energy_range, spectrum=spectrum\n",
-    ")\n",
-    "psf_mean.plot_psf_vs_rad(c=\"k\", ls=\"--\", energy=[500] * u.GeV)\n",
+    "psf_mean = psf.to_image(spectrum=spectrum)\n",
+    "psf_mean.plot_psf_vs_rad(c=\"k\", ls=\"--\", energy_true=[500] * u.GeV)\n",
```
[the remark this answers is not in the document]
adonath: This has been there before, but I agree. It does two things: just plot one line and provide a label. I guess one could improve the behaviour, by choosing different defaults, depending on the size of the energy axis of the `psf_map`...

gammapy/datasets/map.py
```diff
@@ -101,23 +101,25 @@ def create_map_dataset_geoms(
 class MapDataset(Dataset):
     """Perform sky model likelihood fit on maps.
 
+    If an `HDULocation` is passed the map is loaded lazily.
```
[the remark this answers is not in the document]
adonath: It's not really meant to be used by normal users, but yes I can add a few more details...

gammapy/irf/psf/kernel.py
adonath: Yes, I agree those are useful examples...I think we will be able to test them soon?

docs/tutorials/fermi_lat.ipynb
```diff
@@ -530,7 +503,7 @@
     "psf_kernel = psf.get_psf_kernel(\n",
     "    position=geom.center_skydir, geom=geom, max_radius=\"1 deg\"\n",
     ")\n",
-    "psf_kernel.psf_kernel_map.sum_over_axes().plot(stretch=\"log\", add_cbar=True);"
+    "psf_kernel.to_image().psf_kernel_map.plot(stretch=\"log\", add_cbar=True);"
```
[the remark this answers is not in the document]
adonath: Yes, let's re-discuss this and I could improve it in a follow up PR...

gammapy/datasets/map.py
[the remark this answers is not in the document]
adonath: Done

gammapy/irf/psf/core.py
adonath: Done

gammapy/irf/psf/psf_map.py
adonath: I'd prefer to handle this uniformly with other methods such as `RegionNDMap.get_spectrum()` or similar. I'll add this as a discussion point for the next dev meeting...

gammapy/irf/psf/core.py
adonath: BTW: this is something we should check for uniformity in Gammapy. In general I can see how it is convenient, however it adds a lot of code, where we convert to `Quantity` explicitly and also need to document that strings can be passed as well...I'll also add it to the agenda for the next dev meeting.

PR #3208, opened 2021-01-28
Cleanup Dataset.plot_residuals()

AtreyeeS:
**Description**

This pull request cleans-up `Dataset.plot_residuals()` as discussed in a dev call last month.
`FluxPointDataset` now supports `diff` and `diff/model`, and spectral residuals are supported for `diff` and `diff/sqrt(model)`. Spatial residuals are supported for all 3 since no error bars are computed in this case.
For the `diff` case in `MapDataset`, the appropriate count statistics is used.

Also, `SpectrumDataset.to_spectrum_dataset()` is modified to raise a  `NotImplementedError` 

**Dear reviewer**


adonath (approved):
Thanks a lot @AtreyeeS, I think before the next release we need to unify the plotting between `MapDataset` and `SpectrumDataset` in general. So this is good start...

PR #3211, opened 2021-02-01
Refactor notebooks processing

Bultako:
This PR simplifies code in notebooks processing done during tests and documentation building. It mainly r**emoves dependency on the `notebooks.yaml` index file**, processing all notebooks (not only tutorials) present in the documentation by globbing in file extension `*.ipynb`. It also adds a feature to **skip notebooks execution** when the flag in metadata below is present in the notebook.

```
  "gammapy": {
    "skip_run": true
  }
```

Finally it updates related documentation


adonath:
Thanks a lot @Bultako for this simplification, it will help again to reduce the maintenance effort for notebooks. Is the testing of notebooks affect by this change? Or are again just all notebooks "globbed" and tested?

Bultako:
> Thanks a lot @Bultako for this simplification, it will help again to reduce the maintenance effort for notebooks. Is the testing of notebooks affect by this change? Or are again just all notebooks "globbed" and tested?

@adonath 
What it changes wrt tests is that all notebooks are "globbed" and tested, before this PR only notebooks declared in `notebooks.yaml`. And for doc building all notebooks are "globbed" and run, except those having the flag `skip_run` in the metadata.

adonath:
Thanks for the clarification! I think glob all notebook is fine, we don't want to have any un-tested notebooks in our repo anyway...

adonath:
Ok, to merge from your side?

Bultako:
Fine, merging now.

PR #3212, opened 2021-02-01
Remove hard-coded wcs bin size in RegionGeom and make it a parameter

LauraOlivera:
This pull request removes the hard-coded `binsz=0.01` in the `RegionGeom and` instead introduces a user-defined parameter, `binsz_wcs` with a default value of 0.01 to allow advanced users to change this value if needed.

### DIFF
--- a/gammapy/maps/region.py
+++ b/gammapy/maps/region.py
@@ -38,7 +38,8 @@ class RegionGeom(Geom):
     binsz_wcs : `float`
         Angular bin size of the underlying `~WcsGeom` used to evaluate
         quantities in the region. Default size is 0.01 deg. This default
-        value is adequate for the majority of use cases.
+        value is adequate for the majority of use cases. If a wcs object
+        is provided, the input of binsz_wcs is overridden.
     """
 
     is_image = False
--- a/gammapy/maps/region.py
+++ b/gammapy/maps/region.py
@@ -35,7 +35,7 @@ class RegionGeom(Geom):
         Non-spatial data axes.
     wcs : `~astropy.wcs.WCS`
         Optional wcs object to project the region if needed.
-    binsz : `float`
+    binsz_wcs : `float`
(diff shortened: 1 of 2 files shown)


LauraOlivera:
I wrote this now because I plan to explore how much the bin size and the choice between taking IRFs at the region center or not affect the resulting values, to add some  guidelines to the documentation of the `SpectrumDatasetMaker`. This feature is convenient for this.

gammapy/maps/region.py
```diff
@@ -46,16 +51,16 @@ class RegionGeom(Geom):
     _slice_spatial_axes = slice(0, 2)
     _slice_non_spatial_axes = slice(2, None)
     projection = "TAN"
-    binsz = 0.01
 
-    def __init__(self, region, axes=None, wcs=None):
+    def __init__(self, region, axes=None, wcs=None, binsz_wcs=0.01):
```
adonath, line 55: Maybe add the unit for the default by using a string: `"0.01 deg"`

adonath (changes requested):
Thanks @LauraOlivera! I have left one minor inline comment. Here is one more little suggestion: In general the bin size is accessible on the `.wcs`, property however we could could add a read-only property to the `RegionGeom`, so the the used bin size is more readable to users. Like:

```python3
@property
def binsz_wcs(self):
    # access wcs bin size here an create a Quantity (maybe similar to `WcsGeom.pixel_scales`) 
    return binsz

```

gammapy/maps/region.py
LauraOlivera: `WcsGeom.create()` needs `binsz` to be a float though. I can add `binsz_wcs = Angle(binsz_wcs)` before calling it, but maybe it makes more sense to add that line inside `WcsGeom.create()`?

adonath: Yes, please, I think we should support this in `WcsGeom.create()` as well.

adonath (approved):

PR #3216, opened 2021-02-03
Add a notebook to explain model management

AtreyeeS:
**Description**

This pull request adds a new notebook doing a joint Fermi-CTA analysis of the galactic center, explaining how datasets interact with models. 
I have named it `modeling.ipynb` for now, and renamed the previous `modelling.ipynb` to `fitting.ipynb`
This is a draft in progress. Please go through and give your comments @registerrier @adonath @QRemy @luca-giunti 

**Dear reviewer**

- I have not stripped the output cells yet, so that it is easier to see the output for the first round of review
- This notebook needs datasets produced by `fermi_lat.ipynb` and `analysis_3d.ipynb`: How do we distribute it? Do we ship it with `Gammapy-data`? It is 2 + 52 = 54 MB of data. Or do we ask users to compulsorily execute the above two notebooks before running this one (then this fails to run in binder though...)
- The output of `datasets.info_table` does not look ok - see cell [7] I will probably remove this cell from the notebook, but any idea why this is happening @adonath ?
- The notebook takes around 1 min to execute. One option might be to not show the fitting with the `gc-source`, but I believe the notebook is clearer with it than without.
- This notebook must be added to the notebook gallery with a thumbnail view etc. @Bultako, are there some instructions on how to do it?

ps:  I find the present model management very nice once I got used to it


QRemy:
Thanks @AtreyeeS, thanks for this draft. 
- In order to shorten the notebook and speed up its execution you could avoid to print the models/datasets and run the fit when you add the background and diffuse. Just show how to build the global model; define the parameters bounds and frozen status; print models;  add the global model into the datasets; print datasets; run the fit ; print results and `parameters.to_table()`.
- I would also remove TS computation that should be covered in the "3D detailed analysis " notebook otherwise this one will be more detailed than the "3D detailed analysis" notebook.
- As this notebook is more about model management you could show some new features of source selection. I would load 4FGL catalog using gammapy catalog then use `Models.select_mask()` or `Models.select_region` to pick the sources in our ROI; set their datasets_name to fermi_dataset; freeze them all , select by names some sources we want to fit jointly with CTA and unfreeze them (maybe only spectra) and set their datasets_name to None. 
- In order to speed up the fit you can change minuit options, if I remember well strategy 0 is recommended in minuit doc if more than ~10-20 free parameters as evaluation time is about ~O(N) for strategy 0 and ~O(N²) for strategy 1.
 ```
minuit_opts = {"tol": 0.1, "strategy": 0}
result = fit.run(backend="minuit", optimize_opts=minuit_opts)
```
- you can also reduce the number of free parameters by freezing spatial models or background models.

**Comment by adonath** [comment]
Thanks @AtreyeeS, this is an important contribution to our documentation! I'll start looking at the notebook now and think about how to distribute the data and maybe reduce the runtime of the notebook...

**Comment by AtreyeeS** [comment]
Thanks a lot @QRemy for the useful comments! `Models.select_mask()` is indeed a useful feature to show.
Passing minuit options is also a good idea.

I am not very inclined to remove the print statements totally - it is perhaps better if the notebook is readable and self-explanatory in our docs page - but I will try to cut down on the number of prints.

We should show an example of TS computation at least somewhere (and refer to it from the HowTo) - lets discuss this in the dev call.

File: docs/tutorials/fitting.ipynb [hunk 1]
```diff
@@ -0,0 +1,689 @@
+{
```
Bultako, line 1: <p>Thanks for this PR @AtreyeeS&nbsp;</p><p> </p><p>concerning how to select a thumbnail pic and add a tooltip for the gallery, I would recommend to open the notebook in a text editor, and edit the internal code there. It may sound risky, but it is much simpler. Find the code cell that produces&nbsp;the figure that you  would like for the gallery (for example this one :), and then replace the  --- "metadata": {}, ---  bit above the code cell with the snippet below:</p><p> </p><pre class="ql-syntax" spellcheck="false">"metadata": {
 "nbsphinx-thumbnail": {
  "tooltip": "Learn how to do perform a Fit in gammapy."
 }
</pre><p> </p><p>Note that you may write whatever you like after "tooltip"</p><p> </p><p>I will write all this process in the "Developer HowTo" of the docs, still missing.. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/gammapy/gammapy/pull/3216/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='54'/>

adonath:
@AtreyeeS Before I start and extensive review, can you please remove the output from the commit (ideally not with a follow-up commit, but overwrite the previous one). I think notebook output should never be committed, for review we will checkout you branch and execute the notebook anyway...

adonath:
Just a few preliminary comments already:
- Please don't use the "coolwarm" colormap when showing normal sky images, I think the bi-coloured cmaps should be used for residuals only
- There are a number of unused imports, please remove
- Looking at the output of the fermi dataset it seems there are no bins included for fitting, which means either `mask_fit` or `mask_safe` must be set to all `False`, not sure where this comes from. This is the reason for the unexpected `.info_table` output as well
- Setting units on the Fermi Galactic diffuse model shouldn't be needed anymore pleas remove.
- I think there is a number of functionality on the `Models` object that we should show as well:
    - `Models.select_region ()`
    - `Models.select(tag='sky-model')`
    - `Models.select_mask(datasets_names=)`
    - Possibly `Models.select_mask()`
    - `Models.freeze()` and `Models.unfreeze()` etc.


I'll do a more in depth review tomorrow, maybe we can find a way that you can preliminary share the required data.

adonath:
I think in general the focus is currently maybe too much on the specific model definition for this analysis, I don't have a specific proposal right now, but I think the head lines should be rather along the lines of `Assigning models to datasets`, `Selecting model types`, `Model component freezing and unfreezing`, instead of defining specific model components. In general I think the fitting is almost optional and should not be the focus at all, in that sense, I think one could simplify the fitting to only fit norms of the models and freezing all other parameters. Especially don't fit positions and extensions.

adonath:
I think we basically already have a predefined dataset for fermi: https://github.com/gammapy/gammapy-data/tree/master/fermi-3fhl-gc, however it's a bit lengthy to read it: https://docs.gammapy.org/0.18.2/tutorials/detect.html#Read-in-input-images, one way to simplify this would be to support a YAML index file on `MapDataset.read()`, this was already proposed in https://docs.gammapy.org/0.18.2/development/pigs/pig-008.html#dataset-serialization, but never implemented. What do you think @AtreyeeS and @QRemy ?

registerrier:
I am not sure we should introduce another way to serialize an individual dataset (e.g. yaml). Can't we rather have a main fits file with external HDUs? The hierarchical fits is probably a safer (but more cumbersome) way to export datasets.

adonath:
@registerrier I think we should definitely have the possibility to support multiple formats in `MapDataset.read()`, just because even the FITS format will evolve and we will need to support different versions of it anyway. Already now we will need to support e.g. the OGIP format for spectral datasets and soon will need to define another format, that keeps all the associated meta information, that is currently not supported by OGIP. 

Another obvious use case is, that it should be simple to read a dataset produced by Fermipy. As it uses the same `Map` classes, but the structure of the `MapDataset` will certainly be different and e.g. rely on an output folder and a file naming scheme, in which case something like `MapDataset.read("path/to/fermipy/data", format="fermipy"))` would be nice.

However I agree that introducing a YAML based index format is not necessarily the obvious solution. As discussed in the dev call, I'll check whether supporting filenames in `MapDataset.from_dict()` is an option and talk to @QRemy, whether he has time to implement it...

docs/tutorials/modeling.ipynb
```diff
@@ -4,665 +4,784 @@
    "cell_type": "markdown",
```
adonath, line 4: <p>I think the mask should be defined correctly on the fermi dataset already. Fixing it here is not the right place...</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/gammapy/gammapy/pull/3216/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='6'/>

adonath, line 4: <p>"fille_value": None is already the default...no need to redefine it here.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/gammapy/gammapy/pull/3216/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='12'/>

adonath, line 4: <p>Maybe directly use <code>print(models)</code>  here?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/gammapy/gammapy/pull/3216/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='19'/>

adonath, line 4: <p>Maybe just remove?</p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/gammapy/gammapy/pull/3216/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='24'/>

adonath, line 4: <p>Maybe we should add <code>SourceCatalog.to_models()</code>?  Might be useful...</p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/gammapy/gammapy/pull/3216/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='28'/>

adonath:
Thanks @AtreyeeS, I left a few minor comments, I'll see how to distribute the datasets now...

adonath:
@AtreyeeS I added a bundled Fermi dataset to Gammapy data in https://github.com/gammapy/gammapy-data/pull/8. This you can use here...

AtreyeeS:
Thanks @AtreyeeS . I will finish this PR now

AtreyeeS:
I have added `Models` functionalities as we discussed the other day. This now looks like an API notebook.
@QRemy and @adonath can you please review?

adonath:
Many thanks @AtreyeeS, the notebook is really nice and informative. I think it is mostly complete now and demonstrates nicely how to work with complex models. I have a few remaining thoughts, none of which is critical in any way:
- There is a bit of duplication of content with the `models.ipynb` tutorial on `Models.write()` and `Models.extend()`
- I think the plotting of source region is a very common thing to do, we could consider it adding as `Models.plot_regions()` and also include the name of the source in the plot.
- There are 2-3 places where printing the `Models` object might help, e.g. when demonstrating the freezing of parameters.
- The sections on combining / removing / selecting are rather short, they could be possibly combined...

adonath:
One more little thing: maybe use the "afmhot" colormap consistently for now?

docs/tutorials/model_management.ipynb
```diff
@@ -0,0 +1,815 @@
+{
```
QRemy, line 1: <p>We now use <code>Models.select_region()</code> to get a subset of models contributing to a particular region. You can also use <code>Models.select_mask()</code> to get models lying inside the True region of a mask map.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/gammapy/gammapy/pull/3216/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='29'/>

QRemy, line 1: <p>After that I would add a cell to explain the following:</p><p> </p><p>Note that <code>Models.select()</code> combine the different conditions with a AND operator. If one need to combine conditions with a OR operator they can use the <code>Models.selection_mask()</code> method to generate a boolean array that can be used for selection. For example:</p><p> </p><p>selection_mask = models_3fhl.selection_mask(name_substring="1748") | models_3fhl.selection_mask(name_substring="1731")</p><p>models = models_3fhl[selection_mask]</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/gammapy/gammapy/pull/3216/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='44'/>

QRemy:
Thanks @AtreyeeS for the changes. 
I have added two last suggestions about the explanations of `Models.select_region()` and `Model.select()`.

AtreyeeS:
Thanks @QRemy !
I made the changes as requested

adonath (approved):
Thanks @AtreyeeS, no further comments from my side...

PR #3222, opened 2021-02-04
Add alternative parametrization for ShellSpatialModel

QRemy:
This PR add an alternative parametrization for the ShellSpatialModel using outer radius and relative width, so eta=width/Rout  act as a shape parameter similarly than for the GeneralizedGaussian.

### DIFF
--- a/examples/models/spatial/plot_shell2.py
+++ b/examples/models/spatial/plot_shell2.py
@@ -23,6 +23,7 @@ and  :math:`r_{\text{in}}` is the inner radius.
 
 For Shell2SpatialModel, the radius parameter  r_0 correspond to :math:`r_{\text{out}}`
 and the relative width parameter is given as eta = :math:`(r_{\text{out}} - r_{\text{in}})/r_{\text{out}}`
+so we have :math:`r_{\text{in}} = (1-eta) r_{\text{out}}`.
 
 Note that the normalization is a small angle approximation,
 although that approximation is still very good even for 10 deg radius shells.
--- a/examples/models/spatial/plot_shell2.py
+++ b/examples/models/spatial/plot_shell2.py
@@ -1,0 +1,77 @@
+r"""
+.. _shell2-spatial-model:
+
+Shell2 Spatial Model
+===================
+
+This is a spatial model parametrizing a projected radiating shell.
+
+The shell spatial model is defined by the following equations:
+
+.. math::
+    \phi(lon, lat) = \frac{3}{2 \pi (r_{out}^3 - r_{in}^3)} \cdot
+            \begin{cases}
+                \sqrt{r_{out}^2 - \theta^2} - \sqrt{r_{in}^2 - \theta^2} &
+                             \text{for } \theta \lt r_{in} \\
+                \sqrt{r_{out}^2 - \theta^2} &
+                             \text{for } r_{in} \leq \theta \lt r_{out} \\
+                0 & \text{for } \theta > r_{out}
+            \end{cases}
+
+where :math:`\theta` is the sky separation, :math:`r_{\text{out}}` is the outer radius
+and  :math:`r_{\text{in}}` is the inner radius.
+
+For Shell2SpatialModel, the radius parameter  r_0 correspond to :math:`r_{\text{out}}`
+and the relative width parameter is given as eta = :math:`(r_{\text{out}} - r_{\text{in}})/r_{\text{out}}`
+
+Note that the normalization is a small angle approximation,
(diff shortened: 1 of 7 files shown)


gammapy/modeling/models/spatial.py
```diff
@@ -159,11 +159,6 @@ def integrate_geom(self, geom):
 
         return Map.from_geom(geom=geom, data=data.value, unit=data.unit)
 
-    @property
-    def evaluation_radius(self):
```
adonath, line 163: Why did you remove this? The purpose is to provide a default for custom implemented model, so that users are not required to implement an evaluation radius. But it was obviously missing a unit test...

gammapy/modeling/models/spatial.py
```diff
@@ -518,7 +513,7 @@ def evaluation_radius(self):
 
     def to_region(self, **kwargs):
         """Model outline (`~regions.EllipseSkyRegion`)."""
-        minor_axis = Angle(self.r_0.quantity * (1 - self.e.quantity))
+        minor_axis = Angle(self.r_0.quantity * np.sqrt(1 - self.e.quantity ** 2))
```
adonath, line 516: If the previous implementation was incorrect, can you please add a regression test, so that change is covered by a test?

gammapy/modeling/models/tests/test_spatial.py
```diff
@@ -224,6 +225,16 @@ def test_sky_shell():
     assert_allclose(radius.value, rad.value + width.value)
     assert isinstance(model.to_region(), CircleAnnulusSkyRegion)
 
+    model = Shell2SpatialModel(lon_0="1 deg", lat_0="45 deg", r_0=rad + width, eta=0.5)
```
adonath, line 228: Please introduce a separate test for the model...

gammapy/utils/regions.py
```diff
@@ -148,7 +148,14 @@ def f(x, coords):
         center = SkyCoord(lon * u.deg, lat * u.deg)
         return np.sum(center.separation(coords).deg)
 
-    result = minimize(f, x0=[0, 0], args=(positions,), bounds=[(-180, 180), (-90, 90)])
+    eps = 1e-5
```
adonath, line 151: This change seems reasonable, but unrelated? Did you see any test fails that motivate the change?

adonath (changes requested):
Thanks @QRemy, I have left some minor comments. I would propose to add a separate entry in the model Gallery for the `Shell2SpatialModel`, including the formula for the parametrisation.

gammapy/modeling/models/spatial.py
QRemy: It was duplicated

gammapy/utils/regions.py
QRemy: yes it was failing on my machine

gammapy/modeling/models/spatial.py
adonath: Thanks, indeed! I missed that...

gammapy/utils/regions.py
adonath: OK, interesting...this fail didn't appear in any of our CI builds. I guess one could choose the bounds even more reasonable, as the point must lie inside the polygon spanned by the others, but I can do this in a follow up PR. So just leave as is...

gammapy/modeling/models/spatial.py
QRemy: added

examples/models/spatial/plot_shell2.py
```diff
@@ -0,0 +1,77 @@
...
+            \begin{cases}
+                \sqrt{r_{out}^2 - \theta^2} - \sqrt{r_{in}^2 - \theta^2} &
+                             \text{for } \theta \lt r_{in} \\
+                \sqrt{r_{out}^2 - \theta^2} &
+                             \text{for } r_{in} \leq \theta \lt r_{out} \\
+                0 & \text{for } \theta > r_{out}
+            \end{cases}
+
+where :math:`\theta` is the sky separation, :math:`r_{\text{out}}` is the outer radius
+and  :math:`r_{\text{in}}` is the inner radius.
```
adonath, line 22: One last comment: could you at the definition of `R_in` with respect to `eta` here?

adonath (changes requested):
Thanks @QRemy, I have left one last comment,  once this is addressed this PR is ready to merge...

adonath (approved):

Issue #3143, opened 2020-11-30
Documentation list for 1.0


AtreyeeS:
As discussed in the last co-working week, I list here the action items required in gammapy 1.0 documentation effort.
Please feel free to add more items, and/or assign yourself to some particular task.

We decided to split notebooks into 

- Analysis notebooks - which do a full analysis
- API notebooks - which demonstrate a particular API.
The API notebooks need not be executed (but of course, must be tested to ensure that we do not ship broken code), so this releases some pressure on the docs-build time.

Small code examples in the rst pages will still remain.

**Major action items**

1. Have API notebook for each sub-package

- [ ] `gammapy.analysis` - notebook describing all options in HLI
- [x] `gammapy.maps`  - move `maps.ipynb` tutorial into subpage, remove code duplication in rst pages  (@QRemy)
- [x] `gammapy.makers` - notebook explain additional options not shown in current tutorials (@QRemy)
- [x] `gammapy.modeling` - move `models.ipynb`, and `modeling.ipynb` add examples for model management, use of `CompoundModels` (@AtreyeeS )
- [x] `gammapy.datasets` add general API tutorial (@QRemy)
- [ ] `gammapy.estimators` - Add API notebook
- [x] `gammapy.catalogs` - move over `catalogs.ipynb`, extend options in the notebook. (@QRemy)
- [ ] `gammapy.stats` - I find the present documentation quite good (caveat: some output cells not executed at present). I do not know if an API notebook is useful here (this is more conceptual than API based)
- [ ]  `gammapy.data`, `gammapy.irf`, `gammapy.astro`: less critical for API notebooks

2.   Add documentation for
- [x] Overview page to be updated - add a flowchart showing the Gammapy API and data levels (@adonath)
- [x] Add documentation for `RegionGeom` and `RegionNdMaps` (@LauraOlivera )
- [x] Add documentation for `FoVBackgroundMaker`
- [x] Improve documentation for stacking and handling of masks, add cross reference between existing documentation (@AtreyeeS )
- [x] Show how to use the `help()` command in more places / notebooks
- [x] Add plot for `RingBackgroundMaker` - use image from HGPS or Berge paper (@AtreyeeS)
- [x] Decide on what to do for `Dataset.plot_residuals` error bars (@AtreyeeS )
- [x] Generalise the exclusion mask tutorial to explain mask handling in general (boolean operators etc., erosion / dilation etc.), clarify conventions on inside / outside data, move it to the API tutorial section (@QRemy)
[...]

PR #3224, opened 2021-02-05
Unify handling of list of regions

adonath:
**Description**

This pull request unifies the handling of regions in methods like `Models.select_region()`, `EventList.select_region()` and `WcsGeom.region_mask()`. The behaviour is now as following:
- The argument allows for either a ds9 region str, a `Region` object or a list of `Region`
- Internally this always goes through `RegioGeom.from_regions()`, which handles the creation of the required wcs
- If there is no wcs attached to the object (such as `EventList`), then optionally a wcs can be passed by the user

I decided to remove the `inside=` argument from the `.region_mask()` methods, even if it is explicit, I didn't want to add it to the other methods as well. Since we have support for boolean operations on the `Map` object we can easily use the `~` operator, which is a standard Numpy operation. However a bit of documentation is missing on mask `Map` object, possibly we should extend the existing or add another maps tutorial...

### DIFF
--- a/docs/tutorials/analysis_2.ipynb
+++ b/docs/tutorials/analysis_2.ipynb
@@ -230,7 +230,7 @@
     "circle = CircleSkyRegion(\n",
     "    center=SkyCoord(\"83.63 deg\", \"22.14 deg\"), radius=0.2 * u.deg\n",
     ")\n",
-    "exclusion_mask = geom.region_mask(regions=[circle], inside=False)\n",
+    "exclusion_mask = ~geom.region_mask(regions=[circle])\n",
     "maker_fov = FoVBackgroundMaker(method=\"fit\", exclusion_mask=exclusion_mask)"
    ]
   },
@@ -603,9 +603,9 @@
    "autocomplete": true,
    "bibliofile": "biblio.bib",
    "cite_by": "apalike",
-   "current_citInitial": 1,
+   "current_citInitial": 1.0,
    "eqLabelWithNumbers": true,
-   "eqNumInitial": 1,
+   "eqNumInitial": 1.0,
    "hotkeys": {
     "equation": "Ctrl-E",
     "itemize": "Ctrl-I"
@@ -708,7 +708,7 @@
        "description": "Select energy:",
        "description_tooltip": null,
        "disabled": false,
-       "index": 0,
+       "index": 0.0,
        "layout": "IPY_MODEL_095494c994ba43aa96803803704a7cdb",
        "orientation": "horizontal",
        "readout": true,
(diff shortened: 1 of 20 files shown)


QRemy:
Tnanks @adonath, 
I have just minor comments:
- you can remove the comment in test_region_mask:
https://github.com/gammapy/gammapy/blob/37bdbfa446cf8fe4e5d1f855ec04d605a8ceab3d/gammapy/maps/tests/test_wcs.py#L360
- inside argument in boundary mask can also be remove (given how it is defined it do nothing useful anyway)
https://github.com/gammapy/gammapy/blob/37bdbfa446cf8fe4e5d1f855ec04d605a8ceab3d/gammapy/maps/wcs.py#L872

registerrier (approved):
Thanks @adonath . The change of regions list looks good. 

I have one general issue though.

The `inside` argument offers a bit of clarity regarding what the function does. I understand that it is not needed in practice but we must remember that this function is exposed to users and regularly used in tocs and tutorials. Removing it will break a number of users scripts and code-wise the gain is negligible. 

In practice, the `inside=True` is used only internally. All typical uses cases rely on `inside=False`. A minima, if it is to be removed, I would imagine it easier that the default is changed to `inside=False` so that it is not necessary for users to apply the `~` operator.

adonath:
Thanks @registerrier. The main reason why I didn't want to keep `inside=` is that I didn't want to add it as an argument to the other methods as well. If we have `.region_mask(inside=True)`, why not `select_region(inside=True)` as well, or `MapDataset.to_spectrum_dataset(region, inside=True)`? I think the `inside` argument introduces a new API for a very common Python / Numpy operation. While I agree, that its more explicit, I also think that the `~` operator should be know to any Gammapy user. 

I adapted the tutorials accordingly and added more documentation to the docstring. The most important thing is that we handle masks in Gammapy always, that they can be multiplied, i.e. "good data" is flagged with `True`. This is convention is true for how we handle exclusion mask as well. Changing the default behaviour does not make sense to me. I think returning an inverted region selection is even more confusing. And adapting other methods, like `.select_region()` to follow the same inverted behaviour does not make sense either. I guess we should just clarify once  again our convention for exclusion regions, everything else follows from that for users...

Thanks @QRemy, comments are addressed.

adonath:
@registerrier My proposal on documentation would be that we extend the current `exclusion_mask` tutorial and make it a more general "masks" tutorial. Which clarifies the conventions and in addition shows things like boolean operators for masks, `WcsMap.binary_dilation()`, `WcsGeom.boundary_mask()` etc. I'll a a point here: https://github.com/gammapy/gammapy/issues/3143

PR #3227, opened 2021-02-08
Add method  on geom to evaluate memory usage

QRemy:
Add geom.data_nbytes(dtype) in order to estimate memory usage in bytes of the Numpy data array matching this geometry depending on the given type.

### DIFF
--- a/gammapy/maps/geom.py
+++ b/gammapy/maps/geom.py
@@ -1991,7 +1991,7 @@ class Geom(abc.ABC):
         memory : `~astropy.units.Quantity`
             Estimated memory usage in megabytes (MB)
         """
-        return (np.empty(self.data_shape, dtype).nbytes * u.byte).to_("MB")
+        return (np.empty(self.data_shape, dtype).nbytes * u.byte).to("MB")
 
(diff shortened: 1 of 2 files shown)


gammapy/maps/geom.py
```diff
@@ -1962,6 +1977,17 @@ def data_shape(self):
...
+    def data_nbytes(self, dtype="float32"):
+        """Estimate memory usage in bytes of the Numpy data array
+        matching this geometry depending on the given type.
+
+        Parameters
+        ----------
+        dtype : data-type
+            The desired data-type for the array. Default is "float32"
+        """
+        return np.empty(self.data_shape, dtype).nbytes
```
adonath, line 1989: Maybe return a `Quantity` object here and convert to `MB` by default? And add the `Returns` section to the docstring...

adonath (changes requested):
Thanks a lot @QRemy, I have left one inline comment, otherwise looks good to me...

adonath (approved):
Thanks, @QRemy!  No further comments..

PR #3228, opened 2021-02-08
Remove EffectiveAreaTable class

adonath:
**Description**

This pull request removes the `EffectiveAreaTable` class. This class is basically un-used and not supported by our data reduction. The ogip arf serialisation has been fully replaced by the `RegionNDMap` already. The `.from_parametrisation()` method was moved to the `EffectiveAreaTable2D`, so it can be used for data reduction as well. 

**Dear reviewer**


gammapy/datasets/map.py
```diff
@@ -287,6 +287,11 @@ def excess(self):
         """Excess"""
         return self.counts - self.background
 
+    @property
```
registerrier, line 290: Is this really useful? I find it a bit confusing to have `aeff` on a `MapDataset`

gammapy/maps/geom.py
```diff
@@ -629,6 +629,10 @@ def assert_names(self, required_names):
         try:
             for ax, required_name in zip(self, required_names):
                 ax.assert_name(required_name)
+
+            if not len(self) == len(required_names):
```
registerrier, line 633: Shouldn't this be checked before the name assert?

gammapy/irf/effective_area.py
```diff
@@ -491,3 +191,53 @@ def peek(self, figsize=(15, 5)):
         self.plot_energy_dependence(ax=axes[0])
         self.plot_offset_dependence(ax=axes[1])
         plt.tight_layout()
+
+    @classmethod
+    def from_parametrization(cls, energy_axis_true=None, instrument="HESS"):
```
registerrier, line 196: Since this is now a 2D table, one might expect to get correct offset response. Maybe this should not be exposed any more. (After all this is mostly for testing right?)

registerrier (approved):
Thanks @adonath . I think this should work. 

Indeed, the `EffectiveAreaTable` is in practice a reduced IRF.
I wonder whether a simple `RegionNDMap` is a satisfactory container for DL4 IRF since some functionalities might be lost. Take the high and low thresholds, we might/should have maps of high and low thresholds in practice. Storing them in the `meta` might be insufficient.
Also `aeff.plot()` provides a plot without correct name (y label is `Data` and x label is `Energy_true`).

gammapy/irf/effective_area.py
adonath: Yes, this is only used for testing. I think adding a comment that the offset dependence is ignored is sufficient? In general I think such toy-models might be useful in certain situations (mostly testing). In combination with something like `EnergyDispersion2D.from_gauss()` and `EnergyDependentMultiGaussPSF` one can generate a full toy response..

gammapy/datasets/map.py
adonath: I though it was maybe useful, but I'm fine to remove it again...

gammapy/maps/geom.py
adonath: Yes, done.

adonath:
@registerrier Currently there is now information on what is contained in a `Map` object, so one cannot define a more meaningful axis label. But this is a more general problem...

adonath:
Thanks @registerrier!

PR #3229, opened 2021-02-09
Add SpatialModel renormalization for norm correction and performance

QRemy:
This PR adds a sub-class of SpatialModel that may need to be renormalized.
The renormalization can be use to ensure that a model integrates to unity over the sky correcting for norm approximations. 
Moreover the norm update can be disabled during the fit and computed only a posteriori in order to improve performances. It helps for models whose the norm computation require computationally heavy operation as numerical integration.

For now this apply to disk, gauss-general, gauss, shell, shell2 in order to compute the norm correction grids (see figures below). These figures show that the modification is useful only for disk and gauss-general

Also fix shell models that was not returned in sr-1 but in deg-2.

time_comparaisondisk
norm_correctiondisk

time_comparaisongauss-general
norm_correctiongauss-general_eta0dot9
norm_correctiongauss-general
norm_correctiongauss-general_eta0dot1

time_comparaisonshell2
norm_correctionshell2
time_comparaisongauss
norm_correctiongauss

### DIFF
--- a/gammapy/modeling/models/core.py
+++ b/gammapy/modeling/models/core.py
@@ -320,6 +320,11 @@ class DatasetModels(collections.abc.Sequence):
 
         return names
 
+    def renorm(self, geom=None):
+        """Renormalize spatial and spectral models if compatibles"""
+        for m in self._models:
+            m.renorm(geom)
+
     @property
     def names(self):
         return [m.name for m in self._models]
@@ -664,9 +669,7 @@ class DatasetModels(collections.abc.Sequence):
 
         for model in self.select(tag="sky-model"):
             if model.contributes(
-                    mask=mask,
-                    margin=margin,
-                    use_evaluation_region=use_evaluation_region
+                mask=mask, margin=margin, use_evaluation_region=use_evaluation_region
             ):
                 models.append(model)
 
--- a/gammapy/modeling/models/cube.py
+++ b/gammapy/modeling/models/cube.py
@@ -10,7 +10,7 @@ from gammapy.modeling.parameter import _get_parameters_str
 from gammapy.utils.fits import LazyFitsData
 from gammapy.utils.scripts import make_name, make_path
 from .core import Model, Models
-from .spatial import ConstantSpatialModel, SpatialModel
+from .spatial import ConstantSpatialModel, SpatialModel, SpatialModelRenorm
 from .spectral import PowerLawNormSpectralModel, SpectralModel, TemplateSpectralModel
 from .temporal import TemporalModel
 
@@ -144,6 +144,15 @@ class SkyModel(Model):
             parameters.append(self.temporal_model.parameters)
 
         return Parameters.from_stack(parameters)
+
+    def renorm(self, geom=None):
+        """Renormalize spatial and spectral models if compatibles"""
+        if isinstance(self.spatial, SpatialModelRenorm) and hasattr(
+            self.spectral, "amplitude"
+        ):
+            correction = self.spatial.norm_correction(geom)
+            self.spatial.nom.value /= correction
+            self.spectral.amplitude.value *= correction
 
     @property
     def spatial_model(self):
--- a/gammapy/modeling/models/spatial.py
+++ b/gammapy/modeling/models/spatial.py
@@ -330,8 +330,7 @@ class SpatialModelRenorm(SpatialModel):
         frame = kwargs.pop("frame", "icrs")
         update_norm = kwargs.pop("update_norm", True)
         super().__init__(**kwargs)
-        if not hasattr(self, "frame"):
-            self.frame = frame
+        self.frame = frame
         self.set_norm_method(update_norm)
 
     def __init_subclass__(cls, **kwargs):
@@ -363,6 +362,12 @@ class SpatialModelRenorm(SpatialModel):
     def compute_norm_default(self):
         return 1.0 / u.sr
 
+    @property
+    def evaluation_binsz(self):
+        """Evaluation bin size (`~astropy.coordinates.Angle`)."""
+        oversampling_factor = 100.0
+        return self.evaluation_radius / oversampling_factor
+
     def norm_correction(self, geom=None):
         """Norm correction to be multiplied with spectral model amplitude"""
         if geom is None:
@@ -375,7 +380,7 @@ class SpatialModelRenorm(SpatialModel):
             geom = WcsGeom.create(
                 skydir=self.position,
                 width=(2 * radius),
-                binsz=radius / 100.0,
+                binsz=self.evaluation_binsz,
                 frame=self.frame,
             )
         return np.sum(self.evaluate_geom(geom) * geom.solid_angle()).value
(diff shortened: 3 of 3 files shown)


adonath:
Thanks @QRemy, I have tried to understand a bit the motivation and consequences of this proposal and the plots you show. Here is my summary:
- This proposal solves two issues: compute normalisation for spatial models, where there is no analytical solution or where it's complicated to compute one. The second is to provide a faster numerical scheme for the integration, in case we rely on e.g. `scipy.integrate.quad`, which might be slow.
- The main idea is to provide a "default" geometry for each spatial model, that defines a geometry, the model is well sampled on.
- Looking at the performance plots, it seems its only relevant for the `DiskSpatialModel`. One could definitely speed up the computation, by implementing the  analytical solution for `e=0`. In general I wouldn't expect the spatial model evaluation to be the bottleneck and the fft convolution still dominates? But maybe I'm wrong?
- Looking at the correction plots one can see that it's either relevant for small or large extensions. In case of the small extensions, the flux in the "normal" model evaluation is not conserved, because we don't use integration. While this method fixes the issue of normalisation, it doesn't fit the problem, that the shape of the model changes, when it is not integrated. The latter I think is critical, because when fitting small extensions user often try to distinguish between a point source or extended model.
- For large extensions it provides the more correct normalisation for the `GeneralisedGaussianSpatialModel`

In general there are a few things, that I like about this proposal:
- Having a good default integration scheme for spatial models and having the possibility to use a non-normalised spatial model
- Letting the model declare it's default geometry, this is useful in terms of performance and precision to be sure to not under-sample the model and loose information. This will be useful in other situations as well.

What I don't like is a the introduction of a new class of spatial models and the two "modes". I wonder whether we can handle it maybe differently. Here are a few preliminary thoughts:
- I think to resolve the issue of normalisation for small extensions, we should rely on some kind integration scheme, most likely "oversampling". This will solve the normalisation issue for small extensions as well.
- For this it's definitely useful if model declare the default bin size or geometry
- Maybe it can be handled with `normalize` attribute, which would be consistent with the `TemplateSpatialModel` as well?
- I could imagine, that we introduce a general `SpatialModel.evaluate_norm()` method, which defaults to the numerical integration you propose and can be overridden with analytical solutions if available for sub-classes?
- In general the normalisation, can be maybe optimised such, that it is never the bottleneck and there is no need for the two different modes?

I general I like this functionality a lot, but I think we can improve the consistency and clarity of the implementation. I'll make a few more suggestions on this.

adonath:
I have thought a bit more about this and I think in a sense this PR fixes the "symptoms" of more underlying problems. I think we should resolve this problems in the code we currently have one by one first and then see how we fix the remaining ones. My first proposal would be to solve the issue for small bins using oversampling, which is something that we have discussed for a long time:
- In this PR the bin size is hardcoded to `evaluation_radius / 100`, I think this is not fully correct. If you consider e.g. a Shell, the required bin size does not really depend on the size of the shell, but rather the width, which defines the smallest feature in the model, that should be preserved. For the `DiskSpatialModel` it depends on the width of the edge etc.
- Instead we should probably add an `.evaluation_binsz()` property to our models, which declares the minimal bin size that should be used to evaluate the model. E.g. for a Gaussian it would return `sigma / 5` or similar, for the shell something like `width / 2` or similar. In general we have to decide per model I guess, which minimal bin size to use. In general there is a generic solution to determine the required bin size, based on the containment of distribution of the spatial frequencies (see https://github.com/adonath/notebooks-public/blob/master/psf_bin_size.ipynb), but I think this leads too far for now.   
- Either in `.integrate_geom()` (or probably even better in the `MapEvaluator`, where we can work with an upsampled exposure and PSF as well...and check the upsampled geom) we check for the bin size of the geom and upsample to the minimal required bin size. After evaluation of the model we downsample again to the initial bin size. 
- Later this information can be even used to set an optimised bin size for large models to improve performance.

@QRemy Would you agree to work on this first?

QRemy:
ok I will add oversampling first in a separate PR.

adonath:
For the record: yesterday we had a shot call on this with @QRemy, @registerrier and @luca-giunti, with the following conclusions:
- There was a misunderstanding from my side: I thought the incorrect normalisation was because of under-sampling of the model. This is not the case as the normalisation plots are computed using an adaptive bin size of `evaluation_radius / 100`. Instead the effect is likely because of the incorrect definition of the evaluation radius of the disk (which misses the edge parameter) and of the generalised Gaussian (which uses and approximation for eta close to one, which should be improved), This should fix the issues we see for small sources.
- @QRemy Will add the scripts to compute the check plots to https://github.com/gammapy/gammapy-benchmarks and re-run the checks with the fixed definition of the evaluation radius.
- We will discuss once again to possibly unify the definition of our spatial models. Currently the disk and the symmetric Gaussian are normalised on the sphere while all other models are normalised on the tangential plane. This makes the definition in consistent. However the behaviour is documented correctly. 
- To improve the consistency one could always apply the normalisation on the tangential plane, which is correct up to a few deg. To handle the case of very large source for the disk and gaussian model, one could introduce a `SphericalGaussianSpatialModel` and a `ShericalDiskSpatialModel`, carefully documenting, when it should be used.
- For the normalisation of these large spherical models, one can rely on the method proposed in this PR. The only requirement would be to change the implementation such, that it is not exposed to the usre with `.renormalise()` methods, but remains an implementation detail.
- The oversampling of models is still useful and will be implemented independently. To improve the performance of the oversampling @registerrier proposed to use a different cutout for the model evaluation then for the PSF convolution.

QRemy:
Here are the norm correction plots after fixing the evaluation radius.
Note that I changed the range of colorbar to [0.9, 1.1] to better see the small changes

**Disk:**
The evaluation radius is now r_eval = r_0 + edge.
The normalization is still incorrect when the edge width is not much smaller than the radius because the norm computation does not account for the edge.

norm_correctiondisk

**Generalized Gaussian:**
The evaluation radius is defined as r_eval = r_0*(1+8*eta) so it verifies:
      r_eval -> r_0 if eta -> 0 
      r_eval = 5*r_0 = 5*sigma_gauss if eta=0.5
      r_eval = 9*r_0 > 5*sigma_laplace = 5*sqrt(2)*r_0 ~ 7*r_0 if eta = 1
      r_eval -> inf if if eta -> inf

The normalization is now well defined except for very large radius( >~10 degree) but even in this case the error remains lower than few percent.

norm_correctiongauss-general

norm_correctiongauss-general_eta0dot9

norm_correctiongauss-general_eta0dot1

norm_correctiongauss

norm_correctionshell2

adonath:
Thanks a lot @QRemy, this looks excellent! So for the generalised Gaussian even for the larger radii the tangential approximation seems to hold well and the previous effect was again because of my poor parametrisation for the evaluation radius... 

 I'll think about once more, whether there is something we can do about the `DiskSpatialModel`.

adonath:
Remaining action items:
- @adonath or @luca-giunti Fix norm computation for `DiskSpatialModel` to account for the edge
- @QRemy Add a new folder `gammapy-benchmarks/validation/checks/model-norm` and add a `make.py` producing the plots above and save in a results folder. The checks don't run regularly, but are done once and can be extended if new model are implemented.

adonath:
@QRemy  I checked the norm issue for the `DiskSpatialModel` once again and it is not really because the `edge` is not taken into account, but because of the way `r_0` is defined:




`r_0` is defined as the radius where the edge drops to half of its height. So effectively one has to require `r_0 > edge`, otherwise the current definition of the edge does not make any sense. So the `edge` parameter implicitly defines a minimum value for `r_0`. So the issue is basically independent of the absolute value of `r_0`, but just depends on the ratio. So we could e.g. change the definition of `edge` parameter to be relative to `r_0`.

QRemy:
ok with that solution, for now the edge is 1% of r_0 for the default parameters, we could keep that.
We should also mention in the doc that r_0 cannot be smaller than the bin_size so users can set the min accordingly.

adonath:
@QRemy OK, I'll prepare a PR to change the definition of `edge`...

adonath:
@QRemy committed the checks in https://github.com/gammapy/gammapy-benchmarks/pull/105. Everything looks fine now. The only remaining improvement here would be to implement model oversampling, but this can happen in a different PR. Will close this one now. Thanks again @QRemy!

PR #3230, opened 2021-02-09
Expand extended source analysis tutorial

LauraOlivera:
This pull request extends the extended source analysis tutorial to include the recently introduced option of using the `SpectrumDatasetMaker` with `use_region_center=False`, which averages the IRFs over the region. The results are then equivalent to the previous approach, which uses a `MapDatasetMaker` in the data reduction step and transforms it into a `SpectrumDatasetMaker`.

I kept the first part of the notebook almost exactly as it was, but I modified the introduction to explain that there are two ways to do the extended source analysis. Then, after the 3D case, I added the 1D case. A lot of cells are exactly the same as above, because there aren't that many differences between the two once the data reduction is done. In the end, I compared the resulting spectrum from both approaches.

This is the first time I change a notebook, and besides some instructions in https://docs.gammapy.org/dev/development/intro.html#contribute-with-jupyter-notebooks, I couldn't find more information. So maybe there is something else I can do so that the notebook can easily be looked at by a reviewer, please let me know.

This pull request also fixes a number of small issues I found along the way, namely:

- Set the right value for the `livetime` in `MapDataset.to_spectrum_dataset()`
- Downgrade the default `binsz_wcs` in the `RegionGeom` to speed up IRF evaluation. More careful optimization is planned in a follow-up PR. The "old" default is still used in the tests, to avoid having to change the values.
- Add the `binsz_wcs` option to the `RegionNDMap`, which was missing.
- Add a function `RegionGeom.to_binsz_wcs()` to easily change the `binsz_wcs` value in a region geometry.
- Fix a Matplotlib deprecation warning in `SpectralModel.plot()` (keyword `nonposy`, `nonposx` have changed to `nonpositive`)

### DIFF
--- a/docs/tutorials/extended_source_spectral_analysis.ipynb
+++ b/docs/tutorials/extended_source_spectral_analysis.ipynb
@@ -9,7 +9,7 @@
     "## Prerequisites:\n",
     "\n",
     "- Understanding of spectral analysis techniques in classical Cherenkov astronomy.\n",
-    "- Understanding of how the spectrum and cube extraction API works, please refer to the [spectrum extraction notebook](spectrum_analysis.ipynb) and to the [3D analysis notebook](analysis_2.ipynb).\n",
+    "- Understanding the basic data reduction and modeling/fitting processes with the gammapy library API as shown in the [first gammapy analysis with the gammapy library API tutorial](analysis_2.ipynb)\n",
     "\n",
     "## Context\n",
     "\n",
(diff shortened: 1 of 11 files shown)


LauraOlivera:
The counts map that was the thumbnail in the tutorial gallery is now gone... not sure if that is an issue

adonath:
Thanks a lot @LauraOlivera! I don't; have any further comments, just another idea:

For illustrational purpose one could add a plot like this:
image

Here is the basic code:

```python3
from matplotlib.collections import PolyCollection
import numpy as np

def plot_wcs_grid(geom, ax, **kwargs):
    """"""
    geom_wcs = geom.to_wcs_geom()
    coord = geom_wcs.get_coord(mode="edges").skycoord

    # define pixel corners
    low_left = coord[:-1, :-1]
    low_right = coord[1:, :-1]
    up_left = coord[:-1, 1:]
    up_right = coord[1:, 1:]

    vertices = []

    for idx in np.ndindex(low_left.shape):
        vertices.append([
            (low_left[idx].l.deg, low_left[idx].b.deg),
            (low_right[idx].l.deg, low_right[idx].b.deg),
            (up_right[idx].l.deg, up_right[idx].b.deg),
            (up_left[idx].l.deg, up_left[idx].b.deg),
        ])


    grid = PolyCollection(
        vertices,
        transform=ax.get_transform("galactic"),
        **kwargs
        )

    path = geom.region.to_pixel(ax.wcs).as_artist(**kwargs)

    ax.add_artist(path)
    grid.set_clip_path(path)
    
    ax.add_collection(grid)

plt.figure(figsize=(12, 8))
m = Map.read("$GAMMAPY_DATA/fermi-3fhl-gc/fermi-3fhl-gc-counts.fits.gz")

_, ax, _= m.cutout(m.geom.center_skydir, "6 deg").plot(stretch="sqrt");

geom = RegionGeom.create("galactic;circle(0, 0, 2)", binsz_wcs=0.5)

plot_wcs_grid(geom, ax=ax, ec="tab:blue", fc="None", lw=2)
```
This could be added as `RegionGeom.plot_wcs_pixels()` or `WcsGeom.plot_pixel_grid()` or similar. For analysis it's not really useful, but there are maybe 2-3 places in the docs, where one could use this for illustration. Alternatively one could just add the image above as png to the docs and use it this way...

adonath:
However this could be done in any follow-up PR...

LauraOlivera:
ohh I like that. Indeed, I think a follow-up PR would be best for that though...

adonath (approved):
Thanks @LauraOlivera, no further comment from my side.

PR #3234, opened 2021-02-10
Modify bounds handling and scan values generation in estimators

QRemy:
This PR remove hard coded bounds in estimators and simplify their handling.
The parameters `norm_min` and `norm_max` now refers to the limits of the fit and not of the profile scan.
Now the values of the profile scan can only be set using `norm_values`.
I introduced two function `ScanValuesMaker` and `make_scan_bounds`  in order to help generating the bounds/scan values automatically.

 The new default settings  should prevent the upper limit to vanish when the parameter error is missing or poorly defined.

### DIFF
--- a/gammapy/datasets/tests/test_flux_points.py
+++ b/gammapy/datasets/tests/test_flux_points.py
@@ -8,6 +8,7 @@ from gammapy.estimators import FluxPoints
 from gammapy.modeling import Fit
 from gammapy.modeling.models import PowerLawSpectralModel, SkyModel
 from gammapy.utils.testing import mpl_plot_check, requires_data, requires_dependency
+from gammapy.estimators.parameter import ScanValuesMaker
 
 
 @pytest.fixture()
@@ -110,7 +111,9 @@ class TestFluxPointFit:
         fit.backend = "minuit"
         result = fit.run()
 
-        profile = fit.stat_profile("amplitude", nvalues=3, bounds=1)
+        parameter = fit._parameters["amplitude"]
+        values = ScanValuesMaker(bounds=1, n_values=3)(parameter)
+        profile = fit.stat_profile("amplitude", values)
 
         ts_diff = profile["stat_scan"] - result.total_stat
         assert_allclose(ts_diff, [110.1, 0, 110.1], rtol=1e-2, atol=1e-7)
--- a/gammapy/datasets/tests/test_flux_points.py
+++ b/gammapy/datasets/tests/test_flux_points.py
@@ -8,7 +8,7 @@ from gammapy.estimators import FluxPoints
 from gammapy.modeling import Fit
 from gammapy.modeling.models import PowerLawSpectralModel, SkyModel
 from gammapy.utils.testing import mpl_plot_check, requires_data, requires_dependency
-from gammapy.estimators.parameter import ScanValuesMaker
+from gammapy.estimators.parameter import ScanValuesGenerator
 
 
 @pytest.fixture()
@@ -112,7 +112,7 @@ class TestFluxPointFit:
         result = fit.run()
 
         parameter = fit._parameters["amplitude"]
-        values = ScanValuesMaker(bounds=1, n_values=3)(parameter)
+        values = ScanValuesGenerator(bounds=1, n_values=3)(parameter)
         profile = fit.stat_profile("amplitude", values)
 
         ts_diff = profile["stat_scan"] - result.total_stat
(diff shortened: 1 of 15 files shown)


adonath:
Thanks a lot @QRemy! Following up the discussion in the dev meeting, I think it would be good separate this PR into adding the convenience functionality to compute parameter scan values and changing the bounds of the `norm` values for estimators. I think the convenience to create parameter scan values is uncontroversial and nice functionality, changing the norm bounds / UL definition might require some more discussion, because it should be consistent within Gammapy. So a change requires adapting the `TSMapEstimator` etc. as well.

Here are a few further comments on the scan value creation:
- A "Maker" class is by now a well defined API in our data reduction, so I would not use it here.
- As the error information is now available on the `Parameter` object, maybe it's most convenient to add the functionality directly as `Parameter.get_scan_values()` and `Parameter.get_scan_bounds()`?
- I think at some point we should add something like `Parameter(error_default=)`, where we can pre-define error assumptions for parameters for any built-in model. This will help e.g. to choose initial step sizes for optimisers and thus improve convergence as well. 
- Instead of the simple parameter scaling we could possibly introduce something like `Parameter(scale="log")` as well, but this is definitely another PR on it's own...

QRemy:
For now :
- renamed ScanValuesMaker to ScanValuesGenerator
- moved make_scan_bounds() to parameter.get_scan_bounds()
- renamed `norm_min`/`norm_max` to `norm_fit_min`/`norm_fit_max` and `norm_values` to `norm_scan_values`
- restore estimators previous defaults for norm_fit_min/max and added a line in the doctring about `norm_fit_min` : 
" In order to follow Fermi-LAT science tool convention `norm_fit_min` must be set to 0."

For other PRs:
- it's hard to define good default for parameters errors without extensive tests to check the effect on the fit convergence, we should keep that for after 1.0
- I agree about scaling
- FluxPointsEstimator.run() should have an optimize_option
- could add an option ul_method: {"confidence", "profile"} to run the ul computation using  `confidence_scipy` or `stat_profile_ul_scipy`. (and the same for errors ?).
`stat_profile_ul_scipy` uses brentq that fails for complex profiles when f(bounds) are not of opposite signs, maybe we could avoid this by simply slicing the profile in inversible intervals (separated by local min/max) and run brentq in the nearest valid interval where  f(bounds) are of opposite signs.

QRemy:
@adonath could we merge this PR since the big refactoring of estimators seems still far away. This PR simplifies and clarifies the bounds definitions for estimators which is important 0.19

adonath:
@QRemy I'm just coming back to this.  After some thought I think we should directly attach the scan information to the `Parameter` object, like so:

```python3

class Parameter:    
    def __init__(self, ..., scan_min=3, scan_max=3, scan_n_values=1, interp="lin")
    
    @property
    def interp(self):
        """"""
        # interp does not really make sense, but it is the same for MapAxis
        # and Parameter.scale already exists 
        # only used for scan values now
        return self._interp
    
    def scan_values(self):
        """"""
        # generate values here
        # could use MapAxis here 
        pass
    
    @scan_values.setter
    def scan_values(self):
        """"""
        # could allow settinfg scan values directly
        pass
    
    @property
    def scan_n_values(self):
        pass
    
    @property
    def scan_min(self):
        """"""
        # an int is intepreted as n * error
        pass
    
    @scan_min.setter
    def scan_min(self, value):
        """"""
        pass
    
    @property
    def scan_max(self):
        """"""
        pass
```

In future the scan info could be serialised along with the model, for now the in memory representation is just fine.

The plan is still that `Estimator` objects also become serialisable in future (e.g. as YAML), so I'm strongly suggesting to only introduce simple "serialisable" objects for configuration such as `str`, `int`, `dict` etc. This is not the case for the `ScanValueGenerator`.  My proposal would be to make one PR adding the scan info / config to the `Parameter` object and once #3372 is merged, we continue adapting the `Estimator` API and see how the info is best passed on to the `Parameter` object.

In general I like the functionality a lot, I'm just not fond about introducing a new object API, which agains requires a new solution for serialisation and a new API for users to learn. What I prefer about attaching the info to the `Parameter` object is the future potential for serialising it with the model and the hierarchical "per parameter" access that users use to interact with parameters anyway.

QRemy:
ok I convert this PR to draft and will open new ones to:
(1) define or generate scan values on parameter object.
(2) change estimator norm_min and norm_max so it refers to the limits of the fit and not of the profile scan, use the new parameter.scan_values for the profile scan.

adonath:
Thanks @QRemy, sounds good! Maybe just wait with the 2nd PR until #3372 has been merged to avoid merge conflicts...I'll assign you for review as well...

adonath:
Thanks a lot @QRemy for this prototype! As proposed implemented the scan specification on the `Parameter` object in #3406. The separate fit boundary handling can be implemented in a follow up PR. Will close this PR now.

PR #3235, opened 2021-02-11
Add regression tests for background with  missing HDUCLAS keyword

adonath:
**Description**

This pull request adds support for reading background files with missing `HDUCLAS` keywords. I decided to raise a warning, because I think invalid DL3 files should not be read silently.


AtreyeeS (approved):
Thanks a lot @adonath ! Checked the HAP-fr and HAP-hd productions - it works now.

Issue #2482, opened 2019-10-23
Defaults for IRF axes 


AtreyeeS:
We need a good choice of defaults for the IRF axes binning, `rad`, `migra` and `energy_axis_true`. Currently, `energy_axis_true` defaults to `e_reco` (PR #2479), which can be problematic in most cases, and the `rad` and `migra` axes have a fixed default (hard-coded quite randomly). These default values should ideally be chosen from the IRFs.

This is a reminder issue to discuss and implement this properly.

cdeil:
Discussed in the Gammapy weekly call just now.

@AtreyeeS - Please create a brief overview of defaults we have now.
It could be a brief section after this one:
https://docs.gammapy.org/0.14/scripts/index.html#data-reduction-and-datasets

Probably we have duplicated and inconsistent defaults in different places, and also the choices of the defaults can be improved.

Once we have the overview, we can continue the discussion here concerning how to improve.

AtreyeeS:
I think it would be better to edit the docs section once we have a final decision.
For now, I just summarise the present status here.

1.For 1D spectral analysis
(i) In `SpectrumDatasetMaker` : `e_true` defaults to `e_reco` unless explicitly specified.
(ii) In `SpectrumEvaluator`: `e_true` is taken from the IRFs, specifically the effective area (**even when** `e_true` is explicitly passed, the `e_true` is taken from the IRF. When no IRFs are given, the user defined value is taken.)  `e_reco` is taken from the `edisp`. If no `edisp`, then it defaults to `e_true`.
(iii) `SpectrumDataset`: Makes **no** distinction between `e_true` and `e_reco`, and takes the `energy_axis` from the `counts` or the IRFs. The `SpectrumDataset.create()` defaults `e_true` to `e_reco`, unless passed by the user.

Accessing through the high level interface, the `template-1d.yaml` does not provide any distinction between `e_true` and `e_reco`, and allows to user to specify only an `energy` axis. Thus, for practical purposes, `e_true` is always the same as `e_reco`.

2. For 3D Map  analysis
(i) The `energy_axis_true` defaults to the reco axis unless specified (in `MapDatasetMaker` and `MapDataset`.) 
(ii) Binnings for `rad` and `migra` are set completely arbitrarily to 
```
RAD_AXIS_DEFAULT = MapAxis.from_bounds(
    0, RAD_MAX, nbin=66, node_type="edges", name="theta", unit="deg"
)
[...]

PR #3237, opened 2021-02-12
Improve spectral analysis tutorial

cosimoNigro:
This PR improves the spectral analysis tutorial of the H.E.S.S. DL3 data

I simply switched ON the containment correction (was `False` but these are full-enclosure DL3 data) and reduced the number of true energy bins. You had 200 before, I put 30. I realised with other 1D analyses that making too many true energy bins will make the IRFs too noisy and will give trouble when fitting (folding) models more complicated than a power-law. Remember when folding, we integrate over the entire true energy axis - [first formula in IRF theory](https://docs.gammapy.org/dev/irf/theory.html#irf-theory) - so the smoother the IRF in true energy the easier the (folding and) fitting will be.

Before my changes:
tutorials_spectrum_analysis_56_1

After:
index

You can actually fit a power law with cut-off to those four runs, and the comparison with the results from the H.E.S.S. 2006 reference looks much better. 

Let me know what you think.
Thanks.


adonath:
Thanks @cosimoNigro! I only have a few general comments:

- I'd be surprised if a fine true energy binning lead to unstable fit result. Yes, the IRF can be noisy, but during the application of the RMF matrix the fine true energy bins are summed again per reconstructed energy bin and I would expect the noise to cancel out. So the predicted number of counts should be stable and not change with the choice of the true energy bins (given those are fine enough to sample the model well...). 
-  That being said I fully agree 200 bins in true energy is too much and decreases the runtime of the fit for no good reason. Instead of specifying  the total number of bins, one could possibly choose the bins per decade like:  `MapAxis.from_energy_bounds(..., nbin=20, per_decade=True, name="energy_true")`
- Changing to an exponential cutoff is fine from my side. In general I think in a tutorial we don't have to present perfect agreement with references, just because that's not the purpose of a tutorial. However it certainly doesn't hurt either and showing the variety of models  (I think in most tutorials we only use a power law...) is a plus as well.

adonath:
@registerrier I remember you had some plots showing the "ideal" binning to sample a given spectra model well. Are those public anywhere, or could it be made public e.g. in https://github.com/gammapy/gammapy-recipes?

adonath:
Concerning the energy binning: I guess in practice it doesn't make sense to go finer than the binning the IRFs are delivered with because there is no information in the smaller bins. I just checked:

```python3
>>> from gammapy.data import DataStore
>>> from gammapy.maps import MapAxis

>>> datastore = DataStore.from_dir("$GAMMAPY_DATA/hess-dl3-dr1/")
>>> obs = datastore.obs(23523)
>>> print(obs.aeff.axes["energy_true"].nbin_per_decade)
23.99999994175657
```
And the HESS effective area comes with 24 bins per decade. I think this is basically the ideal choice. We already planned to introduce defaults for IRF axes (see https://github.com/gammapy/gammapy/issues/2482), so that the users does not need to care about this at all, but so far we haven't.

cosimoNigro:
Thank you @adonath 

> Instead of specifying the total number of bins, one could possibly choose the bins per decade like

you want me to change to bins per decade or is it fine the value I put?

adonath:
@cosimoNigro Maybe change to `bins_per_decade`? I think we haven't shown this option much yet...

cosimoNigro:
Added bins per decade and clarified the initial description on the utility of the 1D method. Let me know if you are satisfied.

adonath (approved):
Thanks @cosimoNigro, no further comments from my side. I'll request one more review by @registerrier...

adonath:
@registerrier To avoid this PR being open for too long, I'll go ahead and merge it today evening. Please review or write on Slack or open a GitHub issue later  if you have remaining comments...

registerrier (approved):
Thanks @cosimoNigro . This looks good. Indeed correcting for containment is a good thing.

My main comment would be that it might be the occasion to show iterative fitting with comparison of first a PL fit and then an ECPL or log-parabola fit comparing the total stat.

adonath:
I agree with @registerrier here one could add a short section computing the TS difference between both models. So far we haven't shown this anywhere I think. I'll still merge this PR...if you like @cosimoNigro, you can add the model comparison in a follow up PR...

PR #3239, opened 2021-02-15
Allow to choose HDU name in Geom.to_bands_hdu 

registerrier:
**Description**

This pull request tries to correct the behavior of `Geom.to_bands_hdu`.  See issue #3240 .

Currently the `hdu` parameter is unused. This results in the impossibility to properly define the `hdu_bands` to use for `Map.to_hdulist`. 
Here the `hdu` parameter is removed and `hdu_skymap` is replaced with an `extname` keyword.

`Map.to_hdulist` sets the default to `HDU_BANDS` if input `hdu="HDU"` and `hdu_bands = None`. 

**Dear reviewer**

This approach might not be the most relevant. One should be able to fully set the band hdu name. 
The PR is open for discussion.

### DIFF
--- a/gammapy/maps/geom.py
+++ b/gammapy/maps/geom.py
@@ -580,8 +580,8 @@ class MapAxes(Sequence):
         elif format == "fgst-template":
             hdu = "ENERGIES"
         elif format == "gadf" or format is None:
-            if extname:
-                hdu = extname
+            if prefix:
+                hdu = prefix
             else:
                 hdu = "BANDS"
         else:
@@ -2130,8 +2130,8 @@ class Geom(abc.ABC):
 
         return cls.from_header(hdu.header, hdu_bands)
 
-    def to_bands_hdu(self, hdu=None, hdu_skymap=None, format="gadf"):
-        table_hdu = self.axes.to_table_hdu(format=format, prefix=hdu_skymap)
+    def to_bands_hdu(self, hdu_bands=None, format="gadf"):
+        table_hdu = self.axes.to_table_hdu(format=format, prefix=hdu_bands)
         cols = table_hdu.columns.columns
         cols.extend(self._make_bands_cols())
         return fits.BinTableHDU.from_columns(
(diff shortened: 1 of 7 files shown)


adonath:
Thank @registerrier, I must say I much prefer the current behaviour (convention over configuration). I think there is no-good reason why the name of the bands hdu should be chosen individually on write. It will only lead to a variety of FITS file, with different names for HDU bands. So I would even say, let's be strict and completely remove the `hdu_bands` argument for writing.

registerrier:
I think we need this flexibility to be able to share the `hdu_bands` between several `Map` hdus. The hard coding the bands EXTNAME to the map extension name makes this impossible.

Note that there is no convention on the HDUBANDS extension name in the current gadf definition. The diversity of FITS files is there by design.

adonath:
Sorry @registerrier, indeed I forgot about this use-case. 

I just checked once again the spec (https://gamma-astro-data-formats.readthedocs.io/en/latest/skymaps/index.html#bands-hdu) and indeed there is no recommendation on the HDU name, however there is the required `BANDSHDU` keyword for the image / data HDU. As far as I can see, this is correctly used in Gammapy. I guess it has some consequences for #3231, where the corresponding header keywords should point to the bands HDU and region HDU as well. 

So I guess the consequence from this is rather the other way around? So we allow to choose the name of the bands hdu freely, but are more strict on read? Possibly not it's better to not be strict on read...so I guess we keep behaviour proposed in this PR.

The remaining question for me is just which name we choose for the keyword. Here you propose `extname`, while in other places we use `hdu` and `hdu_bands`. I don't care about the actual name but it would be good to keep consistency. I can check and implement this in a follow up PR as well.

adonath:
@registerrier What is the status here?

adonath (approved):
Thanks @registerrier, I added one commit to rename `prefix` in `Axes.to_table_hdu()`, otherwise looks fine...

PR #3241, opened 2021-02-16
Update 4FGL catalogue to DR2

QRemy:
This PR update 4FGL catalogue object to use data release 2 file (see PR#7 in gammapy-data).

### DIFF
--- a/gammapy/catalog/fermi.py
+++ b/gammapy/catalog/fermi.py
@@ -462,7 +462,6 @@ class SourceCatalogObject4FGL(SourceCatalogObjectFermiBase):
         values = self.data[prefix]
         return u.Quantity(values, unit)
 
-    @property
     def lightcurve(self, interval="1-year"):
         """Lightcurve (`~gammapy.estimators.LightCurve`).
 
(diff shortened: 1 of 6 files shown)


adonath:
I just merged https://github.com/gammapy/gammapy-data/pull/7, and will restart the builds now...

gammapy/catalog/fermi.py
```diff
@@ -463,13 +463,31 @@ def _get_flux_values(self, prefix, unit="cm-2 s-1"):
         return u.Quantity(values, unit)
 
     @property
-    def lightcurve(self):
-        """Lightcurve (`~gammapy.estimators.LightCurve`)."""
-        flux = self.data["Flux_History"]
+    def lightcurve(self, interval="1-year"):
```
adonath, line 466: Please still extend the current tests to check for both options...

gammapy/catalog/fermi.py
```diff
@@ -482,8 +500,11 @@ def lightcurve(self):
         # equally-spaced time intervals. This is roughly correct,
         # for plotting the difference doesn't matter, only for analysis
         time_start = Time("2008-08-04T15:43:36.0000")
-        time_end = Time("2016-08-02T05:44:11.9999")
         n_points = len(flux)
+        if n_points in [8, 48]:
```
adonath, line 504: Is there maybe a clearer way to distinguish this case? If not maybe add a comment what the 8 and 48 means...

adonath (changes requested):
Thanks @QRemy, I have left two minor comments...

adonath:
Thanks a lot @QRemy. There is one remaining fail in the catalog notebook: https://github.com/gammapy/gammapy/pull/3241/checks?check_run_id=1918114718#step:6:89

Related to this: can you please make the `.lightcurve()` property a method consistently on all Fermi catalog source classes?

adonath (approved):

PR #3243, opened 2021-02-17
Support I/O for RegionGeom in MapDataset

adonath:
**Description**

This pull request implements the support for `RegionGeom` for the `MapDataset` I/O. It is a follow up to #3231, which introduced a serialisation for `RegionNDMap` with arbitrary extra axes. So far there is only the use case, where e.g. a single PSF is stored in a `RegionNDMap`, which is e.g. the case for a Fermi dataset. See https://github.com/gammapy/gammapy-benchmarks/pull/95#issuecomment-774014871.

### DIFF
--- a/gammapy/datasets/map.py
+++ b/gammapy/datasets/map.py
@@ -940,7 +940,7 @@ class MapDataset(Dataset):
         return hdulist
 
     @classmethod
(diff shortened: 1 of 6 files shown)


adonath:
This blocks the progress on #3216. Will go ahead an merge now...

PR #3244, opened 2021-02-18
Fix evaluation radius for DiskSpatialModel and Generalized GaussianSpatialModel

QRemy:
This PR fix the `evaluation_radius` for `DiskSpatialModel `and `GeneralizedGaussianSpatialModel` in order to prevent cuts of the models (at small radius for disk and large radius for gen-gauss).

The effect of these changes are shown [here](https://github.com/gammapy/gammapy/pull/3229#issuecomment-781318932).


adonath (approved):
Thanks @QRemy, no comments from my side...

Issue #3175, opened 2021-01-07
Retrieve BackgroundMaker status in the data reduction loop


registerrier:
Currently, the various `BackgroundMaker` silently return a possibly incorrect `FoVBackgroundModel` or an empty `counts_off` when no regions where found. 

This will lead to improper behavior in many cases. 

- We first have to increase the log level to warning in these cases. I will change the code accordingly.
- We should also adapt the behavior to either prevent the user using these datasets by returning `None` or by adding a `.success` attribute on the `BackgroundMaker` to let the user proceed one way or the other.

Any opinion @adonath @luca-giunti @AtreyeeS ?

adonath:
I think the appropriate way is to store this info in the meta data or provenance (which we don't really have yet...) somewhere. 
- Raising an error is a bad idea, because the data reduction will fail and interrupt.
- As you said just giving the logging message  is not an ideal solution either, because it cannot really be processed computationally (it can, but only with high effort, such as raising the warning and catching again etc.). 
- I think a status flag on the `Maker` object is non-ideal as well, as the info will be lost after the next pass of the maker. 
- So I think ideally the info is stored in the meta or provenance

As long as we don't have the provenance info we could get a more safe behaviour by setting the `mask_safe` to `False` everywhere, so that the dataset is ignored by default in any likelihood computation and stacking.

registerrier:
As you suggest Axel, I propose to mask the dataset in case of failure as first step. 
This also means we need to change the `_is_stackable()` method so that it does not consider datasets that are fully excluded.

adonath:
The preliminary fixes have beed implemented in #3247 and #3259. A better provenance / history solution remains for after v1.0.

PR #3247, opened 2021-02-19
Change mask_safe to False, when bkg maker fails

registerrier:
**Description**

This pull request proposes a solution to #3175 by making the `mask_safe.data = False` in case the background model fails.

**Dear reviewer**


gammapy/datasets/map.py
```diff
@@ -637,11 +637,17 @@ def stack(self, other):
         if self.exposure and other.exposure:
             self.exposure.stack(other.exposure, weights=other.mask_safe_image)
             # TODO: check whether this can be improved e.g. handling this in GTI
+            livetime=0*u.s
+            if "livetime" in self.exposure.meta:
```
adonath, line 641: To simplify the stacking logic and improve the performance, the mask handling of "self" is in `.to_masked()`. The standard use case I think is to stack into an empty dataset and there is no point in re-applying the mask to self for each dataset, that is stacked into it. So the `.stack()` method only needs to handle `other.mask_safe`.

adonath, line 641: To simplify the stacking logic and improve the performance, the mask handling of "self" is in `.to_masked()`. The standard use case I think is to stack into an empty dataset and there is no point in re-applying the mask to self for each dataset, that is stacked into it. So the `.stack()` method only needs to handle `other.mask_safe`. As `.stack()` is called in `MapDataset.to_masked()`, there is no extra handling need for "self"...

PR #3249, opened 2021-02-22
Add caplog for safe mask maker tests

Jaleleddine:
This pull request adds caplog to test_safe


adonath (approved):
Thanks @Jaleleddine, I wonder whether we should emit this warning at all, but this is independent of the testing...I'll merge now.

PR #3252, opened 2021-02-26
Correct fractional excess variance formula according to reference

JohannesBuchner:
Looking at the [reference](https://arxiv.org/pdf/astro-ph/0307420.pdf), in particular equation 11, it should be σ²err / xbar², not (σ²err / xbar)².

This pull request fixes the code to follow the equation.

### DIFF
--- a/gammapy/stats/variability.py
+++ b/gammapy/stats/variability.py
@@ -45,7 +45,7 @@ def compute_fvar(flux, flux_err):
     sig_square = np.nansum(flux_err ** 2) / n_points
     fvar = np.sqrt(np.abs(s_square - sig_square)) / flux_mean
 
-    sigxserr_a = np.sqrt(2 / n_points) * (sig_square / flux_mean) ** 2
+    sigxserr_a = np.sqrt(2 / n_points) * sig_square / flux_mean ** 2
     sigxserr_b = np.sqrt(sig_square / n_points) * (2 * fvar / flux_mean)
     sigxserr = np.sqrt(sigxserr_a ** 2 + sigxserr_b ** 2)
     fvar_err = sigxserr / (2 * fvar)


adonath (changes requested):
Thanks @JohannesBuchner! I noticed our tests are not sensitive enough to catch this change, so we should lower the test tolerance (`rtol`) to e.g. 1e-3 here: https://github.com/gammapy/gammapy/blob/master/gammapy/stats/tests/test_variability.py#L38 and adapt the reference value. Do you know how to do this?

JohannesBuchner:
To test the correctness of the formula, you need a bright light curve and a faint light curve.

adonath:
@JohannesBuchner I agree this would be the perfect solution to check the formula in both case. However adapting the test tolerance and reference value is the minimal thing to do here...let me know what you prefer.

JohannesBuchner:
Feel free to go ahead. The comment there is confusing me.

adonath:
@JohannesBuchner All functionality in Gammapy is tested. So we make sure executing a method gives a certain result and in case the implementation of the method changes (just as in this PR), we should notice that the reference value in the test changes as well. In this case it doesn't, because the relative change in the reference value is <1e-2. This is non-ideal and I would proposed to change it...I'll do a little follow up commit to fix this.

adonath:
See https://github.com/gammapy/gammapy/commit/2ba98cba50c9ea692fbd1996a0704695cb595d37.

JohannesBuchner:
Yes, but shouldn't it say "the tolerance is very _high_ in the next assert"?

adonath:
@JohannesBuchner Yes, indeed I guess it was meant the precision of the test is very low instead...

PR #3253, opened 2021-02-26
Automate release process with Github actions

Bultako:
This PR adds a remote dispatch Github action that automates a set of steps involved in the release process. Other complementary Github actions have been implemented as *listeners* in the `gammapy-docs` and `gammapy-web` repositories. The set of steps that are automated are the following:

In the gammapy repo:

- Edit docs/index.rst and change the version numbers in the text.

In the gammapy-docs repo:

- Build an updated version of the dev docs using the manual Github action.
- Copy the docs/dev folder as a new docs/0.14 folder.
- In the 0.14/docs/_downloads folder, rename notebooks-dev.tar file as notebooks-0.14.tar.
- Edit stable/index.html to point to 0.14/index.html.

In the gammapy-web repo:

- Mention the release on the news page.
- In the download/install folder, copy gammapy-0.13-environment.yml file as gammapy-0.14-environment.yml

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -13,10 +13,10 @@ jobs:
       - name: modif docs
         run: |
           release="${{ github.event.release.tag_name }}"
-          sed -i -E "s/gammapy-.*-environment.yml/gammapy-$release-environment.yml/" docs/install/index.rst
-          sed -i -E "s/called\s``gammapy-.*``/called ``gammapy-$release``/" docs/install/index.rst
-          sed -i -E "s/activate\sgammapy-.*/activate gammapy-$release/" docs/install/index.rst
-          sed -i -E "s/--release\s.*/--release $release/" docs/install/index.rst
+          sed -i -E "s/gammapy-.*-environment.yml/gammapy-$release-environment.yml/" docs/getting-started/iquickstartndex.rst
+          sed -i -E "s/called\s``gammapy-.*``/called ``gammapy-$release``/" docs/getting-started/quickstart.rst
+          sed -i -E "s/activate\sgammapy-.*/activate gammapy-$release/" docs/getting-started/quickstart.rst
+          sed -i -E "s/--release\s.*/--release $release/" docs/getting-started/quickstart.rst
       - name: dispatch web
         run: |
           curl -XPOST -u "${{ secrets.REMOTE_DISPATCH }}" \
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,37 @@
+name: release
+on:
+  release:
+    types: [published]
+jobs:
+  process:
+    runs-on: ubuntu-latest
+    steps:
+      - name: checkout repo
+        uses: actions/checkout@v2
+        with:
+          ref: master
+      - name: modif docs
+        run: |
+          release="${{ github.event.release.tag_name }}"
+          sed -i -E "s/gammapy-.*-environment.yml/gammapy-$release-environment.yml/" docs/install/index.rst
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,37 @@
...
+  process:
+    runs-on: ubuntu-latest
+    steps:
+      - name: checkout repo
+        uses: actions/checkout@v2
+        with:
+          ref: master
+      - name: modif docs
+        run: |
+          release="${{ github.event.release.tag_name }}"
```
adonath, line 15: Here I'm not sure whether this makes sense: the modification of the docs should happen before defining the tag and not after? So I think this cannot be automatised except maybe with an additional `make release` command or similar, that can be executed by the person taking the release...

adonath (changes requested):
Thanks @Bultako, I have left one comment, I'll checkout the "listeners" in the other repos now. Do you know if there is a ways to test this first?

.github/workflows/release.yml
Bultako: This step modifies the source doc file `index.rst` before dispatching the action to `gammapy-docs` to build the docs. The *release tagging* for the docs is done on `gammapy-docs` repo, which places the sphinx-built docs in its respective version release folder. We could certainly skip this step, to have the good source doc file `index.rst` living in release tag of the code repo, which implies having an additional non-critical manual step and believe that it is going to be properly done. 

I understand that the release tagging is done manually in Github, this triggers the Github action to automates the steps. Another approach could be having a `make release` command that do a POST request to Github API to tag release. In this case, we could make the modifs in `index.rst` as well as sending the two `curls -XPOST` of the dispatched actions with a script called by the `make release` command. I think in this approach we should have to consider dealing with permissions to tag/make a release via the Github API.

Bultako:
> Do you know if there is a ways to test this first?

@adonath 
We could tag a *fake* release and then remove the tag.

.github/workflows/release.yml
adonath: Is there maybe a solution, so that the docs do not mention a specific release at all? I just checked and no other package that is like Gammapy mentions a specific release in the install instructions (I checked e.g. Numpy, Astropy or Scipy). Maybe we keep the normal installation instructions general like `conda install gammapy` or `pip install gammapy` and then we create a dedicated section "Conda Environments", which shows the process using a wildcard for the version?

Bultako: I'd be in favor of simplifying the install process, advertising mainly `conda install gammapy` or `pip install gammapy` We could keep the env yaml files for `gammapy download notebooks` to guarantee reproducibility, and as an alternative less visible in the docs option for install. Moreover, we could also provide Docker images in the Docker Hub.

This said I would not provide commands that will not work when _copy pasted_, so I would not use wildcards.

Bultako: @adonath 
I think we can keep this step, we talked about removing it and do it manually before tagging the release, but I think it does not hurt to keep it. So in case we forgot to do it manually, this part of the text in the release version of the docs will be fixed anyway. And in the case it is done manually, there will be no modifs, since it will replace the text with the same text.

adonath: Yes, that's a fair point...

adonath:
Time is now too short to test this for the v0.19 release. Sorry for that @Bultako, but we will come back to this for v1.0. I'll adapt the milestone now.

registerrier:
A few steps implemented here can be simplified with the PR #4004 and @4005 adapting rest file to add substitution of the release version number in the quick start part and adapting the Makefile.

adonath:
Superseded by #4137.

PR #3258, opened 2021-03-01
Add support for RegionGeom to MapDataset

adonath:
**Description**

This pull request is a proof of concept for the support of `RegionGeom` in the `MapDataset`. This change allows to fully handle a 1D spectral analysis with the `MapDataset` as well. The comparison of performance and results are shown here: https://github.com/adonath/notebooks-public/blob/master/map-dataset-region-geom.ipynb

To achieve comparable performance for the simple circular region / point source case, the `MapEvaluator` was extended to compute the containment of the PSF. The computation of the full PSF kernel and 2D integration, was compuationally to expensive.

In general I think the advantages of this approach are:

- Reduction code duplication
- Containment correction for arbitrary region shapes and spatial models
- Correction of "contamination" by other sources
- 100% API uniformity between 1D and 3D analysis
- Built-in containment correction "on the fly" 

This change in general allows to remove the `SpectrumDataset`, `SpectrumDatasetOnOff` and `SpectrumDatasetMaker` completely and fully unifies the API between 1D, 2D and 3D analysis scenarios. This reduce the future maintenance effort a lot and guarantees technical equivalency of the 1D, 2D and 3D approaches. As one can see from the notebook linked above, the change in API and functionality is minimal. The only major change is how the containment correction works, which was moved from the data reduction step to the fitting / modelling step, which I think is the right place anyway, because it allows the additional treatment of arbitrary region shapes and correction for morphology and contamination by other sources. 

ToDo:
- [x] Add unit tests


adonath:
I updated the notebook to include an illustration of the predicted counts in a given region using a `RegionGeom` and a 3D `WcsGeom`. I find that the predicted counts agree with ~1-2%.

adonath:
This requires a bit more documentation and unit test however the core of the implementation is there and tested. I'll go ahead and merge this and then it can be improved in follow up PRs.

Issue #3259, opened 2021-03-02
ReflectedRegionsBackgroundMaker fails when no OFF are found


registerrier:
**Gammapy version**
v0.18.2 and dev

**Bug description**
When no reflected regions are found `ReflectedRegionsBackgroundMaker.run()` fails with `ZeroDivisionError: division by zero` because of division by 0 in `SpectrumDatasetOnOff.from_spectrum_dataset()`. This is due to the division of floats rather than numpy arrays (when no off regions are found `dataset.acceptance_off` is set to 0.)

Note that this appears only if the `SpectrumDataset` already contains a `background` attribute, e.g. if the option "background" is passed to the `SpectrumDatasetMaker`.

**Expected behavior**
A `SpectrumDatasetOnOff` should be produced with no counts_off and acceptance_off set to 0.

registerrier:
The issue is connected with #3175

adonath:
Thanks @registerrier, I addition we could do the same as proposed in #3247 and also set the `mask_safe` to all false.

For the case where background = None, we already have a unit test: https://github.com/gammapy/gammapy/blob/master/gammapy/makers/background/tests/test_reflected.py#L190

PR #3260, opened 2021-03-02
Correct from_spectrum_dataset bug

registerrier:
**Description**

This pull request proposes a correction for issue #3259 

**Dear reviewer**
The proposed approach is simply to catch the ZeroDivision Error. It might be more rational to modify the background maker behavior so that it directly returns `RegionMap` for `acceptance` and `acceptance_off`.

### DIFF
--- a/gammapy/analysis/core.py
+++ b/gammapy/analysis/core.py
@@ -380,6 +380,8 @@ class Analysis:
             stacked = self.datasets.stack_reduce(name="stacked")
             self.datasets = Datasets([stacked])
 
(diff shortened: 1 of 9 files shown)


adonath:
@registerrier For consistency I decide to remove the possibly to instantiate a `DatasetOnOff` with numbers for `acceptance` and `acceptance_off` and require a `Map` instead. Maybe it can be re-introduced later with a little convenience method like `.from_alpha(alpha=)` or similar...

PR #3268, opened 2021-03-09
Add CompoundSpatialModel

AtreyeeS:
**Description**

This pull request adds a `CompoundSpatialModel`.
As discussed in a previous dev call, this can be useful in dealing with the HGPS 3gauss and 2gauss models.

**Dear reviewer**

One issue here is the behaviour to `.to_region`. A natural option might be to return `CompoundSkyRegion`, but the `CompoundSkyRegion.as_artist()` method is Not implemented yet (https://astropy-regions.readthedocs.io/en/latest/api/regions.CompoundSkyRegion.html#regions.CompoundSkyRegion.as_artist)

This raises issues with plotting.
@adonath what do you suggest?

### DIFF
--- a/gammapy/modeling/models/spatial.py
+++ b/gammapy/modeling/models/spatial.py
@@ -14,6 +14,7 @@ from regions import (
     EllipseSkyRegion,
     PointSkyRegion,
     PolygonSkyRegion,
+    CompoundSkyRegion,
 )
 from gammapy.maps import Map, WcsGeom
 from gammapy.modeling import Parameter
@@ -975,3 +976,61 @@ class TemplateSpatialModel(SpatialModel):
         if geom is None:
             geom = self.map.geom
         super().plot_interative(ax=ax, geom=geom, **kwargs)
+
+
+class CompoundSpatialModel(SpatialModel):
+    """Arithmetic combination of any two spatial models.
+
+    Used to combine two spatial models sharing the same spectral model.
+
+    Works analogously to the :ref:`compound-spectral-model`.
+    """
+
+    tag = ["CompoundSpatialModel", "compound"]
+
+    def __init__(self, model1, model2, operator):
+        self.model1 = model1
+        self.model2 = model2
+        self.operator = operator
+        super().__init__()
+
+    @property
+    def parameters(self):
+        return self.model1.parameters + self.model2.parameters
+
+    def __str__(self):
+        return (
+            f"{self.__class__.__name__}\n"
+            f"    Component 1 : {self.model1}\n"
+            f"    Component 2 : {self.model2}\n"
+            f"    Operator : {self.operator.__name__}\n"
+        )
+
+    def __call__(self, lon, lat):
+        val1 = self.model1(lon, lat)
+        val2 = self.model2(lon, lat)
+        return self.operator(val1, val2)
+
+    def to_dict(self, full_output=False):
+        return {
+            "type": self.tag[0],
+            "model1": self.model1.to_dict(full_output),
+            "model2": self.model2.to_dict(full_output),
+            "operator": self.operator.__name__,
+        }
+
+    @classmethod
+    def from_dict(cls, data):
+        from gammapy.modeling.models import SPATIAL_MODEL_REGISTRY
+
+        model1_cls = SPATIAL_MODEL_REGISTRY.get_cls(data["model1"]["type"])
+        model1 = model1_cls.from_dict(data["model1"])
+        model2_cls = SPATIAL_MODEL_REGISTRY.get_cls(data["model2"]["type"])
+        model2 = model2_cls.from_dict(data["model2"])
+        op = getattr(operator, data["operator"])
+        return cls(model1, model2, op)
+
+    def to_region(self):
+        reg1 = self.model1.to_region()
+        reg2 = self.model2.to_region()
+        return CompoundSkyRegion(reg1, reg2, self.operator)
--- a/gammapy/modeling/models/tests/test_spatial.py
+++ b/gammapy/modeling/models/tests/test_spatial.py
@@ -23,6 +23,7 @@ from gammapy.modeling.models import (
     ShellSpatialModel,
     Shell2SpatialModel,
     TemplateSpatialModel,
+    CompoundSpatialModel,
 )
 from gammapy.utils.testing import mpl_plot_check, requires_data, requires_dependency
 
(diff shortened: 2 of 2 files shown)


QRemy:
Thanks @AtreyeeS
In general I think that models and parameters links can replace compounds models and it would be simpler for complex cases.
For example there is HGPS sources with more than 2 sub-structures, so in that case you would have compounds of compounds... while we already solve this with multiple models sharing the same spectral model. The alternative is to allow a list of models for all `CompoundModel` instead of only two models.

AtreyeeS:
@registerrier : The subclass `CompoundSpectralModel`will then need to inherit from both `CompoundModel` and `SpectralModel` - that is a bit complicated, no?

@QRemy : I agree that most science cases can be performed with a list of Models, but it is very annoying to use the HGPSCatalog where some entries return `Model` and some return `Models` - it should be consistent.
We can have an option `to_compound = True`, and then for users who know what they are doing, they can set it to false and perform accordingly...

adonath:
Thanks a lot @AtreyeeS for this draft! In general I fully agree this is a desirable feature and simplifies bookkeeping of model components. Especially I agree on the consistency of the model object returned by HGPS catalog class. However in detail this requires some more thought:

Now that I look at it (I can't remember whether we discussed this in the dev call already...), I realise the `CompoundSpatialModel` does not solve the problem for the HGPS morphology models, because each component requires a different relative norm / weight. Currently spatial models are all normalised to integrate to unity and this would make the component equally "bright" in the compound model.

So how to solve this? I see a few different options, that we can possibly discuss:
- Introduce a `MultiGaussianSpatialModel` with arbitrary number of components, which allows for these additional norms per component.
- Introduce a more general `SpatialModels` as @QRemy suggested, however in this case it's not 100% clear to me how to introduce the separate norm parameters...
- Introduce a `HGPSMultiGaussianSpatialModel`, with 3 Gaussian components and pre-define the parameters, as for the HGPS the assumption was a symmetrical Gaussian 
- Just convert the morphology to a `TemplateSpatialModel` beforehand, user get access to the individual components by accessing `source.componenets` anyway I think...
- Independently of this we could also add  `SourceCatalog.to_models()`, which takes care of the special case for the HGPS for now.

adonath:
In the list above I forgot your proposal for the `CompoundSpatialModel`. So to make it this completely useful, I think we would need to introduce fixed `norm` parameters for all our spatial models. Which might be an option as well...

AtreyeeS:
Thanks a lot @adonath 
> Now that I look at it (I can't remember whether we discussed this in the dev call already...), I realise the `CompoundSpatialModel` does not solve the problem for the HGPS morphology models, because each component requires a different relative norm / weight. Currently spatial models are all normalised to integrate to unity and this would make the component equally "bright" in the compound model.

Yes, I missed the issue of the normalisation totally.

> So how to solve this? I see a few different options, that we can possibly discuss:
> 
> * Introduce a `MultiGaussianSpatialModel` with arbitrary number of components, which allows for these additional norms per component.

So this will be the only spatial component with an "amplitude normalisation"? that sounds a bit risky to me...


> * Introduce a `HGPSMultiGaussianSpatialModel`, with 3 Gaussian components and pre-define the parameters, as for the HGPS the assumption was a symmetrical Gaussian
I like this solution the best - because otherwise, I see no pressing cause to not used tied parameters and use this compound model thing. What about sources modelled with 2 gaussian? the third gaussian parameter will be zero?
What do you propose for the `to_region` behaviour? 

> * Just convert the morphology to a `TemplateSpatialModel` beforehand, user get access to the individual components by accessing `source.componenets` anyway I think...

Personally, I would not prefer to go from analytic models to a template model...

> * Independently of this we could also add  `SourceCatalog.to_models()`, which takes care of the special case for the HGPS for now.

How will `to_models()` tackle this issue for the HGPS? Some sources will still come with `Models` and some with `Model`, no?

adonath:
> How will to_models() tackle this issue for the HGPS? Some sources will still come with Models and some with Model, no?

Yes, I agree it's not a solution, but just a work-around. One would return the individual components as different `SkyModel` objects, possibly labelled with `source_a`, `source_b` and `source_c` etc. and a linked spectral model by default.

adonath:
> So this will be the only spatial component with an "amplitude normalisation"? that sounds a bit risky to me...

I think one could choose a parametrisation like: amplitude * 1 / ( 1 + norm_1 + norm_2) * (A + norm_1 * B + norm_2 * C), such that the norms are relative to the total amplitude and the spatial model still integrates to unity. This make the norm parameters certainly different from the overall amplitude.

AtreyeeS:
> > So this will be the only spatial component with an "amplitude normalisation"? that sounds a bit risky to me...
> 
> I think one could choose a parametrisation like: amplitude * 1 / ( 1 + norm_1 + norm_2) * (A + norm_1 * B + norm_2 * C), such that the norms are relative to the total amplitude and the spatial model still integrates to unity. This make the norm parameters certainly different from the overall amplitude.

Yes, this is indeed an option. But then, why limit this to `MultiGaussian` and not say, `MultiEllipse` ?
If the primary reason is that the use case comes for the HGPS, then I would still prefer `HGPSMultiGaussianSpatialModel` - just to be explicit

Otherwise, for `CompoundSpatialModel`, maybe we can pass weights as well, and perform a normalisation inside?

registerrier:
> @registerrier Indeed technically this works too, but the bookkeeping is maybe a bit cumbersome, when each model component has a `CompoundSpectralModel`...

I understand this, but we are also making things complex when adding one specific model for each situation that could handled with model composition (e.g. Fermi catalog spectra, HGPS sources and probably lots of other). The bookkeeping is also complex when one has to know that this object is actually the combination of this and this `SkyModel`, one has to introduce specific names for each of them etc.

In the end, I am afraid that to avoid more complexity in one place, we create a lot in others.

I am -1 with the addition of an `HGPSMultiGaussianSpatialModel`, this solves one specific use case with more complexity.

QRemy:
The most general approach would be to have a `Models.to_template()` method that would return a single Model with a `TemplateSpatialModel()` + `NormSpectralModel()`. In addition  we can have a `SourceCatalogue.to_models()` method that will return a Models object so for the case of the HGPS we can add a to_template = False/True option such as:
```

components = self.skymodel()
if isinstance(component, list):
  if to_template:
      components = [components.to_template()]
else:
    components = [components]
models.extend(components)
```

So in one case we have the list of models with each component independent (not linked) and in the other the multicomponent model simplified to only one model.
I can open a PR to implement that.

QRemy:
> A somewhat simpler solution is to have a `ConstantNormSpectralModel` to multiply the shared spectral component for each source.

@registerrier or to avoid using compound we can just have different spectral models with their own amplitude taking into account the different normalisation but a share spectral index

adonath:
@registerrier In general I became somewhat liberal in adding new models, because with the model gallery we have a good way to present and document them. But I also agree that a proliferation of models comes at the cost of users needing to make a choice and understand parameterisations. However I think it cannot fully be avoided, because e.g. already the different parametrisation of the Fermi models were needed to represent the errors correctly.

I support @QRemy 's proposal here, to avoid introducing a new `HGPSMultiGaussianSpatialModel` we could choose the pragmatic solution and use a combination of `TemplateSpatialModel` and `SourceCatalog.to_models()`:
- Modify `SourceCatalogObjectHGPS.sky_model()` to return a `TemplateSpatialModel`
- Implement `SourceCatalog.to_models()`
- Overwrite `SourceCatalogHGPS.to_models(split_components=True)`
- Implement `Models.to_template_spatial_model()`

This achieves an acceptable compromise I think. It has the API uniformity for `SourceCatalogObjectHGPS.sky_model()`, it allows to get the components as individual sources for fitting and even adds new useful convenience like `.to_models()` and `Models.to_template_spatial_model()`

AtreyeeS:
Since we decided against `CompoundModels` in today's dev call, closing this issue

PR #3270, opened 2021-03-10
Minor changes in spatial models docstring

QRemy:
As suggested in the documentation issue, this PR adds details on the nature of the model outline given by .to_region(). It also adds minor comment to evaluation_region() docstring and add docstring for .evaluatiom_geom().

### DIFF
--- a/gammapy/modeling/models/spatial.py
+++ b/gammapy/modeling/models/spatial.py
@@ -450,13 +450,15 @@ class GaussianSpatialModel(SpatialModel):
         exponent = -0.5 * ((1 - np.cos(sep)) / a)
         return u.Quantity(norm * np.exp(exponent).value, "sr-1", copy=False)
 
-    def to_region(self, n_sigma=1, **kwargs):
+    def to_region(self, x_sigma=1.5, **kwargs):
         """Model outline at a given number of :math:`\sigma`.
         
         Parameters
         ----------
-        n_sigma : float
-            Number of :math:`\sigma` (Default is 1). 
+        x_sigma : float
+            Number of :math:`\sigma`
+            Default is :math:`1.5\sigma` which corresonds to about 68%
+            containment for a 2D symetric Gaussian. 
 
         Returns
         -------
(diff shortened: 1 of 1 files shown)


AtreyeeS (dismissed):
Thanks @QRemy !This looks perfect to me.

I was wondering if for the Gaussian models, it would be useful to have an optional `nsigma` argument, which can be `1` by default.  This makes it easier to compare eg, with other models.

adonath:
Thanks @QRemy , no comments from my side. I support @AtreyeeS 's proposal to add the `n_sigma` parameter to the `.to_region()` method. For a 2D Gaussian a 1 sigma radius only corresponds to 38% containment , so often the source appears visually too small. So it might be good to have some flexibility on this...

AtreyeeS (approved):
Thank @QRemy 
I have no further comments

gammapy/modeling/models/spatial.py
```diff
@@ -439,22 +450,34 @@ def evaluate(lon, lat, lon_0, lat_0, sigma, e, phi):
...
+            width=2 * n_sigma * minor_axis,
             angle=self.phi.quantity,
             **kwargs,
         )
 
     @property
     def evaluation_region(self):
-        """Evaluation region"""
+        """Evaluation region consistent with evaluation radius """
         region = self.to_region()
```
adonath, line 479: To really make sure that it's consistent with the `evaluation_radius`, maybe refactor to `return self.to_region(self.evaluation_radius / self.sigma.quantity)`?

gammapy/modeling/models/spatial.py
```diff
@@ -439,22 +450,34 @@ def evaluate(lon, lat, lon_0, lat_0, sigma, e, phi):
...
         return u.Quantity(norm * np.exp(exponent).value, "sr-1", copy=False)
 
-    def to_region(self, **kwargs):
-        """Model outline (`~regions.EllipseSkyRegion`)."""
+    def to_region(self, n_sigma=1, **kwargs):
+        """Model outline at a given number of :math:`\sigma`.
+        
+        Parameters
+        ----------
+        n_sigma : float
```
adonath, line 458: This is nitpicky: `n_sigma` suggests it would be an integer and not a float, maybe there is a different name...

adonath (commented):
Thanks @QRemy, I have left two more inline comments.

adonath (approved):
Thanks @QRemy. no further comments from my side...

PR #3272, opened 2021-03-12
Add SourceCatalog.to_models() and improve models selection from catalogues

QRemy:
This PR implements several of the proposals discussed in #3268
- add option components_status : {'independent', 'linked', 'merged'} to SourceCatalogObjectHGPS.sky_model():
-- Default is "independent" so it preserves the current behavior. 
-- "linked" adds spectral parameters links (except for the normalisations) .
--  "merged" creates a single model given as a TemplateSpatialModel with a PowerLawNormSpectralModel
- Implement Models.to_template_spatial_model()
- implement SourceCatalog.to_models()
- add support for boolean array on getitem of SourceCatalog so one can use subcat = cat[mask]

### DIFF
--- a/gammapy/catalog/hess.py
+++ b/gammapy/catalog/hess.py
@@ -512,34 +512,54 @@ class SourceCatalogObjectHGPS(SourceCatalogObject):
             model.parameters[name].error = value
         return model
 
-    def sky_model(self, which="best"):
+    def sky_model(self, which="best", components_status="independent"):
         """Source sky model.
 
         Parameters
         ----------
         which : {'best', 'pl', 'ecpl'}
             Which spectral model
-
+            
+        components_status : {'independent', 'linked', 'merged'}
+            Relation between the sources components:
+                'independent' : each sub-component of a source is given as 
+                                a diffrent `SkyModel` (Default)
+                'linked' : each sub-component of a source is given as 
+                           a diffrent `SkyModel` but the spectral parameters
+                           except the mormalisation are linked.
+                'merged' : the sub-components are merged into a single `SkyModel`
+                           given as a `~gammapy.modeling.models.TemplateSpatialModel`
+                           with a `~gammapy.modeling.models.PowerLawNormSpectralModel`.
+                           In that case the relave weigths between the components
+                           cannot be adjusted.
         Returns
         -------
-        sky_model : `~gammapy.modeling.models.SkyModel`
-            Sky model of the catalog object.
+        sky_model : `~gammapy.modeling.models.Models`
+           Models of the catalog object.
         """
+
         spatial_type = self.data["Spatial_Model"]
         if spatial_type in {"2-Gaussian", "3-Gaussian"}:
             models = []
+            spectral_model = self.spectral_model(which=which)
             for component in self.components:
-                spectral_model = self.spectral_model(which=which)
+                spec_component = spectral_model.copy()
                 weight = component.data["Flux_Map"] / self.data["Flux_Map"]
-                spectral_model.parameters["amplitude"].value *= weight
+                spec_component.parameters["amplitude"].value *= weight
+                if components_status == 'linked':
+                    for name in spec_component.parameters.names: 
+                         if name not in ["norm", "amplitude"]:
+                             spec_component.__dict__[name] = spectral_model.parameters[name]
                 model = SkyModel(
                     spatial_model=component.spatial_model(),
-                    spectral_model=spectral_model,
+                    spectral_model=spec_component,
                     name=component.name,
                 )
                 models.append(model)
-
-            return Models(models)
+            if components_status == "merged":
+                return models.to_template_spatial_model()
+            else:
+                return Models(models)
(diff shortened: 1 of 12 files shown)


gammapy/catalog/hess.py
```diff
@@ -512,41 +514,87 @@ def spatial_model(self):
...
         else:
             return SkyModel(
                 spatial_model=self.spatial_model(),
                 spectral_model=self.spectral_model(which=which),
                 name=self.name,
             )
 
+    @staticmethod
+    def _get_components_geom(models):
+        energy_axis = MapAxis.from_nodes(
```
adonath, line 579: I think you can get the energy range directly from `.energy_range()`, but maybe one could also choose a constant wider energy range. In the latter case maybe just choose something simple like 100 GeV - 100 TeV. You can also use `MapAxis.from_energy_bounds("100 GeV", "100 TeV", nbin=10, per_decade=True, name="energy_true")`

gammapy/catalog/hess.py
```diff
@@ -512,41 +514,87 @@ def spatial_model(self):
...
+            10 ** np.arange(-0.95, 1.951, 0.1),
+            interp="log",
+            name="energy_true",
+            unit="TeV",
+        )
+        region = list_to_compound_region([m.spatial_model.to_region() for m in models])
```
adonath, line 585: I think this can be reduced to:
```python3
regions = [c.spatial_model().evaluation_region for c in source.components]
geom = RegionGeom.from_regions(regions, binsz_wcs="0.02 deg")
wcs = geom.to_wcs_geom()
```

gammapy/catalog/hess.py
```diff
@@ -693,6 +741,41 @@ def gaussian_component(self, row_idx):
...
+
+        Returns
+        -------
+        models : `~gammapy.modeling.models.Models`
+            Models of the catalog.
+        """
+
+        models = []
+        for _ in self:
+            m = _.sky_model(which, components_status)
```
adonath, line 773: It's always the better pattern to pass arguments by name and not rely on the order, because the order might change when a function is refactored in future. So I recommend to use `.sky_model(which=which, components_status=component_status)`

gammapy/catalog/hess.py
```diff
@@ -512,41 +514,87 @@ def spatial_model(self):
             model.parameters[name].error = value
         return model
 
-    def sky_model(self, which="best"):
+    def sky_model(self, which="best", components_status="independent"):
```
adonath, line 517: Here I'm not a 100% whether to expose the options in `.sky_model()`. Ideally we could achieve uniformity in the returned types. I could imagine e.g. to introduce `SourceCatalogObjectHGPS.to_models(link_components=)`  or `components_models()` as well. `.sky_model()`then reduces to `.to_models().to_template_spatial_model()` an always returns a `SkyModel`

gammapy/modeling/models/core.py
```diff
@@ -791,6 +791,30 @@ def reassign(self, dataset_name, new_dataset_name):
...
+        ----------
+        spectral_model : `~gammapy.modeling.models.SpectralModel`
+            One of the NormSpectralMdel 
+        name : str
+            Name of the new model
+                           
+        """
+        from . import SkyModel, TemplateSpatialModel, PowerLawNormSpectralModel
+
+        unit = u.Unit("1 / (cm2 s sr TeV)")
```
adonath, line 807: Here you could simplify by just evaluation the first component separately, like:
```python3
map_ = self[0].evaluate_geom()

for m in self[1:]:
    map_ += m.evaluate_geom()
```
I think it also misses the selection on  the model type, or raising and error if a `FoVBackgroundModel` is contained.

gammapy/modeling/models/core.py
```diff
@@ -791,6 +791,30 @@ def reassign(self, dataset_name, new_dataset_name):
         models = [m.reassign(dataset_name, new_dataset_name) for m in self]
         return self.__class__(models)
 
+    def to_template_spatial_model(self, geom, spectral_model=None, name=None):
```
adonath, line 794: I think here we should take a decision:
- Either name it `.to_template_spatial_model()` and return a `TemplateSpatialModel`
- Or rename to `.to_template_sky_model()`
I think I prefer the first option, because the latter basically duplicates the API of creating a `SkyModel()` by requiring the `spectral_model` and `name` keywords. But I can see the additional convenience, that saves a few lines of code for the standard use case...

adonath (changes requested):
Thanks @QRemy, I have left a few inline comments.

gammapy/modeling/models/core.py
QRemy: `map_ = self[0].evaluate_geom()`
 this return a quantity not a map

adonath: Indeed, you are right. Could be worth changing at some point, as we have it for `.integrate_geom()` as well...

adonath (approved):

PR #3275, opened 2021-03-17
Cleanup FluxMaps a little bit

adonath:
**Description**

This pull request cleans up the `FluxMaps` object a little bit.

**Dear reviewer**
@registerrier Previously the code allowed any arbitrary extra maps for I/O. I wasn't sure whether this is a good idea or not. In this PR I'm introducing a `COMMON_MAPS` variable that declare the allowed extra quantities. In general this possibly includes a few other such as `is_ul` or `niter`.

### DIFF
--- a/gammapy/estimators/flux_map.py
+++ b/gammapy/estimators/flux_map.py
@@ -27,7 +27,8 @@ REQUIRED_MAPS = {
     "likelihood": ["norm"],
 }
 
-#TODO: add an entry for is_ul?
+
+# TODO: add an entry for is_ul?
 OPTIONAL_MAPS = {
     "dnde": ["dnde_err", "dnde_errp", "dnde_errn", "dnde_ul"],
     "e2dnde": ["e2dnde_err", "e2dnde_errp", "e2dnde_errn", "e2dnde_ul"],
@@ -37,6 +38,7 @@ OPTIONAL_MAPS = {
 }
 
 log = logging.getLogger(__name__)
+
 
 class FluxMaps(FluxEstimate):
     """A flux map container.
@@ -75,7 +77,7 @@ class FluxMaps(FluxEstimate):
     def __init__(self, data, reference_model=None, gti=None):
         self.geom = data['norm'].geom
 
-        if reference_model == None:
+        if reference_model is None:
             log.warning("No reference model set for FluxMaps. Assuming point source with E^-2 spectrum.")
             reference_model = self._default_model()
 
@@ -87,7 +89,10 @@ class FluxMaps(FluxEstimate):
 
     @staticmethod
     def _default_model():
-        return SkyModel(spatial_model=PointSpatialModel(), spectral_model=PowerLawSpectralModel(index=2))
+        return SkyModel(
+            spatial_model=PointSpatialModel(),
+            spectral_model=PowerLawSpectralModel(index=2)
+        )
 
     @property
     def _additional_maps(self):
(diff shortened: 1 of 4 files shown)


gammapy/estimators/flux_map.py
```diff
@@ -27,17 +22,29 @@
...
+        "norm_err",
+        "norm_errn",
+        "norm_errp",
+        "norm_ul",
+        "norm_scan",
+        "stat_scan",
+    ],
 }
 
+COMMON_MAPS = ["ts", "sqrt_ts"]
```
registerrier, line 42: How about other maps: `counts`, `background`, `excess`. 
`ExcessMapEstimator` could be adapted to return flux maps as well but we would need other diagnostics maps in that case.
Shouldn't we leave the flexibility of using any other map?

registerrier (commented):
Thanks @adonath.

My main concern is the decision to remove the flexibility to support other maps than some specified ones.
This means each time we adapt some estimator or introduce some functionality we have to add the new maps in the list of supported maps. We will introduce some for incompatibility with that. If there is no strong gain to introduce this limited list of supported map, I would not do that.

gammapy/estimators/flux_map.py
adonath: This was the main question I had as well. In general I would prefer to declare somehow which other maps are allowed as well, but this requires to define first which those are. I think defining and documenting the additional quantities e.g. a `ExcessMapEstiamtor` returns is a good thing. And this requires also clarifying the terms we use. E.g. especially in the context of the `ExcessMapEstimator` this is not 100% clear to me yet. E.g. currently we return `counts` and `excess` and `background`, while in the context of an `Estimator` I would rather refer to `npred`, `npred_signal` and `npred_background`.

As a first start I could open an PR extending the documentation here https://docs.gammapy.org/dev/estimators/index.html, for quantities we are currently missing.

gammapy/estimators/flux_map.py
```diff
@@ -378,64 +331,59 @@ def _validate_type(maps, sed_type):
...
 
-        if sed_type == "dnde":
-            factor = ref_dnde
-        elif sed_type == "flux":
-            factor = reference_model.spectral_model.integral(e_edges[:-1], e_edges[1:])
-        elif sed_type == "eflux":
-            factor = reference_model.spectral_model.energy_flux(e_edges[:-1], e_edges[1:])
-        elif sed_type == "e2dnde":
-            factor = e_ref ** 2 * ref_dnde
+        with np.errstate(invalid="ignore", divide="ignore"):
```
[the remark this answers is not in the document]
adonath: Yes, I guess the alternative solution is to evaluate the reference flux on the `SkyModel` and integrate over the spatial part, but that's probably too much trouble for almost no benefit....I agree that using a cube template as reference model could be just forbidden.

gammapy/estimators/flux_map.py
adonath: Following #3278 , I added `npred`, `npred_excess` and `npred_null` to the maps. However it might still be incomplete. I think the next step is to check the result quantities of the current estimators and make a PR to unify it and extend for missing quantities. I'm completely open to allow for any other additional quantity, however for this PR I propose to just stick with the defined and documented ones.

adonath:
@registerrier I'll go ahead and merge this PR. Then we can prepare a second one to unify and add missing quantities to the current Estimators...

PR #3277, opened 2021-03-18
Change Estimator optional default selection

adonath:
**Description**

This pull request changes the optional default selection for `Estimator` to `None`, too avoid that expensive computation such as ULs and assym. error are computed by default. In selected place we now use `selection_optional="all"` explicitly.


adonath:
There was agreement in the dev call to not compute optional quantities by default. I'll go ahead and merge this now...

PR #3278, opened 2021-03-19
Extend documentation of quantities returned by estimators

adonath:
**Description**

This pull request adds a proposal for the missing quantities returned by `Estimator` classes. Here is the current proposal:

| Quantity        | Description                           |
|--------------|----------------------------------------------------------------------------------|
| npred        | Predicted counts of the best fit hypothesis, equivalent to correlated `counts` for backward folding                                     |
| npred_null   | Predicted counts of the null hypothesis                                          |
| npred_signal | Predicted counts of the signal over `npred_null`, equivalent to (`npred - npred_null`) and correlated `excess` for backward folding |
| stat         | Fit statistics value of the best fit hypothesis                                  |
| stat_null    | Fit statistics value of the null hypothesis                                      |

I think the information is basically complete now. My proposed naming scheme would be `npred` uniformly and not introduce separate  definitions like `counts_corr` and `excess_corr` for the case of backwards folding as done in the `ExcessMapEstimator` and instead just describe the equivalency in the table. I think in context of hypothesis testing this is reasonable.

### DIFF
--- a/docs/estimators/index.rst
+++ b/docs/estimators/index.rst
@@ -19,10 +19,10 @@ energy.
 The core of any estimator algorithm is hypothesis testing: a reference
 model or counts excess is tested against a null hypothesis. From the
 best fit reference model a flux is derived and a corresponding :math:`\sqrt{\Delta TS}`
-value from the difference in fit statistics to the null hypothesis,
-assuming one degree of freedom. In this case
-:math:`\sqrt{\Delta TS}` represents an approximation of the
-"classical significance".
+value from the difference in fit statistics to the null hypothesis.
+Assuming one degree of freedom, :math:`\sqrt{\Delta TS}` represents an
+approximation (`Wilk's theorem <https://en.wikipedia.org/wiki/Wilks%27_theorem>`_)
+of the "classical significance".
 
 In general the flux can be estimated using methods:
 
@@ -47,13 +47,13 @@ Based on this algorithm most estimators compute the same basic quantities:
 ================= =================================================
 Quantity          Definition
 ================= =================================================
-e_ref			  Reference energy
-e_min			  Minimum energy
-e_max			  Maximum energy
-norm			  Best fit norm with respect to the reference spectral model
-norm_err		  Symmetric error on the norm derived from the Hessian matrix
-ts				  Difference in fit statistics (`stat - stat_null` )
-sqrt_ts			  Square root of ts, in case of one degree of freedom, corresponds to significance (Wilk's theorem)
+e_ref             Reference energy
+e_min             Minimum energy
+e_max             Maximum energy
+norm              Best fit norm with respect to the reference spectral model
+norm_err          Symmetric error on the norm derived from the Hessian matrix
+ts                Difference in fit statistics (`stat - stat_null` )
+sqrt_ts           Square root of ts, in case of one degree of freedom, corresponds to significance (Wilk's theorem)
 ================= =================================================
 
 
@@ -62,18 +62,18 @@ In addition the following optional quantities can be computed:
 ================= =================================================
 Quantity          Definition
 ================= =================================================
-norm_errp		  Positive error of the norm
-norm_errn	      Negative error of the norm
-norm_ul			  Upper limit of the norm
-stat			  Fit statistics value of the best fit hypothesis
-stat_null		  Fit statistics value of the null hypothesis
+norm_errp         Positive error of the norm
+norm_errn         Negative error of the norm
+norm_ul           Upper limit of the norm
+stat              Fit statistics value of the best fit hypothesis
+stat_null         Fit statistics value of the null hypothesis
 ================= =================================================
-npred		  	  Predicted counts of the best fit hypothesis
+npred             Predicted counts of the best fit hypothesis
 npred_null        Predicted counts of the null hypothesis
 npred_signal      Predicted counts of the signal over `npred_null`, equivalent to (`npred - npred_null`)
 ================= =================================================
-norm_scan		  Norm scan
-stat_scan		  Fit statistics scan
+norm_scan         Norm scan
+stat_scan         Fit statistics scan
 ================= =================================================
 
 To compute the assymetric errors as well as upper limits one can
@@ -87,10 +87,10 @@ to the following different SED types:
 ================= =================================================
 Quantity          Definition
 ================= =================================================
-dnde 		      Differential flux at ``e_ref``
-flux 			  Integrated flux between ``e_min`` and ``e_max``
-eflux			  Integrated energy flux between ``e_min`` and ``e_max``
-e2dnde			  Differential energy flux between ``e_ref``
+dnde              Differential flux at ``e_ref``
+flux              Integrated flux between ``e_min`` and ``e_max``
+eflux             Integrated energy flux between ``e_min`` and ``e_max``
+e2dnde            Differential energy flux between ``e_ref``
 ================= =================================================
 
 The same can be applied for the error and upper limit information.
--- a/docs/estimators/index.rst
+++ b/docs/estimators/index.rst
@@ -18,11 +18,13 @@ energy.
 
 The core of any estimator algorithm is hypothesis testing: a reference
 model or counts excess is tested against a null hypothesis. From the
-best fit reference model a flux is derived and a corresponding :math:`\sqrt{\Delta TS}`
+best fit reference model a flux is derived and a corresponding :math:`\Delta TS`
 value from the difference in fit statistics to the null hypothesis.
 Assuming one degree of freedom, :math:`\sqrt{\Delta TS}` represents an
 approximation (`Wilk's theorem <https://en.wikipedia.org/wiki/Wilks%27_theorem>`_)
-of the "classical significance".
+of the "classical significance". In case of a negative best fit flux,
+e.g. when the background is overestimated, the significance is defined
+as :math:`-\sqrt{\Delta TS}` by convention.
 
 In general the flux can be estimated using methods:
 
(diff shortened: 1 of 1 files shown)


adonath:
I think one should add a bit more information on the definition of the null hypothesis, but this varies between the estimators, so it's better documented on the `Estimator` classes...

docs/estimators/index.rst
```diff
@@ -64,13 +65,17 @@ Quantity          Definition
 norm_errp		  Positive error of the norm
 norm_errn	      Negative error of the norm
 norm_ul			  Upper limit of the norm
+stat			  Fit statistics value of the best fit hypothesis
+stat_null		  Fit statistics value of the null hypothesis
+================= =================================================
+npred		  	  Predicted counts of the best fit hypothesis
```
AtreyeeS, line 71: In the GitHub rendering, these lines are not coming inside the table, but rather in a chaotic manner.

docs/estimators/index.rst
```diff
@@ -64,13 +65,17 @@ Quantity          Definition
...
+================= =================================================
 norm_scan		  Norm scan
 stat_scan		  Fit statistics scan
-stat			  Fit statistics value of the best fit model
-null_value		  Fit statistics value of the null hypothesis
 ================= =================================================
 
-
 To compute the assymetric errors as well as upper limits one can
 specify the arguments ``n_sigma`` and ``n_sigma_ul``. The ``n_sigma``
```
AtreyeeS, line 80: Maybe explain what `n_sigma_ul` is as well?

AtreyeeS (dismissed):
Thanks @adonath !
These changes look good to me.

Going through the full file, I think we can improve it a bit more:

- Second paragraph: 
> The core of any estimator algorithm is hypothesis testing: a reference model or counts excess is tested against a null hypothesis. From the best fit reference model a flux is derived and a corresponding \sqrt{\Delta TS} value from the difference in fit statistics to the null hypothesis, assuming one degree of freedom. In this case \sqrt{\Delta TS} represents an approximation of the "classical significance".

This seems to imply that the flux or sqrt(del_ts) is valid only for one degree of freedom, but we start with models, which, naively, can have more dof.
Rather, change to - 
"The core of any estimator algorithm is hypothesis testing: a reference model or counts excess is tested against a null hypothesis. From the best fit reference model a flux is derived and a corresponding \sqrt{\Delta TS} value from the difference in fit statistics to the null hypothesis. Assuming one degree of freedom, \sqrt{\Delta TS} represents an approximation of the "classical significance".

- The tutorials subsection link to only the light curve notebooks. Add the other estimators as well?

docs/estimators/index.rst
adonath: Thanks, indeed this was because of mixture of tabs and whitespace. Fixed.

docs/estimators/index.rst
```diff
@@ -64,13 +65,17 @@ Quantity          Definition
 norm_errp		  Positive error of the norm
 norm_errn	      Negative error of the norm
 norm_ul			  Upper limit of the norm
+stat			  Fit statistics value of the best fit hypothesis
+stat_null		  Fit statistics value of the null hypothesis
+================= =================================================
+npred		  	  Predicted counts of the best fit hypothesis
+npred_null        Predicted counts of the null hypothesis
+npred_signal      Predicted counts of the signal over `npred_bull`, equivalent to (`npred - npred_null`)
```
registerrier: I think `signal` is misleading. Because `npred_null` can contain signal (in the sense of astrophysical photons). I still think `excess` is better, since it is really what it is.

adonath:
Thanks @AtreyeeS! Comments are addressed...

docs/estimators/index.rst
```diff
@@ -19,10 +19,10 @@ energy.
...
 model or counts excess is tested against a null hypothesis. From the
 best fit reference model a flux is derived and a corresponding :math:`\sqrt{\Delta TS}`
-value from the difference in fit statistics to the null hypothesis,
-assuming one degree of freedom. In this case
-:math:`\sqrt{\Delta TS}` represents an approximation of the
-"classical significance".
+value from the difference in fit statistics to the null hypothesis.
+Assuming one degree of freedom, :math:`\sqrt{\Delta TS}` represents an
+approximation (`Wilk's theorem <https://en.wikipedia.org/wiki/Wilks%27_theorem>`_)
+of the "classical significance".
```
registerrier, line 25: Is it an approximation of the significance or its definition in our domain?

docs/estimators/index.rst
```diff
@@ -19,10 +19,10 @@ energy.
 The core of any estimator algorithm is hypothesis testing: a reference
 model or counts excess is tested against a null hypothesis. From the
 best fit reference model a flux is derived and a corresponding :math:`\sqrt{\Delta TS}`
-value from the difference in fit statistics to the null hypothesis,
-assuming one degree of freedom. In this case
-:math:`\sqrt{\Delta TS}` represents an approximation of the
-"classical significance".
+value from the difference in fit statistics to the null hypothesis.
+Assuming one degree of freedom, :math:`\sqrt{\Delta TS}` represents an
```
registerrier, line 23: Maybe one should mention explicitly that we use signed `sqrt_ts`

docs/estimators/index.rst
```diff
@@ -47,33 +47,38 @@ Based on this algorithm most estimators compute the same basic quantities:
...
-norm_err		  Symmetric rrror on the norm derived from the Hessian matrix
-ts				  Difference in fit statistics (`stat_sum - null_value` )
-sqrt_ts			  Square root of TS, corresponds to significance (Wilk's theorem)
+e_ref             Reference energy
+e_min             Minimum energy
+e_max             Maximum energy
+norm              Best fit norm with respect to the reference spectral model
+norm_err          Symmetric error on the norm derived from the Hessian matrix
+ts                Difference in fit statistics (`stat - stat_null` )
+sqrt_ts           Square root of ts, in case of one degree of freedom, corresponds to significance (Wilk's theorem)
```
registerrier, line 56: squared root of ts multiplied by `sign(excess)`

docs/estimators/index.rst
```diff
@@ -47,33 +47,38 @@ Based on this algorithm most estimators compute the same basic quantities:
...
-null_value		  Fit statistics value of the null hypothesis
+norm_errp         Positive error of the norm
+norm_errn         Negative error of the norm
+norm_ul           Upper limit of the norm
+stat              Fit statistics value of the best fit hypothesis
+stat_null         Fit statistics value of the null hypothesis
+================= =================================================
+npred             Predicted counts of the best fit hypothesis
+npred_null        Predicted counts of the null hypothesis
+npred_signal      Predicted counts of the signal over `npred_null`, equivalent to (`npred - npred_null`)
```
registerrier, line 73: Because `signal` is ambiguous, I prefer `npred_excess` with the following definition:

`npred_excess = npred - npred_null`  for forward folding estimators

And similarly:
`excess = counts - npred_null` for backward folding estimators.

registerrier (commented):
Thanks @adonath 

My main comment is that `excess` is more explicit than signal. Or one could use excess_signal for more clarity, but it is a bit too long.

For instance the `n_sig` definition in the statistics doc is confusing for many who think it stands for number of sigma.

docs/estimators/index.rst
adonath: Hm, I guess it's both and there is no either / or? Taken strictly it's an approximation according to Wilk's theorem and it's also the widely used definition in our domain...

docs/estimators/index.rst
adonath: Yes, I agree. My motivation was to stay in-line with what we define here: https://docs.gammapy.org/0.18.2/stats/index.html#notations but I now realise this requires some more thought. I just checked and `npred` is also define here https://gamma-astro-data-formats.readthedocs.io/en/latest/spectra/flux_points/index.html#normalization-representation, where it means the predicted counts from the model component only (so `npred - npred_null` in our definition...).

It's a bit similar with excess, because `MapDataset.excess` is already defined as `counts - background` and does not account for any additional source contributions. So that definition of `excess` is somewhat variable, in the sense that it can either include additional source contributions or not. I wonder whether we can find a fully consistent definition. I'll think about in once more and make a new proposal...

adonath: I think we could rename `sig` -> `excess` consistently here: https://docs.gammapy.org/0.18.2/stats/index.html#notations, to get rid of the ambiguity with n sigma. But this would require adapting the API of `MapDataset` and `CountsStatistic` once again (fine from my side...if not now, we will not do it for v1.0...)

adonath: I think the inconsistency between "our" `npred` (which means the total predicted counts...) and the definition in gadf cannot easily be resolved, it would require to either re-define in Gammapy, which I don't want to do, because `MapDataset.npred()` is already and establish part of our API (`.npred_signal()` possibly not that much...) or in gadf.

adonath: What about if we just stay in line with the `MapDataset` and use `npred`, `npred_background` and `npred_excess`, but clearly state, that for estimators by default a "residual convention" like `npred_background = dataset.npred()` is used? In case there are no source models defined it is equivalent anyway. When looking at the map users will see whether sources are included in the background anyway...

adonath: On 2nd thought maybe `npred`, `npred_excess` and `npred_null` is also fine...

adonath: I renamed to `npred_excess`...

adonath:
@registerrier I'll go ahead and merge this now., but we should have follow up discussion on the `n_sig` and `npred_signal()` names...

PR #3280, opened 2021-03-22
Add figures to illustrate background methods

adonath:
**Description**

This pull request adds the two figures from https://www.aanda.org/articles/aa/abs/2018/04/aa32098-17/aa32098-17.html to illustrate the ring and reflected regions background methods. Albeit we have references for the methods I think it's nice to provide a minimal documentation & description of the methods.


registerrier (dismissed):

adonath:
Thanks @registerrier, I added some more info on the ring background and references to the images...

PR #3282, opened 2021-03-22
Add documentation for stack and joint fit

AtreyeeS:
This PR adds a notebook explaining how dataset stacking is implemented in Gammapy and how it differs from a joint analysis.


docs/datasets/stack.ipynb
```diff
@@ -0,0 +1,362 @@
+{
```
registerrier, line 1: <p>I would explicitly state what is going on here.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/gammapy/gammapy/pull/3282/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='3'/>

registerrier, line 1: <p>Maybe this should go as a general introduction for the notebook.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/gammapy/gammapy/pull/3282/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

registerrier, line 1: <p>Here the actual treatment is a bit more complex in 3D. In particular regarding regions fully excluded (generally because of offset cut).</p><p>Maybe an explanation for what is going on in 3D in needed somewhere.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/gammapy/gammapy/pull/3282/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

registerrier, line 1: <p>Maybe add that if all the dataset is masked, exposure and livetime is not increased. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/gammapy/gammapy/pull/3282/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

registerrier, line 1: <p>I think we need to refer to a general page showing the formula. I suppose that irf theory is right place for that. </p><p>That's for another future PR.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/gammapy/gammapy/pull/3282/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='13'/>

adonath:
Thanks @AtreyeeS, I think this is a very illustrative and instructive notebook and we clearly need to improve the documentation on stacking! In this context I'd like to bring up a few general points:
- This would be there first notebook, that is not in the tutorials section. Even if we have the possibility, I'm a bit hesitant to actually use it now, mostly because we currently have the API section in the tutorials gallery.
- Documentation of stacking is currently spread across multiple places: `MapDataset.stack()`, a comparison of joint vs. stacked is already in the `spectrum_analysis.ipynb` and `analysis_3D.ipynb` notebooks and some documentation is here: https://docs.gammapy.org/0.18.2/makers/index.html#stacking-of-datasets. In addition some documentation will be required for `docs/datasets/index.rst`  as well.
- Even when the toy examples are illustrative, I'm not sure we should use them in the Gammapy documentation. Many parts of the code in the notebook are not really relevant for documentation how stacking actually works, but just to create the plots. In this sense the information is a bit hidden between the code. I think for now we should reserve the API docs for the API (obvious...) and some theoretical background.

So my proposal would be as following:
- Add a new "Stacking of Datasets" section here: https://docs.gammapy.org/0.18.2/datasets/index.html
- Move the "theoretical part" (i.e. which quantities are summed, which masks are applied, how is the weighting done) of your notebook there, without code examples. I could imagine adding a table and some formulae like https://docs.gammapy.org/0.15/api/gammapy.irf.IRFStacker.html#irfstacker,  could help. Or a bullets list or a short text paragraph.
- In addition move the content from `MapDataset.stack()` to the section and replace the docs in `MapDataset.stack()` with a link to the docs page.
- Turn your current notebook with the toy example into a Gammapy-recipe: https://github.com/gammapy/gammapy-recipes (I think this is what we proposed at the last dev call for the stacked npred computation as well...)
- We implement a general solution (maybe sphinx directive, @Bultako ?) to link to gammapy recipes from our documentation. This we can use to link to toy-examples, statistical mini studies, etc. from our documentation.

docs/datasets/index.rst
```diff
@@ -52,10 +52,83 @@ Note that in Gammapy, 2D image analyses are done with 3D cubes with a single
...
 
+To analyse multiple runs, you can either stack the datasets together, or perform
+a joint fit across multiple datasets.
 
-Getting Started
-===============
+Stacking Multiple Datasets
+==========================
 
+Stacking datasets implies that the counts, background and reduced IRFs from all the
```
adonath, line 61: One could put some motivation on why stacking is needed at all: reduction of computation effort.

docs/datasets/index.rst
```diff
@@ -52,10 +52,83 @@ Note that in Gammapy, 2D image analyses are done with 3D cubes with a single
...
+runs are binned together to get one final dataset for which a likelihood is
+computed during the fit.
+
+Counts, background and exposure (lying outside the masked regions are simply summed),
```
adonath, line 65: I think closing parentheses are not set correctly here. In general my proposal would be to only use terms like "region defined by the `mask_safe`" or "data range defined by the `mask_safe`" and not mix it with inside / outside exclusion regions (see later in the table as well), because typically exclusion regions are handled independently from the safe data range.

docs/datasets/index.rst
```diff
@@ -52,10 +52,83 @@ Note that in Gammapy, 2D image analyses are done with 3D cubes with a single
...
+:math:`j` is the dataset number
+
+
+================= ================================== ==================================================================================
+Dataset attribute Behaviour                          Implementation
+================= ================================== ==================================================================================
+
+``livetime``.        Sum of individual livetimes         :math:`\overline{t} = \sum_j t_j`
+
+``mask_safe``        Pixels added with `OR` operation     :math:`\overline{\epsilon_k} = \sum_{j} \epsilon_{jk}`
```
adonath, line 90: In the implementation one could add the definition of: the mask is true if the pixel is included in the safe data range.

docs/datasets/index.rst
```diff
@@ -52,10 +52,83 @@ Note that in Gammapy, 2D image analyses are done with 3D cubes with a single
...
+
+``mask_fit``         Dropped
+
+``counts``           Summed outside exclusion region.     :math:`\overline{\mathrm{counts}_k} = \sum_j \mathrm{counts}_{jk} \cdot \epsilon_{jk}`
```
adonath, line 94: I would propose to use `mask_safe` consistently and not introduce "exclusion region". So something along the lines  of "Summed in the data range defined by `mask_safe`" or similar.

docs/datasets/index.rst
```diff
@@ -52,10 +52,83 @@ Note that in Gammapy, 2D image analyses are done with 3D cubes with a single
...
+
+``background``       Summed outside exclusion region.    :math:`\overline{\mathrm{bkg}_k} = \sum_j \mathrm{bkg}_{jk} \cdot \epsilon_{jk}`
+
+``exposure``         Summed outside spatial exclusion region    :math:`\overline{\mathrm{exposure}_l} = \sum_{j}{\mathrm{exposure}_{jl} \cdot \sum_k \epsilon_{jk}}`
+
+``psf``              Exposure weighted average                 :math:`\overline{\mathrm{psf}_l} = \frac{ \sum_{j}{\mathrm{psf}_{jl} \cdot \mathrm{exposure}_l} {\sum_{j} \cdot \mathrm{exposure}_l}`
+
+``edisp``            Exposure weighted average, with mask on reco energy :math:`\overline{\mathrm{edisp}_kl} = \frac{ \sum_{j}{\mathrm{edisp}_{jkl} \cdot \epsilon_{jk} \cdot \mathrm{exposure}_l} {\sum_{j} \cdot \mathrm{exposure}_l}`
```
adonath, line 102: ...with safe mask on reconstructed energy or just energy.

docs/datasets/index.rst
```diff
@@ -52,10 +52,83 @@ Note that in Gammapy, 2D image analyses are done with 3D cubes with a single
...
+
+- Stacking happens in-place, ie, dataset1.stack(dataset2) will overwrite dataset
+- To properly handle masks, it is necessary to stack onto an empty dataset.
+- Stacking only works for maps with equivalent geometry.
+ Two geometries are called equivalent if one is exactly the same as,
+ or can be obtained from a cutout of, the other.
+- A stacked analysis is reasonable only when adding runs taken by the same instrument.
+
+
+Joint Analysis
```
adonath, line 120: I think this definition is a better version of what we currently have in the glossary. So maybe just move to there for now and add a reference I the text above?

adonath (changes requested):
Thanks a lot @AtreyeeS! The overview table is really informative and seems complete to me. I have left some inline comments mostly on the uniform usage of the term `mask_safe`.

adonath:
@AtreyeeS What is the status here? I think once the remaining comment are implement this can go in...maybe remove the draft status?

AtreyeeS:
@adonath : This needs a little more work. 

- I want to add a plot showing that the `npred` computation is correct. 
- The table rendering is messed up. needs to be fixed.

adonath:
Thanks @AtreyeeS, the plot could be added any time later as well...

adonath:
Preview of the plot: 

image

docs/datasets/index.rst
```diff
@@ -52,10 +52,106 @@ Note that in Gammapy, 2D image analyses are done with 3D cubes with a single
...
+.. _stack:
+
+Stacking Multiple Datasets
+==========================
+
+Stacking datasets implies that the counts, background and reduced IRFs from all the
+runs are binned together to get one final dataset for which a likelihood is
+computed during the fit. Stacking is often useful to reduce the computation effort while
+analysing multiple runs.
+
```
adonath, line 67: Maybe move the table here and just say "the following table lists how the individual quantities are handled..."

docs/datasets/index.rst
```diff
@@ -52,10 +52,106 @@ Note that in Gammapy, 2D image analyses are done with 3D cubes with a single
...
+To ensure that the `npred` (ie, the predicted number of counts) on the stacked
+dataset is the sum expected by stacking the npred of the individual runs,
+a `~gammapy.irf.EDispersionMap` is used, which contains the
+The mask_safe from each dataset is applied on the respective reconstructed energy axis
+of the energy dispersion matrix, and the masked matrices are combined.
+Values lying outside the safe mask of each dataset are lost.
+
+The following plot shows the individual and stacked attributes for two `SpectrumDataset`
+
+.. plot:: datasets/plot_stack.py
```
adonath, line 84: From my perspective I think a plot just showing that npred is conserved would be sufficient. I think just adding two spectra as done for counts / background is maybe too simple to illustrate...

docs/datasets/index.rst
```diff
@@ -52,10 +52,106 @@ Note that in Gammapy, 2D image analyses are done with 3D cubes with a single
...
+It is important to keep in mind that:
+
+- Stacking happens in-place, ie, ``dataset1.stack(dataset2)`` will overwrite ``dataset1``
+- To properly handle masks, it is necessary to stack onto an empty dataset.
+- Stacking only works for maps with equivalent geometry.
+ Two geometries are called equivalent if one is exactly the same as,
+ or can be obtained from a cutout of, the other.
+- A stacked analysis is reasonable only when adding runs taken by the same instrument.
+
+.. _joint:
```
adonath, line 134: I'm not sure this is the right place to explain the difference joint and stacked analysis. I think the datasets documentation is still mostly on technical aspects and API. I think this information should rather go in the overview and a minimal version of it in the glossary...

docs/datasets/index.rst
```diff
@@ -52,10 +52,106 @@ Note that in Gammapy, 2D image analyses are done with 3D cubes with a single
...
+   * - ``psf``
+     - Exposure weighted average
+     - :math:`\overline{\mathrm{psf}_l} = \frac{\sum_{j} \mathrm{psf}_{jl} \cdot \mathrm{exposure}_{jl}} {\sum_{j} \mathrm{exposure}_{jl}}`
+   * - ``edisp``
+     - Exposure weighted average, with mask on reconstructed energy
+     - :math:`\overline{\mathrm{edisp}_{kl}} = \frac{\sum_{j}\mathrm{edisp}_{jkl} \cdot \epsilon_{jk} \cdot \mathrm{exposure}_{jl}} {\sum_{j} \mathrm{exposure}_{jl}}`
+   * - ``gti``
+     - Union of individual `gti`
+     -
+It is important to keep in mind that:
```
adonath, line 125: Maybe we make this a note box, like:

```
.. note:: 
```

adonath (approved):
Thanks @AtreyeeS, I have left some comments...

docs/datasets/index.rst
AtreyeeS: We can remove it if you want, but generally among beginners, there is some confusion as to what stacking really means - showing an example with counts and background can perhaps be helpful, and that npred stacking works really the same way

docs/datasets/index.rst
AtreyeeS: The only issue with putting it in the overview is it comes before people have an actual idea of what stack and joint fit mean. The associated tutorial (I can add one next week) can focus only on the API, but the can/should still explain the idea.

adonath: I think the main idea of the overview is to introduce new users to these concepts, without explaining them in detail and the link to the sub pages with more information. But I'm fine to keep it mostly as is for now and possibly move content later if we think it leads to clearer structure.

docs/datasets/index.rst
adonath: I still think this might be the better place for he table: so focus on "here is how it works"  first and then later on "with this definition the npred computation is conserved". The latter is the non-trivial aspect...

adonath:
@AtreyeeS I don't insist on it, but think for the understanding of the energy thresholds (instead of showing summed counts and background...) it might be more useful to show the resulting weighted edisp matrix side by side with the npred spectrum (see https://github.com/adonath/notebooks-public/blob/master/energy_threshold_stack.ipynb), because this is the non-trivial part, which needs to be explained in more detail...

AtreyeeS:
> @AtreyeeS I don't insist on it, but I think for the understanding of the handling of the energy thresholds (instead of showing summed counts and background...) it might be more useful to show the resulting weighted edisp matrix side by side with the npred spectrum (see https://github.com/adonath/notebooks-public/blob/master/energy_threshold_stack.ipynb), because this is the non-trivial part, which needs to be explained in more detail...

Thanks @adonath : Showing the edisp is a good idea - changed the plot!

adonath:
Thanks a lot @AtreyeeS, the structure and text looks good to me now! Could you check the plot one last time? Here is what it currently looks like:
image
So the first two panels are empty...

AtreyeeS:
Sorry! Fixed it now

adonath:
Thanks @AtreyeeS, I just looked closer at the plot and It think using the edisp from the CTA 1DC is non-ideal, because it was very noisy. However I would propose to merge the PR now and improve the plot e.g. during the co-working week.

AtreyeeS:
sounds good! I keep this in my To-Do list for next week

adonath (approved):
Thanks @AtreyeeS, the plot can be improved in follow up PRs.

PR #3283, opened 2021-03-23
Add Models.plot_spatial()

AtreyeeS:
This PR adds a `Models.plot()` as discussed in a dev call.
I call it `plot_spatial` for clarity.

### DIFF
--- a/gammapy/modeling/models/core.py
+++ b/gammapy/modeling/models/core.py
@@ -3,13 +3,16 @@ import logging
 import collections.abc
 import copy
 from os.path import split
+import yaml
 import numpy as np
 import astropy.units as u
 from astropy.table import Table
-import yaml
+from astropy.visualization.wcsaxes import WCSAxes
+from astropy.coordinates import  SkyCoord
+from regions import PointSkyRegion
 from gammapy.modeling import Covariance, Parameter, Parameters
 from gammapy.utils.scripts import make_name, make_path
-from gammapy.maps import RegionGeom, Map, WcsGeom
+from gammapy.maps import RegionGeom, Map
 
 
 log = logging.getLogger(__name__)
@@ -816,15 +819,32 @@ class DatasetModels(collections.abc.Sequence):
             spectral_model=spectral_model, spatial_model=spatial_model, name=name
         )
 
-    def get_regions(self):
+    @property
+    def positions(self):
+        """Positions of the models (`SkyCoord`)"""
+        positions = []
+
+        for model in self.select(tag="sky-model"):
+            if model.position:
+                positions.append(model.position)
+            else:
+                log.warning(
+                    f"Skipping model {model.name} - no spatial component present"
+                )
+
+        return SkyCoord(positions)
+
+    def to_regions(self):
         """Returns a list of the regions for the spatial models
 
         Returns
         -------
         regions: list of `~regions.SkyRegion`
+            Regions
         """
         regions = []
-        for model in self:
+
+        for model in self.select(tag="sky-model"):
             try:
                 region = model.spatial_model.to_region()
                 regions.append(region)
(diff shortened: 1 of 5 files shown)


gammapy/modeling/models/core.py
```diff
@@ -791,6 +791,57 @@ def reassign(self, dataset_name, new_dataset_name):
...
         return self.__class__(models)
 
+    def plot_spatial(self, ax=None, wcs=None, plot_extent=True, **kwargs):
+        """"Plot spatial models on a given geom
+
+        Parameters
+        ----------
+        ax : `~matplotlib.axes.Axes`, optional
+            Axis
+        wcs: `~astropy.wcs.WCS`
```
adonath, line 801: I think passing an optional wcs is not needed here, because users can always create a `WcsAxes` outside. The question is what to do for the case of `ax=None`. For `RegionGeom.plot_region()` we decided to just create a default wcs. This you can re-use here:
```python3
from gammapy.maps import RegionGeom
from gammapy.utils.regions import make_region
import numpy as np

regions = []

lons = np.random.uniform(-5, 5, 100)
lats = np.random.uniform(-3, 3, 100)
radii = np.random.uniform(0.1, 0.7, 100)

for lon, lat, radius in zip(lons, lats, radii):
    region = make_region(f"icrs;circle({lon}, {lat}, {radius})")
    regions.append(region)

geom = RegionGeom.from_regions(regions)
geom.plot_region()
```
To avoid code duplication you could possibly re-use the `RegionGeom.plot_region()` as well...

adonath (changes requested):
Thanks a lot @AtreyeeS, I think this is a very useful method. I added a few inline comments.

In general I think I might be OK to just start with a `.plot_regions()` method and the extend it later with `.plot_position()`,  `.plot_names()` and possibly `.plot_position_error(). These building blocks could be combined into general `.plot()` methods later, that does everything.

gammapy/modeling/models/core.py
AtreyeeS: `RegionGeom.plot()` creates by default "minimal equivalent WCS geometry", which is good for one model. But for plotting multiple models, a default all-sky option is better no?

AtreyeeS:
Thanks @adonath ! Addressed the changes - I now split it into plot_position() and plot_region()

gammapy/modeling/models/core.py
```diff
@@ -815,6 +816,97 @@ def to_template_sky_model(self, geom, spectral_model=None, name=None):
...
+        -------
+        ax : `~astropy.vizualisation.WcsAxes
+        """
+        import matplotlib.pyplot as plt
+        from astropy.visualization.wcsaxes import WCSAxes
+        from gammapy.modeling.models import PointSpatialModel
+
+        if ax is None or not isinstance(ax, WCSAxes):
+            wcs = WcsGeom.create().wcs
+            ax = plt.gca(projection=wcs)
```
adonath, line 841: Have you checked this minimal example:
```python3
from gammapy.maps import RegionGeom
from gammapy.utils.regions import make_region
import numpy as np

regions = []

lons = np.random.uniform(-5, 5, 100)
lats = np.random.uniform(-3, 3, 100)
radii = np.random.uniform(0.1, 0.7, 100)

for lon, lat, radius in zip(lons, lats, radii):
    region = make_region(f"icrs;circle({lon}, {lat}, {radius})")
    regions.append(region)

geom = RegionGeom.from_regions(regions)
ax = geom.plot_region()
```
In general Gammapy can determine the minimal contained WCS for an arbitrary list of regions...so this allows you to  possibly choose a better selected ROI for the models...

AtreyeeS: Then I can add a `Models.to_spatial_region()` which returns a list of regions and use this inside the plotting code. Is that OK?

AtreyeeS:
Hello @adonath ! As discussed, I changed `TemplateSpatialModel.to_region()` to return a `RectangeSkyRegion` and use the optimal was for plotting.

adonath:
Thanks a lot @AtreyeeS, I added one follow up commit to do a bit of clean up. For consistency renamed `get_regions()` to `to_regions()` and renamed `.geom` to `.wcs_geom` and made it return directly a `WcsGeom`.

PR #3284, opened 2021-03-24
fixing Matplotlib deprecation of nonposx

facero:
This pull request  fixes the deprecated kw nonposx in flux_point.py.
It was apparently recently fixed in spectral.py but not everywhere.
Also the doc here mixes description between nonposx and nonpositive kw:
https://github.com/gammapy/gammapy/blob/722cdd53c6f0768a6b8a3d71a4df91fa689c809e/gammapy/modeling/models/spectral.py#L322

For example, is this allowed for matplotlib >3.3 ?
`plt.loglog(nonposx='clip', nonpositive='clip')`

### DIFF
--- a/gammapy/estimators/flux_point.py
+++ b/gammapy/estimators/flux_point.py
@@ -664,8 +664,8 @@ class FluxPoints:
                 **kwargs,
             )
 
-        ax.set_xscale("log", nonposx="clip")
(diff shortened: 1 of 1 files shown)


adonath (approved):
Thanks @facero, no further comments from my side...

Issue #3292, opened 2021-04-02
Problems in 3HWC catalog


AtreyeeS:
**Gammapy version**

latest dev

**Bug description**

The 3HWC catalog handling is buggy. `SourceCatalog3HWC.to_models()` gives the following error

```
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
<ipython-input-93-7efa69dd73eb> in <module>
----> 1 cat_3hwc.to_models()

~/Gammapy-dev/gammapy/gammapy/catalog/core.py in to_models(self, **kwargs)
    225     def to_models(self, **kwargs):
    226         """ Create Models object from catalogue"""
--> 227         return Models([_.sky_model(**kwargs) for _ in self])
    228 
    229 

~/Gammapy-dev/gammapy/gammapy/catalog/core.py in <listcomp>(.0)
    225     def to_models(self, **kwargs):
    226         """ Create Models object from catalogue"""
--> 227         return Models([_.sky_model(**kwargs) for _ in self])
    228 
    229 

~/Gammapy-dev/gammapy/gammapy/catalog/hawc.py in sky_model(self, which)
    150         return SkyModel(
    151             spatial_model=self.spatial_model(which),
--> 152             spectral_model=self.spectral_model(which),
    153             name=self.name,
    154         )

~/Gammapy-dev/gammapy/gammapy/catalog/hawc.py in spectral_model(self, which)
    237             "index_err": 0.5
    238             * (
--> 239                 self.data[f"spec_{idx}_index_errp"]
    240                 + self.data[f"spec_{idx}_index_errn"]
    241             ),

KeyError: 'spec_0_index_errp'
```

**To Reproduce**
from gammapy.catalog import SourceCatalog3HWC
cat_3hwc = SourceCatalog3HWC()
cat_3hwc.to_models()

PR #3297, opened 2021-04-07
Fix and cleanup 3HAWC catalogue object

QRemy:
This PR fix SourceCatalog3HWC.sky_model() and solves #3292.
The methods .spectral_model(), .spatial_model(), .sky_model() were all broken because they wrongly assumed the same catalog structure for 3HWAC than for 2HWC. The changes include the following:
- As there is only one model per source in 3HAWC we can  set `n_models = 1`, remove `._get_idx() `and remove the `which` options  in the model functions.
- moved specific cases from base class to 2HWAC class
- corrected several typos in parameters and table column names.
- added missing tests for 3HAWC

### DIFF
--- a/gammapy/catalog/hawc.py
+++ b/gammapy/catalog/hawc.py
@@ -16,10 +16,7 @@ __all__ = [
 
 
 class SourceCatalogObjectHWCBase(SourceCatalogObject, abc.ABC):
-    """One source from the HAWC 2HWC catalog.
-
-    Catalog is represented by `~gammapy.catalog.SourceCatalog2HWC`.
-    """
+    """Base class for the HAWC catalogs objects"""
 
     _source_name_key = "source_name"
 
@@ -90,6 +87,13 @@ class SourceCatalogObjectHWCBase(SourceCatalogObject, abc.ABC):
 
         return ss
 
+
+class SourceCatalogObject2HWC(SourceCatalogObjectHWCBase):
+    """One source from the HAWC 2HWC catalog.
+
+    Catalog is represented by `~gammapy.catalog.SourceCatalog2HWC`.
+    """
+
     @property
     def n_models(self):
         """Number of models (1 or 2)."""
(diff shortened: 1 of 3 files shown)


AtreyeeS (changes requested):
Thanks a lot @QRemy!
`SourceCatalog3HWC.info()` is still broken. Can you please fix it?
Also, it would be nice to add a `is_pointlike` attribute like for the Fermi 4fgl catalog.

gammapy/catalog/hawc.py
```diff
@@ -67,29 +64,25 @@ def _info_position(self):
...
-
     def _info_spectrum(self):
         """Print spectral info."""
         ss = "\n*** Spectral info ***\n\n"
-        ss += self._info_spectrum_one(self.data, 0)
+        ss += self._info_spectrum_one(0)
 
         if self.n_models == 2:
-            ss += self._info_spectrum_one(self.data, 1)
+            ss += self._info_spectrum_one(1)
```
AtreyeeS, line 73: For the 3HWC, the info is a bit confusing.
There is a line "No second spectrum available for this source", which seems to imply that for some sources there is a 2nd one available, and then there is none. Perhaps this should not lie in the base class, but the `if` condition should be specific to the 2HWC? What do you think?

AtreyeeS (dismissed):
Thanks @QRemy ! This looks good to me, except for one comment I left

gammapy/catalog/hawc.py
QRemy: or I remove  "for this source" then the statement is ok for both catalogues

registerrier (dismissed):
Thanks @QRemy ! This looks good to me. No specific comment from my side.

gammapy/catalog/hawc.py
AtreyeeS: Yes, fine by me now.

AtreyeeS (approved):

adonath (approved):
Thanks a lot @QRemy, no further comment form my side...

PR #3299, opened 2021-04-09
Improve catalogue notebook

QRemy:
This PR re-arrange the structure of the catalogue notebook, and add or clarify several informations:
- about loading the catalogs
- about table, data, and position(s) properties
- how to select a catalogue subset using boolean mask
- how to create a Models object from a catalogue

### DIFF
--- a/docs/tutorials/api/catalog.ipynb
+++ b/docs/tutorials/api/catalog.ipynb
@@ -506,6 +506,60 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
+    "## Specificities of the HGPS catalog\n",
+    "\n",
+    "Using the `.to_models()` method for the `gammapy.catalog.SourceCatalogHGPS` will return only the models components of the sources retained in the main catalog, several candidate objects appears only in the Gaussian components table (see section 4.9 of the HGPS paper, https://arxiv.org/abs/1804.02432). To acces these components you can do the following:\n"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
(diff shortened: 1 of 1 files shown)


adonath (dismissed):
Thanks a lot @QRemy, I don't have any detailed comments here. I guess one could keep the `Exercises` and `Next steps` sections at the end, but we could also re-introduce those uniformly for all tutorials anyway later...

adonath:
Totally minor: Typically we use American English spelling in Gammapy (at least in the code base...) so for consistency one could always use `catalog` in the notebook as well..

AtreyeeS (dismissed):
Thanks @QRemy ! This looks quite complete now.

Perhaps it would be useful to add some more details on the HGPS calatog - adding examples for getting the large scale diffuse component and the discarded gaussian components.

QRemy:
> Perhaps it would be useful to add some more details on the HGPS calatog - adding examples for getting the large scale diffuse component and the discarded gaussian components.

Added a section about that.

> Totally minor: typically we use American English spelling in Gammapy (at least in the code base...) so for consistency one could always use `catalog` in the notebook as well..

Done for catalog but I did not check the spelling consistancy for the other words

> I guess one could keep the `Exercises` and `Next steps` sections at the end, but we could also re-introduce those uniformly for all tutorials anyway later...

I moved most of the comments from `Next steps` in the introduction as there is not really steps within the API notebooks and the follow-up is to look at analysis notebooks.

For the `Exercises` part, half of the tasks proposed was now somehow shown in the notebook so I removed it. Maybe that instead of many exercise examples dispersed in several notebook we could assemble them into few exercices/solutions notebooks to be used for hands-on sessions (post 1.0).

adonath (approved):
Thanks a lot @QRemy, I have no further comments.

PR #3301, opened 2021-04-12
Change HLI default for Fit range

registerrier:
**Description**

This pull request changes the default fit range min and max of the `AnalysisConfig` to `None` to avoid having a 1-10 TeV range applied silently.

**Dear reviewer**

### DIFF
--- a/gammapy/analysis/tests/test_analysis.py
+++ b/gammapy/analysis/tests/test_analysis.py
@@ -284,7 +284,6 @@ def test_analysis_1d_stacked_no_fit_range():
             method: reflected
     """
     config = AnalysisConfig.from_yaml(cfg)
-#    config = get_example_config("1d")
     analysis = Analysis(config)
     analysis.update_config(cfg)
     analysis.config.datasets.stack = True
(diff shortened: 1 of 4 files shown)


gammapy/analysis/tests/test_analysis.py
```diff
@@ -262,6 +262,44 @@ def test_analysis_1d_stacked():
     assert_allclose(pars["index"].value, 2.76913, rtol=1e-2)
     assert_allclose(pars["amplitude"].value, 5.479729e-11, rtol=1e-2)
 
+@requires_dependency("iminuit")
+@requires_data()
+def test_analysis_1d_stacked_no_fit_range():
```
adonath, line 267: Maybe one could just remove the fit range definition from one of the existing tests to avoid introducing a new one, that runs a full analysis again? Just to save a bit of runtime for the tests...

adonath (approved):
Thanks @registerrier, I just left one minor comment on the test....

adonath (approved):

Issue #3273, opened 2021-03-15
ExcessMapEstimator ignores models with datasets_names attribute set


luca-giunti:
**Gammapy version**
v0.18.2

**Bug description**
Given that in [this line](https://github.com/gammapy/gammapy/blob/1d4bcee496f6884bd99a2476d281fec7f785a23a/gammapy/estimators/excess_map.py#L154) the dataset name is changed (as no `name` parameter is passed to `resample_energy_axis`), the models assignment done three lines later ([here](https://github.com/gammapy/gammapy/blob/1d4bcee496f6884bd99a2476d281fec7f785a23a/gammapy/estimators/excess_map.py#L157)) fails for all models that have a `datasets_names` attribute set. As a result, the `ExcessMapEstimator.run()` ignores those models.

Probably passing the dataset `name` to the `resample_energy_axis` would solve this issue.

adonath:
Thanks @luca-giunti , indeed this looks like a bug. I also agree on your proposed solution. This is exactly how we should handle this in Gammapy internally, whenever we modify a dataset we should keep the name explicitly so that the model is still valid...

registerrier:
Well, the logic should rather be that we resample the background npred and not the background map. I think this was the case initially. 

I don't think having different datasets with identical names is a good design. Also, with a different energy axis, I don't think the npred calculation would be conserved before and after energy axis resampling.

adonath:
> Also, with a different energy axis, I don't think the npred calculation would be conserved before and after energy axis resampling

Naively I think it should. The true energy axis stays untouched and the reco energy axis is just summed up, whether this happens in the application of the RMF matrix or after in `resample_energy_axis()` shouldn't matter. But I have never checked...

adonath:
> I don't think having different datasets with identical names is a good design

I agree in general, however in this case the derived dataset is always internal and never(!) visible to the user. So having this pattern in `Estimator` classes is fine I guess:
```python3
modified_dataset = dataset.resample_energy_axis()
modified_dataset.models = dataset.models.copy()
```

registerrier:
> > Also, with a different energy axis, I don't think the npred calculation would be conserved before and after energy axis resampling
> 
> Naively I think it should. The true energy axis stays untouched and the reco energy axis is just summed up, whether this happens in the application of the RMF matrix or after in `resample_energy_axis()` shouldn't matter. But I have never checked...

Sorry I was unclear. What I meant is that:
[...]

Issue #3286, opened 2021-03-25
Allow to specify spectral model in ExcessMapEstimator


adonath:
Currently the `ExcessMapEstimator` does not allow to define the spectral model, that is used for the flux computation. It is easy to support and should be done...

PR #3303, opened 2021-04-13
Fix npred after energy resampling in ExcessMapEstimator

QRemy:
Issue #3273 pointed out that the ExcessMapEstimator.run() ignores the models that have a datasets_names attribute set, and that in order to preserve the predicted number of background counts during the energy axis resampling, we have to resample the npred_background. This PR implement the solution discussed.

### DIFF
--- a/gammapy/estimators/excess_map.py
+++ b/gammapy/estimators/excess_map.py
@@ -166,11 +166,14 @@ class ExcessMapEstimator(Estimator):
 
         axis = MapAxis.from_energy_edges(energy_edges)
 
-        resampled_dataset = dataset.resample_energy_axis(energy_axis=axis)
-
-        # Beware we rely here on the correct npred background in MapDataset.resample_energy_axis
-        resampled_dataset.models = dataset.models
-
+        resampled_dataset = dataset.resample_energy_axis(
+            energy_axis=axis, name=dataset.name
+        )
+        if isinstance(dataset, MapDatasetOnOff):
+            resampled_dataset.models = dataset.models
+        else:
+            resampled_dataset.background = dataset.npred().resample_axis(axis=axis)
+            resampled_dataset.models = None
         result = self.estimate_excess_map(resampled_dataset)
 
         return result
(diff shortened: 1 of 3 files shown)


adonath (approved):
Thanks a lot @QRemy, no comments from my side. I'll assign @luca-giunti once again, as he opened the issue initially...

adonath:
@QRemy After #3286 this needs a rebase...

QRemy:
@adonath updated

gammapy/estimators/excess_map.py
```diff
@@ -166,11 +166,14 @@ def run(self, dataset):
...
-        # Beware we rely here on the correct npred background in MapDataset.resample_energy_axis
-        resampled_dataset.models = dataset.models
-
+        resampled_dataset = dataset.resample_energy_axis(
+            energy_axis=axis, name=dataset.name
+        )
+        if isinstance(dataset, MapDatasetOnOff):
+            resampled_dataset.models = dataset.models
+        else:
+            resampled_dataset.background = dataset.npred().resample_axis(axis=axis)
```
registerrier, line 175: Why not resampling only the `npred_background` and keep the (non-background ) `models` on the resampled dataset? 
I guess it is fully equivalent anyway.

registerrier (approved):

gammapy/estimators/excess_map.py
QRemy: It is faster this way because the models don't have to be re-evaluated

luca-giunti (approved):

gammapy/estimators/excess_map.py
```diff
@@ -166,11 +166,14 @@ def run(self, dataset):
...
+            resampled_dataset.models = None
```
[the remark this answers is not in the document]
QRemy: For the `MapDatasetOnOff` we cannot set the background by hand so I just set the model that will be re-evaluated. 
For the `MapDataset` as Régis explained we have to re-sample the npred_background not the original background map, but it's faster to resample directly npred that include npred_background and npred_skymodels so we don't have to re-evaluate the models.

adonath (approved):

PR #3306, opened 2021-04-13
Allow to specify spectral model in ExcessMapEstimator

### LINKED ISSUE #3286: Allow to specify spectral model in ExcessMapEstimator

QRemy:
Resolve #3286. Allow to specify spectral model in ExcessMapEstimator for flux computation. 
I can't see any difference in the computed flux from the test but that's done.

### DIFF
--- a/gammapy/estimators/excess_map.py
+++ b/gammapy/estimators/excess_map.py
@@ -89,6 +89,9 @@ class ExcessMapEstimator(Estimator):
         A `~gammapy.datasets.MapDataset.mask_fit` must be present on the input dataset
     correlate_off : Bool
         Correlate OFF events in the case of a MapDatasetOnOff
+    spectral_model : ~gammapy.modeling.models.SpectralModel
+        Spectral model used for the computation of the flux map. 
+        If None, a Power Law of index 2 is assumed (default). 
     """
 
     tag = "ExcessMapEstimator"
@@ -102,7 +105,8 @@ class ExcessMapEstimator(Estimator):
         selection_optional=None,
         energy_edges=None,
         apply_mask_fit=False,
-        correlate_off=False
+        correlate_off=False,
+        spectral_model=None,
     ):
         self.correlation_radius = correlation_radius
         self.n_sigma = n_sigma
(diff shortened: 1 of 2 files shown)


gammapy/estimators/excess_map.py
```diff
@@ -89,6 +89,9 @@ class ExcessMapEstimator(Estimator):
         A `~gammapy.datasets.MapDataset.mask_fit` must be present on the input dataset
     correlate_off : Bool
         Correlate OFF events in the case of a MapDatasetOnOff
+    spectral_model : ~gammapy.modeling.models.SpectralModel
```
adonath, line 92: This misses the backticks so that sphinx will create the link...

gammapy/estimators/tests/test_excess_map.py
```diff
@@ -219,6 +231,19 @@ def test_significance_map_estimator_map_dataset_on_off_with_correlation(simple_d
...
+
+    estimator_mod = ExcessMapEstimator(
+        0.11 * u.deg,
+        apply_mask_fit=False,
+        correlate_off=True,
+        spectral_model=PowerLawSpectralModel(index=10),
+    )
+    result_mod = estimator_mod.run(simple_dataset_on_off)
+
+    assert result_mod["flux"].unit == u.Unit("cm-2s-1")
```
adonath, line 244: Minor: one can actually directly compare to a string like: `unit == "cm-2 s-1"`

gammapy/estimators/tests/test_excess_map.py
```diff
@@ -219,6 +231,19 @@ def test_significance_map_estimator_map_dataset_on_off_with_correlation(simple_d
...
+    assert_allclose(result_mod["flux"].data.sum(), 5.254442192077636e-06, rtol=1e-8)
+    # TODO: find a test where we can actually see a difference with the default case
```
adonath, line 246: I think this would be important. For adding an assert on dnde should be fine? This should be different for both cases...

adonath (commented):
Thanks a lot @QRemy, looks good! I have left some minor inline comments...

registerrier (approved):

adonath (approved):

PR #3307, opened 2021-04-13
Add support for ExcessMapEstimator on HLI

registerrier:
**Description**

This pull request introduces a new config argument for `ExcessMapEstimator` as well as a the corresponding method `Analysis.get_excess_map()`

The default values for `EnergyAxisConfig` arguments is set to None, but default values are added to `EnergyAxesConfig`
**Dear reviewer**

### DIFF
--- a/gammapy/analysis/config.py
+++ b/gammapy/analysis/config.py
@@ -139,8 +139,8 @@ class FitConfig(GammapyBaseConfig):
 
 
 class ExcessMapConfig(GammapyBaseConfig):
-    correlation_radius: AngleType = '0.1 deg'
-    parameters: dict = {"selection_optional": ""}
+    correlation_radius: AngleType = "0.1 deg"
+    parameters: dict = {}
     energy_edges: EnergyAxisConfig = EnergyAxisConfig()
 
 
--- a/gammapy/analysis/config.py
+++ b/gammapy/analysis/config.py
@@ -106,9 +106,9 @@ class SkyCoordConfig(GammapyBaseConfig):
 
 
 class EnergyAxisConfig(GammapyBaseConfig):
-    min: EnergyType = "0.1 TeV"
-    max: EnergyType = "10 TeV"
-    nbins: int = 30
+    min: EnergyType = None
+    max: EnergyType = None
+    nbins: int = None
 
 
 class SpatialCircleConfig(GammapyBaseConfig):
@@ -141,7 +141,7 @@ class FitConfig(GammapyBaseConfig):
 class ExcessMapConfig(GammapyBaseConfig):
     correlation_radius: AngleType = '0.1 deg'
     parameters: dict = {"selection_optional": ""}
-    energy_edges: EnergyAxisConfig = None
+    energy_edges: EnergyAxisConfig = EnergyAxisConfig()
 
 
 class BackgroundConfig(GammapyBaseConfig):
@@ -156,8 +156,8 @@ class SafeMaskConfig(GammapyBaseConfig):
 
 
 class EnergyAxesConfig(GammapyBaseConfig):
-    energy: EnergyAxisConfig = EnergyAxisConfig()
-    energy_true: EnergyAxisConfig = EnergyAxisConfig()
+    energy: EnergyAxisConfig = EnergyAxisConfig(min="1 TeV", max="10 TeV", nbins=5)
+    energy_true: EnergyAxisConfig = EnergyAxisConfig(min="0.5 TeV", max="20 TeV", nbins=16)
 
 
 class SelectionConfig(GammapyBaseConfig):
--- a/gammapy/analysis/config.py
+++ b/gammapy/analysis/config.py
@@ -138,6 +138,12 @@ class FitConfig(GammapyBaseConfig):
     fit_range: EnergyRangeConfig = EnergyRangeConfig()
 
 
+class ExcessMapConfig(GammapyBaseConfig):
+    correlation_radius: AngleType = '0.1 deg'
+    parameters: dict = {"selection_optional": ""}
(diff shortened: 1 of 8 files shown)


gammapy/analysis/config.py
```diff
@@ -138,6 +138,12 @@ class FitConfig(GammapyBaseConfig):
     fit_range: EnergyRangeConfig = EnergyRangeConfig()
 
 
+class ExcessMapConfig(GammapyBaseConfig):
+    correlation_radius: AngleType = '0.1 deg'
+    parameters: dict = {"selection_optional": ""}
```
Bultako, line 143: I think `"selection_optional": ""` can be safely removed, there are other possible parameters allowed, so better declare only those that we want to set by default. Let's make `parameters: dict = {}` like in the other classes.

gammapy/analysis/config.py
```diff
@@ -138,6 +138,12 @@ class FitConfig(GammapyBaseConfig):
     fit_range: EnergyRangeConfig = EnergyRangeConfig()
 
 
+class ExcessMapConfig(GammapyBaseConfig):
+    correlation_radius: AngleType = '0.1 deg'
```
Bultako, line 142: "0.1 deg" double quotes

Bultako (changes requested):
Thanks @registerrier I've added some comments in-between the lines of code.

gammapy/analysis/config.py
registerrier: Done

gammapy/analysis/config.py
registerrier: Indeed. The default is now `selection_optional=None` so we don't need to pass anything.

gammapy/analysis/core.py
```diff
@@ -384,12 +407,15 @@ def _spectrum_extraction(self):
...
-        return MapAxis.from_bounds(
-            name=name,
-            lo_bnd=axis.min.value,
-            hi_bnd=axis.max.to_value(axis.min.unit),
-            nbin=axis.nbins,
-            unit=axis.min.unit,
-            interp="log",
-            node_type="edges",
-        )
+        if axis.min is None and axis.max is None:
```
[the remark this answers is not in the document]
registerrier: I wanted to have the default energy axis to be `None` so that we could pass None arguments to estimators.  Hence, raising `ValueError` is not completely adapted.
I followed your second suggestion. 
We might add a `log.info` or `log.warning` when the axis is left to `None` since the type validation is not complete in this case?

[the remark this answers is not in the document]
Bultako: Ok, thinking more carefully about this I guess we should better do the values checking validation directly in the `EnergyAxisConfig` class in the `config.py` file.  i.e. `min` value declared and `max` value missing, `nbins` missing when both `min` and `max` are declared, etc. for the moment we could just place a simple `log.info` here when we return `None`. Something as what is shown below.

```
if axis.min is None:
    log.info("No energy axis defined.") 
    return None
```

I will make another PR addressing complex validation among fields in the different classes of the model of the HLI, this concerns also other classes than this `EnergyAxisConfig`.

Bultako:
@registerrier 

Two more things I've found:
- I think it would be nice to add the new config block `excess_map` to the file `analysis/config/example-1d.yaml` and analysis/config/docs.yaml`, even if it s not needed, but I think it is good to have it (with the default values is ok) so we can easily inspect allowed params in a sample config file.. so we could maybe use it for doc purposes in the future.
- I see you use `config.excess_map`, `analysis.get_excess_map()`,  `analysis.excess_maps`, etc. could we be more consistent and use only the plural or singular form?

gammapy/analysis/core.py
```diff
@@ -207,6 +207,29 @@ def get_flux_points(self):
...
+        energy_edges = self._make_energy_axis(excess_settings.energy_edges)
+        if energy_edges is not None:
+            energy_edges = energy_edges.edges
+
+        excess_map_estimator = ExcessMapEstimator(
+            correlation_radius=excess_settings.correlation_radius,
+            energy_edges=energy_edges,
+            **excess_settings.parameters
+        )
+        self.excess_maps = excess_map_estimator.run(self.datasets[0])
```
adonath, line 231: I just asked myself which error is thrown if the `ExcessMapEstimator` is called with a `SpectrumDataset`. I found https://github.com/gammapy/gammapy/blob/master/gammapy/estimators/excess_map.py#L154, which is basically fine, but one could maybe improve the message a bit and say which dataset type was passed and which one is expected, just because the HLI is probably more likely to be used by beginners.

adonath (dismissed):
Thanks @registerrier , I have left on minor inline comment, otherwise looks fine...

registerrier:
> @registerrier
> 
> Two more things I've found:
> 
> * I think it would be nice to add the new config block `excess_map` to the files `analysis/config/example-1d.yaml` and `analysis/config/docs.yaml`, even if it s not needed, but I think it is good to have it (with the default values is ok) so we can easily inspect allowed params in a sample config file.. and maybe use it for doc purposes in the future.
> * I see you use `config.excess_map`, `analysis.get_excess_map()`,  `analysis.excess_maps`, etc. could we be more consistent and use only the plural or singular form?

Good points. I have updated `analysis/config/example-3d.yaml` and  `analysis/config/docs.yaml`.
I also changed the result product property to `Analysis.excess_map` for consistency.

gammapy/analysis/core.py
registerrier: Indeed I changed the message to explicitly state that 1D datasets are not supported.

adonath (approved):
Thanks @registerrier, I have no further comments.

Issue #3295, opened 2021-04-06
FluxPointEstimator optional keywords 


AtreyeeS:
**Gammapy version**
latest dev

**Bug description**
in the `selection_optional` keyword for `FluxPointEstimator`, the documentation mentions `norm-scan` as an allowed keyword. However, the accepted keyword is rather `scan`. This should be corrected.

Also, if `norm-scan` is not executed, `.plot_ts_profile` does not work.
This is expected behaviour, but it should be documented. And the error message should be improved. Presently, there is no check and the error happens somewhere in Astropy-packages, and the message is difficult to understand.

QRemy:
Also the doc mentions `selection_optional : list of str`, but`selection_optional = "all"` works, while `selection_optional = ['all']` fails. Maybe the default should rather be selection_optional = None.

 The description of selection_optional should be:  "Which ~~additional~~ specific quantities to estimate", otherwise "additionnal" suggests  estimated on top of the defaults.

adonath:
A few more points from my side:
- I agree the docstring should mention a str for "all" explicitly
- I think the description is correct: `selection_optional` specifies the quantities on top of the default ones (such as `norm`, `ts`, `norm_err`, etc.). The default quantities cannot be specified.
- However the description of the default is incorrect, the doctoring still says: "By default all steps are executed.", which is not the case.
- In addition we should mention which default quantities are computed or at least provide a link to the table in https://docs.gammapy.org/dev/estimators/index.html#introduction

QRemy:
Looking closer to the code I reached the same conclusions than you @adonath and implemented that in #3309 (except you last point).

PR #3309, opened 2021-04-14
Improve doc and cleanup for selection_optional argument in estimators

QRemy:
This PR fix the "all" option for in selection_optional argument of estimators and improve the docstring. Resolve issue #3295.

### DIFF
--- a/gammapy/estimators/flux.py
+++ b/gammapy/estimators/flux.py
@@ -42,12 +42,12 @@ class FluxEstimator(Estimator):
         Re-optimize other free model parameters.
     selection_optional : list of str
         Which additional quantities to estimate. Available options are:
-
+            * "all": all the optional steps are executed
             * "errn-errp": estimate asymmetric errors.
             * "ul": estimate upper limits.
             * "norm-scan": estimate fit statistic profiles.
 
-        By default all steps are executed.
+        Default is None so the optionnal steps are not executed.
     """
 
     tag = "FluxEstimator"
--- a/gammapy/estimators/flux.py
+++ b/gammapy/estimators/flux.py
@@ -42,10 +42,11 @@ class FluxEstimator(Estimator):
         Re-optimize other free model parameters.
     selection_optional : list of str
         Which additional quantities to estimate. Available options are:
+
             * "all": all the optional steps are executed
             * "errn-errp": estimate asymmetric errors.
             * "ul": estimate upper limits.
-            * "norm-scan": estimate fit statistic profiles.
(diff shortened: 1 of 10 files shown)


gammapy/estimators/flux.py
```diff
@@ -42,12 +42,12 @@ class FluxEstimator(Estimator):
         Re-optimize other free model parameters.
     selection_optional : list of str
         Which additional quantities to estimate. Available options are:
-
+            * "all": all the optional steps are executed
             * "errn-errp": estimate asymmetric errors.
             * "ul": estimate upper limits.
             * "norm-scan": estimate fit statistic profiles.
```
AtreyeeS, line 48: It should be `scan` instead of `norm-scan`, no?

gammapy/estimators/flux_point.py
```diff
@@ -784,12 +784,12 @@ class FluxPointsEstimator(Estimator):
         Re-optimize other free model parameters.
     selection_optional : list of str
         Which additional quantities to estimate. Available options are:
-
+            * "all": all the optional steps are executed
             * "errn-errp": estimate asymmetric errors on flux.
             * "ul": estimate upper limits.
             * "norm-scan": estimate fit statistic profiles.
```
AtreyeeS, line 790: same as before - `scan` instead of `norm-scan`

AtreyeeS (commented):
Thanks a lot @QRemy ! I think this PR still needs to homogenise between `scan` and `norm-scan` keywords. Also, as I wrote in #3295, it would be helpful to for a user to have a proper message in `.plot_ts_profile` if `scan` is not execute - maybe searching for the column and throwing an error is useful, instead of incomprehensible message we have now

gammapy/estimators/flux.py
QRemy: yes, fixed in the latest commit

QRemy:
@AtreyeeS, the message I see for `FluxPoints.plot_ts_profiles()` is`"Missing columns for sed type '{}':" " {}".format(sed_type, missing)` which is quite explicit. Do you have an example to reproduce the odd error meaage you have ?

QRemy:
I added a check for optional arguments missing in `._validate_table(table, sed_type, use_optional=False)`.

AtreyeeS (approved):
Thanks again @QRemy ! No further comments from my side

adonath (approved):
Thanks @QRemy and @AtreyeeS for the review. I have no further comments either...

PR #3314, opened 2021-04-15
Implement HpxNDMap.smooth()

LauraOlivera:
**Description**

This pull request implements a function to smooth HpxNDMaps for two different choices of the kernel, gaussian and disk.
It relies on `healpy.sphtfunc.smoothing` for the smoothing and on `healpy.sphtfunc.beam2bl()` to construct the beam function in the harmonics space from the angular radial profile of the disk

**Dear reviewer**
I marked it as draft because I am unsure of the normalization for the disk window_beam function. The method `healpy.sphtfunc.smoothing` has a dedicated implementation of a gaussian beam, which looks like this in l-space
image
When transforming it to angular space via `healpy.sphtfunc.bl2beam()` we get
image
where the x axis are angles in radians.

Analogous to this case, I construct the angular radial disk profile with a step function:
image
and then normalize the corresponding window beam function so that it peaks at 1, like in the Gaussian case:
image

but I am not sure how to test that this is 100% correct.

### DIFF
--- a/gammapy/maps/hpxnd.py
+++ b/gammapy/maps/hpxnd.py
@@ -403,7 +403,11 @@ class HpxNDMap(HpxMap):
         """
         import healpy as hp
 
+        if not self.geom.is_allsky:
+            raise NotImplementedError("Smoothing is only possible for all-sky maps")
+
         nside = self.geom.nside
+        lmax = 3*nside-1 # maximum l of the power spectrum
         nest = self.geom.nest
 
         # The smoothing width is expected by healpy in radians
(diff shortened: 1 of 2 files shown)


gammapy/maps/hpxnd.py
```diff
@@ -379,6 +379,75 @@ def stack(self, other, weights=None):
...
+        kernel : {'gauss', 'disk'}
+            Kernel shape
+
+        Returns
+        -------
+        image : `HpxNDMap`
+            Smoothed image (a copy, the original object is unchanged).
+        """
+        import healpy as hp
+
```
adonath, line 405: I think this currently only works for an allsky map. So maybe raise and error here, if it is a partial HPX map.

gammapy/maps/hpxnd.py
```diff
@@ -379,6 +379,75 @@ def stack(self, other, weights=None):
...
+                # reorder to ring to do the smoothing
+                img = hp.pixelfunc.reorder(img, n2r=True)
+
+            if kernel == "gauss":
+                data = hp.sphtfunc.smoothing(img, sigma=width,pol=False, verbose=False)
+            elif kernel == "disk":
+                lmax = 3*nside-1
+                # create the step function in angular space
+                # (this is not great)
+                theta = np.append(np.linspace(0,width), np.linspace(width+1e-7, width+1, 10))
```
adonath, line 433: I think this can be simplified, the "final" resolution is probably given by the nside and a finer sampling of the edges, has no effect. But maybe check once again...

gammapy/maps/hpxnd.py
```diff
@@ -379,6 +379,75 @@ def stack(self, other, weights=None):
...
+                beam = np.ones(len(theta))
+                beam[theta>width]=0
+                # convert to the spherical harmonics space
+                window_beam = hp.sphtfunc.beam2bl(beam, theta, lmax)
+                # normalize the window beam
+                window_beam = window_beam/window_beam.max()
```
adonath, line 439: I double checked and the normalisation seems correct! So the mean of the data is preserved...

adonath (changes requested):
Thanks a lot @LauraOlivera, I have left some minor inline comments. I think the PR is in good state and can be merged, once the raining comments are addressed.

gammapy/maps/hpxnd.py
LauraOlivera: I checked and indeed it makes no difference, so I simplified it.

LauraOlivera: Also I explored changing the `lmax` parameter and it doesn't make much of a difference and would only add complexity to expose it as a user input I think. I however added a comment in the code explaining what it is in case a very advanced user ever needs it, so one could look at the code and quickly understand what is the value that is being used.

adonath:
Thanks a lot @LauraOlivera. no further comments from my side.  It think documentation wise we can leave it as is for now and maybe add something here later: https://docs.gammapy.org/0.18.2/maps/hpxmap.html...

Issue #3311, opened 2021-04-14
FoVBackgroundMaker fit option


AtreyeeS:
The `FoVBackgroundMaker` takes two options for pre-computing the background norms
- `scale` - the predicted background counts are just scaled to the total counts (outside the exclusion regions)
- `fit` - the  predicted background is fit to the data

The `fit` and `scale` should be identical in case of the `tilt` being frozen, I imagine?
So I would expect that the `fit` option should fit both the `norm` and the `tilt`? 
At present, the tilt is frozen by default.
Should we change  it? @registerrier @adonath @QRemy @luca-giunti

adonath:
@AtreyeeS No, taking it strictly I think the `scale` and `fit` option are only equivalent for either very high statistics per bin, or if there is just a single spatial bin to fit. Both of which are not neccessarilly the case. In practice I guess they deliver similar results most of the time, if only the norm of the background is adjusted, but I would expect different results for low statistics...

AtreyeeS:
I see. But it can still be nice to be able to fit the `tilt` during `FoVBackgroundMaker.run()`, no?

adonath:
@AtreyeeS Yes, definitely! But this already works, if users pre-define the `FoVBackgroundModel`, or do you just propose to change the default behaviour in `FoVBackgroundMaker`?

AtreyeeS:
I feel it would be convenient to be able to fit the tilt without having to pre-define the `FoVBackgroundModel`.

eg: I want to fit the norm and tilt during the data reduction process and create a stacked map, instead of having a 2 step process

```
stacked = MapDataset.create(
    geom=geom, name="stacked"
    )
maker = MapDatasetMaker(
        selection=["counts", "exposure", "edisp", "psf", "background"])

safe_maker = SafeMaskMaker(
    methods=['offset-max'], 
    offset_max=offset_max, aeff_percent=10,
    )
bkg_maker = FoVBackgroundMaker(exclusion_mask=exclusion_mask, method='fit')
for obs in observations:
    cutout = stacked.cutout(
        obs.pointing_radec, width=2 * offset_max, name=f"obs-{obs.obs_id}"
    )
    dataset = maker.run(cutout, obs)
    dataset = safe_maker.run(dataset, obs)
    dataset = bkg_maker.run(dataset, obs) ## nice to stack the fitted background directly.
    stacked.stack(dataset)
```

Maybe it can be made configurable on `init`?

registerrier:
[...]

PR #3319, opened 2021-04-21
Modify FoVBackgroundMaker to take spectral model as argument

registerrier:
**Description**

This pull request solves #3311  by changing the `spectral_model_tag` argument to `spectral_model`. It can now take either a tag such as `"pl-norm"` (still the default). 
On init a `fov_maker.default_spectral_model` is created and is applied if no `FoVBavkgroundModel` is predefined on the `Dataset`.

This has the advantage of being more configurable than manually creating a `FoVBackgroundModel` if one wants to unfreeze one parameter or change its boundaries. 
This is also probably preferable to changing the class default on the fly.

**Dear reviewer**

### DIFF
--- a/gammapy/makers/background/fov.py
+++ b/gammapy/makers/background/fov.py
@@ -28,23 +28,28 @@ class FoVBackgroundMaker(Maker):
         the normalization method to be applied. Default 'scale'.
     exclusion_mask : `~gammapy.maps.WcsNDMap`
         Exclusion mask
-    spectral_model_tag : str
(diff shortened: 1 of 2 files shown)


adonath (approved):
Thanks @registerrier, no further comments from my side. While looking at the PR I noticed we don't have any naming convention on defaults with prefix / suffix, like `default_something` or `something_default`. Could keep in mind for later clean up PR.

adonath:
This needs a changelog entry...

PR #3323, opened 2021-04-27
Implement HpxNDMap.convolve()

LauraOlivera:
This pull requests implements several methods required for the convolution of a HealPix map with a kernel. There are several options to do this: the kernel being a WCS or Hpx map, the convolution happening in a small cutout or in the full sky map.

For now, I've implemented `convolve_wcs()` which is the "simplest" of these approaches in my opinion. For a "small map" (I set a limit of 10 degree width, but we can also refine that), it uses the `HpxToWcsMapping` to project the initial `HpxNDMap` into a `WcsNDMap`, convolves with a kernel using the existing `WcsNDMap.convolve()` and makes use of the mapping to go back to the inital `HpxGeom`. To make this work this I implemented a method `HpxToWcsMapping.fill_hpx_map_from_wcs_data()` which just does the inverse as the existing `HpxToWcsMapping.fill_wcs_map_from_hpx_data()`. Testing for this is not implemented yet. (**TO DO**)

The caveat is that the geometry of the kernel has to be compatible with the intermediate `WcsGeom`, and so the kernel has to be extracted using `hpx_geom.to_wcs_geom()` where `hpx_geom` is the inital map geometry. That means, that if this was used for PSF convolution in a `MapDataset` with Healpix geometry, the geometry of the `PSFKernel` (and Map?) would need to be different than the rest of components.

**TO DO**: The next step would be to, from an inital kernel in an unspecified geometry, extract the radial profile of the kernel in angular space (assumed symmetric), convert to a beam function using `healpy.sphtfunc.beam2bl()` and then make use of `healpy.sphtfunc.smoothing`. This happens in a full-sky map, which would require a cutout to be stacked in a full-sky map as an initial step. I suspect this will be much slower than the case above, but I have to check. For a kernel also in an `HpxMap`,  there might be faster options.

### DIFF
--- a/gammapy/maps/hpxnd.py
+++ b/gammapy/maps/hpxnd.py
@@ -410,12 +410,23 @@ class HpxNDMap(HpxMap):
         """
         import healpy as hp
 
-        if not self.geom.is_allsky:
-            raise NotImplementedError("Smoothing is only possible for all-sky maps")
-
         nside = self.geom.nside
         lmax = 3*nside-1 # maximum l of the power spectrum
         nest = self.geom.nest
+        allsky = self.geom.is_allsky
+        ipix = self.geom._ipix
+
+        if not allsky: #stack into an all sky map
+            full_sky_geom = HpxGeom.create(nside = self.geom.nside,
+                                        nest = self.geom.nest,
+                                        frame = self.geom.frame,
+                                        axes = self.geom.axes
+                                        )
+            full_sky_map = HpxNDMap.from_geom(full_sky_geom)
+            for img, idx in self.iter_by_image():
+                full_sky_map.data[idx][ipix] = img
+        else:
+            full_sky_map = self
 
         # The smoothing width is expected by healpy in radians
         if isinstance(width, (u.Quantity, str)):
@@ -428,7 +439,7 @@ class HpxNDMap(HpxMap):
 
         smoothed_data = np.empty(self.data.shape, dtype=float)
 
-        for img, idx in self.iter_by_image():
+        for img, idx in full_sky_map.iter_by_image():
             img = img.astype(float)
 
             if nest:
@@ -453,9 +464,39 @@ class HpxNDMap(HpxMap):
             if nest:
                 # reorder back to nest after the smoothing
                 data = hp.pixelfunc.reorder(data, r2n=True)
-            smoothed_data[idx] = data
+            smoothed_data[idx] = data[ipix]
 
         return self._init_copy(data=smoothed_data)
+
+    def convolve(self, kernel, use_fft=True, **kwargs):
+        """
+        Convolve map with a WCS kernel.
+
+        It projects the map into a WCS geometry, convolves with a WCS kernel and
+        projects back into the initial Healpix geometry.
+
+        If the kernel is two dimensional, it is applied to all image planes likewise.
+        If the kernel is higher dimensional it must match the map in the number of
+        dimensions and the corresponding kernel is selected for every image plane.
+
+        Parameters
+        ----------
+        kernel : `~gammapy.irf.PSFKernel`
+            Convolution kernel. The pixel size must be upsampled by a factor 2 or bigger
+            with respect to the input map to prevent artifacts in the projection.
+        use_fft : bool
+            Use `scipy.signal.fftconvolve` if True (default)
+            and `ndi.convolve` otherwise.
+        kwargs : dict
+            Keyword arguments passed to `scipy.signal.fftconvolve` or
+            `ndi.convolve`.
+
+        Returns
+        -------
+        map : `HpxNDMap`
+            Convolved map.
+        """
+        return convolve_wcs(self, kernel, use_fft=True, **kwargs)
 
     def convolve_wcs(self, kernel, use_fft=True, **kwargs):
         """
@@ -489,7 +530,7 @@ class HpxNDMap(HpxMap):
             log.warning("Convolution via WCS projection is not recommended for large maps. "\
                 "Perhaps the method `convolve_full()` is more suited for this case.")
         wcs_geom = kernel.psf_kernel_map.geom.to_image()
-        wcs_size = np.min(wcs_geom._cdelt)
+        wcs_size = np.max(wcs_geom.pixel_scales.deg)
         hpx_size = get_pix_size_from_nside(self.geom.nside[0])
         if wcs_size > 0.5*hpx_size:
             raise ValueError("The kernel pixel size has to be smaller by at least a factor 2 than the input map to prevent artifacts.")
@@ -543,7 +584,7 @@ class HpxNDMap(HpxMap):
                                         )
             full_sky_map = HpxNDMap.from_geom(full_sky_geom)
             for img, idx in self.iter_by_image():
-                full_sky_map.data[idx, ipix] = img
+                full_sky_map.data[idx][ipix] = img
         else:
             full_sky_map = self
 
--- a/gammapy/maps/hpxnd.py
+++ b/gammapy/maps/hpxnd.py
@@ -496,7 +496,7 @@ class HpxNDMap(HpxMap):
         map : `HpxNDMap`
             Convolved map.
         """
-        return convolve_wcs(self, kernel, use_fft=True, **kwargs)
+        return self.convolve_wcs(kernel, use_fft, **kwargs)
 
     def convolve_wcs(self, kernel, use_fft=True, **kwargs):
         """
--- a/gammapy/maps/hpxnd.py
+++ b/gammapy/maps/hpxnd.py
@@ -5,6 +5,7 @@ import astropy.units as u
 from astropy.coordinates import SkyCoord
 from regions import PointSkyRegion
 from gammapy.utils.units import unit_from_fits_image_hdu
+import logging
 from .region import RegionGeom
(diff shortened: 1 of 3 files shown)


gammapy/maps/hpxnd.py
```diff
@@ -452,6 +453,47 @@ def smooth(self, width, kernel="gauss"):
...
+        kwargs : dict
+            Keyword arguments passed to `scipy.signal.fftconvolve` or
+            `ndi.convolve`.
+
+        Returns
+        -------
+        map : `HpxNDMap`
+            Convolved map.
+        """
+        if self.geom.width < 10*u.deg:
```
adonath, line 484: I think the number "10 deg" should be introduce as a class level variable like:
```python3

class HpxNDMap:
    convolve_wcs_max_width
```
This allows users in general to overwrite it, in case they know what they are doing...

adonath: Or maybe even better: alternatively we could just say all partial maps are supported and we make a statement in the docstring on the accuracy for larger maps >~10 deg

gammapy/maps/hpxnd.py
```diff
@@ -452,6 +453,47 @@ def smooth(self, width, kernel="gauss"):
...
+            # Project to WCS and convolve
+            wcs_map  = self.to_wcs(fill_nan = False)
```
adonath, line 486: I think `.to_wcs()` takes a pre-computed mapping as well. There might be a performance improvement if the `HpxToWcsMapping` is created outside and re-used for both cases i.e. `.to_wcs()` as well as the later inverse operation.

adonath: I think this even offers the possibly to adjust the WCS geometry used for the conversion and convolution, to the WCS geometry of the kernel, that is passed in. This eliminates the need to compute a matching kernel outside. I haven't thought this through completely, but I think it works...

adonath (changes requested):
Thanks lot @LauraOlivera, I have left some fits comments, I'll check out the functionality now...

gammapy/maps/hpxnd.py
LauraOlivera: oh, you are right! that is a very good point! I'll try that out

gammapy/maps/hpxnd.py
LauraOlivera: I agree. My idea was that ideally we would have this method (ok for small maps) and then a full-sky map one that involves stacking a partial map into a full-sky map if needed (which I expect to be slower, but if it isn't then I need to rethink things), and if you try to use `convolve_wcs` on a partial map too large, you just get a warning suggesting that the other one is more suited or something. But it needs some more thought/work.

gammapy/maps/hpxnd.py
LauraOlivera: It works but I am seeing a lot more issues with artifacts that only go away if the wcs_geom binning is very fine. The nice thing about using `.to_wcs()` is that it guaranteed the oversampling... I guess it could be fixed by requiring a minimum wcs pixel size ratio to hpx pixel size...
```
%%time
energy = MapAxis.from_bounds(1,100, unit='TeV', nbin=2, name='energy')

nside = 1024
hpx_geom = HpxGeom.create(nside=nside,
                         axes=[energy],
                         region='DISK(0,0,2.5)',
                         )

hpx_map = Map.from_geom(hpx_geom)
hpx_map.set_by_coord((0,0, [2,90]), 1)

wcs_geom = WcsGeom.create(width=5,
                          binsz = 0.05,
                         axes = [energy])
kernel = PSFKernel.from_gauss(wcs_geom, 0.5*u.deg)
convolved_map = hpx_map.convolve_wcs(kernel)
convolved_map.plot_interactive(add_cbar=True)
```
image

adonath: Thanks for check! Indeed we should require the oversampling to avoid the artefacts. As an educated guess, I'd say an oversampling of >2 should be sufficient. So +1 to just include the check and document it.

LauraOlivera:
I added another method `convolve_full` (not convinced about the name because it also works for not full-sky maps so suggestions welcome). What this does is extract the radial profile from the input kernel (assuming radial symmetry), convert it into a window beam (like in PR #3314) and convolve using `healpy.sphtfunc.smoothing`. If the input map is not a full-sky map, it first inserts the smaller map into a full sky map, and then extracts the same region once the convolution is over.

Because the convolution happens in a full sky map, the performance is worse if the pixelization is fine. That is why, even if the input map is "small", for a usual analysis pixel size (nside = 1024 is the standard for HAWC, which corresponds to a 0.025deg pixel size), this method is much slower than `convolve_wcs` even if the intermediate WCS geometry has smaller pixels by a factor >2:

```
energy = MapAxis.from_bounds(1,100, unit='TeV', nbin=2, name='energy_true')
nside = 1024
hpx_geom = HpxGeom.create(nside=nside,
                         axes=[energy],
                         region='DISK(0,0,3.5)',
                         )
hpx_map = Map.from_geom(hpx_geom)
hpx_map.set_by_coord((0,0, [2,90]), 1)
wcs_geom = WcsGeom.create(width=5,
                          binsz = 0.01, 
                         axes = [energy])
psf = PSFMap.from_gauss(
    energy_axis_true=energy, sigma=[0.3, 0.2] * u.deg
)
kernel = psf.get_psf_kernel(geom=wcs_geom, max_radius=1*u.deg)

#### Using the WCS projection
%%time 
convolved_map_wcs = hpx_map.convolve_wcs(kernel)
CPU times: user 36.7 ms, sys: 30 µs, total: 36.8 ms
Wall time: 35.8 ms

#### Using the full-sky convolution
%%time 
convolved_map = hpx_map.convolve_full(kernel)
CPU times: user 1min 56s, sys: 3.22 s, total: 2min
Wall time: 10.9 s
```
So my suggestion is to implement now the method `HpxNDMap.convolve()` which essentially decides which of the two are called based on the size of the input map. 

I also compared the maps produced above with one another and this is the result of doing `convolved_map - convolved_map_wcs`
image

adonath:
Thanks a lot for investigating @LauraOlivera! From your results it seems, there is hardly any situation where `.convolve_all()` would be useful, except if it is really an all sky map to be convolved. For the npred evaluation we basically only need the local convolution, because source models are evaluated on small cutouts only. I'll proceed with a short code review and then we discuss in there afternoon call, how to support the different convolve methods (i.e. whether to use a single `.convolve()` method etc.)

gammapy/maps/hpxnd.py
```diff
@@ -452,6 +457,130 @@ def smooth(self, width, kernel="gauss"):
...
+        Returns
+        -------
+        map : `HpxNDMap`
+            Convolved map.
+        """
+        if self.geom.width > 10*u.deg:
+            log.warning("Convolution via WCS projection is not recommended for large maps. "\
+                "Perhaps the method `convolve_full()` is more suited for this case.")
+        wcs_geom = kernel.psf_kernel_map.geom.to_image()
+        wcs_size = np.min(wcs_geom._cdelt)
```
adonath, line 492: There is `WcsGeom.pixel_scales` as well. It might be better to be conservative here and choose `np.max()`

gammapy/maps/hpxnd.py
```diff
@@ -452,6 +457,130 @@ def smooth(self, width, kernel="gauss"):
...
+        convolves via `hp.sphtfunc.smoothing`. Since no projection is applied, this is
+        suited for full-sky and large maps.
+
+        If the kernel is two dimensional, it is applied to all image planes likewise.
+        If the kernel is higher dimensional it must match the map in the number of
+        dimensions and the corresponding kernel is selected for every image plane.
+
+        Parameters
+        ----------
+        kernel : `~gammapy.irf.PSFKernel`
```
adonath, line 520: In general the radial profile is stored directly in a `PSFMap`, so we could consider supporting a `PSFMap` instead of `PSFKernel`...but only if we decide the `.convolve_full()` is useful to keep.

adonath (changes requested):
Thanks @LauraOlivera, for now I just left two minor comments. Let's discuss in the next co-working meeting on the details...

gammapy/maps/hpxnd.py
LauraOlivera: oops you're right, that is what I intended to do - bad brain. I'll fix it

gammapy/maps/hpxnd.py
LauraOlivera: Yes, I was a bit annoyed to be getting the radial profile when the kernel was built from a radial profile... it is obviously more efficient to skip the kernel computation completely. But I wasn't sure if that would work/be worth it for this case, so kept the kernel for the time being.

adonath:
@LauraOlivera To keep the scope of this PR limited maybe just implement `.convolve()` with the WCS method here? Than we could go ahead already and make `MapDataset.npred()` work. The `.convolve_all()` could be added in a follow up PR, that can be used to evaluate the other method and provide better information on the validity of the WCS approximation...

adonath (approved):
Thanks a lot @LauraOlivera! I think the code could need some polishing, however it's not a priority right now. Let's merge the PR and I'll prepare a follow up PR to do the clean up, while you and @vikasj78 can continue to work on the npred computation for HEALPix...

PR #3326, opened 2021-04-29
Add simple installation instructions using conda

maxnoe:
Add simplified instructions for just installing `gammapy` into the current conda environment to the `quickstart` docs.

### DIFF
--- a/docs/getting-started/quickstart.rst
+++ b/docs/getting-started/quickstart.rst
@@ -11,7 +11,15 @@ Installation using Anaconda
 
 The easiest and recommended way to install Gammapy is to install the Anaconda
 distribution from https://www.anaconda.com/download/ and then to install
-Gammapy and it's dependencies by executing these commands in a terminal:
+Gammapy and it's dependencies by executing this command in a terminal:
+
+.. code-block:: bash
+
+    conda install -c conda-forge gammapy
+
+
+Alternatively, we also provide an environment definition file, so you can
+get additional useful packages together with gammapy in an isolated environment:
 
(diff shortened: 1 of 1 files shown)


adonath (approved):
Thanks @maxnoe, no comments from my side. It's a useful addition...

registerrier:
Shouldn't we keep the recommended installation with the environment file?
So the direct conda install might come afterwards.
Most users are not very familiar with environments.

maxnoe:
I would argue the opposite, since most users are not familiar with environments, gammapy shouldn't force users to create one, so just installing gammapy into the current environment (most probably base) is the easiest method of getting gammapy.

"quickstart" is not really the place for arguing about the benefits of environments, right?

adonath:
Sorry @registerrier, for the quick decision here. When I restructured the install docs some time ago with @Bultako, I was thinking the same as @maxnoe. I think it is much more common for other packages as well to just provide the install instructions with `pip` and `conda` and not force users to creating the environment. However it is certainly a good practice to work with environments, so providing users with a low level entry to our environments and describe this as a next step is a good compromise I think.

Bultako:
I would propose then just to move the part speaking about installation with a new environment 
```
Alternatively, we also provide 
[...] 
[...] 
[...] 
base environment, which does not have Gammapy installed.
```

to a new paragraph in the [Installation](https://docs.gammapy.org/dev/getting-started/install.html) page, labelled "Using  versioned environments". I agree "quickstart" should be quick. I can do the modifs if you agree

PR #3328, opened 2021-04-29
Make FluxPoints inherit from FluxEstimate

adonath:
**Description**

This pull request make the `FluxPoints` object inherit from the `FluxEstimate`. It includes the following changes:
- Make the `FluxPoints` object inherit from the `FluxEstimate`
- Introduce `FluxPoints.from_table(table, sed_type=)`, which is equivalent to `FluxMaps.from_dict()`. It allows to create the internal norm representation from a given sed type
- Introduce `FluxPoints.to_table(sed_type=)`, which allows to create a table with any given sed type
- Introduce `FluxPoints.plot(sed_type=)` and `FluxPoints.plot_ts_profiles(sed_type=)`, which allows to plot any sed type, without the need to convert the flux points.
- Removal of `FluxPoints.to_sed_type()`, which becomes obsolete
- Add `SourceCatalogObject.flux_points_table`, to still have direct access to the flux points data
- Related general code cleanup and simplifications

In general, what might be confusing to users is that they have to go though `FluxPoints.to_table()` to go to any "readable" representation of the data. Directly accessing `FluxPoints.table` only shows the internal representation. However the benefit is that we have a uniform internal representation of the data and users can directly interact with the `FluxPoints`, not caring about the underlying sed type. For interactions like plotting and reading tables, they can just go to any desired sed type.

**ToDo**
Adapt tutorials after some other PRs have been merged.


registerrier (commented):
Thanks @adonath.  This looks really promising.

I think we have an issue with the `reference_spectral_model`. It was already a problem working with maps. I think it is worse here.

The issue is of course that the fluxes can be changed on the fly by changing the `reference_spectral_model` parameters.

It is also problematic to have different content after serialization.

So should we create the `TemplateSpectralModel` on `__init__`? Or do we go back and don't rely on a model but simply store all the reference values for all sed types in the `_data`?

docs/tutorials/analysis/1D/spectral_analysis.ipynb
```diff
@@ -549,11 +549,10 @@
    "outputs": [],
    "source": [
     "plt.figure(figsize=(8, 5))\n",
-    "flux_points.table[\"is_ul\"] = flux_points.table[\"ts\"] < 4\n",
-    "ax = flux_points.plot(\n",
-    "    energy_power=2, flux_unit=\"erg-1 cm-2 s-1\", color=\"darkorange\"\n",
-    ")\n",
-    "flux_points.to_sed_type(\"e2dnde\").plot_ts_profiles(ax=ax);"
+    "flux_points.table[\"is_ul\"] = flux_points.ts < 4\n",
```
registerrier, line 552: Shouldn't we get rid of this approach? A minima we have `sqrt_ts` which contains the sign, it is a better choice to use it once we have negative amplitudes.
If we define a level of significance for UL I suppose we might use it as the threshold to consider the point as an upper limit by default.

docs/tutorials/starting/overview.ipynb
```diff
@@ -720,7 +720,7 @@
    "outputs": [],
    "source": [
     "ax = crab_3fhl_spec.plot(energy_range=[10, 2000] * u.GeV, energy_power=2)\n",
-    "crab_3fhl.flux_points.to_sed_type(\"dnde\").plot(ax=ax, energy_power=2);"
+    "crab_3fhl.flux_points.plot(ax=ax, sed_type=\"dnde\", energy_power=2);"
```
registerrier, line 723: Maybe directly plot to `e2dnde`

docs/tutorials/analysis/1D/spectral_analysis.ipynb
adonath: I fully agree, the question is just where to keep the information on the sigma for ULs and errors. In a sense it maybe directly belongs to the `FluxEstimate` object as meta data. For `FluxPoints` it could be stored in the `table.meta` but for `FluxMaps` it is less clear...

adonath:
I think keeping a reference model a `SpectralModel` instance is the most flexible solution. I see that users can break the computation by modifying parameter values. But that's just how it is in Python...the alternative solution is indeed to store the reference values independently e.g. as table for `FluxMaps` as well.

What I already tried locally was to implement a `SEDTemplateSpectralModel`, which allow to handle `dnde`, `flux` and `eflux` separately and this solve the problem of the serialisation and supporting the GADF format directly. I think I can easily open a follow up PR to include this.

adonath:
As discussed today in the co-working meeting: the alternative solution here is to remove the reference spectral model and always keep the `ref_dnde` etc. as table columns and full maps. In the case of maps it is non ideal, because the reference values are spatially constant, however for consistency it is the better solution and opens up the possibility to have spatially varying reference fluxes as well (e.g. if the spectral index varies or similar...)

adonath:
To avoid the PR growing too large, I'll merge this as is and address the remaining comment by @registerrier and the comments summarised in https://github.com/gammapy/gammapy/pull/3328#issuecomment-834356142 in a follow up PR.

PR #3330, opened 2021-04-30
Add min counts options and scale error to FoVBackgroundMaker

AtreyeeS:
This partly solve #3311

- Adds error on the background norm
- Returns the exclusion mask on interpolated geometry
- Allows the user to pass the minimum number of counts/background counts required for scaling.

The ideal option would be to have provenance info as @adonath mentioned, but this can be a temporary solution

### DIFF
--- a/gammapy/makers/background/fov.py
+++ b/gammapy/makers/background/fov.py
@@ -2,9 +2,11 @@
 """FoV background estimation."""
 import logging
 import numpy as np
+from gammapy.maps import Map
 from gammapy.modeling import Fit
 from gammapy.modeling.models import FoVBackgroundModel, Model
 from ..core import Maker
+
 
 __all__ = ["FoVBackgroundMaker"]
 
--- a/gammapy/makers/background/fov.py
+++ b/gammapy/makers/background/fov.py
@@ -1,7 +1,7 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 """FoV background estimation."""
 import logging
-from gammapy.datasets import Datasets
+import numpy as np
 from gammapy.modeling import Fit
 from gammapy.modeling.models import FoVBackgroundModel, Model
 from ..core import Maker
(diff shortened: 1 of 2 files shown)


gammapy/makers/background/fov.py
```diff
@@ -31,20 +36,30 @@ class FoVBackgroundMaker(Maker):
     spectral_model : SpectralModel or str
         Reference norm spectral model to use for the `FoVBackgroundModel`, if none is defined
         on the dataset. By default, use pl-norm.
+    min_counts : float
```
adonath, line 39: Strictly speaking this would be an integer number...

gammapy/makers/background/fov.py
```diff
@@ -31,20 +36,30 @@ class FoVBackgroundMaker(Maker):
...
+        Minimum number of counts required outside the exclusion region
+    min_bkg : float
```
adonath, line 41: Maybe use explicitly `min_npred_background`, so it's compatible with the names we have for `MapDataset`

gammapy/makers/background/fov.py
```diff
@@ -90,6 +107,21 @@ def make_default_fov_background_model(self, dataset):
 
         return dataset
 
+    def make_exclusion_mask(self, dataset):
+        """Project input exclusion mask to dataset geom
+
```
adonath, line 112: Add `Parameters` section...

gammapy/makers/background/fov.py
```diff
@@ -90,6 +107,21 @@ def make_default_fov_background_model(self, dataset):
...
+        Returns
+        -------
+        mask : `~gammapy.maps.WcsNDMap`
+            Projected exclusion mask
+        """
+        geom = dataset._geom
+        if self.exclusion_mask:
+            mask = self.exclusion_mask.interp_to_geom(geom=geom)
+        else:
+            mask = Map.from_geom(geom=geom, data=1)
```
adonath, line 122: Use `dtype=bool` here...

adonath (changes requested):
Thanks @AtreyeeS, I have left a few minor inline comments...

gammapy/makers/background/fov.py
AtreyeeS: Ok changed. I had kept it float because we do not enforce the requirement of `int` anywhere. This `min_counts` requirement can come as a result of some calculation, which can turn out to be float...

AtreyeeS:
Thanks @adonath ! addressed!

gammapy/makers/background/fov.py
adonath: In practice it doesn't matter anyway. However it might be good to mention whether it's greater or greater equal...

PR #3331, opened 2021-05-03
Add Npred formula to dataset docs

adonath:
**Description**

I noticed we don't show the standard formula of our npred data model. So this pull request add the formula how the predicted counts are computed to the datasets docs page. Another good place would the analysis overview page, but for now it does not matter this much. We can move around the content later...

### DIFF
--- a/docs/datasets/index.rst
+++ b/docs/datasets/index.rst
@@ -64,10 +64,11 @@ The total number of predicted counts from a `MapDataset` are computed per bin li
 
 	N_{Pred} = N_{Bkg} + \sum_{Src} N_{Src}
 
-Where :math:`N_{Bkg}` is the expected residual hadronic background model and :math:`N_{Src}`
-the predicted counts from a given source model component.
-
-The predicted counts from a source are given by:
+Where :math:`N_{Bkg}` is the expected counts from the residual hadronic background
+model and :math:`N_{Src}` the predicted counts from a given source model component.
+The predicted counts from the hadronic background are computed directly from
+the model in reconstructed energy and spatial coordinates, while the predicted counts
+from a source are obtained by forward folding with the instrument response:
 
 .. math::
 
--- a/docs/datasets/index.rst
+++ b/docs/datasets/index.rst
@@ -52,9 +52,27 @@ Note that in Gammapy, 2D image analyses are done with 3D cubes with a single
 energy bin, e.g. for modeling and fitting,
 see the `2D map analysis tutorial <./tutorials/image_analysis.html>`__.
 
-
 To analyse multiple runs, you can either stack the datasets together, or perform
 a joint fit across multiple datasets.
+
+
+Predicted counts
+================
+Predicted counts are computed like:
+
+.. math::
+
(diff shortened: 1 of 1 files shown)


docs/datasets/index.rst
```diff
@@ -52,10 +52,28 @@ Note that in Gammapy, 2D image analyses are done with 3D cubes with a single
...
+Predicted counts
+================
+Predicted counts are computed like:
+
+.. math::
+
+	N_{Pred} = N_{Bkg} + \mathrm{PSF} \circledast \mathrm{EDISP(\mathcal{E} \cdot F_{Src}(l, b, E_{True}))}
+
+
+Where :math:`$N_{Bkg}$` is the expected residual hadronic background,
```
registerrier, line 68: isn't the bkg also predicted? It is usually a model. We have a npred_background.

docs/datasets/index.rst
```diff
@@ -52,10 +52,28 @@ Note that in Gammapy, 2D image analyses are done with 3D cubes with a single
...
 see the `2D map analysis tutorial <./tutorials/image_analysis.html>`__.
 
-
 To analyse multiple runs, you can either stack the datasets together, or perform
 a joint fit across multiple datasets.
 
+
+Predicted counts
+================
+Predicted counts are computed like:
```
registerrier, line 61: For each bin in measured position and energy?

docs/datasets/index.rst
```diff
@@ -52,10 +52,28 @@ Note that in Gammapy, 2D image analyses are done with 3D cubes with a single
...
+
+.. math::
+
+	N_{Pred} = N_{Bkg} + \mathrm{PSF} \circledast \mathrm{EDISP(\mathcal{E} \cdot F_{Src}(l, b, E_{True}))}
+
+
+Where :math:`$N_{Bkg}$` is the expected residual hadronic background,
+:math:`$F_{Src}$` the integrated flux of the source model,
```
registerrier, line 69: Normally it should be differential no?

registerrier (approved):
Thansk @adonath !

You might add the integral (or the sum if you want to represent the discrete integral) over `Etrue`

AtreyeeS:
Thanks @adonath ! Calculation of `npred` is partly explained in IRF theory. https://docs.gammapy.org/dev/irf/theory.html#irf-theory
Do you also want to describe it here?

docs/datasets/index.rst
adonath: Hm, the spectral model is integrated in energy and the spatial model at least integrated zero order by multiplying with the solid angle...I think one can keep the statement like this.

adonath:
Thanks @registerrier  and @AtreyeeS. I decided to split the formula and make the sum over the source components explicit. I think having this high level description of the predicted counts is still useful, however I also added a link to "IRF theory", where the detailed expressions and integrals are given.

AtreyeeS:
This looks nice now. Do you also want to explicitly mention that the `N_bkg` is not obtained by convolution with the IRFs?

adonath:
I'll proceed and merge this. I'm happy to follow up on it in other PRs...

PR #3335, opened 2021-05-03
Changed p-value for CountsStatistic

registerrier:
**Description**

This pull request modifies the calculation of the p-value for the CountsStatistics. It considers only positive fluctuations and hence divide the chi-square p-value by two.

The docs are updated to mention that the p-value now considers only positive fluctuations.
 
Is it OK with you @bkhelifi, @tstolarczyk ?

**Dear reviewer**


adonath:
@registerrier Maybe it's even useful to keep both p-values? One and two sided?

adonath (approved):
Thanks @registerrier, no comments from my side...

PR #3336, opened 2021-05-04
Keep infinite values in sparse wcs map serialisation

### LINKED ISSUE #2465: Fix test_wcs_nd_map_data_transpose_issue
I think here we forgot to pass `sparse=True` in `test_wcs_nd_map_data_transpose_issue`?

https://github.com/gammapy/gammapy/pull/1346/files#diff-930fa0f94b49b5d6537396fad7c6512aR137-R140

If I change to `sparse=True` in the write call, I get this fail:
```
E           Arrays are not equal
E           
E           x and y nan location mismatch:
E            x: array([[0., 1., 2.],
E                  [0., 0., 0.]], dtype=float32)
E            y: array([[  0.,   1.,   2.],
E                  [ nan,  inf, -inf]])
```

@adonath - Is there a bug?

Should we spend 1-2 days on maps basics and sort out dtype and missing value issues (#1440, #2374, maybe this issue), e.g. with pair-coding & debugging? Or defer all or some to v1.1?

This is a reminder issue, noticed this accidentally while working on
https://github.com/gammapy/gammapy/pull/2350#issuecomment-542754564

adonath:
**Description**

This pull request fixes #2465, by keeping the infinite values during the serialisation. For the sparsification only values equivalent to zero are skipped.

### DIFF
--- a/gammapy/maps/tests/test_wcsnd.py
+++ b/gammapy/maps/tests/test_wcsnd.py
@@ -149,7 +149,7 @@ def test_wcs_nd_map_data_transpose_issue(tmp_path):
     assert_equal(m2.data, data)
 
     # Data should be unmodified after write / read to sparse image format
-    m.write(tmp_path / "sparse.fits.gz")
+    m.write(tmp_path / "sparse.fits.gz", sparse=True)
     m2 = Map.read(tmp_path / "sparse.fits.gz")
     assert_equal(m2.data, data)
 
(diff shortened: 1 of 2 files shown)


registerrier (approved):
Thanks @adonath. This looks good.

Is the functionality used now?

adonath:
@registerrier Not in Gammapy, as far as I know...I think the change should be backwards compatible, but it might increase the file size in some cases...

PR #3338, opened 2021-05-04
Add fixed_offset attribute to SafeMaskMaker

fabiopintore:
This PR fixes the issue #2704 . It adds a `fixed_offset` attribute to `SafeMaskMaker` to evaluate the safe energy mask from energy dispersion bias or from effective area maximum value at a given offset with respect to the pointing position.

### DIFF
--- a/gammapy/makers/safe.py
+++ b/gammapy/makers/safe.py
@@ -36,7 +36,7 @@ class SafeMaskMaker(Maker):
         Position at which the `aeff_percent` or `bias_percent` are computed. By default,
         it uses the position of the center of the map.
     fixed_offset : `~astropy.coordinates.Angle`
-        offset, calculated from the coordinates of the center of the map, at which 
+        offset, calculated from the pointing position, at which 
         the `aeff_percent` or `bias_percent` are computed.
     offset_max : str or `~astropy.units.Quantity`
         Maximum offset cut.
--- a/gammapy/makers/safe.py
+++ b/gammapy/makers/safe.py
@@ -56,7 +56,7 @@ class SafeMaskMaker(Maker):
         methods=("aeff-default",),
         aeff_percent=10,
         bias_percent=10,
-        position=None,
+        position = None,
         fixed_offset=None,
         offset_max="3 deg",
     ):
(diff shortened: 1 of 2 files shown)


gammapy/makers/safe.py
```diff
@@ -66,6 +70,7 @@ def __init__(
         self.aeff_percent = aeff_percent
         self.bias_percent = bias_percent
         self.position = position
+        self.fixed_offset = fixed_offset
```
adonath, line 73: I think `position` and `fixed_offset` are mutually exclusive, so maybe add a check here in `__init__`

gammapy/makers/safe.py
```diff
@@ -134,6 +139,10 @@ def make_mask_energy_aeff_max(self, dataset):
         """
         geom = dataset._geom
 
+        if self.fixed_offset:
+            self.position = geom.center_skydir.directional_offset_by(position_angle=0.*u.deg,
```
adonath, line 143: Don't set `self.position` here, but just define `position = ...`. I think we should in fact use the offset from the pointing post and not the the offset from the map center...

adonath (changes requested):
Thanks @fabiopintore, I have left some comments.

gammapy/makers/safe.py
fabiopintore: Ok. I've added a condition to check that the two arguments are not passed together.

gammapy/makers/safe.py
```diff
@@ -35,6 +35,9 @@ class SafeMaskMaker(Maker):
     position : `~astropy.coordinates.SkyCoord`
         Position at which the `aeff_percent` or `bias_percent` are computed. By default,
         it uses the position of the center of the map.
+    fixed_offset : `~astropy.coordinates.Angle`
+        offset, calculated from the coordinates of the center of the map, at which
```
adonath, line 39: Adapt docstring here: offset from pointing position...

adonath (changes requested):
Thanks @fabiopintore, I have left one minor comment on the docstring...

gammapy/makers/safe.py
fabiopintore: Done.

adonath (approved):
Thanks @fabiopintore! No further comments from my side...

PR #3339, opened 2021-05-04
Adds Dockerfile to improve reproducibility

Bultako:
This PR adds a `Dockerfile` so that a docker container may be automatically built and stored in the [Gammapy Dockerhub](https://hub.docker.com/repository/docker/gammapy/gammapy) for future tagged releases. In addition, there is also documentation on how to work with this docker container.


adonath:
Thanks @Bultako! Done still need the old docker? Probably not...

Bultako:
> Thanks @Bultako! Do we still need the old docker? Probably not...

No, we do not need it, not only because we do not use Travis for the CI but also I have an error when trying to run that  `Dockerfile`. Do you agree to remove the `dev/docker` folder?

adonath:
@Bultako Yes, please remove!

Bultako:
Ok, done !

Dockerfile
```diff
@@ -0,0 +1,29 @@
+# gammapy docker container
+FROM continuumio/miniconda3:latest
```
maxnoe, line 2: Using miniconda inside docker will result in much larger docker images than say using ubuntu 20.04, installing the system python (which is 3.8).

Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+RUN apt-get update && apt-get install -y build-essential python3-pip
```
maxnoe, line 3: For what do you need build-essential and pip when using conda?

Also, please use docker image best practices, here cleaning up after using apt: https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#run

Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+# define workdir
+WORKDIR /usr/src/app/
+
+# copy content
+COPY . /usr/src/app/
+
+# create conda env
+RUN conda install -c conda-forge mamba
+RUN mamba env create -n gammapy -f environment-dev.yml
+RUN echo 'source activate gammapy' >> ~/.bashrc
```
maxnoe, line 14: This runs should also be combined into a single run statement and `conda clean --all --yes` called at the end.

Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+ENV PATH /opt/conda/envs/gammapy/bin:$PATH
+
+# install gammapy
+RUN pip install -e .
```
maxnoe, line 18: If this is for reproducibility, wouldn't it make more sense to just install a specific version of gammapy and specific versions of all needed packages instead of doing a development install?

Dockerfile
Bultako: The installation process relies on building previously a conda environment described in the `environment-dev.yml`, so `conda` command is needed.

Dockerfile
Bultako: You're right, we can remove `build-essential`, but `pip` is needed to do a development install, as justified in comments below..

Dockerfile
Bultako: ok, done ! thanks !

Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+
+# download datasets
+RUN gammapy download datasets
```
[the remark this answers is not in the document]
Bultako: The datasets size is 192MB out of 3GB of the total image size. We ship the datasets within the docker image together with the Gammapy code and tutorials, so to provide the possibility to play with the tutorials in shipped Jupyterlab. See [here](https://github.com/gammapy/gammapy/pull/3339/files#diff-53d09420b94e4eb52ff6c09b64cfea1439caaa8f2dd67c796cc48e9958500e89R32).

Dockerfile
maxnoe: But pip comes with conda, no  need to install pip into the system

Dockerfile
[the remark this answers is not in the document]
maxnoe: The image is 3Gb??? That is HUGE

maxnoe:
I think we should discuss a bit what the use case for this docker container should be.

If the use case is to provide an exactly reproducible software environment for analysis, the goal should be to make this image as small as possible so it can be used quickly on clusters and the like.

I don't really see a usecase for providing a docker image to run tutorials. This is an anti-pattern. People wanting to try out gammapy and do the tutorials should not have a need to do that in a docker container. It adds many additional hurdles and doesn't solve a problem other than "My python setup is messed up". The chances to "mess up" only increase with docker.

So if the goal is to provide an containerized environment for running  analyses reproducibly, no data should be added and only the bare depedencies, e.g. not jupyter lab.

Dockerfile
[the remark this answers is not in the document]
adonath: This is interesting, I have never checked what the actual size of the dev environment is. Do you have any idea why it is so large?

Bultako:
As I have understood it from internal discussions, the use case is mostly archive oriented, in the sense of archiving an exact replica of the development environment, code and datasets used at the moment of a release. So not only to provide a reproducible software but also the exact environment **used** to develop it at that time. 

Sure, you're very welcome to contribute to this discussion internally :)

Dockerfile
Bultako: Yes, but I have an error when running `pip install -e.` in the docker that does not happen if I install pip with apt-get...

maxnoe: Then you are installing gammapy into the system, not into the conda env

Dockerfile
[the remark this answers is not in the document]
maxnoe: Conda envs get huge fast since they duplicate many system libraries and include e.g. the mal by default. 

Adding nomkl reduces conda env quite a bit.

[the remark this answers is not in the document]
Bultako: just for info

```
$ du -hs /opt/miniconda3/envs/gammapy-0.18.2
1,2G	/opt/miniconda3/envs/gammapy-0.18.2
```

maxnoe:
Just for comparison, a Docker image starting from `python:3.9-slim-buster` and only installing gammapy is "only" 513 MB large:

```dockerfile
❯ cat Dockerfile
FROM python:3.9-slim-buster

RUN apt update && apt install --yes --no-install-recommends \
	gcc \
	libc-dev

RUN useradd -m gammapy
USER gammapy

# region 0.4 does not provide a pyproject.toml or a wheel and
# needs numpy, cython to be present at install time
# this is fixed in current master, so this extra command should be removed
# when the next version of regions is released.
ENV PATH="${HOME}/.local/bin:${PATH}"
RUN python3 -m pip install --no-cache-dir Cython numpy \
	&& pip install --no-cache-dir gammapy \
	&& rm -rf /home/gammapy/.cache/pip

CMD ["/bin/bash", "-l"]
```

which could potentially be slightly smaller with using a build stage (no compilers / dev packages) in the final image.


And one with all optional dependencies  including ipython/jupyterlab and a build stage to further shrink the image size by not having the build requirements in the runtime  is ~700 MB, which is still a lot but considerably smaller than the 3 GB using conda.

conda is a good (most of the time) tool to get binary packages without too much hassle. But docker provides the reproducibility that makes building the binaries a non-issue, most of the packages come as wheels anyway,
the only exceptions are regions and sherpa.


Docker image with all deps and using a build stage:

```dockerfile
FROM python:3.9-slim-buster AS BUILDER

# gcc, libc-dev needed to build regions and sherpa. g++, bison, flex, make needed for sherpa
RUN apt update && apt install --yes --no-install-recommends \
	gcc \
	g++ \
	libc-dev \
	bison \
	flex \
	make

RUN useradd -m gammapy
USER gammapy

# region 0.4 does not provide a pyproject.toml or a wheel and
# needs numpy, cython to be present at install time
# this is fixed in current master, so this extra command should be removed
# when the next version of regions is released.
RUN python3 -m pip install --no-cache-dir --no-warn-script-location Cython numpy \
	&& pip install --no-cache-dir --no-warn-script-location \
		gammapy \
		ipython \
		jupyterlab \
		matplotlib \
		pandas \
		healpy \
		iminuit \
		sherpa \
		naima \
		emcee \
		corner \
		parfive \
	&& rm -rf /home/gammapy/.cache/pip


# second stage to save space
FROM python:3.9-slim-buster

RUN useradd -m gammapy
USER gammapy
WORKDIR /home/gammapy
ENV PATH="/home/gammapy/.local/bin:${PATH}"

COPY --from=BUILDER /home/gammapy/.local /home/gammapy/.local
```


Image sizes:

```
gammapy-all                                          latest                 73993f41e023   About a minute ago   708MB
gammapy                                              latest                 32420a39fd95   39 minutes ago       513MB
```

adonath:
The docker image is not a priority now, I will postpone this to after v1.0.

morcuended:
Here is an updated version of @maxnoe proposal using two stages.

First a build stage to build a wheel, which is later copied and install in the second stage to minimize image size. This is mostly needed because setuptools_scm needs the full version info from git and git itself, but including that in the final image would bloat its size.

```dockerfile
# Python image with tools needed to build and compile python packages
FROM python:3.12 AS builder

# add necessary sources, including .git for version info
COPY . /repo/

# Build Gammapy wheel
RUN python -m pip install --no-cache-dir build \
    && cd repo \
    && python -m build --wheel

# Second stage, copy and install wheel using official python image
# in the slim variant to reduce image size.
FROM python:3.12-slim
COPY --from=builder /repo/dist /tmp/dist

RUN python -m pip install --no-cache-dir /tmp/dist/* \
    && rm -r /tmp/dist

RUN useradd --create-home --system --user-group gammapy
USER gammapy
```

registerrier:
This PR is superseded by #6446 . Closing now.

PR #3342, opened 2021-05-05
Add truncation value to npred in cash statistics calculation

registerrier:
**Description**

This pull request introduces a truncation value for the cash statistics to avoid convergence issues when npred goes to zero.
How should we share the truncation value between the python and the python implementation?

**Dear reviewer**

We should probably adapt cstat/wstat as well to keep the same behavior.

It changes a few things in the `CashCountsStatistics`, the negative error calculation was incorrect before the truncation because cash became 0 when mu_bkg was negative . 
There is also an issue with the excess_matching significance, because the newton algorithm probably does not work well when the slope becomes extremely steep 

Opinions @adonath , @QRemy ?

### DIFF
--- a/gammapy/stats/fit_statistics.py
+++ b/gammapy/stats/fit_statistics.py
@@ -8,8 +8,6 @@ from gammapy.stats.fit_statistics_cython import TRUNCATION_VALUE
 
 __all__ = ["cash", "cstat", "wstat", "get_wstat_mu_bkg", "get_wstat_gof_terms"]
 
-N_ON_MIN = 1e-25
-
 def cash(n_on, mu_on, truncation_value=TRUNCATION_VALUE):
     r"""Cash statistic, for Poisson data.
 
@@ -30,6 +28,7 @@ def cash(n_on, mu_on, truncation_value=TRUNCATION_VALUE):
     truncation_value : array_like
         Minimum value use for ``mu_on``
         ``mu_on`` = ``truncation_value`` where ``n_on`` <= ``truncation_value``
+        Default is 1e-25.
 
     Returns
     -------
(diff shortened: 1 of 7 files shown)


adonath:
@registerrier I guess we could change the `newton` to `brentq` as well...

adonath:
> How should we share the truncation value between the python and the python implementation?

One could e.g. just import from the cython file...

gammapy/stats/fit_statistics_cython.pyx
```diff
@@ -22,13 +21,21 @@ def cash_sum_cython(np.ndarray[np.float_t, ndim=1] counts,
...
-        if npred[i] > 0:
-            sum += npred[i]
-            if counts[i] > 0:
-                sum -= counts[i] * log(npred[i])
+        npr = npred[i]
+        if npr > 0:
+            lognpr = log(npr)
+        else:
+            npr = 1e-25
+            lognpr = -57.56
```
QRemy, line 33: It should be :
```
if npred < truncation_value:
    npr  = truncation_value
lognpr = log()
```
no ?

 Wouldn't be simpler to just avoid fittting negative norm/amplitude by default and do it only for the relevant cases ?

adonath:
@registerrier Here is one remaining doc test fail: https://github.com/gammapy/gammapy/pull/3342/checks?check_run_id=2512363046#step:7:41 I think in this case it would be good to format the output to 3 or less digits...

gammapy/stats/fit_statistics.py
```diff
@@ -4,13 +4,13 @@
 see :ref:`fit-statistics`
 """
 import numpy as np
+from gammapy.stats.fit_statistics_cython import TRUNCATION_VALUE
 
 __all__ = ["cash", "cstat", "wstat", "get_wstat_mu_bkg", "get_wstat_gof_terms"]
 
 N_ON_MIN = 1e-25
```
adonath, line 11: I just noticed that here we already define a truncation value n_on and cstat. Might be good to merge both...but this could be done in a follow up PR as well...

registerrier: It is done

adonath (approved):
Thanks @registerrier, looks good! I have two more general thoughts, but none of it should block merging this PR:
- In general the truncation value could be just derived from the `dtype`? 
- It's not 100% clear to me whether introducing the `truncation_value` as public argument is useful, maybe it could always remain and implementation detail...

PR #3343, opened 2021-05-05
Display progress bar in gammapy

luca-giunti:
**Description**
This PR updates and fixes the old PR #3057. Sorry for the delay :)

I tested it with the available tutorials, and it works fine for me in all cases. The only case in which I get an unwanted line break (for no good reason apparently, since there is no warning message) is for the flux point estimation.

Few things:
- As suggested @adonath , I made the progress bar hidden by default
-  Should we show this funtionality in the tutorials? I think so, maybe only in the "getting started" ones, plus those with long-running loops
- In principle using `from tqdm.auto import tqdm` instead of `from tqdm import tqdm` would be better. On my machine however I can't make it work, so I just used the basic `tqdm` version.


luca-giunti:
Sorry the latest two commits are unrelated - they are for another PR. I think I successfully reverted them

gammapy/modeling/fit.py
```diff
@@ -423,22 +430,24 @@ def stat_surface(self, x, y, x_values, y_values, reoptimize=False, **optimize_op
...
-                    fit_results.append(result)
-                else:
-                    stat = self.datasets.stat_sum()
-
-                stats.append(stat)
+        with pbar(total=len(x_values) * len(y_values), show_pbar=show_pbar, desc="Trial values") as pb:
+            with parameters.restore_status():
+                for x_value, y_value in itertools.product(x_values, y_values):
+                    # TODO: Remove log.info() and provide a nice progress bar
+                    log.info(f"Processing: x={x_value}, y={y_value}")
```
Bultako, line 437: The same as above, if we want to keep the log.info()

gammapy/analysis/core.py
```diff
@@ -331,38 +356,44 @@ def _map_making(self):
...
-                dataset = maker.run(cutout, obs)
-                dataset = maker_safe_mask.run(dataset, obs)
-                if bkg_maker is not None:
-                    dataset = bkg_maker.run(dataset)
-                log.debug(dataset)
-                datasets.append(dataset)
-
+        with pbar(total=len(self.observations), show_pbar=show_pbar, desc="Datasets") as pb:
+            if datasets_settings.stack:
+                for obs in self.observations:
```
Bultako, line 361: I think we can keep the log.info() if show_pbar is False

```
if not show_pbar:
    log.info(f"Processing observation {obs.obs_id}")
```

gammapy/analysis/core.py
```diff
@@ -331,38 +356,44 @@ def _map_making(self):
...
+                        dataset = dataset.to_map_dataset()
+
+                    log.debug(dataset)
+                    stacked.stack(dataset)
+                    pb.update(1)
+                datasets = [stacked]
+            else:
+                datasets = []
+
+                for obs in self.observations:
```
Bultako, line 378: The same as above, if we want to keep the log.info()

gammapy/analysis/core.py
```diff
@@ -411,20 +442,20 @@ def _spectrum_extraction(self):
...
-                    log.info(
-                        f"No OFF region found for observation {obs.obs_id}. Discarding."
-                    )
-                    continue
-            dataset = safe_mask_maker.run(dataset, obs)
-            log.debug(dataset)
-            datasets.append(dataset)
-
+        with pbar(total=len(self.observations), show_pbar=show_pbar, desc="Datasets") as pb:
+            for obs in self.observations:
```
Bultako, line 446: The same as above, if we want to keep the log.info()

gammapy/analysis/core.py
```diff
@@ -411,20 +442,20 @@ def _spectrum_extraction(self):
...
+                dataset = dataset_maker.run(reference.copy(), obs)
+                if bkg_maker is not None:
+                    dataset = bkg_maker.run(dataset, obs)
+                    if dataset.counts_off is None:
+                        log.info(
```
Bultako, line 451: There are two more options here:
- switch to log.debug()
- log if show_pbar is False and dataset.counts_off is None

We can also keep as it is if you consider that this info is important enough to break the progress bar display

Bultako (changes requested):
Thanks @luca-giunti 
I have left some comments in between the lines.

Bultako:
@luca-giunti 

Other change it should be done, is to move up `tqdm` in https://github.com/gammapy/gammapy/blob/master/environment-dev.yml#L33 from the `# test dedendencies` section to `# core dependencies`. This is needed in for the config of our CI in Github actions.

I have also seen that `tqdm` was already used in https://github.com/gammapy/gammapy/blob/master/gammapy/scripts/download.py#L20 as an optional dependency, so since it will no longer be an optional dependency I guess we can move the `from tqdm import tqdm` outside the try/except block.

gammapy/estimators/parameter.py
```diff
@@ -188,7 +188,7 @@ def estimate_errn_errp(self, datasets, parameter):
             f"{parameter.name}_errn": res["errn"],
         }
 
-    def estimate_scan(self, datasets, parameter):
+    def estimate_scan(self, datasets, parameter, show_pbar=False):
```
Bultako, line 191: `with pbar` is not implemented in this script, if we needed it somewhere let's add it, if not let's remove the show_pbar

luca-giunti: I think it makes sense to leave it, because it is passed to `fit.stat_profile` a few lines later. right?

gammapy/analysis/core.py
luca-giunti: I downgraded the log priority to `debug` instead. Do you think it's ok? This way, the user can still see the processed observation ID (even breaking the progress bar) if really needed.

Possibly, as suggested by @registerrier during yesterday's call, we might even show the progress bar only when the log level is set to a high verbosity. This would allow to remove the boolean `show_pbar`. I like this idea and I think it is way to go, but it requires a good cleanup of the logging that we do in gammapy. It is work for the future, I guess

gammapy/estimators/parameter.py
Bultako: yep, you're right

luca-giunti:
> Other change it should be done, is to move up tqdm in https://github.com/gammapy/gammapy/blob/master/environment-dev.yml#L33 from the # test dedendencies section to # core dependencies. This is needed in for the config of our CI in Github actions. Also add as require dependency in setup.cfg https://github.com/gammapy/gammapy/blob/master/setup.cfg#L39

> I have also seen that tqdm was already used in https://github.com/gammapy/gammapy/blob/master/gammapy/scripts/download.py#L20 as an optional dependency, so since it will no longer be an optional dependency I guess we can move the from tqdm import tqdm outside the try/except block.

@Bultako as discussed in the call, I used a similar pattern as in km3pipe. I have tried uninstalling `tqdm` and it seems to behave as expected (no crash, no progress bar)

gammapy/analysis/core.py
Bultako: >  downgraded the log priority to debug instead.

I do not understand what you mean with this, sorry..
when using the HLI the log level is [set to INFO by default](https://github.com/gammapy/gammapy/blob/master/gammapy/analysis/config.py#L212). So, all logs with level info and higher will be displayed. 

I don't think automatically showing the progress bar in debug level would be ok, on the contrary I would display progress bar if log is set to warnings, so only high severity logs will break the display.

Bultako: > downgraded the log priority to debug instead.

ok, now that I can see your modifs, I know what you mean.
fine !

gammapy/utils/pbar.py
```diff
@@ -0,0 +1,25 @@
...
+                disable=True,
+                desc=""
+        ):
+            pass
+        def update(self, x):
+            pass
+
+@contextmanager
+def pbar(total=None, show_pbar=True, desc=None):
+    if total == None and show_pbar == True:
```
Bultako, line 22: replace `if total == None` by `if total is None` since in Python `None` is not a value but an object.

gammapy/utils/pbar.py
```diff
@@ -0,0 +1,25 @@
...
+except ImportError:
+    class tqdm():
+        def __init__(
+                self,
+                total=0,
+                mininterval=0,
+                disable=True,
+                desc=""
+        ):
+            pass
```
Bultako, line 16: shorter 

```
try:
    from tqdm import tqdm
except ImportError:
    class tqdm():
        def __init__(*args, **kwargs):
            pass
        def update(self, x):
            pass
```

gammapy/modeling/fit.py
Bultako: log.debug() here, non?

Bultako (dismissed):

gammapy/estimators/asmooth_map.py
```diff
@@ -135,14 +136,15 @@ def _sqrt_ts_cube(cubes, method):
             )
         return scube
 
-    def run(self, dataset):
+    def run(self, dataset, show_pbar=False):
```
adonath, line 139: I think `show_pbar` could be rather considered as configuration and I would recommend to move it to `Estimator.__init__` consistently.

gammapy/analysis/core.py
```diff
@@ -121,16 +122,22 @@ def get_observations(self):
         for obs in self.observations:
             log.debug(obs)
 
-    def get_datasets(self):
-        """Produce reduced datasets."""
+    def get_datasets(self, show_pbar=False):
```
adonath, line 125: Maybe rename consistently to the more explicit `show_progress_bar`? I think it's not too long but much more explicit...

adonath (changes requested):
Thanks @luca-giunti, I have left some first comments, I'll try out the functionality now...

gammapy/utils/pbar.py
```diff
@@ -0,0 +1,26 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+"""Utilities for progress bar display"""
+import logging
+from contextlib import contextmanager
+
+log = logging.getLogger(__name__)
+
+try:
+    from tqdm import tqdm
```
adonath, line 9: I think importing from `tqdm.auto` would be still good. You mentioned it does not work for you...does it fail, or do you just get the "normal" console progress bar?

luca-giunti: @adonath done. I realized I was using a very old jupyter lab version (~1.0). After updating to jupyterlab 3.0, the bar is correctly rendered

adonath (approved):
Thanks @luca-giunti, no further comments from my side...

gammapy/estimators/asmooth_map.py
registerrier: To avoid replication of the argument in all `__init__` shouldn't we move it to the base class instead?

registerrier (commented):

I wonder whether the scheme where `show_progress_bar` is an argument to all functions that might use it a bit heavy in the long run. Wouldn't it be simpler to have a general setting to use it wherever possible. This would allow less intrusive code in the estimators, with e.g. decorators for the progress bar. No?

adonath:
I think @registerrier  makes good point here: maybe showing the progress bar should be considered as "global" configuration and not individual on the method / estimator level. As we don't have a global config system yet,  one way to achieve this would be to just introduce a flag like:
```python3
from gammapy.utils import pbar

pbar.SHOW_PROGRESS_BAR = True
```
This removes a lot of duplicated and boilerplate code and opens the opportunity to handle this with a global config file later.

luca-giunti:
Hi! @adonath I totally agree. I will probably not have much time to further adapt this, for a couple of weeks... When is the deadline for contributions for v1.0?

luca-giunti:
Hi @adonath, I will have some time tonight to work on this. Could you please elaborate a little bit on your last proposal? It's not clear to me how the progress bar should be handled internally. A decorator, or something else? Thanks :)

luca-giunti:
@registerrier I have tried to implement a decorator, but I run into issues. The main one is: how can we pass the `total` argument (defining the length of the progress bar, for example 100 if one uses 100 energy bins for computing flux points) to the progress bar?

Take for example the `LightCurveEstimator.run()` function. I may decorate it as
```
@progress_bar
def run(self, datasets):
...
```
but the length of the progress bar (i.e. the `total` parameter) has to be set somehow inside the `run()` function itself, as:
```
if self.time_intervals is None:
    gti = datasets.gti
else:
    gti = GTI.from_time_intervals(self.time_intervals)
gti = gti.union(overlap_ok=False, merge_equal=False)
total = len(gti.time_intervals)
```
I don't know if this can be done. Also, I don't know how to `update()` the progress bar from the decorator (right now, [I was doing it manually]( https://github.com/gammapy/gammapy/blob/423d9954ac23776de7b87a8cb28b55d7aedc2522/gammapy/estimators/lightcurve.py#L462)).  Maybe you know a better approach?

Maybe we should only decorate/wrap the `for` loops inside the `Estimator.run()`, instead of the `Estimator.run()` itself.  But I don't know a good way.

Notice that the `tqdm` developers already explored and discarded the possibility of adding a decorator (see https://github.com/tqdm/tqdm/issues/583 and https://github.com/tqdm/tqdm/pull/584).

registerrier:
OK, I see the difficulty. 

Then maybe a less intrusive solution is to go towards the solution suggested in the issue you link and simply add the `tqdm` wrapper on the loop itself. Then this changes only one line of code and can be relatively transparent.
Then it means that your pseudo `tqdm` class must contain the input iterator and be something like:
```
class tqdm():
    def __init__(self, iterable=None, disable=True, **kwargs):
        self.disable = disable
        self._iterable = iterable
        if self.disable == False:
            log.info(
                 f"Tqdm is currently not installed. Visit https://tqdm.github.io/"
             )
    def update(self, x):
        pass
    
    def __iter__(self):
        return self._iterable.__iter__()
    
    def __next__(self):
        return self._iterable.__next__()
```

Then you should be able to do:
```
  for energy in tqdm(self.energy_edges, disable=disable)
```
and this could run in a transparent manner?

luca-giunti:
@registerrier great, this looks much better to me. I will try it! Thank you

luca-giunti:
@adonath an additional question: right now the progress bar is completely undocumented. Maybe I can add a commit to show it in some tutorial? Or there is a better place

luca-giunti:
I don't really understand the new CI fails

adonath:
@luca-giunti Yes, we should definitely show the uses of this in some of our tutorials. However I would propose to do it in a follow up PR , to avoid merge conflicts with #3372, which touches many tutorials.

gammapy/utils/pbar.py
```diff
@@ -0,0 +1,36 @@
...
+log = logging.getLogger(__name__)
+
+try:
+    from tqdm.auto import tqdm
+except ImportError:
+    class tqdm():
+        def __init__(self, iterable, disable=True, **kwargs):
+            self.disable = disable
+            self._iterable = iterable
+            if self.disable == False:
```
adonath, line 15: Comparison of bools is directly done with `if self.disable:` or `if not self.disable`.

gammapy/utils/pbar.py
```diff
@@ -0,0 +1,36 @@
...
+            pass
+
+        def __iter__(self):
+            return self._iterable.__iter__()
+
+        def __next__(self):
+            return self._iterable.__next__()
+
+def progress_bar(iterable, show_progress_bar=False, desc=None):
+    if not isinstance(iterable, Iterable) and show_progress_bar == True:
```
adonath, line 30: This I don't understand: I think this should always fails if there is no iterable, not only if the progress bar shown.

gammapy/utils/pbar.py
```diff
@@ -0,0 +1,36 @@
...
+        raise AttributeError("Can't set up the progress bar if total is None")
```
adonath, line 31: I think the correct error here is `ValueError`, how does `tqdm` fail if users pass and non-iterable object?

adonath (changes requested):
Thanks @luca-giunti, I have left some more comments. I'll investigate the CI fails now..

gammapy/utils/pbar.py
luca-giunti: `TypeError: 'int' object is not iterable`

Maybe I could even remove these three lines (29-30-31), because `tqdm` will take care of it already?

luca-giunti:
@adonath @registerrier I think this should be better now. [Here](https://github.com/luca-giunti/share/blob/master/PROGRESS_BAR.mp4) you can download a short video demonstrating the usage of the progress bar for nested loops

gammapy/utils/pbar.py
```diff
@@ -0,0 +1,36 @@
...
+        def update(self, x):
+            pass
+
+        def __iter__(self):
+            return self._iterable.__iter__()
+
+        def __next__(self):
+            return self._iterable.__next__()
+
+global SHOW_PROGRESS_BAR
```
adonath, line 28: Last comment: this shouldn't be needed. Python automatically looks up a variable in the outer scope, if it is not part of the inner scope (the "progress_bar" method here)...

luca-giunti: @adonath yes indeed. I removed it

adonath (approved):
Thanks a lot @luca-giunti, no further comments from my side.

registerrier (approved):
Thanks @luca-giunti !
 No more comments form my side.

PR #3344, opened 2021-05-05
Add codemeta.json and CITATION.cff description metadata files

Bultako:
This PR adds two files with metadata related to Gammapy software, following formats described in [The CodeMeta Project](https://codemeta.github.io) and [Citation File Format](https://citation-file-format.github.io). The `codemeta.json` file has been created manually with [Codemeta generator](https://codemeta.github.io/codemeta-generator/) and the `CITATION.cff` file has been created automatically with [doi2cff](https://github.com/citation-file-format/doi2cff) from the Zenodo DOI of Gammapy with small modifs additions.


codemeta.json
```diff
@@ -0,0 +1,513 @@
...
+    "relatedLink": [
+        "https://gammapy.org",
+        "https://docs.gammapy.org",
+        "https://zenodo.org/record/4701500#.YJJbsC8lNpQ",
+        "https://ascl.net/gammapy"
+    ],
+    "author": [
+        {
+            "@type": "Person",
+            "givenName": "Christoph",
```
bkhelifi, line 39: Should we keep Christoph for V0.18.2?

codemeta.json
```diff
@@ -0,0 +1,513 @@
...
+                "name": "HeidelbergCement"
+            }
+        },
+        {
+            "@type": "Person",
+            "givenName": "Axel",
+            "familyName": "Donath",
+            "affiliation": {
+                "@type": "Organization",
+                "name": "MPIK"
```
bkhelifi, line 52: The country is not needed?

Bultako: I have not found "country" in the any of the examples provided in the CodeMeta Project website.

codemeta.json
Bultako: We could also use the following snippet for authors and contributors, more easy to maintain..

```
    "author": [
        {
            "@type": "Organization",
            "relatedLink": "https://gammapy.org"
        }
    ],
```

codemeta.json
bkhelifi: So, one can/should remove it...

codemeta.json
bkhelifi: fine with me!

bkhelifi:
Should we wait for the last minute in order to include metadata of the 0.19?

adonath:
@bkhelifi No definitely not. The remaining unresolved question was whether we should rely on https://github.com/citation-file-format/cff-converter-python/tree/1.3.3 for the conversion. I think my current preference would be to just do it...we can always add hand-maintained files later, if there is a problem with `cffconvert`.

adonath:
@Bultako I was wondering whether we can maybe get at least the `CITATION.cff` in here? Such that the GitHub citation works already? https://github.blog/2021-08-19-enhanced-support-citations-github/

Bultako:
I have removed `codemeta.json` file, so to not duplicate maintenance of this info.
We would like to automate the generation of `codemeta.json` file from content in `CITATION.cff` also including info about dependencies. This will come in another PR.

adonath (approved):
Thanks a lot @Bultako! No further comments from my side. Let's merge this in and then update the information and affiliations in follow up PRs.

Issue #3271, opened 2021-03-11
Wrong label in SpectrumDataset.plot_counts()


luca-giunti:
**Gammapy version**
v0.18.2

**Bug description**
I think that the "fit range" label [here](https://github.com/gammapy/gammapy/blob/635ff43540ab39feff9b60c6ecf9a907ae50a3c7/gammapy/datasets/spectrum.py#L74) is wrong, because the `mask_fit` is actually not applied when computing the `MapDataset.energy_range`. This method is called in the `SpectrumDataset.plot_counts()`, that is therefore showing vertical lines in the wrong place.

**Expected behavior**
I expected the vertical lines to display the fit range, not the energy range defined by the safe mask.

**Other information**
I would propose to change [this](https://github.com/gammapy/gammapy/blob/635ff43540ab39feff9b60c6ecf9a907ae50a3c7/gammapy/datasets/spectrum.py#L72) line to actually take into account the `mask_fit`.

AtreyeeS:
Even taking `MapDataset.energy_range` is not ideal behaviour I would say. Because the takes the `[energy_min[mask].min(), energy_max[mask].max()]`, which is fine if the masked bins are at the energy edges. But this breaks if some bins in the middle are masked, say `dataset.mask_safe.data[3:-3]= False`.
This might not be very common, but probably it will be good to change the plot.. maybe a light shaded background for the un-masked bins?

Regarding plotting of `mask_fit` or `mask_safe`, I remember there was already such a discussion on slack. And it seemed the preference depends on the use case.

So probably it can be an option, `MapDataset.plot_counts(apply_mask_fit=False)` like we have in the `ExcessMapEstimator`?

adonath:
@AtreyeeS +1 for the suggestion to plot masks as shaded regions! I'll think about it once more whether there is an alternative way to visualise this, but I think this definitely goes into the right direction. Maybe we can collect a few proposals here and then decide...

adonath:
Here is a quick example how @AtreyeeS 's proposal could look like:

![image](https://user-images.githubusercontent.com/3715928/111498958-bccbdb80-8742-11eb-8336-54e385d7a6c9.png)

adonath:
I think we could implement this shaded area plotting as `RegionNDMap.plot_mask()`.

adonath:
For reference here is the core of the mask plotting code:
```python3

from scipy.ndimage.measurements import label as ndi_label

[...]

PR #3345, opened 2021-05-05
Add plot_mask to RegionNDMap

fabiopintore:
This PR should solve the issue #3271. 
It adds a `plot_mask` function into `RegionNDMap`, in order to take into account the correct region of the mask fit.

### DIFF
--- a/gammapy/maps/regionnd.py
+++ b/gammapy/maps/regionnd.py
@@ -151,6 +151,11 @@ class RegionNDMap(Map):
             equivalent WCS geometry.
         **kwargs : dict
             Keyword arguments forwarded to `~regions.PixelRegion.as_artist`
+
+        Returns
+        -------
+        ax : `~matplotlib.pyplot.Axis`
+            Axis used for plotting
         """
         ax = self.geom.plot_region(ax, **kwargs)
         return ax
@@ -166,6 +171,9 @@ class RegionNDMap(Map):
             Keyword arguments passed to `~matplotlib.pyplot.axvspan`
         """
         import matplotlib.pyplot as plt
+
+        if not self.is_mask:
+            raise ValueError("This is not a mask and cannot be plotted")
 
         kwargs.setdefault("color", "k")
         kwargs.setdefault("alpha", 0.05)
--- a/gammapy/maps/regionnd.py
+++ b/gammapy/maps/regionnd.py
@@ -155,21 +155,21 @@ class RegionNDMap(Map):
         ax = self.geom.plot_region(ax, **kwargs)
         return ax
 
-    def plot_mask(self, ax=None, color="k", label=""):
-        """Plot region
+    def plot_mask(self, ax=None, **kwargs):
+        """Plot the mask as a shaded area in a xmin-xmax range
 
         Parameters
         ----------
-        ax : `~astropy.vizualisation.WCSAxes`
-            Axes to plot on. If no axes are given,
-            the region is shown using the minimal
-            equivalent WCS geometry.
-        color : str
-            color of the shaded region.
-        label : str
-            label of the shaded region.
+        ax : `~matplotlib.axis` 
+            Axis instance to be used for the plot.
+        **kwargs : dict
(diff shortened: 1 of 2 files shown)


gammapy/maps/regionnd.py
```diff
@@ -154,6 +155,37 @@ def plot_region(self, ax=None, **kwargs):
         ax = self.geom.plot_region(ax, **kwargs)
         return ax
 
+    def plot_mask(self, ax=None, color="k", label=""):
```
adonath, line 158: I would propose to use the `**kwargs` pattern here, like we do for the other plot functions. Internally we can set default values for the `kwargs` like:

```python3

def plot_something(ax, **kwargs):
    """
    Parameters
    -----------
    **kwargs : dict
        Keyword arguments passed to `~matplotlib.pyplot.axvspan` 
    """
    kwargs.setdefault("color", "k")
    kwargs.setdefault("alpha", 0.05)
    kwargs.setdefault("label", "mask")
    ax.axvspan(xmin, xmax, **kwargs)


```

gammapy/maps/regionnd.py
```diff
@@ -154,6 +155,37 @@ def plot_region(self, ax=None, **kwargs):
...
+        """Plot region
```
adonath, line 159: Maybe improve the docstring here a bit to be more explicit on plotting the mask as shaded are etc.

gammapy/maps/regionnd.py
```diff
@@ -154,6 +155,37 @@ def plot_region(self, ax=None, **kwargs):
...
+
+        Parameters
+        ----------
+        ax : `~astropy.vizualisation.WCSAxes`
+            Axes to plot on. If no axes are given,
+            the region is shown using the minimal
```
adonath, line 165: This is a copy and paste error: here the ax is not a `WCSAxes` and there is no region plotting, but only a spectrum...

adonath (changes requested):
Thanks @fabiopintore, I have left a few inline comments...

gammapy/maps/regionnd.py
fabiopintore: Done.

gammapy/maps/regionnd.py
fabiopintore: Done.

gammapy/maps/regionnd.py
fabiopintore: Ahh right, my fault! I've corrected it.

gammapy/maps/regionnd.py
```diff
@@ -154,6 +155,37 @@ def plot_region(self, ax=None, **kwargs):
...
+    def plot_mask(self, ax=None, **kwargs):
+        """Plot the mask as a shaded area in a xmin-xmax range
+
+        Parameters
+        ----------
+        ax : `~matplotlib.axis` 
+            Axis instance to be used for the plot.
+        **kwargs : dict
+            Keyword arguments passed to `~matplotlib.pyplot.axvspan`
+        """
```
adonath, line 167: Add `Returns` section here...you'll find example in the same file..

gammapy/maps/regionnd.py
```diff
@@ -154,6 +155,37 @@ def plot_region(self, ax=None, **kwargs):
...
+        import matplotlib.pyplot as plt
+
```
adonath, line 169: Include a check like `if not self.is_mask:` and raise an error to forbid the plotting of non-masks

adonath (changes requested):
Thanks @fabiopintore, I have left two final comments..

gammapy/maps/regionnd.py
fabiopintore: Ok.

gammapy/maps/regionnd.py
fabiopintore: Ok.

adonath (approved):
Thanks @fabiopintore, no further comments from my side...

PR #3346, opened 2021-05-05
Allow Npred computation for Healpix MapDataset

LauraOlivera:
This pull request adds minor fixes required for the npred calculation to work on Healpix maps. I marked it as a draft because, while getting the npred of a dataset runs without errors right now, some more testing is required to make sure that the PSF convolution is actually done correctly. I will update as I investigate further.


adonath (approved):

PR #3347, opened 2021-05-05
Simplify and refactor installation instructions

Bultako:
This PR is a continuation of #3326

It moves the install instructions related to versioned conda environments from [Quickstart](https://docs.gammapy.org/dev/getting-started/quickstart.html) to a new section "Using versioned environments" in [Installation](https://docs.gammapy.org/dev/getting-started/install.html) It also removes the info related to versioned envs that is displayed when `gammapy download notebooks` if the notebooks version downloaded is the same as the gammapy version installed.


Bultako:
Ok, I see as concrete actions:

- Rephrase the text "The easiest and recommended way to install Gammapy" -> "The easiest way to install Gammapy"
- "Alternatively, we also provide" -> "Alternatively, we recommend to make use of an environment definition file that we provide, so you can get useful packages"
- Reformulate content in new section  "Using versioned environments" added in this PR for the Installation page, explaining the more general use of `conda` environments and Python's `venv`.

docs/getting-started/quickstart.rst
```diff
@@ -9,17 +9,18 @@ Quickstart
...
 
     conda install -c conda-forge gammapy
 
-
-Alternatively, we also provide an environment definition file, so you can
-get additional useful packages together with gammapy in an isolated environment:
+Though this is one line command is the standard way to install a software package using Anaconda, **we recommend to
+make use of an environment definition file** that we provide, so you can get additional useful packages together with
+Gammapy in a virtual isolated environment. In order to proceed in this way, just copy and paste in your terminal
+the two lines below:
```
adonath, line 23: Maybe add a link here like "If you want to learn more about using virtual env, see [link to section]"

adonath (dismissed):
Thanks @Bultako! I have left one last comment otherwise looks good to me!

adonath (approved):
Thanks a lot @Bultako, no further comments from my side...

PR #3348, opened 2021-05-05
Distinguish FoV from RoI in gammapy

luca-giunti:
**Description**
As discussed in the #documentation channel on slack, we introduce a clear distinction between the "FoV" and "RoI" terms, both in the documentation and the code.


docs/references.rst
```diff
@@ -43,6 +43,10 @@ Glossary
       of measuring a given reconstructed energy as a function of the true photon
       energy. See :ref:`irf-edisp`
 
+    FoV
+        Stands for "field of view", and indicates the solid angle on the sky that is
+        visible by the telescope with a single pointing.
```
registerrier, line 48: replace telescope by instrument? There are usually more than one telescope

docs/tutorials/analysis/2D/ring_background.ipynb
```diff
@@ -118,10 +118,10 @@
     "    \"lat\": source_pos.dec,\n",
     "    \"frame\": \"icrs\",\n",
     "}  # The WCS geometry - centered on MSH 15-52\n",
-    "config.datasets.geom.wcs.fov = {\"width\": \"3 deg\", \"height\": \"3 deg\"}\n",
+    "config.datasets.geom.wcs.width = {\"width\": \"3 deg\", \"height\": \"3 deg\"}\n",
     "config.datasets.geom.wcs.binsize = \"0.02 deg\"\n",
     "\n",
-    "# The FoV radius to use for cutouts\n",
+    "# The radius to use for cutouts\n",
```
registerrier, line 124: I don't know if this one is very clear. Should we use cutout size? 
Do we really need it?

registerrier (approved):
Thanks @luca-giunti . 

Since we already use `width` for the size of `MapGeom`, I suppose we should stick with it. Technically I suppose it could simple take a simple angle and not split between `width` and `height`.

docs/tutorials/analysis/2D/ring_background.ipynb
luca-giunti: Maybe we don't really need it, and what we need is an additional glossary entry for the offset_max? For the moment I kept it here, with a (hopefully) clearer formulation

docs/references.rst
```diff
@@ -73,6 +77,11 @@ Glossary
     Ring Background
       Background estimation method typically used for image analysis.
 
+    RoI
+      Stands for "region of interest", and indicates the spatial region in which the
```
adonath, line 81: Minor : for abbreviations we could introduce a uniform structure like:
```
RoI
    Short for "Region of Interest":

GTI
    Short for "Good Time Interval":
```
I think there a few other entries that could be adapted to such a scheme.

docs/references.rst
```diff
@@ -73,6 +77,11 @@ Glossary
...
+      data are analyzed. In practice, at each energy it corresponds with the sky region
```
adonath, line 82: "corresponds to"?

docs/references.rst
```diff
@@ -43,6 +43,10 @@ Glossary
       of measuring a given reconstructed energy as a function of the true photon
       energy. See :ref:`irf-edisp`
 
+    FoV
+        Stands for "field of view", and indicates the solid angle on the sky that is
```
adonath, line 47: Strictly speaking it's mostly not given by the solid angle, but by the radius of the (circular) FoV, maybe precise...

adonath (dismissed):
Thanks @luca-giunti, I have left some minor inline comments...

docs/references.rst
luca-giunti: Even in your formulation you are implying that the FoV is circular (i.e. a solid angle) ;) no?

I don't know, sometimes one says "a FoV of [...] sr", and sometimes gives just the radius. Either use is fine I think. I'll try to clarify

adonath: I guess its up to us to define the term, whether we mean solid angle or radius. In the context of IACTs I have mostly heard "FoV of 5 deg" or similar...

PR #3349, opened 2021-05-06
Cleanup MapEvaluator and add more caching

QRemy:
This PR cleanup the MapEvaluator and simplify the choice and order of the methods applied to compute npred.

 It also add a level of caching if only the norm is free (between no parameters changed and if no spatial parameter changed). This is meant to avoid edisp convolution especially for flux points estimation, from my tests on 3FHL validation the gain in fit time is about 25%.

### DIFF
--- a/gammapy/datasets/map.py
+++ b/gammapy/datasets/map.py
@@ -2555,6 +2555,7 @@ class MapEvaluator:
         self._cached_parameter_values_spatial = None
         self._cached_position = (0, 0)
         self._computation_cache = None
+        self._cached_parameter_previous = None
 
     # workaround for the lru_cache pickle issue
     # see e.g. https://github.com/cloudpipe/cloudpickle/issues/178
@@ -2805,6 +2806,7 @@ class MapEvaluator:
         else:
             norm_only_changed, renorm = self.norm_only_changed()
             if not (norm_only_changed and self.use_cache):
+                self._cached_parameter_previous = self.model.parameters.value
                 for method in self.methods_sequence:
                     if inspect.ismethod(method):
                         values = method(self._computation_cache)
@@ -2828,7 +2830,6 @@ class MapEvaluator:
         npred : `~gammapy.maps.Map`
             Predicted counts on the map (in reco energy bins)
         """
-
         if self.parameters_changed or not self.use_cache:
             self._compute_npred.cache_clear()
 
@@ -2894,15 +2895,18 @@ class MapEvaluator:
         """Parameters changed"""
         norm_only_changed = False
         renorm = 1
+        values = self.model.parameters.value
         if self._computation_cache is not None:
-            values = self.model.parameters.value
             names = self.model.parameters.names
-            ind_diff = np.where(self._cached_parameter_values != values)[0]
+            ind_diff = np.where(self._cached_parameter_previous != values)[0]
             if len(ind_diff) == 1:
                 idx = ind_diff[0]
-                norm_only_changed = names[idx] in ["norm", "amplitude"]
+                norm_only_changed = (
+                    names[idx] in ["norm", "amplitude"]
+                    and self._cached_parameter_previous[idx] != 0
+                )
                 if norm_only_changed:
-                    renorm = values[idx] / self._cached_parameter_values[idx]
+                    renorm = values[idx] / self._cached_parameter_previous[idx]
         return norm_only_changed, renorm
 
     @property
--- a/gammapy/datasets/map.py
+++ b/gammapy/datasets/map.py
@@ -2552,11 +2552,12 @@ class MapEvaluator:
         self._compute_npred = lru_cache()(self._compute_npred)
         self._compute_flux_spatial = lru_cache()(self._compute_flux_spatial)
         self._cached_parameter_values = None
+        self._cached_parameter_values_previous = None
         self._cached_parameter_values_spatial = None
         self._cached_position = (0, 0)
         self._computation_cache = None
-        self._cached_parameter_previous = None
         self._neval = 0  # for debugging
+        self._renorm = 1
 
     # workaround for the lru_cache pickle issue
     # see e.g. https://github.com/cloudpipe/cloudpickle/issues/178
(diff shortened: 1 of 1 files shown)


adonath:
Thanks @QRemy! I run our benchmarks locally with master and you branch and this is what I got:
```
 analysis_3d:
-  CPU_max: 101.2
-  CPU_mean: 96.33373493975904
-  memory_peak: 5824.352
-  total_time: 42.472
+  CPU_max: 100.1
+  CPU_mean: 95.58333333333334
+  memory_peak: 5823.285
+  total_time: 42.99
 analysis_3d_joint:
-  CPU_max: 101.7
-  CPU_mean: 99.74426559356137
-  memory_peak: 6246.688
-  total_time: 251.508
+  CPU_max: 101.0
+  CPU_mean: 96.88686679174485
+  memory_peak: 6266.305
+  total_time: 269.996
 io:
-  CPU_max: 99.7
-  CPU_mean: 82.98333333333333
-  memory_peak: 5567.277
-  total_time: 3.553
+  CPU_max: 98.4
+  CPU_mean: 87.22
+  memory_peak: 5559.926
+  total_time: 3.05
 lightcurve_1d:
-  CPU_max: 100.0
-  CPU_mean: 98.35416666666667
-  memory_peak: 5031.898
-  total_time: 12.624
+  CPU_max: 99.7
+  CPU_mean: 94.60357142857143
+  memory_peak: 5033.07
+  total_time: 14.655
 lightcurve_3d:
-  CPU_max: 101.2
-  CPU_mean: 99.6964509394572
-  memory_peak: 6956.422
-  total_time: 242.658
+  CPU_max: 100.7
+  CPU_mean: 97.58179999999999
+  memory_peak: 6941.477
+  total_time: 253.165
 ring_background_estimator:
-  CPU_max: 99.6
-  CPU_mean: 93.04634146341462
-  memory_peak: 5216.305
-  total_time: 21.233
+  CPU_max: 99.7
+  CPU_mean: 93.9325
+  memory_peak: 5193.059
+  total_time: 20.722
 spectrum_1d:
-  CPU_max: 99.9
-  CPU_mean: 99.31666666666669
-  memory_peak: 5068.66
-  total_time: 18.703
+  CPU_max: 99.7
+  CPU_mean: 93.47209302325581
+  memory_peak: 5069.594
+  total_time: 22.245
 spectrum_1d_joint:
   CPU_max: 100.0
-  CPU_mean: 97.49688581314878
-  memory_peak: 5081.836
-  total_time: 471.515
+  CPU_mean: 97.22310231023103
+  memory_peak: 5075.336
+  total_time: 153.605
 tsmap_estimator:
-  CPU_max: 99.7
-  CPU_mean: 87.96206896551725
-  memory_peak: 5303.785
-  total_time: 29.841
+  CPU_max: 99.9
+  CPU_mean: 98.18125000000002
+  memory_peak: 5217.898
+  total_time: 24.748
```
In general the results are not super accurate, so I think the additional runtime in the TS map estimation spectrum 1d might be accidentally. The only significant improvement I see is in the `spectrum_1d_joint` benchmark. I think this is expected, however I don't understand, why e.g. `analysis_3d_joint` does not improve. I'll do a bit of profiling on these cases to understand what's happening and comment again. However the speed up of factor 3 in the `spectrum_1d_joint`benchmark is really nice!

adonath:
Unrelated, but a general thing it noticed: for the benchmarks we should save some key numbers, such as analysis results, number of iterations of the fit etc. to make sure the results are equivalent and can be compared.

adonath:
Note: in the 3D benchmarks we don't define a separate true energy axis https://github.com/gammapy/gammapy-benchmarks/blob/master/benchmarks/analysis_3d_joint.py#L42, that's probably why we don't see any speed up there. In general it might be good to define it...

adonath:
@QRemy I just came back to this PR. My minimal proposal would be:

- Please rebase the PR
- Modify one of our benchmarks to include a different true energy axis and just run the benchmark once compared against master to verify that the performance improvement is there as well in the 3D case. I think this should be quick to do.

I'll do a final review on the implementation, but so far I think only minor comments.

gammapy/datasets/map.py
```diff
@@ -2886,3 +2865,46 @@ def irf_position_changed(self):
 
         return changed
 
+
+    def norm_only_changed(self):
```
adonath, line 2869: This method needs simplification. A few ideas:
- Identify the "norm/amplitude" parameter on MapEvaluator init
- Just check the norm like:
```python3
@property
def parameters_norm_changed(self, reset=True):
    """Parameters changed"""
    value = self.model.parameters.value[self._norm_idx]
    changed = self._cached_parameter_values[sefl._norm_idx] == value

    if changed and reset:
        self._cached_parameter_values = values

    return changed

```

- In addition maybe introduce the re-norm factor like:
```python3

def renorm():
    value = self.model.parameters.value[self._norm_idx]
    value_cached = self._cached_parameter_values[self._norm_idx]
    return value / value_cached    
```

gammapy/datasets/map.py
```diff
@@ -2778,14 +2779,16 @@ def _compute_npred(self):
...
-                npred = self.apply_exposure(npred)
-
-            if self.model.apply_irf["edisp"]:
-                npred = self.apply_edisp(npred)
-
+            norm_only_changed, renorm = self.norm_only_changed()
+            if not (norm_only_changed and self.use_cache):
+                self._cached_parameter_previous = self.model.parameters.value
+                for method in self.methods_sequence:
+                    if len(inspect.signature(method).parameters) == 0:
```
adonath, line 2786: I would just remove the check here and introduce optional "dummy" arguments (like `*args`) to the methods that don't require any arguments

gammapy/datasets/map.py
```diff
@@ -2886,3 +2865,46 @@ def irf_position_changed(self):
...
+                norm_only_changed = (
+                    names[idx] in ["norm", "amplitude"]
+                    and self._cached_parameter_previous[idx] != 0
+                )
+                if norm_only_changed:
+                    renorm = values[idx] / self._cached_parameter_previous[idx]
+        return norm_only_changed, renorm
+
+    @property
+    def methods_sequence(self):
```
adonath, line 2888: Minor: this could be a lazy property: the order never changes during execution...

adonath (changes requested):
Thanks @QRemy, please rebase once again. In addition I have left a few inline comments with suggestions for simplification.

gammapy/datasets/map.py
QRemy: done, not exactly like this but that's simpler now

adonath (approved):
Thanks @QRemy for the simplifications! No further comments from my side...

Issue #2913, opened 2020-06-02
Add helpers for using 2D maps in gammapy


AtreyeeS:
I recently had some exchange with Emma, who wanted to use HESS images (data reduction not done using gammapy), and do a fitting within gammapy.

In this case, it is required to add an `energy` axis on the counts and background maps, and an `energy_true` axis on the exposure. Converting the psf is a bit tricker, the `energy_true` axis should be added on  `PSFKernel.psf_kernel_map`  and then a new `PSFKernel` created.

Perhaps it is difficult to provide helper functions to do this properly for all cases, but maybe we can add some info about this in the documentation for future users?

adonath:
Indeed this is a use case, that came up many times already. Here are a few ideas from my side:

- We already have `Geom.to_cube()`, we could also introduce `Map.to_cube(axes=[])`, which attaches additional axes to a given 2D map. The same could be done by adding `PSFKernel.to_cube()`
- What could be useful as well and is more general is something like `Map.from_images(images=[], axis=energy_axis)`. This would create a map from a list of 2D images and attach the extra axis.

adonath:
And documentation is a good idea anyway...

emmaow:
Hi! thanks @AtreyeeS AtreyeeS for the help. I can provide the notebook whenever im done, if you like. I am using dataset=MapDatasetOnOff as Atreyee suggested. I am still trying to understand how to implemented (do I need to provide exposure on off? models?). Once I understand this it should be done. I can make the example using some public data

registerrier:
I am about to add a notebook to the HESS bitbucket repo for this specific use case. 

We should also add a specific set of examples, probably within the `Dataset` rst page

AtreyeeS:
> * What could be useful as well and is more general is something like `Map.from_images(images=[], axis=energy_axis)`. This would create a map from a list of 2D images and attach the extra axis.

Hi @adonath : How about a `MapDataset.from_images()`? Then it can automatically attach the extra axis as `energy` or `energy_true`...

adonath:
We now also have `Map.from_images()` (see https://github.com/gammapy/gammapy/blob/master/gammapy/maps/core.py#L1105). So I guess the remaining work is to add documentation?

adonath:
Implementing a `MapDataset.from_images()` could be still useful...

adonath:
Will be resolved with HESS internal documentation.

PR #3350, opened 2021-05-06
Add MapDataset.from_images

AtreyeeS:
**Description**

This pull request addresses #2913 

**Dear reviewer**


gammapy/datasets/map.py
```diff
@@ -1128,6 +1134,59 @@ def from_dict(cls, data, lazy=False, cache=True):
         dataset = cls.read(filename, name=data["name"], lazy=lazy, cache=cache)
         return dataset
 
+    @classmethod
+    def from_images(cls, images, energy_min, energy_max, name=None, **kwargs):
+        """Create a MapDataset from a set 2D images which do not have an
```
registerrier, line 1139: from a dictionary of

gammapy/datasets/map.py
```diff
@@ -1128,6 +1134,59 @@ def from_dict(cls, data, lazy=False, cache=True):
...
+        energy axis. An appropriate energy axis is added for all the images.
+
+        This can be useful for modelling and fitting data reduced outside gammapy
+
+        Parameters
+        ----------
+        images: dict
+        Dictionary of images. Allowed keywords are `counts`, `background`,
```
registerrier, line 1147: need to increase indentation here

gammapy/datasets/map.py
```diff
@@ -1893,6 +1959,57 @@ def from_geoms(
...
+         --------
+         datasetonoff: `MapDatasetOnOff`
+            Map dataset.
+         """
+
+        dataset = super().from_images(
+            images, energy_min, energy_max, name=None, **kwargs
+        )
+
+        energy_axis = MapAxis.from_energy_edges([energy_min, energy_max], name="energy")
```
registerrier, line 1993: you can retrieve it from the maps contained in `dataset`

gammapy/datasets/map.py
```diff
@@ -1128,6 +1134,59 @@ def from_dict(cls, data, lazy=False, cache=True):
...
+        --------
+        dataset: `MapDataset`
+            Map dataset.
+        """
+
+        name = make_name(name)
+        kwargs["name"] = name
+
+        energy_axis = MapAxis.from_energy_edges([energy_min, energy_max], name="energy")
+        energy_axis_true = MapAxis.from_energy_edges(
```
registerrier, line 1167: What if the user has a different binning for psf kernel and exposure?

registerrier (commented):
Thanks @AtreyeeS .

This looks good. But I have a problem with the energy axis definition. Here the only supported case is when energy and energy true axes are identical. Isn't it an issue?

AtreyeeS:
Hi @registerrier : Actually I thought about this as well - Then it would need an energy dispersion kernel. Can that normally be available without having an energy axis on the exposure?

gammapy/datasets/map.py
adonath: But we always require that the PSF kernel and exposure have the same energy axis, except if the PSF is in reco energy. Is that what you meant?

gammapy/datasets/map.py
```diff
@@ -1128,6 +1134,59 @@ def from_dict(cls, data, lazy=False, cache=True):
...
+    @classmethod
+    def from_images(cls, images, energy_min, energy_max, name=None, **kwargs):
+        """Create a MapDataset from a set 2D images which do not have an
+        energy axis. An appropriate energy axis is added for all the images.
+
+        This can be useful for modelling and fitting data reduced outside gammapy
+
+        Parameters
+        ----------
+        images: dict
```
adonath, line 1146: Maybe the documentation and usage becomes simpler (especially for the `PSFKernel`  case) if the maps are passed directly like: `.from_images(counts, background, exposure, psf_kernel,  )`. I think there is also some documentation missing...

AtreyeeS:
I removed the lines from the main code base and added them as an example in the H.E.S.S. notebook. More than think should be as internal documentation I guess.

adonath:
Thanks @AtreyeeS for the update. There are a two things I'm not yet happy with:
- I think we should not advertise to use the `MapDatasetOnOff` for spatial fitting, even within HESS I think this has never been done. Instead just use `MapDataset` with the `BACKGROUNDMAP`
- The way the exposure is currently computed from `EXPGAMMAMAP` is incorrect I think. To get correct flux values for fitting  of models, I think one needs to (a) define the energy axis with `e_min = 1 TeV` and `e_max = np.inf * u.TeV` (or at  least a very large number...) and (b) always assume a power-law with spectral index of 2.3 or whatever index  has been chosen respectively.

AtreyeeS:
So do you propose
- add a `datasetonoff.to_mapdataset()` before going for fitting, or directly creating a `MapDataset`? I would slightly prefer the former.
- Energy range can be changed, but there is no Spectral Model involved at this stage. Where do I incorporate this?

adonath:
> add a datasetonoff.to_mapdataset() before going for fitting, or directly creating a MapDataset? I would slightly prefer the former.

I think I would directly create a `MapDataset` here, because the only thing one could do with the `MapDatasetOnOff` is to use the `ExcessMapEstimator`, however all the maps computed by the `ExcessMapEstimator` are already present in the HAP maps. So I think the only fair use case it fitting... 

> Energy range can be changed, but there is no Spectral Model involved at this stage. Where do I incorporate this?

I think this has to remain a comment / note at this stage...

docs/tutorials/data/hess.ipynb
```diff
@@ -12,7 +12,17 @@
     "\n",
```
registerrier, line 12: <p>unnecessary</p><p> </p><p>Do we really want to put the Note written like this? After all we should have the opportunity to introduce data at the DL4 level.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/gammapy/gammapy/pull/3350/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='15'/>

registerrier, line 12: <p>Do we need this? Would it really work like this?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/gammapy/gammapy/pull/3350/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='20'/>

registerrier, line 12: <p>as explained</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/gammapy/gammapy/pull/3350/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='26'/>

adonath:
@AtreyeeS, what do we do about this PR?

adonath:
Just discussed in the dev call: right now there are no public HESS maps we could use to document e.g. a morphology fit. So the proposal is to move the documentation to internal HESS documentation instead. If there is public data at some point we can add some public documentation to Gammapy as well...

PR #3353, opened 2021-05-07
Tweak in MapEvaluator.need_update for performance

QRemy:
This PR introduce a performance tweak in MapEvaluator.need_update that replaces `skycoord.separation` by `angular_separation` in order to avoid unnecessary frame and unit equivalence comparison.

### DIFF
--- a/gammapy/datasets/map.py
+++ b/gammapy/datasets/map.py
@@ -2653,10 +2653,13 @@ class MapEvaluator:
                 self.psf = psf.get_psf_kernel(position=self.model.position, geom=geom)
 
         if self.evaluation_mode == "local":
-            self._init_position = (
-                spatial.lon_0.quantity.to_value(u.rad),
-                spatial.lat_0.quantity.to_value(u.rad),
-            )
+            try:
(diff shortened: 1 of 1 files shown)


QRemy:
Added a test to avoid update if there is no free spatial parameters (solve fail related to `TemplateSpatialModel`)

adonath:
Thanks a lot @QRemy! One last suggestion to try: maybe it is an option to also check for `if parameters_spatial_changed` in `.need_update` and only check for the separation if this is the case?

adonath:
Thanks @QRemy, it seems there are now test fails https://github.com/gammapy/gammapy/pull/3353/checks?check_run_id=2557044541#step:9:1865, I can also reproduce them locally....not sure what is happening. Maybe see if they can be fixed with limited effort...

adonath (approved):
Thanks @QRemy, the remaining test fails seems unrelated. I think adding the `lon_0` and `lat_0` parameters to the `TemplateSpatialModel` is still a good idea, however this can be done in a follow up PR.

PR #3355, opened 2021-05-07
Implement WcsNDMap.plot_mask()

fabiopintore:
This PR adds a new functionality, `plot_mask`, in `~gammapy.Map` as requested in the issue #3271. It allows to plot a `mask` as shaded areas above a given map.

### DIFF
--- a/gammapy/maps/core.py
+++ b/gammapy/maps/core.py
@@ -1111,6 +1111,38 @@ class Map(abc.ABC):
                 fig, ax, cbar = img.plot(stretch=stretch, **kwargs)
                 plt.show()
 
+    def plot_mask(self, ax=None, **kwargs):
+        """Plot the mask as a shaded area
+
+        Parameters
+        ----------
+        ax : `~matplotlib.axis`
+            Axis instance to be used for the plot.
+        **kwargs : dict
+            Keyword arguments passed to `~matplotlib.pyplot.axvspan`
+
+        Returns
+        -------
+        ax : `~matplotlib.pyplot.Axis`
+            Axis used for plotting
+        """
+        import matplotlib.pyplot as plt
+
+        if not self.is_mask:
+            raise ValueError("This is not a mask and cannot be plotted")
+
+        kwargs.setdefault("alpha", 0.05)
+        kwargs.setdefault("colors", "k")
+
+        ax = plt.gca() if ax is None else ax
+
+        data = self.data.astype(float)
+
+        cs = ax.contourf(data, level=[0, 0.5], **kwargs)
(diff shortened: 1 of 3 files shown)


gammapy/maps/core.py
```diff
@@ -1111,6 +1111,38 @@ def _plot_interactive(**ikwargs):
...
+        """
+        import matplotlib.pyplot as plt
+
+        if not self.is_mask:
+            raise ValueError("This is not a mask and cannot be plotted")
+
+        kwargs.setdefault("alpha", 0.05)
+        kwargs.setdefault("colors", "k")
+
+        ax = plt.gca() if ax is None else ax
```
adonath, line 1137: If `ax` is None we need to define `WCSAxis` by default, see https://github.com/gammapy/gammapy/blob/master/gammapy/maps/wcsnd.py#L343

gammapy/maps/core.py
```diff
@@ -1111,6 +1111,38 @@ def _plot_interactive(**ikwargs):
...
+        ax : `~matplotlib.pyplot.Axis`
+            Axis used for plotting
+        """
+        import matplotlib.pyplot as plt
+
+        if not self.is_mask:
+            raise ValueError("This is not a mask and cannot be plotted")
+
+        kwargs.setdefault("alpha", 0.05)
+        kwargs.setdefault("colors", "k")
```
adonath, line 1135: My proposal would be to go for `color="w"` and `alpha=0.5` or similar, this will result in a plot like this:image
Where the masked area is greyed out, which is very intuitive I think...

gammapy/maps/core.py
```diff
@@ -1111,6 +1111,38 @@ def _plot_interactive(**ikwargs):
...
+
+        ax = plt.gca() if ax is None else ax
+
+        data = self.data.astype(float)
+
+        cs = ax.contourf(data, level=[0, 0.5], **kwargs)
+        ax.contour(cs)
```
adonath, line 1142: I think for now there is no need to call `ax.contour()` in addition, so no extra plotting of the boundary...

adonath (changes requested):
Thanks @fabiopintore, I have left some inline comments...

gammapy/maps/core.py
fabiopintore: Done.

adonath:
Thanks @fabiopintore, I attached one commit doing a bit of cleanup. I also noticed we needed to move `.plot_mask()` to `WcsNDMap`, because the creation of the default axis is specific for WCS and does not work for HEALPix. I'll wait for the CI builds to pass and merge...

PR #3361, opened 2021-05-17
Introduce SpatialModel.position_lonlat

adonath:
**Description**

This pull request is a follow up to #3353 and introduces `SpatialModel.position_lonlat` to clean up a bit the implementation. 
It also add a `reset` parameter to `parameters_spatial_changed()` which allows to check for the change without reseting the old values.

### DIFF
--- a/gammapy/modeling/models/cube.py
+++ b/gammapy/modeling/models/cube.py
@@ -187,7 +187,7 @@ class SkyModel(Model):
 
     @property
     def position_lonlat(self):
-        """`~astropy.coordinates.SkyCoord`"""
+        """Spatial model center position `(lon, lat)` in rad and frame of the model"""
         return getattr(self.spatial_model, "position_lonlat", None)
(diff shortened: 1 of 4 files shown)


gammapy/modeling/models/cube.py
```diff
@@ -185,6 +185,11 @@ def position(self):
         """`~astropy.coordinates.SkyCoord`"""
         return getattr(self.spatial_model, "position", None)
 
+    @property
+    def position_lonlat(self):
+        """`~astropy.coordinates.SkyCoord`"""
```
registerrier, line 190: This does not return a SkyCoord.

registerrier (commented):
Looks good @adonath . A small comment inline

gammapy/modeling/models/cube.py
adonath: Done

PR #3364, opened 2021-05-18
Introduce a Maps container

registerrier:
**Description**

This pull request introduces a dictionary for `Map` objects sharing the same `MapGeom`. It inherits from `MutableMapping` and basically simply checks that the `geom` is consistent at each new addition. It also allows serialization of all maps at once all sharing the same `hdu_band`. 

This addition should support the use case not fully covered by the `FluxMap` object, for example excess maps. 

**Dear reviewer**


adonath:
Just for the record: at the last dev meeting the were the following proposals:
- For consistency with other container classes, rename to `Maps`
- Unlike for other "named" objects, we decided to not introduce `Map.name`, but define the the names for `Maps` separately.


Test-wise I explored the option of a `LabelMapAxis` and setting the axis as an "index" for `Map.__getitem__` here: https://github.com/adonath/notebooks-public/blob/master/labeled-map-axis-prototype.ipynb Technically this leads to a similar API and seems to work, but probably requires more implementation effort to support all map methods etc. and definitely limits the data to the exact same geometry. I think in general, it's probably not either `Maps` or `LabeledMapAxis`, but both. The `Maps` abstraction is simpler, while the `LabeledMapAxis` is more general and flexible, there are likely use-cases for both. So it does not affect the implementation proposed in tis PR.

registerrier:
> * For consistency with other container classes, rename to `Maps`

Done.

> * Unlike for other "named" objects, we decided to not introduce `Map.name`, but define the names for `Maps(maps, names=None)` separately (proposed by @QRemy ). The names should remain optional I think and the default access to `Maps` just goes via the index, as we have for the other container classes as well.

Well I am not sure I see the point of this. But well...
What kind of container should I use then? `MutableSequence`? But can I create a `Maps.__setitem__` that will take an optional `key`? Same for `Maps.__getitem__`.
 
> Test-wise I explored the option of a `LabelMapAxis` and setting the axis as an "index" for `Map.__getitem__` here: https://github.com/adonath/notebooks-public/blob/master/labeled-map-axis-prototype.ipynb Technically this leads to a similar API and seems to work, but probably requires more implementation effort to support all map methods etc. and definitely limits the data to the exact same geometry. I think in general, it's probably not either `Maps` or `LabeledMapAxis`, but both. The `Maps` abstraction is simpler, while the `LabeledMapAxis` is more general and flexible, there are likely use-cases for both. So it does not affect the implementation proposed in tis PR.

I think it is confusing to be able to access something with `label_map[i]` or this pattern for slicing needs to be supported for regular maps as well.

gammapy/maps/map_dict.py
```diff
@@ -0,0 +1,139 @@
...
+            self.__setitem__(key, value)
+
+    @property
+    def geom(self):
+        """Map geometry (`Geom`)"""
+        return self._geom
+
+    def __setitem__(self, key, value):
+        if value is not None and not isinstance(value, Map):
+            raise ValueError(f"MapDict can only contain Map objects.")
```
adonath, line 32: Add the actual and expected type here, maybe like  `f"Maps can only contain Map objects, got {type(value)} instead."`

gammapy/maps/map_dict.py
```diff
@@ -0,0 +1,139 @@
...
+
+        if len(self._data) > 0:
+            if value.geom != self._geom:
+                raise ValueError(f"MapDict items must share the same geometry.")
```
adonath, line 36: Again add actual and expected geom here...

gammapy/maps/map_dict.py
```diff
@@ -0,0 +1,139 @@
...
+
+    For maps with different geometries, use a regular dict.
+    """
+
+    def __init__(self, **kwargs):
+        self._geom = None
+        self._data = {}
+        for key, value in kwargs.items():
+            self.__setitem__(key, value)
+
```
adonath, line 24: I think a method like `Maps.from_geom(geom, names=[])` could be useful to create an empty container to be filled with data later...I think there are many use case within Gammapy for this.

gammapy/maps/map_dict.py
```diff
@@ -0,0 +1,139 @@
...
+    def __delitem__(self, key):
+        del self._data[key]
+
+    def __iter__(self):
+        return iter(self._data)
+
+    def __repr__(self):
+        return f"{type(self).__name__}({self._data})"
+
```
adonath, line 57: I think a `__str__` method with a little more output would be nice, like printing the geom and all the available maps...

adonath (commented):
Thanks a lot @registerrier, I have left a few inline comments, I'll checkout and test the branch now...

gammapy/maps/map_dict.py
registerrier: Done

gammapy/maps/map_dict.py
registerrier: OK. Done.

gammapy/maps/map_dict.py
registerrier: Done. The function is `Maps.from_geom(geom, names, kwargs_list=None)` to have the possibility to pass the arguments.

Then it can be called:
```
    geom = WcsGeom.create(npix=5)
    names = ["map1", "map2", "map3"]
    kwargs_list = [
        {"unit":"cm2s", "dtype":"float64"},
        {"dtype" : "bool"},
        {"data" : np.arange(25).reshape(5,5)}
    ]

    maps = Maps.from_geom(geom, names, kwargs_list=kwargs_list)

adonath:
Thanks @registerrier! Maybe rename the actual code file or move `Maps` to `core.py` and then merge?

adonath (approved):

Issue #3004, opened 2020-09-10
Model kernel_radius overwrote in TSMapEstimator


emmaow:
**Gammapy version**
0.17
**Bug description**
When producing TSMaps using a model with a Gaussian kernel of x deg, the init value of TSMapEstimator always uses a fixed value (point?). To solve this one has to init the kernel_width directly on the call TSMapEstimator(model,kernel_radius=x*u.deg)

**To Reproduce**
spatial_model = GaussianSpatialModel(sigma="0.5 deg")
spectral_model = PowerLawSpectralModel(index=2)
model = SkyModel(spatial_model=spatial_model, spectral_model=spectral_model)
ts_image_estimator = TSMapEstimator(model)
images_ts = ts_image_estimator.run(dataset_image)

=> images_ts with a 0.1 deg.

ts_image_estimator = TSMapEstimator(model,kernel_width=0.5*u.deg)

=> images_ts with a 0.5 deg

**Other information** 

https://docs.gammapy.org/0.17/notebooks/cta_data_analysis.html?highlight=tsmapestimator%20gaussian%20model

adonath:
Thanks @emmaow, I think we have to clarify what the `kernel_width` parameters does and probably define a better default behaviour.  The `kernel_width` parameter is independent of the actual spatial shape, but just defines the size of map that is used to create the kernel. It's a useful parameter for TS map estimation, as the runtime will scale with `kernel_width ** 2`.

But I agree there is a large potential for mis-use, e.g. the example that you provided, when a `0.5 deg` Gaussian is only evaluated in a `0.2 deg` window. The minimal thing we could do is raise a warning.

registerrier:
I think that an adaptive kernel size taking into account the PSF and the source model would provide a solid implementation for the `TSMapEstimator`.  It should be possible to find a clean solution for this problem as it is also present in other parts of the code. For instance for the choice of the psf kernel dimension.

adonath:
@registerrier Yes, we already have this code in `MapEvaluator.update()`, probably we should move it somewhere else, so that we can re-use it. On the other hand we already use the `MapEvaluator` in `TSMapEstimator.estimate_kernel()`, so maybe we just have to refactor it partly...

adonath:
I think here we should provide a safe default behaviour by inferring the kernel size automatically by default, however we should keep the `kernel_size` parameter to allow users to overwrite it.

registerrier:
[...]

PR #3366, opened 2021-05-19
Change default kernel_size to None in TSMapEstimator

registerrier:
**Description**

This pull request changes the default behavior of the `TSMapEstimator`. The `kernel_size` default value of 0.2 deg can be very underestimated in a number of cases.

Here we change the default `kernel_size` to `None`. By default, it will use the `MapEvaluator.cutout_width` which is the sum of the PSF kernel width with the model `evaluation_radius`. 
While this avoids grossly underestimated kernel sizes, this can also yield very large sizes since the PSF kernel is by default computed until `rad_max`.

The issue was raised in #3004 .

**Dear reviewer**

Another option would be to impose a containment fraction on the kernel itself (directly in the reco space).

### DIFF
--- a/gammapy/estimators/tests/test_ts_map.py
+++ b/gammapy/estimators/tests/test_ts_map.py
@@ -112,6 +112,10 @@ def test_compute_ts_map(input_dataset):
     ts_estimator = TSMapEstimator(
         model=model, threshold=1, kernel_width="1 deg", selection_optional=[]
     )
+
+    kernel = ts_estimator.estimate_kernel(dataset=input_dataset)
+    assert_allclose(kernel.geom.width, 1.02 * u.deg)
+
     result = ts_estimator.run(input_dataset)
 
     assert_allclose(result["ts"].data[0, 99, 99], 1704.23, rtol=1e-2)
--- a/gammapy/estimators/tests/test_ts_map.py
+++ b/gammapy/estimators/tests/test_ts_map.py
@@ -110,20 +110,20 @@ def test_compute_ts_map(input_dataset):
     spectral_model = PowerLawSpectralModel(index=2)
     model = SkyModel(spatial_model=spatial_model, spectral_model=spectral_model)
     ts_estimator = TSMapEstimator(
-        model=model, threshold=1, kernel_width="1 deg", selection_optional=[]
+        model=model, threshold=1, selection_optional=[]
     )
 
     kernel = ts_estimator.estimate_kernel(dataset=input_dataset)
-    assert_allclose(kernel.geom.width, 1.02 * u.deg)
+    assert_allclose(kernel.geom.width, 1.22 * u.deg)
 
     result = ts_estimator.run(input_dataset)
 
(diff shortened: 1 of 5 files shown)


adonath:
Thanks @registerrier! I started to try out the implementation and I think there are two related changes we need to to:
- Add padding / cropping in the `TSMapEstimator` to avoid that the NaN boundary becomes too large
- Change the PSF kernel size estimation to containment radius. The question there is how to handle the energy dependence and which containment fraction to choose. I think this change should happen in the `MapEvaluator` so that it applies "normal" model fitting as well. I guess for the energy dependence one could just choose the conservative solution of the `max(radius_per_energy)` and the containment fraction could be maybe set 99% as a starting point. 

I think those changes should happen independently before we merge this PR.

adonath:
@registerrier Once #3389 is merged I'll open a PR to implement the padding for TS maps first.

adonath:
Related #3392.

adonath:
Once #3400 is merged the implementation here should become simple. @registerrier do you continue or should I take over?

adonath:
Sorry @registerrier I "hijacked" your PR here. Please review... :-)

gammapy/estimators/tests/test_ts_map.py
```diff
@@ -110,16 +110,20 @@ def test_compute_ts_map(input_dataset):
...
     result = ts_estimator.run(input_dataset)
 
     assert_allclose(result["ts"].data[0, 99, 99], 1704.23, rtol=1e-2)
-    assert_allclose(result["niter"].data[0, 99, 99], 8)
+    assert_allclose(result["niter"].data[0, 99, 99], 7)
     assert_allclose(result["flux"].data[0, 99, 99], 1.02e-09, rtol=1e-2)
     assert_allclose(result["flux_err"].data[0, 99, 99], 3.84e-11, rtol=1e-2)
-    assert_allclose(result["npred"].data[0, 99, 99], 3627.874063, rtol=1e-2)
-    assert_allclose(result["npred_null"].data[0, 99, 99], 2601, rtol=1e-2)
+    assert_allclose(result["npred"].data[0, 99, 99], 4744.020361, rtol=1e-2)
```
adonath, line 125: When the kernel size change the npred must change as well, however the `npred_excess` remains the same..

gammapy/estimators/ts_map.py
```diff
@@ -198,45 +201,44 @@ def estimate_kernel(self, dataset):
...
+            psf=dataset.psf,
+            edisp=dataset.edisp,
+            geom=dataset.counts.geom,
+            mask=dataset.mask_image,
         )
-
         kernel = evaluator.compute_npred()
         kernel.data /= kernel.data.sum()
-
-        if (self.kernel_width >= geom.width).any():
```
adonath, line 224: With padding this case cannot appear anymore. Independent of the size of the kernel, the input map is enlarged correspondingly...

adonath:
@registerrier I'll merge this PR as well. Please let me know of you have any remaining comments in private or open an issue. Thanks!

PR #3371, opened 2021-05-22
Add plot_mask to HpxNDMap

fabiopintore:
This PR adds a new functionality, `plot_mask`, in `~gammapy.HpxNDMap` as an extension of fixing the issue #3271. It allows to plot a mask as shaded areas above a given healpy map.

### DIFF
--- a/gammapy/maps/hpxnd.py
+++ b/gammapy/maps/hpxnd.py
@@ -959,22 +959,44 @@ class HpxNDMap(HpxMap):
         return fig, ax, p
 
 
-    def plot_mask(self, ax=None, proj="AIT", **kwargs):
+    def plot_mask(self,
+            method="raster",
+            ax=None,
+            normalize=False,
+            proj="AIT",
+            oversample=2,
+            width_pix=1000,
+            **kwargs,
+        ):
         """Plot the mask as a shaded area
 
         Parameters
         ----------
-        ax : `~astropy.visualization.wcsaxes.WCSAxes`, optional
-            WCS axis object to plot on.
+        method : {'raster','poly'}
+            Method for mapping HEALPix pixels to a two-dimensional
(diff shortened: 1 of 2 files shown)


adonath:
Thanks @fabiopintore! The following simple test case should work with this PR:

```python3
from gammapy.maps import HpxGeom

geom = HpxGeom.create(nside=16)

region = CircleSkyRegion(
    center=SkyCoord("0d", "0d", frame="galactic"),
    radius=20 * u.deg
)

mask = geom.region_mask([region])

mask.plot_mask()
```
Right now it fails with `AttributeError: 'WcsGeom' object has no attribute 'to_wcs_geom'`. Can you please check the implementation once again and adapt the test accordingly?

adonath:
Thanks @fabiopintore! Unfortunately the result still doesn't look right to me:
```python3
from gammapy.maps import HpxGeom, Map
import numpy as np

geom = HpxGeom.create(nside=64)

region = CircleSkyRegion(
    center=SkyCoord("0d", "0d", frame="galactic"),
    radius=20 * u.deg
)

mask = geom.region_mask([region])

m = Map.from_geom(geom=geom)
m.data += 3
m.data = np.random.poisson(m.data)
m.plot()
mask.plot_mask(alpha=0.2)
```

image
 

Please take a look at `HpxNDMap.plot()` and see how it is implemented:
- Basically copy and paste from `HpxNDMap.plot()`, but only support the `method="raster"` for now.
- Internally you call `.to_wcs()` to create a `WcsNDMap`, but you have to fix the data after, like `m.data = np.nan_to_num(m.data).astype(bool)`
- The internally you can just call `m.plot_mask()`, which calls `WcsNDMap.plot_mask()`

I hope this makes sense...

adonath (approved):
Thanks a lot @fabiopintore! I have added one minor cleanup commit...

PR #3372, opened 2021-05-25
Remove datasets argument from Fit.__init__

adonath:
**Description**

This pull request includes the proposal to remove the `datasets` argument from `Fit` and at the same time expose the `Fit` on the `Estimator` classes. The advantages are:

- A `Fit` instance is defined by it's configuration, such as backend etc. 
- All fit argument can be validate on init and fail early (not really implement yet...but should come in future)
- Exposing the `fit` on the `Estimator` classes allows full configuration of the fit for all estimators. Which allows using all backends, adjusting max iterations etc while using the exact same API as for any "normal" `Fit`
- The same fit instance can be re-used in multiple model / datasets fits

The change is not backwards compatible, but the gain in functionality justifies it I think.


adonath:
I think the most controversial point is whether the `Estimator` classes should directly take a `Fit` object bundling the configuration. I think the alternative way would be to work with a `**kwargs` pattern and pass on keywords...

adonath:
One more question is whether we need separate `backend_optimize`, `backend_convariance` etc. arguments, because not all methods are supported by all backends.

gammapy/estimators/flux_point.py
```diff
@@ -771,16 +769,13 @@ def __init__(
...
-            optimize_opts = {}
-        if covariance_opts is None:
-            covariance_opts = {}
-        self.optimize_opts = optimize_opts
-        self.covariance_opts = covariance_opts
-        self.reoptimize = reoptimize
         self.selection_optional = selection_optional
 
+        if fit is None:
+            fit = Fit(reoptimize=False, confidence_opts={"backend": "scipy"})
```
adonath, line 775: This change solves a real issue: https://docs.gammapy.org/dev/tutorials/analysis/1D/spectral_analysis.html#Compute-Flux-Points In the current dev version the UL of the last flux point in the link above is not computed correctly. This is partly a know issue, as `iminuit` showed convergence problems for "linear" fit statistics. The `scipy` backend  with the Brentq algorithm behaves much more stable in this regard. However I'm not sure what to do about this in general...

gammapy/analysis/core.py
```diff
@@ -172,7 +172,7 @@ def read_models(self, path):
         models = Models.read(path)
         self.set_models(models)
 
-    def run_fit(self, optimize_opts=None):
+    def run_fit(self):
```
registerrier, line 175: Here it is no longer possible options on the `Fit` class. Does it mean we need to update the configuration to introduce parameters on `FitConfig`?

gammapy/estimators/flux.py
```diff
@@ -38,14 +39,6 @@ class FluxEstimator(Estimator):
         Sigma to use for asymmetric error computation.
     n_sigma_ul : int
         Sigma to use for upper limit computation.
-    backend : str
```
registerrier, line 41: Here it seems problematic to assume that users will have to go through passing the full `Fit` object simply to change the `reoptimize` option. In particular this is impossible to configure from the HLI. 
We have to keep the option directly on `__init__`.

gammapy/estimators/lightcurve.py
```diff
@@ -351,6 +344,8 @@ class LightCurveEstimator(Estimator):
             * "scan": estimate fit statistic profiles.
 
         Default is None so the optionnal steps are not executed.
+    fit : `Fit`
+        Fit instance specifying the backend and fit options.
```
registerrier, line 348: Here one needs to explain what happens if left to `None`

gammapy/estimators/flux_point.py
```diff
@@ -742,6 +735,9 @@ class FluxPointsEstimator(Estimator):
             * "scan": estimate fit statistic profiles.
 
         Default is None so the optionnal steps are not executed.
+    fit : `Fit`
+        Fit instance specifying the backend and fit options.
```
registerrier, line 739: The default behavior should be explained.

registerrier (changes requested):
Thanks @adonath. 
I think that passing the `Fit` object to pass the various options is probably a good idea. 
I am less convinced by the addition of `reoptimize` on `Fit` and hiding it from the user on the `Estimators`. 
With this change, you need to make a more complex treatment to cover the case of `reoptimize=True`. At the same time, when exploring a `Fit` you are forced to recreate a fit object just to see the impact of the `reoptimize=True`.

gammapy/estimators/flux.py
adonath: I think the recommended way of working with the `Estimator` becomes slightly different:
```python3

estimator = FluxEstimator()
estimator.fit.reoptimize = True

result_1 = estimator.run(datasets)

estimator.fit.reoptimize = False
result_2 = estimator.run(datasets)
``` 
However we can maybe find an alternative pattern e.g. based on `kwargs` passed on to `Fit`, or re-exposing the most important parameters (however this is non-ideal, as it is not extensible and arbitrary to a certain degree...)

gammapy/analysis/core.py
adonath: This would be the perfect solution, however modifying the `Fit` class in place always works:
```python3
analysis.fit.backend = "minuit"
analysis.fit.optimize_opts = {"something":  2}
```

However we should probably introduce a `Fit.__str__` so that it become easy to see which options are set.

adonath:
Some more feedback from the dev call this morning:
- The global "reoptimize" option does not really make sense, instead it should be configurable for `confidence()` and `stat_profile()` separately and needs to be kept on the `Estimator` level. However the estimator level value is used as default for the fit level arguments. Is that what you meant @registerrier?
- This requires adding `Fit.stat_profile_opts` as well

registerrier:
> Some more feedback from the dev call this morning:
> 
> * The global "reoptimize" option does not really make sense, instead it should be configurable for `confidence()` and `stat_profile()` separately and needs to be kept on the `Estimator` level. However the estimator level value is used as default for the fit level arguments. Is that what you meant @registerrier?

Yes. For `Fit`, `reoptimize` applies only for `confidence`, `stat_profile` and `stat_surface`. Whether the option is limited in scope in `Fit` can be discussed. We can still have a general `Fit.reoptimize` option.

But this option has to be clearly distinct from the `Estimator.reoptimize`. Because the latter applies on the optimization step itself. 

A last option, could be to remove the `reoptimize` option on `confidence`, since after all it does not compute the confidence range if it is set to `False`.

 
> * This requires adding `Fit.stat_profile_opts` as well

Could point. This would alleviate the issue of passing scan values as well.

gammapy/estimators/flux.py
adonath: I moved the `reoptimize` option back to `Estimator.__init__`...

adonath:
@registerrier For now I implemented a compromise here: 
- The backend configuration incl. optimiser options are configured on `__init__`
- Configuration by the estimators, such as `reoptimize`, `n_sigma` etc. is still passed on to the different `Fit.foobar()` methods

To avoid that the PR grows to large, I'd like to merge it soon and work with follow up PRs instead:
- Allow different backends per method, such that Scipy can be used for optimisation
- Improve documentation and code cleanup in `modeling/fit.py`

Issue #3373, opened 2021-05-26
make_theta_squared_table uses the same off position for each observation


maxnoe:
**Gammapy version**

0.18.2

**Bug description**

If no `position_off` is given to `gammapy.makers.utils.make_theta_squared_table`, one is created by mirroring the source position at the pointing position. 

This is only done for the first observation due to how the check is implemented:

```
for obs in observation:
    if position_off is None:
        position_off = ...
``` 
after the first observation, position_off is not None, so all observations use the off position calculated from the first observation.
This yields wrong results when multiple observations with different pointing positions are used (as is usually the case with wobble observations).

**Expected behavior**

Create a new off position for each observation. 

**To Reproduce**

This is using a toy mc approach to trigger this bug:

```python
import numpy as np
import matplotlib.pyplot as plt

import astropy.units as u
from astropy.time import Time
from astropy.table import Table, vstack
from astropy.coordinates import SkyCoord, SkyOffsetFrame


from gammapy.makers.utils import make_theta_squared_table
from gammapy.maps import MapAxis
from gammapy.data import Observation, Observations, EventList, GTI
from gammapy.visualization import plot_theta_squared_table

source = SkyCoord.from_name('Crab')
offset_source = SkyOffsetFrame(origin=source)

observations = []

for offset in [0, 180]:
    
    pointing = source.directional_offset_by(offset * u.deg, 0.4 * u.deg)

    d_ra, d_dec = rng.normal(0, 0.1, (2, 1000)) * u.deg
    reco_radec =  SkyCoord(lat=d_dec, lon=d_ra, frame=offset_source).transform_to('icrs')

    signal = Table()
    signal['RA'] = reco_radec.ra
    signal['DEC'] = reco_radec.dec


    d_ra, d_dec = rng.normal(0, 0.5, (2, 100000)) * u.deg

    offset_pointing = SkyOffsetFrame(origin=pointing)
    reco_radec =  SkyCoord(lat=d_dec, lon=d_ra, frame=offset_pointing).transform_to('icrs')
    background = Table()
    background['RA'] = reco_radec.ra
    background['DEC'] = reco_radec.dec


    events = vstack([signal, background])
    events['TIME'] = Time.now() + np.linspace(0, 10, len(events)) * u.s
    
    obs = Observation(
        obs_id=1,
        events=EventList(events),
        gti=GTI.from_time_intervals([(events['TIME'].min(), events['TIME'].max())]),
[...]

PR #3374, opened 2021-05-26
Fix off position only created for first observation in make_theta_squared_table, fixes 3373

maxnoe:
**Description**

Fix for #3373, see description of issue there

### DIFF
--- a/gammapy/makers/utils.py
+++ b/gammapy/makers/utils.py
@@ -396,12 +396,17 @@ def make_theta_squared_table(
     alpha_tot = np.zeros(len(table))
     livetime_tot = 0
 
+    if position_off is None:
+        create_off = True
+    else:
+        create_off = False
+
     for observation in observations:
         separation = position.separation(observation.events.radec)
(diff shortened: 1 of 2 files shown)


gammapy/makers/utils.py
```diff
@@ -396,12 +396,17 @@ def make_theta_squared_table(
     alpha_tot = np.zeros(len(table))
     livetime_tot = 0
 
+    if position_off is None:
```
adonath, line 399: Could just re-write as `create_off = position_off is None`, but the existing solution is readable as well...

adonath (approved):
Thanks @maxnoe I left one minor inline comment, otherwise looks good to me. Just to confirm: I checked locally and the test fails without the fix...

gammapy/makers/utils.py
maxnoe: right

adonath:
Thanks @maxnoe!

adonath (approved):

maxnoe:
@adonath thanks for taking the time to investigate this with us

PR #3377, opened 2021-05-27
Expand $GAMMAPY_DATA defined in HLI config

Bultako:
This PR expands env variables (i.e. `$GAMMAPY_DATA`) in config fields `BackgroundConfig.exclusion` and `ObservationsConfig.obs_file` of the high-level interface. At this moment, it is no possible to use `$GAMMAPY_DATA` when declaring path values in these fields. 

If the path declared does not exist, it will not be Pydantic but Gammapy functions who will raise the exception, as it is already the case for `ObservationsConfig.datastore`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@ __pycache__
 htmlcov
 .coverage*
 .coverage.subprocess
+.hypothesis
 
 # Sphinx
 _build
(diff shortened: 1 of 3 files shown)


adonath (commented):
Thanks @Bultako! No comments on the code, but can you maybe think of a way how to cover this with a test?

Bultako:
Ok, I will add the tests once https://github.com/gammapy/gammapy-data/pull/13 will be merged.

Bultako:
@adonath 
The test coverage has not changed, since the the new lines got covered with existing tests.

adonath:
@Bultako I'm ok to improve test coverage in future PRs...will merge now.

PR #3378, opened 2021-05-27
Compile extensions against oldest-supported-numpy

maxnoe:
As discussed in the #dev channel in slack, to support older releases of numpy, compilation of the c-extensions must happen against the oldest still supported version of numpy.

See https://pypi.org/project/oldest-supported-numpy/


adonath (approved):
Thanks @maxnoe!

PR #3381, opened 2021-05-31
Fix FoVBackgroundMaker error 

### LINKED ISSUE #3375: FoVBackgroundMaker parameter error not accessible
After the running the `FoVBackgroundMaker`, the error on the fitted parameters is not accessible.

**To reproduce**
Try `print(dataset.background_model.spectral_model.norm.error)` inside the `Data Reduction Loop` in `analysis_2` notebook. It is always `0`

**Reason**
It seems that the error is available at this stage, 
https://github.com/gammapy/gammapy/blob/df3db33e34233369512d57cca5341a7b645802d6/gammapy/makers/background/fov.py#L180

but the info is lost outside the `with models.restore_status` block, eg: in line 187.
I did not understand why, though.

AtreyeeS:
This PR fixes #3375
The covariance is re-written in `Models.restore_status()`, so, the error information would be lost.
Now, restore_status() is applied only on the sky models, so the error on the background model is retained.

`test_fov_bkg_maker_fit` was defined twice, so the first one was not executed. Cleaned up.

### DIFF
--- a/gammapy/makers/background/fov.py
+++ b/gammapy/makers/background/fov.py
@@ -170,14 +170,13 @@ class FoVBackgroundMaker(Maker):
         """
         # freeze all model components not related to background model
 
-        models = dataset.models
+        models = dataset.models.select(tag="sky-model")
 
         with models.restore_status(restore_values=False):
             models.select(tag="sky-model").freeze()
 
             fit = Fit([dataset])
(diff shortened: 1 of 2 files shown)


adonath (approved):
Thanks @AtreyeeS, looks good!

PR #3382, opened 2021-05-31
Fix DEADC definition in datasets.simulate

QRemy:
As discussed in the help channel on slack, the definition of DEADC in gammapy.datasets.simulate is inconsistent with the one in Observation and EventList, this PR solves it.


fabiopintore:
I've seen the discussion on slack. Thank you @QRemy to have fix this issue. I don't have any further comment.

adonath (approved):

PR #3383, opened 2021-06-01
Change  default correlate_off to True for ExcessMapEstimator

registerrier:
**Description**

This pull request reverts the change to have by default `correlate_off=False` for `ExcessMApEstimator`. As noted by @AtreyeeS , this leads to poor results when using stacked datasets.
The main situation where this makes sense is when the `ExcessMapEstimator` is applied on  non stacked datasets. We can therefore leave it as an option but change its default.

**Dear reviewer**

### DIFF
--- a/gammapy/estimators/excess_map.py
+++ b/gammapy/estimators/excess_map.py
@@ -67,10 +67,6 @@ class ExcessMapEstimator(Estimator):
     If a model is set on the dataset the excess map estimator will compute the excess taking into account
     the predicted counts of the model.
 
-    Some background estimation techniques like ring background or adaptive ring background will provide already
-    correlated data for OFF. In the case of already correlated OFF data, the OFF data should not be correlated again,
-    and so the option correlate_off should set to False (default).
-
     Parameters
     ----------
     correlation_radius : ~astropy.coordinate.Angle
@@ -96,7 +92,7 @@ class ExcessMapEstimator(Estimator):
         Apply a mask for the computation.
         A `~gammapy.datasets.MapDataset.mask_fit` must be present on the input dataset
     correlate_off : Bool
-        Correlate OFF events in the case of a MapDatasetOnOff
+        Correlate OFF events in the case of a MapDatasetOnOff. Default is True.
     spectral_model : `~gammapy.modeling.models.SpectralModel`
         Spectral model used for the computation of the flux map. 
         If None, a Power Law of index 2 is assumed (default). 
@@ -113,7 +109,7 @@ class ExcessMapEstimator(Estimator):
         selection_optional=None,
         energy_edges=None,
         apply_mask_fit=False,
-        correlate_off=False,
+        correlate_off=True,
         spectral_model=None,
     ):
         self.correlation_radius = correlation_radius
--- a/gammapy/estimators/excess_map.py
+++ b/gammapy/estimators/excess_map.py
@@ -62,10 +62,19 @@ def convolved_map_dataset_counts_statistics(dataset, kernel, mask, correlate_off
 
 
 class ExcessMapEstimator(Estimator):
-    """Computes correlated excess, sqrt TS (i.e. Li-Ma significance) and errors for MapDatasets.
+    """Computes correlated excess, significance and error maps from a map dataset.
 
-    If a model is set on the dataset the excess map estimator will compute the excess taking into account
-    the predicted counts of the model.
+    If a model is set on the dataset the excess map estimator will compute the
+    excess taking into account the predicted counts of the model.
+
+    ..note::
+        By default the excess estimator correlates the off counts as well to avoid
+        artifacts at the edges of the :term:`FoV` for stacked on-off datasets.
+        However when the on-off dataset has been derived from a ring background
+        estimate, this leads to the off counts being correlated twice. To avoid
+        artifacts and the double correlation, the `ExcessMapEstimator` has to
+        be applied per dataset and the resulting maps need to be stacked, taking
+        the :term:`FoV` cut into account.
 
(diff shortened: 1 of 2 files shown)


mdebony:
Hi,
I don't find the original conversation about the merge request. Would it be possible to have a link to it ?

The reason this parameter was set by default to False, is than it's expected the MapDatasetOnOff is generated by a RingBackgroundMaker. As the ring background estimation technique already correlate OFF, correlate again OFF after the ring background cause than some OFF events will be counts multiple-time. In that case, the OFF count will be too high which can lead to an over estimation of the significance.

registerrier:
> Hi,
> I don't find the original conversation about the merge request. Would it be possible to have a link to it ?
>
Yes it is this one : #3140 

> The reason this parameter was set by default to False, is than it's expected the MapDatasetOnOff is generated by a RingBackgroundMaker. As the ring background estimation technique already correlate OFF, correlate again OFF after the ring background cause than some OFF events will be counts multiple-time. In that case, the OFF count will be too high which can lead to an over estimation of the significance.

 Indeed, you are right. The issue is that one cannot apply it this way when stacking observations because of the various offset cuts. This was noticed by @AtreyeeS . I will paste some example showing the results.

The correct approach in that case should be to keep individual datasets, apply the `ExcessMapEstimator` on each of them and stack the result.
So we still need the functionality not to correlate, but it needs to be properly documented how to do it. And at some point we will need to allow for multiple datasets to be be passed.

adonath:
@registerrier and @AtreyeeS Do we have an example (e.g. in the docs) where the current default leads to a poor result?

AtreyeeS:
Do we want to show ugly results in our docs?

bkhelifi:
In the docstring, there is not yet the recommandations of use of the correlation. Should we add some doc?

adonath:
> Do we want to show ugly results in our docs?

No not at all, I was just asking for a visual example for myself to see the effect once...

adonath:
Maybe you can share a screenshot as well, if there is no example in our docs...

AtreyeeS:
Here is a minimal example that @registerrier  recreated - significance map from simulated wobble observations with a ring background. The edge effects are quite clear.

unknown-8

adonath:
Thanks for the illustration @AtreyeeS! If I understood correctly there is no easy way to make this correct, right? 
- Either we correlate both `counts` and `counts_off` for the stacked case, but then the off counts are twice correlated
- If we don't correlate the `counts_off` again we end up with the artefacts as in @AtreyeeS example
- The only way to make it correct is correlate on the dataset level and stack the datasets again, which is a whole new algorithm

To improve the code structure a bit I was wondering, whether we should introduce `MapDataset.convolve()` and `MapDatasetOnOff.convolve()`?

registerrier:
> Thanks for the illustration @AtreyeeS! If I understood correctly there is no easy way to make this correct, right?
> 
> * Either we correlate both `counts` and `counts_off` for the stacked case, but then the off counts are twice correlated
> * If we don't correlate the `counts_off` again we end up with the artefacts as in @AtreyeeS example
> * The only way to make it correct is correlate on the dataset level and stack the datasets again, which is a whole new algorithm

 That's a good summary. We can possibly show how to do the observation based excess map calculation and stacking in, e.g. a recipe, but this requires some adaptation of the `ExcessMapEstimator` to be done properly as it is necessary to return correlated acceptances as well. This is not the case now.

> To improve the code structure a bit I was wondering, whether we should introduce `MapDataset.convolve()` and `MapDatasetOnOff.convolve()`?

I am not sure I see the exact use case here. We could create an internal method/function to perform correlation of the various quantities stored in the dataset but I don't really see what would be a convolved dataset.

adonath:
> I am not sure I see the exact use case here. We could create an internal method/function to perform correlation of the various quantities stored in the dataset but I don't really see what would be a convolved dataset.

I remember one other use case, where @QRemy extracted probably O(100) spectral datasets of the same size and found that the correlation beforehand sped up the process a lot. But I agree it's a rare use-case and must be taken with care...

adonath:
Just for the record: one could provide the a static method like `ExcessMapEstimator.correlate_dataset(dataset, kernel)`...

gammapy/estimators/excess_map.py
```diff
@@ -67,10 +67,6 @@ class ExcessMapEstimator(Estimator):
     If a model is set on the dataset the excess map estimator will compute the excess taking into account
     the predicted counts of the model.
 
-    Some background estimation techniques like ring background or adaptive ring background will provide already
```
adonath, line 70: Replace with a statement / warning on the fact that it is correlated twice? Whatever this does to the statistics...

adonath:
Again for the record: will show the "correct" way of computing the map in a Gammapy "recipe".

adonath:
@registerrier I think the remaining action item here is to extend the documentation of the `ExcessMapEstimator`?

adonath:
@registerrier and @AtreyeeS I added a note to the docs of the `ExcessMapEstimator` see https://github.com/gammapy/gammapy/pull/3383/files#diff-d51c4d43972776ddc24f7afb00b2dff45134553ab42df8cb573688e25d5682b4R70 

Do you think the will be sufficient?

AtreyeeS:
> I added a note to the docs of the `ExcessMapEstimator` see https://github.com/gammapy/gammapy/pull/3383/files#diff-d51c4d43972776ddc24f7afb00b2dff45134553ab42df8cb573688e25d5682b4R70
> 
> Do you think the will be sufficient?

Thanks @adonath ! This looks clear to me. 
I wonder if we should change our `RingBackground` tutorial to do this accordingly?

PR #3385, opened 2021-06-01
Add Makers object for data reduction chain

QRemy:
This PR adds a Makers object to run makers in a chain. I modified the prototype from @adonath in order to support stacking. In order to have the stacking  that works in parallel the observations are divided into group and a sub-dataset is created for each group. The sub-datasets are built in parallel and then stacked into their parent dataset.


gammapy/makers/reduce.py
```diff
@@ -0,0 +1,187 @@
...
+        if self.stacking:
+            for dataset in results:
+                self.dataset.stack(dataset)
+            self._write(self.dataset)
+            return self.dataset
+
+    @staticmethod
+    def _group_observations(n_groups, observations):
+        """Splint obsevration in multiple groups for parallel run """
+        # TODO: maybe we could introduce observations.select and observations.group
```
adonath, line 162: Thanks @QRemy! Just one preliminary comment here:

Are you sure this "recursive stacking" is required here? From a quick profiling of the `analysis_2.ipynb` notebook it seems we spend 30-40 times more time in computing datasets, then we spend time in stacking them. In this case a simple "sequential" stacking of the datasets, as they "come in" from the parallelised computation should be sufficient and will be much simpler...maybe you can check once if this is true.

However I fully agree with the `TODO` here. Independent of the whether the functionality is used here I think observation selection and grouping is useful in general and ideally it lives on the `Observations` object...

adonath:
Thanks a lot @QRemy, this is a very useful object! I have left two preliminary comments, but will need some time to try out the functionality and comment in more detail...

gammapy/makers/reduce.py
QRemy: The "recursive stacking" is useful if to avoid keeping all the individual datasets in memory, for surveys it is necessary.

gammapy/makers/reduce.py
```diff
@@ -0,0 +1,187 @@
...
+import astropy.units as u
+from astropy.coordinates import SkyCoord, Angle
+from .core import Maker
+from .safe import SafeMaskMaker
+
+
+log = logging.getLogger(__name__)
+
+
+class Makers(Maker):
```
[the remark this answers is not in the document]
QRemy: done

gammapy/makers/reduce.py
QRemy: I pushed a commit using apply_async with callback instead. But I haven't tested the difference in performance yet.

adonath (changes requested):
Thanks @QRemy, I have left a first round of comments.

adonath:
I think one of the main questions here is to define the folder structure, when datasets are written individually. Ideally we can find something that scales to time intervals and event classes. 
- One option is to just stick with the uniques identifier and store the mapping of the id to the obs-id / time interval / event class independently
- We could also choose a hierarchical folder structure like `datasets->obs-1234->dataset-1.fits`
- The same question applies to the default names chosen for the datasets. Ideally it is explicit, but it's unclear whether we can find a uniform solution that works for time intervals / event classes and even multiple instruments...

I need to think about this a bit more, but maybe you have any proposals already @registerrier and @QRemy?

adonath:
Conclusion of the dev meeting today: remove the `write_all` option and just rely on `Datasets.write()`. This has the limitation of requiring all datasets to fit in memory, but avoids to find a "database" solution, with dataset naming etc.

registerrier:
Hello @QRemy . What is the status here?

adonath:
@registerrier I think this PR just misses a few people to try out the functionality, provide feedback and do a review.

registerrier (commented):
Thanks @QRemy , this looks good.

I would suggest not to leave the process number be equal to the number of observations provided.

Looking at the code I wonder whether the `cutout` shouldn't be moved to the `MapDatasetMaker` itself. In a sense, there is no real reason to keep it separate. Any opinion @adonath @QRemy @AtreyeeS ?

Isn't there an issue with memory leak somewhere. I ran into problems with ~400 observations using 4 jobs.

adonath:
@registerrier I think the argument against the cutout in `MapDatasetMaker` was configuration. I.e. there is not reasonable default cutout that works for all instruments (as it also depends on the max offset used...) and not doing the cutout by default is a bad choice as well. So the compromise was just to do the `cutout` explicitly beforehand. I guess the only alternative solution would be to make `cutout_width` a required argument to the `MapDatasetMaker, but I need too think whether there was another argument...

PR #3388, opened 2021-06-08
Add plot_at_energy option for Background3D

AtreyeeS:
This can be useful to see if the background model is asymmetric.

### DIFF
--- a/docs/tutorials/data/cta.ipynb
+++ b/docs/tutorials/data/cta.ipynb
@@ -39,10 +39,10 @@
    "execution_count": null,
    "metadata": {
     "execution": {
-     "iopub.execute_input": "2021-11-02T15:00:14.782635Z",
-     "iopub.status.busy": "2021-11-02T15:00:14.781905Z",
-     "iopub.status.idle": "2021-11-02T15:00:15.060506Z",
-     "shell.execute_reply": "2021-11-02T15:00:15.059972Z"
+     "iopub.execute_input": "2021-11-03T14:30:27.093058Z",
+     "iopub.status.busy": "2021-11-03T14:30:27.092314Z",
+     "iopub.status.idle": "2021-11-03T14:30:27.368026Z",
+     "shell.execute_reply": "2021-11-03T14:30:27.368343Z"
     }
    },
    "outputs": [],
@@ -56,10 +56,10 @@
    "execution_count": null,
    "metadata": {
     "execution": {
-     "iopub.execute_input": "2021-11-02T15:00:15.063627Z",
-     "iopub.status.busy": "2021-11-02T15:00:15.063241Z",
-     "iopub.status.idle": "2021-11-02T15:00:15.239797Z",
-     "shell.execute_reply": "2021-11-02T15:00:15.240084Z"
+     "iopub.execute_input": "2021-11-03T14:30:27.371321Z",
+     "iopub.status.busy": "2021-11-03T14:30:27.370967Z",
+     "iopub.status.idle": "2021-11-03T14:30:27.553125Z",
+     "shell.execute_reply": "2021-11-03T14:30:27.553451Z"
     }
    },
    "outputs": [],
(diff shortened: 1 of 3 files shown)


gammapy/irf/background.py
```diff
@@ -135,6 +135,32 @@ def to_2d(self):
     def peek(self, figsize=(10, 8)):
         return self.to_2d().peek(figsize)
 
+    def plot_at_energy(self, energy=None, ax=None, add_cbar=True, **kwargs):
+        """ Plot the background rate in Field of view co-ordinates at a given energy.
+
+        Parameters
+        -----------
+        energy : `~astropy.units.Quantity`
```
registerrier, line 143: Could it be possible to pass a list of energies and have the result plotted as plot_grid?

registerrier (commented):
Thanks @AtreyeeS ! This looks good. 

I suppose that generally we would need to homogenize a bit the plot methods on `Background2D` and `Background3D`. At the moment, it is not possible to plot an image of `Background2D` but this might be of interest as well. Right?

gammapy/irf/background.py
AtreyeeS: I can do this, but these are not maps, so have to re-implement a part of the code in `Maps.plot_grid` to make an automatic selection of the number of rows and columns in the grid.

AtreyeeS:
> I suppose that generally we would need to homogenize a bit the plot methods on `Background2D` and `Background3D`. At the moment, it is not possible to plot an image of `Background2D` but this might be of interest as well. Right?

Yes, I was just being lazy. I will think a bit how to plot this and add it

AtreyeeS:
I added the plotting in grid option.
For plotting `Background2D`, I think the simplest would be to add `Background2D.to_3d()` and then calling `.plot_at_energy()` for that.

AtreyeeS:
Hi! I adapted to use `meshgrid` instead of `imshow`.
Also added `Background2d.to_3d()`, and use this to plot `Background2d`.

docs/tutorials/data/cta.ipynb
```diff
@@ -37,7 +37,14 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "metadata": {},
+   "metadata": {
```
adonath, line 40: @bultako Is there anything we can do about this? There is auto-generated meta data added, probably dependent non the jupyter or python version. This is a bit annoying...

gammapy/irf/background.py
```diff
@@ -135,6 +135,64 @@ def to_2d(self):
...
+
+        for i, ee in enumerate(energy):
+            if len(energy) == 1:
+                ax = axes
+            else:
+                ax = axes.flat[i]
+            bkg = self.evaluate(energy=ee)
+            Z = bkg.squeeze().value
+            im = ax.pcolormesh(X, Y, Z, **kwargs)
+            ax.set_xlabel(x1.name + " [" + str(x1.unit) + "]")
```
adonath, line 184: You can use `MapAxis.format_plot_xaxis(ax=ax)` here, this will take care of everything...

gammapy/irf/background.py
```diff
@@ -158,6 +216,48 @@ class Background2D(BackgroundIRF):
...
+            data[:, :, i] = val
+
+        return Background3D(
+            axes=[self.axes["energy"], fov_lon, fov_lat], data=data, unit=self.unit,
+        )
+
+    def plot_at_energy(self, energy=None, ax=None, add_cbar=True, ncols=3, **kwargs):
+        """ Plot the background rate in Field of view co-ordinates at a given energy.
+
+            Parameters
```
adonath, line 245: Fix indentation...

adonath (changes requested):
Thanks @AtreyeeS, I have left a few more comments...

docs/tutorials/data/cta.ipynb
adonath: I decided to remove the `execution` meta data keyword during stripping https://github.com/gammapy/gammapy/commit/370fc3e67b8ed2a573cf993f4310710cb737b5cc @AtreyeeS Please rebase and update the PR. Thanks!

gammapy/irf/background.py
```diff
@@ -135,6 +135,64 @@ def to_2d(self):
...
+            figsize=figsize,
+            gridspec_kw={"hspace": 0.2, "wspace": 0.3},
+        )
+
+        x1 = self.axes[1]
+        x2 = self.axes[2]
+
+        x = x1.edges.value
+        y = x2.edges.value
+        X, Y = np.meshgrid(x, y)
```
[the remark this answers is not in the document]
AtreyeeS: Maybe I am doing something wrong, but `self.axes.get_coord()` is node centered. We need the edge values here.

gammapy/irf/background.py
```diff
@@ -136,6 +134,63 @@ def to_2d(self):
...
+        **kwargs : dict
+            Keyword arguments passed to `~matplotlib.pyplot.pcolormesh`.
+        """
+        import matplotlib.pyplot as plt
+
+        n = len(energy)
+        cols = min(ncols, n)
+        rows = 1 + (n - 1) // cols
+        width = 12
+        figsize = (width, width * rows / cols)
```
adonath, line 159: I think it's important to keep `figsize` as parameter to the function, however you can set a default just like in `Map.plot_grid()`

AtreyeeS: Done

adonath (approved):
Thanks @AtreyeeS, no further comments from my side. CI fails are unrelated...

PR #3389, opened 2021-06-16
Remove find_roots from TSMapEstimator for performance

QRemy:
Follow up PR after #3379. Restore previous behaviour in TSMapEstimator for performance.


adonath (approved):
Thanks a lot @QRemy! I added a test on a reference pixel here https://github.com/gammapy/gammapy/commit/0671947a9eb900dc135536ad0ac6f7671fac7715 Please rebase and adjust the test value so that the test passed, otherwise looks good !

adonath (approved):
Thanks a lot for the quick implementation @QRemy!

Issue #3376, opened 2021-05-27
Parameter type not assigned in PiecewiseNormSpectralModel


judevin:
**Gammapy version**
0.18.2

**Bug description**
When using the PiecewiseNormSpectralModel for the FoVBagroundModel, the parameter type is not assigned (which should be "spectral"). That creates problems for serialization due to the parameters_unique_names which uses the par.type and the error is therefore "expected str instance, NoneType found".

**Expected behavior**
The par.type assigned to "spectral"

**To Reproduce**

Example:
```python3
models_bkg = Models([])
Custom_model = PiecewiseNormSpectralModel(energy = geom.axes["energy"].edges)
bkg_model = FoVBackgroundModel(dataset_name=dataset.name, spectral_model=Custom_model)
dataset.models = bkg_models
dataset = maker_bkg.run(dataset)
model_bkg.append(dataset.background_model)
models_bkg.write("./Models_bkg.yaml")
```
```
Error:
~/opt/anaconda3/envs/gammapy-0.18.2/lib/python3.7/site-packages/gammapy/modeling/models/core.py in write(self, path, overwrite, full_output, write_covariance)
    335                 format="ascii.fixed_width", delimiter="|", overwrite=overwrite
    336             )
--> 337             self.write_covariance(base_path / filecovar, **kwargs)
    338             self._covar_file = filecovar
    339 

~/opt/anaconda3/envs/gammapy-0.18.2/lib/python3.7/site-packages/gammapy/modeling/models/core.py in write_covariance(self, filename, **kwargs)
    404 
    405         """
--> 406         names = self.parameters_unique_names
    407         table = Table()
    408         table["Parameters"] = names

~/opt/anaconda3/envs/gammapy-0.18.2/lib/python3.7/site-packages/gammapy/modeling/models/core.py in parameters_unique_names(self)
    248             for par in model.parameters:
    249                 components = [model.name, par.type, par.name]
--> 250                 name = ".".join(components)
    251                 names.append(name)
    252 

TypeError: sequence item 1: expected str instance, NoneType found
```
**Other information**
For the time being, the problem can be bypassed with:

for par in models_bkg.parameters:
    par._type = 'spectral'

adonath:
Thanks a lot @judevin, I can confirm this bug! On closer inspection this might be an issue that affects all models with dynamically generated parameters, such as `NaimaSpectralModel` as well. The fix is exactly as you proposed to define the parameter type "by hand" for these cases...

[...]

PR #3390, opened 2021-06-16
Fix parameter type for PiecewiseNormSpectralModel  and NaimaSpectralModel

QRemy:
Fix parameter type for PiecewiseNormSpectralModel  and NaimaSpectralModel.
Resolve issue #3376.

### DIFF
--- a/gammapy/modeling/models/core.py
+++ b/gammapy/modeling/models/core.py
@@ -40,8 +40,6 @@ class Model:
 
     def __init__(self, **kwargs):
         # Copy default parameters from the class to the instance
-        for p in self.default_parameters:
-            p._type = self._type
         default_parameters = self.default_parameters.copy()
 
         for par in default_parameters:
@@ -54,6 +52,14 @@ class Model:
 
             setattr(self, par.name, par)
         self._covariance = Covariance(self.parameters)
+
+    def __getattribute__(self, name):
+        value = object.__getattribute__(self, name)
+        
+        if isinstance(value, Parameter):
+            return value.__get__(self, None)
+    
+        return value
 
     @property
     def type(self):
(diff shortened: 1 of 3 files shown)


registerrier (commented):
Thanks @QRemy ! 

This solves the issue, but this will appear again when users define models with explicit `Parameter` instantiation in their `__init__` method. Can't we find a way to have them set on init? For instance, adding a `SpectralModel.__init__()` method that will set all parameters type  to `'spectral'`? Could that work?

QRemy:
@registerrier I moved the fix to the `Model` base class such as `parameter._type` is set to match `model._type`

adonath:
Thanks @QRemy! I think we could possibly manually invoke the descriptor protocol, by implementing `__getattribute__` on the `Model` base class like so:

```python3
from gammapy.modeling import Parameter, Parameters
from gammapy.modeling.models import Model

def __getattribute__(self, name):
    value = object.__getattribute__(self, name)
    
    if isinstance(value, Parameter):
        return value.__get__(self, None)

    return value
    
Model.__getattribute__ = __getattribute__

class MyModel(Model):
    _type = "my-model"
    par_a = Parameter("par_a", value=1)    

class MyOtherModel(Model):
    _type = "my-other-model"
    
    def __init__(self):
        par_a = Parameter("par_a", value=1)
        self.default_parameters = Parameters([par_a])
        super().__init__()
        
m = MyModel()
m.par_a
print(m.par_a.type)

m_other = MyOtherModel()
m_other.par_a
print(m_other.par_a.type)
```
I'm not 100% whether this has any side effects, but the toy implementation above seems to work.

The problem with the setting of `_type` on init is that we cannot reset the parameter like:
```python3
model.par = Parameter("par", value=1)
```
Which currently works...

adonath (commented):

adonath (approved):
Looks good!

PR #3393, opened 2021-06-22
Implement TimeMapAxis class

registerrier:
**Description**

This pull request introduces a `TimeAxis` object that contains sorted time intervals that are not expected to be contiguous.
  
For now, it implements the `coord_to_idx` method and a few `from_something` class methods. 

I have created a new file `gammapy.maps.axes.py` since the `geom.py` file is already very long. The `MapAxis` and `MapAxes` code could be moved here as well to clarify a bit.

**Dear reviewer**

This is a draft for discussions.

### DIFF
--- a/gammapy/maps/axes.py
+++ b/gammapy/maps/axes.py
@@ -63,9 +63,6 @@ class TimeMapAxis:
             delta = self._edges_min[1:] - self._edges_max[:-1]
             if np.any(delta.to_value("s")<0):
                 raise ValueError("TimeMapAxis: time intervals must not overlap.")
-#        else:
-#            self._edges_min.reshape((1,))
-#            self._edges_max.reshape((1,))
 
         if interp != "lin":
             raise ValueError("TimeMapAxis: non-linear scaling scheme are not supported yet.")
@@ -199,10 +196,9 @@ class TimeMapAxis:
         idx = np.atleast_1d(self.coord_to_idx(coord))
 
         valid_pix = np.where(idx!=INVALID_INDEX.int)
-        pix = np.asarray(idx, dtype = 'float')
-        if pix.shape == ():
-            pix = pix.reshape((1,))
+        pix = np.atleast_1d(idx).astype('float')
 
+        # TODO: is there the equivalent of np.atleast1d for astropy.time.Time?
         if coord.shape == ():
             coord = coord.reshape((1,))
         relative_time = coord[valid_pix]-self.reference_time
--- a/gammapy/maps/axes.py
+++ b/gammapy/maps/axes.py
@@ -39,8 +39,7 @@ class TimeMapAxis:
         edges_min = u.Quantity(edges_min, ndmin=1)
         edges_max = u.Quantity(edges_max, ndmin=1)
 
-        if not isinstance(reference_time, Time):
-            raise TypeError(f"TimeAxis reference time must be Time object. Got {type(reference_time).__name__}.")
+        reference_time = Time(reference_time)
 
         invalid = [_.unit.name for _ in (edges_min, edges_max) if not _.unit.is_equivalent("d")]
         if len(invalid)>0:
--- a/gammapy/maps/axes.py
+++ b/gammapy/maps/axes.py
@@ -123,6 +123,7 @@ class TimeAxis:
     def upsample(self):
         raise NotImplementedError
 
+    #TODO: how configurable should that be? column names?
     @classmethod
     def from_table(cls, table, name="time"):
         if "TIMESYS" not in table.meta:
@@ -131,6 +132,15 @@ class TimeAxis:
         else:
             scale = table.meta["TIMESYS"]
         format = "mjd"
+
         tmin = Time(table["time_min"], scale=scale, format=format)
         tmax = Time(table["time_max"], scale=scale, format=format)
+        return cls(tmin, tmax, name)
+
+    @classmethod
+    def from_gti(cls, gti, name="time"):
+        """Create a time axis from an input GTI."""
+        tmin = gti.time_start
+        tmax = gti.time_stop
+
         return cls(tmin, tmax, name)
--- a/gammapy/maps/axes.py
+++ b/gammapy/maps/axes.py
@@ -208,7 +208,6 @@ class TimeMapAxis:
             coord = coord.reshape((1,))
         relative_time = coord[valid_pix]-self.reference_time
 
-        print(valid_pix)
         scale = interpolation_scale(self._interp)
         valid_idx = idx[valid_pix]
         s_min = scale(self._edges_min[valid_idx])
--- a/gammapy/maps/axes.py
+++ b/gammapy/maps/axes.py
@@ -121,25 +121,38 @@ class TimeMapAxis:
         return idx
 
     def coord_to_pix(self, coord):
-        time = Time(coord[..., np.newaxis])
-        relative_time = time-self.reference_time
+        """Transform from time to coordinate to pixel position.
+
+        Pixels of time values falling outside time bins will be
+        set to -1.
+
+        Parameters
+        ----------
+        coord : `~astropy.time.Time`
+            Array of axis coordinate values.
+
+        Returns
+        -------
+        pix : `~numpy.ndarray`
+            Array of pixel positions.
+        """
+        idx = self.coord_to_idx(coord)
+
+        valid_pix = np.where(idx!=-1)
+        pix = np.asarray(idx, dtype = 'float')
+        if pix.shape == ():
+            pix = pix.reshape((1,))
+
+        if coord.shape == ():
+            coord = coord.reshape((1,))
+        relative_time = coord[valid_pix]-self.reference_time
 
         scale = interpolation_scale(self._interp)
-        s_min = scale(self._edges_min)
-        s_max = scale(self._edges_max)
+        s_min = scale(self._edges_min[valid_pix])
+        s_max = scale(self._edges_max[valid_pix])
         s_coord = scale(relative_time.to(self._edges_min.unit))
 
-        delta_plus = s_coord - s_min
-        delta_minus = s_coord - s_max
-
-        mask = np.logical_and(delta_plus>0, delta_minus<=0)
-        idx = np.asanyarray(np.argmax(mask, axis=-1))
-        valid = np.any(mask, axis=-1)
-        idx[~valid] = -1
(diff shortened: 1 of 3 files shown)


adonath:
Thanks a lot @registerrier! How do we proceed here? Does @QRemy tale over the PR or do you continue yourself a bit?

adonath:
+1 for the new `maps/axes.py` file!

gammapy/maps/axes.py
```diff
@@ -0,0 +1,145 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+
+import numpy as np
+from astropy.time import Time
+
+
+class TimeAxis:
```
adonath, line 7: Even if it doesn't inherit from `MapAxis` (yet) I think we should call it a `TimeMapAxis`

gammapy/maps/axes.py
```diff
@@ -0,0 +1,145 @@
...
+    time_max : `~astropy.time.Time`
+        Array of edge time values.
+    interp : str
+        Interpolation method used to transform between axis and pixel
+        coordinates.  Valid options are 'log', 'lin', and 'sqrt'.
+    name : str
+        Axis name
+    """
+    node_type = "edges"
+    def __init__(self, time_min, time_max, name="", interp="lin",):
```
adonath, line 26: I think we could safely assume the default name as `"time"`, or even hard-code it to `"time"` (not sure about the latter, but I think there will never be two time axes in the same object...)

gammapy/maps/axes.py
```diff
@@ -0,0 +1,145 @@
...
+    @property
+    def name(self):
+        return self._name
+
+    @property
+    def nbin(self):
+        return len(self._time_min)
+
+    @property
+    def time_min(self):
```
adonath, line 56: Here I'm not 100% sure whether we should introduce the additional properties with names adjusted to the time content. I think for now it;s definitely more important to stay consistent with the `MapAxis` object, because we definitely want to avoid any kind of extra handling...

gammapy/maps/axes.py
```diff
@@ -0,0 +1,145 @@
...
+        Array of edge time values.
+    time_max : `~astropy.time.Time`
+        Array of edge time values.
+    interp : str
+        Interpolation method used to transform between axis and pixel
+        coordinates.  Valid options are 'log', 'lin', and 'sqrt'.
+    name : str
+        Axis name
+    """
+    node_type = "edges"
```
adonath, line 25: Here I'm not sure whether we need a new type, like "intervals". Currently "edges" means a contiguous representation.

gammapy/maps/axes.py
```diff
@@ -0,0 +1,145 @@
...
+        self._time_min = Time(time_min)
+        self._time_max = Time(time_max)
+
+        self._interp = interp
+
+        self._pix_offset = -0.5
+        self._nbin = len(time_min)
+
+    @property
+    def name(self):
```
adonath, line 48: Add docstrings...

gammapy/maps/axes.py
```diff
@@ -0,0 +1,145 @@
...
+        time = Time(time[..., np.newaxis])
+        delta_plus = (time - self.time_min).value > 0.
+        delta_minus = (time - self.time_max).value <= 0.
+        mask = np.logical_and(delta_plus, delta_minus)
+
+        idx = np.asanyarray(np.argmax(mask, axis=-1))
+        idx[~np.any(mask, axis=-1)] = -1
+        return idx
+
+    def time_to_pix(self, coord):
```
adonath, line 99: I think this should return the correct interpolation, however I'm not sure yet how to achieve this...

gammapy/maps/axes.py
```diff
@@ -0,0 +1,145 @@
...
+        if len(invalid)>0:
+            raise TypeError(f"TimeAxis edges must be Time objects. Got {invalid}")
+
+        if not len(time_min) == len(time_max):
+            raise ValueError("Time min and time max must have the same length.")
+
+        if not (np.all(time_min == time_min.sort()) and np.all(time_max == time_max.sort()) ):
+            raise ValueError("TimeAxis: edge values must be sorted")
+
+        self._time_min = Time(time_min)
```
adonath, line 39: It could be advantageous to always go to the same internal representation, it depends a bit on which internal methods we need and whether they support `Time` objects. This is e.g. not the case for interpolation...

gammapy/maps/axes.py
```diff
@@ -0,0 +1,145 @@
...
+        return str_.expandtabs(tabsize=2)
+
+    def upsample(self):
+        raise NotImplementedError
+
+    #TODO: how configurable should that be? column names?
+    @classmethod
+    def from_table(cls, table, name="time"):
+        if "TIMESYS" not in table.meta:
+            print("No TIMESYS information. Assuming UTC scale.")
```
adonath, line 129: Replace by logging statement...

adonath (changes requested):
Thanks a lot @registerrier, I have left a first round of comments.

registerrier:
A general comment here about the API w.r.t. `MapAxis`.
As @adonath mentions this is not an `edge` axis node type. This is really an interval type. 
Therefore `axis.node` becomes somewhat irrelevant and `axis.edges` is ambiguous. There are now two types of edges: `axis.edges_min` and `axis.edge_max`. 
We could define a new type of `IntervalMapAxis` designed for non-contiguous axes. The time axis and label axis would then be specific cases of such axes.

Technically the main functionalities of `MapAxis` do not all fit in this context. See this [gist](https://gist.github.com/registerrier/fe861e08848f9232a99f45633afcf88b)
|

adonath:
@registerrier Thanks a lot for this overview! In general I was wondering whether we could handle the "intervals" case just with a special `node_type`, but I presume this is not the case...it requires special handling in many cases...

gammapy/maps/axes.py
registerrier: What is the best approach?
We could also define a `TimeMapAxis.time_ref` and use all times as quantities similarly to what is done for `GTI` objects. Then interpolation is easier.

adonath: I agree, defining a reference time and relying on this internally is maybe the most promising approach...

gammapy/maps/axes.py
adonath: The main question to solve here is what should happen when users request a time, which is in between two intervals. It could either fail, or still interpolate...

gammapy/maps/axes.py
```diff
@@ -0,0 +1,237 @@
...
+    def center(self):
+        """Return `~astropy.time.Time` at interval centers."""
+        return self.time_mid
+
+    @property
+    def bin_width(self):
+        """Return time interval width."""
+        return self.time_delta
+
+    def __repr__(self):
```
adonath, line 135: Maybe add reference time and total time range to the repr...

adonath:
I noticed that when working with `Time` objects and `MapCoord` there is some problem with `np.broadcast_arrays()`:
```python3
from astropy.time import Time
import numpy as np

energy = np.array([1, 3, 10]) * u.TeV
time = Time(['1999-01-01T00:20:00.123456789', '2005-01-01T00:30:00', '2010-01-01T00:30:00'])
print(np.broadcast_arrays(energy.reshape((-1, 1)), time, subok=True))
```
Which prints:
```
[<Quantity [[ 1.,  1.,  1.],
           [ 3.,  3.,  3.],
           [10., 10., 10.]] TeV>, array([[<Time object: scale='utc' format='isot' value=1999-01-01T00:20:00.123>,
        <Time object: scale='utc' format='isot' value=2005-01-01T00:30:00.000>,
        <Time object: scale='utc' format='isot' value=2010-01-01T00:30:00.000>],
       [<Time object: scale='utc' format='isot' value=1999-01-01T00:20:00.123>,
        <Time object: scale='utc' format='isot' value=2005-01-01T00:30:00.000>,
        <Time object: scale='utc' format='isot' value=2010-01-01T00:30:00.000>],
       [<Time object: scale='utc' format='isot' value=1999-01-01T00:20:00.123>,
        <Time object: scale='utc' format='isot' value=2005-01-01T00:30:00.000>,
        <Time object: scale='utc' format='isot' value=2010-01-01T00:30:00.000>]],
      dtype=object)]
```
So the time is not parsed as a `Time` object anymore. I wanted to remove the forced broadcasting in `MapCoord` for a long time anyway, so I'll go ahead and try to prepare a PR...

gammapy/maps/axes.py
registerrier: I think the time content are useful to obtain real time objects. We can still add the `edges_min` and `edges_max` properties but they will return quantities relative to the reference time.

gammapy/maps/axes.py
registerrier: Currently the approach returns `INVALID_INDEX.int` in between two bins.
The only supported interpolation scheme is `lin` because there is no guarantee that the `edges` will be positive.

gammapy/maps/axes.py
registerrier: It is implemented now.

gammapy/maps/axes.py
registerrier: Changed to `intervals`

adonath (dismissed):
Thanks a lot @registerrier, for now I don't have any further comments. I'd be happy to merge the PR as is and improve the interplay of the axis with the `Map` object in future PRs.

gammapy/maps/axes.py
```diff
@@ -0,0 +1,368 @@
...
+        coordinates.  For now only 'lin' is supported.
+    """
+    node_type = "intervals"
+    def __init__(self, edges_min, edges_max, reference_time, name="time", interp="lin"):
+        self._name = name
+
+        edges_min = u.Quantity(edges_min, ndmin=1)
+        edges_max = u.Quantity(edges_max, ndmin=1)
+
+        if not isinstance(reference_time, Time):
```
adonath, line 42: Just call `Time(reference_time)` here and let Astropy raise the error...

registerrier: Done

adonath (approved):
Thanks @registerrier, I think this is a very solid basis to build on. Let's merge the PR and start using the `TimeMapAxis` a bit and then we adapt step by step and integrate it completely in the `Map` API.

adonath:
I'll submit one follow up PR to move all `MapAxis` classes into the `axes.py` file...

PR #3395, opened 2021-06-22
Remove counts data caching from MapDataset

### LINKED ISSUE #3387: MapDataset._counts_data  property should be reset when counts are modified
The `MapDataset._counts_data` property is used to convert `counts` to `float` for likelihood calculations. 
It is currently defined as `@lazyproperty`. The direct consequence is that when `counts` is modified (e.g. by `MapDataset.fake`) the counts data actually used for likelihood calculations are silently different from the ones actually defined on the `MapDataset`. 
This could be solved by adding a getter/setter definition for `counts` that would reset the lazy property.

adonath:
**Description**

This pull request fixes #3387. For now I decided to remove the caching completely and to accept the performance penalty. I haven't done any detailed profiling, but I expect a difference in order of 10-20% on the stat evaluation. However the npred computation will probably still dominate.

### DIFF
--- a/gammapy/datasets/map.py
+++ b/gammapy/datasets/map.py
@@ -902,13 +902,9 @@ class MapDataset(Dataset):
 
         return ax_spatial, ax_spectral
 
-    @lazyproperty
-    def _counts_data(self):
-        return self.counts.data.astype(float)
-
     def stat_sum(self):
         """Total likelihood given the current model parameters."""
-        counts, npred = self._counts_data, self.npred().data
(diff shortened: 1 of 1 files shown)


registerrier (approved):
Thanks @adonath ! 

Conversely one could also change the `cash_sum_cython` to take a vector of int and avoid the type conversion here. Or at least to do the conversion only when `counts.dtype` is not `int`

adonath:
I think to take the optimal decision here one would need to do a lot of careful profiling. Maybe we postpone this for now and just go with the simplest solution...

PR #3396, opened 2021-06-22
Fix stacking of cutouts 

### LINKED ISSUE #3368: Stacking of Map cutouts is erroneous
** Bug description **

If we have 2 maps which are themselves cutouts (eg; maps obtained from our `MapMaker` process), and then we perform a cutout on them

`map1 = dataset1.counts.cutout(dataset1.counts._geom.center_skydir, width=2.0*u.deg)`
`map2 = dataset2.counts.coutout(dataset2.counts._geom.center_skydir, width=2.0*u.deg)`

and then want to stack `map1` and `map2` onto a large target Geom, 

```
counts = Map.from_geom(geom)
cutout = dataset1.counts.cutout(dataset1.counts._geom.center_skydir, width=2.0*u.deg)
counts.stack(cutout)
```

 just stacks in one corner according to the pixel index, forgetting the geometries.

** Reason **

I discussed this privately with @adonath and the reason is that in
https://github.com/gammapy/gammapy/blob/master/gammapy/maps/wcsnd.py#L789 we only check for the alignment of the WCS and not the actual position. So when stacking a cutout from a cutout into the initial geom, the WCS projections are aligned, but the slices are not valid anymore. The slices from the smallest cutout are only valid for the bigger cutout and not the initial parent geometry.

This is a reminder issue to fix before 1.0 release

adonath:
**Description**

This pull request fixes #3368. The fix includes recomputing the slices using `overlap_slices` from `astropy` in a new method called `WcsGeom.cutout_slices()`. This is slightly slower for stacking and cutouts however it should still be fast enough and correct now. This also allows to remove the `cutout_info` completely and simplify the code in this regard.

The PR also adds a unit test, however in general it might be a good idea to improve the testing to cover all cases like partial cutouts etc.

### DIFF
--- a/gammapy/maps/tests/test_wcs.py
+++ b/gammapy/maps/tests/test_wcs.py
@@ -241,24 +241,14 @@ def test_cutout_info():
     geom = WcsGeom.create(skydir=(0, 0), npix=10)
     position = SkyCoord(0, 0, unit="deg")
     cutout_geom = geom.cutout(position=position, width="2 deg")
-    assert cutout_geom.cutout_info["parent-slices"][0].start == 3
-    assert cutout_geom.cutout_info["parent-slices"][1].start == 3
 
-    assert cutout_geom.cutout_info["cutout-slices"][0].start == 0
-    assert cutout_geom.cutout_info["cutout-slices"][1].start == 0
+    cutout_info = cutout_geom.cutout_slices(geom)
 
-    header = cutout_geom.to_header()
-    assert "PSLICE1" in header
-    assert "PSLICE2" in header
-    assert "CSLICE1" in header
-    assert "CSLICE2" in header
+    assert cutout_info["parent-slices"][0].start == 3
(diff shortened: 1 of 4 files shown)


AtreyeeS:
Just to confirm? Cutouts computed previously will still have the same error, but maps and cutouts created anew should be fine?

adonath:
@AtreyeeS No, the change also fixes the behaviour backwards. Previously we relied on the `cutout_info`, but after having thought about this I realised this is not necessary and slices can be just re-computed based on the `WCS` information instead. Unfortunately I didn't have the idea before and introduced the `cutout_info` instead. So old files will have the `cutout_info` in the header, however it will be just un-used and in new files it is not written anymore.

AtreyeeS (approved):
Thanks a lot @AtreyeeS ! This fixes the issue as far as I could test

adonath:
Thanks @AtreyeeS, the remaining doc test fail is unrelated...

PR #3397, opened 2021-06-23
Use string for model name in npred_signal()

### LINKED ISSUE #3386: `Dataset.npred_signal()` should take model name
`Dataset.npred_signal()` takes a model as argument, whereas in most other places (eg:`FluxPointEstimator` we access models by their names. This create some confusion.

@adonath shall I put in a PR to change this?

AtreyeeS:
This PR fixes #3386


AtreyeeS:
I imagine that the doc test fail is unrelated.

adonath:
Thanks @AtreyeeS, please remove the unrelated `test_stack.ipynb` notebook.

gammapy/datasets/map.py
```diff
@@ -407,10 +407,13 @@ def npred_signal(self, model=None):
         """
         npred_total = Map.from_geom(self._geom, dtype=float)
 
-        for evaluator in self.evaluators.values():
-            if model is evaluator.model:
-                return evaluator.compute_npred()
+        if model is not None:
+            try:
+                return self.evaluators[model].compute_npred()
+            except:
```
adonath, line 413: It's important to always specify the kind of error to catch like `except KeyError`. Also the new error message you give is not really an improvement (it missed which key is not contained). My proposal would be to just keep the `KeyError`, that is raised if the `model_name` does  not exist.

gammapy/datasets/map.py
```diff
@@ -395,8 +395,8 @@ def npred_signal(self, model=None):
 
         Parameters
         -------------
-        model: `~gammapy.modeling.models.SkyModel`, optional
-            Sky model to compute the npred for.
+        model: str
```
adonath, line 398: Maybe rename to `model_name` to make the change more explicit.

adonath (changes requested):
Thanks @AtreyeeS, I have left two minor inline comments...

adonath:
Thanks @AtreyeeS, once the remaining test fails are solved this is real to merge...

adonath (approved):
Thanks @AtreyeeS! Remaining test fails are unrelated...

PR #3399, opened 2021-06-24
Implement WcsGeom.from_aligned

### LINKED ISSUE #3384: Show reference point in WcsGeom.__repr__
Currently when printing a `WcsGeom` the reference point is not shown in the print out, so two `WcsGeom` can look equivalent in the printout, while they are actually not equivalent. Here is an example to reproduce:
```python3
from gammapy.maps import WcsGeom

geom_1 = WcsGeom.create(skydir=(0, 0), width=10)
geom_2 = WcsGeom.create(skydir=(4.75, -4.75), width=10, refpix=(1, 1))

print(geom_1)
print(geom_2)
print(geom_1 == geom_2)
``` 
Which prints:
```
WcsGeom

	axes       : ['lon', 'lat']
	shape      : (20, 20)
	ndim       : 2
	frame      : icrs
	projection : CAR
	center     : 0.0 deg, 0.0 deg
	width      : 10.0 deg x 10.0 deg

WcsGeom

	axes       : ['lon', 'lat']
	shape      : (20, 20)
	ndim       : 2
	frame      : icrs
	projection : CAR
	center     : 0.0 deg, 0.0 deg
	width      : 10.0 deg x 10.0 deg

False
```
A good solution would be to include the reference point of the WCS projection in the printout.

adonath:
**Description**

This pull request implement a `WcsGeom.from_aligned(geom, skydir, width)` method which allows users to create a new aligned geometry for stacking cutouts. This makes it possibly to stack cutouts into and any arbitrary aligned geometry, without needing to know the parent one. I'll add a corresponding `MapDataset.from_aligned()` method. 

This PR also fixes #3384.

### DIFF
--- a/gammapy/maps/wcs.py
+++ b/gammapy/maps/wcs.py
@@ -491,7 +491,7 @@ class WcsGeom(Geom):
         return cls.create(
             skydir=tuple(geom.wcs.wcs.crval),
             npix=tuple(npix),
-            refpix=(xref - 1, yref),
+            refpix=(xref - 1, yref),  # TODO: understand why the -1 is required for consistency with .cutout()
             frame=geom.frame,
             binsz=tuple(geom.pixel_scales.deg),
             axes=geom.axes,
--- a/gammapy/maps/wcs.py
+++ b/gammapy/maps/wcs.py
@@ -1048,6 +1048,7 @@ class WcsGeom(Geom):
         axes = ["lon", "lat"] + [_.name for _ in self.axes]
         lon = self.center_skydir.data.lon.deg
         lat = self.center_skydir.data.lat.deg
+        lon_ref, lat_ref = self.wcs.wcs.crval
 
         return (
             f"{self.__class__.__name__}\n\n"
@@ -1058,6 +1059,7 @@ class WcsGeom(Geom):
             f"\tprojection : {self.projection}\n"
             f"\tcenter     : {lon:.1f} deg, {lat:.1f} deg\n"
             f"\twidth      : {self.width[0][0]:.1f} x {self.width[1][0]:.1f}\n"
+            f"\twcs ref    : {lon_ref:.1f} deg, {lat_ref:.1f} deg\n"
         )
 
     def to_odd_npix(self, max_radius=None):
--- a/gammapy/maps/wcs.py
+++ b/gammapy/maps/wcs.py
@@ -13,6 +13,7 @@ from astropy.wcs.utils import (
     celestial_frame_to_wcs,
     proj_plane_pixel_scales,
     wcs_to_celestial_frame,
+    skycoord_to_pixel
 )
 from gammapy.utils.array import round_up_to_odd
 from .geom import (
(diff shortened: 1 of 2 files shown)


gammapy/maps/wcs.py
```diff
@@ -459,6 +459,45 @@ def footprint(self):
...
+        """
+        width = _check_width(width) * u.deg
+        xref, yref = geom.to_image().coord_to_idx(skydir)
+        npix = np.rint(width / geom.pixel_scales).astype(int)
+        xref = -int(xref) + npix[0] // 2 + geom.wcs.wcs.crpix[0]
+        yref = -int(yref) + npix[1] // 2 + geom.wcs.wcs.crpix[1]
+        return cls.create(
+            skydir=tuple(geom.wcs.wcs.crval),
+            npix=tuple(npix),
+            refpix=(xref - 1, yref),
```
adonath, line 494: Honestly I have no idea why the `-1` here is required. I added it so the resulting geometry is equivalent to one returned by `.cutout()`. See the [test vs cutout](https://github.com/gammapy/gammapy/pull/3399/files#diff-33810c8943a3a35af445d931195c7b17ae552ac859c334dedd1f502b7c14ccedR155). I think this is non-ideal, however the behaviour is established with a test, so possibly OK to keep for now...

adonath:
One remaining question is what to do for `HpxGeom`? By definition the pixels in a HPX projection are aligned, so `.from_aligned()` will basically just call into `.create()`...

gammapy/maps/wcs.py
```diff
@@ -459,6 +459,45 @@ def footprint(self):
...
+        geom : `~WcsGeom`
+            An aligned WCS geometry object with specified size and center.
+
+        """
+        width = _check_width(width) * u.deg
+        xref, yref = geom.to_image().coord_to_idx(skydir)
+        npix = np.rint(width / geom.pixel_scales).astype(int)
+        xref = -int(xref) + npix[0] // 2 + geom.wcs.wcs.crpix[0]
+        yref = -int(yref) + npix[1] // 2 + geom.wcs.wcs.crpix[1]
+        return cls.create(
```
registerrier, line 491: Does it really work like this? 
If one uses a "TAN" projection, the center will be changed. No? 
Is the WCS really the same as that obtained from a cutout for such a projection? 

The test works, but what happens if you take a significantly larger distance for the cutout geometry.

registerrier (commented):
Thanks @adonath . This looks good, but I wonder when it should break. If the new position is far enough from the center parent geometry in TAN projection, the deviation must be large enough to have an inconsistent WCS at some point. I am not sure the tests here can really check that. Am I wrong?

gammapy/maps/wcs.py
adonath: Good point. Indeed this can only work up to certain numerical precision. I'll add an edge case as test as well.

registerrier: Why not just rely on `cutout` here? It guarantees to have an aligned WCS.

adonath: Because `.cutout()` will clip at the data edges? But in general you are right, we should do the same operation as in `Cutout2D`. I'll take a look at the implementation and see how I can improve. I think the current implementation (also because of the spurious `-1` is not yet correct...)

adonath: I managed to find the consistent implementation with `Cutout2D` and added a test case using a TAN projection and a reference point further away. I think this should be fine for now.

adonath:
@registerrier Will go ahead and merge this, please open an issue if further comments / ideas for tests.

PR #3400, opened 2021-06-24
Add containment option to PSFMap.get_psf_kernel()

adonath:
**Description**

This pull request adds a containment option to `PSFMap.get_psf_kernel()`. This option allows users to choose the `max_radius` based on the containment fraction of the kernel. The option can be overwritten with `max_radius` option.
For now I set the default to 99.9% containment, with a lower default I saw bigger changes in the test reference values. However we could optimise this in future. I also introduce a `PSF_CONTAINMENT` variable which allows users to modify the value for fitting if needed.


adonath:
@registerrier I'll go ahead and merge this PR and continue with #3366.

PR #3401, opened 2021-06-24
Use dataset._geom in SafeMaskMaker

AtreyeeS:
`SafeMaskMaker` would fail in the absence of a counts map. adapted to now use `dataset._geom` instead of `dataset.counts.geom`


adonath (approved):
Thanks @AtreyeeS, looks reasonable! It's probably a rare use-case but it might happen to run the `SafeMaskMaker` without counts...

adonath:
Thanks @AtreyeeS, CI fails are unrelated...

PR #3403, opened 2021-06-25
Add sparse option to get_coord() methods

adonath:
**Description**
This PR adds a `get_coord(sparse=)` option to be able to work with sparse coordinate arrays and rely on broadcasting instead. This simplifies the code in a few places (just adapt a few places so far) and makes sure it is always the fastest option.


adonath:
This is a clean up PR and doesn't come with any change in API, will go ahead and merge. CI is green...

PR #3404, opened 2021-06-29
Add is_pointlike property on irfs

registerrier:
**Description**

This pull request adds a `IRF.is_pointlike` property to be able to check whether an IRF is pointlike or not.

**Dear reviewer**

### DIFF
--- a/gammapy/irf/core.py
+++ b/gammapy/irf/core.py
@@ -35,13 +35,14 @@ class IRF:
 
     default_interp_kwargs = dict(bounds_error=False, fill_value=None,)
 
-    def __init__(self, axes, data=0, unit="", meta=None):
+    def __init__(self, axes, data=0, unit="", is_pointlike = False, meta=None):
         axes = MapAxes(axes)
         axes.assert_names(self.required_axes)
         self._axes = axes
         self.data = data
         self.unit = unit
         self.meta = meta or {}
+        self._is_pointlike = is_pointlike
 
     @property
     @abc.abstractmethod
@@ -52,6 +53,11 @@ class IRF:
     @abc.abstractmethod
     def required_axes(self):
         pass
+
+    @property
+    def is_pointlike(self):
+        """Whether the IRF is pointlike of full containment."""
+        return self._is_pointlike
 
     @property
     def is_offset_dependent(self):
@@ -357,7 +363,10 @@ class IRF:
         axes = MapAxes.from_table(table=table, format=format)[cls.required_axes]
         column_name = IRF_DL3_HDU_SPECIFICATION[cls.tag]["column_name"]
         data = table[column_name].quantity[0].transpose()
-        return cls(axes=axes, data=data.value, meta=table.meta, unit=data.unit)
+        is_pointlike = False
+        if "HDUCLAS3" in table.meta and table.meta["HDUCLAS3"]=="POINT-LIKE":
+            is_pointlike = True
+        return cls(axes=axes, data=data.value, meta=table.meta, unit=data.unit, is_pointlike=is_pointlike)
 
     def to_table(self, format="gadf-dl3"):
         """Convert to table
@@ -379,6 +388,8 @@ class IRF:
             spec = IRF_DL3_HDU_SPECIFICATION[self.tag]
             # TODO: add missing required meta data!
             table.meta["HDUCLAS2"] = spec["hduclas2"]
+            if self.is_pointlike:
+                table.meta["HDUCLAS3"] = "POINT-LIKE"
             table[spec["column_name"]] = self.quantity.T[np.newaxis]
         else:
             raise ValueError(f"Not a valid supported format: '{format}'")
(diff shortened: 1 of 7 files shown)


gammapy/irf/tests/test_effective_area.py
```diff
@@ -104,3 +104,12 @@ def test_wrong_axis_order():
...
             axes=[energy_axis_true, offset_axis], data=data, unit="cm2"
         )
+
+def test_aeff2d_pointlike():
+    filename = "$GAMMAPY_DATA/joint-crab/dl3/magic/run_05029748_DL3.fits"
+
+    aeff = EffectiveAreaTable2D.read(filename)
+    hdu = aeff.to_table_hdu()
+
+    assert aeff.is_pointlike is True
```
adonath, line 114: Simplify to `assert aeff.is_pointlike`

gammapy/irf/tests/test_energy_dispersion.py
```diff
@@ -204,3 +204,12 @@ def test_get_bias_energy():
...
     thresh_lo = edisp.get_bias_energy(0.1)
     assert_allclose(thresh_lo.to("TeV").value, 0.9174, rtol=1e-4)
+
+def test_edisp2d_pointlike():
+    filename = "$GAMMAPY_DATA/joint-crab/dl3/magic/run_05029748_DL3.fits"
+
+    aeff = EnergyDispersion2D.read(filename)
+    hdu = aeff.to_table_hdu()
+
+    assert aeff.is_pointlike is True
```
adonath, line 214: See above...

gammapy/makers/spectrum.py
```diff
@@ -57,6 +57,9 @@ def make_exposure(self, geom, observation):
         exposure = super().make_exposure(geom, observation, use_region_center=self.use_region_center)
 
         if self.containment_correction:
+            if observation.aeff.is_pointlike:
```
adonath, line 60: Note the access to `observation.aeff` reloads the IRF from disk, to avoid this, maybe work with a local variable. I also think we should keep the meta info as much as possible. So maybe we copy over the meta data to the reduced maps?
So you could access e.g.`exposure.meta["is_pointlike"]` here...

gammapy/irf/core.py
```diff
@@ -35,13 +35,14 @@ class IRF:
...
 
-    def __init__(self, axes, data=0, unit="", meta=None):
+    def __init__(self, axes, data=0, unit="", is_pointlike = False, meta=None):
         axes = MapAxes(axes)
         axes.assert_names(self.required_axes)
         self._axes = axes
         self.data = data
         self.unit = unit
         self.meta = meta or {}
+        self._is_pointlike = is_pointlike
```
adonath, line 45: Maybe treat the `is_pointlike` info as part of the `meta` data? I guess the motivation for the separate argument is to stay independent of the serialisation format (which is good!), however we can introduce arbitrary internal meta data and corresponding keywords as we like. The advantage is that the meta data can be easier copied over to the reduced data as well and we keep the option for a unified data structure of `IRF` and `Map` in future.

adonath (changes requested):
Thanks @registerrier, the is a useful addition. I have left some inline comments..

gammapy/irf/core.py
adonath: We could possibly add a table in the docs with "allowed" meta keywords for IRFs and a corresponding description. Like:

| Keyword           | Description                 | Type  |
|-------------------|-----------------------------|-------|
| is_pointlike      | Whether IRF is pointlike    | bool  |
| safe_energy_range | Recommend safe energy range | tuple |

gammapy/makers/tests/test_spectrum.py
```diff
@@ -23,6 +23,12 @@ def observations_hess_dl3():
     obs_ids = [23523, 23526]
     return datastore.get_observations(obs_ids)
 
+@pytest.fixture
```
adonath, line 26: Add missing `requires_data()`...

gammapy/makers/utils.py
```diff
@@ -60,10 +62,11 @@ def make_map_exposure_true_energy(pointing, livetime, aeff, geom, use_region_cen
     )
 
     exposure = (exposure * livetime).to("m2 s")
-    meta = {"livetime": livetime}
+    meta = {"livetime": livetime,
+            "is_pointlike": aeff.is_pointlike}
 
     if not use_region_center:
-        data = np.average(exposure.value, axis=1, weights=weights)
+         data = np.average(exposure.value, axis=1, weights=weights)
```
adonath, line 69: Isn't this an indentation error?

adonath (approved):
Thanks @registerrier, I left two more minor comments. After those have been addressed, CI should turn green...

gammapy/makers/utils.py
registerrier: Oops, it is

adonath (approved):

PR #3405, opened 2021-07-01
Improve mask_fit handling in analysis_3d notebook

registerrier:
**Description**

This pull request clarifies the usage of the mask_fit in analysis_3d. It removes the manual creation of the map which is not the recommended approach.

**Dear reviewer**

### DIFF
--- a/docs/tutorials/analysis/3D/analysis_3d.ipynb
+++ b/docs/tutorials/analysis/3D/analysis_3d.ipynb
@@ -254,7 +254,12 @@
     "\n",
     "We choose a point source model with an exponential cutoff power-law spectrum.\n",
     "\n",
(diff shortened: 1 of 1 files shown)


adonath:
Thanks @registerrier, I agree the analysis tutorial is not the right place for this. However I think showing how to create a mask manually is still useful, so maybe we show it in the mask tutorial instead?

registerrier:
Yes indeed. The mask_map notebook lacks completely something for energy. So we need to show how to define a mask in energy manually and with the specific method there. And also how to combine the spatial part and spectral part. I can add that.

I think we also need to refer more frequently to this notebook to help people understand the roles of the different masks and how to create/handle them. I will add a reference here.

adonath (approved):

adonath:
Thanks @registerrier!

PR #3406, opened 2021-07-01
Add scan specification to the Parameter object

adonath:
**Description**

As proposed in https://github.com/gammapy/gammapy/pull/3234#issuecomment-849669197 this pull request adds all the scan configuration to the `Parameter` object:

- Add `.scan_min` and `.scan_max`
- Add `.scan_n_values`, and `.scan_n_sigma`
- Add `.scan_values` and `.interp`

The logic is as following:
- If the parameter error is defined the `.scan_min` and `.scan_max` values are generated from `scan_n_sigma` and the parameter value
- If `.scan_min` and `.scan_max` are defined explicitly, those values are taken
- If `.scan_values` is defined explicitly, those values are taken

I decided to remove the scan specification from the `ParameterEstimator` and just rely on the specification on the `Parameter` object. The other estimators, where the norm parameter is created internally still have the specification for the norm scan. One could create a mechanism to predefine the norm `Parameter` instead or possibly translate the information from the `amplitude` parameter of the model. However this should happen in a follow up PR.

This PR supersedes #3234.

### DIFF
--- a/docs/tutorials/api/fitting.ipynb
+++ b/docs/tutorials/api/fitting.ipynb
@@ -119,7 +119,11 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "scipy_opts = {\"method\": \"L-BFGS-B\", \"options\": {\"ftol\": 1e-4, \"gtol\": 1e-05}, \"backend\": \"scipy\"}\n",
+    "scipy_opts = {\n",
+    "    \"method\": \"L-BFGS-B\",\n",
+    "    \"options\": {\"ftol\": 1e-4, \"gtol\": 1e-05},\n",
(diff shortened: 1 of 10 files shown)


QRemy:
Thanks @adonath, #3234 also offered the possibility to set the parameters norm_min and norm_max for  the fit and not only for the profile scan, but this can be done in the follow-up PR that auto set the scan_values for the norm parameter created internally.

adonath:
Thanks @QRemy, indeed! On the `Parameter` level `.min` and `.scan_min` etc. are already separate so this can be easily implemented in a follow up PR.

adonath:
I also updated `Fit.stat_surface` to rely on the `.scan_values` properties...

PR #3408, opened 2021-07-02
Return optimize_results and covariance_results in Fit.run() 

QRemy:
Modify Fit.run() to return a dict containing optimize_results and covariance_results.

### DIFF
--- a/docs/tutorials/analysis/time/light_curve_simulation.ipynb
+++ b/docs/tutorials/analysis/time/light_curve_simulation.ipynb
@@ -386,7 +386,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "result.parameters.to_table()"
+    "result[\"optimize_result\"].parameters.to_table()"
    ]
   },
   {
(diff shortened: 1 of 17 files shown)


docs/tutorials/analysis/time/light_curve_simulation.ipynb
```diff
@@ -431,7 +431,25 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.7.0"
+   "version": "3.7.3"
+  },
```
adonath, line 435: Can you check your version of `nbconvert`? Like:
```
import nbconvert
print(nbconvert.__version__)
```
I think in more recent versions this meta data is not present anymore...

QRemy: ok upgraded to 6.1.0

adonath:
Thanks @QRemy, I fixed the merge conflict caused by merging #3406. Once CI is green I'll go ahead and merge. Instead of the `dict` we could also consider actually renaming the current `FitResult` to e.g. `FitStepResult` and create a new `FitResult` object which bundles the results from the different steps.  But this only becomes useful if we add custom methods e.g. `FitResult.__str__` and / or `FitResult.write()`, it will be a separate PR anyway...

adonath (approved):
Thanks @QRemy!

PR #3409, opened 2021-07-02
Support iminuit v2.0

adonath:
**Description**

This pull request adds the implementation of the `iminuit>2.0` backend. For now I added it in a separate `iminuit2.py` file, but it seems from the Gammapy side the change is backwards compatible. The remaining question is what we do:
- Support both `iminuit` and `iminuit>2.0` and choose internally
- Support both `iminuit` and `iminuit>2.0` as separate backends
- Remove the support for `iminuit<2.0`

My preferred option would be the last one, as it reduces the maintenance effort for Gammapy in future. Depending on which option we choose, I'll adapt the iminuit file.

**Dear reviewer**


adonath:
@bkhelifi @registerrier @QRemy @AtreyeeS @luca-giunti Any preference on which versions we should support?

bkhelifi:
As you, Axel, I prefer the last option, ie Remove the support for `iminuit<2.0`, for the maintenance.
I think this move is wise, as this upgrade will be unavailable in the future.

AtreyeeS:
I confess that I have not explored either in great details. Is there any specific case to keep support for `iminuit<2.0`? Otherwise, I imagine it is simpler to remove it.

registerrier (commented):
Thanks @adonath!

Looks good. One comment below regarding `errordef`.
It is interesting to see that iminuit now uses CL rather than sigmas.

gammapy/modeling/iminuit2.py
```diff
@@ -0,0 +1,218 @@
...
+
+def setup_iminuit(parameters, function, store_trace=False, **kwargs):
+    from iminuit import Minuit
+    minuit_func = MinuitLikelihood(function, parameters, store_trace=store_trace)
+
+    pars, errors, limits = make_minuit_par_kwargs(parameters)
+
+    minuit = Minuit(minuit_func.fcn, name=list(pars.keys()), **pars)
+    minuit.tol = kwargs.pop("tol", 0.1)
+    minuit.errordef = kwargs.pop("errordef", 1)
```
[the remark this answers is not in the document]
adonath: I guess in general users are free to overwrite the `Dataset.stat_sum()` method and use a different normalisation, but I agree we could just hard-code the `errordef` and document it. If users implement a custom fit statistic, they have to follow the convention.

adonath:
@AtreyeeS As commented here https://github.com/gammapy/gammapy/issues/3118#issuecomment-873110572, we will loose the status info for the contour computation, but this is probably acceptable...

adonath:
@HDembinski I'm now seeing two test fails in our CI: https://github.com/gammapy/gammapy/pull/3409/checks?check_run_id=2990082892#step:9:1952 I can reproduce the fails locally, however only when using `iminuit==2.7.0` as used in the CI. When using `iminuit==2.6.1` the fails are not present. Are you aware of any fixed bug that could cause this difference? What would be your recommended handling here? Thanks!

adonath:
Currently there is a blocking issue, see https://github.com/scikit-hep/iminuit/issues/669. I'll check for a workaround.

adonath:
Meanwhile the bug has bee fixed and a new release has been made. I'll try with `iminuit==2.8.0` now...

adonath:
I just updated to `iminiut=2.8.0` and everything works fine now. However `iminiut=2.8.0` is not available on conda forge yet and can only be installed via `pip`.

adonath:
`iminuit>2.8` is now available on conda, however I see test fails in the CI (https://github.com/gammapy/gammapy/pull/3409/checks?check_run_id=3653998586#step:9:1909), that I cannot reproduce locally. It seems to depend on the specific environment. I'll spend another hour to try to reproduce and understand this, but this might be a blocker.

adonath:
I failed to reproduce the error even in the almost exact environment (same version of Numpy, Astropy, Scipy and iminuit), also from the CI it seems it is an issue specific to the Linux environment and Python 3.7. The issue is almost impossible to debug. So what do we do here @registerrier and @bkhelifi ? Just disable the falling test? All other error estimates seems to be working, except the ones for the `FoVBackgroundMaker`...

adonath:
@registerrier I started to look into the issue of the FoV background estimator and I saw a weird behavior of almost arbitrary fit results, when unfreezing the tilt parameter. The results would change when the estimator was run on the same dataset repeatedly. Then I realized the dataset only has one energy bin and the `tilt` is never constrained. So the test in Gammapy is meaningless and should be changed. However I'd prefer a follow up PR for this...

gammapy/datasets/tests/test_flux_points.py
```diff
@@ -123,7 +123,7 @@ def test_stat_profile(dataset):
         )
 
         ts_diff = profile["stat_scan"] - result.total_stat
-        assert_allclose(ts_diff, [174.358204, 0., 174.418515], rtol=1e-2, atol=1e-7)
+        assert_allclose(ts_diff, [112.368266, 0., 112.416683], rtol=1e-2, atol=1e-7)
```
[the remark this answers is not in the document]
adonath: I added scan value check, and there one can see the values shifted a little bit, causing the change in the likelihood values as well.

gammapy/estimators/tests/test_flux_point_estimator.py
```diff
@@ -433,29 +433,29 @@ def test_run_pwl_parameter_range(fpe_pwl):
...
-    pl.amplitude.max=1e-8
+    pl.amplitude.min = 0
+    pl.amplitude.max = 1e-12
 
     fp = fpe.run(datasets)
     table_with_bounds = fp.to_table()
 
     actual = table_with_bounds["norm"].data
-    assert_allclose(actual, [3.215947e-02, 3.939055e-02, 5.551115e-09], rtol=1e-3)
+    assert_allclose(actual, [0., 0., 0.], atol=1e-2)
```
adonath, line 443: I decide to use a more stable check here, just testing whether the norm is consistent with zero.

gammapy/estimators/tests/test_flux_point_estimator.py
```diff
@@ -433,29 +433,29 @@ def test_run_pwl_parameter_range(fpe_pwl):
...
 
-    actual = table_with_bounds["norm_err"].data
-    assert_allclose(actual, [251.490704, 280.37361 , 404.162784], rtol=1e-2)
+    actual = table_with_bounds["norm_errp"].data
+    assert_allclose(actual, [212.593368, 298.383045, 449.951747], rtol=1e-2)
```
adonath, line 446: As the minimizer is not in a minimum the covariance is not a stable estimate. I change to the confidence based error instead.

adonath:
@registerrier Please do a final review...

gammapy/estimators/tests/test_flux_point_estimator.py
registerrier: Looks good.

registerrier (approved):
Thanks @adonath ! This is good to go.

It is good that the origin of the discrepancies is understood. Adding a better test for tilt fitting will be a good change.

adonath:
@registerrier and @bkhelifi I'll go ahead and merge this now. I think we should just re-run the validation, remaining issue can be fixed in follow up PRs.

PR #3412, opened 2021-07-06
Change definition of disk edge parameter

adonath:
**Description**

This pull request changes the definition of the `edge` parameter to a relative fraction of `r_0`. This solves the problem mentioned in https://github.com/gammapy/gammapy/pull/3229#issuecomment-874640578. To make the change more explicit I decided to rename the parameter to `edge_width` as well.

### DIFF
--- a/gammapy/modeling/models/spatial.py
+++ b/gammapy/modeling/models/spatial.py
@@ -592,7 +592,8 @@ class DiskSpatialModel(SpatialModel):
         Increases counter-clockwise from the North direction.
     edge : `~astropy.coordinates.Angle`
         Width of the edge. The width is defined as the range within the
-        smooth edges of the model drops from 95% to 5% of its amplitude.
+        smooth edges of the model drops from 95% to 5% of its amplitude
+        and is given as fraction of r_0.
     frame : {"icrs", "galactic"}
         Center position coordinate frame
     """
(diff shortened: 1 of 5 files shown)


gammapy/modeling/models/spatial.py
```diff
@@ -590,9 +590,10 @@ class DiskSpatialModel(SpatialModel):
     phi : `~astropy.coordinates.Angle`
         Rotation angle :math:`\phi`: of the major semiaxis.
         Increases counter-clockwise from the North direction.
-    edge : `~astropy.coordinates.Angle`
+    edge_width : float
         Width of the edge. The width is defined as the range within the
-        smooth edges of the model drops from 95% to 5% of its amplitude.
+        smooth edge of the model drops from 95% to 5% of its amplitude.
```
luca-giunti, line 595: Thanks @adonath! The only comment I have is that this sentence should probably be re-formulated

luca-giunti (changes requested):

gammapy/modeling/models/spatial.py
luca-giunti: Or maybe just "within which"

luca-giunti (approved):

adonath:
Thanks @luca-giunti!

PR #3413, opened 2021-07-06
Introduce inheritance for Estimator classes

adonath:
**Description**

This pull request introduce inheritance for the various `Estimator` classes. The goal is to reduce code duplication and inconsistencies. It include the following changes:
- Remove energy range definition from `FluxEstimator`, the energy range I just taken from the passed dataset
- Make `FluxPointsEstimator` inherit from `FluxEstimator`
- Make `LightCurveEstimator` inherit from `FluxPointsEstimator`
- Unify `n_sigma_ul`

### DIFF
--- a/gammapy/estimators/excess_map.py
+++ b/gammapy/estimators/excess_map.py
@@ -2,11 +2,11 @@
 import copy
 import logging
 import numpy as np
-import astropy.units as u
 from astropy.convolution import Tophat2DKernel
 from astropy.coordinates import Angle
 from gammapy.datasets import MapDataset, MapDatasetOnOff
-from gammapy.maps import Map, MapAxis
+from gammapy.maps import Map
+from gammapy.modeling.models import PowerLawSpectralModel
 from gammapy.stats import CashCountsStatistic, WStatCountsStatistic
 from .core import Estimator
 from .utils import estimate_exposure_reco_energy
@@ -128,6 +128,10 @@ class ExcessMapEstimator(Estimator):
         self.selection_optional = selection_optional
         self.energy_edges = energy_edges
         self.correlate_off = correlate_off
+
+        if spectral_model is None:
+            spectral_model = PowerLawSpectralModel()
+
         self.spectral_model = spectral_model
 
     @property
--- a/gammapy/estimators/excess_map.py
+++ b/gammapy/estimators/excess_map.py
@@ -130,7 +130,7 @@ class ExcessMapEstimator(Estimator):
         self.correlate_off = correlate_off
 
         if spectral_model is None:
-            spectral_model = PowerLawSpectralModel()
+            spectral_model = PowerLawSpectralModel(index=2)
 
         self.spectral_model = spectral_model
 
--- a/gammapy/estimators/flux.py
+++ b/gammapy/estimators/flux.py
@@ -56,7 +56,7 @@ class FluxEstimator(ParameterEstimator):
 
     def __init__(
         self,
-        source,
+        source=0,
         norm_min=0.2,
         norm_max=5,
         norm_n_values=11,
@@ -183,7 +183,7 @@ class FluxEstimator(ParameterEstimator):
             models[self.source].spectral_model = model
 
             datasets.models = models
-            result.update(self._parameter_estimator.run(datasets, model.norm))
+            result.update(super().run(datasets, model.norm))
             result["sqrt_ts"] = self.get_sqrt_ts(result["ts"], result["norm"])
 
         return result
--- a/gammapy/estimators/flux.py
+++ b/gammapy/estimators/flux.py
@@ -64,7 +64,7 @@ class FluxEstimator(Estimator):
         norm_n_values=11,
         norm_values=None,
         n_sigma=1,
-        n_sigma_ul=3,
+        n_sigma_ul=2,
         selection_optional=None,
         fit=None,
         # TODO: why the different default here?
--- a/gammapy/estimators/flux.py
+++ b/gammapy/estimators/flux.py
@@ -45,11 +45,11 @@ class FluxEstimator(ParameterEstimator):
             * "ul": estimate upper limits.
             * "scan": estimate fit statistic profiles.
 
-        Default is None so the optionnal steps are not executed.
+        Default is None so the optional steps are not executed.
     fit : `Fit`
         Fit instance specifying the backend and fit options.
     reoptimize : bool
-        Re-optimize other free model parameters. Default is True.
+        Re-optimize other free model parameters. Default is False.
     """
     tag = "FluxEstimator"
     _available_selection_optional = ["errn-errp", "ul", "scan"]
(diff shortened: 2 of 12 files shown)


gammapy/estimators/flux.py
```diff
@@ -58,50 +56,30 @@ class FluxEstimator(Estimator):
...
         norm_n_values=11,
         norm_values=None,
         n_sigma=1,
-        n_sigma_ul=3,
+        n_sigma_ul=2,
         selection_optional=None,
         fit=None,
         # TODO: why the different default here?
-        reoptimize=True
+        reoptimize=False
```
bkhelifi, line 69: The docstring should be adapted accordingly.
I am wondering where we could explain in detail the details of this option, that is in fact quite import...

gammapy/estimators/flux.py
```diff
@@ -195,16 +160,20 @@ def run(self, datasets):
 
         model = self.get_scale_model(models)
 
+        energy_min, energy_max = datasets.energy_ranges
+
         with np.errstate(invalid="ignore", divide="ignore"):
-            result = self.get_reference_flux_values(model.model, energy_min, energy_max)
+            result = self.get_reference_flux_values(
```
bkhelifi, line 166: The general big difficulty for LC of AGN is the change of energy threshold with runs. If 1 dataset is a group of runs within a GTI interval, we will have one threshold per dataset. The question arising is which energy threshold for the whole? If we take the lowest energy threshold (or safe energy threshold), we necessarily do an extrapolation of the model. If we use the highest energy threshold, no extrapolation is done and only the 'measured photons' are used, which is safer at the expense of a strong reduction of the energy range. 
There is no ideal solution and this depends of the Use Case, ie a possible change of the model with time at the extremity of the energy ranges...
I am wondering how to manage this difficulty, as it seems to be 'hardcoded' here. I hope that my comment is clear...

gammapy/estimators/flux.py
adonath: Yes, I think we should try to avoid to repeat the documentation. Maybe a forward reference to the `Fit` class is a good option, where the `reoptimise` is explained in detail.

gammapy/estimators/flux.py
adonath: The current behaviour is to take the full energy range from the dataset as energy range to quote the flux. However the models is fit taking the `mask_safe` into account. The information on the `mask_safe` energy range is lost later, which is non-ideal. WE have to think about where to keep this information. 

This also means the flux is possibly extrapolated, but I think it's only behaviour that makes sense, because the energy integrated flux must be comparable between the different time intervals, even for varying energy range. 

What is currently not possible is to support a different model per time interval (e.g. varying spectral index...), however this could implemented later on the basis of #3418.

adonath: @bkhelifi A good option to maintain the energy range info for `FluxPoints` might be to add a `.energy_range_safe` attribute which is extract from the `dataset.energy_range_safe` attribute. When change the data type to a `Map` we could store safe energy ranges depending on position / time etc.

gammapy/estimators/excess_map.py
```diff
@@ -128,6 +128,10 @@ def __init__(
         self.selection_optional = selection_optional
         self.energy_edges = energy_edges
         self.correlate_off = correlate_off
+
+        if spectral_model is None:
+            spectral_model = PowerLawSpectralModel()
```
registerrier, line 133: Maybe explicitly pass `index=2` in case the default is changed?

gammapy/estimators/flux.py
```diff
@@ -195,16 +160,20 @@ def run(self, datasets):
 
         model = self.get_scale_model(models)
 
+        energy_min, energy_max = datasets.energy_ranges
```
registerrier, line 163: This is unrelated to the PR directly but this method is returning something different from the `MapDataset.energy_range` which takes into account masks. This is confusing.

gammapy/estimators/flux.py
registerrier: This discussion sheds some light also on the question of the `FluxMaps` format for flux points. It can be that the actual safe range is different from one estimate to the other so that technically the energy bins used for fitting could be different e.g. at different times. 
This we can't handle with the `FluxMaps` design but we could do it with a flat table.

One safe solution could be to reject energy bins not fully covered by the `mask`.

gammapy/estimators/tests/test_flux.py
```diff
@@ -84,54 +91,43 @@ def test_flux_estimator_fermi_with_reoptimization(fermi_datasets):
...
+
+    result = estimator.run(datasets)
 
     assert_allclose(result["norm"], 1.218139, atol=1e-3)
     assert_allclose(result["ts"], 527.492959, atol=1e-3)
     assert_allclose(result["norm_err"], 0.095496, atol=1e-3)
     assert_allclose(result["norm_errn"], 0.093204, atol=1e-3)
     assert_allclose(result["norm_errp"], 0.097818, atol=1e-3)
-    assert_allclose(result["norm_ul"], 1.525773, atol=1e-3)
+    assert_allclose(result["norm_ul"], 1.418475, atol=1e-3)
```
registerrier, line 109: Any idea for the change here?

registerrier (approved):

I guess the questions about actual energy range and what to reoptimize and how to document it are general questions that apply even without this PR. It would be good to find a solution for that but I suppose this is not directly connected to the PR.

On a side note, the `MapDataset.energy_range` and `Datasets.energy_ranges` should be harmonized.

gammapy/estimators/tests/test_flux.py
adonath: Yes, the default for `n_sigma_ul` changed from 3 to 2. I guess for consistency we should always chose 2 and the change is OK?

gammapy/estimators/flux.py
adonath: Good catch, indeed we should harmonise and mirror the API from the `MapDataset` level on the `Datasets` level.

gammapy/estimators/excess_map.py
adonath: OK

gammapy/estimators/flux.py
adonath: As I sketched in my comment before, I think one way to handle this could be to return a `Map` in the `energy_range_safe` methods and store the resampled map on the `FluxPoints` as well. This allows us to document how the energy threshold changes with position and time and if needed we could convert the measured fluxes based on the best fit model to the safe energy range afterwards and still store it in a flattened table. Do you see what I mean? I think this approach applies to `FluxMaps` as well...

adonath:
Thanks @registerrier and @bkhelifi, the CI fails are unrelated. I'll go ahead and merge this PR and open an new issue on the unification of energy range handling and possibility to store energy range maps e.g. one the `FluxPoints` / `FluxMaps` objects.

PR #3415, opened 2021-07-07
Improve mask fit documentation

registerrier:
**Description**

This pull request adapt significantly the mask_map.ipynb tutorial to explain in detail how to create a `mask_fit` and clearly distinguish from the exclusion mask.
The Fermi dataset is therefore loaded early in the notebook and we show how create a energy mask as well as a region mask and how to combine them on this dataset using directly the `dataset` `geom`.

I have introduced also two properties on the `MapDataset` for clarity to the user. One is the `.energy_range` which return the full energy range for fitting, the other is the `safe_energy_range` which returns only the safe range range. 
Note that the latter change, breaks the current definition of `energy_range`. We might possibly adapt to `fit_energy_range` so that everything is clear (although the latter might not contain the safe range...).

Comments @AtreyeeS @QRemy @adonath @bkhelifi ? Should it help users who currently simply do stuff on the `mask_safe`?

**Dear reviewer**

### DIFF
--- a/gammapy/datasets/map.py
+++ b/gammapy/datasets/map.py
@@ -337,22 +337,36 @@ class MapDataset(Dataset):
         """Shape of the counts or background data (tuple)"""
         return self._geom.data_shape
 
-    @property
-    # TODO: make this a method to support different methods?
-    def energy_range(self):
-        """Energy range defined by the safe mask"""
+    def _energy_range(self, use_mask_fit = True):
+        """Compute the energy range with or without the mask fit."""
         energy = self._geom.axes["energy"].edges
         energy_min, energy_max = energy[:-1], energy[1:]
 
-        if self.mask_safe is not None:
-            if self.mask_safe.data.any():
-                mask = self.mask_safe.data.any(axis=(1, 2))
+        if use_mask_fit:
+            mask = self.mask
+        else:
+            mask = self.safe_mask
+
+        if mask is not None:
+            if mask.data.any():
+                mask = mask.data.any(axis=(1, 2))
             else:
                 return None, None
         else:
             mask = None
 
         return u.Quantity([energy_min[mask].min(), energy_max[mask].max()])
+
+    @property
+    def energy_range(self):
+        """Energy range defined by the full mask (mask_safe and mask_fit)."""
+        return self._energy_range(use_mask_fit=True)
+    
+    @property
+    def safe_energy_range(self):
+        """Energy range defined by the full mask (mask_safe and mask_fit)."""
+        return self._energy_range(use_mask_fit=False)
+
 
     def npred(self):
         """Predicted source and background counts
--- a/gammapy/datasets/map.py
+++ b/gammapy/datasets/map.py
@@ -337,15 +337,10 @@ class MapDataset(Dataset):
         """Shape of the counts or background data (tuple)"""
         return self._geom.data_shape
 
-    def _energy_range(self, use_mask_fit = True):
+    def _energy_range(self, mask = None):
         """Compute the energy range with or without the mask fit."""
         energy = self._geom.axes["energy"].edges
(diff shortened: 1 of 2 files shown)


adonath (commented):
Thanks @registerrier, this is really helpful! 

A few preliminary comments on the `energy_range` definition:
- Maybe we stick with the mask nomenclature like: `energy_range`, `energy_range_safe`, `energy_range_fit`
- To avoid code duplication one could introduce `Map.mask_energy_range()`

gammapy/datasets/map.py
```diff
@@ -337,23 +337,37 @@ def data_shape(self):
...
         return u.Quantity([energy_min[mask].min(), energy_max[mask].max()])
 
+    @property
+    def energy_range(self):
+        """Energy range defined by the full mask (mask_safe and mask_fit)."""
+        return self._energy_range(use_mask_fit=True)
+    
+    @property
+    def safe_energy_range(self):
+        """Energy range defined by the full mask (mask_safe and mask_fit)."""
```
AtreyeeS, line 367: The doc string should read only `mask_safe`, no?

AtreyeeS (approved):
Thanks a lot @registerrier ! This makes the concept clear. 
I think we should expose the `mask_fit` a bit more in our tutorials, and advertise this notebook more.

Perhaps it is better to have `safe_energy_range`, `fit_energy_range` and`energy_range`(which is the intersection of the previous two) explicitly?
What is being done in the new `plot` method introduced by @fabiopintore ? We should follow a consistent nomenclature...

adonath:
Thanks @registerrier, I think content wise the addition is very nice, however I think the order could maybe be improved. I think the tutorial should start with the basics of maps as masks, incl. creation and arithmetics, move on to more advance things like erosion and dilation and finally explain the interaction with the higher level `MapDataset` structure. From a pedagogical point of view, I think it's non ideal to start with creating `MapDataset`, when the tutorial is about masks...

On 2nd thought I actually prefer  the following: I think  we are still missing a `dataset.ipynb` API tutorial, which explains the basics of the `Dataset` classes (see https://github.com/gammapy/gammapy/issues/3143). What about adding the `.mask_fit` etc. content as the  first content to the a new API tutorial explaining datasets?

registerrier:
> Thanks @registerrier, this is really helpful!
> 
> A few preliminary comments on the `energy_range` definition:
> 
> * Maybe we stick with the mask nomenclature like: `energy_range`, `energy_range_safe`, `energy_range_fit`
> * To avoid code duplication one could introduce `Map.mask_energy_range()`

OK. I am not sure whether it is a good idea to allow an `energy_range_fit`. This could suggest that this is the energy range that will be used for the fit. I suppose this is similar to the mask naming scheme, so it is probably fine though.

I am not found of adding methods on `Maps` that only relate to some specific types of maps (here boolean with an energy axis). Since these methods are exposed to the user, it can be quite confusing on what they can do. I would rather have it a function possibly in a file such as `gammapy/datasets/utils.py`.

adonath:
> OK. I am not sure whether it is a good idea to allow an energy_range_fit. This could suggest that this is the energy range that will be used for the fit. I suppose this is similar to the mask naming scheme, so it is probably fine though.

A agree the choice of name is non-ideal for the man as well, but I think the consistency is more important.

> I am not found of adding methods on Maps that only relate to some specific types of maps (here boolean with an energy axis). Since these methods are exposed to the user, it can be quite confusing on what they can do. I would rather have it a function possibly in a file such as gammapy/datasets/utils.py.

As you prefer. In the helper function one could maybe even return a complete map like `energy_range_maps(mask) > (e_min_map, e_max_map)`, which returns the energy range per pixel. Could be interesting to visualise for datasets, where the threshold changes with position.

registerrier:
After today's meeting we've decided to merge this now and adapt once the dataset API notebook is written.

Issue #3064, opened 2020-10-13
Application of mask_safe in MapDataset and SpectrumDataset does not handle nan values well


lmohrmann:
**Gammapy version**
0.17

**Bug description**
The `mask_safe` is applied in the `MapDataset` and `SpectrumDataset` by multiplying the different maps (counts, background, etc.) with the `mask_safe`. This method fails when, for instance, there are `nan` values present in the background model map (which might occur when the background model is interpolated onto the used geometry). This is because `nan * False = nan`, which results in the corresponding pixels not being properly masked.

**Expected behavior**
All pixels outside the `mask_safe` should be properly masked. This could be achieved by using a scheme like `dataset.background_model.map[~dataset.mask_safe.data] = 0` instead of the multiplication.

adonath:
@lmohrmann I checked once again what the status here is and to apply the mask we consistently use `Map.stack( weights=)`. The `weights` argument however is meant to be more generic and not strictly for boolean types. So I think we cannot apply the proposed solution and have to fix replace `NaN` values beforehand instead. A method for this has been added in #3219

adonath:
However once could check the dtype here: https://github.com/gammapy/gammapy/blob/master/gammapy/maps/wcsnd.py#L782

QRemy:
PR #3416 adds a method "bkg-clip" to the SafeMaskMaker in order create a mask from aberrant values in background maps (not finite and eventually very large values). It also offers the possibility to clip the aberrant values to zero in the background map. This should solve the issue.

adonath:
Solved by #3419.

PR #3416, opened 2021-07-07
Mask invalid background values in SafeMaskMaker

QRemy:
This PR introduces a bkg-clip method to SafeMaskMaker that adds a mask from the aberrant values in background maps

By default only non-finite values are masked.
If`n_95percentile` option  is not None then the values larger than 
`n_95percentile` times the 95% percentile value will also be masked.

### DIFF
--- a/gammapy/makers/safe.py
+++ b/gammapy/makers/safe.py
@@ -59,7 +59,6 @@ class SafeMaskMaker(Maker):
         position=None,
         fixed_offset=None,
         offset_max="3 deg",
-        n_95percentile=None,
     ):
         methods = set(methods)
 
@@ -73,7 +72,6 @@ class SafeMaskMaker(Maker):
         self.position = position
         self.fixed_offset = fixed_offset
         self.offset_max = Angle(offset_max)
-        self.n_95percentile = n_95percentile
         if self.position and self.fixed_offset:
             raise ValueError(
                 "`position` and `fixed_offset` attributes are mutually exclusive"
@@ -261,7 +259,7 @@ class SafeMaskMaker(Maker):
         """
 
         bkg = dataset.background.data
-        mask = np.isfinite(bkg) | (bkg == 0.0)
+        mask = np.isfinite(bkg) & (bkg > 0.0)
         return mask
 
     def run(self, dataset, observation=None):
--- a/gammapy/makers/safe.py
+++ b/gammapy/makers/safe.py
@@ -45,7 +45,7 @@ class SafeMaskMaker(Maker):
         `n_95percentile` times the 95% percentile value.
         Default is None then only non-finite values are masked.
     force_clipping : bool
-       If True the aberrant value found by `bkg-clip` method will be set to zero.
+       If True (default), the aberrant value found by `bkg-clip` method will be set to zero.
     """
 
     tag = "SafeMaskMaker"
@@ -67,7 +67,7 @@ class SafeMaskMaker(Maker):
         fixed_offset=None,
         offset_max="3 deg",
         n_95percentile=None,
-        force_clipping=False,
+        force_clipping=True,
     ):
         methods = set(methods)
 
--- a/gammapy/makers/safe.py
+++ b/gammapy/makers/safe.py
@@ -40,10 +40,6 @@ class SafeMaskMaker(Maker):
         the `aeff_percent` or `bias_percent` are computed.
     offset_max : str or `~astropy.units.Quantity`
         Maximum offset cut.
-    n_95percentile : float
-        If not None the `bkg-clip` method will mask the values larger than 
-        `n_95percentile` times the 95% percentile value.
-        Default is None then only non-finite values are masked.
     """
 
     tag = "SafeMaskMaker"
@@ -53,7 +49,6 @@ class SafeMaskMaker(Maker):
         "edisp-bias",
         "offset-max",
         "bkg-peak",
-        "bkg-clip",
     }
 
     def __init__(
@@ -251,13 +246,8 @@ class SafeMaskMaker(Maker):
         energy_min = energy_axis.pix_to_coord(idx)
         return geom.energy_mask(energy_min=energy_min)
 
-    def make_mask_bkg_clip(self, dataset):
-        """Mask aberrant values in background maps
-
-        By default only non-finite values and zeros are masked.
-        If`n_95percentile` is not None then the values larger than 
-        `n_95percentile` times the 95% percentile value will also be masked.
-       
+    def make_mask_bkg_invalid(self, dataset):
+        """Mask non-finite values and zeros values in background maps
  
         Parameters
         ----------
@@ -270,17 +260,8 @@ class SafeMaskMaker(Maker):
             Safe data range mask.
         """
 
-        bkg = dataset.background.data.copy()
-        bkg[bkg == 0.0] = np.nan
-        mask = np.isfinite(bkg)
-        if self.n_95percentile is not None:
-            thr = self.n_95percentile * np.nanpercentile(
-                bkg, 95, axis=(1, 2), keepdims=True
-            )
-            mask = np.abs(bkg) < thr
-        if np.any(~mask):
-            bad_values = np.unique(bkg[~mask & np.isfinite(bkg)])
-            log.warning(f"Invalid values found in background masked \n {bad_values}")
+        bkg = dataset.background.data
+        mask = np.isfinite(bkg) | (bkg == 0.0)
         return mask
 
     def run(self, dataset, observation=None):
@@ -300,9 +281,9 @@ class SafeMaskMaker(Maker):
         """
         mask_safe = np.ones(dataset._geom.data_shape, dtype=bool)
 
-        if "bkg-clip" in self.methods:
+        if dataset.background is not None and np.any(dataset.background.data):
             # apply it first so only clipped values are removed for "bkg-peak"
-            mask_safe &= self.make_mask_bkg_clip(dataset)
+            mask_safe &= self.make_mask_bkg_invalid(dataset)
             dataset.mask_safe = Map.from_geom(dataset._geom, data=mask_safe)
 
         if "offset-max" in self.methods:
--- a/gammapy/makers/safe.py
+++ b/gammapy/makers/safe.py
@@ -44,6 +44,8 @@ class SafeMaskMaker(Maker):
         If not None the `bkg-clip` method will mask the values larger than 
         `n_95percentile` times the 95% percentile value.
         Default is None then only non-finite values are masked.
+    force_clipping : bool
+       If True the aberrant value found by `bkg-clip` method will be set to zero.
     """
 
     tag = "SafeMaskMaker"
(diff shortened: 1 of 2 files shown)


gammapy/makers/safe.py
```diff
@@ -247,6 +255,41 @@ def make_mask_energy_bkg_peak(dataset):
...
+        Returns
+        -------
+        mask_safe : `~numpy.ndarray`
+            Safe data range mask.
+        """
+
+        bkg = dataset.background.data
+        mask = np.isfinite(bkg)
+        if self.n_95percentile is not None:
+            for k in range(bkg.shape[0]):
```
adonath, line 280: Please avoid the loop and work with `np.nanpercentile(axis=(1, 2))` instead...

gammapy/makers/safe.py
```diff
@@ -247,6 +255,41 @@ def make_mask_energy_bkg_peak(dataset):
...
+                data = bkg[k, :, :]
+                thr = self.n_95percentile * np.nanpercentile(data[data > 0], 95)
+                mask[k, :, :] = np.abs(bkg[k, :, :]) < thr
+        if self.force_clipping:
+            dataset.background.data[~mask] = 0.0
+        if np.any(~mask):
+            bad_values = np.unique(bkg[~mask])
+            log.warning(f"Invalid values found in background masked \n {bad_values}")
+            # TODO: maybe we should store the bad values and their coord somewhere,
```
adonath, line 289: As users explicitly request to fix the bad values, this might not be needed for now. If they are interested in the bad values, they could just keep them in the background map and check afterwards...

gammapy/makers/safe.py
```diff
@@ -40,6 +40,12 @@ class SafeMaskMaker(Maker):
         the `aeff_percent` or `bias_percent` are computed.
     offset_max : str or `~astropy.units.Quantity`
         Maximum offset cut.
+    n_95percentile : float
```
adonath, line 43: What is the advantage of hard-coding to `95%` and then scale up? Maybe we just take the value of percentile directly? And maybe mention `bkg` in the name, e.g like `bkg_clip_percentile`

gammapy/makers/safe.py
```diff
@@ -40,6 +40,12 @@ class SafeMaskMaker(Maker):
...
+        If not None the `bkg-clip` method will mask the values larger than 
+        `n_95percentile` times the 95% percentile value.
+        Default is None then only non-finite values are masked.
+    force_clipping : bool
```
adonath, line 47: I guess this is needed to solve #3064? I can see the need for this, but I'd prefer to solve it in a more general way. Such that e.g. `dataset.to_masked()` handles the `NaN` values correctly. Maybe remove from here and we add an option like `dataset.to_masked(nan_to_num=0)`, which passed down to `Map.stack(, weights=, nan_to_num=0)`?

adonath (changes requested):
Thanks @QRemy, I have left some inline comments...

gammapy/makers/safe.py
QRemy: ok I will remove this comment

gammapy/makers/safe.py
QRemy: It very hard to find the exact percentile that corresponds to the good-to-bad values transition, but we can reasonably assume that bad values are within the 5 last percentile and much larger than the good ones so I opted for that solution.

gammapy/makers/safe.py
QRemy: in a separate PR ?

gammapy/makers/safe.py
```diff
@@ -274,20 +270,15 @@ def make_mask_bkg_clip(self, dataset):
...
         mask = np.isfinite(bkg)
         if self.n_95percentile is not None:
-            for k in range(bkg.shape[0]):
-                data = bkg[k, :, :]
-                thr = self.n_95percentile * np.nanpercentile(data[data > 0], 95)
-                mask[k, :, :] = np.abs(bkg[k, :, :]) < thr
-        if self.force_clipping:
-            dataset.background.data[~mask] = 0.0
+            thr = self.n_95percentile * np.nanpercentile(bkg, 95, axis=(1,2))
+            mask = np.abs(bkg) < thr[:, None, None]
```
adonath, line 278: One minor comment here: you can use `np.nanpercentile(..., keepdims=True)` to avoid the `[:, None, None]`

gammapy/makers/safe.py
adonath: Yes, please!

adonath:
@QRemy Sorry I should have investigated earlier, but I just started to check where the large values actually come from and indeed they come from extrapolation of the background IRF.

This is the current default of extrapolation:
Screenshot 2021-07-09 at 12 13 22

When using `np.nan` as fill value:
Screenshot 2021-07-09 at 12 13 57

So one can see that the background is only defined up to 2.5 deg in for lon/lat. Extrapolation of IRFs will always be unstable and we should definitely not do it by default.  My strong recommendation would be that we change the default behaviour here:
- Don't extrapolate the IRFs, but fill with `np.nan` values instead
- By default add `np.nan` values to the `mask_safe` in the `SafeMaskMaker`, so that they are ignored during fitting and stacking.
- There might be rare use cases where people want to extrapolate IRFs. I think a reasonable solution could be to add `IRF.pad()`, where users can pad an IRF prior to evaluation. 

Does this sounds reasonable to you? If so I'd prefer to not add the `clip_bkg` option and close the PR, otherwise if you and others think it might be still useful to have it we could merge it as well. What do you think  @registerrier @AtreyeeS?

QRemy:
I adapted the PR to fill the IRF maps with nan instead of extrapolating and then mask invalid bkg by default in SafeMaskMaker (so the bkg-clip option is no longer visible and percentile option removed).

gammapy/makers/safe.py
```diff
@@ -59,6 +59,7 @@ def __init__(
         position=None,
         fixed_offset=None,
         offset_max="3 deg",
+        n_95percentile=None,
```
adonath, line 62: Remove the option here...

gammapy/makers/safe.py
```diff
@@ -247,6 +246,24 @@ def make_mask_energy_bkg_peak(dataset):
...
+            Dataset to compute mask for.
+
+        Returns
+        -------
+        mask_safe : `~numpy.ndarray`
+            Safe data range mask.
+        """
+
+        bkg = dataset.background.data
+        mask = np.isfinite(bkg) | (bkg == 0.0)
```
adonath, line 264: I this correct? I think we should require `np.isfinite() & bkg > 0`?

adonath (changes requested):
Thanks @QRemy, I have left tow final comments...

gammapy/makers/safe.py
```diff
@@ -247,6 +244,24 @@ def make_mask_energy_bkg_peak(dataset):
...
+            Dataset to compute mask for.
+
+        Returns
+        -------
+        mask_safe : `~numpy.ndarray`
+            Safe data range mask.
+        """
+
+        bkg = dataset.background.data
+        mask = np.isfinite(bkg) & (bkg > 0.0)
```
adonath, line 262: I just noticed there are test fails now for on-off datasets: it seems there are case where `alpha * n_off` is actually zero, because `n_off` is zero. @registerrier should those bins with `n_off` be included by default in stat computation? I remember there is a bias for this situation...

QRemy: I would say that we have  a poor background estimation if `n_off` is zero that would lead to over-estimate the overall signal-to-background. For now I have fixed the tests assuming that masking zeros is right.

adonath: Sounds reasonable to me, maybe add an explanation in the docstring on the behaviour for on-off datasets. And maybe let's wait for @registerrier to finally comment...

registerrier: Technically, including the empty OFF bins can introduce a bias but removing them will introduce one as well.
Also if you use real ON-OFF 2D datasets you will end up having many empty bins.

I would not set such a mask by default on ON-OFF datasets.

adonath: Thanks @registerrier, so it seems there is no obviously correct way to treat this. In this case we maybe just go with the "established" behaviour?  I guess traditionally people have included the bins and maybe we just do the same to not introduce unexpected behaviour for un-obvious reasons. Are you OK @QRemy to revert the change?

registerrier: I guess the issue here depends on what will be done later. Typically for an ON-OFF dataset, background estimation is carried out later. So this would not apply. 
I think we need to document that this is not to be applied for ON-OFF analysis.

adonath: Yes, I think we have never defined (but maybe should...) on the order of background and safe data range estimation. I'm not even sure if there is a generally preferable order. Ad hoc I'd also say `SafeMaskMaker -> BackgroundMaker`...but there might be cases where the result of the background estimate affects the chosen data range...

QRemy: I will revert the changes and apply the bkg > 0 condition only if the dataset is not a ON-OFF.

adonath: Thanks! @registerrier do we already have a note in the docs on the `wstat` bias? I guess we would need a reference which doesn't exist yet?

adonath (approved):
Thanks @QRemy, I added one commit with a bit of cleanup, otherwise looks good. Will merge now...

gammapy/makers/safe.py
```diff
@@ -281,10 +283,9 @@ def run(self, dataset, observation=None):
         """
         mask_safe = np.ones(dataset._geom.data_shape, dtype=bool)
 
-        if dataset.background is not None and np.any(dataset.background.data):
```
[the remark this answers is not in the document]
adonath: I see. I was wondering about the behaviour and checked there was no test fail. I agree in general there might be the case where users set the background all zero e.g. to remove the contribution from the background model.  I guess it depends a bit on the context, for IACT data working with an all zero background is probably a bad idea, for Fermi dat aim might just be fine.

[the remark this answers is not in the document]
AtreyeeS: I suppose creating an empty dataset leads to a background with zero filled maps, no?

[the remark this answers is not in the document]
QRemy: Running the MapDatasetMaker with "background" option disabled, will return the reference dataset with zero-filled background and then if you apply the MaskSafeMaker before the BackgroundMaker the mask safe will be all False so I think we need this `np.any(dataset.background.data)` for that case.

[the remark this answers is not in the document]
adonath: The actual `background` map is filled by the `MapDatasetMaker`, which is usually run as the first step. So the empty background map can only occur if the background map is omitted  in the `MapDatasetMaker`? The `FoVBackgroundMakler` does not change that. All the other background makers create on-off datasets, so is it really an issue?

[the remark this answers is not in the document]
QRemy: ok then should fine

Issue #3414, opened 2021-07-07
HpxMap read wrong coord and cannot find the right peak


jlzhang001:
**Gammapy version**
System:

        python_executable      : /public/home/jlzhang/anaconda3/envs/pipthreeML/bin/python 
        python_version         : 3.7.10     
        machine                : x86_64     
        system                 : Linux      

Gammapy package:

        version                : 0.18.2     
        path                   : /public/home/jlzhang/anaconda3/envs/pipthreeML/lib/python3.7/site-packages/gammapy 

Other packages:

        numpy                  : 1.20.3     
        scipy                  : 1.6.3      
        astropy                : 4.2.1      
        regions                : 0.4        
        click                  : 8.0.1      
        yaml                   : 5.4.1      
        IPython                : 7.24.0     
        jupyterlab             : not installed 
        matplotlib             : 3.4.2      
        pandas                 : 1.2.4      
        healpy                 : not installed 
        iminuit                : 2.6.1      
        sherpa                 : not installed 
        naima                  : not installed 
        emcee                  : 3.0.2      
        corner                 : 2.2.1      
        parfive                : 1.3.0      

Gammapy environment variables:

        GAMMAPY_DATA           : /public/home/jlzhang/hawc/gammapy-tutorials/datasets 

**Bug description**
A short description what the bug is (usually 1-2 sentences)
I have a hpx all sky map, HpxMap read wrong coord and cann't find the right peak, only find the pixel around one source

**Expected behavior**

read HpxMap with right coord and find all the peaks in the sky map

**To Reproduce**
Steps to reproduce the behavior.

```python3
import healpy as hp
import numpy as np
from gammapy.maps import HpxGeom, HpxNDMap, HpxMap
from gammapy.estimators.utils import find_peaks
import matplotlib.pyplot as plt
sig = hp.read_map('skymap.fits') #which is a healpix all sky map, NSIDE=1024

m = HpxMap.create(nside=1024, nest=False,frame="icrs")
m.data=sig
m_wcs=m.to_wcs()
m_wcs.plot(cmap='nipy_spectral_r')
plt.show()
sources = find_peaks(m_wcs, threshold=5, min_distance="1.00 deg")
print(sources)
```

output:
```
value   x    y      ra      dec   
                   deg      deg   
------ ---- ---- -------- --------
6.3466 8234 9068 83.67101 22.01225
6.3466 8233 9068 83.68506 22.01175
6.3466 8231 9064 83.68985 21.96360
[...]

PR #3417, opened 2021-07-08
Add a deduplication method to SourceCatalogue

QRemy:
This adds a deduplication method to SourceCatalogue that return a new catalogue merging entries within a given radius.
Positions of the merged objects are set to the coordinates mean.
Provide a solution for  #3414

### DIFF
--- a/gammapy/catalog/core.py
+++ b/gammapy/catalog/core.py
@@ -6,6 +6,8 @@ import numpy as np
 from copy import deepcopy
 from astropy.coordinates import SkyCoord
 from astropy.utils import lazyproperty
+from astropy.table import Table
+import astropy.units as u
 from gammapy.utils.table import table_from_row_data, table_row_to_dict
 from gammapy.modeling.models import Models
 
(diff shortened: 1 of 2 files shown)


QRemy:
For now I convert this to a draft, we will come back to improvements of the catalogue module and its relation to object detection and filtering methods after 1.0

adonath:
Just for reference here is a more efficient similar implementation, that I found on my hard disk:
```python3
from gammapy.catalog import SourceCatalog3FGL
from scipy.sparse.csgraph import connected_components
from scipy import ndimage
import numpy as np

catalog = SourceCatalog3FGL()

pos = catalog.positions
separation = pos.separation(pos[:, None])
mask = (separation < 5 * u.deg)
n_groups, labels = connected_components(mask)

lon_mean = ndimage.mean(pos.data.lon, labels=labels, index=range(n_groups))
lat_mean = ndimage.mean(pos.data.lat, labels=labels, index=range(n_groups))
```

QRemy:
Thanks @adonath, I will try with this. I think an extra loop is still needed to make it recursive.

QRemy:
Closing this for now I will probably implement first what commented before https://github.com/gammapy/gammapy/pull/3417#issuecomment-876304491 :

> * introduce a dedicated SeedCatalogue object with some predefined columns names, move the deduplicate method there and make `findpeaks` and any future detection methods return this object.

PR #3418, opened 2021-07-08
Refactor FluxPoints to rely on maps internally

adonath:
**Description**

This pull request refactors `FluxPoints` to rely on the `RegionNDMap` data structure instead of a `Table`. This allows to share all functionality with `FluxMaps` and to extend the flux points for additional axes in future. Here is a summary of the changes:

- Remove `FluxPoints.table` and `FluxPoints.table_formatted`, both are replaced by the `.to_table(sed_type, formatted=True)`
- Add a `meta` attribute to `FluxPoints`, which allows to hold data such as `n_sigma`, `n_sigma_ul` and `ts_threshold_ul`
- Add `RegionNDMap.from_table()` which supports `format = "gadf-sed"`
- Adapt a bit `FluxPoints.__str__`, it now gives the following output:
```
FluxPoints
----------

  geom           : RegionGeom
  axes           : ['lon', 'lat', 'energy']
  shape          : (1, 1, 24)
  quantities     : ['norm', 'norm_err', 'norm_ul', 'ts', 'stat', 'stat_scan', 'stat_null']
  ref. model     : template
  n_sigma        : 
  n_sigma_ul     : 
  ts_theshold_ul : 4
```

In general I think this goes exactly in the right direction. However there are some open points here:
- Further reduce code duplication with `FluxMaps` (can be resolved in a follow up PR)
- Interacting with the data becomes less nice, because the data is stored as 3D array. However users can always rely on `.from_table()` and `.to_table()`. But maybe we can find a way to reduce the dimensionality of the data of the `RegionNDMap`
 again...

**Dear reviewer**
This PR is hard to review. Maybe checkout the branch and start with
```python3

fp = FluxPoints.read("$GAMMAPY_DATA/tests/spectrum/flux_points/binlike.fits")
print(fp)
```


gammapy/estimators/core.py
```diff
@@ -226,110 +253,151 @@ def _validate_data(data, sed_type, check_scan=False):
...
 
-    @property
-    def energy_axis(self):
-        """Energy axis (`MapAxis`)"""
-        return self._energy_axis
+    # TODO: add support for scan
+    def _check_quantity(self, quantity):
+        if quantity not in self.available_quantities:
+            raise AttributeError(
+                f"Cannot compute required flux quantity. {quantity} "
```
registerrier, line 260: Note that when no `counts` or no `ts` are available the message about flux quantity is strange.

gammapy/estimators/flux_point.py
```diff
@@ -302,7 +227,7 @@ def _convert_flux_columns(table, reference_model, sed_type):
         col_ref = table[sed_type]
         factor = fluxes[f"ref_{sed_type}"].to(col_ref.unit)
 
-        data = Table(fluxes)
+        data = Table(fluxes, meta=table.meta)
```
registerrier, line 230: I think that this implementation omits additional columns such as `ts` `counts` etc.

registerrier (commented):
Thanks @adonath .

 For now, I only left few comments. I am still testing the new approach.
Trying to read flux point files produced with v0.18.2 with the new class (reading the table first, or directly with `FluxPoints.read`) it seems that the `FluxPoints.from_table` is overlooking some columns such as ts and others.

gammapy/estimators/flux_point.py
adonath: Thanks, I'll check. It seems there are no unit tests yet that cover this case...

gammapy/estimators/core.py
adonath: Agreed, done.

gammapy/estimators/flux_point.py
adonath: The information is actual there, if one specifies the `sed_type="likelihood"` specifically:
```python3
fp = FluxPoints.read("flux-points.fits", sed_type="likelihood")
print(fp.counts)
```
Part of the problem is, that we wrote mixed sed types in v0.18.2 i.e. tables that contain all quantities like `norm` and `dnde` etc. But I guess we still want to keep quantities like `counts` for the other sed-types as well. So I'll check why those are not read from the table...

adonath: Fixed

adonath:
This will still need a follow PR to further reduce the code duplication between `FluxPoints` and `FluxMaps`. I think the reasonable solution here is to make `FluxMaps` the base class and make `FluxPoints` inherit from it...

adonath:
@registerrier I'll go ahead and merge this now. There will be many opportunities to further test and comment on the new `FluxPoints` data structure.

PR #3419, opened 2021-07-08
Add fill_value option on stack methods to replace invalid values

QRemy:
Add a fill_value option on .to_masked() and .stack() methods to replace non-finite values as suggested in :
https://github.com/gammapy/gammapy/pull/3416#discussion_r665947289

### DIFF
--- a/gammapy/maps/hpxnd.py
+++ b/gammapy/maps/hpxnd.py
@@ -359,7 +359,7 @@ class HpxNDMap(HpxMap):
             geom=geom, data=data, unit=self.unit, meta=self.meta
         )
 
-    def stack(self, other, weights=None):
+    def stack(self, other, weights=None, fill_value=0):
         """Stack cutout into map.
 
         Parameters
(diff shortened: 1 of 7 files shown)


gammapy/maps/hpxnd.py
```diff
@@ -382,8 +384,9 @@ def stack(self, other, weights=None):
                 "Can only stack equivalent maps or cutout of the same map."
             )
 
-        data = other.quantity.to_value(self.unit)
-
+        data = other.quantity.to_value(self.unit).copy()
```
adonath, line 387: Could avoid the copy here, if `nan_to_num = False`...

adonath (approved):
Thanks @QRemy, I have left one minor inline comment otherwise looks fine!

gammapy/maps/hpxnd.py
QRemy: Without the copy I have two tests failing with a message about "data is read only"

adonath: I meant moving the `.copy()` into the `if` block...

adonath (approved):

PR #3420, opened 2021-07-09
Disable IRF extrapolation

QRemy:
Following the discussion in https://github.com/gammapy/gammapy/pull/3416#issuecomment-877083206
this PR disable the extrapolation of IRF maps. This breaks a lot of tests, should it be applied to background maps only ?

### DIFF
--- a/gammapy/irf/core.py
+++ b/gammapy/irf/core.py
@@ -33,7 +33,7 @@ class IRF:
         Meta data
     """
 
-    default_interp_kwargs = dict(bounds_error=False, fill_value=np.nan,)
+    default_interp_kwargs = dict(bounds_error=False, fill_value=None,)
 
     def __init__(self, axes, data=0, unit="", meta=None):
         axes = MapAxes(axes)
--- a/gammapy/irf/core.py
+++ b/gammapy/irf/core.py
@@ -203,28 +203,12 @@ class IRF:
                 coords_default[key] = u.Quantity(coord, copy=False)
         data = self._interpolate(coords_default.values(), method=method)
 
-        eps = 1e-5
-        for k, ax in enumerate(self.axes):
-            coords = coords_default[ax.name]
-            unit = coords.unit
-            if not coords.shape:
-                continue
-            coords = coords.squeeze()
-            if len(np.atleast_1d(coords)) != 1 and coords.shape != data.shape:
-                ind = []
-                for s in coords.shape:
-                    ind.extend(np.where(np.array(data.shape)==s)[0])
-                axis = tuple(
-                    [idx for idx in range(len(data.shape)) if idx not in ind]
-                )  # simpler solution ?
-                coords = np.expand_dims(coords, axis=axis)
-                coords = np.broadcast_to(coords, data.shape) * unit
-            axmin = ax.edges.min()
-            axmax = ax.edges.max()
-            mask = (coords < (1 - eps * np.sign(axmin)) * axmin) | (
-                coords > (1 + eps * np.sign(axmax)) * axmax
-            )
-            data[mask] = np.nan
+        idxs = self.axes.coord_to_idx(coords_default, clip=False)
+        invalid = np.broadcast_arrays(*[idx == -1 for idx in idxs])
+        mask = np.any(invalid, axis=0)
+        if not data.shape:
+            mask = mask.squeeze()
+        data[mask] = np.nan
         return data
 
     def integrate_log_log(self, axis_name, **kwargs):
--- a/gammapy/irf/core.py
+++ b/gammapy/irf/core.py
@@ -33,7 +33,7 @@ class IRF:
         Meta data
     """
 
-    default_interp_kwargs = dict(bounds_error=False, fill_value=None,)
+    default_interp_kwargs = dict(bounds_error=False, fill_value=np.nan,)
 
     def __init__(self, axes, data=0, unit="", meta=None):
         axes = MapAxes(axes)
--- a/gammapy/irf/core.py
+++ b/gammapy/irf/core.py
@@ -211,7 +211,7 @@ class IRF:
             if not data.shape:
                 mask = mask.squeeze()
             data[mask] = self.fill_value
-            data[np.isnan(data)] = self.fill_value
+            data[~np.isfinite(data)] = self.fill_value
         return data
 
     def integrate_log_log(self, axis_name, **kwargs):
--- a/gammapy/irf/core.py
+++ b/gammapy/irf/core.py
@@ -33,16 +33,18 @@ class IRF:
         Meta data
     """
 
-    default_interp_kwargs = dict(bounds_error=False, fill_value=None,)
-    fill_value = 0.
+    default_interp_kwargs = dict(bounds_error=False, fill_value=0.,)
 
-    def __init__(self, axes, data=0, unit="", meta=None):
+    def __init__(self, axes, data=0, unit="", meta=None, interp_kwargs=None):
         axes = MapAxes(axes)
         axes.assert_names(self.required_axes)
         self._axes = axes
         self.data = data
         self.unit = unit
         self.meta = meta or {}
+        if interp_kwargs is None:
+            interp_kwargs = self.default_interp_kwargs.copy()
+        self.interp_kwargs = interp_kwargs
 
     @property
     @abc.abstractmethod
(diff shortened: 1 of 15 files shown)


adonath:
Thanks a lot @QRemy, this is an important change! I'm surprised to see that we rely on extrapolation of IRFs in so many places. I think we should not extrapolate any IRFs at all. So let's run the CI here and then we review the cases where it fails together, I'll include @registerrier  and @AtreyeeS in the discussion.

adonath:
Meanwhile I'll implement `IRF.pad()` to be able to extend the validity range of IRFs for well justified cases...

registerrier:
I agree that relying on irf extrapolation sounds like a bad idea a priori. It would be interesting to see where this causes problem.

adonath:
Let's wait for #3416 and #3419 to be merged and see which fails remain here. The ones related to stacking should be gone after...

adonath:
@QRemy To make it easier to understand the test fails we could go 1 by 1, so disable the extrapolation for the background first, then effective area, then PSF etc.

QRemy:
I fixed the sampling in case of non-finite value in npred (3 tests failure with "ValueError: lam value too large")

adonath:
Looking at the test fails we get I think they can be divided I a few categories:
- Actual bugs: in some tests we define geoms and the pointing is actually outside
- `sum()` and `.mean()`: sometimes we use large geoms, where part of the data is extrapolated and check the assert with `.sum()` and `.mean()`. Those test could be fix by relying on `np.nansum()` and `np.nanmean()` instead.
- `mask` handling: some tests rely on the `mask_safe` being properly defined. Should be fixed after #3416
- For some tests it's a bit more tricky: the axis is defined with edges, while for the interpolation we just rely on the centers. If the value is in between the center and the edge of the last bin, it fails. If we consider the extrapolation within a bin to be OK, we might need a workaround...

gammapy/irf/core.py
```diff
@@ -201,8 +201,30 @@ def evaluate(self, method=None, **kwargs):
...
+                )  # simpler solution ?
+                coords = np.expand_dims(coords, axis=axis)
+                coords = np.broadcast_to(coords, data.shape) * unit
+            axmin = ax.edges.min()
+            axmax = ax.edges.max()
+            mask = (coords < (1 - eps * np.sign(axmin)) * axmin) | (
+                coords > (1 + eps * np.sign(axmax)) * axmax
+            )
+            data[mask] = np.nan
+        return data
```
QRemy, line 227: I tried to add this code to allow the extrapolation within the default range, it fix few tests but I still have problems with array shapes in many cases. @adonath a suggestion to make this simpler ?

adonath: I think something along the lines of:
```python3
from gammapy.maps import MapAxis, MapAxes
import numpy as np

energy = MapAxis.from_energy_bounds("1 TeV", "10 TeV", nbin=10)
offset = MapAxis.from_bounds(0, 5, unit="deg", nbin=5, name="offset")

axes = MapAxes([energy, offset])
coords = {
    "energy": np.geomspace(0.3, 30, 20)[:, None] * u.TeV,
    "offset": np.linspace(0, 6, 10) * u.deg
}

idxs = axes.coord_to_idx(
    coords, clip=False
)

invalid = np.broadcast_arrays(*[idx == -1 for idx in idxs])
mask = np.any(invalid, axis=0)
```
Should work. I think the correct clipping of values should be handled in `MapAxis.coord_to_idx()` if to doesn't work as expected we should probably adapt it there.

adonath: I haven't looked into your solution in detail, but `._interpolate` does the broadcasting of arrays internally as well, so `data` and `mask` should always end up with the same shape.

QRemy: Thanks this is much simple. The evaluation at the edge value sometimes return nan,  I think we need have to add some tolerance, can we handle this doing the clipping in MapAxis.coord_to_idx() ? Otherwise we could just adapt the tests.

adonath: Do you have a minimal test case for this, or which test?

gammapy/irf/core.py
```diff
@@ -201,8 +202,20 @@ def evaluate(self, method=None, **kwargs):
             coord = kwargs.get(key, value)
             if coord is not None:
                 coords_default[key] = u.Quantity(coord, copy=False)
+        data = self._interpolate(coords_default.values(), method=method)
 
-        return self._interpolate(coords_default.values(), method=method)
+        if self.fill_value != self.default_interp_kwargs["fill_value"]:
```
adonath, line 207: Do we need the extra `.fill_value` definition here? Why can't we just rely on `default_interp_kwargs["fill_value"]`?

gammapy/irf/core.py
```diff
@@ -201,8 +202,20 @@ def evaluate(self, method=None, **kwargs):
...
+            idxs = self.axes.coord_to_idx(coords_default, clip=False)
+            invalid = np.broadcast_arrays(*[idx == -1 for idx in idxs])
+            mask = self._mask_out_bounds(invalid)
+            if not data.shape:
+                mask = mask.squeeze()
+            data[mask] = self.fill_value
+            data[~np.isfinite(data)] = self.fill_value
```
adonath, line 214: Hm, within the defined data rang the IRFs should be valid I guess, why this additional masking?

gammapy/irf/edisp/core.py
```diff
@@ -48,9 +48,17 @@ class EnergyDispersion2D(IRF):
 
     tag = "edisp_2d"
     required_axes = ["energy_true", "migra", "offset"]
+    fill_value = 0.0
+
+    def _mask_out_bounds(self, invalid):
```
adonath, line 53: Again why is this necessary? Doesn't it work to just declare `fill_value=0` for this case?

adonath (commented):
Thanks @QRemy, I left a few questions to clarify. I think it's important that we end up with a "minimal invasive" solution here. I could possibly do one follow up PR moving the "within bin extrapolation" to the `ScaledRegularGridInterpolator` for now it's fine to keep it in `IRF.evaluate`

gammapy/irf/core.py
QRemy: yes as default_interp_kwargs["fill_value"] must remain None so the extrapolation is performed.

adonath: I see, maybe we just handle it in `_interpolate` instead , like:
```python3
kwargs = self.default_interpolation_kwargs.copy()
# Allow extrap[olation with in bins
kwargs["fill_value"] = None
return ScaledRegularGridInterpolator(..., **kwargs)
```
And later rely on:
```python3
if self.default_interpolation_kwargs["fill_value"] is not None:
    # replace values outside by fill value
```
The fill values are then just defined on a sub-class level...

gammapy/makers/tests/test_utils.py
```diff
@@ -155,7 +155,9 @@ def bkg_3d_custom(symmetry="constant"):
     energy_axis = MapAxis.from_energy_edges([0.1, 10, 1000] * u.TeV)
     fov_lon_axis = MapAxis.from_edges([-3, -1, 1, 3] * u.deg, name="fov_lon")
     fov_lat_axis = MapAxis.from_edges([-3, -1, 1, 3] * u.deg, name="fov_lat")
-    return Background3D(
+    Background3D_extra = Background3D
+    Background3D_extra.fill_value = None #allow extrapolation for symetry tests
```
adonath, line 159: One last idea: I noticed that we only define `default_interp_kwargs`, but no instance level `interp_kwargs`. So maybe we just add:

```python3

class IRF:
    def __init__(self, ..., interp_kwargs=None):
        if interp_kwargs is None:
            interp_kwargs = self.interp_kwargs_default.copy()
        self.interp_kwargs = interp_kwargs
```

adonath:
Can you check once whether the behaviour is no roughly consistent with what we write here: https://docs.gammapy.org/0.18.2/development/howto.html#interpolation-and-extrapolation

gammapy/irf/edisp/core.py
QRemy: without this addition `TestEDispKernel.test_to_image` and `test_edispkernel_from_1d` fail.  I think the extrapolation is needed if you are in the valid range of at least one of the energy axes.

gammapy/irf/core.py
QRemy: interp_missing_data can also add NaN, I thought that it would be better to keep the NaN there instead of fill_value as it modifies directly the data and then fix it on evaluate.

gammapy/irf/core.py
QRemy: This means we cannot disable completely the extrapolation within bins if necessary but ok.

adonath: I think that's fine for now. It's the default behaviour we want to establish...

adonath:
Thanks @QRemy, looks good to me now! Could finally check, whether the test fail here is related https://github.com/gammapy/gammapy/pull/3420/checks?check_run_id=3088475497#step:7:32?

adonath (approved):
Thanks a lot @QRemy, I'll merge the PR now...

PR #3423, opened 2021-07-12
Implement LabelMapAxis

adonath:
**Description**

This pull request adds a first version of `LabelMapAxis` to allow for handling of labels for map axes. It implements the basic functionality for label to idx conversion and vice versa as well as the basic axis attributes. 

The core of the functionality is to be bale to define an axis like:
```python3
label_axis = LabelMapAxis(labels=["label-1", "label-2", "label-3"], name="label")
```
This allows to include labelled data in a `Map` object. This is useful in the context of e.g. flux points where flux points of multiple sources could be handled with the same object. The same goes for maps, e.g. to include npred maps from multiple datasets.

### DIFF
--- a/gammapy/maps/axes.py
+++ b/gammapy/maps/axes.py
@@ -2712,6 +2712,17 @@ class LabelMapAxis:
         str_ += fmt.format(f"labels", "{0}".format(list(self._labels)))
         return str_.expandtabs(tabsize=2)
 
+    def __eq__(self, other):
+        if not isinstance(other, self.__class__):
+            return NotImplemented
+
+        name_equal = self.name.upper() == other.name.upper()
+        labels_equal = np.all(self.center == other.center)
+        return name_equal & labels_equal
+
+    def __ne__(self, other):
+        return not self.__eq__(other)
+
     # TODO: could create sub-labels here using dashes like "label-1-a", etc.
     def upsample(self, *args, **kwargs):
         """Upsample axis"""
--- a/gammapy/maps/axes.py
+++ b/gammapy/maps/axes.py
@@ -14,7 +14,7 @@ from gammapy.utils.time import time_ref_to_dict, time_ref_from_dict
 from .utils import INVALID_INDEX, edges_from_lo_hi
 
 
-__all__ = ["MapAxes", "MapAxis", "TimeMapAxis"]
+__all__ = ["MapAxes", "MapAxis", "TimeMapAxis", "LabelMapAxis"]
 
 
 def flat_if_equal(array):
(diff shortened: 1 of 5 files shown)


adonath:
This requires a follow-up pPR to reduce code duplication between `MapAxis`, `TimeMapAxis` and `LabelMapAxis` and define the minimal common API probably by a common base class.

gammapy/maps/axes.py
```diff
@@ -2125,3 +2126,177 @@ def from_gti(cls, gti, name="time"):
...
+        """Bin width is unity """
+        return np.ones(self.nbin)
+
+    def __repr__(self):
+        str_ = self.__class__.__name__ + "\n"
+        str_ += "-" * len(self.__class__.__name__) + "\n\n"
+        fmt = "\t{:<10s} : {:<10s}\n"
+        str_ += fmt.format("name", self.name)
+        str_ += fmt.format("nbins", str(self.nbin))
+        str_ += fmt.format("node type", self.node_type)
```
AtreyeeS, line 2276: All `LabelMapAxis` have the same `node_type`. Is it necessary to print?

gammapy/maps/axes.py
```diff
@@ -2125,3 +2126,177 @@ def from_gti(cls, gti, name="time"):
...
+        str_ += fmt.format(f"labels", "{0}".format(list(self._labels)))
+        return str_.expandtabs(tabsize=2)
+
+    # TODO: could create sub-labels here using dashes like "label-1-a", etc.
+    def upsample(self, *args, **kwargs):
+        """Upsample axis"""
+        raise NotImplementedError("Upsampling a LabelMapAxis is not supported")
+
+    # TODO: could merge labels here like "label-1-label2", etc.
```
AtreyeeS, line 2285: For merging labels, I would prefer to call it `.merge_nodes()` of something similar instead of `downsample` - Downsampling seems to indicate it will similarly on all bins, whereas it might be reasonable to merge `A` + `B`, and then `C`+`D`+`E` and just keep `F`.

But this is of course for later. For now, I `NotImplementedError` is reasonable. (Or do we even need to have/expose the `upsample` and `downsample`  options?)

AtreyeeS (dismissed):
Thanks @adonath ! `LabelMapAxis` has the potential to be quite useful. We can merge this in and keep building upon it after 1.0

gammapy/maps/axes.py
adonath: The advantage of exposing it, is we can raise an `NotImplementedError`, otherwise it would fail e.g. with and `AttributeError` if `Map.downsample()` is called and a `LabelMapAxis` is present. 

Note that `.downsample()` is indeed the same for all bins, but I fully agree a more general `.merge_labels()` or `.resample()` method would be useful.

gammapy/maps/axes.py
adonath: I agree, it basically duplicates the information already contained in the name of the class.

adonath:
Thanks @AtreyeeS! I'll go ahead and merge the PR, we will do more testing and use more of the functionality in follow up PRs.

Issue #2267, opened 2019-07-03
Improve reflected regions finder


cdeil:
This issue lists some questions and possible changes and improvements to the reflected regions finder, following up on discussions with @registerrier and @adonath this morning.

This is still one of the most complex pieces of code in Gammapy and I think we should do one more attempt to make it better now, to have it in good shape for v1.0.

[PIG 10](https://docs.gammapy.org/dev/development/pigs/pig-010.html#rotated-regions) and  #2089 are related, and the current code is here:

https://github.com/gammapy/gammapy/blob/fe8ca7d6caac77b8a31efc8bec3b21d09aacf6c1/gammapy/background/reflected.py#L15-L255

The current algorithm is based on a `reference_map` with small pixels that is a local TAN projection centered on the observation pointing position. If the user provides an exclusion mask, that is reprojected onto that `reference_map`, and then for the reflected region finding, the input `exclusion` mask isn't used anymore. I think basically that's fine and all we can achieve, building up spherical sky region contains and overlap functionality and changing to purely spherical computations would require weeks of developments in `astropy-regions`, no-one is working on this, isn't feasible. If we basically stick with this method to have a PixelRegion based algorithm working with  a local TAN reference mask image, I think `ReflectedRegionsFinder` can be improved in several ways.

The current class has conflated `__init__`, `setup` and `run` methods, where part of the setup (making the reference_map) happens in `run`, which is why `run` has to call `setup`. The tests currently make one finder and then use it for multiple jobs, changing out things like which region to use (see [here](https://github.com/gammapy/gammapy/blob/fe8ca7d6caac77b8a31efc8bec3b21d09aacf6c1/gammapy/background/tests/test_reflected.py#L116). I suggest we change the class to do all setup in `__init__` (could call a setup method from `__init__`), and then have `run` be what `find_regions` does now. The only allowed way to use the class would be
```
finder = ReflectedRegionsFinder(...)
reflected_regions = finder.run()
```
or
```
finder = ReflectedRegionsFinder(...)
finder.run()
finder.reflected_regions
finder.other_things_for_debugging
```
which would allow to change `self.reference_map` to `self.exclusion_mask`, and to initialise it by reprojecting the user input exclusion mask onto it and to exclude the pixels in the on region also.

[...]

PR #3424, opened 2021-07-12
Analysis class clean-up

registerrier:
**Description**

This pull request cleans up the structure of the `Analysis` class. 
It modifies the handling of observation selection. Most of it is performed on the `ObservationTable`.
The creation of geometries, reference dataset and makers is now split in dedicated `_make_XXX` methods.
Code duplication could be further reduced once #3385 is merged.

**Dear reviewer**

PR #3426, opened 2021-07-13
Cleanup reflected regions finder

adonath:
**Description**

This pull request partly addresses #2267 and cleans up the implementation of the `ReflectedRegionsFinder`. the most relevant change here is to replace the `.setup()` methods by a `lazyproperty` pattern. This allowed me to replace the derived quantities from the finder configuration by lazy properties. I think the resulting implementation is very simple and clean.

### DIFF
--- a/gammapy/makers/background/reflected.py
+++ b/gammapy/makers/background/reflected.py
@@ -199,22 +199,25 @@ class ReflectedRegionsFinder:
             Reflected regions
         """
         self.reset_cache()
-        curr_angle = self.angle_min + self.min_distance_input
-        reflected_regions = []
+        regions = []
 
-        while curr_angle < self.angle_max:
-            test_reg = self.region_pix.rotate(self.center_pix, curr_angle)
-            if not np.any(test_reg.contains(self.excluded_pix_coords)):
-                region = test_reg.to_sky(self.geom_ref.wcs)
-                reflected_regions.append(region)
+        angle = self.angle_min + self.min_distance_input
 
-                curr_angle += self.angle_min
-                if self.max_region_number <= len(reflected_regions):
+        while angle < self.angle_max:
+            region_test = self.region_pix.rotate(self.center_pix, angle)
+
+            if not np.any(region_test.contains(self.excluded_pix_coords)):
+                region = region_test.to_sky(self.geom_ref.wcs)
+                regions.append(region)
+
+                if len(regions) >= self.max_region_number:
                     break
-            else:
-                curr_angle = curr_angle + self.angle_increment
 
-        return reflected_regions
+                angle += self.angle_min
+            else:
+                angle += self.angle_increment
+
+        return regions
 
 
 class ReflectedRegionsBackgroundMaker(Maker):
--- a/gammapy/makers/background/reflected.py
+++ b/gammapy/makers/background/reflected.py
@@ -3,6 +3,7 @@ import logging
 import numpy as np
 from astropy import units as u
 from astropy.coordinates import Angle
+from astropy.utils import lazyproperty
 from regions import PixCoord
 from gammapy.datasets import SpectrumDatasetOnOff
 from gammapy.maps import RegionGeom, RegionNDMap, WcsNDMap, WcsGeom
@@ -79,6 +80,7 @@ class ReflectedRegionsFinder:
         self.center = center
 
         self.angle_increment = Angle(angle_increment)
+
         if self.angle_increment <= Angle(0, "deg"):
             raise ValueError("angle_increment is too small")
 
@@ -86,10 +88,9 @@ class ReflectedRegionsFinder:
         self.min_distance_input = Angle(min_distance_input)
         self.exclusion_mask = exclusion_mask
         self.max_region_number = max_region_number
-        self.reflected_regions = None
         self.binsz = Angle(binsz)
 
-    @property
+    @lazyproperty
     def exclusion_mask_ref(self):
         """Exlcusion mask reprojected"""
         if self.exclusion_mask:
@@ -99,7 +100,7 @@ class ReflectedRegionsFinder:
 
         return mask
 
-    @property
+    @lazyproperty
     def geom_ref(self):
         """Reference geometry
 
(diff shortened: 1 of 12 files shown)


gammapy/makers/background/reflected.py
```diff
@@ -179,74 +159,69 @@ def _region_angular_size(pixels, center):
...
-        # Extract all pixcoords in the geom
-        X, Y = geom.get_pix()
-        ONpixels = PixCoord(X[mask], Y[mask])
+    @lazyproperty
+    def exclusion_mask_ref(self):
+        """Exclusion mask reprojected"""
+        if self.exclusion_mask:
+            mask = self.exclusion_mask.interp_to_geom(self.geom_ref, fill_value=True)
+        else:
+            mask = WcsNDMap.from_geom(geom=self.geom_ref, data=1)
```
registerrier, line 168: why not creating directly the Map as boolean?

gammapy/makers/background/reflected.py
```diff
@@ -179,74 +159,69 @@ def _region_angular_size(pixels, center):
...
+        return mask
 
+    @lazyproperty
+    def excluded_pix_coords(self):
+        """Excluded pix coords"""
         # find excluded PixCoords
-        mask = self.reference_map.data == 0
-        self.excluded_pixcoords = PixCoord(X[mask], Y[mask])
+        pix_y, pix_x = np.where(~self.exclusion_mask_ref.data)
+        return PixCoord(pix_x, pix_y)
```
registerrier, line 177: Is it always equivalent to X[]?

gammapy/makers/background/reflected.py
```diff
@@ -179,74 +159,69 @@ def _region_angular_size(pixels, center):
...
+    def run(self):
+        """Find reflected regions.
 
-        self.reflected_regions = reflected_regions
+        Returns
+        -------
+        regions : list of `SkyRegion`
+            Reflected regions
+        """
+        self.reset_cache()
```
registerrier, line 205: Does it need to reset the cache since the `RegionsFinder` is re-instantiated at each `run` of the `ReflectedRegionsBackgroundMaker`?
  You could also define getters and no setters for the `center` and `region` to force users to create a new `RegionFinder` for a new set of region and center position.

gammapy/makers/background/tests/test_reflected.py
```diff
@@ -114,19 +110,18 @@ def test_find_reflected_regions(
...
 center = SkyCoord(0.5, 0.0, unit="deg")
 other_region_finder_param = [
     (RectangleSkyRegion(center, 0.5 * u.deg, 0.5 * u.deg, angle=0 * u.deg), 3),
     (RectangleSkyRegion(center, 0.5 * u.deg, 1 * u.deg, angle=0 * u.deg), 1),
-    (RectangleSkyRegion(center, 0.5 * u.deg, 1 * u.deg, angle=90 * u.deg), 0),
+    (RectangleSkyRegion(center, 0.5 * u.deg, 1 * u.deg, angle=90 * u.deg), 1),
     (EllipseSkyRegion(center, 0.1 * u.deg, 1 * u.deg, angle=0 * u.deg), 2),
     (EllipseSkyRegion(center, 0.1 * u.deg, 1 * u.deg, angle=60 * u.deg), 3),
-    (EllipseSkyRegion(center, 0.1 * u.deg, 1 * u.deg, angle=90 * u.deg), 0),
+    (EllipseSkyRegion(center, 0.1 * u.deg, 1 * u.deg, angle=90 * u.deg), 2),
```
registerrier, line 124: Why the change here?

registerrier (dismissed):
Thanks @adonath . It seems indeed clearer with a pattern where everything is defined on `__init__`  and `run` returns the list of regions.
I would even enforce that `center` and `region` cannot be set to remove the need for a clear_cache.

There is conflict with the new version of `Analysis` introduced in #3424

gammapy/makers/background/reflected.py
adonath: Good point, I tried:
```python3
m = Map.from_geom(geom=geom, data=True)
m = Map.from_geom(geom=geom, data=1, dtype=bool)
```
Neither seems to work, I'll investigate why...

gammapy/makers/background/reflected.py
adonath: Yes, I think so...I haven't checked the performance, but using `np.where()` seemed simpler to me...

gammapy/makers/background/reflected.py
adonath: Good point, indeed if `center` and `region` cannot be changed we can probably remove the reset cache pattern. I think in general the pattern is not bad, but I agree further restricting the attributes is even better...

gammapy/makers/background/tests/test_reflected.py
adonath: I will try to understand the change and comment again...

gammapy/makers/background/reflected.py
adonath: Done

gammapy/makers/background/reflected.py
adonath: Looking more closely, it would require to declare all the attributes as non-settable attributes, which creates a lot of extra code. I the only way to further improve and simplify would be to move more code to `RegionGeom` as it already bundles the `region` and `wcs`. This way one could cache on the `RegionGeom` object.

gammapy/makers/background/tests/test_reflected.py
adonath: # Before:
Screenshot 2021-07-14 at 10 14 50
# After:
Screenshot 2021-07-14 at 10 12 35

I think placing the regions for the last column is within the numerical precision of the algorithm, which depends on the bin  size of the tangential WCS. I'm not 100% why the behaviour changed, but result still seems OK.

adonath: In general one could change to a less sensitive test, but maybe it's better to keep the more sensitive one and check the results once by hand.

PR #3428, opened 2021-07-14
Support serialisation of TimeMapAxis

adonath:
**Description**

This pull request is a follow-up PR to #3393 to support serialisation of a `TimeMapAxis`. I includes the following changes:

- Add `.edges_min` and `.edges_max` attribute on the axis objects
- Refactor `TimeMapAxis.from_table()` to support the "gadf" format only. I decided to remove the generic format for now, but I'm happy to introduce it later once we have identified and document an internal format.
- Clean up `TimeMapAxis.__init__` 
- Change `TimeMapAxis.center` to **not** return a `Time` object

I think there are a few open points to discuss:
- How to handle Gammapy internal formats for e.g. tables?
- Should we keep the time scale on I/O or always convert to "tt"?

I opened https://github.com/open-gamma-ray-astro/gamma-astro-data-formats/pull/179 to allow for the additional keywords required for time axis serialisation in GADF.

### DIFF
--- a/gammapy/maps/axes.py
+++ b/gammapy/maps/axes.py
@@ -1850,7 +1850,8 @@ class TimeMapAxis:
         self._pix_offset = -0.5
         self._interp = interp
 
-        if np.any(self.time_delta < 0 * u.s):
+        delta = edges_min[1:] - edges_max[:-1]
+        if np.any(delta < 0 * u.s):
             raise ValueError("Time intervals must not overlap.")
 
     @property
--- a/gammapy/maps/axes.py
+++ b/gammapy/maps/axes.py
@@ -10,7 +10,7 @@ from astropy.io import fits
 from astropy.table import Column, Table, hstack
 from astropy.utils import lazyproperty
 from gammapy.utils.interpolation import interpolation_scale
-from gammapy.utils.time import time_ref_to_dict
+from gammapy.utils.time import time_ref_to_dict, time_ref_from_dict
 from .utils import INVALID_INDEX, edges_from_lo_hi
 
 
@@ -1730,7 +1730,11 @@ class MapAxes(Sequence):
                 if axcols is None:
                     break
 
-                axis = MapAxis.from_table(table, format=format, idx=idx)
+                # TODO: what is good way to check whether it is a time axis?
+                try:
+                    axis = TimeMapAxis.from_table(table, format=format, idx=idx)
+                except (KeyError, ValueError):
+                    axis = MapAxis.from_table(table, format=format, idx=idx)
                 axes.append(axis)
         elif format == "gadf-dl3":
             for column_prefix in IRF_DL3_AXES_SPECIFICATION.keys():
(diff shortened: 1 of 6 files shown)


gammapy/maps/axes.py
```diff
@@ -1797,41 +1818,45 @@ class TimeMapAxis:
...
-        if not len(edges_min.flatten()) == len(edges_max.flatten()):
-            raise ValueError("Time min and time max must have the same length.")
+        if not edges_max.unit.is_equivalent("s"):
+            raise ValueError(f"Time edges max must have a valid time unit, got {edges_max.unit}")
 
-        if not (np.all(edges_min == np.sort(edges_min))
-                and np.all(edges_max == np.sort(edges_max))):
-            raise ValueError("TimeAxis: edges values must be sorted")
+        if not edges_min.shape == edges_max.shape:
+            raise ValueError("Time min and time max must have the same shape,"
```
registerrier, line 1835: Here it should be `edges min and edges mas mast have the same shape`

gammapy/maps/axes.py
```diff
@@ -1797,41 +1818,45 @@ class TimeMapAxis:
...
-            raise ValueError("TimeMapAxis: non-linear scaling scheme are not supported yet.")
+            raise NotImplementedError(f"Non-linear scaling scheme are not supported yet, got {interp}")
+
+        self._edges_min = edges_min
+        self._edges_max = edges_max
+        self._reference_time = Time(reference_time)
+        self._pix_offset = -0.5
         self._interp = interp
 
+        delta = edges_min[1:] - edges_max[:-1]
```
registerrier, line 1853: this will break for 1 bin entry.

gammapy/maps/axes.py
```diff
@@ -1982,7 +2017,7 @@ def pix_to_idx(self, pix, clip=False):
     @property
     def center(self):
         """Return `~astropy.time.Time` at interval centers."""
-        return self.time_mid
+        return self.edges_min + 0.5 * self.time_delta
```
registerrier, line 2020: this does not return a time. So either change docstring or implementation.

registerrier: Note that the docstring of `time_delta` is incorrect too. the difference of `edges` will return a `Quantity` not a `TimeDelta`.

gammapy/maps/axes.py
```diff
@@ -1714,7 +1730,11 @@ def from_table(cls, table, format="gadf"):
                 if axcols is None:
                     break
 
-                axis = MapAxis.from_table(table, format=format, idx=idx)
+                # TODO: what is good way to check whether it is a time axis?
+                try:
```
registerrier, line 1734: Should we add tags for axes too?
I guess that this will break the gadf format.

registerrier (approved):

Indeed we should find a way to distinguish between serialized axes types. Can't the node type be used?

gammapy/maps/axes.py
adonath: Done

gammapy/maps/axes.py
adonath: I changed the docstring, I think returning a Quantity is the consistent solution, with `edges_min`, `edges_max`, `center` and the corresponding time based attributes `time_min`, `time_max`, `time_mid` (maybe we should rename to `time_center`?)

adonath: For `time_delta` I changed to the same pattern: `time_delta` returns a `TimeDelta`, while `bin_width` returns a `Quantity`. This way we are consistent with the standard `MapAxis`, but get the additional `.time_` attributes for convenience.

gammapy/maps/axes.py
adonath: Yes, this seems like a good long-term solution. However I think for now we can stick with the `try` and `error` block. I think it's safe: if there is a reference time present in the axis specification then it will create a `TimeMapAxis` otherwise a `MapAxis`.

gammapy/maps/axes.py
adonath: I think it's fine:
```python3
import numpy as np

a = u.Quantity(1, ndmin=1)
delta = a[1:] - a[:-1]
print(np.any(delta < 0))
```
`delta` will be an empty array that always evaluates to `False`. For a single bin the bins cannot be overlapping...

adonath:
Thanks a lot @registerrier for the review. I think I address all the comments, let me know if you have additional ones...

PR #3431, opened 2021-07-15
Add multiple selection in ObservationTable.select_observations

registerrier:
**Description**

This pull request add the possibility to pass a list of selection dictionaries to `ObservationTable.select_observations`. It would be even simpler if we could add one type of selection dictionary to pass the list of observations ids. 

**Dear reviewer**

This is a proposal. Feel free to comment

### DIFF
--- a/gammapy/data/obs_table.py
+++ b/gammapy/data/obs_table.py
@@ -189,7 +189,7 @@ class ObservationTable(Table):
 
         return self[mask]
 
-    def select_cone(self, center, radius, inverted=False):
+    def select_sky_circle(self, center, radius, inverted=False):
         """Make an observation table, applying a cone selection.
 
         Apply a selection based on the separation between the cone center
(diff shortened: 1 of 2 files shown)


adonath (approved):
Thanks @registerrier, makes sense to me! I just have a few general comments for improvement:
- Maybe we should split out the cone selection into a dedicated `.select_region(region=)` method
- For consistency with `Models` we could rename the highest level method from `select_observations ` to `select`
- For consistency with `datasets` and other places (`ack select_time`) we could rename to `.select_time()`
- We could maybe make the arguments in `select_observations` explicit, like:
```python3
def select_observations(time_interval=None, region=None, par_range= )
```
This is slightly less flexible, but allows for a cleaner documentation. Internally the arguments are passed to the individual `select_something()` methods.

registerrier:
One issue with a `.select_region` method is the necessity to rely on the wcs projection which will yield different results with the cone selection we have now especially for large angles. Large angles is relatively common use case for such selections
I have introduced a `select_cone` method instead.

registerrier:
I think we can merge as is @adonath .

The question of `select_time` vs `select_time_range` is real. One issue is that some `select_time` methods take a `time_interval` argument (or even `time_intervals`), while others (e.g. `Datasets.select_time` takes `time_min` and `time_max` arguments. We can homogenize this in a further PR.

I would keep `ObsTable.select_observations` since `ObsTable` is a table and not a sequence of `Observation`. 
Technically, one could argue that `filter` would be a more appropriate name.

adonath:
For the time handling we make a follow up PR.

Issue #3359, opened 2021-05-12
Unit consistency when plotting flux points and spectral models together


LauraOlivera:
**Gammapy version**
I am using `0.18.dev584+g3a1807fba`

**Bug description**
When plotting a spectral model on top of a `FluxPointsDataset.plot_fit()` the units in the axis change but not the value being plotted.
Taking a snippet from the `FluxPointsDataset` tests file we define one:
```python3
from gammapy.estimators import FluxPoints
from gammapy.modeling.models import LogParabolaSpectralModel, PowerLawSpectralModel, SkyModel
from astropy.table import Table
import matplotlib.pyplot as plt
from gammapy.datasets import FluxPointsDataset
from astropy import units as u


path = "$GAMMAPY_DATA/tests/spectrum/flux_points/diff_flux_points.fits"
data = FluxPoints.read(path)
data.table["e_ref"] = data.energy_ref.to("TeV")
model = SkyModel(
    spectral_model=PowerLawSpectralModel(
        index=2.3, amplitude="2e-13 cm-2 s-1 TeV-1", reference="1 TeV"
    )
)

obs_table = Table()
obs_table["TELESCOP"] = ["CTA"]
obs_table["OBS_ID"] = ["0001"]
obs_table["INSTRUME"] = ["South_Z20_50h"]

dataset = FluxPointsDataset(model, data, meta_table=obs_table) 
```
we have created a flux points dataset with flux points and a model. Doing `dataset.plot_fit()` results in a plot with flux units of `erg · cm-2 · s-1` by default:
```python3
ax1, ax2 = dataset.plot_fit()
ax1.set_ylim(1e-15,)
ax1.grid(which='both')
plt.show()
```
![image](https://user-images.githubusercontent.com/22521727/118002386-da36b380-b347-11eb-98a7-9cdfe090243b.png)

If I now want to plot **the same model** on the same plot:
```python3
ax_spectrum, ax_residuals = dataset.plot_fit()
model.spectral_model.plot(ax = ax_spectrum, energy_range=(1e-4, 5e-1) * u.TeV, energy_power=2)
ax_spectrum.set_ylim(1e-15,)
ax_spectrum.grid(which='both')

plt.show()
```
![image](https://user-images.githubusercontent.com/22521727/118002247-bc694e80-b347-11eb-9f36-c801804130bc.png)

[...]

PR #3438, opened 2021-07-21
Unify plot unit handling

adonath:
**Description**

This pull request unifies the handling of units for plotting in Gammapy. The proposed scheme is based on `quantity_support()`
and the matplotlib keywords `xunits` and `yunits`. See the follwing minimal example:
```python3

filename = "$GAMMAPY_DATA/hawc_crab/HAWC19_flux_points.fits"
fp = FluxPoints.read(filename)
fp.plot(sed_type="e2dnde", xunits=u.erg, yunits=u.Unit("erg cm-2 s-1"))
```

I have removed all cases, where we had arguments related to plot units, such as `flux_unit` and `energy_unit`. It is much easier to handle it with the proposed scheme above. 

Also addresses #3359.


adonath:
Because of the lack of real alternatives, I'll go ahead and merge the PR. A short Slack discussion proposed to do the same for time axis, based on https://docs.astropy.org/en/stable/api/astropy.visualization.time_support.html and unify the plot label format. Both can be addressed in follow up PRs.

PR #3439, opened 2021-07-22
Allow to choose scaling method per parameter 

### LINKED ISSUE #2434: Remove parameter scaling?

QRemy:
Add a `scaler` attribute on parameter so the choice of scaling method is not forced to be the same for all parameters. By default `scaler=None` so no scaling is applied. For amplitudes/energies parameters and all Naima parameters  the default is set to `scaler="scale10"`. Should resolve #2434.

### DIFF
--- a/gammapy/modeling/fit.py
+++ b/gammapy/modeling/fit.py
@@ -97,8 +97,6 @@ class Fit:
         interval can be adapted by modifying the the upper bound of the interval (``b``) value.
     store_trace : bool
         Whether to store the trace of the fit
-    scale_parameters : bool
-        Whether to scale parameters prior to the fit.
     """
 
     def __init__(
@@ -108,7 +106,6 @@ class Fit:
         covariance_opts=None,
         confidence_opts=None,
         store_trace=False,
-        scale_parameters=True
     ):
         self.store_trace = store_trace
         self.backend = backend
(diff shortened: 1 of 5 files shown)


adonath (approved):
Thanks a lot @QRemy, no further comments from my side. The CI test fails seems unrelated...

PR #3442, opened 2021-07-26
Fix region geom width computation

adonath:
**Description**

This pull request fixes a number of test fails that recently appeared in the CI builds:
- In `regions=0.5` the unit returned in `geom.width` change from `deg` to `arcsec`.
- The `compound_region_center` didn't behave stably and I added better boundary handling
- Adapt tests...

### DIFF
--- a/gammapy/maps/region.py
+++ b/gammapy/maps/region.py
@@ -60,6 +60,7 @@ class RegionGeom(Geom):
         if wcs is None and region is not None:
             if isinstance(region, CompoundSkyRegion):
                 center = compound_region_center(region)
+                print(f"Center: {center}")
             else:
                 center = region.center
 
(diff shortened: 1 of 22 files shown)


gammapy/maps/region.py
```diff
@@ -60,6 +60,7 @@ def __init__(self, region, axes=None, wcs=None, binsz_wcs="0.1 deg"):
         if wcs is None and region is not None:
             if isinstance(region, CompoundSkyRegion):
                 center = compound_region_center(region)
+                print(f"Center: {center}")
```
bkhelifi, line 63: We do not use log.info here?

adonath: Oops, this `print` statement was not meant to stay....I'll remove it, thanks!

adonath:
Last CI fail is unrelated...will merge now.

PR #3444, opened 2021-07-27
Clean up region nd map plotting and adapt flux points plotting

adonath:
**Description**

This pull request improves the `RegionNDMap.plot()` method and uses it in `FluxPoints.plot` to reduce code duplication. It include the following changes:

- Add a legend of multiple lines are plotted in `RegionNDMap.plot` (see example below)
- Use `RegionNDMap.plot` in `FluxPoint.plot()` to reduce code duplication
- For now I removed the `energy_power` option. This part I am not sure about. There is the option to provide the `sed_type` which already includes the most common `e2dnde` scaling. I feel like keeping the `energy_power` is maybe too much configuration and a simple handling with the `sed_type` is sufficient and preferable. Another reason is that the arbitrary energy scaling does not easily generalise to the plotting of TS profiles.

**Examples**
The following code shows the minimal example for the new label handling:
```python3
from gammapy.maps import MapAxis, TimeMapAxis, RegionNDMap
from astropy.time import Time
import numpy as np

energy_axis = MapAxis.from_energy_edges([1, 3, 10] * u.TeV)

time_ref =  Time('1999-01-01T00:00:00.123456789')

time_axis = TimeMapAxis(
    edges_min=[0, 1, 3] * u.d,
    edges_max=[0.8, 1.9, 5.4] * u.d,
    reference_time=time_ref
)
time_axis.time_format = "iso"

m = RegionNDMap.create("icrs;circle(0, 0, 1)", axes=[energy_axis, time_axis], unit="cm-2 s-1")
m.data = 10 + np.random.random(m.data.size)
```
The plotting of the time/energy map using `m.plot(axis_name="time")` results now in:
image

So showing the lines and labels per energy...

### DIFF
--- a/gammapy/catalog/gammacat.py
+++ b/gammapy/catalog/gammacat.py
@@ -363,7 +363,7 @@ class SourceCatalogGammaCat(SourceCatalog):
     >>> source.spectral_model().plot_error(energy_range)
     <AxesSubplot:xlabel='Energy [TeV]', ylabel='dnde [1 / (cm2 s TeV)]'>
     >>> source.flux_points.plot()
-    <AxesSubplot:xlabel='Energy (TeV)', ylabel='dnde (1 / (cm2 s TeV))'>
+    <AxesSubplot:xlabel='Energy [TeV]', ylabel='dnde (1 / (cm2 s TeV))'>
     """
 
     tag = "gamma-cat"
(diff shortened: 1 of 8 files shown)


bkhelifi:
Hi @adonath ,
Thanks for this nice feature, that reduces indeed code duplication. I have a question concerning your comment on the option `energy_power`.
As discussed once during a call, some felt that it is still useful to be able to plot fluxes with any power. A classical use case is the plot of E^2.7 * Flux, for electrons for example or in the future CR analysis.... 
Do you think that it is hard to implement now? I guess that in the future people may wish to have it...

adonath:
Thanks @bkhelifi, I added back the `energy_power` option. I don't have any strong preference here anyway...

PR #3445, opened 2021-07-29
Use FluxPoints object in LightcurveEstimator

adonath:
**Description**

This pull request makes the `LightcurveEstimator` return a `FluxPoints` object as well. It includes the following changes:
- Implement `FluxPoints.from_stack()`
- Implement `FluxPoints.slice_by_idx()`
- Implement `GTI.from_stack()`
- Implement `format = "lightcurve"` in `FluxPoints.to_table()`, which corresponds to the previous table format return by the `LightCurveEstimator`

The final `FluxPoints` object in the `LightCurveEstimator` is created from a list of `FluxPoints` and providing the `TimeMapAxis` to define the time intervals. It still misses to implement some more unit tests.


gammapy/estimators/flux_point.py
```diff
@@ -159,40 +159,6 @@ def write(self, filename, sed_type="likelihood", **kwargs):
...
-            tables.append(table[reference.colnames])
-
-        table_stacked = vstack(tables)
-        table_stacked.meta["SED_TYPE"] = "dnde"
-        table_stacked.sort("e_ref")
-        return cls.from_table(table=table_stacked, sed_type="dnde")
-
     @staticmethod
     def _convert_loglike_columns(table):
         # TODO: check sign and factor 2 here
```
bkhelifi, line 164: From the documentation of GADF (https://gamma-astro-data-formats.readthedocs.io/en/latest/spectra/flux_points/index.html#likelihood-columns) the loglike format does not include the 2 ("Log-Likelihood value of the best-fit model.", +log and not -log! So it should be negative).
In `to_table' of the `flux_point' code, the profile is made with the cash statistics (to be checked once again). And thus, the stored data are -2 * log... (see https://cxc.cfa.harvard.edu/sherpa/statistics/#cash). The stored values should be positive
So, I think that we should add the sign minus here.
In order to be sure about the input format (their could not follow the definition of GADF), one could test if the input data are really negative!

PS: I did not find any web documentation for `FluxMaps' and `CountsStatistic'

gammapy/estimators/flux_point.py
```diff
@@ -308,7 +277,9 @@ def to_table(self, sed_type="likelihood", format="gadf-sed", formatted=False):
             Flux points table
         """
         if format == "gadf-sed":
-            idx = (Ellipsis, 0, 0)
+            # TODO: what to do with GTI info?
```
bkhelifi, line 280: astropy.table has a member metadata ("meta: dict, optional -> Metadata associated with the table").
Why is it not possible to use that field?

gammapy/estimators/flux_point.py
bkhelifi, line 164: From the documentation of GADF (https://gamma-astro-data-formats.readthedocs.io/en/latest/spectra/flux_points/index.html#likelihood-columns) the loglike format does not include the 2 ("Log-Likelihood value of the best-fit model.", +log and not -log! So it should be negative).
In to_table' of the flux_point' code, the profile is made with the cash statistics (to be checked once again). And thus, the stored data are -2 * log... (see https://cxc.cfa.harvard.edu/sherpa/statistics/#cash). The stored values should be positive
So, I think that we should add the sign minus here.
In order to be sure about the input format (their could not follow the definition of GADF), one could test if the input data are really negative!

PS: I did not find any web documentation for FluxMaps' and CountsStatistic'

gammapy/estimators/lightcurve.py
```diff
@@ -405,40 +401,31 @@ def run(self, datasets):
...
-                    data = fp_table[column].quantity
-                row[column] = data
-
-            fp_table_flux = fp.to_table(sed_type="flux")
-            for column in fp_table_flux.colnames:
-                if "flux" in column:
-                    row[column] = fp_table_flux[column].quantity
-
-            rows.append(row)
+            # TODO: find a way to handle counts per dataset
```
bkhelifi, line 404: I did not get the issue... Is there an hidden approximation with this sum?

gammapy/estimators/flux_point.py
adonath: In principle this is possible, however the `GTI` information is often a complete table with start and stop times...

gammapy/estimators/lightcurve.py
adonath: Currently the counts are summed over all datasets, it would be good to keep the info per dataset for debugging. But this is future work and will not be addressed in this PR.

gammapy/estimators/flux_point.py
adonath: As discussed in the dev call this morning the test file seems to be incorrect. I'll implement a unit test and open a PR in the GADF repo.

adonath:
Thanks @bkhelifi! I'll go ahead and merge this PR. There will be a number of follow PR where we can always implement further suggestions for improvements...

PR #3446, opened 2021-07-30
corrected axis label units for PSF containment radius

ischigal:
**Description**

This PR fixes the units for the PSF containment radius plots. The unit of the x-axis was wrongly given in the y-axis label and vice versa.

**Dear reviewer**

This seems like a trivial fix but please tell me if I misunderstood something.


adonath (approved):
Thanks a lot @ischigal! No further comments from my side...

PR #3448, opened 2021-08-02
Use FluxPoints as lightcurve in gammapy.catalog

adonath:
**Description**

This pull request refactors `gammapy.catalog` to use `FluxPoints` to represent lightcurves of sources. For this it introduces the `fermi-fgl` format for `TimeMapAxis` to support the `Hist_Start` HDU in the Fermi catalogs.

**Dear reviewer**

### DIFF
--- a/docs/tutorials/api/catalog.ipynb
+++ b/docs/tutorials/api/catalog.ipynb
@@ -635,7 +635,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "lightcurve.table[:3]"
+    "lightcurve.to_table(format=\"lightcurve\", sed_type=\"flux\")"
    ]
   },
   {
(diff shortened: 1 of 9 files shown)


registerrier (approved):
Thanks @adonath. This looks good.

In the long run, we need a better way to pass format options to `Axes.read`.

adonath:
Thanks @registerrier! 

> In the long run, we need a better way to pass format options to Axes.read

I there anything specific you have in mind?

registerrier:
> Thanks @registerrier!
> 
> > In the long run, we need a better way to pass format options to Axes.read
> 
> I there anything specific you have in mind?

Well, probably a registry of external methods, so that this can be added without changing the core code of the axes/geom framework. This way one would not need to expose the specific format of 4fgl which is only specific to gammapy.catalog.

adonath:
Yes, I agree this makes a lot of sense. Methods like https://github.com/gammapy/gammapy/blob/master/gammapy/maps/axes.py#L1118 have grown a lot and a nice way to resolve this is a registry system for different formats.

PR #3449, opened 2021-08-03
Remove LightCurve class

adonath:
**Description**

This pull request is a follow up to #3448 and #3447. It removes the `LightCurve` class, which is then basically un-used. For now this PR contains the changes of #3448 as well, I'll rebase one #3448 has been merged.

**Dear reviewer**


adonath:
Will fix remaining dockets fails independently...

PR #3450, opened 2021-08-04
Add option full to WcsNDMap.convolve

registerrier:
**Description**

This pull request adapts `WcsNDMap.convolve` to use `spicy.signal.convolve`. The `use_fft` flag is replaced with a `convolve_method` to allow `auto` as a possible argument. 
The `mode` is now better taken into account. In particular, `mode='full'` is now supported by padding the map `geom` object with the size of the kernel. 

This will allow evaluating the models on smaller cutouts before applying the psf convolution in `MapEvaluator`.

**Dear reviewer**

### DIFF
--- a/gammapy/maps/hpxnd.py
+++ b/gammapy/maps/hpxnd.py
@@ -506,7 +506,7 @@ class HpxNDMap(HpxMap):
 
         return self._init_copy(data=smoothed_data)
 
-    def convolve(self, kernel, method="wcs-tan", **kwargs):
+    def convolve(self, kernel, convolution_method="wcs-tan", **kwargs):
         """Convolve map with a WCS kernel.
 
         It projects the map into a WCS geometry, convolves with a WCS kernel and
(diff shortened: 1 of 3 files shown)


gammapy/maps/hpxnd.py
```diff
@@ -506,7 +506,7 @@ def smooth(self, width, kernel="gauss"):
 
         return self._init_copy(data=smoothed_data)
 
-    def convolve(self, kernel, use_fft=True, method="wcs-tan", **kwargs):
+    def convolve(self, kernel, method="wcs-tan", **kwargs):
```
bkhelifi, line 509: the field `method' has no docstring

adonath (changes requested):
Thanks a lot @registerrier! One minor comment: maybe we use `method` in `WcsNDMap.convolve()` to be in line with Scipy's convolve? And rename the `method` keyword in `HpxNDMap.convolve()` (that we defined our own...) to `convolve_method` instead?

adonath (approved):
Thanks a lot @registerrier, no further comment s from my side...

PR #3451, opened 2021-08-04
Impose WcsGeom.cutout to have at least one bin.

registerrier:
**Description**

This pull request modifies the behavior of `WcsGeom.cutout` for small cutout sizes. Currently if the size is smaller than the pixel the resulting cutout has no pixels. Here we force the cutout to have at least one pixel. A warning is raised to warn the user.

**Dear reviewer**

### DIFF
--- a/gammapy/maps/wcs.py
+++ b/gammapy/maps/wcs.py
@@ -1,7 +1,6 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 import copy
 from functools import lru_cache
-import logging
 import numpy as np
 import astropy.units as u
 from astropy.coordinates import Angle, SkyCoord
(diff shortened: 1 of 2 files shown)


gammapy/maps/wcs.py
```diff
@@ -913,9 +915,15 @@ def cutout(self, position, width, mode="trim", odd_npix=False):
         """
         width = _check_width(width) * u.deg
 
+        binsz = self.pixel_scales
+        width_npix = (width / binsz).to_value("")
+
+        invalid_size = width_npix<1
+        if np.any(invalid_size):
```
adonath, line 922: Looking at the implementation I'm wondering whether we actually need the log statement. Maybe simplify the implementation to basically one line with `np.clip()` and document the behaviour in the docstring instead? I think the clipping behaviour is very reasonable and without alternative...so it maybe does not justify emitting a warning.

adonath (approved):

PR #3458, opened 2021-08-06
Support binned time series format in FluxPoints.to_table()

adonath:
**Description**

This pull request implements a `binned-time-series` format for `FluxPoints.to_table()` to support interaction with the `BinnedTimeSeries` object in Astropy, like so:
```python3
from astropy.timeseries import BinnedTimeSeries

fp = FluxPoints.read()
data = fp.to_table(sed_type="flux", format="binned-time-series")
series = BinnedTimeSeries(data=data)
```

This PR also improves the docstring  of `FluxPoints.to_table()`, but in general some more documentation is required.


adonath:
Remaining CI fail I unrelated and fixed in #3459.

PR #3461, opened 2021-08-10
Use correct n_sigma_ul for FluxPoints.to_table

registerrier:
**Description**

This pull request corrects a small bug. 
I haven't seen a test for `FluxPoints.to_table` to check for this. Should we add one?

**Dear reviewer**

### DIFF
--- a/gammapy/estimators/flux_point.py
+++ b/gammapy/estimators/flux_point.py
@@ -323,7 +323,7 @@ class FluxPoints(FluxMaps):
             table.meta["SED_TYPE"] = sed_type
 
             if self.n_sigma_ul:
-                table.meta["UL_CONF"] = np.round(1 - 2 * stats.norm.sf(2), 2)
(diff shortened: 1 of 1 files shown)


adonath (changes requested):
Thanks @registerrier! Indeed this doesn't;t seem to be tested, maybe add one assert statement wherever we call `.to_table()` in the tests?

gammapy/estimators/tests/test_flux_point_estimator.py
```diff
@@ -232,6 +232,8 @@ def test_run_pwl(fpe_pwl):
     actual = table["npred_null"].data
     assert_allclose(actual, [[942.], [398.], [14.]])
 
+    actual = table.meta["UL_CONF"]
+    assert_allclose(actual, 0.9544997)
```
adonath, line 236: +1 To keep a higher precision here, GADF does not specify anything...(https://gamma-astro-data-formats.readthedocs.io/en/latest/spectra/flux_points/index.html#header-keywords)

adonath (approved):
Please merge whenever CI is green...

PR #3464, opened 2021-08-11
Change UL threshold condition to sqrt TS

registerrier:
**Description**

This pull request proposes a change in the definition of the UL threshold relying on arts TS rather than TS. It is more natural as it is directly sigmas and it also provides a way to deal with negative fluctuations. Using TS only works if non negative values are forbidden.

**Dear reviewer**

### DIFF
--- a/gammapy/catalog/fermi.py
+++ b/gammapy/catalog/fermi.py
@@ -48,7 +48,7 @@ class SourceCatalogObjectFermiBase(SourceCatalogObject, abc.ABC):
     flux_points_meta = {
         "sed_type_init": "flux",
         "n_sigma": 1,
-        "ts_threshold_ul": 1,
+        "sqrt_ts_threshold_ul": 1,
         "n_sigma_ul": 2
     }
 
--- a/gammapy/catalog/hess.py
+++ b/gammapy/catalog/hess.py
@@ -613,7 +613,7 @@ class SourceCatalogObjectHGPS(SourceCatalogObject):
         table.meta["sed_type_init"] = "dnde"
         table.meta["n_sigma_ul"] = 2
         table.meta["n_sigma"] = 1
-        table.meta["ts_threshold_ul"] = 1
+        table.meta["sqrt_ts_threshold_ul"] = 1
         mask = ~np.isnan(self.data["Flux_Points_Energy"])
 
         table["e_ref"] = self.data["Flux_Points_Energy"][mask]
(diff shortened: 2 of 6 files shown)


adonath (commented):
Thanks @registerrier I have left on minor inline comment. I think you still need to adapt `gammapy.catalog`. I also openrd #3465 to improve a bit the test coverage. We can decide which one to merge first. 

In general I'm fine with the change, the decision for `ts_threhold_ul` was arbitrary to a certain degree, just motivated by the fact that in the reference  papers (HGPS / Fermi catalogs) the TS values is given as a threshold. 

I have one more related comment: negative TS values should not appear. If the TS becomes negative it means the fit failed. The fit should always improve the likelihood over the null hypothesis, except for the rare case where the null hypothesis is the best fit. In this case I guess TS should be zero. So I guess we should probably adapt the handling in the `ParameterEstimator` to prevent the case of the negative TS already early...

registerrier:
> Thanks @registerrier I have left on minor inline comment. I think you still need to adapt `gammapy.catalog`. I also opened #3465 to improve a bit the test coverage. We can decide which one to merge first.
> 
Then let's wait for #3465 to be merged.

> In general I'm fine with the change, the decision for `ts_threhold_ul` was arbitrary to a certain degree, just motivated by the fact that in the reference papers (HGPS / Fermi catalogs) the TS values is given as a threshold.
> 

Indeed. TS is more general since it can apply in any situation. Of course here the issue is what to do with negative fluctuations if we allow for them. As the range for UL is given as n_sigma I though it would be clearer to use the same thing for the significance threshold. 
We can also call it `sigma_threshold_ul` but the `sqrt_ts` is what we have used so far.

> I have one more related comment: negative TS values should not appear. If the TS becomes negative it means the fit failed. The fit should always improve the likelihood over the null hypothesis, except for the rare case where the null hypothesis is the best fit. In this case I guess TS should be zero. So I guess we should probably adapt the handling in the `ParameterEstimator` to prevent the case of the negative TS already early...

I suppose those negative values are simply rounding error noise. The minimum is never exactly the minimum so that fluctuations of delta stat of the order of 1e-9 are likely to occur no?
There might be a better way to handle these though.

adonath:
> I suppose those negative values are simply rounding error noise. The minimum is never exactly the minimum so that fluctuations of delta stat of the order of 1e-9 are likely to occur no?
There might be a better way to handle these though.

Ah, I see. Indeed if the values are that small it could be related to the precision of the optimisation. 

I just merged #3465.

PR #3467, opened 2021-08-12
Catch Runtime Exception from iminuit_confidence

registerrier:
**Description**

This pull request adds `RuntimeException` to the exception caught in `iminuit_confidence`. This avoids flux points failing when the iminuit minos does not properly converges.

**Dear reviewer**


adonath (approved):
Thanks @registerrier, I guess this is hard to test...so maybe we just merge as is.

PR #3468, opened 2021-08-12
Prevents spatial model integration if no psf is present and add test

registerrier:
**Description**

This pull request change the behavior of the `MapEvaluator.compute_flux_spatial` for `RegionGeom` datasets.

At present it can lead to undesired behavior because the spatial integration follows the `binsz_wcs` of the `RegionGeom`. By default the latter is set to 0.1 deg. In this condition, the integration will be incorrect for many situations where the region geom size is comparable.

A priori, `SpectrumDataset` objects have no PSF defined. It is unclear whether the option of  spatial evaluation is meaningful is this case. Hence the choice made here to not perform the spatial evaluation in the absence of a psf.

Note that this issue might be alleviated once a correct/adaptive `SpatialModel.integrate_geom` is in place. See e.g. #3440 

**Dear reviewer**

### DIFF
--- a/gammapy/datasets/map.py
+++ b/gammapy/datasets/map.py
@@ -2756,13 +2756,14 @@ class MapEvaluator:
             Psf-corrected, integrated flux over a given region.
         """
         if self.geom.is_region:
-            if self.geom.region is None:
+            # We don't estimate spatial contributions if no psf are defined
+            if self.geom.region is None or self.psf is None:
                 return 1
 
             wcs_geom = self.geom.to_wcs_geom(width_min=self.cutout_width).to_image()
             values = self.model.spatial_model.integrate_geom(wcs_geom)
 
-            if self.psf and self.model.apply_irf["psf"]:
+            if self.model.apply_irf["psf"]:
                 values = self.apply_psf(values)
             else:
                 axes = [self.geom.axes["energy_true"].squash()]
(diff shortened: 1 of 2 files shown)


adonath:
Thanks @registerrier! I think the behaviour makes sense, the spatial model evaluation is only reasonable when there is a PSF defined as well. Expect maybe for very extended sources...

I already added this regression test https://github.com/gammapy/gammapy/blob/master/gammapy/datasets/tests/test_spectrum.py#L133, which could be adapted here to be sensitive to not evaluating the spatial model...

registerrier:
> Thanks @registerrier! I think the behaviour makes sense, the spatial model evaluation is only reasonable when there is a PSF defined as well. Expect maybe for very extended sources...
> 
 
It is true that one might want to estimate e.g. diffuse emission contribution or pre-convolved spatial model. Then a possible option could be to add a parameter to `MapEvaluator.__init__`, e.g. `compute_spatial_flux` which would be set to `True` by default.

Then we could add that  property on `SpectrumDataset`. So that a user could set it if needed. Does it sound like a more flexible approach? 

> I already added this regression test https://github.com/gammapy/gammapy/blob/master/gammapy/datasets/tests/test_spectrum.py#L133, which could be adapted here to be sensitive to not evaluating the spatial model...

PR #3470, opened 2021-08-13
Implement Parameter.conf_min and Parameter.conf_max

adonath:
**Description**

This pull request modifies the boundary handling for the confidence estimation for the Scipy backend. I noticed that in the current version of the `spectral_analysis.ipynb` tutorial, the UL of the highest flux point is not computed properly. This is because there are hard-coded bounds for the confidence intervals which also depend on the error. There is also a remaining TODO (see https://github.com/gammapy/gammapy/blob/master/gammapy/modeling/scipy.py#L74 ).  

For now I adjusted the boundaries to rely on `scan_min` and `scan_max`. We could keep the behaviour and document it. Maybe you have better proposal for choosing default boundaries @registerrier?

**Dear reviewer**

### DIFF
--- a/gammapy/modeling/scipy.py
+++ b/gammapy/modeling/scipy.py
@@ -69,14 +69,11 @@ def _confidence_scipy_brentq(
     )
 
     lower_bound = parameter.factor
-    upper_bound = parameter.factor_max if upper else parameter.factor_min
-    if np.isnan(upper_bound):
-        # TODO: remove hard coded limits here...
-        upper_bound = parameter.factor
-        if upper:
-            upper_bound += 1e2 * parameter.error / parameter.scale
-        else:
-            upper_bound -= 1e2 * parameter.error / parameter.scale
+
+    if upper:
+        upper_bound = parameter.scan_max / parameter.scale
+    else:
+        upper_bound = parameter.scan_min / parameter.scale
 
     message, success = "Confidence terminated successfully.", True
     kwargs.setdefault("nbin", 1)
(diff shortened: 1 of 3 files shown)


gammapy/modeling/scipy.py
```diff
@@ -69,14 +69,11 @@ def _confidence_scipy_brentq(
...
-    upper_bound = parameter.factor_max if upper else parameter.factor_min
-    if np.isnan(upper_bound):
-        # TODO: remove hard coded limits here...
-        upper_bound = parameter.factor
-        if upper:
-            upper_bound += 1e2 * parameter.error / parameter.scale
-        else:
-            upper_bound -= 1e2 * parameter.error / parameter.scale
+
+    if upper:
```
registerrier, line 73: Why removing the initial condition? It there is a finite range defined it seems fair to stick to it no?

registerrier: In particular, the error message below suggests to change the parameter boundaries.

registerrier (commented):
Thanks @adonath. 

I think the parameter range should be the first interval to consider.

If the issue lies when `counts=0`, the solution might to explicitly treat these points differently in the `Estimator` by changing the range there no?

gammapy/modeling/scipy.py
adonath: The difference is that the parameter min/max value are not necessarily defined, while the `scan_min` / `scan_max` are derived from the parameter errors by default. But indeed one could implement the order like:
- check `par.min` and `par.max` first, if those are defined use those
- if not check `scan_min` and `scan_max` and use it 
- throw an error else...

adonath: For flux points estimation there is currently no access to the norm fit boundaries either...

adonath:
@registerrier I introduced the `conf_min` and `conf_max` attributes which behave as proposed in https://github.com/gammapy/gammapy/pull/3470#discussion_r688503321. Later those could be set independently by users...

PR #3474, opened 2021-08-17
Return min/max energy maps from energy_range methods

### LINKED ISSUE #3427: Dataset energy range handling
All `Dataset` classes define different energy ranges related to the definition of the different `mask` attributes. Currently a single energy range is derived even if the energy range varies with e.g. position. This issue outlines a possible solution to make the handling uniform and more general. First I propose to define three energy ranges, related to the the masks defined on the dataset:
- `.energy_range`, which is the global energy range derived from `.mask`
- `.energy_range_safe`, which is the energy range derived from `.mask_safe`
- `.energy_range_fit`, which is the global energy range derived from `.mask_fit`
Instead of a single value for the min and max energy we should return a `Map` object to keep the information on the position dependency. E.g. it can be useful to plot a map of the lower energy threshold of a flux map.

In addition we should mirror the API on the `Datasets` object. The challenge here is that the datasets can have different types and geometries. So it requires some handling of the energy ranges between the different datasets.

Finally it could be very useful to store the information one the energy ranges used for e.g. flux map and flux points estimation on the corresponding result objects. The logic here is to reduce the energy range maps on the input datasets, e.g.. by taking the min or max energies in each dataset / bin group.

alexiscoutinho:
**Description**
Closes #3427. This PR makes `MapDataset._energy_range` return minimum and maximum energy maps based on the provided `mask` instead of just the min/max values across the whole spatial area. It supports datasets holding maps with more than one non-spatial dimension. Therefore, the returned maps may not be images. When no minimum or maximum energy is appropriate at a certain 'coord', `np.nan` is returned. The test is based on the one from #3020.

### DIFF
--- a/gammapy/datasets/map.py
+++ b/gammapy/datasets/map.py
@@ -364,8 +364,8 @@ class MapDataset(Dataset):
                 energy_max = energy_min.copy()
         else:
             data_shape = geom.data_shape
-            energy_min = np.full(data_shape, energy[0])
-            energy_max = np.full(data_shape, energy[-1])
+            energy_min = np.full(data_shape, energy.value[0])
+            energy_max = np.full(data_shape, energy.value[-1])
 
         map_min = Map.from_geom(geom, data=energy_min, unit=energy.unit)
         map_max = Map.from_geom(geom, data=energy_max, unit=energy.unit)
--- a/gammapy/datasets/map.py
+++ b/gammapy/datasets/map.py
@@ -337,34 +337,53 @@ class MapDataset(Dataset):
         """Shape of the counts or background data (tuple)"""
         return self._geom.data_shape
 
-    def _energy_range(self, mask = None):
-        """Compute the energy range with or without the mask fit."""
-        energy = self._geom.axes["energy"].edges
-        energy_min, energy_max = energy[:-1], energy[1:]
+    def _energy_range(self, mask=None):
+        """Compute the energy range maps with or without the fit mask."""
+        geom = self._geom
+        energy = geom.axes["energy"].edges
+        idx = geom.axes.index_data("energy")
+        geom = geom.drop("energy")
 
         if mask is not None:
-            if mask.data.any():
-                mask = mask.data.any(axis=(1, 2))
-            else:
-                return None, None
-        else:
-            mask = None
+            mask = mask.data
+            if mask.any():
+                energy_shape = [1] * mask.ndim
+                energy_shape[idx] = -1
+                energy_min, energy_max = energy.value[:-1], energy.value[1:]
+                energy_min = energy_min.reshape(energy_shape)
+                energy_max = energy_max.reshape(energy_shape)
 
-        return u.Quantity([energy_min[mask].min(), energy_max[mask].max()])
+                energy_min = energy_min * mask
+                energy_min[~mask] = np.nan
+                energy_min = np.nanmin(energy_min, idx)
+
+                energy_max = (energy_max * mask).max(idx)
+                energy_max[energy_max == 0] = np.nan
+            else:
+                energy_min = np.full(geom.data_shape, np.nan)
+                energy_max = energy_min.copy()
+        else:
+            data_shape = geom.data_shape
+            energy_min = np.full(data_shape, energy[0])
+            energy_max = np.full(data_shape, energy[-1])
+
+        map_min = Map.from_geom(geom, data=energy_min, unit=energy.unit)
+        map_max = Map.from_geom(geom, data=energy_max, unit=energy.unit)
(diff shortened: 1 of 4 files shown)


gammapy/datasets/map.py
```diff
@@ -337,34 +337,53 @@ def data_shape(self):
...
+        """Compute the energy range maps with or without the fit mask."""
+        geom = self._geom
+        energy = geom.axes["energy"].edges
+        idx = geom.axes.index_data("energy")
+        geom = geom.drop("energy")
 
         if mask is not None:
-            if mask.data.any():
-                mask = mask.data.any(axis=(1, 2))
+            mask = mask.data
```
bkhelifi, line 348: Is not a bit strange to use the same variable name for 2 different objects?

gammapy/datasets/map.py
```diff
@@ -337,34 +337,53 @@ def data_shape(self):
...
         return self._geom.data_shape
 
-    def _energy_range(self, mask = None):
-        """Compute the energy range with or without the mask fit."""
-        energy = self._geom.axes["energy"].edges
-        energy_min, energy_max = energy[:-1], energy[1:]
+    def _energy_range(self, mask=None):
+        """Compute the energy range maps with or without the fit mask."""
+        geom = self._geom
+        energy = geom.axes["energy"].edges
```
bkhelifi, line 343: have we decided finally to always use the convention e_reco and e_true? If yes, here, it should be e_reco....

adonath: Its correct we now use consistently `energy` and `energy_true`, see https://docs.gammapy.org/0.18.2/irf/index.html#irf-axis-naming.

gammapy/datasets/map.py
adonath: Agreed one could maybe use `_energy_range(mask_map)` or similar...

gammapy/datasets/map.py
```diff
@@ -337,34 +337,53 @@ def data_shape(self):
...
+        idx = geom.axes.index_data("energy")
+        geom = geom.drop("energy")
 
         if mask is not None:
-            if mask.data.any():
-                mask = mask.data.any(axis=(1, 2))
+            mask = mask.data
+            if mask.any():
+                energy_shape = [1] * mask.ndim
```
adonath, line 350: I think you could simplify the implementation here by relying on `np.argmax()`. See the following minimal example:
```python3
import numpy as np

energy = [1, 3, 10] * u.TeV
data = np.ones((3, 2, 2), dtype=bool)
data[0, 0, 0] = False
data[:1, 1, 1] = False
data[:2, 0, 1] = False

idx = np.argmax(data, axis=0)
energy[idx]
```
`np.argmax()` will directly return the index of the first appearance. To get the upper energy range you could rely on `np.flip` like so:
```python3
data = np.flip(data, axis=0)
idx = np.argmax(data, axis=0)
energy[::-1][idx]
```

adonath (commented):
Thanks a lot @alexiscoutinho, I have left some inline comments to address.

gammapy/datasets/map.py
bkhelifi: Thanks Axel for the reminder!

adonath:
Thanks a lot for the simplifications @alexiscoutinho! The implementation just works as expected:
Screenshot 2021-08-18 at 11 19 12

The corresponding map of the minimal energy looks like:

image

Form my side this is fine maybe @registerrier and @bkhelifi have additional comments?

registerrier (approved):
Thanks @alexiscoutinho . This looks nice.

It is likely not very intuitive to get two (1,1,1) maps when asking for the energy range of a spectrum dataset. One could return quantities in that specific case (although they contain less information obviously).
Maybe in the future one could use `LabelMapAxis` to represent this as a single `Map`.

adonath:
@registerrier I think returning a `Map` object has many advantages, like having the possibility to plot (as I did above...), or getting the value at a certain sky coord. However I agree on using `LabelMapAxis`, or as a compromise maybe return a dict like: `{"energy_min": emin_map, "energy_max": max_map}`, this would be more explicit...

alexiscoutinho:
@adonath Furthermore, `SpectrumDataset`s with extra non-spatial dimensions, for whatever reason, would still require maps to be returned. So the return exception would only be for simple `SpectrumDataset`s.

adonath:
Indeed @alexiscoutinho, we have e.g. the use case where `RegionNDMap` has a time axis for flux points. In this case the energy threshold would be time dependent...

adonath (commented):
Thanks a lot @alexiscoutinho, excellent work!

Issue #3477, opened 2021-08-19
LightcurveEstimator breaks if some time intervals are empty 


registerrier:
**Gammapy version**

dev

**Bug description**
If the time intervals passed to the `LightcurveEstimator` are not all covered by some datasets, it fails.

**Expected behavior**
it should run and leave nan values for empty time bins.

**To Reproduce**
Change the number of time intervals to the light curve validation to 50. 
That is change line [53](https://github.com/gammapy/gammapy-benchmarks/blob/eeb1ff2556dc085d0261d67dc7b3c58a9e855504/validation/lightcurve/make.py#L53) from
```
n_time_bins = 35
```
to 
```
n_time_bins = 50
```
will yield:
```
python make.py run-analyses 1d
INFO:__main__:Run analysis.
INFO:__main__:Filter observations in time intervals
INFO:__main__:Number of observations produced 59.
INFO:__main__:Dataset creation in 1d.
INFO:__main__:Assign model on the datasets.
INFO:__main__:Run LightCurveEstimator in 1d.
Traceback (most recent call last):
  File "make.py", line 278, in <module>
    cli()
  File "/Users/terrier/Code/anaconda3/envs/gammapy-dev/lib/python3.7/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/Users/terrier/Code/anaconda3/envs/gammapy-dev/lib/python3.7/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/Users/terrier/Code/anaconda3/envs/gammapy-dev/lib/python3.7/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/terrier/Code/anaconda3/envs/gammapy-dev/lib/python3.7/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/terrier/Code/anaconda3/envs/gammapy-dev/lib/python3.7/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "make.py", line 69, in run_analyses
    perform_analysis(analysis_type, short_observations, target_position, time_intervals)
  File "make.py", line 114, in perform_analysis
    lc = lc_maker.run(datasets)
  File "/Users/terrier/Code/gammapy-dev/gammapy/gammapy/estimators/lightcurve.py", line 128, in run
[...]

PR #3479, opened 2021-08-19
Correct for empty time bins in LightCurveEstimator

registerrier:
**Description**

This pull request  corrects issue #3477 . 
The choice is here to simply drop empty time intervals rather than filling them with nan values. It seems easier to handle. Possibly the message mentioning empty time bin could be raised from `DEBUG` to `INFO`. 

**Dear reviewer**


adonath:
Thanks a lot @registerrier, fix and test look good to me! I just have the general comment on the consistency with the `FluxPointsEstimator`: the current behaviour for the `FluxPointsEstimator` is that it fills `NaN` for energy intervals which are empty, instead of dropping it. I think consistency would be good and we need to document it somewhere...

PR #3480, opened 2021-08-20
Implement FluxProfileEstimator

adonath:
**Description**

This pull request implement a prototype for a `FluxProfileEstimator`, which internally relies on the `FluxPointsEstimator` as well. It also returns a `FluxPoints` object. This approach saves code duplication and unifies the return types. It still misses implementation of unit tests.


adonath:
I'll go ahead and merge this now and continue working on a tutorial for this. Further improvements can be implemented in follow up PRs.

Issue #3481, opened 2021-08-20
FluxPoints.plot() fails if no ul is defined 


registerrier:
**Gammapy version**
dev

**Bug description**
if the flux points object has no upper limit defined (either from an input table or if the estimator is run without `ul` selection), the plot method will fail.

**Expected behavior**
It should plot the flux points without ul.

**To Reproduce**
A simple test:
```
path = make_path("$GAMMAPY_DATA/tests/spectrum/flux_points/diff_flux_points.fits")
table = Table.read(path)
table.remove_column('dnde_ul')
fp = FluxPoints.from_table(table, sed_type='dnde')
fp.plot()
```
yields:
```
self = <gammapy.estimators.flux_point.FluxPoints object at 0x13e9e05f8>, quantity = 'norm_ul'

    def _check_quantity(self, quantity):
        if quantity not in self.available_quantities:
            raise AttributeError(
>               f"Quantity '{quantity}' is not defined on current flux estimate."
            )
E           AttributeError: Quantity 'norm_ul' is not defined on current flux estimate.
```

**Other information**
Any other information you think will be useful for us to fix the issue can go here.

PR #3482, opened 2021-08-20
Correct for missing UL FluxPoints plot

registerrier:
**Description**

This pull request corrects issue #3481  and adds a test for plots with missing UL values in `FluxPoints`. A check that `fp.is_ul` is not fully NaN is added before the UL plot.

**Dear reviewer**


adonath (approved):
Thanks a lot @registerrier! No comments from my side...

PR #3484, opened 2021-08-24
Modification to region geom for performance

registerrier:
**Description**

This pull request corrects a bug in `SkyModel.contributes` where the `region_mask` was returning `False` inside the region leading the method to always return `False`. Consequently model spatial model evaluation was performed at each `npred` call.

Another change is made to `RegionGeom.center_skydir` to use `region.center` when it exists and the region is not compound. This also improves the run time. 

**Dear reviewer**

### DIFF
--- a/gammapy/modeling/models/cube.py
+++ b/gammapy/modeling/models/cube.py
@@ -252,7 +252,7 @@ class SkyModel(Model):
 
         if mask.geom.is_region and mask.geom.region is not None:
             geom = mask.geom.to_wcs_geom()
-            mask = ~geom.region_mask([mask.geom.region])
+            mask = geom.region_mask([mask.geom.region])
 
         try:
             mask_cutout = mask.cutout(
--- a/gammapy/modeling/models/cube.py
+++ b/gammapy/modeling/models/cube.py
@@ -257,7 +257,7 @@ class SkyModel(Model):
         try:
             mask_cutout = mask.cutout(
                 position=self.position,
-                width=(2 * self.evaluation_radius + CUTOUT_MARGIN) + margin,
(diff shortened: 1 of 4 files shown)


adonath (commented):
Thanks @registerrier! As `RegionGeom.region` is read-only and cannot be modified it's probably simpler and equally safe to just make `self._rectangle_bbox` or even `center_skydir` a `lazyproperty`?

gammapy/modeling/models/cube.py
```diff
@@ -252,7 +252,7 @@ def contributes(self, mask, margin="0 deg"):
 
         if mask.geom.is_region and mask.geom.region is not None:
             geom = mask.geom.to_wcs_geom()
-            mask = geom.region_mask([mask.geom.region])
+            mask = ~geom.region_mask([mask.geom.region])
```
adonath, line 255: There is a related test-fail: https://github.com/gammapy/gammapy/pull/3484/checks?check_run_id=3414904505#step:9:1975

Maybe it's good to add a unit test for this as well...

registerrier:
I have added a unit test for the `SkyModel.contributes` with region mask input. 
I have also reverted to using the rectangle box center setting these as lazyproperties. 
Since this is very inefficient on a single call, one should check whether this is a bottleneck in some cases.

registerrier:
The problem here is actually different. In the `spectrum_1d` benchmark, the dataset are written and read from disk. There is a big in `regions` v0.5 which causes regions to become  `excluded` after `Regions.parse` in fits format. The `RegionGeom` after serialization is therefore an exclusion region. `model.contributes` cannot return `True` in this case, leading to the numerous evaluations causing the performance issue.
See: https://github.com/astropy/regions/issues/400

I will update the PR accordingly.

PR #3486, opened 2021-08-31
Fix doc string in TemplateSpectralModel

AtreyeeS:
As pointed out by Giacomo d'Amico on slack, the arguments are not passed to `scipy.interpolate.interp1d` but to `ScaledRegularGridInterpolator`. I directly link to `scipy.interpolate.RegularGridInterpolator` to avoid links to links as discussed.


registerrier (approved):
Thanks @AtreyeeS

PR #3491, opened 2021-09-01
Use FluxMaps as return type for ExcessMapEstimator

adonath:
**Description**

This pull request changes the `ExcessMapEstimator` to return a `FluxMaps` object as well. This required adding `npred` based quantities to the `FluxMaps` object, such that it can represent the case where there is no exposure and thus no flux estimate present. Here is a list of the quantities:

- `FluxMaps.npred_ref` reference predicted counts, such that `npred_excess = npred_ref * norm`. The definition of `npred_ref` somewhat differs from the other reference spectra, because it needs to vary with position and is represented as a `Map`.
- `FluxMaps.npred_err` error on `npred_excess`
- `FluxMaps.npred_errp` positive error on `npred_excess`
- `FluxMaps.npred_errn` negative error on `npred_excess`
- `FluxMaps.npred_ul` upper limit on `npred_excess`

In the name there is one inconsistency, that I don't like: `npred` already refers to the total counts (including null...), while the npred that relates to the norm is named `npred_excess`. So we might consider renaming either `npred` to e.g. `npred_total` or work with  `npred_excess_err` etc. instead.

Again for illustration I have updated the notebook here https://github.com/adonath/notebooks-public/blob/master/excess-vs-ts-map-estimator.ipynb

In past PRs I have resolved the inconsistencies between the `TSMapEstimator` and `ExcessMapEstimator`:
- #3497
- #3569
- #3519

The results (including the npred quantities defined above) now agree nicely and I propose to use `FluxMaps` as return type for the `ExcessMapEstimator` as well.  The main question is whether `npred` should be introduced as an independent sed type, to be able to represent maps and spectra without knowing the effective area of the instrument.

### DIFF
--- a/gammapy/estimators/utils.py
+++ b/gammapy/estimators/utils.py
@@ -120,6 +120,10 @@ def estimate_exposure_reco_energy(dataset, spectral_model=None, normalize=True):
             the input dataset
     spectral_model: `~gammapy.modeling.models.SpectralModel`
             assumed spectral shape. If none, a Power Law of index 2 is assumed
+    normalize : bool
+        Normalize the exposure to the total integrated flux of the spectral model.
+        When not normalized it directly gives the predicted counts from the spectral
+        model.
 
     Returns
     -------
--- a/gammapy/estimators/utils.py
+++ b/gammapy/estimators/utils.py
@@ -111,7 +111,7 @@ def find_peaks(image, threshold, min_distance=1):
     return table
 
 
-def estimate_exposure_reco_energy(dataset, spectral_model=None):
+def estimate_exposure_reco_energy(dataset, spectral_model=None, normalize=True):
     """Estimate an exposure map in reconstructed energy.
 
     Parameters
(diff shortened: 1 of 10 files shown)


registerrier:
Thanks @adonath . This is a very interesting comparison.

Regarding UL, the error is within `TSMapEstimator`. As can be seen from the figure the UL is lower than the flux in some parts. This is because the UL is only the confidence interval without the best fit flux. This is corrected in #3492 

Indeed the exposure should be correlated in the `ExcessMapEstimator`. Do we correct for this in a separate PR or within this one?

Regarding computation time, both find the confidences with the `find_root` function. I guess one difference is the use of cython for the stat function. A better estimate of the range maybe?

adonath:
Here is and update of the excess map vs ts map estimator cross-check: https://github.com/adonath/notebooks-public/blob/master/excess-vs-ts-map-estimator.ipynb

adonath:
@registerrier, @AtreyeeS, @bkhelifi and @QRemy I updated the PR here and the results in the linked notebook. I'll present the PR again in the next dev call, but maybe you have time to review before.

gammapy/estimators/utils.py
```diff
@@ -111,7 +111,7 @@ def find_peaks(image, threshold, min_distance=1):
     return table
 
 
-def estimate_exposure_reco_energy(dataset, spectral_model=None):
+def estimate_exposure_reco_energy(dataset, spectral_model=None, normalize=True):
```
bkhelifi, line 114: A docstring explaning the argument `normalize` could be usefull

adonath: Done

adonath:
@registerrier Please do a final review. Then I can continue to work on the related PRs, such as #2949 and #3480.

adonath:
I consistently changed to `npred_excess` now. In order to go  ahead with #2949 and #3480, I'll merge this PR now.  I'm happy re-discuss the naming and change in a follow-up PR.

PR #3492, opened 2021-09-02
Correct UL computation for TS map

registerrier:
**Description**

This pull request corrects an error in the UL computation of the `TSMapEstimator`. The best fit flux was not added to the confidence interval, yielding under-evaluated limits.

**Dear reviewer**

do we correct here or use #3491 to do so? The latter being a draft I thought it could be simpler to do it here.


adonath (approved):
Great, thanks a lot @registerrier for investigating! The change looks all fine to me, maybe extend the description of `norm_ul` here https://docs.gammapy.org/dev/estimators/index.html#introduction?

I fully agree it's better to resolve some of the issues listed in #3491 one by one first.

PR #3494, opened 2021-09-03
Move map evaluator to new file

registerrier:
**Description**

This pull request moves the `MapEvaluator` definition and some tests into new files. This should make modifications easier by not mixing `MapDataset` and `MapEvaluator` into the same file.

**Dear reviewer**


adonath (dismissed):
Thanks a lot @registerrier, I agree this improves the file structure. By now the the `datasets/map.py` is way to long, one has to scroll ~2000 lines just to go to the `MapEvaluator`. One could even consider moving the `MapEvaluator` to `gammapy.models` and maybe rename to `ModelEvaluator`, but splitting it out is the first step in any case.

PR #3497, opened 2021-09-03
Fix TSMapEstimator norm error boundary handling

adonath:
**Description**

This pull request fix the boundary behaviour for errors computed by the `TSMapEstimator` mentioned in #3491. The for the failure was that the background was padded with zero, which resulted in `NaN` values for the second derivative with respect to the norm. It also removes the boundary mask, which is not needed anymore after padding is the default. 

It also changes the `n_sigma_ul` default for `ExcessMapEstimator` for consistency with all the other estimators.


adonath:
I will merge this PR now and update the comparison shown in #3941. Remaining issue can be addressed in follow up PRs.

PR #3498, opened 2021-09-08
Pass empty dataset in Datasets.stack_reduce

AtreyeeS:
`Datasets.stack_reduce()` works only if they all have the same Geom. In practice, they can be stackable, but each residing on smaller geoms.
This PR adds an option to pass a large empty dataset to stack onto.

Also, in `info_table` its annoying that the first dataset is called `stacked`. This changed in this PR


gammapy/datasets/core.py
```diff
@@ -433,10 +435,15 @@ def stack_reduce(self, name=None, nan_to_num=True):
             raise ValueError(
                 "Stacking impossible: all Datasets contained are not of a unique type."
             )
+        if self[0].tag == "FluxPointsDataset":
+            raise ValueError("Stacking not defined for FluxPointsDataset")
```
registerrier, line 439: Wouldn't it be better to move this to a `stack` method for `FluxPointsDataset` and return a specific message there with a `NotImplementedError`?

gammapy/datasets/core.py
```diff
@@ -433,10 +435,15 @@ def stack_reduce(self, name=None, nan_to_num=True):
...
 
-        stacked = self[0].to_masked(name=name, nan_to_num=nan_to_num)
+        if empty is not None:
```
registerrier, line 441: Isn't it possible to put the `empty` dataset as first element in the list?

registerrier (changes requested):
Thanks @Atreyee . This is missing tests. They should be added.

Yet, is it the best approach? You could also pass a target `geom` object and first validate that your datasets are all stackable on that geom.

Some other comments inline.

gammapy/datasets/core.py
```diff
@@ -456,12 +463,13 @@ def info_table(self, cumulative=False):
         """
         if not self.is_all_same_type:
             raise ValueError("Info table not supported for mixed dataset type.")
+        if self[0].tag == "FluxPointsDataset":
```
registerrier, line 466: Looking at these tests, it might be that a `Datasets.type` property returning an `Enum` with something like:
* 0 = `heterogeneous`
* 1 = `MapDataset`
etc. might  be of use.

adonath:
Right now this is what users need to do, when they start from a list of cutout map datasets:
```python3
geoms = {name: WcsGeom.from_aligned(geom, skydir=, width=) for name, geom in cutout_dataset.geoms.items()}
empty = MapDataset.from_geoms(**geoms)
datasets.stack_reduce(empty=empty)

```

My proposal would be to move this code in the `.stack_reduce()` and change the API to take `skydir` and `width` as mandatory parameters in case.

AtreyeeS:
Hi @adonath : What you propose might be confusing for spectral datasets, no? Probably the `stack_reduce` is more relevant for MapDatasets, but I have found it rather useful for spectra as well.

adonath:
@AtreyeeS Note the API for spectral datasets remains the same:
```python3
spectral_datasets.stack_reduce()
```
However for map datasets one need to specify the `center` and `width` in addition:
```python3
map_datasets.stack_reduce(center=SkyCoord(), width=[5, 3] * u.deg)
```

adonath:
@AtreyeeS I'll move the milestone here to v1.0. We can implement this in a way such that it is backwards compatible in API with v0.19.

AtreyeeS:
I refactored this to take in width and skydir as discussed. I am getting a lot of unrelated test fails locally (probably related to Minuit version). @adonath If you can implement the rest of the desired changes (if any) directly, I will be very grateful.

AtreyeeS:
> @AtreyeeS I'll move the milestone here to v1.0. We can implement this in a way such that it is backwards compatible in API with v0.19.

AtreyeeS:
Closing this for now. We can implement a method to infer the best, all encompassing `geom` at a later stage

Issue #3490, opened 2021-09-01
Error in `Dataset.npred_signal`


AtreyeeS:
**Gammapy version**

present dev

**Bug description**

`MapDataset.npred_signal(model_name)` fails on the first call.
It works if `MapDataset.npred_signal()` (without arguments) is first called.

**To Reproduce**
 see this : https://github.com/AtreyeeS/test_notebooks/blob/master/error_npred.ipynb

Proposed solutions @registerrier @QRemy ?

QRemy:
Thanks @AtreyeeS,
https://github.com/gammapy/gammapy/blob/86c341315e035ec3347a51afbbace5845d294d12/gammapy/datasets/map.py#L440 
these lines could be modified to filter the evaluators instead of returning directly  `self.evaluators[model_name].computed_npred()`, such as:
```python
evaluators = self.evaluators
 if model_name is not None:
    evaluators = {model.name: self.evaluators[model_name]} 
for evaluator in evaluators.values():
```

In  addition it could be interesting to have `model_names` as a list and add a `evaluators.select` so ones can easily get the stacked npred of multiple models using npred_signal.

AtreyeeS:
Closed via #3500

PR #3500, opened 2021-09-08
Fix MapDataset.npred_signal for single model evaluation at first

QRemy:
Currently single model evaluation with `MapDataset.npred_signal(model_name) ` fails on the first call,  because the default case `MapDataset.npred_signal()`  have to be called first in order to initialize the evaluators. This PR fix this issue and change the order of the tests to cover this problem.

Solves #3490


AtreyeeS (approved):
Thanks @QRemy ! No comments from my side

registerrier (approved):
Thanks @QRemy. This looks good. No comment from my side.

Issue #3503, opened 2021-09-08
Error in TemplateSpatialModel serialisation


AtreyeeS:
**Gammapy version**
dev

**Bug description**
A `TemplateSpatialModel` cannot be read back after writing unless `filename` is explicitly specified. The importance of the `filename` is not specified anywhere, and it is rather annoying to find this error at an advanced stage of analysis.
I do not know if we should 
- solve this through documentation
- raise a `warning` if `filename` is not passed upon init
- make it compulsory to write the map as well

Moreover,  I do not want to use the whole map as the template, but a cutout of the same, which works fine on `__init__` but is not reflected in serialisation. 

opinions @registerrier @adonath @QRemy @bkhelifi ?

**To Reproduce**
see: https://github.com/AtreyeeS/test_notebooks/blob/f2c658786f0e307a07c61ca4612e689630da935e/models.ipynb

- content of `test.yaml`
```
-   name: diffuse
    type: SkyModel
    spectral:
        type: PowerLawSpectralModel
        parameters:
        -   name: index
            value: 2.7
            interp: lin
        -   name: amplitude
            value: 2.0e-13
            unit: cm-2 s-1 TeV-1
            interp: log
            scale_method: scale10
        -   name: reference
            value: 1.0
            unit: TeV
            frozen: true
            interp: lin
    spatial:
        type: TemplateSpatialModel
        frame: galactic
        parameters: []
        filename: null
        normalize: true
        unit: 1 / sr
covariance: test_covariance.dat
```

QRemy:
Thanks @AtreyeeS, using TemplateSpatialModel.read() directly setup the filename correctly, so we should adapt the documentation where it is needed (the Fermi-LAT tutorial for example).

AtreyeeS:
HI @QRemy ! What I mean is, I do not always use the exact template from the file but do some additional operations on the `map` before passing it to `TemplateSpatialModel` - this is not reflected in `read/write`. Should we write the `map` as well then?

QRemy:
yes the modified map should be written as well, and I think the users should do it themself as they have to define a different name.

registerrier:
So maybe, here the solution is to have a warning raised when running `Models.write` if there is no filename present.

QRemy:
> So maybe, here the solution is to have a warning raised when running `Models.write` if there is no filename present.

or in `TemplateSpatialModel.to_dict()` as `Models.write` call this.

AtreyeeS:
[...]

PR #3506, opened 2021-09-10
Safer TemplateSpatialModel init

QRemy:
This PR add the following changes to TemplateSpatialModel init in order to solve #3503:
- if filename  is not None and not found on disk, add IOError 
- if filename is None, add a log.warning("filename is not set, model serialisation will not be possible if map is not written to disk beforehand")
- map attribute as a property and _map is a copy() of the input so it cannot be changed

Writing the map to disk should remain the responsability of the user  as it is a pre-processing operation that should be done only once. It is not the responsability of Models.write() that come later in the analysis chain and may be repeated multiple times.

### DIFF
--- a/gammapy/modeling/models/spatial.py
+++ b/gammapy/modeling/models/spatial.py
@@ -1010,7 +1010,7 @@ class TemplateSpatialModel(SpatialModel):
         self, map, meta=None, normalize=True, interp_kwargs=None, filename=None,
     ):
         if (map.data < 0).any():
-            log.warning("Diffuse map has negative values. Check and fix this!")
+            log.warning("Map has negative values. Check and fix this!")
 
         if filename is not None:
             filename = str(make_path(filename))
--- a/gammapy/modeling/models/spatial.py
+++ b/gammapy/modeling/models/spatial.py
@@ -1014,12 +1014,6 @@ class TemplateSpatialModel(SpatialModel):
 
         if filename is not None:
             filename = str(make_path(filename))
-            if not os.path.isfile(filename):
-                raise IOError("File not found")
-        else:
-            log.warning(
-                "filename is not set, model serialisation will not be possible if map is not written to disk beforehand"
-            )
 
         self.normalize = normalize
 
@@ -1051,6 +1045,7 @@ class TemplateSpatialModel(SpatialModel):
 
     @property
     def map(self):
+        """Template map  (`~gammapy.maps.Map`)"""
         return self._map
 
     @property
@@ -1122,6 +1117,12 @@ class TemplateSpatialModel(SpatialModel):
         data["filename"] = self.filename
         data["normalize"] = self.normalize
         data["unit"] = str(self.map.unit)
+        if self.filename is None:
+            raise IOError("Missing filename")
+        elif os.path.isfile(self.filename) :
+            log.warning("Template file already exits, it will not be overwritten")
+        else:
+            self.map.write(self.filename)
         return data
 
     def to_region(self, **kwargs):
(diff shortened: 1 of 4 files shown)


AtreyeeS:
> * if filename is None, add a log.warning("filename is not set, model serialisation will not be possible if map is not written to disk beforehand")

Maybe TemplateModels should **NOT** be serialised without a `filename`? Stronger than a warning?

gammapy/modeling/models/spatial.py
```diff
@@ -1003,10 +1010,16 @@ def __init__(
         self, map, meta=None, normalize=True, interp_kwargs=None, filename=None,
     ):
         if (map.data < 0).any():
-            log.warning("Diffuse map has negative values. Check and fix this!")
+            log.warning("Map has negative values. Check and fix this!")
 
         if filename is not None:
             filename = str(make_path(filename))
+            if not os.path.isfile(filename):
```
adonath, line 1017: This check doesn't make any sense to me. Why would the file that is written to need to exist beforehand?

gammapy/modeling/models/spatial.py
```diff
@@ -1003,10 +1010,16 @@ def __init__(
...
+                raise IOError("File not found")
+        else:
+            log.warning(
```
adonath, line 1020: Again I think having this check in `__init__` doesn't make sense to me. The model could be created without being serialise...why not move to `.to_dict()` instead?

gammapy/modeling/models/spatial.py
```diff
@@ -1036,6 +1049,10 @@ def __init__(
         self.filename = filename
         super().__init__()
 
+    @property
+    def map(self):
+        return self._map
```
adonath, line 1054: Add one line docstring and return type here.

adonath (changes requested):
Thanks @QRemy, I have left some inline comments...

gammapy/modeling/models/spatial.py
QRemy: My points was that `models.write` should write only the .yaml file and if users define **new maps** that their responsibility to save them  beforehand as it is an operation independent from the saving the model fit results.

 But ok I added a .write() line in to_dict if the file do not exit on disk, however if it already exits I just send a warning.
 I don't think we should overwrite by default or add an overwrite keyword in .to_dict() because then you likely need a different overwrite options for .fits and .yaml files in .write() as the latter may change more often.
 
I'm still convince that the best practice is to save new maps to disk before using them as templates and not rely on models.write to save templates...

gammapy/modeling/models/spatial.py
QRemy: done

gammapy/modeling/models/spatial.py
QRemy: moved to `to_dict()`

gammapy/modeling/models/spatial.py
```diff
@@ -1105,6 +1117,12 @@ def to_dict(self, full_output=False):
         data["filename"] = self.filename
         data["normalize"] = self.normalize
         data["unit"] = str(self.map.unit)
+        if self.filename is None:
+            raise IOError("Missing filename")
+        elif os.path.isfile(self.filename) :
```
AtreyeeS, line 1122: Can we have a `overwrite` option rather? The original file on disc will be overwritten only if explicitly desired?

QRemy: added a `TemplateSpatialModel.write(overwrite=False)` method and added extra keyword `overwrite_templates` to `Models.write()`, `.to_dict()` and `.to_yaml()` to propagates this option.

gammapy/modeling/models/spatial.py
QRemy: finally I added the two overwrite options for .fits and .yaml files in Models.write() ...

registerrier (approved):

PR #3508, opened 2021-09-14
[ci skip] Minor doc string bug fix

AtreyeeS:
Fix doc string example in `ImageProfileEstimator`


AtreyeeS:
I though that the `ExcessProfileEstimator` becomes redundant.
The `ImageProfileEstimator` does something quite simple (but useful) - we remove this as well?

PR #3511, opened 2021-09-14
Add example of computing a livetime map

AtreyeeS:
This PR adds an example of how to compute on-axis equivalent observation duration.
I have added it in the hess notebook for now.

Maybe we can
- move the `make_livetime_map` to `Gammapy.makers.utils`?
- I do not know if passing the `energy_axis_true` is necessary here. It can just compute it directly from the IRF and then create one map. I kept it in case people want the effective exposure above a certain threshold


adonath:
Thanks @AtreyeeS! For now just a few general thoughts from my side:
- I was wondering a bit about the definition of the quantity. I think the name "livetime map" is misleading. I think "on axis equivalent livetime" (as written by @AtreyeeS in the initial post) is already a much better choice. "Livetime map" is probably HESS jargon and we should avoid it.
- Personally I find the definition of the quantity odd, because it suggestions a variation of the lifetime with offset, which is just not the case. The effective area varies instead. Can you maybe comment @AtreyeeS or @registerrier and @bkhelifi on how this quantity is useful? 
- Currently the method does the stacking as well and does not handled any safe data range. Is this intended? E.g. it ignores the offset cut, which is usually applied. So one cannot compute a consistent map with the "normal" data reduction. 
- Independent of this PR I think computing an actual livetime map, just showing how much time is spend where on the sky could be useful as well.

AtreyeeS:
Thanks for the comment @adonath 
Indeed many of these are my doubts as well. Started this PR as a rough prototype to invite comments.

For the moment, I tried to emulate what is provided by ParisAnalysis within H.E.S.S.
From what I understand, it just computes the observation duration in hours, correcting for the acceptance. Since 30 mins at the centre of the camera can be 20 mins at the edge...

How should we handle the safe range? Only using the aeff cuts?

Application wise, I imagine this will be used pre-analysis - to see if a region is worth analysing/preparing proposals etc.
So better to keep it fast and simple.

adonath:
I just talked to @registerrier this morning and here are some more comments:
- Keep the computation of the map in the tutorial, introducing a helper method in Gammapy will not really fit the API design we have and it also seems to functionality is too specific
- In notebooks it's rather unusual to define methods, because one cannot import from it and there is additional overhead (such as docstring...while the notebook is basically already documentation itself). Possibly remove the method definition and just keep the code in a single cell.
- Make sure to use a dedicated term such as "on axis equivalent livetime" everywhere in the tutorial

docs/tutorials/data/hess.ipynb
```diff
@@ -22,31 +22,54 @@
   {
```
adonath, line 22: <p>Please use a more explicit description such as "on axis equivalent livetime" or similar...</p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/gammapy/gammapy/pull/3511/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='15'/>

adonath, line 22: <p>A few comments on the implementation:</p><ul><li>I think we should use cutouts here, otherwise the code will not be usable for larger maps</li><li>Just use the higher level API i.e.  <code>MapDatasetMaker.make_exposure</code></li><li> You can change the unit if a Map like <code>Map.to_unit()</code></li></ul>
<br/>

 _Reply via <a href='https://app.reviewnb.com/gammapy/gammapy/pull/3511/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='16'/>

AtreyeeS:
There are merge conflicts which, when I try to address, seems to corrupt the notebook. Closing this PR and starting anew.

PR #3514, opened 2021-09-15
Create an Observation object from a single file

cosimoNigro:
**Description**

It is a bit impractical to create a `DataStore` object each time we want to manipulate / check `Observation`s, without having the possibility of reading them directly from DL3 files.

`Observations` are what actually goes into the spectral analysis and it seems to me that the `DataStore` provides a class for organising a list of them: **but often I want to quickly check if the information in a single (or few) test DL3 files I am producing is correct, without having to create the `hdu` and `obs` index files each time!**

Note that there is a `from_event_files` method in `DataStore`, but the IRF components - quoting the docstring - "are found only if you have a ``CALDB`` environment variable set"...

*what changes are done?*

I added a `.from_file(cls, event_file, irf_file)` method in `Observation`. It takes different arguments for event and IRF files, in case they are not in the same file.

To search for IRF components in the file I added a function searching for the allowed `HDUCLAS4` keywords by the GADF and returning the corresponding IRF component implemented in gammapy. There is a 1:1 correspondence between them (keyword and IRF class / component). The same is not true for `EXTNAME`, for example, as this can change from production to production (e.g. `AEFF` vs `EFFECTIVE AREA`) and does not identify the class of IRF component produced.

**Dear reviewer**

Let me know if you agree with the changes proposed - @AtreyeeS, @LauraOlivera, @bkhelifi and @maxnoe agreed with (or did not protest) my proposal in slack, I assumed it was a good basis for a PR.

This is roughly what is available now:
```python
obs = Observation.from_file(
    event_file="$GAMMAPY_DATA/hess-dl3-dr1/data/hess_dl3_dr1_obs_id_020136.fits.gz", 
    irf_file="$GAMMAPY_DATA/hess-dl3-dr1/data/hess_dl3_dr1_obs_id_020136.fits.gz"
)
print(obs.events)
print(obs.available_irfs)
```
that gives 
```
EventList
---------

  Instrument       : H.E.S.S. Phase I
  Telescope        : HESS
  Obs. ID          : 20136

  Number of events : 11243
  Event rate       : 6.684 1 / s

  Time start       : 53090.123451203704
  Time stop        : 53090.14291879629

  Min. energy      : 2.23e-01 TeV
  Max. energy      : 1.04e+02 TeV
  Median energy    : 5.37e-01 TeV

  Max. offset      : 36.3 deg

['aeff', 'edisp', 'psf', 'bkg']
```

*P.S.* I seem to have a problem with the `peek` functions, though I changed nothing in the `EventList` or `irf` class. 
Will take a deeper look tomorrow.

### DIFF
--- a/gammapy/data/observations.py
+++ b/gammapy/data/observations.py
@@ -49,7 +49,7 @@ def load_irf_from_hdu_class(irf_file, hdu_class):
             if (read_hdu_class.lower() == hdu_class.lower()):
                 # we have found a hdu with this specific HDUCLAS4 hedaer keyword
                 extname = hdu.header["EXTNAME"]
-                component = IRF_REGISTRY.getattr(hdu_class)
+                component = IRF_REGISTRY.get_cls(hdu_class)
                 return component.read(irf_file, extname)
         except KeyError:
             # no 'HDUCLAS4' header keyword in this HDU (probably EVENTS or GTI)
--- a/gammapy/data/observations.py
+++ b/gammapy/data/observations.py
@@ -9,51 +9,48 @@ from astropy.units import Quantity
 from astropy.io import fits
 from astropy.table import Table
 from gammapy.utils.scripts import make_path
-from gammapy.utils.fits import LazyFitsData, earth_location_from_dict
+from gammapy.utils.fits import LazyFitsData, HDULocation, earth_location_from_dict
 from gammapy.utils.testing import Checker
 from .event_list import EventList, EventListChecker
 from .filters import ObservationFilter
 from .gti import GTI
 from .pointing import FixedPointingInfo
 from .hdu_index_table import HDUIndexTable
-from gammapy.irf import IRF_REGISTRY
 
 __all__ = ["Observation", "Observations"]
 
 log = logging.getLogger(__name__)
 
-def load_irf_from_hdu_class(irf_file, hdu_class):
-    """Search for an IRF component with the specified HDUCLAS4 keyword within
-    the irf_file. Return an instance of the corresponding IRF class implemented 
-    in Gammapy.
-    
-    Parameters
-    ----------
-    irf_file : str, path
-        path to the file containing the IRF components, does not matter if also
-        EVENTS and GTI HDUs are included in the file, these are avoided
-    hdu_class : str
-        one of the allowed HDUCLAS4 header keyword in VALID_HDU_CLASS
-    
-    Returns
-    -------
-    one of the IRF component implementated in `~gammapy.irf`;
-    if the specified HDUCLAS4 is not available, `None` is returned
+def load_irf_dict_from_file(filename):
     """
-    if hdu_class not in HDUIndexTable.VALID_HDU_CLASS:
-        raise KeyError(f"{hdu_class} is not a valid HDU CLASS.")
+    """
+    filename = make_path(filename)
+
+    hdulist = fits.open(make_path(filename))
     
-    for hdu in fits.open(irf_file):
-        try: 
-            read_hdu_class = hdu.header["HDUCLAS4"]
-            if (read_hdu_class.lower() == hdu_class.lower()):
-                # we have found a hdu with this specific HDUCLAS4 hedaer keyword
-                extname = hdu.header["EXTNAME"]
-                component = IRF_REGISTRY.get_cls(hdu_class)
-                return component.read(irf_file, extname)
-        except KeyError:
-            # no 'HDUCLAS4' header keyword in this HDU (probably EVENTS or GTI)
+    irf_dict = {}
+
+    for hdu in hdulist:
+        hdu_class = hdu.header.get("HDUCLAS1", "").lower()
+        
+        if hdu_class == "response":
+            hdu_class = hdu.header.get("HDUCLAS4", "").lower()
+        
+            loc = HDULocation(
+                hdu_class=hdu_class,
+                hdu_name=hdu.name,
+                file_dir=filename.parent,
+                file_name=filename.name
+            )
+            
+            for name in HDUIndexTable.VALID_HDU_TYPE:
+                if name in hdu_class:
+                    data = loc.load()
+                    # TODO: maybe introduce IRF.type attribute...
+                    irf_dict[name] = data
+        else : # not an IRF component
             continue
+    return irf_dict
     
 
 class Observation:
@@ -396,35 +393,31 @@ class Observation:
         -------
         `~gammapy.data.Observation` with the event and irf read from the file
         """
-        event_file = make_path(event_file)
-        irf_file = make_path(irf_file) if irf_file is not None else event_file
         events = EventList.read(event_file)
+        
         gti = GTI.read(event_file) 
-        aeff = load_irf_from_hdu_class(irf_file, "aeff_2d")
-        edisp = load_irf_from_hdu_class(irf_file, "edisp_2d")
-        # non-mandatory IRF components
+
+        irf_file = irf_file if irf_file is not None else event_file
+        irf_dict = load_irf_dict_from_file(irf_file)
+        
+        aeff = None
+        edisp = None
         psf = None
         bkg = None
-        # there are different options for the PSF, the first allowed HDU CLASS 
-        # that gives not None with read_irf_with_hdu_class will be returned
-        # - we assume only one PSF type per file is stored
-        for hdu_class in HDUIndexTable.VALID_HDU_CLASS:
-            if hdu_class.startswith("psf"): 
-                psf_tmp = load_irf_from_hdu_class(irf_file, hdu_class)
-                if psf_tmp is not None: 
-                    psf = psf_tmp
-                    break
-        # there are different options for the BKG, the first allowed HDU CLASS 
-        # that gives not None with read_irf_with_hdu_class will be returned
-        # - we assume only one PSF type per file is stored
-        for hdu_class in HDUIndexTable.VALID_HDU_CLASS:
-            if hdu_class.startswith("bkg"): 
-                bkg_tmp = load_irf_from_hdu_class(irf_file, hdu_class)
-                if bkg_tmp is not None: 
-                    bkg = bkg_tmp
-                    break
-        # the obs_info seems to be a dictionary of the header of the EVENTS
-        obs_info = Table.read(event_file, hdu="EVENTS").meta
+
+        if "aeff" in irf_dict.keys():
+            aeff = irf_dict["aeff"]
+        
+        if "edsip" in irf_dict.keys():
+            edisp = irf_dict["edisp"]
+        
+        if "psf" in irf_dict.keys():
+            psf = irf_dict["psf"]
+        
+        if "bkg" in irf_dict.keys():
+            bkg = irf_dict["bkg"]
+            
+        obs_info = events.table.meta
         return cls(
             obs_id = obs_info.get("OBS_ID"),
             obs_info=obs_info,
--- a/gammapy/data/observations.py
+++ b/gammapy/data/observations.py
@@ -421,7 +421,7 @@ class Observation:
         if "aeff" in irf_dict.keys():
             aeff = irf_dict["aeff"]
         
-        if "edsip" in irf_dict.keys():
+        if "edisp" in irf_dict.keys():
             edisp = irf_dict["edisp"]
         
         if "psf" in irf_dict.keys():
--- a/gammapy/data/observations.py
+++ b/gammapy/data/observations.py
@@ -8,7 +8,7 @@ from astropy.time import Time
 from astropy.units import Quantity
 from gammapy.utils.fits import LazyFitsData, earth_location_from_dict
 from gammapy.utils.testing import Checker
-from gammapy.irf.io import load_irf_dict_from_file
+from gammapy.irf import load_irf_dict_from_file
 from .event_list import EventList, EventListChecker
 from .filters import ObservationFilter
 from .gti import GTI
--- a/gammapy/data/observations.py
+++ b/gammapy/data/observations.py
@@ -6,17 +6,57 @@ import numpy as np
 from astropy.coordinates import SkyCoord
 from astropy.time import Time
 from astropy.units import Quantity
+from astropy.io import fits
+from gammapy.utils.scripts import make_path
 from gammapy.utils.fits import LazyFitsData, earth_location_from_dict
 from gammapy.utils.testing import Checker
-from .event_list import EventListChecker
+from .event_list import EventList, EventListChecker
 from .filters import ObservationFilter
 from .gti import GTI
 from .pointing import FixedPointingInfo
+from .hdu_index_table import HDUIndexTable
+from gammapy import irf
+import matplotlib.pyplot as plt
 
-__all__ = ["Observation", "Observations"]
+__all__ = ["Observation", "Observations", "read_irf_with_hdu_class"]
 
 log = logging.getLogger(__name__)
 
+irf_dictionary = HDUIndexTable.HDU_CLASS_IMPLEMENTATION
+
+def read_irf_with_hdu_class(irf_file, hdu_class):
+    """Search for an IRF component with the specified HDUCLAS4 keyword within
+    the irf_file. Return an instance of the corresponding IRF class implemented 
+    in Gammapy.
+    
+    Parameters
+    ----------
+    irf_file : str, path
+        path to the file containing the IRF components, does not matter if also
+        EVENTS and GTI HDUs are included in the file, these are avoided
+    hdu_class : str
+        one of the allowed HDUCLAS4 header keyword in VALID_HDU_CLASS
+    
+    Returns
+    -------
+    one of the IRF component implementated in `~gammapy.irf`;
+    if the specified HDUCLAS4 is not available, `None` is returned
+    """
+    if hdu_class not in HDUIndexTable.VALID_HDU_CLASS:
+        raise KeyError(f"{hdu_class} is not a valid HDU CLASS.")
+    
+    for hdu in fits.open(irf_file):
+        try: 
+            read_hdu_class = hdu.header["HDUCLAS4"]
+            if (read_hdu_class.lower() == hdu_class.lower()):
+                # we have found a hdu with this specific HDUCLAS4 hedaer keyword
+                extname = hdu.header["EXTNAME"]
+                component = getattr(irf, irf_dictionary[hdu_class])
+                return component.read(irf_file, extname)
+        except KeyError:
+            # no 'HDUCLAS4' header keyword in this HDU (probably EVENTS or GTI)
+            continue
+    
 
 class Observation:
     """In-memory observation.
@@ -341,6 +381,56 @@ class Observation:
         obs = copy.deepcopy(self)
         obs.obs_filter = new_obs_filter
         return obs
+
+    @classmethod
+    def from_file(cls, event_file, irf_file):
+        """Create an Observation from a Event List and IRF file.
+        Ease the creation of a single Observation without creating an entire `DataStore`
+        that is without the need of HDU and OBS index files
+        """
+        log.info(f"reading the Event List and GTI from {event_file}")
+        event_file = make_path(event_file)
+        if not event_file.exists():
+            raise OSError(f"File not found: {event_file}")
+        log.debug(f"Reading {event_file}")
+        events = EventList.read(event_file)
+        gti = GTI.read(event_file) 
+        log.info(f"reading the IRF components in {irf_file}")
+        irf_file = make_path(irf_file)
+        if not irf_file.exists():
+            raise OSError(f"File not found: {irf_file}")
+        aeff = read_irf_with_hdu_class(irf_file, "aeff_2d")
+        edisp = read_irf_with_hdu_class(irf_file, "edisp_2d")
+        aeff.peek()
+        plt.show()
+        # non-mandatory IRF components
+        psf = None
+        bkg = None
+        # there are different options for the PSF, the first not None is returned
+        # we assume a single PSF class is saved in the file
+        for hdu_class in HDUIndexTable.VALID_HDU_CLASS:
+            if hdu_class.startswith("psf"): 
+                psf_tmp = read_irf_with_hdu_class(irf_file, hdu_class)
+                if psf_tmp is not None: 
+                    psf = psf_tmp
+        # there are different options for the BKG, the first not None is returned
+        for hdu_class in HDUIndexTable.VALID_HDU_CLASS:
+            if hdu_class.startswith("bkg"): 
+                bkg_tmp = read_irf_with_hdu_class(irf_file, hdu_class)
+                if bkg_tmp is not None: 
+                    bkg = bkg_tmp
+        
+        return cls(
+            gti=gti,
+            aeff=aeff,
+            bkg=bkg,
+            edisp=edisp,
+            psf=psf,
+            events=events,
+        )
+        
+        
+
 
 
 class Observations(collections.abc.MutableSequence):
--- a/gammapy/data/observations.py
+++ b/gammapy/data/observations.py
@@ -432,7 +432,7 @@ class Observation:
             
         obs_info = events.table.meta
         return cls(
-            obs_id = obs_info.get("OBS_ID"),
+            obs_id=obs_info.get("OBS_ID"),
             obs_info=obs_info,
             gti=gti,
             aeff=aeff,
--- a/gammapy/data/observations.py
+++ b/gammapy/data/observations.py
@@ -383,22 +383,26 @@ class Observation:
         return obs
 
     @classmethod
-    def from_file(cls, event_file, irf_file):
-        """Create an Observation from a Event List and IRF file.
-        Ease the creation of a single Observation without creating an entire `DataStore`
-        that is without the need of HDU and OBS index files
+    def from_file(cls, event_file, irf_file=None):
+        """Create an Observation from a Event List and an (optional) IRF file.
+
+        Parameters
+        ----------
+        event_file : str, Path
+            path to the .fits file containing the event list and the GTI
+        irf_file : str, Path
+            (optional) path to the .fits file containing the IRF components,  
+            if not provided the IRF will be read from the event file
+
+        Returns
+        -------
+        `~gammapy.data.Observation` with the event and irf read from the file
         """
         log.info(f"reading the Event List and GTI from {event_file}")
         event_file = make_path(event_file)
-        if not event_file.exists():
-            raise OSError(f"File not found: {event_file}")
-        log.debug(f"Reading {event_file}")
+        irf_file = make_path(irf_file) if irf_file is not None else event_file
         events = EventList.read(event_file)
         gti = GTI.read(event_file) 
-        log.info(f"reading the IRF components in {irf_file}")
-        irf_file = make_path(irf_file)
-        if not irf_file.exists():
-            raise OSError(f"File not found: {irf_file}")
         aeff = read_irf_with_hdu_class(irf_file, "aeff_2d")
         edisp = read_irf_with_hdu_class(irf_file, "edisp_2d")
         # non-mandatory IRF components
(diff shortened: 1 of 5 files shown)


gammapy/data/observations.py
```diff
@@ -342,6 +382,61 @@ def select_time(self, time_interval):
         obs.obs_filter = new_obs_filter
         return obs
 
+    @classmethod
+    def from_file(cls, event_file, irf_file):
```
maxnoe, line 386: Can we have a default for `irf_file=None` that would mean "Look for the IRFs in the `event_file`"?

gammapy/data/observations.py
```diff
@@ -342,6 +382,61 @@ def select_time(self, time_interval):
...
+        """
+        log.info(f"reading the Event List and GTI from {event_file}")
+        event_file = make_path(event_file)
+        if not event_file.exists():
+            raise OSError(f"File not found: {event_file}")
+        log.debug(f"Reading {event_file}")
+        events = EventList.read(event_file)
+        gti = GTI.read(event_file) 
+        log.info(f"reading the IRF components in {irf_file}")
+        irf_file = make_path(irf_file)
```
maxnoe, line 399: e.g. just change this to
```
irf_file = make_path(irf_file) if irf_file is not None else event_file
```
?

maxnoe (commented):
Thanks @cosimoNigro, in general looks good.

I added a comment regarding making `irf_file` optional in case the IRFs are stored in the events file.

Regarding the IRF reading: wouldn't it be better to just loop over the HDUs in the irf file and check what HDUCLASSES they have instead of looking through all possible HDU formats if they are in the file?

Also, you should probably `break` as soon as you found your psf / bkg HDU.

gammapy/data/observations.py
```diff
@@ -342,6 +382,61 @@ def select_time(self, time_interval):
...
 
+    @classmethod
+    def from_file(cls, event_file, irf_file):
+        """Create an Observation from a Event List and IRF file.
+        Ease the creation of a single Observation without creating an entire `DataStore`
+        that is without the need of HDU and OBS index files
+        """
+        log.info(f"reading the Event List and GTI from {event_file}")
+        event_file = make_path(event_file)
+        if not event_file.exists():
```
adonath, line 393: Catching this error seems unnecessary. If the file does not exist `EventList.read()` should fail the same way.

gammapy/data/observations.py
```diff
@@ -342,6 +382,61 @@ def select_time(self, time_interval):
...
+            raise OSError(f"File not found: {event_file}")
+        log.debug(f"Reading {event_file}")
+        events = EventList.read(event_file)
```
adonath, line 396: Here you can directly use the `HDULocation` system. Here is a minimal example:
```python3
from gammapy.utils.fits import HDULocation

loc = HDULocation(
    hdu_class="events",
    base_dir="$GAMMAPY_DATA/hess-dl3-dr1/data",
    file_dir="",
    file_name="hess_dl3_dr1_obs_id_020136.fits.gz"
)

events = loc.load()
```

gammapy/data/observations.py
```diff
@@ -342,6 +382,61 @@ def select_time(self, time_interval):
...
+        gti = GTI.read(event_file) 
+        log.info(f"reading the IRF components in {irf_file}")
+        irf_file = make_path(irf_file)
+        if not irf_file.exists():
```
adonath, line 400: See comment above...

gammapy/data/observations.py
adonath, line 386: Here I'm leaning towards just calling this `Observation.read()`, I think `.from_file()` is only used in `DataStore.from_file()`, but for all the other classes in Gammapy we use `.read()`.

adonath: I think for simulations it's also useful to make the event file optional as well. This e.g. will allow us to finally get rid of https://github.com/gammapy/gammapy/blob/master/gammapy/irf/io.py#L76

adonath (changes requested):
Thanks a lot @cosimoNigro, I agree this is useful functionality! I have left comments on the implementation. I think it's important to avoid the duplicated IRF registry and rely on the system we already have in Gammapy instead. Otherwise it will be hard to maintain if new / modified / versions of DL3 formats are added.

AtreyeeS:
Thanks @cosimoNigro ! I wonder if `Observation.from_file(irf_file=)` is the best solution.
This assumes that all the different IRFs are shipped in one file. This is not the case for some HESS productions at the moment, and I do not think its a requirement from the GADF either.

maxnoe:
> This assumes that all the different IRFs are shipped in one file.

You have multiple files for each IRF component? That's a bit overcomplicating things and would also be very easy to merge. However, if we really do want to support this, we could use `irf_files` and support a sequence of files.

That would again make it important to look what is in the files, not go through all possible IRF types and see if they are there.

However, as this is convenience functionality in case there is no HDU index, is this really a common usecase? Wouldn't it be simpler yo just add the IRFs to the Events file / a single IRF file?

AtreyeeS:
> Wouldn't it be simpler yo just add the IRFs to the Events file / a single IRF file?

Adding the IRFs to the events file is rather restrictive. As @LauraOlivera mentioned in the dev channel, you can have, eg, in HAWC, only one set of IRFs to be shared by all event files.

I do not know about merging all IRFs together. If, eg, there are different productions of background IRFs, its not desirable to have to keep merging HDUs, since this is at a stage when you are simple testing if the `Observation` is well behaved. This might be the case of some HESS productions. Please correct me if I am mistaken, @bkhelifi

gammapy/data/observations.py
```diff
@@ -6,17 +6,57 @@
...
+from .hdu_index_table import HDUIndexTable
+from gammapy import irf
 
 __all__ = ["Observation", "Observations"]
 
 log = logging.getLogger(__name__)
 
+irf_dictionary = HDUIndexTable.HDU_CLASS_IMPLEMENTATION
+
+def read_irf_with_hdu_class(irf_file, hdu_class):
```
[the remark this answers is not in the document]
adonath: Here is one more idea: `Map.read()` acts as a "factory method" for maps. This means `Map.read()` will try to guess the map type from the file and return the correct instance off the class, such as `WcsNDMap`, `HpxNDMap` or `RegionNDMap`. It might be convenient to introduce the same system for `IRF.read()`...

gammapy/data/observations.py
adonath: As it's a convenience function anyway, one could also simplify the logic and just allow for one filename. In this case it means to just read everything that's contained in the file. If the data is stored in two separate files, the following works as well, just with one additional line of code:

```python3
obs = Observation.read("my-dl3-irfs.fits")
obs.events = EventList.read("my-obs-events.fits")
```

gammapy/data/observations.py
```diff
@@ -342,6 +382,61 @@ def select_time(self, time_interval):
         obs.obs_filter = new_obs_filter
         return obs
 
+    @classmethod
+    def from_file(cls, event_file, irf_file):
+        """Create an Observation from a Event List and IRF file.
+        Ease the creation of a single Observation without creating an entire `DataStore`
+        that is without the need of HDU and OBS index files
```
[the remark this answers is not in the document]
cosimoNigro: Added "Returns" and shortened the docstring.

gammapy/data/observations.py
```diff
@@ -342,6 +382,61 @@ def select_time(self, time_interval):
...
+        """
+        log.info(f"reading the Event List and GTI from {event_file}")
```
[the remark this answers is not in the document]
cosimoNigro: Sorry did not check how is it done in other classes, removed logging.

gammapy/data/observations.py
cosimoNigro: Ok, did not know that the `.read` would have failed, removed the check.

gammapy/data/observations.py
maxnoe: That looks much more complicated than `EventList.read()`, what is the advantage?

gammapy/data/observations.py
cosimoNigro: This is done now.

gammapy/data/observations.py
adonath: Agreed. In the case of event lists there is no clear advantage to go via the registry / lazy loading interface here, because there is only one class to represent the data. The main remaining difference is that `HDULocation` triggers the lazy loading. So events are only read when `Observation.events` is accessed. But again in this case the lazy loading does not matter, so indeed one could keep `EventList.read()` as well.

gammapy/data/observations.py
cosimoNigro: Also done.

cosimoNigro:
After some considerations:

@adonath: I know that my dictionary
```python
HDU_CLASS_IMPLEMENTATION = {
    "events" : "EventList",
    "gti" : "GTI",
    "aeff_2d" : "EffectiveAreaTable2D",
    "edisp_2d" : "EnergyDispersion2D",
    "psf_table" : "PSF3D",
    "psf_3gauss" : "EnergyDependentMultiGaussPSF",
    "psf_king" : "PSFKing",
    "bkg_2d" : "Background2D",
    "bkg_3d" : "Background3D"
}
```
duplicates the `IRFRegistry`. 
But here is the problem: there is this other dictionary
https://github.com/gammapy/gammapy/blob/master/gammapy/irf/io.py#L18
linking the `hdu_class` with some other properties of the FITS files, among them the `extname`s. 
This dictionary is called, along with the `IRFRegistry`, by  this `HDULocation` you suggested. 
But - I cannot figure why - IRF components are searched by their `extname`s that are not standardised!
So we have this problem:
```python
from gammapy.utils.fits import HDULocation

loc = HDULocation(
    hdu_class="aeff_2d",
    base_dir="$GAMMAPY_DATA/cta-1dc/caldb/data/cta/1dc/bcf/South_z20_50h/",
    file_dir="",
    file_name="irf_file.fits"
)
aeff = loc.load()
print(aeff)
```  
gives
```
EffectiveAreaTable2D
--------------------

  axes  : ['energy_true', 'offset']
  shape : (42, 6)
  ndim  : 2
  unit  : m2
  dtype : >f4
```

But when we go to read to read the H.E.S.S. data - that have `AEFF` instead of `EFFECTIVE AREA` for `extname`.
```python
from gammapy.utils.fits import HDULocation

loc = HDULocation(
    hdu_class="aeff_2d",
    base_dir="$GAMMAPY_DATA/hess-dl3-dr1/data",
    file_dir="",
    file_name="hess_dl3_dr1_obs_id_020136.fits.gz"
)
aeff = loc.load()
```
we get
```python
KeyError                                  Traceback (most recent call last)
<ipython-input-3-010e6b0bdd4a> in <module>
      8 )
      9 
---> 10 aeff = loc.load()

~/work/gammapy/gammapy/utils/fits.py in load(self)
     97             cls = IRF_REGISTRY.get_cls(hdu_class)
     98 
---> 99             return cls.read(filename, hdu=hdu)
    100 
    101 

~/work/gammapy/gammapy/irf/core.py in read(cls, filename, hdu, format)
    360         """
    361         with fits.open(str(make_path(filename)), memmap=False) as hdulist:
--> 362             return cls.from_hdulist(hdulist, hdu=hdu)
    363 
    364     @classmethod

~/work/gammapy/gammapy/irf/core.py in from_hdulist(cls, hdulist, hdu, format)
    339             hdu = IRF_DL3_HDU_SPECIFICATION[cls.tag]["extname"]
    340 
--> 341         return cls.from_table(Table.read(hdulist[hdu]), format=format)
    342 
    343     @classmethod

~/software/miniconda3/envs/gammapy-dev/lib/python3.7/site-packages/astropy/io/fits/hdu/hdulist.py in __getitem__(self, key)
    318         try:
    319             return self._try_while_unread_hdus(super().__getitem__,
--> 320                                                self._positive_index_of(key))
    321         except IndexError as e:
    322             # Raise a more helpful IndexError if the file was not fully read.

~/software/miniconda3/envs/gammapy-dev/lib/python3.7/site-packages/astropy/io/fits/hdu/hdulist.py in _positive_index_of(self, key)
    775         """
    776 
--> 777         index = self.index_of(key)
    778 
    779         if index >= 0:

~/software/miniconda3/envs/gammapy-dev/lib/python3.7/site-packages/astropy/io/fits/hdu/hdulist.py in index_of(self, key)
    756 
    757         if (found is None):
--> 758             raise KeyError(f'Extension {key!r} not found.')
    759         else:
    760             return found

KeyError: "Extension 'EFFECTIVE AREA' not found."
```

if this worked properly we could replace the function I introduced in observation just with `HDULocation`.
My function searches for the `HDUCLAS4` first in the header keywords, that is unique; then gets the `extname` and reads the IRF.


@AtreyeeS, concerning IRF components in different files, why would you use this function at that point? Wouldn't it be easier to use the normal `Observation` initialiser and read the IRF components with their own classes from the different files of interest?

maxnoe:
IRFs shouldn't be searched/identified by extname at all. The standard does not prescribe those and there are only some recommendations. 

The `HDUCLAS*` header is all we need. 

Having extnames defined in gammapy makes sense as a default for writing,  but reading should not rely on them

maxnoe:
This snippet shows how to map the `HDUCLAS*` to the correct gammapy class, it is also easy to invert for the opposite direction, if you are looking for a specific class in a file:

```python
from gammapy.irf import EffectiveAreaTable2D, EnergyDispersion2D, PSF3D, Background2D, Background3D, RadMax2D
from gammapy.data import EventList, GTI, PointingInfo, ObservationTable, HDUIndexTable
from astropy.io import fits

import warnings



# hierarchy of HDUCLASX to gammapy classes
GADF_HDUS = {
    'EVENTS': EventList,
    'GTI': GTI,
    'POINTING': PointingInfo,
    'INDEX': {
        'OBS': ObservationTable,
        'HDU': HDUIndexTable,
    },
    'RESPONSE': {
        'EFF_AREA': {
            'POINT-LIKE': {
                'AEFF_2D': EffectiveAreaTable2D,
            },
            'FULL-ENCLOSURE': {
                'AEFF_2D': EffectiveAreaTable2D,
            },
        },
        'EDISP': {
            'POINT-LIKE': {
                'EDISP_2D': EnergyDispersion2D,
            },
            'FULL-ENCLOSURE': {
                'EDISP_2D': EnergyDispersion2D,
            },
        },
        'RPSF': {
            'POINT-LIKE': {
                'PSF_TABLE': PSF3D,
            },
            'FULL-ENCLOSURE': {
                'PSF_TABLE': PSF3D,
            }
        },
        'BKG': {
            'POINT-LIKE': {
                'BKG_2D': Background2D,
                'BKG_3D': Background3D,
            },
            'FULL-ENCLOSURE': {
                'BKG_2D': Background2D,
                'BKG_3D': Background3D,
            }
        },
        'RAD_MAX': {
            'POINT-LIKE': RadMax2D,
        }
    }
}


def find_cls_for_hdu(hdu, hierarchy, level=1):
    hduclas = hdu.header.get(f'HDUCLAS{level}')
    if hduclas is None:
        return None

    if hduclas not in hierarchy:
        warnings.warn(f'Unknown HDUCLAS{level}: "{hduclas}"')
        return None

    result = hierarchy.get(hduclas)
    if isinstance(result, dict):
        return find_cls_for_hdu(hdu, result, level=level + 1)
    else:
        return result


def load_gadf_hdus(path):
    hdus = []
    with fits.open(path) as hdu_list:
        for hdu in hdu_list:
            cls = find_cls_for_hdu(hdu, GADF_HDUS)
            if cls is None:
                continue

            hdus.append(cls.from_table(Table.read(hdu)))
    return hdus
```


I don't know where this fits best? Should I open a PR for gammapy io (without RAD_MAX for now?)
Or just add it here?

cosimoNigro:
Thanks @maxnoe, 

I don't think this solves the duplication problem pointed out by @adonath. 
The thing is, discussing with @registerrier today, that the `.tag` attribute of each IRF it's the same string of the `HDUCLAS4` (see for example the [EffectiveArea2D](https://github.com/gammapy/gammapy/blob/master/gammapy/irf/psf/parametric.py#L359) and [PSFKing](https://github.com/gammapy/gammapy/blob/master/gammapy/irf/psf/parametric.py#L359), so there is already a 1:1 connection between the GADF specification and the `Gammapy` implementation.
 
Can we get the class from the tag? Without incurring in the error I pointed out in the `HDULocation` - i.e. without using `extname`s?

maxnoe:
@cosimoNigro It was not meant to duplicate, but rather augment / replace.

You need to indices: one for finding the right class given the HDUCLAS* objects, one for finding the right HDU when looking for a specific class.

cosimoNigro:
@adonath, finally I understood how to use the `IRFRegistry` and I removed the dictionary with the class implementation I added. I think there should be no duplication now. @adonath, @maxnoe, can you take a second look and tell me if you are satisfied now?

adonath (changes requested):
Thanks for the improvements @cosimoNigro. I have left a few more comments for simplification. By making full use of the `HDULocation` class I think you even simplify further. Here is a what I mean:
```python3
from astropy.io import fits
from gammapy.data import HDUIndexTable
from gammapy.utils.fits import HDULocation

filename = make_path("$GAMMAPY_DATA/hess-dl3-dr1/data/hess_dl3_dr1_obs_id_020136.fits.gz")

hdulist = fits.open(make_path(filename))

kwargs = {}

for hdu in hdulist:
    hdu_class = hdu.header.get("HDUCLAS1", "").lower()
    
    if hdu_class == "response":
        hdu_class = hdu.header.get("HDUCLAS4", "").lower()
    
    loc = HDULocation(
        hdu_class=hdu_class,
        hdu_name=hdu.name,
        file_dir=filename.parent,
        file_name=filename.name
        
    )
    for name in HDUIndexTable.VALID_HDU_TYPE:
        if name in hdu_class:
            data = loc.load()
            # TODO: maybe introduce IRF.type attribute...
            kwargs[name] = data
```

In addition this PR definitely needs a test to execute the functionality once. In general I think the "searching for first valid HDU" should go to `gammapy.irf`. But this could be done in a follow up PR.

cosimoNigro:
Hello @adonath, thanks for the patience.

I slightly modified your snippet to check only for IRF components (we want eventually to allow them to be in different files than the event list). Then I reformatted the docstrings and added a minimal test.

adonath (changes requested):
Thanks a lot @cosimoNigro! I think the implementation is very clean now and re-uses as much existing functionality as possible. I have two last comment on removing the default value handling and moving some of the functionality to `gammapy.irf.io`, otherwise the PR is ready to merge from my side.

gammapy/data/observations.py
```diff
@@ -342,6 +389,62 @@ def select_time(self, time_interval):
...
+        observation : `~gammapy.data.Observation` 
+            observation with the events and the irf read from the file
+        """
+        events = EventList.read(event_file)
+        
+        gti = GTI.read(event_file) 
+
+        irf_file = irf_file if irf_file is not None else event_file
+        irf_dict = load_irf_dict_from_file(irf_file)
+
```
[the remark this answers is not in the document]
cosimoNigro: simplified

gammapy/data/observations.py
```diff
@@ -6,17 +6,64 @@
...
 from .filters import ObservationFilter
 from .gti import GTI
 from .pointing import FixedPointingInfo
+from .hdu_index_table import HDUIndexTable
 
 __all__ = ["Observation", "Observations"]
 
 log = logging.getLogger(__name__)
 
+def load_irf_dict_from_file(filename):
```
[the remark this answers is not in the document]
cosimoNigro: moved

cosimoNigro:
All done Maestro.

cosimoNigro:
I don't understand why this is failing...
https://github.com/gammapy/gammapy/pull/3514/checks?check_run_id=3650421966#step:6:107
:thinking:

adonath:
@cosimoNigro The failure is a circular import. To avoid it just use a "delayed" import here. This means move the `from gammapy.irf.io import load_irf_dict_from_file` directly to the `Observation.read()` method.

cosimoNigro:
I fixed it, all tests are passing now.

adonath:
Thanks @cosimoNigro! Any final comments @registerrier?

gammapy/data/tests/test_observations.py
```diff
@@ -228,6 +228,15 @@ def test_observation():
     assert_allclose(obs.muoneff, 1)
 
 
+@requires_data()
+def test_observation_read():
```
maxnoe, line 232: There is no test for the case of a separate irf file

gammapy/irf/io.py
```diff
@@ -124,3 +129,49 @@ def load_cta_irfs(filename):
...
+            hdu_class = hdu.header.get("HDUCLAS4", "").lower()
+        
+            loc = HDULocation(
+                hdu_class=hdu_class,
+                hdu_name=hdu.name,
+                file_dir=filename.parent,
+                file_name=filename.name
+            )
+            
+            for name in HDUIndexTable.VALID_HDU_TYPE:
```
maxnoe, line 169: Why do we need to loop here? Why not just check `if hdu_class in HDUIndexTable.VALID_HDU_TYPE`?

gammapy/irf/io.py
```diff
@@ -124,3 +129,49 @@ def load_cta_irfs(filename):
...
+                if name in hdu_class:
```
maxnoe, line 170: Are you using `in` here to compare strings? Why not `==`? Also see the above comment.

maxnoe (changes requested):

maxnoe:
Also, @AtreyeeS said multiple IRF files should be supported

gammapy/irf/io.py
maxnoe: Also, the `VALID_HDU_TYPE` and `VALID_HDU_CLASS` should probably be sets, not lists

gammapy/data/observations.py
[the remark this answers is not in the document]
maxnoe: Is `load_cta_irfs` still needed? It basically does the same but with hard-coded IRF types, which CTA are not CTA specific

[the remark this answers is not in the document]
adonath: No, indeed it's not needed anymore, but it can be removed in a follow up PR as well.

gammapy/irf/io.py
adonath: The problem that is solved here is to assign the class instances to the corresponding observation attributes (and attribute names). That way it works here (basically accidentally, but safe...) is that the hdu class (`bkg_2d`, `aeff_2d` etc.) contains the attribute name. The alternative solution here is something like:
```python3
irf = loc.load()

if instance(irf, PSF):
    kwargs["psf"] = irf
elif instance(irf, EffectiveArea2D):
    kwargs["aeff"] = irf
...
```
Or even better introducing a Gammapy internal`IRF.type` class attribute (see TODO in the code), such that the following works:
```python3
irf = loc.load()
kwargs[irf.type] = irf
```
As I mentioned before for `Map` objects we have a factory pattern, where `Map.read()` returns the given instance of `WcsNDMap`, `HpxNDMap` or `RegionNDMap`. Something like this could work for IRFs as well, like:

```python3
try:
    kwargs["psf"] =  PSF.read() # find first valid HDU here...
except ValueError:
    pass
```
But the latter I'd consider as work for a follow up PR.

adonath:
@AtreyeeS I think adding the support for multiple IRF files leads too far for this convenience functionality. Because at the point, where one declares the files one by one I would rather go for an index file again. I talked to @registerrier yesterday and the only clean solution is probably to declare the corresponding IRF files in the event file header / meta information, as done in the OGIP format. However this is currently not supported by GADF. If you strongly prefer having this kind of flexibility one could maybe support a `dict` for `irf_file` as well...

gammapy/irf/io.py
```diff
@@ -124,3 +129,49 @@ def load_cta_irfs(filename):
...
+    
+    Parameters
+    ----------
+    filename : str, Path
+        path to the file containing the IRF components, if EVENTS and GTI HDUs 
+        are included in the file, they are ignored
+
+    Returns
+    -------
+    irf_dict : dict
```
registerrier, line 146: Maybe "dict of `~gammapy.irf.IRF`"

gammapy/irf/io.py
```diff
@@ -124,3 +129,49 @@ def load_cta_irfs(filename):
...
+        dictionary with instances of the Gammapy obejcts corresponding
```
registerrier, line 147: objects

gammapy/irf/io.py
```diff
@@ -124,3 +129,49 @@ def load_cta_irfs(filename):
     psf = EnergyDependentMultiGaussPSF.read(filename, hdu="POINT SPREAD FUNCTION")
 
     return dict(aeff=aeff, bkg=bkg, edisp=edisp, psf=psf)
+
+
+def load_irf_dict_from_file(filename):
```
registerrier, line 134: An issue with this function is what happens if several IRFs of the same type are stored in the same file, e.g. same run, different event type.
A minima, one should document the behavior and log some warning in case there is a doubt.

registerrier (commented):
Thanks @cosimoNigro .

This looks OK.

My mais concerns with this approach are:
- the assumption that all IRFs reside in one file. 
- the assumption that the irf file only contains one IRF per type. What happens if it is not the case?
In the absence of a defined serialization format for an `Observation` (e.g. either based on the FITS hierarchy conventions or more simply with a yaml file), I think another approach is to store the file names and possibly EXTNAME and EXTVER of the corresponding IRFs.

AtreyeeS:
I did not mean that multiple IRFs *should* be supported. I just wanted to imply that this support is easily provided when loading observations through the index tables, and enquired how it could be handled in case of one fits file with different IRFs in the header. 
Thinking about the scope for this convenience function, I  agree that this is an overkill.

gammapy/irf/io.py
cosimoNigro: added

gammapy/irf/io.py
cosimoNigro: corrected

gammapy/irf/io.py
cosimoNigro: I added a Warning message if two irf with the same HDU class type are detected.
It specifies the `EXTNAME` of the one that has been loaded in the dictionary.

gammapy/data/tests/test_observations.py
cosimoNigro: Added now

cosimoNigro:
I have added the test for `Observation.read` in case of two separate file for events and IRF.
I have added a test on the function `load_irf_dict_from_file`, it is a bit a duplication of the one on `load_cta_irf`, but one of the two can be removed in a follow-up PR eliminating `load_cta_irf`.
I have included a warning in `load_irf_dict_from_file` in case two IRF of the same HDU class type are encountered in the same file.
```python
from gammapy.irf import load_irf_dict_from_file
irf = load_irf_dict_from_file("irf_file_020136_two_psfs.fits")
```
gives
```bash
more than one HDU of psf type found
loaded the PSF HDU in the dictionary
```
@adonath, @registerrier can you take another look? Thanks.  
[irf_file_020136_two_psfs.fits.gz](https://github.com/gammapy/gammapy/files/7233846/irf_file_020136_two_psfs.fits.gz)

cosimoNigro:
Hi @registerrier, @maxnoe,

I added the test on `Observation.read` reading file and IRF components from different DL3 files.
I also added the test catching the warning about two different IRF components with the same hdu class in the function loading the IRF component.

Can you check again?

gammapy/irf/tests/test_io.py
```diff
@@ -35,6 +37,49 @@ def test_cta_irf():
...
+    
+    # create a dummy file with the PSF HDU repeated twice
+    f = fits.open(original_file)
+    f.append(f[5].copy())
+    f[7].name = "PSF2"
+    f.writeto(dummy_file)
+    
+    load_irf_dict_from_file(dummy_file)
+    
+    assert "more than one HDU" in caplog.text
```
registerrier, line 81: OK this could have tested that the log level was warning. Will do it a follow up PR.

registerrier (approved):
Thanks @cosimoNigro ! This looks good to go.

PR #3515, opened 2021-09-16
Add RadMax2d IRF

maxnoe:
**Description**

Implement RadMax2d IRF for point like IRFs with energy dependent RAD_MAX

**Dear reviewer**

### DIFF
--- a/gammapy/irf/tests/test_rad_max.py
+++ b/gammapy/irf/tests/test_rad_max.py
@@ -7,12 +7,12 @@ def test_rad_max_roundtrip(tmp_path):
     from gammapy.irf import RadMax2D
 
     n_energy = 10
-    energy_edges = np.geomspace(100 * u.GeV, 100 * u.TeV, n_energy + 1)
-    energy_axis = MapAxis.from_edges(energy_edges, name='energy')
+    energy_axis = MapAxis.from_energy_bounds(
+        50 * u.GeV, 100 * u.TeV, n_energy, name='energy'
+    )
 
     n_offset = 5
-    offset_edges = np.linspace(0, 2, n_offset + 1) * u.deg
-    offset_axis = MapAxis.from_edges(offset_edges, name="offset")
+    offset_axis = MapAxis.from_bounds(0, 2, n_offset, unit=u.deg, name="offset")
 
     shape = (n_energy, n_offset)
     rad_max = np.linspace(0.1, 0.5, n_energy * n_offset).reshape(shape)


gammapy/irf/rad_max.py
```diff
@@ -0,0 +1,8 @@
+from .core import IRF
+
+
+class RadMax2D(IRF):
+    """small class to exploit the IRF class to directly read in the FITS file the
+    RAD_MAX hud, this has identical axes to the EFFECTIVE AREA HDU"""
```
registerrier, line 6: Please explicitly put the expected arguments in the docstring as well as a link to the gadf specifications as done in the `EffectiveAreaTable`.

gammapy/data/observations.py
```diff
@@ -35,6 +35,9 @@ class Observation:
         Point spread function
     bkg : `~gammapy.irf.Background3D`
         Background rate model
+    rad_max: `~gammapy.irf.RadMax2D` or `~astropy.units.Quantity`
+        Only for point-like IRFs: RAD_MAX table (energy dependent RAD_MAX)
+        or a single angle (global RAD_MAX)
```
registerrier, line 40: Will this work with either a `RadMax2D` object and a single quantity? 

I would expect that at some point one would need to use the `RadMax2D.evaluate()` method.  This won't work with a `Quantity` object.

gammapy/irf/tests/test_rad_max.py
```diff
@@ -0,0 +1,33 @@
...
+def test_rad_max_roundtrip(tmp_path):
+    from gammapy.irf import RadMax2D
+
+    n_energy = 10
+    energy_edges = np.geomspace(100 * u.GeV, 100 * u.TeV, n_energy + 1)
+    energy_axis = MapAxis.from_edges(energy_edges, name='energy')
+
+    n_offset = 5
+    offset_edges = np.linspace(0, 2, n_offset + 1) * u.deg
+    offset_axis = MapAxis.from_edges(offset_edges, name="offset")
```
registerrier, line 15: You can use the `MapAxis.from_bounds` directly:
`offset_axis = MapAxis.from_bounds(0*u.deg, 2*u.deg, n_offset, name="offset", interp="lin")`

gammapy/irf/tests/test_rad_max.py
```diff
@@ -0,0 +1,33 @@
...
+import astropy.units as u
+from gammapy.maps import MapAxis
+
+
+def test_rad_max_roundtrip(tmp_path):
+    from gammapy.irf import RadMax2D
+
+    n_energy = 10
+    energy_edges = np.geomspace(100 * u.GeV, 100 * u.TeV, n_energy + 1)
+    energy_axis = MapAxis.from_edges(energy_edges, name='energy')
```
registerrier, line 11: Here you should rather use `MapAxis.from_energy_bounds` since by default `from_edges` will create a `lin` interpolation (and centers won't be properly calculated).
`energy_axis = MapAxis.from_energy_bounds(100*u.GeV, 100*u.TeV, n_energy, unit="TeV")`

gammapy/data/hdu_index_table.py
```diff
@@ -17,7 +17,7 @@ class HDUIndexTable(Table):
     See :ref:`gadf:hdu-index`.
     """
 
-    VALID_HDU_TYPE = ["events", "gti", "aeff", "edisp", "psf", "bkg"]
+    VALID_HDU_TYPE = ["events", "gti", "aeff", "edisp", "psf", "bkg", "rad_max"]
```
registerrier, line 20: Technically, we don't have different format specifications for the HDU index table, so I suppose it is fine to add `rad_max` even though it does not appear in the v0.2 format specification. 
At some point we will need to introduce different conventions for the valid HDU types. But this will be for later PRs.

gammapy/data/observations.py
registerrier: In particular, in the current scheme there is no HDU  for the RAD_MAX in 1D. It is not clear to me how one can identify that it exists.
What would happen if you were to open one of joint Crab data release?

registerrier (changes requested):
Thanks @maxnoe ! This looks good.

Simply add arguments information to the RadMax2D docstring and a reference to the relevant page in the gadf.

I have one general question regarding the expected behavior with the simple RadMax. I don't clearly see how the scheme here can apply with a `Quantity` defined as a header keyword compared to a full `IRF` with a specific entry in the HDU table. A minima, I suppose one needs to check for its existence when creating the `Observation`. What would be the plan here?

gammapy/data/observations.py
maxnoe: @registerrier I don't understand the direction you are going with the second question. AFAIK none of the open crab data sets have a RAD MAX 2D table, so `rad_max` will be `None` for those.

gammapy/data/hdu_index_table.py
maxnoe: As this is a backwards compatible addition, I don't see why this should be excluded. Also I would argue that the omission of RAD MAX for the index table hdu in 0.2 of the spec was a bug in the spec, as all other HDU types were listed.

gammapy/data/observations.py
maxnoe: > In particular, in the current scheme there is no HDU for the RAD_MAX in 1D.

I don't understand this sentence. What do you mean with "1D" in here?

gammapy/irf/tests/test_rad_max.py
maxnoe: done

gammapy/irf/tests/test_rad_max.py
maxnoe: done

PR #3516, opened 2021-09-17
Add support for systematic errors and weights in FluxPointDataset

QRemy:
Preliminary draft PR to follow-up this morning discussion on systematic errors and weights in FluxPointDataset.
I tried to add systematic errors as a model that is bounded to a dataset (somehow like the BackgroundModel for MapDataset), and weights such as the datasets will minimise |sum(chi2)/sum(dof)-1| instead of chi2 (so the optimal syst. errors fitted won't tend to infinity).

### DIFF
--- a/gammapy/datasets/core.py
+++ b/gammapy/datasets/core.py
@@ -69,7 +69,7 @@ class Dataset(abc.ABC):
 
         return np.sum(stat, dtype=np.float64)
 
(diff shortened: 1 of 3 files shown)


QRemy:
Closing this for now, we will have to re-discuss how to deal with weighted likelihood in a more general way.

Issue #3493, opened 2021-09-02
Success flag and FluxMaps/FluxPoints


registerrier:
The success flag is no longer present in the `FluxPoints`. It can be important to track errors.

Should we add a boolean `success` quantity in `FluxMaps`? Or should we put e.g. `nan` values where optimization/confidence has failed?

Opinions @adonath , @QRemy , @bkhelifi @AtreyeeS ?

AtreyeeS:
I vote for `success` flags rather than `nan`

bkhelifi:
I have a preference for a NaN, as one could not check the content of all bins in a FluxMap. In the plot function, this will screw up everything, which is great as it will tell to the user that something went wrong. But maybe one could add an argument to remove the non-converged bins...

One could add a boolean flag additionally, such that the user can understand the reason of this NaN. One might have NaN coming from other reasons (or maybe not).

PS: it is sad that the documentation of `FluxEstimator` is not visible, as the code to compute flux in `FluxPointsEstimator` is actually in `FluxEstimator`
PS2: the same for the documentation of `FluxMaps` that is not visible...
PS2: this is an issue with the documentation of the function sqrt_ts of `FluxPoints`.

QRemy:
I would prefer to keep the bad values rather than using nan as they often contain information on what went wrong by themself (stuck at best fit-value or min/max bounds, errors missing or way too small/large...).

 If we provide the convergence status as a boolean array the users can easily remove them in one line  if they want to: `FluxMap[convergence_failure]=np.nan`. 
 
Moreover we could have one mask for covariance_failure and one array for convergence_failure with different values depending on the success message as convergence_failure = 0:success, 1:estimated distance to minimum (EDM) too small, 2 failure. So we can decide to keep or not the EDM too small case that often correspond to a success.

registerrier:
The good thing with having `nan` is that `Map.plot` will directly plot a blank entry. So that if you do `flux_maps.dnde.plot()` you won't see the pixels without convergence.

This behavior can be achieved with the approach suggested by @QRemy if we modify the various flux properties of `FluxMaps` e.g.:
```
 def _filter_convergence_failure(self, some_map):
     """Put NaN where pixels did not converge."""
     some_map[~self.convergence_success.data] = np.nan
     return some_map

@property
def dnde(self):
"""Return differential flux (dnde) SED values."""
    return self._filter_convergence_failure(self.norm * self.dnde_ref)
```

adonath:
[...]

PR #3518, opened 2021-09-20
Add FluxMaps.fit_status

bkhelifi:
**Description**
This PR proposes an implementation in order to add the feature proposed in #3493, add fit succes maps in the FluxMaps and compute them in the estimators 'ParameterEstimator', 'FluxPointsEstimator' and 'LightCurveEstimator'.

**Dear reviewer**
With this implementation, the 'FluxMaps' are able to contain a Boolean map of estimation success and an integer map containing the estimation success with the following scheme: 0=failed, 1=success of the quantity estimation, 2=success of the estimation of the quantity and its error.
These information are propagated from the 'ParameterEstimator' to the high-level estimators. We can note that I have not made a differentiation between all types of errors (e.g. error in the estimation of ul, errn-errp, err). Here, if there is one issue during the computation of any error of a parameter, a 'Flase' flag is propagated.
During the scan of a parameter (over dataset or GTI), if one issue occurs, all the scanning estimation is set as faulty. This could be improved.
In this implementation, a NaN is added when a parameter (its error) has not being properly estimated. 

This initial PR is a draft for discussion. The associated test functions need to be implemented after validation of the feature(s). The realised tests were on the tutorial notebook 'detect.ipynb' and 'light_curve.ipynb' (with and w/o the reoptimisation option), and all the tests of the estimators module are ok (after one adaptation).
Beyond the implementation that might deserve some clever writing, one should decide:
- [ ] Have only the Boolean map, only the fit_status, or both
- [ ] Use of NaN when there is an error

In this PR, we could also decide whether the 'BrentqFluxEstimator' should be moved from ts_map.py to flux.py.

### DIFF
--- a/gammapy/scripts/download.py
+++ b/gammapy/scripts/download.py
@@ -65,8 +65,8 @@ class DownloadIndex:
 
 def progress_download(source, destination):
     """
-    Note
-    ----
+    Notes
+    -----
     The progress bar can be displayed for this function.
     """
     import requests
--- a/gammapy/utils/parallel.py
+++ b/gammapy/utils/parallel.py
@@ -115,8 +115,8 @@ def run_multiprocessing(
 ):
     """Run function in a loop or in Parallel
 
-    Note
-    ----
+    Notes
+    -----
     The progress bar can be displayed for this function.
 
     Parameters


gammapy/estimators/parameter.py
```diff
@@ -211,10 +221,17 @@ def estimate_scan(self, datasets, parameter):
             parameter=parameter,
             reoptimize=self.reoptimize
         )
+        success = True
+        for res in profile["fit_results"]:
+            success &= res
```
QRemy, line 226: This seems equivalent to `success = False not in profile["fit_results"]`
and redondant with : https://github.com/gammapy/gammapy/blob/832eb734cb6ac7c3d9902bfa454e323fc445a8e6/gammapy/estimators/parameter.py#L366

gammapy/estimators/parameter.py
```diff
@@ -330,4 +358,19 @@ def run(self, datasets, parameter):
                 result.update(self.estimate_scan(datasets, parameter))
 
         result.update(self.estimate_counts(datasets))
+
+        result["fit_status"] = int(result["success"])
+        if result["fit_status"] > 0 and "success_err" in result:
+            result["fit_status"] += int(result["success_err"])
+            if "scan" in self.selection_optional and "success_scan" in result:
+                if False in result["success_scan"]:
+                    result["fit_status"] = 1
```
QRemy, line 367: see comment above, if I get it correctly `success_err` is False `if False in result["success_scan"]` so this is unnecessary.

gammapy/modeling/tests/test_fit.py
```diff
@@ -192,9 +192,10 @@ def test_stat_profile_reoptimize():
 
     assert_allclose(result["x_scan"], [0, 2, 4], atol=1e-7)
     assert_allclose(result["stat_scan"], [4, 0, 4], atol=1e-7)
-    assert_allclose(
-        result["fit_results"][0].total_stat, result["stat_scan"][0], atol=1e-7
-    )
+    # assert_allclose(
+    #     result["fit_results"][0].total_stat, result["stat_scan"][0], atol=1e-7
+    # )
```
QRemy, line 197: is this failing ?

gammapy/modeling/tests/test_fit.py
```diff
@@ -249,9 +250,9 @@ def test_stat_surface_reoptimize():
     ]
 
     assert_allclose(list(result["stat_scan"]), expected_stat, atol=1e-7)
-    assert_allclose(
-        result["fit_results"][0][0].total_stat, result["stat_scan"][0][0], atol=1e-7
-    )
+    # assert_allclose(
+    #     result["fit_results"][0][0].total_stat, result["stat_scan"][0][0], atol=1e-7
+    # )
```
QRemy, line 255: is this failing ?

QRemy (commented):
Thanks @bkhelifi this looks good.  I'm not sure to understand the different success checks related to the `estimate_scan`, but I think it could be simplified. Besides that I mostly left comment about lines that should be uncommented or removed.

I also wonder if fit_status should not rather take into account the different status related to the optimize_result message and introduce a separated map for the error_status.

gammapy/estimators/parameter.py
```diff
@@ -211,10 +221,17 @@ def estimate_scan(self, datasets, parameter):
             parameter=parameter,
             reoptimize=self.reoptimize
         )
+        success = True
+        for res in profile["fit_results"]:
+            success &= res
+            if res is False:
+                log.warning(f"Failed to make the parameter scan for {parameter.name}")
```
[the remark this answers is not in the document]
bkhelifi: This is a good question! I have no idea how often it fails... and this was totally hidden. As people prefer ...

gammapy/estimators/parameter.py
```diff
@@ -82,17 +82,20 @@ def estimate_best_fit(self, datasets, parameter):
                 * parameter.name_err: covariance-based error estimate on parameter value
         """
         value, total_stat, success, error = np.nan, 0, False, np.nan
+        # value, total_stat, success, message, error = np.nan, 0, False, "", np.nan
```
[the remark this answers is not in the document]
bkhelifi: I can propagate all the messages! But should we have a map of messages or log.warning?

gammapy/estimators/parameter.py
```diff
@@ -330,4 +358,19 @@ def run(self, datasets, parameter):
...
+                if False in result["success_scan"]:
+                    result["fit_status"] = 1
+        if "success_err" in result:
+            del result["success_err"]
+        if "success_scan" in result:
+            del result["success_scan"]
+
+        # Temporary DEBUG for this draft
+        if result["fit_status"] > 2:
+            print("\nBKH> LOGIC ERROR fit_status=", result["fit_status"])
```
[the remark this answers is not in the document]
bkhelifi: of course!!!!

gammapy/estimators/parameter.py
bkhelifi: Not exactly.... in parameter.py, one bin can fail. 
then, in the 'run' function I concatenate the fit_status for all estimations (e.g. scan, ul, errn-errp)

gammapy/modeling/tests/test_fit.py
bkhelifi: Yes, because I have overloaded the function assessing to a parameter in order to return a Nan when the estimation fails. Which was the case here!! This is why this PR is usefull

gammapy/estimators/parameter.py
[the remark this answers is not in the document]
QRemy: not as a log.warning for sure, a map of messages could be useful but complicated, so maybe just uncommenting the message in the output of the estimate_... method could be enough to help debugging for now.

QRemy:
A `full_ouptut` option could be introduced such as it return only `fit_status` if False (by default) and otherwise each succes_... maps for the different estimates (ul, err, errp-errn, scan...).

https://github.com/gammapy/gammapy/blob/832eb734cb6ac7c3d9902bfa454e323fc445a8e6/gammapy/estimators/parameter.py#L348
Lines after this ones could be simplified with something like:

```python
fit_status = int(result["success"])
other_status = True
selected = mandatory + selection_optional
for method_name, method in methods_dict.items():
    if method_name in selected:
        result.update(method(...))
        other_status &= result[f”succes_{method_name}”]
        if not full_output:
            del result[f”succes_{method_name}”]
fit_status += int(fit_status*other_status)
```

gammapy/estimators/flux_map.py
```diff
@@ -338,32 +349,36 @@ def npred_excess(self):
 
     @property
     def stat_scan(self):
-        """Fit statistic value"""
+        """Scanned values of fit statistic"""
```
registerrier, line 352: Fit statistic scan values

gammapy/estimators/flux_map.py
```diff
@@ -385,41 +400,41 @@ def sqrt_ts(self):
...
         """
         if "sqrt_ts" in self._data:
-            return self._data["sqrt_ts"]
+            return self._filter_convergence_failure(self._data["sqrt_ts"])
         else:
             with np.errstate(invalid="ignore", divide="ignore"):
                 ts = np.clip(self.ts.data, 0, None)
                 data = np.where(self.norm > 0, np.sqrt(ts), -np.sqrt(ts))
-                return Map.from_geom(geom=self.geom, data=data)
+                return self._filter_convergence_failure(Map.from_geom(geom=self.geom, data=data))
```
registerrier, line 408: here the filtering has been already applied when doing `self.ts.data` no?

gammapy/estimators/flux_map.py
```diff
@@ -455,102 +470,119 @@ def eflux_ref(self):
     @property
     def dnde(self):
         """Return differential flux (dnde) SED values."""
-        return self.norm * self.dnde_ref
+        return self._filter_convergence_failure(self.norm * self.dnde_ref)
```
registerrier, line 473: if it has been filtered for `norm` it is useless to redo it for `flux` quantities

gammapy/estimators/flux_map.py
```diff
@@ -455,102 +470,119 @@ def eflux_ref(self):
...
+        """Fit statistic value"""
+        self._check_quantity("fit_status")
+        return self._data["fit_status"]
+
+    def _filter_convergence_failure(self, some_map, status=1):
+        """Put NaN where pixels did not converge."""
+        if not self.has_fit_status:
+            return some_map
+        # For integer content, NaN is not defined
+        if not np.asanyarray(some_map.data).dtype is np.dtype(float):
```
registerrier, line 581: This line is hard to understand. Is it useful? All fitted quantities are float no?

Also `Map` has a `dtype` attribute directly.

bkhelifi: There are not. A test function returns one integer, and not an ndarray of float. This is why I use np.asanyarray

QRemy: If it appears only in one test function, it would be simpler to adapt it rather than the base code

gammapy/estimators/flux_map.py
```diff
@@ -291,13 +300,15 @@ def energy_max(self):
     def niter(self):
```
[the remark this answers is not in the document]
bkhelifi: Very good point.
I have made the unification between 'niter' and 'ncalls': 'ncalls' is then the only parameter whatever the method of parameter estimation

gammapy/estimators/flux_map.py
bkhelifi: done

gammapy/estimators/flux_map.py
bkhelifi: very good point, thanks

gammapy/estimators/flux_map.py
bkhelifi: This has been fixed along the whole code

gammapy/modeling/tests/test_fit.py
bkhelifi: The object result["fit_results"] contains now the status of the fit (a Boolean), and not the full object 'result = self.optimize'

bkhelifi:
This is not anymore a conflict, but an add-on

gammapy/estimators/parameter.py
```diff
@@ -7,6 +7,13 @@
 
 log = logging.getLogger(__name__)
 
+AVAILABLE_METHODS = [
```
registerrier, line 10: Do we really want this? The list of methods is a common thing for estimators.

gammapy/estimators/ts_map.py
```diff
@@ -773,6 +847,8 @@ def _ts_value(position, counts, exposure, background, kernel, norm, flux_estimat
     norm : `~numpy.ndarray`
         Norm image. The flux value at the given pixel position is used as
         starting value for the minimization.
+    name : str
+        Name of the associated Dataset (optional)
```
registerrier, line 851: Why do you add this? Is it needed at all?

registerrier (changes requested):
Thanks @bkhelifi . 

This looks good but I am afraid that the PR is changing too many at once and becomes complex to review. In particular some aspects require some design discussions as they touch broader part of the problem.

I have some reservations with specific choices:
- I don't think `ncalls` or `niter` are relevant informations to keep in the `estimate` data model.  It was present only in the `TSMapEstimator` result and I think never used. This is rather debug information and could/should be delivered through `log.DEBUG` output to the user. Opinions @adonath @QRemy ?
- The `fit_status` here becomes a very general structure. It might need an individual PR by itself. For easier maintenance and documentation, it could become an `Enum` based class, e.g.:
```
class FitStatus(Enum):
    FAILED = 0
    OPTIMIZE = 1
    COVARIANCE = 2
``` 
Should there be a flag per step (optimize, cover, confidence, scan), possibly allowing a combination or something else?
- `success` becomes redundant if there is a `fit_status`. If we decide to go for the status solution 
- Be careful with the handling of the `FluxMap`. The way it is currently written, you might be modifying the data themselves when applying the success filtering. 

All in all, I think this has to be split into several PRs or it will be very difficult to get in. Maybe, first something adding a simple `success` flag to the `FluxMaps` to simply recover the initial missing information and then introduce a more complex solution for the `fit_status`? Opinions @adonath ?

gammapy/estimators/parameter.py
```diff
@@ -310,24 +380,46 @@ def run(self, datasets, parameter):
...
             if "ul" in self.selection_optional:
                 result.update(self.estimate_ul(datasets, parameter))
+                other_success &= result["success_ul"]
 
             if "scan" in self.selection_optional:
                 result.update(self.estimate_scan(datasets, parameter))
+                other_success &= result["success_scan"]
+
+        if result["fit_status"] == 1 and other_success:
+            result["fit_status"] = 2
```
QRemy, line 411: This can go in the loop over AVAILABLE_METHODS if this is given as a dict that link the key to the methods,
as in my previous comment https://github.com/gammapy/gammapy/pull/3518#issuecomment-924066478

gammapy/estimators/ts_map.py
```diff
@@ -637,18 +663,30 @@ def ts_diff(x):
...
+                    ts_diff,
+                    [min_norm],
+                    [max_norm],
+                    nbin=1,
+                    maxiter=self.max_niter,
+                    rtol=self.rtol,
+                )
+            value = (roots[0] - norm) * factor
+            if "converged" not in res[0].flag:
+                value = np.nan
```
QRemy, line 677: according to the previous comment "Where the root finding fails NaN is set as norm" then I guess that roots[0] is already NaN so the extra if is not necessary.

bkhelifi:
@registerrier : Your comment is fully in phase with my PR description. What do we want? All the pieces are here in this draft... One should decide. Flag of what? Bool for what? which maps to add on? Do we put Nan when there are errors?
The ncalls/niter/nfev were present under 3 different names and non-uniformely over estimators. What to do? Should we use only has debug info or to a map (as it happens for some estimators now and not the other)? Here I have proposed an uniformisation between all estimators and a map creation (that is simple to remove).

gammapy/modeling/scipy.py
```diff
@@ -69,23 +69,21 @@ def _confidence_scipy_brentq(
...
     message, success = "Confidence terminated successfully.", True
     kwargs.setdefault("nbin", 1)
 
+    success = True
     roots, res = find_roots(
         ts_diff.fcn, lower_bound=lower_bound, upper_bound=upper_bound, **kwargs
     )
     result = (roots[0], res[0])
-    if np.isnan(roots[0]):
+    if np.isnan(roots[0]) or "converged" not in res[0].flag:
```
QRemy, line 86: same as above, roots[0] should be NaN if not converged

gammapy/modeling/scipy.py
```diff
@@ -173,4 +171,7 @@ def f(x):
     roots, res = find_roots(
         f, lower_bound=norm_best_fit, upper_bound=value_scan[-1], nbin=1, **kwargs
     )
-    return roots[0]
+    ul = roots[0]
+    if "converged" not in res[0].flag:
+        ul = np.nan
+    return ul
```
QRemy, line 177: same as above, roots[0] should be NaN if not converged

bkhelifi:
For the fit_status, I have no issue with the proposal of a class creation as internal format. Here, the PR is mainly about the serialisation, the output format in maps. What do we propose for that? an enum can not be used in order to create maps that can be read by anyone w/o gammapy.

gammapy/modeling/scipy.py
bkhelifi: should or are?

gammapy/estimators/ts_map.py
bkhelifi: In order to trace where an error occurs.

gammapy/modeling/scipy.py
QRemy: They are initialized with Nan
https://github.com/gammapy/gammapy/blob/0643bb772c86092f758efad745f248a517658013/gammapy/utils/roots.py#L117
and replaced with value only if converged:
https://github.com/gammapy/gammapy/blob/0643bb772c86092f758efad745f248a517658013/gammapy/utils/roots.py#L132

gammapy/estimators/flux_map.py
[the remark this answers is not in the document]
adonath: Note that `niter` and `ncalls` is not necessarily the same: some optimizers require multiple function evaluations per iteration. In practice it might be hard to distinguish. However I'm +1 to unify the name. Would you mind @bkhelifi and do this is a separate PR (as @QRemy suggested as well) first?

adonath:
Just as a general comment. I think a good internal model would be to map the individual success flag to an 4 or 8 bit integer, like:

|        | Case 1  | Case 2   | Case 3  | etc.  |
|--------------------|----|----|---|----|
| Fit success        | 1  | 1  | 0 | 1  |
| Covariance success | 1  | 1  | 0 | 1  |
| Confidence success | 1  | 0  | 1 | 1  |
| UL success         | 1  | 0  | 1 | 0  |
| Total fit status   | 15 | 12 | 3 | 14 |

The mapping can be easily done like:
```python3

def success_to_fit_status(success):
    string = "".join([str(int(_)) for _ in success])
    return int(string, 2)

print(success_to_fit_status([True, False, True, True]))
```

adonath:
Here is the ND Numpy version for the fit status conversion:
```python3
import numpy as np

fit_status = np.array([[1, 2, 3], [4, 5, 6]], dtype=np.uint8)

success = np.unpackbits(fit_status).reshape((8,) + fit_status.shape)
print(bools)

fit_status = np.packbits(success).reshape(success.shape[1:])
print(fit_status)
```

adonath:
@bkhelifi Are you interested to continue this work? I think it would be goo to have the `.fit_status` stored somewhere...

adonath:
Just wanted to mention it would be good to have this...

PR #3519, opened 2021-09-20
Allow for neg. ULs in CountsStatistics

adonath:
**Description**

This pull request changes `compute_upper_limits` top allow for negative upper limits. While physically this doesn't make sense, statistically it is a correct solution and points to underestimated systematics. This change is also done for consistency with the `TSMapEstimator` I noticed the difference in https://github.com/adonath/notebooks-public/blob/master/excess-vs-ts-map-estimator.ipynb.

**Dear reviewer**

### DIFF
--- a/docs/estimators/index.rst
+++ b/docs/estimators/index.rst
@@ -104,6 +104,20 @@ n_sigma_ul        Number of sigma used for upper limit estimation
 ts_threshold_ul   TS threshold to define the use of an upper limit
 ================= =================================================
 
+A note on negative flux and upper limit values:
+
+.. note::
+
+    Gammapy allows for negative flux values and upper limits by default.
+    While those values are physically not valid solutions, they are still
+    valid statistically. Negative flux values either hint at overestimated
+    background levels or underestimated systematic errors in general. Or in
+    case of many measurements, such as pixels in a flux map, they are even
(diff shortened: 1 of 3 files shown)


docs/estimators/index.rst
```diff
@@ -104,6 +104,20 @@ n_sigma_ul        Number of sigma used for upper limit estimation
 ts_threshold_ul   TS threshold to define the use of an upper limit
 ================= =================================================
 
+A note on negative flux and upper limit values:
+
+.. note::
+
+    Gammapy allows for negative flux values and upper limits by default.
```
registerrier, line 111: Technically, this depends on the allowed parameter range, no? 
If you force, `amplitude` and  `norm` to be positive quantities the UL will be positive as well right?

registerrier (dismissed):
Thanks @adonath .

My main comment is regarding is the explanation text which might not be fully true since by defining parameter range the allowed upper bound of the interval will change.

docs/estimators/index.rst
adonath: Indeed I forgot about the recent change, that allows taking into account the boundaries. I'll adapt the text accordingly.

registerrier (approved):
Thanks @adonath !

PR #3521, opened 2021-09-22
Add sample_time to model.temporal.ExpDecayTemporalModel

fabiopintore:
The `sample_time` function is generally needed for event sampling of variable sources. This PR aims at introducing such a function into the `~gammapy.model.temporal.ExpDecayTemporalModel`, where it is currently missing. 
The new function is based on the `numpy.random.Generator.exponential`, which allows to generate random numbers following an exponential function.


registerrier (changes requested):
Thanks @fabiopintore . This PR duplicates #3522 because of an issue with branches I suppose. Maybe close this one and add the changes to the other one?

PR #3522, opened 2021-09-22
Add sample_time to model.temporal.GaussianTemporalModel

fabiopintore:
The `sample_time` function is generally needed for event sampling of variable sources. This PR aims at introducing such a function into the ~gammapy.model.temporal.GaussianTemporalModel, where it is currently missing.
The new function is based on the numpy.random.Generator.normal, which allows to generate random numbers following a normal distribution.

### DIFF
--- a/gammapy/modeling/models/tests/test_temporal.py
+++ b/gammapy/modeling/models/tests/test_temporal.py
@@ -174,7 +174,6 @@ def test_exponential_temporal_model_sample_time():
 
     assert_allclose(times.value, [55555.00000787, 55555.0000118], rtol=1e-8)
 
-
 def test_gaussian_temporal_model_evaluate():
     t = Time(46301, format="mjd")
     t_ref = 46300 * u.d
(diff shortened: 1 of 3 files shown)


gammapy/modeling/models/tests/test_temporal.py
```diff
@@ -187,6 +194,14 @@ def test_gaussian_temporal_model_integral():
     assert_allclose(np.sum(val), 0.682679, rtol=1e-5)
 
 
+def test_gaussian_temporal_model_sample_time():
+    t_ref = Time(55555, format="mjd")
+    temporal_model = GaussianTemporalModel(t_ref=t_ref.mjd * u.d, sigma="2.0 day")
+    times = temporal_model.sample_time(2)
+
+    assert_allclose(times.value, [55555.00000291, 55554.99999694], rtol=1e-8)
```
registerrier, line 202: the difference in the test suggest that the sigma is not 2 days. Maybe an issue with the unit propagation in `sample_time`?

registerrier (approved):
Thanks @fabiopintore . Indeed these functions are missing to be able to sample events.

It seems you made the branch out of the first branch. This PR contains the same modification as those in #3521 . Since they touch the same file it might not be needed to have 2 distinct PRs. Maybe close the other one.

Also there seems to be an issue with the handling of units.

PR #3525, opened 2021-09-23
Add example of DataStore.from_events_files in CTA notebook

QRemy:
This PR copy the usage example of  `DataStore.from_events_files()`
from its docstring  to the CTA notebook. This illustrates an alternative solution if ones can't get the index files  (received report about download link broken and bad file received from someone else...).

### DIFF
--- a/docs/tutorials/data/cta.ipynb
+++ b/docs/tutorials/data/cta.ipynb
@@ -139,7 +139,28 @@
    "outputs": [],
    "source": [
     "data_store = DataStore.from_dir(\"$GAMMAPY_DATA/cta-1dc/index/gps\")\n",
-    "data_store"
+    "print(data_store)"
+   ]
+  },
+  {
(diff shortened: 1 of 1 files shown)


QRemy:
Works locally but fails in the tests because of $CALDB ?
 `No such file or directory: '$CALDB/data/cta/1dc/bcf/South_z20_50h/irf_file.fits'`

registerrier (approved):

Maybe at some point we will have to add the new CTA IRFs BTW.

QRemy:
This notebook focus on CTA 1DC that uses the old IRFFs. Maybe we can add in a separate PR the latest IRFs to gammapy-data and create a how to about "getting the CTA IRFs in gammapy"

registerrier:
We have to see with CTAO whether we can add them to gammapy-data. But a minima, an howto is indeed a good idea.
I merge this PR now.

PR #3528, opened 2021-09-24
Change tutorials to use Models.to_parameters_table() method.

registerrier:
**Description**

This pull request modifies tutorials to make the use `Models.to_parameters_table()` rather than `fit_result.parameters.to_table()`.

This method is barely used in our documentation but it has the advantage of connecting parameters to their parent models which many users require when fitting multi-component models. 

It goes also in the direction of having users interact mostly with `Datasets` and `Models`.  

**Dear reviewer**


AtreyeeS (approved):
Thanks @registerrier !

PR #3531, opened 2021-09-28
Small bug fixes in ImageProfileEstimator

AtreyeeS:
This pull request makes small fixes in `ImageProfileEstimator`

- Fix axis label for plot
- Fix error message in case of wrong inputs

**Dear reviewer**

Probably this class will evolve with the new estimator API, but this is a small change that might still be required.


registerrier (approved):
Thanks @AtreyeeS . This looks good!

PR #3535, opened 2021-10-05
Add TemplateNDSpectralModel

QRemy:
This PR adds a `TemplateNDSpectralModel`, a model spectral generated from a `RegionNDMap` where extra dimensions define the parameters space. The goal is to be able to explore a pre-computed models grid (from complex simulations for example)

The doc example page is still missing, first I have to find a good use case and add the fits file in gammapy-data.
Similarly I could introduce a `TemplateNDSpatialModel` in a separate PR.


gammapy/modeling/models/spectral.py
```diff
@@ -1487,6 +1514,105 @@ def from_region_map(cls, map, **kwargs):
...
+        return self._map
+
+    def evaluate(self, energy, **kwargs):
+        lon = self.map.geom.center_skydir.data.lon.deg
+        lat = self.map.geom.center_skydir.data.lat.deg
+        coord = {"lon": lon, "lat": lat, "energy_true": energy}
+        coord.update(kwargs)
+
+        pixels = list(self.map.geom.coord_to_pix(coord))
+        # TODO : pixels in lon/lat are nan at center coord, why ?
```
adonath, line 1582: The usual behavior is that the lon / lat coordinate are `NaN` if the position is not contained in the region...this should be the case for many regions, but it's not guaranteed for compound regions or more complex shapes such as annuli or similar.

gammapy/modeling/models/spectral.py
```diff
@@ -1533,6 +1536,105 @@ def from_region_map(cls, map, **kwargs):
...
+        val = self.map.interp_by_pix(pixels, **self._interp_kwargs)
+        return u.Quantity(val, self.map.unit, copy=False)
+
+        super().__init__()
+
+    def write(self, overwrite=False):
+        if self.filename is None:
+            raise IOError("Missing filename")
+        elif os.path.isfile(self.filename) and not overwrite:
+            log.warning("Template file already exits, and overwrite is False")
```
registerrier, line 1617: Maybe raise an `IOError` here. Or is it to avoid exceptions when exporting `Models` with template files?
In the latter case, maybe adding a test would be useful.

gammapy/modeling/models/spectral.py
```diff
@@ -1533,6 +1536,105 @@ def from_region_map(cls, map, **kwargs):
...
+        Meta information, meta['filename'] will be used for serialization
+    interp_kwargs : dict
+        Interpolation keyword arguments passed to `gammapy.maps.Map.interp_by_coord`.
+        Default arguments are {'interp': 'linear', 'fill_value': 0}.
+    """
+
+    tag = ["TemplateNDSpectralModel", "templateND"]
+
+    def __init__(self, map, interp_kwargs=None, meta=None, filename=None):
+        self._map = map.copy()
```
registerrier, line 1558: Maybe one should check that an energy axis is present.

gammapy/modeling/models/spectral.py
```diff
@@ -1533,6 +1536,105 @@ def from_region_map(cls, map, **kwargs):
...
+                    value=center,
+                    unit=axe.unit,
+                    scale_method="scale10",
+                    min=axe.bounds[0],
+                    max=axe.bounds[-1],
+                    interp="lin",
+                )
+                points_scale.append("lin")
+                parameters.append(parameter)
+        self.default_parameters = Parameters(parameters)
```
registerrier, line 1584: I suppose an `amplitude` parameter is required. It probably needs to be added to the list of tabulated parameters.

registerrier (changes requested):
Thanks @QRemy ! This looks good, but I have a general comment.

Our `SpectralModel` are usually either regular spectra with an `amplitude` parameter (having flux dimension, an additive model) or a `NormSpectralModel` with a `norm` parameter (dimensionless, a multiplicative model). 
Here the `Model` can be both, or it can have neither `amplitude` nor `norm` parameter.
This has to be clarified.
Possibly by testing the existence of an `amplitude` or `norm` parameter or by creating two distinct types of models.

gammapy/modeling/models/spectral.py
QRemy: > Thanks @QRemy ! This looks good, but I have a general comment.
> 
> Our `SpectralModel` are usually either regular spectra with an `amplitude` parameter (having flux dimension, an additive model) or a `NormSpectralModel` with a `norm` parameter (dimensionless, a multiplicative model). Here the `Model` can be both, or it can have neither `amplitude` nor `norm` parameter. This has to be clarified. Possibly by testing the existence of an `amplitude` or `norm` parameter or by creating two distinct types of models.

The behaviour is the same for the `TemplateSpectralModel`, it has no `amplitude` or `norm` parameter and technically it could be two distinct types of models (normed or not).

QRemy:
> Thanks @QRemy ! This looks good, but I have a general comment.
> 
> Our `SpectralModel` are usually either regular spectra with an `amplitude` parameter (having flux dimension, an additive model) or a `NormSpectralModel` with a `norm` parameter (dimensionless, a multiplicative model). Here the `Model` can be both, or it can have neither `amplitude` nor `norm` parameter. This has to be clarified. Possibly by testing the existence of an `amplitude` or `norm` parameter or by creating two distinct types of models.

The behaviour is the same for the `TemplateSpectralModel`, it has no `amplitude` or `norm` parameter, so technically it could be two distinct types of models (normed or not).

gammapy/modeling/models/spectral.py
QRemy: if `os.path.isfile(self.filename) and not overwrite` then `.write` will return an error but I prefer to avoid

gammapy/modeling/models/spectral.py
QRemy: The behaviour is the same for the `TemplateSpectralModel`, it has no amplitude or norm parameter

gammapy/modeling/models/spectral.py
```diff
@@ -1533,6 +1536,105 @@ def from_region_map(cls, map, **kwargs):
...
+
+        points_scale = [
+            "lin",
+            "lin",
+            "log",
+        ]
+        parameters = []
+        for axe in map.geom.axes:
+            if axe.name not in ["energy_true", "energy"]:
+                center = (axe.bounds[1] + axe.bounds[0]) / 2
```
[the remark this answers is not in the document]
QRemy: `axis.center[0]` is the center of the first bin, that's too close of the min bound. I think the center of the interval is more appropriate.

gammapy/modeling/models/spectral.py
```diff
@@ -1533,6 +1536,105 @@ def from_region_map(cls, map, **kwargs):
...
+
+        pixels = list(self.map.geom.coord_to_pix(coord))
+        # TODO : pixels in lon/lat are nan at center coord, why ?
+        pixels[0][np.isnan(pixels[0])] = 0
+        pixels[1][np.isnan(pixels[1])] = 0
+
+        val = self.map.interp_by_pix(pixels, **self._interp_kwargs)
+        return u.Quantity(val, self.map.unit, copy=False)
+
+        super().__init__()
```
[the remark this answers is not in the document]
QRemy: removed

gammapy/modeling/models/spectral.py
QRemy: added

gammapy/modeling/models/spectral.py
```diff
@@ -1533,6 +1536,105 @@ def from_region_map(cls, map, **kwargs):
...
+        """Template map  (`~gammapy.maps.RegionNDMap`)"""
+        return self._map
+
+    def evaluate(self, energy, **kwargs):
+        lon = self.map.geom.center_skydir.data.lon.deg
+        lat = self.map.geom.center_skydir.data.lat.deg
+        coord = {"lon": lon, "lat": lat, "energy_true": energy}
+        coord.update(kwargs)
+
+        pixels = list(self.map.geom.coord_to_pix(coord))
```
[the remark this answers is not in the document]
QRemy: simplified

registerrier (approved):

We will have to see how to document this.

PR #3488, opened 2021-08-31
Add a Datasets API notebook

AtreyeeS:
This PR adds an API notebook for `Datasets`
I think there is sufficient room for improvement, so suggestions much welcome @registerrier @QRemy @luca-giunti @adonath @LauraOlivera @bkhelifi 
I think the `FluxPoints` can be explained in a dedicated `estimators` notebook, and we can link to that.

PR #3536, opened 2021-10-05
Cleanup table rendering

AtreyeeS:
The table created in #3488 is rendering horribly on the webpage. Fixing that


registerrier (approved):
Thanks @AtreyeeS ! 

This looks much clearer with this syntax. I have left one inline comment for the missing `exposure`

Issue #3533, opened 2021-10-01
Support astropy quantities for Map.data


maxnoe:
**Is your feature request related to a problem? Please describe.**

Creating map complains if you pass an `astropy.units.Quantity` as 

**Describe the solution you'd like**

Map should accept a quantity

maxnoe:
The docstrings of `IRF.data` even state that a `Quantity` is possible, only to raise an error five lines further down:

```
    @data.setter
    def data(self, value):
        """Set data

        Parameters
        ----------
        value : `~astropy.units.Quantity`, array-like
            Data array
        """
        required_shape = self.axes.shape

        if np.isscalar(value):
            value = value * np.ones(required_shape)

        if isinstance(value, u.Quantity):
            raise TypeError("Map data must be a Numpy array. Set unit separately")

```

maxnoe:
@QRemy I think I have a simple fix here.

maxnoe:
@QRemy yes.

A side note: the unit should be a class property, right? All IRFs have their prescribed units.

PR #3538, opened 2021-10-06
Minor change on IRF.data and .quantity setters

QRemy:
Solve #3533
-fix docstring on IRF.data setter
-add unit equivalence check on quantity setter


adonath (approved):

adonath:
Can you please do one last review @registerrier?

adonath:
@registerrier Indeed I missed the overwriting the unit on init should work (like `Quantity` works). But we can implement this in a follow up PR.

Issue #3505, opened 2021-09-09
Cryptic error message when FluxPointsEstimator reaches empty bin


registerrier:
In the current dev, when you pass `energy_edges` that are not fully contained in a single energy axis bin, the FluxPointsEstimator provide an INFO message about a dataset that does not contribute in the energy range and fail. The error message is weird because it relates to the model:
```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-26-5d8a7a5c9b26> in <module>
      1 fpe = FluxPointsEstimator(energy_edges=energy_edges, source="crab", selection_optional="all")
----> 2 flux_points = fpe.run(datasets=datasets)

~/Code/gammapy-dev/gammapy/gammapy/estimators/flux_point.py in run(self, datasets)
    656         ):
    657             row = self.estimate_flux_point(
--> 658                 datasets, energy_min=energy_min, energy_max=energy_max,
    659             )
    660             rows.append(row)

~/Code/gammapy-dev/gammapy/gammapy/estimators/flux_point.py in estimate_flux_point(self, datasets, energy_min, energy_max)
    695 
    696         datasets_sliced.models = datasets.models.copy()
--> 697         return super().run(datasets=datasets_sliced)

~/Code/gammapy-dev/gammapy/gammapy/estimators/flux.py in run(self, datasets)
    129         models = datasets.models.copy()
    130 
--> 131         model = self.get_scale_model(models)
    132 
    133         energy_min, energy_max = datasets.energy_ranges

~/Code/gammapy-dev/gammapy/gammapy/estimators/flux.py in get_scale_model(self, models)
     93             Scale spectral model
     94         """
---> 95         ref_model = models[self.source].spectral_model
     96         scale_model = ScaleSpectralModel(ref_model)
     97 

~/Code/gammapy-dev/gammapy/gammapy/modeling/models/core.py in __getitem__(self, key)
    531             return self.__class__(list(np.array(self._models)[key]))
    532         else:
--> 533             return self._models[self.index(key)]
    534 
    535     def index(self, key):

~/Code/gammapy-dev/gammapy/gammapy/modeling/models/core.py in index(self, key)
    537             return key
    538         elif isinstance(key, str):
--> 539             return self.names.index(key)
    540         elif isinstance(key, Model):
    541             return self._models.index(key)

ValueError: 'crab' is not in list
```

It is difficult for a user to figure out what is going-on in the context. 
[...]

PR #3543, opened 2021-10-08
Correct FluxPointsEstimator to handle empty datasets slices

registerrier:
**Description**

This pull request proposes a correction for issue #3505 . 

Most of the difficulty comes from the fact that, because of the inheritance, information is lost when going to the various `.run()` methods. Since only the lowest level `ParameterEstimator.run()` can compute the resulting dictionary, if it fails at a higher level (e.g. here with the impossibility to find the reference model on an empty `Datasets`), no result dict can be returned. 
This PR re-introduces a `nan_result` method that will produce a `dict` filled with nans for all expected entries. This is non ideal as any change at the lower level will require to update there as well.

At least a test is added to ensure that these issues do not go under the radar.

Opinions @adonath ?

**Dear reviewer**


adonath:
The conclusion here was to try and internal API, where the `datasets` and `models` are passed in to estimators separately. Did you have time to try this @registerrier? Or do you want me to take this over?

registerrier:
> The conclusion here was to try and internal API, where the `datasets` and `models` are passed in to estimators separately. Did you have time to try this @registerrier? Or do you want me to take this over?

Thanks @adonath . I am currently updating the PR to pass `models` along `datasets` in `FluxEstimator` and `ParameterEstimator` as this is a blocker for hess validation. This should be ready before the next dev call.

registerrier:
> The conclusion here was to try and internal API, where the `datasets` and `models` are passed in to estimators separately.

Trying to implement a solution where we pass `Models` down to the `ParameterEstimator` we still have the issue that the `FluxEstimator` has to compute a reference flux for the energy range, which it can't do if an empty `Datasets` is passed to `run`. It is therefore probably unavoidable to compute part of the result dictionary on the highest level.
Otherwise one has to pass the energy interval to `run`.

adonath:
@registerrier I resolved the merge conflict and will merge now.

PR #3544, opened 2021-10-12
Add FluxMaps.success quantity

bkhelifi:
**Description**
This pull request is associated to the Issue #3493 and a simplified version of the obsolete PR #3518.
Here, the success map with a boolean dtype is added in FluxMap. Small bug fixes have to be made in order to deal properly with the dtype of maps.

**Dear reviewer**
This PR is an advanced draft in order to add this feature. Some improvements could be made, even if I have not found a best ways to conserve the dtype of maps within some functions or constructor.
There is one test on the cta_sensitivity that fails and it should not. I have not understand why and help is asked...


bkhelifi:
After having be helped, the source of the cast of boolean maps has been found and fixed. 
to_region_nd_map should be properly called within flux_map, and within wcs/ndmap::to_region_nd_map a cast should be done as the function interp_by_coord does not preserve the dtype

gammapy/estimators/flux_map.py
```diff
@@ -136,6 +143,10 @@ def __init__(self, data, reference_model, meta=None, gti=None):
 
         self.meta = meta
         self.gti = gti
+        self.remove_failed = remove_failed
```
adonath, line 146: Maybe choose a name that is closer to the `success` flag? Such as `filter_success_nan` or similar?

gammapy/estimators/flux_map.py
```diff
@@ -136,6 +143,10 @@ def __init__(self, data, reference_model, meta=None, gti=None):
...
+
+    def set_remove_failed(self, remove_failed):
```
adonath, line 148: Please change to the property pattern like:
```python3

@property
def filter_success(self):
    """""""
    return self._filter_success

@filter_sucess.setter
def filter_success(self, value):
    # check value if needed
    self._filter_success = value


```

gammapy/estimators/flux_map.py
```diff
@@ -552,6 +574,18 @@ def eflux_ul(self):
         """Return energy flux (eflux) SED upper limits."""
         return self.norm_ul * self.eflux_ref
 
+    def _filter_convergence_failure(self, some_map):
+        """Put NaN where pixels did not converge."""
+        if not self.remove_failed:
+            return some_map
+
+        if (not self.has_success) or (not self.success.data.shape == some_map.data.shape):
```
adonath, line 582: The implicit assumption is that the geometries of the maps are compatible in `FluxMaps`, so you could skip the second check here...

adonath (changes requested):
Thanks @bkhelifi, I have left a few minor inline comments, otherwise looks good to me!

adonath:
This might create merge conflicts with #3491, but let's merge this PR first and then I rebase #3491.

gammapy/estimators/flux_map.py
bkhelifi: In the function, flux_point::from_table, there is a look on each individual map. The same occurs for the run function of the lightcurve estimator. During this loop, one map is reduced but not yet the success map. Without this second check, the lign `new_map.data[~self.success.data] = np.nan` will lead to a crash.

adonath (commented):
Thanks @bkhelifi, no further comment from my side...

gammapy/estimators/tests/test_flux_map.py
```diff
@@ -233,6 +240,21 @@ def test_flux_map_properties(wcs_flux_map, reference_model):
...
+    assert_allclose(fluxmap.flux_err.data[:, 0, 1], [np.nan, 9e-14])
+
+    assert_allclose(fluxmap.eflux.data[:, 0, 1], [np.nan, 2.30258509e-12])
+    assert_allclose(fluxmap.e2dnde_err.data[:, 0, 1], [np.nan, 1e-13])
+
+
+def test_flux_map_failed_properties(wcs_flux_map, reference_model):
+    fluxmap = FluxMaps(wcs_flux_map, reference_model)
+    fluxmap.filter_success_nan = False
+
```
registerrier, line 254: I would suggest to add an explicit test for the getter/setter mechanism here, e.g.
```
assert fluxmap.filter_success_nan == False
```

registerrier (approved):
Thanks @bkhelifi . This looks good to go.

Do we prefer `filter_success_nan` over the simpler `filter_success`.
I have one suggestion to add an explicit test for the getter/setter mechanism of `filter_success_nan`. The entry could also be added to the `FluxMaps.__str__()` method.

gammapy/estimators/tests/test_flux_map.py
bkhelifi: done

gammapy/estimators/flux_map.py
adonath: My first thought here is, that the `success` map, **must** have the same dimension and axes as the `norm` map, because each norm pixel is associated to a fit. This is independent of `FluxPoints` or `FluxMaps`. I still haven't understood, why the second would be needed.

adonath: OK, now I think I understand: the failures you see are related to the `counts`, `npred` and `stat_scan` quantities. Those are the quantities, that have a different dimension than `norm`. They either include an additional `norm` or `dataset` axis. In this case the approach of:
```python3
new_map.data[~self.success.data] = np.nan
```
Does not work and raises an error, because the dimension do not agree. The solution here is to rely on Numpy's broadcasting instead, e.g. via multiplication:
```python3
mask_nan = np.ones(success.data.shape)
mask_nan[~success.data] = np.nan
new_map = some_map * mask_nan
```

adonath: On the other hand it might not make sense at all to filter the `counts`, `npred` and `stat_scan` quantities, because those are the quantities one would like to check, when the fit failed...but I guess that's why we introduced the `filter_success_nan` flag.

adonath (changes requested):

bkhelifi:
With the last commits, comments are taken into account and all the test pass.
Let's note that the added complexity in `FluxMap.__filter_convergence_failure` will be simplified afterwards with the PR #3491 by @adonath. We coordinate ourself for this strategy.

adonath (approved):
Thanks @bkhelifi , no further comments from my side.

PR #3545, opened 2021-10-12
Correct bug in integrate_spectrum

registerrier:
**Description**

This pull request corrects a bug in `gammapy.modeling.models.spectral.integrate_spectrum` which cause integral to be 0 if the number of requested bin was larger than 50 per decade. 

This is caused because we rely on `numpy.geomspace(min, max,num, axis=-1)` to produce a 2D array oversampling the input `energy_min` and `energy_max`. If `num` becomes lower than two, `geomspace` returns only one energy edge or, even worse,  an empty array if num=0.
To solve this we force here `num` to be at least 2.

**Dear reviewer**


registerrier:
Merging now.

PR #3561, opened 2021-10-28
Add FluxPoints.is_ul setter and fix its serialisation

QRemy:
Fix FluxPoints.to_table() such as the output table contains meta["sqrt_ts_threshold_ul"] and ` is_ul` column.


adonath (approved):

adonath:
@QRemy There are still test fails: https://github.com/gammapy/gammapy/runs/4206751490?check_suite_focus=true#step:9:2027

Some of them might be introduce by the merge commit that I made. Do you have time to fix those? Otherwise I can do it as well.

QRemy:
4 of the test failures was there before, I think they requires to adapt the reference files in gammapy-data, I will check.

QRemy:
I cannot reproduce the error in `test_flux_point_estimator.py::test_flux_points_estimator_small_edges`.
The other fails should be solved by after merging the PR in gammapy-data.

adonath:
Thanks @QRemy, I'll merge the PR now and do a follow up PR to fix remaining test fails.

PR #3564, opened 2021-11-01
Add unit tests for cta irfs, alpha configuration

fabiopintore:
This PR adds unit tests in `~gammapy.irf` for the new set of CTA IRFS for the alpha configuration.


adonath:
Thanks @fabiopintore! Please fix https://github.com/gammapy/gammapy/runs/4069321465?check_suite_focus=true#step:9:2011

adonath (approved):
Thanks a lot @fabiopintore! At some point we could maybe think about testing IRFs and data files independently, but for now it's fine to keep them in Gammapy.

PR #3537, opened 2021-10-06
Allow quanity for irf data in __init__ and setter, fixes #3533

maxnoe:
**Description**

Allows setting IRF data with a quantity and adds the missing ABCMeta metaclass to the `IRF` baseclass.

PR #3566, opened 2021-11-01
Add missing ABCMeta to IRF base class

maxnoe:
**Description**

This contains the fix for the missing `ABCMeta` of the `IRF` class that was part of the closed without merging #3537.

Before, the interface for subclasses of `IRF` was not enforced.


maxnoe:
@adonath you could test that a dummy IRF class definition throws an error, like this:

```
with pytest.raises(...):
    class FooIrf(IRF):
        pass
    f = FooIrf()
```

adonath:
Thanks @maxnoe, this is probably not worth the effort right now...

PR #3570, opened 2021-11-03
Implement Map broadcasting

adonath:
**Description**

This pull request implements the support for broadcasting when doing map arithmetics. It include the following methods:
- `MapAxis.broadcast()`, which returns the broadcasted axis dimension, i.e. the max of the len of the axis, when either both axes are equal or one of them has exactly one bin.
- `MapAxes.broadcast()`, which follow bumpy broadcasting rules.
- `Geom.broadcast()`, does broadcasting of the extra axes and requires the spatial dimension to be equivalent. 

**Dear reviewer**
Todo:
- Implement broadcasting arithmetics unit tests
- Extend the `maps.ipynb` tutorial

### DIFF
--- a/gammapy/maps/axes.py
+++ b/gammapy/maps/axes.py
@@ -1363,7 +1363,7 @@ class MapAxes(Sequence):
             else:
                 axes.append(ax_other)
 
-        return axes
+        return self.__class__(axes=axes)
 
     @property
     def reverse(self):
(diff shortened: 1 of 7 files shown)


registerrier:
@adonath what is the status here? 
Can this be rebased ? Can it be done for v1.1?

adonath:
@registerrier No, I think this will require more discussion on the actual behavior. E.g. if positions of axes do not align etc. Let's postpone...

registerrier:
Hi @adonath . What is the status here?

QRemy:
The implementation would require more work and there is no strong need for that so closing for now.

Issue #3571, opened 2021-11-03
Printing parameters to table if the first model has a linked parameter


tunbehaun273:
**Gammapy version**
v0.18.2

**Bug description**
KeyError is raised if the first model of a Datasets object has a linked parameter (the to_dict() method includes the 'link' key) but not all parameters have the link key.

**Expected behavior**
The to_table() method should either include the link column for all or no parameter

**To Reproduce**
Link the a parameter of the first model in the Datasets object. Write it to disk and read it again so the link is in the yaml representation. Try datasets.parameters.to_table() and one gets the key error.
**Other information**
Easy work around is to not have the linked model as first model in the datasets.
If this issue is already solved in the developer version this issue can be closed. Just wanted to make you aware in case nobody else noticed.

PR #3573, opened 2021-11-04
Add link info to parameter.to_table()

QRemy:
Add link info to parameter.to_table()
Solves #3571


adonath:
Maybe keep the `link` info the table and just set a placeholder if it is not defined?

adonath (approved):

PR #3574, opened 2021-11-08
Add a few more temporal models

bkhelifi:
**Description**
This pull request associated to the issue #3325 adds 3 temporal models: LinearTemporalModel, PowerLawTemporalModel and SinusTemporalModel.

**Dear reviewer**
One might need to discuss about the exact functions...

PR #3577, opened 2021-11-10
Add serialisation test for temporal models

QRemy:
Follow up of PR #3574
Add serialisation tests for temporal models.

@bkhelifi, note that Model.from_dict() works only for  specific classes.
So if one call model_dict = PowerLawTemporalModel.to_dict()
 then PowerLawTemporalModel.from_dict(model_dict) works but TemporalModel.from_dict(model_dict) create a TemporalModel and not a PowerLawTemporalModel. To prevent this I could raise an error in from_dict if `data['type'] not in cls.tag` (in a separate PR).

### DIFF
--- a/gammapy/modeling/models/tests/test_temporal.py
+++ b/gammapy/modeling/models/tests/test_temporal.py
@@ -7,6 +7,7 @@ from astropy.table import Table
 from astropy.time import Time
 from gammapy.data.gti import GTI
 from gammapy.modeling.models import (
+    TemporalModel,
     ConstantTemporalModel,
     LinearTemporalModel,
     ExpDecayTemporalModel,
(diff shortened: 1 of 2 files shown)


bkhelifi:
If I understand well, one needs to know in advance the type of the model in order to read it correctly? If true, it is not very flexible, is not it?

gammapy/modeling/models/tests/test_temporal.py
```diff
@@ -288,3 +295,21 @@ def test_plot_constant_model():
     constant_model = ConstantTemporalModel(const=1)
     with mpl_plot_check():
         constant_model.plot(time_range)
+
+
+def test_io():
```
adonath, line 300: Minor: maybe move to `gammapy/modeling/models/test_io.py` instead? This where the other I/O tests are...

adonath (approved):
Thanks @QRemy, I have left one minor comment on the test location. Otherwise looks good to me...

gammapy/modeling/models/tests/test_temporal.py
QRemy: done

adonath (approved):
Thanks @QRemy!

PR #3581, opened 2021-11-11
Adapt npred excess definition

adonath:
**Description**

This pull request adapts the current definition of `npred_excess`, based on the finding outlined in https://github.com/adonath/notebooks-public/blob/master/npred_excess_definition.ipynb . The conclusion of the notebook is that in order to get consistent excess definitions in the case of `WStatCountsStatistics` and `SpectrumDatasetOnOff`, one needs to define excess as the excess in case of the **best fit hypothesis**, while currently it is defined as the difference between the best fit npred and the  null npred . In case of an on-off statistics this is not the same, because the estimated background level is not the same for both hypothesis.

This PR corrects the definition of `npred_excess` according to the proposal in the notebook. For now I have decided to remove the `npred_null` completely. One could keep it as additional info, but it is basically not very relevant information.

This change will lead to consistent results between the `FluxProfileEstimator` (proposed in #3480) and the current `ExcessProfileEstimator`, basically making the `ExcessProfileEstimator` obsolete.

### DIFF
--- a/docs/estimators/index.rst
+++ b/docs/estimators/index.rst
@@ -53,8 +53,7 @@ stat_null         Fit statistics value of the null hypothesis
 ts                Difference in fit statistics (`stat - stat_null` )
 sqrt_ts           Square root of ts time sign(norm), in case of one degree of freedom, corresponds to significance (Wilk's theorem)
 npred             Predicted counts of the best fit hypothesis, equivalent to correlated counts for backward folding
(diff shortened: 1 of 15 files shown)


adonath:
@registerrier I'll go ahead and merge this now. I think technically there is no doubt, that this solution yields the consistent results. However let's talk again about the naming and whether to include `npred_null` next Monday.

PR #3585, opened 2021-11-14
Reintroduce iminuit covariance caching

adonath:
**Description**

This pull request re-introduces the caching of the iminuit covariance matrix. @QRemy Can you pleas quickly verify this leads to the desired improvement in performance?


QRemy:
Thanks @adonath, I confirm that the extra covariance computation no longer appears in the analysis_3d_joint benchmark.

adonath:
Thanks @QRemy! There are changes in the test values, but I guess we can live with those as they are rather small...

PR #3588, opened 2021-11-15
Add unit equivalency check to MapAxis.from_energy methods

### LINKED ISSUE #3575: MapAxis.from_energy_bounds does not necessarily return axis with energy unit
**Gammapy version**
dev

**Bug description**
Calling `MapAxis.from_energy_edges` and `MapAxis.from_energy_bounds` without `unit` argument and with regular floats will return an axis without unit. You can also create an axis with any unit actually.

**Expected behavior**
Only axes with energy equivalent units should be returned. An energy axis is not only the name, but also the energy.

So we could set a default unit to TeV. `Quantity.__init__` would then fail if units are not consistent. 

We could also add a `is_equivalent` test.

**To Reproduce**
axis = MapAxis.from_energy_bounds(0.1,10,2, unit = "deg")
axis = MapAxis.from_energy_bounds(0.1,10,2)

**Other information**
Any other information you think will be useful for us to fix the issue can go here.

adonath:
**Description**

This pull request fixes #3575 by adding an `unit.is_equivalent` check to `MapAxis.from_energy_bounds` and `MapAxis.from_energy_edges`.

### DIFF
--- a/gammapy/maps/axes.py
+++ b/gammapy/maps/axes.py
@@ -419,9 +419,8 @@ class MapAxis:
         """
         energy_edges = u.Quantity(energy_edges, unit)
 
-        if unit is None:
-            unit = energy_edges.unit
-            energy_edges = energy_edges.to(unit)
+        if not energy_edges.unit.is_equivalent("TeV"):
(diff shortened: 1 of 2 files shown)


adonath:
CI fails are unrelated.

Issue #3476, opened 2021-08-18
Return simple data types in MapDataset.info_dict


adonath:
I would be useful to return simple data types in `MapDataset.info_dict`. By simple I mean data types that can be serialised to YAML by default or possible with the `JsonQuantityEncoder`. Currently `.info_dict()` returns sometimes `np.array` for single numbers.

adonath:
Addressed in #3589.

PR #3589, opened 2021-11-16
Rerun simple data types in MapDataset.info_dict()

adonath:
**Description**

This pull request fixes issue #3476, by returning only simple data types and verifying the dict is json serializable.


adonath:
CI fails are unrelated.

PR #3590, opened 2021-11-16
Restore "scale10" method as default parameter scaling

QRemy:
This PR changes the default parameter scaling to "scale10" (as previously) instead of `None`. This fix a drop of performance by about 40 % in the `profile_analysis_3d_joint` benchmark. Without scaling  for lon/lat parameters the number of evaluator `.update()` calls required increase by a factor of 6.


adonath (approved):
Thanks a lot @QRemy, no further comments from my side. The CI fail is unrelated...

PR #3591, opened 2021-11-16
Correct FluxPoints upper limit plotting

registerrier:
**Description**

This pull request solves a bug with UL plotting in `FluxPoints`. 
The check in `FluxPoints.plot()` `isnan(is_ul).all()` should read `not isnan(is_ul).all()`: if any value in not nan, you want to plot the UL.

**Dear reviewer**


adonath (approved):

PR #3592, opened 2021-11-16
Avoid copy in background model evaluation

QRemy:
Avoid copy in background model evaluation in order to slightly improve performance.

### DIFF
--- a/gammapy/datasets/map.py
+++ b/gammapy/datasets/map.py
@@ -414,7 +414,7 @@ class MapDataset(Dataset):
             if self._background_cached is None:
                 self._background_cached = background * values
             else:
-                self._background_cached.data = background.data * values.data
+                self._background_cached.data = background.data * values.value
                 self._background_cached.unit = background.unit
             return self._background_cached
         else:
--- a/gammapy/datasets/map.py
+++ b/gammapy/datasets/map.py
@@ -179,6 +179,7 @@ class MapDataset(Dataset):
         self.counts = counts
         self.exposure = exposure
         self.background = background
+        self._background_cached = None
         self.mask_fit = mask_fit
 
         if psf and not isinstance(psf, (PSFMap, HDULocation)):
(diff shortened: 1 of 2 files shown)


gammapy/datasets/map.py
```diff
@@ -408,12 +409,17 @@ def npred_background(self):
             Predicted counts from the background.
         """
         background = self.background
-
         if self.background_model and background:
             values = self.background_model.evaluate_geom(geom=self.background.geom)
-            background = background * values
+            if self._background_cached is None:
+                self._background_cached = background * values
```
adonath, line 415: This caching does not make sense to me. Don't you want to check on the model parameters and avoid the recomputation if not needed?

gammapy/modeling/models/cube.py
```diff
@@ -767,8 +768,9 @@ def evaluate(self):
             Background evaluated on the Map
         """
         value = self.spectral_model(self.energy_center).value
-        back_values = self.map.data * value
-        return self.map.copy(data=back_values)
+        self._evaluation_cache.data = self.map.data * value
```
adonath, line 771: I'd say we don't need the caching on the `.evaluate()` level, this is only used for independent evaluation and not used during fitting.

QRemy: for the `TemplateNPredModel` the evaluator call `.evaluate()` 
https://github.com/gammapy/gammapy/blob/19da7840b81b88092d42cd5ea032a7f9a9b7276f/gammapy/datasets/evaluator.py#L361

adonath: Yes, but we have caching on the `MapEvaluator`, so the effect on the performance should be completely negligible.

gammapy/datasets/map.py
QRemy: yes that could be added for now I just did that as most of the time is spend in copy during `background = background * values` just avoiding this allow to reduce npred_background computation by about 40%.

gammapy/modeling/models/cube.py
QRemy: True

QRemy: removed

gammapy/datasets/map.py
QRemy: added

adonath:
Thanks @QRemy, do you have a rough estimate how much this improve the performance in our benchmarks?

QRemy:
> Thanks @QRemy, do you have a rough estimate how much this improve the performance in our benchmarks?

I tested it only on the `analysis_3d_joint` benchmark, it reduces `npred_background` computation time by about 50% (34s to 17s on my machine)

adonath (approved):

PR #3595, opened 2021-11-16
Workaround for empty column in HDUIndexTable

registerrier:
**Description**

This pull request is changing the `HDUIndexTable.read` to use a `filled` version of the table . This is made necessary because the veritas dataset from the joint-crab paper has an empty `FILE_DIR` column which astropy now transforms into a `MaskColumn`. To avoid this the `read` method will convert the `table` to its filled version using the empty string as the fill value.

**Dear reviewer**


adonath (approved):
Thanks @registerrier!

PR #3597, opened 2021-11-17
Correct confidence level used for contours

registerrier:
**Description**

This pull request corrects the confidence level calculation for contours. 

There is general question one can ask now that `mncontours` takes a confidence level expressed as probability rather than number of sigma. Should we stick with `sigma` as the argument for this method compared to `cl` as done in iminuit?
Opinions @adonath @QRemy @bkhelifi ?

**Dear reviewer**


adonath:
Thanks @registerrier! I guess out tests are not sensitive to this change, because we only use `sigma = 1`? In general we expose `n_sigma` as parameter everywhere, so this we definitely keep, but I agree we could change the signature of the lower level `confidence_iminuit` method.

adonath (approved):
Thanks @registerrier!

PR #3602, opened 2021-11-17
Update affiliation of FP

fabiopintore:
This PR is to update the affiliation of @fabiopintore  in `CITATION.cff`


adonath (approved):
Thanks @fabiopintore!

PR #3620, opened 2021-11-19
Add FoVBackgroundModel and TemplateNPredModel to referenced models

QRemy:
Add FoVBackgroundModel and TemplateNPredModel to referenced models so they will appears in the doc.


adonath (commented):
Thanks @QRemy!

PR #3621, opened 2021-11-19
Drop support for Python 3.6

adonath:
**Description**

With the most recent CI build fails (e.g. https://github.com/gammapy/gammapy/runs/4263760636?check_suite_focus=true) and official support ending Dec 2021 (https://endoflife.date/python) we decided to drop support for Python 3.6.


adonath:
CI passes...

Issue #3555, opened 2021-10-22
OnOff datasets npred computation fails in case no counts is present


AtreyeeS:
**Gammapy version**
present dev, 0.18.2

**Bug description**
 As pointed out by @thierrystolarczyk in the slack channel, `SpectrumDatasetOnOff.npred()` fails in case no counts are present, because in on off datasets we try to compute the marginalised background.

https://github.com/gammapy/gammapy/blob/926d3dae2354859b5bfb008e9f8c2e59fee5d9f4/gammapy/datasets/map.py#L1868

This is annoying as for a simulated dataset (before simulating the counts), `print(dataset_onoff)` or `fake` fails. The only way to get around this is to add a dummy counts map.

**To Reproduce**

In the spectrum simulation notebook, add 
```

ds_onoff = SpectrumDatasetOnOff.from_spectrum_dataset(
                                            dataset=dataset,
                                            acceptance=1,
                                            acceptance_off = 1/0.2)
print(ds_onoff)
```

after cell 10.

**Expected behavior**
What should be the correct behaviour @registerrier @QRemy @adonath ?

adonath:
I think this should be fixed before tomorrow...anybody?

registerrier:
I suppose this means that `counts` should never be `None` in counts based  `Dataset` objects. 

I propose to change `MapDatasetMaker.run` to add an empty `counts`  attribute if `"counts"` is not in the selection.

PR #3623, opened 2021-11-19
Add default zero counts in MapDatasetMaker.run

registerrier:
**Description**

This pull request provides a solution to #3555 . 
In general having no counts entry in a `MapDataset` or `SpectrumDataset` is bad. With ONOff statistic it will lead to errors as in the above issue. 

This situation usually happens when `MapDatasetMaker` is run without the `"counts"` selection applied (usually to make simulations). Here we change the behavior by adding an empty `counts` to the `MapDataset` if `"counts"` is not selected. 

**Dear reviewer**


adonath (approved):

PR #3633, opened 2021-11-20
Expression not assigned

DimitriPapadopoulos:
**Description**

Fix DeepSource.io alert introduced by #3610 / 0538675:

> An expression that is not a function call is assigned to nothing. Probably something else was intended here. We recommend to review this.


adonath (approved):
Thanks!

PR #3634, opened 2021-11-20
Decorate methods with `@staticmethod`

DimitriPapadopoulos:
**Description**

Fixes these DeepSource.io alerts:

> The method doesn't use its bound instance. Decorate this method with `@staticmethod` decorator, so that Python does not have to instantiate a bound method for every instance of this class thereby saving memory and computation. Read more about staticmethods [here](https://docs.python.org/3/library/functions.html#staticmethod).

I have not fixed tests in this merge request.


adonath (approved):
Thanks @DimitriPapadopoulos, those are fair changes.

PR #3635, opened 2021-11-20
Use literal syntax to create dict

DimitriPapadopoulos:
**Description**

Fixes DeepSource.io alerts:
> Using the literal syntax can give minor performance bumps compared to using function calls to create `dict`, `list` and `tuple`.
> [...]
> This is because here, the name `dict` must be looked up in the global scope in case it has been rebound. Same goes for the other two types `list()` and `tuple()`.

This is debatable: one of the advantages of Python is readability, and using `dict()` instead of `{}` might be more readable, especially for empty dictionaries.


adonath (approved):
Thanks!

PR #3636, opened 2021-11-20
Avoid mutable default values for keyword arguments

DimitriPapadopoulos:
**Description**

Address a DeepSource.io alert:
> Do not use a mutable like `list` or `dictionary` as a default value to an argument. Python’s default arguments are evaluated once when the function is defined. Using a mutable default argument and mutating it will mutate that object for all future calls to the function as well.

In these cases, mutables are not mutated, so it's not an actual issue. Still, avoid this anti-pattern - see [Python Mutable Defaults Are The Source of All Evil](https://florimond.dev/en/posts/2018/08/python-mutable-defaults-are-the-source-of-all-evil/).

### DIFF
--- a/gammapy/irf/psf/core.py
+++ b/gammapy/irf/psf/core.py
@@ -71,8 +71,8 @@ class PSF(IRF):
 
     def info(
         self,
-        fraction=[0.68, 0.95],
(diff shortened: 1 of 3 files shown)


adonath (approved):
Thanks!

PR #3637, opened 2021-11-20
Unexpected keyword argument 'psf_map' in constructor call

DimitriPapadopoulos:
**Description**

Fixes a DeepSource.io alert:
> The function call has passed a keyword argument that doesn't correspond to one of the function's parameter names. This is an error.

### DIFF
--- a/gammapy/irf/core.py
+++ b/gammapy/irf/core.py
@@ -627,7 +627,7 @@ class IRFMap:
             exposure_map = Map.from_geom(
                 geom=geom_exposure, data=table["Exposure"].data, unit="cm2 s"
             )
-            return cls(psf_map=psf_map, exposure_map=exposure_map)
+            return cls(irf_map=psf_map, exposure_map=exposure_map)
         else:
             raise ValueError(f"Format {format} not supported")
 


DimitriPapadopoulos:
Mmmh... Both DeepSource.io and myself are missing something here!

DimitriPapadopoulos:
I see, there's a design issue here:
```
IRFMap
└── PSFMap
```
The base class `IRFMap.__init__()` and derived class `PSFMap.__init__()` have different signatures:
https://github.com/gammapy/gammapy/blob/ae5621e7ea0857a8b7dd22df2487165f066934fb/gammapy/irf/core.py#L495
https://github.com/gammapy/gammapy/blob/ae5621e7ea0857a8b7dd22df2487165f066934fb/gammapy/irf/psf/map.py#L70

Now `from_hdulist()` attempts to call the constructor with the signature of `PSFMap.__init__()` instead of the signature of `IRFMap.__init__()`:
https://github.com/gammapy/gammapy/blob/ae5621e7ea0857a8b7dd22df2487165f066934fb/gammapy/irf/core.py#L630

This will work only if we call the class method as `PSFMap.from_hdulist()` instead of `IRFMap.from_hdulist()`.

I think I'll open an issue instead.

PR #3641, opened 2021-11-20
Iterate over the dictionary instead of its keys

DimitriPapadopoulos:
**Description**

This fixes a DeepSource.io alert:

> Consider iterating the dictionary directly instead of calling `.keys()`. Using `for key in dictionary` would always iterate the dictionary keys.


adonath (approved):
Thanks!

PR #3644, opened 2021-11-20
Use `... != ...` instead of `not ... == ...`

DimitriPapadopoulos:
**Description**

Simplifies codes and fixes DeepSource.io alert:
> The boolean expression contains an unneeded negation and can be re-written without using the `not` statement.


adonath (approved):
Thanks!

PR #3645, opened 2021-11-20
Static method with `self` as first argument

DimitriPapadopoulos:
**Description**

Fixes DeepSource.io alerts:
> It is customary for instance or class methods to take `self` or `cls`, respectively, as their first arguments, a method that uses either of these names but is found to be a static method may have been defined incorrectly. Choose names other than `self` or `cls` for arguments to avoid confusing other programmers looking at your code.


adonath (approved):
Thanks @DimitriPapadopoulos!

PR #3646, opened 2021-11-20
Introduce global FitResult object

adonath:
**Description**

This pull request introduce a global `FitResult` object, which just contains the individual results form the covariance and optimize steps. The motivation is a clearer API and future convenience, such as `FitResult.read()`, `FitResult.write()` or `FitResult.status` or `FitResult.check()`.

This is partly a backwards incompatible change, because we previously returned an `OptimizeResult` in `Fit.run()`, which probably was never a good idea. For now I repeated the most common quantities, such as `.total_stat` or `.parameters`. Other quantities such as `.trace` need to be accessed like `FitResult.optimize_result.trace`. Should we fully break the API or repeat all optimize quantities on the `FitResult` to stay backwards compatible?

### DIFF
--- a/gammapy/analysis/tests/test_analysis.py
+++ b/gammapy/analysis/tests/test_analysis.py
@@ -299,7 +299,7 @@ def test_analysis_1d_stacked_no_fit_range():
 
     assert len(analysis.datasets) == 1
     assert_allclose(analysis.datasets["stacked"].counts.data.sum(), 184)
-    pars = analysis.fit_result.parameters
+    pars = analysis.models.parameters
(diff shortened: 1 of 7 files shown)


registerrier (commented):
It is not totally clear why the `models` should be copied on the `FitResult`. 
Won't it create new issues if there are two distinct models?
Also why changing the tests if the `parameters` are still present on the `FitResult`?
 
I am not to see what is gained by reverting to the approach where optimize and covariance results are present on the `FitResult`. Covariance is never really independent of an optimization. You have to be at a minimum to compute the error matrix.

PR #3653, opened 2021-11-23
Change download test to dev notebooks

adonath:
**Description**

This pull request changes the notebooks download test to download the dev version notebooks, which currently fails. 

The issue seems to point in this direction https://stackoverflow.com/q/45027246.

**Dear reviewer**


Bultako:
Continues on #3658

PR #3657, opened 2021-11-23
Fix link to Binder in tutorials

Bultako:
This PR fixes links to Binder notebooks in produced HTML tutorials, due to modifs in tutorials tree-folder structure that had not been taken into account until now.


Bultako:
After modifs done #3658, this PR does not apply.

PR #3658, opened 2021-11-24
Fix untar downloaded tutorials and change test

Bultako:
This PR is a continuation of #3653.

It fixes the way how downloaded tar bundled tutorials are extracted and modifies the notebooks download test to check if the dev version of produced tar bundlle is properly produced, downloaded and uncompressed.


adonath (approved):
Thanks @Bultako, no further comments from my side.

Issue #3664, opened 2021-11-25
Add GeneralizedGaussianTemporalModel


aaguasca:
There is no generalized Gaussian temporal model on gammapy.

I have made one, but the code could be improved. In particular, the way I have defined the parameters that contain time units.

I have recently started to use Github. Is the procedure to open a pull request the following? 
1. Create a branch and apply the changes to it. 
2. Make a pull request from this branch to the main branch.

Any help is welcome!

PR #3669, opened 2021-11-30
Add GeneralizedGaussianTemporalModel

aaguasca:
**Description**
This pull request addresses issue #3664. 
I made a generalized Gaussian temporal profile to describe the temporal component of an AGN flare, BeXRB, etc.
Changes: Add the class "GeneralizedGaussianTemporalModel" to temporal.py file and add the class name to __init__.py file.

**Dear reviewer**
I coded the class following the style of other temporal profiles. The code works and could be merged. The code style could improve, but I do not know how to do it. Is it required? I would like some feedback.

### DIFF
--- a/gammapy/modeling/models/temporal.py
+++ b/gammapy/modeling/models/temporal.py
@@ -364,7 +364,7 @@ class GeneralizedGaussianTemporalModel(TemporalModel):
     eta = Parameter("eta", 1/2, unit = "", frozen=False)
 
     @staticmethod
-    def evaluate(time, t_ref, t_rise, eta, t_decay):
+    def evaluate(time, t_ref, t_rise, t_decay, eta):
         val_rise = np.exp( - 0.5 * (np.abs((time - t_ref).value) ** (1/eta)) / (t_rise.to_value("d") ** (1/eta)))
         val_decay = np.exp( - 0.5 * (np.abs((time - t_ref).value) ** (1/eta)) / (t_decay.to_value("d") ** (1/eta)))
         val = np.where(time < t_ref, val_rise, val_decay)
--- a/gammapy/modeling/models/temporal.py
+++ b/gammapy/modeling/models/temporal.py
@@ -334,6 +334,93 @@ class GaussianTemporalModel(TemporalModel):
         return integral / self.time_sum(t_min, t_max)
 
 
+class GeneralizedGaussianTemporalModel(TemporalModel):
+    r"""A generalized Gaussian temporal profile
+
+    ..math::
+            F(t) = A * exp( - (\frac{|t - t_{maxim}|}{\sigma_r}) ^ v)   for  t < t_max
+            
+            F(t) = A * exp( - (\frac{|t - t_{maxim}|}{\sigma_d}) ^ v)   for  t > t_max            
+
+    Parameters
+    ----------
+    A : `~astropy.units.Quantity`
+        Maximum intensity
+    t_maxim: `~astropy.units.Quantity`
+        The time of the pulse's maximum intensity.
+    sigma_r : `~astropy.units.Quantity`
+        Rise time constant.
+    sigma_d : `~astropy.units.Quantity`
+        Decay time constant.
+    v : `~astropy.units.Quantity`
+        Pulse sharpness -> lower values implies a more peaked pulse
+    
+    """
+
+    tag = ["GeneralizedGaussianTemporalModel", "gengauss"]
+
+    _t_max_default = Time("2000-01-01")
+    t_maxim = Parameter("t_maxim", _t_max_default.mjd, unit = "day", frozen=False)
+    sigma_r = Parameter("sigma_r", "1d", frozen=False)
+    sigma_d = Parameter("sigma_d", "1d", frozen=False)
+    v = Parameter("v", 2, unit = "", frozen=False)
+    A = Parameter("A", 1, unit = "", frozen=False)
+
+    @staticmethod
+    def evaluate(time, t_maxim, sigma_r, v, sigma_d, A):
+        val=np.ones(time.shape)
+        for t in range(len(time)):
+            if time[t] < t_maxim:
+                val[t] = (A * np.exp( - (np.abs((time[t] - t_maxim).value) ** v) / (sigma_r.value ** v)))
+            else:
+                val[t] = (A * np.exp( - (np.abs((time[t] - t_maxim).value) ** v) / (sigma_d.value ** v)))
+        return val
+    
+    def integral(self, t_min, t_max, **kwargs):
+        """Evaluate the integrated flux within the given time intervals
+
+        Parameters
+        ----------
+        t_min: `~astropy.time.Time`
+            Start times of observation
+        t_max: `~astropy.time.Time`
+            Stop times of observation
+
+        Returns
+        -------
+        norm : float
+            Integrated flux norm on the given time intervals
+        """
+        
+        pars = self.parameters
+        sigma_r = pars["sigma_r"].quantity.to_value("day")
+        sigma_d = pars["sigma_d"].quantity.to_value("day")
+        v = pars["v"].quantity
+        A = pars["A"].quantity
+        t_maxim = Time(pars["t_maxim"].quantity, format = "mjd")
+                     
+        def gen_gauss(time):
+            return A * np.exp( - (np.abs(time - t_maxim.mjd) ** v) / (sigma ** v))
+
+        if t_min < t_maxim and t_max <= t_maxim:
+            sigma = sigma_r
+            integral = scipy.integrate.quad(gen_gauss, t_min.mjd, t_max.mjd)[0]
+            return integral / self.time_sum(t_min, t_max).value
+        
+        elif t_min < t_maxim and t_max > t_maxim:
+            sigma = sigma_r
+            integral = scipy.integrate.quad(gen_gauss, t_min.mjd, t_maxim.mjd)[0]
+            
+            sigma = sigma_d
+            integral = integral + scipy.integrate.quad(gen_gauss, t_maxim.mjd, t_max.mjd)[0]  
+            return integral / self.time_sum(t_min, t_max).value
+        
+        elif t_min >= t_maxim and t_max > t_maxim:
+            sigma = sigma_d
+            integral = scipy.integrate.quad(gen_gauss, t_min.mjd, t_max.mjd)[0]
+            return integral / self.time_sum(t_min, t_max).value
+
+
 class LightCurveTemplateTemporalModel(TemporalModel):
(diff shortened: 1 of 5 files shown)


gammapy/modeling/models/temporal.py
```diff
@@ -334,6 +334,93 @@ def integral(self, t_min, t_max, **kwargs):
...
+        Decay time constant.
+    v : `~astropy.units.Quantity`
+        Pulse sharpness -> lower values implies a more peaked pulse
+    
+    """
+
+    tag = ["GeneralizedGaussianTemporalModel", "gengauss"]
+
+    _t_max_default = Time("2000-01-01")
+    t_maxim = Parameter("t_maxim", _t_max_default.mjd, unit = "day", frozen=False)
```
adonath, line 363: Please rename to `t_ref` for consistency with the `GaussianTemporalModel`

gammapy/modeling/models/temporal.py
```diff
@@ -334,6 +334,93 @@ def integral(self, t_min, t_max, **kwargs):
...
+    sigma_r = Parameter("sigma_r", "1d", frozen=False)
+    sigma_d = Parameter("sigma_d", "1d", frozen=False)
+    v = Parameter("v", 2, unit = "", frozen=False)
+    A = Parameter("A", 1, unit = "", frozen=False)
+
+    @staticmethod
+    def evaluate(time, t_maxim, sigma_r, v, sigma_d, A):
+        val=np.ones(time.shape)
+        for t in range(len(time)):
```
adonath, line 372: Please avoid the `for` loop and use plain nunmpy expressions. One way to achieve the definition of the step function is e.g. to use `np.where(condition, value, else)`

gammapy/modeling/models/temporal.py
```diff
@@ -334,6 +334,93 @@ def integral(self, t_min, t_max, **kwargs):
...
+        """
+        
+        pars = self.parameters
+        sigma_r = pars["sigma_r"].quantity.to_value("day")
+        sigma_d = pars["sigma_d"].quantity.to_value("day")
+        v = pars["v"].quantity
+        A = pars["A"].quantity
+        t_maxim = Time(pars["t_maxim"].quantity, format = "mjd")
+                     
+        def gen_gauss(time):
```
adonath, line 402: Why not call into `.evaluate()` directly or at least use `functools.wrap` to avoid duplicating the model expression.

gammapy/modeling/models/temporal.py
```diff
@@ -334,6 +334,93 @@ def integral(self, t_min, t_max, **kwargs):
...
+            return A * np.exp( - (np.abs(time - t_maxim.mjd) ** v) / (sigma ** v))
+
+        if t_min < t_maxim and t_max <= t_maxim:
+            sigma = sigma_r
+            integral = scipy.integrate.quad(gen_gauss, t_min.mjd, t_max.mjd)[0]
+            return integral / self.time_sum(t_min, t_max).value
```
adonath, line 408: Maybe move the return statement to the end? Why is it actually needed to distinguish the cases? Naively I'd assume an adaptive integrate method such as `quad` should deal with the shape of the function correctly. Does the `quad` method yield acceptable performance? If not you could rely on oversampling as well...

adonath (commented):
Thanks @aaguasca! I have left some inline comments. 

Ideally you can add the model to the model gallery as well. See https://github.com/gammapy/gammapy/tree/master/examples/models/temporal for more examples.

gammapy/modeling/models/temporal.py
```diff
@@ -334,6 +334,93 @@ def integral(self, t_min, t_max, **kwargs):
...
+    """
+
+    tag = ["GeneralizedGaussianTemporalModel", "gengauss"]
+
+    _t_max_default = Time("2000-01-01")
+    t_maxim = Parameter("t_maxim", _t_max_default.mjd, unit = "day", frozen=False)
+    sigma_r = Parameter("sigma_r", "1d", frozen=False)
+    sigma_d = Parameter("sigma_d", "1d", frozen=False)
+    v = Parameter("v", 2, unit = "", frozen=False)
+    A = Parameter("A", 1, unit = "", frozen=False)
```
adonath, line 367: By definition all temporal models in Gammapy do not have an amplitude parameter. The amplitude is currently always carried by the `SpectralModel`, which is mandatory in any model definition. Please remove...

gammapy/modeling/models/temporal.py
```diff
@@ -334,6 +334,93 @@ def integral(self, t_min, t_max, **kwargs):
...
+    
+    """
+
+    tag = ["GeneralizedGaussianTemporalModel", "gengauss"]
+
+    _t_max_default = Time("2000-01-01")
+    t_maxim = Parameter("t_maxim", _t_max_default.mjd, unit = "day", frozen=False)
+    sigma_r = Parameter("sigma_r", "1d", frozen=False)
+    sigma_d = Parameter("sigma_d", "1d", frozen=False)
+    v = Parameter("v", 2, unit = "", frozen=False)
```
adonath, line 366: Maybe find a longer, more descriptive parameter name? Even a greek letter might be better...

gammapy/modeling/models/temporal.py
```diff
@@ -334,6 +334,93 @@ def integral(self, t_min, t_max, **kwargs):
...
+    v : `~astropy.units.Quantity`
+        Pulse sharpness -> lower values implies a more peaked pulse
+    
+    """
+
+    tag = ["GeneralizedGaussianTemporalModel", "gengauss"]
+
+    _t_max_default = Time("2000-01-01")
+    t_maxim = Parameter("t_maxim", _t_max_default.mjd, unit = "day", frozen=False)
+    sigma_r = Parameter("sigma_r", "1d", frozen=False)
```
adonath, line 364: Maybe rename to something more explicit? I think even `sigma_left` and `sigma_right` might be better...

gammapy/modeling/models/temporal.py
aaguasca: I am sorry, but I am not familiar with functools.wrap. I have searched on Internet, but I am not sure how to use it. Could you please be more specific?

gammapy/modeling/models/temporal.py
aaguasca: I am not sure if `quad` yields acceptable performance. How could I check it? Is there any maximum lapse time the code should not exceed?

gammapy/modeling/models/temporal.py
```diff
@@ -334,6 +334,74 @@ def integral(self, t_min, t_max, **kwargs):
...
+    
+    """
+
+    tag = ["GeneralizedGaussianTemporalModel", "gengauss"]
+
+    _t_ref_default = Time("2000-01-01")
+    t_ref = Parameter("t_ref", _t_ref_default.mjd, unit = "day", frozen=False)
+    sigma_rise = Parameter("sigma_rise", "1d", frozen=False)
+    sigma_decay = Parameter("sigma_decay", "1d", frozen=False)
+    nu = Parameter("nu", 2, unit = "", frozen=False)
```
QRemy, line 364: For consistency with the [spatial gen-gauss](https://docs.gammapy.org/dev/api/gammapy.modeling.models.GeneralizedGaussianSpatialModel.html) the shape parameter could be eta=1/nu

gammapy/modeling/models/temporal.py
```diff
@@ -334,6 +334,74 @@ def integral(self, t_min, t_max, **kwargs):
...
+
+    @staticmethod
+    def evaluate(time, t_ref, sigma_rise, nu, sigma_decay):
+        val_rise = np.exp( - (np.abs((time - t_ref).value) ** nu) / (sigma_rise.value ** nu))
+        val_decay = np.exp( - (np.abs((time - t_ref).value) ** nu) / (sigma_decay.value ** nu))
+        val = np.where(time < t_ref, val_rise, val_decay)
+        return val
```
QRemy, line 371: for consistency with the [GaussianTemporalModel](https://docs.gammapy.org/dev/_modules/gammapy/modeling/models/temporal.html#GaussianTemporalModel) a factor 2 could be added in the denominator within the exponential, so it would give the same result for equivalent parameters (i.e.  sigma=sigma_rise =sigma_decay and nu=2).

QRemy (commented):
Thanks for the addition of this model, I left two minor comments about the consistency with the other models.

registerrier:
What is the status here @aaguasca , @adonath @QRemy ? Most comments have been adressed no?

gammapy/modeling/models/temporal.py
```diff
@@ -334,6 +334,74 @@ def integral(self, t_min, t_max, **kwargs):
...
+            Integrated flux norm on the given time intervals
+        """
+        
+        pars = self.parameters
+        t_rise = pars["t_rise"].quantity.to_value("day")
+        t_decay = pars["t_decay"].quantity.to_value("day")
+        eta = pars["eta"].quantity
+        t_ref = Time(pars["t_ref"].quantity, format = "mjd")
+                     
+        def gen_gauss(time,t_ref, t_rise, t_decay, eta):
```
adonath, line 395: Why is this re-declaration needed? I think directly calling `.evaluate()` shock work fine...

adonath (changes requested):
Apologies @aaguasca for the dallied review here. I have left one comment on the code duplication on the integral method. Otherwise please add a minimal unit test of the model here: https://github.com/gammapy/gammapy/blob/master/gammapy/modeling/models/tests/test_temporal.py

gammapy/modeling/models/temporal.py
adonath: Would great if you had time to address this remaining comment @aaguasca...Thanks!

aaguasca: I was waiting for an answer in issue #3736 before uploading any commit. Depending on the answer, I should change the code, as I cannot introduce more than one interval of time in quad.

aaguasca:
I hope these commits solve the last comment :). Sorry for the delay.
I changed the units of t_ref, t_rise and t_decay to be compatible with the input units of these variables in evaluate.
As the integral method is defined right now, it cannot compute a set of different intervals all at once. Is it required?  
Let me know anything should be changed.

aaguasca:
Could you please explain to me why some checks were not successful? I do not understand the errors that appear in the details.

registerrier:
> Could you please explain to me why some checks were not successful? I do not understand the errors that appear in the details.

I think the issue is due to the new spicy release which breaks this specific test. There was a recent PR to avoid this error. See #3786
This is unrelated and not an issue here.

adonath (approved):
Thanks a lot @aaguasca, no further comments from my side!

registerrier (approved):
Thanks @aaguasca

Issue #3296, opened 2021-04-07
Support for energy dependent `RAD_MAX` point-like IRFs


maxnoe:
**Is your feature request related to a problem? Please describe.**

Gammapy currently seems to have no support for energy-depedent `RAD_MAX` for point-like IRFs as described in the DL3 format here:
https://open-gamma-ray-astro.github.io/gamma-astro-data-formats/irfs/point_like/index.html#rad-max-2d

**Describe the solution you'd like**

Add support for using the energy-dependent `RAD_MAX` table and corresponding IRFs.

**Describe alternatives you've considered**

No alternatives. Energy-dependent event selection is a needed feature for most if not all IACTs and point-like IRFs are much easier to handle/calculate when applicable. 

**Additional context**

At least LST, MAGIC and FACT would like to use gammapy with point-like IRFs with energy-dependent `RAD_MAX`

chaimain:
Yes, we from LST would like to have this support as well, as we recently started producing standard DL3 files, with fixed selection cuts (gammaness and theta2) and would soon be moving towards producing IRFs with energy-dependent cuts.

maxnoe:
Reading RAD_MAX_2D works with this now: https://github.com/gammapy/gammapy/pull/3515

adonath:
I had the chance to talk to @registerrier yesterday about this and here are a few general comments / questions:
- Fully integrating this in the Gammapy API is hard: adding full support for energy dependent regions to `RegionGeom` with support for methods such as `RegionGeom.contains()`, `RegionGeom.plot()`, `RegionGeom.solid_angle()` and supporting serialisation is a lot of work. Currently there is no support in `regions` for array valued properties, so one would either need a custom class (such as `ArrayCircleSkyRegion`) or write a lot of `for` loops and declarative way such as `RegionGeom(regions, regions_axis=energy_axis)`. In general this would probably need a design document (PIG) and extensive review. If we decide to go for this, we should keep in mind to not only support the case, where regions are energy dependent but can vary with other axes as well.
- However getting this to work at all is simple.  It seems the only missing functionality is the possibility to bin events into a map with energy dependent region size and then it could just work with setting `RegionGeom(region=None)`. This could be written as a prototype in a single notebook.  My recommendation would be to start like this, use it as a basis for discussion and conclude from the prototype which additions to Gammapy are needed and reasonable.
[...]

Issue #3666, opened 2021-11-26
Make RAD_MAX hdu optional?


AtreyeeS:
The `RAD_MAX` hdu is assumed to be a default IRF in 0.19,

https://github.com/gammapy/gammapy/blob/d5d2e12fb2c4bdd3f9c027d97ac7d94056d2f25d/gammapy/data/data_store.py#L229

 leading to a warning message `No HDU found matching: OBS_ID = 33088, HDU_TYPE = rad_max, HDU_CLASS = None` for **every** run for present HESS and CTA datastores. This is rather annoying. Specially because `RAD_MAX` is not listed as one of the `available_irfs`, it appears a bit from nowhere.

https://github.com/gammapy/gammapy/blob/d5d2e12fb2c4bdd3f9c027d97ac7d94056d2f25d/gammapy/data/data_store.py#L260

This can (is) mislead(ing) users into believing that there is some issue with their runs, (and suppressing gammapy warnings is probably not the ideal solution).

Should we make this an optional hdu ? @registerrier @adonath @cosimoNigro @maxnoe @bkhelifi

maxnoe:
@AtreyeeS and we get warnings about missing PSF and Background IRFs for each and every pointlike dataset we load. 

Maybe this warning system as a whole should be removed and replaced with a *clear* error message as soon something is actually trying to use one of these IRFs.

bkhelifi:
rad_max is obviously optional... depending if the IRF is point-like or not. Do we have a metadata in the IRF files (edisp, area) are point-like or not?
By the way, the use cases are a bit more complex:
- full-containment production: we require to have "events", "gti", "aeff", "edisp", "psf". I do not remember for "bkg"
- point-source production: we require to have "events", "gti", "aeff", "edisp" and "rad_max" (the "bkg" is optional, "psf" not needed at all)
So, the design seems to be a bit more complex to handle these 2 use cases. Is not it?

registerrier:
I agree. The logic behind should be changed. 

The `HDULocation` should not be in charge of warning of missing entries, because it cannot know what are the expected ones. What it could do is to emit a debug message about which HDUs it found.

The `Observation` constructor could complain if something is incorrect. For instance, if  `aeff.is_pointlike` but there is no `RADMAX_2D` or no `RADMAX` keyword in the header. But it is likely inefficient or problematic if we force some combinations directly at the level of the `Observation.__init__` (e.g. this forces to load IRFs into memory from the start).
[...]

PR #3684, opened 2021-12-10
Started to implement the energy-dependent 1D spectrum extraction

cosimoNigro:
**Description**

**What is the motivation for this pull request?**
This PR starts the implementation of the energy-dependent spectrum extraction.

**Briefly: what changes are done here?**
* For the moment I have modified the `SpectrumDatasetMaker` - actually the `MapDatasetMaker` from which it inherits - to use the information in `~gammapy.irf.RadMax2D` to create a different ON region for each energy bin and fill the array with the ON counts accordingly.

* The next step would be to modify the `ReflectedRegionsBackgroundMaker` to extract the OFF counts using the same strategy.

**If this is to address a Github issue, mention its number to create a link**
Related to Issue #3296.

**Status**
This is a draft, meant for discussion. No docs and no tests have been added yet.
But find here a snippet of how the energy-dependent spectrum extraction will look with these modifications (I stop at the `SpectrumDatasetMaker.run`).
```python
from regions import PointSkyRegion
from gammapy.maps import MapAxis, RegionGeom
from gammapy.makers import SpectrumDatasetMaker
from gammapy.datasets import SpectrumDataset

# load the observation
observation = Observation.read("$GAMMAPY_DATA/magic/rad_max/data/magic_dl3_run_05029747.fits")
# observation = Observation.read("$GAMMAPY_DATA/hess-dl3-dr1/data/hess_dl3_dr1_obs_id_020323.fits.gz")

# energy axes
energy_axis = MapAxis.from_energy_bounds(
    0.005, 50, nbin=28, per_decade=False, unit="TeV", name="energy"
)
energy_true_axis = MapAxis.from_energy_bounds(
    0.005, 50, nbin=20, per_decade=False, unit="TeV", name="energy_true"
)

# create a point-like geometry for the centre of the ON region
target_position = SkyCoord(ra=83.63, dec=22.01, unit="deg", frame="icrs")
on_center = PointSkyRegion(target_position)
geom = RegionGeom.create(region=on_center, axes=[energy_axis])

# spectrum dataset and its maker
dataset_empty = SpectrumDataset.create(
    geom=geom, energy_axis_true=energy_true_axis
)

dataset_maker = SpectrumDatasetMaker(
    containment_correction=False, selection=["counts", "exposure", "edisp", "rad_max"]
)

# try to run it and plot the counts
dataset = dataset_maker.run(
    dataset_empty.copy(name="trial"), observation
)
dataset.counts.plot(drawstyle="steps-mid", ls="-")
plt.grid(ls=":")
plt.show()
```

Figure_1

It looks in agreement with what I was obtaining with MARS and with my custom scripts
run_5029747_counts_comparison

Thanks!

P.S.
It would help other testing if you merge https://github.com/gammapy/gammapy-data/pull/18.

### DIFF
--- a/gammapy/makers/map.py
+++ b/gammapy/makers/map.py
@@ -83,7 +83,7 @@ class MapDatasetMaker(Maker):
             Counts map.
         """
         if geom.is_region and isinstance(geom.region, PointSkyRegion):
-            counts = make_counts_rad_max(geom, observation)
+            counts = make_counts_rad_max(geom, observation.rad_max, observation.events)
         else:
             counts = Map.from_geom(geom)
             counts.fill_events(observation.events)
--- a/gammapy/makers/map.py
+++ b/gammapy/makers/map.py
@@ -14,7 +14,7 @@ from .utils import (
     make_map_background_irf,
     make_map_exposure_true_energy,
     make_psf_map,
-    make_counts_rad_max
+    make_counts_rad_max,
 )
 
 __all__ = ["MapDatasetMaker"]
@@ -53,7 +53,7 @@ class MapDatasetMaker(Maker):
             selection = self.available_selection
 
         selection = set(selection)
-        
+
         if not selection.issubset(self.available_selection):
             difference = selection.difference(self.available_selection)
             raise ValueError(f"{difference} is not a valid method.")
@@ -64,7 +64,7 @@ class MapDatasetMaker(Maker):
     def make_counts(geom, observation):
         """Make counts map.
 
-        **NOTE for 1D analysis:** if the `~gammapy.maps.Geom` is built from a 
+        **NOTE for 1D analysis:** if the `~gammapy.maps.Geom` is built from a
         `~regions.CircleSkyRegion`, the latter will be directly used to extract
         the counts. If instead the `~gammapy.maps.Geom` is built from a
         `~regions.PointSkyRegion`, the size of the ON region is taken from
@@ -82,9 +82,9 @@ class MapDatasetMaker(Maker):
         counts : `~gammapy.maps.Map`
             Counts map.
         """
-        if isinstance(geom.region, PointSkyRegion):
+        if geom.is_region and isinstance(geom.region, PointSkyRegion):
             counts = make_counts_rad_max(geom, observation)
-        elif isinstance(geom.region, CircleSkyRegion):
+        else:
             counts = Map.from_geom(geom)
             counts.fill_events(observation.events)
         return counts
--- a/gammapy/makers/map.py
+++ b/gammapy/makers/map.py
@@ -64,10 +64,9 @@ class MapDatasetMaker(Maker):
     def make_counts(geom, observation):
         """Make counts map.
 
-        **NOTE:** for 1D analyses, if the `~gammapy.maps.Geom` is built from a
+        **NOTE for 1D analysis:** if the `~gammapy.maps.Geom` is built from a 
         `~regions.CircleSkyRegion`, the latter will be directly used to extract
-        the counts.
-        If instead the `~gammapy.maps.Geom` is built from a
+        the counts. If instead the `~gammapy.maps.Geom` is built from a
         `~regions.PointSkyRegion`, the size of the ON region is taken from
         the `RAD_MAX_2D` table containing energy-dependent theta2 cuts.
 
--- a/gammapy/makers/map.py
+++ b/gammapy/makers/map.py
@@ -1,9 +1,12 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 import logging
+import numpy as np
 import astropy.units as u
 from astropy.table import Table
+from gammapy.maps.region.ndmap import RegionNDMap
+from regions import CircleSkyRegion
 from gammapy.irf import EDispKernelMap, PSFMap
-from gammapy.maps import Map
+from gammapy.maps import MapAxis, Map, RegionGeom
 from .core import Maker
 from .utils import (
     make_edisp_kernel_map,
@@ -35,7 +38,7 @@ class MapDatasetMaker(Maker):
     """
 
     tag = "MapDatasetMaker"
-    available_selection = ["counts", "exposure", "background", "psf", "edisp"]
+    available_selection = ["counts", "exposure", "background", "psf", "edisp", "rad_max"]
 
     def __init__(
         self,
@@ -49,7 +52,7 @@ class MapDatasetMaker(Maker):
             selection = self.available_selection
 
         selection = set(selection)
-
+        
         if not selection.issubset(self.available_selection):
             difference = selection.difference(self.available_selection)
             raise ValueError(f"{difference} is not a valid method.")
@@ -74,6 +77,35 @@ class MapDatasetMaker(Maker):
         """
         counts = Map.from_geom(geom)
         counts.fill_events(observation.events)
+        return counts
+    
+    @staticmethod
+    def make_counts_rad_max(geom, observation):
+        """Extract the counts using for the ON region size the values in the `RAD_MAX_2D` table"""
+        if observation.rad_max is None:
+            raise TypeError("the IRF for this observation does not include a RAD_MAX_2D table")
+
+        # TODO: move this in a function similar to the make_map_<IRF_COMPONENT>
+        on_center = geom.center_skydir
+        offset = on_center.separation(observation.pointing_radec)
+        rad_max = observation.rad_max.evaluate(
+            offset=offset,
+            energy=geom.axes["energy"].center
+        )
+
+        counts_list = []
+        # create and fill a map per each energy bin, fetch the counts
+        for i, rad in enumerate(rad_max):
+            on_region = CircleSkyRegion(center=on_center, radius=rad)
+            energy_range = geom.axes["energy"].slice(i)
+            on_region_geom = RegionGeom(on_region, axes=[energy_range])
+            counts = Map.from_geom(on_region_geom)
+            counts.fill_events(observation.events)
+            counts_list.append(counts.data[0])
+
+        counts = RegionNDMap.from_geom(geom)
+        counts.data = np.asarray(counts_list)
+
         return counts
 
     @staticmethod
@@ -309,6 +341,10 @@ class MapDatasetMaker(Maker):
             counts = self.make_counts(dataset.counts.geom, observation)
         else:
             counts = Map.from_geom(dataset.counts.geom, data=0)
+        
+        if set(["counts", "rad_max"]).issubset(self.selection):
+            counts = self.make_counts_rad_max(dataset.counts.geom, observation)
+
         kwargs["counts"] = counts
 
         if "exposure" in self.selection:
--- a/gammapy/makers/map.py
+++ b/gammapy/makers/map.py
@@ -4,9 +4,9 @@ import numpy as np
 import astropy.units as u
 from astropy.table import Table
 from gammapy.maps.region.ndmap import RegionNDMap
-from regions import CircleSkyRegion
+from regions import PointSkyRegion, CircleSkyRegion
 from gammapy.irf import EDispKernelMap, PSFMap
-from gammapy.maps import MapAxis, Map, RegionGeom
+from gammapy.maps import Map, RegionGeom
 from .core import Maker
 from .utils import (
     make_edisp_kernel_map,
@@ -14,6 +14,7 @@ from .utils import (
     make_map_background_irf,
     make_map_exposure_true_energy,
     make_psf_map,
+    make_counts_rad_max
 )
 
 __all__ = ["MapDatasetMaker"]
@@ -38,7 +39,7 @@ class MapDatasetMaker(Maker):
     """
 
     tag = "MapDatasetMaker"
-    available_selection = ["counts", "exposure", "background", "psf", "edisp", "rad_max"]
+    available_selection = ["counts", "exposure", "background", "psf", "edisp"]
 
     def __init__(
         self,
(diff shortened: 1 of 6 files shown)


gammapy/makers/map.py
```diff
@@ -75,6 +78,35 @@ def make_counts(geom, observation):
         counts = Map.from_geom(geom)
         counts.fill_events(observation.events)
         return counts
+    
+    @staticmethod
+    def make_counts_rad_max(geom, observation):
+        """Extract the counts using for the ON region size the values in the `RAD_MAX_2D` table"""
+        if observation.rad_max is None:
+            raise TypeError("the IRF for this observation does not include a RAD_MAX_2D table")
```
maxnoe, line 86: I don't think this is a `TypeError`. `TypeError` is meant for cases like a function expecting an `int` but getting a `str`.

This is either a `ValueError` or maybe, also in regard to https://github.com/gammapy/gammapy/issues/3666, we should create a ` RequiredIRFMissing` or similar exception in gammapy.

gammapy/makers/map.py
```diff
@@ -75,6 +78,35 @@ def make_counts(geom, observation):
         counts = Map.from_geom(geom)
         counts.fill_events(observation.events)
         return counts
+    
+    @staticmethod
+    def make_counts_rad_max(geom, observation):
```
adonath, line 83: I agree the method make sense in terms of code organisation, but maybe not that much in terms of API? I think ideally the handling of energy dependent region sizes could be almost hidden completely to the user. E.g. like:

```python3
from gammapy.makers.utils import make_counts_rad_max

def make_counts(geom, observation):
      
      # TODO: Maybe add RegionGeom.is_energy_dependent or similar....
      if geom.is_region and isinstance(geom.region, PointSkyRegion):
          counts = make_counts_rad_max(geom, rad_max, events)
      else:
           counts.fill_events()

```

gammapy/makers/map.py
```diff
@@ -35,7 +38,7 @@ class MapDatasetMaker(Maker):
     """
 
     tag = "MapDatasetMaker"
-    available_selection = ["counts", "exposure", "background", "psf", "edisp"]
+    available_selection = ["counts", "exposure", "background", "psf", "edisp", "rad_max"]
```
adonath, line 41: The `rad_max` option here is not really part of the data reduction as the other quantities. I think the `selection` rather refers to "which" data to reduce, while the `rad_max` rather refers to "how" it is reduced. In practice it is actually not optional at all, because once the IRFs are point like and the `rad_max` info is present one has to conduct the data reduction this way. So I would completely remove the option and not expose it to the user.

gammapy/makers/map.py
```diff
@@ -75,6 +78,35 @@ def make_counts(geom, observation):
...
+        # create and fill a map per each energy bin, fetch the counts
+        for i, rad in enumerate(rad_max):
+            on_region = CircleSkyRegion(center=on_center, radius=rad)
+            energy_range = geom.axes["energy"].slice(i)
+            on_region_geom = RegionGeom(on_region, axes=[energy_range])
+            counts = Map.from_geom(on_region_geom)
+            counts.fill_events(observation.events)
+            counts_list.append(counts.data[0])
+
+        counts = RegionNDMap.from_geom(geom)
```
adonath, line 106: You can directly pass the data here, like `RegionNDMap.from_geom(geom=geom, data=np.array(counts_list))`

gammapy/makers/map.py
```diff
@@ -75,6 +78,35 @@ def make_counts(geom, observation):
...
+        )
+
+        counts_list = []
+        # create and fill a map per each energy bin, fetch the counts
+        for i, rad in enumerate(rad_max):
+            on_region = CircleSkyRegion(center=on_center, radius=rad)
+            energy_range = geom.axes["energy"].slice(i)
+            on_region_geom = RegionGeom(on_region, axes=[energy_range])
+            counts = Map.from_geom(on_region_geom)
+            counts.fill_events(observation.events)
```
adonath, line 103: Because of the lazy loading the access to `observation.events` will actually re-load the event from disk. To prevent this just introduce a local variable at the top of the method:

```python3
@staticmethod
def make_counts(geom, observation):
    events = observation.events
    ...
```

gammapy/makers/map.py
adonath, line 106: Creating from the point geometry means the information on which `rad_max` was actually used will be lost later. Of course it can be re-created from the IRFs later, but those might not be accessible anymore. Technically one could work with a `CompoundSkyRegion` here, just to preserve the information at all. But this requires careful documentation to explain how the regions map to the energy bins. Maybe it creates more confusion than it helps...

adonath (changes requested):
Thanks @cosimoNigro, I have left a few more comments. I think most importantly we should not expose any option to the users to choose the `rad_max`, because it is basically mandatory once the IRFs are defined. To leave the option for a later extension / refactoring with energy dependent region sizes on the `RegionGeom`, I would propose to keep the API changes minimal and just introduce `gammapy.makers.utils.make_counts_rad_max`, which is then called in `MapDatasetMaker.make_counts()`, based on the unique identification of the rad max case.

cosimoNigro:
Hello @adonath, @registerrier can you take a look now?
I implemented the energy-dependent extraction of the OFF counts using the `ReflectedRegionsBackgroundMaker` and added a test in `test_reduce`, testing the new functionalities. Let me know if you want me to add more tests and where.

This is how the energy-dependent spectrum extraction would look like. Basically identical to the fixed-energy one but passing a `PointSkyRegion` as centre of the ON region. I'll add a notebook in a follow-up PR.
```python
import astropy.units as u
from regions import PointSkyRegion
from gammapy.maps import MapAxis, RegionGeom
from gammapy.makers import SpectrumDatasetMaker, ReflectedRegionsBackgroundMaker
from gammapy.datasets import SpectrumDataset

# load the observation
observation = Observation.read("$GAMMAPY_DATA/magic/rad_max/data/magic_dl3_run_05029747.fits")

# energy axes
energy_axis = MapAxis.from_energy_bounds(
    0.005, 50, nbin=28, per_decade=False, unit="TeV", name="energy"
)
energy_true_axis = MapAxis.from_energy_bounds(
    0.005, 50, nbin=20, per_decade=False, unit="TeV", name="energy_true"
)

# create a point-like geometry for the centre of the ON region
target_position = SkyCoord(ra=83.63, dec=22.01, unit="deg", frame="icrs")
on_center = PointSkyRegion(target_position)
geom = RegionGeom.create(region=on_center, axes=[energy_axis])

# spectrum dataset and its maker (fills the ON counts)
dataset_empty = SpectrumDataset.create(
    geom=geom, energy_axis_true=energy_true_axis
)
dataset_maker = SpectrumDatasetMaker(
    containment_correction=False, selection=["counts", "exposure", "edisp"]
)
dataset = dataset_maker.run(
    dataset_empty.copy(name="trial"), observation
)

# reflected regions maker (fills the OFF counts)
bkg_maker = ReflectedRegionsBackgroundMaker()
dataset_on_off = bkg_maker.run(dataset, observation)

# try to plot them
dataset_on_off.counts.plot(drawstyle="steps-mid", ls="-", label="ON")
dataset_on_off.counts_off.plot(drawstyle="steps-mid", ls="-", label="OFF")
dataset_on_off.acceptance_off.plot(drawstyle="steps-mid", ls="-", label="acceptance")
plt.legend()
plt.grid(ls=":")
plt.show()
```
Figure_1

maxnoe:
@cosimoNigro these on/off counts look wrong too me, there seems to be an issue. In the central bins you have roughly 3 times more off than on counts? 

And why are all off bins below 30 GeV empty but not the on counts?

cosimoNigro:
Hello @maxnoe 

> In the central bins you have roughly 3 times more off than on counts?

But you have to scale those for the acceptance

> And why are all off bins below 30 GeV empty but not the on counts?

In the low energy bins it fails to find anyOFF regions. That was a mistake from my side.
I stored a too large theta2 cut in the low energy bins, namely 0.2, which translates to theta = 0.45, so larger than the offset.
The pointing position is within the ON region and no symmetric OFF could be found.
Will fix this with the next data.

maxnoe:
@cosimoNigro Wait, does this code use different off regions (not just different sizes of off regions) for each energy bin?

That would be a very surprising behavior to me.

maxnoe:
Ah, and failure to find any off regions should probably result in an exception

cosimoNigro:
> does this code use different off regions (not just different sizes of off regions) for each energy bin?

Thanks Max, I did not think about this. 
Yes, for each estimated energy bins it is simply re-running the `ReflectedRegionsBackgroundFinder`, so trying to fit as many OFF regions as possible.
How can we force it to use the same number of OFF regions? (e.g. 1 or 3).
Specifying e.g. [`max_region_number=1`](https://github.com/gammapy/gammapy/blob/master/gammapy/makers/background/reflected.py#L75) will not work, as it will not place it symmetrically w.r.t. the ON position.

maxnoe:
@cosimoNigro I think this goes back to one of the initial discussions we had of implementing a different version of the `ReflectedRegionsBackgroundFinder` that will use the "classical" approach of distributing n off regions along

using `offset_angles = (np.arange(n_off_positions) + 1) * 2 * np.pi / (n_off_positions + 1)` ?

adonath:
@cosimoNigro and @maxnoe I think this could just be an additional option to the existing `ReflectedRegionsBackgroundMaker`, implemented independently of this PR. How does one handle exclusion regions in this approach?

maxnoe:
You remove positions that intersect with an exclusion Region.

registerrier:
I suppose it is unclear whether the `ReflectedRegionsBackgroundMaker` should support the `radmax2d` at all. At least the fixed RADMAX case should be supported.
 
For clarity, it might be simpler to have another background `Maker` for the wobble/symetric region. Otherwise, the arguments might get pretty complicated.

gammapy/makers/utils.py
```diff
@@ -422,3 +424,161 @@ def make_theta_squared_table(
...
+    offset = on_center.separation(observation.pointing_radec)
+
+    rad_max = observation.rad_max.evaluate(
+        offset=offset, energy=geom.axes["energy"].center
+    )
+
+    return rad_max
+
+
+def make_counts_rad_max(geom, observation):
```
adonath, line 460: One last minor comment here: we typically kept the signature of the `make_something` utility methods to take the IRF classes and not the observation objects. The idea I just that this way they can used a bit easier independently. The API relying on the `Obseervation` object is still available on the `Maker` class, like `MapDatasetMaker.make_off_counts(geom, observation)`. So maybe changee the signature here to take the `geom`, `rad_max` and `events` independently?

gammapy/makers/utils.py
```diff
@@ -422,3 +424,161 @@ def make_theta_squared_table(
...
+        counts = Map.from_geom(on_region_geom)
+        counts.fill_events(events)
+        counts_list.append(counts.data[0])
+
+    counts = RegionNDMap.from_geom(geom, data=np.asarray(counts_list))
+
+    return counts
+
+
+def make_counts_off_rad_max(
```
adonath, line 495: Same here maybe remove the `observation` container from the signature and rely on the individual classes instead..

adonath (changes requested):
Thanks @cosimoNigro, I have left two final minor comments on the signature of the `maker_...` utility methods, otherwise looks good to me know. The proposed changes to the reflected regions background maker can be implemented in a different PR.

gammapy/makers/utils.py
cosimoNigro: Hello @adonath, I changed the function's signature from `make_counts_rad_max(geom, observation)` to `make_counts_rad_max(geom, rad_max, events)`.

gammapy/makers/utils.py
cosimoNigro: Changed the function's signature from `make_counts_off_rad_max(geom, observation)` to `make_counts_off_rad_max(geom, rad_max, events)`

cosimoNigro:
Hi @adonath, implemented the change in the function arguments you requested. Sorry for the delay.
Cheers!

gammapy/makers/utils.py
```diff
@@ -422,3 +424,161 @@ def make_theta_squared_table(
...
+    -------
+    counts : `~gammapy.maps.RegionNDMap`
+        Counts vs estimated energy extracted from the ON region.
+    """
+    rad_max = get_rad_max_vs_energy(rad_max, events.pointing_radec, geom)
+
+    counts_list = []
+
+    # create and fill a map per each energy bin, fetch the counts
+    for i, rad in enumerate(rad_max):
```
registerrier, line 480: Maybe a comment here:
since RAD_MAX is determined independently of any projection, the region based selection is not a necessity. It might be more efficient here to perform the filtering on the `EventList` directly.

gammapy/makers/utils.py
```diff
@@ -422,3 +424,161 @@ def make_theta_squared_table(
...
+        the RAD_MAX_2D table IRF
+    events : `~gammapy.data.EventList`
+        event list to be used to compute the ON counts
+
+    Returns
+    -------
+    counts : `~gammapy.maps.RegionNDMap`
+        Counts vs estimated energy extracted from the ON region.
+    """
+    rad_max = get_rad_max_vs_energy(rad_max, events.pointing_radec, geom)
```
registerrier, line 475: A question here. Is the rad_max determination usually done directly from the event energy or from the bin energy? i.e. is it expected that different energy binning yield different select event lists?

registerrier (approved):
Thanks @cosimoNigro ! This looks good.

I left two questions inline but they can be addressed later.

gammapy/makers/utils.py
maxnoe: Yes, just using `coordinate.separation(source_position) < rad_max` should be more efficient

gammapy/makers/utils.py
maxnoe: Energy binning of what? I'd say you look up the cut value based on the event energy, like we do here: 

https://github.com/cta-observatory/pyirf/blob/3f9cfba672c8a25eb5094d25acf411c6b2104112/pyirf/cuts.py#L104

gammapy/makers/utils.py
cosimoNigro: But the `RegionGeom` serves also the purpose of filtering by energy, if you look at it, it is defined by a single energy bin
If I filter your way then I have to filter also by energy, right?

registerrier:
Thanks @cosimoNigro ! Let's merge now and rediscuss the other aspects in the follow-up PRs.

PR #3687, opened 2021-12-14
Fix Fit.covariance shape in case of linked parameters

QRemy:
Fix the shape of the covariance matrix in `Fit.covariance` in case of linked parameters using the full list of parameters instead of the list of unique parameters.

### DIFF
--- a/gammapy/datasets/tests/test_map.py
+++ b/gammapy/datasets/tests/test_map.py
@@ -689,6 +689,53 @@ def test_map_fit(sky_model, geom, geom_etrue):
     dataset_1.npred()
     assert not dataset_1.evaluators["test-model"].contributes
 
+@requires_dependency("iminuit")
+@requires_data()
+def test_map_fit_linked(sky_model, geom, geom_etrue):
+    dataset_1 = get_map_dataset(geom, geom_etrue, name="test-1")
+    dataset_2 = get_map_dataset(geom, geom_etrue, name="test-2")
+    datasets = Datasets([dataset_1, dataset_2])
+
+    models = Models(datasets.models)
+    models.insert(0, sky_model)
+    sky_model2 = sky_model.copy(name="test-model-2")
+    sky_model2.spectral_model.index = sky_model.spectral_model.index
+    sky_model2.spectral_model.reference = sky_model.spectral_model.reference
+
+    models.insert(0, sky_model2)
+
+    models["test-1-bkg"].spectral_model.norm.value = 0.5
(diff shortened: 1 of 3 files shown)


gammapy/modeling/fit.py
```diff
@@ -241,7 +241,8 @@ def covariance(self, datasets):
         result : `CovarianceResult`
             Results
         """
-        datasets, parameters = self._parse_datasets(datasets=datasets)
+        datasets, _ = self._parse_datasets(datasets=datasets)
+        parameters = datasets.models.parameters
```
adonath, line 245: I'd be surprised if this worked. Directly accessing `datasets.models.parameters` will keep the shared parameter duplicated in the parameter list (as opposed to https://github.com/gammapy/gammapy/blob/master/gammapy/datasets/core.py#L147). This causes the optimizer to see the shared parameter as independent parameters, which makes them correlated 100%. In any case we should start with a unit test showing the parameter sharing works with the fit here: https://github.com/gammapy/gammapy/blob/master/gammapy/modeling/tests/test_fit.py

Then we can reproduce the issue reported in Slack and find the proper fix.

QRemy: Using `parameters `or `unique_parameters` in `compute` makes no difference (entries for duplicated parameters will be set to zero anyway), but `from_factor_matrix` need `parameters ` as input in order to expand correctly the `factor_matrix`

gammapy/datasets/tests/test_map.py
```diff
@@ -689,6 +689,51 @@ def test_map_fit(sky_model, geom, geom_etrue):
...
+    assert result.success
+    assert "minuit" in repr(result)
+
+    assert sky_model2.parameters["index"] is sky_model.parameters["index"]
+    assert sky_model2.parameters["reference"] is sky_model.parameters["reference"]
+
+    assert len(datasets.models.parameters.unique_parameters) == 20 
+    assert datasets.models.covariance.shape == (22,22) 
+
+    models.to_dict()
```
adonath, line 729: This test (and following asserts) seems a bit out of place maybe move to `gammapy.modeling.models` instead or remove completely?

adonath (changes requested):
Thanks @QRemy, I have left one last comment on the location of the link label test. Otherwise looks good to me!

gammapy/datasets/tests/test_map.py
QRemy: removed

registerrier (approved):
Thanks QRemy ! This is good.

Issue #3680, opened 2021-12-07
FoVBackgroundMaker with method=fit ignores min_counts and min_npred_background


luca-giunti:
**Gammapy version**
0.19

**Description**
Reminder issue to improve consistency: the options `min_counts` and `min_npred_background` passed to the `FoVBackgroundMaker` are used only if the method is `scale`, not for `fit`. This should be documented if there is a good reason behind, or they should be applied in both cases.

registerrier:
Hello @AtreyeeS . This issue can be closed now that #3688 is merged right?

AtreyeeS:
Addressed in #3688

PR #3688, opened 2021-12-15
Apply min counts consistently to scale and fit in FoVBackgroundMaker

AtreyeeS:
This PR address #3680 

I am not sure if applying a total min_counts requirement is the best option for fit (probably some requirement on each bin will be more reasonable), hence this was not applied to the `fit` option earlier.
However, this is confusing, and the present PR makes the behaviour consistent.


luca-giunti:
Thank you @AtreyeeS ! I think it makes sense to make this consistent

adonath (approved):

registerrier (approved):
Thanks @AtreyeeS ! This looks good.

Issue #3681, opened 2021-12-08
Implement PSFKernelMap


adonath:
In order to support asymmetric PSF Kernels as well as e.g. pre computed PSF kernels, we should implement a `PSFKernelMap`.

adonath:
@LauraOlivera I remember you had some interest in this in the past. Feel free to un-assign if you are not interested or don't have time...

PR #3689, opened 2021-12-16
Implement PSFKernelMap

LauraOlivera:
This pull request addresses issue #3681 to implement a `PSFKernelMap`. 

So far, I have created the object and implemented a `from_gauss()` method to easily create Gaussian kernel maps. Since one of the uses of this implementation are asymmetric PSF kernels, I added an `AsymmetricGauss2DPDF` to `gammapy.utils.gauss` to easily create asymmetric Gaussian kernels.

I marked the PR as draft because I still have a bunch to do. An incomplete list below:

- [ ] Implement `from_geom` method
- [ ] Implement `get_psf_kernel()` ?
- [ ] Containment radius methods ?
- [ ] tests for everything
- [ ] Orientation of the asymmetric Gaussian ?

### DIFF
--- a/gammapy/irf/psf/map.py
+++ b/gammapy/irf/psf/map.py
@@ -10,7 +10,7 @@ from ..core import IRFMap
 from .core import PSF
 from .kernel import PSFKernel
 
-__all__ = ["PSFMap"]
+__all__ = ["PSFMap", "PSFKernelMap"]
 
 
 class IRFLikePSF(PSF):
@@ -490,3 +490,11 @@ class PSFMap(IRFMap):
 
     def __str__(self):
         return str(self.psf_map)
+
+
+class PSFKernelMap(IRFMap):
+    tag = "psf_kernel_map"
+    required_axes = ["psf_lon","psf_true", "energy_true"]
+
+    def __init__(self, psf_kernel_map, exposure_map=None):
+        super().__init__(irf_map=psf_kernel_map, exposure_map=exposure_map)
--- a/gammapy/irf/psf/map.py
+++ b/gammapy/irf/psf/map.py
@@ -4,7 +4,7 @@ import astropy.units as u
 from astropy.visualization import quantity_support
 from gammapy.maps import Map, MapAxis, MapCoord, WcsGeom
 from gammapy.modeling.models import PowerLawSpectralModel
-from gammapy.utils.gauss import Gauss2DPDF
+from gammapy.utils.gauss import Gauss2DPDF, AsymmetricGauss2DPDF
 from gammapy.utils.random import InverseCDFSampler, get_random_state
 from ..core import IRFMap
 from .core import PSF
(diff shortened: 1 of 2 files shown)


LauraOlivera:
I feel like I wasn't very creative when making the tests... Please let me know if you can think of something a better than what I implemented!

LauraOlivera:
I made the `geom` input optional, as we discussed in the dev call.

I also realised I could construct the `PSFKernelMap.from_gauss()` using `PSFKernel.from_spatial_model()` using a `GaussianSpatialModel` with e>0 for the asymmetric case. From a quick test I think that option would be slower, but maybe it is nicer in terms of less code to maintain? It also would guarantee the kernels are normalized, although they get normalized when doing `.get_psf_kernel()` anyway...
Let me know what you think @adonath

registerrier (commented):
Thanks @LauraOlivera .

This is a really useful addition. With the coming of non-isotropic PSF, this type of `PSFMap` will be needed.

I have the following concerns:
- I don't think this class represents a map of kernels but simply a map of non-isotropic PSF. kernel implies a geometry and a projection (and contain a pdf per bin and not per sr). This cannot be described by the `lon`  and `lat` axes. So this is really a generalized `PSFMap`. It could even possibly inherit from it and the `get_psf_kernel` should follow more closely the implementation of `PSFMap.get_psf_kernel`. 
- A priori, the stored  PSF is in physical coordinates. I assume it should be the same frame as the `IRFMap` geom, right? This would need to be documented somewhere.

adonath:
> I don't think this class represents a map of kernels but simply a map of non-isotropic PSF.

@registerrier Your description would imply that the `.get_psf_kernel()` does the reprojection on the WCS grid, is this correct? I never thought of it this way. I think one huge advantage of the map of kernels is that the computation of the kernel is moved to the data reduction step, which solves one bottleneck of the fitting, where one could just extract the kernel instead of recomputing it, which takes time. The main question for me is whether we would ever need both. A map that really contains kernels (for performance) and one for the asymmetric PSF. My guess is no, because at the data reduction step the geometries are already fixed and don't change after. So one could directly compute the kernel at the data reduction step, just as we do for the `EDispKernelMap`.

On the other hand this makes the `PSFKernelMap` really bound to the geometry of e.g. the `exposure` that is used for the model evaluation.

adonath:
@registerrier Following your understanding of the asymmetric PSF map, we would maybe not need any `PSFKernelMap` at all, right? Just making sure the current `PSFMap` does not hard-code the dependency on `rad` should be sufficient...

adonath:
Just taking the thought a bit further, maybe it still make sense to have both, a `PSFKernelMap` and a `PSFMap` like so:
```python
class PSFMap(IRFMap):
    """Implements all common functionality, should be N-dimensional"""
    
    def to_psf_kernel_map(geom):
        """Compute PSFKernelMap"""
        return PSFKernelMap()


# This corresponds to our current PSFMap
class RadiallySymmetricPSFMap(PSFMap):
    """"""
    required_axes = ["rad", "energy_true"]


class NonRadiallySymmetricPSFMap(PSFMap):
    """"""
    required_axes = ["psf_lon", "psf_lat", "energy_true"]


class PSFKernelMap(IRFMap):
    """Contains precomputed PSFKernel"""
    def __init__(map, exposure, kernel_geom):
        pass
```
But I'm just brain-storming for now...

QRemy:
This requires more thoughts and we will probably need more than ore PR to solve this,
so I'll close this one for now.

Issue #3700, opened 2022-01-17
Incorrect predicted counts for models with negative amplitude.


LauraOlivera:
**Gammapy version**
`0.20.dev18+g735b25cd5.d20220112`

**Bug description**
When computing the predicted counts of a model that has negative amplitude, the "signal" is masked by this line

https://github.com/gammapy/gammapy/blob/3cd64fdb2c53c8e5c697a9b85ef8d0486bff0b76/gammapy/datasets/evaluator.py#L333-L337

The following snippet, with a negative source

```python3
from gammapy.datasets import MapDataset
from gammapy.modeling.models import SkyModel, PointSpatialModel, PowerLawSpectralModel
import matplotlib.pyplot as plt

dataset = MapDataset.read('$GAMMAPY_DATA/cta-1dc-gc/cta-1dc-gc.fits.gz')
 
spatial_model = PointSpatialModel(lon_0 = "0 deg", lat_0 = "0 deg", frame='galactic')
spectral_model = PowerLawSpectralModel(amplitude = "-1e-12 TeV-1 cm-2 s-1", index = 2)

model = SkyModel(spatial_model = spatial_model, spectral_model = spectral_model, name = 'gc negative')
model.apply_irf["psf_after_edisp"] = True

dataset = dataset.to_image()
dataset.models = model

dataset.npred().plot(add_cbar=True)
plt.show()
```
results in 
![image](https://user-images.githubusercontent.com/22521727/149830664-5c67169b-3a8c-4032-b26e-efecbff322c6.png)

**Expected behavior**
The right npred should have a negative excess in the middle, like
![image](https://user-images.githubusercontent.com/22521727/149830767-1db92d4f-cfa3-46f8-8f82-9f6f0f64b7fc.png)

This is an issue, for example when fitting a model where there is no source. The amplitude should be consistent with background fluctuations, which will some times be negative.

adonath:
Thanks a lot @LauraOlivera, indeed there is an obvious issue! Especially I remember there we convergence issues with upper limits at high energies. The clipping was introduced here https://github.com/gammapy/gammapy/pull/2342, because we saw negative predicted counts, which should not happened either. I think the proper fix here might be to move the zero clipping to `dataset.npred()` instead.

registerrier:
This is solved by #3704. Closing the issue.

PR #2342, opened 2019-09-05
Fix PSF convolution

luca-giunti:
This minimal PR fixes a bug in the PSF convolution.

Without this fix, calling dataset.npred() could predict negative counts in some bins, as shown in the attached figure (that whas produced calling `dataset.npred().slice_by_idx({'energy' : 7}).plot(vmax=0, add_cbar=True)
psf_wrong_map

PR #3704, opened 2022-01-19
Fix zero clipping in npred calculation

LauraOlivera:
This pull request addresses #3700 by moving the zero clipping applied to the predicted counts outside of the PSF convolution and into `dataset.npred`.

No tests are necessary because they were already added by #2342, which implemented the clipping in the first place:
https://github.com/gammapy/gammapy/blob/d342c5ac70cb321345f3c5e2c70f88ea8a116809/gammapy/datasets/tests/test_map.py#L250

With this fix, the snippet in #3700 results in

image

which is clipped where the negative values are. However, if the model amplitude is such that the negative model counts expected are less than the background counts (i.e. compatible with background fluctuations), then the result is this
image

which is not clipped to zero as it used to.

### DIFF
--- a/gammapy/datasets/map.py
+++ b/gammapy/datasets/map.py
@@ -393,7 +393,7 @@ class MapDataset(Dataset):
 
         if self.background:
             npred_total += self.npred_background()
-
+        npred_total.data[npred_total.data < 0.0] = 0
         return npred_total
 
     def npred_background(self):
--- a/gammapy/datasets/evaluator.py
+++ b/gammapy/datasets/evaluator.py
@@ -333,7 +333,6 @@ class MapEvaluator:
     def apply_psf(self, npred):
         """Convolve npred cube with PSF"""
         tmp = npred.convolve(self.psf)
-        tmp.data[tmp.data < 0.0] = 0
         return tmp
 
     def apply_edisp(self, npred):
--- a/gammapy/datasets/tests/test_evaluator.py
+++ b/gammapy/datasets/tests/test_evaluator.py
@@ -1,6 +1,7 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 from numpy.testing import assert_allclose
 import astropy.units as u
+import numpy as np
 from astropy.coordinates import SkyCoord
 from regions import CircleSkyRegion
 from gammapy.datasets.evaluator import MapEvaluator
@@ -12,6 +13,7 @@ from gammapy.modeling.models import (
     Models,
     PointSpatialModel,
     SkyModel,
+    PowerLawSpectralModel
 )
 from gammapy.utils.gauss import Gauss2DPDF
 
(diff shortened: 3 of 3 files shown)


gammapy/datasets/map.py
```diff
@@ -393,7 +393,7 @@ def npred(self):
 
         if self.background:
             npred_total += self.npred_background()
-
+        npred_total.data[npred_total.data < 0.0] = 0
```
registerrier, line 396: Is the same command fine? Does `np.clip` provide better performance?

registerrier (commented):
Thanks @LauraOlivera .

I have few comments:
- This will not fully solve the noise introduced by individual `fit_concolve`. Is it better to apply the filtering on the total npred compared to the npred per component?
- It might require a test since your specific problem went unnoticed: a negative flux should yield a negative npred at the `MapEvaluator` level no?

gammapy/datasets/map.py
adonath: I think @luca-giunti checked the implementation at the time and the "normal" masking was faster, if I remember correctly.

adonath:
> This will not fully solve the noise introduced by individual fit_concolve. Is it better to apply the filtering on the total npred compared to the npred per component?

I think there is no solution to this, right? To be able to correctly computed "negative" npred for source components the npred cannot be clipped. Statistically the clipping on the total npred is correct, the positive npred is only required after the background is added.

LauraOlivera:
Thanks @registerrier !

> It might require a test since your specific problem went unnoticed: a negative flux should yield a negative npred at the MapEvaluator level no?

Well I am not sure if we want to predict negative counts, since that is not really physical. The issue I was having is that in the absence of source, a the correct best-fit normalization might be (small) and negative, indicating compatible with a background fluctuation. Which was not possible with the existing implementation. So what I needed was `npred_signal` to be allowed to be negative, but the total `npred` should still be >=0 due to the background. I was of the understanding we don't want to allow negative `npred`. But of course I can change that...

registerrier:
> Thanks @registerrier !
> 
> > It might require a test since your specific problem went unnoticed: a negative flux should yield a negative npred at the MapEvaluator level no?
> 
> Well I am not sure if we want to predict negative counts, since that is not really physical. The issue I was having is that in the absence of source, a the correct best-fit normalization might be (small) and negative, indicating compatible with a background fluctuation. Which was not possible with the existing implementation. So what I needed was `npred_signal` to be allowed to be negative, but the total `npred` should still be >=0 due to the background. I was of the understanding we don't want to allow negative `npred`. But of course I can change that...
> 
> EDIT: yeah, pretty much what Axel said... Only saw his comment after.

OK. Indeed, the correct way to remove fit noise is probably to do it at the end by removing possible negative npred.

Regarding the test, I was thinking to `MapEvaluator.compute_npred()`. This one should return negative entries if fed with a negative flux.

registerrier (approved):
Thanks @LauraOlivera ! This looks good to me.

adonath (approved):
Thanks @LauraOlivera, no further comment from my side either..

PR #3711, opened 2022-01-21
Add ObservationsEventsSampler class

QRemy:
This PR add the following features:
- add a function to read ctools obs-def.xml or .csv files used to generate simulated observations.
- add .read() method to Observations class and .from_table() method to Observation class
- add ObservationsEventsSampler. This class create the datasets used by the MapDatasetEventsSamplers and run them (in parallel). I tried to introduce a way to get a reasonable geom for the dataset based on the observation and irfs properties.


bkhelifi:
No yaml format is planned?

QRemy:
@bkhelifi yes, that's something we have to discuss. For now we get Observations from the DataStore but this require to have the events files defined which is not possible if you want to simulate them. We could use an observation table that contains all the entries needed for simulation in a .fits or a yaml file.

QRemy:
Refactored to use #4917.

gammapy/datasets/tests/test_simulate.py
```diff
@@ -847,7 +847,7 @@ def test_MC_ID_flag(model_alternative):
 
 
 @requires_data()
-def test_simulated_observation_maker(signal_model, tmp_path):
+def test_observation_event_sampler(signal_model, tmp_path):
```
bkhelifi, line 850: Should we make a test with a sky model that has a temporal model? Is it wise?

bkhelifi (approved):
It seems ready to be merged... I have just a small question about tests with a temporal model

gammapy/datasets/tests/test_simulate.py
QRemy: it is tested for the MapDatasetEvent sampler that is used by the ObservationEventSampler, but I added a test for the ObservationsEventSampler.

bkhelifi (dismissed):
Thanks @QRemy

fabiopintore:
Many thanks @QRemy . The PR looks great to me and it is super-useful. 
I don't have any major comments!
My only doubt is why the `ObservationEventSampler` class returns an `observation` object rather than the simulated events as the `MapDatasetEventSampler` does. 

(I've spotted a few typos in the doc-strings, like missing commas or "defalut" instead of "default")

QRemy:
> My only doubt is why the `ObservationEventSampler` class returns an `observation` object rather than the simulated events as the `MapDatasetEventSampler` does.

I think we can introduce an option in a separate PR to avoid re-writting the irf and just refers to the existing file somehow.

Issue #3690, opened 2021-12-16
Crash when extracting flux points with a CompoundSpectralModel


registerrier:
**Gammapy version**
v0.19

**Bug description**
If a compound model is used for the `FluxPointsEstimator`, an error occurs in `FluxEstimator.get_scale_model()` because a `CompoundSpectralModel` has no `norm` or `amplitude` attribute.
 
**Expected behavior**
Using a `CompoundSpectralModel` should work. 

The issue is caused by the mechanism introduced to scale the `norm` according to the input parameter. It does not work in general for a `CompoundSpectralModel` which might have different `amplitude` and `norm` parameters. So that determining the actual range a priori might be ambiguous.

Note that we miss a test for flux points extraction with a compound model.

**To Reproduce**
```
from gammapy.modeling.models import EBLAbsorptionNormSpectralModel, PowerLawSpectralModel, SkyModel
from gammapy.estimators.flux import FluxEstimator

pl = PowerLawSpectralModel()
ebl = EBLAbsorptionNormSpectralModel.read_builtin(reference='finke')
model=SkyModel(spectral_model=pl*ebl)

fe = FluxEstimator()
fe.get_scale_model([model])
```

**Other information**
Any other information you think will be useful for us to fix the issue can go here.

registerrier:
Note that a workaround for v0.19 is shown in #3705

luca-giunti:
I think [this](https://github.com/gammapy/gammapy/blob/25f8aa20e263c7665d6a848c481e28b4b6b245f9/gammapy/estimators/flux.py#L116) still fails if one defines a custom model with a random name for the parameter to be scaled (e.g. `Flux`, or whatever). Also nothing prevents one to define a custom model with two parameters called `amplitude` and a `norm`, in which case we have a problem. I think our implicit convention on the names of parameters that can be scaled is not obvious, and should be better handled or documented.

Possible solutions that I see:
1. Adding an optional argument to `FluxPointEstimator.__init__()` to eventually specify the parameter to be scaled. But this can become problematic if one specifies e.g. `index`
2. Documentation. Specify that a custom model needs either an `amplitude`  or a `norm`  parameter. But this can become problematic e.g. if one defines a wrapper around a model coming from an external library, for which there is no freedom in the choice of parameter names. Fortunately the `naima` models have an `amplitude`, but this may not be the case in general
[...]

PR #3714, opened 2022-01-21
Correct FluxEstimator bug with compound model 

registerrier:
**Description**

This pull request resolves issues #3690 . The `amplitude` parameter is accessed with the `parameters` attribute
to avoid issues when it is not directly accessible on the model itself. 

Tests are added to make sure the scaling is possible with `CompoundSpectralModel` and with `NaimaSpectralModel`.

**Dear reviewer**


adonath (approved):
Thanks @registerrier, no further comments from my side. I think it's still non-ideal to rely on the parameter name, but this is not easy to fix. We would need some semantic description of the parameter to be an amplitude / norm parameter to improve this instead, which is a dedicated coding project on it's own...

PR #3719, opened 2022-01-24
Fix return of covariance result

### LINKED ISSUE #3703: FitResult print output is confusing
A `print(fit_result)` displays both the `covariance_result` and the `optimize_result` as `OptimizeResult`, eg: see cell 19
https://docs.gammapy.org/dev/tutorials/starting/analysis_2.html#Fit-the-model

Reminder issue to fix it during the sprint week

AtreyeeS:
Fixes #3703

### DIFF
--- a/gammapy/modeling/fit.py
+++ b/gammapy/modeling/fit.py
@@ -635,7 +635,7 @@ class FitResult:
     @property
     def covariance_result(self):
         """Optimize result"""
-        return self._optimize_result
+        return self._covariance_result
 
     def __repr__(self):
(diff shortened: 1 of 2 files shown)


adonath (changes requested):
Thanks @AtreyeeS, could maybe add one unit test here https://github.com/gammapy/gammapy/blob/master/gammapy/modeling/tests/test_fit.py#L89

AtreyeeS:
Thanks @adonath . Added

adonath (approved):

PR #3720, opened 2022-01-24
Add convenience method to write EventLists to file

LauraOlivera:
This pull request implements a `EventList.write` method to conveniently write event lists to disk. It has the option to pass a `GTI` object as input, which is saved together with the events in that case.

I've also updated the snippet in the documentation on how to save `EventList` objects to file.

### DIFF
--- a/gammapy/data/event_list.py
+++ b/gammapy/data/event_list.py
@@ -76,11 +76,24 @@ class EventList:
         table = Table.read(filename, **kwargs)
         return cls(table=table)
 
-    def write(self, filename, gti=None, overwrite=False):
+    def write(self, filename, gti=None, overwrite=False, format='gadf'):
         """Write the event list to a FITS file.
 
         If a GTI object is provided, it is saved into
         a second extension in the file.
+
+        Parameters
+        ----------
+        filename : `pathlib.Path`, str
+            Filename
+        gti : `~gammapy.data.GTI`
+            Good Time Intervals object to save to the same file.
+            Default is None.
+        overwrite : bool
+            Overwrite existing file?
+        format : str, optional
+            FITS format convention.  By default files will be written
+            to the gamma-astro-data-formats (GADF) format.
         """
 
         filename = make_path(filename)
--- a/gammapy/data/event_list.py
+++ b/gammapy/data/event_list.py
@@ -99,6 +99,19 @@ class EventList:
         if format != "gadf":
             raise ValueError(f"{format} is not a valid EventList format.")
 
+        meta_dict = self.table.meta
+
+        if "HDUCLAS1" in meta_dict.keys() and meta_dict["HDUCLAS1"].lower() != "events":
+            raise ValueError("The HDUCLAS1 keyword should be EVENTS for an EventList")
+        else:
+            meta_dict["HDUCLAS1"] = "EVENTS"
+
+
+        if "HDUCLASS" in meta_dict.keys() and meta_dict["HDUCLASS"].lower() != format:
+            raise ValueError("The HDUCLASS keyword should match the format")
+        else:
+            meta_dict["HDUCLASS"] = format.upper()
+
         filename = make_path(filename)
 
         primary_hdu = fits.PrimaryHDU()
--- a/gammapy/data/event_list.py
+++ b/gammapy/data/event_list.py
@@ -96,6 +96,9 @@ class EventList:
             to the gamma-astro-data-formats (GADF) format.
         """
 
+        if format != "gadf":
+            raise ValueError(f"{format} is not a valid EventList format.")
+
         filename = make_path(filename)
 
         primary_hdu = fits.PrimaryHDU()
--- a/gammapy/data/event_list.py
+++ b/gammapy/data/event_list.py
@@ -85,18 +85,17 @@ class EventList:
 
         filename = make_path(filename)
 
-        if gti is None: # save just the events
-            if "extname" not in map(str.lower, self.table.meta.keys()):
-                self.table.meta = {"EXTNAME": "EVENTS"}
-            self.table.write(filename, overwrite=overwrite)
+        primary_hdu = fits.PrimaryHDU()
+        hdu_evt = fits.BinTableHDU(self.table, name='EVENTS')
+        hdu_all = fits.HDUList([primary_hdu, hdu_evt])
 
-        else:
-            assert isinstance(gti, GTI)
-            primary_hdu = fits.PrimaryHDU()
-            hdu_evt = fits.BinTableHDU(self.table, name='EVENTS')
+        if gti is not None:
+            if not isinstance(gti, GTI):
+                raise TypeError('gti must be an instance of GTI')
             hdu_gti = fits.BinTableHDU(gti.table, name="GTI")
-            hdu_all = fits.HDUList([primary_hdu, hdu_evt, hdu_gti])
-            hdu_all.writeto(filename, overwrite=True)
+            hdu_all.append(hdu_gti)
+
+        hdu_all.writeto(filename, overwrite=overwrite)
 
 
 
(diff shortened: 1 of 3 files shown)


gammapy/data/event_list.py
```diff
@@ -74,6 +76,30 @@ def read(cls, filename, **kwargs):
...
+    def write(self, filename, gti=None, overwrite=False):
+        """Write the event list to a FITS file.
+
+        If a GTI object is provided, it is saved into
+        a second extension in the file.
+        """
+
+        filename = make_path(filename)
+
+        if gti is None: # save just the events
```
maxnoe, line 88: I think this could be simplified to just appending the gti to the hdu list if given instead of duplicating the writing code

gammapy/data/event_list.py
```diff
@@ -74,6 +76,30 @@ def read(cls, filename, **kwargs):
...
+                self.table.meta = {"EXTNAME": "EVENTS"}
+            self.table.write(filename, overwrite=overwrite)
+
+        else:
+            assert isinstance(gti, GTI)
+            primary_hdu = fits.PrimaryHDU()
+            hdu_evt = fits.BinTableHDU(self.table, name='EVENTS')
+            hdu_gti = fits.BinTableHDU(gti.table, name="GTI")
+            hdu_all = fits.HDUList([primary_hdu, hdu_evt, hdu_gti])
+            hdu_all.writeto(filename, overwrite=True)
```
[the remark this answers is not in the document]
LauraOlivera: damn how did I miss that - thanks

gammapy/data/event_list.py
maxnoe:
```
hdus = [fits.PrimaryHDU(), fits.BinTableHDU(self.table, name='EVENTS)]

if gti is not None:
    hdus.append(fits.BinTableHDU(gti.table, name="GTI")
    
fits.HDUList(hdus).writeto(filename, overwrite=True)
```

gammapy/data/event_list.py
```diff
@@ -74,6 +76,30 @@ def read(cls, filename, **kwargs):
...
+
+        filename = make_path(filename)
+
+        if gti is None: # save just the events
+            if "extname" not in map(str.lower, self.table.meta.keys()):
+                self.table.meta = {"EXTNAME": "EVENTS"}
+            self.table.write(filename, overwrite=overwrite)
+
+        else:
+            assert isinstance(gti, GTI)
```
maxnoe, line 94: Asserts should not be used in library code, better `raise TypeError('`gti` must be an instance of `GTI`')`

gammapy/data/event_list.py
LauraOlivera: oh this comment didn't appear until after I pushed - but my implementation is essentially the same

maxnoe (dismissed):

registerrier (approved):
Thanks @LauraOlivera ! This looks good.

The docstring misses the list of arguments. 
Also, it would be good to have a `format` argument even if the only possible value is `gadf` as there should be different options later.

gammapy/data/event_list.py
```diff
@@ -74,6 +76,42 @@ def read(cls, filename, **kwargs):
...
+        gti : `~gammapy.data.GTI`
+            Good Time Intervals object to save to the same file.
+            Default is None.
+        overwrite : bool
+            Overwrite existing file?
+        format : str, optional
+            FITS format convention.  By default files will be written
+            to the gamma-astro-data-formats (GADF) format.
+        """
+
```
registerrier, line 98: I would simply add a:
```
if format != "gadf":
    raise ValueError(f"{format} is not a valid EventList format.")
```

registerrier (approved):
Thanks @LauraOlivera ! This looks good. 
 
I would simply add a test to ensure that the format is always "gadf". Possibly this argues against having the argument. Opinion @adonath ?

gammapy/data/event_list.py
LauraOlivera: done

maxnoe:
@LauraOlivera Can you check (test?) that the mandatory keywords for the EVENTS / GTI table are actually there when artificially creating an `EventList` / `GTI`?:

https://gamma-astro-data-formats.readthedocs.io/en/latest/events/events.html#mandatory-header-keywords

LauraOlivera:
Which version of the GADF, though? Because I assume we should stick to the stable release, but that would mean that I can't save HAWC event lists with this method... which was a big reason why I wanted to implement it

LauraOlivera:
I would be fine with adding checks for the headers, but would like to be able to check for `RA_PNT`, `DEC_PNT` or `ALT_PNT`, `AZ_PNT`, not just ra-dec

maxnoe:
@LauraOlivera True. But at least the "uncontroversial" ones, like the `HDUCLASN` `HDUDOC` and `HDUVERS` should be there.

Strictly speaking, as long as there is no new release of the GADF, you cannot do this in complicance with the spec. E.g. you would claim to be 0.2 and have missing headers or you cannot specify a version (adding 0.2 would be wrong).

This is connected deeper in gammapy though. Since gammapy at the moment relies on RA_PNT being there, how do you deal with that for HAWC? It's not only a required keyword in the dataspec, it's a required and used property by gammapy.

LauraOlivera:
> This is connected deeper in gammapy though. Since gammapy at the moment relies on RA_PNT being there, how do you deal with that for HAWC? It's not only a required keyword in the dataspec, it's a required and used property by gammapy.

Well it actually doesn't anymore, at least not for the type of things you need to do for the typical HAWC  data analysis. Basically because of this:
https://github.com/gammapy/gammapy/blob/482b182d31c8e02b662233a4ffbfbbd84d9832a6/gammapy/data/event_list.py#L666-L669

 You can read event lists that don't have those headers. You can use them to fill maps. Even the `.peek()` method works. The `MapDatasetMaker`, for example can take maps as IRFs and background instead of IRFs that need to be projected into the sky. So I think it would actually not go against the current compliance with the spec to allow to write event lists that don't have ra-dec pointing (i.e. are not `pointed_observations`).

maxnoe:
@LauraOlivera Ah ok, nice. That's not what I heard in yesterday's meeting. @registerrier mentioned that the HAWC analysis can only start at the map level.

maxnoe:
> return "RA_PNT" in self.table.meta 

Shouldn't that be `self.table.meta.get("OBS_MODE", "POINTING") == "DRIFT"?

Since the standard explicitly allows filling average values for drift observations no and that would then break this logic.

LauraOlivera:
>  @registerrier mentioned that the HAWC analysis can only start at the map level.

This is in a sense true, since like I mentioned, the current framework for data reduction requires background and IRFs to be given to Gammapy as (whole-sky, typically) maps. But event lists can also be read, and used to make said maps.

> Shouldn't that be self.table.meta.get("OBS_MODE", "POINTING") == "DRIFT"?

This is actually a very good point - that snippet precedes the discussion on `OBS_MODE` we had at the gadf. I can update it, to check for both things so as to not break compatibility for files that don't yet have `OBS_MODE`.

maxnoe:
> But event lists can also be read, and used to make said maps.

That's another issue... It is customary for software to be lenient with standards when reading (e.g. just emit warnings) but be strict when writing.

You could add a version argument (we will need this elsewhere as well) and maybe require the keywords for `version=="0.2"` and not for `version=="dev"`. Having  `"dev"` version support will be great for prototyping and further developing the standard.

maxnoe:
@LauraOlivera I think we can just set those when writing. Not requiring them to be already set. Is there a downside to this?

LauraOlivera:
@maxnoe you mean the version header keywords? I just figured that to call something v0.2 then it would need to be true, meaning we'd need to deal with the rest of the headers too...

maxnoe:
No, I mean `HDUCLASS` and `HDUCLAS1`, instead of checking there existence and raising if they are not there, we could just add them always.

LauraOlivera:
That is what I am doing right now. I am just checking before if they already exist and something is set, that it is the right string. That is the only case in which an error is raised. Arguably I could just always rewrite it, but I thought this might be better, to be explicit if a keyword is not well set by the user.

maxnoe:
Ah, sorry, you are right. I misread the code. You only check that they are what you expect if they are there! 

Yes, that is good.

maxnoe (approved):

registerrier (approved):
Thanks @LauraOlivera !

Issue #3668, opened 2021-11-29
Warning about `is_pointlike` when converting to IRFs to HDU


maxnoe:
**Gammapy version**
0.19.0

**Bug description**

Calling `to_hdu_table` raises a FITS warning that `is_pointlike` is longer than 8 characters and thus not valid for a normal FITS header keyword.

**Expected behavior**

No warning and probably not adding `is_pointlike` to the header at all, since this is covered by the `HDUCLAS` in GADF.

**To Reproduce**

```

In [4]: aeff = EffectiveAreaTable2D.read('joint-crab/data/fact/fact_irf.fits')

In [5]: aeff
Out[5]: <gammapy.irf.effective_area.EffectiveAreaTable2D at 0x7fd1a9d21e90>

In [6]: aeff.to_table_hdu()
/home/maxnoe/.local/anaconda/envs/pyirf/lib/python3.7/site-packages/astropy/io/fits/card.py:268: VerifyWarning: Keyword name 'is_pointlike' is greater than 8 characters or contains characters not allowed by the FITS standard; a HIERARCH card will be created.
  keyword), VerifyWarning)
Out[6]: <astropy.io.fits.hdu.table.BinTableHDU at 0x7fd1aa0e9510>
```

adonath:
Thanks @maxnoe ! Just to confirm: https://github.com/gammapy/gammapy/blob/master/gammapy/irf/core.py#L419 naively copies over the meta data from the Gammapy object. One should translate to valid FITS and GADF keywords instead.

PR #3721, opened 2022-01-24
Remove is_pointlike keyword from irf.to_table method

fabiopintore:
This PR solves the issue #3668. 
The keyword `is_pointlike` is an optional one that can be found in the IRFs. Currently, all the metadata information of the IRF are copied when the `~gammapy.irf.to_table_hdu` method is called, not being consistent with the GADF format. 
The PR adds a filter to remove such a keyword from the output HDU.

### DIFF
--- a/gammapy/irf/core.py
+++ b/gammapy/irf/core.py
@@ -422,6 +422,8 @@ class IRF(metaclass=abc.ABCMeta):
             table.meta["HDUCLAS2"] = spec["hduclas2"]
             if self.is_pointlike:
                 table.meta["HDUCLAS3"] = "POINT-LIKE"
+            if "is_pointlike" in table.meta:
+                del table.meta["is_pointlike"]
             table[spec["column_name"]] = self.quantity.T[np.newaxis]
         else:
             raise ValueError(f"Not a valid supported format: '{format}'")
--- a/gammapy/irf/core.py
+++ b/gammapy/irf/core.py
@@ -422,8 +422,7 @@ class IRF(metaclass=abc.ABCMeta):
             table.meta["HDUCLAS2"] = spec["hduclas2"]
             if self.is_pointlike:
                 table.meta["HDUCLAS3"] = "POINT-LIKE"
-            if "is_pointlike" in table.meta:
-                del table.meta["is_pointlike"]
+            table.meta.pop("is_pointlike", None)
             table[spec["column_name"]] = self.quantity.T[np.newaxis]
         else:
             raise ValueError(f"Not a valid supported format: '{format}'")
--- a/gammapy/irf/tests/test_effective_area.py
+++ b/gammapy/irf/tests/test_effective_area.py
@@ -87,7 +87,7 @@ def test_to_table():
 @requires_data()
 def test_to_table_is_pointlike(aeff):
     hdu = aeff.to_table_hdu()
-    assert(bool(hdu.meta.pop("is_pointlike", None)) == False)
+    assert "is_pointlike" in hdu.header
 
 
 def test_wrong_axis_order():
--- a/gammapy/irf/tests/test_effective_area.py
+++ b/gammapy/irf/tests/test_effective_area.py
@@ -84,6 +84,11 @@ def test_to_table():
     assert hdu.header["TUNIT1"] == aeff.axes["energy_true"].unit
 
 
+@requires_data()
+def test_to_table_is_pointlike(aeff):
+    hdu = aeff.to_table_hdu()
+    assert(bool(table.meta.pop("is_pointlike", None)) == False)
+
 def test_wrong_axis_order():
     energy_axis_true = MapAxis.from_energy_bounds(
         "1 TeV", "10 TeV", nbin=10, name="energy_true"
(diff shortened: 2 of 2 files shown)


gammapy/irf/core.py
```diff
@@ -422,6 +422,8 @@ def to_table(self, format="gadf-dl3"):
             table.meta["HDUCLAS2"] = spec["hduclas2"]
             if self.is_pointlike:
                 table.meta["HDUCLAS3"] = "POINT-LIKE"
+            if "is_pointlike" in table.meta:
```
maxnoe, line 425: This line seems superflous

registerrier (changes requested):
Thanks @fabiopintore . This is good but this misses a test to make sure that the meta does not contain the `is_pointlike` keyword.

gammapy/irf/core.py
adonath: You could also  just use `table.meta.pop("is_pointlike", None)`, where the `None` suppresses the `KeyError` if the keyword is not present.

fabiopintore: Ah great, I didn't know this!

fabiopintore:
Thanks @registerrier @maxnoe @adonath , I've implemented the comments and added a test.

gammapy/irf/tests/test_effective_area.py
```diff
@@ -84,6 +84,11 @@ def test_to_table():
     assert hdu.header["TUNIT1"] == aeff.axes["energy_true"].unit
 
 
+@requires_data()
+def test_to_table_is_pointlike(aeff):
+    hdu = aeff.to_table_hdu()
+    assert(bool(table.meta.pop("is_pointlike", None)) == False)
```
registerrier, line 90: Here you don't test on `hdu`. 
You could simply test `assert "is_pointlike" in hdu.header`

registerrier (changes requested):
Thanks @fabiopintore 

The test is probably not testing what it should be testing. See inline suggestion.

gammapy/irf/tests/test_effective_area.py
maxnoe: Also: assert is a statement, not a function. `assert condition`, not `assert(condition)`

fabiopintore: Sorry @registerrier , I've pushed before reading your suggestions. 
I've pushed again.

gammapy/irf/tests/test_effective_area.py
```diff
@@ -87,7 +87,7 @@ def test_to_table():
 @requires_data()
 def test_to_table_is_pointlike(aeff):
     hdu = aeff.to_table_hdu()
-    assert(bool(hdu.meta.pop("is_pointlike", None)) == False)
+    assert "is_pointlike" in hdu.header
```
maxnoe, line 90: But it should not be in there, right?

fabiopintore: You're totally right. I've forgotten the `not`. However, I don't understand why the test passed with this syntax on my machine while it shouldn't have...

maxnoe: You mean the `bool(header.pop(key, None)) == False)` version? 

if key is not in header, `None` is returned and `bool(None) is False`

fabiopintore: No, that one passes as it's correct. I was meaning the `assert "is_pointlike" in hdu.header`, which shouldn't pass at all...

maxnoe: > No, that one passes as it's correct.

No it's not correct. It would also pass if `is_pointlike` is in the header but with a `falsy` value, e.g. `False` or an empty string.
The only version where this assertion would raise is if `is_pointlike` is in the header, but with a `True` value.

maxnoe: Are you sure you ran the test? It has a `@requires_data` decorator, so it might be skipped.

fabiopintore: > > No, that one passes as it's correct.
> 
> No it's not correct. It would also pass if `is_pointlike` is in the header but with a `falsy` value, e.g. `False` or an empty string. The only version where this assertion would raise is if `is_pointlike` is in the header, but with a `True` value.

Ah you're right. I wrongly considered only the condition of is_pointlike` = True!

fabiopintore: > Are you sure you ran the test? It has a `@requires_data` decorator, so it might be skipped.

Aahaa that's why! I forgot to define `GAMMAPY_DATA` and the test was silently skypped.

maxnoe: Best to just remove the relience on test data and just create a dummy effective area like this:

```

energy_axis = MapAxis.from_energy_bounds('1 TeV', '10 TeV', nbin=3, name='energy_true')
offset_axis = MapAxis.from_bounds(0 * u.deg, 2 * u.deg, nbin=2, name='offset')

aeff = EffectiveAreaTable2D(data=np.ones((3, 2)) * u.m**2, axes=[energy_axis, offset_axis])
```

fabiopintore: Thanks @maxnoe for the suggestion.

registerrier (approved):
Thanks @fabiopintore ! This looks good.

PR #3731, opened 2022-01-25
generate RadMax2D from the RAD_MAX keyword of a point-like IRF

### LINKED ISSUE #3717: Load global RAD_MAX header keyword into RAD_MAX hdu
* The PR #3515 added support for the `RAD_MAX_2D` IRF, stored as the RAD_MAX table in GADF.

* PR #3684  uses this to for energy dependent on region calculation relying on this RAD_MAX

It should be possible to use the mechanisms introduced in #3684 also when using a fixed RAD_MAX that is stored in the `RAD_MAX` header keyword in GADF.

Possibly the best way would be to read this value if present into a `RAD_MAX_2D` object with a single energy / offset bin covering the same min/max values as the corresponding IRF.

cosimoNigro:
in case of a point-like IRF without the RAD_MAX_2D HDU, a RadMax2D object with
a single bin in energy and offset is generated from the RAD_MAX header keyword

Co-authored-by: Maximilian Nöthe <maximilian.noethe@tu-dortmund.de>

Fixes #3717

### DIFF
--- a/gammapy/data/observations.py
+++ b/gammapy/data/observations.py
@@ -35,9 +35,9 @@ class Observation:
         Point spread function
     bkg : `~gammapy.irf.Background3D`
         Background rate model
-    rad_max: `~gammapy.irf.RadMax2D` or `~astropy.units.Quantity`
+    rad_max: `~gammapy.irf.RadMax2D`
         Only for point-like IRFs: RAD_MAX table (energy dependent RAD_MAX)
-        or a single angle (global RAD_MAX)
+        For a fixed RAD_MAX, create a RadMax2D with a single bin.
     gti : `~gammapy.data.GTI`
         Table with GTI start and stop time
     events : `~gammapy.data.EventList`
@@ -50,7 +50,7 @@ class Observation:
     edisp = LazyFitsData(cache=False)
     psf = LazyFitsData(cache=False)
     bkg = LazyFitsData(cache=False)
-    rad_max = LazyFitsData(cache=False)
+    _rad_max = LazyFitsData(cache=False)
     _events = LazyFitsData(cache=False)
     _gti = LazyFitsData(cache=False)
 
@@ -73,10 +73,31 @@ class Observation:
         self.edisp = edisp
         self.psf = psf
         self.bkg = bkg
-        self.rad_max = rad_max
+        self._rad_max = rad_max
         self._gti = gti
         self._events = events
         self.obs_filter = obs_filter or ObservationFilter()
+
+
+    @property
+    def rad_max(self):
+        # prevent circular import
+        from gammapy.irf import RadMax2D
+
+        if self._rad_max is not None:
+            return self._rad_max
+
+        # load once to avoid trigger lazy loading it three times
+        aeff = self.aeff
+        if aeff is not None and aeff.is_pointlike:
+            self._rad_max = RadMax2D.from_irf(aeff)
+            return self._rad_max
+
+        edisp = self.edisp
+        if edisp is not None and edisp.is_pointlike:
+            self._rad_max = RadMax2D.from_irf(self.edisp)
+
+        return self._rad_max
 
     @property
     def available_irfs(self):
--- a/gammapy/data/tests/test_data_store.py
+++ b/gammapy/data/tests/test_data_store.py
@@ -4,6 +4,7 @@ from pathlib import Path
 import pytest
 import numpy as np
 from astropy.io import fits
+import astropy.units as u
 from gammapy.data import DataStore
 from gammapy.utils.scripts import make_path
 from gammapy.utils.testing import requires_data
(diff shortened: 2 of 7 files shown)


maxnoe:
@adonath To know whether we had any point-like IRFs in the datastore, we needed to actually load at least one IRF in the `DataStore.obs` method, this partly removing the lazy loading.

Do you know if there is a better solution for this? Neither obs index nor hdu index contain the information about point-like vs. full-enclosure.

gammapy/data/tests/test_data_store.py
```diff
@@ -173,17 +174,30 @@ def test_hdu_table(self):
 
     def test_observation(self, monkeypatch):
         """Check that one observation can be accessed OK"""
+        # Note: IRF access requires the CALDB env var
+        caldb_path = Path(os.environ["GAMMAPY_DATA"]) / Path("cta-1dc/caldb")
```
maxnoe, line 178: This had to be moved up, because the `DataStore.obs` method already loads the IRFs.

maxnoe:
@adonath @registerrier We are not really sure this is the correct approach. Another one would simply be adding a `rad_max` attribute or property to the IRF classes that can either be a `RadMax2D` or a simple value.

gammapy/irf/rad_max.py
```diff
@@ -24,3 +31,51 @@ class RadMax2D(IRF):
...
+        '''
+        Create a RadMax2D instance from another IRF component.
+
+        This reads the RAD_MAX metadata keyword from the irf and creates
+        a RadMax2D with a single bin in energy and offset using the
+        ranges from the input irf.
+
+        Parameters
+        ----------
+        irf: EffectiveAreaTable2D or EnergyDispersion2D
```
adonath, line 47: Please add backticks to the class names here, such that Sphinx links to them in the docs..

adonath (commented):
Thanks @cosimoNigro and @maxnoe! I have left two minor inline comments for now. 

Concerning the location of the creation of the `RadMax2D` object: 
- Indeed I think one has to access the IRF to check whether it is actually point like. I think this info is not available elsewhere.
- However I guess it could work to implement a property `Observation.rad_max` and move the caching of the `rad_max` to a hidden property `_rad_max`, the same way as is done for the events and GTI. The advantage would be that the `rad_max` is only created when it is requested and no other IRF is loaded beforehand, but one can access the cached one on the `Observation` object.

gammapy/irf/rad_max.py
```diff
@@ -24,3 +31,51 @@ class RadMax2D(IRF):
 
     tag = "rad_max_2d"
     required_axes = ["energy", "offset"]
+
+
+    @classmethod
+    def from_irf(cls, irf: Union[EffectiveAreaTable2D, EnergyDispersion2D]):
```
registerrier, line 37: Note that so far we have not used typing in gammapy.

registerrier (commented):
Thanks @maxnoe . 

I agree with @adonath 's suggestion.

gammapy/irf/rad_max.py
maxnoe: Yes, will remove, this did just help my Editor with autocompletion

maxnoe:
@adonath I implemented the property, seems to work nicely.

maxnoe:
I updated the PR with more tests to make the coverage pass (it was complaining justly that I didn't have tests for the error conditions)

maxnoe:
@adonath @registerrier Could you review?

bkhelifi:
@maxnoe Today, there is no metadata into the obs_table or in the hdu_table specifying which type of IRFs is used for a production (or an observation)? If not, it might be a subject of thinking... as it facilities a lot the 'findability' of data (the tricks that Gammapy does is great, not not ideal in reality)

maxnoe:
> @maxnoe Today, there is no metadata into the obs_table or in the hdu_table specifying which type of IRFs is used for a production (or an observation)? I

No, the information of whether IRFs are point-like or not are not stored in the HDU index if only a global RAD_MAX was used.
In the newest version, there is the option to add the `RAD_MAX_2D` HDU to the HDU index.

Since there is no `HDU` in case of a global `RAD_MAX`, I don't think the HDU index table is the correct place. Also, the information is there, it only needs to be read. 

So I don't think we need a change in the standard here. In general, you don't need `RAD_MAX` before actually reading the IRF, so reading it from the IRF is fine.

maxnoe:
> He does not want to have any type of production on this machine... To be discussed else where.

In my mind, this will work like this:
* A user goes to the observatory website where they can download data
* They select what source / dates / event types / kind of analysis they want
* The data organization system creates the corresponding HDU and OBS index on the fly and prepares a download for the user

I think using a giant obs index or hdu index as the index for a whole archive is impractical. That sort of information should be stored in a proper database, from which the queries can be answered and the index files can be created.

gammapy/data/observations.py
```diff
@@ -73,11 +73,28 @@ def __init__(
...
+
+    @property
+    def rad_max(self):
+        # prevent circular import
+        from gammapy.irf import RadMax2D
+
+        if self._rad_max is not None:
+            return self._rad_max
+
+        if self.aeff is not None and self.aeff.is_pointlike:
```
adonath, line 90: Two short comments on performance / caching here:
- The access to `self.aeff` actually calls into the lazy loading interface, which by default does not cache the object. SO the repeated call to `self.aeff` will actually related the IRF from file everytime. I think the solution here is to just introduce a local variable at the top to only load the IRF once. The same is true for `edisp`
- Once could cache the created `RadMax2D` in `_rad_max` to prevent it is recreated on a 2nd access

I haven't checked which one of the mentioned points are actually relevant for the performance in case the `rad_max` use during data reduction. But might be good to keep in mind, once we start using it.

maxnoe: > The access to self.aeff actually calls into the lazy loading interface, which by default does not cache the object. So the repeated call to self.aeff will actually reload the IRF from file everytime.

I thought just checking `self.aeff is None` only verifies that there is indeed a `LazyHDU` and does not load it. Is that wrong?

maxnoe: Ah, sure. It calls `get`, which loads it..

maxnoe: Ok, updated to cache the `_rad_max` and only load the `aeff` and `edisp` once

gammapy/data/observations.py
```diff
@@ -73,11 +73,31 @@ def __init__(
...
+        from gammapy.irf import RadMax2D
+
+        if self._rad_max is not None:
+            return self._rad_max
+
+        # load once to avoid trigger lazy loading it three times
+        aeff = self.aeff
+        if aeff is not None and aeff.is_pointlike:
+            self._rad_max = RadMax2D.from_irf(aeff)
+
```
registerrier, line 94: Why is `self._rad_max` created twice? Shouldn't it be returned once created?

registerrier (approved):
Thanks @maxnoe . This looks good.

I have left just one inline comment.

gammapy/data/observations.py
maxnoe: yes, sorry, there is a return statement missing here.

maxnoe: fixed

registerrier (approved):
Thanks @maxnoe !

adonath (approved):
No further comments from my side either...

Issue #3692, opened 2021-12-21
Show linked parameters when printing model


adonath:
Right now linked parameters are not shown in the printout of the model:
```python3
from gammapy.modeling.models import (
    SkyModel,
    FoVBackgroundModel,
    Models,
    PointSpatialModel,
    PowerLawSpectralModel
)

# define model
point = PointSpatialModel(lon_0="0 deg", lat_0="0 deg", frame="galactic")
pwl = PowerLawSpectralModel()

crab = SkyModel(pwl, point, name="crab")

bkg_d1 = FoVBackgroundModel(dataset_name="fermi-dataset")
bkg_d2 = FoVBackgroundModel(dataset_name="cta-dataset")

# link background norm parameters
bkg_d1.spectral_model.norm = bkg_d2.spectral_model.norm

models = Models([crab, bkg_d1, bkg_d2])
print(models)
```
Gives:
```
Models

Component 0: SkyModel

  Name                      : crab
  Datasets names            : None
  Spectral model type       : PowerLawSpectralModel
  Spatial  model type       : PointSpatialModel
  Temporal model type       : 
  Parameters:
    index                   :      2.000   +/-    0.00             
    amplitude               :   1.00e-12   +/- 0.0e+00 1 / (cm2 s TeV)
    reference    (frozen)   :      1.000       TeV         
    lon_0                   :      0.000   +/-    0.00 deg         
    lat_0                   :      0.000   +/-    0.00 deg         

Component 1: FoVBackgroundModel

  Name                      : fermi-dataset-bkg
  Datasets names            : ['fermi-dataset']
  Spectral model type       : PowerLawNormSpectralModel
  Parameters:
    norm                    :      1.000   +/-    0.00             
    tilt         (frozen)   :      0.000                   
    reference    (frozen)   :      1.000       TeV         

Component 2: FoVBackgroundModel

  Name                      : cta-dataset-bkg
  Datasets names            : ['cta-dataset']
  Spectral model type       : PowerLawNormSpectralModel
  Parameters:
    norm                    :      1.000   +/-    0.00             
    tilt         (frozen)   :      0.000                   
    reference    (frozen)   :      1.000       TeV         

```
While this is a simple feature, I think technically it is not as simple to achieve, because there is no "primary" owner of a model parameter and on the parameter level we don't have access to the parent model. A similar thing applies to sharing e.g. spatial and spectral models of sources. Let's collect a few approaches how to possibly make this work in this issue.

[...]

PR #3732, opened 2022-01-25
Show linked parameters when printing model

QRemy:
Solves #3692.
-Move the code that update linked parameters labels into a function and run it  not only in .to_dict() but also in .__str__().
-display link label in .__str__()

### DIFF
--- a/gammapy/modeling/parameter.py
+++ b/gammapy/modeling/parameter.py
@@ -23,7 +23,12 @@ def _get_parameters_str(parameters):
         else:
             value_format, error_format = "{:10.3f}", "{:7.2f}"
 
-        line = "\t{:12} {:11}: " + value_format + "\t {} {:<12s}\n"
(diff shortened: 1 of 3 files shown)


gammapy/modeling/parameter.py
```diff
@@ -23,7 +23,12 @@ def _get_parameters_str(parameters):
         else:
             value_format, error_format = "{:10.3f}", "{:7.2f}"
 
-        line = "\t{:12} {:11}: " + value_format + "\t {} {:<12s}\n"
+        line = "\t{:21} {:8}: " + value_format + "\t {} {:<12s}\n"
```
QRemy, line 26: why do we use this `_get_parameters_str` ?
It would be simpler to just show `parameters.to_table().pprint(max_lines=-1, max_width=-1)`.

registerrier (approved):
Thanks @QRemy . This looks good!

Issue #3718, opened 2022-01-24
Implement a reflected regions background maker using fixed positions for the OFF regions


cosimoNigro:
When using energy-dependent sizes of the ON and OFF regions it is customary to assume that the OFF regions sit at the same positions in the different energy bins.

The `ReflectedRegionsBackgroundMaker` instead tries to fit as many OFF regions as possible, when estimating the OFF counts. 
Therefore, when using the newly-implemented energy-dependent spectrum extraction (#3684) - that relies on `ReflectedRegionsBackgroundMaker` - different OFF regions are used in different energy bins.

It would be good to have a background maker with the same positions of the OFF regions.

PR #3733, opened 2022-01-25
Add new RegionsFinder that uses a fixed number of regions symmetrically distributed on the circle

cosimoNigro:
Added another finder to the background makers returning symmetric OFF positions.
Added a minimal test.

Co-authored-by: Maximilian Nöthe <maximilian.noethe@tu-dortmund.de>

Fixes: #3718

### DIFF
--- a/gammapy/makers/background/reflected.py
+++ b/gammapy/makers/background/reflected.py
@@ -11,7 +11,12 @@ from gammapy.maps import RegionGeom, RegionNDMap, WcsGeom, WcsNDMap
 from ..core import Maker
 from ..utils import make_counts_off_rad_max
 
-__all__ = ["ReflectedRegionsFinder", "ReflectedRegionsBackgroundMaker"]
+__all__ = [
+    "ReflectedRegionsBackgroundMaker",
+    "ReflectedRegionsFinder",
+    "RegionsFinder",
+    "WobbleRegionsFinder",
+]
 
 log = logging.getLogger(__name__)
 
@@ -19,11 +24,21 @@ FULL_CIRCLE = Angle(2 * np.pi, "rad")
 
 
 class RegionsFinder(metaclass=ABCMeta):
-    '''Baseclass for regions finders'''
+    '''Baseclass for regions finders
+
+
+    Parameters
+    ----------
+    binsz : `~astropy.coordinates.Angle`
+        Bin size of the reference map used for region finding.
+    '''
+    def __init__(self, binsz=0.01 * u.deg):
+        '''Create a new RegionFinder'''
+        self.binsz = Angle(binsz)
 
     @abstractmethod
     def run(self, region, center, exclusion_mask=None):
-        """Find regions to calculate background counts.
+        """Find reflected regions.
 
         Parameters
         ----------
@@ -32,7 +47,8 @@ class RegionsFinder(metaclass=ABCMeta):
         center : `~astropy.coordinates.SkyCoord`
             Rotation point
         exclusion_mask : `~gammapy.maps.WcsNDMap`, optional
-            Exclusion mask
+            Exclusion mask. Regions intersecting with this mask will not be
+            included in the returned regions.
 
         Returns
         -------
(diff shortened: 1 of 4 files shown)


maxnoe:
magic_off_regions

maxnoe:
We should review / merge #3737 first and then add this PR with the new api

maxnoe:
Depends on #3734, so I will rebase when that is merged

adonath:
#3737 is merged please rebase this PR

maxnoe:
Done

gammapy/makers/background/reflected.py
```diff
@@ -41,6 +146,25 @@ def run(self, region, center, exclusion_mask=None):
...
             WCS for the determined regions
         """
+        reference_geom = self._create_reference_geometry(region, center)
+        center_pixel = self._get_center_pixel(center, reference_geom)
+
+        region_pix = self._get_region_pixels(region, reference_geom)
+        excluded_pixels = self._get_excluded_pixels(reference_geom, exclusion_mask)
+
+        n_positions = self.n_off_regions + 1
+        increment = FULL_CIRCLE / n_positions
```
registerrier, line 156: You could also loop over increments:
```
rotation_angles = np.linspace(0*u.rad, FULL_CIRCLE, n_positions, endpoint=False)
for angle in rotation_angles:
    region_test = region_pix.rotate(center_pixel, angle)
```

gammapy/makers/background/reflected.py
```diff
@@ -57,7 +181,7 @@ class ReflectedRegionsFinder(RegionsFinder):
     background estimate for an IACT observation using the reflected regions
     method, see also `~gammapy.makers.ReflectedRegionsBackgroundMaker`
 
-    Parameters
+    Attributes
```
registerrier, line 184: We don't use `Attributes` but rather `Parameters` for `__init__` docstring.

gammapy/makers/background/reflected.py
```diff
@@ -342,19 +419,24 @@ def make_counts_off(self, dataset, observation):
...
+            Rotation point
+        exclusion_mask : `~gammapy.maps.WcsNDMap`, optional
+            Exclusion mask. Regions intersecting with this mask will not be
+            included in the returned regions.
 
         Returns
         -------
-        dataset_on_off : `SpectrumDatasetOnOff`
-            On off dataset.
+        regions : list of `SkyRegion`
```
registerrier, line 436: The return types have not changed here.

registerrier (changes requested):
Thanks @cosimoNigro . This looks good! I have left minor comments inline.

A possibly more serious issue is the absence of checks that regions do not overlap. If the wobble offset is too small and/or if the number of regions is too large, at some point you will end up having overlapping OFF regions. I think this needs to be checked. 
Maybe the `_get_angle_range` method from the `ReflectedRegionsFinder` can be of use?

gammapy/makers/background/reflected.py
maxnoe: But this is a class docstring. You either have to use `Attributes` in the class docstring or `Parameters` in the `__init__` docstring. Using `Parameters` in the class docstring is wrong.

registerrier: So far, we have been using the class docstring for the `__init__` docstring, but we have not used `Attributes` at all in the documentation. Our dev doc here is clearly unfinished on the specific point of class docstrings. 
As far as I understand, using `Parameters` in the class docstring is not wrong in general. It is one possible option.  In astropy, you actually find the two forms, but mostly `__init__` `Parameters` in the class docstring.

gammapy/makers/background/reflected.py
maxnoe: Your proposal is bugged, since it starts at 0, which is the source position.

maxnoe: I think the loop is very clear and easy to reason about.

gammapy/makers/background/reflected.py
maxnoe: Ok

gammapy/makers/background/reflected.py
maxnoe: fixed

gammapy/makers/background/reflected.py
maxnoe: fixed

adonath:
@maxnoe Looks good! 

I guess this could be implemented as a special case in https://github.com/gammapy/gammapy/blob/master/gammapy/maps/region/geom.py#L292

In https://github.com/gammapy/gammapy/blob/master/gammapy/maps/region/geom.py#L371, we internally use some oversampling to get a higher precision for not fully contained pixels. This might be useful here as well.

gammapy/makers/background/reflected.py
registerrier: > Your proposal is bugged, since it starts and ends at 0 / 360°, which is the source position.

It misses the `[1:]` indeed, but it does not contain the end point. But it does not matter.

gammapy/makers/background/tests/test_reflected.py
```diff
@@ -246,3 +247,50 @@ def test_reflected_bkg_maker_no_off_background(reflected_bkg_maker, observations
...
+        position_angle=source_angle + 90 * u.deg,
+    )
+    exclusion_region = CircleSkyRegion(pos, Angle(0.2, "deg"))
+    geom = WcsGeom.create(skydir=pos, binsz=0.01, width=10.0)
+    exclusion_mask = ~geom.region_mask([exclusion_region])
+
+    finder = WobbleRegionsFinder(n_off_regions)
+    regions, _ = finder.run(on_region, pointing, exclusion_mask)
+
+    print(exclusion_region.center)
```
registerrier, line 293: Please remove the print statements.

gammapy/makers/background/tests/test_reflected.py
```diff
@@ -246,3 +247,50 @@ def test_reflected_bkg_maker_no_off_background(reflected_bkg_maker, observations
...
+        center=SkyCoord.from_name("Crab"),
+        radius=0.15 * u.deg,
+    )
+    pointing = on_region.center.directional_offset_by(
+        separation=0.4 * u.deg,
+        position_angle=180 * u.deg + source_angle,
+    )
+
+    assert u.isclose(pointing.position_angle(on_region.center).to(u.deg), source_angle, rtol=0.05)
+    n_off_regions = 3
```
registerrier, line 265: What happens if e.g. `n_off_regions = 15` ? 
I suppose one wants non-overlapping regions, so that the `WobbleRegionsFinder` should fail or return no regions.

This should be tested.

maxnoe: Do you think it should only be a warning or an error to have overlapping regions? Having them overlap is not wrong per se but it violates the poisson assumption made later when fitting since the counts in the single regions are not independent from each other anymore.

So raising an error would probably be the safest option.

maxnoe: Ok, looking at the current code, it logs a warning and sets the counts to None, so maybe we return no (valid) regions found from the finder.

maxnoe:
@registerrier I added a check for overlapping regions using the mask representation of each region. I don't know if that's the best way to check, but it was the first idea I had and it seems to work nicely.

gammapy/makers/background/reflected.py
```diff
@@ -41,6 +155,35 @@ def run(self, region, center, exclusion_mask=None):
...
+        regions = []
+        for i in range(1, n_positions):
+            angle = i * increment
+            region_test = region_pix.rotate(center_pixel, angle)
+
+            if exclusion_mask is None or not np.any(region_test.contains(excluded_pixels)):
+                regions.append(region_test)
+
+
+        # check for overlaps
```
adonath, line 176: This could use https://github.com/gammapy/gammapy/blob/master/gammapy/maps/wcs/ndmap.py#L547 to simplify the code...

maxnoe: mmh. I don't think so. That is for testing if a sky region is included in a sky mask. We are still in pixel space here and the code is pretty straight forward, right? Converting every region repeatedly into the list of pixel coordinates seems also more expensive than calculating the masks once and then checking logical and

adonath: Sorry, my bad. This indeed checks for the overlap of the regions themselves and not for the overlap with the exclusion mask.

adonath: In this case maybe move the check inside the other for loop? I think it's sufficient to check whether the `region_test`, overlaps with any of the previous regions and then break?

adonath: One could also extend the excluded pixels by the previously found regions, but that's maybe a bit too clever...

maxnoe: Yes I thought about that, but like that you will end up with a single region that might overlap with the on region. Actually we should probably just be able to check if the first region overlaps with the on region, right?

adonath: > Actually we should probably just be able to check if the first region overlaps with the on region, right?

Hm, if the offset is really large and the number of requested off regions high, one could still end up with overlapping off regions, that don't overlap with the on region?

maxnoe: I dont think so, since the basis of the code is that the off regions and the on region are symmetrically distributed. If I don't tell you the on region, you should not be able to tell.

adonath: Yes, what I said before doesn't make sense. I had the wrong geometry of the problem in mind...

The symmetry argument is key here. In this case I agree it should be sufficient to check whether the first off region overlaps with the on region and that's it.

registerrier (approved):
Thanks @maxnoe ! 
This looks good! Indeed, one could possibly only check the overlap of the first region with the ON region. But it requires that it is not excluded.

Issue #3677, opened 2021-12-03
Set default required_irfs to None on DataStore.get_observations()


registerrier:
With the addition of the `required_irfs` argument on `Datastore.get_observations()` some observations will be skipped if the irf is missing. By default, the argument is set to `"all"` which requires aeff, psf, edisp and bkg to be present.

This is not adapted to a number of situations e.g. pointlike observations, or `DataStore` devoid of bkg IRF.
While a warning is logged for each observation skipped, users might be confused. It is safer to set the default to `None` and properly document the usage of this argument with a how-to and inside one or two notebooks.

This problem causes issue number 2 reported in this discussion #3673

PR #3735, opened 2022-01-25
Cleanup HDUs required in Datastore.get_observations()

QRemy:
Resolve issues #3666 and #3677
- replace `required_irf` argument by `required_hdu` in Datastore.get_observations()
- set default requirement to None (all hdu will be loaded if possible and a warning is emitted otherwise)
- remove `all` option and add two options `full-enclosure` and `point-like` as suggested in https://github.com/gammapy/gammapy/issues/3666#issuecomment-980095022

### DIFF
--- a/gammapy/data/data_store.py
+++ b/gammapy/data/data_store.py
@@ -14,6 +14,12 @@ from .observations import Observation, ObservationChecker, Observations
 
 __all__ = ["DataStore"]
 
+REQUIRED_IRFS = {
+    "full-enclosure": ["aeff", "edisp", "psf", "bkg"],
+    "point-like": ["aeff", "edisp"],
+    "all-optional":  ["aeff", "edisp", "psf", "bkg", "rad_max"],
+}
+
 log = logging.getLogger(__name__)
 log.setLevel(logging.INFO)
 
@@ -317,15 +323,11 @@ class DataStore:
             Container holding a list of `~gammapy.data.Observation`
         """
 
-        all_hdu = ["aeff", "edisp", "psf", "bkg", "rad_max"]
+        all_hdu = REQUIRED_IRFS["all-optional"]
         is_all_optinal = required_irf == "all-optional"
-        
-        if required_irf == "full-enclosure":
-            required_irf = ["aeff", "edisp", "psf", "bkg"]
-        elif required_irf == "point-like":
-            required_irf = ["aeff", "edisp"]
-        elif required_irf == "all-optional":
-            required_irf = all_hdu
+
+        if isinstance(required_irf, str) and required_irf in REQUIRED_IRFS.keys():
+            required_irf = REQUIRED_IRFS[required_irf]
 
         if not set(required_irf).issubset(all_hdu):
             difference = set(required_irf).difference(all_hdu)
--- a/gammapy/data/data_store.py
+++ b/gammapy/data/data_store.py
@@ -15,6 +15,7 @@ from .observations import Observation, ObservationChecker, Observations
 __all__ = ["DataStore"]
 
 log = logging.getLogger(__name__)
+log.setLevel(logging.INFO)
 
 
 class DataStore:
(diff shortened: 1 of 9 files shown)


gammapy/data/data_store.py
```diff
@@ -242,32 +248,42 @@ def get_observations(self, obs_id=None, skip_missing=False, required_irf="all"):
...
+            * `events` : Events
+            * `gti` :  Good time intervals
             * `aeff` : Effective area
             * `bkg` : Background
             * `edisp`: Energy dispersion
             * `psf` : Point Spread Function
-            By default, all the IRFs are required.
+            * `rad_max` : Maximal radius
+            * `full-containment` :["events", "gti", "aeff", "edisp", "psf", "bkg"]
+            * `point-source` : ["events", "gti", "aeff", "edisp", "rad_max"]
```
maxnoe, line 263: `rad_max` is not required for `point-like` (it should be `point-like`, not `point-source`!) since the `RAD_MAX` header keyword can also be used.

See this issue: #3717  and corresponding PR: #3731

QRemy: thanks modified

gammapy/data/data_store.py
```diff
@@ -273,9 +273,9 @@ def get_observations(self, obs_id=None, skip_missing=False, required_hdu=None):
         point_like = ["events", "gti", "aeff", "edisp"]
         available_hdu = ["events", "gti", "aeff", "edisp", "psf", "bkg", "rad_max"]
         
-        if required_hdu == "full_containment":
+        if required_hdu == "full-enclosure":
             required_hdu = full_containment
```
maxnoe, line 277: Here it's still containmnet

gammapy/data/data_store.py
```diff
@@ -226,14 +229,17 @@ def obs(self, obs_id):
         kwargs = {"obs_id": int(obs_id)}
         kwargs["obs_info"] = table_row_to_dict(row)
 
-        hdu_list = ["events", "gti", "aeff", "edisp", "psf", "bkg", "rad_max"]
+        if required_hdu is None or required_hdu==[]:
+            hdu_list = ["events", "gti", "aeff", "edisp", "psf", "bkg", "rad_max"]
```
AtreyeeS, line 233: So if the `required_hdu` is `None`, is tries to load *all* of them? And raises warning for the missing ones?

maxnoe: That's very contra-intuitive. Requiring `None` or an empty list should not be treated as `all`, since it is semantically the exact opposite.

QRemy: @AtreyeeS yes
@maxnoe I interpret None required as all optional

AtreyeeS: I think having an info table will indeed be very helpful. It seems that in the default case, the user will end up with a bunch of warnings (eg: either a `rad_max` or `psf` warning will be there for *all*) and then if you have say, an `edisp` missing somewhere, it becomes impossible to find. And `MapMaker` breaks later on.

Maybe @maxnoe suggestion of having `full_enclosure` as the default is better?

gammapy/data/data_store.py
```diff
@@ -226,14 +237,19 @@ def obs(self, obs_id):
         kwargs = {"obs_id": int(obs_id)}
         kwargs["obs_info"] = table_row_to_dict(row)
 
-        hdu_list = ["events", "gti", "aeff", "edisp", "psf", "bkg", "rad_max"]
-
-        for hdu in hdu_list:
+        if required_irf == "full-enclosure":
+            hdus = ["events", "gti", "aeff", "edisp", "psf", "bkg"]
```
adonath, line 241: I think it could be good to define a dict at the top of the file, like:

```python3
REQUIRED_HDUS = {
    "full-enclosure": ...,
    "point-like": ...,
    "all": ...,
}
```
To have the definite of those in a central place.

gammapy/data/observations.py
```diff
@@ -79,18 +79,17 @@ def __init__(
...
     @property
-    def available_irfs(self):
-        """Which irfs are available"""
-        available_irf = []
-
-        for irf in ["aeff", "edisp", "psf", "bkg"]:
-            available = self.__dict__.get(irf, False)
-            available_hdu = self.__dict__.get(f"_{irf}_hdu", False)
-
+    def available_hdus(self):
```
adonath, line 82: Why not keep both `available_irfs` and `available_hdus`? I think there might be use-case for both...

registerrier (dismissed):

I just left some small inline comments.

registerrier (approved):
Thanks @QRemy ! 
This looks good. We will just need to document the new behavior in future PRs.

PR #3737, opened 2022-01-25
Change API of ReflectedRegionsFinder, fixes #3734

### LINKED ISSUE #3734: Make `RegionFinder` modular for `ReflectedRegionsBackgroundMaker`

As discussed e.g. here #3296 , there are multiple strategies to find reflected off regions.

My proposal would be that the `ReflectedRegionsBackgroundMaker` gets a `region_finder` argument instead of the arguments for `ReflectedRegionsFinder` and that the API for a generic RegionsFinder should be:

```
class RegionsFinder:
     def __init__(**options_for_finder):
         ...

    def run(self, region, center):
         return list_of_regions
```

Like this, the `ReflectedRegionsBackgroundMaker` can receive a number of `Finder` implementations and call them when needed for the specific combination of `region` and `center`.

Currently, the `region` and `center` arguments are given to the `__init__` method of `ReflectedRegionsFinder`, although these are only used once in the `run` method. The rest of the class is a more or less complicated caching structure, that is not really needed, since `run` clears the cache. So only inside `run`, things are cached.

This could be more simply solved with just having local variables in run eliminating the need for storing `region` and `center` in the class.

maxnoe:
The ReflectedRegionsFinder now takes configuration options
as `__init__` arguments but the `region` and `center` are
now parameters to the `run` method.

This enables creating a `ReflectedRegionsFinder` and calling it
for each region / center in the `ReflectedRegionsBackgroundMaker`
instead of creating new finders over and over.

The run method returns the found regions and their wcs object.

The reliance on caching was removed and the code simplified by that.

Co-authored-by: Axel Donath <axel.donath@mpi-hd.mpg.de>

**Description**

This pull request ...

**Dear reviewer**

### DIFF
--- a/gammapy/makers/background/reflected.py
+++ b/gammapy/makers/background/reflected.py
@@ -1,10 +1,11 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 import logging
 import numpy as np
+from abc import ABCMeta, abstractmethod
+
 from astropy import units as u
 from astropy.coordinates import Angle
-from astropy.utils import lazyproperty
-from regions import PixCoord, PointSkyRegion, CircleSkyRegion
+from regions import PixCoord, PointSkyRegion
 from gammapy.datasets import SpectrumDatasetOnOff
 from gammapy.maps import RegionGeom, RegionNDMap, WcsGeom, WcsNDMap
 from ..core import Maker
@@ -14,8 +15,33 @@ __all__ = ["ReflectedRegionsFinder", "ReflectedRegionsBackgroundMaker"]
 
 log = logging.getLogger(__name__)
 
+FULL_CIRCLE = Angle(2 * np.pi, "rad")
 
-class ReflectedRegionsFinder:
+
+class RegionsFinder(metaclass=ABCMeta):
+    '''Baseclass for regions finders'''
+
+    @abstractmethod
+    def run(self, region, center):
+        """Find regions to calculate background counts.
+
+        Parameters
+        ----------
+        region : `~regions.SkyRegion`
+            Region to rotate
+        center : `~astropy.coordinates.SkyCoord`
+            Rotation point
+
+        Returns
+        -------
+        regions : list of `SkyRegion`
+            Reflected regions
+        wcs: `~astropy.wcs.WCS`
+            WCS for the determined regions
+        """
+
+
+class ReflectedRegionsFinder(RegionsFinder):
     """Find reflected regions.
 
     This class is responsible for placing :ref:`region_reflected` for a given
@@ -31,10 +57,6 @@ class ReflectedRegionsFinder:
 
     Parameters
     ----------
-    region : `~regions.SkyRegion`
-        Region to rotate
-    center : `~astropy.coordinates.SkyCoord`
-        Rotation point
     angle_increment : `~astropy.coordinates.Angle`, optional
         Rotation angle applied when a region falls in an excluded region.
     min_distance : `~astropy.coordinates.Angle`, optional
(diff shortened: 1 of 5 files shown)


gammapy/makers/background/reflected.py
```diff
@@ -14,8 +15,33 @@
...
+        """Find regions to calculate background counts.
+
+        Parameters
+        ----------
+        region : `~regions.SkyRegion`
+            Region to rotate
+        center : `~astropy.coordinates.SkyCoord`
+            Rotation point
+
+        Returns
```
registerrier, line 35: Maybe directly returning a list of `RegionGeom` objects would avoid returning `SkyRegion` and a `WCS`

gammapy/makers/background/reflected.py
```diff
@@ -140,134 +157,112 @@ def region_angular_size(self):
...
-        Minimal distance between two consecutive reflected regions
-    min_distance_input : `~astropy.coordinates.Angle`, optional
-        Minimal distance from input region
-    max_region_number : int, optional
-        Maximum number of regions to use
-    exclusion_mask : `~gammapy.maps.WcsNDMap`, optional
-        Exclusion mask
-    binsz : `~astropy.coordinates.Angle`
-        Bin size of the reference map used for region finding.
+    regions_finder: RegionsFinder
```
registerrier, line 252: Maybe specify list of possible choices.
Also specify default choice.

registerrier: The `kwargs` argument should be listed in the parameters list. 
And possibly an explicit link to the `ReflectedRegionsFinder` if it is the default.

registerrier: Shouldn't the `mask` be an argument of the `Maker` itself since it will be an argument of all `RegionsFinder`?

registerrier (commented):
Thanks @maxnoe . 

Introducing a general `RegionsFinder` is a good idea. 
- There is now an issue with the `Maker` name, which does not reflect on what it actually does. Shouldn't there be a more general `OFFRegionBackgroundMaker` that would take any `RegionsFinder` and a inherited `ReflectedRegionsBackgroundMaker` that would use specifically the reflected regions finder. This would be a clearer API scheme and wouldn't break the API.
- Is the new scheme configurable in `AnalysisConfig`. I assume this relates to the first point.
- Is there performance impact due to the removal of cached properties?

gammapy/makers/background/reflected.py
maxnoe: I discussed this with @adonath yesterday. Unfortunately, creating a `RegionGeom` created problems, as the number of found regions was not knowable by the maker then. 

The number of regions is however required for the alpha calculation.

gammapy/makers/background/reflected.py
maxnoe: Yes, that's a good point. Probably the `exclusion_mask` should also be moved to the `run` method.

maxnoe: Done. How that cleaned up the test was a good sign that again, this is the preferable API

gammapy/makers/background/reflected.py
adonath: One could change the alpha calculation to the ratio of solid angles, but this can be easily done in a follow up PR.

maxnoe: Is there an easy way to get this?

maxnoe:
```

In [4]: geom.solid_angle()
NotImplementedError                       Traceback (most recent call last)
~/Projects/gammapy/gammapy/makers/background/reflected.py in <module>
----> 1 geom.solid_angle()

~/Projects/gammapy/gammapy/maps/region/geom.py in solid_angle(self)
    302             raise ValueError("Region definition required.")
    303 
--> 304         area = self.region.to_pixel(self.wcs).area
    305         solid_angle = area * proj_plane_pixel_area(self.wcs) * u.deg ** 2
    306         return solid_angle.to("sr")

~/.local/anaconda/envs/gammapy-dev/lib/python3.7/site-packages/regions/core/compound.py in area(self)
    162     @property
    163     def area(self):
--> 164         raise NotImplementedError
    165 
    166     def rotate(self, center, angle):

NotImplementedError: 

```

adonath: Ok, this would need to be implemented in `regions`. So we just keep as is for now...

adonath:
Hm, I'm wondering why the docs build passes. There is one example script here https://github.com/gammapy/gammapy/blob/master/docs/makers/make_reflected_regions.py, that requires adaption to the new API.

maxnoe:
@adonath It seems the plot is just silently dropped when the script fails to exectute

adonath (approved):
Thanks @maxnoe, no further comments from my side.

Issue #3619, opened 2021-11-19
Add Python 3.9 and 3.10 to CI setup


adonath:
Reminder issue, see title.

Issue #3614, opened 2021-11-18
Speed up CI builds and tests


adonath:
This is a reminder issue to speed up the execution of the CI builds and tests, maybe by changing to `mambaforge` as a base environment or relying on pip wheels for the setup.

adonath:
I think there are two separate issues here:
- Speed up the setup time
- Speed up the actual run time of tests and tutorials

facero:
Talking about mamba for installation for users I noticed that `micromamba` has been developped in particular for CI jobs.
Micromamba is a pure C++-based CLI, self-contained in a single-file executable.
See here : [micromamba](https://mamba.readthedocs.io/en/latest/user_guide/micromamba.html#micromamba).
Maybe this could be worth investigating.

maxnoe:
Current durations (after merging #3699):

![Screenshot_20220124_153033](https://user-images.githubusercontent.com/5488440/150801984-2947b5c1-9480-4524-bae1-c730c3ee6990.png)
![Screenshot_20220124_153000](https://user-images.githubusercontent.com/5488440/150801990-d5fa6f25-2702-4dc9-b9ec-6d9a3cad5a1e.png)

adonath:
And just for reference here are the 10 slowest tests (measured locally on my machine):

```
22.20s call     gammapy/scripts/tests/test_download.py::test_cli_download_datasets
15.00s call     gammapy/estimators/points/tests/test_sed.py::test_run_map_pwl
11.76s call     gammapy/analysis/tests/test_analysis.py::test_analysis_1d
7.10s call     gammapy/estimators/points/tests/test_profile.py::test_radial_profile
5.98s call     gammapy/estimators/points/tests/test_profile.py::test_profile_content
5.23s call     gammapy/estimators/map/tests/test_ts.py::test_compute_ts_map
4.21s call     gammapy/estimators/points/tests/test_profile.py::test_radial_profile_one_interval
3.56s call     gammapy/maps/hpx/tests/test_ndmap.py::test_smooth[disk]
3.35s call     gammapy/estimators/points/tests/test_lightcurve.py::test_lightcurve_estimator_map_datasets
3.00s call     gammapy/makers/tests/test_safe.py::test_safe_mask_maker
```
The total runtime was ~200s. I think this is still acceptable, however I'm using a recent M1 machine so speeding up the test by ~50% would be useful. 

[...]

PR #3739, opened 2022-01-25
Improve CI

Bultako:
This PR intends to improve the CI as follows:

- speeds CI splitting the conda-build step into its own run
- speeds CI using [micromamba](https://mamba.readthedocs.io/en/latest/user_guide/micromamba.html) reducing the environments creation and setup to half of the time
- tests oldest supported versions of dependencies defined in `setup.cfg` in a Linux Python 3.7 env
- adds test for a Linux Python 3.9 environment

Still some fixes need to be done because the tests for the oldest supported versions of dependencies do not consider `astropy` and `numpy`

Also we could add a run to test development versions of some of the dependencies, but it is likely that this additional run will not help in speeding the CI if the the dev versions of dependencies would have to be built from source and installed via `pip install git+https://` 

Related issues:
- https://github.com/gammapy/gammapy/issues/2270
- https://github.com/gammapy/gammapy/issues/3614
- https://github.com/gammapy/gammapy/issues/3619

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -77,11 +77,11 @@ jobs:
         if: contains(matrix.name, 'all-deps')
         run: |
           gammapy download datasets
-          export GAMMAPY_DATA=/home/runner/work/gammapy/gammapy/gammapy-datasets
+          export GAMMAPY_DATA=./gammapy-datasets
           make test-cov
           codecov -X gcov
       - name: test simple
-        if: matrix.name!='Linux python 3.7 all-deps'
+        if: contains(matrix.name, 'core-deps')
         run: |
           make test
   notebooks:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,7 +39,7 @@ jobs:
           - name: Windows python 3.7 core-deps
             os: windows-latest
             python: 3.7
-          - name: Mac python 3.7 all-deps
+          - name: Mac python 3.7 core-deps
             os: macos-latest
     steps:
       - name: checkout repo
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -110,8 +110,8 @@ jobs:
       - name: test notebooks
         run: |
           make test-nb
-  others:
-    name: Linux python 3.7 others all-deps
+  sphinx:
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -162,18 +162,19 @@ jobs:
...
+          extra-specs: |
+            python=3.7
       - name: install dependencies
         # oldest supported versions according to setup.cfg
         # TODO set astropy and numpy versions
         run: |
-          pip install astropy click==7.0 numpy pydantic==1.0 pyyaml==5.1 regions==0.5 scipy==1.2
-          pip install pytest pytest-astropy requests tqdm
+          mamba install click==7.0 pydantic==1.0 pyyaml==5.1 regions==0.5 scipy==1.2 astropy==4.3 numpy==1.18.1
+          mamba install requests tqdm
```
maxnoe, line 177: why two commands here?

Bultako: fixed with micromaba actions, though still in draft stage..

Bultako:
I have found the minimal versions required for the core packages to pass the CI tests and used them in a github actions job to test oldest packages supported.

```
            click==7.0
            pydantic==1.0
            pyyaml==5.1
            regions==0.5
            scipy==1.2
            astropy==4.3
            numpy==1.18.1
```

ok, there are still very few tests failing but I hope these could be solved somehow, all of them are related with testing the string content of a specific catalog...  https://github.com/gammapy/gammapy/runs/4971854151?check_suite_focus=true#step:6:145

@adonath do you have an idea on how this could be solved?

adonath:
@Bultako I have seen this error before and I think it is related to how a masked value is represented in Astropy table. I'll check what we can do about this...

Bultako:
This PR is now ready for review, green flag in CI.
Minimal versions of core packages supported.


```
            click==7.0
            pydantic==1.0
            pyyaml==5.1
            regions==0.5
            scipy==1.2
            astropy==4.3
            numpy==1.19
```

registerrier (approved):
Thanks @Bultako . Looks good!

@adonath is it fine with you?

adonath (approved):
Thanks @Bultako, no further comments from myself side.

Issue #3675, opened 2021-12-03
Polish axis labels


adonath:
For v0.19 we moved all the axes formatting for plots we make to the `MapAxis` object. This defines a single place where the formatting of plot axes happens. By default the the axis table is chosen as `axis.name.title()`. While this is nice because it guarantees consistency between the actual axis plotted and the axis label, it leads to less pretty axis labels. To solve this I proposed to introduce a dictionary to translate map axis names to plot axis labels. Like:

```python3
PLOT_AXIS_LABEL = {
    "energy": "Energy",
    "energy_true": "True Energy",
    "migra": ...
}
```

Of course we have to agree in which axes labels to chose...

registerrier:
I suppose this should go with a general definition of reserved axes names in gammapy, otherwise we might duplicate names, no? 
So far, this does not really exist. There is a definition for IRF in `IRF_DL3_AXS_SPECIFICATION`. Where would the plot label registry would go?

adonath:
@registerrier We already have a definition of the reserved axis names here: https://docs.gammapy.org/0.19/irf/index.html#irf-axis-naming 

I think the the plot label registry could either go in `gammapy.visualisation` or `gammapy.maps.axes`. I think both is fine by now.

PR #3740, opened 2022-01-25
Common format axis labels

fabiopintore:
This PR fixes the issue #3675. 
A registry of labels is added to `~gammapy.maps.axes` and called by the `format_plot_xaxis` functions of `MapAxis` and `TimeMapAxis`. Two tests were added.
However, it's not clear to me how the `LabelMapAxis` should be treated in this case.

### DIFF
--- a/gammapy/maps/axes.py
+++ b/gammapy/maps/axes.py
@@ -45,6 +45,18 @@ def pix_to_coord(edges, pix, interp="lin"):
     return scale.inverse(interp_fn(pix))
 
 
+PLOT_AXIS_LABEL = {
+    "energy": "Energy",
+    "energy_true": "E$_{\rm true}$",
+    "offset": "Offset",
+    "rad": "$\delta p$",
+    "migra": "$\mu$",
+    "fov_lon": "FOV Lon.",
+    "fov_lat": "FOV Lat.",
+    "time": "Time",
+}
+
+
 class MapAxis:
     """Class representing an axis of a map.
 
@@ -278,7 +290,7 @@ class MapAxis:
         """
         ax.set_xscale(self.as_plot_scale)
 
-        xlabel = self.name.capitalize() + f" [{ax.xaxis.units}]"
+        xlabel = PLOT_AXIS_LABEL[self.name] + f" [{ax.xaxis.units}]"
         ax.set_xlabel(xlabel)
         ax.set_xlim(self.bounds)
         return ax
@@ -2187,7 +2199,7 @@ class TimeMapAxis:
         import matplotlib.pyplot as plt
         from matplotlib.dates import DateFormatter
 
-        xlabel = self.name.capitalize() + f" [{self.time_format}]"
+        xlabel = PLOT_AXIS_LABEL[self.name] + f" [{self.time_format}]"
 
         ax.set_xlabel(xlabel)
 
--- a/gammapy/maps/axes.py
+++ b/gammapy/maps/axes.py
@@ -47,10 +47,10 @@ def pix_to_coord(edges, pix, interp="lin"):
 
 PLOT_AXIS_LABEL = {
     "energy": "Energy",
-    "energy_true": "E$_{\rm true}$",
+    "energy_true": r"E$_{\rm true}$",
     "offset": "Offset",
-    "rad": "$\delta p$",
-    "migra": "$\mu$",
+    "rad": r"$\delta p$",
+    "migra": r"$\mu$",
     "fov_lon": "FOV Lon.",
     "fov_lat": "FOV Lat.",
     "time": "Time",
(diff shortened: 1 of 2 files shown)


gammapy/maps/axes.py
```diff
@@ -45,6 +45,18 @@ def pix_to_coord(edges, pix, interp="lin"):
     return scale.inverse(interp_fn(pix))
 
 
+PLOT_AXIS_LABEL = {
+    "energy": "Energy",
```
maxnoe, line 49: This seems very inconsistent. Mixing short mathematical symbols with text. 

Probably we should provide consistent styles here? One with mathematical symbols and one with text?

gammapy/maps/axes.py
```diff
@@ -45,6 +45,18 @@ def pix_to_coord(edges, pix, interp="lin"):
...
+    "energy_true": r"E$_{\rm true}$",
```
AtreyeeS, line 50: If we are to have a registry, I would go for `True Energy` explicitly

gammapy/maps/axes.py
```diff
@@ -45,6 +45,18 @@ def pix_to_coord(edges, pix, interp="lin"):
...
+    "offset": "Offset",
+    "rad": r"$\delta p$",
```
AtreyeeS, line 52: I don't know is $\delta p$ is clearer than `rad`

gammapy/maps/axes.py
```diff
@@ -2187,7 +2199,7 @@ def format_plot_xaxis(self, ax):
         import matplotlib.pyplot as plt
         from matplotlib.dates import DateFormatter
 
-        xlabel = self.name.capitalize() + f" [{self.time_format}]"
+        xlabel = PLOT_AXIS_LABEL[self.name] + f" [{self.time_format}]"
```
AtreyeeS, line 2202: If the name is not in `PLOT_AXIS_LABEL`, it should use `self.name`. After all, `Maps` are sufficiently flexible and we should not prevent plotting for random axis names

AtreyeeS (changes requested):
Thanks @fabiopintore . I added some comments inline

maxnoe:
Can we at this point maybe also address the units story? 

The label could be formatted like this:


```
DEFAULT_LABEL_TEMPLATE = '{quantity} / {unit}'

....

label = DEFAULT_LABEL_TEMPLATE.format(quantity=PLOT_AXIS_LABEL[axis], unit=unit)
```

This would allow users to override the style with which units are typeset and would allow using the correct notation for units instead of the one usually used in HEP.

gammapy/maps/axes.py
adonath: Yes, I'd use `PLOT_AXIS_LABEL.get(self.name, self.name.capitalize())` to prevent it fails for unknown / uncommon axes.

gammapy/maps/axes.py
adonath: Yes, we should definitely introduce consistency. I personally prefer the text based labels, because of readability and context. We could introduce minimal configuration to let users choose, on the other hand we can just choose one (convention over configuration). Users are always able to overwrite if they don't like it using:
```python
ax = data.plot_something()
ax.set_xlabel("My label")
```

fabiopintore: Thanks @maxnoe @adonath. It is fine to follow just the text format only. I proposed both as in some cases I wasn't sure about the best way to represent the label. As for example the case of migration: what's the best here? "migration", "True Energy / Energy" or "μ"? If we go to the text format only, I would prefer the notation "True Energy / Energy".
About the options "configuration vs convention", I would vote for convention.

gammapy/maps/axes.py
fabiopintore: Thanks @AtreyeeS . If we choose the text only format, I also like `True Energy`.

gammapy/maps/axes.py
fabiopintore: I took this definition from [the IRF axis ](https://docs.gammapy.org/0.19/irf/index.html#irf-axis-naming). Effectively $\delta p$ is not really clear to all users, while `rad` is more common.

gammapy/maps/axes.py
fabiopintore: Good point. I agree on this.

adonath:
@maxnoe I'd be happy to adapt the SI convention for Gammapy and not offer any alternative at all. Users can always overwrite the labels using matplotlib if they don't like what Gammapy provides. On the other the minimal configuration system via the top level variable you propose is probably acceptable as well. It could even remain undocumented but would at least offer an entry point...

fabiopintore:
I've followed your suggestions and I've modified the code accordingly.

adonath:
@registerrier Can you maybe do one final review here? Thanks!

registerrier (approved):
Thanks @fabiopintore

PR #3742, opened 2022-01-26
Merge obs_info from obs index with EVENTS Header, fixes #3724

### LINKED ISSUE #3724: Read information from `EVENTS` header when not present in `OBS_INDEX`

In the GADF, the `OBS_INDEX` table, mainly for convenience I think, has optional columns for many header keywords that are mandatory for the `EVENTS` HDU header.

However, gammapy only relies on the obs index table in most parts to retrieve this information. 

This is suboptimal, as requiring optional columns and ignoring mandatory keywords somewhere else is not what implementers of the standard expect.
 
When creating observations, read the info from the event header, not an obs index.
Maybe check for consistency if both `OBS_INDEX` and the `EVENT` header have the same information.

**Additional context**

Mandatory headers in `EVENTS` specification: https://gamma-astro-data-formats.readthedocs.io/en/latest/events/events.html#mandatory-header-keywords

Columns of the `OBS_INDEX`:  https://gamma-astro-data-formats.readthedocs.io/en/latest/data_storage/obs_index/index.html#optional-columns

maxnoe:
* Turn Observation.obs_info into a property that merges the explicitly
  given obs_info, e.g. as filled by `DataStore` from the obs-index table
  with the meta information in `events.table.meta`, which contains
  the mandadatory information, at least for GADF.

### DIFF
--- a/docs/tutorials/analysis/3D/event_sampling.ipynb
+++ b/docs/tutorials/analysis/3D/event_sampling.ipynb
@@ -362,7 +362,7 @@
    "outputs": [],
    "source": [
     "primary_hdu = fits.PrimaryHDU()\n",
-    "hdu_evt = fits.BinTableHDU(events.table)\n",
+    "hdu_evt = events.to_table_hdu()\n",
     "hdu_gti = fits.BinTableHDU(dataset.gti.table, name=\"GTI\")\n",
     "hdu_all = fits.HDUList([primary_hdu, hdu_evt, hdu_gti])\n",
     "hdu_all.writeto(\"./event_sampling/events_0001.fits\", overwrite=True)"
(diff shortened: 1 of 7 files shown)


maxnoe:
Since the keywords in `meta` are there without units when read from fits, I had to introduce some of the GADF specification here. 

There were also strange test failures when I just updated the obs_info from the metaheader:

```
    
        assert not mask_nonan[0, 0, 0]
    
>       assert_allclose(bkg[mask_nonan].max(), 20.656366)
E       AssertionError: 
E       Not equal to tolerance rtol=1e-07, atol=0
E       
E       Mismatched elements: 1 / 1 (100%)
E       Max absolute difference: 3.6159321e+28
E       Max relative difference: 1.75051705e+27
E        x: array(3.615932e+28)
E        y: array(20.656366)

gammapy/makers/tests/test_safe.py:117: AssertionError
```

Which went away after restricting the obs_info to known important keys and not overwriting info found in the obs table with content from the events header.

That bothers me a bit. Any idea which header keyword could make that test fail so spectactularly? 27 Orders of magnitude off...

maxnoe:
Ok, sorry, this now got a bit larger because doing this correctly and in a compatible way with the information stored in the obs info table required treating units and converting to/from quantities.

I had to fix some parts of the simulator to enter correct metadata and add a `to_table_hdu` method, converting units to plain numbers according to the GADF required units.

maxnoe:
Put on hold for the discussion of #3767

maxnoe:
This was merged as part of another PR

Issue #3685, opened 2021-12-10
Adding utility function to provide rebinned energy axis


registerrier:
A frequent request from users is a tool to determine a good binning for representation of flux point extraction. 

Such a function could determine a `MapAxis` that would provide some requested constraint on energy bins e.g. min number of counts, of off counts, min significance etc.
I suppose this is mostly relevant for `SpectrumDataset` but maybe not only.

adonath:
Solved by #3745.

PR #3745, opened 2022-01-26
Helper function to rebin map axis

luca-giunti:
**Description**
Thi PR addresses #3685 

**Dear reviewer**
A few points:
1. This implementation assumes that the low stat bins are always toward the high-energy end of a spectrum dataset
2. It fails with `IndexError` when the user-defined condition cannot be met (e.g. `min_counts=1e10`)
3. I think it is not too complex to add also significance-based options, using the Cash/WstatCountsStatistic. Possibly in a future PR
4. I raised an error if the input axis has `node_dtype="center"`, even if in principle this could work in that case as well. That's for consistency with the `Dataset.resample_energy_axis()` (to which the axis obtained with this new method is expected to be passed), which works only for `node_dtype="edges".`

### DIFF
--- a/gammapy/datasets/spectrum.py
+++ b/gammapy/datasets/spectrum.py
@@ -237,6 +237,51 @@ class SpectrumDataset(PlotMixin, MapDataset):
     def to_spectrum_dataset(self, *args, **kwargs):
         raise NotImplementedError("Already a Spectrum Dataset. Method not supported")
 
+    def get_resampled_energy_axis(self, name="energy", conditions_kw={}):
+        """Returns an energy axis whose binning satisfies given conditions on the per-bin statistics.
+
+        Parameters
+        ----------
+        name: str
+            Name of the axis to be resampled.
+        conditions_kw : dict
+            Keyword arguments containing the per-bin conditions used to resample the axis.
+            Available options are: "min_counts", "min_excess".
+
+        Returns
+        -------
+        energy_axis : `~gammapy.maps.MapAxis`
+            New energy axis.
+        """
+        available_options = ["min_counts", "min_excess"]
+
+        def _find_idx(key):
+            if key in available_options:
+                data = getattr(self, key.strip("min_")).data[::-1]
+                cumsum_idx = np.where(np.cumsum(data) >= conditions_kw[key])[0][0]
+                data_idx = np.where(data >= conditions_kw[key])[0][0]
+                return max(cumsum_idx, data_idx)
+            else:
+                raise ValueError(
+                    f"Unrecognized option {key}. The available methods are: {available_options}."
+                )
+
+        idx = 0
+        for key in conditions_kw.keys():
+            idx = max(idx, _find_idx(key))
+
+        energy_axis = self._geom.axes[name]
+        if (
+            energy_axis.node_type != "edges"
+        ):  # Here for consistency, since the map axis resampling requires "edges"
+            raise ValueError("Only edge based axes can be rebinned")
+        edges = energy_axis.edges
+        rebinned_energy_axis = energy_axis.copy(
+            nodes=np.delete(edges, range(len(edges) - 1 - idx, len(edges) - 1))
+        )
+
+        return rebinned_energy_axis
+
 
 class SpectrumDatasetOnOff(PlotMixin, MapDatasetOnOff):
     stat_type = "wstat"
--- a/gammapy/datasets/spectrum.py
+++ b/gammapy/datasets/spectrum.py
@@ -237,6 +237,7 @@ class SpectrumDataset(PlotMixin, MapDataset):
     def to_spectrum_dataset(self, *args, **kwargs):
         raise NotImplementedError("Already a Spectrum Dataset. Method not supported")
 
+
 class SpectrumDatasetOnOff(PlotMixin, MapDatasetOnOff):
     stat_type = "wstat"
     tag = "SpectrumDatasetOnOff"
@@ -359,5 +360,3 @@ class SpectrumDatasetOnOff(PlotMixin, MapDatasetOnOff):
             SpectrumDatset with cash statistics
         """
         return self.to_map_dataset(name=name).to_spectrum_dataset(on_region=None)
-
-
--- a/gammapy/datasets/spectrum.py
+++ b/gammapy/datasets/spectrum.py
@@ -237,14 +237,14 @@ class SpectrumDataset(PlotMixin, MapDataset):
     def to_spectrum_dataset(self, *args, **kwargs):
         raise NotImplementedError("Already a Spectrum Dataset. Method not supported")
 
-    def get_resampled_energy_axis(self, name="energy", conditions_kw={}):
+    def get_resampled_energy_axis(self, name="energy", conditions={}):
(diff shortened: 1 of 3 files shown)


adonath:
Thanks a lot @luca-giunti, this is very useful functionality!

The main point I'd like to bring up is the question of API here. I'm not sure whether attaching it directly to the `Dataset` object makes sense. Maybe it is just an independent helper function? I haven't really thought about this yet, I guess it depends on how it will be used in practice and which other methods (e.g. Li&Ma significance) we would like to support in future. 

- Maybe it is likely that the methods event gets an `FluxEstimator` instance at some point to derive significance? 
- What about multiple datasets? Does the condition hold per dataset? Or on the joint / stacked ones?
- It's not 100% clear to me how this method generalises to `MapDataset`, probably one just considers the counts in a given region?

My proposal would be to move this to an independent `estimate_map_axis_resampling()` and possibly a more specific `estimate_energy_edges()` method in `gammapy.estimators.utils` for now. This leaves the option open to maybe treat multiple datasets or an estimator instance later.

gammapy/datasets/spectrum.py
```diff
@@ -237,6 +237,51 @@ def plot_residuals_spatial(self, *args, **kwargs):
...
+        -------
+        energy_axis : `~gammapy.maps.MapAxis`
+            New energy axis.
+        """
+        available_conditions = ["min_counts", "min_excess"]
+
+        def _find_idx(key):
+            if key in available_conditions:
+                data = getattr(self, key.strip("min_")).data[::-1]
+                cumsum_idx = np.where(np.cumsum(data) >= conditions[key])[0][0]
```
adonath, line 261: We already have https://docs.gammapy.org/0.19/api/gammapy.maps.Map.html#gammapy.maps.Map.cumsum, which returns a new `Map` object with correctly adapted axis definition. This might help to simplify the implementation here.

luca-giunti:
Hello @adonath :) Thank you for your comments! I moved it to the `estimators.utils`, and for the moment I raised an error if `MapDataset` or `Datasets` are passed. It will be easy to generalize later if needed

> We already have https://docs.gammapy.org/0.19/api/gammapy.maps.Map.html#gammapy.maps.Map.cumsum, which returns a new `Map` object with correctly adapted axis definition. This might help to simplify the implementation here.

I am not sure about that, because the cumsum here has to be done on the "inverted" energy axis (i.e. starting at the end). So doing it on the data seems easier to me than on the map. Am I missing something?

luca-giunti:
Ah one thing, I left the test in `datasets/tests/test_spectrum.py` because there a `SpectrumDataset` is already available. Should I move the test to `estimators/tests/test_utils.py`?

adonath:
@luca-giunti and @registerrier Here is my proposed draft implementation that works on `Dataset` objects:

```python3
from gammapy.datasets import Datasets, SpectrumDatasetOnOff
import numpy as np

def resample_energy_axis_adaptively(dataset):
    """"""
    axis = dataset.counts.geom.axes["energy"]
    # TODO: could introduce some convenience on dataset for this...
    energy_min_map, energy_max_map = dataset.energy_range
    energy_min_all = energy_min_map.quantity.squeeze()
    energy_max_all = energy_max_map.quantity.squeeze()
    
    energy_edges = [energy_max_all]
    
    while energy_edges[-1] > energy_min_all:
        for energy_min in reversed(axis.edges_min):
            if energy_min >= energy_edges[-1]:
                continue

            sliced = dataset.slice_by_energy(
                energy_min=energy_min, energy_max=energy_edges[-1]
            )
                
            with np.errstate(invalid="ignore"):
                info = sliced.info_dict()

            # TODO: use arbitrary keyword or combinations here...
            if info["sqrt_ts"] > 1:
                energy_edges.append(energy_min)
                break

    return u.Quantity(energy_edges[::-1])

path = "$GAMMAPY_DATA/joint-crab/spectra/hess/"
dataset = SpectrumDatasetOnOff.read(path + "pha_obs23523.fits")
edges = resample_energy_axis_adaptively(dataset)
```

On the specific spectrum dataset with ~80 energy bins it runs in ~200 ms. Which I think is very reasonable. I'm not sure whether it does the exact same thing as your implementation @luca-giunti, but my first impression is, that this might be the better option for the following reasons:
- It has mask handling
- In general consistent handling with what we do in `Dataset`
- It already gives access to all the quantities in `.info_dict()` as criteria
- It give a straight forward solution to be used with `Estimator` objects later, because the `Dataset` objects are already there

In general I wasn't sure how the method is expected work: does it just adapt the energy binning of the dataset, or should it maybe take user defined energy edges and adapt those? Or maybe both?

adonath:
Here are a few more general thoughts:
- I think this is maybe the first use-case of what I would call a "MetaEstimator"
- Instead of a helper method it could actually be an own `AdaptiveBinningEstimator` or similar, which takes the configuration on init and a second `Estimator` class. But this is future work...
- In principle there is no limitation to energy, it could e.g. time bins as well. In this case one has to work with `Dataset` objects in any case, because the time info is only available at the `Dataset` level.
- I think there should be the possibility to provide a custom n=binning by the users, which is then adapted.

registerrier:
> @luca-giunti and @registerrier Here is my proposed draft implementation that works on `Dataset` objects:

Well, why not. This offers the possibility 

But somehow, I think the issue is still there. The counts statistics are poorly handled in `info_dict`. You have to create a `CountsStatistics` that serves only for `sqrt_ts` and the whole summation of `alpha` is coded in `info_dict`. That's not where it should be coded. 

Note that you can do:
```
stat=WStatCountsStatistic(dataset.counts, dataset.counts_off, dataset.alpha)
stat.n_bkg
stat.sqrt_ts
```
The right approach would be to remove most statistics based calculation from the `Dataset` implementation, especially in functions like `info_dict` where this does not belong, and move the correct `alpha` summation in a `CountsStatistics.sum()`. 

> On the specific spectrum dataset with ~80 energy bins it runs in ~200 ms. Which I think is very reasonable. I'm not sure whether it does the exact same thing as your implementation @luca-giunti, but my first impression is, that this might be the better option for the following reasons:
> 
> * It has mask handling
> * It has the alpha handling
> * In general consistent handling with what we do in `Dataset`
> * It already gives access to all the quantities in `.info_dict()` as criteria
> * It give a straight forward solution to be used with `Estimator` objects later, because the `Dataset` objects are already there
> 
> In general I wasn't sure how the method is expected work: does it just adapt the energy binning of the dataset, or should it maybe take user defined energy edges and adapt those? Or maybe both?

I suppose this should start with the dataset binning.

registerrier:
> * I think this is maybe the first use-case of what I would call a "MetaEstimator"

Well, technically it does not estimate any quantity, it is rather a preparatory step to fitting or estimation. That's why I am 50/50 on having it `gammapy.estimators` rather than `gammapy.datasets`.

The question of rebinning could be applied to `FluxMaps` and `FluxPoints` but that would require a specific implementation of a `resample` method.

> * I think there should be the possibility to provide a custom n=binning by the users, which is then adapted.

One would resample with the desired starting binning first and then apply regrouping no?

luca-giunti:
Superseded by #3796.

Issue #3686, opened 2021-12-11
Non-functional code block in FluxPoints docs


mdpunch:
**Gammapy version**
```
System:

        python_executable      : /scr/punch/miniconda3/envs/alto-uproot-gammapy/bin/python 
        python_version         : 3.9.7      
        machine                : x86_64     
        system                 : Linux      


Gammapy package:

        version                : 0.19       
        path                   : /scr/punch/miniconda3/envs/alto-uproot-gammapy/lib/python3.9/site-packages/gammapy 

INFO:numexpr.utils:NumExpr defaulting to 8 threads.

Other packages:

        numpy                  : 1.21.4     
        scipy                  : 1.7.3      
        astropy                : 5.0        
        regions                : 0.5        
        click                  : 8.0.3      
        yaml                   : 6.0        
        IPython                : 7.29.0     
        jupyterlab             : 3.2.4      
        matplotlib             : 3.5.0      
        pandas                 : 1.3.4      
```
**Bug description**
While trying to understand the FluxPoints, I copy/pasted code from the documentation page into jupyter notebook.

From https://docs.gammapy.org/0.19/api/gammapy.estimators.FluxPoints.html, the code example being:
```python
import numpy as np
from astropy import units as u
from astropy.table import Table
from gammapy.estimators import FluxPoints
from gammapy.modeling.models import PowerLawSpectralModel

table = Table()
pwl = PowerLawSpectralModel()
e_ref = np.geomspace(1, 100, 7) * u.TeV

table["e_ref"] = e_ref
table["dnde"] = pwl(e_ref)
table.meta["SED_TYPE"] = "dnde"

flux_points = FluxPoints.from_table(table)
flux_points.plot(sed_type="flux")
```
But, the code gives errors, as follows.
```
No reference model set for FluxMaps. Assuming point source with E^-2 spectrum.

---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
/tmp/ipykernel_405804/603317310.py in <module>
     17 
     18 flux_points = FluxPoints.from_table(table)
---> 19 flux_points.plot(sed_type="flux")

/scr/punch/miniconda3/envs/pyirf/lib/python3.7/site-packages/gammapy/estimators/points/core.py in plot(self, ax, sed_type, energy_power, **kwargs)
    494 
    495         flux = scale_plot_flux(flux=flux.to_unit(flux_unit), energy_power=energy_power)
--> 496         ax = flux.plot(ax=ax, **kwargs)
[...]

PR #3748, opened 2022-01-26
Fix doc strings for makers and datasets

AtreyeeS:
This PR fixes the doc string in 2 sub packages.
Probably good to merge this before I start other packages to avoid changing too many files and creating merge conflicts.

I wanted to add an example for all classes, but it started to become like full tutorials, so avoided in some cases.

### DIFF
--- a/gammapy/datasets/flux_points.py
+++ b/gammapy/datasets/flux_points.py
@@ -15,7 +15,7 @@ __all__ = ["FluxPointsDataset"]
 
 class FluxPointsDataset(Dataset):
     """
-    Bundle a set of flux points with a parametric model, to compute a likelihood. Uses chi2 statistics.
+    Bundle a set of flux points with a parametric model, to compute fit statistic function using chi2 statistics.
 
     Parameters
     ----------
--- a/gammapy/datasets/flux_points.py
+++ b/gammapy/datasets/flux_points.py
@@ -15,7 +15,7 @@ __all__ = ["FluxPointsDataset"]
 
 class FluxPointsDataset(Dataset):
     """
-    Bundle a set of flux points with a parametric model, to perform a fit. Uses chi2 statistics.
+    Bundle a set of flux points with a parametric model, to compute a likelihood. Uses chi2 statistics.
 
     Parameters
     ----------
@@ -207,14 +207,13 @@ class FluxPointsDataset(Dataset):
 
         str_ += "\t{:32}: {} \n".format("Number of models", n_models)
 
-        str_ += "\t{:32}: {}\n".format(
-            "Number of parameters", len(self.models.parameters)
-        )
-        str_ += "\t{:32}: {}\n\n".format(
-            "Number of free parameters", len(self.models.parameters.free_parameters)
-        )
-
         if self.models is not None:
+            str_ += "\t{:32}: {}\n".format(
+                "Number of parameters", len(self.models.parameters)
+            )
+            str_ += "\t{:32}: {}\n\n".format(
+                "Number of free parameters", len(self.models.parameters.free_parameters)
+            )
             str_ += "\t" + "\n\t".join(str(self.models).split("\n")[2:])
 
         return str_.expandtabs(tabsize=2)
@@ -241,7 +240,7 @@ class FluxPointsDataset(Dataset):
         return ((data - model) / sigma.quantity[:, 0, 0]).to_value("") ** 2
 
     def residuals(self, method="diff"):
-        """Compute the flux point residuals ().
+        """Compute flux point residuals
 
         Parameters
         ----------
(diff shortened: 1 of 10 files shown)


AtreyeeS:
Is it possible to add a tolerance limit for the doc-string tests @Bultako? The test fail I am getting here passes locally. Its off by rtol < 1e-12, but I think it is checking as a string and failing.

gammapy/makers/background/reflected.py
```diff
@@ -59,11 +59,11 @@ class ReflectedRegionsFinder:
     >>> on_region = CircleSkyRegion(target_position, theta)
     >>> finder = ReflectedRegionsFinder(min_distance_input='1 rad', region=on_region, center=pointing)
     >>> regions = finder.run()
-    >>> print(regions[0])
+    >>> print(regions[0]) # doctest: +SKIP
```
maxnoe, line 62: Don't need to skip. You can use an ellipsis:
` # doctest: +ELLIPSIS`

and then in the result:

```
radius: 1438.3... arcsec
```

Bultako:
Thanks @AtreyeeS !
Could you also fix `gammapy.estimators.FluxPoints` described in https://github.com/gammapy/gammapy/issues/3686?

AtreyeeS:
As discussed, I will add some examples in Datasets plot. I propose to do it in this PR

AtreyeeS:
I am updated the doc strings (including some code examples for plotting) for datasets.
made some minor changes in the code.
 - One change I mention here is that if no `region` is passed, the full map is used consistently in all plot methods.
 - fixed `FluxPointDataset.__str__` in case no model is present

2 plot methods that remains complicated with `**kwargs` is 
- `SpectrumDataset.plot_fit()`.  It takes  `kwargs_spectrum` and passes it to `plot_excess` which itself takes `kwargs_excess, kwargs_npred_signal, **kwargs`. 
-  `FluxPointDataset.plot_fit()` passes `kwargs_spectrum` to `plot_spectrum()` which passes 
- - `kwargs_fp` to `FluxPoints.plot()` which passes to `RegionNDMap.plot`  
- - and `kwargs_model` to `SpectralModel.plot` and `SpectralModel.plot_error` which finally goes to matplotlib

This inception like behaviour is complex to understand. Needs a dictionary of dictionaries. Can it be simplified in a followup PR @adonath ?

AtreyeeS:
> Could you also fix `gammapy.estimators.FluxPoints` described in #3686?

@Bultako : I propose we merge this PR and I do it in a followup PR because it changes many files. Probably its simpler to do it one package at a time.

gammapy/datasets/map.py
```diff
@@ -818,6 +866,13 @@ def plot_residuals_spatial(
         -------
         ax : `~astropy.visualization.wcsaxes.WCSAxes`
             WCSAxes object.
+
+        Example
+        -------
+        >>> from gammapy.datasets import MapDataset
+        >>> dataset = MapDataset.read("$GAMMAPY_DATA/cta-1dc-gc/cta-1dc-gc.fits.gz")
+        >>> kwargs = {"cmap": "gnuplot2", "add_cbar": True}
```
adonath, line 874: For residual plotting I'd always use `cmap="RdBu"`  and a symmetric data range, like `vmin=-5, vmax=5`

gammapy/datasets/map.py
```diff
@@ -941,13 +1005,26 @@ def plot_residuals(
...
+
+        Examples
+        --------
+        >>> from regions import CircleSkyRegion
+        >>> from astropy.coordinates import SkyCoord
+        >>> import astropy.units as u
+        >>> from gammapy.datasets import MapDataset
+        >>> dataset = MapDataset.read("$GAMMAPY_DATA/cta-1dc-gc/cta-1dc-gc.fits.gz")
+        >>> reg = CircleSkyRegion(SkyCoord(0,0, unit="deg", frame="galactic"), radius=1.0*u.deg)
+        >>> kwargs_spatial = {"cmap": "gnuplot2", "add_cbar": True}
```
adonath, line 1025: Same comment as above...

adonath (commented):
Thanks @AtreyeeS, I have left some minor final comments on the colormap for the residuals...

gammapy/datasets/map.py
maxnoe: I always use `"RdBu_r"` since most people associate blue with negative and red with positive changes (colder / warmer)

AtreyeeS:
Thanks @adonath for the comments. Made the change in the colour map.
I suppose its okay to have dict of dicts now that it is documented

AtreyeeS:
I do not understand the test fails here...

Bultako:
maybe the fail is caused by the new release of pytest

https://github.com/pytest-dev/pytest/releases/tag/7.0.0

AtreyeeS:
@Bultako should I update my dev environment for the new pytest version then? Sorry, I really can't understand the error message here...

AtreyeeS:
Thanks for the fix @Bultako ! 
Shall we merge this PR now @adonath ?

gammapy/datasets/flux_points.py
```diff
@@ -15,14 +15,14 @@
 
 class FluxPointsDataset(Dataset):
     """
-    Fit a set of flux points with a parametric model.
+    Bundle a set of flux points with a parametric model, to compute a likelihood. Uses chi2 statistics.
```
registerrier, line 18: to compute a fit statistic function?

gammapy/datasets/map.py
```diff
@@ -431,9 +466,9 @@ def _background_parameters_changed(self):
         return changed
 
     def npred_signal(self, model_name=None):
-        """ "Model predicted signal counts.
+        """Model predicted signal counts.
 
-        If a model is passed, predicted counts from that component is returned.
+        If a model name is passed, predicted counts from that component is returned.
```
registerrier, line 471: predicted counts from that component are returned

gammapy/datasets/map.py
```diff
@@ -1929,7 +2032,7 @@ def alpha(self):
         return alpha
 
     def npred_background(self):
-        """Prediced background counts estimated from the marginalized likelihood estimate.
+        """Prediced background counts (mu_bkg) estimated from the marginalized likelihood estimate.
```
registerrier, line 2035: Predicted

registerrier (approved):
Thanks @AtreyeeS . This looks good! I have left a few small inline comments.

PR #3752, opened 2022-01-27
Human readable energy units string formatting for plot_interactive & plot_grid

facero:
**Description**
When using data on a large energy range or for the example the Crab dataset, the energy label are not the most readable (e.g. 2.56e8 keV instead of 256 GeV).
This PR adds a helper function in utils to transform a string, list/np array of energy Quantities to a string or list of strings of well formatted energies.
Example:
E = [1.54e2 *u.GeV, 4300 *u.keV, 300.6e12*u.eV]
=> ['154 GeV', '4.30 MeV', '301 TeV']

Example of before/After in plot_interactive:
image

**Dear reviewer**
This helper function has been applied to the `plot_interactive()` and `plot_grid()` functions to show more readable energies.
The function could also be useful in other contexts but I started with these only 2 cases for now.

Tests were added as well. 

One point I'm not sure is how to best check whether the input is iterable or not in `energy_unit_format()`.
Possible inputs are list of quantities or a np.array of quantities.
If I have just E = 1 *u.TeV I don't to return a list.
Any suggestions to improve my first attempt ?


adonath (commented):
Thanks @facero!

While I think it is clearly an improvement of readability in the plots, I must admit I'm not yet happy with the implementation yet. This PR adds ~100 lines of code to achieve this small improvement, which seems a bit "overkill" and it is hard-coded to energy units only. I did a bit of research to see whether there are already similar formatting options available in `numpy` or `astropy`, but it doesn't seem to be the case... 

Maybe we should open an issue in `astropy` first and ask the `astropy.units` experts how to best achieve something like this? I noticed e.g. that `astropy` handles this as a special `PrefixUnit`...

gammapy/utils/units.py
```diff
@@ -64,3 +65,63 @@ def unit_from_fits_image_hdu(header):
...
+        E = E.to('PeV')
+
+    if E.value < 10:
+        prec=2
+    elif (E.value >= 10) and (E.value < 100):
+        prec=1
+    elif (E.value >= 100) and (E.value < 1000):
+        prec=0
+    
+    return f"{E.value:0.{prec}f} {E.unit}"
```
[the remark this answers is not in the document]
nstarman: Something similar can be done for the precision selection.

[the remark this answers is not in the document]
facero: thanks for the suggestion. This really reduces the number of code line.

gammapy/utils/units.py
```diff
@@ -64,3 +66,54 @@ def unit_from_fits_image_hdu(header):
...
+    else:  # get best large unit
+        unit = (u.eV, u.keV, u.MeV, u.GeV, u.TeV, u.PeV)[i]
+
+    v = E.to_value(unit)
+    i=floor(np.log10(v))
+    prec=(2,1,0)[i]
+
+    return f"{v:0.{prec}f} {unit}"
+
+def energy_unit_format(E):
```
facero, line 95: I've implemented the very good suggestions from Nathaniel. Now this function `energy_unit_format` and in maps/core.py could also be optimized a bit. Any suggestions ?

gammapy/maps/core.py
```diff
@@ -1129,7 +1135,17 @@ def plot_interactive(self, rc_params=None, **kwargs):
...
-            options = axis.as_plot_labels
+            if axis.node_type == "center":
+                if axis.name == "energy":
+                    options =  energy_unit_format(axis.center)
+                else:
+                    options = axis.as_plot_labels
+            else:
+                if axis.name == "energy":
+                    E = energy_unit_format(axis.edges)
+                    options = [f"{E[i]} - {E[i+1]}" for i in range(len(E)-1)]
```
nstarman, line 1146: Using ``itertools.pairwise``

```suggestion
                    options = [f"{e1} - {e2}" for e1, e2 in pairwise(E)]
```

gammapy/utils/tests/test_units.py
```diff
@@ -1,8 +1,30 @@
...
 def test_standardise_unit():
     assert standardise_unit("ph cm-2 s-1") == "cm-2 s-1"
     assert standardise_unit("ct cm-2 s-1") == "cm-2 s-1"
     assert standardise_unit("cm-2 s-1") == "cm-2 s-1"
+
+def test_energy_str_formatting():
+    assert energy_str_formatting(1530e4 *u.keV) == "15.3 GeV"
+    assert energy_str_formatting(1530 *u.GeV) == "1.53 TeV"
+    assert energy_str_formatting(1530 *u.GeV) == "1.53 TeV"
+    assert energy_str_formatting(1530.5e8 *u.keV) == "153 TeV"
```
nstarman, line 16: This feels like a ``pytest.parametrize``

```suggestion
@pytest.parametrize("q, expect", [(1530e4 *u.keV, "15.3 GeV"), (1530 *u.GeV, "1.53 TeV"), (1530 *u.GeV, "1.53 TeV"), (1530.5e8 *u.keV, "153 TeV")])
def test_energy_str_formatting(q, expect):
    assert energy_str_formatting(q) == expect
```

the args to parametrize should be on a few lines, of course.

gammapy/utils/tests/test_units.py
```diff
@@ -1,8 +1,30 @@
...
+
+def test_energy_unit_format():
```
nstarman, line 18: Maybe also ``pytest.parametrize``

gammapy/utils/units.py
```diff
@@ -64,3 +66,54 @@ def unit_from_fits_image_hdu(header):
...
+    """    
+    i = floor(np.log10(E.to_value(u.eV)) / 3) # a new unit every 3 decades
+    if i > 4:  # default for very large values
+        unit = u.PeV
+    else:  # get best large unit
+        unit = (u.eV, u.keV, u.MeV, u.GeV, u.TeV, u.PeV)[i]
+
+    v = E.to_value(unit)
+    i=floor(np.log10(v))
+    prec=(2,1,0)[i]
```
nstarman, line 91: A default value is needed if $i >= 3$

```suggestion
    prec = (2,1,0)[i] if i < 3 else 0
```

gammapy/utils/units.py
```diff
@@ -64,3 +66,54 @@ def unit_from_fits_image_hdu(header):
...
+    ----------    
+    E: `~astropy.units.Quantity`
+        Quantity or list of quantities
+    Returns
+    -------
+    str : str
+        Returns a string or list of strings with energy unit formatted        
+    """   
+
+    if (isinstance(E, (list, tuple, np.ndarray)) == False):
```
nstarman, line 108:
```suggestion
    if not np.shape(E):  # scalar
```

gammapy/utils/units.py
```diff
@@ -64,3 +66,54 @@ def unit_from_fits_image_hdu(header):
...
+    prec=(2,1,0)[i]
+
+    return f"{v:0.{prec}f} {unit}"
+
+def energy_unit_format(E):
+    """
+    Format an energy quantity (or a list of quantities) to a string (or list of string) representations.
+    Parameters
+    ----------    
+    E: `~astropy.units.Quantity`
```
nstarman, line 100:
```suggestion
    E: `~astropy.units.Quantity` or sequence thereof
```

gammapy/utils/units.py
```diff
@@ -64,3 +66,54 @@ def unit_from_fits_image_hdu(header):
...
+    str : str
+        Returns a string or list of strings with energy unit formatted        
+    """   
+
+    if (isinstance(E, (list, tuple, np.ndarray)) == False):
+        return energy_str_formatting(E)
+    #Additional test for scalar quantities like 1*u.TeV
+    if isinstance(E, u.quantity.Quantity):
+        if E.shape == ():
+           return energy_str_formatting(E)
```
nstarman, line 113: If the above suggestion is implemented

```suggestion
```

gammapy/utils/units.py
```diff
@@ -64,3 +66,54 @@ def unit_from_fits_image_hdu(header):
...
+    
+    E_fmt = []     
+    for i, E_ in enumerate(E):
+        E_fmt.append(energy_str_formatting(E[i]))
```
nstarman, line 117:
```suggestion
    E_fmt = [energy_str_formatting(e) for e in E]
```

gammapy/utils/tests/test_units.py
facero: Thanks @nstarman. I've implemented the recommended suggestions.
Code is much more compact now.
We'll see if similar features is implemented at some point in astropy as discussed [here](https://github.com/astropy/astropy/issues/6913).

facero:
I've implemented the suggested changes and code is more compact now.
Tests are successful locally (except the known issue related to Regions).
There is still one part that I'm not happy about but don't know if I can do better (see code comment).

gammapy/utils/units.py
```diff
@@ -102,9 +102,12 @@ def energy_unit_format(E):
...
     """   
 
-    if not np.shape(E):  # scalar
-        return energy_str_formatting(E)
-    
-    E_fmt = [energy_str_formatting(e) for e in E]
-    return E_fmt
-    
\ No newline at end of file
+    if (isinstance(E, (list, tuple)) == True):
```
facero, line 105: This is the not easy part.
the input can be different nature:
- astropy quantity (single Q or array of Q)
- list or tuple of Q
I couldn't find a way to simply disentangle between iterable and non iterable inputs.
size, shape, or len doesn't work on  all inputs types.

facero:
Thanks @nstarman for the suggestions.
Especially `return tuple(map(energy_str_formatting, E))` is a nice workaround to handle both scalar and iterables.
With this I reduced the code to one function only which also reduced the tests.
@adonath can you have a look at this latest version ?

gammapy/maps/core.py
```diff
@@ -1065,10 +1066,15 @@ def plot_grid(self, figsize=None, ncols=3, **kwargs):
             image_wcs.plot(ax=ax, **kwargs)
 
             if axis.node_type == "center":
-                info = f"{axis.center[idx]:.1f}"
+                if axis.name == "energy":
```
registerrier, line 1069: Maybe also `energy_true`?

gammapy/maps/core.py
```diff
@@ -1129,7 +1135,16 @@ def plot_interactive(self, rc_params=None, **kwargs):
         interact_kwargs = {}
 
         for axis in self.geom.axes:
-            options = axis.as_plot_labels
+            if axis.node_type == "center":
+                if axis.name == "energy":
```
registerrier, line 1139: Same here check `energy_true`

gammapy/utils/units.py
```diff
@@ -64,3 +66,32 @@ def unit_from_fits_image_hdu(header):
         unit = ""
 
     return standardise_unit(unit)
+
+def energy_unit_format(E):
+    """
+    Format energy quantities to a string representation that is more comfortable to read
+    by swithcing to the most relevant unit (keV, MeV, GeV, TeV) and changing the float precision.
+    Parameters
```
registerrier, line 74: Add blank line before Parameters

gammapy/utils/units.py
```diff
@@ -64,3 +66,32 @@ def unit_from_fits_image_hdu(header):
...
+    ----------    
+    E: `~astropy.units.Quantity`
+        Quantity or list of quantities
+    Returns
```
registerrier, line 78: Add blank line

gammapy/utils/units.py
```diff
@@ -64,3 +66,32 @@ def unit_from_fits_image_hdu(header):
         unit = ""
 
     return standardise_unit(unit)
+
+def energy_unit_format(E):
+    """
+    Format energy quantities to a string representation that is more comfortable to read
+    by swithcing to the most relevant unit (keV, MeV, GeV, TeV) and changing the float precision.
```
registerrier, line 73: switching

registerrier (dismissed):
Thanks @facero . This looks good!

I have left a few minor comments inline.

registerrier:
Merging now. Thanks @facero.

Issue #3679, opened 2021-12-06
makers.utils undocumented online


Tobychev:
**Gammapy version**
0.18.2

**Bug description**
`make_theta_squared_table` docstring does not show up in the documentation search, and I can't find it via the menu either. Probably applies for all functions in that subpackage, but I haven't checked.

The function has a proper docstring, it's just not been caught during generation of the html page, and is thus hard to discover.

**To Reproduce**
Go to docs.gammapy.org and search for make_theta_squared_table, the only hit is in an example.

registerrier:
Note that it appears in hess tutorials.

bkhelifi:
I will add it in the PR about the missing links into the documentation

PR #3753, opened 2022-01-27
Fix broken links and add subpackages in documentation

Bultako:
This PR fixes all broken links found with [make linkcheck](https://docs.gammapy.org/dev/development/doc_howto.html#check-broken-links) in RST files, notebooks and docstrings within scripts, as well as those listed in https://github.com/gammapy/gammapy/issues/3143#issuecomment-1005927960 fixing also others issues from that list. It also adds some subpackages missing in the docs as reported in https://github.com/gammapy/gammapy/issues/3679 and  https://github.com/gammapy/gammapy/issues/3143#issuecomment-1007430661

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -341,7 +341,7 @@ accordingly to an ``EBLAbsorptionNormSpectralModel``. A new
 have been introduced.
 
 Gammapy v0.18 comes now with support for custom energy dependent spatial models.
-An example for this can be found in the models tutorial.
+An example for this can be found in the `models tutorial <https://docs.gammapy.org/0.18/tutorials/models.html>`__.
 The ``SkyDiffuseCube`` has been removed, the same functionality can now be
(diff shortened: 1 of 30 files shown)


CHANGES.rst
```diff
@@ -341,7 +341,7 @@ accordingly to an ``EBLAbsorptionNormSpectralModel``. A new
 have been introduced.
 
 Gammapy v0.18 comes now with support for custom energy dependent spatial models.
-An example for this can be found in the `models tutorial <tutorials/models.html>`__.
```
adonath, line 344: Is it possible to fix the links instead of removing it?

Bultako: Ok, done.
I have replaced the links with absolute links to the versioned docs, instead of removing.

adonath (approved):
Thanks a lot @Bultako! No further comments from my side.

PR #3754, opened 2022-01-27
Fix ci hangs due to  coverage 6.3 by pinning to 6.2

maxnoe:
See nedbat/coveragepy#1310

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,7 @@ jobs:
       - name: install python3.8
         if: contains(matrix.name, '3.8')
         run: |
-          sed -i 's/python==3.7/python==3.8/' environment-dev.yml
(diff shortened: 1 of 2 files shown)


maxnoe:
@Bultako @adonath @registerrier Can we merge this? This is blocking all other PRs at the moment.

Bultako:
Thanks @maxnoe !!

bkhelifi:
At the bottom, one changes maybe change also:
```
  - nbformat==5.0.8
  - h5py==2.10.0

```
to
```
  - nbformat=5.0.8
  - h5py=2.10.0
```

maxnoe:
@bkhelifi That wouldn't be a change in semantics. Changing `python==3.7` to `python=3.7` means any `3.7.x` is ok. But when specifying all three version numbers there is no difference between `=` and `==`

PR #3757, opened 2022-01-27
Use Pool context directly, fixes #3755

### LINKED ISSUE #3755: Use context of Pool directly
**Gammapy version**

Current master

gammapy uses a `multiprocessing.Pool` in two locations and confusingly for me, it is used like this:

```
            with contextlib.closing(Pool(processes=self.n_jobs)) as pool:
                 do_stuff_with_pool
```

Any reason why the context manager of the pool is not used directly?

```
            with Pool(processes=self.n_jobs) as pool:
                 do_stuff_with_pool
```

From the docs:
> Warning
> multiprocessing.pool objects have internal resources that need to be properly managed (like any other resource) by using the pool > as a context manager or by calling close() and terminate() manually. Failure to do this can lead to the process hanging on finalization.

So it seems this might also be the reason for the stuck CI, as using `closing` is not enough, one has to call both `close` and `terminate`? Or just use the pool as context.

maxnoe:


adonath (changes requested):
Thanks @maxnoe! I guess you can remove the `import contextlib` as well. It should be unused now...

maxnoe:
@adonath done

adonath (approved):
Thanks @maxnoe, I have  no further comments.

PR #3759, opened 2022-01-27
Fix print of catalogs

### LINKED ISSUE #3749: Error while printing catalogs
```
from gammapy.catalog import *
catalog = SourceCatalog4FGL()
print(catalog)
```

gives an error
```
AttributeError                            Traceback (most recent call last)
/var/folders/v4/62d6cvbj72qbyw23mq2pwxl00000gn/T/ipykernel_17380/3277138734.py in <module>
      1 catalog = SourceCatalog4FGL()
----> 2 print(catalog)

~/Gammapy-dev/gammapy/gammapy/catalog/core.py in __str__(self)
    110     def __str__(self):
    111         return (
--> 112             f"{self.__class__.__name__}:\n"
    113             f"    name: {self.name}\n"
    114             f"    description: {self.description}\n"

AttributeError: 'SourceCatalog4FGL' object has no attribute 'name'
```

This happens for all catalogs

AtreyeeS:
Fixes #3749 by replacing `catalog.name` by `catalog.tag`
I do not know how best to add a test for a print statement

### DIFF
--- a/gammapy/catalog/core.py
+++ b/gammapy/catalog/core.py
@@ -99,6 +99,11 @@ class SourceCatalog(abc.ABC):
         """Catalog description (str)."""
         pass
 
+    @property
+    @abc.abstractmethod
+    def tag(self):
+        pass
+
     source_object_class = SourceCatalogObject
     """Source class (`SourceCatalogObject`)."""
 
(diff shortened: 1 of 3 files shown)


adonath (changes requested):
Thanks @AtreyeeS! Can you add one unit test to make sure the printing works correctly? And maybe we should introduce `SourceCatalog.tag` as a required abstract property to make sure that all sub-classes actually implement it. See e.g. https://github.com/gammapy/gammapy/blob/master/gammapy/estimators/core.py#L18

adonath (approved):

PR #3762, opened 2022-01-27
Run tests in parallel

maxnoe:
**Description**

Adds pytest-xdist to the dev env and uses it in the CI to run tests in parallel


adonath (approved):
Thanks a lot @maxnoe, no comments from my side.

PR #3768, opened 2022-01-28
Remove non-existent packages from `extern/__init__.py` docstring

maxnoe:
There were comments about to modules that apparently got removed some time ago


adonath (approved):
Thanks @maxnoe!

Issue #3766, opened 2022-01-28
Incorrect label for FluxPoints plots


registerrier:
**Gammapy version**
v0.19 an dev

**Bug description**
When plotting a `FluxPoints` in `dnde` or `e2dnde`, the label is confusing because it gives an interval where it should give the reference energy. 
**Expected behavior**
Give only the reference energy. Possible add it in the y-axis title.

**To Reproduce**
Just look at the flight curve flare tutorial. It is incorrect there.

**Other information**
Any other information you think will be useful for us to fix the issue can go here.

adonath:
Would keep the info in the legend, because it generalizes to plotting multiple light curves at different energies.

adonath:
The strict solution here is actually to change the node type here (in fact for all differential quantities): https://github.com/gammapy/gammapy/blob/master/gammapy/estimators/map/core.py#L570

The label is already handled correctly here: https://github.com/gammapy/gammapy/blob/master/gammapy/maps/axes.py#L237, but internally even for `dnde` maps the `node_type` remains "edges", that's why the label is still shown as an energy range.

PR #3769, opened 2022-01-28
Change flux map attribute axis type 

registerrier:
**Description**

This pull request changes the axis type of the `FluxMaps.dnde` and `FluxMaps.e2dnde` attributes to `center`. This correctly represents differential quantities.

For now a static methods changes the `MapAxis._node_type` keyword. Possibly it could be better to add a setter on the `MapAxis.node_type` itself.

This solves #3766.

**Dear reviewer**

### DIFF
--- a/gammapy/estimators/map/core.py
+++ b/gammapy/estimators/map/core.py
@@ -446,9 +446,11 @@ class FluxMaps:
 
     @staticmethod
     def _change_energy_axis_node_type(input_map, node_type='center'):
-        """This change the node_type of the input map."""
-        input_map.geom.axes["energy"]._node_type = node_type
-        return input_map
+        """Change the node_type of the input map."""
+        energy_axis = input_map.geom.axes["energy"]
+        new_axis = energy_axis.to_node_type(node_type)
+        geom = input_map.geom.replace_axis(axis=new_axis)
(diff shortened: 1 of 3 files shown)


adonath:
Thanks @registerrier! Strictly speaking the change of the `_node_type` is not sufficient, because internally the handling of the nodes and interpolation is different for the different node types. However in practice it should still be fine and I think we can go with the solution you propose. If there are maybe the other use cases to change the node type we could introduce a minimal API like:

```
ax = axis.to_node_type("center")
geom = geom.replace_axis(axis=axis)
Map.from_geom(geom=geom, data=m.data)
```
Where the `geom.replace_axis()` already exists.

registerrier:
I have introduced the solution you suggested, which is indeed cleaner.

Yet there is a serious issue since one cannot create `node_type=center` `MapAxis` with only 1 bin. See #1952

adonath:
I realized the situation is a bit tricky here: internally in Gammapy I think we'd like to keep the info on `e_min` and `e_max` even when changing the node type. However especially in the case of one energy bin  the information will be lost, once the node type is changed. So maybe we have to work with an independent `Map.is_differential_quantity` attribute or similar...

registerrier:
Merging now.

PR #3770, opened 2022-01-29
Make `make docs-show` work not only on macOS

maxnoe:
The `make docs-show` command used `open docs/_build/html/index.html` which is bad for two reasons:

* It only workes on macOS as other systems don't usually have the `open` command available in `sh` which is used by Make
* It opens the index file in the browser, instead of really serving the directory. Which can create problems with loading the custom css or js.

This adds a little python script using the python simple http server to serve the docs directory and open the served page in the default browser, which should work cross platform and handle the real webpage better.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -103,7 +103,8 @@ docs-all:
 	python -m gammapy.utils.notebooks_links --src="$(src)"
 
 docs-show:
-	open docs/_build/html/index.html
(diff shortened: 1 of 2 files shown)


adonath (approved):
Thanks @maxnoe! Nice solution to use the `HTTPServer`...

PR #3773, opened 2022-01-31
Remove uneeded pins to old versions of h5py and nbformat

maxnoe:
* Remove pin to h5py < 3, I guess this was due to https://github.com/astropy/astropy/pull/11359, but this got fixed in the astropy versions we are using

* What was the reason for pinning nbformat?

### DIFF
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -64,7 +64,7 @@ dependencies:
   - twine
   - yamllint
   - nbformat
-  - h5py=3
+  - h5py
   - ruamel.yaml
   - pip:
       - pytest-sphinx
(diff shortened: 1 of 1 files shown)


Bultako:
Thanks @maxnoe 

I think pinning `nbformat` was needed to remove metadata warnings arising with [nbsphinx](http://nbsphinx.readthedocs.io/),  not needed anymore.

Shouldn't we also remove pinning the `h5py`=3 ?

maxnoe:
@Bultako done

adonath (approved):
Thanks @maxnoe and @Bultako! I have no further comments...

Issue #3725, opened 2022-01-25
EnergyDispersion2D does not validate units


maxnoe:
**Gammapy version**
Current master

**Bug description**

`EnergyDispersion2D` happily accepts a data array with arbitrary units

**Expected behavior**

Raises an error for wrong unit

**To Reproduce**

```python
from gammapy.irf import EnergyDispersion2D
from gammapy.maps import MapAxis

energy_axis = MapAxis.from_energy_bounds(1 * u.TeV, 10 * u.TeV, nbin=3, name='energy_true')
offset_axis = MapAxis.from_bounds(0 * u.deg, 2 * u.deg, nbin=2, name='offset')
migra_axis = MapAxis.from_bounds(0.2, 5, nbin=5, interp='log', name='migra')


### OBSERVE THE m**2
data = np.zeros((energy_axis.nbin, migra_axis.nbin, offset_axis.nbin)) * u.m**2

edisp = EnergyDispersion2D(data=data, axes=[energy_axis, migra_axis, offset_axis])
hdu = edisp.to_table_hdu(format='gadf-dl3')
```

registerrier:
Issues resolved by #3777 . Closing now.

PR #3777, opened 2022-02-01
Validate EnergyDispersion2D units

fabiopintore:
The PR solves a bug on the units of the input data for `~gammapy.irf.EnergyDispersion2D`, as reported in the issue #3725 .
It is now added a check on the units, and the code raises an error if the units are wrong.
The bug fix applies on `~gammapy.irf.core.py` although it is probably not the most elegant way to do it...

### DIFF
--- a/gammapy/irf/core.py
+++ b/gammapy/irf/core.py
@@ -44,7 +44,7 @@ class IRF(metaclass=abc.ABCMeta):
         self._axes = axes
         if isinstance(data, u.Quantity):
             self.data = data.value
-            if not self.unit.is_equivalent(data.unit):
+            if not self.default_unit.is_equivalent(data.unit):
                 raise ValueError(f"Error: {data.unit} is not an allowed unit. {self.tag} requires dimensionless data quantities!")
             else:
                 self.unit = data.unit
--- a/gammapy/irf/core.py
+++ b/gammapy/irf/core.py
@@ -45,6 +45,9 @@ class IRF(metaclass=abc.ABCMeta):
         if isinstance(data, u.Quantity):
             self.data = data.value
             self.unit = data.unit
+            if self.axes.names == ["energy_true", "migra", "offset"] :
+                if self.unit != u.dimensionless_unscaled:
+                    raise ValueError(f"Error: {self.unit} is not an allowed unit. EDisp requires dimensionless data quantities!")
         else:
             self.data = data
             self.unit = unit
--- a/gammapy/irf/core.py
+++ b/gammapy/irf/core.py
@@ -44,10 +44,10 @@ class IRF(metaclass=abc.ABCMeta):
         self._axes = axes
         if isinstance(data, u.Quantity):
             self.data = data.value
-            self.unit = data.unit
-            if self.axes.names == ["energy_true", "migra", "offset"] :
-                if self.unit != u.dimensionless_unscaled:
-                    raise ValueError(f"Error: {self.unit} is not an allowed unit. EDisp requires dimensionless data quantities!")
+            if not self.unit.is_equivalent(data.unit):
+                raise ValueError(f"Error: {data.unit} is not an allowed unit. {self.tag} requires dimensionless data quantities!")
+            else:
+                self.unit = data.unit
         else:
             self.data = data
             self.unit = unit
--- a/gammapy/irf/core.py
+++ b/gammapy/irf/core.py
@@ -45,7 +45,7 @@ class IRF(metaclass=abc.ABCMeta):
         if isinstance(data, u.Quantity):
             self.data = data.value
             if not self.default_unit.is_equivalent(data.unit):
-                raise ValueError(f"Error: {data.unit} is not an allowed unit. {self.tag} requires dimensionless data quantities!")
+                raise ValueError(f"Error: {data.unit} is not an allowed unit. {self.tag} requires {self.default_unit} data quantities.")
             else:
                 self.unit = data.unit
         else:
--- a/gammapy/irf/edisp/core.py
+++ b/gammapy/irf/edisp/core.py
@@ -48,7 +48,7 @@ class EnergyDispersion2D(IRF):
 
     tag = "edisp_2d"
     required_axes = ["energy_true", "migra", "offset"]
-    unit = u.Unit("")
+    default_unit = u.one
 
     def _mask_out_bounds(self, invalid):
         return (
--- a/gammapy/irf/edisp/core.py
+++ b/gammapy/irf/edisp/core.py
@@ -48,6 +48,7 @@ class EnergyDispersion2D(IRF):
 
     tag = "edisp_2d"
     required_axes = ["energy_true", "migra", "offset"]
+    unit = u.Unit("")
 
     def _mask_out_bounds(self, invalid):
         return (
--- a/gammapy/irf/effective_area.py
+++ b/gammapy/irf/effective_area.py
@@ -64,7 +64,7 @@ class EffectiveAreaTable2D(IRF):
 
     tag = "aeff_2d"
     required_axes = ["energy_true", "offset"]
-    unit = u.Unit("m**2")
+    default_unit = u.m**2
 
     def plot_energy_dependence(self, ax=None, offset=None, **kwargs):
         """Plot effective area versus energy for a given offset.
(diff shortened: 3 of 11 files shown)


gammapy/irf/core.py
```diff
@@ -45,6 +45,9 @@ def __init__(self, axes, data=0, unit="", meta=None, interp_kwargs=None):
         if isinstance(data, u.Quantity):
             self.data = data.value
             self.unit = data.unit
+            if self.axes.names == ["energy_true", "migra", "offset"] :
+                if self.unit != u.dimensionless_unscaled:
+                    raise ValueError(f"Error: {self.unit} is not an allowed unit. EDisp requires dimensionless data quantities!")
```
maxnoe, line 50: So I guess None of the IRFs specified units?

Then we should just add a `unit` attribute to all of the classes and use that here to check using:
```
if not self.unit.is_equivalent(data.unit):
    raise ValueError(...)
```

adonath: Yes, this should be done in a general way. There is already a `IRF.unit` attribute, which specifies the currently used unit, bot not the expected one. I think the most elegant solution would be introduce a class level attribute `.default_unit` or similar, which specifies the expected unit equivalency for the data.

fabiopintore: Thanks @maxnoe @adonath , I added the default units for all the IRF extensions (Edisp, Aeff, BKG and PSF) and I also modified the tests accordingly.
Unfortunately, I've read the last message only after pushing so I called the default units simply as `unit`...

adonath: Yes, please adapt the name to `.default_unit` and add the check like `u.Unit(val).is_equivalent(self.default_unit)` here: https://github.com/gammapy/gammapy/blob/master/gammapy/irf/core.py#L145

gammapy/irf/effective_area.py
```diff
@@ -64,6 +64,7 @@ class EffectiveAreaTable2D(IRF):
 
     tag = "aeff_2d"
     required_axes = ["energy_true", "offset"]
+    unit = u.Unit("m**2")
```
maxnoe, line 67: Better use the defined units like this: `u.m**2`. No reason to go through string parsing.

gammapy/irf/edisp/core.py
```diff
@@ -48,6 +48,7 @@ class EnergyDispersion2D(IRF):
 
     tag = "edisp_2d"
     required_axes = ["energy_true", "migra", "offset"]
+    unit = u.Unit("")
```
maxnoe, line 51: `u.one`

gammapy/irf/core.py
maxnoe: Shouldn't that setter transform any existing data?

At the moment (as is often the case when having this getter/setter setups) the user can easily create invalid data like this:

```
aeff = EffectiveArea2D(data=[[1, 2, 3, 4]] * u.m**2)
aeff.unit = u.cm**2  # I would expect that to *convert*, not *overwrite*
```

adonath: Good point. Thinking about it again we should probably remove the setter completely? This is the behavior for `Quantity`:
```
q = 5 * u.TeV
q.unit = "GeV"
```
Raises `AttributeError: can't set attribute`, which makes sense to me...

adonath: Instead for maps we have: https://github.com/gammapy/gammapy/blob/master/gammapy/maps/core.py#L1506 Which mirrors a bit the `Quantity.to()` API.

fabiopintore: Sorry,  it's not completely clear to me what's the final suggestion here, i.e. if I have to modify the `unit(self, val)` introducing some filters or not 😅.
For now, I've just implemented everywhere the simplest change (i.e. `default_unit`).

adonath: The use cases to actually change the unit are more than rare, but it might still be reasonable to support it...

adonath: @fabiopintore I think for this PR we probably keep as is, but we might need a follow up discussion to solve the question whether units should be settable at all (for `IRF` as well as `Map`). Right now they are which is fine, as it still a well define behavior. But I agree with @maxnoe that it is maybe unexpected and could cause more harm than benefit (especially because it is different behavior than for `Quantity`)

maxnoe: If you *really* want to do it, you could do `map.quantity = u.Quantity(map.data, u.unit)` but then it's also clear that this is kind of dangerous.

So +1 for removing the unit setter (and probably even the data setter).

gammapy/irf/psf/tests/test_table.py
```diff
@@ -12,6 +14,21 @@ def psf_3d():
...
+    offset_axis = MapAxis.from_edges([0, 1, 2], unit="deg", name="offset")
+
+    rad_axis = MapAxis.from_edges([0, 1, 2], unit="deg", name="rad")
+
+    wrong_unit = u.cm**2 * u.s
+    data = np.ones((energy_axis.nbin, offset_axis.nbin, rad_axis.nbin)) * wrong_unit
+    with pytest.raises(ValueError) as error:
+        psf3d = PSF3D(axes=[energy_axis, offset_axis, rad_axis],
+                 data=data)
+        assert error == (f"Error: {wrong_unit} is not an allowed unit. {psf3d.tag} requires dimensionless data quantities!")
```
registerrier, line 29: is it dimensionless or per `sr`?

registerrier (changes requested):
Thanks @fabiopintore . Checking units homogeneity is indeed important. 

There is some unclear. Why does the warning always states dimensionless quantity when the arguments list `Quantity` as a valid input. Do we want to only support `ndarray` + `Unit` initializer? If yes this should adapted accordingly.

gammapy/irf/psf/tests/test_table.py
maxnoe: No, a psf 3d should have unit `sr^-1`, https://gamma-astro-data-formats.readthedocs.io/en/latest/irfs/full_enclosure/psf/psf_table/index.html

gammapy/irf/core.py
```diff
@@ -44,7 +44,10 @@ def __init__(self, axes, data=0, unit="", meta=None, interp_kwargs=None):
         self._axes = axes
         if isinstance(data, u.Quantity):
             self.data = data.value
-            self.unit = data.unit
+            if not self.default_unit.is_equivalent(data.unit):
+                raise ValueError(f"Error: {data.unit} is not an allowed unit. {self.tag} requires dimensionless data quantities!")
```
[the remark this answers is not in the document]
maxnoe: No, that is probably a copy/paste of the earlier version, it should be `{self.default_unit}` in the error message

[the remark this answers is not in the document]
fabiopintore: Yes, indeed. It was a mistake. It was there left by the first version. I've fixed it!

fabiopintore:
Thank you @registerrier . The "dimensionless" word in the error message was a mistake due to a previous commit. I've corrected it and I've also modified some tests that were not good enough.
 
The only test that is not passing is on the `BKG3D` (`test_bkg_3d_wrong_units`), and I don't understand why. It mentions a `Did you mean to `re.escape()` the regex?` error, which I don't know how to avoid...

maxnoe:
@fabiopintore `match` uses a regular expression match to see if the error is what you expect. Many characters (e.g. `()`) have special meaning in regular expressions.

Since you define the exact error message, you should just compare the message, not use `match` or define a regex that is actually matching, e.g. 
```
assert error.match(f"Error: (.*) is not an allowed unit. (.*) requires (.*) data quantities.")```

**Comment by fabiopintore** [comment]
Ah fantastic, I've got it. Very useful, @maxnoe

File: gammapy/irf/tests/test_background.py [hunk 6]
```diff
@@ -220,6 +220,39 @@ def test_background_3d_read_gadf():
...
+
+    fov_lat = [0, 1, 2, 3] * u.deg
+    fov_lat_axis = MapAxis.from_edges(fov_lat, name="fov_lat")
+
+    wrong_unit = u.cm**2 * u.s
+    data = np.ones((2, 3, 3)) * wrong_unit
+    with pytest.raises(ValueError) as error:
+        Background3D(axes=[energy_axis, fov_lon_axis, fov_lat_axis],
+                 data=data)
+    assert error.match(f"Error: (.*) is not an allowed unit. (.*) requires (.*) data quantities.")
```
registerrier, line 238: This probably does not need the f-string. Apparently this breaks the py flakes CI.

registerrier (approved):
Thanks @fabiopintore ! This looks good. I just left a small inline comment.

registerrier (approved):
Thanks @fabiopintore

PR #3778, opened 2022-02-01
Speed docs building process 

Bultako:
This PR speeds the docs building process by running the needed execution of notebooks in parallel. The documentation building process actually happens as a github action in the `gammapy-docs` repo, though it may also be run locally with the `make docs-all` command.

Tests done locally reduce the time from 19min to 7 min. approx.


docs/tutorials/analysis/1D/cta_sensitivity.ipynb
```diff
@@ -2,6 +2,7 @@
  "cells": [
   {
    "cell_type": "markdown",
+   "id": "fbb1c19b",
```
adonath, line 5: Is there anything we can do about these "random" meta updates, that just depend on the `jupyter` versions?

Bultako: @adonath 

You can also see in this PR that the `nbformat` version metadata have been modified for all notebooks to the values below.

```
  "nbformat": 4,
  "nbformat_minor": 4
```

The "id" tags in the cells are a consequence of the upgrading of `nbformat` in the notebooks. Since we unpinned the version of `nbformat` in the `environement-dev.yaml` file [here](https://github.com/gammapy/gammapy/pull/3773), this upgrading of notebooks is needed if we want to avoid warnings in the step of notebooks manipulation (adding box for links and binder, create link to API ref, etc.),.

It is completely harmless.
These ids and their values will not change from one doc building to the next if the notebooks are not modified.

adonath: OK, thanks for the explanation. My biggest worry here is that the diffs become hard to review and long-term we might blow up the size of the `.git` folder with these updated which are basically unrelated to the actual content. I guess the only solution would to change to a text-based notebook format...

Bultako: This is not a problem.
The id tags and values do not change when the notebook are re-executed. But if we add a new cell, this one will be added with an id tag.

adonath: But if not all developers are on the same `nbformat` version, another PR will remove them again, so we should definitely make sure this is not happening. All in all this is not ideal...I hope we can change to a text based format in future.

Bultako: Yes.

Anyhow, we have a development environment `environment-dev.yml` with the version of the `nbformat` dependency unpinned. This is also the environment of the CI, developers working with old versions of `nbformat` would not comply with this development environment and modifs in the metadata of the notebeook would show up in the PR.

The other option would we to remove the id tags and pin the version of `nbformat`

adonath (commented):
Thanks @Bultako! 

I just noticed one more thing: the `notebooks-dev.tar` is actually committed (but already in the current master as well). I think this should not be the case, right? The tar file will always be downloaded from the docs server i.e. `gammapy-docs`. The current updated adds 100KB... doing this 100 times, we will increase the Gammapy repo size by 10MB for no good reason. Is this correct? Shouldn't we instead remove the tar file from the code base completely and add it to `.gitignore`?

Bultako:
@adonath 
you're right, the `notebooks-dev.tar` file plays the same role than the html files created with sphinx, they only live in the versioned docs in the `gammapy-docs` repo. I will remove it from the code base and add it to `.gitignore`

adonath (dismissed):
Thanks @Bultako, no further comments from my side...

The `id` seems to be required meta data for `nbformat` since 4.5, so it raises a warning if it is not present and we should probably introduce it to make sure `nbformat` keeps working.

Bultako:
@adonath 

I have made some last minor modifs in this PR
I have fixed the `.gitignore`, removed an info box from `analysis_1.ipynb` notebook that was left somehow(?) by mistake, and shorten the message added in that info box of the html version of the notebooks to have a better display in the new docs theme.

from my side this PR is ready.

PR #3780, opened 2022-02-01
Add PYTHONNOUSERSITE=1 to conda env

maxnoe:
This prevents the user site packages (where `pip install --user` installs to) to take precedence over the packages in the environment.

This is a long standing issue with conda, that this is not done by default (unlike e.g. python virtualenvs, which do this by default).

### DIFF
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -11,6 +11,9 @@ channels:
   - conda-forge
   - sherpa
 
+variables:
+  PYTHONNOUSERSITE: "1"
+
 dependencies:
   # core dependencies
   - python=3.7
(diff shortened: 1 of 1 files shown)


adonath:
Thanks @maxnoe!

Here is the oldest discussion I could find as reference for this `https://github.com/conda/conda/issues/394`. I agree this would be the preferred default. 

Our other environment files are here: https://github.com/gammapy/gammapy-webpage/tree/gh-pages/download/install and could be adapted as well.

adonath (approved):
Thanks @maxnoe, I have no further comments.

PR #3782, opened 2022-02-01
Minor update in dev HowTo

AtreyeeS:
I was getting an error `zsh: no matches found: -glob=*/tests` while testing doc strings. This was the solution I found after some searching, so adding it. I don't understand the issue though...


registerrier (approved):
Thanks @AtreyeeS

PR #3785, opened 2022-02-04
Improve doc strings for Gammapy.data

AtreyeeS:
follows from #3748

### DIFF
--- a/gammapy/data/data_store.py
+++ b/gammapy/data/data_store.py
@@ -41,7 +41,7 @@ class DataStore:
     >>> data_store.info()
     Data store:
     HDU index table:
-    BASE_DIR: /home/runner/work/gammapy/gammapy/gammapy-datasets/hess-dl3-dr1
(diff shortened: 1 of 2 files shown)


gammapy/data/data_store.py
```diff
@@ -41,7 +41,7 @@ class DataStore:
     >>> data_store.info()
     Data store:
     HDU index table:
-    BASE_DIR: /home/runner/work/gammapy/gammapy/gammapy-datasets/hess-dl3-dr1
+    BASE_DIR: /Users/ASinha/Gammapy-dev/gammapy-data/hess-dl3-dr1
```
registerrier, line 44: I suppose this should be left untouched but skipped at CI, no?

registerrier (commented):
Thanks @AtreyeeS ! 

There are apparently CI failures. One is connected to an output which shouldn't be tested. The other to a unit issue with `plot_offset2_distribution` which I don't really understand.

PR #3788, opened 2022-02-08
Alternative interface for the hli 

QRemy:
This PR changes how to call the different methods of the analysis class and allows their execution from the cli. 
- the methods .get_xx are replace by specific AnalysisStep classes and can be defined in the configuration (for example `config.general.steps = ["data-reduction", "fit"]`) or called interactively by `analysis.run(["data-reduction", "fit"])`
- by default `analysis.run()` execute the workflow defined in the configuration file, that is set to ["data-reduction", "excess-map", "fit", "flux-points"].
- in the cli `gammapy analysis run` is now equivalent to `analysis.run()` so it can execute all the analysis step available and not only data reduction.
- add `datasets_file`, `models_file`, `overwrite`, `n_jobs`  options in the configuration file 
- add a read/write_datasets methods in the analysis class in order to  restart from files.
-add severals TODO about various configuration options/features that should be added


adonath:
Thanks @QRemy! Would you mind splitting the changes? I think introducing the `DatasetsMaker` is completely independent and worth a separate PR. Thanks!

QRemy:
Splited in multiple PR: #3802, #3820, #3852 so this draft can be closed.

PR #3791, opened 2022-02-10
Include ASTRI mini-array in the list of observatory locations

fabiopintore:
`~gammapy.data.observers` lists the Earth location of the main gamma-ray observatories. This PR includes the location of the future ASTRI mini-array.


registerrier (commented):
Thanks @fabiopintore .

I assume `lon` and `lat` are inverted and `lon` should be negative (-16d). ASTRI is in Tenerife, right?

fabiopintore:
Thanks @registerrier . You're right, I've inverted the coordinates!

registerrier (approved):
Thanks @fabiopintore !

PR #3793, opened 2022-02-10
Implement solid_angle for CompoundSkyRegion by going through pixel mask

maxnoe:
**Description**

As discussed here #3737 and #3733, it would be good to return a region geom directly from the regions finder and use the solid angle to calculate the alpha value. This needs `solid_angle` support for `CompoundSkyRegion` that unfortunately does not implement the `area` property in `regions` but which was easy to implement by going through the pixel mask.

As proposed by @adonath, I use oversampling (hardcoded factor of 10 at the moment) to increase the precision of the calculated area beyond the pixel level of the map.

### DIFF
--- a/gammapy/maps/region/geom.py
+++ b/gammapy/maps/region/geom.py
@@ -64,15 +64,15 @@ class RegionGeom(Geom):
 
         if wcs is None and region is not None:
             if isinstance(region, CompoundSkyRegion):
-                center = compound_region_center(region)
+                self._center = compound_region_center(region)
             else:
-                center = region.center
+                self._center = region.center
 
             wcs = WcsGeom.create(
                 binsz=binsz_wcs,
-                skydir=center,
+                skydir=self._center,
                 proj=self.projection,
-                frame=center.frame.name,
+                frame=self._center.frame.name,
             ).wcs
 
         self._wcs = wcs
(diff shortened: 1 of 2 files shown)


gammapy/maps/region/tests/test_geom.py
```diff
@@ -198,6 +198,35 @@ def test_solid_angle(region):
...
+
+    # fully overlapping regions, expect only area of the larger one
+    expected = RegionGeom.create(region1).solid_angle()
+    region = region1 | region2
+    assert isinstance(region, CompoundSkyRegion)
+
+    geom = RegionGeom.create(region)
+    omega = geom.solid_angle()
+
+    assert u.isclose(omega, expected, rtol=2e-3)
```
maxnoe, line 227: The actual difference for this testcase is `1.1e-3`

gammapy/maps/region/geom.py
```diff
@@ -301,7 +301,25 @@ def solid_angle(self):
...
             raise ValueError("Region definition required.")
 
-        area = self.region.to_pixel(self.wcs).area
+        # compound regions do not implement area()
+        # so we use the mask represenation and estimate the area
+        # from the pixels in the mask using oversampling
+        if isinstance(self.region, CompoundSkyRegion):
+            # oversample by a factor of ten
+            oversampling = 10.0
+            wcs = WcsGeom.create(
```
adonath, line 310: Here is a nitpicky thing: `RegionGeom` also takes optionally a `wcs` object as input, which defines a custom projection, if users want this. Here you create a new one, which is not necessarily aligned (likely not...). However you can just use `self.to_wcs_geom()`, which compute the minimal "enclosing" WCS geometry. Instead of counting pixels then you could also sum up the bin wise solid angle returned by `WcsGeom.solid_angle()`.

maxnoe: So do you suggest to add an oversampling option to `to_wcs_geom`? I didn't use it because I need to create a different but similar `WCSGeom` with finer bins.

maxnoe: Ah, there is also `to_binsz_wcs`, which seems to work fine.

adonath: Yes, or you can use `WcsGeom.upsample()` as well.

adonath (approved):
Thanks @maxnoe, I have no further comments for now. We could maybe open a feature request in https://github.com/astropy/regions to actually support `CompoundRegion.solid_angle()` natively...

adonath:
See https://github.com/astropy/regions/issues/433

PR #3796, opened 2022-02-11
Helper function to rebin map axis (new PR)

luca-giunti:
**Description**
This pull request supersedes #3745. I followed the implementation proposed by @adonath (https://github.com/gammapy/gammapy/pull/3745#issuecomment-1030176642), and moved the `spectrum_dataset()` fixture to `gammapy.conftest.py` to share it both with the `gammapy.datasets.tests.test_spectrum` and `gammapy.estimators.tests.test_utils`.

### DIFF
--- a/gammapy/conftest.py
+++ b/gammapy/conftest.py
@@ -61,7 +61,7 @@ def pytest_configure(config):
         pass
 
 
-@pytest.fixture()
+@pytest.fixture(scope="session")
 def spectrum_dataset():
     name = "test"
     energy = np.logspace(-1, 1, 31) * u.TeV
--- a/gammapy/conftest.py
+++ b/gammapy/conftest.py
@@ -2,8 +2,21 @@
 # this contains imports plugins that configure py.test for astropy tests.
 # by importing them here in conftest.py they are discoverable by py.test
 # no matter how it is invoked within the source tree.
+import pytest
 import os
+import numpy as np
 from pytest_astropy_header.display import PYTEST_HEADER_MODULES
+from gammapy.modeling.models import (
+    Models,
+    PowerLawSpectralModel,
+    ConstantTemporalModel,
+    SkyModel,
+)
+from gammapy.maps import MapAxis, RegionNDMap
+from astropy.time import Time
+from gammapy.data import GTI
+import astropy.units as u
+from gammapy.datasets import SpectrumDataset
 
 # TODO: activate this again and handle deprecations in the code
 # enable_deprecations_as_exceptions(warnings_to_ignore_entire_module=["iminuit", "naima"])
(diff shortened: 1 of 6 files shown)


gammapy/datasets/spectrum.py
```diff
@@ -213,6 +213,12 @@ class SpectrumDataset(PlotMixin, MapDataset):
     stat_type = "cash"
     tag = "SpectrumDataset"
 
+    @property
+    def energy_range(self):
```
adonath, line 217: I think this could be useful for `MapDataset` as well. If we rename e.g. to `energy_range_quantity` (or similar), then it would return `np.min(energy_min_map.quantity)` and `np.max(energy_min_map.quantity)`.

gammapy/estimators/utils.py
```diff
@@ -157,3 +155,71 @@ def estimate_exposure_reco_energy(dataset, spectral_model=None, normalize=True):
...
+        "npred_background_min",
+        "npred_signal_min",
+    ]
+    for key in conditions.keys():
+        if key not in available_conditions:
+            raise ValueError(
+                f"Unrecognized option {key}. The available methods are: {available_conditions}."
+            )
+
+    def _satisfies_conditions(info_dict):
```
adonath, line 198: I think this could be moved outside and just take `conditions` as well. This make the code structure probably a bit simpler...

luca-giunti: @adonath I did move it outside but didn't really get how not to pass it the `info_dict`

adonath: What you did is exactly what I meant!

gammapy/estimators/utils.py
```diff
@@ -157,3 +155,72 @@ def estimate_exposure_reco_energy(dataset, spectral_model=None, normalize=True):
...
+    conditions : dict
+        Keyword arguments containing the per-bin conditions used to resample the axis.
+        Available options are: 'counts_min', 'background_min', 'excess_min', 'sqrt_ts_min', 'npred_min',
+        'npred_background_min', 'npred_signal_min'.
+    Returns
+    -------
+    energy_edges : list of `~astropy.units.Quantity`
+        Energy edges for the resampled energy axis.
+    """
+    if not isinstance(dataset, SpectrumDataset) or isinstance(
```
adonath, line 183: Minor: this could be written shorter as `isinstance(dataset, (SpectrumDataset, SpectrumDatasetOnOff))`

adonath (dismissed):
Thanks a lot @luca-giunti, I have no further comments except that this will require some more documentation. But this can be addresses in future PRs.

gammapy/datasets/map.py
```diff
@@ -381,6 +381,12 @@ def energy_range_fit(self):
         """Energy range maps defined by the mask_fit only."""
         return self._energy_range(self.mask_fit)
 
+    @property
+    def energy_range_total(self):
```
registerrier, line 385: I think this method won't work as soon as you have a `nan` in a mask. How does it work on full `MapDataset`?
Maybe move it to an internal function?

gammapy/estimators/utils.py
```diff
@@ -157,3 +155,72 @@ def estimate_exposure_reco_energy(dataset, spectral_model=None, normalize=True):
         reco_exposure = reco_exposure / ref_flux[:, np.newaxis, np.newaxis]
 
     return reco_exposure
+
+
+def _satisfies_conditions(info_dict, conditions):
```
registerrier, line 160: If this function is used only in `resample_energy_edges` you might as well define it inside the function.

gammapy/estimators/utils.py
```diff
@@ -157,3 +155,72 @@ def estimate_exposure_reco_energy(dataset, spectral_model=None, normalize=True):
...
+    satisfies = True
+    for key in conditions.keys():
+        satisfies &= info_dict[key.strip("_min")] > conditions[key]
+    return satisfies
+
+
+def resample_energy_edges(dataset, conditions={}):
+    """Returns energy edges that defining a binning that satisfies given conditions on the per-bin statistics.
```
registerrier, line 168: I think we always use imperative mood for docstrings. So "Return energy edges that satisfies given condition on the per bin statistics"

registerrier (dismissed):

I have left few inline comments. Note that this might also need some addition in the dev doc to explain fixture usage in tests.

gammapy/estimators/utils.py
adonath: I commented before to move it outside, just because I thought `resample_energy_edges` was already complex enough...

gammapy/conftest.py
```diff
@@ -46,3 +59,45 @@ def pytest_configure(config):
         print('Setting matplotlib backend to "agg" for the tests.')
     except ImportError:
         pass
+
+
+@pytest.fixture()
```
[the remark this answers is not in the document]
adonath: Yes, the correct scope here would be `scope="session"`

gammapy/datasets/map.py
adonath: @luca-giunti Could implement this change to handle NaN values?

luca-giunti:
@adonath do you know why after changing the scope of the `spectrum_dataset` fixture to `"session"` some fitting tests are failing? Does `"session"` mean that the dataset is created only once and then each test can modify it in memory? Or each test has access to the dataset originally created by the fixture

adonath:
@luca-giunti Indeed "session" means that the dataset is just created once. If we currently have tests, that modify the datasets in place, this will lead to test fails. However we should probably never modify fixtures in place, this is a bad practice in any case. SO maybe go back to the normal scope for now and leave a TODO comment?

adonath (approved):
Thanks @luca-giunti, no further comments from my side. This method will require som documentation in a follow up PR.

Issue #3767, opened 2022-01-28
Separate internal data model from specific serialization formats


maxnoe:
Currently, gammapy's internal data structures are very deeply intertwined with the first version(s) of the GADF specification. 

Mostly, data is read as is from FITS into astropy tables and then throughout the code, the data originally stored in the FITS headers is accessed via `table.meta`.

This has several problems:

* No easy support of multiple formats, since the internal data structure is tied to one specific format
* Data is not in the optimal in-memory representation, some examples:
   * observation begin / end (TSTART / TSTOP) should be `astropy.time.Time` instances, not the internal datetime format of the DL3 files
   * quantities in general are pure numbers, not `astropy.units.Quantity`
* Data is not validated on input. Errors can happen deep into the code for something that could have been caught on input file reading.
* Writing data out is harder, because no clear assumptions can be made about the data in `.meta`: it's a mixture of what existed in the input file (if any) and what is set by the user and what is created by gammapy.

We should define the internal data model, via the corresponding data classes (EventList, IRFs, etc. ) and have a clear IO boundary between internal and external data representations.

maxnoe:
Starting the evaluation of what is required by gammapy and how that would be best represented in memory and where that information is currently read from 

### Observation

The current observation relies mostly on the `obs_info` metadata dict, that is taken either from the header of the `EVENTS` table when directly reading a single observation or it is taken from the `OBS-INDEX` table.

As discussed in the data model meeting, this is problematic for two reasons:
* This information is mandatory in the `EVENTS` header but optional in the `OBS-INDEX` table. Right now, in the standard analysis use case of gammapy, going through datastore, the information is only accessed from the optional location.
* There might be inconsistencies between OBS-INDEX and EVENTS header, then the event header should be the one source of truth.


Looking at the places in gammapy that use obs-info, I identified these properties are used in the analysis:

### Observatory location

Read from GEOLON / GEOLAT / (GEOALT or ALTITUDE).

Note: this is incompatible with the current FITS standard and should be expanded to also support the current standard using the `OBSGEO-{X,Y,Z}` keywords:

> The location, if specified, shall be represented as a geocentric Cartesian triple with respect to a standard ellipsoidal geoid at the time of the observation

Or alternatively:

> A set of geodetic coordinates is recognized by the following keywords OBSGEO-{B,L,H}

The in memory representation should be a `astropy.coordinates.EarthLocation`.

[...]

PR #3797, opened 2022-02-11
Refactor pointing information handling and `Observation.create`

maxnoe:
**Description**

As discussed here #3767, the `Observation` class duplicated in a less correct manner code that is implemented in `FixedPointingInfo`. 

Here I remove the code and change the properties to look up the corresponding properties in the `fixed_pointing_info`, making that a `lazyproperty`.

This also made it necessary, to require the `locatation` for all observations. This was not always given, mostly in simulation examples. However, I think requiring the observatory location makes sense in general so I added it where needed.

### DIFF
--- a/gammapy/data/observations.py
+++ b/gammapy/data/observations.py
@@ -6,8 +6,9 @@ import numpy as np
 from astropy.coordinates import SkyCoord
 from astropy.time import Time
 from astropy.units import Quantity
-from gammapy.utils.fits import LazyFitsData, earth_location_from_dict
+from gammapy.utils.fits import LazyFitsData
 from gammapy.utils.testing import Checker
+from astropy.utils import lazyproperty
 from .event_list import EventList, EventListChecker
 from .filters import ObservationFilter
 from .gti import GTI
@@ -240,35 +241,31 @@ class Observation:
         """
         return 1 - self.obs_info["DEADC"]
 
-    @property
-    def pointing_radec(self):
-        """Pointing RA / DEC sky coordinates (`~astropy.coordinates.SkyCoord`)."""
-        lon, lat = (
-            self.obs_info.get("RA_PNT", np.nan),
-            self.obs_info.get("DEC_PNT", np.nan),
-        )
-        return SkyCoord(lon, lat, unit="deg", frame="icrs")
-
-    @property
-    def pointing_altaz(self):
-        """Pointing ALT / AZ sky coordinates (`~astropy.coordinates.SkyCoord`)."""
-        alt, az = (
-            self.obs_info.get("ALT_PNT", np.nan),
-            self.obs_info.get("AZ_PNT", np.nan),
-        )
-        return SkyCoord(az, alt, unit="deg", frame="altaz")
-
-    @property
-    def pointing_zen(self):
-        """Pointing zenith angle sky (`~astropy.units.Quantity`)."""
-        return Quantity(self.obs_info.get("ZEN_PNT", np.nan), unit="deg")
-
-    @property
+    @lazyproperty
     def fixed_pointing_info(self):
         """Fixed pointing info for this observation (`FixedPointingInfo`)."""
         return FixedPointingInfo(self.events.table.meta)
 
     @property
+    def pointing_radec(self):
+        """Pointing RA / DEC sky coordinates (`~astropy.coordinates.SkyCoord`)."""
+        return self.fixed_pointing_info.radec
+
+    @property
+    def pointing_altaz(self):
+        return self.fixed_pointing_info.altaz
+
+    @property
+    def pointing_zen(self):
+        """Pointing zenith angle sky (`~astropy.units.Quantity`)."""
+        return self.fixed_pointing_info.altaz.zen
+
+    @property
+    def observatory_earth_location(self):
+        """Observatory location (`~astropy.coordinates.EarthLocation`)."""
+        return self.fixed_pointing_info.location
+
+    @lazyproperty
     def target_radec(self):
         """Target RA / DEC sky coordinates (`~astropy.coordinates.SkyCoord`)."""
         lon, lat = (
@@ -276,11 +273,6 @@ class Observation:
             self.obs_info.get("DEC_OBJ", np.nan),
         )
         return SkyCoord(lon, lat, unit="deg", frame="icrs")
-
-    @property
-    def observatory_earth_location(self):
-        """Observatory location (`~astropy.coordinates.EarthLocation`)."""
-        return earth_location_from_dict(self.obs_info)
 
     @property
     def muoneff(self):
--- a/gammapy/data/observations.py
+++ b/gammapy/data/observations.py
@@ -6,8 +6,10 @@ import numpy as np
 from astropy.coordinates import SkyCoord
 from astropy.time import Time
 from astropy.units import Quantity
-from gammapy.utils.fits import LazyFitsData
+import astropy.units as u
+from gammapy.utils.fits import LazyFitsData, earth_location_to_dict
 from gammapy.utils.testing import Checker
+from gammapy.utils.time import time_ref_to_dict, time_relative_to_ref
 from astropy.utils import lazyproperty
 from .event_list import EventList, EventListChecker
 from .filters import ObservationFilter
@@ -125,25 +127,34 @@ class Observation:
         return gti
 
     @staticmethod
-    def _get_obs_info(pointing, deadtime_fraction):
+    def _get_obs_info(pointing, deadtime_fraction, time_start, time_stop, reference_time, location):
         """Create obs info dict from in memory data"""
-        return {
+        obs_info = {
             "RA_PNT": pointing.icrs.ra.deg,
             "DEC_PNT": pointing.icrs.dec.deg,
             "DEADC": 1 - deadtime_fraction,
         }
+        obs_info.update(time_ref_to_dict(reference_time))
+        obs_info['TSTART'] = time_relative_to_ref(time_start, obs_info).to_value(u.s)
+        obs_info['TSTOP'] = time_relative_to_ref(time_stop, obs_info).to_value(u.s)
+
+        if location is not None:
+            obs_info.update(earth_location_to_dict(location))
+
+        return obs_info
 
     @classmethod
     def create(
         cls,
         pointing,
+        location=None,
         obs_id=0,
         livetime=None,
         tstart=None,
         tstop=None,
         irfs=None,
         deadtime_fraction=0.0,
-        reference_time="2000-01-01",
+        reference_time=Time("2000-01-01 00:00:00"),
     ):
         """Create an observation.
 
@@ -157,11 +168,13 @@ class Observation:
             Observation ID as identifier
         livetime : ~astropy.units.Quantity`
             Livetime exposure of the simulated observation
-        tstart : `~astropy.units.Quantity`
-            Start time of observation w.r.t reference_time
-        tstop : `~astropy.units.Quantity` w.r.t reference_time
-            Stop time of observation
-        irfs : dict
+        tstart: `~astropy.time.Time` or `~astropy.units.Quantity`
+            Start time of observation as `~astropy.time.Time` or duration
+            relative to `reference_time`
+        tstop: `astropy.time.Time` or `~astropy.units.Quantity`
+            Stop time of observation as `~astropy.time.Time` or duration
+            relative to `reference_time`
+        irfs: dict
             IRFs used for simulating the observation: `bkg`, `aeff`, `psf`, `edisp`
         deadtime_fraction : float, optional
             Deadtime fraction, defaults to 0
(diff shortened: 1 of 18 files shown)


maxnoe:
This fails, because the tests are accessing these properties on simulated observations, where things required by `FixedPointingInfo` are missing.

I am not clear on how to proceed now. These properties are part of the public API but they are not used internally for doing analysis.

Should I adapt the tests? Or should I try to somehow keep the backwards compatibility?

One consistency issue is that `FixedPointingInfo.altaz` does not look at `ALT_PNT`, `AZ_PNT` but transforms the `RA_PNT`, `DEC_PNT` into icrs using the mid of run time. `Observation.pointing_altaz` would use `ALT_PNT` / `AZ_PNT`.

This is the kind of inconsistency that removing the duplication would avoid, but then we have to take a decision here.

gammapy/data/observations.py
```diff
@@ -240,35 +241,31 @@ def observation_dead_time_fraction(self):
         """
         return 1 - self.obs_info["DEADC"]
 
+    @lazyproperty
+    def fixed_pointing_info(self):
+        """Fixed pointing info for this observation (`FixedPointingInfo`)."""
+        return FixedPointingInfo(self.events.table.meta)
```
adonath, line 247: For simulations the event meta will never be there from the beginning. Doesn't it work to just change to `FixedPointingInfo(self.obs_info)` for now and then later improve the handling of the "in-memory" meta and `events.table.meta`, that we planned to do anyway?

maxnoe: That didn't work since the obs info table used for obs info in case I going through datastore does not contain the mjdref values. I think these are in the header of the obs table but are not added to the o s info at the moment.

adonath: I think adding the `MJDREF` info from the observation table header is probably fine for now? It should be a mandatory keyword...

adonath: And for the simulation case one could adapt the `Observation._get_obs_info`...

adonath: Just to add: I think This will not be any final solution, but it will probably help to identify and improve the internal data model.

maxnoe:
The two tests that are failing now are 

* Test with the hess dataset. The `FixedPointingInfo` returns the altaz pointing transformed from the radec at the midpoint of the run, while before `Observation.pointing_altaz` would use `ALT_PNT` and `AZ_PNT`, which it seems are filled by HESS for the start of the run, not the midpoint / average.

* `test_cta_observation_cta_1d`. This test is strange to me. It tests `Observation.create`? By just loading IRFs, not events. Most information required for analysis is not defined and set into the observation. The resulting observation is certainly unusable for mostly anything. I am not sure what this test is supposed to test.

adonath:
Thanks a lot @maxnoe, I think the change you made in your last commit go exactly in the right direction. I'll try look at the remaining test fail and comment again...

adonath:
I looked into remaining test fail and there is a "hot" pixel outside the valid range of the background model, both in offset and energy. This probably appeared in this PR, because the alt-az position slightly changed compared to the exported HESS value and the background is evaluated slightly different coordinates? So it is likely the same reason as the changed test value in `test_observation`. The only way to understand this in more detail would be check how the alt-az coordinates are computed in HESS (maybe @registerrier or @bkhelifi knows?). 

By now I would trust the Astropy coordinate transform. It is well tested against other libraries. So I think we should keep the change in this PR. To fix the test one could add and energy offset cut to the `safe_mask_maker_bkg_invalid` test as a workaround...

maxnoe:
> The only way to understand this in more detail would be check how the alt-az coordinates are computed in HESS (maybe @registerrier or @bkhelifi knows?).

I think this is the difference between using the mid-of run timepoint now vs. the start of run that seems to be filled into `ALT_PNT` / `AZ_PNT`

adonath:
@lmohrmann Can you maybe comment how the ALT_PNT / AZ_PNT keywords are filled in the FITS exporter? Is it the start time or any other time?

maxnoe:
Seems to neither be the start nor the midpoint but something in between:


```
from astropy.coordinates import SkyCoord, EarthLocation, AltAz

obs = Observation.read('./GAMMAPY_DATA/hess-dl3-dr1/data/hess_dl3_dr1_obs_id_020136.fits.gz')

pointing = obs.fixed_pointing_info
location = obs.fixed_pointing_info.location


print(f"ALT_PNT               : {obs.events.table.meta['ALT_PNT']:.4f}°")

altaz_tstart = pointing.radec.transform_to(AltAz(location=location, obstime=obs.tstart))
print(f"From RaDec at tstart  : {altaz_tstart.alt.deg:.4f}°")

midpoint = obs.tstart + 0.5 * (obs.tstop - obs.tstart)
altaz_midpoint = pointing.radec.transform_to(AltAz(location=location, obstime=midpoint))
print(f"From RaDec at midpoint: {altaz_midpoint.alt.deg:.4f}°")

altaz_tstop = pointing.radec.transform_to(AltAz(location=location, obstime=obs.tstop))
print(f"From RaDec at tstop   : {altaz_tstop.alt.deg:.4f}°")
```

```
ALT_PNT               : 51.4870°
From RaDec at tstart  : 52.7327°
From RaDec at midpoint: 51.9039°
From RaDec at tstop   : 50.9365°
```

lmohrmann:
Just checked quickly - I think it is the mean ALT and AZ, i.e. averaged over the event rate, if I'm not mistaken. So it makes sense that it does not match exactly with the start, end, or midpoint time.

adonath:
Thanks @lmohrmann for clarifying!

gammapy/data/observations.py
maxnoe: done

maxnoe:
Converting the radec pointing to altaz using the event time gives something close to the midpoint, not the header value:

```
altaz_events = pointing.radec.transform_to(AltAz(location=location, obstime=obs.events.time))
print(f"Mean event pointing   : {altaz_events.alt.deg.mean():.4f}°")

Mean event pointing   : 51.8915°
```

maxnoe:
And timewise, it looks even stranger: ALT is after the mid point, but az is before:
hess_pointing

lmohrmann:
I think the averaging is done using only events in some ON region. That the average ALT and AZ would correspond to different times is also not entirely inconceivable, is it?

maxnoe:
@lmohrmann yes. But it is kind of strange behavior to have  ALT / AZ pairs that don't correspond to a single point in time.

lmohrmann:
Yes, that is arguably true. We might want to re-think this, but in the end I think these quantities will not be used very often, will they (given that they're not uniquely defined - unless we want to enforce a strict definition, of course)?

maxnoe:
> Yes, that is arguably true. We might want to re-think this, but in the end I think these quantities will not be used very often, > will they (given that they're not uniquely defined - unless we want to enforce a strict definition, of course)?

No, in the schema I proposed for handling this, for pointing observations, always the fixed Ra/Dec coordinates will be transformed to a given time. `ALT_PNT` / `AZ_PNT` are then only used for `DRIFT` observations, where in turn the ra/dec coordinates will alsways be transformed from those.

See the proposal here: https://github.com/gammapy/gammapy/issues/3767#issuecomment-1041467113

registerrier (commented):
Thanks @maxnoe .

One issue here is that the change is API breaking in a commonly used part of gammapy. Is there a way to avoid this?

Maybe the metadata part can be removed from this PR as it is quite independent. We should further discuss the actual format of metadata elsewhere.
 
For some reason, some notebooks are changed here but they are not modified. Could you remove them for the PR?

gammapy/data/meta.py
```diff
@@ -0,0 +1,46 @@
+from dataclasses import dataclass
```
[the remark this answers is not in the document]
maxnoe: Sorry, yes, accidental commit.

gammapy/data/observations.py
```diff
@@ -240,35 +258,34 @@ def observation_dead_time_fraction(self):
...
     def pointing_zen(self):
         """Pointing zenith angle sky (`~astropy.units.Quantity`)."""
-        return Quantity(self.obs_info.get("ZEN_PNT", np.nan), unit="deg")
+        return self.fixed_pointing_info.altaz.zen
 
     @property
-    def fixed_pointing_info(self):
-        """Fixed pointing info for this observation (`FixedPointingInfo`)."""
-        return FixedPointingInfo(self.events.table.meta)
+    def observatory_earth_location(self):
```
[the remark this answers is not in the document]
maxnoe: `FixedPointing` here refers to fixed pointing values given in the header as opposed to the pointing table, that is defined in GADF and for which a class exists in `gammapy.data.pointing`, but which is nowhere used in gammapy for analysis.

So you are right, gammapy currently only supports `FixedPointingInfo`, but this did not change with this PR.

maxnoe:
> One issue here is that the change is API breaking in a commonly used part of gammapy. Is there a way to avoid this?

The API breaking is requiring the location.

One could allow the location to be None and only raise errors when transformations from/to altaz are requested.

docs/tutorials/analysis/3D/analysis_3d.ipynb
```diff
@@ -803,7 +803,7 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "Python 3",
+   "display_name": "Python 3 (ipykernel)",
```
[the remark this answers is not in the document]
maxnoe: Fixed

maxnoe:
> One issue here is that the change is API breaking in a commonly used part of gammapy. Is there a way to avoid this?

The only breaking change is that we now require the location in `Observation.create`. I think it's a very sensible requirement. To avoid the breaking change, there are a couple of possibilities:

* check telescope / instrume headers and use the one location from `gammapy.data.observatory_locations`. This introduces back hidden state and implicitly looking things up *somewhere*, which is why I am not a huge fan of this.
* Do not require location immediately, but raise a meaningful error like `MissingMetaData('Observatory location unknown for observation ...')` as soon as one tries to access something needing it (e.g. the altaz frame)

> Maybe the metadata part can be removed from this PR as it is quite independent. We should further discuss the actual format of metadata elsewhere.

Yes, sorry, that was purely accidental and not meant to be part of this.

> For some reason, some notebooks are changed here but they are not modified. Could you remove them for the PR?

There was one notebook that I opened and saved by accident that didn't need to be changed (analysis_3d), if I am correct, all other do actually have a change (adding the location to `Observation.create`).

maxnoe:
@registerrier I made the `location` option to `Observation.create` optional again (but left it added in the simulation notebooks, since I think it *should* be added) and let the `FixedPointingInfo` fallback to `ALT_PNT/AZ_PNT` and eventually `nan` like `Observation.pointing_altaz` did before.

Does that address your breaking change concerns?

maxnoe:
I now also added back the option to give tstart and tstop as durations relative to `t_ref`. So I think this should be fully backwards compatible now but improved by

* Adding the location to `Observation.create`
* Adding the possibility to give `Time` instances for `tstart`, `tstop` in `Observation.create`
* Removing the duplicated code in the `Observation` and instead using properties looking things up in `fixed_pointing_info`

I think this is a good start for further refactoring the data / metadata coupling.

adonath (dismissed):
Thanks a lot @maxnoe, this look very good to me now! I have only one minor future suggestion:

To make it slightly more convenient for users, once could try to (optionally) extract the location info from the IRF meta data. Some code for this is already available [here](https://github.com/gammapy/gammapy/blob/master/gammapy/datasets/simulate.py#L302), which could be refactored into `earth_location_from_gadf_dict()` or similar. But the PR is fine from my side as is...

adonath:
@registerrier Can you please add your final review here?

docs/tutorials/analysis/1D/spectral_analysis.ipynb
```diff
@@ -110,7 +110,7 @@
     "from regions import CircleSkyRegion\n",
     "from gammapy.maps import MapAxis, RegionGeom, WcsGeom\n",
     "from gammapy.modeling import Fit\n",
-    "from gammapy.data import DataStore\n",
+    "from gammapy.data import DataStore, observatory_locations\n",
```
registerrier, line 113: Here the new import is not necessary

gammapy/data/observations.py
```diff
@@ -124,25 +127,34 @@ def gti(self):
...
         """Create obs info dict from in memory data"""
-        return {
+        obs_info = {
             "RA_PNT": pointing.icrs.ra.deg,
             "DEC_PNT": pointing.icrs.dec.deg,
             "DEADC": 1 - deadtime_fraction,
         }
+        obs_info.update(time_ref_to_dict(reference_time))
+        obs_info['TSTART'] = time_relative_to_ref(time_start, obs_info).to_value(u.s)
+        obs_info['TSTOP'] = time_relative_to_ref(time_stop, obs_info).to_value(u.s)
```
registerrier, line 139: If `obs_info` contains only relative time we have to also store the reference time information, no?

Also it might be simpler to pass the GTI directly.

gammapy/data/observations.py
```diff
@@ -172,15 +186,28 @@ def create(
...
         """
         if tstart is None:
-            tstart = Quantity(0.0, "hr")
+            tstart = reference_time.copy()
 
         if tstop is None:
             tstop = tstart + Quantity(livetime)
 
-        gti = GTI.create([tstart], [tstop], reference_time=reference_time)
+        if not isinstance(tstart, Time):
```
registerrier, line 194: Note that this could be moved to `GTI.create` to make it more general.

registerrier (dismissed):
Thanks @maxnoe .

I have left a few minor comments/questions.

gammapy/data/observations.py
maxnoe: The reference time is stored above in line 137:

`obs_info.update(time_ref_to_dict(reference_time))`

registerrier: My bad. Thanks.

gammapy/data/observations.py
maxnoe: Good idea!

maxnoe: done

docs/tutorials/analysis/1D/spectral_analysis.ipynb
maxnoe: fixed

registerrier (approved):
Thanks @maxnoe !

PR #3802, opened 2022-02-14
Use DatasetsMaker in Analysis class

QRemy:
This PR modifies the analysis class such as it relies internally on the DatasetsMaker in get_datasets().
(implement a part of #3788 separately)

### DIFF
--- a/gammapy/analysis/config.py
+++ b/gammapy/analysis/config.py
@@ -229,6 +229,7 @@ class LogConfig(GammapyBaseConfig):
 class GeneralConfig(GammapyBaseConfig):
     log: LogConfig = LogConfig()
     outdir: str = "."
+    n_jobs: int = 1
 
 
 class AnalysisConfig(GammapyBaseConfig):
--- a/gammapy/analysis/core.py
+++ b/gammapy/analysis/core.py
@@ -19,6 +19,7 @@ from gammapy.makers import (
     RingBackgroundMaker,
     SafeMaskMaker,
     SpectrumDatasetMaker,
+    DatasetsMaker,
 )
 from gammapy.maps import Map, MapAxis, RegionGeom, WcsGeom
 from gammapy.modeling import Fit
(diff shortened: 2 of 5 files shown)


gammapy/analysis/config.py
```diff
@@ -229,6 +229,7 @@ class LogConfig(GammapyBaseConfig):
 class GeneralConfig(GammapyBaseConfig):
     log: LogConfig = LogConfig()
     outdir: str = "."
+    n_jobs: int = 1
```
registerrier, line 232: For now `n_jobs` is used mostly in data reduction, but could be used in later steps so this is likely a valid place to put it. Is that the motivation?

gammapy/analysis/tests/test_analysis.py
```diff
@@ -400,14 +400,14 @@ def test_analysis_3d_joint_datasets():
 
     assert_allclose(
         analysis.datasets[0].background_model.spectral_model.norm.value,
-        1.031743694988066,
+        0.998829, rtol=1e-5
```
registerrier, line 403: Same here?

gammapy/makers/reduce.py
```diff
@@ -102,7 +102,10 @@ def make_dataset(self, dataset, observation):
         log.info(f"Computing dataset for observation {observation.obs_id}")
         for maker in self.makers:
             log.info(f"Running {maker.tag}")
-            dataset_obs = maker.run(dataset=dataset_obs, observation=observation)
+            try:
```
registerrier, line 105: Why this try clause?

registerrier (commented):
Thanks @QRemy ! it looks nice. 

gammapy/analysis/core.py
```diff
@@ -446,36 +447,18 @@ def _map_making(self):
...
-                log.debug(f"Processing observation {obs.obs_id}")
-                cutout = stacked.cutout(obs.pointing_radec, width=2 * offset_max)
-                dataset = maker.run(cutout, obs)
-                dataset = maker_safe_mask.run(dataset, obs)
-                if bkg_maker is not None:
-                    dataset = bkg_maker.run(dataset)
-                log.debug(dataset)
-                datasets.append(dataset)
-        self.datasets = Datasets(datasets)
+        datasets_maker = DatasetsMaker(makers,
```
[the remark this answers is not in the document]
QRemy: yes the RingBackground is supported thanks to the try/except below

gammapy/makers/reduce.py
QRemy: RingBackground and AdaptiveRingBackground don't have observation in their arguemnt contrary to any other makers

gammapy/analysis/config.py
QRemy: yes, I see other potential use cases:
- TSmapEstimator already have this argument,
- the ObservationsEventsSampler from #3711 will,
- run multiple flux points in parallel from a pool of sources and energy bins,
- allow multiple kernel size/shape for the ExcessMapEstimator/TSmapEstimator and run them in parallel
Moreover even if later we want to define a n_jobs per analysis step, I think that we still need to define a general limit at the beginning.

gammapy/analysis/tests/test_analysis.py
```diff
@@ -318,10 +318,10 @@ def test_analysis_ring_background():
     analysis.get_excess_map()
     assert isinstance(analysis.datasets[0], MapDataset)
     assert_allclose(
-        analysis.datasets[0].npred_background().data[0, 10, 10], 0.091799, rtol=1e-2
+        analysis.datasets[0].npred_background().data[0, 10, 10], 0.096767, rtol=1e-2
```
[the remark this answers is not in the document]
QRemy: because the default cutout mode in the DatasetsMaker is "partial" instead of "trim", not sure why it makes a difference. (I don't see an option to change the cutout_mode in the configuration for now)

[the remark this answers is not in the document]
adonath: Maybe keep the same cutout mode for now, to make sure the results are equivalent? Then in a follow up PR the mode could be changed if there is a better default...

gammapy/makers/reduce.py
adonath: Actually for consistency that could just take `.run(dataset, observation=None)`, maybe it is needed later for access to meta data or similar...

gammapy/analysis/tests/test_analysis.py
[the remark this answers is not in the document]
QRemy: ok changed the default to trim.
Then I noticed that the changes also come from the stacking of the MapDatasetOnOff into the MapDataset after the RingBackground because I forgot the conversion .to_map_dataset(). This should return an error no ?

gammapy/makers/reduce.py
QRemy: done

adonath (approved):
Thanks a lot @QRemy, I have no further comments...

adonath:
@registerrier Is the ready to be merged from your side?

registerrier (approved):
Thanks @QRemy !

Issue #2453, opened 2019-10-11
Crash after macOS Catalina update in multiprocess call


cdeil:
I updated to macOS Catalina and now I get this crash
```
gammapy/cube/tests/test_ring.py::TestAdaptiveRingBackgroundEstimator::test_run_const_width libc++abi.dylib: terminating with uncaught exception of type std::runtime_error: Couldn't close file
Fatal Python error: Aborted
```
```
libc++abi.dylib: terminating with uncaught exception of type std::runtime_error: Couldn't close file
Fatal Python error: Aborted
```
from here:
https://github.com/gammapy/gammapy/blob/c9ed2c68ed00ae37b14c8092d6096b0f2dbae95a/gammapy/maps/image_utils.py#L51

It occurs every time I run `pytest -s -v gammapy/cube/tests/`, but never when I run `pytest -s -v gammapy/cube/tests/test_ring.py`, i.e. it depends on Python process state and likely previous multiprocessing calls.

Log: https://gist.github.com/cdeil/a75211856a3bcab751ead707df9708c9#file-gistfile1-txt-L88

There is https://www.anaconda.com/how-to-restore-anaconda-after-macos-catalina-update/ but I couldn't find any issue yet concerning multiprocessing issues after Catalina update.

I'll `rm -r` my Anaconda install now and re-install from scratch, let's see ...
I just wanted to put this issue in the tracker to have a record, and as reference if others run into the same issue.

cdeil:
@adonath - I don't have a Python at the moment. :-)
```
hfm-1804a:gammapy deil$ pytest -s -v gammapy/detect/tests
-bash: /usr/local/bin/pytest: /usr/local/bin/python3.9: bad interpreter: No such file or directory
```

I tried to create a minimal test case for 5 minutes, running some multiprocessing code examples like this through python and pytest:
```
import multiprocessing


def f(x):
    return x ** 2

def run():
    pool = multiprocessing.Pool()
    pool.map(f, range(100))
    pool.close()
    pool.join()


def test_pool():
    for _ in range(30):
        run()
```
But that didn't trigger the crash - I wasn't able to create a minimal test case.

I remember this talk: https://youtu.be/UVL4LFy8ch0
Basically his assessment was that `multiprocessing` was broken in some cases on some platforms, and while some important fixes and improvements are in Python 3.7 and 3.8, at the moment, e.g. `scikit-learn` avoids using it, and probably we should as well.

[...]

PR #3806, opened 2022-02-16
Improve documentation for the DatasetsMaker

QRemy:
Improve documentation for the `DatasetsMaker`:
-extend docsting about the cutout options
-add an example in the makers notebook

### DIFF
--- a/docs/tutorials/api/makers.ipynb
+++ b/docs/tutorials/api/makers.ipynb
@@ -33,6 +33,7 @@
     "    SafeMaskMaker,\n",
     "    SpectrumDatasetMaker,\n",
     "    ReflectedRegionsBackgroundMaker,\n",
+    "    DatasetsMaker\n",
     ")\n",
     "from gammapy.datasets import MapDataset, SpectrumDataset, Datasets\n",
     "from gammapy.data import DataStore\n",
@@ -303,21 +304,21 @@
     "    skydir=(83.63, 22.01), axes=[energy_axis], width=5, binsz=0.02\n",
     ")\n",
     "\n",
-    "maker = MapDatasetMaker()\n",
+    "dataset_maker = MapDatasetMaker()\n",
     "safe_mask_maker = SafeMaskMaker(\n",
     "    methods=[\"aeff-default\", \"offset-max\"], offset_max=\"3 deg\"\n",
     ")\n",
     "\n",
-    "stacked = MapDataset.create(geom)\n",
+    "global_dataset = MapDataset.create(geom)\n",
     "\n",
     "for obs in observations:\n",
-    "    cutout = stacked.cutout(obs.pointing_radec, width=\"6 deg\")\n",
-    "    dataset = maker.run(cutout, obs)\n",
+    "    local_dataset = global_dataset.cutout(obs.pointing_radec, width=\"6 deg\")\n",
+    "    dataset = dataset_maker.run(local_dataset, obs)\n",
     "    dataset = safe_mask_maker.run(dataset, obs)\n",
     "    dataset = fov_bkg_maker.run(dataset)\n",
-    "    stacked.stack(dataset)\n",
+    "    global_dataset.stack(dataset)\n",
     "\n",
-    "print(stacked)"
+    "print(global_dataset)"
    ]
   },
   {
@@ -334,7 +335,22 @@
     "per observation into a larger dataset. During the stacking the safe data\n",
     "range mask (`MapDataset.mask_safe`) is applied by setting data outside to\n",
     "zero, then data is added to the larger map dataset. To stack multiple\n",
-    "observations, the larger dataset must be created first."
+    "observations, the larger dataset must be created first.\n",
+    "\n",
+    "The data reduction loop shown above can be done throught the `DatasetsMaker`.  Its `stack_datasets` option offers the possibily to stack or not the output datasets, and the `n_jobs` option allow to use multiple processes on run:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "global_dataset = MapDataset.create(geom)\n",
+    "makers = [dataset_maker, safe_mask_maker, fov_bkg_maker] #the order matter\n",
+    "datasets_maker = DatasetsMaker(makers, stack_datasets=False, n_jobs=1)\n",
+    "datasets = datasets_maker.run(global_dataset, observations)\n",
+    "print(datasets)"
    ]
   },
   {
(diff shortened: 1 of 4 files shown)


docs/tutorials/api/makers.ipynb
```diff
@@ -317,7 +318,22 @@
     "per observation into a larger dataset. During the stacking the safe data\n",
     "range mask (`MapDataset.mask_safe`) is applied by setting data outside to\n",
     "zero, then data is added to the larger map dataset. To stack multiple\n",
-    "observations, the larger dataset must be created first."
+    "observations, the larger dataset must be created first.\n",
+    "\n",
+    "The data reduction loop shown above can be done throught the `DatasetsMaker`.  Its `stack_datasets` option offers the possibily to stack or not the output datasets, and the `n_jobs` option allow to use multiple processes on run:"
```
registerrier, line 323: I would add in bold face that the order in which you pass the makers is important

docs/tutorials/api/makers.ipynb
```diff
@@ -287,21 +288,21 @@
...
     ")\n",
     "\n",
-    "maker = MapDatasetMaker()\n",
+    "dataset_maker = MapDatasetMaker()\n",
     "safe_mask_maker = SafeMaskMaker(\n",
     "    methods=[\"aeff-default\", \"offset-max\"], offset_max=\"3 deg\"\n",
     ")\n",
     "\n",
-    "stacked = MapDataset.create(geom)\n",
+    "global_dataset = MapDataset.create(geom)\n",
```
registerrier, line 296: why the change to `global_dataset`? `stacked` is quite explicit no?

gammapy/makers/reduce.py
```diff
@@ -22,14 +22,18 @@ class DatasetsMaker(Maker):
...
-        Stack into reference dataset or not
+        If True stack into the reference dataset (see `run` method arguments).
     n_jobs : int
-        Number of processes to run in parallel
-        Default is None
-    cutout_mode : str
-        Cutout mode. Default is "partial"
-    cutout_width : str or `~astropy.coordinates.Angle`,
-        Cutout width. Default is None, If Default is determined
+        Number of processes to run in parallel. Default is None.
```
registerrier, line 27: default should be 1 no?

registerrier (approved):

gammapy/makers/reduce.py
QRemy: yes, I also changed the TSMapEstimator for consistency

docs/tutorials/api/makers.ipynb
QRemy: added

docs/tutorials/api/makers.ipynb
QRemy: changed back to stacked

gammapy/makers/reduce.py
adonath: @registerrier No, the default of `None` was introduced on purpose and I think it should remain! The difference is that `n_jobs=None` implements a second path which avoids into calling into `multiprocessing` at all. The reason was that we have seen issues with `multiprocessing` in the past. See e.g. https://github.com/gammapy/gammapy/issues/2453. That's why the default should remain `None` (i.e. not calling into `multiprocessing` at all), optionally users can then use the parallelisation, but it is not guaranteed to work in all situations!

gammapy/estimators/map/ts.py
```diff
@@ -386,13 +386,12 @@ def estimate_flux_map(self, dataset):
...
-        if self.n_jobs is None:
-            results = list(map(wrap, positions))
-        else:
+        if self.n_jobs > 1 :
             with Pool(processes=self.n_jobs) as pool:
                 log.info("Using {} jobs to compute TS map.".format(self.n_jobs))
                 results = pool.map(wrap, positions)
-
+        else:
+            results = list(map(wrap, positions))
```
QRemy, line 394: @adonath this is equivalent to checking if n_jobs is None

gammapy/makers/reduce.py
QRemy: running multiprocessing only if n_jobs > 1 is equivalent, cf the change below

registerrier (approved):

Regarding `n_jobs` argument, it is probably more clear that `n_job=1` means one process. Otherwise, one needs to explain that `n_job=1` is one process in multiprocessing mode. This might be useful to debug I suppose, but probably confusing to users. Is the solution OK with you @adonath ?

registerrier:
@adonath, do you think the `n_jobs` argument should be `None` for no multiprocessing, or is the current solution with default at 1 and multiprocessing called only for `n_jobs >=2` fine with you?

gammapy/makers/reduce.py
```diff
@@ -22,14 +22,18 @@ class DatasetsMaker(Maker):
...
-        Cutout width. Default is None, If Default is determined
+        Number of processes to run in parallel. Default is 1.
+    cutout_mode : {'trim', 'partial', 'strict'}
+        Used only to cutout the refrence MapDataset around each processed observation.
+        Mode is an option for Cutout2D, for details see `~astropy.nddata.utils.Cutout2D`.
+        Default is "trim"..
+    cutout_width : tuple of `~astropy.coordinates.Angle`
+        Angular sizes of the region in (lon, lat) in that specific order.
+        If only one value is passed, a square region is extracted.
+        If None it is set to 2*offset-max if this option is given to the SafeMaskmaker
```
registerrier, line 35: 2*offset_max might be clearer

gammapy/makers/reduce.py
```diff
@@ -22,14 +22,18 @@ class DatasetsMaker(Maker):
...
-        Default is None
-    cutout_mode : str
-        Cutout mode. Default is "trim"
-    cutout_width : str or `~astropy.coordinates.Angle`,
-        Cutout width. Default is None, If Default is determined
+        Number of processes to run in parallel. Default is 1.
+    cutout_mode : {'trim', 'partial', 'strict'}
+        Used only to cutout the refrence MapDataset around each processed observation.
+        Mode is an option for Cutout2D, for details see `~astropy.nddata.utils.Cutout2D`.
+        Default is "trim"..
```
registerrier, line 31: Remove last dot.

registerrier (approved):
Thanks @QRemy ! I have left two nitpick comments.

The CI failure is unrelated (due to a filename change in gammapy-data)

adonath:
@QRemy and @registerrier I just looked into the `n_jobs` argument again and I now remember that I followed `scikit-learn` when first introducing the argument. `scikit-learn` specifies `n_jobs=None` by default, the reason is given [here](https://scikit-learn.org/stable/glossary.html#term-n_jobs). So `scikit-learn` has backend independent parallelization support and in some cases the number of jobs are configured "outside", like:
```python3
from ray.util.joblib import register_ray  
register_ray()

with parallel_backend("ray", n_jobs=10):  
    print(Parallel()(delayed(neg)(i + 1) for i in range(5)))

```
In which case the `n_jobs=None` corresponds to number of jobs defined in the context, otherwise it defaults to one job. We don't have anything like this in Gammapy (yet), but I think the need for an additional placeholder value is probably still justified for future developments.

registerrier (approved):
Thanks @QRemy ! Fine with the None default.

registerrier:

PR #3807, opened 2022-02-16
Allow MapAxis with single bin and node type center

adonath:
This pull request addresses #1952 by changing to nearest neighbor interpolation for the case of a single axis bin. It adds a unit test for `MapAxis` and unit test for `FluxPoints` serialization and plotting, where the issue came up first.

### DIFF
--- a/gammapy/estimators/map/core.py
+++ b/gammapy/estimators/map/core.py
@@ -156,6 +156,7 @@ class FluxMaps:
         if meta is None:
             meta = {}
 
+        meta.setdefault("sed_type_init", "likelihood")
         self.meta = meta
         self.gti = gti
         self._filter_success_nan = filter_success_nan
@@ -233,6 +234,14 @@ class FluxMaps:
                 return sed_type
 
     @property
+    def is_convertible_to_flux_sed_type(self):
+        """Check whether differential sed type is convertable to integral sed type"""
+        if self.sed_type_init in ["dnde", "e2dnde"]:
+            return self.energy_axis.node_type == "edges"
+
+        return True
+
+    @property
     def has_ul(self):
         """Whether the flux estimate has either sqrt(ts) or ts defined"""
         return "norm_ul" in self._data
@@ -287,6 +296,14 @@ class FluxMaps:
     def sed_type_init(self):
         """Initial sed type"""
         return self.meta.get("sed_type_init")
+
+    @property
+    def sed_type_plot_default(self):
+        """Initial sed type"""
+        if self.sed_type_init == "likelihood":
+            return "dnde"
+
+        return self.sed_type_init
 
     @property
     def geom(self):
@@ -553,6 +570,9 @@ class FluxMaps:
     @property
     def flux_ref(self):
         """Reference integral flux"""
+        if not self.is_convertible_to_flux_sed_type:
+            raise ValueError("Missing energy range definition, cannot convert to flux sed type.")
+
         energy_min = self.energy_axis.edges[:-1]
         energy_max = self.energy_axis.edges[1:]
         result = self.reference_spectral_model.integral(energy_min, energy_max)
@@ -561,6 +581,9 @@ class FluxMaps:
     @property
     def eflux_ref(self):
         """Reference energy flux"""
+        if not self.is_convertible_to_flux_sed_type:
+            raise ValueError("Missing energy range definition, cannot convert to flux sed type.")
+
         energy_min = self.energy_axis.edges[:-1]
         energy_max = self.energy_axis.edges[1:]
         result = self.reference_spectral_model.energy_flux(energy_min, energy_max)
@@ -720,7 +743,7 @@ class FluxMaps:
             gti=self.gti,
         )
 
-    def to_maps(self, sed_type="likelihood"):
+    def to_maps(self, sed_type=None):
         """Return maps in a given SED type.
 
         Parameters
@@ -734,6 +757,9 @@ class FluxMaps:
             Maps object containing the requested maps.
         """
         maps = Maps()
+
+        if sed_type is None:
+            sed_type = self.sed_type_init
 
         for quantity in self.all_quantities(sed_type=sed_type):
             m = getattr(self, quantity, None)
@@ -852,7 +878,7 @@ class FluxMaps:
 
         return cls(data=data, reference_model=reference_model, gti=gti, meta=meta)
 
-    def to_hdulist(self, sed_type="likelihood", hdu_bands=None):
+    def to_hdulist(self, sed_type=None, hdu_bands=None):
         """Convert flux map to list of HDUs.
 
         For now, one cannot export the reference model.
@@ -870,6 +896,9 @@ class FluxMaps:
         hdulist : `~astropy.io.fits.HDUList`
             Map dataset list of HDUs.
         """
+        if sed_type is None:
+            sed_type = self.sed_type_init
+
         exclude_primary = slice(1, None)
 
         hdu_primary = fits.PrimaryHDU()
@@ -926,7 +955,7 @@ class FluxMaps:
         )
 
     def write(
-        self, filename, filename_model=None, overwrite=False, sed_type="likelihood"
+        self, filename, filename_model=None, overwrite=False, sed_type=None
     ):
         """Write flux map to file.
 
@@ -942,6 +971,9 @@ class FluxMaps:
         sed_type : str
             sed type to convert to. Default is `likelihood`
         """
+        if sed_type is None:
+            sed_type = self.sed_type_init
+
         filename = make_path(filename)
 
         if filename_model is None:
--- a/gammapy/estimators/map/core.py
+++ b/gammapy/estimators/map/core.py
@@ -571,7 +571,7 @@ class FluxMaps:
     def flux_ref(self):
         """Reference integral flux"""
         if not self.is_convertible_to_flux_sed_type:
-            raise ValueError("Missing energy range definition, cannot convert to flux sed type.")
+            raise ValueError("Missing energy range definition, cannot convert to sed type 'flux'.")
 
         energy_min = self.energy_axis.edges[:-1]
         energy_max = self.energy_axis.edges[1:]
@@ -582,7 +582,7 @@ class FluxMaps:
     def eflux_ref(self):
         """Reference energy flux"""
         if not self.is_convertible_to_flux_sed_type:
-            raise ValueError("Missing energy range definition, cannot convert to flux sed type.")
+            raise ValueError("Missing energy range definition, cannot convert to sed type 'eflux'.")
 
         energy_min = self.energy_axis.edges[:-1]
         energy_max = self.energy_axis.edges[1:]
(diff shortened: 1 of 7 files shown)


gammapy/estimators/points/core.py
```diff
@@ -322,6 +322,14 @@ def to_table(self, sed_type="likelihood", format="gadf-sed", formatted=False):
 
             idx = (Ellipsis, 0, 0)
             table = self.energy_axis.to_table(format="gadf-sed")
+
+            if self.energy_axis.nbin == 1 and self.energy_axis.node_type == "center":
+                if sed_type in ["flux", "eflux", "likelihood"]:
```
AtreyeeS, line 327: So this clause is because Gammapy has no way of guessing the energy edges given a single centre?

But I suppose we are still guessing the edges given multiple centres? and, say, for flux points, it can be totally incorrect for the last bin?

AtreyeeS (commented):
Thanks @adonath !

So now the fit can proceed with a `FluxPoint` with a single bin, node_type=center, but it can fail for plotting as `flux`?

Ideally we would expect users to have `node_type=edges` in that case?

gammapy/estimators/points/core.py
adonath: Thanks, I now realize this is not a complete solution and probably requires a some more thought. The problem is basically that when computing flux points in Gammapy we know `e_min`, `e_max` and `e_ref`. For completeness we should add this information to the table for the `dnde` case as well. However in the case where we read existing flux points of type `dnde` we should indeed not invent the information. I need to think bit more how to better handle this maybe we need an additional keyword on the `MapAxis` object...

adonath:
@AtreyeeS Yes, for a single bin there is no way we can guess the energy range from the bin center. And you are right we need and additional check for plotting. 

Since I started to work on this PR I'm not even fully convinced the support for single energy bins with node type `center` is a good idea. We probably need a few more checks and special cases. But we can still use the PR to discuss further...

adonath:
One important aspect that became clear to me is: strictly speaking there are no "differential flux points", this quantity does not exist. All SED measurements are based on integrating flux in a given energy range, even if it is very small. When people published flux points as "differential", they just dropped this information. This is non-ideal, but for whatever reason very common. 

Within Gammapy we have to decide what to do...

adonath:
@registerrier Not "inventing" bin edges means we cannot convert from the `dnde` SED type to any integral sed type, which is not ideal either...I'm not sure what to do here.

AtreyeeS:
Probably the GADF flux points columns can be updated to necessarily have `e_min` `e_max` add `e_ref` ? I suppose it is better if the user invents the information somehow rather than Gammapy doing it silently?
If the edges are not present, Gammapy emits a warning like `bin edge information not found. assuming log spaced around the bin centres` and creates them.
@bkhelifi : The assumption of contiguous edges should be ok, no? It is built into our normal `MapAxis`..

Coming back to the single bin case, can we foresee any usage except for flux points? 
I think we do not enforce the binning of IRF axes with appropriate nodes at present ( or `energy_true` as `center` and `energy` as `edges` ). If we do this in future, we need to think a bit.

Regarding usage within `FluxPoints`, this point can be used only for fitting then? It should still plot, but without an `x_err`. Is that sufficient @cosimoNigro ?

bkhelifi:
@atreyee: I don't think that the assumption of contiguous edges is a special case of what could happen.
Ex: 1 differential flux from one experiment (radio, X-ray, HAWC). The notion of edges does not exist if they publish as it is
Ex2: one can have light curves with holes. The edges are then not contiguous....
In fact, we make assumptions of what we (developers) imagine about the data, because we have a software issue...

Conceptually, a MapAxis defined with edges have a centre, whereas a MapAxis defined with centres has no edge. Tricky

registerrier:
Thanks @adonath !
To follow on the discussion, 
- I think the initial notion of integral/differential quantities (or extensive/intensive since we're physicists) on `Maps` through the `MapAxis.node_type` is not really relevant because we never really use it this way. The way IRFs are formatted clearly shows this. We also don't define `exposure` at bin centers for instance. Some issues:
    * A possible issue for instance is `Map.normalize`: there we simply perform a sum to normalize along an axis, i.e. we assume the quantity to be integral. Otherwise we should multiply by the bin size.
    * If the quantity is integral, it should not be possible to interpolate, e.g. you should not interpolate the integrated flux.
    * For methods like `downsample`, we have introduced a `preserve_counts` argument to ensure extensive/intensive quantities are treated correctly. 
    * For spatial axes, we don't really have this notion. We could rely on the location of the reference position (at the center or edge of a pixel) but this would be very tricky to use. Yet some quantities are extensive in space (e.g. background, counts) while others are intensive (e.g. exposure). Note that a flux map produced by an estimator is not only integrated but also oversampled...
- `MapAxis` is therefore by construction an axis of *continuous* intervals that can be created from interval centers or edges. The notion of extensive/intensive quantity is independent.
This globally works for our usual handling, but relying on `Maps` for flux points has made things more complex since we indeed need axes defined on nodes, without any notion of centers/edges. 
1. That is the only way to handle spectra defined at reference energies.
2. Indeed, fluxes are always defined on some range but in most of astronomy (radio, IR, optical) these ranges are never contiguous and very small. So we are back to point 1.
Basically, we need something that defines a graph rather than a histogram.

adonath:
Thanks @registerrier! What I realize with your comment is that there are indeed much more places where we have to distinguish between the two cases. What I also learned is that the distinction by the `node_type` is probably not a good idea, because it possibly looses the information, once once goes from integral to differential quantities and the other way around. Given that we have many more use case such as `.normalize()` I wonder whether we should maybe just introduce an independent meta keyword to `MapAxis`, e.g. like `MapAxis(is_differential=True)`?

adonath:
Conclusion from dev meeting: for this PR we just need to handle the case, where we read `dnde` flux types. If there is no `e_min` and `e_max` information we should not silently invent the flux points, but raise and error on flux conversion...

adonath:
I updated the PR to include the property `FluxPoints.is_convertible_to_flux_sed_type` and slightly change the default sed_type handling:
- If the `sed_type_init` (the sed type the flux points were created from) is differential and there are energy ranges defined, the conversion to any integral sed type fails. The same goes for `.write()` and `.to_table()`...
- By default the `sed_type_init` is conserved, i.e. it will be used as the default for plotting, writing and `.to_table()`. The previous behavior was to always use `dnde`

gammapy/estimators/map/core.py
```diff
@@ -870,6 +896,9 @@ def to_hdulist(self, sed_type="likelihood", hdu_bands=None):
         hdulist : `~astropy.io.fits.HDUList`
             Map dataset list of HDUs.
         """
+        if sed_type is None:
+            sed_type = self.sed_type_init
```
registerrier, line 900: maybe test the default behavior?

gammapy/maps/axes.py
```diff
@@ -26,22 +26,35 @@ def flat_if_equal(array):
...
+        "fill_value": "extrapolate",
+        "x": np.arange(len(edges), dtype=float),
+        "y": scale(edges),
+    }
 
+    # change to nearest neighbour interpolation for single bins
+    if len(edges) == 1:
+        kwargs["kind"] = 0
+
+    interp_fn = scipy.interpolate.interp1d(**kwargs)
```
registerrier, line 57: I suppose that an additional function could reduce code duplication here.

registerrier, line 57: I suppose that an additional function could reduce code duplication here.

gammapy/maps/axes.py
```diff
@@ -291,7 +301,9 @@ def format_plot_xaxis(self, ax):
                                         quantity=PLOT_AXIS_LABEL.get(self.name, self.name.capitalize())
                                         , unit=self.unit )
         ax.set_xlabel(xlabel)
-        ax.set_xlim(self.bounds)
+        xmin, xmax = self.bounds
+        if not xmin == xmax:
+            ax.set_xlim(self.bounds)
```
registerrier, line 306: Note that with center defined axis bounds will stop exactly at the first and last bin centers. Maybe one should add a margin?

registerrier: But that can be another PR

gammapy/maps/region/ndmap.py
```diff
@@ -74,7 +74,9 @@ def plot(self, ax=None, axis_name=None, **kwargs):
         ax = ax or plt.gca()
 
         if axis_name is None:
-            if self.geom.axes.is_unidimensional:
+            if len(self.geom.axes) == 1:
```
registerrier, line 77: Why is this needed now?

gammapy/estimators/map/core.py
registerrier, line 900: maybe test the default behavior?

registerrier (approved):
Thanks @adonath . I think this definitely improves the current issues. I have left some minor comments inline.

gammapy/estimators/map/core.py
```diff
@@ -232,6 +233,14 @@ def _guess_sed_type(quantities):
             if required.issubset(quantities):
                 return sed_type
 
+    @property
+    def is_convertible_to_flux_sed_type(self):
+        """Check whether differential sed type is convertable to integral sed type"""
+        if self.sed_type_init in ["dnde", "e2dnde"]:
```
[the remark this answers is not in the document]
adonath: Yes, I think for this matter it is equivalent. I usually prefer the more explicit implementation...which I think is the case here.

gammapy/estimators/map/core.py
```diff
@@ -156,6 +156,7 @@ def __init__(
         if meta is None:
             meta = {}
 
+        meta.setdefault("sed_type_init", "likelihood")
```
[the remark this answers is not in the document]
adonath: I think this need to be and independent discussion. It is the question for `Map` and `IRF`. It think the best solution would be to introduce dedicated `MapMeta`, `FluxMapsMeta` classes. This bundles the meta data into an own namespace (advantage of `dict`), while keeping the code based documentation and definition (advantage of the attributes...)

gammapy/maps/axes.py
adonath: I will try this...

registerrier:
What is the status here @adonath ?

adonath:
@registerrier I still have to rebase and address your comments. Will start now...

gammapy/maps/axes.py
adonath: I added a little `AxisCoordInterpolator`, which allows to share the configuration...

gammapy/maps/axes.py
adonath: Would prefer a separate OR for this...

gammapy/estimators/map/core.py
adonath: I changed the to `sed_type=None` in the signature, so this behavior is actually tested in many places.

adonath:
@registerrier I addressed you remaining comments. Will go ahead and merge now...

Issue #3457, opened 2021-08-06
Keyword forwarding and documentation


adonath:
In Gammapy we have many places where we forward keywords to other methods like:
```python3

@classmethod
def from_something(cls, something, **kwargs):
    # some code
    return cls(..., **kwargs)

```
There are many examples for this, such as `RegionGeom.create()`, `MapDataset.create()` etc. In some cases we partly repeat keywords and document them again in others we don't. Here are some examples:
- [RegionGeom.from_geoms](https://github.com/gammapy/gammapy/blob/master/gammapy/maps/region.py#L599)
- [WcsGeom.create](https://github.com/gammapy/gammapy/blob/master/gammapy/maps/wcs.py#L342)
- [SkyModel.copy](https://github.com/gammapy/gammapy/blob/master/gammapy/modeling/models/cube.py#L349)
- ...

In general the `kwargs` pattern is very common in Python and we need it as well e.g. to forward keyword arguments to `matplotlib`.  A common pattern to document it is as following:

```python3

@classmethod
def from_something(cls, something, **kwargs):
    """Create class from something
    
    Parameters
    ------------
    something : `Something`
       Instance of something
    **kwargs : dict
        Keyword arguments forwarded to `Class.__init__`

    Returns
    --------
    instance : `Class`
        Instance of the class
    """
    # some code
    return cls(..., **kwargs)

```
See e.g. https://numpydoc.readthedocs.io/en/latest/example.html#source or https://pandas.pydata.org/pandas-docs/stable/development/contributing_docstring.html#section-3-parameters

The downside of this pattern is, that it's slightly inconvenient for users because the documentation of the possible valid keywords is in a different place. Especially if keywords are forward to multiple methods, down the hierarchy. In case where we forward to 3rd party packages this is and will never be under our control. However in Gammapy we can handle it uniformly and decide on what we would like to have. The options are:
- Not use `**kwargs` at all and always repeat arguments. I think this is not an option at all. This will increase the maintenance effort a lot and lead to many inconsistencies, where new keywords are forgotten etc. So it will also decrease the users experience long term.
- Rely on `**kwargs` and repeat the documentation of all valid keywords. This could be an option, but again there is a risk that inconsistencies will be introduced with time.
- Repeat some important keywords and documentation. Could be an option as well, but there it's hard to draw a line which keywords to repeat.
[...]

PR #3808, opened 2022-02-17
Add available formats to docstring of FluxPoints.write()

LauraOlivera:
This pull request makes very small changes to the documentation. Just a couple of things I noticed recently and wanted to fix before forgetting.

* a typo in the .rst page of the estimators sub-package
* the docstring of the `FluxPoints.write()` method was missing two of the available formats. I also copied the definition of each format from the  `FluxPoints.to_table()` method, because from a user perspective it is better to have that info already in the write method documentation without needing to figure out what method is that one calling.

### DIFF
--- a/docs/estimators/index.rst
+++ b/docs/estimators/index.rst
@@ -86,7 +86,7 @@ e_max             Maximum energy
 dnde              Differential flux at ``e_ref``
 flux              Integrated flux between ``e_min`` and ``e_max``
 eflux             Integrated energy flux between ``e_min`` and ``e_max``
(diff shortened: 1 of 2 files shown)


adonath:
Thanks a lot @LauraOlivera! I remember that we discussed and agreed on a solution for documenting keyword forwarding, which is described here: https://github.com/gammapy/gammapy/issues/3457#issuecomment-902575170 However I just realized this does not apply here because the forwarding actually happens to `Table.write()`, so +1 to repeat the documentation if keywords are repeated as well.

adonath (approved):

PR #3809, opened 2022-02-17
Minor change in detect tutorial

AtreyeeS:
Removing left over line from older notebook when we would actually perform a `sum_over_axes()`


adonath (approved):
Thanks @AtreyeeS! For these kind of changes you could put `[ci skip]` in the commit message to prevent the whole CI from running...

PR #3813, opened 2022-02-17
Fix build docs

Bultako:
This PR fixes the build docs CI after #3778 
It creates target folder in `gammapy jupyter tar` when it does not exist.

https://github.com/gammapy/gammapy-docs/runs/5233129211?check_suite_focus=true#step:7:38


adonath (approved):
Thanks @Bultako!

Issue #3747, opened 2022-01-26
FluxPointDataset serialisation


AtreyeeS:
I am a bit confused about reading `FluxPointsDataset`. 

- We have a `FluxPointsDataset.write()`, but no `read`. Should the reading always happen via `FluxPoints` ?
- The spectral model seems to get changed, for ex:

```
from gammapy.datasets import FluxPointsDataset
from gammapy.modeling.models import PowerLawSpectralModel, SkyModel
filename = "$GAMMAPY_DATA/tests/spectrum/flux_points/diff_flux_points.fits"
from gammapy.estimators import FluxPoints
flux_points = FluxPoints.read(filename)
print(flux_points)
```

The output is (as expected)
```
FluxPoints
----------

  geom                   : RegionGeom
  axes                   : ['lon', 'lat', 'energy']
  shape                  : (1, 1, 24)
  quantities             : ['norm', 'norm_err', 'norm_ul']
  ref. model             : pl
  n_sigma                : 1
  n_sigma_ul             : 2.0
  sqrt_ts_threshold_ul   : 2
  sed type init          : dnde
```

Now, I create a `FluxPointDataset`, write it, read it back

```
model = SkyModel(spectral_model=PowerLawSpectralModel())
dataset = FluxPointsDataset(data=flux_points, models=model)
dataset.write("test.fits")
f1 = FluxPoints.read("test.fits")
print(f1)
```

The output is:
```
FluxPoints
----------

  geom                   : RegionGeom
  axes                   : ['lon', 'lat', 'energy']
  shape                  : (1, 1, 24)
  quantities             : ['norm', 'norm_err', 'norm_ul', 'is_ul']
  ref. model             : template
  n_sigma                : 1
  n_sigma_ul             : 2.0
  sqrt_ts_threshold_ul   : 2
  sed type init          : likelihood
```

namely, the ref model is now a `template`

- In the above example, if `models` is not passed on `init` to `FluxPointDataset`, `print(dataset)` fails with 
```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
/var/folders/v4/62d6cvbj72qbyw23mq2pwxl00000gn/T/ipykernel_16809/2347221472.py in <module>
----> 1 print(dataset)

~/Gammapy-dev/gammapy/gammapy/datasets/flux_points.py in __str__(self)
    209 
    210         str_ += "\t{:32}: {}\n".format(
--> 211             "Number of parameters", len(self.models.parameters)
[...]

PR #3814, opened 2022-02-17
Add `FluxPointsDataset.read`

AtreyeeS:
Directly allows to create a `FluxPointsDataset` as proposed in #3747 
This makes the API homogeneous with `MapDataset`

Also, changing `FluxPointsDataset.write` default `overwrite` to `False` to be consistent with `MapDataset.write`

### DIFF
--- a/gammapy/datasets/flux_points.py
+++ b/gammapy/datasets/flux_points.py
@@ -163,8 +163,17 @@ class FluxPointsDataset(Dataset):
 
         filename = make_path(filename)
         table = Table.read(filename)
+        mask_fit = None
+        mask_safe = None
+        if "mask_safe" in table.colnames:
+            mask_safe = table["mask_safe"].data.astype("bool")
+        if "mask_fit" in table.colnames:
+            mask_fit = table["mask_fit"].data.astype("bool")
         return cls(
-            name=make_name(name), data=FluxPoints.from_table(table, format=format)
+            name=make_name(name),
+            data=FluxPoints.from_table(table, format=format),
+            mask_fit=mask_fit,
+            mask_safe=mask_safe
         )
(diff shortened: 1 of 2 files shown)


gammapy/datasets/flux_points.py
```diff
@@ -146,6 +141,32 @@ def write(self, filename, overwrite=True, **kwargs):
...
+
+        Returns
+        -------
+        dataset : `FluxPointsDataset`
+            FluxPointsDataset
+        """
+        from gammapy.estimators import FluxPoints
+
+        filename = make_path(filename)
+        table = Table.read(filename)
```
adonath, line 165: Is there any specific reason to go though `Table.read()` and not `FluxPoints.read()`?

AtreyeeS: `.write` is writing a `Table`, so I did similarly here. Would `FluxPoints.read()` work in all cases?

registerrier (commented):
Thanks @AtreyeeS 

I don't know what should be correct behavior here. 
When a `FluxPointsDataset` is read from disk, do we want to re-serialize it? If yes, should we let the content change? Because this is what should happen because of the `FluxPoints` creation.  

I think we are hitting the same issue as discussed in #3807.

AtreyeeS:
It should behave similarly for `Datasets.read()` and `FluxPointsDataset.read()` (without the models serialisation), no?

AtreyeeS:
Adapted to read the masks as discussed in the dev call

adonath (approved):
Thanks @AtreyeeS, no further comments from my side...

registerrier (approved):
Thanks @AtreyeeS !

PR #3815, opened 2022-02-17
Fix SafeMaskMaker to take mask safe properly into account

### LINKED ISSUE #3787: ReflectedRegions analysis fails when no off regions are found
**Gammapy version**

0.19, dev

**Bug description**

Reflected Regions analysis fails when no off regions are found. While the `ReflectedRegionsBackgroundMaker` correctly raises a warning, the error happens in `SafeMaskMaker`

```
ReflectedRegionsBackgroundMaker failed. No OFF region found outside exclusion mask for jTTbJpk3.
ReflectedRegionsBackgroundMaker failed. Setting jTTbJpk3 mask to False.
TypeError                                 Traceback (most recent call last)
<timed exec> in <module>

~/opt/anaconda3/envs/gammapy-0.19/lib/python3.9/site-packages/gammapy/makers/safe.py in run(self, dataset, observation)
    283         mask_safe = np.ones(dataset._geom.data_shape, dtype=bool)
    284 
--> 285         if dataset.background is not None:
    286             # apply it first so only clipped values are removed for "bkg-peak"
    287             mask_safe &= self.make_mask_bkg_invalid(dataset)

~/opt/anaconda3/envs/gammapy-0.19/lib/python3.9/site-packages/gammapy/datasets/map.py in background(self)
   1971             Background map
   1972         """
-> 1973         return self.alpha * self.counts_off
   1974 
   1975     def stat_array(self):

~/opt/anaconda3/envs/gammapy-0.19/lib/python3.9/site-packages/gammapy/maps/core.py in __mul__(self, other)
   1579 
   1580     def __mul__(self, other):
-> 1581         return self._arithmetics(np.multiply, other, copy=True)
   1582 
   1583     def __imul__(self, other):

~/opt/anaconda3/envs/gammapy-0.19/lib/python3.9/site-packages/gammapy/maps/core.py in _arithmetics(self, operator, other, copy)
   1543                 raise ValueError("Map Arithmetic: Inconsistent geometries.")
   1544         else:
-> 1545             q = u.Quantity(other, copy=False)
   1546 
   1547         out = self.copy() if copy else self

~/opt/anaconda3/envs/gammapy-0.19/lib/python3.9/site-packages/astropy/units/quantity.py in __new__(cls, value, unit, dtype, copy, order, subok, ndmin)
    382             not (value.dtype.kind == 'O' and
    383                  isinstance(value.item(0), numbers.Number))):
--> 384             raise TypeError("The value must be a valid Python or "
    385                             "Numpy numeric type.")
    386 

TypeError: The value must be a valid Python or Numpy numeric type.
```

AtreyeeS:
This fixes #3787


bkhelifi:
It looks OK for me, Thanks Atreyee

gammapy/makers/safe.py
```diff
@@ -281,7 +281,13 @@ def run(self, dataset, observation=None):
         dataset : `Dataset`
             Dataset with defined safe range mask.
         """
-        mask_safe = np.ones(dataset._geom.data_shape, dtype=bool)
+
+        if dataset.mask_safe:
+            mask_safe = dataset.mask_safe.data
+            if np.all(mask_safe == False):
```
adonath, line 287: Maybe even remove this "fast-track" path? Of course the `SafeMaskMaker` does not modify the mask if it is all `False` from the beginning, however it maybe still adds some meta data in future. Performance wise it should be completely negligible...

AtreyeeS: `MapDatasetOnOff.background` errors out if `counts_off` is None (because map arithmetic fails). To avoid this, I adapted map artithmetic such that `map X None = None`, where `X`   is any `operator`.
This is reasonable for multiplication or division, but `a + None = None` is strange (currently it just fails).

The other solution would be to manually set `background`, `npred_background` to `None` in `MapDatasetOnOff` if `counts_off` is None.

What would you suggest?

AtreyeeS: Thinking about this, the behaviour I introduced in maps is not nice. I'm reverting it and rather adapting `MapDatasetOnOff.background`

adonath: Yes, that's the preferable solution. If  `counts_off` is `None` than `npred_background` should return `None` as well. Arithmetics involving `None` should always fail!

gammapy/maps/core.py
```diff
@@ -1536,6 +1536,8 @@ def __repr__(self):
 
     def _arithmetics(self, operator, other, copy):
         """Perform arithmetic on maps after checking geometry consistency."""
+        #if other is None:
```
adonath, line 1539: Just remove the comment here...

gammapy/datasets/map.py
```diff
@@ -2075,6 +2075,8 @@ def background(self):
         background : `Map`
             Background map
         """
+        if self.counts_off is None:
```
adonath, line 2078: Yes, I think this is the correct behavior.

adonath (dismissed):
Thanks @AtreyeeS! Please rebase then it's ready to be merged.

adonath (approved):
Thanks @AtreyeeS, I have no further comments.

Issue #3784, opened 2022-02-04
Remove unit setters on Map and IRF


registerrier:
Both `Map` and `IRF` have unit setter which let the user change the unit without converting the data accordingly. 

For this the `Map.to_unit()` is a good approach. It could be added to the `IRF` base class as well.

PR #3816, opened 2022-02-17
Remove unit setter on IRF and Map 

registerrier:
**Description**

This pull request solves #3784 and removes the unit setter on both `IRF` and  `Map`. 
A `to_unit` method is added on `IRF` to follow the implementation of `Map`. This is not implemented for parametric PSFs which have several parameters and several units. 

**Dear reviewer**


registerrier:
For some reason, the diff on the notebook is even worse than before...

Issue #3818, opened 2022-02-21
Moving documentation content from API Reference


AtreyeeS:
We discussed in the last call that we should have a proposal for moving content from API reference doc to somewhere more discoverable by the users. Here I try to make a few suggestions.

The said content should probably reside in the **User Guide** section. 

1. The bullet points and the buttons should not co-exist. 
- Probably remove the *Gamma-ray data analysis* link. Add some comprehensive references (or link to http://tevcat.uchicago.edu/reviews.html) in the *Glossary and References* section. 
- *Package Structure* should be a big button, where the bulk of the content can be moved.
- I do not know if `Dependencies` should stay here. It can maybe move to the page after `Getting Started` and `Installation`

2. `Package Structure`: This page can be broken into sub-pages, one for each sub-package.
- The first figure remains as is.
- The sections become clickable links with one line descriptions.
- Every sub-package has one page. The paragraph on the present page + the API reference page text moves here.

Please opine @registerrier @adonath @Bultako @bkhelifi @luca-giunti @LauraOlivera @QRemy @cosimoNigro

Bultako:
Thanks @AtreyeeS for opening this issue.
I have also filed PR #3819 to address the first part of the suggestions concerning the User Guide.
Let's keep this one only to discuss details about moving the API Reference.

In general I like the idea of moving the content of the API index pages to individual subpages of the Package structure. Let's see what others can say, if nobody is against I could open a new PR with a proposal.

PR #3819, opened 2022-02-22
Refactor content in User Guide and Getting Started docs

Bultako:
This PR continues with the improvement of the docs, addressing the suggestions issued in the first group of items of #3818 

- cosmetic modifs on the top navigation bar
- adding boxes and refactoring content in User Guide and Getting Started sections

You may access how the docs shows with this PR in the [usual link](http://www.iaa.csic.es/~jer/gammapydocs).


Bultako:
Bringing some comments from #3818 

> Probably remove the Gamma-ray data analysis link. 

I think it could be useful to have a page describing briefly the standard Gamma-ray data analysis process.

> Package Structure should be a big button, where the bulk of the content can be moved.

Agree.

> I do not know if Dependencies should stay here. It can maybe move to the page after Getting Started and Installation

Dependencies is now under the [Development](https://docs.gammapy.org/dev/development/index.html) section.

AtreyeeS (dismissed):
Thank you @Bultako ! I really like the `Getting Started` section now.

The `User Guide` also looks very nice, specially the icons that you have chosen :)

registerrier (approved):
Thanks @Bultako . This looks very nice.

On my browser, the menu bar disappears and is replaced by a single icon if the screen size is too small. Is it expected behavior?

Bultako:
@registerrier 

> On my browser, the menu bar disappears and is replaced by a single icon if the screen size is too small. Is it expected behavior?

Yes, this is called responsive behaviour, which means that for small screen devices like phones or tablets the menu bar is accessible by clicking on the menu icon. This is a feature of the pydata sphinx theme, and you can also see it in the dev docs already published. In this PR I just changed the color of that menu icon to give it more visibility.

registerrier:
Merging now. Sorry for the delay.

PR #3822, opened 2022-02-23
Add headers from obs_table to obs_info for individual observation

maxnoe:
**Description**

This pull request is a breakout from #3797 that should be easy to review and merge. It ensures that we have the information from the header of the `OBS_INDEX` table available in the `obs_info` of the individual observations.

This is required since otherwise the `TSTART` / `TSTOP` columns cannot be interpreted due to missing reference headers.


adonath (commented):
Thanks @maxnoe, no comments from my side...

adonath (approved):

PR #3826, opened 2022-02-23
Rework time handling

### LINKED ISSUE #284: Use utc or tt time scale?
In Gammapy we currently use the `utc` time scale. I think in CTA it is under discussion if the `utc` or `tt` time scale should be used (e.g. for event times, but really all times).

See http://astropy.readthedocs.org/en/latest/time/index.html#time-scale and references there.

@kosack Do you know what the status of the UTC versus TT time scale discussion is in CTA? Even if it's not written up in a spec yet, what should we use in Gammapy for now? Or will the time scale always be put in the FITS header and CTA software (science tools specifically) is supposed to handle both UTC and TT?

maxnoe:
**Description**

Improving the handling of times, to fix #284

to be continued.

### DIFF
--- a/gammapy/data/event_list.py
+++ b/gammapy/data/event_list.py
@@ -13,7 +13,7 @@ from gammapy.maps import MapAxis, MapCoord, RegionGeom, WcsNDMap
 from gammapy.utils.fits import earth_location_from_dict
 from gammapy.utils.scripts import make_path
 from gammapy.utils.testing import Checker
-from gammapy.utils.time import time_ref_from_dict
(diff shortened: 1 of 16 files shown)


maxnoe:
Still missing is converting the time-like columns of things to be actual `Time` columns and creating / reversing the conversion to reference + delta on reading / writing

bkhelifi:
@maxnoe the branch of this PR needs some update. What is the status of this work?

maxnoe:
I guess we can close this PR, it will probably be easier to start again from the current main branch then to update here

PR #3828, opened 2022-02-25
Take into account method argument in WcsNDMap.interp_by_coord

### LINKED ISSUE #3812: interp_by_coord gives same results for "nearest" and "linear" interpolation
**Gammapy version**
version 0.19

**Bug description**
Interpolating a Map using interp_by_coord() gives the same result independent of the method. The method keyword is actually passed to the ScaledRegularGridInterpolator but it doesn't seem to make any difference. The interpolation is linear. Other keywords like points_scale and values_scale are not passed but probably should be i.e. an energy axis should be interpolated in log scale.

**Expected behavior**
Actually performing a nearest or linear interpolation for the respective method. Passing the correct scales to ScaledRegularGridInterpolator.

**To Reproduce**
```
import numpy as np
from gammapy.maps import Map, MapAxis

axis = MapAxis.from_energy_bounds("1 TeV", "10 TeV", nbin=3)

m = Map.create(npix=(3, 3))
m.data += np.arange(9).reshape((3, 3))

print(m.interp_by_coord({"lon": 0.07, "lat": 0.03}, method="linear"))
print(m.interp_by_coord({"lon": 0.07, "lat": 0.03}, method="nearest"))
```

adonath:
**Description**

This pull request fixes #3812, by correctly taking into account the method keyword. I also decided to change the default interpolation method to "linear". Otherwise the default behavior of `interp_by_coord()` would be the same as `.get_by_coord()`. This slight changes some test values, which are adapted as well.


QRemy:
Thanks @adonath, I have no specific comments on the changes here, but looking where `interp_by_coord` is used I found some inconsistencies in the docstings :
- `TemplateSpatialModel`  still  mention `interp` instead of `method`
https://github.com/gammapy/gammapy/blob/dd35f7a27c2f939fa6af3080d9c0ec0d1029260e/gammapy/modeling/models/spatial.py#L1036-L1037

-  same for `TemplateNDSpectralModel` and in addition it mentions using `interp_by_coord` while it relies on `interp_by_pix` 
https://github.com/gammapy/gammapy/blob/dd35f7a27c2f939fa6af3080d9c0ec0d1029260e/gammapy/modeling/models/spectral.py#L1576-L1577

gammapy/utils/interpolation.py
```diff
@@ -79,16 +79,17 @@ def __init__(
...
         Parameters
         ----------
         points : tuple of `~numpy.ndarray` or `~astropy.units.Quantity`
             Tuple of coordinate arrays of the form (x_1, x_2, x_3, ...). Arrays are
             broadcasted internally.
-        method : {"linear", "nearest"}
-            Linear or nearest neighbour interpolation.
+        method : {None, "linear", "nearest"}
+            Linear or nearest neighbour interpolation. None will choose the default
+            defined on init.
```
registerrier, line 92: Maybe this could be made more explicit since `method` is not an explicit argument of `ScaledRegularGridInterpolator.__init__()`

registerrier (approved):
Thanks @adonath !

Maybe change the default `method` on the base class `Map.interp_by_coord` and `Map.interp_by_pix` to `"linear"` as well?

gammapy/utils/interpolation.py
adonath: I extended the docstring.

adonath:
@registerrier  and @QRemy  I adapted the docstings and will merge now...

PR #3829, opened 2022-02-25
Move content in API to User Guide

Bultako:
This PR addresses issue #3818 moving content present in the API Reference pages to the Package Structure section in the User Guide. It also fixes the TOC side bars so do they do not scroll with the central frame and adds a second level of depth in the main TOC menu of the left.

You may access how the docs shows with this PR in the [usual link](http://www.iaa.csic.es/~jer/gammapydocs).


registerrier (approved):
Thanks @Bultako .

 This looks good. I think this is good starting basis for the doc reorganization. We should not wait too long for feedback, but rather merge and make further changes from the new structure.

There is a merge conflict. Can you resolve it?

Bultako:
@registerrier
ok, conflict solved after rebase

Bultako:
@adonath 
I guess you mean updating the content [here](https://www.iaa.csic.es/~jer/gammapydocs/).
It is actually up-to-date to this PR.

adonath:
Thanks indeed! I just haven't noticed that the headlines on https://www.iaa.csic.es/~jer/gammapydocs/userguide/package.html are now links.

 This is probably also my main comment: I think it's the only place where we use headers as links to new pages and at least to me this was very unintuitive. Maybe just adding an explicit link to the sub-page at the end of the paragraph is even more visible. Having the documentation hidden in the "package structure" is maybe non-ideal as well. But for now I don't have a better proposal...

Bultako:
@adonath 
I have replaced the links in the titles with `more info` links at the end of each subsection.
You can check it in the usual preview.

adonath (approved):
Thanks a lot @Bultako, I think this is the more visible solution for now. In general I think we can still improve the overall structure to make the content even easier accessible. But this is definitely future work and not related to this PR. So no further comments from my side.

Bultako:
ok
merging now.

PR #3830, opened 2022-02-25
Improve FOVALIGN check for background

maxnoe:
This makes the check more explicit and always passes the `fixed_pointing_info`.

The unit test before was wrong.
It passed a `FixedPointingInfo`, which was the convoluted signal for
"FOVALIGN" == "ALTAZ" for a Background3D with no FOVALIGN, which is not
allowed by the standard, but gammapy assumes to mean "RADEC".

### DIFF
--- a/gammapy/makers/map.py
+++ b/gammapy/makers/map.py
@@ -3,7 +3,7 @@ import logging
 import astropy.units as u
 from astropy.table import Table
 from regions import PointSkyRegion
-from gammapy.irf import EDispKernelMap, PSFMap, FoVAlignment
+from gammapy.irf import EDispKernelMap, PSFMap
 from gammapy.maps import Map
 from .core import Maker
 from .utils import (
--- a/gammapy/makers/map.py
+++ b/gammapy/makers/map.py
@@ -210,18 +210,13 @@ class MapDatasetMaker(Maker):
         if isinstance(bkg, Map):
             return bkg.interp_to_geom(geom=geom, preserve_counts=True)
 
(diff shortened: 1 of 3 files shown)


adonath:
Thanks @maxnoe, looks good to me! I think it still requires the change from https://github.com/gammapy/gammapy/pull/3797/files#diff-f410b52b9f0c9268f119f5d9c1c05473181cf8d183a0516b3a3a662703f7260dR262 to get the test to pass and handle the case, where there is no events present. This was previously no triggered, because no `FixedPointingInfo` was created...

adonath:
I guess it makes sense to merge #3832 first and rebase after...

registerrier (approved):
Thanks @maxnoe . This looks good.

Apparently the previous behavior was not correct, right?

maxnoe:
> Apparently the previous behavior was not correct, right?

The previous behaviour *in the test* was not correct. In actual code when run through the makers it behaved correctly but it was a very strange and not straight-forward way of checking the fov alignment.

adonath:
Thanks @maxnoe for the update! I think this still requires the fix from https://github.com/gammapy/gammapy/pull/3797/files#diff-f410b52b9f0c9268f119f5d9c1c05473181cf8d183a0516b3a3a662703f7260dR262. You could either include it here, or wait for #3797 to be merged. Do as you prefer...

maxnoe:
@adonath Yes, let's wait for the other PR and then I'll rebase

adonath:
@maxnoe Please rebase...thanks!

adonath (approved):
Thanks @maxnoe, I have no further comments...

PR #3831, opened 2022-02-25
Serialise meta information on Dataset

AtreyeeS:
As an user, I find these information rather useful. Adding them to the meta table.

### DIFF
--- a/gammapy/datasets/map.py
+++ b/gammapy/datasets/map.py
@@ -1116,6 +1116,9 @@ class MapDataset(Dataset):
         if self.gti is not None:
             hdulist.append(fits.BinTableHDU(self.gti.table, name="GTI"))
 
+        if self.meta_table is not None:
+            hdulist.append(fits.BinTableHDU(self.meta_table, name="META_TABLE"))
+
         return hdulist
 
     @classmethod
@@ -1191,6 +1194,10 @@ class MapDataset(Dataset):
         if "GTI" in hdulist:
             gti = GTI(Table.read(hdulist, hdu="GTI"))
             kwargs["gti"] = gti
+
+        if "META_TABLE" in hdulist:
+            meta_table = Table.read(hdulist, hdu="META_TABLE")
+            kwargs["meta_table"] = meta_table
 
         return cls(**kwargs)
 
@@ -2441,6 +2448,10 @@ class MapDatasetOnOff(MapDataset):
         if "GTI" in hdulist:
             gti = GTI(Table.read(hdulist, hdu="GTI"))
             kwargs["gti"] = gti
+
+        if "META_TABLE" in hdulist:
+            meta_table = Table.read(hdulist, hdu="META_TABLE")
+            kwargs["meta_table"] = meta_table
         return cls(**kwargs)
 
     def info_dict(self, in_safe_data_range=True):
--- a/docs/userguide/dl4.rst
+++ b/docs/userguide/dl4.rst
@@ -69,7 +69,9 @@ analyse the following datasets:
      -
      - `chi2`
 
-
+In addition to the above quantities, a dataset can optionally have a
+`meta_table` serialised, which can contain relevant information about the observations
+used to create the dataset.
 In general, `OnOff` datasets should be used when the
 background is estimated from real off counts,
 rather than from a background model.
--- a/gammapy/datasets/tests/test_map.py
+++ b/gammapy/datasets/tests/test_map.py
@@ -1106,6 +1106,7 @@ def get_map_dataset_onoff(images, **kwargs):
 @requires_data()
 def test_map_dataset_on_off_fits_io(images, tmp_path):
     dataset = get_map_dataset_onoff(images)
+    dataset.meta_table = Table(data=[[1.0*u.h], [111]], names=["livetime", "obs_id"])
     gti = GTI.create([0 * u.s], [1 * u.h], reference_time="2010-01-01T00:00:00")
     dataset.gti = gti
 
@@ -1129,6 +1130,7 @@ def test_map_dataset_on_off_fits_io(images, tmp_path):
         "MASK_SAFE",
         "MASK_SAFE_BANDS",
         "GTI",
+        "META_TABLE",
         "COUNTS_OFF",
         "COUNTS_OFF_BANDS",
         "ACCEPTANCE",
@@ -1143,6 +1145,8 @@ def test_map_dataset_on_off_fits_io(images, tmp_path):
 
     dataset_new = MapDatasetOnOff.read(tmp_path / "test.fits")
     assert dataset_new.mask.data.dtype == bool
+    assert dataset_new.meta_table["livetime"] == 1.0*u.h
+    assert dataset_new.meta_table["obs_id"] == 111
 
     assert_allclose(dataset.counts.data, dataset_new.counts.data)
     assert_allclose(dataset.counts_off.data, dataset_new.counts_off.data)


gammapy/makers/map.py
```diff
@@ -340,6 +340,12 @@ def make_meta_table(observation):
         meta_table["OBS_ID"] = [observation.obs_id]
         meta_table["RA_PNT"] = [observation.pointing_radec.icrs.ra.deg] * u.deg
         meta_table["DEC_PNT"] = [observation.pointing_radec.icrs.dec.deg] * u.deg
+        try:
+            meta_table["ZEN_PNT"] = [observation.pointing_zen]
+            meta_table["LIVETIME"] = [observation.observation_live_time_duration]
+        except KeyError:
```
maxnoe, line 346: It's very non-obvious why this would raise a `KeyError`. You are accessing properties here. One needs to know that these properties in turn do a key lookup in the `meta` dict. 

Also logging a warning about missing optional information is not really good I think.

AtreyeeS: I can make it into an info statement. We generally have warnings in Gammapy, so I stayed with warning. But I agree

maxnoe: This is not a warning, it is a log message with level warning. This is an important distinction. Also, if `ZEN_PNT` is not there, you could check for `ALT_PNT` and calculate it.

gammapy/datasets/map.py
```diff
@@ -629,7 +631,9 @@ def create(
         )
 
         kwargs.update(geoms)
-        return cls.from_geoms(reference_time=reference_time, name=name, **kwargs)
+        return cls.from_geoms(
```
registerrier, line 634: If this was not there before, this probably means we miss a test to check that a `meta_table` is actually present if passed on creation.

gammapy/makers/map.py
registerrier: Another possible issue with the implementation is that if the first one is not there then no livetime will be added.
This would require a try-except statement for each entry

registerrier (commented):
Thanks @AtreyeeS . It is important to improve the capabilities associated with the `Dataset.meta_table`. 
It is not clear to determine how much information should be put there. We probably should make it configurable.

In this regard, the current behavior is strange. One can add a `meta_table` on dataset creation, but the current `Maker` creates an empty `Table`. A possible behavior could then be that if a `meta_table` is defined, the `Maker` will try to fill the columns with the corresponding keys in the `observation.obs_info`.

adonath:
> It is not clear to determine how much information should be put there. We probably should make it configurable.

Just wanted to quickly comment here: I think making the selection of meta data configurable is almost certainly a bad idea. There will be much more meta data in future, which makes the selection complex and in general I think the amount of meta data to be stored should be as much as possible (or as much as useful and then a bit more...). In any case I think the selection should be made by Gammapy and the supported data formats and not by users. Otherwise we will end up with very inconsistent DL4 files, where in some cases important meta data might be missing. However there will probably be required and optional meta data information...

registerrier:
> Just wanted to quickly comment here: I think making the selection of meta data configurable is almost certainly a bad idea. There will be much more meta data in future, which makes the selection complex and in general I think the amount of meta data to be stored should be as much as possible (or as much as useful and then a bit more...). In any case I think the selection should be made by Gammapy and the supported data formats and not by users. Otherwise we will end up with very inconsistent DL4 files, where in some cases important meta data might be missing. However there will probably be required and optional meta data information...

I partly agree here. This PR is precisely an example showing that some analysis/people need more informations than the default ones. 

If we prefer not making it configurable, we have to make the `meta_table` maximal and e.g. copy all the entries of the `Observation.obs_info` plus other mandatory attributes of `Observation`.

adonath:
> If we prefer not making it configurable, we have to make the meta_table maximal and e.g. copy all the entries of the Observation.obs_info plus other mandatory attributes of Observation.

Yes, I think this is the better option. Typically one finds that meta information is missing, never the other way around. It's always better to have the information not used, than have it missing when needed.

AtreyeeS:
Thanks for the discussions. I would also prefer to have the maximal info.

 1. `Observation.obs_info` for the H.E.S.S. DL3 observations give a lot of info:
```
{'MJDREFI': 51910,
 'MJDREFF': 0.000742870370370241,
 'TIMEUNIT': 's',
 'TIMESYS': 'TT',
 'EXTNAME': 'OBS_INDEX',
 'TIMEREF': 'local',
 'GEOLON': '16.5002222222222',
 'GEOLAT': '-23.2717777777778',
 'ALTITUDE': '1835.',
 'comments': ['This file is part of the first H.E.S.S. FITS test data release from 2018',
  'Webpage: https://www.mpi-hd.mpg.de/hfm/HESS/pages/dl3-dr1/',
  'Contact: contact@hess-experiment.eu.'],
 'OBS_ID': 23523,
 'RA_PNT': <Quantity 83.63333 deg>,
 'DEC_PNT': <Quantity 21.514444 deg>,
 'GLON_PNT': <Quantity 184.9823 deg>,
 'GLAT_PNT': <Quantity -6.0516925 deg>,
 'ZEN_PNT': <Quantity 48.61021 deg>,
 'ALT_PNT': <Quantity 41.38979 deg>,
 'AZ_PNT': <Quantity 22.481705 deg>,
 'OBJECT': 'Crab Nebula',
 'RA_OBJ': <Quantity 83.63333 deg>,
 'DEC_OBJ': <Quantity 22.014444 deg>,
 'OFFSET_OBJ': <Quantity 0.5 deg>,
 'ONTIME': <Quantity 1687. s>,
 'LIVETIME': <Quantity 1581.7368 s>,
 'DEADC': 0.9376033,
 'TSTART': <Quantity 1.23890826e+08 s>,
 'TSTOP': <Quantity 1.23892513e+08 s>,
 'DATE-OBS': '2004-12-04',
 'TIME-OBS': '22:08:10.184',
 'DATE-END': '2004-12-04',
 'TIME-END': '22:36:17.184',
 'N_TELS': 4,
 'TELLIST': '1,2,3,4',
 'QUALITY': 0,
 'MUONEFF': 0.9070748,
 'EVENT_COUNT': 7613,
 'TARGET_NAME': 'Crab',
 'TARGET_TAG': 'crab',
 'TARGET_OFFSET': <Quantity 0.500049 deg>,
 'SAFE_ENERGY_LO': <Quantity 0.8709636 TeV>,
 'SAFE_ENERGY_HI': <Quantity 100. TeV>}
```
 
carrying over all of this can be redundant (eg: webpage, contact, etc). Even if we agree to have redundant info, I don't think we should copy over  `SAFE_ENERGY` keywords, because the safe energy of the dataset is usually different from that of the observation.

2. For an observation created on the fly
(a) passing parameters on init
```
reference_time = Time("2010-12-31 00:00:00", scale="utc", format="iso")
start = [0 * u.hr]
stop = [48 * u.hr]
gti = GTI.create(reference_time=reference_time, start=start, stop=stop)
o1 = Observation(obs_id=1,
                gti=gti,
                aeff=observations[0].aeff,
                events=observations[0].events)
```
`o1.obs_info` is blank (since no `obs_info=None` on init, it is not re-created. I would say it is okay, and users should always go through `Observation.create` (that is indeed what we show in our notebooks, it is only in some tests that we create through init)

(b) Using `Observation.create` : `obs_info` is incomplete
```
o2 = Observation.create(pointing=observations[0].pointing_radec,
                       obs_id=1,
                       tstart=0 * u.hr,
                       tstop=48 * u.hr,
                       irfs={"aeff":observations[0].aeff})
```
`o2.obs_info` is just `{'RA_PNT': 83.63333129882812, 'DEC_PNT': 21.51444435119629, 'DEADC': 1.0}`
even though `o2.observation_live_time_duration` is correctly computed. `o2.pointing_zen` is `Nan` in this case, which is fine as well.

So either we first improve the `obs_info` to have the maximal(?) information, or we access the available attributes on `Observation` for `Dataset.meta_table` (but I don't know how to automatically loop over all attributes).

maxnoe:
@AtreyeeS That is then very much connected to the restructuring of the data / metadata models.

If we have e.g. classes for the observatory metadata and the observation metadata, we'd know which changes per run and which should be the same for all.

AtreyeeS:
@maxnoe : So how would that work? we have an `Observatory` class and an `Observation` class? or `Observation` has two types of info dicts, `obs_info` and `observatory_info`?

adonath:
> That is then very much connected to the restructuring of the data / metadata models.

I completely agree. A solution for the observation meta data also needs to provide the possibility to pass on the information to higher data levels. Meta data will also be stored on the reduced `IRF` and `Map` classes, some at the `Dataset` level etc.

AtreyeeS:
Realised that the meta table is not serialised !

AtreyeeS:
A minor side comment, maybe we can name the meta information similarly at different level. presently its
- `Observation.obs_info`
- `Dataset.meta_table`
- `Datasets.info_table`

this will be API breaking though, so not too nice.

And also add info on `Observations`?

AtreyeeS:
Leaving aside the discussion on meta information for later, I modify this PR to simply serialise MapDatasets

Again, I do not know what to do for Spectral Datasets. The OGIP meta information is rather different.
Suggestions @registerrier @adonath ?

registerrier (dismissed):

Do you plan to add it in OGIP read/write or would you see it as a future PR?

AtreyeeS:
I am doing it now @registerrier. I will ask for a re-review once it is ready

adonath:
@registerrier and @AtreyeeS My only concern with simply adding more meta data to the OGIP serialization is documentation. I think Ideally we define first and document the allowed meta somewhere, before we actually implement it. I'm not sure where this would happen probably GADF?

AtreyeeS:
I added a meta information on `SpectrumDatasetOnOff`
As discussed in the last dev call, this is just an extra astropy table residing along with the OGIP.

The content of the meta table is a separate discussion, and DL4 serialisation is anyways poorly documented in GADF, so maybe we can merge this without addressing the above?

adonath:
> and DL4 serialisation is anyways poorly documented in GADF

I think this is a very weak argument. Negatively speaking, if we start introducing modifications and additions to the serialization formats and don't even try to document them it is going to end in a disaster. Especially if modifications are introduced on a PR by PR basis. This means there will be many versions of files of which some have the information, some older versions don't, some will have a subset etc.  I think we should try to avoid this situation from the start.

Given there is also activity on the meta data handling for IRFs and observations, I think we now have the chance to do it right and not continue with improvised solutions. But this comment was more about the internal handling of meta data in Gammapy...not really related to this PR.

Concerning the question of serialization: I think the first question to resolve here is whether to extend the existing OGIP or whether to introduce a new "native" Gammapy format. The current implementation, where the table contains lifetime and obs id duplicates the information that is already present in the PHA header (https://github.com/gammapy/gammapy/blob/master/gammapy/datasets/io.py#L105). I think this we should probably avoid... 

When extending OGIP I think we would first propose the addition here: https://gamma-astro-data-formats.readthedocs.io/en/v0.2/spectra/ogip/index.html I.e. document the existence of the meta table in the HDU list, describe its structure and add the allowed content / column names.

If we go into the direction of a native Gammapy format we should at least document the content of the meta table in our documentation (which is a good idea in any case...).

AtreyeeS:
From the discussions at the last dev call, I understood that it would be okay to add a serialisation of the meta table without changing OGIP. But since this is leading to issues, I propose

- to revert this commit and merge this PR adding only a meta table serialisation on `MapDataset`
-  In a follow up PR, add an option on `SpectrumDataset.write(format="gadf")`. Then, spectrum datasets can be serialised exactly the same way as `MapDatasets` (the option is just prohibited now). `SpectrumDatasetOnOff` can then be serialised either as `ogip` , or as `gadf`, and the meta information added accordingly. 
- I don't know if the above step needs to be discussed within the GADF, or it should be enough to document it within Gammapy
- The discussion on the exact content of the meta table can happen later.

Does that sound okay @registerrier @adonath ?

adonath:
Thanks @AtreyeeS, I think this is a good proposal! The introduction of a generic format is useful in any case.

When I implemented the serialization format for region maps I was optimistic and proposed to GADF and already called the format "gadf". However the PR is still open and the activity has stalled, which is bad. I remember at that time @registerrier proposed to name the format different and keep it native to Gammapy. Maybe this would have been the better option. On the other hand the DL4 `WcsNDMap`  and `HpsNDMap` serialization is already in GADF so I thought it was obvious to do the same for the `RegionNDMap`. Maybe we could use your proposal to revive the discussion in GADF again and push for a quick acceptance of https://github.com/open-gamma-ray-astro/gamma-astro-data-formats/pull/170. This is then the basis for serialization of the `SpectrumDataset` as well.

adonath:
> The discussion on the exact content of the meta table can happen later.

OK for me. I think the expectations on stability of the generic format are less than for OGIP or more "official" formats.

AtreyeeS:
Okay, I modified this PR. Once this is merged, I will open a new one for Spectral datasets, and re-start discussion on your GADF PR

registerrier:
Technically, I don't see how the addition of a table HDU to the PHA file breaks the OGIP conventions. The latter only contain required HDUs to allow for spectral fitting  packages. There is nothing about optional data. Is there any other way to serialize a table than an additional table HDU? 

Technically we are already non fully consistent with OGIP convention when we assume log-scale binning of table such as rmf. This should a more serious concern.
  
Note that I don't expect to see anything from gadf on DL4 any time soon.

adonath:
> Technically, I don't see how the addition of a table HDU to the PHA file breaks the OGIP conventions. The latter only contain required HDUs to allow for spectral fitting packages. There is nothing about optional data. Is there any other way to serialize a table than an additional table HDU?

My concern was about the documentation of this, not that it would break any OGIP convention. I think technically it is fine if we add the meta table to the OGIP format but as this format is already documented in GADF we should (!) propose this addition there first and not "silently" introduce it in Gammapy. My main point here is that Gammapy should only write in well defined and documented formats.

Here it gets a bit nitpicky, but the generic dataset serialization is so far not documented in GADF , so I would consider it native to Gammapy. Adding the meta data there I would consider simpler and documentation in Gammapy is sufficient for now. To avoid any confusion we should probably not call the `MapDataset` / `SpectrumDataset` format "gadf" then, but maybe "gp-generic" or similar. I hope this make sense...

> Technically we are already non fully consistent with OGIP convention when we assume log-scale binning of table such as rmf. This should a more serious concern.

I agree, this should be clearly documented here https://gamma-astro-data-formats.readthedocs.io/en/v0.2/spectra/ogip/index.html

gammapy/datasets/map.py
```diff
@@ -1192,6 +1195,10 @@ def from_hdulist(cls, hdulist, name=None, lazy=False, format="gadf"):
             gti = GTI(Table.read(hdulist, hdu="GTI"))
             kwargs["gti"] = gti
 
+        if "META_TABLE" in hdulist:
```
adonath, line 1198: Minor comment: maybe rename just to `"META"`? For the other HDU names we don't put the type of the HDU either (like `COUNTS_MAP` or `COUNTS_IMAGE`)

adonath (commented):
Thanks a lot @AtreyeeS, I have left one minor comment on the HDU name. Otherwise I think it would be important to document the content of the meta table somewhere in the Gammapy docs.

adonath:
>A minor side comment, maybe we can name the meta information similarly at different levels. presently its

> Observation.obs_info
> Dataset.meta_table
> Datasets.info_table

This is not correct, we have:

- `Observation.obs_info`
- `Dataset.meta_table`
- `Datasets.meta_table`

So on the datasets level this is already consistent.


>this will be API breaking though, so not too nice.

We could rename `Observation.obs_info` just to `Observation.meta` and if need keep the alias (but probably not worth it, because users rarely access the info from the `Observation.obs_info`)

>And also add info on Observations?

I agree `Observations.meta_table` would be convenient, however there is the problem, that it would be created on fly (because we decided to make `Observations` a mutable sequence...), which will not be efficient if there are many hundred observations.

gammapy/datasets/map.py
AtreyeeS: I called it `META_TABLE` to have the same name as the class attribute

adonath: OK, makes sense.

adonath:
Thanks @AtreyeeS , let's wait for the CI to pass and then merge...

adonath:
CI fails are unrelated. I'll go ahead and merge this now. However in a follow-up PR I'll add a note in the dataset docs section and mention the meta data serialization format is not stable and likely subject to change in future versions of Gammapy. OK @AtreyeeS?

AtreyeeS:
Thanks @adonath ! I thought of doing that once we have some idea of what goes in, eg, "the meta table contains information on the pointing position, the zenith angle, etc". However, both the content and the format of this information  is not stable and likely subject to change in future versions of Gammapy". But if you want to put it in now, sure!

PR #3832, opened 2022-02-28
Add fov_alignment and is_pointlike attributes to IRF

maxnoe:
**Description**

This pull request continues with the low hanging fruits of #3767, adding the `fov_alignment` and `is_pointlike` attributes to the `IRF` class, parsing/passing the information from the meta in `{to,from}_table`.

### DIFF
--- a/gammapy/irf/core.py
+++ b/gammapy/irf/core.py
@@ -68,7 +68,7 @@ class IRF(metaclass=abc.ABCMeta):
         axes = MapAxes(axes)
         axes.assert_names(self.required_axes)
         self._axes = axes
-        self.fov_alignment = FoVAlignment(fov_alignment)
+        self._fov_alignment = FoVAlignment(fov_alignment)
         self._is_pointlike = is_pointlike
 
         if isinstance(data, u.Quantity):
@@ -104,6 +104,11 @@ class IRF(metaclass=abc.ABCMeta):
     def has_offset_axis(self):
         """Whether the IRF explicitly depends on offset"""
         return "offset" in self.required_axes
+
+    @property
+    def fov_alignment(self):
+        """Alignment of the field of view coordinate axes, see `FoVAlignment`"""
+        return self._fov_alignment
 
     @property
     def data(self):
--- a/gammapy/irf/core.py
+++ b/gammapy/irf/core.py
@@ -13,9 +13,22 @@ from gammapy.utils.interpolation import (
     interpolation_scale,
 )
 from gammapy.utils.scripts import make_path
-from .io import IRF_DL3_HDU_SPECIFICATION, IRF_MAP_HDU_SPECIFICATION
+from enum import Enum
+from .io import IRF_DL3_HDU_SPECIFICATION, IRF_MAP_HDU_SPECIFICATION, gadf_is_pointlike
 
 log = logging.getLogger(__name__)
+
+
+class FoVAlignment(str, Enum):
+    '''
+    Orientation of the Field of View Coordinate System
+
+    Currently, only two possible alignments are supported: alignment with
+    the horizontal coordinate system (ALTAZ) and alignment with the equatorial
+    coordinate system (RADEC).
+    '''
+    ALTAZ = "ALTAZ"
+    RADEC = "RADEC"
 
 
 class IRF(metaclass=abc.ABCMeta):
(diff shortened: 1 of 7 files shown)


gammapy/irf/core.py
```diff
@@ -38,10 +44,22 @@ class IRF(metaclass=abc.ABCMeta):
         fill_value=0.0,
     )
 
-    def __init__(self, axes, data=0, unit="", meta=None, interp_kwargs=None):
+    def __init__(
+        self,
+        axes,
+        data=0,
+        unit="",
+        is_pointlike=False,
```
adonath, line 52: Add documentation to docstring?

gammapy/irf/core.py
```diff
@@ -38,10 +44,22 @@ class IRF(metaclass=abc.ABCMeta):
...
+        fov_alignment=FoVAlignment.RADEC,
+        meta=None,
+        interp_kwargs=None,
+    ):
         axes = MapAxes(axes)
         axes.assert_names(self.required_axes)
         self._axes = axes
+        self.fov_alignment = FoVAlignment(fov_alignment)
```
adonath, line 60: I guess all meta attributes should be read-only as well? I was wondering, whether it's worth to group all meta info in its own namespace? E.g. by introducing an `IRFMeta` class or similar, which is accessible as `IRF.meta`. This keeps the `IRF` base class more general, because I could imaging that the amount of meta information will grow in future, when we introduce e.g. event classes and types...

adonath (commented):
Thanks a lot @maxnoe, I think this goes exactly in the right direction! 

I left two minor comments on documentation and possibly grouping meta data in its own namespace. I think the latter would lead to simpler code structure overall in future...

gammapy/irf/core.py
maxnoe: I am not sure about that. Especially at the moment, we cannot really call it `meta`, since that still needs to be the actual FITS header data.

These attributes really belong to the IRF itself, since otherwise the `.data` cannot be interpreted correctly.

gammapy/irf/core.py
maxnoe: done

gammapy/irf/core.py
maxnoe: > I guess all meta attributes should be read-only as well? 

That would mean creating either properties for all of them, or adding a `__setattr__` that raises an error. If we decide to make IRFs immutable (at least semantically, it's not possible to really do that in python), I'd go for the removed `__setattr__` instead of adding all the properties code.

maxnoe: See e.g. how dataclasses are implementing the `frozen` parameter:

https://docs.python.org/3/library/dataclasses.html#frozen-instances

adonath: > I am not sure about that. Especially at the moment, we cannot really call it meta, since that still needs to be the actual FITS header data.

Yes, for now any proposed solution will lead to cleaner code.

> These attributes really belong to the IRF itself, since otherwise the .data cannot be interpreted correctly.

I agree, however the `IRFMeta` can be mandatory as well. As I said for me it's mostly a question of code organization / definition of namespace. I could possibly imagine that meta information could be read and validated independently of the actual data of the IRF...but lm totally fine to postpone this discussion or open a separate issue to discuss it there.

>  See e.g. how dataclasses are implementing the frozen parameter:

Yes, I think technically this is definitely a good solution for an `IRFMeta` class. Applying this pattern to the `IRF` class, I'm not sure yet. I think historically we have worked in Gammapy with properties, but only because the Gammapy code is older than `dataclasses`. However for consistency it might be preferable to continue with properties for now, except if we find us adding new properties in more than 10-15 places.

maxnoe: Ok, I added a property for `fov_alignment` and a test ensuring that `setattr` doesn't work for them

adonath (dismissed):
Thanks @maxnoe, no further comments from my side! The remaining test fail in the CI seems unrelated...

maxnoe:
Yes, I'll rebase master for the MAGIC DL3 fix

maxnoe:
Done, tests are passing

registerrier (approved):
Thanks @maxnoe ! Looks good.

PR #3834, opened 2022-03-01
Rename IRF.is_offset_dependent to IRF.has_offset_axis

maxnoe:
**Description**

As discussed with @adonath via slack, the `IRF.is_offset_dependent` property is badly named, since a `Background3D` for example is definitely offset dependent, albeit not explicitly via the `offset` axis.

Renamed to the more explicit `has_offset_axis`

### DIFF
--- a/gammapy/data/observations.py
+++ b/gammapy/data/observations.py
@@ -340,7 +340,7 @@ class Observation:
         if "bkg" in self.available_irfs:
             bkg = self.bkg
 
-            if not bkg.is_offset_dependent:
+            if not bkg.has_offset_axis:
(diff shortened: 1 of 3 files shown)


registerrier (approved):
Thanks @maxnoe!

Is an attribute based on axes the more general? I suppose we don't expect IRF including an azimuth axis without any offset axis.

adonath (approved):
Thanks @maxnoe, this is definitely the more clearer name already. In general it could be even more clearer by explicitly declaring the coordinate system. But this can be part of the larger improvement to the IRF meta handling...

PR #3836, opened 2022-03-02
Minor fix to evaluation_bin_size_min for Shell2SpatialModel

luca-giunti:
Fixed this broken line: https://github.com/gammapy/gammapy/blob/dd35f7a27c2f939fa6af3080d9c0ec0d1029260e/gammapy/modeling/models/spatial.py#L890

`self.r_0` -> `self.r_0.quantity`

### DIFF
--- a/gammapy/modeling/models/spatial.py
+++ b/gammapy/modeling/models/spatial.py
@@ -887,7 +887,7 @@ class Shell2SpatialModel(SpatialModel):
 
         The bin min size is defined as r_0*eta.
         """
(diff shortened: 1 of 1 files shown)


QRemy:
Thanks @luca-giunti, maybe you could add a test checking the value of evaluation_bin_size_min for the spatial models that define it.

registerrier (approved):
Thanks @luca-giunti . Looks good to me.

adonath (approved):
Thanks @luca-giunti!

PR #3837, opened 2022-03-02
Allow nearest neighbor interpolation with scalar data

### LINKED ISSUE #3835: IRF.evaluate fails for IRFs with single energy bin
**Gammapy version**

Current master

**Bug description**

`IRF.evaluate` fails in the interpolation code when there is only a single energy bin, as for example is used with the global RAD_MAX:

**Expected behavior**

Returning the single global value.

**To Reproduce**

```
In [1]: from gammapy.irf import RadMax2D
In [2]: from gammapy.maps import MapAxis
In [3]: energy_axis = MapAxis.from_energy_bounds(0.01, 100, 1, unit="TeV")
In [4]: offset_axis = MapAxis.from_bounds(0., 5, 1, unit="deg", name="offset", )
In [5]: import astropy.units as u
In [6]: rad_max = RadMax2D(data=[[0.1]] * u.deg, axes=[energy_axis, offset_axis])
In [7]: rad_max.evaluate(1 * u.TeV)
ValueError                                Traceback (most recent call last)
Input In [7], in <module>
----> 1 rad_max.evaluate(1 * u.TeV)

File ~/Projects/gammapy/gammapy/irf/core.py:239, in IRF.evaluate(self, method, **kwargs)
    237     if coord is not None:
    238         coords_default[key] = u.Quantity(coord, copy=False)
--> 239 data = self._interpolate(coords_default.values(), method=method)
    241 if self.interp_kwargs["fill_value"] is not None:
    242     idxs = self.axes.coord_to_idx(coords_default, clip=False)

File ~/Projects/gammapy/gammapy/utils/interpolation.py:106, in ScaledRegularGridInterpolator.__call__(self, points, method, clip, **kwargs)
    104 if self.axis is None:
    105     points = np.broadcast_arrays(*points)
--> 106     points_interp = np.stack([_.flat for _ in points]).T
    107     values = self._interpolate(points_interp, method, **kwargs)
    108     values = self.scale.inverse(values.reshape(points[0].shape))

File <__array_function__ internals>:180, in stack(*args, **kwargs)

File ~/.local/anaconda/envs/gammapy-dev/lib/python3.8/site-packages/numpy/core/shape_base.py:422, in stack(arrays, axis, out)
    420 arrays = [asanyarray(arr) for arr in arrays]
    421 if not arrays:
--> 422     raise ValueError('need at least one array to stack')
    424 shapes = {arr.shape for arr in arrays}
    425 if len(shapes) != 1:

ValueError: need at least one array to stack
```

adonath:
**Description**

This pull request fixes #3835, by allowing nearest neighbor interpolation with scalar data in the `ScaledRegularGridInterpolator`.

**Dear reviewer**
- Should we raise an error when scalar data is interpolated with `method="linear"` or just silently use `"nearest"`?

### DIFF
--- a/gammapy/data/event_list.py
+++ b/gammapy/data/event_list.py
@@ -769,6 +769,34 @@ class EventList:
         mask &= offset < offset_band[1]
         return self.select_row_subset(mask)
 
+    def select_rad_max(self, rad_max, position=None):
+        """Select energy dependent offset
+
+        Parameters
+        ----------
+        rad_max : `~gamapy.irf.RadMax2D`
+            Rad max definition
+        position : `~astropy.coordinates.SkyCoord`
+            Center position. By default the pointing position is used.
+
+        Returns
+        -------
+        event_list : `EventList`
+            Copy of event list with selection applied.
+        """
+        if position is None:
+            position = self.pointing_radec
+
+        offset = position.separation(self.pointing_radec)
+        separation = position.separation(self.radec)
+
+        rad_max_for_events = rad_max.evaluate(
+            method="nearest", energy=self.energy, offset=offset
+        )
+
+        selected = separation <= rad_max_for_events
+        return self.select_row_subset(selected)
+
     @property
     def is_pointed_observation(self):
         """Whether observation is pointed"""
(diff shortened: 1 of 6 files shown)


maxnoe:
> Should we raise an error when scalar data is interpolated with method="linear" or just silently use "nearest"?

I'd say raise an error. Falling back to nearest is probably not the right choice every time, so it shouldn't be done silently

registerrier (dismissed):

Regarding your question I suppose we don't expect many IRFs to come with only 1-element axes. Raising an exception is probably fine.

PR #3838, opened 2022-03-04
Example of a spectrally varying temporal model

AtreyeeS:
This is a test implementation of a model with time dependent spectral variability. For now, I simply add it as a an example in our models tutorial, as a custom model.

If we decide this is the way to do it, I need to update `TemporalModel.plot()` to accept an energy axis. And of course, polish the tutorial. But before putting in more time, it would be nice to have your opinion if we want to show this @registerrier @adonath @QRemy

### DIFF
--- a/gammapy/modeling/models/temporal.py
+++ b/gammapy/modeling/models/temporal.py
@@ -420,7 +420,6 @@ class GeneralizedGaussianTemporalModel(TemporalModel):
         t_decay = pars["t_decay"].quantity
         eta = pars["eta"].quantity
         t_ref = Time(pars["t_ref"].quantity, format = "mjd")
-
         integral = scipy.integrate.quad(self.evaluate, t_min.mjd, t_max.mjd, args=(t_ref.mjd,t_rise,t_decay,eta))[0]
         return integral / self.time_sum(t_min, t_max).to_value("d")
 
--- a/gammapy/modeling/models/temporal.py
+++ b/gammapy/modeling/models/temporal.py
@@ -34,7 +34,7 @@ class TemporalModel(ModelBase):
 
     _type = "temporal"
 
-    def __call__(self, time):
+    def __call__(self, time, energy=None):
         """Evaluate model
 
         Parameters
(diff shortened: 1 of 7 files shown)


adonath (commented):
Thanks a lot @AtreyeeS, the is very useful functionality. My main comment would be to establish the correct behavior here with unit tests.

registerrier:
A general comment here. 
What is the best approach here?
1. Add an energy evaluation on `TemporalModel` and a `is_energy_dependent` flag as done here?
2. Add a time evaluation on `SpectralModel` and a `is_time_dependent`flag?
I think option 1 adds complexity in the sense that you will still need a `SpectralModel` on top. 
Option 2 does not require an additional model. Also it could in the long run simplify significantly the handling of models by allowing products of `SpectralModel` and `TemporalModel`.

AtreyeeS:
Thinking about this, I tend to agree with @registerrier 
I started with (1) to be analogous with `SpatialModels` - you **always** need to multiply by a `SpectralModel` for the normalisation parameter (a `ConstantSpectralModel` at least). So the API could be consistent.

But probably it is more intuitive to think of Spectral models with some added temporal variability. I propose that I close this PR and start a new one? @adonath ?

adonath:
@registerrier As a long term solution I probably also agree, however given the convention we have now, adding the energy dependence to the temporal model seems a bit clearer to me:

```
model(E, t, l, b) = f(E) * g(l, b, E) * h(t, E)
```
Which defines the energy as the main required coordinate. Instead of:
```
model(E, t, l, b) = f(E, t) * g(l, b, E) * h(t)
```
Which basically removes the need for a spectral model. But it's not a strong argument in any case.

adonath:
What is the status here @AtreyeeS? Would you still like to add those for v1.0rc?

registerrier (changes requested):
Thanks @AtreyeeS it seems test_energy_dependent_model lacks a `requires_data()` decorator

gammapy/modeling/models/temporal.py
```diff
@@ -179,7 +195,7 @@ def evaluate(time):
         """Evaluate at given times."""
         return np.ones(time.shape)
 
-    def integral(self, t_min, t_max):
+    def integral(self, t_min, t_max, energy=None):
```
registerrier, line 198: Where is the energy integration performed?

AtreyeeS:
Closing this PR and opening a new one

Issue #2921, opened 2020-06-12
Fit mask overridden in FoVBackgroundMaker and FluxPointsEstimator


luca-giunti:
**Gammapy version**
dev

**Bug description**
See:
- https://github.com/gammapy/gammapy/blob/9df7a783fc050cf79667deb94d0cc635a2aa5bfb/gammapy/makers/background/fov.py#L49
- https://github.com/gammapy/gammapy/blob/9df7a783fc050cf79667deb94d0cc635a2aa5bfb/gammapy/estimators/flux_point.py#L906

The `dataset.mask_fit` if overridden during the computation of flux points or the field of view background normalization. This is bad, because users might want e.g. to:
- normalize the background in a custom energy range
- compute flux points by masking away some map regions

In those cases, they would set a `dataset.mask_fit` that is silently overridden in the current code. Notice that, for the case of background normalization in a custom energy range, this issue can be overcome by adapting the exclusion mask, instead of touching the `dataset.mask_fit`. But still, it is legitimate to want to do it with the `dataset.mask_fit` so it shouldn't be overridden.

**Other information**
I didn't check if this issue is also present somewhere else in the gammapy code, it might be worth having a look

adonath:
In #2991 the flux points estimation was refactored to use slicing instead of the masks. So this issue should not be present anymore for flux point estimation.

Remaining action item: adapt `FoVBackgroundMaker` to take into account the `mask_fit` information....

adonath:
The main question for me here is: what is the relation between the `exclusion_mask` parameter and the `mask_fit`.

adonath:
@luca-giunti I think the conclusion from the last dev call here was, to handle everything with the `exclusion_mask`. Is this correct?

adonath:
Then the issue is resolved....

luca-giunti:
> @luca-giunti I think the conclusion from the last dev call here was, to handle everything with the `exclusion_mask`. Is this correct?

@adonath @registerrier I don't remember exactly the reasons behind this conclusion, but in any case shouldn't it be better documented? I think it is not an obvious feature. A sentence like `Caveat: This Maker overrides internally the Dataset mask_fit. This means that all spatial and spectral exclusion ranges have to be encoded in the exclusion_mask.` would help

adonath:
[...]

PR #3841, opened 2022-03-08
Modified mask_fit handling in FoVBackgroundMaker

luca-giunti:
**Description**
This pull request addresses #2921: the `mask_fit` was silently overridden by the `FoVBackgroundMaker`. Now it is taken into account.

### DIFF
--- a/gammapy/makers/background/fov.py
+++ b/gammapy/makers/background/fov.py
@@ -18,7 +18,8 @@ class FoVBackgroundMaker(Maker):
     The dataset background model can be simply scaled (method="scale") or fitted (method="fit")
     on the dataset counts.
 
-    The normalization is performed outside the exclusion mask that is passed on init.
+    The normalization is performed outside the exclusion mask that is passed on init. This also internally 
+    takes into account the dataset fit mask.
 
     If a SkyModel is set on the input dataset its parameters
     are frozen during the fov re-normalization.
@@ -183,8 +184,6 @@ class FoVBackgroundMaker(Maker):
 
     def run(self, dataset, observation=None):
         """Run FoV background maker.
-
-        This internally takes into account both the exclusion mask and the dataset fit mask.
         
         Parameters
         ----------
--- a/gammapy/makers/background/fov.py
+++ b/gammapy/makers/background/fov.py
@@ -184,8 +184,8 @@ class FoVBackgroundMaker(Maker):
     def run(self, dataset, observation=None):
         """Run FoV background maker.
 
-        Fit the background model norm
-
+        This internally takes into account both the exclusion mask and the dataset fit mask.
+        
         Parameters
         ----------
         dataset : `~gammapy.datasets.MapDataset`
--- a/gammapy/makers/background/fov.py
+++ b/gammapy/makers/background/fov.py
@@ -193,8 +193,10 @@ class FoVBackgroundMaker(Maker):
 
         """
         mask_fit = dataset.mask_fit
-
-        dataset.mask_fit = self.make_exclusion_mask(dataset)
+        if mask_fit:
+            dataset.mask_fit *= self.make_exclusion_mask(dataset)
+        else:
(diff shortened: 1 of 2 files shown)


gammapy/makers/background/tests/test_fov.py
```diff
@@ -284,3 +284,18 @@ def test_fov_bkg_maker_scale_fail(obs_dataset, exclusion_mask, caplog):
...
+    region = CircleSkyRegion(obs_dataset._geom.center_skydir, Angle(0.4, "deg"))
+    test_dataset.mask_fit = obs_dataset._geom.region_mask(regions=[region])
+
+    dataset = fov_bkg_maker.run(test_dataset)
+
+    model = dataset.models[f"{dataset.name}-bkg"].spectral_model
+
+    assert_allclose(model.norm.value, 0.9975, rtol=1e-3)
+    assert_allclose(model.norm.error, 0.1115, rtol=1e-3)
+    assert_allclose(model.tilt.value, 0.0, rtol=1e-2)
```
registerrier, line 301: Maybe add a check that the mask fit is back to its initial value.

registerrier (approved):
Thanks @luca-giunti . This looks good

See inline comment.

gammapy/makers/background/tests/test_fov.py
```diff
@@ -284,3 +290,20 @@ def test_fov_bkg_maker_scale_fail(obs_dataset, exclusion_mask, caplog):
...
+@requires_data()
+def test_fov_bkg_maker_mask_fit_handling(obs_dataset, exclusion_mask):
+    fov_bkg_maker = FoVBackgroundMaker(method="scale", exclusion_mask=exclusion_mask)
+    test_dataset = obs_dataset.copy(name="test-fov")
+    region = CircleSkyRegion(obs_dataset._geom.center_skydir, Angle(0.4, "deg"))
+    mask_fit = obs_dataset._geom.region_mask(regions=[region])
+    test_dataset.mask_fit = mask_fit
+
+    dataset = fov_bkg_maker.run(test_dataset)
+    assert np.all((test_dataset.mask_fit == mask_fit).data) == True
```
registerrier, line 304: `assert_allclose` should better for this

adonath: I think for booleans there should be no uncertainty. However one should simplify the expression to:
```
assert np.all(test_dataset.mask_fit == mask_fit)
```
`Map` objects are interpreted as `np.ndarray` when put into a numpy function...

registerrier: Can you try this solution @luca-giunti ?

luca-giunti: @registerrier Sure, sorry for the delay

gammapy/makers/background/fov.py
```diff
@@ -193,8 +193,10 @@ def run(self, dataset, observation=None):
 
         """
         mask_fit = dataset.mask_fit
-
-        dataset.mask_fit = self.make_exclusion_mask(dataset)
+        if mask_fit:
+            dataset.mask_fit *= self.make_exclusion_mask(dataset)
+        else:
+            dataset.mask_fit = self.make_exclusion_mask(dataset)
```
adonath, line 199: I noticed this behavior is actually not documented. Could you add one or two sentences in the docstring of `FoVBackgroundMaker` that it modifies the `mask_fit` by combining it with the exclusion mask?

adonath, line 199: I noticed this behavior is actually not documented. Could you add one or two sentences in the docstring of `FoVBackgroundMaker` that it modifies the `mask_fit` by combining it with the exclusion mask?

adonath (commented):
Thanks @luca-giunti, I left one comment on documentation. Otherwise looks good to me...

luca-giunti:
@adonath done!

adonath (approved):
Thanks a lot @luca-giunti, no further comments from my side!

Issue #3765, opened 2022-01-28
Add 4FGL-DR3


AtreyeeS:
In continuation with discussions in #2055, the 4FGL-DR3 catalog has been released (https://fermi.gsfc.nasa.gov/ssc/data/access/lat/12yr_catalog/) and we might update `gammapy-data` and `gammapy.catalog`

PR #3842, opened 2022-03-08
Update SourceCatalog4FGL class to support changes in data release 3

QRemy:
This PR modifies the default input file in `SourceCatalog4FGL` to the latest 4FGL-DR3 catalog. The  `SuperExpCutoffPowerLaw4FGLSpectralModel` is adapted to match the DR3 definition (equation 2 and 3 of https://arxiv.org/pdf/2201.11184.pdf). The previous definition is now given by the `SuperExpCutoffPowerLaw4FGLDR1SpectralModel`.

Resolves #3840 and #3765.


QRemy:
@registerrier could you merge the PR in gammapy-data :  https://github.com/gammapy/gammapy-data/pull/24

registerrier:
> @registerrier could you merge the PR in gammapy-data : [gammapy/gammapy-data#24](https://github.com/gammapy/gammapy-data/pull/24)

Done. Sorry to have missed it.

gammapy/catalog/core.py
```diff
@@ -214,17 +214,28 @@ def _make_source_object(self, index):
...
+
+        if fp_energy_edges:
+            data["fp_energy_edges"] = fp_energy_edges
+        
         hist_table = getattr(self, "hist_table", None)
         hist2_table = getattr(self, "hist2_table", None)
 
+
         if hist_table:
-            data["time_axis"] = TimeMapAxis.from_table(hist_table, format="fermi-fgl")
```
adonath, line 221: Has the time info been removed from the latest data release?

adonath (commented):
Thanks @QRemy! For backwards compatibility I'd prefer to introduce `SuperExpCutoffPowerLaw4FGLDR3SpectralModel` instead and keep the existing model as is.

gammapy/catalog/core.py
QRemy: No but it may be safer to assume that it can be missing  
(for example the 1 year light-curve was add in rev 3 of DR1  see the table https://fermi.gsfc.nasa.gov/ssc/data/access/lat/8yr_catalog/)

registerrier (commented):

I think the docstring of the `SourceCatalog4FGL.__init__`  should mention that the DR versions depend on the `filename` argument, and by default this DR3. Then the class docstring might also mention DR3 rather than DR2.

I think we will need in some future PR to move the models linked to catalogs into some specific section of the model gallery to avoid overloading it.

adonath (approved):
Thanks @QRemy, no other comments from my side...

@registerrier ?

registerrier (approved):
Thanks @QRemy! This looks good to me.

Issue #3694, opened 2022-01-01
FluxEstimator._set_norm_parameter crashes


Tobychev:
**Gammapy version**
0.19

**Bug description**
Any code passing trough the "scan" branch of `FluxPointsEstimator._nan_result()` will crash due the invocation of `_set_norm_parameter` without the argument `scaled_parameter`.

**Expected behavior**
Either `_nan_result` passing the correct arguments, or `_set_norm_parameter` properly handling its default arguments.

registerrier:
Thanks @Tobychev.  The test for missing energy intervals currently does not execute scan. This needs to be added.

One issue with `FluxEstimator._set_norm_estimator` is that it shouldn't fail with its default argument `scaled_parameter=None`. A possible default behavior is to force `norm` range to be positive. This could provide a solution to #3690 as well in case no `scale_parameter` is unambiguously found.

PR #3845, opened 2022-03-09
Fix default arguments for flux_estimator._set_norm_parameter

luca-giunti:
**Description**
This pull request addresses https://github.com/gammapy/gammapy/issues/3694

### DIFF
--- a/gammapy/estimators/flux.py
+++ b/gammapy/estimators/flux.py
@@ -93,9 +93,10 @@ class FluxEstimator(ParameterEstimator):
             norm.max = scaled_parameter.max / scaled_parameter.value
             norm.interp = scaled_parameter.interp
         else:
-            norm.min = np.nan
-            norm.max = np.nan
+            norm.min = self.norm_min
+            norm.max = self.norm_max
             norm.interp = "log"
+
         norm.scan_values = self.norm_values
         norm.scan_min = self.norm_min
         norm.scan_max = self.norm_max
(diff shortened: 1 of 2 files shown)


QRemy:
if we redefine everything (value, frozen, min, max interp,  scan_values, scan_min, scan_max) in this function then norm does not have to be in the arguments

gammapy/estimators/flux.py
```diff
@@ -88,9 +88,15 @@ def _set_norm_parameter(self, norm=None, scaled_parameter=None):
...
-        norm.min = scaled_parameter.min / scaled_parameter.value
-        norm.max = scaled_parameter.max / scaled_parameter.value
-        norm.interp = scaled_parameter.interp
+        if scaled_parameter:
+            norm.min = scaled_parameter.min / scaled_parameter.value
+            norm.max = scaled_parameter.max / scaled_parameter.value
+            norm.interp = scaled_parameter.interp
+        else:
+            norm.min = self.norm_min
+            norm.max = self.norm_max
```
QRemy, line 97: the FluxEstimator defaults for norm_mina and norm_max are 0.2, and 5 respectively, so the constraint may be too strong if the reference model is not optimal then NaN default of the parameter may be better.

adonath:
Conclusion from dev meeting: for we require spectral models to either define a `norm` or `amplitude` parameter. To make it independent from the parameter name we could introduce `Parameter(..., is_amplitude=True)` as well.

adonath:
After #3860 one should modify https://github.com/gammapy/gammapy/blob/master/gammapy/estimators/flux.py#L116 to not check for the amplitude name, but for the `.is_norm` property instead. If it it not present it should fail... @luca-giunti Can you update the PR to do this?

adonath:
Superseded by #3905.

PR #3846, opened 2022-03-09
Allows DataStoreMaker  to be used with IRFs not following CALDB structure

QRemy:
DataStore.from_events_files is used to create HDU and observation index tables from the EVENTS header. But this works only for IRFs following the CALDB structure, and the events file header have to contains the TELESCOP, CALDB, and IRF keywords (not part of GADF).

 This PR improve the generation of the irf filename from the CALDB infos, and alternatively allow to read it directly from an extra keyword IRF_FILE in the events metadata.
Now, the events_sampling notebook show how to add this keyword, so the index files generated are not broken.

### DIFF
--- a/docs/tutorials/analysis/3D/event_sampling.ipynb
+++ b/docs/tutorials/analysis/3D/event_sampling.ipynb
@@ -748,6 +748,7 @@
     "%%time\n",
     "n_obs = len(tstarts)\n",
     "irf_paths = [Path(irf_filename)] * n_obs\n",
+    "events_paths = []\n",
     "for idx, tstart in enumerate(tstarts):\n",
     "    irf_path = load_cta_irfs(irf_paths[idx])\n",
     "    observation = Observation.create(\n",
@@ -761,12 +762,12 @@
     "\n",
     "    dataset = maker.run(empty, observation)\n",
     "    dataset.models = models\n",
-    "\n",
     "    sampler = MapDatasetEventSampler(random_state=idx)\n",
     "    events = sampler.run(dataset, observation)\n",
-    "    events.table.write(\n",
-    "        f\"./event_sampling/events_{idx:04d}.fits\", overwrite=True\n",
-    "    )"
+    "    \n",
+    "    path = Path(f\"./event_sampling/events_{idx:04d}.fits\")\n",
+    "    events_paths.append(path)\n",
+    "    events.table.write(path, overwrite=True)"
    ]
   },
   {
--- a/docs/tutorials/analysis/3D/event_sampling.ipynb
+++ b/docs/tutorials/analysis/3D/event_sampling.ipynb
@@ -787,7 +787,7 @@
    "source": [
     "path = Path(\"./event_sampling/\")\n",
     "events_paths = list(path.rglob(\"events*.fits\"))\n",
-    "data_store = DataStore.from_events_files(events_paths, irfs_paths)\n",
+    "data_store = DataStore.from_events_files(events_paths, irf_paths)\n",
     "data_store.obs_table"
    ]
   },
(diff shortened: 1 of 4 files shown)


adonath:
Related: https://github.com/open-gamma-ray-astro/gamma-astro-data-formats/issues/183

adonath:
Conclusion from the dev meeting:
- Adding new keywords requires changes and discussion in GADF (see link above)
- We can introduce `Observation.write()` to write events and IRF into one file
- We can introduce a helper function like `create_hdu_index_table(observations, irf_filenames={})` to create the HDU index table for the case where only one single set of IRFs is used

QRemy:
I modified the DataStoreMaker to accept an `irf_paths` argument,
 so now we can call `DataStore.from_events_file(events_paths, irf_paths=None)`
If` irfs_paths` is provided it must be the same length than `events_paths` (shown in the notebook).
If None the events file header have to contain CALDB and IRF keywords to locate the IRF file,
otherwise the IRFs are assumed to be contained in the events files.
This should cover all the use cases we have.

adonath:
@QRemy Maybe still support the case where `irf_path` is a single dict internally and document it? Just to make it even more convenient for users?

QRemy:
> @QRemy Maybe still support the case where `irf_path` is a single dict internally and document it? Just to make it even more convenient for users?

As the `irfs` arguement in `Observation.create` ?

adonath:
No, I meant something along the lines of:
```python3
def from_events_files(event_paths, irf_paths):
    """..."""
    if len(irf_paths) == 1:
        irf_paths = [irf_paths] * len(event_paths)
```

QRemy:
@adonath do you understand why the CALDB is not found even if defined in the failing tests ?

docs/tutorials/analysis/3D/event_sampling.ipynb
```diff
@@ -747,8 +746,11 @@
    "outputs": [],
    "source": [
     "%%time\n",
+    "n_obs = len(tstarts)\n",
+    "irf_paths = [Path(irf_filename)] * n_obs\n",
+    "events_paths = []\n",
     "for idx, tstart in enumerate(tstarts):\n",
-    "\n",
+    "    irf_path = load_cta_irfs(irf_paths[idx])\n",
```
registerrier, line 753: Is it a path or a `dict`? I would assume the later

docs/tutorials/analysis/3D/event_sampling.ipynb
```diff
@@ -747,8 +746,11 @@
    "outputs": [],
    "source": [
     "%%time\n",
+    "n_obs = len(tstarts)\n",
+    "irf_paths = [Path(irf_filename)] * n_obs\n",
```
registerrier, line 750: Is it useful here? It is only when reading that you need the path to the irf now, no?

docs/tutorials/analysis/3D/event_sampling.ipynb
```diff
@@ -784,8 +786,8 @@
    "outputs": [],
    "source": [
     "path = Path(\"./event_sampling/\")\n",
-    "paths = list(path.rglob(\"events*.fits\"))\n",
-    "data_store = DataStore.from_events_files(paths)\n",
+    "events_paths = list(path.rglob(\"events*.fits\"))\n",
+    "data_store = DataStore.from_events_files(events_paths, irf_paths)\n",
```
registerrier, line 790: This is probably where you could define the list of irf_path no?

registerrier (approved):
Thanks @QRemy .

I have left some inline comments in the notebook. 

@fabiopintore is this new behavior fine?

docs/tutorials/analysis/3D/event_sampling.ipynb
QRemy: In general observations can have different irf files so the list of irfs_path have to be defined before the loop. I wanted to stress that in the notebook, even if it is not necessary for this use case (but we could use 2 different zenith angles for example).

docs/tutorials/analysis/3D/event_sampling.ipynb
QRemy: a dict yes

fabiopintore:
It seems OK to me and no further comments from my side. Thanks @QRemy to have taken care of this!

registerrier (approved):

PR #3849, opened 2022-03-11
Clean up rad max data reduction code

adonath:
**Description**

This pull request tries to clean up a bit the data reduction code for the point like case. I noticed if we give up the "safety" to not check for overlaps in `make_counts_off_rad_max`, we can remove the code duplication of running the regions finder inside `make_counts_off_rad_max` and share more code with the "normal" data reduction in `ReflectedRegionsBackgroundMaker.make_counts`. There is no change in test values, except warning messages which I slightly adapted. I also moved `are_regions_overlapping_rad_max` to `makers/background/reflected.py`.


gammapy/makers/background/reflected.py
```diff
@@ -395,10 +417,28 @@ def __init__(
...
         else:
-            if len(kwargs) != 0:
-                raise ValueError('No kwargs can be given if providing a region_finder')
+            if kwargs:
+                raise ValueError("No kwargs can be given if providing a region_finder")
             self.region_finder = region_finder
 
+    @staticmethod
+    def _filter_regions_off_rad_max(regions_off, energy_axis, geom, events, rad_max):
+        # check for overlap
```
registerrier, line 426: Maybe add a docstring (even minimal) here.

adonath:
@registerrier This PR should be good now. Merge as soon as the CI is green?

registerrier (dismissed):
Thanks @adonath .

This seems good to me. I have left minor comments about docstring inline.

registerrier:
Maybe remove the unused import in utils.py for pyflakes CI to pass?

Issue #3811, opened 2022-02-17
`FluxPoints.plot()` fails if no error bars are present


AtreyeeS:
`FluxPoints.plot()` fails if no error bars are present on the data. According to the gaff (https://gamma-astro-data-formats.readthedocs.io/en/v0.2/spectra/flux_points/index.html), error columns are optional, so it should not complain.

adonath:
Could you maybe add minimal example that fails? This can be transformed into unit test later...Thanks!

PR #3850, opened 2022-03-14
Allow for flux points plotting without error measurements

registerrier:
**Description**

This pull request solves issue #3811. If no error column is present `FluxPoints` will be plotted without flux errors. 

**Dear reviewer**


adonath (approved):
Thanks @registerrier, no comments from my side.

PR #3852, opened 2022-03-15
Analysis class cleanup to support future extension of the hli

QRemy:
Add the remaining changes proposed in #3788 related to the cleanup of the analysis class :
- the methods .get_xx are replace by specific AnalysisStep classes and can be defined in the configuration file (for example config.general.steps = ["data-reduction", "fit"]), or called interactively by analysis.run(["data-reduction", "fit"])
- in the cli gammapy analysis run is now equivalent to analysis.run() so it can execute all the analysis step available and not only data reduction.


gammapy/analysis/steps.py
```diff
@@ -0,0 +1,498 @@
...
+from gammapy.utils.pbar import progress_bar
+from gammapy.utils.scripts import make_path, make_name
+
+
+class AnalysisStepBase(abc.ABC):
+    tag = "analysis-step"
+    requires_datasets = False
+    requires_models = False
+
+    def __init__(self, analysis, name=None, overwrite=True):
```
adonath, line 35: On a first look taking the analysis class on `__init__` does not make sense to me. This introduces a "cross dependency", while in fact I think it can be hierarchical in the sense that the `Analysis` class is built from `AnalysisStep` classes. It think this can be resolved by slightly refactoring the API of the `AnalysisStep` class, along the lines of:

```python3

class AnalysisStep:
    """Analysis step class"""
    tag = "analysis-step"
    def __init__(self, analysis_sub_config, overwrite=True, log=None):
        self.config = analysis_sub_config
        self.overwrite = overwrite
        
        if log is None:
        
        self.log = log
    
    @property
    def maker_config(self):
        # translate analysis sub config to Gammapy API config here
        return config
    
    def run(self, datasets, models=None):
        maker = Maker(**self.maker_config)
        # returning might be optional...changes could happen in place
        return datasets
    
    


```

gammapy/analysis/steps.py
```diff
@@ -0,0 +1,498 @@
...
+        self.analysis = analysis
+        self.overwrite = overwrite
+        self._name = make_name(name)
```
adonath, line 38: What is the name attribute used for? Is to generate the dataset later? But what happened for multiple datasets?

adonath, line 38: What is the name attribute used for? Is to generate the dataset later? But what happened for multiple datasets?

gammapy/analysis/steps.py
```diff
@@ -0,0 +1,498 @@
...
+                lon=obs_settings.obs_cone.lon,
+                lat=obs_settings.obs_cone.lat,
+                radius=obs_settings.obs_cone.radius,
+                border="0 deg",
+            )
+            selected_obs_table = selected_obs_table.select_observations(cone)
+
+        return selected_obs_table["OBS_ID"].tolist()
+    
+class DatasetsAnalysisStep(AnalysisStepBase):
```
adonath, line 152: This is currently a very large class with lots of responsibility, maybe it could be split further...

adonath, line 152: This is currently a very large class with lots of responsibility, maybe it could be split further...

gammapy/analysis/steps.py
```diff
@@ -0,0 +1,498 @@
...
+            self.analysis.log.debug(dataset)
+            datasets.append(dataset)
+        self.analysis.datasets = Datasets(datasets)
+
+        if datasets_settings.stack:
+            stacked = self.analysis.datasets.stack_reduce(name="stacked")
+            self.analysis.datasets = Datasets([stacked])
+
+
+def make_energy_axis(axis, name="energy"):
```
adonath, line 361: I haven't thought this through, but we could maybe even introduce and API like, `MapAxis.from_analysis_config(config=)` and `Maker.from_analysis_config()`...

adonath, line 361: I haven't thought this through, but we could maybe even introduce and API like, `MapAxis.from_analysis_config(config=)` and `Maker.from_analysis_config()`...

adonath (commented):
Thanks a lot @QRemy, I think this goes exactly in the right direction!  Especially I like the improvement in code organization and possibility to extend the `Analysis` pipeline with a registry.

I have left a few general comments with proposals for the API and some remaining questions for now. I think it might be good to discuss this in a bit more detail again, maybe in a dedicated meeting next week?

gammapy/analysis/config.py
```diff
@@ -55,6 +55,16 @@ def validate(cls, v):
         return Time(v)
 
 
+class AnalysisStepEnum(str, Enum):
+    data_reduction = "data-reduction"
+    observations = "observations"
```
registerrier, line 60: How about calling the step `data-selection`? This might be more explicit?

gammapy/analysis/core.py
```diff
@@ -80,90 +63,48 @@ def config(self, value):
...
-        return selected_obs_table["OBS_ID"].tolist()
-
+            if overwrite is None:
+                overwrite = [True]*len(steps)
+        for k,step in enumerate(steps):
+            if isinstance(overwrite, list):
+                overwrite_step = overwrite[k]
+            else :
+                overwrite_step = overwrite
+            anlysis_step = AnalysisStep.create(step,
```
registerrier, line 78: analysis

gammapy/analysis/core.py
```diff
@@ -80,90 +63,48 @@ def config(self, value):
...
-        # First select input list of observations from obs_table
-        if len(obs_settings.obs_ids):
-            selected_obs_table = self.datastore.obs_table.select_obs_id(
-                obs_settings.obs_ids
-            )
-        elif obs_settings.obs_file is not None:
-            path = make_path(obs_settings.obs_file)
-            ids = list(Table.read(path, format="ascii", data_start=0).columns[0])
-            selected_obs_table = self.datastore.obs_table.select_obs_id(ids)
+    def run(self, steps=None, overwrite=None, **kwargs):
```
registerrier, line 66: add doctring

registerrier: what does the `overwrite` option do?

gammapy/analysis/core.py
```diff
@@ -80,90 +63,48 @@ def config(self, value):
...
+        if steps is None:
```
registerrier, line 67: Maybe steps creation should be done in another method?

gammapy/analysis/core.py
```diff
@@ -274,320 +215,24 @@ def write_datasets(self, overwrite=True, write_covariance=True):
...
+            self.log.info(f"Datasets stored to {filename}.")
+            self.log.info(f"Datasets stored to {filename_models}.")
         else:
             raise RuntimeError("Missing datasets_file in config.general")
 
 
-    def run_fit(self):
-        """Fitting reduced datasets to model."""
-        if not self.models:
+    def check_datasets(self):
```
registerrier, line 224: is dataset reading and/or model creation and setting a specific analysis step?

gammapy/analysis/steps.py
registerrier: Maybe a minima, you could take the `AnalysisConfig` on init and pass the `Analysis` to run().

gammapy/analysis/steps.py
```diff
@@ -0,0 +1,498 @@
...
+    def __init__(self, analysis, name=None, overwrite=True):
+        self.analysis = analysis
+        self.overwrite = overwrite
+        self._name = make_name(name)
+
+    @property
+    def name(self):
+        return self._name
+    
+    def run(self):
```
registerrier, line 44: make `Analysis` an argument?

gammapy/analysis/steps.py
```diff
@@ -0,0 +1,498 @@
...
+            DatasetsAnalysisStep(self.analysis, self.overwrite).get_datasets()
+            self.analysis.write_datasets()
+
+class ObservationsAnalysisStep(AnalysisStepBase):
+    tag = "observations"
+
+    def _run(self):
+        """Fetch observations from the data store according to criteria defined in the configuration."""
+        observations_settings = self.analysis.config.observations
+        self._set_data_store()
```
registerrier, line 83: there it should get the datastore and set it on `analysis`.

gammapy/analysis/steps.py
registerrier: Yes possibly there could be a step responsible for the geometry and dataset creation and one for handling makers and filling them

gammapy/analysis/steps.py
```diff
@@ -0,0 +1,498 @@
...
+            time_intervals = None
+        else:
+            time_intervals = [
+                (t1, t2)
+                for t1, t2 in zip(
+                    lc_settings.time_intervals.start, lc_settings.time_intervals.stop
+                )
+            ]
+
+        self.analysis.check_datasets()
```
registerrier, line 482: why is it needed explicitly here?

registerrier (commented):
Thanks @QRemy. This is a very ambitious change but it is definitely very interesting.

See some inline comments.
To make progress possible and have an intermediate working solution, it might be possible to pass the `Analysis` object to each step rather than passing it on init.
Would that be OK @adonath ?

gammapy/analysis/core.py
QRemy: I had in mind that each AnalysisStep could have a read/write method but for now this is used only in the DataReductionAnalysisStep to read the datasets if exits.

gammapy/analysis/core.py
QRemy: It means they will have to be kept in memory and attached to the analysis class which affects where the information is stored. I will try this later after sorting out the input/output of the analysis steps.

gammapy/analysis/steps.py
QRemy: removed it for now but I will reintroduce a similar system to check if the data required for the each step are well defined

gammapy/analysis/steps.py
QRemy: for now changed to take `AnalysisConfig` on init and pass the `Analysis` to run() but in the next PR I will introduce a specific `AnalysisProducts` container to return outputs and pass data references on run().

gammapy/analysis/steps.py
QRemy: for now the analysis steps mirror the get_xx methods, I could try to further split theses classes in separate PR (one for each)

QRemy: for now the analysis steps mirror the get_xx methods, I could try to further split theses classes in separate PR (one for each)

gammapy/analysis/core.py
QRemy: no but it could be, for now reading of config.general.datasets_file is done through the data-selection step if the file exists, and if overwrite is False.

registerrier:
This PR implements the basis of a large refactoring of the HLI. A number of design choices will have to be made and a full expanded and improved HLI is an objective for v2.0.
For now, even though the changes proposed in this PR do not break the API, we decided during the co-working week to postpone to v1.1 and implement the HLI in parallel with the existing one.

QRemy:
> For now, even though the changes proposed in this PR do not break the API, we decided during the co-working week to postpone to v1.1 and implement the HLI in parallel with the existing one.

I moved the code to a new submodule called `workflow` (that won't show in the doc for now).
Some tests are still failing but I can't reproduce the error locally any idea @registerrier @adonath ?

QRemy:
Now all tests are good, maybe we wait to merge this only after 1.3 ?

QRemy:
Merging now, will continue to improve the workflow module in separate PR.

PR #3854, opened 2022-03-16
Implement Observation.write

maxnoe:
**Description**

This pull request implements the `Observation.write` method as discussed in last week's developer call


gammapy/data/observations.py
```diff
@@ -428,6 +431,37 @@ def read(cls, event_file, irf_file=None):
...
+        events = self.events
+        if events is not None:
+            hdul.append(events.to_table_hdu(format=format))
+
+        gti = self.gti
+        if gti is not None:
+            hdul.append(gti.to_table_hdu(format=format))
+
+        if include_irfs:
+            for irf_name in ['aeff', 'edisp', 'psf', 'bkg', 'rad_max']:
```
adonath, line 458: To avoid the hard coded irf names you can use `Observation.available_irfs` here...

gammapy/data/observations.py
```diff
@@ -428,6 +431,37 @@ def read(cls, event_file, irf_file=None):
             **irf_dict,
         )
 
+    def write(self, path, overwrite=False, format='gadf-dl3', include_irfs=True):
+        """
+        Write this observation into `path` using the specified format
+        """
```
adonath, line 437: Please add Parameter section to docstring

gammapy/data/gti.py
```diff
@@ -107,6 +107,12 @@ def read(cls, filename, hdu="GTI"):
         table = Table.read(filename, hdu=hdu)
         return cls(table)
 
+    def to_table_hdu(self, name="GTI", format='gadf-dl3'):
+        '''
+        Convert this GTI instance to a `astropy.io.fits.BinTableHDU`.
```
adonath, line 112: Please add Parameter section to docstring.

gammapy/data/gti.py
```diff
@@ -107,6 +107,12 @@ def read(cls, filename, hdu="GTI"):
         table = Table.read(filename, hdu=hdu)
         return cls(table)
 
+    def to_table_hdu(self, name="GTI", format='gadf-dl3'):
```
adonath, line 110: In other places we use `hdu_name` maybe change for consistency?

gammapy/data/gti.py
```diff
@@ -107,6 +107,12 @@ def read(cls, filename, hdu="GTI"):
...
+        '''
+        Convert this GTI instance to a `astropy.io.fits.BinTableHDU`.
+        '''
+        return fits.BinTableHDU(self.table, name="GTI")
```
adonath, line 114: This still has the hard-coded HDU name...

gammapy/data/event_list.py
```diff
@@ -102,7 +102,13 @@ def read(cls, filename, **kwargs):
         table = Table.read(filename, **kwargs)
         return cls(table=table)
 
-    def write(self, filename, gti=None, overwrite=False, format='gadf'):
+    def to_table_hdu(self, name="EVENTS", format='gadf-dl3'):
+        '''
+        Convert event list to a `~astropy.io.fits.BinTableHDU`
```
adonath, line 107: Please add Parameter section to docs and add a check for the value of `format`...

adonath (changes requested):
Thanks @maxnoe, this is very convenient functionality. I left a few first inline comments for now...

gammapy/data/observations.py
maxnoe: That does not include `rad_max` at the moment. Is that just a place where we forgot to add it, or is there a specific reason?

gammapy/data/gti.py
maxnoe: I cannot find any method with a `hdu_name` argument in gammapy. All the `to_table_hdu` methods now have `name=` or do not support giving a name (should be changed?):

```
❯ rg 'def to_table_hdu'
gammapy/data/gti.py
110:    def to_table_hdu(self, name="GTI", format='gadf-dl3'):

gammapy/data/event_list.py
105:    def to_table_hdu(self, name="EVENTS", format='gadf-dl3'):

gammapy/maps/axes.py
1096:    def to_table_hdu(self, format="ogip"):
1803:    def to_table_hdu(self, format="gadf", hdu_bands=None):

gammapy/irf/core.py
488:    def to_table_hdu(self, format="gadf-dl3"):
```

maxnoe: Also, we should probably replace `gadf-dl3` with `gadf`, right? `DL3` is not needed, since it is clear from what is being stored.

gammapy/data/observations.py
adonath: No, I think we just forgot it. It is the single places where the availability of irfs is checked without accessing the files on disk.

maxnoe: ok, added `"rad_max"`

gammapy/data/gti.py
adonath: Yes, you are right the argument e.g., used in `gammapy.maps` is just called `hdu` not `hdu_name` (that was probably wishful thinking of a better argument name...).

Concerning the `gadf` vs. `gadf-dl3`. There is one place where there is a distinction between these formats. A `MapAxis` can be serialized to a table both in `gadf` (DL4) and `gadf-dl3` (DL3) format However this distinction is only needed on the lowest level of `MapAxis` (which users basically never use independently...). It is still convenient to just forward the format specification, but I agree that on the high level of `IRF` and `Map` such is distinction is not needed...

adonath: Conclusion from the dev meeting: only specify `format="gadf"` on the high level classes, such as `Observation`, `Map`, `IRF` etc. and then internally for `MapAxis` do the correct translation. Long-term the code on `MapAxis` will move to `Reader` and `Writer` classes.

gammapy/data/event_list.py
```diff
@@ -102,7 +102,23 @@ def read(cls, filename, **kwargs):
         table = Table.read(filename, **kwargs)
         return cls(table=table)
 
-    def write(self, filename, gti=None, overwrite=False, format='gadf'):
+    def to_table_hdu(self, name="EVENTS", format='gadf-dl3'):
```
adonath, line 105: Change to `format="gadf"` here...and maybe remove the possibility to change the extname completely? While I think the name is not a GADF specification it might be good be strict on write and use consistent names when writing with Gammapy. If users really want they can change the EXTNAME later on the `BinTableHDU`...

gammapy/data/gti.py
adonath: On second thought: I think we should remove the `name` argument completely. While there is no specification on the extname, we should be strict and just write consistent names from Gammapy to avoid a proliferation of HDU names in files written by Gammapy...if users really want they can always change the name later based on `BinTableHDU`...

adonath: On second thought: I think we should remove the `name` argument completely. While there is no specification on the extname, we should be strict and just write consistent names from Gammapy to avoid a proliferation of HDU names in files written by Gammapy...if users really want they can always change the name later based on `BinTableHDU`...

adonath (changes requested):
Thanks @maxnoe, I have left a few more inline comments.

maxnoe:
I removed the `name` argument and changed to `format="gadf"` for the touched methods. IRFs still use `format="gadf-dl3"`

adonath (approved):
Thanks a lot @maxnoe! I have no further comments...

gammapy/data/gti.py
maxnoe: done

maxnoe:
Tests should be passing now

adonath (approved):
Thanks @maxnoe, no further comments from my side...

registerrier (approved):
Thanks @maxnoe . This looks good!

Issue #3790, opened 2022-02-09
Make ObservationTable optional on DataStore


adonath:
Technically the `DataStore` should be able to work without the `ObservationTable`. All required information is present in the event list header. So we should add one unit test where the `obs_table` is actually optional and make sure that the Gammapy code handles this case correctly. This will help to improve the separation of responsibilities in the code.

maxnoe:
Related to #3742 and #3767 

Currently, gammapy only reads the information from the obs table when using datastore, never looking at the individual observations.

PR #3856, opened 2022-03-16
Make obs_table optional on DataStore

registerrier:
**Description**

This pull request makes the `ObservationTable` an optional argument in `DataStore`. This solves #3790 .

**Dear reviewer**

### DIFF
--- a/gammapy/data/data_store.py
+++ b/gammapy/data/data_store.py
@@ -96,7 +96,7 @@ class DataStore:
 
         hdu_table = HDUIndexTable.read(filename, hdu=hdu_hdu, format="fits")
 
-        obs_table=None
+        obs_table = None
         if hdu_obs:
             obs_table = ObservationTable.read(filename, hdu=hdu_obs, format="fits")
 
@@ -155,7 +155,7 @@ class DataStore:
         hdu_table.meta["BASE_DIR"] = str(base_dir)
 
         if not obs_table_filename.exists():
-            log.warning("Cannot find default obs-index table.")
+            log.info("Cannot find default obs-index table.")
             obs_table = None
         else:
             log.debug(f"Reading {obs_table_filename}")
--- a/gammapy/data/data_store.py
+++ b/gammapy/data/data_store.py
@@ -92,6 +92,7 @@ class DataStore:
 
         hdu_table = HDUIndexTable.read(filename, hdu=hdu_hdu, format="fits")
 
+        obs_table=None
         if hdu_obs:
             obs_table = ObservationTable.read(filename, hdu=hdu_obs, format="fits")
 
--- a/gammapy/data/data_store.py
+++ b/gammapy/data/data_store.py
@@ -152,6 +152,7 @@ class DataStore:
             log.warning(f"Cannot find observation index file : {obs_table_filename}.")
             obs_table = None
         else:
+            log.debug(f"Reading {obs_table_filename}")
             obs_table = ObservationTable.read(obs_table_filename, format="fits")
 
         return cls(hdu_table=hdu_table, obs_table=obs_table)
@@ -242,16 +243,6 @@ class DataStore:
             raise ValueError(f"OBS_ID = {obs_id} not in HDU index table.")
 
         kwargs = {"obs_id": int(obs_id)}
-
-        if self.obs_table and obs_id in self.obs_table["OBS_ID"]:
-            row = self.obs_table.select_obs_id(obs_id=obs_id)[0]
-
-            # add info from table meta, e.g. the time references
-            kwargs["obs_info"] = {
-                k: v for k, v in self.obs_table.meta.items()
-                if not k.startswith('HDU')  # Ignore GADF structure of index table
-            }
-            kwargs["obs_info"].update(table_row_to_dict(row))
 
         hdu_list = ["events", "gti", "aeff", "edisp", "psf", "bkg", "rad_max"]
 
--- a/gammapy/data/data_store.py
+++ b/gammapy/data/data_store.py
@@ -349,7 +349,8 @@ class DataStore:
             subhdutable.add_index("HDU_CLASS")
             with subhdutable.index_mode("discard_on_copy"):
                 subhdutable = subhdutable.loc[hdu_class]
-        subobstable = self.obs_table.select_obs_id(obs_id)
+        if self.obs_table:
+            subobstable = self.obs_table.select_obs_id(obs_id)
 
         for idx in range(len(subhdutable)):
             # Changes to the file structure could be made here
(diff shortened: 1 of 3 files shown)


adonath (commented):
Thanks @registerrier, I have left two inline comments.

gammapy/data/data_store.py
```diff
@@ -146,9 +149,10 @@ def from_dir(cls, base_dir, hdu_table_filename=None, obs_table_filename=None):
         hdu_table.meta["BASE_DIR"] = str(base_dir)
 
         if not obs_table_filename.exists():
-            raise OSError(f"File not found: {obs_table_filename}")
-        log.debug(f"Reading {obs_table_filename}")
```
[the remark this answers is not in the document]
registerrier: Indeed. Done.

gammapy/data/data_store.py
```diff
@@ -231,21 +238,20 @@ def obs(self, obs_id):
...
-        row = self.obs_table.select_obs_id(obs_id=obs_id)[0]
         kwargs = {"obs_id": int(obs_id)}
 
-        # add info from table meta, e.g. the time references
-        kwargs["obs_info"] = {
-            k: v for k, v in self.obs_table.meta.items()
-            if not k.startswith('HDU')  # Ignore GADF structure of index table
-        }
-        kwargs["obs_info"].update(table_row_to_dict(row))
+        if self.obs_table and obs_id in self.obs_table["OBS_ID"]:
```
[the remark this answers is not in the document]
registerrier: I have introduced a possible approach defining a `lazyproperty` on `Observation`. 
The implementation might be a bit unorthodox though.

registerrier:
Note that if we remove the `obs_info` setting from the `DataStore.obs`and rely instead on the event file,  we might loose the ability to use a `DataStore` to perform observation simulations.

registerrier:
> Good point that I haven't thought of yet. So the logic would be to create an observation table for simulations first? Or is this all done based on in-memory `Observation` objects? This is not clear to me yet...I'll think about it once more and comment again.

I think this is an open possibility. An `ObservationsEventSampler` class such as the one @QRemy proposes in #3711 could have a method building an `Observations` container from  a `DataStore`. 

Probably such an approach might require a `POINTING` extension to be present in the HDU index. (but for some reason `POINTING` does not appear in the list of valid `HDU_TYPE` for the index in gadf v0.2). And we would need to support the pointing extension. 

Thinking about it, this is not sufficient since you still need to provide some obs_info e.g. dead time.

gammapy/data/data_store.py
```diff
@@ -292,7 +289,7 @@ def get_observations(self, obs_id=None, skip_missing=False, required_irf="all"):
             )
 
         if obs_id is None:
-            obs_id = self.obs_table["OBS_ID"].data
+            obs_id = np.unique(self.hdu_table["OBS_ID"].data)
```
maxnoe, line 292: Make this an 

```
@property
def obs_ids(self):
     return np.unique(self.hdu_table["OBS_ID"].data)
```

?

Depending on whether you consider the data store mutable or not, this could also be a `lazyproperty` then.

gammapy/data/observations.py
```diff
@@ -259,6 +259,14 @@ def observation_dead_time_fraction(self):
         """
         return 1 - self.obs_info["DEADC"]
 
+    @lazyproperty
+    def obs_info(self):
+        """Observation info dictionary."""
```
maxnoe, line 264: This makes the lines further down obsolete and takes a different approach: only using the events header if it is present completely ignoring any `obs_info`. 

The code below *merges* obs_info and events table, giving preference to the events header.

We should decide what to do here and then use `self.obs_info` in `fixed_pointing_info`.

gammapy/data/tests/test_data_store.py
```diff
@@ -227,3 +227,22 @@ def test_datastore_header_info_in_obs_info(data_store):
     assert "GEOLAT" in obs.obs_info
     # make sure we don't add the OBS_INDEX HDUCLAS
     assert "HDUCLAS1" not in obs.obs_info
+
+@requires_data()
+def test_datastore_from_dir_no_obs_index(caplog):
+    """Test the `from_dir` method."""
+    data_store = DataStore.from_dir(
+        "$GAMMAPY_DATA/hess-dl3-dr1/", "hdu-index.fits.gz", "bad_name"
```
maxnoe, line 235: I think when users provide explicitly a filename that does not exist, it should raise, not warn.

registerrier: You are right. The correct behavior should be:
- if `obs_index_filename = None`  log a warning if the default obs-index is not there
- raise a warning otherwise.

QRemy:
> I think this is an open possibility. An `ObservationsEventSampler` class such as the one @QRemy proposes in #3711 could have a method building an `Observations` container from a `DataStore`.

But how do you create the Datastore ? For simulations I think it is simpler to create an Observations list from an human-readable configuration file containing also optional arguments for the `ObservationsEventSampler` then the Datastore is create from the events files and hdu-index saved to disk.

adonath:
I agree @QRemy, this sounds like a reasonable approach.

gammapy/data/data_store.py
```diff
@@ -91,7 +96,9 @@ def from_file(cls, filename, hdu_hdu="HDU_INDEX", hdu_obs="OBS_INDEX"):
 
         hdu_table = HDUIndexTable.read(filename, hdu=hdu_hdu, format="fits")
 
-        obs_table = ObservationTable.read(filename, hdu=hdu_obs, format="fits")
+        obs_table=None
```
maxnoe, line 99: Missing spaces

gammapy/data/data_store.py
```diff
@@ -146,9 +155,11 @@ def from_dir(cls, base_dir, hdu_table_filename=None, obs_table_filename=None):
         hdu_table.meta["BASE_DIR"] = str(base_dir)
 
         if not obs_table_filename.exists():
-            raise OSError(f"File not found: {obs_table_filename}")
-        log.debug(f"Reading {obs_table_filename}")
-        obs_table = ObservationTable.read(obs_table_filename, format="fits")
+            log.warning("Cannot find default obs-index table.")
```
maxnoe, line 158: Should this really be a warning if we say it's optional?

adonath:
@registerrier Please address remaining comments by @maxnoe, thanks!

gammapy/data/data_store.py
registerrier: Done

gammapy/data/data_store.py
registerrier: Done

PR #3860, opened 2022-03-22
Introduce Parameter.is_norm property

adonath:
**Description**

This pull request introduces a `"norm"` parameter type, to distinguish the norm type parameters such as `"norm"` and  `"amplitude"`. The PR also addressed implicitly #3845.

The behavior is now, that on init of `SkyModel` the existence of a norm type parameter will be checked and additionally checked that there is only one(!) such a parameter free for fitting. This prevents failing fits because of correlated amplitude values.

### DIFF
--- a/gammapy/modeling/models/cube.py
+++ b/gammapy/modeling/models/cube.py
@@ -73,7 +73,6 @@ class SkyModel(ModelBase):
         is_norm = np.array([par.is_norm for par in spectral_model.parameters])
 
         if not np.any(is_norm):
-            print(spectral_model)
             raise ValueError("Spectral model used with SkyModel requires a norm type parameter.")
 
         is_free_norm = np.array([par.is_norm for par in spectral_model.parameters.free_parameters])
--- a/gammapy/modeling/models/cube.py
+++ b/gammapy/modeling/models/cube.py
@@ -69,6 +69,18 @@ class SkyModel(ModelBase):
         self.apply_irf = apply_irf
         self.datasets_names = datasets_names
         self._check_unit()
+
+        is_norm = np.array([par.is_norm for par in spectral_model.parameters])
+
+        if not np.any(is_norm):
+            print(spectral_model)
+            raise ValueError("Spectral model used with SkyModel requires a norm type parameter.")
+
+        is_free_norm = np.array([par.is_norm for par in spectral_model.parameters.free_parameters])
+
+        if np.sum(is_free_norm) > 1:
+            raise ValueError("Spectral model used with SkyModel can only have a single free norm type parameter.")
+
         super().__init__()
 
     @property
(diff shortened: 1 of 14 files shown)


gammapy/modeling/parameter.py
```diff
@@ -14,6 +16,19 @@
 log = logging.getLogger(__name__)
 
 
+class ParameterTypes(Enum):
```
registerrier, line 19: Assigning the model categories to parameters seems a bad idea to me. It adds complexity but what does it solve?

registerrier (changes requested):
Thanks @adonath .

Adding a flag to determine which is the scalable parameter is useful. Yet I don't see the point of creating `Parameter` types that mimic our model classes. 
It will make user model addition more complex, it will add complexity in many other places.

Also using `norm` as a type is not very clear, as e.g. `amplitude` can be of type `norm`, but the `norm` parameter of another component will not.

gammapy/modeling/parameter.py
adonath: It seemed the more general solution to me. Possibly we get more parameter types in future (could imagine e.g., distinguishing position from morphology parameters). And already now a `SpatialModel` could in fact declare "spectral" parameters e.g. to handle for energy dependent morphology, such as a spatially smoothly varying spectral index. But I see your point, given the current convention we have!

I guess the alternative and even simpler solution here is to just work with the `.is_norm` flag for now? And declare this on init of the parameter?

adonath: I guess it mostly depends on the definition we employ for the `Parameter.type` attribute. One could even introduce `Parameter.model_type` and distinguish both cases...

gammapy/modeling/models/cube.py
```diff
@@ -69,6 +69,18 @@ def __init__(
         self.apply_irf = apply_irf
         self.datasets_names = datasets_names
         self._check_unit()
+
+        is_norm = np.array([par.is_norm for par in spectral_model.parameters])
+
+        if not np.any(is_norm):
+            print(spectral_model)
```
adonath, line 76: Remove print statement....

registerrier (commented):
Thanks for the changes @adonath . 
Just to be sure, the main API change now is that one spectral model parameter at most have `is_norm=True`.  By default all regular `SpectralModel` amplitude parameters do. But the `norm` parameter of `NormSpectralModel` doesn't.
Does it mean that when using a a 3D `TemplateSpatialModel` the user has to change the default `norm` parameter?

adonath:
> Just to be sure, the main API change now is that one spectral model parameter at most have is_norm=True.

Yes

> By default all regular SpectralModel amplitude parameters do. But the norm parameter of NormSpectralModel doesn't.

The norm spectral models define one parameter with `is_norm` as well. This is different from `SpectralModel.is_norm_spectral_model`. I'm happy to implement any better proposal for the parameter property...

> Does it mean that when using a a 3D TemplateSpatialModel the user has to change the default norm parameter?

No, the `TemplateSpatialModel` is used with a norm spectral model in any case. There is no API change related to this...

registerrier (approved):
Thanks @adonath ! This looks good to me.

I suppose one additional change needed will to require that only one `is_norm` parameter be present on the model for any `FluxEstimator` in order to solve the issue of parameter to be scaled. This can be done in another PR.

PR #3861, opened 2022-03-24
Fix interp_to_geom to use correctly fill_value

QRemy:
Fix interp_to_geom to use correctly fill_value
(fill_value was defined outside kwargs and not passed to interp method)

### DIFF
--- a/gammapy/maps/core.py
+++ b/gammapy/maps/core.py
@@ -887,7 +887,7 @@ class Map(abc.ABC):
             data = map_copy.get_by_coord(coords)
             data = np.nan_to_num(data, nan=fill_value).astype(bool)
         else:
-            data = map_copy.interp_by_coord(coords, **kwargs)
+            data = map_copy.interp_by_coord(coords, fill_value=fill_value, **kwargs)
 
         if preserve_counts:
             data *= geom.solid_angle().to_value("deg2")
--- a/gammapy/maps/hpx/ndmap.py
+++ b/gammapy/maps/hpx/ndmap.py
@@ -338,7 +338,7 @@ class HpxNDMap(HpxMap):
         else:
             return self.copy()
 
-    def interp_by_coord(self, coords, method="linear"):
+    def interp_by_coord(self, coords, method="linear", fill_value=None):
         # inherited docstring
         coords = MapCoord.create(coords, frame=self.geom.frame)
 
@@ -349,7 +349,7 @@ class HpxNDMap(HpxMap):
         else:
             raise ValueError(f"Invalid interpolation method: {method!r}")
 
-    def interp_by_pix(self, pix, method=None):
+    def interp_by_pix(self, pix, method=None, fill_value=None):
         """Interpolate map values at the given pixel coordinates."""
         raise NotImplementedError
 
--- a/gammapy/maps/tests/test_core.py
+++ b/gammapy/maps/tests/test_core.py
@@ -447,25 +447,28 @@ def test_interp_to_geom():
     assert interp_hpx_map.geom == hpx_geom_target
 
     # Preserving the counts
+    binsz = 0.2 * u.deg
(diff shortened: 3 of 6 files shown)


QRemy:
Now that when `fill_value=0.` is taken into account properly by `interp_to_geom`  the extrapolation beyond the last bin center is disabled by default.  So if the base map resolution is coarser than the target geom  coarser we loose information between the centers and edges of the last base bins, as shown below :

image

This will also affect the `TemplateSpatialModel` evaluation that has the same default behaviour. 

Extrapolation should probably be allowed up to the geom edges in `interp_by_coord`, similarly to what is done in IRF: https://github.com/gammapy/gammapy/blob/35f4ae77ffd70c475dc8e263b7ad9cfdea74e2ea/gammapy/irf/core.py#L277-L284

adonath:
Thanks @QRemy, first quick comment: please add a unit test.

QRemy:
I adapted the test checking that counts are roughly preserved such as the target geom is larger so this test fails in case of extrapolation.

gammapy/maps/core.py
```diff
@@ -887,8 +887,16 @@ def interp_to_geom(self, geom, preserve_counts=False, fill_value=0, **kwargs):
             data = map_copy.get_by_coord(coords)
             data = np.nan_to_num(data, nan=fill_value).astype(bool)
         else:
-            data = map_copy.interp_by_coord(coords, **kwargs)
-
+            data = map_copy.interp_by_coord(coords, fill_value=None, **kwargs)
+            if fill_value is not None:
```
adonath, line 891: This does not seem to be the right place for this fix to me. Because `.interp_by_coord()` and `.interp_by_pix()` will remain unaffected, even if the both define a `fill_value`. I think the better place is likely https://github.com/gammapy/gammapy/blob/master/gammapy/maps/wcs/ndmap.py#L139

gammapy/maps/core.py
```diff
@@ -887,8 +887,16 @@ def interp_to_geom(self, geom, preserve_counts=False, fill_value=0, **kwargs):
             data = map_copy.get_by_coord(coords)
             data = np.nan_to_num(data, nan=fill_value).astype(bool)
         else:
-            data = map_copy.interp_by_coord(coords, **kwargs)
-
+            data = map_copy.interp_by_coord(coords, fill_value=None, **kwargs)
+            if fill_value is not None: 
+                 idxs = self.geom.coord_to_idx(coords, clip=False) 
+                 invalid = np.broadcast_arrays(*[idx == -1 for idx in idxs])
```
adonath, line 893: I think it is a good idea to introduce something like a `WcsGeom.footprint_mask` (lazy) attribute. I remember we have many of these checks, like `[idx == -1 for idx in idxs]`.It's cleaner and more efficient to just compute it once in a single place. Because it I snot going to change for a given `WcsGeom`.

QRemy: but here `idxs` depends on the `coords` of the target geom here so we have to recompute this

adonath: But you can still extract the value of the pixels of the target geom like `self.geom.footprint_mask.get_by_coord()`?

QRemy: `get_by_coord ` will call `contains` so it would be simpler to do direclty:
`invalid = ~self.geom.contains(coords)`

adonath: `contains` should use `geom.footprint_mask` as well. Let's do it like this: I'll introduce `footprint_mask` in a separate PR myself. There are easily ~5 more places where it could be used. Then you can rebase this PR and use it as well...

registerrier (approved):
Thanks @QRemy. This looks good to me. 

The suggestion of @adonath to create a specific `footprint_mask` method might be implemented in a follow-up PR.

Issue #3843, opened 2022-03-08
Introduce Map.to_frame()


adonath:
It would be convenient to quickly change the coordinate frame of a `Map` and reproject the data at the same time.

adonath:
@registerrier Just adding the convenience, `.to_frame()`, which internally calls `.resample()` or `.interp_to_geom()`

QRemy:
We also have to determine the width of the `geom` in the new frame to avoid cutting the data.

registerrier:
What I don't understand here is what `to_frame()` is supposed to do. 

You would do `wcs_icrs_map.to_frame("galactic")` ? But how do you decide on the new WCS? Isn't it safer to leave it to the user?

Postponing to 2.1.

PR #3862, opened 2022-03-24
Add Map.reproject method

QRemy:
Add Map.reproject(geom) based on oversampling of the base map and re-distribution of the bins to the target geom. 

Note sure how to deal with the energy axis because :
- it may need to be oversampled with a different factor
- upsample along energy axis not implemented for hpx maps
To avoid these issues we could interpolate the base map along the energy axis of the target geom and then reproject  oversampling only the spatial part.

Partly solves #3843, adding Map.to_frame() would also require to determine the geom width needed to avoid cuts of  the original map.


adonath:
Thanks @QRemy, just a quick comment: in the past we also had https://reproject.readthedocs.io/en/stable/ as a dependency. If it provides the functionality (flux preserving reprojection) we can also easily re-introduce it as optional dependency. It is an astropy affiliated package anyway.

adonath:
https://github.com/gammapy/gammapy/pull/2562

QRemy:
-added option to use the `reproject.reproject_excat` method
-support axes interpolation if needed (first reproject with the same axes then interpolate)
-restore the previously removed tests (in #2562) and add new ones

QRemy:
Here are some comparisons of reprojection for this image: 
reprojection_polygon

 - `reproject(geom, method='oversampling')` vs `reproject(geom, method='polygon')`
`oversampling` is the brute-force method implemented here. Note that it is slower but will also work for HpxGeom.
`polygon` is  the  flux-conserving method  based on spherical polygon intersection from`reproject.reproject_exact` . 
rel_err_oversampling_vs_polygon

- `interp_to_geom(geom)` vs `reproject(geom, method='polygon')`
There is use cases where one want to avoid as much as possible adding more errors so it justifies to restore `reproject`.
For example we should show it in the notebooks using diffuse templates.
rel_err_interp_gammapy_vs_polygon

- `reproject.reproject_interp` vs `reproject.reproject_exact`
The interpolation from reproject seems to introduce more artefacts than the one in gammapy, not sure why
rel_err_interp_reprojct_vs_polygon

registerrier (commented):
Thanks @QRemy. 

This looks good. I have two issues:
- Should inconsistent geometries be accepted? if the role of the method is to reproject, it is a bit confusing to have some settings performing interpolation, no?
- Is the addition of `preserve_counts` on the `fill_by_xx` methods necessary? This does not seem adapted to the typical usage.

gammapy/maps/core.py
```diff
@@ -894,11 +894,120 @@ def interp_to_geom(self, geom, preserve_counts=False, fill_value=0, **kwargs):
...
+            geom = geom.copy(axes=axes)
+            axes_eq = True
+        elif geom.ndim != self.geom.ndim:
+            raise TypeError("Invalid axes in the target geom.")
+        else:
+            axes_eq = np.all(
+                [ax0 == ax1 for ax0, ax1 in zip(geom.axes, self.geom.axes)]
+            )
+
+        if not axes_eq:
```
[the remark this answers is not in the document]
QRemy: For templates the axes will rarely be the same so it will fail often, I would prefer to return the reprojection of the spatial part keeping original axes by default and add an option interp_axes  to allow the interpolation along the axes for convenience.

gammapy/maps/core.py
```diff
@@ -894,11 +894,120 @@ def interp_to_geom(self, geom, preserve_counts=False, fill_value=0, **kwargs):
...
+        if base_factor >= oversampling_factor:
+            input_map = self
+        else:
+            factor = int(np.ceil(oversampling_factor / base_factor))
+            input_map = self.upsample(factor=factor, preserve_counts=preserve_counts)
+
+        coords = input_map.geom.get_coord()
+        output_map.fill_by_coord(coords,
+                                 weights=input_map.quantity,
+                                 preserve_counts=preserve_counts
```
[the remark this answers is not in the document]
QRemy: `fill_by_coord` makes sense only for maps of integral quantities but the method exits independently of the type of data, so I think it is useful to have the option preserve_counts=False to do the average over the filled bins

gammapy/maps/core.py
```diff
@@ -894,11 +894,120 @@ def interp_to_geom(self, geom, preserve_counts=False, fill_value=0, **kwargs):
...
+
+        data = np.empty(geom.data_shape)
+
+        for img, idx in self.iter_by_image():
+            vals, footprint = reproject_exact(
+                (img, self.geom.wcs), geom.wcs, shape_out=data[idx].shape
+            )
+            data[idx] = vals
+            
+        if preserve_counts:
```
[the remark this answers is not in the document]
QRemy: changed to raise an error for this option

gammapy/maps/core.py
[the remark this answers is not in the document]
QRemy: we can't simply use the oversampling factor because  the weight is not necessarily the same for all the bins as we may have less pixels that contribute at the edges

gammapy/maps/core.py
```diff
@@ -894,11 +894,120 @@ def interp_to_geom(self, geom, preserve_counts=False, fill_value=0, **kwargs):
...
+            data *= geom.solid_angle().to_value("deg2")
+
+        return self._init_copy(geom=geom, data=data)
+
     def fill_events(self, events):
         """Fill event coordinates (`~gammapy.data.EventList`)."""
         self.fill_by_coord(events.map_coord(self.geom))
 
-    def fill_by_coord(self, coords, weights=None):
+    def fill_by_coord(self, coords, weights=None, preserve_counts=True):
```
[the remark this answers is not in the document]
QRemy: The alternative would be to add another method `???_by_coord(self, coords, weights=None, preserve_counts=False)` that replace this `fill_by_coord`  and then `fill_by_coord` would just return `???_by_coord` with `preserve_counts=True`.
I let you find the name for this method.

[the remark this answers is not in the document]
QRemy: maybe `resample_by_coord` ?

gammapy/maps/axes.py
```diff
@@ -2014,6 +2014,12 @@ def center_coord(self):
         """Center coordinates"""
         return tuple([ax.pix_to_coord((float(ax.nbin) - 1.0) / 2.0) for ax in self])
 
+    def __eq__(self, other):
```
adonath, line 2017: Please add this as an independent PR together with a unit test...

adonath:
A few more suggestions from my side:
- I think we should introduce a general `Map.resample(geom_spatial_coarse, weights=None, ufunc=np.sum)` method, similar to what we have for `Map.resample_axis`, I'll propose a prototype implementation later...(separate PR)
- For API consistency I think we should have the two methods `.interp_to_geom()` and `.reproject_to_geom()` (the later introduced in this PR)
- I think the `interp_axes` flag is a bit misleading because it overrides the information when passing in a 3D geom object. This does not make sense to me, however I can see how it is convenient. In any case I think we should introduce it for both methods `.interp_to_geom()` and `.reproject_to_geom()` consistently (ideally separate PR again!)

adonath:
@QRemy I'll open a separate PR to implement `Map.resample()`. Then you can rebase this one...

adonath:
@QRemy I just realize you already implemented the `.resample()`, based on `.fill_by_coord()` and the `preserve_counts`, which is fine. For consistency and easier review please still split the work:

- Open one PR adding `MapAxes.is_allclose()` and `MapAxes.__eq__` with unit tests
- Open one PR adding `Map.resample(geom, preserve_counts=)`, based on `.fill_by_coord()`. For now it can do the spatial resampling only. I think the `.fill_by_coord()` can remain unchanged. The re-normalisation for `preserve_counts=False` can happen in `.resample()`, if you compute the weights based on a resampled flat (includes only ones...) map. In this case there should be no need for a method`.resample_by_idx()`
- Rebase this PR, which only adds `.reproject_to_geom()`, for now remove the `interp_axes` option.

Let's have a separate discussion on the `interp_axes` option. On 2nd thought I think the current behavior is inconsistent, because `.reproject_to_geom()` does a mixture of things: while for the spatial part it uses flux conserving methods, this is not the case for the non-spatial axes, where only interpolation happens. The interpolation currently does not even take into account the scaling of the axes, which should probably be worked on first.

adonath:
Thanks @QRemy, there is still one facing unit test: https://github.com/gammapy/gammapy/runs/6195848746?check_suite_focus=true#step:9:4197

adonath:
@QRemy I realize the `Map.resample()` method was not working as expected: it did not return anything...I proposed a fix in #4085. Once this PR is merged, please rebase this PR here. I verified that the `.resampled()` methods also works for extra axes.

 I would propose to adapt the following behavior:

- Rename the method here to `.reproject_to_geom()` for consistency with `.interp_to_geom()`
- Then we go for consistent behavior of both methods: if a 2d geom is passed we internally attach the axes, such that the axes remain unchanged. 
- If a 3d geom is passed we interpolated onto to this geometry for `.interp_to_geom()` as we already do and resample extra axes in `.reproject_to_geom()`

I"ll add a few more inline comments now.

gammapy/maps/core.py
```diff
@@ -981,6 +981,61 @@ def _resample_by_idx(self, idx, weights=None, preserve_counts=False):
...
+
+        Returns
+        -------
+        output_map : `Map`
+            Reprojected Map
+        """
+        from .region import RegionGeom
+        from .hpx import HpxGeom
+
+        if isinstance(geom, RegionGeom) or isinstance(self.geom, RegionGeom):
```
adonath, line 1010: This should just work, why forbid it? I think it is basically equivalent to `.to_region_nd_map()`, but still. Or when raising an error, mention `.to_region_nd_map()` as an alternative solution.

gammapy/maps/core.py
```diff
@@ -981,6 +981,61 @@ def _resample_by_idx(self, idx, weights=None, preserve_counts=False):
...
+            raise TypeError(f"Reproject from {type(self.geom)} to {type(geom)} is not supported")
+
+        
+        if geom.is_image:
```
adonath, line 1014: Keep this, but let's support for `.interp_to_geom()` as well...

gammapy/maps/core.py
```diff
@@ -981,6 +981,61 @@ def _resample_by_idx(self, idx, weights=None, preserve_counts=False):
...
+            axes = [ax.copy() for ax in self.geom.axes]
+            geom = geom.copy(axes=axes)
+        else:
+            raise TypeError("Invalid target geometry, use geom.to_image()")
+
+        output_map = Map.from_geom(geom, unit=self.unit)
+        base_factor = geom.pixel_scales.min()/self.geom.pixel_scales.min()
```
adonath, line 1021: Maybe simplify here and just always apply the `oversampling_factor` to `self`? One could e.g. choose a factor of 2 as default as a compromise...

gammapy/maps/core.py
```diff
@@ -981,6 +981,61 @@ def _resample_by_idx(self, idx, weights=None, preserve_counts=False):
...
+            input_map = self
+        else:
+            factor = oversampling_factor / base_factor
+            if isinstance(self.geom, HpxGeom):
+                factor = int(2**np.ceil(np.log(factor)/np.log(2)))
+            else:
+                factor = int(np.ceil(factor))
+            input_map = self.upsample(factor=factor, preserve_counts=preserve_counts)
+
+        output_map.resample(input_map.geom,
```
adonath, line 1032: Adapt API after #4085.

adonath (commented):
Thanks @QRemy, I have left a few more inline comments.

In general should we required that the target geometry has a larger binning?

gammapy/maps/core.py
QRemy: `pixel_scales` is not defined

adonath: I see, one would need to work with `RegionGeom.to_wcs_geom().pixel_scales`...

gammapy/maps/core.py
adonath: I think the comment is obsolete after we agreed to keep the "adaptive resampling factor" and rename it from `oversampling_factor` to something else.

gammapy/maps/axes.py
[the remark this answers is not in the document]
QRemy: already exits  in current master

gammapy/maps/core.py
```diff
@@ -1012,6 +1012,79 @@ def interp_to_geom(self, geom, preserve_counts=False, fill_value=0, **kwargs):
...
+            factor = precision_factor / base_factor
+            if isinstance(self.geom, HpxGeom):
+                factor = int(2 ** np.ceil(np.log(factor) / np.log(2)))
+            else:
+                factor = int(np.ceil(factor))
+            input_map = self.upsample(factor=factor, preserve_counts=preserve_counts)
+
+        output_map = input_map.resample(geom3d, preserve_counts=preserve_counts)
+
+        if not geom.is_image and geom.axes != geom3d.axes:
```
adonath, line 1074: As a side note I think we should support this directly as proposed here https://github.com/gammapy/gammapy/issues/4110, but not for v1.0. So keep as is for now.

adonath (approved):
Thanks @QRemy, no further comments from myself side.

PR #3864, opened 2022-03-29
Restructure getting started page

adonath:
**Description**

This pull request restructures the "Getting Started" page. The new structure is shamelessly inspired from the Pandas documentation. The main changes are:

- Move the simple installation instructions to boxes at the top
- Merge the "install" and "other install" and upgrade page and link to it from the "in depth instructions" box
- Move the "naive" how tos (such as how to do a spectral/spatial/combined analysis) to the getting started page

This is what it looks like:

image


adonath:
image

LauraOlivera:
Hi Axel, I think it looks great! Much clearer and concise.

However, I think the instructions to update an existing environment should be a little bit more visible. At least, mention it among the examples of cases that would require one to click the "learn more" button.

I've already seen it several times, that people with an older version just update through conda and then some of the optional dependencies (imunuit, matplotlib...) don't update, causing errors that are relatively hard to debug if you don't know what to look for. 

In a related note, I would also argue that those two packages (the default fitting backend and the one that provides the ability to plot) should not really be optional since they are part of the workflow of the majority of users... but that is a different discussion.

Bultako (changes requested):
Thanks @adonath !
It's definitely better this way.
I have copied this version of the docs in the [usual place](https://www.iaa.csic.es/~jer/gammapydocs/getting-started/index.html) for others to consult.

I have two minor comments:
- I would remove the subtitle `Using Anaconda / Miniconda` on the Upgrade section.
- You can safely remove the files `install.png`, `using.png`, `gears.png` and `upgrade.png` from the `docs/_statics` folder

adonath:
Thanks @LauraOlivera, good point! I opened a separate issue for the dependency discussion https://github.com/gammapy/gammapy/issues/3865

Thanks for the preview @Bultako! I have never tried but I guess I could also have my own `gammapy-docs-preview` repository and just upload a preview there? Or is there anything special we do for `gammapy-docs`, except for the URL forwarding?

adonath:
@Bultako And if we introduce a second small section on how to update with `pip`?

docs/getting-started/index.rst
```diff
@@ -11,85 +11,163 @@ Getting started
...
-    A short recipe on how to upgrade.
+.. accordion-header::
+    :id: collapseSix
+    :title: How to compute a 2D image
+    :link: ../tutorials/index.html#d-image
 
-    +++
+Gammapy treats 2D maps as 3D cubes with one bin in energy. Sometimes, you might want to use previously
+obtained images lacking an energy axis (eg: reduced using traditional IACT tools) for modeling and fitting
+inside Gammapy. In this case, it is necessary to attach an `energy` axis on as it is showm in the tutorials.
```
AtreyeeS, line 89: shown*

AtreyeeS: Do we have such an example in the tutorials? I started to introduce this here https://github.com/gammapy/gammapy/pull/3350, but then we decided to have it in the internal HESS documentation.

docs/getting-started/index.rst
```diff
@@ -11,85 +11,163 @@ Getting started
...
+    :title: How to compute a lightcurve
+    :link: ../tutorials/analysis/time/light_curve.html
 
-    Installing with pip or other package managers.
+The light curve estimation tutorial shows how to extract a run-wise lightcurve.
 
-    +++
+To perform an analysis in a time range smaller than that of an observation, it
+is necessary to filter the latter with its `~gammapy.data.Observations.select_time` method. This produces
+an new observation containing events in the specified time range. With the new
```
AtreyeeS, line 103: a* new observation

docs/getting-started/index.rst
```diff
@@ -11,85 +11,163 @@ Getting started
...
+    :id: collapseTwelve
+    :title: How to detect sources in an image
+    :link: ../tutorials/analysis/2D/detect.html
+
+Gammapy provides methods to perform source detection in a 2D map. First step is
+to produce a significance map, i.e. a map giving the probability that the flux
+measured at each position is a background fluctuation. For a
+`~gammapy.datasets.MapDataset`, the class `~gammapy.estimators.TSMapEstimator` can be
+used. A simple correlated Li & Ma significance can be used, in particular for
+ON-OFF datasets. The second step consists in applying a peak finer algorithm,
```
AtreyeeS, line 121: peak - finder*

docs/getting-started/index.rst
```diff
@@ -11,85 +11,163 @@ Getting started
...
+.. accordion-footer::
+
+
+.. _download-tutorials:
+
+Download Tutorial Datasets
+--------------------------
+
+You can now proceed to download the Gammapy tutorial notebooks and the example
+datasets. The total size to download is ~180 MB. Select the location where you
```
AtreyeeS, line 143: Maybe we can also mention that the catalogs are also located here. Sometimes people got the idea that the catalogs are accessed directly from the "websites", whereas GAMMAPY_DATA is only for the DL3 data

QRemy:
Thanks @adonath 
maybe move the section "Download Tutorial Datasets" before the introduction ( because it points to tutorials using these datasets)

Bultako:
>  I have never tried but I guess I could also have my own gammapy-docs-preview repository and just upload a preview there?

I'm afraid you have to find your own web server, this one is access restricted for people at the institute ;(

> And if we introduce a second small section on how to update with pip?

I would just keep the small page Upgrade after Troubleshooting as it is [now](https://docs.gammapy.org/dev/getting-started/upgrade.html)

AtreyeeS (commented):
Thanks @adonath ! This looks good now.

docs/getting-started/index.rst
adonath: I agree this would be good to explain, however I think the "Getting Started" page is not the right place...it should not give any details but just help people to get started. I'll think about a better place...

adonath:
Thanks @QRemy, I have thought about that too...I thought as an alternative to move these instructions to the top of https://docs.gammapy.org/dev/tutorials/index.html . Any opinions on that?

docs/getting-started/index.rst
adonath: I think the perfect solution would be something like Pandas has done here: https://pandas.pydata.org/docs/getting_started/index.html#coming-from

So sub-pages like `Coming from HESS`, `Coming from CTA`, `Coming from MAGIC`, etc. But this could be a separate PR...

registerrier (commented):
Thanks @adonath This looks really nice. 

Yet I think the main and most visible installation method should be the environment file. Maybe you could put this first.

Putting analysis use cases before notebooks download might be surprising

QRemy:
> Thanks @QRemy, I have thought about that too...I thought as an alternative to move these instructions to the top of https://docs.gammapy.org/dev/tutorials/index.html . Any opinions on that?

It's fine in the getting started.
In the tutorial gallery page, I would keep the text on top short (even maybe reduce it in the introduction)
so ones can see more images to click-on without scrolling.

docs/getting-started/index.rst
```diff
@@ -11,85 +11,163 @@ Getting started
...
-    upgrade
-    otherinstall
 
+Installation
+------------
 
 .. panels::
-    :card: + intro-card text-center
-    :column: col-lg-6 col-md-6 col-sm-6 col-xs-12 d-flex
+    :column: col-lg-6 col-md-6 col-sm-12 col-xs-12 p-3
```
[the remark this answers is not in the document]
adonath: > Yet I think the main and most visible installation method should be the environment file.

I remember we have discussed about this in the past. While we highlighted this for previous releases, there was also concerns that this is very dependent on the preference of the users and basically no other package does something like this.

adonath:
Thanks everyone! I addressed most comments, for the remaining ones I'll make a reminder issue.

Issue #3866, opened 2022-03-30
`MapDatasetEventSampler` gives wrong MC_ID for the IRF bkg model


fabiopintore:
**Gammapy version**
gammapy-0.19

**Bug description**
`MapDatasetEventSampler` is designed to associate the MC_ID=0 to all the IRF sampled background events. This is correctly done in the table, but in the metadata, where the IRF background has a different MC_ID depending on the number of sampled source sky-models.

**Expected behavior**
The MC_ID of the IRF background has to be consistent in the event list table and in the metadata, and should be 0.

**To Reproduce**
To reproduce:

```
filename = "$GAMMAPY_DATA/cta-caldb/Prod5-North-20deg-AverageAz-4LSTs09MSTs.180000s-v0.1.fits.gz"
IRFS = load_cta_irfs(filename)

pointing = SkyCoord(ra=217.129, dec=42.678, 
                    unit=u.deg, frame="icrs")

bkg_model = FoVBackgroundModel(dataset_name="my-dataset")

spec = PowerLawSpectralModel(
                    amplitude="1e-11 TeV-1 cm-2 s-1",
                    reference="1 TeV", index=2
                    )
spatial = PointSpatialModel(
                    lon_0="217.129 deg", lat_0="42.678 deg",
                    frame="icrs",
                    )
sky = SkyModel(
                spectral_model=spec,
                spatial_model=spatial,
                name="test",
                )

models = Models([sky,
                         bkg_model])


observation = Observation.create(
                  obs_id="0001", pointing=pointing, 
                  livetime="2 hr", 
                  irfs=IRFS
                  )

energy_axis = MapAxis.from_energy_bounds(
    "0.05 TeV", "100 TeV", nbin=10, per_decade=True
    )
energy_axis_true = MapAxis.from_energy_bounds(
    "0.01 TeV", "110 TeV", nbin=20, per_decade=True, name="energy_true"
    )
migra_axis = MapAxis.from_bounds(
    0.5, 2, nbin=150, node_type="edges", name="migra"
    )

geom = WcsGeom.create(
    skydir=pointing,
    width=(2, 2),
    binsz=0.02,
    frame="icrs",
    axes=[energy_axis],
)

empty = MapDataset.create(
        geom,
        energy_axis_true=energy_axis_true,
        migra_axis=migra_axis,
        name="my-dataset",
            )
maker = MapDatasetMaker(selection=["exposure", "background", "psf", "edisp"])
[...]

PR #3867, opened 2022-03-30
Fix bug on MC_ID when sampling IRF background events

fabiopintore:
This PR fixes the bug reported in #3866.
When sampling events from the an IRF background model, `MapDatasetEventSampler` is set like that the MC_ID of these events is always zero. However, this is true only in the table of sample event list, but not in the metadata where it can assume every kind of value.


gammapy/datasets/simulate.py
```diff
@@ -295,8 +295,12 @@ def event_list_meta(dataset, observation):
         meta["CONV_DEC"] = 0
 
         for idx, model in enumerate(dataset.models):
-            meta["MID{:05d}".format(idx + 1)] = idx + 1
-            meta["MMN{:05d}".format(idx + 1)] = model.name
+            if model.name != dataset.background_model.name:
```
registerrier, line 298: Written this way, this can create an empty model ID, if e.g. the background model is second in the list and another source model is defined in third position.
I suppose one could first reorder models and put the background model first and then build the meta table.

registerrier (commented):
Thanks @fabiopintore . See inline comment.

I assume one wants contiguous source model IDs,.
Also what would happen if no background model is defined but a background map is present?

gammapy/datasets/simulate.py
fabiopintore: Thanks  a lot @registerrier . 
I suppose that "reorder the model first and then simulate" can give some troubles to the users, as they have to know that this should be done. However, the tutorial does not currently provide such an information. In addition, doesn't seem strange that MC_ID=0 in the bkg table is hardcoded while, in the metadata, the setting of the correct MC_ID is left to the user? I would suggest to be coherent and make it all hard-coded. 
What about the following solution?

```
idx = 1
for model in dataset.models:
            if model.name != dataset.background_model.name:
                meta["MID{:05d}".format(idx)] = idx
                meta["MMN{:05d}".format(idx)] = model.name
                idx += 1
            else:
                meta["MID{:05d}".format(0)] = 0
                meta["MMN{:05d}".format(0)] = model.name
```

In the case of no background model, the `MapDatasetEventSampler` always crashes. So, as far as I know, we cannot have a background map and no background model.

PR #3869, opened 2022-04-02
Re-computation of upper limits on flux points

AtreyeeS:
This PR adds an `FluxPoints.recompute_ul`. If the `stat_scan` is present, upper limits can be re-computed from the same. 
It might be useful to add the same for asymmetric error computation in future.

### DIFF
--- a/gammapy/estimators/points/core.py
+++ b/gammapy/estimators/points/core.py
@@ -13,6 +13,7 @@ from gammapy.modeling.scipy import stat_profile_ul_scipy
 from gammapy.utils.scripts import make_path
 from gammapy.utils.table import table_standardise_units_copy
 from ..map.core import DEFAULT_UNIT, FluxMaps
+from copy import copy
 
 __all__ = ["FluxPoints"]
 
@@ -617,7 +618,6 @@ class FluxPoints(FluxMaps):
     def recompute_ul(self, n_sigma_ul=2, **kwargs):
         """Recompute upper limits corresponding to the given value.
         The pre-computed stat profiles must exist for the re-computation.
-        Modifies the object in place.
 
         Parameters
         ----------
@@ -625,6 +625,11 @@ class FluxPoints(FluxMaps):
             Number of sigma to use for upper limit computation. Default is 2.
         **kwargs : dict
             Keyword arguments passed to `~scipy.optimize.brentq`.
+
+        Returns
+        --------
+        flux_points : `FluxPoints`
+            A new FluxPoints object with modified upper limits
         """
 
         if self.has_stat_profiles is False:
@@ -634,12 +639,16 @@ class FluxPoints(FluxMaps):
 
         delta_ts = n_sigma_ul ** 2
 
+        flux_points = copy(self)
+
         value_scan = self.stat_scan.geom.axes["norm"].center
         shape_axes = self.stat_scan.geom._shape[slice(3, None)]
         for idx in np.ndindex(shape_axes):
             stat_scan = (
                 self.stat_scan.data[idx].squeeze() - self.stat.data[idx].squeeze()
             )
-            self.norm_ul.data[idx] = stat_profile_ul_scipy(
+            flux_points.norm_ul.data[idx] = stat_profile_ul_scipy(
                 value_scan, stat_scan, delta_ts=delta_ts, **kwargs
             )
+        flux_points.meta["n_sigma_ul"] = n_sigma_ul
+        return flux_points
--- a/gammapy/estimators/points/core.py
+++ b/gammapy/estimators/points/core.py
@@ -13,7 +13,7 @@ from gammapy.modeling.scipy import stat_profile_ul_scipy
 from gammapy.utils.scripts import make_path
 from gammapy.utils.table import table_standardise_units_copy
 from ..map.core import DEFAULT_UNIT, FluxMaps
-from copy import copy
+from copy import deepcopy
 
 __all__ = ["FluxPoints"]
 
@@ -630,6 +630,17 @@ class FluxPoints(FluxMaps):
         --------
         flux_points : `FluxPoints`
             A new FluxPoints object with modified upper limits
+
+        Examples
+        --------
+        >>> from gammapy.estimators import FluxPoints
+        >>> filename = '$GAMMAPY_DATA/tests/spectrum/flux_points/binlike.fits'
+        >>> flux_points = FluxPoints.read(filename)
+        >>> flux_points_recomputed = flux_points.recompute_ul(n_sigma_ul=3)
+        >>> print(flux_points.meta["n_sigma_ul"], flux_points.flux_ul.data[0])
+        2.0 [[3.95451985e-09]]
+        >>> print(flux_points_recomputed.meta["n_sigma_ul"], flux_points_recomputed.flux_ul.data[0])
+        3 [[6.22245374e-09]]
         """
 
         if not self.has_stat_profiles:
@@ -639,12 +650,12 @@ class FluxPoints(FluxMaps):
 
         delta_ts = n_sigma_ul ** 2
 
-        flux_points = copy(self)
+        flux_points = deepcopy(self)
 
         value_scan = self.stat_scan.geom.axes["norm"].center
         shape_axes = self.stat_scan.geom._shape[slice(3, None)]
         for idx in np.ndindex(shape_axes):
-            stat_scan = (
+            stat_scan = np.abs(
                 self.stat_scan.data[idx].squeeze() - self.stat.data[idx].squeeze()
             )
             flux_points.norm_ul.data[idx] = stat_profile_ul_scipy(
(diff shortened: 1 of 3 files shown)


gammapy/estimators/points/core.py
```diff
@@ -612,3 +613,45 @@ def plot_ts_profiles(
...
+
+        norm_ul = np.empty(shape=(ncols1, ncols2))
+        for i in range(ncols1):
+            for j in range(ncols2):
+                stat_scan = np.ravel(stat_scan_full[i][j] - stat_min[i][j])
+                norm_ul[i][j] = stat_profile_ul_scipy(
+                    value_scan, stat_scan, delta_ts=delta_ts, **kwargs
+                )
+
+        self.norm_ul.data = norm_ul.reshape(self.norm.data.shape)
```
QRemy, line 657: maybe try this to simplify :
```python
shape_axes = self.stat_scan.geom._shape[slice(3, None)]
for idx in np.ndindex(shape_axes):
    stat_scan = self.stat_scan.data[idx].squeeze() - self.stat.data[idx].squeeze()
    self.norm_ul.data[idx] = stat_profile_ul_scipy(
                    value_scan, stat_scan, delta_ts=delta_ts, **kwargs
                )
```

AtreyeeS: Thanks @QRemy ! works perfectly

AtreyeeS:
I do not understand the test fails here - they pass locally on my Mac and seem to be unrelated to this PR. Any suggestions @QRemy @registerrier @maxnoe ?

maxnoe:
@AtreyeeS This is due to a problem with the server providing the IERS A bulletin, needed for astropy coordinate transformations.

This is fixed in astropy 5.0.4, so if the failing test uses astropy 5, just rerunning should fix it, as it will get the newer astropy.

If it is using astropy 4, we would need to provide a manual fix, which I can do in a separate PR.

AtreyeeS:
Thanks @maxnoe ! The issue indeed solved after re-running

registerrier (commented):
Thanks @AtreyeeS . 
This sounds like an interesting functionality. I wonder if the UL should be modified in-place or a new `FluxPoints` object should be created.

the `meta` key containing the `n_sigma_ul` should be updated. 

Should there be a check on the number of actual points available on the stat scan?

gammapy/estimators/points/tests/test_lightcurve.py
```diff
@@ -634,3 +634,30 @@ def test_lightcurve_estimator_map_datasets():
...
+@requires_data()
+@requires_dependency("iminuit")
+def test_recompute_ul():
+    datasets = get_spectrum_datasets()
+    selection = ["all"]
+    estimator = LightCurveEstimator(
+        energy_edges=[1, 3, 30] * u.TeV, selection_optional=selection, n_sigma_ul=2
+    )
+    lightcurve = estimator.run(datasets)
+    table = lightcurve.to_table(format="lightcurve", sed_type="dnde")
```
registerrier, line 648: You might not need to convert to a table, but could directly access `light curve.dnde_ul.data`

gammapy/estimators/points/tests/test_lightcurve.py
```diff
@@ -634,3 +634,30 @@ def test_lightcurve_estimator_map_datasets():
     assert_allclose(table["norm_err"][0], [0.031508], rtol=1e-2)
     assert_allclose(table["counts"][0], [[2205, 2220]])
     assert_allclose(table["ts"][0], [2557.346464], rtol=1e-2)
+
+
+@requires_data()
+@requires_dependency("iminuit")
+def test_recompute_ul():
```
registerrier, line 641: Maybe the test could also check that it recovers a sensible `ul` with this technique compared to the one initially computed?

gammapy/estimators/points/tests/test_sed.py
```diff
@@ -498,3 +498,20 @@ def test_flux_points_estimator_small_edges():
...
     assert np.isnan(fp.ts.data[1, 0, 0])
     assert np.isnan(fp.npred.data[1, 0, 0])
+
+
+@requires_dependency("iminuit")
+def test_flux_points_recompute_ul(fpe_pwl):
+    datasets, fpe = fpe_pwl
+    fpe.selection_optional = ["all"]
+    fp = fpe.run(datasets)
+    table = fp.to_table("flux")
```
registerrier, line 508: same comment here. `to_table` is not necessary

registerrier (approved):
Thanks @AtreyeeS .

This looks good. I have left few inline comments to improve a bit the tests

gammapy/estimators/points/core.py
```diff
@@ -612,3 +614,41 @@ def plot_ts_profiles(
...
+        **kwargs : dict
+            Keyword arguments passed to `~scipy.optimize.brentq`.
+
+        Returns
+        --------
+        flux_points : `FluxPoints`
+            A new FluxPoints object with modified upper limits
+        """
+
+        if self.has_stat_profiles is False:
```
adonath, line 635: Just use `if not self.has_stat_profiles`.

gammapy/estimators/points/core.py
```diff
@@ -612,3 +614,41 @@ def plot_ts_profiles(
...
+                "Stat profiles not present. Upper limit computation is not possible"
+            )
+
+        delta_ts = n_sigma_ul ** 2
+
+        flux_points = copy(self)
+
+        value_scan = self.stat_scan.geom.axes["norm"].center
+        shape_axes = self.stat_scan.geom._shape[slice(3, None)]
+        for idx in np.ndindex(shape_axes):
```
adonath, line 646: You could use `self.stat_scan.iter_by_axis()` (see https://docs.gammapy.org/0.19/api/gammapy.maps.RegionNDMap.html#gammapy.maps.RegionNDMap.iter_by_axis)

adonath (commented):
Thanks a lot @AtreyeeS, this is very useful functionality. I have left two minor inline comments to simplify the implementation a tiny bit...

gammapy/estimators/points/core.py
AtreyeeS: I could not manage to simplify it using this while iterating over both `stat_scan` and `stat` with the generators...

adonath: OK

gammapy/estimators/points/tests/test_lightcurve.py
adonath: @AtreyeeS Can you still add this self consistency check?

adonath:
> Should there be a check on the number of actual points available on the stat scan?

I guess we would require at least 3 points to be able to do at least a parabolic interpolation? Probably we should add this check already to the `FluxEstimator` norm values?

gammapy/estimators/points/tests/test_lightcurve.py
AtreyeeS: Can you suggest a good check? The value changes from `3.2` to `3.77`  in going from `2` to `4` sigma, which I think is sensible... But what would be a good automatic check?

QRemy: check if `.recompute_ul(n_sigma_ul=2)` is close to 3.2 with this method (it may be different if the default profile scan sampling is not good enough)

gammapy/estimators/points/core.py
```diff
@@ -612,3 +614,41 @@ def plot_ts_profiles(
...
+        """Recompute upper limits corresponding to the given value.
+        The pre-computed stat profiles must exist for the re-computation.
+
+        Parameters
+        ----------
+        n_sigma_ul : int
+            Number of sigma to use for upper limit computation. Default is 2.
+        **kwargs : dict
+            Keyword arguments passed to `~scipy.optimize.brentq`.
+
```
adonath, line 628: Please add an example here using `"$GAMMAPY_DATA/tests/spectrum/flux_points/binlike.fits"`

adonath (approved):
Thanks @AtreyeeS, can you please add an example to the docstring. Otherwise the PR is ready from my side.

gammapy/estimators/points/core.py
AtreyeeS: Hi @adonath ! How was this file produced? The stat scan values on this seem to be the negative of our current convention

adonath: Yes, but we have the conversion code here: https://github.com/gammapy/gammapy/blob/master/gammapy/estimators/points/core.py#L234

Doesn't this work?

AtreyeeS: Okay added

AtreyeeS: Implicitly it did not. I added an `np.abs` so that a convex function is passed to `stat_profile_ul_scipy`

adonath: Thanks! I wonder why it works for the plotting...I can check...

adonath (approved):

PR #3870, opened 2022-04-04
Enable serialisation of spectrum datasets

AtreyeeS:
This introduces a `format=gadf` option for the serialisation of `SpectrumDataset`. The serialisation was forbidden before, but it works directly as expected for `MapDataset` with a `RegionGeom`. I added the clarification in the doc strings, don't know if it needs to be explicitly mentioned somewhere in the docs as well. Ideally this should go in the  GADF I imagine, but since https://github.com/open-gamma-ray-astro/gamma-astro-data-formats/pull/170 is still in progress, we may add a serialisation subsection in the Package guide DL4 page...

### DIFF
--- a/gammapy/datasets/tests/test_io.py
+++ b/gammapy/datasets/tests/test_io.py
@@ -1,8 +1,10 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 from numpy.testing import assert_allclose
-from gammapy.datasets import Datasets
+from gammapy.datasets import Datasets, SpectrumDataset
 from gammapy.modeling.models import DatasetModels
 from gammapy.utils.testing import requires_data, requires_dependency
+import astropy.units as u
+from regions import CircleSkyRegion
 
 
 @requires_data()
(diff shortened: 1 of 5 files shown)


gammapy/datasets/tests/test_io.py
```diff
@@ -71,3 +73,36 @@ def test_datasets_to_io(tmp_path):
...
+@requires_dependency("iminuit")
+def test_spectrum_datasets_to_io(tmp_path):
+    filedata = "$GAMMAPY_DATA/tests/models/gc_example_datasets.yaml"
+    filemodel = "$GAMMAPY_DATA/tests/models/gc_example_models.yaml"
+
+    datasets = Datasets.read(
+        filename=filedata,
+        filename_models=filemodel,
+    )
+    g = datasets[0]._geom
```
QRemy, line 88: unused

gammapy/datasets/tests/test_spectrum.py
```diff
@@ -70,6 +70,46 @@ def test_set_model(spectrum_dataset):
...
+    spectrum_dataset.write(tmp_path / "test.fits")
+    dataset_new = SpectrumDataset.read(tmp_path / "test.fits", name="test")
+
+    assert_allclose(spectrum_dataset.counts.data, dataset_new.counts.data)
+    assert_allclose(
+        spectrum_dataset.npred_background().data, dataset_new.npred_background().data
+    )
+    assert dataset_new.edisp is None
+    assert dataset_new.edisp is None
+    assert dataset_new.name is "test"
```
QRemy, line 105: assert dataset_new.name == "test"

gammapy/datasets/tests/test_io.py
AtreyeeS: oops sorry

registerrier (commented):
Thanks @AtreyeeS . 

I am not sure I fully understand here. 
We can probably store a `SpectrumDataset` following the OGIP format. But the issue is that currently the `OGIPDatasetReader` creates a OnOff dataset.
As far as I know, the OGIP format does not explicitly provides something to explain that the background is a template/model rather than a measurement.

adonath:
Thanks a lot @AtreyeeS, I agree with @registerrier, here I think the serialization of the `SpectrumDataset` should not go through the `OGIPDatasetWriter`, but just use the same mechanism as `MapDataset`. SO convert all the data to and `HDUList`, based one the standard `RegionNDMap` format.
So `SpectrumDataset.write()` should support `format={"gadf"}`, while `SpectrumDatasetOnOff.write()` supports `format={"gadf", "ogip"}`. I hope this makes sense...

adonath (approved):
Thanks a lot @AtreyeeS, no further comments from my side. After the feature freeze we should take some time to document the dataset serialization in the Gammapy docs, ideally before the next release.

PR #3872, opened 2022-04-05
Fix gaussians with missing size in SourceCatalogHGPS.to_models

QRemy:
Fix SourceCatalogHGPS.to_models for single gaussian sources with missing size in the main table using informations from the components table.


AtreyeeS (approved):
Thanks @QRemy ! No comments from my side.

registerrier (approved):
Thanks @QRemy !

Issue #3851, opened 2022-03-15
`bkg-peak` option in `SafeMaskMaker` always masks the first bin


AtreyeeS:
**Gammapy version**
dev, 0.19

**Bug description**

Running `SafeMaskMaker` with the `bkg-peak` option *always* masks the first bin, irrespective of the lower energy threshold.

**Expected behavior**
No bins should be masked if the lower threshold is above the peak energy

**To Reproduce**

See https://github.com/AtreyeeS/test_notebooks/blob/master/test-bkg-peak.ipynb

adonath:
Fixed by #3873.

PR #3873, opened 2022-04-05
Modify SafeMaskMaker to ignore data only below the background peak 

AtreyeeS:
This PR addresses #3851
The energy bin containing the peak of the background is now kept.
Surprisingly, this did not affect any of the fitting test results...

### DIFF
--- a/gammapy/makers/tests/test_safe.py
+++ b/gammapy/makers/tests/test_safe.py
@@ -73,7 +73,7 @@ def test_safe_mask_maker(observations, caplog):
     assert_allclose(mask_edisp_bias_offset.data.sum(), 1694)
 
     mask_bkg_peak = safe_mask_maker.make_mask_energy_bkg_peak(dataset)
-    assert_allclose(mask_bkg_peak.data.sum(), 1815)
+    assert_allclose(mask_bkg_peak.data.sum(), 1936)
     assert "WARNING" in [_.levelname for _ in caplog.records]
     message1 = "No default thresholds defined for obs 110380"
     assert message1 in [_.message for _ in caplog.records]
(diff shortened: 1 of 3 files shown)


registerrier (commented):
Thanks @AtreyeeS . It is indeed strange that the tests do not change more.

Can it be due to the way the function is designed? Taking the npred on the `Dataset` energy axis is likely insufficient to locate the maximum. 
Currently most safe mask functions rely on the projected IRFs. I think this can be problematic.  Note that the `aeff-default` accesses the `obs.aeff.meta`. 
Another approach could be to find the position of the maximum directly in the `obs.bkg` and exclude all bins with upper edges below the energy.

adonath:
@AtreyeeS @registerrier I think we do not use the background method in tests do we? What I can see in the diff is that at least the tests, that are expected to change do change.

adonath (changes requested):
Thanks @AtreyeeS! Please adapt the documentation of the method in the `SafeMaskMaker`, otherwise looks good to me...

gammapy/makers/tests/test_safe.py
```diff
@@ -86,6 +86,14 @@ def test_safe_mask_maker(observations, caplog):
     assert_allclose(mask_aeff_max_noroot.data.sum(), 1815)
     assert_allclose(mask_edisp_bias_noroot.data.sum(), 1936)
 
+    ## test if the peak is in the first bin
+    axis = MapAxis.from_bounds(1.0, 10, nbin=6, unit="TeV", name="energy", interp="log")
+    geom = WcsGeom.create(npix=(5, 5), axes=[axis], skydir=obs.pointing_radec)
+    empty_dataset = MapDataset.create(geom=geom, energy_axis_true=axis_true)
+    dataset = dataset_maker.run(empty_dataset, obs)
+    mask_bkg_peak = safe_mask_maker.make_mask_energy_bkg_peak(dataset)
+    assert_allclose(np.all(mask_bkg_peak), True)
```
adonath, line 95: Just use `assert np.all(mask_bkg_peak)` here...

AtreyeeS:
Thanks @adonath ! Should be safe to merge now I think

adonath (approved):

PR #3874, opened 2022-04-06
Allow MapDatasetOff in Datasets serialization and avoid read_lazy

QRemy:
Allow `MapDatasetsOff` in Datasets serialization. 
Note that `lazy=True` is only supported for `MapDataset` so in order  to prevent a fail I added a `MapDatasetsOff._read_lazy()` that avoid the problem but I did not actually implemented the lazy loading (could be done in a separate PR).

### DIFF
--- a/gammapy/datasets/__init__.py
+++ b/gammapy/datasets/__init__.py
@@ -5,7 +5,7 @@ from .map import MapDataset, MapDatasetOnOff, create_map_dataset_geoms
 from .simulate import MapDatasetEventSampler
 from .spectrum import SpectrumDataset, SpectrumDatasetOnOff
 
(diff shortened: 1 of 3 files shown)


registerrier (commented):
Thanks @QRemy.

Why not raising a `NotImplementedError` rather that doing something else than required? Is it an issue if the `lazy=True` fails for now?

QRemy:
The default for the `Datasets.read` is `lazy=True` so it will fail by default for the `MapDatasetOnOff`, and setting it to `False` by default could make it long to execute.

registerrier (approved):
Thanks @QRemy .This looks good.

Issue #3859, opened 2022-03-22
Enable temporal model fitting for flux points


AtreyeeS:
It will be useful to be able to fit temporal models directly on the output of the `LightCurveEstimator`, using `FluxPointsDataset`, similar to the available functionality for spectral models.

adonath:
I agree this would be useful. I think it can be achieved rather simple, however there are two different possibilities:
- Allow for `FluxPoints` with a time axis in `FluxPointsDataset` and use the time info from there
- Introduce `FluxPointsDataset.gti` and split the `FluxPoints` per time bin into multiple `FluxPointsDataset` objects. This would be consistent with what we do for the `MapDataset`, this is the less elegant solution, but maybe preferred for consistency for now.

AtreyeeS:
Implemented in #3876

PR #3876, opened 2022-04-07
Add temporal model on flux point dataset

AtreyeeS:
This PR adds a temporal model evaluation on a `FluxPointDataset` as discussed in #3859 

However, to be able to fit a lightcurve with temporal models is still complicated - the user has to create a `FluxPointsDataset` for each time bin of the `FluxPoints` object. We can create a `FluxPoints.split_by_time()` maybe?


AtreyeeS:
@registerrier : Added a test for the dimension as discussed in the dev call. Once this is merged, I can modify our tutorials to show an example

adonath:
Thanks a lot @AtreyeeS, this is a very nice feature! I think `split_by_time()` would be useful, but maybe implement it in a more general way like `Map.split_by_axis(axis_name=)` and `FluxMaps.split_by_axis(axis_name=)`.

registerrier:
A few comments:
- Currently we have no `gti` exported with `FluxPoints`. Maybe the `gti` could be obtained from the time axis itself.
- Note that this may contradict the notion of GTI as provenance info (i.e. list of data time intervals used for creation of the object)
- I am not sure that a `split_by_time` method would be adapted. What would it return? Can't you do the same thing with `FluxMaps.slice_by_idx`. Maybe more useful would be a function that return a `Datasets` filled with `FluxPointsDataset` directly from a `FluxPoints`  object?

gammapy/datasets/flux_points.py
```diff
@@ -259,7 +261,13 @@ def flux_pred(self):
         """Compute predicted flux."""
         flux = 0.0
         for model in self.models:
-            flux += model.spectral_model(self.data.energy_ref)
+            flux_model = model.spectral_model(self.data.energy_ref)
+            if model.temporal_model is not None:
+                integral = model.temporal_model.integral(
+                    self.gti.time_start, self.gti.time_stop
```
adonath, line 267: I think `FluxPointsDataset` currently does not have a `.gti` attribute. However the `FluxMaps` and `FluxPoints` have a `.gti` attribute which, could be re-used. We should think about which way to go and how to serialize the GTI information as well.

gammapy/datasets/tests/test_flux_points.py
```diff
@@ -24,18 +31,18 @@ def dataset():
     path = "$GAMMAPY_DATA/tests/spectrum/flux_points/diff_flux_points.fits"
     table = Table.read(make_path(path))
     table["e_ref"] = table["e_ref"].quantity.to("TeV")
+    gti = GTI.create(start=0 * u.s, stop=30 * u.min)
     data = FluxPoints.from_table(table, format="gadf-sed")
+    data.gti = gti
```
adonath, line 36: Same comment as above: the `.gti` attribute should be added to `__init__` and documented.

gammapy/datasets/flux_points.py
```diff
@@ -89,6 +89,8 @@ def __init__(
         name=None,
         meta_table=None,
     ):
+        if data.geom.ndim != 3 or data.geom.has_energy_axis is False:
```
adonath, line 92: Just use `not data.geom.has_energy_axis`.

adonath (changes requested):
Thanks a @AtreyeeS, I have left some first inline comments.

adonath:
@registerrier I guess for now we stick with the information stored in `dataset.gti`? However we should make sure to document, that currently serializing GTIs in `FluxPointsDataset.write()` is not supported, because there is no form for it. Alternatively we could allow serialising to the `"lightcurve"` format, but this might be confusing.

gammapy/datasets/flux_points.py
registerrier: the `self.gti` property returns `self.data.gti`

gammapy/maps/tests/test_core.py
```diff
@@ -485,3 +485,21 @@ def test_map_plot_mask():
...
+        24,
+        nbin=24,
+        unit="hour",
+        name="time",
+    )
+    energy_axis = MapAxis.from_bounds(
+        1, 100, nbin=4, unit="TeV", name="energy", interp="log"
+    )
+    m_4d = Map.create(
+        binsz=0.02, width=(10, 5), frame="galactic", axes=[energy_axis, time_axis]
```
registerrier, line 502: I would make it smaller to limit the memory requirements of tests. e.g. use number of pixels instead.

gammapy/maps/core.py
```diff
@@ -1442,6 +1442,27 @@ def from_stack(cls, maps, axis=None, axis_name=None):
             data=np.stack(data), geom=geom.to_cube(axes=[axis]), unit=maps[0].unit
         )
 
+    def split_by_axis(self, axis_name):
```
registerrier, line 1445: maybe add a `keepdims` argument to allow for axis dropping.

registerrier: Another question, in terms of API, I wonder whether it would not make more sense to generalize `RegionNDMap.iter_by_axis()` to all `Map` types? In that case, the need for the `split_by_axis` is less strong.
Any idea @adonath ?

AtreyeeS: I wanted to add `keepdims`, but it requires changes in `Geom` and `MapAxis`. I can do it in a followup PR. (Maybe it was a bad idea to do the `split_by_axis` in this PR in the first place)

gammapy/datasets/flux_points.py
AtreyeeS: We are basically re-using the `FluxPoints.gti here`. Should it go on `init` ?

adonath: Sorry, I didn’t realize FluxPointsDataset.gti already existed. I has been added in an earlier PR. I think the current behavior is fine.

gammapy/maps/core.py
adonath: I think it is a good suggestion, however there is one detail:

For historical reasons we have `Map.iter_by_image`, which returns the data as array and the corresponding ND- index. I later implemented `RegionNDMap.iter_by_axis()` with the same API and return values. But I think it is not the API the users would like to have. For users one would rather directly return a sliced or indexed `Map` object. The only clean solution I see would be:
- Rename the existing `.iter_by_image` to `.iter_by_image_data` and `.iter_by_axis_data`
- Change `.iter_by_image` and `.iter_by_axis` to directly return `Map` objects
We could still keep `.split_by_axis()`...

AtreyeeS:
I tried to adapt this PR a bit taking into your comments.
`FluxPointsDataset` can now be created from a lightcurve by:

```
flux_maps = lc_1d.split_by_axis("time")
datasets = Datasets()
for amap in flux_maps:
    datasets.append(FluxPointsDataset(data=amap))
```

and directly used for modelling and fitting.

@adonath and @registerrier : Can you please do a review once again?
If this looks good, I can adapt the tutorials (the `lightcurve_simulation` ?) and add a HowTo in the docs (maybe in a separate PR)

gammapy/maps/core.py
adonath: @AtreyeeS I'll open a PR to implement my suggestion from the comment above. The you can rebase and simplify your implementation...

gammapy/estimators/map/core.py
```diff
@@ -809,6 +813,45 @@ def from_stack(cls, maps, axis, meta=None):
             data=data, reference_model=reference.reference_model, meta=meta, gti=gti
         )
 
+    def split_by_axis(self, axis_name):
```
adonath, line 816: I just opened #3899, after this is merged my proposal would be to implement `FluxPoints.iter_by_axis(axis_name, keepdims=)`. Then the implementation becomes something along the lines of:
```python3
datasets = Datasets()

for fp in flux_points.iter_by_axis(axis_name="time", keepdims=True):
    gti = fp.geom.axes["time"].to_gti()
    dataset = FluxPointsDataset(fp, gti=gti)
    datasets.append(datasets)
```
Where `TimeMapAxis.to_gti()` would still need to be implemented. What do you think @AtreyeeS?

AtreyeeS: We might need to distinguish between the time axis bins and the gti, no?
For example, the `time_axis` has a constant binning, say 6 hrs. And within that 6 hrs, the telescope was operational only for 2 hrs. The time integration (that is taken from the gti) in this case should be 2hrs (and that is what we take within MapEvaluator)

adonath: Yes, you are right. The code example above does not make sense, it would only work for the "flare" case, where we split into smaller GTIs. So the handling of GTIs remains in `FluxPoints.iter_by_axis()`  based on `gti.select_time()` and makes the code example form the user side even simple:
```python3
datasets = Datasets()

for fp in flux_points.iter_by_axis(axis_name="time", keepdims=False):
    dataset = FluxPointsDataset(fp)
    datasets.append(datasets)
```

gammapy/estimators/map/core.py
```diff
@@ -809,6 +813,45 @@ def from_stack(cls, maps, axis, meta=None):
...
+
+        split_maps = {}
+        flux_maps = []
+        for amap in self.available_quantities:
+            split_maps[amap] = getattr(self, amap).split_by_axis(axis_name)
+
+        axis = self.geom.axes["time"]
+
+        gti = self.gti
+        for idx in range(axis.nbin):
```
adonath, line 838: You could use `TimeMapAxis.iter_by_edges` here.

gammapy/estimators/map/core.py
AtreyeeS: Should I remove `FluxPoints.slice_by_axis()` (but keep `Map.slice_by_axis`)?

adonath (approved):
Thanks @AtreyeeS, I  added a comparison plot to the notebook.

PR #3880, opened 2022-04-11
Fix EnergyDispersion2D.from_gauss if sigma is an array

QRemy:
Fix EnergyDispersion2D.from_gauss if sigma is an array. For now this works only if   sigma is given with a shape (N,1,1) which is not obvious to guess, otherwise there is an error about invalid broadcast shapes. So I adapted the test to not assume a (N,1,1)  shape and applied a transposition within .from_gauss().


registerrier (approved):

Issue #3881, opened 2022-04-11
TemplateSpectralModel fails for single energy bin


registerrier:
**Gammapy version**
dev

**Bug description**
`TemplateSpectralModel` fails on init when the input arrays have a single elements. The error raised is:
```
ValueError: Interpolating scalar values requires using method='nearest' explicitely.
```
This is likely connected with the change introduced in PR #3837 

Note that this is problematic because `likelihood` type light curves cannot be serialized from disk.

**Expected behavior**
The model initialization should proceed.

**To Reproduce**
```
import astropy.units as u
from gammapy.modeling.models import TemplateSpectralModel
TemplateSpectralModel([10]*u.TeV, [1])
```

registerrier:
The test pass because it uses the `sed_type=flux`. 

The model `init` works in v0.19 but fails on evaluate.

registerrier:
This is solved by #3882. Closing

PR #3882, opened 2022-04-11
Adapt TemplateSpectralModel to support single bin input

registerrier:
**Description**

This pull request solves issue #3881 .  If a single value is passed to the `TemplateSpectralModel` init, the `nearest` interpolation method will be applied.

This allows read/write of `FluxPoints` with single energy bin in `likelihood` sed format. A test is added to test read/write of the various formats.

**Dear reviewer**


registerrier:
Merging now.

PR #3883, opened 2022-04-14
Fix FluxPointsEstimators to support TemplateNPredModel

QRemy:
For each energy band the FluxPointsEstimators creates a dataset sliced in energy and copies the models but if TemplateNPredModel are present they are not sliced which cause an error in npred computation (invalid dimension to stack the maps). This PR fix the issue.

### DIFF
--- a/gammapy/estimators/points/tests/test_sed.py
+++ b/gammapy/estimators/points/tests/test_sed.py
@@ -400,6 +400,7 @@ def test_run_template_npred(fpe_map_pwl, tmpdir):
     model = TemplateNPredModel(dataset.background, datasets_names=[dataset.name])
     models.append(model)
     dataset.models = models
+    dataset.background.data = 0
 
     fp = fpe.run(dataset)
 
--- a/gammapy/estimators/points/tests/test_sed.py
+++ b/gammapy/estimators/points/tests/test_sed.py
@@ -22,8 +22,10 @@ from gammapy.modeling.models import (
     Models,
     PiecewiseNormSpectralModel,
     PowerLawSpectralModel,
+    TemplateNPredModel,
     SkyModel,
     TemplateSpatialModel,
+    Models,
 )
 from gammapy.utils import parallel
 from gammapy.utils.testing import requires_data, requires_dependency
@@ -389,6 +391,15 @@ def test_run_map_pwl_reoptimize(fpe_map_pwl_reoptimize):
     actual = table["stat_scan"][0] - table["stat"][0]
     assert_allclose(actual, [9.788123, 0.486066, 17.603708], rtol=1e-2)
 
+@requires_dependency("iminuit")
+def test_run_template_npred(fpe_map_pwl, tmpdir):
+    datasets, fpe = fpe_map_pwl
+    dataset = datasets[0]
+    models = Models(dataset.models)
+    model = TemplateNPredModel(dataset.background, datasets_names=[dataset.name])
+    models.append(model)
+    dataset.models = models
+    fpe.run(dataset)
 
 @requires_data()
 def test_reoptimize_no_free_parameters(fpe_pwl, caplog):
--- a/gammapy/estimators/points/tests/test_sed.py
+++ b/gammapy/estimators/points/tests/test_sed.py
@@ -22,10 +22,9 @@ from gammapy.modeling.models import (
     Models,
     PiecewiseNormSpectralModel,
     PowerLawSpectralModel,
-    TemplateNPredModel,
     SkyModel,
+    TemplateNPredModel,
     TemplateSpatialModel,
-    Models,
 )
 from gammapy.utils import parallel
 from gammapy.utils.testing import requires_data, requires_dependency
@@ -391,6 +390,7 @@ def test_run_map_pwl_reoptimize(fpe_map_pwl_reoptimize):
     actual = table["stat_scan"][0] - table["stat"][0]
     assert_allclose(actual, [9.788123, 0.486066, 17.603708], rtol=1e-2)
 
+
 @requires_dependency("iminuit")
 @requires_data()
 def test_run_template_npred(fpe_map_pwl, tmpdir):
@@ -400,10 +400,11 @@ def test_run_template_npred(fpe_map_pwl, tmpdir):
     model = TemplateNPredModel(dataset.background, datasets_names=[dataset.name])
     models.append(model)
     dataset.models = models
-    
+
     fpe.run(dataset)
     fpe.sum_over_energy_groups = True
     fpe.run(dataset)
+
 
 @requires_data()
 def test_reoptimize_no_free_parameters(fpe_pwl, caplog):
@@ -693,7 +694,6 @@ def test_fpe_non_uniform_datasets():
 
 @requires_data()
 def test_flux_points_estimator_norm_spectral_model(fermi_datasets):
-
     energy_edges = [10, 30, 100, 300, 1000] * u.GeV
 
     model_ref = fermi_datasets.models["Crab Nebula"]
--- a/gammapy/estimators/points/tests/test_sed.py
+++ b/gammapy/estimators/points/tests/test_sed.py
@@ -392,6 +392,7 @@ def test_run_map_pwl_reoptimize(fpe_map_pwl_reoptimize):
     assert_allclose(actual, [9.788123, 0.486066, 17.603708], rtol=1e-2)
 
 @requires_dependency("iminuit")
+@requires_data()
 def test_run_template_npred(fpe_map_pwl, tmpdir):
     datasets, fpe = fpe_map_pwl
     dataset = datasets[0]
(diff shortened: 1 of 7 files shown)


QRemy:
> A side comment: this does not fix the problem if e.g. `dataset.slice_by_energy()` is called independently. One way to support this in a more general way would be to support stacking of "energy cutouts" in `Map.stack()`, i.e. allow to stack the map sliced in energy if the energy edges align. What do you think @QRemy and @registerrier?

But Dataset.slice_by_energy do not return the models at all, if we change that here it should be also done on similar method as well (and we decided against it previously)

bkhelifi:
@QRemy , what is the status if this PR?

QRemy:
Requires #5027

gammapy/modeling/models/core.py
```diff
@@ -681,6 +681,30 @@ def copy(self, copy_data=False):
...
+    def slice_by_energy(self, energy_min, energy_max, sum_over_energy_groups=False):
+        """Copy models and slice TemplateNPredModel in energy range
+
+        Parameters
+        ----------
+        energy_min, energy_max : `~astropy.units.Quantity`
+            Energy bounds of the slice
+        sum_over_energy_groups : bool
+            Whether to sum over the energy groups or not.
+
```
bkhelifi, line 693: There is a missing parameter to describe, `sum_over_energy_groups'

gammapy/modeling/models/core.py
```diff
@@ -681,6 +681,30 @@ def copy(self, copy_data=False):
             models=models, covariance_data=self.covariance.data.copy()
         )
 
+    def slice_by_energy(self, energy_min, energy_max, sum_over_energy_groups=False):
+        """Copy models and slice TemplateNPredModel in energy range
```
bkhelifi, line 685:
```suggestion
        """Copy models and slice TemplateNPredModel in energy range.
```

gammapy/modeling/models/core.py
```diff
@@ -681,6 +681,30 @@ def copy(self, copy_data=False):
...
+        ----------
+        energy_min, energy_max : `~astropy.units.Quantity`
+            Energy bounds of the slice
+        sum_over_energy_groups : bool
+            Whether to sum over the energy groups or not.
+
+        Returns
+        -------
+        models : `Models`
+            Sliced models
```
bkhelifi, line 697:
```suggestion
            Sliced models.
```

gammapy/modeling/models/cube.py
```diff
@@ -1093,6 +1093,50 @@ def stack(self, other, weights=None, nan_to_num=True):
...
 
+    def slice_by_energy(self, energy_min=None, energy_max=None, name=None):
+        """Select and slice model template in energy range
+
+        Parameters
+        ----------
+        energy_min, energy_max : `~astropy.units.Quantity`
+            Energy bounds of the slice.
+        name : str
+            Name of the sliced model.
```
bkhelifi, line 1104: Precise the default values for parameters...

gammapy/modeling/models/core.py
```diff
@@ -681,6 +681,30 @@ def copy(self, copy_data=False):
             models=models, covariance_data=self.covariance.data.copy()
         )
 
+    def slice_by_energy(self, energy_min, energy_max, sum_over_energy_groups=False):
```
bkhelifi, line 684: A naive question: this function slices only the `TemplateNPredModel'. Should we have a more explicit function name?

gammapy/modeling/models/cube.py
```diff
@@ -1093,6 +1093,50 @@ def stack(self, other, weights=None, nan_to_num=True):
...
+
+        Returns
+        -------
+        model : `TemplateNpredModel`
+            Sliced Model
```
bkhelifi, line 1109:
```suggestion
            Sliced Model.
```

gammapy/modeling/models/cube.py
```diff
@@ -1093,6 +1093,50 @@ def stack(self, other, weights=None, nan_to_num=True):
...
+        """
+        name = make_name(name)
+
+        energy_axis = self.map._geom.axes["energy"]
+
+        if energy_min is None:
+            energy_min = energy_axis.bounds[0]
+
+        if energy_max is None:
+            energy_max = energy_axis.bounds[1]
```
bkhelifi, line 1120: As this is not tested, I have a doubt... it is bounds[1] or bounds[-1]?

gammapy/modeling/models/cube.py
```diff
@@ -1093,6 +1093,50 @@ def stack(self, other, weights=None, nan_to_num=True):
...
+
+        if name is None:
+            name = self.name
+
+        energy_min, energy_max = u.Quantity(energy_min), u.Quantity(energy_max)
+
+        group = energy_axis.group_table(edges=[energy_min, energy_max])
+
+        is_normal = group["bin_type"] == "normal   "
```
bkhelifi, line 1129: "normal   ": this empty space is really what is expected??

gammapy/estimators/points/tests/test_sed.py
```diff
@@ -390,6 +391,21 @@ def test_run_map_pwl_reoptimize(fpe_map_pwl_reoptimize):
...
+    dataset = datasets[0]
+    models = Models(dataset.models)
+    model = TemplateNPredModel(dataset.background, datasets_names=[dataset.name])
+    models.append(model)
+    dataset.models = models
+
+    fpe.run(dataset)
+    fpe.sum_over_energy_groups = True
+    fpe.run(dataset)
+
```
bkhelifi, line 407: No numerical test is needed here? We test just that it runs and not that it runs well?

bkhelifi (changes requested):
Thanks @QRemy ... I have put some small comments

gammapy/modeling/models/cube.py
QRemy: I copied that part of the code from datasets.slice_by_energy so I expect yes

gammapy/modeling/models/cube.py
```diff
@@ -1093,6 +1093,50 @@ def stack(self, other, weights=None, nan_to_num=True):
         self.spectral_model.norm.value = 1
         self.spectral_model.tilt.value = 0
 
+    def slice_by_energy(self, energy_min=None, energy_max=None, name=None):
+        """Select and slice model template in energy range
+
+        Parameters
+        ----------
+        energy_min, energy_max : `~astropy.units.Quantity`
+            Energy bounds of the slice.
```
QRemy, line 1102:
```suggestion
            Energy bounds of the slice.  Default is None.
```

gammapy/modeling/models/cube.py
QRemy, line 1104:
```suggestion
            Name of the sliced model.  Default is None.
```

gammapy/modeling/models/cube.py
QRemy: it is equivalent

gammapy/modeling/models/cube.py
```diff
@@ -1093,6 +1093,50 @@ def stack(self, other, weights=None, nan_to_num=True):
         self.spectral_model.norm.value = 1
         self.spectral_model.tilt.value = 0
 
+    def slice_by_energy(self, energy_min=None, energy_max=None, name=None):
+        """Select and slice model template in energy range
+
+        Parameters
+        ----------
+        energy_min, energy_max : `~astropy.units.Quantity`
+            Energy bounds of the slice.  Default is None.
```
QRemy, line 1102:
```suggestion
            Energy bounds of the slice. Default is None.
```

gammapy/modeling/models/core.py
```diff
@@ -681,6 +681,30 @@ def copy(self, copy_data=False):
...
+        """Copy models and slice TemplateNPredModel in energy range.
+
+        Parameters
+        ----------
+        energy_min, energy_max : `~astropy.units.Quantity`
+            Energy bounds of the slice
+        sum_over_energy_groups : bool
+            Whether to sum over the energy groups or not.
```
QRemy, line 692:
```suggestion
            Whether to sum over the energy groups or not. Default is False.
```

gammapy/estimators/points/tests/test_sed.py
QRemy: added

gammapy/modeling/models/core.py
QRemy: I would says no to be consistent with dataset.slice_by_energy

bkhelifi (approved):
Thanks a lot @QRemy .... It looks good for me

gammapy/modeling/models/core.py
registerrier, line 684: This function should most likely not be public as it is an implementation detail specific to `TemplateNPredModel`. I would rather have it either private or a stand-alone function.

PR #3885, opened 2022-04-19
Add TemplateNpredModel auto-write on models.write()

QRemy:
Write TemplateNpredModel automatically on models.write().

Requires gammapy-data PR :
https://github.com/gammapy/gammapy-data/pull/25


registerrier (approved):
Thanks @QRemy . Looks good. No comment from my side.

QRemy:
> Side point: Should we add this TemplateNPredModel to our model gallery?

No because we will very likely change it in future (for now I mostly used for performance, caching npred to disk).

PR #3289, opened 2021-03-28
Restructure docs

adonath:
**Description**

This pull request restructures the docs and includes the following changes:
- Introduce sub-pages for overview:
    - gamma-ray data analysis
    - package structure
    - dependencies
- Remove separate "installation" section and move everything in "getting started" with the following sub-pages:
    - Quickstart
    - Installation
    - Using Gammapy
    - Troubleshooting
- Rename the "Getting Started" menu section to just "User guide"
- Move back the changelog and list of releases to the "user guide" section
- A few minor improvements, clarifications of section titles
    

**Dear reviewer**

PR #3886, opened 2022-04-20
Improve docs get started

adonath:
**Description**

This pull request is a follow up to #3289. It includes the following changes:
- Add the tutorials back to the main menu (I think it should get more visibility, it's easily the most important part of the docs)
- Adapt the links on the main page to avoid some duplication
- Adapt the getting started page to include a bit more info and some simple 1D/2D/3D analysis illustrations
- Make the Twitter button white
- Change section title from "Changelog" to "Release Notes", this is consistent with what Scipy / Numpy / Pandas etc. do.

A preview (without notebooks) is available here:
https://adonath.github.io/gammapy-docs-preview


adonath:
Thanks @registerrier, yes I  remember this was an open comment from #3289. I can add a new panel for this...

In addition, the specific issue with regions is already addressed correctly in https://github.com/gammapy/gammapy/blob/master/setup.cfg#L35.

adonath:
@registerrier I tried adding an additional panel for the conda environment install and I'm not very happy (visually) with the result:
Screen Shot 2022-04-22 at 9 43 18 AM
Adding another panel leaves too many options for the simple "getting started", so it misses the point.  

As a compromise we could add a note like:
Screen Shot 2022-04-22 at 9 44 44 AM

Or if we decide the conda environment install is more important, than we should remove the pip and simple conda panels.

facero:
With the setup.cfg modified, does this imply that the `conda install gammapy` will install the 0.5 version of regions ?
I see that this was modified 29days ago but we still had the issue during ISAPP which took place 15d ago.

Another question is should we keep the pip install panel ?
Installing via pip (without any virtual env) is maybe not recommended nowadays no (or at least I would not show it in a top panel) ?
I think the predefined env is the best option for a wide audience now.

(Experts can then do the way they like anyway)

adonath:
> With the setup.cfg modified, does this imply that the conda install gammapy will install the 0.5 version of regions ?

Yes, but only for the next release. I think a release of an upstream dependency can always break the current version of Gammapy. I guess the correct way to deal with this in future is to make a bug fix release for Gammapy, restricting the dependency. This specific issue is completely independent of `conda` or `pip` ( we could also provide a `requirements-v0.19.txt` file with fixed dependency versions and the issue would not appear...)

> Another question is should we keep the pip install panel ?
Installing via pip (without any virtual env) is maybe not recommended nowadays no (or at least I would not show it in a top panel) ?

I'm not sure this is true. Python has a builtin virtual environment management system, together with `pip` this is basically equivalent to conda functionality for Python packages (see https://docs.gammapy.org/0.19/getting-started/install.html#using-virtual-environments). Many projects make use of this for development (astropy as well) e.g. together with the `tox` tool as well.

**Disclaimer: this is my strict expert opinion here...**
I fully agree the use of virtual environments is recommended and the preferred way of working with any software. However the creation of a virtual environment is user and project specific, **not software specific**. The logic is that a virtual environment is created first and then the required software is installed into it, which would include Gammapy, but other (user-specific) packages as well. This means in practice inexperienced users will start from the `gammapy-v0.18` environment and then update `matplotlib`, because they need a new feature, install `seaborn` because it makes nicer plots, install `scikit-image` because they need some algorithm and modify the environment as they like. 

Of course the pre-defined environment files are convenient and work nicely for tutorials, but long-term they are not a good way for users to work with Gammapy. They should always define a new environment for any project they work on. The logic from the user side should not be having `gammapy-v0.18.2` or `gammapy-v0.19` environments, but `my-j1808-analysis` (which includes gammapy v0.18.2) and `my-agn-ml-classification` (which includes gammapy v0.19) environments. And this is the best practice we should teach. So I would even favor if we showed users how to create new environments themself (teaching them to fish, rather then giving them one...).

adonath:
More specific, my preferred solution would be:
- Keep the conda and pip panels
- Rename "Tutorial Dataset Download" to `Tutorials Setup` or similar and introduce a tutorial specific setup which includes the pre-defined conda environment in the download section below
- Rename the pre-defined Gammapy conda environment to e.g. "gammapy-v0.19-tutorials" to make clear it is project specific and not specific to the software

adonath:
I implemented the following changes now:
- Keep the pip / conda panels
- Rename to `Quickstart Setup`
- Add a hint on the recommendation of using virtual environment to the beginning go the install page
- Introduce a separate  more detailed page for working with virtual environments

Issue #3534, opened 2021-10-01
Implement `__eq__` for IRFs


maxnoe:
**Is your feature request related to a problem? Please describe.**

Currently `IRF` does not implement `__eq__`, so it just tests object identity. Would make e.g. writing unit tests easier.

**Describe the solution you'd like**
Implement `__eq__` for IRF.

AtreyeeS:
Hi @maxnoe What would you like to test the equivalency for? `data`, `unit`, and the `axes` like in maps, or also some extra info like meta etc?

adonath:
I agree this is a useful feature and will lead to nice and readable code. In practice one might need to allow for a certain tolerance, which is the downside of the `__eq__`, where one cannot pass additional parameters. On the other hand one can certainly work with a good default choice...

bkhelifi:
We have 4 types of IRFs, so we might need (at least) 4 __eq__ for each type.
For `EffectiveAreaTable2D`, we can compare the 2 axis and the data
For `BackgroundIRF`, the axis and the data.
For the PSF classes, one might need a different comparison for each PSF class.
For `EnergyDispersion2D`, one can compare the 3 axis and the data...
Do I analyse correctly the 4 tasks, one per IRF type?

adonath:
@bkhelifi In practice I don't think we need a different implementation for the different IRF types, except maybe for the parametric PSFs. One pattern that could be useful is to add either a `IRF.compare(other, atol, rtol)` or maybe `IRF.allclose(r, atol, rtol)`. This would allow for configurable comparison and maybe provide a more meaningful printout of what is different. The `__eq__` method then just calls into `.compare()` with a fixed pre-defined (class level defined) precision.

adonath:
And if we implement it for `IRF`, we should do it for `Map` as well.

maxnoe:
@adonath with `Map` I would be probably expect something else. A map is something like an ndarray with named axes, right?

So I would expect it to behave like it, i.e. do such comparisons element wise and return a boolean map? Raise an error if map axes differ?

registerrier:
> @adonath with `Map` I would be probably expect something else. A map is something like an ndarray with named axes, right?
> 
> So I would expect it to behave like it, i.e. do such comparisons element wise and return a boolean map of the same dimensions? Raise an error if map axes differ?

@adonath it actually already exists on `Map` and behaves exactly as @maxnoe suggests (raise an error if map geometries differ).

[...]

PR #3887, opened 2022-04-21
Implement __eq__ for DL3 IRFs

AtreyeeS:
Addresses  #3534

### DIFF
--- a/gammapy/irf/core.py
+++ b/gammapy/irf/core.py
@@ -554,8 +554,30 @@ class IRF(metaclass=abc.ABCMeta):
             data=data, axes=axes, meta=self.meta.copy(), unit=self.unit
         )
 
-    def compare(self, other, **kwargs):
-        """Compare two IRFs for equivalency
+    def axes_all_close(self, other):
+        """Compare two IRF types and axes
+        Parameters
+        ----------
+        other: the irf to compare against
+            `gammapy.irfs.IRF`
+        kwargs: dict
+                keywords passed to `numpy.allclose`
+
+        Returns
+        -------
+        state: bool
+        """
+        if not isinstance(other, self.__class__):
+            return False
+
+        for x1, x2 in zip(self.axes, other.axes):
+            if x1 != x2:
+                return False
+
+        return True
+
+    def data_allclose(self, other, **kwargs):
+        """Compare two data IRFs for equivalency
 
         Parameters
         ----------
@@ -570,19 +592,12 @@ class IRF(metaclass=abc.ABCMeta):
         """
 
         kwargs.setdefault("rtol", 1e-3)
-
-        if not isinstance(other, self.__class__):
+        if self.data.shape != other.data.shape:
             return False
-
-        for x1, x2 in zip(self.axes, other.axes):
-            if x1 == x2:
-                continue
-            return False
-
         return np.allclose(self.quantity, other.quantity, **kwargs)
 
     def __eq__(self, other):
-        return self.compare(other)
+        return bool(self.axes_all_close(other) * self.data_allclose(other))
 
 
 class IRFMap:
--- a/gammapy/irf/core.py
+++ b/gammapy/irf/core.py
@@ -559,14 +559,15 @@ class IRF(metaclass=abc.ABCMeta):
 
         Parameters
         ----------
-        other: the irf to compare against
+        other : the irf to compare against
             `gammapy.irfs.IRF`
-        kwargs: dict
+        kwargs : dict
                 keywords passed to `numpy.allclose`
+                The default tolerance is rtol=1e-3
 
         Returns
         -------
-        state: bool
+        state : bool
         """
 
         kwargs.setdefault("rtol", 1e-3)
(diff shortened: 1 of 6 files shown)


maxnoe (commented):
Why the indirection using a `compare` method? Why not just directly implement `__eq__`?

adonath:
@maxnoe I think the main argument is that a `.compare()` method can take arguments such as relative and absolute tolerance, which is probably required, depending on the use case. However we should maybe think a bit about a uniform pattern. Similar methods are here:
- https://github.com/gammapy/gammapy/blob/master/gammapy/maps/axes.py#L177
- https://github.com/gammapy/gammapy/blob/master/gammapy/maps/axes.py#L153
- https://github.com/gammapy/gammapy/blob/master/gammapy/maps/axes.py#L2276
- https://github.com/gammapy/gammapy/blob/master/gammapy/maps/axes.py#L2912
- https://github.com/gammapy/gammapy/blob/master/gammapy/maps/wcs/geom.py#L1129
- https://github.com/gammapy/gammapy/blob/master/gammapy/maps/wcs/geom.py#L1107

Maybe a method like `IRF.is_allclose(other, atol, rtol)`  and `IRF.axes.is_allclose(other, atol, rtol)` is more flexible for testing. `__eq__` would then be implemented as `self.is_allcose(other, atol=1e-12, rtol=1e-6, atol_axes=, rtol_axes=), with some default tolerance.

gammapy/irf/core.py
```diff
@@ -546,6 +554,31 @@ def pad(self, pad_width, axis_name, **kwargs):
...
+
+        kwargs.setdefault("rtol", 1e-3)
+        if self.data.shape != other.data.shape:
+            return False
+        return np.allclose(self.quantity, other.quantity, **kwargs)
+
+    def __eq__(self, other):
+        class_eq = isinstance(other, self.__class__)
+        axes_eq = self.axes == other.axes
+        return bool(class_eq * axes_eq * self.data_allclose(other))
```
registerrier, line 580: why not using `and` ?

gammapy/irf/core.py
```diff
@@ -546,6 +554,31 @@ def pad(self, pad_width, axis_name, **kwargs):
...
 
+    def data_allclose(self, other, **kwargs):
+        """Compare two data IRFs for equivalency
+
+        Parameters
+        ----------
+        other: the irf to compare against
+            `gammapy.irfs.IRF`
+        kwargs: dict
+                keywords passed to `numpy.allclose`
```
registerrier, line 565: maybe specify the default `rtol`?

gammapy/irf/core.py
```diff
@@ -546,6 +554,31 @@ def pad(self, pad_width, axis_name, **kwargs):
...
+
+        Returns
+        -------
+        state: bool
```
registerrier, line 569: space missing: `state :  bool`

gammapy/irf/psf/parametric.py
```diff
@@ -248,6 +248,29 @@ def evaluate(self, rad, **kwargs):
...
+        Parameters
+        ----------
+        other: the irf to compare against
+                    `gammapy.irfs.IRF`
+        kwargs: dict
+                keywords passed to `numpy.allclose`
+
+        Returns
+        -------
+        state: bool
```
registerrier, line 263: space missing.

gammapy/irf/psf/parametric.py
```diff
@@ -248,6 +248,29 @@ def evaluate(self, rad, **kwargs):
...
+        """
+
+        state = True
+        for key in self.quantity.keys():
+            if self.quantity[key].shape != other.quantity[key].shape:
```
registerrier, line 268: Here you might simply return False

gammapy/irf/psf/parametric.py
```diff
@@ -248,6 +248,29 @@ def evaluate(self, rad, **kwargs):
...
+                state = False
+            state *= np.allclose(self.quantity[key], other.quantity[key], **kwargs)
```
registerrier, line 270: Here also you can only return False as soon as the `allclose` is False.

registerrier (approved):
I have left some small suggestions.

gammapy/irf/core.py
```diff
@@ -546,6 +554,32 @@ def pad(self, pad_width, axis_name, **kwargs):
...
+
+        kwargs.setdefault("rtol", 1e-3)
+        if self.data.shape != other.data.shape:
+            return False
+        return np.allclose(self.quantity, other.quantity, **kwargs)
+
+    def __eq__(self, other):
+        class_eq = isinstance(other, self.__class__)
+        axes_eq = self.axes == other.axes
+        return bool(class_eq & axes_eq & self.data_allclose(other))
```
registerrier, line 581: `&` is a bitwise operator. Here it should work but I think the correct syntax should be:
`class_eq and axes_eq and self.data_allclose(other)`

registerrier (approved):
Thanks @AtreyeeS

Issue #3890, opened 2022-04-22
Wrong coordinates of events when producing a counts map


SimoneMender:
**Gammapy version**
0.20.dev726+g57be15f25

**Bug description**
When producing a counts map, the coordinates of the events seem to be shifted to the source position (current master branch):
![counts_gammapy_0 20 dev726+g57be15f25](https://user-images.githubusercontent.com/18547482/164730999-e1fe6fce-8e18-4f75-b3d6-136f552dfcb7.png)

**Expected behavior**
Background counts are expected to peak at the pointing positions (which is the case for gammapy 0.19):
![counts_gammapy_0 19](https://user-images.githubusercontent.com/18547482/164730945-fd9daa89-b269-4fad-b427-58f383e303bf.png)

**To Reproduce**
```
from gammapy.data import DataStore
from gammapy import __version__
import numpy as np
from astropy.coordinates.angle_utilities import angular_separation
import astropy.units as u
from astropy.coordinates import SkyCoord
from gammapy.makers import MapDatasetMaker
from gammapy.datasets import MapDataset, Datasets
from gammapy.maps import MapAxis, WcsGeom
import matplotlib.pyplot as plt

data_store = DataStore.from_dir(".")
observations = data_store.get_observations(required_irf=['aeff', 'edisp'])

ra_pnt = [obs.events.table.meta['RA_PNT'] for obs in observations] * u.deg
dec_pnt = [obs.events.table.meta['DEC_PNT'] for obs in observations] * u.deg
ra_obj = [obs.events.table.meta['RA_OBJ'] for obs in observations] * u.deg
dec_obj = [obs.events.table.meta['DEC_OBJ'] for obs in observations] * u.deg

separation = angular_separation(ra_pnt, dec_pnt, ra_obj, dec_obj)
mask = np.abs(separation - 1.8 * u.deg) < 0.05 * u.deg
observations18deg = data_store.get_observations(data_store.obs_table['OBS_ID'][mask], required_irf=["aeff", "edisp"])

source_pos = SkyCoord.from_name("Crab")

e_reco = MapAxis.from_energy_bounds(0.07, 2, 1, unit="TeV", name="energy")
geom = WcsGeom.create(skydir=source_pos, axes=[e_reco], width=4, binsz=0.1)
maker = MapDatasetMaker(selection=["counts"]) 

datasets = []

for obs in observations18deg:
    dataset = MapDataset.create(geom, name=str(obs.obs_id))
[...]

PR #3891, opened 2022-04-22
Add notes about order of observations to docstrings

maxnoe:
**Description**

The issue described in #3890 was coming from the *wrong* assumption that `DataStore.get_observations` returns the observations in the same order as they are in `DataStore.obs_table`. 

By using `np.unique` which returns unique values in *sorted* order, `DataStore.get_observations` now returns observations ordered by obs_id, not in the order in the obs table.

I think this behavior is ok, but should be documented.

**Dear reviewer**

### DIFF
--- a/gammapy/data/data_store.py
+++ b/gammapy/data/data_store.py
@@ -76,7 +76,7 @@ class DataStore:
 
     @property
     def obs_ids(self):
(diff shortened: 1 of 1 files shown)


adonath (approved):
Thanks @maxnoe!

Issue #3889, opened 2022-04-22
Support circular region for fixed RADMAX


registerrier:
It is no longer possible to run the joint crab analysis in a consistent manner with the dev version. This is because the `SpectrumDatasetMaker` requires a `PointSkyRegion` to work. 

This is not necessary and a simple `CircleSkyRegion` could be accepted if its radius matches the fixed RADMAX value.

See https://github.com/gammapy/gammapy-benchmarks/issues/124

registerrier:
Solved by #3892 .

PR #3892, opened 2022-04-23
Allow for regular reflected region analysis with fixed rad_max IRFs

registerrier:
**Description**

This pull request solves #3889 by introducing a `RadMax2D.check_geom` that validates the region shape and size w.r.t. to the rad_max. For now, it only allows circular region with identical radius if rad_max is constant and point sky region otherwise. 
A test is then performed in `ReflectedRegionsMaker.run` to check whether the input is valid and raise exception otherwise. The check is basically:
- if no rad_max, OK
- if rad_max and if reflected regions, OK only if rad_max and region agree
- if rad_max and not reflected region, valid only if point sky region

With this the joint crab validation runs.
**Dear reviewer**

### DIFF
--- a/gammapy/makers/background/reflected.py
+++ b/gammapy/makers/background/reflected.py
@@ -1,6 +1,5 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 import logging
-from numpy.testing import assert_allclose
 import numpy as np
 from itertools import combinations
 from abc import ABCMeta, abstractmethod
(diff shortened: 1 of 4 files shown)


gammapy/makers/background/reflected.py
```diff
@@ -467,21 +467,37 @@ def make_counts_off(self, dataset, observation):
...
-
-        if is_point_like and not geom.is_all_point_sky_regions:
-            raise ValueError(
-                "Must use PointSkyRegion on region in point-like analysis,"
-                f" got {type(geom.region)} instead"
-            )
+        is_point_like = False
+
+        if observation.rad_max is not None:
+            if isinstance(self.region_finder, ReflectedRegionsFinder):
```
adonath, line 473: I think the additional check for the `ReflectedRegionsFinder` is not needed, it should work fine with the `WhobbleRegionsFinder` as well, as long as the circle is consistent with the single rad max value.

Issue #3683, opened 2021-12-10
Add peek methods on IRFMaps


registerrier:
Most IRF methods have `peek` methods implemented. `EDispMap` and `PSFMap` are lacking such methods. Yet they would be useful, in particular to have a rapid view from the `MapDataset`.

The issue is to determine what is the interesting plots to provide. Most likely the average containment radius as a function of energy at the center for the `PSFMap` . Possibly the edisp matrix for `EDispMap`.
Some visualization of the map itself would be good. Possibly the associated exposure map or the containment radius map?

Suggestions @adonath @QRemy @AtreyeeS @LauraOlivera ?

adonath:
Solved by #3894.

PR #3894, opened 2022-04-25
Add peek method to IRFMaps and unify peek methods

LauraOlivera:
This pull request:

1.  addresses #3683 and adds a peek method for the IRFMaps, namely psf and edisp. Also adds an example of this to the 3D analysis tutorial.
2. Fixes a bug causing the axis tick labels of quantities with degree units to show 0º if the value was below 1
3. Unifies the docstring and argument of most peek method (except events.peek)
4. Fixes a bug in `EDispMap.get_edisp_kernel()` which had `energy_axis` as optional input but not giving it lead to an error.

The new peek methods look like:
image

and for the edisp
image

### DIFF
--- a/gammapy/irf/psf/map.py
+++ b/gammapy/irf/psf/map.py
@@ -6,7 +6,6 @@ from gammapy.maps import Map, MapAxis, MapCoord, WcsGeom
 from gammapy.modeling.models import PowerLawSpectralModel
 from gammapy.utils.gauss import Gauss2DPDF
 from gammapy.utils.random import InverseCDFSampler, get_random_state
-from matplotlib.ticker import FormatStrFormatter
 from ..core import IRFMap
 from .core import PSF
 from .kernel import PSFKernel
@@ -425,6 +424,7 @@ class PSFMap(IRFMap):
 
         """
         import matplotlib.pyplot as plt
+        from matplotlib.ticker import FormatStrFormatter
 
         ax = plt.gca() if ax is None else ax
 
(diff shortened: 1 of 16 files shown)


adonath:
Thanks a lot @LauraOlivera! 

Concerning the `.peek()` methods I don't really have a strong opinion. I think typically they are not meant to be re-used for paper plots and to be customized. So I think any scheme we introduce will work, especially because users can always get the current figure using `plt.gcf()` or customize the figure beforehand like `plt.figure(figsize=(), ...)`.

In this PR you introduced consistently the `figsize` parameter, which I think is fine. We need it internally to set a good default figure size anyway and it is the most important parameter to configure for users...

adonath (commented):
Thanks @LauraOlivera, please rebase, otherwise looks good to me.

LauraOlivera:
I tried to rebase but I am not 100% that it worked as I intended... :sweat:

gammapy/irf/psf/map.py
```diff
@@ -6,6 +6,7 @@
 from gammapy.modeling.models import PowerLawSpectralModel
 from gammapy.utils.gauss import Gauss2DPDF
 from gammapy.utils.random import InverseCDFSampler, get_random_state
+from matplotlib.ticker import FormatStrFormatter
```
adonath, line 9: Matplotlib is still an optional dependency, so this has to be a delayed import in the method where it's used. This also leads to the remaining test fail...

adonath (commented):
Thanks @LauraOlivera, I have left one minor comment. Otherwise looks good!

gammapy/irf/psf/map.py
LauraOlivera: ah very good point, I missed that. Fixed now!

adonath (approved):
Thanks @LauraOlivera, no further comments from my side!

PR #3895, opened 2022-04-26
Add MapAxes .__eq__, .__neq__, and .copy() 

QRemy:
Add MapAxes .__eq__, __neq__, and .copy() (as suggested in #3862)


adonath (approved):

PR #3896, opened 2022-04-26
Separate handling of safe mask default thresholds

adonath:
**Description**

When looking at the rad max tutorial I noticed that the default low energy thresholds are not taken into account. This is because the IRFs only define the `LO_THRESH` keyword. However Gammapy tries to access lower and upper threshold in the same try/except block. So if the upper threshold is missing, the lower threshold will not be applied either. This PR fixes the behavior to handle both thresholds independently.


AtreyeeS (approved):
Thanks @adonath ! No comments from my side

adonath:
Thanks @AtreyeeS!

PR #3898, opened 2022-04-26
Add Map.resample method

QRemy:
Add Map.resample method (as suggested in #3862)

### DIFF
--- a/gammapy/maps/tests/test_core.py
+++ b/gammapy/maps/tests/test_core.py
@@ -493,7 +493,7 @@ def test_map_plot_mask():
         mask.plot_mask()
 
 
-def test_resample():
+def test_resample_wcs_wcs():
     npix1=3 
     geom1 = WcsGeom.create(npix=npix1, frame="icrs")
     map1 = Map.from_geom(geom1, data=np.eye(npix1))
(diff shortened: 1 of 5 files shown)


adonath (commented):
Thanks a lot @QRemy! As it seems to work already:  can you add one more unit test for the `HpxNDMap`?

gammapy/maps/tests/test_core.py
```diff
@@ -491,3 +491,35 @@ def test_map_plot_mask():
...
+                           npix=7,
+                           frame="galactic"
+                           )
+    map2 = Map.from_geom(geom2)
+    map2.resample(geom1, weights=map1.quantity, preserve_counts=False)
+    assert_allclose(np.sum(map2*geom2.solid_angle()),
+                    np.sum(map1*geom1.solid_angle()),
+                    rtol=1e-3
+                    )
+
```
adonath, line 512:
```suggestion
def test_resample_wcs_hpx():
```

gammapy/maps/tests/test_core.py
```diff
@@ -491,3 +491,35 @@ def test_map_plot_mask():
 
     with mpl_plot_check():
         mask.plot_mask()
+
+
+def test_resample():
```
adonath, line 496:
```suggestion
def test_resample_wcs_wcs():
```

adonath (approved):

PR #3899, opened 2022-04-27
Refactor Map.iter_by_image

adonath:
**Description**

This pull request refactors the `.iter_by_image` mechanism for `Map`:
- Rename the existing `.iter_by_image` to `.iter_by_image_data`
- Rename the existing `.iter_by_axis` to `.iter_by_axis_data`
- Re-implement `.iter_by_image` to return a `Map` object
- Re-implement `.iter_by_axis` to return a `Map` object

This is a backwards incompatible change, but I think it is replaced by a much cleaner and more practical solution.

### DIFF
--- a/docs/tutorials/api/maps.ipynb
+++ b/docs/tutorials/api/maps.ipynb
@@ -851,7 +851,7 @@
    "source": [
     "### Iterating by image\n",
     "\n",
-    "For maps with non-spatial dimensions the `~Map.iter_by_image` method can be used\n",
+    "For maps with non-spatial dimensions the `~Map.iter_by_image_data` method can be used\n",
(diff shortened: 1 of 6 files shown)


QRemy (approved):
Thanks @adonath, no objection from my side even if this is not backward compatible. I guess that won't affect much users as this feature is likely used only internally, and was shown only in the maps api tutorial and not in the analyses tutorials.

AtreyeeS (approved):
Thanks @adonath ! Since `yield` is less commonly known amongst users, it might be helpful to add `Examples` in the doc-strings at some point.

adonath:
Thanks @AtreyeeS , I'll include examples in the docstrings a follow up PR later this week.

Issue #3792, opened 2022-02-10
Save the center of OFF regions, calculated by WobbleRegionsFinder with PointSkyRegions, properly


chaimain:
**Is your feature request related to a problem? Please describe.**
Following the progress in PR #3733, trying with LST-1 DL3 data with RAD_MAX_2D HDU, and using the modified spectral analysis notebook PR #3712, `plot_spectrum_datasets_off_regions()` cannot plot the correct OFF regions.
As @adonath pointed out that the new implementation of this `RegionsFinder` does not store the OFF regions as `RegionGeom` of the `counts_off` object.

**Describe the solution you'd like**
Make the `RegionGeom` handle energy-dependent regions.

**Describe alternatives you've considered**
Save the correct center of the OFF regions, thus generated, so that the `plot_spectrum_datasets_off_regions` can at least plot those correctly.

**Additional context**
![DL4_off_regions](https://user-images.githubusercontent.com/36500590/153431324-1f2b96e4-2afa-4412-ac4f-19386712c938.png)
The current usage of the function, while using LST-1 standard Crab dataset

chaimain:
I could replicate with Max's script using my LST DL3 files, and updated gammapy master.
With 2 OFF regions (just for different cross-check),
![Crab_off_positions_en_dep](https://user-images.githubusercontent.com/36500590/157700507-593674e3-9b56-4255-960d-e71cbfa88378.png)

adonath:
Thanks @maxnoe! Just to confirm, plotting `PointSkyRegion` does not work correctly in Gammapy:
```python3
from gammapy.maps import RegionGeom
from regions import PointSkyRegion
from astropy import units as u
from astropy.coordinates import SkyCoord

centers = SkyCoord([1, 1, -1, -1] * u.deg, [-1, 1, 1, -1] * u.deg)
regions = [PointSkyRegion(_) for _ in centers]
geom = RegionGeom.from_regions(regions=regions)

ax = geom.plot_region()
```

The reason is a bit weird and possibly a bug in `matplotlib`. So the following works:
```python3
import matplotlib.pyplot as plt
from matplotlib.collections import PatchCollection
from regions import PointPixelRegion, PixCoord

ax = plt.subplot()

point = PointPixelRegion(PixCoord(0.5, 0.5))

artist = point.as_artist()

ax.add_artist(artist)
```

While going through the `PatchCollection` (which is the current implementation in Gammapy) does not:
```python3
import matplotlib.pyplot as plt
[...]

PR #3900, opened 2022-04-27
Support plotting of point regions

adonath:
**Description**

This pull request adds the support to plot point regions with `RegionGeom.plot_region()`. Plotting point regions with regions / matplotlib is supported by a `Line2D` objects. Which requires different arguments than a patch.  For this reason I introduced a `kwargs_points` argument. This is consistent with `Models.plot_regions()`.

I adapted the rad max tutorial to show the plot of the off positions. Here is a preview:

image

This PR resolves the remaining to do's in #3792.

### DIFF
--- a/docs/tutorials/analysis/1D/spectral_analysis_rad_max.ipynb
+++ b/docs/tutorials/analysis/1D/spectral_analysis_rad_max.ipynb
@@ -58,14 +58,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "metadata": {
-    "execution": {
-     "iopub.execute_input": "2022-01-21T12:32:11.095008Z",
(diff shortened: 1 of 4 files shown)


cosimoNigro:
Thanks a lot for completing the notebook @adonath.
I am not sure if it's a problem of my browser, but despite "hide outputs" being un-slected in the control bar I do not see ant actual output, so I don't see the nice plot of the regions you added.

adonath:
Thanks @cosimoNigro, this is expected, because the ReviewNB does not actually execute the notebooks. That's why I added the screenshot for now.

adonath:
The only way to re-create the plot would be to checkout my branch locally and run the notebook...

PR #3904, opened 2022-05-03
Fix bug on MC_ID when sampling IRF background events

fabiopintore:
This PR fixes the bug reported in #3866 and replaces the previous PR #3867 .
When sampling events from the an IRF background model, MapDatasetEventSampler is set like that the MC_ID of these events is always zero. However, this is true only in the table of sample event list, but not in the metadata where it can assume every kind of value. 

A solution is to apply an internal re-order of the model so to have always the bkg model as first model.

### DIFF
--- a/gammapy/datasets/simulate.py
+++ b/gammapy/datasets/simulate.py
@@ -7,7 +7,7 @@ from astropy.table import Table
 import gammapy
 from gammapy.data import EventList, observatory_locations
 from gammapy.maps import MapCoord
-from gammapy.modeling.models import ConstantTemporalModel
+from gammapy.modeling.models import ConstantTemporalModel, Models
 from gammapy.utils.random import get_random_state
 
 __all__ = ["MapDatasetEventSampler"]
@@ -295,8 +295,8 @@ class MapDatasetEventSampler:
         meta["CONV_DEC"] = 0
 
         for idx, model in enumerate(dataset.models):
-            meta["MID{:05d}".format(idx + 1)] = idx + 1
-            meta["MMN{:05d}".format(idx + 1)] = model.name
+            meta["MID{:05d}".format(idx)] = idx
+            meta["MMN{:05d}".format(idx)] = model.name
         meta["NMCIDS"] = len(dataset.models)
 
         # Necessary for DataStore, but they should be ALT and AZ instead!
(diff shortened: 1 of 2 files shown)


adonath:
Thanks @fabiopintore! 

I think it might the better solution to actually generate the meta data of the `MC_ID` in the `.sample_sources()` and `.sample_background()` methods instead. This should simplify implementation and make sure the info is consistent between the table and its meta data. Can you check this?

gammapy/datasets/simulate.py
```diff
@@ -88,6 +88,8 @@ def sample_sources(self, dataset):
             table = self._sample_coord_time(npred, temporal_model, dataset.gti)
             if len(table) > 0:
                 table["MC_ID"] = idx + 1
+                table.meta["MID{:05d}".format(idx + 1)] = idx + 1
+                table.meta["MMN{:05d}".format(idx + 1)] = dataset.models[idx].name
```
adonath, line 92: You can directly use:
```suggestion
                table.meta["MMN{:05d}".format(idx + 1)] = evaluator.model.name
```

gammapy/datasets/simulate.py
```diff
@@ -119,6 +121,12 @@ def sample_background(self, dataset):
         table["RA"] = table["RA_TRUE"]
         table["DEC"] = table["DEC_TRUE"]
 
+        for i, mod in enumerate(dataset.models):
+            if mod.tag[0] == "FoVBackgroundModel":
+                break
+        table.meta["MID{:05d}".format(0)] = 0
+        table.meta["MMN{:05d}".format(0)] = dataset.models[i].name
```
adonath, line 128:
```suggestion
        table.meta["MMN{:05d}".format(0)] = dataset.background_model.name
```

gammapy/datasets/simulate.py
```diff
@@ -119,6 +121,12 @@ def sample_background(self, dataset):
         table["RA"] = table["RA_TRUE"]
         table["DEC"] = table["DEC_TRUE"]
 
+        for i, mod in enumerate(dataset.models):
+            if mod.tag[0] == "FoVBackgroundModel":
+                break
```
adonath, line 126:
```suggestion

```

adonath (changes requested):
Thanks @fabiopintore, I have left some code suggestions.

gammapy/datasets/simulate.py
fabiopintore: Thanks @adonath . I've implemented all suggestions.

gammapy/datasets/simulate.py
```diff
@@ -7,7 +7,7 @@
 import gammapy
 from gammapy.data import EventList, observatory_locations
 from gammapy.maps import MapCoord
-from gammapy.modeling.models import ConstantTemporalModel
+from gammapy.modeling.models import ConstantTemporalModel, Models
```
adonath, line 10:
```suggestion
from gammapy.modeling.models import ConstantTemporalModel
```

adonath (approved):
Thanks @fabiopintore, in future we should clean up the tests for this...

Issue #3763, opened 2022-01-28
Require astropy 5.0 for gammapy 1.0


maxnoe:
**Is your feature request related to a problem? Please describe.**

With the recent release of the new LTS release of astropy, bugfix support for older versions of astropy stopped.

To not get into the situation, where gammapy users run into astropy bugs that are not supported with bugfix releases anymore, we should require astropy 5 and see that this LTS releases stays supported by gammapy for the duration of 1.0.

adonath:
Added in #3906.

PR #3906, opened 2022-05-03
Require astropy>=5.0

adonath:
**Description**

This pull request pushes the Astropy version required by Gammapy to >=5.0. This is an LTS version and the version that should be used for Gammapy v1.0.

Addresses #3763.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -182,7 +182,7 @@ jobs:
           channel-priority: true
       - name: install dependencies
         run: |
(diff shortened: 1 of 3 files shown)


adonath:
By requiring `astropy>=5.0` we also have to drop support for Python 3.7.

registerrier (dismissed):
+1 @adonath . What should be the CI grid now?

adonath:
I think we should add Python 3.10 and possibly 3.11 instead.

adonath:
@registerrier I cannot get the CI fails resolved. I'll split the PR and update the Python version first.

Issue #3662, opened 2021-11-25
[0.19] Failures on non-x86 architectures


olebole:
**Gammapy version**

0.19

**Bug description**

The recent version of gammapy fails on all Debian 64-bit platforms except x86_64. 

**Expected behavior**

All tests should pass.

**To Reproduce**

Running the CI test on a non-x86 platform, f.e. on arm64 gives:

```
______________________________ test_binary_erode _______________________________

    def test_binary_erode():
        geom = WcsGeom.create(binsz=0.02, width=2 * u.deg)
        mask = geom.region_mask("icrs;circle(0, 0, 1)")
    
        mask = mask.binary_erode(width=0.2 * u.deg, kernel="disk", use_fft=False)
        assert_allclose(mask.data.sum(), 4832)
    
        mask = mask.binary_erode(width=0.2 * u.deg, kernel="box", use_fft=True)
>       assert_allclose(mask.data.sum(), 3372)
E       AssertionError: 
E       Not equal to tolerance rtol=1e-07, atol=0
E       
E       Mismatched elements: 1 / 1 (100%)
E       Max absolute difference: 15
E       Max relative difference: 0.0044484
E        x: array(3387)
E        y: array(3372)

gammapy/maps/wcs/tests/test_ndmap.py:836: AssertionError
______________________________ test_binary_dilate ______________________________

    def test_binary_dilate():
        geom = WcsGeom.create(binsz=0.02, width=2 * u.deg)
        mask = geom.region_mask("icrs;circle(0, 0, 0.8)")
    
        mask = mask.binary_dilate(width=0.2 * u.deg, kernel="disk", use_fft=False)
        assert_allclose(mask.data.sum(), 8048)
    
        mask = mask.binary_dilate(width=(10, 10), kernel="box")
>       assert_allclose(mask.data.sum(), 9203)
E       AssertionError: 
E       Not equal to tolerance rtol=1e-07, atol=0
E       
E       Mismatched elements: 1 / 1 (100%)
E       Max absolute difference: 11
E       Max relative difference: 0.00119526
E        x: array(9214)
E        y: array(9203)

gammapy/maps/wcs/tests/test_ndmap.py:847: AssertionError
```

**Other information**

This has been observed so far on arm64, ppc64el, s390x, ppc64. It does not happen on 32-bit platforms (but they show other failures, update on https://github.com/gammapy/gammapy/issues/1993#issuecomment-979240280).

adonath:
[...]

PR #3907, opened 2022-05-03
Allow tolerance in fft binary operations

adonath:
**Description**

This pull request solves #3662, by increasing the test tolerance for now.


registerrier (dismissed):
Thanks @adonath. This looks fine. I suppose we should open an issue to properly handle fft round-off errors  in the future.

PR #3909, opened 2022-05-04
Require Python 3.8

adonath:
**Description**

See title.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -182,7 +182,7 @@ jobs:
           channel-priority: true
       - name: install dependencies
         run: |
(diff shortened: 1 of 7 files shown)


.github/workflows/ci.yml
```diff
@@ -179,15 +175,14 @@ jobs:
       - name: create and activate env
         uses: mamba-org/provision-with-micromamba@main
         with:
-          environment-file: environment-dev.yml
```
registerrier, line 182: Why is the environment file not needed here?

adonath: In general the use of the `environment-dev.yml` for the CI is not ideal, because it installs many dependencies that we don't need for testing...

PR #3910, opened 2022-05-05
Fix issue #3801

### LINKED ISSUE #3801: Change the first dataset name in `.info_table()`
When the method `.info_table()` is used in a Datasets object, the name of the first dataset in the output table is always "stacked". This rename is done even if the parameter `cumulative` of `.info_table()` is set to False.

Keep the original name of the first dataset in the returned table if `cumulative`=False and only use "stacked" when `cumulative`=True.

adonath:
**Description**

This pull request fixes #3801.


adonath:
CI fails seem unrelated.

PR #3913, opened 2022-05-06
Print minuit fit detailed infos in Fit API tutorial

QRemy:
As discussed in the help channel printing the FitResults no longer shows the detailed convergence infos from minuit, but it can be displayed printing directly the minuit object attached to the fit. This PR shows this in the Fit API notebook.


lmohrmann (approved):
Thanks!

adonath (approved):
Thanks @QRemy, CI fails are unrelated....

Issue #3868, opened 2022-03-31
Parameters stored in fit results are not stable


luca-giunti:
**Gammapy version**
0.19

**Bug description**
Parameters stored in fit results should not change if one later modifies the fitted model parameters. This is instead the case, because the parameters stored in the fit result point to the fitted model parameters. 

![image](https://user-images.githubusercontent.com/47325742/161095343-c8130e36-eb78-4583-a0f8-827012094ffc.png)

This is completely unexpected, because the fit results should contain the best fit parameters at a given analysis stage, which should always remain the same. 
A simple solution would be to make a copy of the parameters [here](https://github.com/gammapy/gammapy/blob/7e2976713facc0d443ade5279e91c4ae6b38c6c4/gammapy/modeling/fit.py#L221).

adonath:
Thanks @luca-giunti, I agree this should be the preferred behavior.

registerrier:
> I think I'd even prefer to remove the `.parameters` completely. However this is not backwards compatible...

Having a copy of the parameters provides a working mechanism to explore the values of other parameters during e.g. a `stat_scan`. In the longer run, this could provide some useful debug information for flux points too.

adonath:
> Having a copy of the parameters provides a working mechanism to explore the values of other parameters during e.g. a stat_scan. In the longer run, this could provide some useful debug information for flux points too.

OK, but in this case it should still be better to store a copy of the full model. Because e.g. 3D joint fits might have >10 "norm" parameters from background models, where it's completely unclear to which observation they belong, because this info is not stored in the `Parameters` list. I think the isolated `Parameters` list is pretty much useless and even "dangerous" to use, because parameter names are not unique. I guess at some point we will be able to write a `FitResult` to YAML, it is completely unclear what to do with the isolated parameter list in this case.

However storing a copy of the model could be convenient... and then users can access the parameters like `results.models.parameters`. I checked and I proposed this already in https://github.com/gammapy/gammapy/pull/3646, but decided to remove it again (https://github.com/gammapy/gammapy/pull/3646/commits/163133114c4f49245877ae988d579025fcd2019e), because the time was too short for a more detailed discussion.

luca-giunti:
Summary from the dev call:
- the preferred solution will be to store a shallow copy of the` Models` in the optimize results
- this can be made backward compatible by keeping the parameters property, but pointing it to the model copy parameters
- this behavior needs to be documented 
[...]

PR #3918, opened 2022-05-06
Improve fit results

luca-giunti:
**Description**

This pull request addresses issue #3868 

It is the best I could do in the limited available time. I made some quick tests and it seems to work, but I haven't had the time to play a lot with it

### DIFF
--- a/docs/tutorials/api/fitting.ipynb
+++ b/docs/tutorials/api/fitting.ipynb
@@ -101,6 +101,7 @@
     "    datasets.append(dataset)\n",
     "\n",
     "dataset_hess = Datasets(datasets).stack_reduce(name=\"HESS\")\n",
+    "datasets = Datasets(datasets=[dataset_hess])\n",
     "\n",
     "# Set model and fit range\n",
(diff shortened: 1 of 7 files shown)


registerrier (approved):
Thanks @luca-giunti . See minot inline comment.

@adonath @QRemy , can the proposed change, that removes the deep copy of the input map in `TemplateSpatialModel.__init__` be problematic? I am not sure why this was implemented this way initially.

luca-giunti:
@registerrier from my side this should be OK

registerrier (approved):
thanks @luca-giunti .

Exposing `Datasets.parameters` is probably a good idea. I think we rarely show it. 
I think it would be better to use a property for `OptimizeResults.parameters` rather than a `copy` (i.e. use `OptimizeResults.models.parameters`.

adonath (approved):
Thanks @luca-giunti!

I implemented @registerrier comment and renamed `data_deepcopy` to just `copy_data`, where the use of the verb slightly improves readability of the code.

PR #3922, opened 2022-05-09
Remove ConstantSpatialModel from list of models to be plotted

### LINKED ISSUE #3920: plotting `ConstantSpatialModel`
`Models.plot_regions()/plot_positions()` fails if it contains any `ConstantSpatialModel` (or `ConstantFluxSpatialModel`), with a rather complicated error message. I suppose this is because there is no `position` defined for these models. 

It would be nice to have this fixed before the 1.0rc release.

A quick solution can be to simply   add an exception for `ValueError ` as well here https://github.com/gammapy/gammapy/blob/a68be851d6523742627ed293aa1122c795680f8e/gammapy/modeling/models/core.py#L918.

AtreyeeS:
this fixes #3920

### DIFF
--- a/gammapy/modeling/models/core.py
+++ b/gammapy/modeling/models/core.py
@@ -147,8 +147,7 @@ class ModelBase:
         return copy.deepcopy(self)
 
     def to_dict(self, full_output=False):
-        """Create dict for YAML serialisation
-        """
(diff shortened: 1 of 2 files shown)


adonath:
Superseded by #3933.

PR #3927, opened 2022-05-10
Codemeta for the 1.0rc release

bkhelifi:
**Description**

This pull request updates the script that generate the codemeta.json file. The logics is now that a preliminary file is created with https://codemeta.github.io/codemeta-generator/ without the authors list. Then, one add the authors list stored in CITATION.cff.

This PR contains then the updated version of the codemeta.json


bkhelifi:
I did a mistake, because only 2 files should be changed.... Maybe you should reject this PR.

PR #3929, opened 2022-05-10
Change tooltip and minor edits in lightcurve simulation notebook

AtreyeeS:
I think the new image with the lightcurve and model better illustrates that fitting is performed in the notebook


adonath (approved):

Issue #3863, opened 2022-03-28
Remember angular size of regions used with the FluxProfileEstimator


LauraOlivera:
When using the `FluxProfileEstimator` the number of regions used (i.e. their size) will obviously impact the result. However, in the resulting `FluxPoints` object, only the projected distance information is kept. Meaning that this information (the solid angle size of the regions used) is lost.

This is inconvenient when one wants to, for example, compare the flux profile to the prediction of a model, because the exact regions used to create the `FluxProfileEstimator` need to be also saved alongside the profile in order to correct for the region sizes. The prediction of the model can typically be derived in smaller spatial bins to better characterize the shape, but then has to be corrected to match the flux level of the points derived from data.

**Describe the solution you'd like**
A solution could be to maybe add another `Map` to the container. For example,  `FluxPoints.angular_size` that is 0 or nans when dealing with "normal" flux points, and that contains the region size in sr for the flux profiles. The regions resulting of `make_orthogonal_rectangle_sky_regions` all have (almost exactly) the same sizes, so maybe just one number would work as well. But for generality maybe it is better to keep them all.

**Describe alternatives you've considered**
Of course another option is to save the list of regions, or directly their sizes alongside the `FluxPoints` and handle them together every time, but I would argue it is quite unpractical in terms of bookkeeping, and it would be nicer if the flux points container remembered this information.

adonath:
Thanks a lot @LauraOlivera, I fully agree this is important information to keep.

I think there is already a very good place to store this kind of information: the `RegionGeom` itself, which can directly store the `Region` objects. It then also supports plotting, computing solid angles, serialization etc. What is currently missing is the possibility to "assign" regions to bins in a `MapAxis`. Effectively it is the same problem as handling an energy dependent regions size. In one case the region varies with energy, in the second case in varies with label, bin number of maybe projected distance (this remains to be decided). A few month ago in the context for the energy dependent region size, I worked on a little prototype for this. To provide a possible starting point and base for further discussion I'll link to it here again:
- Branch: https://github.com/adonath/gammapy/tree/region_geom_array
- Notebook: https://github.com/adonath/notebooks-public/blob/master/region-geom-array.ipynb

[...]

PR #3933, opened 2022-05-10
Support regions 0.6

### LINKED ISSUE #3920: plotting `ConstantSpatialModel`

adonath:
**Description**

This pull request fixes the issues seen with regions=0.6. It includes the following changes:
- Return a `RectangleSkyRegion` with all sky specification for constant models (fixes #3920).
- Introduce a `radius_min` parameter for `make_concentric_annulus_sky_regions`
- Remove code duplication between `Models.plot_regions()` and `RegionGeom.plot_region()`


gammapy/maps/region/geom.py
```diff
@@ -121,7 +128,12 @@ def _rectangle_bbox(self):
         for region_pix in regions_pix[1:]:
             bbox = bbox.union(region_pix.bounding_box)
 
-        rectangle_pix = bbox.to_region()
+        try:
+            rectangle_pix = bbox.to_region()
+        except ValueError:
+            rectangle_pix = RectanglePixelRegion(
+                center=PixCoord(*bbox.center[::-1]), width=1, height=1
```
registerrier, line 135: Why are the width and height set to 1 in this case?

gammapy/maps/region/geom.py
```diff
@@ -683,16 +697,22 @@ def from_hdulist(cls, hdulist, format="ogip", hdu=None):
             region_hdu = hdu + "_" + region_hdu
 
         if region_hdu in hdulist:
-            region_table = Table.read(hdulist[region_hdu])
-            wcs = WcsGeom.from_header(region_table.meta).wcs
+            try:
```
registerrier, line 700: Maybe add a reminder TODO to remove the `try` once regions 0.5 is no longer supported

gammapy/utils/regions.py
```diff
@@ -197,7 +197,7 @@ def make_orthogonal_rectangle_sky_regions(start_pos, end_pos, wcs, height, nbin=
     return regions
 
 
-def make_concentric_annulus_sky_regions(center, radius_max, nbin=11):
+def make_concentric_annulus_sky_regions(center, radius_max, radius_min=1e-5 * u.deg, nbin=11):
```
registerrier, line 200: another solution would be to make the first region a disk

registerrier (approved):
Thanks @adonath . Looks good to me.

gammapy/maps/region/geom.py
adonath: I think this up to us to define: is the width of the `RegionGeom` for a point sky region zero or the size of one pixel? I decided for the latter case, because this is basically how to treated technically.

gammapy/utils/regions.py
adonath: Yes, however this might become tricky later for bookkeeping: e.g. in #3863 we would then need to handle different shapes of regions for flux profiles in addition. This might impose additional requirement on serialization of flux profiles etc. So I prefer to keep the consistent region shape for, such that the parameters of the individual regions are consistent.

gammapy/maps/region/geom.py
adonath: OK

Issue #3931, opened 2022-05-10
FluxPointEstimator not able to run without presence of a mask


cbadams2:
**Gammapy version**
0.19

**Bug description**
FluxPointEstimator appears to be failing when run on a Fermi-LAT dataset due to lack of a mask.

**Expected behavior**
Expect that a mask should be optional for LAT data.

**To Reproduce**
Issue occurs when running a code block such as the following:
```
# compute Fermi-LAT flux points
energy_edges = MapAxis.from_energy_bounds("10 GeV", "2 TeV", nbin=3).edges

flux_points_fermi = FluxPointsEstimator(
    energy_edges=energy_edges,
    source="Mrk421",
).run([datasets["Fermi"]])
```
on a Fermi dataset that was generated following roughly the procedure outlined by https://docs.gammapy.org/dev/tutorials/data/fermi_lat.html to produce the fermi dataset. 

**Other information**
Stack trace for the above code:
```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
/tmp/ipykernel_3038980/67347065.py in <module>
      2 energy_edges = MapAxis.from_energy_bounds("10 GeV", "2 TeV", nbin=3).edges
      3 
----> 4 flux_points_fermi = FluxPointsEstimator(
      5     energy_edges=energy_edges,
      6     source="Mrk421",

~/anaconda3/envs/gammapy-0.19/lib/python3.9/site-packages/gammapy/estimators/points/sed.py in run(self, datasets)
    100             zip(self.energy_edges[:-1], self.energy_edges[1:]), desc="Energy bins"
    101         ):
--> 102             row = self.estimate_flux_point(
    103                 datasets,
    104                 energy_min=energy_min,

~/anaconda3/envs/gammapy-0.19/lib/python3.9/site-packages/gammapy/estimators/points/sed.py in estimate_flux_point(self, datasets, energy_min, energy_max)
    147         if len(datasets_sliced) > 0:
    148             datasets_sliced.models = datasets.models.copy()
--> 149             return super().run(datasets=datasets_sliced)
    150         else:
    151             log.warning(f"No dataset contribute in range {energy_min}-{energy_max}")

~/anaconda3/envs/gammapy-0.19/lib/python3.9/site-packages/gammapy/estimators/flux.py in run(self, datasets)
    179         # TODO: find a cleaner way of including the npred_excess info
    180         datasets.models[self.source].spectral_model.norm.value = result["norm"]
[...]

PR #3934, opened 2022-05-10
Fix issue #3931

adonath:
**Description**

See title.


gammapy/data/observers.py
```diff
@@ -97,7 +97,7 @@
 
 # communication with ASTRI Project Manager
 observatory_locations["astri"] = EarthLocation(
-    lon="-16d30m20.99s", lat="28d17m60.0s", height="2370m"
+    lon="-16d30m20.99s", lat="28d18m00.0s", height="2370m"
```
registerrier, line 100: Ah that was the reason for the recurrent astropy warnings. Thanks!

PR #3937, opened 2022-05-11
Cleanup fixed rad max checks

adonath:
**Description**

This pull request is a follow up to #3892, to simplify a bit the logic.

### DIFF
--- a/gammapy/makers/background/reflected.py
+++ b/gammapy/makers/background/reflected.py
@@ -45,6 +45,40 @@ def are_regions_overlapping_rad_max(regions, rad_max, offset, e_min, e_max):
     return np.any(separations[np.newaxis, :] < (2 * rad_max_at_offset))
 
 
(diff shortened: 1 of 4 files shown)


gammapy/makers/background/reflected.py
```diff
@@ -45,6 +45,40 @@ def are_regions_overlapping_rad_max(regions, rad_max, offset, e_min, e_max):
...
+        True if region is a PointSkyRegion
+        False otherwise.
+    """
+    if geom.is_all_point_sky_regions:
+        valid = True
+    elif isinstance(geom.region, CircleSkyRegion) and rad_max.is_fixed_rad_max:
+        valid = np.allclose(geom.region.radius, rad_max.quantity, rtol)
+
+        if not valid:
+            raise ValueError(
```
registerrier, line 71: Here the logic is problematic too. The behavior is inconsistent between the test raising an exception for a specific set of parameters and returning a boolean flag the rest of the time.

PR #3940, opened 2022-05-11
return models.positions in same frame

AtreyeeS:
If different models were defined in different frames (`galactic` vs `icrs`, which can be the case when using template models, specially the fermi diffuse ones), `models.positions` fails because `SkyCoord(positions)` needs all `positions` in the same frame.

A very tiny fix - `models.positions` is by default in `icrs` now, can be trivially converted to galactic. Also added a test


adonath (approved):
Good catch! Thanks @AtreyeeS!

Issue #3921, opened 2022-05-09
Implement _repr_html_ for ease of use in Jupyter notebooks


evertrol:
I like to suggest an implementation of `_repr_html_` for various classes, for ease of use in Jupyter notebooks.

Currently, several of the example and tutorial notebooks use `print(some_class_instance)` to display information. More Jupyter-style would be to simply use `some_class_instance` on the last line of a cell. Currently, the latter just shows the `__repr__` of the instance, which is not that informative. The `print(some_class_instance)` instead returns `str(some_class_instance)`, which is more informative.

My suggested path to do this would be
- implement one or a few (can be increased over time) `_repr_html_` methods.
- the `_repr_html_` initially implements `return str(self)`
- over time, explicit `to_html(self)` methods can be added, which essentially return the same information as `__str__(self)`, but wrapped in some HTML tags. The explicit methods can be used elsewhere if wanted, as part of the API (compared to the more private `_repr_html_`).
  I can imagine implementations of `to_html` using a HTML table or a dictionary list (dl, dt, dd), to format the information.
- `_repr_html_` for the relevant classes can then implement `return self.to_html()` instead.

Since this is all fairly straightforward, relying on current implementations of `__str__(self)`, I'd be happy to turn this is into a PR myself.

PR #3947, opened 2022-05-12
Provide initial implementations of _repr_html_

evertrol:
This pull request provides an initial implementation of #3921.

A stub implementation of `_repr_html_` is provided for most of the Gammapy classes (often only through their base classes), for use in Jupyter notebooks (resulting in nicer representations of variables).
Later commits & PRs can implement an explicit `to_html()` method, that will be automatically picked up by `_repr_html_`. If these are not provided (as is the case as of this PR), `_repr_html_` defaults to using the `__str__` implementation, which ultimately defaults to using `__repr__` if `__str__` is not provided.

I have also added a section on `_repr_html_` and `to_html` in the development documentation, just below the "repr, str and info" section at the bottom of the `dev_howto.rst` file, where it seems to fit in nicely.

I have taken the liberty of changing a few `__repr__` implementations to `__str__` implementations, which seems to be more semantically correct; these are `__repr__` implementations that produce nicely formatted ("human readable") output, which is more suited for `__str__`. This did require changing ten unit tests as well.

I have not implemented any unit tests. This does reduce test coverage, but I felt that this would require testing `_repr_html_` for every class (not just base classes), to test that it equals `str(<some_instance>)`.
I think it makes more sense to provide unit tests for each implementation of `to_html()`, where such a unit test can verify that both `to_html()` returns the correct HTML, and `_repr_html_` returns the result of `to_html()`.

### DIFF
--- a/docs/development/dev_howto.rst
+++ b/docs/development/dev_howto.rst
@@ -884,3 +884,58 @@ Example what to put as a test::
         assert str(p).startswith('Hi')
         assert p.info(add_location=True).endswith('Heidelberg')
 
+
+Output in Jupyter notebook cells
+++++++++++++++++++++++++++++++++
+
+In addition to the standard `repr` and `str` outputs, Jupyter notebook
+cells have the option to display nicely formatted (HTML) output, using
+the IPython rich display options (other output options also exist,
+such as LaTeX or SVG, but these are less used). This requires the
+implementation of a `_repr_html_(self)` method (note: single
+underscore) in a class.
+
+By default, this method can just return the string representation of
+the object (which may already produce a nicely formatted output). That
+is often nicer than the default, which is to return the `repr` output,
+which tends to be shorter and may miss details of the object. Thus,
+the following would be a good default for a new class::
+
+    def _repr_html_(self):
+        return f'<pre>html.escape(str(self))</pre>'
+
+The surrounding `<pre>` takes care that the formatting in HTML is the
+same as that for the normal `__str__` method, while `html.escape`
+handles escaping HTML special characters (<, >, &, " and ').
+
+Note that if no `__str__` method is implemented, `__repr__` is used,
+and ultimately, the Python built-in `__repr__` method serves as the
+final fallback (which looks like `<gammapy.data.event_list.EventList
+at 0x129602550>` or similar).
+
+If more specific HTML output is preferred (for example, output
+formatted in a HTML table or list), it is best to create a separate
+`to_html(self)` method in the class, which is more explicit (and can
+be documentated as part of the API), and let `_repr_html_` call this
+method::
+
+    def _repr_html_(self):
+        return self.to_html(self)
+
+Thus very similar to `__str__` calling `info()`, with optional
+parameters if needed.
+
+To allow for both options, the following default `_repr_html_` can be
+implemented instead::
+
+    def _repr_html_(self):
+        try:
+           return self.to_html(self)
+       except AttributeError:
+           return f'<pre>html.escape(str(self))</pre>'
+
+Nearly all base classes in Gammapy implement this default
+`_repr_html_`. If a new class derives from an existing Gammapy class,
+a default implementation is not needed, since it will rely on its
+(grand)parent. As a result, for specific HTML output, only the
+`to_html` method needs to be implented for the relevant class.
--- a/gammapy/analysis/config.py
+++ b/gammapy/analysis/config.py
@@ -1,5 +1,6 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 import json
+import html
 import logging
 from collections import defaultdict
 from enum import Enum
@@ -107,6 +108,12 @@ class GammapyBaseConfig(BaseModel):
             Quantity: lambda v: f"{v.value} {v.unit}",
             Time: lambda v: f"{v.value}",
         }
+
+    def _repr_html_(self):
+        try:
+            return self.to_html()
+        except AttributeError:
+            return f"<pre>{html.escape(str(self))}</pre>"
 
 
 class SkyCoordConfig(GammapyBaseConfig):
--- a/gammapy/analysis/core.py
+++ b/gammapy/analysis/core.py
@@ -1,5 +1,6 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 """Session class driving the high level interface API"""
+import html
 import logging
 from astropy.coordinates import SkyCoord
 from astropy.table import Table
@@ -55,6 +56,12 @@ class Analysis:
         self.fit = Fit()
         self.fit_result = None
         self.flux_points = None
+
+    def _repr_html_(self):
+        try:
+            return self.to_html()
+        except AttributeError:
+            return f"<pre>{html.escape(str(self))}</pre>"
 
     @property
     def models(self):
--- a/gammapy/astro/darkmatter/profiles.py
+++ b/gammapy/astro/darkmatter/profiles.py
@@ -1,6 +1,7 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 """Dark matter profiles."""
 import abc
+import html
 import numpy as np
 import astropy.units as u
 from gammapy.modeling import Parameter, Parameters
@@ -29,6 +30,12 @@ class DMProfile(abc.ABC):
         """Call evaluate method of derived classes."""
         kwargs = {par.name: par.quantity for par in self.parameters}
         return self.evaluate(radius, **kwargs)
+
+    def _repr_html_(self):
+        try:
+            return self.to_html()
+        except AttributeError:
+            return f"<pre>{html.escape(str(self))}</pre>"
 
     def scale_to_local_density(self):
         """Scale to local density."""
--- a/gammapy/astro/darkmatter/utils.py
+++ b/gammapy/astro/darkmatter/utils.py
@@ -1,5 +1,6 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 """Utilities to compute J-factor maps."""
+import html
 import numpy as np
 import astropy.units as u
 
@@ -30,6 +31,12 @@ class JFactory:
         self.profile = profile
         self.distance = distance
         self.annihilation = annihilation
+
+    def _repr_html_(self):
+        try:
+            return self.to_html()
+        except AttributeError:
+            return f"<pre>{html.escape(str(self))}</pre>"
 
     def compute_differential_jfactor(self, ndecade=1e4):
         r"""Compute differential J-Factor.
--- a/gammapy/astro/source/pulsar.py
+++ b/gammapy/astro/source/pulsar.py
@@ -1,5 +1,6 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 """Pulsar source models."""
+import html
 import numpy as np
 from astropy.units import Quantity
 
@@ -35,6 +36,12 @@ class SimplePulsar:
         self.P_dot = P_dot
         self.I = I  # noqa: E741
         self.R = R
+
+    def _repr_html_(self):
+        try:
+            return self.to_html()
+        except AttributeError:
+            return f"<pre>{html.escape(str(self))}</pre>"
 
     @property
     def luminosity_spindown(self):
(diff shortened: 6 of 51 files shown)


registerrier (commented):
Thanks @evertrol . This looks nice.

Would it be possible to limit the amount of boilerplate code since most `_repr_html`methods are identical? maybe using decorators? This would also allow adding simple and limited test

evertrol:
I understand the concern, and I agree that ~60 changed files with 3/5(*) lines feels a bit much (there are a large amount of classes in Gammapy though!).
Then again, I'm more in favour of being explicit. It's similar to `__str__` and `__repr__` methods, just more of a (Jupyter? Pandas?) community invention. So I prefer the five lines instead of a (class) decorator.
(An alternative to a class decorator would be a class mixin, but I think that's only confusing for future developers: every base class needs to include this, for such a small bit of functionality.)

What I could do is remove the methods where there is no (not yet) specific `__str__` or `to_html()` method for that class (or a child class). That would reduce the amount of code. Everything would then simply default to `repr(instance)`.

I had done it the current way, because then it's already available (even if no `__str__` or `to_html()` would be implemented) for the future: if someone implements `__str__`, there will automatically be a nicer variant for `_repr_html_`; plus it doesn't get in the way of things.


(*) I see I haven't even been consistent with a try-except implementation everywhere.

QRemy:
maybe add a `repr_html` function  in gammapy.utils.script and use it in the classes to avoid duplication of the try/except bloc.

adonath:
I think we could even introduce a base class that is common to all objects in Gammapy. So not a mix-in solution but a a single common base class. I would imagine that such a base class could be useful for other use cases (especially for configurable objects later) as well.

@evertrol Do you know how other projects have solved this?

evertrol:
I am a bit wary of adding a utility function, because it takes the logic flow away from the class (and then back again to it): it would become `_repr_html_()` -> `utility_function()` -> `__str__()` (or similar. And while the function could be tested by itself, for correctness of the resulting HTML or text returned from `_repr_html_`, classes will need to be tested separately.

I don't know about a base class: I don't have enough insight into Gammapy for that, although it does feel as if the classes are so diverse that a generic base class might have very limited functionality. If it doesn't implement an `__init__()`, I guess it's effectively a mixin (over-generalizing perhaps). I also wonder how this works for classes that inherit from an abstract base class; I don't have experience mixing an ABC with another class as parents.

I had a quick look at Pandas and Astropy, to see what they do. From what I could see, there are far fewer implementations, probably because there are far fewer classes overall (I could imagine Astropy has skipped a few options to implement `_repr_html_`, but I expect for Pandas all implementations are there, given its usage).

It is simply possible to remove most of the current `_repr_html_` implementations, and just keep the ones for which a good `__str__()` implementation already exists. I did it the current way to preemptively handle the case where `__str__()` might be implemented for a class later on, so that there would automatically be a nicer representation in a notebook.

I guess another point of view is that it may look like a lot of (small) changes, but in the end, no-one will (hardly) notice. Users will perhaps find a nicer way of analysing data in notebooks, and developers will mostly ignore it, until someone implements `to_html()` method. In that sense, it's probably somewhat more like removing trailing whitespace everywhere, or applying a style guide and re-ordering imports in alphabetical order (from the looks of it, that's already the case; near).
It would then only be about the necessity, or not, of unit tests for this change.

These are just my thoughts; I don't have a strong opinion on this. Besides, I'm just dipping my toes into Gammapy, and came across this while working my way through the tutorials.

adonath:
Thanks @evertrol! 

Now that the work is done I would propose to just keep it. One of the most important features of Gammapy is interactive use of the objects in notebooks, so I think having a default implementation here is definitely an improvement to the API.

To avoid at least some code duplication we should make sure to just implement the `_repr_html_` for the base classes (such as `Map`, `Estimator`, `Maker`, but not `WcsNDMap`, `FluxPointsEstimator`, etc), which I think you already did. When trying out the functionality I noticed that the default formatting of the `__str__` representation does not work that well for html:

image

I guess adapting the line breaks and indentations would be needed to make it more readable.

evertrol:
Whoops, that was dumb: making a change, but not checking/using it myself. Thanks @adonath, for checking.

Unfortunately, the (most logical) fix requires surrounding the string version with `<pre>` and `</pre>`, and the actual contents of that needs to be HTML escaped (for example, when the default `__repr__` is used if `__str__` is not defined, the output will be like `<gammapy.module.submodule.SomeClass object at 0x7ff08045c>`, and the < and > characters need escaping). Which then adds another `import html` to the module.

All in all, things now look like
```
import html
...
    def _repr_html_(self):
        try:
            return self.to_html()
        except AttributeError:
            return f'<pre>{html.escape(str(self))}</pre>'
```
which of course is far from pretty looking.

(I could move the `import html` into the except clause, where it's only used. It's a fast import, and is really only needed in that part, while importing it for the whole module feels out of place. But that is less Pythonic than having all imports at the top.)

I've updated the PR to include the above solution.

This should provide proper formatted output in a notebook (when available), but I could imagine some points from the above discussion could be raised again, given e.g. the `return f'<pre>{html.escape(str(self))}</pre>'` line.

Screenshot_20220530_163039

QRemy:
@evertrol would you have time to continue the work on this PR ?

evertrol:
Yes, it's something I could pick again.

I let it rest, because it wasn't entirely clear to me what the preferred implementation would be:
- explicit for every relevant class (lots of code changes in a single PR, but hopefully not much maintenance)
- implementation in a base GammaPy object class
- implementation through a helper function (as noted, my feeling is that makes the flow a tad more annoying to follow)
- other?

If there is no clear preference, I'm happy to stick with the current idea: explicit, with a call to a `.to_html()` method if it exists, and otherwise a HTML-escaped version of `str(<instance>)`.

evertrol:
I've brought this PR up to the current gamma-py version.
Please check if this is still what

evertrol:
I've brought this PR up to the current gamma-py version.

Please check if this is still what people like (or at least are okay with ;-).

gammapy/catalog/hess.py
```diff
@@ -62,6 +63,12 @@ def __str__(self):
         )
         return ss
 
+    def _repr_html_(self):
```
bkhelifi, line 66: A naive question: why should we repeat it these lines when it is implemented in  the mother class `SourceCatalogObject`?

gammapy/estimators/map/ts.py
```diff
@@ -198,6 +199,12 @@ def __init__(
             ts_threshold=threshold,
         )
 
+    def _repr_html_(self):
```
bkhelifi, line 202: Should we repeat it here (mother class `Estimator' has it)

gammapy/estimators/profile.py
```diff
@@ -68,6 +69,12 @@ def __init__(self, x_edges=None, method="sum", axis="lon", center=None):
         self._x_edges = x_edges
         self.parameters = {"method": method, "axis": axis, "center": center}
 
+    def _repr_html_(self):
```
bkhelifi, line 72: The same comment

gammapy/utils/testing.py
```diff
@@ -243,6 +244,12 @@ def run(self, checks="all"):
             method = getattr(self, self.CHECKS[check])
             yield from method()
 
+    def _repr_html_(self):
```
bkhelifi, line 247: Is it really needed in this class?

gammapy/visualization/panel.py
```diff
@@ -36,6 +37,12 @@ def __init__(self, figure, xlim, ylim, npanels=4, **kwargs):
         self.parameters = {"xlim": xlim, "ylim": ylim, "npanels": npanels}
         self.grid_spec = GridSpec(nrows=npanels, ncols=1, **kwargs)
 
+    def _repr_html_(self):
```
bkhelifi, line 40: Really needed?

bkhelifi (commented):
Thank a lot.... I have few questions

evertrol:
@bkhelifi Thanks! Happy to see someone actually took the effort to step through all the edits.

Yes, I clearly went a bit too much on autopilot. I think all your comments are valid, in the sense that these occurrences can be removed (perhaps there is some argument to be made for the last one to keep it, but I'm not even sure what that would be).

I'll remove them. I need to update the PR anyway, as Black has been complaining (the double quotes Black enforces is one of the few things I dislike about it, even if it's consistent).

bkhelifi:
For the recording, maybe you can put in comments here screenshots of the outputs w/o and with this implementation ;)

evertrol:
I'm not sure if there is a use in screenshots: it's essentially just replacing every `print(something)` with `something` in the various example notebooks in the documentation (and make sure `something` is always on the last line, so some cells will have to be split). Which I actually want to do in a separate commit (bit of work to go through all the  examples).

One thing of note I saw is that it makes some results a tad nicer. For example, replacing the line 
```
print(irfs)
```
in https://docs.gammapy.org/dev/tutorials/data/cta.html#irfs with just
```
irfs
```
now produces a dict output spread across lines, instead of one long line. That is, the new output is
```
{'aeff': <gammapy.irf.effective_area.EffectiveAreaTable2D at 0x297e39a50>,
 'psf': <gammapy.irf.psf.parametric.EnergyDependentMultiGaussPSF at 0x2996000d0>,
 'edisp': <gammapy.irf.edisp.core.EnergyDispersion2D at 0x299689610>,
 'bkg': <gammapy.irf.background.Background3D at 0x177b6add0>}
 ```
 (In other words, `repr(something)` is handled nicer because it has some HTML thrown around it.)

**Comment by evertrol** [comment]
Actually, looking a bit further into this: it seems like that for each tutorial, there is a script and a notebook, both generated from the same source. Changing the examples to show the convenience of not having to `print()` various variables would not work in this case (in fact, `plt.show()` is also not necessary in the notebooks; I was a bit puzzled by that, but I see where it comes from): the script version would not show any output.

I think it goes to show that scripts/programs and notebooks are different things.

I would actually be so bold to suggest to drop the scripts, and turn all these tutorials into fully-fledged notebooks, at least as far usage concerns (the source can remain plain text). I don't really see too much of a point of having a fully annotated Python single-file program as a tutorial: perhaps useful for people that can't or won't use a web browser, or for some editors that can parse this differently; but that seem limited use cases.  
It makes sense to have it as a source, since notebooks themselves are annoying with version control systems; but perhaps then have the prints and shows dropped.

**Comment by registerrier** [comment]
> I would actually be so bold to suggest to drop the scripts, and turn all these tutorials into fully-fledged notebooks, at least as far usage concerns (the source can remain plain text). I don't really see too much of a point of having a fully annotated Python single-file program as a tutorial: perhaps useful for people that can't or won't use a web browser, or for some editors that can parse this differently; but that seem limited use cases. It makes sense to have it as a source, since notebooks themselves are annoying with version control systems; but perhaps then have the prints and shows dropped.

Thanks @evertrol.
On this specific point, we decided to move all the notebooks we had to python scripts because this works better with sphinx-gallery. The latter then produces the jupyter notebooks we ditribute to users via gammapy download

**Comment by evertrol** [comment]
> On this specific point, we decided to move all the notebooks we had to python scripts because this works better with sphinx-gallery. The latter then produces the jupyter notebooks we ditribute to users via gammapy download

I guess that makes the print functions mandatory for the Sphinx gallery to capture output and operate properly? (I don't really know, as I've never used that functionality.)

If not, and the scripts are effectively notebooks, I'd be willing to convert all print function calls to end-of-cell variable expressions in the various tutorial and example notebooks. It's not essential, but it feels more Jupyter-style.

**Review by bkhelifi — DISMISSED** [review]
Looks good to me, thanks @evertrol

**Comment by bkhelifi** [comment]
Thanks @evertrol. GitHub mentions many conflicts... can you have a look before making a review?
Thanks

**Comment by bkhelifi** [comment]
Thanks @evertrol. GitHub mentions many conflicts... can you have a look before making a review?
Thanks

**Comment by evertrol** [comment]
Whoops: I rebased onto the wrong branch when trying to bring the PR up to date. Hence all the conflicts and merged files.
Should be fixed now.

**Comment by evertrol** [comment]
Looks like tests such as 
```
assert repr(m) == repr(m_copy)
```
in `test_map_copy` were a bad idea, since the representations often contain the ID / address, which is obviously variable and not the same between copies.

I don't have the testbed set up yet locally, so I'll do that first, and then remove the offending tests where applicable (or fix them otherwise).

registerrier:
Thanks @evertrol . Merging now!

evertrol:
Thanks!

(I had wanted to ask whether I should actually squash the commits. Well, no need now 😉.)

PR #3948, opened 2022-05-12
Add test-cases for loading point-like irfs with data store

maxnoe:
**Description**

This PR adds test cases that should pass but are currently not passing for the loading of point-like IRFs.

In each of the test cases, the first asserts pass but the second do not 

**Dear reviewer**


adonath:
For some reason I think currently only the IRFs specified in `required_irfs` are loaded, however `required_irfs` only represents the sub-set of the required IRFs, optional ones should still be loaded if present.

maxnoe:
@adonath I updated the code to load all available IRFs and check if all the required once are there.

There was also a conceptual logic error in the code before: `REQUIRED_IRFS` contained a list of all irfs for `all-optional`, which doesn't make sense since "all optional" means "none required".

gammapy/data/data_store.py
```diff
@@ -284,15 +289,29 @@ def obs(self, obs_id, required_irf="full-enclosure"):
...
+        required_hdus = {"event", "gti"}.union(required_irf)
+
+        for hdu in ALL_HDUS:
+            hdu_location = self.hdu_table.hdu_location(
+                obs_id=obs_id, hdu_type=hdu, warn_missing=False,
+            )
+            if hdu_location is not None:
+                kwargs[hdu] = hdu_location
+            else:
+                if hdu in required_hdus:
```
QRemy, line 311:
```suggestion
            is_required = hdu in required_hdus
            hdu_location = self.hdu_table.hdu_location(
                obs_id=obs_id, hdu_type=hdu, warn_missing=is_required,
            )
            if hdu_location is not None:
                kwargs[hdu] = hdu_location
            else:
                if is_required:
```

gammapy/data/data_store.py
```diff
@@ -14,12 +14,17 @@
...
 
+ALL_IRFS = ["aeff", "edisp", "psf", "bkg", "rad_max"]
+ALL_HDUS = ["events", "gti"] + ALL_IRFS
 REQUIRED_IRFS = {
-    "full-enclosure": ["aeff", "edisp", "psf", "bkg"],
-    "point-like": ["aeff", "edisp"],
-    "all-optional": ["aeff", "edisp", "psf", "bkg", "rad_max"],
+    "full-enclosure": {"aeff", "edisp", "psf", "bkg"},
+    "point-like": {"aeff", "edisp"},
+    "all-optional": {},
```
QRemy, line 22: maybe we can just delete this option now

gammapy/data/data_store.py
QRemy: display missing warning for required_hdus

QRemy (commented):
Thanks @maxnoe, here are some suggestions.

gammapy/data/data_store.py
maxnoe: I don't understand this Suggestion. An error is raised here. A warning additionally to an exception does not make sense.

gammapy/data/data_store.py
```diff
@@ -284,15 +289,29 @@ def obs(self, obs_id, required_irf="full-enclosure"):
...
+            hdu_location = self.hdu_table.hdu_location(
+                obs_id=obs_id, hdu_type=hdu, warn_missing=False,
+            )
+            if hdu_location is not None:
+                kwargs[hdu] = hdu_location
+            else:
+                if hdu in required_hdus:
+                    raise MissingRequiredHDU(
+                        f"Required HDU {hdu} not found in observation {obs_id}"
+                    )
```
QRemy, line 314: Multiple hdus may be missing while the raise will catch only the first one that fails so I modified my previous suggestion

gammapy/data/data_store.py
QRemy, line 311:
```suggestion
        missing_hdu = []
        for hdu in ALL_HDUS:
            hdu_location = self.hdu_table.hdu_location(
                obs_id=obs_id, hdu_type=hdu, warn_missing= hdu in required_hdus,
            )
            if hdu_location is not None:
                kwargs[hdu] = hdu_location
            else:
               missing_hdu.append(hdu) 
        if missing_hdu :
            raise MissingRequiredHDU(
                f"Required HDUs {missing_hdu } not found in observation {obs_id}"
            )
```

gammapy/data/data_store.py
QRemy, line 314: `raise MissingRequiredHDU` will cath only the first fail but we want to display the warning for all missing hdus : 

```suggestion
        missing_hdu = []
        for hdu in ALL_HDUS:
            hdu_location = self.hdu_table.hdu_location(
                obs_id=obs_id, hdu_type=hdu, warn_missing= hdu in required_hdus,
            )
            if hdu_location is not None:
                kwargs[hdu] = hdu_location
            else:
               missing_hdu.append(hdu) 
        if missing_hdu :
            raise MissingRequiredHDU(
                f"Required HDUs {missing_hdu} not found in observation {obs_id}"
            )
```

maxnoe: Your suggestion would raise an error if any of the known hdu types is not there (including optional HDUs)

QRemy: modified

gammapy/data/data_store.py
maxnoe: That would be an API breaking change

gammapy/data/data_store.py
maxnoe: done

QRemy (approved):
Thanks @maxnoe, no further comments from my side.

adonath (approved):
Thanks @maxnoe, no further comments from my side.

PR #3950, opened 2022-05-13
PIG 23 - Gammapy Release Cycle and Version Numbering

registerrier:
**Description**

This pull request introduces a PIG on the release scheme and development workflow after v1.0

For now it is a first draft.

**Dear reviewer**

### DIFF
--- a/docs/development/pigs/pig-023.rst
+++ b/docs/development/pigs/pig-023.rst
@@ -6,7 +6,7 @@
 PIG 23 - Gammapy release cycle and version numbering
 ****************************************************
 
-* Author: Régis Terrier, Axel Donath and ?
+* Author: Régis Terrier, Axel Donath
 * Created: May 12th, 2022
 * Accepted:
 * Status:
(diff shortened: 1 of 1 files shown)


AtreyeeS (commented):
Thanks @registerrier !

- When and how do we decide to move from `1.x` to `2.0`? Probably that will not happen at least in the next 2 yrs, so we have some time to think?
- I suppose that we will have situations where the same bug fix implementation cannot be applied between `1.0.y` and `1.x.y` because there can be API changes. So the PRs have to be adjusted accordingly?
- I do not know if we are sufficiently well-planned yet to propose deprecation warnings. Maybe we can try putting it, but not make it necessary by adding in the PIG?

registerrier:
> * I suppose that we will have situations where the same bug fix implementation cannot be applied between `1.0.y` and `1.x.y` because there can be API changes. So the PRs have to be adjusted accordingly?

Good question. I have left a question in the text, but this might not need to be explained in the PIG. 
Any idea @adonath ?

> * I do not know if we are sufficiently well-planned yet to propose deprecation warnings. Maybe we can try putting it, but not make it necessary by adding in the PIG?

After the co-working week close out meeting we decided to include them.

registerrier:
This is now ready for review. Please comment! @adonath @AtreyeeS @QRemy @bkhelifi @fabiopintore @LauraOlivera @cosimoNigro and all.

docs/development/pigs/pig-023.rst
```diff
@@ -0,0 +1,128 @@
...
+If issues are found, they should be corrected before the actual release is made. This can lead
+to a delay in the release.
+
+Once the version is tagged, specific branches must be created for the feature and backport
+developments. This process will require a detailed description in the developer documentation.
+
+Deprecation
+===========
+
+API breaking changes can be introduced in feature and LTS releases. This changes must be clearly identified
```
AtreyeeS, line 109:
```suggestion
API breaking changes can be introduced in feature and LTS releases. These changes must be clearly identified
```

AtreyeeS (dismissed):
Thanks @registerrier ! No further comments from my side

contrera:
Hi
My 2 cents.
I have read the PIG and can´t see any obvious flaw or case when this should not work, but I have little experience in managing large packages on long time scales.
It looks more difficult to handle than before, for the developers, but better for users. 

José Luis

bkhelifi (approved):
Thanks @registerrier. This PIG can be merged.

adonath (approved):

Issue #3953, opened 2022-05-16
wrong release date in Zenodo / CITATION.cff


vuillaut:
**Bug description**
The Zenodo record shows Nov. 19, 2020 as a release date.
The error comes from the `CITATION.cff` which is used by Zenodo as a prior source of metadata.

**Expected behavior**
To have the correct release date.

**Other information**
You can modify the metadata directly on the Zenodo platform to fix the release date for v0.20.
To fix this on the long term, you may either:
- automatise the update of `CITATION.cff`
- remove the release date entirely from `CITATION.cff` and let Zenodo get this information from GitHub upon release

Bultako:
Thanks @vuillaut 
I have updated manually the date in the Zenodo record.

I would remove the date from `CITATION.cff` and I guess we should also remove the version number https://github.com/gammapy/gammapy/blob/master/CITATION.cff#L258

@adonath?

Note also that in order to publish the new version in Zenodo we have to draft a **release** in Github. Creating a tag as we usually do for every version of Gammapy is not enough. So, I drafted the release for v0.20 to let Zenodo fetch the info, and then removed it.

vuillaut:
> I would remove the date from `CITATION.cff` and I guess we should also remove the version number https://github.com/gammapy/gammapy/blob/master/CITATION.cff#L258

Ha yes, I am  surprised the version did not get mixed-up as well...

> Note also that in order to publish the new version in Zenodo we have to draft a **release** in Github. Creating a tag as we usually do for every version of Gammapy is not enough. So, I drafted the release for v0.20 to let Zenodo fetch the info, and then removed it.

Out of curiosity, why is that needed?

Bultako:
> Out of curiosity, why is that needed?

Automating publishing of a new release in Zenodo is triggered when **creating a release** using Github web interface, but this does not work if you only tag the release with `git tag`. I don't know why we have never created releases in Github. I just removed the release I created manually for Zenodo to keep consistency with all the previous ones, and I did it after automatic sync and update with Zenodo record was done.

vuillaut:
> > Out of curiosity, why is that needed?
> 
> Automating publishing of a new release in Zenodo is triggered when **creating a release** using Github web interface, but this does not work if you only tag the release with `git tag`. I don't know why we have never created releases in Github. I just removed the release I created manually for Zenodo to keep consistency with all the previous ones, and I did it after automatic sync and update with Zenodo record was done.

[...]

PR #3954, opened 2022-05-16
add action workflow to update citation.cff when doing a release

vuillaut:
**Description**
Introduces an action to update `CITATION.cff` release date when doing a new release in order to solve #3953 

**Dear reviewer**

I am not sure either this is the solution gammapy team would prefer to solve that issue, nor if it would actually work, so some feedback would be welcome.

Potential issue with that PR:
- if that action breaks for some reason (e.g. invalid `CITATION.cff`), it will break the tag release
- the codemeta conversion action should come after
- I am not actually sure if the action are run before Zenodo fetch the new tag and produces a new record

### DIFF
--- a/.github/workflows/updatecff.yml
+++ b/.github/workflows/updatecff.yml
@@ -1,0 +1,30 @@
+name: updatecff
+
+on:
+  tags:
+    - '*'
+
+jobs:
+  update:
+    name: "updatecff"
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out a copy of the repository
+        uses: actions/checkout@v2
+
+      - name: Update CITATION.cff release date
+        run: python dev/citationcff.py
+
+      - name: Check whether the citation metadata from CITATION.cff is valid
+        uses: citation-file-format/cffconvert-github-action@2.0.0
+        with:
+          args: "--validate"
+
+      - name: commit changes
+        uses: stefanzweifel/git-auto-commit-action@v4.1.6
+        if: success()
+        with:
+          commit_author: GitHub Actions <actions@github.com>
+          commit_message: commit metadata files
+
+        
--- a/dev/citationcff.py
+++ b/dev/citationcff.py
@@ -1,0 +1,8 @@
+import yaml
+from pathlib import Path
+from datetime import date
+
+citation_path = Path(__file__).parent.joinpath('CITATION.cff')
+cita = yaml.safe_load(open(citation_path).read())
+cita['date-released'] = date.today()
+yaml.safe_dump(cita, open(citation_path, 'w'))
--- a/dev/citationcff.py
+++ b/dev/citationcff.py
@@ -2,7 +2,7 @@ import yaml
 from pathlib import Path
 from datetime import date
 
-citation_path = Path(__file__).parent.joinpath('CITATION.cff')
+citation_path = Path(__file__).parents[1].joinpath('CITATION.cff')
 cita = yaml.safe_load(open(citation_path).read())
 cita['date-released'] = date.today()
 yaml.safe_dump(cita, open(citation_path, 'w'))


bkhelifi:
Dear all,
The citation scheme will be discussed during the next CC meeting. Maybe one can wait for the outcome before making a merge...

adonath:
I think some solution like this should be in place for v1.0, so I put the v1.0 milestone here...

adonath:
Thanks @vuillaut! What is not clear to me is, whether it make sense to have that as an action that runs on pushing a tag. This basically introduces the change after the tag, however shouldn't it be included with the tag? So the `citationcff.py` should be run manually before tagging?

bkhelifi:
Indeed, there are 3 cases (for the work after the V1.0):
- Just after a feature release, the CITATION.cff should be started from scratch, with global information coming from a TEMPLATE. The first merged PR could lead to the addition of the first author
- Just after a bug release, the CITATION.cff should not erased but appended. 
- In the other case, an append can be done.
Can a CI know that? A very good script should be set up for that...
Do you share this view?

adonath:
@bkhelifi No, I think independent of any version number the script should just always create a new `CITATION.cff` following the policy in PIG 24. Whether the author list is "appended" or not follows completely from the git history, there is no need to append an existing `CITATION.cff` ever. The info can always be created from the git history and the `AUTHORS_ALL.yml` file...

What is less clear to me is **when** the update to the `CITATION.cff` should happen. The options are:
- "Manually" before the tag using `dev/prepapre-release.py`. Then all the info is updated and included for the next version. However this requires "reserving" and DOI in Zenodo and then putting it in the `CITATION.cff`. In this case we do not need a CI as proposed in this PR. We just need the `cffconvert.py` script, which is run manually before defining the release tag.
- After the tag e.g. using a CI, but then the `CITATION.cff` is always lagging  and one version behind. 
- Remove all version specific info from the `CITATION.cff`, I guess mostly the `identifiers` section with the version specific DOIs.

vuillaut:
Hi @adonath 

I have thought about this, let me try to explain:

Here the current situation with the metadata and the release process:

gammapy_current

Metadata are partially handled by gammapy (`CITATION.cff`, `codemeta.json`).
Other metadata directly come from GitHub (repo name, URLs, release version...).

This is more flexible, lighter and less error prone for the Zenodo publication, because the GitHub-Zenodo integration is doing a lot for you in the background, and if the `CITATION.cff` is not readable, Zenodo will just use whatever comes from GitHub. 


If you want to do the publication to Zenodo yourself (required if you want to prereserve a DOI and update your metadata before upload), you then need to have full control over your metadata. If the metadata payload is not exactly what Zenodo expects, the publication will fail. However, this gives you more control - so I let you decide the benefit / risk balance.

The process might then look like this:

gammapy_prereserve

To handle the publication to Zenodo, you need:
- a Zenodo token (created by the Zenodo account owning gammapy)
- create the metadata payload for Zenodo

To handle the upload, you may use the ESCAPE `eossr` library and then code would look like this:

```
from eossr.api import zenodo
zen = zenodo.ZenodoAPI(access_token=token)
deposit = zen.create_new_entry()
doi = deposit.json()['metadata']['prereserve_doi']['doi']
record_id = deposit.json()['record_id']
gammapy_metadata['doi'] = doi
zen.upload_dir_content(path_to_gammapy_dir, record_id, gammapy_metadata)
```
The `gammapy_metadata` should be a dictionary following Zenodo expected metadata schema.
Taking advantage of the ESCAPE environment, you can read this metadata directly from your `codemeta.json` (given that it is correctly written):
```
import json
from eossr.metadata import codemeta2zenodo
gammapy_metadata = codemeta2zenodo.converter(json.load(open('codemeta.json')))
```

You can finish the publication in the script, or later manually in Zenodo API:
``` 
zen.publish_entry(record_id)
```

Note that in this workflow, `CITATION.cff` is not used anymore to handle the upload, the source of truth is the `codemeta.json` file.

vuillaut:
Interestingly, I opened an issue several months ago already to update the `codemeta.json` file with the prereserved DOI: https://gitlab.in2p3.fr/escape2020/wp3/eossr/-/issues/64

That would simplify your life, simply using eossr CLI in gammay CI when doing a release:
``` 
eossr-upload-repository --token $TOKEN --input-dir . 
```
If this is what you would like to have as a workflow, I could work on it. However, I think that would mean dropping `CITATION.cff` entirely, since `eossr` does  not handle that format.

vuillaut:
> Does this code be able to use a template to fill other fields? like this one: https://github.com/bkhelifi/gammapy/blob/Authors/dev/FOOTER_CODEMETA.json

Yes, you would just
1. load the base metadata from what you are calling here `FOOTER_CODEMETA.json`
2. update that metadata with whatever information (here probably release version, DOI, publication date...)
3. write `codemeta.json`
4. do the rest as proposed up there

registerrier:
What is the status here? Is it still applicable? @adonath @vuillaut ?

PR #3957, opened 2022-05-18
Fix DM spectral model

luca-giunti:
**Description**
Just added `is_norm=True` for the `scale` parameter in a dark matter mode, to allow fitting. Not sure about the default  `interp="log"`, does it make sense?


adonath (approved):
Thanks @luca-giunti!

PR #3961, opened 2022-05-22
Avoid evaluator update if model not contributes

QRemy:
Currently if a model contributes inside the dataset but outside the mask+psf_margin its evaluator will be updated at each optimizer call. This cause a serious performance drop if ones want to reduce mask_fit to fit only a subregion within the dataset. This PR simply adds a check on contributes in need_update to solve this.


registerrier (approved):
Thanks @QRemy 

I think we need to seriously work on the documentation and testing of the `MapEvaluator`.  It has become complex and it is easy to inadvertently remove an important feature when modifying it.

Issue #3963, opened 2022-05-23
Observation.peek returns empty subplots


AtreyeeS:
**Gammapy version**
0.20

`Observation.peek` is ugly as the first two subplots are empty (for the hdu corresponding to `events` and `gti`)

adonath:
Thanks @AtreyeeS! I guess the number of axes should be adapted to the actual number of available IRFs here...

registerrier:
Issue solved after #3966. Closing.

PR #3966, opened 2022-05-25
Adapt Observation.peek according to available hdus

AtreyeeS:
This addresses #3963 .

I have removed the warnings for the irrelevant hdus as they are unnecessarily confusing.

While implementing this PR, I realised that our `peek` methods arguments vary (eg `Observation.peek` takes `figsize` whereas `EventList.peek()` does not). It might be good to homogenise.

### DIFF
--- a/gammapy/data/observations.py
+++ b/gammapy/data/observations.py
@@ -19,6 +19,7 @@ from .event_list import EventList, EventListChecker
 from .filters import ObservationFilter
 from .gti import GTI
 from .pointing import FixedPointingInfo
+from itertools import zip_longest
 
 __all__ = ["Observation", "Observations"]
 
@@ -357,7 +358,8 @@ class Observation:
 
         plottable_hds = ["events", "aeff", "psf", "edisp", "bkg", "rad_max"]
 
-        plot_hdus = list(set(self.available_hdus) & set(plottable_hds))
+        plot_hdus = list(set(plottable_hds) & set(self.available_hdus))
+        plot_hdus.sort()
 
         n_irfs = len(plot_hdus)
         nrows = n_irfs // 2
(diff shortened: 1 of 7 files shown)


gammapy/data/observations.py
```diff
@@ -346,51 +346,56 @@ def check(self, checks="all"):
...
 
-        if "edisp" in self.available_hdus:
+        if "edisp" in plot_hdus:
             self.edisp.plot_bias(ax=axes_dict["edisp"], add_cbar=True)
             axes_dict["edisp"].set_title("Energy dispersion")
-        else:
-            logging.warning(f"No energy dispersion found for obs {self.obs_id}.")
+
+        if "rad_max" in plot_hdus:
+            self.rad_max.plot_rad_max_vs_energy(ax=axes_dict["rad_max"])
```
registerrier, line 394: Maybe this would deserve a test

gammapy/data/observations.py
```diff
@@ -346,51 +346,56 @@ def check(self, checks="all"):
...
+    def peek(self, figsize=(15, 10)):
         """Quick-look plots in a few panels.
 
         Parameters
         ----------
         figsize : tuple
             Figure size
         """
-        n_irfs = len(self.available_hdus)
+        plot_hdus = self.available_hdus
```
registerrier, line 357: It might be useful to take the intersection of `available_hdus` with a set of plottable hdus. This avoids removing explicitly `gti` and would ensure that no additional sub_plot is created when another hdu is present.

gammapy/data/observations.py
```diff
@@ -346,51 +346,57 @@ def check(self, checks="all"):
...
         """Quick-look plots in a few panels.
 
         Parameters
         ----------
         figsize : tuple
             Figure size
         """
-        n_irfs = len(self.available_hdus)
+
+        plottable_hds = ["events", "aeff", "psf", "edisp", "bkg", "rad_max"]
```
adonath, line 358: Instead of repeating the IRF names here it would be better to just use `self.available_hdus` and remove "gti" from it, which is the only one which is not plottable.

adonath (changes requested):
Thanks @AtreyeeS, I have left one minor comment for cleanup of the code. Would you mind adding a before / after screenshot to the PR? This is will be useful for future reference...

gammapy/data/observations.py
AtreyeeS: That was my initial implementation, but as @registerrier pointed out, that leaves the risk of adding additional sub plots if hdus are added in future

AtreyeeS:
Hi @adonath  ! Here is a before:
Screenshot 2022-05-25 at 10 39 29

and after
- for full enclosure 
Unknown-1
- varying rad_max
Unknown

registerrier:
There is also an issue with the X-Y axes for the events plot. The wcs axes are not visible.

The last subplot in the first case has empty axes drawn.

gammapy/data/observations.py
adonath: It's close to 50:50 I guess, because one could also argue that if additional HDUs are supported they should be plotted as well, if available. But OK to keep from my side for now...

adonath:
Thanks @AtreyeeS for the preview, it looks much better now! Once comment for the event plot: this currently misses the wcs projection, which is present when using `EventList.plot_image()` directly. This is because in `.peek()` the subplots are re-initialised without wcs by default. As we rely on `plt.subplots()` the fix is maybe not simple and one needs to change to `fig.add_subplot()` instead...maybe it's possible to replace and axes?

adonath:
@AtreyeeS Yes, the PSF plotting issue should be fixed independently.

adonath:
@AtreyeeS Axes in the grid not used for plotting can be hidden with: `ax.set_visible(False)`

AtreyeeS:
This is what the plots look like now

Unknown-3

and

Unknown-2

 Any further comments or should we merge?

gammapy/data/observations.py
```diff
@@ -346,51 +346,65 @@ def check(self, checks="all"):
...
+        )
 
-        if "edisp" in self.available_hdus:
-            self.edisp.plot_bias(ax=axes_dict["edisp"], add_cbar=True)
-            axes_dict["edisp"].set_title("Energy dispersion")
-        else:
-            logging.warning(f"No energy dispersion found for obs {self.obs_id}.")
+        for idx in range(nrows * ncols):
+            ax = axes.flat[idx]
+            try:
```
adonath, line 375: Please write this `try/except` more compact, because the `IndexError` could happen anywhere in the code block and it would go silent. Ideally `try/except` statements only cover a single line, or maybe just use:
```
for ax, name in zip_longest(axes.flat, plot_hdus):
    if name is None:
        ax.set_visible(False)
        continue
```

adonath (approved):
Thanks @AtreyeeS, I have no further comments!

adonath:
CI fail is unrelated.

PR #3968, opened 2022-05-25
Set reasonable error message case of no free parameters for optimisation

AtreyeeS:
This PR sets `FluxEstimator.reoptimize = False` in case there are no free parameters to reoptimize

### DIFF
--- a/gammapy/estimators/parameter.py
+++ b/gammapy/estimators/parameter.py
@@ -126,7 +126,7 @@ class ParameterEstimator(Estimator):
             # compute ts value
             parameter.value = self.null_value
 
-            if self.reoptimize:
+            if self._reoptimize:
                 parameter.frozen = True
                 _ = self.fit.optimize(datasets=datasets)
 
@@ -167,7 +167,7 @@ class ParameterEstimator(Estimator):
             datasets=datasets,
             parameter=parameter,
             sigma=self.n_sigma,
-            reoptimize=self.reoptimize,
+            reoptimize=self._reoptimize,
         )
 
         return {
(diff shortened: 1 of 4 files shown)


gammapy/estimators/parameter.py
```diff
@@ -308,6 +308,13 @@ def run(self, datasets, parameter):
 
         with datasets.parameters.restore_status():
 
+            if len(datasets.parameters.free_parameters.names) == 1:
+                if self.reoptimize:
+                    log.warning(
+                        f"No free parameters to reoptimize. Setting reoptimize to False"
+                    )
+                    self.reoptimize = False
```
adonath, line 316: I think modifying the state of the `Estimator` and overwriting the use input is not a good idea.  When the `Estimator` is re-used with different datasets, the state might change arbitrarily...

I think just emitting the warning that `reoptimize` is ignored in this case should be fine.

adonath: And the fix the actual error as low as possible in the hierarchy, probably in `gammapy.modeling.fit`...

adonath: > I am still unsure how to fix it in modeling.fit, ie, what should a fit do in the absence of any free parameters? 

I think just emitting the warning here is fine. E.g. [this method]( https://github.com/gammapy/gammapy/blob/master/gammapy/modeling/fit.py#L139) is called at the beginning of each method call of `Fit.` Here we could check:
- If there are free parameters at all, if not emit a warning
- Pass the `reoptimize` option to the method and check inside, whether there are other parameters to re-optimize, if not emit a warning and change the value of reoptimize, like:

```python
datasets, parameters, reoptimize = self._parse_datasets(datasets, reoptimize)
```
Or similar...alternatively this could happen on the `Parameters` level (we already have `Parameters.check_limits()`). E.g. like:

```python
parameters.check_reoptimize()
```

adonath:
Requires further review...

adonath:
Requires further review...

AtreyeeS:
Updated PR to fail with `ValueError` in case of no free parameters for fitting. No special handling in case of estimators

adonath (approved):
Thanks @AtreyeeS!

PR #3977, opened 2022-06-07
Fix parameters min/max display in serialization 

QRemy:
This PR fix a bug in model serialization where the min/max different from default were never shown for nan default.
(Also modified model.to_dict such `is_norm` is hidden if `full_output=False`).


adonath (approved):

PR #3979, opened 2022-06-08
Fix inconsistency between NMCID and number of MID* keywords in `MapDatasetEventSampler`

fabiopintore:
I've found the possibility of an inconsistency between the value given in the NMCID keyword and the number of MID* keywords in the header of an event list obtained from `MapDatasetEventSampler`. 
NMCID indicates the total number of MID* keywords. However, currently if `MapDatasetEventSampler` does not simulate any event for a given source (i.e. a very faint source), the corresponding MID keywords is not stored but but the NMCID value counts it anyway.
This PR tries to fix this issue.

### DIFF
--- a/gammapy/datasets/simulate.py
+++ b/gammapy/datasets/simulate.py
@@ -88,11 +88,11 @@ class MapDatasetEventSampler:
             table = self._sample_coord_time(npred, temporal_model, dataset.gti)
             if len(table) > 0:
                 table["MC_ID"] = idx + 1
-                table.meta["MID{:05d}".format(idx + 1)] = idx + 1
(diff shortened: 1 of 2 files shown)


gammapy/datasets/simulate.py
```diff
@@ -88,11 +88,11 @@ def sample_sources(self, dataset):
             table = self._sample_coord_time(npred, temporal_model, dataset.gti)
             if len(table) > 0:
```
adonath, line 89: Is this case still needed?

fabiopintore: No, it's not. 
We would need a condition only to build the astropy table in case of zero sampled events, I think.

adonath (approved):
Thanks @fabiopintore!

PR #3980, opened 2022-06-08
Fix bug in `TemplateSpatialModel.evaluate`

fabiopintore:
In very rare case, it may happen that the evaluation of a `TemplateSpatialModel` model provide negative values. This PR aims at fixing this issue.


adonath:
Does it actually resolve the issue you observed?

fabiopintore:
Yes, the issue does not appear anymore now.

adonath:
Thanks @fabiopintore, can you please add a minimal test case? Create a `TemplateSpatialModel` with negative values in the map and make sure those are not present any more after evaluation. Please also add a comment in the docstring, that negative values are clipped.

adonath (approved):
Thanks @fabiopintore, no further comment from my side.

Issue #3982, opened 2022-06-10
RegionNDMap.fill_events(events) fails for a PointSkyRegion if events is empty 


registerrier:
**Gammapy version**
v0.20 and dev

**Bug description**
The data reduction for radmax spectral analysis can fail in case no off counts are selected because the command:
`counts.fill_events(events)` fails when no events are selected.

**Expected behavior**
The command should work even if no events are found as is the case for other region types.

**To Reproduce**
```
from astropy.coordinates import SkyCoord
import astropy.units as u
from gammapy.data import DataStore
from gammapy.maps import RegionNDMap

ds=DataStore.from_dir("$GAMMAPY_DATA/hess-dl3-dr1/")
obs=ds.obs(23523)

events=obs.events.select_offset([70.,71]*u.deg)

counts=RegionNDMap.create("icrs;point(0,0)")

counts.fill_events(events)
```

**Other information**
The error is cause by the current implementation of `PointSkyRegion.contains(skycoord, wcs)` which always returns a `bool` independently of the `skycoord.shape`.

See:
https://astropy-regions.readthedocs.io/en/stable/_modules/regions/shapes/point.html#PointSkyRegion.contains

registerrier:
See associated issue: https://github.com/astropy/regions/issues/465

adonath:
Thanks @registerrier, I guess we don't wait for the fix in regions but implement a workaround in Gammapy instead...

PR #3986, opened 2022-06-10
Fix bug in RegionNDMap.fill_events with PointSkyRegion 

registerrier:
**Description**

This pull request solves #3982 . The fix introduced should be removed once corrected in regions.

**Dear reviewer**


adonath (approved):
Thanks @registerrier, no comments from my side!

PR #4002, opened 2022-06-20
Modify conf.py to support bugfix versions

registerrier:
**Description**

This pull request corrects the switcher in the docs to support bug fix releases. Currently every version with more than 5 characters is renamed to dev in the switcher. This is not enough for e.g. 0.20.1 . 
The change here relies on the presence of the dev string in the version name.

**Dear reviewer**


adonath (approved):
Thanks @registerrier! I have no comments..

PR #4003, opened 2022-06-20
Add template generation of quickstart in documentation

registerrier:
**Description**

This pull request is a proposal to help simplify the release process w.r.t. doc generation.

- First the Makefile is modified to avoid the necessity of performing part of step 3 described in dev doc:
```
In the 0.14/docs/_downloads folder, rename notebooks-dev.tar file as notebooks-0.14.tar.
```
- Second a python script is added to generate the quick start and gammapy download sections of the getting started
  - The section is included for a release but not for a dev version. This avoids the inconsistency with the versioning in the dev doc.
- Third the Makefile is adapted to automatically create the files before compiling the doc.

**Dear reviewer**

This is a proposal two remove two manual steps in the release process. It is a first step towards a possible automation of the doc deployment.

Some of the solutions proposed are not ideal e.g. having the text of the quick start in a python script is not very intuitive.

Note that it is not possible to use the sphinx substitution mechanism in the `code-block` hence the proposal to programmatically generate the code block. 

Note that this requires testing. But I am not sure where to put the tests.

PR #4004, opened 2022-06-21
Add code block substitution and conditional quick start in doc getting-started

registerrier:
**Description**

This pull request is an alternative approach to PR #4003. As suggested by @adonath it uses a new sphinx directive for substitution in code block and is applied for the quick start section. A conditional statement is added not to include the quick start for dev versions.

Note that here the substitution is added in the code base. Conversely we can rely on the external package: 
https://github.com/adamtheturtle/sphinx-substitution-extensions

**Dear reviewer**


docs/sphinx_utils.py
```diff
@@ -0,0 +1,39 @@
+"""
```
adonath, line 1: I think it's fine to keep this on our side, however the other sphinx extensions we have are in `gammapy/utils/docs.py`, maybe move to there?

docs/getting-started/index.rst
```diff
@@ -55,66 +55,9 @@ Installation
         :text: Learn more
         :classes: btn-secondary stretched-link
 
-.. _quickstart-setup:
+.. ifconfig:: not 'dev' in version
```
adonath, line 58: Does this mean the the Quickstart is not included at all for dev versions? I guess this mostly makes sense, alternatively we could just work with a place-holder like:

```
https://gammapy.org/download/install/gammapy-x.x.x-environment.yml
```

adonath (dismissed):
Thanks @registerrier, this looks like an elegant solution to me!

docs/sphinx_utils.py
registerrier: Done. 
I did not manage to use exactly the same logic as was previously used. So I added the `setup` in the `conf.py`.

Should we test this code? I suppose we should, but it is not clear how. and what to test.

docs/getting-started/index.rst
adonath: I guess the downside of hiding it for dev version is that we don't see the rendered output even when changes are being made? But I have no strong opinion there...

registerrier: Yes Indeed. With the `ifconfig` scheme you can only test the result on a tagged version, so hiding the section completely is a bit annoying.

adonath: +1 for the `X.Y.Z` I think this makes it obvious, that there is no env file for dev versions...

adonath:
Two more things:
- Update dev docs?
- Still include the argument forwarding for the `Makefile` from #4003? It is used for the notebooks...

registerrier:
I can open a new PR for the Makefile as it is a slightly different matter

adonath:
Idea from dev meeting: add a warning box for dev builds saying there is no quickstart setup

adonath (approved):
Thanks @registerrier! I think we should go ahead and merge this. The warning box could optionally be added in a follow-up PR.

PR #4014, opened 2022-06-30
Avoid unnecessary copy in Map.stack

QRemy:
Remove unnecessary copy in Map.stack before replacing nan values


QRemy:
> Is this approach more efficient regarding CPU usage? Is the additional call to `np.isfinite` equivalent?

My thought  was that `stack` is used mostly in `npred_signal` where there is usually no NaN so avoiding the copy is worth in that case even if for the other cases this add one more `np.isfinite` call.

registerrier (dismissed):
Thanks @QRemy 
This looks good to me.

registerrier (approved):
Thanks @QRemy !

PR #4015, opened 2022-06-30
Allow to stack mask_fit in Dataset.stack

QRemy:
Allow to stack mask_fit in Dataset.stack in the same way than mask_safe

### DIFF
--- a/gammapy/datasets/map.py
+++ b/gammapy/datasets/map.py
@@ -703,7 +703,8 @@ class MapDataset(Dataset):
             Masked dataset
         """
         dataset = self.__class__.from_geoms(**self.geoms, name=name)
-        dataset.mask_fit.data = False
+        if self.mask_fit:
+            dataset.mask_fit.data = False
         dataset.stack(self, nan_to_num=nan_to_num)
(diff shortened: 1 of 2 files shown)


gammapy/datasets/map.py
```diff
@@ -786,6 +786,11 @@ def stack(self, other, nan_to_num=True):
         if self.mask_safe and other.mask_safe:
             self.mask_safe.stack(other.mask_safe)
 
+        if self.mask_fit and other.mask_fit:
+            self.mask_fit.stack(other.mask_fit)
+        elif other.mask_fit:
+            self.mask_fit = other.mask_fit
```
adonath, line 792:
```suggestion
            self.mask_fit = other.mask_fit.copy()
```

adonath (changes requested):
Thanks @QRemy, I have left one inline suggestion. I think we should copy the other mask...

registerrier (approved):
Thanks @QRemy

registerrier:
@adonath comment has been taken into account. Merging now.

PR #4017, opened 2022-06-30
Introduce a phase curve model

registerrier:
**Description**

This pull request introduces a possible PhaseCurve template model. Contrary to other template models, it also contains `Parameters` that describe a simple timing solution.

This raises question on the bookkeeping since those parameters would be described elsewhere. I assume the parameters  could be defined as keywords in the `Table` header. But then we have an issue of consistency with the yaml `Models` file.

We could decide that the `models.yaml` parameters take precedence but that can be tricky. 

**Dear reviewer**

does it fit the required use case @fabiopintore ? Are there public files we could use to test?

### DIFF
--- a/gammapy/modeling/models/temporal.py
+++ b/gammapy/modeling/models/temporal.py
@@ -6,7 +6,7 @@ from astropy import units as u
 from astropy.table import Table
 from astropy.time import Time
 from astropy.utils import lazyproperty
-from gammapy.maps import RegionNDMap, TimeMapAxis
+from gammapy.maps import RegionNDMap, TimeMapAxis, MapAxis
 from gammapy.modeling import Parameter, Parameters
 from gammapy.utils.random import InverseCDFSampler, get_random_state
 from gammapy.utils.scripts import make_path
@@ -826,12 +826,25 @@ class TemplatePhaseCurveTemporalModel(TemporalModel):
         ph_min, n_min = self._time_to_phase(t_min.mjd*u.d, **kwargs)
         ph_max, n_max = self._time_to_phase(t_max.mjd*u.d, **kwargs)
 
-        val1 = self._interpolator.antiderivative()(ph_max)-self._interpolator.antiderivative()(0)
-        val2 = self._interpolator.antiderivative()(1)-self._interpolator.antiderivative()(ph_min)
-        total = self._interpolator.antiderivative()(1)-self._interpolator.antiderivative()(0)
+        # here we assume that the frequency does not change during the integration boundaries
+        delta_t = (t_min.mjd - self.t_ref.value)*u.d
+        frequency = self.f0.quantity + delta_t * (self.f1.quantity + delta_t  * self.f2.quantity/2)
 
-        total *= (n_max-n_min)
-        return (total + val1 + val2)/self.time_sum(t_min, t_max)
+        # Compute integral of one phase
+        phase_integral = self._interpolator.antiderivative()(1)-self._interpolator.antiderivative()(0)
+        # Multiply by the total number of phases
+        phase_integral *= (n_max-n_min-1)
+
+        # Compute integrals before first full phase and after the last full phase
+        end_integral = self._interpolator.antiderivative()(ph_max)-self._interpolator.antiderivative()(0)
+        start_integral = self._interpolator.antiderivative()(1)-self._interpolator.antiderivative()(ph_min)
+
+        # Divide by Jacobian (here we neglect variations of frequency during the integration period)
+        total = (phase_integral + start_integral + end_integral)/frequency
+        # Normalize by total integration time
+        integral_norm =  total / self.time_sum(t_min, t_max)
+
+        return integral_norm.to("")
 
     @classmethod
     def from_dict(cls, data):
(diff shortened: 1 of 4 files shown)


gammapy/modeling/models/temporal.py
```diff
@@ -678,3 +679,103 @@ def integral(self, t_min, t_max):
...
+        if filename is not None:
+            filename = str(make_path(filename))
+        self.filename = filename
+        super().__init__(**kwargs)
+
+    @classmethod
+    def read(cls, path):
+        """Read phasecurve model table from FITS file.
+
+        Beware : this does **not** read parameters.
```
bkhelifi, line 738: Do you mean here that there are not parameters but a table?

bkhelifi:
Hi @registerrier,
I do not understand the association of the parameters and the table. The parameters are used only in `evaluate` and the building of the model never uses it.

gammapy/modeling/models/temporal.py
```diff
@@ -678,3 +679,103 @@ def integral(self, t_min, t_max):
...
+        delta_t = time - t_ref
+        phase = (phi_ref + delta_t * (f0 + delta_t / 2. * (f1 + delta_t / 3 * f2))).to_value('')
+
+        phase -= np.floor(phase)
+        return self._interpolator(phase)
+
+    @classmethod
+    def from_dict(cls, data):
+        params = data["temporal"]["parameters"]
+        return cls.read(data["temporal"]["filename"])
```
QRemy, line 774: This should read also the parameters.
Could be similar than the EBl model : 
https://docs.gammapy.org/dev/_modules/gammapy/modeling/models/spectral.html#EBLAbsorptionNormSpectralModel.from_dict

gammapy/modeling/models/temporal.py
```diff
@@ -678,3 +679,103 @@ def integral(self, t_min, t_max):
...
+
+    def to_dict(self, full_output=False):
+        """Create dict for YAML serialisation"""
+        model_dict = super().to_dict()
+        model_dict["filename"] = self.filename
+        return model_dict
```
QRemy, line 780: As for the EBL model the data could save directly in the yaml file if a filename is not provided :
https://docs.gammapy.org/dev/_modules/gammapy/modeling/models/spectral.html#EBLAbsorptionNormSpectralModel.to_dict

gammapy/modeling/models/temporal.py
registerrier: Yes, the `read` method only reads the `Table`. The `Parameters` should be read from the models.yaml file.

gammapy/modeling/models/temporal.py
```diff
@@ -678,3 +679,103 @@ def integral(self, t_min, t_max):
...
+    f2 : `~astropy.units.Quantity`
+        The frequency second derivative at t_ref in s-3
+    """
+    tag = ["TemplatePhaseCurveTemporalModel", "template-phase"]
+    _t_ref_default = Time(48442.5, format="mjd")
+    t_ref = Parameter("t_ref", _t_ref_default.mjd, unit="day", frozen=True)
+    phi_ref = Parameter("phi_ref", 0, unit="", frozen=True)
+    f0 = Parameter("f0", "29.946923 s-1", frozen=True)
+    f1 = Parameter("f1", "-3.77535E-10 s-2", frozen=True)
+    f2 = Parameter("f2", "1.1147E-20 s-3", frozen=True)
```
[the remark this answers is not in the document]
registerrier: Those are parameters representing the `Crab` pulsar in ATNF. That's a possible default.

fabiopintore:
Thanks @registerrier . Unfortunately there are no public models yet. I'm waiting for them. I can pass either the phasecurve model for the binaries or a pulsar phasecurve from the previous CTA DC...

gammapy/modeling/models/temporal.py
registerrier: OK will adapt accordingly.

gammapy/modeling/models/temporal.py
```diff
@@ -662,3 +663,131 @@ def integral(self, t_min, t_max):
...
+            Table.read(filename),
+            filename=filename,
+            t_ref=t_ref,
+            phi_ref=phi_ref,
+            f0=f0,
+            f1=f1,
+            f2=f2,
+        )
+
+    def write(self, path=None, overwrite=False):
```
bkhelifi, line 755: The ``write`` method stores only the table, not the other parameters. What is the logic behind? 
In parallel, the ``read`` method ingests the table stored in the filename and the other parameters have to be put in argument. So the whole is coherent. 
But, why can not the full model be stored? When manipulating the dictionary it seems that everything is handled...

gammapy/modeling/models/tests/test_temporal.py
```diff
@@ -313,3 +316,52 @@ def test_plot_constant_model():
     constant_model = ConstantTemporalModel(const=1)
     with mpl_plot_check():
         constant_model.plot(time_range)
+
+
+def test_phase_curve_model(tmp_path):
```
bkhelifi, line 321: Maybe you can add a test on the filename raise (L759)?

adonath:
This PR does not have a milestone, @registerrier will this be ready for v1.0?

gammapy/modeling/models/temporal.py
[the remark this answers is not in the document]
registerrier: I changed the first and second derivatives default to 0.

gammapy/modeling/models/temporal.py
```diff
@@ -662,3 +663,131 @@ def integral(self, t_min, t_max):
...
+    f1 : `~astropy.units.Quantity`
+        The frequency derivative at t_ref in s-2
+    f2 : `~astropy.units.Quantity`
+        The frequency second derivative at t_ref in s-3
+    """
+
+    tag = ["TemplatePhaseCurveTemporalModel", "template-phase"]
+    _t_ref_default = Time(48442.5, format="mjd")
+    _phi_ref_default = 0
+    _f0_default = 29.946923 * u.s**-1
```
bkhelifi, line 707:
```suggestion
    _f0_default = 29.946923 * u.s**-1 # Crab pulsar value, as default
```

bkhelifi (approved):
Thanks @registerrier . OK for me

gammapy/modeling/models/temporal.py
```diff
@@ -79,6 +80,8 @@ def plot(self, time_range, ax=None, **kwargs):
             times to plot the model
         ax : `~matplotlib.axes.Axes`, optional
             Axis to plot on
+        nbin : int
```
adonath, line 83: In `SpectralModel.plot` we called this `n_points`, so we are inconsistent here. I think the best solution would be to rename to `nbin` consistently, but that's backwards incompatible. The 2nd best solution would probably to at least go for consistency in the `.plot` methods an call this `n_points` here as well.

adonath (approved):
Thanks @registerrier, I only have one minor inline comment...

gammapy/modeling/models/temporal.py
registerrier: modified to `n_points`

adonath (approved):

bkhelifi (approved):
Thanks a lot @registerrier. The merge can be made.

Issue #3707, opened 2022-01-19
TSMapEstimator fails when exposure is zero at the map center


adonath:
**Gammapy version**
v0.19

**Bug description**
To compute the kernel for the TS map computation the exposure at the center of the map is used. However this can fail, when the exposure at the map center is zero, but contains e.g. valid data in the upper part of the map. One could work e.g. with the mean exposure taking the `mask` into account instead.

PR #4018, opened 2022-07-01
Computation of the WcsMap kernel at the nearest valid exposure

bkhelifi:
**Description**

This pull request fixes the bug mentioned in the Issue #3707. It changes the computation of the kernel, previously done with the exposure at the centre of the map and now with the mean non-null exposure over the whole map.

**Dear reviewer**
In addition to the change of the computation logics, I have added a test on the kernel values and also a test with a map with bins with null exposure.

### DIFF
--- a/gammapy/datasets/utils.py
+++ b/gammapy/datasets/utils.py
@@ -23,3 +23,12 @@ def get_axes(ax1, ax2, width, height, args1, args2, kwargs1=None, kwargs2=None):
         raise ValueError("Either both or no Axes must be provided")
 
     return ax1, ax2
+
+
+def get_nearest_valid_exposure_position(exposure, position=None):
+    import numpy as np
+    mask_exposure = exposure > 0.0*exposure.unit
+    mask_exposure = mask_exposure.reduce_over_axes(func=np.logical_or)
+    if not position:
+        position = mask_exposure.geom.center_skydir
+    return mask_exposure.mask_nearest_position(position)
--- a/gammapy/datasets/utils.py
+++ b/gammapy/datasets/utils.py
@@ -1,3 +1,6 @@
+import numpy as np
+
+
 def get_figure(fig, width, height):
     import matplotlib.pyplot as plt
 
@@ -26,7 +29,6 @@ def get_axes(ax1, ax2, width, height, args1, args2, kwargs1=None, kwargs2=None):
 
 
 def get_nearest_valid_exposure_position(exposure, position=None):
-    import numpy as np
     mask_exposure = exposure > 0.0*exposure.unit
     mask_exposure = mask_exposure.reduce_over_axes(func=np.logical_or)
     if not position:
--- a/gammapy/estimators/map/ts.py
+++ b/gammapy/estimators/map/ts.py
@@ -215,7 +215,7 @@ class TSMapEstimator(Estimator):
     def estimate_kernel(self, dataset):
         """Get the convolution kernel for the input dataset.
 
-        Convolves the model with the PSFKernel at the center of the dataset.
+        Convolves the model with the PSFKernel using the mean non-null exposure.
 
         Parameters
         ----------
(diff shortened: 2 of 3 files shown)


gammapy/estimators/map/ts.py
```diff
@@ -236,10 +236,9 @@ def estimate_kernel(self, dataset):
         model = self.model.copy()
         model.spatial_model.position = geom.center_skydir
 
-        # Creating exposure map with exposure at map center
+        # Creating exposure map with the mean non-null exposure
         exposure = Map.from_geom(geom, unit="cm2 s1")
-        exposure_center = dataset.exposure.to_region_nd_map(geom.center_skydir)
-        exposure.data[...] = exposure_center.data
+        exposure.data[...] = np.mean(dataset.exposure.data[dataset.exposure.data > 0.])
```
QRemy, line 241: you could also consider the dataset.mask and only finite values to compute the mean exposure :
```python
exposure_mask = (dataset.mask.data) & (dataset.exposure.data > 0.) & np.isfinite(dataset.exposure.data)
exposure.data[...] = np.mean(dataset.exposure.data[exposure_mask])
```

QRemy: this also address my comment about unit

QRemy: I think we can use  dataset.mask.interp_to_geom( dataset.exposure.geom)  as a the mask in reco energy is dilated compared to the mask in true energy we don't loose information .

bkhelifi: This can not work as the energy axis names are different between the exposure (in true energy) and the mask (in reco energy)... No turn-around found yet.

registerrier (commented):
Thanks @bkhelifi .

I am not sure you need to use a full mask on the exposure, I suppose that the image_mask should be sufficient. Note that the current implementation does not return an array for the exposure.
Is it possible to quantify the result when applying the mean compared to the center value? 

Is it possible that no PSF is available as well?

adonath:
Idea from dev meeting today: make behavior consistent with IRF kernel extraction. So find the nearest valid position to extract the exposure from. I think for a TS map we always need a PSF, so we could use:

```python3
datasets.psf._get_nearest_valid_position(position=, )
```
It's not ideal to rely on a hidden method, but as we do it internally in Gammapy it's acceptable.

gammapy/estimators/map/ts.py
```diff
@@ -236,10 +236,9 @@ def estimate_kernel(self, dataset):
         model = self.model.copy()
         model.spatial_model.position = geom.center_skydir
 
-        # Creating exposure map with exposure at map center
+        # Creating exposure map with the mean non-null exposure
         exposure = Map.from_geom(geom, unit="cm2 s1")
```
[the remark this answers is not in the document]
adonath: This is still unaddressed?

gammapy/datasets/utils.py
```diff
@@ -23,3 +23,12 @@ def get_axes(ax1, ax2, width, height, args1, args2, kwargs1=None, kwargs2=None):
         raise ValueError("Either both or no Axes must be provided")
 
     return ax1, ax2
+
+
+def get_nearest_valid_exposure_position(exposure, position=None):
+    import numpy as np
```
registerrier, line 29: I would import globally here.

gammapy/estimators/map/ts.py
```diff
@@ -236,10 +237,13 @@ def estimate_kernel(self, dataset):
...
-        exposure_center = dataset.exposure.to_region_nd_map(geom.center_skydir)
-        exposure.data[...] = exposure_center.data
+        position = get_nearest_valid_exposure_position(dataset.exposure, geom.center_skydir)
+        exposure_position = dataset.exposure.to_region_nd_map(position)
+        if not np.any(exposure_position.data):
+            raise ValueError("The exposure is null. Impossible to convolve the model with the PSFKernel.")
```
registerrier, line 245: I would use a simpler error message e.g.:
`No valid exposure. Impossible to compute kernel for TS Map.`

gammapy/estimators/map/ts.py
```diff
@@ -215,7 +216,7 @@ def selection_all(self):
     def estimate_kernel(self, dataset):
         """Get the convolution kernel for the input dataset.
 
-        Convolves the model with the PSFKernel at the center of the dataset.
+        Convolves the model with the IRFs.
```
registerrier, line 219: Maybe specify: `with the IRFs at the center of the dataset (or the nearest position with non-null exposure)`.

gammapy/estimators/map/ts.py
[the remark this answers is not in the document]
registerrier: Note that this is a pre-existing code, and possibly buggy indeed because what is passed is the `exposure.data`. 
A possible solution is to replace `"cm2s"` by `dataset.exposure.unit`

registerrier (changes requested):
Thanks @bkhelifi . This looks good. 
I have left a few minor comments. 
Please remove the `assert_raises` which we never use in gammapy and use instead `with pytest.raises`.

gammapy/estimators/map/ts.py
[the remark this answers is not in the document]
bkhelifi: But this contradicts a bit the initial comment of Quentin... What's happening if `data.exposure.unit` is not in `"cm2s"`?
This code allows to fix this case...
What is the consensus then?

PR #3799, opened 2022-02-13
Remove MapAxis name setter

registerrier:
**Description**

This pull request solves #3781 and remove `MapAxis.name` setter.

**Dear reviewer**

Issue #3781, opened 2022-02-01
Remove MapAxis.name setter


adonath:
I think we should remove the `MapAxis.name` setter. In Gammapy we rely internally on the correct definition of the axis names. So modifying an axis name in-place can cause trouble in other places. The recommended pattern should be to make an explicit copy to change the name like `axis.copy(name="new-name")`.

PR #4022, opened 2022-07-06
Introduce consistent .rename_axes and .rename API for maps

QRemy:
Rename Geom axes
-could be used in https://github.com/gammapy/gammapy/pull/4018#discussion_r913135787
-could be used in `Geom.as_energy_true`
-could  also be useful to rename axes of FITS maps that do not follow the expected convention.

### DIFF
--- a/gammapy/maps/axes.py
+++ b/gammapy/maps/axes.py
@@ -340,15 +340,22 @@ class MapAxis:
                 unit=self.unit,
             )
     
-    def rename(self, new_name):
+    def rename(self, new_name, copy=True):
         """Rename the axis.
     
         Parameters
         ----------
         new_name : str
-            The new name for the column
+            The new name for the axis.
+        copy : bool
+            Copy or change in place.
         """
-        self._name = new_name
+        if copy:
+            axis = self.copy()
+        else:
+            axis=self
+        axis._name = new_name
+        return axis
 
     def format_plot_xaxis(self, ax):
         """Format plot axis
@@ -2055,7 +2062,7 @@ class MapAxes(Sequence):
             raise ValueError(message)
 
 
-    def rename(self, names, new_names):
+    def rename(self, names, new_names, copy=True):
         """Rename the axes.
     
         Parameters
@@ -2063,15 +2070,21 @@ class MapAxes(Sequence):
         names : list or str
             Names of the axes
         new_names : list or str
-            New names of the axes (list must be of same length than `names`)
+            New names of the axes (list must be of same length than `names`).
+        copy : bool
+            Copy or change in place.
         """
+        if copy:
+            axes = self.copy()
+        else:
+            axes = self
         if isinstance(names, str):
             names = [names]
         if isinstance(new_names, str):
             new_names = [new_names]
         for name, new_name in zip(names, new_names):
-            self[name].rename(new_name)
-
+            self[name].rename(new_name, copy=False)
+        return axes
 
     @property
     def center_coord(self):
--- a/gammapy/maps/axes.py
+++ b/gammapy/maps/axes.py
@@ -102,7 +102,9 @@ class MapAxis:
 
     # TODO: Cache an interpolation object?
     def __init__(self, nodes, interp="lin", name="", node_type="edges", unit=""):
-        self._name = name
+        
+        if not isinstance(name, str):
+            raise TypeError(f"Name must be a string, got: {type(name)!r}")
 
         if len(nodes) != len(np.unique(nodes)):
             raise ValueError("MapAxis: node values must be unique")
@@ -116,6 +118,7 @@ class MapAxis:
         else:
             nodes = np.array(nodes)
 
+        self._name = name
         self._unit = u.Unit(unit)
         self._nodes = nodes.astype(float)
         self._node_type = node_type
@@ -2060,7 +2063,7 @@ class MapAxes(Sequence):
             raise ValueError(message)
 
 
-    def rename(self, names, new_names):
+    def rename_axes(self, names, new_names):
         """Rename the axes.
     
         Parameters
@@ -2082,7 +2085,7 @@ class MapAxes(Sequence):
             new_names = [new_names]
         for name, new_name in zip(names, new_names):
             axes[name]._name = new_name
-        return axes
+        return  axes
 
     @property
     def center_coord(self):
--- a/gammapy/maps/axes.py
+++ b/gammapy/maps/axes.py
@@ -339,6 +339,16 @@ class MapAxis:
                 node_type=node_type,
                 unit=self.unit,
             )
+    
+    def rename(self, new_name):
+        """Rename the axis.
+    
+        Parameters
+        ----------
+        new_name : str
+            The new name for the column
+        """
+        self._name = new_name
 
     def format_plot_xaxis(self, ax):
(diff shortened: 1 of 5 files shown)


adonath:
Thanks @QRemy! I remember we discussed this functionality before:
- https://github.com/gammapy/gammapy/issues/3781
- https://github.com/gammapy/gammapy/pull/3799

The conclusion was the recommended pattern to rename axis should be `new_axis = axis.copy(name="new-name")`. So making an explicit copy. As we rely on axis name conventions in Gammapy the in-place name change could break things unexpectedly. I think the renaming is typically done from existing objects, such as:

```python3
energy_axis = dataset.psf.psf_map.geom.axes["energy_true"]
energy_axis.rename("energy")

new_map = WcsNDMap.create(axes=[energy_axis])

# This fails then for not so obvious reasons...
dataset.npred()
```
So we decided the `.copy()` pattern should be preferred:

```python3
energy_axis = dataset.psf.psf_map.geom.axes["energy_true"]
energy_axis = energy_axis.copy(name="energy")

new_map = WcsNDMap.create(axes=[energy_axis])

# Does not fail...
dataset.npred()
```

Which advantage do you see in the in-place name update? Or do we just need to document the `.copy(name="new-name")` pattern better?

QRemy:
If we want to modify the axes name attached to a Map or a Geom it is not trivial (just looking how `Geom.as_energy_true` is implement). `Geom.rename_axes` is more generic and closer to the pattern in astropy `Table.rename_columns`. I could add an option `copy=True` by default to match the current behavior.  The in place change could be usefull to change the axes names of Map that do not follows the gammapy conventions without coying them.

adonath:
@QRemy I think there are two separate things to be discussed:
- In-place modification vs copy
- Convenience API for renaming axes / axes in existing geometries

I fully agree that convenience methods such as:
```python

new_axes = axes.rename_axes(names={"energy": "energy_true", "foo": "bar"})
new_geom = geom.rename_axes(names={"energy": "energy_true", "foo": "bar"})
new_map = map.rename_axes(names={"energy": "energy_true", "foo": "bar"})
```
...would be very useful!

I think a dedicated `axis.rename(name="new-name")`, would not be needed, because it will be an alias for `axis.copy(name="new-name")`, but could be introduced for API consistency with the `.rename_axes()`.

However we should still (as shown in the example above) keep the copy and return behavior. This is just how our API is defined now. All methods on `MapAxis`, `Geom` and `Map` return a new modified object and do not support in place operations. There is exactly one exception, which is `Map.stack()`. The only reason it exists is performance. Always creating a full copy of the data would be the bottleneck and slow down stacking a lot. 

It is true that e.g. `Table.rename_column()` does an in-place update, but e.g. `Table.group_by()` does not. I'm not sure whether they tried to achieve API consistency here at all.

gammapy/maps/geom.py
```diff
@@ -349,11 +349,31 @@ def slice_by_idx(self, slices):
...
+            New names of the axes (list must be of same length than `names`).
+
+        Returns
+        -------
+        geom : `~Geom`
+            Renamed geometry.
+        """
+        geom = self.copy()
+        geom._axes = geom.axes.rename(names, new_names)
+        return geom
```
adonath, line 370:
```suggestion
        axes = self.axes.rename(names=names, new_names=new_names)
        return self._init_copy(axes=axes)
```

gammapy/maps/geom.py
```diff
@@ -349,11 +349,31 @@ def slice_by_idx(self, slices):
...
+
     @property
     def as_energy_true(self):
         """If the geom contains an energy axis rename it to energy true"""
-        energy_axis = self.axes["energy"].copy(name="energy_true")
-        return self.to_image().to_cube([energy_axis])
+        return  self.rename_axes("energy", "energy_true", copy=True)
```
adonath, line 375:
```suggestion
        return  self.rename_axes(name="energy", new_name="energy_true")
```

gammapy/maps/core.py
```diff
@@ -132,6 +132,25 @@ def quantity(self, val):
...
+            New names of the axes (list must be of same length than `names`).
+
+        Returns
+        -------
+        geom : `~Map`
+            Renamed Map.
+        """
+        map_copy = self.copy()
+        map_copy._geom = map_copy.geom.rename_axes(names, new_names)
+        return map_copy
```
adonath, line 152:
```suggestion
        geom = self.geom.rename_axes(names=names, new_names=new_names)
        return self._init_copy(geom=geom)
```

gammapy/maps/axes.py
```diff
@@ -2044,6 +2059,31 @@ def assert_names(self, required_names):
...
+        axes : `MapAxes`
+            Renamed Map axes object
+        """
+        axes = self.copy()
+        if isinstance(names, str):
+            names = [names]
+        if isinstance(new_names, str):
+            new_names = [new_names]
+        for name, new_name in zip(names, new_names):
+            axes[name]._name = new_name
```
adonath, line 2084: This does not really make a copy of the `MapAxis`, just use the `MapAxis.rename()` method here you introduced earlier.

gammapy/maps/axes.py
```diff
@@ -339,6 +339,21 @@ def to_node_type(self, node_type):
...
+        Parameters
+        ----------
+        new_name : str
+            The new name for the axis.
+
+        Returns
+        -------
+        axis : `~gammapy.maps.MapAxis`
+            Renamed MapAxis
+        """
```
adonath, line 355: We could maybe add a check to `MapAxis.__init__`, that `name` is actually as string...not sure we have this.

adonath (changes requested):
Thanks @QRemy, I have left some inline comments and suggestions for simplification (please accept those if you agree...).

gammapy/maps/axes.py
QRemy: but here axes is self.copy() which already copied each axis

gammapy/maps/geom.py
```diff
@@ -349,11 +349,30 @@ def slice_by_idx(self, slices):
         axes = self.axes.slice_by_idx(slices)
         return self._init_copy(axes=axes)
 
+
+    def rename_axes(self, names, new_names):
+        """Rename the axes in place.
```
adonath, line 354:
```suggestion
        """Rename axes contained in the geometry
```

QRemy:
Thanks @adonath for all the corrections

adonath:
@registerrier Can you please do the final review here? The PR is fine from my side now...

gammapy/maps/core.py
```diff
@@ -132,6 +132,24 @@ def quantity(self, val):
         self.data = val.value
         self._unit = val.unit
 
+    def rename_axes(self, names, new_names):
+        """Rename the Map axes.
+
```
registerrier, line 137: Maybe specify that this will create a full copy of the `Map` as this might not be obvious from the name.

registerrier (approved):

I have just left a small inline comment.

gammapy/maps/core.py
adonath: By default all maps methods do a copy, except for `.stack()`,  so I think it's fine to just state the exception.

adonath (approved):

Issue #4010, opened 2022-06-27
Move "psf_after_edisp" flag from model to dataset


QRemy:
For now evaluation of models for HAWC datasets requires to set `model.apply_irf["psf_after_edisp"]=True`.
This implies that a different  model have to be defined for HAWC and would complexity joint analyses with other instruments.
The "psf_after_edisp" flag should rather been defined at the dataset level, and ideally not set by the user but deduced from the energy-axis name in the PSF  or from metadata.

PR #4025, opened 2022-07-14
Add support for PSF in reco energy

QRemy:
Add support for PSF in reco energy for HAWC DL3.
Set apply_apply_after_edisp in MapEvaluator directly from psf energy axis name.
(For now it's only a draft and it will require example data to be tested.)

I also noticed that the energy parameter in PSF methods is always called energy_true, this is problematic if both are allowed. To be consistent everywhere I think in class method the energy parameters should always be called energy even if the axis is energy_true, this is typically what we have in model.evaluate(energy). Alternatively a  sub-class can be introduced to redefine only the methods with energy as parameter.


bkhelifi:
Hi Luca,
Can you precise the Use Case associated with the PSF in reconstructed energy?

gammapy/irf/psf/map.py
```diff
@@ -135,15 +145,15 @@ def _get_irf_coords(self, **kwargs):
...
         """Containment at given coords
 
         Parameters
         ----------
         rad : `~astropy.units.Quantity`
             Rad value
-        energy_true : `~astropy.units.Quantity`
-            Energy true value
+        energy : `~astropy.units.Quantity`
+            Energy value
```
LauraOlivera, line 156: Given that we generally call "energy" to the reco energy, I would specify in the docstring that this energy value could be true energy or reco energy depending on how the PSF is given

gammapy/irf/psf/map.py
```diff
@@ -155,19 +165,19 @@ def containment(self, rad, energy_true, position=None):
...
         """Containment at given coords
 
         Parameters
         ----------
         fraction : float
             Containment fraction
-        energy_true : `~astropy.units.Quantity`
-            Energy true value
+        energy : `~astropy.units.Quantity`
+            Energy value
```
LauraOlivera, line 180: Same comment as above

gammapy/irf/psf/map.py
```diff
@@ -16,7 +16,7 @@
 
 
 class IRFLikePSF(PSF):
-    required_axes = ["energy_true", "rad", "lat_idx", "lon_idx"]
+    IRFLikePSF = ["rad",  "energy_true", "rad", "lat_idx", "lon_idx"]
```
LauraOlivera, line 19: why is "rad" here twice?

LauraOlivera: maybe you meant to add "energy"?

LauraOlivera (changes requested):
Hi  @QRemy ! thanks a lot for doing this
I have left a few minor comment, otherwise it looks good so far

Please add/modify the missing test and then I'll take a look again :+1:

gammapy/irf/psf/map.py
LauraOlivera: But I see what you mean about the confusion in general with energy/energy true... I think the docstrings should be clearer that both types of energy are valid as input

adonath:
@LauraOlivera As for HAWC the DL4 level files are used, introducing a `PSFMap` in reconstructed energy would not require any change to GADF. Is this correct?

gammapy/irf/psf/map.py
QRemy: For clarity I introduced a second class PSFMapReco that redefine only the methods that take energy as argument to have a correct docstring

gammapy/irf/psf/map.py
QRemy: removed this change

QRemy:
The PR is ready for review and resolves #4010.

gammapy/irf/psf/map.py
```diff
@@ -523,6 +534,129 @@ def peek(self, figsize=(12, 10)):
...
 
         axes[3].set_title("Containment radius at 1 TeV")
-        self.containment_radius_map(energy_true=2 * u.TeV).plot(
+        kwargs = {self.energy_name:2 * u.TeV}
+        self.containment_radius_map(**kwargs).plot(
             ax=axes[3], add_cbar=True
         )
+
+
+class PSFMapReco(PSFMap):
```
adonath, line 543: I understand correctly, once we have the `PSFMap.energy_axis_name` attribute we don't really need the `PSFMapReco` class except for documentation. I this correct? In this case we could think about just to extend the current documentation of `PSFMap`. and not introduce a new class. What do you think @QRemy  and @LauraOlivera?

gammapy/irf/psf/map.py
```diff
@@ -428,20 +438,21 @@ def plot_containment_radius_vs_energy(
...
-            )
+            radius = self.containment_radius(frac, energy_true, position)
             label = f"Containment: {100 * frac:.1f}%"
             with quantity_support():
                 ax.plot(energy_true, radius, label=label, **kwargs)
 
         ax.semilogx()
         ax.legend(loc="best")
         ax.yaxis.set_major_formatter(FormatStrFormatter("%.2f"))
-        ax.set_xlabel(f"Energy ({ax.xaxis.units})")
```
adonath, line 444: Just replace by `energy_axis.format_plot_xaxis(ax=ax)`

gammapy/irf/psf/map.py
```diff
@@ -67,8 +67,13 @@ class PSFMap(IRFMap):
     """
 
     tag = "psf_map"
+    energy_name = "energy_true"
     required_axes = ["rad", "energy_true"]
 
+    @classmethod
+    def as_energy(cls):
+        return PSFMapReco
```
adonath, line 75: This does not seem to do anything...just remove?

LauraOlivera (dismissed):
Looks good to me, @QRemy ! I only had some minor question.

I also downloaded the branch locally and tested it with the public HAWC data, to start prototyping how the tests will look like once they include real data. I will add new tests once that is ready (in a different PR)

gammapy/irf/psf/map.py
```diff
@@ -523,6 +534,129 @@ def peek(self, figsize=(12, 10)):
...
+    Parameters
+    ----------
+    psf_map : `~gammapy.maps.Map`
+        the input PSF Map. Should be a Map with 2 non spatial axes.
+        rad and energy axes should be given in this specific order.
+    exposure_map : `~gammapy.maps.Map`
+        Associated exposure map. Needs to have a consistent map geometry.
+    """
+
+    tag = "psf_map"
```
LauraOlivera, line 555: isn't the fact that the tag is the same as the `PSFMap` an issue? I am having trouble opening `PSFMapReco` objects from index tables because of it. But maybe I am doing something wrong. Shouldn't it be `psf_map_reco`?

gammapy/irf/psf/map.py
LauraOlivera: I see your point @adonath and I don't really have a strong opinion.

I think it would be good to unify both and remove code duplication, but also this PR is almost ready and we do have the deadline of the feature freeze... so I would rather this go in as is than need to re-do it and miss the cut. 

What do you think, @QRemy?

adonath: > so I would rather this go in as is than need to re-do it and miss the cut.

I think the effort is limited here. Handling both cases with the `PSFMap` will require adapting the docstrings and renaming the argument name. That's basically it...

gammapy/irf/psf/map.py
adonath: Yes, if we keep the `PSFMapReco` it should get a separate tag.

gammapy/irf/psf/map.py
adonath: OK, looking more closely the issue that we use the hard-coded axes names for IRFs, like https://github.com/gammapy/gammapy/blob/master/gammapy/irf/psf/map.py#L70.

So handling both cases with the `PSFMap`, would require a new pattern for allowed axes names, which deviated from put current coding pattern. So lets keep the `PSFMapReco`! However we should rename, because our current scheme of naming Classe puts the base class name at the end. Our convention uses "energy" as default for reconstructed energy, however an `EnergyPSFMap` sounds weird. So maybe the best compromise is something like `EnergyRecoPSFMap`. But @registerrierm should comment once again...

registerrier (commented):
Thanks @QRemy.

Indeed, we should have a distinct class to follow the one class/one set of required axes logic. 
The `energy` vs `energy_true` argument issue is forcing to overload many functions but I assume this is acceptable. At least, it is likely impossible to find a better solution that is not API-breaking.
Regarding the name, indeed it should be `EnergyPSFMap` vs `EnergyTruePSFMap`, if we were to follow the logic fully. But given the latter does not exist,  `EnergyRecoPSFMap` or more simply `RecoPSFMap` is probably acceptable since we only need to indicate this is different from the canonical `PSFMap`.

gammapy/irf/psf/map.py
```diff
@@ -67,8 +67,13 @@ class PSFMap(IRFMap):
     """
 
     tag = "psf_map"
+    energy_name = "energy_true"
```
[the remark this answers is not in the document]
adonath: I agree, that's on place less where to hard-code the axis name.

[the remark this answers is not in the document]
QRemy: done

gammapy/irf/psf/map.py
QRemy: done

gammapy/irf/psf/map.py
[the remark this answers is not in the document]
QRemy: yes removed

gammapy/irf/psf/map.py
```diff
@@ -523,6 +531,132 @@ def peek(self, figsize=(12, 10)):
...
+
+class RecoPSFMap(PSFMap):
+    """Class containing the Map of PSFs in reconstructed energy and allowing to interact with it.
+
+    Parameters
+    ----------
+    psf_map : `~gammapy.maps.Map`
+        the input PSF Map. Should be a Map with 2 non spatial axes.
+        rad and energy axes should be given in this specific order.
+    exposure_map : `~gammapy.maps.Map`
```
adonath, line 548: @LauraOlivera Do we have an exposure map in reconstructed energy here? Computed with the same spectrum as the PSF?

LauraOlivera: I've never seen that needed or used for HAWC... but I guess it is a possibility, and yes it would need to be a reco energy map in that case.

LauraOlivera (approved):
Thanks @QRemy it looks good to me... I tested it locally with the HAWC files and it looks OK as far as I could test.

There are still some small changes needed to make the `RecoPSFMap` usable in the full data reduction chain, like for example creating `MapDataset` objects with the right PSF geometry 
https://github.com/gammapy/gammapy/blob/0d27bd532a1d429286713905bc52352b5ea23f8a/gammapy/datasets/map.py#L564-L565

but I guess this is is something we can do after this PR is merged.

QRemy:
Thanks @LauraOlivera, I added support for the  `RecoPSFMap` in `MapDataset.from_geoms`.

bkhelifi:
Dear all,
With this PR, the processing of HAWC data will be really possible. Thus, if all agree, I propose to you to make some update of the doc in the general parts in order to change the status of "some support to HAWC" to "used to analyse data from HESS, ... and HAWC".

LauraOlivera (approved):
Thanks @QRemy !
I was now able to fully run the analysis of the public HAWC data with this branch. The fit results are virtually identical.
image

adonath:
@QRemy Is this ready to merge from your side?

@LauraOlivera Will you have time to work on a follow up PR which actually tests this with real HAWC data?

QRemy:
ready to merge from my side

adonath:
@registerrier Any final comments from your side?

registerrier (approved):
Thanks @QRemy . This looks good to me. No further comment.

PR #4026, opened 2022-07-15
Fix TemplateSpatialModel overwrite

QRemy:
Fix TemplateSpatialModel overwrite


adonath (approved):
Thanks @QRemy! No comment from my side, I think we can skip the unit test here...

registerrier (approved):
Thanks @QRemy!

PR #4037, opened 2022-07-27
Implement IRF.slice_by_idx()

adonath:
I came across a use-case for this method in context of the Gammapy paper, so I decided to just implement it here. It includes:
- Implement `IRF.slice_by_idx()`
- A unit test in `gammapy/irf/tests/test_core.py`

Currently there is no support for integer indexing such as `.slice_by_idx({"energy": 0})`, as we have more maps. This is because for IRFs we have fixed axes names, so we cannot drop an axes from the IRF definition.

### DIFF
--- a/gammapy/irf/core.py
+++ b/gammapy/irf/core.py
@@ -554,6 +554,30 @@ class IRF(metaclass=abc.ABCMeta):
             data=data, axes=axes, meta=self.meta.copy(), unit=self.unit
         )
 
+    def slice_by_idx(self, slices):
+        """Slice sub IRF from IRF object.
+
+        Parameters
+        ----------
+        slices : dict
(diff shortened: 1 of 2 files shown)


adonath:
This currently blocks the paper build, could somebody please do a quick review here? Thanks!

maxnoe (approved):
Looks good and useful

QRemy (approved):
Looks good, maybe add that integer indexing is not supported the docstring.

adonath:
@QRemy I think the doctsring mentions correctly, that it supports `slice()`, while `Map.slice_by_idx()` additionally mentions the integer index. So I think it's fine like this...

Thanks @QRemy and @maxnoe for the quick response!

Issue #4008, opened 2022-06-25
CompoundSpectralModel Norm issue


MRegeard:
**Gammapy version**
Gammapy 0.20.1

**Bug description**
It appears that while fixing the bug related to FluxPoints when using a CompoundSectralModel, it is now (Gammapy-0.20.1) impossible to create a CompoundSpectralModel with more than one free amplitude in the spectral models. 

**Expected behavior**
I think that it should not be the case, at least for the fitting of the data. In Gammapy-0.19, this was possible, however, it was not possible to compute FluxPoints.
Of course, in the case of degenarated models, amplitude should be fixed, but I think that it is the responsibility of the user rather than an obligation for the usage of CompoundSpectralModel. 
But then, it raise the issue of how do we compute FluxPoints with multiple free amplitude in CompoundSpectralModel. 

**Temporary fix**
A possible fix to this issue is to set _is_norm to False to each amplitude in the CompoundSpectralModel but one. However, this is_norm feature does not seems very clear (at least to me), and should probably be define in an other way.

adonath:
Thanks @MRegeard!

In the Gammapy dev meeting we agreed that it would be the desired behavior to check for the uniqueness of the free amplitude parameter. This is done here: https://github.com/gammapy/gammapy/blob/master/gammapy/modeling/models/cube.py#L86. The reason is that two amplitude parameters will be correlated 100% in a fit and will never yield stable results. I don't understand where the limitation to a single free parameter leads to undesired behavior in your case.

adonath:
>  how do we compute FluxPoints with multiple free amplitude in CompoundSpectralModel.

The answer here is that this is unfeasible and should never be done. The two different amplitude parameters will be correlated 100%. In case you have done this before, you should not trust those previous results and rely on the new behavior instead.

registerrier:
I think the issue here is that we overlooked that some models can have independent and not correlated amplitudes: e.g. PL + gaussian line  or if you do a fit over a sufficiently broad range you can have several PLs whose amplitudes are controlled by different datasets (e.g. Fermi + VHE). 

This type of use case is no longer supported, in particular in the absence of a setter for `Parameter.is_norm`.

registerrier:
> So the current check is strictly speaking only valid if the operator is multiplicative?

Yes I would think so, what we want to avoid is having a model with a free norm multiplying a model with a free amplitude.

adonath:
The action item here for v1.0 is to additionally check the operator to be multiplicative...

[...]

PR #4040, opened 2022-07-27
Usage of Compound spectral models for a SkyModel

bkhelifi:
**Description**
This pull request allows to fix the issue associated to the sum of compound models mentioned in the Issue #4008 . The sum of spectral models is allowed without restriction. Concerning the multiplication, one checks if there is only one free norm parameter in order to avoid degeneracy.
This scheme will allow:
- sum any spectral models without restriction (new feature)
- multiply spectral models if there is only 1 free norm parameter

**Dear reviewer**
The fix is implemented. A new test function about compound model is added.
I have tried to fix the associated model gallery section. Let me know if its content is pedagogical enough. Also, I have not realised the associated CI test.

### DIFF
--- a/gammapy/modeling/models/cube.py
+++ b/gammapy/modeling/models/cube.py
@@ -1,5 +1,7 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 """Cube models (axes: lon, lat, energy)."""
+import operator
+
 import numpy as np
 import astropy.units as u
 from astropy.nddata import NoOverlapError
@@ -11,7 +13,7 @@ from gammapy.utils.fits import LazyFitsData
 from gammapy.utils.scripts import make_name, make_path
 from .core import Model, ModelBase, Models
 from .spatial import ConstantSpatialModel, SpatialModel
-from .spectral import PowerLawNormSpectralModel, SpectralModel, TemplateSpectralModel
+from .spectral import PowerLawNormSpectralModel, SpectralModel, TemplateSpectralModel, CompoundSpectralModel
 from .temporal import TemporalModel
 
 __all__ = [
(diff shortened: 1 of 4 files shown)


gammapy/modeling/models/cube.py
```diff
@@ -77,15 +79,16 @@ def __init__(
...
-            raise ValueError(
-                "Spectral model used with SkyModel can only have a single free norm type parameter."
+        if isinstance(self.spectral_model, CompoundSpectralModel):
+            is_free_norm = np.array(
+                [par.is_norm for par in spectral_model.parameters.free_parameters]
             )
 
+            if self.spectral_model.operator is operator.mul and np.sum(is_free_norm) > 1:
+                raise ValueError(
+                    "The multiplicative compound spectral model can only have a single free norm type parameter."
```
QRemy, line 89: Thanks @bkhelifi, I'm not sure this is the best place to raise this error as the multiplication may make sense if the norm correction applies to a linked model, in that case there is no 100% correlation with the amplitude.
Imagine a joint fit of two dataset with a source fitted by two models, one for each dataset,  with linked amplitude but the second dataset have in addition a norm correction to compensate for systematic bias in flux within a given range (eventually linked between multiple sources for the second instrument).

So I think we should remove the error raise here and force the secondary norm parameters of a compound model to be frozen when reoptimize=True in the FluxPointsEstimator to solve the original problem. 

In a separate PR, the error raise could be introduced at the fit level in a specific method that verify the validity of the full parameters list (so there we can check for links) and we could eventually add other conditions to this method in future (no free parameters, value outside min/max range, bad parameter scaling with value below precision...).

adonath: Thanks @QRemy, indeed this is a use-case I haven't thought of either. So +1 for your proposed solution...

bkhelifi: My initial thought was indeed to make nothing within the SkyModel, but only in the FluxPointEstimator.
For a 3D analysis, it think that the user could/should understand when a model can not be fit because of degeneracy.

So, what is the exact proposal?
- remove simply the currents line 82->89
- make an other PR for the FluxPointEstimator to raise an error when there is a degeneracy? I think that this is up to the user to choose what to froze and not up tp Gammapy to chose what/guess is 'secondary' parameter..

QRemy: Actually only the selected norm appears in the spectral model parameters after we apply the ScaleModel so there is mo issue with compound model.

bkhelifi:
Thanks @QRemy. I think that you can thus make the final review

QRemy (approved):
Thanks @bkhelifi, no further comments.

adonath (approved):
Thanks @bkhelifi, no further comments from my side either...

Issue #4036, opened 2022-07-26
Use equal aspect ratio in Background.plot_at_energy()


adonath:
The current plots are distorted https://docs.gammapy.org/0.20.1/tutorials/data/cta.html#Background One should use an equal aspect ratio here...

PR #4042, opened 2022-07-28
Fix the plot height when there is a color bar in Background.plot_at_energy()

bkhelifi:
**Description**
This pull request is associated to the Issue #4036 . The plot function `background.plot_at_energy()` deforms the sub-panel plot, with an height different from the width. 

**Dear reviewer**
The issue is coming to the fact that the height computation was not taking into account of the existence or not of the color bar. With the color bar, the plot itself is smaller than the sub-panel width.
This PR fixes most of the distorsion, even if one still does not correct for the axis label size and the inter-panel margins.

### DIFF
--- a/codemeta.json
+++ b/codemeta.json
@@ -589,9 +589,9 @@
         }
     ],
     "codeRepository": "https://github.com/gammapy/gammapy",
-    "datePublished": "2020-11-19",
+    "datePublished": "2022-06-17",
     "description": "Gammapy analyzes gamma-ray data and creates sky images, spectra and lightcurves, from event lists and instrument response information; it can also determine the position, morphology and spectra of gamma-ray sources. It is used to analyze data from H.E.S.S., Fermi-LAT, and the Cherenkov Telescope Array (CTA).",
-    "identifier": "https://doi.org/10.5281/zenodo.5721467",
+    "identifier": "https://doi.org/10.5281/zenodo.6552377",
     "keywords": [
         "Astronomy",
         "Gamma-rays",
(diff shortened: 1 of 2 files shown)


adonath:
Thanks @bkhelifi! We can also force the axis ratio using `matplotlib`, see https://matplotlib.org/stable/gallery/subplots_axes_and_figures/axis_equal_demo.html

Maybe this is the simpler option?

bkhelifi:
Great option!
With the initial computation of the height, the color bar is going above/below the plot: 
Capture d’écran du 2022-07-28 20-10-17

If you use both this option and my proposed computation of the height, I think we are almost perfect:
Capture d’écran du 2022-07-28 20-21-14

adonath:
Thanks @bkhelifi! I agree the adjusted size of the colorer looks much nicer, however does it work for all grids? Probably not, it seems to me that with the magic numbers it is adapted to the 2x3 grid in the example. In general users could make it nicer, just by adapting the figure size. So to make this example nice maybe just adapt the figure size from outside? Like:

```python3
plot_at_energy(.., figsize=(.., ...))
```

This is a more generic solution, where we just show users, that they can adapt the figure size as needed. This will also work for any other `.peek()` or `Map.plot_grid()` method we have.

bkhelifi:
A question to @adonath : do you know why the Scrutinizer CI takes so long (>12h)? This is strange...

adonath:
As the issue with colormaps appears in many places, we should maybe introduce a more generic solution like https://matplotlib.org/stable/gallery/axes_grid1/demo_colorbar_with_axes_divider.html

However this can be done in a follow-up PR...

adonath (approved):
Thanks @bkhelifi! Let's merge this now and rather work on a follow up PR to add a general colorbar method...

Issue #3750, opened 2022-01-27
Dataset serialisation should preserve names


AtreyeeS:
`MapDataset.read()` should ideally read back the original dataset name by default instead of generating a random string. Is there a specific reason why we can't read the original name?

adonath:
Good point @AtreyeeS, I think there is actually no good reason why the name of the dataset is not serialized.

adonath:
Would be good to have this for v1.0rc

AtreyeeS:
What would be a good way to do it? Add an information in `meta` (but we decided not to use the meta_table internally in gammapy anywhere).
While serialising through `Datasets` we force the filename to be the dataset name. But that is not a good idea for serialising individual dataset.

registerrier:
There is already a specific behavior implemented for `SpectrumDatasetOnOff` in ogip format where the name of the dataset is taken from  the `OBSID` keyword. This default behavior should be kept.

An option for the other cases is to use a specific keyword that could be serialized in most hdus.

adonath:
Solved in #4085.

PR #4045, opened 2022-08-01
Better management of the meta data and the names of the Datasets

bkhelifi:
**Description**

This pull request aims to our small imperfections of managing the names of datasets. It is associated with the issue #3750 . As it is associated to the serialisation of meta data, this PR corrects some of our small imperfections at this regards.

**Dear reviewer**
As the meta data of the serialisation and the names of the datasets, this PR contains a correction of the management of both the meta data and the dataset names.
I profit of this PR for a small add-on: adding some provenance information for the datasets created from an other dataset. For example, the `dataset` created with `to_image` has now as name `f"{self.name}-image"` if `self.name` exists.
I might need to add a test on the meta data of a `Datasets` build from several datasets.

### DIFF
--- a/gammapy/visualization/utils.py
+++ b/gammapy/visualization/utils.py
@@ -1,8 +1,6 @@
 import numpy as np
 from scipy.interpolate import CubicSpline
 import matplotlib.pyplot as plt
-from gammapy.maps import MapAxis
-from gammapy.maps.utils import edges_from_lo_hi
 
 
 __all__ = [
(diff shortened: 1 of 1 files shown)


adonath:
Conclusion from the dev call:
- The dataset name should be serialized with the FITS file in the header of the primary HDU. The `.meta_table` is currently fully optional, while the `.name` is not.
- We make the dataset name keyword in the `.yaml` serialization optional and it falls back to the name stored in the FITS file. However this way users can overwrite names if need on I/O for datasets.
- The handling of the `Datasets.meta_table` should be an independent PR.

adonath:
@bkhelifi After #4085 has been merged, what is your suggestion for this PR here?

PR #4050, opened 2022-08-04
Test if there are too many free norm parameters in the FluxEstimator

bkhelifi:
**Description**
This pull request is the second step associated to the Issue #4008. In order to avoid unseen parameters' degeneracy during the flux computation, it adds a test on multiple free norm parameters

**Dear reviewer**
This PR is somehow a patch to this estimator. A more coherent correction would be needed, e.g. by adding in the constructor or the run function the name of the norm parameter to adjust. But this is a change of the API, and thus can not go into the v1.0.


gammapy/estimators/flux.py
```diff
@@ -113,6 +113,14 @@ def get_scale_model(self, models):
...
         scale_model = ScaleSpectralModel(ref_model)
 
+        is_free_norm = np.array(
+            [par.is_norm for par in ref_model.parameters.free_parameters]
+        )
+        if np.sum(is_free_norm) > 1:
+            raise ValueError(
+                "Spectral model used can only have a single free norm type parameter."
+            )
+
```
QRemy, line 123: There is always only one norm (and no other parameters) in the scale model so this in not necessary

bkhelifi: I don't think so... Have a look on the tests;)

QRemy: because `is_free_norm` test  the ref_model  but it will be replace by
the scale model in the estimator so the will issue disappears

bkhelifi: The tests work... how can you explain that, then?

bkhelifi:
Thanks a lot @QRemy !!!!

gammapy/estimators/flux.py
```diff
@@ -113,16 +113,16 @@ def get_scale_model(self, models):
         ref_model = models[self.source].spectral_model
         scale_model = ScaleSpectralModel(ref_model)
 
-        for scaled_parameter in ref_model.parameters:
-            if scaled_parameter.is_norm:
-                break
-        else:
+        norms = Parameters([p for p in ref_model.parameters if p.is_norm])
+        if len(norms) == 0 or len(norms.free_parameters)>1 :
```
QRemy, line 117: This still pick the first norm when all parameters are frozen as it should be the case if there only one norm to be backward compatible. But we could also raise if (len(norms>1) and len(norms.free_parameters)==0).

registerrier (approved):
Thanks @bkhelifi . This looks good to me.

The behavior is backward compatible if all norm parameters are frozen, in which case it will take the first one. 

Maybe add a line in the docstring to explain that there should be only one free norm or amplitude  parameter for the Estimator to run?

bkhelifi:
Thanks @registerrier. The docstring contains this information now

adonath (approved):
Thanks @bkhelifi, no further comments from my side either.

Issue #501, opened 2016-04-05
Use astropy table indexing in ObservationTable


joleroi:
https://github.com/gammapy/gammapy/blob/master/gammapy/data/observation.py#L53

registerrier:
This requires adapting `ObservationTable.select_obs_id` to make it work with `Table` indexing, e.g. by introducing an index `obs_table.add_index('OBS_ID')`

registerrier:
This is implemented in #4051 . Closing now.

PR #4051, opened 2022-08-04
Using astropy Table indices on ObservationTable

registerrier:
**Description**

This pull request is a test to solve the old issue #501 . 

In practice, this comes with a significant performance loss and I am not sure to understand why. 
I have noticed that the simple index scheme we have now is more efficient. It might break in a number of occasion though.

Using the old implementation we have:
```
from gammapy.data.tests.test_obs_table import make_test_observation_table
table=make_test_observation_table(n_obs=100000)
%%timeit
table.select_obs_id([732, 3523,9399])

table.add_index("OBS_ID")
%%timeit
ObservationTable(table.loc["OBS_ID", [732, 3523,9399]])
```
The first one runs in 0.252 ms while the second runs in 1.53 ms.

Using only indices:
```
%%timeit
table.get_obs_idx([732, 3523,9399])

%%timeit
table.loc_indices["OBS_ID", [732, 3523,9399]]
```
Here again,  the first one runs in 4.4 mus while the second runs in 168 mus.

Of course a major issue with the current implementation is the following:
```
ds = DataStore.from_dir("$GAMMAPY_DATA/hess-dl3-dr1")

ds.obs_table.select_obs_id([23523])

ds.obs_table.sort('MUONEFF')
ds.obs_table.select_obs_id([23523])
```

The latter case does not yield the correct entry... The new implementation makes this work.

Opinions ?

**Dear reviewer**

### DIFF
--- a/gammapy/data/obs_table.py
+++ b/gammapy/data/obs_table.py
@@ -57,6 +57,26 @@ class ObservationTable(Table):
         """Observation stop time (`~astropy.time.Time`)."""
         return self.time_ref + Quantity(self["TSTOP"], "second")
 
+    def get_obs_idx(self, obs_id):
+        """Get row index for given ``obs_id``.
+
+        Raises KeyError if observation is not available.
+
+        Parameters
+        ----------
+        obs_id : int, list
+            observation ids
+        Returns
+        -------
+        idx : list
+            indices corresponding to obs_id
+        """
+        try:
+            self.indices['OBS_ID']
+        except:
+            self.add_index('OBS_ID')
+        return np.atleast_1d(self.loc_indices['OBS_ID', obs_id])
+
     def select_obs_id(self, obs_id):
         """Get `~gammapy.data.ObservationTable` containing only ``obs_id``.
 
@@ -71,7 +91,8 @@ class ObservationTable(Table):
             self.indices['OBS_ID']
         except:
             self.add_index('OBS_ID')
-        return self.loc['OBS_ID', obs_id]
+        return self.__class__(self.loc['OBS_ID', obs_id])
+#        return self[self.get_obs_idx(obs_id)]
 
     def summary(self):
         """Summary info string (str)."""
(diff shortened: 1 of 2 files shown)


adonath:
Thanks @registerrier!

I think the performance overhead should go away, once one uses larger index arrays. Might be good to double check. In general making the indexing independent of the sorting of the table is probably a very good idea. So right now I'm leaning towards the new implementation...

registerrier:
I have tried to test performances creating an `ObservationTable` with 1e6 entries. I randomized the order. When selecting 1e3 obs_ids randomly chosen in the list, the performance difference seems even larger 2.1 ms for the current implementation and 52 ms for the proposed one.  The performance ratio seems to be the same for larger selections e.g. selecting 1e5 obs_ids takes respectively 326ms and 5s.

```
from gammapy.data.tests.test_obs_table import make_test_observation_table
from gammapy.data import ObservationTable

table=make_test_observation_table(n_obs=1000000)
table.sort('MUONEFF')

obs_ids = np.random.choice(table["OBS_ID"].data,1000,replace=False)

%%timeit
table.select_obs_id(obs_ids)
# 2.16 ms ± 77.5 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

%%timeit
ObservationTable(table.loc["OBS_ID", obs_ids])
# 53.6 ms ± 1.7 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

Note that removing the `lazyproperty` on the `ObservationTable._index_dict` only reduces performances by a factor of 3. So it seems we can get a correct behaviour w.r.t. `Table` manipulation with limited computing cost. It is another possibility.

All in all the computing time is limited and the possible choices might not be that relevant in the end.

adonath:
Here is a notebook with some indexing performance checks:

https://nbviewer.jupyter.org/github/mdmueller/astropy-notebooks/blob/master/table/indexing-profiling.ipynb

adonath (approved):
Thanks @registerrier! I think the performance here is not a concern and the simplification of the implementation helps as well as correct handling of a modified table. No further comments from my side...

adonath (approved):

bkhelifi (approved):
Thanks. Ok for me

adonath:
Thanks @registerrier!

PR #4058, opened 2022-08-25
Serialisation in the primary HDU of the Dataset name

bkhelifi:
**Description**
Based on the discussion described in the PR #4045, this pull request is one of the three aiming to deal correctly with the serialization of the `Dataset` names. This PR is then associated to the issue #3750.

This pull request is about the serialization of the dataset names with the FITS file in the header of the primary HDU

### DIFF
--- a/gammapy/datasets/map.py
+++ b/gammapy/datasets/map.py
@@ -1103,6 +1103,10 @@ class MapDataset(Dataset):
         exclude_primary = slice(1, None)
 
         hdu_primary = fits.PrimaryHDU()
+
+        header = hdu_primary.header
+        header['NAME'] = self.name
+
         hdulist = fits.HDUList([hdu_primary])
         if self.counts is not None:
             hdulist += self.counts.to_hdulist(hdu="counts")[exclude_primary]
@@ -1230,7 +1234,12 @@ class MapDataset(Dataset):
 
     @classmethod
     def _read_lazy(cls, name, filename, cache, format=format):
-        kwargs = {"name": name}
+        ds_name = make_name(name)
+        hdulist = fits.open(str(make_path(filename)))
+        if 'NAME' in hdulist[0].header and name is None:
+            ds_name = hdulist[0].header['NAME']
+        hdulist.close()
+        kwargs = {"name": ds_name}
         try:
             kwargs["gti"] = GTI.read(filename)
         except KeyError:
@@ -1289,7 +1298,6 @@ class MapDataset(Dataset):
         dataset : `MapDataset`
             Map dataset.
         """
-        name = make_name(name)
 
         if lazy:
             return cls._read_lazy(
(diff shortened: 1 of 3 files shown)


gammapy/datasets/map.py
```diff
@@ -1230,7 +1234,12 @@ def write(self, filename, overwrite=False):
 
     @classmethod
     def _read_lazy(cls, name, filename, cache, format=format):
-        kwargs = {"name": name}
+        ds_name = make_name(name)
+        hdulist = fits.open(str(make_path(filename)))
```
adonath, line 1238: As you open the hdulist here anyway there is not advantage in putting it in `_read_lazy`. So the extortion of the name could just always happen in `MapDataset.read()`

adonath: Just remove the modification here...

gammapy/datasets/map.py
```diff
@@ -1289,15 +1298,17 @@ def read(cls, filename, name=None, lazy=False, cache=True, format="gadf"):
...
-        name = make_name(name)
 
         if lazy:
             return cls._read_lazy(
                 name=name, filename=filename, cache=cache, format=format
             )
         else:
             with fits.open(str(make_path(filename)), memmap=False) as hdulist:
-                return cls.from_hdulist(hdulist, name=name, format=format)
+                ds_name = make_name(name)
```
adonath, line 1308: Again just remove the modification here...

gammapy/datasets/map.py
```diff
@@ -1289,15 +1298,17 @@ def read(cls, filename, name=None, lazy=False, cache=True, format="gadf"):
         dataset : `MapDataset`
             Map dataset.
         """
-        name = make_name(name)
```
adonath, line 1301:
```suggestion
if name is None:
    header = fits.getheader(filename)
    name = header.get("NAME", name)
name = make_name(name)
```

adonath (changes requested):
Thanks @bkhelifi I have left some minor comments for simplification...

gammapy/datasets/map.py
bkhelifi: In the read function, the name management is made for the non-lazy method. In the case of the lazy one, why we should not manage it?

gammapy/datasets/tests/test_map.py
```diff
@@ -1147,8 +1150,10 @@ def test_map_dataset_on_off_fits_io(images, lazy, tmp_path):
     if lazy:
         with pytest.raises(NotImplementedError):
             dataset_new = MapDatasetOnOff.read(tmp_path / "test.fits", lazy=lazy)
+            assert dataset_new.name == 'MapDatasetOnOff-test'
```
registerrier, line 1153: Should we expect something after an error is raised?

registerrier (dismissed):
Thanks @bkhelifi . This looks good. See minor inline comment.

One general question though. The ogip format for `SpectrumOnOffDataset.write` sets the  `dataset.name` to the `OBS_ID` keyword and `.read` takes the name from the `OBS_ID` not the `NAME` keyword. Since the two formats are independent that is probably not an issue, but this will require some documentation.

adonath (approved):
Thanks @bkhelifi, no further comments from my side!

registerrier (approved):
Thanks @bkhelifi 

No comments from my side.

PR #4047, opened 2022-08-02
Closing of contours

bkhelifi:
**Description**
This pull request anticipates the passage to `iminuit` V2.13 by testing or not if a contour should be closed before plotting it (issue #4031).

**Dear reviewer**
A very simple PR... I put a relative tolerance of 1.e-2, by hoping that this a good compromise (tests needed)

PR #4064, opened 2022-09-07
Remove iminuit version constraint

registerrier:
Signed-off-by: Terrier <rterrier@apc.in2p3.fr>

**Description**

This pull request removes the iminuit<2.13 constraint which is no longer necessary after #4047 . The failing test is modifed to support both cases.

**Dear reviewer**


adonath (approved):
Thanks @registerrier! I think this is uncontroversial so I'll just go ahead and merge...

Issue #4065, opened 2022-09-07
Circular import in visualization


dguevel:
**Gammapy version**
0.20.1

Are you using the latest stable version of Gammapy? (see https://gammapy.org/news.html)
Yes.

**Bug description**
``import gammapy.visualization'' throws a circular import error. Issue is resolved when using 0.19

**Expected behavior**
I expect to import gammapy.visualization

**To Reproduce**
Run ``import gammapy.visualization'' on v0.20.1

**Other information**

registerrier:
Thanks @dguevel . I confirm the bug is also present in dev.

PR #4066, opened 2022-09-07
Solve circular import in gammapy.visualization

registerrier:
**Description**

This pull request solves issue #4065 . 
Should we create tests for global module imports?

**Dear reviewer**

### DIFF
--- a/gammapy/maps/region/geom.py
+++ b/gammapy/maps/region/geom.py
@@ -26,6 +26,7 @@ from gammapy.utils.regions import (
     compound_region_to_regions,
     regions_to_compound_region,
 )
+from gammapy.visualization.utils import ARTIST_TO_LINE_PROPERTIES
 from ..axes import MapAxes
 from ..core import Map, MapCoord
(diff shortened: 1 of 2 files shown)


adonath (changes requested):
Thanks @registerrier! I think we should resolve the error the other way around:
- All modules should be able to import from `gammapy.visualization`
- Use delayed imports in `gammapy.visualization`

What do you think?

gammapy/maps/region/geom.py
```diff
@@ -794,6 +793,7 @@ def plot_region(self, ax=None, kwargs_point=None, path_effect=None, **kwargs):
         ax : `~astropy.visualization.WCSAxes`
             Axes to plot on.
         """
+        from gammapy.visualization.utils import ARTIST_TO_LINE_PROPERTIES
```
adonath, line 796: I thunk this can remain a non-delayed import...

registerrier: Done

adonath (approved):

PR #4070, opened 2022-09-09
Energy evaluation of temporal model

AtreyeeS:
This is a cleaner implementation of the PR started in #3838 
The following steps are to be implemented

- [x] Enable the energy evaluation 
- [x] Add a test
- [x] Add a numerical integration for temporal models with energy 
- [ ] Add an example in the notebook
- [x] Enable plotting for such models
- [ ] Add support for Template model
- [ ] Implement event sampling

### DIFF
--- a/gammapy/modeling/models/cube.py
+++ b/gammapy/modeling/models/cube.py
@@ -12,7 +12,12 @@ from gammapy.utils.fits import LazyFitsData
 from gammapy.utils.scripts import make_name, make_path
 from .core import Model, ModelBase, Models
 from .spatial import ConstantSpatialModel, SpatialModel
-from .spectral import PowerLawNormSpectralModel, SpectralModel, TemplateSpectralModel
+from .spectral import (
+    PowerLawNormSpectralModel,
+    SpectralModel,
+    TemplateSpectralModel,
+    PiecewiseNormSpectralModel,
+)
 from .temporal import TemporalModel
 
 __all__ = [
@@ -365,12 +370,18 @@ class SkyModel(ModelBase):
             Predicted flux map
         """
         energy = geom.axes["energy_true"].edges
-        value = np.ones(geom.data_shape)
+        value = u.Quantity(np.ones(geom.data_shape))
+        spectral_model = self.spectral_model
         if self.temporal_model:
             integral = self.temporal_model.integral(
                 gti.time_start, gti.time_stop, energy=energy
             )
-            value = value * np.sum(integral)
+            if self.temporal_model.is_energy_dependent:
+                norms = np.sum(integral, axis=-1).reshape((-1, 1, 1))
+                norm_model = PiecewiseNormSpectralModel(energy=energy, norms=norms)
+                spectral_model = spectral_model * norm_model
+            else:
+                value = value * np.sum(integral)
 
         if self.spatial_model:
             value = (
@@ -380,7 +391,7 @@ class SkyModel(ModelBase):
                 ).quantity
             )
 
-        value = value * self.spectral_model.integral(
+        value = value * spectral_model.integral(
             energy[:-1],
             energy[1:],
         ).reshape((-1, 1, 1))
--- a/gammapy/modeling/models/cube.py
+++ b/gammapy/modeling/models/cube.py
@@ -366,10 +366,12 @@ class SkyModel(ModelBase):
             Predicted flux map
         """
         energy = geom.axes["energy_true"].edges
-        value = self.spectral_model.integral(
-            energy[:-1],
-            energy[1:],
-        ).reshape((-1, 1, 1))
+        value = np.ones(geom.data_shape)
+        if self.temporal_model:
+            integral = self.temporal_model.integral(
+                gti.time_start, gti.time_stop, energy=energy
+            )
+            value = value * np.sum(integral)
 
         if self.spatial_model:
             value = (
@@ -379,11 +381,10 @@ class SkyModel(ModelBase):
                 ).quantity
             )
 
-        if self.temporal_model:
-            integral = self.temporal_model.integral(
-                gti.time_start, gti.time_stop, energy=energy
-            )
-            value = value * np.sum(integral)
+        value = value * self.spectral_model.integral(
+            energy[:-1],
+            energy[1:],
+        ).reshape((-1, 1, 1))
 
         return Map.from_geom(geom=geom, data=value.value, unit=value.unit)
 
--- a/gammapy/modeling/models/cube.py
+++ b/gammapy/modeling/models/cube.py
@@ -337,7 +337,7 @@ class SkyModel(ModelBase):
 
         if self.temporal_model:
             integral = self.temporal_model.integral(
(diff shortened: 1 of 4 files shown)


AtreyeeS:
@registerrier @adonath I think this PR is ready for review now.
The numerical integration implemented in the test is not nice, but that should be resolved in the next PR with where the integration is performed in the base class

gammapy/modeling/models/temporal.py
```diff
@@ -166,7 +184,7 @@ def sample_time(self, n_events, t_min, t_max, t_delta="1 s", random_state=0):
 
         return t_min + time
 
-    def integral(self, t_min, t_max, oversampling_factor=100, **kwargs):
+    def integral(self, t_min, t_max, oversampling_factor=100, energy=None, **kwargs):
```
adonath, line 187: In general the definition of only the energy center here is not sufficient. We would need `energy_min` and `energy_max` here as well and the perform the energy integration. However for the purpose of the feature is probably fine for now...

gammapy/modeling/models/temporal.py
```diff
@@ -295,7 +317,7 @@ def evaluate(time, t0, t_ref):
         """Evaluate at given times"""
         return np.exp(-(time - t_ref) / t0)
 
-    def integral(self, t_min, t_max):
+    def integral(self, t_min, t_max, energy=None):
```
adonath, line 320: Please make sure to adapt the docstrings as well..

gammapy/modeling/models/temporal.py
```diff
@@ -91,13 +104,18 @@ def plot(self, time_range, ax=None, **kwargs):
         time_axis = TimeMapAxis.from_time_bounds(
             time_min=time_min, time_max=time_max, nbin=100
         )
-
         m = RegionNDMap.create(region=None, axes=[time_axis])
+        if energy:
+            if self.is_energy_dependent is False:
```
adonath, line 109:
```suggestion
            if not self.is_energy_dependent:
```

adonath (commented):
Thanks @AtreyeeS, I have left some minor comments for now. I think we can merge this PR rather fast and then proceed with the plan discussed in the meeting today:

- Introduce a `t_ref` parameter to the `LightCurveTemplateTemporalModel`
- Refactor `LightCurveTemplateTemporalModel` to use a `RegionNDMap` internally (without region definiton)
- Make sure the serialization `.read()` and `.write()` still works as before
- Let me and other devs know about additional required changes to `RegionNDMap`, such that we can maybe split the work and support you

gammapy/modeling/models/temporal.py
AtreyeeS: That would again change the logic of the `MapEvaluator`, no? This implementation simply calculates the normalisation introduced at the specific (true) energy grid, which is then independently integrated over.

AtreyeeS:
Thanks @adonath ! I addressed your comments, and opened #4072 for the `RegionNdMaps`.
I am working on the template model now.

gammapy/modeling/models/temporal.py
```diff
@@ -295,7 +321,7 @@ def evaluate(time, t0, t_ref):
         """Evaluate at given times"""
         return np.exp(-(time - t_ref) / t0)
 
-    def integral(self, t_min, t_max):
+    def integral(self, t_min, t_max, energy=None):
```
registerrier, line 324: Should the `energy` argument be exposed when models are not energy dependent.

gammapy/modeling/models/tests/test_temporal.py
```diff
@@ -319,3 +323,92 @@ def test_plot_constant_model():
...
+
+    def integral(self, t_min, t_max, energy):
+        pars = self.parameters
+        t_ref = Time(pars["t_ref"].quantity, format="mjd")
+        beta = pars["beta"].quantity
+        E0 = pars["E0"].quantity
+        integral = []
+        for t1, t2 in zip(t_min, t_max):
+            integral_ene = []
+            for ene in energy:
```
registerrier, line 358: Is the integral method meaningful here?
Should it return the integral over time of the spectral norms or the double integral over time and energy as defined here? 
This does not sound obvious to me.

gammapy/modeling/models/cube.py
```diff
@@ -332,7 +336,9 @@ def evaluate_geom(self, geom, gti=None):
             value = value * self.spatial_model.evaluate_geom(geom)
 
         if self.temporal_model:
-            integral = self.temporal_model.integral(gti.time_start, gti.time_stop)
+            integral = self.temporal_model.integral(
```
registerrier, line 339: here you want to have the integral over time at given true energies. Not the integral over time and true energy, right?

gammapy/modeling/models/cube.py
```diff
@@ -373,7 +379,9 @@ def integrate_geom(self, geom, gti=None, oversampling_factor=None):
             )
 
         if self.temporal_model:
-            integral = self.temporal_model.integral(gti.time_start, gti.time_stop)
+            integral = self.temporal_model.integral(
```
registerrier, line 382: here I would rather compute the integral over time and then compute the integral of the spectral model.

registerrier (commented):
Thanks @AtreyeeS . 

I think the exact behavior of `integral` with time and anergy is unclear. I think it should not compute the double integral.

gammapy/modeling/models/tests/test_temporal.py
AtreyeeS: In the case of a normal temporal model, an array of normalisations is returned, ie, one integral for each time interval .

In this case, I am computing the integral between t1 and t2 for each energy. A double integral is not performed.
What do you think @registerrier ?

The implementation is indeed ugly, but this for this I am preparing a PR.

AtreyeeS:
I have implemented `TemporalModel.integral()` in this PR itself.
The integral is computed between the GTIs at the given energy.
I think this makes the logic clearer. Can you do another review @adonath @registerrier ?

I imagine the reasonable way to use these models would be with a `ConstantSpectralModel` to allow a spectral normalisation, while avoiding factorising the integral.

AtreyeeS:
@registerrier Indeed! I was not very happy about the same either. I have updated the PR to use a `PiecewiseNormSpectralModel` now. The test values did not change.
The `PiecewiseNormSpectralModel` is still calculated at the true energies. For analytic models, I dont know what other energy sampling we can use. For a template model, maybe it can be an option to use the sampling of the model itself?

adonath:
In the dev call this morning we proposed the following:
- @AtreyeeS will first open another PR adding the modified `TemplateTemporalModel`, with internal support for `RegionNDMap`
- Then @registerrier and @fabiopintore  will continue to work on implementing the support for event sampling
- At the same time @AtreyeeS will add support for YAML serialisation of the `TemplateTemporalModel`

For event sampling the actual evaluation of the temporal model at given energies is not needed, so we will put this PR on hold for now, because event sampling as the priority.

AtreyeeS:
Continuing in #4272.

Issue #4056, opened 2022-08-16
Can't write DarkMatter Model to file


StFroese:
**Gammapy version**
System:

	python_executable      : /scratch/sfroese/envs/phd/bin/python 
	python_version         : 3.9.13     
	machine                : x86_64     
	system                 : Linux      

Gammapy package:

	version                : 0.20.1     
	path                   : /scratch/sfroese/envs/phd/lib/python3.9/site-packages/gammapy 

INFO:numexpr.utils:Note: NumExpr detected 48 cores but "NUMEXPR_MAX_THREADS" not set, so enforcing safe limit of 8.
INFO:numexpr.utils:NumExpr defaulting to 8 threads.

Other packages:

	numpy                  : 1.22.3     
	scipy                  : 1.9.0      
	astropy                : 5.0.4      
	regions                : 0.6        
	click                  : 8.1.3      
	yaml                   : 6.0        
	IPython                : 8.3.0      
	jupyterlab             : not installed 
	matplotlib             : 3.5.2      
	pandas                 : 1.4.2      
	healpy                 : 1.16.1     
	iminuit                : 2.15.2     
	sherpa                 : not installed 
	naima                  : not installed 
	emcee                  : not installed 
	corner                 : not installed 

Gammapy environment variables:

	GAMMAPY_DATA           : /net/big-tank/POOL/users/sfroese/gammapy-data/ 

**Bug description**
I'm not able to write a `DarkMatterAnnihilationSpectralModel` to a file.
```python
AttributeError: 'DarkMatterAnnihilationSpectralModel' object has no attribute 'tag'
```

**Expected behavior**
See above.

**To Reproduce**
```python
from gammapy.astro.darkmatter import (
    profiles,
    JFactory,
    DarkMatterAnnihilationSpectralModel,
)
from astropy.coordinates import SkyCoord
from gammapy.maps import WcsGeom
from gammapy.maps import WcsNDMap
from gammapy.modeling.models import (
    Models,
    SkyModel,
    TemplateSpatialModel
)
import astropy.units as u
```
```python
profile = profiles.NFWProfile()

position = SkyCoord(0.0, 0.0, frame="galactic", unit="deg")
geom = WcsGeom.create(binsz=0.05, skydir=position, width=3.0, frame="galactic")


jfactory = JFactory(
    geom=geom, profile=profile, distance=profiles.DMProfile.DISTANCE_GC
)
jfact = jfactory.compute_differential_jfactor()

[...]

PR #4071, opened 2022-09-09
Add tag on DM Annihilation spectral model

registerrier:
**Description**

This pull request solves #4056  by simply adding a `tag` on `DarkMatterAnnihilationSpectralModel` with short name `dm-annihilation`.
Note that DM spatial models are not usable in `SkyModel` as they are not regular `SpatialModel`.

**Dear reviewer**

### DIFF
--- a/gammapy/astro/darkmatter/spectra.py
+++ b/gammapy/astro/darkmatter/spectra.py
@@ -177,6 +177,7 @@ class DarkMatterAnnihilationSpectralModel(SpectralModel):
         interp="log",
         is_norm=True,
     )
+    tag = ["DarkMatterAnnihilationSpectralModel", "dm-annihilation"]
 
     def __init__(self, mass, channel, scale=scale.quantity, jfactor=1, z=0, k=2):
         self.k = k
--- a/gammapy/astro/darkmatter/spectra.py
+++ b/gammapy/astro/darkmatter/spectra.py
@@ -4,7 +4,7 @@ import numpy as np
 import astropy.units as u
 from astropy.table import Table
 from gammapy.modeling import Parameter
-from gammapy.modeling.models import SpectralModel, TemplateSpectralModel
+from gammapy.modeling.models import SpectralModel, TemplateSpectralModel, SPECTRAL_MODEL_REGISTRY
 from gammapy.utils.interpolation import LogScale
 from gammapy.utils.scripts import make_path
 
@@ -182,9 +182,9 @@ class DarkMatterAnnihilationSpectralModel(SpectralModel):
     def __init__(self, mass, channel, scale=scale.quantity, jfactor=1, z=0, k=2):
         self.k = k
         self.z = z
-        self.mass = mass
+        self.mass = u.Quantity(mass)
         self.channel = channel
-        self.jfactor = jfactor
+        self.jfactor = u.Quantity(jfactor)
         self.primary_flux = PrimaryFlux(mass, channel=self.channel).table_model
         super().__init__(scale=scale)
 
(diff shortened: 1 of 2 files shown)


adonath:
Thanks @registerrier! My main question is whether adding the tag really solves the problem? The `DarkMatterAnnihilationSpectralModel` as much more associated information , which needs to be serialized with a custom `.to_dict()` methods. Including `k, z, mass, channel`...

registerrier:
> Thanks @registerrier! My main question is whether adding the tag really solves the problem? The `DarkMatterAnnihilationSpectralModel` as much more associated information , which needs to be serialized with a custom `.to_dict()` methods. Including `k, z, mass, channel`...

This is a good point. Technically, some of these quantities could be `Parameter` objects. But not e.g. `channel`.  How to serialize non `Parameter` attributes is not clear. 
Maybe this is a similar case to `NaimaSpectralModel` which has no yaml serialization for now.

adonath:
@registerrier Yes, maybe let's just raise a meaningful error for now?

registerrier:
I have included additional attributes in the out dictionary. It actually works and allows exporting DMmodels to yaml.

Obviously we will need to look at what should be a `Parameter` and what is a simple model attribute. But that could be left for later as the DM models will have to be better included in the `Model` framework (or they will have to be removed if those models are found to be useless).

Opinions @adonath ?

adonath:
Thanks @registerrier, great to see it works with limited effort. I'll make a more thorough review later...

gammapy/astro/darkmatter/spectra.py
```diff
@@ -200,3 +201,27 @@ def evaluate(self, energy, scale):
...
+        data = data["spectral"]
+
+        channel = data["channel"]
+        mass = u.Quantity(data["mass"])
+        jfactor = u.Quantity(data["jfactor"])
+        z = data["z"]
+        k = data["k"]
+
+        scale = [p["value"] for p in data["parameters"] if p["name"] == "scale"][0]
+        return cls(mass, channel, scale=scale, jfactor=jfactor, z=z, k=k)
```
adonath, line 225:
```suggestion
    def from_dict(cls, data):
        """Create spectral model from dict
        
        Parameters
        ----------
        data : dict
            DIct with model data
        
        Returns
        -------
        model : `DarkMatterAnnihilationSpectralModel`
            Dark matter annihilation spectral model
        """
        data = data["spectral"]
        scale = [p["value"] for p in data.pop("parameters") if p["name"] == "scale"][0]
        return cls(scale=scale, **data)
```

gammapy/astro/darkmatter/spectra.py
```diff
@@ -177,13 +177,14 @@ class DarkMatterAnnihilationSpectralModel(SpectralModel):
         interp="log",
         is_norm=True,
     )
+    tag = ["DarkMatterAnnihilationSpectralModel", "dm-annihilation"]
 
     def __init__(self, mass, channel, scale=scale.quantity, jfactor=1, z=0, k=2):
         self.k = k
         self.z = z
-        self.mass = mass
+        self.mass = u.Quantity(mass)
```
adonath, line 185: Maybe add a unit test to establish the quantity here...

adonath (commented):
Thanks @registerrier, I have left one minor comment and one inline suggestion...

adonath:
What is the status here @registerrier ?

gammapy/astro/darkmatter/spectra.py
```diff
@@ -200,3 +201,34 @@ def evaluate(self, energy, scale):
...
+        data : dict
+            Dict with model data
+
+        Returns
+        -------
+        model : `DarkMatterAnnihilationSpectralModel`
+            Dark matter annihilation spectral model
+        """
+        data = data["spectral"]
+        type = data.pop("type")
```
adonath, line 229:
```suggestion
        data.pop("type")
```

adonath (approved):

Issue #4072, opened 2022-09-12
Coordinate support for RegionNDMaps


AtreyeeS:
1. Presently, co-ordinate handling on `RegionNDMap` is broken (eg: `RegionNdMap.interp_by_coord) ` as `MapCoord` compulsorily needs a position argument. 
It would be nice to resolve this for supporting maps where `geom=None`

PR #4073, opened 2022-09-12
Make spatial coordinates optional in RegionNDMap.interp_by_coord()

adonath:
See title. This pull request addresses #4072

### DIFF
--- a/gammapy/maps/region/geom.py
+++ b/gammapy/maps/region/geom.py
@@ -518,7 +518,8 @@ class RegionGeom(Geom):
         if isinstance(coords, dict):
             coords.setdefault("skycoord", self.center_skydir)
         elif isinstance(coords, tuple) and len(coords) == len(self.axes):
-            coords = (0, 0) + coords
+            skydir = self.center_skydir.transform_to(self.frame)
+            coords = (skydir.data.lon, skydir.data.lat) + coords
 
         coords = MapCoord.create(coords, frame=self.frame, axis_names=self.axes.names)
 
--- a/gammapy/maps/region/geom.py
+++ b/gammapy/maps/region/geom.py
@@ -513,6 +513,7 @@ class RegionGeom(Geom):
         return tuple(idxs)
 
     def coord_to_pix(self, coords):
+        # inherited docstring
         coords = MapCoord.create(coords, frame=self.frame, axis_names=self.axes.names)
 
         if self.region is None:
--- a/gammapy/maps/region/geom.py
+++ b/gammapy/maps/region/geom.py
@@ -28,7 +28,8 @@ from gammapy.utils.regions import (
 )
 from gammapy.visualization.utils import ARTIST_TO_LINE_PROPERTIES
 from ..axes import MapAxes
-from ..core import Map, MapCoord
+from ..core import Map
+from ..coord import MapCoord
 from ..geom import Geom, pix_tuple_to_idx
 from ..utils import _check_width
 from ..wcs import WcsGeom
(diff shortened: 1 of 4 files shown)


AtreyeeS (changes requested):
Thanks @adonath ! Maybe you can also fix `get_by_coord`?

adonath:
Thanks @AtreyeeS! I'll include the change to `.get_by_coord()` in this PR...

gammapy/maps/region/ndmap.py
```diff
@@ -395,8 +395,31 @@ def get_by_idx(self, idxs):
...
-        pix = self.geom.coord_to_pix(coords)
+        """Interpolate map values at the given map coordinates.
+
+        Parameters
+        ----------
+        coords : tuple, dict or `~gammapy.maps.MapCoord`
+            Coordinate arrays for each dimension of the map.  Tuple
+            should be ordered as (lon, lat, x_0, ..., x_n) where x_i
+            are coordinates for non-spatial dimensions of the map.
+            "lon" and "lat" are optional and will be take at the center
```
registerrier, line 406: will be taken at

gammapy/maps/region/tests/test_ndmap.py
```diff
@@ -211,6 +211,13 @@ def test_region_nd_map_get(region_map):
     values = region_map.get_by_coord((83.63, 21.51, energies[[0, -1]]))
     assert_allclose(values.squeeze(), [0, 5])
 
+    values = region_map.get_by_coord((energies[[0, -1]],))
+    assert_allclose(values.squeeze(), [0, 5])
+
+    values = region_map.get_by_coord({"energy": energies[[0, -1]]})
+    assert_allclose(values.squeeze(), [0, 5])
```
registerrier, line 218: Does it work with `region=None`? Maybe this should be tested as well?

gammapy/maps/region/tests/test_ndmap.py
```diff
@@ -375,3 +382,32 @@ def test_region_nd_hdulist():
     assert hdulist[1].name == "SKYMAP"
     assert hdulist[2].name == "SKYMAP_BANDS"
     assert hdulist[3].name == "SKYMAP_REGION"
+
+
+def test_region_nd_map_interp_no_region():
```
registerrier, line 387: this answers the  previous comment...

gammapy/maps/region/geom.py
```diff
@@ -513,6 +513,14 @@ def pix_to_idx(self, pix, clip=False):
         return tuple(idxs)
 
     def coord_to_pix(self, coords):
+        # inherited docstring
+
+        if isinstance(coords, dict):
+            coords.setdefault("skycoord", self.center_skydir)
```
registerrier, line 519: what happens here if the `dict` contains already `lat` and `lon` keys?

registerrier (approved):
Thanks @adonath ! This looks good. I have left small inline questions/comments.

gammapy/maps/region/geom.py
adonath: Then `.setdefault()` does nothing...

registerrier: It will add the `skycoord` to the list of keys and it will later be listed as one of the axes.

adonath: Yes, of course! I forgot we treat `skycoord` separately from `lon` and `lat`. It was such a bad idea to support all of those, like `MapCoord`, `dict` (with `skycoord` and `lon` and `lat`) and `tuple`...it's so bad to maintain. We should get rid of it for Gammapy v2.0...

AtreyeeS (approved):
Thanks @adonath ! No further comments from my side

Issue #3764, opened 2022-01-28
Implement MapEvaluator.peek()


adonath:
For debugging and possibly for users as well it would be useful to have a `MapEvaluator.peek()` method. It could plot all of the following quantities:
- PSF kernel
- EDisp kernel
- Exposure map
- Npred map

AtreyeeS:
Currently `MapEvaluator` is not visible in the public API. Maybe making it visible will make the peek more useful.

adonath:
Addressed in #4107.

PR #4074, opened 2022-09-12
Implement PsfKernel.peek() and MapEvaluator.peek()

luca-giunti:
**Description**

This pull request addresses #3764 , by adding a `MapEvaluator.peek()` method. For consistency (and to simplify the code) I also added a plot and a peek method for the `PsfKernel`.

Example plots:

`PsfKernel.peek()`
image

`MapEvaluator.peek()`
image

### DIFF
--- a/docs/tutorials/api/datasets.ipynb
+++ b/docs/tutorials/api/datasets.ipynb
@@ -1047,11 +1047,48 @@
     ")\n",
     "print(datasets_sliced.energy_ranges)"
    ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "049be356-8798-437f-b9f6-9755b4d82e7d",
+   "metadata": {},
+   "source": [
+    "## MapEvaluator"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "906b6b6a-d017-4cff-8844-ad2eda09fca0",
+   "metadata": {},
+   "source": [
+    "The `MapEvaluator` is responsible for evaluating a sky model on a map and returning a map of the predicted counts, by convolving the model with IRFs. This is an essential step for the fitting procedure. \n",
+    "\n",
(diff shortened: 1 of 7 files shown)


registerrier (commented):
Thanks @luca-giunti . Maybe one general comment first. Do we want to expose `MapEvaluator` as user level API? 
I assume the datasets tutorial is a valid place to give an example, but should we expose the full API?

adonath:
I agree with @registerrier, the `MapEvaluator` can remain part of the internal API. The plotting is mostly useful for developers for debugging. That's also why I opened the issue in first place. The `MapEvaluator` is one of the most complex pieces of code in Gammapy and I found myself plotting its data members by hand very often.

registerrier (commented):
One general question, what happens when the dataset is a `SpectrumDataset`

Looking at the content of the plot, I would a minima plot the 2D integrated counts image and the 1D integrated counts spectrum. 
The IRFs plots are interesting too. The energy integrated kernel is less obvious to me as its meaning is not clear (no averaging over energies).

luca-giunti:
> One general question, what happens when the dataset is a `SpectrumDataset`
> 
> Looking at the content of the plot, I would a minima plot the 2D integrated counts image and the 1D integrated counts spectrum. The IRFs plots are interesting too. 

@registerrier for a spectrum dataset only the PSF plots are removed. Here is an example (I may try to remove the blank line, even if this is not very important being a debug plot)
image

Is the 2D integrated counts image still available in a spectrum dataset?

>The energy integrated kernel is less obvious to me as its meaning is not clear (no averaging over energies).

I am open to better suggestions :)

adonath:
@luca-giunti I just realized the change of the tutorial setup created a merge conflict here. I guess it is best resolved by discarding the changes to the tutorial in this PR, like so:
```
git reset HEAD~
git checkout -- .
git fetch upstream
git rebase upstream/master
git push -f origin Implement-MapEvaluator-peek
```
Depending on how your upstream and origin remotes are called...

Issue #4006, opened 2022-06-24
HpxGeom constructor should fail if nside is not a power of two.


registerrier:
Healpix geometry requires that the number of sides is a power of two. Currently the keyword `nside` of `HpxGeom.__init__` accepts any value. But further manipulation will fail, e.g.:
```
from gammapy.maps import HpxGeom
geom=HpxGeom.create(7)
geom.get_coord()
```
fails with `ValueError: [10] is not a valid nside parameter (must be a power of 2, less than 2**30)`

The `nside` validation should take place on `__init__`.

registerrier:
Solved by #4075 . Closing.

PR #4075, opened 2022-09-13
Validate nside parameter for HpxGeom 

luca-giunti:
**Description**

This pull request solves #4006

### DIFF
--- a/gammapy/maps/hpx/geom.py
+++ b/gammapy/maps/hpx/geom.py
@@ -6,6 +6,7 @@ from astropy import units as u
 from astropy.coordinates import SkyCoord
 from astropy.io import fits
 from astropy.units import Quantity
+from healpy.pixelfunc import check_nside
 from gammapy.utils.array import is_power2
 from ..axes import MapAxes
 from ..coord import MapCoord, skycoord_to_lonlat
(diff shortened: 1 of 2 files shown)


luca-giunti:
I just realized that the proposed solution creates problems because it relies on a function contained in `healpy`, which is not a mandatory dependency: 
`from healpy.pixelfunc import check_nside`

Should I copy-paste that function in a `HpxGeom._check_nside()` mehod? It's a pity because it, looks like this, not trivial
```
    if hasattr(nside, "__len__"):
        if not isinstance(nside, np.ndarray):
            nside = np.asarray(nside)
        is_nside_ok = (
            (nside == nside.astype(int)) & (nside > 0) & (nside <= max_nside)
        )
        if nest:
            is_nside_ok &= (nside.astype(int) & (nside.astype(int) - 1)) == 0
    else:
        is_nside_ok = nside == int(nside) and 0 < nside <= max_nside
        if nest:
            is_nside_ok = is_nside_ok and (int(nside) & (int(nside) - 1)) == 0
    return is_nside_ok
```

adonath:
@luca-giunti Just use a delayed import, like:

```python3

def __init__(self, ..):
    from healpy.pixelfunc import check_nside
```

Healpy is required anyway when working with HEALPix maps...

registerrier:
You might need the `@require_dependency` decorator for the test though

luca-giunti:
> You might need the `@require_dependency` decorator for the test though

@registerrier is that really needed? There is an overall `pytest.importorskip("healpy")` in the test file

registerrier (approved):

adonath (approved):
Thanks @luca-giunti!, looks good to me as well!

PR #4076, opened 2022-09-13
Allow for DRIFT mode observations in the MapDatasetMaker

LauraOlivera:
**Description**
Something changed between 0.19 and 0.20 leading to an error in the `MapDatasetMaker` if the observation table has no RA/DEC pointing information. Before, the `observation.pointing_radec` would have `nan` as values, but now it throws an error. Which occurs when constructing the meta table during the data reduction in 0.20.1

This pull request adds the `OBS_MODE` keyword to the dataset meta table, and consequently, the pointing info required is the one defined by the observation mode.

### DIFF
--- a/gammapy/data/pointing.py
+++ b/gammapy/data/pointing.py
@@ -131,8 +131,6 @@ class FixedPointingInfo:
         obs_mode = self.meta.get("OBS_MODE")
         if obs_mode is None:
             return PointingMode.POINTING
-        if np.size(obs_mode)>1 and "POINTING" in [i.upper() for i in obs_mode]:
-            return PointingMode.POINTING
         return PointingMode.from_gadf_string(obs_mode)
 
     @lazyproperty
--- a/gammapy/data/pointing.py
+++ b/gammapy/data/pointing.py
@@ -131,6 +131,8 @@ class FixedPointingInfo:
         obs_mode = self.meta.get("OBS_MODE")
         if obs_mode is None:
             return PointingMode.POINTING
+        if np.size(obs_mode)>1 and "POINTING" in [i.upper() for i in obs_mode]:
+            return PointingMode.POINTING
         return PointingMode.from_gadf_string(obs_mode)
 
     @lazyproperty
--- a/gammapy/makers/map.py
+++ b/gammapy/makers/map.py
@@ -328,9 +328,13 @@ class MapDatasetMaker(Maker):
         meta_table = Table()
         meta_table["TELESCOP"] = [observation.aeff.meta.get("TELESCOP", "Unknown")]
         meta_table["OBS_ID"] = [observation.obs_id]
-        meta_table["RA_PNT"] = [observation.pointing_radec.icrs.ra.deg] * u.deg
-        meta_table["DEC_PNT"] = [observation.pointing_radec.icrs.dec.deg] * u.deg
-
+        meta_table["OBS_MODE"] = [str(observation.fixed_pointing_info.mode).split('.')[-1]]
+        if meta_table["OBS_MODE"] == "POINTING":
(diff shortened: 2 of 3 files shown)


gammapy/data/pointing.py
```diff
@@ -131,6 +131,8 @@ def mode(self):
         obs_mode = self.meta.get("OBS_MODE")
         if obs_mode is None:
             return PointingMode.POINTING
+        if np.size(obs_mode)>1 and "POINTING" in [i.upper() for i in obs_mode]:
```
adonath, line 134: Hm, the need for this statement is not clear to me. In any case one should avoid the hard-coded "POINTING" string here. This is the whole point of the existence of the `PointingMode` enum object. I'll comment in more detail further down...

gammapy/makers/map.py
```diff
@@ -328,9 +328,13 @@ def make_meta_table(observation):
         meta_table = Table()
         meta_table["TELESCOP"] = [observation.aeff.meta.get("TELESCOP", "Unknown")]
         meta_table["OBS_ID"] = [observation.obs_id]
-        meta_table["RA_PNT"] = [observation.pointing_radec.icrs.ra.deg] * u.deg
-        meta_table["DEC_PNT"] = [observation.pointing_radec.icrs.dec.deg] * u.deg
-
+        meta_table["OBS_MODE"] = [str(observation.fixed_pointing_info.mode).split('.')[-1]]
```
adonath, line 331: The logic here should be along the lines of:

```python

if observation.fixed_pointing_info.mode == PointingMode.POINTING:
   ...
elif observation.fixed_pointing_info.mode == PointingMode.DRIFTING:
   ...
```

gammapy/data/pointing.py
adonath: Maybe this should go in `PointingMode.from_gadf_string` instead?

adonath (approved):
Thanks @LauraOlivera! I have left some inline comments...

gammapy/data/pointing.py
LauraOlivera: well I added that because the data used in the `spectral_analysis_rad_max` tutorial have `OBS_MODE = ["POINTING", "WOBBLE"]`, which gives an error because `PointingMode.from_gadf_string` expects a single string and not a list...

LauraOlivera: but indeed I can move that logic inside `PointingMode.from_gadf_string`

LauraOlivera: it would also work to take the first of the entries of the list... I just thought that checking if `POINTING` was in it would be best....

adonath: > well I added that because the data used in the spectral_analysis_rad_max tutorial have OBS_MODE = ["POINTING", "WOBBLE"], which gives an error because PointingMode.from_gadf_string expects a single string and not a list...

Is having a list of pointing modes GADF compliant? Do you know?

LauraOlivera: technically no - GADF specifies it should be a string

LauraOlivera: but `OBS_MODE` is also not mandatory in 0.2... so it is a bit of a gray area

adonath: OK, maybe @cosimoNigro can shortly comment. I think he produced the rad max files...

adonath:
After https://github.com/gammapy/gammapy-data/pull/27 has been merged I restarted the CI builds and the fail related to the pointing mode should disappear...

adonath (approved):
Thanks @LauraOlivera, remaining test fails are unrelated. Will go ahead and merge now...

PR #4077, opened 2022-09-15
Review of CITATION.cff for the LTS release

bkhelifi:
**Description**
In the context of preparation of our first LTS release, **the author list should be reviewed by each of the Gammapy contributors**. Emails will also be sent as reminder.

**Dear reviewer**
Followingthe PIG 24 #3970, the authorship policy for the v1.0 release is special, as the DCO mechanism (see #4035) was not installed since the beginning of the project.

So, each contributor should comment this PR by:
1. if she/her wishes to "OptOut", removing her/his associated data
2. checking her/his name (first and family) and the information about email address
4. checking/adding/removing her/his ORCID (we do recommend to have one and to update your personal information, as Gammapy follows the EU Open Science recommendations)
5. checking/adding/removing her/his affiliation and its associated city, accordingly to your employer agreement and rules _at the time_ of your contribution or latest contribution.

In case of no reaction or answer, the current information will be kept.

PR #3970, opened 2022-05-30
PIG 24: Authorship Policy

bkhelifi:
This pull request is about the Gammapy policy about the authorship rules for all kind of products (code, paper, conferences).This is proposal that has been discussed a bit among the Lead Developers and Project Managers.

A general discussion with the Gammapy community is required up to a consensus. Then, this PIG will be submitted to the CC for validation.

From that, the author list for the V1.0 paper will be updated.

PR #4078, opened 2022-09-15
Update the codemeta file for the LTS release

bkhelifi:
**Description**
In the context of preparation of our first LTS release, the ``codemeta.json`` file should be updated accordingly to our new authorship policy (#3970).

This pull request contains an update of the metadata associated to the projet and an update of the author list.

**Dear reviewer**
The review should be made in two steps:
1. review of the general metadata
2. once the PR #4077 is accepted and merged, we will update the metadata associated to the authors


codemeta.json
```diff
@@ -1,616 +1,809 @@
...
+    "runtimePlatform": [
+        "C",
+        "Cython",
+        "Python 3",
+        "Python 3.8",
+        "Python 3.9",
+        "Python 3.10",
+        "Python Implementation CPython"
+    ],
+    "softwareRequirements": [
```
registerrier, line 75: The software requirements listed here correspond to the dev environment. Many are not needed.
Also should we put minimum requirements as per setup.cfg or the maximal list that will go in the `environment-v1.0.yaml`?

registerrier (commented):
Thanks @bkhelifi .

The dependencies listed here are for the dev environment. What should be provide here? I suppose only the minimal required dependencies, no? So this should be synced with what in the setup.cfg. Right?

codemeta.json
adonath: I think this entry should be auto generated from the `setup.cfg` and just list the minimal requirements. The information in the `setup.cfg` is much more detailed. We shouldn't even try to duplicate it here min any way...

registerrier:
Superseded by #4202

PR #4080, opened 2022-09-15
Use sphinx gallery for tutorials

### LINKED ISSUE #3985: Left pane inconsistent for the tutorials
This is a reminder issue of what we discussed in the dev call today.  In the left pane here https://docs.gammapy.org/0.20/tutorials/analysis/1D/spectral_analysis.html we see the subsections of the user guide, with some in bold. This happens for some tutorials (eg: 3D analysis), but not the others (eg: multi instrument). Any idea why?

adonath:
**Description**

This pull request changes the tutorial handling to rely on https://sphinx-gallery.github.io/stable/index.html as well. This has a number of advantages:

- Fixes #3985 
- Fixes the current tutorial layout issue:  https://docs.gammapy.org/dev/tutorials/index.html
- It avoids the need for a specific `make docs-all` command and includes all examples in the docs build
- It changes from `.ipynb` files to normal `.py` files. Those can be easier maintained and edited with code editors. Some developers might prefer editing in a jupyter notebook, which should be possible using https://jupytext.readthedocs.io/en/latest/
- In a second PR we can remove the custom notebook tooling, which avoids the need to maintain it for v1.0.
- I think the notebook download test can currently not succeed, because there are no updated docs yet.

ToDo:
- [ ] adapt the choice of the thumbnail images to the old ones
- [ ] Activate Binder integration, see https://sphinx-gallery.github.io/stable/configuration.html#generate-binder-links-for-gallery-notebooks-experimental

The preview of the docs are here: https://adonath.github.io/gammapy-docs-preview/

Here is an example for the automatically generated API mini-gallery https://adonath.github.io/gammapy-docs-preview/user-guide/dl3.html#examples-using-gammapy-data-datastore


docs/development/intro.rst
```diff
@@ -171,7 +171,7 @@ To *activate* your development version (branch) of Gammapy in your environment:
 
 .. code-block:: bash
 
-    pip install -e .
+    python -m pip install -e .
```
registerrier, line 174: Why is this changed?

registerrier (commented):
Thanks @adonath ! That's a major effort here.

Just to clarify: what is the development process introduced by this PR for tutorials? Directly write a python script and convert it to a notebook afterwards?

docs/development/intro.rst
adonath: It's just in general the recommended way to install into a specific environment. `pip` can in general live in a different environment than the `python` command. See e.g. http://jakevdp.github.io/blog/2017/12/05/installing-python-packages-from-jupyter/#How-to-use-Pip-from-the-Jupyter-Notebook

adonath:
@registerrier There are some options for the development process. In general tutorials are now stored as plain Python files (using the gallery format with embedded RST as comments). While it is less convenient for inert active editing, I'm convinced it is the much better format for maintenance. It will lead to much more meaningful diffs, they can be "blacked" and "isorted" just as normal Python files.

For contributors there are now multiple options:
- Just edit plain Python files in `examples` using a normal editor, this is a very good and more efficient workflow for small edits and maintenance
- For creating new notebooks and interactive editing one can use https://jupytext.readthedocs.io/en/latest/ and store in the `sphinx-gallery` format
- Or just store in `.ipynb` format and use a tool like https://gist.github.com/chsasank/7218ca16f8d022e02a9c0deb94a310fe (I have used a modified version in this PR, we could maybe include it in `dev/`)
- All notebooks are then automatically generated by `sphinx-gallery` from the `.py` files, including zip archives for download of all and individual notebooks, measuring runtimes memory usage and API mini galleries.

adonath:
I will now proceed and merge this in order to continue the clean up...

PR #4084, opened 2022-09-16
Adapt TemplateTemporalModel to use a RegionNDMap internally

AtreyeeS:
This PR is a work in progress.

### DIFF
--- a/gammapy/datasets/tests/test_simulate.py
+++ b/gammapy/datasets/tests/test_simulate.py
@@ -45,7 +45,7 @@ def models():
     table["NORM"] = norm / norm.max()
     t_ref = Time("2000-01-01")
     table.meta = dict(MJDREFI=t_ref.mjd, MJDREFF=0, TIMEUNIT="s")
-    temporal_model = LightCurveTemplateTemporalModel(table)
+    temporal_model = LightCurveTemplateTemporalModel.from_table(table)
 
     model = SkyModel(
         spatial_model=spatial_model,
@@ -129,17 +129,17 @@ def test_mde_sample_sources(dataset, models):
     sampler = MapDatasetEventSampler(random_state=0)
     events = sampler.sample_sources(dataset=dataset)
 
-    assert len(events.table["ENERGY_TRUE"]) == 88
-    assert_allclose(events.table["ENERGY_TRUE"][0], 2.751205, rtol=1e-5)
+    assert len(events.table["ENERGY_TRUE"]) == 90
+    assert_allclose(events.table["ENERGY_TRUE"][0], 2.383778805, rtol=1e-5)
     assert events.table["ENERGY_TRUE"].unit == "TeV"
 
-    assert_allclose(events.table["RA_TRUE"][0], 266.559566, rtol=1e-5)
+    assert_allclose(events.table["RA_TRUE"][0], 266.56408893, rtol=1e-5)
     assert events.table["RA_TRUE"].unit == "deg"
 
-    assert_allclose(events.table["DEC_TRUE"][0], -28.742429, rtol=1e-5)
+    assert_allclose(events.table["DEC_TRUE"][0], -28.748145, rtol=1e-5)
     assert events.table["DEC_TRUE"].unit == "deg"
 
-    assert_allclose(events.table["TIME"][0], 123.424996, rtol=1e-5)
+    assert_allclose(events.table["TIME"][0], 119.7494479, rtol=1e-5)
     assert events.table["TIME"].unit == "s"
 
     assert_allclose(events.table["MC_ID"][0], 1, rtol=1e-5)
@@ -201,14 +201,14 @@ def test_mde_sample_psf(dataset, models):
     events = sampler.sample_sources(dataset=dataset)
     events = sampler.sample_psf(dataset.psf, events)
 
-    assert len(events.table) == 88
-    assert_allclose(events.table["ENERGY_TRUE"][0], 2.751205, rtol=1e-5)
+    assert len(events.table) == 90
+    assert_allclose(events.table["ENERGY_TRUE"][0], 2.38377880, rtol=1e-5)
     assert events.table["ENERGY_TRUE"].unit == "TeV"
 
-    assert_allclose(events.table["RA"][0], 266.556053, rtol=1e-5)
+    assert_allclose(events.table["RA"][0], 266.542912, rtol=1e-5)
     assert events.table["RA"].unit == "deg"
 
-    assert_allclose(events.table["DEC"][0], -28.746459, rtol=1e-5)
+    assert_allclose(events.table["DEC"][0], -28.78829, rtol=1e-5)
     assert events.table["DEC"].unit == "deg"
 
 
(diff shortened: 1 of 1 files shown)


gammapy/modeling/models/temporal.py
```diff
@@ -472,79 +471,93 @@ def __str__(self):
         )
 
     @classmethod
-    def read(cls, path):
-        """Read lightcurve model table from FITS file.
+    def from_table(cls, table):
+        """Create a Template model from an astropy table"""
```
adonath, line 475: Extend the docstring here for a `Parameters` section...

gammapy/modeling/models/temporal.py
```diff
@@ -472,79 +471,93 @@ def __str__(self):
...
+        columns = [_.upper() for _ in table.colnames]
+        t_ref = table.meta["MJDREFI"]
```
adonath, line 477: This only copies over the integer part of the reference time. Please rely on https://github.com/gammapy/gammapy/blob/master/gammapy/utils/time.py#L18 instead

gammapy/modeling/models/temporal.py
```diff
@@ -472,79 +471,93 @@ def __str__(self):
...
+            t_ref = m.meta["MJDREFI"]
+            return cls(m, filename=filename, t_ref=t_ref)
+
+        else:
+            raise ValueError("Not a valid format")
+
+    def to_table(self):
+        """Convert model to an astropy table"""
+        columns = self.map.geom.axes
+        table = Table(data=self.map.quantity, names=columns,
```
adonath, line 522: This does not work for the extra axes. For now it is sufficient to just support the "old" format and raise an error for a 2d model instead. I think we should introduce the `format` argument here as well, even if only one format is supported for now.

gammapy/modeling/models/temporal.py
```diff
@@ -472,79 +471,93 @@ def __str__(self):
...
+                      meta=self.map.meta)
+        return table
+
+    def write(self, filename, format="table", overwrite=False):
+        """Write a model to disk as per the specified format"""
```
adonath, line 527: Extend the docstring for the additional arguments here and document the selection of formats.

gammapy/modeling/models/temporal.py
```diff
@@ -472,79 +471,93 @@ def __str__(self):
...
+
+        if format == "table":
+            table = self.to_table()
+            table.write(filename, overwrite=overwrite)
+
+        elif format == "map":
+            self.map.write(filename, overwrite=overwrite)
+
+        else:
+            raise ValueError("Not a valid format")
```
adonath, line 540:
```suggestion
            raise ValueError(f"Not a valid format: '{format}', choose from: {'table', 'map'}")
```

gammapy/modeling/models/temporal.py
```diff
@@ -472,79 +471,93 @@ def __str__(self):
...
+
+    def evaluate(self, time, energy=None):
+        """Evaluate the model at given coordinates."""
+        coord = {"time": time.mjd}
+        if energy is not None:
+            coord["energy"] = energy
+
+        coords = self.create_coords(coord)
```
adonath, line 548: Is this working? The `self.create_coords()` method does not seem to exist? In any case it should be part of the non-public API, so please rename to `self._create_coords()`. But I'm not sure the method is even needed it seems it should just work to pass the coordinates to `self.map.interp_by_coord` like you do.

gammapy/modeling/models/temporal.py
```diff
@@ -472,79 +471,93 @@ def __str__(self):
...
+        val = self.map.interp_by_coord(coords)
+        val = np.clip(val, 0)
+        return u.Quantity(val, self.map.unit, copy=False)
+
+    def integral(t_min, t_max, **kwargs):
+        # for t1, t2 in zip(t_min, t_max):
+        #    map_slice = self.map.
+        map_interp = RegionNDMap.create(geom=self.map.geom,
+                                        data=val, unit=self.map.unit)
+        interp_map = map_interp.integral(axis_name="TIME")
```
adonath, line 558: This should take the integration boundaries...

adonath (changes requested):
Thanks @AtreyeeS, I have left a first round of inline comments...

adonath:
@AtreyeeS https://github.com/gammapy/gammapy/pull/4073 is merged, please rebase...

adonath:
@AtreyeeS What is the status here?

bkhelifi:
Hi @AtreyeeS, can you add the features mentioned in #4096, that is closed by @adonath to avoid conflicts? The goal was to fix the spline interpolator, and add the associated tests...
Thanks

gammapy/modeling/models/temporal.py
```diff
@@ -454,97 +445,145 @@ class LightCurveTemplateTemporalModel(TemporalModel):
...
+    def __init__(self, map, t_ref=None):
+
+        if (map.data < 0).any():
+            log.warning("Map has negative values. Check and fix this!")
+
+        self.map = map.copy()
+        if t_ref:
+            self.t_ref.value = Time(t_ref, format="MJD").mjd
+        if "ENERGY" in map.geom.axes.names:
+            self.is_energy_dependent = True
```
adonath, line 460: Please tun this into a property, such that it cannot be modified by users. Please also use https://docs.gammapy.org/0.20.1/api/gammapy.maps.Geom.html#gammapy.maps.Geom.has_energy_axis, to avoid hard-coding axis names.

gammapy/modeling/models/temporal.py
```diff
@@ -454,97 +445,145 @@ class LightCurveTemplateTemporalModel(TemporalModel):
...
         super().__init__()
 
     def __str__(self):
-        norm = self.table["NORM"]
+        start_time = self.map.axes["time"].edges[0]
+        end_time = self.map.axes["time"].edges[-1]
+        energy_min = self.map.axes["energy"].edges[0]
```
adonath, line 466: In general the energy axis optional...

gammapy/modeling/models/temporal.py
```diff
@@ -454,97 +445,145 @@ class LightCurveTemplateTemporalModel(TemporalModel):
...
+
+        elif format == "map":
+            self.map.write(filename, overwrite=overwrite)
+
+        else:
+            raise ValueError("Not a valid format")
+
+    def evaluate(self, time, energy=None):
+        """Evaluate the model at given coordinates."""
+        coords = {"time": time.mjd}
```
adonath, line 569: I think the time needs to be converted into a time relative to the reference time parameter.

gammapy/modeling/models/temporal.py
```diff
@@ -554,6 +593,11 @@ def to_dict(self, full_output=False):
         """Create dict for YAML serialisation"""
         return {self._type: {"type": self.tag[0], "filename": self.filename}}
 
+    def create_coords(self, coords):
```
adonath, line 596: Can be removed now I guess...

adonath (changes requested):
@AtreyeeS I have left some more comments...

adonath:
@AtreyeeS We have not talked about this yet, but my recommendation here would be to split the work:
- One PR to refactor the existing model to use a `RegionMap` internally and introducing the `t_ref` parameter
- A 2nd PR to introduce the energy dependence

All in all this might be a bit simpler to handle...

adonath:
@AtreyeeS We agreed at the dev call to postpone the energy dependent temporal models to v1.1. However its would be have the (small) API breaking change of changing the `TemplateTemporalModel` to rely on `Map` and `t_ref`. Otherwise we would need to go through a depreciation phase. So my proposal here would be to just include refactoring in this PR and make sure the existing tests pass. 

The support for energy will the come for v1.1 and we can take the time for proper testing.

AtreyeeS:
This PR should be ready for review now. 

1. `LightCurveTemplateTemporalModel.__init__` now takes a `map` instead of a `table`, so there are many API breaks. I think the validation scripts will also need to updated
2. I assume the template model to be defined at node centers
3. A bug in `TemplateModel.integral` is fixed

gammapy/modeling/models/temporal.py
```diff
@@ -174,10 +175,12 @@ def integral(self, t_min, t_max, oversampling_factor=100, **kwargs):
...
             Integrated flux norm on the given time intervals
         """
-        t_values = np.linspace(t_min.mjd, t_max.mjd, oversampling_factor, axis=-1)
+        t_values, steps = np.linspace(
+            t_min.mjd, t_max.mjd, oversampling_factor, retstep=True, axis=-1
+        )
         times = Time(t_values, format="mjd")
         values = self(times)
-        integral = np.sum(values / oversampling_factor, axis=-1)
+        integral = np.sum(values, axis=-1) * steps
```
adonath, line 183: This change seems to cause the changes in the test values. But looking at the implementation, it seems, that now it is correct. Previously it was `1 / factor`, so divided by the number of nodes (including bounds), while now it is `1 / (factor - 1)`, which the number of bins themselves. So looks good to me!

adonath (approved):
Thanks a lot @AtreyeeS! I see some minor issues in the docstrings, but those can be fixed uniformly anyway. Otherwise looks good to me!

PR #4085, opened 2022-09-16
Adapt map resampling API

adonath:
**Description**

This pull request adapts the `Map.resampled` API to return a map object. This makes the API consistent with what we have everywhere else.


QRemy (approved):
Thanks @adonath no comments from my side.

adonath:
Thanks @QRemy!

PR #4086, opened 2022-09-17
Remove jupyter tooling

adonath:
**Description**

This pull request is a follow up to #4080. It removes the custom Jupiter notebook tools, which are not needed anymore.


registerrier (commented):
How come this fails only on windows?

adonath:
@registerrier I'm currently investigating. The issue with plotting negative errors is actually a real issue in Gammapy! So I noticed that sometimes errors and upper limits computed by the `FluxPointsEstimator` and `CountsStatistic` actually come out negative. Here is what I understood so far:

- Some of the upper limits computed in the flux profile tutorial are actually negative. However this is not surprising because the Fermi data uses a fixed background model, which is just overestimating the flux in certain regions. Background systematics are neglected. This behavior is understood and mostly fine.
- We have some cases in our tests where the negative error computed by the `CountsStatistics` comes out negative: https://github.com/gammapy/gammapy/blob/master/gammapy/stats/tests/test_counts_statistic.py#L31
- And some where it comes out positive: https://github.com/gammapy/gammapy/blob/master/gammapy/estimators/map/tests/test_excess.py#L139
- The behavior between the `TSMapEsimator` and `ExcessMapEstimator` has been shown to be consistent https://gammapy.github.io/gammapy-recipes/_build/html/notebooks/ts-vs-excess-map-estimator/ts-vs-exess-map-estimator.html, however we actually invert the sign (with respect to `CountsStatistic`) of `errn` here: https://github.com/gammapy/gammapy/blob/master/gammapy/estimators/map/excess.py#L254 

I first propose to actually adapt the sign of the value returned by `CountsStatistic.compute_errn()` for consistency with everything else in Gammapy....but I'm investigating whether we have other inconsistencies.

adonath:
I'll go ahead and merge this PR. I'll work on a follow up PR now to make the `gammapy download notebooks` command work again.

PR #4089, opened 2022-09-18
Change sign of the value returned by CountsStatistic.compute_errn

adonath:
**Description**

This pull request changes the sign return by `CountsStatistic.compute_errn` for consistency with all the other estimators in Gammapy. I adapted the description in the estimator docs as well. It also include the fix to address the recent test fails with plotting of negative error in Matplotlib 3.6.


adonath:
The failed tests will be fixed in #4086, after this PR is merged. However we should get green CI for this PR now.

docs/user-guide/estimators.rst
```diff
@@ -62,9 +62,9 @@ In addition, the following optional quantities can be computed:
 ================= =================================================
 Quantity          Definition
 ================= =================================================
-norm_errp         Positive error of the norm
-norm_errn         Negative error of the norm
-norm_ul           Upper limit of the norm
+norm_errp         Positive error of the norm, given as difference to the best fit norm
+norm_errn         Negative error of the norm, given as difference to the best fit norm
+norm_ul           Upper limit of the norm, fiven as abolsute value
```
registerrier, line 67: given as absolute value

Note: even if negative?

registerrier (commented):
Thanks @adonath .

I assume this can indeed be made consistent with the convention that errors are always positive numbers. One issue is that this is API breaking. We might consider it is worth it, but it has to be clearly visible to users.

docs/user-guide/estimators.rst
adonath: Fixed.

adonath:
> One issue is that this is API breaking. We might consider it is worth it, but it has to be clearly visible to users.
It is not API breaking, but yes, it is backwards incompatible. I will be described in the release notes for v1.0, do you think this will be sufficient?

adonath:
@registerrier I'll go ahead and merge this now, such that we get a green CI again. We can make additional changes to the documentation after and during preparation of the releases notes.

Issue #3639, opened 2021-11-20
Argument passed both by position and keyword in function call


DimitriPapadopoulos:
**Gammapy version**

`@master`

**Bug description**

`LightCurveTemplateTemporalModel.evaluate()` calls `LightCurveTemplateTemporalModel._interpolator()` with two arguments `time` and `ext`:
https://github.com/gammapy/gammapy/blob/ae5621e7ea0857a8b7dd22df2487165f066934fb/gammapy/modeling/models/temporal.py#L434
https://github.com/gammapy/gammapy/blob/ae5621e7ea0857a8b7dd22df2487165f066934fb/gammapy/modeling/models/temporal.py#L455

Yet `LightCurveTemplateTemporalModel._interpolator()` is defined to accept a single `ext` argument:
https://github.com/gammapy/gammapy/blob/ae5621e7ea0857a8b7dd22df2487165f066934fb/gammapy/modeling/models/temporal.py#L418-L419

adonath:
The code does not exist anymore after #4084.

PR #4096, opened 2022-09-20
Fix `LightCurveTemplateTemporalModel.evaluate`

bkhelifi:
**Description**
This pull request fixes the issue mentioned here: #3639 .

**Dear reviewer**
The `_interpolator` function is return and the code is adapted accordingly.
The issue for me is that the interpolator function is built each time the user wants an evaluation at a given time, which is a lost of time. If you agree, I propose that the interpolator function is built in the init function and saves into a variable of the class. Or, there is a magic python decorator...

### DIFF
--- a/gammapy/modeling/models/temporal.py
+++ b/gammapy/modeling/models/temporal.py
@@ -489,7 +489,6 @@ class LightCurveTemplateTemporalModel(TemporalModel):
             self.filename = str(make_path(path))
             self.table.write(self.filename, overwrite=overwrite)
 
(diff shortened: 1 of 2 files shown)


adonath:
Thanks @bkhelifi! In #4084 , @AtreyeeS will actually refactor the temporal model evaluation anyway. If I merge this PR I will just create a merge conflict and the changes will be need to be refactored by @AtreyeeS anyway. If you do not mind, I'd prefer to just close this PR instead. Please re-open if you think otherwise...

Issue #3708, opened 2022-01-20
SafeMaskMaker assumes estimated energy = true energy


maxnoe:
**Gammapy version**
All versions including current master

**Bug description**

The `SafeMaskMaker` supports using effective area to calculate a "safe" energy range.

However, the safe mask is applied in estimated energy where effective area is in true energy,
so this silently assumes no energy bias, which is violated for most experiments especially in the region 
this tool should identify.

**Expected behavior**

Do not **silently** make these assumptions.

Either warn about this or require the needed information (edisp and assumed spectrum) to make the actual conversion 
possible.

PR #4097, opened 2022-09-20
Precise that the `SafeMaskMaker` uses the true energy range to select events for the methods "aeff*"

bkhelifi:
**Description**

This pull request is associated to the issue #3708. It completes the documentation to precise the made assumption.


adonath (dismissed):
Thanks @bkhelifi, looks good to me!

gammapy/makers/safe.py
```diff
@@ -17,6 +17,11 @@
 class SafeMaskMaker(Maker):
     """Make safe data range mask for a given observation.
 
+    
+    .. warning::
+
```
registerrier, line 22: Technically this is true for `aeff-default` `aeff-max` and `edisp-bias`as well.

Maybe :  'Note that currently some methods computing a safe energy range (`aeff-default` `aeff-max` and `edisp-bias`) determine a true energy range and apply it to reconstructed energy, effectively neglecting the energy bias.

examples/tutorials/api/makers.py
```diff
@@ -131,7 +131,11 @@
 #    energy bin with the highest predicted background rate. This method
 #    was introduced in the HESS DL3 validation paper:
 #    https://arxiv.org/pdf/1910.08088.pdf
-# 
+#
```
registerrier, line 134: See comment below

registerrier (approved):
Thanks @bkhelifi , a small clarification below

gammapy/makers/safe.py
maxnoe: energy bias → energy dispersion (it's not only the mean bias that is neglected, but the error in each individual event)

bkhelifi: Indeed, both rms and bias play a role. And 'energy dispersion' is the term used in our IRF theory. Thanks @maxnoe

registerrier (approved):
Thanks @bkhelifi !

adonath (approved):

PR #4087, opened 2022-09-17
Introduce event sampling of spectrally varying temporal model

fabiopintore:
This PR will introduce the possibility to event sample a spectrally varying temporal model with `MapDatasetEventSampler`. The PR is still in draft.

PR #4100, opened 2022-09-20
Implement the _sample_coord_time_energy function in MapDatasetEventSa…

fabiopintore:
This PR will introduce the possibility to event sample a spectrally varying temporal model with `MapDatasetEventSampler`. The PR is still in draft and fix the #4087.


adonath:
Thanks @fabiopintore!

I think the implementation should simplify a lot, once https://github.com/gammapy/gammapy/pull/4084 is merged. You could refactor your implementation, assuming that `TemplateTemporalModel` has a `.map` attribute you can directly sampled from.
There is no need to re-evaluate the model.

registerrier:
> I think the implementation should simplify a lot, once #4084 is merged. You could refactor your implementation, assuming that `TemplateTemporalModel` has a `.map` attribute you can directly sampled from. There is no need to re-evaluate the model.

I am not sure we can avoid resampling the model `Map`.  For two reasons:
1. There is no reason to assume the model `Map` definition is adapted to the `EventSampler`. Sampling from the model `MapGeom` time axis is not valid. Some reprojection is required.
2. Coord sampling is performed by first determining randomly a pixel in  the `Map` geometry. And then randomly determine precise coordinates inside the pixel. To be consistent with regular model evaluation and integration, this **requires** oversampling of the geometry. Otherwise the nearest-neighbor vs linear scheme of sampling/evaluation will be at odds. For spatial template models, this is how it is performed in the `MapEvaluator`, the model is evaluated on the `exposure` map.

adonath:
> There is no reason to assume the model Map definition is adapted to the EventSampler. Sampling from the model Geom time axis is not valid. Some reprojection is required.

I see what you mean. I guess this is mostly related to efficiency. Sampling events is only required for the part where the model overlaps with the requested geom. 

> Coord sampling is performed by first determining randomly a pixel in the Map geometry. And then randomly determine precise coordinates inside the pixel. To be consistent with regular model evaluation and integration, this requires oversampling of the geometry. Otherwise the nearest-neighbor vs linear schemes of sampling/evaluation will be inconsistent. For spatial template models, this is how it is performed in the MapEvaluator, the model is evaluated on the exposure map.

Yes, you are right. The event sampling only does the linear interpolation in the coordinate axis and not the data axis. So the scheme is indeed, that we have to linearly upsample the map to binning sufficiently smaller than the requested geom and then do the sampling.

registerrier:
> The question is also what to do with the spatial coordinates here. To keep the sub-pixel position, one would need to work with a map with 4 pixels instead of the one pixel of the `RegionGeom`.

If the center of the region is the `PointSpatialModel.position` we have directly the position. No need to go sub-pixel no?

adonath:
Indeed one could work directly with a SkyPointRegion when sampling, which should give directly the right coordinate.

gammapy/datasets/simulate.py
```diff
@@ -26,6 +27,66 @@ class MapDatasetEventSampler:
     def __init__(self, random_state="random-seed"):
         self.random_state = get_random_state(random_state)
 
+    def _sample_coord_time_energy(self, dataset, temporal_model, t_delta="1 sec"):
```
registerrier, line 30: Is a fixed `t_delta` the simplest approach? One could take an oversampling factor as well. It could take the mean of the dataset duration and the model time interval

registerrier, line 30: Do you need to pass the full dataset or simply the `MapEvaluator` as in the other `sample_coord_energy`?

gammapy/datasets/simulate.py
```diff
@@ -26,6 +27,66 @@ class MapDatasetEventSampler:
...
+            else ontime.unit
+        )
+
+        t_step = t_delta.to_value(time_unit)
+        t_step = (t_step * u.s).to("d")
+
+        t = Time(np.arange(t_min.mjd, t_max.mjd, t_step.value), format="mjd")
+        time_axis = MapAxis.from_edges(t.value * u.d, name="time")
+
+        target = temporal_model.geom.skycoord #change it when the TemplateTemporalModel is updated
```
registerrier, line 54: There are two places where you can get the coordinate, the region of the temporal model and the `PointSpatialModel` I would use this one

gammapy/datasets/simulate.py
```diff
@@ -26,6 +27,66 @@ class MapDatasetEventSampler:
...
+        on_region = PointSkyRegion(center=target)
+        region_geom = RegionGeom.create(on_region,
+                                        axes=[energy_axis, time_axis])
+    
+        flux = temporal_model.evaluate_geom(region_geom)
+        region_exposure = dataset.exposure.to_region_nd_map(region_geom.center_skydir)
```
registerrier, line 60: Here you could equivalently use the exposure in the `MapEvaluator`

gammapy/datasets/simulate.py
```diff
@@ -26,6 +27,66 @@ class MapDatasetEventSampler:
...
+        
+        npred = flux * region_exposure #check the broadcasting
+        data = npred.data[np.isfinite(npred.data)]
+        n_events = self.random_state.poisson(np.sum(data))
+
+        coords = npred.sample_coord(n_events=n_events, random_state=self.random_state)
+
+        position = SkyCoord(coords["lon"], coords["lat"],
```
registerrier, line 68: `position = coords.skycoord` should work no?

gammapy/datasets/simulate.py
```diff
@@ -26,6 +27,66 @@ class MapDatasetEventSampler:
...
+
+        target = temporal_model.geom.skycoord #change it when the TemplateTemporalModel is updated
+        on_region = PointSkyRegion(center=target)
+        region_geom = RegionGeom.create(on_region,
+                                        axes=[energy_axis, time_axis])
+    
+        flux = temporal_model.evaluate_geom(region_geom)
+        region_exposure = dataset.exposure.to_region_nd_map(region_geom.center_skydir)
+        
+        npred = flux * region_exposure #check the broadcasting
```
registerrier, line 62: technically we are performing integration here so you need to multiply by the bin volume as well

gammapy/datasets/simulate.py
registerrier, line 68: Maybe all the `MapCoord` to table manipulation could be performed utside the function as it is common with the other case

registerrier (commented):
Thanks @fabiopintore . Here are some inline comments.

A test will be needed.

Should the function also support regular `TemporalModel`? or even be applied to all `Model` with a `PointSpatialModel`?

fabiopintore:
As far as I understood, the idea was to have a special function only for the `TemplateTemporalModel` with the `energy` column. Otherwise we should modify the `_sample_coord_time` and discard the new `_sample_coord_time_energy`, right?

gammapy/datasets/simulate.py
fabiopintore: I used the `t_delta` parameter so to be in line with the approach of `_sample_coord_time`. If we agree to follow your suggestion, we should change it in both functions for consistency. For now I implement it in the `_sample_coord_time_energy`.

fabiopintore: Ah right, we can just pass the `MapEvaluator`. I've changed it.

gammapy/datasets/simulate.py
fabiopintore: Ok.

gammapy/datasets/simulate.py
fabiopintore: OK.

gammapy/datasets/simulate.py
fabiopintore: Yes, it works. I've changed it

gammapy/datasets/simulate.py
fabiopintore: OK.

registerrier:
Since we miss the proper model to make any test, it was decided to postpone the addition of this feature to the next feature release

PR #4101, opened 2022-09-21
Correct CountsStatistic documentation plots

registerrier:
**Description**

This pull request corrects a problem with `CountsStatistic` documentation plots that appeared after #4089 was merged.
See https://docs.gammapy.org/dev/user-guide/stats/plot_wstat_errors.png

The documentation in the Makefile is also modified to correctly give the `make docs-sphinx` command instead of `make docs` which is no longer valid.

**Dear reviewer**


bkhelifi (approved):
Thanks @registerrier

adonath (approved):

PR #4102, opened 2022-09-21
Fix weights in resample

QRemy:
This PR fix a bug introduced in #4085 where the weights in resample are ignored and add a test for that case.


adonath (approved):
Thanks a lot @QRemy! No comments from my side...

PR #4103, opened 2022-09-21
Activate binder for tutorials

adonath:
**Description**

This pull request activates the Binder service for our tutorials. To allow for a dat download in Binder, but also because it is useful in general I introduced a `check_tutorials_setup` method, which downloads the data if not present and prints a summary of the environment as given by `gammapy info`.

To make this work for a past versions of stable release one has to manually update the `docs/binder/requirements.txt` file. However this can also happen later manually e.g. in https://github.com/adonath/gammapy-docs-preview/tree/main/binder. We have to see how to integrate this best into the release process...

You can checkout a preview here https://adonath.github.io/gammapy-docs-preview/


adonath:
The setup seems to work on my doc preview repo. I'll now proceed and change it to our docs repo, however it will only become available., once this PR is merged.

adonath:
I'll go ahead and merge this now to check whether this works for our dev version. I'll report back on any issues here and reopen if needed...

Issue #4062, opened 2022-09-01
gammapy/gammapy-data: management of metadata 


bkhelifi:
**Is your feature request related to a problem? Please describe.**
The Gammapy test features are associated to data stored in the project [gammapy/gammapy-data](https://github.com/gammapy/gammapy-data). Because we anticipate changes in the data model associated to these data, each Gammapy release should have its associated gammapy-data release.

**Describe the solution you'd like**
- Create a release CI for gammapy-data
- Create/update the metadata of this repository (a script might be needeed to 'synchronize' these metadata with the ones of the associated Gammapy release)
- Create a workflow to make a Zenodo entry for each gammapy-data

PR #4106, opened 2022-09-23
Reactivate gammapy download command

adonath:
**Description**

This pull request reactivates the `gammapy download` command and adapts it to work with the archive file created by `sphinx-gallery`. For this I introduced a download index file here: https://github.com/gammapy/gammapy-webpage/blob/gh-pages/download/index.json

This allows us to declare the files to download independently of the code base. I additionally added the index for datasets, which allows us to address #4062 even after we have tagged v1.0.


bkhelifi:
Thanks @adonath, it seems great. There is a minor error for one test, but this is a detail...
A general question: should we 'review' the release note to be sure to not have forgotten something?

adonath:
@bkhelifi  I will review and adapt the developer docs on how to make a release next week once again.

PR #4107, opened 2022-09-23
Implement peek methods for map evaluator and psf kernel

luca-giunti:
**Description**

This pull request overrides #4074 (due to merge conflicts)


registerrier (commented):
Thanks @luca-giunti .

Looks good to me. You might just add a test for `MapEvaluator.peek()` with a `WcsNDMap` based dataset.

luca-giunti:
I added the requested test. Probably we could make a better use of fixtures to save lines of code, but I guess it's a general issue that will be addressed consistently at some point

adonath:
@luca-giunti Yes I agree, refactoring of tests should be a priority after v1.0!

adonath (approved):
Thanks @luca-giunti, no further comments from my side!

PR #4108, opened 2022-09-23
Add tests with HAWC data

LauraOlivera:
This pull request adds tests using the public HAWC event release. In particular there are new tests for:

- Reading and handling `RecoPSFMap` objects
- Running the data reduction on a dataset with a reconstructed energy PSF
- Fitting a model to several datasets (one per event class) jointly with a reconstructed energy PSF.
The last test is the slowest, at ~30sec

 It also fixes the logic in the `MapDatasetMaker` because it was in the wrong order before (`RecoPSFMap` is an instance of `PSFMap` due to inheritance)

The data needed for these tests is added by [this PR](https://github.com/gammapy/gammapy-data/pull/28) to gammapy-data


QRemy:
Thanks @LauraOlivera
Maybe you can reduce the geom binsz or the number of bins along the energy axis to speed up the tests.
Eventually the `test_fit_hawc `could go into gammapy-benchmark/validation

adonath:
Thanks @LauraOlivera! I agree with @QRemy, once we know `MapDataset.npred()` is working I think we can be confident enough. The full model fit could go into https://github.com/gammapy/gammapy-benchmarks/tree/master/validation instead.

LauraOlivera:
Thanks!! I've removed the full model fit test and will move it to gammapy-benchmarks

adonath:
Thanks a lot @LauraOlivera!

adonath (approved):

Issue #4019, opened 2022-07-01
Add support for NormSpectralModel in FluxEstimator


AtreyeeS:
Questions about using full 3D `TemplateSpatialModel` with units of, eg `cm-2 s-1 TeV-1 sr-1` seem to be frequent.
We can maybe add a HowTo entry for this before 1.0 release.

QRemy:
Could link to :
https://docs.gammapy.org/dev/tutorials/api/models.html?highlight=powerlawnorm#SkyModel (at the end of the section)
https://docs.gammapy.org/dev/tutorials/data/fermi_lat.html?highlight=templatespatialmodel#Galactic-diffuse-background

adonath:
Thanks @AtreyeeS! Indeed I think norm spectral models are not supported...I guess we should raise a meaningful error for now?

AtreyeeS:
I don't have a better solution for the moment ... but must admit it is annoying for a LTS version, because this can be useful for studies of the the diffuse emission, etc.
Maybe we can suggest a workaround (link to a GitHub discussions?) buy "transferring" the amplitude spectral model and creating a new spatial model? (or maybe you have a better workaround?)

adonath:
On closer inspection all information is actually there: 
- We have `flux_points.reference_model`, which also contains the `TemplateSpatialModel`
- The `flux_points.dnde_ref` (and `flux_ref` etc.) would need to be changed to take into account the spatial model as well.
- The main question is just which convention to use for the latter, probably integrate the spatial model on the full support and then multiply with the spectral norm model.
- The second question is whether the reference model should be kept as a spatial model, or whether it is simplified to an integrated model, stored in a `TemplateSpectralModel`...

adonath:
@QRemy This issue is assigned to you. Do you have time to open a PR to raise and error, that `NormSpecralModel` objects are not supported for now?

adonath:
Nothing left to be done for v1.0

AtreyeeS:
@adonath Maybe re-label this to `1.1` or later and keep open as a reminder? This is something we should support in future

adonath:
Thanks @AtreyeeS, I agree. I just changed the title and will re-open now...

PR #4109, opened 2022-09-23
Raise error in FluxPointsEstimator for NormSpectralModel

QRemy:
Raise an error in FluxPointsEstimator for NormSpectralModel to delay solving the issue in 
https://github.com/gammapy/gammapy/issues/4019#issuecomment-1187637137

### DIFF
--- a/gammapy/estimators/points/sed.py
+++ b/gammapy/estimators/points/sed.py
@@ -95,7 +95,7 @@ class FluxPointsEstimator(FluxEstimator):
         """
         datasets = Datasets(datasets=datasets)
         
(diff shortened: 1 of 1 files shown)


gammapy/estimators/points/sed.py
```diff
@@ -94,8 +94,11 @@ def run(self, datasets):
             Estimated flux points.
         """
         datasets = Datasets(datasets=datasets)
+        
+        if "NormSpectralModel" in datasets.models[self.source].spectral_model.tag[0]:
```
adonath, line 98:
```suggestion
        if datasets.models[self.source].spectral_model.is_norm_spectral_model:
```

adonath (approved):
Thanks, @QRemy!

PR #4112, opened 2022-09-27
Add pre commit hooks and black CI

adonath:
**Description**

This pull request proposes to add two pre-commit hooks to Gammapy:

- Running the `isort` tool
- Running the `black` code formating tool

The motivation is that we used these tool anyway, however only to improve codestyle before releases. This way we miss once of the main advantages: smaller diffs in PRs. Therefore I propose to now enforce the `black` code style, by introducing a C check for this. Together with the "codestyle" check we now enforce a consistent code style in Gammapy. To avoid the extra work for the developers, I introduced the pre-commit hooks, which can be activated by calling `pre-commit install` once.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,6 +2,12 @@ name: CI
 on: [push, pull_request]
 
 jobs:
+  lint:
+    name: Black check
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: psf/black@stable
   ci-runs:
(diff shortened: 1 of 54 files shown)


bkhelifi (approved):
Thanks a lot @adonath !!!! This is a great feature that will help

registerrier (approved):
Thanks @adonath . This looks good!

This will remove the need for large code formatting updates before very release at minimal cost for developers.

adonath:
Thanks @registerrier  and @bkhelifi!

PR #4116, opened 2022-09-27
Adapt Astropy and Numpy versions in tox configuration

adonath:
**Description**

This pull request drops the support for Numpy 1.19 and remove the Astropy 4.0 configuration from `tox.ini`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ jobs:
     env:
       PYTEST_ADDOPTS: --color=yes -n auto --dist=loadscope
     strategy:
-      fail-fast: true
+      fail-fast: false
       matrix:
         include:
           - os: ubuntu-latest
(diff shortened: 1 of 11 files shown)


adonath:
Test are passing now for Numpy 1.23 an all platforms. The modifications for the tests I included were:
- The map sampling test seemed numerically unstable, because the Gaussian source, was center on a pixel and exactly symmetric. This lead to pixel values being the same also at the weak tails of the Gaussian. The seemed to result in a random ordering (or at least differing ordering) of those pixels. By chance one of the events was in one of those pixels.
- The spectrum tests were performed with linear energy binning between 1 - 10 TeV, which I adapted to log binning.
- The analysis 3d test shows very high correlation (0.87) between both the amplitude and index of the power-law model as well as the norm and tilt of the background model. This makes the test also numerically unstable. For now I increase the true energy binning, which seemed to help.

registerrier (approved):
Thanks @adonath ! Good to see CI running fine.

Issue #4117, opened 2022-09-29
3D (Fermi) + 1D (LST/MAGIC PointSkyRegion-style) analysis. Global fit ignores the 1D. 


mireianievas:
**Gammapy version**
0.20.1

Are you using the latest stable version of Gammapy? (see https://gammapy.org/news.html)
yes

**Bug description**
Issue discussed already in slack (and I should probably have opened this ticket back then, my bad). A 3D (e.g. Fermi) and a 1D ON/OFF datasets (PointSkyRegion type, e.g. LST or MAGIC) are analyzed together with N skymodels for the 3D (with spatial and spectral components, including those for the source of interest) and 1 skymodel (the same model as in the 3D for ONLY the source of interest, including spectral and spatial components). The result, after the likelihood fit, is that one of the datasets is completely ignored in the fit. However, no error or warning is raised, so I have no idea how to debug it. 

<img width="730" alt="Captura de Pantalla 2022-09-29 a las 11 35 47" src="https://user-images.githubusercontent.com/2607760/193009687-d5d1bee3-9a8e-4fa0-852e-d625918a63cd.png">

Partial workaround: 
1) define a skymodel for the 1D dataset consisting only on a spectral component, no spatial ones. 
2) The parameters of the spectral components are linked/fixed to the corresponding spectral component of the source model in the 3D dataset (lets call the source name in the 3D dataset XXX). 
3) The source name of the skymodel then needs to be different, e.g. renamed to XXX-2, otherwise gammapy returns an error saying that the names should be unique.
4) perform the fit, and then voila both datasets contribute to the result, the confidence band is much narrower.

<img width="738" alt="Captura de Pantalla 2022-09-29 a las 11 35 57" src="https://user-images.githubusercontent.com/2607760/193009783-7fc169bf-0e19-4e89-887f-4b8cd6335b7f.png">

However, this is not a perfect solution, because the FluxPointEstimator needs to be called for each dataset with their unique source names (even when it is the same source and both datasets should contribute to the energy ranges where both instruments are sensitive). So you have Fermi points on one hand and IACT points on the other. At the overlapping energy range (~20-300 GeV)  you may have 'duplicated' points with larger error bars (because only one dataset contributes at a time). 

**Expected behavior**

[...]

PR #4118, opened 2022-09-30
Issue #4117

adonath:
**Description**

This pull request addresses the issues raised in #4117. It includes the following checks:
- Check that the dataset name is a `str`
- Modify the behavior for `RegionGeom.contains` and `PointSkyRegion`. The implementation in `regions`, does not include the coordinate even if it is exactly at the center of the region. The point sky region here is approximated by a rectangular region of the size of 1x1 pixel.

### DIFF
--- a/gammapy/datasets/core.py
+++ b/gammapy/datasets/core.py
@@ -524,7 +524,7 @@ class Datasets(collections.abc.MutableSequence):
         if np.all([table is None for table in tables]):
             meta_table = Table()
         else:
-            meta_table = vstack(tables)
+            meta_table = vstack(tables).copy()
 
         meta_table.add_column([d.tag for d in self], index=0, name="TYPE")
         meta_table.add_column(self.names, index=0, name="NAME")
--- a/gammapy/datasets/map.py
+++ b/gammapy/datasets/map.py
@@ -2247,6 +2247,7 @@ class MapDatasetOnOff(MapDataset):
 
     def to_map_dataset(self, name=None):
         """Convert a MapDatasetOnOff to  MapDataset
+
         The background model template is taken as alpha * counts_off
 
         Parameters
--- a/gammapy/datasets/map.py
+++ b/gammapy/datasets/map.py
@@ -644,7 +644,9 @@ class MapDataset(Dataset):
         )
 
         kwargs.update(geoms)
-        return cls.from_geoms(reference_time=reference_time, name=name, **kwargs)
+        return cls.from_geoms(
+            reference_time=reference_time, name=name, meta_table=meta_table, **kwargs
+        )
 
     @property
     def mask_safe_image(self):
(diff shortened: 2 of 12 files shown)


registerrier (commented):
Thanks @adonath .  

It is not clear to me what we gain by treating the `PointSkyRegion` as a `Rectangle` brings.  Is the issue with `contain` connected with the necessity of taking into account the varying ON region size? Is returning `True` all the time a possibility as well?

mireianievas:
@registerrier astropy.region developers were in favour of keeping False as a returning value (argument they used: a point does not include anything). Which is somewhat logical. 

Yes, one can force PointSkyRegion.contain to return True by setting the meta information as {'include': False} . But I guess the ideal case is to use something different which behaves like PointSkyRegion except it has some dimensions to include the point and return True.

adonath:
@registerrier The use of the `RectangleSkyRegion` here is related to the fact that we internally rely on a `WCS` with finite pixel size. A point source on a WCS grid is represented by a 1x1 pixel, equivalent to our `PointSpatialModel`. I have not checked where it would be relevant for the `RegionGeom` apart from `.contains(), probably methods like `get_wcs_coord_and_weights`.

registerrier (commented):
Thanks @adonath.

What is the new behaviour introduced for small bins? They are now skipped? How is the flux points axis defined, then?
Maybe you could clarify why this is needed? I don't see the link with the issue #4117 .

adonath:
@registerrier Those changes kind of happened "accidentally", when I refactored the code. What happened was the following:

- The current code fails for the flux point estimation with datasets of different energy ranges, because quantities like `npred` or `excess`, actually have a `dataset` axes. 
- I decided to introduce the same behavior as for the `LightCurveEstimator`, where we just ues a placeholder value, if datasets do not contribute to the given time bin. 
- This happened by using the same pattern with `FluxPoints.from_stack()`
- However it modified the the behavior for the small energy bins

All in all I'm not even sure whether we should actually allow the flux point computation with "mixed" datasets at all. Users will need to be very careful with choices of energy bins for the different datasets and how they align. I think this will require better documentation...

adonath:
@mireianievas I just talked to @registerrier and we realized that right now we should not allow to use datasets from different instruments for flux point computation in Gammapy. Currently Gammapy uses a "hybrid" methods, where flux points are fitted in reconstructed energy bins. Different instruments have varying energy resolution and bias and this way there is not clear definition of an energy bin anymore. In addition when limiting the fit to a small energy range the handling of systematics becomes more important as well, which requires a more flexible likelihood definition as well. I'll now modify the PR to raise a more meaningful error if flux points are computed with heterogeneous datasets. 

The support for multi-instrument flux points will come once we add unfolding of spectra to Gammapy. There is currently one member of the dev team working on this. So it might come with the next minor feature release v1.1 scheduled for spring next year...

PR #4119, opened 2022-09-30
simplification of np.array(set(labels))

mireianievas:
Simplification of unique_labels using np.unique instead of sets. Avoids calling again the conversion to np.array at the end of the function.

**Description**

This pull request was suggested in issue #4117


adonath (approved):
Thanks a lot @mireianievas for the simplification here!

PR #4125, opened 2022-10-05
Add script to perform summary of merged PRs and closed issues 

registerrier:
**Description**

This pull request proposes a script to query the GitHub list of closed PRs and issues to list all merged PRs and count closed issues with a given milestone.

Usage:
```
python dev/github_summary.py  merged_PR 1.0 --token XXXX
python dev/github_summary.py  closed_issues 0.20 --token XXXX --print_issues True
``` 

It requires to install PyGithub (https://github.com/PyGithub/PyGithub)
**Dear reviewer**

### DIFF
--- a/dev/github_summary.py
+++ b/dev/github_summary.py
@@ -5,13 +5,17 @@ import logging
 
 log = logging.getLogger(__name__)
 
+
 @click.group
 @click.option(
-    "--log-level", default="INFO", type=click.Choice(["DEBUG", "INFO", "WARNING"]),
+    "--log-level",
+    default="INFO",
+    type=click.Choice(["DEBUG", "INFO", "WARNING"]),
 )
 def cli(log_level):
     logging.basicConfig(level=log_level)
     log.setLevel(level=log_level)
+
 
 def login(token):
     if token:
@@ -28,6 +32,7 @@ def login(token):
 
     return g
 
+
 def check_requests_number(g):
     remaining, total = g.rate_limiting
     log.info(f"Remaining {remaining} requests over {total} requests.")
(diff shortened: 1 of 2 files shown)


adonath:
@registerrier Be default I included the dev scripts in the CI check as well, just because they don't change often. However I'm not sure the dev scripts are included in the pre-commit hook. Maybe you can check and adapt the PR...

registerrier:
> @registerrier Be default I included the dev scripts in the CI check as well, just because they don't change often. However I'm not sure the dev scripts are included in the pre-commit hook. Maybe you can check and adapt the PR...

It is included in the pre-commit hooks. I was not executing them :).

bkhelifi:
Thanks @registerrier, great implementation.
I have two thoughts:
- the functionality provided by the click function `list_merged_PRs` is the same than the one used to build a contributor list. I am wondering if it should not be into a dedicated function into a ``dev/utils.py``... such that one can re-use it
- this nice implementation with PyGitHub provides the same type of information than `git shortlog --merges --since XX --until YY`. Do I mistaken? If yes, what are the pros and cons of each method?

adonath:
It would be good to have something like this, but we should try to improve the implementation.

registerrier:
The code has been completely changed. It now consists in a a general class `GitHubInfoExtractor` which deals with the github interface (login, token, repo) and whose main method build a summary table of PRs.
This table is then parsed by a helper function to produce an output list of PRs, the list of contributors (for now based on the list of people who opened the PRs).

adonath (dismissed):
Thanks a lot @registerrier, I have no comments! This is a very useful and cleanly written script!

dev/github_summary.py
```diff
@@ -0,0 +1,232 @@
...
+        commits = pull_request.get_commits()
+        result.update(self._get_commits_info(commits))
+        # extract reviews
+        reviews = pull_request.get_reviews()
+        result.update(self._get_reviews_info(reviews))
+
+        return result
+
+    def extract_pull_requests_table(
+        self, state="closed", number_min=0, include_backports=False
```
AtreyeeS, line 110: Github counts from 1, so I guess the default `number_min = 1`

AtreyeeS (approved):
Thanks @registerrier !

dev/github_summary.py
registerrier, line 110:
```suggestion
        self, state="closed", number_min=1, include_backports=False
```

Issue #4124, opened 2022-10-04
Serialising covariance of CompoundSpectralModel with a TemplateSpectralModel


AtreyeeS:
**Bug description**

The covariance matrix is incorrectly written for a CompoundSpectralModel which is a product of a TemplateSpectralModel and a NormSpectralModel

**To Reproduce**

Either serialise the `diffuse_iem` model in the Fermi-LAT tutorial, or

```
from gammapy.modeling.models import *
t = TemplateSpectralModel(energy=[1,3,9]*u.TeV, 
                        values=[20,30,40]*1 / (u.cm*u.cm*u.MeV*u.s))
p = PowerLawNormSpectralModel()
sk = SkyModel(spectral_model=t*p, name="xx")
Models([sk]).write("test.yaml", overwrite=True)
m1 = Models.read("test.yaml")
```

This leads to an error 

```
File ~/Gammapy-dev/gammapy/gammapy/modeling/models/core.py:362, in DatasetModels.covariance(self, covariance)
    359 @covariance.setter
    360 def covariance(self, covariance):
    361     self._check_covariance()
--> 362     self._covariance.data = covariance
    364     for model in self._models:
    365         subcovar = self._covariance.get_subcovariance(model.covariance.parameters)

File ~/Gammapy-dev/gammapy/gammapy/modeling/covariance.py:59, in Covariance.data(self, value)
     57 shape = (npars, npars)
     58 if value.shape != shape:
---> 59     raise ValueError(
     60         f"Invalid covariance shape: {value.shape}, expected {shape}"
     61     )
     63 self._data = value

ValueError: Invalid covariance shape: (13, 12), expected (13, 13)
```

This occurs because the `norm` parameter is present on both the `TemplateSpectralModel` and the normed model, and is incorrectly serialised during writing.

**Other information**

It would be good to sort this before `1.0`
Otherwise, Fermi datasets serialisation is difficult.

adonath:
Just to clarify: I think the issue is not that both parameters are called `norm`, but that the norm parameter of the `TemplateSpectralModel` is not serialized.

registerrier:
I think the `norm` parameter is useful because it provides at least one degree of freedom. One could assume that all template spectral models come with a normalization parameter, which is not the case at the moment.
Of course the behavior can be obtained multiplying by a `NormSpectralModel` as in the example above.

adonath:
[...]

PR #4126, opened 2022-10-05
Fix TemplateSpectralModel serialisation

AtreyeeS:
This fixes the parameter serialisation in `TemplateSpectralModel` as discussed in #4124 

The `TemplateNdSpectralModel` takes the paramters from its axes. If an axes is called norm, then it creates a problem, so I did not adapt it

The issue with the covariance matrix is still unresolved

### DIFF
--- a/gammapy/modeling/models/spectral.py
+++ b/gammapy/modeling/models/spectral.py
@@ -1624,26 +1624,29 @@ class TemplateSpectralModel(SpectralModel):
         return norm * self._evaluate((energy,), clip=True)
 
     def to_dict(self, full_output=False):
-        return {
-            self._type: {
-                "type": self.tag[0],
-                "energy": {
-                    "data": self.energy.data.tolist(),
-                    "unit": str(self.energy.unit),
-                },
-                "values": {
-                    "data": self.values.data.tolist(),
-                    "unit": str(self.values.unit),
-                },
-            }
+        data = super().to_dict(full_output)
+        data["spectral"]["energy"] = {
+            "data": self.energy.data.tolist(),
+            "unit": str(self.energy.unit),
         }
+        data["spectral"]["values"] = {
+            "data": self.values.data.tolist(),
+            "unit": str(self.values.unit),
+        }
+
+        return data
 
     @classmethod
     def from_dict(cls, data):
         data = data["spectral"]
         energy = u.Quantity(data["energy"]["data"], data["energy"]["unit"])
         values = u.Quantity(data["values"]["data"], data["values"]["unit"])
-        return cls(energy=energy, values=values)
+        model = cls(energy=energy, values=values)
+        for idx, p in enumerate(model.parameters):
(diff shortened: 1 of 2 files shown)


AtreyeeS:
This can be implemented independently of the covariance issue.

gammapy/modeling/models/spectral.py
```diff
@@ -1624,26 +1624,29 @@ def evaluate(self, energy, norm):
...
 
+        return data
+
     @classmethod
     def from_dict(cls, data):
         data = data["spectral"]
         energy = u.Quantity(data["energy"]["data"], data["energy"]["unit"])
         values = u.Quantity(data["values"]["data"], data["values"]["unit"])
-        return cls(energy=energy, values=values)
+        model = cls(energy=energy, values=values)
```
registerrier, line 1644: I suppose that `norm` should be instantiated by the `__init__`  abit like what is done for `ABLAbsorptionSpectralModel`.
See : https://github.com/gammapy/gammapy/blob/c5697fd2c91fb9728686f44d8a19bfbfd103e5be/gammapy/modeling/models/spectral.py#L1807

This would avoid modifying parameters afterwards.

gammapy/modeling/models/spectral.py
```diff
@@ -1736,18 +1739,21 @@ def write(self, overwrite=False):
 
     @classmethod
     def from_dict(cls, data):
+        data = data["spectral"]
         filename = data["filename"]
         m = RegionNDMap.read(filename)
         model = cls(m, filename=filename)
         for idx, p in enumerate(model.parameters):
-            p.value = data["spectral"]["parameters"][idx]["value"]
+            par = p.to_dict()
```
registerrier, line 1747: Same comment here modifying the constructor to properly take into account the norm might prove simpler.

registerrier (approved):
Thanks @AtreyeeS . 

My main comment is that the `TemplateSpectralModel.__init__` might be modified to have `norm` as an argument, as it should normally. Then the to/from methods should simplify a bit.

AtreyeeS:
Thanks @registerrier - done!

gammapy/modeling/models/spectral.py
```diff
@@ -1736,18 +1736,21 @@ def write(self, overwrite=False):
...
+            par.update(data["parameters"][idx])
+            setattr(model, p.name, Parameter(**par))
         return model
 
     def to_dict(self, full_output=False):
         """Create dict for YAML serilisation"""
         data = super().to_dict(full_output)
-        data["filename"] = self.filename
-        data["unit"] = str(self.map.unit)
+        data["spectral"]["filename"] = self.filename
```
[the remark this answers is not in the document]
AtreyeeS: because the `from_dict` is also adapted. The filename should be the spectral model filename, which was incorrect before

adonath (approved):
Thanks @AtreyeeS, looks good to me!

PR #4127, opened 2022-10-05
Correct safe mask parameters in spectral analysis with the HLI tutorial

registerrier:
**Description**

This pull request .corrects the configuration applied to the `Analysis` class in the spectral analysis tutorial. The `SafeMaskMaker` parameters were using the maximum offset, where it should be using the effective area criterion.

**Dear reviewer**


adonath (approved):
Thanks @registerrier!

PR #4133, opened 2022-10-11
Fix handling of point sky region for model contribution check

### LINKED ISSUE #4117: 3D (Fermi) + 1D (LST/MAGIC PointSkyRegion-style) analysis. Global fit ignores the 1D.

adonath:
**Description**

This pull request now fixes #4117.

### DIFF
--- a/gammapy/datasets/map.py
+++ b/gammapy/datasets/map.py
@@ -1357,11 +1357,12 @@ class MapDataset(Dataset):
 
         counts = 0
         background, excess, sqrt_ts = np.nan, np.nan, np.nan
+
         if self.counts:
-            summed_stat = self._counts_statistic[mask].sum()
-            counts = summed_stat.n_on
+            counts = self.counts.data[mask].sum()
 
             if self.background:
+                summed_stat = self._counts_statistic[mask].sum()
                 background = summed_stat.n_bkg
                 excess = summed_stat.n_sig
                 sqrt_ts = summed_stat.sqrt_ts
--- a/gammapy/makers/tests/test_reduce.py
+++ b/gammapy/makers/tests/test_reduce.py
@@ -307,7 +307,7 @@ def test_dataset_maker_spectrum_global_rad_max():
     assert counts.unit == ""
     assert counts_off.unit == ""
     assert_allclose(counts.data.sum(), 438, rtol=1e-5)
-    assert_allclose(counts_off.data.sum(), 273, rtol=1e-5)
+    assert_allclose(counts_off.data.sum(), 276, rtol=1e-5)
 
 
 @requires_data()
--- a/gammapy/maps/region/geom.py
+++ b/gammapy/maps/region/geom.py
@@ -17,7 +17,6 @@ from regions import (
     PixCoord,
     PointSkyRegion,
     RectanglePixelRegion,
-    RectangleSkyRegion,
     Regions,
     SkyRegion,
 )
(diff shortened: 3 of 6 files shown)


registerrier (approved):
Thanks @adonath !

I think this is the right behavior that we should support. 
For a spectrum extracted with pointlike IRFs, all the spatial model is already in the IRF. We should not try to estimate whether a model contributes or not. Because we have no way to estimate the actual `npred` of a source that is not a point source centered on the point region. 
So in the end it is the user's responsibility to apply the right model to a dataset extracted in pointlike mode.

adonath:
Thanks @registerrier!

PR #4137, opened 2022-10-13
Implement minimal release pipeline

adonath:
**Description**

This pull request implements a minimal release pipeline, which includes the following steps:

- Minimal  `prepare-release.py` script which will handle updates to the code meta data
- Upload to PyPi
- Create a GitHub release, which then in turn triggers the Zenodo upload
- Emit the dispatch to the `gammapy-docs` and `gammapy-webpage` repo.

### DIFF
--- a/docs/development/release.rst
+++ b/docs/development/release.rst
@@ -6,74 +6,58 @@
 How to make a Gammapy release
 *****************************
 
-This page contains step-by-step instructions how to make a Gammapy release.
-
-Overview
---------
-
-We have structured the procedure in three phases:
-
-#. "Pre release" -- on a day (or several days) before making the release
-#. "Make release" -- on the day of making the release
-   (tag the stable version and make source release on PyPI)
-#. "Post release" -- on the day when announcing the release
-   (ideally only two or three days after making the release)
-
-The purpose of writing the procedure down explicitly is to make it easy for
-anyone to make a release (as opposed to one or a few people with secret
-knowledge how to do it). It's also good to not have to remember this stuff, and
-to avoid errors or forgotten steps.
-
-Note that we currently don't do bugfix releases, i.e. making a release is always simple in the sense
-that you just need to create a branch off of ``master``.
-Making bugfix releases would be more difficult and involve identifying commits with bug fixes and
-backporting those to the stable branches.
-
-In these notes we'll use the Gammapy 0.14 release as an example.
-
-Pre release
------------
-
-Steps to prepare for the release (e.g. a week before) to check that things are in order:
-
-#. Try to get developers to finish up their PRs, try to help fix bugs, and postpone non-critical issues to the next release.
-#. Do these extra checks and clean up any warnings / errors that come up::
-
-       make polish
-       make pylint
-       make flake8
-
-#. Check external HTML links from the docs (see :ref:`here <dev-check_html_links>`).
-#. Check that the GitHub actions CI build is working.
-
-   Links are at https://github.com/gammapy/gammapy#status-shields
-#. Check that the release notes are complete, by going through the list of GitHub issues for the
-   release milestone.
-#. Check the author list in CITATION.cff by running the ``dev/authors.py`` script. Manually
-   fi any mismatches and finally sort by commits.
+This page contains step-by-step instructions how to make a Gammapy release. We mostly follow the
+`Astropy release instructions <https://docs.astropy.org/en/latest/development/releasing.html>`__
+and just lists the additional required steps.
 
 
-Make release
-------------
+Feature Freeze and Branching
+----------------------------
 
-Steps for the day of the release:
+#. Follow the `Astropy feature freeze and branching instructions <https://docs.astropy.org/en/latest/development/releasing.html#start-of-a-new-release-cycle-feature-freeze-and-branching>`__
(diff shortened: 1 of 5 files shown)


docs/development/release.rst
```diff
@@ -6,74 +6,58 @@
...
-Note that we currently don't do bugfix releases, i.e. making a release is always simple in the sense
-that you just need to create a branch off of ``master``.
-Making bugfix releases would be more difficult and involve identifying commits with bug fixes and
-backporting those to the stable branches.
+Releasing the first major release candidate
+-------------------------------------------
 
-In these notes we'll use the Gammapy 0.14 release as an example.
+#. Add an entry for the release candidate like `v1.0rc1` or `v1.1rc1` in the ``download/index.json`` file in the `gammapy-web repo <https://github.com/gammapy/gammapy-webpage>`__.
+   The ``datasets`` entry should point to dev version of `gammapy-data`. This allows to fix bugs in the data during the release candidate testing.
```
registerrier, line 25: mention that notebooks can be left to the same value as dev?

docs/development/release.rst
```diff
@@ -6,74 +6,58 @@
...
 
-Pre release
------------
+#. Fill the changelog ``docs/release-notes/<version>.rst`` for the version you are about to release.
+#. Update the ``CITATION.cff`` by running the ``dev/prepare-release.py`` script.
+#. Open a PR including both changes and mark it with the `backport-v<version>.x` label.
+   Gather feedback from the Gammapy user and dev community and finally merge and backport to the `v<version>.x` branch.
```
registerrier, line 30: This should be after the rc is released no? Why is user feedback for here?

docs/development/release.rst
```diff
@@ -6,74 +6,58 @@
...
-   * Follow the stable documentation procedure.
-   * Edit ``stable/index.html`` to point to ``0.14/index.html``.
+   * Wait for the triggered docs build to finish.
    * Edit ``stable/switcher.json`` to add the new version.
 
 #. In the `gammapy-web repo <https://github.com/gammapy/gammapy-webpage>`__:
 
    * Mention the release on the front page and on the news page.
-   * In the ``download/install`` folder, copy ``gammapy-0.13-environment.yml`` file as ``gammapy-0.14-environment.yml``.
+   * In the ``download/install`` folder, copy ``gammapy-1.0-environment.yml`` file as ``gammapy-1.0-environment.yml``.
```
registerrier, line 56: Maybe it should say when to create the file

docs/development/release.rst
```diff
@@ -107,4 +91,15 @@ Steps for the day to announce the release:
...
    coming weeks and months).
-#. Start working on the next release. :-)
+#. Start working on the next release.
+
+
+Make a Bugfix release
+---------------------
+
+#. Add an entry for the bug-fix release like `v1.0.1` or `v1.1.2` in the ``download/index.json`` file in the `gammapy-web repo <https://github.com/gammapy/gammapy-webpage>`__.
+   The ``datasets`` entry should point to last stabel version, like `v1.0` or `v1.1`. We do not provide buf-fix release for data.
```
registerrier, line 101: stable

registerrier: bug-fix

registerrier (commented):
Thanks @adonath

bkhelifi:
The script `codemeta.py` as it exists do an update of an existing file, and not create it... This is a workflow to be defined to be sure that this script does what we want. Note that an open PR about the content of this metadata file is open, #4078. So, do we plan to run this script or to use this PR? This is not clear to me.

If this script aims to create a codemeta from scratch, one can use a template with basic information. Such template can be found [here](https://github.com/bkhelifi/gammapy/blob/Authors/dev/FOOTER_CODEMETA.json), with up-to-date info. 

There is a least one thing that is missing: the update of the dependencies ("softwareRequirements") and the running platform ("runtimePlatform"), that might change from one release to an other one. A possible implementation can be found [here](https://github.com/bkhelifi/gammapy/blob/5d505b6fa882f236d5ddca66b11f869a9eb6aea7/dev/authors_2.py#L330).

adonath:
Thanks @bkhelifi! The release script in this PR is totally incomplete and that's fine as we still can update manually for v1.0. My goal was just to at least keep the functionality proposed in #3954. For v1.0 we will not get a complete automatic generation of the meta data (or maybe we do?), but we could at least work towards it. At any time you can propose the changes to the `prepare-release.py`, including author list etc.

This PR is just for the release candidate, which will not be uploaded to Zenodo or any other archive. In that sense the content is not as important yet, but we should aim to make it more complete towards the final release!

docs/development/release.rst
adonath: Done

docs/development/release.rst
adonath: Done

docs/development/release.rst
adonath: Yes, at this stage it could be omitted, but it might still be reasonable to have some time for other people to look at the proposed changelog, because we do not auto-generate it. So the content is flexible and can be discussed...

adonath: The same goes for the changes to `CITATION.cff` in case contributors would still like to join the auto-generated author list. I think the review should happen before the first release candidate, because otherwise we will always do `rc2`?

registerrier (approved):

PR #4138, opened 2022-10-13
Correct bug with ReflectedRegionFinder and PointSkyRegion

registerrier:
**Description**

This pull request corrects a bug that was not seen by tests. The `ReflectedRegionFinder` should fail if the region is `PointSkyRegion`. The error raised previously happened not to be raised at the right place. It appeared in the CI of #4133 
This PR corrects for this by explicitly testing if the input region is `PointSkyRegion` and raises a `TypeError` if it is the case.

**Dear reviewer**

### DIFF
--- a/gammapy/makers/background/reflected.py
+++ b/gammapy/makers/background/reflected.py
@@ -384,7 +384,9 @@ class ReflectedRegionsFinder(RegionsFinder):
             WCS for the determined regions
         """
         if isinstance(region, PointSkyRegion):
-            raise TypeError("ReflectedRegionFinder does not work with PointSkyRegion.")
+            raise TypeError(
+                "ReflectedRegionFinder does not work with PointSkyRegion. Use WobbleRegionFinder instead."
+            )
 
         regions = []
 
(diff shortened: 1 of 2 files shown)


bkhelifi (approved):
Thanks!!!

adonath (commented):
Thanks @registerrier! Just for completeness: 

- Technically the `ReflectedRegionFinder` works fine with a `PointSkyRegion`, however it is very sensitive to the `angle_increment` parameter. So it would basically try to place point source regions at every pixel in the tangential map. Finally it would fail because regions are overlapping.
- So I agree it is safer to just forbid it. However we should make users aware to use the `WobbleRegionFInder` instead

gammapy/makers/background/reflected.py
```diff
@@ -383,6 +383,9 @@ def run(self, region, center, exclusion_mask=None):
         wcs: `~astropy.wcs.WCS`
             WCS for the determined regions
         """
+        if isinstance(region, PointSkyRegion):
+            raise TypeError("ReflectedRegionFinder does not work with PointSkyRegion.")
```
adonath, line 387: Please add a comment to use the `WobbleRegionFinder` instead...

Issue #4149, opened 2022-10-20
Covariance matrix is not properly set on Datasets


registerrier:
**Gammapy version**
from at least v0.17 to 1.0rc1

**Bug description**
The covariance matrix is not properly set on `Datasets` when `Models` contain more than one model. Only the sub-covariances of individual models are properly set.  No positive covariance terms between independent model components are set. Thanks @AtreyeeS  for noting this.

**Expected behavior**
Correlations between independent model parameters should be non-zero.

**To Reproduce**
```
import numpy as np
from gammapy.datasets import Datasets
from gammapy.modeling import Covariance

ds = Datasets.read("$GAMMAPY_DATA/fermi-3fhl-crab/Fermi-LAT-3FHL_datasets.yaml",
              filename_models= "$GAMMAPY_DATA/fermi-3fhl-crab/Fermi-LAT-3FHL_models.yaml")

factor_matrix = np.ones((9,9))
ds.models.covariance = Covariance.from_factor_matrix(
        parameters=ds.parameters, matrix=factor_matrix
    )
ds.models.covariance.plot_correlation()
```
yields:
![download-1](https://user-images.githubusercontent.com/5047091/197058054-5829bd1e-c00c-45cf-9018-18de758d4cb2.png)

while:
```
ds[0].models.covariance = Covariance.from_factor_matrix(
        parameters=ds[0].models.parameters, matrix=factor_matrix
    )
ds[0].models.covariance.plot_correlation()
```
yields:
![download-2](https://user-images.githubusercontent.com/5047091/197058469-ebec0ccd-ca05-47c6-afb7-be4efcc3647a.png)

But `ds.models.covariance.plot_correlation()` still plots the incorrect matrix.

adonath:
Thanks @registerrier and @AtreyeeS! I think this is actually expected behavior and there is not really a solution to this until we completely detach the models from the datasets. Look at the following code:

```python3
import numpy as np
from gammapy.datasets import Datasets
from gammapy.modeling import Covariance
from gammapy.modeling.models import Models

ds = Datasets.read("$GAMMAPY_DATA/fermi-3fhl-crab/Fermi-LAT-3FHL_datasets.yaml")

models = Models.read("$GAMMAPY_DATA/fermi-3fhl-crab/Fermi-LAT-3FHL_models.yaml")

factor_matrix = np.ones((9,9))
models.covariance = Covariance.from_factor_matrix(
        parameters=models.parameters, matrix=factor_matrix
)

[...]

PR #4152, opened 2022-10-21
Return correct covariance matrix on CovarianceResult

registerrier:
**Description**

This pull request is the result of the discussion in #4149 . It sets the correct covariance matrix on the `CovarianceResult` plus it adds the `Covariance` object on the `FitResult.models` so that all the information should be available inside it.

One should find a proper place to test this.

**Dear reviewer**

### DIFF
--- a/gammapy/modeling/fit.py
+++ b/gammapy/modeling/fit.py
@@ -169,6 +169,10 @@ class Fit:
 
         covariance_result = self.covariance(datasets=datasets)
 
+        optimize_result.models.covariance = Covariance(
+            optimize_result.models.parameters, covariance_result.matrix
+        )
+
         return FitResult(
             optimize_result=optimize_result,
             covariance_result=covariance_result,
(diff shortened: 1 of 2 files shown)


gammapy/modeling/fit.py
```diff
@@ -169,6 +169,10 @@ def run(self, datasets):
 
         covariance_result = self.covariance(datasets=datasets)
 
+        optimize_result.models.covariance = Covariance(
```
adonath, line 172: I guess we can actually do the same in `Fit.run()`, such that:
```python3
result = fit.run()
result.models
``` 
Already contains the full covariance...

registerrier:
Normally it should be already doing this.
The following gist shows it works:
https://gist.github.com/registerrier/13194716af259037267c90681c0bea00

Note that, now it is clear that the right object to serialize the `Models` and the `Covariance` is the `FitResult`, not the `Datasets`.

adonath:
Yes, you are right  @registerrier! It was even myself that added this, it seems I just forgot about it.

adonath:
> Note that, now it is clear that the right object to serialize the Models and the Covariance is the FitResult, not the Datasets.

Yes, but it has always been like this. Why would one re-write e.g. all the datasets again, which have not changed during a fit?
Maybe we show this not often enough in tutorials?

adonath (dismissed):

AtreyeeS (dismissed):
Thanks @registerrier ! This works well.
I think it would be good to modify the fitting tutorial to show `result.models.covariance.plot_correlation`, even though it will not affect the specific example because there is only model.
Also, maybe we can also add a warning or info in `models.correlation` if models is a `DatasetModels)` ?

adonath (approved):

PR #4154, opened 2022-10-25
Fix failing test because of scipy 1.9.3

registerrier:
**Description**

This pull request introduces a patch for failing test with scipy 1.9.3
The issue is that `multivariate_normal` is now failing with a matrix of rank 1.  The test is changed to have a rank 2 matrix instead. This is likely a bug in scipy.

**Dear reviewer**


adonath (approved):
Thanks @registerrier! No comments from my side...

Issue #4155, opened 2022-10-25
Plot_regions() fails for an empty model object


luca-giunti:
**Gammapy version**
1.0rc1

**Bug description**
`DatasetModels.plot_regions()` fails with `AttributeError: 'list' object has no attribute 'center` if `DatasetModels` is empty (i.e. `DatasetModels.to_regions()` is equal to `[]`)

**Expected behavior**
It should either plot nothing, or crash with a better message

registerrier:
This is now solved by #4158 . Closing

PR #4158, opened 2022-10-25
Better handlling of empty region lists in RegionGeom and DatasetModels

luca-giunti:
**Description**

This pull request fixes issue #4155 .

Now `RegionGeom.from_regions()` can accpt an empty list as an argument, and in that case `RegionGeom.plot_region()` (used inside `DatasetModels.plot_region()`) fails  with a better error message: `Models don't have any associated region.`
`


gammapy/modeling/models/core.py
```diff
@@ -996,6 +996,9 @@ def plot_regions(self, ax=None, kwargs_point=None, path_effect=None, **kwargs):
             WCS axes
         """
         regions = self.to_regions()
+        if isinstance(regions, list) and len(regions) == 0:
+            raise ValueError("Models don't have any associated region.")
```
registerrier, line 1000: Should fail an error or simply log a warning that there are no regions to plot? I think the latter is more appropriate since having no region to plot is a still a regular usage.

registerrier (changes requested):

I would simply suggest to not raise an exception if the list of region to plot is empty. This is a valid case and it should not throw an error. Adding a warning is probably a good compromise.

gammapy/modeling/models/core.py
luca-giunti: If we only raise a warning and then let the plotting function go on,it will fail with `ValueError("Region definition required.")` [here](https://github.com/gammapy/gammapy/blob/560589ab7c80a880401ccb444ecfa64bcbd6fad3/gammapy/maps/region/geom.py#L123). For this reason I raised this (more explicit) error before

luca-giunti: @adonath and @registerrier do you have further opinions about this? It is true that throwing an error here potentially breaks loops in which multiple `Models` regions are plotted, which could be annoying. But the alternative option (to plot a dummy region similarly to what happens when one calls `plt.plot()` with no data) would probably require the definition of some default WCS object

adonath: I think the error is basically the right behavior. And also the error message of `ValueError: Region definition required.` is meaningful. Defining a default WCS seems not reasonable to me, it is completely arbitrary...

I would basically keep the current behavior but maybe allow the exception of `RegionGeom.plot()` to not fail if an `ax` is passed. Because in this case one can just show the image again without any region on top...

registerrier (approved):

PR #4160, opened 2022-10-26
Better handling of regions with inhomogeneous frames in RegionGeom

luca-giunti:
**Description**

This pull request addresses the following issue: `RegionGeom.from_regions(regions=regions)` fails if `regions` contain regions with different center coordinate frames.

To reproduce: 
```
from gammapy.maps.region import RegionGeom
from regions import Regions, CircleSkyRegion
import astropy.units as u

c1 = CircleSkyRegion(SkyCoord(0, 0, unit=u.deg, frame="icrs"), radius=1*u.deg)
c2 = CircleSkyRegion(SkyCoord(0, 0, unit=u.deg, frame="galactic"), radius=1*u.deg)

regions = Regions([c1, c2])
RegionGeom.from_regions(regions=regions)
```
Which yields
```
ValueError: List of inputs don't have equivalent frames: <SkyCoord (Galactic): (l, b) in deg
    (0., 0.)> != <SkyCoord (ICRS): (ra, dec) in deg
    (0., 0.)>
```
To fix it it was sufficient to convert all region centers to the `icrs` frame in the `compound_region_center` function defined in `gammapy.utils.regions`.

### DIFF
--- a/gammapy/utils/regions.py
+++ b/gammapy/utils/regions.py
@@ -55,7 +55,7 @@ def compound_region_center(compound_region):
     if len(regions) == 1:
         return regions[0].center
 
-    positions = SkyCoord([region.center for region in regions])
+    positions = SkyCoord([region.center.icrs for region in regions])
 
     def f(x, coords):
         """Function to minimize"""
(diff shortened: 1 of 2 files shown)


luca-giunti:
I don't understand the fail in `CI / ubuntu-latest, py39-test`, since on my machine the test works. Should I decrease the `atol`?

adonath:
@luca-giunti I think increasing the test tolerance is fine here. The center is computed using numerical minimization, which might yield slightly different results on different architectures. It seems the result is still precise enough...

registerrier (approved):
Thanks @luca-giunti

Issue #4162, opened 2022-10-28
Parameters that are frozen by default are always freezed upon Datasets reading


luca-giunti:
**Gammapy version**
dev

**Bug description**
I noticed that when a `Datasets` object is serialized and then read back (`.write()`, `.read()`), the `FoVBackgroundModel` spectral tilt is frozen even if it was originally un-frozen in the written `Datasets`

![image](https://user-images.githubusercontent.com/47325742/198545472-0eb33b17-8dd7-45e3-b355-e5c47ab5ead3.png)

I verified that this is the case also for all other model parameters that are frozen by default (e.g. spectral reference energy for any spectral model).

luca-giunti:
The issue comes from `Models.to_dict()`  because the frozen key is removed from the dict if `False`, which is not correct for default values with `frozen=True`. I will add a PR for that

registerrier:
This is solved by #4163 . Closing.

PR #4163, opened 2022-10-28
Fix Models.to_dict()

luca-giunti:
**Description**

This pull request addresses #4162


registerrier (commented):
Thanks @luca-giunti .

I will merge now. It is likely that a similar issue exist with other types of template models if they have additional parameters.

PR #4164, opened 2022-10-28
Consistent Interpolation scale for bkg 2d and 3d

QRemy:
This PR change the default values scale for interpolation to log in Background2D  in order to be consistent with Background3D. Test are modified to not include zeros in the test data. Also add a test for the MapDatasetMaker with 2d bkg

### DIFF
--- a/gammapy/makers/map.py
+++ b/gammapy/makers/map.py
@@ -228,8 +228,8 @@ class MapDatasetMaker(Maker):
 
         if self.background_interp_missing_data:
             bkg.interp_missing_data(axis_name="energy")
-            
-        if self.background_pad_offset and  bkg.has_offset_axis:
+
+        if self.background_pad_offset and bkg.has_offset_axis:
             bkg = bkg.pad(1, mode="edge", axis_name="offset")
 
         return make_map_background_irf(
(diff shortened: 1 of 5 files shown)


registerrier (commented):
Thanks @QRemy 

I am not sure the padding option on the `MapDatasetMaker` object is the right way to handle this. Shouldn't this be defined on the `Background2D` IRF itself?

QRemy:
> I am not sure the padding option on the `MapDatasetMaker` object is the right way to handle this. Shouldn't this be defined on the `Background2D` IRF itself?

This way it is similar to the `background_interp_missing_data` option that is not applied by default when reading the bkg map, but only on the `MapDatasetMaker`.

gammapy/makers/map.py
```diff
@@ -35,6 +35,10 @@ class MapDatasetMaker(Maker):
     background_interp_missing_data: bool
         Interpolate missing values in background 3d map.
         Default is True, have to be set to True for CTA IRF.
+    background_pad_offset: bool
+        Pad one bin in offset for 2d background map.
+        This avoid extrapolation at edges and use the nearest value.
+        Default is True, have to be set to True for HESS IRF.
```
registerrier, line 41:
```suggestion
        Default is True.
```

registerrier (approved):
Thanks @QRemy!

PR #4172, opened 2022-11-03
Fix grid plot with label axis

adonath:
**Description**

This pull request fixes a bug that prevented a `LabelMapAxis` being plot with `plot_grid()`. I still have to add a unit test...


registerrier (approved):

Do we need a similar scheme for maps with a time axis?

adonath:
> Do we need a similar scheme for maps with a time axis?

Probably yes, it declares a different node type.

adonath:
I'll include a fix in this PR as well.

registerrier (approved):
Thanks @adonath !

Issue #4171, opened 2022-11-03
Backwards support of OBS_MODE  for GADF v0.2


luca-giunti:
**Gammapy version**
1.0rc1

**Bug description**
`obs.fixed_pointing_info.mode` crashes with `ValueError` if the pointing mode stored in the FITS data file is not equal to "WOBBLE", "POINTING" or "DRIFT". This is coded in [`PointingMode.from_gadf_string`](https://github.com/gammapy/gammapy/blob/2d23261187232ffafaff0b56ebbbe8ac5dc944bf/gammapy/data/pointing.py#L48). This in turn blocks the data reduction.

 This is not a good behavior and not even compatible with GADF, because GADF does not impose any list of acceptable `OBS_MODE` meta data (see [here](https://gamma-astro-data-formats.readthedocs.io/en/v0.2/events/events.html?#obs-mode)). 

Example (using H.E.S.S. data):
![image](https://user-images.githubusercontent.com/47325742/199734428-79516f87-efe4-4dab-bb9a-9fd20ad9d0a8.png)

adonath:
Thanks @luca-giunti! The `OBS_MODE` handling was subject to change in GADF v0.3 (which will be released soon). Here is the latest documentation: https://gamma-astro-data-formats.readthedocs.io/en/latest/events/events.html#obs-mode

As `OBS_MODE` is a mandatory keyword, which is used in the data reduction it should fail for invalid values. I don't see any incorrect behavior in Gammapy here...

registerrier:
> As OBS_MODE is a mandatory keyword, which is used in the data reduction it should fail for invalid values. I don't see any incorrect behavior in Gammapy here...

The question is probably whether Gammapy still supports gadf v0.2 or if it supports only v0.3.
This is probably a strong statement to support only v0.3 only a few weeks after it was released.

luca-giunti:
Thank you @adonath for the reference to the new GADF format. However I still find this problematic, since the current behavior assumes that all FITS DL3 data comply with the latest GADF, which is not necessarily true (yet). The prod I am using, for example, is a standard H.E.S.S. prod with `OBS_MODE=ON`. So it should not crash. At most, raise a warning

adonath:
Yes, I see. I checked and it seems indeed in v0.2 basically arbitrary values are allowed (which is very weird BTW...). This means from the Gammapy side we have to check the GADF version?

registerrier:
[...]

PR #4173, opened 2022-11-03
Introduce gadf version test for OBS_MODE in FixedPointingInfo

registerrier:
**Description**

This pull request is a fix for #4171 .  The current code makes it impossible to perform an analysis with gadf data v0.2 with values of `OBS_MODE` different from `POINTING` and `WOBBLE`.
In v0.2, the list of optional values is not  given and the keyword is not mandatory.

It introduces a gadf version test in `PointingMode.from_gadf_string`. 
The behavior is the following:
if v0.3, only allow `POINTING` and `DRIFT`, as the keyword becomes mandatory.
if v0.2, always assume `POINTING` as this not a mandatory information and it is explicitly written that this is not used for data reduction.

Another option could be not to use the `PointingMode` in the `MapDatasetMaker` or at least only use it for `DRIFT` mode data.

**Dear reviewer**

A side question is whether

### DIFF
--- a/gammapy/data/pointing.py
+++ b/gammapy/data/pointing.py
@@ -30,31 +30,34 @@ class PointingMode(Enum):
 
     For ground-based instruments, the most common options will be:
     * POINTING: The telescope observes a fixed position in the ICRS frame
-    * DRIFT: The telscope observes a fixed position in the AltAz frame
+    * DRIFT: The telescope observes a fixed position in the AltAz frame
 
     OGIP also defines RASTER, SLEW and SCAN. These cannot be treated using
     a fixed pointing position in either frame, so they would require the
     pointing table, which is at the moment not supported by gammapy.
 
-    The H.E.S.S. data releases uses the not-defined value "WOBBLE",
-    which we assume to be the same as "POINTING", making the assumption
-    that one observation only contains a single wobble position.
+    Data releases based on gadf v0.2 do not have consistent OBS_MODE keyword
+    e.g. the H.E.S.S. data releases uses the not-defined value "WOBBLE".
+    For all gadf v0.2 data, we assume OBS_MODE to be the same as "POINTING",
+    making the assumption that one observation only contains a single wobble position.
     """
 
     POINTING = auto()
     DRIFT = auto()
 
     @staticmethod
-    def from_gadf_string(val):
-        # OBS_MODE is not well-defined in GADF 0.2, HESS and MAGIC
-        # filled some variation of "WOBBLE" for the open crab paper
-        if val.upper() in ("POINTING", "WOBBLE"):
+    def from_gadf_string(val, version="0.3"):
+        # OBS_MODE is not well-defined not mandatory in GADF 0.2
+        # We always assume that they are pointing observations
+        if version == "0.3":
+            if val.upper() == "POINTING":
+                return PointingMode.POINTING
+            if val.upper() == "DRIFT":
+                return PointingMode.DRIFT
+
+            raise ValueError(f"Unsupported pointing mode: {val}")
+        else:
             return PointingMode.POINTING
-
-        if val.upper() == "DRIFT":
-            return PointingMode.DRIFT
-
-        raise ValueError(f"Unsupported pointing mode: {val}")
 
 
 class FixedPointingInfo:
@@ -103,7 +106,7 @@ class FixedPointingInfo:
     """
 
     def __init__(self, meta):
-        self.meta = meta
+        self.meta = meta.copy()
 
     @classmethod
     def read(cls, filename, hdu="EVENTS"):
@@ -129,9 +132,8 @@ class FixedPointingInfo:
     def mode(self):
         """See `PointingMode`, if not present, assume POINTING"""
         obs_mode = self.meta.get("OBS_MODE")
-        if obs_mode is None:
-            return PointingMode.POINTING
-        return PointingMode.from_gadf_string(obs_mode)
+        gadf_version = self.meta.get("HDUVERS")
+        return PointingMode.from_gadf_string(obs_mode, gadf_version)
 
     @lazyproperty
     def location(self):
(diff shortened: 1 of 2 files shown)


registerrier:
I have finally reverted the changes. In the current state, it is not possible to properly test for the gadf version:
- to read the gadf version on the `FixedPointingInfo`  you need to have the `HDUVERS` stored  in the meta  that is used to build it. But, these keywords are removed from the `meta` when `Observation`  is read from the `DataStore`  (but not when using `Observation.read`... ).
- There is no gadf version defined when events are simulated nor an OBS_MODE  defined. In these situations, it is necessary to handle `OBS_MODE==None`  in a specific manner by assuming those are pointing observations.
- the HAWC data in gammapy-data use a `HDUVERS==dev1bf00373`, so we would need another case to handle these.

Finally, it seems to me that, for now we have to avoid giving `OBS_MODE` any practical functionality except for the case of `DRIFT`, and that the general assumption so far is that we have only fixed pointing observations.
The behaviour implemented is therefore to always set `PointingMode.POINTING`  except when `OBS_MODE=="DRIFT"`, independently of the gadf version.

bkhelifi (approved):
I think that this behaviour is what is needed to be backward compatible with all data releases.

As written here, the function ``from_gadf_string ``has lost its link with GADF (and its different versions). Also, when we will have other formats (CTAO, VODF, ...), this function name will be a problem. But this should be treated for the v1.1 (and not here).

maxnoe:
> There is no gadf version defined when events are simulated nor an OBS_MODE defined. In these situations, it is necessary to handle OBS_MODE==None in a specific manner by assuming those are pointing observations.

That just means that the simulation has to set the pointing mode. Right now I assume gammapy only simulates `POINTING` observations, so the simulation should just set it. The whole point of this function was to disentangle the data format of the file from gammapy's internal one. So anything that starts internally in gammapy needs to set this.

This for me means that like we did for other values, the correct solution would be to have mode be set on init of `FixedPointingInfo` not in a property relying on meta.
I see that `FixedPointingInfo` still only stores the `meta` dict. So this should be changed to explicitly store the members in the internal representation.

> the HAWC data in gammapy-data use a HDUVERS==dev1bf00373, so we would need another case to handle these.

You want to support data that was made with an intermediate dev release indefinitely? That's not a good idea I think, it will increse the maintenance burden quite a bit. Can we not ask HAWC to update this to 0.3 once it is out? I am in contact with zenodo support to figure out the upload, in principal we can already release it and can try to figure out the zenodo deployment later...

gammapy/data/pointing.py
```diff
@@ -30,31 +30,31 @@ class PointingMode(Enum):
...
-        # OBS_MODE is not well-defined in GADF 0.2, HESS and MAGIC
-        # filled some variation of "WOBBLE" for the open crab paper
-        if val.upper() in ("POINTING", "WOBBLE"):
+        # OBS_MODE is not well-defined not mandatory in GADF 0.2
+        # We always assume that they are pointing observations
+        if val.upper() == "POINTING":
             return PointingMode.POINTING
-
-        if val.upper() == "DRIFT":
+        elif val.upper() == "DRIFT":
```
maxnoe, line 54: If this is the solution we want to use right now, the logic can be simplified to just:

```
if val.upper() == "DRIF":
    return PointingMode.DRIFT

return PointingMode.POINTING
```

gammapy/data/pointing.py
```diff
@@ -30,31 +30,31 @@ class PointingMode(Enum):
...
     a fixed pointing position in either frame, so they would require the
     pointing table, which is at the moment not supported by gammapy.
 
-    The H.E.S.S. data releases uses the not-defined value "WOBBLE",
-    which we assume to be the same as "POINTING", making the assumption
-    that one observation only contains a single wobble position.
+    Data releases based on gadf v0.2 do not have consistent OBS_MODE keyword
+    e.g. the H.E.S.S. data releases uses the not-defined value "WOBBLE".
+    For all gadf v0.2 data, we assume OBS_MODE to be the same as "POINTING",
+    making the assumption that one observation only contains a single wobble position.
```
maxnoe, line 42: Maybe add that this is because gammapy anyways only supports a single fixed pointing position over the whole observation (either in equatorial or horizontal coordinates) at this moment?

adonath:
> You want to support data that was made with an intermediate dev release indefinitely? That's not a good idea I think, it will increse the maintenance burden quite a bit. Can we not ask HAWC to update this to 0.3 once it is out? 

I'm pretty sure this was @LauraOlivera' s plan anyway...

gammapy/data/pointing.py
```diff
@@ -32,26 +32,28 @@ class PointingMode(Enum):
     * POINTING: The telescope observes a fixed position in the ICRS frame
     * DRIFT: The telescope observes a fixed position in the AltAz frame
 
+    Gammapy only supports only fixed pointing positions over the whole observation
```
maxnoe, line 35: doubled only

gammapy/data/pointing.py
```diff
@@ -32,26 +32,28 @@ class PointingMode(Enum):
...
 
     @staticmethod
     def from_gadf_string(val):
-        # OBS_MODE is not well-defined not mandatory in GADF 0.2
-        # We always assume that they are pointing observations
-        if val.upper() == "POINTING":
-            return PointingMode.POINTING
-        elif val.upper() == "DRIFT":
+        # OBS_MODE is not well-defined and not mandatory in GADF 0.2
+        # We always assume that they observations are pointing observations
```
maxnoe, line 54: the observations

LauraOlivera:
> Can we not ask HAWC to update this to 0.3 once it is out?

yes

registerrier:
> This for me means that like we did for other values, the correct solution would be to have mode be set on init of FixedPointingInfo not in a property relying on meta.
I see that FixedPointingInfo still only stores the meta dict. So this should be changed to explicitly store the members in the internal representation.

Yes this we definitely have to do, in v1.1.

registerrier:
CI is green. Merging now.

Issue #4143, opened 2022-10-18
Fix tutorials printing/plotting


AtreyeeS:
This is a reminder issue to continue the cleanup implemented in #4131 and #4132 .
The relevant changes were implemented in `getting started` and `data exploration`, the other sections should be cleaned up before the final 1.0 release

AtreyeeS:
@adonath I think it will take more time to list them rather than fix them! 
I will put in a PR next week

adonath:
That's even better! Thanks @AtreyeeS!

PR #4174, opened 2022-11-04
Cleanup tutorials

AtreyeeS:
Addresses #4143

### DIFF
--- a/examples/tutorials/analysis-1d/cta_sensitivity.py
+++ b/examples/tutorials/analysis-1d/cta_sensitivity.py
@@ -143,7 +143,7 @@ sensitivity_table = sensitivity_estimator.run(dataset_on_off)
 #
 
 # Show the results table
-sensitivity_table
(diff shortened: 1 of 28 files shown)


adonath:
Thanks @AtreyeeS, one way to let Python choose the representation is:

```python3

from IPython.display import display

from astropy.table import QTable
import astropy.units as u
import numpy as np

a = np.array([1, 4, 5], dtype=np.int32)
b = [2.0, 5.0, 8.5]
c = ['x', 'y', 'z']
d = [10, 20, 30] * u.m / u.s

t = QTable([a, b, c, d],
           names=('a', 'b', 'c', 'd'),
           meta={'name': 'first table'})

display(t)
```
This will choose the print in scripts and html in notebooks...however it will be some work to adapt this. From my side I think the simple printing is still OK.

registerrier (approved):
Thanks @AtreyeeS . This is great!

PR #4182, opened 2022-11-05
Adapt tutorial figure handling

adonath:
**Description**

This pull request is a follow up to #4174. It adapts the following:

- Do not use `plt.figure()` where figures are created internally
- Use `display()` to show tables. This defaults to prints for scripts and html for notebooks (however I noticed it does not seem to work with sphinx gallery....)
- Call `plt.show()` at the end of tutorials to make sure plots are shown when executed as script.

ToDo:
- [x] Adapt `data` tutorials
- [x] Adapt `api` tutorials

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -315,7 +315,7 @@ sphinx_gallery_conf = {
     "reset_modules": ("matplotlib",),
     "within_subsection_order": FileNameSortKey,
     "download_all_examples": True,
-    "capture_repr": (),
+    "capture_repr": ("_repr_html_", "__repr__"),
     "nested_sections": False,
     "min_reported_time": 10,
     "show_memory": False,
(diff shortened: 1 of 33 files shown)


registerrier:
data and API tutorials have been adapted along the same lines.
conf.py has been modified but the `display` function does not yield better output.

registerrier:
Also is there a simple way to split code cells to avoid having three consecutive `print` that are display all at once. See e.g.
https://github.com/gammapy/gammapy/blob/89b9fc79727e3ef5323fffed8e8d1f19a97338fc/examples/tutorials/api/models.py#L426

adonath:
@registerrier I think you can use:
```python3
# %%
msg = "Hello World"
print(msg)

# %%
msg = "Hello again"
print(msg)

```

registerrier:
Thanks @adonath , it works. It looks much better like this

registerrier (approved):

PR #4196, opened 2022-11-09
Correct documentation part 3

registerrier:
**Description**

This pull request corrects remaining issues and changes thumbnails in the gallery

**Dear reviewer**

PR #4198, opened 2022-11-09
Fix broken link

AtreyeeS:


registerrier:
Thanks @AtreyeeS . I think this is already corrected in #4196  :)

AtreyeeS:
Oh sorry I did not notice you had a PR open. I was just going through the dev docs and noticed it

PR #4205, opened 2022-11-10
Fix dev docs formatting

adonath:
**Description**

This pull request includes the following changes:
- fix some formatting in the developer docs
- fix links in the `CONTRIBUTING.md`
- add a DCO check CI

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,6 +2,10 @@ name: CI
 on: [push, pull_request]
 
 jobs:
+  dco-check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: tisonkun/actions-dco@v1.1
   lint:
     name: Black check
     runs-on: ubuntu-latest
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,7 +1,7 @@
 Contributing to Gammapy
 =======================
 
-Note: This text is freely inspired by the `Contribution.md <https://raw.githubusercontent.com/astropy/astropy/main/CONTRIBUTING.md>`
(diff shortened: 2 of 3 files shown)


bkhelifi:
Hi @adonath,
Thanks a lot!
Are you sure about the DCO CI? I have seen within the code of some of them that ALL commits should be signed-off, which is a bit constraining.In absolute, only one is needed. 
From what I can see in the lines 8541-8577 of https://github.com/tisonkun/actions-dco/blob/main/dist/index.js, it seems to be the case.
But maybe a check is better, e.g. on your test project you used for testing the PR with our new branching scheme.

bkhelifi (dismissed):
Thanks Axel!
I have no comment, except the general one about the DCO CI.

adonath:
@bkhelifi  and @registerrier I'm proposing a pre-commit hook to handle the sign-off in this PR. Developers can install it using:

```
pre-commit install --hook-type prepare-commit-msg
```

I'm not sure it is a good idea, but it works and by explicitly installing, they would agree to the DCO one time.

registerrier (approved):

We should first consider the DCO an allowed fail.  And we will see how this works.

Maybe a suggestion, the pre-commit hook script could explicitly ask the user to agree to the DCO for future commits to the gammapy repo?

adonath:
> Maybe a suggestion, the pre-commit hook script could explicitly ask the user to agree to the DCO for future commits to the gammapy repo?

I agree this is a good idea, however if this was added to the script users would be asked every time. We would need to find a way to enable this for the installation only...

registerrier:
> > Maybe a suggestion, the pre-commit hook script could explicitly ask the user to agree to the DCO for future commits to the gammapy repo?
> 
> I agree this is a good idea, however if this was added to the script users would be asked every time. We would need to find a way to enable this for the installation only...

Maybe if we add a script that prints a message asking if the user agrees and then performs the `pre-commit install`.

PR #4208, opened 2022-11-14
Add PiecewiseNormSpatialModel

QRemy:
Add `PiecewiseNormSpatialModel` and support for spatial model corrections in `FoVbackgroundModel` and `TemplateNPredModel`.


adonath:
Thanks a lot @QRemy, this is a very interesting feature! Just two superficial comments:

- Strictly speaking we do not have `NormSpatialModels`, because the spatial model are always normed to the integral. However the different abstraction here is, that this is not normed to the integral but just applies a multiplicative factor. So for consistency with `NormSpectralModel` this is probably fine...but we should think about once whether the abstraction makes sense...or whether we need something else.
- Would you mind to split the PR in introducing the `FoVBackgroundModel.spatial_model` independently? I think the first spatial correction model we introduce should be maybe just a simple gradient for testing...

QRemy:
> * Would you mind to split the PR in introducing the `FoVBackgroundModel.spatial_model` independently? I think the first spatial correction model we introduce should be maybe just a simple gradient for testing...

Done in #4209, I'll update this PR once the other one will be merged.

gammapy/modeling/models/spatial.py
```diff
@@ -1282,3 +1286,129 @@ def plot_interactive(self, ax=None, geom=None, **kwargs):
...
+    def evaluate(self, lon, lat, energy=None, **norms):
+        """Evaluate the model at given coordinates."""
+        scale = interpolation_scale(scale=self._interp)
+        v_nodes = scale(self.norms)
+        coords = [value.value for value in self.coords._data.values()]
+        # TODO: apply axes scaling in this loop
+        coords = list(zip(*coords))
+        if self.is_energy_dependent:
+            if energy is None:
+                raise ValueError("Missing nergy value for  energy-dependent model")
```
registerrier, line 1358:
```suggestion
                raise ValueError("Missing energy values for  energy-dependent model")
```

gammapy/modeling/models/spatial.py
```diff
@@ -1282,3 +1286,129 @@ def plot_interactive(self, ax=None, geom=None, **kwargs):
...
+            interpolated = griddata(
+                coords, v_nodes, (lon, lat, energy), method="linear"
+            )
+            scaled_norms = scale.inverse(interpolated)
+        else:
+            interp = CloughTocher2DInterpolator(coords, v_nodes)
```
registerrier, line 1364: why the different interpolation scheme here?

gammapy/modeling/models/spatial.py
```diff
@@ -1282,3 +1286,129 @@ def plot_interactive(self, ax=None, geom=None, **kwargs):
...
+            scaled_norms = scale.inverse(interp(lon, lat))
+        return scaled_norms
+
+    def evaluate_geom(self, geom):
+        """Evaluate model on `~gammapy.maps.Geom`
+
+        Parameters
+        ----------
+        geom : `~gammapy.maps.WcsGeom`
+
```
registerrier, line 1374: add brief description.

gammapy/modeling/models/spatial.py
```diff
@@ -1282,3 +1286,129 @@ def plot_interactive(self, ax=None, geom=None, **kwargs):
...
+        Returns
+        -------
+        `~gammapy.maps.Map`
+
```
registerrier, line 1378: same here

gammapy/modeling/models/spatial.py
```diff
@@ -1282,3 +1286,129 @@ def plot_interactive(self, ax=None, geom=None, **kwargs):
...
+        """
+        coords = geom.get_coord(frame=self.frame, sparse=True)
+
+        if self.is_energy_dependent:
+            return self(*coords)
```
registerrier, line 1383: I would specify the expected coords here.

gammapy/modeling/models/spatial.py
```diff
@@ -1282,3 +1286,129 @@ def plot_interactive(self, ax=None, geom=None, **kwargs):
...
+        else:
+            return self(coords.lon, coords.lat)
```
registerrier, line 1385: It might be more efficient to compute it only on 2D and then resize?

gammapy/modeling/models/tests/test_cube.py
```diff
@@ -783,3 +784,35 @@ def test_spatial_model_io_background(background):
...
     assert isinstance(new_model.spatial_model, ConstantSpatialModel)
+
+
+def test_piecewise_spatial_model_background(background):
+
+    geom = background.geom
+    coords = geom.get_coord().flat
+
+    spatial_model = PiecewiseNormSpatialModel(coords, frame="galactic")
+    reference_npred = TemplateNPredModel(background, spatial_model=None).evaluate()
```
registerrier, line 795: `background.data` is enough no?

gammapy/modeling/models/tests/test_cube.py
```diff
@@ -783,3 +784,35 @@ def test_spatial_model_io_background(background):
...
+    identical_npred = TemplateNPredModel(
```
registerrier, line 796: You might use a `FoVBackgroundModel` here no?

registerrier (commented):
Thanks @QRemy . Here are some comments.

gammapy/modeling/models/spatial.py
QRemy: CloughTocher2DInterpolator is equivalent to use griddata with method="cubic",  by default I would rely on this because it has some nice properties (C1 smooth, curvature-minimization). I simplified the code but left a comment about this.

gammapy/modeling/models/spatial.py
```diff
@@ -1282,3 +1286,129 @@ def plot_interactive(self, ax=None, geom=None, **kwargs):
         if geom is None:
             geom = self.map.geom
         super().plot_interactive(ax=ax, geom=geom, **kwargs)
+
+
+class PiecewiseNormSpatialModel(SpatialModel):
+    """Piecewise spatial correction
+       with a free normalization at each fixed nodes.
+
+       For more information see :ref:`piecewise-norm-spectral`.
```
adonath, line 1295: This reference does not exist yet? Add in follow up PR?

adonath (commented):
Thanks @QRemy, I have left some comments to clarify.

gammapy/modeling/models/tests/test_cube.py
QRemy: yes done

gammapy/modeling/models/tests/test_cube.py
QRemy: the equivalent test for the `FoVBackgroundModel` is done below.

gammapy/modeling/models/spatial.py
```diff
@@ -1282,3 +1286,129 @@ def plot_interactive(self, ax=None, geom=None, **kwargs):
...
+            interpolated = griddata(
+                coords, v_nodes, (lon, lat, energy), method="linear"
+            )
+            scaled_norms = scale.inverse(interpolated)
+        else:
+            interp = CloughTocher2DInterpolator(coords, v_nodes)
+            scaled_norms = scale.inverse(interp(lon, lat))
+        return scaled_norms
+
+    def evaluate_geom(self, geom):
```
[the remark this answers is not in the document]
adonath: Ah, I see the model is meant to be used with `FoVBackgroundModel` and e.g. `TemplateSpatialModel`, but right now we do not support multiplication for spatial models, so does the latter work at all? So we should probably just require an "energy" coordinate on init?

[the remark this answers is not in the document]
QRemy: I did not try but it may even be used instead of the `TemplateSpatialModel` to fit a complex source morphology directly.

gammapy/modeling/models/spatial.py
QRemy: added

gammapy/modeling/models/spatial.py
QRemy: maybe I could check that in a separate PR as it may also be true for all spatial models.

adonath:
My proposal would be to remove the energy dependence for now, just to avoid the question of how to document and advertise this to users. We should probably  get some experience with this first.

adonath (dismissed):

adonath:
Please do a 2nd review @registerrier or @AtreyeeS!

registerrier (approved):
Thanks @QRemy . No more comments from my side.

PR #4209, opened 2022-11-15
Add support for spatial model correction on background models

QRemy:
Follow-up PR from #4208.
Allow to apply `spatial_model` corrections to `FoVBackgroundModel` and `TemplateNPredModel`.
It can possibly work with all the spatial model  already implemented, even is it was designed for the `PiecewiseNormModel` introduced in #4208.

### DIFF
--- a/gammapy/modeling/models/cube.py
+++ b/gammapy/modeling/models/cube.py
@@ -595,7 +595,7 @@ class FoVBackgroundModel(ModelBase):
 
     tag = ["FoVBackgroundModel", "fov-bkg"]
 
-    def __init__(self, spectral_model=None, dataset_name=None):
+    def __init__(self, spatial_model=None, spectral_model=None, dataset_name=None):
         if dataset_name is None:
             raise ValueError("Dataset name a is required argument")
 
@@ -607,6 +607,7 @@ class FoVBackgroundModel(ModelBase):
         if not spectral_model.is_norm_spectral_model:
             raise ValueError("A norm spectral model is required.")
 
+        self._spectral_model = None
         self._spectral_model = spectral_model
         super().__init__()
 
@@ -619,6 +620,11 @@ class FoVBackgroundModel(ModelBase):
     def spectral_model(self):
         """Spectral norm model"""
         return self._spectral_model
+
+    @property
+    def spatial_model(self):
+        """Spatial norm model"""
+        return self._spatial_model
 
     @property
     def name(self):
@@ -651,11 +657,16 @@ class FoVBackgroundModel(ModelBase):
     def evaluate_geom(self, geom):
         """Evaluate map"""
         coords = geom.get_coord(sparse=True)
-        return self.evaluate(energy=coords["energy"])
+        return self.evaluate(**coords._data) 
 
-    def evaluate(self, energy):
+
+    def evaluate(self, energy, lon=None, lat=None):
         """Evaluate model"""
-        return self.spectral_model(energy)
+        value = self.spectral_model(energy) 
+        if self.spatial_model is not None and lon is not None and lat is not None:
+            return self.spatial_model(lon,lat,energy).value * value
+        else:
+            return value
 
     @copy_covariance
     def copy(self, name=None, copy_data=False, **kwargs):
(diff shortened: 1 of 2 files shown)


gammapy/modeling/models/cube.py
```diff
@@ -607,6 +607,7 @@ def __init__(self, spectral_model=None, dataset_name=None):
         if not spectral_model.is_norm_spectral_model:
             raise ValueError("A norm spectral model is required.")
 
+        self._spatial_model = None
```
bkhelifi, line 610: why here do we not have:
self._spatial_model = spatial_model?
Like for the spectral_model....

gammapy/modeling/models/cube.py
```diff
@@ -651,11 +657,19 @@ def __str__(self):
...
-        return self.evaluate(energy=coords["energy"])
+        return self.evaluate(**coords._data)
 
-    def evaluate(self, energy):
+    def evaluate(self, energy, lon=None, lat=None):
         """Evaluate model"""
-        return self.spectral_model(energy)
+        value = self.spectral_model(energy)
+        if self.spatial_model is not None and lon is not None and lat is not None:
+            if self.spatial_model.is_energy_dependent:
```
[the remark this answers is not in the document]
registerrier: Note by the way the background map itself is already an energy dependent spatial model...

gammapy/modeling/models/cube.py
```diff
@@ -859,12 +889,17 @@ def evaluate(self):
         """
         value = self.spectral_model(self.energy_center).value
         back_values = self.map.data * value
+        if self.spatial_model is not None:
+            value = self.spatial_model.evaluate_geom(self.map.geom).value
```
registerrier, line 893: In general, is it safe to use the `value`? It can silently work, where the unit should fail?

registerrier (commented):
Thanks @QRemy . 
Having the possibility to renormailze the bkg spatiallay is an important functionality.

Some questions:
- do you expect to create a `NormSpatialModel`? Should we be able to re-use spatial models?
- this is very close to `CompoundModel` for spatial models.

gammapy/modeling/models/cube.py
```diff
@@ -595,7 +595,7 @@ class FoVBackgroundModel(ModelBase):
 
     tag = ["FoVBackgroundModel", "fov-bkg"]
 
-    def __init__(self, spectral_model=None, dataset_name=None):
+    def __init__(self, spatial_model=None, spectral_model=None, dataset_name=None):
```
adonath, line 598: This change is not backwards compatible as it changes the order of the function arguments. So it should go in the last position instead. However it would be good to make the order consistent with https://docs.gammapy.org/1.0/api/gammapy.modeling.models.SkyModel.html as well. @registerrier do we have a scheme to deprecate argument order?

gammapy/modeling/models/cube.py
QRemy: We already use `value` for the `spectral_model`.
This way we can use the already existing spatial model without introducing `NormSpatialModel ` for each of them

gammapy/modeling/models/cube.py
QRemy: I can move `spatial_model` at the end and we decide later if we want the same order as the `SkyModel` and how to deal with the deprecation in that case.

gammapy/modeling/models/cube.py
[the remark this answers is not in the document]
QRemy: Yes, as for now the spectral model is always created I guess the user will have to freeze its parameters if an energy-dependent SpatialModel is used.

gammapy/modeling/models/cube.py
QRemy, line 598:
```suggestion
    def __init__(self, spectral_model=None, dataset_name=None, spatial_model=None):
```

registerrier (dismissed):
Thanks @QRemy .

See a last inline comment regarding how the model is evaluated.
Otherwise, I think the docstring should explictly say that the norm unit is dropped.

registerrier (dismissed):
Thanks @QRemy . No further comments from my side.

Issue #3967, opened 2022-05-25
Implement WcsNDMap.plot_rgb()


adonath:
For convenience it would nice to have a `WcsNDMap.plot_rgb()` and `HpxNDMap.plot_rgb()` method, which allows users to plot a map with three different energy bins as an RGB image. If it contains more or less bins it will through an error. The logic for more bins would be that users resample the energy axis and plot after...

registerrier:
@adonath Having such a function is definitely a nice to have feature. 
I am not sure it should be a method though. Getting a correct 3 color plot require significant adjustment of the color scales and ranges, the function has to be configurable for that. I think this is a case for a standalone function in `gammapy.visualization`.

luca-giunti:
A draft:
```
from astropy.visualization import make_lupton_rgb
def plot_rgb(map_, energy_edges=None, fig=None, ax=None, **kwargs):
    axis = map_.geom.axes["energy"]
    if not energy_edges:
        if len(axis.center) != 3:
            raise ValueError("Energy edges must be provided for maps with more/less than 3 energy bins")
        else:
            energy_edges = axis.edges
            
    if len(energy_edges) != 4:
        raise ValueError("Energy edges length must be equal to 4")    
    
    axis_rgb = MapAxis.from_energy_edges(
        energy_edges.value, 
        unit=energy_edges.unit,
        name="energy",
        interp=axis.interp
    )
    map_rgb = map_.resample_axis(axis_rgb)
    
    data = make_lupton_rgb(
        map_rgb.data[0] / np.nanmax(map_rgb.data[0]),
        map_rgb.data[1] / np.nanmax(map_rgb.data[1]),
        map_rgb.data[2] / np.nanmax(map_rgb.data[2]),
        **kwargs
    )
    
    if fig is None:
        fig = plt.gcf()

    ax = map_rgb._plot_default_axes(ax=ax)
        
    ax.imshow(data)

    if map_rgb.geom.is_allsky:
        ax = map_rgb._plot_format_allsky(ax)
    else:
        ax = map_rgb._plot_format(ax)
        
    ax.autoscale(enable=False)
    
    return ax
```

- Example with a flux map:
![image](https://user-images.githubusercontent.com/47325742/201727635-99a64eeb-767e-455d-9dae-72e6efddb798.png)


- Example with a counts map:
![image](https://user-images.githubusercontent.com/47325742/201727716-fc551fdb-e375-46f1-97d2-ceea9257569e.png)

PR #4210, opened 2022-11-16
Add plot_rgb() function in gammapy.visualization

luca-giunti:
**Description**

This pull request addresses #3967 

Right now, all plot customization is done via the `kwargs` passed to `astropy.visualization.make_lupton_rgb(...)`. Is that sufficient, or more degrees of freedom should be allowed?

Also, it only works for `Wcs` maps. I guess we can expect the user to convert `Hpx` maps to `Wcs` before plotting...

The docstring example produces the following plot:
image

### DIFF
--- a/gammapy/visualization/utils.py
+++ b/gammapy/visualization/utils.py
@@ -47,6 +47,7 @@ def plot_rgb(map_, energy_edges=None, ax=None, **kwargs):
 
     Examples
     --------
+    >>> from gammapy.visualization.utils import plot_rgb
     >>> from gammapy.maps import Map
     >>> import astropy.units as u
     >>> map_ = Map.read("$GAMMAPY_DATA/cta-1dc-gc/cta-1dc-gc.fits.gz")
--- a/gammapy/visualization/utils.py
+++ b/gammapy/visualization/utils.py
@@ -58,8 +58,7 @@ def plot_map_rgb(map_, ax=None, **kwargs):
     >>> plot_rgb(map_.smooth(0.08*u.deg), **kwargs)
     """
     geom = map_.geom
-    axes = geom.axes
-    if len(axes) != 1 or axes[0].nbin != 3:
+    if len(geom.axes) != 1 or geom.axes[0].nbin != 3:
         raise ValueError(
             "One non-spatial axis with exactly 3 bins is needed to plot an RGB image"
         )
--- a/gammapy/visualization/utils.py
+++ b/gammapy/visualization/utils.py
@@ -1,12 +1,11 @@
 import numpy as np
 from scipy.interpolate import CubicSpline
-import matplotlib.pyplot as plt
 from astropy.visualization import make_lupton_rgb
-from gammapy.maps import MapAxis
+import matplotlib.pyplot as plt
 
 __all__ = [
     "plot_contour_line",
-    "plot_rgb",
+    "plot_map_rgb",
     "plot_spectrum_datasets_off_regions",
     "plot_theta_squared_table",
 ]
@@ -23,11 +22,13 @@ ARTIST_TO_LINE_PROPERTIES = {
 }
 
 
-def plot_rgb(map_, energy_edges=None, ax=None, **kwargs):
+def plot_map_rgb(map_, ax=None, **kwargs):
     """
     Plot RGB image on matplotlib WCS axes.
 
-    Based on the `~astropy.visualization.make_lupton_rgb` function.
+    This function is based on the `~astropy.visualization.make_lupton_rgb` function, and it
+    assumes that the input map has exactly three z-axis bins. If that were not the case, you
+    can resample the z-axis of your map as shown in the code example below.
 
     Parameters
     ----------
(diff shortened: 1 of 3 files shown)


gammapy/visualization/utils.py
```diff
@@ -20,6 +23,69 @@
...
+        if len(axis.center) != 3:
+            raise ValueError("Exactly 3 energy bins are needed to plot an RGB image")
+        else:
+            energy_edges = axis.edges
+
+    if len(energy_edges) != 4:
+        raise ValueError("Exactly 3 energy bins are needed to plot an RGB image")
+
+    axis_rgb = MapAxis.from_energy_edges(
+        energy_edges.value, unit=energy_edges.unit, name="energy", interp=axis.interp
```
QRemy, line 68:
```suggestion
        energy_edges.value, unit=energy_edges.unit, name=axis.name, interp=axis.interp
```
In case the input map is in true energy

gammapy/visualization/utils.py
```diff
@@ -20,6 +23,69 @@
...
+    )
+    map_rgb = map_.resample_axis(axis_rgb)
```
QRemy, line 70: This implicitly  assumes that the map is an integral quantity.
Does plotting a rgb map makes sense if the input map is dnde ?
If yes the `resample` method should be used here, and the preserve_counts option added to `plot_rgb`.

QRemy (changes requested):
Thanks @luca-giunti 
I think for now this support only map intergral maps (counts or npred) given in energy.
Plotting Integral flux in energy true could be also useful, I made a suggestion that should allow this (but a test is probably needed).
I am not sure about what to do if the input is a differential map.

gammapy/visualization/utils.py
```diff
@@ -20,6 +23,69 @@
 }
 
 
+def plot_rgb(map_, energy_edges=None, ax=None, **kwargs):
```
maxnoe, line 26: This function is mixing the plotting with the resampling. I think it would be better to split this. I.e. require  a map with three bins in z-axis in this function and explain in the docs how to use `resample_axis` to create a 3-bin axis from something that has more

maxnoe: E.g. the API would be: `plot_rgb(map_, axis="energy", ax=None, **kwargs)` and inside you do a check that `map_.axes[axis].nbins == 3`

gammapy/visualization/utils.py
```diff
@@ -20,6 +22,63 @@
 }
 
 
+def plot_rgb(map_, axis="energy", ax=None, **kwargs):
```
adonath, line 25: I think the `ax` keyword should come first as in all our other plot methods.

Maybe remove the `axis` keyword completely? The convention could be that RGB plotting can only have one non-spatial axis with exactly three bins.

And maybe rename the function to `plot_map_rgb()`

gammapy/visualization/utils.py
```diff
@@ -20,6 +22,63 @@
...
+
+    This function is based on the `~astropy.visualization.make_lupton_rgb` function, and it
+    assumes that the input map has exactly three z-axis bins. If that were not the case, you
+    can resample the z-axis of your map as shown in the code example below.
+
+    Parameters
+    ----------
+    map_ : `~gammapy.maps.WcsNDMap`
+        WCS map. If energy edges are not provided, the map should contain exactly 3 energy bins.
+    energy_edges : `~astropy.units.Quantity`, float
```
adonath, line 37: Please adapt the docstring to the function signature, they are inconsistent...

gammapy/visualization/utils.py
```diff
@@ -20,6 +22,63 @@
...
+    >>> map_ = map_.resample_axis(axis_rgb)
+    >>> kwargs = {"stretch": 0.5, "Q": 1, "minimum": 0.15}
+    >>> plot_rgb(map_.smooth(0.08*u.deg), axis=axis_rgb.name, **kwargs)
+    """
+    geom = map_.geom
+    if geom.axes[axis].nbin != 3:
+        raise ValueError("Exactly 3 bins are needed to plot an RGB image")
+
+    data = [data_slice / np.nanmax(data_slice.flatten()) for data_slice in map_.data]
+    data = make_lupton_rgb(*data, **kwargs)
```
adonath, line 67: I think it would be good to make the use of the Lupton method optional here, or introduce a `method` keyword. I think a very good approach for RGB plotting is also plotting `norm` maps, i.e. the flux values with respect to a reference spectral model. In this case no special Lupton approach is required...

adonath (commented):
Thanks @luca-giunti, I have left a few inline comments.

gammapy/visualization/utils.py
luca-giunti: @adonath Indeed it makes sense. I propose to add this `method` keyword in another PR. Is that OK?

gammapy/visualization/utils.py
```diff
@@ -20,6 +22,63 @@
...
+
+    ax = map_._plot_default_axes(ax=ax)
+    ax.imshow(data)
+
+    if geom.is_allsky:
+        ax = map_._plot_format_allsky(ax)
+    else:
+        ax = map_._plot_format(ax)
+
+    ax.autoscale(enable=False)
```
QRemy, line 77: added a comment that appears in the other method using this
```suggestion
    # without this the axis limits are changed when calling scatter
    ax.autoscale(enable=False)
```

QRemy (approved):
Thanks, @luca-giunti Looks like all the comments have been implemented.

registerrier (approved):
Thanks @luca-giunti and @QRemy . Looks good.

PR #4212, opened 2022-11-20
Remove unneeded table util function

maxnoe:
**Description**

This pull request removes an unneeded function that nowadays is just supported by astropy out of the box.

**Dear reviewer**

Do we consider these `util` functions as part of the public API? If yes, the function should be deprecated but kept, right?

### DIFF
--- a/gammapy/catalog/core.py
+++ b/gammapy/catalog/core.py
@@ -6,9 +6,10 @@ from copy import deepcopy
 import numpy as np
 from astropy.coordinates import SkyCoord
 from astropy.utils import lazyproperty
+from astropy.table import Table
 from gammapy.maps import TimeMapAxis
 from gammapy.modeling.models import Models
-from gammapy.utils.table import table_from_row_data, table_row_to_dict
+from gammapy.utils.table import table_row_to_dict
 
 __all__ = ["SourceCatalog", "SourceCatalogObject"]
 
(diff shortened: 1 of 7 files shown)


adonath:
Thanks a lot @maxnoe! Yes, this will need a deprecation warning first. I think we agreed on a deprecation period of one minor release, so we could finally remove it for v1.2. So far we do not have a deprecation system in place. @registerrier Do you plan tp work on this soon?

registerrier (approved):
Thanks @maxnoe ! This looks good.

I don't know whether `table_from_row_data` is really part of the public API. I suppose it could go without warning.
Yet it might be good to be broad for now.
I have opened a PR in #4213  for a possible deprecation warning system re-using the one from astropy.

adonath:
@registerrier It is part of the public API, we expose it here: 

https://docs.gammapy.org/1.0/api/gammapy.utils.table.table_from_row_data.html#gammapy.utils.table.table_from_row_data

But it is the perfect use case for testing our new deprecation system... :-)

registerrier:
Thanks @maxnoe . You might rebase now that #4213  is merged and add a deprecation warning on the function instead.

maxnoe:
@adonath @registerrier Rebased and now deprecated instead of removed

registerrier (approved):
Thanks @maxnoe ! Looks good to me.

registerrier:
CI fails are unrelated. Merging now.

PR #4213, opened 2022-11-21
Add deprecation warning system

registerrier:
**Description**

This pull request introduces a deprecation warning system for gammapy.

It reuses the Astropy warning system and replaces the `AstropyDeprecationWarning` by a `GammapyDeprecationWarning`. 

It allows to  add a deprecation to a given function or class
```
from gammapy.utils.deprecation import gammapy_deprecated
@gammapy_deprecated('v1.2')
def some_function():
    return 1
some_function()

<ipython-input-3-6a01c81cb036>:1: GammapyDeprecationWarning: The some_function function is deprecated and may be removed in a future version.
  some_function()
Out[3]: 1
```
and:
```

In [7]: some_function?
Signature: some_function()
Docstring:
.. deprecated:: v1.2
    The some_function function is deprecated and may be removed in a future version.

\ 
File:      ~/Code/gammapy-dev/gammapy/<ipython-input-2-cee7afcf5f38>
Type:      function
```

It also allows to replace one argument with the old one:
```
from gammapy.utils.deprecation import gammapy_deprecated_renamed_argument

@gammapy_deprecated_renamed_argument('old', 'new', 'v1.1')
def dep(new):
    return new

dep(old=2)
<ipython-input-5-230b0d24a1bf>:1: GammapyDeprecationWarning: "old" was deprecated in version v1.1 and will be removed in a future version. Use argument "new" instead.
  dep(old=2)
Out[5]: 2
```

**Dear reviewer**

We can also add the attribute deprecation warning.

The developer doc should be updated to include a description of this system.

### DIFF
--- a/gammapy/utils/deprecation.py
+++ b/gammapy/utils/deprecation.py
@@ -1,0 +1,72 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+from astropy.utils import deprecated
+
+__all__ = ["GammapyDeprecationWarning", "gammapy_deprecated"]
+
+
+class GammapyDeprecationWarning(Warning):
+    """
+    The Gammapy deprecation warning
+    """
+
+
+def gammapy_deprecated(
+    since,
+    message="",
+    name="",
+    alternative="",
+    pending=False,
+    obj_type=None,
+    warning_type=GammapyDeprecationWarning,
+):
+    """
+    Used to mark a function or class as deprecated.
+
+    Reuses Astropy's deprecated decorator
+
+    Parameters
+    ----------
+    since : str
+        The release at which this API became deprecated.  This is required.
+
+    message : str, optional
+        Override the default deprecation message.  The format
+        specifier ``func`` may be used for the name of the function,
+        and ``alternative`` may be used in the deprecation message
+        to insert the name of an alternative to the deprecated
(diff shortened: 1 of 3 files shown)


adonath:
Thanks @registerrier, please add an entry in the developer how to as well: https://docs.gammapy.org/1.0/development/dev_howto.html

adonath (approved):
Thanks a lot @registerrier, this looks all good to me!

AtreyeeS (approved):
Thanks @registerrier ! This looks very good to me

gammapy/utils/deprecation.py
```diff
@@ -0,0 +1,54 @@
...
+
+    Reuses Astropy's deprecated decorator.
+    Check arguments and usage in `~astropy.utils.decorator.deprecated`
+
+    Parameters
+    ----------
+    since : str
+        The release at which this API became deprecated.  This is required.
+    """
+    from astropy.utils import deprecated
```
maxnoe, line 29: Any particular reason why the imports are inline in the function and not at the top?

registerrier: Mostly to avoid confusion with the function names. Maybe importing `astropy.utils` and calling from the namespace would be better.

Issue #4055, opened 2022-08-10
Add Franceschini 2018 EBL model


aleberti:
It would be nice to have the updated Franceschini EBL model, which was published a few years ago.

The values of optical depth can be found in table 4 of https://www.aanda.org/articles/aa/pdf/2018/06/aa29684e-16.pdf

registerrier:
Thanks @aleberti . Are there other EBL models that would be required? I see that agnpy also refers to this reference:
https://ui.adsabs.harvard.edu/abs/2021MNRAS.507.5144S/abstract

Internally, Gammapy relies on the parameteric table model of XSpec to represent the model. (I think agnpy is using the same format. Is that correct @cosimoNigro ?)
So in practice one needs to create such a fits file from the existing table.

cosimoNigro:
Hello @registerrier, yes `agnpy` is using the same format! For the old EBL models I just copied the FITS files in `gammapy-data`. And for the Saldana Lopez 2021, @IlariaViale created a file with the same format starting from the model values publicly available at https://www.ucm.es/blazars/ebl.

If the Franceschini 2018 model is available somewhere I (or @IlariaViale?) can produce another file.
I think the Saldana Lopez one can go directly in.

cosimoNigro:
Hello @aleberti I am looking into this,

It is difficult to include Table 4 of the Franceschini 2018 paper as an opacity model readable by Gammapy.
The redshift values at which the opacity is calculated are not evenly spaced (and are very few - just ten).
Evenly spaced values of the redshift parameter are needed to define the first table of the [XSPEC Table Model](https://heasarc.gsfc.nasa.gov/docs/heasarc/caldb/docs/memos/ogip_92_009/ogip_92_009.pdf) adopted by Gammapy to describe the opacity models.

The paper provides the EBL intensity as a function of energy for a denser grid of z but these values are also not evenly spaced in redshift (delta_z = 0.2 in Table 1 and 2 and delta_z = 0.3 in Table 3...). The intensity should be integrated with the gamma-gamma cross section to obtain the opacity and I think this is not a job for us and the people producing these models can do it better.

Would it be possible for you to obtain from the authors a denser and evenly spaced grid of opacity values? [The other (2008) Franceschini EBL absorption file in gammapy-data](https://github.com/cosimoNigro/gammapy-data/blob/master/ebl/ebl_franceschini.fits.gz) has 2002 spectra (so the absorption is provided for 2002 redshift values).

[...]

PR #4214, opened 2022-11-22
Fix IO for  EBLAbsorptionNormSpectralModel 

QRemy:
This PR fix the IO of the EBLAbsorptionNormSpectralModel in case a filename is not given, and in case the filename point to a built-in model written on another machine.

PR #4215, opened 2022-11-24
Adds built-in Franceschini (2018) and Saldana-Lopez (2021) EBL models

cosimoNigro:
Paired with [PR 29 in gammapy-data](https://github.com/gammapy/gammapy-data/pull/29) this PR adds the EBL models of Franceschini (2018) and Saldana-Lopez (2021).

Solves #4055.

What to do with the tests? It seems to me that no test uses all the built-in models available: there is one testing that a `TemplateSpectralModel` [can read one of the EBL file](https://github.com/gammapy/gammapy/blob/main/gammapy/modeling/models/tests/test_spectral.py#L540) and [one that checks its extrapolation](https://github.com/gammapy/gammapy/blob/main/gammapy/modeling/models/tests/test_spectral.py#L589). Then another [one tests the effect of the absorption](https://github.com/gammapy/gammapy/blob/main/gammapy/modeling/models/tests/test_spectral.py#L550). There are some files untouched by the tests, should we add a test involving all the built-in models?

### DIFF
--- a/examples/models/spectral/plot_absorbed.py
+++ b/examples/models/spectral/plot_absorbed.py
@@ -40,6 +40,12 @@ franceschini = EBLAbsorptionNormSpectralModel.read_builtin(
     "franceschini", redshift=redshift
 )
 finke = EBLAbsorptionNormSpectralModel.read_builtin("finke", redshift=redshift)
+franceschini17 = EBLAbsorptionNormSpectralModel.read_builtin(
+    "franceschini17", redshift=redshift
+)
+saldana21 = EBLAbsorptionNormSpectralModel.read_builtin(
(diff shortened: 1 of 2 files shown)


QRemy:
Thanks @cosimoNigro.
In the read_builtin method,  should we rename the previous ones to also include the year in the name ?
(That's backward incompatible but may be nicer at the end).

registerrier:
@cosimoNigro could you rebase and use the registry introduced in #4214 ?

QRemy (approved):
Thanks @cosimoNigro, I made the necessary change after the rebase, so this can be merged if the models files are in $GAMMAPY_DATA

Issue #3702, opened 2022-01-18
Add information on computing TS of a specific model


AtreyeeS:
Getting the TS of a particular model (from a list of many models) seems to a common question. What we mention in https://docs.gammapy.org/0.19/howto.html#compute-source-significance is probably not sufficient for people starting with multiple models in the initial model definition.

Maybe we can either add some more detailed documentation, or some helper function in our code base.

QRemy:
We could mention in the how-to that the significance of a source assuming that the TS follows a chi2 distribution (see scipy.stats.chi2) can be computed computed as follows:
`Significance = sqrt(chi2.isf(chi2.sf(TS, DoF), 1)`
with
`DoF = len(model.parameters.free_parameters) #could be a model property`
(Note that the survival function sf = 1 - cdf can be more accurate numerically than the cdf.)

We could also add a `TSestimator` to ease the calculation and define some presets for the choice of the frozen parameters.
For the CTA-GPS catalogue I use the following tests:
Test global model with and without each candidate source :
-TSnull : test against global model, all frozen (tested source amplitude frozen or zero)
-TSbkg: test with background norm free (and with tested source amplitude free or zero)
-TSall: test with background norm free + nearby sources amplitude free if d < Robject + Rother + 1 deg (and with tested source amplitude free or zero)

registerrier:
> We could also add a `TSestimator` to ease the calculation and define some presets for the choice of the frozen parameters. For the CTA-GPS catalogue I use the following tests: Test global model with and without each candidate source : -TSnull : test against global model, all frozen (tested source amplitude frozen or zero) -TSbkg: test with background norm free (and with tested source amplitude free or zero) -TSall: test with background norm free + nearby sources amplitude free if d < Robject + Rother + 1 deg (and with tested source amplitude free or zero)

Technically, such a utility would not really be an `Estimator` since it would not estimate a physical parameter but only the statistical significance of a specific model. Of course, this can be extended to be a complete `SkyModelEvaluator` that would take a given component, put its amplitude to 0, then fit it and yield TS as well as all best fit parameters. I suppose that providing detailed error estimates or likelihood scans for each parameter is less obvious. 

This is indeed a feature that is frequently requested.

adonath:
For v1.0 this will remain as a documentation example.

registerrier:
I assume the last missing piece here is an example showing the usage of the function introduced by @QRemy in #4216

registerrier:
[...]

PR #4216, opened 2022-11-24
Add TestStatisticNested class

QRemy:
As discussed in #3702 this PR add a method to compute TS between nested models.
In a separate PR I could add examples in a tutorial to show how to use that for source detection significance, spectral curvature test, or ellipticity test.


registerrier (commented):
Thanks @QRemy . This is an important aspect that must be covered indeed.

I have several very general comments:
- maybe it should not live in `gammapy.stats` because it already contains references to `gammapy.fit` and `gammapy.dataset`
- should such a class perform the fit directly? Or should it rather use the fit results?
We have seen recently with the covariance matrix that the `FitResult` is probably a more important object that we have considered so far. 
One could introduce the same functionality as a `FitResult` comparator. No?

bkhelifi:
Good point, Régis. The class `TestStatisticNested` is probably not well located

AtreyeeS (commented):
I tend to agree with @registerrier here. I had imagined this to be a simple function `significance_nested( result1, result2)` (with probably a better choice of name, residing in gammapy.modelling) which returns the significance correcting for the number of degrees of freedom.

I dont know if this function should have the responsibility of ensuring "nestedness", or we keep it on the user to do not do incorrect things.

adonath:
Thanks @QRemy! I agree with @registerrier and @AtreyeeS , that probably a more suitable place would be `gammapy.estimators`. Given that it does already a very similar thing to the `ParameterEstimator`. The difference is that it does this for multiple parameters at the same time, correct?

QRemy:
For now I moved the `TestStatisticNested` class into `gammapy.modeling.selection` and the two sigma<->ts conversion functions into `gammapy.stats.utils`

gammapy/modeling/selection.py
```diff
@@ -0,0 +1,94 @@
...
+
+    def __init__(
+        self, parameters, null_values, n_sigma=2, n_free_parameters=None, fit=None
+    ):
+        self.parameters = parameters
+        self.null_values = null_values
+        self.n_sigma = n_sigma
+
+        if n_free_parameters is None:
+            n_free_parameters = len(parameters)
```
bkhelifi, line 38: In the docstring, it is mentioned that "parameters: List of parameters frozen for the null hypothesis". So its length is not the number of free parameters, is not it?

gammapy/modeling/tests/test_selection.py
```diff
@@ -0,0 +1,48 @@
...
+def test_test_statistic_link(fermi_datasets):
+
+    # TODO: better test with simuated data ?
+    model = fermi_datasets.models["Crab Nebula"]
+    model2 = model.copy(name="other")
+    model2.spectral_model.alpha.value = 2.4
+    fermi_datasets.models = fermi_datasets.models + [model2]
+
+    fit = Fit()
+    minuit_opts = {"tol": 10, "strategy": 0}
```
bkhelifi, line 38: Why a so large tolerance?

gammapy/modeling/selection.py
QRemy: I should clarify this in the docsting as : "parameters: List of parameters frozen for the null hypothesis but free for the test hypothesis".

gammapy/modeling/tests/test_selection.py
QRemy: just to run the test faster

gammapy/modeling/selection.py
bkhelifi: with, if this list contain frozen and not frozen, its length is still not the number of free parameters. We need instead `len(parameters.free_parameters)`, no?

gammapy/modeling/selection.py
```diff
@@ -0,0 +1,94 @@
...
+
+class TestStatisticNested:
+    """Compute the test statistic (TS) between two nested hypothesis.
+    The null hypothesis is the minimal one, for which a set of parameters
+    are frozen to given values.
+
+    Parameters
+    ----------
+    parameters : `~gammapy.modeling.Parameters` or list of `~gammapy.modeling.Parameter`
+        List of parameters frozen for the null hypothesis.
```
QRemy, line 14:
```suggestion
        List of parameters frozen for the null hypothesis  but free for the test hypothesis.
```

AtreyeeS (approved):
Thanka @QRemy ! It would be nice to expose this in our tutorials. Maybe in a follow up PR?

adonath:
@QRemy Please introduce the TS helper functions in an independent PR and add minimal documentation. Then rebase this one...

adonath:
I think the API here needs a bit more thought. E.g. hardcoding the fit in the `TestStatisticNested` is not a good idea. There is no access to the fit results, all the important information (even on the best fit model parameters) is just thrown away. The approach does nor really scale to testing multiple models either. Also the API to modify the model parameters in place based on the `ts_threshold` seems weird to me. The behavior is not documented either. 
 
I think the approach that @AtreyeeS suggested is much cleaner, e.g. based on the `FitResult` object. Like:

```python3
datasets.models =  models_null
result_null = fit.run(datasets)

datasets.models =  models_1
result_1 = fit.run(datasets)

datasets.models =  models_2
result_2 = fit.run(datasets)

significance_nested(model_null, models_1)
significance_nested(model_null, models_2)

# or for direct model selection
model = select_model(result_null, [result_1, result_2], ts_threshold=...)
```
If we are worried about the nested assumption we could e.g. introduce a check like `Model.is_nested(other_model)`.

adonath:
Right now I don't have the time to think about this, but it would be good to figure out the relation of this functionality to e.g. the `ParameterEstimator` which does model testing as well, but limited to a single parameter. If there is a way to generalize the current `ParameterEstimator` this would be preferable...

registerrier:
> Right now I don't have the time to think about this, but it would be good to figure out the relation of this functionality to e.g. the `ParameterEstimator` which does model testing as well, but limited to a single parameter. If there is a way to generalize the current `ParameterEstimator` this would be preferable...

I think an argument for having the fit done inside the `run` method of the class is to guarantee that the models are nested, which is more difficult to validate by comparing the `Models`. 

Indeed, working with an `Estimator` would broaden the scope since one would be able to keep track of the fitted parameters and their errors. An option could be to use a `ModelEstimator` that takes a model component name and performs a fit with and without the model component (possibly with `reoptimize` option). Then besides the TS this would yield parameter errors, likelihood profiles etc. When designing the `ParameterEstimator` this was a plausible option but was not so easy to use as parent class for the `FluxPointsEstimator`.  

Of course, such an `Estimator` can remove the need of using directly `Fit` for quite a few use cases. This might be a good thing (or bad).

adonath:
> I think an argument for having the fit done inside the run method of the class is to guarantee that the models are nested, which is more difficult to validate by comparing the Models.

Yes, but it also limits the scope. As far as I understood the current `TestStatisticNested` could not compare e.g. models where we use `PowerLawSpectralModel` and `ExponentialCutoffPowerLawSpectralModel`, even if the models are nested. But I agree that a general scheme for checking nested models is a bit of work, but it is not hard if we just declare it:

```python3
class ExponentialCutoffPowerLawSpectralModel:
    _includes_nested = ["pwl"]
```

QRemy:
> > I think an argument for having the fit done inside the run method of the class is to guarantee that the models are nested, which is more difficult to validate by comparing the Models.
> 
> Yes, but it also limits the scope. As far as I understood the current `TestStatisticNested` could not compare e.g. models where we use `PowerLawSpectralModel` and `ExponentialCutoffPowerLawSpectralModel`, even if the models are nested.

You can using an ExponentialCutoffPowerLawSpectralModel with a valid  null value for the cutoff energy (far outside the energy max of the dataset). With the support of parameter linking we can even test broken power law vs. power law by linking the two index of the broken power-law. In principle you can always  find a null value for the complex model that match the simpler model if they are nested.

> But I agree that a general scheme for checking nested models is a bit of work, but it is not hard if we just declare it:
> 
> ```python
> class ExponentialCutoffPowerLawSpectralModel:
>     _includes_nested = ["pwl"]
> ```
This will requires more maintenance each time new models are added.

Moreover checking nested models also means that all the other parameters have to be the same which is more complex to check  when different `Models` instances are compared especially if the sources name change (for example one could use source1-pl, source1-ecpl).

QRemy:
> There is no access to the fit results, all the important information (even on the best fit model parameters) is just thrown away. 

The `fit` object is attached to the class, I could  also attach `fit_results` and `fit_results_null` so we have still access to all fitted parameters.

>```
># or for direct model selection
>model = select_model(result_null, [result_1, result_2], ts_threshold=...)
>```
The `FitResult` is not part of the public API (same for `OptimizeResult` and `CovarianceResult`).

>Also the API to modify the model parameters in place based on the ts_threshold seems weird to me. 

I can attach a `.select_model(ts_threshold)` method to this class to do it separately.
And when possible, as here, I prefer to modify a model parameters in place rather than setting a new models instance on the datasets so the cache of all the other models is not reset.

gammapy/modeling/selection.py
```diff
@@ -0,0 +1,94 @@
...
+            fit.optimize_opts = minuit_opts
+        self.fit = fit
+
+    @property
+    def ts_threshold(self):
+        """Threshold value in TS corresponding to `n_sigma`.
+        This assumes that the TS follows a chi squared distribution
+        with a number of degree of freedom equal to `len(parameters)`.
+        """
+        return sigma_to_ts(self.n_sigma, len(self.parameters))
```
QRemy, line 54:
```suggestion
        with a number of degree of freedom equal to `n_free_parameters`.
        """
        return sigma_to_ts(self.n_sigma, self.n_free_parameters)
```

AtreyeeS:
> The FitResult is not part of the public API (same for OptimizeResult and CovarianceResult).

As a side issue, would be good to expose this, no? We have already seen that the full covariance information is accessible only on the FitResult

registerrier:
Now that #4346 is merged. This needs to be updated.

gammapy/modeling/selection.py
```diff
@@ -0,0 +1,105 @@
...
+    n_sigma : float
+        Threshold in number of sigma to switch from the null hypothesis
+        to the alternative one. Default is 2.
+    n_free_parameters : int
+        Number of free parameters to consider between the two hypothesis
+        in order to estimate the `ts_threshold` from the `n_sigma` threshold.
+        Default is len(parameters).
+    fit : `Fit`
+        Fit instance specifying the backend and fit options.
+    """
```
bkhelifi, line 30: the `n_sigma` to DeltaTS conversion supposes that the Wilk's theorem is valid for this dataset. Maybe one should precise this into the docstring (otherwise, as mentioned by @adonath, simulations should be done)

QRemy: added

registerrier (approved):
Thanks @QRemy . No further comments.

PR #4217, opened 2022-11-24
FluxPointsDataset support model with spatial template and NormSpectralModel

### LINKED ISSUE #4019: Add support for NormSpectralModel in FluxEstimator

QRemy:
This PR change the `.flux_pred()` method in FluxPointsDataset in order to support reference models relying on spatial template and NormSpectralModel.
This should resolve #4019.

### DIFF
--- a/gammapy/datasets/flux_points.py
+++ b/gammapy/datasets/flux_points.py
@@ -1,7 +1,6 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 import logging
 import numpy as np
-from scipy.interpolate import interp1d
 from astropy import units as u
 from astropy.io import fits
 from astropy.table import Table
@@ -9,7 +8,7 @@ from astropy.visualization import quantity_support
 import matplotlib.pyplot as plt
 from matplotlib.gridspec import GridSpec
 from gammapy.maps.axes import UNIT_STRING_FORMAT
-from gammapy.modeling.models import DatasetModels
+from gammapy.modeling.models import DatasetModels, TemplateSpatialModel
 from gammapy.utils.scripts import make_name, make_path
 from .core import Dataset
 
@@ -286,24 +285,10 @@ class FluxPointsDataset(Dataset):
         for model in self.models:
             if isinstance(model.spatial_model, TemplateSpatialModel):
                 geom = model.spatial_model.map.geom
-                spatial_integ = model.spatial_model.integrate_geom(geom)
-                uspatial = spatial_integ.unit
-                spatial_integ.data[~np.isfinite(spatial_integ.data)] = np.nan
-                if model.spatial_model.is_energy_dependent:
-                    spatial_integ = np.nansum(spatial_integ.data, axis=(1, 2))
-                    energy = geom.axes[0].center.to(self.data.energy_ref.unit)
-                    interp = interp1d(
-                        np.log10(energy.value), np.log10(spatial_integ), kind="linear"
-                    )
-                    spatial_integ = (
-                        10 ** interp(np.log10(self.data.energy_ref.value)) * uspatial
-                    )
-                else:
-                    spatial_integ = np.nansum(spatial_integ.data) * uspatial
-                spectral_values = model.spectral_model(self.data.energy_ref)
-                flux_model = spectral_values * spatial_integ
+                reference_model = model.to_template_spectral_model(geom)
             else:
-                flux_model = model.spectral_model(self.data.energy_ref)
+                reference_model = model.spectral_model
+            flux_model = reference_model(self.data.energy_ref)
 
             if model.temporal_model is not None:
                 integral = model.temporal_model.integral(
--- a/gammapy/datasets/flux_points.py
+++ b/gammapy/datasets/flux_points.py
@@ -17,14 +17,15 @@ log = logging.getLogger(__name__)
 __all__ = ["FluxPointsDataset"]
 
 
-def _get_reference_model(energy_edges, model):
+def _get_reference_model(energy_edges, model, margin_percent=70):
     if isinstance(model.spatial_model, TemplateSpatialModel):
         geom = model.spatial_model.map.geom
-        if geom.is_image:
-            energy_axis = MapAxis.from_energy_bounds(
-                energy_edges[0], energy_edges[-1], nbin=10, per_decade=True
-            )
-            geom = geom.to_cube([energy_axis])
+        emin = energy_edges[0] * (1 - margin_percent / 100)
+        emax = energy_edges[-1] * (1 + margin_percent / 100)
+        energy_axis = MapAxis.from_energy_bounds(
+            emin, emax, nbin=20, per_decade=True, name="energy_true"
+        )
+        geom = geom.to_image().to_cube([energy_axis])
         return Models([model]).to_template_spectral_model(geom)
     else:
         return model.spectral_model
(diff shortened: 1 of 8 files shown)


registerrier (commented):
Thanks @QRemy !

I think this is not really a bug fix but rather a new feature, so I remove the backport label.

Is the approach here solving the issue? 
The problem is already that the reference model of the `FluxPoints` object is not correct.
Maybe the operation should rather be applied when creating the `FluxPoints` to create a `TemplateSpectralModel` containing the integrated spatial template multiplied by the norm model. Maybe a method to directly a `TemplateSpectralModel` from an energy dependent `SpatialTemplateModel` and a region?

Then I think it is difficult to consider the spatial model on a `FluxPointsDataset`. How to perform the spatial integration in general? Here you integrate over the whole geometry of the model, but it is not necessarily relevant in all cases.

AtreyeeS (commented):
Thanks @QRemy for going back to adding new features.

So the model will be integrated over the full geometry if there is a `TemplateSpatialModel`, but not for other spatial models? Is the the best option?

Maybe `FluxPoints.reference_model` should always be converted to a  `TemplateSpectralModel` ? This then also addresses the difference introduced during serialisation as discussed in https://github.com/gammapy/gammapy/issues/3747#issuecomment-1022790580

AtreyeeS:
I guess it would be nice to have this in 1.1. Is the feasible @QRemy ?

adonath:
> Is the approach here solving the issue?
The problem is already that the reference model of the FluxPoints object is not correct.

To be more specific: the reference model is stored correctly on the `FluxPoints` object, however the spatial model is not used to compute the reference fluxes. So I think the correct fix for #4019 is to change from evaluating the spectral model only in `FluxPoints.dnde_ref`, `FluxPoints.flux_ref` etc. to evaluating the `SkyModel` instead, by integrating over the spatial part (for now over the full support).

>Maybe the operation should rather be applied when creating the FluxPoints to create a TemplateSpectralModel containing 

This could be an explicit method on the `FluxPoints` object. E.g. like:

```python
fp_new = fp.to_reference_model(model=None)
```

In case `model=SkyModel()` one would recompute all quantities and transform the `FluxPoints` to the new reference model. The case `model=None` could then support the option to convert to a `TemplateSpectralModel` by default. 

@QRemy Supporting spatial models on the `FluxPointsDataset` seems a useful feature in general, but I think the current implementation has the code in the wrong location. I think the the evaluation should just go through `SkyModel.evaluate_geom() / .integrate_geom()`. Please do not introduce any new hard coded interpolation / integration methods.

bkhelifi:
@QRemy , can you update your code from the main for the CI tests?

gammapy/modeling/models/core.py
```diff
@@ -990,6 +990,41 @@ def to_template_sky_model(self, geom, spectral_model=None, name=None):
...
+        """Merge a list of models into a single `~gammapy.modeling.models.TemplateSpectralModel`
+
+        Parameters
+        ----------
+        geom : `Geom`
+            Map geometry the evaluate model.
+        mask :  `Map`
+            Evaluate the model only where the mask is True.
+        maks : `~gammapy.modeling.models.SpectralModel`
+            One of the NormSpectralMdel
```
QRemy, line 1003:
```suggestion
        maks : `~gammapy.modeling.models.SpectralModel`
            One of the NormSpectralMdel
```

QRemy:
The new solution introduces  `Models.to_template_spectral_model`, this allows to get a valid spectrum for flux points evaluation when the models includes spatial templates. 
it would be nice to have this in the next release.

gammapy/datasets/flux_points.py
```diff
@@ -17,6 +17,20 @@
 __all__ = ["FluxPointsDataset"]
 
 
+def _get_reference_model(model, energy_bounds, margin_percent=70):
+    if isinstance(model.spatial_model, TemplateSpatialModel):
```
registerrier, line 21: Technically, this spatial integration should be performed for any spatial model that is energy dependent right? Maybe this should be the criterion applied?

registerrier: Or maybe also if the model does not integrate to unity...

gammapy/datasets/flux_points.py
```diff
@@ -17,6 +17,20 @@
...
+        geom = model.spatial_model.map.geom
```
registerrier, line 22: What happens here if the template is defined on a much larger geometry than the actual geom used?

registerrier (commented):
Thanks @QRemy ! I have left some comments inline.

More generally, is the behavior correct for large templates? How far should one integrate to convert to a spectral model?
Is there a possibility to have a test with a diffuse feature?

gammapy/datasets/flux_points.py
QRemy: I think that's the same for any spatial model the extent of the model could be larger than the mask fit and yet we consider the full model.

gammapy/datasets/flux_points.py
QRemy: in the case the geom could re-use what we have for spatial model plotting :https://github.com/gammapy/gammapy/blob/0afcf2ab0bbfae1559b0546c156e9544e7c19389/gammapy/modeling/models/spatial.py#L277-L284

gammapy/datasets/flux_points.py
QRemy: but I can add the a `mask` argument here and pass it to `to_template_spectral_model`

gammapy/estimators/points/tests/test_sed.py
```diff
@@ -609,3 +625,56 @@ def test_fpe_non_uniform_datasets():
...
         fpe.run(datasets=[dataset_1, dataset_2])
+
+
+@requires_data()
+def test_flux_points_estimator_norm_spectral_model(fermi_datasets):
+
+    filename = "$GAMMAPY_DATA/fermi-3fhl-crab/Fermi-LAT-3FHL_datasets.yaml"
+    filename_models = "$GAMMAPY_DATA/fermi-3fhl-crab/Fermi-LAT-3FHL_models.yaml"
+
+    fermi_datasets = Datasets.read(filename=filename, filename_models=filename_models)
```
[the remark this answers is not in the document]
QRemy: removed

gammapy/modeling/models/core.py
```diff
@@ -990,6 +990,39 @@ def to_template_sky_model(self, geom, spectral_model=None, name=None):
             spectral_model=spectral_model, spatial_model=spatial_model, name=name
         )
 
+    def to_template_spectral_model(self, geom, mask=None):
```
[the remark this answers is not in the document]
QRemy: Added a test.
I don't understand your other point,  for `to_template_sky_model` the spatial integration is not needed

gammapy/estimators/points/tests/test_sed.py
```diff
@@ -19,13 +24,24 @@
...
+    PiecewiseNormSpectralModel,
     PowerLawSpectralModel,
     SkyModel,
+    TemplateSpatialModel,
 )
 from gammapy.utils import parallel
 from gammapy.utils.testing import requires_data, requires_dependency
 
 
+@pytest.fixture(scope="session")
```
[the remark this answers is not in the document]
QRemy: pytest doesn't find the fixture from gammapy/modeling/tests, so I duplicate it here.
We have the same problem in gammapy/estimators/tests/test_flux.py.

gammapy/datasets/flux_points.py
QRemy: Let's keep that for a separate PR.

registerrier (approved):
Thanks @QRemy. No major comments from my side.

Issue #4027, opened 2022-07-18
Implement parameter scaling for fitting


adonath:
We already have the `Parameter.interp` property which is used to compute the parameter grid for stat profiles. It would be nice to use the same information to actually scale the parameter value that is visible to the fitter. E.g. some parameters would benefit from a log scaling for faster convergence.

registerrier:
This is done in #4218 . Closing.

PR #4218, opened 2022-11-25
Add interp_scale to parameter scaling

QRemy:
Use the interpolation scale given by Parameter.interp in the transform from the parameter value to the factor seen by the optimizer, as proposed in #4027.
Not sure how to combine this with the scale method, for now I apply interp_scale first and then the scale method.

### DIFF
--- a/gammapy/modeling/parameter.py
+++ b/gammapy/modeling/parameter.py
@@ -518,7 +518,7 @@ class Parameter:
         return output
 
     def update_scale(self, value):
-        """update the parameter scale.
+        """Update the parameter scale.
 
         Set ``factor`` and ``scale`` according to ``scale_method`` attribute.
 
@@ -555,11 +555,11 @@ class Parameter:
         return value
 
     def autoscale(self):
-        "apply interpolation_scale and scale_method to the parameter"
+        "Apply `interpolation_scale' and `scale_method' to the parameter."
         self.factor = self.transform(self.value, update_scale=True)
 
     def reset_autoscale(self):
-        "reset scaling such as factor=value, scale=1"
+        "Reset scaling such as factor=value, scale=1."
         self._factor = self._value
         self._scale = 1.0
 
--- a/gammapy/modeling/parameter.py
+++ b/gammapy/modeling/parameter.py
@@ -251,9 +251,10 @@ class Parameter:
 
     @property
     def factor_min(self):
-        """Factor minimum as a float.
+        """Factor minimum as a float (used by the optimizer).
 
-        This ``factor_min = transform(min)`` is for the optimizer interface.
+        By default when no transform is applied, ``factor_min = min / scale``,
+        otherwise ``factor_min = transform(min)``.
         """
         return self.transform(self.min)
 
@@ -272,9 +273,10 @@ class Parameter:
 
     @property
     def factor_max(self):
-        """Factor maximum as a float.
+        """Factor maximum as a float (used by the optimizer).
 
-        This ``factor_max = transform(max)`` is for the optimizer interface.
+        By default when no transform is applied, ``factor_max = max / scale``,
+        otherwise ``factor_max = transform(max)``.
         """
         return self.transform(self.max)
 
--- a/gammapy/modeling/parameter.py
+++ b/gammapy/modeling/parameter.py
@@ -128,8 +128,6 @@ class Parameter:
         self._scale_interp = scale_interp
         self.interp = interp
         self._scale = float(scale)
-        self.min = min
-        self.max = max
         self.frozen = frozen
         self._error = error
         self._type = None
@@ -222,7 +220,7 @@ class Parameter:
     @factor.setter
     def factor(self, val):
         self._factor = float(val)
-        self._value = self.inverse_transform(self._factor)
+        self._value = float(self.inverse_transform(self._factor))
 
     @property
     def scale(self):
--- a/gammapy/modeling/parameter.py
+++ b/gammapy/modeling/parameter.py
@@ -90,13 +90,13 @@ class Parameter:
      scan_values: `numpy.array`
          Scan values. Overwrites all the scan keywords before.
      scale_method : {'scale10', 'factor1', None}
-         Method used to set ``factor`` and ``scale``.
+         Method used to set ``factor`` and ``scale``.  Default is "scale10".
      interp : {"lin", "sqrt", "log"}
-         Parameter scaling to use for the scan.
+         Parameter scaling to use for the scan.  Default is "lin".
      prior : `~gammapy.modeling.models.Prior`
          Prior set on the parameter.
     scale_transform : {"lin", "sqrt", "log"}
-        Parameter scaling.
+        Parameter scaling. Default is "lin".
     """
 
     def __init__(
@@ -538,8 +538,8 @@ class Parameter:
         ----------
         value : float
             Parameter value
-        update_scale : bool
-            Update the scaling (used by the autoscale)
+        update_scale : bool, optional
+            Update the scaling (used by the autoscale). Default is False.
         """
         interp_scale = interpolation_scale(self.scale_transform)
         factor = interp_scale(value)
--- a/gammapy/modeling/parameter.py
+++ b/gammapy/modeling/parameter.py
@@ -356,7 +356,7 @@ class Parameter:
 
     @value.setter
     def value(self, val):
-        self._value = val
+        self._value = float(val)
         self._factor = self.transform(val)
 
     @property
--- a/gammapy/modeling/parameter.py
+++ b/gammapy/modeling/parameter.py
@@ -113,6 +113,7 @@ class Parameter:
         scan_n_sigma=2,
         scan_values=None,
         scale_method="scale10",
+        scale_interp="lin",
         interp="lin",
         prior=None,
     ):
@@ -122,6 +123,7 @@ class Parameter:
         self._name = name
         self._link_label_io = None
         self.interp = interp
+        self.scale_interp = scale_interp
         self.scale = scale
         self.frozen = frozen
         self._error = error
@@ -524,7 +526,7 @@ class Parameter:
             self.scale = value
 
     def transform(self, value, update_scale=False):
-        interp_scale = interpolation_scale(self.interp)
+        interp_scale = interpolation_scale(self.scale_interp)
         factor = interp_scale(value)
         if update_scale:
             self.update_scale(factor)
@@ -532,7 +534,7 @@ class Parameter:
 
     def inverse_transform(self, factor):
         factor = self.scale * factor
-        interp_scale = interpolation_scale(self.interp)
+        interp_scale = interpolation_scale(self.scale_interp)
         value = interp_scale.inverse(factor)
         return value
 
--- a/gammapy/modeling/parameter.py
+++ b/gammapy/modeling/parameter.py
@@ -555,10 +555,20 @@ class Parameter:
         value : float
             Parameter factor
         """
-        factor = self.scale * factor
         interp_scale = interpolation_scale(self.scale_transform)
-        value = interp_scale.inverse(factor)
+        value = interp_scale.inverse(self.scale * factor)
         return value
+
+    def _inverse_transform_derivative(self, factor):
+        """Inverse tranform from factor (used by the optimizer) to value.
+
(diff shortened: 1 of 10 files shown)


adonath:
Thanks @QRemy! Should deprecate the "old scaling" methods? @registerrier

QRemy:
For now they are still applied after the scaling from the .interp argument, not sure if it makes sense.

adonath:
@QRemy I think we never really discussed this PR in our dev call, would you mind setting it on the agenda for this Friday? https://github.com/gammapy/gammapy-meetings/tree/master/dev-meetings/2023/2023-01-27 Thanks!

QRemy:
For backward compatibility I added a `scale_interp="lin"` argument instead of using direclty `interp` so the transform is done correctly for previously written models.

gammapy/modeling/parameter.py
```diff
@@ -134,16 +140,14 @@ def __init__(
             self.value = val.value
             self.unit = val.unit
         else:
-            self.factor = value
+            self.value = float(value)
```
QRemy, line 143: This looks like a bug, I will open PR to solve it separately.

QRemy: Done in #4324.

bkhelifi:
@QRemy what is the status of this work?

gammapy/modeling/parameter.py
```diff
@@ -447,15 +463,34 @@ def autoscale(self):
...
+        interp_scale = interpolation_scale(self.scale_interp)
+        value = interp_scale.inverse(factor)
+        return value
+
+    def autoscale(self):
+        "apply interpolation_scale and scale_method to the parameter"
+        self.factor = self.transform(self.value, update_scale=True)
+
+    def reset_autoscale(self):
+        "reset sacling such as factor=value, scale=1"
```
QRemy, line 491:
```suggestion
        "reset scacling such as factor=value, scale=1"
```

Astro-Kirsty:
@QRemy, what is the status here?
Once rebased, ready for review?

gammapy/modeling/parameter.py
```diff
@@ -97,6 +97,8 @@ class Parameter:
         Whether the parameter represents the flux norm of the model.
     prior : `~gammapy.modeling.models.Prior`
         Prior set on the parameter.
+    scale_interp : {"lin", "sqrt", "log"}
```
bkhelifi, line 100: Can you profit of this PR to add in the docstring all the default values? Thanks

gammapy/modeling/parameter.py
```diff
@@ -483,15 +499,34 @@ def autoscale(self):
...
+        return factor / self.scale
+
+    def inverse_transform(self, factor):
+        factor = self.scale * factor
+        interp_scale = interpolation_scale(self.scale_interp)
+        value = interp_scale.inverse(factor)
+        return value
+
+    def autoscale(self):
+        "apply interpolation_scale and scale_method to the parameter"
```
bkhelifi, line 523:
```suggestion
        "Apply `interpolation_scale' and `scale_method' to the parameter."
```

gammapy/modeling/parameter.py
```diff
@@ -466,8 +482,8 @@ def to_dict(self):
             output["prior"] = self.prior.to_dict()["prior"]
         return output
 
-    def autoscale(self):
-        """Autoscale the parameters.
+    def update_scale(self, value):
+        """update the parameter scale.
```
bkhelifi, line 486:
```suggestion
        """Update the parameter scale.
```

gammapy/modeling/parameter.py
```diff
@@ -483,15 +499,34 @@ def autoscale(self):
...
-                scale = np.power(10.0, exponent)
-                self.factor = value / scale
-                self.scale = scale
+                self._scale = np.power(10.0, exponent)
 
         elif self.scale_method == "factor1":
-            self.factor, self.scale = 1, self.value
+            self._scale = value
+
+    def transform(self, value, update_scale=False):
```
bkhelifi, line 509: What does exactly this function? Do not forget the default value of the argument

gammapy/modeling/parameter.py
```diff
@@ -483,15 +499,34 @@ def autoscale(self):
...
+        interp_scale = interpolation_scale(self.scale_interp)
+        factor = interp_scale(value)
+        if update_scale:
+            self.update_scale(factor)
+        return factor / self.scale
+
+    def inverse_transform(self, factor):
```
bkhelifi, line 516: What does exactly this function? A docstring might be great

gammapy/modeling/parameter.py
```diff
@@ -483,15 +499,34 @@ def autoscale(self):
...
+        interp_scale = interpolation_scale(self.scale_interp)
+        value = interp_scale.inverse(factor)
+        return value
+
+    def autoscale(self):
+        "apply interpolation_scale and scale_method to the parameter"
+        self.factor = self.transform(self.value, update_scale=True)
+
+    def reset_autoscale(self):
+        "reset scaling such as factor=value, scale=1"
```
bkhelifi, line 527:
```suggestion
        "Reset scaling such as factor=value, scale=1."
```

gammapy/modeling/parameter.py
```diff
@@ -824,12 +859,16 @@ def __init__(
         min=np.nan,
         max=np.nan,
         error=0,
+        scale_method="scale10",
```
bkhelifi, line 862: AS this class is exposed, it might be great to add a docstring describing it with its parameters (and default values)

bkhelifi (changes requested):
Thanks a lot @QRemy. It seems more coherent now. I have just put some suggestions on the docstrings

QRemy:
Now the inverse transform of the covariance is implemented using `propagate` from `jacobi` package
So it's ready for review.

gammapy/modeling/parameter.py
QRemy: added

gammapy/modeling/parameter.py
QRemy: added

gammapy/modeling/parameter.py
QRemy: inherited from Parameter

gammapy/modeling/parameter.py
```diff
@@ -248,6 +249,14 @@ def min(self, val):
         else:
             self._min = self._set_quantity_str_float(val)
 
+    @property
+    def factor_min(self):
```
bkhelifi, line 253: Per curiosity, for which purposes the two function factor_min and factor_max have been introduced? It seems not directly related to the PR topics

QRemy:
One last review before merging this old PR ?

gammapy/modeling/parameter.py
```diff
@@ -843,12 +894,16 @@ def __init__(
         min=np.nan,
         max=np.nan,
         error=0,
+        scale_method="scale10",
```
bkhelifi, line 897: This is really sad that this important class has no docstring... A reference to some tutorials could be also appreciated by users.
Even if it is not the scope of this PR, one could add that now OR make quickly an issue to update the documentation.

bkhelifi (dismissed):
OK for me, thanks @QRemy .
I have a comment on the docstring of the class ParameterPrior.

gammapy/modeling/parameter.py
```diff
@@ -261,6 +270,14 @@ def max(self, val):
         else:
             self._max = self._set_quantity_str_float(val)
 
+    @property
+    def factor_max(self):
+        """Factor maximum as a float.
+
+        This ``factor_max = transform(max)`` is for the optimizer interface.
```
registerrier, line 277: For clarity, I would keep the previous definition and add the new one as most of the time transform is identity:
```
In general, when no transform is applied, ``factor_min = min / scale`` else ``factor_min = transform(min)``
```

gammapy/modeling/parameter.py
```diff
@@ -115,13 +117,17 @@ def __init__(
...
         if not isinstance(name, str):
             raise TypeError(f"Name must be string, got '{type(name)}' instead")
 
         self._name = name
         self._link_label_io = None
-        self.scale = scale
+        self._scale_method = scale_method
+        self._scale_transform = scale_transform
+        self.interp = interp
+        self._scale = float(scale)
```
registerrier, line 130: I must say that `Parameter` argument have become very complex. Is there a use case for a different scaling for scan and for parameter transform?
Is `scale_method` only relevant for `scale_transform=="lin"`?

gammapy/modeling/parameter.py
```diff
@@ -499,15 +515,50 @@ def autoscale(self):
...
+
+        Parameters
+        ----------
+        value : float
+            Parameter value
+        update_scale : bool, optional
+            Update the scaling (used by the autoscale). Default is False.
+        """
+        interp_scale = interpolation_scale(self.scale_transform)
+        factor = interp_scale(value)
```
registerrier, line 536: I understand this is not the same `factor` as before, and I must say this is confusing when trying to understand the code.

gammapy/modeling/parameter.py
```diff
@@ -499,15 +515,50 @@ def autoscale(self):
...
+        if update_scale:
+            self.update_scale(factor)
```
registerrier, line 538: Does `update_scale` take the factor or the value in input?

registerrier (commented):
Thanks @QRemy. I have left some comments inline. 

My main comment is that this PR is introducing significant additional complexity, changes the serialization format of models and introduce an additional dependency (currently in all cases or you have to forbid setting the transform_scale to anything other than `"lin"` if `jacobi` is not installed).

Before merging, I think it would be important to clarify to what extent this PR brings a clear gain? In which use cases?  Can't priors produce similar results since changing the measure of a parameter must be equivalent to setting a prior?

gammapy/modeling/parameter.py
QRemy: the scale can also be applied `scale_transform!="lin"` but I have no idea if this is useful.

QRemy:
> priors produce similar results since changing the measure of a parameter must be equivalent to setting a prior

priors change the likelihood while this change only how the parameter space is explored

setup.cfg
```diff
@@ -59,6 +59,7 @@ all =
     ipywidgets
     ray[default]>=2.9
     ultranest
+    jacobi
```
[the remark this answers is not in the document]
QRemy: if we want to avoid new dependancy we could use the jacobian from scipy:
https://docs.scipy.org/doc/scipy/reference/generated/scipy.differentiate.jacobian.html
and add in gammapy.utils a copy of the propagate method which is quite small:
https://github.com/HDembinski/jacobi/blob/main/src/jacobi/_propagate.py

gammapy/modeling/parameter.py
```diff
@@ -309,8 +310,21 @@ def scale_method(self):
...
+        self.reset_autoscale()
         self._scale_method = val
 
+    @property
+    def scale_transform(self):
+        """scale interp : {"lin", "sqrt", "log"}"""
+        return self._scale_transform
+
+    @scale_transform.setter
+    def scale_transform(self, val):
```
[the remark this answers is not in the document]
QRemy: removed jacobi

setup.cfg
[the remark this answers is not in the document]
QRemy: removed jacobi and now the derivative are computed analytically

gammapy/modeling/parameter.py
QRemy: renamed transformed

bkhelifi:
What is the status here, @QRemy ?

QRemy:
> What is the status here, @QRemy ?

Ready

registerrier (approved):
Thanks @Qremy . Looks good. No further comments.

Since the commits are spread over a long period, isn't it beter for the git history here to squash and merge?

QRemy:
> Since the commits are spread over a long period, isn't it beter for the git history here to squash and merge?

not sure about that I think we can just merge it like this.

registerrier:
> > Since the commits are spread over a long period, isn't it beter for the git history here to squash and merge?
> 
> not sure about that I think we can just merge it like this.

OK Let's make a simple merge once the CI is finished.

PR #4219, opened 2022-11-25
Allow reading of IRF files with single-value axes

lmohrmann:
**Description**
I found that it is not possible to read IRF files if they have a single-valued axis. Specifically, I was trying to read [this file](https://github.com/gammapy/gammapy/files/10093286/aeff.fits.gz) using `EffectiveAreaTable2D.read`, which has a single entry in the `THETA_LO` / `THETA_HI` axis. This resulted in the following error:

```python
----> 1 aeff = EffectiveAreaTable2D.read('aeff.fits.gz')

File /lfs/l7/hess/users/mohrmann/software/anaconda3/envs/gammapy-1.0/lib/python3.9/site-packages/gammapy/irf/core.py:434, in IRF.read(cls, filename, hdu, format)
    417 """Read from file.
    418 
    419 Parameters
   (...)
    431     IRF class
    432 """
    433 with fits.open(str(make_path(filename)), memmap=False) as hdulist:
--> 434     return cls.from_hdulist(hdulist, hdu=hdu)

File /lfs/l7/hess/users/mohrmann/software/anaconda3/envs/gammapy-1.0/lib/python3.9/site-packages/gammapy/irf/core.py:413, in IRF.from_hdulist(cls, hdulist, hdu, format)
    410 if hdu is None:
    411     hdu = IRF_DL3_HDU_SPECIFICATION[cls.tag]["extname"]
--> 413 return cls.from_table(Table.read(hdulist[hdu]), format=format)

File /lfs/l7/hess/users/mohrmann/software/anaconda3/envs/gammapy-1.0/lib/python3.9/site-packages/gammapy/irf/core.py:452, in IRF.from_table(cls, table, format)
    436 @classmethod
    437 def from_table(cls, table, format="gadf-dl3"):
    438     """Read from `~astropy.table.Table`.
    439 
    440     Parameters
   (...)
    450         IRF class.
    451     """
--> 452     axes = MapAxes.from_table(table=table, format=format)
    453     axes = axes[cls.required_axes]
    454     column_name = IRF_DL3_HDU_SPECIFICATION[cls.tag]["column_name"]

File /lfs/l7/hess/users/mohrmann/software/anaconda3/envs/gammapy-1.0/lib/python3.9/site-packages/gammapy/maps/axes.py:1999, in MapAxes.from_table(cls, table, format)
   1997 for column_prefix in IRF_DL3_AXES_SPECIFICATION:
   1998     try:
-> 1999         axis = MapAxis.from_table(
   2000             table, format=format, column_prefix=column_prefix
   2001         )
   2002     except KeyError:
   2003         continue

File /lfs/l7/hess/users/mohrmann/software/anaconda3/envs/gammapy-1.0/lib/python3.9/site-packages/gammapy/maps/axes.py:1307, in MapAxis.from_table(cls, table, format, idx, column_prefix)
   1304 edges_hi = table[f"{column_prefix}_HI"].quantity[0]
   1306 if np.allclose(edges_hi, edges_lo):
-> 1307     axis = MapAxis.from_nodes(edges_hi, interp=interp, name=name)
   1308 else:
   1309     edges = edges_from_lo_hi(edges_lo, edges_hi)

File /lfs/l7/hess/users/mohrmann/software/anaconda3/envs/gammapy-1.0/lib/python3.9/site-packages/gammapy/maps/axes.py:615, in MapAxis.from_nodes(cls, nodes, **kwargs)
    598 @classmethod
    599 def from_nodes(cls, nodes, **kwargs):
    600     """Generate an axis object from a sequence of nodes (bin centers).
    601 
    602     This will create a sequence of bins with edges half-way
   (...)
    613         coordinates.  Default: 'lin'.
    614     """
--> 615     if len(nodes) < 1:
    616         raise ValueError("Nodes array must have at least one element.")
    618     return cls(nodes, node_type="center", **kwargs)

File /lfs/l7/hess/users/mohrmann/software/anaconda3/envs/gammapy-1.0/lib/python3.9/site-packages/astropy/units/quantity.py:1249, in Quantity.__len__(self)
   1247 def __len__(self):
   1248     if self.isscalar:
-> 1249         raise TypeError("'{cls}' object with a scalar value has no "
   1250                         "len()".format(cls=self.__class__.__name__))
   1251     else:
   1252         return len(self.value)

TypeError: 'Quantity' object with a scalar value has no len()
```

This PR fixes this problem by checking the dimensionality of the array read from the FITS file.

### DIFF
--- a/gammapy/maps/axes.py
+++ b/gammapy/maps/axes.py
@@ -1,6 +1,7 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 import copy
 import inspect
+import logging
 from collections.abc import Sequence
 import numpy as np
 import scipy
@@ -15,6 +16,8 @@ from gammapy.utils.time import time_ref_from_dict, time_ref_to_dict
 from .utils import INVALID_INDEX, edges_from_lo_hi
 
 __all__ = ["MapAxes", "MapAxis", "TimeMapAxis", "LabelMapAxis"]
+
+log = logging.getLogger(__name__)
 
 
 def flat_if_equal(array):
@@ -1300,12 +1303,15 @@ class MapAxis:
             if hduclass in {"BKG", "RAD_MAX"} and column_prefix == "ENERG":
                 name = "energy"
 
-            edges_lo = table[f"{column_prefix}_LO"].quantity
-            edges_hi = table[f"{column_prefix}_HI"].quantity
-            # usually the axes arrays are stored with one extra (single-column) dimension
-            if edges_lo.ndim > 1:
-                edges_lo = edges_lo[0]
-                edges_hi = edges_hi[0]
+            edges_lo = table[f"{column_prefix}_LO"].quantity[0]
+            edges_hi = table[f"{column_prefix}_HI"].quantity[0]
+            # for a single-valued array, it can happen that the value is stored/extracted as a scalar
+            if edges_lo.isscalar:
+                log.warning(
+                    f"'{column_prefix}' axis is stored as a scalar -- converting to 1D array."
+                )
+                edges_lo = edges_lo[np.newaxis]
+                edges_hi = edges_hi[np.newaxis]
 
             if np.allclose(edges_hi, edges_lo):
                 axis = MapAxis.from_nodes(edges_hi, interp=interp, name=name)
--- a/gammapy/maps/axes.py
+++ b/gammapy/maps/axes.py
@@ -1300,8 +1300,12 @@ class MapAxis:
             if hduclass in {"BKG", "RAD_MAX"} and column_prefix == "ENERG":
                 name = "energy"
 
-            edges_lo = table[f"{column_prefix}_LO"].quantity[0]
-            edges_hi = table[f"{column_prefix}_HI"].quantity[0]
+            edges_lo = table[f"{column_prefix}_LO"].quantity
+            edges_hi = table[f"{column_prefix}_HI"].quantity
+            # usually the axes arrays are stored with one extra (single-column) dimension
+            if edges_lo.ndim > 1:
+                edges_lo = edges_lo[0]
+                edges_hi = edges_hi[0]
 
             if np.allclose(edges_hi, edges_lo):
                 axis = MapAxis.from_nodes(edges_hi, interp=interp, name=name)
--- a/gammapy/maps/tests/test_axes.py
+++ b/gammapy/maps/tests/test_axes.py
@@ -757,3 +757,9 @@ def test_time_map_axis_format_plot_xaxis(time_intervals):
 
     ax1 = axis.format_plot_xaxis(ax=ax)
     assert ax1.xaxis.label.properties()["text"] == "Time [iso]"
+
+
+def test_single_valued_axis():
+    theta_values = np.array([0.5]) * u.deg
+    table = Table(data=[theta_values, theta_values], names=["THETA_LO", "THETA_HI"])
+    _ = MapAxis.from_table(table, format="gadf-dl3", column_prefix="THETA")
--- a/gammapy/maps/tests/test_axes.py
+++ b/gammapy/maps/tests/test_axes.py
@@ -760,6 +760,13 @@ def test_time_map_axis_format_plot_xaxis(time_intervals):
 
 
 def test_single_valued_axis():
+    # this will be interpreted as a scalar value
+    # that is against the specifications, but we allow it nevertheless
     theta_values = np.array([0.5]) * u.deg
+    table = Table(data=[theta_values, theta_values], names=["THETA_LO", "THETA_HI"])
+    _ = MapAxis.from_table(table, format="gadf-dl3", column_prefix="THETA")
+
+    # this is a proper array-like axis with just a single value
+    theta_values = np.array([[0.5]]) * u.deg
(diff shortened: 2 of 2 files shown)


AtreyeeS:
Thanks @lmohrmann ! We have had this issue reported on slack often.
It would be good to add a test.

adonath:
Thanks @lmohrmann!

Now it becomes nit-picky, but GADF explicitly states that the entries need to be `ndim=1` (see e.g. https://gamma-astro-data-formats.readthedocs.io/en/v0.3/irfs/full_enclosure/aeff/index.html#effective-area-vs-true-energy), this means there is no support for scalar values but only one dimensional arrays with a single entry (and those should be supported by Gammapy already now...)

The file you linked is not compatible with GADF. However I remember there was a bug in Astropy table which stored single entry arrays as scalars. Maybe the file was written using the Astropy version affected by this bug. Maybe @maxnoe has an opinion here as well...

maxnoe:
Be lenient when parsing, strict when writing...

And yes, astropy pre I think 4.2 would write a 1d array with a single entry as a scalar into fits.

So I'd not be opposed to allowing this when reading, but then it should immediately be converted into a 1d array.

However, it's also quite simple to fix the file, just add the `TDIM<N> = (1)` keyword to the header for the corresponding column(s)

gammapy/maps/axes.py
```diff
@@ -1300,8 +1300,12 @@ def from_table(cls, table, format="ogip", idx=0, column_prefix=""):
             if hduclass in {"BKG", "RAD_MAX"} and column_prefix == "ENERG":
                 name = "energy"
 
-            edges_lo = table[f"{column_prefix}_LO"].quantity[0]
-            edges_hi = table[f"{column_prefix}_HI"].quantity[0]
+            edges_lo = table[f"{column_prefix}_LO"].quantity
```
maxnoe, line 1303: While solving the problem, the code here somewhat has the wrong semantics. GADF always stores the IRFs as a single row in a binary table, so to get the first (and until now only) IRF GADF file, `[0]` is correct, so I'd not change this here.

You want to add support for the case that the content of this row is not an array, but a single number. So instead of changing the two lines above and then checking for `ndim > 1`, check for a scalar and upcast if needed:

```
if edges_lo.shape == ():
    edges_lo = edges_lo[np.newaxis]
    edges_hi = edges_hi[np.newaxis]

```
You could also add a warning there.

Simpler would be just doing `edges_lo = np.atleast_1d(table[f"{column_prefix}_LO"].quantity[0])`, but then you can't issue the warning.

maxnoe (changes requested):

gammapy/maps/axes.py
lmohrmann: Good point - I've changed the code and added a warning (although this is now the only place in the entire file where a warning is being issued...)

lmohrmann:
Thanks for the feedback! I've changed the code according to @maxnoe's suggestion, and also added a test as suggested by @AtreyeeS.
Note that in the test I create a table with an array, but the original code still fails (and this is with Astropy 5.1.1. I believe).

maxnoe (approved):

adonath (approved):
Thanks @lmohrmann, no further comments from my side either.

lmohrmann:
Great 👍 
Do you want me to do the merging, or should I leave that to you?

adonath:
> Do you want me to do the merging, or should I leave that to you?

Please leave it open for now. 

We still have to decide whether this get's a back-port to v1.0.x. What do you think @registerrier, I guess we could consider it a bug as well, it is a backwards compatible change...

gammapy/maps/tests/test_axes.py
```diff
@@ -757,3 +757,9 @@ def test_time_map_axis_format_plot_xaxis(time_intervals):
 
     ax1 = axis.format_plot_xaxis(ax=ax)
     assert ax1.xaxis.label.properties()["text"] == "Time [iso]"
+
+
+def test_single_valued_axis():
+    theta_values = np.array([0.5]) * u.deg
```
maxnoe, line 763: Ah, here we are even mixing more things! You have a 1d array with one element which you pass to `Table`, which for astropy means "a column with a single row of scalar value".

This is the case that strictly speaking is not GADF compatible, since the format requires it to be an array (that could have only one element).

Could you also add the test for the compatible case of a single element? I.e. `theta_values = np.array([[0.5]]) * u.deg`?

lmohrmann: > which for astropy means "a column with a single row of scalar value".

I guess I hadn't realised this. In any case, I've expanded the test such that it also tests for the specs-compliant case. OK like so?

maxnoe: Yes, perfect

registerrier:
> > Do you want me to do the merging, or should I leave that to you?
> 
> Please leave it open for now.
> 
> We still have to decide whether this get's a back-port to v1.0.x. What do you think @registerrier, I guess we could consider it a bug as well, it is a backwards compatible change...

Yes I think we can consider this is a bug fix. Let's add the flag and relevant milestone.

registerrier (approved):
Thanks @lmohrmann ! No comment from my side.

registerrier:

PR #4220, opened 2022-11-25
New API for FixedPointingInfo, decouple from GADF metadata

maxnoe:
**Description**

This pull request is here to further work on #3767 and to move towards a API of the pointing information that is easier to reason about.

The properties / methods that are not directly related to pointing are deprecated (and should be accessed directly from `Observation` or the `EventList` etc.

**Dear reviewer**

### DIFF
--- a/gammapy/data/observations.py
+++ b/gammapy/data/observations.py
@@ -314,16 +314,12 @@ class Observation:
             self._pointing = FixedPointingInfo.from_fits_header(self.obs_info)
         return self._pointing
 
-    def get_pointing_altaz(self, time=None):
+    def get_pointing_altaz(self, time):
         """Get the pointing in altaz for given time"""
-        if time is None:
-            time = self.tmid
         return self.pointing.get_altaz(time, self.observatory_earth_location)
 
-    def get_pointing_icrs(self, time=None):
+    def get_pointing_icrs(self, time):
         """Get the pointing in icrs for given time"""
-        if time is None:
-            time = self.tmid
         return self.pointing.get_icrs(time, self.observatory_earth_location)
 
     @lazyproperty
--- a/gammapy/data/observations.py
+++ b/gammapy/data/observations.py
@@ -220,9 +220,11 @@ class Observation:
         )
 
         if not isinstance(pointing, FixedPointingInfo):
-            warnings.warn("Pointing will be required to be provided as FixedPointingInfo", DeprecationWarning)
+            warnings.warn(
+                "Pointing will be required to be provided as FixedPointingInfo",
+                DeprecationWarning,
+            )
             pointing = FixedPointingInfo.from_gadf_header(obs_info)
-
 
         return cls(
             obs_id=obs_id,
--- a/gammapy/data/observations.py
+++ b/gammapy/data/observations.py
@@ -18,6 +18,7 @@ from gammapy.utils.fits import LazyFitsData, earth_location_to_dict
 from gammapy.utils.scripts import make_path
 from gammapy.utils.testing import Checker
 from gammapy.utils.time import time_ref_to_dict, time_relative_to_ref
+from gammapy.utils.deprecation import GammapyDeprecationWarning
 from .event_list import EventList, EventListChecker
 from .filters import ObservationFilter
 from .gti import GTI
@@ -222,9 +223,9 @@ class Observation:
         if not isinstance(pointing, FixedPointingInfo):
             warnings.warn(
                 "Pointing will be required to be provided as FixedPointingInfo",
-                DeprecationWarning,
+                GammapyDeprecationWarning,
             )
-            pointing = FixedPointingInfo.from_gadf_header(obs_info)
+            pointing = FixedPointingInfo.from_fits_header(obs_info)
 
         return cls(
             obs_id=obs_id,
@@ -483,7 +484,7 @@ class Observation:
             gti=gti,
             obs_info=obs_info,
             obs_id=obs_info.get("OBS_ID"),
-            pointing=FixedPointingInfo.from_gadf_header(obs_info),
+            pointing=FixedPointingInfo.from_fits_header(obs_info),
             **irf_dict,
         )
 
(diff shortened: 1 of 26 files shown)


adonath:
Thanks a lot @maxnoe! I haven't looked into any details here (can do this later...). However in general I was wondering whether you have thought about using on a `pydantic` model for defining the meta data model. It will solve the validation and type declarations...

maxnoe:
@registerrier @adonath 

The tests are passing now in principle. The errors are due to the deprecation warnings of all that I have deprecated and are mainly occuring in the test setup code, not gammapy itself.

Before I go start fixing the 90+ deprecation warnings this is raising now, could you please review the API changes and say if you agree with them?


Main changes are:

* the new API of FixedPointingInfo (if you ignore the deprecated parts) is essentially:
```
class FixedPointingInfo:
    mode: PointingMode
    fixed_icrs : None | SkyCoord[ICRS]
    fixed_altaz : None | SkyCoord[AltAz]

    def get_icrs(self, time, location) -> SkyCoord[AltAz]
    def get_altz(self, time, location) -> SkyCoord[AltAz]
    def from_fits_header(cls, format, version) -> FixedPointingInfo
    def to_fits_header(self, format, versoin) -> fits.Header
    def read(cls, path) -> FixedPointingInfo
```

All other parts are deprecated as they are either suboptimal (don't force giving a valid location / obstime) or have nothing to do with pointing (tstart, tstop, obstime, location).


I added pointing as a data member to `Observation` and reading it using a `LazyFits` using the event header.

I also added `get_pointing_icrs` and `get_pointing_altaz` to `Observation`, which are very quick shorthands, just taking the `location` from observation and by default the `mid of run time`. This is the current use in gammapy, but there are several open issues connected to this.

The other pointing related properties of Observation are now also deprecated to make it clear that one should access `Observation.pointing` or `Observation.get_pointing_{altaz, icrs}`.

Questions I have:

* Do we want to  fallback to reading pointing info from IRF HDUs if there is no events HDU? Or is that a case where users have to manually supply a pointing?
* Do we want to require the `time` argument in `Observation.get_pointing_{altaz,icrs}` explicitly? This would make it clear that a time is required and that using the "mid of observation time" really is a (over-) simplification.

registerrier (commented):
Thanks @maxnoe. We just discussed the new API and it looks fine.

maxnoe:
@adonath @registerrier Tests should be passing now

gammapy/data/observations.py
```diff
@@ -491,7 +542,9 @@ def write(self, path, overwrite=False, format="gadf", include_irfs=True):
 
         events = self.events
         if events is not None:
-            hdul.append(events.to_table_hdu(format=format))
+            events_hdu = events.to_table_hdu(format=format)
```
bkhelifi, line 545: I have a question that might be related to an other PR.
As for each 'format' the versions can evolve, should be 'extended' the parameters list of the `to_table_hdu` function with a version ID?
The version of the format is written into the DL3 meta data and this information should be kept into the internal meta data such that it can be re-used.

maxnoe: Yes, we should

maxnoe: Relevent issue #4239

gammapy/data/observations.py
```diff
@@ -286,29 +308,56 @@ def obs_info(self):
...
+        message="Use observation.pointing or observation.get_pointing_{{altaz,icrs}} instead",
+    )
     def fixed_pointing_info(self):
         """Fixed pointing info for this observation (`FixedPointingInfo`)."""
-        return FixedPointingInfo(self.obs_info)
+        return self._pointing
 
     @property
+    @deprecated("v1.1", message="Use observation.get_pointing_icrs(time) instead")
     def pointing_radec(self):
```
adonath, line 336: For convenience one could  still introduce `obs.pointing_icrs_tmid`. Otherwise many people will end up writing `obs.get_pointing_icrs(obs.tmid)` it is not that bad, but the former is still better readable...

gammapy/data/pointing.py
```diff
@@ -67,45 +70,194 @@ class FixedPointingInfo:
...
+            raise ValueError(f'Only the "gadf" format supported, got {format}')
+
+        if version not in {"0.2", "0.3"}:
+            raise ValueError(f"Unsupported version {version} for format {format}")
+
+        if self.mode == PointingMode.DRIFT and version == "0.2":
+            raise ValueError("mode=DRIFT is only supported by GADF 0.3")
+
+        header = fits.Header()
+        if self.mode == PointingMode.POINTING:
```
adonath, line 233: Could move this code to the `PointingMode` class directly...would lead to a shorter implementation here...

gammapy/data/pointing.py
```diff
@@ -67,45 +70,194 @@ class FixedPointingInfo:
...
+            location=location,
+            fixed_icrs=fixed_icrs,
+            fixed_altaz=pointing_altaz,
+            time_start=time_start,
+            time_stop=time_stop,
+            time_ref=time_ref,
+            legacy_altaz=legacy_altaz,
+        )
+
+    def to_fits_header(self, format="gadf", version="0.3", time_ref=None):
```
adonath, line 222: Same here: if the method is public please add a docstring...

adonath (commented):
Thanks @maxnoe, I have left some more minor comments...

gammapy/data/observations.py
maxnoe: Actually, my reasoning was to not make it convenient to assume the pointing doesn't change. People should get the "oh right, it might be time-dependent" when trying to get the pointing.

gammapy/data/pointing.py
maxnoe: What could be moved? I don't see what part of the code is dealing with the `PonitingMode` itself, the `PointingMode` just decides which information stored in `PointingInfo` is put into the header.

adonath: Oh yes, you are right. As we already have `PointingMode.from_gadf_string()` I thought it might make sense to have `PointingMode.to_fits_header()` as well, but it would contain effectively only two lines of code and the returned header would be barely useful. So fine as is...

gammapy/data/pointing.py
maxnoe: Same? Is there another one missing? 

I added one for this method now.

maxnoe: Ah, from_fits_header, fixing now

maxnoe:
Ok, this should be it: comments addressed and rebased

registerrier (approved):
Thanks @maxnoe . No further comment from my side.

PR #4225, opened 2022-12-01
Replace deprecated numpy type aliases

QRemy:
Replace deprecated numpy type aliases.
See https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations


registerrier (dismissed):
thanks @QRemy ! This looks good and should be backported to ensure v1.0 will stay compatible with numpy>=1.24

We could probably remove the dev build from the CI matrix, to avoid failing all tests if the CI does not work for the future versions. It could be run independently. (Possibly independently of a PR)

registerrier (approved):
Thanks @QRemy

PR #4226, opened 2022-12-01
Fix inhomogeneous array shape deprecation

QRemy:
Fix `SpectralModel._propagate_error` method to avoid passing array with inhomogeneous shape to Quantity which is now deprecated :
"VisibleDeprecationWarning: Creating an ndarray from ragged nested sequences (which is a list-or-tuple of lists-or-tuples-or ndarrays with different lengths or shapes) is deprecated."


registerrier (commented):
Thanks @QRemy 

Maybe it would be simpler to make all the changes in the same PR as they all have to be backported to stay consistent with future numpy as astropy versions.

registerrier (approved):
Thanks @QRemy

PR #4231, opened 2022-12-02
Fix bug in safe mask computation for SpectrumDatasetOnOff

lmohrmann:
**Description**
This PR fixes a bug in the safe mask computation, specifically for the `edisp-bias` method. The edges of an energy axis were passed instead of the axis itself.
A test for this bug has been added as well.


QRemy (approved):
Thanks @lmohrmann, no comments from my side.

PR #4236, opened 2022-12-05
Make GammapyDeprecationWarnings and AstropyDepecrationWarning errors in tests

maxnoe:
Signed-off-by: Maximilian Linhoff <maximilian.linhoff@tu-dortmund.de>

**Description**

See title

**Dear reviewer**

### DIFF
--- a/gammapy/catalog/tests/test_core.py
+++ b/gammapy/catalog/tests/test_core.py
@@ -25,7 +25,7 @@ def make_test_catalog():
 
 
 class TestSourceCatalog:
-    def setup(self):
+    def setup_method(self):
         self.cat = make_test_catalog()
 
     def test_str(self):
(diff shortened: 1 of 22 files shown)


registerrier:
Thanks @maxnoe ! We have obviously some effort to make to get rid of deprecations.

What is the best approach here? Should it be part of the main CI (run on each PR), or could be run less frequently?

maxnoe:
It doesn't seem much, the CI has 4 failures only If I saw it correctly (at least in the first failing test). I will see if I can fix those.

I think we should have these options enabled for all CI runs. Where those happen in third party code, we can add the `pytest.warns` and open issues with the third party

QRemy (approved):
Thanks @maxnoe for the changes.

PR #4245, opened 2022-12-07
Added an evaluate method for CompoundSpectralModel

lgreaux:
Signed-off-by: Lucas Gréaux <lucas.greaux@ijclab.in2p3.fr>

This pull request adds an evaluate method to the CompoundSpectralModel. This method allows to evaluate spectral models without having to modify the values of the parameters. This could be a first step towards the evaluation of spectral models on numpy arrays, which would greatly enhance the speed at which multiple evaluations are performed.

### DIFF
--- a/gammapy/modeling/models/spectral.py
+++ b/gammapy/modeling/models/spectral.py
@@ -3,6 +3,7 @@
 import logging
 import operator
 import os
+from pathlib import Path
 import numpy as np
 import scipy.optimize
 import scipy.special
@@ -12,7 +13,6 @@ from astropy.table import Table
 from astropy.utils.decorators import classproperty
 from astropy.visualization import quantity_support
 import matplotlib.pyplot as plt
-from pathlib import Path
 from gammapy.maps import MapAxis, RegionNDMap
 from gammapy.modeling import Parameter, Parameters
 from gammapy.utils.integrate import trapz_loglog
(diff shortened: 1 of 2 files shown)


QRemy (approved):
Thanks, no comments from my side.

adonath:
Thanks @lgreaux! Can you quickly comment on the concern of the performance? Spectral models are already evaluated on numpy arrays...or do you mean arrays for parameters?

lgreaux:
Hello @adonath, as @QRemy explained, the use case is to evaluate the model on parameters that are themselves arrays. I want ot evaluate spectral models on multiple points in the parameter space (around 100 000 points).

Currently, the only way to proceed is, for each point in the parameter space, to manually change the values of the parameters, evaluate the spectral model by calling it on the energy, and then to gather the evaluation.

Thanks to this PR, it is now possible to evaluate a CompoundSpectralModel without changing the parameters values. Using broadcasting, it is now possible to evaluate arrays of parameters, which greatly improves the speed of this evaluation.

You will find bellow a Jupyter Notebook working with GammaPy 1.0 implementing both ways of evaluating. On my machine, using the evaluate method from the PR runs 100 times faster than calling the model. 

[https://github.com/lgreaux/gammapy/blob/Notebook_CompoundSpectralModel/Evaluate%20method%20on%20CompoundSpectralModel.ipynb](https://github.com/lgreaux/gammapy/blob/Notebook_CompoundSpectralModel/Evaluate%20method%20on%20CompoundSpectralModel.ipynb)

PR #4252, opened 2022-12-09
Map dataset on off in phase maker

MRegeard:
**Description**

This pull request modify the PhaseBackgroundMaker to support a MapDatasetOnOff dataset. This is useful to be able to produce pulsed map. 
I also modify the test_phase file to be able to test the run method of the maker in the case of a MapDatasetOnOff.
Because of the current handling of the phase in the data (namely adding column(s) to the DL3 event file), I also added a column_name ("PHASE" by default) in the init to be able to choose the phase column if there are multiples. For this I put it in the init of the maker, but I am not sure that it should be here or in the run method directly. 
I was also wandering whether it is a good idea to put a method that check if the phase column is non-zeros. This thought is mainly motivated by the fact that in HESS DL3 event file, a PHASE column exist by default but is filled with zeros. Is this a relevant feature or do we consider that it is the responsibility of the user to check it. 
I haven't checked what happens if you give a column name that does not exist in the event file. My guess would be that an error occur, but that's probably something that I can check. 

**Dear reviewer**

### DIFF
--- a/gammapy/makers/background/phase.py
+++ b/gammapy/makers/background/phase.py
@@ -7,7 +7,7 @@ from gammapy.datasets import (
     SpectrumDataset,
     SpectrumDatasetOnOff,
 )
-from gammapy.maps import RegionNDMap, WcsNDMap
+from gammapy.maps import Map
 from ..core import Maker
 
 __all__ = ["PhaseBackgroundMaker"]
@@ -43,20 +43,17 @@ class PhaseBackgroundMaker(Maker):
         return s
 
     @staticmethod
-    def _make_counts(self, dataset, observation, phases):
+    def _make_counts(dataset, observation, phases, column_name):
 
         event_lists = []
         for interval in phases:
             events = observation.events.select_parameter(
-                parameter=self.column_name, band=interval
+                parameter=column_name, band=interval
             )
             event_lists.append(events)
 
         events = EventList.from_stack(event_lists)
-        if isinstance(dataset, SpectrumDataset):
-            counts = RegionNDMap.from_geom(dataset.counts.geom)
-        elif isinstance(dataset, MapDataset):
-            counts = WcsNDMap.from_geom(dataset.counts.geom)
+        counts = Map.from_geom(dataset.counts.geom)
         counts.fill_events(events)
         return counts
 
@@ -75,7 +72,7 @@ class PhaseBackgroundMaker(Maker):
         counts_off : `RegionNDMap`
             Off counts.
         """
-        return self._make_counts(self, dataset, observation, self.off_phase)
+        return self._make_counts(dataset, observation, self.off_phase, self.column_name)
 
     def make_counts(self, dataset, observation):
         """Make on counts.
@@ -92,7 +89,7 @@ class PhaseBackgroundMaker(Maker):
         counts : `RegionNDMap`
             On counts.
         """
-        return self._make_counts(self, dataset, observation, self.on_phase)
+        return self._make_counts(dataset, observation, self.on_phase, self.column_name)
 
     def run(self, dataset, observation):
         """Run all steps.
@@ -112,13 +109,13 @@ class PhaseBackgroundMaker(Maker):
         counts_off = self.make_counts_off(dataset, observation)
         counts = self.make_counts(dataset, observation)
 
+        acceptance = Map.from_geom(geom=dataset.counts.geom)
+        acceptance.data = np.sum([_[1] - _[0] for _ in self.on_phase])
+
+        acceptance_off = Map.from_geom(geom=dataset.counts.geom)
+        acceptance_off.data = np.sum([_[1] - _[0] for _ in self.off_phase])
+
         if isinstance(dataset, SpectrumDataset):
-            acceptance = RegionNDMap.from_geom(geom=dataset.counts.geom)
-            acceptance.data = np.sum([_[1] - _[0] for _ in self.on_phase])
-
-            acceptance_off = RegionNDMap.from_geom(geom=dataset.counts.geom)
-            acceptance_off.data = np.sum([_[1] - _[0] for _ in self.off_phase])
-
             dataset_on_off = SpectrumDatasetOnOff.from_spectrum_dataset(
                 dataset=dataset,
                 counts_off=counts_off,
@@ -129,12 +126,6 @@ class PhaseBackgroundMaker(Maker):
             return dataset_on_off
 
         elif isinstance(dataset, MapDataset):
-            acceptance = WcsNDMap.from_geom(geom=dataset.counts.geom)
-            acceptance.data = np.sum([_[1] - _[0] for _ in self.on_phase])
-
-            acceptance_off = WcsNDMap.from_geom(geom=dataset.counts.geom)
-            acceptance_off.data = np.sum([_[1] - _[0] for _ in self.off_phase])
-
             dataset_on_off = MapDatasetOnOff.from_map_dataset(
                 dataset=dataset,
                 counts_off=counts_off,
--- a/gammapy/makers/background/phase.py
+++ b/gammapy/makers/background/phase.py
@@ -1,12 +1,7 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 import numpy as np
 from gammapy.data import EventList
-from gammapy.datasets import (
-    MapDataset,
-    MapDatasetOnOff,
-    SpectrumDataset,
-    SpectrumDatasetOnOff,
-)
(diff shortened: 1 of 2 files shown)


gammapy/makers/background/phase.py
```diff
@@ -19,32 +24,39 @@ class PhaseBackgroundMaker(Maker):
...
     def __str__(self):
         s = self.__class__.__name__
         s += f"\n{self.on_phase}"
         s += f"\n{self.off_phase}"
+        s += f"\n{self.column_name}"
         return s
 
     @staticmethod
-    def _make_counts(dataset, observation, phases):
+    def _make_counts(self, dataset, observation, phases):
```
AtreyeeS, line 46: pass the column name instead of accessing `self`

gammapy/makers/background/phase.py
```diff
@@ -100,20 +112,37 @@ def run(self, dataset, observation):
...
 
-        acceptance = RegionNDMap.from_geom(geom=dataset.counts.geom)
-        acceptance.data = np.sum([_[1] - _[0] for _ in self.on_phase])
+        if isinstance(dataset, SpectrumDataset):
+            acceptance = RegionNDMap.from_geom(geom=dataset.counts.geom)
+            acceptance.data = np.sum([_[1] - _[0] for _ in self.on_phase])
 
-        acceptance_off = RegionNDMap.from_geom(geom=dataset.counts.geom)
-        acceptance_off.data = np.sum([_[1] - _[0] for _ in self.off_phase])
+            acceptance_off = RegionNDMap.from_geom(geom=dataset.counts.geom)
```
AtreyeeS, line 119: Same as before. No need for the `if`, create a `Map` directly

gammapy/makers/background/phase.py
```diff
@@ -100,20 +112,37 @@ def run(self, dataset, observation):
...
 
-        dataset_on_off = SpectrumDatasetOnOff.from_spectrum_dataset(
-            dataset=dataset,
-            counts_off=counts_off,
-            acceptance=acceptance,
-            acceptance_off=acceptance_off,
-        )
-        dataset_on_off.counts = counts
-        return dataset_on_off
+            dataset_on_off = SpectrumDatasetOnOff.from_spectrum_dataset(
```
AtreyeeS, line 122: Maybe simplify a bit by
```
dataset_on_off = MapDatasetOnOff.from_spectrum_dataset(
             dataset=dataset,
             counts_off=counts_off,
             acceptance=acceptance,
             acceptance_off=acceptance_off,
         )

if `dataset._geom.is_region:
     dataset_on_off = dataset_on_off.to_spectrum_dataset(dataset._geom.region)

return dataset_on_off
```

gammapy/makers/background/phase.py
```diff
@@ -1,8 +1,13 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 import numpy as np
 from gammapy.data import EventList
-from gammapy.datasets import SpectrumDatasetOnOff
-from gammapy.maps import RegionNDMap
+from gammapy.datasets import (
```
AtreyeeS, line 4: these imports are not necessary

AtreyeeS (changes requested):
Thank you @MRegeard !

This looks good. I have left some inline comments to decrease the lines of code a bit.
I like the idea of passing the name of the phase column - the present logic is good, it should happen on `init`

I do not think that we should worry about zero or other invalid values in the phase column at present. Creating this column is rather specific to analysis of a specific pulsar, and at this point it is the users' responsibility to ensure that the phasing is done correctly.

gammapy/makers/background/phase.py
MRegeard: Hi @AtreyeeS,
I have made all the modifications that you suggest apart from this one. 
If I understand correctly what you suggest, it means that in both case (SpectrumonOff or MapOnOff), the user has to pass a MapDataset to the maker? (I guessed that it MapDatasetOnOff.from_map_dataset([...]) and not MapDatasetOnOff.from_spectrum_dataset([...])).
If I understand correctly, I'm not sure that it is very intuitive for the user to build a MapDataset and then to have the maker transform it into a SpectrumDataset if the user wants to do a SpectrumDataset from the start. 
Let me know

AtreyeeS: Hi @MRegeard ! Yes indeed, it should be `MapDatasetOnOff.from_map_dataset` as you correctly pointed out. Sorry! 

The user does not need to create a `MapDataset` explicitly.
`MapDatasetOfOff.xyz()` functionalities are available on `SpectrumDataset` due to inheritance. So, `MapDatasetOnOff.from_map_dataset` will create a MapDatasetOnOff. Now, if dataset was initially a spectral one to start off with, gammapy will change to a spectral datasetonoff

AtreyeeS: Thinking a bit about this, I think it is better to change 

```
if dataset._geom.is_region:
     dataset_on_off = dataset_on_off.to_spectrum_dataset(dataset._geom.region)
```

to 

```
if isInstance(dataset, SpectrumDataset):
    dataset_on_off = dataset_on_off.to_spectrum_dataset(dataset._geom.region)
```

MRegeard: Oh I see, I always forget about inheritance...
Changes were made !

AtreyeeS (dismissed):
Thanks @MRegeard ! No further comments from my side

gammapy/makers/background/phase.py
```diff
@@ -19,32 +19,36 @@ class PhaseBackgroundMaker(Maker):
         on-phase defined by the two edges of each interval (edges are excluded)
     off_phase : `tuple` or list of tuples
         off-phase defined by the two edges of each interval (edges are excluded)
+    column_name : `str`
+        The name of the column in the event file from which the phase informations are extracted
```
registerrier, line 23:
```suggestion
        The name of the column in the event file from which the phase informations are extracted. Default is 'PHASE'.
```

gammapy/makers/background/phase.py
```diff
@@ -19,32 +19,36 @@ class PhaseBackgroundMaker(Maker):
...
+    def __init__(self, on_phase, off_phase, column_name="PHASE"):
         self.on_phase = self._check_intervals(on_phase)
         self.off_phase = self._check_intervals(off_phase)
+        self.column_name = column_name
 
     def __str__(self):
         s = self.__class__.__name__
         s += f"\n{self.on_phase}"
         s += f"\n{self.off_phase}"
+        s += f"\n{self.column_name}"
```
registerrier, line 37: Note that the `__str__` method could be more talkative.

gammapy/makers/background/phase.py
```diff
@@ -19,32 +19,36 @@ class PhaseBackgroundMaker(Maker):
...
 
-    def __init__(self, on_phase, off_phase):
+    def __init__(self, on_phase, off_phase, column_name="PHASE"):
         self.on_phase = self._check_intervals(on_phase)
         self.off_phase = self._check_intervals(off_phase)
+        self.column_name = column_name
 
     def __str__(self):
         s = self.__class__.__name__
         s += f"\n{self.on_phase}"
```
registerrier, line 35:
```suggestion
        s += f"\nON phase interval : {self.on_phase}"
```

gammapy/makers/background/phase.py
registerrier, line 37:
```suggestion
        s += f"\nOn phase interval : {self.on_phase}"
        s += f"\nOff phase interval : {self.off_phase}"
        s += f"\nPhase column name : {self.column_name}"
```

registerrier (approved):

I have left some inline suggestions. 

Once this is merged, the associated tutorial should be updated.

gammapy/makers/background/phase.py
```diff
@@ -19,32 +19,36 @@ class PhaseBackgroundMaker(Maker):
         on-phase defined by the two edges of each interval (edges are excluded)
     off_phase : `tuple` or list of tuples
         off-phase defined by the two edges of each interval (edges are excluded)
+    column_name : `str`
```
adonath, line 22: Rename to the even more explicit `phase_column_name` or `column_name_phase`?

MRegeard:
@registerrier changes done !
Yes the update of the tutorial is planned.

gammapy/makers/background/phase.py
MRegeard: Yes I guess it would be better.

registerrier (approved):
Thanks @MRegeard !

registerrier:
The CI fail is unrelated. Merging now.

PR #4253, opened 2022-12-10
PIG 25 - Unbinned analysis

tunbehaun273:
**Description**

This pull request will add a pig for the unbinned analysis

**Dear reviewer**

### DIFF
--- a/docs/development/pigs/pig-025.rst
+++ b/docs/development/pigs/pig-025.rst
@@ -16,7 +16,7 @@ PIG 025 - Unbinned Dataset
 Abstract
 ========
 
-Our goal is to be able to perform an unbinned analysis using gammapy.
+Our goal is to be able to perform an unbinned analysis using Gammapy.
 To this end we like to propose to add a new EventDataset class with a dedicated fit
 statistic and an EventDatasetEvaluator to compute the differential model prediction (flux) at each event's reconstructed coordinates.
 
--- a/docs/development/pigs/pig-025.rst
+++ b/docs/development/pigs/pig-025.rst
@@ -83,7 +83,7 @@ Class requirements
        * "-" Cannot inherit from MapDataset (might complicate stacking)
        
 * Want to store and evaluate models using an UnbinnedEvaluator. 
-* Which "convenience fucntions" similar to the ``MapDataset``: 
+* Which "convenience functions" similar to the ``MapDataset``: 
 
   * copy: Can be inherited?
   * create: Modification needed wrt ``MapDataset``
--- a/docs/development/pigs/pig-025.rst
+++ b/docs/development/pigs/pig-025.rst
@@ -43,11 +43,11 @@ The EventDataset still contains all the reconstructed properties of the events (
 ****************************************
 ****************************************
 
-* Spectral analysis with narrow features
-* Pulsar analysis
-* Flare detection
-* Time variablity
-* Energy-temporal analysis
+* spectral analysis with narrow features,
+* pulsar analysis,
+* flare detection,
+* time variablity,
+* energy-temporal analysis.
 * ..
 
 
--- a/docs/development/pigs/pig-025.rst
+++ b/docs/development/pigs/pig-025.rst
@@ -50,21 +50,77 @@ The EventDataset still contains all the reconstructed properties of the events (
 * Energy-temporal analysis
 * ..
 
-Implementation
-==============
-EventDataset:
 
-**********************************
-**********************************
+Class requirements 
+==================
+* Individual event information (position, energy, time); therefore the ``EventList`` seems to be a good choice
+* IRF information (should support time dependence) 
+  General requirement: Memory usage of the IRFs should not be too big, e.g. we only want to store the IRFs at the resolution of the instrument. Also, building of the kernel should be fast for many events and fine integration grids. 
+  Open question: Do we want projected IRFs?
+  Pros/Cons: 
+  
+  * "+" Could use existing classes
+  * "+" Memory consumption is under control
+  * "-" Would require binning
+  * "-" Implementation of time dependence is not straightforward
+  * "-" Information loss due to interpolations
+        
+  In case we use projected IRFs we should only support the ``EDispMap`` and ``PSFMap`` and not the "kernel" version of those for simplicity and precision
+  
+  Alternatives:
+  
+  1. Event-wise IRFs: Interpolate unprojected IRFs to the event coordinates
+       * "+" Processing for the UnbinnedEvaluator would be fast
+       * "+" No information loss 
+       * "-" Classes would need to be implemented
+       * "-" Might be too memory intensive esp. in cases of many events that are close to each other (that could use the same binned IRF)
+        
+  2. Unprojected IRFs: Information of the observation
+       * "+" No information loss
+       * "+" Classes exist
+       * "+" Fast to build the Dataset
+       * "-" Slow to build the kernels for each event
+       * "-" Cannot inherit from MapDataset (might complicate stacking)
+       
+* Want to store and evaluate models using an UnbinnedEvaluator. 
+* Which "convenience fucntions" similar to the ``MapDataset``: 
 
-* DL4 (Eventlist + projected IRFs)
-* We need a maker class
-* Models
-* unbinned likelihood (stat_sum)
-`$-2 \\log \\mathcal{L} =  2 N_{pred} - 2 \\sum_{i} \\log \\phi( E_i, \\vec{r}_i )$`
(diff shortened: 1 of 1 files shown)


docs/development/pigs/pig-025.rst
```diff
@@ -0,0 +1,96 @@
+.. include:: ../../references.txt
+
+.. _pig-xxx:
+
+
+**********************************
```
bkhelifi, line 6: the lenght of the '***' should exactly match the text lenght, if I do not mistaken (rst formatting).
The same applies below for the '===='

tunbehaun273: I tried to change the length of === and *** but found no noticable difference in the appearance of the file.

docs/development/pigs/pig-025.rst
```diff
@@ -0,0 +1,96 @@
...
+* Accepted: -
+* Status: draft
+* Discussion: 
+
+Abstract
+========
+
+Our goal is to be able to perform an unbinned analysis using gammapy.
+To this end we like to propose to add a new EventDataset class with a dedicated fit
+statistic and an EventDatasetEvaluator to compute the differential model flux at each event's position.
```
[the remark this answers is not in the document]
tunbehaun273: I changed the text

adonath:
A few notes from the discussion today:

- Prototype is available here https://github.com/gammapy/gammapy-unbinned-analysis/tree/main/EventDataset
- In the current implementation the `EventDataset` is very similar to the `MapDataset`, to avoid code duplication the PIG should sketch how the `EventMapDataset` could e.g. inherit from `MapDataset`:
   - Which methods would be supported? E.g. `EventMapDataset.to_image()`, `EventMapDataset.cutout()` etc.
   - Which methods would be different and how? E.g. `EventMapDataset.npred()`, also document return types, such as whether the method returns a `np.ndarray`, `Quantity` or maybe `MapCoord`
   - Which methods would just raise an error?
   - Is stacking supported?
   - For simplicity and precision we should just support the `EDispMap` and `PSFMap` and not the "kernel" version of those.
- The `EventMapDataset` could compute a counts maps "on the fly", this is convenient for visualization.
- Clarify whether we need a dedicated `EventMapDatasetMaker` (probably not!)
- The most complex code in the prototype is [here](https://github.com/gammapy/gammapy-unbinned-analysis/blob/main/EventDataset/utils.py) and [here](https://github.com/gammapy/gammapy-unbinned-analysis/blob/main/EventDataset/UnbinnedEvaluator.py). We should have an idea on how to simplify it and make it more re-usable. Here are some ideas:
    - The utility computes "per event" IRFs. It could make sense to introduce a separate abstraction for this, like `EventPSF`, or `EventEDisp` with an indexed `MapAxis` or `LabelMapAxis`  to handle the event axis. But this might be tricky in many ways...
    - One place where we already use a per event IRF is in https://github.com/gammapy/gammapy/blob/main/gammapy/irf/psf/map.py#L285 and https://github.com/gammapy/gammapy/blob/main/gammapy/irf/edisp/map.py#L193, we should check if and how this could be re-used.
- The prototype introduce the concept of and "acceptance" map, which is different from what we have e.g. in `MapDatasetOnOff.acceptance` this should be resolved.
- I think the alternative describe an `EventDataset` with IRFs stored per event, but list the reasons why this is impractical (@tunbehaun273 has a lot of experience, which should be documented somewhere)
- @registerrier also  raised the point pf whether the `EventMapDataset` is the more fundamental abstraction (to be clarified)
- @tunbehaun273 and @giacomodamico24 agreed to continue writing the PIG and the we will continue discussion in this PR.

docs/development/pigs/pig-025.rst
```diff
@@ -0,0 +1,159 @@
...
+* Author: Giacomo D'Amico, Julia Djuvsland, Tim Unbehaun (in alphabetical order)
+* Created: May 9, 2022
+* Accepted: -
+* Status: draft
+* Discussion: 
+
+Abstract
+========
+
+Our goal is to be able to perform an unbinned analysis using gammapy.
```
bkhelifi, line 19:
```suggestion
Our goal is to be able to perform an unbinned analysis using Gammapy.
```

docs/development/pigs/pig-025.rst
```diff
@@ -0,0 +1,159 @@
...
+they can be useful when very narrow features of a spectrum are expected. Then an unbinned 
+fit can provide more information than a binned one, as the latter one can't be more precise 
+than the bin width. Admittedly, this advantage is small when narrow bins are chosen especially
+as the IRFs are binned.
+Secondly, when performing time analyses the instruments response can be assumed to be perfect. 
+So all features of the light curve can be fully taken into account with the unbinned analysis.  
+Thirdly, using an unbinned data set can save computing time in case of low event numbers.
+While the computing time is dependent on the number of bins of the binned dataset, 
+it is dependent on the number of events in the unbinned case. Introducing unbinned versions of the 
+currently existing gammapy datasets (as we propose here) therefore gives the user the freedom to
```
bkhelifi, line 35:
```suggestion
currently existing Gammapy datasets (as we propose here) therefore gives the user the freedom to 
```

docs/development/pigs/pig-025.rst
```diff
@@ -0,0 +1,159 @@
...
+The EventDataset still contains all the reconstructed properties of the events (energy, position, time) and can therefore be used for
+
+****************************************
+****************************************
+
+* Spectral analysis with narrow features
+* Pulsar analysis
+* Flare detection
+* Time variablity
+* Energy-temporal analysis
```
bkhelifi, line 50:
```suggestion
* spectral analysis with narrow features,
* pulsar analysis,
* flare detection,
* time variablity,
* energy-temporal analysis.
```

docs/development/pigs/pig-025.rst
```diff
@@ -0,0 +1,159 @@
...
+  * stat_array + stat_sum: Modification needed wrt ``MapDataset``. Use the unbinned likelihood (see below). ``stat_array`` might not make sense because of the 0-dimensional Npred term.
+  * to_image: only for visualisation
+  * to_masked: Modification needed wrt ``MapDataset`` and would require stack to be implemented. 
+  * to_region_map_dataset/ to_spectrum_dataset: Want to have links to binned datasets as well as 1D datasets. 
+    
+  All in all, many methods are useful but need adaptation. Need to discuss how to avoid code duplication while maintaining transparency. 
+  
+
+(Current) Implementation
+==============
```
bkhelifi, line 111:
```suggestion
=====================
```

adonath:
Thanks @tunbehaun273 and @giacomodamico24 for the update! I think the only way to reduce the code duplication would be to move functionality down in the hierarchy to an un-binned data structure, basically an `UnbinnedMap` or, very similar, a `SparseMap`, but my current impression is that this is out of reach for this project as we are limited by the time you can provide for implementation. However I wanted to mention it for completeness.

registerrier:
Hello @tunbehaun273 @giacomodamico24 , the PIG is open for a while. It lead to important design discussions and prototypes.
Because the effort is stalled, we propose to merge it as is, and set its status to Withdrawn. (which does not mean we give up on unbinned analysis).

docs/development/pigs/pig-032.rst
```diff
@@ -0,0 +1,190 @@
...
+Things that would need implementing:
+
+* Time-dependent evaluator
+* 1-D evaluator
+* Unit tests
+* Benchmarks
+
+Decision
+========
+The authors decided to withdraw the PIG.
```
registerrier:
```suggestion
Given the lack of activity, it has been decided to withdraw the PIG.
```

docs/development/pigs/pig-032.rst
```diff
@@ -0,0 +1,190 @@
...
+It is clear we need to support an unbinned analysis but the exact implementation is under review.
```
registerrier:
```suggestion
The case for unbinned analysis, this draft PIG provides interesting use cases and design discussion. The exact design will require further iterations.
```

docs/development/pigs/pig-032.rst
```diff
@@ -0,0 +1,190 @@
...
+
+* Time-dependent evaluator
+* 1-D evaluator
+* Unit tests
+* Benchmarks
+
+Decision
+========
+Given the lack of activity, it has been decided to withdraw the PIG.
+The case for unbinned analysis, this draft PIG provides interesting use cases and design discussion. The exact design will require further iterations.
```
registerrier:
```suggestion
The case for unbinned analysis is clear and this draft PIG provides interesting use cases and design discussions. The exact design will require further iterations.
```

PR #4256, opened 2022-12-14
Exposure correction for MapDataset.to_region_map_dataset()

luca-giunti:
**Description**

If `MapDataset.to_region_map_dataset(region)` is passed a `region` which is not fully contained inside the dataset `geom`, the exposure has to be corrected (taking into account the fraction of region solid angle contained in the `geom`). Otherwise, the exposure may be over-estimated and therefore the gamma-ray flux under-estimated.

### DIFF
--- a/gammapy/datasets/map.py
+++ b/gammapy/datasets/map.py
@@ -1471,6 +1471,14 @@ class MapDataset(Dataset):
         """
         from .spectrum import SpectrumDataset
 
+        if self.geoms["geom"].is_region and on_region is not None:
+            raise TypeError(
+                " `on_region` "
+                "argument is not supported for datasets with "
+                " `RegionGeom` "
+                "geometry."
+            )
+
         dataset = self.to_region_map_dataset(region=on_region, name=name)
 
         if containment_correction:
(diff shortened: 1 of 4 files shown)


registerrier (commented):
Thanks @luca-giunti . 

I left some comments inline. 

More generally, I think the docstring should explain the mechanism.  I wonder if this should be configurable. Are there cases where one does not want to modify the exposure to account for leaking events?
 
A side question is whether this should be backported. If the exposure correction is the only sensible option, then probably yes, otherwise, it would be a new option.

gammapy/datasets/map.py
```diff
@@ -1548,7 +1556,22 @@ def to_region_map_dataset(self, region, name=None):
             )
 
         if self.exposure:
-            kwargs["exposure"] = self.exposure.to_region_nd_map(region, func=np.mean)
```
QRemy, line 1551: In `to_region_nd_map` we have :
https://github.com/gammapy/gammapy/blob/92a6eb9f440aa92c73269edcffe8bad4d6493662/gammapy/maps/wcs/ndmap.py#L555-L557
these lines should handle correctly the case where the region do not fully overlap with the dataset.

For me the main difference is that here the function is simply `np.mean` while you use `np.sum()` weighted by solid angle. You can probably define  `weights = self.exposure.geom.solid_angle().self.exposure.geom.solid_angle().data.sum(axis=(1,2))`
 and the masking should be handled correctly by `to_region_nd_map`.

gammapy/datasets/map.py
```diff
@@ -1548,7 +1556,22 @@ def to_region_map_dataset(self, region, name=None):
...
+            geom = self.geoms["geom_exposure"]
+            if geom.is_region:
+                weights = None
```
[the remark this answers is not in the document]
luca-giunti: Yes. When you do `SpectrumDatasetOnOff.to_spectrum_dataset()` it intenally does `SpectrumDatasetOnOff.to_map_dataset().to_spectrum_dataset()` which corresponds to this case

gammapy/datasets/map.py
luca-giunti: > You can probably define weights = self.exposure.geom.solid_angle().self.exposure.geom.solid_angle().data.sum(axis=(1,2))
and the masking should be handled correctly by to_region_nd_map.

This is basically what I did right? With a few more lines of code of course, I agree my implementation can be simplified... But the idea should be the same

QRemy: yes,  but if the weight correction is done  after the application of the region mask within `to_region_nd_map` it would probably be simpler.

registerrier:
This won't be corrected in time for v1.0.1. Setting to v1.0.2 milestone.

Astro-Kirsty:
Closed as this is more than just a bug and needs further discussion. 
The new issue is here #5252

PR #4262, opened 2022-12-20
Modify ci.yml to allow devdeps run to fail

registerrier:
**Description**

This pull request modifies the ci-runs matrix to use the `continue-on-error` variable.  It is set to `false` for all runs except `devdeps`.

To remove deprecation warnings in the ci, the version number of actions/setup-python and actions/checkout have been upgraded.

**Dear reviewer**


registerrier:
CI is now working as expected: i.e. devdeps is an allowed fail. Merging now.

PR #4264, opened 2023-01-02
Deprecate load_cta_irfs, replace usage with load_irf_dict_from_file

maxnoe:
**Description**

In a recent exchange, people were trying to use `load_cta_irfs` for IRFs for CTA they produced themselves using pyirf, however, the function hard codes both the expected IRF classes and the HDU names and performs no checks if the actual data file contains those or other entries.

This method is generally named (for CTA! but has actually nothing CTA specific other than the combination of IRFs and HDU names used) but has hard coded specifics. It should be deprecated / removed to avoid such confusion in the future.

**Dear reviewer**

### DIFF
--- a/docs/user-guide/irf/index.rst
+++ b/docs/user-guide/irf/index.rst
@@ -100,7 +100,7 @@ Using gammapy.irf
     :add-heading:
 
 
-.. minigallery:: gammapy.irf.load_cta_irfs
+.. minigallery:: gammapy.irf.load_irf_dict_from_file
     :add-heading:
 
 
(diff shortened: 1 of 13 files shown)


maxnoe:
The CTA 1DC IRF files are actually not GADF compliant in their `HDUCLAS4` names. 

I now added a workaround, so that also `load_irf_dict_from_file` loads these correctly.

adonath (approved):
Thanks a lot @maxnoe!

Mid-term we could also think about deprecating the method completely and always go through `Observation.read()` instead. Otherwise I fully agree with the change in name and functionality. Looks good to me!

registerrier (approved):
Thanks @maxnoe . This looks good. No comment on my side.

registerrier:

PR #4266, opened 2023-01-05
Correct issue with test_safe_mask_maker_dc1

registerrier:
Signed-off-by: Régis Terrier <rterrier@apc.in2p3.fr>

**Description**

This pull request corrects an issue with test_safe_mask_maker_dc1 in makers/tests/test_spectrum.py

The `assert_allclose(dataset.energy_range[0], 1, rtol=1e-3)` statement is incorrect since `dataset.energy_range[0]`is a `RegionMap`. The test was silently passing up to now but might the origin of the recent fail in ubuntu python3.9 environment.

**Dear reviewer**


registerrier:
This change is actually not needed. The `assert_allclose` already works as expected since the `Map` behaves like an array. This does not solve the CI issue.
Closing now.

Issue #4255, opened 2022-12-12
PointSpatialModel: min and max of lon_0


bkhelifi:
**Gammapy version**
1.0 +

**Bug description**
By default, there is no lon_0 range on the PointSpatialModel, ie they are nan. And it seems that iminuit does not like this (ie no convergence)

**Expected behavior**
The constructor sets limits for lat_0 but not lon_0. Maybe one could harmonize this...

**To Reproduce**
The following nb: [pks2155.ipynb](https://github.com/bkhelifi/mynbs/blob/master/random/class.ipynb)

AtreyeeS:
Hi @bkhelifi ! I do not think that this is a bug. In general, we do not have parameter ranges fixed by default for gammapy, except when it is a natural hard limit, ie the longitude for spatial models (not specifically for the `PointSpatialModel`).

Having parameter limits of course improves the fit convergence and the performance, and in general we do need to specify it in most cases. However, I think we discussed in a dev call that we will not have min and max defaults because it is difficult to specify correct ones satisfying all use cases, but rather explain how to do this in our tutorials

registerrier:
I don't think `lon_0` should have parameter bounds by default. It is a periodic quantity (contrary to `lat_0`) and there is no way to properly define a good default. If minuit explores a part that is not covered by the geometry it will get stuck anyway.

registerrier:
A solution for this is now given in the how to  introduced in #4268. Closing.

PR #4268, opened 2023-01-09
Add an HowTo for the fit non-convergence

bkhelifi:
Signed-off-by: Bruno Khélifi <khelifi@in2p3.fr>

**Description**
This PR is associated with the issue #4255. It adds an entry in the HowTo list of the documentation that explains how to solve the issue

### DIFF
--- a/docs/user-guide/howto.rst
+++ b/docs/user-guide/howto.rst
@@ -159,7 +159,7 @@ one for each time bin.
 
 .. accordion-header::
     :id: collapseHowToOne
-    :title: Non-convergence of a fit because of IRFs validity range
+    :title: Improve fit convergence with constrains on the source position
 
 It happens that a 3D fit does not converge with warning messages indicating that the
 scanned positions of the model is outside valid IRF map range. The type of warning message is:
--- a/docs/user-guide/howto.rst
+++ b/docs/user-guide/howto.rst
@@ -162,12 +162,19 @@ one for each time bin.
     :title: Non-convergence of a fit because of IRFs validity range
 
 It happens that a 3D analysis does not converge with warning messages indicating that the
-scanned positions of the model is outside valid IRF map range. This issue might happen when
-the position of a model has no defined range. The minimizer might scan positions outside the
-spatial range in which the IRFs are computed and then it gets lost.
+scanned positions of the model is outside valid IRF map range. The type of warning message is:
+::
+
+    Position <SkyCoord (ICRS): (ra, dec) in deg
+      (329.71693826, -33.18392464)> is outside valid IRF map range, using nearest IRF defined within
+
+This issue might happen when the position of a model has no defined range. The minimizer
+might scan positions outside the spatial range in which the IRFs are computed and then it gets lost.
 
 The simple solution is to add a physically-motivated range on the model's position, which
-solves the issue in most of the times.
+solves the issue in most of the times. The documentation of the
+`models sub-package <https://docs.gammapy.org/1.0/tutorials/api/models.html#modifying-model-parameters>`_
+explains how to add a validity range of a model parameter.
 
 .. accordion-footer::
 
--- a/docs/user-guide/howto.rst
+++ b/docs/user-guide/howto.rst
@@ -162,7 +162,7 @@ one for each time bin.
     :title: Improve fit convergence with constraints on the source position
 
 It happens that a 3D fit does not converge with warning messages indicating that the
-scanned positions of the model is outside valid IRF map range. The type of warning message is:
+scanned positions of the model are outside the valid IRF map range. The type of warning message is:
 ::
 
     Position <SkyCoord (ICRS): (ra, dec) in deg
--- a/docs/user-guide/howto.rst
+++ b/docs/user-guide/howto.rst
@@ -171,8 +171,9 @@ scanned positions of the model is outside valid IRF map range. The type of warni
 This issue might happen when the position of a model has no defined range. The minimizer
 might scan positions outside the spatial range in which the IRFs are computed and then it gets lost.
 
-The simple solution is to add a physically-motivated range on the model's position, which
-solves the issue in most of the times. The documentation of the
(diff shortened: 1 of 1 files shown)


docs/user-guide/howto.rst
```diff
@@ -157,6 +157,20 @@ one for each time bin.
...
+.. accordion-header::
+    :id: collapseHowToOne
+    :title: Non-convergence of a fit because of IRFs validity range
+
+It happens that a 3D analysis does not converge with warning messages indicating that the
+scanned positions of the model is outside valid IRF map range. This issue might happen when
+the position of a model has no defined range. The minimizer might scan positions outside the
+spatial range in which the IRFs are computed and then it gets lost.
+
+The simple solution is to add a physically-motivated range on the model's position, which
```
adonath, line 169: Maybe also restrict the model position to the FoV size?

AtreyeeS (commented):
Thanks @bkhelifi ! Maybe keep the title more general, like "tips for improving fit convergence".
and say that adding physical motivated parameters ranges is usually quite helpful, eg, on the amplitude or the positions. Link to https://docs.gammapy.org/1.0/tutorials/api/models.html#modifying-model-parameters to see how to do the same.  And mention explicitly about the warning on the IRF validity here.

docs/user-guide/howto.rst
bkhelifi: I have hesitated to mention that. Indeed, Quentin have reminded that some extended source could have their center outside the FoV and their extension is such that a part lies into the FoV. In this case, this could work, no?
This is why I wrote this generic formulation...

adonath: Makes sense, however one could still include some reasonable margin...

bkhelifi:
Hi Atreyee,
I had a doubt on the title... I understand your remark. 
Your title proposal seems to me a bit too generic... The HowTo content is not about an improvement of fit convergence, but only about one type of failure. 
Good suggestion for the link;)

docs/user-guide/howto.rst
```diff
@@ -157,6 +157,28 @@ one for each time bin.
 
 .. accordion-footer::
 
+.. accordion-header::
+    :id: collapseHowToOne
+    :title: Non-convergence of a fit because of IRFs validity range
+
+It happens that a 3D analysis does not converge with warning messages indicating that the
```
registerrier, line 164:
```suggestion
It happens that a 3D fit does not converge with warning messages indicating that the
```

docs/user-guide/howto.rst
```diff
@@ -157,6 +157,28 @@ one for each time bin.
...
+scanned positions of the model is outside valid IRF map range. The type of warning message is:
+::
+
+    Position <SkyCoord (ICRS): (ra, dec) in deg
+      (329.71693826, -33.18392464)> is outside valid IRF map range, using nearest IRF defined within
+
+This issue might happen when the position of a model has no defined range. The minimizer
+might scan positions outside the spatial range in which the IRFs are computed and then it gets lost.
```
registerrier, line 172: I think that the minimizer gets lost because the contribution to the npred of a model located outside is null. The IRF range is a side effect not the cause. e.g. if a region where you have zero exposure, but still a PSF and edisp defined this would fail the same way but without the warning message

docs/user-guide/howto.rst
```diff
@@ -157,6 +157,28 @@ one for each time bin.
 
 .. accordion-footer::
 
+.. accordion-header::
+    :id: collapseHowToOne
+    :title: Non-convergence of a fit because of IRFs validity range
```
registerrier, line 162: I am not sure the cause is the IRF validity range. It could happen with a dataset having PSF and edisp defined over the whole sky but a more limited exposure.

registerrier, line 162: We should try to stick to the (How to) Do something  naming scheme for Howtos

registerrier (commented):
Thanks @bkhelifi . See a few inline comments.

docs/user-guide/howto.rst
bkhelifi: Really? If everything is defined but with a null exposure, npred will be zero and be compared to the data... Is it a failure case?

docs/user-guide/howto.rst
bkhelifi: The warning message is about "outside valid IRF map range". So the HowTo is about this specific case, only.
If the cause is not the the map range, then the warning message is bad! And one should change it.
Do you have any suggestion then?

AtreyeeS: I think that the warning message is correct. It says that the model position is outside the IRF validity range, so the closest valid kernel is used. As @registerrier mentioned, the `npred` is zero because of no exposure

docs/user-guide/howto.rst
AtreyeeS: I think the point is that the `npred` remains zero at all nearby positions, so the fit does not know which direction to follow.

docs/user-guide/howto.rst
```diff
@@ -157,6 +157,28 @@ one for each time bin.
 
 .. accordion-footer::
 
+.. accordion-header::
+    :id: collapseHowToOne
+    :title: Improve fit convergence with constrains on the source position
```
registerrier, line 162:
```suggestion
    :title: Improve fit convergence with constraints on the source position
```

docs/user-guide/howto.rst
```diff
@@ -157,6 +157,28 @@ one for each time bin.
...
+
+It happens that a 3D fit does not converge with warning messages indicating that the
+scanned positions of the model is outside valid IRF map range. The type of warning message is:
```
registerrier, line 165:
```suggestion
scanned positions of the model are outside the valid IRF map range. The type of warning message is:
```

registerrier (approved):
Thanks @bkhelifi .

Two small inline comments (typos).

registerrier:
Fail is unrelated. Merging now.

PR #4269, opened 2023-01-10
Use linebreaks in passenv list in tox.ini

adonath:
**Description*

The current dev docs build fails, because in the latest version of `tox` the `passenv` variable does not allow to use whitespace anymore in the same line. This PR introduces line breaks instead, which is better readable anyway...


registerrier (approved):
Thanks @adonath ! Looks good.

This needs to be backported, right?

bkhelifi (approved):
Thanks @adonath

registerrier:
Merging now.

PR #4272, opened 2023-01-12
Add energy support for a template temporal model

AtreyeeS:
As we previously discussed, this enables to 
1. read the cat-sdc files. I have put the read method in utils now, because the file format is a bit annoying and might change in the future (we should encourage a "map" based format)
2. use adapt the evaluation for such model
3. enable plotting such models

It might be necessary to add some documentation in the future, maybe as a separate tutorial showing a full simulation with such models

### DIFF
--- a/gammapy/modeling/models/utils.py
+++ b/gammapy/modeling/models/utils.py
@@ -39,6 +39,7 @@ def _read_cta_sdc(filename):
                 region=PointSkyRegion(center=position),
                 axes=[energy_axis, time_axis],
                 data=np.array(list(data.data) * u.Unit(data.header["UNITS"])),
+                meta=time_header,
             ),
             time_ref,
         )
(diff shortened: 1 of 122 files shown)


gammapy/modeling/models/utils.py
```diff
@@ -0,0 +1,37 @@
...
+        energy_axis = MapAxis.from_nodes(
+            nodes=energy_hdu.data,
+            unit=energy_hdu.header["TUNIT1"],
+            name="energy",
+            interp="log",
+        )
+        time_hdu = hdul["TIMES"]
+        time_header = time_hdu.header
+        time_header.setdefault("MJDREFI", 0.5)
+        time_header.setdefault("MJDREFF", 55555)
```
fabiopintore, line 23: Should this be the opposite? I think that MJDREFI is integer part while MJDREFF is the float one.

fabiopintore (commented):
Thanks @AtreyeeS . This is very useful and I don't have any major comment.

gammapy/modeling/models/utils.py
AtreyeeS: Indeed, my bad! Thanks @fabiopintore

registerrier:
Closing before a new PR is opened.

Issue #4207, opened 2022-11-14
MapDataset.create() sets mask to False by default


AtreyeeS:
**Gammapy version**
`gammapy-1.0`

**Bug description**
By default, a fully empty `mask_safe` is created when making a dataset, as per
https://github.com/gammapy/gammapy/blob/647c0001a08fbda055a7aed4bf62747e08ea38e7/gammapy/datasets/map.py#L575

This is confusing, as, it leads to non-intuitive results for the background while stacking synthetic datasets, eg:

```python3
from gammapy.datasets import MapDataset
from gammapy.maps import *
import numpy as np

axis = MapAxis.from_energy_bounds("0.1 TeV", "10 TeV", nbin=3)
geom = WcsGeom.create(binsz=0.5, width=(2, 2), axes=[axis])
bkg1 = Map.from_geom(geom)
bkg1.data = np.ones(bkg1.data.shape)
cnt1 = bkg1.copy()
dataset1 = MapDataset(counts=cnt1, background=bkg1, name="dataset-1")

stacked = MapDataset.from_geoms(**dataset1.geoms)
for i in range(3):
    stacked.stack(dataset1)
    print(stacked.counts.data.sum(), stacked.background.data.sum())
```

Output being:

```
48.0 48.0
96.0 48.0
144.0 48.0
```

Since `MapDatasetMaker.run()` creates a `True` `mask_safe` by default, there should be no erroneous results for users going through the standard data reduction loop. Still, I think it would be good to be consistent and set the mask to True in all cases by default.

adonath:
Thanks @AtreyeeS! 

> Since MapDatasetMaker.run() creates a True mask_safe by default, there should be no erroneous results for users going through the standard data reduction loop. Still, I think it would be good to be consistent and set the mask to True in all cases by default.

This is not a bug, but the convention we introduced for `MapDataset.create()` and `.from_geoms()`. The `.mask_safe` will be set to all false by default. This is needed because while stacking, masks are combined using a logical or. If we started with a mask set all true, stacking could never modify the mask: it would just remain all true. Which is even worse default behavior I think.


However looking at `MapDataset.stack()` in noticed indeed one inconsistency. For some reason the background stacking applies the `self.mask_safe` while all the others quantities do not. So I think the actual bug here is the multiplication with `self.mask_safe` here: https://github.com/gammapy/gammapy/blob/main/gammapy/datasets/map.py#L786.

adonath:
[...]

PR #4275, opened 2023-01-17
Remove safe mask in background stacking

AtreyeeS:

As discussed in #4207

### DIFF
--- a/gammapy/datasets/map.py
+++ b/gammapy/datasets/map.py
@@ -783,7 +783,7 @@ class MapDataset(Dataset):
 
         if self.stat_type == "cash":
             if self.background and other.background:
-                background = self.npred_background() * self.mask_safe
+                background = self.npred_background()
                 background.stack(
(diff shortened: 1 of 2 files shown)


bkhelifi:
Hi @AtreyeeS , does your test include or deal with the issue mentioned in #4207 during the Fermi analysis? Do we have also tests for the 2 cases of staking into an empty dataset and a non-empty one?

AtreyeeS:
Hi @bkhelifi 

- This PR and associated test should address the Fermi LAT issue mentioned in quantity
- stacking behaves correctly only when stacking into an empty dataset. There is probably no good way to prevent users from doing the wrong thing, except maybe documentation and constant re-iteration in hands on sessions

registerrier (approved):
Thanks @AtreyeeS !
No comment from my side.

adonath (approved):
Thanks @AtreyeeS, no further comments from me.

PR #4276, opened 2023-01-20
Allow to set psf max radius in evaluator

QRemy:
Currently the psf max radius in the evaluator is derived as the max 0.999 containment radius but this can lead to aberrantly  high values if the psf is not well-defined everywhere or noisy (I noticed that for HAWC psf where it can go from <1deg in one bin in energy and >10deg in the next). This will affect seriously the performance increasing the psf convolution time.

This PR allow to set a PSF_MAX_RADIUS variable in evaluator in order to limit the size of the kernel. If the user set the mask_safe  such as the psf is well-defined they can evaluate the max radius only in the mask_safe.

### DIFF
--- a/gammapy/datasets/evaluator.py
+++ b/gammapy/datasets/evaluator.py
@@ -9,6 +9,7 @@ import matplotlib.pyplot as plt
 from gammapy.maps import HpxNDMap, Map, RegionNDMap, WcsNDMap
 from gammapy.modeling.models import PointSpatialModel, TemplateNPredModel
 
+PSF_MAX_RADIUS = None
 PSF_CONTAINMENT = 0.999
 CUTOUT_MARGIN = 0.1 * u.deg
 
@@ -192,6 +193,7 @@ class MapEvaluator:
                     position=self.model.position,
                     geom=geom_psf,
                     containment=PSF_CONTAINMENT,
+                    max_radius=PSF_MAX_RADIUS,
                 )
 
         if self.evaluation_mode == "local":


adonath (commented):
Thanks @QRemy, I agree this makes sense! Can you verify the code in `.get_psf_kernel()` is tested for this case?

QRemy:
> Can you verify the code in `.get_psf_kernel()` is tested for this case?

https://github.com/gammapy/gammapy/blob/21307cca8dd3337157036316801e1e877678a298/gammapy/irf/psf/tests/test_map.py#L138-L142

adonath (approved):
Thanks, @QRemy! No further comments from my side...

adonath:
Can somebody please to a 2nd review here? @AtreyeeS or @registerrier?

registerrier (approved):
Thanks @QRemy .
 At some point, we might want to have a more easily configurable approach for the `MapEvaluator` on the dataset but I agree this solution is fine for now.

registerrier:
One question though: should this really be backported to 1.0.1?  I suppose this is to allow some HAWC usage with v1.0.x variants right @QRemy ?

adonath:
> One question though: should this really be backported to 1.0.1? I suppose this is to allow some HAWC usage with v1.0.x variants right @QRemy ?

Given that the PSF containment radius seems unstable under certain conditions this could be consider a bug fix. However then we should label it like this...

QRemy:
> Given that the PSF containment radius seems unstable under certain conditions this could be consider a bug fix. However then we should label it like this...

yes we can see it like that

PR #4278, opened 2023-01-20
Fix bkgs  spectral model io

QRemy:
Fix bkgs  spectral model io


registerrier (approved):

One possible issue: wouldn't the new test pass without the change? Shouldn't it also check the dictionary structure?

QRemy:
> One possible issue: wouldn't the new test pass without the change? Shouldn't it also check the dictionary structure?

The test fails without the change (which affects only from_dict and not the dict structure)

PR #4280, opened 2023-01-20
Allow to load observations with only IRFs defined

QRemy:
Add a "has_only_irf" option in DataStore.get_obervation in order to be able to load observations without events defined (useful with HAWC data)

### DIFF
--- a/gammapy/data/data_store.py
+++ b/gammapy/data/data_store.py
@@ -329,9 +329,10 @@ class DataStore:
 
         # TODO: right now, gammapy doesn't support using the pointing table of GADF
         # so we always pass the events location here to be read into a FixedPointingInfo
-        pointing_location = copy(kwargs["events"])
-        pointing_location.hdu_class = "pointing"
-        kwargs["pointing"] = pointing_location
+        if "events" in kwargs:
+            pointing_location = copy(kwargs["events"])
+            pointing_location.hdu_class = "pointing"
+            kwargs["pointing"] = pointing_location
 
         return Observation(**kwargs)
 
--- a/gammapy/data/data_store.py
+++ b/gammapy/data/data_store.py
@@ -259,7 +259,7 @@ class DataStore:
         else:
             return s
 
-    def obs(self, obs_id, required_irf="full-enclosure", has_only_irf=False):
+    def obs(self, obs_id, required_irf="full-enclosure", require_events=True):
         """Access a given `~gammapy.data.Observation`.
 
         Parameters
@@ -281,6 +281,8 @@ class DataStore:
 
             * `"full-enclosure"` : includes `["events", "gti", "aeff", "edisp", "psf", "bkg"]`
             * `"point-like"` : includes `["events", "gti", "aeff", "edisp"]`
+        require_events : bool
+            Require events and gti table or not.
 
         Returns
         -------
@@ -303,10 +305,10 @@ class DataStore:
                 f"{difference} is not a valid hdu key. Choose from: {ALL_IRFS}"
             )
 
-        if has_only_irf:
-            required_hdus = required_irf
-        else:
+        if require_events:
             required_hdus = {"events", "gti"}.union(required_irf)
+        else:
+            required_hdus = required_irf
 
         missing_hdus = []
         for hdu in ALL_HDUS:
@@ -338,7 +340,7 @@ class DataStore:
         obs_id=None,
         skip_missing=False,
         required_irf="full-enclosure",
-        has_only_irf=False,
+        require_events=True,
     ):
         """Generate a `~gammapy.data.Observations`.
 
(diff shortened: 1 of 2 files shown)


adonath:
Thanks a lot @QRemy, I agree this is useful. Also in the context of performing simulations...

Currently I do not really like the various arguments to `.obs()` we end up with. Would it make sense to deprecate `required_irfs` and introduce `required_hdus` instead? The we could introduce an alias for the case without events...

gammapy/data/data_store.py
```diff
@@ -259,14 +260,15 @@ def info(self, show=True):
         else:
             return s
 
-    def obs(self, obs_id, required_irf="full-enclosure"):
+    @deprecated_renamed_argument("required_irf", "required_hdus", "v1.1")
+    def obs(self, obs_id, required_hdus="full-enclosure", required_irf=None):
```
QRemy, line 264: Is it the correct deprecation warning ? because not only the name change but the also expected entries, and I added
``` # deprecated
        if required_irf is not None:
            required_hdus = {"events", "gti"}.union(required_irf) 
```

gammapy/analysis/config.py
```diff
@@ -218,7 +218,14 @@ class ObservationsConfig(GammapyBaseConfig):
     obs_file: Path = None
     obs_cone: SpatialCircleConfig = SpatialCircleConfig()
     obs_time: TimeRangeConfig = TimeRangeConfig()
-    required_irf: List[RequiredHDUEnum] = ["aeff", "edisp", "psf", "bkg"]
+    required_hdus: List[RequiredHDUEnum] = [
```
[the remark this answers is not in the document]
registerrier: A priori this is a different attribute in `ObservationConfig` so I would say that https://docs.gammapy.org/dev/development/dev_howto.html#removing-an-attribute
But I don't know whether this will work in this specific case.

registerrier (commented):
There is an issue with a change that was introduced in the meantime I suppose. See inline comment.

The only thing we need to be sure about is the terminology. If `required_hdu` seems fine, we will have to stick to it. 
I have no clear idea how to handle the deprecation in the ConfigFile. Maybe keep the option possible, which should raise the deprecation warning later, no?

gammapy/data/data_store.py
```diff
@@ -324,14 +329,20 @@ def obs(self, obs_id, required_irf="full-enclosure"):
 
         # TODO: right now, gammapy doesn't support using the pointing table of GADF
         # so we always pass the events location here to be read into a FixedPointingInfo
-        pointing_location = copy(kwargs["events"])
-        pointing_location.hdu_class = "pointing"
-        kwargs["pointing"] = pointing_location
+        if "pointing" not in kwargs and "events" in kwargs:
```
QRemy, line 332:
```suggestion
        if "events" in kwargs:
```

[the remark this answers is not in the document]
QRemy: changed to have only `if "events" in kwargs`

maxnoe:
I think a similar change should then be made to `Observation.read`, right? And the required irf / required hdu mechanism?

This should then also probably remove the need for `load_irf_dict_from_file`, as you should just use `Observation.read` and use the IRFs?

gammapy/data/data_store.py
```diff
@@ -17,9 +18,9 @@
 
 ALL_IRFS = ["aeff", "edisp", "psf", "bkg", "rad_max"]
 ALL_HDUS = ["events", "gti", "pointing"] + ALL_IRFS
-REQUIRED_IRFS = {
-    "full-enclosure": {"aeff", "edisp", "psf", "bkg"},
-    "point-like": {"aeff", "edisp"},
+REQUIRED_HDUS = {
+    "full-enclosure": {"events", "gti", "aeff", "edisp", "psf", "bkg"},
+    "point-like": {"events", "gti", "aeff", "edisp"},
```
bkhelifi, line 23: For point-like, rad_max is not required? W/o it, how to use correctly aeff?

bkhelifi:
I have not understood why it was necessary to replace the `required_irf` by the `required_hdu` in order to achieve the wished feature. It seems to be unrelated...

In term of user interface, using the wording `hdu` is a bit hard, technical, especially for non-fits experts. The working `irf` is more common, more intuitive... I do not understand the motivation of this change of the user interface.

adonath:
> In term of user interface, using the wording hdu is a bit hard, technical, especially for non-fits experts. The working irf is more common, more intuitive... I do not understand the motivation of this change of the user interface.

The change is because events are not an IRF, so it would make sense to include it in the `required_irf` keyword. However one could maybe keep the `required_irf` keyword as is and just make an additional argument to allow for the events being optional...

adonath:
From the dev call today: introduce `require_events` and keep the `required_irfs` as is.

adonath (approved):

adonath:
Please submit a 2nd review @bkhelifi, @AtreyeeS or @registerrier.

AtreyeeS (approved):
Looks good to me! Thanks @QRemy

PR #4281, opened 2023-01-20
Optional meta_table creation in MapDatasetMaker

QRemy:
This PR add meta_table  in to the list of available_selection for the MapDatasetMaker so it can run even if the observations meta_table are not well-defined.

### DIFF
--- a/gammapy/analysis/config.py
+++ b/gammapy/analysis/config.py
@@ -95,6 +95,7 @@ class MapSelectionEnum(str, Enum):
     background = "background"
     psf = "psf"
     edisp = "edisp"
+    meta_table = "meta_table"
 
 
 class GammapyBaseConfig(BaseModel):
--- a/gammapy/makers/map.py
+++ b/gammapy/makers/map.py
@@ -99,7 +99,14 @@ class MapDatasetMaker(Maker):
     """
 
     tag = "MapDatasetMaker"
-    available_selection = ["counts", "exposure", "background", "psf", "edisp"]
+    available_selection = [
+        "counts",
+        "exposure",
+        "background",
+        "psf",
+        "edisp",
+        "meta_table",
+    ]
 
     def __init__(
         self,
@@ -378,7 +385,8 @@ class MapDatasetMaker(Maker):
             Map dataset.
         """
         kwargs = {"gti": observation.gti}
-        kwargs["meta_table"] = self.make_meta_table(observation)
+        if "meta_table" in self.selection:
+            kwargs["meta_table"] = self.make_meta_table(observation)
 
         mask_safe = Map.from_geom(dataset.counts.geom, dtype=bool)
         mask_safe.data[...] = True
--- a/gammapy/makers/map.py
+++ b/gammapy/makers/map.py
@@ -28,7 +28,7 @@ class MapDatasetMaker(Maker):
     ----------
     selection : list
         List of str, selecting which maps to make.
-        Available: 'counts', 'exposure', 'background', 'psf', 'edisp'
+        Available: 'counts', 'exposure', 'background', 'psf', 'edisp', 'meta_table'
         By default, all maps are made.
     background_oversampling : int
         Background evaluation oversampling factor in energy.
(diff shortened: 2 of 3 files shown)


adonath:
Thanks @QRemy! I don't think making meta information configurable is a good idea.

 However the meta information should be (with some exceptions) mostly be optional. So the logic should rather be: `dataset.meta_table` should be always fill with all the information that is available, but we do not rely on this meta info later anywhere (again with some exceptions). So if the info is not available from the DL3 data it should just not be filled...and the `.meta_table` just contains the minimal info.

What is the purpose of the configuration here?

registerrier:
Initially the proposition was to have an `include_meta_table` boolean argument to `MapDatasetMaker.run()`.  I have suggested along the line here with the idea that it could be made more configurable. I suppose this is not the right place to discuss this issue. 
Maybe indeed, the better approach is to let the maker continue even if it can't find the relevant info. 
We nevertheless still miss a mechanism to extract relevant information (not only from the DL3, but also from the various makers which might export important data for debugging or dataset selection/clustering).

QRemy:
> What is the purpose of the configuration here?
I wanted to prevent the maker to fail if the meta informations are not available on the observation.

bkhelifi:
Hi @QRemy,
For me, it is surprising that Gammapy would treat all meta data as optional. Some are really mandatory (like the list of retained obs, or later the provenance) and the user should not be able to not propagate them. So, I do not understand this PR and its motivation.

The work on the meta data is not yet done, from their definition for each product (that needs somehow to be propagated) to the used SkyModel or exclusion masks, and down to the maker configuration options describing the analysis. A dedicated project should be started on it... And so, it is out-of-scope of this work.

However, one could think all together about the list of mandatory and optional meta data associated to the Gammapy data reduction, at first. A feature-request issue could be the place for such collective work. I give here just some examples of questions I have:
- For the datasermaker, the meta data "TELESCOP" is stored for each observation. But during the stacking, this meta data will be repeated N times. Is it wise? This meta data should not be a 'global' meta data of this stacked dataset? 
- Why the meta data "TELESCOP" is extracted from aeff? and not from event? Is the rad_max propagated correctly from aeff? In which case this meta data can be optional?
- Why the meta data "INSTRUME" or "CREATOR" stored into events are not propagated?
- Should we store the maker options, e.g. the 'available_selection' of the datasetmaker? Are these king of meta data optional? Or should we stack them optionally with the provenance information coming from the DL3 prod?
- Who writes the gammapy version into the meta data of the dataset? Which is mandatory because one can serialize them...

adonath:
>> What is the purpose of the configuration here?

> I wanted to prevent the maker to fail if the meta informations are not available on the observation.

I see, but I still think the logic here should be to keep the meta data optional as much as possible. So currently it seems we require (see https://github.com/gammapy/gammapy/blob/main/gammapy/makers/map.py#L334):

- Observation ID information
- Fixed pointing information

All the other info is currently optional and that's how we should proceed. In which scenario does the maker currently fail, because of the **required** missing meta info?

QRemy:
The problem I have is related to #4280, if we want to create an Observation without events just to bundle irfs then the meta table infos are missing. For HAWC analysis it is convenient to create one Observation per event type and use the dataset maker only to interp the irfs to the target geom. Then events are filled fill from the events lists reading only the relevant events type and energy bins. For now we don't really have a way to do these kind of selections at the DataStore level.

QRemy:
Maybe we could just set the meta table only if "counts" in selection, so it won't appear as optional, and will not fail if we process only irfs.

adonath:
> The problem I have is related to https://github.com/gammapy/gammapy/pull/4280, if we want to create an Observation without events just to bundle irfs then the meta table infos are missing. For HAWC analysis it is convenient to create one Observation per event type and use the dataset maker only to interp the irfs to the target geom. Then events are filled fill from the events lists reading only the relevant events type and energy bins. For now we don't really have a way to do these kind of selections at the DataStore level.

Makes sense, I think an `Observation` object without events is generally useful for simulations as well. However I'm sure there is other meta info that we would like to still add in this case, such as e.g. event types or psf classes, which applies to IRFs as well. So I think coupling the meta to the counts is not a good idea either. What should still work is adding the info by hand to the `Observation` object, like we do here: https://docs.gammapy.org/1.0/tutorials/analysis-3d/event_sampling.html#define-an-observation

adonath:
In general we could still make the pointing an optional info...I guess it would not be needed for HAWC observations...

QRemy:
Let's close this assuming that the solution is rather to add the missing infos by hand for now.

PR #4282, opened 2023-01-20
add from_region()

AtreyeeS:

This PR adds a `DiskSpatialModel.from_region()` to create a model directly from an elliptical region. I am using this often these days, and felt having a convenience function can be useful for others as well...

### DIFF
--- a/gammapy/modeling/models/spatial.py
+++ b/gammapy/modeling/models/spatial.py
@@ -775,6 +775,28 @@ class DiskSpatialModel(SpatialModel):
             **kwargs,
         )
 
+    @classmethod
+    def from_region(cls, region, **kwargs):
+        """Create a `DiskSpatialModel from a ~regions.EllipseSkyRegion`
+
+        Parameters
+        ----------
+        region : `~regions.EllipseSkyRegion`
+            region to create model from
+
+        Returns
+        -------
+        spatial_model : `~gammapy.modeling.models.DiskSpatialModel`
+        """
+        frame = kwargs.pop("frame", region.center.frame)
+        model = cls.from_position(getattr(region.center, frame))
+        r0 = max(region.width, region.height)
+        model.r_0.quantity = r0 / 2.0
+        r1 = min(region.width, region.height)
+        model.e.value = np.sqrt(1.0 - np.power(r1 / r0, 2))
+        model.phi.quantity = 90 * u.deg + region.angle
+        return model
+
 
 class ShellSpatialModel(SpatialModel):
     r"""Shell model.
--- a/gammapy/modeling/models/tests/test_spatial.py
+++ b/gammapy/modeling/models/tests/test_spatial.py
@@ -223,6 +223,21 @@ def test_sky_disk_edge():
     assert_allclose((value_edge_nwidth / value_center).to_value(""), 0.95)
 
 
+def test_disk_from_region():
+    region = EllipseSkyRegion(
+        center=SkyCoord(20, 17, unit="deg"),
+        height=0.3 * u.deg,
+        width=1.0 * u.deg,
+        angle=30 * u.deg,
+    )
+    disk = DiskSpatialModel.from_region(region, frame="galactic")
+    assert_allclose(disk.parameters["lon_0"].value, 2.315473, rtol=1e-2)
+    assert_allclose(disk.parameters["lat_0"].value, -0.791178, rtol=1e-2)
+    assert_allclose(disk.parameters["r_0"].value, 0.5, rtol=1e-2)
+    assert_allclose(disk.parameters["e"].value, 0.9539, rtol=1e-2)
+    assert_allclose(disk.parameters["phi"].quantity, 120 * u.deg)
+
+
 def test_sky_shell():
     width = 2 * u.deg
     rad = 2 * u.deg


gammapy/modeling/models/spatial.py
```diff
@@ -775,6 +775,28 @@ def to_region(self, **kwargs):
...
+
+        Parameters
+        ----------
+        region : `~regions.EllipseSkyRegion`
+            region to create model from
+
+        Returns
+        -------
+        spatial_model : `~gammapy.modeling.models.DiskSpatialModel`
+        """
```
adonath, line 790: Please add a check like `isinstance(region, EllipseSkyRegion)`...

gammapy/modeling/models/spatial.py
```diff
@@ -775,6 +775,28 @@ def to_region(self, **kwargs):
...
+        frame = kwargs.pop("frame", region.center.frame)
+        model = cls.from_position(getattr(region.center, frame))
```
adonath, line 792: Shouldn't `.from_position(position=region.center)` work directly? The frame handling is done in `.from_position()`...

gammapy/modeling/models/spatial.py
```diff
@@ -775,6 +775,28 @@ def to_region(self, **kwargs):
...
+        r0 = max(region.width, region.height)
```
adonath, line 793: I think in regions there is no convention on which is the major or minor axis, so this could potential flip the ellipse by 90 deg?

adonath (changes requested):
Thanks @AtreyeeS, I agree this is a very convenient feature! Maybe it would be good to check for self consistency with `.to_region()`...

gammapy/modeling/models/spatial.py
AtreyeeS: Done! I thought our dev docs dissuade adding instance checks https://docs.gammapy.org/1.0/development/dev_howto.html#what-checks-and-conversions-should-i-do-for-inputs

gammapy/modeling/models/spatial.py
AtreyeeS: `.from_position()` takes the frame of the `SkyCoord` as the frame of the model. The regions can be defined in a different frame than that of the analysis, so added this...

gammapy/modeling/models/spatial.py
AtreyeeS: Thank you @adonath ! Indeed my bad. fixed it now

gammapy/modeling/models/spatial.py
adonath: Oh, but this should be handled in Gammpay correctly (I hope)? It should be possible to define a model in ICRS while the maps are in galactic coordinates...so just copying over the frame from the region should be fine. If not we should make it work...

adonath: Yes, we are not fully consistent here. However in this case the input is very well defined and only one class is possible. So I think the check is simple and helpful in this case...

gammapy/modeling/models/spatial.py
```diff
@@ -775,6 +775,38 @@ def to_region(self, **kwargs):
...
+        Returns
+        -------
+        spatial_model : `~gammapy.modeling.models.DiskSpatialModel`
+        """
+        if isinstance(region, CircleSkyRegion):
+            region = EllipseSkyRegion(
+                center=region.center, width=region.radius, height=region.radius
+            )
+        if not isinstance(region, EllipseSkyRegion):
+            raise TypeError("Please provide a Circular or Elliptic region")
```
adonath, line 796: It is very good practice to also tell users what they actually passed as input, like:

```suggestion
            raise ValueError(f"Please provide a `CircleSkyRegion` or `EllipseSkyRegion`, got {type(region) instead.}")
```

gammapy/modeling/models/spatial.py
```diff
@@ -775,6 +775,38 @@ def to_region(self, **kwargs):
...
+        frame = kwargs.pop("frame", region.center.frame.name)
+        model = cls.from_position(getattr(region.center, frame))
+        if region.height > region.width:
+            r0, r1 = region.height, region.width
+            phi = region.angle
+        else:
+            r1, r0 = region.height, region.width
+            phi = 90 * u.deg + region.angle
+        model.r_0.quantity = r0 / 2.0
+        model.e.value = np.sqrt(1.0 - np.power(r1 / r0, 2))
```
QRemy, line 806:
```suggestion
        if region.height > region.width:
            major_axis, minor_axis = region.height, region.width
            phi = region.angle
        else:
            minor_axis, major_axis = region.height, region.width
            phi = 90 * u.deg + region.angle
        model.r_0.quantity = major_axis / 2.0
        model.e.value = np.sqrt(1.0 - np.power(minor_axis / major_axis, 2))
```

gammapy/modeling/models/spatial.py
```diff
@@ -775,6 +775,38 @@ def to_region(self, **kwargs):
...
+        -------
+        spatial_model : `~gammapy.modeling.models.DiskSpatialModel`
+        """
+        if isinstance(region, CircleSkyRegion):
+            region = EllipseSkyRegion(
+                center=region.center, width=region.radius, height=region.radius
+            )
+        if not isinstance(region, EllipseSkyRegion):
+            raise TypeError("Please provide a Circular or Elliptic region")
+        frame = kwargs.pop("frame", region.center.frame.name)
```
QRemy, line 797: if the given `frame` is different from the region frame the rotation angle seems incorrect.  Here is an example:


```
from regions import EllipseSkyRegion
import astropy.units as u
from gammapy.modeling.models import DiskSpatialModel

region = EllipseSkyRegion(
    center=SkyCoord(0, -10, unit="deg"),
    height=0.3 * u.deg,
    width=1.0 * u.deg,
    angle=0 * u.deg,
)
kwargs={}
model = DiskSpatialModel.from_region(region, **kwargs)
plt.figure()
ax = model.plot(add_cbar=True)

# illustrate size parameter
plot_region = model.to_region().to_pixel(ax.wcs)
artist = plot_region.as_artist(facecolor="none", edgecolor="b", lw=2)
ax.add_artist(artist)

kwargs={"frame":"galactic"}
model = DiskSpatialModel.from_region(region, **kwargs)

# illustrate size parameter
plot_region = model.to_region().to_pixel(ax.wcs)
artist = plot_region.as_artist(facecolor="none", edgecolor="m", lw=2)
ax.add_artist(artist)
```
image

AtreyeeS: I think you need to use a different `ax` here for the 2 cases. Plotting on the same figure is creating the apparent issue,

Otherwise,

```
from regions import EllipseSkyRegion
import astropy.units as u
from gammapy.modeling.models import DiskSpatialModel

region = EllipseSkyRegion(
    center=SkyCoord(0, -10, unit="deg"),
    height=0.3 * u.deg,
    width=1.0 * u.deg,
    angle=-40.0 * u.deg,
)
kwargs={}
model = DiskSpatialModel.from_region(region, **kwargs)
plt.figure()
ax = model.plot(add_cbar=True)
plot_region = model.to_region().to_pixel(ax.wcs)
artist = plot_region.as_artist(facecolor="none", edgecolor="b", lw=2)
ax.add_artist(artist)
```

gives 

image

and
```
# change frame
region = EllipseSkyRegion(
    center=SkyCoord(0, -10, unit="deg"),
    height=0.3 * u.deg,
    width=1.0 * u.deg,
    angle=0 * u.deg,
)

kwargs={"frame":"galactic"}
model = DiskSpatialModel.from_region(region, **kwargs)

ax = model.plot()
# illustrate size parameter
plot_region = model.to_region().to_pixel(ax.wcs)
artist = plot_region.as_artist(facecolor="none", edgecolor="m", lw=2)
ax.add_artist(artist)
```

gives 
image

gammapy/modeling/models/tests/test_spatial.py
```diff
@@ -223,6 +223,46 @@ def test_sky_disk_edge():
...
+    assert_allclose(disk.parameters["phi"].quantity, 30 * u.deg)
+
+    region = CircleSkyRegion(center=region.center, radius=1.0 * u.deg)
+    disk = DiskSpatialModel.from_region(region)
+    assert_allclose(disk.parameters["e"].value, 0.0, rtol=1e-2)
+    assert_allclose(disk.parameters["lon_0"].value, 20, rtol=1e-2)
+    assert disk.frame == "galactic"
+
+    region = PointSkyRegion(center=region.center)
+    with pytest.raises(TypeError):
```
AtreyeeS, line 262:
```suggestion
    with pytest.raises(ValueError):
```

gammapy/modeling/models/spatial.py
QRemy: but if the region are equivalent `model.to_region().to_pixel(ax.wcs)` should do the projection correctly for any `ax.wcs`

AtreyeeS: yes indeed @QRemy ! The frame transfer is incorrect in this PR. I will fix this. Thanks a lot

AtreyeeS:
Thank you @adonath and @QRemy ! I fixed the issues discussed. Can you do another review please

QRemy (dismissed):

gammapy/modeling/models/spatial.py
```diff
@@ -775,6 +775,49 @@ def to_region(self, **kwargs):
...
 
+    @classmethod
+    def from_region(cls, region, **kwargs):
+        """Create a `DiskSpatialModel from a ~regions.EllipseSkyRegion`
+
+        Parameters
+        ----------
+        region : `~regions.EllipseSkyRegion` or ~regions.CircleSkyRegion`
+            region to create model from
+
```
registerrier, line 786: Why is the `frame` parameter hidden?

registerrier (commented):
Thanks @AtreyeeS . 

Why are you using a `kwargs` argument and not explicitly  the `frame` argument? 

If the behaviour is not transparent regarding e.g. what is the rotation angle and what axis it is measured from, it should be clarified in the docstring.
The default behaviour here is that the angle is computed from the longitude axis of the region center frame. It should be explained.
If nothing in the `EllipseSkyRegion` defines the celestial `frame` whose longitude axis is used to define the angle, then `frame` might not have a default a default value at all.

QRemy:
>  If nothing in the `EllipseSkyRegion` defines the celestial `frame` whose longitude axis is used to define the angle, then `frame` might not have a default a default value at all.

I think the `SkyRegion` takes the `frame` from the center given as a `SkyCoord` so it is always defined and ICRS is the default.

AtreyeeS:
> Why are you using a kwargs argument and not explicitly the frame argument?

To forward the kwargs to `from_position`, and because the `SpatialModel` has `frame` inside kwargs in general.

> If the behaviour is not transparent regarding e.g. what is the rotation angle and what axis it is measured from, it should be clarified in the docstring.

The figure is self explanatory, no?

> If nothing in the EllipseSkyRegion defines the celestial frame whose longitude axis is used to define the angle, then frame might not have a default a default value at all.

The default is `icrs` as Quentin mentioned

registerrier (approved):
Thanks @AtreyeeS .

gammapy/modeling/models/spatial.py
```diff
@@ -775,6 +775,49 @@ def to_region(self, **kwargs):
...
+        if region.height > region.width:
+            major_axis, minor_axis = region.height, region.width
+            phi = region.angle
+        else:
+            minor_axis, major_axis = region.height, region.width
+            phi = 90 * u.deg + region.angle
+        model.r_0.quantity = major_axis / 2.0
+        model.e.value = np.sqrt(1.0 - np.power(minor_axis / major_axis, 2))
+        model.phi.quantity = phi
+        return model
```
adonath, line 819: Here are some more advanced ideas: e.g. it is usually a good idea to work with a pattern like:

```python3
kwargs.setdefault("r_0", major_axis / 2.0)
kwargs.setdefault("phi", phi)

return cls.from_position(region.position, **kwargs)
```
This way you could even overwrite quantities of the disk model.

For the frame handling you could maybe introduce a general utility method `region_to_frame()` in `gammapy.utils.regions` (and maybe make a feature request in `regions`). For an arbitrary region you can actually skip the computation of the width. The code here would simplify to:

```python3
frame = kwargs.pop("frame", region.center.frame)
region = region_to_frame(region, frame=frame)
```

AtreyeeS:
Thanks a lot @adonath for the advice! The implementation looks much cleaner now

adonath (approved):
Thanks a lot @AtreyeeS, no further comments from my side!

Issue #4287, opened 2023-01-23
Using TSMapEstimator on a MapDatasetOnOff should raise an error


MRegeard:
**Gammapy version**
1.0 / dev

**Bug description**
Using `TSMapEstimator`on a `MapDatasetOnOff`leads to error : 
```
...
-> dataset = dataset.pad(pad_width, name=dataset.name)
...
TypeError: pad() got an unexpected keyword argument 'name'
```
This is because `pad()`is not implemented in `MapDatasetOnOff`. 
Moreover, one should not use `TSMapEstimator`on a `MapDatasetOnOff` so I think that a proper error should be raise to tell this to the user.

PR #4288, opened 2023-01-24
Add MapDatasetOnOff type test and associated error for TSMapEstimator

MRegeard:

This pull request is associated with issue #4287.

I added a `TypeErrors`if the type of `dataset`in `TSMapEstimator.run`is `MapDatasetOnOff`. 

I also added a test for that.

### DIFF
--- a/gammapy/estimators/map/ts.py
+++ b/gammapy/estimators/map/ts.py
@@ -8,7 +8,6 @@ import numpy as np
 import scipy.optimize
 from astropy.coordinates import Angle
 from astropy.utils import lazyproperty
-from gammapy.datasets import MapDatasetOnOff
 from gammapy.datasets.map import MapEvaluator
 from gammapy.datasets.utils import get_nearest_valid_exposure_position
 from gammapy.maps import Map, Maps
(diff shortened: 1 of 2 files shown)


gammapy/estimators/map/ts.py
```diff
@@ -468,6 +469,8 @@ def run(self, dataset):
                 * flux_ul : upper limit map
 
         """
+        if isinstance(dataset, MapDatasetOnOff):
```
registerrier, line 472: As the issue is the necessity to use the `cash` statistic, you could test `dataset.stat_type != "cash"`

registerrier (approved):
Thanks @MRegeard . This looks good. See the small inline comment.

gammapy/estimators/map/ts.py
MRegeard: Done !
Should I add "with stat type {dataset.stat_type}" at the end of the error print or is it good enough as it is?

adonath (approved):
Thanks @MRegeard! No further comments from my side...

registerrier (approved):
Thanks @MRegeard ! No further comment either.

PR #4292, opened 2023-01-26
Do not run release and docs-dev actions in forks

maxnoe:
**Description**

This pull request ...

**Dear reviewer**


adonath (approved):
Thanks @maxnoe, I have no comments.

registerrier (approved):
Thanks  @maxnoe . Looks good. I assume the change on the release workflow is less necessary, but it is probably better to have the same behavior.

PR #4294, opened 2023-01-26
Addition of a Map.dot operator

registerrier:
**Description**

This pull request proposes a new operator on `Map` e.g. to allow for convolution  such as the `apply_edisp`. 
In all generality, it is not trivial to deal properly with the multiplicity of axes. 
For numpy arrays, the `a.dot(b)` syntax will sum along the last axis of `a` and second to last axis of `b`.  In our case, summation is done along the only common axis. This therefore requires moving axes around. The solution so far is not satisfactory.

**Dear reviewer**


gammapy/maps/core.py
```diff
@@ -1907,3 +1907,76 @@ def sample_coord(self, n_events, random_state=0):
         cdict = OrderedDict(zip(self.geom.axes_names, coords))
 
         return MapCoord.create(cdict, frame=self.geom.frame)
+
+    def dot(self, other):
```
maxnoe, line 1911: This should probably be `__matmul__`.

adonath: Or probably both. There might be cases where we need an additional `.dot(other, common_axis_name="")` to specify which axis to sum over. If it is ambiguous it is currently just implemented to fail, which is fine. But we might change that in future...

adonath: But I'm definitely +1 to support the `@` operator...!

adonath:
Thanks @registerrier!  think this could be turned from a draft to review ready?

adonath:
I think it would be good to introduce `Map.move_axis()` and maybe `Map.insert_axis()` as independent PRs.

adonath (approved):
As far as I can tell the remaining work is to refactor to use the new `reorder_axes()` method? Bur this could also happen later in a follow up PR. So no further comments from my side.

Please provide a 2nd review @QRemy or @AtreyeeS...

QRemy (approved):
Thanks @registerrier, I agree the remaining changes can be done in a separate PR.

Issue #4257, opened 2022-12-17
Incorrect serialisation of light curves time axis


cosimoNigro:
**Gammapy version**
1.0

**Bug description**
After saving a light curve (a `FluxPoints` object) to file and reading it back, the time axis changes scale from `tt` to `utc`.
This is a problem if one is manually plotting light curves with times in MJD - as very commonly done.

**To Reproduce**
Here a snippet to produce few LC bins, based on the tutorials

```python
import astropy.units as u
from gammapy.analysis import Analysis, AnalysisConfig
from gammapy.modeling.models import PowerLawSpectralModel, SkyModel
from gammapy.estimators import LightCurveEstimator, FluxPoints
import matplotlib.pyplot as plt

# use gammapy.analysis to quickly obtain datasets
conf_1d = AnalysisConfig()
conf_1d.observations.obs_ids = [23523, 23526, 23559, 23592]
conf_1d.datasets.type = "1d"
conf_1d.datasets.stack = False
conf_1d.datasets.on_region = dict(
    frame="icrs",
    lon=83.63308 * u.deg,
    lat=22.01450 * u.deg,
    radius=0.1 * u.deg,
)
conf_1d.datasets.containment_correction = True
conf_1d.datasets.geom.axes.energy = dict(min=0.7 * u.TeV, max=10 * u.TeV, nbins=5)
conf_1d.datasets.geom.axes.energy_true = dict(min=0.3 * u.TeV, max=20 * u.TeV, nbins=20)

analysis_1d = Analysis(conf_1d)
analysis_1d.get_observations()
analysis_1d.get_datasets()
datasets = analysis_1d.datasets

# let us compute the LC
spectral_model = PowerLawSpectralModel(
    index=2.702,
    amplitude=4.712e-11 * u.Unit("1 / (cm2 s TeV)"),
    reference=1 * u.TeV,
)
sky_model = SkyModel(spectral_model=spectral_model, name="crab")
datasets.models = [sky_model]

# compute the LC
lc_estimator = LightCurveEstimator(
    energy_edges=[1, 10] * u.TeV, source="crab", reoptimize=False
)
lc_1 = lc_estimator.run(datasets)

# now write it and read it again
lc_1.write("lc.fits", format="lightcurve", overwrite=True)
lc_2 = FluxPoints.read("lc.fits", format="lightcurve")

import IPython; IPython.embed()
```

if we open an `Ipython` session and examine the two LC objects we get:

```
[...]

PR #4301, opened 2023-01-30
Add TIMESYS to lightcurve table meta

registerrier:
Signed-off-by: Régis Terrier <rterrier@apc.in2p3.fr>

**Description**

This pull request corrects issue #4257 . To sort the inconsistent timescale between the serialized `TimeMapAxis` and the initial one, a `TIMESYS` key is added to the `table.meta`.

Serializing directly `Time` objects, rather than MJD values, would of course avoid this kind of issues.

**Dear reviewer**


bkhelifi:
The current recommandation for LCs in GADF is:
https://gamma-astro-data-formats.readthedocs.io/en/v0.3/lightcurves/index.html
`MJDREFI`, `MJDREFF`, `TIMEUNIT` are needed, and `TIMESYS` highly encouraged

bkhelifi:
And also, as mentioned in https://gamma-astro-data-formats.readthedocs.io/en/v0.3/general/hduclass.html,
one needs to add in the header:
- `HDUCLASS` : General identifier of the data format. Recommended value: “GADF” (for “gamma-astro-data-formats”)
- `HDUDOC` : Link to the DL3 specifications documentation
- `HDUVERS` : Version of the DL3 specification format

registerrier:
While  adding `TIMEUNIT`, `MJDREFF` makes sense, it is less clear for the `HDUCLASS`  and `HDUVERS` keywords as there is no specific format defined for lightcurve.
Also as far as I know, there is no requirement for these keywords on the DL4 products described in gadf, e.g. maps or flux points.

bkhelifi:
no explicit format is given, indeed.
But, is is controversial to precise which 'time' format is used?

maxnoe:
You cannot claim GADF for something GADF does not define. As I read the current GADF standard, it does not define a light curve format, just gives some general recommendations. So filling in `HDUCLAS` `GADF` would be wrong. Howabout `HDUCLAS = "Gammapy"` for all formats that are only internally defined by gammapy.

Maybe include a version : `HDUCLAS = "Gammapy 1.0"`

maxnoe:
And yes, the header keywords that define the time columns should always be included when writing times.

bkhelifi:
I have a question: for this format 'lightcurve', is it backward compatible? If not, a version number is not needed?

adonath:
@registerrier I'm +1 to add the `HDUCLAS = "Gammapy" / HDUVERS = "1.0"` for the generic Gammapy formats. The clarifiers their origin and context.

adonath:
@bkhelifi This formats are not really meant to evolve, they are just there because they were "always" there. So those formats should only contain minimal backwards compatible bug-fixes if needed.

registerrier:
Adding a specific `HDUCLASS` `HDUVERS` set of keywords for gammapy specific formats seems indeed a good idea. I suppose it requires a bit of work beforehand to determine what objects require that. In particular, while SEDs/flux points or skymaps are relatively well described in gadf there is no specification for the format keywords. Does this mean that we should consider this requires `HDUCLASS=="Gammapy"` as well?
Also an `HDUCLAS1` (and a possibly deeper hierarchy) )will be needed e.g. to distinguish between the various serialized objects.

This requires a full independent issue and PR. What would be a reasonable milestone: 1.1? I guess this would not be backported to v1.0.x. No?

bkhelifi (dismissed):
Thanks @registerrier

gammapy/estimators/points/core.py
```diff
@@ -387,6 +387,12 @@ def to_table(self, sed_type=None, format="gadf-sed", formatted=False):
                 tables.append(table_flat)
 
             table = vstack(tables)
+
+            table.meta["TIMESYS"] = time_axis.reference_time.scale
```
adonath, line 391: Was there any reason why we could not use https://github.com/gammapy/gammapy/blob/main/gammapy/utils/time.py#L99?

adonath: Or even better we could introduce `TimeMapAxis.to_table()`, with the given format...

registerrier: For now, I have used the `time_ref_to_dict` function.

bkhelifi (approved):
Thanks @registerrier . Nothing to add.

QRemy (approved):
Thanks @registerrier, no comments from my side.

PR #4302, opened 2023-01-31
Adapt condition on release workflow

registerrier:
**Description**

This pull request solves an issue with the release.yml workflow. The two independent successive conditions seem to break the syntax. 
Combining them with a `and` operator should solve the issue. Is there a way to test the workflow?

**Dear reviewer**


adonath (approved):
Thanks @registerrier! For local testing I have sometimes used https://github.com/nektos/act, which kind of worked. It creates a docker and runs the action in this environment locally...

PR #4303, opened 2023-01-31
Add  __str__ on CountStatisticsObjects

AtreyeeS:

I find it helpful to have a quick look at these objects, hence the PR

### DIFF
--- a/gammapy/stats/counts_statistic.py
+++ b/gammapy/stats/counts_statistic.py
@@ -61,6 +61,15 @@ class CountsStatistic(abc.ABC):
         """
         return 0.5 * chi2.sf(self.ts, 1)
 
+    def __str__(self):
+
+        str_ = "\t{:32}: {:.2f} \n".format("On counts", self.n_on)
+        str_ += "\t{:32}: {:.2f} \n".format("Background counts", self.n_bkg)
+        str_ += "\t{:32}: {:.2f} \n".format("Excess counts", self.n_sig)
+        str_ += "\t{:32}: {:.2f} \n".format("Significance", self.sqrt_ts)
+
+        return str_.expandtabs(tabsize=2)
+
     def compute_errn(self, n_sigma=1.0):
         """Compute downward excess uncertainties.
 
@@ -250,6 +259,12 @@ class CashCountsStatistic(CountsStatistic):
         """Stat value for best fit hypothesis, i.e. expected signal mu = n_on - mu_bkg"""
         return cash(self.n_on, self.n_on)
 
+    def __str__(self):
+        str_ = f"{self.__class__.__name__}\n"
+        str_ += "\t{:32}: {:.2f} \n".format("Predicted background counts", self.mu_bkg)
+        str_ += super().__str__()
+        return str_.expandtabs(tabsize=2)
+
     def _stat_fcn(self, mu, delta=0, index=None):
         return cash(self.n_on[index], self.mu_bkg[index] + mu) - delta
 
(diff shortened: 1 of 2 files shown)


adonath:
Thanks @AtreyeeS, I agree this is a very helpful method! I'd like to suggest one change to the implementation:

- A pattern that has worked well for the `Dataset` was split the string generation and content generation.
- So introduce `.info_dict()` (see https://github.com/gammapy/gammapy/blob/main/gammapy/datasets/map.py#L1337)m which returns a dict with the stats info.
- Use that dict to format a string in `__str__` (see https://github.com/gammapy/gammapy/blob/main/gammapy/datasets/map.py#L246)

The advantage here is two-fold:
- The dict could e.g. be serialized to json or yaml. Which can be useful for certain diagnostic data...
- We could reuse `CountsStastitic.info_dict()` in `MapDataset.info_dict()` and just add the info using the `dict.update()` method. 
- Finally we could then make the names of the quantities consistent see also: https://docs.gammapy.org/1.0/user-guide/stats/index.html#notations

registerrier (commented):
Thanks @AtreyeeS . This is a good addition.

One general comment, the various members are usually arrays. 
- It could be useful to have informations on their dimensions. 
- Printing the full arrays might produce a very cluttered output. Maybe print the sum? Yet, it is difficult to know whether the input can be summed in general.

AtreyeeS:
Indeed, for the acceptance, mean or sum does not make sense...
If someone wants to visualise text for large arrays, that is kind on an illegible use case, no?

Maybe an option can be to have `.info_dict(detailed=True)` which prints the full arrays, and only the summed (and min and max acceptance) if `False`?

gammapy/stats/tests/test_counts_statistic.py
```diff
@@ -249,3 +249,36 @@ def test_wstat_sum():
     assert stat_sum.n_on.shape == (3,)
     assert_allclose(stat_sum.n_on, (20, 40, 60))
     assert_allclose(stat_sum.n_bkg, (10, 14, 26))
+
+
+def test_CountStatistic_str():
```
adonath, line 254:
```suggestion
def test_counts_statistic_str():
```

gammapy/stats/tests/test_counts_statistic.py
```diff
@@ -249,3 +249,36 @@ def test_wstat_sum():
...
+    assert "Total significance" in str(cash)
+
+    wstat = WStatCountsStatistic(n_on=5, n_off=4, alpha=0.2, mu_sig=2)
+    assert "Off counts" in str(wstat)
+    assert "alpha " in str(wstat)
+    assert "Total counts " in str(wstat)
+    assert "WStatCountsStatistic" in str(wstat)
+
+
+def test_CountStatistic_infodict():
```
adonath, line 267:
```suggestion
def test_counts_statistic_info_dict():
```

gammapy/stats/counts_statistic.py
```diff
@@ -61,6 +61,48 @@ def p_value(self):
...
+        str_ += "\t{:32}: {{significance:.2f}}\n".format("Total significance")
+        str_ += "\t{:32}: {{p_value:.3f}}\n".format("p - value")
+        str_ += "\t{:32}: {{n_bins:.0f}}\n".format("Total number of bins")
+        info = self.info_dict()
+        info["n_bins"] = np.array(self.n_on).size
+        str_ = str_.format(**info)
+
+        return str_.expandtabs(tabsize=2)
+
+    def info_dict(self, summed=True, axis=None):
```
adonath, line 77: Actually I think we could just remove the `summed` option. User can achieve the same, like:
```python3
stat.sum().info_dict()
```
Instead of:
```python3
stat.info_dict(summed=True)
```
I think there is no clear advantage for the latter case. Not that we have the pattern of "hierarchical" function calls also elsewhere, e.g. `map.smooth(1).plot()` or `map.sum_over_axes().plot()`.

adonath (commented):
Thanks @AtreyeeS, this looks nice and clean now. I have left one last comment on the need for the `summed` argument. I think we could actually remove it...

adonath (approved):
Thanks @AtreyeeS, looks good to me!

adonath:
Can someone do a 2nd review here? Maybe @registerrier or @QRemy

QRemy (approved):

PR #4304, opened 2023-02-02
Fix zero length shape after  WcsGeom.cutout()

QRemy:
This fix a bug where WcsGeom.cutout() creates a geom with zero length shape.
In the test case added the problem appears when the width passed to `astropy.nddata.Cutout2D` is in degrees but not if it is in pixels.

### DIFF
--- a/gammapy/maps/wcs/tests/test_geom.py
+++ b/gammapy/maps/wcs/tests/test_geom.py
@@ -622,8 +622,9 @@ def test_wcs_geom_to_even_npix():
     assert geom_even.data_shape == (4, 4)
 
 
-def test_wcs_geom_no_zero_shape_cut(skydir):
+def test_wcs_geom_no_zero_shape_cut():
     geom = WcsGeom.create(skydir=(0, 2.5), binsz=(360, 5), width=(360, 180))
+    skydir = SkyCoord(110.0, 75.0, unit="deg", frame="icrs")
 
     geom_cut = geom.cutout(skydir, width=5)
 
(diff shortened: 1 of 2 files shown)


gammapy/maps/wcs/tests/test_geom.py
```diff
@@ -620,3 +620,13 @@ def test_wcs_geom_to_even_npix():
...
     assert geom_even.data_shape == (4, 4)
+
+
+def test_wcs_geom_no_zero_shape_cut():
+    geom = WcsGeom.create(skydir=(0, 2.5), binsz=(360, 5), width=(360, 180))
+    skydir = SkyCoord(110.0, 75.0, unit="deg", frame="icrs")
+
+    geom_cut = geom.cutout(skydir, width=5)
+
+    assert geom_cut.data_shape != (1, 0)
```
registerrier, line 631: why the two tests?

registerrier (dismissed):
Thanks @QRemy. Looks good to me. I have just left a small inline comment.

Please rebase to avoid the black CI fails. Thanks!

gammapy/maps/wcs/tests/test_geom.py
QRemy: I added this one such as it is more explicit that what we want to avoid in this test is the zero shape.

registerrier (approved):
Thanks @QRemy!

adonath (approved):
Thanks @QRemy, no comments from my side...

PR #4306, opened 2023-02-03
Add a parameter with is_norm=True for PiecewiseNormSpectralModel

QRemy:
Set is_norm=True for parameters in PiecewiseNormSpectralModel

### DIFF
--- a/gammapy/modeling/models/spectral.py
+++ b/gammapy/modeling/models/spectral.py
@@ -1098,6 +1098,9 @@ class PiecewiseNormSpectralModel(SpectralModel):
     def __init__(self, energy, norms=None, interp="log"):
         self._energy = energy
         self._interp = interp
+        self._norm = Parameter(
+            "_norm", 1, unit="", interp="log", is_norm=True, frozen=True
+        )
 
         if norms is None:
             norms = np.ones(len(energy))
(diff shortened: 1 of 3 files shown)


adonath:
Hm, I think this somehow in conflict with the purpose of the `.is_norm` parameter: identify the single parameter for the overall norm of a spectral model. We do not really have the notion of an energy dependent norm at this time. I think the correct way to handle this would be the introduction of a single "global" norm parameter, like:

```
f(E) = norm * (1 * A_1(E) + norm_2 * A_2(E) + norm_3 * A_3(E))
```
Instead of:
```
f(E) = norm _1 * A_1(E) + norm_2 * A_2(E) + norm_3 * A_3(E)
```

I hope this make sense...

QRemy:
> ```
> f(E) = norm * (1 * A_1(E) + norm_2 * A_2(E) + norm_3 * A_3(E))
> ```
Parameters will probably be too correlated if norm is free maybe I could keep norm_1 and introduce a norm_0 with `is_norm = True` frozen by default  so it is used only for the flux points.

> ```
> f(E) = norm_0 * (norm_1 * A_1(E) + norm_2 * A_2(E) + norm_3 * A_3(E))
> ```

On a longer term we should think about modifying the flux points such as they scale the total flux and not the spectral_model then `is_norm` won't be needed anymore.

adonath:
> Parameters will probably be too correlated if norm is free maybe I could keep norm_1 and introduce a norm_0 with is_norm = True frozen by default so it is used only for the flux points.

This sounds like a pragmatic solution now. I think `nom_0` we already have, but you could name it e.g. `norm` without index...

> On a longer term we should think about modifying the flux points estimator such as it scales the total flux and not the spectral_model then is_norm won't be needed anymore.

Yes, this will most likely come with moving the `.amplitude` to `SkyModel`.

registerrier (approved):
Thanks @QRemy . This looks good. No further comment.

Issue #4318, opened 2023-02-09
First command line in developer guide/documentation how to doesn't work anymore


MRegeard:
**Gammapy version**
dev

**Bug description**
The first command line `make docs-all`does not work anymore to build the documentation. 
One should use `toc -e build_docs`as it is shown in developer guide/how to contribute to gammapy.

PR #4319, opened 2023-02-09
updating documentation how to

MRegeard:

Pull request related to issue #4318. 

Changing `make docs-all`to `tax -e build_docs`in documentation how to.

### DIFF
--- a/docs/development/doc_howto.rst
+++ b/docs/development/doc_howto.rst
@@ -11,7 +11,7 @@ Documentation building
 
 Generating the HTML docs for Gammapy is straight-forward::
 
-    tox -e build_docs
+    make docs-sphinx
     make docs-show
 
 Generating the PDF docs is more complex.
--- a/docs/development/doc_howto.rst
+++ b/docs/development/doc_howto.rst
@@ -11,7 +11,7 @@ Documentation building
 
 Generating the HTML docs for Gammapy is straight-forward::
 
-    make docs-all
+    tox -e build_docs
     make docs-show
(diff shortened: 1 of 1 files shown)


docs/development/doc_howto.rst
```diff
@@ -11,7 +11,7 @@ Documentation building
 
 Generating the HTML docs for Gammapy is straight-forward::
 
-    make docs-all
+    tox -e build_docs
```
registerrier, line 14: technically `make docs-sphinx` works (docs-all is no longer required since notebooks have been replaced by python scripts). 
So I would keep it here and propose `tox -e build_docs` as another way to build the docs.

registerrier (commented):
Thanks @MRegeard. See inline comment.

docs/development/doc_howto.rst
MRegeard: But then we should put `make docs-sphinx`instead of `make docs-all`right ?

docs/development/doc_howto.rst
```diff
@@ -11,7 +11,7 @@ Documentation building
 
 Generating the HTML docs for Gammapy is straight-forward::
 
-    make docs-all
+    make docs-sphinx
     make docs-show
```
registerrier, line 16:
```suggestion

Or one can equivalently use tox::

     tox -e build_docs
```

QRemy (approved):
Thanks @MRegeard and @registerrier, no further comment.

Issue #4320, opened 2023-02-09
4fgl_dr3 equation is still broken in docs + phi_0 in broken power law spectral model


MRegeard:
**Gammapy version**
dev

**Bug description**
Even though equation of 4fgl_dr3 has been fixed in #4315, the rendering is still not good in the documentation. 

Also, phi_0 in broken power law spectral model is rendered as "phi_0" and not the greek character.

AtreyeeS:
It seems okay to me https://docs.gammapy.org/dev/user-guide/model-gallery/spectral/plot_super_exp_cutoff_powerlaw_4fgl.html#sphx-glr-user-guide-model-gallery-spectral-plot-super-exp-cutoff-powerlaw-4fgl-py

Can you elaborate exactly the issue

MRegeard:
Hi @AtreyeeS, It has already been merged from the pull request above (https://github.com/gammapy/gammapy/pull/4321). So that why this is now fine. I'm closing this issue.

PR #4321, opened 2023-02-09
fixing 4fgl and broken power law equations in docs

MRegeard:

Pull request related to #4320.

Fixing 4fgl equation by adding an indent before the equation.
Fixing broken power law equation by adding "\" before "phi_0".


QRemy (approved):
Thanks @MRegeard for this follow-up fix.

PR #4324, opened 2023-02-10
Fix Parameter init if scale is not one

QRemy:
The parameter init set the value to the factor which is incorrect if the given scale is not one. This PR fix this bug and correct the associated tests.

### DIFF
--- a/gammapy/modeling/models/tests/data/examples.yaml
+++ b/gammapy/modeling/models/tests/data/examples.yaml
@@ -30,14 +30,14 @@ components:
     type: PointSpatialModel
     parameters:
     - name: lon_0
-      value: -50.
+      value: -0.5
       scale: 0.01
       unit: deg
       min: -180.0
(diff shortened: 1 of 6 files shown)


AtreyeeS (approved):
Thanks @QRemy ! Looks good to me

adonath:
Thanks @QRemy! Will this affect the I/O of models? So when the users modified the scale value, will this change be backward compatible?

QRemy:
> Thanks @QRemy! Will this affect the I/O of models? So when the users modified the scale value, will this change be backward compatible?

It would only break files where users figured out that there is a bug and set the factor as the value if scale is different than one.
Models created from files with a scale different than one and value as value were incorrect and this will be fixed by this patch.

registerrier (approved):
Thanks @QRemy.

Can we prepare some text for the changelog to warn users of the potential issue if they manually changed the scale value.

PR #4325, opened 2023-02-10
Fix typo in SpatialModel.plot_interactive()

QRemy:
Correct a typo in the name of some plot_interactive methods definition.

### DIFF
--- a/gammapy/modeling/models/spatial.py
+++ b/gammapy/modeling/models/spatial.py
@@ -20,6 +20,7 @@ import matplotlib.pyplot as plt
 from gammapy.maps import Map, WcsGeom
 from gammapy.modeling import Parameter
 from gammapy.modeling.covariance import copy_covariance
+from gammapy.utils.deprecation import deprecated
 from gammapy.utils.gauss import Gauss2DPDF
 from gammapy.utils.regions import region_circle_to_ellipse, region_to_frame
 from gammapy.utils.scripts import make_path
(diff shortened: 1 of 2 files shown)


AtreyeeS (dismissed):
Thanks @QRemy ! 

I don't know how we plan to maintain the deprecated functions. Should they remain in the tests (and also testing that the correct warning is raised) till the next release?

adonath (dismissed):
Ops, that's embarrassing...:-) But the fix looks correct!

registerrier (commented):

We just have one issue with the deprecation. Do we apply it to a bug fix as well? 
In that case, I would add `v1.0.1` as the version.

adonath:
@registerrier I think this is a corner case. It is a bug-fix but also an API change. I'm not sure what our policy in this case see e.g. https://github.com/astropy/astropy-APEs/blob/main/APE2.rst#releases-and-backwards-compatibility

gammapy/modeling/models/spatial.py
```diff
@@ -1246,7 +1271,13 @@ def plot(self, ax=None, geom=None, **kwargs):
             geom = self.map.geom
         super().plot(ax=ax, geom=geom, **kwargs)
 
+    @deprecated("v1.1", alternative="plot_interactive")
```
QRemy, line 1274:
```suggestion
    @deprecated("v1.0.1", alternative="plot_interactive")
```

gammapy/modeling/models/spatial.py
```diff
@@ -301,6 +302,7 @@ def plot(self, ax=None, geom=None, **kwargs):
             )
         return m.plot(ax=ax, **kwargs)
 
+    @deprecated("v1.1", alternative="plot_interactive")
```
QRemy, line 305:
```suggestion
    @deprecated("v1.0.1", alternative="plot_interactive")
```

registerrier (approved):
Thanks @QRemy . Looks good.

PR #4326, opened 2023-02-10
Introduce internal data model for GTI

registerrier:
**Description**

This pull request modifies `GTI` to rely on a `Table` object that does not follow GADF. Instead, it uses start and stop `Time` columns. The `read` and `write` methods so far support only gadf.

**Dear reviewer**

### DIFF
--- a/gammapy/data/gti.py
+++ b/gammapy/data/gti.py
@@ -96,6 +96,7 @@ class GTI:
             the reference time to use in GTI definition
         """
         reference_time = Time(reference_time)
+        reference_time.format = "mjd"
 
         if not isinstance(start, Time):
             start = reference_time + u.Quantity(start)
@@ -165,7 +166,8 @@ class GTI:
         if format != "gadf":
             raise ValueError(f'Only the "gadf" format supported, got {format}')
 
-        meta = time_ref_to_dict(self.time_ref, scale="tt")
+        # Don't impose the scale. GADF does not require it to be TT
+        meta = time_ref_to_dict(self.time_ref, scale=self.time_ref.scale)
         start = self.time_start - self.time_ref
         stop = self.time_stop - self.time_ref
         table = Table({"START": start.to("s"), "STOP": stop.to("s")}, meta=meta)
--- a/gammapy/data/gti.py
+++ b/gammapy/data/gti.py
@@ -58,13 +58,10 @@ class GTI:
     - Stop: 2015-08-02T23:14:24.184 (time standard: TT)
     """
 
-    def __init__(self, table, meta=None):
+    def __init__(self, table, reference_time="2000-01-01"):
         self.table = self._validate_table(table)
-        if meta is None:
-            meta = table.meta
 
-        if "MJDREFFI" in meta:
-            self._time_ref = time_ref_from_dict(meta)
+        self._time_ref = Time(reference_time)
 
     @staticmethod
     def _validate_table(table):
@@ -101,20 +98,17 @@ class GTI:
         """
         reference_time = Time(reference_time)
 
-        if isinstance(start, Time):
-            start = (start - reference_time).to(u.s)
+        if isinstance(start, u.Quantity):
+            start = reference_time + start
 
-        if isinstance(stop, Time):
-            stop = (stop - reference_time).to(u.s)
+        if isinstance(stop, u.Quantity):
+            stop = reference_time + stop
 
-        start = u.Quantity(start, ndmin=1)
-        stop = u.Quantity(stop, ndmin=1)
-        meta = time_ref_to_dict(reference_time)
-        table = Table({"START": start.to("s"), "STOP": stop.to("s")}, meta=meta)
-        return cls(table)
+        table = Table({"START": start, "STOP": stop})
+        return cls(table, reference_time=reference_time)
 
     @classmethod
-    def read(cls, filename, hdu="GTI"):
+    def read(cls, filename, hdu="GTI", format="gadf"):
         """Read from FITS file.
 
         Parameters
@@ -123,10 +117,16 @@ class GTI:
             Filename
         hdu : str
             hdu name. Default GTI.
+        format: str
+            Input format, currently only "gadf" is supported
         """
+        if format != "gadf":
+            raise ValueError(f'Only the "gadf" format supported, got {format}')
+
         filename = make_path(filename)
         table = Table.read(filename, hdu=hdu)
-        return cls(table)
+        time_ref = time_ref_from_dict(table.meta, format="mjd", scale="tt")
+        return cls.create(table["START"].quantity, table["STOP"].quantity, time_ref)
 
     def to_table_hdu(self, format="gadf"):
         """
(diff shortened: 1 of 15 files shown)


gammapy/data/gti.py
```diff
@@ -78,21 +96,20 @@ def create(cls, start, stop, reference_time="2000-01-01"):
...
-        meta = time_ref_to_dict(reference_time)
-        table = Table({"START": start.to("s"), "STOP": stop.to("s")}, meta=meta)
-        return cls(table)
+        table = Table({"START": np.atleast_1d(start), "STOP": np.atleast_1d(stop)})
+
+        return cls(table, reference_time=reference_time)
 
     @classmethod
-    def read(cls, filename, hdu="GTI"):
+    def read(cls, filename, hdu="GTI", format="gadf"):
```
bkhelifi, line 112: The hdu name "GTI" is for the GADF format... Should we assume that the hdu name is independent of the format?

gammapy/data/gti.py
```diff
@@ -101,10 +118,36 @@ def read(cls, filename, hdu="GTI"):
...
+        with fits.open(str(make_path(filename)), memmap=False) as hdulist:
+            return cls.from_table_hdu(hdulist[hdu], format=format)
+
+    @classmethod
+    def from_table_hdu(cls, table_hdu, format="gadf"):
+        """Read from table HDU.
+
+        Parameters
+        ----------
+        table : `~astropy.io.fits.BinTableHDU`
```
bkhelifi, line 134:
```suggestion
        table_hdu : `~astropy.io.fits.BinTableHDU`
```

registerrier:
Regarding the impact on `Observations.select_time`, the computing time is actually reduced with the new approach. The following code snippet runs in 9s with the current main branch and 4s with the proposed model.
```
from astropy.time import Time

ds = DataStore.from_dir("$GAMMAPY_DATA/hess-dl3-dr1/")
observations = ds.get_observations()

interval1 = [Time('2004-12-04 22:10:10.184'), Time('2004-12-04 22:18:10.184')]
interval2 = [Time('2004-12-04 22:19:00'), Time('2004-12-04 22:38:00')]
interval3 = [Time('2004-12-04 22:50:00'), Time('2004-12-04 23:01:00')]

%timeit filtered = observations.select_time([interval1, interval2, interval3])
```

gammapy/data/gti.py
```diff
@@ -172,14 +220,22 @@ def time_sum(self):
...
     def time_stop(self):
         """GTI end times (`~astropy.time.Time`)."""
-        met = u.Quantity(self.table["STOP"].astype("float64"), "second")
-        return self.time_ref + met
+        return self.table["STOP"]
+
+    @property
+    def met_start(self):
+        """GTI start time difference with reference time in sec, MET (`~astropy.units.Quantity`)."""
+        return (self.time_start - self.time_ref).to("s")
```
maxnoe, line 233: Why go through `self.time_start` here if this is by definition what is stored in the table?

maxnoe:
Any specific reason to use a table with time columns over two time arrays?

gammapy/data/gti.py
registerrier: I am not sure the hduname is really imposed by the gadf.

registerrier:
> Any specific reason to use a table with time columns over two time arrays?

The main reason would be that it guarantees that start and stop times stay in phase and cannot be sorted independently. A minor one would be that we still have to allow access to the `GTI.table` but we could deprecate it.

AtreyeeS (commented):
Thanks @registerrier ! This looks good to me. I suppose the internal format should be defined somewhere - how does the user know the default time scale / the table column names...

registerrier:
I have updated the PR and it is ready. 

- Now the `GTI` are represented internally by a `Table` with `START` and `STOP` columns of type `Time`. This has the advantage that start and stop stay in phase when stacking or reordering GTIs.
- The table is validated on init. For now, no specific scale is imposed. Note that previously, it was mostly `TT`. Do we want to enforce that all `GTIs` are in `TT`?
- The reference time is no longer required and should become de-facto metadata. Since the scheme for meta is not yet defined, I have made it a property to stay backward compatible with `GTI.time_ref`
- I have introduced a default value for the reference time which is defined as a global variable `TIME_REF_DEFAULT`. Ideally, this should be part of a set of global configurable variables. 
- I have introduced `met_start` and `met_stop` methods for backward compatibility  with the previous `GTI.table["START"]`

Do you have time for a review @maxnoe or @kosack ?

gammapy/data/gti.py
```diff
@@ -190,27 +258,29 @@ def time_intervals(self):
...
-        start = Time([_[0] for _ in time_intervals]) - reference_time
-        stop = Time([_[1] for _ in time_intervals]) - reference_time
-        meta = time_ref_to_dict(reference_time)
-        table = Table({"START": start.to("s"), "STOP": stop.to("s")}, meta=meta)
-        return cls(table=table)
+        start = Time([_[0] for _ in time_intervals])
+        stop = Time([_[1] for _ in time_intervals])
+
+        if reference_time is None:
+            reference_time = start.min()
```
adonath, line 281: The behavior is well documented, but maybe we should still use the same `TIME_REF_DEFAULT` instead?

adonath (approved):
Thanks @registerrier! I have left two minor comments, otherwise looks good to me!

AtreyeeS (approved):
Thanks @registerrier ! No comments at this stage. It will be nice to try this out

gammapy/data/gti.py
```diff
@@ -101,10 +128,36 @@ def read(cls, filename, hdu="GTI"):
             Filename
         hdu : str
             hdu name. Default GTI.
+        format: str
+            Input format, currently only "gadf" is supported
         """
         filename = make_path(filename)
-        table = Table.read(filename, hdu=hdu)
-        return cls(table)
+        with fits.open(str(make_path(filename)), memmap=False) as hdulist:
```
maxnoe, line 135: Why was this change needed? Everywhere else, we use the simple `Table.read`

registerrier: To call directly the `GTI.from_table_hdu()` method that actually does the gadf to gammapy format conversion.
Alternatively a staticmethod such as `gadf_table_to_gammapy(table)` could also be used.

adonath: I guess the question is whether meta data accessed via `table.meta` or `hdu.header`. I have not noticed this before but the pattern of `.from_table_hdu()` is unusual. We typically rely on the meta data being present in `table.meta`.

There is still https://github.com/gammapy/gammapy/blob/main/gammapy/maps/axes.py, but if I remember correctly this is only there for backwards compatibility, so we could actually deprecate it. 

@registerrier I think it is worth doing a follow-up PR to change to `GTI.from_table(..., format="gadf")` and not introduce `.from_table_hdu()`...

adonath: We could keep `.from_table_hdu()` but all it does is really:

```python3
table = Table.read(table_hdu)
return cls.from_table(table, )
```

adonath: Not sure it's worth to keep, however it saves one line of code...

Issue #4239, opened 2022-12-06
EventList definition


maxnoe:
**Is your feature request related to a problem? Please describe.**

Part of the ongoing work to support multiple data formats / versions and #3767 

* [ ] Define columns of EventList
* [ ] remove reliance on meta dict in properties
* [ ] add methods to convert to / from fits / GADF
* [ ] validate table in `__init__`

PR #4334, opened 2023-02-15
Replace deprecated property Observation.pointing_radec

QRemy:
Replace deprecated method Observation.pointing_radec() by Observation.get_pointing_icrs() in get_irfs_features

### DIFF
--- a/examples/tutorials/analysis-3d/cta_data_analysis.py
+++ b/examples/tutorials/analysis-3d/cta_data_analysis.py
@@ -129,7 +129,7 @@ maker = MapDatasetMaker(selection=["counts", "background", "exposure", "psf"])
 maker_safe_mask = SafeMaskMaker(methods=["offset-max"], offset_max=2.5 * u.deg)
 
 for obs in observations:
-    cutout = stacked.cutout(obs.pointing_radec, width="5 deg")
+    cutout = stacked.cutout(obs.get_pointing_icrs(obs.tmid), width="5 deg")
     dataset = maker.run(cutout, obs)
     dataset = maker_safe_mask.run(dataset, obs)
     stacked.stack(dataset)
@@ -292,7 +292,7 @@ plot_spectrum_datasets_off_regions(datasets, ax=ax)
 # Model fit
 # ~~~~~~~~~
 #
-# The next step is to fit a spectral model, using all data (i.e. a
+# The next step is to fit a spectral model, using all data (i.e. a
 # “global” fit, using all energies).
 #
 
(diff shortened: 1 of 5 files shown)


QRemy:
On a side note maybe we could have `time` optional in `obsservation.get_pointing_icrs()` and if time is None : time =self.tmid
because in most of cases we just use tmid.

bkhelifi:
Hi @QRemy ,

If you start to replace this deprecated function, maybe this PR should update at the same time much more files:
- documentation: EventList.rst, ObservationTable.rst, Observation.rst,
- in the tutorials cta.py, hess.py, makers.py, analysis_2.py, cta_data_analysis.py

A side comment:
In `gammapy/data/event_list.py`, we have the function `pointing_radec` that is not deprecated. Is it normal?

adonath:
> On a side note maybe we could have time optional in obsservation.get_pointing_icrs() and if time is None : time =self.tmid because in most of cases we just use tmid.

I commented about this as well https://github.com/gammapy/gammapy/pull/4220/files#r1082899314

I think the compromise would be to introduce a property `obs.pointing_icrs_tmid`, which make it explicit that the mid of runtime is used. What do you think @maxnoe?

QRemy:
> I think the compromise would be to introduce a property `obs.pointing_icrs_tmid`, which make it explicit that the mid of runtime is used. What do you think @maxnoe?

and should `pointing_radec` in `EventList` and `ObservationTable` be renamed in the same way for consistency ?

maxnoe:
> What do you think @maxnoe?

My opinion is still that we should it make really explicit that an assumption is made in the code that the pointing doesn't change although in general it could / does change with time in at least one reference system. 

Code should make it easy to use correct (pointing changes with time) and hard to be used wrong (assume fixed pointing).

You could argue that introduction a `observation.pointing_icrs_tmid` makes it easier to later find the places where we do make this assumption. But it also makes it convenient to introduce this assumption to new places, which is probably something we don't want.

This is connected to e.g. these TODOs here:

https://github.com/gammapy/gammapy/blob/32393ccdba5b5665a7c51b9e4d3111009e84e35e/gammapy/makers/utils.py#L154-L161
for which I didn't find an issue, so maybe one should be created.

maxnoe:
> and should pointing_radec in EventList and ObservationTable be renamed in the same way for consistency ?

It should probably be deprecated on the EventList, see also #4239 

Since in general, pointing can be separate from the `EventList`, only the `FixedPointingInfo` lives on the `EventList` and could be exposed there optionally.

adonath (approved):
Thanks @QRemy, the PR can go in as is. Let's keep question of introducing the shortcut for `pointing_icrc_tmid` independent...

registerrier (approved):
Thanks @QRemy. Looks good!

registerrier:
CI fails are unrelated. docs build works. Merging.

PR #4335, opened 2023-02-15
Fixing docstring for stat_sum in several dataset class

MRegeard:
The docstring for `stat_sum`in sevral dataset class was saying "Likelihood", which is misleading. 

Fixed by replacing it by "statistic function value". 

Question: Do we stick to this also in `Dataset` which says "Total statistic"?

### DIFF
--- a/gammapy/datasets/core.py
+++ b/gammapy/datasets/core.py
@@ -216,7 +216,7 @@ class Datasets(collections.abc.MutableSequence):
         return np.array(contributions)
 
     def stat_sum(self):
-        """Compute joint likelihood"""
+        """Compute joint statistic function value"""
         stat_sum = 0
         # TODO: add parallel evaluation of likelihoods
         for dataset in self:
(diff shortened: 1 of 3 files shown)


adonath:
Thanks @MRegeard, indeed this was misleading. Could even add the correct definition of `stat = -2 * log(L)`... what do you think @registerrier ?

MRegeard:
> Thanks @MRegeard, indeed this was misleading. Could even add the correct definition of `stat = -2 * log(L)`... what do you think @registerrier ?

But from what I understand it's not even a proper Likelihood, because it is not normalised, right ?

registerrier:
> > Thanks @MRegeard, indeed this was misleading. Could even add the correct definition of `stat = -2 * log(L)`... what do you think @registerrier ?
> 
> But from what I understand it's not even a proper Likelihood, because it is not normalised, right ?

Well it is defined as `-2 log(L) + cst` strictly speaking. 
I think it is fine like this.

registerrier (approved):

adonath (approved):
No further comments from my side either.

registerrier:

Issue #3743, opened 2022-01-26
Allow following BiPM recommendations on axis labels


maxnoe:
**Is your feature request related to a problem? Please describe.**

Gammapy plotting functions typeset units as `quantity [unit]`.

 Which, while being common in HEP along with the alternative `quantity (unit)`, is wrong¹ and against the recommendations of the BiPM on how our unit system works:

See section 5.4, Rules and style conventions for expressing values of quantities of the SI Brochure here:

https://www.bipm.org/documents/20126/41483022/SI-Brochure-9-EN.pdf

> Symbols for units are treated as mathematical entities. In expressing the value of a quantity as
the product of a numerical value and a unit, both the numerical value and the unit may be
treated by the ordinary rules of algebra. This procedure is described as the use of quantity
calculus, or the algebra of quantities. For example, the equation p = 48 kPa may equally be
written as p/kPa = 48. It is common practice to write the quotient of a quantity and a unit in
this way for a column heading in a table, so that the entries in the table are simply numbers.
For example, a table of the velocity squared versus pressure may be formatted as shown
below. The axes of a graph may also be labelled in this way, so that the tick marks are labelled
only with numbers, as in the graph below.

¹ According to the people who's job it is to define our unit system

**Describe the solution you'd like**

Allow to use the recommended format for typesetting quantities. If not by default, then at least optionally.

adonath:
Thanks @maxnoe! While I don't have a string opinion here I'd be happy to adopt the recommendation and put the SI brochure as a reference. This mostly concerns #3740...

adonath:
Now one can actually do:

```python3
from gammapy.maps import axes

axes.DEFAULT_LABEL_TEMPLATE = "{quantity} / {unit}"
```
Do you reuest anything else here @maxnoe?

maxnoe:
no, great!

maxnoe:
It seems that this is not fully implemented yet. 

E..g. `Model.plot` seems to ignore this setting, reopening

maxnoe:
Same for most of the IRF plotting methods, e.g. `EffectiveArea2D.plot_energy_depedence`

PR #4336, opened 2023-02-16
Change label units within parentheses to brackets

aaguasca:
**Description**

Depending on the plots, units are displayed within parentheses or brackets. This PR homogenises the representation of units in plot labels using brackets instead of parentheses.

**Dear reviewer**

The PR is ready for review. I think I did not forget any parentheses.

### DIFF
--- a/docs/user-guide/astro/population/plot_spiral_arm_models.py
+++ b/docs/user-guide/astro/population/plot_spiral_arm_models.py
@@ -32,8 +32,8 @@ for spiralarm_index in range(4):
 
 ax_cartesian.plot(vallee_spiral.bar["x"], vallee_spiral.bar["y"])
 
-ax_cartesian.set_xlabel("x (kpc)")
-ax_cartesian.set_ylabel("y (kpc)")
+ax_cartesian.set_xlabel("x [kpc]")
+ax_cartesian.set_ylabel("y [kpc]")
 ax_cartesian.set_xlim(-12, 12)
 ax_cartesian.set_ylim(-15, 12)
 ax_cartesian.legend(ncol=2, loc="lower right")
--- a/examples/tutorials/analysis-1d/cta_sensitivity.py
+++ b/examples/tutorials/analysis-1d/cta_sensitivity.py
@@ -175,8 +175,8 @@ ax.plot(
 )
 
 ax.loglog()
-ax.set_xlabel(f"Energy ({t['energy'].unit})")
-ax.set_ylabel(f"Sensitivity ({t['e2dnde'].unit})")
+ax.set_xlabel(f"Energy [{t['energy'].unit}]")
+ax.set_ylabel(f"Sensitivity [{t['e2dnde'].unit}]")
 ax.legend()
 
 
@@ -192,11 +192,11 @@ fig, ax1 = plt.subplots()
 ax1.plot(t["energy"], t["background"], "o-", color="black", label="blackground")
 
 ax1.loglog()
-ax1.set_xlabel(f"Energy ({t['energy'].unit})")
+ax1.set_xlabel(f"Energy [{t['energy'].unit}]")
 ax1.set_ylabel("Expected number of bkg counts")
 
 ax2 = ax1.twinx()
-ax2.set_ylabel(f"ON region radius ({on_radii.unit})", color="red")
+ax2.set_ylabel(f"ON region radius [{on_radii.unit}]", color="red")
 ax2.semilogy(t["energy"], on_radii, color="red", label="PSF68")
 ax2.tick_params(axis="y", labelcolor="red")
 ax2.set_ylim(0.01, 0.5)
--- a/examples/tutorials/analysis-3d/flux_profiles.py
+++ b/examples/tutorials/analysis-3d/flux_profiles.py
@@ -201,7 +201,7 @@ fig, ax = plt.subplots()
 for quantity in quantities:
     profile[quantity].plot(ax=ax, label=quantity.title())
 
-ax.set_ylabel("Counts ")
+ax.set_ylabel("Counts")
 
 
 ######################################################################
--- a/examples/tutorials/api/fitting.py
+++ b/examples/tutorials/api/fitting.py
@@ -236,7 +236,7 @@ for ax, par in zip(axes, datasets.parameters.free_parameters):
     name = datasets.models.parameters_unique_names[idx]
     profile = fit.stat_profile(datasets=datasets, parameter=par)
     ax.plot(profile[f"{name}_scan"], profile["stat_scan"] - total_stat)
-    ax.set_xlabel(f"{par.name} {par.unit}")
+    ax.set_xlabel(f"{par.name} [{par.unit}]")
     ax.set_ylabel("Delta TS")
     ax.set_title(f"{name}:\n {par.value:.1e} +- {par.error:.1e}")
 
--- a/gammapy/data/event_list.py
+++ b/gammapy/data/event_list.py
@@ -475,7 +475,7 @@ class EventList:
         time = self.table["TIME"]
         time = time - np.min(time)
 
-        ax.set_xlabel("Time (sec)")
+        ax.set_xlabel("Time [sec]")
         ax.set_ylabel("Counts")
         y, x_edges = np.histogram(time, bins=20)
 
@@ -559,7 +559,7 @@ class EventList:
         with quantity_support():
             ax.hist(offset2, **kwargs)
 
-        ax.set_xlabel(f"Offset^2 ({ax.xaxis.units})")
+        ax.set_xlabel(f"Offset^2 [{ax.xaxis.units}]")
         ax.set_ylabel("Counts")
         return ax
 
--- a/gammapy/estimators/points/core.py
+++ b/gammapy/estimators/points/core.py
@@ -614,7 +614,7 @@ class FluxPoints(FluxMaps):
 
         axis.format_plot_xaxis(ax=ax)
 
-        ax.set_ylabel(f"{sed_type} ({ax.yaxis.units})")
+        ax.set_ylabel(f"{sed_type} [{ax.yaxis.units}]")
         ax.set_yscale("log")
 
         if add_cbar:
--- a/gammapy/estimators/profile.py
+++ b/gammapy/estimators/profile.py
@@ -325,7 +325,7 @@ class ImageProfile:
         kwargs.setdefault("alpha", 0.5)
 
         ax.fill_between(x, ymin, ymax, **kwargs)
-        ax.set_xlabel("x (deg)")
+        ax.set_xlabel("x [deg]")
         ax.set_ylabel("profile")
         return ax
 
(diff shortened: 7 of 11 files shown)


registerrier (commented):
Thanks @aaguasca . This is nice.

What is the best convention, brackets do not seem to be generally accepted as a good convention. 

In journals we find some suggesting to use parenthesis (e.g. https://journals.aps.org/authors/axis-labels-and-scales-on-graphs-h18). 

The NIST convention is even different (https://www.nist.gov/pml/special-publication-811/nist-guide-si-chapter-7-rules-and-style-conventions-expressing-values) requiring to write v / (km/s). BTW, this document uses a convention where [v] means the unit of v.

Maybe the parenthesis is a better default. Opinions @adonath @bkhelifi @maxnoe ?

QRemy:
Maybe we could add an helper function that adapt the label formatting for different journal conventions
`def format_label(name, unit, convention)`

maxnoe:
> The NIST convention is even different 

This is the only "correct" solution, it's the one prescribed by the people defining our unit system.

See https://github.com/gammapy/gammapy/issues/3743 and references linked there-in.

adonath:
Thanks a lot @aaguasca! 

Before v1.0 we already had one PR to unify the plot labels, that rely on `MapAxis`. There was a similar discussion on which convention to use. So we decided to just introduce a helper variable [`DEFAULT_LABEL_TEMPLATE`](https://github.com/gammapy/gammapy/blob/main/gammapy/maps/axes.py#L70). Would you mind just changing the remaining places to use the variable as well? Then there is a single (!) place where users can modify this for all the plots.

adonath:
@QRemy This basically exists here https://github.com/gammapy/gammapy/blob/main/gammapy/maps/axes.py#L70

docs/user-guide/astro/population/plot_spiral_arm_models.py
```diff
@@ -32,8 +32,8 @@
 
 ax_cartesian.plot(vallee_spiral.bar["x"], vallee_spiral.bar["y"])
 
-ax_cartesian.set_xlabel("x (kpc)")
-ax_cartesian.set_ylabel("y (kpc)")
+ax_cartesian.set_xlabel("x [kpc]")
+ax_cartesian.set_ylabel("y [kpc]")
```
adonath, line 36: For example change to `DEFAULT_LABEL_TEMPLATE.format(quantity="x", unit="kpc")` here...

examples/tutorials/analysis-1d/cta_sensitivity.py
```diff
@@ -175,8 +175,8 @@
 )
 
 ax.loglog()
-ax.set_xlabel(f"Energy ({t['energy'].unit})")
-ax.set_ylabel(f"Sensitivity ({t['e2dnde'].unit})")
+ax.set_xlabel(f"Energy [{t['energy'].unit}]")
+ax.set_ylabel(f"Sensitivity [{t['e2dnde'].unit}]")
```
adonath, line 179: There is the question, whether we should make the `DEFAULT_LABEL_TEMPLATE` public and use it here...maybe not...I guess it distracts from the content too much.

adonath:
As we tried to unify before v1.0 but seem to have missed some places this PR could go as a bug fix...

adonath:
@aaguasca When you find circular imports, then we could move `DEFAULT_LABEL_TEMPLATE` to `gammapy.visualization`.

adonath:
If we think this is important to show we could have an "how to entry" of how to modify `DEFAULT_LABEL_TEMPLATE`.

maxnoe:
> If we think this is important to show we could have an "how to entry" of how to modify DEFAULT_LABEL_TEMPLATE.

Modifying a module variable is tricky, because:

```
from gammapy.visualization import DEFAULT_LABEL_TEMPLATE

# this does not work, it sets a new local variable instead of modifying the module
# variable
DEFAULT_LABEL_TEMPLATE = "{label} / {unit}"
```

this works:

```
from gammapy import visualization

# this does not work, it sets a new local variable instead of modifying the module
# variable
visualization.DEFAULT_LABEL_TEMPLATE = "{label} / {unit}"
```


So functions are preferable I'd say. (set_label_format / format_label)

examples/tutorials/analysis-1d/cta_sensitivity.py
maxnoe: It's also dangerous, because of how python globals work.

I would prefer functions, as these are "safer" for users.

How about the a function:

```
def format_label(quantity, unit=None, template=None, convention="divide")
```

that would also allow for the cases where we don't have units (and where the current approach would render something wrong like "foo / " or "foo []".

adonath:
Thanks for laying out the options @maxnoe! For now I don't have a strong opinion on how to support it. I think in this case the documentation is more important...

aaguasca:
Thank you to you all for this very interesting discussion. I am not sure which is the best solution, but maybe using a function is safer to use, clearer, and more versatile. 

Following @maxnoe suggestion, I tried to explore the function implementation (it is just a draft):

```
def set_label_format(quantity, unit=None, template=None, convention="divide"):

    DEFAULT_CONVENTION_LIST=[
        "divide",
        "parentheses",
        "brackets"
    ]
    
    if template is None:
    
        if unit is None:
            label = f"{quantity}"
        else:
            if convention not in DEFAULT_CONVENTION_LIST:
                raise NameError(f"Specified convention is not known.")            
            else:
                if convention == "divide":
                    unit_form = f"\ ({unit})"

                elif convention == "parentheses":
                    unit_form = f"({unit})"

                elif convention == "brackets":
                    unit_form = f"[{unit}]"
                
                label = f"{quantity} {unit_form}"

    else:
        label=template.format(
            quantity=quantity,
            unit=unit
        )
        
    return label
```

, which it is easy to use:

```
ylabel=set_label_format(
    quantity="e2dnde", 
    unit=ax.yaxis.units, 
    convention="divide",
)
ax.set_ylabel(ylabel)
```

aaguasca:
> I'd call that `format_label` and the "convention" case needs a case for `unit=None` I think, other than that it looks good.

I am not sure if the case of "convention" with `unit=None` is required. If one does not specify the units, for example, a histogram with counts (not normalised), the units are not required, and you only display the quantity. So whatever the "convention" you specify, you do not need to plot the units. I think this is the same situation you mentioned before with "foo / " or "foo []".
 

> The `set_label_format` was meant to set the Gammapy default for plots where the users don't have direct control.
> 
> I.e. in `gammapy.visualization`:
> 
> ```python
> _PLOTTING_OPTIONS = {
>     "label_format": "divide"
> }
> 
> def set_label_format(convention="divide", template=None):
>     # modify _PLOTTING_OPTIONS
> ```

I am sorry, but I am not sure I understood it. Do you mean, for example, having in `gammapy.visualization`:


```python
_PLOTTING_OPTIONS = {
    "convention": "divide"
}

def set_label_format(convention="divide", template=None):
    # modify _PLOTTING_OPTIONS
    
    if template is None:
        _PLOTTING_OPTIONS["convention"] = convention
    
    else:
        _PLOTTING_OPTIONS.pop('convention', None)
        _PLOTTING_OPTIONS["template"] = template
        
    return _PLOTTING_OPTIONS
```

Then, when Gammapy plots a default figure without the user's control, for example, in the method `plot()` from`gammapy.estimators.points.core`: https://github.com/gammapy/gammapy/blob/a0da8a27a50c0296dfb7fc740886616481ea04f3/gammapy/estimators/points/core.py#L466
Should the code do the following?

```python
def plot(self, ax=None, sed_type=None, energy_power=0, **kwargs):
    # ...
    # ... up to L524
    axis_label_format = set_label_format()
    ylabel = format_label(quantity=sed_type, unit=ax.yaxis.units, **axis_label_format)
    ax.set_ylabel(ylabel)
    # ...
```

maxnoe:
> So whatever the "convention" you specify, you do not need to plot the units. I think this is the same situation you mentioned before with "foo / " or "foo []".

Exactly, so you need a special case, right? To prevent the trailing "/" or empty "[]"

> Should the code do the following?

The idea would be that *users* call `set_label_format` to modify the default gammapy uses and gammapy plotting functions just call `format_label` which will lookup the current default I think.

aaguasca:
> Exactly, so you need a special case, right? To prevent the trailing "/" or empty "[]"

Not really. In the function above (https://github.com/gammapy/gammapy/pull/4336#issuecomment-1433424882), if I am not wrong, this line prevents the trailing  "/" or empty "[]":
```python
if unit is None:
    label = f"{quantity}"
```

> The idea would be that users call set_label_format to modify the default gammapy uses and gammapy plotting functions just call format_label which will lookup the current default I think.

Okay, I think I understand it now. Thanks, @maxnoe!

In the end, which implementation do you prefer to use, function or module variable, @adonath @maxnoe?

maxnoe:
> Not really. In the function above (https://github.com/gammapy/gammapy/pull/4336#issuecomment-1433424882), if I am not wrong, this line prevents the trailing "/" or empty "[]":

Yes, this is for the built-in conventions case. What about the "template" case. I.e. the user sets a custom template for gammapy and gammapy uses it for the "unit=None" case?

aaguasca:
> What about the "template" case. I.e. the user sets a custom template for gammapy and gammapy uses it for the "unit=None" case?

Okay, you are right. Displaying None units using the template makes no sense. Below you can find the draft of the function with the template section more developed (but without documentation):

```python
def format_label(quantity, unit=None, template=None, convention="divide"):

    DEFAULT_CONVENTION_LIST=[
        "divide",
        "parentheses",
        "brackets"
    ]
    
    if template is None:
        if unit is None:
            label = f"{quantity}"
        else:
            if convention not in DEFAULT_CONVENTION_LIST:
                raise NameError(f"Specified convention is not known.")
            else:
                if convention == "divide":
                    unit_form = f"\ ({unit})"

                elif convention == "parentheses":
                    unit_form = f"({unit})"

                elif convention == "brackets":
                    unit_form = f"[{unit}]"
                
                label = f"{quantity} {unit_form}"

    else:
        if "{quantity}" not in template and "{unit}" not in template:
            raise NameError("Template does not contain neither " +\
                            "{quantity} nor {unit}. Specify at least one.")
        else:
            if "{quantity}" not in template:
                log.warning("Quantity attribute is not in the template, "+\
                            "if required, add {quantity} to the template.")
                
                if unit is None:
                    label = ""  
                else:
                    label = template.format(
                        unit=unit
                    )                
                
            elif "{unit}" not in template:
                log.warning("Unit attribute is not in the template, "+\
                            "if required, add {unit} to the template.")     
        
                label = template.format(
                    quantity=quantity,
                )
            
            else:
                label = template.format(
                    quantity=quantity,
                    unit=unit
                )                  
    return label
```

adonath:
@aaguasca and @maxnoe 

Let's not try to "over-engineer" the problem, simplicity counts! I think the `DEFAULT_LABEL_TEMPLATE` approach is just fine, as long as it's properly documented. Yes, there are other cases to be covered, but keep in mind that users can always overwrite the label with one (!) line of code like `ax.set_xlabel()` or `ax.set_ylabel()`.  I think the justification to introduce a method with > 3 lines of code is not really there...

registerrier (approved):
Thanks @aaguasca . This looks good.

After discussion in the dev call, it was decided that this is a follow-up of the uniformisation of the plot labels performed before v1.0. It is therefore considered a bug-fix. The milestone is v1.0.1.
Let's keep this PR simple. Further options for configuration can be kept for later with a different milestone, hence in another PR/issue.

registerrier:
CI fails are unrelated. Merging now.

PR #4344, opened 2023-02-20
Fix norm_only_changed in MapEvaluator

QRemy:
This PR fix MapEvaluator.norm_only_changed property which is likely always False currently as noticed by @tunbehaun273.

### DIFF
--- a/gammapy/datasets/evaluator.py
+++ b/gammapy/datasets/evaluator.py
@@ -105,6 +105,15 @@ class MapEvaluator:
         return self.exposure.geom
 
     @property
+    def _geom_reco(self):
+        if self.edisp is not None:
+            energy_axis = self.edisp.axes["energy"].copy(name="energy")
+        else:
+            energy_axis = self.geom.axes["energy_true"].copy(name="energy")
+        geom = self.geom.to_image().to_cube(axes=[energy_axis])
+        return geom
+
+    @property
     def needs_update(self):
         """Check whether the model component has drifted away from its support."""
         # TODO: simplify and clean up
@@ -361,7 +370,7 @@ class MapEvaluator:
                 self._norm_idx is not None
                 and self.model.parameters.value[self._norm_idx] == 0
             ):
-                npred = Map.from_geom(self.geom, data=0)
+                npred = Map.from_geom(self._geom_reco, data=0)
             elif not self.parameter_norm_only_changed or not self.use_cache:
                 for method in self.methods_sequence:
                     values = method(self._computation_cache)
--- a/gammapy/datasets/evaluator.py
+++ b/gammapy/datasets/evaluator.py
@@ -357,7 +357,12 @@ class MapEvaluator:
         if isinstance(self.model, TemplateNPredModel):
             npred = self.model.evaluate()
         else:
-            if not self.parameter_norm_only_changed or not self.use_cache:
+            if (
+                self._norm_idx is not None
+                and self.model.parameters.value[self._norm_idx] == 0
+            ):
+                npred = Map.from_geom(self.geom, data=0)
+            elif not self.parameter_norm_only_changed or not self.use_cache:
                 for method in self.methods_sequence:
                     values = method(self._computation_cache)
                     self._computation_cache = values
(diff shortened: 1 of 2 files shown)


gammapy/datasets/evaluator.py
```diff
@@ -406,8 +406,8 @@ def parameter_norm_only_changed(self):
         norm_only_changed = False
         idx = self._norm_idx
         values = self.model.parameters.value
-        if idx and self._computation_cache is not None:
-            changed = self._cached_parameter_values_previous == values
+        if idx is not None and self._computation_cache is not None:
+            changed = self._cached_parameter_values_previous != values
             norm_only_changed = sum(changed) == 1 and changed[idx]
```
maxnoe, line 411: `np.count_nonzero` instead of `sum`

maxnoe:
```
In [3]: %timeit sum(changed)
3 µs ± 109 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)

In [4]: %timeit np.sum(changed)
4.06 µs ± 140 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)

In [5]: %timeit np.count_nonzero(changed)
706 ns ± 40 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

tunbehaun273:
This will probably fix the issue for most cases but I think the lines `if changed:
            self._cached_parameter_values = values` might also be problematic in some cases.
For the caching of the UnbinnedEvaluator (https://github.com/gammapy/gammapy-unbinned-analysis/blob/main/EventDataset/UnbinnedEvaluator.py) I went for a different approach: 

1. I check if a parameter changed and if this is only the norm parameter **but** I do not reset the cached parameter values
2. Dependent on the case I return the cache or simply rescale the cache or fully re-compute the cache.
3. Only after the cache is rescaled or re-computed I set the cached parameter values to the ones for which the cache is computed.
Like this I also don't need the "previous" parameters for renormalization. To me this approach feels more natural and robust but maybe I am also missing some reasons for the existing method.

QRemy:
The reason we have the `_cached_parameter_previous` rather than only `_cached_parameter`  in the `MapEvaluator` is that `_computation_cache` is updated only when `norm_only_changed` is False, while in the `UnbinnedEvaluator` you also update the `_computation_cache` if `norm_only_changed` is True.

Looking at the code in the `UnbinnedEvaluator` I have two questions :

`_cached_parameter_previous` seems to be used but never updated
https://github.com/gammapy/gammapy-unbinned-analysis/blob/432eb63661595e80843622184212be8757dcb405/EventDataset/UnbinnedEvaluator.py#L332

Where values is defined ?
https://github.com/gammapy/gammapy-unbinned-analysis/blob/432eb63661595e80843622184212be8757dcb405/EventDataset/UnbinnedEvaluator.py#L289
https://github.com/gammapy/gammapy-unbinned-analysis/blob/432eb63661595e80843622184212be8757dcb405/EventDataset/UnbinnedEvaluator.py#L295

tunbehaun273:
@QRemy that makes sense. That is of course also a way of doing it. 
Concerning your questions, those were all bugs in my code and I pushed a fix.

gammapy/datasets/evaluator.py
```diff
@@ -406,9 +404,9 @@ def parameter_norm_only_changed(self):
         norm_only_changed = False
         idx = self._norm_idx
```
registerrier, line 405: Looking at this part here, I wonder if the definition of `_norm_idx` is optimal. It does not rely on the `Parameter.is_norm` property but on the names of the parameters. 
This could be improved since `norm_idx` is `None` if both `amplitude` and `norm` parameters are present.

registerrier (approved):

I have left a question inline.

Does it change the performance? I suppose the `norm_only_changed` was never true. In particular, for flux point computation do you see an improvement with this?

QRemy:
> Does it change the performance? I suppose the `norm_only_changed` was never true. In particular, for flux point computation do you see an improvement with this?

Yes, `compute_npred` in `the spectrum_1d_joint` benchmark is about 3x faster.

Before
Screenshot 2023-03-03 at 13 54 09

After
Screenshot 2023-03-03 at 13 54 41

In that case I think the profiler do not even report the  psf and edisp operations as they happened only once.

registerrier (approved):
Thanks @QRemy . No further comment.

PR #4348, opened 2023-02-24
Add a run function in DataStoreChecker

bkhelifi:
**Description**

This pull request adds a run function that propagate correctly the outputs of the checks. It was astonishingly missing!

**Dear reviewer**
In the current implementation, the dictionary `CHECKS` is not used and maybe a better implementation can be done.

Also, this class is hidden. Should we expose it?


registerrier (changes requested):
Thanks @bkhelifi . Normally this is already working fine. What `DataStore.check()` returns is a generator.

PR #4350, opened 2023-02-28
Always use FixedPointingInfo from events header in DataStore

maxnoe:
**Description**

This pull request fixes the  behavior of the data store for files that do contain the `POINTING` table.
Gammapy doesn't support using the pointing table currently, so the code in the data store that didn't read the `FixedPointingInfo` from the events header in case the `POINTING` table is there does not make sense currently.

To be changed once Gammapy actually uses the pointing table.

### DIFF
--- a/gammapy/data/data_store.py
+++ b/gammapy/data/data_store.py
@@ -322,11 +322,11 @@ class DataStore:
                 f"Required HDUs {missing_hdus} not found in observation {obs_id}"
             )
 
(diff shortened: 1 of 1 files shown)


maxnoe:
> Thanks @maxnoe . Shouldn't gammapy fail if a `pointing` hdu is required?

I am not sure what you mean. 

In this case it was enough that the pointing hdu existed in the file, it was not included in `required_irf` if that's what you mean

registerrier (approved):
Thanks @maxnoe . This looks fine.

Issue #4340, opened 2023-02-17
PhaseBackgroundMaker does not support energy-dependent directional cuts


alvmas:
**Gammapy version**
 Gammapy v1.0

**Bug description**
When one tries to do the analysis estimating OFF counts based on the PHASE information and using an energy-dependent rad_max value, the PhaseBackgroundMaker does not work correctly since the geometry is defined as a PointSkyRegion.

**Expected behavior**
It should compute the OFF counts based on the phase information, but using different rad_max for the ON region, in different energy bins. 

**Other information**
We should have an option to compute the OFF counts in the variable ON region, for the case of using a PointSkyRegion geometry. Something similar to what is already implemented in ReflectedRegionsBackgroundMaker.

MRegeard:
Hi @alvmas, 

The rad max cuts should be implemented with the pull request above.

alvmas:
Hi @MRegeard ,
Thanks, I was actually implementing a PR that was exactly the same, so I think this should work!

registerrier:
This is solved in #4352 . Closing now.

PR #4352, opened 2023-02-28
Adding rad max cut in PhaseBackgroundMaker

MRegeard:
This is a pull request related to issue #4340. 

It allows for rad max cut in the `PhaseBackgroundMaker`. 

I still need to create test for that.

### DIFF
--- a/gammapy/makers/background/tests/test_phase.py
+++ b/gammapy/makers/background/tests/test_phase.py
@@ -4,25 +4,33 @@ import numpy as np
 from numpy.testing import assert_allclose
 from astropy import units as u
 from astropy.coordinates import Angle, SkyCoord
-from regions import PointSkyRegion
+from regions import CircleSkyRegion, PointSkyRegion
 from gammapy.data import DataStore, EventList
 from gammapy.datasets import MapDataset, SpectrumDataset
 from gammapy.makers import MapDatasetMaker, PhaseBackgroundMaker, SpectrumDatasetMaker
 from gammapy.maps import MapAxis, RegionGeom, WcsGeom
-from gammapy.utils.regions import SphericalCircleSkyRegion
 from gammapy.utils.testing import requires_data
+
+POS_CRAB = SkyCoord(83.6331, +22.0145, frame="icrs", unit="deg")
+POS_VELA = SkyCoord("08h35m20.65525s", "-45d10m35.1545s", frame="icrs")
 
 
 @pytest.fixture(scope="session")
-def observations():
-    """Example observation list for testing."""
+def observations_cta():
     datastore_cta = DataStore.from_dir("$GAMMAPY_DATA/cta-1dc/index/gps")
+    return datastore_cta.get_observations([111630])
+
+
+@pytest.fixture(scope="session")
+def observations_magic():
     datastore_magic = DataStore.from_dir("$GAMMAPY_DATA/magic/rad_max/data")
+    return datastore_magic.get_observations(required_irf="point-like")[0]
+
+
+@pytest.fixture(scope="session")
+def observations_hess():
     datastore_hess = DataStore.from_dir("$GAMMAPY_DATA/hess-dl3-dr1/")
-    observations = datastore_cta.get_observations([111630])
-    observations.append(datastore_magic.get_observations(required_irf="point-like")[0])
-    observations.append(datastore_hess.obs(23523))
-    return observations
+    return datastore_hess.obs(23523)
 
 
 @pytest.fixture(scope="session")
@@ -37,21 +45,20 @@ def test_basic(phase_bkg_maker):
 
 
 @requires_data()
-def test_run_spectrum(observations, phase_bkg_maker):
+def test_run_spectrum(observations_cta, phase_bkg_maker):
 
     maker = SpectrumDatasetMaker()
 
     e_reco = MapAxis.from_edges(np.logspace(0, 2, 5) * u.TeV, name="energy")
     e_true = MapAxis.from_edges(np.logspace(-0.5, 2, 11) * u.TeV, name="energy_true")
 
-    pos = SkyCoord("08h35m20.65525s", "-45d10m35.1545s", frame="icrs")
     radius = Angle(0.2, "deg")
-    region = SphericalCircleSkyRegion(pos, radius)
+    region = CircleSkyRegion(POS_VELA, radius)
 
     geom = RegionGeom.create(region=region, axes=[e_reco])
     dataset_empty = SpectrumDataset.create(geom=geom, energy_axis_true=e_true)
 
-    obs = observations["111630"]
+    obs = observations_cta[0]
     dataset = maker.run(dataset_empty, obs)
     dataset_on_off = phase_bkg_maker.run(dataset, obs)
 
@@ -63,20 +70,18 @@ def test_run_spectrum(observations, phase_bkg_maker):
 
 
 @requires_data()
-def test_run_map(observations, phase_bkg_maker):
+def test_run_map(observations_cta, phase_bkg_maker):
 
     maker = MapDatasetMaker()
 
     e_reco = MapAxis.from_edges(np.logspace(0, 2, 5) * u.TeV, name="energy")
     e_true = MapAxis.from_edges(np.logspace(-0.5, 2, 11) * u.TeV, name="energy_true")
 
-    pos = SkyCoord("08h35m20.65525s", "-45d10m35.1545s", frame="icrs")
-
     binsz = Angle(0.02, "deg")
-    geom = WcsGeom.create(binsz=binsz, skydir=pos, width="2 deg", axes=[e_reco])
+    geom = WcsGeom.create(binsz=binsz, skydir=POS_VELA, width="2 deg", axes=[e_reco])
     dataset_empty = MapDataset.create(geom=geom, energy_axis_true=e_true)
 
-    obs = observations["111630"]
+    obs = observations_cta[0]
     dataset = maker.run(dataset_empty, obs)
     dataset_on_off = phase_bkg_maker.run(dataset, obs)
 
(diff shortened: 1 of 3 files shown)


registerrier (commented):
Thanks @MRegeard . Some comments:

- This case needs a proper test. 
- The piece of code added here is nearly the same as `MapDatasetMaker.make_counts` but the latter takes an `Observation` rather than an `EventList`.
- This shows that the current logic is probably not the good one and that phase handling should probably be performed before the dataset maker is run. But that's another story.

adonath:
Thanks @MRegeard, please add a test for this case.

MRegeard:
Here is a test for this !

However, it seems to me that there is no test function for the `make_counts_rad_max` function of `gammapy.makers.utils`. Am I right about this? In such case I could make one !

gammapy/makers/background/tests/test_phase.py
```diff
@@ -81,6 +85,143 @@ def test_run_map(observations, phase_bkg_maker):
...
+    dataset = maker.run(dataset_empty, obs)
+    dataset_on_off = phase_bkg_maker.run(dataset, obs)
+
+    counts_on = np.reshape(
+        np.array([30, 16, 14, 5, 10, 4, 0, 0, 0, 1, 1, 0, 0, 0, 0, 0, 0]), (17, 1, 1)
+    )
+    counts_off = np.reshape(
+        np.array(
+            [
+                73,
```
AtreyeeS, line 122: Maybe the test can be performed with lesser bins?

AtreyeeS (commented):
Thanks @MRegeard ! I suppose the tests can be condensed using `pytest.mark.parametrize`.

MRegeard:
Hi @AtreyeeS, I've done the tests with `pytest.mark.parametrize`.

gammapy/makers/background/tests/test_phase.py
```diff
@@ -95,3 +99,127 @@ def test_check_phase_intervals(pars):
...
+    maker = SpectrumDatasetMaker(
+        containment_correction=False, selection=["counts", "exposure", "edisp"]
+    )
+
+    e_reco = MapAxis.from_energy_bounds(0.05, 100, nbin=6, unit="TeV", name="energy")
+    e_true = MapAxis.from_energy_bounds(
+        0.01, 300, nbin=15, unit="TeV", name="energy_true"
+    )
+
+    if pars["p_in"][0] == "111630":
```
AtreyeeS, line 202: Maybe you can use the HESS DL3 Crab runs to get rid of this if else block

gammapy/makers/background/tests/test_phase.py
```diff
@@ -95,3 +99,127 @@ def test_check_phase_intervals(pars):
...
+        pos = SkyCoord(128.83606354, -45.17643181, frame="icrs", unit="deg")
+        obs = observations[pars["p_in"][0]]
+    elif pars["p_in"][0] == "5029747":
+        pos = SkyCoord(083.6331144560900, +22.0144871383400, frame="icrs", unit="deg")
+        obs = observations[pars["p_in"][0]]
+        table = obs.events.table
```
AtreyeeS, line 208: phase addition should be outside the if else block, no?

gammapy/makers/background/tests/test_phase.py
```diff
@@ -95,3 +99,127 @@ def test_check_phase_intervals(pars):
...
+        table["PHASE"] = np.linspace(0, 1, len(table["TIME"]))
+        obs._events = EventList(table)
+
+    if pars["p_in"][1] == "point-like":
+        on_region = PointSkyRegion(pos)
+    elif pars["p_in"][1] == "not-point-like":
+        on_region = SphericalCircleSkyRegion(pos, radius=0.1 * u.deg)
```
AtreyeeS, line 215: I imagine you can have the region directly in `p_in` instead of the `point-like` string

MRegeard: Let me test this !

gammapy/makers/background/tests/test_phase.py
MRegeard: I think I can't because I need a rad_max to test the feature and this is not in the HESS DL3 gammapy data. 
Maybe I'm mistaken on that

gammapy/makers/background/tests/test_phase.py
MRegeard: I don't think I can do better that this for the geom, but tell me !

gammapy/makers/background/tests/test_phase.py
AtreyeeS: Full enclosure IRFs should not have a `rad_max`, no? This tests the "standard analysis" case, which should be equivalent for the hess and cta datasets I think

MRegeard: But this also test that the rad_max is taken into account in the making of counts, so we need a run with rad_max to test that, hence the MAGIC run right?

bkhelifi (approved):
Sounds good for me! Thanks @MRegeard

AtreyeeS (approved):
Thanks @MRegeard ! No further comments from my side

gammapy/makers/background/tests/test_phase.py
```diff
@@ -95,3 +101,62 @@ def test_check_phase_intervals(pars):
...
+        0.01, 300, nbin=15, unit="TeV", name="energy_true"
+    )
+
+    obs = observations[pars["p_in"][0]]
+    pos = SkyCoord(083.6331144560900, +22.0144871383400, frame="icrs", unit="deg")
+    table = obs.events.table
+    table["PHASE"] = np.linspace(0, 1, len(table["TIME"]))
+    obs._events = EventList(table)
+
+    on_region = SphericalCircleSkyRegion(pos, radius=0.1 * u.deg)
```
registerrier, line 150: it would be simpler to define the on region directly in the pytest parameters.

gammapy/makers/background/tests/test_phase.py
```diff
@@ -95,3 +101,62 @@ def test_check_phase_intervals(pars):
...
+    [
+        {
+            "p_in": ["5029747", PointSkyRegion],
+            "p_out": [
+                np.reshape(np.array([58, 19, 2, 2, 0, 0]), (6, 1, 1)),
+                np.reshape(np.array([163, 46, 15, 1, 0, 0]), (6, 1, 1)),
+            ],
+        },
+        {
+            "p_in": ["5029747", SphericalCircleSkyRegion],
```
registerrier, line 117: Please use `CircleSkyRegion` instead.

gammapy/makers/background/tests/test_phase.py
```diff
@@ -15,8 +16,13 @@
 @pytest.fixture(scope="session")
 def observations():
     """Example observation list for testing."""
-    datastore = DataStore.from_dir("$GAMMAPY_DATA/cta-1dc/index/gps")
-    return datastore.get_observations([111630])
+    datastore_cta = DataStore.from_dir("$GAMMAPY_DATA/cta-1dc/index/gps")
```
registerrier, line 19: Here I would rather make a single pytest fixture for each type of observations

gammapy/makers/background/tests/test_phase.py
```diff
@@ -95,3 +101,62 @@ def test_check_phase_intervals(pars):
     assert_allclose(
         PhaseBackgroundMaker._check_intervals(pars["p_in"]), pars["p_out"], rtol=1e-5
     )
+
+
+@pytest.mark.parametrize(
+    "pars",
+    [
+        {
+            "p_in": ["5029747", PointSkyRegion],
```
registerrier, line 110: Maybe pass directly the observation rather than the obs id

gammapy/makers/tests/test_utils.py
```diff
@@ -332,6 +350,44 @@ def test_make_edisp_kernel_map():
     assert_allclose(kernel.pdf_matrix[:, 2], (0.0, 0.0, 0.0, 0.0, 1.0, 1.0), atol=1e-14)
 
 
+@requires_data()
+def test_make_counts_rad_max(observations):
```
registerrier, line 354: But this is already tested no?

gammapy/makers/tests/test_utils.py
```diff
@@ -332,6 +350,44 @@ def test_make_edisp_kernel_map():
...
+
+    assert_allclose(
+        counts.data, np.reshape(np.array([547, 188, 52, 8, 0, 0]), (6, 1, 1))
+    )
+
+
+@requires_data()
+def test_make_counts_off_rad_max(observations):
+
+    pos = SkyCoord(083.6331144560900, +22.0144871383400, unit="deg", frame="icrs")
```
registerrier, line 372:
```suggestion
    pos = SkyCoord(83.6331, 22.0145, unit="deg", frame="icrs")
```

gammapy/makers/tests/test_utils.py
```diff
@@ -332,6 +350,44 @@ def test_make_edisp_kernel_map():
...
+    region_finder = WobbleRegionsFinder(n_off_regions=3)
+    region_off, wcs = region_finder.run(on_region, pos)
+    geom_off = RegionGeom.from_regions(regions=region_off, axes=[energy_axis], wcs=wcs)
+
+    counts_off = make_counts_off_rad_max(
+        geom_off=geom_off, rad_max=observations.rad_max, events=observations.events
+    )
+
+    assert_allclose(
+        counts_off.data, np.reshape(np.array([1641, 564, 156, 24, 0, 0]), (6, 1, 1))
```
registerrier, line 387:
```suggestion
        np.squeeze(counts_off.data), [1641, 564, 156, 24, 0, 0])
```

registerrier (approved):
Thanks @MRegeard . See some minor inline comments.

gammapy/makers/tests/test_utils.py
MRegeard: If it is already tested I didn't find the test.

MRegeard:
@registerrier changes done !

PR #4359, opened 2023-03-06
Fix interpolation values_scale in TemplateSpatialModel

QRemy:
This PR allows to change the `values_scale` in the interpolation of the `TemplateSpatialModel`. The default interp_kwargs is also set to `log` instead of `lin` to be consistent with  the `TemplateSpectralModel`.

The fix is done in the evaluate method of the `TemplateSpatialModel` to avoid API breaking changes but ideally  all `Map.interp_by_coord` methods  should accept the `values_scale` argument instead.

### DIFF
--- a/gammapy/maps/wcs/ndmap.py
+++ b/gammapy/maps/wcs/ndmap.py
@@ -126,15 +126,43 @@ class WcsNDMap(WcsMap):
         idx = pix_tuple_to_idx(idx)
         return self.data.T[idx]
 
-    def interp_by_coord(self, coords, method="linear", fill_value=None):
+    def interp_by_coord(
+        self, coords, method="linear", fill_value=None, values_scale="lin"
+    ):
+        """Interpolate map values at the given map coordinates.
+
+        Parameters
+        ----------
+        coords : tuple, dict or `~gammapy.maps.MapCoord`
+            Coordinate arrays for each dimension of the map.  Tuple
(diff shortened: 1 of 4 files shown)


adonath (commented):
Thanks @QRemy, maybe we should introduce the `default_iterp_kwargs` directly to the `Map` object. E.g. as we have for IRFs too? Like: https://github.com/gammapy/gammapy/blob/main/gammapy/irf/background.py#L20

This way we would avoid the code duplication of handling the value scale in `TemplateSpatialModel`

QRemy:
Simplified by adding `fill_value` as an argument for `WcsNDMap.interp_by_coord`.

registerrier (approved):
Thanks @QRemy . This looks good and was indeed missing.

Is it a bugfix really or is this adding a new feature? Can this change the behaviour of e.g. existing serialized models?

QRemy:
> Is it a bugfix really or is this adding a new feature? Can this change the behaviour of e.g. existing serialized models?

Discussing with @adonath we considered that it can be a bug fix as the previous behaviour was not the one intended (especially as regiondmap supports the value_scale but not wcsndmap)
Existing serialized models can still be opened but the values_scale of TemplateSpatialModel will be in 'log' by default instead of 'lin'.

registerrier:
> @registerrier I think the motivation for the bug fix here is to have e.g. log-log interpolation for the fermi-diffuse model in energy.

Sounds good. Should I proceed and merge the PR?

PR #4369, opened 2023-03-10
Pulsar analysis tutorial

MRegeard:
This is a pull request for the update of the pulsar analysis tutorial. 

Note that I have added a notebook to the tutorial directory so that it is easier to see the changes and run the tutorial, but it will be removed once the tutorial is transformed into a python script and ready to be merged. 

This is a 1st draft, changes still have to be made.


adonath:
Thanks a lot @MRegeard! On a side note: I think changes will be much easier to see if you commit the python script instead. The diff on GitHub will be much more meaningful, than the json file. Reviewer can always checkout you branch and open the Python script as notebook using e.g. `jupytext`. My recommendation here would be to commit the `.py` file instead.

MRegeard:
> Thanks a lot @MRegeard! On a side note: I think changes will be much easier to see if you commit the python script instead. The diff on GitHub will be much more meaningful, than the json file. Reviewer can always checkout you branch and open the Python script as notebook using e.g. `jupytext`. My recommendation here would be to commit the `.py` file instead.

I was think of doing the big changes on the notebook and then when it is in an end phase I transform and commit the script and make the small modification from the reviewer. But I will try and see if it easier doing it directly with the script. 
Moreover, since some other people might also contribute in to this branch I thought it might be easier with the notebook way. But I might be wrong.

adonath:
> Moreover, since some other people might also contribute in to this branch I thought it might be easier with the notebook way. But I might be wrong.

With the `.ipynb` format you will run very likely into merge conflicts if others contribute. Just executing the notebook (without any modifications) will change the meta data of the `.ipynb` file. Merge conflicts in `.ipynb` files are very hard to resolve. 

> But I will try and see if it easier doing it directly with the script.

Once you get used to the format, small (up to mid sized) changes are much easier to do in the script.

In case you still prefer editing with a notebook server, you can use: https://jupytext.readthedocs.io/en/latest/index.html I guess this should be better documented in our developers docs.

AtreyeeS:
Thanks @MRegeard ! A few comments:

1. Can you suppress the warnings? Makes the notebook unreadable
2. Why do you use `edisp-bias` in the safe mask maker?
3. Why do you use a correlation radius of 0.15 deg
4. I think there is no need to show so many maps - only the excess and the sort_ts should be enough
5. Maybe it is good to explicitly mention that the only difference is in the makers stage where the `PhaseBackgroundMaker` has to be used. Once the `Dataset` is created, modelling and fitting can proceed exactly as for the standard cases...

AtreyeeS (dismissed):
Looks good to me! Thanks @MRegeard

examples/tutorials/analysis-time/pulsar_analysis.py
```diff
@@ -198,99 +209,176 @@
...
+
+spectrum_datasets = Datasets()
 
 for obs in obs_list_vela:
-    dataset = dataset_maker.run(dataset_empty, obs)
-    dataset_on_off = phase_bkg_maker.run(dataset, obs)
-    dataset_on_off = safe_mask_maker.run(dataset_on_off, obs)
-    datasets.append(dataset_on_off)
+    spectrum_dataset = spectrum_dataset_maker.run(spectrum_dataset_empty, obs)
+    spectrum_dataset_on_off = phase_bkg_maker.run(spectrum_dataset, obs)
```
registerrier, line 366: Same question here. Why is the safe run after the bkg? It was like this before, was there a specific reason?

registerrier (approved):
Thanks @MRegeard. Looks good. See inline comments.

examples/tutorials/analysis-time/pulsar_analysis.py
MRegeard: Yes it was like that before. I changed it the other way around.

AtreyeeS: I imagine the order is not critical here? It is important in case of, ex: the `bkg-peak`

registerrier (approved):
Thanks @MRegeard . No more comments from my side.?

PR #4295, opened 2023-01-27
Implement the _sample_coord_time_energy function in MapDatasetEventSampler

fabiopintore:
This re-open the closed PR #4100.

PR #4370, opened 2023-03-10
Implement the _sample_coord_time_energy function in MapDatasetEventSampler

fabiopintore:
This PR replaces the #4295

### DIFF
--- a/gammapy/datasets/simulate.py
+++ b/gammapy/datasets/simulate.py
@@ -174,9 +174,6 @@ class MapDatasetEventSampler:
                     dataset.mask,
                 )
 
-            flux = evaluator.compute_flux()
-            npred = evaluator.apply_exposure(flux)
-
             if evaluator.model.temporal_model is None:
                 temporal_model = ConstantTemporalModel()
             else:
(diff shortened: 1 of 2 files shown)


gammapy/datasets/simulate.py
```diff
@@ -85,7 +182,12 @@ def sample_sources(self, dataset):
             else:
                 temporal_model = evaluator.model.temporal_model
 
-            table = self._sample_coord_time(npred, temporal_model, dataset.gti)
+            if (
+                temporal_model == "TemplateTemporalModel"
+            ):  # check the condition when the format model is defined
+                table = self._sample_coord_time_energy(dataset, evaluator)
+            else:
+                table = self._sample_coord_time(npred, temporal_model, dataset.gti)
```
registerrier, line 190: Maybe the line computing the `flux` and the `npred` should be moved in this `else` clause to avoid computing `npred` for a time varying spectral model.

registerrier (dismissed):
Thanks @fabiopintore . This looks good. I have left a minor suggestion inline.

gammapy/datasets/simulate.py
fabiopintore: Thank you @registerrier . I've implemented your suggestion.

registerrier (approved):
Thanks @fabiopintore . This looks good.

In the next step the method `_evaluate_timevar_source` will be reimplemented and tested, right?

fabiopintore:
Thanks @registerrier . 
Yes, the next step will be to (re-)implement the function `_evaluate_timevar_source` into `_sample_coord_time_energy`.

PR #4380, opened 2023-03-14
Adapt default offset for plotting point like IRFs

AtreyeeS:
The plot for `edisp.peek()` looks bad for point like IRFs (eg: https://github.com/gammapy/gammapy-recipes/blob/2ee2588268dccc266cebafca32979f410068e68a/recipes/pulsar_phase/pulsar_phase_computation.ipynb)

This is because of the choice of the default angle. This PR corrects it. The plot now looks like

Screenshot 2023-03-14 at 19 03 32

@registerrier : I dont know if we can't still put it in 1.0.1.

### DIFF
--- a/gammapy/irf/edisp/core.py
+++ b/gammapy/irf/edisp/core.py
@@ -52,6 +52,14 @@ class EnergyDispersion2D(IRF):
     required_axes = ["energy_true", "migra", "offset"]
     default_unit = u.one
 
+    @property
+    def _default_offset(self):
+        if self.axes["offset"].nbin == 1:
+            default_offset = self.axes["offset"].center
+        else:
+            default_offset = [1.0] * u.deg
+        return default_offset
+
     def _mask_out_bounds(self, invalid):
         return (
             invalid[self.axes.index("energy_true")] & invalid[self.axes.index("migra")]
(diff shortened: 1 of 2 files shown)


AtreyeeS:
> Wouldn't it be possible to plot the effective area as a regular line plot (1D) rather than a 2D plot in this case?

That is done in `aeff.peek()`. For me, `peeks` are quick functions which show that data exists properly etc, so not too many `if` statements... Better if they plot as similar things as possible. I removed the offset dependence plot because having empty plots seems to imply something is not fully correctly...

A 2D plot here quickly shows that there is no offset dependence... I can change it if you prefer...

gammapy/irf/edisp/core.py
```diff
@@ -52,6 +52,14 @@ class EnergyDispersion2D(IRF):
     required_axes = ["energy_true", "migra", "offset"]
     default_unit = u.one
 
+    @property
+    def _default_offset(self):
```
registerrier, line 56: Note that this might have to evolve once we introduce more complex IRF with FoV spatial dependency

registerrier (approved):
Thanks @AtreyeeS

PR #4381, opened 2023-03-15
PIG 16 - Model Priors API

nbiederbeck:
This pull request adds a rough PIG for `Prior` (and `FitStatistic`).  I anticipate I will add details with your assistance.


docs/development/pigs/pig-026.rst
```diff
@@ -0,0 +1,91 @@
...
+
+Abstract
+========
+
+This PIG is intended to introduce priors on models that are evaluated during fitting.
+
+Motivation
+==========
+
+Using priors on models and on parameters is the next step in full-fledged statistical analyses.
```
bkhelifi, line 23:
```suggestion
Using priors on models or on parameters is the next step in full-fledged statistical analyses.
```

docs/development/pigs/pig-026.rst
```diff
@@ -0,0 +1,91 @@
...
+A prior can incorporate the analysers' knowledge of the topic at hand, and yield more realistically and trustworthy results.
```
bkhelifi, line 24: Maybe not only the analyser... Indeed, one idea to deal with the IRFs systematics was to use the priors mechanics. So, if it is implemented like that, the knowledge comes from the people making the IRFs...

By the use, do we add as use case the inclusion of IRFs systematics into priors?

adonath:
Thanks a lot @nbiederbeck for this minimal draft. The overall structure is good and the content should be extended for details now:

- We need the definition of an abstract Prior base class.
- Clarify whether priors can have fittable parameters (yes!)
- Some priors do not have fitable parameters... 
- Clarify serialisation of priors with the model e.g. via an example YAML file
- Mention prior registry system
- Add a list of proposed standard priors: `GaussianPrior`, `UniformPrior`, `MultivariateGaussianPrior`
- Strictly speaking one would not need the  `L2Regularization` as it presents a multivariate Gaussian prior.
- Clarify which Gammapy classes will support a `.prior` attribute: `Models`, `SpectralModel`, `SkyModel` etc.
- Clarify how (hierarchical) priors are combined in the fit-statistic function
- Do we need e.g. a class like `PriorFitStatistic(beta=1.)`?

These are just some thoughts for now, I'm sure I'll have some more...

docs/development/pigs/pig-026.rst
```diff
@@ -0,0 +1,91 @@
...
+            return ((value - self.mu) / self.sigma) ** 2
+
+    dataset.models.parameters["x"].prior = GaussianPrior()
+
+
+Priors on models:
+
+.. code:: python
+
+    class L2Regularization():
```
adonath, line 62: Note that this case actually corresponds to the `GaussianPrior` on each of the individual parameters, where `mu=0 and sigma=1`. The most general case I could imagine would be `MultiVariateGaussianPrior` on a list of parameters. The case of `L2` regularization would be included there...but maybe we just keep it for convenience.

adonath:
@katrinstreil kindly agreed to take over the writing and further development of the PIG. Please let us of know if you have any questions.

registerrier (approved):
Thanks @katrinstreil  @nbiederbeck . This already quite nice.

I would like to clarify a number of potential issues:
1. it seems a bit complicated to me that priors can live and be defined and evaluated at different levels in the modeling package structure. I don't think that the `SkyModel` is the correct place anyway. We could have something to set priors on the `Models` instance with e.g. `models.set_prior(parameters, prior)`. 
2. Parameters have units. How do we make sure that units are correctly handled?
3. Is the `PriorFitStatistic` really needed? As far as I can see, it is mostly a single function. 
4. Connected to this, I wonder if the mixing of two use cases in a single prior framework is not a bit confusing. We have:
  - priors on individual parameters as usually done in e.g. sherpa, bxa, 3ML. 
  - penalization terms for the ML which is used for  e.g. regularization in unfolding

docs/development/pigs/pig-026.rst
```diff
@@ -0,0 +1,328 @@
...
+            from numpy import linalg
+            return values @ linalg.inv(self.cov) @ values
+        
+        @property
+        def prior_parameters(self):
+            return [("cov", self.cov)]    
+            
+The prior classes are supposed to be set as ``.prior`` on:
+
+* Model-like classes, for instance the ``SkyModel`` class:
```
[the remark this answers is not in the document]
adonath: The reason why it should be attached to the `Model` is bookkeeping. Priors should be serialized with the model. Parameters do not know, to which model they belong. The alternative is to handle it completely independent as we do for the covariance matrix and we would rely on the order of parameters. But in this case it would still be attached to the `Model`.

docs/development/pigs/pig-026.rst
```diff
@@ -0,0 +1,328 @@
...
+          """Total statistic given the current model parameters."""
+          stat = self.stat_array()
+          
+          if self.mask is not None:
+              stat = stat[self.mask.data]
+              
+          return np.sum(stat, dtype=np.float64) + PriorFitStatistic(weight=1).stat_sum(models = self.models)
+    
+    
+Where the ``PriorFitStatistic`` iterates the models and parameters, evaluates and combines the priors with the possibility to assign a weight.
```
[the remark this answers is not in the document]
adonath: The idea of the additional class is just to make it explicit, that one changes to a MAP estimate. It's like a global switch to account for the prior or not. The alternative is indeed to always evaluate priors, if present...

docs/development/pigs/pig-026.rst
```diff
@@ -0,0 +1,328 @@
...
+
+.. code:: python 
+
+ class PriorFitStatistic():
+       
+      def __init__(self, weight=1):
+            self.weight = weight
```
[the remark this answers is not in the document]
adonath: This is the most common one, where people introduce a weight parameter (equivalent to a regularization strength...).

docs/development/pigs/pig-026.rst
```diff
@@ -0,0 +1,328 @@
...
+=========
+
+In the past priors were a regularly requested feature or the solution to a problem.
+See the following issues and PRs of Gammapy:
+
+- Add sigma v estimator functionality for dark matter use case `#2075`_
+- Background systematics as a nuisance parameter `#3955`_
+- Support unfolding methods for spectral flux points `#4122`_
+- Add PiecewiseNormSpatialModel `#4208`_
+
```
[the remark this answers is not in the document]
adonath: This is also my main remaining comment. Adding the uses cases in code. E.g. for the unfolding it could look like:

```python3
import numpy as np
from gammapy.modeling.model import PiecewiseNormSpectralModel, MultivariateGaussianPrior, PowerLawSpectralModel

n_points = 10

pwl = PowerLawSpectraModel()

energy = np.geomspace(1, 10, n_points) * u.TeV
norm =  PiecewiseNormSpectralModel(energy=energy)

prior = MultivariateGaussianPrior.from_(means=np.ones(10), covariance_type="diagonal")
norm.prior = prior

spectral_model = pwl * norm
```

adonath:
> Connected to this, I wonder if the mixing of two use cases in a single prior framework is not a bit confusing. We have:
priors on individual parameters as usually done in e.g. sherpa, bxa, 3ML.
penalization terms for the ML which are used for e.g. regularization in unfolding
Maybe this should be distinct features?

I don't think they are not distinct features. If you do the math you will find the L2 regularization corresponds to a multivariate Gaussian prior in a Maximum A-Posteriori estimation. The information is all in the covariance matrix. I think the right approach here is to just introduce a `MultivariateGaussianPrior` and some convenience methods for the creation of covariance matrices. See my sample above.

docs/development/pigs/pig-026.rst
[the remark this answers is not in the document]
katrinstreil: Thanks @registerrier and @adonath! I will move the use cases up and also add a similar example as you showed here!

docs/development/pigs/pig-026.rst
```diff
@@ -0,0 +1,287 @@
...
+        """Uniform Prior
+        """
+        tag = ["UniformPrior"]
+        uni = PriorParameter(name="uni", value = 0, min = 0, max = 10, unit = '' )
+        
+        @staticmethod
+        def evaluate(value, uni):
+            return uni
+
+
```
facero, line 173: A prior that can be quite useful is the LogUniform prior.
For example for the normalization of a source, giving a linear prior between 1e-14 and 1e-11 is not ideal. Having the LogUniform priors will become handy if at some point we perform Bayesian analyses.
It could be something like this : 

```
class LogUniformPrior(PriorModel):
         """LogUniform Prior
         """
         tag = ["LogUniformPrior"]
         uni = PriorParameter(name="loguni", value = 1, min = 0.1, max = 10, unit = '' )
```

A note on the fact that in the Uniform and LogUniform priors the `value` parameter itself is a not useful as we just want to define (min, max) boundaries.
Is there a reason we should have a `value` parameter in this case ?

katrinstreil: Thanks for the comment! 
Yes, I think this kind of prior could be helpful and is very straightforward to implement. 
The reason for the ``value`` parameter was to gain more flexibility in the use cases. In some cases, one might the prior to be stronger than in others.

docs/development/pigs/pig-026.rst
```diff
@@ -0,0 +1,328 @@
...
+--------------
+
+.. code:: python 
+
+  class Prior():
+  
+     def __call__(self, base):
+     
+        if isinstance(base, Parameter):
+            return self.evaluate(base.value)
```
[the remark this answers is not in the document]
adonath: I guess priors should be defined in such a way, that they are unit less (i.e. a likelihood after evaluation). But maybe we should add a check for this...

docs/development/pigs/pig-026.rst
```diff
@@ -0,0 +1,328 @@
...
+  class GaussianPrior(Prior):
+  
+      """Gaussian Prior with mu and sigma set on a single parameter
+      """
+      
+      def __init__(self, mu=0, sigma=1):
+          self.mu = mu
+          self.sigma = sigma
+
+      def evaluate(self, value):
```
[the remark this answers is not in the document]
adonath: Yes, I agree that the `.evaluate()` method should be static if possible. This will allows us to easier move to other tensor libraries later...The stateful call can happen in `Prior.__call__`

docs/development/pigs/pig-026.rst
```diff
@@ -0,0 +1,328 @@
...
+     
+        if isinstance(base, Parameter):
+            return self.evaluate(base.value)      
+        else: # model-like instance
+            return self.evaluate(base.parameters.value)  
+         
+
+Future additional properties of the ``Prior`` classes: 
+
+- Fittable parameters
```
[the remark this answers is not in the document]
adonath: Yes, I think in general a prior should be treated much like a model. In general prior parameter can be fitted in a likelihood fit (maximum a posteriori) as well...so I think they should handle `Parameter` objects.

docs/development/pigs/pig-026.rst
```diff
@@ -0,0 +1,287 @@
...
+
+   pwl = PowerLawSpectraModel()
+   prior = StepPrior(xmin = "-inf", xmax = "0",  value = "1")
+   pwl.set_prior([pwl.parameters['amplitude']], [prior])
+  
+
+
+
+Case 3: Support unfolding methods for spectral flux points `#4122`_
+--------------
```
registerrier, line 71:
```suggestion
Case 3: Support unfolding methods for spectral flux points `#4122`_
----------------------------------------------------------
```

docs/development/pigs/pig-026.rst
registerrier: Does this prior need a range for the parameter , e.g.:
```
 min = PriorParameter(name="min", value = -np.inf, unit = '' )
 max = PriorParameter(name="min", value = np.inf, unit = '' )
```

docs/development/pigs/pig-026.rst
```diff
@@ -0,0 +1,287 @@
...
+          
+          def prior_stat_sum(self):
+              parameters_stat_sum = 0
+              for par in self:
+                  if par.prior is not None:
+                      parameters_stat_sum += par.prior_stat_sum()
+              return parameters_stat_sum
+
+
+During the Fit the ``stat_sum`` of the Datasets is evaluated where now the statistics due to the priors set on the parameters is added.
```
registerrier, line 212:
```suggestion
During the Fit the ``Datasets.stat_sum()`` function is evaluated.  Now the function has to compute the statistics due to the priors set on the parameters and add to the stat_sum. 
```

docs/development/pigs/pig-026.rst
```diff
@@ -0,0 +1,287 @@
...
+   'interp': 'lin',
+   'scale_method': 'scale10',
+   'is_norm': False,
+   'prior': {'tag': 'GaussianPrior',
+            'parameters': [{'name': 'mu', 'value': 0},
+                           {'name': 'sigma', 'value': 0.1}]}}
+
+
+Implementation Outline
+--------------
```
registerrier, line 265:
```suggestion
Implementation Outline
-----------------------
```

docs/development/pigs/pig-026.rst
```diff
@@ -0,0 +1,287 @@
...
+2. ``PriorParameter`` and ``PriorParameters`` (See `#4620`_)
+3. ``.prior`` attribute in the ``Parameter``, ``Parameters`` and ``Model`` classes
+4.  Evaluation of the prior in the ``Datasets`` class
+
+
+
+Decision
+========
+
+
```
registerrier, line 279: The PIG was discussed and reviewed among Gammapy developers. It has been reviewed by the CC and is now considered accepted.

registerrier (dismissed):
Thanks @katrinstreil . 

Let's now send it to the CC.

bkhelifi (approved):
Thank you, @katrinstreil and @nbiederbeck

docs/development/pigs/pig-026.rst
```diff
@@ -0,0 +1,287 @@
...
```
registerrier, line 279:
```suggestion
The PIG was discussed and reviewed among Gammapy developers. It has been reviewed by the CC and is now considered accepted.
```

docs/development/pigs/pig-026.rst
```diff
@@ -0,0 +1,287 @@
...
+Future additional properties of the ``PriorModel`` classes: 
+
+- Write/read from/to a yaml file, ideally also when the corresponding model is written/read (see serialisation example below)
+- Prior registry system
+- Different prior subclasses depending on the use cases 
+
+Exemplary additional prior subclasses:
+---------------------------------------
+.. code:: python
```
registerrier, line 144:
```suggestion
Exemplary additional prior subclasses:
--------------------------------------

.. code:: python
```

docs/development/pigs/pig-026.rst
```diff
@@ -0,0 +1,287 @@
...
+
+.. _pig-026:
+
+********************************
+PIG 26 - Model Priors API 
+********************************
+
+* Author: Noah Biederbeck, Katrin Streil
+* Created: ``2023-03-13``
+* Accepted:
```
registerrier, line 11:
```suggestion
* Accepted: ``2023-11-03``
```

docs/development/pigs/pig-026.rst
```diff
@@ -0,0 +1,287 @@
...
+* Status: draft
```
registerrier, line 12:
```suggestion
* Status: Accepted
```

registerrier:
Review period is over. The PIG is accepted. Merging now.

PR #4382, opened 2023-03-15
Fix error message in interp_to_geom()

adonath:
**Description**

See title.

**Dear reviewer**

### DIFF
--- a/gammapy/maps/core.py
+++ b/gammapy/maps/core.py
@@ -998,7 +998,7 @@ class Map(abc.ABC):
         if preserve_counts:
             if geom.ndim > 2 and geom.axes[0] != self.geom.axes[0]:
                 raise ValueError(
-                    f"Energy axis do not match, expected: \n {self.geom.axes[0]},"
+                    f"Energy axes do not match, expected: \n {self.geom.axes[0]},"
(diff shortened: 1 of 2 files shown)


gammapy/maps/core.py
```diff
@@ -998,8 +998,8 @@ def interp_to_geom(self, geom, preserve_counts=False, fill_value=0, **kwargs):
         if preserve_counts:
             if geom.ndim > 2 and geom.axes[0] != self.geom.axes[0]:
                 raise ValueError(
-                    f"Energy axis do not match: expected {self.geom.axes[0]},"
-                    " but got {geom.axes[0]}."
+                    f"Energy axis do not match, expected: \n {self.geom.axes[0]},"
```
bkhelifi, line 1001: maybe "Energy axis does not match" or "Energy axes do not match"...

bkhelifi (approved):
Thanks @adonath. I have just propose a minor suggestion

gammapy/maps/core.py
adonath, line 1001:
```suggestion
                    f"Energy axes do not match, expected: \n {self.geom.axes[0]},"
```

PR #4384, opened 2023-03-17
Correct default in FluxPointsEstimator

AtreyeeS:
The doc string was wrong


adonath (commented):
Thanks @AtreyeeS, no comments. Can you maybe check the other estimators for consistency?

AtreyeeS:
@adonath Good point! From a quick look, I see that the doc strings have some missing/incorrect info indeed. I can do it systematically for the estimators next week. Maybe we merge this one now?

bkhelifi (approved):
Well seen! Thanks @AtreyeeS

Issue #4383, opened 2023-03-15
Minuit optimization message sometimes misleading


cbadams2:
**Gammapy version**
1.0

**Bug description**
In the gammapy code for iminuit:
https://github.com/gammapy/gammapy/blob/3bcbfce287f02e03eabfe6413cdf67f0fa07c734/gammapy/modeling/iminuit.py#L178-L190
a user is informed that their optimization has failed if their minuit optimization returns `False` for `minuit.accurate`. However, it is a default behavior for the `minuit.accurate` to return `False` if the fit strategy was set to `0`.

This case will often lead to the somewhat confusing output of:
```
OptimizeResult

	backend    : minuit
	method     : migrad
	success    : True
	message    : Optimization failed.
```

**Expected behavior**
It appears that it would be a simple fix to duplicate what iminuit is currently doing:
https://github.com/scikit-hep/iminuit/blob/v2.21.0/src/iminuit/minimize.py#L124-L136
which will clarify that the optimization has succeeded, but that the uncertainties are not trustworthy.

registerrier:
Solved with #4393

PR #4393, opened 2023-03-21
Iminuit output

bkhelifi:
**Description**
This pull request is associated to the Issue #4383 permitting to fix the iminuit message.

### DIFF
--- a/gammapy/modeling/iminuit.py
+++ b/gammapy/modeling/iminuit.py
@@ -178,24 +178,21 @@ def contour_iminuit(parameters, function, x, y, numpoints, sigma, **kwargs):
 # This code is a mix from https://github.com/iminuit/iminuit/blob/master/iminuit/_minimize.py#L95
 # and https://github.com/scikit-hep/iminuit/blob/v2.21.0/src/iminuit/minimize.py#L124-L136
 def _get_message(m, parameters):
-    success = m.accurate
+    success = m.valid
     success &= np.all(np.isfinite([par.value for par in parameters]))
-    if not success:
+    if success:
+        message = "Optimization terminated successfully."
+        if m.accurate:
+            message += "."
+        else:
+            message += ", but uncertainties are unreliable."
+    else:
         message = "Optimization failed."
         fmin = m.fmin
         if fmin.has_reached_call_limit:
             message += " Call limit was reached."
         if fmin.is_above_max_edm:
             message += " Estimated distance to minimum too large."
-    elif m.valid:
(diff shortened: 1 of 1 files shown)


cbadams2:
I don't know if this is a functional solution to the issue. The way I'm reading the new code, I believe that it will still yield the same output for the problem case of choosing a fit strategy of 0:

It will check for `m.accurate`, find it to be `False`, and enter the first conditional statement, causing it to skip over `elif m.valid`. 

The trouble I think comes from associating the variable `success` with `minuit.accurate`, which is somewhat misleading because it seems that `accurate` is related to the covariance rather than convergence of the fit.  The iminuit code associates success with `minuit.valid` instead. I would maybe suggest the following, which is more or less the same as in https://github.com/scikit-hep/iminuit/blob/v2.21.0/src/iminuit/minimize.py#L124-L136:
```
def _get_message(m, parameters): 
    success = m.valid
    success &= np.all(np.isfinite([par.value for par in parameters])) 
    if success:
        message = "Optimization terminated successfully."
        if m.accurate:
            message += "."
        else:
            message += ", but uncertainties are unrealiable."
    else: 
        message = "Optimization failed." 
        fmin = m.fmin 
        if fmin.has_reached_call_limit: 
            message += " Call limit was reached." 
        if fmin.is_above_max_edm: 
            message += " Estimated distance to minimum too large." 
    return message 
```

AtreyeeS (approved):
Thanks @bkhelifi 
It would be nice to have this tested to ensure all the cases are correctly covered, but it might be complex coming up with a simple case where the fit fails

gammapy/modeling/iminuit.py
```diff
@@ -175,12 +175,18 @@ def contour_iminuit(parameters, function, x, y, numpoints, sigma, **kwargs):
     }
 
 
-# this code is copied from https://github.com/iminuit/iminuit/blob/master/iminuit/_minimize.py#L95
+# This code is a mix from https://github.com/iminuit/iminuit/blob/master/iminuit/_minimize.py#L95
+# and https://github.com/scikit-hep/iminuit/blob/v2.21.0/src/iminuit/minimize.py#L124-L136
```
QRemy, line 179:
```suggestion
# This code is a copied from https://github.com/scikit-hep/iminuit/blob/v2.21.0/src/iminuit/minimize.py#L124-L136
```

The first link given as reference is invalid

registerrier (approved):
Thanks @bkhelifi . Looks good!

Issue #4387, opened 2023-03-17
Using ObservationFilter on the pulsar phase does not change the livetime of the observation


MRegeard:
**Gammapy version**
dev

**Bug description**
When using `ObservationFilter`on the pulsar phase, the `observation.observation_live_time_duration`is not modified to account for the phase selection. 

**Fix**
It can be easily fixed by checking if `observation.obs_filter`is not None and then if `obs_filter.event_filter`as 'PHASE', get the band in phase and multiply the return of `observation.observation_live_time_duration` by the range in phase. 

**Alongside discussion**
I think that in the end we would like to store the filter applied by the `ObservationFilter` somewhere in the `Observation`. Then the discussion is where to store these information. Is `Observation.table.meta` the right place ?

PR #4394, opened 2023-03-21
Obs filter live time

MRegeard:
This pull request is related to issue #4387. 

I added a staticmethod which check if the `ObservationFilter`applied is on the phase. If it's the case it returns the range of phase that has been filtered (i.e. phase range is [0.2, 0.8], it returns 0.6). If no phase filtering has been applied to `ObservationFilter`, it just return 1. 
And then this method is called when computing `observation_live_time_duration`.

I also added a test that check the observation livetime durationn, and especially check that the livetime is corrected when filtering on phase. 
I also added a test for the staticmethod.

### DIFF
--- a/gammapy/data/filters.py
+++ b/gammapy/data/filters.py
@@ -50,7 +50,7 @@ class ObservationFilter:
 
     @property
     def livetime_fraction(self):
-        """Fraction of the livetime resulting from event_filters."""
+        """Fraction of the livetime kept when applying the event_filters."""
         return self._check_filter_phase(self.event_filters)
 
     def filter_events(self, events):
--- a/gammapy/data/filters.py
+++ b/gammapy/data/filters.py
@@ -47,7 +47,11 @@ class ObservationFilter:
     def __init__(self, time_filter=None, event_filters=None):
         self.time_filter = time_filter
         self.event_filters = event_filters or []
-        self.phase_range = self._check_filter_phase(event_filters)
+
+    @property
+    def livetime_fraction(self):
+        """Fraction of the livetime resulting from event_filters."""
+        return self._check_filter_phase(self.event_filters)
 
     def filter_events(self, events):
         """Apply filters to an event list.
--- a/gammapy/data/filters.py
+++ b/gammapy/data/filters.py
@@ -47,6 +47,7 @@ class ObservationFilter:
     def __init__(self, time_filter=None, event_filters=None):
         self.time_filter = time_filter
         self.event_filters = event_filters or []
+        self.phase_range = self._check_filter_phase(event_filters)
 
     def filter_events(self, events):
         """Apply filters to an event list.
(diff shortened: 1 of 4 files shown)


gammapy/data/filters.py
```diff
@@ -97,3 +98,14 @@ def _filter_by_time(self, data):
     def copy(self):
         """Copy the `ObservationFilter` object."""
         return copy.deepcopy(self)
+
+    @staticmethod
```
registerrier, line 102: I think it would safer to make this a `property` and not define the value on init.

gammapy/data/filters.py
```diff
@@ -47,6 +47,7 @@ class that corresponds to the filter type
     def __init__(self, time_filter=None, event_filters=None):
         self.time_filter = time_filter
         self.event_filters = event_filters or []
+        self.phase_range = self._check_filter_phase(event_filters)
```
registerrier, line 50: Same comment this should be a `property`

registerrier (commented):
Thanks @MRegeard . This is better like this. It would be better to have the `ObservationFilter.phase_range` as a property.

Is the name `phase_range` clear? It is the livetime fraction resulting from applying `ObservationFilter`.

MRegeard:
Done 
I changed the name to `livetime_fraction`and made it a property.

registerrier (dismissed):
Thanks @MRegeard . No further comment.

@adonath , is the approach fine with you?

gammapy/data/filters.py
```diff
@@ -48,6 +48,11 @@ def __init__(self, time_filter=None, event_filters=None):
         self.time_filter = time_filter
         self.event_filters = event_filters or []
 
+    @property
+    def livetime_fraction(self):
```
AtreyeeS, line 52: Is this a good name? Off hand, it seems like a `livetime_duration/observation_duration`... like `deadtime_fraction`
Also, the doc string `resulting from event_filters` is a bit confusing.

gammapy/data/filters.py
```diff
@@ -97,3 +102,14 @@ def _filter_by_time(self, data):
     def copy(self):
         """Copy the `ObservationFilter` object."""
         return copy.deepcopy(self)
+
+    @staticmethod
+    def _check_filter_phase(event_filter):
```
AtreyeeS, line 107: Why is this function hidden?

AtreyeeS (approved):
Thanks @MRegeard ! I like the implementation, the only point is `livetime_fraction` sounds a bit like `1-deadtime_fraction`. I dont know if it would be better to have a different name

gammapy/data/filters.py
MRegeard: Well, this is really the livetime that you keep in your observation. But I am open to other suggestion !

gammapy/data/filters.py
MRegeard: I assume it had to be hidden because it is static and doesn't do anything interesting on it's own. 
But it could be public

Issue #4258, opened 2022-12-17
Light curve plot with MJD on the time axis


cosimoNigro:
It is standard to plot gamma-ray light curves with the time axis in MJD.

This is extremely useful for very large periods of time (e.g. multi-year monitoring), where the [current default plotting (see this tutorial)](https://docs.gammapy.org/1.0/tutorials/analysis-time/light_curve.html?highlight=light%20curves#light-curve-estimation-by-observation) with ISO time will be too detailed for such a wide time axis.

Could we add the option to have the time axis in MJD to the `FluxPoints.plot()` function?

Incidentally, when I was trying to plot myself manually a LC in MJD I run into #4257.

adonath:
The way to configure this currently is on the `TimeMapAxis` itself:

```
from gammapy.maps import TimeMapAxis

TimeMapAxis.time_format = "mad"
```

This is e.g. shown here: https://docs.gammapy.org/1.0/tutorials/analysis-time/light_curve_simulation.html#check-setup

The reason is that the plot formatting happens on the `TimeMapAxis` and because there are more places where the info is used. But I agree that its would be convenient to have it in `FluxPoints.plot(..., time_format=)` as well.

registerrier:
This has been implementd in #4395 . Closing now.

PR #4395, opened 2023-03-21
Add the possibility to plot in MJD the light curves

bkhelifi:
**Description**
This pull request is associated to the Issue #4258 . It adds an argument in the plot function of the FluxPoints, `time_format`. This argument is propagated to the `NDRegionMap` and the `TimeMapAxis` is correctly formatted for mjd.

### DIFF
--- a/gammapy/estimators/points/core.py
+++ b/gammapy/estimators/points/core.py
@@ -471,7 +471,7 @@ class FluxPoints(FluxMaps):
 
         return y_errn, y_errp
 
-    def plot(self, ax=None, sed_type=None, energy_power=0, **kwargs):
+    def plot(self, ax=None, sed_type=None, energy_power=0, time_format="iso", **kwargs):
         """Plot flux points.
 
         Parameters
@@ -482,6 +482,8 @@ class FluxPoints(FluxMaps):
             Sed type
         energy_power : float
             Power of energy to multiply flux axis with
+        time_format : {"iso", "mjd"}
+            Used time format is a time axis is present. Default: "iso"
         **kwargs : dict
             Keyword arguments passed to `~RegionNDMap.plot`
 
@@ -529,6 +531,8 @@ class FluxPoints(FluxMaps):
             kwargs.setdefault("yerr", None)
 
         flux = scale_plot_flux(flux=flux.to_unit(flux_unit), energy_power=energy_power)
+        if "time" in flux.geom.axes_names:
+            flux.geom.axes["time"].time_format = time_format
         ax = flux.plot(ax=ax, **kwargs)
         ax.set_ylabel(f"{sed_type} [{ax.yaxis.units}]")
         ax.set_yscale("log")
--- a/gammapy/maps/axes.py
+++ b/gammapy/maps/axes.py
@@ -2370,7 +2370,7 @@ class TimeMapAxis:
         ax : `~matplotlib.pyplot.Axis`
             Formatted plot axis
         """
-        from matplotlib.dates import DateFormatter
+        from matplotlib.dates import DateFormatter, ticker
 
         xlabel = DEFAULT_LABEL_TEMPLATE.format(
             quantity=PLOT_AXIS_LABEL.get(self.name, self.name.capitalize()),
@@ -2380,13 +2380,16 @@ class TimeMapAxis:
 
         if self.time_format == "iso":
             ax.xaxis.set_major_formatter(DateFormatter("%Y-%m-%d %H:%M:%S"))
-            plt.setp(
-                ax.xaxis.get_majorticklabels(),
-                rotation=30,
-                ha="right",
-                rotation_mode="anchor",
-            )
-
+        elif self.time_format == "mjd":
+            ax.xaxis.set_major_formatter(ticker.StrMethodFormatter("{x:,.5f}"))
+        else:
+            raise ValueError(f"Not a supported format: {self.time_format}")
+        plt.setp(
+            ax.xaxis.get_majorticklabels(),
+            rotation=30,
+            ha="right",
+            rotation_mode="anchor",
+        )
         return ax
 
     def assert_name(self, required_name):
--- a/gammapy/maps/axes.py
+++ b/gammapy/maps/axes.py
@@ -2160,10 +2160,10 @@ class TimeMapAxis:
     """
 
     node_type = "intervals"
-    time_format = "iso"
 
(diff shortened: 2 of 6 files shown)


gammapy/estimators/points/core.py
```diff
@@ -482,6 +482,8 @@ def plot(self, ax=None, sed_type=None, energy_power=0, **kwargs):
             Sed type
         energy_power : float
             Power of energy to multiply flux axis with
+        time_format : {"iso", "mjd"}
+            Default: "iso"
```
adonath, line 486: Maybe just clarify that this only applies if a time axis is present...

gammapy/maps/region/ndmap.py
```diff
@@ -82,6 +84,8 @@ def plot(self, ax=None, axis_name=None, **kwargs):
                 )
 
         axis = self.geom.axes[axis_name]
+        if isinstance(axis, TimeMapAxis):
+            axis.time_format = time_format
```
adonath, line 88: This will change the time format permanently on the object, not just for plotting. I'll make a suggestion in a different comment...

adonath (commented):
Thanks @bkhelifi! I think should treat the `TimeMapAxis` as default. So I suggest a behavior along the lines of:

```python

def format_plot_axis(..., time_format=None):
    if time_format is None:
        time_format = self.time_format
```

gammapy/maps/region/ndmap.py
```diff
@@ -52,7 +52,7 @@ def __init__(self, geom, data=None, dtype="float32", meta=None, unit=""):
         self.meta = meta
         self._unit = u.Unit(unit)
 
-    def plot(self, ax=None, axis_name=None, **kwargs):
+    def plot(self, ax=None, axis_name=None, time_format="iso", **kwargs):
```
AtreyeeS, line 55: Since our Maps are very general, I dont know how nice it is to have explicit arguments for configuring each axis type... Maybe having another `kwargs` would be better?

adonath: So you propose something along the lines of:

```python3
map.plot(axis_name="time", axis_plot_kwargs={"time_format": "mjd"} )
``` 
Where `axis_plot_kwargs` would be forwarded to `axes[axis_name].format_plot_xaxis(**axis_plot_kwargs)`.

I tend to agree, except that right now we do not have arguments for e.g. `MapAxis.format_plot_xaxis()`, so right now there is no other uses case than the time format. If we can imagine to have other use cases then it is probably a good idea.

AtreyeeS: @adonath : Yes, we just discussed this, and at present we dont see other use cases. So maybe this can remain as is now, and in future we can move it to kwargs. This should be backwards compatible

gammapy/maps/region/ndmap.py
adonath: I see, but isn't it better then to just modify the solution we currently have? So we could change the `time_format` from as class attribute to an instance attribute. This way use can more easily change the time format between to plots. We could also implement a setter to check for the valid formats...
 
I think the implicit change of `MapAxis.time_format` by calling `.plot()` is not a good idea. Then we should rather use the explicit solution:

```
m.geom.axes["time"].time_format = "mjd"
m.plot(axis_name="time")

# instead of

m.plot(axis_name="time", time_format="mjd")
 
```
If we prefer the latter solution in terms of API we have to choose the solution you propose and add the support to all plot methods...

gammapy/maps/region/ndmap.py
AtreyeeS: @adonath : Yes, we just discussed this, and at present we dont see other use cases. So maybe this can remain as is now, and in future we can move it to kwargs. This should be backwards compatible

gammapy/maps/axes.py
```diff
@@ -2380,13 +2382,16 @@ def format_plot_xaxis(self, ax):
...
             ax.xaxis.set_major_formatter(DateFormatter("%Y-%m-%d %H:%M:%S"))
-            plt.setp(
-                ax.xaxis.get_majorticklabels(),
-                rotation=30,
-                ha="right",
-                rotation_mode="anchor",
-            )
-
+        elif self.time_format == "mjd":
+            ax.xaxis.set_major_formatter(ticker.StrMethodFormatter("{x:,.5f}"))
```
registerrier, line 2386: I find the output not so easy to read. Isn't the standard labeling of matplotlib sufficient for this case? 
Maybe the slanted labeling is not necessary as well.

gammapy/maps/tests/test_axes.py
```diff
@@ -763,6 +763,18 @@ def test_map_axis_format_plot_xaxis():
...
 
 
+def test_time_map_axis_plot_center(time_intervals):
+    axis = TimeMapAxis(
+        time_intervals["t_min"],
+        time_intervals["t_max"],
+        time_intervals["t_ref"],
+        name="time",
+    )
+    axis.time_format = "null"
```
registerrier, line 773: Again it would make more sense if this is the code that raises the exception.

registerrier (commented):
Thanks @bkhelifi . 

My main comment here is that it is weird to add multiple exception raise and multiple test for the case where `time_format` is incorrect.
It would seem more rational to create a `setter` property that would fail if the input value is incorrect, i.e when running `axis.time_format = "null"`

gammapy/maps/axes.py
bkhelifi: If I remove if, the output seems to me not very nice:

Capture d’écran du 2023-04-17 21-52-51

I prefer this one:

Capture d’écran du 2023-04-17 22-02-18

This is a question of taster. I leave you decide...
In any case, I have simplified the formatting. Thanks

gammapy/maps/core.py
```diff
@@ -51,9 +51,9 @@ def __init__(self, geom, data, meta=None, unit=""):
             self._unit = u.Unit(unit)
 
         if meta is None:
-            self.meta = {}
+            self._meta = {}
```
registerrier, line 54: why this change here?

registerrier (approved):
Thanks @bkhelifi . Looks good. 
I have only one comment, why do you change the `Map.meta` attribute? I suppose this was not intended to go in this PR.

gammapy/maps/core.py
bkhelifi: fixed

registerrier (approved):
Thanks @bkhelifi . No more comments from my side.

Issue #4222, opened 2022-11-29
plot_spectrum_datasets_off_regions failing with many datasets


LukasNickel:
**Gammapy version**
```
In [3]: gammapy.__version__
Out[3]: '1.0'

In [5]: matplotlib.__version__
Out[5]: '3.6.2'
```

**Bug description**
If you pass too many datasets to `plot_spectrum_datasets_off_regions` it fails with a key_error:

```python
----> 1 plot_spectrum_datasets_off_regions(datasets)

File /scratch/lnickel/conda/envs/M87/lib/python3.9/site-packages/gammapy/visualization/utils.py:90, in plot_spectrum_datasets_off_regions(datasets, ax, legend, legend_kwargs, **kwargs)
     88 plot_kwargs = kwargs.copy()
     89 plot_kwargs["facecolor"] = "None"
---> 90 plot_kwargs.setdefault("edgecolor", props.pop("color"))
     91 plot_kwargs.update(props)
     93 dataset.counts_off.plot_region(ax, **plot_kwargs)

KeyError: 'color'
```
This is due to there being more datasets than colors in the `prop_cycle`

**Expected behavior**
It should reuse colors or default to the first/last one or just black.

**To Reproduce**
This MWE is really only matplotlib, but it shows the problem:
```python
for i, c in zip(range(20), plt.rcParams["axes.prop_cycle"]()):
    print(i, c.pop("color"))
```
Note that if you dont create the itertools object (so just `plt.rcParams["axes.prop_cycle"]`), zip will stop if there are no more colors left,
but this way it will continue and try to start from the beginning again. But it cannot do that, because we popped the elements before.

So if you instead write:
```python
for i, c in zip(range(20), plt.rcParams["axes.prop_cycle"]()):
    print(i, c)
```
it will loop around and if you write

```python
for i, c in zip(range(20), plt.rcParams["axes.prop_cycle"]):
    print(i, c)
```
it stops after all colors have been used once

**Other information**
This is the relevant code:
https://github.com/gammapy/gammapy/blob/647c0001a08fbda055a7aed4bf62747e08ea38e7/gammapy/visualization/utils.py#L85-L93

This did work with the same datasets on an older gammapy-version, but I do not see any changes there. Maybe the behaviour of matplotlib changed at some point regarding the popping of elements.

adonath:
[...]

PR #4397, opened 2023-03-22
Fix plot_spectrum_datasets_off_regions with too many regions

bkhelifi:
**Description**
This pull request deals with the Issue #4222 . 
When the dataset list is too long, it creates a new list of color with the correct dimension.

### DIFF
--- a/gammapy/estimators/points/core.py
+++ b/gammapy/estimators/points/core.py
@@ -471,7 +471,7 @@ class FluxPoints(FluxMaps):
 
         return y_errn, y_errp
 
-    def plot(self, ax=None, sed_type=None, energy_power=0, **kwargs):
+    def plot(self, ax=None, sed_type=None, energy_power=0, time_format="iso", **kwargs):
         """Plot flux points.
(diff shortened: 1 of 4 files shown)


gammapy/estimators/points/core.py
```diff
@@ -482,6 +482,8 @@ def plot(self, ax=None, sed_type=None, energy_power=0, **kwargs):
             Sed type
         energy_power : float
             Power of energy to multiply flux axis with
+        time_format : {"iso", "mjd"}
```
AtreyeeS, line 485: This should belong to PR #4395 , no?

adonath:
Thanks @bkhelifi! I think the required fix for #4222 would be to just get rid of the `props.pop()` statement. By default the color cycling of matplotlib is "infinite". There is no need to create new colors. If users still would like to do this, they can pass in the cycler from the outside.

registerrier (approved):
Thanks @bkhelifi ! This looks good. No further comments.

PR #4398, opened 2023-03-22
Add profilers to dev environment

QRemy:
Add profilers to dev environment


registerrier (commented):
Thanks @QRemy . I have left a few comments.

We have to see whether the daily run is enough, though. Conversely, one could add a label `profile-benchmarks` on the PR to trigger the profiling.

registerrier (approved):
Thanks @QRemy!

No more comments from my side.

PR #4402, opened 2023-03-22
Support for parallel evaluation in FluxPointsEstimator

QRemy:
Support for parallel evaluation in FluxPointsEstimator

### DIFF
--- a/gammapy/analysis/core.py
+++ b/gammapy/analysis/core.py
@@ -309,6 +309,7 @@ class Analysis:
             energy_edges=energy_edges,
             source=fp_settings.source,
             fit=self.fit,
+            n_jobs=self.config.general.n_jobs,
             **fp_settings.parameters,
         )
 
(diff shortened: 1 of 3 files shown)


registerrier (dismissed):
Thanks @QRemy ! This looks good. No major comment from my side. 

We have a `n_jobs` parameter defined on the `AnalysisConfig` in the general section. Maybe, one would expect that this is used to change the call in the `Analysis` class. (otherwise defining `n_jobs` among the parameters of the `FluxPointsEstimator` works too).

adonath:
Thanks @QRemy! As this becomes a new pattern, can you maybe propose a new reusable function to handle the pool context and fall-back? 

I'm thinking of something along the lines of:

```python3

results = execute_parallel(func, arguments, backend="")
```
Where `execute_parallel()` would then live in `gammapy.utils`. 

Can we keep the progress for parallel execution as well, maybe via a callback functions?

QRemy:
> Thanks @QRemy! As this becomes a new pattern, can you maybe propose a new reusable function to handle the pool context and fall-back?
> Can we keep the progress bar for parallel execution as well, maybe via a callback function?

I added this in #4406

registerrier (approved):
Thanks @QRemy . No more comments.

PR #4403, opened 2023-03-22
Fix issue in metadata when using `RegionNDMap.write`

fabiopintore:
This PR solves a bug `RegionNDMap.write` that affects the metadata related to a reference time (`MJDREFF` and `MJDREFI`). Now, if this keyword are present in the map header, they are transferred to the written file.

### DIFF
--- a/gammapy/maps/region/ndmap.py
+++ b/gammapy/maps/region/ndmap.py
@@ -9,6 +9,7 @@ from astropy.visualization import quantity_support
 import matplotlib.pyplot as plt
 from gammapy.utils.interpolation import ScaledRegularGridInterpolator, StatProfileScale
 from gammapy.utils.scripts import make_path
+from gammapy.utils.time import time_ref_from_dict, time_ref_to_dict
 from ..axes import MapAxes
 from ..core import Map
 from ..geom import pix_tuple_to_idx
@@ -516,6 +517,10 @@ class RegionNDMap(Map):
         if format in ["ogip", "ogip-sherpa", "ogip-arf", "ogip-arf-sherpa"]:
             hdulist.append(fits.BinTableHDU(table))
         elif format == "gadf":
+            if self.meta.setdefault("MJDREFI"):
+                t_ref = time_ref_from_dict(self.meta)
+                time_ref_dict = time_ref_to_dict(t_ref)
+                table.meta.update(time_ref_dict)
             table.meta.update(self.geom.axes.to_header())
             hdulist.append(fits.BinTableHDU(table, name=hdu))
         else:


adonath:
Thanks @fabiopintore! Can you please double check, why your case is not covered by https://github.com/gammapy/gammapy/blob/main/gammapy/maps/axes.py#L1896?

gammapy/maps/region/ndmap.py
```diff
@@ -516,6 +517,10 @@ def to_hdulist(self, format="gadf", hdu="SKYMAP", hdu_bands=None, hdu_region=Non
         if format in ["ogip", "ogip-sherpa", "ogip-arf", "ogip-arf-sherpa"]:
             hdulist.append(fits.BinTableHDU(table))
         elif format == "gadf":
+            if self.meta.setdefault("MJDREFI"):
```
adonath, line 520: This does not seem right to me. This kind of meta data should not be in `RegionNDMap.meta`, but the stored in the `TimeMapAxis` instead.

adonath:
After private communication with @fabiopintore:

- The issue is that in the `LightcurveTemplateTemporalModel` the `RegionNDMap` handles the (relative) time as a normal `MapAxis`
- The reference time is stored in `LightcurveTemplateTemporalModel.t_ref` and not in a `TimeMapAxis`
- String any reference time information in `map.meta` is a bad idea, because it is not possible to sync with the `t_ref`

I think the behavior should be:
- For consistency with other models the reference time should be serialized in YAML because it is a model parameter
- For convenience it could be stored in the FITS file as well, but note that this may not be a valid GADF data format. At least none we can handle right now, because we do not have a `TimeMapAxis` that supports center bins.
- In this case the reference time should be handled in `LightcurveTemplateTemporalModel.write()` and `.read()` instead

adonath:
Looking at the current implementation, there are some bugs:
- `LightcurveTemplateTemporalModel.to_table()` should take the current `t_ref` and add the info to the `table.meta`
- `LightcurveTemplateTemporalModel.write()` should store the current `t_ref` in the header of the map
- I think adding the table meta here should be removed https://github.com/gammapy/gammapy/blob/main/gammapy/modeling/models/temporal.py#L525, to avoid the duplicated information with `t_ref`
-

bkhelifi:
About GADF, the LC format is not really specified. So, the compliance with GADF is not really a question, unfortunately I would say.
However, there are very good advice and recommendations for this timing information in GADF: https://gamma-astro-data-formats.readthedocs.io/en/v0.3/general/time.html?highlight=TIMEREF#formats. And from that, I do not see any real problem to store the above-mentioned quantities into our FITS header.
Are you sharing this view?

fabiopintore:
Thansk @adonath and @bkhelifi .
Till now we don't have even a consesus on the format for this kind of energy-dependent temporal model, therefore I would suggest to implement Axel's suggestions but also to keep the time_ref information in the fits header.

AtreyeeS:
This is being handled in the update of the `LightCurveTemplateTemporalModels` and the metadata pig

PR #4406, opened 2023-03-23
Add configuration and helper function to run multiprocessing or ray

QRemy:
Add support for multiprocessing with ray and add global configuration options

### DIFF
--- a/gammapy/estimators/core.py
+++ b/gammapy/estimators/core.py
@@ -4,7 +4,6 @@ import inspect
 from copy import deepcopy
 import numpy as np
 from astropy import units as u
-import gammapy.utils.parallel as parallel
 from gammapy.maps import MapAxis
 from gammapy.modeling.models import ModelBase
 
@@ -24,17 +23,6 @@ class Estimator(abc.ABC):
     @abc.abstractmethod
     def run(self, datasets):
         pass
-
-    @property
-    def n_jobs(self):
-        if self._n_jobs is None:
-            return parallel.N_PROCESSES
-        else:
-            return self._n_jobs
-
-    @n_jobs.setter
-    def n_jobs(self, value):
-        self._n_jobs = value
 
     @property
     def selection_optional(self):
--- a/gammapy/estimators/core.py
+++ b/gammapy/estimators/core.py
@@ -4,6 +4,7 @@ import inspect
 from copy import deepcopy
 import numpy as np
 from astropy import units as u
+import gammapy.utils.parallel as parallel
 from gammapy.maps import MapAxis
 from gammapy.modeling.models import ModelBase
 
@@ -23,6 +24,17 @@ class Estimator(abc.ABC):
     @abc.abstractmethod
     def run(self, datasets):
         pass
+
+    @property
+    def n_jobs(self):
+        if self._n_jobs is None:
+            return parallel.N_PROCESSES
+        else:
+            return self._n_jobs
+
+    @n_jobs.setter
+    def n_jobs(self, value):
+        self._n_jobs = value
 
     @property
     def selection_optional(self):
--- a/gammapy/estimators/points/lightcurve.py
+++ b/gammapy/estimators/points/lightcurve.py
@@ -128,19 +128,20 @@ class LightCurveEstimator(FluxPointsEstimator):
 
             valid_intervals.append([t_min, t_max])
 
-            if self.n_jobs is None or self.n_jobs == 1:
+            if self.n_jobs == 1:
                 fp = self.estimate_time_bin_flux(datasets_to_fit, dataset_names)
                 rows.append(fp)
             else:
                 parallel_datasets.append(datasets_to_fit)
 
-        if self.n_jobs is not None and self.n_jobs > 1:
+        if self.n_jobs > 1:
             rows = parallel.run_multiprocessing(
                 self.estimate_time_bin_flux,
                 zip(
                     parallel_datasets,
                     repeat(dataset_names),
                 ),
+                backend=self.parallel_backend,
                 pool_kwargs=dict(processes=self.n_jobs),
                 task_name="Time intervals",
             )
--- a/gammapy/estimators/points/lightcurve.py
+++ b/gammapy/estimators/points/lightcurve.py
@@ -79,10 +79,10 @@ class LightCurveEstimator(FluxPointsEstimator):
 
     tag = "LightCurveEstimator"
 
-    def __init__(self, time_intervals=None, atol="1e-6 s", n_jobs=1, **kwargs):
+    def __init__(self, time_intervals=None, atol="1e-6 s", **kwargs):
         self.time_intervals = time_intervals
         self.atol = u.Quantity(atol)
-        self.n_jobs = n_jobs
+
         super().__init__(**kwargs)
 
     def run(self, datasets):
@@ -128,13 +128,13 @@ class LightCurveEstimator(FluxPointsEstimator):
 
             valid_intervals.append([t_min, t_max])
 
-            if self.n_jobs == 1:
+            if self.n_jobs is None or self.n_jobs == 1:
                 fp = self.estimate_time_bin_flux(datasets_to_fit, dataset_names)
                 rows.append(fp)
             else:
                 parallel_datasets.append(datasets_to_fit)
 
-        if self.n_jobs > 1:
+        if self.n_jobs is not None and self.n_jobs > 1:
             rows = parallel.run_multiprocessing(
                 self.estimate_time_bin_flux,
                 zip(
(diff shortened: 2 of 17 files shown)


gammapy/utils/pbar.py
```diff
@@ -38,7 +38,6 @@ def progress_bar(iterable, desc=None):
     return tqdm(
         iterable_to_list,
         total=total,
-        mininterval=0,
```
QRemy, line 41: because without this line the tsmap_estimator benchmark is twice faster

gammapy/estimators/points/lightcurve.py
```diff
@@ -133,14 +135,15 @@ def run(self, datasets):
...
         if self.n_jobs > 1:
-            with Pool(processes=self.n_jobs) as pool:
-                rows = pool.starmap(
-                    self.estimate_time_bin_flux,
-                    zip(
-                        parallel_datasets,
-                        repeat(dataset_names),
-                    ),
-                )
+            rows = parallel.run_multiprocessing(
```
registerrier, line 138: Maybe the lightcurve multiprocessing could be tested as well?

gammapy/makers/reduce.py
```diff
@@ -58,12 +58,23 @@ def __init__(
                 self._apply_cutout = False
             else:
                 self.cutout_width = 2 * self.offset_max
-        self.n_jobs = n_jobs
+        self._n_jobs = n_jobs
```
registerrier, line 61: Maybe rely on the `setter` rather than set the internal variable. This way the setter will be tested.

gammapy/utils/parallel.py
```diff
@@ -0,0 +1,89 @@
...
+    inputs,
+    backend=None,
+    pool_kwargs=None,
+    method="starmap",
+    method_kwargs=None,
+    task_name="",
+):
+    """Run function in a loop or in parralel"""
+
+    if method not in ["starmap", "apply_async"]:
```
registerrier, line 42: Maybe add a test for the incorrect method.

gammapy/utils/parallel.py
```diff
@@ -0,0 +1,89 @@
...
+        raise ValueError("Invalid multiprocessing method")
+
+    multiprocessing = get_multiprocessing(backend)
+    if method_kwargs is None:
+        method_kwargs = {}
+    if pool_kwargs is None:
+        pool_kwargs = {}
+    pool_kwargs.setdefault("processes", N_PROCESSES)
+    if backend == "ray":
+        pool_kwargs.setdefault("ray_adress", "auto")
```
registerrier, line 52: is it the correct argument name? 
Maybe add a test here as well.

gammapy/utils/parallel.py
```diff
@@ -0,0 +1,89 @@
...
+            return run_pool(pool, func, inputs, method, method_kwargs, task_name)
+
+
+def run_loop(func, inputs, method_kwargs=None, task_name=""):
+    """Loop over inputs an run function"""
+
+    results = []
+    for arguments in progress_bar(inputs, desc=task_name):
+        output = func(*arguments)
+        if "callback" in method_kwargs:
```
registerrier, line 70: Maybe add a test here?

registerrier (commented):
Thanks @QRemy. This looks really nice. I have left some inline comments.

gammapy/makers/reduce.py
QRemy, line 61:
```suggestion
        self.n_jobs = n_jobs
```

gammapy/utils/parallel.py
QRemy: done

gammapy/makers/reduce.py
```diff
@@ -156,24 +167,20 @@ def run(self, dataset, observations, datasets=None):
...
-                            obs,
-                        ),
-                        callback=self.callback,
-                        error_callback=self.error_callback,
-                    )
-                    results.append(result)
-                # wait async run is done
-                [result.wait() for result in results]
+            log.info("Using {} jobs.".format(n_jobs))
+            parallel.run_multiprocessing(
```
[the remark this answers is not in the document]
QRemy: yes simplified

gammapy/estimators/points/lightcurve.py
QRemy: fixed, the n_jobs argument was not set correctly

gammapy/utils/parallel.py
QRemy: corrected the argument
not sure how I can add a test for that

gammapy/utils/parallel.py
QRemy: for now it is tested only with the DatasetsMaker

registerrier (commented):
Thanks @QRemy. 

I have a last comment. To avoid introducing a `n_job` property on all `Estimator` maybe create a Mixin class that will simply provide the `n_job` setter and getter. It can then be also used in `DatasetMaker`.
You might even add a multiprocessing_backend setter and getter.

gammapy/estimators/core.py
```diff
@@ -24,6 +25,17 @@ def tag(self):
...
+    @property
+    def n_jobs(self):
+        if self._n_jobs is None:
+            return parallel.N_PROCESSES
+        else:
+            return self._n_jobs
+
+    @n_jobs.setter
+    def n_jobs(self, value):
+        self._n_jobs = value
```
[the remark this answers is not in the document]
QRemy: i think it need to be done in the property in case `parallel.N_PROCESSES changed`

adonath:
@QRemy Please rebase after #4504 and check if the test fail here goes away...Thanks!

QRemy:
> py310-test-alldeps-cov



> @QRemy Please rebase after #4504 and check if the test fail here goes away...Thanks!

Done, the previous fail is solved.
Now py310-test-alldeps-cov run for ever not sure if it is related.

registerrier (approved):
Thanks @QRemy . 

This looks good. It seems there is unstability with ray and the lightcurve calculation. So we should explicitly say that ray support is a only a prototype and is not guaranteed to work.

Shouldn't we add a warning when setting `ray` as a backend? Opinion @adonath ?

registerrier (approved):
Thanks @QRemy . 

This looks good. It seems there is unstability with ray and the lightcurve calculation. So we should explicitly say that ray support is a only a prototype and is not guaranteed to work.

Shouldn't we add a warning when setting `ray` as a backend? Opinion @adonath ?

QRemy:
I added a `warning_prototype` function so it could be re-used, for example in #3047, or if more backends are added.

gammapy/utils/parallel.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+MULTIPROCESSING_BACKEND = "multiprocessing"
+N_PROCESSES = 1
+N_THREADS = 1
+
+
+def warning_prototype(module="ray"):
+    log.warning(
+        "gammapy supports for parallelisation with {module} is still a prototype and is not fully functionnal."
```
QRemy, line 17:
```suggestion
        "gammapy supports for parallelisation with {module} is still a prototype and is not fully functional."
```

gammapy/utils/parallel.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+MULTIPROCESSING_BACKEND = "multiprocessing"
+N_PROCESSES = 1
+N_THREADS = 1
+
+
+def warning_prototype(module="ray"):
+    log.warning(
+        "gammapy supports for parallelisation with {module} is still a prototype and is not fully functional."
```
QRemy, line 17:
```suggestion
        f"gammapy supports for parallelisation with {module} is still a prototype and is not fully functional."
```

gammapy/utils/parallel.py
```diff
@@ -0,0 +1,128 @@
...
+        raise ValueError("Invalid multiprocessing method")
+
+    multiprocessing = get_multiprocessing(backend)
+    if method_kwargs is None:
+        method_kwargs = {}
+    if pool_kwargs is None:
+        pool_kwargs = {}
+    pool_kwargs.setdefault("processes", N_PROCESSES)
+    if backend == "ray":
+        pool_kwargs.setdefault("ray_address", "auto")
```
QRemy, line 84:
```suggestion
    if backend == "ray":
        from ray import is_initialized
        if is_initialized():
            address= "auto"
        else:
            address = None
        pool_kwargs.setdefault("ray_address", "auto")
```

gammapy/utils/parallel.py
```diff
@@ -0,0 +1,133 @@
...
+    if pool_kwargs is None:
+        pool_kwargs = {}
+    pool_kwargs.setdefault("processes", N_PROCESSES)
+    if backend == "ray":
+        from ray import is_initialized
+        if is_initialized():
+            address= "auto"
+        else:
+            address = None
+        pool_kwargs.setdefault("ray_address", "auto")
```
QRemy, line 89:
```suggestion
        pool_kwargs.setdefault("ray_address", address)
```

adonath (approved):
Thanks @QRemy! I have no further comments, I think the mixin is a very elegant solution for now! The remaining issue with the covariance we will fix independently.

registerrier (approved):
Thanks @QRemy. No further comments.

Issue #3903, opened 2022-04-29
Livetime info is misleading?


luca-giunti:
**Gammapy version**
0.19

**Bug description**
I find that the current handling of the `livetime` entry in the Map/SpectrumDataset `info_dict` is a bit misleading. It can lead to unexpected situations like the following: an empty dataset cutout or a spectrum dataset extracted from an empty region can have nonzero `livetime`, since the latter is propagated by simply adding up the `livetime` value every time that two datasets are stacked. I show this in an [example notebook](https://github.com/luca-giunti/share/blob/master/debug.ipynb). 

Basically, the `livetime` value shown in the `info_dict` corresponds to the total livetime of the parent stacked dataset, which is well defined but (I think) a bit useless. In any case it should perhaps not be propagated as-is to cutouts or spectrum datasets.

adonath:
Thank @luca-giunti! I agree storing the livetime as a single number is not the best solution as it gives rise to issue like the one you reported here. I remember we discussed a similar issue in the context of stacking datasets before. @registerrier proposed to not stack the livetime if the mask safe was all zero. However this is not a good solution either, because the different value of even a single pixel would change the behavior. I think the only correct and reasonable solution is to work with a livetime map here. It‘s the only way to store the information correctly and pass it on to other datasets.

registerrier:
Livetime maps creation functions have been introduced. We can close now.

PR #4407, opened 2023-03-23
Add a livetime map on dataset

AtreyeeS:
This addresses #3903 by adding a new member on the `MapDataset`.
The serialisation is not implemented.

### DIFF
--- a/gammapy/datasets/map.py
+++ b/gammapy/datasets/map.py
@@ -127,6 +127,8 @@ class MapDataset(Dataset):
     meta_table : `~astropy.table.Table`
         Table listing information on observations used to create the dataset.
         One line per observation for stacked datasets.
+    livetime_map : `~gammapy.maps.WcsNDMap`
+        The (non-acceptance corrected) livetime at each position of the map
 
     If an `HDULocation` is passed the map is loaded lazily. This means the
     map data is only loaded in memory as the corresponding data attribute
@@ -180,6 +182,7 @@ class MapDataset(Dataset):
     psf = LazyFitsData(cache=True)
     mask_fit = LazyFitsData(cache=True)
     mask_safe = LazyFitsData(cache=True)
+    livetime_map = LazyFitsData(cache=True)
 
     _lazy_data_members = [
         "counts",
@@ -189,6 +192,7 @@ class MapDataset(Dataset):
         "mask_fit",
         "mask_safe",
         "background",
+        "livetime_map",
     ]
 
     def __init__(
@@ -204,6 +208,7 @@ class MapDataset(Dataset):
         gti=None,
         meta_table=None,
         name=None,
+        livetime_map=None,
     ):
         self._name = make_name(name)
         self._evaluators = {}
@@ -234,6 +239,7 @@ class MapDataset(Dataset):
         self.gti = gti
         self.models = models
         self.meta_table = meta_table
+        self.livetime_map = livetime_map
 
     # TODO: keep or remove?
     @property
@@ -573,6 +579,9 @@ class MapDataset(Dataset):
             "gti", GTI.create([] * u.s, [] * u.s, reference_time=reference_time)
         )
         kwargs["mask_safe"] = Map.from_geom(geom, unit="", dtype=bool)
+        kwargs["livetime_map"] = kwargs["mask_safe"].reduce_over_axes(
+            func=np.logical_or
+        ) * (0.0 * u.s)
         return cls(**kwargs)
 
     @classmethod
@@ -813,6 +822,9 @@ class MapDataset(Dataset):
             self.meta_table = hstack_columns(self.meta_table, other.meta_table)
         elif other.meta_table:
             self.meta_table = other.meta_table.copy()
+
+        if self.livetime_map and other.livetime_map:
+            self.livetime_map.stack(other.livetime_map, weights=other.mask_safe_image)
 
     def stat_array(self):
         """Statistic function value per bin given the current model parameters"""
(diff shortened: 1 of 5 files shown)


gammapy/datasets/map.py
```diff
@@ -204,6 +208,7 @@ def __init__(
         gti=None,
         meta_table=None,
         name=None,
+        livetime_map=None,
```
adonath, line 211: Please rename to `livetime`, the other attributes do not carry the `_map` prefix either. The type can be clarified in the docstring.

gammapy/datasets/map.py
```diff
@@ -573,6 +579,9 @@ def from_geoms(
             "gti", GTI.create([] * u.s, [] * u.s, reference_time=reference_time)
         )
         kwargs["mask_safe"] = Map.from_geom(geom, unit="", dtype=bool)
+        kwargs["livetime_map"] = kwargs["mask_safe"].reduce_over_axes(
```
adonath, line 582: The livetime can be initialized as a non-masked map. Just as for the other qualities the mask will be applied in `.stack()` and `.to_masked()`

gammapy/makers/map.py
```diff
@@ -412,4 +414,9 @@ def run(self, dataset, observation):
 
             kwargs["edisp"] = edisp
 
+        kwargs["livetime_map"] = (
+            kwargs["mask_safe"].reduce_over_axes(func=np.logical_or)
```
adonath, line 418: I would leave the livetime map un-masked here...

adonath (commented):
Thanks @AtreyeeS, this is useful information added to the `MapDataset`. One more general comment: I think the livetime map also has to be stored as 3D map. When stacking different observations can have different energy thresholds resulting in an energy dependent livetime.

Currently the scalar livetime is stored in the exposure map as well, this could probably be deprecated. But I would need to think about how...

AtreyeeS:
@adonath We discussed this a bit yesterday, and if we correct for the energy, why not for the acceptance? This information is then correctly available in the exposure map, and, we show how to calculate a livetime map in the HESS tutorial.

This map is useful to because people often want to quote "there were x hours of observations of this region"

I added the `_map` tag to not confuse with the `livetime` present on the `info_dict` (which is taken from the  exposure map). Even if deprecated, we need to support both for the next version, no?

gammapy/datasets/map.py
adonath: No, but `Map.from_geom(geom, unit="s")` is simpler I guess..

gammapy/makers/map.py
AtreyeeS: The livetime map for a single observation should take the mask into account

adonath: For single observations we leave the quantities unmasked. The mask is only applied when stacking or calling `.to_masked()`. This has the huge advantage that the safe data range can be modified after data reduction if needed.

AtreyeeS:
I made the `livetime_map` hidden. We can expose this once we have some clarity on the meta handling. I can start to prepare a PIG for this. Maybe this PR can be merged independently ?

gammapy/makers/map.py
AtreyeeS: Yes indeed, my bad. If the counts are there, then the livetime should also be.

gammapy/datasets/map.py
```diff
@@ -127,6 +127,8 @@ class MapDataset(Dataset):
     meta_table : `~astropy.table.Table`
         Table listing information on observations used to create the dataset.
         One line per observation for stacked datasets.
+    livetime_map : `~gammapy.maps.WcsNDMap`
```
bkhelifi, line 130: As the livetime_map is hidden, should we describe it into the docstring?

gammapy/datasets/map.py
```diff
@@ -1877,6 +1887,7 @@ def resample_energy_axis(self, energy_axis, name=None):
         else:  # None or EDispMap
             kwargs["edisp"] = self.edisp
 
+        kwargs["_livetime_map"] = self._livetime_map
```
bkhelifi, line 1890: Maybe one should add a `if self._livetime_map is not None:` ...

gammapy/datasets/map.py
```diff
@@ -2239,7 +2255,7 @@ def to_map_dataset(self, name=None):
         The background model template is taken as alpha * counts_off
 
         Parameters
-        ----------
+        ----------c
```
bkhelifi, line 2258:
```suggestion
        ----------
```

gammapy/datasets/tests/test_map.py
```diff
@@ -930,6 +934,7 @@ def test_stack(sky_model):
     assert_allclose(stacked.mask_safe.data.sum(), 4600)
     assert_allclose(stacked.mask_fit.data.sum(), 4600)
     assert_allclose(stacked.exposure.data.sum(), 1.6e11)
+    assert_allclose(stacked._livetime_map.data.sum(), 240000.0)
 
     assert_allclose(stacked.meta_table["OBS_ID"][0], [0, 1])
```
bkhelifi, line 940: One could profit of this PR to check if the metadata livetime exists and contains the correct value

bkhelifi:
A side question, do you test that the mask is correctly applied when computing the livetime_map?

adonath:
The livetime map could be handled as meta information in its own data structure. See #4491.

bkhelifi:
Based on the work associated to the PIG 25 about metadata container, this information could be added in the metadata of the dataset class.
What is the best strategy? Wait for this DatasetMetadataClass or do the merge and then update the code with the future metadata class?

AtreyeeS:
Tutorial showed in #4817 
Meta information will be added later

PR #4408, opened 2023-03-23
Add plot function for 1D distribution of map data

MRegeard:
This pull request is for the implementation of a function that allows to plot the 1D distribution of the data of a map. 

It takes as main parameter an instance of `gammapy.maps.Map`. If the dimension of map.data is less than 3, it will display a single plot. Otherwise it will plot a grid of plot as in the `plot_grid` methods that we have. 

I give the possibility to make fit of the distribution using `scipy.stats`. The function that is used for fitting can be taken from all the continuous distribution that `scipy.stats` offers. 

For the handling of the `**kwargs`, the user can put both kwargs for `matplotlib.pyplot.hist` and `matplotlib.axes.Axes`. I then automatically detect the kwargs that are for `hist` and the ones that are for `Axes`.

The main use case for this function is probably going to be plotting the distribution of significance over a map as it is shown here : https://docs.gammapy.org/dev/tutorials/analysis-2d/ring_background.html .
But since it is a general function it can actually plot any data as a 1D histogram distribution. 

I still need to make some tests.


gammapy/visualization/utils.py
```diff
@@ -280,3 +286,111 @@ def plot_theta_squared_table(table):
...
+        if idx > len(data) - 1:
+            axe.set_visible(False)
+            continue
+        d = data[idx]
+        axe.hist(d.flatten(), **hist_dict)
+
+        fit_parameters = []
+        if fit:
+
+            function = getattr(scipy.stats._continuous_distns, fit_function)
```
adonath, line 379: We should not rely on private namespaces like `_continuous_distns`, probably the distributions are just visible in `scipy.stats` as well? If not users can also just pass the method themselves (so not via a string but just the actual function object).

MRegeard: The thing is that there are not continuous distribution in `scipy.stats` , and I am not sure that we would want that. 
But I guess it's fine if the user pass directly the object.

MRegeard:
> 

Ok noted. I am going to look at this. 
For the fit I originally did that because when we plot the significance distribution, we usually do a quick fit of the data to get the parameters of the normal distribution. 
But I get your point.

adonath:
@MRegeard Ar you working on this? If it helps I could provide an independent implementation of the `.to_region_nd_map_hist()` method...

MRegeard:
@adonath not yet ...
Indeed it would helpful, if it's not to much for you.

adonath:
@MRegeard I'm working on it now. I will CC for review...

gammapy/visualization/utils.py
```diff
@@ -280,3 +286,111 @@ def plot_theta_squared_table(table):
...
+    axes : `~numpy.ndarray` of `~matplotlib.pyplot.Axes`
+        Array of Axes.
+    """
+
+    if not isinstance(map_, Map):
+        raise TypeError(
+            f"map_ must be an instance of gammapy.maps.Map, given {type(map_)}"
+        )
+
+    hist_args = list(inspect.signature(plt.hist).parameters)
```
adonath, line 325: This is a very interesting approach of separating the keyword arguments. Right now I'm not sure if this really works, if there are non-unique parameters it must fail I guess. 

In Gammapy we typically have have approach to introduce different `kwargs` arguments like, `kwargs_hist`, `kwargs_axes`, `kwargs_patch`, this is slightly more inconvenient for the users, but simpler, easier to document and guaranteed to work.

MRegeard: Yes I have tested it so it is working here ! I have checked that there are no shared parameters between these two objects. But yes it's probably not working if there are common parameters. 

I should probably stick to Gammapy way of handeling `kwargs` ! 

(I working on this function now !)

MRegeard:
@adonath the other question here is should we keep the fit functionality or let the user do it ?

adonath:
@MRegeard I think it is definitely convenient for users! The only technical concern I have is that it is build into the plotting function. This means for example if a fit fails, user have no way of interacting with the fit.  Can we re-use https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.fit.html? The returned `FitResult` object has a `.plot(ax=ax)` method. Such that something like this works:

```python3
from scipy import stats

result = stats.fit(...)
result.plot(ax=ax)
```
I just noticed that for the Scipy fit we need the unbinned data. I will think about it once again...

registerrier:
What is the status here @MRegeard ? Do we need to merge #4508 first?
Could you rebase to solve the merge conflicts?

MRegeard:
Yes we need to merge #4508 first and then it should be pretty straight forward. 
Yes I will rebase.

MRegeard:
I thought I would be able to use the `plot_grid()` method of `RegionNDMap` but when I apply `to_region_nd_map_histogram` to a `WcsNDMap`, since it creates another non spatial axis (bins) to the map I cannot use `plot_grid`.

So at the moment I create a "plot_grid" in the plot function. If the dimension of the map is 1 then it is just a normal plot. 

Maybe it would be nice to have a kwargs that specify the axis on which to grid in `plot_grid` to allow for multiple non-spatial dimension map. 

I think that for now it can stay like this and we can apply modification later. 

For the kwargs handeling, should I create two different dictionaries, to stick with gammapy way or is it good like this?

gammapy/visualization/utils.py
```diff
@@ -178,3 +183,102 @@ def plot_theta_squared_table(table):
...
+
+        axe = axes.flat[idx]
+        if idx > len(data) - 1:
+            axe.set_visible(False)
+            continue
+        d = data[idx][np.isfinite(data[idx])]
+        axe.hist(d, **hist_dict)
+
+        if fit:
+
```
registerrier, line 270:
```suggestion
```

registerrier (commented):
Thanks @MRegeard . This looks good. If the algorithm is limited `WcsNDMap` specify it directly.

registerrier (commented):
Thanks @MRegeard . This looks good. If the algorithm is limited `WcsNDMap` specify it directly.

gammapy/visualization/utils.py
```diff
@@ -178,3 +183,101 @@ def plot_theta_squared_table(table):
...
+        returns None.
+    axes : `~numpy.ndarray` of `~matplotlib.pyplot.Axes`
+        Array of Axes.
+    """
+
+    from gammapy.maps import WcsNDMap  # import here because of circular import
+
+    if not isinstance(wcs_map, WcsNDMap):
+        raise TypeError(
+            f"map_ must be an instance of gammapy.maps.Map, given {type(wcs_map)}"
```
registerrier, line 221:
```suggestion
            f"map_ must be an instance of gammapy.maps.WcsNDMap, given {type(wcs_map)}"
```

gammapy/visualization/utils.py
```diff
@@ -178,3 +183,101 @@ def plot_theta_squared_table(table):
...
+
+
+def plot_distribution(wcs_map, ax=None, ncols=3, fit=True, dist=stats.norm, **kwargs):
+    """
+    Plot the 1D distribution of data inside a map as a histogram. If the dimension of the map is smaller than 2,
+    a unique plot will be displayed. Otherwise, if the dimension is 3 or greater, a grid of plot will be displayed.
+
+    Parameters
+    ----------
+    wcs_map : an instance of `~gammapy.maps.WcsMap`
```
registerrier, line 195:
```suggestion
    wcs_map : an instance of `~gammapy.maps.WcsNDMap`
```

registerrier (dismissed):
Thanks @MRegeard . There are still two minor typos. Otherwise looks good to me.

gammapy/visualization/utils.py
```diff
@@ -178,3 +183,101 @@ def plot_theta_squared_table(table):
...
+    if not isinstance(wcs_map, WcsNDMap):
+        raise TypeError(
+            f"map_ must be an instance of gammapy.maps.WcsNDMap, given {type(wcs_map)}"
+        )
+
+    hist_args = list(inspect.signature(plt.hist).parameters)
+    patches_args = list(inspect.signature(matplotlib.patches.Patch).parameters)
+    all_hist_args = np.unique(hist_args + patches_args)
+
+    hist_dict = {k: kwargs.pop(k) for k in dict(kwargs) if k in all_hist_args}
```
registerrier, line 228: These two dictionaries should be made optional arguments of the method.

gammapy/visualization/utils.py
```diff
@@ -178,3 +183,101 @@ def plot_theta_squared_table(table):
...
+        A map that contains data to be plotted.
+    ax : `~matplotlib.axes.Axes` or list of `~matplotlib.axes.Axes`
+        Axis object to plot on. If a list of Axis is provided it has to be the same length as the length of _map.data.
+    ncols : int
+        Number of columns to plot if a "plot grid" was to be done.
+    fit : bool
+        Whether to perform a fit of the distribution of data. If True, the fit is performed by `scipy.stats`.
+    dist : `scipy.stats.rv_continuous` or `scipy.stats.rv_discrete`
+        The distribution to use for the fit.
+    **kwargs : dict
```
registerrier, line 205: Please make this argument two distinct arguments e.g. `hist_kwargs` and `axes_kwargs`

gammapy/visualization/utils.py
```diff
@@ -178,3 +183,101 @@ def plot_theta_squared_table(table):
...
+
+            x = np.linspace(np.min(d), np.max(d), 100)
+            y = dist.pdf(x, *result.params)
+
+            axe.plot(x, y, label="Fit")
+
+        axe.set(**axes_dict)
+        axe.legend()
+
+        print(result)
```
registerrier, line 280: It might be better to use logging for this: `log.info(result)`

registerrier (commented):
Sorry @MRegeard , still a few remaining issues

registerrier (commented):
Sorry @MRegeard , still a few remaining issues

adonath:
@MRegeard One last comment, please use the following pattern instead:

```
def func(..., kwargs_some=None):
    kwargs_some = {} or kwargs_some
```
The reason is e.g. explained here: https://florimond.dev/en/posts/2018/08/python-mutable-defaults-are-the-source-of-all-evil/

registerrier:
Can you rebase @MRegeard ? The CI fail is now corrected in main.

MRegeard:
I move back to fitting directly on the scipy distribution and not using the `scipy.stats.fit` because it sometime does bad things (and says that the fit is good)

MRegeard:
I modify the fitting interface to `scipy.optimize.curve_fit`. Now the user has to define a function to pass to `curve_fit` himself, and put it in the `func` argument of the plot function. I think it is better because it allows for more flexibility. 
This also enable access to more information on the fit (covariance, "fit result", etc). 
I added a argument for kwargs to pass to `curve_fit`. 
I also added an example in the docstring.

gammapy/visualization/utils.py
```diff
@@ -178,3 +181,135 @@ def plot_theta_squared_table(table):
...
+            continue
+        d = data[idx][np.isfinite(data[idx])]
+        n, bins, _ = axe.hist(d, **kwargs_hist)
+
+        if fit:
+            centers = 0.5 * (bins[1:] + bins[:-1])
+
+            pars, cov, infodict, message, _ = curve_fit(func, centers, n, **kwargs_fit)
+
+            result_dict = {"param": pars, "covar": cov, "info_dict": infodict}
```
registerrier, line 301: to be useful in case of cubes, the dict might need the axes indices or their edges/center

registerrier (approved):

I have left some small inline comments. The main remaining one is that `fit` can be removed from arguments as the presence of `func` is enough to determine whether a fit is necessary or not.

gammapy/visualization/utils.py
```diff
@@ -178,3 +181,135 @@ def plot_theta_squared_table(table):
...
+    wcs_map : an instance of `~gammapy.maps.WcsNDMap`
+        A map that contains data to be plotted.
+    ax : `~matplotlib.axes.Axes` or list of `~matplotlib.axes.Axes`
+        Axis object to plot on. If a list of Axis is provided it has to be the same length as the length of _map.data.
+    ncols : int
+        Number of columns to plot if a "plot grid" was to be done.
+    fit : bool
+        Whether to perform a fit of the distribution of data. If True, the fit is performed by `scipy.stats`.
+    func : function object
+        The function to pass to `scipy.optimize.curve_fit`
```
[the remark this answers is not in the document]
MRegeard: I am using `curve_fit` because you can have access to the covariance matrix of the fit, therefore to the error of the parameters. Moreover, I think it is more flexible, user can input some fancy function that they would have define before. The fit will basically work with every function.

gammapy/visualization/tests/test_utils.py
```diff
@@ -71,3 +73,35 @@ def test_plot_map_rgb():
...
+
+    energy_axis_10 = MapAxis.from_energy_bounds(1 * u.TeV, 10 * u.TeV, 10)
+    map_empty = WcsNDMap.create(npix=(100, 100), axes=[energy_axis_10])
+
+    def fit_func(x, mu, sigma):
+        return norm.pdf(x, mu, sigma)
+
+    with mpl_plot_check():
+        res, axes = plot_distribution(
+            wcs_map=map_, func=fit_func, kwargs_hist={"bins": 40}
```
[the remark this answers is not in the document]
MRegeard: Because of `curve_fit`, you need to define a python function that take the parameter to fit (here in the case of `norm` : `mu`and `sigma`), so that `curve_fit` knows which parameter to fit.

gammapy/visualization/utils.py
```diff
@@ -178,3 +181,135 @@ def plot_theta_squared_table(table):
...
+
+    Parameters
+    ----------
+    wcs_map : an instance of `~gammapy.maps.WcsNDMap`
+        A map that contains data to be plotted.
+    ax : `~matplotlib.axes.Axes` or list of `~matplotlib.axes.Axes`
+        Axis object to plot on. If a list of Axis is provided it has to be the same length as the length of _map.data.
+    ncols : int
+        Number of columns to plot if a "plot grid" was to be done.
+    fit : bool
```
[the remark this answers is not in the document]
MRegeard: Yes, true!

gammapy/visualization/utils.py
MRegeard: Done ! Not sure of the naming I put though.

PR #4409, opened 2023-03-23
Add a function that plot the npred_signal of models of a dataset

MRegeard:
This pull request is for the implementation of a function that plot the npred_signal() of a dataset. 
It takes a dataset as a parameter and return a plot of the npred_signal of the models of the dataset. 
By default only the 5 most contributing models. The user can also sum all the models together. 

I haven't tested yet so there might be some bugs.


adonath:
Thanks that's a very insightful plot! Again one proposal for making things more generally useful:

Modify `dataset.npred_signal(model_names=)` to take a list of model names and return a `Map` with a `LabelMapAxis` for the model names. This way you will save a lot of code later when plotting, because you can re-use `RegionNDMap.plot()`. E.g. things like:

```python3
npred_spec = npred_per_model.to_region_ndmap()
npred_spec.plot(axis_name="energy")
npred_spec.sum_over_axis(axis_name="model-name")
```
Will just work...

registerrier:
Hi @MRegeard, what is the status here? How much is there left to finish the PR?

MRegeard:
From my side the code is done. I just need to do a small test. 

I have added a new file in visualization for datasets plot, where I move this function but also `plot_spectrum_dataset_off_regions`.

gammapy/visualization/datasets.py
```diff
@@ -0,0 +1,175 @@
...
+        The dataset from which to plot the npred_signal
+    ax : `~matplotlib.axes.Axes`
+        Axis object to plot on.
+    model_names : list of str
+        The list of model for which the npres_signal is plotted. If set to None, it will take all the models
+        of the `dataset
+    stack : bool
+        Whether to sum the npred_signal of all the model.
+    region: `~regions.Region` or `~astropy.coordinates.SkyCoord`
+        Region.
```
registerrier, line 135: Region used to reproject predicted counts. If None, use full dataset geometry.
Default is None.

registerrier (commented):
Thanks @MRegeard . See some inline comments.

If you add a datasets.py file in visualization, don't forget to move the corresponding tests in tests_datasets.py.

gammapy/visualization/datasets.py
```diff
@@ -0,0 +1,175 @@
...
+
+    Plot that cycles through colors (``edgecolor``) and line styles together::
+
+        plot_spectrum_datasets_off_regions(datasets, ax, prop_cycle=plt.cycler(color=list('rgb'), ls=['--', '-', ':']))  # noqa: E501
+
+    Plot that uses a modified `~matplotlib.rcParams`, has two legend columns, static and
+    dynamic colors, but only shows labels for ``datasets1`` and ``datasets2``. Note that
+    ``legend_kwargs`` only applies if it's given in the last function call with ``legend=True``::
+
+        plt.rc('legend', columnspacing=1, fontsize=9)
```
AtreyeeS, line 56: Example code lines have to be preceded by `>>>` for doc tests I think.
See https://github.com/gammapy/gammapy/blob/5fbb60ba48ab9a04143b1b451b1085a4193f8d12/gammapy/estimators/map/excess.py#L108

gammapy/visualization/datasets.py
```diff
@@ -0,0 +1,175 @@
...
+
+
+def plot_spectrum_datasets_off_regions(
+    datasets, ax=None, legend=None, legend_kwargs=None, **kwargs
+):
+    """Plot the off regions of spectrum datasets.
+
+    Parameters
+    ----------
+    datasets : `~gammapy.datasets.Datasets` or list of `~gammapy.datasets.SpectrumDatasetOnOff`
```
AtreyeeS, line 16: Is `~gammapy.datasets.Datasets` apt here? Only `SpectrumDatasetOnOff`s are supported, so `list of ~gammapy.datasets.SpectrumDatasetOnOff` should suffice.

gammapy/visualization/datasets.py
```diff
@@ -0,0 +1,175 @@
...
+__all__ = [
+    "plot_spectrum_datasets_off_regions",
+    "plot_npred_signal",
+]
+
+
+def plot_spectrum_datasets_off_regions(
+    datasets, ax=None, legend=None, legend_kwargs=None, **kwargs
+):
+    """Plot the off regions of spectrum datasets.
```
AtreyeeS, line 12: this can be clarified better

gammapy/visualization/datasets.py
```diff
@@ -0,0 +1,175 @@
...
+    -----
+    Properties from the ``prop_cycle`` have maximum priority, except ``color``,
+    ``edgecolor``/``color`` is selected from the sources below in this order:
+    ``kwargs["edgecolor"]``, ``kwargs["prop_cycle"]``, ``matplotlib.rcParams["axes.prop_cycle"]``
+    ``matplotlib.rcParams["patch.edgecolor"]``, ``matplotlib.rcParams["patch.facecolor"]``
+    is never used.
+
+    Examples
+    --------
+    Plot forcibly without legend and with thick circles::
```
AtreyeeS, line 40: forcibly seems a bit too strong a word here!

gammapy/visualization/datasets.py
```diff
@@ -0,0 +1,175 @@
...
+
+        plot_spectrum_datasets_off_regions(datasets, ax, legend=False, linewidth=2.5)
+
+    Plot that quantifies the overlap of off regions::
```
AtreyeeS, line 44:
```suggestion
    Plot to visualise the overlap of off regions::
```

AtreyeeS (approved):
Thanks @MRegeard ! It is useful to start filling up the datasets visualisation

AtreyeeS (approved):
Thanks @MRegeard ! It is useful to start filling up the datasets visualisation

adonath:
Thanks @MRegeard! One minor side comment: please keep PR smalls in future: so first submit a PR moving the existing functionally and then open a 2nd PR introducing the new method.

MRegeard:
@AtreyeeS I haven't done this code, I just moved it from `gammapy.visualization.utils` to `gammapy.visualization.datasets`. But I can take a look at it. Probably in another PR would be better though.

MRegeard:
For the test, I have copy-paste some pytest fixture from `gammapy.datasets.tests.test_map.py`. Is there a way to import them instead ?

AtreyeeS:
I think you can use the precomputed dataset in `$GAMMAPY_DATA/cta-1dc-gc/cta-1dc-gc.fits.gz`

gammapy/visualization/tests/test_datasets.py
```diff
@@ -0,0 +1,87 @@
...
+    )
+
+    actual = ax.patches[0].get_edgecolor()
+    assert_allclose(actual, (0.121569, 0.466667, 0.705882, 1.0), rtol=1e-2)
+
+    actual = ax.patches[2].get_edgecolor()
+    assert_allclose(actual, (1.0, 0.498039, 0.054902, 1.0), rtol=1e-2)
+    assert ax.lines[0].get_color() in ["green", "C0"]
+
+
```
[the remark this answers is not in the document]
MRegeard: Ah yes right, thanks !

gammapy/visualization/datasets.py
```diff
@@ -0,0 +1,177 @@
...
+    -----
+    Properties from the ``prop_cycle`` have maximum priority, except ``color``,
+    ``edgecolor``/``color`` is selected from the sources below in this order:
+    ``kwargs["edgecolor"]``, ``kwargs["prop_cycle"]``, ``matplotlib.rcParams["axes.prop_cycle"]``
+    ``matplotlib.rcParams["patch.edgecolor"]``, ``matplotlib.rcParams["patch.facecolor"]``
+    is never used.
+
+    Examples
+    --------
+    Plot forcibly without legend and with thick circles::
```
QRemy, line 41:
```suggestion
    Plot without legend and with thick circles::
```

gammapy/visualization/datasets.py
QRemy: The examples should be fully executable, no.

gammapy/visualization/datasets.py
```diff
@@ -0,0 +1,177 @@
...
+        names = npred_region.geom.axes["models"].center
+    else:
+        npred_iter = [npred_region]
+        names = ["stacked models"]
+
+    if ax is None:
+        fig = plt.figure()
+        axes = fig.add_subplot(111)
+    else:
+        axes = ax
```
adonath, line 166: Just simplify to:

```suggestion
    if ax is None:
       ax = plt.gca()
```

gammapy/visualization/datasets.py
```diff
@@ -0,0 +1,177 @@
...
+
+    for (npred, name) in zip(npred_iter, names):
```
adonath, line 168: `RegionNDMap.plot()` already features the loop over the secondary axis. This is triggered by specifying the x-axis, like `RegionNDMap.plot(axis_name="energy")`. Please try it put and simplify the code here even further...

gammapy/visualization/datasets.py
```diff
@@ -0,0 +1,177 @@
...
+
+    axes : `~matplotlib.axes.Axes`
+        Axis object
+    """
+
+    npred_region = dataset.npred_signal(
+        model_names=model_names, stack=stack
+    ).to_region_nd_map(region)
+
+    if not stack:
```
adonath, line 155: My proposal here would be to just simplify and plot always everything:

- Plot stacked npred
- Plot npred per model
- Plot background npred 

An remove the configuration arguments. I think plotting more information is always better in this case.

gammapy/visualization/datasets.py
```diff
@@ -0,0 +1,177 @@
...
+        axes = ax
+
+    for (npred, name) in zip(npred_iter, names):
+        npred.plot(axes, label=name, **kwargs)
+    if plot_background:
+        dataset.npred_background().to_region_nd_map(region).plot(
+            ax=axes, label="background", **kwargs
+        )
+    axes.set_ylabel("Predicted counts")
+    axes.legend(loc="upper left", bbox_to_anchor=(1.04, 1), borderaxespad=0)
```
adonath, line 175: I would avoid the hard-coded legend position and just use whatever matplotlib does as a default. It is unclear whether the legend positions is the best in all cases...

adonath (changes requested):
Thanks @MRegeard, I have left some more inline comments. My recommendation here would be to simplify and just have the function always plot everything. In almost all cases it gives a more useful plot.

gammapy/visualization/datasets.py
MRegeard: done

gammapy/visualization/datasets.py
MRegeard: done

MRegeard:
I just noticed that this function was probably not supporting 1D datasets because the `to_nd_region_map` was hard coded. 
The check on the npred map instance fix this.

adonath:
@MRegeard I think the check might not be necessary, there is https://github.com/gammapy/gammapy/blob/main/gammapy/maps/region/ndmap.py#L773, which just returns the map itself. But I have not checked whether it applies in your case...

adonath:
@MRegeard In general it is very hard to write re-usable plotting code, because in most cases there are way to many options. The usual approach in Gammapy is to provide a default visualization and then users can modify the plot to their needs using the `matplotlib` API. If they need something fundamentally different or publication quality plots, users are encouraged to just rely on `Map.plot()` or the matplotlib API.

MRegeard:
@adonath So now I always plot the stacked models and the background. I just kept the check of lenght of model and if it's one I only plot the model (because the model and the stacked will obviously be the same). Is it want you meant ?

gammapy/visualization/datasets.py
```diff
@@ -0,0 +1,167 @@
...
+        Axis object to plot on.
+    model_names : list of str
+        The list of models for which the npred_signal is plotted. Default is None.
+        If None, all models are considered.
+    region: `~regions.Region` or `~astropy.coordinates.SkyCoord`
+        Region used to reproject predicted counts. Default is None.
+        If None, use the full dataset geometry.
+    **kwargs : dict
+        Keyword arguments to pass to `~gammapy.maps.RegionNDMap.plot`.
+
```
registerrier, line 138:
```suggestion
```

gammapy/visualization/datasets.py
```diff
@@ -0,0 +1,167 @@
...
+
+    Returns
+    -------
+
```
registerrier, line 142:
```suggestion
```

gammapy/visualization/datasets.py
```diff
@@ -0,0 +1,167 @@
...
+    axes : `~matplotlib.axes.Axes`
+        Axis object
+    """
+
+    npred_stack = dataset.npred_signal(
+        model_names=model_names, stack=True
+    ).to_region_nd_map(region)
+    npred_not_stack = dataset.npred_signal(
```
registerrier, line 150: Isn't it simpler to first compute `npred` and then stack it with a `sum_over_axes`?

gammapy/visualization/datasets.py
```diff
@@ -0,0 +1,167 @@
...
+        model_names=model_names, stack=False
+    ).to_region_nd_map(region)
+    npred_background = dataset.npred_background().to_region_nd_map(region)
+
+    if ax is None:
+
```
registerrier, line 156:
```suggestion
```

gammapy/visualization/datasets.py
```diff
@@ -0,0 +1,167 @@
...
+    **kwargs,
+):
+    """
+    Plot the energy distribution of predicted counts of a selection of models assigned to a dataset.
+    Plot the energy distribution of predicted counts of the background and of the sum of all the considered models
+    on the same axis.
+
+    Parameters
+    ----------
+    dataset : an instance of `~gammapy.datasets.dataset`
```
registerrier, line 126:
```suggestion
    dataset : an instance of `~gammapy.datasets.MapDataset`
```

gammapy/visualization/datasets.py
```diff
@@ -0,0 +1,167 @@
...
+def plot_npred_signal(
+    dataset,
+    ax=None,
+    model_names=None,
+    region=None,
+    **kwargs,
+):
+    """
+    Plot the energy distribution of predicted counts of a selection of models assigned to a dataset.
+    Plot the energy distribution of predicted counts of the background and of the sum of all the considered models
```
registerrier, line 121:
```suggestion
   
     The background and the sum of all the considered models predicted counts are plotted on top of individual
     contributions of the considered models. 
```

gammapy/visualization/datasets.py
```diff
@@ -0,0 +1,167 @@
...
+    on the same axis.
```
registerrier, line 122:
```suggestion
```

registerrier (approved):
Thanks @MRegeard . This looks good. I have left minor comments/suggestions.

registerrier (approved):
Thanks @MRegeard . This looks good. I have left minor comments/suggestions.

MRegeard:
@registerrier done !

PR #4412, opened 2023-03-24
LightCurveTemplateModel serialisation

AtreyeeS:
This implements @adonath 's  proposal in #4403

### DIFF
--- a/gammapy/maps/axes.py
+++ b/gammapy/maps/axes.py
@@ -2000,7 +2000,7 @@ class MapAxes(Sequence):
                 except (KeyError, TypeError):
                     try:
                         axis = TimeMapAxis.from_table(table, format=format, idx=idx)
-                    except (KeyError, ValueError):
+                    except (KeyError, ValueError, IndexError):
                         axis = MapAxis.from_table(table, format=format, idx=idx)
 
                 axes.append(axis)
--- a/gammapy/modeling/models/temporal.py
+++ b/gammapy/modeling/models/temporal.py
@@ -525,8 +525,8 @@ class LightCurveTemplateTemporalModel(TemporalModel):
         nodes = table["TIME"]
         if table["TIME"].unit:
             ax_unit = table["TIME"].unit
-        elif table.meta["TIMEUNIT"]:
-            ax_unit = table.meta["TIMEUNIT"]
+        elif table.meta.get("TIMEUNIT", None):
+            ax_unit = table.meta.get("TIMEUNIT", None)
         else:
             raise ValueError("Time unit not found in the table")
         time_axis = MapAxis.from_nodes(nodes=nodes, name="time", unit=ax_unit)
--- a/gammapy/modeling/models/temporal.py
+++ b/gammapy/modeling/models/temporal.py
@@ -524,8 +524,8 @@ class LightCurveTemplateTemporalModel(TemporalModel):
         t_ref = time_ref_from_dict(table.meta, scale="utc")
         nodes = table["TIME"]
      ax_unit = nodes.quantity
-        elif table.meta.get("TIMEUNIT", None):
-            ax_unit = table.meta.get("TIMEUNIT", None)
+    if not ax_unit.is_equivalent("d"):
+        ax_unit = table.meta.get("TIMEUNIT", None)
         else:
             raise ValueError("Time unit not found in the table")
         time_axis = MapAxis.from_nodes(nodes=nodes, name="time", unit=ax_unit)
--- a/gammapy/modeling/models/temporal.py
+++ b/gammapy/modeling/models/temporal.py
@@ -523,8 +523,7 @@ class LightCurveTemplateTemporalModel(TemporalModel):
 
         t_ref = time_ref_from_dict(table.meta, scale="utc")
         nodes = table["TIME"]
-        if table["TIME"].unit:
-            ax_unit = table["TIME"].unit
+     ax_unit = nodes.quantity
         elif table.meta.get("TIMEUNIT", None):
             ax_unit = table.meta.get("TIMEUNIT", None)
         else:
--- a/gammapy/modeling/models/temporal.py
+++ b/gammapy/modeling/models/temporal.py
@@ -528,9 +528,7 @@ class LightCurveTemplateTemporalModel(TemporalModel):
             try:
                 ax_unit = u.Unit(table.meta["TIMEUNIT"])
             except KeyError:
-                ax_unit = None
-        if ax_unit is None:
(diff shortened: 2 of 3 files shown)


gammapy/modeling/models/temporal.py
```diff
@@ -587,7 +596,10 @@ def write(self, filename, format="table", overwrite=False):
             table = self.to_table()
             table.write(filename, overwrite=overwrite)
         elif format == "map":
-            self.map.write(filename, overwrite=overwrite)
+            # RegionNDMap.from_hdulist does not update the header
+            hdulist = self.map.to_hdulist()
+            hdulist[0].header.update(time_ref_to_dict(self.tref_mjd))
```
registerrier, line 601: Same comment here. Don't export the time reference to the primary header. Use the time axis hdu instead.

gammapy/modeling/models/temporal.py
```diff
@@ -517,13 +523,15 @@ def from_table(cls, table, filename=None):
...
         nodes = table["TIME"]
-        time_axis = MapAxis.from_nodes(
-            nodes=nodes, name="time", unit=table.meta["TIMEUNIT"]
-        )
+        if table["TIME"].unit:
+            ax_unit = table["TIME"].unit
+        elif table.meta["TIMEUNIT"]:
+            ax_unit = table.meta["TIMEUNIT"]
+        else:
+            raise ValueError("Time unit not found in the table")
```
registerrier, line 531: This should be tested.

gammapy/modeling/models/tests/test_temporal.py
```diff
@@ -53,8 +53,12 @@ def ph_curve(x, amplitude=0.5, x0=0.01):
 @requires_data()
 def test_light_curve_to_from_table(light_curve):
     table = light_curve.to_table()
-    lc1 = LightCurveTemplateTemporalModel.from_table((table))
+    assert_allclose(table.meta["MJDREFI"], 59000)
+    assert_allclose(table.meta["MJDREFF"], 0.5, rtol=1e-2)
```
registerrier, line 57: why such a large `rtol`? Isn't it an issue of `tt` vs `utc`?

gammapy/modeling/models/temporal.py
```diff
@@ -495,6 +496,11 @@ def __str__(self):
 
         return prnt
 
+    @property
```
registerrier, line 499: If this is used only in tests, this is probably not needed.

registerrier: By the way, here it implicitly assumes the scale to be `utc`.

registerrier (changes requested):
Thanks @AtreyeeS !

The main comment I have is that the reference time keywords should be serialized in the `TIME` axis hdu.

A side question, I think there is an issue in the general time model approach because `t_ref` is handled as a parameter, i.e. as quantity. But its scale is not properly defined. Similarly the scale of the time object used in the `__call__` method is not checked. I think this is a serious issue.

AtreyeeS:
Merci @registerrier ! I fixed the serialisation. The issue about the time scale is indeed more serious, and relevant for analytic temporal models as well. I will try to fix that in a follow-up PR

fabiopintore:
Thanks @AtreyeeS . I don't have any further comment on this PR. It looks ok to me. 
Anyway, I'm also worried about the missing scale of the reference time. This is a quite serious issue that should be fixed soon.

gammapy/modeling/models/temporal.py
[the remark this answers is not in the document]
registerrier: Does the time need to be in mjd?

gammapy/maps/axes.py
```diff
@@ -2000,7 +2000,7 @@ def from_table(cls, table, format="gadf"):
                 except (KeyError, TypeError):
                     try:
                         axis = TimeMapAxis.from_table(table, format=format, idx=idx)
-                    except (KeyError, ValueError):
+                    except (KeyError, ValueError, IndexError):
```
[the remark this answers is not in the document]
AtreyeeS: The `TimeMapAxis.from_table()` expects the table meta to have TIME_MIN and TIME_MAX instead of just TIME.
Thus, it tries to read `colnames[1]`, which does not exist for a node centered normal `MapAxis`. The addition of `IndexError` ensures that `Map.read` tries to read it as a normal `MapAxis`

gammapy/modeling/models/temporal.py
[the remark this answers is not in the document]
AtreyeeS: Yes, I can adapt this in a cleaner manner in the scale PR. Just waiting for this to be merged to avoid merge conflicts

gammapy/modeling/models/temporal.py
```diff
@@ -550,7 +558,8 @@ def read(cls, filename, format="table"):
 
         elif format == "map":
             m = RegionNDMap.read(filename)
-            t_ref = time_ref_from_dict(m.meta)
+            header = fits.getheader(filename, "SKYMAP_BANDS")
```
registerrier, line 561: Here I would suggest to use `hdulist = fits.open(filename)` and then `RegionNDMap.from_hdulist()`

registerrier: But is this necessary? On `read` the `RegionNDMap.meta` is updated no?

registerrier (commented):
Thanks @AtreyeeS . 

See inline comments.

registerrier (dismissed):

gammapy/modeling/models/temporal.py
```diff
@@ -523,9 +523,12 @@ def from_table(cls, table, filename=None):
...
-    if not ax_unit.is_equivalent("d"):
-        ax_unit = table.meta.get("TIMEUNIT", None)
+        ax_unit = nodes.quantity.unit
+        if not ax_unit.is_equivalent("d"):
+            try:
+                ax_unit = u.Unit(table.meta["TIMEUNIT"])
+            except KeyError:
+                ax_unit = None
         if ax_unit is None:
             raise ValueError("Time unit not found in the table")
```
fabiopintore: I suggest to include the `ValueError` directly into the `except KeyError`

fabiopintore (commented):
Thanks @AtreyeeS , I've left only one minor comment. For the rest, it looks good to me!

registerrier (approved):
Thanks @AtreyeeS

registerrier:
After discussion, we realize that this is a bug fix that is part of the scale issue. This should therefore be backported to v1.0.x along with the second PR that will introduce the scale attribute to `TemporalModel` see: #4453 .

PR #4418, opened 2023-03-24
Update the `_evaluate_timevar_source` function in `MapDatasetEventSampler`

fabiopintore:
This PR re-implements the `_evaluate_timevar_source` function in `MapDatasetEventSampler`. The function evaluates a energy-dependent temporal model over a given time and energy range, assuming the a point-like morphology, and returns a npred.
The PR is still in draft. @AtreyeeS @adonath @registerrier @QRemy do you have any comment or advice?

### DIFF
--- a/gammapy/datasets/simulate.py
+++ b/gammapy/datasets/simulate.py
@@ -7,7 +7,7 @@ from astropy.table import Table
 from regions import PointSkyRegion
 import gammapy
 from gammapy.data import EventList, observatory_locations
-from gammapy.maps import MapCoord, RegionNDMap, TimeMapAxis
+from gammapy.maps import MapAxis, MapCoord, RegionNDMap, TimeMapAxis
 from gammapy.modeling.models import (
     ConstantSpectralModel,
     ConstantTemporalModel,
@@ -54,8 +54,9 @@ class MapDatasetEventSampler:
 
         table["TIME"] = (coords["time"] - time_ref).to("s")
         table["ENERGY_TRUE"] = energy
-        table["RA_TRUE"] = coords.skycoord.icrs.ra.to("deg")
-        table["DEC_TRUE"] = coords.skycoord.icrs.dec.to("deg")
+
+        table["RA_TRUE"] = coords.skycoord.icrs.ra.deg
+        table["DEC_TRUE"] = coords.skycoord.icrs.dec.deg
 
         return table
 
@@ -90,15 +91,21 @@ class MapDatasetEventSampler:
             min_timebin = np.min(
                 evaluator.model.temporal_model.map.geom.axes["time"].bin_width
             )
-            time_axis = TimeMapAxis.from_time_bounds(
+            time_axis_eval = TimeMapAxis.from_time_bounds(
                 time_min=tstart,
                 time_max=tstop,
                 nbin=int(((tstop - tstart) / min_timebin).to("")),
             )
+            time_axis = MapAxis.from_bounds(
+                (tstart[0].mjd - dataset.gti.time_ref.mjd) * u.d,
+                (tstop[0].mjd - dataset.gti.time_ref.mjd) * u.d,
+                nbin=int(((tstop.mjd - tstart.mjd) * u.d / min_timebin.to("d")).to("")),
+                name="time",
+            )
 
         flux = (
             evaluator.model.temporal_model.evaluate(
-                time_axis.time_mid, energy=energy_true.center
+                time_axis_eval.time_mid, energy=energy_true.center
             )
             * evaluator.model.spectral_model.parameters[0].quantity
         )
@@ -145,7 +152,22 @@ class MapDatasetEventSampler:
                 raise TypeError(
                     f"Event sampler expects ConstantSpectralModel for a time varying source. Got {evaluator.model.spectral_model} instead."
                 )
-            raise NotImplementedError("The functionality is not yet implemented")
+
+            npred = self._evaluate_timevar_source(dataset, evaluator)
+            data = npred.data[np.isfinite(npred.data)]
+            n_events = self.random_state.poisson(np.sum(data))
+
+            coords = npred.sample_coord(
+                n_events=n_events, random_state=self.random_state
+            )
+
+            coords["time"] = (
+                (coords["time"].value - dataset.gti.time_ref.mjd) * u.d
+            ).to("s")
+
+            table = self._make_table(coords, dataset.gti.time_ref)
+
+        return table
 
         return
 
--- a/gammapy/datasets/simulate.py
+++ b/gammapy/datasets/simulate.py
@@ -128,6 +128,7 @@ class MapDatasetEventSampler:
         pred = (
             (
                 region_exposure.quantity[:, None, :, :]
+                / time_axis.nbin
                 * flux_pred.interp_by_coord(mapcoord)
                 * flux_pred.unit
             )
--- a/gammapy/datasets/simulate.py
+++ b/gammapy/datasets/simulate.py
@@ -118,7 +118,7 @@ class MapDatasetEventSampler:
 
         flux_diff = (
             evaluator.model.temporal_model.evaluate(
-                time_axis_eval.time_mid, energy=energy_new.center, method=self.method
+                time_axis_eval.time_mid, energy=energy_new.center
             )
             * evaluator.model.spectral_model.parameters[0].quantity
         )
--- a/gammapy/datasets/simulate.py
+++ b/gammapy/datasets/simulate.py
@@ -74,7 +74,7 @@ class MapDatasetEventSampler:
 
         Returns
         -------
-        npred : `~RegionNDMap`
+        npred : `~gammapy.maps.RegionNDMap`
             Npred map.
         """
         energy = MapAxis.from_edges(
@@ -87,20 +87,25 @@ class MapDatasetEventSampler:
         if not time_axis:
             tstart = dataset.gti.time_start
             tstop = dataset.gti.time_stop
+            min_timebin = np.min(
+                evaluator.model.temporal_model.map.geom.axes["time"].bin_width
+            )
             # how do we choose the number of bins?
             time_axis = TimeMapAxis.from_time_bounds(
-                time_min=tstart, time_max=tstop, nbin=10
+                time_min=tstart,
+                time_max=tstop,
+                nbin=((tstop - tstart) / min_timebin).to(""),
             )
 
         flux = (
             evaluator.model.temporal_model.evaluate(
                 time_axis.time_mid, energy=energy.center
             )
-            * evaluator.model.spectral_model.parameters[0].unit
+            * evaluator.model.spectral_model.parameters[0].quantity
         )
 
         pred = (
-            (region_exposure.data[:, 0, 0, None] * dataset.exposure.unit * flux)
+            (region_exposure.data[:, 0, 0, None] * flux)
             * np.diff(energy.edges)[:, None]
         ).to("")
 
--- a/gammapy/datasets/simulate.py
+++ b/gammapy/datasets/simulate.py
@@ -1,7 +1,6 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 """Simulate observations"""
 import numpy as np
-from scipy import interpolate
 import astropy.units as u
 from astropy.coordinates import SkyCoord, SkyOffsetFrame
 from astropy.table import Table
@@ -85,9 +84,6 @@ class MapDatasetEventSampler:
         energy_true = dataset.edisp.edisp_map.geom.axes["energy_true"]
         energy_new = energy_true.upsample(10)
         target = evaluator.model.spatial_model.position
-        edisp = dataset.edisp.get_edisp_kernel(
-            dataset.geoms["geom"].axes["energy"], position=target
-        )
         region_exposure = dataset.exposure.to_region_nd_map(target)
 
         if not time_axis:
@@ -115,17 +111,28 @@ class MapDatasetEventSampler:
 
         flux_inte = flux_diff * energy_new.bin_width[:, None]
 
-        f = interpolate.interp1d(energy_new.center, flux_inte, axis=0)
-        flux = f(energy_true.center) * flux_inte.unit
+        flux_pred = RegionNDMap.create(
+            region=PointSkyRegion(center=target),
+            axes=[time_axis, energy_new],
+            data=np.array(flux_inte),
+            unit=flux_inte.unit,
+        )
 
-        pred = ((region_exposure.quantity[:, 0, 0, None] * flux)).to("")
+        mapcoord = flux_pred.geom.get_coord()
+        mapcoord["energy_true"] = energy_true.center[:, None, None, None]
 
-        pred_eres = pred.T @ edisp.quantity
+        pred = (
+            (
+                region_exposure.quantity[:, None, :, :]
+                * flux_pred.interp_by_coord(mapcoord)
+                * flux_pred.unit
+            )
+        ).to("")
 
         npred = RegionNDMap.create(
             region=PointSkyRegion(center=target),
-            axes=[time_axis, dataset.geoms["geom"].axes["energy"]],
-            data=np.array(pred_eres.T),
+            axes=[time_axis, energy_true],
+            data=np.array(pred),
         )
 
         return npred
@@ -234,7 +241,7 @@ class MapDatasetEventSampler:
             else:
                 temporal_model = evaluator.model.temporal_model
 
-            if temporal_model.is_energy_dependent == True:
+            if temporal_model.is_energy_dependent:
                 table = self._sample_coord_time_energy(dataset, evaluator)
             else:
                 flux = evaluator.compute_flux()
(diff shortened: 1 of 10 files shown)


registerrier (commented):
Thanks @fabiopintore

gammapy/datasets/simulate.py
```diff
@@ -70,25 +79,58 @@ def _evaluate_timevar_source(self, dataset, evaluator, time_axis=None):
...
         Returns
         -------
-        npred : `~gammapy.maps.Map`
+        npred : `~gammapy.maps.RegionNDMap`
             Npred map.
         """
-        energy_axis = MapAxis.from_edges(
-            dataset.geoms["geom"].axes["energy"].edges, name="energy_true"
+        energy_true = dataset.edisp.edisp_map.geom.axes["energy_true"]
+        energy_new = energy_true.upsample(10)
```
adonath, line 86: We should avoid the hardcoded number here..

gammapy/datasets/simulate.py
```diff
@@ -70,25 +79,58 @@ def _evaluate_timevar_source(self, dataset, evaluator, time_axis=None):
...
+        edisp = dataset.edisp.get_edisp_kernel(
+            dataset.geoms["geom"].axes["energy"], position=target
         )
+        region_exposure = dataset.exposure.to_region_nd_map(target)
+
+        if not time_axis:
+            tstart = dataset.gti.time_start
+            tstop = dataset.gti.time_stop
+            nbin = int(((tstop - tstart) / t_delta).to(""))
+            time_axis_eval = TimeMapAxis.from_time_bounds(
```
adonath, line 97: This does not seem right to me:
- `tstart` is an array of times, while `.from_time_bounds` would require a single value.
- You can maybe use https://github.com/gammapy/gammapy/blob/main/gammapy/maps/axes.py#L2734 instead

gammapy/datasets/simulate.py
```diff
@@ -70,25 +79,58 @@ def _evaluate_timevar_source(self, dataset, evaluator, time_axis=None):
...
+                time_min=tstart,
+                time_max=tstop,
+                nbin=nbin,
+            )
+            time_axis = MapAxis.from_bounds(
```
adonath, line 102: I think it would be could to introduce some convenience like `MapAxis.from_time_map_axis()` which handles the conversion.

gammapy/datasets/simulate.py
```diff
@@ -70,25 +79,58 @@ def _evaluate_timevar_source(self, dataset, evaluator, time_axis=None):
...
+        flux_diff = (
+            evaluator.model.temporal_model.evaluate(
+                time_axis_eval.time_mid, energy=energy_new.center
+            )
+            * evaluator.model.spectral_model.parameters[0].quantity
+        )
+
+        flux_inte = flux_diff * energy_new.bin_width[:, None]
+
+        f = interpolate.interp1d(energy_new.center, flux_inte, axis=0)
```
adonath, line 118: Please avoid the "manual" interpolation here. You could create the `RegionNDMap` earlier and rely on `RegionNDMap.interpolate_by_coord()` instead.

gammapy/datasets/simulate.py
```diff
@@ -70,25 +79,58 @@ def _evaluate_timevar_source(self, dataset, evaluator, time_axis=None):
...
+        flux = f(energy_true.center) * flux_inte.unit
 
-        flux = evaluator.model.evaluate_geom(region_geom.to_wcs_geom(), gti=dataset.gti)
-        region_exposure = dataset.exposure.to_region_nd_map(region_geom.center_skydir)
+        pred = ((region_exposure.quantity[:, 0, 0, None] * flux)).to("")
 
-        npred = flux * region_exposure * dataset.geoms["geom"].bin_volume()
+        pred_eres = pred.T @ edisp.quantity
```
adonath, line 123: Why is the edisp applied here already? The sampling of the energy dispersions happens later independently, no?

gammapy/datasets/simulate.py
```diff
@@ -179,9 +234,7 @@ def sample_sources(self, dataset):
             else:
                 temporal_model = evaluator.model.temporal_model
 
-            if (
-                temporal_model == "TemplateTemporalModel"
-            ):  # check the condition when the format model is defined
+            if temporal_model.is_energy_dependent == True:
```
adonath, line 237:
```suggestion
            if temporal_model.is_energy_dependent:
```

adonath (changes requested):
Thanks @fabiopintore, I have left some first comments.

gammapy/datasets/simulate.py
```diff
@@ -70,21 +74,41 @@ def _evaluate_timevar_source(self, dataset, evaluator, time_axis=None):
...
+            evaluator.model.temporal_model.evaluate(
+                time_axis.time_mid, energy=energy.center
+            )
+            * evaluator.model.spectral_model.parameters[0].unit
+        )
 
-        npred = flux * region_exposure * dataset.geoms["geom"].bin_volume()
+        pred = (
+            (region_exposure.data[:, 0, 0, None] * dataset.exposure.unit * flux)
+            * np.diff(energy.edges)[:, None]
```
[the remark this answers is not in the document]
adonath: Why is the edisp applied here at all? I thought this happened per event in https://github.com/gammapy/gammapy/blob/main/gammapy/datasets/simulate.py#L232

A dataset can currently not handle time dependent IRFs anyway.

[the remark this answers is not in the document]
fabiopintore: Yes @adonath , good point. The edisp is indeed applied event by event. Hence no need to apply it when sampling the events, it's a mistake.

gammapy/datasets/simulate.py
fabiopintore: I agree, but I'm not sure what's the best way to avoid it. Here we need to oversample the energy to have a precise estimate of `npred`, so how could we do it in a universal approach? Should we left this to the users? Any suggestion @adonath , @registerrier , @AtreyeeS ?

gammapy/datasets/simulate.py
fabiopintore: If I'm not wrong, `from_gti` creates a `TimeMapAxis` with a number of time bins equal to the number of `tstart` and `tstop` of the GTIs. However, this is not what we need in the event sampler... 
`tstart` and `tstop` are arrays of times in general, but the datasets we use for the simulator usually store only one value. Do we expect to simulate from a dataset with gaps in its GTIs? I suppose this functionality is not implemented yet in the event sampler..

gammapy/datasets/simulate.py
fabiopintore: @Atreyee was working on this but she found some issues that made it not very useful. @Atreyee, maybe you can comment on this.

gammapy/datasets/simulate.py
fabiopintore: Done.

gammapy/datasets/simulate.py
fabiopintore: Correct, this is a mistake. I've removed it from the code.

gammapy/datasets/simulate.py
fabiopintore: Done.

gammapy/datasets/simulate.py
adonath: I think we always said that the geometry and energy binning of of the `MapDataset` should be adapted to the required precision. So when in doubt users should choose a fine energy binning. I do not see how the behavior would be different compared to a "normal" spectral model.

fabiopintore: I understand your point and I agree when we talk about normal models. In this case however models are, in some sense, not common. They can be more complex than standard ones and users can do mistakes: for example spectra in different times bins may have very different shapes and it cannot be excluded that a (fine) energy binning is ok for one spectrum but not enough for the other. 
Upsampling the energies is a sort of advice that we give to the user...

gammapy/datasets/simulate.py
AtreyeeS: I think it is just `MapAxis.from_edges(time_map_axis.edges)` which works only if `time_map_axis` is contiguous. And, it loses all information about the `scale` and the reference time, so I dont know if its really very useful

adonath: I agree, it's not very useful for users, it might still improve code readability...

fabiopintore:
I've realized that `npred` is estimated by multiplying the integrated flux (which is defined in terms of energy and time) for the total exposure. This is not correct: the flux in each time bin should be multiplied by the exposure in that time bin. I applied a correction in `_evaluate_timevar_source`:

```
        pred = (
            (
                region_exposure.quantity[:, None, :, :]
                / time_axis.nbin
                * flux_pred.interp_by_coord(mapcoord)
                * flux_pred.unit
            )
        ).to("")
```
the division for `time_axis.nbin` assures that the exposure is calculated for a given time bin. Of course, we are assuming that the bins are linearly spaced.

fabiopintore:
I don't understand why the tests are failing. Locally on my machine, they are all passing!

gammapy/datasets/tests/test_simulate.py
```diff
@@ -126,22 +159,61 @@ def dataset():
...
     sampler = MapDatasetEventSampler(random_state=0)
-    with pytest.raises(NotImplementedError):
-        sampler._sample_coord_time_energy(dataset, evaluator)
+    events = sampler._sample_coord_time_energy(dataset, evaluator)
+
+    assert_allclose(len(events), 918)
+
+    assert_allclose(
+        [events[0][0], events[0][1], events[0][2], events[0][3]],
+        [568.238666, 7.687285, 266.404988, -28.936178],
```
AtreyeeS, line 214: The test fail is on the simulated event energy?

gammapy/datasets/tests/test_simulate.py
```diff
@@ -126,22 +159,61 @@ def dataset():
...
+    assert_allclose(
+        npred.data[:, 1000, 0, 0] / 1e-13,
+        [0.038113, 0.033879, 0.010938],
+        rtol=2e-4,
+    )
+
+
+@requires_data()
+def test_sample_coord_time_energy(dataset, enedip_temporal_model):
+    enedip_temporal_model.spatial_model = None
```
AtreyeeS, line 192: The sampling cannot work if there is no spatial model (ie, it does not assume a point spatial model?)

gammapy/modeling/models/temporal.py
```diff
@@ -493,7 +493,7 @@ class LightCurveTemplateTemporalModel(TemporalModel):
     _t_ref_default = Time("2000-01-01")
     t_ref = Parameter("t_ref", _t_ref_default.mjd, unit="day", frozen=True)
 
-    def __init__(self, map, t_ref=None, filename=None):
+    def __init__(self, map, t_ref=None, filename=None, method=None, values_scale=None):
```
AtreyeeS, line 496: `method` and `values_scale` can be added to the doctoring

gammapy/modeling/models/temporal.py
```diff
@@ -647,7 +658,13 @@ def write(self, filename, format="table", overwrite=False):
             raise ValueError("Not a valid format, choose from ['map', 'table']")
 
     def evaluate(self, time, t_ref=None, energy=None):
-        """Evaluate the model at given coordinates."""
+        """Evaluate the model at given coordinates.
+        time: ~astropy.Time; array of times where the model is
+            evaluate;
```
AtreyeeS, line 663:
```suggestion
            evaluated;
```

gammapy/modeling/models/temporal.py
```diff
@@ -647,7 +658,13 @@ def write(self, filename, format="table", overwrite=False):
...
+        t_ref: ~astropy.Time; reference time. Default is None;
+        energy: ~astropy.Quantity; array of energies where the
+            model is evaluate;
```
AtreyeeS, line 666:
```suggestion
            model is evaluated;
```

gammapy/modeling/models/temporal.py
```diff
@@ -647,7 +658,13 @@ def write(self, filename, format="table", overwrite=False):
             raise ValueError("Not a valid format, choose from ['map', 'table']")
 
     def evaluate(self, time, t_ref=None, energy=None):
-        """Evaluate the model at given coordinates."""
+        """Evaluate the model at given coordinates.
+        time: ~astropy.Time; array of times where the model is
```
AtreyeeS, line 662:
```suggestion
        time: `~astropy.time.Time`
        array of times where the model is
```

gammapy/modeling/models/temporal.py
```diff
@@ -647,7 +658,13 @@ def write(self, filename, format="table", overwrite=False):
...
+            evaluate;
+        t_ref: ~astropy.Time; reference time. Default is None;
```
AtreyeeS, line 664:
```suggestion
        t_ref: `~astropy.time.Time`; reference time. Default is None;
```

AtreyeeS, line 664:
```suggestion
        t_ref: ~astropy.time.Time; reference time. Default is None;
```

gammapy/modeling/models/temporal.py
```diff
@@ -647,7 +658,13 @@ def write(self, filename, format="table", overwrite=False):
...
+        energy: ~astropy.Quantity; array of energies where the
```
AtreyeeS, line 665:
```suggestion
        energy: ~astropy.units.Quantity; array of energies where the
```

gammapy/modeling/models/tests/test_temporal.py
```diff
@@ -55,7 +55,7 @@ def test_energy_dependent_lightcurve(tmp_path):
 
     t = Time(55555.6157407407, format="mjd")
     val = mod.evaluate(t, energy=[0.3, 2] * u.TeV)
-    assert_allclose(val.data, [[2.389591e-21], [4.399548e-23]], rtol=1e-5)
+    assert_allclose(val.data, [[2.278068e-21], [4.280503e-23]], rtol=1e-5)
```
AtreyeeS, line 58: why did this change? because of the interpolation scheme?

AtreyeeS (commented):
Thanks @fabiopintore ! I think looks good

gammapy/datasets/tests/test_simulate.py
fabiopintore: Yes, but locally it doesn't fail. So I don't understand why... any idea of what's going on?

gammapy/datasets/tests/test_simulate.py
fabiopintore: It works only for point-like morphologies. It does not assume a point-like one if the `spatial_model` is `None`: it only raises an error! 
I've also added a test when the morphology is Gaussian.

gammapy/modeling/models/temporal.py
fabiopintore: I've added a note about them in the docstring.

gammapy/modeling/models/temporal.py
fabiopintore: Ok.

gammapy/modeling/models/temporal.py
fabiopintore: Ok.

gammapy/modeling/models/temporal.py
fabiopintore: Ok.

gammapy/modeling/models/temporal.py
fabiopintore: Ok.

fabiopintore: Ok.

gammapy/modeling/models/temporal.py
fabiopintore: Ok.

gammapy/modeling/models/tests/test_temporal.py
fabiopintore: Yes! Previously the interpolation scheme was `method=linear` and `values_scale=linear`. Now it's `method=linear` and `values_scale=log`. This causes the different values in the test.

gammapy/datasets/tests/test_simulate.py
AtreyeeS: Does event sampling work for point like IRFs?

registerrier (dismissed):
Thanks @fabiopintore . I have left some suggestions to clarify the implementation.

AtreyeeS:
@fabiopintore : The test is failing locally on my machine as well, but with different values from the CI! Maybe you can try an assert on some event other than the 0th event?

fabiopintore:
Thansk @AtreyeeS . I've tried to sort the events by time and assert the first element again. If the event sampler is working properly, I think the event time array should be the same on all machine... let's see if this works

adonath:
@fabiopintore I will look into the issue now.

adonath:
@fabiopintore The test you implemented was using a constant model. This was a non-ideal choice, because it does not really test the functionality we are interested in. It also seems that the `InverserCDFSampler` was maybe sensitive to numerical issues when using all constant small values. I have changed know to a real energy dependent temporal model (power law and exponential decay...), which seems to work. Once the CI is all green, I'll merge this PR, however we should check whether the validation in `gammapy-benchmarks` still works correctly. The test coverage in this PR it not yet great.

PR #4419, opened 2023-03-24
Addition of an ExtensionEstimator

registerrier:
**Description**

This pull request proposes an `ExtensionEstimator` based on the parameter estimator.

A possible improvement compared to the typical functionalities of the `ParameterEstimator` would be to handle a case where the extension parameter is kept fixed between all energies and then released to test the energy dependent morphology.

**Dear reviewer**

### DIFF
--- a/gammapy/estimators/extension.py
+++ b/gammapy/estimators/extension.py
@@ -1,0 +1,140 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+import astropy.units as u
+from astropy.coordinates import Angle
+from gammapy.datasets import Datasets
+from gammapy.estimators.parameter import ParameterEstimator
+from gammapy.modeling import Fit
+from gammapy.utils.pbar import progress_bar
+
+
+class ExtensionEstimator(ParameterEstimator):
+    """Source extension estimator.
+
+    Estimates source size for a given list of datasets and a given model.
+
+    The size can be estimated in various energy intervals.
+
+    Parameters
+    ----------
+    energy_edges : `~astropy.units.Quantity`
+        Edges of the energy intervals where the extension is estimated.
+    source : str or int
+        For which source in the model to compute the extension.
+    size_min : float
+        Minimum value for the size used for the fit statistic profile evaluation.
+    size_max : float
+        Maximum value for the size used for the fit statistic profile evaluation.
+    size_n_values : int
+        Number of size values used for the fit statistic profile.
+    size_values : `~astropy.coordinates.Angle`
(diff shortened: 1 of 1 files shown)


gammapy/estimators/extension.py
```diff
@@ -0,0 +1,140 @@
...
+            reoptimize=reoptimize,
+        )
+
+    def run(self, datasets):
+        """Run."""
+        datasets = Datasets(datasets)
+        # find extension parameter
+        model = datasets.models[self.source].spatial_model
+
+        if hasattr(model, "sigma"):
```
adonath, line 91: I think this code block should go into the `estimate_size()` function instead and use the size parameter just as local variable.

adonath:
Thanks @registerrier, this is very interesting and covers an important use case. As this is still a draft I just have a few preliminary comments:

- Depending on the parametrization of the model there is typically a strong correlation between the amplitude and the size parameter. So I think the default behavior here should be to always re-fit the amplitude, when varying the size parameter.
- In general we could think about  having an (optional) `Map`, that stores the size parameter and size scan in `FluxMaps` as well.

bkhelifi:
@registerrier, should you update your branch such that the tests pass?

gammapy/estimators/extension.py
```diff
@@ -0,0 +1,140 @@
...
+
+    The size can be estimated in various energy intervals.
+
+    Parameters
+    ----------
+    energy_edges : `~astropy.units.Quantity`
+        Edges of the energy intervals where the extension is estimated.
+    source : str or int
+        For which source in the model to compute the extension.
+    size_min : float
```
bkhelifi, line 23: Maybe the default values should be added in the docstring...

gammapy/estimators/extension.py
```diff
@@ -0,0 +1,140 @@
...
+        Fit instance specifying the backend and fit options.
+    reoptimize : bool
+        Re-optimize other free model parameters. Default is True.
+    """
+
+    tag = "ExtensionEstimator"
+
+    def __init__(
+        self,
+        energy_edges=[1, 10] * u.TeV,
```
bkhelifi, line 54: Maybe these default values should be removed... to be generic and to force the user to add edges

bkhelifi (commented):
Beyond the few comments on the docstring and initial values, it seems OK to be merged...

bkhelifi:
@registerrier can you update your code from the main?

registerrier:
Closing for now. We need to think more generally how to work with morphology estimation.

PR #4426, opened 2023-03-27
Use datastore maker in Observation.read

QRemy:
Attempt to re-use datastore code for Observation.read

### DIFF
--- a/gammapy/data/observations.py
+++ b/gammapy/data/observations.py
@@ -19,7 +19,7 @@ from gammapy.utils.fits import LazyFitsData, earth_location_to_dict
 from gammapy.utils.scripts import make_path
 from gammapy.utils.testing import Checker
 from gammapy.utils.time import time_ref_to_dict, time_relative_to_ref
-from .event_list import EventList, EventListChecker
+from .event_list import EventListChecker
 from .filters import ObservationFilter
 from .gti import GTI
 from .pointing import FixedPointingInfo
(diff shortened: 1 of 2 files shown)


adonath:
Thanks @QRemy! This looks basically good, except that the overall logic is a bit twisted. I'd be fine to merge this as it reduces code duplication, but I think the we should also try to improve the overall code structure.


I would need to think about it a bit more, but here are some first ideas:

- In a sense I think we should not have the `DataStoreMaker` at all
- We should introduce `Observation.read(..., lazy=True)`
    - This creates `HDULocation` objects instead and triggers the lazy loading of `Observation`.
- Then we could introduce `Observations.read(event_files, irf_files=None, lazy=True)`.
- And introduce `Observations.to_obs_table()` and `Observations.to_hdu_index_table()` (the latter would fail for in-memory observations).
- Then refactor `DataStore.from_event_files()` to rely on `Observations.read()`

Let me know what you think, I can also help with the refactoring.

QRemy:
I just tried this after @maxnoe  [comment](https://github.com/gammapy/gammapy/pull/4280#issuecomment-1483181590) to see if it is possible to not use `load_irf_dict_from_file ` and just rely on the datastore maker code, but this does not work. If I understand correctly we still need workarounds for the hdu names that do not match exactly the format.

QRemy:
Closing as it is broken but we can further discuss this issue in #4995

Issue #4254, opened 2022-12-11
Simulations in the absence of a background model


AtreyeeS:
It can be useful to simulate DL4 observations in the absence of a background model, ie, where the off counts are obtained from real off observations (eg, for pulsars or OnOff Spectral datasets). The observed background here is one realisation of the random variable.

The main point here is to handle the statistics correctly. I dont know if it can be done by simply passing the correct `npred_background` to the `fake` method, or it requires some additional code.

opinions @registerrier @bkhelifi @QRemy @facero

AtreyeeS:
I think the simplest thing to here would be

- Update https://docs.gammapy.org/1.0/tutorials/analysis-1d/spectrum_simulation.html#on-off-analysis to show an example with the MAGIC Crab runs. 
- `dataset_on_off.fake` should take `npred_background=dataset_on_off.background`. The `npred_background` gives the marginalised background estimate, so it changes with the source model.

Or, we can also update https://docs.gammapy.org/1.0/tutorials/analysis-1d/spectral_analysis_rad_max.html to include a section on simulations.

registerrier:
In the absence of a background model, the only possible thing would be to bootstrap on the `counts_off` themselves. For instance, by using `np.random.choice`.

adonath:
Just for completeness I wanted to mention the following approach:

- One could work with 2 datasets: one "on" and one "off" dataset.
- One could define a parametric model for the npred of the  background, e.g. some power law with a turn over at low energies
- The background model would be shared between the on and off dataset, while the source model is only fitted to the on dataset
- This gives an estimation for the background npred model.

AtreyeeS:
@adonath This is indeed interesting, but I think there are still some advantages to using on-off stat in certain cases.
I added a line about your proposal in #4427

adonath:
@AtreyeeS One problem of "wstat" that is often ignored, is the lack of a physical prior assumption / model in the background. The energy bins in the background are assumed to be uncorrelated, which is not true. "wstat" always "overfits" to the background, thus the background model does not generalize to other observations. This problem is apparent for cubes (but for spectra too), where one would end up with a different estimate for the background predicted counts in each pixels. However a real or physical background model should typically behave smoothly. Unfortunately I have never seen anybody discuss this problem, so I'm not aware of any references and magnitude of this effect. I think some assumption about smoothness of the background should be required.

adonath:
[...]

PR #4427, opened 2023-03-27
Add example for on-off simulation

AtreyeeS:
This addresses #4254

### DIFF
--- a/examples/tutorials/analysis-1d/spectral_analysis_rad_max.py
+++ b/examples/tutorials/analysis-1d/spectral_analysis_rad_max.py
@@ -349,4 +349,31 @@ ax.legend()
 ax.set_ylim([1e-13, 1e-10])
 plt.show()
 
+
+######################################################################
+# Dataset simulations
+# -------------------
+#
+# A common way to check is a fit is biased is to simulate multiple datasets with
+# the obtained best fit model, and check the distribution of the fitted parameters.
+# Here, we show how to perform one such simulation
+#
+
+dataset_simulated = datasets.stack_reduce().copy(name="simulated_ds")
+simulated_model = best_fit_model.copy(name="simulated")
+dataset_simulated.models = simulated_model
+dataset_simulated.fake(
+    npred_background=dataset_simulated.counts_off * dataset_simulated.alpha
+)
+plt.figure()
+dataset_simulated.peek()
+plt.show()
+
+# The important thing to note here is that while this samples the on-counts, the off counts are
+# not sampled. If you have multiple measurements of the off counts, they should be used.
+
+result = fit.run(datasets=[dataset_simulated])
+print(result.models.to_parameters_table())
+
+
 # sphinx_gallery_thumbnail_number = 4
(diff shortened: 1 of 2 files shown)


examples/tutorials/analysis-1d/spectral_analysis_rad_max.py
```diff
@@ -349,4 +349,32 @@
...
+# A common way to check is a fit is biased is to simulate multiple datasets with
+# the obtained best fit model, and check the distribution of the fitted parameters.
+# Here, we show how to perform one such simulation
+#
+
+dataset_simulated = datasets.stack_reduce().copy(name="simulated_ds")
+simulated_model = best_fit_model.copy(name="simulated")
+dataset_simulated.models = simulated_model
+dataset_simulated.fake(
+    npred_background=dataset_simulated.counts_off * dataset_simulated.alpha
```
adonath, line 366: This line ignores the data from the in measurement. The predicted background counts in the Wstat case depend on the predict counts for the source. Instead they are given by https://docs.gammapy.org/1.0.1/api/gammapy.stats.get_wstat_mu_bkg.html#gammapy.stats.get_wstat_mu_bkg

However this is only valid for a given source model! We should at least mention this assumption...

QRemy (approved):
Thanks @AtreyeeS that's an interesting point to show in the doc.
In future we could have this in gammapy as a function that run a set of simulations+fit in parallel and return the distribution of fitted parameters. 
Maybe later we could also think of a tutorial dedicated to systematic studies based on simulations but also other methods like Jackknife masking part of the FoV or observations.

examples/tutorials/analysis-1d/spectral_analysis_rad_max.py
```diff
@@ -349,4 +349,32 @@
 ax.set_ylim([1e-13, 1e-10])
 plt.show()
 
+
+######################################################################
+# Dataset simulations
+# -------------------
+#
+# A common way to check is a fit is biased is to simulate multiple datasets with
+# the obtained best fit model, and check the distribution of the fitted parameters.
```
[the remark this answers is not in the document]
AtreyeeS: Indeed that would be more useful. But we show this in the spectral_simulation notebook already, and I dont know how much of stuff we want to repeat..

QRemy (approved):

examples/tutorials/analysis-1d/spectral_analysis_rad_max.py
```diff
@@ -349,4 +349,32 @@
...
 plt.show()
 
+
+######################################################################
+# Dataset simulations
+# -------------------
+#
+# A common way to check if a fit is biased is to simulate multiple datasets with
+# the obtained best fit model, and check the distribution of the fitted parameters.
+# Here, we show how to perform one such simulation
```
registerrier, line 359:
```suggestion
# Here, we show how to perform one such simulation assuming the measured off counts provide a good distribution of the background.
```

registerrier (approved):
Thanks @AtreyeeS . This is good.

I just left a small suggestion to insist on the assumption that off counts provide a correct background distribution.

QRemy (approved):

Issue #4145, opened 2022-10-18
Use negative powers for the representation of units


aaguasca:
Dear developers,

**Context**
Currently, the astropy units are used directly to represent the units of a given axis.

**Describe the solution you'd like**
I propose to use `to_string('latex_inline')` to represent the units of the y-axis with negative powers instead of fractions. Using `.to_string('latex_inline')` may make the unit representation clearer.

registerrier:
Hello @aaguasca . Have you tried to explore the solution you propose?

aaguasca:
Hello @registerrier. Sorry for the delay. Yes, I tried this solution. Please, find below the SED plot from tutorial "analysis-1d/spectral_analysis.ipynb" using `.to_string('latex_inline')`:
![image](https://user-images.githubusercontent.com/82101094/218340935-37f6be81-f233-4ea2-b1b6-b2df1b775641.png)

For this example, I added `.to_string('latex_inline')` to the following lines:
https://github.com/gammapy/gammapy/blob/74c6a76de09ef7c5189ca6e61804469ab96c3fef/gammapy/estimators/points/core.py#L525
https://github.com/gammapy/gammapy/blob/74c6a76de09ef7c5189ca6e61804469ab96c3fef/gammapy/estimators/points/core.py#L617

Adding `.to_string('latex_inline')` should not raise errors because None "units" are prevented, for example by this:
https://github.com/gammapy/gammapy/blob/74c6a76de09ef7c5189ca6e61804469ab96c3fef/gammapy/estimators/points/core.py#L494

If it is OK, I can take care of the PR. I will add `.to_string('latex_inline')` to all relevant axis with expected negative powers.

registerrier:
Hi @aaguasca I suppose it is more efficient to directly use `quantity_support`  from astropy.visualization: 
https://docs.astropy.org/en/stable/api/astropy.visualization.quantity_support.html

aaguasca:
Hi @registerrier, very interesting feature! I suppose using `quantity_support` is necessary to handle the representation of data with different units (but equivalent), but I think we have to define the axis label as well. As far as I know, `quantity_support` only plots the units (with negative powers), but not the name of the variable that has these units. Thus, I think one must specify set_{x,y}label too. Then, when one uses set_{x,y}label, the axis label shows the units without negative powers again.

PR #4428, opened 2023-03-28
Add .to_string() to axis y/xlabel

aaguasca:
**Description**

This pull request addresses Issue #4145. 

The changes consist of adding `.to_string('latex_inline')` to the unit representation in x/y axis labels using the units in ax.xaxis.units, ax.yaxis.units, or directly the units of the object displayed in that axis. In particular, it is added to axes where a power representation is required.

**Dear reviewer**

The PR is ready for review. 

Most of the code where I add `.to_string('latex_inline')` already uses `with quantity_support()`, which prevents any possible error when the ax.yaxis.units or ax.xaxis.units are NoneType. For the case of methods `plot` and `plot_ts_profiles` in "gammapy/estimators/points/core.py", the function variable `DEFAULT_UNIT` prevents NoneType ax.yaxis.units.

### DIFF
--- a/gammapy/irf/psf/map.py
+++ b/gammapy/irf/psf/map.py
@@ -497,8 +497,8 @@ class PSFMap(IRFMap):
                 ax.plot(rad, psf_value, label=label, **kwargs)
 
         ax.set_yscale("log")
-        ax.set_xlabel(f"Rad [{ax.xaxis.units}.to_string(UNIT_STRING_FORMAT)]")
-        ax.set_ylabel(f"PSF [{ax.yaxis.units}.to_string(UNIT_STRING_FORMAT)]")
+        ax.set_xlabel(f"Rad [{ax.xaxis.units.to_string(UNIT_STRING_FORMAT)}]")
+        ax.set_ylabel(f"PSF [{ax.yaxis.units.to_string(UNIT_STRING_FORMAT)}]")
         ax.xaxis.set_major_formatter(FormatStrFormatter("%.2f"))
         plt.legend()
         return ax
--- a/gammapy/irf/psf/map.py
+++ b/gammapy/irf/psf/map.py
@@ -5,6 +5,7 @@ from astropy.visualization import quantity_support
 import matplotlib.pyplot as plt
 from matplotlib.ticker import FormatStrFormatter
 from gammapy.maps import Map, MapAxes, MapAxis, MapCoord, WcsGeom
+from gammapy.maps.axes import UNIT_STRING_FORMAT
 from gammapy.modeling.models import PowerLawSpectralModel
 from gammapy.utils.gauss import Gauss2DPDF
 from gammapy.utils.random import InverseCDFSampler, get_random_state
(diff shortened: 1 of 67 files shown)


registerrier:
Thanks @aaguasca !
Apparently there is a formatting issue in the CI. Do you use the correct settings for the pre-commit? What version of black do you have?

adonath:
Thanks @aaguasca, this looks mostly good! I think it is still a good idea to introduce a global variable e.g. here https://github.com/gammapy/gammapy/blob/main/gammapy/maps/axes.py#L71

Like so:
```python3

UNIT_STRING_FORMAT = "latex_inline"

ax.set_ylabel(f"Sensitivity [{t['e2dnde'].unit.to_string(UNIT_STRING_FORMAT)}]")
```
Such that it could be potentially modified by users, if needed.

adonath:
On second thought I think we should use is consistently (not only for powers...). There potentially a few more places to be adapted https://github.com/gammapy/gammapy/search?q=set_ylabel

registerrier:
Hi @aaguasca , there is a merge conflict here because the tests you modify were adapted in #4429 because they were breaking with future versions of astropy and matplotlib. The order of the units in the label will change in the future, so we no longer test the unit in the label string, but simply that the unit seen by matplotlib is correct.
So the modifications in test_spectral should be reverted and rebased. Then maybe simply test that the unit part of the label string is latex?

aaguasca:
Hi @registerrier. Thank you for the clarification. I will revert the changes in the test test_spectral. 

I created a PR to fix another problem with the tests (PR #4446). Once it is merged, I will push all commits. Please, let me know if I should modify something in PR #4446.

> Then maybe simply test that the unit part of the label string is latex?

Maybe if we know the convention of the unit representation, for example, within brackets, checking the default string format representation of the units should be relatively easy using split.

registerrier (commented):
Thanks @aaguasca . This looks good.

Shouldn't all `MapAxis` have a `Unit` defined? Even `LabelMapAxis` does.
For `rad`, the issue is the content of the `IRF`, but here it should have a unit too, no?

aaguasca:
I think the problem is `psf_value`. The returned array by the method `self.psf_map.interp_by_coord` has no units because `self.psf_map` is a `RegionNDMap` object.

Maybe a workaround would be to multiply the output `psf_value` by `self.psf_map.unit`. What do you think the best solution is?

gammapy/irf/psf/map.py
```diff
@@ -493,9 +497,10 @@ def plot_psf_vs_rad(self, ax=None, energy_true=[0.1, 1, 10] * u.TeV, **kwargs):
             with quantity_support():
                 ax.plot(rad, psf_value, label=label, **kwargs)
 
+        format_NoneType_axes_units(ax)
```
registerrier, line 500: yes here it should  be sufficient to do `psf_value *= self.psf_map.unit`

registerrier (approved):
Thansk @aaguasca . This looks good. Indeed, you could multiply here the `psf_value` by the `psf_map` unit. 
`interp_by_coord` returns an array...

registerrier (approved):
Thanks @aaguasca . No further comment.

PR #4429, opened 2023-03-28
Fix CI fail in test_spectrum

AtreyeeS:
This should fix the CI break which checks the string instead of the unit

### DIFF
--- a/gammapy/makers/tests/test_spectrum.py
+++ b/gammapy/makers/tests/test_spectrum.py
@@ -223,7 +223,7 @@ def test_safe_mask_maker_dc1(spectrum_dataset_gc, observations_cta_dc1):
     maker = SpectrumDatasetMaker()
     dataset = maker.run(spectrum_dataset_gc, obs)
     dataset = safe_mask_maker.run(dataset, obs)
-    assert_allclose(dataset.energy_range[0], 1, rtol=1e-3)
+    assert_allclose(dataset.energy_range[0].data, 1.0, rtol=1e-3)
     assert dataset.energy_range[0].unit == "TeV"
 
 
--- a/gammapy/makers/tests/test_spectrum.py
+++ b/gammapy/makers/tests/test_spectrum.py
@@ -217,9 +217,9 @@ def test_safe_mask_maker_dl3(spectrum_dataset_crab, observations_hess_dl3):
 
 @requires_data()
 def test_safe_mask_maker_dc1(spectrum_dataset_gc, observations_cta_dc1):
-    safe_mask_maker = SafeMaskMaker(methods=["edisp-bias", "aeff-max"])
+    safe_mask_maker = SafeMaskMaker(methods=["aeff-max"])
     empty = SpectrumDataset.from_geoms(**spectrum_dataset_gc.geoms)
-    obs = observations_cta_dc1[0]
+    obs = observations_cta_dc1[1]
     maker = SpectrumDatasetMaker()
     dataset = maker.run(empty, obs)
     dataset = safe_mask_maker.run(dataset, obs)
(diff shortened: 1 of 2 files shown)


registerrier (dismissed):
Thanks @AtreyeeS ! This is important to solve these issues.

No comment from my side. 
Just one question, what is the issue with  `dataset.energy_range` in test_spectrum.py?

AtreyeeS:
I can't reproduce the error locally, so its difficult to debug. The CI builds mask the first bin, 3.16 is the starting energy of the next bin..

AtreyeeS:
Ok apparently the problem was with the `edisp` option on the safe mask maker. Still dont know why!

adonath:
> Ok apparently the problem was with the edisp option on the safe mask maker. Still dont know why!

I think this is a sparse fail. Right no it is all green.

gammapy/makers/tests/test_spectrum.py
```diff
@@ -217,13 +217,13 @@ def test_safe_mask_maker_dl3(spectrum_dataset_crab, observations_hess_dl3):
 
 @requires_data()
 def test_safe_mask_maker_dc1(spectrum_dataset_gc, observations_cta_dc1):
-    safe_mask_maker = SafeMaskMaker(methods=["edisp-bias", "aeff-max"])
```
AtreyeeS, line 220: @adonath Its green because I removed the edisp-bias here

gammapy/modeling/models/tests/test_spectral.py
```diff
@@ -437,26 +437,34 @@ def test_model_plot_sed_type():
...
+        assert ax1.axes.axes.get_ylabel().split()[0] == "dnde"
+        assert ax2.yaxis.units.is_equivalent(u.Unit("1 / (s cm2 TeV)"))
+        assert ax2.axes.axes.get_ylabel().split()[0] == "dnde"
 
     with mpl_plot_check():
         ax1 = pwl.plot((1 * u.TeV, 100 * u.TeV), sed_type="e2dnde")
         ax2 = pwl.plot_error((1 * u.TeV, 100 * u.TeV), sed_type="e2dnde")
-        assert ax1.axes.axes.get_ylabel() == "e2dnde [erg / (cm2 s)]"
-        assert ax2.axes.axes.get_ylabel() == "e2dnde [erg / (cm2 s)]"
+        assert ax1.yaxis.units.is_equivalent(u.Unit("erg / (cm2 s)"))
```
adonath, line 448: I remember I had some discussions with Christoph in the past on how to test something like this. In the end he convinced me that one should test, that the unit does not change at all. E.g. in our code (for what ever reason...) the units could change to `"TeV / (m2 s)` and the test would pass.

So my proposal here would be to just cover both cases:
```python3
label = ax2.axes.axes.get_ylabel()
assert label in ["e2dnde [erg / (cm2 s)]", "..." ]
```

registerrier (dismissed):
Thanks @AtreyeeS .

Removing the failing test is of course a possibility. It sounds fine to me. I would just suggest to open an issue to create a test for this `edisp-bias` option.

gammapy/modeling/models/tests/test_spectral.py
adonath: > Do you know if we can check if units are same, ie, not equivalent?

Just comparing the units seems to work as expected:

```python

a = u.Unit("erg / (cm2 s)")
b = u.Unit("erg / (s cm2)")
c = u.Unit("TeV / (s cm2)")

print(a == b)
print(a == c)
```

gammapy/makers/tests/test_spectrum.py
adonath: That's fine be me as well. However we should maybe make a reminder issue to investigate at some point why it failed...

registerrier (approved):
Thanks @AtreyeeS !

PR #4430, opened 2023-03-28
Add squash method to LabelMapAxis

MRegeard:
This PR is related to PR #4409. 

I added a squash method the `LabelMapAxis`in order to be able to use `sum_over_axes` on a `LabelMapAxis`. 

For the label of the squashed axis, I currently put 'squash_' + the name of the axis. But if anyone has a better, more explicit proposition for this, let me know.

### DIFF
--- a/gammapy/maps/axes.py
+++ b/gammapy/maps/axes.py
@@ -3196,15 +3196,20 @@ class LabelMapAxis:
 
         return LabelMapAxis(merged_labels, self.name)
 
-    def squash(self):
+    def squash(self, label=[""]):
         """Create a new axis object by squashing the axis into one bin.
 
-        The label of the new axis is given as "squash_" + the name of the axis.
+        Parameters
+        ----------
+        label : list of str
+            The label of the new axis. If label is not given, the label will be set to "first_label...last_label"
+            of the axis.
 
         Returns
         -------
         axis : `~MapAxis`
             Sliced axis object.
         """
-
-        return LabelMapAxis(labels=["squash_" + self._name], name=self._name)
+        if label == [""]:
+            label = [self.center[0] + "..." + self.center[-1]]
+        return LabelMapAxis(labels=label, name=self._name)
--- a/gammapy/maps/axes.py
+++ b/gammapy/maps/axes.py
@@ -3195,3 +3195,15 @@ class LabelMapAxis:
         merged_labels = np.append(self.center, axis.center)
 
         return LabelMapAxis(merged_labels, self.name)
+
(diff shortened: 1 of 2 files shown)


adonath:
Thanks a lot @MRegeard! So far we do not have any convention how the labels are handled when summed or resampled (see e.g. https://github.com/gammapy/gammapy/blob/main/gammapy/maps/axes.py#L3118) So it is open to us to decide!

I can imagine 2-3 possibilities:

- Merge all the labels into a single one using a string separator. For example `["a", "b", "c"]`, becomes `"a-b-c"` or `"a,b,c"`. This has the advantage of keeping some meta information on the former labels.
- Just use a new label by convention, as you are proposing in this PR
- Combine one of the approaches above with the optional possibility to overwrite the label, like `LabelMapAxis.squash(label="my-label")` 

Whatever we decide we should make sure that the behavior could be later consistently introduced in e.g. `LabelMapAxis.resample()` as well. 

Do you have any preference @registerrier, @AtreyeeS or @QRemy?

adonath:
Personally I'm leaning towards the solution of merging the labels using a string separator, because it keeps the original information. But this could lead to potentially very long labels...

MRegeard:
> Personally I'm leaning towards the solution of merging the labels using a string separator, because it keeps the original information. But this could lead to potentially very long labels...

Yes I thought about that but it could indeed be really long and therefore not that readable.

AtreyeeS:
Having a very large name for a MapAxis is probably not a good idea. I would prefer a default name like in this PR, with a possibility to overwrite as @adonath suggested. 

It would be a good idea to come up with a meta handling for Maps, and I suppose the individual label information should go in there. Even in the case of a normal `MapAxis.squash`, we lose information on the number of bins squashed.

gammapy/maps/axes.py
```diff
@@ -3195,3 +3195,15 @@ def append(self, axis):
         merged_labels = np.append(self.center, axis.center)
 
         return LabelMapAxis(merged_labels, self.name)
+
+    def squash(self):
+        """Create a new axis object by squashing the axis into one bin.
```
registerrier, line 3200:
```suggestion
        """Create a new axis object by squashing the axis into one bin.
        
```

registerrier (dismissed):

I am also not very confident in appending all the labels together as the length of the string can become extremely long. How about "first...last" ?

adonath:
Thanks @MRegeard, now that we have agreed on a label, we maybe just remove the possibility to modify it? Typically convention is better then configuration...

Issue #4230, opened 2022-12-02
FluxPoints serialisation loses GTI information


AtreyeeS:
**Gammapy version**
1.0

`FluxPoints.write()` does not serialise the GTI information.
For example, in the light curve flare tutorial, 
`lc_1d.gti`  exists.
But,

```
lc_1d.write("lightcurve_pks2155.fits", format="lightcurve", overwrite=True)
lightcurve = FluxPoints.read("lightcurve_pks2155.fits", format="lightcurve")
print(light curve.gti)
```
gives `None`.

This is rather problematic as the light curve cannot be converted into `FluxPointsDataset`s for modelling and fitting.

AtreyeeS:
Some of the meta information like `n_sigma` and `n_sigma_ul` is lost as well.
Only the table is serialised

adonath:
Thanks @AtreyeeS! The sad truth is that we do not have any official or documented data format for light curves that allows for this. As long as the definition of data formats is behind, the serialization in Gammapy will be incomplete. We can start to introduce and document our own Gammapy formats, but then we might do the work twice. I'll change the label from bug to feature request, because I think right now we do not claim anywhere that out serialization is complete...

However users should be able serialize the gti independently using e.g. `GTI.write()`?

AtreyeeS:
> However users should be able serialize the gti independently using e.g. `GTI.write()`?

Yes indeed @adonath.  So the minimal thing to do would be to mention this explicitly in our tutorials, and also show how to explicitly attach a GTI to a FluxPoints object. As far as I remember, GTI handling is not exposed very well in our tutorials, because it is mostly handled by gammapy and not left to the external user.

registerrier:
Some remarks here:
- indeed `GTI` should be serialized along with `FluxPoints` for provenance
- For lightcurve modeling, time intervals of the lightcurve should rather be used, no? The issue rather lies with the current choice of evaluating time from `GTI` rather than from the dataset time geometry.

PR #4432, opened 2023-03-28
Serialise gti table to flux points object

AtreyeeS:
This is intended to address #4230 

There is a test fail here with the ecsv format that I am not sure how to handle.
I am serialising the gti table as an additional hdu, but that does not work for non-fits formats.

The issue, I guess, is that FluxPoints format is not very well specified in the GADF.
We can decide to serialise the GTI only if the output is a fits format (any quick way to check @adonath ?)
Or, we can write the GTI table separately and have a yaml file connecting the FluxPoints and the GTI.
Suggestions please @adonath @registerrier @bkhelifi

### DIFF
--- a/gammapy/estimators/points/core.py
+++ b/gammapy/estimators/points/core.py
@@ -7,7 +7,6 @@ from astropy.io import fits
 from astropy.io.registry import IORegistryError
 from astropy.table import Table, vstack
 from astropy.time import Time
-from astropy.utils.exceptions import AstropyDeprecationWarning
 from astropy.visualization import quantity_support
 import matplotlib.pyplot as plt
 from gammapy.data import GTI
@@ -135,17 +134,20 @@ class FluxPoints(FluxMaps):
             Flux points
         """
         filename = make_path(filename)
-
+        gti = None
+        kwargs.setdefault("format", "ascii.ecsv")
         try:
             table = Table.read(filename, **kwargs)
-        except IORegistryError:
-            kwargs.setdefault("format", "ascii.ecsv")
-            table = Table.read(filename, **kwargs)
-
-        try:
-            gti = GTI.read(filename)
-        except (TypeError, AstropyDeprecationWarning):
-            gti = None
+        except (IORegistryError, UnicodeDecodeError):
+            with fits.open(filename) as hdulist:
+                hdu_names = [_.name for _ in hdulist]
+                if "FLUXPOINTS" in hdu_names:
+                    fp = hdulist["FLUXPOINTS"]
+                else:
+                    fp = hdulist[""]  # to handle older files
+                table = Table.read(fp)
+                if "GTI" in hdu_names:
+                    gti = GTI.read(filename)
 
         return cls.from_table(
             table=table,
@@ -180,13 +182,18 @@ class FluxPoints(FluxMaps):
         overwrite : bool
             Overwrite existing file`.
         """
+        filename = make_path(filename)
         if sed_type is None:
             sed_type = self.sed_type_init
-
-        filename = make_path(filename)
-        primary_hdu = fits.PrimaryHDU()
         table = self.to_table(sed_type=sed_type, format=format)
-        hdu_evt = fits.BinTableHDU(table, name="FLUXPOINT")
+
+        # TODO: rather ugly - better method?
+        if str(filename).split(".")[-1] != "fits":
+            table.write(filename)
+            return
+
+        primary_hdu = fits.PrimaryHDU()
+        hdu_evt = fits.BinTableHDU(table, name="FLUXPOINTS")
         hdu_all = fits.HDUList([primary_hdu, hdu_evt])
         if self.gti:
             hdu_all.append(self.gti.to_table_hdu())
--- a/gammapy/estimators/points/core.py
+++ b/gammapy/estimators/points/core.py
@@ -187,7 +187,7 @@ class FluxPoints(FluxMaps):
         table = self.to_table(sed_type=sed_type, format=format)
 
         # TODO: rather ugly - better method?
-        if not ".fits" in filename.suffixes:
+        if ".fits" not in filename.suffixes:
             table.write(filename)
             return
 
--- a/gammapy/estimators/points/core.py
+++ b/gammapy/estimators/points/core.py
@@ -7,6 +7,7 @@ from astropy.io import fits
 from astropy.io.registry import IORegistryError
 from astropy.table import Table, vstack
 from astropy.time import Time
+from astropy.utils.exceptions import AstropyDeprecationWarning
 from astropy.visualization import quantity_support
 import matplotlib.pyplot as plt
 from gammapy.data import GTI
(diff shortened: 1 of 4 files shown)


gammapy/estimators/points/core.py
```diff
@@ -139,14 +142,20 @@ def read(
             kwargs.setdefault("format", "ascii.ecsv")
             table = Table.read(filename, **kwargs)
 
+        try:
+            gti = GTI.read(filename)
+        except (TypeError, AstropyDeprecationWarning):
```
registerrier, line 147: Why catching the `AstropyDeprecationWarning`?
This might be problematic because we need to see those warnings to remove the deprecated functionality.

gammapy/estimators/points/core.py
```diff
@@ -139,14 +142,20 @@ def read(
             kwargs.setdefault("format", "ascii.ecsv")
             table = Table.read(filename, **kwargs)
 
+        try:
+            gti = GTI.read(filename)
```
registerrier, line 146: Also you could rather open the `HDUList` and then read the `Table` and `GTI` at their correct locations. 
Of course one has to be careful maintaining backward compatibility with files having no extname defined.

registerrier (commented):
Thanks @Atreyee. 
See inline comments.

gammapy/estimators/points/core.py
AtreyeeS: If the required HDU is not present, ie, GTI, then it tries to read the first HDU. This leads to a warning and a OSError much deeper down. Catching the OSError did not work.

bkhelifi:
In the event FITS file, the GTI are in a separated HDU. Is it also the case here?

AtreyeeS:
Thanks for the comments.

The `FluxPoints` serialisation now works as follows
- If the format is `BinnedTimeSeries`, the `GTI` table is NOT serialised
- For the formats `lightcurve` and `profile`, the FluxPoints and GTIs are serialised as two separate HDUs in the same file, named `FLUXPOINTS` and `GTI`, resp. To support files created with prev versions, HDUs without names are read as flux points.
- For format `gadf-sed`, the above scheme is followed for fits files. For other formats, eg: ecsv, the serialisation works as an satrapy table and the GTI table is not serialised.

bkhelifi:
Thanks a lot @AtreyeeS! 
Based on your explanations and as the GADF format for the DL5 is not totally set up, I am wondering if one should not expend a bit the docstring, or the rst/doc page about the format (and maybe make an interlink). What do you think @registerrier , @adonath and @maxnoe ?

AtreyeeS:
I realised that the GTI table will be necessary event after https://github.com/gammapy/gammapy/issues/4480 is implemented - we would still need information on when the source was observed between the given time bins.

The light curve has a `time_axis` defining the bins for the light curve extraction (say, daily bins). We also need to account for the source observation duration each day, which is contained in the GTIs.

Maybe you can do a (re-)review @registerrier @adonath ?

bkhelifi:
Hi @AtreyeeS ,
You have added an exception in the read function. Is there a test for that?

gammapy/estimators/points/core.py
```diff
@@ -168,15 +179,26 @@ def write(self, filename, sed_type=None, format="gadf-sed", **kwargs):
             * "profile": Gammapy internal format to store energy dependent
                 flux profiles. Basically a generalisation of the "gadf" format, but
                 currently there is no detailed documentation available.
-        **kwargs : dict
-            Keyword arguments passed to `astropy.table.Table.write`.
+        overwrite : bool
+            Overwrite existing file`.
```
bkhelifi, line 183:
```suggestion
            Overwrite existing file.
```

gammapy/estimators/points/core.py
```diff
@@ -168,15 +179,26 @@ def write(self, filename, sed_type=None, format="gadf-sed", **kwargs):
...
+        filename = make_path(filename)
         if sed_type is None:
             sed_type = self.sed_type_init
-
-        filename = make_path(filename)
         table = self.to_table(sed_type=sed_type, format=format)
-        table.write(filename, **kwargs)
+
+        # TODO: rather ugly - better method?
+        if str(filename).split(".")[-1] != "fits":
```
bkhelifi, line 191: Maybe this: 
`if ".fits" in str(filename):`

gammapy/estimators/points/core.py
```diff
@@ -132,21 +134,30 @@ def read(
...
+        except (IORegistryError, UnicodeDecodeError):
+            with fits.open(filename) as hdulist:
+                hdu_names = [_.name for _ in hdulist]
+                if "FLUXPOINTS" in hdu_names:
+                    fp = hdulist["FLUXPOINTS"]
+                else:
+                    fp = hdulist[""]  # to handle older files
+                table = Table.read(fp)
+                if "GTI" in hdu_names:
+                    gti = GTI.read(filename)
```
registerrier, line 150: Here you could avoid reopening the file:
```
if "GTI" in hdulist:
    gti = GTI.from_table_hdu(hdulist["GTI"])
```

gammapy/estimators/points/core.py
registerrier, line 191:
```suggestion
        if not ".fits" in filename.suffixes:
```

registerrier (approved):
Thanks @AtreyeeS . This looks good. I have left a few minor inline comments.

gammapy/estimators/points/core.py
maxnoe:
```
".fits" in Path(filename).suffixes
```

gammapy/estimators/points/core.py
```diff
@@ -132,21 +134,30 @@ def read(
...
+        kwargs.setdefault("format", "ascii.ecsv")
         try:
             table = Table.read(filename, **kwargs)
-        except IORegistryError:
-            kwargs.setdefault("format", "ascii.ecsv")
-            table = Table.read(filename, **kwargs)
+        except (IORegistryError, UnicodeDecodeError):
+            with fits.open(filename) as hdulist:
+                hdu_names = [_.name for _ in hdulist]
+                if "FLUXPOINTS" in hdu_names:
```
registerrier, line 144:
```suggestion
                if "FLUXPOINTS" in hdulist:
```

gammapy/estimators/points/core.py
```diff
@@ -168,15 +178,26 @@ def write(self, filename, sed_type=None, format="gadf-sed", **kwargs):
...
+        filename = make_path(filename)
         if sed_type is None:
             sed_type = self.sed_type_init
-
-        filename = make_path(filename)
         table = self.to_table(sed_type=sed_type, format=format)
-        table.write(filename, **kwargs)
+
+        # TODO: rather ugly - better method?
+        if not ".fits" in filename.suffixes:
```
registerrier, line 190:
```suggestion
        if ".fits" not in filename.suffixes:
```

registerrier:

PR #4414, opened 2023-03-24
Expose more methods in ExcessMapEstimator

QRemy:
Expose methods to compute kernel, mask, and reco_exposure in ExcessMapEstimator. This will be needed in an other PR to add support for estimation over datasets.

PR #4433, opened 2023-03-28
Add a function to combine excess maps

QRemy:
Computes correlated excess and significance for a set of datasets.
The significance computation assumes that the model contains
 one degree of freedom per valid bin in energy in each dataset.
 
 To add an estimate for the flux #4414 is needed.

### DIFF
--- a/gammapy/estimators/map/excess.py
+++ b/gammapy/estimators/map/excess.py
@@ -129,6 +129,11 @@ class ExcessMapEstimator(Estimator):
     spectral_model : `~gammapy.modeling.models.SpectralModel`
         Spectral model used for the computation of the flux map.
         If None, a Power Law of index 2 is assumed (default).
+    sum_over_energy_groups : bool
+        Only used if energy_edges is None.
+        If False apply the estimator in each energy bin of the parent dataset.
+        If True apply the estimator in only one bin defined by the energy edges of the parent dataset.
+        Default is False.
 
     Examples
     --------
@@ -169,6 +174,7 @@ class ExcessMapEstimator(Estimator):
         gamma_min_sensitivity=10,
         bkg_syst_fraction_sensitivity=0.05,
         apply_threshold_sensitivity=False,
+        sum_over_energy_groups=False,
     ):
         self.correlation_radius = correlation_radius
         self.n_sigma = n_sigma
@@ -179,6 +185,7 @@ class ExcessMapEstimator(Estimator):
         self.apply_threshold_sensitivity = apply_threshold_sensitivity
         self.selection_optional = selection_optional
         self.energy_edges = energy_edges
+        self.sum_over_energy_groups = sum_over_energy_groups
         self.correlate_off = correlate_off
 
         if spectral_model is None:
--- a/gammapy/estimators/map/excess.py
+++ b/gammapy/estimators/map/excess.py
@@ -8,12 +8,11 @@ from gammapy.datasets import MapDataset, MapDatasetOnOff
 from gammapy.maps import Map
 from gammapy.modeling.models import PowerLawSpectralModel, SkyModel
 from gammapy.stats import CashCountsStatistic, WStatCountsStatistic
-from gammapy.stats.utils import ts_to_sigma
 from ..core import Estimator
 from ..utils import estimate_exposure_reco_energy
 from .core import FluxMaps
 
-__all__ = ["ExcessMapEstimator", "get_joint_excess_maps"]
+__all__ = ["ExcessMapEstimator"]
 
 log = logging.getLogger(__name__)
 
@@ -75,45 +74,6 @@ def convolved_map_dataset_counts_statistics(dataset, kernel, mask, correlate_off
         npred = dataset.npred() * mask
         background_conv = npred.convolve(kernel_data)
         return CashCountsStatistic(n_on_conv.data, background_conv.data)
-
-
-def get_joint_excess_maps(estimator, datasets):
-    """Computes correlated excess and significance for a set of datasets.
-    The significance computation assumes that the model contains
-    one degree of freedom per valid bin in energy in each dataset.
-
-    Parameters
-    ----------
-    estimator : `~gammapy.estimator.ExcessMapEstimator`
-        Excess Map Estimator
-    dataset : `~gammapy.datasets.Datasets`
-        Datasets
-
-    """
-
-    geom = datasets[0].counts.geom.to_image()
-    for dataset in datasets[1:]:
-        if dataset.counts.geom.to_image() != geom:
-            raise ValueError
-
-    ts_sum = Map.from_geom(geom)
-    npred_excess_sum = Map.from_geom(geom)
-    significance = Map.from_geom(geom)
-
-    dof = 0
-    for dataset in datasets:
-        results = estimator.run(dataset)
-
-        dof += np.sum(results["ts"].data > 0, axis=0)
-        ts = results["ts"].reduce_over_axes()
-
-        npred_excess = results["npred_excess"].reduce_over_axes()
-
-        ts_sum += ts
-        npred_excess_sum += npred_excess
-
-    significance.data = ts_to_sigma(ts_sum.data, dof) * np.sign(npred_excess_sum.data)
-    return dict(ts=ts_sum, npred_excess=npred_excess_sum, significance=significance)
 
 
 class ExcessMapEstimator(Estimator):
--- a/gammapy/estimators/map/excess.py
+++ b/gammapy/estimators/map/excess.py
@@ -8,13 +8,12 @@ from gammapy.datasets import MapDataset, MapDatasetOnOff
 from gammapy.maps import Map
 from gammapy.modeling.models import PowerLawSpectralModel, SkyModel
 from gammapy.stats import CashCountsStatistic, WStatCountsStatistic
+from gammapy.stats.utils import ts_to_sigma
 from ..core import Estimator
 from ..utils import estimate_exposure_reco_energy
 from .core import FluxMaps
 
-__all__ = [
-    "ExcessMapEstimator",
-]
+__all__ = ["ExcessMapEstimator", "get_joint_excess_maps"]
 
 log = logging.getLogger(__name__)
 
@@ -76,6 +75,45 @@ def convolved_map_dataset_counts_statistics(dataset, kernel, mask, correlate_off
         npred = dataset.npred() * mask
         background_conv = npred.convolve(kernel_data)
         return CashCountsStatistic(n_on_conv.data, background_conv.data)
+
+
+def get_joint_excess_maps(estimator, datasets):
+    """Computes correlated excess and significance for a set of datasets.
+    The significance computation assumes that the model contains
+    one degree of freedom per valid bin in energy in each dataset.
+
+    Parameters
+    ----------
+    estimator : `~gammapy.estimator.ExcessMapEstimator`
+        Excess Map Estimator
+    dataset : `~gammapy.datasets.Datasets`
+        Datasets
+
+    """
+
+    geom = datasets[0].counts.geom.to_image()
+    for dataset in datasets[1:]:
+        if dataset.counts.geom.to_image() != geom:
+            raise ValueError
+
+    ts_sum = Map.from_geom(geom)
+    npred_excess_sum = Map.from_geom(geom)
+    significance = Map.from_geom(geom)
+
+    dof = 0
+    for dataset in datasets:
+        results = estimator.run(dataset)
+
+        dof += np.sum(results["ts"].data > 0, axis=0)
+        ts = results["ts"].reduce_over_axes()
+
+        npred_excess = results["npred_excess"].reduce_over_axes()
+
+        ts_sum += ts
+        npred_excess_sum += npred_excess
+
+    significance.data = ts_to_sigma(ts_sum.data, dof) * np.sign(npred_excess_sum.data)
+    return dict(ts=ts_sum, npred_excess=npred_excess_sum, significance=significance)
 
 
 class ExcessMapEstimator(Estimator):
--- a/gammapy/estimators/utils.py
+++ b/gammapy/estimators/utils.py
@@ -706,17 +706,25 @@ def get_joint_significance_maps(
 
     Parameters
     ----------
-    estimator : `~gammapy.estimator.ExcessMapEstimator`
-        Excess Map Estimator
+    estimator : `~gammapy.estimator.ExcessMapEstimator` or `~gammapy.estimator.TSMapEstimator`
+        Excess Map Estimator or TS Map Estimator
     dataset : `~gammapy.datasets.Datasets`
-        Datasets
+        Datasets containing only `~gammapy.maps.MapDataset`.
     energy_edges : list of `~astropy.units.Quantity`, optional
         Edges of the target maps energy bins. The resulting bin edges won't be exactly equal to the input ones,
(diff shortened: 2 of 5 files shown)


adonath:
Thanks @QRemy!

I have some concerns of the validity of this method. Poisson statistics behaves non-linear and TS values can typically not just be summed. I have done some literature  research in the past and I'm not aware of any reference that would describe this method and it's related approximations. I'll invite @registerrier and @giacomodamico24 to comment as well. 

Especially in the low counts regime, such a method will never provide correct errors and upper limits, I think this makes it basically useless. One might gain sensitivity by combining the results, but you loose sensitivity by assuming linearity in the TS value. It is completely unclear which benefit the method should bring, compared to stacking.

I think the correct ways to compute such a map are:
- Optimize the norm in a joint-likelihood fit, basically a "spectral fit" per pixel. Which directly computes norm, errors and ULs
- Let the `ExcessMapEstimator` compute likelihood profiles, combine those and find the new norm, errors and ULs.

We could provide such a method, but I think we should:
- (a) Have a reference for it, that shows under which circumstances the approximation is valid.
- (b) Provide it as a method that combines `FluxMaps` objects, such that it works in a general way for results from all flux estimators. But only if (a) is fulfilled.

QRemy:
The reference would be Lancaster (1961) 
https://onlinelibrary.wiley.com/doi/10.1111/j.1467-842X.1961.tb00058.x
it is also mentioned in section 2.1.2 of https://www.mdpi.com/2076-3417/12/1/322
and the introduction of https://www.sciencedirect.com/science/article/abs/pii/S0167947313003204

So the method implemented here is valid under the assumption that the TS in each bin follows a Chi2 distribution.

gammapy/estimators/map/excess.py
```diff
@@ -62,6 +61,45 @@ def convolved_map_dataset_counts_statistics(dataset, kernel, mask, correlate_off
...
+        if dataset.counts.geom.to_image() != geom:
+            raise ValueError
+
+    ts_sum = Map.from_geom(geom)
+    npred_excess_sum = Map.from_geom(geom)
+    significance = Map.from_geom(geom)
+
+    dof = 0
+    for dataset in datasets:
+        results = estimator.run(dataset)
```
bkhelifi, line 89: Is it not useful to check that the estimator is an excessMapEstimator? By safety

registerrier:
I also think this is a valid method here.

You are absolutely right @adonath that this is not correct in the low count regime, but unfortunately as are all methods based on the maximum likelihood. Even via full forward folding you will get non-gaussian distribution in the very low count regime. In that sense, it does not seem to me worse than what we (and most of the field) currently do.

Conversely, stacking counts will degrade the sensitivity in many cases (e.g. mixing a measurement with relatively good signal to noise ratio with one with a poor ratio).

See e.g. the following example:
```
import numpy as np
from gammapy.stats import CashCountsStatistic, WStatCountsStatistic
from gammapy.stats.utils import ts_to_sigma

ndof = 2
n_sim = 1000000
size=(n_sim, ndof)

mu_s = np.array([4, 5])
mu_b = np.array([0.75, 15])

on = np.random.poisson(mu_s+mu_b, size=size)

stat = CashCountsStatistic(on, np.resize(mu_b, size))

plt.hist(stat.sqrt_ts, bins=50, log=True);

stacked = stat.sum(axis=1).sqrt_ts
joint = ts_to_sigma(stat.ts.sum(axis=1), ndof) * np.sign(stat.n_sig.sum(axis=1))

joint = joint[~np.isnan(stacked)]
stacked = stacked[~np.isnan(stacked)]

plt.subplot(121)
plt.hist(joint, bins=50, log=True);
plt.subplot(122)
plt.hist(stacked, bins=50, log=True);

plt.hist(joint-stacked, bins=50, log=True);
```

giacomodamico24:
Hi all,
I did a quick check using MC simulation and assuming On and Off observation (Li&Ma) and indeed for strong signal (TS >~ 20) the TS obtained by the whole sample (without dividing it into small subsamples) and the TS obtained by summing all TS from each subsample coincides.
For weak signals, as expected, the "sum over all subsamples" method tends to overestimate the TS value. 
See image below.

Maybe it would be good to add a 'warning' in the docstring that the TS value obtained might be overestimated for a weak signal.

Screenshot 2023-03-29 at 15 55 13

adonath:
Thanks a lot @giacomodamico24, this is really helpful!

adonath:
I can see the usefulness of the method to get fast results, but I really think we have to be careful here. We have seen in so many cases, that users typically do not bring the detailed statistical knowledge. That is totally fine and that’s exactly why we try to take care of that in Gammapy for our users and why users trust our software.

The bias that @giacomodamico24 has demonstrated is not that large, but as long as astronomers are after the 5 sigma threshold it will make the difference between a detection and a non-detection. If we introduce such a method in Gammapy
it will be used to identify new detections. If not...what is the point in trying to improve sensitivity with such a method? 

I think there is a contradiction: we would introduce the method to improve sensitivity close to the detection threshold, but this is also the regime where the method is biased. This does not make any sense to me. I guess the question here is, whether the gain in sensitivity exceeds the bias on average. 
 
> You are absolutely right @adonath that this is not correct in the low count regime, but unfortunately as are all methods based on the maximum likelihood. Even via full forward folding you will get non-gaussian distribution in the very low count regime. In that sense, it does not seem to me worse than what we (and most of the field) currently do.

But that's exactly why we have e.g. the stat scans. So we can keep track of the likelihood profile and make use of it later, when combining the results. But yes, this has problems too, like neglecting correlations etc., but at least we have tried to make it right!

registerrier:
Thanks @giacomodamico24 !

Are you sure this is a bias? Is the TS of the whole (stacked?) sample really the expected limit? I would not think so. 
Actually, I don't see what is really incorrect in the approach @QRemy  is proposing, as long as one corrects for the number of dof that is introduced.

> But that's exactly why we have e.g. the stat scans. So we can keep track of the likelihood profile and make use of it later, when combining the results. But yes, this has problems too, like neglecting correlations etc., but at least we have tried to make it right!

But even likelihood profiles make sense only in the asymptotic limit.

adonath:
> Are you sure this is a bias? Is the TS of the whole (stacked?) sample really the expected limit? I would not think so.
Actually, I don't see what is really incorrect in the approach @QRemy is proposing, as long as one corrects for the number of dof that is introduced.

But the test hypothesis changes, when you take into account more data. The measurements are just not independent. It boils down to:

```python3
ts_sum_approx = ts_value(norm_1, dataset_1) + ts_value(norm_2, dataset_2) 
# vs
ts_sum_joint =  ts_value(norm_joint, dataset_1) + ts_value(norm_joint, dataset_2)
```
Where `ts_value` is some non-linear function. For strong signals the different estimate for the `norm`, becomes the same. But for weak signals I presume `norm_joint` is some complex weighted average between `norm_1` and `norm_2`. Yes, the computation of significance assumes the chi2 approximation, which also allows to combine the ts values, but we would not have any corresponding estimate for the norm, norm_err, upper limits, etc.

Isn't that the whole point of doing a joint analysis in Gammapy? If considered the proposed approximation sufficient, we could throw away all of the joint likelihood framework in Gammapy. One of the most important features we "sell" to users I the joint-likelihood framework. I don't think we are doing us a favor by introducing a method to circumvent that. I know I'm strict here, but why didn't we just do an SED based on flux points for the "joint crab? It would have been less effort and faster too!

QRemy:
> If we considered the proposed approximation sufficient, we could throw away all of the joint likelihood framework in Gammapy.

The two are not incompatible, the approximation can be sufficient as a first step to locate the candidates sources before going into the full joint likelihood  framework fitting them precisely.

adonath:
> The two are not incompatible, the approximation can be sufficient as a first step to locate the candidates sources before going into the full joint likelihood framework fitting them precisely.

Yes, but it is still unclear whether the improvement in sensitivity dominates over the bias. The net gain might be zero. You find more source candidates because of improved sensitivity but all of those are false detections because the method overestimates TS.

AtreyeeS:
I join a bit late into the conversation, sorry, but from @giacomodamico24 simulations, I think the results are rather consistent, no? If we talk about the 5-sigma threshold, then in the range of around TS 25-40, the "bias" can be corrected for taking into account the number of dof ? Will this really lead to a many false detections?

adonath:
From the comments so far I see a tendency in favor of such functionality. Here is my proposal in this case:

- Introduce an explicit `FluxMaps.to_gaussian_approx_errors()` (or similar / better name), which throws away the asymmetric error definitions and likelihood profiles and just keeps the Gaussian approximation to the likelihood.
- Introduce a helper function in `estimator.utils` like `combine_estimator_results([], method={"gauss-approx", "stat-profile"})`, which takes a list of `FluxMaps` / `FluxPoints` objects and combines them internally using the Gaussian approximation (and later stat profiles as option) and returns a "joint" `FluxMap` object. The norm and errors / UL should be combined as well corresponding to the hypothesis the significance then corresponds to (would need to do the math, but it probably comes out as some error weighted average).
- Clearly document when the approximation is valid, give reference and possible modify the `success` (for a lack of an alternative) pixel flag if the approximation is just not applicable.

gammapy/estimators/map/excess.py
QRemy: added

gammapy/estimators/utils.py
```diff
@@ -6,16 +6,18 @@
...
     ConstantFluxSpatialModel,
     PowerLawSpectralModel,
     SkyModel,
 )
 from gammapy.stats import compute_flux_doubling, compute_fpp, compute_fvar
+from gammapy.stats.utils import ts_to_sigma
 from .map.core import FluxMaps
 
 __all__ = [
+    "get_joint_significance_maps",
```
AtreyeeS, line 20: Is it better to call it `get_joint_TS_maps` rather than `get_joint_significance_maps`?
Because that is what we show mostly. The Estimators return `sqrt_ts` maps rather than `significance` maps.

gammapy/estimators/utils.py
```diff
@@ -590,3 +592,74 @@ def get_rebinned_axis(fluxpoint, axis_name="energy", method=None, **kwargs):
...
+    that the TS in each independent bin follows a Chi2 distribution,
+    then the sum of the TS also follows a Chi2 distribution (with the sum of degree of freedom).
+
+    See, Zhen (2014): https://www.sciencedirect.com/science/article/abs/pii/S0167947313003204,
+    Lancaster (1961): https://onlinelibrary.wiley.com/doi/10.1111/j.1467-842X.1961.tb00058.x
+
+
+    Parameters
+    ----------
+    estimator : `~gammapy.estimator.ExcessMapEstimator`
```
AtreyeeS, line 613: Also `TSMapEstimator`

gammapy/estimators/utils.py
```diff
@@ -590,3 +592,74 @@ def get_rebinned_axis(fluxpoint, axis_name="energy", method=None, **kwargs):
...
+        Excess Map Estimator
+    dataset : `~gammapy.datasets.Datasets`
+        Datasets
```
AtreyeeS, line 616: Should contain only `MapDataset/MapDatasetOnOff`

gammapy/estimators/utils.py
```diff
@@ -590,3 +592,74 @@ def get_rebinned_axis(fluxpoint, axis_name="energy", method=None, **kwargs):
...
+    energy_edges : list of `~astropy.units.Quantity`, optional
```
AtreyeeS, line 617: `energy_edges` information is already contained in the `estimator`, no?

gammapy/estimators/utils.py
```diff
@@ -590,3 +592,74 @@ def get_rebinned_axis(fluxpoint, axis_name="energy", method=None, **kwargs):
...
+        Edges of the target maps energy bins. The resulting bin edges won't be exactly equal to the input ones,
+        but rather the closest values to the energy axis edges of the parent dataset.
+        Default is None: apply the estimator in each energy bin of the parent dataset.
+        For further explanation see :ref:`estimators`.
+    sum_over_energy_groups : bool
```
AtreyeeS, line 622: same as above

gammapy/estimators/utils.py
```diff
@@ -590,3 +592,74 @@ def get_rebinned_axis(fluxpoint, axis_name="energy", method=None, **kwargs):
...
+        Whether to sum over the energy groups or not. Default is True.
+
```
AtreyeeS, line 624: what it returns is also useful to know

AtreyeeS (changes requested):
Thanks @QRemy ! This will need a test

gammapy/estimators/utils.py
QRemy: yes removed

gammapy/estimators/utils.py
QRemy: This one was not defined on the ExcessMapEstimator, moved there instead.

gammapy/estimators/utils.py
QRemy: added

gammapy/estimators/utils.py
```diff
@@ -686,3 +688,68 @@ def get_rebinned_axis(fluxpoint, axis_name="energy", method=None, **kwargs):
         edges = np.append(edges_min, edges_max[-1])
         axis_new = MapAxis.from_edges(edges, name=axis_name, interp=ax.interp)
     return axis_new
+
+
+def get_combined_significance_maps(estimator, datasets):
+    """Computes excess and significance for a set of datasets.
+    The significance computation assumes that the model contains
+    one degree of freedom per valid bin in energy in each dataset.
```
registerrier, line 696:
```suggestion
    one degree of freedom per valid energy bin in each dataset.
```

gammapy/estimators/utils.py
```diff
@@ -686,3 +688,68 @@ def get_rebinned_axis(fluxpoint, axis_name="energy", method=None, **kwargs):
         edges = np.append(edges_min, edges_max[-1])
         axis_new = MapAxis.from_edges(edges, name=axis_name, interp=ax.interp)
     return axis_new
+
+
+def get_combined_significance_maps(estimator, datasets):
```
registerrier, line 693: Is there a specific reason for such an interface?
A priori you could simply combine the results of estimators run independently no? 
Is it to ensure that significance maps where obtained in a consistence manner?

registerrier (dismissed):

I mostly have general questions about the API, as one would expect a function to combine different maps rather than a method that performs the estimation itself.

gammapy/estimators/utils.py
QRemy: yes consistency was the goal.
The function could accept both and use directly a list of flux maps if provided.
Alternatively I could use only flux maps here and modify the estimator.run to return a list of results if a datasets is given (and eventually parallelise that loop in a separate PR).

gammapy/estimators/map/excess.py
```diff
@@ -131,6 +129,11 @@ class ExcessMapEstimator(Estimator):
     spectral_model : `~gammapy.modeling.models.SpectralModel`
         Spectral model used for the computation of the flux map.
         If None, a Power Law of index 2 is assumed (default).
+    sum_over_energy_groups : bool
```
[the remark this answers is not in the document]
QRemy: I can remove that here but I will open an other PR to suggest energy_edges = [axis.edges[0], axis.edges[-1]] as default for the excess map estimator as it what we want most of the time.

registerrier:
Waiting for CI to run and will merge now. Thanks @QRemy !

PR #4440, opened 2023-03-31
Fix example code of ring background maker

AtreyeeS:
The stacking should be done on an on-off dataset. Thanks to Emma for pointing this out!


adonath (approved):
Thanks @AtreyeeS, I agree this is more reasonable! Technically it is fine and works but usually one would keep the on-off info.

registerrier (approved):
Thanks @AtreyeeS ! This looks good. No comment for my side.

PR #4442, opened 2023-03-31
Add stack option to npred_signal()

MRegeard:
This pull request is related to PR #4409.

It aims to introduce several features to the `npred_signal()`methods. 

- First, it adds the possibility to input a list of model name instead of only one. 

- Secondly, it adds a new parameter `stack` which is True by default to keep the former behaviour in the calls of `npred_signal` in the other place of the code. 

If stack is True, the npred maps are stack one onto the others. If stack is False, I first create a new map with the geometry of the Dataset and stack the npred map on this new map. This is the simplest and fastest way I found to make all the npred maps have the same geometry (in order to do a from_stack over those maps). Then I create a `LabelMapAxis` named "models" with the label set to the name of the model. I "append" this axis to the geometry, and repeat.
Once all the npred have been computed, I do a `from_stack` of all the npred maps over the `LabelMapAxis`.

### DIFF
--- a/gammapy/datasets/map.py
+++ b/gammapy/datasets/map.py
@@ -502,7 +502,8 @@ class MapDataset(Dataset):
             evaluators = {name: self.evaluators[name] for name in model_name}
 
         npred_list = []
-        for evaluator_name, evaluator in zip(evaluators.keys(), evaluators.values()):
+        labels = []
+        for evaluator_name, evaluator in evaluators.items():
             if evaluator.needs_update:
                 evaluator.update(
                     self.exposure,
@@ -519,12 +520,12 @@ class MapDataset(Dataset):
                 else:
                     npred_geom = Map.from_geom(self._geom, dtype=float)
                     npred_geom.stack(npred)
-                    label = LabelMapAxis(labels=[evaluator_name], name="models")
-                    npred_geom = npred_geom.to_cube([label])
+                    labels.append(evaluator_name)
                     npred_list.append(npred_geom)
 
         if npred_list != []:
-            npred_total = Map.from_stack(npred_list, axis_name="models")
+            label_axis = LabelMapAxis(labels=labels, name="models")
+            npred_total = Map.from_stack(npred_list, axis=label_axis)
 
         return npred_total
 
--- a/gammapy/datasets/map.py
+++ b/gammapy/datasets/map.py
@@ -8,7 +8,7 @@ from regions import CircleSkyRegion
 import matplotlib.pyplot as plt
 from gammapy.data import GTI
 from gammapy.irf import EDispKernelMap, EDispMap, PSFKernel, PSFMap, RecoPSFMap
-from gammapy.maps import Map, MapAxis
+from gammapy.maps import LabelMapAxis, Map, MapAxis
 from gammapy.modeling.models import DatasetModels, FoVBackgroundModel
 from gammapy.stats import (
     CashCountsStatistic,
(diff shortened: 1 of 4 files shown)


gammapy/datasets/map.py
```diff
@@ -497,9 +499,10 @@ def npred_signal(self, model_name=None):
 
         evaluators = self.evaluators
         if model_name is not None:
-            evaluators = {model_name: self.evaluators[model_name]}
+            evaluators = {name: self.evaluators[name] for name in model_name}
 
-        for evaluator in evaluators.values():
+        npred_list = []
+        for evaluator_name, evaluator in zip(evaluators.keys(), evaluators.values()):
```
adonath, line 505: This is usually written as:
```suggestion
        for evaluator_name, evaluator in evaluators.items():
```

adonath (commented):
Thanks @MRegeard! I have left one minor suggestion. Otherwise I think the implementation would slightly simplify, when defining the `LabelMapAxis` first using all labels and later using `MapAxis.from_stack(npred_list, axis=axis)` instead. This avoids the repeated creation of `LabelMapAxis` and calls to `.to_cube()`.

gammapy/datasets/map.py
MRegeard: done

gammapy/datasets/map.py
```diff
@@ -475,18 +475,20 @@ def _background_parameters_changed(self):
...
+        If stack is set to True, a map of the sum of all the predicted counts is returned.
+        If stack is set to False, a map with an additional axis representing the models is returned.
 
         Parameters
         ----------
-        model_name: str
-            Name of  SkyModel for which to compute the npred for.
-            If none, the sum of all components (minus the background model)
-            is returned
+        model_name: list of str
```
AtreyeeS, line 487: I guess it should be `model_names` here (with a deprecation warning) to be consistent with our datasets and models naming schemes

AtreyeeS (commented):
Thanks @MRegeard ! This looks very clean and useful

gammapy/datasets/map.py
MRegeard: Done. 
For the deprecation version I put 1.1 but I don't know it is the right one to put.

QRemy (commented):
Thanks @MRegeard, maybe passing only an `str` could still be supported.
We have this pattern in few other functions:

```
if model_names and not isinstance(model_names, list):
    model_names = [model_names]
```

gammapy/datasets/map.py
```diff
@@ -475,18 +476,21 @@ def _background_parameters_changed(self):
             self._background_parameters_cached = values
         return changed
 
-    def npred_signal(self, model_name=None):
+    @deprecated_renamed_argument("model_name", "model_names", "1.1")
+    def npred_signal(self, model_names=None, stack=True):
```
AtreyeeS, line 480: Do we need a `stack` option here? Users can do `dataset.npred_signal(model_names).stack()` if necessary. This is the general scheme we follow, no?

AtreyeeS (approved):
Thanks @MRegeard ! I am just a little bit unsure about doing the stacking inside `npred_signal()` rather than in `LabelMapAxis.stack()` 
Apart from that, it would be good to demonstrate
 this in some tutorial (can be a future PR). Maybe the `modelling` tutorial..

gammapy/datasets/map.py
MRegeard: I have done this to avoid to have to rewrite the code everywhere where `npred_signal()` is called. 
That's why it is set to `True` by default, so that by default it behave exactly as before.

AtreyeeS: Yes indeed, I forgot this is necessary for backwards compatibility.

QRemy: Having `stack=True` by default also save memory.

QRemy (approved):
Thanks @MRegeard, no further comments.

Issue #4329, opened 2023-02-14
Calculate Integral Flux Sensitivities


STSpencer:
**Is your feature request related to a problem? Please describe.**
As far as I can tell in the documentation or source code, the gammapy _SensitivityEstimator_ class lacks the capability to calculate integral flux sensitivities from GADF IRF files, and can only calculate differential sensitivities. This is needed, for example, for CTA source detectability studies. ctools has this functionality built into its cssens tool.

**Describe the solution you'd like**
Could this possibly be implemented?

**Describe alternatives you've considered**
Alternatively, is there a conceivable hack that could be used with the existing gammapy classes? Thanks in advance.

adonath:
Thanks @STSpencer! You can e.g. use our flux points class to convert the sensitivities:

```python3
from gammapy.estimators import FluxPoints

table = energy_axis.to_table(format="gadf-sed")
table["e2dnde"] = sensitivity_table["e2dnde"]

fp = FluxPoints.from_table(
    table=table,
    format="gadf-sed",
    sed_type="e2dnde",
    reference_model=sensitivity_estimator.spectrum
    
)

fp.to_table(sed_type="flux")
```
Given you have an `energy_axis` from e.g. a `SpectrumDatasetOnOff` you passed to the `SensitivityEstimator`. I hop ethos helps...

adonath:
I know realize that you probably refer to the integrated flux above a certain energy threshold. This is not as simple to achieve...

bkhelifi:
Hi @adonath ,
Looking at the details, I see that the function `SensitivityEstimator.estimate_min_excess` could be extended to compute the array of counts integrated over bins instead of the array of counts per bin. And then the function `estimate_min_e2dnde` can be extended accordingly. 
The changes are not so difficult but they are needed. This can be a nice small project for beginners to have an option 'integral' in this estimator.

adonath:
@bkhelifi Yes, I agree it is indeed not too complex. Looking at the table the sensitivity estimator returns I think we should probably make the columns compatible with the `"gadf-sed"` format as well. So use column names `e_ref` and later `e_min` and `e_max`. For the integral sensitivity `e_min` will be set to the energy threshold and `e_max` to a very large value or infinity...

AtreyeeS:
Do we need to adapt the estimator here? Can this not be simply achieved by having one energy bin in the dataset?

adonath:
[...]

PR #4444, opened 2023-03-31
Integral sensitivity in FluxPointsEstimator

AtreyeeS:
The PR address #4329

The issue is how to keep backwards compatibility while renaming `energy` to `e_min` in the `Table` column.
One option might be to have both columns in the table, but that is rather ugly.

### DIFF
--- a/examples/tutorials/analysis-1d/cta_sensitivity.py
+++ b/examples/tutorials/analysis-1d/cta_sensitivity.py
@@ -209,16 +209,12 @@ plt.show()
 #
 # It is often useful to obtain the integral sensitivity above a certain
 # threshold. In this case, it is simplest to use a dataset with one energy bin
-# while setting the high energy edge to a very large value
+# while setting the high energy edge to a very large value.
+# Here, we simply squash the previously created dataset into one with a single
+# energy
 #
 
-energy_axis1 = MapAxis.from_energy_bounds("0.03 TeV", "300 TeV", nbin=1)
-geom = RegionGeom.create("icrs;circle(0, 0.5, 0.1)", axes=[energy_axis1])
-empty_dataset1 = SpectrumDataset.create(geom=geom, energy_axis_true=energy_axis_true)
-dataset1 = spectrum_maker.run(empty_dataset1, obs)
-dataset_on_off1 = SpectrumDatasetOnOff.from_spectrum_dataset(
-    dataset=dataset1, acceptance=1, acceptance_off=5
-)
+dataset_on_off1 = dataset_on_off.to_image()
 sensitivity_estimator = SensitivityEstimator(
     gamma_min=5, n_sigma=3, bkg_syst_fraction=0.10
 )
@@ -232,7 +228,7 @@ flux_points = FluxPoints.from_table(
     sensitivity_table, sed_type="e2dnde", reference_model=sensitivity_estimator.spectrum
 )
 print(
-    f"Integral sensitivity in {livetime:.2f} above {energy_axis1.edges[0]:.2e} "
+    f"Integral sensitivity in {livetime:.2f} above {energy_axis.edges[0]:.2e} "
     f"is {np.squeeze(flux_points.flux.quantity):.2e}"
 )
 
--- a/examples/tutorials/analysis-1d/cta_sensitivity.py
+++ b/examples/tutorials/analysis-1d/cta_sensitivity.py
@@ -39,7 +39,7 @@ import matplotlib.pyplot as plt
 from IPython.display import display
 from gammapy.data import Observation, observatory_locations
 from gammapy.datasets import SpectrumDataset, SpectrumDatasetOnOff
-from gammapy.estimators import SensitivityEstimator
+from gammapy.estimators import FluxPoints, SensitivityEstimator
 from gammapy.irf import load_irf_dict_from_file
 from gammapy.makers import SpectrumDatasetMaker
 from gammapy.maps import MapAxis, RegionGeom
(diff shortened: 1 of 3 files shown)


adonath:
Thanks a lot @AtreyeeS! Indeed to get a single value for the integral sensitivity we can just use a single energy bin. In general (and also requested in #4329) I think it would be good to have this for an array of minimum energies.  

Concerning this PR:
- I think adding the `e_min` and `e_max` to the table is a very good idea. For backwards compatibility we could keep `energy` as well. But we should probably emit a deprecation warning in the `.run()` method.
- I think the example with a single energy bins is useful, but probably not as useful as having the same information fo an array of energies.

So to fully address the issue we would need a few more modifications (but this can be a separate PR):

- Adapt the `SensitivityEstimator` to compute the sensitive for the energy bins defined e.g. by `e_min = [1, 2, 3] * u.TeV` and `e_max = [np.inf, np.inf, np.inf] * u.TeV`. I think the definition of an `e_ref` would not apply here.
- Currently the `SensitivityEstimator` just takes the energy binning from the dataset. The `SensitivityEstimator`could be modified to take `energy_min` and `energy_max` and `energy` arguments separately. For the integral sensitivity one would then explicitly use `SensitivityEstimator(energy_min=[1, 2, 3] * u.TeV, energy_max=np.inf * u.TeV)`
- In this case one would internally have a loop of energy bins and resample the dataset to the given energy range.
- Alternatively we could just keep the logic of  taking the energy binning from the dataset.
- In this case it could be very useful to have a method like `CountsStatistic.cumsum()`, which cumulates the counts along the given axis. Maybe this is worth an independent PR.

AtreyeeS:
Is it very useful to have the integral sensitivity as a function of energy? Wouldn't it be more practical then to compute a differential sensitivity?

adonath:
@AtreyeeS I think this is useful in the context of source detection and finding the energy threshold which corresponds to the best sensitivity. But maybe @STSpencer can comment again.

STSpencer:
Thanks for working on this @AtreyeeS !

So the reason we want this feature is to determine whether or not particular molecular clouds with various emission scenarios will be visible with CTA, in a similar manner to Mitchell et al. 2021 (https://arxiv.org/abs/2103.01787). For this you need to compute the predicted integrated flux from the source to the integrated sensitivity above a given energy threshold. In principle this is just one number per threshold, but I think @adonath is right (I should say I'm not an expert in this) in that being able to compute this number for arrays of minimum energy thresholds is the way to go, as this seems to be important for transient studies (see (Fioretti et al. 2016) https://arxiv.org/abs/1608.04992).

AtreyeeS:
Thanks! Indeed, I see now. 
Maybe we merge this PR now for 1.1, so that users can at least manually loop over thresholds and get the integral sensitivity curve. We keep #4329 open for 1.2, to be implemented with @adonath suggestions.

QRemy (dismissed):
Thanks @AtreyeeS, looks good,  no further comments from my side. I agree the remaining comments can be implemented in a separate PR.

gammapy/estimators/points/sensitivity.py
```diff
@@ -135,6 +141,11 @@ def run(self, dataset):
         criterion = self._get_criterion(
             excess.data.squeeze(), dataset.background.data.squeeze()
         )
+        logging.warning(
```
registerrier, line 144: Does this require a deprecation warning? It is a change in the output format, not in the API.

gammapy/estimators/points/tests/test_sensitivity.py
```diff
@@ -41,28 +41,63 @@ def test_cta_sensitivity_estimator(spectrum_dataset):
     sens = SensitivityEstimator(gamma_min=25, bkg_syst_fraction=0.075)
     table = sens.run(dataset_on_off)
 
+    warning_message = (
```
registerrier, line 44: I am not sure you should test for this.

examples/tutorials/analysis-1d/cta_sensitivity.py
```diff
@@ -188,20 +189,52 @@
...
+######################################################################
+# Obtaining an integral flux sensitivity
+# --------------------------------------
+#
+# It is often useful to obtain the integral sensitivity above a certain
+# threshold. In this case, it is simplest to use a dataset with one energy bin
+# while setting the high energy edge to a very large value
+#
+
+energy_axis1 = MapAxis.from_energy_bounds("0.03 TeV", "300 TeV", nbin=1)
```
registerrier, line 215: here you could squash the `energy_axis` defined above

registerrier: Or even simpler `geom.squash("energy")`

examples/tutorials/analysis-1d/cta_sensitivity.py
```diff
@@ -188,20 +189,52 @@
...
+geom = RegionGeom.create("icrs;circle(0, 0.5, 0.1)", axes=[energy_axis1])
+empty_dataset1 = SpectrumDataset.create(geom=geom, energy_axis_true=energy_axis_true)
```
registerrier, line 217: or `dataset_on_off.to_image()` or `dataset.resample_energy_axis`

registerrier (commented):

I have left some inline comments to simplify the reduction of the dataset in the notebook.
Also I don't think we need deprecation here. You change the output table format, this is different.

A question might be whether or not to use the `FluxPoints` object instead of a table directly.

gammapy/estimators/points/sensitivity.py
AtreyeeS: Thanks @registerrier ! Interpreting backwards compatibility loosely, I tend to see it as "something which breaks my script which used to work on a prev version". In this case, users might like to be fore-warned that `table["energy"]` would yield errors in future versions?

bkhelifi:
I would have expected in the tutorial that the integral sensitivity is similar to the differential one, ie the minimal flux above an energy threshold as a function of threshold. The standard output should look like the differential sensitivity, ie a plot with X=energy and Y=flux [/cm2/s] .
With this implementation, one needs to re-run n times the `SensitivityEstimator`. It seems to me not very efficient, but maybe we have no choice. At least one should add the parallel processing for the loop on energy thresholds

registerrier (approved):

adonath:
@AtreyeeS You could add one more comment in the tutorial to mention that the in future we will support varying energy thresholds.

PR #4448, opened 2023-04-03
Catch invalid conversion warning in pix_tuple_to_idx

registerrier:
**Description**

This pull request introduces a method to catch invalid rounding warnings from numpy that are now raised systematically since numpy 1.24. They occur mostly in `gammapy.maps.geom.pix_tuple_to_idx`.  (See for instance outputs in v1.0.1 tutorials). 

**Dear reviewer**

### DIFF
--- a/gammapy/maps/geom.py
+++ b/gammapy/maps/geom.py
@@ -47,8 +47,8 @@ def pix_tuple_to_idx(pix):
         else:
             with np.errstate(invalid="ignore"):
                 p_idx = np.rint(p).astype(int)
-                p_idx[~np.isfinite(p)] = INVALID_INDEX.int
-                idx += [p_idx]
+            p_idx[~np.isfinite(p)] = INVALID_INDEX.int
+            idx += [p_idx]
 
     return tuple(idx)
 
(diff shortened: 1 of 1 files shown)


gammapy/maps/geom.py
```diff
@@ -45,9 +45,10 @@ def pix_tuple_to_idx(pix):
         if np.issubdtype(p.dtype, np.integer):
             idx += [p]
         else:
-            p_idx = np.rint(p).astype(int)
-            p_idx[~np.isfinite(p)] = INVALID_INDEX.int
-            idx += [p_idx]
+            with np.errstate(invalid="ignore"):
+                p_idx = np.rint(p).astype(int)
```
adonath, line 49: Minor: ideally the with context only covers the line in question. Or is the warning emitted from all three lines?

registerrier: Indeed. Corrected.

registerrier:
> Thanks, @registerrier the warning may also appears in the ExcessMapEstimator
> 
> https://github.com/gammapy/gammapy/blob/9718fc54202baee7077cd727e50e606d1265d21f/gammapy/estimators/map/excess.py#L31
> 
> 

The warning occurs because of the type conversion not the `np.rint`. So here it should not matter.

> and we also use np.rint in Wcsgeom but the warning probably never appears in that case
> https://github.com/gammapy/gammapy/blob/9718fc54202baee7077cd727e50e606d1265d21f/gammapy/maps/wcs/geom.py#L386-L387

In this case, I don't know. I suppose that infinite values should break somewhere else here.

QRemy (approved):

Issue #4425, opened 2023-03-27
TemporalModel miss a scale attribute


registerrier:
**Gammapy version**
v1.0.1 and previous versions

**Bug description**
The `t_ref` parameter on `TemporalModel` does not have a `scale` attached. The scale used will in practice be the one from `Time` object passed on evaluation.

**Expected behavior**
The model reference time should have a `scale` associated as the `SpatialModel` has a `frame` attribute.

**To Reproduce**
```
from astropy.time import Time
from gammapy.modeling.models import ExpDecayTemporalModel

model = ExpDecayTemporalModel(t0 = "1h", t_ref="55000 d")
time = Time(55000.0000001, format="mjd", scale="tt")

print(model(time))
print(model(time.utc))
```
**Other information**

registerrier:
This is now solved in #4453. Closing.

PR #4453, opened 2023-04-04
Add scale in temporal model

AtreyeeS:
This address #4425 but adding a scale attribute

### DIFF
--- a/gammapy/modeling/models/temporal.py
+++ b/gammapy/modeling/models/temporal.py
@@ -43,7 +43,7 @@ class TemporalModel(ModelBase):
         scale = kwargs.pop("scale", "utc")
         if scale not in Time.SCALES:
             raise ValueError(
-                f"{scale} is not a valid time scale. Choose" f"from {Time.SCALES}"
+                f"{scale} is not a valid time scale. Choose from {Time.SCALES}"
             )
         super().__init__(**kwargs)
         if not hasattr(self, "scale"):
--- a/gammapy/modeling/models/temporal.py
+++ b/gammapy/modeling/models/temporal.py
@@ -62,25 +62,20 @@ class TemporalModel(ModelBase):
         return self._type
 
     @property
-    def t_ref_mjd(self):
+    def reference_time(self):
         """Reference time in mjd"""
         return Time(self.t_ref.value, format="mjd", scale=self.scale)
 
-    @t_ref_mjd.setter
-    def t_ref_mjd(self, t_ref):
+    @reference_time.setter
+    def reference_time(self, t_ref):
         """Reference time"""
-        self.t_ref_mjd = getattr(t_ref, self.scale, "mjd")
+        self.reference_time = getattr(t_ref, self.scale, "mjd")
 
     def to_dict(self, full_output=False):
         """Create dict for YAML serilisation"""
         data = super().to_dict(full_output)
         data["temporal"]["scale"] = self.scale
-        data["temporal"]["parameters"] = data["temporal"].pop("parameters")
         return data
-
-    def from_dict(cls, data):
-        data = data["temporal"]
-        return super().from_dict(data)
 
     @staticmethod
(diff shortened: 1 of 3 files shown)


gammapy/modeling/models/temporal.py
```diff
@@ -54,6 +60,27 @@ def __call__(self, time):
     def type(self):
         return self._type
 
+    @property
+    def t_ref_mjd(self):
```
registerrier, line 64: Maybe use `reference_time` instead.

registerrier (commented):
Thanks @AtreyeeS

gammapy/modeling/models/temporal.py
```diff
@@ -55,6 +61,22 @@ def __call__(self, time):
...
 
+    @property
+    def reference_time(self):
+        """Reference time in mjd"""
+        return Time(self.t_ref.value, format="mjd", scale=self.scale)
+
+    @reference_time.setter
+    def reference_time(self, t_ref):
+        """Reference time"""
+        self.reference_time = getattr(t_ref, self.scale, "mjd")
```
registerrier, line 72: Does this work?
I think you should add a check that this is a `Time` object. Then set `t_ref = reference_time.mjd` and `self.scale = t_ref.scale`

gammapy/modeling/models/tests/test_temporal.py
```diff
@@ -405,3 +411,19 @@ def test_phasecurve_DC1():
 
     with mpl_plot_check():
         model.plot_phasogram(n_points=200)
+
+
+def test_model_scale():
```
registerrier, line 416: Please add a test for the `reference_time` setter.

registerrier (commented):
Thanks @AtreyeeS ! Looks good. See minor inline comments.

gammapy/modeling/models/temporal.py
AtreyeeS: Should the scale of the model change according to the `Time` passed, or the other way round.
In `SpatialModels`, its the latter, see https://github.com/gammapy/gammapy/blob/610e9d4202b8bef70e61e83a8e73b5b4ca6422b8/gammapy/modeling/models/spatial.py#L104

registerrier: Yes we should follow the same convention. It is still possible to change manually the `scale`.

AtreyeeS: Added a test

fabiopintore:
Thanks @AtreyeeS , it looks very good to me. I don't have further comments.

gammapy/modeling/models/temporal.py
```diff
@@ -39,6 +39,12 @@ class TemporalModel(ModelBase):
 
     _type = "temporal"
 
+    def __init__(self, **kwargs):
+        scale = kwargs.pop("scale", "utc")
+        super().__init__(**kwargs)
+        if not hasattr(self, "scale"):
+            self.scale = scale
```
registerrier, line 46: It might be worth checking that scale is in `Time.SCALES`. Or at least check that the reference_time exists.

AtreyeeS: A ConstantTemporalModel has no reference time

gammapy/modeling/models/temporal.py
```diff
@@ -39,6 +39,16 @@ class TemporalModel(ModelBase):
 
     _type = "temporal"
 
+    def __init__(self, **kwargs):
+        scale = kwargs.pop("scale", "utc")
+        if scale not in Time.SCALES:
+            raise ValueError(
+                f"{scale} is not a valid time scale. Choose" f"from {Time.SCALES}"
```
registerrier, line 46:
```suggestion
                f"{scale} is not a valid time scale. Choose from {Time.SCALES}"
```

registerrier (approved):
Thanks @AtreyeeS . Perfect.

PR #4455, opened 2023-04-06
Test SpatialModel.from_position()

AtreyeeS:
In v1.0.1, `GaussianSpatialModel.from_position` gives an error `SkyModel` creation 

```
155 def evaluate_geom(self, geom):
    156     """Evaluate model on `~gammapy.maps.Geom`
    157 
    158     Parameters
   (...)
    165 
    166     """
--> 167     coords = geom.get_coord(frame=self.frame, sparse=True)
    169     if self.is_energy_dependent:
    170         return self(coords.lon, coords.lat, energy=coords["energy_true"])

TypeError: unhashable type: 'ICRS'
```

The issue was fixed (unintentionally) in #4282 - the frame argument should be `position.frame.name` rather than `position.frame`

The function was also untested.

This PR adds a test. I intended to fix it, but the issue does not exist on the main branch. Should I rebase on the `1.0.x` branch @adonath @registerrier ?

### DIFF
--- a/gammapy/modeling/models/spatial.py
+++ b/gammapy/modeling/models/spatial.py
@@ -407,7 +407,7 @@ class SpatialModel(ModelBase):
     @classmethod
     def from_position(cls, position, **kwargs):
         """Define the position of the model using a sky coord
-
+           The model will be created in the frame of the sky coord
         Parameters
         ----------
         position : `~astropy.coordinates.SkyCoord`
--- a/gammapy/modeling/models/tests/test_spatial.py
+++ b/gammapy/modeling/models/tests/test_spatial.py
@@ -19,8 +19,10 @@ from gammapy.modeling.models import (
     GaussianSpatialModel,
     GeneralizedGaussianSpatialModel,
     PointSpatialModel,
+    PowerLawSpectralModel,
     Shell2SpatialModel,
     ShellSpatialModel,
+    SkyModel,
     TemplateSpatialModel,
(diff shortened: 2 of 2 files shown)


AtreyeeS:
As discussed with @registerrier , this PR can go in for v1.2, without back port.
For the bug fix, a new PR based on the 1.0.x branch can be opened

QRemy (approved):

registerrier:
> > As discussed with @registerrier , this PR can go in for v1.2, without back port.
> > For the bug fix, a new PR based on the 1.0.x branch can be opened
> 
> As this PR does not contain any new feature it could be back-ported. So you could include the fix here...what do you think @registerrier ?

I understand that this is correct in dev:
https://github.com/gammapy/gammapy/blob/edb09c8e84e1104051c5498f868e4ea09fd62e8d/gammapy/modeling/models/spatial.py#L422

So this cannot be added here, but should be modified directly into v1.0.x. 
Then maybe, first open a PR on v1.0.x modyfing the line and then we merge and backport this one?

adonath:
> So this cannot be added here, but should be modified directly into v1.0.x.

Thanks, now I understand. This PR is against `dev` so there is no change...however having the test in v1.0.2 would be good as well. Not sure how we achieve this most efficiently, but copy and paste might be fine as well.

registerrier (approved):
Thanks @AtreyeeS

Issue #4457, opened 2023-04-06
There is no weights options for Map.fill_events


MRegeard:
In `Map`, there is no weights option for `Map.fill_events` while there is one for all the other "fill" methods. 
Moreover, `fill_events` uses `fill_by_coord` which has a weights option so the implementation should be straight forward.

MRegeard:
That true. However, if the case of an unbinned dataset it is not the case I think.

MRegeard:
We had the use case yesterday where we tried the unbinned analysis prototype and calculated the weights of each events for a given spectral model. At some point we wanted to put the event in a map but we could apply the calculated weights for each events.

PR #4458, opened 2023-04-06
adding weights option to fill_events

MRegeard:
This pull request is related to #4457. 
I added a weights option to `Map.fill_events`. Since `Map.fill_events` uses `Map.fill_by_coord` which has a weights option, the implementation is straight forward.

### DIFF
--- a/gammapy/maps/core.py
+++ b/gammapy/maps/core.py
@@ -1089,7 +1089,15 @@ class Map(abc.ABC):
         return output_map
 
     def fill_events(self, events, weights=None):
-        """Fill event coordinates (`~gammapy.data.EventList`)."""
+        """Fill event coordinates (`~gammapy.data.EventList`).
+
+        Parameters
+        ----------
+        events : `~gammapy.data.EventList`
+            Events to be fill in the map.
+        weights : `~numpy.ndarray`
+            Weights vector. Default is weight of one.
+        """
         self.fill_by_coord(events.map_coord(self.geom), weights=weights)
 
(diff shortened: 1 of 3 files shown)


adonath:
Thanks @MRegeard! I'm not sure we would really like to expose this, where we explicitly refer to "events" (see [my comment ](https://github.com/gammapy/gammapy/issues/4457#issuecomment-1499038177)). What do you think @registerrier ?

gammapy/maps/core.py
```diff
@@ -1088,9 +1088,17 @@ def reproject_to_geom(self, geom, preserve_counts=False, precision_factor=10):
...
-        self.fill_by_coord(events.map_coord(self.geom))
+    def fill_events(self, events, weights=None):
+        """Fill event coordinates (`~gammapy.data.EventList`).
+
+        Parameters
+        ----------
+        events : `~gammapy.data.EventList`
+            Events to be fill in the map.
+        weights : `~numpy.ndarray`
+            Weights vector. Default is weight of one.
```
adonath, line 1099: Add that it should have the same shape as `events` or be broadcastable (not sure what is correct...)

adonath (commented):
Thanks @MRegeard! Please add a minimal test to establish the behavior...

registerrier (commented):
Thanks @MRegeard . Looks good. Please a a test for this new functionality.

registerrier (approved):
Thanks @MRegeard . No further comment.

PR #4460, opened 2023-04-06
Fix SpatialModel.from_position()

AtreyeeS:
I think once this is merged, #4455 can be back ported if necessary.


registerrier (approved):

Issue #4327, opened 2023-02-10
Consistently suppress output after plot calls


adonath:
To avoid getting meaningless output after `.plot()` calls (see e.g. https://docs.gammapy.org/1.0/tutorials/starting/analysis_1.html#data-reduction), we can consistently add a semicolon like `.plot();`. This cleans up the output in the tutorials...

AtreyeeS:
Addressed in #4483

PR #4466, opened 2023-04-13
Add tutorial for the use of HAWC data

LauraOlivera:
This pull request adds a tutorial showing how to open and use the HAWC data from the public data release.

The tutorial goes slightly beyond the usual "data" tutorial by showing the data reduction because there are some subtleties that are different for HAWC.

It also adds/fixes some things that I encountered along the way, namely:

- a typo in the HESS tutorial
- the ability to `.peek()` a PSFMap when no exposure map is set
- the ability to `.create()` an empty `MapDataset` where the psf is a `RecoPSFMap`

### DIFF
--- a/examples/tutorials/data/hawc.py
+++ b/examples/tutorials/data/hawc.py
@@ -182,6 +182,9 @@ safemask_maker = SafeMaskMaker(methods=["aeff-max"], aeff_percent=10)
 
 ######################################################################
 # Create empty Mapdataset
+# The keyword reco_psf=True is needed because the HAWC PSF is
+# derived in reconstructed energy.
+
 dataset_empty = MapDataset.create(
     geom, energy_axis_true=energy_axis_true, name="fHit " + str(fHit), reco_psf=True
 )
--- a/examples/tutorials/data/hawc.py
+++ b/examples/tutorials/data/hawc.py
@@ -38,16 +38,16 @@ This is how to access data and IRFs from the HAWC Crab event data release.
 
 """
 
+import numpy as np
 import astropy.units as u
 from astropy.coordinates import SkyCoord
 import matplotlib.pyplot as plt
-import numpy as np
 from IPython.display import display
-from gammapy.datasets import MapDataset
-from gammapy.makers import MapDatasetMaker, SafeMaskMaker
 from gammapy.data import DataStore, HDUIndexTable, ObservationTable
-from gammapy.maps import Map, MapAxis, WcsGeom
+from gammapy.datasets import MapDataset
 from gammapy.estimators import ExcessMapEstimator
+from gammapy.makers import MapDatasetMaker, SafeMaskMaker
+from gammapy.maps import Map, MapAxis, WcsGeom
 
 ######################################################################
 # Check setup
@@ -60,7 +60,7 @@ check_tutorials_setup()
 ######################################################################
 # Chose which estimator we will use
 
-which = 'NN'
+which = "NN"
 
 
 ######################################################################
@@ -85,21 +85,21 @@ which = 'NN'
 # Load the tables
 
 data_path = "$GAMMAPY_DATA/hawc/crab_events_pass4/"
-hdu_filename = 'hdu-index-table-' + which + "-Crab.fits.gz"
-obs_filename = 'obs-index-table-' + which + "-Crab.fits.gz"
+hdu_filename = "hdu-index-table-" + which + "-Crab.fits.gz"
+obs_filename = "obs-index-table-" + which + "-Crab.fits.gz"
 
 # there is only one observation table
-obs_table = ObservationTable.read(data_path+obs_filename)
+obs_table = ObservationTable.read(data_path + obs_filename)
 
 # we read the hdu index table of fHit bin number 6
 fHit = 6
-hdu_table = HDUIndexTable.read(data_path+hdu_filename, hdu=fHit)
+hdu_table = HDUIndexTable.read(data_path + hdu_filename, hdu=fHit)
 
 
 ######################################################################
 # From the tables, we can create a Datastore
 
-data_store = DataStore( hdu_table=hdu_table, obs_table=obs_table)
+data_store = DataStore(hdu_table=hdu_table, obs_table=obs_table)
 
 data_store.info()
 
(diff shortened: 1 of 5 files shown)


examples/tutorials/data/hawc.py
```diff
@@ -0,0 +1,257 @@
...
+# process inside of a for loop that modifies the variable fHit.
+
+
+######################################################################
+# First we define the geometry and axes
+
+# create the energy reco axis
+energy_axis = MapAxis.from_edges([1.00,1.78,3.16,5.62,10.0,17.8,31.6,56.2,100,177,316] * u.TeV,
+        name="energy",
+        interp="log")
```
QRemy, line 153: Maybe mention that this is the energy binning used to create the background template and if different values are given the template will be interpolated during the dataset creation by the MapDatasetMaker.

examples/tutorials/data/hawc.py
```diff
@@ -0,0 +1,257 @@
...
+# of transits can be derived using the good time intervals (GTI) stored
+# with the event list. For convenience, the HAWC data release already
+# included this quantity as a map
+
+transit_map = Map.read(data_path + 'irfs/TransitsMap_Crab.fits.gz')
+transit_number = transit_map.get_by_coord(geom.center_skydir)
+
+# Correct the quantities
+dataset.background.data*=transit_number
+dataset.exposure.data*=transit_number
```
QRemy, line 204: could this be done automatically by the MapDatasetMaker ?

QRemy (commented):
Thanks @LauraOlivera, this looks good.  I wonder if some of the steps could be done internally within gammapy rather that by hand, but we can figure this out later.
I left some minor suggestion that I will push directly.
And you should probably run `pip install pre-commit`, `pre-commit install` to have the style formatting done automatically for each commit.

examples/tutorials/data/hawc.py
```diff
@@ -0,0 +1,257 @@
...
+                                  reco_psf=True )
+# run the map dataset maker
+dataset = maker.run(dataset_empty, obs)
+
+# The livetime info is used by the SafeMask maker to retrieve the
+# effective area from the exposure. The HAWC effective area is computed
+# for one source transit above 45º zenith, which is around 6h
+# Note that since the effective area condition used here is relative to
+# the maximum, this value does not actually impact the result
+dataset.exposure.meta["livetime"] = "6 h"
```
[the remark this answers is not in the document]
LauraOlivera: Something similar is used yes. Bins with less events than a certain fraction are ignored and deemed "bad". Which is analogous to using the effective area as a threshold... And you need some way to remove these "nonsensical bins" (e.g. high fHit, very high energy etc)

examples/tutorials/data/hawc.py
LauraOlivera: see comment above - yes it could and would be nice, but would also mean a significant new feature (i.e. computing exposure from GTIs)

examples/tutorials/data/hawc.py
LauraOlivera: good point, I will add this

bkhelifi:
For HAWC, do we need to change the default spatial binning of the IRFs (==0.2deg)?

QRemy:
> For HAWC, do we need to change the default spatial binning of the IRFs (==0.2deg)?

Yes it should not be much smaller than in the original geometry of the edisp and  psf maps
Alternatively one can use directly their original geometries to init the dataset:

```
geom_psf = obs.psf.psf_map.geom
geom_edisp = obs.edisp.edisp_map.geom
geom_exposure = geom.to_image().to_cube([energy_axis_true])

dataset_empty = MapDataset.from_geoms(geom, geom_exposure=geom_exposure, geom_psf=geom_psf, geom_edisp=geom_edisp, name="fHit " + str(fHit))
```

LauraOlivera:
> Yes it should not be much smaller than in the original geometry of the edisp and psf maps.

I actually don't necessary agree with this - the HAWC IRFs are computed in 1deg declination bands but are expected to behvave smoothly in between. So having bins in scales smaller than the 1 deg (e.g. steps of 0.2 deg) and interpolating is actually not bad. Why do you think so?

examples/tutorials/data/hawc.py
LauraOlivera: done

examples/tutorials/data/hawc.py
```diff
@@ -0,0 +1,268 @@
...
+
+######################################################################
+# Peek the background for one transit
+plt.figure()
+obs.bkg.reduce_over_axes().plot(add_cbar=True)
+
+######################################################################
+# Peek the effective exposure for one transit
+plt.figure()
+obs.aeff.reduce_over_axes().plot(add_cbar=True)
```
AtreyeeS, line 135: The background and aeff are per transit, but the psf or edisp are not?

examples/tutorials/data/hawc.py
```diff
@@ -0,0 +1,268 @@
...
+# process inside of a for loop that modifies the variable fHit.
+
+
+######################################################################
+# First we define the geometry and axes
+
+# create the energy reco axis
+# Note that this axis is the one used to create the background model map. If
+# different edges are used, the MapDatasetMaker will interpolate between
+# them, which might lead to unexpected behavior.
```
AtreyeeS, line 153: The same issue could arise in case of, eg, H.E.S.S. analysis, no? Why is the interpolation a bigger concern here?

examples/tutorials/data/hawc.py
```diff
@@ -0,0 +1,268 @@
...
+energy_axis = MapAxis.from_edges(
+    [1.00, 1.78, 3.16, 5.62, 10.0, 17.8, 31.6, 56.2, 100, 177, 316] * u.TeV,
+    name="energy",
+    interp="log",
+)
+
+
+# and energy true axis
+energy_axis_true = MapAxis.from_energy_bounds(
+    1e-3, 1e4, nbin=140, unit="TeV", name="energy_true"
```
AtreyeeS, line 163: is the edisp so poor at low energies that you need to start from 1e-3 TeV for 1TeV ereco?

examples/tutorials/data/hawc.py
```diff
@@ -0,0 +1,268 @@
...
+    axes=[energy_axis],
+    binsz=0.05,
+)
+
+
+######################################################################
+# Define the makers we will use
+
+maker = MapDatasetMaker(selection=["counts", "background", "exposure", "edisp", "psf"])
+safemask_maker = SafeMaskMaker(methods=["aeff-max"], aeff_percent=10)
```
AtreyeeS, line 180: No need for `bkg-peak` here ?

AtreyeeS (approved):
Thanks @LauraOlivera ! Excellent tutorial, and I have more questions (see inline) than useful comments.

Would it be too time consuming to show the analysis for 2 nhit bins, and then do a joint analysis? Then the notebook can be self sufficient.

And a very minor point - maybe you can add a ";" after the plot commands (see #4327)

examples/tutorials/data/hawc.py
QRemy: For HAWC, contrary to HESS, if the background template is used with its default binning it should match exactly without need for scaling . The interpolation might just add systematics or numerical noise depending on how it is performed .

examples/tutorials/data/hawc.py
QRemy: no by default the background template scales well in each energy bin.

QRemy (approved):

examples/tutorials/data/hawc.py
AtreyeeS: Maybe it can be better to show how to get the energies then? `ereco = obs.bkg.geom.axes["energy"]` and then `print(ereco.edges)`

examples/tutorials/data/hawc.py
AtreyeeS: During fitting, do you  keep the background norm/tilt free?

examples/tutorials/data/hawc.py
LauraOlivera: Everything is for one transit, but the PSF and EDisp don't depend on the number of transits, whereas the effective exposure and the background do.

examples/tutorials/data/hawc.py
LauraOlivera: Ah indeed, this is just the same axis I always use (which matches the IRFs) - you're right that in principle one could start at e.g. 1e-2 TeV without much issue (see Fig 4 of https://arxiv.org/pdf/2109.04097.pdf). But the edisp is quite poor at low energies so I wouldn't go higher than that...

examples/tutorials/data/hawc.py
LauraOlivera: no, the background in HAWC is usually very well characterized and so there is in principle no need for this correction (if the background is estimated ok)

examples/tutorials/data/hawc.py
AtreyeeS: yes indeed! stupid question, sorry

AtreyeeS:
> The reason I didn't is that the other tutorials in the "data" section are quite restricted to opening and exploring the data, and so I wasn't sure if it made sense to have the full HAWC analysis when once you have `MapDatasets` it is in principle like any other instrument. But if you think it would be better, I can easily add it (maybe in a later PR?)

Yes, it is like any other instrument, but there are these nuances that you dont need to fit the background, you need to do a joint fit, etc. The other `data` tutorials are smaller because there is no difference between HESS or CTA analysis! 
I think it would be nice to have the fitting included to avoid confusion. Of course, you are best equipped to decide how many PRs you would want to split it into.

registerrier (approved):
Thanks @LauraOlivera ! This looks good.

PR #4468, opened 2023-04-13
Serialisation of MapAxes

AtreyeeS:
In the present version, `MapAxes.from_table_hdu()` tries to instantiate different axes, and keeps the one which does not fail https://github.com/gammapy/gammapy/blob/d98b5b2b547a87270ead6d13c99359fba8d2da94/gammapy/maps/axes.py#L1997

This is annoying, see: https://github.com/gammapy/gammapy/blob/18c2597f23816f465189f6dde4e6e75fd9e50fe6/gammapy/modeling/models/temporal.py#L595

This PR introduces a AXNODE keyword in the header, and the axis type is chosen accordingly. For older files, the guess work still remains.

In a broader context, we should decide on the `to_table` and `from_table` behaviour of the map classes in general, eg

1. The arguments in to and from table are not always the same (eg: `RegionNDMap`)
2. Support of the `ogip` format. (Its the default in `MapAxis.to_table`, why?)

### DIFF
--- a/gammapy/maps/axes.py
+++ b/gammapy/maps/axes.py
@@ -13,7 +13,7 @@ from astropy.utils import lazyproperty
 import matplotlib.pyplot as plt
 from gammapy.utils.interpolation import interpolation_scale
 from gammapy.utils.time import time_ref_from_dict, time_ref_to_dict
-from .utils import INVALID_INDEX, edges_from_lo_hi
+from .utils import INVALID_INDEX, edges_from_lo_hi, guess_axis
 
 __all__ = ["MapAxes", "MapAxis", "TimeMapAxis", "LabelMapAxis"]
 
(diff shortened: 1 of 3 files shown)


adonath:
Thanks @AtreyeeS! One first quick comment: the addition of a new header keyword would require a change to GADF and first be proposed there...do you have any suggestion @registerrier how to handle this?

adonath:
From the dev call today: we agreed to emit custom errors for the cases where the axes cannot be parsed correctly. This cleans up the current implementation.

AtreyeeS:
If and when GADF proposes an addition of a header keyword, this PR can be re-opened

adonath:
@AtreyeeS I think you should actually propose this is in https://github.com/open-gamma-ray-astro/gamma-astro-data-formats, but recently there was just not much activity. But I'd be happy to support this addition...

PR #4475, opened 2023-04-18
Set a default Xmax for plot_offset2_distribution

bkhelifi:
**Description**

This pull request fixes the fact that the event.peek shows an Offset distribution with only 1 bin:

Capture d’écran du 2023-04-18 17-19-27

This is due to badly reconstructed events at very large offset.

**Dear reviewer**
In this PR, I set a default xmax to the histogram (98% quantile). 

The issue of the latex rendering will be fixed in the PR #4428 .

### DIFF
--- a/gammapy/data/event_list.py
+++ b/gammapy/data/event_list.py
@@ -552,9 +552,11 @@ class EventList:
             center = self._plot_center
 
         offset2 = center.separation(self.radec) ** 2
+        max2 = np.percentile(offset2, q=98)
 
         kwargs.setdefault("histtype", "step")
         kwargs.setdefault("bins", 30)
+        kwargs.setdefault("range", (0.0, max2.value))
 
         with quantity_support():
             ax.hist(offset2, **kwargs)
--- a/gammapy/data/event_list.py
+++ b/gammapy/data/event_list.py
@@ -489,7 +489,9 @@ class EventList:
 
         return ax
 
-    def plot_offset2_distribution(self, ax=None, center=None, **kwargs):
+    def plot_offset2_distribution(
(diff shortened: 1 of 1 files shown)


gammapy/data/event_list.py
```diff
@@ -552,9 +552,11 @@ def plot_offset2_distribution(self, ax=None, center=None, **kwargs):
             center = self._plot_center
 
         offset2 = center.separation(self.radec) ** 2
+        max2 = np.percentile(offset2, q=98)
```
adonath, line 555: I think it would be good to make the quantile a parameter to `plot_offset2_distribution()`. I don't expect it to be changed much, but just in case it good to avoid the hard-coded number...

registerrier (commented):
Thanks @bkhelifi . I am mixed here, since `peek` directly shows that there are some outliers by default. It is easy to avoid this with a `select_offset`.

Note that one could also apply a similar correction to the event distribution

bkhelifi:
`peek` directly shows that there are some outliers by default, agree.
In other hand, showing outlaiers might be not the top goal of the `peek` function ;) 
I would prefer that it provides an insight of the distribution (shape, counts)

adonath (approved):
Thanks @bkhelifi, no further comments from my side!

AtreyeeS (approved):
Thanks @bkhelifi ! I kind of agree with @registerrier that it is nice to know if there are outliers, but at the same time, having empty ugly plots on `peek` is rather annoying (specially during hands on sessions!). So finally +1 to merge this PR

adonath:
Thanks @AtreyeeS!

PR #4483, opened 2023-04-21
Suppress some matplotlib outputs in tutorials

bkhelifi:
**Description**

This pull request is associated to the Issue #4327. It adds semicolons when necessary

**Dear reviewer**
During my commits, I have noticed that `black` automatically suppresses some semicolons. One might need to configure back if possible (a new issue?)

### DIFF
--- a/examples/tutorials/analysis-1d/cta_sensitivity.py
+++ b/examples/tutorials/analysis-1d/cta_sensitivity.py
@@ -179,7 +179,8 @@ ax.plot(
 ax.loglog()
 ax.set_xlabel(f"Energy [{t['e_ref'].unit.to_string(UNIT_STRING_FORMAT)}]")
 ax.set_ylabel(f"Sensitivity [{t['e2dnde'].unit.to_string(UNIT_STRING_FORMAT)}]")
-ax.legend();
+ax.legend()
+plt.show()
 
 
 ######################################################################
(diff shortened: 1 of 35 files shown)


adonath (dismissed):
Thanks @bkhelifi, no comments from my side.

AtreyeeS (changes requested):
Thanks @bkhelifi ! Can you also please adapt the following notebooks while you are at it? Then we can close the issue 
- analysis-1d/sed_fitting.py
- analysis-3d/analysis_3d.py, analysis-3d/analysis_mwl.py
- analysis-time/light_curve_flare.py
- api/fitting.py
- data/*

adonath:
@bkhelifi I just realized that the problem with `black` is maybe a non-issue. Let me explain:

- As we want our tutorials to work with notebooks as well as scripts. We always have to call `plt.show()`, when a plot is ready to be shown. 
- This means all plotting cells should end with a `plt.show()` statement
- It seems the cases where we end e.g. with `plt.legend()` miss the `plt.show()` statement anyway.
- In cases where the definition of a plot is split into multiple cells it should be merged into a single one. 
- Once all the plotting cells end correctly with `plt.show()`, it should suppress the matplotlib output as well as be compliant with `black`

AtreyeeS (dismissed):
Thanks a lot @bkhelifi ! Its noteworthy how many typos survived in our docs after the extensive 1.0 checking !!

adonath (dismissed):
Thanks @bkhelifi, I have no further comments. I just need to double check whether this affects any other open PR.

AtreyeeS:
Hi @bkhelifi ! Can you please add a `plt.show` after the `.peek()` in `cta.py`
This is specially necessary for `irfs["psf"].peek()` , otherwise the results of `plot_containment_radius_vs_energy` and the last axis of peek merge to give a very confusing plot as in:  https://docs.gammapy.org/1.0.1/tutorials/data/cta.html#point-spread-function

QRemy (approved):

Issue #4480, opened 2023-04-21
Support additional axes in SkyModel.evaluate_geom(geom)


registerrier:
**Is your feature request related to a problem? Please describe.**
Currently, `SkyModel.evaluate_geom(geom)` does not support axes that are not `energy_true`.  Calling it on a `Geom` object with additional axes will return an array with incorrect shape.

**Describe the solution you'd like**
`SkyModel.evaluate_geom(geom)` should return an array of the correct size copying the contents for the additional axes.

Special caution should be taken for `time` axis. For three reasons:
- currently the time is evaluated via the `gti` argument. This is an incorrect approach as `gti` is relevant on the `MapEvaluator`, not on the model. Note that a `TimeMapAxis.from_gti()` factory method exists. 
-  `evaluate_geom` actually integrates over time. Is this still necessary? Does it make sense?
- A `MapAxis` can represent a time-like quantity but does not contain a reference time. This has to be verified. 

Another question, is how to handle `energy` axis (instead of `energy_true`).

Astro-Kirsty:
This should be closed by #4828 ?

PR #4485, opened 2023-04-21
Implement TimeMapAxis.pix_to_coord

AtreyeeS:
I am not happy with this implementation, mainly because
- What to do when the pixel is outside the range? For normal `MapAxis`, we extrapolate, but here that is not feasible. I am raising a `ValueError`. Maybe it would be better to return `nan`, but instantiating a `Time` object with `nan` value does not work
- The `for` loop is ugly! Suggestions for improvement please!


adonath:
>  but instantiating a Time object with nan value does not work

I double checked and indeed this does no work. Time values are required to be finite. I'm not sure why this is the case, because e.g. `Quantity` handles `np.nan` values just fine.

gammapy/maps/axes.py
```diff
@@ -2485,6 +2485,34 @@ def coord_to_idx(self, coord, **kwargs):
...
+        interp_scale = interpolation_scale(self.interp)
+        for p in np.atleast_1d(pix):
+            if p < 0.0 or p > self.nbin:
+                raise ValueError("Pixel is outside range")
+            ipix = int(np.floor(p))
+            x = [ipix, ipix + 1]
+            y = [self.time_min[ipix].mjd, self.time_max[ipix].mjd]
+            interp_fn = scipy.interpolate.interp1d(x=x, y=y, kind=1)
+            val = interp_scale.inverse(interp_fn(p))
+            coords.append(Time(val, scale=self.reference_time.scale, format="mjd"))
```
registerrier, line 2513: I think it should be simpler to build the array of time-like quantities and then add the reference time.

AtreyeeS: I can use `self.edges_max` instead of `time_max`, but since we want to return a `Time` object, `nan` will still not work. And `int(x)` does not work if `x` is an array, so I used the loop...

AtreyeeS:
I modified the value error to use np.ma.masked

adonath:
@AtreyeeS In the dev call we quickly discussed this and agreed it would be good to only support the contiguous case for now. This avoid the problem of missing values.

AtreyeeS:
If we support only the contiguous case, implementing  #4480 will be complex.
The time_axis that we encounter is mostly non-contiguous.

The problem of missing value comes in the case of extrapolation, otherwise pixel values must be contiguous (missing pixels not possible, right?). eg: Asking for a coord at `pix=20` when the axis has 15 pixels. But this case is anyways non-physical, and we want it fail in a decent way, no?

adonath:
Thanks @AtreyeeS, so it seems we are limited by the Astropy time handling. We could define a placeholder value  for invalid times, such as year 42 before Christ. We have similar definitions for other data types: https://github.com/gammapy/gammapy/blob/main/gammapy/maps/utils.py#L77

maxnoe:
In ctapipe, we now use mjd=0 as invalid time marker, until I can come back to Solling the astropy issue.

The conclusion of that issue was that it's ok to add, but I couldn't finish the PR and it Slipper my mind for a while.

registerrier (commented):
Thanks @AtreyeeS . This is an important addition.

Indeed the loop over pix to recreate an 1d interpolator is likely inefficient. I suppose you can directly obtain the same behaviour by computing fractional part and index directly with e.g. `np.modf`.

Then something like:
```
frac, idx = np.modf(pix)
valid = ~np.logical_or(idx<0, idx>=self.nbin)
coords[valid] = frac[valid]*self.time_delta[idx[valid]] + self.edges_min[valid]
```

AtreyeeS:
Thanks @registerrier ! I adapted as per your suggestions. This will work only for time axis with linear interp, but since that is what we support at the moment, I guess it is okay.

In future, we would have to support other interpolation scheme I think, to support GRB studies...

gammapy/maps/axes.py
```diff
@@ -2485,6 +2484,33 @@ def coord_to_idx(self, coord, **kwargs):
         idx[~np.any(mask, axis=-1)] = INVALID_INDEX.int
         return idx
 
+    def pix_to_coord(self, pix):
+        """Transform from pixel position to time coordinate
+        Currently works only for lin inperp
+        Parameters
```
registerrier, line 2490:
```suggestion
        Currently works only for linear interpolation scheme.
        
        Parameters
```

registerrier (approved):
Thanks @AtreyeeS . This looks good. I have left a small inline comment.

Regarding interpolation scheme, I suppose we can apply the scaling function when computing from `frac` to `coord.`

registerrier (approved):
Thanks @AtreyeeS . No further comments.

PR #4486, opened 2023-04-21
Scale handling in temporal models

AtreyeeS:
Further fixes to #4453

### DIFF
--- a/gammapy/modeling/models/tests/test_temporal.py
+++ b/gammapy/modeling/models/tests/test_temporal.py
@@ -299,10 +299,11 @@ def test_generalized_gaussian_temporal_model_integral():
     )
     start = 1 * u.day
     stop = 2 * u.day
-    t_ref = Time(50000, format="mjd")
+    t_ref = Time(50000, format="mjd", scale="utc")
     gti = GTI.create(start, stop, reference_time=t_ref)
+    assert gti.time_start.scale == t_ref.scale
     val = temporal_model.integral(gti.time_start, gti.time_stop)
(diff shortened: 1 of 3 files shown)


registerrier (commented):

My main comment would be not to touch `GTI` in this PR. There os another PR re-factoring it, and failing at the same place as here.

AtreyeeS:
Thanks @registerrier ! I removed the use of the GTI class.
It would still be good to ensure that the scale handling in GTI is fixed before the next release (maybe even a bug fix?), because `GTI.create` silently changes the scale

gammapy/modeling/models/tests/test_temporal.py
```diff
@@ -299,10 +291,11 @@ def test_generalized_gaussian_temporal_model_integral():
...
     stop = 2 * u.day
-    t_ref = Time(50000, format="mjd")
-    gti = GTI.create(start, stop, reference_time=t_ref)
-    val = temporal_model.integral(gti.time_start, gti.time_stop)
-    assert_allclose(val, 0.759115, rtol=1e-4)
+    t_ref = Time(50000, format="mjd", scale="utc")
+    time_start = t_ref + start
+    time_stop = t_ref + stop
+    val = temporal_model.integral(time_start, time_stop)
+    assert_allclose(val, 0.758918, rtol=1e-4)
```
registerrier, line 298: The difference here is due to confusion between `utc` and `tt` right?

registerrier (commented):
Thanks @AtreyeeS . This looks good. Maybe add a test case where times are with a different scale than the reference time?

PR #4491, opened 2023-04-27
PIG 27 - Metadata structure

registerrier:
**Description**

This pull request introduces the PIG on metadata structure.

**Dear reviewer**


docs/development/pigs/pig-028.rst
```diff
@@ -0,0 +1,221 @@
...
+agregates the `meta_table` of its members. After estimation, the ``FluxPoints`` don't
+contain any specific meta information.
+
+The algorithm classes (`Makers` and `Estimators`) don't contain any meta information so far.
+This might be an issue since some information  could be transfered on their various products
+as well.
+
+A minimal information that needs to be present on every Gammapy product and serialized
+in various formats is the `CREATOR` which is the software and software version used,
+as well as the `DATE` when the object was created, and possibly the `ORIGIN` (the user,
```
AtreyeeS, line 42: Maybe also the data models version? ie, GADF2.0, etc? This might be useful for the DL4/DL5 products (even though we obviously dont have a formal definition yet)

adonath:
Thanks a lot @registerrier! I think this is very convincing approach and will lead to much improved meta data handling for Gammapy. Here are some first smaller comments:

- I think we should emphasize that the meta data representation is independent of the serialization format. So it is always a superset of what can be serialized.
- Maybe also emphasize that meta data is optional to tun the code. So we try to minimize the places where we rely on the information in the code. In almost all cases it is optional, but there might be a handful of cases where it is required. Maybe we can identify those beforehand.
- I think we should come up with a list of proposed meta classes. Typically associated with our existing data classes. And we should also clarify that meta data will get it's get it's own namespace, like `MapDataset.meta` or `IRF.meta`
- What is less clear to me is whether we need a generic container on the `MetaData` class to store un-validated data? In general we do not really want this, but there is the case of round-tripping unknown meta data. My proposed answer here is no. So we only read and write known, validated meta data.

registerrier:
> * I think we should emphasize that the meta data representation is independent of the serialization format. So it is always a superset of what can be serialized.

OK

> * Maybe also emphasize that meta data is optional to tun the code. So we try to minimize the places where we rely on the information in the code. In almost all cases it is optional, but there might be a handful of cases where it is required. Maybe we can identify those beforehand.

I don't whether we can fully address the question, since some processing information might not be represented by a regular attribute of the class. For instance, the phase selection information will be needed to properly compute a pulsar flux or build a phasogram. 
The `Datasets.select` could also use meta informations on the datasets.
Probably the issue is mostly for products such `Dataset` or `FluxMaps`.

> * I think we should come up with a list of proposed meta classes. Typically associated with our existing data classes. And we should also clarify that meta data will get it's get it's own namespace, like `MapDataset.meta` or `IRF.meta`

Yes, indeed. You mean proposing a naming scheme and a list of foreseen classes, but not their actual content, right?

> * What is less clear to me is whether we need a generic container on the `MetaData` class to store un-validated data? In general we do not really want this, but there is the case of round-tripping unknown meta data. My proposed answer here is no. So we only read and write known, validated meta data.

It still might be useful to leave the option to add some meta information to the user. Of course, it is not clear how these informations can then be serialized.

There is already some flexibility in e.g. g.a.d.f. event header keywords. Many are optional, but they should all be present in the `EventList.meta`. We can validate them when reading from file keeping only documented entries with a validation scheme for the I/O methods. But different formats will come with different lists of optional and non-optional meta information.  Do we want to have meta objects like `GADFEventMetaData` or `VODFEventMetaData` ?

adonath:
> I don't whether we can fully address the question, since some processing information might not be represented by a regular attribute of the class. For instance, the phase selection information will be needed to properly compute a pulsar flux or build a phasogram. The Datasets.select could also use meta informations on the datasets.
Probably the issue is mostly for products such Dataset or FluxMaps.

Yes, I agree there are some cases (but not many) were the meta data is required. I think it would be good to identify these cases (our best guess) and list them in the PIG.

> Yes, indeed. You mean proposing a naming scheme and a list of foreseen classes, but not their actual content, right?

Yes, exactly. Just the list of class names no content yet. Once we have a namespace for the meta data, we can be very liberal in adding more meta data. Meta data memory I think it is rarely controversial and we can in most cases just have more of. This is less clear for serialization. But this is not up to us to decide.

> Do we want to have meta objects like GADFEventMetaData or VODFEventMetaData ?

That's an interesting question. My initial thought was to just have `EventListMetaData` and then corresponding serialization methods to and from multiple formats. However classes like `GADFEventMetaData` might be a good way to maintain the serialization formats. Right now I don't have a clear answer yet.

bkhelifi:
Hi all,
The PIG does not deal with the data model of the meta data, as stated in the text. However, one could introduce the motion of the need of a "meta data reduction", making the parallel with the "data reduction". Indeed, this concept will be important in order to not create an inflation of meta data and also to make the distinction between meta data and provenance data.
Should we evoke this concept here, as it could be made into functions of this class? Maybe into the requirement section...

docs/development/pigs/pig-028.rst
```diff
@@ -0,0 +1,298 @@
...
+  A single generic class could be used for all actual IRFs.
+
+Makers
+------
+
+It is unclear whether stateless algorithm classes such as ``Maker`` actually need meta
+information beyond their actual attributes. They will have to create or update `meta`
+information of the ``Dataset`` they create or modify. For now, we don't propose
+any metadata for ``Maker`` objects.
+
```
bkhelifi, line 257: I think that it might be important to store the options of the makers (e.g. fit or scale for the FoVBkgMaker).
However, these information could be classified as provenance information... or not. Maybe this is a point to discuss.

docs/development/pigs/pig-028.rst
registerrier: One could assume that the internal data format can be deduced from the gammapy version. I don't know whether we want to introduce an additional versioning.
Regarding serialization data formats, I think the IO code should be the place where the correct keywords are defined.

registerrier:
> Hi all, The PIG does not deal with the data model of the meta data, as stated in the text. However, one could introduce the motion of the need of a "meta data reduction", making the parallel with the "data reduction". Indeed, this concept will be important in order to not create an inflation of meta data and also to make the distinction between meta data and provenance data. Should we evoke this concept here, as it could be made into functions of this class? Maybe into the requirement section...

Indeed, metadata generation and propagation in general will have to defined. We can add a paragraph to evoke the problem. Here is a proposal:
```
DL3 products should come with their pre-defined metadata (unless generated by Gammapy for instance during event simulations). But all all other data levels will have metadata generated by gammapy.
Algorithm classes (Makers, Estimators) produce new data containers (DL4 and DL5), they will generate new metadata to be stored on the container and will propagate some of the metadata from the lower level products they manipulate. What metadata will be passed or discarded, how metadata will be restructured in this process (i.e. how propagation and reduction will be performed) is beyond the scope of this PIG. For now, the important point is that metadata handling becomes a task of algorithm classes. The actual definition of the metadata classes will have to support the propagation and reduction process.
An obvious case is `Dataset` stacking. The associated `meta` class will have to support the stacking mechanism.    
```

bkhelifi:
Perfect! Thanks
Here a minor correction:
"But all all other data levels will have metadata generated by gammapy." -> "But all other data levels will have metadata generated by Gammapy."

bkhelifi (approved):
Thanks a lot @registerrier . Looks OK for me

PR #4495, opened 2023-05-03
Move axis

registerrier:
**Description**

This pull request is a follow-up of #4294 which implement `Map.move_axis`. 

I have not added the `Map.insert_axis(idx, axis)` functionality. I woul assume the latter requires a `MapAxes.insert()` method, but this would make `MapAxes` a mutable sequence. What do you think @adonath ?

**Dear reviewer**

### DIFF
--- a/gammapy/maps/core.py
+++ b/gammapy/maps/core.py
@@ -1911,15 +1911,29 @@ class Map(abc.ABC):
 
         return MapCoord.create(cdict, frame=self.geom.frame)
 
-    def move_axis(self, axis):
-        """
+    def move_axis(self, axes_names):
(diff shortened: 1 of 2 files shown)


adonath (dismissed):
Thanks a lot @registerrier, I have no comments!

adonath:
> I have not added the Map.insert_axis(idx, axis) functionality. I woul assume the latter requires a MapAxes.insert() method, but this would make MapAxes a mutable sequence. What do you think @adonath ?

It does not have to work this way. Typically we follow the Numpy API, which has https://numpy.org/doc/stable/reference/generated/numpy.insert.html it does return a copy...

QRemy (approved):
Thanks @registerrier, looks good !

Issue #4499, opened 2023-05-10
SpectrumDatasetOnOff.stat_sum fails when counts_off is None


Astro-Kirsty:
**Gammapy version**
version 1.0

**Bug description**
Fit fails for a given collection of `SpectrumDatasetOnOff` when one dataset has no `counts_off`.
```DatasetOnOff.stat_array(self)
   2116 """Statistic function value per bin given the current model parameters"""
   2117 mu_sig = self.npred_signal().data
   2118 on_stat_ = wstat(
   2119     n_on=self.counts.data,
-> 2120     n_off=self.counts_off.data,
   2121     alpha=list(self.alpha.data),
   2122     mu_sig=mu_sig,
   2123 )
   2124 return np.nan_to_num(on_stat_)

AttributeError: 'NoneType' object has no attribute 'data'
```

**Expected behavior**
The Fit should ignore the dataset (the `mask_safe` is set to false by the reflected regions maker)

**To Reproduce**
```from gammapy.datasets import SpectrumDatasetOnOff
from gammapy.maps import RegionGeom, MapAxis

axis = MapAxis.from_energy_bounds(0.1, 10, 5, unit='TeV')
geom = RegionGeom.create(None, axes=[axis])
dataset = SpectrumDatasetOnOff.create(geom)
dataset.counts_off = None
dataset.stat_sum()
```

PR #4500, opened 2023-05-12
Fix SpectrumDatasetOnOff.stat_sum to support when counts_off is None

Astro-Kirsty:
**Description**
This pull request fixes the bug from Issue  #4499
`stat_sum` in `SpectrumDatasetOnOff` has been adapted to account for this issue

### DIFF
--- a/gammapy/datasets/core.py
+++ b/gammapy/datasets/core.py
@@ -64,7 +64,7 @@ class Dataset(abc.ABC):
     def stat_sum(self):
         """Total statistic given the current model parameters."""
 
-        if self.counts_off is None and np.all(self.mask_safe.data) is False:
+        if self.counts_off is None and not np.any(self.mask_safe.data) is False:
             return 0
 
         stat = self.stat_array()
--- a/gammapy/datasets/core.py
+++ b/gammapy/datasets/core.py
@@ -63,6 +63,10 @@ class Dataset(abc.ABC):
 
     def stat_sum(self):
         """Total statistic given the current model parameters."""
+
+        if self.counts_off is None and np.all(self.mask_safe.data == False):
+            return 0
+
         stat = self.stat_array()
 
         if self.mask is not None:
--- a/gammapy/datasets/core.py
+++ b/gammapy/datasets/core.py
@@ -64,7 +64,7 @@ class Dataset(abc.ABC):
     def stat_sum(self):
         """Total statistic given the current model parameters."""
 
-        if self.counts_off is None and np.all(self.mask_safe.data == False):
+        if self.counts_off is None and np.all(self.mask_safe.data) is False:
             return 0
 
         stat = self.stat_array()
--- a/gammapy/datasets/core.py
+++ b/gammapy/datasets/core.py
@@ -63,10 +63,6 @@ class Dataset(abc.ABC):
 
     def stat_sum(self):
         """Total statistic given the current model parameters."""
-
-        if self.counts_off is None and not np.any(self.mask_safe.data) is False:
-            return 0
-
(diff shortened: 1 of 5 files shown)


gammapy/datasets/core.py
```diff
@@ -63,6 +63,10 @@ def mask(self):
 
     def stat_sum(self):
         """Total statistic given the current model parameters."""
+
+        if self.counts_off is None and not np.any(self.mask_safe.data) is False:
```
registerrier, line 67: In pratice, this is causing because `counts_off` does not belong to all the `Dataset` subclasses. This change has to be applied on the `MapdatasetOnOff` instead.

registerrier (changes requested):
Thanks @Astro-Kirsty . This requires a change because the correction should be applied on the subclass that contains `counts_off` i.e. `MapDatasetOnOff`

gammapy/datasets/core.py
Astro-Kirsty: Okay, thanks. I will adapt and push

registerrier (commented):
Thanks @Astro-Kirsty . Last comment: the test you've added is part of test_map.py. Since it touches mostly `SpectrumDatasetOnOff`, I would suggest to move it to test_spectrum.py instead.

gammapy/datasets/tests/test_map.py
```diff
@@ -1909,3 +1909,11 @@ def test_peek(images):
...
     dat = MapDataset.create(geom, reco_psf=True)
     assert isinstance(dat.psf, RecoPSFMap)
+
+
+def test_stat_sum():
+    axis = MapAxis.from_energy_bounds(0.1, 10, 5, unit="TeV")
+    geom = RegionGeom.create(None, axes=[axis])
+    dataset = SpectrumDatasetOnOff.create(geom)
+    dataset.counts_off = None
+    return dataset.stat_sum()
```
registerrier, line 1919:
```suggestion
    stat = dataset.stat_sum()
    assert stat == 0
```
Actually a test should assert something. (and run obviously).

gammapy/datasets/map.py
```diff
@@ -2375,6 +2375,8 @@ def stack(self, other, nan_to_num=True):
 
     def stat_sum(self):
         """Total statistic function value given the current model parameters."""
+        if self.counts_off is None and not np.any(self.mask_safe.data) is False:
```
registerrier, line 2378:
```suggestion
        if self.counts_off is None and not np.any(self.mask_safe.data):
```

registerrier (commented):
See two additional comments:
- the test misses an `assert` to check the result is what is expected
- the condition in `stat_sum` probably not correct
The latter point suggest that a test is missing to make sure the `stat_sum` still fails where it should.

I suggest that you add a few lines of code in your test where you modify the `mask_safe` so that it is no longer completely False and check that `stat_sum` is failing as it should: 
```
dataset.mask_safe[0] = True
pytest.raises(ValueError):
    dataset.stat_sum()
```

registerrier (commented):
See two additional comments:
- the test misses an `assert` to check the result is what is expected
- the condition in `stat_sum` probably not correct
The latter point suggest that a test is missing to make sure the `stat_sum` still fails where it should.

I suggest that you add a few lines of code in your test where you modify the `mask_safe` so that it is no longer completely False and check that `stat_sum` is failing as it should: 
```
dataset.mask_safe[0] = True
pytest.raises(ValueError):
    dataset.stat_sum()
```

AtreyeeS (dismissed):
Thanks @Astro-Kirsty ! I think the behaviour here can be documented in the doc-string

PR #4504, opened 2023-05-12
Deprecate Fit.minuit member

adonath:
**Description**

This pull request deprecates the `Fit.minuit` member and moves it to the `OptimizeResult` instead. The motivation is to get rid of the stateful dependency of the `Fit` object, which is likely the reason for the remaining test fails in #4406.

### DIFF
--- a/gammapy/modeling/fit.py
+++ b/gammapy/modeling/fit.py
@@ -270,7 +270,7 @@ class Fit:
 
         kwargs = self.covariance_opts.copy()
 
-        if optimize_result is not None:
+        if optimize_result is not None and optimize_result.backend == "minuit":
             kwargs["minuit"] = optimize_result.minuit
 
         backend = kwargs.pop("backend", self.backend)
--- a/gammapy/modeling/fit.py
+++ b/gammapy/modeling/fit.py
@@ -642,6 +642,11 @@ class FitResult:
 
         self._covariance_result = covariance_result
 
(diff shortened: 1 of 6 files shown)


gammapy/modeling/fit.py
```diff
@@ -249,6 +256,9 @@ def covariance(self, datasets):
         ----------
         datasets : `Datasets` or list of `Dataset`
             Datasets to optimize.
+        optimize_result : `OptimizeResult`
+            Optimization result. Can be optionally used to pass the state of the IMinuit object
+            to the convariance estimatiomn. This might save computation time in certain cases.
```
registerrier, line 261:
```suggestion
            to the covariance estimation. This might save computation time in certain cases.
```

registerrier (dismissed):
Thanks @adonath . Looks good. No comment from my side except for docstring typos.

registerrier (commented):
Looking at it again, the deprecation scheme is not really respected since the `_iminuit` attribute is no longer set. 
Is it possible to keep but not use it in tests? Will this prevent the multiprocessing to run properly?

adonath:
@registerrier Good point! No, I think I can just add it back...

adonath:
@registerrier I added test to access the private member `._minuit`.

QRemy (approved):
Thanks @adonath, no comments from my side.

PR #4507, opened 2023-05-15
Update dev docs for sphinx gallery

adonath:
**Description**

This pull request update the developer docs for the Sphinx gallery and tutorials. It mostly remove the custom notebook handling and just refers to sphinx gallery. It still provides some minimal examples for the most common use cases.

### DIFF
--- a/docs/development/doc_howto.rst
+++ b/docs/development/doc_howto.rst
@@ -21,8 +21,6 @@ Or one can equivalently use tox::
 Generating the PDF docs is more complex.
 This should work::
 
-    # build the notebooks
-    python -m gammapy.utils.notebooks_process
     # build the latex file
     cd docs
     python -m sphinx . _build/latex -b latex -j auto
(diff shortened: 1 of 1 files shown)


adonath:
Do we need to back port this? @registerrier

registerrier (approved):

I suppose this can be backported as the current v1.0.1 dev doc is not valid for the notebooks.

I suppose another element that will have to be added is an how to convert a notebook into a python tutorial.

adonath:
> I suppose another element that will have to be added is an how to convert a notebook into a python tutorial.

Yes, good point. The options we have for this are `jupytext` or a custom script we would need to provide. In any case I'd propose to do this in a separate PR.

MRegeard:
I think we might need to go with a custom script (something like this : https://gist.github.com/chsasank/7218ca16f8d022e02a9c0deb94a310fe). 
`jupytext` does not support conversion to `sphinx-gallery` format. 
I actually could not find anything that convert from markdown to sphinx-gallery.

PR #4508, opened 2023-05-15
Implement WcsNDMap.to_region_nd_map_histogram

adonath:
**Description**

This pull request implements `WcsNDMap.to_region_nd_map_histogram`. The idea is as following:

- `WcsNDMap.to_region_nd_map_histogram` creates a `RegionNDMap` with a bins axis to present a histogram.
- This way it allows to represent "ND histograms": so a histogram for each extra axis in the map.
- This is mostly useful for plotting and serialization

### DIFF
--- a/gammapy/maps/wcs/ndmap.py
+++ b/gammapy/maps/wcs/ndmap.py
@@ -608,6 +608,83 @@ class WcsNDMap(WcsMap):
 
         return RegionNDMap(geom=geom, data=data, unit=self.unit, meta=self.meta.copy())
 
+    def to_region_nd_map_histogram(
+        self, region=None, bins_axis=None, nbin=100, density=False
+    ):
+        """Convert map into region map by histogramming.
+
+        By default it creates a linearly spaced axis with 100 bins between
+        (-max(abs(data)), max(abs(data))) within the given region.
+
+        Parameters
+        ----------
+        region: `~regions.Region`
+            Region to histogram over.
+        bins_axis : `MapAxis`
+            Binning of the histogram.
+        nbin : int
+            Number of bins to use if no bins_axis is given.
+        density : bool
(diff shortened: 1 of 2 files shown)


adonath:
CC @MRegeard, please test the functionality if you have time. E.g. you can start by using this code:

```python3
import numpy as np

random_state = np.random.RandomState(seed=0)

energy_axis = MapAxis.from_energy_bounds("1 TeV", "10 TeV", nbin=3)

data = Map.create(axes=[energy_axis], width=10, unit="cm2 s-1", binsz=0.02)
data.data = random_state.normal(
    size=data.data.shape, loc=0, scale=np.array([1.0, 2.0, 3.0]).reshape((-1, 1, 1))
)

hist = data.to_region_nd_map_histogram()
hist.plot(axis_name="bins")
```
Which should create the following plot:

image

adonath:
This turned out to be a bit more complex than expected. There is also some code duplication with `.to_region_nd_map()`, I'm happy for any suggestions. Maybe we could introduce a shared method, like `Map.cutout_region()`, that cutout the map based on some region definition and masks out values, that are not inside the region or similar...

MRegeard:
Let me check your code.

adonath:
@MRegeard I think for fitting we also need a solution to represent the un-binned data. So this is one more reason to introduce something like `Map.cutout_and_mask_region()`. I can propose another PR...

adonath:
See #4510.

adonath:
Will rebase after #4510 has been merged.

registerrier (dismissed):
I suppose the code that was merged in #4510 should be removed from this PR.

I have left one suggestion to try to build the histogram with a `fill_by_coord`

registerrier (approved):
Thanks @adonath . Looks good

PR #4514, opened 2023-05-19
Patch plot npred

MRegeard:
I changes the way the if statement works in `plot_npred_signal`. 
Before I was looking at the length of `dataset.models` but often it contains the background model, however the background model is plotted separately  from the "source models". So instead I have to look at the number of bins in the `LabelMapAxis` that contains the model names. 

I also move the stacking of the models to create the npred_stack inside the if statement.


adonath (approved):
Thanks, @MRegeard!

PR #4520, opened 2023-05-23
Support Astropy 5.3

adonath:
**Description**

This pull request modifies out tests to support Astropy 5.3, where the default unit order changed when printing. For this it introduces a `modify_unit_order_astropy_5_3` helper method, which will remain until we support older version than 5.3.


registerrier (approved):
Thanks @adonath . This looks good! No comments.

adonath:
@registerrier There is the remaining test fail in the sphinx build...I'll look into this now.

adonath:
Right no I cannot reproduce the error locally...I'll go ahead and merge and come back to it later.

PR #4521, opened 2023-05-23
Add covariance copy to support ray

adonath:
**Description**

This pull request is simpler alternative to #4516.  It includes the following changes:

- Most commits by @QRemy from #4516 
- Adding `covariance_data` to `Models`
- Some minor docstring additions
- Always making a copy of `_data` in `Covariance.set_subcovariance`

The latter is slightly inefficient, but when using ray I think there is basically no way to avoid this. If there are performance concerns we could introduce an option to control the copy behavior. 

Please review @QRemy and @registerrier.

### DIFF
--- a/gammapy/makers/tests/test_reduce.py
+++ b/gammapy/makers/tests/test_reduce.py
@@ -154,6 +154,13 @@ def makers_spectrum(exclusion_mask):
             "n_jobs": 2,
             "backend": "multiprocessing",
         },
+        {
+            "dataset": get_mapdataset(name="parallel_staking"),
+            "stack_datasets": True,
+            "cutout_width": None,
+            "n_jobs": 2,
+            "backend": "ray",
+        },
     ],
 )
 @requires_data()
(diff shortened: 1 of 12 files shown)


gammapy/makers/tests/test_reduce.py
```diff
@@ -154,9 +154,17 @@ def makers_spectrum(exclusion_mask):
...
+            "dataset": get_mapdataset(name="parallel_staking"),
+            "stack_datasets": True,
+            "cutout_width": None,
+            "n_jobs": 2,
+            "backend": "ray",
+        },
     ],
 )
 @requires_data()
+@requires_dependency("ray")
```
QRemy, line 167: Is it possible to have requires_dependency("ray") only for one of the parametrized tests ? or it has to be written separately ?

adonath: Probably it's possible. One can use `pytest.skip()` and an in-line if statement. But I think it is easier to just split out the test in this case. The benefit is that the test will not be collected, the test printout becomes more meaningful when the test is not skipped at execution.

registerrier (approved):
Why is the covariance handling necessary for the `TemplateSpatial Model`? It has no parameter, no?

QRemy (approved):
Thanks @adonath, I let you decide if it is worth or not to split the ray case from the parametrized test.

adonath:
@QRemy I'll open one follow up PR with some more clean up to the tests...

PR #4522, opened 2023-05-23
Fix MapDatasetEventSampler.sample_source

fabiopintore:
These PR fix a bug in `MapDatasetEventSampler.sample_source` when an energy-dependent temporal model is sampled.


adonath (approved):
Thanks @fabiopintore, test fails are unrelated.

PR #4529, opened 2023-05-25
Correct docstring in LightCurveEstimator

AtreyeeS:
I verified that the other estimator docs are corrected now


registerrier (approved):
Thanks @AtreyeeS . This is better to put the correct defaults... :)

adonath (approved):
Thank @AtreyeeS!

Issue #4470, opened 2023-04-18
Simple source detection function for FluxMaps 


registerrier:
**Is your feature request related to a problem? Please describe.**
The current find_peaks algorithm offers a minimalist set of data from a significance map. Much more can be extracted from `FluxMaps`: position in radec, and associated error, flux end flux error (in various sed type), excess counts. 
  
**Describe the solution you'd like**
A first approach could be to create a function in `gammapy.estimators.utils` to compute this.

**Additional context**
Note we have to understand whether this type of algorithm really belongs to the `Estimator` abstraction (DL4 to DL5). It is rather a tool extracting information from an estimate and building some higher level information. I am not sure this is really DL6 data, though, as this would rather be a list of source candidates without spectral information.

Similarly, the data format for such a product is not fully clear. At first a regular `Table` is fine. It could be partly converted to a `Models` object to allow for easy fitting.

adonath:
I agree this would be very useful. Here are just some links to existing functionality in `photutils`:

- https://photutils.readthedocs.io/en/stable/detection.html
- https://photutils.readthedocs.io/en/stable/segmentation.html
- https://photutils.readthedocs.io/en/stable/morphology.html
- https://photutils.readthedocs.io/en/stable/aperture.html

With a combination of the linked functionality above one could support extended sources as well or perform a simple deblending, segmentation and estimation of morphological properties.

Astro-Kirsty:
I have added a PR to address the initial request. 
To extract the sub-pixel location and therefore further features from the maps is more difficult. The method I have attempted is to utilise the astropy `modeling.functional_models.Gaussian2D `to fit the peaks (found through `find_peaks_in_flux_map`). In general, this works well when you have peaks which are separated by some distance. It would be good to discuss this further, and if anyone has suggestions to let me know.

registerrier:
Once this new funciton is exposed on detect.py notebook, we can close this issue.

Astro-Kirsty:
This functionality is now exposed, closing the issue

PR #4540, opened 2023-05-30
Add function to extract values from FluxMaps

Astro-Kirsty:
**Description**
This pull request is an initial function as the first approach for Issue #4470.
The function `find_peaks_in_flux_map` has been added to `gammapy.estimators.utils`, which utilises the methods from `find_peaks`.
The column names for the table are chosen to match the `FluxPointsEstimator` table column names. If anyone has better naming ideas, please let me know.

### DIFF
--- a/gammapy/estimators/utils.py
+++ b/gammapy/estimators/utils.py
@@ -127,7 +127,7 @@ def find_peaks_in_flux_map(maps, threshold, min_distance=1):
     ----------
     maps : `~gammapy.estimators.map.FluxMaps`
     threshold : float or array-like
-        The data value or pixel-wise data values to be used for the
+        The TS data value or pixel-wise TS data values to be used for the
         detection threshold.  A 2D ``threshold`` must have the same
         shape as the map ``data``.
     min_distance : int or `~astropy.units.Quantity`
--- a/gammapy/estimators/utils.py
+++ b/gammapy/estimators/utils.py
@@ -121,9 +121,7 @@ def find_peaks(image, threshold, min_distance=1):
 def find_peaks_in_flux_map(maps, threshold, min_distance=1):
     """Find local TS peaks for a given Map.
 
-    Simple peak finder, that finds local TS maximums using a sliding window
-    in a given map, above a given ``threshold`` within a given ``min_distance``
-    around each given pixel.
+    Utilises the find_peaks function to find various parameters from FluxMaps.
 
     Parameters
     ----------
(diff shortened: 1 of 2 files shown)


gammapy/estimators/utils.py
```diff
@@ -116,6 +118,99 @@ def find_peaks(image, threshold, min_distance=1):
...
+    # Remove non-finite values to avoid warnings or spurious detection
+    data = image.sum_over_axes(keepdims=False).data
+    data[~np.isfinite(data)] = np.nanmin(data)
+
+    # Handle edge case of constant data; treat as no peak
+    if np.all(data == data.flat[0]):
+        return Table()
+
+    # Run peak finder
+    data_max = scipy.ndimage.maximum_filter(data, size=size, mode="constant")
```
adonath, line 170: Why not just rely on `find_peaks`? You can later use the pixel positions to extract the values from the `npred` and excess maps. And just extend the returned table.

gammapy/estimators/utils.py
```diff
@@ -116,6 +118,99 @@ def find_peaks(image, threshold, min_distance=1):
...
+    flux_peaks = flux_data[0, y, x]
+    flux_err_data = maps["flux_err"].data
+    flux_err_peaks = flux_err_data[0, y, x]
+
+    table = Table()
+    table["x"] = x
+    table["y"] = y
+    table["ra"] = coord.ra
+    table["dec"] = coord.dec
+    table["sqrt_ts"] = sqrt_ts_values * image.unit
```
adonath, line 194: Is it possible to replace this by a loop over all `FluxMap` quantities? Note that some quantities are optional, but those would only be added if available.

adonath (changes requested):
Thanks a lot @Astro-Kirsty, this is a very useful feature! I have left some comment to simplify the implementation and make more use of already existing functionality.

gammapy/estimators/utils.py
Astro-Kirsty: I created a loop over the `available_quantities` as well as adding in the `flux` and `flux_err`. I guess the question is whether we want to loop over `all_quantities` or leave it as is?

registerrier (commented):
Thanks @Astro-Kirsty . 

If you want to extract all information from the `FluxMaps` object you can use the `get_flux_points()` method. See https://github.com/gammapy/gammapy/blob/e45cb44c8bfb6b3029e134c571ca0bdacb76cbfb/gammapy/estimators/map/core.py#L720
Interestingly, the results of your function could be also returned in a list of `FluxPoints`...

gammapy/estimators/utils.py
```diff
@@ -116,6 +118,80 @@ def find_peaks(image, threshold, min_distance=1):
...
 
+def find_peaks_in_flux_map(maps, threshold, min_distance=1):
+    """Find local TS peaks for a given Map.
+
+    Utilises the find_peaks function to find various parameters from FluxMaps.
+
+    Parameters
+    ----------
+    maps : `~gammapy.estimators.map.FluxMaps`
+    threshold : float or array-like
```
bkhelifi, line 129: The docstring does not state is the threshold is in TS... can you precise it?

gammapy/estimators/utils.py
```diff
@@ -116,6 +118,80 @@ def find_peaks(image, threshold, min_distance=1):
...
 
 
+def find_peaks_in_flux_map(maps, threshold, min_distance=1):
+    """Find local TS peaks for a given Map.
+
+    Utilises the find_peaks function to find various parameters from FluxMaps.
+
+    Parameters
+    ----------
+    maps : `~gammapy.estimators.map.FluxMaps`
```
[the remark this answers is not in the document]
Astro-Kirsty: Thanks @bkhelifi. 
It is a FluxMaps object which contains multiple quantities (i.e maps). Happy to change to be 'map' if you think it is confusing for the user

gammapy/estimators/utils.py
```diff
@@ -116,6 +118,80 @@ def find_peaks(image, threshold, min_distance=1):
...
+        peaks = values[0, y, x]
+        table[name] = peaks
+
+    flux_data = maps["flux"].quantity
+    table["flux"] = flux_data[0, y, x]
+    flux_err_data = maps["flux_err"].quantity
+    table["flux_err"] = flux_err_data[0, y, x]
+
+    for column in table.colnames:
+        if column.startswith(("flux", "flux_err")):
```
registerrier, line 173: Is there a specific reason for defining a different format for the flux columns?

registerrier (approved):
Thanks @Astro-Kirsty . This looks good. I have left some small inline suggestions.

gammapy/estimators/utils.py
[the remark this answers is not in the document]
Astro-Kirsty: Thanks @registerrier, I have implemented those changes

gammapy/estimators/utils.py
Astro-Kirsty: For readability of the flux values -- typical fluxes of 1e-13, for example

registerrier (approved):

registerrier:

PR #4546, opened 2023-06-01
Remove is_ul column in FluxPointsEstimator if no upper limit is defined

Astro-Kirsty:
**Description**
The default for `selection_optional` in `FluxPointsEstimator` is `None`. This means that the `is_ul` column shows all `False`. This current behaviour seems non-intuitive because it is not `False`, rather it has not been calculated. 

This PR proposes an alternative, such that if `has_ul` is `False`, then the `is_ul` column will be removed, as to avoid confusion for the user.

### DIFF
--- a/gammapy/estimators/points/core.py
+++ b/gammapy/estimators/points/core.py
@@ -9,7 +9,7 @@ from astropy.time import Time
 from astropy.visualization import quantity_support
 import matplotlib.pyplot as plt
 from gammapy.maps import MapAxis, Maps, RegionNDMap, TimeMapAxis
-from gammapy.maps.axes import flat_if_equal, UNIT_STRING_FORMAT
+from gammapy.maps.axes import UNIT_STRING_FORMAT, flat_if_equal
 from gammapy.modeling.models import TemplateSpectralModel
 from gammapy.modeling.models.spectral import scale_plot_flux
 from gammapy.modeling.scipy import stat_profile_ul_scipy
@@ -370,6 +370,8 @@ class FluxPoints(FluxMaps):
                 table["stat_scan"] = self.stat_scan.data[idx]
 
             table["is_ul"] = self.is_ul.data[idx]
+            if not self.has_ul:
+                table.remove_columns("is_ul")
 
         elif format == "lightcurve":
             time_axis = self.geom.axes["time"]
--- a/gammapy/estimators/points/tests/test_core.py
+++ b/gammapy/estimators/points/tests/test_core.py
@@ -338,6 +338,7 @@ def test_flux_points_plot_no_error_bar():
         _ = flux_points.plot(sed_type="dnde")
 
 
+@requires_data()
 def test_fp_no_is_ul():
     path = make_path("$GAMMAPY_DATA/tests/spectrum/flux_points/flux_points.fits")
     table = Table.read(path)
@@ -347,5 +348,4 @@ def test_fp_no_is_ul():
     fp = FluxPoints.from_table(table)
 
     if not fp.has_ul:
-        with pytest.raises(AttributeError):
-            table.is_ul
+        assert "is_ul" not in table.colnames
--- a/gammapy/estimators/points/tests/test_core.py
+++ b/gammapy/estimators/points/tests/test_core.py
@@ -6,9 +6,8 @@ import astropy.units as u
 from astropy.table import Table
 import matplotlib.pyplot as plt
 from gammapy.catalog.fermi import SourceCatalog3FGL, SourceCatalog4FGL
-from gammapy.estimators import FluxPoints, FluxPointsEstimator
+from gammapy.estimators import FluxPoints
 from gammapy.estimators.map.core import DEFAULT_UNIT
-from gammapy.estimators.points.tests.test_sed import simulate_map_dataset
 from gammapy.modeling.models import PowerLawSpectralModel, SpectralModel
 from gammapy.utils.scripts import make_path
 from gammapy.utils.testing import (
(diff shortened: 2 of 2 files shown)


gammapy/estimators/points/core.py
```diff
@@ -370,6 +370,8 @@ def to_table(self, sed_type=None, format="gadf-sed", formatted=False):
                 table["stat_scan"] = self.stat_scan.data[idx]
 
             table["is_ul"] = self.is_ul.data[idx]
+            if not self.has_ul:
+                table.remove_columns("is_ul")
```
adonath, line 374: I think this would be even simpler:

```suggestion
            if self.has_ul:
                table["is_ul"] = self.is_ul.data[idx]
```

Astro-Kirsty: This was also my initial idea, however, this section of code already creates the column `is_ul` and fills that column with data.

https://github.com/gammapy/gammapy/blob/0655a46e352c8d4ccf38168e89d3b7bb06b038c4/gammapy/estimators/points/core.py#L361-L364

Open to further suggestion

adonath: You are right, just leave the suggestion open....

gammapy/estimators/points/tests/test_core.py
```diff
@@ -336,3 +337,22 @@ def test_flux_points_plot_no_error_bar():
...
+        norm_n_values=3,
+        source="source",
+    )
+
+    FluxPoints = fpe.run(datasets=dataset_1)
+    table = FluxPoints.to_table()
+
+    if not FluxPoints.has_ul:
+        with pytest.raises(AttributeError):
+            table.is_ul
```
registerrier, line 358:
```suggestion
    with pytest.raises(ValueError):
        "is_ul" in table.colnames
```

gammapy/estimators/points/tests/test_core.py
```diff
@@ -336,3 +337,22 @@ def test_flux_points_plot_no_error_bar():
     flux_points = FluxPoints.from_table(table)
     with mpl_plot_check():
         _ = flux_points.plot(sed_type="dnde")
+
+
+def test_fp_no_is_ul():
+    dataset_1 = simulate_map_dataset(random_state=0, name="dataset_1")
```
registerrier, line 343: It might be better not to import and use here the estimator. I would rather suggest to import a `FluxPoints` from a table where you remove the UL column.
See e.g. `test_plot_fp_no_ul`

registerrier, line 343: It might be better not to import and use here the estimator. I would rather suggest to import a `FluxPoints` from a table where you remove the UL column.
See e.g. `test_plot_fp_no_ul`

registerrier (commented):
Thanks @Astro-Kirsty . See inline comments.

gammapy/estimators/points/tests/test_core.py
Astro-Kirsty: I utilised the `simulate_map_dataset`, so that the `FluxPoints` has the `has_ul` set to False.
By importing the flux points from `GAMMAPY_DATA` mean `has_ul` is set to True. 
Open to suggestion on if you still wish to import the `FluxPoints` from a table

registerrier: Even if you remove the `ul` column from the input table before creating the `FluxPoints` object?

Astro-Kirsty: Apologies, it works now. Thanks @registerrier

gammapy/estimators/points/tests/test_core.py
Astro-Kirsty: Correct me if I wrong but, `"is_ul" in table.colnames` should return False, therefore it will not be a ValueError

registerrier: Yes indeed. But `table.is_ul` will always raise an `AttributeError` since columns are not accessible as attributes on `Table`.
So the correct thing to do is:
```
assert "is_ul" in table.colnames.
```

gammapy/estimators/points/tests/test_core.py
```diff
@@ -336,3 +336,16 @@ def test_flux_points_plot_no_error_bar():
     flux_points = FluxPoints.from_table(table)
     with mpl_plot_check():
         _ = flux_points.plot(sed_type="dnde")
+
+
+def test_fp_no_is_ul():
```
registerrier, line 341:
```suggestion
@requires_data()
def test_fp_no_is_ul():
```

The decorator is required for CI to run.

gammapy/estimators/points/tests/test_core.py
```diff
@@ -336,3 +336,16 @@ def test_flux_points_plot_no_error_bar():
...
+    path = make_path("$GAMMAPY_DATA/tests/spectrum/flux_points/flux_points.fits")
+    table = Table.read(path)
+    table.remove_column("is_ul")
+    table.remove_column("flux_ul")
+
+    fp = FluxPoints.from_table(table)
+
+    if not fp.has_ul:
+        assert "is_ul" not in table.colnames
```
registerrier, line 351:
```suggestion
    fp = FluxPoints.from_table(table)
    fp_table = fp.to_table()

    assert "is_ul" not in fp_table.colnames
```

registerrier (dismissed):
Thanks @Astro-Kirsty . 

the test was not testing anything. I have left an inline suggestion to make it work: you have to create a new table from the `FluxPoints` object and check that `is_ul` is not one of its columns.

registerrier (approved):
Thanks @Astro-Kirsty . No further comment on my side.

PR #4548, opened 2023-06-01
Update the CITATION.cff file

registerrier:
**Description**

This pull request modifies the CITATION.cff for the v1.1 release. Following the requirements of PIG 24, it creates a new list of authors that will be increase with minor releases until the next LTS.

**Dear reviewer**

PR #4553, opened 2023-06-01
Authors script lts option

adonath:
**Description**

This pull request adds a `--since-last-lts` option and hard-codes the authors for the Gammapy CC and the ones that did not opt in for v1.0. This will allow to check #4548.


registerrier (commented):
Thanks @adonath . 
My main comment would be that we don't need to follow-up on the non-opt-in list for LTS. I would remove the list.

adonath:
Thanks @registerrier, I removed the list. I missed that with the `--since-last-lts` option it is actually not needed.

PR #4561, opened 2023-06-02
Export Datastore to Obscore table

paulakx:
**Description**

This pull request tries to implement obscore Table inside gammapy

It builds an integrated Obscore Table taking a given observation or a set of observations in four well defined tasks (one function for each)  inside the DataStore module. 
Only one function is accessible for the user to_obscore_table, the other three (_obscore_def, _obscore_row, _write_obscore_file) are hidden.

Dear reviewer, I still have some failures in the tests, it would be good to discuss with you to solve them. Thank you


adonath:
Thanks @paulakx, can you please provide a description of what this PR is supposed to implement?

registerrier (commented):
Thanks @paulakx . This is an interesting new functionality. I think we need to clarify a bit the use cases.

From discussions on Slack, the objective is to be able to build an Obscore table from an in-memory list of observations/datastore/obs_table. This obscore table will then be used to set up a VO service to query and access this collection. So in the long run there should also be the opposite situation where you want to build a `DataStore` from the VO service directly. Right?

For now, the main function is `DataStore.to_obscore_table`, which loops over the selected obs to build the Obscore table. This raises a few questions:
- there is nothing specific to the `DataStore` used here since you extract the observations one by one. You could equivalently have this function living directly on the `Observations` or the `ObservationTable`. I understand that there is some information that is more general such as the `access_url` or `publisher_id`, but those two are not present (so far) on the `DataStore` either. 
- Currently our datastore is tightly coupled with the GADF format (as is most of our data subpackage). This has to evolve to accommodate  different formats. Our global view so far is that we will have Reader/Writer classes or functions to convert from the specific format to the internal data model. This PR can then be a first element in a `VODataStore` converter.  

Those reasons suggest that placing this function on the `DataStore` is not the most practical choice. I would propose to make this a function in e.g. gammapy/data/io.py.

gammapy/data/data_store.py
```diff
@@ -455,6 +457,336 @@ def check(self, checks="all"):
...
+        if selected_obs is None:
+            obscore_tab = tab_default
+
+        else:
+            obscore_tab = Table()
+            for i in range(0, len(selected_obs)):
+                obscore_row = self._obscore_row(selected_obs[i], **kwargs)
+                obscore_tab = vstack([obscore_tab, obscore_row])
+
+        self._write_obscore_file(obscore_tab, format)
```
registerrier, line 763: Why write it on disk here?

gammapy/data/data_store.py
```diff
@@ -455,6 +457,336 @@ def check(self, checks="all"):
...
+
+    def _obscore_def(self):
+        """Generate the Obscore default table
+
+        In case the obscore standard changes, this function should be changed consistently
+
+
+        Returns
+        -------
+        Astropy Table length=0
```
QRemy, line 471: Why not using the VO table from astropy ? 
https://docs.astropy.org/en/stable/io/votable/index.html#building-a-new-table-from-scratch
It should better match what you need, for example missing values are supported
https://docs.astropy.org/en/stable/io/votable/index.html#missing-values

QRemy (commented):
Thanks @paulakx, for my comments are mostly questions :

gammapy/data/data_store.py
```diff
@@ -455,6 +457,336 @@ def check(self, checks="all"):
         checker = DataStoreChecker(self)
         return checker.run(checks=checks)
 
+    ##########################################################################################
+    # My contribution:
+
+    def _obscore_def(self):
```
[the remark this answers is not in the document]
paulakx: Regis, thank you for your comments.

I agree, I'm moving this to a separate set of functions in a script called ivoa.py in gammapy/data/ similar to what you suggested.

gammapy/data/data_store.py
```diff
@@ -455,6 +457,336 @@ def check(self, checks="all"):
...
+        tab["obs_publisher_did"] = obs_publisher_did
+        tab["access_url"] = access_url
+        tab["access_format"] = "application/fits"
+
+        tab["access_estsize"] = size
+        tab["s_ra"] = observation.get_pointing_icrs(observation.tmid).ra.value
+        tab["s_dec"] = observation.get_pointing_icrs(observation.tmid).dec.value
+
+        tab["s_fov"] = 10.0
+        #    tab["s_region"] = 'null'
```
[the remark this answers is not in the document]
paulakx: done

gammapy/data/data_store.py
paulakx: I removed for now the function to write the table, currently to_oscore_table returns an astropy table object.

gammapy/data/data_store.py
```diff
@@ -455,6 +457,336 @@ def check(self, checks="all"):
         checker = DataStoreChecker(self)
         return checker.run(checks=checks)
 
+    ##########################################################################################
+    # My contribution:
+
+    def _obscore_def(self):
+        """Generate the Obscore default table
+
+        In case the obscore standard changes, this function should be changed consistently
```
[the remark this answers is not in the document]
paulakx: @QRemy  Thank you for your comments.

In the next commit you will see many changes, most of them following Regis's suggestions.
Regarding this question, for now we must support one version, and maybe in the future this can evolve, anyway I need to define the bases somewhere and I decided to compact it in a "_obscore_def" function.

gammapy/data/data_store.py
paulakx: I was evaluating your suggestion, I understand that VOtable is more focused on the format of the tables and its connection with the xml format. Here I propose something more general, which is the creation of a table that can be written in different formats and that is why it seemed appropriate to use astropy table. In the next commit you will see that the new function only returns an astropy table object and then the user can write it in the format they need.
Regarding missing values, astropy table also has ways to treat them with masks. But currently we prioritize keeping the dtype of each obscore field (eg 0.0 for an empty double precision field according to obscore).

bkhelifi: Indeed, the astropy method is not adapted, proposing here only one serialization type. One should be more generic and be able to serialise this table in FITS, yaml, xml, ...

bkhelifi:
Have you ran the code style on the files? e.g. `tox -e codestyle'
It seems that they are some extra empty lines

paulakx:
>

paulakx:
> Have you ran the code style on the files? e.g. `tox -e codestyle' It seems that they are some extra empty lines

Yes, and I think is OK. Here the result of thi command:
**$ tox -e codestyle**
codestyle: commands[0] gammapy/data> flake8 gammapy --count --max-line-length=100
0
  codestyle: OK (2.76=setup[0.03]+cmd[2.72] seconds)
  congratulations :) (3.05 seconds)

bkhelifi (dismissed):
Thanks

cgalelli (dismissed):

gammapy/data/ivoa.py
```diff
@@ -0,0 +1,338 @@
...
+    tab_default = _obscore_def()
+
+    if selected_obs is None:
+        obscore_tab = tab_default
+
+    else:
+        obscore_tab = Table()
+        for i in range(0, len(selected_obs)):
+            obscore_row = _obscore_row(base_dir, selected_obs[i], **kwargs)
+            obscore_tab = vstack([obscore_tab, obscore_row])
```
registerrier, line 336: I find the logic a bit puzzling. Why do you create a table and fill it with one element at each step e.g. with `Table.add_row()`

gammapy/data/ivoa.py
```diff
@@ -0,0 +1,338 @@
...
+    tab["em_min"] = observation.events.energy.min().value
+    tab["em_max"] = observation.events.energy.max().value
+    tab["facility_name"] = observation.obs_info["TELESCOP"]
+    tab["instrument_name"] = observation.obs_info["TELLIST"]
+
+    return tab
+
+
+def to_obscore_table(base_dir, selected_obs=None, **kwargs):
+
```
registerrier, line 307: remove empty line between definition and docstring

registerrier (commented):
Thanks @paulakx ! Sorry for the late review. This is more clear this way. 

1. This still misses a few tests. Because the code will evolve in the future we really need to have unit tests to make sure we don't break the functionalities. 
2. I don't see the point of creating a table with 1 element before stacking them. You could keep the empty table creation but simply fill it with an `add_row`. Or equivalently build the list of rows and fill the table.

gammapy/data/ivoa.py
```diff
@@ -0,0 +1,338 @@
...
+    tab["t_exptime"] = observation.observation_live_time_duration.value
+    tab["em_min"] = observation.events.energy.min().value
+    tab["em_max"] = observation.events.energy.max().value
+    tab["facility_name"] = observation.obs_info["TELESCOP"]
+    tab["instrument_name"] = observation.obs_info["TELLIST"]
+
+    return tab
+
+
+def to_obscore_table(base_dir, selected_obs=None, **kwargs):
```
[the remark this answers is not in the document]
paulakx: ok, done.

gammapy/data/ivoa.py
```diff
@@ -0,0 +1,338 @@
...
+    tab["dataproduct_type"] = observation.obs_info["EXTNAME"]
+    tab["calib_level"] = 2  # Look into the data
+    tab["target_name"] = observation.obs_info["OBJECT"]
+    tab["obs_id"] = str(observation.obs_info["OBS_ID"])
+    tab["obs_collection"] = "DL3"
+    tab["obs_publisher_did"] = obs_publisher_did
+    tab["access_url"] = access_url
+    tab["access_format"] = "application/fits"
+    tab["access_estsize"] = size
+    tab["s_ra"] = observation.get_pointing_icrs(observation.tmid).ra.value
```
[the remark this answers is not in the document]
paulakx: Done

gammapy/data/ivoa.py
paulakx: Thanks for your reviews. In fact, using a table "vstack" was the only way I found in the beginning to be able to achieve the IVOA requests where they recommend keeping the empty "values" consistent with each column data type. Now, because of how the script evolved, it seems reasonable to me to do it using .add_row(). 

I pushed a new version of ivoa.py where now _obscore_row() does not generate a table of length “one” but directly adds a row to the already existing table for each ObsID.

gammapy/data/ivoa.py
```diff
@@ -0,0 +1,327 @@
...
+    tab = table
+    observation = data_store.obs(single_obsID)
+    obs_mask = data_store.obs_table["OBS_ID"] == observation.obs_id
+    obs_pos = data_store.obs_table[obs_mask]["EVENTS_FILENAME"]
+    path = make_path(os.environ["GAMMAPY_DATA"] + "/" + obs_pos[0])
+    size = int(os.path.getsize(path) / 1000.0)
+    tab.add_row(
+        {
+            "dataproduct_type": observation.obs_info["EXTNAME"],
+            "calib_level": 2,  # Look into the data
```
AtreyeeS, line 273: This would be DL3 data, no? Calib level 2

gammapy/data/ivoa.py
```diff
@@ -0,0 +1,327 @@
...
+            "obs_collection": "DL3",
+            "obs_publisher_did": str(obs_publisher_did),
+            "access_url": str(access_url),
+            "access_format": "application/fits",
+            "access_estsize": size,
+            "s_ra": observation.get_pointing_icrs(observation.tmid).ra.to_value("deg"),
+            "s_dec": observation.get_pointing_icrs(observation.tmid).dec.to_value(
+                "deg"
+            ),
+            "s_fov": 10.0,
```
AtreyeeS, line 285: The FoV is not fixed. It should be guessed from the IRFs?

gammapy/data/ivoa.py
```diff
@@ -0,0 +1,327 @@
...
+        }
+    )
+
+    return tab
+
+
+def to_obscore_table(
+    base_dir, selected_obs=None, obs_publisher_did=None, access_url=None
+):
+    """Generate the complete obscore Table by adding one row per observation using DataStore._obscore_row()
```
AtreyeeS, line 302: `_obscore_row` is not in DataStore

gammapy/data/ivoa.py
```diff
@@ -0,0 +1,327 @@
...
+
+    Parameters
+    ----------
+    selected_obs : list or array of Observation ID(int)
+    (default of ``None`` means ``no obaservation ``)
```
AtreyeeS, line 307:
```suggestion
    (default of ``None`` means ``no observation ``)
```

gammapy/data/ivoa.py
```diff
@@ -0,0 +1,327 @@
...
+    If not given, the obscore default table is returned.
```
AtreyeeS, line 308: So an empty Table is returned. Is that useful?

AtreyeeS (commented):
Thank you very much @paulakx 

I have a few questions in general.

1. How is this intended to be used? The Obscore table will be created once during the data release, and be used during all queries? Or will it be generated for the requested obs_ids during a query?
2. The user needs to know the relevant obs_ids before generating this Table?
3. Can you not use the `obs_table` directly?
4. Your `obscore_row` does not fill some columns, eg: o_ucd, em_res_power etc. The idea is to leave them blank?

gammapy/data/ivoa.py
bkhelifi: `calib_level` is a VO terminology that is not associated to our TeV one

gammapy/data/ivoa.py
```diff
@@ -0,0 +1,327 @@
...
+
+
+log = logging.getLogger(__name__)
+log.setLevel(logging.INFO)
+
+
+def _obscore_def():
+    """Generate the Obscore default table
+    In case the obscore standard changes, this function should be changed consistently
+
```
bkhelifi, line 19: Can you add a link towards the documentation of the obsscore format?

bkhelifi:
Hi Atreyee,
You was not involved in the discussion, so thank you for your questions.
1. The obscore table is a table that is used for VO TAP server. As you have guessed, it is our equivalent of our obs_table, but in a format precised by the IVOA. This table permits to the VO server to provide the discoverability of the data and their findability. When CTAO will prepare the TAP server for one of the DL3 releases, the CTAO staff will use this function to create the obsscore table... I think that for the SDC CTAO wish to make some tests by providing a TAP server.
2. By passing the obs_id, this function can be used to generate a table that can be concatenated to an existing one or to generate the table for a full production. This is then fully generic and allows many possibilities

gammapy/data/ivoa.py
paulakx: Hi @AtreyeeS, thank you very much for responding and sorry for my delay!  I see that @bkhelifi  has answered your first two questions well and I agree with his answers. Just a remark in general terms, yes, this tool has to be as general as possible for different types of users who need to generate both an obscore table for the entire data set as well as a subset of that data.

Regarding the other two questions:
3) We cannot use it because it is not a "standard".
4) Yes exactly, the idea is to leave them blank because they are mandatory columns in obscore, then we need to keep them anyway.

gammapy/data/ivoa.py
paulakx: Yes, I agree, but we have not yet defined how to do it, so for now it is fixed. If you agree we can add a comment on this line.

gammapy/data/ivoa.py
paulakx: Ok, this is left over from a previous version. I already changed the comment.

gammapy/data/ivoa.py
paulakx: sure

gammapy/data/ivoa.py
```diff
@@ -0,0 +1,327 @@
...
+    Parameters
+    ----------
+    selected_obs : list or array of Observation ID(int)
+    (default of ``None`` means ``no observation ``)
+    If not given, the obscore default table is returned.
+
+
+
+    **kwargs : `str` {obs_publisher_did, access_url}
```
AtreyeeS, line 312: Maybe mention what `obs_publisher_did` is supposed to be. it is not intuitively obvious. Same for access_url

AtreyeeS (commented):
Thanks for the explanations @paulakx and @bkhelifi ! I understand a lot better now.

1.  I understand that the format of our obs_table is different from the Obscore implemented here, but I was wondering if it's possible to use the values directly from the obs_table instead of accessing from the observation itself. The former is much faster.
 

2. I was wondering about the default behaviour of  `to_obscore_table`
To get an empty table, users can always call `obscore_def` (does it need to be hidden?).
If `selected_obs=None`, I would naively expect the full table. Or, have an explicit option (eg `selected_obs=all`) to get the same.

3. It seems the `obs_publisher_did` and `access_url` can differ for each observation. However, in the way it's implemented now, these values will be same for all rows for each creation of the table. Is that intended?

paulakx:
Hi @AtreyeeS , thanks for your comments!

1) Do you mean using directly _gammapy.data import ObservationTable_
and obs_table = ObservationTable.read(directory) ?
If it is the case, the problem is that there is missing information in obs_table and we need to read metadata that is outside as ["EXTNAME"] or ["OBJECT"], unless we can add this information to obs_table.
2) Yes you are right, it doesn't  need to be hidden, I will change that in order to give the empty table by obscore_def() and if  selected_obs=None it will give the full Table.
3) In fact they are the same for each observation because they belong to the same dataset provided by the same publisher. But the same dataset may however have more than one publisher if it is published in more than one location by a different publisher. I changed the warnings so that they only show once now and I added some links to the warnings in order to give some information about how these values should be built.

registerrier (commented):
Thanks @paulakx . The function still lacks unit tests. This is to ensure that future changes in the code will always maintain the expected behaviour.
So having a test that makes sure the main elements are present in the output table would be important here.

gammapy/data/ivoa.py
```diff
@@ -23,217 +31,224 @@ def obscore_def():
...
+    for var in obscore_table:
         tab_default.add_column(var)
     return tab_default
 
 
-def _obscore_row(base_dir, single_obsID, obs_publisher_did, access_url, table):
+def _obscore_row(
+    base_dir,
+    single_obsID,
+    obs_publisher_did,
```
bkhelifi, line 247: Only the parameter `single_obsID' is described in the docstring. Even if this function is not exposed, the description of all parameters and their default value is important for the other developers.

gammapy/data/ivoa.py
```diff
@@ -284,7 +314,11 @@ def _obscore_row(base_dir, single_obsID, obs_publisher_did, access_url, table):
 
 
 def to_obscore_table(
-    base_dir, selected_obs=None, obs_publisher_did=None, access_url=None
+    base_dir,
+    selected_obs=None,
```
bkhelifi, line 318: All parameters should be described in the docstring

bkhelifi (changes requested):
Thanks @paulakx . It looks almost ready for me. I have only a small request change on the doctrings.

gammapy/data/ivoa.py
paulakx: Done!

AtreyeeS (dismissed):
Thanks @paulakx ! No more comments from my side

registerrier (commented):
Thanks @paulakx . I have left small inline comments.

The handling of `DataStore` and `table` is a bit too complex. I might propose some changes directly to your branch PR if you agree.

gammapy/data/ivoa.py
```diff
@@ -0,0 +1,394 @@
...
+
+    Returns
+    -------
+    obscore_tab : ~astropy.table.Table
+        Obscore table with number of rows = len(selected_obs)
+    """
+
+    if obs_publisher_did is None:
+        log.warning(
+            "Insufficient publisher information: 'obs_publisher_did' obscore value will be empty. Giving this values is highly recommended. obs_publisher_did is the Dataset identifier given by the publisher. For more information about these arguments please check the IVOA recommendations: https://www.ivoa.net/documents/ObsCore and https://www.ivoa.net/documents/IVOAIdentifiers/20160523/REC-Identifiers-2.0.html"
```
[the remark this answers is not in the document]
paulakx: done !

gammapy/data/tests/test_ivoa.py
```diff
@@ -0,0 +1,43 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+
+from gammapy.data.ivoa import obscore_structure, to_obscore_table
+from gammapy.utils.testing import requires_data
+
+
+def test_obscore_structure():
+    obscore_default_tab = obscore_structure()
+    assert len(obscore_default_tab.columns) == 29
+    assert len(obscore_default_tab["dataproduct_type"]) == 0
```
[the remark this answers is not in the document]
paulakx: Done !  I added a test to check the name of the first and last column, the dtype and part of its metadata.

registerrier (approved):
Thanks @paulakx . I have cleaned-up and simplified the code. The general interface to `to_obscore_table` in unchanged. The _obscore_row is renamed and simplified. `obscore_structure` is renamed `empty_obscore_table`.

registerrier:

bkhelifi (approved):
Thanks @paulakx and @registerrier .... So sad that one can not add the CreatorMetaData information in such dictionary ;)

PR #4566, opened 2023-06-07
Sample time for phase template models

registerrier:
**Description**

This pull request introduces a `sample_time` method for `TemplatePhaseCurveTemporalModel`. This method determines
the best `t_delta` to use and then call the parent `sample_time` method. This has the advantage of providing correct phase 
distributions independently of the user input. 
This might be important when dealing with the `MapDatasetEventSampler` because you can provide a sample `t_delta` for all models. Computing down to the milliseconds for e.g. background models is useless. 

**Dear reviewer**

### DIFF
--- a/gammapy/modeling/models/temporal.py
+++ b/gammapy/modeling/models/temporal.py
@@ -1120,7 +1120,7 @@ class TemplatePhaseCurveTemporalModel(TemporalModel):
         """Sample arrival times of events.
 
         To fully cover the phase range, t_delta is the minimum between the input
-        and 5 percent of the period at 0.5*(t_min + t_max).
+        and product of the period at 0.5*(t_min + t_max) and the table bin size.
 
         Parameters
         ----------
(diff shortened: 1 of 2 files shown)


gammapy/modeling/models/temporal.py
```diff
@@ -1115,3 +1115,43 @@ def plot_phasogram(self, ax=None, n_points=100, **kwargs):
...
+        # Determine period at the mid time
+        t_mid = Time(t_min, scale=self.scale) + 0.5 * (t_max - t_min)
+        delta_t = (t_mid - self.reference_time).to(u.d)
+        frequency = self.f0.quantity + delta_t * (
+            self.f1.quantity + delta_t * self.f2.quantity / 2
+        )
+        period = 1 / frequency
+
+        # Take minimum time delta between user input and 5% of the period
+        t_delta = np.minimum(0.05 * period, t_delta)
```
fabiopintore, line 1155: Thanks @registerrier .
It seems Ok to me. I just wonder if there is a reason to have 5% and not 10% or 1%, for example...

Issue #2877, opened 2020-04-30
Comments and ideas on flux points estimation


QRemy:
The current implementation of the `FluxPointsEstimator` is well suited for 1D case or simple 3D models but I see several limitations : 
- works only if the source is defined as a `SkyModel`, does not for the `SkyDiffuseCube` and `BackgroundModel`
- does not work for linked sources as it does not account for the normalisation or amplitude parameters links. 
- the estimator has to be run for each source independently which is slow and can cause inconsistencies. For example if an excess is not well accounted by the baseline models multiple sources can be biased upward. If the re-optimize option is used then each source flux points are computed based on a different global model and the errorbars do not reflect the crosstalk between the components.
- slow and encounters convergence issues, especially when the re-optimize option is True.

I think several of these issues can be avoided by simply pre-computing the models and just re-optimizing a norm for a set of selected models. I prepared a notebook that show the behaviour that I would expect when dealing with multiple model components. I started to prototype a solution to give an option that keep the links, it works but it is not simple and clean. I also compare different ways to compute the dnde points: as done in the `FluxPointsEstimator`, or converting flux points with `to_sed_type`, or using gradient.

https://nbviewer.jupyter.org/github/QRemy/gammapy/blob/multi_flux_points/examples/CustomFluxbands_test.ipynb

Any comments/ideas ?
@registerrier, @adonath

registerrier:
Thanks @QRemy. I agree, there are numerous issues that we need to solve for the flux points. The following ones are likely connected:
- `FluxPointsEstimator` should be able to work with `SkyDiffuseCube` indeed.
- we need to make it faster.
At the moment, we define a `ScaleSpectralModel` on top of the `SpectralModel`. This ensures that things function smoothly between 1D and 3D. In practice, as you suggest in your example, we should have a `ScaleSkyModel` that precomputes all `npred` in each energy bin and adds a `norm` factor.

For the inclusion of the other parameters, I see the point. As a check, one might want to keep track of other normalization parameters. Yet, it might become very complex at some point. Probably correlation is important as well. Maybe one could store the result somewhere but not return it as a `FluxPoints` object.

adonath:
Just for the record, there was already some progress in this:
- The `FluxPointsEstimator` now takes a `Fit` instance on input, which allow users to deal with convergence issues if present.
- Now that the `SkyDIffuseCube` is gone, the flux poin estimation just works with the `TemplateSpatialModel` and a spectral model
[...]

Issue #4472, opened 2023-04-18
Investigate fail in FluxPointsEstimator


AtreyeeS:
**Gammapy version**
1.0, 1.0.1

While running `FluxPointsEstimator` with multiple models present, and `reoptimize=True`, a confusing error has been sometimes reported

```

File ~/Gammapy-dev/gammapy/gammapy/datasets/evaluator.py:301, in MapEvaluator._compute_flux_spatial(self)
    299     value = (values.quantity * weights).sum(axis=(1, 2), keepdims=True)
    300 else:
--> 301     value = self._compute_flux_spatial_geom(self.geom)
    303 return value

File ~/Gammapy-dev/gammapy/gammapy/datasets/evaluator.py:309, in MapEvaluator._compute_flux_spatial_geom(self, geom)
    307 if not self.model.spatial_model.is_energy_dependent:
    308     geom = geom.to_image()
--> 309 value = self.model.spatial_model.integrate_geom(geom)
    311 if self.psf and self.model.apply_irf["psf"]:
    312     value = self.apply_psf(value)

File ~/Gammapy-dev/gammapy/gammapy/modeling/models/spatial.py:245, in SpatialModel.integrate_geom(self, geom, oversampling_factor)
    243     # Finally stack result
    244     result._unit = integrated.unit
--> 245     result.stack(integrated)
    246 else:
    247     values = self.evaluate_geom(wcs_geom)

File ~/Gammapy-dev/gammapy/gammapy/maps/wcs/ndmap.py:878, in WcsNDMap.stack(self, other, weights, nan_to_num)
    876     cutout_slices = Ellipsis, slices[0], slices[1]
    877 else:
--> 878     raise ValueError(
    879         "Can only stack equivalent maps or cutout of the same map."
    880     )
    882 data = other.quantity[cutout_slices].to_value(self.unit)
    883 if nan_to_num:

ValueError: Can only stack equivalent maps or cutout of the same map.
```

I tried to produce a minimal notebook here where the error shows (on cell 24)
https://github.com/AtreyeeS/test_notebooks/blob/master/fpe_fail.ipynb

from what i understood, the wcs_geom  here (https://github.com/gammapy/gammapy/blob/18c2597f23816f465189f6dde4e6e75fd9e50fe6/gammapy/modeling/models/spatial.py#L227)  ends up with width      : 0.0 deg x 0.0 deg when the spatial model position is very close to the lat/lon allowed range, probably leading to a situation with very little overlap on geom

[...]

PR #4567, opened 2023-06-07
Freeze source parameters in FluxEstimator

registerrier:
**Description**

This pull request modifies the behaviour of `FluxEstimator` freezing all the source of interest parameters to only leave the `norm`  of the `ScaleSpectralModel` free.
This was discussed in #2877 but we never implemented it. It does not seem meaningful to leave the source spatial parameters free. Can we guarantee that it does not fit a different source?

Because the `Fit` and the `Estimator` classes are stateless, it does not seem possible to properly test that the source spatial parameters do not change during the various steps. Could there be a way to test this?

**Dear reviewer**
This changes the behavior rather strongly, but I can't really see a good reason to leave the non spectral parameters of the source of interest free. Opinions?


AtreyeeS:
This is the issue reported in #4472
Specially for upper limit computation, this becomes a big problem as the extension continuously keeps changing, so each upper limit corresponds to a different region. And, the fit sometimes goes ends on a different source.

@adonath suggested we leave it to the user to freeze all spatial parameters before running the estimator.
Then we should explain the behaviour and the caveats clearly.

bkhelifi:
The suggestion of @adonath is wise, indeed. However, are the non-expert aware of the issue? Is there properly documented?
The other possibility is to add a comment or a note in the doctring...

QRemy (dismissed):
I vote to merge this.

bkhelifi (changes requested):
For me, this behaviour seems very reasonable and logical. So, I do approuve this PR.

One should however precise this behaviour in the docstring (and maybe in global rst page), and precising also that a global fit is needed as prior

AtreyeeS (approved):
I see no good reason to not merge this, so +1 from my side as well. 
But this should be added in the estimator user guide, and the change in behaviour specifically mentioned in the release notes.

bkhelifi (approved):
Thanks a lot @QRemy

AtreyeeS (approved):
Thank you @QRemy ! Merging this

PR #4578, opened 2023-06-13
Fix ExcessMapEstimator to account for mask safe

QRemy:
Fix ExcessMapEstimator to account for mask safe

### DIFF
--- a/gammapy/estimators/map/tests/test_excess.py
+++ b/gammapy/estimators/map/tests/test_excess.py
@@ -36,6 +36,18 @@ def simple_dataset():
 
 
 @pytest.fixture
+def simple_dataset_mask_safe():
+    axis = MapAxis.from_energy_bounds(0.1, 10, 3, unit="TeV")
+    geom = WcsGeom.create(npix=20, binsz=0.02, axes=[axis])
+    dataset = MapDataset.create(geom)
+    dataset.mask_safe += np.ones(dataset.data_shape, dtype=bool)
+    dataset.mask_safe.data[0, :, :] = False
+    dataset.counts += 2
+    dataset.background += 1
+    return dataset
+
+
+@pytest.fixture
 def simple_dataset_on_off():
     axis = MapAxis.from_energy_bounds(0.1, 10, 2, unit="TeV")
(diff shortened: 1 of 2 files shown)


gammapy/estimators/map/tests/test_excess.py
```diff
@@ -142,6 +154,27 @@ def test_significance_map_estimator_map_dataset(simple_dataset):
     assert_allclose(result["npred_excess_ul"].data[0, 10, 10], 107.806275, atol=1e-3)
 
 
+def test_significance_map_estimator_map_dataset_mask_safe(simple_dataset_mask_safe):
+    simple_dataset_mask_safe.exposure = None
+    energy_edges = simple_dataset_mask_safe.counts.geom.axes[0].edges
+    energy_edges = [energy_edges[0], energy_edges[-1]]
+    estimator = ExcessMapEstimator(
+        0.1 * u.deg, selection_optional=["all"], energy_edges=energy_edges
+    )
```
registerrier, line 163:
```suggestion
    estimator = ExcessMapEstimator(0.1 * u.deg, selection_optional=["all"])
```
If you take min and max edges, you don't need to specify them.

registerrier (commented):
Thanks @QRemy . Looks good.

Should we use the `mask_safe` or the full `mask`?

QRemy:
> Should we use the `mask_safe` or the full `mask`?

It should be consistent with `MapDataset.resample_energy_axis` that apply only mask_safe.
If we change the behaviour of `MapDataset.resample_energy_axis` to include the mask_fit by default it should probably done in a separate PR. I don't have a strong opinion about that personally I just redefine the mask_safe when needed.

AtreyeeS (approved):
Thanks @QRemy ! 

I think in a future PR, it will be good to take the fit mask into account.
Usually, we discourage users from modifying the mask safe.

registerrier:
Indeed in practice, we usually need an exclusion mask to compute the excess maps. Using the `mask_fit` is likely the simpler approach.

registerrier (approved):
Thanks @QRemy . No further comment.

PR #4592, opened 2023-06-16
Improve the documentation of the reoptimise argument

bkhelifi:
**Description**

This simple PR describes a bit more the behaviour of this argument.

### DIFF
--- a/gammapy/estimators/points/lightcurve.py
+++ b/gammapy/estimators/points/lightcurve.py
@@ -89,6 +89,9 @@ class LightCurveEstimator(FluxPointsEstimator):
         self.time_intervals = time_intervals
         self.atol = u.Quantity(atol)
 
+        if "reoptimize" not in kwargs:
+            kwargs["reoptimize"] = False
+
         super().__init__(**kwargs)
 
     def run(self, datasets):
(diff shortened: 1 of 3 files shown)


MRegeard (approved):
Thanks @bkhelifi,
No comment on my side.

MRegeard:
> Thanks @bkhelifi ! This is indeed a common query.
> 
> I think the `FluxPointsEstimator` needs to be adapted, rather than the `FluxEstimator`. It is also useful to specify that the spectral parameters are always kept frozen.
> 
> Probably how the estimators work can go in the documentation, and the doc strings can refer to the said section

Hi @AtreyeeS,

Why not also do this on the `FluxEstimator` ? It the same thing for this object right ? Even though it's not public, for homogeneity we could repeat this everywhere ?

bkhelifi:
I have implemented the comments of @AtreyeeS.

gammapy/estimators/points/lightcurve.py
```diff
@@ -87,6 +89,9 @@ def __init__(self, time_intervals=None, atol="1e-6 s", **kwargs):
         self.time_intervals = time_intervals
         self.atol = u.Quantity(atol)
 
+        if "reoptimize" not in kwargs:
+            kwargs["reoptimize"] = False
+
```
QRemy, line 94:
```suggestion
```

the default is already False as it is inherited from the `FluxEstimator`

registerrier (approved):
Thanks @bkhelifi . No comments from my side.

Issue #3878, opened 2022-04-08
Safe mask functions should not rely on the projected IRFs


AtreyeeS:
Continuing from the discussions in #3873, maybe safe mask functions should not use the binned IRFs, but rather the DL3 irfs directly when possible. This is a reminder issue for v1.0

adonath:
@AtreyeeS This is a backwards incompatible change and it would be bad to introduce it between v1.0rc and v1.0, because it might change analysis results. I'll label the issue after v1.0.

PR #4599, opened 2023-06-20
Add a SafeMaskMaker at DL3 level

AtreyeeS:
This PR implements a SafeMaskMaker to run on DL3 IRFs as discussed in #3878 .
The cuts are applied directly on the event list.
Since HAWC and SWGO are directly using DL4 IRFs, I guess there is no need to adapt this PR for non pointing instruments.

In this case, the `SafeMakerDL3` must be run before the `MapDatasetMaker`, as opposed to the `SafeMaskMaker`

The energy dispersion is not applied as yet.

### DIFF
--- a/gammapy/makers/safe.py
+++ b/gammapy/makers/safe.py
@@ -153,6 +153,28 @@ class SafeMaskMaker(Maker):
 
         return dataset._geom.energy_mask(energy_min=energy_min, energy_max=energy_max)
 
+    def _get_offset(self, observation):
+        offset = self.fixed_offset
+        if offset is None:
+            if self.position:
+                offset = observation.get_pointing_icrs(observation.tmid).separation(
+                    self.position
+                )
+            else:
+                offset = 0.0 * u.deg
+        return offset
+
+    def _get_position(self, observation, geom):
+        if self.fixed_offset is not None and observation is not None:
+            pointing = observation.get_pointing_icrs(observation.tmid)
+            return pointing.directional_offset_by(
+                position_angle=0 * u.deg, separation=self.fixed_offset
+            )
+        elif self.position is not None:
+            return self.position
+        else:
+            return geom.center_skydir
+
     def make_mask_energy_aeff_max(self, dataset, observation=None):
         """Make safe energy mask from effective area maximum value.
 
@@ -168,17 +190,16 @@ class SafeMaskMaker(Maker):
         mask_safe : `~numpy.ndarray`
             Safe data range mask.
         """
+
+        if self.fixed_offset is not None and observation is None:
+            raise ValueError(
+                f"{observation} argument is mandatory with {self.fixed_offset}"
+            )
+
         geom, exposure = dataset._geom, dataset.exposure
 
         if self.irfs == "DL3":
-            offset = self.fixed_offset
-            if offset is None:
-                if self.position:
-                    offset = observation.get_pointing_icrs(observation.tmid).separation(
-                        self.position
-                    )
-                else:
-                    offset = 0.0 * u.deg
+            offset = self._get_offset(observation)
 
             values = observation.aeff.evaluate(
                 offset=offset, energy_true=observation.aeff.axes["energy_true"].edges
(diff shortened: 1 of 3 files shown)


registerrier (commented):
Thanks @AtreyeeS . This is interesting, but I think we have to understand how this fits in the workflow.

A few general comments/questions:
- technically this is more an event filtering scheme than a mask creation algorithm.  One could use the existing `ObservationFilter`.
- one can probably split event filtering from mask creation, but the latter is still necessary for further treatment and analysis.
- the advantage of creating the masks on the input geometry and apply them to filter events is that you don't have to worry about fractional exposures in the bins. If not, `mask_safe` has to contain weights between 0 and 1 rather than booleans.  This might be useful in a number of situation but requires heavy modifications in `stacking` and other functionalities. Currently `edisp` is the place where we do this (and once stacked, the `mask_safe` is partly redundant with the `EdispKernelMap`).
- The idea of having event filtering as part of the list of `Makers` is interesting because it could allow reading observations from the `DataStore` within the loop over observations (or in that case obs_id). 

One option could be first to adapt the `SafeMaskMaker` to use the observation `IRF` rather than the `Dataset` IRF to build the mask. This would alleviate some of the difficulties we have currently.

Opinions @adonath @QRemy  ?

AtreyeeS:
This should be ready for review now

gammapy/makers/safe.py
```diff
@@ -162,45 +170,64 @@ def make_mask_energy_aeff_max(self, dataset, observation=None):
...
-                )
-            else:
-                raise ValueError(
-                    f"observation argument is mandatory with {self.fixed_offset}"
-                )
+        if self.irfs == "DL3":
+            offset = self.fixed_offset
+            if offset is None:
+                if self.position:
+                    offset = observation.get_pointing_icrs(observation.tmid).separation(
```
QRemy, line 177: test if observation is None ? as for the DL4 case

QRemy (commented):
Thanks @AtreyeeS, I left few minor comments. It may be further simplified to have less nested if tests, but if it is not possible it may be nicer to move the DL3/DL4 in separated functions. I will have a closer look into that.

gammapy/makers/safe.py
```diff
@@ -220,38 +247,54 @@ def make_mask_energy_edisp_bias(self, dataset, observation=None):
...
+                        self.position
+                    )
+                else:
+                    offset = 0.0 * u.deg
+
+            edisp = observation.edisp.to_edisp_kernel(offset=offset)
+            energy_min = edisp.get_bias_energy(self.bias_percent / 100)[-1]
```
QRemy, line 262:
```suggestion
            energy_min = edisp.get_bias_energy(self.bias_percent / 100)[0]
```
as for the DL4 case ?

gammapy/makers/safe.py
```diff
@@ -264,17 +307,29 @@ def make_mask_energy_bkg_peak(dataset):
...
-        background_spectrum = dataset.npred_background().get_spectrum()
-        idx = np.argmax(background_spectrum.data, axis=0)
-        energy_axis = geom.axes["energy"]
-        energy_min = energy_axis.edges[idx]
+        if self.irfs == "DL3":
+            bkg = observation.bkg.to_2d()
+            background_spectrum = np.ravel(
+                bkg.integral(axis_name="offset", offset=bkg.axes["offset"].bounds[1])
+            )
+            energy_min = bkg.axes["energy"].center[np.argmax(background_spectrum)]
```
QRemy, line 325: center and not edges like DL4 case ?

registerrier (approved):
Thanks @AtreyeeS . This looks good. I have only minor comments.

gammapy/makers/safe.py
```diff
@@ -83,6 +87,10 @@ def __init__(
                 "`position` and `fixed_offset` attributes are mutually exclusive"
             )
 
+        if irfs not in ["DL3", "DL4"]:
+            raise IOError("Invalid option for irfs. Choose 'DL3' or 'DL4'.")
```
AtreyeeS, line 91:
```suggestion
            ValueError("Invalid option for irfs: expected 'DL3' or 'DL4', got {irfs} instead.")
```

AtreyeeS:
Thanks @registerrier ! Commited your suggestions. Should be good to merge now

registerrier (approved):
Thanks @AtreyeeS . No further comment.

registerrier:

PR #4608, opened 2023-06-23
Add a slice_by_coord function for FluxMaps

cgalelli:
This pull request adds a slice_by_coord function for FluxMaps, alongside (and exploiting) the slice_by_idx function. 
A use case example is slicing a lightcurve by using directly the time start and end of the desired portion instead of passing through axis indices.


MRegeard:
Thanks @cgalelli, 
I was wondering wether there are use cases where we would like to `slice_by_coord` over multiple axes. 
Since `slice_by_idx` can handle several axes, it could be supported, but it would drastically change this function, and probably making it not very user friendly. 
So my initial guess would be no, but I'm still asking !

cgalelli:
> I was wondering wether there are use cases where we would like to slice_by_coord over multiple axes.
Since slice_by_idx can handle several axes, it could be supported, but it would drastically change this function, and probably making it not very user friendly.
So my initial guess would be no, but I'm still asking !

Hi @MRegeard,
for sure it would be possible to slice by coordinates over multiple axes, in the same way as it is done by indexes. It is also true that the situation would become a bit more confusing for the user, but it would probably still be manageable as long as the different axes are easily distinguishable.

QRemy:
> Thanks @cgalelli, I was wondering wether there are use cases where we would like to `slice_by_coord` over multiple axes. Since `slice_by_idx` can handle several axes, it could be supported, but it would drastically change this function, and probably making it not very user friendly. So my initial guess would be no, but I'm still asking !

It would be nice to have the same behaviour than `slice_by_idx` for `slice_by_coord` just passing a dict with axes names as keys and tuple of quantities or times as values. The `slice_by_coord` should also be added to all the classes where `slice_by_idx` is defined, but this can be done in another PR.

In this PR you could also add for convenience two methods `slice_by_time(time_min=None, time_max=None)` and `slice_by_energy(energy_min=None, energy_min=None))` (like we have for the datasets).

(Tests are missing).

gammapy/estimators/map/core.py
```diff
@@ -1084,6 +1085,70 @@ def slice_by_idx(self, slices):
             gti=self.gti,
         )
 
+    def slice_by_coord(self, slices):
+        """Slice flux maps by coordinate values
+
+        Parameters
+        ----------
+        slices : dict
+            Dict of axes names and `qstropy.Quantity` or `qstropy.Time` or `slice` object pairs.
```
QRemy, line 1094:
```suggestion
            Dict of axes names and `astropy.Quantity` or `astropy.Time` or `slice` object pairs.
```

bkhelifi (approved):
Thanks a lot, @cgalelli

MRegeard (approved):

MRegeard:
Thanks @cgalelli, seems good to me !

gammapy/estimators/map/core.py
```diff
@@ -1084,6 +1085,70 @@ def slice_by_idx(self, slices):
...
+
+        for key, interval in zip(slices.keys(), slices.values()):
+            imin = self.geom.axes[key].coord_to_idx(interval.start)
+            imax = self.geom.axes[key].coord_to_idx(interval.stop)
+
+            idx_intervals.append(slice(imin, imax))
+
+        return self.slice_by_idx(dict(zip(slices.keys(), idx_intervals)))
+
+    def slice_by_time(self, time_min, time_max):
```
registerrier, line 1116: The two methods `slice_by_time` and `slice_by_energy` are each only one line of code. They are not tested also.

I think it is better to have them as example usage in the docstring of `slice_by_coord`.

Issue #4511, opened 2023-05-17
Introducing logics in ObservationFilter


MRegeard:
In order to use the `ObservationFilter` for pulsar analysis it would be good to have logics supported by it. 

The obvious thing I am thinking about is the logical `OR`. For instance if one wants to take two non-continuous phase range, let's say [0.4, 0.5] and [0.7, 0.8]. 
Currently I don't think there is any way to do that with ObservationFilter` . 

Probably the logical `AND` could also be useful. Let's say I want to filter through all the events that have energy > 1 TeV  and offset < 1 deg.

PR #4616, opened 2023-06-27
introducing union support in ObservationFilter

MRegeard:
This pull request introduce "union" filtering in `ObservationFilter`, as mentioned in #4511. 

First of all, I introduce a new input parameter to the class, `event_filter_method`, which can take either the value `'intersect'`or `'union'`. By default it is set to `'intersect'` to stick to the former behaviour. 

Then I modified the `filter_events` method to be able to do "union filtering". For that, I take the event filtered by each filter one by one, `vstack`their table, sort the table by `'TIME'`, apply `unique` to table and then put this in a new `EventList` which is retuned.

I also modified the `_check_filter_phase` (that check for "phase filter" and apply the livetime correction if needed) so that it can support multiple filtering. 

Finally, I also added a staticmethod `_check_overlap_phase`, which check if there are overlap in the define `band`of "phase filter". This is mandatory to avoid having livetime that are not corresponding to the really fraction of livetime that is kept.

### DIFF
--- a/gammapy/data/filters.py
+++ b/gammapy/data/filters.py
@@ -1,9 +1,9 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 import copy
 import logging
+from itertools import groupby
 import numpy as np
 from astropy.table import unique, vstack
-import pandas as pd
 
 __all__ = ["ObservationFilter"]
 
@@ -53,8 +53,8 @@ class ObservationFilter:
         self, time_filter=None, event_filters=None, event_filter_method="intersect"
     ):
         self.time_filter = time_filter
-        self.event_filters = event_filters or []
-        self._check_overlap_phase(self.event_filters)
+        event_filters = event_filters or []
+        self.event_filters = self._merge_overlapping(event_filters)
         self.event_filter_method = event_filter_method
 
     @property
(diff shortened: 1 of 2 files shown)


gammapy/data/filters.py
```diff
@@ -23,8 +26,10 @@ class ObservationFilter:
           class that corresponds to the filter type
           (see `~gammapy.data.ObservationFilter.EVENT_FILTER_TYPES`)
 
-        The filtered event list will be an intersection of all filters. A union
-        of filters is not supported yet.
+    event_filter_method: str
```
bkhelifi, line 29: Should we add an example with this argument in the docstring exemple?

gammapy/data/tests/test_filters.py
```diff
@@ -102,6 +102,13 @@ def test_filter_gti(observation):
             "p_in": [],
             "p_out": 1,
         },
+        {
+            "p_in": [
```
bkhelifi, line 106: Should we add other test with an union and an intersection?

registerrier:
What is the status here @MRegeard ? 
Is the event filtering working with GTIs that have holes?

MRegeard:
@registerrier, I think that the logic that I introduce here is not satisfactory. That's why didn't went further. 
I need to think about it again. 
For the GTI, I'm not sure.

QRemy:
What's the status here ? 
Should we close this until there is a better solution ?
@MRegeard

PR #4620, opened 2023-06-28
Adding prior class

katrinstreil:
**Description**

This pull request adds the prior base class as well as some exemplary subclasses. For details see the PR for the corresponding PIG (https://github.com/gammapy/gammapy/pull/4381). 

**Dear reviewer**


QRemy (commented):
Thanks @katrinstreil, this looks promising !
I wonder if it would be interesting to re-use the Model/Parameter API, I left few comments in that direction.
That may be useful if we consider the prior parameters as hyper-parameters that could be tuned.

gammapy/modeling/prior.py
```diff
@@ -0,0 +1,111 @@
...
+    def __init__(self, mu=0, sigma=1):
+        self.mu = mu
+        self.sigma = sigma
+
+    def evaluate(self, value):
+        return ((value - self.mu) / self.sigma) ** 2
+
+    @property
+    def prior_parameters(self):
+        return [("mu", self.mu), ("simga", self.sigma)]
```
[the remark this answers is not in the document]
bkhelifi: I am wondering whether one should not add a "StepPrior", ie 0 for X<Xmin and X>Xmax (Xmin and Xmax could be infinite) and 1 else where...
This can be nice to e.g. force a flux to be positive.

gammapy/modeling/prior.py
```diff
@@ -0,0 +1,111 @@
+from . import Parameter
+
+
+class Prior:
```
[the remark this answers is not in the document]
katrinstreil: Thanks @QRemy! This was a very helpful comment! I changed the PR to have a ``PriorParameter`` inheriting from the ``Parameter`` class. I think having attributes like ``frozen`` (and the later used ``.prior``) does not make sense for the ``PriorParameters`` so I changed that. The ``PriorModel`` is now a subclass of ``ModelBase`` and the models do get evaluated as a static method similar to the model API.

gammapy/modeling/prior.py
```diff
@@ -0,0 +1,188 @@
...
+
+    def __repr__(self):
+        return self.__class__.__name__
+
+
+class GaussianPrior(PriorModel):
+    """Gaussian Prior with mu and sigma."""
+
+    tag = ["GaussianPrior"]
+    _type = "prior"
```
QRemy, line 146:
```suggestion
```

gammapy/modeling/prior.py
```diff
@@ -0,0 +1,188 @@
...
+        return table
+
+
+class PriorModel(ModelBase):
+    """
+    Base prior class.
+    """
+
+    _weight = 1
+
```
QRemy, line 84:
```suggestion
_type = "prior"

```

gammapy/modeling/prior.py
```diff
@@ -0,0 +1,188 @@
...
+        return table
+
+
+class PriorModel(ModelBase):
+    """
+    Base prior class.
+    """
+
+    _weight = 1
+_type = "prior"
```
QRemy, line 84:
```suggestion
    _type = "prior"
```

gammapy/modeling/prior.py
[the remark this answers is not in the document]
QRemy: Thanks the changes look good, could you add some tests to check the init, evaluation, and serialisation  ?

gammapy/modeling/parameter.py
```diff
@@ -113,6 +113,7 @@ def __init__(
         scale_method="scale10",
         interp="lin",
         is_norm=False,
+        prio=None,
```
bkhelifi: Is it not `prior` here?

gammapy/modeling/parameter.py
```diff
@@ -144,6 +145,7 @@ def __init__(
         self.scan_n_sigma = scan_n_sigma
         self.interp = interp
         self.scale_method = scale_method
+        self.prior = None
```
bkhelifi:
```suggestion
        self.prior = prior
```

bkhelifi:
@katrinstreil, should you update your branch from the main such that the tests pass? I do not know if it is the right explanation...

katrinstreil:
Thanks @QRemy and @bkhelifi for the review and the comments! I added some tests, maybe you could have a look! 
And btw, updating my branch did unfortunately not help with the unsuccessful checks, idk

bkhelifi:
Hello @katrinstreil , are you sure that your branch is uptodate regarding the main? The CI failure seems related to a problem that has been recently fixed. @registerrier , do you confirm?

gammapy/modeling/models/prior.py
```diff
@@ -0,0 +1,149 @@
...
+import numpy as np
+import astropy.units as u
+from gammapy.modeling import PriorParameter, PriorParameters
+from .core import ModelBase
+
+log = logging.getLogger(__name__)
+
+
+def _build_priorparameters_from_dict(data, default_parameters):
+    """Build Parameters object from input dict and default  parameter values."""
```
bkhelifi, line 11:
```suggestion
    """Build Parameters object from input dict and default parameter values."""
```

bkhelifi, line 11: Maybe one can describe in the docstring the parameters, even for the developers...

gammapy/modeling/models/prior.py
```diff
@@ -0,0 +1,149 @@
...
+                    # "scale_method",
+                ]:
+                    default = init[item]
+
+                    if par[item] == default or (
+                        np.isnan(par[item]) and np.isnan(default)
+                    ):
+                        del par[item]
+
+                # if par["frozen"] == init["frozen"]:
```
bkhelifi, line 96: Should these lines be removed completely?

gammapy/modeling/models/prior.py
```diff
@@ -0,0 +1,149 @@
...
+        # assuming the same unit as the PriorParamter here
+        kwargs = {par.name: par.value for par in self.parameters}
+        return self.weight * self.evaluate(value.value, **kwargs)
+
+    def to_dict(self, full_output=False):
+        """Create dict for YAML serialisation"""
+        tag = self.tag[0] if isinstance(self.tag, list) else self.tag
+        params = self.parameters.to_dict()
+
+        if not full_output:
```
bkhelifi, line 80: What happens when the full_output is asked? 
If it is not yet implemented, one should raise a nonImplemented exception

gammapy/modeling/parameter.py
```diff
@@ -113,6 +113,7 @@ def __init__(
         scale_method="scale10",
         interp="lin",
         is_norm=False,
+        prior=None,
```
bkhelifi, line 116: Add the description of this new parameter in the docstring

gammapy/modeling/parameter.py
```diff
@@ -396,7 +415,7 @@ def __repr__(self):
         return (
             f"{self.__class__.__name__}(name={self.name!r}, value={self.value!r}, "
             f"factor={self.factor!r}, scale={self.scale!r}, unit={self.unit!r}, "
-            f"min={self.min!r}, max={self.max!r}, frozen={self.frozen!r}, id={hex(id(self))})"
+            f"min={self.min!r}, max={self.max!r}, frozen={self.frozen!r}, prior={self.prior!r}, id={hex(id(self))})"
         )
```
bkhelifi, line 420: @registerrier , in this PR should we implement here what is under review in the PR https://github.com/gammapy/gammapy/pull/3947?

gammapy/modeling/parameter.py
```diff
@@ -755,3 +791,85 @@ def __exit__(self, type, value, traceback):
...
+        }
+        return output
+
+    def __repr__(self):
+        return (
+            f"{self.__class__.__name__}(name={self.name!r}, value={self.value!r}, "
+            f"factor={self.factor!r}, scale={self.scale!r}, unit={self.unit!r}, "
+            f"min={self.min!r}, max={self.max!r})"
+        )
+
```
bkhelifi, line 842: Cf my comment about https://github.com/gammapy/gammapy/pull/3947

gammapy/modeling/prior.py.save
bkhelifi, line 1: Why is this file within the PR?

bkhelifi:
Could add the description of all parameters and their default values in the docstring of the new classes?

gammapy/modeling/models/prior.py
```diff
@@ -0,0 +1,149 @@
...
+                f"PriorParameter '{par_dict['name']}' not defined in YAML file."
+                f" Using default value: {par_dict['value']} {par_dict['unit']}"
+            )
+        par_data.append(par_dict)
+
+    return PriorParameters.from_dict(par_data)
+
+
+class PriorModel(ModelBase):
+    _weight = 1
```
QRemy, line 32: should this be a parameter ? at least it should be serialised

QRemy (commented):
Thanks @katrinstreil, it looks nice I just have a question about the usage of the weight :

gammapy/modeling/models/prior.py
katrinstreil: Yes, you are right! The weight is now a parameter which can be set with or after the initialisation and it gets serialised.

gammapy/modeling/models/prior.py
katrinstreil: Both `full_output = True` and `full_output = False` are implemented and work. If the full output is asked the "min",                     "max" and "error" are set in the dict.

gammapy/modeling/prior.py.save
katrinstreil: I don't know what happened here, I deleted the file

gammapy/modeling/parameter.py
```diff
@@ -427,7 +455,10 @@ def to_dict(self):
 
         if self._link_label_io is not None:
             output["link"] = self._link_label_io
-
+        if self.prior is not None:
+            output["prior"] = self.prior.to_dict()["prior"]
+        else:
+            output["prior"] = None
```
registerrier, line 461: Maybe you don't need those two lines. If no prior is defined it does not need to be explicit.

gammapy/modeling/parameter.py
```diff
@@ -483,6 +514,17 @@ def check_limits(self):
...
 
+    @property
+    def prior(self):
+        return [par.prior for par in self]
+
+    def prior_stat_sum(self):
+        parameters_stat_sum = 0
+        for par in self:
+            if par.prior is not None:
+                parameters_stat_sum += par.prior(par)
```
registerrier, line 525: This looks a bit weird to write `par.prior(par)`. The prior set on a parameter should always evaluate on the parameter value, no?

gammapy/modeling/parameter.py
```diff
@@ -755,3 +799,85 @@ def __exit__(self, type, value, traceback):
...
+
+    def __repr__(self):
+        return (
+            f"{self.__class__.__name__}(name={self.name!r}, value={self.value!r}, "
+            f"factor={self.factor!r}, scale={self.scale!r}, unit={self.unit!r}, "
+            f"min={self.min!r}, max={self.max!r})"
+        )
+
+
+class PriorParameters(Parameters):
```
registerrier, line 852: It is not totally clear to me whether this class is really needed, as one does not need to interact with the list of all prior parameters set. Yet I suppose the `from_dict` method to build the `Prior` justifies it.

gammapy/modeling/models/prior.py
```diff
@@ -0,0 +1,168 @@
...
+            log.warning(
+                f"PriorParameter '{par_dict['name']}' not defined in YAML file."
+                f" Using default value: {par_dict['value']} {par_dict['unit']}"
+            )
+        par_data.append(par_dict)
+
+    return PriorParameters.from_dict(par_data)
+
+
+class PriorModel(ModelBase):
```
registerrier, line 31: While this inherits from `ModelBase`, this is not really a model. This would be closer to a statistic object since it is a log-likelihood that it computes. Why not calling it simply `Prior`?

It would be less ambiguous for a user.

registerrier (commented):
Thanks @katrinstreil . This looks good!

I have left some comments and small suggestions. My main question is whether this should be called `PriorModel` or `Prior`. I think the latter is more explicit. The former makes sense only because of the inheritance. But a prior is not a model, it is a log-probability.

gammapy/modeling/parameter.py
registerrier: Yes, we should implement this. Maybe in a follow-up PR?

gammapy/modeling/parameter.py
registerrier: same here

registerrier (approved):
Thanks @katrinstreil ! This looks good to me!

registerrier:
Merging now.

PR #4621, opened 2023-06-28
Added the PriorFitStatistic class

katrinstreil:
**Description**

This pull request adds the PriorFitStatistic class within stats. It is used to evaluate the priors which are set on models or parameters. See the corresponding PIG here: https://github.com/gammapy/gammapy/pull/4381

### DIFF
--- a/gammapy/stats/prior_statistic.py
+++ b/gammapy/stats/prior_statistic.py
@@ -1,0 +1,21 @@
+class PriorFitStatistic:
+    """Class to compute statistics due to the priors set on models and their parameters.
+
+    Parameters
+    ----------
+    weight : float
+        Weight of the prior
+    """
+
+    def __init__(self, weight=1):
+        self.weight = weight
+
+    def stat_sum(self, models):
+        stat = 0
+        for m in models:
+            if m.prior is not None:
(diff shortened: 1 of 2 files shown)


gammapy/stats/prior_statistic.py
```diff
@@ -0,0 +1,21 @@
...
+        self.weight = weight
+
+    def stat_sum(self, models):
+        stat = 0
+        for m in models:
+            if m.prior is not None:
+                stat += m.prior_stat_sum()
+            for par in m.parameters:
+                if par.prior is not None:
+                    stat += par.prior_stat_sum()
```
QRemy, line 20: given the changes in your other PR it could be something like :
if isinstance(m, `PriorModel`):
    stat += m.evaluate()
for par in m.parameters:
                if par.prior is not None:
                    stat += par.prior.evaluate()

katrinstreil: Yes, it basically is the same. With the new way of handling the ``prior_stat_sum`` this ``PriorFitStatistic`` class is no longer needed. I will close the PR. Thanks @QRemy for reviewing it!

PR #4628, opened 2023-06-30
Match energy binning per decade to pyirf's

JBernete:
**Description**

This pull request has the purpose of matching what pyirf does when making an energy binning per decade, i.e. to make sure that when using the same energy limits and number of bins per decade, the resulting bins are exactly the same in both pyirf and gammapy.
Before these changes, the value of energy_max was always included as the last edge, even when it didn't match the bins-per-decade requirement, so the edges were shifted and the actual number of bins per decade was a fraction higher. Now, energy_max is not forced to be an edge and the last edge is recalculated.

**Dear reviewer**

This PR is ready for review.

### DIFF
--- a/gammapy/maps/axes.py
+++ b/gammapy/maps/axes.py
@@ -583,7 +583,7 @@ class MapAxis:
             )
 
         if per_decade:
-            nbin = np.ceil(np.log10(energy_max / energy_min).value * nbin)
+            nbin = np.floor(np.log10(energy_max / energy_min).value * nbin)
 
         if name is None:
             name = "energy"
--- a/gammapy/maps/axes.py
+++ b/gammapy/maps/axes.py
@@ -583,7 +583,10 @@ class MapAxis:
             )
 
         if per_decade:
-            nbin = np.floor(np.log10(energy_max / energy_min).value * nbin)
+            bin_per_decade = nbin
+            nbin = np.floor(np.log10(energy_max / energy_min).value * bin_per_decade)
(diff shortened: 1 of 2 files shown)


QRemy:
> Maybe another factory method with this specific behavior can be introduced if this is necessary.

or another option `strict_bounds`=True that produce the current behaviour by default  so it is backward compatible.

JBernete:
> > Maybe another factory method with this specific behavior can be introduced if this is necessary.
> 
> or another option `strict_bounds=True` that produce the current behaviour by default so it is backward compatible.

I like this idea. I added the option. Please review it again.

gammapy/maps/axes.py
```diff
@@ -564,6 +565,10 @@ def from_energy_bounds(
             Whether `nbin` is given per decade.
         name : str
             Name of the energy axis, either 'energy' or 'energy_true'
+        strict_bounds : bool
+            Whether to strictly end the binning at the max energy when
+            `per_decade=True`. If True, the number of bins per decade
+            might be slightly adjusted.
```
AtreyeeS, line 571: Maybe document explicitly the `False` option as well

registerrier (commented):
Thanks @JBernete . This is nice.

I think you need to run black on your commit. You could install the pre-commit hooks following: 
https://docs.gammapy.org/1.1/development/intro.html#get-set-up

Could you just add a simple test to make sure the option is correctly taken into account?

registerrier (approved):
Thanks @JBernete . This looks good. I have no further comments.

Issue #4016, opened 2022-06-30
Need for general decorrelation energy for any Spectral Model


chaimain:
**Is your feature request related to a problem? Please describe.**
Decorrelation energy is only calculated for the Power Law Spectral Model and not for any other Spectral Model.

**Describe the solution you'd like**
Probably a numerical estimation of the correlation energy of the flux amplitude for any given Spectral Model could work. But, maybe there is a smarter way to evaluate such correlation energy.

**Additional context**
We use the decorrelation energy of a given model as the reference energy of a spectral model, assuming the correlation of the flux amplitude with the other free parameters is minimum at that energy. For Spectral Models distinctively different from the Power Law, using the same decorrelation energy as that for Power Law will be a wrong assumption.

PR #4635, opened 2023-07-03
Add function to determine pivot energy for any spectral model

Astro-Kirsty:
**Description**
This PR is addressing the feature request from issue [#4016.](https://github.com/gammapy/gammapy/issues/4016) 
I have added a `pivot_energy` function to the `SpectralModel` class that finds the energy at which the error butterfly is minimum. This function is then 'overridden' in the `PowerLawSpectralModel` where there is an analytical equation available.

Please note: this must be performed in log space (as I understand) because the butterfly minimum occurs in the log space plot. 

**Dear reviewer**
I am certainly open to suggestions/feedback on this issue.

### DIFF
--- a/gammapy/modeling/models/spectral.py
+++ b/gammapy/modeling/models/spectral.py
@@ -213,25 +213,24 @@ class SpectralModel(ModelBase):
         """
         return self._propagate_error(epsilon=epsilon, fct=self, energy=energy)
 
-    def pivot_energy(self, energy_bounds, nbins=1000):
+    @property
+    def pivot_energy(self):
         """The decorrelation energy for a given spectral model calculated numerically.
-
-        Parameters
-        ----------
-        energy_bounds : `~astropy.units.Quantity`
-            Energy bounds to evaluate the spectral model and error.
 
         Returns
         -------
         pivot energy : `~astropy.units.Quantity`
             The energy at which the statistical error is smallest.
         """
-        energy_min, energy_max = energy_bounds
-        energy = np.geomspace(energy_min, energy_max, nbins)
 
-        dnde, dnde_error = self.evaluate_error(energy)
+        def min_energy(x):
+            x = np.exp(x)
+            x = u.Quantity(x, "GeV")
+            dnde, dnde_error = self.evaluate_error(x)
+            energy_minimum = dnde_error / dnde
+            return energy_minimum
 
-        return energy[np.argmin(dnde_error / dnde)]
+        return u.Quantity(np.exp(scipy.optimize.fsolve(min_energy, x0=1e-4)), "GeV")[0]
 
     def integral(self, energy_min, energy_max, **kwargs):
         r"""Integrate spectral model numerically if no analytical solution defined.
--- a/gammapy/modeling/models/spectral.py
+++ b/gammapy/modeling/models/spectral.py
@@ -213,6 +213,7 @@ class SpectralModel(ModelBase):
         """
         return self._propagate_error(epsilon=epsilon, fct=self, energy=energy)
 
+    @property
     def pivot_energy(self):
         """The decorrelation energy for a given spectral model calculated numerically.
 
@@ -220,21 +221,27 @@ class SpectralModel(ModelBase):
         -------
         pivot energy : `~astropy.units.Quantity`
             The energy at which the statistical error is smallest.
+            If no minimum is found, NaN will be returned.
         """
 
-        x_unit = u.GeV
+        x_unit = self.reference.unit
 
-        def min_energy(x):
+        def min_func(x):
+            """Function to minimize"""
             x = np.exp(x)
             dnde, dnde_error = self.evaluate_error(x * x_unit)
-            energy_minimum = dnde_error / dnde
-            return energy_minimum
+            return dnde_error / dnde
 
-        bounds = np.log([self.reference.value - 5, self.reference.value + 5])
-        minimizer = scipy.optimize.minimize_scalar(min_energy, bounds=bounds)
+        bounds = [np.log(self.reference.value) - 3, np.log(self.reference.value) + 3]
+
+        std = np.std(min_func(x=np.linspace(bounds[0], bounds[1], 100)))
+        if std < 1e-5:
+            return np.nan * x_unit
+
+        minimizer = scipy.optimize.minimize_scalar(min_func, bounds=bounds)
 
         if not minimizer.success:
-            return print("Minimizer terminated unsuccessfully.")
+            return np.nan * x_unit
         else:
             return np.exp(minimizer.x) * x_unit
 
@@ -838,6 +845,7 @@ class PowerLawSpectralModel(SpectralModel):
         base = value / self.amplitude.quantity
         return self.reference.quantity * np.power(base, -1.0 / self.index.value)
 
+    @property
     def pivot_energy(self):
         r"""The decorrelation energy is defined as:
 
@@ -846,6 +854,11 @@ class PowerLawSpectralModel(SpectralModel):
             E_D = E_0 * \exp{cov(\phi_0, \Gamma) / (\phi_0 \Delta \Gamma^2)}
 
         Formula (1) in https://arxiv.org/pdf/0910.4881.pdf
+
+        Returns
+        -------
+        pivot energy : `~astropy.units.Quantity`
+            If no minimum is found, NaN will be returned.
         """
         index_err = self.index.error
         reference = self.reference.quantity
@@ -930,6 +943,7 @@ class PowerLawNormSpectralModel(SpectralModel):
         base = value / self.norm.quantity
         return self.reference.quantity * np.power(base, -1.0 / self.tilt.value)
 
+    @property
     def pivot_energy(self):
         r"""The decorrelation energy is defined as:
 
@@ -938,6 +952,11 @@ class PowerLawNormSpectralModel(SpectralModel):
             E_D = E_0 * \exp{cov(\phi_0, \Gamma) / (\phi_0 \Delta \Gamma^2)}
 
         Formula (1) in https://arxiv.org/pdf/0910.4881.pdf
+
+        Returns
+        -------
+        pivot energy : `~astropy.units.Quantity`
+            If no minimum is found, NaN will be returned.
         """
         tilt_err = self.tilt.error
         reference = self.reference.quantity
--- a/gammapy/modeling/models/spectral.py
+++ b/gammapy/modeling/models/spectral.py
@@ -215,7 +215,8 @@ class SpectralModel(ModelBase):
 
     @property
     def pivot_energy(self):
-        """The pivot or decorrelation energy for a given spectral model calculated numerically.
+        """The pivot or decorrelation energy, for a given spectral model calculated numerically.
+        It is defined as the energy at which the correlation between the spectral parameters is minimized.
 
         Returns
         -------
@@ -236,13 +237,17 @@ class SpectralModel(ModelBase):
 
         std = np.std(min_func(x=np.linspace(bounds[0], bounds[1], 100)))
         if std < 1e-5:
-            log.warning("No minimum was found.")
+            log.warning(
+                "The relative error on the flux does not depend on energy. No pivot energy found."
+            )
             return np.nan * x_unit
 
         minimizer = scipy.optimize.minimize_scalar(min_func, bounds=bounds)
 
         if not minimizer.success:
-            log.warning("The minimizer was not successful.")
+            log.warning(
+                "No minima found in the relative error on the flux. Pivot energy computation failed."
+            )
             return np.nan * x_unit
         else:
             return np.exp(minimizer.x) * x_unit
--- a/gammapy/modeling/models/spectral.py
+++ b/gammapy/modeling/models/spectral.py
@@ -215,12 +215,12 @@ class SpectralModel(ModelBase):
 
     @property
     def pivot_energy(self):
-        """The decorrelation energy for a given spectral model calculated numerically.
+        """The pivot or decorrelation energy for a given spectral model calculated numerically.
 
         Returns
         -------
         pivot energy : `~astropy.units.Quantity`
-            The energy at which the statistical error is smallest.
+            The energy at which the statistical error in the computed flux is smallest.
             If no minimum is found, NaN will be returned.
         """
 
@@ -849,7 +849,7 @@ class PowerLawSpectralModel(SpectralModel):
 
     @property
     def pivot_energy(self):
-        r"""The decorrelation energy is defined as:
+        r"""The pivot or decorrelation energy is defined as:
 
         .. math::
 
@@ -947,7 +947,7 @@ class PowerLawNormSpectralModel(SpectralModel):
 
     @property
     def pivot_energy(self):
-        r"""The decorrelation energy is defined as:
+        r"""The pivot or decorrelation energy is defined as:
 
         .. math::
 
--- a/gammapy/modeling/models/spectral.py
+++ b/gammapy/modeling/models/spectral.py
@@ -228,15 +228,7 @@ class SpectralModel(ModelBase):
         """
         dnde, dnde_error = self.evaluate_error(energy)
 
-        y_lo = dnde - dnde_error
-        y_hi = dnde + dnde_error
-
-        log_y_lo = np.log10(y_lo[y_lo > 0].value)
-        log_y_hi = np.log10(y_hi[y_lo > 0].value)
-
-        energy = energy[y_lo > 0]
-
-        return energy[np.argmin(log_y_hi - log_y_lo)]
+        return energy[np.argmin(dnde_error / dnde)]
 
     def integral(self, energy_min, energy_max, **kwargs):
         r"""Integrate spectral model numerically if no analytical solution defined.
--- a/gammapy/modeling/models/spectral.py
+++ b/gammapy/modeling/models/spectral.py
@@ -214,21 +214,23 @@ class SpectralModel(ModelBase):
         return self._propagate_error(epsilon=epsilon, fct=self, energy=energy)
 
     def pivot_energy(self, energy):
-        """Return the pivot energy for a given spectral model.
-        The pivot energy is the narrowest point of the butterfly.
+        """The decorrelation energy for a given spectral model calculated numerically.
 
         Parameters
         ----------
         energy : `~astropy.units.Quantity`
-            The energy to evaluate the spectral model and error.
+            Array of energies to evaluate the spectral model and error.
 
+        Returns
+        -------
+        pivot energy : `~astropy.units.Quantity`
+            The energy at which the statistical error is smallest.
         """
         dnde, dnde_error = self.evaluate_error(energy)
 
         y_lo = dnde - dnde_error
         y_hi = dnde + dnde_error
 
-        # Add here to get rid of any negative fluxes
         log_y_lo = np.log10(y_lo[y_lo > 0].value)
         log_y_hi = np.log10(y_hi[y_lo > 0].value)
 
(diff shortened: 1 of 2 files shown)


gammapy/modeling/models/spectral.py
```diff
@@ -213,6 +213,29 @@ def evaluate_error(self, energy, epsilon=1e-4):
...
 
+    def pivot_energy(self, energy):
+        """Return the pivot energy for a given spectral model.
+        The pivot energy is the narrowest point of the butterfly.
+
+        Parameters
+        ----------
+        energy : `~astropy.units.Quantity`
+            The energy to evaluate the spectral model and error.
+
```
MRegeard, line 224: Could you add the "returns docstring" here ?

MRegeard (changes requested):
Thanks @Astro-Kirsty,
Really helpful feature ! See inline comments.

gammapy/modeling/models/spectral.py
```diff
@@ -213,6 +213,29 @@ def evaluate_error(self, energy, epsilon=1e-4):
...
+        """
+        dnde, dnde_error = self.evaluate_error(energy)
+
+        y_lo = dnde - dnde_error
+        y_hi = dnde + dnde_error
+
+        # Add here to get rid of any negative fluxes
```
bkhelifi, line 231: Is there a way to test that in order to be sure that it works well?

gammapy/modeling/models/spectral.py
```diff
@@ -213,6 +213,29 @@ def evaluate_error(self, energy, epsilon=1e-4):
         """
         return self._propagate_error(epsilon=epsilon, fct=self, energy=energy)
 
+    def pivot_energy(self, energy):
+        """Return the pivot energy for a given spectral model.
+        The pivot energy is the narrowest point of the butterfly.
```
bkhelifi, line 218: The `butterfly' might not always have a shape of a butterfly.... I am wondering whether this is not too much into our jargon. Maybe one can just say energy with the smallest statistical error..
In the same manner, is `pivot_energy' a standard statistical word?

bkhelifi:
Thanks a lot @Astro-Kirsty . With this PR, I think that it would be great to have some tutorials with this new feature.... e.g. the spectral analysis or 3D detailed analysis

gammapy/modeling/models/spectral.py
Astro-Kirsty: I compared values from the analytical equation in the `PowerLawSpectralModel` to this new method, and was achieving the same pivot energy

gammapy/modeling/models/spectral.py
Astro-Kirsty: Okay, I agree with the use of the word butterfly. 
As for pivot energy, this is what the already defined function in `PowerLawSpectralModel` was called. So I kept it this way, open to suggestions. It has also been called the 'decorrelation energy'

gammapy/modeling/models/spectral.py
```diff
@@ -213,6 +213,31 @@ def evaluate_error(self, energy, epsilon=1e-4):
         """
         return self._propagate_error(epsilon=epsilon, fct=self, energy=energy)
 
+    def pivot_energy(self, energy):
+        """The decorrelation energy for a given spectral model calculated numerically.
+
+        Parameters
+        ----------
+        energy : `~astropy.units.Quantity`
```
AtreyeeS, line 221: should there be a default choice of energy?

gammapy/modeling/models/spectral.py
```diff
@@ -213,6 +213,31 @@ def evaluate_error(self, energy, epsilon=1e-4):
...
+            Array of energies to evaluate the spectral model and error.
+
+        Returns
+        -------
+        pivot energy : `~astropy.units.Quantity`
+            The energy at which the statistical error is smallest.
+        """
+        dnde, dnde_error = self.evaluate_error(energy)
```
AtreyeeS, line 229: why not the min of `dnde_error` directly ?

AtreyeeS (commented):
Thanks @Astro-Kirsty ! 
Is is sufficient to see the thinnest band among the range of energies supplied by the user? The decorrelation energy may not even lie in the given range.

If the energy array is too coarsely binned, the estimate will be incorrect. Finding the minima of the error might be better, no?

gammapy/modeling/models/spectral.py
Astro-Kirsty: A default energy array could be a good idea. 
Or the other option is for the user to define the `energy_bounds` instead, then there will be no issue with bins.

I believe there was some discussion recently about the validity of certain energy ranges for different functions. In this PR, I had to ensure that we did not have any negative fluxes `y_lo[y_lo > 0]`. Negative values for y_lo are possible if the energy array chosen for this function is not valid for the specific dataset used. For example, using a 'nonsensical' energy range for the Fermi data.

gammapy/modeling/models/spectral.py
Astro-Kirsty: This will not work, taking for example the `PowerLawSpectralModel`, the argmin  of dnde_error, would simply be the last energy bin in the plot (as we have a curve which is constantly decreasing.)

AtreyeeS: I meant that the function  can be replaced by 

```
dnde, dnde_error = self.evaluate_error(energy)
energy[np.argmin(dnde_error/dnde)]
```
This will automatically include the condition `y_lo > 0`

Astro-Kirsty: Apologies, I see what you mean. 

Good pick up, happy to implement it in this way! Thanks @AtreyeeS

AtreyeeS:
Thanks @Astro-Kirsty ! This looks better, but it is still an issue that this does not guarantee you are returning the decorrelation energy. This is just the minimum error point in the given range (without letting users know if the actual pivot lies outside). I tried with `fsolve`, and the results look stable. Maybe this is not the best optimiser, but something along the lines of

```
lp = LogParabolaSpectralModel(
        amplitude="5.82442e-11 cm-2 s-1 GeV-1",
        reference="17.337 GeV",
        alpha="1.9134",
        beta="0.2217",
    )
lp.amplitude.error = "2.8804e-12 cm-2 s-1 GeV-1"
lp.alpha.error = "0.1126"
lp.beta.error = "0.0670"

def f1(x, unit=u.TeV):
    x=np.exp(x)
    dnde, dnde_error = lp.evaluate_error(x*unit)
    ene = dnde_error / dnde
    return ene

print("pivot = ", np.exp(fsolve(f1, x0=0)))
```

maybe an option? What do you think?

Astro-Kirsty:
I agree with that @AtreyeeS. Using `fsolve` would be a nice way so that the use does not have to define any energy and therefore always gives the decorrelation energy for that SpectralModel

Astro-Kirsty:
Implemented the code using `fsolve`, I tried the function on multiple spectra to find a reasonable initial guess.

gammapy/modeling/models/spectral.py
```diff
@@ -213,6 +213,25 @@ def evaluate_error(self, energy, epsilon=1e-4):
...
+        """
+
+        def min_energy(x):
+            x = np.exp(x)
+            x = u.Quantity(x, "GeV")
+            dnde, dnde_error = self.evaluate_error(x)
+            energy_minimum = dnde_error / dnde
+            return energy_minimum
+
+        return u.Quantity(np.exp(scipy.optimize.fsolve(min_energy, x0=1e-4)), "GeV")[0]
```
registerrier, line 233: what will happen if `fsolve` fails to find a minimum?

registerrier (commented):
Thanks @Astro-Kirsty .

`fsolve` is a root finder, not a minimizer. Is it really adapted here? Shouldn't one use something like `minimize_scalar` instead?

Those methods require an interval to search for the minimum. We have to see if it is possible to keep `pivot_energy` a simple property in this context.

One should probably check that a proper minimum is found and yield an error if not.

Astro-Kirsty:
An option would be to implement:
`minimizer = scipy.optimize.minimize_scalar(min_energy, bounds=bounds)`

where the bounds are predefined by default as the `reference` value from the `SpectralModel` plus or minus some small value. But making this an optional argument, so that the user can specific their own bounds?

And as you mention, to add that if the minimizer is not successful, then it should alert the user

gammapy/modeling/models/spectral.py
```diff
@@ -213,6 +213,31 @@ def evaluate_error(self, energy, epsilon=1e-4):
...
+    def pivot_energy(self):
+        """The decorrelation energy for a given spectral model calculated numerically.
+
+        Returns
+        -------
+        pivot energy : `~astropy.units.Quantity`
+            The energy at which the statistical error is smallest.
+        """
+
+        x_unit = u.GeV
```
bkhelifi, line 225: Why Gev and not TeV? ;)

gammapy/modeling/models/spectral.py
```diff
@@ -213,6 +213,31 @@ def evaluate_error(self, energy, epsilon=1e-4):
...
+            x = np.exp(x)
+            dnde, dnde_error = self.evaluate_error(x * x_unit)
+            energy_minimum = dnde_error / dnde
+            return energy_minimum
+
+        bounds = np.log([self.reference.value - 5, self.reference.value + 5])
+        minimizer = scipy.optimize.minimize_scalar(min_energy, bounds=bounds)
+
+        if not minimizer.success:
+            return print("Minimizer terminated unsuccessfully.")
```
bkhelifi, line 237: In one case, the function return a string and in the other a Quantity. 
I think that it is best that in this case:
- there is a log.warning message (that will precise from which function the message is coming from)
- and return a nan * x_unit

gammapy/modeling/models/spectral.py
```diff
@@ -213,6 +213,40 @@ def evaluate_error(self, energy, epsilon=1e-4):
         """
         return self._propagate_error(epsilon=epsilon, fct=self, energy=energy)
 
+    @property
+    def pivot_energy(self):
+        """The decorrelation energy for a given spectral model calculated numerically.
```
MRegeard, line 218: It is a little weird to me that the function is named `pivot_energy` and that the description is `The decorrelation energy`. I know that the name `pivot_energy` was already in the code before. But maybe for clarity it is worse to rename everything coherently ?

MRegeard (approved):
Thanks @Astro-Kirsty. 
Apart for my comments on the name, I'm ok with this.

gammapy/modeling/models/spectral.py
```diff
@@ -213,6 +213,40 @@ def evaluate_error(self, energy, epsilon=1e-4):
...
+
+        Returns
+        -------
+        pivot energy : `~astropy.units.Quantity`
+            The energy at which the statistical error is smallest.
```
AtreyeeS, line 223:
```suggestion
            The energy at which the statistical error in the computed flux is smallest.
```

gammapy/modeling/models/spectral.py
```diff
@@ -213,6 +213,40 @@ def evaluate_error(self, energy, epsilon=1e-4):
...
+            x = np.exp(x)
+            dnde, dnde_error = self.evaluate_error(x * x_unit)
+            return dnde_error / dnde
+
+        bounds = [np.log(self.reference.value) - 3, np.log(self.reference.value) + 3]
+
+        std = np.std(min_func(x=np.linspace(bounds[0], bounds[1], 100)))
+        if std < 1e-5:
+            log.warning("No minimum was found.")
+            return np.nan * x_unit
```
AtreyeeS, line 240: do nans need units?

AtreyeeS: What is this warning case?

gammapy/modeling/models/spectral.py
```diff
@@ -822,6 +856,11 @@ def pivot_energy(self):
             E_D = E_0 * \exp{cov(\phi_0, \Gamma) / (\phi_0 \Delta \Gamma^2)}
 
         Formula (1) in https://arxiv.org/pdf/0910.4881.pdf
+
+        Returns
+        -------
+        pivot energy : `~astropy.units.Quantity`
+            If no minimum is found, NaN will be returned.
```
AtreyeeS, line 863: Does the nan case arise here?

gammapy/modeling/models/spectral.py
```diff
@@ -915,6 +954,11 @@ def pivot_energy(self):
...
```
AtreyeeS, line 961: same as before

gammapy/modeling/models/tests/test_spectral.py
```diff
@@ -614,14 +614,29 @@ def test_ecpl_integrate():
...
     pwl.covariance = [
         [0.0318377**2, 6.56889442e-14, 0],
         [6.56889442e-14, 0, 0],
         [0, 0, 0],
     ]
+    assert_quantity_allclose(pwl.pivot_energy, 3.3540034 * u.TeV)
 
-    assert_quantity_allclose(pwl.pivot_energy, 3.3540034240210987 * u.TeV)
+
+def test_num_pivot_energy():
```
AtreyeeS, line 627: maybe test a failing case as well

AtreyeeS (commented):
Thanks @Astro-Kirsty ! Left some small inline questions

gammapy/modeling/models/spectral.py
bkhelifi: This is super important! minimize_scalar fails when the function is constant...

gammapy/modeling/models/spectral.py
bkhelifi: yes

gammapy/modeling/models/spectral.py
bkhelifi: yes

AtreyeeS: For the simple power law, this is an analytical result, no?

gammapy/modeling/models/spectral.py
AtreyeeS: ok

gammapy/modeling/models/spectral.py
bkhelifi: If there are no error, there is no minimum. So a NaN should be returned

AtreyeeS: Yes indeed! that's the behaviour in the present version. 
Thanks Bruno

gammapy/modeling/models/tests/test_spectral.py
Astro-Kirsty: Do you mean for when the minimizer is not successful?

gammapy/modeling/models/spectral.py
Astro-Kirsty: > do nans need units?

Keeping units to be consistent with the analytical output for the `PowerLawSpectralModel`

gammapy/modeling/models/tests/test_spectral.py
```diff
@@ -624,6 +624,20 @@ def test_pwl_pivot_energy():
...
 
 
+def test_num_pivot_energy():
+    lp = LogParabolaSpectralModel(
+        amplitude="5.82442e-11 cm-2 s-1 GeV-1",
+        reference="17.337 GeV",
+        alpha="1.9134",
+        beta="0.2217",
+    )
+    lp.amplitude.error = "2.8804e-12 cm-2 s-1 GeV-1"
```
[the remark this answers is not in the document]
Astro-Kirsty: Making a parameter frozen will not effect anything here, it is about the errors utilised. 
So, I am not sure of the reason behind doing this?

MRegeard (approved):
Thanks @Astro-Kirsty !
Seems good to me.

AtreyeeS (approved):
Thanks @Astro-Kirsty ! I tend to agree with @registerrier suggestion during the dev call of improving the warning message to make it more explicit

Astro-Kirsty:
Could you please remind me of which warning needs to be more specific @AtreyeeS ?

gammapy/modeling/models/spectral.py
```diff
@@ -213,6 +213,40 @@ def evaluate_error(self, energy, epsilon=1e-4):
         """
         return self._propagate_error(epsilon=epsilon, fct=self, energy=energy)
 
+    @property
+    def pivot_energy(self):
+        """The pivot or decorrelation energy for a given spectral model calculated numerically.
```
AtreyeeS, line 218:
```suggestion
        """The pivot or decorrelation energy, for a given spectral model calculated numerically.
         It is defined as the energy at which the correlation between the spectral parameters is minimized. 
```

gammapy/modeling/models/spectral.py
```diff
@@ -213,6 +213,40 @@ def evaluate_error(self, energy, epsilon=1e-4):
...
+            """Function to minimize"""
+            x = np.exp(x)
+            dnde, dnde_error = self.evaluate_error(x * x_unit)
+            return dnde_error / dnde
+
+        bounds = [np.log(self.reference.value) - 3, np.log(self.reference.value) + 3]
+
+        std = np.std(min_func(x=np.linspace(bounds[0], bounds[1], 100)))
+        if std < 1e-5:
+            log.warning("No minimum was found.")
```
AtreyeeS, line 239:
```suggestion
            log.warning("The relative error on the flux does not depend on energy. No pivot energy found.")
```

gammapy/modeling/models/spectral.py
```diff
@@ -213,6 +213,40 @@ def evaluate_error(self, energy, epsilon=1e-4):
...
+            return np.nan * x_unit
+
+        minimizer = scipy.optimize.minimize_scalar(min_func, bounds=bounds)
+
+        if not minimizer.success:
+            log.warning("The minimizer was not successful.")
```
AtreyeeS, line 245:
```suggestion
            log.warning("No minima found in the relative error on the flux. Pivot energy computation failed.")
```

AtreyeeS (approved):

registerrier:
Thanks @Astro-Kirsty . There is a test failing, could you increase the tolerance? I suppose the method won't be precise below the percent.

Astro-Kirsty:
Thanks @registerrier. It was not failing on my laptop which is odd. I have increased the tolerance

registerrier (commented):

The only missing thing to me is a test with some correlation. e.g. use an exp-cutoff with the same parameters and correlation matrix as the PWL test and check you get the same result.

gammapy/modeling/models/spectral.py
```diff
@@ -213,6 +213,45 @@ def evaluate_error(self, energy, epsilon=1e-4):
...
+
+        def min_func(x):
+            """Function to minimize"""
+            x = np.exp(x)
+            dnde, dnde_error = self.evaluate_error(x * x_unit)
+            return dnde_error / dnde
+
+        bounds = [np.log(self.reference.value) - 3, np.log(self.reference.value) + 3]
+
+        std = np.std(min_func(x=np.linspace(bounds[0], bounds[1], 100)))
```
QRemy, line 238: does the choice of the number of steps in linespace can affect the precision ?
If yes it should be configurable.

QRemy (commented):
Thanks @Astro-Kirsty this is a nice feature, I have just one question :

gammapy/modeling/models/spectral.py
Astro-Kirsty: It does not

AtreyeeS (approved):
Thanks! no more comments from me

bkhelifi (approved):
Thanks @Astro-Kirsty, all good for me.
I am still wondering if one could not have an optional argument about the energy range of computation. But this can be added later if needed.

MRegeard (approved):
Thanks @Astro-Kirsty, 
Seems good to me !

Issue #4637, opened 2023-07-04
Adding a `concatenate` method to `Observations` 


MRegeard:
I think it would be interesting  to have a proper `concatenate` `classmethod` on `Observations`.  
Sometimes, `DataStore` are split by "epoque".  Therefore, if you want to do an analysis over multiple "epoque" you end up having multiple `Observations` while you want to do a classic analysis with all the `Observation` combined (not a joint-analysis).

registerrier:
This is solved by #4638 . The only remaining thing is to document this somewhere.

Astro-Kirsty:
@MRegeard I think we can close this now? 
I think we do not need further document as it will be present in the API reference documentation?

PR #4638, opened 2023-07-04
Add a `from_stack` method on `Observations`

MRegeard:
This PR implement #4637.

It allows to create an `Observations` from a list of `Observations`. It act like the `concatenate` method of `numpy`.

### DIFF
--- a/gammapy/data/observations.py
+++ b/gammapy/data/observations.py
@@ -715,6 +715,22 @@ class Observations(collections.abc.MutableSequence):
             obs_groups[f"group_{label}"] = observations
         return obs_groups
 
+    @classmethod
+    def concatenate(cls, observations_list):
+        """Create a new `Observations` instance by concatenating a list of `Observations` objects.
+
+        Parameters
+        ----------
+        observations_list : list of `~gammapy.data.Observations`
+            The list of `Observations` to concatenate.
+
+        Returns
+        -------
+        observations : `~gammapy.data.Observations`
+            The `Observations` object resulting from the concatenation of all the `Observations` in `observation_list`.
+        """
+        return cls(list(np.concatenate(observations_list)))
+
 
 class ObservationChecker(Checker):
     """Check an observation.
--- a/gammapy/data/observations.py
+++ b/gammapy/data/observations.py
@@ -2,6 +2,7 @@
 import collections.abc
 import copy
 import inspect
+import itertools
 import logging
 import warnings
 from itertools import zip_longest
@@ -716,7 +717,7 @@ class Observations(collections.abc.MutableSequence):
         return obs_groups
 
     @classmethod
-    def concatenate(cls, observations_list):
+    def from_stack(cls, observations_list):
         """Create a new `Observations` instance by concatenating a list of `Observations` objects.
 
         Parameters
@@ -729,7 +730,8 @@ class Observations(collections.abc.MutableSequence):
         observations : `~gammapy.data.Observations`
             The `Observations` object resulting from the concatenation of all the `Observations` in `observation_list`.
         """
-        return cls(list(np.concatenate(observations_list)))
+        obs = itertools.chain(*observations_list)
+        return cls(list(obs))
 
 
 class ObservationChecker(Checker):
--- a/gammapy/data/observations.py
+++ b/gammapy/data/observations.py
@@ -718,20 +718,46 @@ class Observations(collections.abc.MutableSequence):
 
     @classmethod
     def from_stack(cls, observations_list):
+        # TODO : Do more check when stacking observations when we have metadata.
         """Create a new `Observations` instance by concatenating a list of `Observations` objects.
 
         Parameters
         ----------
         observations_list : list of `~gammapy.data.Observations`
-            The list of `Observations` to concatenate.
+            The list of `Observations` to stack.
 
         Returns
         -------
         observations : `~gammapy.data.Observations`
-            The `Observations` object resulting from the concatenation of all the `Observations` in `observation_list`.
+            The `Observations` object resulting from the stacking of all the `Observations` in `observation_list`.
         """
         obs = itertools.chain(*observations_list)
-        return cls(list(obs))
+        return cls.check_uniqueness(cls(list(obs)))
+
+    @staticmethod
+    def check_uniqueness(obs):
+        # TODO : Change this once we support event type.
+        """
+        Check that there are no duplicate of `Observation` in the `Observations` object. If there are duplicates,
+        only raise a warning to signal the user.
+
+        Parameters
+        ----------
+            obs : `~gammapy.data.Observations`
+            The `Observations` to check for uniqueness.
+
+        Returns
+        -------
+            obs : `~gammapy.data.Observations`
+            The `Observations` that has been checked, left unchanged.
+        """
+        if len(set(obs.ids)) == len(obs.ids):
+            return obs
+        else:
+            log.warning(
+                "Found observation with the same `obs_id`. Check that all observations are different."
+            )
+            return obs
 
 
 class ObservationChecker(Checker):
--- a/gammapy/data/observations.py
+++ b/gammapy/data/observations.py
@@ -622,7 +622,17 @@ class Observations(collections.abc.MutableSequence):
     """
 
     def __init__(self, observations=None):
-        self._observations = observations or []
+        if all(isinstance(obs, Observation) for obs in observations or []):
+            self._observations = observations or []
+            if len(set(observations)) != len(observations):
+                log.warning(
+                    "Found duplicates of the same observation in observations. "
+                    "Check that all observations are different."
+                )
+        else:
+            raise TypeError(
+                "Invalid type: at least one element in the input list is not of type 'Observation'."
+            )
 
     def __getitem__(self, key):
         return self._observations[self.index(key)]
@@ -638,6 +648,8 @@ class Observations(collections.abc.MutableSequence):
 
     def insert(self, idx, obs):
         if isinstance(obs, Observation):
+            if obs in self:
+                log.warning(f"{obs.obs_id} already belongs to Observations.")
             self._observations.insert(idx, obs)
         else:
             raise TypeError(f"Invalid type: {type(obs)!r}")
@@ -732,32 +744,7 @@ class Observations(collections.abc.MutableSequence):
             The `Observations` object resulting from the stacking of all the `Observations` in `observation_list`.
         """
         obs = itertools.chain(*observations_list)
-        return cls.check_uniqueness(cls(list(obs)))
-
-    @staticmethod
-    def check_uniqueness(obs):
-        # TODO : Change this once we support event type.
-        """
-        Check that there are no duplicate of `Observation` in the `Observations` object. If there are duplicates,
-        only raise a warning to signal the user.
-
-        Parameters
-        ----------
-            obs : `~gammapy.data.Observations`
-            The `Observations` to check for uniqueness.
-
-        Returns
-        -------
-            obs : `~gammapy.data.Observations`
-            The `Observations` that has been checked, left unchanged.
-        """
-        if len(set(obs.ids)) == len(obs.ids):
-            return obs
-        else:
-            log.warning(
-                "Found observation with the same `obs_id`. Check that all observations are different."
-            )
-            return obs
+        return cls(list(obs))
 
 
 class ObservationChecker(Checker):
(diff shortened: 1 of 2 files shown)


gammapy/data/observations.py
```diff
@@ -715,6 +715,22 @@ def group_by_label(self, labels):
...
+        ----------
+        observations_list : list of `~gammapy.data.Observations`
+            The list of `Observations` to concatenate.
+
+        Returns
+        -------
+        observations : `~gammapy.data.Observations`
+            The `Observations` object resulting from the concatenation of all the `Observations` in `observation_list`.
+        """
+        return cls(list(np.concatenate(observations_list)))
```
bkhelifi, line 732: Should we not add a test function??
if an obs_id is common, check if the metadata are different. Unfortunately there are no meta data in observation, e.g. DST_VER, ANA_VER, CAL_VER, EVENT_CLASS, EVENT_TYPE, INSTRUME

registerrier (commented):
Thanks @MRegeard . 

I am not sure I see the point here. The gain compared to `__iadd__` is limited and this approach does prevent inserting two identical observations in the the same `Observations`.

gammapy/data/observations.py
AtreyeeS: What is the advantage over `observations.concatenate(observations_extra)` ?

MRegeard:
> Thanks @MRegeard .
> 
> I am not sure I see the point here. The gain compared to `__iadd__` is limited and this approach does prevent inserting two identical observations in the the same `Observations`.

Hi @registerrier, I agree that this doesn't change a lot as compared to `__iad__`. 
However, I think that this as two major advantage :

- First, it will appear in the documentation. Currently, the `__iadd__` method does not appear directly in the documentation because it is a built-in method of `Sequence`. So if you don't know that it exists (typically you don't know a lot of python, which I think is the case of non negligible fraction of gammapy users), you cannot "concatenate" `Observations`. 

- Moreover, the big difference with `__iadd__` is that it create a new `Observations`. I think that this is less dangerous than 'iadding' on one of the `Observations` that you have.

I agree that in any case it would be good to be able to check that in an `Observations` object there are not multiple occurrence of the same `Observation`.

QRemy:
This is similar to the `from_stack` method we have for other container classes, maybe it should be renamed  and work in the same way : 
https://docs.gammapy.org/dev/_modules/gammapy/modeling/parameter.html#Parameters.from_stack

MRegeard:
> This is similar to the `from_stack` method we have for other container classes, maybe it should be renamed and work in the same way : https://docs.gammapy.org/dev/_modules/gammapy/modeling/parameter.html#Parameters.from_stack

True !
It's seems to be a better wording.

gammapy/data/tests/test_observations.py
```diff
@@ -501,3 +507,15 @@ def test_filter_live_time_phase(data_store):
...
 
     assert_allclose(live_time_filter, default_obs_live_time * (0.8 - 0.2))
+
+
+@requires_data()
+def test_observations_concatenate(data_store):
+    obs_1 = data_store.get_observations([20136, 20137, 20151])
+    obs_2 = data_store.get_observations([20275, 20282])
+
+    concatenate_obs = Observations.concatenate([obs_1, obs_2])
```
AtreyeeS, line 517: Should it not test `from_stack` here?

MRegeard: Ah yes !
Thanks !

AtreyeeS (approved):
Thanks @MRegeard! It would be good to expose this in a tutorial.

gammapy/data/observations.py
```diff
@@ -715,6 +716,23 @@ def group_by_label(self, labels):
...
+        ----------
+        observations_list : list of `~gammapy.data.Observations`
+            The list of `Observations` to concatenate.
+
+        Returns
+        -------
+        observations : `~gammapy.data.Observations`
+            The `Observations` object resulting from the concatenation of all the `Observations` in `observation_list`.
+        """
+        obs = itertools.chain(*observations_list)
```
bkhelifi, line 733: I would propose to add a function `duplication_check` before the concatenation, because this function is soooo dangerous.
Today, only the obs_id can be checked. Later on,  when we will have metadata, better checks  could be added.

MRegeard:
I added a function that check if the list of ids is unique. 
If it is not the case, I simply raise a warning that informe the user. 
Is it ok like this or should I raise an Error?

registerrier (commented):
Thanks @MRegeard .

I am not sure that we can or should at the moment impose uniqueness in the `Observations`. There are many other ways this can fail and what you introduce it does not cover all cases.
 
I can still do:
 ```
data_store = DataStore.from_dir("$GAMMAPY_DATA/hess-dl3-dr1/")
observations = data_store.get_observations([20136, 20137, 20151])
observations.append(observations[0])
```

In many places, you might have different `Observation` objects sharing the same `obs_id`. For instance if you split an observation by event type, or in time/phase. Checking for uniqueness of `obs_id` is not the correct approach.

One could check that objects are not identical, but even this does not work, because the same observation can be loaded twice (with 2 different in-memory object). For instance, you can do this:
```
data_store = DataStore.from_dir("$GAMMAPY_DATA/hess-dl3-dr1/")
observations = data_store.get_observations([20136, 20137, 20151, 20136])
print(observations[0] == observations[-1])
```
yields `False`.

**Comment by bkhelifi** [comment]
Hi Régis,
I think I do not understand fully your points.
"In many places, you might have different Observation objects sharing the same obs_id". 
You mention event type, but we don't have release with different event types. Once, we will have, I really hope that the event type ID will be stored as metadata somewhere... and then, at that time, one could add checks on metadata.
Today, I do not see a use case with the same obs_id in the Observations list.
You mention also time filtering... I hope that each small observation stores in its gti the new time interval. And then, you're right, one could add in the check function a gti check. This use case has to be handle, indeed...
In the check function, the concatenation is always possible... it just sends a warning, which is fair enough.

**Comment by registerrier** [comment]
> Once, we will have, I really hope that the event type ID will be stored as metadata somewhere... and then, at that time, one could add checks on metadata. (...)  one could add in the check function a gti check. 

The issue is that there are many possible ways to check that observations are identical.  Such a check mechanism will complexify with time and be backward incompatible. Introducing configurability will be painful (e.g. `check_uniqueness(check_obsid, check_gti, check_type, check_phase)` ). So I think this solution is worse than the problem itself.

As usual, we cannot provide a flexible but fool-proof scheme. We should try to correct real bugs and document the rest. Here, it seems to me that:

- `DataStore.get_observations(obs_ids)` should check that obs_ids are unique and emit a warning if not. At least, there is a requirement in the GADF that `OBS_ID` be unique to a given observation.
- If there is a consensus to make sure that a single observation cannot appear twice in an `Observations`, one would have to check that in-memory objects are not identical in `__setitem__` and `insert`. For instance, with something like:
```
def insert(self, idx, obs):
    if isinstance(obs, Observation):
        if obs in self:
            raise ValueError(f"{obs.obs_id} already belongs to Observations.")
        self._observations.insert(idx, obs)
    else:
        raise TypeError(f"Invalid type: {type(obs)!r}")
```

**Comment by MRegeard** [comment]
I totally agree with you that there could some use case where we would like to have observation with the same ids. 
What I implemented here is just a check that raise a warning to inform the user that there are multiple `Observation` with the same ids. But in any case the staking happened and if the user knows what he is doing he can continue. 

It is aslo true that currently we are not checking anything on this class. This is a bit surprising to me considering it is the containers class that is used in almost all the analyses before going to the dataset making. 
I am pretty sure that if we want to introduce proper check, the check function that I propose is irrelevant. But in the mean time, I would says that it is probably better that nothing, and it does not restrict in any case what the user can do with the class. 

Then of course, the question is why only doing this on this method. Considering our workflow, where we usually extract the `Observations` from the `DataStore` (often using the obs_table), it seems less probable to get an `Observations` object with some duplicates. While when stacking `Observations`, It is easy to mess up with the variable names, etc. At least that is the feeling that I have.

**Comment by MRegeard** [comment]
I added some check on `__init__` and `insert` and remove the check function that I had created for `from_stack` .
Now the `__init__` check that first all the elements of the input list are instance of `Observation`, and then that all the observations in the list are in-memory unique. If it is not the case, the later check will only raise a warning to inform the user. 
The check on `insert` also raise a warning if the `Observation` to insert is already inside the `Observations` object (again in-memory the same).

File: gammapy/data/observations.py [hunk 4]
```diff
@@ -621,7 +622,17 @@ class Observations(collections.abc.MutableSequence):
     """
 
     def __init__(self, observations=None):
-        self._observations = observations or []
+        if all(isinstance(obs, Observation) for obs in observations or []):
```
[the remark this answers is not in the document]
registerrier: Then the warning and the exception become unnecessary in the `__init__`

gammapy/data/observations.py
```diff
@@ -637,6 +648,8 @@ def __setitem__(self, key, obs):
 
     def insert(self, idx, obs):
         if isinstance(obs, Observation):
+            if obs in self:
```
registerrier, line 651: I think the `__setitem__` method should have the same test as well.

registerrier (commented):
Thanks @MRegeard . I have left a few additional comments to simplify the `__init__` function and make sure that `__setitem__` is also ckeched.

gammapy/data/observations.py
```diff
@@ -637,6 +648,8 @@ def __setitem__(self, key, obs):
...
+                log.warning(f"{obs.obs_id} already belongs to Observations.")
```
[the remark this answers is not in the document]
registerrier: You should use the `caplog` fixture of pytest.
See https://docs.gammapy.org/1.1/development/dev_howto.html#caplog-fixture

It is used in several places in the tests.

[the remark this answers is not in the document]
MRegeard: Oh thank you, I missed that in the doc ...

MRegeard:
Should I make proper test for `insert` and `__setitem__` warning/error ?

gammapy/data/tests/test_observations.py
```diff
@@ -501,3 +507,26 @@ def test_filter_live_time_phase(data_store):
...
+    obs_1 = data_store.get_observations([20136, 20137, 20151])
+    obs_2 = data_store.get_observations([20275, 20282])
+
+    obs12 = Observations.from_stack([obs_1, obs_2])
+
+    assert len(obs12) == 5
+    assert isinstance(obs12[0], Observation)
+    assert isinstance(obs12[0].events, EventList)
+
+    assert "WARNING" in [_.levelname for _ in caplog.records]
```
registerrier, line 523:
```suggestion
    obs122 = Observations.from_stack([obs12, obs2])
    
    assert len(obs122) == 7 
    assert "WARNING" in [_.levelname for _ in caplog.records]
```

gammapy/data/tests/test_observations.py
```diff
@@ -501,3 +507,26 @@ def test_filter_live_time_phase(data_store):
...
+@requires_data()
+def test_stack_observations(data_store, caplog):
+    obs_1 = data_store.get_observations([20136, 20137, 20151])
+    obs_2 = data_store.get_observations([20275, 20282])
+
+    obs12 = Observations.from_stack([obs_1, obs_2])
+
+    assert len(obs12) == 5
+    assert isinstance(obs12[0], Observation)
+    assert isinstance(obs12[0].events, EventList)
```
registerrier, line 521: I don't think checking on `events` is really useful here.

registerrier, line 521:
```suggestion
```

gammapy/data/tests/test_observations.py
```diff
@@ -501,3 +507,33 @@ def test_filter_live_time_phase(data_store):
...
+    assert len(obs122) == 7 
+    assert "WARNING" in [_.levelname for _ in caplog.records]
+    assert "Observation with obs_id 20275 already belongs to Observations." in [
+        _.message for _ in caplog.records
+    ]
+    
+    caplog.clear()
+    obs[3] = obs[0]
+    
+    assert "WARNING" in [_.levelname for _ in caplog.records]
```
registerrier, line 533:
```suggestion

    caplog.clear()
    obs[3] = obs[0]

    assert "WARNING" in [_.levelname for _ in caplog.records]
```

registerrier:
Thanks @MRegeard . I have implemented the few remaining comments. 
The CI fails are unrelated. Merging now.

Issue #4602, opened 2023-06-21
Piecewise Norm Spatial Model problems at Galactic Longitude Coordinate 0 deg


katrinstreil:
**Gammapy version**
1.1

**Bug description**

I followed the example plot for the piecewise Norm Spatial Model:
https://docs.gammapy.org/dev/user-guide/model-gallery/spatial/plot_piecewise_norm_spatial.html

If changing the sky coordinate of the geom so that it contains the Galactic Center i.e. (0,0) deg the model evaluation returns NaN. 

**To Reproduce**
Using 
`geom = WcsGeom.create(skydir=(0., 0.), npix=(120, 120), binsz=0.03, frame="galactic")`
in the linked example plot

QRemy:
a simple fix to work at the GC would be to redefine the coords such as:
`coords["lon"][coords["lon"]>180*u.deg]-=360*u.deg`

Astro-Kirsty:
I had a similar idea but using the `wrap_at` function 
`self._coords['lon']=Angle(self._coords.lon).wrap_at(180*u.deg)`
However, then the plot looks like this 
![image](https://github.com/gammapy/gammapy/assets/60767360/b589404c-f76a-4830-bbfb-0804839300e3)

Astro-Kirsty:
I did something similar using `wrap_at` function
`self._coords['lon']=Angle(self._coords.lon).wrap_at(180*u.deg)`

Will make a PR soon

PR #4652, opened 2023-07-12
Adapt `PiecewiseNormSpatialModel` to wrap the longitude

Astro-Kirsty:
**Description**
This PR is to fix [#4602](https://github.com/gammapy/gammapy/issues/4602)
Adapted so that the code 'wraps' the Galactic longitude at 180 deg, so there is no longer an issue at the Galactic Centre. 

**Dear reviewer**
Hoping for some feedback. 
I tried to change the test for the `PiecewiseNormSpatialModel` such that it is testing at the GC, however, the following line of code does not return the expected
`assert_allclose(model(*geom.to_image().get_coord()), expected, atol=1e-5)`

The other tests work fine. So any explanation on this would be good

### DIFF
--- a/gammapy/modeling/models/spatial.py
+++ b/gammapy/modeling/models/spatial.py
@@ -1311,7 +1311,8 @@ class PiecewiseNormSpatialModel(SpatialModel):
 
     def __init__(self, coords, norms=None, interp="lin", **kwargs):
         self._coords = coords
-        self._coords["lon"] = Angle(self._coords.lon).wrap_at(180 * u.deg)
+        self._wrap_angle = (coords.lon.max() - coords.lon.min()) / 2
+        self._coords["lon"] = Angle(coords.lon).wrap_at(self._wrap_angle)
         self._interp = interp
 
         if norms is None:
@@ -1357,6 +1358,7 @@ class PiecewiseNormSpatialModel(SpatialModel):
         coords = [value.value for value in self.coords._data.values()]
         # TODO: apply axes scaling in this loop
         coords = list(zip(*coords))
+        lon = Angle(lon).wrap_at(self._wrap_angle)
         # by default rely on CloughTocher2DInterpolator
         # (Piecewise cubic, C1 smooth, curvature-minimizing interpolant)
         interpolated = griddata(coords, v_nodes, (lon, lat), method="cubic")
@@ -1377,7 +1379,7 @@ class PiecewiseNormSpatialModel(SpatialModel):
 
         """
         coords = geom.get_coord(frame=self.frame, sparse=True)
-        return self(Angle(coords.lon).wrap_at(180 * u.deg), coords.lat)
+        return self(coords.lon, coords.lat)
 
     def to_dict(self, full_output=False):
         data = super().to_dict(full_output=full_output)
--- a/gammapy/modeling/models/spatial.py
+++ b/gammapy/modeling/models/spatial.py
@@ -1311,6 +1311,7 @@ class PiecewiseNormSpatialModel(SpatialModel):
 
     def __init__(self, coords, norms=None, interp="lin", **kwargs):
         self._coords = coords
+        self._coords["lon"] = Angle(self._coords.lon).wrap_at(180 * u.deg)
         self._interp = interp
 
         if norms is None:
@@ -1376,7 +1377,7 @@ class PiecewiseNormSpatialModel(SpatialModel):
 
         """
         coords = geom.get_coord(frame=self.frame, sparse=True)
-        return self(coords.lon, coords.lat)
+        return self(Angle(coords.lon).wrap_at(180 * u.deg), coords.lat)
 
     def to_dict(self, full_output=False):
         data = super().to_dict(full_output=full_output)
--- a/gammapy/modeling/models/spatial.py
+++ b/gammapy/modeling/models/spatial.py
@@ -1310,7 +1310,8 @@ class PiecewiseNormSpatialModel(SpatialModel):
     tag = ["PiecewiseNormSpatialModel", "piecewise-norm"]
 
     def __init__(self, coords, norms=None, interp="lin", **kwargs):
-        self._coords = coords
+        self._coords = coords.copy()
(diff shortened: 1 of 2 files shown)


gammapy/modeling/models/spatial.py
```diff
@@ -1376,7 +1377,7 @@ def evaluate_geom(self, geom):
 
         """
         coords = geom.get_coord(frame=self.frame, sparse=True)
-        return self(coords.lon, coords.lat)
+        return self(Angle(coords.lon).wrap_at(180 * u.deg), coords.lat)
```
QRemy, line 1380: this fix should be applied to the whitin `evaluate` function adding lon=Angle(lon).wrap_at(self.wrap_angle)`

gammapy/modeling/models/spatial.py
```diff
@@ -1311,6 +1311,7 @@ class PiecewiseNormSpatialModel(SpatialModel):
 
     def __init__(self, coords, norms=None, interp="lin", **kwargs):
         self._coords = coords
+        self._coords["lon"] = Angle(self._coords.lon).wrap_at(180 * u.deg)
```
QRemy, line 1314: this will move the problem at the anticenter, the wrap_angle could be computed such as is is always outside the range of the input coords maybe
self._wrap_angle = (coords.lon.max()-coords.lon.min())/2

gammapy/modeling/models/spatial.py
QRemy, line 1380: this fix should be applied to the whitin evaluate function adding lon=Angle(lon).wrap_at(self._wrap_angle)`

QRemy (changes requested):
Thanks @Astro-Kirsty, I added few suggestions to make the fix more general

gammapy/modeling/models/spatial.py
Astro-Kirsty: Okay, that is a good idea

gammapy/modeling/models/spatial.py
Astro-Kirsty: Do you mean inside the `evaluate_geom` function? As far as I understand, this is an issue with the geom not the evaluation?

QRemy: `evaluate_geom ` call `evaluate` and the problem already appears here if the given lon is not valid
https://github.com/gammapy/gammapy/blob/fc886b94357e131de35787657c7592f7d42f232e/gammapy/modeling/models/spatial.py#L1352C9-L1352C17

Astro-Kirsty: Oh yes, sorry. 
I have fixed it. Should I update the test to be at the GC, or leave it as it is?

QRemy: yes you can update the existing test to be at the GC

QRemy: I added tests at galactic center an anticenter and I don't see any fails, same trying the doc example in these regions.
Do you have an simple example to reproduce the fail ?

bkhelifi:
@QRemy, can we add a test with a fov frame? see PR #4667 ? Just to be sure that it works well...

gammapy/modeling/models/spatial.py
Astro-Kirsty: The issue occurs when you chose the GC: 
`geom = WcsGeom.create(skydir=(0, 0), npix=(2, 2), binsz=0.3, frame="galactic")`

Then `new_model.evaluate_geom(geom.to_image()) = [[NaN, 3], [NaN, 2]]`

QRemy:
> @QRemy, can we add a test with a fov frame? see PR #4667 ? Just to be sure that it works well...

Maybe we need a FoVGeom for that.
 I was able to create a `WcsGeom` with a `SkyOffsetFrame` but in `evaluate_geom` I get an error message from:
 ```
 coords = geom.get_coord(frame=self.frame, sparse=True)

TypeError: unhashable type: 'SkyOffsetICRS'
 ```

gammapy/modeling/models/spatial.py
QRemy: ok the definition of the `wrap_angle` was not working if the wrap angle was not 0 in the input coords, now it should work for all longitude and I added a test that loop over -360<l<360

gammapy/modeling/models/spatial.py
```diff
@@ -1310,7 +1310,10 @@ class PiecewiseNormSpatialModel(SpatialModel):
     tag = ["PiecewiseNormSpatialModel", "piecewise-norm"]
 
     def __init__(self, coords, norms=None, interp="lin", **kwargs):
-        self._coords = coords
+        self._coords = coords.copy()
```
Astro-Kirsty, line 1313: What is the purpose of copying the coords?

Astro-Kirsty:
Another comment I had, is that this spatial model will only work for a 'square' geometry. i.e `npix=(100, 120)` gives a result with no values in the y from 100-120 pixels. 
This is something that either needs to be fixed or documented?

gammapy/modeling/models/spatial.py
QRemy: to avoid the modification of the input coords

QRemy:
> Another comment I had, is that this spatial model will only work for a 'square' geometry. i.e `npix=(100, 120)` gives a result with no values in the y from 100-120 pixels. This is something that either needs to be fixed or documented?

In the example the model edges are defined by `geom.footprint` on init but is seems that this does not match the edges of the map so maybe the problem comes from `geom.wcs. calc_footprint()`

registerrier:
> In the example the model edges are defined by `geom.footprint` on init but is seems that this does not match the edges of the map so maybe the problem comes from `geom.wcs.calc_footprint()`

Yes I think there is a bug in `geom.footprint`. Doesn't it reverse the order?

Astro-Kirsty:
@QRemy, can we merge this now? 
I think this problem is solved and the issue now comes from the bug in the `geom.footprint` ?

PR #4653, opened 2023-07-12
Fix points scaling in TemplateNDSpectralModel

QRemy:
Remove `points_scale` from the `interp_kwargs` because the transform is already applied by axis.coord_to_pix.


registerrier (approved):
Thanks @QRemy . Looks good.

The tests differences are due to the scaling being applied twice in the previous version?

PR #4654, opened 2023-07-12
Add TemplateNDSpatialModel

QRemy:
Add TemplateNDSpatialModel : a model generated from a ND map where extra dimensions define the parameter space.

### DIFF
--- a/gammapy/modeling/models/spatial.py
+++ b/gammapy/modeling/models/spatial.py
@@ -18,7 +18,7 @@ from regions import (
     RectangleSkyRegion,
 )
 import matplotlib.pyplot as plt
-from gammapy.maps import Map, MapCoord, WcsGeom, WcsNDMap
+from gammapy.maps import HpxNDMap, Map, MapCoord, WcsGeom, WcsNDMap
 from gammapy.modeling import Parameter, Parameters
 from gammapy.utils.deprecation import deprecated
 from gammapy.utils.gauss import Gauss2DPDF
@@ -1318,6 +1318,8 @@ class TemplateNDSpatialModel(SpatialModel):
         filename=None,
         copy_data=True,
     ):
+        if not isinstance(map, (HpxNDMap, WcsNDMap)):
+            raise TypeError("Map should be a HpxNDMap or WcsNDMap")
         if copy_data:
             self._map = map.copy()
         else:
@@ -1386,7 +1388,7 @@ class TemplateNDSpatialModel(SpatialModel):
     def from_dict(cls, data):
         data = data["spatial"]
         filename = data["filename"]
-        m = WcsNDMap.read(filename)
+        m = Map.read(filename)
         model = cls(m, filename=filename)
         for idx, p in enumerate(model.parameters):
             par = p.to_dict()
--- a/gammapy/modeling/models/spatial.py
+++ b/gammapy/modeling/models/spatial.py
@@ -1395,7 +1395,7 @@ class TemplateNDSpatialModel(SpatialModel):
         return model
 
     def to_dict(self, full_output=False):
-        """Create dict for YAML serilisation"""
+        """Create dict for YAML serialisation"""
         data = super().to_dict(full_output)
         data["spatial"]["filename"] = self.filename
         data["spatial"]["unit"] = str(self.map.unit)
--- a/gammapy/modeling/models/spatial.py
+++ b/gammapy/modeling/models/spatial.py
@@ -1295,7 +1295,7 @@ class TemplateNDSpatialModel(SpatialModel):
 
     Parameters
     ----------
-    map : `~gammapy.maps.WcsNDMap`
+    map : `~gammapy.maps.WcsNDMap` or `~gammapy.maps.HpxNDMap`
         Map template.
     meta : dict, optional
         Meta information, meta['filename'] will be used for serialization
@@ -1354,7 +1354,7 @@ class TemplateNDSpatialModel(SpatialModel):
 
     @property
     def map(self):
-        """Template map  (`~gammapy.maps.WcsNDMap`)"""
+        """Template map  (`~gammapy.maps.WcsNDMap` or `~gammapy.maps.HpxNDMap`)"""
         return self._map
 
     @property
--- a/gammapy/modeling/models/spatial.py
+++ b/gammapy/modeling/models/spatial.py
@@ -18,7 +18,7 @@ from regions import (
     RectangleSkyRegion,
 )
 import matplotlib.pyplot as plt
-from gammapy.maps import Map, MapCoord, WcsGeom
+from gammapy.maps import Map, MapCoord, WcsGeom, WcsNDMap
 from gammapy.modeling import Parameter, Parameters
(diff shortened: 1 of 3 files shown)


registerrier (commented):
Thanks @QRemy ! 

A few general comments and questions:
- the functionality is very similar to the `TemplateNDSpectralModel` that you introduced in #3535 .  The `evaluate` methods are specific but large parts of the code are very similar. Are there ways to avoid code duplication here?
- I am not sure we have examples documenting the usage of the spectral ND template. If we add more such models we have to think on example usage.
- Regular template models could be considered simple cases of ND models. Is it possible to also share some functionalities there?

QRemy:
> * the functionality is very similar to the `TemplateNDSpectralModel` that you introduced in [Add TemplateNDSpectralModel #3535](https://github.com/gammapy/gammapy/pull/3535) .  The `evaluate` methods are specific but large parts of the code are very similar. Are there ways to avoid code duplication here?

> * Regular template models could be considered simple cases of ND models. Is it possible to also share some functionalities there?

I agree it would be nice to have only one model for the templates but I would prefer to do that in a separate PR.

> * I am not sure we have examples documenting the usage of the spectral ND template. If we add more such models we have to think on example usage.

For now it is not documented but In the tests we have an example to create an EBL model using the TemplateNDSpectralModel, so if we add something like `RegionNDMap.read(format="xspec")` or `TemplateNDSpectralModel.read(format="xspec")` then the EBL model could be removed (or internally rely on the TemplateNDSpectralModel)
https://github.com/gammapy/gammapy/blob/d44a1c240dfab9bf9a44e7ce066cc534f3ea6365/gammapy/modeling/models/tests/test_spectral.py#L1119

For the `TemplateNDSpatialModel` the use case I have in mind is to generate a small sample of diffuse models with hermes varing the parameters of the radial profiles in CR sources or spectral index...

gammapy/modeling/models/spatial.py
```diff
@@ -1289,6 +1290,104 @@ def plot_interactive(self, ax=None, geom=None, **kwargs):
...
+        Meta information, meta['filename'] will be used for serialization
+    interp_kwargs : dict
+        Interpolation keyword arguments passed to `gammapy.maps.Map.interp_by_pix`.
+        Default arguments are {'method': 'linear', 'fill_value': 0, "values_scale": "log"}.
+    """
+
+    tag = ["TemplateNDSpatialModel", "templateND"]
+
+    def __init__(self, map, interp_kwargs=None, meta=None, filename=None):
+        self._map = map.copy()
```
registerrier, line 1310: Shouldn't you check the type of the `Map`? One should be able to evaluate on spatial quantities.

gammapy/modeling/models/spatial.py
```diff
@@ -1289,6 +1290,104 @@ def plot_interactive(self, ax=None, geom=None, **kwargs):
...
+        elif os.path.isfile(self.filename) and not overwrite:
+            log.warning("Template file already exits, and overwrite is False")
+        else:
+            self.map.write(self.filename)
+
+    @classmethod
+    def from_dict(cls, data):
+        data = data["spatial"]
+        filename = data["filename"]
+        m = WcsNDMap.read(filename)
```
registerrier, line 1375: Can't there be other types? e.g. `HpxNDMap`? You could use `Map.read` instead.

Otherwise you can create a model with such a map, write it but not read it back.

gammapy/modeling/models/spatial.py
```diff
@@ -1289,6 +1290,104 @@ def plot_interactive(self, ax=None, geom=None, **kwargs):
...
+        model = cls(m, filename=filename)
+        for idx, p in enumerate(model.parameters):
+            par = p.to_dict()
+            par.update(data["parameters"][idx])
+            setattr(model, p.name, Parameter(**par))
+        return model
+
+    def to_dict(self, full_output=False):
+        """Create dict for YAML serilisation"""
```
registerrier, line 1384:
```suggestion
        """Create dict for YAML serialisation"""
```

gammapy/modeling/models/tests/test_spatial.py
```diff
@@ -582,3 +583,45 @@ def test_piecewise_spatial_model_3d():
...
+    assert len(template.parameters) == 2
+    assert_allclose(template.parameters["norm"].value, 5)
+    assert_allclose(template.parameters["cste"].value, 0)
+    assert_allclose(
+        template.evaluate(
+            geom2d.center_skydir.ra, geom2d.center_skydir.dec, norm=0, cste=0
+        ),
+        [0],
+    )
+    assert_allclose(template.evaluate_geom(geom2d), 5 * map_.data, rtol=0.2)
```
registerrier, line 614: Why such a large tolerance?

registerrier (approved):
Thanks @QRemy . This looks good. I have left some inline comments, in particular concerning the types of `Map` used. 

Of course the amount of code duplication with `TemplateNDSpectralModel` is very large. I suggest that once this is merged, you open an follow-up issue to clarify the code structure with the various template models.

gammapy/modeling/models/spatial.py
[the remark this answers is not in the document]
QRemy: added `copy_data` option as for `TemplateSpatialModel`

gammapy/modeling/models/tests/test_spatial.py
QRemy: changing the interpolation values_scales to linear we have now rtol <0.03

gammapy/modeling/models/spatial.py
```diff
@@ -1289,6 +1290,120 @@ def plot_interactive(self, ax=None, geom=None, **kwargs):
         super().plot_interactive(ax=ax, geom=geom, **kwargs)
 
 
+class TemplateNDSpatialModel(SpatialModel):
+    """A model generated from a ND map where extra dimensions define the parameter space.
+
+    Parameters
+    ----------
+    map : `~gammapy.maps.WcsNDMap`
```
registerrier, line 1298:
```suggestion
    map : `~gammapy.maps.WcsNDMap` or `~gammapy.maps.HpxNDMap`
```

gammapy/modeling/models/spatial.py
```diff
@@ -1289,6 +1290,120 @@ def plot_interactive(self, ax=None, geom=None, **kwargs):
...
+        interp_kwargs = interp_kwargs or {}
+        interp_kwargs.setdefault("method", "linear")
+        interp_kwargs.setdefault("fill_value", 0)
+        interp_kwargs.setdefault("values_scale", "log")
+        self._interp_kwargs = interp_kwargs
+        super().__init__()
+
+    @property
+    def map(self):
+        """Template map  (`~gammapy.maps.WcsNDMap`)"""
```
registerrier, line 1357:
```suggestion
        """Template map  (`~gammapy.maps.WcsNDMap` or `~gammapy.maps.HpxNDMap`)"""
```

registerrier (approved):
Thanks @QRemy ! Looks good.

PR #4657, opened 2023-07-12
Fix the FluxProfileEstimator to take into account models

### LINKED ISSUE #4650: Models are lost in FluxProfileEstimator
**Gammapy version**
gammapy 1.0.1

**Bug description**
All models attached to my datasets are discarded when I calculate a flux profile with `FluxProfileEstimator`.

**Expected behavior**
The predicted counts from the models should become part of the background when `to_spectrum_datasets` is called in `FluxProfileEstimator.run()`

QRemy:
This PR fixes the FluxProfileEstimator such as the models attached to the dataset are taken into account.
Resolves #4650

### DIFF
--- a/gammapy/estimators/points/profile.py
+++ b/gammapy/estimators/points/profile.py
@@ -134,7 +134,7 @@ class FluxProfileEstimator(FluxPointsEstimator):
             Profile flux points.
         """
 
-        datasets_copy = Datasets(datasets=datasets)
+        datasets = Datasets(datasets=datasets)
         datasets_copy = datasets.copy()
         for dataset in datasets_copy:
             dataset.background = dataset.npred()
--- a/gammapy/estimators/points/profile.py
+++ b/gammapy/estimators/points/profile.py
@@ -133,12 +133,15 @@ class FluxProfileEstimator(FluxPointsEstimator):
         profile : `~gammapy.estimators.FluxPoints`
             Profile flux points.
         """
-        datasets = Datasets(datasets=datasets)
+
+        datasets_copy = Datasets(datasets=datasets)
+        datasets_copy = datasets.copy()
+        for dataset in datasets_copy:
+            dataset.background = dataset.npred()
 
         maps = []
-
         for region in self.regions:
-            datasets_to_fit = datasets.to_spectrum_datasets(region=region)
+            datasets_to_fit = datasets_copy.to_spectrum_datasets(region=region)
             datasets_to_fit.models = SkyModel(self.spectrum, name="test-source")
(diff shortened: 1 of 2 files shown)


gammapy/estimators/points/profile.py
```diff
@@ -133,12 +133,15 @@ def run(self, datasets):
         profile : `~gammapy.estimators.FluxPoints`
             Profile flux points.
         """
+
         datasets = Datasets(datasets=datasets)
+        datasets_copy = datasets.copy()
```
bkhelifi, line 138: Why is a copy needed. beyond the fact of an increased memory usage?

gammapy/estimators/points/profile.py
```diff
@@ -133,12 +133,15 @@ def run(self, datasets):
...
+        for dataset in datasets_copy:
+            dataset.background = dataset.npred()
```
bkhelifi, line 140: I have not catched the logics behind...

gammapy/estimators/points/profile.py
AtreyeeS: > Why are initially models lost?

@bkhelifi I think this is our design, that dataset manipulation discards the models, which is the case here on calling `to_spectrum_dataset`.

AtreyeeS (approved):
Thank @QRemy ! 
Maybe we need to support the old behaviour as well? Add a `models_contribution=False` argument for backwards consistency?

gammapy/estimators/points/profile.py
QRemy: Let's say that you have a point source nearby your extended source of interest. Instead of masking the point source you can include it in the models. The simplest way to do that is to add the npred to the background (because the other source won't be reoptimized as it would be too degenerated in spectrum dataset used internally by the estimator).

gammapy/estimators/points/profile.py
QRemy: modified to avoid copy

QRemy:
> Maybe we need to support the old behaviour as well? Add a `models_contribution=False` argument for backwards consistency?

I would say that the old behaviour is a bug because for the user it is impossible to know that the FluxPointEstimator keep the models while the FluxProfileEstimator doesn't.

gammapy/estimators/points/profile.py
AtreyeeS: It can be useful to document this logic in the docstring

AtreyeeS (approved):
Thanks! No further comments

registerrier (approved):
Thanks @Qremy. No further comments.

PR #4658, opened 2023-07-13
Add utility function to find containment region of maps

QRemy:
Add two functions in utils :
`containment_region` that find the iso-contours region corresponding to a given containment for a 2d map.
`containment_radius` that compute containment radius from the center of the map.

The use case is for example to find  the containment surface and radius of the (psf-convolved) models.

### DIFF
--- a/gammapy/maps/measure.py
+++ b/gammapy/maps/measure.py
@@ -8,7 +8,7 @@ import matplotlib.pyplot as plt
 
 def containment_region(map_, fraction=0.393, apply_union=True):
     """Find the iso-contours region corresponding to a given containment
-        for a map of integral quantities.
+        for a map of integral quantities with a flat geometry.
 
     Parameters
     ----------
@@ -25,6 +25,13 @@ def containment_region(map_, fraction=0.393, apply_union=True):
     regions : list of ~regions.PolygonSkyRegion` or `~regions.CompoundSkyRegion`
         regions from iso-contours matching containment fraction.
     """
+    from . import WcsNDMap
+
+    if not isinstance(map_, WcsNDMap):
+        raise TypeError(
+            f"This function is only supported for WcsNDMap, got {type(map_)} instead."
+        )
+
     fmax = np.nanmax(map_.data)
     if fmax > 0.0:
         ordered = np.sort(map_.data.flatten())[::-1]
@@ -73,12 +80,15 @@ def containment_region(map_, fraction=0.393, apply_union=True):
 
 
 def containment_radius(map_, fraction=0.393, position=None):
-    """Compute containment radius from the center of a map with integral quantities.
+    """Compute containment radius from a position in a map
+        with integral quantities and flat geometry.
 
     Parameters
     ----------
+    map_ : `~gammapy.maps.WcsNDMap`
+        Map of integral quantities.
     fraction : float
-        Containment fraction.
+        Containment fraction. Default is 0.393.
     position : `~astropy.coordinates.SkyCoord`
         Position from where the containment is computed.
         Default is the center of the Map.
@@ -89,6 +99,12 @@ def containment_radius(map_, fraction=0.393, position=None):
         Minimal radius required to reach the given containement fraction.
 
     """
+    from . import WcsNDMap
+
+    if not isinstance(map_, WcsNDMap):
+        raise TypeError(
+            f"This function is only supported for WcsNDMap, got {type(map_)} instead."
+        )
 
     if position is None:
         position = map_.geom.center_skydir
--- a/gammapy/maps/measure.py
+++ b/gammapy/maps/measure.py
@@ -1,0 +1,91 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+import numpy as np
+from astropy.coordinates import SkyCoord
+from regions import PolygonSkyRegion
+import matplotlib.pyplot as plt
+
+
+def containment_region(map_, fraction=0.68, apply_union=True):
+    """Find the iso-contours region corresponding to a given containment
+        for a map of integral quantities.
+
+    Parameters
+    ----------
+    map_ : `~gammapy.maps.WcsNDMap`
+        Map of integral quantities
+    fraction : float
+        Containment fraction
+    apply_union : bool
+        It True return a compound region otherwise return a list of polygon regions.
+        Default is True. Note that compound regions cannot be written in ds9 format but can always be saved with numpy.savez.
+
+    Returns
+    -------
+    regions : list of ~regions.PolygonSkyRegion` or `~regions.CompoundSkyRegion`
+        regions from iso-contours matching containment fraction
+    """
+    fmax = np.nanmax(map_.data)
+    if fmax > 0.0:
+        ordered = np.sort(map_.data.flatten())[::-1]
+        cumsum = np.nancumsum(ordered)
+        ind = np.argmin(np.abs(cumsum / cumsum.max() - fraction))
+        fval = ordered[ind]
+
+        plt.ioff()
+        fig = plt.figure()
+        cs = plt.contour(map_.data.squeeze(), [fval])
+        plt.close(fig)
+        plt.ion()
+        regions_pieces = []
+        for pp in cs.collections[0].get_paths():
+            vertices = []
+            for v in pp.vertices:
+                v_coord = map_.geom.pix_to_coord(v)
+                vertices.append([v_coord[0], v_coord[1]])
+            vertices = SkyCoord(vertices, frame=map_.geom.frame)
+            regions_pieces.append(PolygonSkyRegion(vertices))
+
+        if apply_union:
+            regions_union = regions_pieces[0]
+            for region in regions_pieces[1:]:
+                regions_union = regions_union.union(region)
+            return regions_union
+        else:
+            return regions_pieces
+    else:
+        raise ValueError("No positive values in the map.")
+
+
+def containment_radius(map_, fraction=0.68, position=None):
+    """Compute containment radius from the center of a map with integral quantities
+
+    Parameters
+    ----------
+    fraction : float
+        Containment fraction
+    n_levels : int
+        Numbers of contours levels used to find the required containment radius.
+    position : `~astropy.coordinates.SkyCoord`
+        Position from where the containment is conputed.
+        Default is the center of the Map.
+
+    Returns
+    -------
+    radius : `~astropy.coordinates.Angle`
+        Minimal radius required to reach the given containement fraction.
+
+    """
+
+    if position is None:
+        position = map_.geom.center_skydir
+
+    fmax = np.nanmax(map_.data)
+    if fmax > 0.0:
+        sep = map_.geom.separation(position).flatten()
+        order = np.argsort(sep)
+        ordered = map_.data.flatten()[order]
+        cumsum = np.nancumsum(ordered)
+        ind = np.argmin(np.abs(cumsum / cumsum.max() - fraction))
+    else:
+        raise ValueError("No positive values in the map.")
+    return sep[order][ind]
--- a/gammapy/maps/measure.py
+++ b/gammapy/maps/measure.py
@@ -13,9 +13,9 @@ def containment_region(map_, fraction=0.393, apply_union=True):
     Parameters
     ----------
     map_ : `~gammapy.maps.WcsNDMap`
-        Map of integral quantities
+        Map of integral quantities.
     fraction : float
-        Containment fraction
+        Containment fraction. Default is 0.393.
     apply_union : bool
         It True return a compound region otherwise return a list of polygon regions.
         Default is True. Note that compound regions cannot be written in ds9 format but can always be saved with numpy.savez.
@@ -23,7 +23,7 @@ def containment_region(map_, fraction=0.393, apply_union=True):
     Returns
     -------
     regions : list of ~regions.PolygonSkyRegion` or `~regions.CompoundSkyRegion`
-        regions from iso-contours matching containment fraction
+        regions from iso-contours matching containment fraction.
     """
     fmax = np.nanmax(map_.data)
     if fmax > 0.0:
@@ -73,16 +73,14 @@ def containment_region(map_, fraction=0.393, apply_union=True):
 
 
 def containment_radius(map_, fraction=0.393, position=None):
-    """Compute containment radius from the center of a map with integral quantities
+    """Compute containment radius from the center of a map with integral quantities.
 
     Parameters
     ----------
     fraction : float
-        Containment fraction
-    n_levels : int
-        Numbers of contours levels used to find the required containment radius.
+        Containment fraction.
     position : `~astropy.coordinates.SkyCoord`
-        Position from where the containment is conputed.
+        Position from where the containment is computed.
         Default is the center of the Map.
 
     Returns
--- a/gammapy/maps/measure.py
+++ b/gammapy/maps/measure.py
@@ -37,9 +37,9 @@ def containment_region(map_, fraction=0.68, apply_union=True):
         plt.close(fig)
         plt.ion()
         regions_pieces = []
-        for pp in cs.collections[0].get_paths():
+        for pp in cs.allsegs[0]:
             vertices = []
-            for v in pp.vertices:
+            for v in pp:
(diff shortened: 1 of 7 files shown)


bkhelifi:
Maybe you could also add this functionality into a tutorial, e.g. the 2D analysis...

gammapy/utils/regions.py
```diff
@@ -275,3 +277,94 @@ def region_circle_to_ellipse(region):
...
+            return regions_pieces
+
+
+def containment_radius(integral_map, fraction=0.68, n_levels=100):
+    """Compute containment radius from the center of a map with integral quantities
+
+    Parameters
+    ----------
+    integral_map : `~gammapy.maps.WcsNDMap`
+        Map of integral quantities
```
registerrier, line 341: Isn't there a positivity requirement to compute the containment this way? The cumulative sum has to be increasing, I presume.

gammapy/utils/regions.py
```diff
@@ -275,3 +277,94 @@ def region_circle_to_ellipse(region):
...
+        Numbers of contours levels used to find the required containment region.
+
+    Returns
+    -------
+    regions : list of ~regions.PolygonSkyRegion` or `~regions.CompoundSkyRegion`
+        regions from iso-contours matching containment fraction
+    """
+    integral_map = integral_map.reduce_over_axes()
+    fmax = np.nanmax(integral_map.data)
+    if fmax != 0.0:
```
registerrier, line 302:
```suggestion
    if fmax > 0.0:
```

gammapy/utils/regions.py
```diff
@@ -275,3 +277,94 @@ def region_circle_to_ellipse(region):
...
+        frange = np.linspace(fmax / n_levels, fmax, n_levels)
+        fsum = integral_map.data.sum()
+        for fval in frange:
+            S = np.sum(integral_map.data[integral_map.data > fval]) / fsum
```
registerrier, line 306: Same comment as below, you could argsort the `integral_map.data`  and compute the cumulative sum.

registerrier (commented):
Thanks @QRemy ! I have left some propositions to improve the code and a number of questions inline .
 
More generally, does this belong to `gammapy.utils.region`? Isn't it better placed in `maps`?

gammapy/utils/regions.py
```diff
@@ -275,3 +277,94 @@ def region_circle_to_ellipse(region):
...
+    """
+    integral_map = integral_map.reduce_over_axes()
+    coords = integral_map.geom.get_coord()
+    grid = SkyCoord(coords["lon"], coords["lat"], frame=integral_map.geom.frame)
+    center = integral_map.geom.center_skydir
+    hwidth = np.max(integral_map.geom.width) / 2.0
+
+    radius = np.nan
+    fmax = np.nanmax(integral_map.data)
+    if fmax != 0.0:
```
QRemy, line 363:
```suggestion
    if fmax > 0.0:
```

gammapy/utils/regions.py
```diff
@@ -275,3 +277,94 @@ def region_circle_to_ellipse(region):
...
+    integral_map : `~gammapy.maps.WcsNDMap`
+        Map of integral quantities
+    fraction : float
+        Containment fraction
+    n_levels : int
+        Numbers of contours levels used to find the required containment radius.
+
+    Returns
+    -------
+    radius : `~regions.CompoundSkyRegion`
```
QRemy, line 349:
```suggestion
    radius : `~astropy.coordinates.Angle`
```

gammapy/utils/regions.py
```diff
@@ -275,3 +277,94 @@ def region_circle_to_ellipse(region):
...
+        Containement radius
+    regions_pieces : list of `~regions.PolygonSkyRegion`
```
QRemy, line 351:
```suggestion
```

gammapy/utils/regions.py
```diff
@@ -151,10 +153,10 @@ class SphericalCircleSkyRegion(CircleSkyRegion):
       TAN WCS and not have true cone select at all?
     """
 
-    def contains(self, skycoord, wcs=None):
+    def contains(integral_map, skycoord, wcs=None):
```
QRemy, line 156:
```suggestion
    def contains(self, skycoord, wcs=None):
```

gammapy/utils/regions.py
QRemy: I would say no because if the map is noisy cutting only negative values will bias the containment upward

gammapy/utils/regions.py
```diff
@@ -275,3 +277,94 @@ def region_circle_to_ellipse(region):
...
+
+    Returns
+    -------
+    radius : `~regions.CompoundSkyRegion`
+        Containement radius
+    regions_pieces : list of `~regions.PolygonSkyRegion`
+
+
+    """
+    integral_map = integral_map.reduce_over_axes()
```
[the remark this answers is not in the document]
QRemy: yes and then it will be safe to add this to map

adonath:
Thanks @QRemy, this is both useful functionality! 

If haven't looked at the implementation in detail, but we had at some point: https://github.com/gammapy/gammapy/blob/v0.8/gammapy/image/measure.py

So my recommendation here would be to introduce `gammapy/maps/measure.py` (or similar) and double check your implementation against the one I linked above.

gammapy/utils/regions.py
QRemy: done it's faster but slightly less precise especially for the radius computation

gammapy/utils/regions.py
```diff
@@ -275,3 +277,94 @@ def region_circle_to_ellipse(region):
...
+    """Compute containment radius from the center of a map with integral quantities
+
+    Parameters
+    ----------
+    integral_map : `~gammapy.maps.WcsNDMap`
+        Map of integral quantities
+    fraction : float
+        Containment fraction
+    n_levels : int
+        Numbers of contours levels used to find the required containment radius.
```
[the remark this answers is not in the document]
QRemy: removed

gammapy/maps/wcs/core.py
```diff
@@ -302,3 +305,87 @@ def _make_hdu_sparse(data, npix, hdu, header):
...
+        fmax = np.nanmax(self.data)
+        if fmax > 0.0:
+            ordered = np.sort(self.data.flatten())[::-1]
+            cumsum = np.nancumsum(ordered)
+            ind = np.argmin(np.abs(cumsum / cumsum.max() - fraction))
+            fval = ordered[ind]
+
+            plt.ioff()
+            fig = plt.figure()
+            cs = plt.contour(self.data.squeeze(), [fval])
```
bkhelifi, line 337:
```suggestion
            cs = plt.contour(self.data.squeeze(), [fval]);
```

gammapy/maps/wcs/core.py
```diff
@@ -302,3 +305,87 @@ def _make_hdu_sparse(data, npix, hdu, header):
...
+            for kp, pp in enumerate(cs.collections[0].get_paths()):
+                vertices = []
+                for v in pp.vertices:
+                    v_coord = self.geom.pix_to_coord(v)
+                    vertices.append([v_coord[0], v_coord[1]])
+                vertices = SkyCoord(vertices, frame=self.geom.frame)
+                regions_pieces.append(PolygonSkyRegion(vertices))
+
+            if apply_union:
+                # compound from union seems not supported to write in ds9 format
```
bkhelifi, line 350: Should we add this ds9 issue into the docstring?

gammapy/maps/wcs/core.py
```diff
@@ -302,3 +305,87 @@ def _make_hdu_sparse(data, npix, hdu, header):
...
+
+        Parameters
+        ----------
+        self : `~gammapy.maps.WcsNDMap`
+            Map of integral quantities
+        fraction : float
+            Containment fraction
+        apply_union : bool
+            It True return a compound region otherwise return a list of polygon regions.
+            Default is True.
```
QRemy, line 321:
```suggestion
            Default is True. Note that compound regions cannot be written in ds9 format but can always be saved with numpy.savez.
```

gammapy/maps/wcs/core.py
```diff
@@ -302,3 +305,87 @@ def _make_hdu_sparse(data, npix, hdu, header):
...
+                for v in pp.vertices:
+                    v_coord = self.geom.pix_to_coord(v)
+                    vertices.append([v_coord[0], v_coord[1]])
+                vertices = SkyCoord(vertices, frame=self.geom.frame)
+                regions_pieces.append(PolygonSkyRegion(vertices))
+
+            if apply_union:
+                # compound from union seems not supported to write in ds9 format
+                # so regions_pieces contains a list that is supported
+                # while regions_full can be saved as .npz
```
QRemy, line 352:
```suggestion
```

gammapy/maps/wcs/core.py
QRemy: done

gammapy/maps/wcs/core.py
```diff
@@ -302,3 +305,84 @@ def _make_hdu_sparse(data, npix, hdu, header):
...
+        fmax = np.nanmax(self.data)
+        if fmax > 0.0:
+            ordered = np.sort(self.data.flatten())[::-1]
+            cumsum = np.nancumsum(ordered)
+            ind = np.argmin(np.abs(cumsum / cumsum.max() - fraction))
+            fval = ordered[ind]
+
+            plt.ioff()
+            fig = plt.figure()
+            cs = plt.contour(self.data.squeeze(), [fval]);
```
QRemy, line 337:
```suggestion
            cs = plt.contour(self.data.squeeze(), [fval])
```

gammapy/maps/wcs/core.py
QRemy: here it is not necessary

gammapy/maps/wcs/core.py
```diff
@@ -302,3 +305,84 @@ def _make_hdu_sparse(data, npix, hdu, header):
...
+            Containement radius
+
+        """
+        coords = self.geom.get_coord()
+        grid = coords.skycoord
+        if position is None:
+            position = self.geom.center_skydir
+
+        fmax = np.nanmax(self.data)
+        if fmax > 0.0:
```
adonath, line 382: This needs a different error handling. If `fmax<=0`, the variables in the if block remain undefined and the return statement fails with a cryptic error. Better check in advance and raise a meaningful error.

gammapy/maps/wcs/core.py
```diff
@@ -302,3 +305,84 @@ def _make_hdu_sparse(data, npix, hdu, header):
...
+            sep = grid.separation(position).flatten()
+            order = np.argsort(sep)
+            ordered = self.data.flatten()[order]
+            cumsum = np.nancumsum(ordered)
+            ind = np.argmin(np.abs(cumsum / cumsum.max() - fraction))
```
adonath, line 387: If you do not require the the cumsum to be strictly increasing (as @registerrier suggested) the solution here is not unique. `argmin` will return the first solution. This at least needs documentation. The need for the `abs` statement is unclear.

adonath (changes requested):
I left some more comments. I think in general it might make more sense to introduce the functionality in an independent namespace `gammapy.maps.measure`. The functionality seems too specific to be on the `Map` object: to be meaningful it requires the data to be specifically integrable...

adonath:
Another reason for an independent namespace: the way it is implemented it should basically work with HEALPix maps as well. However for `RegionNDMap` it does not make sense...

gammapy/maps/wcs/core.py
```diff
@@ -302,3 +305,84 @@ def _make_hdu_sparse(data, npix, hdu, header):
...
+        Returns
+        -------
+        regions : list of ~regions.PolygonSkyRegion` or `~regions.CompoundSkyRegion`
+            regions from iso-contours matching containment fraction
+        """
+        fmax = np.nanmax(self.data)
+        if fmax > 0.0:
+            ordered = np.sort(self.data.flatten())[::-1]
+            cumsum = np.nancumsum(ordered)
+            ind = np.argmin(np.abs(cumsum / cumsum.max() - fraction))
```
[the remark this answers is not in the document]
QRemy: 2 lines are the same, the others are slightly different so I think it is fine to have a bit of duplication here, at least for now.

bkhelifi:
Thanks @QRemy... It seems that your implementation is not compatible with matplotlib. Can you have a look on the CI failure?

QRemy:
> Thanks @QRemy... It seems that your implementation is not compatible with matplotlib. Can you have a look on the CI failure?

Now it works for the latest version.

gammapy/maps/measure.py
```diff
@@ -0,0 +1,107 @@
...
+def containment_region(map_, fraction=0.393, apply_union=True):
+    """Find the iso-contours region corresponding to a given containment
+        for a map of integral quantities.
+
+    Parameters
+    ----------
+    map_ : `~gammapy.maps.WcsNDMap`
+        Map of integral quantities
+    fraction : float
+        Containment fraction
```
registerrier, line 18:
```suggestion
        Containment fraction. Default is 0.393.
```

gammapy/maps/measure.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+def containment_region(map_, fraction=0.393, apply_union=True):
+    """Find the iso-contours region corresponding to a given containment
+        for a map of integral quantities.
+
+    Parameters
+    ----------
+    map_ : `~gammapy.maps.WcsNDMap`
+        Map of integral quantities
```
registerrier, line 16:
```suggestion
        Map of integral quantities.
```

gammapy/maps/measure.py
```diff
@@ -0,0 +1,107 @@
...
+    fraction : float
+        Containment fraction
+    apply_union : bool
+        It True return a compound region otherwise return a list of polygon regions.
+        Default is True. Note that compound regions cannot be written in ds9 format but can always be saved with numpy.savez.
+
+    Returns
+    -------
+    regions : list of ~regions.PolygonSkyRegion` or `~regions.CompoundSkyRegion`
+        regions from iso-contours matching containment fraction
```
registerrier, line 26:
```suggestion
        regions from iso-contours matching containment fraction.
```

gammapy/maps/measure.py
```diff
@@ -0,0 +1,107 @@
...
+                regions_union = regions_union.union(region)
+            return regions_union
+        else:
+            return regions_pieces
+    else:
+        raise ValueError("No positive values in the map.")
+
+
+def containment_radius(map_, fraction=0.393, position=None):
+    """Compute containment radius from the center of a map with integral quantities
```
registerrier, line 76:
```suggestion
    """Compute containment radius from the center of a map with integral quantities.
```

gammapy/maps/measure.py
```diff
@@ -0,0 +1,107 @@
...
+
+    Parameters
+    ----------
+    fraction : float
+        Containment fraction
```
registerrier, line 81:
```suggestion
        Containment fraction.
```

gammapy/maps/measure.py
```diff
@@ -0,0 +1,107 @@
...
+    n_levels : int
+        Numbers of contours levels used to find the required containment radius.
+    position : `~astropy.coordinates.SkyCoord`
+        Position from where the containment is conputed.
```
registerrier, line 85:
```suggestion
        Position from where the containment is computed.
```

gammapy/maps/measure.py
```diff
@@ -0,0 +1,107 @@
...
+
+def containment_radius(map_, fraction=0.393, position=None):
+    """Compute containment radius from the center of a map with integral quantities
+
+    Parameters
+    ----------
+    fraction : float
+        Containment fraction
+    n_levels : int
+        Numbers of contours levels used to find the required containment radius.
```
registerrier, line 83:
```suggestion
```

registerrier (changes requested):
Thanks @QRemy . This looks good. I have left some inline comments.

The main one is that the functions will work only for flat geometries. They should raise an explicit exception. The docstring should also reflect this.

registerrier (approved):
Thanks @QRemy. This looks good to me. No further comments from my side.

PR #4664, opened 2023-07-17
Add a faster reprojection method  : reproject_by_image

QRemy:
This PR adds a method `reproject_by_slice` that reprojects each slice of a 3d map independently  in parallel. Even without the parallelisation the loop over the slices  is more efficient and consumes less memory than using directly  `reproject_to_geom`.

### DIFF
--- a/gammapy/maps/core.py
+++ b/gammapy/maps/core.py
@@ -1129,7 +1129,7 @@ class Map(abc.ABC):
 
         output_map = Map.from_geom(geom.to_cube(self.geom.axes))
         maps = parallel.run_multiprocessing(
-            self._reproject_slice,
+            self._reproject_image,
             zip(
                 self.iter_by_image(),
                 repeat(geom),
--- a/gammapy/maps/core.py
+++ b/gammapy/maps/core.py
@@ -1116,7 +1116,7 @@ class Map(abc.ABC):
             if the map is an integral quantity (e.g. counts) and false if
             the map is a differential quantity (e.g. intensity)
         precision_factor : int
-           Minimal factor between the bin size of the output map and the oversampled base map.
+             Minimal factor between the bin size of the output map and the oversampled base map.
            Used only for the oversampling method.
 
         Returns
--- a/gammapy/maps/core.py
+++ b/gammapy/maps/core.py
@@ -1105,6 +1105,8 @@ class Map(abc.ABC):
     ):
         """Reproject each image of a ND map to input 2d geometry.
 
+           For large maps this method is faster than `reproject_to_geom`.
+
         Parameters
         ----------
         geom : `~gammapy.maps.Geom`
--- a/gammapy/maps/core.py
+++ b/gammapy/maps/core.py
@@ -10,6 +10,7 @@ import numpy as np
 from astropy import units as u
 from astropy.io import fits
 import matplotlib.pyplot as plt
+import gammapy.utils.parallel as parallel
 from gammapy.utils.deprecation import deprecated
 from gammapy.utils.random import InverseCDFSampler, get_random_state
 from gammapy.utils.scripts import make_path
@@ -1140,10 +1141,11 @@ class Map(abc.ABC):
                 repeat(preserve_counts),
                 repeat(precision_factor),
             ),
-            backend=self.parallel_backend,
-            pool_kwargs=dict(processes=self.n_jobs),
+            # backend=self.parallel_backend,
+            # pool_kwargs=dict(processes=self.n_jobs),
             task_name="Reprojection",
         )
+        # TODO: uncomment after parallel support is added on maps
         return Map.from_stack(maps, axis=axis)
 
     @staticmethod
--- a/gammapy/maps/core.py
+++ b/gammapy/maps/core.py
@@ -5,6 +5,7 @@ import html
 import inspect
 import json
 from collections import OrderedDict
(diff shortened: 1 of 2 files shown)


gammapy/maps/core.py
```diff
@@ -1088,6 +1116,63 @@ def reproject_to_geom(self, geom, preserve_counts=False, precision_factor=10):
...
+        Parameters
+        ----------
+        geom : `~gammapy.maps.Geom`
+            Target slice geometry (2d)
+        preserve_counts : bool
+            Preserve the integral over each bin.  This should be true
+            if the map is an integral quantity (e.g. counts) and false if
+            the map is a differential quantity (e.g. intensity)
+        precision_factor : int
+           Minimal factor between the bin size of the output map and the oversampled base map.
```
bkhelifi, line 1136: This is not super clear for me....
You have a base map, then it is oversampled (there is a first bin factor). And then you add a second bin factor? 
I sure that I have not understand.

QRemy: The oversampling factor is not given by the user it is computed by dividing the `precision_factor` by the scale ratio of the output and input maps. Because the precision depends only of the factor between the oversampled map and the output map.

QRemy:
I removed the parallel options in maps from that PR but I still  use `run_multiprocessing` in  `reproject_by_image` so we can easily activate the parallel support later by uncommenting two lines.

gammapy/maps/core.py
```diff
@@ -1088,6 +1090,60 @@ def reproject_to_geom(self, geom, preserve_counts=False, precision_factor=10):
...
+        maps = parallel.run_multiprocessing(
+            self._reproject_image,
+            zip(
+                self.iter_by_image(),
+                repeat(geom),
+                repeat(preserve_counts),
+                repeat(precision_factor),
+            ),
+            # backend=self.parallel_backend,
+            # pool_kwargs=dict(processes=self.n_jobs),
```
QRemy, line 1133:
```suggestion
```

Astro-Kirsty:
What is the status with this PR?

gammapy/maps/core.py
```diff
@@ -1095,6 +1097,60 @@ def reproject_to_geom(self, geom, preserve_counts=False, precision_factor=10):
...
```
QRemy, line 1140:
```suggestion
```

gammapy/maps/core.py
```diff
@@ -1095,6 +1097,60 @@ def reproject_to_geom(self, geom, preserve_counts=False, precision_factor=10):
...
+            task_name="Reprojection",
+        )
+        # TODO: uncomment after parallel support is added on maps
```
QRemy, line 1143:
```suggestion
```

QRemy:
The PR is ready, I removed the comments related to the parallel options (after https://github.com/gammapy/gammapy/pull/4845 is merged the loop could run in parallel without any change so this PR can go in before).

gammapy/maps/core.py
```diff
@@ -1095,6 +1097,57 @@ def reproject_to_geom(self, geom, preserve_counts=False, precision_factor=10):
...
 
+    def reproject_by_image(
+        self,
+        geom,
+        preserve_counts=False,
+        precision_factor=10,
+    ):
+        """Reproject each image of a ND map to input 2d geometry.
+
+           For large maps this method is faster than `reproject_to_geom`.
```
registerrier, line 1108:
```suggestion
       For large maps this method is faster than `reproject_to_geom`.
```

gammapy/maps/core.py
```diff
@@ -1095,6 +1097,57 @@ def reproject_to_geom(self, geom, preserve_counts=False, precision_factor=10):
...
+        Parameters
+        ----------
+        geom : `~gammapy.maps.Geom`
+            Target slice geometry (2d)
+        preserve_counts : bool
+            Preserve the integral over each bin.  This should be true
+            if the map is an integral quantity (e.g. counts) and false if
+            the map is a differential quantity (e.g. intensity)
+        precision_factor : int
+           Minimal factor between the bin size of the output map and the oversampled base map.
```
registerrier, line 1119:
```suggestion
        precision_factor : int
             Minimal factor between the bin size of the output map and the oversampled base map.
```

gammapy/maps/core.py
registerrier: Maybe mention in the docstring that this can be sued with mutli-processing context manager ?

gammapy/maps/tests/test_core.py
```diff
@@ -717,6 +717,37 @@ def test_map_reproject_wcs_to_wcs_with_axes():
...
+    )
+    with pytest.raises(TypeError):
+        m.reproject_by_image(geom_wcs_wrong)
+
+    m_r = m.reproject_by_image(geom_wcs)
+    actual = m_r.get_by_coord(
+        {"lon": 0, "lat": 0, "energy_true": [1.0, 3.16227766, 10.0]}
+    )
+    assert_allclose(actual, [287.5, 1055.5, 1823.5], rtol=1e-3)
+
```
registerrier, line 749: Maybe add a TODO to remind adding a test with multiprocessing?

registerrier (dismissed):
Thanks @QRemy . This looks good to me. I have left a few inline comments.

registerrier:
Merging now.

PR #4665, opened 2023-07-17
`plot_distribution` in tutorials

MRegeard:
This PR is a follow up of #4408. 
It introduce the `plot_distribution` visualization function in the tutorials.

### DIFF
--- a/examples/tutorials/analysis-2d/ring_background.py
+++ b/examples/tutorials/analysis-2d/ring_background.py
@@ -59,6 +59,7 @@ from gammapy.analysis import Analysis, AnalysisConfig
 from gammapy.datasets import MapDatasetOnOff
 from gammapy.estimators import ExcessMapEstimator
 from gammapy.makers import RingBackgroundMaker
+from gammapy.visualization import plot_distribution
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


MRegeard:
The second commit on the 3D analysis tutorial change slightly the output, because it plot the distribution per bin.

QRemy (dismissed):
Thanks @MRegeard, looks good, no comments from my side.

MRegeard:
By adapting the two tutorials with this function, I get the feeling that it is heavy to use... I am not entirely sure that we should advertise it right now.

registerrier:
> By adapting the two tutorials with this function, I get the feeling that it is heavy to use... I am not entirely sure that we should advertise it right now.

Maybe part of the difficulty is with the `lambda` function that is a bit complex to explain. Maybe, one could define a specific case where `func="gauss"` which will internally create the 2 parameters gaussian function.

MRegeard:
> > By adapting the two tutorials with this function, I get the feeling that it is heavy to use... I am not entirely sure that we should advertise it right now.
> 
> Maybe part of the difficulty is with the `lambda` function that is a bit complex to explain. Maybe, one could define a specific case where `func="gauss"` which will internally create the 2 parameters gaussian function.

Yes I thought about that. A new function that would only deal with gaussian distribution. Something like `plot_gaussian_distribution.

bkhelifi:
And this new function could add a legend with the fitted parameters ;)

MRegeard:
> And this new function could add a legend with the fitted parameters ;)

Yes of course, in this case it would be really easy to implement !

bkhelifi (approved):
Thanks for the update!

MRegeard:
Pending bug fix in #4775 to be sure that everything is ok !

MRegeard:
It gives this for the two tutorial that I have updated :

Capture d’écran 2023-10-02 à 14 18 41
and 

Capture d’écran 2023-10-02 à 14 19 02

QRemy (approved):

registerrier (approved):
Thanks @MRegeard . This looks fine.

I think these examples show that we could find good defaults for these plots, or maybe provide specialized plot functions for specific types of maps that would alleviate the interaction with `plt.hist` and the `Axes`.

Maybe it could be worth opening a follow-up issue for this.

PR #4666, opened 2023-07-17
force color and linewidth in the plotting of the fit of `plot_distribution`

MRegeard:
This PR follows #4408. 
It is a small fix that force the color and the linewidth of the plotting of the fit to black and 2 respectively. 
I put the option that were used in the ring background tutorial (https://docs.gammapy.org/dev/tutorials/analysis-2d/ring_background.html).

### DIFF
--- a/gammapy/visualization/utils.py
+++ b/gammapy/visualization/utils.py
@@ -313,7 +313,7 @@ def plot_distribution(
             xmin, xmax = kwargs_hist.get("range", (np.min(d), np.max(d)))
             x = np.linspace(xmin, xmax, 1000)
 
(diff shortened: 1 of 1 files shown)


QRemy (approved):
These arguments could be given by a `kwargs_plot` but there is already a lot of kwargs so I'm fine with these hard coded values.

MRegeard:
Yes I think that another kwargs dictionary would be to much. 
Then Black seems ok, it will (probably) never be used as a color for a histogram !

Astro-Kirsty (approved):
Thanks @MRegeard. I'm happy with this.

Issue #3510, opened 2021-09-14
Inconsistency between gammapy.utils.coordinates.fov and gamma.makers.utils


maxnoe:
**Gammapy version**

Current master

**Bug description**

In `gammapy/utils/coordinates/fov`, `sky_to_fov` is defined like this:

```python
    # Create a frame that is centered on the pointing position
    center = SkyCoord(lon_pnt, lat_pnt)
    fov_frame = SkyOffsetFrame(origin=center)

    # Define coordinate to be transformed.
    target_sky = SkyCoord(lon, lat)

    # Transform into FoV-system
    target_fov = target_sky.transform_to(fov_frame)

    # Switch sign of longitude angle since this axis is
    # reversed in our definition of the FoV-system
    return -target_fov.lon, target_fov.lat

```

Observe the `-` in front of `target_fov.lon`.

The only place this function is used is in `gammapy/makers/utils.py`, where there is an if condition and the same transformation is basically done "by hand" again: 

https://github.com/gammapy/gammapy/blob/a3f6b0945b15cfea1d5fc3f711ef58929161d49a/gammapy/makers/utils.py#L60-L89

But this time, no `-` in front of `lon`.

This seems inconsistent. Is this a bug? Or intentional? Then there should proabbly be some kind of comment why it is negated in the one function but not in the "by hand" approach, which essentially duplicates the same code again.

adonath:
Thanks @maxnoe for noticing!  The [code here](https://github.com/gammapy/gammapy/blob/master/gammapy/makers/map.py#L151) uses the `FOVALIGN` keyword in the background meta data to choose the coordinate transformation. The keyword is documented here: https://gamma-astro-data-formats.readthedocs.io/en/latest/general/coordinates.html#field-of-view. There it says "FOV LON should increase with decreasing OTHER LON", so I guess in both cases the minus sign is required? 

I think the ideal solution would be to just get rid of the "pseudo FoV" coordinates. Right now I can't even remember why it has been there at all. I guess because it comes from a time, where background model where radially symmetric anyway? And because both have been used in the past? Maybe you remember @lmohrmann or @registerrier?

In any case it would be good to check whether the definitions are consistent between Gammapy and pyirf / ctapipe.

maxnoe:
In ctapipe, we right now just use the SkyOffsetFrame for `TelescopeCoordinate` (origin in AltAz). We don't have a frame aligned with `ICRS`.

[...]

PR #4667, opened 2023-07-18
Use sky_to_fov() also for RADEC FoVAlignment in make_map_background_irf

SimoneMender:
Suggestion to fix issue #3510

### DIFF
--- a/gammapy/irf/core.py
+++ b/gammapy/irf/core.py
@@ -33,9 +33,8 @@ class FoVAlignment(str, Enum):
 
     ALTAZ = "ALTAZ"
     RADEC = "RADEC"
-    REVERSE_LON_RADEC = (
-        "REVERSE_LON_RADEC"  # used for backward compatibility of old HESS data
-    )
+    # used for backward compatibility of old HESS data
+    REVERSE_LON_RADEC = "REVERSE_LON_RADEC"
 
 
 class IRF(metaclass=abc.ABCMeta):
--- a/gammapy/irf/core.py
+++ b/gammapy/irf/core.py
@@ -33,6 +33,9 @@ class FoVAlignment(str, Enum):
 
     ALTAZ = "ALTAZ"
     RADEC = "RADEC"
+    REVERSE_LON_RADEC = (
+        "REVERSE_LON_RADEC"  # used for backward compatibility of old HESS data
+    )
 
 
 class IRF(metaclass=abc.ABCMeta):
--- a/gammapy/makers/utils.py
+++ b/gammapy/makers/utils.py
@@ -14,6 +14,8 @@ from gammapy.stats import WStatCountsStatistic
 from gammapy.utils.coordinates import sky_to_fov
 from gammapy.utils.regions import compound_region_to_regions
 
+REVERSE_FOV_LON = False
+
 __all__ = [
     "make_counts_rad_max",
     "make_edisp_kernel_map",
@@ -89,7 +91,8 @@ def _get_fov_coords(pointing, irf, geom, use_region_center=True, obstime=None):
                 pointing_icrs.icrs.ra,
                 pointing_icrs.icrs.dec,
             )
-            fov_lon = -fov_lon
+            if REVERSE_FOV_LON:
+                fov_lon = -fov_lon
         else:
             raise ValueError(
                 f"Unsupported background coordinate system: {irf.fov_alignment!r}"
--- a/gammapy/makers/utils.py
+++ b/gammapy/makers/utils.py
@@ -84,7 +84,10 @@ def _get_fov_coords(pointing, irf, geom, use_region_center=True, obstime=None):
             )
         elif irf.fov_alignment == FoVAlignment.RADEC:
             fov_lon, fov_lat = sky_to_fov(
-                sky_coord.ra, sky_coord.dec, pointing_icrs.ra, pointing_icrs.dec
+                sky_coord.icrs.ra,
+                sky_coord.icrs.dec,
+                pointing_icrs.icrs.ra,
+                pointing_icrs.icrs.dec,
             )
         else:
             raise ValueError(
--- a/gammapy/makers/utils.py
+++ b/gammapy/makers/utils.py
@@ -89,6 +89,7 @@ def _get_fov_coords(pointing, irf, geom, use_region_center=True, obstime=None):
                 pointing_icrs.icrs.ra,
                 pointing_icrs.icrs.dec,
             )
+            fov_lon = -fov_lon
         else:
             raise ValueError(
                 f"Unsupported background coordinate system: {irf.fov_alignment!r}"
--- a/gammapy/makers/utils.py
+++ b/gammapy/makers/utils.py
@@ -14,8 +14,6 @@ from gammapy.stats import WStatCountsStatistic
 from gammapy.utils.coordinates import sky_to_fov
 from gammapy.utils.regions import compound_region_to_regions
 
-REVERSE_FOV_LON = False
-
 __all__ = [
     "make_counts_rad_max",
     "make_edisp_kernel_map",
@@ -84,14 +82,14 @@ def _get_fov_coords(pointing, irf, geom, use_region_center=True, obstime=None):
             fov_lon, fov_lat = sky_to_fov(
                 altaz_coord.az, altaz_coord.alt, pointing_altaz.az, pointing_altaz.alt
             )
-        elif irf.fov_alignment == FoVAlignment.RADEC:
+        elif irf.fov_alignment in [FoVAlignment.RADEC, FoVAlignment.REVERSE_LON_RADEC]:
             fov_lon, fov_lat = sky_to_fov(
                 sky_coord.icrs.ra,
                 sky_coord.icrs.dec,
                 pointing_icrs.icrs.ra,
                 pointing_icrs.icrs.dec,
             )
-            if REVERSE_FOV_LON:
+            if irf.fov_alignment == FoVAlignment.REVERSE_LON_RADEC:
                 fov_lon = -fov_lon
         else:
             raise ValueError(
--- a/gammapy/makers/utils.py
+++ b/gammapy/makers/utils.py
@@ -3,7 +3,7 @@ import logging
 import warnings
 import numpy as np
 import astropy.units as u
-from astropy.coordinates import Angle, SkyOffsetFrame
+from astropy.coordinates import Angle
 from astropy.table import Table
 from gammapy.data import FixedPointingInfo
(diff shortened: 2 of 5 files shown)


bkhelifi:
Hello @SimoneMender ,
Thank you for this PR. @maxnoe, is the `SkyOffsetFrame(origin=pointing_icrs)` compatible with the CTAO official frame in which the IRF can be defined? The link with pyrirf is obvious here and a coordination is needed.

maxnoe:
Unfortunayley, there are no official CTAO celestial coordinate definitions yet

maxnoe:
Unfortunayley, there are no official CTAO celestial coordinate definitions yet

bkhelifi:
It was my fear... And do you think that, at least temporary in Gammapy&pyirf, this definition from astropy could be used? At least for testing purposes....

SimoneMender:
The issue is that according to the GADF definition (https://gamma-astro-data-formats.readthedocs.io/en/latest/general/coordinates.html#coords-fov),  "FOV LON should increase with decreasing OTHER LON". As this is the coordinate system in which a background model has to be stored, it is mandatory to switch the sign of longitude angle when creating a background map. This is exactly covered by the `sky_to_fov()` function.

maxnoe:
I think that CTA(O) definitions are not relevant here.

As of now, Gammapy only supports GADF as input format. So what applies are the definitions of the field of view coordinate system of GADF.

And GADF defines the FOV longitude to have the opposite sign to the system it is aligned with, for both alignments (ALTAZ / RADEC).

We should probably  implement this using astropy coordinate frames like `GADFFoVAltAz` and `GADFFoVICRS`.

SimoneMender:
Isn't the parameter `sky_coord` already following the definition of the right coordinate system (in the case of RADEC FoVAlignment), as it is created from the reference geometry `geom`, which is a `~gammapy.maps.WcsGeom` object?
https://github.com/SimoneMender/gammapy/blob/474ad6c23d6e152b60298a489ada5ca6e3a1d331/gammapy/makers/utils.py#L191C21-L191C30

gammapy/makers/utils.py
```diff
@@ -217,11 +217,9 @@ def make_map_background_irf(
                 altaz_coord.az, altaz_coord.alt, pointing_altaz.az, pointing_altaz.alt
             )
         elif bkg.fov_alignment == FoVAlignment.RADEC:
-            # Create OffsetFrame
-            frame = SkyOffsetFrame(origin=pointing_icrs)
-            pseudo_fov_coord = sky_coord.transform_to(frame)
-            fov_lon = pseudo_fov_coord.lon
-            fov_lat = pseudo_fov_coord.lat
+            fov_lon, fov_lat = sky_to_fov(
+                sky_coord.ra, sky_coord.dec, pointing_icrs.ra, pointing_icrs.dec
```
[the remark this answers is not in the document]
registerrier: Also is it guaranteed that `sky_coord` is in RADEC system?

registerrier (changes requested):
Thanks @SimoneMender . 
As far as I know, `sky_to_fov` does the same thing as what was done here (except for the sign of fov_lon). But it does not check the input systems of the coordinates. 

The issue lies probably in the `sky_to_fov` implementation, taking ra and dec instead of a full SkyCoord.

bkhelifi:
@SimoneMender can you update your code from the main, such that the CI tests can pass successfully?

gammapy/makers/utils.py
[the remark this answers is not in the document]
QRemy: done

gammapy/makers/utils.py
```diff
@@ -82,11 +82,13 @@ def _get_fov_coords(pointing, irf, geom, use_region_center=True, obstime=None):
...
-            pseudo_fov_coord = sky_coord.transform_to(frame)
-            fov_lon = pseudo_fov_coord.lon
-            fov_lat = pseudo_fov_coord.lat
+            fov_lon, fov_lat = sky_to_fov(
+                sky_coord.icrs.ra,
+                sky_coord.icrs.dec,
+                pointing_icrs.icrs.ra,
+                pointing_icrs.icrs.dec,
+            )
+            fov_lon = -fov_lon
```
QRemy, line 91: so what's the correct sign ?

maxnoe: GADF specs are here: https://gamma-astro-data-formats.readthedocs.io/en/v0.3/general/coordinates.html#field-of-view

They say:

> FOV LON should increase with decreasing OTHER LON

maxnoe: But that's true for both `FOV_ALIGNMENTS`, so `sky_to_fov` should already produce the correct sign

maxnoe: sky_to_fov already has a `-` for fov_lon, so here should be None, which is the actual fix

lmohrmann: Would still be important that the coordinates come out the same way as they did before (which is still what we expect, right?)

maxnoe: No, this was opened to make gammapy match the GADF definition, which wasn't the case before

QRemy: Does the HESS background models match the GADF definition ?
If not we need a special handling to still be able to use them.

maxnoe:
@lmohrmann Thanks a lot for this very thorough check!

So indeed, it is as we assumed: Gammapy and H.E.S.S. deviate from the GADF definition for the FoV RaDec alignment.

I am not sure as well on how to proceed. Changing the GADF definition so that it matches current Gammapy and H.E.S.S. behavior is a bit weird, since I think the sign flipping of the FoV coordinate was introduced especially for Ra/Dec, not for AltAz. So having the definition be exactly the opposite... is really weird.


So I see a couple of possibilities:

* We bring Gammapy in-line with the GADF definition. This should probably happen in a major release, i.e. Gammapy 2.0
* We should think about how to best deal with existing files. Do HESS DL3 files have proper metadata set e.g. for `CREATOR`, so that we could do something like:

```
class ChangedBkgCoordinatesWarning(GammapyWarning):
    pass


if the_hess_software in header["CREATOR"] and get_version(header["CREATOR"]) < THE_HESS_VERSION_WHERE_IT_WAS_FIXED:
    warnings.warn("Reading legacy H.E.S.S. data file, assuming flipped definition of field of view coordinates")
    header['FOV_ALIGN"] = FovAlignment.LEGACY_RADEC
```

maxnoe:
Checking the files of the hess-dr1, there is no version number in those. Just `SASH FITS::EventListWriter`.

However, I see `HDUVERS = '0.2'`

@lmohrmann is there HESS DL3 data that specifies `HDUVERS = '0.3'`, if not, maybe we could use that as signed for "HESS data after coordinates were fixed"?

lmohrmann:
Yes, I think we will be able to use the `CREATOR` and `HDUVERS` keywords to be able to distinguish between existing and future background model files :+1:

lmohrmann:
Ah, and on this:

>  Changing the GADF definition so that it matches current Gammapy and H.E.S.S. behavior is a bit weird, since I think the sign flipping of the FoV coordinate was introduced especially for Ra/Dec, not for AltAz. So having the definition be exactly the opposite... is really weird.

I didn't necessarily mean that we should bring GADF in line with current Gammapy, but rather that this might be a good point to reflect upon whether or not the definition as it is makes sense or should be changed in any way. But it's also not super important I think – as long as we make sure that all relevant software follow the specs 😅

gammapy/irf/core.py
```diff
@@ -33,6 +33,9 @@ class FoVAlignment(str, Enum):
 
     ALTAZ = "ALTAZ"
     RADEC = "RADEC"
+    REVERSE_LON_RADEC = (
+        "REVERSE_LON_RADEC"  # used for backward compatibility of old HESS data
+    )
```
maxnoe, line 38:
```suggestion
    # used for backward compatibility of old HESS data
    REVERSE_LON_RADEC = "REVERSE_LON_RADEC"
```

QRemy:
I implemented @maxnoe suggestion. 
However the `"CREATOR"`  and `"HDUVERS"` keyword are on the event header while `"FOVALIGN"` is on the background header, so it is not straightforward to change the `fov_alignment` and the bkg object creation. Instead  I did the modification within the bkg property of the observation using the meta infos.

maxnoe:
> However the "CREATOR" and "HDUVERS" keyword are on the event header while "FOVALIGN" is on the background header

At least the `HDUVERS` should also be present in the `BACKGROUND` header.

registerrier (commented):
Thanks @QRemy @lmohrmann @maxnoe . 

I think we should avoid solutions where different behavior are applied at the data handling and makers level.  I would rather put the patch at the irf IO level e.g. reverting fov_lon axis in `Background.read`. In general, future versions of data format might have different conventions. 
We have to clarify the internal gammapy convention and stick to it. The IRF objects we use should follow these conventions independently of their initial format.

bkhelifi (changes requested):
Beyond the implementation, respecting GADF standards is wise. One should maybe add numerical tests to associated the standards to values

QRemy:
> I think we should avoid solutions where different behavior are applied at the data handling and makers level. I would rather put the patch at the irf IO level e.g. reverting fov_lon axis in `Background.read`. In general, future versions of data format might have different conventions. We have to clarify the internal gammapy convention and stick to it. The IRF objects we use should follow these conventions independently of their initial format.

It cannot be done in in `Background.read` because the necessary meta key are not on the background header.

QRemy:
I'll propose to go with this fix for version 1.3 and if someone else want to find a better solution that works also at `Background.read` level we add it later.

maxnoe:
@bkhelifi please read the discussion above, especially the post by @lmohrmann 

This PR is specifically to bring Gammapy in line with the GADF spec.

bkhelifi:
> @bkhelifi please read the discussion above, especially the post by @lmohrmann
> 
> This PR is specifically to bring Gammapy in line with the GADF spec.

@registerrier will be then reassured;)

QRemy (approved):
I see no objection so far  then I'll merge it now.

maxnoe:
@QRemy Sorry for only writing now that this has been merged, but the notification triggered me to have a look again.

If I see this correctly, here this is only solved on the `Observation`, which means everything going through `DataStore` or `Observation.read`, correct?

I think we should add the same logic to `read_irf_dict_from_file`, right?

QRemy:
by `read_irf_dict_from_file ` do you mean  `load_irf_dict_from_file` ?
The problem is  that the necessary meta keys to identify the problematic version are not on the background header but on the events header, that's why it can only by added at the Observation level.

PR #4669, opened 2023-07-18
Add the progress bar for the DataStore

bkhelifi:
**Description**
This pull request add the progress bar in the function `get_observations`.

**Dear reviewer**
I am wondering if, in general, one should mention in the docstring that a progress bar is possible for a function...

### DIFF
--- a/gammapy/data/obs_table.py
+++ b/gammapy/data/obs_table.py
@@ -74,7 +74,9 @@ class ObservationTable(Table):
 
     def summary(self):
         """Summary info string (str)."""
-        obs_name = self.meta.get("OBSERVATORY_NAME", "N/A")
+        obs_name = self.meta.get(
+            "OBSERVATORY_NAME", "N/A"
+        )  # This is not GADF compliant
         return (
             f"Observation table:\n"
             f"Observatory name: {obs_name!r}\n"
--- a/gammapy/data/obs_table.py
+++ b/gammapy/data/obs_table.py
@@ -77,6 +77,9 @@ class ObservationTable(Table):
         obs_name = self.meta.get(
             "OBSERVATORY_NAME", "N/A"
         )  # This is not GADF compliant
+        if "N/A" in obs_name:
+            obs_name = self.meta.get("OBSERVER", "N/A")
+
         return (
             f"Observation table:\n"
             f"Observatory name: {obs_name!r}\n"
--- a/gammapy/data/obs_table.py
+++ b/gammapy/data/obs_table.py
@@ -78,7 +78,9 @@ class ObservationTable(Table):
             "OBSERVATORY_NAME", "N/A"
         )  # This is not GADF compliant
         if "N/A" in obs_name:
-            obs_name = self.meta.get("OBSERVATORY_NAME", self.meta.get("OBSERVER", "N/A"))
+            obs_name = self.meta.get(
+                "OBSERVATORY_NAME", self.meta.get("OBSERVER", "N/A")
+            )
 
         return (
             f"Observation table:\n"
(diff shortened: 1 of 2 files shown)


gammapy/data/obs_table.py
```diff
@@ -74,7 +74,12 @@ def select_obs_id(self, obs_id):
 
     def summary(self):
         """Summary info string (str)."""
-        obs_name = self.meta.get("OBSERVATORY_NAME", "N/A")
+        obs_name = self.meta.get(
+            "OBSERVATORY_NAME", "N/A"
+        )  # This is not GADF compliant
+        if "N/A" in obs_name:
+            obs_name = self.meta.get("OBSERVER", "N/A")
```
AtreyeeS, line 81:
```suggestion
            obs_name = self.meta.get("OBSERVATORY_NAME", self.meta.get("OBSERVER", "N/A"))
```

AtreyeeS (approved):
Thanks @bkhelifi !

gammapy/data/obs_table.py
```diff
@@ -74,7 +74,12 @@ def select_obs_id(self, obs_id):
 
     def summary(self):
         """Summary info string (str)."""
-        obs_name = self.meta.get("OBSERVATORY_NAME", "N/A")
+        obs_name = self.meta.get(
```
AtreyeeS, line 77: Lines 77-80 are not necessary :-)

Astro-Kirsty (approved):
Look good from my side @bkhelifi

gammapy/data/obs_table.py
```diff
@@ -74,7 +74,12 @@ def select_obs_id(self, obs_id):
 
     def summary(self):
         """Summary info string (str)."""
-        obs_name = self.meta.get("OBSERVATORY_NAME", "N/A")
```
registerrier, line 77: I tend to agree with @AtreyeeS . If there is no keyword specification for the observatory name in the specifications, we shouldn't try to guess one. 
I would competely remove this part.

registerrier (commented):
Thanks @bkhelifi . See small suggestion.

QRemy (commented):
I don't think this is useful because `get_observations` does not read the files as we use lazy loading so it should be very fast and the display time of the progress bar probably add too much overhead.

bkhelifi:
The lazy loading is set to False since its introduction...
On my computer, this function takes more than 4 min, which I realised by introducing this optional feature.
It costs nothing and it is useful;)

gammapy/data/obs_table.py
bkhelifi: Agree, even if the gammapy team and friends are leading the data formatting. It is not 'others' that will make specifications.

registerrier:
> The lazy loading is set to False since its introduction... On my computer, this function takes more than 4 min, which I realised by introducing this optional feature. It costs nothing and it is useful;)

It is the caching of the loaded content that is set to `False` in the lazy loading mechanism. So we re-read from disk each time we access one of the HDU. But when you extract the observation from the datastore, its content is not yet loaded.  

Technically the operation `get_observations()` is long only when you extract a very large number of observations from the `DataStore` (typically >10000).

QRemy:
> So we re-read from disk each time we access one of the HDU

Does that mean we also re-decompress the whole file each time we access one of the HDU ? (if yes that probably takes most of the time)

morcuended:
> Possibly I am mistaken, but did you not face similarly large times to load observations for the LST-1 dataset @morcuended ?

Yes, I think the problem was the unzipping of the files. We reduced the run time from the order of 30 mins to just several minutes when using already unzipped files.

QRemy (approved):
Thanks, no further comments.
I corrected the formatting and will merge now.

Issue #4651, opened 2023-07-11
Add option in `SpatialModels` to plot the error of the model


Astro-Kirsty:
It would be useful to be able to plot the error of the `SpatialModel`. 
Currently you can use `to_region` to plot the `SpatialModel` region, and `plot_error` to plot the position error, however, it would also be nice to be able to plot the shaded error band of your `SpatialModel` too. 

**Describe the solution you'd like**
Change naming of the current `plot_error` so that it is more clear.
Then add function to utilise the `AnnulusSkyRegions` to plot the shaded error of the `SpatialModel` of choice.

**Additional context**
An example for the GaussianSpatialModel here showing the shaded error:
![image](https://github.com/gammapy/gammapy/assets/60767360/0fef6028-f957-41f7-9fe0-463e4bac6a24)

Also, I notice that for the current `plot_error` function, it is plotting the position as an 'x' and then placing a region (circle-like) that shows the error of the position. I wonder if it is also useful to plot the actual error of position, i.e. as a '+' where the length and width of the '+' are the errors.

AtreyeeS:
A shading around the Gaussian looks nice @Astro-Kirsty ! 
Since these plots are often added on top of some map (eg: a significance map or a map at some other wavelengths), I don't know how visible a shaded colour might be. But it would be nice to have, and then users can try to adapt the colours accordingly.
Adding a + with relevant sizes can also be useful indeed.

We can adapt the names accordingly `plot_error_position` and `plot_error_extension` maybe

Astro-Kirsty:
Yes, that sounds good. 

The SpatialModels have an option `to_region`

PR #4670, opened 2023-07-19
Add function to plot error of SpatialModels

Astro-Kirsty:
**Description**
This PR aims to resolve [#4651](https://github.com/gammapy/gammapy/issues/4651). 
I have changed the `plot_error` function to be `plot_position_error`, and added a new `plot_error` function which allows the user to plot either/both the position and extension errors for each `SpatialModel`. 
For any `SpatialModel` which has no extension error define it will just default to the base class and `pass`.

I adapted the test, and added a new one for this PR.

**Dear reviewer**
It might be nice to show the use of this somewhere too

### DIFF
--- a/gammapy/modeling/models/spatial.py
+++ b/gammapy/modeling/models/spatial.py
@@ -13,6 +13,7 @@ from astropy.utils import lazyproperty
 from regions import (
     CircleAnnulusSkyRegion,
     CircleSkyRegion,
+    EllipseAnnulusSkyRegion,
     EllipseSkyRegion,
     PointSkyRegion,
     RectangleSkyRegion,
@@ -353,7 +354,7 @@ class SpatialModel(ModelBase):
             raise TypeError("Use .plot() for 2D Maps")
         m.plot_interactive(ax=ax, **kwargs)
 
-    def plot_error(self, ax=None, **kwargs):
+    def plot_position_error(self, ax=None, **kwargs):
         """Plot position error
 
         Parameters
@@ -386,6 +387,40 @@ class SpatialModel(ModelBase):
             ax.add_artist(artist)
 
         return ax
+
+    def plot_extension_error(self, ax=None):
+        pass
+
+    def plot_error(self, ax, selection_optional="position", **kwargs):
+        """Plot the errors of the spatial model.
+
+        Parameters
+        ----------
+        ax : `~matplotlib.axes.Axes`, optional
+            Axis
+        selection_optional: list of str
+            Which errors to plot.
+            Available options are:
+                * "all": all the optional steps are plotted
+                * "position": plot the position error of the spatial model.
+                * "extension": plot the extension error of the spatial model.
+        **kwargs : dict
+            Keyword arguments passed to `~gammapy.maps.WcsMap.plot()`
+
+        Returns
+        -------
+        ax : `~matplotlib.axes.Axes`, optional
+            Axis
+        """
+        if "all" in selection_optional:
+            self.plot_position_error(ax, **kwargs)
+            self.plot_extension_error(ax, **kwargs)
+
+        if "position" in selection_optional:
+            self.plot_position_error(ax, **kwargs)
+
+        if "extension" in selection_optional:
+            self.plot_extension_error(ax, **kwargs)
 
     def plot_grid(self, geom=None, **kwargs):
         """Plot spatial model energy slices in a grid.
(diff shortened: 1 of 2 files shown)


gammapy/modeling/models/spatial.py
```diff
@@ -352,7 +353,7 @@ def plot_interactive(self, ax=None, geom=None, **kwargs):
             raise TypeError("Use .plot() for 2D Maps")
         m.plot_interactive(ax=ax, **kwargs)
 
-    def plot_error(self, ax=None, **kwargs):
```
bkhelifi, line 355: Maybe a deprecation warning is needed here, as the API changes

bkhelifi:
We have an well-known technical issue that the docstring of the child classes does not contain the docstring of the parent class. So, I fear that one needs to copy the docstring in the child classes... To be discussed

gammapy/modeling/models/spatial.py
```diff
@@ -386,6 +387,40 @@ def plot_error(self, ax=None, **kwargs):
 
         return ax
 
+    def plot_extension_error(self, ax=None):
+        pass
+
+    def plot_error(self, ax, selection_optional="position", **kwargs):
```
AtreyeeS, line 393: `selection_optional` is more suited when there is a set of default options which are executed, like in the `Estimators`. Maybe `which`?

gammapy/modeling/models/spatial.py
```diff
@@ -612,6 +647,55 @@ def evaluation_region(self):
...
+        r"""Plot model error at a given number of :math:`\sigma`.
+
+        Parameters
+        ----------
+        ax : `~matplotlib.axes.Axes`, optional
+                Axis
+        x_sigma : float
+            Number of :math:`\sigma
+            Default is :math:`1.5\sigma` which corresponds to about 68%
+            containment for a 2D symmetric Gaussian.
```
AtreyeeS, line 660: This is just the error on the extension that is plotted, no? Plotting the 1-sigma error would be more standard I think

AtreyeeS (commented):
Thanks @Astro-Kirsty ! Looking at the code duplication, I think it would be nice to have Gammapy understand which are the extension parameters.
We discussed having parameter types in the past. The simplest would be to have an `is_extension` attribute, akin to `is_norm` for the spectral models. Opinions @bkhelifi @registerrier @adonath @QRemy ?

Otherwise, I agree with @bkhelifi that the individual functions would need to be documented.

Astro-Kirsty:
The following SpatialModels have a `plot_extension_error` implemented:
- `GaussianSpatialModel`, `DiskSpatialModel`, `GeneralizedGaussianSpatialModel`

The following SpatialModels will not have a `plot_extension_error`:
- `PointSpatialModel` (position only, no extension), 
- `ConstantSpatialModel`, `PiecewiseNormSpatialModel`, `TemplateSpatialModel` (each of these are a map)
_It could be good to include some warning message to say that it is not possible to plot the extension in these cases?_

Finally, for the two shell models: `ShellSpatialModel`, `Shell2SpatialModel`. I don't see a purpose for wanting to plot the error band. It can be implemented, but I think it is an odd thing to want to plot. I propose we also do not include the `plot_extension_error` for these either. 

image


**Open to suggestions and discussion around this!**

QRemy:
For the shell the errors on Rout and Rin could be in different colors.

AtreyeeS:
Maybe breaking this function into smaller parts will help with the duplication

- Create a `plot_region` in `gammapy.visualisation`  which plots a given region
- Create a `extension_error_region(extension_parameter)` in the `SpatialModel` base (or this can also reside in visualisations, in which case the model needs to be passed as well) which creates the Annular Region for using the said extension
- For each model, call the above functions by specifying the extension parameter.

gammapy/modeling/models/spatial.py
```diff
@@ -612,6 +654,55 @@ def evaluation_region(self):
         """Evaluation region consistent with evaluation radius"""
         return self.to_region(x_sigma=5)
 
+    def plot_extension_error(self, ax=None, x_sigma=1.5, **kwargs):
```
registerrier, line 657: Most of the plotting code is duplicated, only the region definition differs from one model to the other.

This could be simplified by putting most of the plotting code in the base `SpatialModel` and create e.g. hidden `GaussianSpatialModel._region_error` that will return the error ellipse (or maybe just its parameters).

registerrier (commented):
Thanks @Astro-Kirsty .  Those plots are useful and nice. A possible question is whether plotting the error on sigma conveys the better view on the source size.

My main comment concerns the code duplication and I would propose to move all the code for plotting to the base class and leave just the region definition on the child ones.

Astro-Kirsty:
Thanks @QRemy, @AtreyeeS and @registerrier. 
I guess we need to agree here on the correct placement of the code. Whether it should be kept in the `SpatialModels` or partly moved to `gammapy.visualisation`

One of my beginning ideas was to have a `to_region_error` function which was similar to the `to_region` functions for each SpatialModel. This sounds similar to what you are suggesting Regis. 

Also, I do not see a purpose for wanting to plot the errors for the shell models. Can either of you think of a use case?

Astro-Kirsty:
> For the shell the errors on Rout and Rin could be in different colors.

Happy to implement it this, but with the current implementation it would require something like:

```
kwargs = {}
prop_cycle = kwargs.pop("prop_cycle", plt.rcParams["axes.prop_cycle"])

if len(regions)>1:
    for region, props in zip(regions, prop_cycle()):
        artist_inner = region.to_pixel(ax.wcs).as_artist(color=props.pop("color"), fill=True, alpha=0.2)
        ax.add_artist(artist_inner)
```

Astro-Kirsty:
@QRemy and @AtreyeeS , for now the purpose of this PR is to plot the extension error for different spatial models. As the shell spatial models do not specifically have an extension, and would require a proper expression for error propagation (which is beyond the scope of this function) I will leave the PR as it is. However, I will create an issue for the potential to plot these more complex shell models in the future.

PR #4675, opened 2023-07-20
Add parallel options support in Map classes 

QRemy:
Add parallel options support to Map classes.
Required for #4664
In future this could also be useful to convolution with multiple kernels of a PSFMap in parallel.

### DIFF
--- a/gammapy/maps/core.py
+++ b/gammapy/maps/core.py
@@ -8,6 +8,7 @@ import numpy as np
 from astropy import units as u
 from astropy.io import fits
 import matplotlib.pyplot as plt
+from gammapy.utils import parallel
 from gammapy.utils.deprecation import deprecated
 from gammapy.utils.random import InverseCDFSampler, get_random_state
 from gammapy.utils.scripts import make_path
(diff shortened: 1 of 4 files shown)


bkhelifi (approved):
Thank you very much. This is a very useful feature

adonath:
I agree with @registerrier that the Map API should be mostly agnostic from a distributed computing, except if we decide this to a design based on e.g. `dask`. If we would like to introduce the parallel support now  we could use a context manager to keep the API separate:

```python3

from gammapy.maps.utils import compute_parallel
from gammapy.maps import Map

m = Map.create()

with compute_parallel(m, backend="ray", axis="energy") as mp:
    m_new = mp.smooth(5)

# OR

with compute_parallel(m, backend="ray", axis="energy"):
    m_new = mp.smooth(5)

# with a decorator on `Map.smooth` and all other methods which should support parallelism

# OR simpler

from gammapy.maps.utils import compute_parallel

m = Map.create()

m_new = compute_parallel(m.smooth, kwargs_func={}, backend="ray", axis="energy")

```

QRemy:
The context manager approach is implemented in #4845. Closing this now.

PR #4678, opened 2023-07-21
Fixing the test on `plot_distribution`

MRegeard:
It appears that the test on `plot_distribution` was failing in some case. Since it was not failing locally, the only thing I could think of was the random draw. Therefore I use a seed for the random draw.

### DIFF
--- a/gammapy/visualization/tests/test_utils.py
+++ b/gammapy/visualization/tests/test_utils.py
@@ -6,6 +6,7 @@ import astropy.units as u
 from astropy.table import Table
 import matplotlib.pyplot as plt
 from gammapy.maps import Map, MapAxis, WcsNDMap
+from gammapy.utils.random import get_random_state
 from gammapy.utils.testing import mpl_plot_check, requires_data
 from gammapy.visualization import (
     plot_contour_line,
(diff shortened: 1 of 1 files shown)


bkhelifi (approved):
Thanks @MRegeard ... I hope that the CI will work whatever the OS

AtreyeeS (approved):
Thanks @MRegeard ! I don't know if `nfev` is a good parameter to assert on. It reflects the number of evaluations of the objective function, so it might vary depending on the machine or the OS. An assert on the `param` would be more physical.

The other test fail seems unrelated. The caplog is picking some other warning before the gammapy warning I think.

MRegeard:
> Thanks @MRegeard ! I don't know if `nfev` is a good parameter to assert on. It reflects the number of evaluations of the objective function, so it might vary depending on the machine or the OS. An assert on the `param` would be more physical.
> 
> The other test fail seems unrelated. The caplog is picking some other warning before the gammapy warning I think.

It's true that now that you say it, it si probably not a very good idea to test nfev.
For the test on `param`, at the beginning I was test the parameter and the covariance matrix, but they were not the same depending on the OS. However, now that I am using a seed for the random drawing, I guess the parameter and covariance matrix may be more stable.

AtreyeeS:
> For the test on `param`, at the beginning I was test the parameter and the covariance matrix, but they were not the same depending on the OS. 

The `param` should be `[0,1]` within a given tolerance (1% on the mean and ~2% on the sigma I think). Just put a relatively high `rtol` around 5% in the assert

PR #4681, opened 2023-07-21
Fix MapEvaluator for regions

QRemy:
Fix a fails in MapEvaluator with regions because of update_spatial_oversampling_factor


AtreyeeS (approved):
Thanks @QRemy ! No comments from my side

Astro-Kirsty (approved):
As per discussions in dev call, no comments on my side either

QRemy:
CI fail is unrelated merging now

Issue #4691, opened 2023-07-24
Error in logging capture and ray


registerrier:
Several tests using caplog to check logging warning messages are failing. See e.g.:
https://github.com/gammapy/gammapy/actions/runs/5621343729/job/15235574153?pr=4678

I can reproduce the issue locally, and even have a bit more tests failing after changing the multiprocessing import logic . See https://github.com/registerrier/gammapy/actions/runs/5647077306/job/15296415904

Uninstalling ray completely removes the issue. 

We need to understand better how logging and logging capture are impacted by the multiprocessing in general and ray in particular.

QRemy:
That's probably due to the log deduplication
https://docs.ray.io/en/master/ray-observability/user-guides/configure-logging.html#log-deduplication

registerrier:
So far the issue with CI has gone. Closing the issue.

PR #4693, opened 2023-07-25
Introduce a no-ray tox environment 

registerrier:
**Description**

This pull request introduces a set of dependencies without `ray` in the setup.cfg, as well as `alldeps_nroay` tox environment to allow for testing without ray installed. This is a temporary measure to limit CI fails due to inconsistencies between ray and the logging system. See issue #4691 .
 
The CI matrix workflow has been adapted accordingly. I have left ray on ubuntu with python 3.10 and coverage.
Opinion @QRemy , @adonath ?

**Dear reviewer**

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,11 +22,11 @@ jobs:
         include:
           - os: ubuntu-latest
             python: '3.9'
-            tox_env: 'py39-test-alldeps_noray'
+            tox_env: 'py39-test-alldeps-cov'
             allowed_fail: false
           - os: ubuntu-latest
             python: '3.10'
-            tox_env: 'py310-test-alldeps-cov'
+            tox_env: 'py310-test-alldeps_noray'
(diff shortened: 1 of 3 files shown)


MRegeard:
On that, I was wondering wether we can (or not) benefit from a lazy import mechanism on `ray`? Of course, the well-known consequence of lazy importing is the time needed to import the module at each call. It would need some testing. 
However, I'm not really sure that it will solve our issues with the tests.

QRemy (approved):
Thanks @registerrier, this is fine for me

QRemy:
> On that, I was wondering wether we can (or not) benefit from a lazy import mechanism on `ray`?

For now it is only imported if the user specify to use ray as backend and if it is not available we fall back to multiprocessing.

registerrier:
The tests are running with the full deps on python 3.9. For now, all the python 3.10 CI will use the no_ray environments.
Issue #4691 should stay open until we find a proper solution.

There remain a CI fail from the sampling methods. The tests need to be changed as testing sampled values seems too unstable. This should tackled in a follow-up PR.

PR #4694, opened 2023-07-25
Fix event sampling writing

AtreyeeS:
This was recently brought up by Antonio Tutone in the help channel

### DIFF
--- a/examples/tutorials/analysis-3d/event_sampling.py
+++ b/examples/tutorials/analysis-3d/event_sampling.py
@@ -309,8 +309,9 @@ events.table.meta["OBJECT"] = dataset.models[0].name
 #
 
 primary_hdu = fits.PrimaryHDU()
-hdu_evt = fits.BinTableHDU(events.table)
-hdu_gti = fits.BinTableHDU(dataset.gti.table, name="GTI")
+hdu_gti = dataset.gti.to_table_hdu()
+hdu_evt = events.to_table_hdu()
+hdu_evt.header.update(observation.pointing.to_fits_header())
 hdu_all = fits.HDUList([primary_hdu, hdu_evt, hdu_gti])
 hdu_all.writeto("./event_sampling/events_0001.fits", overwrite=True)
 
@@ -561,6 +562,7 @@ display(data_store.obs_table)
 observations = data_store.get_observations()
 observations[0].peek()
 plt.show()
+print(observations[0])
 
 
 ######################################################################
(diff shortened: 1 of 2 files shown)


examples/tutorials/analysis-3d/event_sampling.py
```diff
@@ -309,8 +309,9 @@
 #
 
 primary_hdu = fits.PrimaryHDU()
-hdu_evt = fits.BinTableHDU(events.table)
-hdu_gti = fits.BinTableHDU(dataset.gti.table, name="GTI")
+hdu_gti = dataset.gti.to_table_hdu()
+hdu_evt = events.to_table_hdu()
+hdu_evt.header.update(observation.pointing.to_fits_header())
```
registerrier, line 314: Shouldn't this be moved into the event sampler code itself?
It is currently failing in make docs.

registerrier (commented):
Thanks @AtreyeeS . Shouldn't the `FixedPointingInfo` header writing be directly moved to the events meta writer?

fabiopintore:
Hi @AtreyeeS , I've seen the issue from Antonio but I couldn't investigate it, so thanks to you to have done it.
So, from what I've understood, it seems that his issue is related to the missing extension `GTI` that it's not recognised by `Datastore`. So I'm  a bit confused now, because you're acting on the keywords in the header of the `EVENTS` extension. What are the missing keywords in the sampled event files?

examples/tutorials/analysis-3d/event_sampling.py
maxnoe: This should be added to EventList.to_table_hdu directly, right?

maxnoe:
Aside from that we need to improve the simulator in light of #3767, the current code should work fine for GADF, it includes the pointing information here:

https://github.com/gammapy/gammapy/blob/3bb9cab6a4a76fa777983995bc6fa4f719b2b78e/gammapy/datasets/simulate.py#L412-L414 

into the meta dict, which is included into the table here:
https://github.com/gammapy/gammapy/blob/3bb9cab6a4a76fa777983995bc6fa4f719b2b78e/gammapy/datasets/simulate.py#L545

So what exactly is this fixing? Can we add a regression test for the specific encountered issue?

maxnoe:
This is not the correct fix. The actual issue is that for some reason, `ALT_PNT` and `AZ_PNT` are filled as string by the event sample code:

```
OBJECT  = 'point-pwl'                                                           
RA_OBJ  =    265.9185627628803                                                  
DEC_OBJ =    -28.6747888816758                                                  
...                                                 
ALT_PNT = '-13.534507893517661'                                                 
AZ_PNT  = '228.82981599564576'
```

and this trips up Gammapy when reading back in. The fix should be making sure that the correct type is used in the fits header!

AtreyeeS:
Closing for now. Addressed in #4698 and #4698

PR #4695, opened 2023-07-26
Add jupytext to dev env

maxnoe:
**Description**

This pull request adds jupytext to the dev env for working on the docs notebooks

**Dear reviewer**

### DIFF
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -53,6 +53,7 @@ dependencies:
   - flake8
   - isort
   - jinja2
+  - jupytext
   - nbsphinx
   - numdifftools
   - pandoc


AtreyeeS:
It would be good to add some info on how to use jupytext in this context in the dev docs, https://docs.gammapy.org/dev/development/doc_howto.html#choose-a-thumbnail-and-tooltip-for-the-tutorial-gallery

maxnoe:
Actually, you don't do anything special... You just open the jupyter notebook as usual. It just enables jupyter to load the python files we now use as notebook.

registerrier (approved):
Thanks @maxnoe !

Maybe we should introduce a systematic check on the notebooks too.

PR #4697, opened 2023-07-26
Deduce pointing mode from arguments in FixedPointingInfo

maxnoe:
**Description**

Checking up on #4694, I realized that the mode argument of the `FixedPointingInfo` is actually redundant and can be deduced from the given arguments.

So here, I deprecate the mode argument and implement the deduction from whether fixed_icrs and fixed_altaz are given.

I also realized that we didn't actually validate the given pointing directions, so this is now also done.

**Dear reviewer**

### DIFF
--- a/examples/tutorials/analysis-3d/event_sampling.py
+++ b/examples/tutorials/analysis-3d/event_sampling.py
@@ -74,7 +74,6 @@ We will work with the following functions and classes:
 -  `~gammapy.data.EventList`
 """
 
-
 ######################################################################
 # Setup
 # -----
@@ -90,10 +89,13 @@ from astropy.io import fits
 from astropy.time import Time
 from regions import CircleSkyRegion
 import matplotlib.pyplot as plt
-
-# %matplotlib inline
 from IPython.display import display
-from gammapy.data import DataStore, Observation, observatory_locations
+from gammapy.data import (
+    DataStore,
+    FixedPointingInfo,
+    Observation,
+    observatory_locations,
+)
 from gammapy.datasets import MapDataset, MapDatasetEventSampler
 from gammapy.irf import load_irf_dict_from_file
 from gammapy.makers import MapDatasetMaker
@@ -117,10 +119,9 @@ from gammapy.utils.check import check_tutorials_setup
 
 check_tutorials_setup()
 
-
 ######################################################################
 # Define an Observation
-# ~~~~~~~~~~~~~~~~~~~~~
+# ---------------------
 #
 # You can firstly create a `~gammapy.data.Observations` object that
 # contains the pointing position, the GTIs and the IRF you want to
@@ -131,21 +132,23 @@ check_tutorials_setup()
 # Galactic Center. We also fix the exposure time to 1 hr.
 #
 # Let’s start with some initial settings:
-#
 
 path = Path("$GAMMAPY_DATA/cta-caldb")
 irf_filename = "Prod5-South-20deg-AverageAz-14MSTs37SSTs.180000s-v0.1.fits.gz"
 
-pointing = SkyCoord(0.0, 0.0, frame="galactic", unit="deg")
+# telescope is pointing at a fixed position in ICRS for the observation
+pointing = FixedPointingInfo(
+    fixed_icrs=SkyCoord(0.0, 0.0, frame="galactic", unit="deg").icrs,
+)
 livetime = 1 * u.hr
+location = observatory_locations["cta_south"]
 
+
+irfs = load_irf_dict_from_file(path / irf_filename)
 
 ######################################################################
 # Now you can create the observation:
 #
-
-irfs = load_irf_dict_from_file(path / irf_filename)
-location = observatory_locations["cta_south"]
 
 observation = Observation.create(
     obs_id=1001,
@@ -154,11 +157,11 @@ observation = Observation.create(
     irfs=irfs,
     location=location,
 )
-
+print(observation)
 
 ######################################################################
 # Define the MapDataset
-# ~~~~~~~~~~~~~~~~~~~~~
+# ---------------------
 #
 # Let’s generate the `~gammapy.datasets.Dataset` object (for more info
 # on `~gammapy.datasets.Dataset` objects, please checkout
@@ -187,13 +190,12 @@ energy_axis_true = MapAxis.from_energy_bounds(
 migra_axis = MapAxis.from_bounds(0.5, 2, nbin=150, node_type="edges", name="migra")
 
 geom = WcsGeom.create(
-    skydir=pointing,
+    skydir=pointing.fixed_icrs,
     width=(2, 2),
     binsz=0.02,
     frame="galactic",
     axes=[energy_axis],
 )
-
 
 ######################################################################
 # In the following, the dataset is created by selecting the effective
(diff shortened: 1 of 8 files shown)


examples/tutorials/analysis-3d/event_sampling.py
```diff
@@ -117,10 +115,9 @@
 
 check_tutorials_setup()
 
-
 ######################################################################
 # Define an Observation
-# ~~~~~~~~~~~~~~~~~~~~~
+# ---------------------
```
maxnoe, line 120: This was needed to make the notebook render correctly in jupyter, it should also work in sphinx...

gammapy/data/pointing.py
```diff
@@ -25,6 +27,22 @@
 __all__ = ["FixedPointingInfo", "PointingInfo", "PointingMode"]
 
 
+def _check_coord(coord_or_frame, expected_frame, name):
```
bkhelifi, line 30: Could you add a small docstring?

gammapy/data/pointing.py
```diff
@@ -25,6 +27,22 @@
...
+    is_coord = isinstance(coord_or_frame, SkyCoord)
+    is_frame = isinstance(coord_or_frame, BaseCoordinateFrame)
+
+    if not (is_frame or is_coord):
+        raise TypeError(f"{name} must be a {SkyCoord} or {expected_frame} instance")
```
bkhelifi, line 35: {SkyCoord} -> `SkyCoord` ?

gammapy/data/pointing.py
bkhelifi, line 30: name is not used in this function. One could remove it, is not it?

examples/tutorials/analysis-3d/event_sampling.py
```diff
@@ -327,10 +318,9 @@
 counts.sum_over_axes().plot(add_cbar=True)
 plt.show()
 
-
 ######################################################################
 # Fit the simulated data
-# ~~~~~~~~~~~~~~~~~~~~~~
+# ------------------------------
```
bkhelifi, line 323:
```suggestion
# ----------------------
```

gammapy/data/pointing.py
maxnoe: Using `{SkyCoord}` will give the full import path:

```
In [3]: f"{SkyCoord}"
Out[3]: "<class 'astropy.coordinates.sky_coordinate.SkyCoord'>"
```

maxnoe: Which might be a bit too verbose... will change

gammapy/data/pointing.py
maxnoe: It is used in the error messages f-strings

AtreyeeS (commented):
Thank you @maxnoe! Indeed, the pointing mode is redundant, but it is probably easier for the user to deal with `DRIFT` & `POINTING` modes, instead of gammapy classes... I have to think a bit...

examples/tutorials/analysis-3d/event_sampling.py
```diff
@@ -131,34 +128,36 @@
...
 #
 # Let’s start with some initial settings:
-#
 
 path = Path("$GAMMAPY_DATA/cta-caldb")
 irf_filename = "Prod5-South-20deg-AverageAz-14MSTs37SSTs.180000s-v0.1.fits.gz"
 
-pointing = SkyCoord(0.0, 0.0, frame="galactic", unit="deg")
+# telescope is pointing at a fixed position in ICRS for the observation
+pointing = FixedPointingInfo(
```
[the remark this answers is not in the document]
maxnoe: I disagree, people should be forced to think about what kind of pointing they simulate

examples/tutorials/analysis-3d/event_sampling.py
```diff
@@ -314,7 +306,6 @@
 hdu_all = fits.HDUList([primary_hdu, hdu_evt, hdu_gti])
```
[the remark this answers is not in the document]
maxnoe: I can do that, but I'd prefer to do that in a different PR, this was not meant to be an update of the notebook, just a fix for the actual simulator

gammapy/data/pointing.py
```diff
@@ -125,40 +149,40 @@ def __init__(
             self.__dict__.update(self.from_fits_header(meta).__dict__)
             return
 
-        if not isinstance(mode, PointingMode):
-            raise TypeError(f"mode must be an instance of PointingMode, got {mode!r}")
+        if mode is not None:
+            warnings.warn(
+                "Passing mode is deprecated and the argument will be removed in Gammapy 2.0."
```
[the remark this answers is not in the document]
maxnoe: Since 1.1 is already released and we require the deprecation notice to be in at least one Gammapy release before it is removed, we cannot remove it in 1.2.

bkhelifi (approved):
Thanks @maxnoe.

gammapy/data/pointing.py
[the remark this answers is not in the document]
AtreyeeS: We need to support this in 1.2, and it will be removed since 1.3. I think we planned 4 minor releases between LTS releases, so 2.0 should come after that.

Would it be possible to use `deprecated_renamed_argument` here, eg: https://github.com/gammapy/gammapy/blob/5764a25fcca02815300c45a058f28799f118fe3d/gammapy/utils/deprecation.py#L35

registerrier (commented):
Thanks @maxnoe ! I have left a few inline suggestions.

The `FixedPointingInfo` can actually be further simplified.  In the end, it is only a `pointing` coordinate. The coordinate system `ICRS` or `AltAz` yields the `PointingMode`. A single `pointing` argument and attribute is enough to describe it.

gammapy/data/pointing.py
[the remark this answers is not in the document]
maxnoe: It's not really renamed, its removed, but if it is common, I could add a decorator also for that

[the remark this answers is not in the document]
maxnoe: > We need to support this in 1.2, and it will be removed since 1.3. I think we planned 4 minor releases between LTS releases, so 2.0 should come after that.

Ah ok, I thought only two minor between major releases

gammapy/data/pointing.py
```diff
@@ -125,40 +149,40 @@ def __init__(
...
-                raise ValueError("fixed_icrs is excluded for PointingMode.DRIFT")
-
-            self._fixed_altaz = fixed_altaz
-            self._fixed_icrs = None
         else:
-            raise ValueError(f"Unsupported pointing mode for FixedPointingInfo: {mode}")
+            _check_coord_frame(fixed_altaz, AltAz, "fixed_altaz")
+            self._mode = PointingMode.DRIF
+            self._fixed_icrs = None
+            self._fixed_altaz = fixed_altaz
```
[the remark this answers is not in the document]
maxnoe: It makes the code in the getters much easier and it doens't really cost much, right? Instead of checking in the getters the frame, these can just return the internal state.

bkhelifi (approved):
Thanks a lot @maxnoe.

Astro-Kirsty (approved):
Thanks @maxnoe, this is a nice update to the code.
It all looks good to me, are we ready to merge?

registerrier:
Merging now.

PR #4698, opened 2023-07-26
Use correct type for alt/az pointing in event sampler, add observatory location info

maxnoe:
**Description**

This pull request fixes the data type of `{ALT,AZ}_PNT` filled by the event sampler, which was for some reason stored as string, not as float.

Additionally to the existing checks, the test now also writes the observation out and reads it back in.

For convenience, I added a setter for `Observatoin.events`, so that we can now do:
```
observation.events = sampler.run(observation)
```

I also noticed now that the observatory location info is not added, which I fixed as well here.
**Dear reviewer**

### DIFF
--- a/gammapy/datasets/tests/test_simulate.py
+++ b/gammapy/datasets/tests/test_simulate.py
@@ -582,7 +582,9 @@ def test_mde_run(dataset, models, tmp_path):
     path = tmp_path / "obs.fits.gz"
     obs.write(path)
     obs_back = Observation.read(path)
-    assert obs_back.observatory_earth_location == LOCATION
+    assert u.isclose(obs_back.observatory_earth_location.lon, LOCATION.lon)
+    assert u.isclose(obs_back.observatory_earth_location.lat, LOCATION.lat)
+    assert u.isclose(obs_back.observatory_earth_location.height, LOCATION.height)
 
 
 @requires_data()
--- a/gammapy/datasets/tests/test_simulate.py
+++ b/gammapy/datasets/tests/test_simulate.py
@@ -479,7 +479,7 @@ def test_event_det_coords(dataset, models):
 
 
 @requires_data()
-def test_mde_run(dataset, models):
+def test_mde_run(dataset, models, tmp_path):
     irfs = load_irf_dict_from_file(
         "$GAMMAPY_DATA/cta-1dc/caldb/data/cta/1dc/bcf/South_z20_50h/irf_file.fits"
     )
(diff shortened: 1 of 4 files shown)


gammapy/datasets/tests/test_simulate.py
```diff
@@ -540,14 +540,21 @@ def test_mde_run(dataset, models):
...
     assert meta["INSTRUME"] == "1DC"
     assert meta["N_TELS"] == ""
     assert meta["TELLIST"] == ""
 
+    # test writing out and reading back in works
+    obs.events = events
+    path = tmp_path / "obs.fits.gz"
+    obs.write(path)
+    obs_back = Observation.read(path)
+    assert obs_back.observatory_earth_location == LOCATION
```
AtreyeeS, line 556: maybe add an assert on obs.events to ensure it is being set correctly?

AtreyeeS (approved):
Thank you very much @maxnoe . This makes #4694 rather redundant, I will close it now

maxnoe:
The windows CI failure is unrelated and comes from the ray log issue

gammapy/datasets/tests/test_simulate.py
maxnoe: Done

gammapy/datasets/tests/test_simulate.py
```diff
@@ -551,14 +551,21 @@ def test_mde_run(dataset, models):
...
```
bkhelifi, line 567: This test is failing in the CI. It seems that the __eq__ is not implemented for the class `ObservationLocation`.

bkhelifi:
Thanks for the fix! It looks good...
If you could add some additional tests, it will pass completely the codecov test.

registerrier (dismissed):
Thanks @maxnoe . This looks good to me.

fabiopintore:
Green light from my side, I don't have further comments. Thanks @maxnoe

bkhelifi (dismissed):
Thanks @maxnoe, looks good for me

Astro-Kirsty:
Thanks @maxnoe. This will need a rebase!
Then we should be good to merge

maxnoe:
@Astro-Kirsty Done, sorry for the long inactivity

fabiopintore:
Thanks @maxnoe . This looks good to me. No further comments from my side.

Astro-Kirsty (approved):

registerrier (approved):
Thanks @maxnoe ! Sorry for having let this slip though.

PR #4701, opened 2023-07-28
Add 'norm' option to `plot_distribution`

MRegeard:
This PR introduce a 'norm' option to `plot_distribution` to directly fit a normal distribution, as mentioned in #4665.

### DIFF
--- a/gammapy/visualization/utils.py
+++ b/gammapy/visualization/utils.py
@@ -218,6 +218,9 @@ def plot_distribution(
 
     Returns
     -------
+    axes : `~numpy.ndarray` of `~matplotlib.pyplot.Axes`
+        Array of Axes.
+
     result_list : list of dict
         List of dictionnary that contains the results of `scipy.optimize.curve_fit`. The number of elements in the list
         correspond to the dimension of the non-spatial axis of the map.
@@ -228,9 +231,6 @@ def plot_distribution(
             * `covar` : the covariance matrix for the fitted parameters `param`
             * `info_dict` : the `infodict` return of `scipy.optimize.curve_fit`
 
-    axes : `~numpy.ndarray` of `~matplotlib.pyplot.Axes`
-        Array of Axes.
-
     Examples
     --------
     >>> from gammapy.datasets import MapDataset
@@ -239,10 +239,10 @@ def plot_distribution(
     >>> from gammapy.visualization import plot_distribution
     >>> dataset = MapDataset.read("$GAMMAPY_DATA/cta-1dc-gc/cta-1dc-gc.fits.gz")
     >>> tsmap_est = TSMapEstimator().run(dataset)
-    >>> res, ax = plot_distribution(tsmap_est.sqrt_ts, func="norm", kwargs_hist={'bins': 75, 'range': (-10, 10), 'density': True})
+    >>> axs, res = plot_distribution(tsmap_est.sqrt_ts, func="norm", kwargs_hist={'bins': 75, 'range': (-10, 10), 'density': True})
     >>> # Equivalently, one can do the following:
     >>> func = lambda x, mu, sig : norm.pdf(x, loc=mu, scale=sig)
-    >>> res, ax = plot_distribution(tsmap_est.sqrt_ts, func=func, kwargs_hist={'bins': 75, 'range': (-10, 10), 'density': True})
+    >>> axs, res = plot_distribution(tsmap_est.sqrt_ts, func=func, kwargs_hist={'bins': 75, 'range': (-10, 10), 'density': True})
     """
 
     from gammapy.maps import WcsNDMap  # import here to avoid circular import
@@ -345,4 +345,4 @@ def plot_distribution(
         axe.set(**kwargs_axes)
         axe.legend()
 
-    return result_list, axes
+    return axes, result_list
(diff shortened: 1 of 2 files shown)


gammapy/visualization/utils.py
```diff
@@ -293,9 +297,32 @@ def plot_distribution(
...
+
+                def func(x, mu, sigma):
+                    return norm.pdf(x, mu, sigma)
+
+                pars, cov, infodict, message, _ = curve_fit(
+                    func, centers, n, **kwargs_fit
+                )
+
+                mu, sig = pars[0], pars[1]
+                err_mu, err_sig = cov[0][0] ** 2, cov[1][1] ** 2
```
registerrier, line 313: it should be the square root no?

gammapy/visualization/utils.py
```diff
@@ -293,9 +297,32 @@ def plot_distribution(
...
+
+                label_norm = (
+                    r"$\mu$ = {:.2f} ± {:.2E}, $\sigma$ = {:.2f} ± {:.2E}".format(
+                        mu, err_mu, sig, err_sig
+                    )
+                )
+                kwargs_plot_fit["label"] = label_norm
```
registerrier, line 320: Doesn't this clutter a bit the output plot?

gammapy/visualization/utils.py
```diff
@@ -293,9 +297,32 @@ def plot_distribution(
         n, bins, _ = axe.hist(d, **kwargs_hist)
 
         if func is not None:
+            kwargs_plot_fit = {"label": "Fit"}
             centers = 0.5 * (bins[1:] + bins[:-1])
 
-            pars, cov, infodict, message, _ = curve_fit(func, centers, n, **kwargs_fit)
+            if func == "norm":
```
registerrier, line 303: I would rather do this early and set `func` to the new function.

registerrier (commented):

How does the addition of the specific label clutter the resulting plot? Maybe printing the result would be good enough?

gammapy/visualization/utils.py
MRegeard: Yes indeed, thanks !

gammapy/visualization/utils.py
MRegeard: Because you said that I change the label so that it is written on 2 lines. It looks like this :
plot_dist_norm_label

I suppose it's ok, right ?

bkhelifi:
The I/O function (dict, yaml) seem to not be covered by the tests. Is it really the case?

bkhelifi (commented):
I have left a comment in the discussion. Otherwise, it seems ready to be merged.

MRegeard:
@bkhelifi, I added some more tests on the result dictionary. Is that what you wanted ?

bkhelifi:
Yes, thanks. Unfortunately, the CI tests block the whole...

bkhelifi (approved):
Thank you @MRegeard

gammapy/visualization/utils.py
```diff
@@ -204,9 +205,10 @@ def plot_distribution(
         Axis object to plot on. If a list of Axis is provided it has to be the same length as the length of _map.data.
     ncols : int
         Number of columns to plot if a "plot grid" was to be done.
-    func : function object
-        The function to pass to `scipy.optimize.curve_fit`. Default is None.
-        If None, no fit will be done.
+    func : function object or str
+        The function to used to fit a map data histogram or "norm". Default is None.
```
Astro-Kirsty, line 209: The wording here does not make sense
`The function is used`?

gammapy/visualization/utils.py
```diff
@@ -204,9 +205,10 @@ def plot_distribution(
...
+        If None, no fit will be done. If "norm" is given, `scipy.stats.norm.pdf`
```
Astro-Kirsty, line 210: `If None, no fit will be performed.`

Astro-Kirsty (changes requested):
Thanks @MRegeard. I added a few minor comments, otherwise looks good

gammapy/visualization/utils.py
MRegeard: I actually think `The function used to ...` is better, what do you think ?

Astro-Kirsty (dismissed):

bkhelifi (dismissed):
Thank you @MRegeard . Looks good for me!
Should we make a new PR to insert this function in tutorials?

MRegeard:
> Thank you @MRegeard . Looks good for me! Should we make a new PR to insert this function in tutorials?

@bkhelifi, there is a PR pending this one to insert this in the tutorials !

gammapy/visualization/utils.py
```diff
@@ -216,6 +218,9 @@ def plot_distribution(
 
     Returns
     -------
+    axes : `~numpy.ndarray` of `~matplotlib.pyplot.Axes`
+        Array of Axes.
+
```
registerrier, line 223:
```suggestion
```

bkhelifi:
> > Thank you @MRegeard . Looks good for me! Should we make a new PR to insert this function in tutorials?
> 
> @bkhelifi, there is a PR pending this one to insert this in the tutorials !

I think that it could be done afterwards....

gammapy/visualization/utils.py
registerrier: Well I think it would have been simpler. We'll reorganize later.

registerrier (approved):
Thanks @MRegeard . This is nice.

A bit of clean-up might be good to avoid the too many indents. But we can do that later.

Issue #4709, opened 2023-07-31
Issues with RandomState and reproducible inverse CDF sampling


registerrier:
Currently a number of tests in event sampler fail with numpy 1.25 and ubuntu. The exact event coordinates are not reproduced even thought they are sampled from a fully constant npred map. (see e.g. test_mde_sample_background in test_simulate.py).

Introducing a similar test directly on `InverseCDFSampler` (e.g. testing sampled values with a predefined `RandomState`) already fails even though the CDF is the same and `RandomState.get_state()` returns the same tuple.  
See this test batch: https://github.com/registerrier/gammapy/actions/runs/5716269531  which works on all platforms/environment except ubuntu.

From the numpy documentation, it seems that the stability of legacy random generators is not fully guaranteed on al platforms. Besides, they advise changing to `Generator` (which does not have a version to version compatibility). 
https://numpy.org/doc/stable/reference/random/compatibility.html

We might have to remove all tests checking specific sampled values to avoid this happening gain and again. We could then change from `numpy.random.RandomState` to `numpy.random.Generator`.

Opinions @adonath , @fabiopintore , @QRemy , @AtreyeeS ?

registerrier:
No issues recently in the CI. Closing for now.

PR #4713, opened 2023-08-03
Remove sampling value tests

registerrier:
**Description**

This pull request modifies a number of tests in datasets/tests/test_simulate.py . Because the random generator is completely not stable from version to version of numpy, the tests that check a specific value of a sampled event quantities (e.g. energy or position) can fail. This PR modifies the failing tests to measure mean values or min and max values of the quantities. 

It should help solve #4709 

**Dear reviewer**


AtreyeeS (approved):
Thanks @registerrier ! Indeed, this should be more stable!

QRemy (approved):

PR #4714, opened 2023-08-04
Replace CashCountsStatistic error calculation by analytical expression

registerrier:
**Description**

This pull request replaces the current asymmetric error calculation as well as the `n_sig_matching_significance` function by analytical expression for the `CashCountsStatistic`. This uses the `scipy.special.lambertw` function.
The upper limit calculation should be adapted too.

This makes the error calculation more than 1e3 times faster than the current root searching scheme.

I don't know if there is an analytical solution for the WStat. 

**Dear reviewer**


QRemy (approved):
Thanks @registerrier, looks good !

adonath:
Nice @registerrier! This is especially useful for the  `ExcessMapEstimator`, where the root finding is run for each pixel.

gammapy/stats/counts_statistic.py
```diff
@@ -315,6 +316,37 @@ def sum(self, axis=None):
     def __getitem__(self, key):
         return CashCountsStatistic(n_on=self.n_on[key], mu_bkg=self.n_bkg[key])
 
+    def compute_errn(self, n_sigma=1.0):
```
bkhelifi, line 319: Should the docstrings for the 3 functions be repeated?

bkhelifi (commented):
Thanks a lot... Except a small question about the docstring, it is ok for me!

adonath (approved):
Thanks @registerrier! I have nor further comments, I think this is ready to be merged?

registerrier:
I have added the `compute_upper_limit` calculation as well and modified the tolerance of sensitivity tests in test_ excess.py.

PR #4719, opened 2023-08-18
Fix devdeps test setup

QRemy:
See tox issues 3096

https://github.com/tox-dev/tox/issues/3096

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -7,6 +7,7 @@ envlist =
     build_docs
     linkcheck
     codestyle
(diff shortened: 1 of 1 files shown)


QRemy:
Now the test on the devdeps env are running but there is a new fail.
Should we fix it in this PR or a separate ? @adonath, @registerrier

registerrier (approved):

I think the corrections for the devdeps test fails can be kept for a future PR.

Issue #4515, opened 2023-05-19
Expose NormSpectralModels in Model Gallery


AtreyeeS:
In the present dev, 1.0.1, only the `PiecewiseNormSpectralModel` (https://docs.gammapy.org/dev/user-guide/model-gallery/index.html#spectral-models ) is shown among all norm models.
I understand the motivation, but this gives an incomplete info about the number of supported models.
I think should show all the norm models.

Astro-Kirsty:
I am happy to implement this, however, could you please explain the motivation behind wanting to expose all the `NormSpectralModels`? 
They don't seem to be used a lot, and I wonder if exposing them in this specific manner might confuse the user into wanting to use them in incorrect situations?

AtreyeeS:
We advertise the Models Gallery as a place to know the available models in gammapy, so it would be good to have it complete. A reason why these models are not used is because people don't know that they are there! (kind of like the discussion we had at our last dev call about not advertising our features)


What can be useful is to show how these models affect a Spectral Model, eg, the original model (a power law) and original model*norm model on the same plot..

like... (with better plot labels)

![Unknown](https://github.com/gammapy/gammapy/assets/32677370/93c9afae-fc69-41d2-b10b-1fe602a1c95d)

PR #4721, opened 2023-08-21
Exposing NormSpectralModels

Astro-Kirsty:
This PR is to fix issue [#4515](https://github.com/gammapy/gammapy/issues/4515)
I have implemented three new scripts to expose the `NormSpectralModels` in the Model Gallery

**Dear reviewer**
I have been thinking about this a bit, although this solves issue #4515, I would like some further discussion. Do we really want to expose them in this manner?
For example, we wouldn't typically add the `PowerLawNormSpectralModel` to a `PowerLawSpectralModel`. One very obvious use case is adding the `ExpCutoffPowerLawNormSpectralModel` to a `PowerLawSpectralModel`. Or creating a `TemplateSpectralModel` and then applying the `NormSpectralModel`. 
Therefore, I think it is a good idea to discuss further the use cases of these `NormSpectralModels`, and therefore the best way to expose them in the model gallery.

### DIFF
--- a/examples/models/spectral/plot_exp_cutoff_powerlaw_norm_spectral.py
+++ b/examples/models/spectral/plot_exp_cutoff_powerlaw_norm_spectral.py
@@ -1,0 +1,46 @@
+r"""
+.. _exp-cutoff-powerlaw-norm-spectral-model:
+
+Exponential cutoff power law norm spectral model
+========================
+
+This model parametrises a cutoff power law spectral correction with a norm parameter.
+
+"""
+
+# %%
+# Example plot
+# ------------
+# Here is an example plot of the model:
+
+from astropy import units as u
+import matplotlib.pyplot as plt
+from gammapy.modeling.models import (
+    ExpCutoffPowerLawNormSpectralModel,
+    ExpCutoffPowerLawSpectralModel,
+    Models,
+    SkyModel,
+)
+
+energy_bounds = [0.1, 100] * u.TeV
+
+model = ExpCutoffPowerLawSpectralModel()
+norm = ExpCutoffPowerLawNormSpectralModel(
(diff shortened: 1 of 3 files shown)


QRemy:
An example with a `TemplateSpectralModel` is already shown here :
https://docs.gammapy.org/dev/tutorials/api/models.html?#norm-spectral-models
The initial use case was to use them with the `FoVBacgroundModel` or  for diffuse models using 3D `TemplateSpatialModel`.
So if we want to add them in a tutorial I would show a `FoVbackgroundMaker` fit with a `PiecewiseNormSpectralModel`.
Moreover the Fermi-LAT tutorial could be extended adding a fit of the diffuse with a `PiecewiseNormSpectralModel` (for now it is a PowerlawNormSpectralModel) then we could show a plot comparing the two spectral corrections.

Astro-Kirsty:
Thanks @QRemy. The purpose of this PR was only to expose the available `NormSpectralModel` options within the [Model Gallery](https://docs.gammapy.org/dev/user-guide/model-gallery/index.html#spectral-models).

Of course, I am open to adapting a tutorial to add the model.

QRemy:
For the model gallery we should just just repoduce the example with the `TemplateSpectralModel` (https://docs.gammapy.org/dev/tutorials/api/models.html?#norm-spectral-models) because the multiplication with any other spectral model will lead to degenerated parameters.

QRemy (approved):
Thanks @Astro-Kirsty, no further comments from my side.
The remaining tests fail is unrelated.

cgalelli (approved):

cgalelli:
Thanks @Astro-Kirsty. Tried the examples. No comments from my side

Issue #4702, opened 2023-07-29
get_overlap_fraction fails for different units in true and reco axis


tunbehaun273:
The current version of [gammapy/irf/edisp/map.py/get_overlap_fraction](https://github.com/gammapy/gammapy/blob/v1.2.dev/gammapy/irf/edisp/map.py#L11) fails to give proper values when the unit of the true and reco axis are not the same as the unit (i.e. MeV/TeV) is returned. This affects for example the `from_diagonal_response` of the EDispKernel.

As solution I would suggest to implement the following:
```
def get_overlap_fraction(energy_axis, energy_axis_true):
    a_min = energy_axis.edges[:-1]
    a_max = energy_axis.edges[1:]

    b_min = energy_axis_true.edges[:-1][:, np.newaxis]
    b_max = energy_axis_true.edges[1:][:, np.newaxis]

    xmin = np.fmin(a_max, b_max)
    xmax = np.fmax(a_min, b_min)
    fractions = np.clip(xmin - xmax, 0, np.inf) / (b_max - b_min)
    return fractions.to_value('')
```

PR #4722, opened 2023-08-21
Fix get_overlap_fraction for different energies

Astro-Kirsty:
This PR resolves issue [#4702](https://github.com/gammapy/gammapy/issues/4702)
Namely, it allows different units for the `energy_true_axis` and `energy_axis` to be defined and still achieve the correct result. 

I added to the `test_kernel.py` to test this behaviour.

(Side note: I just fixed a typo in another test function too)

### DIFF
--- a/gammapy/irf/edisp/map.py
+++ b/gammapy/irf/edisp/map.py
@@ -17,7 +17,7 @@ def get_overlap_fraction(energy_axis, energy_axis_true):
 
     xmin = np.fmin(a_max, b_max)
     xmax = np.fmax(a_min, b_min)
-    return np.clip(xmin - xmax, 0, np.inf) / (b_max - b_min)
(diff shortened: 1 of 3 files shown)


gammapy/irf/edisp/map.py
```diff
@@ -17,7 +17,7 @@ def get_overlap_fraction(energy_axis, energy_axis_true):
 
     xmin = np.fmin(a_max, b_max)
```
bkhelifi, line 18: What happens if the units of a_max and b_max are different here? Can it happen?

bkhelifi (approved):
Thanks @Astro-Kirsty

AtreyeeS (approved):
Thanks @Astro-Kirsty ! No comments from my side

PR #4726, opened 2023-08-24
Introduce a function to compute the point-to-point fractional variance

cgalelli:
After the merging of pr #4501 I wrote a parallel script providing the computation for the point-to-point fractional variation, another estimator of temporal variability, very similar to the excess fractional variation but quite different in code implementation. 

This PR includes the code for the computation of the variable itself in `stats/variability.py`, as well as a wrapper in `estimators/utils.py`, each with its test.

PR #4501, opened 2023-05-12
Exposing computation of the fractional excess variance

cgalelli:
**Description**

This pull request adds a wrapper function in estimators/utils.py to access the computation of the fractional excess variance contained in stats/variability.py useful for the evaluation of temporal variability. This addresses #4471

PR #4727, opened 2023-08-24
Introduce a function to compute the doubling/halving time for a lightcurve

cgalelli:
Similarly to pr #4501 and pr #4726, I wrote a script providing the computation of the doubling/halving time, useful for studies of temporal variability in lightcurves (example: AGN flares)

This PR includes the code for the computation of the variable itself in stats/variability.py, as well as a wrapper in estimators/utils.py, each with its test.

### DIFF
--- a/gammapy/datasets/map.py
+++ b/gammapy/datasets/map.py
@@ -2292,6 +2292,8 @@ class MapDatasetOnOff(MapDataset):
         """
         name = make_name(name)
 
+        background = self.counts_off * self.alpha if self.counts_off else None
+
         return MapDataset(
             counts=self.counts,
             exposure=self.exposure,
@@ -2301,7 +2303,7 @@ class MapDatasetOnOff(MapDataset):
             gti=self.gti,
             mask_fit=self.mask_fit,
             mask_safe=self.mask_safe,
-            background=self.counts_off * self.alpha,
+            background=background,
             meta_table=self.meta_table,
         )
 
--- a/gammapy/datasets/tests/test_map.py
+++ b/gammapy/datasets/tests/test_map.py
@@ -1667,6 +1667,10 @@ def test_to_map_dataset():
     assert isinstance(dataset, MapDataset)
     assert dataset.counts == dataset_onoff.counts
 
+    dataset_onoff.counts_off = None
+    dataset2 = dataset_onoff.to_map_dataset(name="ds2")
+    assert dataset2.background is None
+
 
 def test_downsample_onoff():
     axis = MapAxis.from_energy_bounds(1, 10, 4, unit="TeV")
--- a/gammapy/makers/background/reflected.py
+++ b/gammapy/makers/background/reflected.py
@@ -569,7 +569,7 @@ class ReflectedRegionsBackgroundMaker(Maker):
         Returns
         -------
         dataset_on_off : `SpectrumDatasetOnOff`
-            On off dataset.
+            On-Off dataset.
         """
         counts_off, acceptance_off = self.make_counts_off(dataset, observation)
         acceptance = RegionNDMap.from_geom(geom=dataset.counts.geom, data=1)


gammapy/stats/variability.py
```diff
@@ -82,3 +84,76 @@ def compute_chisq(flux):
...
+        * np.log(2)
+        / flux[..., :-1]
+        * np.log(flux[..., 1:] / flux[..., :-1]) ** 2
+    )
+    times_err = np.sqrt(
+        (flux_err[..., 1:] * times_err_1) ** 2 + (flux_err[..., :-1] * times_err_2) ** 2
+    )
+
+    imin = np.argmin(
+        np.where(np.logical_and(np.isfinite(times), times > 0), times, 1e15 * u.s),
```
AtreyeeS, line 145: you can use `np.infty` instead of `1e15`

AtreyeeS, line 145: you can use `np.infty` instead of 1e15

gammapy/stats/variability.py
```diff
@@ -82,3 +84,76 @@ def compute_chisq(flux):
...
+    ----------
+    ..[Brown2013] "Locating the γ-ray emission region
+    of the flat spectrum radio quasar PKS 1510−089", Brown et al. (2013)
+    https://academic.oup.com/mnras/article/431/1/824/1054498
+    """
+
+    flux = np.atleast_2d(flux).swapaxes(0, axis).T
+    flux_err = np.atleast_2d(flux_err).swapaxes(0, axis).T
+
+    times = (time[1:] - time[:-1]) / np.log2(flux[..., 1:] / flux[..., :-1])
```
AtreyeeS, line 127: In this case, I suppose passing time edges is necessary, not time centers?

gammapy/stats/variability.py
```diff
@@ -82,3 +84,76 @@ def compute_chisq(flux):
...
+        np.where(np.logical_and(np.isfinite(times), times < 0), times, -1e15 * u.s),
+        axis=-1,
+        keepdims=True,
+    )
+    index = np.concatenate([imin, imax], axis=-1)
+
+    dtime = np.take_along_axis(times, index, axis=-1)
+    dtime_err = np.take_along_axis(times_err, index, axis=-1)
+
+    return dtime, dtime_err
```
AtreyeeS, line 159: Maybe it is also useful to know which flare/at what time this doubling scale is obtained. For a typical AGN light curve, you can have multiple flares...

gammapy/estimators/utils.py
```diff
@@ -346,3 +347,55 @@ def compute_lightcurve_fvar(lightcurve, flux_quantity="flux"):
...
+        ],
+        names=(
+            "min_energy",
+            "max_energy",
+            "doubling_time",
+            "dtime_err",
+            "halving_time",
+            "htime_err",
+        ),
+        meta=lightcurve.meta,
```
AtreyeeS, line 398: I don't know if this meta info is relevant for the table...

gammapy/estimators/utils.py
```diff
@@ -346,3 +347,55 @@ def compute_lightcurve_fvar(lightcurve, flux_quantity="flux"):
     )
 
     return table
+
+
+def compute_lightcurve_dtime(lightcurve, flux_quantity="flux"):
+    r"""Compute the doubling time of the input lightcurve.
```
AtreyeeS, line 353: Should mention that this is the minimum doubling time scale...

gammapy/stats/variability.py
```diff
@@ -82,3 +84,76 @@ def compute_chisq(flux):
...
+    axis : int, optional
+        Axis along which the value is computed.
+        The default is to compute the value on axis 0.
+
+    Returns
+    -------
+    dtime, dtime_err : `~numpy.ndarray`
+        Characteristic doubling time, halving time and errors.
+
+    References
```
AtreyeeS, line 117: Since users will be more exposed to the wrapper function, this detailed doc string can lie there..

AtreyeeS (commented):
Thanks @cgalelli for this necessary functionality! 

One general comment is that `compute_dtime` might be more useful than simply only for light curves, eg, for flux profiles, it can be interesting to know the spatial scales at which the fluxes change. In this case, maybe

1. You can adapt `compute_dtime` to take any axis as an input. The naming can be adapted accordingly, eg, `compute_flux_doubling_scale` etc
2. The `compute_lightcure_dtime` can reside in `LightCurve` directly, rather than in utils. It improves the visibility...

cgalelli:
> Thanks @cgalelli for this necessary functionality!
> 
> One general comment is that `compute_dtime` might be more useful than simply only for light curves, eg, for flux profiles, it can be interesting to know the spatial scales at which the fluxes change. In this case, maybe
> 
> 1. You can adapt `compute_dtime` to take any axis as an input. The naming can be adapted accordingly, eg, `compute_flux_doubling_scale` etc
> 2. The `compute_lightcure_dtime` can reside in `LightCurve` directly, rather than in utils. It improves the visibility...

Thanks for all the feedback @AtreyeeS ! I did not think about the expanded use cases for this function. I can surely make the necessary changes if you think this is interesting for more than just temporal variability. Do you think I should keep the "core" function in `variability.py`?

I'll also work on all the other points you raised. Thanks again!

gammapy/estimators/utils.py
cgalelli: Absolutely, it is mostly a placeholder for future metadata or to not leave the table without

gammapy/stats/variability.py
cgalelli: Done, thanks!

gammapy/stats/variability.py
AtreyeeS: My point was that from `compute_lightcurve_dtime` you pass the time centers. As per your logic here, you should be passing the edges, no?

gammapy/stats/variability.py
cgalelli: Added in the new commit, thanks!

bkhelifi (approved):
Thanks @cgalelli . It seems good!

gammapy/stats/variability.py
cgalelli: Sorry for the delay in the response. I don't think any between edges (min or max) or centers are perfect. I would still prefer to pass the centers of the bins in this case as they better represent an "average coordinate" of the flux point. Sorry if I am missing something obvious, but why are edges more logic here?

bkhelifi (approved):
All good!

registerrier (changes requested):
Thanks @cgalelli . This is a nice addition.

I think the implementation is too complex as is and should be simplified. Some suggestions:
- the idea of having the possibility to deal with other axes is interesting but I would keep that for the generic function working with numpy arrays/Quantity. The object working with `FluxPoints` should be kept simple and compute only flux doubling time scales.
- the `compute_flux_doubling` function implementation is too complex and should be clarified.
- there is no explicit expression given for the flux doubling time and its error in the reference given. If you have to derive it manually I think one should give the expression in the docstring.

As a more general comment, I wonder what the best approach could be. Technically, what is done is fitting a temporal model with one dof per time bin. Would it be possible or make sense to extract the information by fitting directly a dataset?

gammapy/estimators/utils.py
```diff
@@ -346,3 +348,107 @@ def compute_lightcurve_fvar(lightcurve, flux_quantity="flux"):
     )
 
     return table
+
+
+def compute_lightcurve_flux_doubling(lightcurve, axis_name, flux_quantity="flux"):
```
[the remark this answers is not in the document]
cgalelli: Thanks Regis, in the latest commit I removed this possibility and left only the doubling time as a wrapper for FluxPoints objects. The base function can still use whatever axis.

gammapy/stats/variability.py
```diff
@@ -82,3 +83,76 @@ def compute_chisq(flux):
...
+    axis : int, optional
+        Axis along which the value is computed.
+
+    Returns
+    -------
+    doubling, doubling_err, coord : `~numpy.ndarray`
+        Characteristic flux doubling, halving and errors,
+        with coordinates at which they were found.
+    """
+
```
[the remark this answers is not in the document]
cgalelli: After researching possible alternatives, it seems in my opinion that axis swapping is still the best solution for this.

registerrier:
Merging #4726 has introduced some merge conflicts unfortunately. 
Can you rebase @cgalelli ?

PR #4729, opened 2023-08-28
Change default index for  NormSpectralModel

QRemy:
This PR change to 0 the default index value of the `ExpCutoffPowerLawNormSpectralModel ` and `LogParabolaNormSpectralModel`  such as it match the default tilt value of the `PiecewiseNormSpectralModel`.
Do we need/have a deprecation message for that ?

### DIFF
--- a/gammapy/modeling/models/spectral.py
+++ b/gammapy/modeling/models/spectral.py
@@ -3,6 +3,7 @@
 import logging
 import operator
 import os
+import warnings
 from pathlib import Path
 import numpy as np
 import scipy.optimize
@@ -16,6 +17,7 @@ import matplotlib.pyplot as plt
 from gammapy.maps import MapAxis, RegionNDMap
 from gammapy.maps.axes import UNIT_STRING_FORMAT
 from gammapy.modeling import Parameter, Parameters
+from gammapy.utils.deprecation import GammapyDeprecationWarning
 from gammapy.utils.integrate import trapz_loglog
 from gammapy.utils.interpolation import (
     ScaledRegularGridInterpolator,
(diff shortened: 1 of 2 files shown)


bkhelifi:
Can you also precise these default values into the docstring? Even with a deprecation message, the documentation should also provide this information.

QRemy:
> Can you also precise these default values into the docstring? Even with a deprecation message, the documentation should also provide this information.

None of the models show the default parameters so this should be done in a dedicated PR.

bkhelifi:
As mentioned in a dev call, one might raise a warning message to inform the user that the default behaviour has changed... At least, for one release as for the depreciation warning

Astro-Kirsty:
I have adapted the functions such that a deprecation warning shows when the default values are used

registerrier (approved):
Thanks @QRemy and @Astro-Kirsty .  This looks good.

I think this should not be backported as it changes the API but should go in  1.2 only.

Note that there are tests using the default values. They should be adapted and deprecation warnings must be caught as the CI fails if tests emit a deprecation warning.

gammapy/modeling/models/spectral.py
```diff
@@ -1332,6 +1334,29 @@ class ExpCutoffPowerLawNormSpectralModel(SpectralModel):
     lambda_ = Parameter("lambda_", "0.1 TeV-1")
     alpha = Parameter("alpha", "1.0", frozen=True)
 
+    def __init__(
+        self, index=None, norm=None, reference=None, lambda_=None, alpha=None, **kwargs
+    ):
+
+        if index is None:
+            warnings.warn(
+                "The default index value changed from 1.5 to 0 since v1.0.1",
```
registerrier, line 1343:
```suggestion
                "The default index value changed from 1.5 to 0 since v1.2",
```

gammapy/modeling/models/spectral.py
```diff
@@ -1594,15 +1619,42 @@ class LogParabolaNormSpectralModel(SpectralModel):
...
-    alpha = Parameter("alpha", 2)
-    beta = Parameter("beta", 1)
+    alpha = Parameter("alpha", 0)
+    beta = Parameter("beta", 0)
+
+    def __init__(self, norm=None, reference=None, alpha=None, beta=None, **kwargs):
+
+        if alpha is None:
+            warnings.warn(
+                "The default alpha value changed from 2 to 0 since v1.0.1",
```
registerrier, line 1638:
```suggestion
                "The default alpha value changed from 2 to 0 since v1.2",
```

gammapy/modeling/models/spectral.py
```diff
@@ -1594,15 +1619,42 @@ class LogParabolaNormSpectralModel(SpectralModel):
...
+                GammapyDeprecationWarning,
+            )
+
+        if beta is None:
+            warnings.warn(
+                "The default beta value changed from 1 to 0 since v1.0.1",
```
registerrier, line 1644:
```suggestion
                "The default beta value changed from 1 to 0 since v1.2",
```

registerrier (approved):
Thanks @Astro-Kirsty & @QRemy

registerrier:

Issue #4672, opened 2023-07-20
Replace Observation.obs_info dictionary  with proper MetaData object


registerrier:
Currently all `Observation` metadata are stored in the `obs_info` dictionary. On should instead:
- define an `ObservationMetaData` class with required and optional attributes
- replace the `obs_info` with this object stored as `Observation.meta`

Expected attributes should be at least:
- obs_id : Union[str, int]
- telescope : str
- instrument : Optional[str] = None
- observation_mode : Optional[str] = None
- location : Optional[Union[str, EarthLocation]] = None
- creation : Optional[CreatorMetaData] = None

What else should we add?

registerrier:
This is now solved by #4738. Closing.

PR #4738, opened 2023-08-30
Introduce Observation metadata container

registerrier:
**Description**

This pull request introduces the `ObservationMetaData` as per #4672 

**Dear reviewer**

The current list of attributes of the meta data is inspired by the current properties of the Observation object and the GADF model. Please comment!


gammapy/data/metadata.py
```diff
@@ -0,0 +1,90 @@
...
+        the telescope/observatory name
+    instrument : str, optional
+        the specific instrument used
+    observation_mode : str, optional
+        the observation mode
+    location : `~astropy.coordinates.EarthLocation` or str, optional
+        the observatory location
+    deadtime : float
+        the observation deadtime. Default is 1.
+    muon_efficiency : float
```
bkhelifi, line 27: This quantity is really associated to HESS. Can we not add it later when loading the HESS data?
Also, for HESS, a useful quantity is TC

gammapy/data/metadata.py
```diff
@@ -0,0 +1,90 @@
...
+    telescope : str, optional
+        the telescope/observatory name
+    instrument : str, optional
+        the specific instrument used
+    observation_mode : str, optional
+        the observation mode
+    location : `~astropy.coordinates.EarthLocation` or str, optional
+        the observatory location
+    deadtime : float
+        the observation deadtime. Default is 1.
```
bkhelifi, line 26: This default means that there is no observation time. Is it what we want? 0 is better
I think that no default is wished, this quantity should be filled all the time...

gammapy/data/observations.py
```diff
@@ -92,8 +96,16 @@ def __init__(
         self._gti = gti
         self._events = events
         self._pointing = pointing
-        self._location = location
+        self._location = location  # this is part of the meta or is it data?
```
bkhelifi, line 99: For me, meta!

gammapy/data/observations.py
```diff
@@ -223,6 +235,15 @@ def create(
             location=location,
         )
 
+        meta = ObservationMetaData(
+            deadtime=1 - deadtime_fraction,
```
bkhelifi, line 239: The names are surprising.... spontaneously, with this subtraction we retrieve the livetime_fraction, and not the deadtime

gammapy/data/metadata.py
```diff
@@ -74,17 +76,69 @@ def validate_time(cls, v):
...
+            return SkyCoord(np.nan, np.nan, unit="deg", frame="icrs")
+        elif isinstance(v, SkyCoord):
+            return v
+        else:
+            raise ValidationError(
+                f"Incorrect position. Expect SkyCoord got {type(v)} instead."
+            )
+
     @classmethod
     def from_gadf_header(cls, events_hdr):
```
bkhelifi, line 91: Should we not start to add the version ID of a format? In order to deal with format evolution...

gammapy/data/observations.py
```diff
@@ -336,7 +336,7 @@ def obs_info(self):
     @property
     def pointing(self):
         if self._pointing is None:
-            self._pointing = FixedPointingInfo.from_fits_header(self.obs_info)
+            self._pointing = FixedPointingInfo.from_fits_header(self.events.table.meta)
```
bkhelifi, line 339: The pointing information should come logically from the data stored in the HDU pointing, no?

gammapy/data/metadata.py
registerrier: I have included this for backward compatibility because there is currently a muoneff property on `Observation`. It might make more sense to have it as part of optional entries since it is quite instrument specific.

Regarding the transparency coefficient one issue is that it does not correspond to any specific keyword in gadf.

gammapy/data/metadata.py
registerrier: Yes I followed the convention used in gadf  to store the deadtime correction factor. This might not be a good choice, especially if the attribute is called `deadtime`...

What do we prefer: storing the deadtime or the correction factor?

gammapy/data/observations.py
registerrier: OK. I tend to agree.
Currently there is a `location` argument in the `Observation` constructor (not mentioned in the docstring though). 
We can impose that `location` is passed in the helper `Observation.create` method and be stored in the `meta` there.

Opinions @bkhelifi , @maxnoe , @adonath ?

gammapy/data/metadata.py
registerrier: I suppose that in the long run we might have to introduce several such functions for different formats and subsequent version. I suppose we won't have to deal with a huge variety of formats.

But in general I would find it more convenient to have all I/O related functions in specific namespaces/classes. The factory method could find the correct format from the header itself.

In the meantime, we can rename this `ObservationMetaData.from_header(cls, events_hdr, format="gadf")`.

gammapy/data/observations.py
registerrier: If there is one yes, but so far we have only fixed pointing which is stored in the events header.

gammapy/data/metadata.py
maxnoe: Store livetime, start of observation and end?

gammapy/data/observations.py
maxnoe: How is pointing information relevant for the WCD experiments?

LauraOlivera: I think this is fine, for HAWC/SWGO we would just set the FixedPointingInfo to fixed_altaz (so in DRIFT mode). Would also be nice to not need a pointing HDU if the pointing is fixed anyway.

gammapy/data/metadata.py
```diff
@@ -0,0 +1,160 @@
...
+    optional : dict
+        additional optional metadata
+    """
+
+    telescope: Optional[str]
+    instrument: Optional[str]
+    observation_mode: Optional[str]
+    location: Optional[Union[str, EarthLocation, None]]
+    deadtime_fraction: float = Field(0.0, ge=0, le=1.0)
+    time_start: Optional[Union[Time, str, None]]
```
maxnoe, line 51: Optional[...] is a shorthand for Union[None, ...]

So either include None in the Union and remove optional or remove None

gammapy/data/metadata.py
```diff
@@ -0,0 +1,160 @@
...
+        if not format == "gadf":
+            raise ValueError(
+                f"Metadata creation from format {format} is not supported."
+            )
+
+        kwargs = {}
+        kwargs["telescope"] = events_hdr.get("TELESCOP")
+        kwargs["instrument"] = events_hdr.get("INSTRUME")
+        kwargs["observation_mode"] = events_hdr.get("OBS_MODE")
+        kwargs["deadtime_fraction"] = 1 - events_hdr.get("DEADC")
```
maxnoe, line 110: get returns None if the key is not found and then 1 - None wil raise an error

gammapy/data/metadata.py
registerrier: thanks.

PR #4740, opened 2023-09-04
Add DarkMatterDecaySpectralModel

### LINKED ISSUE #4739: Add DarkMatterDecaySpectralModel
One of our bachelor students (@JanaKonrad) implemented a `DarkMatterDecaySpectralModel` based on the already existing `DarkMatterAnnihilationSpectralModel`.
### This would basically just be a copy of the existing model with a few changes:

1. Change of the flux from 
https://github.com/gammapy/gammapy/blob/c71cb4707087c5278a15913b46828718562a8b10/gammapy/astro/darkmatter/spectra.py#L198-L207
to
```python
flux = ( 
     scale 
     * self.dfactor 
     / self.LIFETIME_AGE_OF_UNIVERSE 
     * self.primary_flux(energy=energy * (1 + self.z)) 
     / self.k
     / self.mass 
     / (4 * np.pi) 
 ) 
```

2. Init change from
https://github.com/gammapy/gammapy/blob/c71cb4707087c5278a15913b46828718562a8b10/gammapy/astro/darkmatter/spectra.py#L187
to
```python
def __init__(self, mass, channel, scale=scale.quantity, dfactor=1, z=0, k=1): 
```
3. `self.LIFETIME_AGE_OF_UNIVERSE = 10**17 * u.Unit("s")` roughly

### We also need to calculate the J-Factor for decay which is also called the D-Factor:
The only difference between them is a power of 2 for the provided density.
J-Factor: density**2 inside the l.o.s. integral
D-Factor: density inside the l.o.s. integreal

Unfortunately, this seems not be part of the `JFactory` but of the densities themselves:
https://github.com/gammapy/gammapy/blob/c71cb4707087c5278a15913b46828718562a8b10/gammapy/astro/darkmatter/profiles.py#L45-L64

I guess it would be smart to move the square to the actual integration happening in the `JFactory` or change the densities to either provide the integrated square or simple integration

StFroese:
**Description**
This PR fixes #4739 as proposed in the issue.
Add DarkMatterDecaySpectralModel

Ready for review

### DIFF
--- a/examples/tutorials/api/astro_dark_matter.py
+++ b/examples/tutorials/api/astro_dark_matter.py
@@ -37,6 +37,7 @@ import matplotlib.pyplot as plt
 from matplotlib.colors import LogNorm
 from gammapy.astro.darkmatter import (
     DarkMatterAnnihilationSpectralModel,
+    DarkMatterDecaySpectralModel,
     JFactory,
     PrimaryFlux,
     profiles,
@@ -84,7 +85,7 @@ print("DISTANCE_GC:", profiles.DMProfile.DISTANCE_GC)
 #
 # There are utilities to compute J-Factor maps that can serve as a basis
 # to compute J-Factors for certain regions. In the following we compute a
-# J-Factor map for the Galactic Centre region
+# J-Factor annihilation map for the Galactic Centre region
 #
 
 profile = profiles.NFWProfile(r_s=20 * u.kpc)
@@ -133,6 +134,46 @@ print(
 )
 
 ######################################################################
+# The J-Factor can also be computed for dark matter decay
+jfactory = JFactory(
+    geom=geom,
+    profile=profile,
+    distance=profiles.DMProfile.DISTANCE_GC,
+    annihilation=False,
+)
+jfact_decay = jfactory.compute_jfactor()
+
+jfact_map = WcsNDMap(geom=geom, data=jfact_decay.value, unit=jfact_decay.unit)
+plt.figure()
+ax = jfact_map.plot(cmap="viridis", norm=LogNorm(), add_cbar=True)
+plt.title(f"J-Factor [{jfact_map.unit}]")
+
+# 1 deg circle usually used in H.E.S.S. analyses without the +/- 0.3 deg band around the plane
+sky_reg = CircleSkyRegion(center=position, radius=1 * u.deg)
+pix_reg = sky_reg.to_pixel(wcs=geom.wcs)
+pix_reg.plot(ax=ax, facecolor="none", edgecolor="red", label="1 deg circle")
+
+sky_reg_rec = RectangleSkyRegion(center=position, height=0.6 * u.deg, width=2 * u.deg)
+pix_reg_rec = sky_reg_rec.to_pixel(wcs=geom.wcs)
+pix_reg_rec.plot(ax=ax, facecolor="none", edgecolor="orange", label="+/- 0.3 deg band")
+
+plt.legend()
+plt.show()
+
+total_jfact_decay = (
+    pix_reg.to_mask().multiply(jfact_decay).sum()
+    - pix_reg_rec.to_mask().multiply(jfact_decay).sum()
+)
+total_jfact_decay = (
+    pix_reg.to_mask().multiply(jfact_decay).sum()
+    - pix_reg_rec.to_mask().multiply(jfact_decay).sum()
+)
+print(
+    "J-factor in 1 deg circle without the +/- 0.3 deg band around GC assuming a "
+    f"{profile.__class__.__name__} is {total_jfact_decay:.3g}"
+)
+
+######################################################################
 # Gamma-ray spectra at production
 # -------------------------------
 #
@@ -168,7 +209,7 @@ plt.show()
 
 
 ######################################################################
-# Flux maps
+# Flux maps for annihilation
 # ---------
 #
 # Finally flux maps can be produced like this:
@@ -179,6 +220,30 @@ massDM = 10 * u.TeV
 diff_flux = DarkMatterAnnihilationSpectralModel(mass=massDM, channel=channel)
 int_flux = (
     jfact * diff_flux.integral(energy_min=0.1 * u.TeV, energy_max=10 * u.TeV)
+).to("cm-2 s-1")
+
+flux_map = WcsNDMap(geom=geom, data=int_flux.value, unit="cm-2 s-1")
+plt.figure()
+ax = flux_map.plot(cmap="viridis", norm=LogNorm(), add_cbar=True)
+plt.title(
+    f"Flux [{int_flux.unit}]\n m$_{{DM}}$={fluxes.mDM.to('TeV')}, channel={fluxes.channel}"
+)
+
+plt.show()
+
+
+######################################################################
+# Flux maps for decay
+# ---------
+#
+# Finally flux maps for decay can be produced like this:
+#
+
+channel = "Z"
+massDM = 10 * u.TeV
+diff_flux = DarkMatterDecaySpectralModel(mass=massDM, channel=channel)
+int_flux = (
+    jfact_decay * diff_flux.integral(energy_min=0.1 * u.TeV, energy_max=10 * u.TeV)
 ).to("cm-2 s-1")
 
 flux_map = WcsNDMap(geom=geom, data=int_flux.value, unit="cm-2 s-1")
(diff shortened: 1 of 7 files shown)


gammapy/astro/darkmatter/profiles.py
```diff
@@ -34,19 +34,22 @@ def scale_to_local_density(self):
         scale = (self.LOCAL_DENSITY / self(self.DISTANCE_GC)).to_value("")
         self.parameters["rho_s"].value *= scale
 
-    def _eval_squared(self, radius, separation):
-        """Squared density at given radius together with the substitution part"""
+    def _eval_substitution(self, radius, separation, squared):
+        """Density at given radius together with the substitution part"""
```
bkhelifi, line 38: It might be important to describe all the parameters in the docstring (and their type).

gammapy/astro/darkmatter/profiles.py
```diff
@@ -34,19 +34,22 @@ def scale_to_local_density(self):
...
-            self(radius) ** 2
+            self(radius) ** exponent
             * radius
             / np.sqrt(radius**2 - (self.DISTANCE_GC * np.sin(separation)) ** 2)
         )
 
-    def integral(self, rmin, rmax, separation, ndecade):
-        r"""Integrate squared dark matter profile numerically.
+    def integral(self, rmin, rmax, separation, ndecade, squared):
+        r"""Integrate dark matter profile numerically.
```
bkhelifi, line 47: The same here for the docstring... Generally speaking, improving the API documentation can help a lot the Gammapy non-experts

gammapy/astro/darkmatter/spectra.py
```diff
@@ -237,4 +237,106 @@ def from_dict(cls, data):
...
+        \frac{\Gamma}{4\pi m_{\mathrm{DM}}}
+        \frac{\mathrm d N}{\mathrm dE} \times J(\Delta\Omega)
+
+    Parameters
+    ----------
+    mass : `~astropy.units.Quantity`
+        Dark matter mass
+    channel : str
+        Annihilation channel for `~gammapy.astro.darkmatter.PrimaryFlux`
+    scale : float
```
bkhelifi, line 256: An example below of the scale might help to understand what is it exactly

gammapy/astro/darkmatter/spectra.py
```diff
@@ -237,4 +237,106 @@ def from_dict(cls, data):
...
+        massDM = 5000*u.Unit("GeV")
+        jfactor = 3.41e19 * u.Unit("GeV cm-2")
+        modelDM = DarkMatterDecaySpectralModel(mass=massDM, channel=channel, jfactor=jfactor)  # noqa: E501
+
+    References
+    ----------
+    * `2011JCAP...03..051 <https://ui.adsabs.harvard.edu/abs/2011JCAP...03..051C>`_
+    """
+
+    LIFETIME_AGE_OF_UNIVERSE = 4.3e17 * u.Unit("s")
```
bkhelifi, line 281: This quantity should be be configurable? Is there some science case that could lead to a change of it?

gammapy/astro/darkmatter/spectra.py
```diff
@@ -237,4 +237,106 @@ def from_dict(cls, data):
...
+    def __init__(self, mass, channel, scale=scale.quantity, jfactor=1, z=0):
+        self.z = z
+        self.mass = u.Quantity(mass)
+        self.channel = channel
+        self.jfactor = u.Quantity(jfactor)
+        self.primary_flux = PrimaryFlux(mass, channel=self.channel).table_model
+        super().__init__(scale=scale)
+
+    def evaluate(self, energy, scale):
+        """Evaluate dark matter decay model."""
```
bkhelifi, line 302: description of the parameters in the docstring...

StFroese:
@bkhelifi regarding the docstrings, should I go ahead and change the ones of annihilation model too in that case? Like I said in the issue, it's just a copy with a few changes here and there

Regarding the lifetime, I'm not sure. I also just did this analogous to the thermal relic cross section in the annihilation model

gammapy/astro/darkmatter/spectra.py
StFroese: Not sure but the decay width is the inverse of the lifetime and that means that when finding an upper limit for the width you can easily have upper limits being much smaller than 1e-17 (normally something like 1e-25). 
This results in the scale parameter being very small during the fitting procedure. Maybe it's numerically more stable to let the user choose a smaller lifetime in respect to having a larger scale parameter

gammapy/astro/darkmatter/profiles.py
StFroese: This already done a few lines down

bkhelifi:
Regarding to your first question, I would say yes... as it concerns the same file, a unique commit will be done. To be confirmed by @registerrier 

About `DEFAULT_LIFETIME', my question was more scientific than technical. As coded here, it is ok. It is then up to you, based on the science analyse use cas.

bkhelifi (dismissed):
From my side, it seems already OK, beyond the last comment.

registerrier:
Thanks @StFroese . 
I have the same comments here as in #4741 . Check the pre-commit hooks and don't forget to sign-off your commits. Thanks!

examples/tutorials/api/astro_dark_matter.py
```diff
@@ -189,3 +230,27 @@
...
+).to("cm-2 s-1")
+
+flux_map = WcsNDMap(geom=geom, data=int_flux.value, unit="cm-2 s-1")
+plt.figure()
+ax = flux_map.plot(cmap="viridis", norm=LogNorm(), add_cbar=True)
+plt.title(
+    f"Flux [{int_flux.unit}]\n m$_{{DM}}$={fluxes.mDM.to('TeV')}, channel={fluxes.channel}"
+)
+
+plt.show()
```
bkhelifi, line 256:
```suggestion
plt.show();
```

bkhelifi (approved):
Sound good for me, even if I am not an expert

bkhelifi (dismissed):

jdjuvsla:
Hi, 

I had a first look at the commit. Overall it looks good to me. There is one concern I want to raise: For the DM decay, the J-factor is typically called D-factor (as initially also stated in the issueL https://github.com/gammapy/gammapy/issues/4739) . So it might be a bit odd to use the Jfactory to implement the Dfactors. At the same time, it is just a naming issue and I don't know if this is a good reason for code duplication (I guess not). Did you discuss this point?

Best, 

Julia

StFroese:
Hi @jdjuvsla, I understand that concern but maybe we can leave a comment on that in the docstring and not duplicate the code

jdjuvsla:
Hi @StFroese,

this is fine with me as long as it is a conscious decision.  I don't think one gains too much from making the distinction between J and D. All I want to make sure is that people are aware of it and that the users don't get confused.

Best,
Julia

gammapy/astro/darkmatter/utils.py
```diff
@@ -47,17 +55,20 @@ def compute_differential_jfactor(self, ndecade=1e4):
...
                     np.arctan(_.value / self.distance.value),
                     ndecade,
+                    self.annihilation,
                 )
             )
             for _ in rmin.ravel()
         ]
-        jfact = u.Quantity(val).to("GeV2 cm-5").reshape(rmin.shape)
+        inegral_unit = u.Unit("GeV2 cm-5") if self.annihilation else u.Unit("GeV cm-2")
+        jfact = u.Quantity(val).to(inegral_unit).reshape(rmin.shape)
```
registerrier, line 71:
```suggestion
        integral_unit = u.Unit("GeV2 cm-5") if self.annihilation else u.Unit("GeV cm-2")
        jfact = u.Quantity(val).to(integral_unit).reshape(rmin.shape)
```

registerrier (approved):
Thanks @StFroese ! This looks nice.

I have left a few comments inline. I must say I am not completely clear regarding how to handle J/D Factor code duplication. Maybe have `compute_d_factor` and `compute_j_factor` in the `JFactory`?

In general, we should understand how this part can better connect to the rest of the modeling framework.

gammapy/astro/darkmatter/utils.py
bkhelifi: A naive question.... why in GeV?

StFroese: `GeV2 cm-5` is the standard unit for this. There's a package called clumpy (https://clumpy.gitlab.io/CLUMPY/v3.1.1/gallery.html) where you can see that too. 

Clumpy will get python binding soon. Maybe it would be of interest to also takes this into account when refactoring the dark matter stuff

registerrier (approved):
Thanks @StFroese . This is good to go.
 
Since #4741 was merged first, there is a merge conflict. Can you resolve it?

registerrier (approved):
Thanks @StFroese .

PR #4741, opened 2023-09-06
Add Zhao profile

StFroese:
**Description**
Adding the Zhao profile to the available dark matter profiles.

reference: https://ui.adsabs.harvard.edu/abs/1996MNRAS.278..488Z/abstract

### DIFF
--- a/gammapy/astro/darkmatter/profiles.py
+++ b/gammapy/astro/darkmatter/profiles.py
@@ -21,7 +21,7 @@ class DMProfile(abc.ABC):
     """DMProfile model base class."""
 
     LOCAL_DENSITY = 0.3 * u.GeV / (u.cm**3)
-    """Local dark matter density as given in refenrece 2"""
+    """Local dark matter density as given in reference 2"""
     DISTANCE_GC = 8.33 * u.kpc
     """Distance to the Galactic Center as given in reference 2"""
 
@@ -91,9 +91,10 @@ class DMProfile(abc.ABC):
 class ZhaoProfile(DMProfile):
     r"""Zhao Profile.
     This is taken from equation 1 from Zhao (1996). It is a generalization of the NFW profile. The volume density is parametrized with
-    a double power-law. Scale radii smaller than the scale radiu are described with a slope of :math:`-\gamma` and scale radii larger than the scale radius are described with a slope of :math:`-\beta`. :math:`\alpha` is a measure for the width of the transition region.
+    a double power-law. Scale radii smaller than the scale radius are described with a slope of :math:`-\gamma` and scale radii larger than the scale radius are described with a slope of :math:`-\beta`. :math:`\alpha` is a measure for the width of the transition region.
+
     .. math::
-        \rho(r) = \rho_s \left(\frac{r_s}{r}\right)^\gamma \left(1 + \left(\frac{r}{r_s}\right)^\frac{1}{\alpha} \right)^{(\gamma - \beta) * \alpha}
+        \rho(r) = \rho_s \left(\frac{r_s}{r}\right)^\gamma \left(1 + \left(\frac{r}{r_s}\right)^\frac{1}{\alpha} \right)^{(\gamma - \beta) \alpha}
 
     Parameters
     ----------
--- a/gammapy/astro/darkmatter/profiles.py
+++ b/gammapy/astro/darkmatter/profiles.py
@@ -91,7 +91,7 @@ class DMProfile(abc.ABC):
 class ZhaoProfile(DMProfile):
     r"""Zaho Profile.
     .. math::
-        \rho(r) = \rho_s \frac{r_s}{r}^\gamma \left(1 + \frac{r}{r_s}^\alpha \right)^{\frac{\beta - \gamma}{\alpha}}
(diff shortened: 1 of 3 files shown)


gammapy/astro/darkmatter/profiles.py
```diff
@@ -87,6 +88,59 @@ def integrate_spectrum_separation(self, func, xmin, xmax, separation, ndecade):
         return val.sum()
 
 
+class ZhaoProfile(DMProfile):
+    r"""Zaho Profile.
+    .. math::
+        \rho(r) = \rho_s \frac{r_s}{r}^\gamma \left(1 + \frac{r}{r_s}^\alpha \right)^{\frac{\beta - \gamma}{\alpha}}
```
bkhelifi, line 94: Have you check locally that the formula is displayed correctly?

bkhelifi (dismissed):
Thanks a lot @StFroese

gammapy/astro/darkmatter/profiles.py
StFroese: Yes it, but I think I should add some parentheses to the fraction of r/rs to make it more readable

StFroese: It compiled but the equation was wrong. Should be correct now

bkhelifi (dismissed):
Thanks @StFroese. I have not checked the implemented formula, otherwise all good for me

registerrier (commented):
Thanks @StFroese !

There is apparently an issue with the code style. Can you check that the pre-commit hooks are correctly set?
See the end of this section:
https://docs.gammapy.org/1.1/development/intro.html#get-set-up

registerrier:
Another thing would be to sign your commits according the following section:
https://docs.gammapy.org/1.1/development/intro.html#acceptation-of-the-developer-certificate-of-origin-dco

Thanks!

StFroese:
@registerrier both done

jdjuvsla:
Hi all! 

Quentin asked me to look at this and here are my comments:

1.) The reference in line 112 does not work for me (404) could you please check this?
2.) Which formula do you try to implement? Is it equation (1) from the first reference? I see some inconsistencies between line 94 and line 144 (and the mentioned equation). Please double check this.
3.) Since this model is not very widely known maybe it is worth to describe the parameters a bit more in lines 96ff?

Best regards,

Julia

StFroese:
Hi, @jdjuvsla :) Regarding your points:

1. It worked when I created this PR. This is really odd because this is exactly the same as the reference for the NFW profile which is already implemented. Maybe an ADS problem?
2. Yes it's (1). Actually it's all the same:
    - C is the normalisation constant, so $\rho_s$ in our case
    - Due to this normalisation $r$ has to be substituted with $r/r_s$
    - I substituted $\alpha=\frac{1}{\alpha^\prime}$ and called it $\alpha$ again
    - I wrote the equation in the docstring and in the function call different in order to be consistent with the equations you already wrote for other profiles like NFW
3. Can do :)

gammapy/astro/darkmatter/profiles.py
```diff
@@ -87,6 +88,62 @@ def integrate_spectrum_separation(self, func, xmin, xmax, separation, ndecade):
         return val.sum()
 
 
+class ZhaoProfile(DMProfile):
+    r"""Zaho Profile.
```
Astro-Kirsty, line 92: 'Zhao'

gammapy/astro/darkmatter/profiles.py
```diff
@@ -87,6 +88,62 @@ def integrate_spectrum_separation(self, func, xmin, xmax, separation, ndecade):
...
+    .. math::
```
Astro-Kirsty, line 93: Add a small description here, to say that which equation this utilises. i.e.
`Equation 1 from Zhao (1996)`

Astro-Kirsty (commented):
Hello, just a few comments regarding the docstring.

jdjuvsla:
Hi @StFroese ,

thanks for the clarifications! I could convince myself now that the implementation is correct. 
Concerning the second reference, I would prefer to only to push things that work, so it would be great if you would fix the link.
Concerning the formula in the docstring, I wonder why you rewrote the formula. Now we have 3 different formulations (eq. (1) from the reference, the docstring and the actual implementation). Is this really necessary or are we over complicating things? Especially the renaming of alpha took me by surprise and could be mentioned somewhere just to help people understand faster.

Best,
Julia

StFroese:
@jdjuvsla fixed the reference and the $\alpha$ for more consistency. Turns out that ADS changed the reference and it was missing a `C` in the string

bkhelifi:
Hello Stefan,
Can you run `tox -e codestyle` and update your code such that the CI passes the test? Thank you

StFroese:
@bkhelifi @jdjuvsla I just saw that the `DMProfile`s are not in the API documentation. Can we fix this somehow?

StFroese:
Ok, I actually don't like the solution since the profiles are no longer imported with `from gammapy.astro.darkmatter.profiles import ...`

registerrier:
Merging now. Thanks for the review @jdjuvsla

PR #4746, opened 2023-09-13
Raise error if the predicted event number is too large in event sampling

fabiopintore:
For a set of DC energy-dependent temporal models, the number of predicted events at some energies and time intervals may be very large (>>1e6). The reason is due to the creation methods of such models. 
The event sampling of them does not work since `MapDatasetEventSampler` raises a `ValueError` that is not very clear for the user (`Error: lam value is too large`) and does not help much in finding the issue. 
This PR introduces a clearer way to determine which model cannot be properly sampled and why.

### DIFF
--- a/gammapy/datasets/simulate.py
+++ b/gammapy/datasets/simulate.py
@@ -170,12 +170,12 @@ class MapDatasetEventSampler:
         npred = self._evaluate_timevar_source(dataset, model=model)
         data = npred.data[np.isfinite(npred.data)]
 
-        if np.sum(data) > 1e6:
-            raise TypeError(
+        try:
+            n_events = self.random_state.poisson(np.sum(data))
+        except ValueError:
+            raise ValueError(
                 f"The number of predicted events for the model {model.name} is too large. No event sampling will be performed for this model!"
             )
-
-        n_events = self.random_state.poisson(np.sum(data))
 
         coords = npred.sample_coord(n_events=n_events, random_state=self.random_state)
 
--- a/gammapy/datasets/simulate.py
+++ b/gammapy/datasets/simulate.py
@@ -169,6 +169,12 @@ class MapDatasetEventSampler:
 
         npred = self._evaluate_timevar_source(dataset, model=model)
         data = npred.data[np.isfinite(npred.data)]
+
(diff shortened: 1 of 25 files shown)


bkhelifi:
A naive question, hardcoding a threshold value is not a bit dangerous?
Otherwise, it might be good to precise this feature in the docstring as a small note or PS.

fabiopintore:
@bkhelifi , in general it can be dangerous. However, it might be quite impossible having a source whose predicted events are larger than the threshold of a million.

gammapy/datasets/simulate.py
```diff
@@ -169,6 +169,12 @@ def _sample_coord_time_energy(self, dataset, model):
 
         npred = self._evaluate_timevar_source(dataset, model=model)
         data = npred.data[np.isfinite(npred.data)]
+
+        if np.sum(data) > 1e6:
```
registerrier, line 173: 1e6 seems a very low value. I would have assumed this would work up to 1e16 or more. Are there really issues occurring for such low values?

gammapy/datasets/simulate.py
```diff
@@ -169,6 +169,12 @@ def _sample_coord_time_energy(self, dataset, model):
...
+            raise TypeError(
+                f"The number of predicted events for the model {model.name} is too large. No event sampling will be performed for this model!"
```
registerrier, line 175: Another approach is to try the poisson random and catch the exception to raise the one you want with its correct message.

registerrier (commented):
Thanks @fabiopintore .

My main comment is on the low threshold value you set. Can't you instead catch the exception from `random.poisson` and adapt the message?

fabiopintore:
Thanks @registerrier . I've implemented your suggestion. However, I do not know how to adapt the test: indeed, assuming a very large Npred value to be sampled, it implies that the test will be too slow. Any suggestions?

registerrier (commented):
Thanks @fabiopintore . Looks good.

Regarding the test, I would simply suggest to use something like `test_sample_coord_time_energy` except that you copy the temporal model and multiply it by e.g. 1e10 or more if needed to make the `poisson` call fail. It should not be long since the events won't be sampled.

Issue #4754, opened 2023-09-21
CI failing with matplotlib 3.8


registerrier:
Since the release of matplotlib 3.8 on September 14th, the CI fails at several places because of unit conversion errors. See e.g. https://github.com/gammapy/gammapy/actions/runs/6254270014/job/16981454002?pr=4753

registerrier:
The issue is caused by `pcolormesh` in `Background3D.plot_at_energy()`. There must be an issue in `astropy.visualization.quantity_support` and matplotlib 3.8.

A work around is to apply `pcolormesh` to the values and not the quantity.

registerrier:
This is solved by #4755 . closing.

PR #4755, opened 2023-09-21
Separate units in pcolormesh

registerrier:
**Description**

This pull request solves #4754 by applying `pcolormesh` on the values rather than the quantity object.

**Dear reviewer**


registerrier:
For compatibility this should be backported.

morcuended (approved):

registerrier:
During dev meeting, we agreed to keep a unit test to check for pcolormesh and Quantity. This will be in a future PR.

Issue #4732, opened 2023-08-29
Show default model parameters values into the docstring


QRemy:
> Can you also precise these default values into the docstring? Even with a deprecation message, the documentation should also provide this information.

None of the models show their default parameters in the docstring so this should be done in a dedicated PR.

_Originally posted by @QRemy in https://github.com/gammapy/gammapy/issues/4729#issuecomment-1695800786_

PR #4759, opened 2023-09-22
Display the default model parameters in docstrings

Astro-Kirsty:
Resolves issue [#4732](https://github.com/gammapy/gammapy/issues/4732)

### DIFF
--- a/gammapy/modeling/models/spectral.py
+++ b/gammapy/modeling/models/spectral.py
@@ -672,6 +672,7 @@ class ConstantSpectralModel(SpectralModel):
     ----------
     const : `~astropy.units.Quantity`
         :math:`k`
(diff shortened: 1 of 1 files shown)


bkhelifi (approved):
Thanks a lot @Astro-Kirsty for the effort.

QRemy (approved):
Thanks @Astro-Kirsty, do you want to add also the defaults for spatial and temporal models in this PR ? (we can also do than in a separate one)

Astro-Kirsty:
Thanks @bkhelifi and @QRemy.

Let's add a separate PR for the spatial and temporal (don't know I'll have time this week to implement those)

registerrier (approved):
Thanks @Astro-Kirsty . 

I wonder if there is a way to build the docstring programmatically for the parameter class attributes.

PR #4761, opened 2023-09-25
Regenerate name to avoid starting with underscore

QRemy:
Prevent a random fail in the CI when auto-generated name start with underscore :
FAILED ..\..\.tox\py310-test-alldeps_noray\lib\site-packages\gammapy\visualization\tests\test_datasets.py::test_plot_spectrum_datasets_off_regions - matplotlib._api.deprecation.MatplotlibDeprecationWarning: An artist whose label starts with an underscore was passed to legend(); such artists will no longer be ignored in the future.  To suppress this warning, explicitly filter out such artists, e.g. with `[art for art in artists if not art.get_label().startswith('_')]`.

This should fix the break seen in #4759.

### DIFF
--- a/gammapy/data/event_list.py
+++ b/gammapy/data/event_list.py
@@ -4,8 +4,7 @@ import copy
 import logging
 import numpy as np
 from astropy import units as u
-from astropy.coordinates import AltAz, Angle, SkyCoord
-from astropy.coordinates.angle_utilities import angular_separation
+from astropy.coordinates import AltAz, Angle, SkyCoord, angular_separation
 from astropy.io import fits
 from astropy.table import Table
 from astropy.table import vstack as vstack_tables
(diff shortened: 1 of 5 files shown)


bkhelifi:
The CI fails because of  `ModuleNotFoundError: No module named 'astropy.coordinates.angle_utilities`. But this is unrelated to this PR.

bkhelifi (approved):
Thanks @QRemy

QRemy:
> The CI fails because of `ModuleNotFoundError: No module named 'astropy.coordinates.angle_utilities`. But this is unrelated to this PR.

I tried to fix that direclty in this PR but I could open an other one instead.

registerrier (approved):

I think having the angular_utilities change in another PR would be better. But this can go as is.
This part of the PR can be backported too to ensure CI runs on bugfix branches with future versions.

Issue #4763, opened 2023-09-25
Show default model (spatial and temporal) parameter values in docstring


Astro-Kirsty:
Thanks @Astro-Kirsty, do you want to add also the defaults for spatial and temporal models in this PR ? (we can also do that in a separate one)

_Originally posted by @QRemy in https://github.com/gammapy/gammapy/pull/4759#pullrequestreview-1642499306_

PR #4769, opened 2023-09-26
Display default model parameters (temporal and spatial) in docstrings

Astro-Kirsty:
Resolves issue [#4763](https://github.com/gammapy/gammapy/issues/4763)

### DIFF
--- a/gammapy/modeling/models/spatial.py
+++ b/gammapy/modeling/models/spatial.py
@@ -510,7 +510,7 @@ class PointSpatialModel(SpatialModel):
     ----------
     lon_0, lat_0 : `~astropy.coordinates.Angle`
         Center position
-        Default is 0,0 deg
+        Default is "0 deg", "0 deg"
     frame : {"icrs", "galactic"}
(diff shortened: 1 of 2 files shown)


gammapy/modeling/models/spatial.py
```diff
@@ -510,6 +510,7 @@ class PointSpatialModel(SpatialModel):
     ----------
     lon_0, lat_0 : `~astropy.coordinates.Angle`
         Center position
+        Default is 0,0 deg
```
bkhelifi, line 513:
```suggestion
        Default is "0 deg" , "0 deg"
```

bkhelifi (approved):
Thanks a lot @Astro-Kirsty

gammapy/modeling/models/spatial.py
morcuended: minor question: are the quotes needed? As all the rest of the defaults go without them

QRemy (approved):

PR #4772, opened 2023-09-27
Fix plot spectrum function

SimoneMender:
Actual behavior:
The function `plot_spectrum() ` https://github.com/gammapy/gammapy/blob/4f7f73e493df78ceee59c9884e0ad4750e1fa97c/gammapy/irf/background.py#L398 calculates the integrated background rate and the output is in units of `deg / (MeV s sr)`.

Expected behavior: 
Calculate integrated background rate in units of `1 / (MeV s)`.

Fix:
- Change `cumsum()` function so that the Jacobian is applied when integrating over the offset axis. -> integrated background rate  is calculated in units of `deg^2 / (MeV s sr)`.
- Change `plot_spectrum()`: convert unit of the integrated bg rate to  `1 / (MeV s)`.

### DIFF
--- a/gammapy/irf/background.py
+++ b/gammapy/irf/background.py
@@ -411,7 +411,7 @@ class Background2D(BackgroundIRF):
         energy_axis = self.axes["energy"]
 
         bkg = self.integral(offset=offset_axis.bounds[1], axis_name="offset")
-        bkg = bkg.to(u.Unit("s-1")/energy_axis.unit)
+        bkg = bkg.to(u.Unit("s-1") / energy_axis.unit)
 
         with quantity_support():
             ax.plot(energy_axis.center, bkg, label="integrated spectrum", **kwargs)
--- a/gammapy/irf/background.py
+++ b/gammapy/irf/background.py
@@ -411,7 +411,7 @@ class Background2D(BackgroundIRF):
         energy_axis = self.axes["energy"]
 
         bkg = self.integral(offset=offset_axis.bounds[1], axis_name="offset")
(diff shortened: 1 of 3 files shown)


gammapy/irf/background.py
```diff
@@ -411,6 +411,7 @@ def plot_spectrum(self, ax=None, **kwargs):
         energy_axis = self.axes["energy"]
 
         bkg = self.integral(offset=offset_axis.bounds[1], axis_name="offset")
+        bkg = bkg.to(u.Unit("s-1 MeV-1"))
```
bkhelifi, line 414: @registerrier , should we hardcoded MeV? Why not TeV?

registerrier: As in the future we have to support more formats than gadf, one could consider that the energy axis unit of the bkg IRF is a good guess.
So `bkg.to(Unit("s-1")/energy_axis.unit)`  could work.

registerrier (commented):
Thanks @SimoneMender . This looks good. See inline comments.

The only missing thing is a test to make sure `offset` axis cumsum is correctly performed, as the plot test does not really check the values themselves.

registerrier (approved):
Thanks @SimoneMender . This looks good.!

registerrier:
I think this should be backported as it was already incorrect in v1.0.x

registerrier:
@meeseeksdev backport to v1.0.x

registerrier:
@meeseeksdev backport to v1.1.x

PR #4775, opened 2023-09-28
Fixing overwrite func in plot_distribution

MRegeard:
This PR is a very small fix for `plot_distribution`. 

Because `func` was overwritten, in case of map with more than one non-spatial bins, the code had not the same behaviour for the first bin and for the other. See pictures, where the second bin does not have the fit information. 
Capture d’écran 2023-09-28 à 16 35 50

I fix that by not overwritting `func` and putting the actual function to fit in `func_to_fit`.

### DIFF
--- a/gammapy/visualization/utils.py
+++ b/gammapy/visualization/utils.py
@@ -301,11 +301,11 @@ def plot_distribution(
 
             if func == "norm":
 
-                def func(x, mu, sigma):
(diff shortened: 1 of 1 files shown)


bkhelifi (approved):
Thanks for this fix!

bkhelifi (approved):
Thanks

registerrier (approved):
Thanks @MRegeard !

Issue #4707, opened 2023-07-31
Addition of EnergyDependenceEstimator


Astro-Kirsty:
Introducing a tool to investigate the potential energy-dependent morphology from a spatial map. This type of algorithm belongs in the `Estimator` class. 
As we know, it is possible for gamma-ray sources to exhibit energy-dependent morphology defined as the spatial morphology changing in different energy bins. Many different techniques of measuring this energy-dependence have been utilised over the years. The idea of this estimator is to provide a generic initial look at the energy-dependence for a gamma-ray source. 

**Methodology**
The technique chosen here is to fit a spatial model and check the difference in extension and position in different energy bins. Firstly, the user must define an initial fit for the `SkyModel` for their source of interest, along with energy bands they are interested in. Wilks theorem is utilised, in which the initial hypothesis (H0) is a joint fit where the entire energy range is fit with the spatial model of choice. The alternative hypothesis (H1) is an individual fit of the chosen spatial model in each energy band. This allows the `delta_ts` to be calculated and converted to a significance value using the degrees of freedom.  In both cases the spectral parameters are frozen except the amplitude. 

The user can define which spatial parameters to keep free, in the examples to be presented however, a `GaussianSpatialModel` will be utilised with the position and sigma extension free, whilst other parameters are frozen. 

The significance of the results can be expressed as both the `delta_ts` value from Wilks theorem, or as the chi-square values for each parameter.

Once the error extension function is applied to the `SpatialModels` it will be possible to plot each of the `SpatialModels` in each energy band. 

**Extra information**
For this feature we will require an `Estimator` to be added, a tutorial on how to use the `Estimator`, documentation and a recipe which shows the validation of the results. 
The `Estimator` and tutorial have been created and a PR will be made.

PR #4783, opened 2023-10-05
Add EnergyDependentMorphologyEstimator

Astro-Kirsty:
Add the `EnergyDependentEstimator` to resolve [#4707](https://github.com/gammapy/gammapy/issues/4707)

This is an updated PR to include all recent upgrade. Old PR [here](https://github.com/gammapy/gammapy/pull/4708)

**Description**
An example is included for MSH 1552 to show how the estimator works: 
https://gist.github.com/Astro-Kirsty/f74ee75ca543bd8f763cb9225d3b84f8

I've added here a notebook showing the validation for the method/results:
https://gist.github.com/Astro-Kirsty/33efa8847e5f756fbe214f7d12f54bfc

Documentation and a recipe for the validation of results need to be added -- but in a separate PR

### DIFF
--- a/examples/tutorials/analysis-3d/energy_dependence_estimation.py
+++ b/examples/tutorials/analysis-3d/energy_dependence_estimation.py
@@ -201,24 +201,24 @@ estimator = EnergyDependenceEstimator(energy_edges=energy_edges, source="MSH1552
 # The results of the source signal above the background in each energy bin
 # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 #
-# The table shows the ∆(TS) value, the number of degrees of freedom (df)
+# Firstly, the estimator is run to produce the results.
+# The table here shows the ∆(TS) value, the number of degrees of freedom (df)
 # and the significance (σ) in each energy bin. The significance values here show that each
 # energy band has significant signal above the background.
 #
 
-result_bkg_src = estimator.estimate_source_significance(datasets)
-table_bkg_src = Table(result_bkg_src)
+results = estimator.run(datasets)
+table_bkg_src = Table(results["src_above_bkg"])
 display(table_bkg_src)
 
 ######################################################################
 # The results for testing energy dependence
 # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 #
-# Next, the estimator is run to produce the results of the energy-dependent estimator.
+# Next, the results of the energy-dependent estimator are shown.
 # The table shows the various free parameters for the joint fit for :math:`H_0` across the entire
 # energy band and for each energy bin shown for :math:`H_1`.
 
-results = estimator.run(datasets)
 ts = results["energy_dependence"]["delta_ts"]
 df = results["energy_dependence"]["df"]
 sigma = ts_to_sigma(ts, df=df)
--- a/examples/tutorials/analysis-3d/energy_dependence_estimation.py
+++ b/examples/tutorials/analysis-3d/energy_dependence_estimation.py
@@ -201,8 +201,8 @@ estimator = EnergyDependenceEstimator(energy_edges=energy_edges, source="MSH1552
 # The results of the source signal above the background in each energy bin
 # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 #
-# The table shows the delta TS value, the number of degrees of freedom (df)
-# and the significance in each energy bin. The significance values here show that each
+# The table shows the ∆(TS) value, the number of degrees of freedom (df)
+# and the significance (σ) in each energy bin. The significance values here show that each
 # energy band has significant signal above the background.
 #
 
@@ -213,6 +213,10 @@ display(table_bkg_src)
 ######################################################################
 # The results for testing energy dependence
 # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+#
+# Next, the estimator is run to produce the results of the energy-dependent estimator.
+# The table shows the various free parameters for the joint fit for :math:`H_0` across the entire
+# energy band and for each energy bin shown for :math:`H_1`.
 
 results = estimator.run(datasets)
 ts = results["energy_dependence"]["delta_ts"]
@@ -230,7 +234,9 @@ display(results_table)
 # The chi-squared value for each parameter of interest
 # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 #
-# The sigma values.
+# We can also utilise the `~gammapy.estimators.energydependence.weighted_chi2_parameter` function for each parameter.
+#
+# The weighted chi-squared significance for the sigma values.
 #
 
 display(
@@ -242,7 +248,7 @@ display(
 )
 
 ######################################################################
-# The lat_0 values.
+# The weighted chi-squared significance for the lat_0 values.
 
 display(
     Table(
@@ -253,7 +259,7 @@ display(
 )
 
 ######################################################################
-# The lon_0 values.
+# The weighted chi-squared significance for the lon_0 values.
 
 display(
     Table(
--- a/examples/tutorials/analysis-3d/energy_dependence_estimation.py
+++ b/examples/tutorials/analysis-3d/energy_dependence_estimation.py
@@ -178,7 +178,7 @@ model.spatial_model.sigma.frozen = False
 model.spectral_model.amplitude.frozen = False
 model.spectral_model.index.frozen = True
 
-datasets.model = model
+datasets.models = model
 
 estimator = EnergyDependenceEstimator(energy_edges=energy_edges, source="MSH1552")
 
@@ -190,9 +190,7 @@ estimator = EnergyDependenceEstimator(energy_edges=energy_edges, source="MSH1552
 # -----------------------------------------------
 #
 
-estimator = EnergyDependenceEstimator(
-    energy_edges=energy_edges, source=source, selection_optional=["src-sig"]
-)
+estimator = EnergyDependenceEstimator(energy_edges=energy_edges, source="MSH1552")
 table_bkg_src = estimator.estimate_source_significance(datasets)
 table_bkg_src
 
--- a/examples/tutorials/analysis-3d/energy_dependence_estimation.py
+++ b/examples/tutorials/analysis-3d/energy_dependence_estimation.py
@@ -35,8 +35,11 @@ from astropy.coordinates import SkyCoord
 from astropy.table import Table
 from regions import CircleSkyRegion
 from gammapy.data import DataStore
-from gammapy.datasets import Datasets, MapDataset
-from gammapy.estimators.energydependence import EnergyDependenceEstimator
+from gammapy.datasets import MapDataset
+from gammapy.estimators.energydependence import (
+    EnergyDependenceEstimator,
+    weighted_chi2_parameter,
+)
 from gammapy.makers import (
     DatasetsMaker,
     FoVBackgroundMaker,
@@ -186,13 +189,13 @@ estimator = EnergyDependenceEstimator(energy_edges=energy_edges, source="MSH1552
 # Show the results tables
 # -----------------------
 #
-# The results for the source above the background
+# The results of the source signal above the background in each energy bin.
 # -----------------------------------------------
 #
 
 estimator = EnergyDependenceEstimator(energy_edges=energy_edges, source="MSH1552")
-table_bkg_src = estimator.estimate_source_significance(datasets)
-table_bkg_src
+result_bkg_src = estimator.estimate_source_significance(datasets)
+Table(result_bkg_src)
 
 ######################################################################
 # The results for testing energy dependence
@@ -201,9 +204,28 @@ table_bkg_src
 results = estimator.run(datasets)
 ts = results["energy_dependence"]["delta_ts"]
 df = results["energy_dependence"]["df"]
-print(f"The significance value is {ts_to_sigma(ts, df=df)}")
+sigma = ts_to_sigma(ts, df=df)
+print(f"The delta_ts for the energy-dependent study is {ts:.3f}.")
+print("")
+print(f"Converting this to a significance gives {sigma:.3f} \u03C3")
 table = Table(results["energy_dependence"]["result"])
 table
+
+# The chi-squared value for each parameter of interest
+# ----------------------------------------------------
+
+Table(
+    weighted_chi2_parameter(results["energy_dependence"]["result"], parameter="sigma")
+)
+
+Table(
+    weighted_chi2_parameter(results["energy_dependence"]["result"], parameter="lat_0")
+)
+
+Table(
+    weighted_chi2_parameter(results["energy_dependence"]["result"], parameter="lon_0")
+)
+
 
 ######################################################################
 # Plotting the results
--- a/examples/tutorials/analysis-3d/energy_dependence_estimation.py
+++ b/examples/tutorials/analysis-3d/energy_dependence_estimation.py
@@ -1,0 +1,246 @@
+"""
+Energy Dependence Estimation
+============================
+
+Learn how to test for energy dependent morphology in your dataset.
+
+Prerequisites
+-------------
+Knowledge on data reduction and datasets used in Gammapy, for example see
+the :doc:`/tutorials/data/hess` and :doc:`/tutorials/analysis-2d/ring_background` tutorials.
+
+
+Context
+-------
+
+A tool to investigate the potential of energy-dependent morphology from spatial maps. This tutorial consists
+of two main steps.
+
+Firstly, the user defines both the initial SkyModel as determined through previous investigate along with the
+energy-band of interest to test for energy dependence. The null hypothesis (H0) is defined as only the background
+component being free (norm). The alternative hypothesis (H1) is adding the source model back in. The results of this
+first step show the significance of your source above the background in each energy band.
+
+The second step is for the purpose of quantifying any energy dependent morphology. The null hypothesis here is defined
+through a joint fit of parameters. Giving a base for the analysis. The alternative hypothesis is where the free
+parameters of the model are fit individually in each energy band.
+
+Setup
+-----
+
+"""
+
+from astropy import units as u
+from astropy.coordinates import SkyCoord
+from astropy.table import Table
+from regions import CircleSkyRegion
+from gammapy.data import DataStore
+from gammapy.datasets import Datasets, MapDataset
+from gammapy.estimators.energydependence import EnergyDependenceEstimator
+from gammapy.makers import (
+    DatasetsMaker,
+    FoVBackgroundMaker,
+    MapDatasetMaker,
+    SafeMaskMaker,
+)
+from gammapy.maps import MapAxis, WcsGeom
+from gammapy.modeling.models import (
+    GaussianSpatialModel,
+    PowerLawSpectralModel,
+    SkyModel,
+)
+from gammapy.stats.utils import ts_to_sigma
+
+######################################################################
+# Check setup
+# -----------
+
+# from gammapy.utils.check import check_tutorials_setup
+# check_tutorials_setup()
+
+######################################################################
+# Obtain the data to use
+# ----------------------
+#
+# Create the data store and obtain the observations from the HESS DL3 DR1 for MSH 1552.
+#
+# P.S.: do not forget to setup your environment variable $GAMMAPY_DATA to your local directory
+
+data_store = DataStore.from_dir("$GAMMAPY_DATA/hess-dl3-dr1")
+obs_id = data_store.obs_table["OBS_ID"][data_store.obs_table["OBJECT"] == "MSH 15-5-02"]
+observations = data_store.get_observations(obs_id)
+
+######################################################################
+# Setting the exclusion mask
+# --------------------------
+#
+# First, we define the energy range to obtain the dataset within. The geometry is also
+# defined, based on the position of MSH 1552 (the source of interest here).
+
+energy_axis = MapAxis.from_energy_bounds(0.2, 100, nbin=15, unit="TeV")
+energy_axis_true = MapAxis.from_energy_bounds(
+    0.05, 110, nbin=30, unit="TeV", name="energy_true"
+)
+
+source_pos = SkyCoord(320.33, -1.19, unit="deg", frame="galactic")
+geom = WcsGeom.create(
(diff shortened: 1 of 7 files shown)


Astro-Kirsty:
Did not include an example a general fit to determine the overall parameters for the model (see [here](https://github.com/gammapy/gammapy/pull/4708#discussion_r1334007550)) as I believe the user should already have a handle on doing this before they try to use the estimator. 

[This](https://github.com/gammapy/gammapy/pull/4708#discussion_r1347230550) also need implementation or discussion.

I will add tests soon

bkhelifi:
Once this PR is merged, indeed a tutorial would be needed... or insert this functionality within an existing tutorial

gammapy/estimators/energydependence.py
```diff
@@ -0,0 +1,245 @@
...
+from gammapy.datasets import Datasets
+from gammapy.modeling import Fit
+from gammapy.modeling.models import FoVBackgroundModel
+from gammapy.modeling.selection import TestStatisticNested
+from gammapy.stats.utils import ts_to_sigma
+from .core import Estimator
+
+
+def weighted_chi2_parameter_results(table_edep, parameter="sigma"):
+    """Calculate the weighted chi2 value for the parameter of interest"""
```
bkhelifi, line 11: Can you define in the docstring the parameters?

gammapy/estimators/energydependence.py
```diff
@@ -0,0 +1,245 @@
...
+    """Test if there is any energy-dependent morphology in a map dataset for a given set of energy bins.
+
+    Parameters
+    ----------
+    energy_edges : `~astropy.units.Quantity`
+        Energy edges for the energy-dependence test.
+    source : str or int
+        For which source in the model to compute the estimator.
+    fit : `Fit`
+        Fit instance specifying the backend and fit options.
```
bkhelifi, line 41: In the previous PR, we were speaking about adding in the docstring the list of available fitting backend for all estimators. Can you put an issue to remind that?

bkhelifi, line 41: You could precise that when None iminuit is used

gammapy/estimators/energydependence.py
```diff
@@ -0,0 +1,245 @@
...
+            Table with the results of the null hypothesis with no source, and alternative
+            hypothesis with the source added in. Entries are:
+            * "Emin" : the minimum energy of the energy band
+            * "Emax" : the maximum energy of the energy band
+            * "delta_ts" : difference in ts
+            * "df" : the degrees of freedom between null and alternative hypothesis
+            * "significance" : significance of the result
+        """
+        for dataset in datasets:
+            dataset.mask_fit = dataset.counts.geom.energy_mask(
```
bkhelifi, line 78: here, the initial mask of the dataset is erased. Is it the wished behaviour? or an AND is preferable?

gammapy/estimators/energydependence.py
```diff
@@ -0,0 +1,245 @@
...
+        Returns
+        -------
+        results : `dict`
+            Dict with results of the energy-dependence test.
+            * "delta_ts" : difference in ts between fitting each energy band individually (sliced fit) and the joint fit
+            * "df" : the degrees of freedom between fitting each energy band individually (sliced fit) and the joint fit
+            * "result" : the results for the fitting each energy band individually (sliced fit) and the joint fit
+        """
+        for dataset in datasets:
+            dataset.mask_fit = dataset.counts.geom.energy_mask(
```
bkhelifi, line 146: The same remark as above..

bkhelifi:
I am wondering if some visualization utils of the results are needed. Any thoughts?

gammapy/estimators/energydependence.py
registerrier: I don't think this is a good idea to duplicate the list of available backends in the documentation. This will become impossible to maintain.
What should be done is to put a proper link to `~gammapy.modeling.Fit` in the argument that will directly link to the list of available backends.

Astro-Kirsty:
Thanks @bkhelifi. We have the `SpatialModel` plotting implemented, however, we could also add other visualisation i.e. parameter vs energy plot

Astro-Kirsty:
I've added some tests, however these utilise the data in the example. Not sure if there is another dataset that can be used or if it would be useful to save the MSH 15-52 dataset in gammapy data. 
Suggestions are welcome

adonath:
@Astro-Kirsty Ideally you use a small (simulated) toy dataset with a source that has energy depend morphology to check whether you can approximately reconstruct the true values. Here is in example of how to define such a model: https://docs.gammapy.org/1.1/tutorials/api/models.html#models-with-energy-dependent-morphology

Astro-Kirsty:
Thanks @adonath, I already have a dataset for this so I will adapt the test accordingly

gammapy/estimators/energydependence.py
Astro-Kirsty: This was the wished behaviour, however, I am not sure whether it is preferable

examples/tutorials/analysis-3d/energy_dependence_estimation.py
```diff
@@ -0,0 +1,266 @@
...
+
+safe_mask_maker = SafeMaskMaker(
+    methods=["aeff-default", "offset-max"], offset_max=2.5 * u.deg
+)
+
+dataset_maker = MapDatasetMaker()
+
+fov_bkg_maker = FoVBackgroundMaker(method="fit", exclusion_mask=exclusion_mask)
+
+global_dataset = MapDataset.create(geom)
```
MRegeard, line 120: Don't you need to put the energy_axis_true here ?

```suggestion
global_dataset = MapDataset.create(geom, energy_axis_true=energy_axis_true)
```

examples/tutorials/analysis-3d/energy_dependence_estimation.py
```diff
@@ -0,0 +1,266 @@
...
+
+######################################################################
+# Show the results tables
+# -----------------------
+#
+# The results of the source signal above the background in each energy bin.
+# -----------------------------------------------
+#
+
+estimator = EnergyDependenceEstimator(energy_edges=energy_edges, source="MSH1552")
```
MRegeard, line 196: This is not needed, either this one or the one above should be remove.

examples/tutorials/analysis-3d/energy_dependence_estimation.py
```diff
@@ -0,0 +1,266 @@
...
+result_bkg_src = estimator.estimate_source_significance(datasets)
+Table(result_bkg_src)
+
+######################################################################
+# The results for testing energy dependence
+# -----------------------------------------
+
```
MRegeard, line 203: Maybe a little bit of context is needed here. Especially, what is `df`?

examples/tutorials/analysis-3d/energy_dependence_estimation.py
```diff
@@ -0,0 +1,266 @@
...
+print("")
+print(f"Converting this to a significance gives {sigma:.3f} \u03C3")
+table = Table(results["energy_dependence"]["result"])
+table
+
+# The chi-squared value for each parameter of interest
+# ----------------------------------------------------
+
+Table(
+    weighted_chi2_parameter(results["energy_dependence"]["result"], parameter="sigma")
```
MRegeard, line 218: Why not using `table` here instead of `results["energy_dependence"]["result"]`?

examples/tutorials/analysis-3d/energy_dependence_estimation.py
```diff
@@ -0,0 +1,266 @@
...
+# -----------------------------------------
+
+results = estimator.run(datasets)
+ts = results["energy_dependence"]["delta_ts"]
+df = results["energy_dependence"]["df"]
+sigma = ts_to_sigma(ts, df=df)
+print(f"The delta_ts for the energy-dependent study is {ts:.3f}.")
+print("")
+print(f"Converting this to a significance gives {sigma:.3f} \u03C3")
+table = Table(results["energy_dependence"]["result"])
```
MRegeard, line 211: If `table` is used below (see next comment), maybe it would be wise to renamed it to a more specific name, i.e. `results_table`.

MRegeard (commented):
Thanks @Astro-Kirsty, this is a very nice and useful feature !
See inline comments.

gammapy/estimators/energydependence.py
```diff
@@ -0,0 +1,262 @@
...
+from gammapy.stats.utils import ts_to_sigma
+from .core import Estimator
+
+
+def weighted_chi2_parameter(results_edep, parameter="sigma"):
+    """Calculate the weighted chi2 value for the parameter of interest
+
+    Parameters
+    ----------
+    result_edep : dict
```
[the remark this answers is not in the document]
Astro-Kirsty: The idea was to return a dict, similar to how it is done in the `~gammapy.modeling.select_nested_models`.
Open to other ideas, because this was a long debated topic. What are the most important values and how to show them!

Astro-Kirsty:
Note: need to wait for #4927 to be merged, so that the test can be fixed. 
This PR will also need a rebase - but I will wait until other reviews are received and we are ready for merge.

registerrier (commented):
Thanks @Astro-Kirsty . Here are some further (but old) comments.

The CI is broken because of the angular_separation import with astropy 6

gammapy/estimators/tests/test_energydependence.py
```diff
@@ -0,0 +1,202 @@
+import numpy as np
+from numpy.testing import assert_allclose
+from astropy import units as u
+from astropy.coordinates import SkyCoord
+from astropy.coordinates.angle_utilities import angular_separation
```
registerrier, line 5:
```suggestion
from astropy.coordinates import angular_separation
```

gammapy/estimators/energydependence.py
```diff
@@ -0,0 +1,264 @@
...
+
+        # Prepare results dictionary
+        joint_values = [result_joint.parameters[param].value for param in parameters]
+        joint_errors = [result_joint.parameters[param].error for param in parameters]
+
+        parameter_values = np.empty((len(parameters), self.num_energy_bands))
+        parameter_errors = np.empty((len(parameters), self.num_energy_bands))
+        for i in range(self.num_energy_bands):
+            parameter_values[:, i] = [
+                results_src[i].parameters[param].value for param in parameters
```
registerrier, line 217: Is it always working? Shouldn't you rather access `results[i].models[self.source].free_parameters`?

registerrier: You could use `Parameters.value` to directly access all parameters values.

gammapy/estimators/tests/test_energydependence.py
```diff
@@ -0,0 +1,201 @@
...
+
+        return norm * np.exp(exponent)
+
+    @property
+    def evaluation_radius(self):
+        """Evaluation radius (`~astropy.coordinates.Angle`)."""
+        return 2 * np.max([self.sigma_1TeV.value, self.sigma_10TeV.value]) * u.deg
+
+
+def create_dataset():
```
registerrier, line 71: you could make this a fixture

gammapy/estimators/tests/test_energydependence.py
```diff
@@ -0,0 +1,201 @@
...
+    dataset = maker_safe_mask.run(dataset, obs)
+
+    # Add the model on the dataset and Poission fluctuate
+    dataset.models = model
+    dataset.fake(random_state=42)
+
+    return dataset
+
+
+source_pos = SkyCoord(5.58, 0.2, unit="deg", frame="galactic")
```
registerrier, line 127: You should not leave test code outside functions or class.

You have two options:
- create a fixture/function to create the model and then in each test function call the `create_model` and `create_dataset` functions
- Create a general Test class that will perform this code on init and then have test methods for the various things required.

registerrier (commented):
Thanks @Astro-Kirsty . I have left some inline comments. The main one concerns the test whic should be structured differently. You shouldn't have code executed outside functions or class. You could restructure it to go into a test class.

gammapy/estimators/energydependence.py
```diff
@@ -0,0 +1,264 @@
...
+            * "delta_ts" : difference in ts
+            * "df" : the degrees of freedom between null and alternative hypothesis
+            * "significance" : significance of the result
+        """
+        slices_src = self._slice_datasets(datasets)
+
+        # Norm is free and fit
+        test_results = []
+        for sliced in slices_src:
+            parameters = [param for param in sliced.models.parameters.free_parameters]
```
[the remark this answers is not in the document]
Astro-Kirsty: This will not work because we are working with sliced datasets that do not have the name `source`.
See line 100

gammapy/estimators/energydependence.py
```diff
@@ -0,0 +1,264 @@
...
+            null_values = [0] + [
+                param.value
+                for param in sliced.models[0].spatial_model.parameters.free_parameters
```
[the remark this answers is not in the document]
Astro-Kirsty: We need to know access specifically the `spatial_model`, so it should not be `parameters`

gammapy/estimators/energydependence.py
```diff
@@ -0,0 +1,264 @@
...
+            Sliced datasets.
+        """
+        model = datasets.models[self.source]
+
+        slices_src = Datasets()
+        for emin, emax in zip(self.energy_edges[:-1], self.energy_edges[1:]):
+            for dataset in datasets:
+                sliced_src = dataset.slice_by_energy(emin, emax)
+                bkg_sliced_model = FoVBackgroundModel(dataset_name=sliced_src.name)
+                sliced_src.models = [model.copy(), bkg_sliced_model]
```
registerrier, line 101: But what if there are more models set on the dataset?

gammapy/estimators/energydependence.py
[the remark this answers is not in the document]
Astro-Kirsty: Also, same as above the names are no longer so we take `sliced.models[0]` which is the `SkyModel` rather than the `FoVBackgroundModel`

gammapy/estimators/energydependence.py
```diff
@@ -0,0 +1,264 @@
...
+        )
+        df_src = free_x * free_y
+
+        # Calculate the joint fit
+        parameters = model.spatial_model.parameters.free_parameters.names
+        slice0 = slices_src[0]
+        for slice_j in slices_src[1:]:
+            for param in parameters:
+                setattr(
+                    slice_j.models[0].spatial_model,
```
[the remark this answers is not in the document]
Astro-Kirsty: Same as above!

gammapy/estimators/energydependence.py
Astro-Kirsty: Currently it is implemented through the parameters variable on line 193, such that the results will only include the spatial model!

examples/tutorials/analysis-3d/energy_dependence_estimation.py
```diff
@@ -0,0 +1,310 @@
...
+the :doc:`/tutorials/data/hess` and :doc:`/tutorials/analysis-2d/ring_background` tutorials.
+
+
+Context
+-------
+
+A tool to investigate the potential of energy-dependent morphology from spatial maps. This tutorial consists
+of two main steps.
+
+Firstly, the user defines the initial `~gammapy.modeling.models.SkyModel` based on previous investigations
```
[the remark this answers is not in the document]
Astro-Kirsty: Okay, should I move it somewhere else or remove it?

gammapy/estimators/energydependence.py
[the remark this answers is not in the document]
Astro-Kirsty: I cannot seem to get this to work with the naming, I believe because there are multiple parameters looped over and the names would be the same?

examples/tutorials/analysis-3d/energy_dependence_estimation.py
```diff
@@ -0,0 +1,310 @@
...
+    )
+)
+
+######################################################################
+# The weighted chi-squared significance for the ``lon_0`` values.
+
+display(
+    Table(
+        weighted_chi2_parameter(
+            results["energy_dependence"]["result"], parameter="lon_0"
```
AtreyeeS, line 267: Do you need 3 Tables with one row each? Showing all the parameters in a Table simultaneously will be better

gammapy/estimators/energydependence.py
```diff
@@ -0,0 +1,275 @@
...
+    tag = "EnergyDependenceEstimator"
+
+    def __init__(self, energy_edges, source=0, fit=None):
+
+        self.energy_edges = energy_edges
+        self.source = source
+        self.num_energy_bands = len(self.energy_edges) - 1
+
+        if fit is None:
+            fit = Fit(optimize_opts={"print_level": 1})
```
AtreyeeS, line 77: Why configure `print_level`?

gammapy/estimators/energydependence.py
```diff
@@ -0,0 +1,275 @@
...
+        for i, (emin, emax) in enumerate(
+            zip(self.energy_edges[:-1], self.energy_edges[1:])
+        ):
+            for dataset in datasets:
+                sliced_src = dataset.slice_by_energy(
+                    emin, emax, name=f"{self.source}_{i}"
+                )
+                bkg_sliced_model = FoVBackgroundModel(dataset_name=sliced_src.name)
+                sliced_src.models = [
+                    model.copy(name=f"{sliced_src.name}"),
```
AtreyeeS, line 106: This will remove the other `models` from the fit. Is that what you want? 
One might want to search for source extension in the presence of other sources (eg, diffuse emission)

gammapy/estimators/energydependence.py
```diff
@@ -0,0 +1,275 @@
...
+            Sliced datasets.
+        """
+        model = datasets.models[self.source]
+
+        slices_src = Datasets()
+        for i, (emin, emax) in enumerate(
+            zip(self.energy_edges[:-1], self.energy_edges[1:])
+        ):
+            for dataset in datasets:
+                sliced_src = dataset.slice_by_energy(
```
AtreyeeS, line 101: How is the `mask_fit` for `sliced_src` handled?

AtreyeeS (commented):
Thanks @Astro-Kirsty ! Sorry for the delay, I had forgotten to finish a review I had started long back.

Please see inline comments for some specific questions.

In general, I think it can be useful if you add a dataset with an energy dependent model in `gammapy-data`. A major part of the test is defining a model, doing a simulation etc. Same for the tutorial, where the data reduction takes up a significant part.

gammapy/estimators/energydependence.py
```diff
@@ -0,0 +1,262 @@
...
+from astropy.table import Table
+from gammapy.datasets import Datasets
+from gammapy.modeling import Fit
+from gammapy.modeling.models import FoVBackgroundModel
+from gammapy.modeling.selection import TestStatisticNested
+from gammapy.stats.utils import ts_to_sigma
+from .core import Estimator
+
+
+def weighted_chi2_parameter(results_edep, parameter="sigma"):
```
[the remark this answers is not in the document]
Astro-Kirsty: The idea was that this is an additional calculation that could be made. 
One should really utilise the results_edep significance values because these include any correlation between parameters.

gammapy/estimators/energydependence.py
Astro-Kirsty: I believe this comment is outdated? There is no `mask_fit` defined within the code now?

gammapy/estimators/energydependence.py
```diff
@@ -0,0 +1,262 @@
...
+        result_bkg_src : dict
+            Dictionary with the results of the null hypothesis with no source, and alternative
+            hypothesis with the source added in. Entries are:
+            * "Emin" : the minimum energy of the energy band
+            * "Emax" : the maximum energy of the energy band
+            * "delta_ts" : difference in ts
+            * "df" : the degrees of freedom between null and alternative hypothesis
+            * "significance" : significance of the result
+        """
+        for dataset in datasets:
```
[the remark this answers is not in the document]
Astro-Kirsty: This comment is possibly outdated?

examples/tutorials/analysis-3d/energy_dependence_estimation.py
Astro-Kirsty: The way the `weighted_chi2_parameter` function is created means that one table is created for a given parameter. I will adjust the implementation a bit!

Astro-Kirsty:
Thanks for the comments @AtreyeeS.

> In general, I think it can be useful if you add a dataset with an energy dependent model in gammapy-data. 

I agree, it would certainly speed up the tutorial also.

gammapy/estimators/energydependence.py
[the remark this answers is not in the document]
AtreyeeS: If there are caveats to keep in mind, it should be clearly mentioned in the doc string and the tutorial, otherwise everyone will definitely use it.

gammapy/estimators/energydependence.py
AtreyeeS: I was unsure if the slicing takes the masks correctly into account. Inow checked that it does, so the comment is indeed irrelevant

AtreyeeS:
> > In general, I think it can be useful if you add a dataset with an energy dependent model in gammapy-data.
> 
> I agree, it would certainly speed up the tutorial also.

Or, if we don't want to add to gammapy data for one particular use case, you can do the data reduction in the tutorial using a config file to avoid those extra lines of code

Astro-Kirsty:
I decided to add a file to the `gammapy-data` [repo](https://github.com/gammapy/gammapy-data/pull/46), as this is useful in both the tests and the tutorial, significantly decreasing the unnecessary parts of the the tutorial. 
This will need to be merged and then I will push the new commits for this

gammapy/estimators/energydependence.py
Astro-Kirsty: No, that is not the desired behaviour. 
I will fix this!

bkhelifi (changes requested):
Thanks @Astro-Kirsty... Except very minor suggestions, I think that this is ready to be merged

AtreyeeS (dismissed):

gammapy/estimators/energydependence.py
```diff
@@ -0,0 +1,309 @@
...
+        -------
+        slices_src : `~gammapy.datasets.Datasets`
+            Sliced datasets.
+        """
+        model = datasets.models[self.source]
+
+        filtered_names = [name for name in datasets.models.names if name != self.source]
+        other_models = Models()
+        for name in filtered_names:
+            other_models.append(datasets.models[name])
```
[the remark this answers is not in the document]
Astro-Kirsty: There is no `Models` attribute, only the `DatasetModels`. Therefore this is not possible, hence the weird way I had to implement it

Astro-Kirsty:
I have implemented all comments here. 
@AtreyeeS, @registerrier, @bkhelifi any final comments?

AtreyeeS (approved):

registerrier (approved):
Thanks @Astro-Kirsty . This looks good. No further comments.

PR #4785, opened 2023-10-06
Use interpolation for dark matter mass

### LINKED ISSUE #4781: Nearest neighbour interpolation for mass of `PrimaryFlux` produces high fluctuations for ULs
**Problem**
I was trying to compute some upper limits using the `DarkMatterAnnihilationSpectralModel` which uses the PPPC4DM gamma fluxes via the `PrimaryFlux` class. 
The PPPC fluxes are tabular data of flux points for given (mass, energy). 
In the current implementation of `PrimaryFlux` the mass parameter `mDM` is interpolated using nearest neighbour interpolation.

https://github.com/gammapy/gammapy/blob/cecb6f88328e894de1afb374a181f6203e0e9f08/gammapy/astro/darkmatter/spectra.py#L91-L96

This results in high fluctuations when calculating upper limits:

Screenshot 2023-10-04 at 11 45 37

I took a look a the ratio of the interpolated masses and the masses that are used for this plot and this totally matches.

Screenshot 2023-10-04 at 11 46 53

**Suggested Solution**
Find out how to interpolate the values properly

StFroese:
Fixes #4781 

The `PrimaryFlux` is now based on `TemplateNDSpectralModel`. My upper limit plot looks better now haha :shipit: 

Before:
272549014-0be46766-b6d0-48dc-9372-8acc16eb342d

After:
Screenshot 2023-10-06 at 16 39 19

### DIFF
--- a/gammapy/astro/darkmatter/spectra.py
+++ b/gammapy/astro/darkmatter/spectra.py
@@ -2,27 +2,40 @@
 """Dark matter spectra."""
 import numpy as np
 import astropy.units as u
+from astropy.coordinates import SkyCoord
 from astropy.table import Table
+from regions import CircleSkyRegion
+from gammapy.maps import MapAxis, RegionNDMap
 from gammapy.modeling import Parameter
 from gammapy.modeling.models import (
     SPECTRAL_MODEL_REGISTRY,
     SpectralModel,
+    TemplateNDSpectralModel,
     TemplateSpectralModel,
 )
+from gammapy.utils.deprecation import deprecated
 from gammapy.utils.interpolation import LogScale
 from gammapy.utils.scripts import make_path
 
 __all__ = ["PrimaryFlux", "DarkMatterAnnihilationSpectralModel"]
 
 
-class PrimaryFlux:
+class PrimaryFlux(TemplateNDSpectralModel):
     """DM-annihilation gamma-ray spectra.
 
     Based on the precomputed models by Cirelli et al. (2016). All available
     annihilation channels can be found there. The dark matter mass will be set
     to the nearest available value. The spectra will be available as
-    `~gammapy.modeling.models.TemplateSpectralModel` for a chosen dark matter mass and
-    annihilation channel.
+    `~gammapy.modeling.models.TemplateNDSpectralModel` for a chosen dark matter mass and
+    annihilation channel. Using a `~gammapy.modeling.models.TemplateNDSpectralModel`
+    allows the interpolation beween different dark matter masses.
+
+    Parameters
+    ----------
+    mDM : `~astropy.units.Quantity`
+        Dark matter mass
+    channel: str
+        Annihilation channel
 
     References
     ----------
@@ -63,6 +76,8 @@ class PrimaryFlux:
 
     table_filename = "$GAMMAPY_DATA/dark_matter_spectra/AtProduction_gammas.dat"
 
+    tag = ["PrimaryFlux", "dm-pf"]
+
     def __init__(self, mDM, channel):
 
         self.table_path = make_path(self.table_filename)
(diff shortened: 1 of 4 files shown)


StFroese:
@adonath can you review? :)

gammapy/astro/darkmatter/spectra.py
```diff
@@ -114,19 +143,15 @@ def channel(self, channel):
         else:
             self._channel = channel
 
-    @property
-    def table_model(self):
```
adonath, line 118: Please keep this method for backwards compatibility. We have to deprecate it for 1-2 minor version before we can remove it. See https://docs.gammapy.org/1.1/development/dev_howto.html?highlight=deprecation#how-to-handle-api-breaking-changes

adonath (changes requested):
Thanks @StFroese, this looks very good! I left one comment with regards to backwards compatibility.

adonath (changes requested):
Thanks @StFroese! One last comment from my side: please add a "regression test", this means a test, that would fail for nearest-neighbor interpolation (as it was before) and now passed correctly. Just evaluating the model on a 2-3 values and asserting the results is fine.

gammapy/astro/darkmatter/spectra.py
```diff
@@ -80,20 +95,37 @@ def __init__(self, mDM, channel):
...
+
+        mass_axis = MapAxis.from_nodes(masses, name="mass", interp="log", unit="GeV")
+        log10x_axis = MapAxis.from_nodes(log10x, name="energy_true")
+
+        channel_name = self.channel_registry[self.channel]
+        regionMap = RegionNDMap.create(
+            region, axes=[log10x_axis, mass_axis], data=self.table[channel_name]
+        )
+
+        super().__init__(regionMap)
```
adonath, line 116:
```suggestion
        super().__init__(region_map)
```

gammapy/astro/darkmatter/spectra.py
```diff
@@ -80,20 +95,37 @@ def __init__(self, mDM, channel):
...
-        self.mDM = mDM
         self.channel = channel
 
+        # create RegionNDMap for channel
+
+        masses = np.unique(self.table["mDM"])
+        log10x = np.unique(self.table["Log[10,x]"])
+
+        coord = SkyCoord(0, 0, unit="deg", frame="galactic")
+        region = CircleSkyRegion(coord, radius=1 * u.deg)
```
adonath, line 106:
```suggestion
```

gammapy/astro/darkmatter/spectra.py
```diff
@@ -80,20 +95,37 @@ def __init__(self, mDM, channel):
...
+
+        mass_axis = MapAxis.from_nodes(masses, name="mass", interp="log", unit="GeV")
+        log10x_axis = MapAxis.from_nodes(log10x, name="energy_true")
+
+        channel_name = self.channel_registry[self.channel]
+        regionMap = RegionNDMap.create(
+            region, axes=[log10x_axis, mass_axis], data=self.table[channel_name]
```
adonath, line 113:
```suggestion
            region=None, axes=[log10x_axis, mass_axis], data=self.table[channel_name]
```

adonath (approved):
Thanks @StFroese! I left some inline edits to merge...otherwise this is ready now from my side!

adonath (approved):
Thanks, I have no further comments! 

@registerrier or @QRemy please provide a 2nd review...

gammapy/astro/darkmatter/spectra.py
```diff
@@ -3,26 +3,37 @@
...
-    `~gammapy.modeling.models.TemplateSpectralModel` for a chosen dark matter mass and
-    annihilation channel.
+    `~gammapy.modeling.models.TemplateNDSpectralModel` for a chosen dark matter mass and
+    annihilation channel. Using a `~gammapy.modeling.models.TemplateNDSpectralModel`
+    allows the interpolation beween different dark matter masses.
+
+    Parameters
+    ----------
+    mDM : `~astropy.units.Quantity`
+        Dark matter mass
```
registerrier, line 34: While this may be obvious, it could be worth to mention that the mass is given as rest mass energy (ie m c^2)

registerrier (dismissed):
Thanks @StFroese . This looks good!

Maybe one general question regarding extrapolation. Should it be allowed by default (as is the case here) or should it be forbidden?

registerrier:
> @registerrier I haven't thought about that and my first thought is to not allow it since we don't know if it's possible

I suppose that extrapolation in `mDM` should be avoided, but they are necessary in `log10x` since you might want to estimate the flux above the DM particle mass.
So probably this requires a check on the `mDM` setter but we cannot enforce `axtrapolate=False`.

StFroese:
@registerrier yes I agree. I will check if `mDM` is out of range

registerrier (approved):
Thanks @StFroese . This looks good!

PR #4799, opened 2023-10-17
Fix plotting for Energy Dependent TemplateSpatialModel

AtreyeeS:
A simple edit to get the map geom for Template Spatial Models for plotting.

Side question: Should Spatial Models be strict about the `energy_true` axis name, specially for Template Maps? The templates usually have an `energy` axis, rather than `energy_true`.

### DIFF
--- a/gammapy/modeling/models/spatial.py
+++ b/gammapy/modeling/models/spatial.py
@@ -275,10 +275,13 @@ class SpatialModel(ModelBase):
             )
 
         if geom is None:
-            width = 2 * max(self.evaluation_radius, 0.1 * u.deg)
-            geom = WcsGeom.create(
-                skydir=self.position, frame=self.frame, width=width, binsz=0.02
-            )
+            if isinstance(self, TemplateSpatialModel):
+                geom = self.map.geom
+            else:
+                width = 2 * max(self.evaluation_radius, 0.1 * u.deg)
+                geom = WcsGeom.create(
+                    skydir=self.position, frame=self.frame, width=width, binsz=0.02
+                )
         data = self.evaluate_geom(geom)
         return Map.from_geom(geom, data=data.value, unit=data.unit)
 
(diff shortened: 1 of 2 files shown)


registerrier (approved):

Is it possible to test the new plots supported?

AtreyeeS:
Added a test @registerrier . I think the test fails here are unrelated

gammapy/modeling/models/spatial.py
```diff
@@ -275,10 +275,13 @@ def _get_plot_map(self, geom):
             )
 
         if geom is None:
-            width = 2 * max(self.evaluation_radius, 0.1 * u.deg)
-            geom = WcsGeom.create(
-                skydir=self.position, frame=self.frame, width=width, binsz=0.02
-            )
+            if isinstance(self, TemplateSpatialModel):
+                geom = self.map.geom
```
bkhelifi, line 279: A very naive question: a spatial model has a map as internal attribute?

registerrier:
> Added a test @registerrier . I think the test fails here are unrelated

There was a CI issue solved now. If you rebase it should be fine.

gammapy/modeling/models/spatial.py
registerrier: a `TemplateSpatialModel` does, yes.

gammapy/modeling/models/spatial.py
```diff
@@ -275,10 +275,13 @@ def _get_plot_map(self, geom):
             )
 
         if geom is None:
-            width = 2 * max(self.evaluation_radius, 0.1 * u.deg)
-            geom = WcsGeom.create(
-                skydir=self.position, frame=self.frame, width=width, binsz=0.02
-            )
+            if isinstance(self, TemplateSpatialModel):
```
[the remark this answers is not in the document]
AtreyeeS: The hidden function `_get_plot_map` is not checked explicitly. But this is indirectly checked through plotting the individual Spatial models eg: https://github.com/gammapy/gammapy/blob/e6e5dd848b17962b4b9019e06f4617a6df7c9692/gammapy/modeling/models/tests/test_spatial.py#L461

Issue #4711, opened 2023-08-02
Documentation of `SpectrumDatasetMaker` is misleading.


MRegeard:
**Version** : dev

Because `SpectrumDatasetMaker` inherit from `MapDatasetMaker`, the return of the `run` method of the maker states that a `MapDataset` is returned. 
This is obviously misleading for the user. However, I am not sure that a clean solution exist to avoid this.

PR #4800, opened 2023-10-17
Documentation of `SpectrumDatasetMaker`, `NotImplementedError` methods and cleanup

MRegeard:
This PR fix issue #4711. I redefine methods that are inherited from the parent class, write a new docstring and call-return the parent methods. This way, the behaviour is not changed, but the docstring is more specific to the child class. If I miss some places in the code where it should be done, let me know, and it can be added to this PR.

In addition:

- I added docstring on methods that are raising `NotImplementedError`. 

- I cleaned up the `PhaseBackgroundMaker` so that the docstring are in agreements with gammapy standard.

### DIFF
--- a/gammapy/datasets/spectrum.py
+++ b/gammapy/datasets/spectrum.py
@@ -292,11 +292,11 @@ class SpectrumDatasetOnOff(PlotMixin, MapDatasetOnOff):
     tag = "SpectrumDatasetOnOff"
 
     def cutout(self, *args, **kwargs):
-        """Not supported for `SpectrumDataset`"""
+        """Not supported for `SpectrumDatasetOnOff`"""
         raise NotImplementedError("Method not supported on a spectrum dataset")
 
     def plot_residuals_spatial(self, *args, **kwargs):
-        """Not supported for `SpectrumDataset`"""
+        """Not supported for `SpectrumDatasetOnOff`"""
         raise NotImplementedError("Method not supported on a spectrum dataset")
 
     @classmethod
--- a/gammapy/datasets/spectrum.py
+++ b/gammapy/datasets/spectrum.py
@@ -265,12 +265,15 @@ class SpectrumDataset(PlotMixin, MapDataset):
     tag = "SpectrumDataset"
 
     def cutout(self, *args, **kwargs):
+        """Not supported for `SpectrumDataset`"""
         raise NotImplementedError("Method not supported on a spectrum dataset")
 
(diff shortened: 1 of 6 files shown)


gammapy/datasets/spectrum.py
```diff
@@ -289,9 +292,11 @@ class SpectrumDatasetOnOff(PlotMixin, MapDatasetOnOff):
     tag = "SpectrumDatasetOnOff"
 
     def cutout(self, *args, **kwargs):
+        """Not supported for `SpectrumDataset`"""
```
registerrier, line 295:
```suggestion
        """Not supported for `SpectrumDatasetOnOff`"""
```

gammapy/datasets/spectrum.py
```diff
@@ -289,9 +292,11 @@ class SpectrumDatasetOnOff(PlotMixin, MapDatasetOnOff):
...
         raise NotImplementedError("Method not supported on a spectrum dataset")
 
     def plot_residuals_spatial(self, *args, **kwargs):
+        """Not supported for `SpectrumDataset`"""
```
registerrier, line 299:
```suggestion
        """Not supported for `SpectrumDatasetOnOff`"""
```

gammapy/makers/spectrum.py
```diff
@@ -91,3 +91,43 @@ def make_exposure(self, geom, observation):
             exposure.quantity *= containment.reshape(geom.data_shape)
 
         return exposure
+
+    @staticmethod
+    def make_counts(geom, observation):
+        """Make counts map.
+
+        If the `~gammapy.maps.Geom` is built from a `~regions.CircleSkyRegion`,
```
registerrier, line 99:
```suggestion
        If the `~gammapy.maps.RegionGeom` is built from a `~regions.CircleSkyRegion`,
```

gammapy/makers/spectrum.py
```diff
@@ -91,3 +91,43 @@ def make_exposure(self, geom, observation):
...
+        the latter will be directly used to extract the counts.
+        If instead the `~gammapy.maps.Geom` is built from a `~regions.PointSkyRegion`,
```
registerrier, line 101:
```suggestion
        If instead the `~gammapy.maps.RegionGeom` is built from a `~regions.PointSkyRegion`,
```

registerrier (dismissed):
Thanks @MRegeard . This looks good. See some inline suggestions.

registerrier (approved):
Thanks @MRegeard .

Issue #4768, opened 2023-09-26
Arguments of the class `Observation' to be described 


bkhelifi:
Some arguments of the class `Observation' have no description in the docstring of the class. There are:
- pointing
- location

PR #4805, opened 2023-10-17
Description of the arguments of the class `Observation'

bkhelifi:
**Description**

This pull request completes the docstring description of all arguments of the class and of the `create` function. It is associated to the PR #4768

### DIFF
--- a/gammapy/data/observations.py
+++ b/gammapy/data/observations.py
@@ -36,26 +36,30 @@ class Observation:
     Parameters
     ----------
     obs_id : int
-        Observation id
+        Observation id.
     obs_info : dict
-        Observation info dict
+        Observation info dict.
     aeff : `~gammapy.irf.EffectiveAreaTable2D`
-        Effective area
+        Effective area.
     edisp : `~gammapy.irf.EnergyDispersion2D`
-        Energy dispersion
+        Energy dispersion.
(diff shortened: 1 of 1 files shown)


morcuended (approved):

registerrier (approved):
Thanks @bkhelifi . This looks good. 
Do we have a scheme defined for punctuation of parameters and returned objects in docstrings?

morcuended:
> Do we have a scheme defined for punctuation of parameters and returned objects in docstrings?

I see phrases ending or not with period indistinctly in the docstrings. I guess we should follow https://peps.python.org/pep-0257/ (which recommends the use of periods)

Also realize that in some cases in which parameter description is indicated in several lines, are rendered like if they were in the same line:
image

docstring looks like:
image

registerrier:
> > Do we have a scheme defined for punctuation of parameters and returned objects in docstrings?
> 
> I see phrases ending or not with period indistinctly in the docstrings. I guess we should follow https://peps.python.org/pep-0257/ (which recommends the use of periods)
> 
> Also realize that in some cases in which parameter description is indicated in several lines, are rendered like if they were in the same line: image
> 

At some point, we will need to make all docstrings uniform. This might require a reminder issue.

Merging for now.

Astro-Kirsty:
Resolves PR #4768

Issue #4563, opened 2023-06-06
Improve error traceback for DatasetsMaker


registerrier:
**Is your feature request related to a problem? Please describe.**
When using `DatasetsMaker` with `n_jobs>1`, if an error occurs during a specific dataset creation, the execution stops with a message `RuntimeError: Execution of a sub-process failed` which does not give any information neither on which process failed nor for what reason it failed.

**Describe the solution you'd like**
Get minimal error traceback and the dataset name that caused trouble.
Ideally, an option could be set to simply skip failing datasets with a log.warning info.

AtreyeeS:
I think this is not limited to the `DatasetsMaker`, but occurs for other cases involving multiple processors (eg: the estimators)

Astro-Kirsty:
I found a related issue in DatasetsMaker which is hard to trace back, caused by the following warning 
`ReflectedRegionsBackgroundMaker failed. No OFF region found outside exclusion mask for dataset 'wkPpkwsV'.
ReflectedRegionsBackgroundMaker failed. Setting wkPpkwsV mask to False.`
![Screenshot from 2023-09-28 17-23-43](https://github.com/gammapy/gammapy/assets/60767360/91d60ac6-93a1-4c7d-afd3-7bda2e72dbc6)

adonath:
I propose to return an empty map for `counts_off` as well: https://github.com/gammapy/gammapy/blob/main/gammapy/makers/background/reflected.py#L530

The alternative would be to check for `counts_off is None` in many other places...

bkhelifi:
There are 2 issues:

- [ ] the function `makers.callback` used by `DatasetsMaker` in case of multi-processions calls the function `map.to_map_dataset` that is not protected when `counts_off is None` 
- [ ] the multiprocessing functions should catch any error properly and terminate in a clean manner the full execution

PR #4807, opened 2023-10-17
Fix the case of none `counts_off` for the function `to_map_dataset`

bkhelifi:
**Description**
This pull request corrects the bug in the function `to_map_dataset` when the `counts_off` is None

It is associated to the Issue #4563


MRegeard (approved):
@bkhelifi Thanks ! Seems good to me.

registerrier (approved):
Thanks @bkhelifi . This looks good to me.

Issue #4806, opened 2023-10-17
Add a generator on Observations to yield full in memory observation


registerrier:
**Is your feature request related to a problem? Please describe.**
Currently observations do not go out of scope with the for loop. Hence we need to remove the caching of the data read from disk. This causes multiple access to a single file in a regular data reduction process which reduces performance especially for larges zipped files.

**Describe the solution you'd like**
A generator could defined on `Observations` that would yield a single in-memory `Observation` object. Since it would go out of scope at the end of the loop, the memory issues should go away.

**Describe alternatives you've considered**
We could also use an iterator? 
Only calling observation.events once is not really possible as we use the dataset maker and the background makers in a regular loop (esp. in 1D with `ReflectedRegionsBackgroundMaker`)

adonath:
A context manager and a global "switch" might work as well:

```python 

with cache_observation():
    for obs in observations:
       ...
```
But I guess the iterator is simpler and cleaner...

PR #4808, opened 2023-10-17
adding `Observations` in memory generator

MRegeard:
This PR adds an iterator to `Observations` as discuss in #4806.

### DIFF
--- a/gammapy/data/observations.py
+++ b/gammapy/data/observations.py
@@ -787,17 +787,9 @@ class Observations(collections.abc.MutableSequence):
         obs = itertools.chain(*observations_list)
         return cls(list(obs))
 
-    def generator(self, idx=None):
-        """Iterate over observations.
-
-        Parameters
-        ----------
-        idx : slice
-            Slice object defining which observations to return. Default is None, which means all observations.
-        """
-        if idx is None:
-            idx = slice(None)
-        for obs in self[idx]:
+    def in_memory_generator(self):
+        """A generator that iterates over observation. It yields an in memory copy of the observation."""
+        for obs in self:
             obs_copy = obs.copy(in_memory=True)
             yield obs_copy
             del obs
--- a/gammapy/data/observations.py
+++ b/gammapy/data/observations.py
@@ -792,8 +792,6 @@ class Observations(collections.abc.MutableSequence):
         for obs in self:
             obs_copy = obs.copy(in_memory=True)
             yield obs_copy
-            del obs
-            del obs_copy
 
 
 class ObservationChecker(Checker):
--- a/gammapy/data/observations.py
+++ b/gammapy/data/observations.py
@@ -787,15 +787,17 @@ class Observations(collections.abc.MutableSequence):
         obs = itertools.chain(*observations_list)
         return cls(list(obs))
 
-    def iter(self, slice=None):
+    def iter(self, idx=None):
         """Iterate over observations.
 
         Parameters
         ----------
-        slice : slice
+        idx : slice
             Slice object defining which observations to return. Default is None, which means all observations.
         """
-        for obs in self[slice]:
+        if idx is None:
+            idx = slice(None)
+        for obs in self[idx]:
             yield obs.copy(in_memory=True)
 
 
--- a/gammapy/data/observations.py
+++ b/gammapy/data/observations.py
@@ -787,6 +787,17 @@ class Observations(collections.abc.MutableSequence):
         obs = itertools.chain(*observations_list)
         return cls(list(obs))
 
+    def iter(self, slice=None):
+        """Iterate over observations.
+
+        Parameters
+        ----------
+        slice : slice
+            Slice object defining which observations to return. Default is None, which means all observations.
+        """
+        for obs in self[slice]:
+            yield obs.copy(in_memory=True)
+
 
 class ObservationChecker(Checker):
     """Check an observation.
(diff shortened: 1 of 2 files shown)


gammapy/data/observations.py
```diff
@@ -744,6 +744,19 @@ def from_stack(cls, observations_list):
         obs = itertools.chain(*observations_list)
         return cls(list(obs))
 
+    def iter(self, idx=None):
```
registerrier: Maybe call this `generator()` not to confuse with the `__iter__()` method?

I would prefer no to pass the slice as argument and rely on `Observations[slice]` instead, but I understand that the latter returns a `list` and not an `Observations`. Maybe this is an issue that should be corrected.

gammapy/data/observations.py
```diff
@@ -744,6 +744,19 @@ def from_stack(cls, observations_list):
...
+
+        Parameters
+        ----------
+        idx : slice
+            Slice object defining which observations to return. Default is None, which means all observations.
+        """
+        if idx is None:
+            idx = slice(None)
+        for obs in self[idx]:
+            yield obs.copy(in_memory=True)
```
registerrier: do you need to explicitly delete the copy here? 
or do someting like 
```
obs_copy = obs.copy(in_memory=True)
yield obs_copy
```

gammapy/data/observations.py
MRegeard: Done. For the slice, see #4815.

gammapy/data/observations.py
MRegeard: Don't know ...

adonath: I think there is no need to trigger the `del` explicitly, as long as the observation goes out of scope. Which it typically does...

adonath:
@MRegeard You can use a tool called `psrecord` (https://pypi.org/project/psrecord/) to check the memory use and runtime of some example data reduction and share the result here.

gammapy/data/tests/test_observations.py
```diff
@@ -529,3 +535,15 @@ def test_stack_observations(data_store, caplog):
 
     with pytest.raises(TypeError):
         Observations.from_stack([obs_1, ["a"]])
+
+
+@requires_data()
+def test_observations_iter(data_store):
```
morcuended, line 541:
```suggestion
def test_observations_generator(data_store):
```

gammapy/data/tests/test_observations.py
```diff
@@ -529,3 +535,15 @@ def test_stack_observations(data_store, caplog):
...
+    """Test Observations.iter()"""
```
morcuended, line 542:
```suggestion
    """Test Observations.generator()"""
```

morcuended (commented):
@MRegeard for consistency with the name of the method I think these should be changed

morcuended:
I did some tests of `Observations.generator()`(+ some basic 1d spectral fitting) using LST-1 Crab dataset (~35 h of observations split in around 100 observation DL3 files). Memory and CPU usage attached below.

Using unzipped files takes around 2 min (with v1.1 it is around 3 min)
generator_unzip

Using zipped files takes around 10 min (with v1.1 it is around 30 min) letting Gammapy doing the unzipping.
generator_zip

adonath:
Thanks @MRegeard and @morcuended for the measurements, this looks like a clear improvement! So certainly worth the work...

MRegeard:
I change the name to `in_memory_generator`. Do you have opinion on this name ? Is it explicit enough ?
@adonath, @registerrier, @bkhelifi.

morcuended:
@MRegeard, I added the plots of memory usage without the in memory observations generator as a comparison to the ones produced using it (see my comment above).

gammapy/data/observations.py
```diff
@@ -787,6 +787,14 @@ def from_stack(cls, observations_list):
         obs = itertools.chain(*observations_list)
         return cls(list(obs))
 
+    def in_memory_generator(self):
+        """A generator that iterates over observation. It yields an in memory copy of the observation."""
+        for obs in self:
+            obs_copy = obs.copy(in_memory=True)
+            yield obs_copy
+            del obs
```
registerrier, line 795: Normally this should not be needed.

gammapy/data/observations.py
```diff
@@ -787,6 +787,14 @@ def from_stack(cls, observations_list):
...
+            del obs_copy
```
registerrier, line 796: What happens if you remove this?

MRegeard: need to check !

morcuended: I can test removing the two del's

MRegeard: Would be great ! Thanks

morcuended: Checking this. Sorry for the long delay

MRegeard: No worries !

morcuended:
For gzipped files:

 - using the generator with `del`
generator_with_del_gzip

 - using the generator without `del`'s
generator_without_del_gzip

For unzipped files:

 - using the generator with `del`
generator_with_del_unzip

 - using the generator without `del`'s
generator_without_del_unzip

I think it does not make any difference whether to include or not the `del`'s.

(In both cases I reduced >100 runs)

registerrier:
Thanks @morcuended ! @MRegeard, can you remove the dels?

MRegeard:
Thanks @morcuended !
@registerrier done !

registerrier (approved):

MRegeard:
Closes #4806

Issue #4742, opened 2023-09-08
Please revise get_wstat_mu_bkg


moralejo:
https://github.com/gammapy/gammapy/blob/c71cb4707087c5278a15913b46828718562a8b10/gammapy/stats/fit_statistics.py#L211

I think the denominator here should be just 2 (1+alpha) 
If that is the case, the current function will only provide the bext mu_bkg for the case alpha=1

moralejo:
Hi,

the code is correct, but it return the Poisson parameter of the background **in the OFF region**

Whereas the same formula in the documentation: 
https://docs.gammapy.org/1.1/user-guide/stats/wstat_derivation.html

is used to obtain the Poisson parameter of the background **in the ON region** - that is what mu_bkg means in the documentation. So probably only the docs are wrong.

Here: https://github.com/gammapy/gammapy/blob/c71cb4707087c5278a15913b46828718562a8b10/gammapy/datasets/map.py#L2121-L2126

the return of the function (for the whole OFF region) is scaled by alpha, to get the value for the ON region, so it seems correct.
It would anyway be worth that an expert checks it. I also think it would be better to use, for the background Poisson parameters, variable names which make it clear straightaway whether they refer to the ON region or to the OFF region.

adonath:
Thanks @moralejo! Which modifications would you propose to https://docs.gammapy.org/1.1/user-guide/stats/index.html#notations, to make the description more clear?

moralejo:
> Thanks @moralejo! Which modifications would you propose to https://docs.gammapy.org/1.1/user-guide/stats/index.html#notations, to make the description more clear?

Hi,

I would say 

"mu_bkg: predicted background counts in the **ON** region" ( this is what it means here, since below we have mu_off = mu_bkg / alpha) 

Then, for consistency, the function get_wstat_mu_bkg should either be called get_wstat_mu_off , or be modified to return mu_bkg (and the code using the function should be changed as well). 

In the current documentation https://docs.gammapy.org/1.1/user-guide/stats/wstat_derivation.html (not sure if everywhere else) mu_bkg has the same meaning as in the glossary - ok!    
But then the formulae here:
https://docs.gammapy.org/1.1/user-guide/stats/wstat_derivation.html#profile-likelihood 
need fixing (they were taken from a reference in which mu_bkg meant what in gammapy is called mu_noff). These would be the changes (crosscheck is welcome):

[...]

PR #4809, opened 2023-10-17
Fix WSTAT docs

AtreyeeS:
This addresses #4742
The inconsistency was between `mu_bkg` and `mu_bkg/alpha`

I am not sure about the requested change in https://github.com/gammapy/gammapy/issues/4742#issuecomment-1711617402 (ie, change in the arguments of L). What we have now give the likelihood of observing `n1` given `n1` photons in the region. Replacing `mu` by the general case is correct at this step?

### DIFF
--- a/docs/user-guide/stats/index.rst
+++ b/docs/user-guide/stats/index.rst
@@ -29,7 +29,7 @@ Variable          Dataset attribute name Definition
 ``n_sig``		  ``excess``			 Estimated signal counts defined as ``n_on`` - ``n_bkg``
 ``mu_on``         ``npred``              Predicted counts
 ``mu_off``        ``npred_off``          Predicted counts in the off region
-``mu_bkg``        ``npred_background``   Predicted background counts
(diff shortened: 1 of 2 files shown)


registerrier:
> I am not sure about the requested change in [#4742 (comment)](https://github.com/gammapy/gammapy/issues/4742#issuecomment-1711617402) (ie, change in the arguments of L). What we have now give the likelihood of observing `n1` given `n1` photons in the region. Replacing `mu` by the general case is correct at this step?

I think the point in the goodness of fit section is that the `L` function should take the same arguments, therefore it should not be `L(n_on, n_off, n_on, n_off)` but rather `L(n_on, n_off, alpha, n_on-alpha n_off, n_off)`

registerrier (approved):

Maybe simply modify the goodness of fit part as well regarding argument of `L`.

registerrier (approved):
I have changed the `L` expression in two additional ratios.

PR #4819, opened 2023-10-18
Management of metadata for `Models`

bkhelifi:
**Description**
This pull request aims to deal with metadata for `Models`. 

For the moment, models metadata are not kept in a dedicated metadata class. They are just written in file, and read when a basic check of the Gammapy version.

### DIFF
--- a/gammapy/utils/metadata.py
+++ b/gammapy/utils/metadata.py
@@ -76,7 +76,33 @@ class CreatorMetaData(MetaData):
 
     @validator("date")
     def validate_time(cls, v):
-        return Time(v)
+        if v is not None:
+            return Time(v)
+        else:
+            return v
+
+    def to_header(self, format="gadf"):
+        """Convert creator metadata to fits header.
+
+        Parameters
+        ----------
+        format : str
+            header format. Default is 'gadf'.
+
+        Returns
+        -------
+        header : dict
+            the header dictionary
+        """
+        if format != "gadf":
+            raise ValueError(f"Creator metadata: format {format} is not supported.")
+
+        hdr_dict = {}
+        hdr_dict["CREATED"] = self.date.iso
+        hdr_dict["CREATOR"] = self.creator
+        hdr_dict["ORIGIN"] = self.origin
+
+        return hdr_dict
 
     @classmethod
     def from_default(cls):
@@ -84,3 +110,22 @@ class CreatorMetaData(MetaData):
         date = Time.now()
         creator = f"Gammapy {version}"
         return cls(creator=creator, date=date)
+
+    @classmethod
+    def from_header(cls, hdr, format="gadf"):
+        """Builds creator metadata from fits header.
+
+        Parameters
+        ----------
+        hdr : dict
+            the header dictionary
+        format : str
+            header format. Default is 'gadf'.
+        """
+        if format != "gadf":
+            raise ValueError(f"Creator metadata: format {format} is not supported.")
+
+        date = hdr.get("CREATED", None)  # note regular FITS keyword is DATE
+        origin = hdr.get("ORIGIN", None)
+        creator = hdr.get("CREATOR", None)
+        return cls(creator=creator, date=date, origin=origin)
--- a/gammapy/data/__init__.py
+++ b/gammapy/data/__init__.py
@@ -5,6 +5,7 @@ from .event_list import EventList
 from .filters import ObservationFilter
 from .gti import GTI
 from .hdu_index_table import HDUIndexTable
+from .metadata import ObservationMetaData
 from .obs_table import ObservationTable
 from .observations import Observation, Observations
 from .observers import observatory_locations
@@ -19,6 +20,7 @@ __all__ = [
     "Observation",
     "ObservationFilter",
     "Observations",
+    "ObservationMetaData",
     "ObservationTable",
     "observatory_locations",
     "PointingInfo",
--- a/gammapy/data/metadata.py
+++ b/gammapy/data/metadata.py
@@ -1,0 +1,164 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+from typing import Optional, Union
+import numpy as np
+import astropy.units as u
+from astropy.coordinates import EarthLocation, SkyCoord
+from astropy.time import Time
+from pydantic import Field, ValidationError, validator
+from gammapy.utils.fits import earth_location_from_dict
+from gammapy.utils.metadata import CreatorMetaData, MetaData
+from gammapy.utils.time import time_ref_from_dict
+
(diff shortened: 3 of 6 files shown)


gammapy/modeling/models/tests/test_io.py
```diff
@@ -483,3 +484,19 @@ def test_compound_models_io(tmp_path):
...
+
+def test_meta_io(caplog, tmp_path):
+    m = PowerLawSpectralModel()
+    sk = SkyModel(spectral_model=m, name="model")
+    Models([sk]).write(tmp_path / "test.yaml")
+
+    sk_dict = read_yaml(tmp_path / "test.yaml")
+    assert "metadata" in sk_dict
+
+    sk_dict["metadata"]["creator"] = "Gammapy 1.1"
```
morcuended, line 497: Here, if you do not define it as "Gammapy 1.1" it would be set by default as dev version by the `from_default` metadata creator? Perhaps you can also test that the warning is present in that case in `test.yaml`

morcuended (commented):
thanks @bkhelifi. I just left a few comments regarding the log message and the test.

gammapy/modeling/models/tests/test_io.py
bkhelifi: done

gammapy/modeling/models/core.py
```diff
@@ -428,6 +430,18 @@ def read(cls, filename):
...
         data = yaml.safe_load(yaml_str)
+        if "metadata" in data.keys():
+            creator_version = data["metadata"]["creator"].split()[1]
+            if check_version(creator_version) < 0:
+                log.warning(
+                    f"The currently used Gammapy version is older than the one used to create the model ({creator_version})"
+                )
+            else:
+                log.debug(
+                    f"The currently used Gammapy version is compatible with the one used to create the model ({creator_version})"
```
AtreyeeS, line 441: Is it better to say "compatible" or just "same". Older versions are not necessarily incompatible

AtreyeeS: This also includes the case of check_version > 1 ? Older versions may not support new models

gammapy/modeling/models/core.py
```diff
@@ -428,6 +430,18 @@ def read(cls, filename):
     def from_yaml(cls, yaml_str, path=""):
         """Create from YAML string."""
         data = yaml.safe_load(yaml_str)
+        if "metadata" in data.keys():
```
AtreyeeS, line 433: This meta information will be created only if a model is read from disc? Just 

```
m = PowerLawSpectralModel()
sk = SkyModel(spectral_model=m, name="model")
sk.to_dict()["metadata"]
```
is `None` ?

AtreyeeS (commented):
Thanks @bkhelifi !

gammapy/modeling/models/core.py
```diff
@@ -428,6 +430,18 @@ def read(cls, filename):
...
+            creator_version = data["metadata"]["creator"].split()[1]
```
registerrier, line 434: If version is important, I would suggest to split it  directly in the `CreatorMetaData`:
maybe having `creation.software` and `creation.version` would be simpler

gammapy/modeling/models/core.py
```diff
@@ -509,7 +523,9 @@ def write(
             self.write_covariance(base_path / filecovar, **kwargs)
             self._covar_file = filecovar
 
-        path.write_text(self.to_yaml(full_output, overwrite_templates))
+        creation = CreatorMetaData.from_default()
+        text = self.to_yaml(full_output, overwrite_templates) + creation.to_yaml()
```
registerrier, line 527: Maybe this makes more sense to move this directly into `Models.to_yaml()` no?

gammapy/modeling/models/core.py
registerrier, line 434: Here you could create the `CreatorMetaData` from the dict.

gammapy/modeling/models/core.py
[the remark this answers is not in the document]
registerrier: Probably printing the version of the model opened is already enough info?

gammapy/modeling/models/core.py
```diff
@@ -428,6 +430,18 @@ def read(cls, filename):
     def from_yaml(cls, yaml_str, path=""):
         """Create from YAML string."""
         data = yaml.safe_load(yaml_str)
+        if "metadata" in data.keys():
+            creator_version = data["metadata"]["creator"].split()[1]
+            if check_version(creator_version) < 0:
```
registerrier, line 435: If there are other places where a check on the version is to be performed, this might be defined as a method on `CreatorMetaData` .

gammapy/utils/metadata.py
```diff
@@ -52,8 +52,9 @@ def from_header(cls, hdr):
     def to_yaml(self):
         """Dumps metadata content to yaml."""
         meta = json.loads(self.json())
+        metadict = {"metadata": meta}
```
registerrier, line 55: I wonder if a single level is enough here. Or if we need a general `metadata` level with hierarchical entries:
e.g.  ["metadata"]["creation"]

registerrier (commented):
Thanks @bkhelifi . I have left some inline comments.

If the product creator version comparison is to be present in various places, I would suggest to make this available at the level of the metadata object.

gammapy/modeling/models/core.py
bkhelifi: They can never support them! except if the code is changed... But as you can see, this test gives only a log message. There is no stop-over

bkhelifi: > Is it better to say "compatible" or just "same". Older versions are not necessarily incompatible

I do not understand your point... because the code does not say that here

gammapy/utils/check.py
```diff
@@ -22,3 +23,27 @@ def check_tutorials_setup(download_datasets_path="./gammapy-data"):
...
+
+    Parameters
+    ----------
+    testversion : str
+        Version to test.
+
+    Returns
+    -------
+    check : int
+        0 if identical, -1 if the current version is older than `testversion` and +1 if newer.
```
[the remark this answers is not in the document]
bkhelifi: And then, when used in an if statement, we test string instead of an enum? "identical", "older", "newer"?

gammapy/utils/check.py
```diff
@@ -22,3 +23,27 @@ def check_tutorials_setup(download_datasets_path="./gammapy-data"):
...
+    """
+
+    if testversion == version:
+        return 0
+    elif "dev" in version:
```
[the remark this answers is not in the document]
bkhelifi: I don't

Issue #4788, opened 2023-10-09
Expose undocumented new features from v1.1 and v1.0


registerrier:
Several features have been introduced in v1.0 and v1.1 but lack proper documentation. Ideally this should sorted for 1.2.

The main features are
- [ ] `gammapy.estimators.resample_energy_edges` : the question of providing minimum number of counts of significance per bin comes regularly. This could be exposed in several places (e.g. in an how-to and in some tutorials)
- [ ] `select_nested_models` can be demonstrated for different use cases, e.g. source TS estimation and/or spectral feature detection
- [ ] observation clustering should also be presented in its own notebook.
- [ ] we should expose the `DatasetsMaker`
- [ ] we could demonstrate using different spectral normalization models for `FoVBackgroundModel`

Are there others? 

I propose we discuss possible presentations in this issue's discussion

Astro-Kirsty:
I can create the notebook for the observation clustering as this is something I have been utilising recently

Astro-Kirsty:
I can implement the different spectral normalisation models for the `FoVBackgroundModel`. 
I think this requires some discussion. 

Initially I would say it makes sense to adapt the doc [here](url), such that a `PowerLawNormSpectralModel` is added into the `FoVBackgroundModel` and the `tilt` and `norm` are allowed to be free.  This is quite a typical process when using this model. 

Then I would suggest adding a separate tutorial which shows this and also the implementation of a more complex spectral model (i.e. the `PiecewiseNormSpectralModel`) along with how these affect the background, and how to show the various parameters of the background.

Astro-Kirsty:
For `select_nested_models` it is possible that @QRemy already has some scripts doing this, as per #4216 ?

PR #4821, opened 2023-10-18
Adapt tutorials to include PowerLawNormSpectralModel in the FoVBackgroundMaker

Astro-Kirsty:
This is to include a spectral model (`PowerLawNormSpectralModel`) in which both the tilt and norm are unfrozen, and fit for the `FoVBackgroundMaker`. This partly resolves task number 5 in #4788 

I also noticed some broken links so fixed those.

### DIFF
--- a/docs/user-guide/makers/fov.rst
+++ b/docs/user-guide/makers/fov.rst
@@ -19,7 +19,8 @@ Gammapy provides the `~gammapy.makers.FoVBackgroundMaker`. The latter creates a
 and a `~gammapy.modeling.models.NormSpectralModel` which allows to renormalize the background cube, and
 possibly to change its spectral distribution. By default, only the `norm` parameter of a
 `~gammapy.modeling.models.PowerLawNormSpectralModel` is left free. Here we show the addition of a `~gammapy.modeling.models.PowerLawNormSpectralModel`
-in which the `norm` and `tilt` parameters are unfrozen as an example.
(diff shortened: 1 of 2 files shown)


bkhelifi (approved):
Looks good to me, thanks @Astro-Kirsty

cgalelli (approved):
Thanks @Astro-Kirsty looks good

AtreyeeS:
Maybe it would be better to show a different model, like the piecewise norm? Or at least mention that there are other normed models

Astro-Kirsty:
> Maybe it would be better to show a different model, like the piecewise norm? Or at least mention that there are other normed models

So there was some discussion whether we should add a separate tutorial which shows a more complex spectral model for the background i.e. piecewise norm. And then this dedicated tutorial would include more information specifically regarding the correction of the background. Maybe @registerrier can comment here?

AtreyeeS:
I agree that we tend to have too many things in our tutorials, and that makes them hard to find unless specifically linked from the HowTo. Small tutorials showing specific examples is something I personally like, but having too many tutorials can be daunting. I don't know the best thing here. Maybe we can mention in the tutorial some other models that exist, and add code examples in the doc strings of each class to make them self sustained?

cgalelli (approved):
Looks good to me.  Thanks @Astro-Kirsty

bkhelifi (approved):
Thanks @Astro-Kirsty

Issue #4784, opened 2023-10-05
Adding a GTI.difference or GTI.remove_interval method


registerrier:
**Is your feature request related to a problem? Please describe.**
Removing a bad time interval is necessary for several use cases. The current API requires to create manually a new GTI object.

**Describe the solution you'd like**
A method such as `GTI.difference(intervals)` or `GTI.remove_interval(interval)` would support this use case easily.

**Additional context**
A similar issue is discussed in #4780

PR #4822, opened 2023-10-18
Add a function to delete a time interval from GTI

cgalelli:
As requested by the issue #4784 I added a function that removes a "bad time interval" from the GTIs with correlated test. A future addition will be a more complex handling of lists of "bad periods"

### DIFF
--- a/gammapy/data/gti.py
+++ b/gammapy/data/gti.py
@@ -323,7 +323,7 @@ class GTI:
 
         Parameters
         ----------
-        time_interval : `astropy.time.Time`
+        time_interval : [`astropy.time.Time`, `astropy.time.Time`]
             Start and stop time for the selection.
 
         Returns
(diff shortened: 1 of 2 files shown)


registerrier:
The CI issue is corrected. You can rebase to make it run.

registerrier (commented):
Thanks @cgalelli . This looks good!

I think there is one issue because you don't copy the parent Table. You are probably modifying the GTI.

Otherwise you might add a test when the bad interval is partly outside or fully outside the range of your GTI?

gammapy/data/gti.py
```diff
@@ -318,6 +318,35 @@ def select_time(self, time_interval):
 
         return self.__class__(gti_within)
 
+    def delete_interval(self, time_interval):
+        """Select and crop GTIs in time interval.
+
+        Parameters
+        ----------
+        time_interval : `astropy.time.Time`
```
bkhelifi, line 326: The docstring says that this is an interval, but the type is just one Time. Should not it be a [Time, Time]?

registerrier (approved):
Thanks @cgalelli ! Looks good to me.

Issue #4830, opened 2023-10-19
Fixing vmin to 0 and vmax to 1 when plotting mask


MRegeard:
I think that we could fix the vmin and vmax of imshow when plotting mask. 
This would avoid confusing plots like this one : 
<img width="983" alt="Capture d’écran 2023-10-19 à 15 07 26" src="https://github.com/gammapy/gammapy/assets/108189156/0e3bd2cc-efbf-4641-b931-fd374d50a506">
where the first 3 bins are completely mask and the next ones are not, but visually it is only noticeable if you look at the colorbar scale and value. 

It is really easy to do since we have an `is_mask` attributes on maps.

PR #4831, opened 2023-10-19
fix vmin vmax to (0, 1) when plotting mask

MRegeard:
This PR is related to issue #4830. 

It set vmin and vmax to (0, 1) resp. when plotting mask. This simple fix transform the example shown in the issue to this :
Capture d’écran 2023-10-19 à 15 15 33

### DIFF
--- a/gammapy/maps/wcs/ndmap.py
+++ b/gammapy/maps/wcs/ndmap.py
@@ -10,6 +10,7 @@ from astropy.coordinates import SkyCoord
 from astropy.io import fits
 from astropy.nddata import block_reduce
 from regions import PixCoord, PointPixelRegion, PointSkyRegion, SkyRegion
+import matplotlib.colors as mpcolors
 import matplotlib.pyplot as plt
 from gammapy.utils.interpolation import ScaledRegularGridInterpolator
 from gammapy.utils.units import unit_from_fits_image_hdu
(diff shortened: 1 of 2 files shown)


adonath:
I agree this is much more reasonable. We could also define a custom colormap: see e.g. https://stackoverflow.com/a/65305266/19802442

Then the colorer would also make more sense.

MRegeard:
> I agree this is much more reasonable. We could also define a custom colormap: see e.g. https://stackoverflow.com/a/65305266/19802442
> 
> Then the colorbar would also make more sense.

Yes I also thought about that. I can try to see what it looks like !

MRegeard:
with a custom map it looks like this : 
Capture d’écran 2023-10-19 à 16 13 00
I don't really know which one I prefer.

morcuended:
> I don't really know which one I prefer.

I like this last one better than the non-discrete colormap from the first entry.

AtreyeeS (dismissed):
Thanks @MRegeard ! Quite useful

adonath:
I'm also +1 for the binary colormap. The other colors in the continuous colormap are meaningless...

Astro-Kirsty (approved):

morcuended (dismissed):
thanks @MRegeard

registerrier (commented):
Thanks @MRegeard . This is indeed a good idea. Could you just add a unit test ? Just in case...

MRegeard:
> Thanks @MRegeard . This is indeed a good idea. Could you just add a unit test ? Just in case...

Done ! I could not find a way to get the value of vmin, vmax after plotting though ...

Astro-Kirsty (approved):

registerrier (approved):

Issue #4673, opened 2023-07-20
Add metadata class for EventList


registerrier:
Currently, the `EventList` meta data is stored on the `table` meta object. It should have its own `meta` class based on the  `MetaData` base class, following PIG 25 (#4491).

Metadata are partly duplicated from `Observation` attributes (but act here as meta info). They should contain:
- obs_id : Union[str, int]
- telescope : Optional[str] = None   # here optional
- instrument : Optional[str] = None
- observation_mode : Optional[str] = None
- observation_time_start : Optional[Time] = None
- observation_time_stop :  Optional[Time] = None
- observation_time_ref : Optional[Time] = None
- livetime : Time
- deadtime : float
- location : Optional[Union[str, EarthLocation]] = None
- creation : Optional[CreatorMetaData] = None

Are there other things? Should we use names close to current `EventList` properties, e.g. `observatory_earth_location` rather than `location`?

bkhelifi:
Thanks Régis...
What is exactly here observation_mode? For to be sure... 
There are also:
- the array_pointing: parallel, convergent, divergent? if yes, an angle is needed.
- the tracking_mode: pointing, drift-scan (fix alt/az)
- precision pointing boolean
 But should we have a metadata class associated to the HDU pointing? In this case, the 3 later should not be here
- some sky quality
 But should we have a metadata class associated to monitoring data?

A DataRelease name is probably needed: for the current HESS fits prod, we have a prod per analysis config type (ie cuts). But, in other hand, this information is a common metadata to many EventList, like instrument, location, creation... Should we create a `DataReleaseMetaClass`?

Maybe @kosack and @maxnoe can help here...

PR #4832, opened 2023-10-19
Introduce `EventListMetaData` Part 1/2

Astro-Kirsty:
**Description**
Part 1: This PR introduces the `EventListMetaData` described in #4673 

Part 2: will be deprecating the write function. PR #5059

This is connected to #5042.

### DIFF
--- a/gammapy/data/event_list.py
+++ b/gammapy/data/event_list.py
@@ -202,6 +202,7 @@ class EventList:
         """
         tables = [_.table for _ in event_lists]
         stacked_table = vstack_tables(tables, **kwargs)
+        log.warning("The meta information will be empty here.")
         return cls(stacked_table)
 
     def stack(self, other):
--- a/gammapy/data/event_list.py
+++ b/gammapy/data/event_list.py
@@ -114,7 +114,6 @@ class EventList:
         kwargs.setdefault("hdu", "EVENTS")
         table = Table.read(filename, **kwargs)
 
-        # build meta from the table.meta
         meta = EventListMetaData.from_header(table.meta)
 
         return cls(table=table, meta=meta)
@@ -221,16 +220,13 @@ class EventList:
         info = self.__class__.__name__ + "\n"
         info += "-" * len(self.__class__.__name__) + "\n\n"
 
-        # -----------------------------------  these are then read directly from the self.meta !
-
-        # instrument = self.table.meta.get("INSTRUME")
         instrument = self.meta.instrument
         info += f"\tInstrument       : {instrument}\n"
 
-        telescope = self.table.meta.get("TELESCOP")
+        telescope = self.meta.telescope
         info += f"\tTelescope        : {telescope}\n"
 
-        obs_id = self.table.meta.get("OBS_ID", "")
+        obs_id = self.meta.obs_id
         info += f"\tObs. ID          : {obs_id}\n\n"
 
         info += f"\tNumber of events : {len(self.table)}\n"
--- a/gammapy/data/event_list.py
+++ b/gammapy/data/event_list.py
@@ -18,6 +18,7 @@ from gammapy.utils.scripts import make_path
 from gammapy.utils.testing import Checker
 from gammapy.utils.time import time_ref_from_dict
 from .gti import GTI
+from .metadata import EventListMetaData
 
 __all__ = ["EventList"]
 
@@ -86,8 +87,9 @@ class EventList:
 
     """
 
-    def __init__(self, table):
+    def __init__(self, table=None, meta=None):
         self.table = table
+        self.meta = meta
 
     def _repr_html_(self):
         try:
@@ -111,7 +113,11 @@ class EventList:
         filename = make_path(filename)
         kwargs.setdefault("hdu", "EVENTS")
         table = Table.read(filename, **kwargs)
-        return cls(table=table)
+
+        # build meta from the table.meta
+        meta = EventListMetaData.from_header(table.meta)
+
+        return cls(table=table, meta=meta)
 
     def to_table_hdu(self, format="gadf"):
         """
@@ -215,7 +221,10 @@ class EventList:
         info = self.__class__.__name__ + "\n"
         info += "-" * len(self.__class__.__name__) + "\n\n"
 
-        instrument = self.table.meta.get("INSTRUME")
+        # -----------------------------------  these are then read directly from the self.meta !
+
+        # instrument = self.table.meta.get("INSTRUME")
+        instrument = self.meta.instrument
         info += f"\tInstrument       : {instrument}\n"
 
         telescope = self.table.meta.get("TELESCOP")
--- a/gammapy/data/event_list.py
+++ b/gammapy/data/event_list.py
@@ -58,7 +58,9 @@ class EventList:
     Parameters
     ----------
     table : `~astropy.table.Table`
-        Event list table.
+        Event list table
+    meta : `~gammapy.data.EventListMetaData`
+        The metadata. Default is None.
 
     Examples
     --------
@@ -87,7 +89,7 @@ class EventList:
 
     """
 
-    def __init__(self, table=None, meta=None):
+    def __init__(self, table, meta=None):
         self.table = table
         self.meta = meta
 
--- a/gammapy/data/event_list.py
+++ b/gammapy/data/event_list.py
@@ -58,7 +58,7 @@ class EventList:
     Parameters
     ----------
     table : `~astropy.table.Table`
-        Event list table
+        Event list table.
     meta : `~gammapy.data.EventListMetaData`
         The metadata. Default is None.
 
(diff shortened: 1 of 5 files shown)


Astro-Kirsty:
The `DEADC` does not exist for the Fermi `EventList` [i.e. this](https://docs.gammapy.org/1.1/tutorials/starting/overview.html#event-lists), so I guess maybe this parameter should also be optional?

gammapy/data/event_list.py
```diff
@@ -86,8 +87,9 @@ class EventList:
 
     """
 
-    def __init__(self, table):
+    def __init__(self, table=None, meta=None):
```
registerrier, line 90: Why the `table` can is None by default? There should not be a default argument here.

registerrier (commented):
Thanks @Astro-Kirsty . 
A very general comment here, this is too close to the `ObservationMetaData`. So we either have to use the same `meta` object on the two structures or we have to clarify better what is connected to `EventList` and what is connected to `Observation`.

gammapy/data/event_list.py
Astro-Kirsty: Oops, a mistake on my behalf

gammapy/data/event_list.py
```diff
@@ -86,8 +87,9 @@ class EventList:
...
         self.table = table
+        self.meta = meta
```
bkhelifi, line 92: This new parameter should be described in the docstring of the class

gammapy/data/event_list.py
```diff
@@ -109,7 +111,10 @@ def read(cls, filename, **kwargs):
         filename = make_path(filename)
         kwargs.setdefault("hdu", "EVENTS")
         table = Table.read(filename, **kwargs)
-        return cls(table=table)
+
```
bkhelifi, line 114: Can you add a test for the read function?

gammapy/data/metadata.py
```diff
@@ -162,3 +162,135 @@ def from_header(cls, events_hdr, format="gadf"):
         kwargs["optional"] = optional
 
         return cls(**kwargs)
+
+
+class EventListMetaData(MetaData):
```
bkhelifi, line 167: I do not see the point of such redundancy with the `ObservationMetaData`. Why not including it in this class?

gammapy/data/metadata.py
```diff
@@ -162,3 +162,135 @@ def from_header(cls, events_hdr, format="gadf"):
...
+        else:
+            raise ValueError("Incorrect time input value.")
+
+    @classmethod
+    def from_header(cls, events_hdr, format="gadf"):
+        """Create and fill the metadata from the event list metadata.
+
+        Parameters
+        ----------
+        format : str
```
bkhelifi, line 236: `event_hdr` should be described also...

Astro-Kirsty:
This has been updated as per discussions in the dev call. 

We don't want `EventListMetaData` to be a repeat of the `ObservationMetaData` and therefore only very limited amount of keywords can/should be defined. 

This is a first step to include metadata for v1.2 but further adjustments will be made for future versions. 

The options keywords: "DST_VER", "ANA_VER", "CAL_VER" could potentially be placed in another `gammapy.utils.metadata` class called "DataProcessingMetadata". 

Note: tests still need to be added

Please comment!

Astro-Kirsty:
The error is coming from the `test_dataset_hawc` in the `gammapy/makers/tests/test_map.py`. I am a little stumped
Note: all other tests run fine.

Can someone help in solving this?
@QRemy @adonath @LauraOlivera @AtreyeeS @registerrier @bkhelifi

gammapy/data/metadata.py
```diff
@@ -32,6 +32,13 @@
 METADATA_FITS_KEYS["observation"] = OBSERVATION_METADATA_FITS_KEYS
 
 
+EVENTLIST_METADATA_FITS_KEYS = {
+    "event_class": "EV_CLASS",
```
bkhelifi, line 36: what is the evt class?
In parallel, do we have event types? If yes, the evt types should be a list, in general. Is not it?

gammapy/data/event_list.py
```diff
@@ -191,9 +198,7 @@ def from_stack(cls, event_lists, **kwargs):
         Parameters
         ----------
         event_lists : list
-            List of `~gammapy.data.EventList` to stack.
-        **kwargs : dict, optional
-            Keyword arguments passed to `~astropy.table.vstack`.
+            List of `~gammapy.data.EventList` to stack
         """
         tables = [_.table for _ in event_lists]
         stacked_table = vstack_tables(tables, **kwargs)
```
registerrier, line 204: In this function, one would need to pass some meta data too. 
As far as I can, there was no real check on the `table.meta`. 
I assume that the correct behavior is to check that specific meta contents are consistent. 
Maybe simply raise a warning that `meta` will be empty here.

gammapy/data/metadata.py
registerrier: It depends if one event list is provided per event type or if event list is a general object with a single column.
The SDC will use the former, but in general we might be better off with the latter.

In all generality, it could be a list of `str` indeed.

gammapy/data/metadata.py
```diff
@@ -124,3 +131,35 @@ def from_header(cls, header, format="gadf"):
...
+    _tag: ClassVar[Literal["EventList"]] = "eventlist"
+    event_class: Optional[str] = None
+    creation: Optional[CreatorMetaData] = None
+    optional: Optional[dict] = None
+
+    @classmethod
+    def from_header(cls, header, format="gadf"):
+        meta = super(EventListMetaData, cls).from_header(header, format)
+
+        meta.creation = CreatorMetaData()
```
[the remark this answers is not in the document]
registerrier: Yes this should be read from the header.

registerrier (commented):
Thanks @Astro-Kirsty . This is nice. I have left a few comments.

gammapy/data/metadata.py
Astro-Kirsty: So you'd prefer the `event_class` to be a list of `str` instead of a `str`?

From the formatting we could actually have two fields here `EV_CLASS` (same for all events in a given event list) and `EVENT_TYPE` (can have a different value for each event).
https://gamma-astro-data-formats.readthedocs.io/en/v0.3/events/events.html#iact-events-class-type

gammapy/data/event_list.py
Astro-Kirsty: `vstack` has a `metadata_conflicts` argument with default set to `warn`.
So it should be handled within that?

I added a warning for `meta` being empty

gammapy/data/event_list.py
```diff
@@ -111,7 +115,10 @@ def read(cls, filename, **kwargs):
         filename = make_path(filename)
         kwargs.setdefault("hdu", "EVENTS")
         table = Table.read(filename, **kwargs)
-        return cls(table=table)
+
+        meta = EventListMetaData.from_header(table.meta)
+
+        return cls(table=table, meta=meta)
```
registerrier, line 121: This would need a test.

gammapy/data/event_list.py
registerrier: Maybe add a TODO as well

registerrier (commented):
Thansk @Astro-Kirsty . Can you rebase here? I have left an additional comment.

gammapy/data/metadata.py
AtreyeeS: Maybe we can start with the most minimal requirement for the SDC, and then allow lists in the future? Otherwise, if we decide otherwise, we would need to through the entire deprecation process?

gammapy/data/metadata.py
```diff
@@ -124,3 +131,34 @@ def from_header(cls, header, format="gadf"):
...
+    """
+
+    _tag: ClassVar[Literal["EventList"]] = "eventlist"
+    event_class: Optional[str] = None
+    creation: Optional[CreatorMetaData] = None
+    optional: Optional[dict] = None
+
+    @classmethod
+    def from_header(cls, header, format="gadf"):
+        meta = super(EventListMetaData, cls).from_header(header, format)
```
AtreyeeS, line 150: I think in python 3 `super` does not need arguments 
`meta = super().from_header(header, format)` should work

AtreyeeS (dismissed):
Thanks @Astro-Kirsty ! Looks good to me

gammapy/data/metadata.py
Astro-Kirsty: Yes that is true! 
Thanks @AtreyeeS 

I adapted in the other cases too

Astro-Kirsty:
This is ready for another review @registerrier, @AtreyeeS, @bkhelifi, @adonath

registerrier (approved):
Thanks @Astro-Kirsty . No further comment.

PR #4834, opened 2023-10-19
Adding prior stat sum to datasets

katrinstreil:
**Description**
This PR is to incorporate any priors which are set on the model parameters into the stat_sum of the dataset. 
This is based on the PIG #4381 and the prior definition from #4620.

This pull request is not yet ready for review. For all the tests to be successful #4620 has to be merged.

### DIFF
--- a/gammapy/datasets/core.py
+++ b/gammapy/datasets/core.py
@@ -68,13 +68,15 @@ class Dataset(abc.ABC):
             return self.mask_safe
 
     def stat_sum(self):
-        """Total statistic given the current model parameters."""
+        """Total statistic given the current model parameters and priors."""
         stat = self.stat_array()
-
         if self.mask is not None:
             stat = stat[self.mask.data]
-
-        return np.sum(stat, dtype=np.float64)
+        if self.models is not None:
+            prior_stat_sum = self.models.parameters.prior_stat_sum()
+        else:
+            prior_stat_sum = 0.0
+        return np.sum(stat, dtype=np.float64) + prior_stat_sum
 
     @abc.abstractmethod
     def stat_array(self):
--- a/gammapy/datasets/core.py
+++ b/gammapy/datasets/core.py
@@ -75,7 +75,7 @@ class Dataset(abc.ABC):
             stat = stat[self.mask.data]
         prior_stat_sum = 0.0
         if self.models is not None:
-            prior_stat_sum = self.models.parameters.prior_stat_sum()    
+            prior_stat_sum = self.models.parameters.prior_stat_sum()
         return np.sum(stat, dtype=np.float64) + prior_stat_sum
 
     @abc.abstractmethod
--- a/gammapy/datasets/core.py
+++ b/gammapy/datasets/core.py
@@ -73,10 +73,9 @@ class Dataset(abc.ABC):
 
         if self.mask is not None:
             stat = stat[self.mask.data]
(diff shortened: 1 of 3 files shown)


gammapy/datasets/map.py
```diff
@@ -1114,13 +1114,21 @@ def plot_residuals(
         return ax_spatial, ax_spectral
 
     def stat_sum(self):
-        """Total statistic function value given the current model parameters."""
+        """Total statistic function value given the current model parameters and priors."""
+        if self.models is not None:
+            prior_stat_sum = self.models.parameters.prior_stat_sum()
+        else:
+            prior_stat_sum = 0.0
```
registerrier, line 1121:
```suggestion
        prior_stat_sum = 0.0
        if self.models is not None:
            prior_stat_sum = self.models.parameters.prior_stat_sum()     
```

gammapy/datasets/core.py
```diff
@@ -68,13 +68,16 @@ def mask(self):
...
         stat = self.stat_array()
 
         if self.mask is not None:
             stat = stat[self.mask.data]
-
-        return np.sum(stat, dtype=np.float64)
+        if self.models is not None:
+            prior_stat_sum = self.models.parameters.prior_stat_sum()
+        else:
+            prior_stat_sum = 0.0
```
registerrier, line 79:
```suggestion
        prior_stat_sum = 0.0
        if self.models is not None:
            prior_stat_sum = self.models.parameters.prior_stat_sum()    
```

registerrier (commented):
Thanks @katrinstreil . This looks good but a simple unit test is missing!

gammapy/datasets/map.py
```diff
@@ -1114,13 +1114,20 @@ def plot_residuals(
         return ax_spatial, ax_spectral
 
     def stat_sum(self):
-        """Total statistic function value given the current model parameters."""
+        """Total statistic function value given the current model parameters and priors."""
+        prior_stat_sum = 0.0
+        if self.models is not None:
+            prior_stat_sum = self.models.parameters.prior_stat_sum()
```
adonath, line 1120: I just wanted to repeat the point I made in the dev call I while ago:

```
self.models.parameters.prior_stat_sum() 
```
Only works for prior that are on individual parameters. This means we will not be able to handle e.g. prior with off-diagonal elements in the covariance matrix. As it is e.g. need for the unfolding use-case. The `Models.parameters`
 attributed is re-created on access and thus cannot hold any state. Thus I I think it is important to move the prior stat sum to:
 
```
Models.prior_stat_sum()
```

adonath: However this could happen in a follow up PR, but it should happen before we advertise the new API.

katrinstreil: Yes, I agree this is a preliminary version to include only one-dimensional priors and should not be advertised yet. Only the drafted follow-up PR with multi-dimensional priors, the `PriorStatistics` and the attachment of the priors to the models should be.

registerrier (approved):
Thanks @katrinstreil . This looks good but there is a style issue. Can you make sure that black and flake8 pre-commit hooks are run before commiting?

registerrier (approved):
Thnaks @katrinstreil . Looks good! 
This is of course a temporary solution before the multi-dimensional priors are implemented.

PR #4835, opened 2023-10-19
Management of metadata for `Models`

bkhelifi:
**Description**

This pull request is the continuation of the closed #4819, after an issue of rebasing...

This pull request aims to deal with metadata for Models.

For the moment, models metadata are not kept in a dedicated metadata class. They are just written in file, and read when a basic check of the Gammapy version.

### DIFF
--- a/gammapy/modeling/models/core.py
+++ b/gammapy/modeling/models/core.py
@@ -12,10 +12,8 @@ import matplotlib.pyplot as plt
 import yaml
 from gammapy.maps import Map, RegionGeom
 from gammapy.modeling import Covariance, Parameter, Parameters
-from gammapy.utils.check import check_version
 from gammapy.utils.metadata import CreatorMetaData
 from gammapy.utils.scripts import make_name, make_path
-from gammapy.version import version
 
 __all__ = ["Model", "Models", "DatasetModels", "ModelBase"]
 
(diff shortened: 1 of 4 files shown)


QRemy:
Could we have the metadata in a separate file with only a reference in the model file (as we have for the covariance).
This would preserve the readability of the models files if the amount metadata grows later.

bkhelifi:
I guess that one should have the same scheme for all meta data, and not only for Models. Any thought, @registerrier

bkhelifi:
@registerrier : done

gammapy/modeling/models/core.py
```diff
@@ -431,6 +434,18 @@ def read(cls, filename):
     def from_yaml(cls, yaml_str, path=""):
         """Create from YAML string."""
         data = yaml.safe_load(yaml_str)
+        if "metadata" in data.keys():
+            creator_version = data["metadata"]["creator"].split()[1]
```
registerrier, line 438: I think this will cause issues because there is no guarantee that creator metadata will always be structured this way. `split()[1]` can result in unwanted failures if `split()` returns only one element.

registerrier (changes requested):
Thanks @bkhelifi .

On second thoughts, given the expected level of change in the metadata design, I prefer not to include a version checking mechanism for now. 
This is too early and is likely to cause issues in the future. We cannot be sure that version will stored as it is now. 
I can update the PR to do so.

registerrier (approved):
Thanks @bkhelifi ! Merging now.

PR #4836, opened 2023-10-19
Fix a division per zero in `TemplateSpatialModel` normalisation

bkhelifi:
**Description**

Everything is in the title.

### DIFF
--- a/gammapy/modeling/models/spatial.py
+++ b/gammapy/modeling/models/spatial.py
@@ -1292,6 +1292,12 @@ class TemplateSpatialModel(SpatialModel):
     ):
         if (map.data < 0).any():
             log.warning("Map has negative values. Check and fix this!")
+        if (map.data == 0.0).all():
+            log.warning("Map values are all zeros. Check and fix this!")
+        if not map.geom.is_image and (map.data.sum(axis=(1, 2)) == 0).any():
+            log.warning(
+                "Map values are all zeros in at least one energy bin. Check and fix this!"
+            )
 
         if filename is not None:
             filename = str(make_path(filename))
--- a/gammapy/modeling/models/spatial.py
+++ b/gammapy/modeling/models/spatial.py
@@ -1306,7 +1306,7 @@ class TemplateSpatialModel(SpatialModel):
                 # Normalize in each energy bin
                 data_sum = map.data.sum(axis=(1, 2)).reshape((-1, 1, 1))
 
-            data = map.data / data_sum
+            data = map.data / data_sum if data_sum > 0 else map.data * 0.0
             data /= map.geom.solid_angle().to_value("sr")
             map = map.copy(data=data, unit="sr-1")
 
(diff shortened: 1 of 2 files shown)


gammapy/modeling/models/spatial.py
```diff
@@ -1303,7 +1303,7 @@ def __init__(
                 # Normalize in each energy bin
                 data_sum = map.data.sum(axis=(1, 2)).reshape((-1, 1, 1))
 
-            data = map.data / data_sum
+            data = map.data / data_sum if data_sum > 0 else map.data * 0.0
```
morcuended, line 1306: Tests complained. I think it has to be `data_sum.any()`

morcuended (changes requested):

gammapy/modeling/models/spatial.py
```diff
@@ -1303,7 +1303,7 @@ def __init__(
                 # Normalize in each energy bin
                 data_sum = map.data.sum(axis=(1, 2)).reshape((-1, 1, 1))
 
-            data = map.data / data_sum
+            data = map.data / data_sum if data_sum.any() > 0 else map.data * 0.0
```
morcuended, line 1306: not strong opinion but maybe we should raise a warning in the `else` situation?

AtreyeeS: I agree. It is not clear what purpose an empty template can have, so it would be good to warn users that something is probably wrong with the input map

bkhelifi: With the safemask, we have always energy bin with no data. Should we warn systematically the users? Because at this end, it is expected in most case...

In fact, I saw this warning in one CI output. And in this case, there were no data at all... This fix allows to eliminate this case, but not all. And this does not change any result as Python takes care of the division per zero

QRemy: this won't work if `data_sum` is zero in one energy bin but not in the other, the fix should be :

`data = np.divide(map.data, data_sum, out=np.zeros_like(map.data), where=data_sum!=0)`

QRemy (changes requested):
Thanks @bkhelifi, I propose another fix :

QRemy:
> I would really recommend to not suppress this warning here. If the data is all zero the associated model parameter are not constrained. The fit will never converge. Users should know about this...

For that we can have a separate PR to add a check in the MapEvaluator such as `contributes` is False if the template data are all zeros in the evaluation geom.

adonath:
Thanks @QRemy, your proposal makes much more sense to me. I would still raise a warning, as the template would lead to zero predicted flux in affected energy ranges. And at this point one cannot know whether those are masked our correctly later in the likelihood computation.

adonath:
> For that we can have a separate PR to add a check in the MapEvaluator such as contributes is False if the template data are all zeros in the evaluation geom.

I agree this is a possible addition to achieve a save behavior for users, however I think still think it should not happen silently. Users should know, whether the input is wrong and ideally fox it in first place.

QRemy (approved):
@adonath as you suggested I added a warning for the cases where all values are zeros or all values are zeros in at least one bin.
@bkhelifi this should be ready to merge.

Astro-Kirsty:
https://github.com/gammapy/gammapy/pull/4836#discussion_r1368414566

Should we add a test to catch this sort of behaviour then?

QRemy:
> Should we add a test to catch this sort of behaviour then?

Added

PR #4837, opened 2023-10-19
Add checksum argument to gammapy products write functions

registerrier:
**Description**

This pull request adds a checksum argument to all write methods in gammapy. This should support fixity metadata requirements from CTAO. Right @kosack ?
For now the checksum is set to False by default.

**Dear reviewer**

This has tests now. 
Note that checksum verification on read must be added too. But this will be in a follow-up PR.

### DIFF
--- a/gammapy/data/observations.py
+++ b/gammapy/data/observations.py
@@ -557,7 +557,9 @@ class Observation:
             **irf_dict,
         )
 
-    def write(self, path, overwrite=False, format="gadf", include_irfs=True):
+    def write(
+        self, path, overwrite=False, format="gadf", include_irfs=True, checksum=False
+    ):
         """
         Write this observation into `path` using the specified format
 
@@ -571,6 +573,9 @@ class Observation:
             Output format, currently only "gadf" is supported
         include_irfs: bool
             Whether to include irf components in the output file
+        checksum : bool
+            When True adds both DATASUM and CHECKSUM cards to the headers written to the file.
+            Default is False.
         """
         if format != "gadf":
             raise ValueError(f'Only the "gadf" format is supported, got {format}')
@@ -599,7 +604,7 @@ class Observation:
                 if irf is not None:
                     hdul.append(irf.to_table_hdu(format="gadf-dl3"))
 
-        hdul.writeto(path, overwrite=overwrite)
+        hdul.writeto(path, overwrite=overwrite, checksum=checksum)
 
     def copy(self, in_memory=False, **kwargs):
         """Copy observation
--- a/gammapy/data/observations.py
+++ b/gammapy/data/observations.py
@@ -568,11 +568,11 @@ class Observation:
         path: str or `~pathlib.Path`
             Path for the output file
         overwrite: bool
-            If true, existing files are overwritten.
+            If true, existing files are overwritten. Default is False.
         format: str
-            Output format, currently only "gadf" is supported
+            Output format, currently only "gadf" is supported. Default is "gadf".
         include_irfs: bool
-            Whether to include irf components in the output file
+            Whether to include irf components in the output file. Default is True.
         checksum : bool
             When True adds both DATASUM and CHECKSUM cards to the headers written to the file.
             Default is False.
--- a/gammapy/data/tests/test_event_list.py
+++ b/gammapy/data/tests/test_event_list.py
@@ -3,6 +3,7 @@ import pytest
 from numpy.testing import assert_allclose
 from astropy import units as u
 from astropy.coordinates import SkyCoord
+from astropy.io import fits
 from astropy.table import Table
 from regions import CircleSkyRegion, RectangleSkyRegion
 from gammapy.data import GTI, EventList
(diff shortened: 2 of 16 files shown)


gammapy/datasets/flux_points.py
```diff
@@ -149,7 +153,17 @@ def write(self, filename, overwrite=False, **kwargs):
...
+        filename = make_path(filename)
+
+        if "fits" in filename.suffixes:
+            primary_hdu = fits.PrimaryHDU()
+            hdu_table = fits.BinTableHDU(table, name="TABLE")
+            fits.HDUList([primary_hdu, hdu_table]).writeto(
+                filename, overwrite=overwrite, checksum=checksum
+            )
+        else:
+            table.write(make_path(filename), overwrite=overwrite, **kwargs)
```
morcuended, line 166: codecov complain about these lines not being covered?

gammapy/datasets/tests/test_map.py
```diff
@@ -1246,6 +1247,18 @@ def test_map_datasets_on_off_fits_io(images, tmp_path):
     assert_allclose(dataset.mask_safe, dataset_new.mask_safe)
 
 
+@requires_data()
+def test_map_datasets_on_off_checksum(images, tmp_path):
+    dataset = get_map_dataset_onoff(images)
+    Datasets([dataset]).write(tmp_path / "test.yaml", checksum=True)
+    file = open(tmp_path / "test.yaml")
+    print(file.read())
```
morcuended, line 1255: is this print needed?

morcuended (commented):
I see it is still WIP. I just leave here a few comments

gammapy/maps/core.py
```diff
@@ -416,9 +416,15 @@ def write(self, filename, overwrite=False, **kwargs):
...
             This option is only compatible with the 'gadf' format.
+        checksum : bool
+            When True adds both DATASUM and CHECKSUM cards to the headers written to the file.
+            Default is False.
         """
+        checksum = kwargs.pop("checksum", False)
         hdulist = self.to_hdulist(**kwargs)
-        hdulist.writeto(str(make_path(filename)), overwrite=overwrite)
+        hdulist.writeto(
+            str(make_path(filename)), overwrite=overwrite, checksum=checksum
```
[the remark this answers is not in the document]
registerrier: Indeed thanks!

gammapy/datasets/tests/test_map.py
registerrier: no and the line above neither. Thanks :)

gammapy/datasets/flux_points.py
registerrier: yes, we have nowhere a test to export a fluxpoints dataset to an ascii table. In theory it was possible, but untested. I left this possibility open. But you're right it would need a test to check.

Astro-Kirsty:
Just a general comment, we should try to be consistent with the docstring for `overwrite`
We currently have multiple wordings throughout the code:

- Overwrite file if it exists.
- Overwrite file on disk if present
- Overwrite

Plus many more!

gammapy/data/observations.py
```diff
@@ -557,7 +557,9 @@ def read(cls, event_file, irf_file=None):
             **irf_dict,
         )
 
-    def write(self, path, overwrite=False, format="gadf", include_irfs=True):
+    def write(
+        self, path, overwrite=False, format="gadf", include_irfs=True, checksum=False
```
bkhelifi, line 561: Can you add in the docstring also the default value of the parameters?

gammapy/data/tests/test_event_list.py
```diff
@@ -82,6 +83,12 @@ def test_write(self):
             dummy_events.table.meta["HDUCLAS1"] = "events"
             dummy_events.write("test.fits", overwrite=True)
 
+    def test_write_checksum(self):
+        self.events.write("test.fits", overwrite=True, checksum=True)
+        hdu = fits.open("test.fits")["EVENTS"]
+        assert "CHECKSUM" in hdu.header
```
bkhelifi, line 89: Can we test also the values of the CHECKSUM and DATASUM? One should be aware when any change will happen

gammapy/datasets/core.py
```diff
@@ -420,7 +420,12 @@ def read(cls, filename, filename_models=None, lazy=True, cache=True):
         return datasets
 
     def write(
-        self, filename, filename_models=None, overwrite=False, write_covariance=True
+        self,
```
bkhelifi, line 423: Can you profit of this PR to add the default value of the parameters in the docstring?

registerrier:
> Just a general comment, we should try to be consistent with the docstring for `overwrite` We currently have multiple wordings throughout the code:
> 
> * Overwrite file if it exists.
> * Overwrite file on disk if present
> * Overwrite
> 
> Plus many more!

For this we need a specific PR to solve all issues at once.

gammapy/data/tests/test_event_list.py
registerrier: I'd rather test that when reading after writing everything is fine. Testing the exact checksum/datasum will break if anything changes in the file content without tolerance.

gammapy/datasets/core.py
registerrier: done

gammapy/data/observations.py
registerrier: done

bkhelifi (approved):
Thanks a lot @registerrier

PR #4839, opened 2023-10-19
No extrapolation for `PrimaryFlux`

StFroese:
I removed the extrapolation in energy for the `PrimaryFlux` since it doesn't makes sense extrapolate beyond the limit of the mass of the particle. We also don't know if we can extrapolate to zero.

### DIFF
--- a/gammapy/astro/darkmatter/spectra.py
+++ b/gammapy/astro/darkmatter/spectra.py
@@ -108,7 +108,8 @@ class PrimaryFlux(TemplateNDSpectralModel):
             region=None, axes=[log10x_axis, mass_axis], data=self.table[channel_name]
         )
 
-        super().__init__(region_map)
+        interp_kwargs = {"extrapolate": True, "fill_value": 0, "values_scale": "lin"}
+        super().__init__(region_map, interp_kwargs=interp_kwargs)
         self.mDM = mDM
         self.mass.frozen = True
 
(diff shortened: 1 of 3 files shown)


registerrier (commented):
Thanks @StFroese . Sorry for the late review here.
Is this really removing extrapolation? 
Apparently this is changing test values by more than the expected tolerance. Is this expected?

StFroese:
Hi @registerrier, sry for the late reply. I see, it is very confusing 😄 
1. We actually have to activate extrapolation, hence `extrapolate=True`, since deactivating it will result an out-of-bounds error when you want to integrate the spectrum for example and you don't know the smallest energy that is allowed.
2. The `TemplateNDSpectralModel` uses the `interp_by_pixel` method of the RegionNDMap, which therefore uses the `ScaledRegularGridInterpolator`. The problem is that the default for `values_scale` is `'log'`. So the flux values are interpolated and extrapolated on the grid of `log(values)`. The `__call__` method of the interpolator evaluates the grid at the point of interest and uses the inverse of the scale to produce the inter/extrapolated value `exp(value)`. The problem is that we can't extrapolate with `0`'s in this case since `exp(x)=0` doesn't have a solution.

Solution:
Change `values_scale` to `lin` and use `fill_value=0`. 
I think the tests fail because of change from log to linear space for the interpolation. The test actually raise with a very small absolute difference in the values, i.e. 
```python
args = (<function assert_allclose.<locals>.compare at 0x12ca3d4c0>, array([9.3319318e-05]), array(9.328234e-05))
```

Let me also include two plots to show the extrapolation on this branch (channel='W', mass='1 TeV'):
on main:
gammapy_pr_master.pdf

on this branch:
gammapy_pr.pdf

registerrier (approved):
Thanks @StFroese . This looks good.

PR #4844, opened 2023-10-20
Fixing observation copy example

MRegeard:
This PR fix the example code of `Observation.copy`.


registerrier (approved):
Thanks @MRegeard . 
Could we make sure the doctest is run of this part?

PR #4849, opened 2023-10-20
Add TimeMapAxis.to_gti()

AtreyeeS:
Since time handling methods (intersection, union, etc) mostly reside on GTI, it is useful to be able to convert back and forth between time axis and GTIs. This is necessary, eg: in #4828 to find the intersection of the time bin with the GTIs

### DIFF
--- a/gammapy/maps/axes.py
+++ b/gammapy/maps/axes.py
@@ -2891,6 +2891,20 @@ class TimeMapAxis:
             name=name,
         )
 
+    def to_gti(self):
+        """Convert the axis to a GTI.
+
+        Returns
+        -------
+        gti : `GTI`
+        GTI table
+        """
+        from gammapy.data import GTI
+
+        return GTI.create(
+            self.edges_min, self.edges_max, reference_time=self.reference_time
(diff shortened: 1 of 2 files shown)


cgalelli (dismissed):
Thanks @AtreyeeS, this is a very useful addition. Everything seems linear and fine to me!

gammapy/maps/axes.py
```diff
@@ -2891,6 +2891,20 @@ def from_time_bounds(cls, time_min, time_max, nbin, unit="d", name="time"):
             name=name,
         )
 
+    def to_gti(self):
+        """Convert the axis to a GTI.
+
+        Returns
+        -------
+        gti : `GTI`
+        GTI table
```
adonath, line 2900:
```suggestion
            GTI table
```

adonath (approved):
Thanks @AtreyeeS! While this is effectively a one line implementation, I would still agree that this leads to slightly more readable code. So OK to merge from my side.

cgalelli (approved):

PR #4851, opened 2023-10-21
Parallel support for WcsNDMap map convolution

QRemy:
This PR replaces the loop over multiple images/kernels  by a call to run_multiprocessing so the convolution can be executed in parallel using the context manager introduced in #4845

### DIFF
--- a/gammapy/maps/wcs/ndmap.py
+++ b/gammapy/maps/wcs/ndmap.py
@@ -1,5 +1,6 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 import logging
+from itertools import repeat
 import numpy as np
 import scipy.interpolate
 import scipy.ndimage as ndi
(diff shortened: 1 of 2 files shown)


registerrier (approved):

I assume one has to wait for #4845 to be merged to implement a test for this.

gammapy/maps/wcs/ndmap.py
```diff
@@ -871,18 +869,34 @@ def convolve(self, kernel, method="fft", mode="same"):
...
+                repeat(mode),
+            ),
+            task_name="Convolution",
+        )
+        data = np.empty(geom.data_shape, dtype=np.float32)
+        for idx_res, idx in enumerate(indexes):
+            data[idx] = convolved[idx_res]
         return self._init_copy(data=data, geom=geom)
 
+    def _convolve(self, image, kernel, method, mode):
```
QRemy, line 896:
```suggestion
        @staticmethod
        def _convolve(image, kernel, method, mode):
```

registerrier:
Merging now.

Issue #4792, opened 2023-10-11
Define and add metadata on MapDataset


registerrier:
MapDataset metadata need a specific `MapDatasetMetaData` container. We currently have a minimal `meta_table` that is not serialized. It should be replaced by a proper container with well defined entries and tested serialization.

A difficulty specific to `MapDataset` is that we have to design a mechanism for stacking the metadata. We have to understand how each stored information can be stacked.  Currently the `meta_table` is stacked horizontally which relies on `np.stack` for each column.  
A similar approach could be used if one stores the metadata attributes as arrays. But this makes validation more problematic.  Any opinion @adonath ?

Currently, we keep track of:
- telescope : str
- obs_id : str
- pointing coordinates : SkyCoord

There are much more information that we can transfer from the parent observation metadata if they exist there such as:
- instrument : str
- observation_mode : str
- event_type : str
- others?

Currently the proposition from the `ObservationMetadata` PR (#4738) is to have a `dict` for `optional` metadata. It is not clear if we can have this here as well. In particular w.r.t. stacking.

There is also the PR on livetime map (#4407), and a possibility to is have it as a meta attribute. 

Opinions, suggestions @adonath , @bkhelifi @QRemy @AtreyeeS and others?

adonath:
@registerrier I think `pydantic` deals just fine with with lists of objects. So indeed the design can be, that attributes such as `.telescope` have a type of `Union[str, list[str]]`. The only risk I see is performance when we get very large lists and maybe some inconvenience of accessing specific meta data.

bkhelifi:
Here is coming the question of metadata reduction...
- The `instrument` is obviously to be added... as well as `observation_mode`, `event_type`
- For `telescope`, I think that one should totally avoid to repeat N times this parameter. If there is a mixture of sub-arrays, one should have only one instance for each sub-array. The  list of `obs_id` will allow to know exactly the relative weight... 
- From the obs_id one can have the observation times, but I am wondering whether one should not have a TSTART and TSTOP at least. What do you think @kosack ? If there is a GTI selection, I am also wondering if one should not add it, as well as a PHASE selection
- The livetime map has to be added also, good point! Otherwise it is hard to rebuild it... PS: the livetime map is corrected from the radial acceptance -> its name should be more explicit, e.g. effective livetime or somethink else....

Issue #4791, opened 2023-10-10
Adding metadata on Datasets


registerrier:
If each dataset contain its own metadata, we might nevertheless need some metadata for the `Datasets` object. 

I think it does not need to be more than the creation information (gammapy version and date) and can be only added at serialization and not on the in-memory object. 

Opinions?

bkhelifi:
For Datasets, indeed, I do not yet see what else to add... knowing that each of the associated `Dataset`  will contain metadata describing what it is... The hardest is on this case class

bkhelifi:
@registerrier I think that one can close this issue, as we have created the associated metadata class.
The closing of this issue will then close th project `Introduce metadata handling in Gammapy'

PR #4853, opened 2023-10-24
Add MapDatasetMetaData container

AtreyeeS:
Addresses  #4792. 

This PR adds a meta container on the MapDataset object. 

1. The stacking is not done in place currently. 
2. The `optional` dict keys are just stacked - and then users can decide what to do with it 
3. Stacking multiple instruments raises a warning - change to error and forbid?
4. The GTI table is already present on the map dataset, not replicating it here
5. Maybe another property `maker` can be added? To know how this dataset was created, eg: the outcome of some maker process or created by hand by the user. Or is that provenance information @bkhelifi 
6. In https://github.com/gammapy/gammapy/blob/main/gammapy/maps/utils.py#L77 we define INVALID_VALUE.int = np.nan
However, `isinstance(np.nan, int)` returns `False`, which raises `ValidationError`. I have put -999 as default for the event types, any better solution?

opinions please @registerrier @adonath @QRemy @maxnoe 

Still needs in future PRs:

- [ ] Deprecate the meta table?
- [ ] MapDatasetMaker.create_metadata()
- [ ] FluxPointsDataset meta object

### DIFF
--- a/gammapy/datasets/metadata.py
+++ b/gammapy/datasets/metadata.py
@@ -14,11 +14,8 @@ __all__ = ["MapDatasetMetaData"]
 
 MapDataset_METADATA_FITS_KEYS = {
     "MapDataset": {
-        "creation": "CREATION",
         "event_types": "EVT_TYPE",
         "optional": "OPTIONAL",
-        "pointing": "POINTING",
-        "obs_info": "OBS_INFO",
     },
 }
 
--- a/gammapy/datasets/metadata.py
+++ b/gammapy/datasets/metadata.py
@@ -1,4 +1,3 @@
-import logging
 from typing import ClassVar, Literal, Optional, Union
 from pydantic import ConfigDict
 from gammapy.utils.metadata import (
@@ -61,51 +60,7 @@ class MapDatasetMetaData(MetaData):
     event_type: Optional[Union[str, list[str]]] = None
     optional: Optional[dict] = None
 
-    def _stack_linear(self, obj_name, other, kwargs):
-        obj = getattr(self, obj_name)
-        if obj:
-            if not isinstance(obj, list):
-                obj = [obj]
-            obj.append(getattr(other, obj_name))
-        else:
-            obj = getattr(other, obj_name)
-        kwargs[obj_name] = obj
-        return kwargs
-
-    def _stack_unique(self, obj_name, other, kwargs):
-        obj = getattr(self, obj_name)
-        obj_other = getattr(other, obj_name)
-        if obj:
-            if not isinstance(obj, list):
-                obj = [obj]
-            if obj_other not in obj:
-                obj.append(obj_other)
-        else:
-            obj = obj_other
-        kwargs[obj_name] = obj
-        return kwargs
-
     def stack(self, other):
         kwargs = {}
-        kwargs["creation"] = self.creation
-        linear_stack = ["pointing", "event_type", "obs_ids"]
-        for i in linear_stack:
-            kwargs = self._stack_linear(i, other, kwargs)
-        unique_stack = ["observation_mode", "instrument", "telescope"]
-        for i in unique_stack:
-            if i == "instrument":
-                if self.instrument != other.instrument:
-                    logging.warning(
-                        f"Stacking data from different instruments {self.instrument} and {other.instrument}"
-                    )
-            kwargs = self._stack_unique(i, other, kwargs)
-
-        if self.optional:
-            optional = self.optional
-            for k in other.optional.keys():
-                if not isinstance(optional[k], list):
-                    optional[k] = [optional[k]]
-                optional[k].append(other.optional[k])
-            kwargs["optional"] = optional
-
+        kwargs["creation"] = CreatorMetaData()
         return self.__class__(**kwargs)
--- a/gammapy/datasets/metadata.py
+++ b/gammapy/datasets/metadata.py
@@ -6,6 +6,7 @@ from gammapy.utils.metadata import (
     METADATA_FITS_KEYS,
     CreatorMetaData,
     MetaData,
+    ObsInfoMetaData,
     PointingInfoMetaData,
 )
 
@@ -14,13 +15,9 @@ __all__ = ["MapDatasetMetaData"]
 MapDataset_METADATA_FITS_KEYS = {
     "MapDataset": {
         "creation": "CREATION",
-        "instrument": "INSTRUM",
-        "telescope": "TELESCOP",
-        "observation_mode": "OBS_MODE",
-        "pointing": "POINTING",
-        "obs_id": "OBS_ID",
         "event_types": "EVT_TYPE",
         "optional": "OPTIONAL",
+        "pointing": "POINTING",
     },
 }
 
@@ -28,24 +25,18 @@ METADATA_FITS_KEYS.update(MapDataset_METADATA_FITS_KEYS)
 
 
 class MapDatasetMetaData(MetaData):
-    """Metadata containing information about the GTI.
+    """Metadata containing information about the Dataset.
 
     Parameters
     ----------
     creation : `~gammapy.utils.CreatorMetaData`, optional
          The creation metadata.
-    instrument : str
-        the instrument used during observation.
-    telescope : str
-        The specific telescope subarray.
-    observation_mode : str
-        observing mode.
-    pointing : ~astropy.coordinates.SkyCoord
-        Telescope pointing direction.
-    obs_id : int
-        Observation id in the dataset.
-    event_types : int
+    obs_info : list of `~gammapy.utils.ObsInfoMetaData`
+        info about the observation.
+    event_types : list of int or str
         Event types used in analysis.
+    pointing: list of `~gammapy.utils.PointingInfoMetaData`
+        Telescope pointing directions.
     optional : dict
         Additional optional metadata.
     """
@@ -54,21 +45,18 @@ class MapDatasetMetaData(MetaData):
 
     _tag: ClassVar[Literal["MapDataset"]] = "MapDataset"
     creation: Optional[CreatorMetaData] = CreatorMetaData()
-    instrument: Optional[Union[str, list[str]]] = None
-    telescope: Optional[Union[str, list[str]]] = None
-    observation_mode: Optional[Union[str, list]] = None
+    obs_info: Optional[Union[ObsInfoMetaData, list[ObsInfoMetaData]]] = None
     pointing: Optional[Union[PointingInfoMetaData, list[PointingInfoMetaData]]] = None
-    obs_id: Optional[Union[str, list[str]]] = None
     event_type: Optional[Union[str, list[str]]] = None
     optional: Optional[dict] = None
 
     def stack(self, other):
         kwargs = {}
-        kwargs["creation"] = CreatorMetaData()
+        kwargs["creation"] = self.creation
         return self.__class__(**kwargs)
 
     @classmethod
-    def from_meta_table(cls, table):
+    def _from_meta_table(cls, table):
         """Create MapDatasetMetaData from MapDataset.meta_table
 
         Parameters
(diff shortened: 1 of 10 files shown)


bkhelifi:
Thanks a lot, @AtreyeeS .
About 2, it is not clear how the metadata reduction will be made. In order to avoid an exponential growing of metadata, I would not stack them per default now. I propose to think about them later when exchanging about metata reduction (with @kosack and @maxnoe ).
About 3, I would have say that this is forbidden, when instrument is not None. I do not see a reasonable use case for which it will be possible. However, one could change this behaviour in a future PR.
About 5, I think that this is indeed a provenance information.

For information, I am working on the `FluxMetaData` container. Once, we will have the main containers of the DL3 and DL4, one could work on the makers... the same for the estimators when we will have the main one for the DL4 and DL5.

gammapy/datasets/metadata.py
```diff
@@ -0,0 +1,157 @@
...
+    ----------
+    creation : `~gammapy.utils.CreatorMetaData`
+        the creation metadata
+    instrument : str
+        the instrument used during observation
+    telescope : str
+        The specific telescope subarray
+    observation_mode : str
+        observing mode
+    pointing : ~astropy.coordinates.SkyCoord
```
bkhelifi, line 24: I think that this is not proper to a dataset, but to an individual observation. Is not it?

gammapy/datasets/metadata.py
```diff
@@ -0,0 +1,157 @@
...
+        Telescope pointing direction
+    obs_ids : int
+        Observation ids stacked in the dataset
+    event_types : int
```
bkhelifi, line 28: What is missing is the origin of the data, the data release name associated to the dir of the datastore.

bkhelifi, line 28: There are other types of selection that can be made beyond the event types, the zen, the gti, the azimuth... I am wondering how and where one can 'store' this information in the workflow. Any idea?

gammapy/datasets/metadata.py
```diff
@@ -0,0 +1,157 @@
...
+        Event types used in analysis
+    optional : dict
```
bkhelifi, line 30: I share your suggestion to not store the individuals GTIs, but I am wondering whether a TSTART and a TSTOP can not be still useful. Any thought?

gammapy/datasets/metadata.py
```diff
@@ -0,0 +1,157 @@
...
+from typing import Optional, Union
+import numpy as np
+from astropy.coordinates import SkyCoord
+from pydantic import ValidationError, validator
+from gammapy.utils.metadata import CreatorMetaData, MetaData
+
+__all__ = ["MapDatasetMetaData"]
+
+
+class MapDatasetMetaData(MetaData):
```
bkhelifi, line 11: Should we not create a class for the Dataset object? and having daughter classes with their specificities? This would avoid code duplication...

gammapy/datasets/metadata.py
```diff
@@ -0,0 +1,157 @@
...
+    creation: Optional[CreatorMetaData]
+    instrument: Optional[str]
+    telescope: Optional[Union[str, list[str]]]
+    observation_mode: Optional[Union[str, list]]
+    pointing: Optional[Union[SkyCoord, list[SkyCoord]]]
+    obs_ids: Optional[Union[int, list[int]]]
+    event_type: Optional[Union[int, list[int]]]
+    optional: Optional[dict]
+
+    @validator("creation")
```
registerrier, line 43: I am not sure you need a validation here.
Technically, in general, I think the `CreatorMetaData` can be None until the object is serialized (unless the information is read from disk).

gammapy/datasets/metadata.py
```diff
@@ -0,0 +1,157 @@
...
+        if v is None:
+            return CreatorMetaData.from_default()
+        elif isinstance(v, CreatorMetaData):
+            return v
+        else:
+            raise ValidationError(
+                f"Incorrect pointing. Expect CreatorMetaData got {type(v)} instead."
+            )
+
+    @validator("instrument")
```
registerrier, line 54: Here no validation is needed either as `str` is a standard type.

gammapy/datasets/metadata.py
```diff
@@ -0,0 +1,157 @@
...
+    optional : dict
+        Any other meta information
+    """
+
+    creation: Optional[CreatorMetaData]
+    instrument: Optional[str]
+    telescope: Optional[Union[str, list[str]]]
+    observation_mode: Optional[Union[str, list]]
+    pointing: Optional[Union[SkyCoord, list[SkyCoord]]]
+    obs_ids: Optional[Union[int, list[int]]]
```
registerrier, line 39: Are we sure `obs_id` can only be `int`?

gammapy/datasets/metadata.py
```diff
@@ -0,0 +1,157 @@
...
+            return v
+        else:
+            raise ValidationError(
+                f"Incorrect pointing. Expect SkyCoord got {type(v)} instead."
+            )
+
+    @validator("obs_ids", "event_type")
+    def validate_obs_ids(cls, v):
+        if v is None:
+            return -999
```
registerrier, line 94: Why would you use a value like -999. It could stay None. No?

registerrier (commented):
Thanks @AtreyeeS .

Maybe a very general comment first. We already see that informations are replicated between many metadata objects.  Also, it seems useless to validate the same information in many different places.

This means that we need a higher granularity in metadata structures. Our `ObjectMetaData` classes will then mostly contain lower level metadata objects + some informations specific to the object.

Then stacking would mostly be building a list of e.g. `ObsInfoMetaData` and `EventMetaData`. 

It would be good to also identify entries that are specific to the `Dataset`, if there are. 
Maybe the `name`, possibly the `target_name`

gammapy/datasets/metadata.py
AtreyeeS: I thought `MapDataset.empty()` can add the `CreatorMetaData` and the rest of the information filled in by the `MapDatasetMaker`

gammapy/datasets/metadata.py
AtreyeeS: I had the same question about event types as well.

gammapy/datasets/metadata.py
AtreyeeS: Yes, this I will remove. I wanted to be sure we forbid a list of strings here

gammapy/datasets/metadata.py
AtreyeeS: Or a more general `ObjectMetaData` as @registerrier suggested

gammapy/datasets/metadata.py
AtreyeeS: I guess it depends a bit on how this meta information will be used

gammapy/datasets/metadata.py
AtreyeeS: This will be copied from the `ObservationMetaData`?

AtreyeeS:
I updated to use pydantic 2.0. 

I think the present stacking logic is too repetitive. Can it be simplified @adonath ?

adonath:
Thanks a lot @AtreyeeS! Probably the following code structure is simpler?
```python
from pydantic import BaseModel, RootModel
from typing import List, Optional


class MapDatasetMetaDataItem(BaseModel):
    name: str
    event_type: Optional[str] = None


class MapDatasetMetaData(RootModel):
    root: List[MapDatasetMetaDataItem]
    instrument: Optional[str] = None
    creation: ...

    def __iter__(self):
        return iter(self.root)

    def __getitem__(self, item):
        # could use the obs id here
        return self.root[item]
    
    def to_table(self):
        pass

    def from_table(self):
        pass
        
    def stack(self, item):
        # just append the list here
        pass


meta = StackedMapDatasetMetaData(
    [
        {"name": "hess"},
        {"name": "magic"},
        {"name": "veritas"},
        {"name": "fact"},
        {"name": "cta"},
    ]
)

meta[3].name
```
Having parallel lists is usually not a good idea, as it leads to these repetitive code statements. Introducing two separate classes might be the better solution here.

bkhelifi:
If we follow the ideaof @adonath , maybe one should keep the same logics of naming scheme between the class and the metadata class, ie:
MapDatasetMetaData, MapDatasetsMetaData (then leading to a change of the name of the python file)

In the list of instruments, one could add hawc, fermilat, astri....

Otherwise, I am wondering whether this class with this content is more a DatasetMetaData class... It seems very generic to all dataset child classes. It not it?

AtreyeeS:
I added a `from_meta_table` function to create a meta object from a meta table, and added computation of the meta in makers.

So the current behaviour is that there is a mechanism 

1. to compute the meta container from a meta table, and the meta table keeps information on all dataset operations (stacking, to_image, slicing etc). The meta container just contains the creation info.
2. Create the meta container during data reduction using the meta table

registerrier (commented):
Thanks @AtreyeeS . I have left a few comments.

AtreyeeS:
This now uses the `ObsInfoMetaData`.
Ready for final review from side

gammapy/datasets/metadata.py
```diff
@@ -0,0 +1,103 @@
...
+)
+
+__all__ = ["MapDatasetMetaData"]
+
+MapDataset_METADATA_FITS_KEYS = {
+    "MapDataset": {
+        "creation": "CREATION",
+        "event_types": "EVT_TYPE",
+        "optional": "OPTIONAL",
+        "pointing": "POINTING",
```
AtreyeeS, line 20: without adding `pointing`, `to_header` fails if `altaz` is not present

gammapy/datasets/tests/test_metadata.py
```diff
@@ -0,0 +1,131 @@
...
+    assert_allclose(meta.pointing[1].radec_mean.ra.deg, 83.1287)
+    assert meta.obs_info[0].obs_id == "111"
+    assert meta.obs_info[1].obs_id == "112"
+    assert meta.optional is None
+    assert meta.event_type is None
+
+
+def test_mapdataset_metadata_stack():
+    input1 = {
+        "obs_info": ObsInfoMetaData(**{"obs_id": 111}),
```
registerrier, line 91:
```suggestion
        "obs_info": ObsInfoMetaData(obs_id=111),
```
might be simpler

registerrier (approved):
Thanks @AtreyeeS . Looks good to me! No further comment.

AtreyeeS:
I checked that the meta information is not serialised with the dataset object.
I can add that in a following PR.

gammapy/utils/metadata.py
```diff
@@ -193,9 +193,10 @@ class ObsInfoMetaData(MetaData):
         The observation mode.
     """
 
+    model_config = ConfigDict(coerce_numbers_to_str=True)
     _tag: ClassVar[Literal["obs_info"]] = "obs_info"
 
-    obs_id: int
+    obs_id: str
```
registerrier, line 199: Why is it changed to `str`?

AtreyeeS: We discussed previously that obs_ids don't need to be necessarily integers, and can be strings, and thus the MapDatasetMetaData was taking strings for obs_id

Having a `str` accepts both `int` and `str` on input

registerrier: Currently the gadf requirement is that this is an int. That's why I kept this constrain when introducing the `ObsInfoMetaData`. 
Unless there is a strong constraint for the `MapDatasetMetaData` I'd rather keep the same solution and update later if necessary. No?

AtreyeeS: I have no opinions on this.. I can then change the behaviour for the MapDataset as well...

AtreyeeS: @bkhelifi The current implementation allows to read `int` as seen in the tests

registerrier: Indeed, but pydantic is changing the type to `str`. 
For now, I think it is simpler to stick to `int`.

bkhelifi:
A test is failing:
FAILED ../../.tox/py311-test-alldeps_noray/lib/python3.11/site-packages/gammapy/data/tests/test_observations.py::test_observation_read - AssertionError: assert '20136' == 20136
 +  where '20136' = <gammapy.data.observations.Observation object at 0x7febafbb75d0>.obs_id

gammapy/datasets/metadata.py
registerrier: See #5075 and #5076.

AtreyeeS:
1. reverted obs_id to int
2. a minimal serialisation of the meta implemented by extending the PRIMARY header
3. the serialisation does not work correctly when `obs_info` is a `list of ObsInfoMetaData` - I can try to fix that for 1.3

registerrier (approved):
Thanks @AtreyeeS . This looks good! No further comment from my side.

registerrier:
I have corrected an issue with `PointingInfoMetaData.from_header()` directly on your branch @AtreyeeS . The issue was that the default value was set to `SkyCoord(np.nan, no.nan)` which was breaking fits serialization. 
This is now solved with default values set to `None` if the header keywords are not defined.

registerrier (approved):
Thanks @AtreyeeS . Let's wait for the CI to run and merge.

bkhelifi:
This PR is associated to the issue #4791

PR #4869, opened 2023-10-27
Adapt overwrite in docstring to be consistent

### LINKED ISSUE #4858: Consistency with the docstring for `overwrite`
Just a general comment, we should try to be consistent with the docstring for `overwrite`
We currently have multiple wordings throughout the code:

- Overwrite file if it exists.
- Overwrite file on disk if present
- Overwrite

Plus many more!

_Originally posted by @Astro-Kirsty in https://github.com/gammapy/gammapy/issues/4837#issuecomment-1772495483_
            
We should decided upon the wording in this issue, then I can adapt the docstrings

Astro-Kirsty:
This PR is to fix #4858

### DIFF
--- a/gammapy/analysis/core.py
+++ b/gammapy/analysis/core.py
@@ -271,7 +271,7 @@ class Analysis:
         Parameters
         ----------
         overwrite : bool, optional
-            Overwrite existing FITS file. Default is True.
+            Overwrite existing file. Default is True.
         write_covariance : bool
             save covariance or not
         """
--- a/gammapy/analysis/core.py
+++ b/gammapy/analysis/core.py
@@ -271,7 +271,7 @@ class Analysis:
         Parameters
         ----------
         overwrite : bool, optional
-            Overwrite existing file. Default is False.
+            Overwrite existing FITS file. Default is True.
         write_covariance : bool
             save covariance or not
         """
--- a/gammapy/analysis/core.py
+++ b/gammapy/analysis/core.py
@@ -270,8 +270,8 @@ class Analysis:
 
         Parameters
         ----------
-        overwrite : bool
-            overwrite datasets FITS files
+        overwrite : bool, optional
+            Overwrite existing file. Default is False.
         write_covariance : bool
             save covariance or not
         """
--- a/gammapy/datasets/core.py
+++ b/gammapy/datasets/core.py
@@ -436,7 +436,7 @@ class Datasets(collections.abc.MutableSequence):
         filename_models : str or `Path`
             File path or name of models yaml file. Default is None.
         overwrite : bool, optional
-            Overwrite existing file. Default is False.
+            Overwrite existing FITS file. Default is False.
         write_covariance : bool
             save covariance or not. Default is False.
         checksum : bool
--- a/gammapy/datasets/core.py
+++ b/gammapy/datasets/core.py
@@ -435,8 +435,8 @@ class Datasets(collections.abc.MutableSequence):
             File path or name of datasets yaml file
         filename_models : str or `Path`
             File path or name of models yaml file. Default is None.
-        overwrite : bool
-            overwrite datasets FITS files. Default is False.
+        overwrite : bool, optional
+            Overwrite existing file. Default is False.
         write_covariance : bool
             save covariance or not. Default is False.
         checksum : bool
--- a/gammapy/datasets/core.py
+++ b/gammapy/datasets/core.py
@@ -436,7 +436,7 @@ class Datasets(collections.abc.MutableSequence):
         filename_models : str or `Path`
             File path or name of models yaml file. Default is None.
         overwrite : bool, optional
-            Overwrite existing FITS file. Default is False.
+            Overwrite existing file. Default is False.
         write_covariance : bool
             save covariance or not. Default is False.
         checksum : bool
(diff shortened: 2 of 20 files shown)


adonath (dismissed):
Great, thanks @Astro-Kirsty!

gammapy/analysis/core.py
```diff
@@ -270,8 +270,8 @@ def write_datasets(self, overwrite=True, write_covariance=True):
 
         Parameters
         ----------
-        overwrite : bool
-            overwrite datasets FITS files
+        overwrite : bool, optional
+            Overwrite existing file. Default is False.
```
QRemy, line 274: The default is True here.

gammapy/analysis/core.py
```diff
@@ -270,8 +270,8 @@ def write_datasets(self, overwrite=True, write_covariance=True):
 
         Parameters
         ----------
-        overwrite : bool
-            overwrite datasets FITS files
```
QRemy, line 274: Here it was specified FITS file because the YAML files are always overwritten.

gammapy/analysis/core.py
adonath: Probably we should make it consistent as well, however for the HLI it might have been a deliberate decision.

gammapy/analysis/core.py
adonath: I guess "overwrite" should apply to both, the YAML and FITS file?

QRemy (changes requested):
Thanks @Astro-Kirsty, I have just few comments

gammapy/analysis/core.py
Astro-Kirsty: Happy to change it to 'False' unless there is a reason we can think why the default should be True

gammapy/analysis/core.py
QRemy: yes pushed a commit for that

QRemy:
There is also no docstring for the write method of `TemplateSpatialModel`, `TemplateNDSpatialModel`, `TemplateNDSpectralModel`, `TemplateNPredModel`.
For these `overwrite` behave differently as `False` won't overwrite and won't raise an error but just emit a warning
Maybe this should be mentioned in their docstring and for the `overwrite_templates` option on `Models.write`.

gammapy/datasets/core.py
```diff
@@ -455,6 +455,9 @@ def write(
             )
             data["datasets"].append(d)
 
+        if path.exists() and not overwrite:
```
registerrier, line 458: Maybe add a test for this?

gammapy/modeling/models/spatial.py
```diff
@@ -1554,6 +1563,15 @@ def evaluate(self, lon, lat, energy=None, **kwargs):
...
 
     def write(self, overwrite=False):
+        """
+        Write the map.
+
+        Parameters
+        ----------
+        overwrite: bool, optional
+            Overwrite existing file.
+            Default is False, which will raise a warning if the template file exists already.
```
registerrier, line 1573: Maybe this can be explained in the main part of the docstring or as a note since this is a different behavior compared to most cases: i.e. it does not fail if there is an existing file and `overwrite=False`.

registerrier (approved):
Thanks @Astro-Kirsty I have left 2 minor inline comments.

gammapy/datasets/core.py
Astro-Kirsty: @QRemy added these lines, but I am actually not sure why. We already pick up on the overwrite?

When I try to test this it has an error at line 453. And the final error is actually from astropy. Maybe I am doing something wrong

QRemy: It added this so the overwrite option apply also to the datasets.yaml file otherwise it is always overwritten

Astro-Kirsty:
I have added a test, but it will need review. 

The only way to ensure this specific line is tested it to first create the dataset and remove the fits file saved, such that it raises the warning regarding the yaml only. Otherwise the warning will be for the fits file overwrite

gammapy/datasets/tests/test_datasets.py
```diff
@@ -87,3 +92,45 @@ def test_datasets_info_table():
     table = datasets_hess.info_table(cumulative=True)
     assert table["name"][0] == "stacked"
     assert table["name"][1] == "stacked"
+
+
+def test_datasets_write(tmp_path):
```
registerrier, line 97: This requires a `@requires_data()`  decorator

registerrier (approved):
Thanks @Astro-Kirsty . The test misses a `requires_data` decorator (from gammay.utils.testing) not to be run when the data are not available.

Astro-Kirsty:
Should be ready to merge now!

gammapy/datasets/tests/test_datasets.py
```diff
@@ -87,3 +92,46 @@ def test_datasets_info_table():
...
+    dataset_1.models = [model]
+
+    datasets.write(
+        filename=tmp_path / "test",
+        filename_models=tmp_path / "test_model",
+        overwrite=False,
+    )
+    os.remove(tmp_path / "test-1.fits")
+
+    with pytest.raises(OSError, match=f"File exists already: {tmp_path / 'test'}"):
```
registerrier, line 132:
```suggestion
    with pytest.raises(OSError):
```

PR #4875, opened 2023-10-30
Remove is_norm from spectral model and add norm to FluxPointEstimator

QRemy:
Remove is_norm from spectral model and add a norm parameter to FluxPointEstimator.
In my opinion this is simpler and easier to configure.
For now it is just a prototype to see how much it would change.


QRemy:
Two bug fix commits from this PR have been moved into separate PR  : #4875 and #4876 
I will drop these commits once the other PR are merged.

registerrier:
Closing now. Replaced by #4902 #4903 #4904

Issue #4811, opened 2023-10-17
Unify scheme for punctuation of parameters and returned objects in docstrings


morcuended:
Raised in #4805.

We should unify the scheme for punctuation of parameters, returned objects in the doctrings.

Also sometimes the description of a parameter contain several lines which are rendered in the docs as a single line. See for example the `selection` or `use_region_center` parameters in docstring of [SpectrumDatasetMaker](https://docs.gammapy.org/dev/api/gammapy.makers.SpectrumDatasetMaker.html?highlight=spectrumdatasetmaker#gammapy.makers.SpectrumDatasetMaker). I do not know if it's possible to show them as item list in the documentation but at least they should be ended with periods to make them more clear.

More information: https://peps.python.org/pep-0257/

MRegeard:
Thanks @morcuended, 
`PEP8` and `PEP257` clearly states that docstring description should end with a full stop. I have made this task list bellow. If one want to take a module and check that docstring description end with a full stop (also don't forget to check for the Upper case at the beginning), they should add their name after the module name in the list. This way it is easy to keep track of who is going what !

- [ ] gammapy.analysis
- [ ] gammapy.astro
- [ ] gammapy.catalog
- [ ] gammapy.data (@MRegeard)
- [ ] gammapy.datasets
- [ ] gammapy.estimators
- [ ] gammapy.extern
- [ ] gammapy.irf
- [ ] gammapy.makers
- [ ] gammapy.maps
- [ ] gammapy.modeling
- [ ] gammapy.script
- [ ] gammapy.stats
- [ ] gammapy.utils
- [ ] gammapy.visualization

registerrier:
should we use a static docstring checker like [pydocstyle](https://pydocstyle.org/en/stable/)?

morcuended:
Thanks @MRegeard for creating the tasklist

> should we use a static docstring checker like [pydocstyle](https://pydocstyle.org/en/stable/)?

Yes, that is a good idea. Maybe we could even think of moving to Ruff (https://astral.sh/ruff) for the linting and all kind of checks

registerrier:
> Thanks @MRegeard for creating the tasklist
> 
> > should we use a static docstring checker like [pydocstyle](https://pydocstyle.org/en/stable/)?
> 
> Yes, that is a good idea. Maybe we could even think of moving to Ruff (https://astral.sh/ruff) for the linting and all kind of checks

[...]

PR #4878, opened 2023-10-31
Cleanup docstrings for `gammapy.catalog`

Astro-Kirsty:
This addresses part of issue https://github.com/gammapy/gammapy/issues/4811, specifically for `gammapy.catalog`

### DIFF
--- a/gammapy/catalog/core.py
+++ b/gammapy/catalog/core.py
@@ -65,18 +65,18 @@ class SourceCatalogObject:
 
     @property
     def name(self):
-        """Source name (str)"""
+        """Source name as a string."""
         name = self.data[self._source_name_key]
         return name.strip()
 
     @property
     def row_index(self):
-        """Row index of source in catalog (int)"""
+        """Row index of source in catalog as an integer."""
         return self.data[self._row_index_key]
 
     @property
     def position(self):
-        """Source position (`~astropy.coordinates.SkyCoord`)."""
+        """Source position as an `~astropy.coordinates.SkyCoord` object."""
         table = Table([self.data])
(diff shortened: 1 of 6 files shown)


QRemy (dismissed):
Thanks a lot @Astro-Kirsty, no comment from my side.

registerrier (dismissed):

Have you tried running pydocstyle to check?

Astro-Kirsty:
I ran pydocstyle, and there are few warnings. Specifically when we do not specify a docstring, we actually tend to have a number of functions without docstring (lots of `__init__` functions). Another common warning is when we have a one line docstring that references a 'signature'.

pydocstyle warning: `D402: First line should not be the function's "signature"`
Examples related to error: 
`"""Source positions (`~astropy.coordinates.SkyCoord`)."""`
`"""Catalog description (str)."""`

The explanation of fixing this can be found here
https://peps.python.org/pep-0257/#one-line-docstrings

gammapy/catalog/core.py
```diff
@@ -274,7 +274,7 @@ def _lookup_extended_source_idx(self):
 
     @property
     def positions(self):
-        """Source positions (`~astropy.coordinates.SkyCoord`)."""
+        """Source positions as a list of `~astropy.coordinates.SkyCoord` objects."""
```
QRemy, line 277:
```suggestion
        """Source positions a `~astropy.coordinates.SkyCoord` object."""
```
it's not a list of SkyCoord objects but only one that includes a list of coordinates

registerrier (approved):
Thanks @Astro-Kirsty !

PR #4879, opened 2023-10-31
Fix energy dependent temporal model simulation

QRemy:
Fix energy dependent temporal model simulation in order to take into account the spectral model value. In principle this fix would allow the spectral model to be any `NormSpectralModel` (which is not allowed for now).

It seems also odd to me that the validity of the model is checked within the sampler
`TypeError: Event sampler expects ConstantSpectralModel for a time varying source. Got PowerLawNormSpectralModel`
This should be done within the `SkyModel` as for spatial templates.

### DIFF
--- a/gammapy/datasets/simulate.py
+++ b/gammapy/datasets/simulate.py
@@ -125,13 +125,12 @@ class MapDatasetEventSampler:
             time_axis_eval.time_mid, energy=energy_new.center
         )
 
-        norm_parameters = model.spectral_model.parameters.norm_parameters
-        norm = norm_parameters[0].quantity
+        norm = model.spectral_model(energy=energy_new.center)
 
         if temp_eval.unit.is_equivalent(norm.unit):
-            flux_diff = temp_eval.to(norm.unit)
+            flux_diff = temp_eval.to(norm.unit) * norm.value[:, None]
         else:
-            flux_diff = temp_eval * norm
+            flux_diff = temp_eval * norm[:, None]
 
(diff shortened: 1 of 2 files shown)


QRemy:
> Here I am not totally sure of the expected behavior. I think the approach was that the `SpectralModel` was used only to apply to correct unit here.

For me if the template temporal model has a unit then the spectral model should a NormSpectral model,
while if the template temporal model gives just a normalisation then the spectral model should be a SpectralModel.
In both case the information provided from the spectral model should be used by evaluating it.

QRemy:
@AtreyeeS, the integration is done by simply multiplying with the energy : https://github.com/gammapy/gammapy/blob/dedaab09f744bc760bfd0ecd86b25d1ccc118a7b/gammapy/datasets/simulate.py#L135
so if the spectrum is evaluated on same the oversampled energies then the multiplication with the template can happen before or after it is the same.

AtreyeeS:
Thanks @QRemy,  I understand now. So adding a `NormSpectralModel` would be akin to adding a spectral correction on the template. That can be useful.

This would then be a feature, and not be back ported I think.

I am still not sure what multiplying a template with a normal spectral model would mean

fabiopintore:
Thanks for the suggestion @QRemy .
I don't have a strong opinion in merit, as it is not very clear to me which means to multiply a template model for a `NormSpectralModel`...

registerrier (approved):
Thansk @QRemy . Looks good to me.

registerrier:
Energy dependent temporal models are not supported in v1.0.x. Backporting is unnecessary.

PR #4881, opened 2023-10-31
Remove fit from events sampling tutorial

QRemy:
Remove the part showing a fit from the events sampling tutorial so the aim of the tutorial is more focused.

### DIFF
--- a/examples/tutorials/analysis-3d/event_sampling.py
+++ b/examples/tutorials/analysis-3d/event_sampling.py
@@ -34,7 +34,7 @@ of predicted counts (`npred`) per bin of the given Sky model, and the
 output of the event-sampler will be a set of events having information
 about their true coordinates, true energies and times of arrival.
 
-To these events, IRF corrections (i.e. PSF and energy dispersion) can
+To these events, IRF corrections (i.e. PSF and energy dispersion) can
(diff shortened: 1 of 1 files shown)


bkhelifi (approved):
This is a valid point. Maybe you can just refer at the end the tutorial to explore the simulated data, but this is not mandatory.

QRemy:
> This is a valid point. Maybe you can just refer at the end the tutorial to explore the simulated data, but this is not mandatory.

The last sentence of the tutorial already mention this

` 
Then you can create the observations from the data store and make your own analysis following the instructions in the [Low level API](https://docs.gammapy.org/dev/tutorials/starting/analysis_2.html) tutorial.`

adonath (approved):
Thanks @QRemy, I agree the tutorial is already long enough.

registerrier:
I propose not to backport this change as it is only touching documentation.

fabiopintore:
thanks @QRemy . The tutorial is indeed quite long, so if you all agree that the fitting steps are not necessary, then it's fine to me as well. 
However, I suppose that, if not strictly needed, removing information may be not an optimal option. Even if redundant because shown in other tutorials, cross-topic information are appreciated by the beginner users, which may have a broader view in one-shot rather than searching for explanations in other tutorials.

PR #4867, opened 2023-10-26
Addition of the class `FluxMetaData`

bkhelifi:
**Description**

This pull request is associated to the Issue #4790. It contains the basic structure of this class, requiring still some discussion with the reviewers. 
Separating the internal data format of Gammapy with the serialization, the functions `from_dict`, `to_table` are however associated to the serialization. And with these class attributes, we are going beyond the GADF format.

**Dear reviewer**
The implementation has been achieved by making the full integration of this class in the `estimators` sub-package. For the reviewing purpose, I have splitted my work into two:
- this PR with the basic class
- a second one about the usage of the class in `estimators` sub-package (with some impact on the map sub-package because of missing propagation of metadata in some functions). The draft of this work can be found the branch 'Issue4790' in my fork (https://github.com/bkhelifi/gammapy/tree/Issue4790)

PR #4886, opened 2023-11-02
Introduce hierarchical metadata structures

registerrier:
**Description**

This pull request introduces hierarchical metadata structures as discussed during dev call.

A registry of FITS header exporters in introduced to follow on the ideas developed in #4867 and during the dev call.

**Dear reviewer**

### DIFF
--- a/gammapy/data/observations.py
+++ b/gammapy/data/observations.py
@@ -17,7 +17,7 @@ from astropy.utils import lazyproperty
 import matplotlib.pyplot as plt
 from gammapy.utils.deprecation import GammapyDeprecationWarning, deprecated
 from gammapy.utils.fits import LazyFitsData, earth_location_to_dict
-from gammapy.utils.metadata import CreatorMetaData
+from gammapy.utils.metadata import CreatorMetaData, TargetMetaData
 from gammapy.utils.scripts import make_path
 from gammapy.utils.testing import Checker
 from gammapy.utils.time import time_ref_to_dict, time_relative_to_ref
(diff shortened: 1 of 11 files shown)


adonath:
@registerrier I'd propose that we merge this PR fast and I rebase and continue with #4750...

registerrier:
The `obs_info` argument on init is now deprecated and for now recreated from the `meta` with `meta.to_header()`. It does not contain exactly the same members as the full events list meta. Some tests were adapted.

The remaining issue is `obs_id` which is present both on the `meta` and as an attribute. Removing it from the main class, requires creating a `meta` but not filling it since we have to wait for the fits file to be opened. This might be solved later.

gammapy/data/observations.py
```diff
@@ -331,19 +337,15 @@ def observation_dead_time_fraction(self):
         """
         return self.meta.deadtime_fraction
 
-    @lazyproperty
+    #    @deprecated("v1.2", alternative="meta")
```
registerrier, line 340:
```suggestion
```

gammapy/data/observations.py
```diff
@@ -550,7 +552,7 @@ def read(cls, event_file, irf_file=None):
         return cls(
             events=events,
             gti=gti,
-            obs_info=obs_info,
+            #            obs_info=obs_info,
```
registerrier, line 555:
```suggestion
```

tox.ini
```diff
@@ -77,7 +77,7 @@ deps =
     oldestdeps: astropy==5.0.*
     oldestdeps: click==7.0.*
     oldestdeps: regions==0.5.*
-    oldestdeps: pydantic==1.4.*
+    oldestdeps: pydantic==1.5.*
```
registerrier, line 80:
```suggestion
    oldestdeps: pydantic==1.4.*
```

registerrier:
Merging now.

PR #4895, opened 2023-11-07
Add `DatasetsMaker` to HowTo

Astro-Kirsty:
This PR aims to address one task in https://github.com/gammapy/gammapy/issues/4788

This is one idea to expose the `DatasetsMaker`. 
Most of our tutorials utilise the analysis config, so could not be adapted to utilise the `DatasetsMaker`.
We already demonstrate the  `DatasetsMaker` in the api makers tutorial. So I believe mentioning this in the HowTo is enough


registerrier (approved):

PR #4898, opened 2023-11-10
Add support for astropy 6.0.0rc1

maxnoe:
**Description**

This pull request ...

**Dear reviewer**

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,6 +48,10 @@ jobs:
             tox_env: 'codestyle'
             allowed_fail: false
           - os: ubuntu-latest
+            python: '3.10'
+            tox_env: 'astropy60'
+            allowed_fail: false
+          - os: ubuntu-latest
             python: '3.9'
             tox_env: 'py39-test-alldeps_noray-astropylts-numpy121'
             allowed_fail: false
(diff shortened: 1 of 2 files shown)


registerrier:
Thanks @maxnoe ! Indeed we have an issue with the coming astropy v6

maxnoe:
This seems mostly to be related with the removal for support of adding unit-less quantities to times.

In astropy 5.x, this was always interpreted as days, astropy v6 raises an error and requires units.

AtreyeeS:
The issue is in adding an empty list to a Time object

```
start = [] * u.s
reference_time + u.Quantity(start)
```

This give the error `Unsupported operand type(s) for ufunc add: 'Time,Quantity'`

This breaks most of our tests because creating an empty dataset calls `GTI.create` with an empty `start` and `stop` list.

If the `start` list is non empty, there is no problem.

To solve this, we can change the default to `[0]*u.s`. @registerrier @bkhelifi @maxnoe @adonath

maxnoe:
@AtreyeeS the old behavior was to create an empty time object:

```
>>> Time.now() + [] * u.s
<Time object: scale='utc' format='datetime' value=[]>
```

So you could do:

```
start = self.reference_time + start if len(start) > 0 else Time([], format='mjd')
```

registerrier:
This is now corrected in astropy. devdeps is now running. 
See https://github.com/astropy/astropy/pull/15604

Thanks @maxnoe !

registerrier:
CI is now running with devdeps. Closing now. Thanks @maxnoe

PR #4902, opened 2023-11-15
Add norm attribute to estimators and deprecate previous norm related attributes

QRemy:
Add norm attribute to estimators and deprecate previous norm related attributes.
Split #4875 part 1.

### DIFF
--- a/examples/tutorials/analysis-3d/flux_profiles.py
+++ b/examples/tutorials/analysis-3d/flux_profiles.py
@@ -272,9 +272,8 @@ flux_profile_estimator = FluxProfileEstimator(
     spectrum=PowerLawSpectralModel(index=2.3),
     energy_edges=[10, 100, 2000] * u.GeV,
     selection_optional=["ul", "scan"],
-    norm_values=np.linspace(-1, 5, 11),
 )
-
+flux_profile_estimator.norm.scan_values = np.linspace(-1, 5, 11)
 profile = flux_profile_estimator.run(datasets=dataset)
 
 
--- a/examples/tutorials/analysis-3d/flux_profiles.py
+++ b/examples/tutorials/analysis-3d/flux_profiles.py
@@ -15,7 +15,7 @@ Context
 
 A useful tool to study and compare the spatial distribution of flux in
 images and data cubes is the measurement of flux profiles. Flux profiles
-can show spatial correlations of gamma-ray data with e.g. gas maps or
+can show spatial correlations of gamma-ray data with e.g. gas maps or
 other type of gamma-ray data. Most commonly flux profiles are measured
 along some preferred coordinate axis, either radially distance from a
 source of interest, along longitude and latitude coordinate axes or
@@ -178,7 +178,7 @@ plt.show()
 
 ######################################################################
 # Based on the spectral model we specified above we can also plot in any
-# other sed type, e.g. energy flux and define a different threshold when
+# other sed type, e.g. energy flux and define a different threshold when
 # to plot upper limits:
 #
 
@@ -265,7 +265,6 @@ plt.show()
 ######################################################################
 # This time we define two energy bins and include the fit statistic
 # profile in the computation:
-#
 
 flux_profile_estimator = FluxProfileEstimator(
     regions=regions,
@@ -273,12 +272,18 @@ flux_profile_estimator = FluxProfileEstimator(
     energy_edges=[10, 100, 2000] * u.GeV,
     selection_optional=["ul", "scan"],
 )
+######################################################################
+# The configuration of the fit statistic profile is done throught the norm parameter:
 flux_profile_estimator.norm.scan_values = np.linspace(-1, 5, 11)
+
+######################################################################
+# Now we can run the estimator,
+
 profile = flux_profile_estimator.run(datasets=dataset)
 
 
 ######################################################################
-# We can directly plot the result:
+# and plot the result:
 #
 
 profile.plot(axis_name="projected-distance", sed_type="flux")
--- a/gammapy/estimators/flux.py
+++ b/gammapy/estimators/flux.py
@@ -127,8 +127,8 @@ class FluxEstimator(ParameterEstimator):
         norm.max = scaled_parameter.max / scaled_parameter.value
         norm.interp = scaled_parameter.interp
         norm.scan_values = self.norm.scan_values
-        norm.scan_min = self.norm.min
-        norm.scan_max = self.norm.max
+        norm.scan_min = self.norm.scan_min
+        norm.scan_max = self.norm.scan_max
         norm.scan_n_values = self.norm.scan_n_values
         return norm
 
--- a/gammapy/estimators/flux.py
+++ b/gammapy/estimators/flux.py
@@ -156,7 +156,9 @@ class FluxEstimator(ParameterEstimator):
         norms = ref_model.parameters.norm_parameters
 
         if len(norms.free_parameters) == 1:
-            self.norm = self._set_norm_parameter(self.norm, norms.free_parameters[0])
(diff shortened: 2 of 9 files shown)


gammapy/estimators/flux.py
```diff
@@ -68,12 +85,23 @@ def __init__(
         selection_optional=None,
         fit=None,
         reoptimize=False,
+        norm=None,
```
adonath, line 88: I was just wondering whether we should make the name a little bit more clearer, to avoid the confusion of an estimator having a `Parameter` object. For example `norm_spec` or similar.

gammapy/estimators/flux.py
```diff
@@ -68,12 +85,23 @@ def __init__(
...
     ):
-        self.norm_values = norm_values
-        self.norm_min = norm_min
-        self.norm_max = norm_max
-        self.norm_n_values = norm_n_values
+
         self.source = source
+        if norm is None:
+            norm = Parameter(
```
adonath, line 93: To further clarify that the parameter is not a parameter in the model sense, but more a configuration class we could also allow users to pass a dict as well and create the parameter internally.  Support both.

adonath (commented):
Thanks @QRemy, I have left two minor comments. I think it could be beneficial to make it a bit clearer, that the norm parameter acts more as configuration here then a model parameter.

gammapy/estimators/flux.py
QRemy: done

gammapy/estimators/flux.py
QRemy: The existing docstring already explains that a "norm" parameter is used to rescale the reference model so it should not be that surprising for the user :
`The amplitude is re-normalized using the "norm" parameter,
    which specifies the deviation of the flux from the reference model in this
    energy range.`
 and I would prefer to keep it short so it is easier to set the attributes.

gammapy/estimators/points/sed.py
```diff
@@ -81,6 +82,11 @@ class FluxPointsEstimator(FluxEstimator, parallel.ParallelMixin):
 
     tag = "FluxPointsEstimator"
 
+    norm_min = deprecated_attribute("norm_min", "1.2")
```
AtreyeeS, line 85: The docstring of Parameters for the FluxPointsEstimator needs to be updated

AtreyeeS (dismissed):
Thanks @QRemy ! Left some small inline comments

gammapy/estimators/flux.py
QRemy: added more details in the docsting

gammapy/estimators/flux.py
```diff
@@ -52,10 +53,28 @@ class FluxEstimator(ParameterEstimator):
     reoptimize : bool
         Re-optimize other free model parameters. Default is False.
         If True the available free parameters are fitted together with the norm of the source of interest in each bin independently, otherwise they are frozen at their current values.
+    norm : ~gammapy.modeling.Parameter` or dict
+        Norm parameter used for the fit
+        Default is None and a new parameter is created automatically,
+        with value=1, name="norm", scan_min=0.2, scan_max=5, and scan_n_values = 11.
```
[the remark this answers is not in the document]
QRemy: no, added an error if norm.name!="norm"

registerrier (commented):
Thanks @QRemy ! See inline comments.

gammapy/modeling/parameter.py
```diff
@@ -98,6 +99,8 @@ class Parameter:
         Prior set on the parameter.
     """
 
+    norm_parameters = deprecated_attribute("norm_parameters", "1.2")
```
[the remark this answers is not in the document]
QRemy: Parameter.is_norm will be deprecated by #4904.
For `norm_parameters` I did it here because it is used only by the estimators

gammapy/estimators/flux.py
```diff
@@ -68,12 +87,29 @@ def __init__(
...
+                scan_min=norm_min,
+                scan_max=norm_max,
+                scan_n_values=norm_n_values,
+                scan_values=norm_values,
+            )
+            if isinstance(norm, dict):
+                norm_kwargs.update(norm)
+            norm = Parameter(**norm_kwargs)
+        if norm.name != "norm":
+            raise ValueError("norm should be a norm parameter")
```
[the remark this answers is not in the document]
QRemy: I have to check but I think it is necessary to set it on the scale model

examples/tutorials/analysis-3d/flux_profiles.py
```diff
@@ -265,21 +265,25 @@
...
 
 flux_profile_estimator = FluxProfileEstimator(
     regions=regions,
     spectrum=PowerLawSpectralModel(index=2.3),
     energy_edges=[10, 100, 2000] * u.GeV,
     selection_optional=["ul", "scan"],
-    norm_values=np.linspace(-1, 5, 11),
 )
+######################################################################
+# The configuration of the fit statistic profile is done throught the norm parameter:
```
Astro-Kirsty, line 276:
```suggestion
# The configuration of the fit statistic profile is done through the norm parameter:
```

gammapy/estimators/flux.py
```diff
@@ -52,10 +53,28 @@ class FluxEstimator(ParameterEstimator):
     reoptimize : bool
         Re-optimize other free model parameters. Default is False.
         If True the available free parameters are fitted together with the norm of the source of interest in each bin independently, otherwise they are frozen at their current values.
+    norm : ~gammapy.modeling.Parameter` or dict
+        Norm parameter used for the fit
```
Astro-Kirsty, line 57:
```suggestion
    norm : `~gammapy.modeling.Parameter` or dict
        Norm parameter used for the fit.
```

gammapy/estimators/flux.py
```diff
@@ -52,10 +53,28 @@ class FluxEstimator(ParameterEstimator):
...
+        Default is None and a new parameter is created automatically,
+        with value=1, name="norm", scan_min=0.2, scan_max=5, and scan_n_values = 11.
+        By default the min and max are not set and derived from the source model,
```
Astro-Kirsty, line 60:
```suggestion
        By default the minimum and maximum are not set and derived from the source model,
```

gammapy/estimators/points/lightcurve.py
```diff
@@ -78,6 +79,14 @@ class LightCurveEstimator(FluxPointsEstimator):
         of jobs is limited to the number of physical CPUs.
     parallel_backend : {"multiprocessing", "ray"}
         Which backend to use for multiprocessing. Defaults to `~gammapy.utils.parallel.BACKEND_DEFAULT`.
+    norm : ~gammapy.modeling.Parameter` or dict
+        Norm parameter used for the fit
```
Astro-Kirsty, line 83:
```suggestion
    norm : `~gammapy.modeling.Parameter` or dict
        Norm parameter used for the fit.
```

gammapy/estimators/points/sed.py
```diff
@@ -77,10 +78,23 @@ class FluxPointsEstimator(FluxEstimator, parallel.ParallelMixin):
         limited to the number of physical CPUs.
     parallel_backend : {"multiprocessing", "ray"}
         Which backend to use for multiprocessing.
+    norm : ~gammapy.modeling.Parameter` or dict
+        Norm parameter used for the fit
```
Astro-Kirsty, line 82:
```suggestion
    norm : `~gammapy.modeling.Parameter` or dict
        Norm parameter used for the fit.
```

Astro-Kirsty (commented):
Thanks @QRemy. 
Looks good, just a few minor inline comments

PR #4903, opened 2023-11-15
Remove useless norm parameters from spectral model

QRemy:
Remove useless norm parameters from spectral model.
Split #4875 part 2.
Requires #4902.


MRegeard (approved):
Thanks @QRemy. 
No comments from my side !

PR #4905, opened 2023-11-15
Correct typo in init of fov background maker

MRegeard:
This PR correct a typo on method in the init of the fov background maker. The method was set as `self.method` in the init, meaning that the property and its setter where not working at all.

### DIFF
--- a/gammapy/makers/background/fov.py
+++ b/gammapy/makers/background/fov.py
@@ -56,7 +56,7 @@ class FoVBackgroundMaker(Maker):
         min_npred_background=0,
         fit=None,
     ):
-        self.method = method
+        self._method = method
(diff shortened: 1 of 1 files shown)


adonath (commented):
Thanks @MRegeard, can you please add a regression test? Meaning a test that would have failed before the fix..

registerrier (commented):
Thanks @MRegeard . I am not sure I understand why this would be causing problems. The setter mechanisms was working, hence the current CI test fail.

MRegeard:
I was confused because I the `_method` is not defined anywhere and being returned. So I thought it was a mistake, and did not even tested this... But it is working as expected !
So I'm just closing this issue.

PR #4906, opened 2023-11-15
Cleanup docstrings for `gammapy.maps.region`

Astro-Kirsty:
This addresses part of issue https://github.com/gammapy/gammapy/issues/4811, specifically for `gammapy.maps.region`

### DIFF
--- a/gammapy/maps/region/geom.py
+++ b/gammapy/maps/region/geom.py
@@ -41,6 +41,7 @@ __all__ = ["RegionGeom"]
 
 class RegionGeom(Geom):
     """Map geometry representing a region on the sky.
+
     The spatial component of the geometry is made up of a
     single pixel with an arbitrary shape and size. It can
     also have any number of non-spatial dimensions. This class
@@ -53,11 +54,11 @@ class RegionGeom(Geom):
     axes : list of `MapAxis`
         Non-spatial data axes.
     wcs : `~astropy.wcs.WCS`
-        Optional wcs object to project the region if needed.
+        Optional WCS object to project the region if needed.
     binsz_wcs : `float`
         Angular bin size of the underlying `~WcsGeom` used to evaluate
-        quantities in the region. Default size is 0.01 deg. This default
-        value is adequate for the majority of use cases. If a wcs object
+        quantities in the region. Default is "0.1 deg". This default
+        value is adequate for the majority of use cases. If a WCS object
         is provided, the input of binsz_wcs is overridden.
     """
 
@@ -111,8 +112,7 @@ class RegionGeom(Geom):
 
     @property
     def frame(self):
-        """Coordinate system, either Galactic ("galactic") or Equatorial
-        ("icrs")."""
+        """Coordinate system, either Galactic ("galactic") or Equatorial ("icrs")."""
         if self.region is None:
             return "icrs"
         try:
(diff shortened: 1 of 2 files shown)


gammapy/maps/region/ndmap.py
```diff
@@ -243,48 +247,52 @@ def create(
...
+            Dictionary to store metadata.
+            Default is None.
+        wcs : `~astropy.wcs.WCS`, optional
+            WCS projection to use for local projections of the region.
+            Default is None.
+        binsz_wcs: `~astropy.units.Quantity` or str, optional
             Bin size used for the default WCS, if wcs=None.
-        data : `~numpy.ndarray`
-            Data array
+            Default is "0.1deg".
```
registerrier, line 265:
```suggestion
            Default is "0.1 deg".
```

registerrier (approved):
I have left a few inline comments.

gammapy/maps/region/ndmap.py
Astro-Kirsty: The default value is actually set to "0.1deg", is that a mistake then?

gammapy/maps/region/geom.py
```diff
@@ -417,35 +418,37 @@ def to_binsz_wcs(self, binsz):
 
         Parameters
         ----------
-        binzs : float, string or `~astropy.quantity.Quantity`
+        binsz : float, string or `~astropy.quantity.Quantity`
```
MRegeard, line 421:
```suggestion
        binsz : float or string or `~astropy.quantity.Quantity`
```

gammapy/maps/region/ndmap.py
```diff
@@ -33,13 +34,15 @@ class RegionNDMap(Map):
     geom : `~gammapy.maps.RegionGeom`
         Region geometry object.
     data : `~numpy.ndarray`
-        Data array. If none then an empty array will be allocated.
+        Data array. If None then an empty array will be allocated.
```
MRegeard, line 37:
```suggestion
        Data array. If None, an empty array will be allocated.
```

gammapy/maps/region/ndmap.py
```diff
@@ -407,10 +417,10 @@ def interp_by_coord(self, coords, **kwargs):
             "lon" and "lat" are optional and will be taken at the center
             of the region by default.
         method : {"linear", "nearest"}
-            Method to interpolate data values. By default linear
+            Method to interpolate data values. By default, linear
```
MRegeard, line 420: I think that I prefer the default value to be clearly stated and not periphrased such that the user clearly sees what is put as the default value. It is also an other way to instinctively guide the user to the type of the parameter.
```suggestion
            Method to interpolate data values. Default is 'linear'.
```

MRegeard (commented):
Thanks @Astro-Kirsty, see inline comments.

gammapy/maps/region/ndmap.py
Astro-Kirsty: I decided to keep the 'then' format because there were lots of other docstrings written in this way

gammapy/maps/region/geom.py
Astro-Kirsty: When you have a list, the 'or' is only written between the last two quantities.

MRegeard (approved):
Thanks @Astro-Kirsty !

registerrier (approved):

PR #4907, opened 2023-11-16
Prior statistics and set-up for multidimensional priors

katrinstreil:
**Description**

This pull request modifies how priors are evaluated with a separate stat method: `prior_fit_statistic(priors)`.
The priors themselves are no longer saved on the `parameters` object but on `models`.  
To allow the evaluation of multidimensional priors, the information on the model parameters is saved on the prior instance as `modelparameters`.

### DIFF
--- a/gammapy/datasets/core.py
+++ b/gammapy/datasets/core.py
@@ -79,19 +79,19 @@ class Dataset(abc.ABC):
             prior_stat_sum = self.models.parameters.prior_stat_sum()
         return np.sum(stat, dtype=np.float64) + prior_stat_sum
 
-    def set_priors_from_models(self, models):
-        if models is not None:
-            self._priors = [
-                par.prior for par in models.parameters if par.prior is not None
-            ]
-        else:
-            self._priors = None
+    # def set_priors_from_models(self, models):
+    #    if models is not None:
+    #        self._priors = [
+    #            par.prior for par in models.parameters if par.prior is not None
+    #        ]
+    #    else:
+    #        self._priors = None
 
-    @property
-    def priors(self):
-        """Priors set on model parameters (list)."""
-        self.set_priors_from_models(self.models)
-        return self._priors
+    # @property
+    # def priors(self):
+    #    """Priors set on model parameters (list)."""
+    #    self.set_priors_from_models(self.models)
+    #    return self._priors
 
     @abc.abstractmethod
     def stat_array(self):
@@ -186,19 +186,20 @@ class Datasets(collections.abc.MutableSequence):
 
         return DatasetModels(list(models.keys()))
 
-    @property
-    def priors(self):
+        # @property
+        # def priors(self):
         """Priors (list).
 
         Duplicate prior objects have been removed.
         """
-        priors = {}
 
-        for dataset in self:
-            if dataset.models is not None:
-                for prior in dataset.priors:
-                    priors[prior] = prior
-        return priors
+    #   priors = {}
+
+    #   for dataset in self:
+    #       if dataset.models is not None:
+    #           for prior in dataset.priors:
+    #               priors[prior] = prior
+    #   return priors
 
     @models.setter
     def models(self, models):
@@ -261,8 +262,8 @@ class Datasets(collections.abc.MutableSequence):
         for dataset in self:
             stat_sum += dataset.stat_sum()
             # remove prior_fit_statistic from individual dataset to avoid double counting
-            stat_sum -= prior_fit_statistic(dataset.priors)
-        return stat_sum + prior_fit_statistic(list(self.priors.values()))
+            stat_sum -= prior_fit_statistic(dataset.models.priors)
+        return stat_sum + prior_fit_statistic(list(self.models.priors.values()))
 
     def select_time(self, time_min, time_max, atol="1e-6 s"):
         """Select datasets in a given time interval.
--- a/gammapy/datasets/core.py
+++ b/gammapy/datasets/core.py
@@ -230,11 +230,16 @@ class Datasets(collections.abc.MutableSequence):
         """Compute joint statistic function value."""
         stat_sum = 0
         # TODO: add parallel evaluation of likelihoods
+        prior_stat_sum = 0
         for dataset in self:
             stat_sum += dataset.stat_sum()
             # remove prior_fit_statistic from individual dataset to avoid double counting
-            stat_sum -= prior_fit_statistic(dataset.models.priors)
-        return stat_sum + prior_fit_statistic(self.models.priors)
+            if dataset.models is not None:
+                prior_stat_sum -= prior_fit_statistic(dataset.models.priors)
+        # compute prior_fit_statistics from all datasets
+        if self.models is not None:
+            prior_stat_sum += prior_fit_statistic(self.models.priors)
+        return stat_sum + prior_stat_sum
 
     def select_time(self, time_min, time_max, atol="1e-6 s"):
         """Select datasets in a given time interval.
--- a/gammapy/datasets/core.py
+++ b/gammapy/datasets/core.py
@@ -9,6 +9,7 @@ from astropy import units as u
 from astropy.table import Table, vstack
 from gammapy.data import GTI
 from gammapy.modeling.models import DatasetModels, Models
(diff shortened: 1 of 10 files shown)


gammapy/datasets/core.py
```diff
@@ -229,9 +230,16 @@ def stat_sum(self):
...
+        prior_stat_sum = 0
         for dataset in self:
             stat_sum += dataset.stat_sum()
-        return stat_sum
+            # remove prior_fit_statistic from individual dataset to avoid double counting
+            if dataset.models is not None:
+                prior_stat_sum -= prior_fit_statistic(dataset.models.priors)
+        # compute prior_fit_statistics from all datasets
+        if self.models is not None:
+            prior_stat_sum += prior_fit_statistic(self.models.priors)
```
adonath, line 241: Why does one prior count positive the other negative?

adonath: `Datasets.model` includes all models, why is there an independent sum for `dataset.models`?

gammapy/datasets/map.py
```diff
@@ -1114,12 +1113,11 @@ def plot_residuals(
...
     def stat_sum(self):
-        """Total statistic function value given the current model parameters and priors."""
+        """Total statistic function value given the current model parameters and priors set on the models."""
+        counts, npred = self.counts.data.astype(float), self.npred().data
         prior_stat_sum = 0.0
         if self.models is not None:
-            prior_stat_sum = self.models.parameters.prior_stat_sum()
-
-        counts, npred = self.counts.data.astype(float), self.npred().data
+            prior_stat_sum = prior_fit_statistic(self.models.priors)
```
adonath, line 1120: Wouldn't this double count the prior, when it is evaluated again at the `Datasets.stat_sum()` level?

gammapy/modeling/models/prior.py
```diff
@@ -32,7 +32,15 @@ def _build_priorparameters_from_dict(data, default_parameters):
 class Prior(ModelBase):
     _unit = ""
 
-    def __init__(self, **kwargs):
+    def __init__(self, modelparameters, **kwargs):
```
adonath, line 35: I think taking the `Parameter` object here on init is rather confusing to establish the linking between parameters and their priors. I think this also deviates from the API proposed in the PIG.

gammapy/modeling/models/prior.py
```diff
@@ -144,8 +162,7 @@ class GaussianPrior(Prior):
     mu = PriorParameter(name="mu", value=0)
     sigma = PriorParameter(name="sigma", value=1)
 
-    @staticmethod
-    def evaluate(value, mu, sigma):
+    def evaluate(self, value, mu, sigma):
```
adonath, line 165: `self` I snot used, this can remain a static method.

gammapy/modeling/models/prior.py
```diff
@@ -172,8 +189,7 @@ class UniformPrior(Prior):
     min = PriorParameter(name="min", value=-np.inf, unit="")
     max = PriorParameter(name="max", value=np.inf, unit="")
 
-    @staticmethod
-    def evaluate(value, min, max):
+    def evaluate(self, value, min, max):
```
adonath, line 192: Same as above...

adonath (commented):
Thanks @katrinstreil! I have left some first comments in line.

adonath:
There are some aspects, that are confusing to me in the current implementation.

- I'm not sure we should bind the prior evaluation to `.stat_sum()`. Currently priors might be double counted as they are evaluated in `Dataset.stat_sum()` as well as `Datasets.stat_sum()`
- We could possibly introduce `.stat_sum_prior()` and `.stat_sum_posterior()` on both `Dataset` as well as `Datasets` if we want to distinguish. 

Next question is how the evaluate hierarchical priors. I think that priors exist on multiple levels of the model hierarchy:

- `Parameter.prior`
- `SpectralModel.prior`, `SpatialModel.prior`, etc.
- `SkyModel.prior`
- `Models.prior`

And I think they can be treated independently. So e.g. if a users defines:

```python3
norm_pl = PowerLawNormSpectralModel()
norm_pl.prior = MultivariateGaussianPrior()
```

Then `norm_pl.norm_1.prior1 is still `None`. This is for example different than the covariance handling we have. 

In general I think it makes sense to define a `MultivariatePrior` class. This class could be created "on the fly" on the objects higher in the hierarchy and gather the priors from the models lower in the hierarchy. So the information flows bottom to top only(!). The `MultivariatePrior` class could indeed take the `Parameters` object internally, very similar to the `Covariance` class we already have. However users are never required to pass the parameters themselves, the linking happens internally.

These are a a few half-structured thoughts, maybe they help a bit in developing a clearer vision of the implementation.

katrinstreil:
Thanks @adonath for the comments! 

**About the first thing you mention concerning the evaluation of the prior in `dataset` and `datasets`:**

- In the here proposed implementation, the `prior_stat_sum` is always added to the `stat_sum`. When the `stat_sum` of a `datasets` instance gets computed, the `prior_stat_sum` is counted multiple times. To avoid this, I subtract the individual `prior_stat_sums` and only count the joint `prior_stat_sums`. I agree that this could be more elegant. 
- I like the proposed way of having a `.prior_stat_sum()`, `.stat_sum()` and `.posterior_stat_sum() = .prior_stat_sum() + .stat_sum()`. However, this might be complex for the user. 

**About the hierarchical treatment of the prior:** 
I am not sure if we could cover all possible use cases without the user having to define the `Parameters` on which a multivariate prior sits.  

For instance, coupling two indices from two Skymodels? 
If one passes the `Parameters` to the `MultiVariatePrior` it would look like this:

```
model1 = PowerLawSpectralModel()
model2 = PowerLawSpectralModel()
MulitVariatePrior(covariance_matrix = C, 
                  modelparameters = [model1.parameters['index'], 
                                     model2.parameters['index'])
```
Which intrinsically sets: `.prior` on the two `Parameter` instances. 

How would this is implemented without passing the model parameters to the prior instance? The only possibility I can think of is the following: 
``` 
multivprior = MultiVariatePrior(covariance_matrix = C)
model1.parameters['index'].prior = multivprior
model2.parameters['index'].prior = multivprior
```
And then the mapping happens internally. 


In your example:
 ```
norm_pl = PowerLawNormSpectralModel()
norm_pl.prior = MultivariateGaussianPrior()
```
Would the `MulitvariateGaussianPrior` be applied to all three parameters of `PowerLawNormSpectralModel`? Even the `reference`? How could one control this? 

**About the non-static evaluation of the priors:**
- I know that this deviates from the implementation outline in the PIG. However, a static evaluation is no longer possible if the model parameters are passed to the multidimensional prior, as suggested in the PR. 
- Even if the linking happens internally, priors depending on the values of multiple parameters cannot be evaluated statically. 
- I think it is easier for the user if priors sitting on single parameters and priors sitting on multiple parameters/models are handled consistently. Therefore, I also removed the static evaluation from the `GaussianPrior` and `UniformPrior`.

**Comment by adonath** [comment]
Your first example with the coupled indices is an interesting use case, that will help us! I think we are going in the exact same direction, with some small differences. Your first use case I think should look like:

```python3
models = Models([model1, model2])

# this creates a full covariance matrix internally
prior = MultiVariateGaussianPrior.from_subcovariance(
    pars=[model1.spectral_model.index, model2.spectral_model.index],
    subcovar=C_sub
)

models.prior = prior


# but this should also work:
prior = MultiVariateGaussianPrior(
    parameters=None
    covariance=C_full
)

# just based on the fact that the number parameters in models is the same size as the full covariance
# matrix one can assign it internally. See 
models.prior = prior
```
But yes internally the `MultiVariateGaussianPrior` needs to store a covariance matrix and a list of parameters.

adonath:
> Which intrinsically sets: .prior on the two Parameter instances.

This is the part I'm less sure about. This would require passing information both direction: up and down the hierarchy. I implemented this for the covariance handling of models and it is rather complex, with not such a huge benefit. Or do you propose to only pass the information down?

registerrier:
Any news here @katrinstreil ?

QRemy:
Is this replaced by #5468 ? @Astro-Kirsty 
Should we close this one ?

registerrier:
This is superseded by #5468 and #5838. Closing.

PR #4912, opened 2023-11-17
Impose that the black version in the YAML environment corresponds to …

cgalelli:
This PR introduces a force on the version of black installed by `environment-dev.yml`, so that it is the same as the one used for the CI and pre-commit hooks. This ensures that if somebody locally runs `black <SOME_FOLDER>`, no undesired formatting is introduced.


registerrier (approved):
Thanks @cgalelli !

PR #4913, opened 2023-11-17
Fix NoOverlapError in SpatialModel.integrate_geom

QRemy:
Fix a NoOverlapError bug in  SpatialModel.integrate_geom that seems to appears when the models is outside the geom but contributes inside only because of a very large PSF radius.


registerrier (commented):
Thanks @QRemy. Looks good. 
Is it possible to have a test for this?

QRemy:
> Is it possible to have a test for this?
 It is quite hard to reproduce with the test data we have

AtreyeeS (approved):
Thanks @QRemy ! No comments

registerrier (approved):
Thanks @QRemy

PR #4917, opened 2023-11-20
Add new class to directly simulate observations

maxnoe:
**Description**
This is relatively old, from one of the previous coding sprints and I never opened the pull request but was now reminded in the context of the running CTA simulations.

This adds a new simulator class, which what we discussed at the time is a better API: treating the `MapDataset` as internal implementation detail of the simulator and making the API take `observation, models`, returning a new observation with the filled-in events.

**Dear reviewer**

### DIFF
--- a/gammapy/datasets/simulate.py
+++ b/gammapy/datasets/simulate.py
@@ -563,7 +563,6 @@ class MapDatasetEventSampler:
         events = EventList.from_stack([events_bkg, events_src])
 
         events = self.event_det_coords(observation, events)
-        events.table.sort("TIME")
         events.table["EVENT_ID"] = np.arange(len(events.table))
         events.table.meta.update(
             self.event_list_meta(dataset, observation, self.keep_mc_id)
--- a/gammapy/datasets/simulate.py
+++ b/gammapy/datasets/simulate.py
@@ -583,6 +583,15 @@ class MapDatasetEventSampler:
 class SimulatedObservationMaker(MapDatasetEventSampler):
     """
     Sample event lists for a given observation and signal models.
+
+    Signal events are sampled from the given sky models and then
+    folded with the instrument response functions.
+    To improve performance, IRFs are evaluated on a pre-defined binning,
+    not at each individual event energy / coordinate.
+
+    Parameters
+    ----------
+
     """
 
     def __init__(
@@ -625,7 +634,7 @@ class SimulatedObservationMaker(MapDatasetEventSampler):
         if models is None:
             models = Models()
 
-        if observation.pointing.pointing_mode is not PointingMode.POINTING:
+        if observation.pointing.mode is not PointingMode.POINTING:
             raise NotImplementedError(
                 "Only observations with fixed pointing in ICRS are supported"
             )
@@ -644,7 +653,7 @@ class SimulatedObservationMaker(MapDatasetEventSampler):
         )
         if observation.edisp is not None:
             components.append("edisp")
-            axes["mira_axis"] = observation.edisp.axes["migra"]
+            axes["migra_axis"] = observation.edisp.axes["migra"]
         if observation.bkg is not None:
             components.append("background")
         if observation.psf is not None:
@@ -658,11 +667,10 @@ class SimulatedObservationMaker(MapDatasetEventSampler):
 
         maker = MapDatasetMaker(selection=components)
         dataset = maker.run(dataset, observation)
-        models.append(FoVBackgroundModel(dataset_name="simulated-dataset"))
         dataset.models = models
 
         events = None
-        if len(models) > 1:
+        if len(models) > 0:
             events = self.sample_sources(dataset)
 
             if len(events.table) > 0:
@@ -672,12 +680,17 @@ class SimulatedObservationMaker(MapDatasetEventSampler):
                 if dataset.edisp is not None:
                     events = self.sample_edisp(dataset.edisp, events)
 
+        # add background
         if dataset.background is not None:
+            dataset.models.append(FoVBackgroundModel(dataset_name="simulated-dataset"))
             events_bkg = self.sample_background(dataset)
             if events is not None:
                 events = EventList.from_stack([events_bkg, events])
             else:
                 events = events_bkg
+
+        if events is None:
+            events = EventList(Table())
 
         geom = dataset._geom
         selection = geom.contains(events.map_coord(geom))
--- a/gammapy/datasets/simulate.py
+++ b/gammapy/datasets/simulate.py
@@ -655,10 +655,7 @@ class SimulatedObservationMaker(MapDatasetEventSampler):
         pointing_icrs = observation.pointing.fixed_icrs
         geom = WcsGeom.create(
             skydir=pointing_icrs,
-            width=(
-                self.spatial_width.to_value(u.deg),
-                self.spatial_width.to_value(u.deg),
-            ),
+            width=self.spatial_width,
             binsz=self.spatial_bin_size.to_value(u.deg),
             frame="icrs",
             axes=[self.energy_axis],
--- a/gammapy/datasets/simulate.py
+++ b/gammapy/datasets/simulate.py
@@ -303,20 +303,23 @@ class MapDatasetEventSampler:
         events : `gammapy.data.EventList`
             Background events.
         """
-        background = dataset.npred_background()
 
-        temporal_model = ConstantTemporalModel()
+        table = Table()
+        if dataset.background:
+            background = dataset.npred_background()
 
-        table = self._sample_coord_time(background, temporal_model, dataset.gti)
+            temporal_model = ConstantTemporalModel()
 
-        table["ENERGY"] = table["ENERGY_TRUE"]
-        table["RA"] = table["RA_TRUE"]
-        table["DEC"] = table["DEC_TRUE"]
+            table = self._sample_coord_time(background, temporal_model, dataset.gti)
 
-        if self.keep_mc_id:
-            table["MC_ID"] = 0
-            table.meta["MID{:05d}".format(0)] = 0
-            table.meta["MMN{:05d}".format(0)] = dataset.background_model.name
+            table["ENERGY"] = table["ENERGY_TRUE"]
+            table["RA"] = table["RA_TRUE"]
+            table["DEC"] = table["DEC_TRUE"]
+
+            if self.keep_mc_id:
+                table["MC_ID"] = 0
+                table.meta["MID{:05d}".format(0)] = 0
+                table.meta["MMN{:05d}".format(0)] = dataset.background_model.name
 
         return EventList(table)
 
--- a/gammapy/datasets/simulate.py
+++ b/gammapy/datasets/simulate.py
@@ -8,7 +8,7 @@ from astropy.coordinates import SkyCoord, SkyOffsetFrame
 from astropy.table import Table
 from astropy.time import Time
 from gammapy import __version__
-from gammapy.data import EventList, Observation, PointingMode, observatory_locations
+from gammapy.data import EventList, PointingMode, observatory_locations
 from gammapy.maps import MapAxis, MapCoord, RegionNDMap, TimeMapAxis, WcsGeom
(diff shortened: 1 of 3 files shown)


gammapy/datasets/simulate.py
```diff
@@ -576,3 +578,99 @@ def run(self, dataset, observation=None):
...
+    def __init__(
+        self,
+        energy_axis_true,
+        energy_axis,
+        migra_axis,
+        map_width,
+        map_bin_size=0.02 * u.deg,
+        random_state="random-seed",
+    ):
+        self.random_state = get_random_state(random_state)
```
maxnoe, line 597: validation / type hints / documentation missing for arguments

gammapy/datasets/simulate.py
```diff
@@ -576,3 +578,99 @@ def run(self, dataset, observation=None):
...
+                events = EventList.from_stack([events_bkg, events])
+            else:
+                events = events_bkg
+
+        geom = dataset._geom
+        selection = geom.contains(events.map_coord(geom))
+        events = events.select_row_subset(selection)
+        events.table.sort("TIME")
+        events.table["EVENT_ID"] = np.arange(len(events.table))
+        observation._events = events
```
maxnoe, line 676: Note: better return a copy with events set instead of mutating the input observation

gammapy/datasets/simulate.py
```diff
@@ -576,3 +578,99 @@ def run(self, dataset, observation=None):
...
+        if observation.psf is not None:
+            components.append("psf")
+
+        maker = MapDatasetMaker(selection=components)
+        dataset = maker.run(dataset, observation)
+        models.append(FoVBackgroundModel(dataset_name="simulated-dataset"))
+        dataset.models = models
+
+        events = None
+        if len(models) > 1:
```
maxnoe, line 654: Note: while correct this logic is confusing. At this point I am asking "why only simulate stuff if we have atleast 2 models?

Ah, because 1 is always the background... there should be a better way to express this.

gammapy/datasets/simulate.py
```diff
@@ -576,3 +578,99 @@ def run(self, dataset, observation=None):
...
+
+        Returns
+        -------
+        events : `~gammapy.data.EventList`
+            Event list.
+        """
+        if models is None:
+            models = Models()
+
+        pointing_icrs = observation.pointing.get_icrs(
```
maxnoe, line 621: Since we explicitly only support fixed pointing in ra/dec, I think this should check and directly use `fixed_icrs` instead of asking for the icrs at mid of run time.

maxnoe:
I left some review comments for my past self, but please also have a look @registerrier @fabiopintore

QRemy:
This seems to be a duplicate of #3711.

gammapy/datasets/simulate.py
```diff
@@ -576,3 +577,102 @@ def run(self, dataset, observation=None):
...
+class SimulatedObservationMaker(MapDatasetEventSampler):
+    """
+    Sample event lists for a given observation.
+    """
+
+    def __init__(
+        self,
+        energy_axis_true,
+        energy_axis,
+        migra_axis,
```
registerrier, line 591: Just take the IRF migra-axis as it is an internal detail.

gammapy/datasets/simulate.py
```diff
@@ -574,3 +578,107 @@ def run(self, dataset, observation=None):
...
+            axes=[self.energy_axis],
+        )
+
+        components = ["exposure"]
+        axes = dict(
+            energy_axis_true=self.energy_axis_true,
+        )
+        if observation.edisp is not None:
+            components.append("edisp")
+            axes["mira_axis"] = observation.edisp.axes["migra"]
```
fabiopintore, line 641: This should be `migra_axis`

gammapy/datasets/simulate.py
fabiopintore: Yes, the logic would support the fact that the model always has the bacgrkound. Probably not the best solution, but I don't have any smart idea now...

fabiopintore (commented):
Thanks @maxnoe . My main doubt is why it's not taken into account the possibility to pass the IRF to the class.

gammapy/datasets/simulate.py
maxnoe: No:

```
In [3]: obs = Observation.read("./gammapy-datasets/dev/hess-dl3-dr1/data/hess_dl3_dr1_obs_id_020136.fits.gz")

In [4]: obs.edisp.axes["migra"]
Out[4]: 
MapAxis

	name       : migra     
	unit       : ''        
	nbins      : 160       
	node type  : edges     
	edges min  : 2.0e-01   
	edges max  : 5.0e+00   
	interp     : lin       

In [5]: obs.edisp.axes["migra_axis"]
KeyError                                  Traceback (most recent call last)
Cell In[5], line 1
----> 1 obs.edisp.axes["migra_axis"]

File ~/Projects/gammapy/gammapy/maps/axes.py:1741, in MapAxes.__getitem__(self, idx)
   1739         if ax.name == idx:
   1740             return ax
-> 1741     raise KeyError(f"No axes: {idx!r}")
   1742 elif isinstance(idx, slice):
   1743     axes = self._axes[idx]

KeyError: "No axes: 'migra_axis'"
```

maxnoe:
> Thanks @maxnoe . My main doubt is why it's not taken into account the possibility to pass the IRF to the class.

Because IRFs are a property of the observation. So you pass in an observation that contains everything but the events and for the given information, the events are then sampled.

registerrier:
What is the status here @maxnoe ?

maxnoe:
@registerrier I started adding in the missing docstrings and validation and I added a test, which currently fails... 

Deep down in `DataSetEvaluator`, it is missing an `exposure` and I cannot figure out why, the `dataset` created in the `run` method contains the exposure.

QRemy:
> Deep down in `DataSetEvaluator`, it is missing an `exposure` and I cannot figure out why, the `dataset` created in the `run` method contains the exposure.

@maxnoe, that was because a model was not contributing within in the dataset geom.
It should work now.

gammapy/datasets/simulate.py
```diff
@@ -574,3 +574,146 @@ def run(self, dataset, observation=None):
...
+    spatial_width : float or `~astropy.units.Quantity`
+        Spatial window size. A float is assumed to be in degree.
+    spatial_spatial_bin_size : float or `~astropy.units.Quantity`, optional
+        Pixel size used for sampling. A float is assumed to be in degree.
+        Default is 0.01 degree.
+    random_state : {int, 'random-seed', 'global-rng', `~numpy.random.RandomState`}, optional
+        Defines random number generator initialisation via the `~gammapy.utils.random.get_random_state` function.
+    oversample_energy_factor : int, optional
+        Defines an oversampling factor for the energies; it is used only when sampling
+        an energy-dependent time-varying source.
```
registerrier, line 603:
```suggestion
        an energy-dependent time-varying source. Default is 10.
```

gammapy/datasets/simulate.py
```diff
@@ -574,3 +574,146 @@ def run(self, dataset, observation=None):
...
         selection = geom.contains(events.map_coord(geom))
         return events.select_row_subset(selection)
+
+
+class SimulatedObservationMaker(MapDatasetEventSampler):
+    """
+    Sample event lists for a given observation and signal models.
+
+    Signal events are sampled from the given sky models and then
+    folded with the instrument response functions.
```
registerrier, line 584:
```suggestion
    Signal events are sampled from the predicted counts distribution given by the product of the sky models and the 
    expected exposure. They are then folded with the instrument response functions.
```
I propose this modification to make the process clearer.

gammapy/datasets/simulate.py
```diff
@@ -574,3 +574,146 @@ def run(self, dataset, observation=None):
...
+    spatial_spatial_bin_size : float or `~astropy.units.Quantity`, optional
+        Pixel size used for sampling. A float is assumed to be in degree.
+        Default is 0.01 degree.
+    random_state : {int, 'random-seed', 'global-rng', `~numpy.random.RandomState`}, optional
+        Defines random number generator initialisation via the `~gammapy.utils.random.get_random_state` function.
+    oversample_energy_factor : int, optional
+        Defines an oversampling factor for the energies; it is used only when sampling
+        an energy-dependent time-varying source.
+    t_delta : `~astropy.units.Quantity`, optional
+        Time interval used to sample the time-dependent source.
```
registerrier, line 605:
```suggestion
    t_delta : `~astropy.units.Quantity`, optional
        Time interval used to sample the time-dependent source. Default is 0.5 s.
```

registerrier: Note that for consistency with other arguments one could also allow for `float` assuming seconds.

gammapy/datasets/simulate.py
```diff
@@ -574,3 +574,146 @@ def run(self, dataset, observation=None):
...
+        self.energy_axis_true = energy_axis_true
+        self.energy_axis = energy_axis
+        self.spatial_width = u.Quantity(spatial_width, "deg")
+        self.spatial_bin_size = u.Quantity(spatial_bin_size, "deg")
+        self.oversample_energy_factor = oversample_energy_factor
+        self.t_delta = t_delta
+        self.keep_mc_id = keep_mc_id
+
+    def _create_dataset(self, observation, models=None, dataset_name=None):
+        """create dataest used for sampling"""
```
registerrier, line 632:
```suggestion
        """create dataset used for sampling."""
```

gammapy/datasets/simulate.py
```diff
@@ -574,3 +574,146 @@ def run(self, dataset, observation=None):
...
+
+        if observation.pointing.mode is not PointingMode.POINTING:
+            raise NotImplementedError(
+                "Only observations with fixed pointing in ICRS are supported"
+            )
+        pointing_icrs = observation.pointing.fixed_icrs
+        geom = WcsGeom.create(
+            skydir=pointing_icrs,
+            width=(
+                self.spatial_width.to_value(u.deg),
```
registerrier, line 659: You don't need to convert to degree here. `WcsGeom.create` accepts quantities too.

gammapy/datasets/simulate.py
```diff
@@ -574,3 +574,146 @@ def run(self, dataset, observation=None):
...
+            name=dataset_name,
+            **axes,
+        )
+
+        maker = MapDatasetMaker(selection=components)
+        dataset = maker.run(dataset, observation)
+        dataset.models = models
+        return dataset
+
+    def run(self, observation: Observation, models=None, dataset_name=None):
```
registerrier, line 690:
```suggestion
    def run(self, observation, models=None, dataset_name=None):
```

gammapy/datasets/simulate.py
```diff
@@ -574,3 +574,146 @@ def run(self, dataset, observation=None):
...
+        in true coordinates and energy. The true quantities are
+        then folded with the IRFs to obtain the observable quantities.
+
+        Parameters
+        ----------
+        observation : `~gammapy.data.Observation`
+            Observation to be simulated.
+        models : `~gammapy.modeling.Models`, optional
+            Models to simulate.
+            Can be None to only sample background events.
```
registerrier, line 703:
```suggestion
            Can be None to only sample background events. Default is None.
```

registerrier (commented):
Thanks @maxnoe & @QRemy . This looks good. I have left minor comments.

gammapy/datasets/simulate.py
```diff
@@ -574,3 +574,146 @@ def run(self, dataset, observation=None):
...
+            raise NotImplementedError(
+                "Only observations with fixed pointing in ICRS are supported"
+            )
+        pointing_icrs = observation.pointing.fixed_icrs
+        geom = WcsGeom.create(
+            skydir=pointing_icrs,
+            width=(
+                self.spatial_width.to_value(u.deg),
+                self.spatial_width.to_value(u.deg),
+            ),
```
QRemy, line 661:
```suggestion
            width=self.spatial_width,
```

PR #4925, opened 2023-11-23
Todo models from dict

MRegeard:
This PR deals with the two TODOs of `Line 233` and `Line 237` of `gammapy.modeling.models.core.py`. 

I refactor the two if statements by adding a new methods to `TemporalModel` and `SpatialModel` respectively to deal with the if statements that were in `ModelBase`. Then I call the `from_dict` method of `ModelBase` with `super()` in the two new methods.
 
I had to add a check (`spatial_data = data.get("spatial", data)` and `temporal_data = data.get("temporal", data)`) because sometime the whole dictionary is passed (i.e. `{"spatial": {...}}`) and sometime only the sub-dict is passed (`{...}`) in the previous example), cf. `TemplateNPredModel.from_dict` method. 

Finally I added `**kwargs` in the `BaseModel.from_dict` method to be able to pass the kwargs from `SpatialModel.from_dict` and `TemporalModel.from_dict` to the base class method. I added `**kwargs` in the signature of all the `from_dict` methods of class that inherit from `BaseModel` to match the base class signature, even though it is never used. But I don't really know what is our convention on this. 

I checked and normally all the tests passed on my computer. 
I think that this implementation don't break our API because only **kwargs are added.


cgalelli (approved):
Thanks @MRegeard good solution

gammapy/modeling/models/temporal.py
```diff
@@ -100,6 +100,23 @@ def to_dict(self, full_output=False):
...
+        """Create a temporal model from a dictionary.
+
+        Parameters
+        ----------
+        data : dict
+            Dictionary containing the model parameters.
+        **kwargs : dict
+            Keyword arguments passed to `~TemporalModel.from_parameters`.
+        """
+        kwargs = kwargs or {}
```
QRemy, line 114: in which case do we use kwargs ?

QRemy (dismissed):
Thanks @MRegeard, looks good, I have just one question :

Issue #4887, opened 2023-11-03
Remove deprecated features


AtreyeeS:
This is a reminder issue to remove the features deprecated since 1.1 before the next release

Astro-Kirsty:
I've made a list of all the deprecation's that need to be taken care of (and the tests associated):


- [gammapy.table.utils.table_from_row_data](https://github.com/gammapy/gammapy/blob/e43ec2735e74886283d672d59d874da9d43bf060/gammapy/utils/table.py#L96C33-L96C33)
- [gammapy.irf.io.load_cta_irfs](https://github.com/gammapy/gammapy/blob/e43ec2735e74886283d672d59d874da9d43bf060/gammapy/irf/io.py#L146)
- number of deprecation in [gammapy.data.observations](https://github.com/gammapy/gammapy/blob/e43ec2735e74886283d672d59d874da9d43bf060/gammapy/data/observations.py#L363C1-L386)
- [gammapy.astro.darkmatter.spectra.table_model](https://github.com/gammapy/gammapy/blob/e43ec2735e74886283d672d59d874da9d43bf060/gammapy/astro/darkmatter/spectra.py#L167)
- [gammapy.maps.core.apply_edisp](https://github.com/gammapy/gammapy/blob/e43ec2735e74886283d672d59d874da9d43bf060/gammapy/maps/core.py#L1407)
- [gammapy.estimators.points.sensitivity](https://github.com/gammapy/gammapy/blob/e43ec2735e74886283d672d59d874da9d43bf060/gammapy/estimators/points/sensitivity.py#L144)
- [gammapy.modeling.models.spatial.plot_iterative](https://github.com/gammapy/gammapy/blob/e43ec2735e74886283d672d59d874da9d43bf060/gammapy/modeling/models/spatial.py#L312)
- [gammapy.datasets.map.npred_signal](https://github.com/gammapy/gammapy/blob/e43ec2735e74886283d672d59d874da9d43bf060/gammapy/datasets/map.py#L489)
- [gammapy.modeling.fit.minuit](https://github.com/gammapy/gammapy/blob/e43ec2735e74886283d672d59d874da9d43bf060/gammapy/modeling/fit.py#L148)
- [gammapy.maps.axes attribute append](https://github.com/gammapy/gammapy/blob/e43ec2735e74886283d672d59d874da9d43bf060/gammapy/maps/axes.py#L109)
[...]

PR #4927, opened 2023-11-24
Remove deprecations in `gammapy.data`

Astro-Kirsty:
This PR partly resolves https://github.com/gammapy/gammapy/issues/4887

### DIFF
--- a/gammapy/data/observations.py
+++ b/gammapy/data/observations.py
@@ -368,9 +368,7 @@ class Observation:
     @property
     def observatory_earth_location(self):
         """Observatory location as an `~astropy.coordinates.EarthLocation` object."""
-        if self._location is not None:
-            return self._location
-        return self.pointing.location
+        self._location
 
     @lazyproperty
     def target_radec(self):
--- a/gammapy/data/observations.py
+++ b/gammapy/data/observations.py
@@ -365,43 +365,12 @@ class Observation:
         """Get the pointing in ICRS for given time."""
         return self.pointing.get_icrs(time, self.observatory_earth_location)
 
-    @lazyproperty
-    @deprecated(
-        "v1.1",
-        message="Use observation.pointing or observation.get_pointing_{{altaz,icrs}} instead",
-    )
-    def fixed_pointing_info(self):
-        """Fixed pointing information for this observation as a `~gammapy.data.FixedPointingInfo` object."""
-        return self._pointing
-
-    @property
-    @deprecated("v1.1", message="Use observation.get_pointing_icrs(time) instead")
-    def pointing_radec(self):
-        """Pointing RA / DEC sky coordinates as a `~astropy.coordinates.SkyCoord` object."""
-        return self.fixed_pointing_info.radec
-
-    @property
-    @deprecated("v1.1", message="Use observation.get_pointing_altaz(time) instead")
-    def pointing_altaz(self):
-        """Pointing ALT / AZ sky coordinates as a `~astropy.coordinates.SkyCoord` object."""
-        return self.fixed_pointing_info.altaz
-
-    @property
-    @deprecated("v1.1", message="Use observation.get_pointing_altaz(time).zen instead")
-    def pointing_zen(self):
-        """Pointing zenith angle sky as a `~astropy.units.Quantity`."""
-        return self.get_pointing_altaz(self.tmid).zen
-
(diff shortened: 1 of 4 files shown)


Astro-Kirsty:
@maxnoe, it would be great if you could check that I haven't missed anything, or misunderstood anything

gammapy/data/observations.py
```diff
@@ -365,43 +365,12 @@ def get_pointing_icrs(self, time):
...
-    @deprecated("v1.1", message="Use observation.get_pointing_altaz(time).zen instead")
-    def pointing_zen(self):
-        """Pointing zenith angle sky as a `~astropy.units.Quantity`."""
-        return self.get_pointing_altaz(self.tmid).zen
-
     @property
     def observatory_earth_location(self):
-        """Observatory location as an`~astropy.coordinates.EarthLocation` object."""
+        """Observatory location as an `~astropy.coordinates.EarthLocation` object."""
         if self._location is not None:
```
maxnoe, line 371: This is now required, so this property should just simplify to `return _self.location`.

gammapy/data/observations.py
```diff
@@ -365,43 +365,12 @@ def get_pointing_icrs(self, time):
...
             return self._location
-
-        # for now we catch the deprecation warning until we store location on the observation properly
-        with warnings.catch_warnings():
-            warnings.simplefilter("ignore", category=GammapyDeprecationWarning)
-            return self.fixed_pointing_info.location
+        return self.pointing.location
```
maxnoe, line 373: This is the wrong behavior. The whole section here should be removed. This will just raise an error as you removed the `location` property from the pointing.

gammapy/data/observations.py
```diff
@@ -365,43 +365,10 @@ def get_pointing_icrs(self, time):
...
-        """Observatory location as an`~astropy.coordinates.EarthLocation` object."""
-        if self._location is not None:
-            return self._location
-
-        # for now we catch the deprecation warning until we store location on the observation properly
-        with warnings.catch_warnings():
-            warnings.simplefilter("ignore", category=GammapyDeprecationWarning)
-            return self.fixed_pointing_info.location
+        """Observatory location as an `~astropy.coordinates.EarthLocation` object."""
+        self._location
```
registerrier, line 371:
```suggestion
        return self._location
```

registerrier (commented):
Thanks @Astro-Kirsty . See inline comment.

Astro-Kirsty:
I have adapted the tests accordingly. 

There is one failing test, which I believe can be completely removed but I wanted a second opinion:
`gammapy/data/tests/test_pointing.py::test_altaz_without_location`

QRemy:
> test_altaz_without_location

Maybe the test should use `pointing.get_altaz()` instead, but then if fails.
So do we need  `pointing.get_altaz()` to return `_legacy_location` in that case for backward compatibly ?

Astro-Kirsty:
I've added a small fix for if the location is None, because this is ultimately where the issue comes from. 
Let me know what you think

maxnoe:
@Astro-Kirsty If you are now going over the deprecations in `gammapy.data.pointing`, there is a lot more connected in `__init__` and `{to,from}_gadf_header`, the points where a `GammapyDeprecationWarning` is issues manually and not via the method decorators.

The failing tests is one for these deprecated behaviors, so should just be removed.

The new, wanted API is tested in separate tests. So I think it's fine to remove the deprecated code paths and remove the tests in `gammapy.data.tests.test_pointing` that fail due to that.

There should be no other tests failing, since we run with the deprecation warnings as errors, so nothing else should use these deprecated functionality internally.

Astro-Kirsty:
Okay thanks @maxnoe. 

Maybe I missed it but I do not see a test for the new API which has no location?

maxnoe:
> Maybe I missed it but I do not see a test for the new API which has no location?

Yes, that's because the new API requires the location intentionally as location is a property of the observatory, not the pointing.

See discussion in https://github.com/gammapy/gammapy/issues/3767 and https://github.com/gammapy/gammapy/pull/4220

Astro-Kirsty:
Okay, thank you! 
I will read into those

I've adapt the PR now

QRemy (approved):
Thanks @Astro-Kirsty, no further comment

PR #4934, opened 2023-11-24
Fix broken links in docs

### LINKED ISSUE #4922: Broken links in the documentation
**Gammapy version**
The dev version

**Bug description**
I have found two issues:
- [x] The link towards the ReflectedBackgroundMaker in https://docs.gammapy.org/dev/tutorials/api/makers.html#background-estimation seems to point towards the RindgBgMaker
- [x] In the page https://docs.gammapy.org/dev/api/gammapy.modeling.Parameter.html?highlight=priors, a link towards the `Prior` class is proposed. However the class is called `PriorParameter` and it seems to be not yet exposed.
- [x] In https://docs.gammapy.org/dev/user-guide/model-gallery/spectral/plot_absorbed.html, the link towards the EBL models list is broken

MRegeard:
This PR fix #4922. 

In order for the `EBL_DATA_BUILTIN` link to work, I have to expose the `EBL_DATA_BUILTIN` in `gammapy.modeling.models`. I don't think there is any other solution for that. 

Also the link for `prior` is working "virtually". It just that the prior classes are not exposed. But when they will, the link will work. 

PS: I also added `TemplateNDSpectralModel` in the `__all__` of `gammapy.modeling.models.spectral.py` to remove the complaint for accessing a protected member in `gammapy.modeling.models.__init__.py`.

### DIFF
--- a/gammapy/modeling/models/spectral.py
+++ b/gammapy/modeling/models/spectral.py
@@ -55,14 +55,18 @@ __all__ = [
     "SuperExpCutoffPowerLaw4FGLDR3SpectralModel",
     "SuperExpCutoffPowerLaw4FGLSpectralModel",
     "TemplateSpectralModel",
+    "TemplateNDSpectralModel",
+    "EBL_DATA_BUILTIN",
 ]
 
-EBL_DATA_BUILTIN = {}
-EBL_DATA_BUILTIN["franceschini"] = "$GAMMAPY_DATA/ebl/ebl_franceschini.fits.gz"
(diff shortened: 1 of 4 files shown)


gammapy/modeling/models/spectral.py
```diff
@@ -55,14 +55,18 @@
     "SuperExpCutoffPowerLaw4FGLDR3SpectralModel",
     "SuperExpCutoffPowerLaw4FGLSpectralModel",
     "TemplateSpectralModel",
+    "TemplateNDSpectralModel",
```
registerrier, line 58: So finally it is in. It is a new addition. No problem for me.

registerrier (approved):

PR #4936, opened 2023-11-27
PiecewiseNormSpectralModel serialising interp

katrinstreil:
**Description**

This pull request slightly changes the serialisation of the PiecewiseNormSpectralModel. The interpolation scaling is written/read to/from the dict.


QRemy (changes requested):
Thanks @katrinstreil, just made a suggestion to set the default if interp is not given in the dict. Could you also add a test ?

QRemy (approved):
Thanks @katrinstreil, no further comment.

PR #4937, opened 2023-11-29
Fix import of angular_separation for astropy 6

maxnoe:
**Description**

This pull request fixes the import of `angular_separation` for astropy 6 (but it also works with 5.x).

**Dear reviewer**


Astro-Kirsty (approved):
Thanks @maxnoe

PR #4940, opened 2023-11-30
Fix MapEvaluator for the apply_edisp=False case

QRemy:
Fix MapEvaluator for the apply_edisp=False case, to avoid inconsistant geometry error because npred has the exposure geom for the models with apply_edisp=False and the counts geom for the others.

### DIFF
--- a/gammapy/datasets/evaluator.py
+++ b/gammapy/datasets/evaluator.py
@@ -7,6 +7,7 @@ from astropy.coordinates import angular_separation
 from astropy.utils import lazyproperty
 from regions import CircleSkyRegion
 import matplotlib.pyplot as plt
+from gammapy.irf import EDispKernel
 from gammapy.maps import HpxNDMap, Map, RegionNDMap, WcsNDMap
 from gammapy.modeling.models import PointSpatialModel, TemplateNPredModel
 from .utils import apply_edisp
@@ -188,6 +189,7 @@ class MapEvaluator:
             self.edisp = edisp.get_edisp_kernel(
                 position=self.model.position, energy_axis=energy_axis
             )
+            del self._edisp_diagonal
 
         # lookup psf
         if psf and self.model.spatial_model:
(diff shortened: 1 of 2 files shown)


gammapy/modeling/models/tests/test_cube.py
```diff
@@ -532,6 +532,18 @@ def test_compute_npred(diffuse_evaluator):
         assert_allclose(out.data.sum(), 1.106403e12, rtol=5e-3)
         assert_allclose(out.data[0, 0, 0], 8.778828e09, rtol=5e-3)
 
+    @staticmethod
+    def test_apply_edisp_false(diffuse_evaluator):
+        # TODO: needs_update=True if apply_irf change ?
+        diffuse_evaluator.model.apply_irf["edisp"] = False
+        diffuse_evaluator.use_cache = False
+        del diffuse_evaluator.methods_sequence
```
registerrier, line 540: Why do you change these settings here?

registerrier: I would rather recreate a `MapEvaluator` as modifying a fixture might cause issues depending on the order the tests are run.

registerrier: Could this rather be tested in test_evaluator.py directly? It does not necessarily need a template model, right?

registerrier (commented):

gammapy/modeling/models/tests/test_cube.py
QRemy: ok I created a separate fixture, but I would prefer to keep the test here to avoid core duplication recreating psf/edisp/model.

gammapy/datasets/evaluator.py
```diff
@@ -366,7 +375,13 @@ def apply_edisp(self, npred):
...
             Predicted counts in reconstructed energy bins.
         """
-        return apply_edisp(npred, self.edisp)
+        if self.model.apply_irf["edisp"]:
+            return apply_edisp(npred, self.edisp)
+        else:
+            if "energy_true" in npred.geom.axes.names:
+                return apply_edisp(npred, self._edisp_diagonal)
+            else:
+                return npred
```
AtreyeeS, line 384: This was the previous behaviour, no? Strange it is not covered by the tests

AtreyeeS (approved):
Thanks @QRemy !
Why can't you use the `evaluator` fixture in the `test_evaluator.py` ?

gammapy/datasets/evaluator.py
QRemy: the previous behaviour was equivalent to return npred even if "energy_true" in npred.geom.axes.names, which lead to a geom inconsistency

PR #4942, opened 2023-11-30
Parallel support for FluxProfileEstimation

QRemy:
Replacing the loop over regions to use run_multiprocessing instead is straightforward. However I left a  TODO for later because I have to test if it would be more efficient to not compute datasets_to_fit in parallel to avoid copying the full datasets. For example for diffuse emission the dataset would be the full galactic plane so we may run into memory issue or just be slower because the copy takes too long.

### DIFF
--- a/gammapy/estimators/points/profile.py
+++ b/gammapy/estimators/points/profile.py
@@ -1,8 +1,10 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 """Tools to create profiles (i.e. 1D "slices" from 2D images)."""
+from itertools import repeat
 import numpy as np
 from astropy import units as u
 from regions import CircleAnnulusSkyRegion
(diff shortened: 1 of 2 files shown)


QRemy:
I noticed than setting `n_jobs=2` on the class will also set `n_jobs=2` to the `FluxPointEstimator` created by the `FluxProfileEstimator` but using the context manager set only `n_jobs=2` for the `FluxProfileEstimator` and the original default on the `FluxPointEstimator`.  I don't think it's a bug , my guess is that the child processes don't see the modifications of the global variables as they store a copy of them.

registerrier (commented):
Thanks @QRemy . I suppose the choice of computing the spectra at first or in parallel depends a bit on the situation. Actually computing the spectre might even be the bottleneck in most use cases since profiles are frequently used with few energy bins.

registerrier (approved):

PR #4944, opened 2023-12-01
Expose more features

QRemy:
Expose more features in particular from gammapy.utils


Astro-Kirsty (commented):
Thanks @QRemy. 
To correctly expose functions from the `utils` folder we also need to adapt the `docs/api-reference/utils.rst` to include the file name there. Then it should be correctly exposed in the API!

PR #4945, opened 2023-12-04
Return the coordinate for the doubling time as an astropy.Time object

cgalelli:
This pull request modifies the `estimators.utils.compute_lightcurve_doublingtime` function so that the coordinates at which the doubling and halving time are found are returned as a time object (relative to the reference time of the lightcurve)


Astro-Kirsty (approved):
Thanks @cgalelli

AtreyeeS (approved):
Thanks @cgalelli ! No more comments

PR #4963, opened 2023-12-07
Expose remaining `utils` functions in API

Astro-Kirsty:
**Description**
There were a number of utils functions which were not included in the rst therefore were unexposed.


QRemy (approved):

PR #4964, opened 2023-12-07
Adapt accordion id in HowTo

Astro-Kirsty:
**Description**
This PR is to fix the accordion IDs for the HowTo. Currently, they follow a numbering scheme which is inconvenient when one wants to add an extra HowTo to the list. This PR changes the id's such that the names are applicable to each HowTo. 

Note: the order of the HowTo's determines the order on the docs page, NOT the numbering system.

### DIFF
--- a/docs/user-guide/howto.rst
+++ b/docs/user-guide/howto.rst
@@ -18,7 +18,7 @@ Please give feedback and suggest additions to this page!
 
 
 .. accordion-header::
-    :id: collapseHowToOne
+    :id: HowToGammapy
     :title: Spell and pronounce Gammapy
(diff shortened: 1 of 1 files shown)


docs/user-guide/howto.rst
```diff
@@ -386,13 +385,3 @@ using a dummy phase column.
 
 .. accordion-footer::
 
-.. accordion-header::
```
registerrier, line 389: why removing this?

Astro-Kirsty: Because it was a duplicate -- see line 43

PR #4965, opened 2023-12-07
Expose more features useful for UL computation

QRemy:
Expose more features useful for UL computation : 
-Parameter estimator 
-stat_profile_ul_scipy


Astro-Kirsty (approved):
Thanks @QRemy. 
This is a good idea!

As you suggest, I agree with adding it to the tutorial

AtreyeeS (approved):
Thanks @QRemy ! Merging this PR now, can you please modify the tutorial as you proposed a follow up PR?

PR #4966, opened 2023-12-08
Fix circular import issue in `data/utils`

Astro-Kirsty:
**Description**
After exposing the `get_irfs_features` there is a circular import issue. 
This PR fixes this by simply moving the related import inside the function.

### DIFF
--- a/gammapy/data/utils.py
+++ b/gammapy/data/utils.py
@@ -2,7 +2,6 @@
 import numpy as np
 import astropy.units as u
 from astropy.table import Table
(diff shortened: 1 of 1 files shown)


QRemy:
Do you have an example to trigger the circular import ? I can't reproduce it

Astro-Kirsty:
It was occurring when I did `Map.read()`

registerrier:
It looks cleaner to delay the import of IRF classes if we are in the utils subpackage. Let's merge.

PR #4967, opened 2023-12-08
Fix TODO for docstring in `excess.py`

Astro-Kirsty:
**Description**

This pull request is to fix one of the TODO.
It also fixes the docstring in `ExcessMapEstimator` and the `copy()` example in `Observation`.

**Dear reviewer**
Do we want to expose `convolved_map_dataset_counts_statistics`, because it is currently hidden from the API?

### DIFF
--- a/gammapy/estimators/map/excess.py
+++ b/gammapy/estimators/map/excess.py
@@ -20,8 +20,24 @@ log = logging.getLogger(__name__)
 
 
 def convolved_map_dataset_counts_statistics(dataset, kernel, mask, correlate_off):
-    """Return CountsDataset objects containing smoothed maps from the MapDataset."""
-    # TODO: Add a full docstring
+    """Return CountsDataset objects containing smoothed maps from the MapDataset.
+
+    Parameters
+    ----------
+    dataset : `~gammapy.datasets.MapDataset` or `~gammapy.datasets.MapDatasetOnOff`
+        Map dataset.
+    kernel : `~astropy.convolution.Tophat2DKernel`
+        Kernel.
+    mask : `~gammapy.maps.Map`
+        Mask map.
+    correlate_off : bool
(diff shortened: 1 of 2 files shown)


QRemy (changes requested):
Thanks @Astro-Kirsty, the description of the function is outdated :

gammapy/estimators/map/excess.py
```diff
@@ -20,8 +20,24 @@
...
 def convolved_map_dataset_counts_statistics(dataset, kernel, mask, correlate_off):
-    """Return CountsDataset objects containing smoothed maps from the MapDataset."""
-    # TODO: Add a full docstring
+    """Return a `CountsStatistic` object.
+
+    Parameters
+    ----------
+    dataset : `~gammapy.datasets.MapDataset` or `~gammapy.datasets.MapDatasetOnOff`
+        Map dataset.
+    kernel : `~astropy.convolution.Tophat2DKernel`
```
AtreyeeS, line 29: I think this function should work correctly for other kernels as well

AtreyeeS (dismissed):
Thanks @Astro-Kirsty !

AtreyeeS (approved):
I think the popular use case would be to use this functionality thought the `MapEstimators`, which does provide the correlated maps. I don't see the point in exposing this `convolved_map_dataset_counts_statistics`, but have no strong opinions on this

I guess the docstring adaptations in `gammapy/data/observations.py` are not related to this PR?

registerrier (approved):

I would not expose the convolved_map_count_statistic function as it is an implementation detail.
do you confirm that the change in observation.py is expected here?

Astro-Kirsty:
Yes, the `observations.py` was not building correctly in the docs. So I just decided to fix that in this PR too

PR #4951, opened 2023-12-05
BLD: cleanup unused build-time dependencies

neutrinoceros:
**Description**
Replaces #4949

**Dear reviewer**

PR #4969, opened 2023-12-08
MNT: fix an implicit incompatibility with Python 3.12 (pkg_resources isn't shipped with stdlib)

neutrinoceros:
**Description**

Noticed this while working on #4951:
`__init__.py` has an undeclared dependency on `pkg_resources` which is a subpackaged shipped with `setuptools`. Python 3.12 was the first release which did `setuptools` builtin (it was never part of the standard library, but for a long time it looked like it did, so I've seen exactly this problem before).
I would recommend testing systematically against Python 3.12, but this is out of scope for this PR.

### DIFF
--- a/gammapy/__init__.py
+++ b/gammapy/__init__.py
@@ -32,13 +32,14 @@ the following sub-packages (e.g. `gammapy.makers`):
 
 __all__ = ["__version__", "song"]
 
-import pkg_resources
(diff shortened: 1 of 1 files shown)


registerrier (approved):
Thanks @neutrinoceros . This is helpful.

Indeed we need to test for 3.11 and 3.12. This is now done in #4998

PR #4970, opened 2023-12-08
Adapt FluxPointsDataset to fit light curves

AtreyeeS:
After #4828, `FluxPointsDataset` can now work with a time axis.
This PR adapt `FluxPointsDataset`  to directly use lightcurves instead of a joint fitting of multiple flux points datasets, and adapts the tutorial

The `plot_residuals` visualisation is complex for multi dimensional flux points, so I raise an error. Another option could be to plot a grid, but I don't know if its useful.

### DIFF
--- a/gammapy/datasets/core.py
+++ b/gammapy/datasets/core.py
@@ -379,7 +379,7 @@ class Datasets(collections.abc.MutableSequence):
         return copy.deepcopy(self)
 
     @classmethod
-    def read(cls, filename, filename_models=None, lazy=True, cache=True):
+    def read(cls, filename, filename_models=None, lazy=True, cache=True, **kwargs):
         """De-serialize datasets from YAML and FITS files.
 
         Parameters
@@ -392,6 +392,9 @@ class Datasets(collections.abc.MutableSequence):
             Whether to lazy load data into memory. Default is True.
         cache : bool
             Whether to cache the data after loading. Default is True.
+        kwargs : dict
+            Keyword arguments passed to individual datasets write methods
+
 
         Returns
         -------
@@ -411,7 +414,7 @@ class Datasets(collections.abc.MutableSequence):
                 data["filename"] = str(make_path(path / data["filename"]))
 
             dataset_cls = DATASET_REGISTRY.get_cls(data["type"])
-            dataset = dataset_cls.from_dict(data, lazy=lazy, cache=cache)
+            dataset = dataset_cls.from_dict(data, lazy=lazy, cache=cache, **kwargs)
             datasets.append(dataset)
 
         datasets = cls(datasets)
@@ -428,6 +431,7 @@ class Datasets(collections.abc.MutableSequence):
         overwrite=False,
         write_covariance=True,
         checksum=False,
+        **kwargs,
     ):
         """Serialize datasets to YAML and FITS files.
 
@@ -444,6 +448,9 @@ class Datasets(collections.abc.MutableSequence):
         checksum : bool
             When True adds both DATASUM and CHECKSUM cards to the headers written to the FITS files.
             Default is False.
+        kwargs : dict
+            Keyword arguments passed to individual datasets write methods
+
         """
         path = make_path(filename)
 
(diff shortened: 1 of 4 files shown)


gammapy/datasets/core.py
```diff
@@ -392,6 +392,9 @@ def read(cls, filename, filename_models=None, lazy=True, cache=True):
             Whether to lazy load data into memory. Default is True.
         cache : bool
             Whether to cache the data after loading. Default is True.
+        kwargs : dict
+            Keyword arguments passed to individual datasets write methods
```
QRemy, line 396: what append if the different datasets have different kwargs  ?

gammapy/datasets/core.py
```diff
@@ -379,7 +379,7 @@ def copy(self):
         return copy.deepcopy(self)
 
     @classmethod
-    def read(cls, filename, filename_models=None, lazy=True, cache=True):
+    def read(cls, filename, filename_models=None, lazy=True, cache=True, **kwargs):
```
QRemy, line 382: I would remove that and ensure that the format is handled correctly by each read/write method.

gammapy/datasets/flux_points.py
```diff
@@ -485,6 +519,8 @@ def plot_spectrum(self, ax=None, kwargs_fp=None, kwargs_model=None):
         kwargs_model : dict, optional
             Keyword arguments passed to `gammapy.modeling.models.SpectralModel.plot` and
             `gammapy.modeling.models.SpectralModel.plot_error` to configure the plot style. Default is None.
+        axis_name : str
+            Axis along which to plot the flux points for multiple axes. Default is energy
```
cgalelli, line 523:
```suggestion
            Axis along which to plot the flux points for multiple axes. Default is energy.
```

gammapy/datasets/flux_points.py
```diff
@@ -218,14 +238,21 @@ def read(cls, filename, name=None, format="gadf-sed"):
         )
 
     @classmethod
-    def from_dict(cls, data, **kwargs):
+    def from_dict(cls, data, format="gadf-sed", **kwargs):
```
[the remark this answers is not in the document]
cgalelli: Maybe the possibility of setting it in the function call should be left open, in case the dict (for some reason) does not contain the format. I agree that the first choice should be read from dict.

It should be easy just adding a `if data.get("format")` statement at the start

cgalelli (commented):
Thanks, @AtreyeeS this is an extremely useful addition that will simplify a lotthe treatment of lightcurves! Seems very good to me, apart from the issue on serialization already raised by Quentin

AtreyeeS:
Yes, the issue with the formats is annoying. Actually this should be fixed in serialisation of lightcurves. Currently, `lc_1d.write(filename)` leads to an error
```
and you have to rather specify `lc_1d.write(filename, format="lightcurve")`

I had opened an issue about this proposing to set a `format` on each class, and @adonath suggested `FluxPoints` could also guess the format... see: https://github.com/gammapy/gammapy/issues/3871

File: gammapy/datasets/flux_points.py (hunk 4 above)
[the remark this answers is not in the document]
Comment (QRemy) [reply gammapy/datasets/flux_points.py:241]: if the dict or yaml file does not provide the correct format and the user know about it then they should update the dict/ yaml file (rather than setting the format on read/from_dict)

**Comment by AtreyeeS** [comment]
I rebased after #4992 and now the serialisation should be better.

File: gammapy/datasets/core.py [hunk 5]
```diff
@@ -428,6 +428,7 @@ def write(
         overwrite=False,
         write_covariance=True,
         checksum=False,
+        **kwargs,
```
QRemy, line 431:
```suggestion
```

gammapy/datasets/tests/test_flux_points.py
```diff
@@ -119,10 +118,46 @@ def test_flux_point_dataset_creation():
...
+        rtol=1e-3,
+    )
+    assert_allclose(flux_points_dataset.stat_sum(), 193.8093, rtol=1e-3)
+    assert_allclose(
+        flux_points_dataset.residuals()[0][0].value, 9.94782173e-12, rtol=1e-5
+    )
+    Datasets([flux_points_dataset]).write(
+        filename=tmp_path / "tmp_datasets.yaml",
+        filename_models=tmp_path / "tmp_models.yaml",
+        format="lightcurve",
```
QRemy, line 144:
```suggestion
```

QRemy (approved):
Thanks @AtreyeeS no further comments.

PR #4971, opened 2023-12-08
Use `FixedPointingInfo` in notebook

AtreyeeS:
This PR adapts tutorials to use `FixedPointingInfo` instead of `SkyCoord`.
@registerrier the current implementation creates a `FixedPointingInfo` from a `SkyCoord` while raising the warning. I did not change it.
One option might be to not remove support for `SkyCoord` in 1.2

### DIFF
--- a/examples/tutorials/analysis-1d/spectrum_simulation.py
+++ b/examples/tutorials/analysis-1d/spectrum_simulation.py
@@ -91,7 +91,12 @@ check_tutorials_setup()
 # Define simulation parameters parameters
 livetime = 1 * u.h
 
-pointing = SkyCoord(0, 0, unit="deg", frame="galactic")
+pointing_position = SkyCoord(0, 0, unit="deg", frame="galactic")
+# We want to simulate an observation pointing at a fixed position in the sky.
+# For this, we use the `FixedPointingInfo` class
+pointing = FixedPointingInfo(
+    fixed_icrs=pointing_position.icrs,
+)
 offset = 0.5 * u.deg
 
 # Reconstructed and true energy axis
--- a/examples/tutorials/analysis-3d/simulate_3d.py
+++ b/examples/tutorials/analysis-3d/simulate_3d.py
@@ -61,7 +61,7 @@ import matplotlib.pyplot as plt
 
 # %matplotlib inline
 from IPython.display import display
-from gammapy.data import Observation, observatory_locations
+from gammapy.data import FixedPointingInfo, Observation, observatory_locations
 from gammapy.datasets import MapDataset
 from gammapy.irf import load_irf_dict_from_file
 from gammapy.makers import MapDatasetMaker, SafeMaskMaker
@@ -92,7 +92,12 @@ irfs = load_irf_dict_from_file(
 
 # Define the observation parameters (typically the observation duration and the pointing position):
 livetime = 2.0 * u.hr
-pointing = SkyCoord(0, 0, unit="deg", frame="galactic")
+pointing_position = SkyCoord(0, 0, unit="deg", frame="galactic")
+# We want to simulate an observation pointing at a fixed position in the sky.
+# For this, we use the `FixedPointingInfo` class
+pointing = FixedPointingInfo(
+    fixed_icrs=pointing_position.icrs,
+)
 
 # Define map geometry for binned simulation
 energy_reco = MapAxis.from_edges(


bkhelifi:
Hello @AtreyeeS ,
The change is straightforward... good for me, then.
You mention a warning, what is it? Is it a Deprecation warning? It seems strange as your usage seems to respect perfectly the documentation, so a warning should not be raised. Is not it? Do I miss something?

AtreyeeS:
Hi @bkhelifi ! There are two issues related to the introduction of `FixedPointingInfo` and deprecation of  `SkyCoord` as pointing arguments.

The first is that we have been exposing deprecated features in our notebook, which is obviously a bad idea and what this PR fixes. It would be nice to have the CI automatically complain about deprecated features to prevent such issues in future

The second is, should the `SkyCoord` support be at all deprecated? A `SkyCoord` pointing is self evident,  a `FixedPointingInfo` is internal Gammapy API and not very transparent to the user. In that case, we can simply change the warning from deprecation to warning here:

https://github.com/gammapy/gammapy/blob/994098e2b3c7b8fcdb98cc6557ecb9fab18cb343/gammapy/data/observations.py#L270-L275

I see no obvious issue to this for the standard IACT use case. Of course, this fails for WCD and drift observations, but we don't support those correctly anyways.

Astro-Kirsty (commented):
Thanks @AtreyeeS. 

The following tutorials also need adapting I believe:
[cta_sensitivity](https://docs.gammapy.org/dev/tutorials/analysis-1d/cta_sensitivity.html), [event_sampling_nrg_depend_models](https://docs.gammapy.org/dev/tutorials/analysis-3d/event_sampling_nrg_depend_models.html), [light_curve_simulation](https://docs.gammapy.org/dev/tutorials/analysis-time/light_curve_simulation.html)

maxnoe:
> The second is, should the SkyCoord support be at all deprecated? 
I think so, yes. Because of consistency. It's easier to understand for users, that what you pass as argument to the constructor is the same as the property of `Observation`.

> A SkyCoord pointing is self evident, a FixedPointingInfo is internal Gammapy API and not very transparent to the user.

A `SkyCoord` is too general. It can be in `AltAz`, it can be in `ICRS`, it can be even lots of coordinates over time (a skycoord in ICRS with many coordinates and different `obstimes` could be a time dependent pointing table for example).

So you at least have to verify the `SkyCoord`, that it is a single skycoord in either ICRS or AltAz (and not any other of the dozens of frames astropy supports...).
Which is exactly what `FixedPointingInformation` does.

Astro-Kirsty (commented):
[This line](https://github.com/gammapy/gammapy/blob/25a920335d425b072eb9afe444eb3bba297b7550/examples/tutorials/analysis-3d/event_sampling_nrg_depend_models.py#L265) also needs adapting?

Astro-Kirsty (dismissed):

examples/tutorials/analysis-3d/event_sampling_nrg_depend_models.py
```diff
@@ -262,7 +262,8 @@
 path = Path("$GAMMAPY_DATA/cta-caldb")
 irf_filename = "Prod5-South-20deg-AverageAz-14MSTs37SSTs.180000s-v0.1.fits.gz"
 
-pointing = SkyCoord(100.0, 30.0, frame="icrs", unit="deg")
+pointing_position = SkyCoord(100.0, 30.0, frame="icrs", unit="deg")
+pointing = FixedPointingInfo(fixed_icrs=pointing_position.icrs)
```
maxnoe, line 266: You don't need `.icrs` if the original position is already in `icrs`.

examples/tutorials/analysis-3d/event_sampling_nrg_depend_models.py
```diff
@@ -262,7 +262,8 @@
 path = Path("$GAMMAPY_DATA/cta-caldb")
 irf_filename = "Prod5-South-20deg-AverageAz-14MSTs37SSTs.180000s-v0.1.fits.gz"
 
-pointing = SkyCoord(100.0, 30.0, frame="icrs", unit="deg")
+pointing_position = SkyCoord(100.0, 30.0, frame="icrs", unit="deg")
```
maxnoe, line 265: ICRS is the default frame, I'd also highly recommend always using keyword arguments for coordinates, as it is not always obvious which order they are in (e.g. the argument order of `AltAz` coordinates is actually az, alt).

`SkyCoord(ra=100 * u.deg, dec=30 * u.deg)`

examples/tutorials/analysis-3d/event_sampling_nrg_depend_models.py
```diff
@@ -261,7 +262,8 @@
...
```
QRemy, line 265:
```suggestion
pointing_position = SkyCoord(ra=100 * u.deg, dec=30 * u.deg)
```

examples/tutorials/analysis-3d/event_sampling_nrg_depend_models.py
```diff
@@ -261,7 +262,8 @@
...
+pointing = FixedPointingInfo(fixed_icrs=pointing_position.icrs)
```
QRemy, line 266:
```suggestion
pointing = FixedPointingInfo(fixed_icrs=pointing_position)
```

QRemy (approved):
Implemented the last comments, merging now

Issue #4920, opened 2023-11-22
Cannot use `TemplateSpatialModel` as spatial model in `TSMapEstimator`


ArmelleJB:
**Gammapy version**
gammapy 1.1

**Bug description**
Trying to use a template instead of a gaussian or point source to make significance map with `TSMapEstimator` - the position of the template exists, but there is an error for setting the position (no setter) :
```
model.spatial_model.position = geom.center_skydir
[...]
AttributeError: can't set attribute
```

**Expected behavior**
Should work as `GaussianSpatialModel` but a `position` function is redefined that maybe overwrite the inherited one. Adding
```
@position.setter
    def position(self, skycoord):
        """Spatial model center position"""
        coord = skycoord.transform_to(self.frame)
        self.lon_0.quantity = coord.data.lon
        self.lat_0.quantity = coord.data.lat
```
to the `TemplateSpatialModel` may solve the problem. 

**To Reproduce**
```
ts_spatial_model = TemplateSpatialModel(image)
ts_spectral_model = PowerLawSpectralModel(
    index=2.4, amplitude="1e-11 cm-2 s-1 TeV-1", reference="1 TeV"
)

model_ts = SkyModel(
    spatial_model=ts_spatial_model,
    spectral_model=ts_spectral_model,
    name="model-ts",
)

estimator = TSMapEstimator(
    model=model_ts,
    energy_edges=[1.0,100] * u.TeV,
)

dataset.models=bkg_model
result = estimator.run(dataset)
print(result)
result["sqrt_ts"].plot(add_cbar=True, stretch="linear")
```

QRemy:
One solution could be to create internally a new map, shifting the geom center to the new position.

adonath:
@QRemy I think we can actually just add positional parameters to the `TemplateSpatialModel`. It is a well defined use case to fit the position of a template as well.

QRemy:
> @QRemy I think we can actually just add positional parameters to the `TemplateSpatialModel`. It is a well defined use case to fit the position of a template as well.

I agree if frozen but default.

QRemy:
> > @QRemy I think we can actually just add positional parameters to the `TemplateSpatialModel`. It is a well defined use case to fit the position of a template as well.
> 
> I agree if frozen but default.

AtreyeeS:
Resolved for v1.2 in https://github.com/gammapy/gammapy/pull/4973

PR #4973, opened 2023-12-11
Add position as a parameter for TemplateSpatialModel

AtreyeeS:
This PR allows the fitting of the position for template spatial models, and addresses #4920

### DIFF
--- a/gammapy/estimators/map/tests/test_ts.py
+++ b/gammapy/estimators/map/tests/test_ts.py
@@ -356,6 +356,7 @@ def test_with_TemplateSpatialModel():
     estimator = TSMapEstimator(
         model=sky_model,
         energy_edges=[1.0, 5.0] * u.TeV,
+        n_jobs=4,
     )
 
     result = estimator.run(dataset)
--- a/gammapy/estimators/map/tests/test_ts.py
+++ b/gammapy/estimators/map/tests/test_ts.py
@@ -3,7 +3,7 @@ import pytest
 import numpy as np
 from numpy.testing import assert_allclose
 import astropy.units as u
-from astropy.coordinates import Angle
+from astropy.coordinates import Angle, SkyCoord
 from gammapy.datasets import MapDataset, MapDatasetOnOff
 from gammapy.estimators import TSMapEstimator
 from gammapy.irf import EDispKernelMap, PSFMap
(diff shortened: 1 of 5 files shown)


QRemy (commented):
Thanks @AtreyeeS, could you fix the failing test and add you to check if we can read a yaml model where the position is not defined ?

AtreyeeS:
Thanks @QRemy ! Indeed the serialisation was broken.
I have added a test for the to and from dict now.

QRemy (changes requested):
I think it is possible to have it work without addins the parameters argument on init :

gammapy/modeling/models/spatial.py
```diff
@@ -1325,8 +1327,36 @@ def __init__(
...
         super().__init__(**kwargs)
 
+    def __str__(self):
+        width = self.map.geom.width
+        data_min = np.nanmin(self.map.data)
+        data_max = np.nanmax(self.map.data)
+
+        prnt = (
+            f"{self.__class__.__name__} model summary:\n "
+            f"Reference position: {self.position} \n "
```
adonath, line 1344: I think this is confusing. The WCS reference position doe snot change and is not equivalent to `self.position`.

gammapy/modeling/models/spatial.py
```diff
@@ -1325,8 +1327,36 @@ def __init__(
...
```
AtreyeeS, line 1344:
```suggestion
            f"Model center: {self.position} \n "
```

gammapy/modeling/models/spatial.py
AtreyeeS: @adonath The reference was a bad naming. ( I was following the `LightCurveTemplateTemporalModel`, where it makes more sense). Updated to `Model position`

QRemy (dismissed):
Thanks @AtreyeeS, no further comment from my side.

registerrier (approved):
Thanks @AtreyeeS . This is nice. 

The way the coordinate rotation is applied should be explained with a caveat on the applicability. Or conversely one can determine the rotation angles and apply them to the coords.

gammapy/estimators/map/tests/test_ts.py
```diff
@@ -341,3 +342,21 @@ def test_MapDatasetOnOff_error():
     ts_estimator = TSMapEstimator()
     with pytest.raises(TypeError):
         ts_estimator.run(dataset=dataset_on_off)
+
+
+@requires_data()
+def test_with_TemplateSpatialModel():
```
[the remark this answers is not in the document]
AtreyeeS: I downsampled the dataset. Now it takes 2s

QRemy (approved):
Thanks @AtreyeeS, merging now

PR #4974, opened 2023-12-11
Remove unnecessary `# TODO`

Astro-Kirsty:
This PR is to remove some `# TODO` that are no longer necessary because they are already implemented.

For `gammapy/modeling/fit.py`, I believe all the convenience access functions are needed. I personally utilise them all the time.


registerrier (approved):
Thanks @Astro-Kirsty Looks good.

PR #4975, opened 2023-12-11
Reduce memory usage of get_psf_kernel

QRemy:
Reduce memory usage of get_psf_kernel by looping over energy to compute the psf_kernel. So upsampling and coordinates computation are only performed for one image at the time. The loop could eventually be parallelised later. 

For the test I did with a large dataset on my laptop there was already a gain in time because it did not swap.

I will add a commit to check if removing the upsampling change a lot the results.

### DIFF
--- a/gammapy/irf/psf/map.py
+++ b/gammapy/irf/psf/map.py
@@ -17,17 +17,18 @@ from .kernel import PSFKernel
 __all__ = ["PSFMap", "RecoPSFMap"]
 
 
-PSF_MAX_OVERSAMPLING = 4
+PSF_MAX_OVERSAMPLING = 4  # for backward compatibility
 
 
 def _psf_upsampling_factor(psf, geom, position, energy=None, precision_factor=12):
+    """Minimal factor between the bin half-width of the geom and the median R68% containment radius."""
     if energy is None:
         energy = geom.axes[psf.energy_name].center
     psf_r68 = psf.containment_radius(
         0.68, geom.axes[psf.energy_name].center, position=position
     )
-    psf_r68 = np.percentile(psf_r68, 50)
-    base_factor = (2 * psf_r68 / geom.pixel_scales.max()).to_value("")
+    psf_r68_median = np.percentile(psf_r68, 50)
+    base_factor = (2 * psf_r68_median / geom.pixel_scales.max()).to_value("")
     factor = np.minimum(
         int(np.ceil(precision_factor / base_factor)), PSF_MAX_OVERSAMPLING
     )
@@ -275,7 +276,7 @@ class PSFMap(IRFMap):
             Oversampling factor to compute the PSF.
             Default is None and it will be computed automatically.
         precision_factor : int, optional
-            Minimal factor between the bin half-width of the geom and the R68% containment radius.
+            Factor between the bin half-width of the geom and the median R68% containment radius.
             Used only for the oversampling method. Default is 10.
 
         Returns
--- a/gammapy/irf/psf/map.py
+++ b/gammapy/irf/psf/map.py
@@ -20,7 +20,7 @@ __all__ = ["PSFMap", "RecoPSFMap"]
 PSF_MAX_OVERSAMPLING = 4
 
 
-def _psf_upsampling_factor(psf, geom, position, energy=None, precision_factor=10):
+def _psf_upsampling_factor(psf, geom, position, energy=None, precision_factor=12):
     if energy is None:
         energy = geom.axes[psf.energy_name].center
     psf_r68 = psf.containment_radius(
@@ -252,7 +252,7 @@ class PSFMap(IRFMap):
         max_radius=None,
         containment=0.999,
         factor=None,
-        precision_factor=10,
+        precision_factor=12,
     ):
         """Return a PSF kernel at the given position.
 
--- a/gammapy/irf/psf/map.py
+++ b/gammapy/irf/psf/map.py
@@ -273,6 +273,7 @@ class PSFMap(IRFMap):
             the `max_radius` argument. Default is 0.999.
         factor : int, optional
             Oversampling factor to compute the PSF.
+            Default is None and it will be computed automatically.
         precision_factor : int, optional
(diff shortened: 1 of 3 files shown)


adonath:
Is the code within the loop the bottleneck? Can you please double check how the performance behaves for a smaller dataset?

adonath:
Thanks @QRemy, looks plausible! Is it worth adding a more general option to `Map.interp_by_coord()?` Or can't we directly rely on `Map.resample()` doesn't it use the upsampling internally as well?

registerrier:
The PSF kernel size should not really depend on the parent dataset geometry. I find it surprising that his can become the memory bottleneck. Having a loop over energy might be a CPU time issue as well.

QRemy:
> The PSF kernel size should not really depend on the parent dataset geometry. I find it surprising that his can become the memory bottleneck. Having a loop over energy might be a CPU time issue as well.

The bin width should be the same between the kernel map and the parent dataset geometry, no ?

registerrier:
> > The PSF kernel size should not really depend on the parent dataset geometry. I find it surprising that his can become the memory bottleneck. Having a loop over energy might be a CPU time issue as well.
> 
> The bin width should be the same between the kernel map and the parent dataset geometry, no ?

Yes. But the size of the kernel does not need to be the same as the the global map.

QRemy:
I remove the loop but still computes the coords in 2d and the energy is given by the axis center (instead of the ND array from get_coords), I added a global config to set the upsampling factor.

PSF_UPSAMPLING_FACTOR = 4 (Default)
newplot (3)

PSF_UPSAMPLING_FACTOR = 1
newplot (2)

gammapy/irf/psf/map.py
```diff
@@ -243,14 +262,19 @@ def get_psf_kernel(
...
+            Oversampling factor to compute the PSF.
+            If None it is given by PSF_UPSAMPLING_FACTOR which is 4 by default.
+            If PSF_UPSAMPLING_FACTOR is set to None it will be automatically estimated.
         Returns
         -------
         kernel : `~gammapy.irf.PSFKernel`
             The resulting kernel.
         """
-        # TODO: try to simplify...is the oversampling needed?
+        if factor is None:
```
adonath, line 273: I think this behavior is confusing. What about:

- Introduce a new `precision_factor` and `factor_max` argument
- Forward both to `_psf_upsampling_factor`
- Deprecate `factor`...

gammapy/irf/psf/map.py
```diff
@@ -243,14 +272,19 @@ def get_psf_kernel(
             across all energies is used. The radius can be overwritten using
             the `max_radius` argument. Default is 0.999.
         factor : int, optional
-            Oversampling factor to compute the PSF. Default is 4.
+            Oversampling factor to compute the PSF.
```
QRemy, line 275:
```suggestion
            Oversampling factor to compute the PSF.
            Default is None and it will be computed automatically.
```

gammapy/irf/psf/map.py
```diff
@@ -16,6 +17,25 @@
...
+PSF_MAX_OVERSAMPLING = 4
+
+
+def _psf_upsampling_factor(psf, geom, position, energy=None, precision_factor=12):
+    if energy is None:
+        energy = geom.axes[psf.energy_name].center
+    psf_r68 = psf.containment_radius(
+        0.68, geom.axes[psf.energy_name].center, position=position
+    )
+    psf_r68 = np.percentile(psf_r68, 50)
```
bkhelifi, line 29: In fact, a mediane of the psf_r68 is computed, is not it? maybe you can rename the variable...

gammapy/irf/psf/map.py
```diff
@@ -16,6 +17,25 @@
...
+    base_factor = (2 * psf_r68 / geom.pixel_scales.max()).to_value("")
+    factor = np.minimum(
+        int(np.ceil(precision_factor / base_factor)), PSF_MAX_OVERSAMPLING
+    )
+    if isinstance(geom, HpxGeom):
```
bkhelifi, line 34: Maybe an "if else" can be introduced, in order to avoid to compute psf.containment_radius in case of HEALPix

gammapy/irf/psf/map.py
```diff
@@ -16,6 +17,25 @@
...
+        factor = int(2 ** np.ceil(np.log(factor) / np.log(2)))
+    return factor
+
+
 class IRFLikePSF(PSF):
```
bkhelifi, line 39: Can you add a small docstring to this class?

bkhelifi (changes requested):
Thanks @QRemy... It seems ready to be merged. I have left very minor comments...

gammapy/irf/psf/map.py
```diff
@@ -16,6 +17,25 @@
 __all__ = ["PSFMap", "RecoPSFMap"]
 
 
+PSF_MAX_OVERSAMPLING = 4
```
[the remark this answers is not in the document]
QRemy: added for backward compatibility

gammapy/irf/psf/map.py
```diff
@@ -16,6 +17,25 @@
...
+
+
+def _psf_upsampling_factor(psf, geom, position, energy=None, precision_factor=12):
```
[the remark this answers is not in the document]
QRemy: added  a description

gammapy/irf/psf/map.py
QRemy: added

gammapy/irf/psf/map.py
QRemy: here the if statement makes the computed factor a multiple of 2 so it can work in Healpix but we still need to compute the factor as in the wcs case

gammapy/irf/psf/map.py
QRemy: it is unrelated so it can be done in a sperate PR

bkhelifi (approved):
Thanks a lot @QRemy . It seems OK for me

PR #4977, opened 2023-12-12
Reduce memory usage of solid_angle 

QRemy:
Computes the solid angle only in 2d to save memory for regular geom.
(the multi-resolution maps use case is untested anyway).

### DIFF
--- a/gammapy/maps/wcs/geom.py
+++ b/gammapy/maps/wcs/geom.py
@@ -139,6 +139,11 @@ class WcsGeom(Geom):
         return self.axes.shape[::-1] + (1, 1)
 
     @property
+    def data_shape_image(self):
+        """Shape of data of the spatial axes and unit non-spatial axes."""
+        return (1,) * len(self.axes) + self.data_shape[len(self.axes) :]
+
+    @property
     def _shape(self):
         npix_shape = tuple([np.max(self.npix[0]), np.max(self.npix[1])])
         return npix_shape + self.axes.shape
@@ -811,7 +816,7 @@ class WcsGeom(Geom):
 
     @lazyproperty
     def _solid_angle(self):
-        coord = self.get_coord(mode="edges").skycoord
+        coord = self.to_image().get_coord(mode="edges").skycoord
 
         # define pixel corners
         low_left = coord[..., :-1, :-1]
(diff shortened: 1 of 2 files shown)


QRemy:
Reference (PR #4975)
Screenshot 2023-12-11 at 20 02 35
memray lg PR4975

PR4977
Screenshot 2023-12-12 at 10 32 51
newplot

gammapy/maps/wcs/geom.py
```diff
@@ -138,6 +138,11 @@ def data_shape_axes(self):
         """Shape of data of the non-spatial axes and unit spatial axes."""
         return self.axes.shape[::-1] + (1, 1)
 
+    @property
+    def data_shape_image(self):
```
registerrier, line 142: Maybe this would be better if private.

registerrier (commented):
thanks @QRemy . I think this should keep the same default behavior as before.

Maybe you could add a `sparse` argument on `solid_angle` to return only the flat array or the full one by resizing the array. This would reduce performance in the default case and would require adapting the various places where we use `solid_angle`. 
 
For irregular arrays (and this case should be tested indeed) it should still compute the full solid angle.

gammapy/maps/wcs/geom.py
QRemy: this is similar to `data_shape_axes` which is public

gammapy/maps/wcs/geom.py
```diff
@@ -811,7 +816,7 @@ def solid_angle(self):
 
     @lazyproperty
     def _solid_angle(self):
-        coord = self.get_coord(mode="edges").skycoord
+        coord = self.to_image().get_coord(mode="edges").skycoord
```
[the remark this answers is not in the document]
QRemy: Now it is applied only `if self.is_regular`
I'm strongly against keeping the old behaviour as default for regular maps because there is no gain from that, just a waste of memory.

bkhelifi:
Thanks @QRemy... So, with the implemented code, the heap size is still around 6-8G?

QRemy:
> Thanks @QRemy... So, with the implemented code, the heap size is still around 6-8G?

The part of the profile around 6-8G is not affected by this PR but by #4975

registerrier (approved):

I suppose we could give more information in the docstring regarding the new behaviour.

gammapy/maps/wcs/geom.py
```diff
@@ -801,7 +806,9 @@ def to_binsz(self, binsz):
     def solid_angle(self):
         """Solid angle array as a `~astropy.units.Quantity` in ``sr``.
 
-        The array has the same dimension as the WcsGeom object.
+        The array has the same dimension as the WcsGeom object
+        if the spatial shape is not unique along the extra axis,
+        otherwise the array shape matches of the spatial dimensions.
```
registerrier, line 811:
```suggestion
        otherwise the array shape matches the spatial dimensions.
```

registerrier (approved):
Thanks @QRemy !

PR #4979, opened 2023-12-12
Consistency format Part 1

### LINKED ISSUE #4870: Consistency with `format` in docstring

For example here:

Astro-Kirsty:
This PR is part 1 to resolve #4870.

**Description**

**Dear reviewer**
In this PR there are a number of `format` options. Could you please double check that all options are actually listed within the docstring.

### DIFF
--- a/gammapy/estimators/map/asmooth.py
+++ b/gammapy/estimators/map/asmooth.py
@@ -39,9 +39,9 @@ class ASmoothMapEstimator(Estimator):
     spectrum : `SpectralModel`
         Spectral model assumption.
     method : {'asmooth', 'lima'}
-        Significance estimation method.
+        Significance estimation method. Default is 'lima'.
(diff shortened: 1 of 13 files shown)


cgalelli (dismissed):
Thanks @Astro-Kirsty !

gammapy/estimators/map/asmooth.py
```diff
@@ -39,9 +39,9 @@ class ASmoothMapEstimator(Estimator):
     spectrum : `SpectralModel`
         Spectral model assumption.
     method : {'asmooth', 'lima'}
```
registerrier, line 41:
```suggestion
    method : {'lima', 'asmooth'}
```
Typical syntax is that the first entry in the list is the default.

registerrier (approved):

PR #4980, opened 2023-12-12
Consistency with format Part 2

### LINKED ISSUE #4870: Consistency with `format` in docstring

Astro-Kirsty:
This PR is part 2 to resolve #4870.

**Description**

### DIFF
--- a/gammapy/maps/hpx/core.py
+++ b/gammapy/maps/hpx/core.py
@@ -303,7 +303,7 @@ class HpxMap(Map):
             Set INDXSCHM to SPARSE and sparsify the map by only
             writing pixels with non-zero amplitude.
             Default is False.
-        format : {'fgst-ccube', 'fgst-template', 'gadf', None}
+        format : {None, 'fgst-ccube', 'fgst-template', 'gadf'}
             FITS format convention. If None this will be set to the
             default convention of the map. Default is None.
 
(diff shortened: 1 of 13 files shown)


gammapy/maps/region/geom.py
```diff
@@ -678,7 +678,7 @@ def to_hdulist(self, format="ogip", hdu_bands=None, hdu_region=None):
 
         Parameters
         ----------
-        format : {"gadf", "ogip", "ogip-sherpa"}, optional
+        format : {"gadf", "ogip", "ogip-sherpa"}
```
registerrier, line 681:
```suggestion
        format : {"ogip", "gadf", "ogip-sherpa"}
```

gammapy/maps/hpx/core.py
```diff
@@ -303,7 +303,7 @@ def to_hdu(self, hdu=None, hdu_bands=None, sparse=False, format=None):
             Set INDXSCHM to SPARSE and sparsify the map by only
             writing pixels with non-zero amplitude.
             Default is False.
-        format : {'fgst-ccube', 'fgst-template', 'gadf', None}, optional
+        format : {'fgst-ccube', 'fgst-template', 'gadf', None}
```
registerrier, line 306:
```suggestion
        format : {None, 'fgst-ccube', 'fgst-template', 'gadf'}
```

registerrier (commented):
My main concern regards the list of of possible choices, giving just the list of choices suggest s that the expected input is a single string among those provided.

Astro-Kirsty:
> Thanks @Astro-Kirsty . This looks good. My main concern regards the list of of possible choices, giving just the list of choices suggest s that the expected input is a single string among those provided.

Yes, in general I agree too. The other option is to use the following, however, I feel sometimes the options get lost in the docstring that way:
image

registerrier (approved):

PR #4981, opened 2023-12-12
Consistency with format Part 3

### LINKED ISSUE #4870: Consistency with `format` in docstring

Astro-Kirsty:
This PR is part 3 to resolve #4870.

**Description**

### DIFF
--- a/gammapy/modeling/parameter.py
+++ b/gammapy/modeling/parameter.py
@@ -733,7 +733,7 @@ class Parameters(collections.abc.Sequence):
         ----------
         name : str or list, optional
             Name of the parameter. Default is None.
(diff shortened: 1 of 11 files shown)


gammapy/modeling/parameter.py
```diff
@@ -733,7 +733,7 @@ def select(
         ----------
         name : str or list, optional
             Name of the parameter. Default is None.
-        type : {None, spatial, spectral, temporal}, optional
+        type : {None, spatial, spectral, temporal}
```
registerrier, line 736:
```suggestion
        type : {None, "spatial", "spectral", "temporal"}
```

registerrier (dismissed):

PR #4982, opened 2023-12-13
Control jobs creation by child processes

### LINKED ISSUE #4968: Nested Estimators create child processes
I noticed than setting `n_jobs=2` on the `FluxProfileEstimator ` will also set `n_jobs=2` to the `FluxPointEstimator` created by the `FluxProfileEstimator` but using the context manager set only `n_jobs=2` for the `FluxProfileEstimator` and the original default  `n_jobs=1` on the `FluxPointEstimator`.  I don't think it's a bug , my guess is that the child processes don't see the modifications of the global variables as they store a copy of them.

Same problem for the `LightCurveEstiamtor`.

QRemy:
This PR prevents nested Estimators to create child processes with ray by default.
Resolves #4968 as the default behaviour is now the same setting n_jobs on the estimator or with the parallel_manager.
Requires #4942 as the FluxProfileEstimator is also affected.


QRemy:
Another way could be to force processes=1 in run_multiprocessing if there is not only the main process running but that will kill use cases where it is reasonable to have other ray actors running in parallel of those launched by run_multiprocessing. Here we want to avoid it because these jobs are small tasks so I think the current solution is fine.

bkhelifi (approved):
Thanks a lot @QRemy

PR #4983, opened 2023-12-13
Adjust getting-started page

### LINKED ISSUE #4838: Review/Idea for the documentation getting started page
Currently the `getting started` page of the [gammapy documentation](https://docs.gammapy.org/dev/getting-started/index.html) displays mostly informations on how to install gammapy and on tutorials. 

The first thing that one sees is how to install (quickly) gammapy by using conda or pip, without creating a virtual environment, even though we know that this is not how one should do it. This part of the page is referred as `Installation`, which don't makes it clear that it is not the way we recommend to install gammapy. 
The second part of the page is called `Quickstart Setup` and shows how to install gammapy with the environment file that we provide, how to activate the environment, download the data, the notebooks, export the gammapy variable, etc. 
The last part is just an overview of the tutorials. 
We also have pages that are dedicated to the topics like installation of gammapy, virtuam environment, etc. 

Overall, I find it really confusing, thinking as a beginner in gammapy or even in python, on what I should do to get started with gammapy. 

This page is really the front door of the package, so it should be extremely clear on what one should do depending on its knowledge of python and gammapy. Especially on the installation part, on the main page we show two ways of installing gammapy. We also have a page `installation`, that show how to install python with different package manager but still without environment. And then we have a page that show in details how to install gammapy with a virtual environment. 

My idea would be to just show a quick (but good) way of installing gammapy with an environment on the main page. This would be something like the `Quickstart Setup` section that we already have. And then enter in the details in the other page that we have.

Astro-Kirsty:
**Description**
This PR is to resolve #4838. It adjusts the `Getting started` page to be far more simple and realistic to how we would typically install gammapy. 

Then the other pages (`Installation`, `Virtual Environments` etc) are adjusted accordingly. 

https://astro-kirsty.github.io/gammapy-docs-preview/

### DIFF
--- a/docs/getting-started/index.rst
+++ b/docs/getting-started/index.rst
@@ -21,26 +21,18 @@ There are various ways for users to install Gammapy. We recommend setting up an
 environment using either conda or mamba. Here are two methods to quickly install Gammapy.
 
 .. panels::
-    :column: col-12 p-3
+    :column: col-lg-6 col-md-6 col-sm-12 col-xs-12 p-3
 
     Working with conda?
     ^^^^^^^^^^^^^^^^^^^
 
-    First set up and activate your environment:
-
-    .. code-block:: bash
-
-        $ conda create -n gammapy
-        $ conda activate gammapy
-
-    Gammapy can then be installed with `Anaconda <https://docs.continuum.io/anaconda/>`__:
+    Gammapy can be installed with `Anaconda <https://docs.continuum.io/anaconda/>`__:
 
     .. code-block:: bash
 
         $ conda install -c conda-forge gammapy
 
     ---
-    :column: col-12 p-3
 
     Prefer pip?
     ^^^^^^^^^^^
--- a/docs/getting-started/index.rst
+++ b/docs/getting-started/index.rst
@@ -17,19 +17,30 @@ Getting started
 Installation
 ------------
 
+There are various ways for users to install Gammapy. We recommend setting up an
+environment using either conda or mamba. Here are two methods to quickly install Gammapy.
+
 .. panels::
-    :column: col-lg-6 col-md-6 col-sm-12 col-xs-12 p-3
+    :column: col-12 p-3
 
     Working with conda?
     ^^^^^^^^^^^^^^^^^^^
 
-    Gammapy can be installed with `Anaconda <https://docs.continuum.io/anaconda/>`__ or Miniconda:
+    First set up and activate your environment:
+
+    .. code-block:: bash
+
+        $ conda create -n gammapy
+        $ conda activate gammapy
+
+    Gammapy can then be installed with `Anaconda <https://docs.continuum.io/anaconda/>`__:
 
     .. code-block:: bash
 
         $ conda install -c conda-forge gammapy
 
     ---
+    :column: col-12 p-3
 
     Prefer pip?
     ^^^^^^^^^^^
--- a/docs/getting-started/index.rst
+++ b/docs/getting-started/index.rst
@@ -17,8 +17,8 @@ Getting started
 Installation
 ------------
 
-There are various ways for users to install Gammapy. We recommend setting up an
-environment using either conda or mamba. Here are two methods to quickly install Gammapy.
+There are various ways for users to install Gammapy. **We recommend setting up a virtual
+environment using either conda or mamba.** Here are two methods to quickly install Gammapy.
 
 .. panels::
     :column: col-lg-6 col-md-6 col-sm-12 col-xs-12 p-3
--- a/docs/getting-started/quickstart.rst
+++ b/docs/getting-started/quickstart.rst
@@ -19,9 +19,10 @@ execute the following commands in the terminal:
 
 .. note::
 
-    For Apple silicon M1 (`arm64`) architectures you also have to open the
-    environment file and delete the `sherpa` entry, as currently there are
-    no conda packages available. However you can later install `sherpa`
+    For gammapy versions **prior to v1.2**, if you're using
+    Apple silicon M1 (`arm64`) architectures you have to open the
+    environment file and delete the `sherpa` entry. Currently there are
+    no conda packages available. You can then install `sherpa`
     in the environment using `python -m pip install sherpa`.
 
 
--- a/docs/getting-started/quickstart.rst
+++ b/docs/getting-started/quickstart.rst
@@ -11,6 +11,19 @@ execute the following commands in the terminal:
     $ curl -O https://gammapy.org/download/install/gammapy-|release|-environment.yml
     $ conda env create -f gammapy-|release|-environment.yml
 
+.. note::
+
+    On Windows, you have to open up the conda environment file and delete the
+    lines with ``sherpa`` and ``healpy``. Those are optional dependencies that
+    currently aren't available on Windows.
+
+.. note::
+
+    For Apple silicon M1 (`arm64`) architectures you also have to open the
+    environment file and delete the `sherpa` entry, as currently there are
+    no conda packages available. However you can later install `sherpa`
+    in the environment using `python -m pip install sherpa`.
+
 
 The best way to get started and learn Gammapy are the :ref:`tutorials`.
 You can download the Gammapy tutorial notebooks and the example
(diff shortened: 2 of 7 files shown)


MRegeard:
Since we have a few dependencies in the explicit pip channel of the environment file, I moved sherpa in this section so that we don't need the note on sherpa with Apple Silicon in the quickstart page of the documentation.
Now sherpa installation should work on all Linux and Apple OS.

cgalelli (dismissed):
Thanks @Astro-Kirsty for this useful set of adjustments!

adonath:
Thanks @Astro-Kirsty! Can you locally build the documentation and push a preview to your GitHub? I did this in the past e.g. here: https://github.com/adonath/gammapy-docs-preview

It should work like:

```bash
make docs-sphinx
cp -r docs/_build/html path-to/gammapy-docs-preview
```

Then we can access the docs online at ```Astro-Kirsty.github.io/gammapy-docs-preview```.

Astro-Kirsty:
I've made the repo [here](https://github.com/Astro-Kirsty/gammapy-docs-preview/tree/main/html), trying to figure out how to create the clickable link

AtreyeeS (commented):
Hi @Astro-Kirsty ! Looking at your generated docs page, the instructions on the "most common way" ie, https://docs.gammapy.org/dev/getting-started/index.html#quickstart-setup are a bit hidden. 

Your `QuickStart` section only shows how to install the data and tutorials https://astro-kirsty.github.io/gammapy-docs-preview/getting-started/index.html#quickstart-setup

I would propose you can simply rename `QuickStart` to `Recommended SetUp`, which should make it clearer, no?

environment-dev.yml
```diff
@@ -82,3 +81,4 @@ dependencies:
       - ray
       - PyGithub
       - pypandoc
+      - sherpa
```
bkhelifi, line 84: Why sherpa is a dev dependency, as stated l49?

Astro-Kirsty: I'm unsure why that file changed, will fix! Thanks for picking up on it

Astro-Kirsty:
Thanks @AtreyeeS, I adjusted it

Note: the doc link on my github is no longer updated!

cgalelli (dismissed):
Thanks @Astro-Kirsty . The page seems intuitive and easy to read for me now. No further comments from me

MRegeard (dismissed):
Thanks @Astro-Kirsty, this looks good!

registerrier (commented):
Thanks @Astro-Kirsty . I am not sure I understand why you removed the instructions for the creation of the environment in the recommended setup. It is no longer visible from the main page then, no?

Astro-Kirsty:
It was according to #4838. 
One of the reasons was because the page seemed too confusing for beginners to gammapy. As there is a 'Virtual Environments' section we thought it best to leave this out, especially because the conda install would install the latest version of gammapy. 

If you think it would be best to change it back, let us know!

bkhelifi:
Also, there is no mention of mamba. Is it intended?

adonath:
> Also, there is no mention of mamba. Is it intended?

The lates versions of conda include the mamba solver. I'm not sure it is needed anymore.

Astro-Kirsty:
mamba is not mentioned on the initial page but is mentioned in "installation" where a full guide of how to install is included.

Astro-Kirsty:
What is the current opinion here? Can we merge?
@bkhelifi @cgalelli @MRegeard @registerrier @AtreyeeS

cgalelli:
What would be the conflicts with #5113  ?

Astro-Kirsty:
No major conflict. That changes the 'Installation' description, which can be adjusted once this PR has been merged. 
Or @bkhelifi can add commit his changes to this PR?

Astro-Kirsty:
I wouldn't have been in the call. 
Present preview is the correct one, see here https://astro-kirsty.github.io/gammapy-docs-preview/

MRegeard:
Hi @Astro-Kirsty, 
Should we put at least these two lines in the `Recommended setup` of the getting started pages :
```
$ curl -O https://gammapy.org/download/install/gammapy-X.Y.Z-environment.yml
$ conda env create -f gammapy-X.Y.Z-environment.yml
```
and redirect user to the `virtual environment` page for further informations ? 

Because currently this part of the page only shows how to download the tutorials and datasets, and furthermore assume a conda environment (without showing how to create one). 

I think installation with in an environment should appear in this page !

Astro-Kirsty:
Good idea @MRegeard! 
I will adapt accordingly

docs/getting-started/quickstart.rst
```diff
@@ -1,39 +1,19 @@
...
 
 .. substitution-code-block:: bash
 
     $ curl -O https://gammapy.org/download/install/gammapy-|release|-environment.yml
     $ conda env create -f gammapy-|release|-environment.yml
 
-.. note::
-
-    On Windows, you have to open up the conda environment file and delete the
-    lines with ``sherpa`` and ``healpy``. Those are optional dependencies that
```
AtreyeeS, line 20: Windows users will still face issues with `healpy`, no?

AtreyeeS (approved):
This looks ok from my side (please see inline question about specific instructions to windows users)
@adonath can you comment once if the instructions about conda environments are okay?

docs/getting-started/quickstart.rst
MRegeard: I think so. There are no file for windows on conda-forge an PyPi states OS: POSIX.

docs/getting-started/index.rst
```diff
@@ -17,19 +17,30 @@ Getting started
...
+
 .. panels::
-    :column: col-lg-6 col-md-6 col-sm-12 col-xs-12 p-3
+    :column: col-12 p-3
 
     Working with conda?
     ^^^^^^^^^^^^^^^^^^^
 
-    Gammapy can be installed with `Anaconda <https://docs.continuum.io/anaconda/>`__ or Miniconda:
+    First set up and activate your environment:
```
adonath, line 29: I would not show this here. Creating virtual environments and installing the software are two different things. I would keep the current panels and just add a note on the recommendation of the using a virtual environment like:

Screenshot 2024-02-28 at 10 40 50 AM

docs/getting-started/quickstart.rst
adonath: I think the note for windows should remain where it is.

docs/getting-started/quickstart.rst
```diff
@@ -1,39 +1,19 @@
...
-    currently aren't available on Windows.
-
-.. note::
```
adonath, line 23: Are sherpa builds for M1 available by now for all OS versions? If not we should keep the comment or adapt all env files in https://github.com/gammapy/gammapy-webpage/tree/gh-pages/download/install to instal sherpa via pip (but only if there are wheels, I have not checked lately...)

docs/getting-started/quickstart.rst
```diff
@@ -1,39 +1,19 @@
...
-    For Apple silicon M1 (`arm64`) architectures you also have to open the
-    environment file and delete the `sherpa` entry, as currently there are
-    no conda packages available. However you can later install `sherpa`
-    in the environment using `python -m pip install sherpa`.
-
-Once the environment has been created you can activate it using:
-
-.. substitution-code-block:: bash
-
-    $ conda activate gammapy-|release|
```
adonath, line 34: Activating the environment needs to happen first, otherwise the later download will fail, because the `gammapy` command is not available.

adonath (changes requested):
Thanks @Astro-Kirsty, I have left some comments.

docs/getting-started/quickstart.rst
MRegeard: Indeed sherpa M1 build is not available with conda. I moved the sherpa to the pip channel (in this PR). However, I'm not sure about the wheels. But this was also done in the 1.2rc1 environment file and the installation succeed.

cgalelli (approved):
Thanks @Astro-Kirsty this final version looks nice to me

adonath (approved):
Thanks a lot @Astro-Kirsty, no further comments from my side!

MRegeard (approved):
Thank you very much @Astro-Kirsty !

PR #4986, opened 2023-12-14
Update mainly `gammapy.astro` TODO

Astro-Kirsty:
**Description**

This pull request fixes the TODO and docstrings in `gammapy.astro`, as well as some other simple TODO in `gammapy.datasets`, `gammapy.modeling` and `gammapy.utils`.

### DIFF
--- a/gammapy/astro/population/simulate.py
+++ b/gammapy/astro/population/simulate.py
@@ -47,14 +47,15 @@ def make_catalog_random_positions_cube(
 
     Parameters
     ----------
-    size : int
-        Number of sources.
+    size : int, optional
+        Number of sources. Default is 100.
     dimension : {1, 2, 3}
-        Number of dimensions.
-    distance_max : `~astropy.units.Quantity`
-        Maximum distance.
+        Number of dimensions. Default is 3.
+    distance_max : `~astropy.units.Quantity`, optional
+        Maximum distance. Default is "1 pc".
     random_state : {int, 'random-seed', 'global-rng', `~numpy.random.RandomState`}
         Defines random number generator initialisation.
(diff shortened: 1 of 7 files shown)


Astro-Kirsty:
As the galactocentric value has been changed to the one from astropy, the tests in astro folder needed to be changed. 
This is something we need to be aware of in other cases too, do we truly want the one from astropy then?

bkhelifi:
Hi,
I am wondering why we are not using anymore `astropy.coordinates.Galactocentric`... We rely on the astropy coordinate library in our code. Why not here?

adonath:
>I am wondering why we are not using anymore astropy.coordinates.Galactocentric... We rely on the astropy coordinate library in our code. Why not here?

This code was written before Astropy existed at all and never refactored.

Astro-Kirsty:
I have decided I will open a new PR

The issue with having `D_SUN_TO_GALACTIC_CENTER` set to a specific value is that if astropy ever changes the value, then all related tests will need to change. Which can be quite a hassle 

Actually after some thought and discussion with Regis, I think this is need of bigger discussion than just changing this one value. 
I plan to open an issue regarding this. Basically we have various parameters through out the code, such as `D_SUN_TO_GALACTIC_CENTER`, which means it could actually be useful to have a config file where all these parameters are defined. Then the user can change them if they wish on their local machines.

bkhelifi:
I understand... But if astropy changes it, it is super important to know! is not it?
Maybe one can do both in our test... use a fixed value (compliant with the current value of astropy) in the test, and add a test to be sure that the astropy value does not change (I mean, not at the sub-AU accuracy! we are not so precise in gamma astronomy...). This is just an idea

PR #4990, opened 2023-12-15
Adapt TODO in `gammapy.astro` and other

Astro-Kirsty:
**Description**
This pull request fixes the TODO and docstrings in `gammapy.astro`, as well as some other simple TODO in `gammapy.datasets`,  and `gammapy.modeling`.

### DIFF
--- a/gammapy/astro/population/simulate.py
+++ b/gammapy/astro/population/simulate.py
@@ -47,14 +47,15 @@ def make_catalog_random_positions_cube(
 
     Parameters
     ----------
-    size : int
-        Number of sources.
+    size : int, optional
+        Number of sources. Default is 100.
     dimension : {1, 2, 3}
-        Number of dimensions.
-    distance_max : `~astropy.units.Quantity`
-        Maximum distance.
+        Number of dimensions. Default is 3.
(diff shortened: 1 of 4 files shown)


Astro-Kirsty:
Leaving the TODO regarding the tolerance for this PR as there were many errors occurring in tests. 
It can be handled at another stage

AtreyeeS (approved):
Thank you @Astro-Kirsty for taking up this clean-up!

Usually, I find an `Examples` section in the doc-string extremely useful to understand how to execute and understand the results of a particular function. If you can add it on the functions users are expected to interact with directly, it would be nice.

About the tolerance here, I think `==` is doing a very stringent check, and `changed` is almost always `True`. When you try with a "high" tolerance like `rtol= 1e-3`,  the `changed=False` option is invoked and the test values change.
You can try with a lower tolerance value, or leave it for a separate PR while adapting the tests as you proposed

Astro-Kirsty:
Thanks @AtreyeeS. 
Another PR can be opened on that, I will be back from holiday in January!

QRemy (approved):
Thanks  @Astro-Kirsty, I agree the TODO about the tests tolerance can be addressed in a separate PR.

Issue #3871, opened 2022-04-04
Can FluxPoints know their own table format


AtreyeeS:
Currently, it is necessary to specify the `format` in `FluxPoints.to_table()`. The default, `gadf-sed` obviously fails for light curves and flux profiles. This is a bit annoying - something generated by the `LightCurveEstimator` should know it its format is `lightcurve`. Of course, this is not critical, but I wonder if it makes sense to have a `FluxPoints.format` which is automatically set by the respective estimator?

adonath:
@AtreyeeS More like `guess_format()`:

- If only an energy axis is present -> `format= "gadf-sed"`
- If a "time" axis is present -> `format= "lightcurve"`
- If a "projected-distance" axis is present -> `format= "profile"`

AtreyeeS:
Fixed in #4992

PR #4992, opened 2023-12-15
Adding a function to guess the format of a FluxPoints object for serialization

cgalelli:
In this PR I introduce a method for `estimators.FluxPoints` which tries to guess the format - "gadf-sed", "lightcurve" or "profile" - of the object itself for serialization.

This change was suggested by @AtreyeeS and @adonath in a previous issue, see #3871

### DIFF
--- a/gammapy/estimators/points/core.py
+++ b/gammapy/estimators/points/core.py
@@ -110,9 +110,7 @@ class FluxPoints(FluxMaps):
     """
 
     @classmethod
-    def read(
-        cls, filename, sed_type=None, format="gadf-sed", reference_model=None, **kwargs
-    ):
+    def read(cls, filename, sed_type=None, format=None, reference_model=None, **kwargs):
         """Read precomputed flux points.
 
         Parameters
@@ -121,8 +119,9 @@ class FluxPoints(FluxMaps):
             Filename.
         sed_type : {"dnde", "flux", "eflux", "e2dnde", "likelihood"}
             SED type.
-        format : {"gadf-sed", "lightcurve"}
-            Format string. Default is "gadf-sed".
+        format : {"gadf-sed", "lightcurve", "profile"}, optional
+            Format string. If None, the format is extracted from the input.
+            Default is None.
         reference_model : `SpectralModel`
             Reference spectral model.
         **kwargs : dict, optional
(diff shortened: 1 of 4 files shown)


gammapy/estimators/points/core.py
```diff
@@ -452,7 +470,7 @@ def to_table(self, sed_type=None, format="gadf-sed", formatted=False):
                 table_flat["x_ref"] = [(x_max + x_min) / 2]
 
                 fp = self.slice_by_idx(slices={"projected-distance": idx})
-                table = fp.to_table(sed_type=sed_type, format="gadf-sed")
+                table = fp.to_table(sed_type=sed_type)
```
AtreyeeS, line 473: Same as above. Here, the format should be `gadf-sed`

gammapy/estimators/points/core.py
```diff
@@ -405,7 +423,7 @@ def to_table(self, sed_type=None, format="gadf-sed", formatted=False):
                 table_flat["time_max"] = [time_max.mjd]
 
                 fp = self.slice_by_idx(slices={"time": idx})
-                table = fp.to_table(sed_type=sed_type, format="gadf-sed")
+                table = fp.to_table(sed_type=sed_type)
```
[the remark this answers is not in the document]
AtreyeeS: The tests pass because once you do a `slice_by_idx` on the time axis, the time axis is removed in the resultant flux point object, and `_guess_format` returns `gadf-sed`. but this is a bit risky and it is better to be explicitly correct

AtreyeeS (approved):
Thank you @cgalelli ! Left two small inline comments.
Otherwise, good to merge

registerrier (commented):
Thanks @cgalelli . This is nice.

I have left one inline comment regarding logging. Otherwise, I agree with @AtreyeeS that it is safer to keep explicit format inside the IO functions.

cgalelli:
Thanks @AtreyeeS and @registerrier for your comments. I adapted the code accordingly. I also added a similar guessing function for input Tables and files.

gammapy/estimators/points/core.py
```diff
@@ -121,8 +119,8 @@ def read(
             Filename.
         sed_type : {"dnde", "flux", "eflux", "e2dnde", "likelihood"}
             SED type.
-        format : {"gadf-sed", "lightcurve"}, optional
-            Format string. Default is "gadf-sed".
+        format : {"gadf-sed", "lightcurve"}
```
AtreyeeS, line 122:
```suggestion
        format : {"gadf-sed", "lightcurve", "profile"}
```

AtreyeeS (approved):
Thank you! Looks good.Please re-solve the merge conflict

QRemy (approved):
Thanks @cgalelli, no comment from my side, merging now.

PR #4993, opened 2023-12-15
Add support for checksum in yaml files

registerrier:
**Description**

This pull request is part of the effort to support fixity metadata in Gammapy products. For FITS files we rely on the CHECKSUM and DATASUM keywords with astropy. We lack a solution for yaml files storing `Models` and `Datasets`.

This PR proposes a mechanism to compute a MD5 hash for the content of the yaml file and include it as the last entry in the yaml dictionary.

**Dear reviewer**

### DIFF
--- a/gammapy/modeling/models/core.py
+++ b/gammapy/modeling/models/core.py
@@ -3,6 +3,7 @@ import collections.abc
 import copy
 import html
 import logging
+import warnings
 from os.path import split
 import numpy as np
 import astropy.units as u
@@ -431,7 +432,12 @@ class DatasetModels(collections.abc.Sequence):
         data = yaml.safe_load(yaml_str)
         checksum_str = data.pop("checksum", None)
         if checksum:
-            verify_checksum(yaml_str, checksum_str)
+            yaml_str = yaml.dump(
+                data, sort_keys=False, indent=4, width=80, default_flow_style=False
+            )
+            if not verify_checksum(yaml_str, checksum_str):
+                warnings.warn("Checksum verification failed.", UserWarning)
+
         # TODO : for now metadata are not kept. Add proper metadata creation.
         data.pop("metadata", None)
         return cls.from_dict(data, path=path)
--- a/gammapy/modeling/models/core.py
+++ b/gammapy/modeling/models/core.py
@@ -12,7 +12,7 @@ import matplotlib.pyplot as plt
 import yaml
 from gammapy.maps import Map, RegionGeom
 from gammapy.modeling import Covariance, Parameter, Parameters
-from gammapy.utils.check import add_checksum
+from gammapy.utils.check import add_checksum, verify_checksum
 from gammapy.utils.metadata import CreatorMetaData
 from gammapy.utils.scripts import make_name, make_path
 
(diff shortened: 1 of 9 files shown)


gammapy/modeling/models/tests/test_io.py
```diff
@@ -124,6 +124,17 @@ def test_sky_models_io(tmpdir, models):
...
 
+@requires_data()
+def test_sky_models_checksum(tmpdir, models):
+    models.write(
+        tmpdir / "tmp.yaml", full_output=True, overwrite_templates=False, checksum=True
+    )
+    file = open(tmpdir / "tmp.yaml", "rb")
+    yaml_content = file.read()
+
+    assert "checksum: " in str(yaml_content)
```
bkhelifi, line 135: Maybe you can add a test on the checksum it-self. IN this way, we are sure that the checksum computation is stable over time.

AtreyeeS (approved):
Thanks @registerrier ! No comments from my side

QRemy (commented):
Thanks @registerrier, the PR is missing few things
- checksum on datasets.write
- verify the checksum on datasets/Models.read such as the yaml str without the checksum entry match the entry if it exits

Astro-Kirsty (approved):
Thanks @registerrier, looks good with your adjustments

gammapy/modeling/models/core.py
```diff
@@ -411,16 +413,32 @@ def names(self):
...
-    def read(cls, filename):
-        """Read from YAML file."""
+    def read(cls, filename, checksum=False):
+        """Read from YAML file.
+
+        Parameters
+        ----------
+        filename : str
+            input filename
+        checksum : bool
```
adonath, line 423: Astropy has a more options for fits: https://docs.astropy.org/en/stable/io/fits/api/verification.html

I think `exception`, `ignore` and `warn` would all make sense here as well.

gammapy/modeling/models/core.py
```diff
@@ -484,6 +503,9 @@ def write(
             Overwrite templates FITS files. Default is False.
         write_covariance : bool, optional
             Whether to save the covariance. Default is True.
+        checksum : bool
```
adonath, line 506: Is it really worth having this options?  It think we should be strict on write and always add a checksum. I can see not harm in adding it...

adonath (commented):
Thanks @registerrier, I have left some comments.

gammapy/modeling/models/core.py
bkhelifi: I agree

gammapy/modeling/models/core.py
registerrier: As far as I understand, the verification scheme you mention concerns the compliance with the FITS standard not the data integrity. 
I have seen only this regarding data integrity: https://docs.astropy.org/en/stable/io/fits/usage/verification.html#verification-using-the-fits-checksum-keyword-convention
The behavior is to raise an  `AstropyUserWarning` in case checksum verification failed.
I have tried to do the same (except we don't have a `GammapyUserWarning`, but we could create one).

adonath: Yes, you are right. Thanks for pointing it out, please ignore the previous comment.

gammapy/modeling/models/core.py
registerrier: Again I have followed the astropy handling here. And now all gammpy methods writing FITS files have a boolean `checksum` argument. See #4837

PR #4994, opened 2023-12-20
Spatial parameters in FovBackgroundModel

katrinstreil:
**Description**

This pull request adds the parameters of the spatial model to the parameters of the FOVBackground model if the spatial model is set. Additionally, if set, the spatial model is added to the string representation.


QRemy (commented):
Thanks @katrinstreil, could you add a test for this ?

QRemy (approved):
The remaining fail is unrelated so it can be merged.

PR #4997, opened 2023-12-21
Fix RootResults default for scipy dev version

QRemy:
Fix RootResults default for scipy dev version

### DIFF
--- a/gammapy/utils/roots.py
+++ b/gammapy/utils/roots.py
@@ -6,7 +6,12 @@ from scipy.optimize import RootResults, root_scalar
 import astropy.units as u
 from gammapy.utils.interpolation import interpolation_scale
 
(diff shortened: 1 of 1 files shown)


gammapy/utils/roots.py
```diff
@@ -6,7 +6,12 @@
 import astropy.units as u
 from gammapy.utils.interpolation import interpolation_scale
 
-BAD_RES = RootResults(root=np.nan, iterations=0, function_calls=0, flag=0)
+try:
```
registerrier, line 9: Why is the try needed? Why does this yield a `TypeError` in some cases?

QRemy: method is a new argument in the dev version, is there a better way to do this ?

registerrier:
@MeeseeksDev backport to v1.0.x

registerrier:
@meeseeksdev backport to v1.1.x

Issue #5005, opened 2024-01-08
Behaviour of gammapy.utils `interpolate_profile`


AtreyeeS:
**Gammapy version**
Latest dev

**Bug description**
There seems to be small interpolation effects near the minima for the ScaledRegularGridInterpolator.

**To Reproduce**
```
from scipy.interpolate import interp1d
import matplotlib.pyplot as plt
from gammapy.utils.interpolation import interpolate_profile

x = [0.2,   0.27594593, 0.38073079, 0.52530556, 0.72477966, 1., 1.37972966, 1.90365394, 2.6265278,  3.62389832, 5.        ]
y = [ 45.19685382,  26.28875168,  13.47343231,   9.56132843,
        18.27203015,  44.63282147,  95.50949058, 180.32676044,
       312.05478467, 508.56544908, 794.50056902]

f1 = interp1d(x, y, kind="quadratic")
f2 = interpolate_profile(x, y)

x1 = np.linspace(0.2, 3, 100)
plt.plot(x1, f1(x1), label="scipy.interpolate.interp_1d")
plt.plot(x1, f2([x1, x1]), label="gammapy.utils.interpolation.interpolate_profile")
plt.legend()
```

shows
![image](https://github.com/gammapy/gammapy/assets/32677370/e47f3587-c2be-4dd2-bab9-03c0a797438d)

It seems that the gammapy interpolate_profile ensures the minima at zero. Is that the desired behaviour or am I using it wrong?
Help @adonath or @QRemy or anyone else please?

adonath:
@AtreyeeS Yes, the `interpolate_profile` assumes the parabola minimum to be at zero. Which should b the case for likelihood profiles by definition. To make the function a bit more generally useful we could allow the profile to be shifted as well, however in you specific case, it might be worth figuring out, why the profile does not go through zero.

AtreyeeS:
>Yes, the interpolate_profile assumes the parabola minimum to be at zero

Where is the assumption? In the `ScaledRegularGridInterpolator`? 

> Which should b the case for likelihood profiles by definition. 

Presence of a constant additive term should not affect the likelihood minimisation, no?

adonath:
> Where is the assumption? In the ScaledRegularGridInterpolator?

Not only.  The assumption is in the square root scaling and the following line: https://github.com/gammapy/gammapy/blob/main/gammapy/utils/interpolation.py#L262

[...]

PR #5006, opened 2024-01-08
Fix interpolate_profile

QRemy:
Resolve issue #5005

### DIFF
--- a/gammapy/utils/interpolation.py
+++ b/gammapy/utils/interpolation.py
@@ -239,7 +239,7 @@ class LinearScale(InterpolationScale):
         return values
 
 
-def interpolate_profile(x, y, interp_scale="sqrt"):
+def interpolate_profile(x, y, interp_scale="sqrt", method="pchip", **kwargs):
     """Helper function to interpolate one-dimensional profiles.
 
     Parameters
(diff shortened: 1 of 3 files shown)


gammapy/utils/interpolation.py
```diff
@@ -253,13 +253,17 @@ def interpolate_profile(x, y, interp_scale="sqrt"):
...
+        Default is “pchip” (better for profile of parabolic shape).
+    **kwargs : dict
+        Keyword arguments passed to `RegularGridInterpolator`.
 
     Returns
     -------
     interp : `ScaledRegularGridInterpolator`
         Interpolator.
     """
-    sign = np.sign(np.gradient(y))
```
QRemy, line 262: Why would we need this line ?

QRemy:
The test fails because the oldest version of scipy tested does not support the default method set to pchip interpolator.

gammapy/utils/interpolation.py
adonath: The `interpolate_profile` does a square root scaling of the values. The profile is assumed to be a parabola and one gets the `abs` function if one does not correct for the sign. With the `abs` function you cannot interpolate linearly across zero. I think @AtreyeeS interpretation is correct, that the `interpolate_profile` assume the profiles goes through zero. So shifting the profile is probably the correct fix. Or just change to scipy quadratic interpolation, as in @AtreyeeS example.

adonath: This is because you have a sampling point at, or very close to the minimum. The sign handling allows to interpolate correctly between samples, even if the minimum is between them. See:

```python3
import numpy as np
from gammapy.utils.interpolation import interpolate_profile, ScaledRegularGridInterpolator

x = np.linspace(-5, 5, 6)
y = x ** 2

x_fine = np.linspace(-5, 5, 100)

interp = interpolate_profile(x, y)
interp_2 = ScaledRegularGridInterpolator(
        points=(x,), values=y, values_scale="sqrt"
    )

plt.plot(x_fine, interp((x_fine,)), color="tab:orange")
plt.plot(x_fine, interp_2((x_fine,)), color="tab:green")
plt.scatter(x, y)
```
Which gives:

image

QRemy: Using cubic spline with lin scale seems ok for that case :
image

adonath:
Thanks @QRemy, no further comments from my side. @AtreyeeS please confirm you can now use `interpolate_profile`...

AtreyeeS (approved):
Thanks! But since this is now just a one line call to `scipy.interpolate.interp1d`, why don't we do away with this function all together and directly use scipy?

registerrier (approved):
Thanks @QRemy . No comments from my side.

Issue #4473, opened 2023-04-18
Add functionality to rebin flux points to have a minimum significance


AtreyeeS:
It is common for users to "rebin the data to have a minimum significance of n sigma for each flux point". This can be done easily by combining the `TS` available on our `FluxPoints`.

It would be convenient to implement  `FluxPoints.rebin(min_sigma=2)`, instead of having people write a variety of code with sometime inconsistent statistics. 

Of course, having other rebinning methods would be required (min counts, min bkg, etc) which would be better implemented at the `EventDataset` level

adonath:
> Of course, having other rebinning methods would be required (min counts, min bkg, etc) which would be better implemented at the EventDataset level 

Just for completeness, we already have: https://docs.gammapy.org/1.0.1/api/gammapy.estimators.utils.resample_energy_edges.html#gammapy.estimators.utils.resample_energy_edges

But I think it is poorly documented. Users can just work with a fine energy binning and get all the functionality they need.

AtreyeeS:
Yes, I understand your concerns, but say, I have a daily binned light curve with mostly upper limits. It can be very convenient to rebin to weekly scales, without having to rerun the estimator. And this is computationally quite light

registerrier:
This is now solved in #5007 . Closing.

PR #5007, opened 2024-01-08
Add functionality to rebin flux points

AtreyeeS:
Implements #4473
Three types of rebinning are supported.

The implementation could probably be cleaner, but I decided to open the PR because this can still be useful, specially for light curves

Needs https://github.com/gammapy/gammapy-data/pull/45 for tests to pass

### DIFF
--- a/examples/tutorials/analysis-time/light_curve_flare.py
+++ b/examples/tutorials/analysis-time/light_curve_flare.py
@@ -78,6 +78,7 @@ log = logging.getLogger(__name__)
 from gammapy.data import DataStore
 from gammapy.datasets import Datasets, SpectrumDataset
 from gammapy.estimators import LightCurveEstimator
+from gammapy.estimators.utils import get_rebinned_axis
 from gammapy.makers import (
     ReflectedRegionsBackgroundMaker,
     SafeMaskMaker,
@@ -272,14 +273,17 @@ datasets.models = sky_model
 # We first create the `~gammapy.estimators.LightCurveEstimator` for the
 # list of datasets we just produced. We give the estimator the name of the
 # source component to be fitted.
-#
+# By default the likelihood scan is computed from 0.2 to 5.0.
+# Here, we increase the max value to 10.0, because we are
+# dealing with a large flare.
 
 lc_maker_1d = LightCurveEstimator(
     energy_edges=[0.7, 20] * u.TeV,
     source="pks2155",
     time_intervals=time_intervals,
-    selection_optional=None,
+    selection_optional="all",
 )
+lc_maker_1d.norm.scan_max = 10
 
 
 ######################################################################
@@ -298,4 +302,20 @@ lc_1d = lc_maker_1d.run(datasets)
 #
 plt.figure(figsize=(8, 6))
 lc_1d.plot(marker="o")
+plt.show()
+
+
+######################################################################
+# Light curves once obtained can be rebinned.
+# Here, we rebin 4 adjacent bins together, to get 40 min bins
+#
+
+axis_new = get_rebinned_axis(lc_1d, method="fixed-bins", value=3, axis_name="time")
+print(axis_new)
+
+lc_new = lc_1d.resample_axis(axis_new)
+plt.figure(figsize=(8, 6))
+ax = lc_1d.plot(label="original")
+lc_new.plot(ax=ax, label="rebinned")
+plt.legend()
 plt.show()
--- a/examples/tutorials/analysis-time/light_curve_flare.py
+++ b/examples/tutorials/analysis-time/light_curve_flare.py
@@ -307,7 +307,7 @@ plt.show()
 
 ######################################################################
 # Light curves once obtained can be rebinned.
-# Here, we rebin 4 adjacent bins together, to get 40 min bins
+# Here, we rebin 4 adjacent bins together, to get 30 min bins
 #
 
 axis_new = get_rebinned_axis(lc_1d, method="fixed-bins", group_size=3, axis_name="time")
--- a/examples/tutorials/analysis-time/light_curve_flare.py
+++ b/examples/tutorials/analysis-time/light_curve_flare.py
@@ -310,7 +310,7 @@ plt.show()
 # Here, we rebin 4 adjacent bins together, to get 40 min bins
 #
 
-axis_new = get_rebinned_axis(lc_1d, method="fixed-bins", value=3, axis_name="time")
+axis_new = get_rebinned_axis(lc_1d, method="fixed-bins", group_size=3, axis_name="time")
 print(axis_new)
 
 lc_new = lc_1d.resample_axis(axis_new)
--- a/gammapy/estimators/points/core.py
+++ b/gammapy/estimators/points/core.py
@@ -59,18 +59,17 @@ def squash_fluxpoints(flux_point, axis):
         ul = stat_profile_ul_scipy(value_scan, stat_scan, delta_ts=delta_ts)
         maps["norm_ul"] = Map.from_geom(geom, data=ul.value)
 
+    maps["stat"] = Map.from_geom(geom, data=f(minimizer.x))
+
     maps["stat_scan"] = Map.from_geom(
         geom=geom.to_cube([MapAxis.from_nodes(value_scan, name="norm")]), data=stat_scan
     )
     try:
         maps["stat_null"] = Map.from_geom(geom, data=np.sum(flux_point.stat_null.data))
+        maps["ts"] = maps["stat_null"] - maps["stat"]
     except AttributeError:
-        pass
+        maps["ts"] = Map.from_geom(geom, data=np.sum(flux_point.ts.data))
 
-    maps["stat"] = Map.from_geom(geom, data=f(minimizer.x))
-
-    # maps["ts"] = maps["stat"] - maps["stat_null"]
-    maps["ts"] = Map.from_geom(geom, data=np.sum(flux_point.ts.data))
     maps["success"] = Map.from_geom(geom=geom, data=minimizer.success, dtype=bool)
 
     combined_fp = FluxPoints.from_maps(
--- a/gammapy/estimators/points/core.py
+++ b/gammapy/estimators/points/core.py
@@ -789,120 +789,25 @@ class FluxPoints(FluxMaps):
         flux_points.meta["n_sigma_ul"] = n_sigma_ul
         return flux_points
 
-    def get_rebinned_axis(self, method, value, axis_name="energy"):
-        """Get the rebinned axis for rebinning
-         the flux point object along the mentioned axis.
+    def resample_axis(self, axis_new):
+        """Rebin the flux point object along the new axis.
         The likelihoods in each bin are combined to compute the
-        resultant maps.
+        resultant quantities.
+        Stat profiles must be present on the fluxpoints object for
+        this method to work.
+
+        For now, works only for flat fluxpoints.
 
         Parameters
         ----------
-        method : {"fixed_bins", "fixed_edges", "min_significance"}
-            The binning method requested. Options are
-            fixed_bins : Combine `value` adjacent bins together
-            fixed_edges : Combine bins within the given edges
-            min_significance : Combine bins to get a minimum significance in each bin
-        value : The value corresponding to the method
-            The relevant matches are
-            fixed_bins : int
-            fixed_edges : MapAxis edges for the relevant axis
-            min_sig : float
-                Minimum TS desired
-        axis_name : The axis name to combine along
-
-        Returns
-        -------
-        flux_points : `~gammapy.estimators.FluxPoints`
-            A new FluxPoints object with modified binning.
-        """
-        # TODO: Make fixed_bins and fixed_edges work for multidimensions
-        if not self.geom.axes.is_unidimensional:
-            raise ValueError(
-                "Rebinning is supported only for Unidimensional FluxPoints \n "
-                "Please use `iter_by_axis` to create Unidimensional FluxPoints"
-            )
-        ax = self.geom.axes[axis_name]
-        nbin = ax.nbin
-
-        if method == "fixed_bins":
-            if not isinstance(value, int):
-                raise ValueError("Only integer number of bins can be combined")
-            idx = np.arange(0, nbin, value)
-            if idx[-1] < nbin:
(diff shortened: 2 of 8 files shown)


gammapy/estimators/points/core.py
```diff
@@ -740,3 +788,127 @@ def recompute_ul(self, n_sigma_ul=2, **kwargs):
...
+            axis_new = TimeMapAxis.from_time_edges(
+                time_min=edges_min + ax.reference_time,
+                time_max=edges_max + ax.reference_time,
+            )
+        else:
+            edges = np.append(edges_min, edges_max[-1])
+            axis_new = MapAxis.from_edges(edges, name=axis_name, interp=ax.interp)
+        return axis_new
+
+    def rebin_on_axis(self, method, value, axis_name="energy"):
```
adonath, line 880: I think for consistency with `Map` and other places (https://github.com/gammapy/gammapy/blob/main/gammapy/estimators/utils.py#L254) we should name the method `resample_axis`. I think it should take a new axis object, with the requested sampling. How users get this new sampling is a different question.

gammapy/estimators/points/core.py
```diff
@@ -740,3 +788,127 @@ def recompute_ul(self, n_sigma_ul=2, **kwargs):
             )
         flux_points.meta["n_sigma_ul"] = n_sigma_ul
         return flux_points
+
+    def get_rebinned_axis(self, method, value, axis_name="energy"):
```
adonath, line 792: I think for now I would male this an independent helper method. We have the same for datasets: https://github.com/gammapy/gammapy/blob/main/gammapy/estimators/utils.py#L254

gammapy/estimators/points/core.py
```diff
@@ -740,3 +788,127 @@ def recompute_ul(self, n_sigma_ul=2, **kwargs):
...
+        # TODO: Make fixed_bins and fixed_edges work for multidimensions
+        if not self.geom.axes.is_unidimensional:
+            raise ValueError(
+                "Rebinning is supported only for Unidimensional FluxPoints \n "
+                "Please use `iter_by_axis` to create Unidimensional FluxPoints"
+            )
+        ax = self.geom.axes[axis_name]
+        nbin = ax.nbin
+
+        if method == "fixed_bins":
```
adonath, line 827: I think you will arrive at a clean implementation if you use a "registry" instead: for each method you move the code to and independent function, such as `get_edges_fixed_bins()`, `get_edges_min_sqrt_ts()` etc. The you access the method like:

```python3

RESAMPLE_METHODS = {
    "fixed-bins": get_edges_fixed_bins,
    "min-sqrt-ts": get_edges_min_sqrt_ts,
    ...
}

func = RESAMPLE_METHODS[method](**kwargs)
```

This is much more extensible if future and users can even add their own methods if needed.

gammapy/estimators/points/core.py
```diff
@@ -740,3 +788,127 @@ def recompute_ul(self, n_sigma_ul=2, **kwargs):
...
+            if isinstance(value, Time):
+                edges_min = value[:-1] - ax.reference_time
+                edges_max = value[1:] - ax.reference_time
+            else:
+                edges_min, edges_max = value[:-1], value[1:]
+
+        else:
+            raise ValueError(
+                "Incorrect option. Choose from "
+                "{fixed_bins, fixed_edges, min_significance}"
```
adonath, line 867: Once you have the `SAMPLE_METHODS` dict I suggest above you can just print the keys as valid options.

adonath (commented):
Thanks a lot @AtreyeeS, resampling flux points "after the fact" is very interesting functionality and convenience for the users. For now I put some suggestions to slightly refactor the code. 

If you feel strongly about the added convenience, I would suggest to introduce both `resample_axis` and `resample_axis_by_method`. Where the former takes a new axis object, while the latter the method type and keyword arguments as you have now.

gammapy/estimators/utils.py
```diff
@@ -461,3 +462,131 @@ def compute_lightcurve_doublingtime(lightcurve, flux_quantity="flux"):
...
+    value : int
+        Number of bins to combine
+
+    Returns:
+    -------
+    edges_min : `Quantity` or `Time`
+    edges_max : `Quantity` or `Time`
+        Low and High edges of the new axis
+    """
+
```
adonath, line 483: Cant't you just use `Axis.downsample()` here?

gammapy/estimators/utils.py
```diff
@@ -461,3 +462,131 @@ def compute_lightcurve_doublingtime(lightcurve, flux_quantity="flux"):
     )
 
     return table
+
+
+def get_edges_fixed_bins(fluxpoint, value, axis_name="energy"):
```
adonath, line 467: Maybe choose a more explicit name such as `group_size` instead of `value`.

gammapy/estimators/utils.py
```diff
@@ -461,3 +462,131 @@ def compute_lightcurve_doublingtime(lightcurve, flux_quantity="flux"):
...
+        raise ValueError("Only integer number of bins can be combined")
+    idx = np.arange(0, nbin, value)
+    if idx[-1] < nbin:
+        idx = np.append(idx, nbin)
+    edges_min = ax.edges_min[idx[:-1]]
+    edges_max = ax.edges_max[idx[1:] - 1]
+    return edges_min, edges_max
+
+
+def get_edges_min_sqrt_ts(fluxpoint, value, axis_name="energy"):
```
adonath, line 496: Again maybe choose `sqrt_ts_threshold` instead of the generic `value`.

gammapy/estimators/utils.py
```diff
@@ -461,3 +462,131 @@ def compute_lightcurve_doublingtime(lightcurve, flux_quantity="flux"):
...
+    edges_max = edges_max[:i1]
+    edges_min = edges_min[:i1]
+
+    return edges_min, edges_max
+
+
+RESAMPLE_METHODS = {
+    "fixed-bins": get_edges_fixed_bins,
+    "min-sqrt-ts": get_edges_min_sqrt_ts,
+}
```
adonath, line 540: I think your initial implementation offered a third method?

gammapy/estimators/utils.py
```diff
@@ -461,3 +462,131 @@ def compute_lightcurve_doublingtime(lightcurve, flux_quantity="flux"):
...
+
+
+def get_rebinned_axis(fluxpoint, method, value, axis_name="energy"):
```
adonath, line 543: Just use `**kwargs` here and forward them to the respective method. I think
```python3
new_axis = get_rebinned_axis(..., method="fixed-bins", group_size=4)
```
Is more readable then:
```python3
new_axis = get_rebinned_axis(..., method="fixed-bins", value=4)
```
If you don't know what the method parameters are.

adonath (commented):
Thanks @AtreyeeS! The new version is much more readable now. I have left a 2nd round of comments, and please check all docstrings once more for consistency and completeness.

cgalelli (approved):
Thanks @AtreyeeS , this is a very useful feature addition. I think the separation of the axis resampling methods from the actual work on the `FluxPoints` is very clever, in this way it is very easy to use this as a base for more complex resamplings!

gammapy/estimators/utils.py
AtreyeeS: Yes, but that is when the user passes fixed bin edges (eg, in the case of recomputing light curves with Bayesian Block edges). In the present scheme, the user can directly create the `MapAxis` for that case add pass it to the resample function

adonath: Makes sense!

gammapy/estimators/utils.py
AtreyeeS: Downsample only works if it's divisible. However, it is a common use case to move from daily bins to weekly bins, where you might have a large number of (not a factor of 7) bins or something

registerrier: Then maybe we could change the `downsample` behavior to have a `strict=False` option which when set to `True` will simply put the remainder in the last bin as you do.

gammapy/estimators/utils.py
```diff
@@ -461,3 +462,127 @@ def compute_lightcurve_doublingtime(lightcurve, flux_quantity="flux"):
...
+
+
+def get_edges_min_sqrt_ts(fluxpoint, sqrt_ts_threshold, axis_name="energy"):
+    """Rebin the flux point to combine adjacent bins
+       till a minimum significance is obtained
+
+    Parameters
+    ----------
+    fluxpoint : `FluxPoints`
+            The flux point to rebin
```
registerrier, line 503:
```suggestion
        The flux point to rebin
```

gammapy/estimators/utils.py
```diff
@@ -461,3 +462,127 @@ def compute_lightcurve_doublingtime(lightcurve, flux_quantity="flux"):
...
+    return edges_min, edges_max
+
+
+def get_edges_min_sqrt_ts(fluxpoint, sqrt_ts_threshold, axis_name="energy"):
+    """Rebin the flux point to combine adjacent bins
+       till a minimum significance is obtained
+
+    Parameters
+    ----------
+    fluxpoint : `FluxPoints`
```
registerrier, line 502:
```suggestion
    fluxpoint : `~gammapy.estimators.FluxPoints`
```

gammapy/estimators/utils.py
```diff
@@ -461,3 +462,127 @@ def compute_lightcurve_doublingtime(lightcurve, flux_quantity="flux"):
...
+            The flux point to rebin
+    sqrt_ts_threshold : float
+        The minimum significance desired
```
registerrier, line 505: Here I am bit annoyed.

Summing `ts` is indeed the correct approach to combine the max likelihoods. But one should keep in mind that the number of do.f. increases when doing so. Using the square root of the summed `ts` is therefore not a significance in the usual sense. 

So you could either define a minimum threshold in `ts` and not bother with the number of `dof` avoiding the reference to significance.
Conversely you can convert the summed `ts` to a p-value accounting for the number of `dof` and transform it back to a significance. One possible issue here is the fact that p-value calculation will quickly yield 0 for large `ts`.

gammapy/estimators/utils.py
```diff
@@ -461,3 +462,127 @@ def compute_lightcurve_doublingtime(lightcurve, flux_quantity="flux"):
...
+
+
+def get_rebinned_axis(fluxpoint, axis_name="energy", **kwargs):
+    """Get the rebinned axis for resampling
+     the flux point object along the mentioned axis.
+
+
+    Parameters
+    ----------
+    kwargs : Dict
```
registerrier, line 550: Why not explictly describe the arguments in the `Parameters`? 
`kwargs` is not optional here since you need to pass the `method`. You should rather list it as a required argument.
I would also list the other arguments as parameters.

registerrier (changes requested):
Thanks @AtreyeeS . This is a really nice addition!

I have one major concern regarding `TS` calculation. 
- You cannot sum the individual ts when summing over several bins. Since you are determining the new best fit `norm` from the sum of stat profiles, the new `ts` must correspond to this new minimum and not to the sum of the `ts` with distinct minima.
- When rebinning with minimum significance, taking the square root of the sum of the ts, is not the significance itself since the number of dof is larger than one.

gammapy/estimators/utils.py
AtreyeeS: Sorry about missing the d.o.f.

Completely agree about the  p-value calculation for large TS. But If we define a minimum threshold in TS, we should also tell the user the number of bins combined. I can't find a good place for that.

adonath: I guess we could add a column with dof? This is interesting in other situations as well. However this should be an independent PR.

examples/tutorials/analysis-time/light_curve_flare.py
```diff
@@ -299,3 +303,19 @@
 plt.figure(figsize=(8, 6))
 lc_1d.plot(marker="o")
 plt.show()
+
+
+######################################################################
+# Light curves once obtained can be rebinned.
+# Here, we rebin 4 adjacent bins together, to get 40 min bins
```
AtreyeeS, line 310:
```suggestion
# Here, we rebin 4 adjacent bins together, to get 30 min bins
```

gammapy/estimators/utils.py
AtreyeeS: Created an issue for this #5047

gammapy/estimators/utils.py
AtreyeeS: Added

gammapy/estimators/points/core.py
```diff
@@ -25,6 +29,59 @@
...
+    maps["stat"] = Map.from_geom(geom, data=f(minimizer.x))
+
+    maps["stat_scan"] = Map.from_geom(
+        geom=geom.to_cube([MapAxis.from_nodes(value_scan, name="norm")]), data=stat_scan
+    )
+    try:
+        maps["stat_null"] = Map.from_geom(geom, data=np.sum(flux_point.stat_null.data))
+        maps["ts"] = maps["stat_null"] - maps["stat"]
+    except AttributeError:
+        maps["ts"] = Map.from_geom(geom, data=np.sum(flux_point.ts.data))
```
AtreyeeS, line 71:
```suggestion
        maps["ts"] = None
```

registerrier (commented):
Thanks @AtreyeeS !

Any idea why the LC object has no `ts` column?

AtreyeeS:
The test fail was due to https://github.com/gammapy/gammapy-data/pull/47 not being merged.

Implemented the latest comments

gammapy/estimators/points/core.py
```diff
@@ -25,6 +28,66 @@
...
+    maps["stat"] = Map.from_geom(geom, data=f(minimizer.x))
+
+    maps["stat_scan"] = Map.from_geom(
+        geom=geom.to_cube([MapAxis.from_nodes(value_scan, name="norm")]), data=stat_scan
+    )
+    try:
+        maps["stat_null"] = Map.from_geom(geom, data=np.sum(flux_point.stat_null.data))
+        maps["ts"] = maps["stat_null"] - maps["stat"]
+    except AttributeError:
+        log.info(
```
registerrier, line 75:
```suggestion
        log.warning(
```

registerrier (approved):
Thanks @AtreyeeS . Looks good!

PR #5013, opened 2024-01-10
Fix FluxMap.slice_by_coord

AtreyeeS:
MapAxis.coord_to_idx returns an array of 1 element, which breaks slicing.
This PR ensures that a number is passed to the slice, instead of an array.


QRemy (approved):
Thanks @AtreyeeS, no comment from my side.

bkhelifi (approved):
Thanks @AtreyeeS for this fix!

registerrier:
Note: backport is not possible nor necessary as this function was introduced after 1.1.

PR #5014, opened 2024-01-10
Fixing None objects in `Parameters.to_table()`

### LINKED ISSUE #5012: Parameters.to_table() does not handle priors correctly
**Gammapy version**
dev
**Bug description**
Creating a table from parameters without priors will produce a `"prior"` column containing `None` with `object` type. This is problematic as e.g. this `Table` cannot be written to disk.

**Expected behavior**
When a prior is set. The expected type is `str`, the content of the table when no prior is set should be `""`.

**To Reproduce**
```
from gammapy.modeling.models import SkyModel
model = SkyModel.create(spectral_model="pl")
model.parameters.to_table().write("test.fits")
```

MRegeard:
This pull request fix #5012. 

As mentioned in the issue, a table can be created with `None` object in it but not written to disk. This means that if you have a column of the table that is filled for the first time with `None`, it's inferred `dtype` will be `Object`. 

To fix that I have created a staticmethod that created a `Table` with predefined column name and associated dtype. Then to fill the table I pass a dictionary defined as `{name: value}` (where name match a column name) to `Table.add_row` . This way, only the field that are in the dictionary are filled. 

In the end it is exactly the same as before, but with stricter type, enabling to write the table to disk.

### DIFF
--- a/gammapy/modeling/tests/test_parameter.py
+++ b/gammapy/modeling/tests/test_parameter.py
@@ -2,6 +2,7 @@
 import pytest
 import numpy as np
 from numpy.testing import assert_allclose
+from astropy.table import Table
 from gammapy.modeling import Parameter, Parameters, PriorParameter, PriorParameters
 
 
@@ -155,7 +156,7 @@ def test_parameters_getitem(pars):
         pars[Parameter("bam!", 99)]
 
 
-def test_parameters_to_table(pars):
+def test_parameters_to_table(pars, tmp_path):
     pars["ham"].error = 1e-10
     pars["spam"]._link_label_io = "test"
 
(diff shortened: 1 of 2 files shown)


QRemy (dismissed):
Thanks @MRegeard , no comment from my side.

gammapy/modeling/tests/test_parameter.py
```diff
@@ -165,6 +172,49 @@ def test_parameters_to_table(pars):
     assert table["link"][0] == "test"
     assert table["link"][1] == ""
 
+    assert table["prior"][0] == ""
+    assert table["prior"][1] == "GaussianPrior"
+    assert table["type"][1] == ""
+
+    table.write(tmp_path / "test_parameters.fits")
+    Table.read(tmp_path / "test_parameters.fits")
```
bkhelifi, line 180: Should we add a least one test on the read table?

MRegeard: Well there is no `from_table` method on `Parameters`. There is one on `DatasetModels` -> `update_parameter_from_table` but them it is implicitly tested in `test_set_parameters_from_table()` of `gammapy/modelling/models/tests/test_core.py`

Issue #4952, opened 2023-12-05
radially asymmetric IRFs


mizorafa:
The current (v1.1) gammapy IRF classes provide functions only for radially symmetric IRFs (in the energy-offset axes) except for Background3D. However, asymmetric response evaluation will be more important and necessary for the analysis of data from large-zenith-angle and/or stereoscopic observations, thus requesting something like `EffectiveAreaTable3D`, `EnergyDispersion3D`, `PSF4D` in addition to `EffectiveAreaTable2D`, `EnergyDispersion2D`, `PSF3D`. Accordingly, `gammapy.makers.utils` needs to be updated, I guess.

registerrier:
Thanks for bringing this up. I think there is a discussion #4725 started by @maxnoe  to which we have not been very responsive so far.  I think we could probably continue the discussion there. Or point to this one there.

> I also strongly disagree on the part of "once it is finalized, it can be implemented" We can't just define a format and then, after a format has been specified, implementation in Gammapy comes next.

I tend to agree here. It is very difficult to define a priori a format without any experience on the object. Most of our formats for DL4 and DL5 products are ad-hoc formats. And they will need adaptation to make them perennial.

> 
> It would be really great if there was some kind of tutorial what would be needed to implement a new IRF component parametrization in Gammapy. This doesn't mean it has to be part of a release / be merged into main for the study to be made, but I think just waiting for an official proposal from CTAO will not be on the timescale that MAGIC and LST need to analyze the existing data and also the other way around: CTAO doesn't have the real observed data to meaningfully test the new implementations.

I agree here as well. 
Creating a new IRF has become quite easy since 0.20. But this feature is indeed barely documented. 
Of course, this is without any support for IO and that's why as @AtreyeeS wrote, we cannot really provide a proper public implementation of more complex IRFs in gammapy for now.
But, indeed, we should expose how to create a new IRF somewhere. There are even use cases beyond asymmetric IRFs.

There are no specific API tutorial on gammapy.irf . We could add one. Recipes could be an option (but they are still broken).
 
> I am happy to give support for the computation of the IRFs side. Somebody else, from the Gammapy team, should be able to advise what is needed on the Gammapy side. On the example datasets we can use to test this, we can certainly use prod5b CTA simulations at 60 degrees zenith or the recent divergent pointing simulation set, which is currently being analysed by @Iburelli.

On the gammapy side, I think we need to adapt various elements:
- document user defined IRF addition 
[...]

PR #5015, opened 2024-01-10
Examples of radially asymmetric IRFs

AtreyeeS:
After the discussions in #4952 I tried to test some asymmetric IRFs implementation

The good news is that we can support asymmetric aeff and edisp with almost minimal effort.
Defining new IRFs is easy, and we can support building DL4 data products with this PR.

The issue of the PSF is more tricky. We need the `PSFKernelMap`.
Also, the present `PSF` classes are hardcoded to `offset` axis, so that will need some adaptation.

If we decide to support these classes internally, we would need a good inheritance scheme to avoid code duplication. All `IRF3D` classes will have some common functionalities like `plot_at_energy` and `to_2d`. Similarly for all `IRF2D` classes.
Also, having `EffectiveAreaIRF` `EnergyDispersionIRF` classes like our present `BackgroundIRF` will be good.

We also need a defined format to understand where to raise warnings about the IRFs. In this PR, I kept the old behaviour and raise them only for the Background classes, because there is indeed no point in raising the same issue for all the maps.

comments @maxnoe @bkhelifi @registerrier @QRemy @adonath @LauraOlivera


examples/tutorials/api/irfs.py
```diff
@@ -0,0 +1,289 @@
...
+# -----------------
+#
+
+
+class EnergyDispersion3D(IRF):
+    tag = "edisp_3d"
+    required_axes = ["energy_true", "migra", "fov_lon", "fov_lat"]
+    default_unit = u.one
+
+    @classmethod
```
registerrier, line 114: you might not expose this in a tutorial as this is quite complex. Showing a simple diagonal one would be enough.

examples/tutorials/api/irfs.py
```diff
@@ -0,0 +1,289 @@
...
+# reduction
+#
+
+import numpy as np
+import scipy.special
+import astropy.units as u
+from astropy.coordinates import SkyCoord
+from gammapy.irf import *
+from gammapy.irf.io import COMMON_IRF_HEADERS, IRF_DL3_HDU_SPECIFICATION
+from gammapy.makers.utils import *
```
registerrier, line 25: not needed here

gammapy/irf/background.py
```diff
@@ -11,7 +11,7 @@
 from .core import IRF
 from .io import gadf_is_pointlike
 
-__all__ = ["Background3D", "Background2D"]
+__all__ = ["Background3D", "Background2D", "BackgroundIRF"]
```
registerrier, line 14: Why do you want to expose `BackgroundIRF`? Because of documentation?

gammapy/makers/utils.py
```diff
@@ -28,6 +28,74 @@
 log = logging.getLogger(__name__)
 
 
+def _get_fov_coords(pointing, irf, geom, use_region_center=True, obstime=None):
```
registerrier, line 31: Is it the right place for this function? It could also live in pointing or irf.

registerrier (commented):
Thanks @AtreyeeS . This is nice!

Just a few rapid comments for a start:
- I think this should be two distinct PRs:
  - the first one should introduce and test support for the different FoV systems (offset or lon,lat) in the `Makers`
  - the second would then introduce the documentation for the IRF
- Regarding the former, would it be worth creating functions such `coord_to_fov_lonlat` and `coord_to_fov_offset` that could take a `FoValignement` as argument in order to have it independent from `IRF`?
- For the tutorial, I think the content could be lighter, e.g. I would not expose a `from_gauss`. Maybe it misses also a general discussion on how IRFs are implemented with some examples of regular IRFs.

examples/tutorials/api/irfs.py
AtreyeeS: Indeed! This is just to show that most of the functions can be easily generalised to the 3D case.

gammapy/irf/background.py
maxnoe: I think presence in `__all__` is required for it to show up in the API documentation.

I think the best way here would be to expand the docstring of the IRF baseclasses to explain to developers what is necessary to implement a new implementation.

AtreyeeS:
Thanks @registerrier ! My main aim here was to show that 

>   * the first one should introduce and test support for the different FoV systems (offset or lon,lat) in the `Makers`
>   * the second would then introduce the documentation for the IRF

If we want to test 3D irfs, it would mean adding these classes in the test files and then testing.
To test the makers, it would be better to have some 3D IRFs to use for testing (@maxnoe do you have some examples?). Otherwise, we can create some from the standard IRFs by rotating around the offset axis.

Or shall we directly add support for 3D IRFs in our main code base? That would need a decision on the inheritance scheme, for ex. And a defined format.

maxnoe:
> To test the makers, it would be better to have some 3D IRFs to use for testing (@maxnoe do you have some examples?). 

Not yet, but I have a new master's student who is now starting implementation of 3d effective area and energy dispersion in pyirf.

bkhelifi:
As the `BackgroundIRF' is now exposed into the documentation, it might be good to improve the docstring, by explaining the content of this base class

gammapy/makers/utils.py
bkhelifi: @AtreyeeS: should we then follow these recommendations ?

examples/tutorials/api/irfs.py
```diff
@@ -0,0 +1,458 @@
...
+        migra_max = coords["migra"][:, 1:, :]
+
+        # Analytical formula for integral of Gaussian
+        s = np.sqrt(2) * sigma
+        t1 = (migra_max - 1 - bias) / s
+        t2 = (migra_min - 1 - bias) / s
+        pdf = (scipy.special.erf(t1) - scipy.special.erf(t2)) / 2
+        pdf = pdf / (migra_max - migra_min)
+
+        r1 = np.rollaxis(pdf, -1, 1)
```
bkhelifi, line 262: As this function is exposed, can you add a simple comment to explain what do these 2 lines?

examples/tutorials/api/irfs.py
```diff
@@ -0,0 +1,458 @@
...
+migra = np.array([0.98, 0.97, 0.7])
+fov_lon = [0.1, 1.5] * u.deg
+fov_lat = [0.0, 0.3] * u.deg
+
+edisp3d.evaluate(
+    energy_true=energy.reshape(-1, 1, 1, 1),
+    migra=migra.reshape(1, -1, 1, 1),
+    fov_lon=fov_lon.reshape(1, 1, -1, 1),
+    fov_lat=fov_lat.reshape(1, 1, 1, -1),
+)[0][2]
```
bkhelifi, line 342: Can you explain the [0][2] to the user?
Can you precise also how the interpolations over the axis and the data are made? Or make a link towards an other tutorial...

examples/tutorials/api/irfs.py
```diff
@@ -0,0 +1,458 @@
...
+# While Gammapy does not have inbuilt classes for supporting asymmetric
+# IRFs (except for `Background3D`), custom classes can be created. For
+# this to work correctly with the `MapDatasetMaker`, only variations
+# with `fov_lon` and `fov_lat` can be allowed.
+#
+# The main idea is that the list of required axes should be correctly
+# mentioned in the class definition.
+#
+# Asymmetric PSFs are not correctly supported at present in the data
+# reduction scheme.
```
bkhelifi, line 126: Maybe you can add that one can however create an asymmetric PSF with the current API, as exemplified here... In this way, the user will not be surprised to see that in the tuto

examples/tutorials/api/irfs.py
```diff
@@ -0,0 +1,458 @@
...
+######################################################################
+# PSF
+# ---
+#
+# A higher dimensional PSF Table can also be created, as shown here.
+# Support for higher dimensional analystic PSF will come later
+#
+
+
+class PSFnD(IRF):
```
bkhelifi, line 393:
```suggestion
class PSF4D(IRF):
```

bkhelifi (changes requested):
Thanks a lot @AtreyeeS ... I have made some suggestions

examples/tutorials/api/irfs.py
```diff
@@ -0,0 +1,458 @@
...
+# The peek methods give quick looks into the irfs
+aeff.peek()
+
+
+######################################################################
+# Similarly, we can access other IRFs as well
+#
+
+psf = EnergyDependentMultiGaussPSF.read(irf_filename, hdu="Point Spread Function")
+bkg = Background3D.read(irf_filename, hdu="BACKGROUND")
```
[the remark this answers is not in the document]
AtreyeeS: The CTA IRFs are in 3D. They are actually symmetric when you plot them, but are supplied as per `Background3D` specifications. If I remember correctly a conversation with Gernot, the IRFs were initially generated in `energy_true` and `offset`, but then supplied as 3D

examples/tutorials/api/irfs.py
AtreyeeS: Removed now

AtreyeeS:
Thanks @bkhelifi for the detailed review. I addressed most of your comments, can you take a look again please?

examples/tutorials/api/irfs.py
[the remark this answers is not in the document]
maxnoe: Public CTAO IRFs should be 2d for Prod3, and pseudo-3d for Prob5.

examples/tutorials/api/irfs.py
```diff
@@ -0,0 +1,495 @@
...
+#
+
+bkg.interp_kwargs
+
+# Evaluate background
+energy = [1, 10, 100, 1000] * u.TeV
+fov_lon = [
+    1,
+    2,
+] * u.deg
```
registerrier, line 98: Maybe try to format it as the line below

examples/tutorials/api/irfs.py
```diff
@@ -0,0 +1,495 @@
...
+fov_lat = [1, 2] * u.deg
+ev = bkg.evaluate(
```
registerrier, line 100: Maybe add a comment explaining that evaluation supports broadcasting numpy-style

As the command might look cryptic

examples/tutorials/api/irfs.py
```diff
@@ -0,0 +1,495 @@
...
+# ~~~~~~~~~~~~~
+#
+
+IRF_DL3_HDU_SPECIFICATION["psf_nd"] = {
+    "extname": "POINT SPREAD FUNCTION",
+    "column_name": "MATRIX",
+    "mandatory_keywords": {
+        **COMMON_IRF_HEADERS,
+        "HDUCLAS2": "PSF",
+        "HDUCLAS3": "FULL-ENCLOSURE",  # added here to have HDUCLASN in order
```
registerrier, line 478:
```suggestion
        "HDUCLAS3": "FULL-ENCLOSURE",
```

examples/tutorials/api/irfs.py
```diff
@@ -0,0 +1,495 @@
...
+# ---
+#
+# Asymmetric PSFs are not correctly supported at present in the data
+# reduction scheme. However, higher dimensional PSF Table can be created
+# as a data container, as shown here.
+#
+
+
+class PSF_assym(IRF):
+    tag = "psf_nd"
```
registerrier, line 428: Maybe use `psf_asym` for clarity?

examples/tutorials/api/irfs.py
```diff
@@ -0,0 +1,495 @@
...
+    fov_lat=fov_lat.reshape(1, 1, 1, -1),
+)
+
+
+######################################################################
+# Serialisation
+# ~~~~~~~~~~~~~
+#
+
+IRF_DL3_HDU_SPECIFICATION["psf_nd"] = {
```
registerrier, line 472: `"psf_asym"`?

gammapy/makers/utils.py
registerrier: For now, I would suggest adding a TODO and open an issue about this.

registerrier (approved):
Thanks @AtreyeeS . I have left a few minor comments, in particular on the tutorial.

gammapy/makers/utils.py
AtreyeeS: Created #5041

registerrier (approved):
Thanks @AtreyeeS . This is ready to merge. No further comment.

registerrier:

PR #5023, opened 2024-01-17
Adapt tolerance and remove others

Astro-Kirsty:
Adjusting tolerances mentioned [here](https://github.com/gammapy/gammapy/pull/4990#pullrequestreview-1786498409). It was not appropriate in the end to adapt the `datasets/evaluator` to include the tolerance. 
It works okay for the `datasets/map` instance.

### DIFF
--- a/gammapy/datasets/map.py
+++ b/gammapy/datasets/map.py
@@ -489,10 +489,9 @@ class MapDataset(Dataset):
 
     def _background_parameters_changed(self):
         values = self.background_model.parameters.value
-        # TODO: possibly allow for a tolerance here?
-        # changed = not np.allclose(self._background_parameters_cached, values, rtol=1e-3)
-
-        changed = ~np.all(self._background_parameters_cached == values)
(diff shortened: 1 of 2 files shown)


gammapy/datasets/map.py
```diff
@@ -489,10 +489,9 @@ def npred_background(self):
 
     def _background_parameters_changed(self):
         values = self.background_model.parameters.value
-        # TODO: possibly allow for a tolerance here?
-        # changed = not np.allclose(self._background_parameters_cached, values, rtol=1e-3)
-
-        changed = ~np.all(self._background_parameters_cached == values)
+        changed = not np.allclose(
+            self._background_parameters_cached, values, rtol=1e-15
+        )
```
QRemy, line 494: I would just remove the TODO and keep the previous behaviour because if the parameters did not change the cached values should be exactly the same.

QRemy (commented):
Thanks @Astro-Kirsty, just one comment :

PR #5024, opened 2024-01-17
Remove potential (0, 0) phase interval by check_intervals method

MRegeard:
The way the `_check_intervals` method was done meant that if you put as an input a phase interval of (*, 1), it would create a supplementary (0, 0) interval. It doesn't break the code or the analysis, but it's useless. I changed that by adding an additional if statement. I also added a test for that. 
I changed the type of created interval to tuple to match the input type.


QRemy (approved):
Thanks @MRegeard, no comment from my side

cgalelli (approved):
Thanks @MRegeard no comments

PR #5027, opened 2024-01-19
Fix copy of TemplateSpatialModel when position is not the center of the map 

QRemy:
Get position from parameters on copy otherwise it is taken from the center of the map


AtreyeeS (approved):
Thanks @QRemy ! No comments

PR #5028, opened 2024-01-19
Fix overwrite on TemplateNpredModel

### LINKED ISSUE #5025: Writing of TemplateNPredModel allows no overwrite
The option `overwrite` is not passed to this line https://github.com/gammapy/gammapy/blob/6321261682b83528858a57baa34fd21b1100b9f4/gammapy/modeling/models/cube.py#L1007  
As a result an error is raised if the map is already written and one cannot overwrite it even if the option in the .write method is set to True
Expected: `self.map.write(self.filename, overwrite=overwrite)`

QRemy:
Fix overwrite on TemplateNpredModel
Resolves #5025.


tunbehaun273 (approved):

AtreyeeS (approved):
Thanks! No comments

Issue #5003, opened 2024-01-05
FluxMaps.slice_by_coord behaviour for different Axis types


AtreyeeS:
The current behaviour of `FluxPoints.slice_by_coord` is rather confusing while slicing along the time axis. 
eg: 

```
from gammapy.estimators import FluxPoints
lc_1d = FluxPoints.read(
    "$GAMMAPY_DATA/estimators/pks2155_hess_lc/pks2155_hess_lc.fits", format="lightcurve"
)
l0 = lc_1d.slice_by_coord({"time":slice(lc_1d.geom.axes["time"].edges[0], lc_1d.geom.axes["time"].edges[1])})
l1 = lc_1d.slice_by_coord({"time":slice(lc_1d.geom.axes["time"].edges[1], lc_1d.geom.axes["time"].edges[2])})
```

`l0.geom` has 0 bins in time and `l0.geom.axes["time"]` raises an Error. `l1` does not.
If the `time` axis is non-contiguous, both `l0` and `l1` lead to an error

This is because `TimeMapAxis.coord_to_pix` assumes an open interval on the first edge, and a closed on the last edge, which is OK in itself, but leads to different behaviour from the `MapAxis`. For example, 

```
start = np.linspace(0,10,5) * u.day
stop = np.linspace(0.1,10.1,5) * u.day
t_ref = Time(55555, format="mjd")
gti = GTI.create(start, stop, reference_time=t_ref)
time_ax = TimeMapAxis.from_gti(gti)
time_ax.coord_to_idx(time_ax.edges_max), time_ax.coord_to_idx(time_ax.edges_min)
```

results `(array([0, 1, 2, 3, 4]), array([-1, -1, -1, -1, -1]))`

while 

```
time_ax2 = TimeMapAxis.from_time_bounds(Time(55555,format='mjd'), Time(55558,format='mjd'),
                                        nbin=5)
time_ax2.coord_to_idx(time_ax2.edges_max), time_ax2.coord_to_idx(time_ax2.edges_min)
```
results `(array([0, 1, 2, 3, 4]), array([-1,  0,  1,  2,  3]))`

(For a `MapAxis`, the behaviour is as expected - no invalid indices
```
energy_axis = MapAxis.from_energy_bounds(
         "1 TeV", "10 TeV", nbin=5, name="energy_true"
     )
energy_axis.coord_to_idx(energy_axis.edges_max), energy_axis.coord_to_idx(energy_axis.edges_min)
```
[...]

PR #5035, opened 2024-01-22
Fixing the different behavior of FluxMaps.slice_by_coord for MapAxis and TimeMapAxis

cgalelli:
As reported in issue #5003, `FluxMaps.slice_by_coord` had different behaviors for `MapAxis` and `TimeMapAxis`, due to the difference in the behavior of the `coord_to_idx` methods between the former and the latter. 

This PR fixes this difference by implementing a `TmeMapAxis.group_table` method and using it  in `FluxMaps.slice_by_coord`, implementing the same logic as `Dataset.slice_by_energy`.


gammapy/maps/axes.py
```diff
@@ -2973,6 +2973,56 @@ def to_header(self, format="gadf", idx=0):
...
+        -------
+        groups : `~astropy.table.Table`
+            Group table.
+        """
+
+        for _, edge in enumerate(edges):
+            if not isinstance(edge, Time):
+                edges[_] = self.reference_time + edges[_]
+
+        time_intervals = [(edges[i], edges[i + 1]) for i in range(len(edges) - 1)]
```
AtreyeeS, line 2994: Here you are assuming contiguous edges. That need not be, no?

gammapy/maps/tests/test_axes.py
```diff
@@ -879,6 +879,26 @@ def test_time_map_axis_format_plot_xaxis(time_intervals):
...
+
+    assert_allclose(groups["idx_min"], [10])
+    assert_allclose(groups["idx_max"], [11])
+    assert_allclose(groups["time_min"], [58932.263158])
+    assert_allclose(groups["time_max"], [58932.83114])
+
+    groups2 = axis.group_table(edges=[11 * u.d, 12 * u.d])
+
+    assert_allclose(groups2["idx_min"], [-1])
+
```
AtreyeeS, line 900: Can you add some more tests

1. Passing times instead of quantity
2. With new edges exactly aligned to the old axis edges
3. New edge not aligned to old axis edges

AtreyeeS (changes requested):
Thank you very much @cgalelli for fixing this! Can you please extend the tests on `FluxMaps.slice_by_coord` to consider both a normal and a time map axis?

gammapy/maps/tests/test_axes.py
cgalelli: Added the tests!

gammapy/maps/axes.py
cgalelli: Whoops, you are totally right! `TimeMapAxis` is interval-based so this should be reflected. Fixed in the most recent commit

cgalelli:
Thanks for the comments and suggestions @AtreyeeS! Wouldn't the normal `MapAxis` behavior for `slice_by_coord` already be tested by the energy slicing test?

gammapy/estimators/map/core.py
```diff
@@ -1120,10 +1120,16 @@ def slice_by_coord(self, slices):
...
+            axis = self.geom.axes[key]
 
-            idx_intervals.append(slice(imin, imax))
+            group = axis.group_table(edges=[interval.start, interval.stop])
+
+            is_normal = group["bin_type"] == "normal   "
+            group = group[is_normal]
+
+            idx_intervals.append(
+                slice(int(group["idx_min"][0]), int(group["idx_max"][0]) + 1)
```
AtreyeeS, line 1131: Does this work correctly for the last bin?

gammapy/estimators/map/tests/test_core.py
```diff
@@ -490,3 +490,6 @@ def test_slice_by_coord():
 
     sliced_map2 = ref_map.slice_by_coord({"energy": slice(0.5 * u.TeV, 5.0 * u.TeV)})
     assert sliced_map2.geom.axes["energy"].nbin == 1
+
+    sliced_map3 = ref_map.slice_by_coord({"energy": slice(3.0 * u.TeV, 11.0 * u.TeV)})
```
AtreyeeS, line 494:
```suggestion
    sliced_map3 = ref_map.slice_by_coord({"energy": slice(1.0 * u.TeV, 10.0 * u.TeV)})
```

AtreyeeS: testing on exact edges might be useful since `sliced_map2` already tests the other case

AtreyeeS: why don't you add a test for the slicing on time in an analogous fashion to the energy?

gammapy/maps/axes.py
```diff
@@ -2973,6 +2973,56 @@ def to_header(self, format="gadf", idx=0):
...
+            mask = mask1 & mask2
+            if np.any(mask):
+                idx_min = np.where(mask)[0][0]
+                idx_max = np.where(mask)[0][-1]
+                time_min = self.time_min[idx_min]
+                time_max = self.time_max[idx_max]
+                bin_type = "normal   "
+            else:
+                idx_min = idx_max = -1
+                time_min = self.time_min[idx_min]
```
AtreyeeS, line 3012: these 2 lines can be outside the if else block to avoid repetition, no?

gammapy/maps/axes.py
```diff
@@ -2973,6 +2973,56 @@ def to_header(self, format="gadf", idx=0):
...
+                time_max = self.time_max[idx_max]
+                if np.any(mask1):
+                    bin_type = "overflow"
+                elif np.any(mask2):
+                    bin_type = "underflow"
+                else:
+                    bin_type = "outflow"
```
AtreyeeS, line 3019: add in the documentation what the different bin types mean.
do you need 3 cases or just outflow suffices?

AtreyeeS (changes requested):
Thanks @cgalelli ! Left some minor comments inline

gammapy/maps/axes.py
cgalelli: In the use case of slicing, just the outflow is enough. Adding under and over might be useful for other cases.

gammapy/maps/axes.py
cgalelli: Modified in the latest commit, thanks!

AtreyeeS (approved):
Thanks a lot @cgalelli ! 
Please add in the docstring what `overflow`, `underflow` and `outflow` mean.

Good to merge from my side

AtreyeeS (approved):
Good from my side

Astro-Kirsty (approved):

gammapy/maps/axes.py
```diff
@@ -2973,6 +2973,59 @@ def to_header(self, format="gadf", idx=0):
 
         return header
 
+    def group_table(self, interval_edges):
```
registerrier, line 2976: Can't you re-use the code in `GTI.group_table`? Is it possible to avoid the code duplication here?

registerrier (commented):
Thanks @cgalelli . This is nice. 
My main comment concerns the `group_table` method for `TimeMapAxis`. It is very similar to `GTI.group_table`. Is there a specific motivation to have two distinct functions?

gammapy/maps/axes.py
cgalelli: The initial idea was to wrap `GTI.group_table`, but unfortunately the method does not handle bin indexes. This would result in either having to use `coord_to_idx`, defeating the point of the change, or writing an equivalent translator code that would probably require more space.

registerrier (commented):
Thanks @cgalelli ! No further comment from my side.

registerrier:
CI fail is unrelated.

PR #5039, opened 2024-01-22
Fix `datasets.info_table()` discrepancy between cumulative mode.

### LINKED ISSUE #5019: datasets.info_table() returns inconsistent values.
**Gammapy version**
Current master / 1.1
Probably its in all version.

**Bug description**
The code for `datasets.info_table()` returns a table with "wrong" values.
For the first dataset, `to_masked` is called before retrieving the info_dict, for all others it is not.

This produces inconsistent output if any masks are attached to the dataset.
Also seems to matter for an attached FoVBackgroundModel, but I don't see why right now.

https://github.com/gammapy/gammapy/blob/0b099a7c68b702ff6e0b5451abd4e5b155530f6f/gammapy/datasets/core.py#L502-L530

**Expected behavior**
Either use `to_masked` everywhere or never.

MRegeard:
This PR fix #5019. 

The discrepancy describe in the issue come from different part of the code: 

- In `info_table` the first dataset was always stacked (alone) whereas the other ones where only stacked if cumulative was `True`. This means that if cumulative was `False`, the first dataset of the table loose its possible model information, therefore `stat_sum = nan`, while it would be some value for the other datasets of the table. 
- In the `_count_statistic` method, `self.background`was passed instead of `self.npred_background()`, therefore the excess, background and sqrt_ts information where not the same if cumulative is `True` or `False`. This is because if cumulative is `True`, the datasets are stack and while stacking `self.background` take the stack of `self.npred_background()` and not `self.background`.

I also added a more robust check for `npred_signal()` so that it is not computed if there are no "signal model".

I still need to add some test for these changes.

### DIFF
--- a/gammapy/datasets/core.py
+++ b/gammapy/datasets/core.py
@@ -505,8 +505,8 @@ class Datasets(collections.abc.MutableSequence):
         Parameters
         ----------
         cumulative : bool
-            Cumulate info across all observations. If True, all the information that depend on a
-            model, such as 'stat_sum', will be lost. Default is False.
+            Cumulate info across all observations. If True, all information depending
+            on a model will be lost. Default is False.
 
         Returns
         -------
--- a/gammapy/datasets/core.py
+++ b/gammapy/datasets/core.py
@@ -505,7 +505,8 @@ class Datasets(collections.abc.MutableSequence):
         Parameters
         ----------
         cumulative : bool
-            Cumulate info across all observations. Default is False.
+            Cumulate info across all observations. If True, all the information that depend on a
+            model, such as 'stat_sum', will be lost. Default is False.
 
         Returns
         -------
@@ -515,19 +516,18 @@ class Datasets(collections.abc.MutableSequence):
         if not self.is_all_same_type:
             raise ValueError("Info table not supported for mixed dataset type.")
 
-        name = "stacked" if cumulative else self[0].name
-        stacked = self[0].to_masked(name=name)
+        rows = []
 
-        rows = [stacked.info_dict()]
-
-        for dataset in self[1:]:
-            if cumulative:
+        if cumulative:
+            name = "stacked"
+            stacked = self[0].to_masked(name=name)
+            rows.append(stacked.info_dict())
+            for dataset in self[1:]:
                 stacked.stack(dataset)
-                row = stacked.info_dict()
-            else:
-                row = dataset.info_dict()
-
-            rows.append(row)
+                rows.append(stacked.info_dict())
+        else:
+            for dataset in self:
+                rows.append(dataset.info_dict())
 
         return Table(rows)
 
--- a/gammapy/datasets/core.py
+++ b/gammapy/datasets/core.py
@@ -525,7 +525,7 @@ class Datasets(collections.abc.MutableSequence):
                 stacked.stack(dataset)
                 row = stacked.info_dict()
             else:
-                row = dataset.info_dict()
+                row = dataset.to_masked(name=dataset.name).info_dict()
 
             rows.append(row)
 
--- a/gammapy/datasets/core.py
+++ b/gammapy/datasets/core.py
@@ -505,8 +505,8 @@ class Datasets(collections.abc.MutableSequence):
         Parameters
         ----------
         cumulative : bool
-            Cumulate information across all datasets. If True, all information depending
-            on a model will be lost. Default is False.
+            Cumulate information across all datasets. If True, all model-dependent
+            information will be lost. Default is False.
 
         Returns
         -------
--- a/gammapy/datasets/core.py
+++ b/gammapy/datasets/core.py
@@ -505,7 +505,7 @@ class Datasets(collections.abc.MutableSequence):
         Parameters
         ----------
         cumulative : bool
-            Cumulate info across all observations. If True, all information depending
+            Cumulate information across all datasets. If True, all information depending
             on a model will be lost. Default is False.
 
         Returns
--- a/gammapy/datasets/core.py
+++ b/gammapy/datasets/core.py
@@ -525,7 +525,7 @@ class Datasets(collections.abc.MutableSequence):
                 stacked.stack(dataset)
                 row = stacked.info_dict()
             else:
-                row = dataset.to_masked(name=dataset.name).info_dict()
+                row = dataset.info_dict()
 
             rows.append(row)
 
--- a/gammapy/datasets/tests/test_datasets.py
+++ b/gammapy/datasets/tests/test_datasets.py
@@ -1,6 +1,7 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 import os
 import pytest
+import numpy as np
 from numpy.testing import assert_allclose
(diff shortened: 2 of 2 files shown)


gammapy/datasets/map.py
```diff
@@ -1572,7 +1572,7 @@ def info_dict(self, in_safe_data_range=True):
 
             if self.background:
                 summed_stat = self._counts_statistic[mask].sum()
```
adonath, line 1574: This already includes the mask, why the separate handling for the background?

gammapy/datasets/map.py
```diff
@@ -1541,7 +1541,7 @@ def from_dict(cls, data, lazy=False, cache=True):
     @property
     def _counts_statistic(self):
         """Counts statistics of the dataset."""
-        return CashCountsStatistic(self.counts, self.background)
+        return CashCountsStatistic(self.counts, self.npred_background())
```
adonath, line 1544: `npred_background` includes the norm and spectrum correction, `background` does not. @registerrier which one did we intend to use for the dataset?

MRegeard: Because if we keep the change above: `npred_background` in `_count_statistic`, then the background field of info_dict become `npred_background`, but we already have a field `npred_background`. 
Does that answer your question @adonath ?

gammapy/datasets/map.py
MRegeard: Because if we keep the change above: `npred_background` in `_count_statistic`, then the background field of info_dict become `npred_background`, but we already have a field `npred_background`. 
Does that answer your question @adonath ?

gammapy/datasets/map.py
bkhelifi: This looks for the me correct behaviour! self.npred_background() contains the adjusted bkg, and self.background only the projected IRF (or a measured OFF from a reflected...).
In the case of a 3D dataset, computing a TS with only the bkg projected IRF is a non-sense: this is the whole spirit of the FoVbackgroundmaker within our data reduction scheme. And if people this using info_table function for physics, in particular for a source detection status, this will lead to wrong result that we do not wish.
I think that this line is an important bug fix within Gammapy!

gammapy/datasets/tests/test_datasets.py
```diff
@@ -88,10 +89,15 @@ def test_datasets_info_table():
     table = datasets_hess.info_table()
     assert table["name"][0] == "23523"
     assert table["name"][1] == "23526"
+    assert np.isnan(table["npred_signal"][0])
 
-    table = datasets_hess.info_table(cumulative=True)
-    assert table["name"][0] == "stacked"
-    assert table["name"][1] == "stacked"
+    table_cumul = datasets_hess.info_table(cumulative=True)
```
bkhelifi, line 94: Maybe one should add numerical tests for the cumulative and non-cumulative case, such that the mentioned issue will never happen (the excess is here, but maybe also the counts and the background as there are the root values to compute the excess).
One should maybe test the mask usage... Be sure that the behaviour is correct without mask and with mask

gammapy/datasets/core.py
```diff
@@ -515,19 +516,18 @@ def info_table(self, cumulative=False):
...
+            stacked = self[0].to_masked(name=name)
+            rows.append(stacked.info_dict())
+            for dataset in self[1:]:
                 stacked.stack(dataset)
-                row = stacked.info_dict()
-            else:
-                row = dataset.info_dict()
-
-            rows.append(row)
+                rows.append(stacked.info_dict())
```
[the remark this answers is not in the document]
MRegeard: I don't think there is any issue here. When cumulative is True, we want each row to contain one more dataset (stacked) than the previous row. So I think this is exactly what you describe.

gammapy/datasets/core.py
```diff
@@ -505,7 +505,8 @@ def info_table(self, cumulative=False):
         Parameters
         ----------
         cumulative : bool
-            Cumulate info across all observations. Default is False.
+            Cumulate info across all observations. If True, all the information that depend on a
+            model, such as 'stat_sum', will be lost. Default is False.
```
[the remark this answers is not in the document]
MRegeard: Now that I think about it, this docstring is not good. Because in the case of OnOff datasets, the stat_sum will be the stat_sum of the stacked dataset (since it's just WStat of the dataset). But in the case where it's not an OnOff datasets (MapDataset, SpectrumDataset), then you effectively loose the stat_sum because the models are not stacked.

bkhelifi (dismissed):
Thanks a lot @MRegeard

gammapy/datasets/core.py
```diff
@@ -505,7 +505,8 @@ def info_table(self, cumulative=False):
         Parameters
         ----------
         cumulative : bool
-            Cumulate info across all observations. Default is False.
+            Cumulate info across all observations. If True, all information depending
+            on a model will be lost. Default is False.
```
Astro-Kirsty, line 509: Change 'info' to 'information'
Also the 2nd sentence needs to be clearer.

Astro-Kirsty (commented):
Thanks @MRegeard!
Just a small comment

gammapy/datasets/map.py
adonath: I see it is for the backwards compatibility.

adonath (commented):
Thanks @MRegeard! I think I'd prefer if you could split out the bug fix for the background handling in a separate PR and then rebase this one.

gammapy/datasets/core.py
```diff
@@ -505,7 +505,8 @@ def info_table(self, cumulative=False):
         Parameters
         ----------
         cumulative : bool
-            Cumulate info across all observations. Default is False.
+            Cumulate information across all datasets. If True, all information depending
+            on a model will be lost. Default is False.
```
Astro-Kirsty, line 509: `If True, all information depending on a model will be lost.`

I'm still not sure what is meant by this, can you clarify?
`If True, all information associated with the model will be lost.` ??

Astro-Kirsty (commented):
Just the one remaining comment, thanks @MRegeard

gammapy/datasets/core.py
MRegeard: It means that because the models are not copied when you stack datasets, informations that depends on model are lost. This is for instance `stat_sum`, or `npred_signal` in a `MapDataset` or `SpectrumDataset`. So this part of the docstring is here to notify the user that it is normal behaviour to get a `NaN` in the `stat_sum` column in these cases. 
If you have a better formulation for this @Astro-Kirsty, I will be please to changed it.

Astro-Kirsty: Okay, your comment makes sense but I would suggest to change the docstring slightly. Even as you described in your comment:
"If True, all information that depends on the model are lost"
OR
"If True, all model-dependent information will be lost."
Or another option, if you have it. But maybe avoid using the wording 'depending'

Astro-Kirsty (approved):

registerrier (approved):
Thanks @MRegeard . This looks good. No further comment.

PR #5045, opened 2024-01-25
Modify config definition in modeling_2D

registerrier:
**Description**

This pull request corrects the docs CI fail due to the way `parameters` arguments of the `SafeMaskMaker` are passed to the `AnalysisConfig`.

Such dicts were never properly validated. But now after the change to pydantic v2. in #4750 Quantities stored in the `parameter` dict will break the export to yaml (via json).

The short term solution is to define them as strings. A longer term solution is to properly validate the full set of arguments of the `Maker`  

**Dear reviewer**


bkhelifi (approved):
Thanks @registerrier

PR #5055, opened 2024-01-29
Add function to scale colorbar height

### LINKED ISSUE #4052: Scaling the colour bar to the plot's height
The colour bars might randomly be larger or smaller than the plot height. This is a known feature of the matplotlib colour bar (see e.g. [link](https://stackoverflow.com/questions/26034777/matplotlib-same-height-for-colorbar-as-for-plot)). In the worse case, the existence of a colour bar can change the size of the plot itself (see e.g. [link](https://stackoverflow.com/questions/57530042/matplotlib-let-color-bar-not-affect-size-and-proportions-of-the-plot)).

Once the plot size is set, the colour bar should match exactly its size. An utility function (`add_scaled_cbar`) should be added in `visualization.utils`, and then all the plot functions of Gammapy should be updated to use it.

**Additional context**
A possible solution can be found [here](https://www.geeksforgeeks.org/set-matplotlib-colorbar-size-to-match-graph/).

Astro-Kirsty:
**Description**
 This PR is to resolve #4052

The [suggestion](https://github.com/gammapy/gammapy/issues/4052#issuecomment-1235445757) by Axel did not work for the `plt.tight_layout()` in the `peek` methods, even with `use_gridspec` set to `True`. So I had to find another possibility.

Instead, I utilised the `make_axes_locatable` functionality in the [mpl_toolkits](https://matplotlib.org/stable/gallery/axes_grid1/demo_colorbar_with_axes_divider.html) of matplotlib.

**Note**: `divider = make_axes_locatable(ax)` makes them the same as the ax
`size` is a percentage of the main axis width to be set as the cbar width
`pad` is similarly determined as the percentage of the main axes width to pad by

### DIFF
--- a/gammapy/irf/background.py
+++ b/gammapy/irf/background.py
@@ -172,7 +172,7 @@ class Background3D(BackgroundIRF):
 
     def plot_at_energy(
         self,
-        energy=None,
+        energy=1 * u.TeV,
         add_cbar=True,
         ncols=3,
         figsize=None,
@@ -184,8 +184,8 @@ class Background3D(BackgroundIRF):
 
         Parameters
         ----------
-        energy : `~astropy.units.Quantity`
-            List of energies.
+        energy : `~astropy.units.Quantity`, optional
+            List of energies. Default is 1 TeV.
         add_cbar : bool, optional
             Add color bar. Default is True.
         ncols : int, optional
@@ -294,14 +294,14 @@ class Background2D(BackgroundIRF):
         )
 
     def plot_at_energy(
-        self, energy=None, add_cbar=True, ncols=3, figsize=None, **kwargs
+        self, energy=1 * u.TeV, add_cbar=True, ncols=3, figsize=None, **kwargs
     ):
         """Plot the background rate in FoV coordinates at a given energy.
 
         Parameters
         ----------
-        energy : `~astropy.units.Quantity`
-            List of energy.
+        energy : `~astropy.units.Quantity`, optional
+            List of energy. Default is 1 TeV.
         add_cbar : bool, optional
             Add color bar. Default is True.
         ncols : int, optional
--- a/gammapy/irf/background.py
+++ b/gammapy/irf/background.py
@@ -319,7 +319,26 @@ class Background2D(BackgroundIRF):
     def plot(
         self, ax=None, add_cbar=True, axes_loc=None, kwargs_colorbar=None, **kwargs
     ):
-        """Plot energy offset dependence of the background model."""
+        """Plot energy offset dependence of the background model.
+
+        Parameters
+        ----------
+        ax : `~matplotlib.axes.Axes`, optional
+            Matplotlib axes. Default is None.
+        add_cbar : bool, optional
+            Add a colorbar to the plot. Default is True.
+        axes_loc : dict, optional
+            Keyword arguments passed to `~mpl_toolkits.axes_grid1.axes_divider.AxesDivider.append_axes`.
+        kwargs_colorbar : dict, optional
+            Keyword arguments passed to `~matplotlib.pyplot.colorbar`.
+        kwargs : dict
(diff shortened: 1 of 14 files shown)


bkhelifi (dismissed):
So amazing!!!! Thanks a lot @Astro-Kirsty

gammapy/visualization/utils.py
```diff
@@ -26,6 +28,16 @@
 }
 
 
+def add_colorbar(img, ax, label=None):
+    """Add colorbar to a given axis."""
+    divider = make_axes_locatable(ax)
+    cax = divider.append_axes(position="right", size="5%", pad="2%")
+    cbar = plt.colorbar(
+        img, cax=cax, orientation="vertical", use_gridspec=True, label=label
```
adonath, line 36: Maybe just forward the keywords `**kwargs` to `plt.colorbar()`?

gammapy/visualization/utils.py
```diff
@@ -26,6 +28,16 @@
 }
 
 
+def add_colorbar(img, ax, label=None):
+    """Add colorbar to a given axis."""
+    divider = make_axes_locatable(ax)
+    cax = divider.append_axes(position="right", size="5%", pad="2%")
```
adonath, line 34: One could introduce an argument `axes_loc={}` with the keywords for this method, such that users can modify it. The defaults can best set using `axes_loc.setdefult()`. But keeping the hard-coded values is probably also fine.

adonath (commented):
Thanks @Astro-Kirsty, I left some minor comments, otherwise look good!

gammapy/irf/effective_area.py
```diff
@@ -149,7 +150,9 @@ def plot_offset_dependence(self, ax=None, energy=None, **kwargs):
         ax.legend(loc="best")
         return ax
 
-    def plot(self, ax=None, add_cbar=True, **kwargs):
+    def plot(
+        self, ax=None, add_cbar=True, axes_loc=None, kwargs_colorbar=None, **kwargs
+    ):
```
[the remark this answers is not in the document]
Astro-Kirsty: Probably wise, I'll implement it

gammapy/irf/background.py
```diff
@@ -300,7 +316,9 @@ def plot_at_energy(
             energy=energy, add_cbar=add_cbar, ncols=ncols, figsize=figsize, **kwargs
         )
 
-    def plot(self, ax=None, add_cbar=True, **kwargs):
+    def plot(
+        self, ax=None, add_cbar=True, axes_loc=None, kwargs_colorbar=None, **kwargs
+    ):
```
bkhelifi, line 321: Same comment about the docstring...

bkhelifi (dismissed):
Thanks a lot @Astro-Kirsty! All good for me

registerrier (changes requested):
Thanks @Astro-Kirsty . 
This PR misses tests. If you explicitly use some matplotlib functionalities, they should be tested. Otherwise we could see failures appearing if the syntax is changed in matplotlib.

Maybe an example usage might be of help in docstrings.

registerrier (dismissed):
Thanks @Astro-Kirsty . No further comment from my side. 

In general, I think the approach of helper function to add configuration to the plots might be a good option. e.g. one could have used only the `add_colorbar` function to add the required functionality instead of adding the `axes_loc` argument to `plot` methods. 

In the long run, a general configuration and/or context managers to do this kind of things might be the way to go.

registerrier:
Can you rebase @Astro-Kirsty ?

Astro-Kirsty:
Thanks @registerrier. 
I have rebased

bkhelifi (dismissed):
Thank you very much!

Astro-Kirsty:
This is not ready for merge, I am solving the error in CI and also the colorbar when we have WCS axes

Astro-Kirsty:
Should be all good now!

registerrier (approved):

PR #4699, opened 2023-07-26
Fermi 2nd and 3rd Pulsar Catalog

MRegeard:
This pull request introduce the 2PC Fermi-LAT catalog to `gammapy.catalog` sub-package. 

The catalog fits file consist of 2 main HDUs:

- One is for the actual pulsar catalog. It gathers intrinsic information on the pulsars, such as the position, the period, derivative of the period, etc.
- The other one is for the spectral parameters of the pulsars. 

Therefore, I had to modify a little bit the core class `SourceCatalog` and `SourceCatalogObject` to account for the spectral parameters to be in another HDU (as with the `extended_source_table`). 

I still need to make the `SourceCatalogObject` class for this catalog.

PR #5056, opened 2024-01-30
Fermi pulsar catalogs base class

MRegeard:
Split of PR #4699. Here is the base class part. 
 
This introduces a base class for object of the pulsar catalog of Fermi (both 2PC and 3PC).

The part where I modify the core class `SourceCatalog` could be extracted into another base class, say `SourceCatalogPC`, but it will add another level of inheritance. I don't have a strong opinion on that.

### DIFF
--- a/gammapy/catalog/core.py
+++ b/gammapy/catalog/core.py
@@ -52,10 +52,11 @@ class SourceCatalogObject:
     _source_name_key = "Source_Name"
     _row_index_key = "_row_index"
 
-    def __init__(self, data, data_extended=None):
+    def __init__(self, data, data_extended=None, data_spectral=None):
         self.data = Bunch(**data)
         if data_extended:
             self.data_extended = Bunch(**data_extended)
+        self.data_spectral = Bunch(**data_spectral) if data_spectral else None
 
     def _repr_html_(self):
         try:
@@ -257,18 +258,44 @@ class SourceCatalog(abc.ABC):
             name_extended = data["Source_Name"].strip()
         else:
             name_extended = None
+
+        if self._source_name_key in data and self._source_name_key == "PSR_Name":
+            name_spectral = data[self._source_name_key].strip()
+        elif self._source_name_key in data and self._source_name_key == "PSRJ":
+            name_spectral = f"PSR{data[self._source_name_key].strip()}"
+        elif "Source_name" in data:
+            name_spectral = data["Source_name"].strip()
+        else:
+            name_spectral = None
+
         try:
             idx = self._lookup_extended_source_idx[name_extended]
             data_extended = table_row_to_dict(self.extended_sources_table[idx])
         except (KeyError, AttributeError):
             data_extended = None
 
-        source = self.source_object_class(data, data_extended)
+        try:
+            idx = self._lookup_spectral_source_idx[name_spectral]
+            data_spectral = table_row_to_dict(self.spectral_table[idx])
+        except (KeyError, AttributeError):
+            data_spectral = None
+
+        source = self.source_object_class(data, data_extended, data_spectral)
         return source
 
     @lazyproperty
+    def _lookup_spectral_source_idx(self):
+        if self._source_name_key == "PSRJ":
+            source_name_key = "NickName"
+        else:
+            source_name_key = self._source_name_key
+        names = [_.strip() for _ in self.spectral_table[source_name_key]]
+        idx = range(len(names))
+        return dict(zip(names, idx))
+
+    @lazyproperty
     def _lookup_extended_source_idx(self):
-        names = [_.strip() for _ in self.extended_sources_table["Source_Name"]]
+        names = [_.strip() for _ in self.extended_sources_table[self._source_name_key]]
         idx = range(len(names))
         return dict(zip(names, idx))
 
@@ -293,6 +320,8 @@ def _skycoord_from_table(table):
         lon, lat, frame = "RA", "DEC", "icrs"
     elif {"ra", "dec"}.issubset(keys):
         lon, lat, frame = "ra", "dec", "icrs"
+    elif {"RAJD", "DECJD"}.issubset(keys):
+        lon, lat, frame = "RAJD", "DECJD", "icrs"
     else:
         raise KeyError("No column GLON / GLAT or RA / DEC or RAJ2000 / DEJ2000 found.")
 
--- a/gammapy/catalog/core.py
+++ b/gammapy/catalog/core.py
@@ -283,12 +283,12 @@ class SourceCatalog(abc.ABC):
             name_spectral = None
         return name_spectral
 
-    def _get_source_name_key(self):
+    def _get_spectral_table_source_name_key(self):
         return self._source_name_key
 
     @lazyproperty
     def _lookup_spectral_source_idx(self):
-        source_name_key = self._get_source_name_key()
+        source_name_key = self._get_spectral_table_source_name_key()
         names = [_.strip() for _ in self.spectral_table[source_name_key]]
         idx = range(len(names))
         return dict(zip(names, idx))
--- a/gammapy/catalog/core.py
+++ b/gammapy/catalog/core.py
@@ -259,14 +259,7 @@ class SourceCatalog(abc.ABC):
         else:
             name_extended = None
 
-        if self._source_name_key in data and self._source_name_key == "PSR_Name":
-            name_spectral = data[self._source_name_key].strip()
-        elif self._source_name_key in data and self._source_name_key == "PSRJ":
-            name_spectral = f"PSR{data[self._source_name_key].strip()}"
-        elif "Source_name" in data:
-            name_spectral = data["Source_name"].strip()
-        else:
-            name_spectral = None
+        name_spectral = self._get_name_spectral(data)
 
         try:
             idx = self._lookup_extended_source_idx[name_extended]
@@ -283,12 +276,19 @@ class SourceCatalog(abc.ABC):
         source = self.source_object_class(data, data_extended, data_spectral)
         return source
 
+    def _get_name_spectral(self, data):
+        if "Source_name" in data:
+            name_spectral = data["Source_name"].strip()
+        else:
+            name_spectral = None
+        return name_spectral
+
+    def _get_source_name_key(self):
+        return self._source_name_key
+
     @lazyproperty
     def _lookup_spectral_source_idx(self):
-        if self._source_name_key == "PSRJ":
-            source_name_key = "NickName"
-        else:
-            source_name_key = self._source_name_key
+        source_name_key = self._get_source_name_key()
         names = [_.strip() for _ in self.spectral_table[source_name_key]]
         idx = range(len(names))
         return dict(zip(names, idx))
--- a/gammapy/catalog/core.py
+++ b/gammapy/catalog/core.py
@@ -51,12 +51,12 @@ class SourceCatalogObject:
     Parameters
     ----------
     data : dict
-        Dictionnary of data from a catalog for a given source.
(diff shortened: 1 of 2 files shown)


gammapy/catalog/core.py
```diff
@@ -257,18 +258,44 @@ def _make_source_object(self, index):
...
+
+        source = self.source_object_class(data, data_extended, data_spectral)
         return source
 
+    @lazyproperty
+    def _lookup_spectral_source_idx(self):
+        if self._source_name_key == "PSRJ":
+            source_name_key = "NickName"
+        else:
+            source_name_key = self._source_name_key
```
QRemy, line 291: maybe move this to a separate private function
`_get_source_name_key()` that return `self._source_name_key` here
 and re-write it in the subclass to return `NickName`
 (and why _source_name_key is different  from  source_name_key in that case ?)

gammapy/catalog/fermi.py
```diff
@@ -41,6 +41,111 @@ def compute_flux_points_ul(quantity, quantity_errp):
...
+        dec = source_pos.dec
+
+        model = PointSpatialModel(lon_0=ra, lat_0=dec, frame="icrs")
+        return model
+
+    def sky_model(self, name=None):
+        """Sky model (`~gammapy.modeling.models.SkyModel`)."""
+        spectral_model = self.spectral_model()
+        if spectral_model is None:
+            return None
```
QRemy, line 124: does catalog.to_models work if one of the models is None ?

gammapy/catalog/core.py
```diff
@@ -257,18 +258,44 @@ def _make_source_object(self, index):
...
             name_extended = None
+
+        if self._source_name_key in data and self._source_name_key == "PSR_Name":
+            name_spectral = data[self._source_name_key].strip()
+        elif self._source_name_key in data and self._source_name_key == "PSRJ":
+            name_spectral = f"PSR{data[self._source_name_key].strip()}"
+        elif "Source_name" in data:
+            name_spectral = data["Source_name"].strip()
+        else:
+            name_spectral = None
```
QRemy, line 269: maybe move to a private function _get_name_spectral that you rewrite in the sub-classes corresponding to different cases

QRemy (commented):
Thanks @MRegeard, maybe some of the code in the base class could be moved in private functions that can be redefined in the sub classes.

gammapy/catalog/core.py
MRegeard: The difference is because in the 3PC, the column that matches the column "PSRJ" of the main table in the spectral table is `Nickname`, while it is the same in the 2PC.

gammapy/catalog/fermi.py
MRegeard: I don't think so, but I re-wrote to-models in the `SourceCatalog` class.

QRemy: in which  case the spectral model is None ?

MRegeard: There are a few pulsars in the 3PC which are in the catalog fits file but don't have associated spectral model.

MRegeard: Same in 2PC.

bkhelifi (commented):
Thanks @MRegeard ... In addition to my small comments, there is no dedicated test to these new functionalities?

gammapy/catalog/core.py
```diff
@@ -257,18 +258,44 @@ def _make_source_object(self, index):
...
         return source
 
+    def _get_name_spectral(self, data):
+        if "Source_name" in data:
+            name_spectral = data["Source_name"].strip()
+        else:
+            name_spectral = None
+        return name_spectral
+
+    def _get_source_name_key(self):
```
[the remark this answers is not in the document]
MRegeard: This function will be overrode by the child class if needed. In the 3PC, we want `"PSRJ"` as the column name for the default catalog table and `"Nickname"` as the column name in the spectral table. So this function is here as a default behaviour (used in 2PC) and will be override by SourceCatalog3PC as I described.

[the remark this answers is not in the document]
MRegeard: I see what is the issue with this method though. I renamed to `_get_spectral_table_source_name_key` it is longer but it explain that this only concern the spectral table (of pulsar catalogs).

Astro-Kirsty (commented):
Thanks for the effort @MRegeard. 
Do some tests need to be added for this?

Also a correction throughout: dictionnary --> dictionary

gammapy/catalog/core.py
```diff
@@ -47,15 +47,27 @@ class SourceCatalogObject:
     it doesn't hold a reference back to it, except for a key
     ``_row_index`` of type ``int`` that links to the catalog table
     row the source information comes from.
+
+    Parameters
+    ----------
+    data : dict
+        Dictionnary of data from a catalog for a given source.
```
registerrier, line 54:
```suggestion
        Dictionary of data from a catalog for a given source.
```

gammapy/catalog/core.py
```diff
@@ -47,15 +47,27 @@ class SourceCatalogObject:
...
+    data_extended : dict
+        Dictionnary of data from a catalog for a given source in the case where the
```
registerrier, line 56:
```suggestion
        Dictionary of data from a catalog for a given source in the case where the
```

gammapy/catalog/core.py
```diff
@@ -47,15 +47,27 @@ class SourceCatalogObject:
...
+        catalog contains an extended sources table (Fermi-LAT).
+    data_spectral : dict
+        Dictionnary of data from a catalof for a given source in the case where the
```
registerrier, line 59:
```suggestion
        Dictionary of data from a catalog for a given source in the case where the
```

gammapy/catalog/fermi.py
```diff
@@ -41,6 +41,111 @@ def compute_flux_points_ul(quantity, quantity_errp):
...
+            spatial_model=self.spatial_model(),
+            spectral_model=spectral_model,
+            name=name,
+        )
+
+    @property
+    def flux_points(self):
+        """Flux points (`~gammapy.estimators.FluxPoints`)."""
+        flux_points_table = self.flux_points_table
+        if flux_points_table is None:
```
registerrier, line 139: You could keep the `self.flux_points_table`

gammapy/catalog/fermi.py
```diff
@@ -41,6 +41,111 @@ def compute_flux_points_ul(quantity, quantity_errp):
...
+            return ss
+        lines = format_flux_points_table(self.flux_points_table).pformat(
+            max_width=-1, max_lines=-1
+        )
+        ss += "\n".join(lines)
+        ss += "\n"
+        return ss
+
+    def _info_phasogram(self):
+        return ""
```
registerrier, line 110: Why does this serve?

gammapy/catalog/fermi.py
```diff
@@ -41,6 +41,111 @@ def compute_flux_points_ul(quantity, quantity_errp):
...
+        ss += "{:<20s} : {:.3f}\n".format("GLAT", source_pos.galactic.b)
+        return ss
+
+    def _info_spectral_fit(self):
+        return "\n"
+
+    def _info_spectral_points(self):
+        ss = "\n*** Spectral points ***\n\n"
+        flux_points_table = self.flux_points_table
+        if flux_points_table is None:
```
registerrier, line 99:
```suggestion
        if self.flux_points_table is None:
```

gammapy/catalog/fermi.py
```diff
@@ -41,6 +41,111 @@ def compute_flux_points_ul(quantity, quantity_errp):
...
+        return ss
+
+    def _info_basic(self):
+        ss = "\n*** Basic info ***\n\n"
+        ss += "Catalog row index (zero-based) : {}\n".format(self.row_index)
+        ss += "{:<20s} : {}\n".format("Source name", self.name)
+        return ss
+
+    def _info_more(self):
+        return ""
```
registerrier, line 79: What for?

gammapy/catalog/core.py
```diff
@@ -257,18 +269,50 @@ def _make_source_object(self, index):
...
+        """
+        source_name_key = self._get_spectral_table_source_name_key()
+        names = [_.strip() for _ in self.spectral_table[source_name_key]]
+        idx = range(len(names))
+        return dict(zip(names, idx))
+
     @lazyproperty
     def _lookup_extended_source_idx(self):
-        names = [_.strip() for _ in self.extended_sources_table["Source_Name"]]
+        """Return a dictionnary of names-idx pairs corresponding to the
```
registerrier, line 312:
```suggestion
        """Return a dictionary of names-idx pairs corresponding to the
```

gammapy/catalog/core.py
```diff
@@ -257,18 +269,50 @@ def _make_source_object(self, index):
...
+        else:
+            name_spectral = None
+        return name_spectral
+
+    def _get_spectral_table_source_name_key(self):
+        return self._source_name_key
+
+    @lazyproperty
+    def _lookup_spectral_source_idx(self):
+        """Return a dictionnary of names-idx pairs corresponding to the
```
registerrier, line 302:
```suggestion
        """Return a dictionary of names-idx pairs corresponding to the
```

gammapy/catalog/core.py
```diff
@@ -257,18 +269,50 @@ def _make_source_object(self, index):
...
 
+    def _get_name_spectral(self, data):
+        if "Source_name" in data:
+            name_spectral = data["Source_name"].strip()
+        else:
+            name_spectral = None
+        return name_spectral
+
+    def _get_spectral_table_source_name_key(self):
+        return self._source_name_key
```
registerrier, line 298: This does not seem necessary. The result is longer than the direct call.
If this is there for daughter classes, maybe it could be a simple attribute rather than a function?

gammapy/catalog/core.py
```diff
@@ -257,18 +269,50 @@ def _make_source_object(self, index):
...
+
+    @lazyproperty
+    def _lookup_spectral_source_idx(self):
+        """Return a dictionnary of names-idx pairs corresponding to the
+        entry of the spectral table (for Fermi-LAT pulsar catalogs).
+        """
+        source_name_key = self._get_spectral_table_source_name_key()
+        names = [_.strip() for _ in self.spectral_table[source_name_key]]
+        idx = range(len(names))
+        return dict(zip(names, idx))
```
registerrier, line 308: The code is very similar between `_lookup_spectral_source_idx` and `_lookup_extended_source_idx`. It could be turned a single function taking either `self.spectral_table[source_name_key]` or `self.extended_sources_table[self._source_name_key]`as argument

registerrier (commented):
Thanks @MRegeard . This is nice. I have left a number of comments inline

gammapy/catalog/fermi.py
MRegeard: If you look at the 2PC and 3PC PR, I have put information like the number of peak, the phase of the peaks, the peak separation etc.

gammapy/catalog/fermi.py
MRegeard: Here I just followed the same scheme as the FermiBaseClass that we have.

gammapy/catalog/core.py
MRegeard: I honestly cannot remember why I did that.

MRegeard:
> Thanks for the effort @MRegeard. Do some tests need to be added for this?
> 
> Also a correction throughout: dictionnary --> dictionary

@Astro-Kirsty, I don't think test are needed because the base class is never other than by the daughter classes. So the tests happen in the daughter class. That is also the scheme for the other Fermi base class.

registerrier (approved):

Issue #5042, opened 2024-01-24
Modify EventList serialization


registerrier:
Currently one can serialize an `EventList` in two ways. Namely:
- `EventList.write()` which will call into `EventList.to_table_hdu`.
- `Observation.write()` with `include_irfs` set to `False`. 
This is problematic for various reasons:
1. `EventList.write` relies on the `EventList.table.meta` to export the various metadata and keywords. We cannot guarantee they follow the gadf standard. It should change to `EventList.meta` instead.
2. `EventList.meta` should most likely not contain all meta informations that belong to `ObservationMetaData`. `ObsInfoMetaData` belongs to `Observation`, not `EventList`. Duplicating seems a bad idea. Yet without these informations, it is not possible to serialize the `EventList` to gadf. 
3. Currently `MapDatasetEventSampler` relies on a specific function `event_list_meta` to fill the `EventList.table.meta`. This should not be necessary.

In the long run, we want to have a clean separation of data format and data model. IO should be performed by a specific `GADFWriter` object to avoid spreading format specification and code all over the code.

I propose we deprecate `EventList.write` and instead rely on `Observation.write` to serialize event lists. Most of the metadata export should be performed there. Then we have a cleaner scheme to implement a proper GADF exporter.

Does it sound reasonable @fabiopintore, @adonath, @LauraOlivera @AtreyeeS @Astro-Kirsty @bkhelifi ?

AtreyeeS:
I guess there is indeed no other reasonable solution that going through `Observation.write`, but if I remember correctly, `Event_list.write` was introduced because it was the only gammapy object that could not write itself. Users were still serialising it, but missing some keywords which were creating issues, which is why we thought it would be okay to add this method.

We will have to document this really well to prevent the same in future...

fabiopintore:
Thanks @registerrier . The approach you're suggesting sounds good and clean. I agree to deprecate the `MapDatasetEventSampler.event_list_meta` as it should be often updated, can be in conflict with other metadata and it's a not crucial code duplication. So +1 to just rely on `Observation.write`.

PR #5059, opened 2024-01-30
Deprecate `EventList.write()` Part 2/2

Astro-Kirsty:
This is connected to #5042

This is Part 2 of adapting the `EventList` and its metadata. 
Part 1: #4832 needs to be merged before this one.

**List of what needs adapting: after deprecation**
- Needs to be well documented so that users understand to utilise `Observations.write()`
- test_write will need to removed (or adapted for include_irfs=False)

### DIFF
--- a/docs/user-guide/dl3.rst
+++ b/docs/user-guide/dl3.rst
@@ -138,12 +138,11 @@ Writing event lists and GTIs to file
 ------------------------------------
 
 To write the events or GTIs separately, one can just save the underlying
-`astropy.table.Table`. There is also a ``write`` method available for
-`~gammapy.data.Observation` which will write the `~gammapy.data.EventList`
-if ``include_irfs`` is set to ``False``.
-It is usually best to save the events and their associated GTIs together in the
-same FITS file. This can be done using the `~gammapy.data.Observation.write`
-method:
+`astropy.table.Table`. To have an event file written in a correct DL3 format, it is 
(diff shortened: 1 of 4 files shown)


Astro-Kirsty:
I've adapted the tests, but they should potentially be moved to the `Observation` test file instead

registerrier (dismissed):

docs/user-guide/dl3.rst
```diff
@@ -138,25 +138,28 @@ Writing event lists and GTIs to file
...
 To write the events or GTIs separately, one can just save the underlying
-`astropy.table.Table`. However, it is usually best to save the events and
-their associated GTIs together in the same FITS file. This can be done using
-the `~gammapy.data.EventList.write` method:
+`astropy.table.Table`. There is also a ``write`` method available for
+`~gammapy.data.Observation` which will write the `~gammapy.data.EventList`
+if ``include_irfs`` is set to ``False``.
+It is usually best to save the events and their associated GTIs together in the
+same FITS file. This can be done using the `~gammapy.data.Observation.write`
+method:
```
registerrier, line 146:
```suggestion
`astropy.table.Table`. To have an event file written in a correct DL3 format, it is 
necessary to utilise the  ``write`` method available for`~gammapy.data.Observation`.
It will write the `~gammapy.data.EventList` and their associated GTIs together in the
same FITS file according to the format specifications. To avoid writing IRFs along the 
``EventList`` one has to set ``include_irfs`` to ``False``. See the example below:
```

PR #5060, opened 2024-01-30
TODO: wcs bug in `WcsGeom.from_header`

MRegeard:
This PR deals with the TODO in `WcsGeom.from_header`. This todo was there because of a bug on WCS where `WCS.dropaxis` would not properly drop the axis. It is now corrected. 
However, I decided to go with `sub(2)` instead of `dropaxis(2)` to deal with the eventuality of more than 3 axes. `dropaxis(2)` would only drop the 3rd axis while `sub(2)` only take the first 2 axis. 
Do we want a specific test for that ?

### DIFF
--- a/gammapy/maps/wcs/geom.py
+++ b/gammapy/maps/wcs/geom.py
@@ -491,9 +491,7 @@ class WcsGeom(Geom):
         wcs : `~WcsGeom`
             WCS geometry object.
         """
-        wcs = WCS(header, naxis=2)
-        # TODO: see https://github.com/astropy/astropy/issues/9259
-        wcs._naxis = wcs._naxis[:2]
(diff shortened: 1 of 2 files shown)


adonath (commented):
Thanks @MRegeard! Yes maybe having a specific unit test for this would be good. Or alternatively, temporarily remove the `.sub()` and see whether any existing test fails.

MRegeard:
None of the tests were failing so I created a new one.

Astro-Kirsty (approved):
Thanks @MRegeard 
It's fine from my side.

registerrier (approved):
Thanks @MRegeard . Looks good. No comment.

registerrier:
Should the astropy  issue mentioned be closed?

Issue #5053, opened 2024-01-29
Uncorrect format  in energy dependent `LightCurveTemplateTemporalModel`


fabiopintore:
The `LightCurveTemplateTemporalModel.to_dict` allows to chose the format (i.e. `table` or `map`, with the former as default) of the model. 
However, the `SkyModel.to_dict()`, which inherits `LightCurveTemplateTemporalModel.to_dict`, only allows for the default format (`table`). This is not suitable in case of energy-dependent `LightCurveTemplateTemporalModel`, where the `map` format should be selected.

A possibility could be to modify `temporal_model.to_dict` in order to remove the `format` argument and let the function to select the proper format.

To replicate the issue:

```
pointing_position = SkyCoord("100 deg", "30 deg", frame="icrs")
spatial_model = PointSpatialModel.from_position(pointing_position)
spectral_model = ConstantSpectralModel(const="1 cm-2 s-1 TeV-1")
temporal_model = LightCurveTemplateTemporalModel.read("$GAMMAPY_DATA/gravitational_waves/GW_example_DC_map_file.fits.gz", format="map")

model = SkyModel(
    spatial_model=spatial_model,
    spectral_model=spectral_model,
    temporal_model=temporal_model,
    name="test",
)

model.to_dict()
```

AtreyeeS:
So the issue here is that the `SkyModel.to_dict` does not propagate options to the individual `to_dict` functions it calls?

For this specific case, I suppose it is reasonable to add a `LightCurveTemplateTemporalModel.guess_format` which return `map` if an energy column is present, and the other functions use the same if `format` is unspecified

registerrier:
This is solved in #5062. Closing.

PR #5062, opened 2024-01-30
Propagate temporal model format

fabiopintore:
This PR fixes the issue #5053.
It allows to correctly propagate the format (i.e. `map` or `table`) of the `LightCurveTemplateTemporalModel`.

### DIFF
--- a/gammapy/modeling/models/tests/test_temporal.py
+++ b/gammapy/modeling/models/tests/test_temporal.py
@@ -503,3 +503,16 @@ def test_model_scale():
         model = GaussianTemporalModel(
             t_ref=50003.2503033 * u.d, sigma="2.43 day", scale="ms"
         )
+
+
+def test_templatetemporalmodelformat():
+    temporal_model = LightCurveTemplateTemporalModel.read(
+        "$GAMMAPY_DATA/gravitational_waves/GW_example_DC_map_file.fits.gz", format="map"
+    )
+    mod_dict = temporal_model.to_dict()
+    assert mod_dict["temporal"]["format"] == "map"
+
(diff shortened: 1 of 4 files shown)


AtreyeeS (commented):
Thanks @fabiopintore !

I am wondering if automatically assigning the format should happen in `SkyModel`, or `LightCurveTemplateTemporalModel` can have a `_guess_format` similar to what was implemented in #4992

fabiopintore:
Thanks @AtreyeeS . The `guess_format` function should be only available for the `LightcurveTemplateModel`, right? This would imply to remove the `format` argument in `LightcurveTemplateModel.to_dict` and let the function to find automatically the format. Is this your suggestion?

AtreyeeS (approved):
Thanks @fabiopintore ! No further comments from my side

gammapy/modeling/models/tests/test_temporal.py
```diff
@@ -503,3 +503,16 @@ def test_model_scale():
         model = GaussianTemporalModel(
             t_ref=50003.2503033 * u.d, sigma="2.43 day", scale="ms"
         )
+
+
+def test_templatetemporalmodelformat():
```
registerrier, line 508: This requires `@requires_data()`.

registerrier (approved):
Thanks @fabiopintore . See inline the comment about CI

gammapy/modeling/models/tests/test_temporal.py
QRemy, line 508:
```suggestion
@requires_data()
def test_template_temporal_model_format():
```

PR #5063, opened 2024-01-31
Adapt class validation in metadata

Astro-Kirsty:
This PR is to adapt the validation of the `from_header` function in metadata. 

Currently, if a header does not conform to the correct type there will be a validation error. 
This fixes this, by trying to proceed with the normal type, however, if that raises a ValidationError allowing the function to pass without validation. 

Previously this was failing with the `test_dataset_hawc` in `gammapy/makers/tests/test_map.py`, as the date from the CreatorMetaData was not of the correct type. This passes now, and I have also added a test

This will need to be merged before #4832


registerrier (dismissed):
Thanks @Astro-Kirsty . This looks good to me.

PR #5067, opened 2024-02-01
Fix tutorial `plt.show()` not showing figures

Astro-Kirsty:
If the `plt.show()` functionality is used in the same cell more than once, then it is not rendered correctly in the documentation.

This PR solves this by creating separate cells for each one of those and a subplot for the other. 

See https://github.com/sphinx-gallery/sphinx-gallery/issues/1025, sphinx-gallery do not plan on changing this functionality


AtreyeeS (approved):

registerrier (approved):

Issue #5075, opened 2024-02-02
PointingInfoMetaData serialization is failing


registerrier:
**Gammapy version**
dev

**Bug description**
`None` is a possible value for `PointingInfoMetaData` attributes. If it is the case, the serialization scheme will fail because no `SkyCoord` is present.

**Expected behavior**
If one of the attribute is `None`, no header keywords should be added.

**To Reproduce**
```

from gammapy.utils.metadata import PointingInfoMetaData
pointing = PointingInfoMetaData()
pointing.to_header()
```
**Other information**
Any other information you think will be useful for us to fix the issue can go here.

PR #5076, opened 2024-02-02
Check if metadata attribute is not None before serialization

registerrier:
**Description**

This pull request solves issue #5075 . 
Before serializing a specific metadata attribute to a set of header keywords, the `to_header()` method now checks that it is not `None`.

**Dear reviewer**


bkhelifi (dismissed):
Thanks a lot @registerrier

PR #5083, opened 2024-02-07
Add TimeInfoMetaData

registerrier:
**Description**

This pull request fills a missing part in the `MetaData` scheme. So far, time range informations were stored directly in `ObservationMetaData` without a proper serialization scheme. 
Here we introduce a specific object to store observation time information with its associated `to_header` and `from_header`.

**Dear reviewer**

### DIFF
--- a/gammapy/data/observations.py
+++ b/gammapy/data/observations.py
@@ -17,7 +17,7 @@ from astropy.utils import lazyproperty
 import matplotlib.pyplot as plt
 from gammapy.utils.deprecation import GammapyDeprecationWarning, deprecated
 from gammapy.utils.fits import LazyFitsData, earth_location_to_dict
-from gammapy.utils.metadata import CreatorMetaData, TargetMetaData
+from gammapy.utils.metadata import CreatorMetaData, TargetMetaData, TimeInfoMetaData
 from gammapy.utils.scripts import make_path
 from gammapy.utils.testing import Checker
 from gammapy.utils.time import time_ref_to_dict, time_relative_to_ref
@@ -263,12 +263,16 @@ class Observation:
             location=location,
         )
 
-        meta = ObservationMetaData(
-            deadtime_fraction=deadtime_fraction,
-            location=location,
+        time_info = TimeInfoMetaData(
             time_start=gti.time_start[0],
             time_stop=gti.time_stop[-1],
             reference_time=reference_time,
+        )
+
+        meta = ObservationMetaData(
+            deadtime_fraction=deadtime_fraction,
+            location=location,
+            time_info=time_info,
             creation=CreatorMetaData(),
             target=TargetMetaData(),
         )
--- a/gammapy/utils/metadata.py
+++ b/gammapy/utils/metadata.py
@@ -2,11 +2,17 @@
 """Metadata base container for Gammapy."""
 import json
 from typing import ClassVar, Literal, Optional, get_args
+import astropy.units as u
 from astropy.coordinates import SkyCoord
 from astropy.time import Time
 import yaml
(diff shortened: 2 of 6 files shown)


gammapy/data/observations.py
```diff
@@ -263,12 +263,16 @@ def create(
             location=location,
         )
 
-        meta = ObservationMetaData(
-            deadtime_fraction=deadtime_fraction,
-            location=location,
+        time_info = TimeInfoMetaData(
```
bkhelifi, line 266: Should we not check here that time_stop>time_start, per safety for the whole pipeline?

gammapy/utils/metadata.py
```diff
@@ -230,6 +237,53 @@ class PointingInfoMetaData(MetaData):
...
+    Parameters
+    ----------
+    time_start : `~astropy.time.Time` or str
+        The observation start time.
+    time_stop : `~astropy.time.Time` or str
+        The observation stop time.
+    reference_time : `~astropy.time.Time` or str
+        The observation reference time."""
+
+    _tag: ClassVar[Literal["pointing"]] = "time_info"
```
bkhelifi, line 252: The tag is really in the 'pointing' section? Or this is just a copy/paste typo?

gammapy/utils/metadata.py
```diff
@@ -230,6 +237,53 @@ class PointingInfoMetaData(MetaData):
...
+    time_start: Optional[TimeType] = None
+    time_stop: Optional[TimeType] = None
+    reference_time: Optional[TimeType] = None
+
+    def to_header(self, format="gadf"):
+        if self.reference_time is None:
+            return {}
+        result = time_ref_to_dict(self.reference_time)
+        if self.time_start is not None:
+            result["TSTART"] = time_relative_to_ref(self.time_start, result).to_value(
```
bkhelifi, line 263: Should we not create a dictionary of keys instead of hiding these keys in the code?

gammapy/utils/metadata.py
registerrier: it is a typo indeed

gammapy/utils/metadata.py
registerrier: The issue here is that the serialization scheme is not one attribute equals one independent set of keywords.
If `reference_time` is exported to `MJDREFFI` `MJDREFF`, `TIMESCALE` etc, `time_start` and `time_stop` need the `reference_time` to be exported to `TSTART` and `TSTOP`.
The scheme with dictionaries does not work here. 
We could add a scheme for additional keywords later though.

bkhelifi: `MJDREFFI`, `MJDREFF`, `TIMESCALE` are FITS standards in any case... So, these ones are not controversial.
TSTART and TSTOP, I do not remember. I have to check... But indeed, one can do that later;)

gammapy/utils/metadata.py
```diff
@@ -230,6 +237,53 @@ class PointingInfoMetaData(MetaData):
...
+    time_stop : `~astropy.time.Time` or str
+        The observation stop time.
+    reference_time : `~astropy.time.Time` or str
+        The observation reference time."""
+
+    _tag: ClassVar[Literal["time_info"]] = "time_info"
+
+    time_start: Optional[TimeType] = None
+    time_stop: Optional[TimeType] = None
+    reference_time: Optional[TimeType] = None
```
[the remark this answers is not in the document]
registerrier: I agree that this is redundant. The reference time is mostly there for serialization.

gammapy/data/observations.py
registerrier: Well maybe. But currently we don't validate that GTIs are correctly ordered. This might be something to add.

PR #5088, opened 2024-02-09
Fix TemplateSpatialModel lon_0 and lat_0 default

QRemy:
Fix TemplateSpatialModel lon_0 and lat_0 default

### DIFF
--- a/gammapy/modeling/models/spatial.py
+++ b/gammapy/modeling/models/spatial.py
@@ -1284,8 +1284,8 @@ class TemplateSpatialModel(SpatialModel):
     """
 
     tag = ["TemplateSpatialModel", "template"]
-    lon_0 = Parameter("lon_0", "0 deg", frozen=True)
-    lat_0 = Parameter("lat_0", "0 deg", min=-90, max=90, frozen=True)
+    lon_0 = Parameter("lon_0", np.nan, unit="deg", frozen=True)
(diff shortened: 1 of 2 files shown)


AtreyeeS (commented):
Thanks @QRemy ! I suppose this should not affect the tests because in practice the default values are the centre of the map?

QRemy:
That affects from_fict when the values are not given in the dict,  I will add a test for that.

bkhelifi:
Can you precise this behaviour in the docstring? I think that this is important for the users

QRemy:
added test and more details in the docstring

gammapy/modeling/models/spatial.py
```diff
@@ -1257,7 +1257,8 @@ class TemplateSpatialModel(SpatialModel):
     """Spatial sky map template model.
 
     For more information see :ref:`template-spatial-model`.
-    The center of the model position can be fit on the data.
+    By default the position of the model is fixed at the center of the map.
+    The position can be fittted unfreezing the `lon_0` and `lat_0 `parameters.
```
AtreyeeS, line 1261:
```suggestion
    The position can be fittted by unfreezing the `lon_0` and `lat_0 `parameters.
```

AtreyeeS (approved):
Thanks for the fix @QRemy !

PR #5089, opened 2024-02-09
Fix another NoOverlapError in spatial model integrate_geom

QRemy:
Fix another NoOverlapError in spatial model integrate_geom.


bkhelifi (approved):
Thanks a lot @QRemy

QRemy:
I confirm that the alignment check fail due to precision error on the position after upsampling/dowmsampling with larger odd value in oversampling factor.
The fix prevents it by simply filling the upsampled/dowmsampled data into the map created previously rather than directly stacking  upsampled/dowmsampled map.

registerrier (approved):
Thanks @QRemy . Looks good!

Issue #5094, opened 2024-02-13
Integrate_geom is causing large memory load


registerrier:
**Gammapy version**
dev and 1.0.2

**Bug description**
`SpatialModel.integrate_geom()` will require very large amount of memory when oversampling is necessary and when the model is outside the required `geom`.

**Expected behavior**
The amount of memory should stay within reasonable values since those large geometries are usually not needed.

**To Reproduce**
```
from astropy.coordinates import SkyCoord
from gammapy.maps import Map, WcsGeom
from gammapy.modeling.models import GaussianSpatialModel
import astropy.units as u

sigma=1e-4*u.deg

enter = SkyCoord("0d", "0d", frame="icrs")
position = SkyCoord(8.234, -0.172, unit="deg", frame="icrs")
model = GaussianSpatialModel(lon_0="0.234 deg", lat_0="-0.172 deg", sigma=sigma, frame="icrs")

model.position = position
geom = WcsGeom.create(skydir=center, npix=100, binsz=0.02)

integrated = model.integrate_geom(geom)
print(1-integrated.data.sum())
```

**Other information**
This is causing issue https://github.com/gammapy/gammapy-benchmarks/issues/146 in validation.

adonath:
Thanks @registerrier! If I understood correctly the problem is the `pass` statement here: https://github.com/gammapy/gammapy/blob/main/gammapy/modeling/models/spatial.py#L236

So if there is an overlap error it will use the full geometry instead, which indeed leads to a large memory usage. I fully agree that the cutout handling should not happen in `integrate_geom`, our models classes should stay "dump".

adonath:
@QRemy Why  is the cutout mechanism triggered at all in the `integrate_geom` method? I though it should be all handled in the `MapEvaluator`?

registerrier:
The mechanism is triggered to avoid oversampling the whole geom which can be very costly.

QRemy:
> @QRemy Why is the cutout mechanism triggered at all in the `integrate_geom` method? I though it should be all handled in the `MapEvaluator`?

The cutout here is only model dependant as it takes into account only the evaluation radius soso model is computed only where it is defined.

The problem is that if the source is totally outside the target geom there will be an overlap error
I guess in that case we can just return an empty map.
Because anyway if a source is outside the dataset geom and contributes inside only due to the psf, we can't evaluate its contribution properly.

adonath:
[...]

PR #5096, opened 2024-02-13
Fix for sources evaluation outside dataset geom

QRemy:
See #5094


registerrier (approved):
Thanks @QRemy . This solves the issue in validation.

I have added a test to check that `integrate_geom` does not fail when the model is outside the geom and that `np.nan` will return a map filled with `np.nan`.

I have also corrected the test for `integrate_geom` to test more with values as well as several models. There are a number of issues. In particular with the disk model. This needs to be checked and corrected in a future PR.

PR #5097, opened 2024-02-14
Make `jfactor` a parameter for DarkMatter models

StFroese:
It is common in dark matter analyses to have a prior on the J-Factor (see [here](https://arxiv.org/pdf/2111.15009.pdf)). 
In order to achieve this, the `jfactor` of the `DarkMatterAnnihilationSpectralModel` has to become a parameter.

I also renamed the J-Factor to D-Factor for the `DarkMatterDecaySpectralModel` and did the same there, this is also very common :)

### DIFF
--- a/examples/tutorials/api/astro_dark_matter.py
+++ b/examples/tutorials/api/astro_dark_matter.py
@@ -219,7 +219,7 @@ channel = "Z"
 massDM = 10 * u.TeV
 diff_flux = DarkMatterAnnihilationSpectralModel(mass=massDM, channel=channel)
 int_flux = (
-    jfact * diff_flux.integral(energy_min=0.1 * u.TeV, energy_max=10 * u.TeV)
+    jfact.to_value("GeV2 cm-5") * diff_flux.integral(energy_min=0.1 * u.TeV, energy_max=10 * u.TeV)
 ).to("cm-2 s-1")
 
 flux_map = WcsNDMap(geom=geom, data=int_flux.value, unit="cm-2 s-1")
@@ -243,7 +243,7 @@ channel = "Z"
 massDM = 10 * u.TeV
 diff_flux = DarkMatterDecaySpectralModel(mass=massDM, channel=channel)
 int_flux = (
-    jfact_decay * diff_flux.integral(energy_min=0.1 * u.TeV, energy_max=10 * u.TeV)
+    jfact_decay.to_value("GeV cm-2") * diff_flux.integral(energy_min=0.1 * u.TeV, energy_max=10 * u.TeV)
 ).to("cm-2 s-1")
 
 flux_map = WcsNDMap(geom=geom, data=int_flux.value, unit="cm-2 s-1")
--- a/examples/tutorials/api/astro_dark_matter.py
+++ b/examples/tutorials/api/astro_dark_matter.py
@@ -219,7 +219,8 @@ channel = "Z"
 massDM = 10 * u.TeV
 diff_flux = DarkMatterAnnihilationSpectralModel(mass=massDM, channel=channel)
 int_flux = (
-    jfact.to_value("GeV2 cm-5") * diff_flux.integral(energy_min=0.1 * u.TeV, energy_max=10 * u.TeV)
+    jfact.to_value("GeV2 cm-5")
+    * diff_flux.integral(energy_min=0.1 * u.TeV, energy_max=10 * u.TeV)
 ).to("cm-2 s-1")
 
 flux_map = WcsNDMap(geom=geom, data=int_flux.value, unit="cm-2 s-1")
@@ -243,7 +244,8 @@ channel = "Z"
 massDM = 10 * u.TeV
 diff_flux = DarkMatterDecaySpectralModel(mass=massDM, channel=channel)
 int_flux = (
-    jfact_decay.to_value("GeV cm-2") * diff_flux.integral(energy_min=0.1 * u.TeV, energy_max=10 * u.TeV)
+    jfact_decay.to_value("GeV cm-2")
+    * diff_flux.integral(energy_min=0.1 * u.TeV, energy_max=10 * u.TeV)
 ).to("cm-2 s-1")
 
 flux_map = WcsNDMap(geom=geom, data=int_flux.value, unit="cm-2 s-1")
--- a/gammapy/astro/darkmatter/spectra.py
+++ b/gammapy/astro/darkmatter/spectra.py
@@ -250,30 +250,6 @@ class DarkMatterAnnihilationSpectralModel(SpectralModel):
         data["spectral"]["k"] = self.k
         return data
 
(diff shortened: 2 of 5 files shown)


StFroese:
Maybe the jfactor should be given as `log10(J)` to avoid large values

StFroese:
> Maybe the jfactor should be given as `log10(J)` to avoid large values

Or a scale factor which is optimized instead. 
$\langle \sigma \nu \rangle$ is also maximally degenerate to the J-Factor, would make to linear scale factors to be optimised. 
Any opinions on how to do this nicely?

examples/tutorials/api/astro_dark_matter.py
```diff
@@ -244,7 +245,8 @@
 massDM = 10 * u.TeV
 diff_flux = DarkMatterDecaySpectralModel(mass=massDM, channel=channel)
 int_flux = (
-    jfact_decay * diff_flux.integral(energy_min=0.1 * u.TeV, energy_max=10 * u.TeV)
+    jfact_decay.to_value("GeV cm-2")
```
alessandromontanari, line 248: Wouldn't it be better to directly rewrite the definition of `jfact_decay` to `dfact` in the tutorial notebook? Anyway, this is just a nomenclature suggestion.

gammapy/astro/darkmatter/spectra.py
```diff
@@ -216,22 +214,28 @@ class DarkMatterAnnihilationSpectralModel(SpectralModel):
...
-    def __init__(self, mass, channel, scale=scale.quantity, jfactor=1, z=0, k=2):
+    def __init__(self, mass, channel, scale=scale.quantity, jfactor=jfactor, z=0, k=2):
         self.k = k
         self.z = z
         self.mass = u.Quantity(mass)
         self.channel = channel
-        self.jfactor = u.Quantity(jfactor)
         self.primary_flux = PrimaryFlux(mass, channel=self.channel)
-        super().__init__(scale=scale)
+        super().__init__(scale=scale, jfactor=jfactor)
```
alessandromontanari, line 232: I agree on all four changes above, along the philosophy of making the `jfactor` a parameter

alessandromontanari (approved):
As I already wrote before in the few comments, I agree with all these changes. The only thing that I would suggest to implement, is the change of `jfact_decay` to `dfact` in the couple of instances where this appears -- in the tutorial and the test_utils. But as I said before, this is just a minor matter of nomenclature.
Thanks a lot for all this work!

gammapy/astro/darkmatter/spectra.py
```diff
@@ -198,8 +197,7 @@ class DarkMatterAnnihilationSpectralModel(SpectralModel):
 
         >>> channel = "b"
         >>> massDM = 5000*u.Unit("GeV")
-        >>> jfactor = 3.41e19 * u.Unit("GeV2 cm-5")
-        >>> modelDM = DarkMatterAnnihilationSpectralModel(mass=massDM, channel=channel, jfactor=jfactor)  # noqa: E501
+        >>> modelDM = DarkMatterAnnihilationSpectralModel(mass=massDM, channel=channel)  # noqa: E501
```
QRemy, line 200: why removing the jfactor from the example, this pattern should still work no ?

gammapy/astro/darkmatter/spectra.py
```diff
@@ -291,23 +298,21 @@ class DarkMatterDecaySpectralModel(SpectralModel):
...
+    This is how to instantiate a `DarkMatterDecaySpectralModel` model::
 
         >>> import astropy.units as u
         >>> from gammapy.astro.darkmatter import DarkMatterDecaySpectralModel
 
         >>> channel = "b"
         >>> massDM = 5000*u.Unit("GeV")
-        >>> jfactor = 3.41e19 * u.Unit("GeV cm-2")
-        >>> modelDM = DarkMatterDecaySpectralModel(mass=massDM, channel=channel, jfactor=jfactor)  # noqa: E501
+        >>> modelDM = DarkMatterDecaySpectralModel(mass=massDM, channel=channel)  # noqa: E501
```
QRemy, line 315: same here why removing the jfactor from the example ? this pattern should still work

gammapy/astro/darkmatter/spectra.py
```diff
@@ -325,21 +330,27 @@ class DarkMatterDecaySpectralModel(SpectralModel):
...
+        "dfactor",
+        1,
+        unit="GeV cm-2",
+        interp="log",
+    )
+
     tag = ["DarkMatterDecaySpectralModel", "dm-decay"]
 
-    def __init__(self, mass, channel, scale=scale.quantity, jfactor=1, z=0):
+    def __init__(self, mass, channel, scale=scale.quantity, dfactor=dfactor, z=0):
```
QRemy, line 342: the mass could be a parameter frozen by default, but it could be done in a separate PR.

QRemy (commented):
Thanks @StFroese, I left some minor comments and pushed a commit to simply the serialisation.

examples/tutorials/api/astro_dark_matter.py
```diff
@@ -220,7 +220,8 @@
 massDM = 10 * u.TeV
 diff_flux = DarkMatterAnnihilationSpectralModel(mass=massDM, channel=channel)
 int_flux = (
-    jfact * diff_flux.integral(energy_min=0.1 * u.TeV, energy_max=10 * u.TeV)
+    jfact.to_value("GeV2 cm-5")
```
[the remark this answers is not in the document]
registerrier: I think it is needed because the change of default unit in `jfactor` changes the dimensionality of `diff_flux`.

registerrier (commented):
Thanks @StFroese . This is nice. Some general comments:

- Regarding the notebook. It might be better to have a mechanism that produces a `TemplateSpatialModel` from the `JFactory` so that unit handling is more straightforward.
- Regarding the `from_dict` method with additional `kwargs` from the dictionnary: this might be unsafe as any change will break forward/backward compatibility for additional entries. Opinion @QRemy ?

gammapy/modeling/models/core.py
```diff
@@ -279,6 +279,10 @@ def from_dict(cls, data, **kwargs):
             data["parameters"], cls.default_parameters
         )
 
+        for key in data.keys():
```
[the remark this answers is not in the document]
QRemy: it applies only to DM models.
I have no strong opinion about that so I can drop this commit if you prefer

registerrier:
What is the status here @StFroese ?

StFroese:
@registerrier I currently have only limited time due to finishing my thesis, I think I will pick this up again in mid October

registerrier:
Any update here @StFroese ? What should we do?

registerrier:
Becasue this is now diverging a lot, we propose to close this PR and let the team looking into DM and gammapy look into this again.

PR #5113, opened 2024-02-19
Small update of the installation instructions

bkhelifi:
**Description**

This pull request tries to orientate users to use the environment file method to install Gammapy as a stand-alone projet. Otherwise, users should check the dependency versions.

It removes also the '$' in the bash commands... Some are making a copy/paste with them and do not understand immediately why it is not working.

### DIFF
--- a/examples/tutorials/analysis-1d/spectral_analysis.py
+++ b/examples/tutorials/analysis-1d/spectral_analysis.py
@@ -418,6 +418,12 @@ flux_points.plot(ax=ax, sed_type="e2dnde", color="darkorange")
 flux_points.plot_ts_profiles(ax=ax, sed_type="e2dnde")
 plt.show()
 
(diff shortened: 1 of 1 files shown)


QRemy:
Is it in conflict with #4983 ?

adonath (commented):
Thanks @bkhelifi, I have left some inline comments.

Astro-Kirsty:
#4983 has been merged, so this will need to be rebased. 
I suggest removing the adaptions in `index.rst` as this has since been discussed and adjusted in #4983 

Otherwise I agree with removing the `$`

QRemy:
This was close automatically because nothing new was left solving the conflicts with https://github.com/gammapy/gammapy/pull/4983.
Removing the $ in bash commands will be done in a separate PR for the whole documentation.

PR #5117, opened 2024-02-22
Fix `meta_table` times for `FixedPointingInfo`

StFroese:
Simple fix, `TSTART` was written twice into the meta_table


QRemy (approved):
Thanks @StFroese, could you add a test ?

registerrier (approved):
Thanks @StFroese . Let's merge now.

PR #5120, opened 2024-02-26
TODO: for datasets/evaluator

Astro-Kirsty:
Adjusting to simplify if statements. Ran the tests and all of them are okay.
For the 2nd TODO, not sure if anything further can be done?

### DIFF
--- a/gammapy/datasets/evaluator.py
+++ b/gammapy/datasets/evaluator.py
@@ -126,16 +126,17 @@ class MapEvaluator:
     @property
     def needs_update(self):
         """Check whether the model component has drifted away from its support."""
-        if self.exposure is None:
+        if isinstance(self.model, TemplateNPredModel):
+            return False
+        elif not self.contributes:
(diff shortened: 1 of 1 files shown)


adonath (approved):
Thanks @Astro-Kirsty, I think the old version was a bit more readable, but this is just a matter of taste...so I'm fine to merge.

registerrier (changes requested):
Thanks @Astro-Kirsty . The new version is causing some test fails because all conditions have to be tested at once with the `or` syntax.

Astro-Kirsty:
Given that we have test fails and it seems a matter of opinion on the style of code, I would suggest we keep the old version and remove the TODO. Any issues?

Astro-Kirsty:
Do we want to create an issue for the long term goal of rewriting `MapEvaluator` then?

registerrier (approved):
Thanks @Astro-Kirsty

PR #5122, opened 2024-02-27
Fix example and docstrings in irfs

Astro-Kirsty:
This pull request fixes the example in `gammapy/irf/edisp/kernel` so that it actually runs when testing docstrings.

Also fixed a few docstrings


MRegeard (approved):
Thanks @Astro-Kirsty.

registerrier (approved):
Thanks @Astro-Kirsty . This looks good! No comment.

PR #5126, opened 2024-02-27
Remove unneeded TODO from `gammapy/irf`

Astro-Kirsty:
**Description**
This pull request removes some TODO and creates issues out of them.

### DIFF
--- a/gammapy/estimators/points/core.py
+++ b/gammapy/estimators/points/core.py
@@ -268,7 +268,6 @@ class FluxPoints(FluxMaps):
             sed_type = self.sed_type_init
         table = self.to_table(sed_type=sed_type, format=format)
 
-        # TODO: rather ugly - better method?
         if ".fits" not in filename.suffixes:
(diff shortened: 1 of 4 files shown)


adonath:
Thanks @Astro-Kirsty! I think the comments are still valid, would you mind transferring those to issues instead? Thus they become more visible...

gammapy/estimators/points/core.py
```diff
@@ -268,7 +268,6 @@ def write(
             sed_type = self.sed_type_init
         table = self.to_table(sed_type=sed_type, format=format)
 
-        # TODO: rather ugly - better method?
```
Astro-Kirsty, line 271: @adonath this TODO was written when the code was different, and then it was updated but the TODO was never removed. Do you think it is still ugly?

registerrier (approved):
Thanks @Astro-Kirsty . 

The corresponding issues are #5141 and #5142

PR #5127, opened 2024-02-27
TODO: simplifications made to `gammapy.maps`

Astro-Kirsty:
See title

Did my best to simplify 

The workaround in `maps/region/geom` line 785 has been fixed. It should be okay now, but @registerrier  or @adonath can you confirm this?

### DIFF
--- a/gammapy/maps/region/geom.py
+++ b/gammapy/maps/region/geom.py
@@ -782,6 +782,10 @@ class RegionGeom(Geom):
             regions = []
 
             for region_pix in regions_pix:
+                # TODO: remove workaround once regions issue with fits serialization is sorted out
+                # see https://github.com/astropy/regions/issues/400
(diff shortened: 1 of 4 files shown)


gammapy/maps/region/geom.py
```diff
@@ -782,9 +782,6 @@ def from_hdulist(cls, hdulist, format="ogip", hdu=None):
             regions = []
 
             for region_pix in regions_pix:
-                # TODO: remove workaround once regions issue with fits serialization is sorted out
-                # see https://github.com/astropy/regions/issues/400
-                region_pix.meta["include"] = True
```
adonath, line 787: I think this requires to update regions to >=v0.6. Currently we only require >=v0.5.

registerrier (approved):
Thanks @Astro-Kirsty .

PR #5129, opened 2024-02-28
Copy on and off phase intervals in `PhaseBackgroundMaker`

MRegeard:
This PR modify the `PhasebackgroundMaker` so that the phase intervals that are passed to the maker are copied. This is because `_check_intervals` can modify the phase intervals, which can lead to undefined behaviour, especially since this methods is private.

### DIFF
--- a/gammapy/makers/background/phase.py
+++ b/gammapy/makers/background/phase.py
@@ -1,4 +1,5 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
+from copy import deepcopy
 import numpy as np
 from regions import PointSkyRegion
 from gammapy.data import EventList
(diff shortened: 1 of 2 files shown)


AtreyeeS (commented):
I am a bit confused. Isn't it the job the `_check_intervals` to do the modification such that there are no bins beyond 0 and 1? If you want to prevent that, why call `check_intervals` ?

MRegeard:
> I am a bit confused. Isn't it the job the `_check_intervals` to do the modification such that there are no bins beyond 0 and 1? If you want to prevent that, why call `check_intervals` ?

Yes indeed it is the job of `_check_intervals` to modify the intervals for the `PhaseBackgroundMaker`. However doing so it also changes the input intervals (if you pass it via a variable). :
```
on_phase = (0.2, 0.4)
off_phase = (0.5, 0.6)
PhaseBackgroundMaker(on_phase, off_phase)
```
Then if the user want to re-use these variable in the code, they might not be in the way they intended. 
Is it clearer @AtreyeeS ?

AtreyeeS:
Indeed I see now! Thanks @MRegeard ! Then maybe `_check_intervals` should be modified to return a new `intervals`?
Otherwise, Ok from me

adonath (approved):
Thanks @MRegeard, no comments from my side.

registerrier (approved):
Thanks @MRegeard . No comment.

registerrier:
Is it a bug correction? Does it need to be backported?

PR #5135, opened 2024-02-29
Add different options to compute stat_array on FluxPointsDatasets

QRemy:
Add different options to compute stat_array on FluxPointsDatasets :
- chi2 : etimate from chi2 statistics.
- profile : estimate from interpolation of the likelihood profile.
- distrib : estimate from probability distributions,
             assumes that flux points correspond to asymmetric gaussians and upper limits complemantary error functions.

 Default is `chi2`, in that case upper limits are ignored and the mean of asymetrics error is used.
The `distrib` case provides an approximation if the `profile` is not available
 which allows to take into accounts upper limits and asymetrics errors.

Updated /tutorials/analysis-1d/spectral_analysis.py to present the different cases:
image
Without ul the results are the same:
image

### DIFF
--- a/examples/tutorials/analysis-1d/spectral_analysis.py
+++ b/examples/tutorials/analysis-1d/spectral_analysis.py
@@ -380,23 +380,13 @@ plt.show()
 #
 # To round up our analysis we can compute flux points by fitting the norm
 # of the global model in energy bands.
-# We can utilise the `~gammapy.estimators.utils.resample_energy_edges`
-# for defining the energy bins in which the minimum number of `sqrt_ts` is 2.
-# To do so we first stack the individual datasets, only for obtaining the energies:
-#
-
-dataset_stacked = Datasets(datasets).stack_reduce()
-energy_edges = resample_energy_edges(dataset_stacked, conditions={"sqrt_ts_min": 2})
-energy_edges = energy_axis.edges
-
-######################################################################
-# Now we create an instance of the
+# We create an instance of the
 # `~gammapy.estimators.FluxPointsEstimator`, by passing the dataset and
 # the energy binning:
 #
 
 fpe = FluxPointsEstimator(
-    energy_edges=energy_edges, source="crab", selection_optional="all"
+    energy_edges=energy_axis.edges, source="crab", selection_optional="all"
 )
 flux_points = fpe.run(datasets=datasets)
 
@@ -430,7 +420,9 @@ plt.show()
 # quickly made like this:
 #
 
-flux_points_dataset = FluxPointsDataset(data=flux_points, models=model_best_joint)
+flux_points_dataset = FluxPointsDataset(
+    data=flux_points, models=model_best_joint.copy()
+)
 flux_points_dataset.plot_fit()
 plt.show()
 
@@ -504,6 +496,82 @@ plt.show()
 # sphinx_gallery_thumbnail_number = 5
 
 ######################################################################
+# A note on statistics
+# ---------
+#
+# Different statistic are available for the FluxPointDataset :
+# * chi2 : etimate from chi2 statistics.
+# * profile : estimate from interpolation of the likelihood profile.
+# * distrib : estimate from probability distributions,
+#             assumes that flux points correspond to asymmetric gaussians
+#             and upper limits complemantary error functions.
+# Default is `chi2`, in that case upper limits are ignored and the mean of asymetrics error is used.
+# So it is recommended to use `profile` if `stat_scan` is available on flux points.
+# The `distrib` case provides an approximation if the `profile` is not available
+# which allows to take into accounts upper limit and asymetrics error.
+#
+# In the example below we can see that the `profle` case match exactly the result
+# from the joint analysis (from the ON/OFF dataset using labelled as wstat).
+
+
+def plot_stat(fp_dataset):
+    fig, ax = plt.subplots()
+
+    plot_kwargs = {
+        "energy_bounds": [0.1, 30] * u.TeV,
+        "sed_type": "e2dnde",
+        "ax": ax,
+    }
+
+    fp_dataset.data.plot(energy_power=2, ax=ax)
+    model_best_joint.spectral_model.plot(
+        color="b", lw=0.5, **plot_kwargs, label="wstat"
+    )
+
+    stat_types = ["chi2", "profile", "distrib"]
+    colors = ["red", "g", "c"]
+    lss = ["--", ":", "--"]
+
+    for ks, stat in enumerate(stat_types):
+
+        fp_dataset.stat_type = stat
+
+        fit = Fit()
+        fit.run([fp_dataset])
+
+        fp_dataset.models[0].spectral_model.plot(
+            color=colors[ks], ls=lss[ks], **plot_kwargs, label=stat
+        )
+        fp_dataset.models[0].spectral_model.plot_error(
+            facecolor=colors[ks], **plot_kwargs
+        )
+        plt.legend()
+
+
+plot_stat(flux_points_dataset)
+
+######################################################################
+
+# In order to avoid discrepancies due to the treatment of upper limits
+# we can utilise the `~gammapy.estimators.utils.resample_energy_edges`
+# for defining energy bins in which the minimum number of `sqrt_ts` is 2.
+# In that case the all the statistics definitions gives equivalent results.
+#
+
+energy_edges = resample_energy_edges(dataset_stacked, conditions={"sqrt_ts_min": 2})
+
+fpe_no_ul = FluxPointsEstimator(
+    energy_edges=energy_edges, source="crab", selection_optional="all"
+)
+flux_points_no_ul = fpe_no_ul.run(datasets=datasets)
+flux_points_dataset_no_ul = FluxPointsDataset(
+    data=flux_points_no_ul,
+    models=model_best_joint.copy(),
+)
+
+plot_stat(flux_points_dataset_no_ul)
+
+######################################################################
 # Exercises
 # ---------
 #
@@ -533,122 +601,3 @@ plt.show()
 # the :doc:`/tutorials/analysis-1d/extended_source_spectral_analysis`
 # tutorial.
 #
-#%%
-
-fig, ax = plt.subplots()
-
-plot_kwargs = {
-    "energy_bounds": [0.1, 30] * u.TeV,
-    "sed_type": "e2dnde",
-    "ax": ax,
-}
-
-flux_points_dataset.data.plot(energy_power=2, ax=ax)
-model_best_joint.spectral_model.plot(color="b", lw=0.5, **plot_kwargs, label="wstat")
-
-#%%
-
-flux_points_dataset.stat_type = "chi2"
-
-fit = Fit()
-result = fit.run([flux_points_dataset])
-print(result)
-model_best_fp = flux_points_dataset.models[0].copy()
-print(flux_points_dataset.models[0].parameters.to_table())
-
-flux_points_dataset.models[0].spectral_model.plot(
-    color="red", ls="--", **plot_kwargs, label="chi2"
-)
-
-
-stop
-flux_points_dataset.stat_type = "profile"
-fit = Fit()
-result = fit.run([flux_points_dataset])
-print(result)
-
-flux_points_dataset.models[0].spectral_model.plot(
-    color="g", ls=":", **plot_kwargs, label="profile"
-)
-flux_points_dataset.models[0].spectral_model.plot_error(facecolor="g", **plot_kwargs)
-model_best_fp_profile = flux_points_dataset.models[0].copy()
-plt.legend()
-plt.show()
-print(flux_points_dataset.models[0].parameters.to_table())
-
-
-#%%
-pars = []
-
-
-def fit_sample(k):
-    flux_points_dataset = FluxPointsDataset(
-        data=flux_points,
-        models=model_best_joint.copy(),
-        stat_type="chi2_samples",
-        stat_kwargs=dict(n_samples=100, random_state=k),
-    )
-    fit = Fit()
-    fit.run([flux_points_dataset])
-    return flux_points_dataset.models[0].parameters.free_parameters.value
-
-
-from gammapy.utils.parallel import run_multiprocessing
-
-results = run_multiprocessing(
-    fit_sample,
-    zip(
-        range(1000),
-    ),
-    backend="ray",
-    pool_kwargs=dict(processes=4),
-    task_name="fp sampling",
-)
-
-#%%
-import corner
-
-samples = np.array(results)
-ndim = samples.shape[1]
-# data fit parameters
-value0 = model_best_joint.parameters.free_parameters.value
-value1 = model_best_fp.parameters.free_parameters.value
-
-# fp samples fit parameters
-value2 = model_best_fp_profile.parameters.free_parameters.value
-
-# fp samples fit parameters
-value3 = np.median(samples, axis=0)
-
-# Make the base corner plot
-figure = corner.corner(samples)
-
-# Extract the axes
-axes = np.array(figure.axes).reshape((ndim, ndim))
-
-# Loop over the diagonal
-for i in range(ndim):
-    ax = axes[i, i]
-    ax.axvline(value0[i], color="b")
-    ax.axvline(value1[i], color="r")
-    ax.axvline(value2[i], color="g", ls="--")
-    ax.axvline(value3[i], color="c", ls="--")
-# Loop over the histograms
-for yi in range(ndim):
-    for xi in range(yi):
-        ax = axes[yi, xi]
-        ax.axvline(value0[xi], color="b")
-        ax.axvline(value1[xi], color="r")
-        ax.axvline(value2[xi], color="g", ls="--")
-        ax.axhline(value3[yi], color="c", ls="--")
-        ax.axhline(value0[yi], color="b")
-        ax.axhline(value1[yi], color="r")
-        ax.axhline(value2[yi], color="g", ls="--")
-        ax.axhline(value3[yi], color="c", ls="--")
-
-        # ax.plot(value0[xi], value0[yi], "sb")
-        # ax.plot(value1[xi], value1[yi], "sr")
-        # ax.plot(value2[xi], value2[yi], "sg")
-
-
-#%%
--- a/examples/tutorials/analysis-1d/spectral_analysis.py
+++ b/examples/tutorials/analysis-1d/spectral_analysis.py
@@ -510,8 +510,8 @@ plt.show()
 # The `distrib` case provides an approximation if the `profile` is not available
 # which allows to take into accounts upper limit and asymetrics error.
 #
-# In the example below we can see that the `profle` case match exactly the result
-# from the joint analysis (from the ON/OFF dataset using labelled as wstat).
+# In the example below we can see that the `profile` case matches exactly the result
+# from the joint analysis of the ON/OFF datasets using `wstat` (as labelled)).
 
 
 def plot_stat(fp_dataset):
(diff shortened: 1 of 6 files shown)


gammapy/datasets/flux_points.py
```diff
@@ -122,11 +139,52 @@ def __init__(
         self.models = models
         self.meta_table = meta_table
 
-        if mask_safe is None:
-            mask_safe = (~data.is_ul).data
+        self.available_stat_type = dict(
```
AtreyeeS, line 142: `FluxPointsDataset.available_stat_types` can be a property like in `FluxPoints.available_quantities`

gammapy/datasets/flux_points.py
```diff
@@ -55,6 +58,19 @@ class FluxPointsDataset(Dataset):
...
+    stat_type : bool
+        Method used to compute the statistics:
+        * chi2 : etimate from chi2 statistics.
+        * profile : estimate from interpolation of the likelihood profile.
+        * distrib : estimate from probability distributions
+                    assuming that flux points correspond to asymmetric gaussians
+                    and upper limits complemantary error functions.
+        Default is `chi2`, in that case upper limits are ignored and the mean of asymetrics error is used.
+        However it is recommended to use `profile` if `stat_scan` is available on flux points.
+    stat_kwargs : int
```
AtreyeeS, line 70:
```suggestion
    stat_kwargs : dict
```

gammapy/datasets/flux_points.py
```diff
@@ -55,6 +58,19 @@ class FluxPointsDataset(Dataset):
...
+        Exta arguments passed to the stat function
```
AtreyeeS, line 71: It is difficult to understand what can be configured. Maybe 

Extra arguments specifying the interpolation scheme of the likelihood profile. Used only in case `stat_type=="profile"`

gammapy/datasets/flux_points.py
```diff
@@ -122,11 +139,52 @@ def __init__(
...
+            distrib=self._stat_array_distrib,
+        )
+
+        if stat_kwargs is None:
+            stat_kwargs = dict()
+        self.stat_kwargs = stat_kwargs
 
+        self.stat_type = stat_type
+
+        if mask_safe is None:
```
AtreyeeS, line 154: In a different PR, maybe we should make the masks as Maps to be consistent with the `MapDataset`

gammapy/datasets/flux_points.py
```diff
@@ -122,11 +139,52 @@ def __init__(
...
+
+        if stat_type not in self.available_stat_type:
+            raise ValueError(
+                "Invalid stat_type: possible options are {list(self.available_stat_type.keys())}"
+            )
+
+        if stat_type == "chi2":
+            self.mask_valid = (~self.data.is_ul).data & np.isfinite(self.data.dnde)
+        elif stat_type == "distrib":
+            self.mask_valid = self.data.is_ul.data | np.isfinite(self.data.dnde)
```
AtreyeeS, line 173: How does it work when the flux is `nan` ?

gammapy/datasets/flux_points.py
```diff
@@ -122,11 +139,52 @@ def __init__(
...
+    @stat_type.setter
+    def stat_type(self, stat_type):
+
+        if stat_type not in self.available_stat_type:
+            raise ValueError(
+                "Invalid stat_type: possible options are {list(self.available_stat_type.keys())}"
+            )
+
+        if stat_type == "chi2":
+            self.mask_valid = (~self.data.is_ul).data & np.isfinite(self.data.dnde)
```
AtreyeeS, line 171: Both `mask_safe` and `mask_valid` are necessary?

gammapy/datasets/flux_points.py
```diff
@@ -310,13 +368,82 @@ def flux_pred(self):
...
+        """Estimate statistic from probability distributions,
+        assumes that flux points correspond to asymmetric gaussians
+        and upper limits complemantary error functions.
+        """
+        model = self.flux_pred()
+
+        is_ul = self.data.is_ul.data.flatten()
+        n_points = len(is_ul)
+        stat = np.zeros(model.shape)
+        for k in range(n_points):
```
AtreyeeS, line 420: Does this work correctly for `FluxPoints` with both `energy` and `time` axes?

AtreyeeS (changes requested):
Thanks @QRemy ! The `profile` is an elegant solution for considering upper limits during the fit!

I don't understand the implementation of the `distrib` option. Maybe you can add some information in `gammapy.stat` docs page?

Otherwise, only minor comments inline

gammapy/datasets/flux_points.py
```diff
@@ -55,6 +58,19 @@ class FluxPointsDataset(Dataset):
...
+        * chi2 : etimate from chi2 statistics.
+        * profile : estimate from interpolation of the likelihood profile.
+        * distrib : estimate from probability distributions
+                    assuming that flux points correspond to asymmetric gaussians
+                    and upper limits complemantary error functions.
+        Default is `chi2`, in that case upper limits are ignored and the mean of asymetrics error is used.
+        However it is recommended to use `profile` if `stat_scan` is available on flux points.
+    stat_kwargs : int
+        Exta arguments passed to the stat function
+        Used only if `stat_type=="distrib"`. In that case the default is :
```
QRemy, line 72:
```suggestion
        Used only if `stat_type=="profile"`. In that case the default is :
```

gammapy/datasets/flux_points.py
QRemy, line 71:
```suggestion
        Extra arguments specifying the interpolation scheme of the likelihood profile.
```

gammapy/datasets/flux_points.py
adonath: Please make `mask_valid` a lazy property then, to make it read only.

gammapy/datasets/flux_points.py
QRemy: fixed

QRemy:
Thanks @AtreyeeS I implemented most of the comments.

> I don't understand the implementation of the `distrib` option. Maybe you can add some information in `gammapy.stat` docs page?

Assuming gaussian errors the likelihood is given by the probability density function of the normal distribution.
For the upper limit case it is necessary to marginalize over the unknown measurement so we integrate the normal distribution up to the upper limit value which gives the complementary error function.

Screenshot 2024-03-06 at 17 03 23
see eq. C7 from https://iopscience.iop.org/article/10.1088/0004-637X/773/2/168/pdf
(in eq C6 they are doing something more complicated as they also marginalised over the unknown true value)

adonath:
Thanks @QRemy, the statistical approach to handle the ULs looks reasonable to me. I was wonderign whether we can maybe find a more descriptive name. I think even "chi2-marginalize-ul" would be better. We should definitely give the paper reference in addition.

gammapy/datasets/flux_points.py
QRemy: done

AtreyeeS (dismissed):
Thank you @QRemy ! Elegant indeed.
Maybe you can add a little info the documentation (what you wrote in the comment, and the paper reference). 
And add `stat_type` option in the example here https://github.com/gammapy/gammapy/blob/fdd05df7453c294a8a898e7f790a0da17372f9d2/examples/tutorials/analysis-time/light_curve_simulation.py#L295

gammapy/datasets/flux_points.py
```diff
@@ -122,11 +145,60 @@ def __init__(
...
+            profile=self._stat_array_profile,
+            distrib=self._stat_array_distrib,
+        )
+
+    @property
+    def stat_type(self):
+        return self._stat_type
+
+    @stat_type.setter
+    def stat_type(self, stat_type):
```
AtreyeeS, line 171: Does this not need a `stat_kwargs` for the `distrib` option?

gammapy/datasets/flux_points.py
```diff
@@ -310,13 +382,84 @@ def flux_pred(self):
...
+        return stat
+
+    def _get_valid_profile_interpolators(self):
+        value_scan = self.data.stat_scan.geom.axes["norm"].center
+        shape_axes = self.data.stat_scan.geom._shape[slice(3, None)]
+        interpolators = np.empty(shape_axes, dtype=object)
+        self._mask_valid = np.ones(self.data.dnde.data.shape, dtype=bool)
+        for idx in np.ndindex(shape_axes):
+            stat_scan = np.abs(
+                self.data.stat_scan.data[idx].squeeze()
```
AtreyeeS, line 412: I think this is not working correctly. Running locally on my machine, I get this error 

```
IndexError                                Traceback (most recent call last)
Cell In[14], line 1
----> 1 dataset_fp.stat_type =  "profile"
      2 dataset_fp.stat_type
      3 dataset_fp.models = model

File ~/Gammapy-dev/gammapy/gammapy/datasets/flux_points.py:187, in FluxPointsDataset.stat_type(self, stat_type)
    185     self.stat_kwargs.setdefault("interp_scale", "sqrt")
    186     self.stat_kwargs.setdefault("extrapolate", True)
--> 187     self._profile_interpolators = self._get_valid_profile_interpolators()
    188 self._stat_type = stat_type

File ~/Gammapy-dev/gammapy/gammapy/datasets/flux_points.py:412, in FluxPointsDataset._get_valid_profile_interpolators(self)
    409 self._mask_valid = np.ones(self.data.dnde.data.shape, dtype=bool)
    410 for idx in np.ndindex(shape_axes):
    411     stat_scan = np.abs(
--> 412         self.data.stat_scan.data[idx].squeeze()
    413         - self.data.stat.data[idx].squeeze()
    414     )
    415     self._mask_valid[idx] = np.all(np.isfinite(stat_scan))
    416     interpolators[idx] = interpolate_profile(
    417         value_scan,
    418         stat_scan,
    419         interp_scale=self.stat_kwargs["interp_scale"],
    420         extrapolate=self.stat_kwargs["extrapolate"],
    421     )

IndexError: index 3 is out of bounds for axis 1 with size 3
```

Possibly the same issue as #5155

gammapy/datasets/tests/test_flux_points.py
```diff
@@ -158,6 +171,9 @@ def test_flux_point_dataset_with_time_axis(tmp_path):
     with pytest.raises(ValueError):
         flux_points_dataset.plot_residuals()
 
+    flux_points_dataset.stat_type = "distrib"
```
AtreyeeS, line 174: Add a test for the `profile` case as well

gammapy/datasets/flux_points.py
```diff
@@ -122,11 +145,60 @@ def __init__(
...
+    @property
+    def stat_type(self):
+        return self._stat_type
+
+    @stat_type.setter
+    def stat_type(self, stat_type):
+
+        if stat_type not in self.available_stat_type:
+            raise ValueError(
+                "Invalid stat_type: possible options are {list(self.available_stat_type.keys())}"
```
AtreyeeS, line 175: Missing the `f` here.. the list does not render correctly

AtreyeeS (changes requested):
The `distrib` option is often returning a `nan` for some bins, leading to the total stat becoming `nan` and a failure of the minimisation. Any idea why?

registerrier (commented):
Thanks @QRemy ! I have left some comments inline.
 
My main comment is that you could compute most of the necessary arrays in advance. I suppose that for large SEDs this might be much more efficient.

gammapy/datasets/flux_points.py
QRemy, line 175:
```suggestion
                f"Invalid stat_type: possible options are {list(self.available_stat_type.keys())}"
```

gammapy/datasets/flux_points.py
QRemy: no only for the profile case I modified the doc

gammapy/datasets/flux_points.py
```diff
@@ -310,13 +382,84 @@ def flux_pred(self):
...
+            stat[idx] = interp(model[idx])
+        return stat
+
+    def _get_valid_profile_interpolators(self):
+        value_scan = self.data.stat_scan.geom.axes["norm"].center
+        shape_axes = self.data.stat_scan.geom._shape[slice(3, None)]
+        interpolators = np.empty(shape_axes, dtype=object)
+        self._mask_valid = np.ones(self.data.dnde.data.shape, dtype=bool)
+        for idx in np.ndindex(shape_axes):
+            stat_scan = np.abs(
```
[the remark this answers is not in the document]
QRemy: here I just coped the code from FluxPoints.recompute_ul

gammapy/datasets/flux_points.py
QRemy: The shape was reversed solved with :
`shape_axes = self.stat_scan.geom._shape[slice(3, None)][::-1]`

gammapy/datasets/flux_points.py
```diff
@@ -310,13 +382,84 @@ def flux_pred(self):
...
+        assumes that flux points correspond to asymmetric gaussians
+        and upper limits complemantary error functions.
+        """
+        model = self.flux_pred()
+
+        stat = np.zeros(model.shape)
+        for idx in np.ndindex(model.shape):
+            value = model[idx].to_value(self.data.dnde.unit)
+
+            if not self.data.is_ul.data[idx]:
```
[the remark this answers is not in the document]
QRemy: This can be done by adapting sqrt_threshold_ul, so for me it is user responsibility.

gammapy/datasets/flux_points.py
```diff
@@ -310,13 +382,84 @@ def flux_pred(self):
...
+                if np.isnan(self.data.dnde_ul.data[idx]):
+                    continue
+
+                loc = 0
+                loc_lim = self.data.dnde_ul.data[idx]
+                scale = self.data.dnde_ul.data[idx]
+                stat[idx] = 2 * np.log(
+                    (erfc((loc_lim - value) / scale) / 2)
+                    / (erfc((loc_lim - loc) / scale) / 2)
+                )
```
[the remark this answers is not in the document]
QRemy: done

gammapy/datasets/flux_points.py
```diff
@@ -310,13 +382,84 @@ def flux_pred(self):
...
+        shape_axes = self.data.stat_scan.geom._shape[slice(3, None)]
+        interpolators = np.empty(shape_axes, dtype=object)
+        self._mask_valid = np.ones(self.data.dnde.data.shape, dtype=bool)
+        for idx in np.ndindex(shape_axes):
+            stat_scan = np.abs(
+                self.data.stat_scan.data[idx].squeeze()
+                - self.data.stat.data[idx].squeeze()
+            )
+            self._mask_valid[idx] = np.all(np.isfinite(stat_scan))
+            interpolators[idx] = interpolate_profile(
```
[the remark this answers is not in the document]
QRemy: no it's directly the likelihood, same as the code for recompute_ul

examples/tutorials/analysis-1d/spectral_analysis.py
```diff
@@ -498,6 +492,82 @@
 
 # sphinx_gallery_thumbnail_number = 5
 
+######################################################################
```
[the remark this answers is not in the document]
QRemy: I would suggest to do that in a separate PR.

examples/tutorials/analysis-1d/spectral_analysis.py
```diff
@@ -497,6 +491,82 @@
...
+# A note on statistics
+# --------------------
+#
+# Different statistic are available for the FluxPointDataset :
+# * chi2 : estimate from chi2 statistics.
+# * profile : estimate from interpolation of the likelihood profile.
+# * distrib : estimate from probability distributions,
+#             assumes that flux points correspond to asymmetric gaussians
```
bkhelifi, line 502:
```suggestion
#             assuming that flux points correspond to asymmetric gaussians
```

examples/tutorials/analysis-1d/spectral_analysis.py
```diff
@@ -497,6 +491,82 @@
...
+#             and upper limits complementary error functions.
+# Default is `chi2`, in that case upper limits are ignored and the mean of asymetrics error is used.
+# So it is recommended to use `profile` if `stat_scan` is available on flux points.
+# The `distrib` case provides an approximation if the `profile` is not available
+# which allows to take into accounts upper limit and asymetrics error.
+#
+# In the example below we can see that the `profile` case matches exactly the result
+# from the joint analysis of the ON/OFF datasets using `wstat` (as labelled)).
```
bkhelifi, line 510:
```suggestion
# from the joint analysis of the ON/OFF datasets using `wstat` (as labelled).
```

gammapy/utils/interpolation.py
```diff
@@ -239,7 +239,7 @@ def _inverse(cls, values):
         return values
 
 
-def interpolate_profile(x, y, interp_scale="sqrt"):
+def interpolate_profile(x, y, interp_scale="sqrt", extrapolate=False):
```
bkhelifi, line 242: Can you add in the docstring the type of x and y, as well as the return value/quantity?

gammapy/datasets/tests/test_flux_points.py
```diff
@@ -102,6 +102,19 @@ def test_flux_point_dataset_flux_pred(dataset):
...
 
 
+@requires_data()
+def test_flux_point_dataset_stat(dataset):
+    dataset.stat_type = "chi2"
+    fit = Fit()
+    fit.run([dataset])
+    assert_allclose(dataset.stat_sum(), 25.205933, rtol=1e-3)
+
+    dataset.stat_type = "distrib"
```
bkhelifi, line 112: There is no test here with "profile"?

gammapy/estimators/points/tests/test_lightcurve.py
```diff
@@ -738,6 +738,43 @@ def test_recompute_ul():
...
+    )
+    lightcurve = estimator.run(datasets)
+    assert_allclose(
+        lightcurve.dnde_ul.data[0], [[[3.260703e-13]], [[1.159354e-14]]], rtol=1e-3
+    )
+
+    lightcurve_dataset = FluxPointsDataset(
+        data=lightcurve, models=[lightcurve.reference_model]
+    )
+    lightcurve_dataset.stat_type = "profile"
```
bkhelifi, line 756: Why tests are needed here when they are tests in test_flux_points?

bkhelifi (commented):
Thanks @QRemy.. I have some questions for you.

gammapy/datasets/tests/test_flux_points.py
QRemy: `test_flux_points` read a flux points objects for which stat_can is not computed so the test for profile can not be done there and is done in `test_lightcurve` instead.

gammapy/estimators/points/tests/test_lightcurve.py
QRemy: see comment above

gammapy/utils/interpolation.py
QRemy: Added a docstring entry for extrapolate.
For x, y the existing docstring already mentions they are arrays but we cannot say anything about the unit as it is a generic class.

bkhelifi (approved):
Thanks @QRemy . All good for me;)

registerrier (dismissed):

I have left a simple comment regarding not providing the dictionary of stat functions but simply their keys.

Otherwise, this would be much cleaner with a stat functions being defined as external objects. This can be a further PR though.

gammapy/datasets/flux_points.py
```diff
@@ -55,6 +58,25 @@ class FluxPointsDataset(Dataset):
     meta_table : `~astropy.table.Table`
         Table listing information on observations used to create the dataset.
         One line per observation for stacked datasets.
+    stat_type : bool
```
QRemy, line 61:
```suggestion
    stat_type : str
```

PR #5136, opened 2024-03-01
Adapt spectral analysis tutorial to include sed type flux

### LINKED ISSUE #5086: Plot of a SpectralModel with the sed type `flux'
**Gammapy version**
The dev version (I have not tested older versions)

**Bug description**
Using the Spectral analysis tutorial, the script works any issue. I have superposed the SpectralModel on the TS scan plot with the sed_type `e2dnde':
e2dnde

When I use the sed_type `flux', the result is the following:
flux

Astro-Kirsty:
This PR is to resolve #5086

Screenshot shows the doc build of the tutorial

Screenshot from 2024-03-01 10-09-01

### DIFF
--- a/examples/tutorials/analysis-1d/spectral_analysis.py
+++ b/examples/tutorials/analysis-1d/spectral_analysis.py
@@ -418,6 +418,22 @@ flux_points.plot(ax=ax, sed_type="e2dnde", color="darkorange")
 flux_points.plot_ts_profiles(ax=ax, sed_type="e2dnde")
 plt.show()
 
+######################################################################
+# We can also plot the flux distribution with the spectral model overlaid:
+#
+
+fig, ax = plt.subplots()
+flux_points.plot(ax=ax, sed_type="flux", color="darkorange")
+flux_points.plot_ts_profiles(ax=ax, sed_type="flux")
+axis = flux_points.energy_axis
+model_best_joint.spectral_model.plot(
+    ax=ax,
+    energy_bounds=(axis.edges_min[0], axis.edges_max[-1]),
+    sed_type="flux",
+    n_points=axis.nbin,
+)
+plt.show()
+
 
 ######################################################################
 # The final plot with the best fit model, flux points and residuals can be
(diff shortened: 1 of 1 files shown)


examples/tutorials/analysis-1d/spectral_analysis.py
```diff
@@ -419,6 +419,22 @@
 flux_points.plot_ts_profiles(ax=ax, sed_type="e2dnde")
 plt.show()
 
+######################################################################
+# We can also plot the flux distribution with the spectral model overlaid:
+#
```
adonath, line 424: I think it would be good to add a comment on the importance of choosing the same energy binning as for the flux points for integral fluxes.

registerrier (commented):
Thanks @Astro-Kirsty . 
I think one issue here is to plot the spectral model in the form of a continuous line when comparing with integrated fluxes.

AtreyeeS:
Thanks @Astro-Kirsty ! Maybe we can add this (to avoid future queries) with a *strong* caveat on the plotting a model for the integral flux, and explaining the issue.

bkhelifi:
There is a small shift between the continuous flux line and the flux points. As mentioned by Regis, this is due to to the fact that the 'integration' bins between both is different, the bin width for the line is smaller than the one of the bin....
If we decide to add this in this tutorial, one should add a warning/disclaimer to warm users. Otherwise, users could complain about the small shift or believe that there is a bug

Astro-Kirsty:
This PR will be updated once we merge #5222

Astro-Kirsty:
I think we decided to leave out plotting the flux distribution, and just noting it could be done. 

We should merge this now and close  #5086

AtreyeeS (commented):
Maybe you can add a plotting of the spectral model while passing the MapAxis. Then users can just change to `sed_type=flux` if they desire

examples/tutorials/analysis-1d/spectral_analysis.py
```diff
@@ -419,6 +418,28 @@
...
+model_best_joint.spectral_model.plot(
+    ax=ax, sed_type=sed_type, energy_bounds=energy_axis
+)
+model_best_joint.spectral_model.plot_error(
+    ax=ax, sed_type=sed_type, energy_bounds=energy_axis
+)
+plt.show()
+
+######################################################################
+# Note: it is also possible to plot the flux distribution with the spectral model overlaid,
```
QRemy, line 437:
```suggestion
# Note: it is also possible to plot the flux distribution with the spectral model overlaid
```

QRemy (dismissed):
Thanks @Astro-Kirsty, no further comment, merging now.

PR #5145, opened 2024-03-04
LightCurveTemplateTemporalModel method to generate a model from a periodogram by using the Timmer algorithm

cgalelli:
This adds a simple method for `LightCurveTemplateTemporalModel` that creates the model from the user's periodogram/power spectrum. To do so, it runs the algorithm described in [Timmer & Konig, 1995](https://adsabs.harvard.edu/full/1995A%26A...300..707T). The algorithm itself is implemented separately in `stats.variability.py`.

This initial PR state is rough and mostly shows the algorithm implementation and one example case of use.


gammapy/modeling/models/temporal.py
```diff
@@ -617,6 +618,28 @@ def from_table(cls, table, filename=None):
 
         return cls(m, t_ref=t_ref, filename=filename)
 
+    @classmethod
+    def from_powerspectrum(cls, powerspectrum, map, **kwargs):
+        """Create a template model from a power spectrum using the Timmer algorithm.
+
+        Parameters
+        ----------
+        powerspectrum: `~numpy.ndarray`
```
AtreyeeS, line 627: I think that Timmer Konig algorithm creates only power law periodograms, so allowing users to pass a random power spectrum is not safe. Rather take model parameters as input and generate the power spectrum internally

gammapy/modeling/models/temporal.py
```diff
@@ -617,6 +618,28 @@ def from_table(cls, table, filename=None):
...
+            Array representing the power spectrum from which to derive the model.
+        map:  `RegionNDMap`
+            Map containing the desired geometry of the model.
```
AtreyeeS, line 630: I am not sure that it will work correctly if there are gaps in the data. So a map with a `TimeMapAxis` might be an issue

gammapy/modeling/models/temporal.py
```diff
@@ -617,6 +618,28 @@ def from_table(cls, table, filename=None):
...
+
+        Returns
+        -------
+        model : `LightCurveTemplateTemporalModel`
+            Light curve template model.
+        """
+        evenness = map.geom.axes["time"].nbin % 2 == 0
```
AtreyeeS, line 637: this should be determined inside `TimmerAlgorithm`

gammapy/stats/variability.py
```diff
@@ -228,3 +229,65 @@ def compute_flux_doubling(flux, flux_err, coords, axis=0):
     }
 
     return doubling_dict
+
+
+def TimmerAlgorithm(power_spectrum, even_freq=True):
```
AtreyeeS, line 234: I would take the number of points and the slope of the power spectrum as input. `even_freq` then becomes redundant

gammapy/stats/variability.py
```diff
@@ -228,3 +229,65 @@ def compute_flux_doubling(flux, flux_err, coords, axis=0):
...
+    """Implementation of the Timmer algorithm to simulate a time series from a power spectrum
```
AtreyeeS, line 235: Also, take `random_state` as input to configure the random number generator. See `MapDataset.fake` for ex

gammapy/stats/variability.py
```diff
@@ -228,3 +229,65 @@ def compute_flux_doubling(flux, flux_err, coords, axis=0):
...
+    Returns
+    -------
+    time_series: `~numpy.ndarray`
+        Simulated time series.
+
+    References
+    ----------
+    ..[Timmer1995]"On generating power law noise", J. Timmer and M, Konig, section 3.
+    """
+    random_numbers = np.random.normal(
```
AtreyeeS, line 255: You are assuming a time binning of 1.0. Should be configurable, no?

AtreyeeS (commented):
Thank you @cgalelli ! Having a functionality to generate `LightCurves` according to a given power spectrum is quite important for timing studies.

Maybe it would be even be useful to have Models defined in frequencies. This is a topic broader than the scope of this PR.

Here, I am not sure if `LightCurveTemplateTemporalModel` is the best container for a simulated time-series. What the `TimmerAlgorithm` generates is in units of `counts`, with noise included correctly. Simply adding Poissionian noise on top of it (with our `.fake`) method would destroy the power spectrum. 

Maybe you can start with an implementation of the algorithm in `gammapy.stats` which returns a `RegionNDMap`?

gammapy/modeling/models/temporal.py
bkhelifi: For our fields, we will have time gaps in general. So, I am wondering if instead one could not use a list of GTI as input.

cgalelli:
Thanks, @AtreyeeS for all your comments! As said at the start, this was more of a "show the concept" PR, and your suggestions are very helpful in shaping it towards a real usable utility. 

Regarding a library of models defined in frequencies (`PowerLawModel`?), I brought up the idea a couple of weeks ago to @registerrier, maybe we can explore that in a different or future PR/issue.

gammapy/modeling/models/temporal.py
cgalelli: Thanks for the comments @AtreyeeS, @bkhelifi.
 I don't think the algorithm itself is limited to power law noise. To report a paragraph at the end of section 2 of the paper _"For linear stochastic processes the spectrum and the periodogram are obtained by multiplying the results for the white noise process by the filter function (here: a power law) that describes the process. For these processes the variance of the complex random variable f(w) in eq. (9) becomes frequency dependent and is determined by the spectrum. Eq. (13) remains valid. This is also true for nonlinear stochastic processes and even for most chaotic processes."_
The _"here: a power law"_ comment is realized in the algorithm described by the paper by substituting sqrt(1/2 S(w_i)) with the power law case (1/w)^b/2 . I don't do this substitution in my implementation. 

Apart from that, I think the use of an array as input is helpful in the case of a periodogram derived from a dataset (for example using scipy or numpy) - so that the user doesn't have to pass through fitting. I feel that both ideas have their space - maybe as alternative inputs (either a set of parameters - or future "models", or an array/template?).

gammapy/stats/variability.py
cgalelli: Thanks, this is a very important suggestion.

gammapy/stats/variability.py
cgalelli: Absolutely right, thanks. I can resolve this by changing the rng using `random_state`

gammapy/stats/variability.py
cgalelli: As expressed above I feel that passing an array representing the periodogram is sometimes helpful - I can add alternative inputs based on your suggestions. I agree that passing the number of points in this case becomes necessary if the input is analytical, and more compact/elegant also in the case of discrete input.

gammapy/modeling/models/temporal.py
cgalelli: But maybe I misinterpreted the passage of the paper I quoted. We can discuss it!

AtreyeeS: @bkhelifi : I don't know the best way to handle time gaps in the data. During my PhD, I used to simulate time series with uniform time binning, and then apply masks as per the light curves being studied. Passing GTIs would emulate something along that line I imagine. Where the time series is internally generated without gaps and then the GTI selection is applied on the simulated data?

cgalelli:
You can find the `stingray` implementation of this algorithm here: 
[docstring](https://docs.stingray.science/en/stable/_modules/stingray/simulator/simulator.html#Simulator.simulate), and [source](https://docs.stingray.science/en/stable/api.html#stingray.simulator.simulator.Simulator.simulate)

Unfortunately the Timmer - Konig algorithm itself is hidden (as `_timmerkoenig`), and the method itself returns a proprietary format, `stingray.Lightcurve`.

However, their implementation supports the idea that the algorithm can be used for whatever shape of the power spectrum, not only power-law, which is very convenient.

gammapy/modeling/models/temporal.py
bkhelifi: Indeed, it seems the right approach for the simulation. In fact, I was referring to the API and the user workflow, for which I have not yet catched the whole logics...

cgalelli:
In the most recent commit i removed the method from `LightCurveTemplateTemporalModel`, but kept the base algorithm, which then will be able to be used by a `Lightcurve` class

gammapy/stats/tests/test_variability.py
```diff
@@ -143,3 +144,10 @@ def test_lightcurve_flux_doubling():
         [2271.34711286, 21743.98603654] * u.s,
     )
     assert_allclose(dtime_err, [425.92375713, 242.80234065] * u.s)
+
+
+def test_tk():
+    time_series = TKAlgorithm(3, 20, 1, type="powerlaw")
+
+    assert len(time_series) == 20
```
AtreyeeS, line 152: Maybe add test both for odd and even number if bins

gammapy/stats/variability.py
```diff
@@ -228,3 +230,101 @@ def compute_flux_doubling(flux, flux_err, coords, axis=0):
...
+    type="discrete",
+    random_state="random-seed",
+    normalization=1.0,
+):
+    """Implementation of the Timmer-Koenig algorithm to simulate a time series from a power spectrum.
+
+    Parameters
+    ----------
+    power_spectrum : float, `~numpy.ndarray`
+        Power spectrum used to generate the time series. It can either be presented analytically, as a parameter or set
```
AtreyeeS, line 248: It is unclear how to pass the `power_spectrum` analytically. What are the parameters and how are they ordered?

AtreyeeS (approved):
Thank you @cgalelli ! This can go in now, and can be added to the `LightCurve` later

gammapy/stats/variability.py
cgalelli: Is it accepted to have sub-indentations in the docstrings? That would be the simplest way to add the parameter explanations for each supported analytical shape.

gammapy/stats/variability.py
```diff
@@ -228,3 +230,102 @@ def compute_flux_doubling(flux, flux_err, coords, axis=0):
     }
 
     return doubling_dict
+
+
+def TKAlgorithm(
```
registerrier, line 235: I would use an explicit name. The name should convey the information on what the function does.

gammapy/stats/variability.py
```diff
@@ -228,3 +230,102 @@ def compute_flux_doubling(flux, flux_err, coords, axis=0):
...
+
+    random_state = get_random_state(random_state)
+
+    frequencies = np.fft.fftfreq(npoints, spacing)
+
+    # To obtain real data only the positive or negative part of the frequency is necessary.
+    real_frequencies = np.sort(np.abs(frequencies[frequencies < 0]))
+
+    if type == "discrete":
+        periodogram = power_spectrum
```
registerrier, line 287: I think this is a bit dangerous as is. This is not tested and there is no way to make sure frequencies and power spectrum match.

registerrier (changes requested):
Thanks @cgalelli ! Sorry for the late review here.

I think this can be a nice addition but the implementation seems a bit delicate. There is no way to make sure frequencies and input power spectrum really match. 
I can see two easy solutions: 
- remove the `power_spectrum` option
- allow only functions as input. After all passing lambda x: x**-index is relatively easy

In any case, this case should be tested.

gammapy/stats/variability.py
```diff
@@ -228,3 +230,101 @@ def compute_flux_doubling(flux, flux_err, coords, axis=0):
...
+
+    Returns
+    -------
+    time_series: `~numpy.ndarray`
+        Simulated time series.
+
+    References
+    ----------
+    ..[Timmer1995]"On generating power law noise", J. Timmer and M, Konig, section 3.
+    """
```
registerrier, line 274: An example usage in the docstring would be very helpful.

gammapy/stats/variability.py
```diff
@@ -228,3 +230,101 @@ def compute_flux_doubling(flux, flux_err, coords, axis=0):
...
+    real_frequencies = np.sort(np.abs(frequencies[frequencies < 0]))
+
+    if function == "powerlaw":
+        periodogram = (1 / real_frequencies) ** power_spectrum
+    elif function == "white":
+        periodogram = np.full_like(real_frequencies, power_spectrum)
+    elif function == "custom":
+        periodogram = power_spectrum(real_frequencies)
+    else:
+        raise ValueError("Power spectrum function not accepted!")
```
registerrier, line 292: I think it is simpler/safer to accept only a callable and explain in the docstring that to generate a PL noise, one needs to pass `power_spectrum = lambda f : f**-index` or `power_spectrum = lambda f : sigma`

registerrier (commented):
Thanks @cgalelli . I have added some further cleanup comments to improve readability and documentation.
I think that the code can be simplified w.r.t. to power_spectrum. This could be only a `callable`, then the docstring can explicitly write how to define it for simple red/white noises.

cgalelli:
Thanks for the comments @registerrier ! In the latest version the PSD is passed only as a callable function and added examples to help the user.

gammapy/stats/variability.py
```diff
@@ -228,3 +230,117 @@ def compute_flux_doubling(flux, flux_err, coords, axis=0):
...
+    Parameters
+    ----------
+    power_spectrum : function
+        Power spectrum used to generate the time series. It is expected to be
+        a function mapping the input frequencies to the periodogram.
+    npoints : float
+        Number of points in the output time series.
+    spacing : '~astropy.Quantity'
+        Sample spacing, inverse of the sampling rate.
+    t_ref : '~astropy.Time', optional
```
AtreyeeS, line 254: In other places, we take a default `t_ref` if unspecified. https://github.com/gammapy/gammapy/blob/5774c346f72c035b969ecdfe726ede9b68aaf0bf/gammapy/utils/time.py#L21
Should we keep the same behaviour here?

AtreyeeS (dismissed):
Thank you @cgalelli ! no further comments

registerrier (commented):
Thanks @cgalelli . This looks good.

I agree with @AtreyeeS that the behavior with respect to `t_ref` is a bit strange. Adding an argument which changes the return type is a bit weird. So I would either suggest to always return a `Time` object, enforcing a default time value. Or to always return a time-like `Quantity` and let the user add the reference time. The latter is probably simplest.

gammapy/stats/tests/test_variability.py
```diff
@@ -143,3 +144,28 @@ def test_lightcurve_flux_doubling():
...
+        return norm * x ** (-index)
+
+    params = {"norm": 1.5, "index": 3}
+
+    time_series3, time_axis3 = TimmerKonig_lightcurve_simulator(
+        temp, 15, 1 * u.h, power_spectrum_params=params
+    )
+
+    assert len(time_series) == 20
+    assert isinstance(time_axis, u.Quantity) == True
```
registerrier, line 167:
```suggestion
    assert isinstance(time_axis, u.Quantity)
```

gammapy/stats/tests/test_variability.py
```diff
@@ -143,3 +144,28 @@ def test_lightcurve_flux_doubling():
...
+    assert time_axis.unit == u.s
+    assert len(time_series2) == 21
+    assert isinstance(time_axis2, Time) == True
```
registerrier, line 170:
```suggestion
    assert isinstance(time_axis2, Time)
```

cgalelli:
Thanks @AtreyeeS and @registerrier ! I remmoved the `t_ref` parameter as suggested by @registerrier , I agree that , between the two options presented, always returning a Quantity axis is the simplest one.

bkhelifi (approved):
Thanks a lot! This is a nice feature....
The dedicated tutorial will be important to expose this feature

QRemy (approved):
Thanks @cgalelli looks good, I just made some style simplifications and rebased.

gammapy/stats/variability.py
```diff
@@ -281,3 +283,105 @@ def structure_function(flux, flux_err, time, tdelta_precision=5):
...
+        [fourier_coeffs, np.conjugate(fourier_coeffs[idx0::-1])]
+    )
+
+    fourier_coeffs = np.insert(fourier_coeffs, 0, 0)
+    time_series = np.fft.ifft(fourier_coeffs).real
+
+    if time_series.max() < np.abs(time_series.min()):
+        time_series = -time_series
+
+    time_series = 0.5 + 0.5 * time_series / time_series.max()
```
bkhelifi, line 383: Is it not faster to make` 0.5 + 0.5 / time_series.max() * time_series` ? it avoids to access to the array twice...

QRemy: Not sure if it makes a significative difference but I added a commit to define tmax=time_series.max() so it is not computed twice.

QRemy (approved):

AtreyeeS (approved):
Thanks @cgalelli ! no further comments

PR #5146, opened 2024-03-04
Fix `geom.energy_mask` docstring

### LINKED ISSUE #5144: Incorrect return type for `Geom.energy_mask`
`Geom.energy_mask` returns a `Map` wit the same geometry as the parent `Geom`. This is incorrect in the currentl documentation which states the return type is `np.array`: https://github.com/gammapy/gammapy/blob/e5aecb334e7aebe304affd96b545a636019f7626/gammapy/maps/geom.py#L628

MRegeard:
This PR fix #5144.

### DIFF
--- a/gammapy/maps/geom.py
+++ b/gammapy/maps/geom.py
@@ -625,8 +625,9 @@ class Geom(abc.ABC):
 
         Returns
         -------
-        mask : `~numpy.ndarray`
-            Energy mask.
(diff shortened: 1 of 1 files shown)


adonath (approved):
Thanks @MRegeard!

adonath:
@MRegeard For these kind of small PRs we could actually skip the CI. You can do this by adding `"... [ci skip] "` to the commit message.

PR #5151, opened 2024-03-05
Change tuple to list for default value of methods in `SafeMaskMaker`

### LINKED ISSUE #5150: Defaults for `methods` in signature of `SafeMaskMaker.__init__` is confusing
```
    def __init__(
        self,
        methods=("aeff-default",),
        aeff_percent=10,
        bias_percent=10,
        position=None,
        fixed_offset=None,
        offset_max="3 deg",
        irfs="DL4",
    ):
```

In the signature of the `SafeMaskMaker`, the methods arguments is defaults to `("aeff-default",)` which is confusing because the coma is necessary for the code to work. If one don't put the coma while using a tuple, in the code latter, the instruction `set(methods)` will give back `{'-', 'a', 'd', 'e', 'f', 'l', 't', 'u'}`. To make it less confusing I think it would be good change the tuple to a list, for which `set(["aeff-default"])` give back `{'aeff-default'}`.

MRegeard:
This PR fix #5150.

### DIFF
--- a/gammapy/makers/safe.py
+++ b/gammapy/makers/safe.py
@@ -62,7 +62,7 @@ class SafeMaskMaker(Maker):
 
     def __init__(
         self,
-        methods=("aeff-default",),
+        methods=["aeff-default"],
         aeff_percent=10,
         bias_percent=10,
         position=None,


MRegeard:
@AtreyeeS Yes it will work, the only thing is that is a tuple is used with only one element it needs the coma, as currently in the code.

AtreyeeS (approved):
Thanks! No comments then

Astro-Kirsty (approved):
Thanks @MRegeard

PR #5156, opened 2024-03-07
Introduction of the structure function for variability studies

cgalelli:
This PR is a draft for the introduction of structure functions. The implementation is that of the discrete function presented in [Emmanoulopoulos et al.](https://academic.oup.com/mnras/article/404/2/931/968488). 
This method is useful for timing analyses, in particular studies on AGN flares.

As for different functions implemented for the time variability domain, the base implementation is in `gammapy/stats/variabilty.py`. A wrapper for a utility function for `FluxPoints` can be added following the same schema as for, e.g., the doubling/halving time.


AtreyeeS (dismissed):
Thank you @cgalelli ! I suppose the changes in `gammapy/estimators/utils.py` are not necessary as the  `compute_structure_function` can reside inside the light curve class once its ready

gammapy/estimators/utils.py
```diff
@@ -24,6 +29,9 @@
     "compute_lightcurve_fvar",
     "compute_lightcurve_fpp",
     "compute_lightcurve_doublingtime",
+    "find_peaks_in_flux_map",
```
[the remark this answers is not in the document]
cgalelli: Probably a merge error, had to rebase the branch

cgalelli:
Thanks @AtreyeeS , you are correct, since the lightcurve class will be in sooner or later, the wrapper in utils for this is not needed. I will move it in the next commits.

gammapy/stats/variability.py
```diff
@@ -228,3 +229,54 @@ def compute_flux_doubling(flux, flux_err, coords, axis=0):
...
+
+    Parameters
+    ----------
+    flux : `~astropy.units.Quantity`
+        The measured fluxes.
+    flux_err : `~astropy.units.Quantity`
+        The error on measured fluxes.
+    time : `~astropy.units.Quantity`
+        The time coordinates at which the fluxes are measured.
+    tdelta_precision : int
```
bkhelifi, line 245:
```suggestion
    tdelta_precision : int, optional
```

gammapy/stats/variability.py
```diff
@@ -228,3 +229,54 @@ def compute_flux_doubling(flux, flux_err, coords, axis=0):
...
+    dist_matrix = (time[np.newaxis, :] - time[:, np.newaxis]).round(
+        decimals=tdelta_precision
+    )
+    distances = np.unique(dist_matrix)
+    distances = distances[distances > 0]
+    shape = distances.shape + flux.shape[1:]
+    factor = np.zeros(shape)
+    norm = np.zeros(shape)
+
+    for i, distance in enumerate(distances[distances > 0]):
```
bkhelifi, line 269: the check of positive distances is already made l264

gammapy/stats/tests/test_variability.py
```diff
@@ -143,3 +144,51 @@ def test_lightcurve_flux_doubling():
...
+def test_structure_function():
+    flux = np.array(
+        [
+            [1e-11, 4e-12],
+            [3e-11, 2.5e-12],
+            [1e-11, 1e-12],
+            [0.8e-11, 0.8e-12],
+            [1e-11, 1e-12],
+        ]
+    )
```
bkhelifi, line 158: the flux units are missing.... the same for their errors

bkhelifi (changes requested):
Thanks a lot @cgalelli . I have left some comments inline

gammapy/stats/tests/test_variability.py
cgalelli: The units on flux and flux errors are missing in all tests of similar type. They are not necessary in a sense (all these functions could work on an dimensional normalization). Should this be addressed ina  separate PR?

cgalelli:
In the latest commits I tried to implement an analytical error computation for the structure function. In the end, also reading from the source [https://academic.oup.com/mnras/article/404/2/931/968488](url), it came clear to me that this is not doable without using monte-carlo simulations, which would be too heavy to implement for a similar utility function. It would be left to the user to run simulations if they need errorbars for single data points in the resulting structure function.

gammapy/stats/tests/test_variability.py
bkhelifi: If it works for both cases, I think that there is no need of an other PR... If it works with units;)

bkhelifi:
Indeed Claudio, like PSD, the SF can not be interpreted without making simulations. A notebook could help the users. And once it is done, a reference of this nb into the docstring of this (merged) function could be made

bkhelifi (approved):
Looks good for me, thanks @cgalelli

gammapy/stats/variability.py
```diff
@@ -228,3 +229,55 @@ def compute_flux_doubling(flux, flux_err, coords, axis=0):
...
+
+    Parameters
+    ----------
+    flux : `~astropy.units.Quantity`
+        The measured fluxes.
+    flux_err : `~astropy.units.Quantity`
+        The error on measured fluxes.
+    time : `~astropy.units.Quantity`
+        The time coordinates at which the fluxes are measured.
+    tdelta_precision : int, optional
```
AtreyeeS, line 245: Is `int` the correct option here, or should it be in units of time.
Of course, this conversion can also be done while calling this function

AtreyeeS (approved):
Thanks @cgalelli ! Good from my side

gammapy/stats/variability.py
cgalelli: Thanks @AtreyeeS! Yes, it is quite easy to do the conversion during the call. I decided to pass an int as it is the direct way it is utilized in the code, and I don't think the parameter will be modified much often by the user.

registerrier: @cgalelli . Does it really work if you specify the unit? What do you gain with the rounding? Can't one use the full precision?

cgalelli: This would not affect most cases in which the time bins are uniform, or most of the cases with very different inhomogeneous bins. Since the structure function is binned on the time distances between measurements, it can happen that in cases with inhomogeneous but similar spacing, you can have two sets of measures with distances, purely for example, 1.7845685 hours and 1.7845632 hours. The 'tdelta_precision' parameter I use is there to specify at which decimal to check if the distances are "the same". Maybe in some use cases a rough binning is fine, or for example it is necessary due to having few data points, so one can relax the check by lowering the precision. In other cases instead the user has plenty of measurements or needs precision.

PR #5160, opened 2024-03-08
Fix select_nested_models if there is no free parameters for the null hypothesis

QRemy:
If there is no free parameter the fit of the null hypothesis fails.
This PR fix it by returning directly the result.


Astro-Kirsty (approved):
Thanks @QRemy. Any confusion is now eliminated for the user

bkhelifi (approved):
Thanks a lot @QRemy

PR #5161, opened 2024-03-09
Flux maps combination

QRemy:
Create a new `FluxMaps` by combining a list of flux maps with the same geometry.
Under the gaussian error approximation the likelihood is given by the gaussian distribution. The product of gaussians is also a gaussian so can derive dnde, dnde_err, and ts for the combined map.

### DIFF
--- a/gammapy/estimators/map/tests/test_core.py
+++ b/gammapy/estimators/map/tests/test_core.py
@@ -9,7 +9,7 @@ from astropy.io import fits
 from astropy.time import Time
 from gammapy.data import GTI
 from gammapy.estimators import FluxMaps
-from gammapy.estimators.utils import joint_flux_maps
+from gammapy.estimators.utils import combine_flux_maps
 from gammapy.maps import MapAxis, Maps, RegionGeom, TimeMapAxis, WcsNDMap
 from gammapy.modeling.models import (
     LogParabolaSpectralModel,
@@ -212,7 +212,7 @@ def test_map_properties(map_flux_estimate):
     assert_allclose(fe.eflux_ul.quantity.value[:, 2, 2], [4.60517e-10, 4.60517e-10])
 
 
-def test_joint_flux_maps(map_flux_estimate, wcs_flux_map, reference_model):
+def test_combine_flux_maps(map_flux_estimate, wcs_flux_map, reference_model):
     start = u.Quantity([1, 2], "min")
     stop = u.Quantity([1.5, 2.5], "min")
     gti = GTI.create(start, stop)
@@ -222,13 +222,13 @@ def test_joint_flux_maps(map_flux_estimate, wcs_flux_map, reference_model):
     iE = 0
     energy = fe.geom.axes[0].center[iE]
 
-    fe_new = joint_flux_maps([fe, fe])
+    fe_new = combine_flux_maps([fe, fe])
     assert_allclose(fe_new.dnde.quantity, fe.dnde.quantity)
     assert_allclose(
         fe_new.dnde_err.quantity.value, fe.dnde_err.quantity.value / np.sqrt(2)
     )
 
-    fe_new = joint_flux_maps(
+    fe_new = combine_flux_maps(
         [fe, fe], reference_model=FluxMaps.reference_model_default.spectral_model
     )
     assert_allclose(fe_new.dnde.quantity, fe.dnde.quantity)
@@ -246,7 +246,7 @@ def test_joint_flux_maps(map_flux_estimate, wcs_flux_map, reference_model):
     assert fe_new.meta == fe.meta
 
     fe = FluxMaps(wcs_flux_map, reference_model)
-    fe_new = joint_flux_maps([fe, fe], reference_model=reference_model)
+    fe_new = combine_flux_maps([fe, fe], reference_model=reference_model)
     assert_allclose(fe_new.dnde.quantity, fe.dnde.quantity)
     assert_allclose(
         fe_new.dnde_err.quantity.value, fe.dnde_err.quantity.value / np.sqrt(2)
@@ -259,7 +259,7 @@ def test_joint_flux_maps(map_flux_estimate, wcs_flux_map, reference_model):
     )
     assert_allclose(fe_new.ts, ts * 2)
 
-    fe_new = joint_flux_maps(fe)
+    fe_new = combine_flux_maps(fe)
 
 
 def test_flux_map_properties(wcs_flux_map, reference_model):
--- a/gammapy/estimators/map/tests/test_core.py
+++ b/gammapy/estimators/map/tests/test_core.py
@@ -213,8 +213,12 @@ def test_map_properties(map_flux_estimate):
 
 
 def test_joint_flux_maps(map_flux_estimate, wcs_flux_map, reference_model):
+    start = u.Quantity([1, 2], "min")
+    stop = u.Quantity([1.5, 2.5], "min")
+    gti = GTI.create(start, stop)
+
     model = SkyModel(PowerLawSpectralModel(amplitude="1e-10 cm-2s-1TeV-1", index=2))
-    fe = FluxMaps(data=map_flux_estimate, reference_model=model)
+    fe = FluxMaps(data=map_flux_estimate, reference_model=model, gti=gti)
     iE = 0
     energy = fe.geom.axes[0].center[iE]
 
@@ -250,6 +254,8 @@ def test_joint_flux_maps(map_flux_estimate, wcs_flux_map, reference_model):
         / stats.norm.pdf(fe.dnde.data, loc=fe.dnde.data, scale=fe.dnde_err.data)
     )
     assert_allclose(fe_new.ts, ts * 2)
+
+    fe_new = joint_flux_maps(fe)
 
 
 def test_flux_map_properties(wcs_flux_map, reference_model):
--- a/gammapy/estimators/map/tests/test_core.py
+++ b/gammapy/estimators/map/tests/test_core.py
@@ -2,6 +2,7 @@
 import pytest
 import numpy as np
 from numpy.testing import assert_allclose
+from scipy import stats
 import astropy.units as u
 from astropy.coordinates import SkyCoord
 from astropy.io import fits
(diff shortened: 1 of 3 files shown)


bkhelifi:
Hello @QRemy,
Thanks a lot for this nice utility function. I have two questions:
- maybe one could add a check and a test of maps with different geometries. This might happen frequently
- I am wondering whether one could not extend this function for non-gaussian errors, when the scan profile is present. Maybe one could make that later into an other PR, and add a ToDo in this one....

QRemy:
> * I am wondering whether one could not extend this function for non-gaussian errors, when the scan profile is present. Maybe one could make that later into an other PR, and add a ToDo in this one....

Yes I had that in mind but first we need to implement the stat_scan option ont the ExcessMapEstimator and TSMapEstimator.

registerrier:
From the dev call, let's try to find a more explicit name for the function. It could be possibly an independent function rather than a classmethod.

registerrier (commented):
Thanks @QRemy. This is nice. I have left some inline comments.

A general question is whether applying this to a regular `FluxMaps` is safe or not. What use cases does this cover?

gammapy/estimators/utils.py
```diff
@@ -686,3 +687,68 @@ def get_rebinned_axis(fluxpoint, axis_name="energy", method=None, **kwargs):
...
     return axis_new
+
+
+def joint_flux_maps(maps, method="gaussian_errors", **kwargs):
+    """Create a FluxMaps by combining a list of flux maps with the same geometry.
+     This assumes the flux maps are independent measurements of the same true value.
+
+    Parameters
+    ----------
+    maps : list of `~gammapy.estimators.FluxMaps` or ~gammapy.estimators.FluxMaps`
```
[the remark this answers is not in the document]
QRemy: to combine along the axes so it will give a weighted normalisation

gammapy/estimators/utils.py
```diff
@@ -686,3 +687,68 @@ def get_rebinned_axis(fluxpoint, axis_name="energy", method=None, **kwargs):
...
+        List of maps with the same geometry.
+    kwargs : dict
```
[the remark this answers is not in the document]
QRemy: now I stack the GTI but `FluxMapsMetaData.stack`  and `FluxPointsMetaData.stack` don't not exit yet right ? so we keep it for a follow up PR.

gammapy/estimators/utils.py
```diff
@@ -686,3 +687,68 @@ def get_rebinned_axis(fluxpoint, axis_name="energy", method=None, **kwargs):
         edges = np.append(edges_min, edges_max[-1])
         axis_new = MapAxis.from_edges(edges, name=axis_name, interp=ax.interp)
     return axis_new
+
+
+def joint_flux_maps(maps, method="gaussian_errors", **kwargs):
```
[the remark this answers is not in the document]
QRemy: tried `.to_maps(sed_type)` but got an error. I would keep that for a separate PR if possible.

gammapy/estimators/utils.py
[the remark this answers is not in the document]
QRemy: I don't want to advertise this but it would be nice to have this options for testing the method

registerrier:
There is a merge conflict here. Can you rebase @QRemy ?

gammapy/estimators/utils.py
```diff
@@ -753,3 +754,79 @@ def get_combined_significance_maps(estimator, datasets):
...
+        mean.data[mask_k] = mean_k[mask_k]
+        sigma.data[mask_k] = sigma_k[mask_k]
+
+    ts = -2 * np.log(
+        stats.norm.pdf(0, loc=mean, scale=sigma)
+        / stats.norm.pdf(mean, loc=mean, scale=sigma)
+    )
+
+    kwargs["sed_type"] = "dnde"
+    return FluxMaps.from_maps(dict(dnde=mean, dnde_err=sigma, ts=ts), gti=gti, **kwargs)
```
AtreyeeS, line 832: The `dict` should contain Maps, no? Right now it contains numpy arrays

QRemy: I added a line to redefine ts as a map, the others are already maps

gammapy/estimators/utils.py
```diff
@@ -753,3 +754,80 @@ def get_combined_significance_maps(estimator, datasets):
...
+def joint_flux_maps(maps, method="gaussian_errors", **kwargs):
+    """Create a FluxMaps by combining a list of flux maps with the same geometry.
+     This assumes the flux maps are independent measurements of the same true value.
+
+    Parameters
+    ----------
+    maps : list of `~gammapy.estimators.FluxMaps` or ~gammapy.estimators.FluxMaps`
+        List of maps with the same geometry.
+    kwargs : dict
+        Options passed to `~gammapy.estimators.FluxMaps.from_maps`
```
AtreyeeS, line 768: Maybe the `kwargs` here is a bit misleading because one can't pass a `gti`, and the `sed_type` will always be overwritten. 
Does one need to pass a reference model separately? The individual `FluxMaps` should have a `reference_model` contained. 
What is the new reference model? Does this only work for the same `reference_model` for all maps?

So what is passed through `kwargs` is mainly the meta?

AtreyeeS (approved):
Thanks @QRemy ! I have some questions inline, but it is okay from my side

gammapy/estimators/utils.py
```diff
@@ -753,3 +754,80 @@ def get_combined_significance_maps(estimator, datasets):
         npred_excess=npred_excess_sum,
         estimator_results=results,
     )
+
+
+def joint_flux_maps(maps, method="gaussian_errors", **kwargs):
+    """Create a FluxMaps by combining a list of flux maps with the same geometry.
+     This assumes the flux maps are independent measurements of the same true value.
```
QRemy, line 761:
```suggestion
     This assumes the flux maps are independent measurements of the same true value.
     The GTI is stacked in the process.
```

gammapy/estimators/utils.py
```diff
@@ -753,3 +754,80 @@ def get_combined_significance_maps(estimator, datasets):
...
+    -------
+    flux_maps : `~gammapy.estimators.FluxMaps`
+        Joint flux map.
+    """
+    if method != "gaussian_errors":
+        raise ValueError(
+            f'Invalid method : {method}, available methods are : "gaussian_errors"'
+        )
+
+    if isinstance(maps, FluxMaps):
```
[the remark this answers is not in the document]
QRemy: I modified the line in the docstring that mentions this a a possible option in order to keep it as a hidden feature.
I will do some test to check in which case it makes sense to use it.

gammapy/estimators/utils.py
QRemy: Now the `reference_model` is added to the arguments list instead of `kwargs` (there is no issues using a different reference model).
The meta is added taking the info from the parent FluxMaps (in a very simple way) but we will have to come back to this later when `FluxMapsMetaData.stack` and `FluxPointsMetaData.stack` are implemented.

AtreyeeS: Yes, we can keep the meta for later. It currently stores the `n_sigma_ul` which should be the same between the different maps I think.

AtreyeeS (approved):
Thanks @QRemy for the updates! No further comments

gammapy/estimators/map/tests/test_core.py
```diff
@@ -210,6 +212,56 @@ def test_map_properties(map_flux_estimate):
...
+    start = u.Quantity([1, 2], "min")
+    stop = u.Quantity([1.5, 2.5], "min")
+    gti = GTI.create(start, stop)
+
+    model = SkyModel(PowerLawSpectralModel(amplitude="1e-10 cm-2s-1TeV-1", index=2))
+    fe = FluxMaps(data=map_flux_estimate, reference_model=model, gti=gti)
+    iE = 0
+    energy = fe.geom.axes[0].center[iE]
+
+    fe_new = joint_flux_maps([fe, fe])
```
registerrier, line 225:
```suggestion
    fe_new = average_flux_maps([fe, fe])
    
```

gammapy/estimators/utils.py
```diff
@@ -753,3 +754,95 @@ def get_combined_significance_maps(estimator, datasets):
         npred_excess=npred_excess_sum,
         estimator_results=results,
     )
+
+
+def joint_flux_maps(maps, method="gaussian_errors", reference_model=None):
```
registerrier, line 759:
```suggestion
def average_flux_maps(maps, method="gaussian_errors", reference_model=None):
```

gammapy/estimators/map/tests/test_core.py
```diff
@@ -210,6 +212,56 @@ def test_map_properties(map_flux_estimate):
     assert_allclose(fe.eflux_ul.quantity.value[:, 2, 2], [4.60517e-10, 4.60517e-10])
 
 
+def test_joint_flux_maps(map_flux_estimate, wcs_flux_map, reference_model):
```
registerrier, line 215:
```suggestion
def test_average_flux_maps(map_flux_estimate, wcs_flux_map, reference_model):
```

gammapy/estimators/map/tests/test_core.py
```diff
@@ -210,6 +212,56 @@ def test_map_properties(map_flux_estimate):
...
+    iE = 0
+    energy = fe.geom.axes[0].center[iE]
+
+    fe_new = joint_flux_maps([fe, fe])
+    assert_allclose(fe_new.dnde.quantity, fe.dnde.quantity)
+    assert_allclose(
+        fe_new.dnde_err.quantity.value, fe.dnde_err.quantity.value / np.sqrt(2)
+    )
+
+    fe_new = joint_flux_maps(
```
registerrier, line 231:
```suggestion
    fe_new = average_flux_maps(
```

gammapy/estimators/map/tests/test_core.py
```diff
@@ -210,6 +212,56 @@ def test_map_properties(map_flux_estimate):
...
+        energy
+    ) / FluxMaps.reference_model_default.spectral_model(energy)
+
+    assert_allclose(
+        fe_new.norm.quantity.value[iE, :, :] / fe.norm.quantity.value[iE, :, :], ratio
+    )
+    assert fe_new.meta == fe.meta
+
+    fe = FluxMaps(wcs_flux_map, reference_model)
+    fe_new = joint_flux_maps([fe, fe], reference_model=reference_model)
```
registerrier, line 249:
```suggestion
    fe_new = average_flux_maps([fe, fe], reference_model=reference_model)
```

registerrier (approved):

I simply propose to call the function `average_flux_maps` as this is more explicit. Joint flux map would rather be applied to flux maps extracted jointly from a group of datasets.Actually maybe `stack` is even more clear as this is consistent with typical gammapy semantics.

QRemy:
> I simply propose to call the function `average_flux_maps` as this is more explicit. Joint flux map would rather be applied to flux maps extracted jointly from a group of datasets.

Maybe `combine_flux_maps` as we have `get_combined_significance_maps`. I think the term combined is often used when combining likehoods from different experiment a posteriori, especially in cosmology like in  https://arxiv.org/pdf/1807.06209 .

registerrier (approved):
Thanks @QRemy. Looks good!

PR #5167, opened 2024-03-11
Fix SourceCatalogObject4FGL.to_models() for DR4

QRemy:
Fix templates path.
Fix output map dtype for templates.


MRegeard (dismissed):
Thanks @QRemy ! Seems good now.

MRegeard (approved):
Thanks @QRemy. The test failing is not related right ?

QRemy:
> Thanks @QRemy. The test failing is not related right ?

yes test fails is unrelated

PR #5169, opened 2024-03-13
Add boundary condition to create a PeriodicMapAxis

AtreyeeS:
This PR adds a PeriodicMapAxis.
Its rather simple, only one function, so arguably it does not need its own class at all.
But I think its conceptually simpler to have a separate name for such an axis.
It can be useful for phase analysis, or for asymmetric IRFs defined not in (Lon, lat) but (offset, theta), as brought up in one of the CTAC ASWG calls...

### DIFF
--- a/gammapy/maps/axes.py
+++ b/gammapy/maps/axes.py
@@ -3427,3 +3427,49 @@ class LabelMapAxis:
         return LabelMapAxis(
             labels=[self.center[0] + "..." + self.center[-1]], name=self._name
         )
+
+
+class PeriodicMapAxis(MapAxis):
+    """
+    A periodic MapAxis. Only linear interpolation is supported.
+
+    Parameters
+    ----------
+    nodes : `~numpy.ndarray` or `~astropy.units.Quantity`
+        Array of node values.  These will be interpreted as either bin
+        edges or centers according to ``node_type``.
+    name : str, optional
+        Axis name. Default is "".
+    node_type : str, optional
+        Flag indicating whether coordinate nodes correspond to pixel
+        edges (node_type = 'edges') or pixel centers (node_type =
+        'center').  'center' should be used where the map values are
+        defined at a specific coordinate (e.g. differential
+        quantities). 'edges' should be used where map values are
+        defined by an integral over coordinate intervals (e.g. a
+        counts histogram). Default is "edges".
+    unit : str, optional
+        String specifying the data units. Default is "".
+    """
+
+    def wrap_coords(self, coords):
+        """Wrap coords between axis edges"""
+        m1, m2 = self.edges_min[0], self.edges_max[-1]
+        coords_copy = copy.deepcopy(coords)
+        coords_copy[coords_copy >= m2] = (coords_copy[coords_copy >= m2] - m1) % (
+            m2 - m1
+        ) + m1
+        coords_copy[coords_copy < m1] = (coords_copy[coords_copy < m1] - m1) % (
+            m2 - m1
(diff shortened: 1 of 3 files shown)


MRegeard (approved):
Thanks @AtreyeeS, this is a great feature

MRegeard:
On top of that, an idea that we could add following this feature is a method on maps to offset the data that are indexed through a periodic axis. This is usual in pulsar analysis to put the peak of the pulsar that would be around 0 (or 1) at the center of the phasogram.

gammapy/maps/tests/test_axes.py
```diff
@@ -962,3 +969,15 @@ def test_energy_bin_per_decade_not_strict_bounds():
     )
 
     assert_allclose(axis.edges[0:-nbin] * 10.0, axis.edges[nbin:], rtol=1e-5, atol=0)
+
+
+def test_phase_map_axis():
+    axis1 = PeriodicMapAxis.from_bounds(-0.5, 0.5, 5)
+    coords = np.array([-1.0, -0.7, 0.1, 1.2])
+    assert_allclose(axis1.wrap_coords(coords), [0.0, 0.3, 0.1, 0.2], rtol=1e-5)
```
bkhelifi, line 977: A naive question.. the nodes of the final axis are not crescent?

AtreyeeS:
I created an option on `MapAxis` as `boundary_type` to specify the boundary condition.
This still assumes that the edges wrap, or should that be configurable?

gammapy/maps/axes.py
```diff
@@ -803,7 +836,7 @@ def pix_to_idx(self, pix, clip=False):
 
         return idx
 
-    def coord_to_pix(self, coord):
+    def _coord_to_pix(self, coord):
```
QRemy, line 839: I don't think that the private definition is necessary here

gammapy/maps/axes.py
```diff
@@ -778,6 +799,18 @@ def pix_to_coord(self, pix):
         values = self._transform.pix_to_coord(pix=pix)
         return u.Quantity(values, unit=self.unit, copy=False)
 
+    def _wrap_coords(self, coords):
```
QRemy, line 802: could be public

gammapy/maps/axes.py
```diff
@@ -820,7 +853,7 @@ def coord_to_pix(self, coord):
         pix = self._transform.coord_to_pix(coord=coord)
         return np.array(pix + self._pix_offset, ndmin=1)
 
-    def coord_to_idx(self, coord, clip=False):
+    def _coord_to_idx(self, coord, clip=False):
```
QRemy, line 856: same

QRemy (commented):
Thanks @AtreyeeS,  looks good, I made some change to remove the private functions.

adonath (changes requested):
Thanks @AtreyeeS and @QRemy! I have left some minor comment to clean up the implementation...

QRemy:
Thanks @adonath for the suggestions I implemented them and I will merge this once the CI pass if no objection.

gammapy/maps/axes.py
```diff
@@ -122,11 +141,18 @@ def __init__(self, nodes, interp="lin", name="", node_type="edges", unit=""):
         else:
             nodes = np.array(nodes)
 
+        available_boundary = {"periodic", "monotonic"}
+        if boundary_type not in available_boundary:
```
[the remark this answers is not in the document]
QRemy: added

PR #5174, opened 2024-03-19
Introduction of a new specialized class for lightcurves

cgalelli:
In this draft PR I introduce the basic structure for a specialized `LightCurve` class which inherits from `FluxPoints`.
I tried to keep the code duplication as minimal as possible.

The first implementation introduces read/write and from/to_table method adaptations from `FluxPoints`, and simple methods for variability tools. It also includes the translator for `stingray` (as an optional inclusion) previously presented in #5170 . 

The class could also use utilities proposed in #5145 and #5156 once they are added, and any future specific utilities.


gammapy/estimators/points/core.py
```diff
@@ -336,8 +341,15 @@ def from_table(
         """
         table = table_standardise_units_copy(table)
 
-        if format is None:
-            format = cls._table_guess_format(table)
+        guessed_format = cls._table_guess_format(table)
+
+        if format is not None and format != guessed_format:
```
AtreyeeS, line 346: Is this necessary? The user can have the liberty to override the guessed format.

gammapy/estimators/points/core.py
```diff
@@ -367,13 +379,24 @@ def from_table(
...
-        return cls.from_maps(
-            maps=maps,
-            reference_model=reference_model,
-            meta=meta,
-            sed_type=sed_type,
-            gti=gti,
-        )
+
+        if format == "lightcurve":
+            return LightCurve.from_maps(
```
AtreyeeS, line 384: Is this the responsibility of the parent class?

gammapy/estimators/points/core.py
```diff
@@ -850,3 +873,79 @@ def resample_axis(self, axis_new):
...
+        Only objects with a single energy bin are allowed.
+        """
+
+        if self.geom.axes["energy"].nbin > 1:
+            raise ValueError("Only lightcurves with 1 energy bin are supported.")
+
+        try:
+            from stingray import Lightcurve
+
+            counts = self.counts.data
```
AtreyeeS, line 908: I am still uncomfortable with exporting just the counts as a reliable unit.
Since the `Lighcurve` object contains a spectral model, maybe we can scale the flux by the model and use some "scaled counts"?

gammapy/estimators/points/core.py
```diff
@@ -850,3 +873,79 @@ def resample_axis(self, axis_new):
...
+                "Requested format is not supported as output for LightCurve"
+            )
+        kwargs.setdefault("format", format)
+        return super().to_table(**kwargs)
+
+    def fvar(self, flux_quantity="flux"):
+        return compute_lightcurve_fvar(self, flux_quantity)
+
+    def fpp(self, flux_quantity="flux"):
+        return compute_lightcurve_fpp(self, flux_quantity)
```
AtreyeeS, line 948: Do these functions work correctly for multi-dimensional light curves?
Otherwise, the class can ensure that it has only one energy bin and no additional axes.

AtreyeeS (commented):
Thank you @cgalelli ! I think this can go in and then we work on the improvements
I would add a customised `.plot`

gammapy/estimators/points/core.py
cgalelli: Yes these functions return tables which can accomodate multi-dimensional objects.

gammapy/estimators/points/core.py
cgalelli: This is a good idea, I should probably investigate stingray to see what would be the best alternative here.

gammapy/estimators/points/core.py
cgalelli: The idea here was not only to give "responsability" to the parent, but to create a Lightcurve "by force" when using FluxPoints with a time axis (e.g. to keep compatibility with old code but still access the new functionalities and checks)

gammapy/estimators/points/core.py
cgalelli: Yes perhaps this check is too harsh

gammapy/estimators/points/core.py
```diff
@@ -850,3 +865,90 @@ def resample_axis(self, axis_new):
...
+        table = table_standardise_units_copy(table)
+
+        if super()._table_guess_format(table) != "lightcurve":
+            raise ValueError("The input table is not a lightcurve!")
+
+        kwargs.setdefault("format", "lightcurve")
+
+        return super().from_table(table, **kwargs)
+
+    def to_stingray_lightcurve(self):
```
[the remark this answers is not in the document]
cgalelli: I can add a translator function on that direction as well. I would still be wary of actually using the stingray functions with flux (and they say so in the paragraph), but if they themselves show an example it means that it's not unheard of, just not exactly the standard approach. I can put a disclaimer in the docstring for both to and `from-stingray`

gammapy/estimators/points/core.py
```diff
@@ -849,3 +865,98 @@ def resample_axis(self, axis_new):
             fluxpoints.append(fp_new)
 
         return self.__class__.from_stack(fluxpoints, axis=axis_new)
+
+
+class LightCurve(FluxPoints):
+    """Specialized 'FluxPoints' heir class for lightcurves"""
```
bkhelifi, line 871: To be developed: description, exemple, parameters and attributes, ....

gammapy/estimators/points/core.py
```diff
@@ -849,3 +865,98 @@ def resample_axis(self, axis_new):
...
+    def __init__(
+        self, data, reference_model, meta=None, gti=None, filter_success_nan=True
+    ):
+        if "time" not in data["norm"].geom.axes_names:
+            raise ValueError("Trying to build a lightcurve without time axis!")
+        super().__init__(data, reference_model, meta, gti, filter_success_nan)
+
+    @classmethod
+    def read(cls, filename, **kwargs):
+        kwargs.setdefault("format", "lightcurve")
```
bkhelifi, line 882: One would need to add a metadata class, as for the flux class, or re-used this metadata class

gammapy/estimators/points/core.py
```diff
@@ -849,3 +865,98 @@ def resample_axis(self, axis_new):
...
+                "Requested format is not supported as output for LightCurve"
+            )
+        kwargs.setdefault("format", format)
+        return super().to_table(**kwargs)
+
+    def plot(self, **kwargs):
+        kwargs.setdefault("axis_name", "time")
+        return super().plot(**kwargs)
+
+    def fvar(self, flux_quantity="flux"):
```
bkhelifi, line 947: docstring to add

gammapy/estimators/points/core.py
```diff
@@ -849,3 +865,98 @@ def resample_axis(self, axis_new):
...
+        return compute_lightcurve_fvar(self, flux_quantity)
+
+    def fpp(self, flux_quantity="flux"):
```
bkhelifi, line 950: docstring to add

gammapy/estimators/points/core.py
```diff
@@ -849,3 +865,98 @@ def resample_axis(self, axis_new):
...
+        return compute_lightcurve_fpp(self, flux_quantity)
+
+    def doubling_time(self, flux_quantity="flux"):
+        return compute_lightcurve_doublingtime(self, flux_quantity)
```
bkhelifi, line 954: docstring to add

gammapy/estimators/points/core.py
```diff
@@ -849,3 +865,98 @@ def resample_axis(self, axis_new):
...
+
+    def structure_function(self, flux_quantity="flux", tdelta_precision=5):
```
bkhelifi, line 956: docstring to add

gammapy/estimators/points/core.py
```diff
@@ -849,3 +865,98 @@ def resample_axis(self, axis_new):
             fluxpoints.append(fp_new)
 
         return self.__class__.from_stack(fluxpoints, axis=axis_new)
+
+
+class LightCurve(FluxPoints):
```
bkhelifi, line 870: Add all the test functions

bkhelifi (commented):
Thanks @cgalelli ... I have put some comments in order to make a step forward

gammapy/estimators/points/core.py
```diff
@@ -849,3 +865,98 @@ def resample_axis(self, axis_new):
...
+
+        return super().from_table(table, **kwargs)
+
+    def to_stingray_lightcurve(self):
+        """Transform to a 'stingray.Lightcurve' object.
+        Only objects with a single energy bin are allowed.
+        """
+
+        if self.geom.axes["energy"].nbin > 1:
+            raise ValueError("Only lightcurves with 1 energy bin are supported.")
```
VascoSch92, line 902:
```suggestion
            raise ValueError(f"Only lightcurves with 1 energy bin are supported, but got {self.geom.axes["energy"].nbin} energy bins.")
```

gammapy/estimators/points/core.py
```diff
@@ -849,3 +865,98 @@ def resample_axis(self, axis_new):
...
+        except ImportError:
+            raise ImportError("stingray is not installed in the environment.")
+
+    def write(self, filename, **kwargs):
+        format = kwargs.pop("format", None)
+        if format is None:
+            format = "lightcurve"
+        elif format not in ["lightcurve", "binned-time-series"]:
+            raise ValueError(
+                "Requested format is not supported as output for LightCurve"
```
VascoSch92, line 927: you can list the supported format in the error message

bkhelifi:
@cgalelli are the CI errors reproducible on your machine? If yes, can you fix the code?

registerrier:
We need to take a decision here. What is the status? For now, let's postpone for 2.0.

registerrier:
Conversion to stingray has proven to be not feasible. We are also considering restructuring plotting code. This leaves not so much inside the LightCurve object. Let's close it for now and revisit if clear case can be made for a specific class.

PR #5176, opened 2024-03-22
Fix `observatory_earth_location`

StFroese:
If an Observation is created by the DataStore.obs method, no location is given as an argument. Therefore the `observatory_earth_location` will be None.

Simple fix of catching the case by reading the location from the meta data.

Maybe the location should always be given by the meta data and not at the __init__ but I wasn't sure and wanted to keep it compatible with older versions

### DIFF
--- a/gammapy/data/observations.py
+++ b/gammapy/data/observations.py
@@ -378,6 +378,8 @@ class Observation:
     @property
     def observatory_earth_location(self):
         """Observatory location as an `~astropy.coordinates.EarthLocation` object."""
+        if self._location is None:
+            return self.meta.location
         return self._location
 
     @lazyproperty


gammapy/data/observations.py
```diff
@@ -554,7 +557,6 @@ def write(
             Path for the output file.
         overwrite : bool, optional
             Overwrite existing file. Default is False.
-        format : {"gadf"}
             Output format, currently only "gadf" is supported. Default is "gadf".
```
bkhelifi, line 560: Maybe one can put this line at the start of the docstring, instead than in the middle...

StFroese: Oh I actually did not want to remove it😅
Was a mistake

registerrier:
Please @StFroese can you check that black has been applied?

registerrier (approved):
Thanks @StFroese . Looks good.

Issue #5179, opened 2024-03-25
WCSNDMap reorders input data


maxnoe:
**Gammapy version**

Current main branch and also released versions.

**Bug description**

The `WCSNDMap` silently reorders the array data by calling `data.reshape(geom.data_shape)` if the input data shape does not match the expected shape.

This is most certainly wrong, as you cannot just reshape the same memory into the desired shape and assume the values are correct.

E.g. in the following example, the correct thing would probably have been a transpose operation (I assume WCSGeom assumes FITS / Fortran shape order), not the reshaping operation.

```python
In [1]: from gammapy.maps import WcsNDMap, WcsGeom
   ...: import numpy as np
   ...: from astropy.coordinates import SkyCoord

In [2]: data = np.arange(15).reshape(5, 3)

In [3]: center = SkyCoord.from_name("Crab Nebula")

In [4]: geom = WcsGeom.create(
   ...:     npix=(5, 3),
   ...:     binsz=0.5,
   ...:     refpix=(2, 1),
   ...:     skydir=center,
   ...: )

In [5]: wcsmap = WcsNDMap(geom, data)

In [6]: wcsmap.data
Out[6]: 
array([[ 0,  1,  2,  3,  4],
       [ 5,  6,  7,  8,  9],
       [10, 11, 12, 13, 14]])

In [7]: data
Out[7]: 
array([[ 0,  1,  2],
       [ 3,  4,  5],
       [ 6,  7,  8],
       [ 9, 10, 11],
       [12, 13, 14]])

In [8]: wcsmap.data.shape == data.T.shape
Out[8]: True

In [9]: wcsmap.data == data.T
Out[9]: 
array([[ True, False, False, False, False],
       [False, False,  True, False, False],
       [False, False, False, False,  True]])

```

**Expected behavior**

An error in case the data shape does not match instead of silently corrupting the input data.

**To Reproduce**

See example above.

**Other information**

CC @StFroese @LukasNickel

registerrier:
This is solved since #5180 . Closing.

PR #5180, opened 2024-03-25
Raise error on invalid input shape in Map

maxnoe:
**Description**

Connected to #5179 I wanted to see what fails if instead of silently converting the input data into the expected shape, an error is raised.

It seems most cases where this fails is just tests "being lazy" with toy data not in the correct shape and relying on this reshaping, not any actual data files being loaded that require this treatment.

### DIFF
--- a/gammapy/maps/core.py
+++ b/gammapy/maps/core.py
@@ -110,10 +110,10 @@ class Map(abc.ABC):
         if not value.shape == self.geom.data_shape:
             try:
                 value = np.broadcast_to(value, self.geom.data_shape, subok=True)
-            except Exception:
+            except ValueError as exc:
                 raise ValueError(
                     f"Input shape {value.shape} is not compatible with shape from geometry {self.geom.data_shape}"
-                ) from None
+                ) from exc
 
         self._data = value
 
--- a/gammapy/maps/core.py
+++ b/gammapy/maps/core.py
@@ -1735,7 +1735,7 @@ class Map(abc.ABC):
             data.append(m.quantity.to_value(maps[0].unit))
 
         return cls.from_geom(
-            data=np.stack(data), geom=geom.to_cube(axes=[axis]), unit=maps[0].unit
+            data=np.concatenate(data), geom=geom.to_cube(axes=[axis]), unit=maps[0].unit
         )
 
     def split_by_axis(self, axis_name):
--- a/gammapy/maps/core.py
+++ b/gammapy/maps/core.py
@@ -108,7 +108,9 @@ class Map(abc.ABC):
             raise TypeError("Map data must be a Numpy array. Set unit separately")
 
         if not value.shape == self.geom.data_shape:
-            value = value.reshape(self.geom.data_shape)
+            raise ValueError(
+                f"Input shape {value.shape} does not match expected shape from geometry {self.geom.data_shape}"
+            )
 
         self._data = value
 
--- a/gammapy/maps/core.py
+++ b/gammapy/maps/core.py
@@ -1734,9 +1734,10 @@ class Map(abc.ABC):
 
             data.append(m.quantity.to_value(maps[0].unit))
 
-        return cls.from_geom(
-            data=np.stack(data), geom=geom.to_cube(axes=[axis]), unit=maps[0].unit
-        )
+        new_geom = geom.to_cube(axes=[axis])
+        data = np.concatenate(data).reshape(new_geom.data_shape)
+
+        return cls.from_geom(data=data, geom=new_geom, unit=maps[0].unit)
 
     def split_by_axis(self, axis_name):
         """Split a Map along an axis into multiple maps.
--- a/gammapy/maps/core.py
+++ b/gammapy/maps/core.py
@@ -1735,7 +1735,7 @@ class Map(abc.ABC):
             data.append(m.quantity.to_value(maps[0].unit))
 
         return cls.from_geom(
-            data=np.concatenate(data), geom=geom.to_cube(axes=[axis]), unit=maps[0].unit
+            data=np.stack(data), geom=geom.to_cube(axes=[axis]), unit=maps[0].unit
         )
 
     def split_by_axis(self, axis_name):
--- a/gammapy/maps/core.py
+++ b/gammapy/maps/core.py
@@ -108,9 +108,12 @@ class Map(abc.ABC):
             raise TypeError("Map data must be a Numpy array. Set unit separately")
 
         if not value.shape == self.geom.data_shape:
-            raise ValueError(
-                f"Input shape {value.shape} does not match expected shape from geometry {self.geom.data_shape}"
-            )
+            try:
+                value = np.broadcast_to(value, self.geom.data_shape, subok=True)
+            except Exception:
+                raise ValueError(
+                    f"Input shape {value.shape} is not compatible with shape from geometry {self.geom.data_shape}"
+                ) from None
 
         self._data = value
 
--- a/gammapy/astro/darkmatter/spectra.py
+++ b/gammapy/astro/darkmatter/spectra.py
@@ -3,7 +3,7 @@
 import numpy as np
 import astropy.units as u
 from astropy.table import Table
-from gammapy.maps import MapAxis, RegionNDMap
+from gammapy.maps import Map, MapAxis, RegionGeom
 from gammapy.modeling import Parameter
 from gammapy.modeling.models import SpectralModel, TemplateNDSpectralModel
 from gammapy.utils.scripts import make_path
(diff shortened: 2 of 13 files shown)


AtreyeeS:
Thanks @maxnoe ! I agree it is rather strange to silently re-order the data (and a relief it does not affect any major results!) I think it might still be useful to allow for broadcastable shapes instead of the hard equality you introduced.

adonath:
Thanks @maxnoe, I agree this is the safer behavior! I'm not sure when this was introduced, but it must have been there for many years already. This tells us at least that it did not lead to frequent problems and people possibly relied on it without issues (we don't know...). So I am rather inclined to formally deprecate this behavior instead. On the other hand it seems the behavior is not documented either...

adonath:
@AtreyeeS I don't have a strong opinion here, but the fact that this went without complaint for such a long time might mean that some users rely on the behavior (as I said...we don't really know). In any case it will be only a very tiny fraction, so we can change without deprecation.

maxnoe:
I allowed broadcasting now, but the tests still fail on shapes that we might think of as compatible but are not to the numpy broadcasting rules:

```
FAILED gammapy/maps/hpx/tests/test_ndmap.py::test_partial_hpx_map_cutout - ValueError: Input shape (1, 1, 25) is not compatible with shape from geometry (1, 25)
FAILED gammapy/maps/hpx/tests/test_ndmap.py::test_hpx_map_to_region_nd_map - ValueError: Input shape (10,) is not compatible with shape from geometry (10, 1, 1)
FAILED gammapy/maps/hpx/tests/test_ndmap.py::test_hpxmap_read_healpy - ValueError: Input shape (12288, 1024) is not compatible with shape from geometry (12582912,)
FAILED gammapy/maps/region/tests/test_ndmap.py::test_region_nd_map_plot_two_axes - ValueError: Input shape (6,) is not compatible with shape from geometry (3, 2, 1, 1)
FAILED gammapy/maps/region/tests/test_ndmap.py::test_label_axis_io - ValueError: Input shape (10,) is not compatible with shape from geometry (2, 5, 1, 1)
FAILED gammapy/maps/region/tests/test_ndmap.py::test_region_nd_io_ogip - ValueError: Input shape (12,) is not compatible with shape from geometry (12, 1, 1)
FAILED gammapy/maps/region/tests/test_ndmap.py::test_region_nd_io_ogip_arf - ValueError: Input shape (12,) is not compatible with shape from geometry (12, 1, 1)
FAILED gammapy/maps/region/tests/test_ndmap.py::test_region_nd_io_gadf - ValueError: Input shape (2,) is not compatible with shape from geometry (2, 1, 1)
FAILED gammapy/maps/region/tests/test_ndmap.py::test_region_nd_io_gadf_no_region - ValueError: Input shape (2,) is not compatible with shape from geometry (2, 1, 1)
FAILED gammapy/maps/region/tests/test_ndmap.py::test_region_nd_io_gadf_rad_axis - ValueError: Input shape (6,) is not compatible with shape from geometry (3, 2, 1, 1)
FAILED gammapy/maps/region/tests/test_ndmap.py::test_region_nd_map_interp_no_region - ValueError: Input shape (2, 3) is not compatible with shape from geometry (3, 2, 1, 1)
FAILED gammapy/maps/wcs/tests/test_ndmap.py::test_get_spectrum - ValueError: Input shape (3,) is not compatible with shape from geometry (3, 1, 1)
FAILED gammapy/maps/wcs/tests/test_ndmap.py::test_get_spectrum_type - ValueError: Input shape (3,) is not compatible with shape from geometry (3, 1, 1)
FAILED gammapy/maps/wcs/tests/test_ndmap.py::test_get_spectrum_weights - ValueError: Input shape (3,) is not compatible with shape from geometry (3, 1, 1)
FAILED gammapy/maps/wcs/tests/test_ndmap.py::test_to_region_nd_map_histogram_basic - ValueError: Input shape (3, 100) is not compatible with shape from geometry (3, 100, 1, 1)
FAILED gammapy/maps/wcs/tests/test_ndmap.py::test_to_region_nd_map_histogram_advanced - ValueError: Input shape (2, 3, 4) is not compatible with shape from geometry (2, 3, 4, 1, 1)
```

as numpy only allows broadcasting on the left side, the last dimensions have to match.

I.e.  (3, 100) is not broadcastable to (3, 100, 1, 1), but would be to (1, 1, 3, 100).

maxnoe:
Ah, btw.: I should mention that these are only the tests failing in `gammapy.maps`, running the whole test suite, many, many, many more tests are failing.

maxnoe:
These are the unique shape combinations that make the tests fail:

```
(1000,), (1000, 1, 1)
(100050,), (2001, 50, 1, 1)
(100,), (100, 1, 1)
(100, 3, 3), (100, 1, 3, 3)
(10, 10, 10), (10, 1, 10, 10)
(10,), (10, 1, 1)
(10, 1), (10, 1, 1)
(10, 1, 1), (10, 1, 1, 1)
(10,), (2, 5, 1, 1)
(10, 25, 17), (10, 1, 25, 17)
(108, 1, 1), (108, 1, 1, 1)
(109, 1, 1), (109, 1, 1, 1)
(11098,), (62, 179, 1, 1)
(11,), (11, 1, 1)
(1, 1, 128, 128), (1, 128, 128)
(1, 1, 25), (1, 25)
(1, 1, 260, 260), (1, 260, 260)
(11, 300), (11, 300, 1, 1)
(1, 1, 32, 40), (1, 32, 40)
(1, 1, 40, 50), (1, 40, 50)
(1, 1, 50, 50), (1, 50, 50)
(1, 1, 64, 64), (1, 64, 64)
(12, 1, 1), (12, 1, 1, 1)
(1, 2), (1, 2, 1, 1)
(12,), (12, 1, 1)
(12288, 1024), (12582912,)
(1, 3), (1, 3, 1, 1)
(13,), (13, 1, 1)
(132,), (12, 11, 1, 1)
(14,), (14, 1, 1, 1)
(14, 20, 20), (14, 1, 20, 20)
(15, 1, 1), (15, 1, 1, 1)
(16, 1, 1), (16, 1, 1, 1)
(167,), (167, 1, 1)
(17, 300), (17, 300, 1, 1)
(198,), (3, 66, 1, 1)
(200,), (200, 1, 1)
(20, 1, 1), (20, 1, 1, 1)
(2, 10), (2, 10, 1, 1)
(21,), (7, 3, 1, 1)
(2, 20, 20), (2, 1, 20, 20)
(2,), (2, 1, 1)
(2,), (2, 1, 1, 1)
(2, 2), (2, 2, 1, 1)
(2, 25, 25), (2, 1, 25, 25)
(2, 25, 50), (2, 1, 25, 50)
(2, 3), (3, 2, 1, 1)
(2, 3, 4), (2, 3, 4, 1, 1)
(24,), (24, 1, 1)
(2551,), (2551, 1, 1)
(2870,), (70, 41, 1, 1)
(2, 8, 8), (2, 1, 8, 8)
(30, 22, 22), (30, 1, 22, 22)
(30, 28, 28), (30, 1, 28, 28)
(30,), (30, 1, 1)
(3, 100), (3, 100, 1, 1)
(3, 1, 2), (3, 1, 1, 2)
(32, 3, 3), (32, 1, 3, 3)
(33, 10, 10), (33, 1, 10, 10)
(3,), (3, 1, 1)
(34, 1), (34, 1, 1, 1)
(3, 4, 4), (3, 1, 4, 4)
(3, 5, 10), (3, 1, 5, 10)
(3, 66), (3, 66, 1, 1)
(3, 8, 8), (3, 1, 8, 8)
(4, 1, 90), (4, 90)
(4, 25, 25), (4, 1, 25, 25)
(4,), (4, 1, 1)
(4, 5, 5), (4, 1, 5, 5)
(48,), (48, 1, 1, 1)
(500,), (5, 10, 10, 1, 1)
(50, 1, 1), (50, 1, 1, 1)
(5,), (5, 1, 1)
(6, 1, 5986), (6, 5986)
(6,), (3, 2, 1, 1)
(6,), (6, 1, 1)
(66,), (1, 66, 1, 1)
(70, 41), (70, 41, 1, 1)
(7,), (7, 1, 1)
(80,), (80, 1, 1)
(8,), (8, 1, 1)
(8,), (8, 1, 1, 1)
(9, 1, 1), (9, 1, 1, 1)
```

I don't see anything that is not additional 1 element size dimensions.

registerrier:
Thanks @maxnoe . I think the reshape was added when `RegionNDMap` were introduced because of their peculiar shapes. That was probably not the best way to handle this. 
Maybe a better solution is to treat the specific case of `RegionNDMap` with a proper `data` setter and not the base class version.

gammapy/maps/core.py
```diff
@@ -108,7 +108,12 @@ def data(self, value):
             raise TypeError("Map data must be a Numpy array. Set unit separately")
 
         if not value.shape == self.geom.data_shape:
-            value = value.reshape(self.geom.data_shape)
+            try:
```
VascoSch92, line 111: this can be dangerous. Because you are catching all the exceptions. Let's say that there is an error which is caused by something other than the shape. Then you will have a cryptic error message about the shape and not the correct error message.

registerrier:
CI fail is unrelated. See https://github.com/pypa/setuptools/issues/4480

gammapy/maps/core.py
registerrier: modified to raise a `ValueError` after a `ValueError`

QRemy (approved):
Thanks @registerrier, no further comment.

AtreyeeS (approved):
Thank you @registerrier !
I guess back porting it would be chaotic, so I would prefer not to. I wonder if we need to inform users about this in advance, at least `fermipy` and `agnpy`?

PR #5184, opened 2024-03-26
Support additional axes in Mapdataset.create

registerrier:
**Description**

This pull request introduces support for more complex `Geom` objects for `MapDataset.create`. This will allow higher dimensionality products e.g. (ra, dec, energy, event_type) or (ra, dec, energy, phase) or for 1D datasets (energy, obs_id).

**Dear reviewer**


AtreyeeS (approved):
Thank you @registerrier ! I don't have any specific comments on the implementation here, but do the IRFs need the extra axis? It depends on the use case no? For event types or a time axis, yes. But for phase, probably not.

PR #5186, opened 2024-03-29
Adjust `to_edisp_kernel` input to `MapAxis`

### LINKED ISSUE #5142: Expect the `MapAxis` directly in `to_edisp_kernel`
Should we expect directly the `MapAxis` here? (Note: this is a TODO converted into an issue)

https://github.com/gammapy/gammapy/blob/e5aecb334e7aebe304affd96b545a636019f7626/gammapy/irf/edisp/core.py#L145-L148

**Solution**
'energy' should be a `MapAxis` type. This needs a small API changes and deprecations so that 'energy' is a `MapAxis` and not a `Quantity`.

Astro-Kirsty:
This PR is to resolve #5142. 
I left it as a draft as I am not sure if this is the correct procedure to take. 
This is two part issue, we want to change the variables to be consistent with the other naming for axis https://github.com/gammapy/gammapy/blob/86488ece61f41e92ff3212b9b870f5a9f7d68b79/gammapy/irf/edisp/map.py#L116
We also need to change the variable to no longer be a quantity but a `MapAxis`. 

This was the best way I could think to achieve this.

Note: there are a number of failing tests, but I did not want to fix them all with the deprecation warning until a method was decided upon.

I am open to suggestions.

### DIFF
--- a/gammapy/irf/edisp/core.py
+++ b/gammapy/irf/edisp/core.py
@@ -4,11 +4,10 @@ import numpy as np
 import scipy.special
 from astropy import units as u
 from astropy.coordinates import Angle, SkyCoord
-from astropy.units import Quantity
 from astropy.visualization import quantity_support
 import matplotlib.pyplot as plt
 from matplotlib.colors import PowerNorm
-from gammapy.maps import MapAxes, MapAxis, RegionGeom
+from gammapy.maps import MapAxes, RegionGeom
 from gammapy.utils.deprecation import deprecated_renamed_argument
 from gammapy.visualization.utils import add_colorbar
 from ..core import IRF
@@ -153,21 +152,6 @@ class EnergyDispersion2D(IRF):
         from gammapy.makers.utils import make_edisp_kernel_map
 
         offset = Angle(offset)
-
-        if isinstance(energy_axis, Quantity):
-            log.warning(
-                "From v1.3 energy_axis should be given as a MapAxis, not a Quantity."
-            )
-            energy_axis = MapAxis.from_energy_edges(energy_axis)
-
-        if isinstance(energy_axis_true, Quantity):
-            log.warning(
-                "From v1.3 energy_axis_true should be given as a MapAxis, not a Quantity."
-            )
-            energy_axis_true = MapAxis.from_energy_edges(
-                energy_axis_true,
-                name="energy_true",
-            )
 
         if energy_axis is None:
             energy_axis = self.axes["energy_true"].copy(name="energy")
--- a/gammapy/irf/edisp/core.py
+++ b/gammapy/irf/edisp/core.py
@@ -4,10 +4,11 @@ import numpy as np
 import scipy.special
 from astropy import units as u
 from astropy.coordinates import Angle, SkyCoord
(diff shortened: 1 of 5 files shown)


gammapy/irf/edisp/core.py
```diff
@@ -117,7 +124,13 @@ def from_gauss(
             data=data.value,
         )
 
-    def to_edisp_kernel(self, offset, energy_true=None, energy=None):
+    @deprecated_renamed_argument(
+        ["energy_true", "energy"],
+        ["energy_axis_true", "energy_axis"],
```
AtreyeeS, line 129: Do you want to rename the argument name or keep the name and change the type?
Either of the two should suffice, no?

If a `DepricationWarning` is raised about the argument name, then the `warning` below is not necessary.

I think we can keep the argument same same, and just allow a `MapAxis` on input. Then, we can raise a deprecation warning if a Quantity is passed

AtreyeeS (changes requested):
Thanks @Astro-Kirsty ! 

I do not think you need to adapt internet tests to check for the deprecation warning, eg as done in `gammapy/datasets/tests/test_map.py` 
It only needs to be checked when the user passes a quantity instead of a `MapAxis` in the `to_edisp_kernel` function, and internally within Gammapy we should pass a `MapAxis`.

gammapy/irf/edisp/core.py
Astro-Kirsty: I changed the argument name to be consistent with others in the code, for example [here](https://github.com/gammapy/gammapy/blob/18b65d5cce21876d53aa7faf54aaeccd68e8b76c/gammapy/irf/edisp/map.py#L116).

Astro-Kirsty: Okay sure thing! Will do

Astro-Kirsty: So `deprecated_renamed_argument` will only let the user know that there was a name change, but not that it should now be a `MapAxis`

QRemy (dismissed):
Thanks @Astro-Kirsty looks good.

AtreyeeS (approved):

gammapy/irf/edisp/core.py
```diff
@@ -141,18 +153,11 @@ def to_edisp_kernel(self, offset, energy_true=None, energy=None):
 
         offset = Angle(offset)
 
-        if energy is None:
+        if energy_axis is None:
```
AtreyeeS, line 156: Will this work correctly for the old behaviour? We have to support passing quantities for 1.3

AtreyeeS (commented):
Please check that the deprecation works correctly, ie, the old behaviour is still supported

gammapy/irf/edisp/core.py
Astro-Kirsty: Apologies I removed a few lines when removed the warning.
Fixing now.

AtreyeeS (approved):
No further comments. Thanks!

registerrier (approved):
Thanks @Astro-Kirsty . This looks good!

PR #5188, opened 2024-03-30
Raise an error if the geometry of the exclusion mask passed to `ReflectedRegionsBackgroundMaker` is not an image

### LINKED ISSUE #5185: Raise an error when passing an energy dependent exclusion map to `ReflectedRegionsBackgroundMaker`
If one pass an energy dependent exclusion map to the `ReflectedRegionsBackgroundMaker`, the error that is raised is really unclear (see picture below). We should add a check on the maker class to check whether the exclusion map contains an energy axis (or any other axis actually), and raise an explicit error to the user. 

Capture d’écran 2024-03-29 à 14 05 28

MRegeard:
This PR solves resolves #5185 . 

I added a check that raise a `ValueError` if the geometry of the exclusion mask passed to the `ReflectedRegionsBackgroundMaker` is not an image.


registerrier:
Can you rebase @MRegeard ?

MRegeard:
@registerrier Done!

gammapy/makers/background/reflected.py
```diff
@@ -451,6 +451,9 @@ def __init__(
         if exclusion_mask and not exclusion_mask.is_mask:
             raise ValueError("Exclusion mask must contain boolean values")
 
+        if exclusion_mask and not exclusion_mask.geom.is_image:
```
registerrier, line 454:
```suggestion
        if exclusion_mask and not exclusion_mask.geom.is_flat:
```
this will work with 1D axis too.

gammapy/makers/background/tests/test_reflected.py
```diff
@@ -456,3 +456,18 @@ def test_reflected_bkg_maker_fixed_rad_max_bad(
     dataset = maker.run(dataset_empty, obs)
     with pytest.raises(TypeError):
         reflected_bkg_maker.run(dataset, obs)
+
+
+def test_reflected_bkg_exclsion_error(exclusion_mask):
+
+    energy = MapAxis.from_energy_bounds(1 * u.TeV, 2 * u.TeV, nbin=1, name="energy")
```
registerrier, line 463: It should with only one bin in energy right?

gammapy/makers/background/reflected.py
registerrier: Maybe log a warning if flat but not image to say that the additional axis edges are not taken into account.

registerrier (commented):
Thanks @MRegeard . Looks good.
I think you should specifiy in the doc that the exclusion mask has to be flat. And possibly accept flat geometries rather than only image geom. I assume it works right?

MRegeard:
@registerrier, 
We don't support flat geometry, we only support image. Is this something that we want ?
In this case we could drop the extra axis at instantiation if the geometry is flat.

registerrier:
> @registerrier, We don't support flat geometry, we only support image. Is this something that we want ? In this case we could drop the extra axis at instantiation if the geometry is flat.



> @registerrier, We don't support flat geometry, we only support image. Is this something that we want ? In this case we could drop the extra axis at instantiation if the geometry is flat.

I think we should. Does the maker fail currently with a flat geometry? 
It can be done in a subsequent PR, if you prefer.

MRegeard:
Yes the maker fails if the geometry is flat. 
I can do it in this PR !

bkhelifi:
A naive question: what is the difference between image and flat?

adonath:
> A naive question: what is the difference between image and flat?

A cube with a single energy dimension is flat for example shape `(1, 500, 500)`, while and image has `(500, 500)`

gammapy/makers/background/reflected.py
```diff
@@ -451,6 +451,13 @@ def __init__(
         if exclusion_mask and not exclusion_mask.is_mask:
             raise ValueError("Exclusion mask must contain boolean values")
 
+        if exclusion_mask and not exclusion_mask.geom.is_flat:
+            raise ValueError("Exclusion mask must only contain spatial dimension")
+
+        if exclusion_mask:
+            exclusion_mask = exclusion_mask.sum_over_axes(keepdims=False)
```
bkhelifi, line 458: To summarize,
- if the non-spatial axis has a dimension equal to 1 (ie is_flat), then there is an error
- if the non-spatial axis dimension is strictly larger than, then we sum over the axes.
Is it correct?

With event_type, then, it should also work in principle...

registerrier: This will accept only 1 bin long non spatial axes. If we want to support multiple label axes for datasets with label axes, we will have to introduce the feature in general. But in another PR, once `Makers` support additional axes.

registerrier (approved):

MRegeard:
@registerrier, I added a few lines in the docstring to explain the behaviour. 
So I leave it like this and open a follow up issue regarding your comments ?

registerrier (approved):

PR #5193, opened 2024-04-02
Fix docs

maxnoe:
**Description**

This pull request forces the light theme to be used by the docs regardless of system preference.

**Dear reviewer**

PR #5197, opened 2024-04-03
Remove sphinx from environment-dev.yml

registerrier:
**Description**

This pull request tries to solve the CI issues. 

In general, the dev environment  does not have to the same as the one used to build docs. This PR separates them by removing sphinx and associated packages from `environment-dev.yml`. The correct approach used in CI to build the docs is with `tox -e build_docs`.

The PR also solves an issue appearing with astropy 6.0.1 because of a `Unit` (rather than a `Quantity`) raised to a power.

**Dear reviewer**


maxnoe:
@registerrier for the astropy issue also see #5198

maxnoe:
I think it would still be good to have a working docs building setup in the env, e.g. to be able to run `sphinx-autobuild` when working on the docs. Always running in tox will result in a full rebuild that takes ages.

maxnoe:
In #5193, I fixed the requirements in environment.yaml to work again

registerrier:
Closing since issue is solved by #5193

Issue #5202, opened 2024-04-08
More informative errors for `MapDatasetEventSampler`


fabiopintore:
**Is your feature request related to a problem? Please describe.**
When sampling a number of sources, `MapDatasetEventSampler` can raise a `ValueError` if it fails to sample a source. However, the error is not very informative on which source cannot be sampled. 

**Describe the solution you'd like**
The `ValueError` should indicate the source which is failing to sample.

**Additional context**
To replicate:

```
spec_model1 = PowerLawSpectralModel(amplitude="-1e-12 cm-2 s-1 TeV-1")
spatial_model1 = PointSpatialModel()

model1 = SkyModel(spectral_model = spec_model1,
                  spatial_model = spatial_model1,
                  name="source1")

model2 = SkyModel(spectral_model = spec_model1,
                  spatial_model = spatial_model1,
                  name="source2")

bkg = FoVBackgroundModel(dataset_name="my-dataset")

models1 = Models([model1, model2, bkg])


path = Path("$GAMMAPY_DATA/cta-caldb")
location = observatory_locations["cta_south"]
irf_filename = "Prod5-South-20deg-AverageAz-14MSTs37SSTs.180000s-v0.1.fits.gz"
irfs = load_irf_dict_from_file(path / irf_filename)

observation = Observation.create(
            obs_id="0001",
            pointing=spatial_model1.position,
            livetime=1 * u.hr,
            irfs=irfs,
            location=location,
            )


energy_axis = MapAxis.from_energy_bounds("1 TeV", "10 TeV",
                                         nbin=10, per_decade=True)
energy_axis_true = MapAxis.from_energy_bounds(
    "0.7 TeV", "20 TeV", nbin=10, per_decade=True, name="energy_true"
)
migra_axis = MapAxis.from_bounds(0.5, 2, nbin=150, node_type="edges", name="migra")

geom = WcsGeom.create(
    skydir=spatial_model1.position,
    width=(1,1),
    binsz=0.02 * u.deg,
    frame="galactic",
    axes=[energy_axis],
)

empty = MapDataset.create(
    geom,
    energy_axis_true=energy_axis_true,
    migra_axis=migra_axis,
    name="my-dataset",
)
maker = MapDatasetMaker(selection=["exposure", "background", "psf", "edisp"])
dataset = maker.run(empty, observation)

dataset.models = models1

sampler = MapDatasetEventSampler(random_state=0)
events = sampler.run(dataset, observation)
```
which gives only the following error:

[...]

PR #5205, opened 2024-04-08
Specify the sampled model in `MapDatasetEventSampler`

fabiopintore:
This PR is intended to solve the issue #5202 and it introduces the possibility to print out the name of the model sampled by the `MapDatasetEventSampler`.

### DIFF
--- a/gammapy/datasets/simulate.py
+++ b/gammapy/datasets/simulate.py
@@ -1,6 +1,7 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 """Simulate observations."""
 import html
+import logging
 from copy import deepcopy
 import numpy as np
 import astropy.units as u
@@ -20,6 +21,8 @@ from gammapy.utils.random import get_random_state
 from .map import create_map_dataset_from_observation
 
 __all__ = ["MapDatasetEventSampler", "ObservationEventSampler"]
+
+log = logging.getLogger(__name__)
 
 
 class MapDatasetEventSampler:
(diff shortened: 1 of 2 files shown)


gammapy/datasets/simulate.py
```diff
@@ -21,6 +22,9 @@
 
 __all__ = ["MapDatasetEventSampler", "ObservationEventSampler"]
 
+logging.basicConfig(level=logging.INFO)
```
adonath, line 25: Please remove the configuration statement here. This usually happens on a global call in the user script.

gammapy/datasets/tests/test_simulate.py
```diff
@@ -590,6 +590,43 @@ def test_mde_run(dataset, models, tmp_path):
     assert u.isclose(obs_back.observatory_earth_location.height, LOCATION.height)
 
 
+@requires_data()
+def test_sampler_output(dataset, model_alternative, caplog):
```
adonath, line 594: Probably it's better to add the testing of the logging output to one of the existing test. Sampling a whole new event list just for testing the log output is a bit too much.

adonath (commented):
Thanks @fabiopintore! I have left some comment...

gammapy/datasets/simulate.py
fabiopintore: If I remove it, there is no more output from `logger.info`. I converted all the `logger.info` to simple `print`.

adonath: No, please keep the logging statements. We do not use print statements anywhere in the Gammapy code. My points was that the configuration should happen in the user scripts, or in you case, move the logger configuration to the test file.

bkhelifi (approved):
Thanks @fabiopintore ... Looks good for me.

Issue #5178, opened 2024-03-25
`MapDatasetEventSampler._sample_coord_time` fails when `Npred` is negative or NaN


fabiopintore:
**Gammapy version**
1.2

**Bug description**
`MapDatasetEventSampler._sample_coord_time` fails when the predicted number of events of a given model is negative or NaN. The fail is in the randomization of `n_events = self.random_state.poisson(np.sum(data))`.

**Expected behavior**
 I think we should set a condition in the event sampler that, when the predicted number is negative, skips the model from the sampling.

**To Reproduce**
Issues found with some binary or pulsar models of the SDC.

fabiopintore:
Hi @registerrier , not sure yet on the reasons of the issue but it seems to me a bad evaluation of the `PhasecurveTemporalModel` when the source is very weak and the phaseogram has bins with normalization zero...

PR #5206, opened 2024-04-08
Convert negative `npred` values to zero in `MapDatasetEventSampler`

fabiopintore:
It may happen that the `npred` for a given model assume negative values. It can introduce some issues in the correct total number of events which will be sampled for the model. The PR introduces the possibility to convert all the negative `npred` values into zeros. It would solve also the issue #5178 .

### DIFF
--- a/gammapy/datasets/simulate.py
+++ b/gammapy/datasets/simulate.py
@@ -191,6 +191,7 @@ class MapDatasetEventSampler:
 
         npred = self._evaluate_timevar_source(dataset, model=model)
         data = npred.data[np.isfinite(npred.data)]
+        data[data < 0] = 0
 
         try:
             n_events = self.random_state.poisson(np.sum(data))
@@ -225,6 +226,7 @@ class MapDatasetEventSampler:
             Table of sampled events.
         """
         data = npred.data[np.isfinite(npred.data)]
+        data[data < 0] = 0
         n_events = self.random_state.poisson(np.sum(data))
 
         coords = npred.sample_coord(n_events=n_events, random_state=self.random_state)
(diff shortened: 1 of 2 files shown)


gammapy/datasets/simulate.py
```diff
@@ -225,6 +226,7 @@ def _sample_coord_time(self, npred, temporal_model, gti):
             Table of sampled events.
         """
         data = npred.data[np.isfinite(npred.data)]
+        data[data < 0] = 0
```
registerrier, line 229: maybe `np.clip` will be faster.

gammapy/datasets/simulate.py
```diff
@@ -191,6 +191,7 @@ def _sample_coord_time_energy(self, dataset, model):
 
         npred = self._evaluate_timevar_source(dataset, model=model)
         data = npred.data[np.isfinite(npred.data)]
+        data[data < 0] = 0
```
registerrier, line 194: Maybe use `np.clip`

registerrier (commented):
Thanks @fabiopintore . Looks good, but you might use clipping instead

bkhelifi:
Do we really understand why they are negative values?

fabiopintore:
Unfortunately not yet. This is something that happens only in very rare cases, so my suspicion is that there is an interpolation issue that affect the evaluation during the phase bins with zero normalization.
I haven't had enough time to investigate it in detail but I count to do it after next week.

bkhelifi:
Thanks Fabio

cgalelli (approved):
Thanks @fabiopintore ! Would be nice to understand where the problem is coming from but I think a fix like this is required anyway to block other occurrences of the problem

registerrier (approved):
Thanks @fabiopintore . Looks good to me.

PR #5215, opened 2024-04-15
Add wrapper for astromodels

QRemy:
This PR add wrappers  to use astromodels in gammapy.
Not sure if this should be included in gammapy or astromodels, any preferences  ?  @omodei @adonath

For now this is a draft I haven't tested all models but only one for each type of model.
I am not completely sure that the unit and frame  handling works properly.


QRemy:
Closing this PR as the interface between gammapy, 3ML and astromodels should be done in a separate package.

PR #5216, opened 2024-04-16
Compile against numpy 2.0 when building wheels, define target api version

maxnoe:
This is needed to provide wheels that are compatible with both numpy 2.0 and 1.x.

Building will happen always against the newest version of numpy and wheels will be ABI compatible with numpy >= `NPY_TARGET_VERSION`

**Description**

This pull request ...

**Dear reviewer**

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,9 +2,8 @@
 requires = [
 	"setuptools >= 65",
 	"Cython",
-	'oldest-supported-numpy',
+	'numpy>=2.0.0rc1',
 	'setuptools_scm[toml] >= 8',
-	"extension-helpers",
 ]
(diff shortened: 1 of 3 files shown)


registerrier:
The devdeps tests now run locally once removing the astropy and numpy constraints.

There are numerous errors appearing. Most of them are due to the change in `Quantity(copy=False)` behavior. See: https://github.com/astropy/astropy/issues/16167
We should follow the numpy 2.0 (and astropy) scheme of using `Quantity(copy=None)` to keep the same behavior.

registerrier (approved):
Thanks @maxnoe . The devdeps tests are now properly run. I propose to solve the fails in a subsequent PR.

PR #5220, opened 2024-04-17
Fix errors to documentation after the new docs have been implemented

Astro-Kirsty:
This PR is to adapt the docs since the update to the css


cgalelli (approved):
Thanks @Astro-Kirsty !

registerrier (approved):
The test fails seem unrelated.

PR #5221, opened 2024-04-17
Add wrapper around 3ML minimizer wrappers 

QRemy:
This PR add a wrapper around 3ML minimizer wrappers so we can use them via gammapy.modelling.Fit(backend="3ml").

For now this is a draft, it works for the simple test dataset with minuit and scipy but I haven't installed the other optimizers yet to test locally.

### DIFF
--- a/gammapy/modeling/tests/test_fit.py
+++ b/gammapy/modeling/tests/test_fit.py
@@ -180,6 +180,25 @@ def test_optimize(backend):
     assert len(result.trace) == result.nfev
 
 
+@pytest.mark.parametrize("method", ["MINUIT", "SCIPY"])
+def test_optimize_3ml(method):
+    backend = "3ml"
+    dataset = MyDataset()
+
+    kwargs = {"method": method}
+
+    fit = Fit(store_trace=True, backend=backend, optimize_opts=kwargs)
+    result = fit.optimize([dataset])
+
+    pars = dataset.models.parameters
+
+    assert_allclose(result.total_stat, 0, atol=1)
+
+    assert_allclose(pars["x"].value, 2, rtol=1e-2)
+    assert_allclose(pars["y"].value, 3e2, rtol=1e-3)
+    assert_allclose(pars["z"].value, 4e-2, rtol=1e-2)
+
+
 @pytest.mark.parametrize("backend", ["minuit"])
 def test_confidence(backend):
     dataset = MyDataset()
--- a/gammapy/modeling/threeML_wrapper.py
+++ b/gammapy/modeling/threeML_wrapper.py
@@ -1,0 +1,53 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+"""3ml minimization wrapper"""
+import numpy as np
+from astromodels import Parameter as AstroParameter
+from .iminuit import MinuitLikelihood
+
+
+def optimize_3ml(parameters, function, store_trace=False, **kwargs):
+    """iminuit optimization.
+
+    Parameters
+    ----------
+    parameters : `~gammapy.modeling.Parameters`
+        Parameters with starting values.
+    function : callable
+        Likelihood function.
+    store_trace : bool, optional
+        Store trace of the fit. Default is False.
+    **kwargs : dict
+        Options passed to `iminuit.Minuit` constructor. If there is an entry
+        'migrad_opts', those options will be passed to `iminuit.Minuit.migrad()`.
+
+    Returns
+    -------
+    result : (factors, info, optimizer)
+        Tuple containing the best fit factors, some information and the optimizer instance.
+    """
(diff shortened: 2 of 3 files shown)


gammapy/modeling/threeML_wrapper.py
```diff
@@ -0,0 +1,53 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+"""3ml minimization wrapper"""
```
bkhelifi, line 2: Some reference of 3ML might be interesting to have

gammapy/modeling/threeML_wrapper.py
```diff
@@ -0,0 +1,53 @@
...
+    **kwargs : dict
+        Options passed to `iminuit.Minuit` constructor. If there is an entry
+        'migrad_opts', those options will be passed to `iminuit.Minuit.migrad()`.
+
+    Returns
+    -------
+    result : (factors, info, optimizer)
+        Tuple containing the best fit factors, some information and the optimizer instance.
+    """
+    from threeML.minimizer.minimization import get_minimizer
```
bkhelifi, line 28: I am wondering whether one should not protect this import by raising a clear error message if threeML is not installed...

gammapy/modeling/threeML_wrapper.py
```diff
@@ -0,0 +1,53 @@
...
+    for p in parameters:
+        min_ = None if np.isnan(p.factor_min) else p.factor_min
+        max_ = None if np.isnan(p.factor_max) else p.factor_max
+
+        parameters_3ml[p.name] = AstroParameter(
+            name="".join(filter(str.isalnum, p.name)),
+            value=p.factor,
+            min_value=min_,
+            max_value=max_,
+            free=not p.frozen,
```
bkhelifi, line 44: No need to propagate the priors?

gammapy/modeling/threeML_wrapper.py
```diff
@@ -0,0 +1,53 @@
...
+        )
+
+    optimizer = optimizer_class(
+        parameters=parameters_3ml, function=func_3ml.fcn, **kwargs
+    )
+    factors, fval = optimizer._minimize()
+    info = {"success": None, "message": None, "trace": func_3ml.trace, "nfev": None}
```
bkhelifi, line 51: Why success, message and nfev are None?

gammapy/modeling/tests/test_fit.py
```diff
@@ -180,6 +180,25 @@ def test_optimize(backend):
     assert len(result.trace) == result.nfev
 
 
+@pytest.mark.parametrize("method", ["MINUIT", "SCIPY"])
```
bkhelifi, line 183: Should we not add a test about the similarity of the fitted resultats between different backends, ie scipy, iminuit, 3ml? This high level test is quite important

bkhelifi (commented):
Thanks @QRemy ... I have started to view this PR. I guess one needs to merge #5215 to see if the CI is good or not. Is not it?

QRemy:
The CI will fail because importing 3ML trigger a number of deprecation warning.
Moreover to properly test this we would need all the 3ML dependencies so at the end it would be simpler if this could go in 3ML instead. Then a gammapy user that want to use minimizers supported by 3ML could simply import this wrapper from 3ML and add it to gammapy optimizer registry.  The approach would be the similar to what we have to add custom models.

Any opinion @omodei ? would you be interested to take this PR into 3ML ?

omodei:
I will have a look, but, if I have understood the idea of this code, I don’t think it should go in 3M. 3ML idea is to add plug in for whichever instrument wants to use it. Then you don’t need to add any wrapper to the fitting engines, because they will be naturally available.  The plug in itself, to some extent, is the wrapper between threeml modeling (astromodels) and the instrument modeling. There are ongoing projects to develop gammapy plug-in in 3ML, which is the cleaner way to do it (that will also allow you to perform multi instrument fitting (gammapy+ all the other instrument supported by 3ML). You are welcome to participate to those efforts.

QRemy:
Thanks @omodei for the feedback, so I guess that #5215 is more of interest for you as it would allows to convert any astromodels into a model that your gammapy plug-in for 3ML can understand.

omodei:
Yes. JMichael did something similar for the gammapy plugin. I’ll try to find a common place where to add this, so that we can all work on the plug-in.

bkhelifi:
#5215 brings a science use case interesting for the Gammapy user:  an astromodel fitted by 3ML can be transmitted to Gammapy to make further analysis. This is then about the interoperability astromodel -> gammapy.models. Is not it?
By the way, can 3ML serialise properly fitted astromodels? I imagine than yes...

QRemy:
Closing this PR as the interface between gammapy, 3ML and astromodels should be done in a separate package.

PR #5222, opened 2024-04-19
Allow `MapAxis` to be passed in `SpectralModel.plot()`

### LINKED ISSUE #5171: Allow `MapAxis` to be passed in `SpectralModel.plot()` ?
Currently in the `SpectralModel.plot()` we allow the user to define the `energy_bounds`. (This would also apply to the `plot_error` function)
Should we also allow them to pass a `MapAxis`? 
This relates to discussions here: https://github.com/gammapy/gammapy/issues/5086#issuecomment-1935949104 and https://github.com/gammapy/gammapy/issues/3078#issuecomment-715320223. 

If we wish to allow this there will be an API change. 
Both `energy_bounds` and `map_axis` would need to be set to `None` inside the function:
` def plot(self, energy_bounds=None, map_axis=None......)`

Then the user would defined either the `energy_bounds` or `map_axis`. I am not sure on this solution however because I do not think it makes sense to have them set to None, this would be confusing to the user and they may think the function can be run without defining either. 

Open to further discussion on whether we wish to allow the functionality to pass a `MapAxis`, and how we could implement that.

Astro-Kirsty:
This PR resolves #5171

We can now accept a `MapAxis` for spectral plotting both in `plot` and `plot_error`.

Note: `if energy_bounds is not None:` must be used rather just `if energy_bounds` due to the quantity handling in [astropy](https://github.com/astropy/astropy/pull/14124).

### DIFF
--- a/gammapy/modeling/models/spectral.py
+++ b/gammapy/modeling/models/spectral.py
@@ -456,9 +456,7 @@ class SpectralModel(ModelBase):
         if self.is_norm_spectral_model:
             sed_type = "norm"
 
-        if (
-            isinstance(energy_bounds, (tuple, list, Quantity))
-        ):
+        if isinstance(energy_bounds, (tuple, list, Quantity)):
             energy_min, energy_max = energy_bounds
             energy = MapAxis.from_energy_bounds(
                 energy_min,
@@ -539,9 +537,7 @@ class SpectralModel(ModelBase):
         if self.is_norm_spectral_model:
             sed_type = "norm"
 
-        if (
-            isinstance(energy_bounds, (tuple, list, Quantity))
-        ):
+        if isinstance(energy_bounds, (tuple, list, Quantity)):
             energy_min, energy_max = energy_bounds
             energy = MapAxis.from_energy_bounds(
                 energy_min,
--- a/gammapy/modeling/models/spectral.py
+++ b/gammapy/modeling/models/spectral.py
@@ -487,8 +487,7 @@ class SpectralModel(ModelBase):
 
     def plot_error(
         self,
-        energy_bounds=None,
-        energy_axis=None,
+        energy_bounds,
         ax=None,
         sed_type="dnde",
         energy_power=0,
--- a/gammapy/modeling/models/spectral.py
+++ b/gammapy/modeling/models/spectral.py
@@ -428,9 +428,9 @@ class SpectralModel(ModelBase):
         Parameters
         ----------
         energy_bounds : `~astropy.units.Quantity`
-            Plot energy bounds passed to `~gammapy.maps.MapAxis.from_energy_bounds`.
-        energy_axis : `~gammapy.maps.MapAxis`
-            Plot the energy axis.
+            Energy bounds between which the model is to be plotted.
+        energy_axis : `~gammapy.maps.MapAxis`, optional
+            Axis defining the energy bounds between which the model is to be plotted.
         ax : `~matplotlib.axes.Axes`, optional
             Matplotlib axes. Default is None.
         sed_type : {"dnde", "flux", "eflux", "e2dnde"}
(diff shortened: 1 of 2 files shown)


gammapy/modeling/models/spectral.py
```diff
@@ -427,7 +428,9 @@ def plot(
         Parameters
         ----------
         energy_bounds : `~astropy.units.Quantity`
-            Plot energy bounds passed to MapAxis.from_energy_bounds.
+            Plot energy bounds passed to `~gammapy.maps.MapAxis.from_energy_bounds`.
```
bkhelifi, line 431: The description with  gammapy.maps.MapAxis.from_energy_bounds is very technical and only experts can understand...
Also you have not mentioned the default values

gammapy/modeling/models/spectral.py
```diff
@@ -443,20 +446,31 @@ def plot(
...
         if self.is_norm_spectral_model:
             sed_type = "norm"
 
-        energy_min, energy_max = energy_bounds
-        energy = MapAxis.from_energy_bounds(
-            energy_min,
-            energy_max,
-            n_points,
-        )
+        if energy_bounds is not None:
```
bkhelifi, line 459: If both are defined, which one to use? Why the priority is given to energy_bounds?

gammapy/modeling/models/spectral.py
```diff
@@ -500,6 +515,8 @@ def plot_error(
         ----------
         energy_bounds : `~astropy.units.Quantity`
             Plot energy bounds passed to `~gammapy.maps.MapAxis.from_energy_bounds`.
+        energy_axis : `~gammapy.maps.MapAxis`
```
bkhelifi, line 518: See my comment above

bkhelifi (changes requested):
Thanks @Astro-Kirsty ... I have just some comments on the docstrings.

AtreyeeS (commented):
Thanks @Astro-Kirsty ! 
My main comment is that `energy_axis` should be given priority over `energy_bounds`. Also, to reduce the code duplication between `plot` and `plot_error`, maybe you can create a small hidden function `_get_axis_for_plot` and move the common code you introduced.

registerrier (commented):
Thanks @Astro-Kirsty . 

If there are two possible exclusive arguments, the default behavior should be to fail if both are defined.
Note the change modifies makes all arguments optional  and modify their order. This is an API breaking change.

Another option could be to let `energy_bounds` be either a `MapAxis` or a two valued energy `Quantity`. In itself, `energy_bounds` does not tell much about its type.

Astro-Kirsty:
> Another option could be to let `energy_bounds` be either a `MapAxis` or a two valued energy `Quantity`. In itself, `energy_bounds` does not tell much about its type.

I think this could be a good idea. Do others have an opinion? @AtreyeeS @bkhelifi

Astro-Kirsty:
Updated the PR to allow for different types in the energy_bounds rather than another keyword as per discussions in the dev call last Friday

gammapy/modeling/models/spectral.py
```diff
@@ -443,20 +445,32 @@ def plot(
         -------
         ax : `~matplotlib.axes.Axes`, optional
             Matplotlib axes.
+
+        Notes
+        -----
+        If ``energy_bounds`` is supplied as a tuple or Quantity, an ``energy_axis`` is created internally with
```
bkhelifi, line 451: Or a list also;)

bkhelifi (dismissed):
Thanks a lot @Astro-Kirsty ... Looks ready to be merged!

gammapy/modeling/models/spectral.py
```diff
@@ -443,20 +445,32 @@ def plot(
...
-        energy_min, energy_max = energy_bounds
-        energy = MapAxis.from_energy_bounds(
-            energy_min,
-            energy_max,
-            n_points,
-        )
+        if (
+            isinstance(energy_bounds, tuple)
+            or isinstance(energy_bounds, Quantity)
+            or isinstance(energy_bounds, list)
```
QRemy, line 462:
```suggestion
            isinstance(energy_bounds, (tuple, list, Quantity))
```

gammapy/modeling/models/spectral.py
```diff
@@ -515,20 +530,32 @@ def plot_error(
...
```
QRemy, line 547:
```suggestion
            isinstance(energy_bounds, (tuple, list, Quantity))
```

QRemy (approved):
Thanks @Astro-Kirsty, looks good I have only minor style comments.

bkhelifi (approved):
All good!

PR #5236, opened 2024-04-24
Fix ExcessMapEstimator reco exposure if IRFs geom different from dataset geom

QRemy:
This applies correctly  the `safe_mask` in the computation of `ExcessMapEstimator.reco_exposure`  if the IRFs geom is different from the dataset geom and prevent the output map to be empty.


QRemy:
Solved by #5271

PR #5241, opened 2024-04-29
Fix test fail

Astro-Kirsty:


gammapy/estimators/points/tests/test_sed.py
```diff
@@ -656,25 +656,25 @@ def test_flux_points_parallel_ray_actor_map(fpe_map_pwl):
     assert_allclose(actual, [0.343332, 3.162278, 29.126327], rtol=1e-5)
 
     actual = table["norm"].data
-    assert_allclose(actual, [0.974726, 0.96342, 0.994251], rtol=1e-2)
+    assert_allclose(actual, [0.95551238, 0.96532761, 0.9952374], rtol=1e-2)
 
     actual = table["norm_err"].data
-    assert_allclose(actual, [0.067637, 0.052022, 0.087059], rtol=3e-2)
+    assert_allclose(actual, [0.08350122, 0.05249954, 0.08750922], rtol=3e-2)
```
bkhelifi, line 662: There is a rather large changes here... And the test are also here to point out issues.
What cause this change? Do we understand it? Are they due to a subtle bug introduction?

gammapy/estimators/points/tests/test_sed.py
```diff
@@ -656,25 +656,25 @@ def test_flux_points_parallel_ray_actor_map(fpe_map_pwl):
...
     actual = table["counts"].data
     assert_allclose(actual, [[44611, 0], [1923, 0], [282, 0]])
 
     actual = table["norm_ul"].data
-    assert_allclose(actual, [1.111852, 1.07004, 1.17829], rtol=1e-2)
+    assert_allclose(actual, [1.1245298, 1.07300974, 1.17931627], rtol=1e-2)
 
     actual = table["sqrt_ts"].data
-    assert_allclose(actual, [16.681221, 28.408676, 21.91912], rtol=1e-2)
+    assert_allclose(actual, [12.19487811, 27.37400989, 21.30540009], rtol=1e-2)
```
bkhelifi, line 671: The change is very significant here, and towards a lower significance!!! Why?

gammapy/estimators/points/tests/test_sed.py
Astro-Kirsty: I am not entirely sure, but it appears to be something with the DatasetsActor because other tests utilising the same datasets methods are not failing.
@QRemy, do you have an idea?

registerrier (commented):
Thanks @Astro-Kirsty . Where does it fail exactly? Not in CI, right?

registerrier:
> Thanks @Astro-Kirsty . Where does it fail exactly? Not in CI, right?

I think that ray testing is actually only partly made in CI. See #5243

Astro-Kirsty:
Ahh okay @registerrier , it fails locally

QRemy:
The fail depends on the execution order of `test_flux_points_parallel_ray_actor_map(fpe_map_pwl)`and `test_run_map_pwl(fpe_map_pwl)`, if `test_flux_points_parallel_ray_actor_map` use its fixture the problem is gone and the results values are the same than `test_run_map_pwl`.
I'm going to push a commit with that fix.

QRemy:
@Astro-Kirsty can you check it works for you locally ?

Astro-Kirsty:
All good on my end, thanks @QRemy

QRemy:
@meeseeksdev backport to v1.2.x

PR #5243, opened 2024-04-30
Update CI

registerrier:
**Description**

This pull request modifies CI to support ray testing and drops python 3.9.

**Dear reviewer**

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,7 +22,7 @@ jobs:
         include:
           - os: ubuntu-latest
             python: '3.11'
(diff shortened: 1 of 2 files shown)


registerrier:
Some notes here:
- the coverage was incorrect since several months because gammapy-data was not properly installed in the corresponding CI run. It seems fine now.
- It seems impossible to get sherpa to compile on python 3.11 with ray at the same time on ubuntu-latest. I moved the full test on python 3.10 where it seems to work

registerrier:
The CI failure within test hypothesis is unclear but should be unrelated. Maybe an issue with a modified fixture? 
Merging for now.

PR #5250, opened 2024-05-07
Fix example in new timmerkonig

Astro-Kirsty:
These are some small adaptions that didnt make it into the PR by @cgalelli before merge


cgalelli (approved):
Many thanks @Astro-Kirsty !

QRemy (approved):

PR #5254, opened 2024-05-07
Make non-FITS table file from FluxPoints overwritable

### LINKED ISSUE #5253: FluxPoints.write() is ignoring overwrite when file extension is not FITS
**Gammapy version**

1.2

**Bug description**

When writing a `FluxPoints` table (_.ecsv_) to file it is not possible to overwrite it.

**Expected behavior**

The `overwrite` argument should be taken into account.

**To Reproduce**

From any analysis get a `FluxPoints` instance (in my case it was from obtained from `LightCurveEstimator.run()`)
and call the `write()` method with `overwrite=True`.

You will get the following error,

`OSError: File xxxxxx.ecsv already exists. If you mean to replace it then use the argument "overwrite=True".`

**Other information**

The bug is here: https://github.com/gammapy/gammapy/blob/2475a006e02ac1497eccc362de61e68a5f7a10eb/gammapy/estimators/points/core.py#L270

It should be sufficient to pass the `overwrite` argument to `table.write()`, because the default (inherited from `astropy.table.Table.write()`) is `False`.

HealthyPear:
**Description**

See #5253.

Propagated `astropy.table.Table.write()` overwrite to `FluxPoints.write()` when file extension is not `.fits`.

This pull request closes #5253.

### DIFF
--- a/gammapy/estimators/points/core.py
+++ b/gammapy/estimators/points/core.py
@@ -267,7 +267,7 @@ class FluxPoints(FluxMaps):
         table = self.to_table(sed_type=sed_type, format=format)
 
         if ".fits" not in filename.suffixes:
-            table.write(filename)
+            table.write(filename, overwrite=overwrite)
             return
 
         primary_hdu = fits.PrimaryHDU()
--- a/gammapy/estimators/points/tests/test_core.py
+++ b/gammapy/estimators/points/tests/test_core.py
@@ -211,12 +211,14 @@ class TestFluxPoints:
         assert str(flux_points) == str(actual)
 
     def test_write_ecsv(self, tmp_path, flux_points):
+        filename = tmp_path / "flux_points.ecsv"
+        filename.touch()
         flux_points.write(
-            tmp_path / "flux_points.ecsv",
+            filename,
             sed_type=flux_points.sed_type_init,
             overwrite=True,
         )
-        actual = FluxPoints.read(tmp_path / "flux_points.ecsv")
+        actual = FluxPoints.read(filename)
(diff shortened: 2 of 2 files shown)


registerrier (approved):
Thanks @HealthyPear . This looks good. Could you maybe check why the test is not failing with the current version?

HealthyPear:
Sure. Could you point me to the test function? I am not familiar with the test suite of this project.

registerrier:
> Sure. Could you point me to the test function? I am not familiar with the test suite of this project.

Here it is:
https://github.com/gammapy/gammapy/blob/2475a006e02ac1497eccc362de61e68a5f7a10eb/gammapy/estimators/points/tests/test_core.py#L213

HealthyPear:
I think the problem is that you are writing that file in the `tmp_path`, but you never stored it there before, you load it from `"$GAMMAPY_DATA/tests/spectrum/flux_points/`, which is not the same path, so there is not overwriting in the first place.

We should either copy the  original file in `tmp_path` (or maybe even touch a dummy one) and only then attempt to write the table file.

registerrier:
> We should either copy the original file in `tmp_path` (or maybe even touch a dummy one) and only then attempt to write the table file.

touching a dummy file with the same name sounds good to me.

registerrier:
This cannot be backported to v1.0.x because export to non-FITS file was not available in v1.0.

https://github.com/gammapy/gammapy/blob/547b342cd5327f851a85899abb4a8ab32fac0809/gammapy/estimators/points/core.py#L174-L179

PR #5255, opened 2024-05-07
Expose `FitResult`

### LINKED ISSUE #5154: Exposing `FitResult` ?
Should we expose the `FitResult` class ? 

I think it would be a good idea. It could help debugging and understanding why a fit failed.

Astro-Kirsty:
**Description**
This PR is to resolve #5154

### DIFF
--- a/gammapy/modeling/fit.py
+++ b/gammapy/modeling/fit.py
@@ -2,11 +2,9 @@
 import html
 import itertools
 import logging
-import os
 import numpy as np
 from astropy.table import Table
 from gammapy.utils.pbar import progress_bar
-from gammapy.utils.scripts import make_path
 from .covariance import Covariance
 from .iminuit import (
     confidence_iminuit,
@@ -17,7 +15,7 @@ from .iminuit import (
 from .scipy import confidence_scipy, optimize_scipy
 from .sherpa import optimize_sherpa
 
-__all__ = ["Fit", "FitResult"]
+__all__ = ["Fit", "FitResult", "OptimizeResult", "CovarianceResult"]
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


bkhelifi:
Because this PR expose this class, I am wondering whether one should not describe the content of the FitResult class in the main docstring..

gammapy/modeling/fit.py
```diff
@@ -721,6 +723,24 @@ def covariance_result(self):
         """Optimize result."""
         return self._covariance_result
 
+    def write(self, filename, overwrite=False):
```
registerrier, line 726: I think that writing the `FitResult` means exporting the `Models` and the covariance matrix plus the actual information on the fit result. Not just the string.
This could be a yaml file exactly like `Models.to_yaml()` with additional entries for `nfev`, `success`, `message` etc.

registerrier (commented):
Thanks @Astro-Kirsty . 

Yes the docstring should give the correct info. This probably means that `OptimizeResult` and `CovarianceResult` need to be exposed too...

Regarding writing the `FitResult` to disc, we should check that we can fit everything in a yaml file. I suppose this is the easiest solution for this. This might be a second PR.

gammapy/modeling/fit.py
```diff
@@ -25,10 +25,6 @@ class Registry:
 
     Gives users the power to extend from their scripts.
     Used by `Fit` below.
-
-    Not sure if we should call it "backend" or "method" or something else.
-    Probably we will code up some methods, e.g. for profile analysis ourselves,
-    using scipy or even just Python / Numpy?
```
QRemy, line 31: Why did you remove this note ?
I would suggest to keep it and have an other PR to decide what to do this. 
We may have to change the name as we have already have a registry class in `gammapy/utils/registry.py`.
Also this `Registry` is not public API so the sentence above that say it allows user to extend supported backends is not really true.

QRemy (commented):
Thanks @Astro-Kirsty, I have only a minor comment about the change in the Registry class :

QRemy (approved):

registerrier (approved):

PR #5257, opened 2024-05-08
Pass `events.time_ref` to `observation.pointing.to_fits_header`in `observation.write`

### LINKED ISSUE #5256: Writing `EventList` with `Observation.write` set `MJDREFI` and `MJDREFF` to 0
When one write an `EventList` with `Observation.write`, `MJDREFI` and `MJDREFF` are set to 0 in the header of the `EventList`. 
This obviously jeopardize the time of the written `EventList`.

This behaviour originates from updating the header of the `EventList` with pointing information without passing the reference time: 
https://github.com/gammapy/gammapy/blob/2475a006e02ac1497eccc362de61e68a5f7a10eb/gammapy/data/observations.py#L581
Therefore, https://github.com/gammapy/gammapy/blob/2475a006e02ac1497eccc362de61e68a5f7a10eb/gammapy/utils/time.py#L25 is used.

MRegeard:
This PR fix #5256. 

I pass the time reference of the eventlist to `pointing.to_fits_header` in `observation.write` so that the MJDREFF and MJDREFI written in the header of the eventlist is not set to 0 but to the good value.

### DIFF
--- a/gammapy/data/observations.py
+++ b/gammapy/data/observations.py
@@ -578,7 +578,9 @@ class Observation:
         events = self.events
         if events is not None:
             events_hdu = events.to_table_hdu(format=format)
-            events_hdu.header.update(self.pointing.to_fits_header())
+            events_hdu.header.update(
+                self.pointing.to_fits_header(time_ref=events.time_ref)
+            )
             hdul.append(events_hdu)
 
         gti = self.gti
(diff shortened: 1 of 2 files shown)


MRegeard:
I think it also need a backport in 1.1.x .

registerrier (approved):
Thanks @MRegeard . This looks good. Is the bug affecting v1.0.2? Does this need backporting to v1.0.x too?

v1.1.x is no longer maintained. No backport there.

bkhelifi (approved):
Thanks a lot @MRegeard

PR #5258, opened 2024-05-10
Add write function to `FitResult`

Astro-Kirsty:
This PR is to include a write function in the `FitResult` and partly resolves https://github.com/gammapy/gammapy/issues/5154

The following is implemented
https://github.com/gammapy/gammapy/pull/5255#discussion_r1592789236

Not sure why 'total_stat' is not saving as a value but instead like [this](https://github.com/gammapy/gammapy/assets/60767360/4a2fa8dc-81a3-4124-871e-701fd7933f69)

### DIFF
--- a/gammapy/modeling/fit.py
+++ b/gammapy/modeling/fit.py
@@ -733,7 +733,7 @@ class FitResult:
         )
 
         return {
-            "components": models_dict,
+            "models": models_dict,
             "covariance_result": {
                 "backend": self.covariance_result.backend,
                 "method": self.covariance_result.method,
--- a/gammapy/modeling/fit.py
+++ b/gammapy/modeling/fit.py
@@ -2,9 +2,11 @@
 import html
 import itertools
 import logging
+import os
 import numpy as np
 from astropy.table import Table
 from gammapy.utils.pbar import progress_bar
+from gammapy.utils.scripts import make_path
 from .covariance import Covariance
 from .iminuit import (
     confidence_iminuit,
@@ -720,6 +722,24 @@ class FitResult:
     def covariance_result(self):
         """Optimize result."""
         return self._covariance_result
+
+    def write(self, filename, overwrite=False):
+        """Write to file.
+
+        Parameters
+        ----------
+        filename : str
+            Input filename.
+        overwrite : bool, optional
+            Overwrite existing file. Default is False.
+        """
+        path = make_path(filename)
+        if os.path.isfile(path) and not overwrite:
+            log.warning("File already exits, and overwrite is False")
+        else:
+            with open(path, "w") as file:
+                string = self.__str__()
+                file.write(string)
 
     def __str__(self):
         string = ""
--- a/gammapy/modeling/fit.py
+++ b/gammapy/modeling/fit.py
@@ -2,9 +2,11 @@
 import html
 import itertools
 import logging
-import os
 import numpy as np
 from astropy.table import Table
+import yaml
+from gammapy.utils.check import add_checksum
+from gammapy.utils.metadata import CreatorMetaData
 from gammapy.utils.pbar import progress_bar
 from gammapy.utils.scripts import make_path
 from .covariance import Covariance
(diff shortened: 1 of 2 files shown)


QRemy (changes requested):
Thanks @Astro-Kirsty, could you add a test for that ?
How do you produce the bad outpout for total_stat ?
I tried with the analysis_2 toturial and the output seems fine.

gammapy/modeling/fit.py
```diff
@@ -721,6 +725,75 @@ def covariance_result(self):
         """Optimize result."""
         return self._covariance_result
 
+    def to_dict(self, full_output=False, overwrite_templates=False):
+        """Convert to dictionary."""
```
bkhelifi, line 729: Should we describe here the arguments into the docstring?

gammapy/modeling/fit.py
```diff
@@ -721,6 +725,75 @@ def covariance_result(self):
...
+                "method": self.optimize_result.method,
+                "success": self.optimize_result.success,
+                "message": self.optimize_result.message,
+                "nfev": self.optimize_result.nfev,
+                "total_stat": self.optimize_result._total_stat,
+            },
+        }
+
+    def to_yaml(self, full_output=False, overwrite_templates=False):
+        """Convert to YAML string."""
```
bkhelifi, line 754: The same comment for the arguments

bkhelifi (approved):
I have some minor comments. This PR is globally ready to be merged! Thanks you a lot

bkhelifi:
Just as a comment, some information stored as data are in fact metadata, e.g. 
```
            "optimize_result": {
                "backend": self.optimize_result.backend,
                "method": self.optimize_result.method,

```
I propose to fix that later on a separated PR about metadata handling.

Astro-Kirsty:
> Thanks @Astro-Kirsty, could you add a test for that ? How do you produce the bad outpout for total_stat ? I tried with the analysis_2 toturial and the output seems fine.

Using this [code snippet](https://github.com/gammapy/gammapy/assets/60767360/572eef2b-78da-4bc0-9ed0-fca0b593d0d7)

The print is fine, but then the output in the yaml file is not

gammapy/modeling/fit.py
Astro-Kirsty: For both this and the `to_yaml` I did not include them because it not done in other instances of the code for these functions

gammapy/modeling/fit.py
```diff
@@ -721,6 +725,75 @@ def covariance_result(self):
...
 
+    def to_dict(self, full_output=False, overwrite_templates=False):
+        """Convert to dictionary."""
+
+        models_dict = self.models.to_dict(
+            full_output=full_output, overwrite_templates=overwrite_templates
+        )
+
+        return {
+            "components": models_dict,
```
QRemy, line 736:
```suggestion
            "models": models_dict,
```

QRemy (approved):
Thanks @Astro-Kirsty, the yaml file written by the test seems ok.

gammapy/modeling/fit.py
```diff
@@ -721,6 +725,75 @@ def covariance_result(self):
...
         return self._covariance_result
 
+    def to_dict(self, full_output=False, overwrite_templates=False):
+        """Convert to dictionary."""
+
+        models_dict = self.models.to_dict(
+            full_output=full_output, overwrite_templates=overwrite_templates
+        )
+
+        return {
```
registerrier, line 735: Maybe start with `optimize_results` then `covariance_result` and finally `models`.

gammapy/modeling/fit.py
```diff
@@ -721,6 +725,75 @@ def covariance_result(self):
...
+            "models": models_dict,
+            "covariance_result": {
```
registerrier, line 737: What happens if covariance has not been run?

registerrier (commented):
Thnaks @Astro-Kirsty . My review is a bit late.
Some general comments:
- What about a read method? The point of the write method is to be able to retrieve the result later.
- I think we would need a more generic scheme to be able to handle different steps. I would suggest to implement a `to_dict()` method on the `FitStepResult`.
- There is code duplication between this `write` and `to_yaml` methods and those implemented on `Models.write` and `Datasets.write`. Maybe we should improve `gammapy.utils.scripts.write_yaml` to support all these cases.

registerrier:
> Just as a comment, some information stored as data are in fact metadata, e.g.
> 
> ```
>             "optimize_result": {
>                 "backend": self.optimize_result.backend,
>                 "method": self.optimize_result.method,
> ```
> 
> I propose to fix that later on a separated PR about metadata handling.

`FitResult` is not meant to be a data product. I don't think we should try to have a scheme of data vs metadata here.

QRemy:
> * What about a read method? The point of the write method is to be able to retrieve the result later.

Do we really need a read method for objects that are only used for diagnostics and not configuration ?

PR #5264, opened 2024-05-16
Introduce COPY_IF_NEEDED to allow compatibility with numpy 2.0

registerrier:
**Description**

This pull request makes gammapy compatible with numpy 2.0 while keeping backward compatibility. 
The main issue is connected with the `np.array` copy semantics . See https://github.com/numpy/numpy/pull/25168
`Quantity(copy=False)` is now strictly interpreted and will fail if a copy is unavoidable. Because the behavior has been fully ported in astropy. See https://github.com/astropy/astropy/pull/16142. 
This creates a hundred fails in the devdeps CI. 
This is solved in a backward compatible manner by following the same approach as https://github.com/astropy/astropy/issues/16167 introducing a global variable `COPY_IF_NEEDED` set to `False` if numpy<2.0 and `None` otherwise.

An additional issue appeared in `WcsNDMap.to_nd_region_map_histogram` probably because units aren't propagated the same way with `np.apply_along_axis`. The backward compatible solution is to tear the units out before applying it.

**Dear reviewer**

### DIFF
--- a/gammapy/data/pointing.py
+++ b/gammapy/data/pointing.py
@@ -18,6 +18,7 @@ from astropy.io import fits
 from astropy.table import Table
 from astropy.units import Quantity
 from astropy.utils import lazyproperty
+from gammapy.utils.compat import COPY_IF_NEEDED
 from gammapy.utils.deprecation import GammapyDeprecationWarning
 from gammapy.utils.fits import earth_location_from_dict, earth_location_to_dict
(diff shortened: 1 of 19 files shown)


adonath (approved):
Thanks a lot @registerrier, I have no comments!

registerrier:
Seeing green light again in CI is a pleasant sight :)

PR #5269, opened 2024-05-17
Add leakage protection factor in the Timmer Konig algorithm

cgalelli:
This PR adds a small modification to the TimmerKonig algorithm. Now there is the possibility of using an additional (optional) input which is a multiplicative factor of the time series length `npoints` to avoid spectral leakage. The final time series of length `npoints` is extracted randomly at the end from the longer time series of length `npoints * leakage_protection`.
A specific test was added.

### DIFF
--- a/gammapy/stats/tests/test_variability.py
+++ b/gammapy/stats/tests/test_variability.py
@@ -164,11 +164,16 @@ def test_tk():
         temp, 15, 1 * u.h, power_spectrum_params=params
     )
 
+    time_series4, time_axis4 = TimmerKonig_lightcurve_simulator(
+        lambda x: x ** (-3), 20, 1 * u.s, leakage_protection=100
+    )
+
     assert len(time_series) == 20
     assert isinstance(time_axis, u.Quantity)
     assert time_axis.unit == u.s
     assert len(time_series2) == 21
     assert len(time_series3) == 15
+    assert len(time_series4) == 20
 
 
 def test_structure_function():
--- a/gammapy/stats/tests/test_variability.py
+++ b/gammapy/stats/tests/test_variability.py
@@ -165,7 +165,7 @@ def test_tk():
     )
 
     time_series4, time_axis4 = TimmerKonig_lightcurve_simulator(
-        lambda x: x ** (-3), 20, 1 * u.s, leakage_protection=100
+        lambda x: x ** (-3), 20, 1 * u.s, leakage_protection=100.0
     )
 
     assert len(time_series) == 20
@@ -174,6 +174,18 @@ def test_tk():
     assert len(time_series2) == 21
     assert len(time_series3) == 15
     assert len(time_series4) == 20
+
+
+def test_tk_badleakage():
+    with pytest.raises(ValueError):
+        TimmerKonig_lightcurve_simulator(
+            lambda x: x ** (-3), 20, 1 * u.s, leakage_protection=0.1
+        )
+
+    with pytest.raises(Warning):
+        TimmerKonig_lightcurve_simulator(
+            lambda x: x ** (-3), 20, 1 * u.s, leakage_protection=1.001
+        )
 
 
 def test_structure_function():
--- a/gammapy/stats/variability.py
+++ b/gammapy/stats/variability.py
@@ -383,6 +383,12 @@ def TimmerKonig_lightcurve_simulator(
     fourier_coeffs = np.insert(fourier_coeffs, 0, 0)
     time_series = np.fft.ifft(fourier_coeffs).real
 
+    if leakage_protection > 1:
+        extract = random_state.randint(
+            npoints - 1, leakage_protection * npoints - npoints
+        )
+        time_series = np.take(time_series, range(extract, extract + npoints))
+
     tmax = time_series.max()
     if tmax < np.abs(time_series.min()):
         time_series = -time_series
@@ -390,11 +396,5 @@ def TimmerKonig_lightcurve_simulator(
     time_series = 0.5 + 0.5 * time_series / tmax
 
     time_axis = np.linspace(0, npoints * spacing.value, npoints) * spacing.unit
-
-    if leakage_protection > 1:
-        extract = random_state.randint(
-            npoints - 1, leakage_protection * npoints - npoints
-        )
-        time_series = np.take(time_series, range(extract, extract + npoints))
 
     return time_series, time_axis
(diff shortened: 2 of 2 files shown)


gammapy/stats/variability.py
```diff
@@ -303,6 +304,8 @@ def TimmerKonig_lightcurve_simulator(
         Number of points in the output time series.
     spacing : `~astropy.units.Quantity`
         Sample spacing, inverse of the sampling rate. The units are inherited by the resulting time axis.
+    leakage_protection : int
```
bkhelifi, line 307: Add the used default value in the docstring and precise that it should be larger than 1
I think also that using a float is more efficient for this purpose, there is no need to create artificially a too large number of bins.... If accepted, then a default value of 1.2 or 1.5 might be wise (maybe some tests are needed to derive a 'reasonable' default value)

gammapy/stats/tests/test_variability.py
```diff
@@ -164,11 +164,16 @@ def temp(x, norm, index):
         temp, 15, 1 * u.h, power_spectrum_params=params
     )
 
+    time_series4, time_axis4 = TimmerKonig_lightcurve_simulator(
```
bkhelifi, line 167: Looking at the code, there is a `if leakage_protection > 1`. So, should we have 2 different tests?

bkhelifi (changes requested):
Well seen! This feature is numerically important... I give here some comments

gammapy/stats/variability.py
```diff
@@ -289,6 +289,7 @@ def TimmerKonig_lightcurve_simulator(
     power_spectrum,
     npoints,
     spacing,
+    leakage_protection=1.0,
```
bkhelifi, line 292: Why the default is 1, and not 1.5 or 1.3?

gammapy/stats/variability.py
```diff
@@ -303,6 +304,8 @@ def TimmerKonig_lightcurve_simulator(
         Number of points in the output time series.
     spacing : `~astropy.units.Quantity`
         Sample spacing, inverse of the sampling rate. The units are inherited by the resulting time axis.
+    leakage_protection : float, optional
+        Factor by which to multiply the length of the time series to avoid red noise leakage.
```
bkhelifi, line 308: Precise the value of the default value in the docstring...

bkhelifi (changes requested):
Some last comments...

gammapy/stats/variability.py
```diff
@@ -346,7 +349,7 @@ def TimmerKonig_lightcurve_simulator(
 
     random_state = get_random_state(random_state)
 
-    frequencies = np.fft.fftfreq(npoints, spacing.value)
+    frequencies = np.fft.fftfreq(int(npoints * leakage_protection), spacing.value)
```
QRemy, line 352: maybe define `npoints_extended= int(npoints * leakage_protection)` before,
check that `if leakage_protection!=1: npoints_extended>npoints`, 
and if not send a warning or force `npoints_extended = npoints+1`

QRemy (commented):
Thanks @cgalelli,  is it better to have this as a multiplicative factor rather than an additive number of margin points ?
I left a comment about the case where  leakage_protection<1, or >1 but too small to add any points when converted to int :

gammapy/stats/variability.py
cgalelli: I think the default behavior should be to simulate the normal lightcurve, both for time efficiency and for consistency. A useful size of the multiplier could be 10 or 100.

QRemy (dismissed):
Thanks @cgalelli  no further comment.

bkhelifi (dismissed):
Thanks a lot @cgalelli

gammapy/stats/variability.py
```diff
@@ -346,7 +349,15 @@ def TimmerKonig_lightcurve_simulator(
...
     random_state = get_random_state(random_state)
 
-    frequencies = np.fft.fftfreq(npoints, spacing.value)
+    if leakage_protection < 1.0:
+        raise ValueError("The leakage protection factor must be at least 1.0")
+
+    npoints_extended = int(npoints * leakage_protection)
+    if leakage_protection > 1.0 and npoints_extended == npoints:
+        raise Warning(
+            "The extended number of points is the same length as the desired number"
```
Astro-Kirsty, line 358: Is there anything more that could be added here i.e when would this occur and how to fix that?

bkhelifi: Good point @Astro-Kirsty. The if condition is True when the leakage_protection is too small. User should increase it or we just add 1 new point instead of raising an error;)

cgalelli: I don't think raising an error or forcing a new point would be the correct thing to do. 
Let's say that some user has an automated tool which returns that the optimal factor is 1.01 or something similar, which then for short enough series results in the same amount of points. This in general means that the original algorithm would not result in noise leakage (typical factors could be 5-100 if there is leakage). In this case I would simply throw the warning up, so that the user is aware and can either increase the factor himself or leave as is because he is confident that there is no leakage.
I could add a more detailed warning description

Astro-Kirsty: If you choose to raise the warning, then I suggest to improve the message a little bit.

bkhelifi (dismissed):

gammapy/stats/variability.py
```diff
@@ -346,7 +349,16 @@ def TimmerKonig_lightcurve_simulator(
 
     random_state = get_random_state(random_state)
 
-    frequencies = np.fft.fftfreq(npoints, spacing.value)
+    if leakage_protection < 1.0:
```
registerrier, line 352: Could you add a test for this?

gammapy/stats/variability.py
```diff
@@ -346,7 +349,16 @@ def TimmerKonig_lightcurve_simulator(
...
+        raise ValueError("The leakage protection factor must be at least 1.0.")
+
+    npoints_extended = int(npoints * leakage_protection)
+    if leakage_protection > 1.0 and npoints_extended == npoints:
+        raise Warning(
```
registerrier, line 357: This should be tested as well.

registerrier (commented):
Thanks @cgalelli . If exception and warning are important features, they should be tested for so that we don't remove the functionality in further updates of the code.

gammapy/stats/variability.py
```diff
@@ -346,7 +349,16 @@ def TimmerKonig_lightcurve_simulator(
...
+            "The extended time series is the same length as the final desired time series."
+            "To have an effective protection from noise leakage and avoid aliasing, increase <leakage_protection>."
```
QRemy, line 359:
```suggestion
            "To have an effective protection from noise leakage and avoid aliasing, increase the leakage_protection option."
```

Astro-Kirsty (approved):

Issue #5261, opened 2024-05-15
Consistency between MapDataset.stack and Datasets.stack_reduce


registerrier:
Currently `Datasets.stack_reduce()` will first apply the `to_masked()` method to the first dataset. 
Using `MapDataset.stack` the mask is not applied to `self`, the assumption being that the first dataset used is empty. 
While this is probably true in general, this might lead to some issues. At least the docstring should more clearly state this.
Note also that `MapDatasetOnOff.stack` applies the mask to  `self.counts_off`, which is then inconsistent.
https://github.com/gammapy/gammapy/blob/011cefef8916ebc9584fa06f52d999ca01af78ff/gammapy/datasets/map.py#L2601
 
See associated PRs : #3058 #3164

Is there any reason for this specific behavior?

AtreyeeS:
If I remember correctly, the behaviour in `MapDataset.stack` is intentional. It was the cleanest way to be able to stack maps during the data reduction process while handling the safe mask correctly.
This is included in the Note here https://docs.gammapy.org/1.2/user-guide/datasets/index.html#stacking-multiple-datasets - "To properly handle masks, it is necessary to stack onto an empty dataset."
`Datasets.stack_reduce` was introduced to avoid users stacking onto non-empty maps. The idea was general users should only use `Datasets.stack_reduce` and not `MapDataset.stack`. I guess we did not expose these consistently.

Not sure what is happening in `MapDatasetOnOff`.

adonath:
> Using MapDataset.stack the mask is not applied to self, the assumption being that the first dataset used is empty.

If I remember the first implementation of `MapDataset.stack()` had the in-place application of the mask. This lead to a lot of complex code, because one needed to apply the mask to all quantities in the dataset. The same code was then repeated for the "other" dataset. It was much simpler to introduce the `.to_masked()` method, which all the mask application in a single place. 

Applying the mask for in-place stacking would mean to always apply it, because one cannot know whether the stacking happens the first time. This means, except for the first time, one would un-necessarily apply the (stacked) mask over and over again.

But independently, `stack_reduce()` is the higher level method anyway. So I think there is no need for consistency. However the behavior of the on-off dataset should be fixed.

registerrier:
OK thanks. Not re-masking all the time seems a reasonable approach indeed. While solving #5245 , I try to adapt the docstring and the behaviour of `MapDatasetOnOff.stack()`

Issue #5245, opened 2024-05-03
Acceptance is not properly propagated during MapDatasetOnOff stacking 


registerrier:
**Gammapy version**
v1.0 and beyond

**Bug description**
The ring background analysis that is demonstrated in the corresponding tutorial is incorrect. The OFF is correlated twice (one with the `RingBackgroundMaker` and the other with the `ExcessMapEstimator` because the latter use incorrectly the default value `correlate_off=True`. The estimated significance is therefore artificially large.

Wile using instead `correlate_off=False` removes the double OFF counting, it creates residual features where the acceptance is discontinuous (mostly because of FoV masking).

**Expected behavior**
`ExcessMapEstimator` should provide images without artifacts using the proper `correlate_off=False` and the tutorial should be updated accordingly.

**To Reproduce**
Modify the ring background tutorial to include an offset-max cut and add the correct option to compute Li-Ma excess and significance map `correlate=False`.

```
# Cutout size (for the run-wise event selection)
config.datasets.geom.selection.offset_max = 3.0 * u.deg
config.datasets.safe_mask.parameters = {"offset_max" : 1.5*u.deg}
config.datasets.safe_mask.methods = ["offset-max"]
```

You will obtain this figure:
![download-5](https://github.com/gammapy/gammapy/assets/5047091/0014c91c-5a11-4b83-aeed-d8d6cdc31428)

**Other information**
Any other information you think will be useful for us to fix the issue can go here.

registerrier:
The issue most likely comes from the way we compute stacked acceptances.

#### Reminder, for excess calculation with ON-OFF data:
- for a single dataset we have $N_{excess}$ and $\alpha$:
	$$N_{excess} = N_{ON} - \alpha N_{OFF} $$
- with  $\alpha = \frac{Acc_{ON}}{Acc_{OFF}}$ the ratio of ON and OFF acceptances
- to guarantee correct number of excess when stacking one has to compute an $N_{OFF}$-weighted $\alpha$, such that:
	$$ \sum_i N_{excess,i} = \sum_i N_{ON,i} - \bar{\alpha} \sum_i{N_{OFF,i}}$$
	- with:
	$$\bar{\alpha} = \frac{\sum_i \alpha_i N_{OFF,i}}{\sum_i N_{OFF,i}}$$

#### Usage in Gammapy 
-  This principle is applied :
	- when stacking ON-OFF datasets
	- when computing the correlated maps in `ExcessMapEstimator`

#### A caveat	
[...]

PR #5270, opened 2024-05-17
Modify acceptance stacking behavior

registerrier:
**Description**

This pull request tries to solve #5245 by modifying the way `acceptance` is stacked for `MapDatasetOnOff`. 
Now, instead of assuming the total acceptance to be equal to one, the `acceptance` is properly stacked. Because `alpha`
is unchanged `acceptance_off` is therefore different now.

The PR also corrects an inconsistency noted in #5261 . `MapDataset.stack` does not apply the mask of `self` when stacking. This avoids recomupting empty stacks all the time where in practice the first dataset is always empty. 
The `MapDataset.OnOff` had a different assumption. This is now made consistent.

The docstrings where previously incorrect and have been modified. The formula are probably still incorrect. 

**Dear reviewer**

This is ready, but docstrings need to be checked. Also the tutorial for Ring Background should be updated.


samanthalwong (approved):
Works well in resolving #5245 for RBM analysis

gammapy/datasets/map.py
```diff
@@ -2572,8 +2575,16 @@ def _is_stackable(self):
...
+        Safe mask is applied to the other dataset to compute the stacked counts data.
+        Counts outside the safe mask are lost.
+
+        The ``acceptance`` of the stacked dataset is obtained by stacking the acceptance weighted
+        by the other mask_safe onto the current unwieghted acceptance.
+
+        Note that the masking is not applied to the current dataset. If masking needs
+        to be applied to it, use `~gammapy.MapDataset.to_masked()` first.
+
+        The stacked ``acceptance_off`` is scaled so that:
```
bkhelifi, line 2587: Maybe one can put below the formulae for both acceptance, adding the use of the masks (in the `MapDataset.stack' docstring, one uses \epsilon for the mask

gammapy/datasets/map.py
```diff
@@ -2572,8 +2575,16 @@ def _is_stackable(self):
     def stack(self, other, nan_to_num=True):
         r"""Stack another dataset in place.
 
-        The ``acceptance`` of the stacked dataset is normalized to 1,
-        and the stacked ``acceptance_off`` is scaled so that:
+        Safe mask is applied to the other dataset to compute the stacked counts data.
+        Counts outside the safe mask are lost.
```
bkhelifi, line 2579: I propose to add the formula, as in `MapDataset.stack'

bkhelifi (changes requested):
Thanks a lot. I have just few suggestions for the docstrings

gammapy/datasets/map.py
Astro-Kirsty: We could instead put that information here https://docs.gammapy.org/1.2/user-guide/makers/ring.html ? 
But, no strong opinion on if it should also be in the docstring

samanthalwong: As a user, I think it would definitely be helpful to have the formulae in the user guide (and maybe linked in the docs). Having it in the docstring might be nice too.

AtreyeeS (approved):
Thanks @registerrier ! I think the tutorial can be adapted in a separate PR

gammapy/datasets/map.py
QRemy, line 2579:
```suggestion
        Safe mask is applied to the other dataset to compute the stacked counts data,
        counts outside the safe mask are lost (as for `~gammapy.MapDataset.stack`).
```

QRemy (approved):
Thanks @registerrier, no further comment, merging now.

PR #5271, opened 2024-05-17
Allow extrapolation only along spatial dimension in `mask_safe_edisp`

QRemy:
Support for extrapolation in `interp_to_geom` if geom is mask.
Allow extrapolation only along spatial dimension in `mask_safe_edisp`
to support case where `mask_safe` geom and irfs geom are different.

This a better  solution to resolve the problem described in #5236.

### DIFF
--- a/gammapy/maps/tests/test_core.py
+++ b/gammapy/maps/tests/test_core.py
@@ -437,6 +437,24 @@ def test_interp_to_geom():
     assert isinstance(interp_wcs_map, WcsNDMap)
     assert interp_wcs_map.geom == wcs_geom_target
 
+    # WcsNDMap is_mask
+    geom_wcs = WcsGeom.create(
+        npix=(5, 3), proj="CAR", binsz=0.1, axes=[energy], skydir=(0, 0)
+    )
+
+    wcs_map = Map.from_geom(geom_wcs, unit="", data=True)
+    wcs_geom_target = WcsGeom.create(
+        skydir=(30, 30),
+        width=(10, 10),
+        binsz=0.1 * u.deg,
+        axes=[energy_target],
+        frame="galactic",
+    )
+    interp_wcs_map = wcs_map.interp_to_geom(
+        wcs_geom_target, method="linear", fill_value=None
+    )
+    assert np.all(interp_wcs_map.data)
+
     # HpxNDMap
     geom_hpx = HpxGeom.create(binsz=60, axes=[energy], skydir=(0, 0))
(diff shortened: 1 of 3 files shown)


bkhelifi:
Hi @QRemy,
I propose to add in the docstring of `interp_to_geom' something like "If None, values outside the domain are extrapolated", as made in `interp_by_pix'

Astro-Kirsty (dismissed):
Thanks @QRemy. This is a nice solution.

gammapy/maps/tests/test_core.py
```diff
@@ -437,6 +437,24 @@ def test_interp_to_geom():
...
     assert interp_wcs_map.geom == wcs_geom_target
 
+    # WcsNDMap is_mask
+    geom_wcs = WcsGeom.create(
+        npix=(5, 3), proj="CAR", binsz=0.1, axes=[energy], skydir=(0, 0)
+    )
+
+    wcs_map = Map.from_geom(geom_wcs, unit="", data=True)
+    wcs_geom_target = WcsGeom.create(
+        skydir=(30, 30),
```
bkhelifi, line 447: the  map from geom_wcs and the one from wcs_geom_target are not overlapping, is not it? If yes, can you precise the goal of this test?

QRemy: yes that's the use case I want to test
for example in HAWC you have the PSF define in icrs frame with  1 bin in lon
```
WcsGeom

	axes       : ['lon', 'lat', 'rad', 'energy']
	shape      : (1, 36, 500, 11)
	ndim       : 4
	frame      : icrs
	projection : CAR
	center     : 0.0 deg, 2.5 deg
	width      : 360.0 deg x 180.0 deg
	wcs ref    : 0.0 deg, 2.5 deg
```
that won't necessarily overlap with the geom of your analysis region
```
WcsGeom

	axes       : ['lon', 'lat', 'energy']
	shape      : (80, 80, 11)
	ndim       : 3
	frame      : galactic
	projection : CAR
	center     : 6.8 deg, -4.8 deg
	width      : 4.8 deg x 4.8 deg
	wcs ref    : 120.0 deg, 0.0 deg
```
you can still interpolated the PSF at the position of the source within the analysis geom but the current behaviour of `self.mask_safe.interp_to_geom(geom)` does not allow to interpolated the mask_safe at the position of the IRF geom.

bkhelifi: Perfect, this test then makes sense.

bkhelifi (approved):
Many thanks @QRemy

PR #5275, opened 2024-05-22
Add a notebook for EBL correction example

AtreyeeS:
This PR adds a notebook to show how to correct for EBL absorption. 
The main idea is to show how to correctly set the covariance as discussed in #5218 

I decided to add a new notebook because it gets difficult to find otherwise. I am trying to see if, for this specific case, we can give a way to get the correct covariance for a free alpha/redshift

PR #5279, opened 2024-05-31
Add copy method on FluxMaps

QRemy:
Add copy method on FluxMaps


gammapy/estimators/map/core.py
```diff
@@ -1093,6 +1105,28 @@ def read(cls, filename, checksum=False):
...
+            of a point source with a power law spectrum of index 2 is assumed.
+
+        Returns
+        -------
+        flux_maps : `FluxMaps`
+            Copied flux maps object.
+
+        """
+
+        new = deepcopy(self)
```
bkhelifi, line 1125: Why are you not calling the copy function of Map here?

AtreyeeS (approved):
Thanks @QRemy !
No comments

gammapy/estimators/map/core.py
QRemy: The FluxMaps does not inherit of Map

gammapy/estimators/map/core.py
```diff
@@ -1093,6 +1105,28 @@ def read(cls, filename, checksum=False):
...
             return cls.from_hdulist(hdulist, checksum=checksum)
 
+    def copy(self, reference_model=None):
+        """Deep copy
+
+        Parameters
+        ----------
+
+        reference_model : `~gammapy.modeling.models.SkyModel`, optional
+            The reference model to use for conversions. If None, a model consisting
```
AtreyeeS, line 1115: The docstring is not correct here. If None, the reference model on self is copied

gammapy/estimators/map/core.py
```diff
@@ -1093,6 +1105,28 @@ def read(cls, filename, checksum=False):
...
+            of a point source with a power law spectrum of index 2 is assumed.
```
QRemy, line 1116:
```suggestion
            The reference model to use for conversions. If None, the copied model is used.
```

gammapy/estimators/map/core.py
```diff
@@ -330,6 +331,17 @@ def reference_model(self):
         """Reference model as a `~gammapy.modeling.models.SkyModel`."""
         return self._reference_model
 
+    @reference_model.setter
+    def reference_model(self, reference_model):
+        """Reference model as a `~gammapy.modeling.models.SkyModel`.
+
+        Parameters
+        ----------
+        value : `~gammapy.modeling.models.SkyModel`
```
AtreyeeS, line 340:
```suggestion
        reference_model : `~gammapy.modeling.models.SkyModel`
```

gammapy/estimators/map/core.py
```diff
@@ -1093,6 +1105,27 @@ def read(cls, filename, checksum=False):
...
             return cls.from_hdulist(hdulist, checksum=checksum)
 
+    def copy(self, reference_model=None):
+        """Deep copy
+
+        Parameters
+        ----------
+
+        reference_model : `~gammapy.modeling.models.SkyModel`, optional
+            The reference model to use for conversions. If None, the copied model is used.
```
AtreyeeS, line 1115:
```suggestion
            The reference model to use for conversions. If None, the original model is used.
```

AtreyeeS (dismissed):

AtreyeeS (dismissed):
Thanks! no more comments

gammapy/estimators/map/core.py
```diff
@@ -330,6 +331,17 @@ def reference_model(self):
         """Reference model as a `~gammapy.modeling.models.SkyModel`."""
         return self._reference_model
 
+    @reference_model.setter
```
registerrier, line 334: I would not include a setter for the reference model, as this is can lead to unexpected results if poorly handled.

gammapy/estimators/map/core.py
```diff
@@ -1093,6 +1105,27 @@ def read(cls, filename, checksum=False):
...
+        Returns
+        -------
+        flux_maps : `FluxMaps`
+            Copied flux maps object.
+
+        """
+
+        new = deepcopy(self)
+        if reference_model is not None:
+            new.reference_model = reference_model
```
registerrier, line 1126: Shouldn't one copy the reference model here?

registerrier, line 1126: Here I would log a warning about changing the reference model.

gammapy/estimators/map/core.py
registerrier: I would add a word of caution here. Something like : "Flux maps have been obtained for a specific reference model. Changing it will change the fluxes. Handle with care."

registerrier (commented):
Thanks @QRemy . This is nice but I think we have to be extra-careful in giving warnings to the user. 
In particular, I think that having the `reference_model` setter is not a good idea, as this is something one should do only in very specific situations.

gammapy/estimators/map/core.py
adonath: Yes, I agree. The reference model should read only. Changing the reference model requires adapting the flux norm values. 

However, we could think about an API, where the flux maps can be transferred to a new reference model, such as `FluxMaps.to_reference_model()` to similar..

gammapy/estimators/map/core.py
QRemy: added

gammapy/estimators/map/core.py
```diff
@@ -1093,6 +1094,31 @@ def read(cls, filename, checksum=False):
         ) as hdulist:
             return cls.from_hdulist(hdulist, checksum=checksum)
 
+    def copy(self, reference_model=None):
+        """Deep copy
+
+        Parameters
+        ----------
+
+        reference_model : `~gammapy.modeling.models.SkyModel`, optional
```
adonath, line 1103: If with a new reference model the fluxes are not modified, what is the use case at all?

QRemy: the use case came from here : was suggested here
https://github.com/gammapy/gammapy/pull/5275#discussion_r1621864916

gammapy/estimators/map/core.py
```diff
@@ -1093,6 +1094,31 @@ def read(cls, filename, checksum=False):
...
+            The reference model to use for conversions. If None, the originial model is copied.
+            Flux maps have been obtained for a specific reference model.
+            Changing it will change the fluxes. Handle with care.
+        Returns
+        -------
+        flux_maps : `FluxMaps`
```
bkhelifi, line 1109:
```suggestion
        flux_maps : `~gammapy.estimators.FluxMaps`
```

gammapy/estimators/map/core.py
adonath: Thanks, that make sense! I would maybe just mention the EBL absorption as a valid use case.

PR #5280, opened 2024-05-31
Add min_pix option in WcsGeom.cutout and set it to 3 for IRFs 

QRemy:
Add min_pix option in WcsGeom.cutout and set it to 3 for IRFs in order to ensure that IRF interpolation is done correctly if the binning of the IRFs is larger than the width of the analysis region.

### DIFF
--- a/gammapy/irf/core.py
+++ b/gammapy/irf/core.py
@@ -963,9 +963,11 @@ class IRFMap:
             Cutout IRF map.
         """
 
-        irf_map = self._irf_map.cutout(position, width, mode, min_npix)
+        irf_map = self._irf_map.cutout(position, width, mode, min_npix=min_npix)
         if self.exposure_map:
-            exposure_map = self.exposure_map.cutout(position, width, mode, min_npix)
+            exposure_map = self.exposure_map.cutout(
+                position, width, mode, min_npi=min_npix
+            )
         else:
             exposure_map = None
         return self.__class__(irf_map, exposure_map=exposure_map)
--- a/gammapy/irf/core.py
+++ b/gammapy/irf/core.py
@@ -966,7 +966,7 @@ class IRFMap:
         irf_map = self._irf_map.cutout(position, width, mode, min_npix=min_npix)
         if self.exposure_map:
             exposure_map = self.exposure_map.cutout(
-                position, width, mode, min_npi=min_npix
+                position, width, mode, min_npix=min_npix
             )
         else:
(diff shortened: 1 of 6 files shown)


bkhelifi:
Thanks @QRemy... What happens where there is not enough bin for the interpolation?

QRemy:
> What happens where there is not enough bin for the interpolation?
It uses the single bin that is left in the psf cutout. So if we reduce the region size the psf used at a given position can be different to the one used before cutout.

bkhelifi (approved):
Thanks @QRemy! Looks good for me

gammapy/maps/wcs/ndmap.py
```diff
@@ -992,14 +992,21 @@ def cutout(self, position, width, mode="trim", odd_npix=False):
         odd_npix : bool, optional
             Force width to odd number of pixels.
             Default is False.
+        min_npix : bool, optional
```
registerrier, line 995:
```suggestion
        min_npix : int, optional
```

gammapy/maps/wcs/ndmap.py
```diff
@@ -992,14 +992,21 @@ def cutout(self, position, width, mode="trim", odd_npix=False):
...
+            Force width to a minimmum number of pixels.
```
registerrier, line 996:
```suggestion
            Force width to a minimum number of pixels.
```

registerrier: Maybe specify that the minimum number applies in both axes.

gammapy/irf/core.py
```diff
@@ -958,9 +958,10 @@ def cutout(self, position, width, mode="trim"):
         cutout : `IRFMap`
             Cutout IRF map.
         """
-        irf_map = self._irf_map.cutout(position, width, mode)
+
+        irf_map = self._irf_map.cutout(position, width, mode, min_npix=3)
```
registerrier, line 962: I think the option should be explicitly available on the `cutout` arguments. A minima it should be explained that the minimum cutout is 3 pixels wide.

registerrier (commented):
I think I understand the use case. I just wonder whether it should be configurable as an argument of the `IRFMap.cutout` method. The docstring could briefly explain why 3 is a good default. No?

Does it need to be backported?

gammapy/irf/core.py
QRemy: done

PR #5285, opened 2024-06-03
Add support for stat_scan in TSMapEstimator

QRemy:
Add support for stat_scan in TSMapEstimator with 2 methods:
* "stat_scan_local": estimate likelihood profile around best norm
* "stat_scan": estimate likelihood profile on a fixed norm range

### DIFF
--- a/gammapy/estimators/map/core.py
+++ b/gammapy/estimators/map/core.py
@@ -83,6 +83,9 @@ VALID_QUANTITIES = [
     "counts",
     "success",
     "n_dof",
+    "norm_scan_values",
+    "stat_scan_local",
+    "stat_scan",
 ]
 
 
@@ -98,6 +101,9 @@ OPTIONAL_QUANTITIES_COMMON = [
     "counts",
     "success",
     "n_dof",
+    "norm_scan_values",
+    "stat_scan_local",
+    "stat_scan",
 ]
 
 
@@ -487,6 +493,18 @@ class FluxMaps:
         """Fit statistic scan value."""
         self._check_quantity("stat_scan")
         return self._data["stat_scan"]
+
+    @property
+    def norm_scan_values(self):
+        """Fit statistic norm scan values."""
+        self._check_quantity("norm_scan_values")
+        return self._data["norm_scan_values"]
+
+    @property
+    def stat_scan_local(self):
+        """Fit statistic scan value around best norm."""
+        self._check_quantity("stat_scan_local")
+        return self._data["stat_scan_local"]
 
     @property
     def stat(self):
--- a/gammapy/estimators/map/core.py
+++ b/gammapy/estimators/map/core.py
@@ -79,7 +79,6 @@ VALID_QUANTITIES = [
     "stat_scan",
     "stat_null",
     "stat_scan",
-    "stat_scan_local",
     "dnde_scan_values",
     "niter",
     "is_ul",
@@ -97,7 +96,6 @@ OPTIONAL_QUANTITIES_COMMON = [
     "stat",
     "stat_null",
     "stat_scan",
-    "stat_scan_local",
     "dnde_scan_values",
     "niter",
     "is_ul",
@@ -499,12 +497,6 @@ class FluxMaps:
         """Fit statistic norm scan values."""
         self._check_quantity("dnde_scan_values")
         return self._data["dnde_scan_values"]
-
-    @property
-    def stat_scan_local(self):
-        """Fit statistic scan value around best norm."""
-        self._check_quantity("stat_scan_local")
-        return self._data["stat_scan_local"]
 
     @property
     def stat(self):
--- a/gammapy/estimators/map/core.py
+++ b/gammapy/estimators/map/core.py
@@ -78,14 +78,14 @@ VALID_QUANTITIES = [
     "stat",
     "stat_scan",
     "stat_null",
+    "stat_scan",
+    "stat_scan_local",
+    "dnde_scan_values",
     "niter",
     "is_ul",
     "counts",
     "success",
     "n_dof",
-    "norm_scan_values",
-    "stat_scan_local",
-    "stat_scan",
 ]
 
 
@@ -96,14 +96,14 @@ OPTIONAL_QUANTITIES_COMMON = [
     "npred_excess",
     "stat",
     "stat_null",
+    "stat_scan",
+    "stat_scan_local",
+    "dnde_scan_values",
     "niter",
     "is_ul",
     "counts",
     "success",
     "n_dof",
-    "norm_scan_values",
-    "stat_scan_local",
-    "stat_scan",
 ]
 
 
@@ -495,10 +495,10 @@ class FluxMaps:
         return self._data["stat_scan"]
 
     @property
-    def norm_scan_values(self):
+    def dnde_scan_values(self):
         """Fit statistic norm scan values."""
-        self._check_quantity("norm_scan_values")
-        return self._data["norm_scan_values"]
+        self._check_quantity("dnde_scan_values")
+        return self._data["dnde_scan_values"]
 
     @property
     def stat_scan_local(self):
--- a/gammapy/estimators/map/tests/test_ts.py
+++ b/gammapy/estimators/map/tests/test_ts.py
@@ -311,15 +311,6 @@ def test_ts_map_stat_scan(fake_dataset):
     norm = maps.dnde_scan_values.data[ij, ind_best, ik, il] / dnde_ref.value
     assert_allclose(norm[success], maps_ref.norm.data[success], rtol=1e-5)
 
-    assert maps.stat_scan.geom.data_shape == (1, 109, 2, 2)
-    ts = np.abs(maps["stat_scan"].data.min(axis=1))
-    assert_allclose(ts[success], maps_ref.ts.data[success], rtol=1e-3)
-
-    ind_best = maps.stat_scan.data.argmin(axis=1)
-    ij, ik, il = np.indices(ind_best.shape)
-    dnde = maps.dnde_scan_values.quantity[ij, ind_best, ik, il]
-    assert_allclose(dnde[success], maps_ref.norm.data[success] * dnde_ref, rtol=1e-5)
-
 
 def test_ts_map_with_model(fake_dataset):
     model = fake_dataset.models["source"]
--- a/gammapy/estimators/map/tests/test_ts.py
+++ b/gammapy/estimators/map/tests/test_ts.py
@@ -297,7 +297,10 @@ def test_ts_map_stat_scan(fake_dataset):
 
     ind_best = maps.stat_scan_local.data.argmin(axis=1)
     ij, ik, il = np.indices(ind_best.shape)
-    norm = maps.dnde_scan_values.data[ij, ind_best, ik, il].value / maps.dnde_ref.value
+
+    dnde_ref = maps.dnde_ref.squeeze()
+    assert maps.dnde_scan_values.unit == dnde_ref.unit
+    norm = maps.dnde_scan_values.data[ij, ind_best, ik, il] / dnde_ref.value
     assert_allclose(norm[success], maps.norm.data[success], rtol=1e-5)
 
     maps.stat_scan.geom.data_shape == (1, 4001, 2, 2)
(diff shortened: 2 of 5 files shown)


gammapy/estimators/map/core.py
```diff
@@ -82,6 +82,9 @@
     "counts",
     "success",
     "n_dof",
+    "norm_scan_values",
+    "stat_scan_local",
+    "stat_scan",
```
AtreyeeS, line 87: `stat_scan` is already present (line 78)

AtreyeeS (commented):
Why do you want to provide two options `stat_scan_local` and `stat_scan`? Why not a similar API as for the `FluxPointsEstimator` where we have a configurable `norm` parameter which, by default corresponds to the `stat_scan_local` case ?

QRemy:
> Why do you want to provide two options `stat_scan_local` and `stat_scan`? Why not a similar API as for the `FluxPointsEstimator` where we have a configurable `norm` parameter which, by default corresponds to the `stat_scan_local` case ?

The API is the same, there is  a `norm_local` and `norm` the difference it how they are applied.
The problem is to define a norm range and sampling that is acceptable for all the bin of the map so I have two approach :

- The `stat_scan_local` uses the  `norm_local` parameter  that is multiplied with the best fit norm map and has a small range and only few points because we start close to the solution. In that case the outputs have  are "stat_scan_local" and "norm_scan_values" a map that gives the different norm values for each (E, l,b) pixels.

- The `stat_scan` uses the `norm` parameter that apply to the reference_model and has a wider range of values and finer sampling because a priori we don't know how far the reference_model is from the best model. In that case the output is just "stat_scan" a map where the norm values are given by the norm axis (same for all l,b pixels). This is quite convenient because if two flux maps were computed with the same reference_model and norm parameter their "stat_scan" maps can be summed to compute a joint flux maps as done in #5289.

bkhelifi (commented):
Thank @QRemy ... Nice features.
Beyond my small comments, you can add few simple test to improve the code coverage.

QRemy:
> Can you please describe again the motivation for the second "local" stat scan? I read your comment above, however I have still not understood why it is needed. Stat scans can have arbitrary sampling so if we need a finer sampling close to the minimum we could still do it with a single array...

The main motivations are better performances and easier configurability.

The best-fit norm can wary largely accros the map  so if we want to use a single norm values array with a finer binning and a wide range as implemented in "stat_scan" the computation time can be very high. Moreover we have to run a first time the estimator without the stat_scan to get the optimal norm range and then update estimator norm parameter and run the stat_scan. 
 
As "stat_scan_local" is defined with respect to the best-fit norm of each (l,b) pixel even if we scan a very small range of norms and it should be well defined for all pixels.

adonath:
Thanks @QRemy, that makes sense. I guess for flux points this was much less of a problem than for a flux map. However one main motivation for the TS profiles is the possibility to combine them afterwards, which requires them to be on the same grid (at least it is much simpler). Alternatively one could allow for a spatially varying reference model, but I guess this is currently not possible with `FluxMaps`.

Naively I would assume a log grid with like ~11 bins should be sufficient. Do you maybe have a distribution of norm values for a typical map? 

In any case it will be good to check, what "gttscube" does and who this problem is solved for the "TS Cube" format I linked above.

QRemy:
> Thanks a lot @QRemy, this is really nice! Please also take a look at https://gamma-astro-data-formats.readthedocs.io/en/v0.3/spectra/binned_likelihoods/index.html (Section "Likelihood SED Cube") to see whether the data could be serialized in the documented format.

the required columns for the scan are:
``
dloglike_scan – ndim: 3, Dimension: nrows x nebins x nnorms
norm_scan – ndim: 3, Dimension: nrows x nebins x nnorms
```
This is similar to the output of "stat_scan_local" except that the spatial dimensions are flatten.
For "stat_scan" I return only the stat map because the norm values are simply defined by the geom.axes["norm"].center

It should be easy to convert the gammapy output to this format if the reshaping from 2d to 1d is done in the same way.

File: gammapy/estimators/map/ts.py [hunk 2]
```diff
@@ -745,6 +831,61 @@ def estimate_errn_errp(self, dataset, result):
         )
         return {"norm_errn": flux_errn, "norm_errp": flux_errp}
 
+    def sample_norm(self, result):
+        """sample different scales to compute profile"""
```
bkhelifi, line 835:
```suggestion
        """Sample different scales to compute profile."""
```

gammapy/estimators/map/tests/test_ts.py
```diff
@@ -277,6 +277,50 @@ def test_compute_ts_map_downsampled(input_dataset):
...
+    assert maps.dnde_scan_values.unit == dnde_ref.unit
+    norm = maps.dnde_scan_values.data[ij, ind_best, ik, il] / dnde_ref.value
+    assert_allclose(norm[success], maps_ref.norm.data[success], rtol=1e-5)
+
+    assert maps.stat_scan.geom.data_shape == (1, 109, 2, 2)
+    ts = np.abs(maps["stat_scan"].data.min(axis=1))
+    assert_allclose(ts[success], maps_ref.ts.data[success], rtol=1e-3)
+
+    ind_best = maps.stat_scan.data.argmin(axis=1)
+    ij, ik, il = np.indices(ind_best.shape)
```
registerrier, line 319: This is duplicated from lines 306-307

gammapy/estimators/map/tests/test_ts.py
```diff
@@ -277,6 +277,50 @@ def test_compute_ts_map_downsampled(input_dataset):
...
+    dnde = maps.dnde_scan_values.quantity[ij, ind_best, ik, il]
+    assert_allclose(dnde[success], maps_ref.norm.data[success] * dnde_ref, rtol=1e-5)
```
registerrier, line 321: Technically this is the same test as l311-312, right?

gammapy/estimators/map/ts.py
```diff
@@ -443,10 +484,34 @@ def estimate_flux_map(self, dataset):
...
+
+                if name == "dnde_scan_values":
+                    unit = dnde_ref.unit
+                    factor = dnde_ref.value
+                else:
+                    unit = ""
+                    factor = 1
+
+                m = Map.from_geom(geom_scan, data=np.nan, unit=unit)
+                m.data[:, 0, j, i] = np.array([_[name] for _ in results]).T * factor
```
registerrier, line 510: Is the transposition enough here? have you tried with non-square maps?

gammapy/estimators/map/ts.py
```diff
@@ -745,6 +812,88 @@ def estimate_errn_errp(self, dataset, result):
...
+        stat_scan = spline(self.norm.scan_values)
+
+        ts = -stat_scan.min()
+        ind = stat_scan.argmin()
+        norm = self.norm.scan_values[ind]
+
+        maskp = self.norm.scan_values > norm
+        invalid_value = 999
+        stat_diff = stat_scan - stat_scan.min()
+        ind = np.abs(stat_diff + invalid_value * maskp - self.n_sigma**2).argmin()
```
registerrier, line 875: why not taking the argmin after applying the mask?

gammapy/estimators/map/ts.py
```diff
@@ -745,6 +812,88 @@ def estimate_errn_errp(self, dataset, result):
...
+        norm_errn = norm - self.norm.scan_values[ind]
+
+        ind = np.abs(stat_diff + invalid_value * (~maskp) - self.n_sigma**2).argmin()
```
registerrier, line 878: see above.

registerrier (commented):
Thanks @QRemy.  See inline comments.

I am a bit puzzled by the two different scans that are required. Is it an issue with our `FluxMaps` format ?

gammapy/estimators/map/tests/test_ts.py
QRemy: yes removed

gammapy/estimators/map/tests/test_ts.py
QRemy: yes removed

gammapy/estimators/map/ts.py
```diff
@@ -166,12 +177,32 @@ def __init__(
...
         n_jobs=None,
         parallel_backend=None,
+        norm=None,
     ):
         if kernel_width is not None:
             kernel_width = Angle(kernel_width)
 
         self.kernel_width = kernel_width
 
+        if norm is None or isinstance(norm, dict):
```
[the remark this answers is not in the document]
QRemy: added `_get_default_norm` un utils

gammapy/estimators/map/ts.py
QRemy: the map from the results object here is always 2d here and it has the dimension (n_valid_pixels, n_scan_values=109)

gammapy/estimators/map/ts.py
```diff
@@ -745,6 +812,88 @@ def estimate_errn_errp(self, dataset, result):
...
+                norm + np.linspace(-10, 10, 21) * norm_err,
+                np.abs(norm) * np.linspace(-10, 10, 21),
+                np.linspace(-10, 10, 21),
+                np.linspace(self.norm.scan_values[0], self.norm.scan_values[-1], 2),
+            )
+        )
+        sparse_norms = np.unique(sparse_norms)
+        if len(sparse_norms) != 109:
+            rand_norms = 20 * np.random.rand(109 - len(sparse_norms)) - 10
+            sparse_norms = np.concatenate((sparse_norms, rand_norms))
```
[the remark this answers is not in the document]
QRemy: The len of stat_scan should be the shape so it can be stored into a map. But is it not always the case, if norm is divisible by norm_err line 826 and 826 can give the same value.

gammapy/estimators/map/ts.py
```diff
@@ -745,6 +812,88 @@ def estimate_errn_errp(self, dataset, result):
...
+
+        ts = -stat_scan.min()
+        ind = stat_scan.argmin()
+        norm = self.norm.scan_values[ind]
+
+        maskp = self.norm.scan_values > norm
+        invalid_value = 999
+        stat_diff = stat_scan - stat_scan.min()
+        ind = np.abs(stat_diff + invalid_value * maskp - self.n_sigma**2).argmin()
+        norm_errn = norm - self.norm.scan_values[ind]
```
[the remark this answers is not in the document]
QRemy: it is reused after
`norm_err = (norm_errn + norm_errp) / 2`

gammapy/estimators/map/ts.py
QRemy: done

gammapy/estimators/map/ts.py
```diff
@@ -745,6 +812,88 @@ def estimate_errn_errp(self, dataset, result):
...
+            sparse_norms, stat_scan_local, k=1, ext="raise", check_finite=True
+        )
+        stat_scan = spline(self.norm.scan_values)
+
+        ts = -stat_scan.min()
+        ind = stat_scan.argmin()
+        norm = self.norm.scan_values[ind]
+
+        maskp = self.norm.scan_values > norm
+        invalid_value = 999
```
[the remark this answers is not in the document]
QRemy: removed

gammapy/estimators/map/ts.py
```diff
@@ -745,6 +812,88 @@ def estimate_errn_errp(self, dataset, result):
...
+        sparse_norms = np.unique(sparse_norms)
+        if len(sparse_norms) != 109:
+            rand_norms = 20 * np.random.rand(109 - len(sparse_norms)) - 10
+            sparse_norms = np.concatenate((sparse_norms, rand_norms))
+        return np.sort(sparse_norms)
+
+    def estimate_scan(self, dataset, result):
+        """Compute likelihood profile on a fixed norm range.
+
+        Parameters
```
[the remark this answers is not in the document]
QRemy: removed ` on a fixed norm range`

QRemy:
> I am a bit puzzled by the two different scans that are required. Is it an issue with our `FluxMaps` format ?

The large grid from `_generate_scan_values` is used to find the norm, norm_err,.. at the require precision, while small grid from `_get_norm_scan_values` allows to sample the profile with much fewer values so the computation is faster.

registerrier (approved):
Thanks @QRemy . This looks good to me!

PR #5297, opened 2024-06-11
Precompute the PSF Kernel if the PSFMap has only one bin

QRemy:
Improve PSF  kernel computation performance
- precompute psf kernel on the map dataset if the psf map has a region geom
- add a multiscale option on get_psf_kernel, if true it returns a list of kernels with energy dependent max_radius (having instead a single irregular map with different spatial shape for each energy bins is broken and it would take a lot more work to fix it than this patch) 

This should drastically speed-up Fermi-LAT analyses.
For now I only tested it on the dataset that David Green sent to @AtreyeeS
Could you also test it for your analysis @facero ?

### DIFF
--- a/gammapy/datasets/evaluator.py
+++ b/gammapy/datasets/evaluator.py
@@ -210,10 +210,7 @@ class MapEvaluator:
         if (
             psf
             and self.model.spatial_model
-            and not (
-                isinstance(self.psf, PSFKernel)
-                and psf.psf_map.geom.to_image().data_shape == (1, 1)
-            )
+            and not (isinstance(self.psf, PSFKernel) and psf.psf_map.single_spatial_bin)
         ):
             energy_name = psf.energy_name
             geom_psf = geom if energy_name == "energy" else exposure.geom
--- a/gammapy/datasets/evaluator.py
+++ b/gammapy/datasets/evaluator.py
@@ -210,7 +210,7 @@ class MapEvaluator:
         if (
             psf
             and self.model.spatial_model
-            and not (isinstance(self.psf, PSFKernel) and psf.psf_map.single_spatial_bin)
+            and not (isinstance(self.psf, PSFKernel) and psf.has_single_spatial_bin)
         ):
             energy_name = psf.energy_name
             geom_psf = geom if energy_name == "energy" else exposure.geom
--- a/gammapy/datasets/evaluator.py
+++ b/gammapy/datasets/evaluator.py
@@ -7,7 +7,7 @@ from astropy.coordinates import angular_separation
 from astropy.utils import lazyproperty
 from regions import CircleSkyRegion
 import matplotlib.pyplot as plt
-from gammapy.irf import EDispKernel
+from gammapy.irf import EDispKernel, PSFKernel
 from gammapy.maps import HpxNDMap, Map, RegionNDMap, WcsNDMap
 from gammapy.modeling.models import PointSpatialModel, TemplateNPredModel
 from .utils import apply_edisp
@@ -207,23 +207,28 @@ class MapEvaluator:
 
         # lookup psf
         if psf and self.model.spatial_model:
-            energy_name = psf.energy_name
-            geom_psf = geom if energy_name == "energy" else exposure.geom
-
-            if self.use_psf_containment(geom=geom_psf):
-                energy_values = geom_psf.axes[energy_name].center.reshape((-1, 1, 1))
-                kwargs = {energy_name: energy_values, "rad": geom.region.radius}
-                self.psf_containment = psf.containment(**kwargs)
+            if isinstance(psf, PSFKernel):
+                self.psf = psf
             else:
-                if geom_psf.is_region or geom_psf.is_hpx:
-                    geom_psf = geom_psf.to_wcs_geom()
+                energy_name = psf.energy_name
+                geom_psf = geom if energy_name == "energy" else exposure.geom
 
-                self.psf = psf.get_psf_kernel(
-                    position=self.position,
-                    geom=geom_psf,
-                    containment=PSF_CONTAINMENT,
-                    max_radius=PSF_MAX_RADIUS,
-                )
+                if self.use_psf_containment(geom=geom_psf):
+                    energy_values = geom_psf.axes[energy_name].center.reshape(
+                        (-1, 1, 1)
+                    )
+                    kwargs = {energy_name: energy_values, "rad": geom.region.radius}
+                    self.psf_containment = psf.containment(**kwargs)
+                else:
+                    if geom_psf.is_region or geom_psf.is_hpx:
+                        geom_psf = geom_psf.to_wcs_geom()
+
+                    self.psf = psf.get_psf_kernel(
+                        position=self.position,
+                        geom=geom_psf,
+                        containment=PSF_CONTAINMENT,
+                        max_radius=PSF_MAX_RADIUS,
+                    )
 
         self.exposure = exposure
         if self.evaluation_mode == "local":
--- a/gammapy/datasets/evaluator.py
+++ b/gammapy/datasets/evaluator.py
@@ -223,9 +223,6 @@ class MapEvaluator:
                 kwargs = {energy_name: energy_values, "rad": geom.region.radius}
                 self.psf_containment = psf.containment(**kwargs)
             else:
-                if geom.is_region or geom.is_hpx:
-                    geom = geom.to_wcs_geom()
-
                 self.psf = psf.get_psf_kernel(
                     position=self.position,
                     geom=geom_psf,
--- a/gammapy/datasets/evaluator.py
+++ b/gammapy/datasets/evaluator.py
@@ -231,9 +231,6 @@ class MapEvaluator:
                     kwargs = {energy_name: energy_values, "rad": geom.region.radius}
                     self.psf_containment = psf.containment(**kwargs)
                 else:
-                    if geom_psf.is_region or geom_psf.is_hpx:
-                        geom_psf = geom_psf.to_wcs_geom()
-
                     self.psf = psf.get_psf_kernel(
                         position=self.position,
                         geom=geom_psf,
--- a/gammapy/datasets/evaluator.py
+++ b/gammapy/datasets/evaluator.py
@@ -200,36 +200,35 @@ class MapEvaluator:
         # lookup edisp
         del self._edisp_diagonal
         if edisp:
-            if isinstance(self.edisp, EDispKernel):
-                pass
-            else:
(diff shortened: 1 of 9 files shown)


adonath:
Thanks @QRemy for PR! A couple of general comments first:

- In general I am skeptical about these very specialized performance improvements. They add a lot of code, with high complexity that sooner or later will break and needs to be refactored or removed again for maintenance.
- The list of PSF kernels does not really offer a performance improvement for the actual convolution. The performance of the fft based convolution barely depends on the kernel size. However I would indeed expect that the computation of the kernel is faster. 
- It is almost always better to solve an issue with an abstraction...

So my concrete proposal here would be to work on a minimal version of the `PSFKernelMap` instead (see also #3689). It can remain very simple, so just support a region geom and a single PSF kernel. This would move the computation of the kernel to the data reduction. For the fitting step the kernel is just extracted using indexing. This avoids the re-computation for different model components in the same dataset. To speed up the data reduction you can still use the "multi scale" approach, but instead of returning a list I would just insert into a kernel map and zero pad for the small kernels. For Fermi analysis there can also be a simple way to just create a kernel map from a normal PSF map in the gtpsf format.

QRemy:
making the upsampling factor energy dependent helps a lot, the kernel computation is no longer the bottleneck
Screenshot 2024-06-12 at 23 28 03

facero:
Hey,
I tested the computation time with your latest PR of this morning and it's an amazing speed improvment !
I have more than 30 sources in my  Fermi ROI and previously the first Npred call was taking several minutes.
Now it's about a few sec !
Then the fit take a similar time which is expected as I fit only spectral parameters.
But this would have a huge impact if I were to fit the spatial parameters as well.

QRemy:
Thanks @facero for the feedback.
The computation time for the flux points could be further improved. 
For now the slicing of the dataset in energy range  of each flux points impose to reset the models and clear the psf convolution cache. Using making instead of slicing to select the energy range would prevent that
https://github.com/gammapy/gammapy/blob/ddf72e06a15cf995363c86d1acb9dce4eabdbb91/gammapy/estimators/points/sed.py#L201-L206
This will be a separate PR.

gammapy/datasets/map.py
```diff
@@ -439,6 +438,17 @@ def __init__(
         else:
             self._meta = meta
 
+    @property
+    def _psf_kernel(self):
+        """Precompute PSFkernel if possible"""
+        if self.psf and self.psf.psf_map.geom.to_image().data_shape == (1, 1):
```
adonath, line 444: Maybe we can introduce a convenience property on the `Map` object here. Such as `Map.has_only_one_spatial_bin` (I'm open to better suggestion for names...:-)

gammapy/datasets/tests/test_map.py
```diff
@@ -276,7 +277,7 @@ def test_fake(sky_model, geom, geom_etrue):
 
     assert real_dataset.counts.data.shape == dataset.counts.data.shape
     assert_allclose(real_dataset.counts.data.sum(), 9525.299054, rtol=1e-5)
-    assert_allclose(dataset.counts.data.sum(), 9711)
+    assert_allclose(dataset.counts.data.sum(), 9709)
```
adonath, line 280: This PR should not change any test values. I think it is related to my comment above, that you start out with a single  PSF for all model components, instead of extracting the one at the components position.

adonath (commented):
Thanks @QRemy! Can you maybe split the PR to implement the change to the `MapEvaluator` and to `PSFMap.get_psf_kernel()` independently. Right now we cannot track which change introduces the changes in test values.

gammapy/datasets/map.py
QRemy: I called it `single_spatial_bin` so it's shorter

gammapy/datasets/tests/test_map.py
QRemy: The tests changes are not related to the precomputation of the PSFKernel.
After I moved the modifications of the get_psf_kernel to a separate PR (https://github.com/gammapy/gammapy/pull/5314) the tests failed there and are all good in this one.
Actually they come from the changes in upsampling factor, and in the previous PR that modified this I also had to change few test values ([#4975](https://github.com/gammapy/gammapy/pull/4975/files)).

gammapy/datasets/map.py
```diff
@@ -539,15 +549,16 @@ def excess(self):
     def models(self, models):
         """Models setter."""
         self._evaluators = {}
-
         if models is not None:
             models = DatasetModels(models)
             models = models.select(datasets_names=self.name)
-
+            if models:
+                psf = self._psf_kernel
```
registerrier, line 556: what happens if the psf map does not contain a one bin PSFMap?

gammapy/datasets/map.py
```diff
@@ -439,6 +438,17 @@ def __init__(
         else:
             self._meta = meta
 
+    @property
```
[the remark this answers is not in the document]
registerrier: Maybe this could be made a standalone helper function as this depends only on self.psf?

gammapy/datasets/map.py
```diff
@@ -439,6 +438,17 @@ def __init__(
...
+    def _psf_kernel(self):
+        """Precompute PSFkernel if possible"""
```
registerrier, line 443: Maybe specify what possible means here

gammapy/datasets/tests/test_map.py
```diff
@@ -1946,6 +1949,17 @@ def test_dataset_mixed_geom(tmpdir):
...
+    assert isinstance(dataset.psf, PSFMap)
+    assert isinstance(dataset._psf_kernel, PSFKernel)
+
+    geom_psf = WcsGeom.create(npix=1, axes=[rad_axis, energy_axis_true])
+    dataset = MapDataset.from_geoms(
+        geom=geom, geom_exposure=geom_exposure, geom_psf=geom_psf, geom_edisp=geom_edisp
+    )
+
+    assert isinstance(dataset.psf, PSFMap)
+    assert isinstance(dataset._psf_kernel, PSFKernel)
```
registerrier, line 1961: Maybe add a test for the case where no kernel should be returned.

gammapy/maps/wcs/tests/test_ndmap.py
```diff
@@ -639,6 +639,30 @@ def test_convolve_nd():
     assert_allclose(values_full, values_same, rtol=1e-5)
 
 
+def test_get_psf_kernel_multiscale():
```
registerrier, line 642: I find strange that this test in within gammapy.maps rather than gammpay.irf. 
It tests `PSFMap` and `PSFKernelMap`

registerrier: Is it really related to this PR?

registerrier (commented):
Thanks @QRemy. This is nice. Maybe the title of the PR should be `Precompute the PSF Kernel if the PSFMap has only one bin`.

registerrier:
How much backporting is needed here? Performance improvement wouldn't require a backport a priori.

gammapy/datasets/evaluator.py
```diff
@@ -209,9 +213,6 @@ def update(self, exposure, psf, edisp, geom, mask):
                 kwargs = {energy_name: energy_values, "rad": geom.region.radius}
                 self.psf_containment = psf.containment(**kwargs)
             else:
-                if geom_psf.is_region or geom_psf.is_hpx:
```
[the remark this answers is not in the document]
QRemy: I moved this into `get_psf_kernel`, see below

gammapy/maps/wcs/tests/test_ndmap.py
QRemy: forgot to move the new tests into #5314

gammapy/datasets/map.py
QRemy: `self._psf_kernel ` return `None` which is the default of the `MapEvaluator`

gammapy/datasets/map.py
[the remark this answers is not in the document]
QRemy: it uses exposure or counts to get the geom

gammapy/datasets/map.py
QRemy: added

gammapy/maps/wcs/tests/test_ndmap.py
QRemy: moved

gammapy/datasets/tests/test_map.py
QRemy: added

QRemy:
> How much backporting is needed here? Performance improvement wouldn't require a backport a priori.

At least for 1.2 it should go in without much conflicts.

gammapy/maps/core.py
```diff
@@ -87,6 +87,10 @@ def geom(self):
         """Map geometry as a `~gammapy.maps.Geom` object."""
         return self._geom
 
+    @property
+    def single_spatial_bin(self):
```
registerrier, line 91:
```suggestion
    def has_single_spatial_bin(self):
```
But shouldn't this live on `Geom` instead?

gammapy/datasets/evaluator.py
```diff
@@ -200,7 +200,11 @@ def update(self, exposure, psf, edisp, geom, mask):
             del self._edisp_diagonal
 
         # lookup psf
-        if psf and self.model.spatial_model:
+        if (
+            psf
+            and self.model.spatial_model
+            and not (isinstance(self.psf, PSFKernel) and psf.psf_map.single_spatial_bin)
```
registerrier, line 206: Maybe one does not need the `is_single_spatial_bin` property on `Map` but directly on `IRFMap`?

gammapy/maps/core.py
```diff
@@ -87,6 +87,10 @@ def geom(self):
...
+        return self.geom.to_image().data_shape == (1, 1)
```
registerrier, line 92: I would avoid calling `to_image` in the `MapEvaluator.update` function because it can take time for no reason if e.g. you have a `RegionGeom`. If `is_single_spatial_bin` is a property on `Geom` then it can be always `True`.

registerrier (commented):
The final comments are connected to the `single_spatial_bin` property:
- use a more explicit name such `is_single_spatial_bin` or `has_single_spatial_bin`
- Maybe place it on `Geom` instead of `Map`
- Maybe you only need it on `IRFMap`

gammapy/maps/core.py
QRemy: moved to IRFmap as a lazy property and renamed `has_single_spatial_bin `

PR #5298, opened 2024-06-11
Coherent units in `TemporalModel.sample_time`

fabiopintore:
This is a very minimal fix to preserve units coherence in the `TemporalModel.sample_time`. 

The time sampling is indeed performed over a number of time bins which is calculated internally on the basis of the `ontime` and the `t_delta`. 
In the current version, the `ontime` is calculated in seconds but it takes the units from `t_delta`. The latter is by default in seconds, but it can be passed by the user. 
If the units of `ontime` and `t_delta` are not consistent, this introduces an issue.

### DIFF
--- a/gammapy/modeling/models/temporal.py
+++ b/gammapy/modeling/models/temporal.py
@@ -198,7 +198,7 @@ class TemporalModel(ModelBase):
         t_delta = u.Quantity(t_delta)
         random_state = get_random_state(random_state)
 
-        ontime = u.Quantity((t_max - t_min).sec, t_delta.to("s").unit)
+        ontime = (t_max - t_min).to("s")
         n_step = (ontime / t_delta).to_value("").item()
         t_step = ontime / n_step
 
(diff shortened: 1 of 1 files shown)


bkhelifi (approved):

gammapy/modeling/models/temporal.py
```diff
@@ -198,7 +198,7 @@ def sample_time(self, n_events, t_min, t_max, t_delta="1 s", random_state=0):
         t_delta = u.Quantity(t_delta)
         random_state = get_random_state(random_state)
 
-        ontime = u.Quantity((t_max - t_min).sec, t_delta.unit)
+        ontime = u.Quantity((t_max - t_min).sec, t_delta.to("s").unit)
```
registerrier, line 201:
```suggestion
        ontime = (t_max - t_min).to("s")
```

registerrier (changes requested):
Thanks @fabiopintore . I think it would be simpler to convert the `TimeDelta` to a Quantity in seconds.

cgalelli (approved):
Thanks @fabiopintore . An alternative solution could be to use .decompose() but this is completely fine (and probably more efficient)

QRemy (approved):
Thanks @fabiopintore, no comment from my side.

QRemy:
Backport to 1.0.x is not needed the code has correct a unit `ontime = u.Quantity((t_max - t_min).sec, "s")`

Issue #4748, opened 2023-09-14
`DarkMatterAnnihilationSpectralModel` not readable from `Models.read` 


StFroese:
**Gammapy version**
v1.1

**Bug description**
`DarkMatterAnnihilationSpectralModel` can't be instantiated when read from .yaml during execution of `Models.read`

If you combine a `DarkMatterAnnihilationSpectralModel` and a `SpatialModel` (i.e. `TemplateSpatialModel`for j-factor) to `SkyModel` and save this as `Models([sky_model]).write('some_file.yaml')`, you can not read in the yaml again with `Models.read('some_file.yaml').`
This is due to the problem that the `DarkMatterAnnihilationSpectralModel` is not found in the `SPECTRAL_MODEL_REGISTRY` since it is only appended inside the gammapy/astro/darkmatter/spectra.py.
The only solution right now is to import the file before de-serialization
```python
import gammapy.astro.darkmatter.spectra

...

Models.read('some_file.yaml')
```

**Solution**
Move the append of the `DarkMatterAnnihilationSpectralModel ` directly to the `SPECTRAL_MODEL_REGISTRY` at definition

Astro-Kirsty:
This is not possible as the `DarkMatterAnnihilationSpectralModel` relies on `modeling` directly, which would lead to a circular import. The only way to get around this would be to adapt the location of the DM spectral models including API change which is not something we want to do. 

Instead I moved the appending to `gammapy.astro.darkmatter` but it means you still need to import it. 
Possibly we need further documentation?

PR #5299, opened 2024-06-11
Small adaption to appending DM spectra models

Astro-Kirsty:
Related to #4748 

See comment there


registerrier (commented):
I don't know whether a change in the import is an API change or not.

adonath:
I think in principle this is a backwards incompatible change. However the import should be so rarely used, so it should be fine...

bkhelifi (approved):

QRemy (approved):
Thanks @Astro-Kirsty, I agree we can merge it like that even if no fully backward compatible.

PR #5303, opened 2024-06-12
QOL changes in Timmer&Konig

cgalelli:
This PR adds 3 quality of life changes for the timer and konig algorithm:

- A better handling of the red noise leakage through chunks
- Possibility of setting a custom mean and standard deviation
- Possibility of adding a poissonian noise (off by default)

These changes are very beneficial for the recipe presented in the Heidelberg coding sprint

### DIFF
--- a/gammapy/stats/tests/test_variability.py
+++ b/gammapy/stats/tests/test_variability.py
@@ -152,7 +152,7 @@ def test_tk():
         lambda x: x ** (-3), 20, 1 * u.s
     )
     time_series2, time_axis2 = TimmerKonig_lightcurve_simulator(
-        lambda x: x**0.5, 21, 2 * u.h
+        lambda x: x**0.5, 21, 2 * u.h, nchunks=1, poisson=True
     )
 
     def temp(x, norm, index):
@@ -161,11 +161,11 @@ def test_tk():
     params = {"norm": 1.5, "index": 3}
 
     time_series3, time_axis3 = TimmerKonig_lightcurve_simulator(
-        temp, 15, 1 * u.h, power_spectrum_params=params
+        temp, 15, 1 * u.h, nchunks=100, power_spectrum_params=params
     )
 
     time_series4, time_axis4 = TimmerKonig_lightcurve_simulator(
-        lambda x: x ** (-3), 20, 1 * u.s, leakage_protection=100.0
+        lambda x: x ** (-3), 200, 1 * u.s, mean=2.0, std=0.1
     )
 
     assert len(time_series) == 20
@@ -173,19 +173,9 @@ def test_tk():
     assert time_axis.unit == u.s
     assert len(time_series2) == 21
     assert len(time_series3) == 15
-    assert len(time_series4) == 20
-
-
-def test_tk_badleakage():
-    with pytest.raises(ValueError):
-        TimmerKonig_lightcurve_simulator(
-            lambda x: x ** (-3), 20, 1 * u.s, leakage_protection=0.1
-        )
-
-    with pytest.raises(Warning):
-        TimmerKonig_lightcurve_simulator(
-            lambda x: x ** (-3), 20, 1 * u.s, leakage_protection=1.001
-        )
+    assert len(time_series4) == 200
+    assert_allclose(time_series4.mean(), 2.0)
+    assert_allclose(time_series4.std(), 0.1)
 
 
 def test_structure_function():
--- a/gammapy/stats/variability.py
+++ b/gammapy/stats/variability.py
@@ -292,8 +292,8 @@ def TimmerKonig_lightcurve_simulator(
     nchunks=10,
     random_state="random-seed",
     power_spectrum_params=None,
-    mean=0,
-    std=1,
+    mean=0.0,
+    std=1.0,
     poisson=False,
 ):
     """Implementation of the Timmer-Koenig algorithm to simulate a time series from a power spectrum.
--- a/gammapy/stats/variability.py
+++ b/gammapy/stats/variability.py
@@ -292,8 +292,8 @@ def TimmerKonig_lightcurve_simulator(
     nchunks=10,
     random_state="random-seed",
     power_spectrum_params=None,
-    mean=None,
-    std=None,
+    mean=0,
+    std=1,
     poisson=False,
 ):
     """Implementation of the Timmer-Koenig algorithm to simulate a time series from a power spectrum.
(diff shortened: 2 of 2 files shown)


AtreyeeS (approved):
Thank you @cgalelli ! I have no specific comments on the implementation here, but it would be very helpful to have some detailed documentation on the implementation. Maybe in the user guide for gammapy.stats

gammapy/stats/variability.py
```diff
@@ -304,13 +307,20 @@ def TimmerKonig_lightcurve_simulator(
...
+    random_state : {int, 'random-seed', 'global-rng', `~numpy.random.RandomState`}, optional
         Defines random number generator initialisation.
         Passed to `~gammapy.utils.random.get_random_state`. Default is "random-seed".
     power_spectrum_params : dict, optional
         Dictionary of parameters to be provided to the power spectrum function.
+    mean : float, `~astropy.units.Quantity`
+        Desired mean of the final series. Default is 0.
+    std : float, `~astropy.units.Quantity`
+        Desired standard deviation of the final series. Default is 1.
+    poisson : bool
```
registerrier, line 321:
```suggestion
    poisson : bool, optional
```

gammapy/stats/variability.py
```diff
@@ -304,13 +307,20 @@ def TimmerKonig_lightcurve_simulator(
...
-        Factor by which to multiply the length of the time series to avoid red noise leakage. Default is 1.0.
-    random_state : {int, 'random-seed', 'global-rng', `~numpy.random.RandomState`}
+    nchunks : int, optional
+        Factor by which to multiply the length of the time series to avoid red noise leakage. Default is 10.
+    random_state : {int, 'random-seed', 'global-rng', `~numpy.random.RandomState`}, optional
         Defines random number generator initialisation.
         Passed to `~gammapy.utils.random.get_random_state`. Default is "random-seed".
     power_spectrum_params : dict, optional
         Dictionary of parameters to be provided to the power spectrum function.
+    mean : float, `~astropy.units.Quantity`
```
registerrier, line 317:
```suggestion
    mean : float, `~astropy.units.Quantity`, optional
```

gammapy/stats/variability.py
```diff
@@ -304,13 +307,20 @@ def TimmerKonig_lightcurve_simulator(
...
+        Desired mean of the final series. Default is 0.
+    std : float, `~astropy.units.Quantity`
```
registerrier, line 319:
```suggestion
    std : float, `~astropy.units.Quantity`, optional
```

gammapy/stats/variability.py
```diff
@@ -349,16 +359,9 @@ def TimmerKonig_lightcurve_simulator(
 
     random_state = get_random_state(random_state)
 
-    if leakage_protection < 1.0:
-        raise ValueError("The leakage protection factor must be at least 1.0.")
+    npoints_ext = npoints * nchunks
```
registerrier, line 362: shouldn't there be a smiliar check for `nchunks`. I suppose 0.5 is a bad value, right?

registerrier: You should convert to int also.

gammapy/stats/variability.py
```diff
@@ -289,9 +289,12 @@ def TimmerKonig_lightcurve_simulator(
     power_spectrum,
     npoints,
     spacing,
-    leakage_protection=1.0,
+    nchunks=10,
     random_state="random-seed",
     power_spectrum_params=None,
+    mean=0.0,
```
registerrier, line 295: Does this follow the initial Timmer and König paper? 
Are there references to support the schemes introduced?

gammapy/stats/variability.py
```diff
@@ -289,9 +289,12 @@ def TimmerKonig_lightcurve_simulator(
...
+    std=1.0,
+    poisson=False,
```
registerrier, line 297: Does it always apply? What is it supposed to do for flux like quantities?

gammapy/stats/tests/test_variability.py
```diff
@@ -152,7 +152,7 @@ def test_tk():
         lambda x: x ** (-3), 20, 1 * u.s
     )
     time_series2, time_axis2 = TimmerKonig_lightcurve_simulator(
-        lambda x: x**0.5, 21, 2 * u.h
+        lambda x: x**0.5, 21, 2 * u.h, nchunks=1, poisson=True
```
registerrier, line 155: if not additional assert is present `poisson=True` is not really tested right?

gammapy/stats/tests/test_variability.py
```diff
@@ -161,31 +161,21 @@ def temp(x, norm, index):
...
-    with pytest.raises(ValueError):
-        TimmerKonig_lightcurve_simulator(
-            lambda x: x ** (-3), 20, 1 * u.s, leakage_protection=0.1
-        )
-
-    with pytest.raises(Warning):
-        TimmerKonig_lightcurve_simulator(
-            lambda x: x ** (-3), 20, 1 * u.s, leakage_protection=1.001
-        )
+    assert len(time_series4) == 200
```
registerrier, line 176: Shouldn't the LC be shrunked to the input size?

registerrier (commented):
Thanks @cgalelli . 
I have left inline comments. 
I think this needs a bit more documentation and testing.

gammapy/stats/tests/test_variability.py
cgalelli: In this case input size is 200 so it is the correct length

registerrier: My bad :)
Then it would be worth testing that changing nchunks always return the correct size.

gammapy/stats/variability.py
cgalelli: You are right, in the case of flux quantities the result of the poisson noise will probably be 0: the option is expected to be used with "norm"-like arrays. Added a warning for the case

gammapy/stats/variability.py
```diff
@@ -347,18 +358,24 @@ def TimmerKonig_lightcurve_simulator(
             "The power spectrum has to be provided as a callable function."
         )
 
+    if not isinstance(npoints * nchunks, int):
+        raise TypeError("npoints and nchunks must be integers")
```
VascoSch92, line 362:
```suggestion
        raise TypeError("The value ``npoints * nchunks`` must be an integer.")
```
you are checking that the multiplication is an integer not that the single variables are

cgalelli: The only way that the multiplication is an integer is if both factors are integers, it's just a more compact way of doing the check.

QRemy (approved):
Thanks @cgalelli no comment from my side.

PR #5304, opened 2024-06-12
introducing a `filename` argument to TemplateSpatialModel.write

### LINKED ISSUE #4747: Change `write` API for `TemplateSpatialModel`
I tried to use the `write` function of the `TemplateSpatialModel` and noticed that it has a different API than other `SpatialModel`s. 
Instead of having a function 
```python
def write(self, filename, overwrite=False)
```
the function looks like this

https://github.com/gammapy/gammapy/blob/da122ea83dc441ffc327b767329b36146923b7f3/gammapy/modeling/models/spatial.py#L1409-L1415

So it's using `self.filename` from the instantiation of the class. 
Can we change that?

fabiopintore:
This PR fixes #4747.

### DIFF
--- a/gammapy/modeling/models/spatial.py
+++ b/gammapy/modeling/models/spatial.py
@@ -1507,8 +1507,9 @@ class TemplateSpatialModel(SpatialModel):
             Overwrite existing file.
             Default is False, which will raise a warning if the template file exists already.
         filename: str, optional
-            Set the filename of the template  model, if needed. By default, the template model
-            will be saved with the TemplateSpatialModel.filename.
+            Filename of the template model. By default, the template model
+            will be saved with the `TemplateSpatialModel.filename` attribute,
+            if `filename` is provided this attribute will be updated.
         """
         if filename is not None:
             self.filename = filename
--- a/gammapy/modeling/models/spatial.py
+++ b/gammapy/modeling/models/spatial.py
@@ -1497,7 +1497,7 @@ class TemplateSpatialModel(SpatialModel):
         data["spatial"]["unit"] = str(self.map.unit)
         return data
 
-    def write(self, overwrite=False):
+    def write(self, overwrite=False, filename=None):
         """
         Write the map.
 
@@ -1507,8 +1507,13 @@ class TemplateSpatialModel(SpatialModel):
             Overwrite existing file.
             Default is False, which will raise a warning if the template file exists already.
         """
-        if self.filename is None:
+        if self.filename is None and filename is None:
             raise IOError("Missing filename")
+        elif filename:
+            if os.path.isfile(make_path(filename)) and not overwrite:
+                log.warning("Template file already exits, and overwrite is False")
+            else:
+                self.map.write(filename, overwrite=overwrite)
         elif os.path.isfile(make_path(self.filename)) and not overwrite:
             log.warning("Template file already exits, and overwrite is False")
         else:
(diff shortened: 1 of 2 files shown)


gammapy/modeling/models/spatial.py
```diff
@@ -1507,8 +1507,13 @@ def write(self, overwrite=False):
...
             Default is False, which will raise a warning if the template file exists already.
         """
-        if self.filename is None:
+        if self.filename is None and filename is None:
             raise IOError("Missing filename")
+        elif filename:
+            if os.path.isfile(make_path(filename)) and not overwrite:
+                log.warning("Template file already exits, and overwrite is False")
+            else:
+                self.map.write(filename, overwrite=overwrite)
```
QRemy, line 1516:
```suggestion
        if filename is not None:
            self.filename = filename
        if self.filename is None:
            raise IOError("Missing filename")
```

gammapy/modeling/models/spatial.py
```diff
@@ -1497,7 +1497,7 @@ def to_dict(self, full_output=False):
         data["spatial"]["unit"] = str(self.map.unit)
         return data
 
-    def write(self, overwrite=False):
+    def write(self, overwrite=False, filename=None):
```
QRemy, line 1500: the docstring is missing for filename

fabiopintore:
One test is failing (`test_spatial::test_sky_diffuse_map`) although my changes shouldn't affect it. I note that it fails also in the main...

gammapy/modeling/models/spatial.py
```diff
@@ -1506,10 +1506,14 @@ def write(self, overwrite=False):
         overwrite: bool, optional
             Overwrite existing file.
             Default is False, which will raise a warning if the template file exists already.
+        filename: str, optional
+            Set the filename of the template  model.
```
AtreyeeS, line 1510: Maybe specify that this is an added convenience, and by default the template will be written to `self.filename`

gammapy/modeling/models/spatial.py
```diff
@@ -1506,10 +1506,15 @@ def write(self, overwrite=False):
         overwrite: bool, optional
             Overwrite existing file.
             Default is False, which will raise a warning if the template file exists already.
+        filename: str, optional
+            Set the filename of the template  model, if needed. By default, the template model
+            will be saved with the TemplateSpatialModel.filename.
         """
+        if filename is not None:
```
registerrier, line 1513: In general, what is the expected behaviour. Should the `write` function really modify the attribute? I would think not, but there might be a good reason.
A minima this should be documented.

registerrier (commented):
Thanks @fabiopintore . This looks good.
My main question is whether the write function should change the filename attribute.

gammapy/modeling/models/spatial.py
QRemy: `Models.read` use the filename from the yaml file so write has to change it.

gammapy/modeling/models/spatial.py
```diff
@@ -1506,10 +1506,15 @@ def write(self, overwrite=False):
         overwrite: bool, optional
             Overwrite existing file.
             Default is False, which will raise a warning if the template file exists already.
+        filename: str, optional
+            Set the filename of the template  model, if needed. By default, the template model
+            will be saved with the TemplateSpatialModel.filename.
```
QRemy, line 1511:
```suggestion
            Filename of the template model. By default, the template model
            will be saved with the `TemplateSpatialModel.filename` attribute,
            if `filename` is provided this attribute will be updated.
```

QRemy (approved):
Thanks @fabiopintore, no further comment.

Issue #4780, opened 2023-10-03
Support multiple time intervals in `gammapy.data.ObservationFilter`


jlenain:
**Is your feature request related to a problem? Please describe.**
Suppose a cloud is passing by during a given IACT observation. The user would like to filter out such a bad time interval, defining two, or multiple good time interval(s) for this observation. Currently, `gammapy.data.ObservationFilter` only supports a single time interval.

**Describe the solution you'd like**
A possibility would be to accept a list of of couple of time intervals in the `obs_filter` attribute of the `gammapy.data.Observation` class.

**Describe alternatives you've considered**
Another, manual way out, suggested by @registerrier , consists in created a `gammapy.data.GTI` table and apply it to the observation with:
```
from gammapy.data import GTI
new_gti = GTI.create(...)
obs._gti = new_gti
```
but this is less elegant, as we access an attribute which is not supposed to be exposed.

PR #5305, opened 2024-06-12
Support multiple time intervals during observation filtering

bkhelifi:
This PR is associate with the feature request #4780. It permits to have a list of time intervals to generate a new observation of a reduced event list and an updated GTI list.


gammapy/data/event_list.py
```diff
@@ -382,22 +384,37 @@ def select_energy(self, energy_range):
...
 
-    def select_time(self, time_interval):
-        """Select events in time interval.
+    def select_time(self, time_intervals):
+        """Select events in time intervals.
 
         Parameters
         ----------
-        time_interval : `astropy.time.Time`
+        time_intervals : `astropy.time.Time` or list of time_interval
```
AtreyeeS, line 392: If you have a `list of time_interval`, `time_intervals.value` won't work

gammapy/data/event_list.py
```diff
@@ -382,22 +384,37 @@ def select_energy(self, energy_range):
...
+
         time = self.time
-        mask = time_interval[0] <= time
-        mask &= time < time_interval[1]
+        mask = np.full(len(time), False)
+        if time_intervals.value.shape != (2,):
+            for time_interval in time_intervals:
+                submask = time_interval[0] <= time
+                submask &= time < time_interval[1]
+                mask |= submask
```
AtreyeeS, line 414: The mask would get written to the latest one.

AtreyeeS (commented):
Thanks @bkhelifi ! Would it not be better if the user could pass a `GTI` object instead of times ?

gammapy/data/event_list.py
```diff
@@ -382,22 +384,37 @@ def select_energy(self, energy_range):
         mask &= energy < energy_range[1]
         return self.select_row_subset(mask)
 
-    def select_time(self, time_interval):
-        """Select events in time interval.
+    def select_time(self, time_intervals):
```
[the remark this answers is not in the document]
bkhelifi: Damned, correct! Thanks

gammapy/data/event_list.py
registerrier: Here I think we probably have to consider checking what is an acceptable group of time intervals.
- do we accept only non overlapping intervals? 
  - If not what has to be done `and` or `or`? I think we should avoid the situation where filtering in a loop over time intervals and filtering over the list yield completely different results.
- what is the expected format of time intervals in general in gammapy? 
  - we are using lists of `Time` objects with shape (2,) in several places already. 
  - we could consider having a specific container for that to limit duplicate checks. After all, `GTI` and `TimeMapAxis` are already such kind of objects

gammapy/data/gti.py
```diff
@@ -306,36 +306,48 @@ def from_time_intervals(cls, time_intervals, reference_time=None):
 
         return cls.create(start, stop, reference_time)
 
-    def select_time(self, time_interval):
+    def select_time(self, time_intervals):
```
registerrier, line 309: technically you are implementing an `intersection` method, right?

registerrier (commented):
Thanks @bkhelifi ! This is a really important issue that needs to be solved.

My main comment here is whether we should not rely on something like `GTI` to perform validation of time intervals (e.g no overlaps) and to perform `union` and `intersection` operations.

gammapy/data/event_list.py
bkhelifi: I did not get your point... This is an OR made for each interval and it is updated at each loop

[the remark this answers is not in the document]
bkhelifi: Overlapping time intervals as input are forbidden... I precise in addition this behaviour as input

bkhelifi:
@registerrier, the PR is now ready to be reviewed!
I have added the feature associated to #5349.

For a very strange reason, the CI fails. But when testing the failed tests on my local machine, there is no error.

bkhelifi:
> CI fail is not due to #5354 ?

No idea, to be frank. I initially saw these errors with the dev version (exactly the same) and I fixed them. So, I do not understand these errors... If you have an idea, do not hesitate.

gammapy/data/utils.py
```diff
@@ -151,3 +154,36 @@ def get_irfs_features(
         features.meta["frame"] = "galactic"
 
     return features
+
+
+def check_time_intervals(time_intervals):
```
registerrier, line 159: This function is not working with the test as designed. Is it really needed?

Isn't the mechanism used for lightcurve, i.e. with `GTI.from_time_intervals` adapted here?

registerrier: I would also assume that a list of `Time` object (of length 2) is closer to what one would expect.

registerrier (commented):
Thanks @bkhelifi . There is apparently an issue with the `check_interval` method. I would suggest to try re-use functionality like `GTI.from_time_intervals` or at least to follow similar assumptions e.g. time intervals is a list of `Time` object.

gammapy/data/utils.py
bkhelifi: I have fixed a logic error in `check_time_intervals` (for a strange reason, it was not appearing in my local branch, I had to independently clone the branch somewhere else).
Otherwise, `GIT.from_time_intervals` does no check at all. This is why a new function is needed.

The latest CI fails are unrelated....

bkhelifi:
@registerrier : the CI py311-alldeps_noray, as well as the python 3.9 sphinx all-deps, crashes with "Error: Process completed with exit code 143", which is probably associated with memory/RAM issue. I can reproduce it on my computer...
Do you have a trick to go beyond and to have more output log?

gammapy/data/tests/test_data_store.py
```diff
@@ -335,3 +335,5 @@ def test_data_store_no_events():
     for obs in observations:
         assert not obs.events
         assert not obs.gti
+
+    utils.CHECK_OVERLAPPING_TI = True
```
QRemy, line 339: Why is this needed ? Does the user have to know is this is needed or not ? can't we determine automatically if the check if needed ?

examples/tutorials/data/hawc.py
```diff
@@ -100,6 +100,13 @@
 
 data_store.info()
 
+######################################################################
+# In this dataset, there are unfortunately overlapping GTIs. So
+# we are disabling the standard check of overlapping of GTIs.
+
+from gammapy.data import utils
+
+utils.CHECK_OVERLAPPING_TI = False
```
QRemy, line 109: What's the problem here ? is it specific to this HAWC dataset or something related to event types.
HAWC users should not have to set such global property. Opinion @LauraOlivera  ?

gammapy/irf/psf/tests/test_map.py
```diff
@@ -567,6 +567,11 @@ def test_psf_map_reco(tmpdir):
 
 @requires_data()
 def test_psf_map_reco_hawc():
+    # There are overlapping GTIs in this HAWC dataset
+    from gammapy.data import utils
+
+    utils.CHECK_OVERLAPPING_TI = False
+
```
QRemy, line 574: Here the test is only about the PSF, why do we need a gti check here ?

examples/tutorials/data/hawc.py
bkhelifi: This is an issue with our HAWC datasets, rather to HAWC in general

gammapy/irf/psf/tests/test_map.py
bkhelifi: good point! fixed..

examples/tutorials/data/hawc.py
LauraOlivera: why?

gammapy/data/utils.py
```diff
@@ -151,3 +159,39 @@ def get_irfs_features(
...
+    else:
+        for xx in ti:
+            if not np.all([isinstance(_, Time) for _ in xx]):
+                return False
+        if not CHECK_OVERLAPPING_TI:
+            return True
+        for xx in combinations(ti, 2):
+            i1 = [xx[0][0].to_value("gps"), xx[0][1].to_value("gps")]
+            i2 = [xx[1][0].to_value("gps"), xx[1][1].to_value("gps")]
+            if interval_overlap_length(i1, i2) > 0.0:
```
QRemy, line 192: this might trigger precision error, maybe it explains why it fails for the HAWC dataset.

gammapy/data/tests/test_data_store.py
QRemy: The overlap catched by the check is of the order of 0.0001s for gtis whitin one observation, so I suspect the failure just comes from a precision error.

gammapy/data/utils.py
bkhelifi: No, for HAWC there are really overlapping GTI

gammapy/data/observations.py
```diff
@@ -166,6 +168,8 @@ def available_irfs(self):
     @property
     def events(self):
         """Event list of the observation as an `~gammapy.data.EventList`."""
+        if self.obs_filter.time_filter is None and self._gti is not None:
+            self.obs_filter.time_filter = self._gti.time_intervals
```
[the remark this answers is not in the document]
bkhelifi: This line implements the feature mentionned in #5349. To my knowleged, there is no precision error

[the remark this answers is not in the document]
QRemy: Depends at which precision the events time is known/saved.

gammapy/data/gti.py
bkhelifi: yes, indeed... the intersection between the observattory GTIs and the user's time intervals

gammapy/data/observations.py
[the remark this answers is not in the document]
bkhelifi: The accuracy of the times of GTIs and events should be identical, so there is no in-depth reason to not have proper GTIs. One could add an accuracy tolerance, that I will set to 1/20 of the lowest psr period (20ms/10=1ms? is it enough?)

gammapy/data/tests/test_data_store.py
bkhelifi: events and GTIs should have the same accuracy. Have we made an error on the events time accuracy?

gammapy/data/observations.py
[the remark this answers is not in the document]
QRemy: Setting `interval_overlap_length(i1, i2) > 1e-3` in `check_time_intervals` seems to not fail right away but the loop over combinations of gti ` combinations(ti, 2)` is extremely long to compute because there is about 4e4 gtis in one obs so 16e8 combinations to test.

It should be part of the format convention that events are encompassed in the provided gti and done at the DL3 production level. I would argue that it is not the responsibility of gammapy to check that every time someone open a file, so by default we should not filter the events.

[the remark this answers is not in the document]
bkhelifi: I agree... even if I put it here for statistical reasons. It is up to us to decide.

registerrier:
Where do we stand here @bkhelifi ?

bkhelifi:
@registerrier: the PR is updated accordingly to the dev call comments and can be reviewed. Additional tests have been added to increase the code coverage.
Once merged, the PR #5511 can be continued...

gammapy/data/tests/test_utils.py
```diff
@@ -0,0 +1,36 @@
...
+    names = ["edisp-bias", "edisp-res", "psf-radius"]
+    features_irfs = get_irfs_features(
+        obs,
+        energy_true="1 TeV",
+        position=position,
+        names=names,
+    )
+    assert_allclose(features_irfs[0]["edisp-bias"], 0.11587, rtol=1.0e-4)
+    assert_allclose(features_irfs[0]["edisp-res"], 0.36834, rtol=1.0e-4)
+    assert_allclose(features_irfs[0]["psf-radius"], 0.14149, rtol=1.0e-4)
```
QRemy, line 36: This seems unrelated to time filtering. Please move to a separate PR.

QRemy (approved):
Thanks @bkhelifi, please next time open multiple PR.

gammapy/data/filters.py
```diff
@@ -65,16 +64,20 @@ def filter_events(self, events):
 
         Parameters
         ----------
-        events : `~gammapy.data.EventListBase`
+        events : `gammapy.data.EventListBase`
```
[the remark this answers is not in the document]
QRemy: As did not agree on how to show this in the documentation see #5490
this should should not be done here.

bkhelifi:
@QRemy, @AtreyeeS and/or @registerrier, could you review and merge quickly this PR, such that I can finish the PR #5511???
Time is running.... Thanks per advance

gammapy/data/tests/test_utils.py
bkhelifi: Indeed, it allows to add some missing time coverage

QRemy: If this test add missing time coverage it means that unit tests for time are missing because `get_irfs_features` doesn't do time filtering so if we modify/remove this test for other reason we will loose coverage on time related functionnalities.

QRemy: do you improve the coverage simply form these 3 lines using HESS data where gtis are defined ?

```
    data_store = DataStore.from_dir("$GAMMAPY_DATA/hess-dl3-dr1/")
    obs_table = data_store.obs_table.select_observations(selection)
    obs = data_store.get_observations(obs_table["OBS_ID"][:1])
```

bkhelifi: OK, I remove it... Does not matter

registerrier (commented):
Thanks @bkhelifi . Sorry for the delay. This is just a rapid comment. In general, it is very difficult to review PR that touch 20 files. It is important to separate things as much as possible. For instance avoid changing unrelated docstrings and try to split the necessary changes.

Here my main issues (sorry for being annoying but this is important) are connected with:
- the ambiguity this introduces between two different methods: `Observations.select_time()` and `Observation.select_time()`, which is introduced here. The first one will split an observation for every time interval passed, yielding an observation per time interval, while the second, will yield one observation filtered over all time intervals. This is a very confusing situation and we need to clarify how this should be handled in general.
- the implicit definition of what are time intervals should be discussed further, especially if they are to be checked. I don't think that a list of `Time` is the best choice. Ideally a `Time` object with the correct shape, or something that can be turned into this seems a more natural choice.

gammapy/analysis/tests/test_analysis.py
```diff
@@ -406,7 +406,7 @@ def test_analysis_3d_joint_datasets():
     assert_allclose(
         analysis.datasets[0].background_model.spectral_model.norm.value,
         1.031743694988066,
-        rtol=1e-6,
+        rtol=1e-3,
```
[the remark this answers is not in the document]
bkhelifi: well seen! A mistake. Thanks

bkhelifi:
> Thanks @bkhelifi . Sorry for the delay. This is just a rapid comment. In general, it is very difficult to review PR that touch 20 files. It is important to separate things as much as possible. For instance avoid changing unrelated docstrings and try to split the necessary changes.
> 
> Here my main issues (sorry for being annoying but this is important) are connected with:
> 
> * the ambiguity this introduces between two different methods: `Observations.select_time()` and `Observation.select_time()`, which is introduced here. The first one will split an observation for every time interval passed, yielding an observation per time interval, while the second, will yield one observation filtered over all time intervals. This is a very confusing situation and we need to clarify how this should be handled in general.

The aims of these functions have not changed from the original code. The PR just aims to use list of time intervals instead of one time interval. The confusing situation is originel. 

> * the implicit definition of what are time intervals should be discussed further, especially if they are to be checked. I don't think that a list of `Time` is the best choice. Ideally a `Time` object with the correct shape, or something that can be turned into this seems a more natural choice.

It would have been better to know that this PR is not good since june 2024! Thanks a lot

QRemy:
> * the ambiguity this introduces between two different methods: `Observations.select_time()` and `Observation.select_time()`, which is introduced here.

This can by solve easily by renaming `Observation.select_time()` to `Observation.filter_time()`.

PR #5308, opened 2024-06-13
Set a diagonal Edisp when `MapDataset.edisp` is `None`

### LINKED ISSUE #5274: FluxPointsEstimator fails if no edisp is set
**Gammapy version**
present dev

**Bug description**

Running `FluxPointsEstimator` on a dataset without an `edisp` gives a very confusing error message `ValueError: Can only stack equivalent maps or cutout of the same map.`

**To Reproduce**

Set `edisp=None` for any dataset (with e_true=e_reco) and proceed with flux points estimation in multiple energy bands (normal use case).

**Expected behaviour**
The analysis fails (as expected) during the fitting stage if `energy_true` has a different binning from the `energy` axis with a more understandable error 
```
File ~/Gammapy-dev/gammapy/gammapy/datasets/evaluator.py:245, in MapEvaluator._edisp_diagonal(self)
    242 @lazyproperty
    243 def _edisp_diagonal(self):
    244     return EDispKernel.from_diagonal_response(
--> 245         energy_axis_true=self.edisp.axes["energy_true"],
    246         energy_axis=self.edisp.axes["energy"],
    247     )

AttributeError: 'NoneType' object has no attribute 'axes'
```
However, if the `energy_axis` and the `energy_axis_true` have same binning, the fit proceeds fine and then the very confusing error is raised during flux points estimation.

1. Either the flux estimator (and also the fit) should fail with a decent error message consistently for no edisp OR
2. the diagonal edisp can be adapted with the `energy_axis` here 
https://github.com/gammapy/gammapy/blob/e9b22cdff573768ac2ab1e2a4a18250071ba3b60/gammapy/datasets/evaluator.py#L243
to use the dataset energy axis.

fabiopintore:
This PR should resolve #5274.
It introduces the possibility to create a diagonal edisp matrix when the `dataset` does not have one. This will avoid that `Fit` and `FluxPointsEstimator ` will fail when the edisp is not defined.
Tests have been updated accordingly.

### DIFF
--- a/gammapy/datasets/tests/test_map.py
+++ b/gammapy/datasets/tests/test_map.py
@@ -2097,6 +2097,7 @@ def test_to_masked():
     assert_allclose(d1.counts.data.sum(), 170)
 
 
+@requires_data()
 def test_diagonal_edisp(geom_etrue, caplog):
     axis = MapAxis.from_energy_bounds(1, 10, 2, unit="TeV")
     geom = WcsGeom.create(npix=(10, 10), binsz=0.05, axes=[axis])
--- a/gammapy/datasets/tests/test_map.py
+++ b/gammapy/datasets/tests/test_map.py
@@ -2104,5 +2104,5 @@ def test_diagonal_edisp(geom_etrue, caplog):
     psf = get_psf()
     exposure = get_exposure(geom_etrue)
     dataset = MapDataset(counts=counts, exposure=exposure, edisp=None, psf=psf)
-    #    assert "Edisp is not defined. A diagonal response matrix will be set." in [record.levelname for record in caplog.records]
+    assert "WARNING" in [record.levelname for record in caplog.records]
     assert dataset.edisp is not None
--- a/gammapy/datasets/tests/test_map.py
+++ b/gammapy/datasets/tests/test_map.py
@@ -2095,3 +2095,14 @@ def test_to_masked():
     )
     d1 = datasetonoff.to_masked()
     assert_allclose(d1.counts.data.sum(), 170)
+
+
+def test_diagonal_edisp(geom_etrue, caplog):
+    axis = MapAxis.from_energy_bounds(1, 10, 2, unit="TeV")
+    geom = WcsGeom.create(npix=(10, 10), binsz=0.05, axes=[axis])
+    counts = Map.from_geom(geom, data=1)
+    psf = get_psf()
+    exposure = get_exposure(geom_etrue)
+    dataset = MapDataset(counts=counts, exposure=exposure, edisp=None, psf=psf)
+    #    assert "Edisp is not defined. A diagonal response matrix will be set." in [record.levelname for record in caplog.records]
+    assert dataset.edisp is not None
--- a/gammapy/datasets/tests/test_spectrum.py
+++ b/gammapy/datasets/tests/test_spectrum.py
@@ -72,6 +72,7 @@ def test_spectrum_dataset_fits_io(spectrum_dataset, tmp_path):
     )
     hdulist = spectrum_dataset.to_hdulist()
     actual = [hdu.name for hdu in hdulist]
+    print(actual)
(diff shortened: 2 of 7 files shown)


gammapy/datasets/tests/test_map.py
```diff
@@ -2095,3 +2095,14 @@ def test_to_masked():
...
+
+
+def test_diagonal_edisp(geom_etrue, caplog):
+    axis = MapAxis.from_energy_bounds(1, 10, 2, unit="TeV")
+    geom = WcsGeom.create(npix=(10, 10), binsz=0.05, axes=[axis])
+    counts = Map.from_geom(geom, data=1)
+    psf = get_psf()
+    exposure = get_exposure(geom_etrue)
+    dataset = MapDataset(counts=counts, exposure=exposure, edisp=None, psf=psf)
+    #    assert "Edisp is not defined. A diagonal response matrix will be set." in [record.levelname for record in caplog.records]
```
AtreyeeS, line 2107: Why did you comment this?

gammapy/datasets/tests/test_spectrum.py
```diff
@@ -96,8 +101,6 @@ def test_spectrum_dataset_fits_io(spectrum_dataset, tmp_path):
     assert_allclose(
         spectrum_dataset.npred_background().data, dataset_new.npred_background().data
     )
-    assert dataset_new.edisp is None
-    assert dataset_new.edisp is None
     assert dataset_new.name == "test"
```
AtreyeeS, line 104: maybe add `dataset_new.edisp is not None`

AtreyeeS (dismissed):
Thanks @fabiopintore ! Very small comments inline

gammapy/datasets/tests/test_spectrum.py
```diff
@@ -72,6 +72,7 @@ def test_spectrum_dataset_fits_io(spectrum_dataset, tmp_path):
     )
     hdulist = spectrum_dataset.to_hdulist()
     actual = [hdu.name for hdu in hdulist]
+    print(actual)
```
bkhelifi, line 75: remove the print

bkhelifi (changes requested):

MRegeard:
Thanks @fabiopintore. I am wondering wether the edisp code in the `__init__` function should go in a setter. It would make the `__init__` simpler and also ensure type check when setting an edisp after instantiation. 
I would also add a comment in the docstring that specify behaviour when edisp is None (in addition to the warning raised), so that it is transparent to doc readers.

bkhelifi:
Creating a default edisp when nothing is precised looks strange for me: if someone forgets to set an edisp, it should realize with a crash.... If someone wants a diagonal edisp, it can not it easly.
This fix for the FluxPointsEstimator migh be dangeourous for the Data Reduction step... I am wondering whether one should not add a diagonal matrix if edisp is None inside the FluxPointsEstimator. Any opinion?

QRemy:
> @bkhelifi : The issue is that in the current implementation, `MapEvaluator` silently creates a diagonal edisp in case no `edisp` is present, and, if `e_true = e_reco` the fit proceeds normally and a very confusing error is raised during the flux estimation. This PR makes the process cleaner and consistently warns the user about diagonal edisp creation.
> 

In principle the `MapEvaluator` silently creates a diagonal edisp only if model.apply_irf["edisp"]==False

AtreyeeS:
> In principle the MapEvaluator silently creates a diagonal edisp only if model.apply_irf["edisp"]==False

But the fit still works if `energy_true = energy_reco` , model.apply_irf["edisp"]==True and `dataset.edisp=None`
I think getting this case to fail with a `ValueError(model.apply_irf["edisp"] is True but no edisp set on dataset)` is also an acceptable solution

AtreyeeS:
Super seeded by https://github.com/gammapy/gammapy/pull/5342

Issue #5306, opened 2024-06-12
Expose `FitResults` in the documentation


AtreyeeS:
In our documentation, we expose `datasets.models` , `dataset.models`, `models`, `model` to access the parameter values/errors. All of this should be replaced by consistently calling `FitResult.models`. 
It should also be mentioned that `datasets.write` does not write the complete covariance, but `FitResult.models.covariance` does.

bkhelifi:
The read method of the FitResult should be added into a separated PR

bkhelifi:
@registerrier, @adonath and @QRemy: in order to prevent using Datasets.models.covariance that is not correct after the fit, one should obviously update the tutorials by using FitResult.models.covariance. This should be done into a dedicated PR.
But, for users knowing Gammapy, they might not read the new tutorials or the release logs... So, I propose that:
- one renders all the 'models' parameters private members of the dataset classes,
- remove the setter 'models',
- add a function 'set_models()'.
This is a (easy) change of API that will prevent further confusions... 
Any thought?

Astro-Kirsty:
I will add the read function and work on making sure we are calling `FitResult.models` rather than `dataset.models` in the docs

Astro-Kirsty:
With the update in https://github.com/gammapy/gammapy/pull/5317 we do not need to change the above. 

See #5319 for the read function

Astro-Kirsty:
With the new implementation by @QRemy, these issues should be fixed?

QRemy:
Resolved by #5317

PR #5311, opened 2024-06-13
Add the covariance in the write function of FitResult

bkhelifi:
**Description**

This pull request is associated to the issue #5306. All is in the title...

**Dear reviewer**
The read function of FitResult needs to be created... into a separated PR.

### DIFF
--- a/gammapy/modeling/fit.py
+++ b/gammapy/modeling/fit.py
@@ -2,6 +2,7 @@
 import html
 import itertools
 import logging
+from os.path import split
 import numpy as np
 from astropy.table import Table
 import yaml
@@ -165,7 +166,7 @@ class Fit:
 
         Returns
         -------
-        fit_result : `FitResult`
+        fit_result : `~gammapy.modeling.FitResult`
             Fit result.
         """
 
@@ -799,6 +800,7 @@ class FitResult:
         full_output=False,
         overwrite_templates=False,
         checksum=False,
+        covariance=True,
     ):
         """Write to file.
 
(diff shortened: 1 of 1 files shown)


QRemy:
Looking at the code of FitResult.write it seems that there is some duplication with `Models.write()`.
Why not adding `optimize_results` and `covariance_results` options to  `Models.write()` so it can add these entries in the model file directly and so the `FitResult.write` will be reduced to `Models.write(..., optimize_results=optimize_results.to_dict(), covariance_results..to_dict()) `

registerrier (commented):
Thanks @bkhelifi . This looks good. It probably misses a test though.

Maybe a general question: can't we include the covariance directly into the main file?

registerrier:
> Looking at the code of FitResult.write it seems that there is some duplication with `Models.write()`. Why not adding `optimize_results` and `covariance_results` options to `Models.write()` so it can add these entries in the model file directly.
> 
> Then the `FitResult.write` will be reduced to `return Models.write(optimize_results=optimize_results.to_dict() covariance_results.to_dict(), **kwargs) ` and we can remove `FitResult.to_dict` and `FitResult.to_yaml`

This would mean that `FitResult` is not really needed, no?

At the moment, we don't really provide a specific object for `confidence`, `contours`, `scan` results. These might go into the `FitResult` too, no? Or is this too connected to a single parameter, hence to parameter estimation?

QRemy:
> At the moment, we don't really provide a specific object for `confidence`, `contours`, `scan` results. These might go into the `FitResult` too, no? Or is this too connected to a single parameter, hence to parameter estimation?

Any of these quantities make sense only with respect to a given model so they have to be exported with them, then we can just give an extra dict to  Models.write as done in #5325.

QRemy:
I will close this as #5325 resolves it.

PR #5312, opened 2024-06-13
More complex background spectral model exposed

### LINKED ISSUE #4947: Tutorial for more complex background models
Currently we expose how to utilise the `FoVBackgroundMaker` with a `PowerLawNormSpectralModel` that have both tilt and norm unfrozen. 

It could also be useful for users to understand how to correctly implement a more complex background. For example utilising the `PiecewiseNormSpectralModel`.

Astro-Kirsty:
This resolves  #4947.

It adds in the docs that it is possible to utilise a more complex model (PieceWise) in your FoVbkgmaker. 

I am open to also adding the priors as mentioned in the coding sprint, however, if we had this is the data reduction step it could force some of the fitted values to be better than they are. And therefore people will not know this is the case.
i.e. setting a prior on the `norm` such that the values vary between +-0.2, it could mean without the prior the norm wants to fit at 1.5, but if the prior is added it would not reach this and only be at  i.e. 1.2

Let me know your thoughts?

### DIFF
--- a/docs/user-guide/makers/fov.rst
+++ b/docs/user-guide/makers/fov.rst
@@ -87,6 +87,9 @@ in the `fov_bkg_maker`.
     )
 
 
+Note: to prevent poorly constrained `norm` parameters or large variance in
+the last bins, the binning should be adjusted to have wider bins at higher energies.
+This ensures there are enough statistics per bin, enabling the fit to converge.
 
 
 .. minigallery:: gammapy.makers.FoVBackgroundMaker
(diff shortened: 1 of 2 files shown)


AtreyeeS:
Adding in the `rst` page is a nice addition! I think you can show the use of priors in the rst file.

> it could mean without the prior the norm wants to fit at 1.5, but if the prior is added it would not reach this and only be at i.e. 1.2
That's the point of using priors. Otherwise, say, the last bin of the piecewise model wants to put norm at 1.5, and the bin before that at 0.1... that's not how we think our background is, so we "force" to be behave within our expectations. 
Of course, the `sigma` of the prior has to be obtained logically, and this should be mentioned in the docs

bkhelifi (approved):

Astro-Kirsty:
@AtreyeeS, but would that be wrongly setting a norm of 1.2 when it should really be 1.5?
To me, this is something to be adjusted afterwards, i.e. you obtain all your norms and tilts and then investigate the outliers within your observations

QRemy (commented):
Thanks @Astro-Kirsty, maybe mentions that in order to avoid having poorly constrained norms parameters or the large variance in the last bins one can set wider bins at high energy so there is enough stat per bin for the fit to converge.

QRemy (approved):
Thanks no further comments, let's discuss how to advertise the prior later (in a separate PR).

PR #5313, opened 2024-06-13
Draft example of a possible event read write with validation

registerrier:
**Description**

This pull request ...

**Dear reviewer**

### DIFF
--- a/gammapy/data/io.py
+++ b/gammapy/data/io.py
@@ -4,84 +4,7 @@ from astropy.io import fits
 from astropy.table import Table
 from pydantic import BaseModel, ValidationError, validator
 from gammapy.utils.scripts import make_path
-
-
-class ColumnType(BaseModel):
-    dtype: str
-    unit: Optional[str]
-    required: bool = False
-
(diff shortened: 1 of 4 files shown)


gammapy/data/io.py
```diff
@@ -0,0 +1,112 @@
...
+    CONV_RA: Optional[float]
+    CONV_DEC: Optional[float]
+    TRGRATE: Optional[float]
+    ZTRGRATE: Optional[float]
+    MUONEFF: Optional[float]
+    BROKPIX: Optional[float]
+    AIRTEMP: Optional[float]
+    PRESSURE: Optional[float]
+    RELHUM: Optional[float]
+    NSBLEVEL: Optional[float]
```
bkhelifi, line 56: Maybe, this long list could be into a file, the filename having also the version of the format... If you add a version id in the constructor of this class, you can have a generic validator of a given format and for all its versions.

registerrier:
@MSCarrasco is preparing a detailed framework to perform format validation in I/O. Closing this prototype.

PR #5316, opened 2024-06-14
Add CovarianceMixin for multicomponent models classes

QRemy:
Move the covariance related properties to a CovarianceMixin class and use it for  all multi-component models classes including CompoundSpectralModel.

### DIFF
--- a/gammapy/modeling/covariance.py
+++ b/gammapy/modeling/covariance.py
@@ -210,3 +210,32 @@ class Covariance:
 
     def __array__(self):
         return self.data
+
+
+class CovarianceMixin:
+    """Mixin class for covariance property on multi-components models"""
+
+    def _check_covariance(self):
+        if not self.parameters == self._covariance.parameters:
+            self._covariance = Covariance.from_stack(
+                [model.covariance for model in self._models]
+            )
+
+    @property
+    def covariance(self):
+        """Covariance as a `~gammapy.modeling.Covariance` object."""
+        self._check_covariance()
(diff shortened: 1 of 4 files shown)


adonath (commented):
Thanks @QRemy, this looks reasonable to me! However I think we should only back port the bugfix and not the new mixin class to avoid future merge conflicts when back porting other fixes...

QRemy:
Requires #5323 before merging.

gammapy/modeling/covariance.py
```diff
@@ -210,3 +210,32 @@ def __str__(self):
...
+        """Covariance as a `~gammapy.modeling.Covariance` object."""
+        self._check_covariance()
+
+        for model in self._models:
+            self._covariance.set_subcovariance(model.covariance)
+
+        return self._covariance
+
+    @covariance.setter
+    def covariance(self, covariance):
```
bkhelifi, line 235: Can you add a docstring defining the type of `covariance`

bkhelifi (changes requested):
Thanks for this code cleaning and rationalization. Beyond minor comments on docstrings, it looks ok for me.

gammapy/modeling/covariance.py
QRemy: The setter just display the same output as its associated property and the type of `Covariance` is described there. I think we have the same behaviour everywhere. Probably better not change too much docstring in a bugfix that has to beck backported.

PR #5317, opened 2024-06-14
Fix datasets.models.covariance after fit to match FitResults.models.covariance

QRemy:
Fix datasets.models.covariance after fit to match FitResults.models.covariance.
It simply add a  private `_covariance` attribute on datasets that will be used to set the covariance properly in `.models` property  if the parameters of the covariance and the datasets.models match.

### DIFF
--- a/gammapy/datasets/core.py
+++ b/gammapy/datasets/core.py
@@ -184,7 +184,7 @@ class Datasets(collections.abc.MutableSequence):
         The order of the unique models remains.
         """
         if models:
-            self._covariance = models.covariance
+            self._covariance = DatasetModels(models).covariance
         for dataset in self:
             dataset.models = models
 
--- a/gammapy/datasets/core.py
+++ b/gammapy/datasets/core.py
@@ -188,9 +188,6 @@ class Datasets(collections.abc.MutableSequence):
         for dataset in self:
             dataset.models = models
 
-    def set_covariance(self, covariance):
-        self._covariance = covariance
-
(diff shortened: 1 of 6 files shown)


bkhelifi (dismissed):
Thanks a lot. It looks good! Apparently an if statement is not tested correctly. Could you add a test, just to be sure? The issue is an important one and one should be sure to fix it definitely

QRemy:
>  Could you add a test, just to be sure? The issue is an important one and one should be sure to fix it definitely

Done also modified a bit the code so the solution for actors is simpler.

gammapy/datasets/core.py
```diff
@@ -178,6 +183,8 @@ def models(self, models):
         Duplicate model objects have been removed.
```
bkhelifi, line 183: For me, the docstring is a bit obscure... The setter is about unique models: what is a unique model? The docstring describes in fact what does the setter and the input parameter is not precised. Can you enlight a bit this docstring?

bkhelifi (approved):
In between the different PRs about covariance, it is a bit hard to have a global view. However, it seems wise and looks ok for me.
I have only a minor comment on a docstring.

gammapy/datasets/core.py
QRemy: The setter just display the same output as its associated property. I think we have the same behaviour everywhere. Probably better not change too much docstring in a bugfix that has to beck backported.

PR #5318, opened 2024-06-14
[JAX prototype]: some bits on adapting maps

registerrier:
**Description**

This pull request is work made with @Astro-Kirsty during the coding sprint to explore what it takes to have Maps carry a JAX Array as `data`.
It is there for discussion and not meant to be merged.

**Dear reviewer**


QRemy:
I will close all the JAX prototype PR and edit their name so we can easily find them back (and maybe add a label)

PR #5319, opened 2024-06-14
Add read to `FitResult`

Astro-Kirsty:
For some cases the total_stat was saving as a `numpy.float64` which the yaml dump cannot understand
[See here](https://github.com/gammapy/gammapy/assets/60767360/d900ba2e-06fd-4e79-be9b-f1f3dde0743c)
Therefore we must ensure the `total_stat` is a `float` in the dictionary. 

For now I have a read function of the yaml. But it is unclear to me how we would like the output
Feel free to add your opinions here.

### DIFF
--- a/gammapy/modeling/fit.py
+++ b/gammapy/modeling/fit.py
@@ -2,10 +2,12 @@
 import html
 import itertools
 import logging
+import warnings
+from os.path import split
 import numpy as np
 from astropy.table import Table
 import yaml
-from gammapy.utils.check import add_checksum
+from gammapy.utils.check import add_checksum, verify_checksum
 from gammapy.utils.metadata import CreatorMetaData
 from gammapy.utils.pbar import progress_bar
 from gammapy.utils.scripts import make_path
@@ -776,6 +778,9 @@ class FitResult:
         output = {}
         if self.optimize_result is not None:
             output["optimize_result"] = self.optimize_result.to_dict()
+            output["optimize_result"]["total_stat"] = float(
+                output["optimize_result"]["total_stat"]
+            )
         if self.covariance_result is not None:
             output["covariance_result"] = self.covariance_result.to_dict()
         output["models"] = models_dict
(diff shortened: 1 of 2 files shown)


gammapy/modeling/tests/test_fit.py
```diff
@@ -356,3 +356,11 @@ def test_write(tmpdir):
...
     assert "covariance_result" not in result_dict
     result.write(path=tmpdir / "test-fit-result.yaml", overwrite=True)
+
+    result_read = FitResult.read(filename=tmpdir / "test-fit-result.yaml")
+
+    assert result_read["optimize_result"]["nfev"] == result.optimize_result.nfev
+    assert (
+        result_read["optimize_result"]["total_stat"]
+        == result.optimize_result.total_stat
+    )
```
cgalelli, line 366: I would just add a test catching the checksum fail here

cgalelli (commented):
Thanks @Astro-Kirsty , as annotated I would maybe just add the test for the checksum verification fail.

gammapy/modeling/tests/test_fit.py
Astro-Kirsty: Thanks @cgalelli 

This actually needs to be updated after all the covariances changes have been added!

gammapy/modeling/fit.py
```diff
@@ -825,6 +830,35 @@ def write(
...
+        data = yaml.safe_load(yaml_str)
+        checksum_str = data.pop("checksum", None)
+        if checksum:
+            yaml_str = yaml.dump(
+                data, sort_keys=False, indent=4, width=80, default_flow_style=False
+            )
+            if not verify_checksum(yaml_str, checksum_str):
+                warnings.warn("Checksum verification failed.", UserWarning)
+
+        return data
```
QRemy, line 860: This method is equivalent to `gammapy.utils.scripts.read_yaml`, so if the goal is just to  get a dict we don't need more.

Here .read and .to_yaml  returns a dict not an instance of the class so they are not classmethod.
The write method does not export all the mandatory arguments of OptimizeResult and CovarianceResult __init__ so we cannot really create a read method right now.

QRemy (commented):
Thanks @Astro-Kirsty, I see several issues with this right now :

gammapy/modeling/fit.py
QRemy: I would also say that with the covariance bug fixed in  #5317, the need to read back the output of FitResult is less obvious.

Astro-Kirsty: I agree with not really needing the read of the FitResult, now with all the other updates in the PRs have been made.

I suggest we close?

PR #5320, opened 2024-06-14
Fix parameters unique names

QRemy:
Attempt to fix parameters unique names.

### DIFF
--- a/gammapy/modeling/models/cube.py
+++ b/gammapy/modeling/models/cube.py
@@ -167,7 +167,7 @@ class SkyModel(ModelBase):
     def parameters_unique_names(self):
         """List of unique parameter names. Return formatted as par_type.par_name."""
         names = []
-        for model in enumerate(self._models):
+        for model in self._models:
             for par_name in model.parameters_unique_names:
                 components = [model.type, par_name]
                 name = ".".join(components)
@@ -723,7 +723,7 @@ class FoVBackgroundModel(ModelBase):
     def parameters_unique_names(self):
         """List of unique parameter names. Return formatted as par_type.par_name."""
         names = []
-        for model in enumerate(self._models):
+        for model in self._models:
             for par_name in model.parameters_unique_names:
                 components = [model.type, par_name]
                 name = ".".join(components)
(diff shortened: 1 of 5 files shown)


Astro-Kirsty (approved):
Thanks @QRemy. This is a really nice addition

cgalelli (approved):
Thanks @QRemy very nice change

gammapy/modeling/models/cube.py
```diff
@@ -703,6 +719,17 @@ def parameters(self):
             parameters.append(self.spatial_model.parameters)
```
registerrier, line 719: technically if you introduce the `_models` property, you could simplify this too.

registerrier (commented):
Thanks @QRemy. 
Could there be a way to avoid the code duplication of the parameter_unique_names mechanism? 
Does it require an adaptation in the API tutorial or in the user guide, to explain how the parameter names are built?

gammapy/modeling/models/cube.py
```diff
@@ -689,6 +700,11 @@ def spatial_model(self):
         """Spatial norm model."""
         return self._spatial_model
 
+    @property
+    def _models(self):
+        models = self.spectral_model, self.spatial_model
```
AtreyeeS, line 705: needs the temporal model as well

gammapy/modeling/tests/test_fit.py
```diff
@@ -221,7 +221,7 @@ def test_stat_profile():
     dataset.models.parameters["x"].scan_n_values = 3
     result = fit.stat_profile(datasets=[dataset], parameter="x")
 
-    assert_allclose(result["test.model.x_scan"], [0, 2, 4], atol=1e-7)
+    assert_allclose(result["test.x_scan"], [0, 2, 4], atol=1e-7)
```
AtreyeeS, line 224: what is the naming logic here? does this solve the issue of whether `x` belongs to `model1` or `model2`?

AtreyeeS (commented):
Thanks @QRemy 
Can `def parameters_unique_names` reside in `ModelBase` to avoid code duplication?

gammapy/modeling/models/cube.py
QRemy: not for FoVBackgroundModel

gammapy/modeling/models/cube.py
QRemy: OK I will see if a larger cleanup is possible after all the covariance related PRs are in.

QRemy:
> Thanks @QRemy Can `def parameters_unique_names` reside in `ModelBase` to avoid code duplication?

It is defined in `ModelBase` but has to be refined in for each composite models.

PR #5321, opened 2024-06-14
Fix cutout_template_models to work if models is None

### LINKED ISSUE #5213: split_dataset fails if no template model is set and split_template_models=True
**Gammapy version**
v1.2

**Bug description**
When running `split_dataset(dataset, 0.3*u.deg)` on a `MapDataset` with no `TemplateSpatialModel` set, an exception is raised.
This is because by default `split_template_models=True`. 

**Expected behavior**
It shouldn't break and proceed as if `split_template_models=False`.

**To Reproduce**

QRemy:
Resolves #5213


QRemy:
Backport to v1.0.x is not needed `cutout_template_models` feature was introduced after.

PR #5323, opened 2024-06-18
Fix covariance for CompoundSpectralModel 

QRemy:
Set correctly the sub-covariance on the individual models of   CompoundSpectralModel.
As suggested by @adonath this should go in before  #5316, then the Mixin class can be introduced to reduce code duplication.


AtreyeeS (approved):
Thank you @QRemy ! No comments

Astro-Kirsty (approved):
Thanks @QRemy 
I doubled check within my code too, and this works nicely!

QRemy:
Fails are unrelated merging now

PR #5324, opened 2024-06-18
Fix tests fail due to precision

QRemy:
-Fix tests fail due to precision
-impose numpy<2 otherwise fails with conda and docs build.


registerrier:
My bad. I had not seen that part of the issues in #5326 were addressed here.
I think we should try not to impose numpy<2.0 for all cases. See #5327

registerrier (approved):
Thanks @QRemy ! I have removed the general change on the setup.cfg, as we should try to support numpy 2.0.

registerrier:
Merging now. Remaining issues should hopefully be solved in #5327

PR #5325, opened 2024-06-18
Simplify and cleanup yaml IO

QRemy:
Simplify and cleanup yaml IO:
-use more the functions in utlis so we don't rewrite the code related to checksum, overwrite, log,  (and maybe meta data creator? )...
-simplify FitResult. write to use directly Models.write with an extra_dict (so we don't need #5311)

### DIFF
--- a/gammapy/analysis/config.py
+++ b/gammapy/analysis/config.py
@@ -243,19 +243,40 @@ class AnalysisConfig(GammapyBaseConfig):
 
     @classmethod
     def read(cls, path):
-        """Read from YAML file."""
+        """Read from YAML file.
+
+        Parameters
+        ----------
+        path : str
+            input filepath
+        """
         config = read_yaml(path)
         config.pop("metadata", None)
         return AnalysisConfig(**config)
 
     @classmethod
     def from_yaml(cls, config_str):
-        """Create from YAML string."""
+        """Create from YAML string.
+
+        Parameters
+        ----------
+        config_str : str
+            yaml str
+
+        """
         settings = yaml.safe_load(config_str)
         return AnalysisConfig(**settings)
 
     def write(self, path, overwrite=False):
-        """Write to YAML file."""
+        """Write to YAML file.
+
+        Parameters
+        ----------
+        path : `pathlib.Path` or str
+            Path to write files.
+        overwrite : bool, optional
+            Overwrite existing file. Default is False.
+        """
         yaml_str = self.to_yaml()
         write_yaml(yaml_str, path, overwrite=overwrite)
 
--- a/gammapy/analysis/config.py
+++ b/gammapy/analysis/config.py
@@ -256,7 +256,8 @@ class AnalysisConfig(GammapyBaseConfig):
 
     def write(self, path, overwrite=False):
         """Write to YAML file."""
-        write_yaml(self.to_yaml(), path, overwrite=overwrite)
+        yaml_str = self.to_yaml()
+        write_yaml(yaml_str, path, overwrite=overwrite)
 
     def to_yaml(self):
         """Convert to YAML string."""
--- a/gammapy/analysis/config.py
+++ b/gammapy/analysis/config.py
@@ -11,7 +11,7 @@ from astropy import units as u
 import yaml
 from pydantic import BaseModel, ConfigDict
 from gammapy.makers import MapDatasetMaker
-from gammapy.utils.scripts import make_path, read_yaml
+from gammapy.utils.scripts import read_yaml, to_yaml, write_yaml
 from gammapy.utils.types import AngleType, EnergyType, PathType, TimeType
 
 __all__ = ["AnalysisConfig"]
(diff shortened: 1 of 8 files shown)


bkhelifi:
This feature looks nice... can you finish the PR such that all the tests are OK? Take care of the coverage. Thanks @QRemy

Astro-Kirsty (approved):
Thanks @QRemy 
These are some nice adaptations

bkhelifi (approved):
Thanks @QRemy. It looks OK for me ;)

gammapy/analysis/config.py
```diff
@@ -255,19 +256,12 @@ def from_yaml(cls, config_str):
 
     def write(self, path, overwrite=False):
```
bkhelifi, line 257: Can you profit of this PR to detail the arguments of the function?

gammapy/modeling/fit.py
```diff
@@ -767,37 +765,13 @@ def covariance_result(self):
...
-        )
-        creation = CreatorMetaData()
-        return text + creation.to_yaml()
-
     def write(
         self,
         path,
         overwrite=False,
-        full_output=False,
+        full_output=True,
```
bkhelifi, line 772: The docstring should be updated ;)

gammapy/modeling/models/core.py
```diff
@@ -518,6 +518,8 @@ def write(
         checksum : bool
             When True adds a CHECKSUM entry to the file.
             Default is False.
+        extra_dict : dict
+            Additionnal informations to write with the models.
```
bkhelifi, line 522: Precise the default

gammapy/utils/scripts.py
```diff
@@ -67,8 +68,8 @@ def read_yaml(filename, logger=None, checksum=False):
     return data
 
 
-def write_yaml(dictionary, filename, logger=None, sort_keys=True, checksum=False):
-    """Write YAML file.
+def to_yaml(dictionary, sort_keys=False):
```
bkhelifi, line 71: The docstring should be updated because of the change of the parameters...
Should we add an API change decorator?

registerrier (commented):
Thanks @QRemy . Yaml serialization is indeed something that needs to be cleaned-up and clarified. 
I have a few questions here:
- I am not totally fond of having models.write() having a possible extra dictionary. This can make the format definition uncertain.
- `write_yaml(to_yaml())` is not very nice.

gammapy/modeling/fit.py
```diff
@@ -586,10 +582,12 @@ def _repr_html_(self):
     def to_dict(self):
         """Convert to dictionary."""
         return {
-            "backend": self.backend,
-            "method": self.method,
-            "success": self.success,
-            "message": self.message,
+            self.__class__.__name__: {
```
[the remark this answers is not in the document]
QRemy: I introduced that so the to_dict of CovarianceResult and FitResult already write their name rather than having this done within FitResult

gammapy/modeling/fit.py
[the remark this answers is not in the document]
QRemy: I added this to have also the min/max on the parameters even if they are default because the FitResult should give the full information about the fit conditions.

gammapy/modeling/fit.py
```diff
@@ -680,8 +678,8 @@ def __str__(self):
     def to_dict(self):
         """Convert to dictionary."""
         output = super().to_dict()
-        output["nfev"] = self.nfev
-        output["total_stat"] = self._total_stat
+        output[self.__class__.__name__]["nfev"] = self.nfev
```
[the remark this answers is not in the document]
QRemy: this has been introduced after 1.2 so it's fine no ?

gammapy/utils/scripts.py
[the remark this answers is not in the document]
QRemy: I introduced the global value just for the cleanup, not sure we want the other options to be configurable

gammapy/modeling/models/core.py
QRemy: removed from public API

QRemy:
> * `write_yaml(to_yaml())` is not very nice.

modified so it looks like :
```
yaml_str = to_yaml()
write_yaml(yaml_str)
```

 I changed the `write_yaml` from taking a dict to a yaml input because it was used almost nowhere as most of the classes relied on their `to_yaml` in their write methods.

QRemy:
> * I am not totally fond of having models.write() having a possible extra dictionary. This can make the format definition uncertain.

I introduced a private function `_write_models` that is used by both `Models` and `FitResult` so the `extra_dict` mechanism is not exposed.

gammapy/utils/scripts.py
registerrier: This is a new function here.

registerrier (commented):
See the few remaining inline comments. In particular, it misses a `deprecated_argument` decorator for `write_yaml`, as it is exposed in the doc

PR #5331, opened 2024-06-19
Temporary fix for #5290

### LINKED ISSUE #5290: FluxPointsEstimator fails on list of multiple stacked datasets because meta_table cannot be created
**Gammapy version**
1.2

**Bug description**
I am trying to run the `FluxPointsEstimator` on a list of (stacked) `MapDataset`s that each contain a different number of observations. This fails with the following error message:
```
TableMergeError                           Traceback (most recent call last)
File <timed exec>:4

File /nfs/scratch/mohrmann/software/miniconda3/envs/gammapy-1.2/lib/python3.10/site-packages/gammapy/estimators/points/sed.py:139, in FluxPointsEstimator.run(self, datasets)
    136 if not datasets.energy_axes_are_aligned:
    137     raise ValueError("All datasets must have aligned energy axes.")
--> 139 if "TELESCOP" in datasets.meta_table.colnames:
    140     telescopes = datasets.meta_table["TELESCOP"]
    141     if not len(np.unique(telescopes)) == 1:

File /nfs/scratch/mohrmann/software/miniconda3/envs/gammapy-1.2/lib/python3.10/site-packages/gammapy/datasets/core.py:561, in Datasets.meta_table(self)
    559     meta_table = Table()
    560 else:
--> 561     meta_table = vstack(tables).copy()
    563 meta_table.add_column([d.tag for d in self], index=0, name="TYPE")
    564 meta_table.add_column(self.names, index=0, name="NAME")

File /nfs/scratch/mohrmann/software/miniconda3/envs/gammapy-1.2/lib/python3.10/site-packages/astropy/table/operations.py:684, in vstack(tables, join_type, metadata_conflicts)
    681     return tables[0]  # no point in stacking a single table
    682 col_name_map = OrderedDict()
--> 684 out = _vstack(tables, join_type, col_name_map, metadata_conflicts)
    686 # Merge table metadata
    687 _merge_table_meta(out, tables, metadata_conflicts=metadata_conflicts)

File /nfs/scratch/mohrmann/software/miniconda3/envs/gammapy-1.2/lib/python3.10/site-packages/astropy/table/operations.py:1479, in _vstack(arrays, join_type, col_name_map, metadata_conflicts)
   1475     raise NotImplementedError(
   1476         f"vstack unavailable for mixin column type(s): {col_cls.__name__}"
   1477     )
   1478 try:
-> 1479     col = col_cls.info.new_like(cols, n_rows, metadata_conflicts, out_name)
   1480 except metadata.MergeConflictError as err:
   1481     # Beautify the error message when we are trying to merge columns with incompatible
   1482     # types by including the name of the columns that originated the error.
   1483     raise TableMergeError(
   1484         "The '{}' columns have incompatible types: {}".format(
   1485             out_name, err._incompat_types
   1486         )
   1487     ) from err

File /nfs/scratch/mohrmann/software/miniconda3/envs/gammapy-1.2/lib/python3.10/site-packages/astropy/table/column.py:488, in ColumnInfo.new_like(self, cols, length, metadata_conflicts, name)
    463 def new_like(self, cols, length, metadata_conflicts="warn", name=None):
    464     """
    465     Return a new Column instance which is consistent with the
    466     input ``cols`` and has ``length`` rows.
   (...)
    486 
    487     """
--> 488     attrs = self.merge_cols_attributes(
    489         cols, metadata_conflicts, name, ("meta", "unit", "format", "description")
    490     )
    492     return self._parent_cls(length=length, **attrs)

File /nfs/scratch/mohrmann/software/miniconda3/envs/gammapy-1.2/lib/python3.10/site-packages/astropy/utils/data_info.py:757, in BaseColumnInfo.merge_cols_attributes(cols, metadata_conflicts, name, attrs)
    755 uniq_shapes = {col.shape[1:] for col in cols}
    756 if len(uniq_shapes) != 1:
--> 757     raise TableMergeError("columns have different shapes")
    758 out["shape"] = uniq_shapes.pop()
    760 # "Merged" output name is the supplied name

TableMergeError: columns have different shapes
```

Evidently, the creation of a `meta_table` on the combined `Datasets` object fails. I checked that each of the datasets in the list has a `meta_table` – this is the case, and all tables have the same columns. The only difference between the tables that I can see is the different number of rows, e.g. for the first two datasets:
```
tables = [ds.meta_table for ds in datasets]
>>> print(tables[0]['TELESCOP'].shape)
(1, 68)
>>> print(tables[1]['TELESCOP'].shape)
(1, 84)
```
This should, in my view, not prevent the stacking of the tables, though.

Astro-Kirsty:
This PR is to resolve #5290.

For now we can loop over each dataset and be sure that the 'TELESCOP' is unique for that dataset. 
This is a short term fix.

My solution is quite clunky. I am open to suggestion

### DIFF
--- a/gammapy/estimators/points/sed.py
+++ b/gammapy/estimators/points/sed.py
@@ -137,13 +137,10 @@ class FluxPointsEstimator(FluxEstimator, parallel.ParallelMixin):
             raise ValueError("All datasets must have aligned energy axes.")
 
         telescopes = []
-        for dataset in datasets:
-            if (
-                dataset.meta_table is not None
-                and "TELESCOP" in dataset.meta_table.colnames
-            ):
-                telescopes.extend(list(dataset.meta_table["TELESCOP"].flatten()))
-        if not len(np.unique(telescopes)) == 1:
+        for d in datasets:
+            if d.meta_table is not None and "TELESCOP" in d.meta_table.colnames:
+                telescopes.extend(list(d.meta_table["TELESCOP"].flatten()))
+        if len(np.unique(telescopes)) > 1:
             raise ValueError(
                 "All datasets must use the same value of the"
                 " 'TELESCOP' meta keyword."
--- a/gammapy/estimators/points/sed.py
+++ b/gammapy/estimators/points/sed.py
@@ -136,29 +136,18 @@ class FluxPointsEstimator(FluxEstimator, parallel.ParallelMixin):
         if not datasets.energy_axes_are_aligned:
             raise ValueError("All datasets must have aligned energy axes.")
 
-        if datasets[0].meta_table is not None and any(
-            dataset.meta_table["TELESCOP"].shape
-            != datasets[0].meta_table["TELESCOP"].shape
-            for dataset in datasets
-        ):
-            for dataset in datasets:
-                if "TELESCOP" in dataset.meta_table.colnames:
-                    telescopes = dataset.meta_table["TELESCOP"]
-                    if not np.all(telescopes == telescopes[0]):
-                        raise ValueError(
-                            "All datasets must use the same value of the"
-                            " 'TELESCOP' meta keyword."
-                        )
-        else:
-            if "TELESCOP" in datasets.meta_table.colnames:
-                telescopes = datasets.meta_table["TELESCOP"]
-                if not len(np.unique(telescopes)) == 1:
-                    raise ValueError(
-                        "All datasets must use the same value of the"
-                        " 'TELESCOP' meta keyword."
-                    )
-
-        rows = []
+        telescopes = []
+        for dataset in datasets:
+            if (
+                dataset.meta_table is not None
+                and "TELESCOP" in dataset.meta_table.colnames
+            ):
+                telescopes.extend(list(dataset.meta_table["TELESCOP"].flatten()))
+        if not len(np.unique(telescopes)) == 1:
+            raise ValueError(
+                "All datasets must use the same value of the"
+                " 'TELESCOP' meta keyword."
+            )
 
         meta = {
             "n_sigma": self.n_sigma,
--- a/gammapy/estimators/points/sed.py
+++ b/gammapy/estimators/points/sed.py
@@ -136,13 +136,14 @@ class FluxPointsEstimator(FluxEstimator, parallel.ParallelMixin):
         if not datasets.energy_axes_are_aligned:
             raise ValueError("All datasets must have aligned energy axes.")
 
-        if "TELESCOP" in datasets.meta_table.colnames:
-            telescopes = datasets.meta_table["TELESCOP"]
-            if not len(np.unique(telescopes)) == 1:
-                raise ValueError(
-                    "All datasets must use the same value of the"
-                    " 'TELESCOP' meta keyword."
-                )
+        for dataset in datasets:
+            if "TELESCOP" in dataset.meta_table.colnames:
+                telescopes = dataset.meta_table["TELESCOP"]
+                if not len(np.unique(telescopes)) == 1:
+                    raise ValueError(
+                        "All datasets must use the same value of the"
+                        " 'TELESCOP' meta keyword."
+                    )
 
         rows = []
 
--- a/gammapy/estimators/points/sed.py
+++ b/gammapy/estimators/points/sed.py
@@ -136,9 +136,26 @@ class FluxPointsEstimator(FluxEstimator, parallel.ParallelMixin):
         if not datasets.energy_axes_are_aligned:
             raise ValueError("All datasets must have aligned energy axes.")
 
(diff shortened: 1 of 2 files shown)


Astro-Kirsty:
For the test we would need to create two sets of Datasets that have different length observations, these are not available in gammapy-data

AtreyeeS:
Do we really need this hard check on the `TELESCOP` ? I think we introduced it to avoid fitting different instruments together, but this seems to create more confusion than solutions. I remember there was some discussions on the hess slack as well. Can we remove this check now and introduce this cleanly with the meta handling? @registerrier @bkhelifi @adonath

gammapy/estimators/points/sed.py
```diff
@@ -136,13 +136,31 @@ def run(self, datasets):
...
+            and any(
+                dataset.meta_table["TELESCOP"].shape
+                != datasets[0].meta_table["TELESCOP"].shape
+                for dataset in datasets
+            )
+        ):
+            for dataset in datasets:
+                if "TELESCOP" in dataset.meta_table.colnames:
+                    telescopes = dataset.meta_table["TELESCOP"]
+                    if not len(np.unique(telescopes)) == 1:
```
VascoSch92, line 151: I will extract this part before the first if. In this way you don't have duplicate code

Astro-Kirsty: The code is not duplicated, note that we extract from `dataset.meta_table` here but below it is from `datasets.meta_table`

VascoSch92: yes but this part is duplicate:

```python
telescopes = datasets.meta_table["TELESCOP"]
                if not len(np.unique(telescopes)) == 1:
                    raise ValueError(
                        "All datasets must use the same value of the"
                        " 'TELESCOP' meta keyword."
                    )
```

gammapy/estimators/points/sed.py
```diff
@@ -136,13 +136,31 @@ def run(self, datasets):
...
             raise ValueError("All datasets must have aligned energy axes.")
 
-        if "TELESCOP" in datasets.meta_table.colnames:
-            telescopes = datasets.meta_table["TELESCOP"]
-            if not len(np.unique(telescopes)) == 1:
-                raise ValueError(
-                    "All datasets must use the same value of the"
-                    " 'TELESCOP' meta keyword."
-                )
+        if (
```
registerrier, line 139: It does not seem necessary to test for all these conditions here. 
You only need to loop over all datasets in the `Datasets`, build the set of "TELESCOP" values and make sure it contains one element only.

gammapy/estimators/points/sed.py
```diff
@@ -136,13 +136,31 @@ def run(self, datasets):
...
+            len(datasets) > 1
+            and datasets[0].meta_table is not None
+            and any(
+                dataset.meta_table["TELESCOP"].shape
+                != datasets[0].meta_table["TELESCOP"].shape
+                for dataset in datasets
+            )
+        ):
+            for dataset in datasets:
+                if "TELESCOP" in dataset.meta_table.colnames:
```
registerrier, line 149: This is testing that every dataset has only one single TELESCOP value, not that these values are always the same.

registerrier (commented):
Thanks @Astro-Kirsty . The code can be simplified, see inline comments.

Maybe you could remove the exception and replace with a warning mentioning that computing joint flux points from different instruments is potentially yielding incorrect results.

gammapy/estimators/points/tests/test_sed.py
```diff
@@ -786,3 +791,83 @@ def test_flux_points_estimator_norm_spectral_model(fermi_datasets):
     flux_points_dataset = FluxPointsDataset(data=flux_points, models=model)
     flux_pred = flux_points_dataset.flux_pred()
     assert_allclose(flux_pred, flux_pred_ref, rtol=2e-4)
+
+
+def test_fpe_diff_lengths():
```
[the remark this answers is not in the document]
registerrier: But you can create it and add it on the datasets right?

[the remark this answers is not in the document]
registerrier: Also this needs a `requires_data` decorator

gammapy/estimators/points/sed.py
Astro-Kirsty: I found issues without the very specific statements. In particular, `dataset.meta_table` was sometimes None which means `dataset.meta_table.colnames` raises a TypeError

gammapy/estimators/points/sed.py
QRemy: modified and simplified

QRemy (approved):
Thanks @Astro-Kirsty, I made some modification to simplify the code taking into account the remaining comment and added few more tests, merging now.

PR #5334, opened 2024-06-19
Propagate metadata to cutout

### LINKED ISSUE #5225: Propagation of metadata in MapDataset.cutout
Use of the MetaData classes instead of the dictionaries.
This impacts the cutout functions into gammapy.maps

Astro-Kirsty:
This PR resolves  #5225

Further discussion will be needed with exactly what we want from the meta
We want to make sure their is a new 'creatormetadata' for when the cutout is created. But what else do we want to keep from the parent dataset?

I also get a little confused between metadata and provenance. Are the options for the cutout actually provenance?

### DIFF
--- a/gammapy/datasets/map.py
+++ b/gammapy/datasets/map.py
@@ -20,6 +20,7 @@ from gammapy.stats import (
     wstat,
 )
 from gammapy.utils.fits import HDULocation, LazyFitsData
+from gammapy.utils.metadata import CreatorMetaData
 from gammapy.utils.random import get_random_state
 from gammapy.utils.scripts import make_name, make_path
 from gammapy.utils.table import hstack_columns
@@ -1841,29 +1842,35 @@ class MapDataset(Dataset):
             Cutout map dataset.
         """
         name = make_name(name)
-        kwargs = {"gti": self.gti, "name": name, "meta_table": self.meta_table}
-        cutout_kwargs = {"position": position, "width": width, "mode": mode}
+        kwargs = {
+            "gti": self.gti,
+            "name": name,
+            "meta_table": self.meta_table,
+            "meta": self.meta,
+        }
+        kwargs["meta"].creation = CreatorMetaData()
+        kwargs["meta"].optional = {"position": position, "width": width, "mode": mode}
 
         if self.counts is not None:
-            kwargs["counts"] = self.counts.cutout(**cutout_kwargs)
+            kwargs["counts"] = self.counts.cutout(position, width, mode)
 
         if self.exposure is not None:
-            kwargs["exposure"] = self.exposure.cutout(**cutout_kwargs)
+            kwargs["exposure"] = self.exposure.cutout(position, width, mode)
 
         if self.background is not None and self.stat_type == "cash":
-            kwargs["background"] = self.background.cutout(**cutout_kwargs)
+            kwargs["background"] = self.background.cutout(position, width, mode)
 
         if self.edisp is not None:
-            kwargs["edisp"] = self.edisp.cutout(**cutout_kwargs)
+            kwargs["edisp"] = self.edisp.cutout(position, width, mode)
 
         if self.psf is not None:
-            kwargs["psf"] = self.psf.cutout(**cutout_kwargs)
+            kwargs["psf"] = self.psf.cutout(position, width, mode)
 
         if self.mask_safe is not None:
-            kwargs["mask_safe"] = self.mask_safe.cutout(**cutout_kwargs)
+            kwargs["mask_safe"] = self.mask_safe.cutout(position, width, mode)
 
         if self.mask_fit is not None:
-            kwargs["mask_fit"] = self.mask_fit.cutout(**cutout_kwargs)
+            kwargs["mask_fit"] = self.mask_fit.cutout(position, width, mode)
 
         return self.__class__(**kwargs)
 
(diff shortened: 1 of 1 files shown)


gammapy/datasets/map.py
```diff
@@ -1841,29 +1842,35 @@ def cutout(self, position, width, mode="trim", name=None):
...
         name = make_name(name)
-        kwargs = {"gti": self.gti, "name": name, "meta_table": self.meta_table}
-        cutout_kwargs = {"position": position, "width": width, "mode": mode}
+        kwargs = {
+            "gti": self.gti,
+            "name": name,
+            "meta_table": self.meta_table,
+            "meta": self.meta,
+        }
+        kwargs["meta"].creation = CreatorMetaData()
```
registerrier, line 1851: In general, I wonder whether, for data products created by gammapy, the creation metadata could be kept to None until serialization.

registerrier (commented):
Thanks @Astro-Kirsty ! Some questions for discussion:
- Shouldn't meta propagation be optional on cutout? There are cases where it makes sense to propagate and others where it might not. e.g. extracting a sub-dataset from a larger one or performing data reduction. 
- I think we should keep creation metadata on DL4+ objects mostly for serialization. Hence it could be kept to None and add it when calling `write`.

gammapy/datasets/map.py
bkhelifi: The metadata propagation should be made easely possible at any step of the analysis. If at each maker, classes, etc, the user should say `meta_propagation=True`, this will be a very inconvenient API. 
However, the feature to switch on and off the propagation is interesting.
I propose to set up a global Gammapy variable, such as the progress bas, to permit this feature... Then, we can expose this variable into the HowTo

bkhelifi: Otherwise, indeed, the metadata creation should be set-up at the serialisation step

gammapy/datasets/map.py
```diff
@@ -1841,29 +1842,35 @@ def cutout(self, position, width, mode="trim", name=None):
             Cutout map dataset.
         """
         name = make_name(name)
-        kwargs = {"gti": self.gti, "name": name, "meta_table": self.meta_table}
-        cutout_kwargs = {"position": position, "width": width, "mode": mode}
```
AtreyeeS, line 1845: You can keep `cutout_kwargs` and directly set `kwargs["meta"].optional = cutout_kwargs` to avoid the later changes

AtreyeeS (commented):
Thanks @Astro-Kirsty for starting this. 

I suppose the propagation of meta is not limited only to `cutout` but to all Dataset operations like slicing, resampling, copying, `to_spectrum_dataset` etc. Similar question would arise at the DL5 levels as well

I was thinking of a general scheme of defining functions
1. `propagate_meta(meta1)` which  will copy the previous meta info, and reset the creation. The individual functions can then add optional quantities if required. 
2. `stack_meta(meta1, meta2)` which will linearly stack all meta keywords if `meta1` and `meta2` belong to the the same class. During serialisation, we can decide whether to keep only the unique ones, or all.

I agree with @registerrier that maybe the `creation` is useful only for serialisation

bkhelifi:
I have started a discussion about the metadata handling: https://github.com/gammapy/gammapy/discussions/5343

gammapy/datasets/map.py
Astro-Kirsty: Yes but is there some reason to have the `cutout_kwargs`? They cannot be accessed from outside

adonath: Using `cutout_kwargs` is syntactically shorter, when the same keywords are passed to multiple methods. Passing arguments by name is usually preferred to passing by position, because its is more explicit. However using `.cutout(position=position, width=width, mode=mode)` all over again is cumbersome. Hence the `kwargs` provides the cleanest solution.

QRemy (commented):
Thanks @Astro-Kirsty, why do we need  both `meta_table` and `meta` in kwargs ?
If `meta_table` is deprecated in favor of `meta` we should not use it here.
I agree with @adonath about keeping cutout_kwargs so I pushed a commit to restore that.

gammapy/datasets/map.py
```diff
@@ -1841,29 +1842,35 @@ def cutout(self, position, width, mode="trim", name=None):
...
+        kwargs = {
+            "gti": self.gti,
+            "name": name,
+            "meta_table": self.meta_table,
+            "meta": self.meta,
```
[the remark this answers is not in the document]
Astro-Kirsty: I think we should deprecate meta_table, but we have not done that yet in the datasets metadata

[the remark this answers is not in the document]
adonath: If I remember correctly we kept the `meta_table` for now, because we have not settled on a meta data structure for for multiple datasets yet (where the table comes in handy...). See https://github.com/gammapy/gammapy/pull/4853#issuecomment-1906605913 for some more thoughts...

gammapy/datasets/map.py
Astro-Kirsty: The point of adding the 'creation' here is that then it will signify the creation of the cutout itself. 

So instead I should just move the creator to the write functionality and therefore on the MapDataset in general?

AtreyeeS:
Following the ideas in https://github.com/gammapy/gammapy/discussions/5343#discussioncomment-9840658 (and subsequent comments), I don't understand how this cutout information on the meta relates to the planned CTAO meta data model. This might fall in some `ScienceAnalysisMetaData`, but I am not sure if the cutout information is really relevant at the higher level. For example, during data reduction, datasets are always cutout and then stacked onto an empty one, and having this entire cutout info would be a clutter.

Maybe the entire cutout `cutout-kwargs` should be stored as provenance, and a correct `reference-metadata` (it can be the `dataset.name` or some additional keyword, TBD) can relate the dataset to the provenance?

PR #5342, opened 2024-06-21
Fix evaluation if no edisp is set

### LINKED ISSUE #5274: FluxPointsEstimator fails if no edisp is set

QRemy:
Resolves #5274.
Better solution than #5308 :)

### DIFF
--- a/gammapy/estimators/points/tests/test_sed.py
+++ b/gammapy/estimators/points/tests/test_sed.py
@@ -449,8 +449,8 @@ def test_run_no_edip(fpe_pwl, tmpdir):
     assert_allclose(actual, [1.081434, 0.91077, 0.922176], rtol=1e-3)
 
     datasets[0].models["source"].apply_irf["edisp"] = True
-    with pytest.raises(ValueError):
-        fp = fpe.run(datasets)
+    actual = table["norm"].data
+    assert_allclose(actual, [1.081434, 0.91077, 0.922176], rtol=1e-3)
 
 
 @requires_dependency("iminuit")
--- a/gammapy/estimators/points/tests/test_sed.py
+++ b/gammapy/estimators/points/tests/test_sed.py
@@ -430,6 +430,29 @@ def test_run_map_pwl_reoptimize(fpe_map_pwl_reoptimize):
     assert_allclose(actual, [9.788123, 0.486066, 17.603708], rtol=1e-2)
 
 
+def test_run_no_edip(fpe_pwl, tmpdir):
+    datasets, fpe = fpe_pwl
+
+    datasets = datasets.copy()
+
+    datasets[0].models["source"].apply_irf["edisp"] = False
+    fp = fpe.run(datasets)
+    table = fp.to_table()
+    actual = table["norm"].data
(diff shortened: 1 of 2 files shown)


AtreyeeS (approved):
Thanks @QRemy ! This is a very clean solution.
Maybe it would still be useful to raise a warning in `MapDataset` if `edisp is None` ?

gammapy/estimators/points/tests/test_sed.py
```diff
@@ -430,6 +430,29 @@ def test_run_map_pwl_reoptimize(fpe_map_pwl_reoptimize):
...
+    assert_allclose(actual, [1.081434, 0.91077, 0.922176], rtol=1e-3)
+
+    datasets[0].edisp = None
+
+    fp = fpe.run(datasets)
+    table = fp.to_table()
+    actual = table["norm"].data
+    assert_allclose(actual, [1.081434, 0.91077, 0.922176], rtol=1e-3)
+
+    datasets[0].models["source"].apply_irf["edisp"] = True
```
fabiopintore, line 451: Shouldn't you add:
```    fp = fpe.run(datasets)
         table = fp.to_table()```
here?

**Review by fabiopintore — COMMENTED** [review]
Thanks @QRemy . It looks definitely better than the other solution. It looks good to me. I have only a small comment for the test.

File: gammapy/estimators/points/tests/test_sed.py (hunk 1 above)
Comment (QRemy) [inline gammapy/estimators/points/tests/test_sed.py:451]:
```suggestion
    datasets[0].models["source"].apply_irf["edisp"] = True
    fp = fpe.run(datasets)
    table = fp.to_table()
```

QRemy: yes added

Issue #5291, opened 2024-06-06
Fit.stat_contour and stat_surface yield incorrect parameter names


registerrier:
**Gammapy version**
1.2, dev and probably older versions too

**Bug description**
When using complex models with models shared among several `SkyModel` in `Models`, the schemes of `Models.unique_parameter_names` and `Parameters.unique_parameters` become inconsistent. This becomes problematic when running `Fit.stat_contours` since the parameter names that are extracted do not correspond to the ones actually tested.
 
**Expected behavior**
The unique parameters names used should be the correct ones.

**To Reproduce**
This needs a complex modeling of dataset to reproduce. I don't have a simple example for now.

**Other information**

AtreyeeS:
Does it work correctly if `models.parameters[xx]` is not passed to `fit.stat_contour` but rather `sky_model.parameters[xx]`?

AtreyeeS:
The decision here is to modify the fitting tutorial (https://github.com/gammapy/gammapy/blob/main/examples/tutorials/api/fitting.py) to always use `model.parameters["par_name"]` instead of going through `datasets.parameters` or `models.parameters`

AtreyeeS:
It will also be nice to add the examples in the docstrings for `fit.stat_profile` , `stat_surface`, `stat_contour` showing the same

Astro-Kirsty:
> The decision here is to modify the fitting tutorial (https://github.com/gammapy/gammapy/blob/main/examples/tutorials/api/fitting.py) to always use `model.parameters["par_name"]` instead of going through `datasets.parameters` or `models.parameters`

We need to wait for the unique parameters issue to be solved potentially #5320

Astro-Kirsty:
@QRemy @AtreyeeS, is there anything more that needs to be done here?

QRemy:
Bug is fixed and documentation is adapted so I think we are done.

PR #5347, opened 2024-06-24
Modify API fitting tutorial

Astro-Kirsty:
This PR is to address https://github.com/gammapy/gammapy/issues/5291#issuecomment-2163313687

We want to avoid accessing `datasets.parameters[name]` but rather take directly from the model

### DIFF
--- a/examples/tutorials/api/fitting.py
+++ b/examples/tutorials/api/fitting.py
@@ -230,12 +230,14 @@ total_stat = result_minuit.total_stat
 
 fig, axes = plt.subplots(nrows=1, ncols=3, figsize=(14, 4))
 
-for ax, par in zip(axes, datasets.parameters.free_parameters):
+for ax, par in zip(axes, crab_model.parameters.free_parameters):
     par.scan_n_values = 17
-    idx = datasets.parameters.index(par)
-    name = datasets.models.parameters_unique_names[idx]
+    idx = crab_model.parameters.index(par)
+    name = crab_model.parameters_unique_names[idx]
     profile = fit.stat_profile(datasets=datasets, parameter=par)
-    ax.plot(profile[f"{name}_scan"], profile["stat_scan"] - total_stat)
+    ax.plot(
+        profile[f"{crab_model.name}.{name}_scan"], profile["stat_scan"] - total_stat
+    )
     ax.set_xlabel(f"{par.name} [{par.unit}]")
     ax.set_ylabel("Delta TS")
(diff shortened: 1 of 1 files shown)


Astro-Kirsty:
I am also going to adjust the plotting of the `stat_contour` because these lines can be confusing for the user

Astro-Kirsty:
https://astro-kirsty.github.io/gammapy-docs-preview/tutorials/api/fitting.html#sphx-glr-tutorials-api-fitting-py

bkhelifi:
I have a side question: the discusssion outcomes was not to use the FitResult object within the tutorials?

Astro-Kirsty:
I thought the end decision was to use `datasets.models` or `fit_result.models`

AtreyeeS:
> I have a side question: the discusssion outcomes was not to use the FitResult object within the tutorials?

@bkhelifi : Thanks to @QRemy recent efforts, `datasets.models` is now complete.
We can, of course, expose better `fit_results` as a matter of habit, but it is not urgent

AtreyeeS (commented):
Thank you @Astro-Kirsty !

While having the stat_contours in a loop looks impressive, the feedback that I got is that because it loops over the parameters, users find it rather difficult to understand how to compute it for some specific parameter and not all combinations of all for parameters.

I would say we show something like

```
par1  = crab_model.parameters["alpha"]
par2 = ...
contour = fit.stat_contour(
                 datasets=datasets,
                 x=par1,
                 y=par2,
                 numpoints=11,
                 sigma=1,
             )
```

then explain what `contour` contains and show how to plot it using `plot_contour_line`.

Similarly for the `stat_surface` and the 1d `stat_profile`

examples/tutorials/api/fitting.py
```diff
@@ -357,69 +358,52 @@ def make_contours(fit, datasets, result, npoints, sigmas):
...
-        "cx": (lambda ct: ct["contour_alpha_beta"]["alpha"]),
-        "cy": (lambda ct: ct["contour_alpha_beta"]["beta"]),
-    },
-]
-
 
-######################################################################
-# Finally we produce the confidence contours figures.
+#####################################################################
+# Produce the confidence contours figures.
```
QRemy, line 388: This could also go in a function that take `contours`, `params`, `sigmas` as input.
(We could even have that somewhere in gammapy together with the make_contours function but that would be for another PR)

QRemy (approved):
Thanks @Astro-Kirsty looks good (especially the simplification  of the ugly plotting code that I may have written).

PR #5350, opened 2024-06-25
Adapt `RadMax2D.plot_rad_max_vs_energy`

### LINKED ISSUE #5333: y-axis value not visible in RadMax2D.plot_rad_max_vs_energ
In `RadMax2D.plot_rad_max_vs_energy` , the values on the y-axis are not visible. See, eg:  https://docs.gammapy.org/1.2/tutorials/analysis-1d/spectral_analysis_rad_max.html#load-data

Astro-Kirsty:
To resolve #5333

When we utilise `quantity_support` it overrides any formating, so to continue with this method to ensure the units are handled correctly we can simply set a formatter on the axis. 
This is also the method in `plot_containment_radius_vs_energy`.

### DIFF
--- a/gammapy/irf/rad_max.py
+++ b/gammapy/irf/rad_max.py
@@ -2,6 +2,7 @@
 import astropy.units as u
 from astropy.visualization import quantity_support
 import matplotlib.pyplot as plt
+from matplotlib.ticker import FormatStrFormatter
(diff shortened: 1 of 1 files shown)


registerrier (approved):

registerrier:
CI fail is due to healpy 1.17 not being compatible with the latest scipy 1.14

PR #5353, opened 2024-06-26
Add support for joint TSmap estimation

QRemy:
Allows the TSmapEstimator to perform computations on datasets.

### DIFF
--- a/gammapy/estimators/map/ts.py
+++ b/gammapy/estimators/map/ts.py
@@ -7,6 +7,7 @@ import scipy.optimize
 from astropy.coordinates import Angle
 from astropy.utils import lazyproperty
 import gammapy.utils.parallel as parallel
+from gammapy.datasets import Datasets
 from gammapy.datasets.map import MapEvaluator
 from gammapy.datasets.utils import get_nearest_valid_exposure_position
 from gammapy.maps import Map, Maps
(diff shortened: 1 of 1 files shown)


adonath:
Thanks a lot @QRemy, this looks very good at first glance! I am wondering which data to use best for testing. I think we do not have any pre-reduced data (datasets) for event types yet. Any suggestion @registerrier or @AtreyeeS?

QRemy:
> Thanks a lot @QRemy, this looks very good at first glance! I am wondering which data to use best for testing. I think we do not have any pre-reduced data (datasets) for event types yet. Any suggestion @registerrier or @AtreyeeS?

For now testing this doing a joint map of the the "MSH 15-52" observations from the HESS public data release,  I'm preparing a notebook to give a preview of all the new features related to the TSmapEstimator.
(I also check it everything works for the HAWC datasets that include event types).

gammapy/estimators/map/ts.py
```diff
@@ -479,36 +483,56 @@ def run(self, dataset):
...
+        for dataset in datasets:
+            pad_width_dataset = self.estimate_pad_width(dataset=dataset)
+            pad_width = tuple(np.maximum(pad_width, pad_width_dataset))
+
+        datasets_padded = Datasets()
+        for dataset in datasets:
+            dataset = dataset.pad(pad_width, name=dataset.name)
+            dataset = dataset.downsample(self.downsampling_factor, name=dataset.name)
+            datasets_padded.append(dataset)
+        datasets = datasets_padded
```
VascoSch92, line 510: you could do this in one for loop instead of two

adonath: I think he uses the first loop to determine the max. width for the padding across all datasets first. Then the same pad width is used later for all datasets. I think the two loop approach is correct.

VascoSch92: ah ok thank you now I see

gammapy/estimators/map/ts.py
```diff
@@ -408,26 +409,29 @@ def estimate_fit_input_maps(self, dataset):
...
 
-    def estimate_flux_map(self, dataset):
+    def estimate_flux_map(self, datasets):
         """Estimate flux and test statistic maps for single dataset.
 
         Parameters
         ----------
-        dataset : `~gammapy.datasets.MapDataset`
+        dataset : `~gammapy.datasets.Datasets`
             Map dataset.
```
AtreyeeS, line 418: list of `MapDataset` objects

AtreyeeS (approved):
Thank you @QRemy ! This looks good. 
Maybe you can add DL4 HAWC datasets to gammapy data and then add a test for this

QRemy:
> Maybe you can add DL4 HAWC datasets to gammapy data and then add a test for this

Done !

adonath (approved):
Thanks @QRemy, I have no further comments!

PR #5356, opened 2024-06-29
Brought gammapy.visualization.plot_distribution in line with its documentation

GerritRo:
Running 
```
fig, ax = plt.subplots(nrows=1, ncols=1)
gammapy.visualization.plot_distribution(ts_map, ax=ax)
```
results in an error, since gammapy.visualization.plot_distribution expects an array of axes in its current form, which does not fit the description, which explicitly allows singular axes objects and lists of axes objects:
```
Parameters
----------
wcs_map : `~gammapy.maps.WcsNDMap`
    A map that contains data to be plotted.
ax : `~matplotlib.axes.Axes` or list of `~matplotlib.axes.Axes`
    Axis object to plot on. If a list of Axis is provided it has to be the same length as the length of _map.data.
```
This pull request implements a fix, but maybe it is not in line with what the original author of the function intended (and still makes implicit assumptions about the input. As implemented, it expects an object or an array like object, on which np.array([obj]) can be called.)

### DIFF
--- a/gammapy/visualization/utils.py
+++ b/gammapy/visualization/utils.py
@@ -341,8 +341,8 @@ def plot_distribution(
         )
         cells_in_grid = rows * cols
     else:
-        axes = ax
-        cells_in_grid = len(ax.flatten())
+        axes = np.array([ax])
+        cells_in_grid = axes.size
 
     if not isinstance(axes, np.ndarray):
         axes = np.array([axes])
--- a/gammapy/visualization/tests/test_utils.py
+++ b/gammapy/visualization/tests/test_utils.py
@@ -138,3 +138,6 @@ def test_plot_distribution():
         axes, res = plot_distribution(
             wcs_map=map_, func="norm", kwargs_hist={"bins": 40}
         )
+
+        fig, ax = plt.subplots(nrows=1, ncols=1)
+        plot_distribution(map_empty, ax=ax)


AtreyeeS (commented):
Thanks @GerritRo 
Maybe you can add modify the tests to make sure that this works.

gammapy/visualization/utils.py
```diff
@@ -341,8 +341,8 @@ def plot_distribution(
         )
         cells_in_grid = rows * cols
     else:
-        axes = ax
-        cells_in_grid = len(ax.flatten())
+        axes = np.array([ax])
+        cells_in_grid = axes.size
```
MRegeard, line 345:
```suggestion
```

gammapy/visualization/utils.py
```diff
@@ -341,8 +341,8 @@ def plot_distribution(
...
 
     if not isinstance(axes, np.ndarray):
         axes = np.array([axes])
```
MRegeard, line 348:
```suggestion
        axes = np.array([axes])
    cells_in_grid = axes.size
```

MRegeard (changes requested):
Thanks @GerritRo for spotting this (my 😄 ) mistake !
Since we are checking if axes is a numpy array, I would suggest to just initialise `cells_in_grid` after this type check. It should fix the bug. 
See inline comments.

gammapy/visualization/utils.py
GerritRo: In this case, cells in grid is initialized over a previous instance if ax=None (see line 342). Should we just change this as well?

MRegeard: I see what you mean, let me redo my comments then!

MRegeard:
@GerritRo, What you can do to make the code cleaner is to assign `axes = ax` before the first if statement (line 328). Then you do the if statement on axes. You also remove the `cells_in_grid` assignment in the if statment. This way you can completely remove the else statement and assign `cells_in_grid = axes.size` after (outside) of the instance check.
This should fix the issue and make the code cleaner at the same time. 
What you have to do anyway is to add a test (or modify the existing one) for this behaviour, as mentioned by @AtreyeeS.

QRemy (approved):
Thanks @GerritRo, added test and merging now.

Issue #5344, opened 2024-06-21
Task list for Sensitivity estimations


AtreyeeS:
After the discussions in #4718 and the prototype in by @registerrier in  #4712 , decided to implement our gammapy API with the new sensitivity estimation.
This can be done using (at least) 3 PRs. Please feel free to assign yourselves 

- [ ] Add a `Dataset.stat_sum_asimov` to get the stat for an Asimov dataset. Also add on `Datasets` for convenience  @AtreyeeS 
- [ ] Add optiional quantity `sensitivity` on `FluxPointsEstimator`, `TSMapEstimator`
- [ ] Also, adapt the `ParameterEstimator` to compute sensitivity
- [ ] Adapt the current `SensitivityEstimator` to use implementation from Cowan et al

adonath:
Thanks @AtreyeeS, the plan looks very reasonable! I will not have time to implement anything myself in the coming weeks, but I can review PRs.

registerrier:
After #5637 #5639 #5679 the first three points have been addressed. The remaining aspect is adapting the `SensitivityEstimator`

bkhelifi:
To be linked with the discussion: https://github.com/gammapy/gammapy/discussions/5417

AtreyeeS:
Can we put this in for 2.0? @registerrier @QRemy @bkhelifi ?

AtreyeeS:
I think we should not modify the `SensitivityEstimator` as it conforms to the default sensitivity definitions for CTA (eg: what pyirf does), but create a new `JointSensitivityEstimator` (probably inheriting from the `FluxPointsEstimator`) to compute sensivity with the Cowan approach

PR #5361, opened 2024-07-04
Add stat_sum_asimov on MapDataset and MapDatasetOnOff

AtreyeeS:
As discussed towards efforts for a sensitivity estimator #5344 

The tests are not very intelligible. Maybe it is better to test with dummy values ?

### DIFF
--- a/gammapy/datasets/core.py
+++ b/gammapy/datasets/core.py
@@ -250,11 +250,10 @@ class Datasets(collections.abc.MutableSequence):
         """
         from .flux_points import FluxPointsDataset
 
-        """Test statistic for the given model assuming an Asimov dataset"""
         stat_sum_asimov = 0
         for dataset in self:
             if isinstance(self, FluxPointsDataset):
-                raise ValueError("Not defined for a flux points dataset")
+                raise TypeError("Not defined for a flux points dataset")
             stat_sum_asimov += dataset.stat_sum_asimov(model_name=model_name)
         return stat_sum_asimov
 
--- a/gammapy/datasets/core.py
+++ b/gammapy/datasets/core.py
@@ -240,6 +240,24 @@ class Datasets(collections.abc.MutableSequence):
             stat_sum += dataset.stat_sum()
         return stat_sum
 
+    def stat_sum_asimov(self, model_name=None):
+        """Test statistic for the given model assuming Asimov datasets
+
+        Parameters
+        ----------
+        model_name : str
+            Name of the sky model for which to compute the statistic
+        """
+        from .flux_points import FluxPointsDataset
+
+        """Test statistic for the given model assuming an Asimov dataset"""
+        stat_sum_asimov = 0
+        for dataset in self:
+            if isinstance(self, FluxPointsDataset):
+                raise ValueError("Not defined for a flux points dataset")
+            stat_sum_asimov += dataset.stat_sum_asimov(model_name=model_name)
+        return stat_sum_asimov
+
     def select_time(self, time_min, time_max, atol="1e-6 s"):
         """Select datasets in a given time interval.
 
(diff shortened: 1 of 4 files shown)


gammapy/datasets/core.py
```diff
@@ -240,6 +240,24 @@ def stat_sum(self):
...
+        """Test statistic for the given model assuming Asimov datasets
+
+        Parameters
+        ----------
+        model_name : str
+            Name of the sky model for which to compute the statistic
+        """
+        from .flux_points import FluxPointsDataset
+
+        """Test statistic for the given model assuming an Asimov dataset"""
```
VascoSch92, line 253:
```suggestion
```

I think this comment is not necessary :-)

gammapy/datasets/core.py
```diff
@@ -240,6 +240,24 @@ def stat_sum(self):
...
+        stat_sum_asimov = 0
+        for dataset in self:
+            if isinstance(self, FluxPointsDataset):
+                raise ValueError("Not defined for a flux points dataset")
```
VascoSch92, line 257: is not a `TypeError` ?

gammapy/datasets/map.py
```diff
@@ -1344,6 +1344,29 @@ def stat_sum(self):
         else:
             return cash_sum_cython(counts.ravel(), npred.ravel()) + prior_stat_sum
 
+    def stat_sum_asimov(self, model_name=None):
```
VascoSch92, line 1347: what happens if `model_name=None` ?

gammapy/datasets/core.py
QRemy, line 257:
```suggestion
                raise TypeError("Not defined for a flux points dataset")
```

gammapy/datasets/map.py
StFroese: It gets passed to the `npred_signal`, which in general reduces the list of evaluators used here to the ones of the model given.
```python
        if model_names is not None:
            if isinstance(model_names, str):
                model_names = [model_names]
            evaluators = {name: self.evaluators[name] for name in model_names}
```
only problem I see is that `model_names=[model_name]` in case of `model_name=None` is actually not None and the evaluators dict from above will be emtpy
not sure if this is the wanted behaviour

gammapy/datasets/map.py
```diff
@@ -1344,6 +1344,29 @@ def stat_sum(self):
...
+
+        if self.mask is not None:
+            mask = self.mask.data
+            npred = npred[mask]
+            npred_h0 = npred_h0[mask]
+
+        L0 = cash_sum_cython(npred, npred)
+        L1 = cash_sum_cython(npred, npred_h0)
+
+        return np.sum(L1 - L0)
```
StFroese, line 1368: I'm a little bit confused about the notation here 😅 
Let me sum this up for my own understanding:
1. You have a Map dataset built from an observation
2. `npred` is the number of the total number of signal and background counts predicted from all models
3. `npred_source` are the counts from one specific model
4. `npred_h0` is the null hypothesis, alias signal=0 and only bkg

So in total is `L0` the best fit profiled log-likelihood, or cash stat, and `L1` as the null hypo for only bkg. 
I have the following concerns:
- `npred_h0` being plugged into `L1` is a bit confusing, I would swap the names maybe
- this method is name `stat_sum_asimov` in reference to the normal `stat_sum` of the dataset although this is not a stat_sum but the test statistic itself for the null hypo, named $q_0$ in the Cowen paper
- This is not flexible enough and only computes the $q_0$ it is not possible to compute upper limits with this for example since one would need $q_\mu$ with $\mu$ being a strength parameter  that scales the signal model
- Priors are missing in this test statistic, they need be evaluated too and do not cancel since, see eq 26 cowen paper, the nuisance params are profiled for the null hypo and taken as given by the alternative hypo

StFroese: The last point actually includes reoptimizing the models for s=0

bkhelifi:
It would be nice to describe this statistics into the Fit.Statistics documentation page

AtreyeeS:
Superseded by #5637

PR #5370, opened 2024-07-15
Add offset mask to make_effective_livetime_map

QRemy:
Add offset mask to make_effective_livetime_map


LauraOlivera (approved):
thanks a lot Quentin for turning my little complaint into a PR in less than 15 min :joy: 
it looks fine to me

registerrier (approved):
Thanks @Qremy. Looks good to me.

I note that because of the `SafeMaskMaker` design we cannot really reuse the offset max function. We should keep this in mind when we refactor it.

PR #5378, opened 2024-07-18
Observations `__getitem__` method addition 

MRegeard:
This PR improves the `__getitem__` method of `Observations`. 
It add the possibilitY to access with array indexing. This is particularly useful to access part of the Observations using a mask. 
I also added the possibility to access part of the array using a list of string. Typically, this can be use to access part of the observations with obs_id.

### DIFF
--- a/gammapy/data/tests/test_observations.py
+++ b/gammapy/data/tests/test_observations.py
@@ -635,12 +635,23 @@ def test_observations_getitem(data_store):
     assert len(obs_2) == 2
     assert obs_2.ids == ["20136", "20151"]
 
-    obs_3 = obs_1[np.array(mask)]
+    ind = [0, 2]
+    obs_2 = obs_1[ind]
 
-    assert len(obs_3) == 2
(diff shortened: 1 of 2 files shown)


gammapy/data/tests/test_observations.py
```diff
@@ -628,3 +618,29 @@ def test_event_setter():
...
     obs.events = events
     assert obs.events is events
+
+
+@requires_data()
+def test_observations_getitem(data_store):
+    """Test mask indexing on Observations"""
+    obs_1 = data_store.get_observations([20136, 20137, 20151])
+
+    assert isinstance(obs_1[0], Observation)
```
VascoSch92, line 628: This cover the case with numpy array right?

MRegeard: Line 638-641 does it!

QRemy (approved):
Thanks @MRegeard, I simplified it a bit and added more tests.

PR #5380, opened 2024-07-19
Add max_niter as option in TSMapEstimator

QRemy:
Add max_niter as option in TSMapEstimator (passed to BrentqFluxEstimator) and increase max default to 100 instead of 20 (joint maps converge for niter~20-30).


bkhelifi (approved):
Thanks!

PR #5382, opened 2024-07-19
Fix and tweaks for the stat_scan in TSMapEstimator

QRemy:
Fix and tweaks for the stat_scan in TSMapEstimator.
Follow up of #5285.


bkhelifi (approved):
Thanks @QRemy ....
I guess that you found that after seeing an issue. Could add this use case into a test function?

PR #5385, opened 2024-07-21
Fix reco_exposure computation with mask safe in ExcessMapEstimator

QRemy:
Fix reco_exposure computation with mask safe in ExcessMapEstimator

### DIFF
--- a/gammapy/estimators/map/excess.py
+++ b/gammapy/estimators/map/excess.py
@@ -228,6 +228,17 @@ class ExcessMapEstimator(Estimator):
         resampled_dataset = dataset.resample_energy_axis(
             energy_axis=axis, name=dataset.name
         )
+
+        if dataset.exposure:
+            reco_exposure = estimate_exposure_reco_energy(
+                dataset, self.spectral_model, normalize=False
+            )
+            reco_exposure = reco_exposure.resample_axis(
+                axis=axis, weights=dataset.mask_safe
+            )
+        else:
+            reco_exposure = None
+
         if isinstance(dataset, MapDatasetOnOff):
             resampled_dataset.models = dataset.models
         else:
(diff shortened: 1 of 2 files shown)


gammapy/estimators/map/tests/test_excess.py
```diff
@@ -170,20 +173,44 @@ def test_significance_map_estimator_map_dataset(simple_dataset):
...
+
+    simple_dataset_mask_safe.exposure = None
+
+    result = estimator.run(simple_dataset_mask_safe)
+
+    assert_allclose(result["npred_excess"].data[0, 10, 10], 208)
+    assert_allclose(result["npred_background"].data[0, 10, 10], 208)
+    assert_allclose(result["sqrt_ts"].data[0, 10, 10], 12.676681, atol=1e-5)
+
+    assert_allclose(result["npred_excess_err"].data[0, 10, 10], 20.396078, atol=1e-3)
```
bkhelifi, line 207: I do not understand why the errors are larger, when having less masked bins... Any idea?

bkhelifi (changes requested):
Thanks @QRemy ... I have minor comments before the merging

gammapy/estimators/map/excess.py
```diff
@@ -228,6 +228,17 @@ def run(self, dataset):
         resampled_dataset = dataset.resample_energy_axis(
             energy_axis=axis, name=dataset.name
         )
+
+        if dataset.exposure:
+            reco_exposure = estimate_exposure_reco_energy(
+                dataset, self.spectral_model, normalize=False
+            )
+            reco_exposure = reco_exposure.resample_axis(
+                axis=axis, weights=dataset.mask_safe
```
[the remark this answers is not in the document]
QRemy: No, otherwise we have to add it everywhere,  I think it makes sense to have it only for the class that create the safe mask.

gammapy/estimators/map/tests/test_excess.py
QRemy: npred_excess is also larger with less masked bin so the relative error npred_excess_err/npred_excess is actually lower of about 1%.

bkhelifi (approved):
Thanks @QRemy

PR #5386, opened 2024-07-21
Cleanup MapEvaluator

QRemy:
Cleanup MapEvaluator

### DIFF
--- a/gammapy/datasets/evaluator.py
+++ b/gammapy/datasets/evaluator.py
@@ -373,8 +373,7 @@ class MapEvaluator:
 
     def apply_psf(self, npred):
         """Convolve npred cube with PSF."""
-        tmp = npred.convolve(self.psf)
-        return tmp
+        return npred.convolve(self.psf)
 
     def apply_edisp(self, npred):
         """Convolve map data with energy dispersion.
--- a/gammapy/datasets/evaluator.py
+++ b/gammapy/datasets/evaluator.py
@@ -92,9 +92,6 @@ class MapEvaluator:
         self._cached_parameter_values_spatial = None
         self._cached_position = (0, 0)
         self._computation_cache = None
-        self._spatial_oversampling_factor = 1
-        if exposure is not None:
-            self.update_spatial_oversampling_factor(self.geom)
 
     def _repr_html_(self):
         try:
@@ -240,7 +237,6 @@ class MapEvaluator:
                 self.exposure = exposure._cutout_view(
                     position=self.position, width=self.cutout_width, odd_npix=True
                 )
-        self.update_spatial_oversampling_factor(self.geom)
 
         self.reset_cache_properties()
         self._computation_cache = None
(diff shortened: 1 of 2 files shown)


bkhelifi (approved):
Thanks a lot @QRemy

gammapy/datasets/evaluator.py
```diff
@@ -318,14 +300,10 @@ def _compute_flux_spatial(self):
...
 
-            wcs_geom = self.geom.to_wcs_geom(width_min=self.cutout_width).to_image()
+            wcs_geom = self.geom.to_wcs_geom(width_min=self.cutout_width)
+            values = self._compute_flux_spatial_geom(wcs_geom)
 
-            if self.psf and self.model.apply_irf["psf"]:
-                values = self._compute_flux_spatial_geom(wcs_geom)
-            else:
-                values = self.model.spatial_model.integrate_geom(
-                    wcs_geom, oversampling_factor=1
```
registerrier, line 327: I understand why you remove here because it is basically unused. But `oversampling=1` is a bug right?

registerrier (commented):
Thanks @QRemy. I am not sure I completely follow the motivation. There is a bug here right? the oversampling_factor shouldn't be set to 1 when calling integrate_geom.

gammapy/datasets/evaluator.py
QRemy: yes it is probably a bug, it was not tested before, should I open a separate for that or we backport this PR ?

registerrier: there has been a mix-up when working on https://github.com/gammapy/gammapy/pull/3440

Well I would try to see what happens with the correct oversampling factor.

QRemy: This line is triggered only for region geom when self.model.apply_irf["psf"] is False. If psf is None we directly return 1 (line 301). With this modification and the test added it also returns 1 for the case self.model.apply_irf["psf"] is False.

QRemy: doen in #5390

gammapy/datasets/tests/test_evaluator.py
```diff
@@ -32,8 +32,8 @@ def evaluator():
     )
 
     spectral_model = ConstantSpectralModel()
-    spatial_model = PointSpatialModel(
-        lon_0=0 * u.deg, lat_0=0 * u.deg, frame="galactic"
+    spatial_model = GaussianSpatialModel(
+        lon_0=0 * u.deg, lat_0=0 * u.deg, sigma=0.001 * u.deg, frame="galactic"
```
QRemy, line 36:
```suggestion
    spatial_model = PointSpatialModel(
        lon_0=0 * u.deg, lat_0=0 * u.deg, frame="galactic"```

gammapy/datasets/evaluator.py
QRemy: rebased after #5390 so now this PR contains only cleanup

PR #5389, opened 2024-07-24
Fix `LabelMapAxis` so it doesn't reorder the labels

### LINKED ISSUE #5384: LabelMapAxis re-orders labels
**Gammapy version**
v1.1

**Bug description**
When instancing a `LabelMapAxis` the labels are re-ordered / sorted. This is problematic because it should keep the order, that corresponds to the  data.

**Expected behavior**
Do not re-order / sort the labels.

**To Reproduce**
```python
import numpy as np
from gammapy.maps import LabelMapAxis

labels = np.array([3, 2, 1, 4, 5])
print(labels)
print(LabelMapAxis(labels).center)
```
Which prints:
```
[3 2 1 4 5]
[1 2 3 4 5]
```

Astro-Kirsty:
Resolves #5384

Note: had to ensure that labels was an array type, as this is what np.unique would do

### DIFF
--- a/gammapy/maps/tests/test_axes.py
+++ b/gammapy/maps/tests/test_axes.py
@@ -797,12 +797,16 @@ def test_label_map_axis_basics():
     labels = np.array([1, 4, 5])
     assert_allclose(labels, [1, 4, 5])
     assert_allclose(LabelMapAxis(labels).center, labels)
+    assert_allclose(LabelMapAxis(labels)._labels, labels)
 
     # Test case with non unique labels
     with pytest.raises(ValueError) as exc_info:
         LabelMapAxis([1, 1, 1])
-
     assert str(exc_info.value) == "Node labels must be unique"
+
+    # Ensure non alphabetical ordering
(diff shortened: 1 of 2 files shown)


gammapy/maps/tests/test_axes.py
```diff
@@ -793,6 +793,17 @@ def test_label_map_axis_basics():
     axis_copy = axis.copy()
     assert axis_copy.name == "label-axis"
 
+    # Ensure order is correct
+    labels = np.array([1, 4, 5])
```
registerrier, line 797: Could you add a test checking that alphabetical order of labels is not imposed?

registerrier (approved):
Thanks @Astro-Kirsty . Looks good. It just misses one test for non re-ordering.

adonath (approved):
Thanks @Astro-Kirsty, I have no further comments.

QRemy (approved):
Thanks @Astro-Kirsty merging now

PR #5390, opened 2024-07-24
Fix region evaluation without psf convolution

QRemy:
See https://github.com/gammapy/gammapy/pull/5386/files#r1689548126

### DIFF
--- a/gammapy/datasets/tests/test_evaluator.py
+++ b/gammapy/datasets/tests/test_evaluator.py
@@ -31,8 +31,8 @@ def evaluator():
     )
 
     spectral_model = ConstantSpectralModel()
-    spatial_model = PointSpatialModel(
-        lon_0=0 * u.deg, lat_0=0 * u.deg, frame="galactic"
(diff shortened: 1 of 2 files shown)


gammapy/datasets/tests/test_evaluator.py
```diff
@@ -31,8 +31,8 @@ def evaluator():
     )
 
     spectral_model = ConstantSpectralModel()
-    spatial_model = PointSpatialModel(
-        lon_0=0 * u.deg, lat_0=0 * u.deg, frame="galactic"
+    spatial_model = GaussianSpatialModel(
```
bkhelifi, line 34: I propose to keep the test with the PointSpatialModel... We had issues in the past with such model

bkhelifi (changes requested):
Thanks @QRemy

gammapy/datasets/tests/test_evaluator.py
QRemy: done

PR #5391, opened 2024-07-25
Backport PR #5381 on branch v1.2.x (Apply safe mask in TSMapEstimator)

QRemy:
Backport PR #5381 on branch v1.2.x (Apply safe mask in TSMapEstimator)

### DIFF
--- a/gammapy/estimators/map/ts.py
+++ b/gammapy/estimators/map/ts.py
@@ -456,6 +456,7 @@ class TSMapEstimator(Estimator, parallel.ParallelMixin):
             repeat(maps["background"].data.astype(float)),
             repeat(maps["kernel"].data),
             repeat(maps["norm"].data),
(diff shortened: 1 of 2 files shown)


gammapy/estimators/map/ts.py
```diff
@@ -497,9 +527,6 @@ def run(self, dataset):
                 energy_min=energy_min, energy_max=energy_max, name=dataset.name
             )
 
-            if self.sum_over_energy_groups:
-                dataset_sliced = dataset_sliced.to_image(name=dataset.name)
```
QRemy, line 501: origin of #6448
yet I'm still not sure what should be the correct behaviour of sum_over_axes for the TSmapEstimator
because this squash the dataset before computing npred so the edisp will be different and it ignores the mask_fit

PR #5408, opened 2024-08-01
Add alpha maps to `ExcessMapEstimator`

### LINKED ISSUE #5247: Returning Alpha Map from 'ExcessMapEstimator'
Relating to #5245, identifying problems relating to acceptance in `ExcessMapEstimator` is difficult because only returns a `FluxMap` which contains no alpha information. 

I would like the ability to produce an alpha map with the output of  `ExcessMapEstimator.run()`. 

You can generate an implied alpha map using the TS map and the On, and scaled Off but this is rather clunky and requires a user defined inverse significance function.  

**Additional context**
Alpha is often a critical tool for assessing issues and uncertainties in significance calculations. Returning this information as a diagnostic immediately highlights potential issues when a user has selected parameters that contain issues (i.e. too small a width in their ring background). It feels clunky to add this information to the 'FluxMap' because you should be able to generate maps and significances where alpha isn't a significant factor in the calculation of the background. It could be just an optional return of 'ExcessMapeEstimator' which is what I currently do.

Astro-Kirsty:
This resolves #5247.
The general idea is that we sometime want access to the correlated on and off and alpha values. We currently only handle these values within `convolved_map_dataset_counts_statistics`. The idea is to create a `Maps` object that is a dictionary of maps containing the required maps. Then if the `full_output` option is selected it will create the maps, provided that a `MapDatasetOnOff` is provided.
This was one idea of implementation that @registerrier and I discussed. 
I think this seems like the easiest way that won't break API.  Because we just have the default `full_output=False`, which is as it was before.

### DIFF
--- a/gammapy/estimators/map/core.py
+++ b/gammapy/estimators/map/core.py
@@ -597,19 +597,19 @@ class FluxMaps:
 
     @property
     def alpha(self):
-        """Alpha maps."""
+        """The normalisation, alpha, for differences between the on and off regions."""
         self._check_quantity("alpha")
         return self._data["alpha"]
 
     @property
     def acceptance_on(self):
-        """acceptance_on."""
+        """The acceptance in the on region."""
         self._check_quantity("acceptance_on")
         return self._data["acceptance_on"]
 
     @property
     def acceptance_off(self):
-        """acceptance_off."""
+        """The acceptance in the off region"""
         self._check_quantity("acceptance_off")
         return self._data["acceptance_off"]
 
--- a/gammapy/estimators/map/excess.py
+++ b/gammapy/estimators/map/excess.py
@@ -156,13 +156,13 @@ class ExcessMapEstimator(Estimator):
         Which additional maps to estimate besides delta TS, significance and symmetric error.
         Available options are:
 
-            * "all": all the optional steps are executed
+            * "all": all the optional steps are executed.
             * "errn-errp": estimate asymmetric errors.
             * "ul": estimate upper limits.
-            * "sensitivity": estimate sensitivity for a given significance
-            * "alpha": normalisation factor to accounts for differences between the on and off regions
-            * "acceptance_on": acceptance from the on region
-            * "acceptance_off": acceptange from the off region
+            * "sensitivity": estimate sensitivity for a given significance.
+            * "alpha": normalisation factor to accounts for differences between the on and off regions.
+            * "acceptance_on": acceptance from the on region.
+            * "acceptance_off": acceptange from the off region.
 
         Default is None so the optional steps are not executed.
         Note: "alpha", "acceptance_on" and "acceptance_off" can only be selected if the dataset is a


adonath:
Thanks a lot @Astro-Kirsty! One question first: have you discussed the option with @registerrier to make theses maps just part of `FluxMaps`?

Astro-Kirsty:
That is another option we discussed, to include them as user option in the `selection_optional`! 
Open to opinions on what we think is best

gammapy/estimators/map/excess.py
```diff
@@ -67,9 +72,17 @@ def convolved_map_dataset_counts_statistics(dataset, kernel, mask, correlate_off
...
+            maps = Maps(
+                acceptance_on=acceptance_on, acceptance_off=acceptance_off, alpha=alpha
+            )
+            return WStatCountsStatistic(
+                n_on_conv.data, n_off.data, alpha.data, npred_sig_convolve.data
+            ), maps
+        else:
+            return WStatCountsStatistic(
+                n_on_conv.data, n_off.data, alpha.data, npred_sig_convolve.data
+            )
```
QRemy, line 85: maybe introduce a private function get_convolved_maps(dataset, kernel, mask, correlate_off) that returns a dict of `convolved_maps` and use it in `convolved_map_dataset_counts_statistics(convolved_maps, staty_type=dataset.stat_type)` (it is not part of the public api so it think it is fine to change its arguments).
Then in `estimate_excess_map` you add `if alpha in selection_optional : maps['alpha']=...`

QRemy (changes requested):
Thanks @Astro-Kirsty, my proposal would be to add alpha in `selection_optional` and to have it also as an optional  entry in the flux maps. Then to avoid having two outputs in `convolved_map_dataset_counts_statistics`  it could be splited in two functions as in the comment below :

adonath:
I'm with @QRemy here, just add them as optional selection and data members to `FluxMaps`, this way we achieve a more consistent and convenient API.

I think changing the return type based on configuration / input to a function is an anti pattern in programming. It typically leads to more code complexity later. It is much better to stick with dictionary like output classes with optional quantities.

Astro-Kirsty:
Okay, thanks for the input. I will adapt the code accordingly

Astro-Kirsty:
I've made the adaptions @adonath and @QRemy. 
I still need to create proper tests, and make sure docstrings are good, but thought you could give your general thoughts?

Also tests will fail for `selection_optional=["all"]` because we can only calculate the new parameters if there it is a `MapDatasetOnOff`. Not certain how we want to proceed there, have empty maps for that option when MapDataset or the 'all' option doesn't select those ones

gammapy/estimators/map/excess.py
```diff
@@ -128,12 +173,15 @@ class ExcessMapEstimator(Estimator):
...
     sum_over_energy_groups : bool
         Only used if energy_edges is None.
-        If False apply the estimator in each energy bin of the parent dataset.
-        If True apply the estimator in only one bin defined by the energy edges of the parent dataset.
+        If False, apply the estimator in each energy bin of the parent dataset.
+        If True, apply the estimator in only one bin defined by the energy edges of the parent dataset.
         Default is False.
+    full_output : bool
+        Whether to return the full output for a `~MapDatasetOnOff` which includes a `~Maps` object
+        for correlated on, correlated off and alpha. Default is False.
```
QRemy, line 184:
```suggestion
```
No longer needed

gammapy/estimators/map/excess.py
```diff
@@ -175,6 +230,7 @@ def __init__(
         bkg_syst_fraction_sensitivity=0.05,
         apply_threshold_sensitivity=False,
         sum_over_energy_groups=False,
+        full_output=False,
```
QRemy, line 233:
```suggestion
```

gammapy/estimators/map/excess.py
```diff
@@ -187,6 +243,7 @@ def __init__(
         self.energy_edges = energy_edges
         self.sum_over_energy_groups = sum_over_energy_groups
         self.correlate_off = correlate_off
+        self.full_output = full_output
```
QRemy, line 246:
```suggestion
```
```suggestion
```

gammapy/estimators/map/excess.py
```diff
@@ -396,6 +452,27 @@ def estimate_excess_map(self, dataset, reco_exposure):
...
+                    )
+                else:
+                    maps["acceptance_on"] = convolved_maps["acceptance_on"]
+            if "acceptance_off" in self.selection_optional:
+                if not isinstance(dataset, MapDatasetOnOff):
+                    raise ValueError(
+                        "This option can only be selected for 'MapDatasetOnOff."
+                    )
+                else:
+                    maps["acceptance_off"] = convolved_maps["acceptance_off"]
```
QRemy, line 475:
```suggestion
            if isinstance(dataset, MapDatasetOnOff):
                keys_onoff = set(["alpha", "acceptance_on", "acceptance_off"])
                for key in keys_onoff.intersection(self.selection_optional):
                    maps[key] = convolved_maps[key]
```

Here I think we can avoid silently ignore the keys that are not applicable to the `MapDaset` as long as the docstring states clearly that alpha is available only for the  `MapDatasetOnOff`.

QRemy (commented):
Thanks @Astro-Kirsty, I made some simplifications.
In particular I think we can silently ignore the keys that are not applicable to the MapDaset as long as the docstring states clearly that alpha is available only for the MapDatasetOnOff.

gammapy/estimators/map/excess.py
```diff
@@ -117,8 +160,13 @@ class ExcessMapEstimator(Estimator):
             * "errn-errp": estimate asymmetric errors.
             * "ul": estimate upper limits.
             * "sensitivity": estimate sensitivity for a given significance
+            * "alpha": normalisation factor to accounts for differences between the on and off regions
+            * "acceptance_on": acceptance from the on region
+            * "acceptance_off": acceptange from the off region
 
-        Default is None so the optional steps are not executed.
+        Default is None so the optional steps are not executed.2
```
QRemy, line 167:
```suggestion
        Default is None so the optional steps are not executed.
```

gammapy/estimators/map/core.py
```diff
@@ -138,6 +141,9 @@ class FluxMaps:
         * sqrt_ts : optional, the square root of the test statistic, when relevant.
         * success : optional, a boolean tagging the validity of the estimation.
         * n_dof : optional, the number of degrees of freedom used in TS computation
+        * alpha : optional, normalisation factor to accounts for differences between the test region and the background
+        * acceptance_off : optional, acceptance from the off region
+        * acceptance_on : optional, acceptance from the oon region
```
QRemy, line 146:
```suggestion
        * acceptance_on : optional, acceptance from the on region
```

QRemy (approved):

AtreyeeS (approved):

PR #5409, opened 2024-08-01
Add strict option to `MapAxis.downsample`

### LINKED ISSUE #5047: Modify `MapAxis.downsample`
Modify `MapAxis.downsample` to have an option strict=False option which when set to True will simply put the remainder in the last bin
 https://github.com/gammapy/gammapy/pull/5007#discussion_r1455938626_

Astro-Kirsty:
This PR resolves #5047

Adding a strict option to the downsample, such that when it is true, it places the remainder in the last bin

### DIFF
--- a/gammapy/maps/axes.py
+++ b/gammapy/maps/axes.py
@@ -1113,7 +1113,7 @@ class MapAxis:
     def downsample(self, factor, strict=True):
         """Downsample map axis by a given factor.
 
-        When downsampling each n-th (given by the factor) bin is selected from
+        When downsampling, each n-th (given by the factor) bin is selected from
         the axis while preserving the axis limits. For node type "edges" this
         requires nbin to be dividable by the factor, for node type "center" this
         requires nbin - 1 to be dividable by the factor.
@@ -1123,7 +1123,7 @@ class MapAxis:
         factor : int
             Downsampling factor.
         strict : bool
-            Whether the number of bins, is strictly divisible by the factor.
+            Whether the number of bins is strictly divisible by the factor.
             If True, ``nbin`` must be divisible by the ``factor``.
             If False, the reminder bins are put into the last bin of the new axis.
             Default is True.
--- a/gammapy/maps/axes.py
+++ b/gammapy/maps/axes.py
@@ -1133,34 +1133,20 @@ class MapAxis:
         axis : `MapAxis`
             Downsampled map axis.
         """
-        if self.node_type == "edges":
-            nbin = self.nbin / factor
-            if not strict and np.mod(nbin, 1) > 0:
-                idx = np.arange(0, self.nbin, factor)
-                if idx[-1] < self.nbin:
-                    idx = np.append(idx, self.nbin)
-                edges = self.edges[idx]
+        if self.node_type == "center":
+            raise ValueError(
+                f"Downsampling is not possible for {self.name} axis with node_type=center"
+            )
+
+        edges = self.edges[::factor]
+        if edges[-1] != self.edges[-1]:
+            if strict is True:
+                raise ValueError(
+                    f"Number of {self.name} bins is not divisible by {factor}"
+                )
             else:
-                if np.mod(nbin, 1) > 0:
-                    raise ValueError(
-                        f"Number of {self.name} bins is not divisible by {factor}"
-                    )
-                edges = self.edges[::factor]
-            return self.from_edges(edges, name=self.name, interp=self.interp)
-        else:
-            nbin = (self.nbin - 1) / factor
-            if not strict and np.mod(nbin, 1) > 0:
-                idx = np.arange(0, self.nbin - 1, factor)
-                if idx[-1] < self.nbin - 1:
-                    idx = np.append(idx, self.nbin - 1)
-                nodes = self.center[idx]
-            else:
-                if np.mod(nbin, 1) > 0:
-                    raise ValueError(
-                        f"Number of {self.name} bins - 1 is not divisible by {factor}"
-                    )
-                nodes = self.center[::factor]
-            return self.from_nodes(nodes, name=self.name, interp=self.interp)
+                edges = np.append(edges, self.edges[-1])
+        return self.from_edges(edges, name=self.name, interp=self.interp)
 
     def to_header(self, format="ogip", idx=0):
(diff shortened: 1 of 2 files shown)


QRemy (commented):
Thanks, @Astro-Kirsty, intuitively I would have expected that `strict=True` impose that the number bins is strictly divisible by factor.
Do you plan to have this used also in the Map.dowmsample ? because having the domsampling factor different for the last bins will break the code there.

gammapy/maps/axes.py
```diff
@@ -1122,6 +1122,10 @@ def downsample(self, factor):
         ----------
         factor : int
             Downsampling factor.
+        strict : bool
+
+            If True, then the remainder is placed in the last bin
```
AtreyeeS, line 1127: The current docstring `requires nbin to be dividable by the factor..`, must be adapted.
I guess the behaviour should be
- `strict=True` : nbin must to be divisible by the factor 
- `strict=False` the reminder bins are put in the last bin of the new axis

gammapy/maps/axes.py
```diff
@@ -1131,23 +1135,38 @@ def downsample(self, factor):
         if self.node_type == "edges":
             nbin = self.nbin / factor
 
-            if np.mod(nbin, 1) > 0:
-                raise ValueError(
-                    f"Number of {self.name} bins is not divisible by {factor}"
-                )
+            if strict and np.mod(nbin, 1) > 0:
```
AtreyeeS, line 1138: this should be `strict=False`

gammapy/maps/axes.py
```diff
@@ -1131,23 +1135,38 @@ def downsample(self, factor):
...
+                edges = self.edges[idx]
 
-            edges = self.edges[::factor]
-            return self.from_edges(edges, name=self.name, interp=self.interp)
+            else:
+                if np.mod(nbin, 1) > 0:
+                    raise ValueError(
+                        f"Number of {self.name} bins is not divisible by {factor}"
+                    )
+                edges = self.edges[::factor]
```
AtreyeeS, line 1149: Do you still need this?
```
idx = np.arange(0, self.nbin, factor)
 if idx[-1] < self.nbin:
      idx = np.append(idx, self.nbin)
 edges = self.edges[idx]
```
should suffice.
The `ValueError` can be raised inside the `if` statement `if strict is True`

gammapy/maps/axes.py
```diff
@@ -1131,23 +1135,38 @@ def downsample(self, factor):
...
+            downsampled = self.from_edges(edges, name=self.name, interp=self.interp)
+            return downsampled
         else:
             nbin = (self.nbin - 1) / factor
 
-            if np.mod(nbin, 1) > 0:
-                raise ValueError(
-                    f"Number of {self.name} bins - 1 is not divisible by {factor}"
-                )
+            if strict and np.mod(nbin, 1) > 0:
```
AtreyeeS, line 1155: I don't think you need the separate handling for node type anymore either

gammapy/maps/tests/test_axes.py
```diff
@@ -180,6 +180,26 @@ def test_up_downsample_consistency(factor):
...
+def test_strict_downsample():
+    axis1 = MapAxis.from_bounds(
+        10.0, 2e3, 6, interp="log", name="energy_true", unit="GeV"
+    )
+    with pytest.raises(ValueError) as exc_info:
+        axis1.downsample(4)
+    assert str(exc_info.value) == "Number of energy_true bins is not divisible by 4"
+
+    axis1_strict_down = axis1.downsample(4, strict=True)
+    assert_allclose(axis1_strict_down.nbin, 0.5 * (axis1.nbin) - 1)
```
AtreyeeS, line 192: What is this number you are checking against? It should be `(axis1.nbin)/factor + 1`, no?

AtreyeeS (changes requested):
Thank you @Astro-Kirsty ! Left some comments inline. 
You can also adapt `get_edges_fixed_bins` in `estimators.utils`

gammapy/maps/axes.py
AtreyeeS: On second thoughts, I think you do. To be able to preserve the `node_type`
But then, this logic needs modification

gammapy/maps/axes.py
Astro-Kirsty: When running the tests, it appears I still need it

gammapy/maps/tests/test_axes.py
Astro-Kirsty: That would be `(axis1.nbin)/4 + 1 = 2.5`

AtreyeeS:
The current behaviour for `node_type=center` looks strange to me.
1. Why does it need `nbin-1` to be divisible by `factor` ? 
2. The rebinned axis is binned from the left, and the bins look strange - see last 2 lines of example 
3. Is it at all reasonable to downsample an axis defined at node centres?

```
from gammapy.maps import MapAxis
ax1 = MapAxis.from_nodes(np.linspace(0,3,6))

ax1.downsample(2) # Leads to an error - ValueError: Number of  bins - 1 is not divisible by 2

ax2 = ax1.downsample(5)
# ax2 now has 2 bins

# or
ax1 = MapAxis.from_nodes(np.linspace(0,9,10))
ax2= ax1.downsample(3)
# ax2 has 4 bins
ax2.edges # this starts from -1.5 whereas ax1.edges starts from -0.5

ax2.center # [0. 3. 6. 9.]
ax1.center # [0. 1. 2. 3. 4. 5. 6. 7. 8. 9.]
```

Is there any motivation for this @registerrier @adonath ? Otherwise, I propose to forbid downsampling for node_type=center

adonath:
> ax2.edges # this starts from -1.5 whereas ax1.edges starts from -0.5

Yes, this is because the bin size has increased, while the center has been kept the same. I hope this make sense.

AtreyeeS:
Thanks @adonath for the clarification. I understand now that downsampling makes perfect sense, it's simply skipping every nth bin as you said. The `while preserving the axis limits` wording confused me.

I have kept the old behaviour.

registerrier (approved):
Thanks @AtreyeeS @Astro-Kirsty . This looks good to me.

PR #5412, opened 2024-08-13
Adjust counts type in fake

Astro-Kirsty:
Should we make the data type the same in the `fake` operation as the initial counts?

I came across this when trying to combing a `fake` dataset with a normal one. Do we want this kind of behaviour always?

### DIFF
--- a/gammapy/datasets/map.py
+++ b/gammapy/datasets/map.py
@@ -1374,6 +1374,7 @@ class MapDataset(Dataset):
         npred = self.npred()
         data = np.nan_to_num(npred.data, copy=True, nan=0.0, posinf=0.0, neginf=0.0)
         npred.data = random_state.poisson(data)
+        npred.data = npred.data.astype("float64")
         self.counts = npred
 
     def to_hdulist(self):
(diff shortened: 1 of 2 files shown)


adonath:
Thanks, I agree this is the most reasonable choice.

registerrier (commented):
Thanks @Astro-Kirsty . 
You might use simple `float` here, no?
A simple test checking the type of faked counts could be useful.

adonath:
@Astro-Kirsty But does the behavior suggested by @registerrier  resolve the problem you encountered? Can you please add a unit test for the error you got?

Astro-Kirsty:
Unrelated I believe but when I run pytest locally it crashes my computer when trying to run `test_map_dataset_region_geom_npred()`.  The line in question is 2105 `npred = dataset_spec.npred()`

registerrier (approved):

Does the test fail for stacking without your change?

Astro-Kirsty:
Hmm, no. I'll try for a different test

registerrier (approved):

PR #5420, opened 2024-08-28
Support for array valued regions

adonath:
This is a prototype for the support of array valued regions in Gammapy, that I worked on a few years back. I did not have time to continue the work and will not in the near future, however I think it is worth opening the PR as a basis for discussions. 

CC @Tobychev and @kosack 

See also https://github.com/gammapy/gammapy/discussions/5417

Notebook demonstrating the core functionality: https://github.com/adonath/notebooks-public/blob/master/region-geom-array.ipynb

The overall vision / design is the following:

- `regions` does not support array valued regions (for no good reason). See also https://github.com/astropy/regions/issues/497
- From the Gammapy side we could introduce `CirclePixelRegionArray` and `CircleSkyRegionArray` as a workaround and overwrite or re-implement the functionality we need. For now there is no real need for other shapes than a circle.
- The `RegionGeom` gets a `is_regular` attribute, which handles the changing region size the same way we handle changing bin sizes with energy in `WcsGeom`
- The key technical part here is the broadcasting: the energy dependent quantities (center / radius) of the region have to be broadcasted to the same dimension as the energy array. Then everything becomes very simple and "just works" (see the linked notebook above)
- The biggest challenges are serialization, testing and documentation.
- We can introduce `CircleSkyRegionArray.from_rad_max()` for convenience.

### DIFF
--- a/gammapy/utils/regions.py
+++ b/gammapy/utils/regions.py
@@ -619,7 +619,7 @@ class ArrayQuantityLength(RegionAttr):
 
     def _validate(self, value):
         if not isinstance(value, u.Quantity):
-            raise ValueError(f"The {self.name} must be an astropy " "Quantity object")
+            raise ValueError(f"The {self.name} must be an astropy  `Quantity object`")
 
 
 class ArrayLength(RegionAttr):
--- a/gammapy/utils/regions.py
+++ b/gammapy/utils/regions.py
@@ -618,7 +618,7 @@ class ArrayQuantityLength(RegionAttr):
     """
 
     def _validate(self, value):
-        if not (isinstance(value, u.Quantity)):
+        if not isinstance(value, u.Quantity):
             raise ValueError(f"The {self.name} must be an astropy " "Quantity object")
 
 
--- a/gammapy/utils/regions.py
+++ b/gammapy/utils/regions.py
@@ -18,7 +18,6 @@ from astropy.table import Table
 from gammapy.utils.scripts import make_path
 from regions import (
     CircleAnnulusSkyRegion,
-    CirclePixelRegion,
     CircleSkyRegion,
     CirclePixelRegion,
     CompoundSkyRegion,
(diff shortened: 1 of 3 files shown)


adonath:
@Tobychev In the (spatial) background integration we rely on `geom.get_wcs_coord_and_weights()`(see https://github.com/gammapy/gammapy/blob/main/gammapy/makers/utils.py#L265). With the changes introduced in this PR, the energy dependent region size background integration should mostly work out of the box (minus small issues...not tested).

gammapy/utils/regions.py
```diff
@@ -274,3 +279,54 @@ def region_circle_to_ellipse(region):
...
+
+
+class ArrayQuantityLength(RegionAttr):
+    """
+    Descriptor class for `~regions.SkyRegion`, which takes a scalar
+    `~astropy.units.Quantity` object.
+    """
+
+    def _validate(self, value):
+        if not (isinstance(value, u.Quantity)):
```
VascoSch92, line 294:
```suggestion
        if not isinstance(value, u.Quantity):
```

gammapy/utils/regions.py
```diff
@@ -274,3 +279,54 @@ def region_circle_to_ellipse(region):
...
+            raise ValueError(f"The {self.name} must be an astropy " "Quantity object")
```
VascoSch92, line 295:
```suggestion
            raise ValueError(f"The {self.name} must be an astropy  `Quantity object`")
```

adonath:
Thanks @VascoSch92!

adonath:
@registerrier How should we proceed with this? Write a PIG first and the continue implementation? The core functionality is already in place, but to support it everywhere in Gammapy more work is needed. Especially on the serialization...

QRemy (approved):
Thanks @adonath, added unit tests and fixed few bugs after rebase.

Astro-Kirsty:
@QRemy, can you explain a bit the use cases for these type of regions? Maybe I miss something but why is this better than just having a list of `Regions`?

PR #5423, opened 2024-09-03
Function for the discrete cross correlation function

cgalelli:
This PR introduces a function in `variability.py` to compute the discrete correlation function as proposed [by Edelson et al](https://ui.adsabs.harvard.edu/abs/1988ApJ...333..646E/abstract). I added the usual wrapper function to use directly `FluxPoints` objects.


bkhelifi (approved):
Thanks a lot! This is a nice new algorithms for the time variability. Tutorials will be interesting to make and to promote.

gammapy/estimators/utils.py
```diff
@@ -536,6 +542,62 @@ def compute_lightcurve_doublingtime(lightcurve, flux_quantity="flux"):
     return table
 
 
+def compute_dcf(lightcurve1, lightcurve2=None, flux_quantity="flux", tau=1 * u.h):
+    r"""Compute the discrete correlation function for two lightcurves of the discrete autocorrelation if only one lightcurve is provided.
```
morcuended, line 546:
```suggestion
    r"""Compute the discrete correlation function for two lightcurves or the discrete autocorrelation if only one lightcurve is provided.
```

gammapy/stats/variability.py
```diff
@@ -286,6 +286,78 @@ def structure_function(flux, flux_err, time, tdelta_precision=5):
     return sf, distances
 
 
+def discrete_correlation_function(
+    flux1, flux_err1, flux2, flux_err2, time1, time2, tau, axis=0
```
Astro-Kirsty, line 290: Given the ordering of your docstring I would say it should be:

```suggestion
    flux1, flux2, flux_err1, flux_err2, time1, time2, tau, axis=0
```

Astro-Kirsty (changes requested):
Thanks @cgalelli. I would suggest to recall it `discrete_correlation` without the `function`. Also instead of abbreviations, such as `dcf`, I would write the full name. Otherwise (at least for me) it is a bit unclear exactly what the parameter is.

gammapy/stats/variability.py
cgalelli: Whoops! I think it's better to change the docstrings in this case.

cgalelli:
Thanks for the comments @Astro-Kirsty . The discrete correlation function as described here is a way to compute correlation between two data trains, or autocorrelation in one, while accounting for lossy sampling and measurement errors. The first use case here is the use for non continuously sampled lightcurves, but I am sure it has other useful cases

gammapy/estimators/utils.py
```diff
@@ -536,6 +542,62 @@ def compute_lightcurve_doublingtime(lightcurve, flux_quantity="flux"):
...
+    flux_err1 = getattr(lightcurve1, flux_quantity + "_err")
+    coords1 = lightcurve1.geom.axes["time"].center
+    axis = flux1.geom.axes.index_data("time")
+
+    if lightcurve2:
+        flux2 = getattr(lightcurve2, flux_quantity)
+        flux_err2 = getattr(lightcurve2, flux_quantity + "_err")
+        coords2 = lightcurve2.geom.axes["time"].center
+        return discrete_correlation(
+            flux1.data,
```
bkhelifi, line 578: Can you add a dedicated test for this `if` case?

bkhelifi (changes requested):
Thanks @cgalelli

gammapy/estimators/tests/test_utils.py
```diff
@@ -12,6 +12,7 @@
     compute_lightcurve_doublingtime,
     compute_lightcurve_fpp,
     compute_lightcurve_fvar,
+    compute_dcf,
```
Astro-Kirsty, line 15:
```suggestion
    compute_discrete_correlation,
```

Astro-Kirsty: Plus all other instances

gammapy/estimators/utils.py
```diff
@@ -536,6 +542,62 @@ def compute_lightcurve_doublingtime(lightcurve, flux_quantity="flux"):
...
+        The second lightcurve object. If not provided, the autocorrelation for the first lightcurve will be computed. Default is None.
+    flux_quantity : str
+        Flux quantity to use for calculation. Should be 'dnde', 'flux', 'e2dnde' or 'eflux'. Default is 'flux'.
+    tau : `~astropy.units.Quantity`
+        Size of the bins to compute the discrete correlation. Default is 1 hour.
+
+
+    Returns
+    -------
+    bincenters, dcf, dcf_err : `~numpy.ndarray`, `~astropy.units.Quantity`
```
Astro-Kirsty, line 564:
```suggestion
    bincenters, discrete_correlation, discrete_correlation_err : `~numpy.ndarray`, `~astropy.units.Quantity`
```

Astro-Kirsty: Plus all other instances

Astro-Kirsty (changes requested):
Instead of abbreviations I would still use the full names

gammapy/stats/variability.py
```diff
@@ -286,6 +287,86 @@ def structure_function(flux, flux_err, time, tdelta_precision=5):
...
+    bins = (
+        np.linspace(
+            minfactor, maxfactor, int(np.abs(maxfactor) + np.abs(minfactor) + 1)
+        )
+        * tau
+    )
+
+    bin_indices = np.digitize(dist, bins).flatten()
+
+    udcf = np.reshape(udcf, (udcf.shape[0] * udcf.shape[1], -1))
```
Astro-Kirsty, line 350:
```suggestion
    udcf = udcf.reshape(-1, udcf.shape[-1])
```

Astro-Kirsty (changes requested):

gammapy/stats/variability.py
Astro-Kirsty: I think this simplification should work

cgalelli: Thanks, unfortunately, it does not seem to produce the same results

Astro-Kirsty (approved):

gammapy/estimators/tests/test_utils.py
```diff
@@ -199,6 +233,52 @@ def test_compute_lightcurve_doublingtime():
     )
 
 
+def test_compute_dcf():
+    lighcurve = lc()
+    lightcurve2 = lc2()
```
morcuended, line 238: I do not know if that'd make a difference in practice, but shouldn't lc and lc2 be fixtures?

bkhelifi (approved):
Thanks a lot! It looks fully ready to be checked by @registerrier or @AtreyeeS

AtreyeeS (commented):
Thank you for this useful function @cgalelli ! I have some questions (mostly about bad user inputs) left inline.

In general, for such functionalities, it would be nice to update the user guide with some details on the implementation and the limits of validity. It can be a separate PR.

gammapy/stats/variability.py
```diff
@@ -286,6 +287,86 @@ def structure_function(flux, flux_err, time, tdelta_precision=5):
...
+    ----------
+    .. [Edelson1988] "THE DISCRETE CORRELATION FUNCTION: A NEW METHOD FOR ANALYZING
+    UNEVENLY SAMPLED VARIABILITY DATA", Edelson et al. (1988)
+    https://ui.adsabs.harvard.edu/abs/1988ApJ...333..646E/abstract
+    """
+
+    flux1 = np.rollaxis(flux1, axis, 0)
+    flux2 = np.rollaxis(flux2, axis, 0)
+
+    tau = tau.to(time1.unit)
```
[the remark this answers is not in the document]
cgalelli: Yes that was the assumption. You are right that this should not be trusted, I should probably either write a warning/error or force one of the two to convert. Probably just converting is fine?

gammapy/stats/variability.py
```diff
@@ -286,6 +287,86 @@ def structure_function(flux, flux_err, time, tdelta_precision=5):
...
+
+    mean1, mean2 = np.nanmean(flux1, axis=0), np.nanmean(flux2, axis=0)
+    sigma1, sigma2 = np.nanstd(flux1, axis=0), np.nanstd(flux2, axis=0)
+
+    udcf1 = (flux1 - mean1) / np.sqrt((sigma1**2 - np.nanmean(flux_err1, axis=0) ** 2))
```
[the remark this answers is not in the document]
cgalelli: Well the input is the full error, but then yes I use only the mean. This actually stems from a mistake in the paper which was then corrected in following notes. The first paper uses the actual errors associated to each data point while the correct way to do this is to use just the mean. Maybe I can switch to simply requesting for the mean as input.

gammapy/stats/variability.py
```diff
@@ -286,6 +287,86 @@ def structure_function(flux, flux_err, time, tdelta_precision=5):
...
+    udcf2 = (flux2 - mean2) / np.sqrt((sigma2**2 - np.nanmean(flux_err2, axis=0) ** 2))
+
+    udcf = np.empty(((flux1.shape[0],) + flux2.shape))
```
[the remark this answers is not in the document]
cgalelli: Should add a check with error or warning. I am not sure it is safe to allow arrays with different non-trivial shapes.

gammapy/stats/tests/test_variability.py
```diff
@@ -257,3 +258,87 @@ def test_structure_function():
...
+            [0.360007, 0.360725],
+            [0.754145, 1.38026],
+        ],
+        rtol=1e-5,
+    )
+
+    assert_allclose(
+        bins,
+        u.Quantity([-22500.0, -7500.0, 7500.0, 22500.0], unit="s"),
+    )
```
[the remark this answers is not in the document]
cgalelli: The problem is that at tau=0, since tau is the size of lag bins, you would just get empty bins...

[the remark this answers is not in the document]
AtreyeeS: but isn't it true by definition for autocorrelation?

gammapy/stats/variability.py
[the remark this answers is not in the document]
AtreyeeS: The input is fine. I was checking from the original paper and got confused about the implementation. maybe add a citation to the corrected notes.

gammapy/stats/variability.py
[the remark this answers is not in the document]
AtreyeeS: Yes, please raise an error. I think its better to avoid giving results which are ill understood

gammapy/stats/tests/test_variability.py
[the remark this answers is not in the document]
cgalelli: Tau=0 would mean bin size is 0, not 0 lag - maybe I misunderstood the comment

gammapy/estimators/tests/test_utils.py
```diff
@@ -199,6 +235,49 @@ def test_compute_lightcurve_doublingtime():
     )
 
 
+def test_compute_dcf():
```
morcuended, line 238: then you use those fixtures defined above as:

```python
def test_compute_dcf(lc, lc2):
    lighcurve = lc
    lightcurve2 = lc2
    ...
```
so it's clear what inputs the test needs

gammapy/estimators/utils.py
```diff
@@ -536,6 +542,79 @@ def compute_lightcurve_doublingtime(lightcurve, flux_quantity="flux"):
...
+    Parameters
+    ----------
+    lightcurve1 : `~gammapy.estimators.FluxPoints`
+        The first lightcurve object.
+    lightcurve2 : `~gammapy.estimators.FluxPoints`, optional
+        The second lightcurve object. If not provided, the autocorrelation for the first lightcurve will be computed. Default is None.
+    flux_quantity : str
+        Flux quantity to use for calculation. Should be 'dnde', 'flux', 'e2dnde' or 'eflux'. Default is 'flux'.
+    tau : `~astropy.units.Quantity`
+        Size of the bins to compute the discrete correlation. Default is 1 hour.
```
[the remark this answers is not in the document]
AtreyeeS: I did not mean that tau shoudl always be obtained automatically. I was thinking of replacing the default 1 hr with something more aligned with the time bins of the light curves

gammapy/estimators/utils.py
```diff
@@ -536,6 +542,79 @@ def compute_lightcurve_doublingtime(lightcurve, flux_quantity="flux"):
...
+    Internally calls the `~gammapy.stats.discrete_correlation` function
+
+    Parameters
+    ----------
+    lightcurve1 : `~gammapy.estimators.FluxPoints`
+        The first lightcurve object.
+    lightcurve2 : `~gammapy.estimators.FluxPoints`, optional
+        The second lightcurve object. If not provided, the autocorrelation for the first lightcurve will be computed. Default is None.
+    flux_quantity : str
+        Flux quantity to use for calculation. Should be 'dnde', 'flux', 'e2dnde' or 'eflux'. Default is 'flux'.
```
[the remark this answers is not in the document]
AtreyeeS: Then maybe you can mention that in the doc-string as well

gammapy/estimators/utils.py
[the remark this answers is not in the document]
cgalelli: Thinking more about it, a static default is not great, as you point out. A default in general is hard to fix and I guess would be overridden most of the times. to keep it very simple, it could be something like (length of the time axis)/(N/2) where N is the number of current bins.

AtreyeeS (dismissed):
Thanks @cgalelli ! No further comments

registerrier (commented):
Thanks @cgalelli . I have left a number of inline comments. Most are cosmetic (docstring, code style). 

- Do the time arrays have to be the same? If yes, that should reflect in the docstring. And `discrete_correlation` should have only one time input. Otherwise, the other case should be tested. 
- Same question regarding the lightcurve lengths?. Do they have to be the same?
- Couldn't you use meshgrids for the correlation calculations?

registerrier (approved):
Thanks @cgalelli . This looks good.

We need to experiment a bit with this, and document it somewhere in a future PR.

Issue #3616, opened 2021-11-19
Spectral fitting within LightCurveEstimator


mcerruti:
I'd like to study spectral variability in AGN light-curves: I start from theoretical models and would like to extract, in addition to the integral light-curve, also the change in i.e. photon index or cut-off energy 

**Describe the solution you'd like**
Currently the LightCurveEstimator fits only the normalization of the model. I would like to have the option to also leave the index free  (or any other parameter of the input model)

**Describe alternatives you've considered**
I can of course loop over the models myself and compute indexes and integral fluxes per single simulated spectra, but it will much better if it is just an option of the light-curve core estimator 

**Additional context**

adonath:
Thanks @mcerruti, indeed this is a very valid use case. Technically this is rather simple to implement as and extension to the `LightCurveEstimator`, but it requires a bit of work on the bookkeeping side. This means we have to extend the `FluxPoints` object to handle time varying reference models (which works by introducing a time axis to e.g. `FluxPoints.dnde_ref`) and add a data member to handle additional model parameters. Long term we probably also have to extend the serialization formats (https://gamma-astro-data-formats.readthedocs.io/en/latest/spectra/flux_points/index.html) to describe this. But as far as I can tell there is nothing that would block this from the Gammapy side.

registerrier:
The estimate produced in such a context cannot be an extension of `FluxPoints`. It is a different object since it is mostly a collection of parameters (or even model) ordered with time. No energy binning is involved.

Isn't such a utility close to a possible `SkyModelEstimator` as discussed in #3702 . If an estimator is designed to evaluate a model component TS and the associated best fit parameters, it could directly be extended to perform the same analysis in time intervals, exactly as the `LightCurveEstimator` with the `FluxPointsEstimator`.

maxnoe:
Sorry, I am not as familiar with the internals, but I thought with `reoptimize=True`, the `LightCurveEstimator` not only fits the amplitude but also the other parameters of the model. 

So obtaining index / break energy for a power law or powerlaw with cutoff already works with `LightCurveEstimator`, correct? 

Is the problem that this information is not stored when running the `LightCurveEstimator`?

registerrier:
> Sorry, I am not as familiar with the internals, but I thought with `reoptimize=True`, the `LightCurveEstimator` not only fits the amplitude but also the other parameters of the model.
> 
[...]

PR #5427, opened 2024-09-10
A time resolved spectroscopy estimator

AtreyeeS:
As mentioned by @cgalelli  in the last dev call, there is a necessity for a Time resolved spectroscopy estimator.
After internal discussions, we think having the output as a `FluxMaps` container is more cumbersome.

This PR is a first sketch of such an estimator.
I have put it as a tutorial now without removing the output because it can help to understand the workflow.

If the idea is ok, the estimator can be moved to gammapy.estimators and the tutorial can show only the latter parts with some additional plots.

@cgalelli can you take over from here please?


examples/tutorials/analysis-time/TimeResolvedSpectroscopyEstimator.ipynb
```diff
@@ -0,0 +1,948 @@
+{
```
bkhelifi, line 1: <pre class="ql-syntax" spellcheck="false">Line #84.            result = self.fit.run(datasets_in_bin)
</pre><p>The result of the fit is not propagated to higher level? What about is it does not converge?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/gammapy/gammapy/pull/5427/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

bkhelifi:
Thank a lot! This future estimator is so expected by the AGN/GRB community... Well done ;)

examples/tutorials/analysis-time/TimeResolvedSpectroscopyEstimator.ipynb
AtreyeeS: good point @bkhelifi ! We can have a status message column in the table.
Do we need to save all the fit_results ? How should that be serialized? 

@cgalelli Do we need some measure of fit statistics in every bin? What should that be, del TS with and without the model?

registerrier (commented):
Thanks @AtreyeeS . This is indeed a long standing issue which we should tackle. I remind of issue #3616 where we started discussing this.

I think the main difficulties remain to be solved. Namely:
- How detailed should the end-product be? (correlations between parameters, stats surface, confidence contours etc). No specifications no far.
- How to handle non-converging fits? It is expected that fits will have difficulty converging in numerous time bins. What kind of information can be used? Should one use a simpler model (e.g. PL rather than ECPL) in the absence of convergence etc?

Note also that the problematic is more general. You could use a similar algorithm to perform fits in a collection of region based spectra to build a spectral index map for instance. Or spectra in rings around a position. 

Maybe looking into generalizing the `TestStatisticNested` class approach is an option. This returns a dictionary with ts value and `FitResult` objects for both h0 and H1 hypotheses.

AtreyeeS:
Thanks @registerrier ! My main motivation was to show a recipe that making hysteresis plots can be done rather easily with Gammapy. One option might be to show this as a recipe while the standing issues are addressed?

> I think the main difficulties remain to be solved. Namely:
> 
>     * How detailed should the end-product be? (correlations between parameters, stats surface, confidence contours etc). No specifications no far.

Lightcurves made with fermipy do not provide confidence contours etc (as far as I know).
If the fit models are serialised, then the correlation matrix should also be available, no?
Is there a use case for having stat surfaces for multiple bins?

>     * How to handle non-converging fits? It is expected that fits will have difficulty converging in numerous time bins. What kind of information can be used? Should one use a simpler model (e.g. PL rather than ECPL) in the absence of convergence etc?
,
Should it not be the job of the user to choose a good model?
I agree that a flag should be raised if the fit does not converge, but more than that is difficult to automate

cgalelli:
Hi @registerrier , me and @AtreyeeS were just discussing in private the general output and in particular how to tackle non-converging fits. In my idea, just how @AtreyeeS said in her previous comment, it would not be the goal of this estimator to find the best model for the analysis, that would be on the user. In my opinion, the estimator could present a boolean option:

- If False, store nothing, even when the fit does not converge
- If True, a sort of "diagnostic/verbose mode" where we could store the whole fit result. This might seem heavy but would give the user all the information possible about the fit

I think in general going into more complex options would be confusing for the user.

Regarding the general output, models are easily serialized and give a lot of information - I think if we start giving more complicated outputs, we might start getting into "contamination" territory with LightCurve

examples/tutorials/analysis-time/TimeResolvedSpectroscopyEstimator.ipynb
cgalelli: Hi @bkhelifi and @AtreyeeS , good question. In general as I wrote below I think it would be a good idea at least to have an option where user could ask for the full serialization of fit_results

AtreyeeS:
> Maybe looking into generalizing the TestStatisticNested class approach is an option. This returns a dictionary with ts value and FitResult objects for both h0 and H1 hypotheses.

What is the H0 here? It should be a "background only" or equivalent, ie, the significance of the source in each bin. The current `TestStaticticNested` has H0 set at some fixed parameter value. 

> Note also that the problematic is more general. You could use a similar algorithm to perform fits in a collection of region based spectra to build a spectral index map for instance. Or spectra in rings around a position.

Interesting point! This can be more complex, because there can be a need to fit overlapping models. We will need something like our standard estimators with a reoptimise option.

examples/tutorials/analysis-time/TimeResolvedSpectroscopyEstimator.ipynb
bkhelifi: Indeed, this option could be great.

 And the new (mandatory) column should contain the status... but one should handle:
- what about if migrad converges but not hess (the errors)? maybe the status should be the logical AND (to be documented).
- what about the sherpa and scipy backends?

bkhelifi: I have seen that... I was speaking about their success flag. They are well defined? Just as a check...

cgalelli:
> > Maybe looking into generalizing the TestStatisticNested class approach is an option. This returns a dictionary with ts value and FitResult objects for both h0 and H1 hypotheses.
> 
> What is the H0 here? It should be a "background only" or equivalent, ie, the significance of the source in each bin. The current `TestStaticticNested` has H0 set at some fixed parameter value.

Philosophically, wouldn't a H0 be the result of the model fit over the whole time axis, without binning in time? Although I am not sure this is exactly what `TestStaticticNested` would imply.

registerrier:
| Regarding the general output, models are easily serialized and give a lot of information - I think if we start giving more 
| complicated outputs, we might start getting into "contamination" territory with LightCurve

Yes, that why I think that `FitResult` is the easiest and most complete output that can be returned.

 | What is the H0 here? It should be a "background only" or equivalent, ie, the significance of the source in each bin. The current TestStaticticNested has H0 set at some fixed parameter value.

Well you can test various situations like ECPL vs PL. Or a fixed index PL, vs a varying one. 
In general, we have a TS associated to an estimated flux quantity, you might have a similar info here.

Technically the main operation done here is splitting the `Datasets` object into time intervals. This might be the only convenience to provide actually.

AtreyeeS:
> Yes, that why I think that FitResult is the easiest and most complete output that can be returned.

I agree, but in this case it will be a list of FitResults. How should that be serailised?

For convenience, it might still be good to provide a Table with the spectral parameters at every time bin, because that is the most common use case.

 Regarding the TS in each bin, I think what is often done is using a Power Law (the simplest model), the detection significance is computed in each bin, More complex models/Finer time bins are chosen accordingly.
It might be sufficient to provide simply the TS in the bin.
It would then be the users responsibility to compare it against another hypothesis by rerunning the estimator with the relevant model.


We have been having this complain from the AGN community for a long time, so it cant be worthwhile to have a simple case ready for the 1.3 version for feedback.

AtreyeeS (commented):
Thanks @cgalelli !
This approach looks clear.

You need to add some markdown cells explaining what is going on.

I think you can first introduce the function `time_resolved_spectroscopy`, explain it returns a list of `FitResult` objects and show what they contain. Then add the `create_table` as a convenience functionality to extract  the spectral parameters and the success tag.

At the end, you can show how to, for ex, check the significant of the spectral break in each time bin

examples/tutorials/analysis-time/Time_resolved_spectroscopy.ipynb
```diff
@@ -0,0 +1,324 @@
...
+    "    \n",
+    "    col_unit.append(\"MJD\")\n",
+    "    col_unit.append(\"MJD\")\n",
+    "\n",
+    "    for par in fit_result[0].models.parameters.free_parameters:\n",
+    "        col_names.append(par.name)\n",
+    "        col_names.append(par.name + \"_err\")\n",
+    "        unt = par.unit\n",
+    "        if unt is u.Unit():\n",
+    "            unt = \"\"\n",
```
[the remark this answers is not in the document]
cgalelli: I tried looking for this issue on the astropy documentation but it seems handling dimensionless quantities together with quantities with units is convoluted. Couldn't find a simpler way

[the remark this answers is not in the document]
cgalelli: Modified the table function in general, now it is fine.

AtreyeeS:
Thanks @cgalelli ! I cleaned up the PR. I cant leave a review since its on my branch, but its okay from my side now

cgalelli (approved):
Thanks @AtreyeeS I think now the tutorial is very clear and quite complete!

bkhelifi (approved):
Thank you a lot, @AtreyeeS .
This tutorial will be very important for the community.

examples/tutorials/analysis-time/Time_resolved_spectroscopy.py
```diff
@@ -0,0 +1,351 @@
...
+Perform spectral fits of a blazar in different time bins to investigate
+spectral changes during flares.
+
+Context
+-------
+
+The `LightCurveEstimator` in Gammapy (see
+:doc:`light curve notebook </tutorials/analysis-time/light_curve>`,
+and
+:doc:`light curve for flares notebook </tutorials/analysis-time/light_curve_for_flares>`.)
```
Astro-Kirsty, line 17:
```suggestion
:doc:`light curve for flares notebook </tutorials/analysis-time/light_curve_flare>`.)
```

examples/tutorials/analysis-time/Time_resolved_spectroscopy.py
```diff
@@ -0,0 +1,351 @@
...
+
+Proposed approach
+-----------------
+
+The main idea behind doing a time resolved spectroscopy is to
+
+-  Select relevant `~gammapy.data.Observations` from the
+   `~gammapy.data.DataStore`
+-  Define time intervals in which to fit the spectral model
+-  Apply the above time seclections on the data to obatin new
```
Astro-Kirsty, line 31:
```suggestion
-  Apply the above time selections on the data to obtain new
```

examples/tutorials/analysis-time/Time_resolved_spectroscopy.py
```diff
@@ -0,0 +1,351 @@
...
+   `~gammapy.data.Observations`
+-  Perform standard data reduction on the above data
+-  Define a source model
+-  Fit the reduced data in each time bin with the source model
+-  Extract relevant informations in a table
```
Astro-Kirsty, line 36:
```suggestion
-  Extract relevant information in a table
```

examples/tutorials/analysis-time/Time_resolved_spectroscopy.py
```diff
@@ -0,0 +1,351 @@
...
+
+Here, we will use the PKS 2155-304 observations from the
+`H.E.S.S. first public test data release <https://www.mpi-hd.mpg.de/hfm/HESS/pages/dl3-dr1/>`\ \_\_.
```
Astro-Kirsty, line 39:
```suggestion
`H.E.S.S. first public test data release <https://www.mpi-hd.mpg.de/hfm/HESS/pages/dl3-dr1/>`__.
```

examples/tutorials/analysis-time/Time_resolved_spectroscopy.py
```diff
@@ -0,0 +1,351 @@
...
+)
+
+log = logging.getLogger(__name__)
+
+
+######################################################################
+# Data selection
+# ~~~~~~~~~~~~~~
+#
+# We select all runs pointing within 2 degrees of PKS2155-304.
```
Astro-Kirsty, line 85:
```suggestion
# We select all runs pointing within 2 degrees of PKS 2155-304.
```

examples/tutorials/analysis-time/Time_resolved_spectroscopy.py
```diff
@@ -0,0 +1,351 @@
...
+obs_ids = data_store.obs_table.select_observations(selection)["OBS_ID"]
+observations = data_store.get_observations(obs_ids)
+print(f"Number of selected observations : {len(observations)}")
+
+
+######################################################################
+# The flaring observations were taken during July 2006. We define
+# 15-minutes time intervals as lists of `~astropy.Time` start and stop
+# objects, and apply the intervals to the observations by using
+# `~observations.select_time`
```
Astro-Kirsty, line 106:
```suggestion
# 15-minute time intervals as lists of `~astropy.time.Time` start and stop
# objects, and apply the intervals to the observations by using
# `~gammapy.data.Observations.select_time`
```

examples/tutorials/analysis-time/Time_resolved_spectroscopy.py
```diff
@@ -0,0 +1,351 @@
...
+print(short_observations[1].gti)
+
+
+######################################################################
+# Data reduction
+# --------------
+#
+# In this example, we perform a 1D analysis with a reflected regions
+# background estimation. For details, see
+# :doc:/tutorials/analysis-1d/spectral_analysis tutorial.
```
Astro-Kirsty, line 129:
```suggestion
# background estimation. For details, see the
# :doc:`/tutorials/analysis-1d/spectral_analysis` tutorial.
```

examples/tutorials/analysis-time/Time_resolved_spectroscopy.py
```diff
@@ -0,0 +1,351 @@
...
+for obs in short_observations:
+    dataset = dataset_maker.run(dataset_empty.copy(), obs)
+
+    dataset_on_off = bkg_maker.run(dataset, obs)
+    dataset_on_off = safe_mask_masker.run(dataset_on_off, obs)
+    datasets.append(dataset_on_off)
+
+
+######################################################################
+# This gives us list of `SpectrumDatasetOnOff` which can now be
```
Astro-Kirsty, line 162:
```suggestion
# This gives us list of `~gammapy.datasets.SpectrumDatasetOnOff` which can now be
```

examples/tutorials/analysis-time/Time_resolved_spectroscopy.py
```diff
@@ -0,0 +1,351 @@
...
+sky_model = SkyModel(spatial_model=None, spectral_model=spectral_model, name="pks2155")
+print(sky_model)
+
+
+######################################################################
+# Time resolved spectroscopy algorithm
+# ------------------------------------
+#
+# The following function is the crux of this tutorial. The `sky_model`
+# is fit in each bin and a list of `fit_results` stores the fit
```
Astro-Kirsty, line 197:
```suggestion
# The following function is the crux of this tutorial. The ``sky_model``
# is fit in each bin and a list of ``fit_results`` stores the fit
```

examples/tutorials/analysis-time/Time_resolved_spectroscopy.py
```diff
@@ -0,0 +1,351 @@
...
+
+######################################################################
+# Or, to access the fitted models,
+#
+
+print(results[0].models)
+
+
+######################################################################
+# To better visualize the data, we can create a table by extracting some
```
Astro-Kirsty, line 252: visualise?

examples/tutorials/analysis-time/Time_resolved_spectroscopy.py
```diff
@@ -0,0 +1,351 @@
...
+
+    return t
+
+
+table = create_table(valid_times, results)
+print(table)
+
+
+######################################################################
+# Visualisting the results
```
Astro-Kirsty, line 283:
```suggestion
# Visualising the results
```

examples/tutorials/analysis-time/Time_resolved_spectroscopy.py
```diff
@@ -0,0 +1,351 @@
...
+# ~~~~~~~~~~~~~~~~~~~~~~~~
+#
+# We can plot the spectral index and the amplitude as a function of time.
+# For convenience, we will convert the times into a `TimeMapAxis`.
```
Astro-Kirsty, line 287:
```suggestion
# For convenience, we will convert the times into a `~gammapy.maps.TimeMapAxis`.
```

examples/tutorials/analysis-time/Time_resolved_spectroscopy.py
```diff
@@ -0,0 +1,351 @@
...
+plt.ylabel("index")
+plt.colorbar()
+plt.show()
+
+
+######################################################################
+# Exercises
+# ---------
+#
+# 1. Quantify the variablity in the spectral index
```
Astro-Kirsty, line 341:
```suggestion
# 1. Quantify the variability in the spectral index
```

examples/tutorials/analysis-time/Time_resolved_spectroscopy.py
```diff
@@ -0,0 +1,346 @@
...
+from astropy.table import QTable
+from astropy.time import Time
+from regions import CircleSkyRegion
+
+# %matplotlib inline
+import matplotlib.pyplot as plt
+
+log = logging.getLogger(__name__)
+
+from gammapy.data import GTI, DataStore
```
Astro-Kirsty, line 64:
```suggestion
from gammapy.data import DataStore
```

Astro-Kirsty (commented):
Thanks for the effort @AtreyeeS and @cgalelli. As the science has already been discussed, I only suggest some cosmetic changes!

examples/tutorials/analysis-time/Time_resolved_spectroscopy.py
```diff
@@ -0,0 +1,346 @@
...
+Aim
+---
+
+Perform spectral fits of a blazar in different time bins to investigate
+spectral changes during flares.
+
+Context
+-------
+
+The `LightCurveEstimator` in Gammapy (see
```
Astro-Kirsty, line 14:
```suggestion
The `~gammapy.estimators.LightCurveEstimator` in Gammapy (see
```

Astro-Kirsty (dismissed):
Thank you @AtreyeeS!

examples/tutorials/analysis-time/Time_resolved_spectroscopy.py
```diff
@@ -0,0 +1,346 @@
...
+######################################################################
+# Data reduction
+# --------------
+#
+# In this example, we perform a 1D analysis with a reflected regions
+# background estimation. For details, see the
+# :doc:`/tutorials/analysis-1d/spectral_analysis` tutorial.
+#
+
+# Target definition
```
Astro-Kirsty, line 132:
```suggestion
```

Astro-Kirsty (approved):

cgalelli (approved):
Thanks @AtreyeeS ! No more modifications from me

registerrier:
Sorry for the late comment. 
Looking at the resulting final plot, I think one would expect error bars and a legend for the z axis. It took me some time to understand what it represents.

PR #5434, opened 2024-09-16
Include further keywords in `codemeta.json`

### LINKED ISSUE #4617: Improve codemeta.py script to include releveant keywords for codemeta.json
This issue is a followup to PR #4594 .

Here we should list the keywords we want to include on the codemeta.json file and include them in the scripts that updates the codemeta file produced by cffconvert.

### LINKED ISSUE #5095: Update `codemeta.py'
The code that generates the codemeta.json file should be updated in order to add:
  -  "programmingLanguage": "Python",
  -  "referencePublication": "https://doi.org/10.1051/0004-6361/202346488"

Astro-Kirsty:
This resolves #4617 and resolves #5095

Most of the keywords we want to add can be directly to the `json` file as they will not change across different versions. Do we actually need anything else defined in the `codemeta.py` if is the same across versions?
- Not sure about the `runtimePlatform`?
- In v1.0 we had: `dateCreated`, `datePublished` and `dateModified`. But now we do not use `dateModified`, how do we want to proceed?
- Do we want the `programmingLanguage` to be `Python` or `Python 3`

### DIFF
--- a/codemeta.json
+++ b/codemeta.json
@@ -253,7 +253,7 @@
         }
     ],
     "codeRepository": "https://github.com/gammapy/gammapy",
-    "dateCreated": "2014-01-01",
+    "dateCreated": "2013-08-19",
     "datePublished": "2014-08-25",
     "dateModified": "2024-02-29",
     "description": "Gammapy analyzes gamma-ray data and creates sky images, spectra and lightcurves, from event lists and instrument response information; it can also determine the position, morphology and spectra of gamma-ray sources. It is used to analyze data from H.E.S.S., Fermi-LAT, HAWC, and the Cherenkov Telescope Array (CTA).",
@@ -269,7 +269,7 @@
     "url": "https://gammapy.org/",
     "softwareVersion": "v1.2",
     "applicationCategory": "Astronomy",
-    "applicationSubCategory": "Very-high energy gamma rays",
+    "applicationSubCategory": "Very-high-energy gamma rays",
     "programmingLanguage": "Python 3",
     "operatingSystem": "Linux, Mac OS, Windows",
     "referencePublication": "https://doi.org/10.1051/0004-6361/202346488",
--- a/codemeta.json
+++ b/codemeta.json
@@ -253,18 +253,30 @@
         }
     ],
     "codeRepository": "https://github.com/gammapy/gammapy",
-    "datePublished": "2024-02-29",
+    "dateCreated": "2014-01-01",
+    "datePublished": "2014-08-25",
+    "dateModified": "2024-02-29",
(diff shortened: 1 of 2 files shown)


codemeta.json
```diff
@@ -253,18 +253,31 @@
...
-        "Data analysis"
+        "Data analysis",
+        "Scientific/Engineering"
     ],
     "license": "https://spdx.org/licenses/BSD-3-Clause",
     "name": "Gammapy: Python toolbox for gamma-ray astronomy",
     "url": "https://gammapy.org/",
     "softwareVersion": "v1.2",
+    "applicationCategory": "Astronomy",
+    "applicationSubCategory": "Very-high energy gamma rays",
```
cgalelli, line 272: Any possibility of being more broad here? e.g. HE gamma rays, VHE gamma rays, neutrino physics etc

registerrier (commented):
Thanks @Astro-Kirsty . 
The json file is updated on the fly by the codemeta.py script. The idea is rather to adapt the script to include necessary changes, in particular the dates that have to be updated at the time of the release.

Astro-Kirsty:
Okay @registerrier, but some of these are defined within the python file. For example, the key words. Is that because the file was edited afterwards and I should add anything that is missing? Or do I misunderstand something

bkhelifi (changes requested):
Thanks a lot @Astro-Kirsty. I have some minor comments

bkhelifi (approved):
Thanks a lot!

registerrier (approved):

I think the definition of publication date is a bit unclear in the codemeta specifications (since the parameter is associated to different versions). I would rather have e.g. the date of last LTS, which, I think, is more relevant than that of v0.1. Opinion @bkhelifi @adonath ?

registerrier:
Merging now. We will see if further changes are needed in another PR.

vuillaut:
> > Thanks @Astro-Kirsty . Looks good.
> > I think the definition of publication date is a bit unclear in the codemeta specifications (since the parameter is associated to different versions). I would rather have e.g. the date of last LTS, which, I think, is more relevant than that of v0.1. Opinion @bkhelifi @adonath ?
> 
> Not strongly! each release is a publication... But there are issues about that on which @vuillaut has worked: https://gitlab.in2p3.fr/escape2020/wp3/eossr/-/issues/13 and [codemeta/codemeta#285](https://github.com/codemeta/codemeta/pull/285). What do you recommand @vuillaut ? the date of the release (ie `time.now()`)?

Hi
I would recommend the date of the release but you could make a different choice that is more relevant to gammapy.

Concerning the keywords, I'd recommend you use the ones from https://astrothesaurus.org/ that has the advantage to be used across different services (but again, there is nothing *wrong* using a different vocabulary).

PR #5436, opened 2024-09-16
Add an inverted option in `Observations.select_time`

### LINKED ISSUE #5428: Add an `inverted` parameter on `Observations.select_time`
What about adding an `inverted` parameter on `Observations.select_time` ? 
That would allow the user to select all observations that lies outside of a given time range. We already support this kind of feature on `ObservationTable.select_time_range`

MRegeard:
This pull request fix #5428.

I simply added an argument on `Observations.select_time` method. When it is set to True, the method return all observations that lies outside of the time_interval.

### DIFF
--- a/gammapy/data/observations.py
+++ b/gammapy/data/observations.py
@@ -716,7 +716,7 @@ class Observations(collections.abc.MutableSequence):
         """List of observation IDs (`list`)."""
         return [str(obs.obs_id) for obs in self]
 
-    def select_time(self, time_intervals):
+    def select_time(self, time_intervals, inverted=False):
         """Select a time interval of the observations.
 
         Parameters
@@ -735,10 +735,14 @@ class Observations(collections.abc.MutableSequence):
 
         for time_interval in time_intervals:
             for obs in self:
-                if (obs.tstart < time_interval[1]) & (obs.tstop > time_interval[0]):
-                    new_obs = obs.select_time(time_interval)
-                    new_obs_list.append(new_obs)
-
+                if not inverted:
+                    if (obs.tstart < time_interval[1]) & (obs.tstop > time_interval[0]):
+                        new_obs = obs.select_time(time_interval)
+                        new_obs_list.append(new_obs)
+                else:
+                    if (obs.tstart > time_interval[1]) | (obs.tstop < time_interval[0]):
+                        new_obs = obs.select_time(time_interval)
+                        new_obs_list.append(new_obs)
         return self.__class__(new_obs_list)
 
     def _ipython_key_completions_(self):
--- a/gammapy/data/tests/test_observations.py
+++ b/gammapy/data/tests/test_observations.py
@@ -227,6 +227,10 @@ def test_observations_select_time_time_intervals_list(data_store):
     assert_time_allclose(new_obss[1].gti.time_start[0], time_intervals[1][0])
     assert_time_allclose(new_obss[1].gti.time_stop[-1], time_intervals[1][1])
 
+    out_obs = obss.select_time(time_intervals[-1], inverted=True)
+
+    assert len(out_obs) == 8
+
 
 @requires_data()
 def test_observation_cta_1dc():


gammapy/data/observations.py
```diff
@@ -735,10 +737,14 @@ def select_time(self, time_intervals):
...
-                    new_obs = obs.select_time(time_interval)
-                    new_obs_list.append(new_obs)
-
+                if not inverted:
+                    if (obs.tstart < time_interval[1]) & (obs.tstop > time_interval[0]):
+                        new_obs = obs.select_time(time_interval)
+                        new_obs_list.append(new_obs)
+                else:
+                    if (obs.tstart > time_interval[1]) | (obs.tstop < time_interval[0]):
+                        new_obs = obs.select_time(time_interval)
```
registerrier, line 746: This cannot work here. You apply the opposite condition to the events.

gammapy/data/tests/test_observations.py
```diff
@@ -227,6 +227,10 @@ def test_observations_select_time_time_intervals_list(data_store):
     assert_time_allclose(new_obss[1].gti.time_start[0], time_intervals[1][0])
     assert_time_allclose(new_obss[1].gti.time_stop[-1], time_intervals[1][1])
 
+    out_obs = obss.select_time(time_intervals[-1], inverted=True)
+
+    assert len(out_obs) == 8
```
registerrier, line 232: You should add a test on events too.

gammapy/data/observations.py
```diff
@@ -716,13 +716,15 @@ def ids(self):
         """List of observation IDs (`list`)."""
         return [str(obs.obs_id) for obs in self]
 
-    def select_time(self, time_intervals):
+    def select_time(self, time_intervals, inverted=False):
```
registerrier, line 719: We should be consistent in the code base. I think we more frequently use `exclude` as an argument name.

registerrier (changes requested):
Thanks @MRegeard . I think there is an issue in this implementation. `Observations.select_time` select observations in time intervals and events in individual observations. The same condition should therefore be applied to event selection. 

Also note the changes in #5305

gammapy/data/observations.py
MRegeard: I used `inverted` because it is the argument name used in `ObservationTable.select_time_range`. But I agree that we should be consistent in the code base, whether it is `excude` or `inverted`.

MRegeard:
Thanks @registerrier, I completely missed this part of the code ...

MRegeard:
I will rebase and make relevant changes once https://github.com/gammapy/gammapy/pull/5305 is merged.

registerrier:
Any news here @MRegeard ?

MRegeard:
@registerrier, I think this is good and just need to be merged, but it has been a while so let me check in more details !

Astro-Kirsty:
I have updated the PR. I wonder if we should not use the same conversion everywhere with `inverted_time` rather than both this and `inverted`.

gammapy/data/tests/test_event_list.py
```diff
@@ -5,6 +5,7 @@
 from astropy import units as u
 from astropy.coordinates import SkyCoord
 from astropy.table import QTable, Table
+from astropy.io import fits
```
AtreyeeS: Why do you need this?

gammapy/data/tests/test_event_list.py
```diff
@@ -247,6 +248,10 @@ def test_peek(self):
         with mpl_plot_check():
             self.events.peek()
 
+    def test_select_time(self):
+        interval = Time([53090.12346118, 53090.12431144], scale="tt", format="mjd")
+        assert len(self.events.select_time(interval).time) == 499
```
AtreyeeS: You can test for the `inverted=True` case as well

gammapy/data/gti.py
```diff
@@ -310,13 +310,16 @@ def from_time_intervals(cls, time_intervals, reference_time=None):
...
         return cls.create(start, stop, reference_time)
 
-    def select_time(self, time_interval):
+    def select_time(self, time_interval, inverted=False):
         """Select and crop GTIs in time interval.
 
         Parameters
         ----------
         time_interval : `astropy.time.Time`
             Start and stop time for the selection.
```
AtreyeeS: For the gti, we mention 
`Start time (inclusive) and stop time (exclusive) for the selection.`
this might be good to clarify everywhere, ie, how the edge values are handled

gammapy/data/observations.py
```diff
@@ -785,10 +792,14 @@ def select_time(self, time_intervals):
...
-                if (obs.tstart < time_interval[1]) & (obs.tstop > time_interval[0]):
-                    new_obs = obs.select_time(time_interval)
-                    new_obs_list.append(new_obs)
-
+                if not inverted:
+                    if (obs.tstart < time_interval[1]) & (obs.tstop > time_interval[0]):
+                        new_obs = obs.select_time(time_interval)
+                        new_obs_list.append(new_obs)
+                else:
+                    if (obs.tstart > time_interval[1]) | (obs.tstop < time_interval[0]):
```
AtreyeeS, line 803: is this completely correct?
lets say t1<t2<t3<t4...

time interval is beween t2 and t3
obs.tsart = t1, obs.tstop=t4

We should keep t1-t2 and t3-t4 for `inverted=True`.

This if statement will not be triggered, though

gammapy/data/gti.py
Astro-Kirsty: Where is this defined? I do not find it

gammapy/data/tests/test_event_list.py
```diff
@@ -247,6 +247,22 @@ def test_peek(self):
...
+        interval = Time([53090.12346118, 53090.12431144], scale="tt", format="mjd")
+        selected_time = self.events.select_time(interval)
+        assert len(selected_time.time) == 499
+        assert_allclose(selected_time.time.value[0], 53090.12346118442, rtol=1e-3)
+        assert_allclose(selected_time.time_ref.value, 51910.00074287037, rtol=1e-3)
+        selected_time_inverted = self.events.select_time(interval, inverted=True)
+
+        assert len(selected_time_inverted.time) == 10744
+        assert_allclose(
+            selected_time_inverted.time.value[0], 53090.12346118442, rtol=1e-3
```
AtreyeeS: It is same as the non inverted one ???

gammapy/data/event_list.py
```diff
@@ -339,13 +339,16 @@ def select_energy(self, energy_range):
...
         return self.select_row_subset(mask)
 
-    def select_time(self, time_interval):
+    def select_time(self, time_interval, inverted=False):
         """Select events in time interval.
 
         Parameters
         ----------
         time_interval : `astropy.time.Time`
             Start time (inclusive) and stop time (exclusive) for the selection.
```
AtreyeeS, line 348: @Astro-Kirsty this is what I meant. Sorry its on the event list, not gti

gammapy/data/tests/test_event_list.py
Astro-Kirsty: Yes but only because it was the 0th one! I should select another

gammapy/data/observations.py
Astro-Kirsty: I'm not sure I understand, can you explain further?

gammapy/data/tests/test_event_list.py
```diff
@@ -247,6 +247,22 @@ def test_peek(self):
         with mpl_plot_check():
             self.events.peek()
 
+    def test_select_time(self):
+        interval = Time([53090.12346118, 53090.12431144], scale="tt", format="mjd")
```
AtreyeeS: is this correct? the interval does not need to be increasing?

gammapy/data/observations.py
AtreyeeS: Lets say (starting from some reference time)

```
obs.tstart = 1
obs.tstop = 10
time_interval = [5,8]  

obs.select_time(time_interval, inverted=False) # returns a new obs between [5, 8]

obs.select_time(time_interval, inverted=True) # should return a new obs between [1,5] and [8,10]
```

however, the 2nd will not happen as obs.tstart < time_interval[1] , obs.tstop > time_interval[0]. So, the new observation will not be correctly created.

You can add this in the tests...

gammapy/data/tests/test_event_list.py
AtreyeeS: Guess it is good to be clear which is the open interval and which is the closed one and ensure its implemented consistently and correctly

gammapy/data/tests/test_event_list.py
Astro-Kirsty: There should be some time difference for sure! There is a slight increase here

gammapy/data/observations.py
Astro-Kirsty: Ahh true, I will look into it today

gammapy/data/tests/test_event_list.py
AtreyeeS: sorry! Misread 43 and 34!

gammapy/data/event_list.py
Astro-Kirsty: Ahh okay! Do we want to do that in this PR or another?

AtreyeeS: This would be just putting it in the docs.

I think its important, so that we are clear where the edge times go

gammapy/data/observations.py
Astro-Kirsty: Okay it because how we want the behaviour, because it does correctly do this but not as a separate object, i.e. in the example you give it will return:
`[1,5,8,10]'

gammapy/data/tests/test_event_list.py
```diff
@@ -247,6 +247,32 @@ def test_peek(self):
...
+        assert_allclose(selected_time.time.value[1], 53090.123652505696, rtol=1e-3)
+        assert_allclose(selected_time.time_ref.value, 51910.00074287037, rtol=1e-3)
+
+        selected_time_inverted = self.events.select_time(interval, inverted=True)
+        # One less because the interval is defined as start time (inclusive) and stop time (exclusive)
+        expected = np.concatenate([events_time[:100].value, events_time[200:].value])
+        assert len(selected_time_inverted.time) == len(expected)
+        assert_allclose(selected_time_inverted.time.value, expected)
+        assert_allclose(
+            selected_time_inverted.time.value[0], 53090.12346069794, rtol=1e-3
```
AtreyeeS: instead of checking values, maybe you can check directly on the `event[id].time.mjd`  to be sure of the correct event

gammapy/data/tests/test_event_list.py
```diff
@@ -247,6 +247,32 @@ def test_peek(self):
...
+        )
+        assert_allclose(
+            selected_time_inverted.time.value[1], 53090.12346118442, rtol=1e-3
```
AtreyeeS: We need lower tolerances here. 0.1% tol is too large.
Or maybe subtract the tref and assert

gammapy/data/gti.py
```diff
@@ -338,8 +341,11 @@ def select_time(self, time_interval):
         )
 
         gti_within["STOP"] = np.clip(gti_within["STOP"], interval_start, interval_stop)
+        gti_selected = gti_within
+        if inverted:
+            gti_selected = self.table[~mask]
```
AtreyeeS: The clipping is not necessary here?

gammapy/data/observations.py
```diff
@@ -527,14 +527,17 @@ def peek(self, figsize=(15, 10)):
...
 
-    def select_time(self, time_interval):
+    def select_time(self, time_interval, inverted=False):
         """Select a time interval of the observation.
 
         Parameters
         ----------
         time_interval : `astropy.time.Time`
-            Start and stop time of the selected time interval.
+            Start time (exclusive) and stop time (exclusive) of the selected time interval.
```
AtreyeeS: Start time is inclusive, right?

gammapy/data/tests/test_event_list.py
Astro-Kirsty: We cannot do it directly on the event, as it is an `EventList` which is not subscriptable i.e. you cannot access its elements via the square brackets

gammapy/data/tests/test_event_list.py
Astro-Kirsty: Another option is to use `assert_time_allclose` i.e.
`assert_time_allclose(selected_time_inverted.time[0], Time(53090.12346069794, format="mjd", scale="tt"))`

gammapy/data/tests/test_event_list.py
AtreyeeS: you have the times already in `events_time`.
So, `assert_allclose(selected_time.time.value[0], events_time[100].mjd], rtol)`

gammapy/data/tests/test_event_list.py
AtreyeeS: You will need to lower the tolerance there as well, The default is ms

gammapy/data/tests/test_event_list.py
AtreyeeS: or you can also assert on the energy of the event...

gammapy/data/observations.py
Astro-Kirsty: In observations `select_time` it is without equal signs:
`(obs.tstart > time_interval[1]) | (obs.tstop < time_interval[0]):`

AtreyeeS: We should have the same behaviour in `obs.select_time` and `events.select_time`! Otherwise it would be too chaotic to interpret !

Astro-Kirsty: I agree, but the issue actually starts with the fact that these were defined differently prior to the introduction of this PR :/

Should we go with: 
`Start time (inclusive) and stop time (exclusive)`
which is currently consistent with event list

AtreyeeS: Looks like a bug to me but we can't backport a feature PR :( 

I would go with this as a feature and mention in the changelog that start and stop time handling were homogenised between `select_time` functions on the event list, gti and observations.

I think `Start time (inclusive) and stop time (exclusive)` is reasonable as it follows numpy rules

QRemy: This change should be done in another PR, don't add the inclusive/exclusive details here. Anyway I bet it doesn't really matter because given the float precision in practice it might be always exclusive (and the number of events selected might even change from one machine to another).

Astro-Kirsty:
We need to wait for #6497 to be merged

gammapy/data/gti.py
```diff
@@ -338,8 +341,11 @@ def select_time(self, time_interval):
...
 
-        return self.__class__(gti_within)
+        return self.__class__(gti_selected)
```
QRemy:
```suggestion
        return self.__class__(gti_within)
```

gammapy/data/gti.py
QRemy, line 349:
```
        if inverted:
           return self.__class__(self.table[~mask])
      
        gti_within = self.table[mask]
``````

File: gammapy/data/gti.py (hunk 15 above)
Comment (Astro-Kirsty) [reply gammapy/data/gti.py]: It is not clear to me whether we need the clipping for the inverted case or not. Any ideas?

Comment (QRemy) [reply gammapy/data/gti.py]: If we mask something that is within START and STOP we don't need the cropping but if the masked interval include one of the edges the clipping is necessary.

Comment (Astro-Kirsty) [reply gammapy/data/gti.py]: Okay given our general definition of `time_start <= interval_stop` and `time_stop > interval_start` for the normal case, I guess we need the clipping?

File: gammapy/data/gti.py [hunk 16]
```diff
@@ -330,6 +333,10 @@ def select_time(self, time_interval):
         # get GTIs that fall within the time_interval
         mask = self.time_start <= interval_stop
         mask &= self.time_stop > interval_start
+
+        if inverted:
+            return self.table[~mask]
```
QRemy: I think the clipping for the inverted case should be done like this:
```suggestion
        if inverted:
            gti_selected = self.table[~mask]
            mask =  (gti_selected["START"] >= interval_start) & (gti_selected["STOP"] >= interval_stop])
            gti_selected["START"][mask] = interval_stop
            mask =  (gti_selected["START"] <= interval_start) & (gti_selected["STOP"] <= interval_stop])
            gti_selected["STOP"][mask] = interval_start
            return  self.__class__(gti_selected)

PR #5437, opened 2024-09-16
Remove mask in residual plotting

### LINKED ISSUE #5411: `SpectrumDatasetOnOff.plot_fit` shows residuals only within `mask_safe`
**Gammapy version**
1.2

**Bug description**
`SpectrumDatasetOnOff.plot_fit` shows the residuals (lower panel in the example plot below) only within the range that enters the fit, i.e. where `mask_safe` is `True`.
image

**Expected behavior**
I would expect the residuals between model and data also outside the fit range.

AtreyeeS:
This PR resolves #5411 by removing the mask in `dataset.plot_residuals_spectral`

I also adapted `dataset.plot_fit` to internally call `dataset.plot_masks()` and two of the associated tutorials

### DIFF
--- a/examples/tutorials/analysis-1d/extended_source_spectral_analysis.py
+++ b/examples/tutorials/analysis-1d/extended_source_spectral_analysis.py
@@ -342,5 +342,4 @@ reduced.models = model
 
 # Plot the result
 ax_spectrum, ax_residuals = reduced.plot_fit()
-reduced.plot_masks(ax=ax_spectrum)
 plt.show()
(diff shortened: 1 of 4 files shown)


lmohrmann:
Many thanks @AtreyeeS ! This looks good to me.

lmohrmann (approved):

registerrier (approved):
Thanks @AtreyeeS . 

Showing the full residuals is indeed more clear. As is the choice to plot the mask over the residuals as well.

Issue #5203, opened 2024-04-08
Speed-up sampling of large numbers of sky-models with `MapDatasetEventSampler`


fabiopintore:
**Is your feature request related to a problem? Please describe.**
`MapDatasetEventSampler` is quite slow when sampling a very large number of sky-models (hundreds of sources).

**Describe the solution you'd like**
Evaluate IRF values for all models just once?

adonath:
It would be good to do some profiling here and actually find out what the bottleneck is.

In general I think this is a bit tricky: when you have a large number of low flux sources, most of them do not create any event. However in general you cannot know which one. So you have to evaluate the model, sample the flux and see if you get an event or not. So you still have to evaluate the predicted counts for all source models. 

The sampling of the IRFs then scales with the number of events, not the number of sources. So if you have large number of low flux sources, the IRF sampling should not significantly increase in runtime. The runtime should be completely dominated by the high flux sources, which create most of the events.

registerrier:
What can we do for v1.3 @fabiopintore ?

fabiopintore:
@registerrier , I did some profiling of the simulator script with `pyinstrument`. I tested it for the simulations of about 50 sources in one of the SDC pointings. I got that most of the time was spent in the function `MapDatasetEventSampler.sample_sources` (as expected), with the bottleneck to be associated to `MapEvaluator.update`. In particular, to `PSFMap.get_psf_kernel`. I'm not sure that we can speed-up the simulating process with the current event sampler structure...

This is a summary of what I found:

<img width="1437" alt="Schermata 2024-09-14 alle 16 19 56" src="https://github.com/user-attachments/assets/416d022e-2692-4fb8-8569-9939f8befb57">
<img width="1436" alt="Schermata 2024-09-14 alle 16 20 50" src="https://github.com/user-attachments/assets/22a08b3f-8c1b-4244-b0a3-09870377ef56">
<img width="1437" alt="Schermata 2024-09-14 alle 16 21 18" src="https://github.com/user-attachments/assets/9173c9c5-d75f-4366-99d1-8242c04ea573">

adonath:
Thanks @fabiopintore, this is good to know! For event sampling one should not compute a PSF kernel at all, because the PSF is applied later in the sampling process. So the either the mechanism in the event sampling or `MapEvaluator.update` should be adjusted to avoid the computation of the kernel at all.

fabiopintore:
Hi @adonath . I think this issue can be closed.

PR #5438, opened 2024-09-16
Speed up event sampler

fabiopintore:
These PR aims at speeding up the event sampler when a large number of sources has to be simulated, as reported in #5203. The approach is to switch-off the PSF update of the `MapDataset`.

### DIFF
--- a/gammapy/datasets/simulate.py
+++ b/gammapy/datasets/simulate.py
@@ -1,5 +1,6 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 """Simulate observations."""
+
 import html
 import logging
 from copy import deepcopy
@@ -264,10 +265,11 @@ class MapDatasetEventSampler:
         events_all = EventList(Table())
         for idx, evaluator in enumerate(dataset.evaluators.values()):
             log.info(f"Evaluating model: {evaluator.model.name}")
+            psf_empty = None
             if evaluator.needs_update:
                 evaluator.update(
                     dataset.exposure,
-                    dataset.psf,
+                    psf_empty,
                     dataset.edisp,
                     dataset._geom,
                     dataset.mask,


adonath:
Thanks @fabiopintore! Can you please share the performance report before and after this change?

fabiopintore:
I've ran the simulations on my laptop. 
For `gammapy-1.2`, this is the report:
[output_gammapy-1.2.pdf](https://github.com/user-attachments/files/17020188/output_gammapy-1.2.pdf)

while for gammapy-dev:
[output_gammapy-dev.pdf](https://github.com/user-attachments/files/17020191/output_gammapy-dev.pdf)

The simulator seems now significantly faster, according to my test.

adonath:
Thanks @fabiopintore, indeed the absolute runtime is almost a factor of three smaller. Great! Can you add one small unit test and assert that the PSF is indeed none? To avoid that in future some other contributor accidentally reverts the change.

gammapy/datasets/simulate.py
```diff
@@ -247,13 +248,16 @@ def _sample_coord_time(self, npred, temporal_model, gti):
 
         return table
 
-    def sample_sources(self, dataset):
+    def sample_sources(self, dataset, psf_update=None):
```
AtreyeeS, line 251: It might be simpler if `update_psf` is a bool.
If `True`, pass the `evaluator.update` will take `dataset.psf`, else None

AtreyeeS (changes requested):
Thank you @fabiopintore ! This should be very useful for the event sampling! Nice that this did not affect any of the test values... 

I left an inline comment to simplify the API

gammapy/datasets/tests/test_simulate.py
```diff
@@ -354,6 +354,14 @@ def test_mde_sample_sources(dataset, models):
     assert_allclose(events.table["MC_ID"][0], 1, rtol=1e-5)
 
 
+@requires_data()
+def test_mde_sample_sources_psf_update(dataset, models):
+    dataset.models = models
+    sampler = MapDatasetEventSampler(random_state=0)
+    events = sampler.sample_sources(dataset=dataset, psf_update=None)
```
registerrier, line 361:
```suggestion
    events = sampler.sample_sources(dataset=dataset, psf_update=False)
```

gammapy/datasets/simulate.py
```diff
@@ -247,27 +248,34 @@ def _sample_coord_time(self, npred, temporal_model, gti):
...
 
-    def sample_sources(self, dataset):
+    def sample_sources(self, dataset, psf_update=False):
         """Sample source model components.
 
         Parameters
         ----------
         dataset : `~gammapy.datasets.MapDataset`
             Map dataset.
+        psf_update : `str` or `~gammapy.datasets.MapDataset`
```
registerrier, line 258:
```suggestion
        psf_update : bool
```

gammapy/datasets/simulate.py
```diff
@@ -247,27 +248,34 @@ def _sample_coord_time(self, npred, temporal_model, gti):
...
+            Parameter to switch-off (on) the update of the PSF
+            in the dataset; available options: `None` or `dataset.psf`.
```
registerrier, line 260:
```suggestion
            in the dataset; default is False.
```

registerrier (approved):
Thanks @fabiopintore !

registerrier:

adonath:
Thanks @fabiopintore! I have one additional comment: I would suggest to remove the `psf_update` flag completely in a follow-up PR. I think it has literally has no effect (as the PSF kernel is not used during event sampling), or did I miss something?

PR #5440, opened 2024-09-18
Fix TSmapEstimator stat_scan with sum_over_energy_group=False 

QRemy:
Fix TSmapEstimator for options sum_over_energy_group=False and selection_optional=["stat_scan"]

### DIFF
--- a/gammapy/estimators/map/tests/test_ts.py
+++ b/gammapy/estimators/map/tests/test_ts.py
@@ -580,6 +580,12 @@ def test_joint_ts_map_hawc():
     result = estimator.run(datasets)
     assert_allclose(result["flux"].data[0, 59, 59], 1.909396e-13, rtol=1e-3)
     assert_allclose(result["sqrt_ts"].data[0, 59, 59], 10.878956, rtol=1e-3)
+    assert result.stat_scan.geom.data_shape == (1, 109, 120, 120)
+    assert result.dnde_scan_values.geom.data_shape == (1, 109, 120, 120)
+    assert_allclose(
+        result["dnde_scan_values"].data[0, 0, 59, 59], -3.164557e-13, rtol=1e-3
+    )
+    assert_allclose(result["stat_scan"].data[0, 0, 59, 59], 5193.588657, rtol=1e-3)
 
     estimator = TSMapEstimator(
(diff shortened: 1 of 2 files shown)


gammapy/estimators/map/tests/test_ts.py
```diff
@@ -564,12 +564,36 @@ def test_joint_ts_map_hawc():
...
 
-    estimator = TSMapEstimator(kernel_width=2 * u.deg, sum_over_energy_groups=False)
+    estimator = TSMapEstimator(
+        kernel_width=2 * u.deg, sum_over_energy_groups=False, n_jobs=4
+    )
+    result = estimator.run(datasets)
+    assert_allclose(result["flux"].data[0, 59, 59], 1.909396e-13, rtol=1e-3)
+    assert_allclose(result["sqrt_ts"].data[0, 59, 59], 10.878956, rtol=1e-3)
+
+    estimator = TSMapEstimator(
```
AtreyeeS, line 574: Isnt the test getting repeated here?

gammapy/estimators/map/ts.py
```diff
@@ -511,7 +511,7 @@ def estimate_flux_map(self, datasets):
         j, i = zip(*positions)
 
         geom = maps[0]["counts"].geom.squash(axis_name="energy")
-        energy_axis = maps[0]["counts"].geom.axes["energy"]
+        energy_axis = geom.axes["energy"]
```
AtreyeeS, line 514: This seems a cleanup.

AtreyeeS (commented):
What was the error and what is the fix here? It seems you made one cleanup in the estimator.

gammapy/estimators/map/ts.py
QRemy: geom is squashed but not on the energy axis line

gammapy/estimators/map/tests/test_ts.py
QRemy: `selection_optional=["stat_scan"]` is added independently in the second case to be sure it returns the same.
I could add a test on the scan output.

AtreyeeS (approved):
Thanks @QRemy !

Issue #5379, opened 2024-07-18
TemplateSpatialModel.__init__ does not check for nan values in input map


lmohrmann:
**Gammapy version**
1.2

**Bug description**
When initializing a `TemplateSpatialModel`, there is no check for `nan` values in the input map. In case `normalize=True` (the default), this will result in a map attached to the model that is `nan` in all pixels, as the normalization involves a sum over all pixels, which is `nan` if there is a single pixel with `nan` entry.

**Expected behavior**
I see two solutions here: either a warning should be issued if any of the pixels in the input map has a `nan` entry (such a warning already exists e.g. for negative pixels), or pixels with `nan` entries should be ignored in the normalization (i.e. use `np.nansum` [here](https://github.com/gammapy/gammapy/blob/ba580c7926a7e5bd44e2933323f11d8b370660d6/gammapy/modeling/models/spatial.py#L1321) and [here](https://github.com/gammapy/gammapy/blob/ba580c7926a7e5bd44e2933323f11d8b370660d6/gammapy/modeling/models/spatial.py#L1324)). I think I would prefer the second option (at least I see nothing wrong per se with `nan` values in the map).

I'm happy to help with fixing this, but would like to hear some opinions about the preferred solution first.

AtreyeeS:
Thanks @lmohrmann ! 
I think we need to issue a `warning` for `nan` pixels - what is the expected counts for a `nan` value? Should those bins be masked during the likelihood computation? 
Maybe it is better to leave it to the user to use a `np_to_num` and clean up the template and configure appropriately the `mask_fit`.

PR #5441, opened 2024-09-18
Add a warning in TemplateSpatialModel for presence of nan values

AtreyeeS:
Presence of any nan value in a Template map silently creates all nan values in the normalised template as reported in #5379 This becomes very difficult to debug.

This PR adds a warning similar to some other invalid inputs.
It remains the users responsibiltity to understand how to interpret and fix the nans


registerrier (approved):
Thanks @AtreyeeS . Looks good to me!

QRemy (approved):
Thanks @AtreyeeS, no comment from my side.

PR #5443, opened 2024-09-20
Introduction of the FitResults container class for lists of FitResult objects and axes

cgalelli:
This draft PR introduces the FitResults class as a way to store and access arrays of FitResult objects, with an optional correlated axis.
This could be useful in the case of iterative fitting procedures for example pr #5427 
The main issue is serialization.

### DIFF
--- a/gammapy/modeling/fit.py
+++ b/gammapy/modeling/fit.py
@@ -3,6 +3,8 @@ import html
 import itertools
 import logging
 import collections.abc
+from abc import ABC
+
 import numpy as np
 from astropy.table import Table, QTable
 from gammapy.utils.pbar import progress_bar
@@ -899,7 +901,7 @@ class FitResult:
             return f"<pre>{html.escape(str(self))}</pre>"
 
 
-class FitResults(collections.abc.MutableSequence):
+class FitResults(collections.abc.MutableSequence, ABC):
     def __init__(self, results, axis_name=None):
         if np.array([not isinstance(result, FitResult) for result in results]).any():
             raise TypeError(f"Elements in {results!r} are not FitResult objects")
--- a/gammapy/modeling/fit.py
+++ b/gammapy/modeling/fit.py
@@ -901,12 +901,10 @@ class FitResult:
 
 
 class FitResults(collections.abc.Sequence):
-    def __init__(self, results, axis=None):
-        if axis and axis.nbin != len(results):
-            raise ValueError("Axis length does not correspond to number of results")
-
+    def __init__(self, results):
+        if np.array([not isinstance(result, FitResult) for result in results]).any():
+            raise TypeError(f"Elements in {results!r} are not FitResult objects")
         self.results = results
-        self.axis = axis
 
     def __add__(self, other):
         if isinstance(other, FitResult):
@@ -935,19 +933,21 @@ class FitResults(collections.abc.Sequence):
     def optimize_results(self):
         return [f.optimize_result for f in self.results]
 
-    def select_by_interval(self, coord_min, coord_max):
-        if self.axis is None:
-            raise ValueError("No axis to convert coordinate to index!")
+    def select_by_interval(self, axis, coord_min, coord_max):
+        if axis and axis.nbin != len(self.results):
+            raise ValueError("Axis length does not correspond to number of results")
 
-        keymin = self.axis.coord_to_idx(coord_min)
-        keymax = self.axis.coord_to_idx(coord_max)
+        keymin = axis.coord_to_idx(coord_min)
+        keymax = axis.coord_to_idx(coord_max)
 
         return self[keymin:keymax]
 
     def write_models(self, path, **kwargs):
         self.models().write(path, **kwargs)
 
-    def create_model_table(self):
+    def create_model_table(self, axis=None):
+        if axis and axis.nbin != len(self.results):
+            raise ValueError("Axis length does not correspond to number of results")
         t = QTable()
 
         t["convergence"] = [result.success for result in self.results]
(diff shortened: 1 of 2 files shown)


gammapy/modeling/fit.py
```diff
@@ -895,3 +899,82 @@ def _repr_html_(self):
...
+            return self.__class__(self.results[key])
+
+    def __len__(self):
+        return len(self.results)
+
+    def success(self):
+        return [f.success for f in self.results]
+
+    def models(self):
+        return Models([model for f in self.results for model in f.models])
```
AtreyeeS, line 931: each FitResult can have multiple models. Is it safe to put all of them in a single Models object?  It is not guaranteed that all the fitresult.models will have the same number of models either

gammapy/modeling/fit.py
```diff
@@ -895,3 +899,82 @@ def _repr_html_(self):
...
+            col_data = []
+
+            for name in self.models()[i].parameters.free_parameters.names:
+                col_data.append(self.models()[i].parameters[name].value)
+                col_data.append(self.models()[i].parameters[name].error)
+
+            t.add_row(col_data)
+
+        if isinstance(self.axis, TimeMapAxis):
+            t.add_columns(self.axis.to_gti().table.columns, indexes=[0, 0])
```
AtreyeeS, line 976: Maybe we should have a `TimeMapAxis.to_table` directly

AtreyeeS (commented):
Thank you @cgalelli !
Looks very useful.

Do we need to ensure that this works only if all the `fitresults` have the same set of Models?

gammapy/modeling/fit.py
```diff
@@ -895,3 +899,82 @@ def _repr_html_(self):
             return self.to_html()
         except AttributeError:
             return f"<pre>{html.escape(str(self))}</pre>"
+
+
+class FitResults(collections.abc.Sequence):
+    def __init__(self, results, axis=None):
+        if axis and axis.nbin != len(results):
```
AtreyeeS, line 906: I dont know if we should have an axis here

gammapy/modeling/fit.py
cgalelli: You are completely right, I was thinking mostly of a possible output from an estimator (which we can suppose to have the same output model "shape") but the FitResults object could be used for many other cases... I think an option ("flatten"-like) + logging if the models don't have the same length.

gammapy/modeling/fit.py
cgalelli: Well yes, this way is quite convoluted...

gammapy/modeling/fit.py
cgalelli: Maybe instead of an attribute of the class, it could be an input in relevant methods? (e.g. table creation)

gammapy/modeling/fit.py
```diff
@@ -895,3 +897,69 @@ def _repr_html_(self):
...
+
+class FitResults(collections.abc.Sequence):
+    def __init__(self, results):
+        if np.array([not isinstance(result, FitResult) for result in results]).any():
+            raise TypeError(f"Elements in {results!r} are not FitResult objects")
+        self.results = results
+
+    def __add__(self, other):
+        if isinstance(other, FitResult):
+            return FitResult([*self, other])
```
AtreyeeS, line 910: should be `FitResults`, no?

gammapy/modeling/fit.py
```diff
@@ -895,3 +897,69 @@ def _repr_html_(self):
             return self.to_html()
         except AttributeError:
             return f"<pre>{html.escape(str(self))}</pre>"
+
+
+class FitResults(collections.abc.Sequence):
```
AtreyeeS, line 902: if you use `collections.abc.MutableSequence` instead, then `append` etc should work automatically, no

gammapy/modeling/fit.py
```diff
@@ -895,3 +897,69 @@ def _repr_html_(self):
...
+        return Models([model for f in self.results for model in f.models])
+
+    def covariance_results(self):
+        return [f.covariance_result for f in self.results]
+
+    def optimize_results(self):
+        return [f.optimize_result for f in self.results]
+
+    def select_by_interval(self, axis, coord_min, coord_max):
+        if axis and axis.nbin != len(self.results):
```
AtreyeeS, line 936: simply passing the `axis` name can also work

gammapy/modeling/fit.py
```diff
@@ -895,3 +897,69 @@ def _repr_html_(self):
...
+
+        keymin = axis.coord_to_idx(coord_min)
+        keymax = axis.coord_to_idx(coord_max)
+
+        return self[keymin:keymax]
+
+    def write_models(self, path, **kwargs):
+        self.models().write(path, **kwargs)
+
+    def create_model_table(self, axis=None):
```
AtreyeeS, line 947: maybe `col_names` can be an argument with some defaults?

gammapy/modeling/fit.py
cgalelli: Oops, stupid mistake!

gammapy/modeling/fit.py
cgalelli: Wouldn't you need the axis itself to slice?

AtreyeeS: Its not guaranteed that the axis correctly corresponds this fit result. 
I dont know if its safe to allow a slicing like this

cgalelli: Ok, so you mean adding the axis name as a parameter of the object? (self.axis_name) And check that the axis name corresponds in case of slicing?

QRemy:
@cgalelli will you have time to work on this or someone should takeover ?

cgalelli:
Sorry @QRemy (and everyone),  I don't foresee an opening to work on this, at least for this year.

QRemy:
Thanks @cgalelli, I'll close this PR for now.

PR #5444, opened 2024-09-20
ResolvedEstimator to fit models on datasets in bins defined by an axis.

cgalelli:
This draft pr is a generalization of #5427 into a class that can do the same type of fitting of Models onto a list of datasets, selected by intervals defined by an axis (energy, time, label...).
The actual class is a proposal, and probably missing some features and options (especially related to morphology).
This also could benefit from the introduction of classes for the output such as the one I propose in #5443

### DIFF
--- a/gammapy/estimators/__init__.py
+++ b/gammapy/estimators/__init__.py
@@ -1,5 +1,6 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 """Estimators."""
+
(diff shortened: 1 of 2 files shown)


QRemy:
@cgalelli will you have time to work on this or someone should takeover ?

cgalelli:
As for #5443  I don't think I will have time to work on this for 2026, sorry!

PR #5445, opened 2024-09-24
Add a to_table() method to TimeMapAxis

cgalelli:
This simple PR adds a `to_table` method for TimeMapAxis that produces the same table as the one produced by the equivalent GTIs.
Mentioned by @AtreyeeS as a simple possibility to spare a couple of function calls in PR #5443 

Probably needs to be completed with some format and validity checks.


gammapy/maps/axes.py
```diff
@@ -3030,6 +3030,25 @@ def to_gti(self):
...
+
+        Parameters
+        ----------
+        format : {"gadf"}
+            Format specification. Default is "gadf".
+
+        Returns
+        -------
+        table : `~astropy.table.Table`
+            Table with axis data.
```
bkhelifi, line 3044: The table contains astopy.Time object only. Is not it?

gammapy/maps/axes.py
```diff
@@ -3030,6 +3030,25 @@ def to_gti(self):
             self.edges_min, self.edges_max, reference_time=self.reference_time
         )
 
+    def to_table(self, format="gadf"):
+        """Create table.
+
+        Parameters
+        ----------
+        format : {"gadf"}
+            Format specification. Default is "gadf".
```
registerrier, line 3039: Is there a specification for such a table in GADF? Usually time quantities are stored with a reference value stored as MJDREFF in the header and start and stops are just time like quantities.

registerrier (commented):
Thanks @cgalelli . In general, it is good to uniformize the API between regular `MapAxis` and `TimeMapAxis` as the latter should be subclasses of the former.

Practically, it is not clear to me that serialization method should live on basic building blocks. But that's probably another question.

gammapy/maps/axes.py
AtreyeeS, line 3039: We can also use `self.to_gti().to_table()` to ensure a similar serialisation

gammapy/maps/axes.py
cgalelli: Yes the table would contain only times, at least in the basic format

gammapy/maps/axes.py
cgalelli: You are right, the specification would be to give the time as a quantity, like doing `to_gti().table`. That is probably the best solution.

AtreyeeS (commented):
Thanks @cgalelli ! Can you add a test please

cgalelli:
Thanks, @AtreyeeS , tests are also missing for `MapAxis.to_table()`. Should they be added as well? And how should they look? Maybe I can make a dedicated PR

AtreyeeS:
>Tests are also missing for `MapAxis.to_table()`. Should they be added as well? And how should they look? Maybe I can make a dedicated PR

Yes should definitely be added! I think you can simply assert that the table column name matches the one of the MapAxis, and that the table values correspond to the MapAxis nodes

gammapy/maps/axes.py
```diff
@@ -3030,6 +3030,23 @@ def to_gti(self):
             self.edges_min, self.edges_max, reference_time=self.reference_time
         )
 
+    def to_table(self, format="gadf"):
+        """Create table.
```
AtreyeeS, line 3034: The format argument is unused

AtreyeeS (commented):
Please remove the  `format` argument. Good from my side

gammapy/maps/axes.py
bkhelifi: Should we precise it explicitly? An open question...

bkhelifi (approved):
Thanks a lot!

AtreyeeS (approved):
Thanks @cgalelli ! Merging

PR #5446, opened 2024-09-25
Adapt docstring for `TemplateSpectralModel`, tutorial and tests

Astro-Kirsty:
[In this tutorial](https://docs.gammapy.org/dev/tutorials/data/fermi_lat.html#isotropic-diffuse-background), we are using 'fill_value': 'extrapolate' to extrapolate the model above 500 GeV . But actually we have 'fill_value':'None', and this is not the correct implementation.

The `kwargs` are the input to `TemplateSpectralModel`, which are then passed to `gammapy.utils.interpolation.ScaledRegularGridInterpolator`. 
I have updated the docstring, tutorial and added tests.

### DIFF
--- a/gammapy/modeling/models/spectral.py
+++ b/gammapy/modeling/models/spectral.py
@@ -1758,7 +1758,7 @@ class TemplateSpectralModel(SpectralModel):
         Array with the values of the model at energies ``energy``.
     interp_kwargs : dict
         Interpolation option passed to `~gammapy.utils.interpolation.ScaledRegularGridInterpolator`.
-        By default, all values outside the interpolation range are set to zero.
+        By default, all values outside the interpolation range are set to NaN.
         If you want to apply linear extrapolation you can pass `interp_kwargs={'extrapolate':
         True, 'method': 'linear'}`. If you want to choose the interpolation
         scaling applied to values, you can use `interp_kwargs={"values_scale": "log"}`.
--- a/gammapy/modeling/models/spectral.py
+++ b/gammapy/modeling/models/spectral.py
@@ -1,5 +1,6 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 """Spectral models for Gammapy."""
+
 import logging
 import operator
 import os
@@ -1397,7 +1398,6 @@ class ExpCutoffPowerLawNormSpectralModel(SpectralModel):
     def __init__(
         self, index=None, norm=None, reference=None, lambda_=None, alpha=None, **kwargs
     ):
-
         if index is None:
             warnings.warn(
                 "The default index value changed from 1.5 to 0 since v1.2",
@@ -1708,7 +1708,6 @@ class LogParabolaNormSpectralModel(SpectralModel):
     beta = Parameter("beta", 0)
 
     def __init__(self, norm=None, reference=None, alpha=None, beta=None, **kwargs):
-
         if alpha is None:
             warnings.warn(
                 "The default alpha value changed from 2 to 0 since v1.2",
@@ -1758,10 +1757,10 @@ class TemplateSpectralModel(SpectralModel):
     values : `~numpy.ndarray`
         Array with the values of the model at energies ``energy``.
     interp_kwargs : dict
-        Interpolation keyword arguments passed to `scipy.interpolate.RegularGridInterpolator`.
-        By default all values outside the interpolation range are set to zero.
-        If you want to apply linear extrapolation you can pass `interp_kwargs={'fill_value':
-        'extrapolate', 'kind': 'linear'}`. If you want to choose the interpolation
+        Interpolation option passed to `~gammapy.utils.interpolation.ScaledRegularGridInterpolator`.
+        By default, all values outside the interpolation range are set to zero.
+        If you want to apply linear extrapolation you can pass `interp_kwargs={'extrapolate':
+        True, 'method': 'linear'}`. If you want to choose the interpolation
         scaling applied to values, you can use `interp_kwargs={"values_scale": "log"}`.
     meta : dict, optional
         Meta information, meta['filename'] will be used for serialisation.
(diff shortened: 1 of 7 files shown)


gammapy/modeling/models/tests/test_cube.py
```diff
@@ -659,6 +659,13 @@ def test_fermi_isotropic():
     assert flux.unit == "MeV-1 cm-2 s-1 sr-1"
     assert isinstance(model.spectral_model, CompoundSpectralModel)
 
+    model_options = create_fermi_isotropic_diffuse_model(
+        filename=filename,
+        interp_kwargs={"extrapolate": True, "method": "nearest"},
+    )
+    flux_options = model_options(**coords)
+    assert_allclose(flux_options.value, 1.456e-13, rtol=1e-3)
```
bkhelifi, line 667: Can we test also an other extrapolation algorithm? If something changes in astropy, we will be then informed...

bkhelifi:
I am then so surprised that the Fermi tutorial was showing data points up to 2TeV. Can you explain me why something was happening?

Astro-Kirsty:
> I am then so surprised that the Fermi tutorial was showing data points up to 2TeV. Can you explain me why something was happening?

The default is actually `True`, but I decided to explicitly write it here to avoid confusion.

gammapy/modeling/models/tests/test_cube.py
AtreyeeS: Is 50 GeV outside the interpolation range? It would be good to test with an arrays of energies covering values inside and outside the interpolation range, such that the `interp_kwargs` make a difference

Astro-Kirsty: It doesn't seem possible to have extrapolate False with this specific example, even if I take the energy directly from the file:

```
vals = np.loadtxt(make_path(filename))
energy = u.Quantity(vals[:, 0], "MeV")
coords_inner = {"lon": 0 * u.deg, "lat": 0 * u.deg, "energy": energy[7]}
flux = model(**coords_inner)
```
[interp_kwargs](https://github.com/user-attachments/assets/fa6a09a8-320f-44b0-b00c-a91e7d4bf5e8)

Astro-Kirsty: Sorry the error actually comes from the 
`model = create_fermi_isotropic_diffuse_model(filename=filename,  interp_kwargs={"extrapolate": False})`

bkhelifi (approved):
Thanks a lot @Astro-Kirsty. All good for me ;)

gammapy/modeling/models/tests/test_cube.py
```diff
@@ -650,15 +650,37 @@ def test_sky_point_source():
...
+        filename=filename,
+        interp_kwargs={"extrapolate": False, "bounds_error": False},
+    )
+    coords_inner = {"lon": 0 * u.deg, "lat": 0 * u.deg, "energy": energy[7]}
+    flux = model(**coords_inner)
+    assert_allclose(flux.value, values[7].value, rtol=1e-3)
     assert flux.unit == "MeV-1 cm-2 s-1 sr-1"
     assert isinstance(model.spectral_model, CompoundSpectralModel)
 
+    # No extrapolation with bounds_error
```
AtreyeeS, line 668: Is this test necessary?

gammapy/modeling/models/tests/test_cube.py
```diff
@@ -650,15 +650,37 @@ def test_sky_point_source():
...
-
-    assert_allclose(flux.value, 1.463e-13, rtol=1e-3)
+    vals = np.loadtxt(make_path(filename))
+    energy = u.Quantity(vals[:, 0], "MeV")
+    values = u.Quantity(vals[:, 1], "MeV-1 s-1 cm-2")
+
+    # No extrapolation
+    model = create_fermi_isotropic_diffuse_model(
+        filename=filename,
+        interp_kwargs={"extrapolate": False, "bounds_error": False},
```
AtreyeeS, line 660: The default option for `bounds_error` should not fail. I think that is the behaviour you implemented, so maybe just remove the bounds_error option in this PR

AtreyeeS (commented):
Thanks @Astro-Kirsty ! As a clean up and to improve readibility, you can set 

```
energy = [0.01, 1, 10, 100, 1000]* u.GeV
coords =  {"lon": 0 * u.deg, "lat": 0 * u.deg, "energy": energy}
model__noextrapolate = 
model_extrapolate= ..
assert_allclose(model__noextrapolate(**coords), [0, 0 ... ] )
```

and so on

ie, compare the computed values at the same energies with and without extrapolation

gammapy/modeling/models/tests/test_cube.py
Astro-Kirsty: I want to ensure a good coverage of options because we were missing this issue previously!

bkhelifi (dismissed):
Great! Thanks

gammapy/modeling/models/spectral.py
```diff
@@ -1758,10 +1757,10 @@ class TemplateSpectralModel(SpectralModel):
     values : `~numpy.ndarray`
         Array with the values of the model at energies ``energy``.
     interp_kwargs : dict
-        Interpolation keyword arguments passed to `scipy.interpolate.RegularGridInterpolator`.
-        By default all values outside the interpolation range are set to zero.
-        If you want to apply linear extrapolation you can pass `interp_kwargs={'fill_value':
-        'extrapolate', 'kind': 'linear'}`. If you want to choose the interpolation
+        Interpolation option passed to `~gammapy.utils.interpolation.ScaledRegularGridInterpolator`.
+        By default, all values outside the interpolation range are set to zero.
```
Astro-Kirsty, line 1761:
```suggestion
        By default, all values outside the interpolation range are set to NaN.
```

AtreyeeS (approved):
No more comments!

bkhelifi (approved):
Thanks @Astro-Kirsty

PR #5449, opened 2024-09-26
combine_flux_maps  supports for different energy axes

QRemy:
Minor change such as `combine_flux_maps` can support maps  with different energy axes.
Works only if the first map energy range encompass the one of the other maps.


bkhelifi (approved):
Thanks a lot @QRemy... a simple fix

Issue #5452, opened 2024-09-28
Wrong time scale for events sampled with `MapDatasetEventSampler` 


fabiopintore:
**Gammapy version**
Gammapy 1.2

**Bug description**
There is a possible bug in the time of the events sampled with `MapDatasetEventSampler`. 
It may happen only with energy-dependent temporal models: events are sampled correctly but their time could be referred to a wrong time scale (i.e. either `tt` or `utc` ). The bug is in `_sample_coord_time_energy`.
Source photons can have a lag with respect to the real time start of the observation (and also with the bkg photons time as well). This effect is quite evident for very short observations as those of the GW/GRB of the SDC...

**Expected behavior**
Event times have to be all inside the time start and time stop.

**To Reproduce**
Sample a sky-model with an energy-dependent temporal model plus background. 

The bug was spotted from the analysis of data from the CTA SDC.

PR #5453, opened 2024-09-28
Fix time shift from sampled events with `MapDatasetEventSampler`

fabiopintore:
This PR solves the issue #5452 .
The times of sampled events from energy-dependent temporal models were affected by a time shift due to a mismatch between the time scales when applying `MapDatasetEventSampler._sample_coord_time_energy`. 
The default `scale` for the times of the sampled events is `tt`. However, the times may be in a different `scale` if the `dataset.gti` times are in a different `scale`.

### DIFF
--- a/gammapy/datasets/simulate.py
+++ b/gammapy/datasets/simulate.py
@@ -206,7 +206,9 @@ class MapDatasetEventSampler:
 
         coords = npred.sample_coord(n_events=n_events, random_state=self.random_state)
 
-        coords["time"] = Time(coords["time"], format="mjd", scale="tt")
+        coords["time"] = Time(
+            coords["time"], format="mjd", scale=dataset.gti.time_ref.scale
+        )
 
         table = self._make_table(coords, dataset.gti.time_ref)
 
--- a/gammapy/datasets/tests/test_simulate.py
+++ b/gammapy/datasets/tests/test_simulate.py
@@ -308,7 +308,7 @@ def test_sample_coord_time_energy_random_seed(
 
     assert_allclose(
         [events[0][0], events[0][1], events[0][2], events[0][3]],
-        [0.2998, 1.932196, 266.404988, -28.936178],
+        [0.29982, 1.932196, 266.404988, -28.936178],
         rtol=1e-3,
     )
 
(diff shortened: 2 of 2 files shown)


gammapy/datasets/tests/test_simulate.py
```diff
@@ -366,6 +366,7 @@ def test_mde_sample_sources_psf_update(dataset, models):
 def test_sample_sources_energy_dependent(dataset, energy_dependent_temporal_sky_model):
     dataset.models = energy_dependent_temporal_sky_model
 
+    print(dataset.gti.time_start.scale)
```
registerrier, line 369: Please remove print statement

gammapy/datasets/simulate.py
```diff
@@ -206,7 +206,9 @@ def _sample_coord_time_energy(self, dataset, model):
 
         coords = npred.sample_coord(n_events=n_events, random_state=self.random_state)
 
-        coords["time"] = Time(coords["time"], format="mjd", scale="tt")
+        coords["time"] = Time(
+            coords["time"], format="mjd", scale=dataset.gti.time_ref.scale
```
registerrier, line 210: You could also use the `time_ref` on the `npred` time axis no? This should be the closest to the time scales of events sampled from `npred`.

registerrier (commented):
Thanks @fabiopintore ! I have left a few inline comments.

gammapy/datasets/simulate.py
fabiopintore: thanks @registerrier . Just to understand, how do you access to the `time_ref` of the `npred`? The `time` axis of `npred` seems to have only a list of times without any specification on `scales` and `format`..

registerrier: There should be a `reference_time` attribute in `TimeMapAxis`:
https://docs.gammapy.org/dev/api/gammapy.maps.TimeMapAxis.html#gammapy.maps.TimeMapAxis.reference_time

So accessing it via `npred.geom.axes["time"].reference_time` should be possible, no?

fabiopintore: Actually, it does not have such an attribute. Because `npred` (for energy-dependent temporal models) has a time axis created by `MapAxis` and not `TimeMapAxis`.

gammapy/datasets/tests/test_simulate.py
```diff
@@ -378,6 +379,9 @@ def test_sample_sources_energy_dependent(dataset, energy_dependent_temporal_sky_
 
     assert_allclose(events.table["TIME"][0], 95.464699, rtol=1e-5)
 
+    dt = np.max(events.table["TIME"]) - np.min(events.table["TIME"])
+    assert dt <= dataset.gti.time_sum.to("s").value + sampler.t_delta.to("s").value
+
```
AtreyeeS, line 384: maybe also assert that the time scales match for the gti and the events

registerrier (approved):
Thanks @fabiopintore ! Looks good.

PR #5407, opened 2024-07-30
Add an API tutorial for Estimators

Astro-Kirsty:
This PR is to revive the [notebook](https://github.com/gammapy/gammapy-meetings/blob/master/user-meetings/2022/2022-01-20/estimators.ipynb) from @adonath 

A copy of the notebook can be found at this [gist](https://gist.github.com/Astro-Kirsty/4402da2315bf4e23a42f59c951c30556)

PR #5454, opened 2024-10-01
Add SED to the glossary

Astro-Kirsty:
Adding SED to the gammapy glossary as per https://github.com/gammapy/gammapy/pull/5407#discussion_r1780901365

### DIFF
--- a/docs/user-guide/references.rst
+++ b/docs/user-guide/references.rst
@@ -134,8 +134,8 @@ Glossary
 
     SED
       Short for "spectral energy distribution". For a spectral model or flux points
-      object, the type of plot (e.g. dN/dE, E^2 dN/dE) is typically adjusted through
-      the `sed_type` quantity.
+      object, the type of plot (e.g. :math:`dN/dE`, :math:`E^2\ dN/dE`) is typically adjusted
+      through the `sed_type` quantity. See :ref:`sedtypes` for a list of options.
 
     Stacked Analysis
       In a stacked analysis individual observations are reduced to datasets which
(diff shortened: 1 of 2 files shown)


docs/user-guide/references.rst
```diff
@@ -132,6 +132,10 @@ Glossary
       data are analyzed. In practice, at each energy it corresponds with the sky region
       in which the dataset mask is True.
 
+    SED
+      Short for "spectral energy distribution". For a spectral model or flux points object
```
bkhelifi, line 136:
```suggestion
      Short for "spectral energy distribution". For a spectral model or flux points object, 
```

Astro-Kirsty:
Are you happy with how this looks in the documentation?
image

bkhelifi (approved):
Thanks a lot

docs/user-guide/references.rst
```diff
@@ -132,6 +132,11 @@ Glossary
       data are analyzed. In practice, at each energy it corresponds with the sky region
       in which the dataset mask is True.
 
+    SED
+      Short for "spectral energy distribution". For a spectral model or flux points
+      object, the type of plot (e.g. dN/dE, E^2 dN/dE) is typically adjusted through
```
registerrier, line 137:
```suggestion
      object, the type of plot (e.g. :math:`dN/dE`, :math:`E^2\ dN/dE`) is typically adjusted through
```
using inline math type should look better.

registerrier (approved):
Thanks @Astro-Kirsty . Looks good see inline comment.

Note that we good point to a table listing the possible `sed_type` values. I think this the closest we have to that:
https://docs.gammapy.org/1.2/user-guide/estimators.html#flux-quantities

AtreyeeS (approved):
Thanks @Astro-Kirsty !

bkhelifi (approved):
Thanks a lot!

PR #5455, opened 2024-10-01
Fix formatting of HAWC tutorial

Astro-Kirsty:
Some fixes in the formatting of the HAWC tutorial I noticed

### DIFF
--- a/examples/tutorials/data/hawc.py
+++ b/examples/tutorials/data/hawc.py
@@ -17,7 +17,7 @@ This dataset is 1 GB in size, so only a subset will be used here as an example.
 
 This notebook is a quick introduction to HAWC data analysis with Gammapy.
 It briefly describes the HAWC data and how to access it, and then uses a
-subset of it to produce a MapDataset, to show how the data reduction is performed.
+subset of it to produce a `~gammapy.datasets.MapDataset`, to show how the data reduction is performed.
 
 The HAWC data release contains events where energy is estimated using
 two different algorithms, referred to as "NN" and "GP" (see this `paper <https://iopscience.iop.org/article/10.3847/1538-4357/ab2f7d>`__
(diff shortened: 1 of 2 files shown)


bkhelifi:
Thanks @Astro-Kirsty .
@QRemy and @LauraOlivera, should we add the tricks that we discussed by email/slack into this tutorial with a pedagogical explanation? If yes, in this PR or into an other one?

examples/tutorials/data/hawc.py
```diff
@@ -80,21 +90,24 @@
...
 obs_filename = f"obs-index-table-{energy_estimator}-Crab.fits.gz"
 
-# there is only one observation table
+######################################################################
+# There is only one observation table
 obs_table = ObservationTable.read(data_path + obs_filename)
 
-# we read the hdu index table of fHit bin number 6
+######################################################################
+# Now, we read the HDU index table of fHit bin number 6
```
registerrier, line 104: Maybe here specify that the remainder of the tutorial is made using only one single `fHit` value, but that a regular analysis would combine them all. (which is suggested in the exercices.

examples/tutorials/data/hawc.py
```diff
@@ -1,7 +1,12 @@
 """
 HAWC with Gammapy
-=====================
-Explore HAWC event lists and IRFs and perform the data reduction steps.
+=================
+
+Explore HAWC event lists and instrument response functions (IRFs), then perform the
+data reduction steps.
+
+Introduction
```
registerrier, line 8: Note that you might add the `Introduction ` section in the HESS tutorial as well as it lacks it.

registerrier (approved):
Thanks @Astro-Kirsty . I have left a few inline comments.

cgalelli (approved):
Thanks @Astro-Kirsty !

bkhelifi (approved):
Thanks @Astro-Kirsty.
If @LauraOlivera or @QRemy has time to quickly improve it a bit for the v1.3 relative to performance issues (see the issue #5484 ), one could make it here. OR into an other tuto later... ;)

Astro-Kirsty:
> Thanks @Astro-Kirsty. If @LauraOlivera or @QRemy has time to quickly improve it a bit for the v1.3 relative to performance issues (see the issue #5484 ), one could make it here. OR into an other tuto later... ;)

I think we decided in the previous dev call that there is not much we can add to this tutorial, as it is just an introduction really. But rather that we would create another tutorial/recipe for that purpose.

bkhelifi:
> 
> I think we decided in the previous dev call that there is not much we can add to this tutorial, as it is just an introduction really. But rather that we would create another tutorial/recipe for that purpose.

Exact, I did forget! sorry and thanks

PR #5461, opened 2024-10-03
Small adaption to `plot_energy_dependence` docstring

Astro-Kirsty:
Commonly people try to use a `offset=0.25*u.deg` but actually we accept only a list  `offset=[0.25*u.deg]`. This could be do with the incorrect docstring


AtreyeeS (approved):
Thanks @Astro-Kirsty !

bkhelifi (approved):

PR #5463, opened 2024-10-04
Add information in guide to add links to documentation

Astro-Kirsty:
We already show how to link to tutorials, but we can also link to other pieces of documentation with the following lines. 
I think this is very useful to show as it saves times trying to google how to do this everytime!

image


bkhelifi (approved):
Perfect, Thanks

PR #5468, opened 2024-10-04
Multi-dimension prior

Astro-Kirsty:
This is an update of #4907

I decided it would be best to keep the original work there, as we have forseen a number of important changes. 

This is a draft as it is still being worked on!


registerrier (commented):
Thanks @Astro-Kirsty . Please rebase since changes on `stat_sum` have already been introduced in the context of the nested sampling. See #5662 .

registerrier:
This is superseded by #5838 . Closing for now.

PR #5474, opened 2024-10-08
Fixing Matomo inclusion

bkhelifi:
**Description**
This PR adds the use of templates by Sphinx and add the Matomo banner in the main readme.


bkhelifi:
Merging such that Pei Yu can make tests

Issue #5233, opened 2024-04-22
Fix the sherpa installation


bkhelifi:
**Gammapy version**
gammapy 1.2, under Linux 

**Bug description**
The installation of the environment with mamba/conda fails because of sherpa. This has been reported by several users.

```
      checking dependency style of gcc... gcc3
      checking for bison... no
      checking for byacc... no
      checking for yacc... no
      configure: error: bison/byacc/yacc not found.
                Please install bison
      configure: error: ./configure failed for region-4.14
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
  ERROR: Failed building wheel for sherpa

```

Even afterwards after an installation without sherpa, the command `pip install sherpa==4.16` fails.

registerrier:
One option here is to remove the version constraint on sherpa, or use 4.15 instead (in which case conda should work).

Astro-Kirsty:
Do we know what versions of Linux this happens for?
Just curious, because I did not encounter this with Ubuntu 22.04

bkhelifi:
I have the same ubuntu version.
Within  the env of the version 1.2, `pip install sherpa==4.15` leads to the same error... 
I have in the output "configure: error: bison/byacc/yacc not found" for the configuration of region-4.14

vmharvey:
I have the same issue on Ubuntu 22.04.4. I found a bug on the sherpa issue tracker which shows the same problem occurring: https://github.com/sherpa/sherpa/issues/828. From this comment https://github.com/sherpa/sherpa/issues/828#issuecomment-644990512, assuming it's still true after a few years, it appears that installing sherpa via pip triggers a source install, which requires the user to have all the sherpa dependencies, including the build system bison. 

Modifying the env file by removing `sherpa==4.16` from the `pip` section and putting `sherpa=4.15` in the conda section fixed the problem for me:
```
  - tqdm
  - sherpa=4.15
  - pip:
    - ray[default]==2.9
```

It appears that the official sherpa channel on anaconda only has version 4.15, hence the small downgrade required.

Astro-Kirsty:
Sherpa was initially moved to the pip installation due to both apple and windows systems not being able to install it through this method. 
https://github.com/gammapy/gammapy/pull/4983#issuecomment-1854379008
As a user of these systems sherpa cannot be installed outside pip, so this line of the environment file needs to be removed. 

Not sure how we wish to proceed here? 
Appears to be some issue with any combination with at least one operating system :(

[...]

PR #5479, opened 2024-10-10
Solve issues with sherpa 4.17 and pin version for doc build

registerrier:
**Description**

This pull request solves issues caused by the sherpa v4.17, solving an issue with the sherpa backend that must have appeared with the numpy v2.0. For some reason, the all dependencies environment in the CI never installed sherpa (at least since we moved to numpy>=2) and silently skipped the test on sherpa. 

- The default `np.nan` for parameter min and maxes are causing issues with sherpa. They are now changed to `-np.inf` and `np.inf`.
- sherpa >4.16 is imposed in environment-dev.yml to make sure we use numpy>=2
- a restriction in the docs requirement is introduced in setup.cfg (sherpa<4.16). This is to keep a working environment with numpy <2 . We are still stuck to astropy<=5.3 for docs build. This should be solved but it will require an independent PR.

**Dear reviewer**

### DIFF
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -39,7 +39,7 @@ dependencies:
   - requests
   - tqdm
   # extra dependencies
-  - sherpa=4.16
+  - sherpa
   - healpy
   - ipython
   - jupyter
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -9,7 +9,7 @@ name: gammapy-dev
 
 channels:
   - conda-forge
-  - sherpa
+  - https://cxc.cfa.harvard.edu/conda/sherpa
 
 variables:
   PYTHONNOUSERSITE: "1"
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -39,6 +39,7 @@ dependencies:
   - requests
   - tqdm
   # extra dependencies
+  - sherpa<=4.15
   - healpy
   - ipython
   - jupyter
@@ -77,7 +78,6 @@ dependencies:
   - pyinstrument
   - memray
   - pip:
-      - sherpa
       - pytest-sphinx
       - ray[default]>=2.9
       - PyGithub
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -39,7 +39,7 @@ dependencies:
   - requests
   - tqdm
   # extra dependencies
-  - sherpa<=4.15
+  - sherpa=4.16
   - healpy
   - ipython
   - jupyter
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -39,7 +39,6 @@ dependencies:
   - requests
   - tqdm
   # extra dependencies
-  - sherpa
   - healpy
   - ipython
   - jupyter
@@ -78,6 +77,7 @@ dependencies:
   - pyinstrument
   - memray
   - pip:
+      - sherpa>=4.17
       - pytest-sphinx
       - ray[default]>=2.9
       - PyGithub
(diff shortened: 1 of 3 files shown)


setup.cfg
```diff
@@ -51,7 +51,7 @@ console_scripts =
 [options.extras_require]
 all =
     naima
-    sherpa; platform_system != "Windows"
+    sherpa<4.17; platform_system != "Windows"
```
bkhelifi, line 54:
```suggestion
    sherpa<4.16; platform_system != "Windows"
```

bkhelifi: if the title of your PR is correct...

bkhelifi:
For information, on Ubuntu, I have launch the command `conda install -c https://cxc.cfa.harvard.edu/conda/sherpa sherpa=4.15`.
It finds the library, which is an improvement.

BUT, it forces to downgrade numpy to 1.26!!!
```
The following packages will be DOWNGRADED:

  algopy                                 0.7.2-pyhd8ed1ab_0 --> 0.6.1-pyhd8ed1ab_0 
  numpy                               2.1.1-py311h71ddf71_0 --> 1.26.4-py311h64a7726_0 

```

And this is in conflict with our environement that requires `numpy>1.20`.

It think that this PR does not solve yet #5233

bkhelifi:
the same issue with sherpa v4.16. It also requires numpy 1.26!

environment-dev.yml
```diff
@@ -77,7 +77,7 @@ dependencies:
   - pyinstrument
   - memray
   - pip:
-      - sherpa
+      - sherpa>=4.17
```
MRegeard, line 80:
```suggestion
      - sherpa<=4.17
```

setup.cfg
```diff
@@ -51,15 +51,15 @@ console_scripts =
 [options.extras_require]
 all =
     naima
-    sherpa; platform_system != "Windows"
+    sherpa>=4.17; platform_system != "Windows"
```
MRegeard, line 54:
```suggestion
    sherpa<=4.17; platform_system != "Windows"
```

setup.cfg
```diff
@@ -51,15 +51,15 @@ console_scripts =
...
     healpy; platform_system != "Windows"
     requests
     tqdm
     ipywidgets
     ray[default]>=2.9
 all_no_ray =
     naima
-    sherpa; platform_system != "Windows"
+    sherpa>=4.17; platform_system != "Windows"
```
MRegeard, line 62:
```suggestion
    sherpa<=4.17; platform_system != "Windows"
```

MRegeard (changes requested):
Thanks @registerrier.
Putting `sherpa<=4.17` here does the job for me.
You can also put `sherpa<4.17` in the docs section.
This way you don't have conflicts and it will install 4.17 (numpy2) for the tests and 4.16 (numpy<2) for the docs.

MRegeard (approved):
@registerrier Thank you very much !

environment-dev.yml
bkhelifi: I am not sure about that... The dev should handle new versions! 
The same type of modifications in the other subsections have been merged, but I am not sure about it.

setup.cfg
```diff
@@ -80,6 +80,7 @@ test =
 docs =
     astropy>=5.0,<5.3
     numpy<2.0
+    sherpa<4.16
```
bkhelifi, line 83: See my comment above

bkhelifi (changes requested):
Thanks a lot @registerrier. You are solving a long-standing issue! I have just a remaining test

setup.cfg
registerrier: Thanks @bkhelifi . I have updated the motivation of the PR to explain why.

bkhelifi (approved):

PR #5488, opened 2024-10-11
Fix docs build error and correct further docstrings

Astro-Kirsty:
Fixes one of the build errors from a docstring plus other docstring I found along the way !


gammapy/estimators/map/core.py
```diff
@@ -1191,7 +1188,7 @@ def slice_by_idx(self, slices):
         )
 
     def slice_by_coord(self, slices):
-        """Slice flux maps by coordinate values
+        """Slice flux maps by coordinate valuesy.
```
Astro-Kirsty, line 1191:
```suggestion
        """Slice flux maps by coordinate values.
```

bkhelifi (approved):

PR #5490, opened 2024-10-11
Fix a couple of parameters links in the doc

bkhelifi:
A tiny PR....


gammapy/astro/darkmatter/utils.py
```diff
@@ -12,7 +13,7 @@ class JFactory:
 
     J-Factors are computed for annihilation and D-Factors for decay.
     Set the argument `annihilation` to `False` to compute D-Factors.
-    The assumed dark matter profiles will be centered on the center of the map.
+    The assumed dark matter profiles will be centered in the center of the map.
```
Astro-Kirsty, line 16: The first one is correct!

gammapy/data/event_list.py
```diff
@@ -528,7 +528,7 @@ def plot_offset2_distribution(
         >>> bins = np.linspace(start=0, stop=0.3 ** 2, num=30) * u.deg ** 2
         >>> events.plot_offset2_distribution(center=center, bins=bins) # doctest: +SKIP
 
-        Note how we passed the ``bins`` option of `matplotlib.pyplot.hist` to control
+        Note how we passed the ``bins`` option of `~matplotlib.pyplot.hist` to control
```
Astro-Kirsty, line 531: I would leave it as the first option, as it explicitly shows the entire link

gammapy/datasets/actors.py
```diff
@@ -19,15 +19,15 @@ class DatasetsActor(Datasets):
...
+    datasets : `~gammapy.datasets.Datasets`
         Datasets
     """
 
     def __init__(self, datasets=None):
         from ray import get
 
         log.warning(
-            "Gammapy support for parallelisation with ray is still a prototype and is not fully functional."
+            "Gammapy support for parallelization with ray is still a prototype and is not fully functional."
```
Astro-Kirsty, line 30:
```suggestion
            "Gammapy support for parallelisation with ray is still a prototype and is not fully functional."
```

gammapy/datasets/flux_points.py
```diff
@@ -71,7 +71,7 @@ class FluxPointsDataset(Dataset):
                     https://iopscience.iop.org/article/10.1088/0004-637X/773/2/168/pdf
 
         Default is `chi2`, in that case upper limits are ignored and the mean of asymetrics error is used.
-        However it is recommended to use `profile` if `stat_scan` is available on flux points.
+        However, it is recommended to use `profile` if `stat_scan` is available on flux points.
```
Astro-Kirsty, line 74:
```suggestion
        However, it is recommended to use ``profile`` if ``stat_scan`` is available on flux points.
```

gammapy/datasets/map.py
```diff
@@ -1709,15 +1709,15 @@ def info_dict(self, in_safe_data_range=True):
         return info
 
     def to_spectrum_dataset(self, on_region, containment_correction=False, name=None):
-        """Return a ~gammapy.datasets.SpectrumDataset from on_region.
+        """Return a `~gammapy.datasets.SpectrumDataset` from on_region.
```
Astro-Kirsty, line 1712:
```suggestion
        """Return a `~gammapy.datasets.SpectrumDataset` from ``on_region``.
```

gammapy/datasets/map.py
```diff
@@ -199,7 +199,7 @@ def create_map_dataset_from_observation(
     models : `~gammapy.modeling.Models`, optional
         Models. Default is None.
     dataset_name : str, optional
-        If `models` contains one or multiple `FoVBackgroundModel`
+        If `models` contains one or multiple `~gammapy.modeling.models.FoVBackgroundModel`
         it should match the `dataset_name` of the background model to use.
```
Astro-Kirsty, line 203:
```suggestion
        If ``models`` contains one or multiple `~gammapy.modeling.models.FoVBackgroundModel`
        it should match the ``dataset_name`` of the background model to use.
```

gammapy/datasets/map.py
```diff
@@ -2867,7 +2867,7 @@ def info_dict(self, in_safe_data_range=True):
         return info
 
     def to_spectrum_dataset(self, on_region, containment_correction=False, name=None):
-        """Return a ~gammapy.datasets.SpectrumDatasetOnOff from on_region.
+        """Return a `~gammapy.datasets.SpectrumDatasetOnOff` from on_region.
```
Astro-Kirsty, line 2870:
```suggestion
        """Return a `~gammapy.datasets.SpectrumDatasetOnOff` from ``on_region``.
```

gammapy/datasets/spectrum.py
```diff
@@ -334,7 +334,7 @@ def read(cls, filename, format="ogip", checksum=False, **kwargs):
     def write(self, filename, overwrite=False, format="ogip", checksum=False):
         """Write spectrum dataset on off to file.
 
-        Can be serialised either as a `MapDataset` with a `RegionGeom`
+        Can be serialised either as a `~gammapy.datasets.MapDataset` with a `RegionGeom`
```
Astro-Kirsty, line 337: Should RegionGeom also have a link?

gammapy/makers/reduce.py
```diff
@@ -34,7 +34,7 @@ class DatasetsMaker(Maker, parallel.ParallelMixin):
     cutout_width : tuple of `~astropy.coordinates.Angle`, optional
         Angular sizes of the region in (lon, lat) in that specific order.
         If only one value is passed, a square region is extracted.
-        If None it returns an error, except if the list of makers includes a `SafeMaskMaker`
+        If None it returns an error, except if the list of makers includes a `~gammapy.makers.SafeMaskMaker`
```
Astro-Kirsty, line 37:
```suggestion
        If None, an error is returned except if the list of makers includes a `~gammapy.makers.SafeMaskMaker`
```

gammapy/modeling/models/spectral.py
```diff
@@ -659,7 +659,7 @@ def f(x):
     def inverse_all(self, values, energy_min=0.1 * u.TeV, energy_max=100 * u.TeV):
         """Return energies for multiple function values of the spectral model.
 
-        Calls the `scipy.optimize.brentq` numerical root finding method.
+        Calls the `~scipy.optimize.brentq` numerical root finding method.
```
Astro-Kirsty, line 662: Again -- leave the full path (i.e. previous implementation)

gammapy/modeling/models/spectral.py
```diff
@@ -1206,9 +1206,9 @@ class PiecewiseNormSpectralModel(SpectralModel):
     ----------
     energy : `~astropy.units.Quantity`
         Array of energies at which the model values are given (nodes).
-    norms : `~numpy.ndarray` or list of `Parameter`
+    norms : `~numpy.ndarray` or list of `~gammapy.modeling.Parameter`
         Array with the initial norms of the model at energies ``energy``.
-        Normalisation parameters are created for each value.
+        Normalization parameters are created for each value.
```
Astro-Kirsty, line 1211: en-GB = 's'

Astro-Kirsty (commented):
Thanks for the huge effort @bkhelifi. 
Is it necessary to specify the full path if you are in the same file as that function/class?

gammapy/modeling/models/core.py
```diff
@@ -379,7 +379,7 @@ class DatasetModels(collections.abc.Sequence, CovarianceMixin):
 
     Parameters
     ----------
-    models : `SkyModel`, list of `SkyModel` or `Models`
+    models : `~gammapy.modelling.SkyModels`, list of `SkyModel` or `Models`
```
registerrier, line 382:
```suggestion
    models : `SkyModel`, list of `SkyModel` or `Models`
```

gammapy/modeling/models/spatial.py
```diff
@@ -1738,7 +1738,7 @@ def evaluate(self, lon, lat, energy=None, **norms):
         lon = Angle(lon).wrap_at(0 * u.deg)
         lon = Angle(lon).wrap_at(self._wrap_angle)
         # by default rely on CloughTocher2DInterpolator
-        # (Piecewise cubic, C1 smooth, curvature-minimizing interpolant)
+        # (Piecewise cubic, C1 smooth, curvature-minimizing interpolation)
```
registerrier, line 1741:
```suggestion
        # (Piecewise cubic, C1 smooth, curvature-minimizing interpolant)
```

gammapy/modeling/models/temporal.py
```diff
@@ -934,14 +935,14 @@ class TemplatePhaseCurveTemporalModel(TemporalModel):
...
-    The phasecurve is given as a table with columns ``phase`` and ``norm``.
+    The phase curve is given as a table with columns ``phase`` and ``norm``.
 
     The ``norm`` is supposed to be a unit-less multiplicative factor in the model,
     to be multiplied with a spectral model.
 
     The model does linear interpolation for times between the given ``(phase, norm)`` values.
 
-    The implementation currently uses `scipy.interpolate. InterpolatedUnivariateSpline`,
+    The implementation currently uses `scipy.interpolate.InterpolatedUnivariateSpline`,
```
registerrier, line 945:
```suggestion
    The implementation currently uses `~scipy.interpolate.InterpolatedUnivariateSpline`,
```

registerrier (commented):
Thanks @bkhelifi . In practice keeping `~ClassName` works well in the same subpackage. It can have the advantage of working well if we rename the subpackage. Technically, many links you correct were already working.

Now, this is not uniform in the documentation and we have not made a specific choice as how to do this in general. Are there opinions @AtreyeeS @QRemy @Astro-Kirsty ?

gammapy/data/metadata.py
```diff
@@ -122,7 +122,7 @@ class GTIMetaData(MetaData):
 
     Parameters
     ----------
-    reference_time : Time, str
+    reference_time : str
```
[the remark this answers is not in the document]
bkhelifi: It was like this, but @Astro-Kirsty prefered like this...

gammapy/modeling/models/temporal.py
bkhelifi: It was like this, but @Astro-Kirsty prefered like this...

gammapy/modeling/models/core.py
bkhelifi: Why not putting the link towards the documentation? while in general all parameters are links...

Astro-Kirsty:
> Thanks @bkhelifi . In practice keeping `~ClassName` works well in the same subpackage. It can have the advantage of working well if we rename the subpackage. Technically, many links you correct were already working.

Yes, this is what I was also mentioning in my comment. I think we should leave it as it was with `~ClassName` without the entire path.

morcuended (commented):
Thanks for the huge effort @bkhelifi. I left a couple of minor comments.

Regarding whether to use `~ClassName` or the full path, I would also use simply `~ClassName` for better readability. I would keep the full path when the context is not clear and you need to specify where the method is coming from.

registerrier:
Regarding the way we write argument types. Maybe it is useful to check the Sphinx documentation. See:
- [target specification](https://www.sphinx-doc.org/en/master/usage/domains/python.html#target-specification)
- [target resolution](https://www.sphinx-doc.org/en/master/usage/domains/python.html#target-resolution)

From the former the `~` prefix, means that what is shown in the doc is only the last part of the path e.g. `~gammapy.maps.MapAxis` will render `MapAxis` with a link. So `~MapAxis` is probably not needed.

From the latter,  it is written that: `Names in these roles are searched first without any further qualification, then with the current module name prepended, then with the current module and class name (if any) prepended.`
So we don't have to put the full path when the documented object is in the current module. 

External links to other packages docs is done through the intersphinx mappings defined in conf.py. 

The LSST dev doc is also useful [here](https://developer.lsst.io/restructuredtext/style.html#links-to-python-objects).  Note that it suggests to always provide the full namespace [here](https://developer.lsst.io/python/numpydoc.html#concrete-types)

We have to decide on a systematic approach, document it in the dev doc and implement it consistently in docstrings (code reviews in particular have to be clear on this).

My opinion on this is that we should stick to what we have implemented most of time so far: full namespace links with `~` prefix. 

Opinions?

gammapy/modeling/models/core.py
registerrier: Right but here you introducing an incorrect namespace. In current doc all links are OK. See https://docs.gammapy.org/dev/api/gammapy.modeling.models.DatasetModels.html

gammapy/modeling/models/temporal.py
registerrier: OK. The change I propose would remove the full path in the doc, while here I think we need it. So don't apply.

morcuended:
> My opinion on this is that we should stick to what we have implemented most of time so far: full namespace links with ~ prefix, when linking to classes in arguments description.

> For methods or attributes inside a class, this is less clear. See for instance the proposed change:
Normalization used to compute the residuals, see `FluxPointsDataset.residuals`. to Normalization used to compute the residuals, see `~gammapy.datasets.FluxPointsDataset.residuals`.
The issue here is that only residuals will show in the final docs, while it would be useful to have the explicit class name where this belongs, i.e. FluxPointsDataset.residuals as is the case in the [current doc version](https://docs.gammapy.org/dev/api/gammapy.datasets.FluxPointsDataset.html#gammapy.datasets.FluxPointsDataset.plot_residuals).

Thanks for the references @registerrier. I agree with your proposal. We should include these guidelines in the dev doc as you propose.

registerrier (commented):
Thanks @bkhelifi .  So finally, you implemented the opposite of what I suggested to limit changes in this [comment](https://github.com/gammapy/gammapy/pull/5490#issuecomment-2414831951). 
So we really have to define a clear guideline. 

Reviewing 128 modified files is practically impossible unfortunately.

Issue #5358, opened 2024-07-01
Keyword argument in `FluxPoints.read` shares name with arbitrary keyword argument.


cbadams2:
In the following code, the keyword argument `format` carries the same name as a variable which can be added to the `kwargs`. 
https://github.com/gammapy/gammapy/blob/2829b45f3a8b8c4885cc8a3d9984e919c171bab5/gammapy/estimators/points/core.py#L174-L227
Unfortunately this means that the user has no control over the `kwargs` value for `format`, which is set to `ascii.ecsv`, and passed to the astropy `Table.Read` function a few lines later. In some cases, this leads (I would argue unnecessarily) to the `UnicodeDecodeError`.

registerrier:
Maybe adding a `table_format` argument can solve the issue?

lmohrmann:
I ran into this issue as well.

@registerrier you mean replacing `**kwargs` with a dict called `table_kwargs`, right?

Another option would be to rename `format` to `fp_format` ("fp" for FluxPoints) or similar.

PR #5495, opened 2024-10-14
Add `table_format` argument in `FluxPoints.read` to disentangle from arbitrary kwarg 

cgalelli:
This PR is a proposed solution to #5358 
I added the `table_format` argument that can steer the `format` argument of the `astropy.Table` call.


registerrier (approved):
Thanks @cgalelli . 

It would be nice to test with an additional format to check if this works.

bkhelifi (approved):
Thanks @cgalelli

PR #5496, opened 2024-10-14
Change to use `spectral_model` instead of `spectrum`

Astro-Kirsty:
There was some inconsistency with our naming conventions. 
We are calling it `spectrum` in some places but it makes more sense to have it called `spectral_model` as in other places of the code. This PR fixes this. 
It would be good to include this in v1.2 such that we can deprecate it at 2.0


cgalelli (approved):
Thanks @Astro-Kirsty , this kind of consistency checks maybe should be extended to all naming schemes!

registerrier (commented):
Thanks @Astro-Kirsty .

Could you add a few tests to make sure that the warnings are correctly raised? This will also help when removing the deprecation.

bkhelifi (approved):
Looks nice, thank you

PR #5500, opened 2024-10-15
Include pydocstyle and note on bullets points

Astro-Kirsty:
Before v1.2 there was a large effort made to fix the formatting of docstrings. We did not keep an eye on this as we introduced new functionality throughout the past 8 months. 
Therefore I think it is useful to include in the documentation how to check the docstrings with pydocstyle. We should all be using this before we push a PR so that they are in the correct format

Another thing I added was how to handle bullet points in the docstrings. This is often done incorrectly and therefore not rendered correctly in the docs. Each time I adjust these I have to dig through all the code to find an example which is time wasting. I give here a simple example how to implement it!

### DIFF
--- a/docs/development/doc_howto.rst
+++ b/docs/development/doc_howto.rst
@@ -150,6 +150,14 @@ If you get a zsh error try using putting to ignore block inside quotes
 
     pytest --doctest-modules "--ignore-glob=*/tests" gammapy
 
+It is also important to check that you have correctly formatted your docstring. An easy way to check this
+is with the following for your specific file, i.e.:
+
+.. code-block:: bash
+
+    pydocstyle gammapy/data/event_list.py
+
+
 Sphinx gallery extension
 ------------------------
 
@@ -289,6 +297,27 @@ or Gammapy code a bit (either locally with your editor or online on GitHub or vi
 or search the Numpy or Astropy documentation guidelines mentioned above.
 If that doesn't quickly turn up something useful, please ask by putting a comment on the issue or
 pull request you're working on GitHub, or email the Gammapy mailing list.
+
+
+Correct format for bullet point list
+++++++++++++++++++++++++++++++++++++
+
+To correctly add a bullet point list to the docstring, the following can be implemented:
+
+.. testcode::
+
+        """
+        Docstring explanation.
+
+        Parameters
+        ----------
+        parameter_name : parameter_type
+            Description of the parameter with entries:
+                    * option1 : description1
+                    * option2 : description2
+                    * option3 : description3 over more than
+                    one line.
+        """
 
 Functions or class methods that return a single object
 ++++++++++++++++++++++++++++++++++++++++++++++++++++++


docs/development/doc_howto.rst
```diff
@@ -290,6 +298,27 @@ or search the Numpy or Astropy documentation guidelines mentioned above.
...
+To correctly add a bullet point list to the docstring, the following can be implemented:
+
+.. testcode::
+
+        """
+        Docstring explanation.
+
+        Parameters
+        ----------
+        parameter_name : parameter_type
```
bkhelifi, line 314: we do not precise to put the link ? `~gammapy.XX.XX`

docs/development/doc_howto.rst
```diff
@@ -150,6 +150,14 @@ If you get a zsh error try using putting to ignore block inside quotes
 
     pytest --doctest-modules "--ignore-glob=*/tests" gammapy
 
+It is also important to check that you have correctly formatted your docstring. An easy way to check this
+is with the following for your specific file, i.e.:
+
+.. code-block:: bash
+
+    pydocstyle gammapy/data/event_list.py
```
AtreyeeS, line 158: Maybe this can be mandated for every commit, like we have for `black` or `flake8`?

AtreyeeS (approved):
Thank you @Astro-Kirsty !

docs/development/doc_howto.rst
morcuended: Can you please give more details on this @Astro-Kirsty? So then we cannot enforce following a docstring format? BTW, ruff also integrates with pycodestyle, so we could just use it directly after defining the rules.

Astro-Kirsty: We have a number of `pydocstyle`s that we do not implement in gammapy and do not want to. For example the two following
- D107 - Missing docstring in __init__
- D104: Missing docstring in public package (often coming from the __init__.py file being empty, which is the case for the 'tests' folders)

So those will always show incorrectly in the pydocstyle. I understand that you can run `pydocstyle` with the `--ignore=<codes>` but it does not work as expected for me. Therefore I am not sure how we should proceed on this. 

It is also important to note that there are a large number of other errors within the docstrings! Some things I can fix, but it will take time to go over this and make some decision of how to proceed. I am open to ideas!

Astro-Kirsty: OR contradicting errors!
`D213: Multi-line docstring summary should start at the second line`
Fix this and we receive:
`D212: Multi-line docstring summary should start at the first line`

morcuended: I see. Many thanks for the explanation. Then, I'd say we should follow some of those rules and ignore others in the pydocstyle configuration.

Astro-Kirsty: Maybe I should also add a note here about that?

morcuended: I think it would be helpful yes. Something like we follow these particular pydocstyle rules.

Astro-Kirsty: So we tend to follow the numpydoc style, but I do not know if there is a command to check for that as cleanly as the pydocstyle command works

QRemy (approved):

Issue #5493, opened 2024-10-11
WcsNDMap.downsample() `weights` parameter normalization issue


k-egg:
**Gammapy version**
1.2

**Bug description**
When downsampling a WcsNDMap with `preserve_counts=False`, i.e. averaging, the weighted average is not properly calculated. Instead an average of `weights.data*self.data` is returned.

**Expected behavior**
The expected formula for a weighted average would be `np.sum(weights.data*self.data)/np.sum(weights.data)`.

**To Reproduce**

Loading an example dataset from the tutorials:
```python
dataset_cta = MapDataset.read(
    "$GAMMAPY_DATA/cta-1dc-gc/cta-1dc-gc.fits.gz", name="dataset-cta")
```
Compare the scales of
```python
dataset_cta.psf.psf_map.plot_interactive(add_cbar=True)
```
and

```python
dataset_cta.psf.psf_map.downsample(
    2,preserve_counts=False,weights=dataset_cta.psf.exposure_map).plot_interactive(add_cbar=True)
```

The result should be a downsampled map weighted with the exposure map, the resulting scales are, however, around 1e9 too high, as the exposure map values have not been normalized out.

**Other information**
The downsampled map data is calculated in this line in the WcsNDMap definition:
```python
data = block_reduce(self.data * weights, tuple(block_size), func=func)
```
which works only for `preserve_counts=True`, where `func=np.nansum`, but not for `preserve_counts=False`, where `func=np.nanmean`.

A solution could be, for instance:
```python
data = block_reduce(self.data * weights, tuple(block_size), func=np.nansum)
if not preserve_counts:
    weight_sum = block_reduce(weights, tuple(block_size), func=np.nansum)
    data=np.divide(data,weight_sum, where=np.where(data>0.0,True,False))
```

Please tell me if I've missed anything here! Thank you!

adonath:
I would honestly deprecate the `preserve_counts` keyword and fully change to the more general `func` approach. This is more in line with Numpy and avoids the domain specific jargon of preserving counts in a map.

The question of weights is mostly about documentation. The `np.sum(weights)` is just a constant factor that can be precomputed and be part of the weights definition. This has the advantage of not needing to re-compute the normalization constant, where it typically does not change. When you force the normalization user do not have the choice to do something else. On the level of the `Map` as a basic data structure I think it is important to keep flexibility. So my recommendation here would be to improve the documentation and give examples on how to achieve which behavior.

k-egg:
[...]

PR #5507, opened 2024-10-17
IRFmap.downsample error fix

### LINKED ISSUE #5492: IRFMap.downsample() sums instead of averaging
**Gammapy version**
1.2

**Bug description**
As was mentioned in a past dev meeting, IRFMap.downsample() sums instead of averaging over the pixels it merges. This leads to wrong values for EDispKernelMaps and PSFMaps.

**Expected behavior**
IRFMap.downsample should average, not sum over the values. The average could optionally also be weighted using the accompanying exposure map.

**To Reproduce**
Loading an example dataset from the tutorials:
```python
dataset_cta = MapDataset.read(
    "$GAMMAPY_DATA/cta-1dc-gc/cta-1dc-gc.fits.gz", name="dataset-cta")
```
By comparing the scales of the colorbars in
```python
dataset_cta.psf.psf_map.plot_interactive(add_cbar=True)
```
and
```python
dataset_cta.psf.downsample(2).psf_map.plot_interactive(add_cbar=True)
```
it becomes apparent that the scale of the values has grown by a factor of around 4.

**Other information**
This error could be fixed by changing `preserve_counts` from `True` to `False` in the following line in the general IRFMap definition:
```python
irf_map = self._irf_map.downsample(
            factor=factor, axis_name=axis_name, preserve_counts=True, weights=weights
        )
```
Weighting the average with the exposure map could be achieved by specifying `weights=self.exposure_map`, which could be made the default if this is the desired behavior.

(However there seems to be another problem with the `weights` parameter in WcsNDMaps that I will address in a separate issue.)

Please tell me if I've missed anything here or if you'd like me to do a pull request! Thanks!

k-egg:
**Description**

This pull request fixes #5492. To avoid further errors it should be fixed together with #5493.
`preserve_counts` was changed from `True` to `False`, weighting with the exposure map was made the default.
Additionally a warning against using this function for downsampling a PSFMap rad axis was added, as for this case the weights are not calculated correctly.

**Dear reviewer**

The pull request is ready for review! Please let me know if anything should be changed/reworded or if I missed anything.
Thank you!

### DIFF
--- a/gammapy/irf/core.py
+++ b/gammapy/irf/core.py
@@ -987,16 +987,20 @@ class IRFMap:
             Downsampling factor.
         axis_name : str
             Axis to downsample. By default, spatial axes are downsampled.
+            It is not recommended to use this function on a PSFMap rad axis.
         weights : `~gammapy.maps.Map`, optional
-            Map with weights downsampling. Default is None.
+            Map with weights downsampling. Default is IRFMap exposure map.
 
         Returns
         -------
         map : `IRFMap`
             Downsampled IRF map.
         """
+        if not weights:
+            weights=self.exposure_map
+
         irf_map = self._irf_map.downsample(
-            factor=factor, axis_name=axis_name, preserve_counts=True, weights=weights
+            factor=factor, axis_name=axis_name, preserve_counts=False, weights=weights
         )
         if axis_name is None:
             exposure_map = self.exposure_map.downsample(


gammapy/irf/core.py
```diff
@@ -987,16 +987,20 @@ def downsample(self, factor, axis_name=None, weights=None):
             Downsampling factor.
         axis_name : str
             Axis to downsample. By default, spatial axes are downsampled.
+            It is not recommended to use this function on a PSFMap rad axis.
```
Astro-Kirsty, line 990:
```suggestion
            It is not recommended to use this function on a `~gammapy.irf.PSFMap` rad axis.
```

gammapy/irf/core.py
```diff
@@ -987,16 +987,20 @@ def downsample(self, factor, axis_name=None, weights=None):
...
         weights : `~gammapy.maps.Map`, optional
-            Map with weights downsampling. Default is None.
+            Map with weights downsampling. Default is IRFMap exposure map.
```
Astro-Kirsty, line 992:
```suggestion
            Map with weights downsampling. Default is `~gammapy.irf.IRFMap` exposure map.
```

gammapy/irf/core.py
```diff
@@ -987,16 +987,20 @@ def downsample(self, factor, axis_name=None, weights=None):
...
 
         Returns
         -------
         map : `IRFMap`
             Downsampled IRF map.
         """
+        if not weights:
+            weights=self.exposure_map
```
Astro-Kirsty, line 1000: I think the error with formatting it here. You could implement this and see
```suggestion
            weights = self.exposure_map
```

Astro-Kirsty (changes requested):
Thanks @k-egg. 
Just some minor inline comments, and potentially a fix to the formatting.

gammapy/irf/core.py
k-egg: Great, thank you for the suggestions! That could be the issue!

gammapy/irf/core.py
```diff
@@ -989,16 +989,27 @@ def downsample(self, factor, axis_name=None, weights=None):
             Downsampling factor.
         axis_name : str
             Axis to downsample. By default, spatial axes are downsampled.
+            It is not recommended to use this function on a `~gammapy.irf.PSFMap` rad axis.
         weights : `~gammapy.maps.Map`, optional
-            Map with weights downsampling. Default is None.
+            Map with weights downsampling. Default is `~gammapy.irf.IRFMap` exposure map.
```
registerrier, line 994:
```suggestion
            Map with downsampling weights if different from `~gammapy.irf.IRFMap.exposure_map`. Default is None.
```

registerrier: I think you could also explain that the weights geometry has to be the same than of the exposure_map. i.e. no weighing along non-spatial/energy axes. 
Is it correct?

registerrier (changes requested):
Thanks @k-egg ! Sorry for the delay. 

This is nice. I have left some inline comments. 
I think the issue #5492 reveals also that IRFmap.dowsample is under (if at all) tested. Could you a minima add a test for the specific behaviour that should be enforced?

gammapy/irf/core.py
```diff
@@ -981,14 +981,16 @@ def cutout(self, position, width, mode="trim", min_npix=3):
...
     def downsample(self, factor, axis_name=None, weights=None):
-        """Downsample the spatial dimension by a given factor.
+        """Downsample the dimension of the spatial axes or a non-spatial axis by a given factor.
+        It is not recommended to use this function on a `~gammapy.irf.PSFMap` rad axis.
 
         Parameters
         ----------
         factor : int
             Downsampling factor.
         axis_name : str
```
QRemy, line 991:
```suggestion
        axis_name : str, optional
```

QRemy (approved):
Thanks @k-egg, looks good, merging now.

bkhelifi (approved):
Many thanks

PR #5511, opened 2024-10-19
Event filtering in the MapDatasetMaker

bkhelifi:
**Description**

This pull request adds options to filter event during the data reduction through the `MapDatasetMaker`. By default, events are filtered with the observatory GTIs.

**Dear reviewer**
This is a draft and needs to be finished with tests once the PR #5305 is merged.

@MRegeard, for the pulsar data flow, should we add something elsewhere?

### DIFF
--- a/gammapy/makers/map.py
+++ b/gammapy/makers/map.py
@@ -116,7 +116,7 @@ class MapDatasetMaker(Maker):
         background_oversampling=None,
         background_interp_missing_data=True,
         background_pad_offset=True,
-        evt_filter={"gti": True},
+        event_filter={"gti": True},
         check_overlapping_intervals=True,
     ):
         self.background_oversampling = background_oversampling
@@ -132,11 +132,11 @@ class MapDatasetMaker(Maker):
             raise ValueError(f"{difference} is not a valid method.")
         self.selection = selection
 
-        if not set(evt_filter.keys()).issubset(self.available_filter):
+        if not set(event_filter.keys()).issubset(self.available_filter):
             raise ValueError(
-                f"Available methods associated to the 'evt_filter' argument are {self.available_filter}"
+                f"Available methods associated to the 'event_filter' argument are {self.available_filter}"
             )
-        self.evt_filter = evt_filter
+        self.event_filter = event_filter
 
         self.check_overlapping_intervals = check_overlapping_intervals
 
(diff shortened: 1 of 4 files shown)


gammapy/makers/map.py
```diff
@@ -100,13 +108,16 @@ class MapDatasetMaker(Maker):
...
     available_selection = ["counts", "exposure", "background", "psf", "edisp"]
+    available_filter = ["gti", "uti", "none"]
 
     def __init__(
         self,
         selection=None,
         background_oversampling=None,
         background_interp_missing_data=True,
         background_pad_offset=True,
+        evt_filter={"gti": True},
```
QRemy, line 119:
```suggestion
        event_filter={"gti": True},
```

adonath:
Thanks a lot @bkhelifi for this functionality! I have one high level comment: I think conceptionally the filters belong on the `MapDataset` itself. I have not checked in detail how it currently works, but conceptionally the filters are not different from defining e.g. a spatial / energy geometry. I wonder whether we can work more in this direction already.

bkhelifi:
Hi Axel,
Thanks for the return... I am not sure to well understand: for me, the Geom is not a filtering.
Sometimes, events lie outside the geom, and then it acts as a hidden filter, indeed. 
However, I understand your point by considering instead the dataset mask: it conceptually permits some filtering (like the `offset_max` of the SafeMaskMaker). But this is NOT at the event level, but rather at the bin level (after projection).
But, adding a dimension to the mask (e.g. gti, event_type) that does not reflect the associated geom seems weird for me, and even might not work.
And adding extra-dimension to the geom to allow a "mask extension" is, for me, both a mix of concepts and an over-complication (compared to my simple and extendable proposed solution into a maker). And we still need a `maker` that does the event filtering: which one in this case?
But, I might have missed the depth of your point... to be discussed;)

adonath:
Thanks for the quick reply @bkhelifi! I was thinking from a user perspective. So when I start my analysis I would ideally define the empty dataset like:

```python

empty = MapDataset.create(geom, energy_axis_true, event_filters={"gti": (...), "event-type": "psf-1", "phase-interval": (...)})
```
Then the corresponding filters / selections are stored as meta data on the `MapDataset` and actually are applied during the dataset creation process. This keeps all the configuration / event selection in a single place and keeps the meta data alive down the data reduction pipeline. The mask is indeed a pixel based filter, that remains as it is.

Looking at the code you suggest I agree that there could be an additional flag on the maker, whether the filter is applied or not. But I think the definition of the filter itself should be associated in some way to the dataset definition, as I have outline above.

bkhelifi:
OK, very interesting! 
Indeed, defining everything into the DL4 object will be of great help (provenance, serialisation). I like this concept.
With it, the "filtering" function should be coded where? 
A. In the MapDatasetMaker class, I guess, having only a Boolean flag `apply_filter`
B. or maybe with a filtering function into the MapDataset class that is called by the MapDatasetMaker (but then, we have to pass the `obs` object to the MapDataset filtering function)

I have a parallel thought on one future feature that one should add soon: the handling of `evt_type`. A typical use case will be: from one observation, one creates N stacked MapDataset instances (one per chosen `evt_type`). One can have different designs:
1. for each `evt_type` associated to 1 MapDataset with all its metadata (following your idea), we do a data reduction loop (-> ugly)
2. having the same apparent data reduction loop, but instead of having one reference Dataset one have a Datasets object that is a list of DataSet instances (one per `evt_type`). The staking function should be then be intelligent, allowing an individual stacking only for similar Dataset metadata . All the job is made then by the MapDatasetMaker that becomes complex. One should then check how to handle the parallelism of the DatasetsMaker.
3. Maybe a similar scheme, but with the DatasetsMaker (centgral management of the parallelism)
4. Other idea?

I'm trying to project myself a little bit...

Issue #5501, opened 2024-10-15
Rename cta -> ctao in observatory_locations?


morcuended:
https://github.com/gammapy/gammapy/blob/2a4d540738e6a25450c0655ca5d9b2b57903e91e/gammapy/utils/observers.py#L55

For both `cta_south` and `cta_north` and all references to these key names.

bkhelifi:
Indeed, one needs to update the names and put a log.ingo message. Could you do it?

Astro-Kirsty:
Are you still working on this @morcuended?

morcuended:
No, I did not follow it up.

PR #5513, opened 2024-10-21
Use `ctao_south` and `ctao_north` in observatory locations defined in Gammapy

morcuended:
**Description**

This PR is to addresses #5501. 

**Dear reviewer**

@bkhelifi, regarding the log info message, should it go inside the definition of the `observatory_locations` dictionary? Maybe using `@deprecated_renamed_argument`?


bkhelifi (dismissed):
It looks good for me! Thanks a lot

morcuended:
> a warning and internally overload the parameter to ctao-north such that the code continues to run

To be able to do this I propose creating a class that handles the deprecation warning. Let me know your opinions. I'll add a test for this deprecation warning as well.

Astro-Kirsty:
Ahh I see the issue! 
This is not really a deprecation as it is a dictionary.

I would suggest to make it a little less complicated and stay with the 'log.warning' idea (based off your class idea):

```
class ObservatoryDict(dict):
    def __missing__(self, key):
        if key == "cta_north":
            log.warning("The key 'cta_north' is incorrect. Automatically changing the location to 'ctao_north'.")
            return self["ctao_north"]
        else:
            raise KeyError(f"'{key}' not found in observatory locations.")

observatory_locations = ObservatoryDict()
"""Gamma-ray observatory locations (dict).....
"""
```

bkhelifi:
Indeed, this is a depreciation of a default value inside a dictionnary. This issue has happened couple of time and we had the only solution to send a log.info or log.warning to inform that the dict entry has changed.
One might add a dict_entry_depreciation in the utils... because this will happen in the future again and again

morcuended:
Thanks for the feedback @Astro-Kirsty, @bkhelifi! I'll work it out following your suggestions.

AtreyeeS:
What is the status here @morcuended ? 
I think the modification of deprecation warning can be an independent PR.

morcuended:
I stopped working on it. I will resume it next week.

gammapy/datasets/simulate.py
```diff
@@ -532,11 +532,11 @@ def event_list_meta(dataset, observation, keep_mc_id=True):
         if loc is None:
             if telescope == "CTA":
```
bkhelifi, line 533: For the initial release, the stored metadata is really CTA?
For the next release, it might be CTAO. So I propose to add an OR condition with telescope == "CTAO"

To @morcuended and Matthias , should we keep in mind that the CTAO data releases will use the following metadata : 
- TELESCOP: "CTAO"
- INSTRUME: "Southern Array" and "Northern Array"
Is it the anticipated standard values?

gammapy/utils/deprecation.py
```diff
@@ -51,3 +51,24 @@ def deprecated_attribute(name, since, **kwargs):
 
     kwargs["warning_type"] = GammapyDeprecationWarning
     return deprecated_attribute(name, since, **kwargs)
+
+
+def deprecated_key(key, old_keys, new_keys, since):
+    """
+    Raise a deprecation warning if provided key is deprecated.
```
bkhelifi, line 58:
```suggestion
    Raise a deprecation warning if the provided key is deprecated.
```

bkhelifi (commented):
Thank you a lot... I have some comments to this PR.

gammapy/utils/observers.py
```diff
@@ -2,109 +2,163 @@
...
+    Class that defines a dict with observatory names as keys
+    and values of type `~astropy.coordinates.EarthLocation`.
+
+    Note that with ``EarthLocation`` the orientation of angles is as follows:
+
+    - longitude is east for positive values and west for negative values
+    - latitude is north for positive values and south for negative values
+
+    Available observatories (alphabetical order):
+
```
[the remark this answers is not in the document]
Astro-Kirsty: Maybe we can add these in a separate PR?

gammapy/utils/observers.py
```diff
@@ -2,109 +2,163 @@
...
+            "astri": EarthLocation(
+                lon="-16d30m20.99s", lat="28d18m00.0s", height="2370m"
+            ),
+            # coordinates from fact-tools (based on google earth)
+            "fact": EarthLocation(
+                lat=28.761647 * u.deg,
+                lon=-17.891116 * u.deg,
+                height=2200 * u.m,
+            ),
+        }
```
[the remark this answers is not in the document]
Astro-Kirsty: As above

gammapy/utils/deprecation.py
```diff
@@ -51,3 +51,24 @@ def deprecated_attribute(name, since, **kwargs):
 
     kwargs["warning_type"] = GammapyDeprecationWarning
     return deprecated_attribute(name, since, **kwargs)
+
+
+def deprecated_key(key, old_keys, new_keys, since):
```
QRemy, line 56: In my opinion the is not needed the keys are not deprecated but the module variable `observation_locations` is.

gammapy/utils/observers.py
```diff
@@ -1,110 +1,152 @@
...
 """Location of gamma-ray observatories."""
+
 import astropy.units as u
 from astropy.coordinates import EarthLocation
+from gammapy.utils.deprecation import deprecated_key
 
 __all__ = ["observatory_locations"]
 
 
-observatory_locations = {}
```
QRemy, line 9: This should not be deleted but instead

simply introduce:
```
class DeprecatedDict(dict):
    old_keys_to_new_keys = {'cta_south': 'ctao_south', 'cta_north': 'ctao_north'}
    def __getitem__(self, key):
        if key in self.old_keys_to_new_keys:
            msg = 'Please use the new key: `{}` instead of `{}`'.format(self.old_keys_to_new_keys[key], key)
            warn(msg)
        return super().__getitem__(key)
    def __setitem__(self, key, value):
        if key in self.old_keys_to_new_keys:
            msg = 'Please use the new key: `{}` instead of `{}`'.format(self.old_keys_to_new_keys[key], key)
            warn(msg, GammapyDeprecationWarning)
        return super().__setitem__(key, value)
        
observatory_locations = DeprecatedDict()        
```        

and you need nothing else to achieve what is in the title of the PR.

gammapy/utils/observers.py
```diff
@@ -1,110 +1,152 @@
...
-    lon="-16d30m20.99s", lat="28d18m00.0s", height="2370m"
-)
-
-# coordinates from fact-tools (based on google earth)
-observatory_locations["fact"] = EarthLocation(
-    lat=28.761647 * u.deg,
-    lon=-17.891116 * u.deg,
-    height=2200 * u.m,
-)
+class Observatories:
```
QRemy, line 11: `Observatories` should be a list of `Observatory`, each  `Observatory` being a class including location as a property among others.

QRemy (changes requested):
Is this PR only to rename ctao_south and ctao_north in observatory locations or introduce a larger concept of Observatory class, the latter should be a separate PR with a better design.

registerrier (commented):
Thanks @morcuended and @Astro-Kirsty . This is good.

In general, I would even propose to remove this from gammapy and rely instead on astropy-data, though `EarthLocation.from_site()` adding entries there:
https://github.com/astropy/astropy-data/blob/gh-pages/coordinates/sites.json

Is there a good reason for us to have the list in gammapy rather than more broadly available to the community in astropy? Opinions @QRemy @AtreyeeS @bkhelifi @adonath ?

gammapy/utils/observers.py
Astro-Kirsty: Yes, but the idea was to introduce a deprecation class for dictionaries in general. 

The only reason why we introduced a class is such that it can be correctly deprecated, while still behaving like a dictionary, which is why there are so many functions added.

QRemy:
> In general, I would even propose to remove this from gammapy and rely instead on astropy-data, though `EarthLocation.from_site()` adding entries there: https://github.com/astropy/astropy-data/blob/gh-pages/coordinates/sites.json

I agree if we can remove it that's even better. I guess we can still have a deprecation using the `DeprecatedDict ` class I mentioned above adding a warning on init that recommend to use `EarthLocation.from_site()` instead.

gammapy/utils/observers.py
QRemy: `old_keys_to_new_keys` can be moved to init so this `DeprecatedDict` would be a general deprecation class for dictionaries

```
class DeprecatedDict(dict):
    old_keys_to_new_keys = {}
    def __init__(self, data, old_keys_to_new_keys):
        if old_keys_to_new_keys:
            self.old_keys_to_new_keys=old_keys_to_new_keys
        return super().__init__(data)
    
    def __getitem__(self, key):
        if key in self.old_keys_to_new_keys:
            msg = 'Please use the new key: `{}` instead of `{}`'.format(self.old_keys_to_new_keys[key], key)
            warn(msg)
        return super().__getitem__(key)
    def __setitem__(self, key, value):
        if key in self.old_keys_to_new_keys:
            msg = 'Please use the new key: `{}` instead of `{}`'.format(self.old_keys_to_new_keys[key], key)
            warn(msg, GammapyDeprecationWarning)
        return super().__setitem__(key, value)
 
observatory_locations = DeprecatedDict(observatory_locations, {'cta_south': 'ctao_south', 'cta_north': 'ctao_north'})   

```
observatory_locations = DeprecatedDict()

Astro-Kirsty: Okay, but dont we still need the other items to ensure it acts exactly like a dictionary?

QRemy: no it inherits from dict class

QRemy: only the new key will defined in `observatory_locations` but the old key will also work and emit a warning.

Astro-Kirsty: Yes, apologies, my script was outdated

gammapy/utils/deprecation.py
```diff
@@ -51,3 +51,59 @@ def deprecated_attribute(name, since, **kwargs):
...
```
QRemy, line 56: this is no longer needed right ?

gammapy/utils/deprecation.py
```diff
@@ -51,3 +51,59 @@ def deprecated_attribute(name, since, **kwargs):
...
+            f'"{key}" was deprecated in version {since} and will be removed in a future version. '
+            f'Use "{new_key}" instead.',
+            GammapyDeprecationWarning,
+            stacklevel=2,
+        )
+        return new_key
+    return key
+
+
+class DeprecatedDict(dict):
```
QRemy, line 77: this should probably go in a separate PR first with its own unit tests.

QRemy (commented):
Thanks @Astro-Kirsty for the update, just few remaining comments:

registerrier:
On second thought, I fear that this is getting way too complex for such a small problem.

What this shows is that the whole idea of the observatory dictionary is not a good design as it does not allow several names for a given observatory nor for deprecation. (and in general all global variables suffer from this).
 
My proposal for a simple short term solution is to duplicate ctao entries with `"ctao_north"` and  `"ctao_south"` as keys and only show the latter in the docs, if the name change is considered important enough. 
Then we have to decide on what functionalities and design for the observatory locations registry in gammapy.

QRemy:
> My proposal for a simple short term solution is to duplicate ctao entries with `"ctao_north"` and `"ctao_south"` as keys and only show the latter in the docs, if the name change is considered important enough. Then we have to decide on what functionalities and design for the observatory locations registry in gammapy.

I agree just pushed a commit to remove the deprecation system and just duplicated the entries, will merge as soon as the CI pass.

PR #5515, opened 2024-10-21
Link to the safe maker api

Astro-Kirsty:
Last week during the CAD school we had some questions about the SafeMask and I noticed that we do not link to the information in the API tutorial within the `SafeMaskMaker`. 

This provide this **very useful** link as well as a few clean ups of the two documents these relate to.


bkhelifi (approved):
All good

morcuended (approved):
Thanks, @Astro-Kirsty, it looks good. Very helpful link, indeed.

registerrier (approved):
Thanks @Astro-Kirsty .

Issue #5191, opened 2024-04-02
Docs are unreadable in dark mode and there is no button to switch to light


maxnoe:
**Gammapy version**

Current developer documentation

**Bug description**

The current styling of the Gammapy documentation results in unreadable text when using dark mode, which is used by default if the system is configured to be dark:

![docs_theme_dark](https://github.com/gammapy/gammapy/assets/5488440/5e2f5d61-1fee-4fa3-97a4-0f066ad701ad)

**Expected behavior**

Use a styling that results in well-readable text in dark mode or force light mode.

**To Reproduce**

Visit the documentation with system configured to use dark mode.

morcuended:
I think the docs in dark mode are still not readable (at least the current dev version)


![immagine](https://github.com/user-attachments/assets/785f6483-2d50-4013-9959-9ba99714f968)

The text color in the main body should change

Astro-Kirsty:
@morcuended, it is not explained here, but the decision was that we do not 'support' the dark theme. Hence why we decide to force the light theme.

morcuended:
Ah! thanks @Astro-Kirsty but the dev version of the docs still allows setting the dark theme. I think the switch button should then be removed, no?

Astro-Kirsty:
Yes that is probably a good idea. Can you make a PR please? 
This choice was for v1.3, I am not sure what the plan is v2.0. 
So lets just move the milestone here for v2.0 and we can discuss later

PR #5517, opened 2024-10-21
Remove switch for chosing between light/dark themes in docs

morcuended:
**Description**

This pull request simply removes the possibility of switching among light or dark themes for the documentation in the sphinx configuration (options in `"navbar_end"`).

**Dear reviewer**

The current dev version of the documentation still allows to switch between dark and light modes, while it was decided to only support and force the use of the light one. If not supported, it should not be possible to change it. See #5191.

If later on we decide to support the dark mode, we can bring this option back.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -237,7 +237,7 @@ html_theme_options = {
         "json_url": "https://docs.gammapy.org/stable/switcher.json",
         "version_match": switch_version,
     },
-    "navbar_end": ["version-switcher", "theme-switcher", "navbar-icon-links"],
+    "navbar_end": ["version-switcher", "navbar-icon-links"],
     "navigation_with_keys": True,
     # footers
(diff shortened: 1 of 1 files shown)


bkhelifi (approved):
Thanks a lot. If the css can not be changed, this is a better to remove this option

Astro-Kirsty:
The docs look good!

bkhelifi:
I am wondering whether one should not add an issue to offer back the dark theme by changing the css... One of the issues was the text color in this case. To be investaged for the v2.0

morcuended:
> I am wondering whether one should not add an issue to offer back the dark theme by changing the css... One of the issues was the text color in this case. To be investaged for the v2.0

Yes, definitely

Astro-Kirsty:
See https://github.com/gammapy/gammapy/issues/5520 for this new issue

PR #5519, opened 2024-10-22
Improve the lightcurve notebook

AtreyeeS:
I think we unnecessarily introduced a lot of misunderstanding by not fitting the model before running the estimator. This is not time consuming, so I adapted the notebook for clarity. Also some other minor modifications to expose more features

### DIFF
--- a/examples/tutorials/analysis-time/light_curve_flare.py
+++ b/examples/tutorials/analysis-time/light_curve_flare.py
@@ -304,6 +304,8 @@ lc_1d = lc_maker_1d.run(datasets)
 # Finally we plot the result for the 1D lightcurve:
 #
 plt.figure(figsize=(8, 6))
+plt.tight_layout()
+plt.subplots_adjust(bottom=0.3)
 lc_1d.plot(marker="o", axis_name="time", sed_type="flux")
 plt.show()
 
@@ -326,6 +328,8 @@ print(axis_new)
 
 lc_new = sliced_lc.resample_axis(axis_new)
 plt.figure(figsize=(8, 6))
+plt.tight_layout()
+plt.subplots_adjust(bottom=0.3)
 ax = sliced_lc.plot(label="original")
 lc_new.plot(ax=ax, label="rebinned")
 plt.legend()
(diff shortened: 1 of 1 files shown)


Astro-Kirsty (changes requested):
Thanks @AtreyeeS, just some minor formatting comments

AtreyeeS:
Thanks @Astro-Kirsty ! I adapted the format..

@cgalelli! Yeah I agree. But the rebinning was there already.. Here I also added the slicing to show how to get a flat lc, and then perform some operationon it... We can keep this now and remove it for 2.0

cgalelli (dismissed):
Thanks @AtreyeeS ! 
I don't think there is necessarily the need to remove parts of this tutorial, my comment was more generic in scope, on the organization and distribution of the concepts in the different notebooks.

bkhelifi (approved):
Nice improvement that shows more features. Thanks a lot

examples/tutorials/analysis-time/light_curve_flare.py
```diff
@@ -301,21 +304,37 @@
...
-# Light curves once obtained can be rebinned.
-# Here, we rebin 4 adjacent bins together, to get 30 min bins
+# Light curves once obtained can be rebinned using the likelihood profiles.
+# Here, we rebin 3 adjacent bins together, to get 30 minute bins.
+#
+# We will first slice `lc_1d` to obtain the lightcurve in the first energy bin
 #
 
-axis_new = get_rebinned_axis(lc_1d, method="fixed-bins", group_size=3, axis_name="time")
+slices = {"energy": slice(0, 1)}
```
registerrier, line 320: This could be shown in the estimators notebook instead, no?

registerrier (commented):
Thanks @AtreyeeS . How does this change the CI time? In general fitting with all independent datasets might be lengthy.

examples/tutorials/analysis-time/light_curve_flare.py
QRemy: we can have it repeated here

QRemy (approved):
Thanks @AtreyeeS, merging now.

Issue #5512, opened 2024-10-21
Precise into the LC/Flux Points tutorial why and how to adjust the scan values limits


bkhelifi:
**Is your feature request related to a problem? Please describe.**
Nan values can appear from the flux points computations (spectrum or light curves).

**Describe the solution you'd like**
- Explain why and show how to change the limits of the scan values.
- Put a link towards the solution within an HowTo

AtreyeeS:
Hi @bkhelifi ! @Astro-Kirsty added some info here https://docs.gammapy.org/dev/tutorials/api/estimators.html#a-fully-configured-flux-points-estimation
Would it suffice to add a link from the HowTo, or would you prefer more explanations?

bkhelifi:
I missed that! perfect... thanks
Maybe an Howto is just enough. Any thought?

QRemy:
Resolved by #5522

PR #5522, opened 2024-10-22
Add HowTo for configuring norm min/max values 

AtreyeeS:
Addresses #5512

### DIFF
--- a/docs/user-guide/howto.rst
+++ b/docs/user-guide/howto.rst
@@ -279,6 +279,17 @@ warning like so:
 .. accordion-footer::
 
 .. accordion-header::
+    :id: HowToAvoidNaninFp
+    :title: Avoid `nan` results in Flux Point estimation
+    :link: ../tutorials/api/estimators.html#a-fully-configured-flux-points-estimation
+
+Sometimes, upper limit values may show as `nan` while running a `FluxPointsEstimator` or a `LightCurveEstimator`.
+This often arises because the range of the norm parameter being scanned over is not sufficient. Increasing this range
+usually solves the problem. In some cases, you can also consider configuring the estimator with a different `Fit` backend.
+
+.. accordion-footer::
+
+.. accordion-header::
     :id: HowToProgressBar
     :title: Display a progress bar
 
--- a/docs/user-guide/howto.rst
+++ b/docs/user-guide/howto.rst
@@ -283,9 +283,10 @@ warning like so:
     :title: Avoid NaN results in Flux Point estimation
     :link: ../tutorials/api/estimators.html#a-fully-configured-flux-points-estimation
 
-Sometimes, upper limit values may show as `nan` while running a `FluxPointsEstimator` or a `LightCurveEstimator`.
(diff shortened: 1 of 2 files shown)


examples/tutorials/api/estimators.py
```diff
@@ -116,10 +116,10 @@
 #
 
 fp_estimator.norm.scan_min = 0.1
-fp_estimator.norm.scan_max = 3
+fp_estimator.norm.scan_max = 10
 
 ######################################################################
-# Note: The default scan range of the norm parameter is between 0.1 to 10. In case the upper
+# Note: The default scan range of the norm parameter is between 0.1 to 5. In case the upper
```
Astro-Kirsty, line 122:
```suggestion
# Note: The default scan range of the norm parameter is between 0.2 to 5. In case the upper
```

docs/user-guide/howto.rst
```diff
@@ -278,6 +278,17 @@ warning like so:
 
 .. accordion-footer::
 
+.. accordion-header::
+    :id: HowToAvoidNaninFp
+    :title: Avoid `nan` results in Flux Point estimation
```
Astro-Kirsty, line 283:
```suggestion
    :title: Avoid NaN results in Flux Point estimation
```

docs/user-guide/howto.rst
```diff
@@ -278,6 +278,17 @@ warning like so:
...
+    :link: ../tutorials/api/estimators.html#a-fully-configured-flux-points-estimation
+
+Sometimes, upper limit values may show as `nan` while running a `FluxPointsEstimator` or a `LightCurveEstimator`.
+This often arises because the range of the norm parameter being scanned over is not sufficient. Increasing this range
+usually solves the problem. In some cases, you can also consider configuring the estimator with a different `Fit` backend.
```
Astro-Kirsty, line 288:
```suggestion
Sometimes, upper limit values may show as ``nan`` while running a `~gammapy.estimators.FluxPointsEstimator`
or a `~gammapy.estimators.LightCurveEstimator`. This often arises because the range of the norm parameter
being scanned over is not sufficient. Increasing this range usually solves the problem. In some cases,
you can also consider configuring the estimator with a different `~gammapy.modeling.Fit` backend.
```

Astro-Kirsty (changes requested):
Thanks @AtreyeeS. Nice addition!

Good pick up on the incorrect max, the min was also incorrect actually.
I also suggest to add the full paths for the objects.

docs/user-guide/howto.rst
bkhelifi: `~gammapy.estimators.FluxPointsEstimator` -> `gammapy.estimators.FluxPointsEstimator` ? the full path might be great into a HowTo

bkhelifi: the same for the LCestimator

bkhelifi (approved):
This HowTo is so important. Thank you... I have just a minor comment (that is not a no-go!)

QRemy (approved):
Thanks @AtreyeeS, no further comment

PR #5525, opened 2024-10-24
Norm `TemplatePhaseCurveTemporalModel`

### LINKED ISSUE #5524: `TemplatePhaseCurveTemporalModel` does not normalise the phasecurve template
The `TemplatePhaseCurveTemporalModel` does not normalise the template at init. This leads to big error, because the value of the norm per phase bin might be very different from 1, which means that the flux is falsely estimated, in particular when simulating data.

MRegeard:
This PR fixes #5524. 

I create a method to normalise the "NORM" of table. I need to create the interpolator once to compute the integral. 
This way, we can keep the `lazyproperty` on `_interpolator`.  I also added the possibility to use weights with the "WEIGHT" keyword in the table.

I also slightly modify the `integral` method because in the old one, if you integrate over time that correspond to less than a phase (phi_min=0, phi_max=0.4), the integral was wrong (>> than integral over 1 period).

### DIFF
--- a/gammapy/modeling/models/temporal.py
+++ b/gammapy/modeling/models/temporal.py
@@ -1,5 +1,6 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 """Time-dependent models."""
+
 import logging
 import numpy as np
 import scipy.interpolate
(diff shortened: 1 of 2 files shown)


bkhelifi (approved):
Thanks @MRegeard . These are important fixes

QRemy (approved):
Thanks @MRegeard, no further comment.

PR #5497, opened 2024-10-14
Add a Data Privacy mention in the doc

bkhelifi:
**Description**

In this pull request, the Data Privacy is mentioned, after the use of Matomo. It is made by making a link towards the main site gammapy.org with a link put in the footer.

PR #5527, opened 2024-10-25
Remove 2nd static path

MRegeard:
This PR removes the second static path in conf.py which seems useless.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -336,7 +336,6 @@ sphinx_gallery_conf = {
     },
 }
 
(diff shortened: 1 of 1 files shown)


morcuended (approved):
Indeed, it was already defined above. Thanks @MRegeard

bkhelifi:
To report a chat with Maxime to all, this is already included into the PR #5497...

MRegeard:
Closing as it is already implemented in https://github.com/gammapy/gammapy/pull/5497.

Issue #5510, opened 2024-10-18
Call to plt.legend() within gammapy.maps.region.ndmap.py triggering a matplotlib info/warning message.


mireianievas:
**Gammapy version**
1.3.dev726+g6c6d15956

(this is present in the stable release too)

**Bug description**
Loading FluxPointDatasets from a Light Curve file previously generated with the LightCurveEstimator and then running the .plot() method triggers this annoying warning in matplotlib:

`No artists with labels found to put in legend.  Note that artists whose label start with an underscore are ignored when legend() is called with no argument.`.

If you call the method N times for N FluxPointsDatasets, you get that message repeated N times. 

The issue comes from gammapy.maps.region.ndmap.py:L132

```
if len(self.geom.axes) > 1:
    plt.legend()
```

Here for the LC the geometry contains 2 axes (time and energy) therefore fulfilling that condition. 

**Expected behavior**
probably that if should be different, or avoid calling there plt.legend() at all and letting the user handle the legend plotting on demand. 

**To Reproduce**

```
LC_Swift_XRT = FluxPointsDataset.read("XRT_2.0-10keV.ecsv",name='Swift_XRT_HE')
fig,axs = plt.subplots(nrows=8,ncols=1,figsize=(10,15),dpi=120,sharex=True)
name = LC_Swift_XRT.name
LC_Swift_XRT.data.plot(ax=axs[3],sed_type='eflux',
    color=colors[name],
    marker=markers[name],markersize=4,mfc='white',
    label=name,
)
```

Example LC file

```
# %ECSV 1.0
# ---
# datatype:
# - {name: time_min, datatype: float64}
# - {name: time_max, datatype: float64}
# - {name: e_ref, unit: keV, datatype: string, subtype: 'float64[1]'}
# - {name: e_min, unit: keV, datatype: string, subtype: 'float64[1]'}
# - {name: e_max, unit: keV, datatype: string, subtype: 'float64[1]'}
# - {name: ref_dnde, unit: 1 / (keV s cm2), datatype: string, subtype: 'float64[1]'}
# - {name: ref_flux, unit: 1 / (s cm2), datatype: string, subtype: 'float64[1]'}
# - {name: ref_eflux, unit: keV / (s cm2), datatype: string, subtype: 'float64[1]'}
# - {name: norm, unit: '', datatype: string, subtype: 'float64[1]'}
# - {name: norm_err, unit: '', datatype: string, subtype: 'float64[1]'}
[...]

PR #5529, opened 2024-10-25
Remove call to plt.legend() in ndmap's plot() method

mireianievas:
For LC, datasets consist on two axes: time and energy. Making a call to plt.legend() without arguments triggers the following the following message in gammapy:

"No artists with labels found to put in legend.  Note that artists whose label start with an underscore are ignored when legend() is called with no argument.."

If you plot N FluxPoints, the message is printed N times. 

Related issue: #5510

### DIFF
--- a/gammapy/maps/region/ndmap.py
+++ b/gammapy/maps/region/ndmap.py
@@ -159,9 +159,6 @@ class RegionNDMap(Map):
         if "energy" in axis_name:
             ax.set_yscale("log", nonpositive="clip")
 
-        if len(self.geom.axes) > 1:
-            plt.legend()
-
         return ax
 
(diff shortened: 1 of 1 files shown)


bkhelifi:
Can we check with the git history when we have added the plt.legend? And to see whether it does not impact the plots of nbs, tutorials (peek functions), etc?
It might be sad that some useful legends disappear.

mireianievas:
Here:

https://github.com/gammapy/gammapy/commit/df6d7212d292027f2b80cf064418eed08bf05ece?after=b213dbf488429db171df379f8af2ebf602632ecc+34

QRemy (approved):
Thanks @mireianievas, no comment from my side.

PR #5533, opened 2024-10-30
Improve utility function to create map dataset from irfs

QRemy:
Add a function `create_empty_map_dataset_from_irfs` and reuse it into `create_map_dataset_from_observation` (before everthing was done in `create_map_dataset_from_observation`). 
Now `create_empty_map_dataset_from_irfs` also support `PSFMap` and `EDispMap`.

### DIFF
--- a/gammapy/datasets/map.py
+++ b/gammapy/datasets/map.py
@@ -199,7 +199,7 @@ def _default_width(observation, spatial_width_max=12 * u.deg):
     # width estimated from the rad_max or the offset_max
     if getattr(observation, "rad_max", False) and observation.rad_max is not None:
         width = 2.0 * observation.rad_max.quantity.max()
-    elif getattr(observation, "axes", False):
+    elif hasattr(observation.psf, "axes"):
         width = 2.0 * observation.psf.axes["offset"].edges[-1]
     else:
         width = 2.0 * np.max(observation.psf.psf_map.geom.width)
@@ -208,7 +208,6 @@ def _default_width(observation, spatial_width_max=12 * u.deg):
 
 def create_empty_map_dataset_from_irfs(
     data,
-    models=None,
     dataset_name=None,
     energy_axis_true=None,
     energy_axis=None,
@@ -228,8 +227,6 @@ def create_empty_map_dataset_from_irfs(
     ----------
     data : `~gammapy.data.Observation` or `~gammapy.data.MapDataset`
         Observation or Dataset containing the IRFs.
-    models : `~gammapy.modeling.Models`, optional
-        Models. Default is None.
     dataset_name : str, optional
         If `models` contains one or multiple `FoVBackgroundModel`
         it should match the `dataset_name` of the background model to use.
@@ -274,19 +271,11 @@ def create_empty_map_dataset_from_irfs(
         if energy_axis_true is None:
             energy_axis_true = energy_axis_true_
 
-    if models is None:
-        models = Models()
-
     if dataset_name is None:
-        dataset_name = f"obs_{getattr(data, 'obs_id', data.name)}"
-
-    if not np.any(
-        [
-            isinstance(m, FoVBackgroundModel) and m.datasets_names[0] == dataset_name
-            for m in models
-        ]
-    ):
-        models.append(FoVBackgroundModel(dataset_name=dataset_name))
+        if hasattr(data, "obs_id"):
+            dataset_name = f"obs_{data.obs_id}"
+        else:
+            dataset_name = f"{data.name}"
 
     if position is None:
         if hasattr(data, "pointing"):
@@ -315,11 +304,12 @@ def create_empty_map_dataset_from_irfs(
         else:
             axes["migra_axis"] = data.edisp.edisp_map.geom.axes["migra"]
 
-    return MapDataset.create(
+    dataset = MapDataset.create(
         geom,
         name=dataset_name,
         **axes,
     )
+    return dataset
 
 
 def create_map_dataset_from_observation(
@@ -377,7 +367,6 @@ def create_map_dataset_from_observation(
 
     dataset = create_empty_map_dataset_from_irfs(
         observation,
-        models=models,
         dataset_name=dataset_name,
         energy_axis_true=energy_axis_true,
         energy_axis=energy_axis,
@@ -389,6 +378,16 @@ def create_map_dataset_from_observation(
         position=position,
         frame=frame,
     )
+
+    if models is None:
+        models = Models()
+    if not np.any(
+        [
+            isinstance(m, FoVBackgroundModel) and m.datasets_names[0] == dataset.name
+            for m in models
+        ]
+    ):
+        models.append(FoVBackgroundModel(dataset_name=dataset.name))
 
     components = ["exposure"]
     if observation.edisp is not None:
--- a/gammapy/datasets/map.py
+++ b/gammapy/datasets/map.py
@@ -149,7 +149,7 @@ def _default_energy_axis(observation, energy_bin_per_decade_max=30, position=Non
     print(eres)
 
     eres = eres[np.isfinite(eres) & (eres > 0.0)]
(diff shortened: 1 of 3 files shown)


QRemy:
Requires #5545 to be merged  first.

registerrier:
Thanks @QRemy . Is this for 1.3 or later?

QRemy:
> Thanks @QRemy . Is this for 1.3 or later?

This could go in with 1.3

gammapy/datasets/map.py
```diff
@@ -169,13 +197,121 @@ def _default_binsz(observation, spatial_bin_size_min=0.01 * u.deg):
...
+        True energy axis. Default is None.
+        If None it is determined from the observation IRFs.
+    energy_axis : `~gammapy.maps.MapAxis`, optional
+        Reconstructed energy axis. Default is None.
+        If None it is determined from the observation IRFs.
+    energy_bin_per_decade_max : int, optional
+        Maximal number of bin per decade in energy for the reference dataset
+    spatial_width : `~astropy.units.Quantity`, optional
+        Spatial window size. Default is None.
+         If None it is determined from the observation offset max or rad max.
```
registerrier, line 244:
```suggestion
        Spatial window size. Default is None.
        If None it is determined from the observation offset max or rad max.
```

gammapy/datasets/map.py
```diff
@@ -169,13 +197,121 @@ def _default_binsz(observation, spatial_bin_size_min=0.01 * u.deg):
...
+    if data.edisp is not None:
+        if hasattr(data.edisp, "axes"):
+            axes["migra_axis"] = data.edisp.axes["migra"]
+        else:
+            axes["migra_axis"] = data.edisp.edisp_map.geom.axes["migra"]
+
+    dataset = MapDataset.create(
+        geom,
+        name=dataset_name,
+        **axes,
```
registerrier, line 310: What about the `irf_binsz` ? I think this one is important to cover as well.

gammapy/datasets/map.py
```diff
@@ -220,64 +358,45 @@ def create_map_dataset_from_observation(
         If None it is determined from the observation PSF R68.
     spatial_bin_size_min : `~astropy.quantity.Quantity`, optional
         Minimal spatial bin size. Default is 0.01 degree.
-
+    position : `~astropy.coordinates.SkyCoord`, optional
+        Center of the geometry. Defalut is the observation pointing.
+    frame: str, optional
+        frame of the coordinate system. Defalut is icrs.
```
registerrier, line 364:
```suggestion
        frame of the coordinate system. Default is "icrs".
```

gammapy/datasets/map.py
```diff
@@ -169,13 +197,121 @@ def _default_binsz(observation, spatial_bin_size_min=0.01 * u.deg):
...
         width = 2.0 * np.max(observation.rad_max)
-    else:
+    elif hasattr(observation.psf, "axes"):
         width = 2.0 * observation.psf.axes["offset"].edges[-1]
+    else:
+        width = 2.0 * np.max(observation.psf.psf_map.geom.width)
     return np.minimum(width, spatial_width_max)
 
 
+def create_empty_map_dataset_from_irfs(
```
registerrier, line 209: I think the function is doing too much which makes it complex to understand.
- it can create a `MapDataset` without any IRF, contrary to its name, if optional arguments are all passed. This use case is covered by `MapDataset.create`.
- it deals with DL3 and DL4 IRFs at the same time. I think those are different use cases. The latter is closer to a `MapDataset.copy()`

registerrier (commented):
Thanks @QRemy. I must say the use case of the new function here is not very clear.  It seems to be doing too much at the same time. 

Maybe it should be doing what it says: take a group of IRFs and return the correct geometry, or even maybe only a dictionary containing the axes, bin sizes etc.

gammapy/datasets/map.py
```diff
@@ -169,13 +197,121 @@ def _default_binsz(observation, spatial_bin_size_min=0.01 * u.deg):
...
+    position=None,
+    frame="icrs",
+):
+    """Create a MapDataset, if energy axes, spatial width or bin size are not given
+    they are determined automatically from the IRFs,
+    but the estimated value cannot exceed the given limits.
+
+    Parameters
+    ----------
+    data : `~gammapy.data.Observation` or `~gammapy.data.MapDataset`
```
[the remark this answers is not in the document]
QRemy: The original goal was to support DL4 IRFs because that what we have for HAWC observations, the fact that is works for a MapDataset is a side effect.

gammapy/datasets/map.py
```diff
@@ -169,13 +197,121 @@ def _default_binsz(observation, spatial_bin_size_min=0.01 * u.deg):
 
 def _default_width(observation, spatial_width_max=12 * u.deg):
     # width estimated from the rad_max or the offset_max
-    if observation.rad_max is not None:
+    if hasattr(observation, "rad_max") and observation.rad_max is not None:
```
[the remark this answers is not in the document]
QRemy: maybe we can remove this case completely

gammapy/datasets/map.py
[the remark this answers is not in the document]
QRemy: could remove this from the docstring.

gammapy/datasets/map.py
QRemy: yes I will have to think how to define good default for that in an other PR.

gammapy/datasets/map.py
```diff
@@ -169,13 +197,121 @@ def _default_binsz(observation, spatial_bin_size_min=0.01 * u.deg):
...
+    they are determined automatically from the IRFs,
+    but the estimated value cannot exceed the given limits.
+
+    Parameters
+    ----------
+    data : `~gammapy.data.Observation` or `~gammapy.data.MapDataset`
+        Observation or Dataset containing the IRFs.
+    dataset_name : str, optional
+        If `models` contains one or multiple `FoVBackgroundModel`
+        it should match the `dataset_name` of the background model to use.
```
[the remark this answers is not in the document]
QRemy: remove

gammapy/datasets/map.py
```diff
@@ -169,13 +197,121 @@ def _default_binsz(observation, spatial_bin_size_min=0.01 * u.deg):
...
 def _default_width(observation, spatial_width_max=12 * u.deg):
     # width estimated from the rad_max or the offset_max
-    if observation.rad_max is not None:
+    if hasattr(observation, "rad_max") and observation.rad_max is not None:
         width = 2.0 * np.max(observation.rad_max)
-    else:
+    elif hasattr(observation.psf, "axes"):
         width = 2.0 * observation.psf.axes["offset"].edges[-1]
+    else:
+        width = 2.0 * np.max(observation.psf.psf_map.geom.width)
```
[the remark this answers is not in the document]
QRemy: modified

gammapy/datasets/map.py
QRemy: removed the support for MapDatasest but kept the support for DL4 irfs

gammapy/datasets/map.py
```diff
@@ -169,13 +206,112 @@ def _default_binsz(observation, spatial_bin_size_min=0.01 * u.deg):
 
 def _default_width(observation, spatial_width_max=12 * u.deg):
     # width estimated from the rad_max or the offset_max
-    if observation.rad_max is not None:
-        width = 2.0 * observation.rad_max.quantity.max()
-    else:
+    if isinstance(observation.psf, PSFMap):
+        width = 2.0 * np.max(observation.psf.psf_map.geom.width)
+    elif hasattr(observation.psf, "axes"):
         width = 2.0 * observation.psf.axes["offset"].edges[-1]
```
QRemy, line 212:
```suggestion
    elif hasattr(observation.psf, "axes"):
        width = 2.0 * observation.psf.axes["offset"].edges[-1]
    else:
        width =  spatial_width_max
```

PR #5538, opened 2024-11-04
Add a helper function to create a global dataset from a list of dataset.

QRemy:
Add a helper function to create a global dataset from a list of dataset.
Follow-up of #5533 that introduces  a function `create_empty_map_dataset_from_irfs`.

### DIFF
--- a/gammapy/datasets/utils.py
+++ b/gammapy/datasets/utils.py
@@ -2,7 +2,9 @@
 import numpy as np
 from astropy.coordinates import SkyCoord
 from gammapy.data import Observation
-from gammapy.maps import Map
+import astropy.units as u
+from gammapy.maps import Map, MapAxis, WcsGeom
+from gammapy.maps.utils import _check_width, _check_binsz
 from gammapy.modeling.models.utils import cutout_template_models
 from . import Datasets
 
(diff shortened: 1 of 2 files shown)


gammapy/datasets/utils.py
```diff
@@ -203,3 +205,124 @@ def split_dataset(dataset, width, margin, split_template_models=True):
...
+
+    Parameters
+    ----------
+    datasets : `~gammapy.datasets.Datasets`
+        Datasets
+    name : str, optional
+            Name of the output dataset. Default is None.
+    position : `~astropy.coordinates.SkyCoord`
+        Center position of the output dataset.
+        Defalut is None, and the average postion is taken.
```
registerrier, line 232:
```suggestion
        Default is None, and the average position is taken.
```

gammapy/datasets/utils.py
```diff
@@ -203,3 +205,124 @@ def split_dataset(dataset, width, margin, split_template_models=True):
...
+    nbin_per_decade=None,
+):
+    """Create an empty dataset emcompassing the input datasets.
+
+    Parameters
+    ----------
+    datasets : `~gammapy.datasets.Datasets`
+        Datasets
+    name : str, optional
+            Name of the output dataset. Default is None.
```
registerrier, line 229:
```suggestion
        Name of the output dataset. Default is None.
```

gammapy/datasets/utils.py
```diff
@@ -203,3 +205,124 @@ def split_dataset(dataset, width, margin, split_template_models=True):
...
+        Default is None, the maximum is taken.
+    Returns
+    -------
+    datasets : `~gammapy.datasets.MapDatset`
+        Empy global dataset.
+    """
+    from gammapy.datasets import MapDataset
+
+    if position is None:
+        positions = SkyCoord([d.exposure.geom.center_skydir for d in datasets])
```
registerrier, line 263: Does this work if frames are not equivalent?

gammapy/datasets/utils.py
```diff
@@ -203,3 +205,124 @@ def split_dataset(dataset, width, margin, split_template_models=True):
...
+        position = SkyCoord(positions.cartesian.mean(), frame=positions.frame)
```
registerrier, line 264: For this to work, the `positions` must be in the same frame.

gammapy/datasets/utils.py
registerrier: Why do you use `exposure.geom` rather than `counts.geom`?

gammapy/datasets/utils.py
```diff
@@ -203,3 +205,124 @@ def split_dataset(dataset, width, margin, split_template_models=True):
...
+    energy_true_max_list = []
+    nbin_per_decade_list = []
+    for d in datasets:
+        binsz_list.append(np.abs(d.counts.geom.pixel_scales).min())
+        width_list.append(
+            d.counts.geom.width.max()
+            + 2 * d.counts.geom.center_skydir.separation(position)
+        )
+        energy_min_list.append(d.counts.geom.axes[0].edges.min())
+        energy_max_list.append(d.counts.geom.axes[0].edges.max())
```
registerrier, line 280: It would be safer to call `axes["energy"]` here.

gammapy/datasets/utils.py
```diff
@@ -203,3 +205,124 @@ def split_dataset(dataset, width, margin, split_template_models=True):
...
+        energy_true_min_list.append(d.exposure.geom.axes[0].edges.min())
+        energy_true_max_list.append(d.exposure.geom.axes[0].edges.max())
```
registerrier, line 282: Same comment here

gammapy/datasets/utils.py
```diff
@@ -203,3 +205,124 @@ def split_dataset(dataset, width, margin, split_template_models=True):
...
+    if energy_true_min is None:
+        energy_true_min = np.min(u.Quantity(energy_true_min_list))
+    if energy_true_max is None:
+        energy_true_max = np.max(u.Quantity(energy_true_max_list))
+    if energy_min is None:
+        energy_min = np.min(u.Quantity(energy_min_list))
+    if energy_max is None:
+        energy_max = np.max(u.Quantity(energy_max_list))
+    energy_min = np.maximum(energy_min, energy_true_min)
+    energy_max = np.minimum(energy_max, energy_true_max)
```
registerrier, line 304: Why is this needed?

gammapy/datasets/utils.py
```diff
@@ -203,3 +205,124 @@ def split_dataset(dataset, width, margin, split_template_models=True):
...
+    energy_axis = MapAxis.from_energy_bounds(
+        energy_min, energy_max, nbin_per_decade, unit="TeV", per_decade=True
+    )
+
+    geom = WcsGeom.create(
+        skydir=position.galactic,
+        binsz=binsz,
+        width=width,
+        frame="galactic",
+        proj="CAR",
```
registerrier, line 316: why this choice? Is it always valid?

gammapy/datasets/utils.py
```diff
@@ -203,3 +205,124 @@ def split_dataset(dataset, width, margin, split_template_models=True):
...
+    energy_min = np.maximum(energy_min, energy_true_min)
+    energy_max = np.minimum(energy_max, energy_true_max)
+    if nbin_per_decade is None:
+        nbin_per_decade = np.max(nbin_per_decade_list)
+    energy_axis = MapAxis.from_energy_bounds(
+        energy_min, energy_max, nbin_per_decade, unit="TeV", per_decade=True
+    )
+
+    geom = WcsGeom.create(
+        skydir=position.galactic,
```
registerrier, line 312: why galactic here?

gammapy/datasets/utils.py
```diff
@@ -203,3 +205,124 @@ def split_dataset(dataset, width, margin, split_template_models=True):
...
+    if energy_true_max is None:
+        energy_true_max = np.max(u.Quantity(energy_true_max_list))
+    if energy_min is None:
+        energy_min = np.min(u.Quantity(energy_min_list))
+    if energy_max is None:
+        energy_max = np.max(u.Quantity(energy_max_list))
+    energy_min = np.maximum(energy_min, energy_true_min)
+    energy_max = np.minimum(energy_max, energy_true_max)
+    if nbin_per_decade is None:
+        nbin_per_decade = np.max(nbin_per_decade_list)
```
registerrier, line 306:
```suggestion
    nbin_per_decade = nbin_per_decade if nbin_per_decade is not None else np.max(nbin_per_decade_list)
```
Maybe this usage would reduce a bit the lines of code

gammapy/datasets/utils.py
```diff
@@ -203,3 +205,124 @@ def split_dataset(dataset, width, margin, split_template_models=True):
...
+        nbin_per_decade_list.append(np.ceil(nbins / ndecade))
+
+    if binsz is None:
+        binsz = np.min(u.Quantity(binsz_list))
+    binsz = _check_binsz(binsz)
+    if width is None:
+        width = np.max(u.Quantity(width_list))
+    width = _check_width(width)
+    if energy_true_min is None:
+        energy_true_min = np.min(u.Quantity(energy_true_min_list))
```
registerrier, line 296:
```suggestion
    energy_true_min = energy_true_min if energy_true_min is not None else u.Quantity(energy_true_min_list).min()
```

gammapy/datasets/utils.py
```diff
@@ -203,3 +205,124 @@ def split_dataset(dataset, width, margin, split_template_models=True):
...
+    datasets : `~gammapy.datasets.MapDatset`
+        Empy global dataset.
+    """
+    from gammapy.datasets import MapDataset
+
+    if position is None:
+        positions = SkyCoord([d.exposure.geom.center_skydir for d in datasets])
+        position = SkyCoord(positions.cartesian.mean(), frame=positions.frame)
+
+    binsz_list = []
```
registerrier, line 266: globally isn't it simpler to use list comprehension to build those lists. Is there a case where you want to load the datasets lazily?

registerrier (commented):
Thanks @QRemy. I have left a number of comments. Maybe you could clarify the scope and use case of the function? 
Does it work in all situations? Why do you create a dataset in galactic Plate-Carrée projection in the end?

gammapy/datasets/utils.py
QRemy: Not sure it will really make a real difference in performance.

QRemy:
> Maybe you could clarify the scope and use case of the function?
The long term goal is to determine automatically the dataset geoms from the irfs instead of recommending specific defaults in the configuration files of the hli.
It is intended to be used after  `create_empty_map_dataset_from_irfs` (which create one dataset with a binning optimised for each irf)  in case we want a single dataset to stack all observations.

> Does it work in all situations?
Could be refined later if it breaks for special cases.

> Why do you create a dataset in galactic Plate-Carrée projection in the end?
modified, now it uses the frame and projection of the first dataset in the list

QRemy:
I'll merge this, for now it remains an undocumented helper function that could be refined later.

Issue #5531, opened 2024-10-30
Plotting issues in some tutorials


bkhelifi:
**Gammapy version**
dev

**Bug description**
I have scanned all tutorials to check whether they are any obvious rendering issues. Here are my returns:

- [x]  [spectral_analysis](https://docs.gammapy.org/dev/tutorials/analysis-1d/spectral_analysis.html#a-note-on-statistics): list is not correctly rendered, strange annotation about asymetric Gaussian, strange error for wstat in the first plot for the section. PR: #5534 
- [x]  [spectral_analysis_hli](https://docs.gammapy.org/dev/tutorials/analysis-1d/spectral_analysis_hli.html#exploration-of-the-fit-results): Two mask fit ranges are plotted, and 2 mask safe (but not in residuals as expected) (PR: #5543)
- [x]  Remove the test `print(psf_new == psf_assym)` in https://docs.gammapy.org/dev/tutorials/api/irfs.html#id2? PR: #5537
- [x]  [accessing-contents-of-a-dataset](https://docs.gammapy.org/dev/tutorials/api/datasets.html#accessing-contents-of-a-dataset): Adding c_bar will not hurt PR: #5537
- [x]  [lightcurves]( https://docs.gammapy.org/dev/tutorials/api/catalog.html#lightcurves): x-axis labels are not visible - margin issue. PR: #5561
- [x]  Plot margin issue in the covariance plot (e.g. https://docs.gammapy.org/dev/tutorials/api/fitting.html#sphx-glr-tutorials-api-fitting-py)  PR: #5537
- [x]  missing legends (as feared for the closed PR #5529 ) in https://docs.gammapy.org/dev/tutorials/analysis-3d/flux_profiles.html#plotting-results and for the following plots
    - [x]  the plot of LCs should be fixed to add legends (identical issue than previously mentioned) in ALL timing tutorials that need to be checked once per once
- [x] y-label on flux points disappears on doc rendering (appears correctly when running the script) in [Low Level API](https://docs.gammapy.org/dev/tutorials/starting/analysis_2.html#sphx-glr-tutorials-starting-analysis-2-py). PR: #5584

QRemy:
> * the plot of LCs should be fixed to add legends (identical issue than previously mentioned) in ALL timing tutorials that need to be checked once per once

Not showing the legend by default is consistent with the current behaviour for flux poins.
one can just add plt.legend() with custom options afterward.

QRemy:
[...]

PR #5541, opened 2024-11-05
Make sure FluxPints.plot gets user-defined y-axis units

HealthyPear:
**Description**

When calling FluxPoints.plot() y-axis units are always in erg because the current axis is never taken into account, this PR fixes this behavior.

**Dear reviewer**

ready for review


QRemy:
Thanks @HealthyPear, could you add an unit test for this and check the doc example which is broken by this change.

gammapy/estimators/points/tests/test_core.py
```diff
@@ -241,6 +241,19 @@ def test_plot(self, flux_points):
...
+        ax = fig.add_axes([0.2, 0.2, 0.7, 0.7])
+        ax.xaxis.set_units(u.eV)
+
+        yunit = u.Unit("TeV cm-2 s-1")
+        ax.yaxis.set_units(yunit)
+
+        with mpl_plot_check():
+            ax = flux_points.plot(ax=ax, sed_type="e2dnde")
+
+        assert ax.yaxis.units == yunit
```
QRemy, line 255: this test does not fail on main branch. Is there really a problem setting custom units ?

morcuended:
I think there is still a problem. In the tutorial https://docs.gammapy.org/dev/tutorials/starting/analysis_2.html (dev) the two last spectrum plots do not show the units in the y-axis (unless there is a problem with figsize).

QRemy:
> I think there is still a problem. In the tutorial https://docs.gammapy.org/dev/tutorials/starting/analysis_2.html (dev) the two last spectrum plots do not show the units in the y-axis (unless there is a problem with figsize).

Running the tutorial script I see the ylabel so that's probably a problem rendering the doc page, I added if to the list in #5531 (this is unrelated to this PR).
As the initial problem of this PR cannot be reproduced I'll close it now.

Issue #5544, opened 2024-11-06
ObservationsEventsSampler does not seem to support directly the point-like use case


HealthyPear:
**Is your feature request related to a problem? Please describe.**

Simulate a point-like observation with an event list and not just faking the counts from a real observation dataset.

**Describe the solution you'd like**

Make `ObservationsEventsSampler` work from IRFs which do not contain PSF (i.e. point-like).

**Describe alternatives you've considered**

Simulate from full-enclosure (with the fixes described in #5542 ) and later cut/analyze as point-like, but not sure if this is easier and what are the limitations of this approach.

PR #5545, opened 2024-11-06
Fix ObservationsEventsSampler for observations based on full-enclosure IRFs

### LINKED ISSUE #5542: ObservationEventSampler not working
**Gammapy version**

1.2

**Bug description**

Use of ObservationEventSampler fails because it's attempting to get the maximum value from the class, not from its data

**Expected behavior**

According to the source code it should be compatible with observations providing RadMax2D (otherwise there would be no reason to look for a maximum value)

**To Reproduce**

```python
from gammapy.data import ObservationsEventsSampler
observation = # any pointlike observation from a data store
model = # any sky model
obs_evt_sampler.simulate_observation(observation, models=Models(models=model))
```

```
TypeError                                 Traceback (most recent call last)
Cell In[14], line 1
----> 1 obs_evt_sampler.simulate_observation(observations[0], models=Models(models=model))

File [~/Software/gammapy/gammapy/gammapy/data/simulate.py:69](http://localhost:8888/lab/tree/~/Software/gammapy/gammapy/gammapy/data/simulate.py#line=68), in ObservationsEventsSampler.simulate_observation(self, observation, models)
     64 from gammapy.datasets import ObservationEventSampler
     66 sampler = ObservationEventSampler(
     67     **self.sampler_kwargs, dataset_kwargs=self.dataset_kwargs
     68 )
---> 69 observation = sampler.run(observation, models=models)
     71 if self.outdir is not None:
     72     observation.write(
     73         self.outdir [/](http://localhost:8888/) f"obs_{observation.obs_id}.fits",
     74         overwrite=self.overwrite,
     75     )

File [~/Software/gammapy/gammapy/gammapy/datasets/simulate.py:681](http://localhost:8888/lab/tree/~/Software/gammapy/gammapy/gammapy/datasets/simulate.py#line=680), in ObservationEventSampler.run(self, observation, models, dataset_name)
    657 def run(self, observation, models=None, dataset_name=None):
    658     """Sample events for given observation and signal models.
    659 
    660     The signal distribution is sampled from the given models
   (...)
    678         A copy of the input observation with event list filled.
    679     """
--> 681     dataset = create_map_dataset_from_observation(
    682         observation, models, dataset_name, **self.dataset_kwargs
    683     )
    684     events = super().run(dataset, observation)
    686     observation = deepcopy(observation)

File [~/Software/gammapy/gammapy/gammapy/datasets/map.py:228](http://localhost:8888/lab/tree/~/Software/gammapy/gammapy/gammapy/datasets/map.py#line=227), in create_map_dataset_from_observation(observation, models, dataset_name, energy_axis_true, energy_axis, energy_bin_per_decade_max, spatial_width, spatial_width_max, spatial_bin_size, spatial_bin_size_min)
    225 from gammapy.makers import MapDatasetMaker
    227 if spatial_width is None:
--> 228     spatial_width = _default_width(observation, spatial_width_max)
    229 if spatial_bin_size is None:
    230     spatial_bin_size = _default_binsz(observation, spatial_bin_size_min)

File [~/Software/gammapy/gammapy/gammapy/datasets/map.py:173](http://localhost:8888/lab/tree/~/Software/gammapy/gammapy/gammapy/datasets/map.py#line=172), in _default_width(observation, spatial_width_max)
    170 def _default_width(observation, spatial_width_max=12 * u.deg):
    171     # width estimated from the rad_max or the offset_max
    172     if observation.rad_max is not None:
--> 173         width = 2.0 * np.max(observation.rad_max)
    174     else:
    175         width = 2.0 * observation.psf.axes["offset"].edges[-1]

TypeError: unsupported operand type(s) for *: 'float' and 'RadMax2D'
```

HealthyPear:
**Description**

This pull request fixes two small bugs that are preventing `gammapy.data.ObservationsEventsSampler` to work on full-enclosure-based observations (for the point-like case see #5544 ).

I can add a new tutorial with a short snippet on how to use this class (either as a separate PR after this fix PR is merged or directly here if preferred by the maintainers).

Regarding the testing (to be integrated):
- the fix for `dataset_kwargs` is needed because in the test this is always defined, never `None`
- the energy axes fix is needed because of the fix above (if the axis is None when `create_map_dataset_from_observation` is called then the (faulty) function `_default_energy_axis` is called

Closes #5542.

### DIFF
--- a/gammapy/datasets/simulate.py
+++ b/gammapy/datasets/simulate.py
@@ -647,7 +647,7 @@ class ObservationEventSampler(MapDatasetEventSampler):
         n_event_bunch=10000,
         dataset_kwargs=None,
     ):
-        self.dataset_kwargs = dataset_kwargs if dataset_kwargs else {}
+        self.dataset_kwargs = dataset_kwargs or {}
         self.random_state = get_random_state(random_state)
         self.oversample_energy_factor = oversample_energy_factor
         self.t_delta = t_delta
--- a/gammapy/datasets/simulate.py
+++ b/gammapy/datasets/simulate.py
@@ -647,7 +647,7 @@ class ObservationEventSampler(MapDatasetEventSampler):
         n_event_bunch=10000,
         dataset_kwargs=None,
     ):
-        self.dataset_kwargs = dataset_kwargs
+        self.dataset_kwargs = dataset_kwargs if dataset_kwargs else {}
         self.random_state = get_random_state(random_state)
         self.oversample_energy_factor = oversample_energy_factor
         self.t_delta = t_delta
--- a/gammapy/datasets/map.py
+++ b/gammapy/datasets/map.py
@@ -125,7 +125,7 @@ def create_map_dataset_geoms(
 def _default_energy_axis(observation, energy_bin_per_decade_max=30):
     # number of bins per decade estimated from the energy resolution
     # such as diff(ereco.edges)/ereco.center ~ min(eres)
-    etrue = observation.psf.axes[0].edges  # only where psf is defined
+    etrue = observation.psf.axes[0].center  # only where psf is defined
     eres = observation.edisp.to_edisp_kernel(0 * u.deg).get_resolution(etrue)
     eres = eres[np.isfinite(eres)]
     if eres.size > 0:
--- a/gammapy/datasets/map.py
+++ b/gammapy/datasets/map.py
@@ -125,8 +125,8 @@ def create_map_dataset_geoms(
 def _default_energy_axis(observation, energy_bin_per_decade_max=30):
     # number of bins per decade estimated from the energy resolution
     # such as diff(ereco.edges)/ereco.center ~ min(eres)
-    etrue = observation.psf.axes[0].center  # only where psf is defined
-    eres = observation.edisp.to_edisp_kernel(0 * u.deg).get_resolution(etrue)
+    etrue = observation.psf.axes[0]  # only where psf is defined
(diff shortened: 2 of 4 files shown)


HealthyPear:
thanks! I'll add the case with no dataset kwargs in ObservationsEventsSampler

QRemy:
> I can add a new tutorial with a short snippet on how to use this class (either as a separate PR after this fix PR is merged or directly here if preferred by the maintainers).

Please open a separate PR for this, will try to have this one merged fast and back-ported.

HealthyPear:
can I add myself to the mailmap in this PR?

HealthyPear:
I think the workflows should be reloaded as the failure appears to be something related to the remote runner and not to the changes introduced here. I do not seem to have enough permissions to do it myself.

QRemy:
The test of defaults CTA was taking too much time and memory to compute the EdispMap (shape 500x500x100x100)  so it was cancelled, using HESS observation instead is fine.

gammapy/datasets/simulate.py
```diff
@@ -647,7 +647,7 @@ def __init__(
         n_event_bunch=10000,
         dataset_kwargs=None,
     ):
-        self.dataset_kwargs = dataset_kwargs
+        self.dataset_kwargs = dataset_kwargs if dataset_kwargs else {}
```
Astro-Kirsty, line 650: Can we simply utilise the following:
```suggestion
        self.dataset_kwargs = dataset_kwargs or {}
```

PR #5548, opened 2024-11-07
Remove is_norm in parameter.from_dict for backward compatibility

QRemy:
Remove is_norm in parameter.from_dict for backward compatibility


registerrier (approved):
Thanks @QRemy . Well spotted!

I think we need different models files created since 1.0 to test and make sure we can read them in CI.

Issue #5562, opened 2024-11-12
Make `_build_parameters_from_dict` public


MRegeard:
`_build_parameters_from_dict` is mandatory to deserialize model. 
So it should be made public in order to create custom models.

registerrier:
This should be discussed as part model refactoring, see #5668

QRemy:
Closing now see : 
https://github.com/gammapy/gammapy/pull/5563#issuecomment-4573974139

PR #5563, opened 2024-11-12
make build_parameters public

MRegeard:
This PR solves #5562.

### DIFF
--- a/gammapy/modeling/models/__init__.py
+++ b/gammapy/modeling/models/__init__.py
@@ -1,7 +1,8 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 """Built-in models in Gammapy."""
+
 from gammapy.utils.registry import Registry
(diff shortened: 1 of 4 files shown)


MRegeard:
@registerrier, this is needed if you have custom models that are not `Spectral`, `Temporal` or `Spatial`. To be able to read them from file you'll need to implement a `from_dict` method which uses this function. 
For instance [here](https://github.com/MRegeard/okkie/blob/6b770887c27719620a898f34d0796755df3a0f97/okkie/modeling/models/phase.py#L60).

registerrier (commented):
Thanks @MRegeard . Should we see how the discussions on the modeling PIG go, here?
I am not sur I understand what we gain by making this function public in general.

QRemy:
As such It's very unlikely that people understand how to use this function for custom model serialisation without being already familiar with the model serialisation source code in gammapy. So I think it's not a sufficient solution to help users. Closing until new proposal.

PR #5566, opened 2024-11-12
Scheduled build

MRegeard:
This PR adds a github action that will run ci.yml everyday at 7 am UTC.

### DIFF
--- a/.github/workflows/scheduled_build.yml
+++ b/.github/workflows/scheduled_build.yml
@@ -1,0 +1,20 @@
+name: Scheduled build
+
+on:
+  # Allow manual runs through the web UI
+  workflow_dispatch:
+  schedule:
+    #        ┌───────── minute (0 - 59)
+    #        │ ┌───────── hour (0 - 23)
+    #        │ │ ┌───────── day of the month (1 - 31)
+    #        │ │ │ ┌───────── month (1 - 12 or JAN-DEC)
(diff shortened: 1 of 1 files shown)


registerrier (commented):
Thanks @MRegeard . We can indeed re-think how the CI matrix is performed.

Is running the full CI matrix on daily basis on top of the pull request CI completely useful?

MRegeard:
@registerrier, yes we should probably rework the CI matrix. 
We could use github action workflow (https://github-actions-workflows.openastronomy.org/en/stable/tox.html) that produces CI runs based on tox environment. It is basically what we already have but more straight forward. 
See sunpy repo for usage. 
Anyway this is definitely for 2.0.

bkhelifi (approved):
It is very useful to track external libraries changes. Thanks a lot.

registerrier (commented):
Thanks @MRegeard . We have to see if we really need  scheduled CI run. I think it could be useful to have a full doc build + linkcheck as scheduled run.
It would allow removing linkcheck on a regular build
Adding failure on linkcheck errors would enforce rigorous checking  which we don't do now.

morcuended:
Other use case for this scheduled CI might be the check of conda build #5666

registerrier:
This has been implemented in several new workflows. Closing now.

PR #5569, opened 2024-11-15
Add legend in correct place and remove calls to plt.legend()

Astro-Kirsty:
This allows the legends to be automatically plotted but with the correct call to `ax.legend`.

Related to #5510 which removed the legends completely. 

I also remove the calls to `plt.legend` in two other cases.

### DIFF
--- a/examples/tutorials/analysis-3d/flux_profiles.py
+++ b/examples/tutorials/analysis-3d/flux_profiles.py
@@ -203,6 +203,7 @@ for quantity in quantities:
     profile[quantity].plot(ax=ax, label=quantity.title())
 
 ax.set_ylabel("Counts")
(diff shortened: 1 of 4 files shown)


morcuended (approved):
Thanks @Astro-Kirsty. I just tested this with the example in #5510 and in principle works as expected. Looks good to me.

bkhelifi:
Hard to see the rendering or to know whether other codes have not the same issue or if this fix does not break other plots.

We start to optimise the plots, but we have no test on their content, their stability after code changes. I do remind that this fix is a consequence of a recent PR. And I don't know if it is technically possible to automise that or if it can be only by-eye. 
A though to keep in mind beyond this fix...

PR #5581, opened 2024-11-18
Adapt release instructions

Astro-Kirsty:
This PR fixes some broken links in the release procedure. 

Personally, I think we should update this page such that we just include all instructions that we follow. There are certain steps that we do not follow from astropy, and it makes sense to have our own complete instructions. 
I am happy to implement this other are okay with it, as I took extensive notes in the candidate release meeting on Friday!


registerrier (approved):
Thanks @Astro-Kirsty

PR #5582, opened 2024-11-18
Change css and conf to allow for smaller screens

### LINKED ISSUE #5196: Docs display on small window sizes broken
**Gammapy version**

Current main.

**Bug description**

The documentation does not behave correctly on window size transitions anymore.

Comparing current dev docs with the Gammapy 1.2 docs:

### Gammapy 1.2
old_small_gammapy

### Gammapy dev

gammapy_new_small

**Expected behavior**
Correctly adapt content to window size

**To Reproduce**

Visit dev docs, decrease window width

Astro-Kirsty:
This makes some changes to fixes #5196
I also adjust the number of headers along the top

In the end utilising the default works well, however, it causes other issues! :( 
Actually our css is quite complicated (and it should not be!), we can work on that further for v2.0

### DIFF
--- a/docs/_static/custom.css
+++ b/docs/_static/custom.css
@@ -111,6 +111,10 @@ p {
   max-width: 77.25%;
 }
 
+.bd-sidebar, .bd-toc {
+  position: relative;
+}
+
 /* pre */
 .pre {
   font-size: 12px;
--- a/docs/_static/custom.css
+++ b/docs/_static/custom.css
@@ -437,8 +437,8 @@ a.dropdown-item {
 }
 
 @media screen and (max-width: 768px) {
-    .pysidebar {
-        display: none;  /* Hides the sidebar */
+    .secondary-sidebar {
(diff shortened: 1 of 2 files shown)


Astro-Kirsty:
This is a temporary fix. As shown here it works but the secondary sidebar is still visible. 
(It doesn't however overlap with the content which is good)
Screenshot from 2024-11-18 15-52-27

image

Astro-Kirsty:
Screenshot from 2024-11-18 15-52-27
Screenshot from 2024-11-18 15-52-33

bkhelifi:
It looks rather good now, is not it? If there is remaining issue, you can add a small issue as reminder.
Thanks a lot @Astro-Kirsty !!!

bkhelifi (approved):
Thanks a million

registerrier (approved):
Thanks @Astro-Kirsty . Looks good. Let's have a look at the dev docs.

PR #5588, opened 2024-11-20
Updates to documentation conf and css

### LINKED ISSUE #5486: Documentation and sphinx-astropy
One should investigate whether one can remove our dependency from `sphinx-astropy`, This would allow using the same libraries for the code and the documentation

Astro-Kirsty:
This PR is to start the update of the documentation.

I have been able to remove sphinx-astropy, which resolves #5486. This mainly involved updating the `extensions` and `intersphinx_mapping`. I have tested the `conf` a number of times to check that we have all the necessary elements. 

I started to adapt the `css` but this needs further work and a future dedicated discussion on how we want to proceed. If we wish to keep our current theme or not etc. 

Note: There are some nice additions we could think about using:
https://pydata-sphinx-theme.readthedocs.io/en/stable/examples/kitchen-sink/admonitions.html

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -80,8 +80,6 @@ test =
     sphinx
 docs =
     astropy
-    numpy
-    sherpa
     numpydoc
     sphinx
     sphinx_automodapi
--- a/setup.cfg
+++ b/setup.cfg
@@ -80,8 +80,8 @@ test =
     sphinx
 docs =
     astropy
-    numpy<2.0
-    sherpa<4.17
+    numpy
+    sherpa
     numpydoc
     sphinx
     sphinx_automodapi
--- a/setup.cfg
+++ b/setup.cfg
@@ -79,15 +79,16 @@ test =
     docutils
     sphinx
 docs =
-    astropy>=5.0,<5.3
+    astropy
     numpy<2.0
     sherpa<4.17
-    sphinx-astropy
+    numpydoc
     sphinx
+    sphinx_automodapi
     sphinx-click
-    sphinx-gallery
-    sphinx-design
     sphinx-copybutton
+    sphinx-design
+    sphinx-gallery
     pydata-sphinx-theme
     nbformat
     docutils
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -147,7 +147,7 @@ jobs:
           gammapy download datasets
       - name: test build docs
         run: |
-          tox -e build_docs -- -j auto
+          tox -e build_docs
       - name: check links
         continue-on-error: true
         run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -119,7 +119,7 @@ jobs:
           file: ./coverage.xml
           verbose: true
   sphinx:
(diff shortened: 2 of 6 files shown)


setup.cfg
```diff
@@ -79,15 +79,16 @@ test =
     docutils
     sphinx
 docs =
-    astropy>=5.0,<5.3
+    astropy
```
bkhelifi, line 82: Why changing the astropy version here?

Astro-Kirsty: We needed these specific conditions as we were using `sphinx-astropy` but now that we do not rely on that we can use the most up to date version!

bkhelifi: The failing test is not associated to that? to be checked ;)

registerrier:
> Having a temporary building of the doc would help so much to understand the result

I have modified the CI workflow to include a step uploading the docs build result to a gammapy-docs-html.zip file. Let's see if this works...

setup.cfg
```diff
@@ -79,15 +79,16 @@ test =
...
     numpy<2.0
     sherpa<4.17
```
registerrier, line 84:
```suggestion
    numpy
    sherpa
```

setup.cfg
registerrier: I think the issue comes from the constraint on `numpy<2.0` as the error message is connected with multiprocessing. I have removed the constraint. Let's see.

setup.cfg
```diff
@@ -79,15 +79,16 @@ test =
     docutils
     sphinx
 docs =
-    astropy>=5.0,<5.3
-    numpy<2.0
-    sherpa<4.17
-    sphinx-astropy
+    astropy
+    numpy
+    sherpa
```
registerrier, line 84:
```suggestion
```

morcuended:
> I have modified the CI workflow to include a step uploading the docs build result to a gammapy-docs-html.zip file. Let's see if this works...

I think this did not work yet, right? The job in charge of uploading the artifact was failing before

registerrier:
> > I have modified the CI workflow to include a step uploading the docs build result to a gammapy-docs-html.zip file. Let's see if this works...
> 
> I think this did not work yet, right? The job in charge of uploading the artifact was failing before

It finally worked after I removed the multiprocessing run of sphinx...

See https://github.com/Astro-Kirsty/gammapy/actions/runs/11969784425/artifacts/2223426845

registerrier:
Note that now there is a CI because of the release of astropy 7.0 which occurred 1h ago :exploding_head:

morcuended:
Artifacts are uploaded to the Actions of the forked repository from which the PR is created. It would be good to serve this deployed documentation on the fly providing the URL to access it maybe using GH Pages

morcuended:
HTML preview looks very good btw!

AtreyeeS (approved):
Yes lets merge this and proceed in further PRs

registerrier (approved):
Thanks @Astro-Kirsty . LEt's merge. 

There is still a CI issue, so we have to impose matplotlib<3.10 for now and try to solve the issues after that. Lets' merge #5659 first.

PR #5589, opened 2024-11-20
Fix select_nested_models docstring fit

Astro-Kirsty:
This is to update the `select_nested_models` docstring which does not specify what the `Fit` object is by default.

I also made the links in the CTAO tutorial more clear


bkhelifi (approved):

MRegeard (approved):
Thanks @Astro-Kirsty.

PR #5590, opened 2024-11-20
Fix missing errors after using select_nested_models

QRemy:
Cahche also the covariacne matrix otherwise errors are not restored ( might have the same problem with restore_status).

### DIFF
--- a/gammapy/modeling/selection.py
+++ b/gammapy/modeling/selection.py
@@ -72,9 +72,9 @@ class TestStatisticNested:
          (for example considereing diffuse background or nearby source).
         """
         stat = datasets.stat_sum()
-        object_cache, prev_pars, prev_covar = self._apply_null_hypothesis(datasets)
+        cache = self._apply_null_hypothesis(datasets)
         stat_null = datasets.stat_sum()
-        self._restore_status(datasets, object_cache, prev_pars, prev_covar)
+        self._restore_status(datasets, cache)
         return stat_null - stat
 
     def ts_asimov(self, datasets):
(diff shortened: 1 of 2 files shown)


bkhelifi:
Can you add the associated test to validate the identified issues?

QRemy:
Tests added  but I did not test for linked models yet but we can go without it for now as usage is limited.

QRemy:
Should be ok for linked params (but the test is probably not the best)

gammapy/modeling/selection.py
```diff
@@ -143,31 +143,38 @@ def run(self, datasets, apply_selection=True):
...
+        cache["values"] = [p.value for p in datasets.models.parameters]
+        cache["error"] = [p.error for p in datasets.models.parameters]
         for p, val in zip(self.parameters, self.null_values):
             if isinstance(val, Parameter):
                 p.__dict__ = val.__dict__
             else:
                 p.value = val
-                p.frozen = True
-        return object_cache, prev_pars
+        p.frozen = True
```
registerrier, line 163: is it the correct indentation?

gammapy/modeling/tests/test_selection.py
```diff
@@ -65,12 +69,16 @@ def test_test_statistic_detection_other_frozen(fermi_datasets):
...
     # TODO: better test with simulated data ?
-    model = fermi_datasets.models["Crab Nebula"]
+
+    models = Models.read("$GAMMAPY_DATA/fermi-3fhl-crab/Fermi-LAT-3FHL_models.yaml")
+
+    model = models["Crab Nebula"]
     model2 = model.copy(name="other")
     model2.spectral_model.alpha.value = 2.4
-    fermi_datasets.models = fermi_datasets.models + [model2]
+    model2.spectral_model.alpha.error = 1
```
registerrier, line 79: why do you set the error manually?

registerrier (commented):
Thanks @QRemy . Looks good. I have left two minor comments inline.

gammapy/modeling/tests/test_selection.py
QRemy: not necessary, removed

gammapy/modeling/selection.py
QRemy: fixed

PR #5631, opened 2024-12-06
Clear cache immediately after npred_signal computation If USE_NPRED_CACHE = False

QRemy:
Currently when USE_NPRED_CACHE is false the cache is reset only before new computations.
In this PR the cache is cleared immediately after npred_signal computation. This is particularly useful for simulation in which npred is computed only once for many sources so caching is not necessary and can be very memory consuming

### DIFF
--- a/gammapy/datasets/evaluator.py
+++ b/gammapy/datasets/evaluator.py
@@ -103,6 +103,8 @@ class MapEvaluator:
         """Reset cached properties."""
         del self._compute_npred
         del self._compute_flux_spatial
+        self._computation_cache = None
+        self._cached_parameter_previous = None
 
     @property
     def geom(self):
@@ -239,8 +241,6 @@ class MapEvaluator:
                 )
 
         self.reset_cache_properties()
-        self._computation_cache = None
-        self._cached_parameter_previous = None
 
     @lazyproperty
     def _edisp_diagonal(self):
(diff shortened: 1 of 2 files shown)


gammapy/datasets/tests/test_map.py
```diff
@@ -1166,6 +1166,58 @@ def test_npred(sky_model, geom, geom_etrue):
...
+    assert_allclose(dataset.npred_background().data.sum(), 4000.0, rtol=1e-3)
+    assert_allclose(dataset._background_cached.data.sum(), 4000.0, rtol=1e-3)
+
+    assert_allclose(dataset.npred().data.sum(), 9676.047906, rtol=1e-3)
+    assert_allclose(dataset.npred_signal().data.sum(), 5676.04790, rtol=1e-3)
+    assert_allclose(
+        dataset.npred_signal(model_names=[model1.name, sky_model.name]).data.sum(),
+        5676.04790,
+        rtol=1e-3,
+    )
```
registerrier, line 1206: I am not sure I completely follow the point of the test. Isn't it enough to call `npred()` and then check that the cached maps in the evaluators are correct?

registerrier (commented):

Maybe the test could be simplified as it is not very clear why some asserts are needed. Could you clarify?

gammapy/datasets/tests/test_map.py
QRemy: yes simplified

PR #5632, opened 2024-12-06
Improve DL4 irfs support in gammapy.makers

QRemy:
Improve DL4 IRFs support in gammapy.makers implementing these features:
- Support for background maps given as a rate not only counts.
- Support for aeff and background given per transit.
- Add n_transits property in gti.

This is required to use SWGO IRFs.

### DIFF
--- a/gammapy/data/gti.py
+++ b/gammapy/data/gti.py
@@ -249,6 +249,11 @@ class GTI:
         return self._meta.reference_time
 
     @property
+    def time_sum(self):
+        """Sum of GTIs in seconds as a `~astropy.units.Quantity`."""
+        return self.time_delta.sum()
+
+    @property
     def time_start(self):
         """GTI start times as a `~astropy.time.Time` object."""
         return self.table["START"]
@@ -275,16 +280,6 @@ class GTI:
             (t_start, t_stop)
             for t_start, t_stop in zip(self.time_start, self.time_stop)
         ]
-
-    @property
-    def time_sum(self):
-        """Sum of GTIs in seconds as a `~astropy.units.Quantity`."""
-        return self.time_delta.sum()
-
-    @property
-    def n_transits(self):
-        with u.add_enabled_units([u.def_unit("transit", u.sday)]):
-            return self.gti.time_sum.to("transit")
 
     @classmethod
     def from_time_intervals(cls, time_intervals, reference_time=None):
--- a/gammapy/data/gti.py
+++ b/gammapy/data/gti.py
@@ -249,11 +249,6 @@ class GTI:
         return self._meta.reference_time
 
     @property
-    def time_sum(self):
-        """Sum of GTIs in seconds as a `~astropy.units.Quantity`."""
-        return self.time_delta.sum()
-
-    @property
     def time_start(self):
         """GTI start times as a `~astropy.time.Time` object."""
         return self.table["START"]
@@ -280,6 +275,16 @@ class GTI:
             (t_start, t_stop)
             for t_start, t_stop in zip(self.time_start, self.time_stop)
         ]
+
+    @property
+    def time_sum(self):
+        """Sum of GTIs in seconds as a `~astropy.units.Quantity`."""
+        return self.time_delta.sum()
+
+    @property
+    def n_transits(self):
+        with u.add_enabled_units([u.def_unit("transit", u.sday)]):
+            return self.gti.time_sum.to("transit")
 
     @classmethod
     def from_time_intervals(cls, time_intervals, reference_time=None):
--- a/gammapy/data/observations.py
+++ b/gammapy/data/observations.py
@@ -194,6 +194,17 @@ class Observation:
         gti = self.obs_filter.filter_gti(self._gti)
         return gti
 
+    @gti.setter
+    def gti(self, value):
+        if not isinstance(value, GTI):
+            raise TypeError(f"GTI must be an GTI instance, got: {type(value)}")
+        self._gti = value
+
+    @property
+    def n_transits(self):
+        with u.add_enabled_units([u.def_unit("transit", u.sday)]):
+            return self.gti.time_sum.to("transit")
+
     @staticmethod
     def _get_obs_info(
         pointing, deadtime_fraction, time_start, time_stop, reference_time, location
--- a/gammapy/data/observations.py
+++ b/gammapy/data/observations.py
@@ -196,6 +196,8 @@ class Observation:
 
     @gti.setter
     def gti(self, value):
+        if self.gti is not None and self.events is not None:
+            raise TypeError("GTI cannot be redefined if events is not None.")
         if not isinstance(value, GTI):
             raise TypeError(f"GTI must be an GTI instance, got: {type(value)}")
         self._gti = value
--- a/gammapy/data/observations.py
+++ b/gammapy/data/observations.py
@@ -197,7 +197,7 @@ class Observation:
     @gti.setter
     def gti(self, value):
         if self.gti is not None and self.events is not None:
-            raise TypeError("GTI cannot be redefined if events is not None.")
+            raise ValueError("GTI cannot be redefined if events is not None.")
         if not isinstance(value, GTI):
             raise TypeError(f"GTI must be an GTI instance, got: {type(value)}")
         self._gti = value
--- a/gammapy/data/observations.py
+++ b/gammapy/data/observations.py
@@ -200,11 +200,6 @@ class Observation:
             raise TypeError(f"GTI must be an GTI instance, got: {type(value)}")
         self._gti = value
 
-    @property
-    def n_transits(self):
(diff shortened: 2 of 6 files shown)


gammapy/data/gti.py
```diff
@@ -281,6 +276,16 @@ def time_intervals(self):
...
 
+    @property
+    def time_sum(self):
+        """Sum of GTIs in seconds as a `~astropy.units.Quantity`."""
+        return self.time_delta.sum()
+
+    @property
+    def n_transits(self):
+        with u.add_enabled_units([u.def_unit("transit", u.sday)]):
+            return self.gti.time_sum.to("transit")
```
QRemy, line 287:
```suggestion
            return self.time_sum.to("transit")
```

adonath:
Thanks @QRemy, it would be great if @LauraOlivera could review!

LauraOlivera (dismissed):
It looks good to me! I left an inline comment but I think the current implementation is fine as is. We should then build on this by implementing a function that derives `n_transits` as a function of R.A. (which is more time consuming...)

Thanks a lot @QRemy

gammapy/makers/map.py
```diff
@@ -166,9 +167,17 @@ def make_exposure(geom, observation, use_region_center=True):
             Exposure map.
         """
         if isinstance(observation.aeff, Map):
-            return observation.aeff.interp_to_geom(
-                geom=geom,
-            )
+            aeff = observation.aeff
+            if aeff.unit.is_equivalent(u.Unit("m2 s")):
+                factor = 1.0
+            elif aeff.unit.is_equivalent(u.Unit("m2 s sday-1")):
```
registerrier, line 173: This does not guarantee that the aeff map will be properly normalized as exposure per transit as any map with a area unit will comply.
Maybe a meta information could be used instead? Or a dedicated IRF object

registerrier (commented):
Thanks @Qremy and @LauraOlivera .

Having `n_transit` as a property on `GTI` is confusing, as it applies only to specific situations and is rather a time unit conversion. I would rather have a new unit defined and perform a time conversion (`GTI.time_sum.to("transit")`). But if the unit is the same as `day` maybe it is even not really useful no?

Also, the choice of having the exposure per transit stored as a simple `Map` does not really guarantee that you can properly handle it, no? 
The unit is equivalent to a surface which does not really tell you that this is per transit. Isn't it better to have a dedicated IRF? Storing effective are as a function of zenith angle is not an option here?

gammapy/makers/map.py
QRemy: now it also checks that "transit" is in the unit

QRemy:
Test added now it requires https://github.com/gammapy/gammapy-data/pull/66.

gammapy/data/observations.py
```diff
@@ -194,6 +194,12 @@ def gti(self):
         gti = self.obs_filter.filter_gti(self._gti)
         return gti
 
+    @gti.setter
+    def gti(self, value):
```
bkhelifi, line 198: At the GADF DL3, the data should contain the GTI (see https://gamma-astro-data-formats.readthedocs.io/en/v0.3/general/hduclass.html). However, a science analysis tool can change the GTI for its internal purpose (see 
https://gamma-astro-data-formats.readthedocs.io/en/v0.3/events/gti.html).
So, having a setter makes sense.
However:
1. one should be sure very cautious to not araise existing GTI (for provenance)
2. one should create our own Gammapy time interval (or a `UserTimeInterval`) in order to not mix the concepts. This becomes more and more important.

I propose here to add in the code a test forbidding to erase existing DL3 GTI

gammapy/maps/core.py
```diff
@@ -22,6 +22,8 @@
 
 __all__ = ["Map"]
 
+u.add_enabled_units([u.def_unit("transit", u.sday)])
```
bkhelifi, line 25: We are creating here an unit, that is not listed as an international unit. Is it wise? Is it really the proper way to do it?
A full documentation is needed into Gammapy in case we accept to create a standard

bkhelifi (changes requested):
Thank you Quentin. I put some comments to be discussed and some improvement proposal.

gammapy/data/gti.py
```diff
@@ -281,6 +276,16 @@ def time_intervals(self):
...
 
+    @property
+    def time_sum(self):
+        """Sum of GTIs in seconds as a `~astropy.units.Quantity`."""
+        return self.time_delta.sum()
+
+    @property
+    def n_transits(self):
+        with u.add_enabled_units([u.def_unit("transit", u.sday)]):
+            return self.time_sum.to("transit")
```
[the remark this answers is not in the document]
QRemy: > This is the only part that makes me slightly nervous, because the definition of a transit as a unit is not very well defined for fractional transits (generally non-integer n_transits < 10 I would say).

This is for transient right ? In that case I think the irf should not be given per transit but integrated over the specific time range they have been derived for.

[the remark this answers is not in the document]
QRemy: > This equivalence of total time in GTI = number of transits observed is only true if the exposure is flat in right ascension.
> For the current HAWC and SWGO analyses, this is at the moment not a concern because in HAWC we select our data so that n_transits is constant in R.A., and since SWGO doesn't exist, we assume that is also the case.

I guess we can check that there is only one RA bin in the irfs and rise a `NotImplementedError` if that's not the case

gammapy/maps/core.py
QRemy: The problem is that the irfs for simulations are typically defined per transit while for real data they integrated over the full livetime.
@LauraOlivera prefers to not use sday explicitly in the unit, I'll let her explain why.

LauraOlivera: The DL4 IRFs are integrated over one transit (for the simplest analysis use case). A transit is defined through a zenith threshold (which should be part of the metadata), and it corresponds to the path a source does over the observatory from when it rises above the zenith threshold to when it goes below it.

In the span of 1 sidereal day, all visible sources will complete their transits. But the amount of *time* that one transit corresponds to depends on what is the declination of the source. Some sources will spend a minute above the zenith threshold, some others will spend the maximum amount (which is ~6h for a 45º zenith cut, so much less than 23.9h).

So it would be incorrect to say that the exposure is 1 sday, because this corresponds to 23.9h ,but the real exposure depends on the declination. So the correct unit is one transit.

If one has >>10 transits in the data and relatively flat exposure (which is common for WCDs), it is safe to assume that N sday in the exposure correspond to N transits, like Quentin implemented. This is not true for short exposures of course (which is what I mentioned above)

bkhelifi: Personally, I have no issue to "define" what is a transit. Fermi has the notion of "orbit", is not it?
If we use it, one should define it properly into the documentation, e.g.:
- the data reduction user guide
- the glossary

gammapy/data/observations.py
QRemy: Maybe we can raise an error only if a GTI and the events exits because for simulation we can use flexibility in redefining the GTI.

bkhelifi: Indeed, when both exist, a error could be raised. This guaranties to not detroy DL3 data

gammapy/maps/core.py
QRemy: glossary :  done #5735
the data reduction user guide : it's very light so I don't see how it would fit in there, I will suggest to have that in the HAWC or SWGO notebook instead in a separate PR.

gammapy/data/observations.py
QRemy: done

gammapy/makers/map.py
```diff
@@ -170,10 +171,22 @@ def make_exposure(geom, observation, use_region_center=True):
...
-                geom=geom,
-            )
+        with u.add_enabled_units([u.def_unit("transit", u.sday)]):
+            if isinstance(observation.aeff, Map):
+                aeff = observation.aeff
+                if aeff.unit.is_equivalent(u.Unit("m2 s")):
+                    factor = 1.0
+                elif aeff.unit.is_equivalent(u.Unit("m2 s transit-1")):
+                    if u.Unit("transit") in aeff.unit.bases:
+                        factor = observation.gti.time_sum.to("transit")
```
LauraOlivera, line 181: maybe put the TODO here too so that it doesn't slip past?

LauraOlivera (dismissed):
Thanks, @QRemy, looks good to me besides something very minor

gammapy/data/gti.py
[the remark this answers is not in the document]
QRemy: added TODO

gammapy/makers/map.py
QRemy: added TODO

QRemy:
All commets are implemented, merging now.

PR #5633, opened 2024-12-06
Generic mask creation on `Geom`

### LINKED ISSUE #5630: Method to create mask over any axes on `Geom`
Currently, we can create an energy_mask from a `Geom` object using `Geom.energy_mask`. 
We should extend this feature to any axes by passing the axes name and the min and max value desired.

MRegeard:
This PR resolves #5630.

It adds a method that takes an axis name and edges and return a map representing a mask defined by `edge_min` and `edge_max`. 
It also adds support for mutli-(more than 3D)-dimensional geometry. 

Question is wether we should keep `energy_mask` as it is calling this method and the API is straight forward. 
Should we expose this method in the mask map tutorial ?


codemeta.json
```diff
@@ -359,7 +359,7 @@
         "active"
     ],
     "email": "GAMMAPY-COORDINATION-L@IN2P3.FR",
-    "dateModified": "2024-11-26",
+    "dateModified": "2024-12-06",
```
QRemy, line 362: seems unrelated could you remove that

codemeta.json
```diff
@@ -371,4 +371,4 @@
         "iminuit>=2.8.0",
         "matplotlib>=3.4"
     ]
-}
\ No newline at end of file
+}
```
QRemy, line 374:
```suggestion
}

```

gammapy/maps/geom.py
```diff
@@ -631,29 +631,79 @@ def energy_mask(self, energy_min=None, energy_max=None, round_to_edge=False):
...
+        """
+        from . import Map
+
+        axes_names = self.axes_names
+
+        try:
+            axis = self.axes[axis_name]
+        except KeyError:
+            raise ValueError(
+                f"Axis with name `{axis_name}` not in list of axis name: {axes_names}."
```
adonath, line 677: Please move the error message to here: https://github.com/gammapy/gammapy/blob/main/gammapy/maps/axes.py#L1821 (ideally in an independent PR)

This will solve the problem once and for all future cases.

gammapy/maps/geom.py
```diff
@@ -631,29 +631,79 @@ def energy_mask(self, energy_min=None, energy_max=None, round_to_edge=False):
...
+            )
+        shape = tuple([-1] + [1 for n in axes_names if n != axis_name])
```
adonath, line 679: Having this as a little helper function on the `MapAxes` object would be great. Alternatively you can just use `axes.get_coord(mode="edges", axis_name=axis_name)[axis_name]`.

gammapy/maps/geom.py
```diff
@@ -631,29 +631,79 @@ def energy_mask(self, energy_min=None, energy_max=None, round_to_edge=False):
...
+        axis_edges = axis.edges.reshape(shape)
+        edge_min = edge_min if edge_min is not None else axis_edges[0]
+        edge_max = edge_max if edge_max is not None else axis_edges[-1]
+
+        if axis.unit != "":
```
adonath, line 684: I think this check is not really need. It will fail later in the comparison operator with a similar message:
```python

edges = [1, 2, 3, 4] * u.TeV
edges > 1 * u.m
```
```python
UnitConversionError: Can only apply 'greater' function to quantities with compatible dimensions
```
If you feel strongly about the improved error message, maybe introduce a little helper function in `gammapy.utils` like `check_unit`.

gammapy/maps/geom.py
```diff
@@ -631,29 +631,79 @@ def energy_mask(self, energy_min=None, energy_max=None, round_to_edge=False):
...
+
+        if round_to_edge:
+            edge_min, edge_max = axis.round([edge_min, edge_max])
 
-        mask = (energy_edges[:-1] >= energy_min) & (energy_edges[1:] <= energy_max)
+        axes_names.reverse()
+        mask = (axis_edges[:-1] >= edge_min) & (axis_edges[1:] <= edge_max)
+        mask = np.reshape(
+            mask,
+            tuple(len(axis_edges[1:]) if n == axis_name else 1 for n in axes_names),
```
adonath, line 706: Again a little helper function in `MapAxes` would be very useful here. Or just use `np.expand_dims()` which should be simpler.

adonath (commented):
Thanks @MRegeard, I have left some inline comments to address!

gammapy/maps/geom.py
MRegeard: Actually, I just realised that we don't even need this step as we can create the mask over a 1D array. 
I didn't realise that I could use `MapAxes.get_coord`. I don't really think a method is need then as the API is pretty clear and simple.

gammapy/maps/geom.py
MRegeard: I don't think a method on `MapAxes` would help here as if you are using e.g.`WcsGeom`, `.axes` only gather the non spatial axes . But here we need every dimension to broadcast correctly the mask. 
I now use `expand_dims` however.

codemeta.json
```diff
@@ -359,7 +359,7 @@
         "active"
     ],
     "email": "GAMMAPY-COORDINATION-L@IN2P3.FR",
-    "dateModified": "2024-11-26",
+    "dateModified": "2024-12-20",
```
registerrier, line 362: this should be reverted.

gammapy/maps/geom.py
```diff
@@ -631,29 +631,58 @@ def energy_mask(self, energy_min=None, energy_max=None, round_to_edge=False):
         ----------
         energy_min, energy_max : `~astropy.units.Quantity`
             Energy range.
+        round_to_edge: bool, optional
+            Wether to round `energy_min` and `energy_max` to the closest axis bin value.
```
registerrier, line 635:
```suggestion
            Whether to round `energy_min` and `energy_max` to the closest axis bin value.
```

gammapy/maps/geom.py
```diff
@@ -631,29 +631,58 @@ def energy_mask(self, energy_min=None, energy_max=None, round_to_edge=False):
...
+        """Create a mask over a given axis.
+
+        Bins must be fully contained to be included in the mask.
+
+        Parameters
+        ----------
+        edge_min, edge_max : `~astropy.units.Quantity` or float
+            Bounds of the mask. If the axis used for the mask has a unit, passing `~astropy.units.Quantity` is mandatory.
+        round_to_edge: bool, optional
+            Wether to round `energy_min` and `energy_max` to the closest axis bin value.
```
registerrier, line 661:
```suggestion
            Whether to round `energy_min` and `energy_max` to the closest axis bin value.
```

gammapy/maps/geom.py
```diff
@@ -631,29 +631,58 @@ def energy_mask(self, energy_min=None, energy_max=None, round_to_edge=False):
...
+
+    def create_mask(self, axis_name, edge_min=None, edge_max=None, round_to_edge=False):
+        """Create a mask over a given axis.
+
+        Bins must be fully contained to be included in the mask.
+
+        Parameters
+        ----------
+        edge_min, edge_max : `~astropy.units.Quantity` or float
+            Bounds of the mask. If the axis used for the mask has a unit, passing `~astropy.units.Quantity` is mandatory.
```
registerrier, line 659: I would simply write Quantity which unit is consistent with axis.

gammapy/maps/geom.py
```diff
@@ -631,29 +631,58 @@ def energy_mask(self, energy_min=None, energy_max=None, round_to_edge=False):
...
-        # get energy axes and values
-        energy_axis = self.axes["energy"]
+        axes_names = self.axes_names
 
-        if round_to_edge:
-            energy_min, energy_max = energy_axis.round([energy_min, energy_max])
+        axis = self.axes[axis_name]
 
-        # TODO: make this more general
-        shape = (-1, 1) if self.is_hpx else (-1, 1, 1)
```
registerrier, line 650: Can you completely ignore the special case for hpx here?

gammapy/maps/tests/test_core.py
```diff
@@ -990,3 +990,31 @@ def test_data_shape_broadcast():
...
     with pytest.raises(ValueError):
         map2.data = np.ones((1, 1, 2))
+
+
+def test_make_mask_geom():
+    energy = MapAxis.from_energy_bounds("1 TeV", "10 TeV", nbin=5)
+    phase = MapAxis.from_bounds(0, 1, nbin=6, name="phase")
+    freq = MapAxis.from_bounds(1, 10, nbin=4, unit=u.Hz, name="freq")
+
+    geom = WcsGeom.create(10, axes=[energy, phase, freq])
```
registerrier, line 1000: I think you should test other geoms. A minima HpxGeom as it has a different shape that WCSgeom and RegionGeom

registerrier (commented):
Thanks @MRegeard . I have left some comments inline. It seems CI is failing is there an issue with the changes you propose?

gammapy/maps/geom.py
MRegeard: Yes I think we can.

gammapy/maps/tests/test_core.py
MRegeard: Done

MRegeard:
@registerrier Yes it seems that safe mask test are failing due to broadcasting issue. I have to check it!

MRegeard:
@registerrier tests for the `safe_mask_maker.make_mask_energy_bkg_peak` are failling because the `energy_min` pass to the `geom.energy_mask` is an `array_like`, e.g. `[[0.1]] TeV`. 
Should I change the `create_mask` code to handle this case (which would make no sense) or should I modify the `safe_mask_maker` code.

registerrier (approved):
Thanks @MRegeard . This looks good to me now.

registerrier:

PR #5637, opened 2024-12-09
Add Datasets.to_asimov_datasets method

QRemy:
This could replace #5361


gammapy/datasets/core.py
```diff
@@ -331,6 +331,12 @@ def to_spectrum_datasets(self, region):
 
         return datasets
 
+    def to_asimov_datasets(self):
+        """Create Asimov datasets from the current models."""
+        return Datasets(
+            [d._to_asimov_dataset(name=d.name, models=self.models) for d in self]
+        )
+
```
StFroese, line 339: Usually you want to create the asimov dataset for a specific alternative hypothesis. 
I think there are two approaches:
1. Set the desired hypo first for the model attached to the og datasets, then create the asimov dataset and restore the parameters on the original dataset (requires using context manager for user to avoid refitting)
2. Make it possible to give this method a parameters set to create the dataset from

I like this second one better, because you directly get the hypo asimov dataset without touching the original dataset

What do you think?

adonath:
Thanks @QRemy, I agree this is a clear approach and clean API. My only comment would comprise to include more detailed documentation and context on the use of this method with examples and references, once we agreed it is the preferable approach.

gammapy/datasets/map.py
```diff
@@ -1473,6 +1473,41 @@ def stat_sum(self):
         else:
             return cash_sum_cython(counts.ravel(), npred.ravel()) + prior_stat_sum
 
+    def _to_asimov_dataset(self, name=None, models=None):
+        """Create Asimov dataset"""
+        npred = self.npred()
+        data = np.nan_to_num(npred.data, copy=True, nan=0.0, posinf=0.0, neginf=0.0)
```
AtreyeeS, line 1479: I wonder if `MapDataset.npred()` should directly take some kwargs and perform a `nan_to_num` (defaults with Nones).

gammapy/datasets/core.py
AtreyeeS: One can always do
```
dataset.models = models1
dataset_asimov = dataset.to_asimov_dataset()
ts1 = dataset_asimov.stat_sum()

dataset.models = models2
...
```

but maybe 
```
dataset_asimov = dataset.to_asimov_dataset(models)
```
can be simpler...

gammapy/datasets/map.py
```diff
@@ -2694,6 +2729,36 @@ def to_map_dataset(self, name=None):
             meta_table=self.meta_table,
         )
 
+    def _to_asimov_dataset(self, name=None, models=None):
+        """Create Asimov dataset"""
+        npred = self.npred()
+        data = np.nan_to_num(npred.data, copy=True, nan=0.0, posinf=0.0, neginf=0.0)
+        npred.data = data.astype("float")
```
AtreyeeS, line 2736: you can use `super` to avoid doing this twice?

gammapy/datasets/map.py
```diff
@@ -2694,6 +2729,36 @@ def to_map_dataset(self, name=None):
...
+
+        npred_off = get_wstat_mu_bkg(
```
AtreyeeS, line 2738: I am a bit confused. The `npred_off` gives the predicted counts in the off region, whereas `get_wstat_mu_bkg` adjusts this by the acceptance to return the background counts in the on region, no?

gammapy/datasets/map.py
```diff
@@ -2694,6 +2729,36 @@ def to_map_dataset(self, name=None):
...
+            exposure=self.exposure,
+            acceptance=self.acceptance,
+            acceptance_off=self.acceptance_off,
+            psf=self.psf,
+            edisp=self.edisp,
+            mask_safe=self.mask_safe,
+            mask_fit=self.mask_fit,
+            gti=self.gti,
+            name=name,
+            meta=self.meta,
```
AtreyeeS, line 2759: Maybe `MapDataset.copy()` can be adjusted like `Observation.copy` with an `in_memory` option. In a separate PR

gammapy/datasets/tests/test_map.py
```diff
@@ -262,6 +262,20 @@ def test_map_dataset_str(sky_model, geom, geom_etrue):
...
+def test_map_dataset_to_asimov(sky_model, geom, geom_etrue):
+    dataset = get_map_dataset(geom, geom_etrue)
+
+    bkg_model = FoVBackgroundModel(dataset_name=dataset.name)
+    dataset.models = [sky_model, bkg_model]
+
+    npred_sum = dataset.npred().data.sum()
+
+    asimov_dataset = dataset.to_asimov_dataset()
+    counts_asimov = asimov_dataset.counts.data.sum()
```
AtreyeeS, line 274: maybe also assert that the model is correctly reassigned

gammapy/datasets/core.py
```diff
@@ -331,6 +331,12 @@ def to_spectrum_datasets(self, region):
 
         return datasets
 
+    def to_asimov_datasets(self):
```
AtreyeeS, line 334: this needs a test

gammapy/datasets/core.py
```diff
@@ -331,6 +331,12 @@ def to_spectrum_datasets(self, region):
...
+        """Create Asimov datasets from the current models."""
+        return Datasets(
+            [d._to_asimov_dataset(name=d.name, models=self.models) for d in self]
```
AtreyeeS, line 337: why are you calling the hidden function here?

AtreyeeS (commented):
Thanks @QRemy ! Much cleaner implementation indeed!

gammapy/datasets/map.py
QRemy: I think the `in_memory` option of `Observation.copy` has a different purpose (overwrite the lazy fits data loading)

gammapy/datasets/map.py
QRemy: here I just copied what is done for .fake

gammapy/datasets/map.py
QRemy: looking at the property definitions `npred_background=self.alpha.data * get_wstat_mu_bkg` and `npred_off = npred_background/alpha` so `npred_off = get_wstat_mu_bkg`

gammapy/datasets/core.py
QRemy: The hypothesis testing can be defined in dependently of the datasets:
`test = TestStatisticNested([parameter], [null_value])`
then one can get :
`ts_asimov = test.ts_asimov(datasets)`

gammapy/datasets/core.py
QRemy: removed

gammapy/datasets/map.py
QRemy: I don't think it is applicable here.

gammapy/datasets/core.py
QRemy: added

gammapy/datasets/tests/test_map.py
QRemy: done

gammapy/datasets/map.py
QRemy: could add a `_init_copy` method in separate PR.

gammapy/datasets/map.py
```diff
@@ -2694,6 +2723,46 @@ def to_map_dataset(self, name=None):
...
+    def _to_asimov_dataset(self):
+        """Create Asimov dataset from the current models.
+
+        Parameters
+        ----------
+        name : str, optional
+            Name of the new dataset. Default is None.
+
+        """
+        npred = self.npred()
```
AtreyeeS, line 2735: this will not work if self.counts is None

AtreyeeS, line 2735: Is `self.counts` is None, this will not work 
```
1 dataset_on_off.counts = None
----> 2 dataset_on_off.npred()

File ~/anaconda3/envs/gammapy-1.2/lib/python3.10/site-packages/gammapy/datasets/map.py:639, in MapDataset.npred(self)
    636 npred_total = self.npred_signal()
    638 if self.background:
--> 639     npred_total += self.npred_background()
    640 npred_total.data[npred_total.data < 0.0] = 0
    641 return npred_total
```

gammapy/datasets/map.py
```diff
@@ -1473,6 +1473,35 @@ def stat_sum(self):
...
+        data = np.nan_to_num(npred.data, copy=True, nan=0.0, posinf=0.0, neginf=0.0)
+        npred.data = data.astype("float")
+
+        asimov_dataset = self.__class__(
+            models=self.models,
+            counts=npred,
```
AtreyeeS, line 1491: is it okay to have non integral counts?

AtreyeeS: I mean, it is of course valid for Asimov datasets, but can we just use the same in our normal datasets?

gammapy/datasets/map.py
QRemy: I think empty counts should be given has an empty map so the geom is taken properly

gammapy/datasets/map.py
registerrier: But is it really what we should be doing?

If you want to compute the sensitivity of a given observation with observed `counts_off` I think you have to keep the specific realization.

If it is a priori, you probably need to do it from a background model, like fake, no?

gammapy/datasets/map.py
```diff
@@ -2694,6 +2723,46 @@ def to_map_dataset(self, name=None):
...
+            n_off=self.counts_off.data,
+            alpha=self.alpha.data,
+            mu_sig=npred.data,
+        )
+        npred_off = Map.from_geom(geom=self._geom, data=npred_off)
+
+        asimov_dataset = self.__class__(
+            models=self.models,
+            counts=npred,
+            counts_off=npred_off,
```
registerrier, line 2750: I am not sure you should change `counts_off`

registerrier (commented):
Thanks @QRemy . This is nice.

I have one issue: should you really change the `counts_off`? I don't see a reason for that.

Technically you are computing the Asymov dataset for the specific OFF realization at hand. You cannot change it based on the new `counts`. 
If you have no OFF, you would need a background model, no?

StFroese:
> Thanks @QRemy . This is nice.
> 
> 
> 
> I have one issue: should you really change the `counts_off`? I don't see a reason for that.
> 
> 
> 
> Technically you are computing the Asymov dataset for the specific OFF realization at hand. You cannot change it based on the new `counts`. 
> 
> If you have no OFF, you would need a background model, no? 

You also always have to create the Asimov dataset with the nuisance parameters (bkg norm for example) that are fitted on the measurement dataset with the proposed strength parameter set

registerrier (approved):

Could it be an issue that the asymov dataset has the same name as the parent one?

QRemy:
> Could it be an issue that the asymov dataset has the same name as the parent one?
I think it simpler to keep the same names to avoid reassigning the models.
and in principle we will never use them together with so parent ones so they can have the same name.

PR #5638, opened 2024-12-10
Add a new method to compute ts and sigma in gammapy.stats.utils

QRemy:
Add a new method "cowan" in `ts_to_sigma` and `sigma_to_ts` considering
that the  ts  distribution follows a noncentral chi-square distribution for r-degrees of freedom with noncentrality parameter equal to ts (applicable for the asimov dataset).

I'm not sure if this is valid when df!=1.

### DIFF
--- a/gammapy/stats/utils.py
+++ b/gammapy/stats/utils.py
@@ -1,11 +1,11 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 
 import numpy as np
-from scipy.stats import chi2
+from scipy.stats import chi2, ncx2
 
 
-def sigma_to_ts(n_sigma, df=1):
-    """Convert number of sigma to delta ts according to the Wilks' theorem.
+def sigma_to_ts(n_sigma, df=1, method="wilk"):
+    """Convert number of sigma to delta ts.
 
     The theorem is valid only if:
     - the two hypotheses tested can be defined in the same parameters space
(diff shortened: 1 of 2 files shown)


StFroese:
> I'm not sure if this is valid when df!=1.

I thinks so, yes, the wald limits are actually calculated in r-dimensional d.o.f space.
It's stated here on the last two pages of the original publication, where this all comes from

https://www.pp.rhul.ac.uk/~cowan/stat/wald1943.pdf

The assumption about the TS is in general not a one but multivariate normal distribution. 
and $\lamdba$ is denoted for parameter point $\theta_n$, so $\lambda_n$.
This is also given in the cowen paper on p.10, directly before 3.2

StFroese:
$\int_{N_\sigma}^\infty \mathcal(N(x\vert 0,1))dx \overset{!}{=} \int_{q_{0,\text{obs}}}^\infty f(q_0\vert \mu, \theta)dq_0$ and $f(q_0\vert \mu, \theta)$ can be a non-central chi2 with df!=0

Astro-Kirsty:
> The point that bother me is that for df=1 using the non-central chi2 distribution or the chi2 distribution give the same significance, but not for df>1 so in that case how to choose a priori the distribution ?

The non-central chi2 should only become 'central' when both df=1 and the non-centrality factor is 0. 
At least this is the way I understand it!


Also just a note, it should be 'Wilks' everywhere, the 's' is actually in the name :)

StFroese:
> The non-central chi2 should only become 'central' when both df=1 and the non-centrality factor is 0. At least this is the way I understand it!

Isn't the non-central chi2 always central when the non-centrality parameter is 0. It is non-central relative to the chi2 with ndof

QRemy:
> Thanks @QRemy ! Should the user be allowed to choose the non centrality factor? Is there any other use case? I dont understand your concerns for df>1...

what is implemented here is basically :

```
p=ncx2.sf(ts, df=df, nc=ts)
sigma = np.sqrt(ncx2.isf(p, df=1, nc=ts))
```

the non centrality factor is set to `nc=ts` following Eq. 29 of Cowan et al. 
Screenshot 2024-12-11 at 21 06 27

which is equivalent to 

```
p=chi2.sf(ts, df=df)
sigma = np.sqrt(chi2.isf(p, df=1))
```

only if df=1

If instead we always consider nc=0 we don't need this PR at all.

Astro-Kirsty:
> If instead we always consider nc=0 we don't need this PR at all.

Yes of course. My point was addressing the concern about sigma being equal for both when df=1
Just not convinced that should have been the case

StFroese:
> > Thanks @QRemy ! Should the user be allowed to choose the non centrality factor? Is there any other use case? I dont understand your concerns for df>1...
> 
> what is implemented here is basically :
> 
> ```
> p=ncx2.sf(ts, df=df, nc=ts)
> sigma = np.sqrt(ncx2.isf(p, df=1, nc=ts))
> ```
> 
> the non centrality factor is set to `nc=ts` following Eq. 29 of Cowan et al.
> 
> Screenshot 2024-12-11 at 21 06 27
> which is equivalent to
> 
> ```
> p=chi2.sf(ts, df=df)
> sigma = np.sqrt(chi2.isf(p, df=1))
> ```
> 
> only if df=1
> 
> If instead we always consider nc=0 we don't need this PR at all.

But I think you can calculate the sigma with the first equation in the cowen paper, which is just based on a given p value for any chi2 or ncx2 with some df
Screenshot 2024-12-12 at 11 49 29

StFroese:
I you look here, Cowen just plugs the p value for the q_0 case into the above equation

Screenshot 2024-12-12 at 11 51 33

registerrier (changes requested):
Thanks @QRemy.  I think there is a problem on the statistical basis and the usage could be clarified.

The fact that you use `ts` both for the non-central parameter and for the ts value will more or less always yield a p_value of 0.5. This is the approach used in sensitivity calculation as it gives 50% chance of a positive outcome.  

In general, `nc` would be given by the `Asymov`  dataset TS while the value is the actual TS. So instead of passing a method keyword you could pass a `ts_asymov` instead.

This needs more documentation as in 99% of the cases the user shouldn't use Wald.

gammapy/stats/utils.py
```diff
@@ -17,39 +17,64 @@ def sigma_to_ts(n_sigma, df=1):
         Significance in number of sigma.
     df : int
         Number of degree of freedom.
+    n_sigma_asymov : float
```
adonath, line 20: Adding that it uses a non-central chi2 in this case would be good.

registerrier (approved):

Maybe it would be worth clarifying what the function `sigma_to_ts` produces when `n_sigma_asymov` is non zero.

Issue #5251, opened 2024-05-07
Consistency with 'references' in docstrings


Astro-Kirsty:
Currently we have a number of different ways to implement a reference. We should make sure we utilise the same method each time.

**Example that links to the actual references page – personally I like this option**
https://docs.gammapy.org/1.2/api/gammapy.estimators.TSMapEstimator.html

Full reference in docstring
https://docs.gammapy.org/1.2/api/gammapy.catalog.SourceCatalog1LHAASO.html
https://docs.gammapy.org/1.2/api/gammapy.catalog.SourceCatalog2HWC.html

List of multiple references
https://docs.gammapy.org/1.2/api/gammapy.stats.cash.html
https://docs.gammapy.org/1.2/api/gammapy.astro.darkmatter.ZhaoProfile.html

I suggest we decide on a specific format and then add that to the documentation guide.
https://docs.gammapy.org/1.2/development/doc_howto.html

There are also a number of things that we could add to the documentation guide to ensure we are producing the correct docstrings and examples. Also as a note to everyone, if we are creating new functions or example codes we should make sure to build the docs and the run the test for the docstring examples i.e.
`pytest --doctest-modules gammapy/estimators/parameter.py`

PR #5639, opened 2024-12-10
Add ParameterSensitivityEstimator and support sensitivity option on FluxpointsEstimator

QRemy:
- Add ParameterSensitivityEstimator a helper class used by the ParameterEstimator to compute sensitivity
 (and expose both classes)
- Add sensitivity option on FluxPointsEstimator

Requires #5638 to use the ts defined in Cowan et al.
Requires #5637 for the new implementation of ts_asimov that should fix an issue with ray.


gammapy/estimators/parameter.py
```diff
@@ -345,5 +377,104 @@ def run(self, datasets, parameter):
...
+        self.parameter = parameter
+        self.n_sigma = n_sigma
+        self.rtol = rtol
+        self.max_niter = max_niter
+
+    def _fcn(self, value, datasets):
+        """Call the Test Statistics function."""
+        self.parameter.value = value
+        ts_asimov = self.test.ts_asimov(datasets)
+        return ts_to_sigma(ts_asimov, method="wald") - self.n_sigma
```
QRemy, line 443:
```suggestion
        return ts_to_sigma(ts_asimov, ts_asimov=ts_asimov) - self.n_sigma
```

gammapy/estimators/parameter.py
```diff
@@ -60,19 +70,21 @@ class ParameterEstimator(Estimator):
     """
 
     tag = "ParameterEstimator"
-    _available_selection_optional = ["errn-errp", "ul", "scan"]
+    _available_selection_optional = ["errn-errp", "ul", "scan", "sensitivity"]
```
bkhelifi, line 73: the docstring should be also updated

gammapy/estimators/parameter.py
```diff
@@ -262,6 +274,27 @@ def estimate_ul(self, datasets, parameter):
...
+        Parameters
+        ----------
+        datasets : `~gammapy.datasets.Datasets`
+            Datasets.
+
+        Returns
+        -------
+        result : dict
+            Dictionary with an array with one entry per dataset with the sum of the
+            masked npred.
```
bkhelifi, line 289: can you describe the content of the dictionary? One needs to be now more explicit and clear for the dict returned values, following users feedbacks

gammapy/estimators/parameter.py
```diff
@@ -345,5 +377,104 @@ def run(self, datasets, parameter):
...
+    to have 50% of measurements above a given significance threshold.
+
+
+    Parameters
+    ----------
+    parameter : `~gammapy.modeling.Parameter`
+       Parameter to test
+    null_value : float or `~gammapy.modeling.Parameter`
+        Value of the parameter for the null hypothesis.
+    n_sigma : int, default=5
```
bkhelifi, line 404: Having a float would be more generic...

gammapy/estimators/parameter.py
```diff
@@ -345,5 +377,104 @@ def run(self, datasets, parameter):
...
+        Relative precision of the estimate. Used as a stopping criterion.
+        Default is 0.01.
+    max_niter : int
+        Maximal number of iterations used by the root finding algorithm.
+        Default is 100.
+
+    References
+    ----------
+        Cowan et al. (2011), European Physical Journal C, 71, 1554.
+        doi:10.1140/epjc/s10052-011-1554-0.
```
bkhelifi, line 416: Does Sphinx generate a clickable link with this syntax?

gammapy/estimators/parameter.py
```diff
@@ -345,5 +377,104 @@ def run(self, datasets, parameter):
...
+    """
+
+    tag = "ParameterSensitivityEstimator"
+
+    def __init__(
+        self,
+        parameter,
+        null_value,
+        n_sigma=5,
+        n_free_parameters=None,
```
bkhelifi, line 427: This parameter is not described in the docstring... Why?

gammapy/estimators/parameter.py
```diff
@@ -345,5 +377,104 @@ def run(self, datasets, parameter):
...
+        self.parameter = parameter
+        self.n_sigma = n_sigma
+        self.rtol = rtol
+        self.max_niter = max_niter
+
+    def _fcn(self, value, datasets):
+        """Call the Test Statistics function."""
+        self.parameter.value = value
+        ts_asimov = self.test.ts_asimov(datasets)
+        return ts_to_sigma(ts_asimov, ts_asimov=ts_asimov) - self.n_sigma
```
bkhelifi, line 443: I have a small doubt... Looking at the dev code, the function gammapy.stats.utils.ts_to_sigma  does not have `ts_asimov` as argument: `def ts_to_sigma(ts, df=1):`. Can you explain me?

PS: the function `ts_to_sigma` is not exposed into the documentation (https://docs.gammapy.org/dev/api-reference/stats.html) ... surprising!

gammapy/estimators/parameter.py
```diff
@@ -345,5 +377,104 @@ def run(self, datasets, parameter):
...
+
+    def parameter_matching_significance(self, datasets):
+        """Parameter value  matching the target significance"""
```
bkhelifi, line 446: Can you precise in the docstring what is returned in case of failure?

gammapy/estimators/parameter.py
QRemy: I would propose to change that consistently for all estimators in a separate PR.

gammapy/estimators/parameter.py
QRemy: maybe @Astro-Kirsty or @adonath know

gammapy/estimators/parameter.py
QRemy: requires #5638

gammapy/estimators/parameter.py
Astro-Kirsty: I can look into it, but overall we actually need to unify our implementation of the `references` https://github.com/gammapy/gammapy/issues/5251

QRemy:
This is ready now.
However I noticed that `_to_asimov_datasets` is quite inefficient when using ray via the `DatasetsActor` because it recreate a datasets in the main process  at each iteration of the brentq method. I'll try into to optimize this that later.

gammapy/estimators/parameter.py
```diff
@@ -345,5 +377,107 @@ def run(self, datasets, parameter):
...
+
+class ParameterSensitivityEstimator:
+    """Estimate the sensitivity to a given parameter
+
+    Computes the TS distribution in the non-null hypothesis using the
+    log likelihood of the Asimov dataset (i.e. a dataset with counts = npred)
+    and the non central chi2 distribution.
+    Once the TS distribution under the testing hypothesis is known,
+    one can compute the required parameter value
+    to have 50% of measurements above a given significance threshold.
```
registerrier, line 395: Shouldn't this be a parameter?
For instance the "brazilian flag" plots show the expectation (the 50% chance probability) along the 1 and 2 sigma ranges hence the 2.5%, 16%, 84%, and 97.5% chance proabilities

gammapy/estimators/parameter.py
```diff
@@ -345,5 +377,107 @@ def run(self, datasets, parameter):
...
+        Relative precision of the estimate. Used as a stopping criterion.
+        Default is 0.01.
+    max_niter : int
+        Maximal number of iterations used by the root finding algorithm.
+        Default is 100.
+
+    References
+    ----------
+        Cowan et al. (2011), European Physical Journal C, 71, 1554.
+        doi:10.1140/epjc/s10052-011-1554-0.
```
registerrier, line 416: Maybe try to follow the reference scheme introduced in #5683

gammapy/estimators/parameter.py
```diff
@@ -345,5 +377,107 @@ def run(self, datasets, parameter):
...
+        # Where the root finding fails NaN is set as norm
+        roots = roots[roots > 0]
+        if roots.size > 0:
+            return roots[0]
+        else:
+            return np.nan
+
+    def run(self, datasets):
+        """Parameter sensitivity
+        given as the diffrence between vaue matching the target significance and the null value.
```
registerrier, line 478:
```suggestion
        given as the difference between value matching the target significance and the null value.
```

gammapy/estimators/parameter.py
registerrier: I must say that I don't see where in the code this should appear. But I think at some point it should.

gammapy/estimators/parameter.py
```diff
@@ -345,5 +377,107 @@ def run(self, datasets, parameter):
...
+        parameter,
+        null_value,
+        n_sigma=5,
+        n_free_parameters=None,
+        rtol=0.01,
+        max_niter=100,
+    ):
+        self.test = TestStatisticNested(
+            [parameter], [null_value], n_free_parameters=n_free_parameters
+        )
```
registerrier, line 433: A general design question here. You might have moved most the content of the ParameterSensitivityEstimator into the `TestStatisticNested`. Could it make sense?

gammapy/estimators/points/tests/test_sed.py
```diff
@@ -211,7 +211,8 @@ def fpe_map_pwl_reoptimize():
 
 @pytest.fixture(scope="session")
 def fpe_pwl():
-    return create_fpe(PowerLawSpectralModel())
+    fpe_pwl = create_fpe(PowerLawSpectralModel())
+    return fpe_pwl
```
registerrier, line 215: Why the change here?

registerrier (commented):
Thanks @QRemy. I have left some comments inline.

One general question is whether the 50% chance detection as definition of sensitivity can be a parameter.

gammapy/estimators/parameter.py
```diff
@@ -22,6 +30,8 @@ class ParameterEstimator(Estimator):
         Sigma to use for asymmetric error computation. Default is 1.
     n_sigma_ul : int
         Sigma to use for upper limit computation. Default is 2.
+    n_sigma_sensitivity : int
+        Sigma to use for sensitivity computation. Default is 5.
```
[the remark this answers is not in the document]
QRemy: I just used the same than the `SensitivityEstimator` (and same than Excess/TS estimators).
Could be changed consistently in a separate PR.

gammapy/estimators/points/tests/test_sed.py
QRemy, line 215:
```suggestion
    return create_fpe(PowerLawSpectralModel())
```

gammapy/estimators/parameter.py
QRemy, line 416:
```suggestion
        * `Cowan et al. (2011), "Asymptotic formulae for likelihood-based tests of new physics"
        <https://arxiv.org/abs/1007.1727>`_
```

gammapy/estimators/parameter.py
QRemy: I think that the ParameterSensitivityEstimator is higher level than the TestStatisticNested, and how rtol and max_niter option should be passed if moved on TestStatisticNested ? as a kwarg_sensitivity (seems more complicated to me)

gammapy/estimators/parameter.py
QRemy: The p-value is calculated in `ts_to_sigma` and setting ts=nc=ts_asimov with 1df gives 50%.
If we want the  (2.5%, 16%, 84%, and 97.5%) chance proabilities it should be equivalent run the evaluator with n_sigma - (-2,-1,+1+2)

registerrier: OK. 
I think it is fine for most of the n_sigma range but maybe not for all. Maybe we will have to check that.

registerrier (approved):

PR #5640, opened 2024-12-11
Helper function for apply_threshold_sensitivity

QRemy:
Small cleanup adding helper function for apply_threshold_sensitivity.


QRemy:
This could be potentially used in the method .to_asimov_dataset from #5637 in order to modify the excess_counts and to reproduce the behaviour of the current SensitivityEstimator.
However I'm not sure if these conditions are applicable for datasets. Should it be applied to each dataset or to the total excess,  in the latter case how to do so.

QRemy:
Merging directly as it's only cleanup.

PR #5641, opened 2024-12-12
Fix nan output in ts_to_sigma and sigma_to_ts

QRemy:
Fix `nan` output in `ts_to_sigma` and `sigma_to_ts` caused by limited precision in `scipy.stats.chi2` functions. The solution proposed here rely on `mpmath` package to strore values with a large number of digits when `scipy` fails.

### DIFF
--- a/gammapy/stats/tests/test_utils.py
+++ b/gammapy/stats/tests/test_utils.py
@@ -4,7 +4,6 @@ from gammapy.stats.utils import sigma_to_ts, ts_to_sigma
 
 
 def test_sigma_ts_conversion():
-
     sigma_ref = 3
     ts_ref = 9
     df = 1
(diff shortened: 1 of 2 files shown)


AtreyeeS:
For very high TS (which leads to nan), sigma = sqrt(TS) usually suffices, no?

QRemy:
> For very high TS (which leads to nan), sigma = sqrt(TS) usually suffices, no?

not if df >1

QRemy:
Closing now, this modified functions will be added in a recipe instead.
The only function where infinite significance can be an issue is `get_combined_significance_maps`.
#5644 add the ts to the output of this function so the significance can always be recomputed a posteriori with more precise method as this one.

PR #5644, opened 2024-12-14
Add ts in combine_significance_maps output

QRemy:
As discussed in dev call add ts in combine_significance_maps output
so if the significance is inf due to precision error it can be recomputed a posteriori (for example with the method of #5641)


QRemy:
Merging directly as it is a minor change discussed in the dev call.

PR #5645, opened 2024-12-17
Fix `expfactor` unit to `MeV**-index_2`

### LINKED ISSUE #5628: `SuperExpCutoffPowerLaw4FGLSpectralModel` provides wrong results if the energy units are different from MeV
**Gammapy version**
gammapy dev version.

**Bug description**
The class `SuperExpCutoffPowerLaw4FGLSpectralModel` provides wrong results when `__call__()` is used if the units of `energy` are different from the units of the reference energy. This affects methods like `energy_flux`, `plot`, etc.

When using `_convert_evaluate_unit()` inside `__call__()`, the units of `reference` are changed to the `energy` units. Since `expfactor` is defined in units of `reference.unit**index_2`, the units are changed, but the value of `expfactor` is not modified accordingly.  

**Expected behavior**
The value of `expfactor` should change accordingly when the units of `reference` are changed. Another option is to force it to a fixed energy unit, e.g., MeV**`index_2`.

**Possible solution**
Since accounting for varying units in `reference` would affect `_convert_evaluate_unit()` by including ad hoc code for this function, I propose to fix the units of `expfactor` to MeV**`index_2`. This approach should be the easiest solution and in line with the units provided in 4FGL DR1 and DR2.

I can take care of the PR.

**To Reproduce**
```
from gammapy.modeling.models import SuperExpCutoffPowerLaw4FGLSpectralModel
import astropy.units as u

secpl=SuperExpCutoffPowerLaw4FGLSpectralModel(
    amplitude=2e-12*u.Unit("MeV-1 s-1 cm-2"),
    reference=1e+03*u.MeV,
    expfactor=5e-03,
    index_1=2e+00,
    index_2=6.6667e-01,
)

print(secpl.energy_flux(100*u.MeV,1000*u.MeV).to_value("MeV cm-2 s-1"),secpl.energy_flux(0.1*u.GeV,1*u.GeV).to_value("MeV cm-2 s-1"))

secpl.plot(sed_type="dnde",energy_bounds=[100*u.MeV,100000*u.MeV],label="OK")
secpl.plot(sed_type="dnde",energy_bounds=[100*u.MeV,100*u.GeV],label="wrong")
plt.legend()
```

aaguasca:
**Description**

This PR is to resolve #5628. 

**Dear reviewer**

This PR is ready for review.

### DIFF
--- a/examples/models/spectral/plot_super_exp_cutoff_powerlaw_4fgl_dr1.py
+++ b/examples/models/spectral/plot_super_exp_cutoff_powerlaw_4fgl_dr1.py
@@ -36,7 +36,7 @@ model = SuperExpCutoffPowerLaw4FGLSpectralModel(
     index_2=2,
     amplitude="1e-12 TeV-1 cm-2 s-1",
     reference="1 TeV",
-    expfactor=1e-2,
+    expfactor=1e-14,
 )
 model.plot(energy_range)
 plt.grid(which="both")
--- a/gammapy/modeling/models/spectral.py
+++ b/gammapy/modeling/models/spectral.py
@@ -1526,8 +1526,8 @@ class SuperExpCutoffPowerLaw4FGLSpectralModel(SpectralModel):
         :math:`E_0`. Default is 1 TeV.
     expfactor : `~astropy.units.Quantity`
         :math:`a`, given as dimensionless value but
-        internally assumes unit of :math:`E_0^{-\Gamma_2}`.
-        Default is 1e-2.
+        internally assumes unit of :math:`{\rm MeV}^{-\Gamma_2}`.
+        Default is 1e-14.
     """
 
     tag = ["SuperExpCutoffPowerLaw4FGLSpectralModel", "secpl-4fgl"]
@@ -1538,7 +1538,7 @@ class SuperExpCutoffPowerLaw4FGLSpectralModel(SpectralModel):
         interp="log",
     )
     reference = Parameter("reference", "1 TeV", frozen=True)
-    expfactor = Parameter("expfactor", "1e-2")
+    expfactor = Parameter("expfactor", "1e-14")
     index_1 = Parameter("index_1", 1.5)
     index_2 = Parameter("index_2", 2)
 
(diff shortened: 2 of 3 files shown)


QRemy (commented):
Thanks @aaguasca, could you add a regression test that fails for the main and works on this branch. This would ensure that the bug you encountered is fixed and prevent to reintroduce it future.

aaguasca:
Thanks @QRemy, for spotting the missing test. I added a test forcing `SpectralModel.__call__()` to change the unit of `reference` through `SpectralModel._convert_evaluate_energy()`. I think this should be enough. Let me know if this is not the case.

registerrier (commented):
Thanks @aaguasca . Can you also check that the evaluation of models extracted from the 4FGL is unchanged?

aaguasca:
Thanks @registerrier. I added the evaluation of 4FGL models at 10 GeV since the evaluation at the reference energy only checks the normalization value.

registerrier (commented):
Thanks @aaguasca . This is good.
One last question it seems to me that among the 4FGL source one evaluation value has changed with the changes you propose. 0002.8+6217. Do you confirm? Was the previous result incorrect?

aaguasca:
Yes, I confirm evaluation value for 0002.8+6217 has to change. The evaluation results at 10 GeV without considering this PR was not OK for the SuperExpCutoffPowerLaw4FGLSpectralModel. You can see the difference in the dN/dE in the plot between main and this PR. The vertical line is at 10 GeV, where the test evaluates the model.

image
image

code:
```
from gammapy.catalog import SourceCatalog4FGL
import numpy as np
import astropy.units as u

cata=SourceCatalog4FGL("$GAMMAPY_DATA/catalogs/fermi/gll_psc_v20.fit.gz")
model=cata["4FGL J0002.8+6217"].spectral_model()

fig,ax=plt.subplots()
ax.yaxis.units=u.Unit("GeV-1 s-1 cm-2")
model.plot(energy_bounds=[100*u.MeV,1e5*u.MeV],label="energy bounds in MeV")
model.plot(energy_bounds=[0.1*u.GeV,1000*u.GeV],label="energy bounds in GeV")
plt.legend()
plt.vlines(1e4*u.MeV,1e-21*u.Unit("GeV-1 cm-2 s-1"),1e-7*u.Unit("GeV-1 cm-2 s-1"),ls="--",color="gray")
plt.xlim((100*u.MeV,1e6*u.MeV))
plt.ylim((1e-23*u.Unit("GeV-1 cm-2 s-1"),8.e-07*u.Unit("GeV-1 cm-2 s-1")))
```

adonath (approved):
Thanks @aaguasca, looks good to me now!

PR #5646, opened 2024-12-17
Move datasets parser to gammapy.modeling.utils

QRemy:
Move datasets parser to gammapy.modeling.utils


AtreyeeS (approved):
Thanks @QRemy ! Good to reduce code duplication between the Fit and Sampler classes

PR #5647, opened 2024-12-18
Set  apply_irf["edisp"]=False in create_fermi_isotropic_diffuse_model

QRemy:
Set  apply_irf["edisp"]=False in create_fermi_isotropic_diffuse_model

See https://fermi.gsfc.nasa.gov/ssc/data/access/lat/BackgroundModels.html :

Screenshot 2024-12-18 at 14 55 21

### DIFF
--- a/gammapy/modeling/models/cube.py
+++ b/gammapy/modeling/models/cube.py
@@ -1234,5 +1234,5 @@ def create_fermi_isotropic_diffuse_model(filename, **kwargs):
         spatial_model=spatial_model,
         spectral_model=spectral_model,
         name="fermi-diffuse-iso",
-        apply_irf={"psf": False, "exposure": True, "edisp": True},
+        apply_irf={"psf": False, "exposure": True, "edisp": False},
     )


adonath (approved):
Thanks @QRemy, maybe put the link as a comment for reference?

facero:
ah good point. Thanks @QRemy .
Did you notice some changes in the spectral results ?

QRemy:
@adonath the page link is already in the docstring and I don't see a way to point this specific section except with a highlight https://fermi.gsfc.nasa.gov/ssc/data/access/lat/BackgroundModels.html#:~:text=unbinned%20likelihood%20analysis.-,Disabling%20Energy%20Dispersion,-For%20certain%20data which is a bit long (can we use a markdown link in the docstring ?)

MRegeard (approved):
Thanks @QRemy, that's a good catch !

MRegeard:
@facero, I think it shouldn't change anything for Gammapy right now since we don't provide a way to read the energy dispersion of the LAT.

PR #5648, opened 2024-12-18
Adding inverse_cdf to existing priors

facero:
Splitting the sampler PR #5643 into smaller chunks with what concerns the Priors here for simplicity.

**Description**

Samplers usually benefit from LogUniform priors for the normalization for example.
`inverse_cdf` is needed for the Nested Sampling algorithm to perform random uniform sampling from [0, 1].
So all priors (uniform, gaussian, or LogUniform) need a function to transform from min, max to  [0, 1] (i.e. the inverse_cdf).

This pull request moves what existed in #5643 into this PR.
Todo list: 
- Add the inverse_cdf for the Gaussian prior
- Add tests for the inverse_cdf
- Add tests for the LogUnifom
- Update doc concerning priors with these new functions and priors

### DIFF
--- a/gammapy/modeling/models/__init__.py
+++ b/gammapy/modeling/models/__init__.py
@@ -9,7 +9,7 @@ from .cube import (
     TemplateNPredModel,
     create_fermi_isotropic_diffuse_model,
 )
-from .prior import GaussianPrior, Prior, UniformPrior
+from .prior import GaussianPrior, Prior, UniformPrior, LogUniformPrior
 from .spatial import (
     ConstantFluxSpatialModel,
     ConstantSpatialModel,
--- a/gammapy/modeling/models/__init__.py
+++ b/gammapy/modeling/models/__init__.py
@@ -9,7 +9,7 @@ from .cube import (
     TemplateNPredModel,
     create_fermi_isotropic_diffuse_model,
 )
-from .prior import GaussianPrior, Prior, UniformPrior, LogUniformPrior
+from .prior import GaussianPrior, Prior, UniformPrior
 from .spatial import (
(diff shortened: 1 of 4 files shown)


adonath (commented):
Thanks @facero, ideally you could split the PR into adding the `LogUniform` prior and adding the `inverse_cdf` consistently in a follow-up PR.

facero:
Need to add unit test for inverse cdf and then this should be good

facero:
For all the inverse cdf we are using the scipy functions for easier use (as we will do for the LogUniform as well).

facero:
Added the tests 
Ready for merging 
@adonath , @QRemy can you make a quick review ?

adonath (dismissed):
Thanks @facero, looks good to me!

QRemy:
Thanks @facero. fine for me too, merging now.

facero:
> Thanks @facero. fine for me too, merging now.

Ah maybe just one thing

https://github.com/gammapy/gammapy/blob/bd254d267f51c0d4e169786cb1ab51a13cb27259/gammapy/modeling/models/prior.py#L195

Should this be value instead of val to be consistent with other calls ?

QRemy:
> > Thanks @facero. fine for me too, merging now.
> 
> Ah maybe just one thing
> 
> https://github.com/gammapy/gammapy/blob/bd254d267f51c0d4e169786cb1ab51a13cb27259/gammapy/modeling/models/prior.py#L195
> 
> Should this be `value` instead of `val` to be consistent with other calls ?

Yes that would be nicer.

facero:
Done

PR #5649, opened 2024-12-19
Add functions to create MapDataset from Fermi-LAT files

QRemy:
Add functions to create Observation and MapDataset from Fermi-LAT files

- `create_observation_from_fermi_files`
- `_drm_to_edisp` (support for edisp, thanks @MRegeard)
- `create_map_dataset_from_dl4_irfs` (should work for any observation containing DL4 irfs including also HAWC/SWGO @LauraOlivera)
- `add_fermi_isotropic_model` (little helper function to attach the isotropic model to a dataset)
- `create_map_dataset_from_fermi_files` (uses all of the above)

Tests are missing, first I have to check if we have Fermi-LAT data with all the files necessary in gammapy-data.
Requires https://github.com/gammapy/gammapy-data/pull/60

### DIFF
--- a/gammapy/data/__init__.py
+++ b/gammapy/data/__init__.py
@@ -1,6 +1,5 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 """Data and observation handling."""
-
 from gammapy.utils.observers import observatory_locations
 from .data_store import DataStore
 from .event_list import EventList
--- a/gammapy/data/__init__.py
+++ b/gammapy/data/__init__.py
@@ -9,13 +9,12 @@ from .gti import GTI
 from .hdu_index_table import HDUIndexTable
 from .metadata import EventListMetaData, ObservationMetaData
 from .obs_table import ObservationTable
-from .observations import Observation, Observations, create_observation_from_fermi_files
+from .observations import Observation, Observations
 from .pointing import FixedPointingInfo, PointingInfo, PointingMode
 from .simulate import ObservationsEventsSampler
 from .utils import get_irfs_features
 
 __all__ = [
-    "create_observation_from_fermi_files",
     "DataStore",
     "EventList",
     "EventListMetaData",
--- a/gammapy/data/__init__.py
+++ b/gammapy/data/__init__.py
@@ -1,5 +1,6 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 """Data and observation handling."""
+
 from gammapy.utils.observers import observatory_locations
 from .data_store import DataStore
 from .event_list import EventList
@@ -8,12 +9,13 @@ from .gti import GTI
 from .hdu_index_table import HDUIndexTable
 from .metadata import EventListMetaData, ObservationMetaData
 from .obs_table import ObservationTable
-from .observations import Observation, Observations
+from .observations import Observation, Observations, create_observation_from_fermi_files
 from .pointing import FixedPointingInfo, PointingInfo, PointingMode
 from .simulate import ObservationsEventsSampler
 from .utils import get_irfs_features
 
 __all__ = [
+    "create_observation_from_fermi_files",
     "DataStore",
     "EventList",
     "EventListMetaData",
--- a/gammapy/data/observations.py
+++ b/gammapy/data/observations.py
@@ -6,7 +6,6 @@ import inspect
 import itertools
 import logging
 import warnings
-from pathlib import Path
 from itertools import zip_longest
 import numpy as np
 import astropy.units as u
@@ -16,9 +15,6 @@ from astropy.time import Time
 from astropy.units import Quantity
 from astropy.utils import lazyproperty
 import matplotlib.pyplot as plt
-from gammapy.irf.edisp.map import _drm_to_edisp
-from gammapy.irf import PSFMap
-from gammapy.maps import Map
 from gammapy.utils.deprecation import GammapyDeprecationWarning
 from gammapy.utils.fits import LazyFitsData, earth_location_to_dict
 from gammapy.utils.metadata import CreatorMetaData, TargetMetaData, TimeInfoMetaData
@@ -32,7 +28,7 @@ from .metadata import ObservationMetaData
 from .pointing import FixedPointingInfo
 
 
-__all__ = ["Observation", "Observations", "create_observation_from_fermi_files"]
+__all__ = ["Observation", "Observations"]
 
 log = logging.getLogger(__name__)
 
@@ -915,58 +911,3 @@ class ObservationChecker(Checker):
         except Exception:
             yield self._record(level="warning", msg="Loading psf failed")
             return
-
-
-def create_observation_from_fermi_files(
-    path,
-    events_filename="ft1_00.fits",
-    counts_filename="ccube_00.fits",
-    exposure_filename="bexpmap_roi_00.fits",
-    psf_filename="psf_00.fits",
-    drm_filename="drm_00.fits",
-    obs_id=0,
-):
-    """Read Fermi-LAT files and create an Observation.
-
-    Parameters
-    ----------
-    path : str
-        path to files
-    events_filename : str
-        Events filename. Default is ft1_00.fits
-    counts_filename : str, optional
-        Counts filename. Default is ccube_00.fits
-    exposure_filename : str, optional
-        exposure filename. Default is bexpmap_roi_00.fits
-    psf_filename : str, optional
-        PSF filename. Default is psf_00.fits
-    drm_filename : str, optional
-        DRM filename. Default is drm_00.fits
-    obs_id : int, optional
-        Observation ID. Defalut is 0.
-
-    Returns
-    -------
-    observation : `~gammapy.data.Observation`
-        Observation with DL4 IRFs
-    """
-
-    path = Path(path)
-
-    events = EventList.read(path / events_filename)
-    counts = Map.read(path / counts_filename)
-    exposure = Map.read(path / exposure_filename)
-    psf = PSFMap.read(path / psf_filename, format="gtpsf")
-
-    edisp = _drm_to_edisp(
-        path / drm_filename, counts.geom, exposure.geom, psf.psf_map.geom
-    )
-
-    return Observation(
-        events=events,
-        psf=psf,
-        edisp=edisp,
-        aeff=exposure,
-        pointing=FixedPointingInfo(fixed_icrs=counts.geom.center_skydir.icrs),
-        obs_id=obs_id,
-    )
--- a/gammapy/data/observations.py
+++ b/gammapy/data/observations.py
@@ -18,7 +18,7 @@ from astropy.utils import lazyproperty
 import matplotlib.pyplot as plt
 from gammapy.irf.edisp.map import _drm_to_edisp
 from gammapy.irf import PSFMap
-from gammapy.map import Map
+from gammapy.maps import Map
 from gammapy.utils.deprecation import GammapyDeprecationWarning
 from gammapy.utils.fits import LazyFitsData, earth_location_to_dict
 from gammapy.utils.metadata import CreatorMetaData, TargetMetaData, TimeInfoMetaData
--- a/gammapy/data/observations.py
+++ b/gammapy/data/observations.py
@@ -6,6 +6,7 @@ import inspect
 import itertools
 import logging
 import warnings
+from pathlib import Path
 from itertools import zip_longest
 import numpy as np
 import astropy.units as u
@@ -15,6 +16,9 @@ from astropy.time import Time
 from astropy.units import Quantity
 from astropy.utils import lazyproperty
 import matplotlib.pyplot as plt
+from gammapy.irf.edisp.map import _drm_to_edisp
+from gammapy.irf import PSFMap
+from gammapy.map import Map
 from gammapy.utils.deprecation import GammapyDeprecationWarning
 from gammapy.utils.fits import LazyFitsData, earth_location_to_dict
 from gammapy.utils.metadata import CreatorMetaData, TargetMetaData, TimeInfoMetaData
@@ -27,7 +31,8 @@ from .gti import GTI
 from .metadata import ObservationMetaData
 from .pointing import FixedPointingInfo
 
-__all__ = ["Observation", "Observations"]
+
+__all__ = ["Observation", "Observations", "create_observation_from_fermi_files"]
 
 log = logging.getLogger(__name__)
 
@@ -910,3 +915,58 @@ class ObservationChecker(Checker):
         except Exception:
             yield self._record(level="warning", msg="Loading psf failed")
             return
+
+
+def create_observation_from_fermi_files(
+    path,
+    events_filename="ft1_00.fits",
+    counts_filename="ccube_00.fits",
+    exposure_filename="bexpmap_roi_00.fits",
+    psf_filename="psf_00.fits",
+    drm_filename="drm_00.fits",
+    obs_id=0,
+):
+    """Read Fermi-LAT files and create an Observation.
+
+    Parameters
+    ----------
+    path : str
+        path to files
+    events_filename : str
+        Events filename. Default is ft1_00.fits
+    counts_filename : str, optional
+        Counts filename. Default is ccube_00.fits
+    exposure_filename : str, optional
+        exposure filename. Default is bexpmap_roi_00.fits
+    psf_filename : str, optional
+        PSF filename. Default is psf_00.fits
+    drm_filename : str, optional
+        DRM filename. Default is drm_00.fits
+    obs_id : int, optional
+        Observation ID. Defalut is 0.
+
+    Returns
+    -------
+    observation : `~gammapy.data.Observation`
+        Observation with DL4 IRFs
+    """
+
+    path = Path(path)
+
+    events = EventList.read(path / events_filename)
+    counts = Map.read(path / counts_filename)
+    exposure = Map.read(path / exposure_filename)
+    psf = PSFMap.read(path / psf_filename, format="gtpsf")
+
+    edisp = _drm_to_edisp(
+        path / drm_filename, counts.geom, exposure.geom, psf.psf_map.geom
+    )
+
+    return Observation(
+        events=events,
+        psf=psf,
+        edisp=edisp,
+        aeff=exposure,
+        pointing=FixedPointingInfo(fixed_icrs=counts.geom.center_skydir.icrs),
+        obs_id=obs_id,
+    )
--- a/gammapy/data/observations.py
+++ b/gammapy/data/observations.py
@@ -27,7 +27,6 @@ from .gti import GTI
 from .metadata import ObservationMetaData
 from .pointing import FixedPointingInfo
 
-
 __all__ = ["Observation", "Observations"]
 
 log = logging.getLogger(__name__)
--- a/gammapy/datasets/io.py
+++ b/gammapy/datasets/io.py
@@ -1,20 +1,26 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
+import os
 import abc
 import numpy as np
+from pathlib import Path
 from astropy import units as u
 from astropy.io import fits
 from astropy.table import Table
 from gammapy.data import GTI
-from gammapy.irf import EDispKernel, EDispKernelMap
-from gammapy.maps import RegionNDMap
-from gammapy.utils.scripts import make_path
+from gammapy.irf import EDispKernel, EDispKernelMap, PSFMap
+from gammapy.maps import RegionNDMap, Map
+from gammapy.modeling.models import add_fermi_isotropic_diffuse_model
+from gammapy.utils.scripts import read_yaml, make_path
 from .spectrum import SpectrumDatasetOnOff
+from .utils import create_map_dataset_from_dl4
+from .datasets import MapDataset, Datasets
 
 __all__ = [
     "DatasetReader",
     "DatasetWriter",
     "OGIPDatasetReader",
     "OGIPDatasetWriter",
+    "FermipyDatasetsReader",
 ]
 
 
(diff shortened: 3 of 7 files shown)


gammapy/irf/edisp/map.py
```diff
@@ -538,3 +540,50 @@ def peek(self, figsize=(15, 5)):
...
+        )
+    if pdf_matrix.shape[0] != len(geom_true.axes["energy_true"].center) + 1:
+        raise ValueError(
+            "Energy true axes of exposure and DRM do not match. Check that edisp_bins are the same"
+        )
+
+    energy = geom.axes["energy"]
+    energy_true = axis.copy(name="energy_true", unit="MeV")
+
+    edisp = EDispKernel(axes=[energy_true, energy], data=pdf_matrix)
```
QRemy, line 584:
```suggestion
    edisp = EDispKernel(axes=[energy_true, energy], data=pdf_matrix.astype(np.float32))
```

gammapy/datasets/utils.py
```diff
@@ -203,3 +214,165 @@ def split_dataset(dataset, width, margin, split_template_models=True):
...
+    geom_exposure = geom.to_image().to_cube([energy_axis_true])
+
+    # create dataset and run data reduction / irfs interpolation
+    dataset = MapDataset.from_geoms(
+        geom,
+        geom_exposure=geom_exposure,
+        geom_psf=geom_psf,
+        geom_edisp=geom_edisp,
+        name=name,
+    )
```
LauraOlivera, line 269: @QRemy, did you figure out why sometimes having the IRFs in the original geom in HAWC analyses (so only 1 R.A. bin) caused issues with some of the estimators?

gammapy/data/observations.py
```diff
@@ -910,3 +915,58 @@ def check_psf(self):
...
+def create_observation_from_fermi_files(
+    path,
+    events_filename="ft1_00.fits",
+    counts_filename="ccube_00.fits",
+    exposure_filename="bexpmap_roi_00.fits",
+    psf_filename="psf_00.fits",
+    drm_filename="drm_00.fits",
+    obs_id=0,
+):
+    """Read Fermi-LAT files and create an Observation.
```
AtreyeeS, line 929: It will be good to mention that the events files is DL3 and the rest at DL4

gammapy/data/observations.py
```diff
@@ -910,3 +915,58 @@ def check_psf(self):
         except Exception:
             yield self._record(level="warning", msg="Loading psf failed")
             return
+
+
+def create_observation_from_fermi_files(
+    path,
+    events_filename="ft1_00.fits",
```
AtreyeeS, line 922: the defaults here are meaningless, no?

gammapy/data/observations.py
```diff
@@ -910,3 +915,58 @@ def check_psf(self):
...
+    Returns
+    -------
+    observation : `~gammapy.data.Observation`
+        Observation with DL4 IRFs
+    """
+
+    path = Path(path)
+
+    events = EventList.read(path / events_filename)
+    counts = Map.read(path / counts_filename)
```
AtreyeeS, line 957: So you dont really need the counts, only a geom?

gammapy/data/observations.py
```diff
@@ -910,3 +915,58 @@ def check_psf(self):
...
+    edisp = _drm_to_edisp(
+        path / drm_filename, counts.geom, exposure.geom, psf.psf_map.geom
+    )
+
+    return Observation(
+        events=events,
+        psf=psf,
+        edisp=edisp,
+        aeff=exposure,
+        pointing=FixedPointingInfo(fixed_icrs=counts.geom.center_skydir.icrs),
```
AtreyeeS, line 970: Why is this `FixedPointingInfo`?

gammapy/datasets/utils.py
```diff
@@ -203,3 +214,165 @@ def split_dataset(dataset, width, margin, split_template_models=True):
...
+        geom_image = obs.aeff.geom.to_image()
+        geom = geom_image.to_cube([obs.edisp.edisp_map.geom.axes["energy"]])
+
+    energy_axis = geom.axes["energy"]
+
+    if energy_axis_true is None:
+        # energy_axis_true = obs.edisp.edisp_map.geom.axes["energy_true"]
+        energy_axis_true = obs.aeff.geom.axes["energy_true"]
+
+    # ensure that DL4 IRFs have the axes
```
AtreyeeS, line 254: Why do DL4 irfs need the same energy_true?

gammapy/data/observations.py
QRemy: I think it's the default names of the files created by fermipy

gammapy/data/observations.py
QRemy: yes, counts are used only to get the `energy_axis` and check that the drm was generated with the same `edisp_bins` option than the exposure/counts (not by the default).
Note also that the counts map created by fermipy is by default in a different projection than exposure (AIT vs CAR) so we cannot use them as such in gammapy.

gammapy/datasets/utils.py
QRemy: I think in any case exposure and edsip must have the same energy_true for the evaluator to work.

gammapy/data/observations.py
QRemy: somehow it is optional but break the meta if not there

gammapy/data/observations.py
facero: This is also I question I had. In the gammapy tuto for Fermi, the counts cube is barely used and recreated from event list and I did not know why.
So you say it's related to this projection issue ?

facero:
Thanks @QRemy . Clearly this would help a lot.
One general comment is related to the event type analysis.
As you know many advanced Fermi analysis use the summed LL approach. 
Fermipy generates a list of products  `_00.fits`, `_01.fits`, etc 
For now you default every filenames to `_00.fits` which is the right to do.  

To carry out a summedLL, do you think one could just load 4 `Dataset` (00,01,02,03 id for example) into a `Datasets` for fitting ?

gammapy/datasets/utils.py
```diff
@@ -203,3 +214,165 @@ def split_dataset(dataset, width, margin, split_template_models=True):
...
+        DRM filename. Default is drm_00.fits
+    obs : `~gammapy.data.Observation`
+        Observation with DL4 IRFs
+    geom : `~gammapy.maps.WcsGeom`, optional
+        Output dataset maps geometry. The default is None, and it is derived from IRFS
+    energy_axis_true : `~gammapy.maps.MapAxis`, optional
+        True energy axis used for IRF maps. The default is None, and it is derived from IRFS
+    name : str, optional
+        Dataset name. The default is None, and the name is randomly generated.
+    isotropic_as_bkg : bool, optional
```
facero, line 356: This is another interesting point.
Should the iso and galdiff considered as bkg or not. 

The way I tried it, everything was loaded as a Sky model. But then the issue is that there bkg is empty and for example for the peek() the excess attribute doesn't work because there is no background.

It is not critical because the rest of the fitting & plotting functions but it introduces a difference between the IACT and Fermi Dataset if Fermi has empty bkg.
Both options are valid in the end, just a question of choosing the default behavior. 
In my opinion if `iso` is default in the background, `galdiff` should be as well no?

gammapy/irf/edisp/map.py
```diff
@@ -538,3 +540,50 @@ def peek(self, figsize=(15, 5)):
 
         """
         self.get_edisp_kernel().peek(figsize)
+
+
+def _drm_to_edisp(filename, geom, geom_true, geom_psf):
```
adonath, line 545: This function would be worth adding in an independent PR. We can support it as a new data format for DL4 IRFs similar to `gtpsf` for the PSF, see here: https://github.com/gammapy/gammapy/blob/main/gammapy/irf/core.py#L792

gammapy/datasets/utils.py
adonath: I am not sure there needs to be a convention for this. For IACTs we typically consider background as instrumental (and not astrophysical). If necessary the same logic could apply to Fermi, which would have no background set.

gammapy/datasets/utils.py
```diff
@@ -203,3 +214,165 @@ def split_dataset(dataset, width, margin, split_template_models=True):
...
+        models = Models()
+
+    diffuse_iso = create_fermi_isotropic_diffuse_model(
+        filename=isotropic_filepath, interp_kwargs={"extrapolate": True}
+    )
+    diffuse_iso.apply_irf["edisp"] = False
+
+    dataset.models = models + [diffuse_iso]
+
+    if as_bkg:
```
adonath, line 311: What is the need for this option? I think with all the caching and support for npred models there is no technical difference. In terms of book-keeping I would rather argue `.background` is considered an IRF and represents the instrumental background (misclassified events), which is negligible for Fermi.

gammapy/datasets/utils.py
```diff
@@ -203,3 +214,165 @@ def split_dataset(dataset, width, margin, split_template_models=True):
                 d.models = dataset.models
             datasets.append(d)
     return datasets
+
+
+def create_map_dataset_from_dl4_irfs(obs, geom=None, energy_axis_true=None, name=None):
```
adonath, line 219: I would recommend to introduce this as another DL4  data format in `MapDataset.read()`. I would check the fermipy documentation again, but ideally this would go as a `yaml` index file, pointing to the file locations and names (however there could be defaults, if fermipy filenames are fixed). Maybe there can be additional support for the single `.npy` session file, that fermipy has.

gammapy/datasets/utils.py
```diff
@@ -203,3 +214,165 @@ def split_dataset(dataset, width, margin, split_template_models=True):
...
+    diffuse_iso.apply_irf["edisp"] = False
+
+    dataset.models = models + [diffuse_iso]
+
+    if as_bkg:
+        dataset.background = dataset.npred()
+        dataset.models = models + [FoVBackgroundModel(dataset_name=dataset.name)]
+
+
+def create_map_dataset_from_fermi_files(
```
adonath, line 316: See my comment above. I think this should move to `datasets/io.py` as a new `FermiPyDatasetReader`. Such that `MapDataset.read("fermipy-index.yaml")` or `MapDataset.read("path/to/fermipy/session")` (in case of fixed filenames) works.

gammapy/datasets/utils.py
adonath: Yes, but this should be guaranteed for fermipy as well. Or does it output inconsistent energy axis definitions?

adonath (changes requested):
Thanks @QRemy! This is useful functionality and simplifies the setup for users of Fermipy. I would strongly recommend to split up the functionality and turn it into independent PRs introducing new DL4 data formats:

- Add a new data format for `EDispKernelMap` to read the output of `gtdrm`
- Add a new data format for `MapDataset.read(..., format='fermipy')`, which supports reading the files generated by `fermipy`. Possibly support for the session `.npy` file as well (however I am not sure what is actually stored there)

The remaining question is what to do with the re-projection / event binning step. I think using the `Observation` object is only weakly motivated, especially because event filtering etc. will not work. There is also no need for lazy loading etc. because Fermi is almost always a single observation. We will never represent Fermi data in GTIs and do stacking or similar operations. The meaning of obs-id is also weak and pointing needs to be invented etc.

I would recommend to skip the re-projection and just keep what we have in the https://docs.gammapy.org/1.3/tutorials/data/fermi_lat.html#sphx-glr-tutorials-data-fermi-lat-py tutorial. Where the re-projection is very limited effort. 

This gives us some time to think about conceptional cleaner approaches, while already providing a lot of convenience (support for fermipy files) to users.

gammapy/datasets/utils.py
QRemy: As far as I understand by default fermipy `compute_drm` use `edisp_bins=1` and not the value set in the config see [here](https://fermipy.readthedocs.io/en/latest/_modules/fermipy/gtanalysis.html#GTAnalysis.compute_drm).
Moreover from what I tested using negative edisp_bins<0 in `compute_drm` also add extra bins to the drm but that's not the case for the exposure. So only `edisp_bins>0` is compatible with what is expected by gammapy.

adonath: I see, but is the binning still aligned? So a simple slicing would remove the additional bin and make it consistent?

gammapy/irf/edisp/map.py
```diff
@@ -538,3 +540,50 @@ def peek(self, figsize=(15, 5)):
...
+        raise ValueError(
+            "Energy axes of counts and DRM do not match. Check that edisp_bins are the same"
+        )
+    if pdf_matrix.shape[0] != len(geom_true.axes["energy_true"].center) + 1:
+        raise ValueError(
+            "Energy true axes of exposure and DRM do not match. Check that edisp_bins are the same"
+        )
+
+    energy = geom.axes["energy"]
+    energy_true = axis.copy(name="energy_true", unit="MeV")
```
adonath, line 582: Is the DRM really always a square matrix? I guess not, it depends on the `edisp_bins` parameter. I have checked and the HDU contains the `DETCHANS` keyword, which tells you the number of channels in the "reco" axis. Bases on this information you should be able to construct both axes correctly. As far as I understood, the "true" axis is aligned with the "reco" axis, but just contains the additional ("appended") bins.

However this still has the issue, that the `gtdrm` does not output a matrix that aligns with the exposure. So some kind of resampling is needed for the RMF matrix, to have it on the same true energy axis as the exposure.

adonath: I have implement a "quick and dirty" version here: https://github.com/adonath/snakemake-workflow-fermi-lat/commit/f9d1f32563b752dea8bfa472acf9ee3aa3c1f110#diff-c8c28ef1325ba68f3569569e41d0f245f8a80cab8b810a30dd46c3bd2fda9779R21

This might be a good starting point. I would be happy to delete it once the drm file format is supported in Gammapy.

gammapy/datasets/utils.py
QRemy: no they are shifted by half a bin, exposure centres are edisp edges (same as counts but with `edisp_bins` more in true energy)

gammapy/datasets/utils.py
facero: agreed
but If the bkg is always None. Then some functions don't work like the excess in the `peek()` functions or when printing the `Dataset`. 
Maybe be not a big issue, but the first time I swa these errors it took me sometime to understand why.

gammapy/data/observations.py
```diff
@@ -27,6 +27,7 @@
 from .metadata import ObservationMetaData
 from .pointing import FixedPointingInfo
 
+
```
QRemy, line 30:
```suggestion
```

gammapy/data/__init__.py
```diff
@@ -1,5 +1,6 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 """Data and observation handling."""
+
```
QRemy, line 3:
```suggestion
```

gammapy/datasets/io.py
```diff
@@ -472,3 +477,167 @@ def read(self):
...
+            )
+
+        if isotropic_file:
+            model = create_fermi_isotropic_diffuse_model(
+                isotropic_file, datasets_names=[dataset.name]
+            )
+            dataset.models = Models([model])
+        return dataset
+
+    def read(self):
```
AtreyeeS, line 594: Isnt it more reasonable to have the filename passed on read, rather than on init?

gammapy/datasets/io.py
```diff
@@ -472,3 +477,167 @@ def read(self):
...
+                path = Path(data["fileio"]["outdir"])
+            else:
+                path = Path("")
+
+            if "model" in component and "isodiff" in component["model"]:
+                isotropic_file = Path(component["model"]["isodiff"])
+                name = isotropic_file.stem[4:]
+            elif "model" in data and "isodiff" in data["model"]:
+                isotropic_file = Path(data["model"]["isodiff"])
+                name = isotropic_file.stem[4:]
```
AtreyeeS, line 629: This reading is tied strictly to the current fermipy file format?
We should have an info about the version we support.

gammapy/datasets/utils.py
```diff
@@ -203,3 +208,82 @@ def split_dataset(dataset, width, margin, split_template_models=True):
                 d.models = dataset.models
             datasets.append(d)
     return datasets
+
+
+def create_map_dataset_from_dl4(data, geom=None, energy_axis_true=None, name=None):
```
AtreyeeS, line 213: It should be `obs`, not `data`. Or, edit the docstring.

gammapy/datasets/utils.py
```diff
@@ -203,3 +208,82 @@ def split_dataset(dataset, width, margin, split_template_models=True):
...
+    """Create a map dataset from a map dataset or an observation containing DL4 IRFs
+
+    Parameters
+    ----------
+    obs : `~gammapy.dataset.MapDataset` or `~gammapy.data.Observation`
+        MapDataset or Observation containing DL4 IRFs
+    geom : `~gammapy.maps.WcsGeom`, optional
+        Output dataset maps geometry. The default is None, and it is derived from IRFS
```
AtreyeeS, line 221:
```suggestion
        Output dataset maps geometry. The default is None, and it is derived from IRFs
```

gammapy/datasets/utils.py
```diff
@@ -203,3 +208,82 @@ def split_dataset(dataset, width, margin, split_template_models=True):
...
+    energy_axis_true : `~gammapy.maps.MapAxis`, optional
+        True energy axis used for IRF maps. The default is None, and it is derived from IRFS
```
AtreyeeS, line 223:
```suggestion
        True energy axis used for IRF maps. The default is None, and it is derived from IRFs
```

gammapy/datasets/utils.py
```diff
@@ -203,3 +208,82 @@ def split_dataset(dataset, width, margin, split_template_models=True):
...
+
+        geom = geom_image.to_cube([data.edisp.edisp_map.geom.axes["energy"]])
+
+    energy_axis = geom.axes["energy"]
+
+    if energy_axis_true is None:
+        energy_axis_true = data.edisp.edisp_map.geom.axes["energy_true"]
+
+    # ensure that DL4 IRFs have the axes
+    rad_axis = data.psf.psf_map.geom.axes["rad"]
```
AtreyeeS, line 250: Why not allow the user to configure `rad_axis` as well?

gammapy/datasets/utils.py
```diff
@@ -203,3 +208,82 @@ def split_dataset(dataset, width, margin, split_template_models=True):
...
+        selection.append("background")
+
+    maker = MapDatasetMaker(selection=selection)
+    dataset = maker.run(dataset, data)
+
+    if isinstance(data, MapDataset) and data.counts:
+        if dataset.counts.geom == data.counts.geom:
+            dataset.counts.data = data.counts.data
+        else:
+            raise ValueError(
```
AtreyeeS, line 279: We can also interpolate the counts map, no?

gammapy/datasets/utils.py
```diff
@@ -203,3 +208,82 @@ def split_dataset(dataset, width, margin, split_template_models=True):
                 d.models = dataset.models
             datasets.append(d)
     return datasets
+
+
+def create_map_dataset_from_dl4(data, geom=None, energy_axis_true=None, name=None):
+    """Create a map dataset from a map dataset or an observation containing DL4 IRFs
```
AtreyeeS, line 214: The DL4 IRFs should be specified, ie `EdispMap` `PSFMap` and `WcsNDMap`

AtreyeeS (commented):
Thanks @QRemy ! Please see inline comments

gammapy/datasets/io.py
QRemy: I used the same logic than for the `OGIPDatasetReader`

gammapy/datasets/io.py
```diff
@@ -472,3 +477,167 @@ def read(self):
...
 
         return SpectrumDatasetOnOff(name=name, exposure=exposure, **kwargs)
+
+
+class FermipyDatasetsReader(DatasetReader):
+    """Create datasets from Fermi-LAT configuration file.
+
+    Parameters
+    ----------
+    path : str
```
QRemy, line 487:
```suggestion
    filename : str
```

gammapy/datasets/io.py
```diff
@@ -472,3 +477,167 @@ def read(self):
...
+        Configuration file path
+    edisp_bins : int
+        Number of margin bins to slice in energy. Default is 0.
+        If fermipy was configured with edisp_bins=0, it should be set to a value
+        edisp_bins>0 here in order to apply the energy dispersion correclty.
```
QRemy, line 492:
```suggestion

        For now only maps created in fermipy with edisp_bins=0 are supported,
         in that case the emin/emax in the fermipy configuration will correspond to the true energy range for gammapy,
          and  a value edisp_bins>0 should be set here in order to apply the energy dispersion correctly.
```

QRemy, line 492:
```suggestion
        For now only maps created in fermipy with edisp_bins=0 are supported,
         in that case the emin/emax in the fermipy configuration will correspond to the true energy range for gammapy,
          and  a value edisp_bins>0 should be set here in order to apply the energy dispersion correctly.
        With a binning of 8 to 10 bins per decade, it is recommended to use edisp_bins ≥ 2
        (See https://fermi.gsfc.nasa.gov/ssc/data/analysis/documentation/Pass8_edisp_usage.html)
```

gammapy/datasets/utils.py
QRemy: We could reproject to be sure it's preserving counts but I prefer to avoid it.

gammapy/datasets/io.py
```diff
@@ -472,3 +477,167 @@ def read(self):
...
+        Returns
+        -------
+        dataset : `~gammapy.datasets.MapDataset`
+            Map dataset.
+
+        """
+        from gammapy.datasets import MapDataset
+
+        path = Path(path)
+        counts = Map.read(path / f"ccube_0{str(file_id)}.fits")
```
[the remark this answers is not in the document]
QRemy: I think it is simpler to explain how to configure and run fermipy for the conversion to dataset to work. We will have only one way to create and read Fermi-LAT data to be used with gammapy but we will be sure this is a good way.

gammapy/datasets/io.py
```diff
@@ -472,3 +477,167 @@ def read(self):
...
+class FermipyDatasetsReader(DatasetReader):
+    """Create datasets from Fermi-LAT configuration file.
+
+    Parameters
+    ----------
+    path : str
+        Configuration file path
+    edisp_bins : int
+        Number of margin bins to slice in energy. Default is 0.
+        If fermipy was configured with edisp_bins=0, it should be set to a value
```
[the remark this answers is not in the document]
QRemy: I added more explainations here

gammapy/datasets/io.py
```diff
@@ -472,3 +477,167 @@ def read(self):
...
+        dataset : `~gammapy.datasets.Datasets`
+            Map datasets.
+
+        """
+        from gammapy.datasets import Datasets
+
+        filename = self.filename.resolve()
+        cwd = os.getcwd()
+        data = read_yaml(filename)
+        os.chdir(filename.parent)
```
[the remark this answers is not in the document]
QRemy: yes now it should work without that

gammapy/datasets/io.py
```diff
@@ -472,3 +476,172 @@ def read(self):
...
+        and  a value edisp_bins>0 should be set here in order to apply the energy dispersion correctly.
+        With a binning of 8 to 10 bins per decade, it is recommended to use edisp_bins ≥ 2
+        (See https://fermi.gsfc.nasa.gov/ssc/data/analysis/documentation/Pass8_edisp_usage.html)
+
+    """
+
+    tag = "fermipy"
+
+    def __init__(self, filename, edisp_bins=0):
+        self.filename = make_path(filename)
```
AtreyeeS, line 501: Somehow, for reading a file, I think its easier to specify the filename in the function rather than through the class

AtreyeeS (commented):
Thanks @QRemy ! My only remaining issue is how strictly dependent we want to be on the fermi-lat naming/file output scheme.

The current functions might break with changes in fermipy, and we are not enforcing fermipy versioning anywhere. 

@bkhelifi @facero Do you think this is safe or we should have some checks?

gammapy/datasets/io.py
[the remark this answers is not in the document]
QRemy: No, we don't have add fermipy as optinal dependency but we can show an example configuration file in the tutorial.

gammapy/datasets/io.py
QRemy: I used the same logic than for the OGIPDatasetReader, so if we change that in should be modified consitently for all Reader class  in a separated PR.
On a longer term I expect the mismatch between true energy axis of the different irfs for edisp_bins>0 to be fixed on the fermipy side. Then won't need that option and so user won't have to use this `FermiDatasetsReader` but would use directly  `Datasets.read(format="fermipy")`.

QRemy:
> Thanks @QRemy ! My only remaining issue is how strictly dependent we want to be on the fermi-lat naming/file output scheme.
> 

Thanks @AtreyeeS for the review
I changed  `create_dataset` to take all the file names

gammapy/datasets/io.py
```diff
@@ -472,3 +476,180 @@ def read(self):
             )
 
         return SpectrumDatasetOnOff(name=name, exposure=exposure, **kwargs)
+
+
+class FermipyDatasetsReader(DatasetReader):
```
MRegeard, line 481: Some Fermi-LAT user are not using Fermipy but only the Fermitools with which you define your own name for each file. So it is probably a good idea to support that. 
One thought would be to init the class with a filename for each file (i.e. counts_file, exposure_file, etc.) and create a classmethod `from_fermipy_config` that would init the class as you currently do.

MRegeard (commented):
Thanks @QRemy. 
See my inline comment.

gammapy/datasets/io.py
QRemy: Now `create_dataset` takes each file names as input and its a static method so it can be used without init.
The init keep only the fermipy config file for simplicity.

MRegeard: I see, clever !
What about the name though ? Do we want to keep Fermipy or simply Fermi ?

mireianievas:
I also think keeping the possibility to specify filenames is useful and the way to go. I personally don't use fermipy and I don't plan to migrate it, probably there are others in the same situation.

gammapy/datasets/io.py
facero: I think the current implementation is good. Have some default to fermipy and allow other names.

gammapy/datasets/io.py
```diff
@@ -472,3 +476,180 @@ def read(self):
...
+    """Create datasets from Fermi-LAT configuration file.
+
+    Parameters
+    ----------
+    filename : str
+        Configuration file path
```
AtreyeeS, line 487:
```suggestion
        Path to Fermipy  (v1.3.1) configuration file 
```

AtreyeeS (dismissed):
Thank you @QRemy for the changes! Looks good to me now.

My only remaining comment is that we should mention the supported Fermipy version somewhere

gammapy/datasets/io.py
QRemy: added

QRemy:
Now that all the comments are implemented i think we are good to merge.

Issue #5484, opened 2024-10-11
Add a tutorial showing the Gammapy use for data with large PSF


bkhelifi:
**Is your feature request related to a problem? Please describe.**
When data are associated to large PSFs (typically >0.5deg), one can face to several issues: large computing time, large RAM use, non-converging git. This can happen for some IACT event types, HAWC/SWGO, Fermi <1GeV, or in the future maybe KM3NeT. These difficulties are well known.

Here is a preliminary list made by @QRemy : 

> Regarding the fit I use the exact same workflow for CTA, HESS, and HAWC catalogues.
> The tricks are mostly the same :
> - start with good guess an the parameters and always set limits
> - start with norm only, then spectral fit + radius, then everything
> - fit together only groups of sources that are nearby (overlapping via psf)
> - use strategy zero, with large tolerance at first, then go to strategy 1 with lower tolerance, and eventually strategy 2.
> - in complex regions start from catalogues don't do the iterative fitting from scratch  eventually add or remove sources depending on your source of interest (like in Fermi-LAT).
> - freeze sources that are in the dataset but outside of mask_fit
> 
> For HAWC the specificity are :
> - the PSF is in reco energy so the caching works if only the norm is free.
> - the 99.9 containment radius computed automatically by gammapy to set the evaluation region does not take into account the mask safe so it can be very large if the PSF has cray values (which is the case in some bins). For that we can do :
>  `gammapy.evaluators import PSF_MAX_RADIUS`
>  `PSF_MAX_RADIUS="a good value"` # typically I compute the 99% containment within the mask safe
> - the mask safe has to discard the  bins (in fhit/energy/dec) where the PSF and EDisp have aberrant values or large error on the fitted model.
>  (for now there is no safe mask maker for that, but the datasets that I provide already include the mask_safe with that kind of selection)
> -the default binning of the maker for the psf/edisp is not suited for HAWC
>  but there is no need to use the maker one can directly use the IRFS provided in ra-dec
>  (it has to be on the same frame that the dataset only for the TSmap computation then interpolation to a downsampled dataset geom with only 4 spatial bin is enough) 
> 

As seen, it impacts both the data reduction step and the fitting one.

**Additional context**
This future tutorial should be then be mentioned into : 
- the HAWC tutorial, 
- the HowTo "Improve fit convergence with constraints on the source position",
- the HowTo "Reduce memory budget for large datasets",

[...]

PR #5650, opened 2024-12-19
Use global variable as default in  dataset._psf_kernel computation

QRemy:
Use global variable as default in  dataset._psf_kernel computation for consistency with MapEvaluator. This does not change the default behaviour.


adonath (commented):
Thanks @QRemy, makes sense! Can you pleas add a small unit test to establish the behavior for the future.

adonath:
One could create or read PSF Map with large rad_max and make sure the kernel is cropped correctly as a test.

registerrier (approved):

bkhelifi:
Is it a bit associated with the feature request #5484 ?

QRemy:
> Is it a bit associated with the feature request #5484 ?

Partly It affects only Fermi-LAT analysis where a `psf_kernel` is pre-computed each dataset but I introduced that only for consistency with the behaviour of the MapEvaluator that already rely on these global properties.

PR #5651, opened 2024-12-20
Overwrite name in ogip spectrum dataset reading

MRegeard:
This PR add the possibility to overwrite ogip spectrum dataset name during reading.

### DIFF
--- a/codemeta.json
+++ b/codemeta.json
@@ -359,7 +359,7 @@
         "active"
     ],
     "email": "GAMMAPY-COORDINATION-L@IN2P3.FR",
-    "dateModified": "2024-11-26",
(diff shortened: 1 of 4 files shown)


adonath (changes requested):
Thanks @MRegeard, this make a lot of sense. I would just document the `name` argument explicitly instead of the implicit `**kwargs`. There is only the `name` argument anyway. Otherwise LGTM.

adonath (approved):
Thanks @MRegeard, looks good to me now!

adonath:
Please add a 2nd review @registerrier  or @QRemy  or @AtreyeeS!

QRemy (approved):
Thanks @MRegeard looks good.

PR #5653, opened 2025-01-02
Preserve model names in cutout_template_models

QRemy:
Preserve model names in cutout_template_models


AtreyeeS (approved):
Thanks @QRemy ! No comments

registerrier (approved):

registerrier (commented):
On second thought given the CI error, given the error in `split_dataset` the behavior has to be changed. Maybe make the name propagation optional?

BTW the docstring might be more explicit on what is done, and for instance what is the role of the `dataset_models`  argument.

registerrier (approved):

PR #5654, opened 2025-01-02
Support "gtdrm" format in .read() of EDispKernelMap and EDispKernel

QRemy:
Support "gtdrm" format in .read() of EDispKernelMap and EDispKernel
Spitted from #5649

I'll add the test once the drm a test file is added in gammapy.data 
Requires https://github.com/gammapy/gammapy-data/pull/60


adonath (approved):
Thanks @QRemy, look good to me!

gammapy/irf/edisp/map.py
```diff
@@ -538,3 +538,34 @@ def peek(self, figsize=(15, 5)):
...
+        Returns
+        -------
+        edisp : `EDispKernelMap`
+            Energy dispersion kernel map.
+
+        """
+
+        if format == "gadf":
+            return super().read(filename, format, hdu, checksum)
+        elif format == "gtdrm":
```
registerrier, line 566: Is there some possibility to verify the checksum in the drm case or is it irrelevant?

gammapy/irf/edisp/map.py
```diff
@@ -538,3 +538,34 @@ def peek(self, figsize=(15, 5)):
 
         """
         self.get_edisp_kernel().peek(figsize)
+
+    @classmethod
+    def read(cls, filename, format="gadf", hdu=None, checksum=False):
+        """Read an EDispKernelMap from file and create corresponding object.
```
registerrier, line 544: Here I would explain that this does not only read a map but create one assuming spatial uniformity when format is drm. 
Ideally this would deserve a different classmethod, but I suppose this is fine like this.

gammapy/irf/edisp/map.py
registerrier: Maybe a log.info to explicitly write that a single bin EDispKernel map is created?

registerrier (approved):
Thanks @QRemy . I have left a few small inline comments. 

In general, a clarifcation of the API regarding formats will be needed as the optional formats vary a lot between different methods. But that is subsequent work.

gammapy/irf/edisp/map.py
QRemy: added

adonath: @registerrier The good thing is, that the this information is still explicit in the geom of the resulting map. So no meta info is lost and creating missing one on I/O is fine.

PR #5655, opened 2025-01-02
Support irf exposure map creation from dl4 irf with MapDatasetMaker

QRemy:
Support irf exposure map creation from dl4 irf with MapDatasetMaker
Related to #5649


gammapy/makers/map.py
```diff
@@ -319,10 +325,15 @@ def make_psf(self, geom, observation):
...
-        elif isinstance(psf, PSFMap):
-            return PSFMap(psf.psf_map.interp_to_geom(geom))
+        if isinstance(psf, PSFMap):
+            if psf.exposure_map:
+                exposure_map = psf.exposure_map.interp_to_geom(
+                    geom.squash(axis_name="rad")
+                )
+            else:
+                exposure_map = None
+            return psf.__class__(psf.psf_map.interp_to_geom(geom), exposure_map)
```
AtreyeeS, line 335: If exposure_map is None, the interp would return all nans, no?

AtreyeeS (approved):
Thanks @QRemy !

gammapy/makers/map.py
QRemy: `exposure_map` can be None bur not `psf_map`

PR #5656, opened 2025-01-02
Implicit support of map dataset instead of observation in DatasetMaker

QRemy:
Implicit support of map dataset instead of observation in DatasetMaker.
Related to #5649 : If ones don't want to create an Observation object for Fermi-LAT data (and  in general data with DL4 irfs), we could instead support using a map dataset as input irfs container in makers.  This would require very little change to work as shown in this draft PR.

### DIFF
--- a/gammapy/makers/map.py
+++ b/gammapy/makers/map.py
@@ -166,7 +166,7 @@ class MapDatasetMaker(Maker):
         exposure : `~gammapy.maps.Map`
             Exposure map.
         """
-        if getattr(observation, "exposure"):
+        if getattr(observation, "exposure", None):
             return observation.exposure.interp_to_geom(
                 geom=geom,
             )
--- a/gammapy/makers/map.py
+++ b/gammapy/makers/map.py
@@ -401,7 +401,7 @@ class MapDatasetMaker(Maker):
         if isinstance(observation, Observation):
             kwargs["meta_table"] = self.make_meta_table(observation)
             kwargs["meta"] = self._make_metadata(kwargs["meta_table"])
-        elif isinstance(observation, MapDataset):
+        elif getattr(observation, "meta"):
             kwargs["meta"] = observation.meta
 
         mask_safe = Map.from_geom(dataset.counts.geom, dtype=bool)
--- a/gammapy/makers/map.py
+++ b/gammapy/makers/map.py
@@ -5,6 +5,7 @@ from astropy.table import Table
 from regions import PointSkyRegion
 from gammapy.datasets import MapDatasetMetaData
 from gammapy.irf import EDispKernelMap, PSFMap
+from gammapy.data import Observation
 from gammapy.maps import Map
 from .core import Maker
 from .utils import (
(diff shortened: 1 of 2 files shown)


gammapy/makers/map.py
```diff
@@ -165,6 +166,10 @@ def make_exposure(geom, observation, use_region_center=True):
         exposure : `~gammapy.maps.Map`
             Exposure map.
         """
+        if getattr(observation, "exposure"):
```
QRemy, line 169:
```suggestion
        if getattr(observation, "exposure", None):
```

AtreyeeS (commented):
Thanks @QRemy !
I agree that this is useful.

I think this needs some more documentation (a minima, editing the doc string to say that `observation` can also be a `MapDataset` object).
We need to make it clear how/when to use this with a Dataset.
In general, we have often emphasied that Makers take a DL3 object to the DL4 level, and this breaks that logic. I guess it is more evident for people from Fermi/HAWC, but if we are officially supporting this scheme, it should be better explained.

gammapy/makers/map.py
```diff
@@ -382,8 +387,11 @@ def run(self, dataset, observation):
             Map dataset.
         """
         kwargs = {"gti": observation.gti}
-        kwargs["meta_table"] = self.make_meta_table(observation)
-        kwargs["meta"] = self._make_metadata(kwargs["meta_table"])
+        if isinstance(observation, Observation):
+            kwargs["meta_table"] = self.make_meta_table(observation)
+            kwargs["meta"] = self._make_metadata(kwargs["meta_table"])
+        else:
+            kwargs["meta"] = observation.meta
```
bkhelifi, line 394: If `observation` is not an Observation, what is it then???

QRemy: a map dataset

gammapy/makers/map.py
```diff
@@ -5,6 +5,7 @@
 from regions import PointSkyRegion
 from gammapy.datasets import MapDatasetMetaData
```
QRemy, line 6:
```suggestion
from gammapy.datasets import MapDatasetMetaData, MapDataset
```

gammapy/makers/map.py
```diff
@@ -382,8 +387,11 @@ def run(self, dataset, observation):
             Map dataset.
         """
         kwargs = {"gti": observation.gti}
-        kwargs["meta_table"] = self.make_meta_table(observation)
-        kwargs["meta"] = self._make_metadata(kwargs["meta_table"])
+        if isinstance(observation, Observation):
+            kwargs["meta_table"] = self.make_meta_table(observation)
+            kwargs["meta"] = self._make_metadata(kwargs["meta_table"])
+        else:
```
QRemy, line 393:
```suggestion
        elif isinstance(observation, MapDataset):
```

gammapy/makers/map.py
bkhelifi: But, the first argument of the function is a dataset:
> def run(self, dataset, observation):
 So, one can have 2 dataset objects when we call the function?

QRemy: yes

bkhelifi: So the docstring should be updated!
And even wondering if the name of the argument (`observation`) is now obsolete...

PR #5657, opened 2025-01-05
Add `datasets_names` argument to `create_fermi_isotropic_diffuse_model`

QRemy:
Splitted from #5649


AtreyeeS (commented):
Thanks @QRemy !
I am wondering if it is better to have the changes directly in `create_isotropic_diffuse_model()` which can optionally take a `dataset_name` as well.
Returning a `model` maybe be easier that directly setting it on `dataset.models`, so it can be maintained in some `models` list to keep track of later..
```
models = Models()
datasets.models = models
```

QRemy:
Thanks @AtreyeeS I simplified the PR as you suggested.

AtreyeeS (commented):
Thanks @QRemy ! Maybe you can also set the default kwargs you had earlier? Like `apply_edisp=None` or the inertpolation scale?

gammapy/modeling/models/cube.py
```diff
@@ -1202,7 +1202,7 @@ def unfreeze(self, model_type="spectral"):
             self._spectral_model.unfreeze()
 
 
-def create_fermi_isotropic_diffuse_model(filename, **kwargs):
+def create_fermi_isotropic_diffuse_model(filename, dataset_name=None, **kwargs):
```
AtreyeeS, line 1205: I dont know if its better to have `datasets_names` here, because thats what most models take as argument, and there might be cases (eg: fitting an EBL model to multiple blazar observations simultaneously) where there are multiple fermi datasets

AtreyeeS (approved):
Thanks @QRemy ! 
I left a couple of comments inline, but you can decide if they are relevant

gammapy/datasets/tests/test_map.py
```diff
@@ -2312,3 +2313,16 @@ def test_create_empty_map_dataset_from_irfs(geom, geom_etrue):
...
+@requires_data()
+def test_add_fermi_iso():
+    dataset = MapDataset.read(
+        "$GAMMAPY_DATA/fermi-3fhl-gc/fermi-3fhl-gc.fits.gz", format="gadf"
+    )
+    filename = "$GAMMAPY_DATA/fermi_3fhl/iso_P8R2_SOURCE_V6_v06.txt"
+    model = create_fermi_isotropic_diffuse_model(filename, dataset_name=dataset.name)
+    assert dataset.name in model.datasets_names
+    dataset.models = model
+    assert "isotropic" in dataset.models.names[0]
```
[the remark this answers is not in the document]
QRemy: here we that that "isotropic is in the name not in the list of names

gammapy/modeling/models/cube.py
AtreyeeS: And if there are multiple psf classes being fit simulatneously?

adonath: @QRemy But I guess there is the use case of e.g.  a joint HESS / Fermi analysis with PSF classes. How is this handled correctly? In this case you only want the iso model to go to the multiple  Fermi datasets and not to the HESS ones.

AtreyeeS (approved):
Thanks @QRemy ! No further comments!

adonath (approved):
Looks good to me as well!

PR #5658, opened 2025-01-05
Adapt hierarchical_clustering test

QRemy:
Adapt the test of `gammapy.utils.cluster.hierarchical_clustering` in order to fix the CI. This is related to changes is scipy.cluster with version 1.15.

### DIFF
--- a/gammapy/data/tests/test_observations.py
+++ b/gammapy/data/tests/test_observations.py
@@ -515,7 +515,7 @@ def test_observations_clustering(data_store):
         names=names,
         apply_standard_scaler=True,
     )
-    features = hierarchical_clustering(features)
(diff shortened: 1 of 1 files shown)


registerrier (commented):
Thanks @QRemy . Looks good.

Does it mean that more arguments should be passed to the `hierarchical_clustering` function? Does the API tutorial require some change?

QRemy:
> Does it mean that more arguments should be passed to the `hierarchical_clustering` function? Does the API tutorial require some change?

No need for more argument but the result of the clustering might change in some case as a bug was fixed in the algorithm in scipy.  For the tutorial the max number of clusters is already 2 (`fcluster_kwargs={"t": 2}`)  and nothing change.

registerrier (approved):

PR #5499, opened 2024-10-14
PIG Bayesian inference for nested sampling

facero:
This pull request starts the draft of a PIG about Bayesian Inference in gammapy.
The draft is still incomplete and will be completed soon but enables to collaborate on the code and the justification.

Will comment when a first draft is complete

PR #5662, opened 2025-01-07
Splitting original stat_sum with & without priors

facero:
This PR is a small brick of the larger Nested Sampling discussed in the PIG #5499 

We introduce `stat_sum_no_prior` that returns the statistics without the prior term. 
`stat_sum_no_prior` is now reused in the `stat_sum` function.
This function is private to avoid overloading.

### DIFF
--- a/gammapy/datasets/core.py
+++ b/gammapy/datasets/core.py
@@ -69,14 +69,18 @@ class Dataset(abc.ABC):
 
     def stat_sum(self):
         """Total statistic given the current model parameters and priors."""
+        prior_stat_sum = 0.0
+        if self.models is not None:
+            prior_stat_sum = self.models.parameters.prior_stat_sum()
+        return self._stat_sum_no_prior() + prior_stat_sum
+
+    def _stat_sum_no_prior(self):
+        """Total statistic given the current model parameters without the priors."""
         stat = self.stat_array()
 
         if self.mask is not None:
             stat = stat[self.mask.data]
-        prior_stat_sum = 0.0
-        if self.models is not None:
-            prior_stat_sum = self.models.parameters.prior_stat_sum()
-        return np.sum(stat, dtype=np.float64) + prior_stat_sum
+        return np.sum(stat, dtype=np.float64)
 
     @abc.abstractmethod
     def stat_array(self):
--- a/gammapy/datasets/core.py
+++ b/gammapy/datasets/core.py
@@ -72,9 +72,9 @@ class Dataset(abc.ABC):
         prior_stat_sum = 0.0
         if self.models is not None:
             prior_stat_sum = self.models.parameters.prior_stat_sum()
-        return self._stat_sum_no_prior() + prior_stat_sum
+        return self._stat_sum_likelihood() + prior_stat_sum
 
-    def _stat_sum_no_prior(self):
+    def _stat_sum_likelihood(self):
         """Total statistic given the current model parameters without the priors."""
         stat = self.stat_array()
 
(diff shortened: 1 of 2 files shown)


AtreyeeS (commented):
Thanks @facero ! Can you please add a test?

Dont we need this on the `MapDataset` as well?

adonath:
Thanks @facero! In terms of naming I think the clearest option would be to have:

- `stat_sum()`
- `stat_sum_prior()`
- `stat_sum_posterior()`

But I think we already deviated from that by adding the prior to `.stat_sum()`, so `.stat_sum_no_prior()` is not the the worst option, but I don't really like it in terms of how it maps to the statistical concept. The only other term might be `.stat_sum_likelihood()`.

facero:
> Thanks @facero! In terms of naming I think the clearest option would be to have:
> 
>     * `stat_sum()`
> 
>     * `stat_sum_prior()`
> 
>     * `stat_sum_posterior()`
> 
> 
> But I think we already deviated from that by adding the prior to `.stat_sum()`, so `.stat_sum_no_prior()` is not the the worst option, but I don't really like it in terms of how it maps to the statistical concept. The only other term might be `.stat_sum_likelihood()`.

I think `stat_sum_likelihood()` could work indeed. We ideally want to have it transparent if no priors are defined and to keep `stat_sum` for with/without priors.

gammapy/datasets/core.py
```diff
@@ -237,12 +237,19 @@ def contributes_to_stat(self):
...
+            prior_stat_sum = self.models.parameters.prior_stat_sum()
+        return self._stat_sum_likelihood() + prior_stat_sum
+
+    def _stat_sum_likelihood(self):
+        """Total statistic given the current model parameters without the priors."""
+        stat = self.stat_array()
+
+        if self.mask is not None:
+            stat = stat[self.mask.data]
+        return np.sum(stat, dtype=np.float64)
```
QRemy, line 252: `stat_sum` should stay the same, and add
```
     def _stat_sum_likelihood(self):
         """Total statistic given the current model parameters without the priors."""
        stat_sum = 0
        for dataset in self:
            stat_sum += dataset. _stat_sum_likelihood()
        return stat_sum
```

facero: indeed my bad, a bad copy/paste
fixed now

facero:
I think this PR is now ready for review

QRemy (approved):
Thanks @facero no further comment.

gammapy/datasets/tests/test_map.py
```diff
@@ -755,6 +755,7 @@ def test_prior_stat_sum(sky_model, geom, geom_etrue):
 
     uniformprior = UniformPrior(min=0, max=np.inf, weight=1)
     datasets.models.parameters["amplitude"].prior = uniformprior
+    assert_allclose(datasets._stat_sum_likelihood(), 12825.9370, rtol=1e-3)
     assert_allclose(datasets.stat_sum(), 12825.9370, rtol=1e-3)
```
bkhelifi, line 760: as `(datasets._stat_sum_likelihood()` is equal to `datasets.stat_sum()`, why is this test function really testing the prior stat_sum? 
Should we not use a prior that lead to a non-nul prior_stat_sum?

QRemy: Here it is the same because it tests a uniform prior with weight=1,
at the end of the test there is also a check with weight=100 that actually make a difference.

adonath:
🎉

PR #5663, opened 2025-01-08
Support weighted likelihood in MapDataset

QRemy:
Support weighted likelihood in MapDataset allowing mask to be a weight map (as done in Fermi-LAT analyses)

### DIFF
--- a/gammapy/datasets/core.py
+++ b/gammapy/datasets/core.py
@@ -61,7 +61,7 @@ class Dataset(abc.ABC):
     def mask(self):
         """Combined fit and safe mask."""
         if self.mask_safe is not None and self.mask_fit is not None:
-            return self.mask_safe & self.mask_fit
+            return self.mask_safe * self.mask_fit
         elif self.mask_fit is not None:
             return self.mask_fit
         elif self.mask_safe is not None:
--- a/gammapy/datasets/core.py
+++ b/gammapy/datasets/core.py
@@ -79,7 +79,7 @@ class Dataset(abc.ABC):
         stat = self.stat_array()
 
         if self.mask is not None:
-            stat = stat[self.mask.data]
+            stat = stat[self.mask.data.astype(bool)]
         return np.sum(stat, dtype=np.float64)
 
     @abc.abstractmethod
--- a/gammapy/datasets/core.py
+++ b/gammapy/datasets/core.py
@@ -79,7 +79,10 @@ class Dataset(abc.ABC):
         stat = self.stat_array()
 
         if self.mask is not None:
-            stat = stat[self.mask.data.astype(bool)]
+            if isinstance(self.mask, np.ndarray):
+                stat = stat[self.mask.astype(bool)]
+            else:
+                stat = stat[self.mask.data.astype(bool)]
         return np.sum(stat, dtype=np.float64)
 
     @abc.abstractmethod
--- a/gammapy/datasets/map.py
+++ b/gammapy/datasets/map.py
@@ -1474,7 +1474,7 @@ class MapDataset(Dataset):
         counts, npred = self.counts.data.astype(float), self.npred().data
 
         if self.mask is not None:
-            mask = self.mask.data == True  # noqa
+            mask = ~(self.mask.data == False)
             counts = counts[mask]
             npred = npred[mask]
             if self.mask.data.dtype == bool:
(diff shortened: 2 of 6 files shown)


facero:
Thanks, this is indeed an important point for the Galactic analysis !

For future reference in the docs, one possible reference for this method is in 3FGL paper 2015ApJS..218...23A (Sect 3.1.2. Galactic diffuse normalization and unweighting) Although they don't go into much details...
Or this technical document ([link](https://indico.cern.ch/event/783425/contributions/3335026/attachments/1814171/2966040/binned_like.pdf )) in Sect 2.7.

Also in the docs we should to a Fermi page with an example on how to generate the weight map but so fare I couldn't find a good link

gammapy/datasets/map.py
```diff
@@ -1467,12 +1468,17 @@ def stat_sum(self):
         counts, npred = self.counts.data.astype(float), self.npred().data
 
         if self.mask is not None:
-            return (
-                cash_sum_cython(counts[self.mask.data], npred[self.mask.data])
-                + prior_stat_sum
-            )
+            mask = self.mask.data == True  # noqa
```
facero, line 1471: Not sure I fully understand what this line here does. Does it converts values to boolean ?

AtreyeeS:
Would it be better to have a separate `weights` argument directly?

QRemy:
> Would it be better to have a separate `weights` option explicitly?

We would need to have one more map in memory and in the fits files this way it can be avoided. 
We could even consider renaming mask/mask_fit/mask_safe... to weight/weight_fit/weight_safe...

AtreyeeS:
> We would need to have one more map in memory and in the fits files, but it can be avoided this way. We could even consider renaming mask/mask_fit/mask_safe... to weight/weight_fit/weight_safe...

I dont see the advantage of renaming `mask_safe`. It still remains a boolean mask, no?
Yes, maybe renaming `mask_fit` to `weights_fit`  can be better

gammapy/datasets/map.py
QRemy, line 1471:
```suggestion
            mask = ~(self.mask.data == False)
```

AtreyeeS:
Having a `use_mask_as_weigths` option as discussed would be useful indeed, as doing a `interp_to_geom`, eg as done in the FoVBackgroundMaker 

https://github.com/gammapy/gammapy/blob/bc69342c8de7ccfa4f53093619f93fa8eb06ce9f/gammapy/makers/background/fov.py#L133

leads to non integer values in certain bins.

QRemy:
Now it applies the weights only if the `stat_type="cash_weighted"` and I added a child class `MapDatasetWeighted` with this set as default.

gammapy/datasets/core.py
```diff
@@ -79,7 +79,10 @@ def _stat_sum_likelihood(self):
         stat = self.stat_array()
 
         if self.mask is not None:
-            stat = stat[self.mask.data]
+            if isinstance(self.mask, np.ndarray):
```
AtreyeeS, line 82: Maybe it is better to have the mask as a Map always. I remember there was some confusion when we shifted from ndarray to map

AtreyeeS (approved):
Thanks @QRemy ! This is really nice.
I dont know how things will evolve with the new FitStatistics API later.

Should #5649 create a `MapDatasetWeighted` ?

facero:
> > Should #5649 create a `MapDatasetWeighted` ?
> 
> I will do it in a separate PR afterward. For now all weight maps I created are filled with ones so I will have to discuss with @facero.

Would you need a test dataset that has a valid weight map for your tests ? I think I should have one at hand

AtreyeeS:
I tried this out, and I think there is a need for both `mask_fit` and `weights`
I used some precomputed `weights`, and also wanted to mask out some badly modelled pixels, so I had to set 
```
weights = dataset.mask_fit
dataset.mask_fit = weights*my_mask_fit
``` 
and that made the book keeping rather ugly.

Maybe setting weights as the `mask_safe` can be an option? The safe mask is also something we compute beforehand, and not keep changing during the analysis.

QRemy:
> Maybe setting weights as the `mask_safe` can be an option? The safe mask is also something we compute beforehand, and not keep changing during the analysis.

The current implementation already support that. I added a test to check it explicitly.

QRemy:
> Would you need a test dataset that has a valid weight map for your tests ? I think I should have one at hand

Not for this PR, but later for the tutorial it could be nice to show the effect of the weighted likelihood on the fitted parameters.

registerrier (approved):
Thanks @QRemy! This looks good to me. 

This will be adapted once we have a FitStatistic class. The question of how to serialize properly the the type of statistic to use can be solved later.

PR #5667, opened 2025-01-10
Refactor `min` and `max` setter on `Parameter` and `set_lim` method

MRegeard:
This PR refactor the setter of `min` and `max` setter of `Parameter` to support assignment with `Quantity` and `str`. 

It also add a method to set both `min` and `max` value at the same time.


gammapy/modeling/parameter.py
```diff
@@ -138,6 +136,8 @@ def __init__(
             self.value = float(value)
             self.unit = unit
 
+        self.min = min
+        self.max = max
```
AtreyeeS, line 140: To avoid having to check between `str` and `Quantity` in every place, you can convert the str to a quantity here

```
if isinstance(min, str):
      self.min = u.Quantity(min)
```

gammapy/modeling/parameter.py
```diff
@@ -264,10 +258,42 @@ def max(self):
...
+        Set the min and/or max value for the parameter.
+
+        Parameters
+        ----------
+        min, max: float, `~astropy.units.Quantity`, str or tuple of float/`~astropy.units.Quantity`/str
+            Minimum and Maximum value to assign to the parameter `min` and `max`. If a tuple is
+            passed and `max` is set to None, the tuple will be decomposed into min and max value.
+            Default is None, which set `min` and `max` to `np.nan`.
+        """
+        if max is None and np.iterable(min):
```
AtreyeeS, line 285: Here you should raise an error if min is a tuple and max is also present. Otherwise it will fail with a confusing message later on

AtreyeeS (commented):
Thanks @MRegeard ! I think this needs to be tested better

gammapy/modeling/parameter.py
MRegeard: I understand your argument but at the same time I think it's cleaner to put this in the setter. Moreover, one most probably set min and max after the `Parameter` initialisation.

AtreyeeS: i agree. Then you can just simplify the setter by changing the str to a qty

gammapy/modeling/parameter.py
MRegeard: Agree !
By the way, I did this because this method is largely inspired by `plt.x/ylim`, `Axes.set_x/ylim`. 
In a way I like that you can have tuples for min and max values and directly pass them to this method. But it makes the API a bit weird, so if people have strong opinion against it, we could remove this possibility. 
Otherwise, we could also show the call signature in the docstring, as it is done in the [x/ylim documentation](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.xlim.html).

gammapy/modeling/parameter.py
MRegeard: Done !

gammapy/modeling/tests/test_parameter.py
```diff
@@ -369,3 +375,33 @@ def test_priorparameters_to_table(priorpars):
...
+    assert par.min == 1
+
+    par.set_lim((2, 4))
+    assert par.min == 2
+    assert par.max == 4
+
+    assert par.error == 0
+
+
+def test_set_quantity_str_float(default_parameter):
```
AtreyeeS, line 400: Can you add a check on the `par.error`? 
What happens if a user does

gammapy/modeling/parameter.py
AtreyeeS: I agree with you about the convenience, and also about the API being a bit weird to have an iterable for min.
Maybe the user can either set `lims = ()` or min and max separately?

gammapy/modeling/tests/test_parameter.py
```diff
@@ -369,3 +375,33 @@ def test_priorparameters_to_table(priorpars):
...
+    par = default_parameter
+    assert par._set_quantity_str_float(1) == 1
```
AtreyeeS, line 402: You can test on `par.set_lim` directly such that the hidden functions are executed correctly internally

gammapy/modeling/tests/test_parameter.py
```diff
@@ -369,3 +375,33 @@ def test_priorparameters_to_table(priorpars):
...
+    assert par._set_quantity_str_float("3 TeV") == 3
+    assert par._set_quantity_str_float(2 * u.TeV) == 2
+    assert par._set_quantity_str_float(4e3 * u.GeV) == 4
```
AtreyeeS, line 405: Add asserts on `par.error` when units are passed.

What happens when one of the parameters has units, the other not, eg `par.set_lim(min=1, max=1*u.GeV)`?
Is the error message reasonable?

AtreyeeS (commented):
Thanks @MRegeard !
I suppose the `min` and `max` setter should be deprecated and users should only interact with `set_lim`?

gammapy/modeling/tests/test_parameter.py
MRegeard: It does not raise any error. Should it ? It is easy to implement if we want it.

MRegeard:
@AtreyeeS, I think that we should keep min and max setter. It is always useful. For instance you can use in-place operators like `+=` or `*=` with setter.
Moreover, all the logic has been implemented on the setter. So in the end `set_lim` is so sort of syntactic sugar for `.min =` and `.max =`

AtreyeeS (approved):
Thanks! No comments

registerrier (approved):

PR #5668, opened 2025-01-14
PIG 29: A generalised sky model

### LINKED ISSUE #5377: A generalised 1D model
During the last coding sprint, @cgalelli and I tried to implement a generalised 1D model to 

1. Avoid code duplication between temporal and spectral models
2. Enable an easy API for fitting other parameters, eg, phase, frequency, etc

I have an example notebook here: https://github.com/AtreyeeS/test_notebooks/blob/master/general_model.ipynb
(there is some ugly jugglery in the integral function, but there should be more elegant solutions with bumpy)

I think this works, and we might consider implementing it for 2.0

AtreyeeS:
This PIG proposes to refactor the models API to allow fitting on arbitrary parameters.
An example notebook can be found here: https://github.com/AtreyeeS/test_notebooks/blob/master/general_model.ipynb

The changes to the MapEvaluator are more detailed

### DIFF
--- a/docs/development/pigs/pig-029.rst
+++ b/docs/development/pigs/pig-029.rst
@@ -1,0 +1,207 @@
+********************************
+PIG 29 - SkyModel Restructuring
+********************************
+
+* Author: Atreyee Sinha, Maxime Regard
+* Created: ``2024-12-13``
+* Accepted: ``xx``
+* Status: xx
+* Discussion: `#4381`_
+
+Abstract
+========
+
+This PIG intends to resturcture `gammapy.modelling.models` to allow for flexible model definitions on arbitrary variables.
+
+Motivation
+==========
+
+Currently, gammapy supports Temporal, Spectral and Spatial models, which are bundled together in the `SkyModel` class for modelling and fitting. Despite the underlying `maps` framework being completely flexible, the modelling framework follows a fixed API allowing only spectral (compulsory), temporal and spatial models. If a user wants to model the behaviour in other parameters, say phase or frequency, it requires expert knowledge and is often performed using other fitting packages, eg: see https://github.com/cgalelli/gammapy_SyLC/tree/main/gammapy_SyLC.
+
+Currently, the `MapDataset` and the `FluxPointsDataset` support additional axes, so the main changes are required in the `MapEvaluator` and in the model definitions
+
+
+Use cases
+=========
+
+1. Fitting other parameters like phase, frequency, etc
+2. Reduction of code duplication: Analytic shapes like the power law is defined separately for `PowerLawSpectralModel`, `PowerLawTeLmporalModel` and `PowerLawNormSpectralModel`, which differ only in the unit handling
+3. Separation of units and values: This is a side benefit which can be useful if move to using JAX arrays in the future.
+
+
+Proposed API
+============
+
+We propose to have a general unit-less analytic class definition, inheriting from `ModelBase` The unit handling responsibility lies on the user.
+Internal support and checks and be provided for common input parameters (eg: time/energy) and the expected output units. 
+
+So, we move from 
+
+`SkyModel(spectral_model=, temporal_model=)`
+to 
+
+`SkyModelGeneral(submodels={"energy_true": p, "time":t})`
+
+where the user defines the axes the models are supposed affect.
+
+
+For a simple power law, it will be:
+
+"$\\phi(x) = \\phi_0 \\cdot \\left( \\frac{x-x_{ref}}{x_0} \\right)^{-\\alpha}$\n",
+
+
+```
+def integrate_linear(func, range_min, range_max, ndecade=100, oversampling_factor=100, **kwargs):
+        if isinstance(range_min, Time):
+            range_min=range_min.mjd * u.d
+            range_max=range_max.mjd * u.d
+        x, steps = np.linspace(
+            range_min, range_max, oversampling_factor, retstep=True, axis=-1
+        )
+        values = func(x)
+        integral = np.sum(values, axis=-1) * steps
+        return integral/np.sum(range_max - range_min) 
+
+class PowerLawModel(ModelBase):
+
+    tag = "Generalised_po"
+    alpha = Parameter("alpha", 2, frozen=True)
+    x0 = Parameter("x0", 1, frozen=True)
+    x_ref = Parameter("x_ref", 1, frozen=True)
+    phi0 = Parameter("phi0", 1, frozen=True, scale_method="scale10",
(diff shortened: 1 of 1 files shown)


docs/development/pigs/pig-029.rst
```diff
@@ -0,0 +1,207 @@
...
+        kwargs = {par.name: par.quantity for par in self.parameters}
+        if isinstance(x, Time):
+            x=Time(x, scale=x.scale).mjd * u.d
+        x = x.to(self.axis_units)
+        return self.evaluate(x, **kwargs)
+    
+    @staticmethod
+    def evaluate(x, x_ref, x0, alpha, phi0):
+        """Evaluate the model (static function)"""
+        return phi0*np.power((x-x_ref)/x0, -alpha)
```
QRemy, line 93: did we discuss at some point to have evaluate the unit separately ?

docs/development/pigs/pig-029.rst
```diff
@@ -0,0 +1,207 @@
...
+class PowerLawModel(ModelBase):
+
+    tag = "Generalised_po"
+    alpha = Parameter("alpha", 2, frozen=True)
+    x0 = Parameter("x0", 1, frozen=True)
+    x_ref = Parameter("x_ref", 1, frozen=True)
+    phi0 = Parameter("phi0", 1, frozen=True, scale_method="scale10",
+                    interp="log")
+    
+    def __init__(self, input_units="", norm_units="", **kwargs):
```
QRemy, line 74: why do you need input_units="", norm_units="" ?
they can be determined from the unit of the parameters given in input

docs/development/pigs/pig-029.rst
```diff
@@ -0,0 +1,207 @@
...
+
+We propose to have a general unit-less analytic class definition, inheriting from `ModelBase` The unit handling responsibility lies on the user.
+Internal support and checks and be provided for common input parameters (eg: time/energy) and the expected output units. 
+
+So, we move from 
+
+`SkyModel(spectral_model=, temporal_model=)`
+to 
+
+`SkyModelGeneral(submodels={"energy_true": p, "time":t})`
```
QRemy, line 43: what happens to the spatial model ?

docs/development/pigs/pig-029.rst
AtreyeeS: Indeed we did!
So should we work like `np.power((x.value-x_ref.value)/x0.value, -alpha)` or should the parameters be devoid of units and only attached in the end?
The latter would required modifications in the `Parameter` class

docs/development/pigs/pig-029.rst
AtreyeeS: Yes! But I thought its safer to set the units on init, rather than during the call so that things dont change silently.
Otherwise, you would suggest to set it during the first call to the function?

docs/development/pigs/pig-029.rst
AtreyeeS: What we do for the 2D models is a question left open. One can say, eg, ` submodels={spatial: GaussianSpatialModel}` if we dont want to change the spatial models.

2D general models maybe useful, eg in the camera plane instead of on the sky. But I dont know how practically useful it can be

docs/development/pigs/pig-029.rst
```diff
@@ -0,0 +1,207 @@
...
+* Author: Atreyee Sinha, Maxime Regard
+* Created: ``2024-12-13``
+* Accepted: ``xx``
+* Status: xx
+* Discussion: `#4381`_
+
+Abstract
+========
+
+This PIG intends to resturcture `gammapy.modelling.models` to allow for flexible model definitions on arbitrary variables.
```
VascoSch92, line 14:
```suggestion
This PIG intends to restructure `gammapy.modelling.models` to allow for flexible model definitions on arbitrary variables.
```

docs/development/pigs/pig-029.rst
```diff
@@ -0,0 +1,207 @@
...
+
+Motivation
+==========
+
+Currently, gammapy supports Temporal, Spectral and Spatial models, which are bundled together in the `SkyModel` class for modelling and fitting. Despite the underlying `maps` framework being completely flexible, the modelling framework follows a fixed API allowing only spectral (compulsory), temporal and spatial models. If a user wants to model the behaviour in other parameters, say phase or frequency, it requires expert knowledge and is often performed using other fitting packages, eg: see https://github.com/cgalelli/gammapy_SyLC/tree/main/gammapy_SyLC.
+
+Currently, the `MapDataset` and the `FluxPointsDataset` support additional axes, so the main changes are required in the `MapEvaluator` and in the model definitions
```
VascoSch92, line 21:
```suggestion
Currently, the `MapDataset` and the `FluxPointsDataset` support additional axes, so the main changes are required in the `MapEvaluator` and in the model definitions.
```

docs/development/pigs/pig-029.rst
```diff
@@ -0,0 +1,207 @@
...
+
+
+Use cases
+=========
+
+1. Fitting other parameters like phase, frequency, etc
```
VascoSch92, line 27:
```suggestion
1. Fitting other parameters, e.g., phase, frequency, ...
```

docs/development/pigs/pig-029.rst
```diff
@@ -0,0 +1,207 @@
...
+2. Reduction of code duplication: Analytic shapes like the power law is defined separately for `PowerLawSpectralModel`, `PowerLawTeLmporalModel` and `PowerLawNormSpectralModel`, which differ only in the unit handling
+3. Separation of units and values: This is a side benefit which can be useful if move to using JAX arrays in the future.
+
+
+Proposed API
+============
+
+We propose to have a general unit-less analytic class definition, inheriting from `ModelBase` The unit handling responsibility lies on the user.
```
VascoSch92, line 35:
```suggestion
We propose to have a general unit-less analytic class definition, inheriting from `ModelBase`. The unit handling responsibility lies on the user.
```

docs/development/pigs/pig-029.rst
```diff
@@ -0,0 +1,207 @@
...
+Internal support and checks and be provided for common input parameters (eg: time/energy) and the expected output units. 
+
+So, we move from 
+
+`SkyModel(spectral_model=, temporal_model=)`
+to
```
VascoSch92, line 41:
```suggestion
`SkyModel(spectral_model=..., temporal_model=...)`

to 
```

registerrier (commented):
Here we propose to withdraw the PIG for now.

PR #5675, opened 2025-01-22
Clarifications in Fermi-LAT notebook

AtreyeeS:
I was doing some Fermi Analysis and had my exposure map on a larger region than the counts map, which is the recommendation on the FSSC webpage but Gammapy does not support. I had a rather confusing message during the npred computation, so maybe it is good to warn users about this in advance?

### DIFF
--- a/examples/tutorials/data/fermi_lat.py
+++ b/examples/tutorials/data/fermi_lat.py
@@ -12,29 +12,12 @@ Gammapy. As an example, we will look at the Galactic center region using
 the high-energy dataset that was used for the 3FHL catalog, in the
 energy range 10 GeV to 2 TeV.
 
-We note that support for Fermi-LAT data analysis in Gammapy is very
-limited. For most tasks, we recommend you use
+We note that support for Fermi-LAT data analysis in Gammapy is currently
+limited. For more details on Fermi-LAT analysis, please see
 `Fermipy <http://fermipy.readthedocs.io/>`__, which is based on the
 `Fermi Science
 Tools <https://fermi.gsfc.nasa.gov/ssc/data/analysis/software/>`__
 (Fermi ST).
-
-Using Gammapy with Fermi-LAT data could be an option for you if you want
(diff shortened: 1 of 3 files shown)


QRemy:
> I had a rather confusing message during the npred computation

What was the message ?

QRemy:
I think the underlying problem is that the `geom` used as starting point for `npred` computation is the exposure `geom` and it must be aligned with the ones of counts and background otherwise we can't compute `npred` and `stat_sum`.
Maybe we can deal with that in the MapEvaluator :
```
if geom.to_image() != gem_reco.to_image():
    geom = gem_reco.to_image().to_cube(geom.axes) 
```

Regarding the recommendation to use larger exposure map for Fermi-LAT the approach in gammapy would rather be to use the same map but rely on the mask_fit to define a smaller ROI in reco energy as shown here: https://docs.gammapy.org/dev/tutorials/api/mask_maps.html#mask-some-sky-regions

AtreyeeS:
> I think the underlying problem is that the `geom` used as starting point for `npred` computation is the exposure `geom` and it must be aligned with the ones of counts and background otherwise we can't compute `npred` and `stat_sum`. Maybe we can deal with that in the MapEvaluator :
> 
> ```
> if geom.to_image() != gem_reco.to_image():
>     geom = gem_reco.to_image().to_cube(geom.axes) 
> ```

Maybe we can do an `interp_to_geom` internally. I agree this is better than raising a warning

> Regarding the recommendation to use larger exposure map for Fermi-LAT the approach in gammapy would rather be to use the same map but rely on the mask_fit to define a smaller ROI in reco energy as shown here: https://docs.gammapy.org/dev/tutorials/api/mask_maps.html#mask-some-sky-regions

Yes, I remembered it when the `npred` failed. So I imagine some others might have the same issue as well... I can add a note about this in the Fermi-LAT notebook if you think so

AtreyeeS:
Closing this for now, Will wait for open PRs on Fermi analysis to be merged first

PR #5676, opened 2025-01-23
Cleanup `MapDataset.__init__`

MRegeard:
This is a small PR that put the meta logic in the `__init__` function of `MapDataset` in the setter of meta. 
Plus some corrections in the error message of psf and edisp

### DIFF
--- a/gammapy/datasets/map.py
+++ b/gammapy/datasets/map.py
@@ -540,7 +540,7 @@ class MapDataset(Dataset):
 
         if psf and not isinstance(psf, (PSFMap, HDULocation)):
             raise ValueError(
-                f"'psf' must be a `PSFMap` or `HDULocation` object, got `{type(psf)}` instead."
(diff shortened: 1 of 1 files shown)


QRemy (commented):
Thanks @MRegeard, in most of the other error message we don't use "`" but nothing around the f-sting like {type(psf)}.

gammapy/datasets/map.py
```diff
@@ -540,13 +540,13 @@ def __init__(
 
         if psf and not isinstance(psf, (PSFMap, HDULocation)):
             raise ValueError(
-                f"'psf' must be a 'PSFMap' or `HDULocation` object, got {type(psf)}"
+                f"'psf' must be a `PSFMap` or `HDULocation` object, got `{type(psf)}` instead."
```
QRemy, line 543:
```suggestion
                f"'psf' must be a `PSFMap` or `HDULocation` object, got {type(psf)} instead."
```

PR #5677, opened 2025-01-23
1st implementation of the `EventDataset`

MRegeard:
This PR implements the first brick of code for the Unbinned analysis: the `EventDataset`. 

At the moment it is missing a lot, but I think it is good to start somewhere. It will also probably bring some discussion on the implementation of the different part needed. 

I used the `MapDataset` and the `EventDataset` from https://github.com/gammapy/gammapy-unbinned-analysis/blob/main/EventDataset/EventDataset.py#L209 for this implementation. 

Please fill free to push to this branch to help in the process !


gammapy/datasets/events.py
```diff
@@ -0,0 +1,128 @@
...
+        """Models setter."""
+        self._evaluators = {}
+        if models is not None:
+            models = DatasetModels(models)
+            models = models.select(datasets_names=self.name)
+        self._models = models
+
+    @property
+    def events(self):
+        return self._events.select_row_subset(self.event_mask)
```
bkhelifi, line 106: One should discuss about the logics of event selection in order to understand whether this implementation makes sense

QRemy:
Thanks @MRegeard , closing this prototype until activity on unbinned analyses restart.

PR #5678, opened 2025-01-23
Visualisation function to plot stat profile

MRegeard:
This PR adds a function to plot the stat profile of a fit.
The code comes mostly from the [fitting tutorial](https://docs.gammapy.org/dev/tutorials/api/fitting.html).


gammapy/visualization/utils.py
```diff
@@ -410,3 +410,47 @@ def func_to_fit(x, mu, sigma):
...
+    width = 12
+    figsize = (width, width * rows / cols)
+
+    fig, axes = plt.subplots(
+        nrows=rows,
+        ncols=cols,
+        figsize=figsize,
+    )
+
+    for ax, par in zip(axes.flatten(), free_parameters):
```
bkhelifi, line 446: maybe one should add here a parallelization, as well as a progress bar

gammapy/visualization/utils.py
```diff
@@ -410,3 +410,47 @@ def func_to_fit(x, mu, sigma):
...
+        par.scan_n_values = 17
```
bkhelifi, line 447: why not having this value as a kwargs?

gammapy/visualization/utils.py
```diff
@@ -410,3 +410,47 @@ def func_to_fit(x, mu, sigma):
...
+        idx = model.parameters.index(par)
+        name = model.parameters_unique_names[idx]
+        profile = fit.stat_profile(datasets=datasets, parameter=par)
+        ax.plot(profile[f"{model.name}.{name}_scan"], profile["stat_scan"] - total_stat)
+        ax.set_xlabel(f"{par.name} [{par.unit}]")
+        ax.set_ylabel("Delta TS")
+        ax.set_title(f"{name}:\n {par.value:.1e} +- {par.error:.1e}")
+
```
bkhelifi, line 455: depending on the dof, should we have horizontal lines corresponding to 1,2 and 3 sigma?

AtreyeeS (commented):
Thanks @MRegeard 
Having all the profiles together is nice, but I dont know if that is really the best thing to expose.
For me, it is better to expose the use of `fit.stat_profile`
Maybe the visualisation function can take as input a `fit.stat_profile` and generate the plot.
This plot is very simple though..

An inbuilt visualization for `stat_contour` or `stat_surface` is for sure useful.
And there, it is probably a rather bad idea to try to generate all possible contours in one go

MRegeard:
@AtreyeeS, I see your point and I agree with you. 
I primary idea was to provide a function that reproduce the code of the fitting tutorial. Because if you want to check the quality of a fit, looking at the profile is a good idea. This means that the user has to "carry around" this specific cell that we provide in each analysis. In the end, they probably will put this into a function anyway. 
I thought it might be a good addition to our visualisation module.

registerrier (commented):
Thanks @MRegeard . This is an interesting proposal. I have a few general comments and possible suggestions:
- in general the responsibility of a function should be limited. I think computing the profiles and plotting them are different actions.
- maybe having a FitStatisticProfile object might useful. One could then:
  - plot it with `profile.plot()` for this use case
  - interpolate on it to evaluate the approximated fit stat value for a given parameter value (see for instance the profile fit statistic now used for `FluxPointsDataset`)
- The `Fit.stat_profile` could possibly work with a collection of parameters.

Astro-Kirsty:
Closing as this will be handled by the plotter class

PR #5679, opened 2025-01-23
Add optional sensitivity computation on TSMapEstimator

QRemy:
Add optional sensitivity computation on TSMapEstimator.
Requires #5638.


gammapy/estimators/map/ts.py
```diff
@@ -1042,3 +1076,38 @@ def _ts_value(
         norm_guess=norm_guess,
     )
     return flux_estimator.run(dataset)
+
+
+class BrentqSensitivityEstimator(BrentqFluxEstimator):
+    """Single parameter sensitivity estimator."""
+
+    def _confidence(self, dataset, n_sigma, result, positive):
```
adonath, line 1084: As the method is hidden away, just move it to the `BrentqFluxEstimator`? I think the motivation for a  new "BrentqSensitivityEstimator" is weak.

adonath (dismissed):
Thanks @QRemy, looks good to me! I just have one minor comment...

gammapy/estimators/map/ts.py
QRemy: yes I removed it.

adonath:
Thanks @QRemy, looks good! I think you can remove the draft label...

gammapy/estimators/map/ts.py
```diff
@@ -75,6 +76,8 @@ class TSMapEstimator(Estimator, parallel.ParallelMixin):
         Number of sigma for flux error. Default is 1.
     n_sigma_ul : int
         Number of sigma for flux upper limits. Default is 2.
+    n_sigma_sensitivity : int
+        Number of sigma for flux  sensitivity. Default is 5.
```
bkhelifi, line 80: Is it the integral sensitivity above the energy threshold given by the safe_mask or the fit_mask? This energy threshold is a key parameter to give to the user. Is it the case?

bkhelifi: Argh, damned. I have not checked that... 
In any case, should we implement an utility plot function that give the exact energy threshold? Whatever the origin of the mask, the value should be quoted. If you think that this is necessary, could you make a dedicated feature-request issue?

registerrier (approved):
Thanks @QRemy. This looks nice and well structured.

Maybe the multiplicative factor might be configurable as in some cases (bright source) you might have a sensitivity way lower than this no? That might be kept for later though if even needed

PR #5683, opened 2025-01-27
Unify references

### LINKED ISSUE #5251: Consistency with 'references' in docstrings

Astro-Kirsty:
This resolves #5251 

In places such as [here](https://docs.gammapy.org/dev/api/gammapy.estimators.TSMapEstimator.html)
We have the reference which links to the reference page, while this is nice while within the documentation, it is not from the docstring. i.e. if you are coding jupyter and query the class you cannot see the reference properly. 

My preference would be that for the docstring we are explicit in the referencing and leave the `References` list in the `User Guide` for references which are within the documentation descriptions themselves.

### DIFF
--- a/gammapy/catalog/lhaaso.py
+++ b/gammapy/catalog/lhaaso.py
@@ -155,8 +155,8 @@ class SourceCatalog1LHAASO(SourceCatalog):
 
     References
     ----------
-    .. [1] 1LHAASO: The First LHAASO Catalog of Gamma-Ray Sources,
-       https://ui.adsabs.harvard.edu/abs/2023arXiv230517030C/abstract
+    .. [1] `Cao et al. (2023), "1LHAASO: The First LHAASO Catalog of Gamma-Ray Sources"
+       <https://ui.adsabs.harvard.edu/abs/2023arXiv230517030C/abstract>`_
     """
 
     tag = "1LHAASO"
--- a/gammapy/estimators/map/ts.py
+++ b/gammapy/estimators/map/ts.py
@@ -166,7 +166,8 @@ class TSMapEstimator(Estimator, parallel.ParallelMixin):
 
     References
     ----------
-    [Stewart2009]_
+    .. [Stewart2009]_ `Stewart (2009), “Maximum-likelihood detection of sources among Poissonian noise”
+       <https://ui.adsabs.harvard.edu/abs/2009A%26A...495..989S/abstract>`_
     """
 
     tag = "TSMapEstimator"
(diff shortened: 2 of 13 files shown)


gammapy/estimators/map/ts.py
```diff
@@ -166,7 +166,8 @@ class TSMapEstimator(Estimator, parallel.ParallelMixin):
 
     References
     ----------
-    [Stewart2009]_
+    .. [Stewart2009]_ `Stewart (2009), “Maximum-likelihood detection of sources among Poissonian noise”
```
bkhelifi, line 169: Why do you keep the "[Stewart2009]_" here, where in other python script you have removed the "[REF]" ?

gammapy/catalog/lhaaso.py
```diff
@@ -155,8 +155,8 @@ class SourceCatalog1LHAASO(SourceCatalog):
 
     References
     ----------
-    .. [1] 1LHAASO: The First LHAASO Catalog of Gamma-Ray Sources,
-       https://ui.adsabs.harvard.edu/abs/2023arXiv230517030C/abstract
+    .. [1] `Cao et al. (2023), "1LHAASO: The First LHAASO Catalog of Gamma-Ray Sources"
```
bkhelifi, line 158: Should we keep the "[1]", when in other scripts it does not appear?

bkhelifi (approved):
Thank you a lot for the effort!

registerrier (approved):

Issue #5685, opened 2025-01-29
TemporalModel event sampling produces spikes of events at edges of sampling range


registerrier:
**Gammapy version**
1.3 and dev

**Bug description**
When sampling a `TemporalModel` with large number of events, one can observe that several events have the same time. They are located at both ends of the interval defined for sampling.

This is caused when the sampler which distributes over a +-0.5 around the bin index. The template must be computed at centers rather than bin edges and  a shift must be applied to recover the correct distribution.

**Expected behavior**
Events must be distributed uniformly in the interval and events with exactly the same time should be be rare.

**To Reproduce**
```
import matplotlib.pyplot as plt
import numpy as np
from gammapy.modeling.models import ConstantTemporalModel
from astropy.time import Time
from astropy.units import Quantity

cst = ConstantTemporalModel()
t0 = Time.now()
t1 = t0 + Quantity("100 s")
times = cst.sample_time(50000, t0, t1, t_delta="0.5 s").sort()

n_same_time = np.sum((times[1:]-times[:-1]).to("s")==0)
print(n_same_time)

_ = plt.hist((times-t0).to("s"), bins=500)
```
yields 234 events with the same time and the following histogram:

![Image](https://github.com/user-attachments/assets/2a3c90de-9ee6-4e0f-926a-4207697ac0a1)

**Other information**
Any other information you think will be useful for us to fix the issue can go here.

PR #5686, opened 2025-01-29
Correct time sampling issue #5685

registerrier:
**Description**

This pull request solves #5685 . The issue was that the time interval was split into n pixels and the model evaluated at the edges of these pixels to build the `InverseCDFSampler`. The latter therefore returned sample pixel coordinates between -0.5  and n+0.5. And all pixel coordinates below 0 and beyond n where put to 0 and n.

This solves this by sampling of pixels centers and shifting by half a pixel the sampled pixel coordinates.

**Dear reviewer**

### DIFF
--- a/gammapy/modeling/models/tests/test_temporal.py
+++ b/gammapy/modeling/models/tests/test_temporal.py
@@ -162,6 +162,17 @@ def test_time_sampling_template():
     assert_allclose(std - sigma.to("d").value, 0.0, atol=3e-4)
 
 
+def test_time_sampling_uniform():
+    cst = ConstantTemporalModel()
+    t0 = Time.now()
+    t1 = t0 + u.Quantity("100 s")
+    times = cst.sample_time(10000, t0, t1, t_delta="0.5 s").sort()
+    time_diff = (times[1:] - times[:-1]).to_value("s")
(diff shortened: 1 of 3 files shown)


fabiopintore:
Thanks @registerrier . It looks very good to me. 
No further comments from my side.

gammapy/modeling/models/tests/test_temporal.py
```diff
@@ -162,6 +162,17 @@ def test_time_sampling_template():
     assert_allclose(std - sigma.to("d").value, 0.0, atol=3e-4)
 
 
+def test_time_sampling_uniform():
```
VascoSch92, line 165: I will add a comment linking the issue. So the next person that come here know why you choose that particular test case :-)

registerrier:
> Thanks @registerrier! In addition one could think about just overwriting `ConstantTemporalModel.sample_time()` using `np.random.uniform`. It is a bit more code, but I think in other cases (exponential etc.) Numpy has already figured out the numerically stable algorithms for sampling.

Indeed! In fact, we could rely on most `scipy.stats` classes and their `rvs()` method to generate random variables. Even for the sampler one can create a `rv_histogram` from a LC for instance.

registerrier:
Merging now.

PR #5687, opened 2025-01-30
[PROTOTYPE] Potential modifications to support a future format 

QRemy:
Few ideas on the modifications required in the HDU-Index and OBS-Index table for a future data format and the changes needed in gammapy.

### Format modifications
1. **Independent data partition**:
   - HDU-Index and OBS-Index should contain a single `event_class` with potentially multiple `event_type` or `observation_block`.
   - Each combination of `event_type` and `observation_block` (or any future independent data partition) should be associated with a fixed `irf_name`.
   - A given `irf_name` might not be associated with a well-defined IRF; in such cases, the `FILE_NAME` would be empty.

2. **HDU_NAME Differentiation**:
   - If independent data partitions are stored in the same file, the `HDU_NAME` should be different, maybe combining `HDU_NAME` and `irf_name` to differentiate them.

3. **Identifiers in OBS-Table**:
   - Any identifier used to select independent data partitions, such as `OBS_ID`, `EVENT_TYPE`, `OBS_BLOCK_ID`, and `IRF_NAME`, should only appear in the `obs-index` and not in the `hdu-index`.

4. **Unique Identifier for Independent data partition**:
   - Introduce a new unique identifier on both `obs-index` and `hdu-index` to link each independent data entry in the `obs-table` to the corresponding files in the `hdu-table`. This could be `OBS_TABLE_ROW`.

### Code Modifications in Gammapy
The first changes needed to support the format modification previously described are available in the pull request. Follow-up PRs could include : 

1. **Datastore Method Changes**:
   - Remove the `obs_id` argument from `datastore.get_observations`.
   - Implement all selections using a new method `datastore.select`.

2. **Stacking Entries**:
   - `datastore.get_observations` should stack entries with the same `IRF_name`.
   - Allow `obs.events` to be a list of event lists to avoid having everything in memory and to prevent repeating data reduction of the IRFs.

### DIFF
--- a/gammapy/data/data_store.py
+++ b/gammapy/data/data_store.py
@@ -94,7 +94,10 @@ class DataStore:
     @property
     def obs_ids(self):
         """Return the sorted obs_ids contained in the datastore."""
-        return np.unique(self.hdu_table["OBS_ID"].data)
+        if "OBS_ID" in self.hdu_table.keys():
+            return np.unique(self.hdu_table["OBS_ID"].data)
+        else:
+            return np.unique(self.obs_table["OBS_ID"].data)
 
     @classmethod
     def from_file(cls, filename, hdu_hdu="HDU_INDEX", hdu_obs="OBS_INDEX"):
@@ -260,7 +263,13 @@ class DataStore:
         else:
             return s
 
-    def obs(self, obs_id, required_irf="full-enclosure", require_events=True):
+    def obs(
+        self,
+        obs_id,
+        required_irf="full-enclosure",
+        require_events=True,
+        obs_table_row=None,
+    ):
         """Access a given `~gammapy.data.Observation`.
 
         Parameters
@@ -286,6 +295,8 @@ class DataStore:
             Default is `"full-enclosure"`.
         require_events : bool, optional
             Require events and gti table or not. Default is True.
+        obs_table_row : int, optional
+            Row number in the observation table. Default is None
 
         Returns
         -------
@@ -293,8 +304,11 @@ class DataStore:
             Observation container.
 
         """
-        if obs_id not in self.hdu_table["OBS_ID"]:
-            raise ValueError(f"OBS_ID = {obs_id} not in HDU index table.")
+        if obs_id not in self.obs_ids:
+            raise ValueError(f"OBS_ID = {obs_id} not in index tables.")
+
+        if obs_table_row is None:
+            obs_table_row = obs_id
 
         kwargs = {"obs_id": int(obs_id)}
 
@@ -316,7 +330,7 @@ class DataStore:
         missing_hdus = []
         for hdu in ALL_HDUS:
             hdu_location = self.hdu_table.hdu_location(
-                obs_id=obs_id,
+                obs_id=obs_table_row,
                 hdu_type=hdu,
                 warn_missing=False,
             )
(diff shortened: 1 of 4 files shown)


QRemy:
Further examples and ideas are given here  :

https://github.com/QRemy/gammapy-data/blob/add_notebook/tests/format/swgo/test_format.ipynb

QRemy:
Requires https://github.com/gammapy/gammapy-data/pull/66.

bkhelifi:
Unfortunately, this could not work for CTAO, to be checked. Indeed, 
- there is **no** notion of `event_class`
- IRFs are associated to events not only per `event_type` but per `event_type` AND `SOI` (Stable Time of Interval)
- there will be events not associated to IRFs that will be stored into the HDU events

Also, I am wondering how this scheme is compliant with IVOA standards. The discovery of data is generally based on query on an RoI, a Time Interval, an observatory and sometimes an instrument (or subarray). It returns then a list of files, associated to a minimal data set (whatever the choice of data curation/repartition defined by the observatory). Each of some have an ObsID, and in general with several files.
The main question is how to link a physical observation to this list of files, at the level of the VO TAP server and as the output of a data server query (VO, CTAO, NED, etc)? This is a question that the IVOA HEIG is currently working. There is a clear link with the ObsCore format... our GADF obs-index table is about that. But it was assuming that there was only 1 event file per observation (which is not general enough).
What is up to VODF is to link, for a given file, the event HDU to IRF HDUs. One should have within each file a 'hdu table' that makes the link between all hdu.
Do we need for Gammapy a global HDU table containing all observations and all files? This is an open question... And for performance reason, do we need for Gammapy a global observations table? This is an open question... And are the formats of these 2 tables associated to the IVOA and VODF formats, or are they an Gammapy format? This is also an open question.

Maybe, one way to work is to think about the Gammapy needs (requirement or feature request) on the offered functionalities, and on performances. And them, use these needs to see which existing format on can use, or which format should be extended, or which proposal one can make to VODF (karl, laura and jutta). It would be great to define cleary our needs to analyse HESS, CTAO, SWGO (and KM3NeT). I made the assumption that Gammapy will be its compatibility with the GADF and Fermi-LAT format in the future... (which is also a need).

QRemy:
> * there is **no** notion of `event_class`
each different reconstruction is an event_class so there will be different event class in CTA
> * IRFs are associated to events not only per `event_type` but per `event_type` AND `SOI` (Stable Time of Interval)
that the observation block here inlcude both Stable Time of Interval and Unstable Time of Interval
> * there will be events not associated to IRFs that will be stored into the HDU events
same as previous

So want I suggest  actually work for all these tree points

bkhelifi:
> each different reconstruction is an event_class so there will be different event class in CTA
Each different reconstruction will lead to a different data release. To my knowledge, DPPS will not plan to have different reconstruction for the same release. This is not a requirement, is not it, @kosack ?

> Regarding the ObsCore relation it's simple : 1 ObsCore entry = 1 obs-index row = 1 minimal independent data partition
This is clear: what I say is that you need the 2 tables to create a link to one ObsCore entry (where as in GADF it was only 1).

bkhelifi:
For the Gammapy requirements, it is clear that our code should handle GADF and VODF. To be checked, naturally... Also, Gammapy should be able to propagate and reduce metadata at each step of analysis (DL3->4, DL4->5), as well as creating the provenance for each written data.

Concerning the future VODF format, there are some general principles to check: individual files should be FAIR and associated with provenance (some hierarchical mechanism will be naturally of use in order to not have 10G of history log per file!), should handle any data producer choice of partitioning of data obtained after an astrophysical observation (that is associated to many metadata, for a pointing instrument there is the notion of proposal).

kosack:
I would suggest before thinking about re-design, make a list of required features.   It's hard to see exactly what problems are being solved, and which ones are missing.

QRemy:
Closing this, I'll open separate PRs to include the minimal modifications needed for SWGO
and the dataset selection and observations grouping helper functions shown in the notebook.

PR #5688, opened 2025-01-30
[PROTOTYPE] Introduce Fit Statistic classes

registerrier:
**Description**

This pull request takes over from #4237 and proposes a possible `FitStatistic` class. It is meant to help support a number of recent changes or current PR:

-  Handling different type of statistics for the same type of Dataset
   - Currently we use different chi square-like statistics in `FluxPointsDataset` which results in a complex handling in `stat_sum`
	- `stat_array_chi2` : regular $\chi^2$
	- `stat_array_profile` : use the likelihood profile to compute the likelihood per data point 
	- `stat_array_distrib` : use asymmetric errors to compute an asymmetric $\chi^2$ and UL
  - Support weighted cash statistic for (Fermi) event types analysis
      - See the following PR:  [Support weighted likelihood in MapDataset](https://github.com/gammapy/gammapy/pull/5663)
		- A new possible `stat` attribute for the `MapDataset` is introduced
		- For clarity a new type of `MapDataset` is introduced
       - This could be handled with a different statistic function 
- ML, penalized ML and Bayesian analyses:
- In general we might want to support:
	- Regular ML 
		- using -2 log L  fit statistic
	- MAP (Maximum A Posteriori):
		- regular maximum likelihood fit statistic with the addition of -2 log priors on all parameters
	- Penalized Maximum Likelihood:
		- Use a more complex regularization scheme for ML
		- e.g. Tikhonov regularization
		- this might be a simpler and more natural approach than handling complex multi-parameter priors
	- Bayesian analysis and Nested Sampling:
		- Compute log likelihood stat_sum rather than -2 log like without priors
		- that a prior is set on every parameter and to compute inverse cdf on the global Parameters prior

Here, we propose to use stateless classes to implement the main functions `stat_array`, `stat_sum` and `loglikelihood` (as static methods).

A registry can be used to access the correct class from a given dataset. 

For now `FitStatistic` can be simply limited to `Dataset.stat_array` and `Dataset.stat_sum`, but we can introduce later a more general `JointFitStatistic` that can be defined on `Datasets`.

Opinions @QRemy @adonath @AtreyeeS @facero @bkhelifi ?

### DIFF
--- a/gammapy/datasets/core.py
+++ b/gammapy/datasets/core.py
@@ -67,6 +67,16 @@ class Dataset(abc.ABC):
         elif self.mask_safe is not None:
             return self.mask_safe
 
+    def _get_stat_sum_inputs(self):
+        """Return only requested keys, computing `npred` dynamically if needed."""
+        inputs = {}
+        for key in self._fit_statistic.required_inputs():
+            if "npred" in key:
+                inputs[key] = getattr(self, key)()  # Compute npred dynamically
+            else:
+                inputs[key] = getattr(self, key)
+        return inputs
+
     def stat_sum(self):
         """Total statistic given the current model parameters and priors."""
         prior_stat_sum = 0.0
--- a/gammapy/datasets/core.py
+++ b/gammapy/datasets/core.py
@@ -8,6 +8,7 @@ import numpy as np
 from astropy import units as u
 from astropy.table import Table, vstack
 from gammapy.data import GTI
+from gammapy.maps import Map
 from gammapy.modeling.models import DatasetModels, Models
 from gammapy.utils.scripts import make_name, make_path, read_yaml, to_yaml, write_yaml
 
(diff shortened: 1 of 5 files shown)


gammapy/datasets/core.py
```diff
@@ -71,7 +71,9 @@ def stat_sum(self):
         """Total statistic given the current model parameters and priors."""
         prior_stat_sum = 0.0
         if self.models is not None:
-            prior_stat_sum = self.models.parameters.prior_stat_sum()
+            prior_stat_sum = (
+                self.models.parameters.prior_stat_sum()
+            )  # This shouldn't be computed here
```
QRemy, line 76: it could if we redefine
`datasets.stat_sum = datasets_stat_sum_likelihood+self.models.parameters.prior_stat_sum()`

gammapy/datasets/map.py
```diff
@@ -1206,7 +1207,7 @@ def stack(self, other, nan_to_num=True):
 
     def stat_array(self):
         """Statistic function value per bin given the current model parameters."""
-        return cash(n_on=self.counts.data, mu_on=self.npred().data)
+        return self._fit_statistic.stat_array(self.counts.data, self.npred().data)
```
facero, line 1210: Wouldn't the argument of fit_statistics change depending on which stat func is called ?

facero:
Thanks
So just that I understand one would define a `FitStatistic` object to point to the appropriate stat function depending on the type of analysis right ? 
The user does not have to select anything, the good default would be defined under the hood right

facero:
Also in the case of fitting multi-instrument with `JointFitStatistic` for example Fermi 3D + HESS SED flux point, one would need two different `FitStatistic` for each right. 
There is probably no show stopper here but just to need to make sure it works as expected

registerrier:
The solution here is likely too complex. See instead #5696 . Closing this PR draft.

PR #5697, opened 2025-02-10
Add docstrings for Models.plot_regions/plot_positions

AtreyeeS:
It seems that the `color` keyword was not being handled correctly through `kwargs_point` for the plot_regions

### DIFF
--- a/gammapy/maps/region/geom.py
+++ b/gammapy/maps/region/geom.py
@@ -29,6 +29,7 @@ from gammapy.utils.regions import (
     regions_to_compound_region,
 )
 from gammapy.visualization.utils import ARTIST_TO_LINE_PROPERTIES
+from gammapy.visualization.utils import ARTIST_TO_LINE_PROPERTIES
 from ..axes import MapAxes
 from ..coord import MapCoord
 from ..core import Map
--- a/gammapy/maps/region/geom.py
+++ b/gammapy/maps/region/geom.py
@@ -852,7 +852,16 @@ class RegionGeom(Geom):
             kwargs.setdefault("facecolor", "None")
             kwargs.setdefault("edgecolor", "tab:blue")
             kwargs_point.setdefault("marker", "*")
-            kwargs_point.setdefault("color", "tab:blue")
+
+            for key, value in kwargs.items():
+                key_point = ARTIST_TO_LINE_PROPERTIES.get(key, None)
+                if key_point and key_point not in kwargs_point:
+                    kwargs_point[key_point] = value
+                    
+            if "color" in kwargs:
+                kwargs_point.setdefault("color", kwargs["color"])
+            if "color" in kwargs_point:
+                kwargs_point["markeredgecolor"] = kwargs_point["color"]
 
             for region in compound_region_to_regions(self.region):
                 region_pix = region.to_pixel(wcs=ax.wcs)
--- a/gammapy/maps/region/geom.py
+++ b/gammapy/maps/region/geom.py
@@ -29,7 +29,6 @@ from gammapy.utils.regions import (
     regions_to_compound_region,
 )
 from gammapy.visualization.utils import ARTIST_TO_LINE_PROPERTIES
-from gammapy.visualization.utils import ARTIST_TO_LINE_PROPERTIES
 from ..axes import MapAxes
 from ..coord import MapCoord
 from ..core import Map
@@ -859,7 +858,7 @@ class RegionGeom(Geom):
                 key_point = ARTIST_TO_LINE_PROPERTIES.get(key, None)
                 if key_point and key_point not in kwargs_point:
                     kwargs_point[key_point] = value
-                    
+
             if "color" in kwargs:
                 kwargs_point.setdefault("color", kwargs["color"])
             if "color" in kwargs_point:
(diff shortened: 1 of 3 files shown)


Astro-Kirsty (changes requested):
Nice pick up @AtreyeeS. 

To allow the examples to be testable you need to add either `>>>` before each line or `.. testcode::` just after the title. See [here](https://docs.gammapy.org/1.3/development/doc_howto.html#code-in-docstrings-in-python-files)
You can then test if the example code works correctly through 
`pytest --doctest-modules --ignore-glob=*/tests gammapy/maps/region/geom.py` for example

AtreyeeS:
Thanks @Astro-Kirsty ! Added the `>>>`

In future, it maybe good to move these functions to gammapy.visualisations

gammapy/maps/region/geom.py
```diff
@@ -853,11 +852,7 @@ def plot_region(self, ax=None, kwargs_point=None, path_effect=None, **kwargs):
             kwargs.setdefault("facecolor", "None")
             kwargs.setdefault("edgecolor", "tab:blue")
             kwargs_point.setdefault("marker", "*")
-
-            for key, value in kwargs.items():
-                key_point = ARTIST_TO_LINE_PROPERTIES.get(key, None)
-                if key_point:
-                    kwargs_point[key_point] = value
```
QRemy, line 860: this is suppose to give the same properties to points using kwargs is kwargs_point is not specified, it should not be remove

AtreyeeS: I found that this makes it impossible to set the colour of the points through `kwargs_point`.
Removing these lines makes no difference to the default plot, and the choices can still be passed cleanly

gammapy/maps/region/geom.py
```diff
@@ -853,11 +852,7 @@ def plot_region(self, ax=None, kwargs_point=None, path_effect=None, **kwargs):
...
+            kwargs_point.setdefault("color", "tab:blue")
```
QRemy, line 855:
```suggestion
            kwargs_point_default = kwargs.copy()            
            kwargs_point_default.update(kwargs_point)

            for key, value in kwargs_point_default.items():
                key_point = ARTIST_TO_LINE_PROPERTIES.get(key, None)
                if key_point and key_point not in kwargs_point:
                    kwargs_point[key_point] = value
                    
            if "color" in kwargs:
                kwargs_point.setdefault("color", kwargs["color"])
            if "color" in kwargs_point:
                kwargs_point["markeredgecolor"] = kwargs_point["color"]
```

QRemy: so in the example :

```
from gammapy.catalog import SourceCatalog3FHL
fermi_dataset = MapDataset.read("$GAMMAPY_DATA/fermi-3fhl-gc/fermi-3fhl-gc.fits.gz", name="fermi_dataset")
catalog = SourceCatalog3FHL()
models = catalog.to_models().select_from_geom(fermi_dataset.geoms["geom"])
ax = fermi_dataset.excess.sum_over_axes().smooth("0.2 deg").plot(add_cbar=True, cmap="Blues")
ax = models.plot_regions(ax=ax, linewidth=1,  color="red")

```

points should also be red and not blue

QRemy: and 

```
ax = models.plot_regions(ax=ax, linewidth=1,  color="red", edgecolor="green"
kwargs_point={"marker":"d", "markersize":5, "markeredgecolor":"green"})
```
will be red because color has priority over edgecolor as for the lines

gammapy/maps/region/geom.py
```diff
@@ -28,7 +28,6 @@
     compound_region_to_regions,
     regions_to_compound_region,
 )
-from gammapy.visualization.utils import ARTIST_TO_LINE_PROPERTIES
```
registerrier, line 31: This seems to break the code.

registerrier (commented):
Thanks @AtreyeeS . What is the status here?

gammapy/maps/region/geom.py
```diff
@@ -28,7 +28,6 @@
     compound_region_to_regions,
     regions_to_compound_region,
 )
```
QRemy, line 30:
```suggestion
)
from gammapy.visualization.utils import ARTIST_TO_LINE_PROPERTIES
```

QRemy:
> Thanks @AtreyeeS . What is the status here?

Fixed and tests added, should be read now.

registerrier (approved):
Thanks @AtreyeeS and @QRemy . This looks good.

The examples read well.

Astro-Kirsty:
Hello, related to this in the 1D spectral tutorial when running:
`plot_spectrum_datasets_off_regions(ax=ax, datasets=datasets)`

We have the warning:
`/home/runner/work/gammapy-docs/gammapy-docs/gammapy/.tox/build_docs/lib/python3.9/site-packages/regions/shapes/circle.py:161: UserWarning: Setting the 'color' property will override the edgecolor or facecolor properties.
  return Circle(xy=xy, radius=radius, **mpl_kwargs)
/home/runner/work/gammapy-docs/gammapy-docs/gammapy/.tox/build_docs/lib/python3.9/site-packages/gammapy/visualization/datasets.py:84: UserWarning: Setting the 'color' property will override the edgecolor or facecolor properties.
  handle = Patch(**plot_kwargs)`
  
  Is it related?

PR #5708, opened 2025-02-19
Use openastronomy workflow to build wheels

### LINKED ISSUE #4656: Consider distributing wheels on PyPI

At the moment there are no wheels on PyPI for `gammapy` (e.g. c.f. https://pypi.org/project/gammapy/1.1/#files for `gammapy` `v1.1`). As `gammapy` is pure-Python it should be straightforward to build wheels along with the `sdist` with `build`. You're already using `build` though in

https://github.com/gammapy/gammapy/blob/d775861ff491f1a0301d5505aa039f24eb3402bc/.github/workflows/release.yml#L21-L25

so is there a reason why you don't build and distribute wheels?

Build and distribute `wheels` to PyPI (can use a workflow like https://github.com/scikit-hep/pyhf/blob/4845a154c18edfd5e2b29ef94167b2818fa6e674/.github/workflows/publish-package.yml).

There is no alternative, but there might be a reason that you intentionally don't distribute wheels that I'm missing.

**Additional context**
N/A

maxnoe:
**Description**

This pull request replaces the custom pypi deployment step in the github CI yaml with the open astronomy action that builds wheels.

See https://github-actions-workflows.openastronomy.org/en/stable/publish.html

Fixes #4656 

**Dear reviewer**


maxnoe:
I reduced the number of builds now by excluding pypy builds, windows 32 bit and less common linux variantes (musl, armv7, etc.)

maxnoe:
Note: the last CI run didn't build the wheels because I re-enabled the workflow rules as they were before.

In the previous commit, you can see which wheels were build:


- https://github.com/gammapy/gammapy/actions/runs/13416279534/job/37477914860?pr=5708
- https://github.com/gammapy/gammapy/actions/runs/13416279534/job/37477915402?pr=5708
- https://github.com/gammapy/gammapy/actions/runs/13416279534/job/37477915897?pr=5708

maxnoe:
Maybe we could even skip the macos 10 builds?

.github/workflows/release.yml
```diff
@@ -8,26 +8,14 @@ on:
...
-          user: __token__
-          password: ${{ secrets.PYPI_TOKEN }}
+    uses: OpenAstronomy/github-actions-workflows/.github/workflows/publish.yml@v1
+    with:
+      targets: |
+        - cp3*-manylinux_x86_64*
+        - cp3*-macosx*
+        - cp3*-win_amd64*
+    secrets:
+      pypi_token: ${{ secrets.PYPI_TOKEN }}
```
matthewfeickert, line 18: Does this GitHub Action not use [trusted publishers](https://docs.pypi.org/trusted-publishers/) yet? I'm in general not a fan of these "build and publish all in one go" workflows as they usually end up causing a lot of pain when they mess up. It is better to build your distributions in a different step from publishing. It also reduces the permissions that you need to give to any one workflow, which is better for security.

adonath: No, I think we haven't configured trusted publishers yet, I'll tag @registerrier on this. I agree on the separate build and publishing step. Alternatively we could either cron or independently trigger the builds for testing as described in the link. This way we can still keep the workflow @maxnoe proposed.

maxnoe: The pipeline is actually multi stage, this is not a single action building and publishing wheels, see:
https://github.com/gammapy/gammapy/actions/runs/13416279534

I thought it makes sense to re-use the open astronomy workflow, which is e.g. also used by astropy:

https://github.com/astropy/astropy/blob/60aa54bc65f59f97c02b26d7c00063a3fa5a52a1/.github/workflows/publish.yml#L23-L30

But if you prepare to roll out a custom workflow or one based on another upstream resource, fine with me

adonath: Thanks @maxnoe for clarifying! I agree we should re-use existing workflows from the astronomy community. I guess the difference to Astropy is that they have configured nightly builds? This way they always see, when builds fail and can fix the next day. While for us it will only trigger on a release tag. But as we usually do release candidates I am not too worried, so we could still fix things...

adonath (dismissed):
Thanks @maxnoe, no further comments from my side!

Do you have any idea whether we can deploy wheels retroactively for past versions? I think one can attach new files to pypi, as long as the they don't modify existing ones. However it probabaly requires to do this manually...have you ever done this before?

maxnoe:
Alternatively, we could backport these changes to the still supported versions and release `.post` or a point release

adonath:
> Alternatively, we could backport these changes to the still supported versions and release .post or a point release

Thanks for the suggestion, I think that's the best compromise! Backporting should be easy and then we just trigger a new point release. What do you think @registerrier?

morcuended (dismissed):
Thanks @maxnoe. I also agree with reusing the existing workflow from openastronomy.

registerrier (dismissed):
Thanks @maxnoe !  This is good.
I agree that re-using astropy/openastronomy frameworks is a good thing. 
Regarding publishing wheels for past releases, we have to see which solution is less time costly as making a bugfix release still requires some manual work.

Let's backport to 1.3.x and 1.0.x and see how things go.

PR #5712, opened 2025-02-26
Add parameter limits on phi for Spatial models

AtreyeeS:
Having limits on the rotation angle for the asymmetry makes the fit more stable.
Also, adapted the warning message for parameter range to warn when the parameter falls strictly outside the range (rather than equals) because otherwise the fit starts with a warning when unfreezing the `e`.


Astro-Kirsty (changes requested):
Thanks @AtreyeeS, this is a good idea. 

Can you also adjust the docstring so that it specifies the min and max, like we have for `e` ?

gammapy/modeling/models/spatial.py
```diff
@@ -736,10 +736,10 @@ class GeneralizedGaussianSpatialModel(SpatialModel):
         Shape parameter within (0, 1). Special cases for disk: ->0, Gaussian: 0.5, Laplace:1
         Default is 0.5.
     e : `float`
-        Eccentricity (:math:`0< e< 1`).
+        Eccentricity (:math:`0<= e< =1`).
         Default is 0.
     phi : `~astropy.coordinates.Angle`
-        Rotation angle :math:`\phi`: of the major semiaxis.
+        Rotation angle :math:`\phi`: of the major semiaxis (0<=phi<=180).
```
Astro-Kirsty, line 742:
```suggestion
        Rotation angle :math:`\phi`: of the major semiaxis (:math:`0<=phi<=180`).
```

gammapy/modeling/models/spatial.py
```diff
@@ -859,10 +859,10 @@ class DiskSpatialModel(SpatialModel):
         :math:`a`: length of the major semiaxis, in angular units.
         Default is 1 deg.
     e : `float`
-        Eccentricity of the ellipse (:math:`0< e< 1`).
+        Eccentricity of the ellipse (:math:`0<= e<= 1`).
         Default is 0.
     phi : `~astropy.coordinates.Angle`
-        Rotation angle :math:`\phi`: of the major semiaxis.
+        Rotation angle :math:`\phi`: of the major semiaxis (0<=phi<=180).
```
Astro-Kirsty, line 865:
```suggestion
        Rotation angle :math:`\phi`: of the major semiaxis (:math:`0<=phi<=180`).
```

registerrier (commented):
Thanks @AtreyeeS . This can help indeed. What is not clear to me is how the fitter will behave for solutions close to phi=0 or 180 degrees. 
Have you tested what this yields when fitting an ellipse with phi=0? Maybe a somewhat larger range would alleviate the problem. Or a minima documenting what to do for these cases.

bkhelifi:
Should we then add at least a circular [0,360]deg or [-180,180]deg? This should be better for minimizers...

AtreyeeS:
I added a wrap function, but just having the wrap inside the evaluate function does not seem to suffice.
So I kept the bounds, but increased them.
Fitting around phi=0 now works with reasonable errors

gammapy/modeling/models/spatial.py
```diff
@@ -736,10 +739,11 @@ class GeneralizedGaussianSpatialModel(SpatialModel):
         Shape parameter within (0, 1). Special cases for disk: ->0, Gaussian: 0.5, Laplace:1
         Default is 0.5.
     e : `float`
-        Eccentricity (:math:`0< e< 1`).
+        Eccentricity (:math:`0<= e< =1`).
         Default is 0.
     phi : `~astropy.coordinates.Angle`
-        Rotation angle :math:`\phi`: of the major semiaxis.
+        Rotation angle :math:`\phi`: of the major semiaxis (:math:`-20<=phi<=200`).
+        phi is wrapped between [0, 180] with the larger range enabled to enable fits for phi close to 0
```
QRemy, line 746: "with the larger range enabled to enable fits for phi close to 0"  : I think this is an implementation detail.

gammapy/modeling/models/spatial.py
```diff
@@ -859,10 +864,11 @@ class DiskSpatialModel(SpatialModel):
         :math:`a`: length of the major semiaxis, in angular units.
         Default is 1 deg.
     e : `float`
-        Eccentricity of the ellipse (:math:`0< e< 1`).
+        Eccentricity of the ellipse (:math:`0<= e<= 1`).
         Default is 0.
     phi : `~astropy.coordinates.Angle`
-        Rotation angle :math:`\phi`: of the major semiaxis.
+        Rotation angle :math:`\phi`: of the major semiaxis (:math:`-20<=phi<=200`).
+        phi is wrapped between [0, 180] with the larger range enabled to enable fits for phi close to 0
```
QRemy, line 871: "with the larger range enabled to enable fits for phi close to 0"  : I think this is an implementation detail.

gammapy/modeling/models/spatial.py
```diff
@@ -622,7 +624,7 @@ class GaussianSpatialModel(SpatialModel):
     lat_0 = Parameter("lat_0", "0 deg", min=-90, max=90)
     sigma = Parameter("sigma", "1 deg", min=0)
     e = Parameter("e", 0, min=0, max=1, frozen=True)
-    phi = Parameter("phi", "0 deg", frozen=True)
+    phi = Parameter("phi", "0 deg", frozen=True, min=-20, max=200)
```
QRemy, line 627: does this limit the fitted error if the true error is larger than 20 ?

QRemy (commented):
Thanks @AtreyeeS, just few more comments:

gammapy/modeling/models/spatial.py
AtreyeeS: Thanks @QRemy ! Actually I could not simulate a source with significant ellipticity, but error on angle greater than 20.
But you are right, 20 is a bad number.
I now put it between -180 and 180.

gammapy/modeling/models/spatial.py
AtreyeeS: removed

gammapy/utils/units.py
```diff
@@ -1,15 +1,41 @@
...
+
+    Returns
+    -------
+    wrapped_val : float or `~astropy.units.Unit`
+        Wrapped value
+    """
+
+    # Note: Astropy wrap_at currently works only for wrapping
+    # between [0, 360] and [-180, 180]
+    range_size = max_value - min_value
```
bkhelifi, line 34: Should we check that the min is really the min?

gammapy/utils/units.py
```diff
@@ -1,15 +1,41 @@
...
+    wrapped_val = ((value - min_value) % range_size) + min_value
```
bkhelifi, line 35: The function `wrap_at` will be called at each parameter evaluation, meaning a lot during a fit! 
Does this function a bottleneck? Or should this line be optimised for speed? This is an open question... Some time processing evaluation might be needed

bkhelifi (changes requested):
Thank you for this PR, closed to be accepted. I put an inline comment...

gammapy/utils/units.py
QRemy: the function is not complicated should be fast enough compared to the other parts of the evaluation function

QRemy (approved):
Thanks @AtreyeeS no further comment, for me it's ready to merge.

gammapy/utils/tests/test_units.py
```diff
@@ -39,3 +40,8 @@ def test_standardise_unit():
 @pytest.mark.parametrize("q, expect", values)
 def test_energy_unit_format(q, expect):
     assert energy_unit_format(q) == expect
+
+
+def test_wrap_at():
+    computed = wrap_at(6.5 * u.deg, 1.0 * u.deg, 3.5 * u.deg)
+    assert_allclose(1.5 * u.deg, computed, rtol=1e-4)
```
QRemy, line 47:
```suggestion
    computed = wrap_at(6.5 * u.deg, 1.0 * u.deg, 3.5 * u.deg)
    assert_allclose(1.5 * u.deg, computed, rtol=1e-4)
    
    computed = wrap_at(np.arange(10), 1.0, 3.5)
    assert_allclose(computed.min(), (1.0, rtol=1e-4)
    assert_allclose(computed.max(), (3.5, rtol=1e-4)

```

gammapy/utils/units.py
QRemy: added a test for that

bkhelifi (approved):
Thanks for this PR. No further comment

PR #5714, opened 2025-02-28
Fix caching in WcsGeom

### LINKED ISSUE #5706: Possible memory leak in FoVBackgroundMaker
I am having a rather strange issue that is a bit difficult to reproduce. I have a (large) number of pre-reduced datasets, and I want to read them in a loop, apply a  `FoVBackgroundMaker`, and stack them. A minimal example (in this example, I am reading the same file N times).

```
import numpy as np
import astropy.units as u

from gammapy.maps import Map, MapAxis, WcsGeom
from gammapy.modeling.models import *

from gammapy.makers import FoVBackgroundMaker, SafeMaskMaker


dataset = MapDataset.read( "test1.fits.gz")
exclusion_mask = empty._geom.to_image().region_mask([reg], inside=False)

safe_maker = SafeMaskMaker(methods=['bkg-peak'], )
PowerLawNormSpectralModel.tilt.frozen = False
pwl = PowerLawNormSpectralModel()
print(pwl)

bkg_maker = FoVBackgroundMaker(method='fit', 
                               exclusion_mask=exclusion_mask,
                               spectral_model=pwl,
                              min_counts=100)

N = 500

for i in range(N):
    if i%10 == 0:
        print('Running for ',i, '/',N)
    ds = MapDataset.read("test_ds.fits.gz")
    #ds = safe_maker.run(ds) # The issue remains if this line is commented out
    ds = bkg_maker.run(ds) 
    empty.stack(ds)

```

The problem I am getting is, after around 60-200 iterations (varies randomly), I get a message `The kernel died. It will restart automatically`.
There is no helpful message on the terminal either, just `The kernel died. It will restart automatically`

I am using jupyterlab 4.2.6 on an Ubuntu 22.02 machine with 32 GB RAM, core i7 processor.
Problem persists if I run it as an ipython script

I have not been able to reproduce this on a Mac OS yet

The file test1.fits.gz can be found here: https://drive.google.com/file/d/1lsmsmzczRkF_5lNVXAdX-o3DTXnw1sHp/view?usp=sharing
I cant reproduce the issue with the files in gammapy-data !

I have no idea how to understand this issue. Any ideas @adonath @maxnoe ?

AtreyeeS:
This seems to fix #5706

### DIFF
--- a/gammapy/maps/wcs/geom.py
+++ b/gammapy/maps/wcs/geom.py
@@ -114,8 +114,8 @@ class WcsGeom(Geom):
         self._crpix = crpix
 
         # define cached methods
-        self.get_coord = lru_cache()(self.get_coord)
-        self.get_pix = lru_cache()(self.get_pix)
+        self.get_coord = lru_cache()(self._get_coord)
+        self.get_pix = lru_cache()(self._get_pix)
 
     def __setstate__(self, state):
         for key, value in state.items():
@@ -596,6 +596,22 @@ class WcsGeom(Geom):
         coord : tuple
             Map pixel coordinate tuple.
         """
+        return self._pix(idx=idx, mode=mode)
+
+    def _get_pix(self, idx=None, mode="center"):
+        """Get map pixel coordinates from the geometry.
+
+        Parameters
+        ----------
+        mode : {'center', 'edges'}, optional
+            Get center or edge pix coordinates for the spatial axes.
+            Default is "center".
+
+        Returns
+        -------
+        coord : tuple
+            Map pixel coordinate tuple.
+        """
         pix = self._get_pix_all(idx=idx, mode=mode)
         coords = self.pix_to_coord(pix)
         m = np.isfinite(coords[0])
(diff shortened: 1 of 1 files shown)


gammapy/maps/wcs/geom.py
```diff
@@ -585,6 +585,22 @@ def _get_pix_all(
...
+        mode : {'center', 'edges'}, optional
+            Get center or edge pix coordinates for the spatial axes.
+            Default is "center".
+
+        Returns
+        -------
+        coord : tuple
+            Map pixel coordinate tuple.
+        """
+        return self._pix(idx=idx, mode=mode)
```
QRemy, line 599:
```suggestion
        return self._get_pix(idx=idx, mode=mode)
```

QRemy: this should break so this function is not use all all
could be simply `pass`

QRemy: Moreover I don't see any improvement, a reference to self is still give to lru_cache so the leak is still there

gammapy/maps/wcs/geom.py
```diff
@@ -114,8 +114,8 @@ def __init__(self, wcs, npix=None, cdelt=None, crpix=None, axes=None):
         self._crpix = crpix
 
         # define cached methods
-        self.get_coord = lru_cache()(self.get_coord)
-        self.get_pix = lru_cache()(self.get_pix)
+        self.get_coord = lru_cache()(self._get_coord)
+        self.get_pix = lru_cache()(self._get_pix)
```
QRemy, line 118: I don't see improvement we the proposed solution but I think we can fix it by forcing garbage collection here: 
```suggestion
        gc.collect()
        self.get_coord = lru_cache()(self.get_coord)
        self.get_pix = lru_cache()(self.get_pix)
```

For my test I run
```
import astropy.units as u

dataset = MapDataset.read( "test1.fits")

N=10
for i in range(N):
    ds = MapDataset.read("test1.fits")  

```
and I get for this PR
pr

while adding gc.collect just before defining the cached method
geom_collect

registerrier (commented):
Thanks @AtreyeeS . 
If you implement this solution then I suppose the method `get_coord` and `get_pix` should not be explicitly implemented right? They are defined in __init__ by the lru_cache call. No?

registerrier:
Additional comments from the dev call:
- Simply calling `gc.collect()` on `__init__` might be a way out, but we need to make sure that the cost is not too large
- Additionally, we could also add a how to to suggest users having memory leak issues to use this solution after a large loop. Matplotlib does something similar [here](https://matplotlib.org/stable/users/prev_whats_new/whats_new_3.6.0.html#garbage-collection-is-no-longer-run-on-figure-close)  
- there might be a performance gain in caching ourselves coordinates and solid angles as they could be cutout/copied and not recomputed each time a Map is copied or cutout.

AtreyeeS:
This does not work! Sorry, I think I had tested something wrong the first time around.
Adding `gc.collect` on init does solve this issue, but it does increase the runtime around 15% for this simple example, so I dont know if we should do it.
Calling `gc.collect` inside the use for loop also works, but I dont know if we should leave it to the users. 
I am closing this PR, we can continue discussion on the issue maybe?

PR #5715, opened 2025-03-04
Modified Polygon region for use as ON region

ParsonsRD:
**Description**

This is a minimal reimplementation of the polygon region class to allow it to be used as a signal extraction region in analysis. The significant addition to this was the addition of a center class member which is missing in the regions implementation of this class. This centre is calculated as the geometric centre of the points when the class is created. The Polygon object was creasted as a sky region and a pixel region to match the regions class behaviour and the centre is propagated between them.

**Dear reviewer**

I created this as a draft to check that I was taking the correct approach here for implementation, also it is not clear to me how best to implement tests for this class.

### DIFF
--- a/gammapy/utils/regions.py
+++ b/gammapy/utils/regions.py
@@ -292,7 +292,6 @@ class PolygonPointsPixelRegion(PolygonPixelRegion):
         center = self.center.rotate(center, angle)
 
         return self.copy(vertices=vertices, center=center)
-    
 
 def make_orthogonal_rectangle_sky_regions(start_pos, end_pos, wcs, height, nbin=1):
     """Utility returning an array of regions to make orthogonal projections.
--- a/gammapy/utils/regions.py
+++ b/gammapy/utils/regions.py
@@ -143,8 +143,12 @@ def regions_to_compound_region(regions):
     return region_union
 
 def get_centroid(vertices):
-    """Compute centroid of a polygon.
+    """Compute centroid of a polygon. Implicitly assumes a flat
+    cartesian projection, will probably break for very large polygons.
     
+    Code comes from:
+    https://stackoverflow.com/questions/75699024/finding-the-centroid-of-a-polygon-in-python
+
     Parameters
     ----------
     vertices : `~astropy.coordinates.SkyCoord`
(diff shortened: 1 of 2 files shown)


QRemy:
Thanks @ParsonsRD, maybe it should be added to regions rather than gammapy, opinions @adonath @registerrier ?

adonath:
@QRemy I have discussed this with @ParsonsRD shortly in private and the problem with adding this to `regions` right now, is that there is no single agreed upon definition of what e.g. the center is, than there is the question of spherical polygons, which algorithm to use for "contains" etc. I think it is fine to provide a prototype / reference implementation in Gammapy and then strat discussions and finally move upstream. This is true for most functionality in `gammapy/utils/regions.py`. At the same time it provides a real benefit for Gammapy users already.

bkhelifi:
I am wondering whether the exact definition of this region centre is very critical. How thought about that? This implementation looks promising to me...
In the mid-run, should we also propose a feature request to astropy?

adonath:
> I am wondering whether the exact definition of this region centre is very critical. How thought about that? This implementation looks promising to me...

I think it is mostly a matter of convention. E.g. I have introduce a helper method to compute the "center" of a set of points via it's geometric median: https://github.com/gammapy/gammapy/blob/main/gammapy/utils/regions.py#L38

Not sure whether this give the same results as what @ParsonsRD introduces in this PR. We do not necessarily have to strive for consistency, but be should be ware that there are different possible approaches.

registerrier:
Thanks @ParsonsRD . This is interesting. To proceed this would require tests to be implemented.
Can we compare the center found in simple cases with expected values?

ParsonsRD:
Indeed this was what I was thinking, to create a region that is a equivalent to another implemented region shape and check we get the same result. Do you know if there is an example of a test like this that I could try to reproduce?

QRemy:
Added AI generated unit tests.

bkhelifi (dismissed):
Thanks @ParsonsRD and @QRemy

QRemy (approved):
Thanks @ParsonsRD, merging now.

ParsonsRD:
One stupid point, I think the code for calculating the polygon area is stack overflow code. Do you usually include an acknowledgment in the code comments or something for gammapy?

gammapy/utils/regions.py
```diff
@@ -137,6 +144,44 @@ def regions_to_compound_region(regions):
     return region_union
 
 
+def get_centroid(vertices):
+    """Compute centroid of a polygon. Implicitly assumes a flat
+    cartesian projection, will probably break for very large polygons.
+
+    Code comes from:
+    https://stackoverflow.com/questions/75699024/finding-the-centroid-of-a-polygon-in-python
```
QRemy, line 152: @ParsonsRD it's fine to acknowledge it like this, we have the same in other places in gammapy

PR #5716, opened 2025-03-05
Documentation appearance

### LINKED ISSUE #5520: Enable dark theme in Gammapy docs with proper visualization
Being able to activate the dark mode of the docs.

So far the button to switch the color theme was disabled from the docs conf.py

**Additional context**
The problem with the current implementation is that the font color of the main body of the docs is very similar to the background. The CSS template has to be worked out.

### LINKED ISSUE #5586: Documentation appearance
This issue aims to list all the current issues with the documentation since the update of PyData Sphinx Theme to  0.16.0.
Please add any other issues you see!

We need to update the conf.py and css files. As per https://github.com/gammapy/gammapy/issues/5506#issuecomment-2447073088 , we should start from scratch with css and make it as simple as possible. 
The theme works fine if we utilise the default, therefore we really only need to implement the correct gammapy colours. 

- the secondary side bar is always shown on the right hand side of the screen (temporary fix:  #5582)
- allow for dark theme (#5520), this requires a completely separate dark theme (see [here](https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/light-dark.html))
- accordion issue (should be fixed by utilising the default. Temporary fix: #5526)

Astro-Kirsty:
This PR resolves #5586 and resolves #5520

It tries to make our css as small as possible such that we utilise the functionality of the theme more.
A major point is the support of the dark theme. 

There is a preview of the docs available with this PR.
If you go to the CI page, then Summary, there is a downloadable zip :)

And here: https://astro-kirsty.github.io/gammapy-docs-preview/

One remaining 'issue' is the formatting of the drop down menus in the HowTo. These are currently fully custom, but we should instead just rely on the drop downs which are available to us. 
The only difference is that we cannot have the button 'straight to tutorial' in the title of the drop down. 

You can find [here](https://github.com/user-attachments/files/19159793/gammapy_docs.pdf) some notes which discuss each element which has been removed or kept.


adonath:
Thanks @Astro-Kirsty this is important clean-up! Concerning the support for the dark theme: do you have any idea on how to support alternative plots? Currently all plots in our tutorials would use a white background. Maybe this can be fixed just by rendering them with a transparent background instead?

Astro-Kirsty:
I can look into it, but I don't think it will be possible as the plots all have black axes, which meaning setting some transparency would mean also needing to change them to white

registerrier:
> I can look into it, but I don't think it will be possible as the plots all have black axes, which meaning setting some transparency would mean also needing to change them to white

It seems that most pydata docs don't change the plots on dark mode. I don't think we should put much effort there.

adonath:
@registerrier @Astro-Kirsty It seems Matplotlib found an acceptable compromise:

- if you e.g. go to https://matplotlib.org/stable/tutorials/pyplot.html#sphx-glr-tutorials-pyplot-py
- switch to dark mode
- you should see that the plots actually dim a bit, the reduced contrast is much more pleasant to the eyes

I have not checked, but it seems this is the default for the dark theme anyway? It was not when I tried a few years back, the images were much more brighter.  In matplotlib the background of the gallery thumbnails is also adapted. This is achieved by https://github.com/matplotlib/matplotlib/blob/v3.10.0/doc/_static/mpl.css#L61, seem nice to me.

Astro-Kirsty:
For me, the figures look the same in that tutorial as what we have in this new PR. Maybe I misunderstood what you mean?
image

adonath:
Thanks @Astro-Kirsty, this is perfect! As I said, in previous versions of the dark pydata theme, the images kept their bright white appearance (compare the image background to the white of the GitHub text box), which was non-ideal. The new default is much better!

Maybe add the little change to adjust the bkg color of the gallery thumbnails and we should be good...

Astro-Kirsty:
Yes, the default for the gallery thumbnails looks good as well. 

There is actually a preview of the docs available with this PR. 
If you go to the CI page, then Summary, there is a downloadable zip :)

And here: https://astro-kirsty.github.io/gammapy-docs-preview/

morcuended:
Thanks @Astro-Kirsty, great job!

I briefly checked and had the problems with those broken icons in the user guide. Do you also see it like this?

image

Astro-Kirsty:
Thanks @morcuended, I did not see this locally nor on the [docs preview ](https://astro-kirsty.github.io/gammapy-docs-preview/user-guide/index.html)

morcuended:
I see it correctly in the downloadable html preview tough.

Another thing, the background in the drop downs boxes in the How To section, with dark mode, the does not change, so the white text is not visible. But I understand that this part is quite customized

image

I think we should try to adapt something more standard (e.g. https://sphinx-design.readthedocs.io/en/pydata-theme/dropdowns.html)

Astro-Kirsty:
Thanks @morcuended 
This is mentioned in the PR text and was also discussed at the last dev call. With the choice not yet decided whether to use the default or keep custom. (there is no point in making all the effort to change the custom code to allow for dark if we decide on the default)
I already have the changes, but I am making a separate PR for that, as this one is already large.

registerrier (approved):
Thanks @Astro-Kirsty . This looks really nice.

Given the evolutions of the default py-data theme, I fully support removing most of the additional code we introduced as it increases maintainability.

AtreyeeS (approved):
Thank you very much @Astro-Kirsty for this huge effort!
I propose we merge this soon and then make further changes if necessary.

PR #5718, opened 2025-03-11
Use default dropdown for HowTo

Astro-Kirsty:
This utilises the default dropdown instead of our custom design which we took from Pandas. 

Changing to the default allows for a cleaner looking dropdown and support for the dark theme without the need to alter. 

The css file will also need to be adapted to remove the code.

#5716 needs to be merged before this one

Examples: 
[Light theme](https://github.com/user-attachments/assets/60cb79dc-ba07-4928-adb1-5e251d00c1b3)
[Dark theme](https://github.com/user-attachments/assets/92cfa83c-2602-415c-a740-39ea16cec607)

### DIFF
--- a/docs/user-guide/howto.rst
+++ b/docs/user-guide/howto.rst
@@ -17,83 +17,86 @@ how to do everything in Gammapy.
 Please give feedback and suggest additions to this page!
 
 
-.. accordion-header::
-    :id: HowToGammapy
-    :title: Spell and pronounce Gammapy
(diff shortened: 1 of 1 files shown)


registerrier (approved):
Thanks @Astro-Kirsty . This is much simpler and still looks nice

bkhelifi (approved):
Thank you a lot, @Astro-Kirsty

PR #5719, opened 2025-03-12
Limit misusage of `to_region_map_dataset` 

QRemy:
`to_region_map_dataset` can only be applied if the mask
is spatially uniform within the region for each enegy bin
 (otherwise the exposure is not averaged over the same region than counts and background are summed)
This PR raises an error for that case.


examples/tutorials/analysis-3d/flux_profiles.py
```diff
@@ -107,11 +107,11 @@
 #
 
 regions = make_orthogonal_rectangle_sky_regions(
-    start_pos=SkyCoord("10d", "0d", frame="galactic"),
-    end_pos=SkyCoord("350d", "0d", frame="galactic"),
+    start_pos=SkyCoord("9d", "0d", frame="galactic"),
+    end_pos=SkyCoord("351d", "0d", frame="galactic"),
```
QRemy, line 111: The `FluxProfileEstimator` should not use `to_spectrum_dataset` but update the `mask_fit` the could be done with a context manager as in #5387 but more general such as it that takes a mask in input not only an energy range.

gammapy/datasets/map.py
```diff
@@ -1961,6 +1961,17 @@ def to_region_map_dataset(self, region, name=None):
         name = make_name(name)
         kwargs = {"gti": self.gti, "name": name, "meta_table": self.meta_table}
 
+        if self.mask and not self.mask.geom.is_region:
+            region_mask = self.mask.geom.region_mask(region)
+            for mask_image, region_mask_image in zip(
+                self.mask.iter_by_image(), region_mask.iter_by_image()
+            ):
+                if len(np.unique(mask_image.data[region_mask_image.data])) > 1:
```
registerrier, line 1969: why not checking `np.all(mask_image.data[region_mask_image.data])`

gammapy/datasets/map.py
```diff
@@ -1961,6 +1961,17 @@ def to_region_map_dataset(self, region, name=None):
...
+                    raise Exception(
+                        """`to_region_map_dataset` can only be applied if the mask
+                        is spatially uniform within the region for each enegy bin"""
```
registerrier, line 1972:
```suggestion
                        """`to_region_map_dataset` can only be applied if the mask
                        is spatially uniform within the region for each energy bin"""
```

gammapy/datasets/map.py
registerrier: A priori, masks are only booleans. If not I would even simply raise an error here. What is the meaning of the projection in this case becomes very unclear in my opinion.

examples/tutorials/analysis-3d/flux_profiles.py
registerrier: Really? 
But if you work with a 3D dataset, then you need a specific model for the flux you try to determine. So you could use a spatial model template following the region and fit it. If you do a joint fit then you could even determine the leakage from one region to the other. 
This could be a valid approach but this seems a bit different in terms of use case.

k-egg:
I think the mask is not yet actually applied to the exposure map at all. (See `gammapy/datasets/map.py`, line 1980/1991).

``` python
if self.exposure:
    kwargs["exposure"] = self.exposure.to_region_nd_map(region, func=np.mean)
```

A fitting mask for the exposure map could e.g. be created by copying `self.mask_image` into the correct exposure geom (now that non-uniform masks over the energy range have been forbidden).

gammapy/datasets/map.py
```diff
@@ -1961,6 +1961,17 @@ def to_region_map_dataset(self, region, name=None):
...
+                    )
+
         if self.mask_safe:
```
registerrier, line 1975: Can't you use this to perform the test above?

registerrier (commented):
Thanks @QRemy. I have left some inline comments.

An additional one is the partial overlap of the region with the geom. This is probably not covered by this test right?

gammapy/datasets/map.py
QRemy: mask_image combines all energy with a or condition but we need to check each energy bin in case the mask change with energy.
and I use np.unique instead of np.all because all False in one energy bin is also acceptable.

gammapy/datasets/map.py
```diff
@@ -1961,6 +1961,17 @@ def to_region_map_dataset(self, region, name=None):
         name = make_name(name)
         kwargs = {"gti": self.gti, "name": name, "meta_table": self.meta_table}
 
+        if self.mask and not self.mask.geom.is_region:
+            region_mask = self.mask.geom.region_mask(region)
```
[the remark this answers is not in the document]
QRemy: done

gammapy/datasets/map.py
QRemy: no it combines all energy with a or condition but we need to check each energy bin in case the mask change with energy.

QRemy:
> An additional one is the partial overlap of the region with the geom. This is probably not covered by this test right?

I added a check for that by creating a region mask from the counts geom padded with one bin and verifying that all edges are empty.

registerrier (approved):

PR #5720, opened 2025-03-12
Fix unknown arguments defined in Models

### LINKED ISSUE #5413: Named arguments aren't checked for some functions (leading to fat-fingering bugs)
**Gammapy version**
Both in version 0.19 and 1.2

**Bug description**

`PowerLawSpectralModel` has three named arguments: `index, amplitude, reference`.

If one of these is mistyped, then the default value is taken for the argument, and the mistyped argument is ignored.  

This leads to silent errors.

**Expected behavior**

Normally in Python, if a function has named arguments, then if a "misnomer" is given, then an error of this type occurs:
```
TypeError: call_by_named() got an unexpected keyword argument 'val3'
```
In this way there are no silent errors.

**To Reproduce**
Try this code:
```

model_simu = PowerLawSpectralModel(
        index=5.32,
        amplitude = u.Quantity("2.04e-8 TeV-1 cm-2 s-1"),
        reference = 80*u.GeV,)

bad_simu = PowerLawSpectralModel(
        index=5.32,
        amplitude = u.Quantity("2.04e-8 TeV-1 cm-2 s-1"),
        referecne = 80*u.GeV,)

energy_bounds = [0.01, 1.0] * u.TeV

from matplotlib import pyplot as plt
model_simu.plot(energy_bounds,sed_type="e2dnde",label="Good spectrum, reference")
bad_simu.plot(energy_bounds,sed_type="e2dnde",label="Misnomer spectrum, referecne")
plt.grid(which="both")
plt.plot(0.08,2.1e-10,'o',label="Good spectrum point")

plt.legend()
plt.show()
```

**Other information**
Not urgent, but this kind of unexpected and non-Pythonic behaviour can lead to loss of time.

Astro-Kirsty:
This fixes #5413

Currently a test fails `gammapy/modeling/models/tests/test_io.py` but I am not sure how to fix that 

`SpectralModels`:
- `ExpCutoffPowerLawNormSpectralModel` requires the kwargs on super because we have the deprecation there
- super init is required on the `EBLAbsorptionNormSpectralModel` -- because it needs to be able to 'read_builtin' --> this should be fine like this because we do not define a kwargs
- all other inits are empty 

SpatialModels:
- `TemplateSpatialModel` has the kwargs in super init (not sure it can be removed?)
- `PiecewiseNormSpatialModel` same as above?

### DIFF
--- a/gammapy/modeling/models/core.py
+++ b/gammapy/modeling/models/core.py
@@ -135,7 +135,7 @@ class ModelBase:
         default_parameters = self.default_parameters.copy()
 
         for key in kwargs.keys():
-            if key not in default_parameters.names:
+            if key != "covariance_data" and key not in default_parameters.names :
                 raise NameError(f"Unknown Parameter name '{key}'")
 
         for par in default_parameters:
--- a/gammapy/modeling/models/core.py
+++ b/gammapy/modeling/models/core.py
@@ -134,6 +134,10 @@ class ModelBase:
         # Copy default parameters from the class to the instance
         default_parameters = self.default_parameters.copy()
 
+        for key in kwargs.keys():
+            if key not in default_parameters.names:
+                raise NameError(f"Unknown Parameter name '{key}'")
+
         for par in default_parameters:
             value = kwargs.get(par.name, par)
 
--- a/gammapy/modeling/models/tests/test_temporal.py
+++ b/gammapy/modeling/models/tests/test_temporal.py
@@ -387,7 +387,7 @@ def test_with_skymodel(light_curve):
 
 def test_plot_constant_model():
     time_range = [Time.now(), Time.now() + 1 * u.d]
-    constant_model = ConstantTemporalModel(const=1)
+    constant_model = ConstantTemporalModel()
     with mpl_plot_check():
         constant_model.plot(time_range)
 
--- a/gammapy/modeling/models/tests/test_temporal.py
+++ b/gammapy/modeling/models/tests/test_temporal.py
@@ -387,7 +387,7 @@ def test_with_skymodel(light_curve):
 
 def test_plot_constant_model():
     time_range = [Time.now(), Time.now() + 1 * u.d]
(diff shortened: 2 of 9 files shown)


gammapy/modeling/models/core.py
```diff
@@ -134,6 +134,10 @@ def __init__(self, **kwargs):
         # Copy default parameters from the class to the instance
         default_parameters = self.default_parameters.copy()
 
+        for key in kwargs.keys():
+            if key not in default_parameters.names:
```
QRemy, line 138:
```suggestion
            if key != "covariance_data" and key not in default_parameters.names :
```

Astro-Kirsty: I dont think this description can work

gammapy/modeling/models/tests/test_temporal.py
```diff
@@ -387,7 +387,7 @@ def test_with_skymodel(light_curve):
 
 def test_plot_constant_model():
     time_range = [Time.now(), Time.now() + 1 * u.d]
-    constant_model = ConstantTemporalModel(const=1)
+    constant_model = ConstantTemporalModel()
```
Astro-Kirsty, line 390: So the `ConstantTemporalModel` is just the default, it cannot actually have some value applied to it?

registerrier: It is still possible here, but it needs a default parameter value

QRemy: yes it seems so

registerrier (approved):
Thanks @Astro-Kirsty ! This looks good. 

In particular, it shows that we had undetected issues in the tests and the docs. So a solution to the issue is really important. You should just remove the argument in the `ConstantTemporalModel` gallery and this is fine.

QRemy:
Thanks @Astro-Kirsty  all tests are good now, merging.

PR #5726, opened 2025-03-14
[PIG] Introduce Plotter classes

registerrier:
**Description**

This pull request proposes a PIG for plotting code.

**Dear reviewer**


bkhelifi:
These classes should rely only on our internal data model, our internal data classes. In this way, whatever the format of the read data, the functions will work well... This is just a small caution to take.

AtreyeeS:
Thank you @registerrier ! I think it will be nice to have this before the next release. Later, we might think if `xx.plot` is necessary to maintain or the users can go through `xx.plotter.plot`.

AtreyeeS (dismissed):
Thank you @registerrier ! I think it will be nice to have this before the next release. Later, we might think if `xx.plot` is necessary to maintain or the users can go through `xx.plotter.plot`.

AtreyeeS:
Suggestion from @adonath during the dev call: Have stateless classes as in eg: https://github.com/gammapy/gammapy/blob/main/gammapy/visualization/panel.py as this allows to plot multiple maps with same config by reusing a plotter object, instead of creating a new plotter class for each of these plots.

docs/development/pigs/pig-029.rst
```diff
@@ -0,0 +1,171 @@
...
+
+A more modular approach, using dedicated `Plotter` classes, would resolve these issues while preserving all current plotting functionalities.
+
+Note that SunPy `ndcube` implements as a similar approach `MatplotlibPlotter`_
+
+Proposed Changes
+================
+
+- Introduce separate `Plotter` classes for different Gammapy objects (e.g., `MapDatasetPlotter`, `SpectrumDatasetPlotter`).
+  - A `Plotter` would take the object to plot on init and reimplement its plot methods.
```
bkhelifi: I  hope by avoiding a copy in order to not overload the RAM...

docs/development/pigs/pig-029.rst
```diff
@@ -0,0 +1,171 @@
...
+  - The `Plotter` would be an attribute of the parent object.
+  - It could be lazily instantiated to import matplotlib only when needed.
+  - Existing plot methods will silently call the `self.plotter` to ensure backward compatibility
```
bkhelifi: self=?, I do not understand...

docs/development/pigs/pig-029.rst
```diff
@@ -0,0 +1,171 @@
...
+        def plot_mask(ax, mask, **kwargs):
+            if mask is not None:
+                mask.plot_mask(ax=ax, **kwargs)
+
+        fig, axes = plt.subplots(
+            ncols=2,
+            nrows=2,
+            subplot_kw={"projection": self._geom.wcs},
+            figsize=figsize,
+            gridspec_kw={"hspace": 0.25, "wspace": 0.1},
```
bkhelifi: This option can lead to superposition of the cbar with the adjacent plot. I won't recommend that!

docs/development/pigs/pig-029.rst
```diff
@@ -0,0 +1,171 @@
...
+    def peek(self,  figsize=(12, 8)):
+        """Quick-look summary plots.
+
+        Parameters
+        ----------
+        figsize : tuple
+            Size of the figure. Default is (12, 10).
+
+        """
+
```
bkhelifi, line 84: Add the cbar in this example

docs/development/pigs/pig-029.rst
```diff
@@ -0,0 +1,171 @@
...
+================
+
+- Introduce separate `Plotter` classes for different Gammapy objects (e.g., `MapDatasetPlotter`, `SpectrumDatasetPlotter`).
+  - A `Plotter` would take the object to plot on init and reimplement its plot methods.
+  - The `Plotter` would be an attribute of the parent object.
+  - It could be lazily instantiated to import matplotlib only when needed.
+  - Existing plot methods will silently call the `self.plotter` to ensure backward compatibility
+- Move existing plotting methods from core objects to these new `Plotter` classes. Move corresponding tests as well.
+  - Existing plotting functionalities will remain accessible but may be deprecated in future releases.
+
```
bkhelifi, line 59: Maybe, add a generic way to add cbar with the following requirements:
- when added, creation of extra margin between plots
- be able to add a configurable title of the cbar
- limit the cbar length to the plot height

docs/development/pigs/pig-029.rst
```diff
@@ -0,0 +1,171 @@
...
+        def peek(self, figsize=(12, 8)):
+            self.plotter.peek(figsize)
+
+An example for `gammapy.data` objects is proposed in the PR `#5725`_.
+
+The property getter/setter mechanism could use a `PlotMixin` approach to avoid repeating the boilerplate code.
+
+If a user wants to modify `MapDataset.peek()` to provide more suited visualization of e.g. Fermi data (`#5570`_), they could
+create a class inheriting from `MapDatasetPlotter` anv overload its `peek` method. Similarly, different plotting
```
bkhelifi:
```suggestion
create a class inheriting from `MapDatasetPlotter` and overload its `peek` method. Similarly, different plotting
```

docs/development/pigs/pig-029.rst
```diff
@@ -0,0 +1,171 @@
...
+
+A general drawback of this approach is the large number of plotter classes necessary. Yet given the specialization of plot functions this seems
+unavoidable.
+
+Other strategies can be envisioned:
+- Keep the current approach: This would maintain the status quo but does not address the issues outlined.
+- Use mixin classes: This could modularize the code but still tightly couples plotting with core objects.
+  We have this already with `PlotMixin` that is used by `SpectrumDataset`. But this does not allow to have you own plotters
+  nor to have delayed imports.
+- Define standalone plotting functions instead of classes: This would reduce complexity but may lead to less structured and reusable code
```
bkhelifi, line 139: What is reusable with the proposed scheme? It seems that not a lot... If there is almost none, do we really need the overcomplexity of classes?

docs/development/pigs/pig-029.rst
```diff
@@ -0,0 +1,171 @@
...
+  and probably less maintainable.
```
bkhelifi, line 140: I can't understand this argument. A function is as maintainable as a class!

docs/development/pigs/pig-029.rst
registerrier: yes, and making them stateless, as proposed by @adonath , clarifies this more. They should contain only the plots configuration.

docs/development/pigs/pig-029.rst
registerrier: this is how it is done in the current code and independent of the proposal here.

docs/development/pigs/pig-029.rst
registerrier: The example is just a copy and paste of the `MapDataset.peek()`

docs/development/pigs/pig-029.rst
registerrier: In what is shown here, relatively little indeed. The point is actually that it allows to configure plots on the `Plotters.__init__` and could alleviate the issue of e.g. passing many arguments to various matplotlib functions inside a single function call.
We can rewrite a bit to make this clearer.

docs/development/pigs/pig-029.rst
registerrier: A priori, yes. But plot functions are complex to configure and require lots of arguments (see e.g. the collection of different kwargs). This is difficult to understand for users and developers as this becomes a bit unpredictable. The interest of classes would be that they can serve as somewhat simpler configuration interfaces for the plots.

docs/development/pigs/pig-029.rst
```diff
@@ -0,0 +1,232 @@
+.. include:: ../../references.txt
+
+.. _pig-029:
+
+*************************************************************
+PIG 29 - Separation of Plotting Methods using Plotter classes
```
AtreyeeS:
```suggestion
PIG 30 - Separation of Plotting Methods using Plotter classes
```

AtreyeeS:
This is ready for review now

MRegeard:
I support this 100% !

Currently plot methods completely bloat core classes. 
The plot methods that we currently have are relatively simple, i.e. they usually access data at 1 or 2 layer down their attached class (`spectrum_dataset.counts.data`) and have default (ugly) `matplotlib` output while taking a lot of code space. 
Moreover, they are usually not useful to the class itself but only for the user. And a core class should not be "broken" because of a plotting function !

Having a plotter interface would make the core classes much cleaner. It would also allow for more customisation (imagine having a gammapy configuration file that provides your favorite colormap, fontsize, figsize, unit for energy axes etc.
This would also open Gammapy to other plotting library (`plotly`, `bokeh`) as plotting backend or even user define plotting library. 

Implementation detail: The [protocol](https://typing.python.org/en/latest/spec/protocol.html) typing could be use here.

MRegeard:
I would be more than happy to help with the implementation !

AtreyeeS:
This PIG is accepted and will be introduced slowly.

AtreyeeS (approved):
Thanks ! Merging now

PR #5727, opened 2025-03-14
Fix Models.read to support relative template path from anywhere

### LINKED ISSUE #5682: Template models are not read correctly
Template models are not read correctly if filename attributes of the model is not absolute.

When you write a model with a template, you need to indicate a filename for the FITS file that will store the template. So you will usually write `filename = "my_template.fits"`.  

Then when you want to read the model, the yaml file will only contains `filename = "my_template.fits"`. So if, you read from another directory, it will try to open `"./my_template.fits"`. This is because the path is not passed to the different `from_dict` methods that create the model. 

To fix that, we can pass the path to the filename as it is done for the covariance. 

To reproduce:

```python
from gammapy.modeling.models import PowerLawSpectralModel, TemplateSpatialModel, SkyModel, Models
from gammapy.maps import Map
import numpy as np
import os

map_ = Map.create(width=(10, 10))
map_.data = np.ones_like(map_.data)

spatial = TemplateSpatialModel(map=map_, filename="spatial_template.fits")
spectral = PowerLawSpectralModel()
sky_model = SkyModel(spectral_model=spectral, spatial_model=spatial, name='test')
m = Models([sky_model])
m.write("test_template_readingyaml")

os.chdir("../")
Models.read("<MODEL_DIR>/test_template_readingyaml")
```

QRemy:
Resolves #5682

### DIFF
--- a/gammapy/modeling/models/core.py
+++ b/gammapy/modeling/models/core.py
@@ -20,6 +20,28 @@ __all__ = ["Model", "Models", "DatasetModels", "ModelBase"]
 log = logging.getLogger(__name__)
 
 
+def _recursive_dict_filename_update(d, path):
+    """update model filename to full path if exits"""
+    for k, v in d.items():
+        if isinstance(v, dict):
+            _recursive_dict_filename_update(v, path)
+        elif k == "filename":
+            filename = d[k]
+            if (path / filename).exists():
+                d[k] = path / filename
+
+
+def _recursive_model_filename_update(model, path):
+    """update model filename to relative path if child of path"""
+    if hasattr(model, "filename") and path == make_path(model.filename).parent:
+        _, filename = split(model.filename)
+        model.filename = filename
+
+    if hasattr(model, "_models"):
+        for m in model._models:
+            _recursive_model_filename_update(m, path)
+
+
 def _set_link(shared_register, model):
     for param in model.parameters:
         name = param.name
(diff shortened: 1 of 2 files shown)


gammapy/modeling/models/core.py
```diff
@@ -20,6 +20,28 @@
 log = logging.getLogger(__name__)
 
 
+def _recursive_dict_filename_update(d, path):
```
bkhelifi, line 23: I am wondering whether these 2 functions should not be into the utils directory...

gammapy/modeling/models/core.py
```diff
@@ -20,6 +20,28 @@
...
+    """update model filename to full path if exits"""
+    for k, v in d.items():
+        if isinstance(v, dict):
+            _recursive_dict_filename_update(v, path)
+        elif k == "filename":
+            filename = d[k]
+            if (path / filename).exists():
+                d[k] = path / filename
```
QRemy, line 31:
```suggestion
    for key, value in d.items():
        if isinstance(value, dict):
            _recursive_dict_filename_update(value, path)
        elif key == "filename":
            filename = d[key]
            if (path / filename).exists():
                d[key] = path / filename
```

gammapy/modeling/models/core.py
QRemy, line 23:
```suggestion
def _recursive_dict_filename_update(dict_, path):
```

MRegeard (approved):
Thanks @QRemy. It seems good to me.

PR #5731, opened 2025-03-17
fix: install sherpa using conda instead of pip in dev env

### LINKED ISSUE #5730: Creating the dev environment is failing because of sherpa
**Gammapy version**
Gammapy main branch.

**Bug description**
Creating the Gammapy dev environment is failing due to `sherpa` dependency:
```
Failed to build sherpa
ERROR: Failed to build installable wheels for some pyproject.toml based projects (sherpa)
```

The problem seems to be similar to this issue: https://github.com/sherpa/sherpa/issues/1528

**Fix**
Installing sherpa using conda fix the issue.

**Expected behavior**
No error expected.

**To Reproduce**

**Other information**

natthan-pigoux:
**Description**

Creating the Gammapy dev environment was failing due to sherpa package installed using pip. 

Following: https://sherpa.readthedocs.io/en/latest/install.html, I moved the sherpa dependency in the conda channel.

Closes #5730

**Dear reviewer**

Maybe there is a good reason to install sherpa using pip instead of conda 'm not aware of.

### DIFF
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -9,6 +9,7 @@ name: gammapy-dev
 
 channels:
   - conda-forge
+  - https://cxc.cfa.harvard.edu/conda/sherpa # c.f. https://sherpa.readthedocs.io/en/latest/install.html
 
 variables:
   PYTHONNOUSERSITE: "1"
@@ -59,6 +60,7 @@ dependencies:
   - pydocstyle
   - pylint
   - setuptools_scm
+  - sherpa>=4.17
   - sphinx
   - sphinx-astropy
   - sphinx-click
@@ -79,7 +81,6 @@ dependencies:
   - pyinstrument
   - memray
   - pip:
-      - sherpa>=4.17
       - pytest-sphinx
       - ray[default]>=2.9
       - PyGithub


bkhelifi:
Hello Natthan,
This was not working on ubuntu, couple of months ago. One should check that this works on the 3 platforms

natthan-pigoux:
> Hello Natthan, This was not working on ubuntu, couple of months ago. One should check that this works on the 3 platforms

Hello Bruno, thank you for the info. I will comment when tested.

registerrier (commented):
Thanks @natthan-pigoux . 

This comes after a series of changes regarding sherpa :
- #5610 (sherpa removed from yaml installation file because of the silicon mac issues)
- #5479 (forces >=4.17 after transition to numpy 2)
- #5243  (difficulty in building env for CI with sherpa and ray on python 3.11)
- #4983 (moved from conda to pip installation because of MacOS silicon)

So it seems that the problem here can be connected with the problem that appeared in #5243 
While the proposed change might work now on most platforms, I don't know how long it will take before we need to change again :) . 

I wonder if we shouldn't have a scheduled action to test environment-dev conda installation on various platforms. And in general, a clean-up of the file might be necessary.

AtreyeeS:
Do we need to support the sherpa-backend beyond 2.0 ?  I dont know if it really useful,  and the installation is a constant issue. Maybe we can have this as an external library and have a recipe if necessary?

Astro-Kirsty:
This worked for me on `Ubuntu 24.04.2 LTS`

registerrier (approved):
Thanks @natthan-pigoux . This is good. We should go for a more regular testing of the dev environment on the different platforms.

PR #5733, opened 2025-03-18
Remove the unneeded css code for the dropdowns

Astro-Kirsty:
This removes the unneeded css code (the HowTo PR was merged before this was added #5718)


AtreyeeS (approved):
Thanks @Astro-Kirsty !

Maybe it would be useful if you add some info in the dev docs for future reference (separate PR)

bkhelifi (approved):
Thanks for the clean up.

registerrier (approved):
Thanks @Astro-Kirsty . The results look nice and the dedicated css is no longer required.

PR #5736, opened 2025-03-21
Change documentation of `ConstantTemporalModel`

### LINKED ISSUE #5721: Should we allow for a constant parameter in the `ConstantTemporalModel`?
For the other constant models ([ConstantSpectralModel](https://docs.gammapy.org/dev/user-guide/model-gallery/spectral/plot_constant_spectral.html#constant-spectral-model), [ConstantSpatialModel](https://docs.gammapy.org/dev/user-guide/model-gallery/spatial/plot_constant.html#constant-spatial-model)) we have a `const` parameter which is implemented. However, for the ConstantTemporalModel it is not implemented correctly, but our [documentation](https://docs.gammapy.org/dev/user-guide/model-gallery/temporal/plot_constant_temporal.html#constant-temporal-model) implies it is. 

The question is whether we should implement the `const` value properly or if we should leave it as it, and change our documentation?

Astro-Kirsty:
Resolve #5721

### DIFF
--- a/examples/models/temporal/plot_constant_temporal.py
+++ b/examples/models/temporal/plot_constant_temporal.py
@@ -6,7 +6,7 @@ Constant temporal model
 
 This model parametrises a constant time model.
 
-.. math:: F(t) = k
+.. math:: F(t) = 1
 
 """
 


registerrier (approved):
Thanks @Astro-Kirsty

PR #5740, opened 2025-03-27
Apply mask in MapDataset.plot_residuals_spectral

QRemy:
Apply mask in `MapDataset.plot_residuals_spectral` to be consistent with  `MapDataset.plot_residuals_spatial`.


bkhelifi (approved):
Well seen!

PR #5743, opened 2025-03-28
Helper method for observation selection at Datastore level

QRemy:
Helper method for observation selection at Datastore level
Follow-up of #5687
Requires https://github.com/gammapy/gammapy-data/pull/67

### DIFF
--- a/gammapy/data/data_store.py
+++ b/gammapy/data/data_store.py
@@ -392,7 +392,7 @@ class DataStore:
         require_events : bool, optional
             Require events and gti table or not. Default is True.
         selection : `~numpy.ndarray`, optional
-            Boolean array of the same lenght than the ``obs_table``.
+            Boolean array of the same length than the ``obs_table``.
             Observation is kept if  corresponding selection mask element is True
             and if it is in the list of obs_id.
             If None, default is all observations ordered by OBS_ID are returned.
@@ -434,7 +434,15 @@ class DataStore:
         log.info(f"Observations selected: {len(obs_list)} out of {len(obs_id)}.")
         return Observations(obs_list)
 
-    def get_observation_groups(self, key, **kwargs):
+    def get_observation_groups(
+        self,
+        key,
+        obs_id=None,
+        skip_missing=False,
+        required_irf="full-enclosure",
+        require_events=True,
+        selection=None,
+    ):
         """Generate groups of `~gammapy.data.Observations` with a shared property.
 
         Parameters
@@ -442,9 +450,41 @@ class DataStore:
         key : str
             Key of the observation table used to apply the grouping.
             For example, "EVENT_TYPE" will return group observations
-            with the same event type.
-        kwargs : dict, optional
-            Keyword arguments passed to `~gammapy.data.DataStore.get_observations`.
+        obs_id : list, optional
+            Observation IDs.
+            If None, default is all observations ordered by OBS_ID are returned.
+            This is not necessarily the order in the ``obs_table``.
+        skip_missing : bool, optional
+            Skip missing observations. Default is False.
+        required_irf : list of str or str, optional
+            Runs will be added to the list of observations only if the
+            required HDUs are present. Otherwise, the given run will be skipped
+            The list can include the following options:
+
+            * `"events"` : Events
+            * `"gti"` :  Good time intervals
+            * `"aeff"` : Effective area
+            * `"bkg"` : Background
+            * `"edisp"` : Energy dispersion
+            * `"psf"` : Point Spread Function
+            * `"rad_max"` : Maximal radius
+
+            Alternatively single string can be used as shortcut:
+
+            * `"full-enclosure"` : includes `["events", "gti", "aeff", "edisp", "psf", "bkg"]`
+            * `"point-like"` : includes `["events", "gti", "aeff", "edisp"]`
+            * `"all-optional"` : no HDUs are required, only warnings will be emitted
+              for missing HDUs among all possibilities.
+
+            Default is `"full-enclosure"`.
+        require_events : bool, optional
+            Require events and gti table or not. Default is True.
+        selection : `~numpy.ndarray`, optional
+            Boolean array of the same length than the ``obs_table``.
+            Observation is kept if  corresponding selection mask element is True
+            and if it is in the list of obs_id.
+            If None, default is all observations ordered by OBS_ID are returned.
+            Default is None.
 
         Returns
         -------
@@ -467,20 +507,21 @@ class DataStore:
                 "obs_table attribute must not be None to select groups of observations"
             )
 
-        observations = self.get_observations(**kwargs)
-        obs_table = self.obs_table[
-            [
-                np.where(self.obs_table["OBS_ID"].astype(str) == _)[0]
-                for _ in observations.ids
-            ]
-        ]
-        observations_group = observations.group_by_label(
-            obs_table[key].astype(str).squeeze()
-        )
-        return {
-            f"{key}{old_key[5:]}": value
-            for old_key, value in observations_group.items()
-        }
+        if obs_id is not None:
+            grouped = self.obs_table.select_obs_id(obs_id).group_by(key)
+        else:
+            grouped = self.obs_table.group_by(key)
+
+        result = {}
+        for key_value, group in zip(grouped.groups.keys, grouped.groups):
+            result[f"{key}_{key_value[0]}"] = self.get_observations(
+                group["OBS_ID"],
+                skip_missing=skip_missing,
+                required_irf=required_irf,
+                require_events=require_events,
+                selection=selection,
+            )
+        return result
 
     def copy_obs(self, obs_id, outdir, hdu_class=None, verbose=False, overwrite=False):
         """Create a new `~gammapy.data.DataStore` containing a subset of observations.
--- a/gammapy/data/data_store.py
+++ b/gammapy/data/data_store.py
@@ -80,7 +80,10 @@ class DataStore:
 
     def __init__(self, hdu_table=None, obs_table=None):
         self.hdu_table = hdu_table
-        self.obs_table = table.unique(obs_table, keys="OBS_ID")
+        if obs_table is not None:
+            self.obs_table = table.unique(obs_table, keys="OBS_ID")
+        else:
+            self.obs_table = None
 
     def __str__(self):
         return self.info(show=False)
--- a/gammapy/data/data_store.py
+++ b/gammapy/data/data_store.py
@@ -450,6 +450,16 @@ class DataStore:
         -------
         groups : dict of `~gammapy.data.Observations`
             Dictionary of Observations instance, one instance for each group.
+
+
+        Examples
+        --------
+        Create groups of observation with the same event type.
+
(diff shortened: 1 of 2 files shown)


gammapy/makers/tests/test_map.py
```diff
@@ -580,16 +580,15 @@ def test_map_dataset_maker_swgo():
...
-    datastore_redu = DataStore(
-        hdu_table=datastore.hdu_table[hdu_selection],
-        obs_table=datastore.obs_table[obs_selection],
-    )
-    observations = datastore_redu.get_observations()
+    datastore_redu = datastore.select(dict(EVENT_TYPE=event_type))
+    observations_redu = datastore_redu.get_observations()
 
-    obs = observations[0]
+    assert observations_redu.obs_ids == observation_groups["event_type"].obs_ids
```
QRemy, line 589:
```suggestion
    assert observations_redu.obs_ids == observation_groups[event_type].obs_ids
```

gammapy/data/data_store.py
```diff
@@ -412,6 +474,29 @@ def get_observations(
...
+        groups : dict of `~gammapy.data.Observations`
+            Dictionary of Observations instance, one instance for each group.
+        """
+
+        observations = self.get_observations()
+        observations_group = observations.group_by_label(self.obs_table[key])
+        reformated = dict()
+        for old_key, value in observations_group.items():
+            reformated[f"{key}{old_key[5:]}"] = value
+        return reformated
```
VascoSch92, line 498:
```suggestion
       return {
           f"{key}{old_key[5:]}":  value
           for old_key, value in observations_group.items()
       } 
```
should be faster than an explicit `for-loop`

QRemy:
based on discussion #5780, I simplified this to the strict minimum keeping only `get_observation_groups ` and adding a `selection_mask` to `get_observations`

QRemy:
This would need a review
(the SWGO tests will fail until this is we merged as the format changed to be closer to what we discussed for VODF) @bkhelifi, @registerrier, @LauraOlivera

bkhelifi:
Should we wait for the merge of #5855 to update and review this PR?

QRemy:
> Should we wait for the merge of #5855 to update and review this PR?

Not necessarily, the selection option introduced here could be extended later in order to also accept a string and pass it to the logic parser from #5855 (that would be only one more condition in the first if statement of get_observations).

registerrier (commented):
Thanks @QRemy . A general comment for now.

I find the function does too many things and the scope of the function possibly too narrow (e.g. extracting observations and grouping per event_type). 

The grouping of observations could be performed outside the function. 

In general, we should reflect a bit more on the handling of our data collections using operators like `group_by` or `filter`, or with operators behaving like them. These are usually very powerful and versatile with their function argument.

If in the end you do grouping on the observations after creating the full list,  you could have a function such as:
```
def group_by(observations, key):
    from itertools import groupby
    groups = groupby(observations, key)
    result = {}
    for name, obslist in groups:
        if name in result.keys():
            result[name] += Observations(obslist)
        else:
            result[name] = Observations(obslist)
    return result
```
and use it like this:
```
>>> groups = group_by(observations, lambda obs : obs.meta.target.name)
>>> print(group["Crab Nebula"])

Observations
Number of observations: 4
Observation

	obs id            : 23523 
 	tstart            : 53343.92
	tstop             : 53343.94
	duration          : 1687.00 s
	pointing (icrs)   : 83.6 deg, 21.5 deg

	deadtime fraction : 6.2%
```
or 
```
>>> group_by(observations, lambda obs : round(obs.meta.time_info.time_start.byear))

{2004: <gammapy.data.observations.Observations at 0x159bc9850>,
 2005: <gammapy.data.observations.Observations at 0x15f163790>,
 2006: <gammapy.data.observations.Observations at 0x15f14ee50>,
 2007: <gammapy.data.observations.Observations at 0x15f14ec90>,
 2009: <gammapy.data.observations.Observations at 0x15b8490d0>}
``` 
And you can already do something like:
```
def pnt_separation(target, distance=5*u.deg):
    return  lambda _ : _.meta.pointing.radec_mean.separation(position) < distance
filtered = Observations([_ for _ in filter(pnt_separation(crab_pos), observations)])
```

gammapy/data/data_store.py
```diff
@@ -80,7 +81,10 @@ class DataStore:
 
     def __init__(self, hdu_table=None, obs_table=None):
         self.hdu_table = hdu_table
-        self.obs_table = obs_table
+        if obs_table is not None:
+            self.obs_table = table.unique(obs_table, keys="OBS_ID")
```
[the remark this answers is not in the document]
QRemy: it is specified in the format https://gamma-astro-data-formats.readthedocs.io/en/v0.3/data_storage/obs_index/index.html?highlight=obs_id#required-columns
OBS_ID  could be eventually added to the glossary in a separate PR giving the link to the definition in GADF and VODF.

gammapy/data/data_store.py
```diff
@@ -416,6 +434,44 @@ def get_observations(
...
+            For example "EVENT_TYPE" will return group observations
+            with the same event type.
+        kwargs : dict, optional
+            Keyword arguments passed to `~gammapy.data.DataStore.get_observations`.
+
+        Returns
+        -------
+        groups : dict of `~gammapy.data.Observations`
+            Dictionary of Observations instance, one instance for each group.
+        """
```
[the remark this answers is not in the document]
QRemy: Added a SWGO example with event types.

LauraOlivera (approved):
Hi Quentin, this looks good (and very useful) to me.

The only question I have is how will this be documented: is the plan to make a SWGO/event types tutorial?

bkhelifi:
Hi @LauraOlivera,
This function is useful for any instrument. And as you mention, a new technical tutorial might be great...

LauraOlivera:
indeed - I just mentioned SWGO explicitly thinking about examples in $GAMMAPY_DATA that use `EVENT_TYPE` keyword. But if I remember correctly, the HAWC data also has it, although I am not sure if in the right table...

gammapy/data/data_store.py
```diff
@@ -387,26 +392,40 @@ def get_observations(
             Default is `"full-enclosure"`.
         require_events : bool, optional
             Require events and gti table or not. Default is True.
+        selection : `~numpy.ndarray`, optional
+            Boolean array of the same lenght than the ``obs_table``.
```
registerrier:
```suggestion
            Boolean array of the same length than the ``obs_table``.
```

gammapy/data/data_store.py
```diff
@@ -387,26 +392,40 @@ def get_observations(
...
+        if selection is None:
+            obs_id_selection = self.obs_ids
+        else:
+            obs_id_selection = np.array(self.obs_ids)[selection]
+
         if obs_id is None:
-            obs_id = self.obs_ids
+            obs_id = obs_id_selection
+        else:
+            for _ in obs_id:
```
registerrier, line 415: Since all obs_id are unique, this could be handled simply with a `set` rather than a `list`

gammapy/data/data_store.py
```diff
@@ -387,26 +392,40 @@ def get_observations(
...
         Returns
         -------
         observations : `~gammapy.data.Observations`
             Container holding a list of `~gammapy.data.Observation`.
         """
+
+        if selection is None:
+            obs_id_selection = self.obs_ids
+        else:
+            obs_id_selection = np.array(self.obs_ids)[selection]
```
registerrier: I am not sure I understand the actual gain here.  

The current logic is to perform the obs_id selection directly from the `ObservationTable` and this requires the same thing as the `selection` mask to be built from the `obs_table`.

registerrier: How do you make sure the `selection` mask is correctly ordered w.r.t. the `obs_table`

gammapy/data/data_store.py
```diff
@@ -416,6 +435,54 @@ def get_observations(
...
+        >>> observation_groups = datastore.get_observation_groups("EVENT_TYPE")
+        """
+
+        if self.obs_table is None:
+            raise ValueError(
+                "obs_table attribute must not be None to select groups of observations"
+            )
+
+        observations = self.get_observations(**kwargs)
+        obs_table = self.obs_table[
```
registerrier: This can be simplified using directly `ObservationTable.select_obs_id([_ for _ in observations.ids])`.

gammapy/data/data_store.py
```diff
@@ -416,6 +435,54 @@ def get_observations(
         log.info(f"Observations selected: {len(obs_list)} out of {len(obs_id)}.")
         return Observations(obs_list)
 
+    def get_observation_groups(self, key, **kwargs):
```
registerrier: Here I would keep the explicit arguments.

gammapy/data/data_store.py
```diff
@@ -416,6 +435,54 @@ def get_observations(
...
+        >>> datastore = DataStore.from_dir(path, "hdu-index.fits.gz", "obs-index.fits.gz")
+        >>> observation_groups = datastore.get_observation_groups("EVENT_TYPE")
+        """
+
+        if self.obs_table is None:
+            raise ValueError(
+                "obs_table attribute must not be None to select groups of observations"
+            )
+
+        observations = self.get_observations(**kwargs)
```
registerrier: this seems a bit strange to extract observations first, then build a mask to filter the `ObservationTable` to then apply the `Observations.group_by_label`. 

You can already use `ObservationTable.group_by(key)` and then loop over `ObservationTable.groups` to get the observations. e.g.

```
if obs_ids is not None:
    grouped = self.select_obs_id(obs_ids).group_by(key)
else:
    grouped = self.obs_table.group_by(key)

result = {}
for key, group in zip(grouped.groups.keys, grouped.groups):
    result[key[0]]= self.get_observations(group["OBS_ID"], **kwargs)

```

registerrier (commented):
Thanks @QRemy . I have left a few specific commentsinline. 

In general:
- I am not sure we gain much with the addition of a `selection` mask on `get_observations`. Is it safe w.r.t. ordering for instance?
- I propose a slightly different way to perform the grouping to avoid getting all observations and then re-creating an observation table, but rather using the `Table.group_by` functionality.

gammapy/data/data_store.py
QRemy: now the `obs_table` is forced to be ordered and unique on init sur the order will be the same as `self.obs_ids`
https://github.com/gammapy/gammapy/blob/dd2b64d77f28a523e753f3d3d1635291313ac71a/gammapy/data/data_store.py#L85

gammapy/data/data_store.py
QRemy: done

gammapy/data/data_store.py
QRemy: we still need the list comprehension to preserve the current behaviour that returns the observations in the order given by `obs_id` but `obs_id_selection` can be a set so the lookup is faster.
`selection_set = set(obs_id_selection)`
`obs_id_selection = [_ for _ in obs_id if _ in selection_set]`

registerrier (approved):
Thanks @QRemy. Looks good!

PR #5748, opened 2025-04-03
Replace cython code by jitted code

QRemy:
This PR replace the cython code in `fit_statistics_cython` by  jitted code in `fit_statistics_jit`.
I did not run the benchmarks yet but the performance seems to improve using jit by about 30% for the tests in `datasets` and  10% for the TSmap tests in `estimators/map/test_ts.py`. 
The option nogil=True seems to be responsible to about half of the improvement.

Caveat : requires to impose numpy<2.2

Please try it to see if it does not degrade the performances for your use cases.


QRemy:
> When you say ` about 30% for the tests in datasets`. Is this new numba vs old cython ?
> So the new code is even faster than the Cython code ?

yes from what I tested so far the new code using numba seems faster.

gammapy/stats/fit_statistics_jit.py
```diff
@@ -0,0 +1,145 @@
...
+            lognpr = logtrunc
+
+        stat_sum += npr
+        if counts[i] > 0.0:
+            stat_sum -= counts[i] * lognpr
+
+    return 2 * stat_sum
+
+
+@jit(nopython=True, parallel=True, nogil=True, cache=True)
```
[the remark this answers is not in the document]
facero: I think it'll be better to have them all to False for 1) consistency and 2) if you're running gammapy on a cluster it might be better that each process is mono-core
(also true for Ultranest MPI btw)

facero:
Useful piece of info for the review of this PR: 
ctapipe also uses numba (for a while I think) and so removing cython and adding numba aligns with ctapipe dependencies.

Concerning the numpy < 2.2
This issues mentions that this should be fixed in a matter of weeks:
https://github.com/numba/numba/issues/9838

gammapy/stats/fit_statistics_jit.py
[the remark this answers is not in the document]
QRemy: done

QRemy:
I had to force the array type to float for the CI to pass on all systems.
By the way the tests seem to be set for float64 but I thought our default was float32 ?

facero:
Removing Cython makes gammapy a pure python package right ?
I imagine that has a lot of benefits for installation, distributions, and CI.
For the dev version for example the `pip install -e` would no longer be necessary I think.

For conda distribution this will reduce the variety of wheels we have to distribute.
This may also impact CI in a positive way (no platform test ?)

Maybe we should make a list of cascading impact of this change.

bkhelifi:
Hi all,
I am wondering whether it is wise today to remove completely the current cpython code, as we have not yet a view on the longterm stability or compatibility with numpy or etc... This is an open question that I would like to raise

QRemy:
TS cython
Screenshot 2025-04-04 at 11 00 06

TS jit
Screenshot 2025-04-04 at 12 31 23

3d-joint fit cython
Screenshot 2025-04-04 at 10 58 54

3d-joint fit jit
Screenshot 2025-04-04 at 10 58 15

facero:
@kosack : There is this proposal to replace current cython code by numba.

Do you have a positive feedback after using numba in ctapipe for years now ?
Did you have issues with new numpy version or anything that after years you regret ? ;-)

AtreyeeS:
I did some tests by comparing the dev version vs this branch on my Ubuntu laptop, and the run times are almost compatible with a small improvement (~1%) using numba as opposed to cython.

QRemy:
Now this should support both jit and cython but I removed the `..._cython`function from the public API.

kosack:
> Do you have a positive feedback after using numba in ctapipe for years now ?

It's been quite nice: we don't have to make compiled packages for each architecture, and get the benefits of a quite large speedup for "bottleneck" functions. The overhead is that the first time you call a function it can be a bit slower due to the JITing, but as long as you are careful with how you run the code, it's great.  I.e. in docker containers or on clusters, make sure you have a shared cache folder per architecture, not one per node otherwise you will re-compile many times. 

I wonder however why you decided to re-write all your functions in "pure" python for-loops, when numba supports most of numpy natively? That may make the code simpler in some cases, and numba will still optimize how it's used - it doesn't really use numpy itself, but has it's own optimized implementations of a subset of numpy functions that get further optimized by the compiler. For example, in pure numpy if you do something like :

```python
@jit(nopython=True)
def some_code(X: np.ndarray,Y: np.ndarray) → np.ndarray
   return (X*Y).sum()
```

You would get a  temporary array for the `X*Y` part, that then gets summed.  But numba will optimize this to have no temporary, and thus will be more memory efficient.  Sometimes using the numpy functions can even give speed boosts for numba compared to hand-rolled for-loops, since it can do optimizations that might not be possible without more hints.

Where we did use a mix of for-loops and numpy was for complicated integrations over multidimensional arrays, where we found writing the function for one row rather than the full array was simpler to read than a fancier vector-based approach, e.g. https://github.com/cta-observatory/ctapipe/blob/7e40bd23019a06793b4ea2d5c29cbf1403093467/src/ctapipe/image/extractor.py#L58C1-L69C2

maxnoe:
To add to what @kosack wrote, some caveats we found:

- enable caching for all methods where it is possible (i.e add `cache=True` to all jit decorators).
- Do you use the functions from numpy that are supported by numba instead of re-implementing them
- We found that `parallel=True` is sometimes slower than not using `parellel` due to multi-processing overhead. Be carefull when benchmarking that. You also need to make users aware of the [environment variables](https://numba.pydata.org/numba-doc/dev/reference/envvars.html#envvar-NUMBA_NUM_THREADS) for steering the number of threads (but that's the same as for cython with e.g. `OMP_NUM_THREADS`. 
- Computing coverage for numba jitted code is currently not supported and debugging or performance profiling is not really great. https://github.com/numba/numba/issues/5028 https://github.com/numba/numba/issues/4268

This project might also be relevant for gammapy: https://github.com/HDembinski/numba-stats

kosack:
Also, I'm not sure if you are familiar with the timeit module or better the `%timeit` Jupyter magic function, but that is a nice way to check the speed differences without running your whole code.  For example I use the following method a lot:

```python
t = {}
t['numpy']  = %timeit -o some_function_numpy(N,M)
t['numba']  = %timeit -o some_function_numba(N,M)

def plot_performance(time_dict):
    mean = [t.average for t in time_dict.values()]
    std = [t.stdev for t in time_dict.values()]
    x = range(len(time_dict))
    plt.errorbar(x, mean, yerr=std, lw=3, fmt="o")
    plt.xticks(np.arange(len(time_dict)), time_dict.keys())
    plt.ylabel("average execution time (s)")
    plt.grid()

plot_performance(t)
````
e.g.
image

**Comment by facero** [comment]
Btw, the numba issue with numpy 2.2 is now solved :
https://numba.readthedocs.io/en/stable/release/0.61.2-notes.html#support-for-numpy-2-2

File: environment-dev.yml [hunk 2]
```diff
@@ -21,7 +21,7 @@ dependencies:
   - astropy
   - click>8.0
   - cython
-  - numpy>1.20
+  - numpy>1.20,<2.2
```
facero, line 24: The requirement can now be <=2.2 right ?

Also regarding should we impose an < to the python version as well ? (If numba lags behind 1 python version)

maxnoe: I wouldn't do that. `numba` correctly pins the numpy and python versions. You, as in gammapy itself, doesn't have this requirement. 

Since you made the dependencies on numba even optional, it makes even less sense to restrict required dependencies based on what an optional dependency supports.

pip will resolve correctly when doing `pip install gammapy` what versions are compatible

environment-dev.yml
```diff
@@ -21,7 +21,7 @@ dependencies:
   - astropy
   - click>8.0
   - cython
-  - numpy>1.20
+  - numpy>1.20,<=2.2
```
QRemy:
```suggestion
  - numpy>1.20
```

QRemy:
> I wonder however why you decided to re-write all your functions in "pure" python for-loops, when numba supports most of numpy natively?

For now I just copied and re-adapted the cython code into pure python equivalent.
Trying simplifications and optimisations using numpy could be done in a separate PR.

facero:
> > Do you have a positive feedback after using numba in ctapipe for years now ?
> 
> It's been quite nice: we don't have to make compiled packages for each architecture, and get the benefits of a quite large speedup for "bottleneck" functions. 

Thanks for the feedback !
About this point, what are the cascades of implications in being a `no-arch` package ? 
Easier conda distributions ? (no more compiled version for each) 

And what about CI ? Do this mean you have less CI runs ? (I see on ctapipe that you still have 2mac+multiple Linux CI)

QRemy:
> Might also be worth seeing if `parallel=True` can help, since with that you can get automatic (if using numpy functions), or manual (using `prange()` instead of `range()`) multi-threading for things that can be parallelized).

I tried to play a bit with that but didn't see significant improvement with `parallel=True` in most of the benchmarks.

kosack:
> I tried to play a bit with that but didn't see significant improvement with parallel=True in most of the benchmarks.

Yeah, there is always a cost-benefit curve for parallelization, as there is overhead involved.  If the data size is too small, or the individual jobs complete too quickly, the cost of sending data to different threads dominates.   So not too surprising.

facero:
Maybe better to keep `parallel=False` by default here.

Probably the best future path to try to speed up here would to replace python loops with numpy as suggested.
But as said earlier this would probably be in a 2nd stage.

maxnoe:
Note that using `prange` has implications also in case of `parallel=False`, as it might help the compiler to vectorize the loop.

Also, fastmath can be an option, depending on precision needs.

QRemy:
Tweaking the parameters for each method or optimizing how they are written could be done follow up PRs,
this one simply introduces the possibility to switch between cython and jit.

gammapy/utils/compilation.py
```diff
@@ -0,0 +1,88 @@
...
+        """Get enum from string."""
+
+        if value == "jit" and not is_numba_available():
+            log.warning("numba is not installed, falling back to cython backend")
+            value = "cython"
+
+        return cls(value)
+
+
+def _get_fit_statistics_cython():
```
registerrier, line 36: The part below might be implemented inside `gammapy.stats` instead?

gammapy/stats/fit_statistics_jit.py
```diff
@@ -0,0 +1,144 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+import numpy as np
+from numba import jit
+
+global TRUNCATION_VALUE  # compile time constant
+TRUNCATION_VALUE = 1e-25
```
registerrier, line 6: Can we avoid declaring the `TRUNCATION_VALUE` twice? Can it be imported from fit_statistic_cython? OR conversely imported there from here?

gammapy/stats/fit_statistics.py
```diff
@@ -305,10 +312,14 @@ def stat_sum_dataset(cls, dataset):
             npred = npred[mask]
 
             weights = dataset.mask.data[mask].astype("float")
-            return weighted_cash_sum_cython(counts, npred, weights)
+            return get_fit_statistics_compiled()["weighted_cash_sum_compiled"](
```
registerrier, line 315: does it slow down the computation?
the compiled fit stat function could also be declared as a class attribute, no?

registerrier (commented):
Thanks @QRemy. This looks good. My main suggestions would be to: 
- have a single declaration of `TRUNCATION_VALUE`
- maybe move the `_get_fit_statistic` functions declarations in `gammapy.stats`?

PR #5754, opened 2025-04-07
Fix `e_peak` implementation of `SuperExpCutoffPowerLaw4FGLDR3SpectralModel`

MRegeard:
My former implementation (#5694) of the energy at the peak of the SED in the `SuperExpCutoffPowerLaw4FGLDR3SpectralModel` is wrong. 
I corrected it according to  https://iopscience.iop.org/article/10.3847/1538-4357/acee67 Eq. 21. 
@AtreyeeS, I think that you mentioned this in the former PR, but I didn't got it at the time.
I also added more test accordingly which I think cover most cases.


bkhelifi (approved):

MRegeard:
@registerrier, I guess we could implement some kind of dichotomy algorithms if no analytical formula is given.

PR #5759, opened 2025-04-07
`FluxProfileEstimator` fails if `reoptimize=True`

fabiopintore:
The `FluxProfileEstimator` fails when the `reoptimize` argument is set to True. This is due to the fact that the `datasets` is converted into a `to_spectrum_datasets`, therefore there is no need to make all the other models free parameters free to vary. 
The PR fixes this issue.

(still in draft)


gammapy/estimators/points/profile.py
```diff
@@ -103,6 +109,14 @@ def __init__(self, regions, spectral_model=None, **kwargs):
             spectral_model = PowerLawSpectralModel()
 
         self.spectral_model = spectral_model
+
+        if "reoptimize" in kwargs:
+            if kwargs["reoptimize"] is True:
+                raise ValueError(
+                    f"reoptimize=True is not available for {self}. Set to `False` instead."
```
AtreyeeS, line 116: I think we can  have a warning  "Reoptimisation not allowed for backward folding estimator. Setting reoptimize=False" and proceed by internally setting reoptimize=False always

bkhelifi: Using an error will then prevent to never be able to re-optimise. This is the wished behaviour?

fabiopintore: I think the answer is yes. 
Currently, it is not possible to re-optimise anyway, as you would get an error which is however much more unclear.

gammapy/estimators/points/profile.py
```diff
@@ -103,6 +111,14 @@ def __init__(self, regions, spectral_model=None, **kwargs):
...
+                )
+        else:
+            kwargs["reoptimize"] = False
```
registerrier, line 121:
```suggestion
        reoptimize = kwargs.get("reoptimize", False)
        if reoptimize is True:
                raise ValueError(
                    f"reoptimize=True is not available for {self}. Set to `False` instead."
                )
```
this might be simpler

registerrier (approved):
Thanks @fabiopintore . This looks good to me.

I have left a small comment inline

registerrier:
Hi @fabiopintore , unfortunately the change I suggested broke the codestyle... Can you check? I can also try to modify directly.

PR #5756, opened 2025-04-07
3PC pulse profile

MRegeard:
I added pulse profiles, which were missing from the 3PC implementation (#5058). 
3PC pulse profile are given in different energy bins, namely:
- \> 0.1 GeV
- 0.05 - 0.1 GeV
- 0.1 - 0.3 GeV
- 0.3 - 1 GeV
- 1 - 3 GeV
- 3 - 1000 GeV
- 10 - 1000 GeV

For this reason I could not bundle everything in one `RegionNDMap`. Therefore I created a dict for each of the above energy bins and an associated map. The keys of the dict are the same as in the 3PC auxiliary file. 
The radio profile and the best fit are simply `RegionNDMap`. For some pulsar, there is no radio emission (Radio-Quiet pulsar), therefore the property raise a `ValueError`. Is this exception good enough or do we prefer a simple warning ? Or a custom Exception ?

Each of the 3 table of the auxiliary file contains metadata such as follow:
```python
OrderedDict([('EXTNAME', 'RADIO_PROFILE'),
             ('LCFREQ', 1400.0),
             ('TELESCOP', 'JBO'),
             ('CHECKSUM', '8F59A9562C568956'),
             ('DATASUM', '3511893980')])
``` 
for Radio 
Or
```python
OrderedDict([('EXTNAME', 'GAMMA_LC'),
             ('TELESCOP', 'Fermi'),
             ('INSTRUME', 'LAT'),
             ('CHECKSUM', '4fQO6ZPL4fPL4ZPL'),
             ('DATASUM', '3858353760')])
```
for LAT pulse profile.

Should this go somewhere ? I think it is especially useful for Radio as it gives the frequency and telescope.

PR #5761, opened 2025-04-07
Plot function for 3PC pulse profile

MRegeard:
This PR adds a plot function to plot 3PC pulse profile. It is largely inspired by the 3PC HTML page plot (see e.g. [here](https://fermi.gsfc.nasa.gov/ssc/data/access/lat/3rd_PSR_catalog/3PC_HTML/J0534+2200.html)).

For now it looks like this:
Capture d’écran 2025-04-07 à 19 12 54
Capture d’écran 2025-04-07 à 19 11 59

I am not sure that we want this in gammapy as it is not really configurable but more of a catalog end products. Maybe it is one of the case where this function would be better as a method on `SourceCatalog3PCObject`. 
Needs test, cleanup and docstring.
Needs https://github.com/gammapy/gammapy/pull/5756 merged.


gammapy/visualization/utils.py
```diff
@@ -422,3 +424,174 @@ def func_to_fit(x, mu, sigma):
...
+        "3000_100000_WtCt",
+        "10000_100000_WtCt",
+    ]
+
+    if not isinstance(source, SourceCatalogObject3PC):
+        raise TypeError(
+            f"`source` must be an instance of `~gammapy.catalog.SourceCatalogObject3PC`, got {type(source)}."
+        )
+    if n_period > 2:
+        raise ValueError(f"`n_preiod` must be either 1 or 2, got {n_period}.")
```
bkhelifi, line 454:
```suggestion
        raise ValueError(f"`n_period` must be either 1 or 2, got {n_period}.")
```

registerrier (commented):
Thanks @MRegeard . I wonder if this is not too specific as a 300 loc function. Shouldn't we have first a generic solution for pulse profiles in gammapy?

MRegeard:
@registerrier, I think that this function is a good and useful plotting utility on the 3PC Catalog. Once you have your catalog source, you can straight forward plot the pulse profiles in different energy bands. 
Otherwise, with the current implementation of 3PC pulse profiles, you can use the `.plot_hist` method of `RegionNDMap` which plot the pulse profiles.

registerrier:
@MRegeard Could you add a test here?

registerrier:
@Astro-Kirsty . Can you add a test here?

gammapy/visualization/catalogs.py
```diff
@@ -0,0 +1,226 @@
...
+        ncols=1,
+        nrows=6,
+        figsize=(7, 12),
+        gridspec_kw={"height_ratios": [1.6, 1, 1, 1, 1, 1]},
+    )
+
+    plt.subplots_adjust(wspace=0, hspace=0)
+    for ax in axes[:-1]:
+        ax.tick_params(
+            axis="x", direction="inout", labelbottom=False, top=True, bottom=True
```
QRemy, line 88: The codespell seems to create a false positive making changes  for `inout` a str variable, any idea why ? 
@morcuended @natthan-pigoux

morcuended: will open a PR to avoid it

morcuended: https://github.com/gammapy/gammapy/pull/6702

QRemy (approved):
Thanks @MRegeard and @Astro-Kirsty, ok to merge for me.
Is there is anything else you want to add

Issue #5753, opened 2025-04-07
gammapy docs and search engines: stop them to show outdated docs.


mireianievas:
**Is your feature request related to a problem? Please describe.**
Right now if you look for documentation of gammapy in google directly, it will sometimes show outdated documentation pages. 

**Describe the solution you'd like**
It seems a good starting point could be to add something like 
` <meta name="robots" content="noindex, nofollow">`

for all outdated documentation pages. That should tell good-behaved search engines to stop going over them and showing results. 

**Additional context**

<img width="1103" alt="Image" src="https://github.com/user-attachments/assets/74fafcca-52c6-4a37-8afd-ddb0ed162bec" />

bkhelifi:
From https://developers.google.com/search/docs/crawling-indexing/robots-meta-tag,

- `noindex` :  Do not show this page, media, or resource in search results. If you don't specify this rule, the page, media, or resource may be indexed and shown in search results.
  - Google should still be able to show the page index.html of our documentation page;) 
- `nofollow` : Do not follow the links on this page. If you don't specify this rule, Google may use the links on the page to discover those linked pages. 
  - Google should then not be able to the likelihood page into our doc.

However, as discussed, one can try that the latest release is put in top, instead of the dev.

PR #5763, opened 2025-04-07
Push the web search engine to display first the stable version

bkhelifi:
**Description**

This pull request is linked to the issue #5753.
The canonization of urls (see thie [link](https://developers.google.com/search/docs/crawling-indexing/canonicalization)) permits to search engine to know which pages are preferred compared to others. This mechanism is used here to tag the `stable` version as the preferred one when displaying search results.
To achieve that, the `html_baseurl` sphinx command permits that... Normally ;)

** Other possible solution
One could use [sphinx-sitemap](https://sphinx-sitemap.readthedocs.io/en/latest/), but it requires adding a library...
You can see the pros and cons on all methods [here](https://developers.google.com/search/docs/crawling-indexing/consolidate-duplicate-urls).

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -95,6 +95,10 @@ source_suffix = '.rst'
 # The master toctree document.
 master_doc = 'index'
 
+# Allow to add the canonical html flag on all the pages
+# This permits to precise to web search engine that the pages for the stable version are privileged
+html_baseurl = 'https://docs.gammapy.org/stable/'
+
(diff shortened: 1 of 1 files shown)


Astro-Kirsty:
Thanks @bkhelifi , this is a nice addition. I have seen this in other packages i.e. astropy. So it should be good!

registerrier (approved):
Thanks @bkhelifi ! This looks like a simple and practical solution

Issue #5173, opened 2024-03-19
to_spectrum_dataset() loses background model


LukasNickel:
**Gammapy version**
1.1, 1.2

**Bug description**
Running `to_spectrum_dataset()` on a `Mapdataset` with a background model produces a spectrum dataset, that is constructed with just the IRF bkg and not the fitted model.

**Expected behavior**
It should take the fit into account. This is the case, when the mapdataset is first stacked / `to_masked` is run, because then the model is correctly applied.

Compare these two code pieces:

https://github.com/gammapy/gammapy/blob/86488ece61f41e92ff3212b9b870f5a9f7d68b79/gammapy/datasets/map.py#L1009-L1017

https://github.com/gammapy/gammapy/blob/86488ece61f41e92ff3212b9b870f5a9f7d68b79/gammapy/datasets/map.py#L1783-L1787

As the `background` does not know of the model, it is not applied.

From my understanding of the intended behavior, the first case is correct and the second one should be fixed.

QRemy:
Yes  indeed`to_spectrum_dataset` should `self.npred_background`instead of `self.background`.

PR #5764, opened 2025-04-08
Fix bug in `dataset.to_region_map_dataset`

fabiopintore:
This PR solves the issue #5173.

### DIFF
--- a/gammapy/datasets/map.py
+++ b/gammapy/datasets/map.py
@@ -1940,7 +1940,7 @@ class MapDataset(Dataset):
             )
 
         if self.stat_type == "cash" and self.background:
-            kwargs["background"] = self.background.to_region_nd_map(
+            kwargs["background"] = self.npred_background().to_region_nd_map(
                 region, func=np.sum, weights=self.mask_safe
             )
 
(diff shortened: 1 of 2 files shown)


bkhelifi (approved):
Thanks for the fix

registerrier (approved):
Thanks @fabiopintore !

AtreyeeS:
Hello, sorry for commenting late on this.
I think this is a general feature of dataset operations (eg:`MapDatset.cutout`, `MapDataset.to_masked`, etc etc) that all models, including the background model is lost.
I agree that this is quite annoying in certain case, (specially for the `to_spectrum_dataset` and `cutout` personally), but since this was the intended behaviour, maybe we should not tag this as a bug-fix and silently. Also, it might be difficult to remember what happens to the models for different operations.

Maybe we can keep the default behaviour, and add an argument (`keep_background_model=False`), and retain the model if True?

fabiopintore:
Thanks @AtreyeeS . I see your point. 
If that was the intended behaviour, then it's fine to apply such a suggestion. If we are all happy with that, I can proceed opening a new PR.

AtreyeeS:
Thanks @fabiopintore ! I think we should at least document the intended behaviour for all dataset operations (in the docstrings and maybe in the user guide as well)

Issue #5570, opened 2024-11-15
Improving peek() on 3D datasets 


mireianievas:
As of now, the peek on a 3D dataset (e.g. from Fermi-LAT) gives something like this

![image](https://github.com/user-attachments/assets/f033481a-f666-46f1-8ee1-b9ed19fa3606)

This is not very nice, and is rather incomplete as it shows very little information about the IRFs (only the exposure). 

For this work https://arxiv.org/abs/2409.20487 I built something a little bit more informative

![image](https://github.com/user-attachments/assets/a684ed1a-180f-43a7-8d50-c24ce60b87d3)

That shows 1D and 2D projections of basically all the IRFs. The only missing part would be possibly the background, which for Fermi-LAT is None (it is included as part of the modelling) so I skipped it.

Maybe something in this line could be considered to replace what peek() does in this case?

AtreyeeS:
Thanks @mireianievas ! Your plot looks very nice, but the tradeoff is between having something very simple (which we wanted the peek to do) vs having something fitting a variety of use cases.
I dont know if the psf/exposure at the center is a useful quantity for IACTs, as we will expect significant variation across the FoV.
Also, one will probably use the peek() before defining any models, just to ensure that the dataset is correctly created

Do you have some ideas on what would work for both IACT and Fermi-LAT (and maybe even WCD)?`
`peek` functions are generally not configurable, so it should preferably not take any arguments

bkhelifi:
The general idea of the `peek' functions is to have a easy and quickly look of quantities. Having them configurable goes beyond our initial goal. 
As our IRFs are a function of many parameters (and we do not have yet the event types!), I think that it would be hard to have plotting functions with kwargs that permit to handle any type of plots for papers. This will depend on the source, the objectives of the plot within a paper, etc

AtreyeeS:
The minimum thing to do here can be to ensure that the `peek` function works for a Fermi-LAT dataset, or any dataset with any missing attributes. So, eg, one can check `if self.excess` before this block https://github.com/gammapy/gammapy/blob/4a4bcc2c80198edd3b02ceac0acfb35c89ed0066/gammapy/datasets/map.py#L2381

and so on for other attributes.

We can also add a `psf.plot_containment_radius_vs_energy` and `edisp.get_edisp_kernel().plot_matrix()`

registerrier:
This is solved in #5766 . Closing.

PR #5766, opened 2025-04-08
Modification of peek() in MapDataset to show more information

mireianievas:
Currently showing:
counts, npred, psf, edisp matrix at center, exposure map, exposure at center

Solves issue #5570

Examples:

Fermi-LAT

image

CTAO DC1 GC sample

image

### DIFF
--- a/gammapy/datasets/map.py
+++ b/gammapy/datasets/map.py
@@ -5,8 +5,10 @@ from scipy.stats import median_abs_deviation as mad
 import astropy.units as u
 from astropy.io import fits
 from astropy.table import Table
-from regions import CircleSkyRegion
+from regions import CircleSkyRegion, RectangleSkyRegion
 import matplotlib.pyplot as plt
+from matplotlib import colormaps
+from matplotlib.colors import LogNorm
 import gammapy.datasets.evaluator as meval
 from gammapy.data import GTI, PointingMode
 from gammapy.irf import EDispKernelMap, EDispMap, PSFKernel, PSFMap, RecoPSFMap
@@ -2344,8 +2346,14 @@ class MapDataset(Dataset):
         energy_axis = self._geom.axes["energy"].squash()
         return self.resample_energy_axis(energy_axis=energy_axis, name=name)
 
-    def peek(self, figsize=(12, 8)):
-        """Quick-look summary plots.
+    def peek(self, figsize=(11, 6.5)):
+        """Quick-look summary plots for a given MapDataset:
+        - Counts map
+        - Npred map
+        - Energy dispersion matrix
+        - Exposure map
+        - Exposure profile
+        - PSF containment radius
 
         Parameters
         ----------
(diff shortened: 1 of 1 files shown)


gammapy/datasets/map.py
```diff
@@ -2321,37 +2329,300 @@ def peek(self, figsize=(12, 8)):
...
+            -------
+            QuadMesh
+                The resulting QuadMesh object (useful if you need a colorbar reference).
+            """
+            mesh = edisp_kernel.plot_matrix(ax=ax, add_cbar=True)
+            ax.set_xscale("log")
+            ax.set_yscale("log")
+            ax.set_title("Energy Dispersion (at FoV center)")
+            ax.set_xlabel("True Energy (MeV)")
+            ax.set_ylabel("Energy (MeV)")
```
k-egg, line 2484: Here for example instead of setting the axis units to MeV it would be better to use the given units from `edisp_kernel.plot_matrix`.

mireianievas:
Fixed, thanks @k-egg for checking !

gammapy/datasets/map.py
```diff
@@ -2321,37 +2329,293 @@ def peek(self, figsize=(12, 8)):
 
         """
 
+        def make_rgb_transparent(rgba, bg_rgb=(1, 1, 1)):
```
bkhelifi, line 2332: Should we move this function and the following 2 in visualization?

gammapy/datasets/map.py
```diff
@@ -2321,37 +2329,293 @@ def peek(self, figsize=(12, 8)):
...
+            index = int(exposure_map.geom.axes["energy_true"].nbin / 2)
+            # Dynamically scale the exposure by powers of 10 for improved readability
+            exponent = int(np.log10(exposure_map.data.mean()))
+            exp_data = exposure_map.get_image_by_idx([index]).to_unit(
+                f"10^{exponent} cm2 s"
+            )
+            energy_center = exposure_map.geom.axes["energy_true"].center[index]
+
+            exp_data.plot(ax=ax, cmap=cmap, add_cbar=True)
+            ax.set_title(f'Exposure map at {energy_center.to("GeV"):.1f}')
```
bkhelifi, line 2508: I would prefer TeV ;)

mireianievas: solved.

mireianievas:
Now it shows

image

gammapy/datasets/map.py
```diff
@@ -2321,37 +2329,298 @@ def peek(self, figsize=(12, 8)):
...
+            Parameters
+            ----------
+            ax : `matplotlib.axes.Axes`
+                The matplotlib Axes to plot on.
+            psf : `PSFKernel` or `PSFMap`
+                The Gammapy PSF object.
+            """
+            psf.plot_containment_radius_vs_energy(ax=ax)
+            ax.legend(fontsize="small")
+            ax.set_title("Containment radius")
```
QRemy, line 2546: should also mention (at FoV center)

gammapy/datasets/map.py
QRemy: I agree `adjust_color_limits`, `create_custom_colormap`, and `make_rgb_transparent`
should go in `gammapy.visualization` preferably in a separate PR that go first so its easier to review and we can make them a bit more general.

gammapy/datasets/map.py
```diff
@@ -2321,37 +2329,298 @@ def peek(self, figsize=(12, 8)):
...
+            return [alpha * c1 + (1 - alpha) * c2 for c1, c2 in zip(rgb, bg_rgb)]
+
+        def create_custom_colormap(npred_min, cmap_name="viridis"):
+            """
+            Create two custom colormaps based on the given colormap and "GnBu".
+
+            This function:
+            - Resamples the specified colormap to 256 colors.
+            - Applies a power-law transform to the RGBA values for a custom transparency ramp.
+            - Sets any "bad" values in the colormap to the color representing `npred_min`.
```
QRemy, line 2359: not sure to understand the logic here
bad values should not be in the range of the colormap, typically nan should appear in white or black

gammapy/datasets/map.py
```diff
@@ -2321,37 +2329,298 @@ def peek(self, figsize=(12, 8)):
...
+            - Returns a second default colormap ("GnBu") for variety.
+
+            Parameters
+            ----------
+            npred_min : float
+                The minimum value of the npred data for color setting.
+            counts_data : ndarray
+                The counts data (used for scaling).
+            cmap_name : str
+                The base colormap name (default is 'viridis').
```
QRemy, line 2369: I would keep "afmhot" as default for consistency with the default of the other plot.

gammapy/datasets/map.py
```diff
@@ -2321,37 +2329,298 @@ def peek(self, figsize=(12, 8)):
...
+            ----------
+            ax : `matplotlib.axes.Axes`
+                The matplotlib Axes to plot on.
+            exposure_map : `Map`
+                Gammapy Map containing the exposure data (with energy axis).
+            cmap : Colormap
+                The colormap to use for plotting.
+            """
+            index = int(exposure_map.geom.axes["energy_true"].nbin / 2)
+            # Dynamically scale the exposure by powers of 10 for improved readability
```
QRemy, line 2500: why not just have it in log scale like the others ?

QRemy (commented):
Thanks @mireianievas 
Few general comments:
-Maybe change the plot order 
top row : exposure , counts, npred (all the maps)
bottom row : exposure vs energy, psf, edsip (so it is also the order in which they are applied)
-the units should be displayed in bracket for consistency with the other plots in gammapy

mireianievas:
> Thanks @mireianievas Few general comments: -Maybe change the plot order top row : exposure , counts, npred (all the maps) bottom row : exposure vs energy, psf, edsip (so it is also the order in which they are applied) -the units should be displayed in bracket for consistency with the other plots in gammapy

Done partially (order of figures). Which units specifically do you refer to?

gammapy/datasets/map.py
mireianievas: deprecated, if using afmhot and correctly setting bad to black, there is no need to introduce this function.

gammapy/datasets/map.py
mireianievas: actually adjust_color_limits is a bit misleading function name, as its sole purpose is to determine some reasonable vmin,vmax to be used in event maps. Not sure how useful it would be outside the scope of this peek() since it basically uses both the npred and counts maps. Maybe I could even move the content of the function to peek() directly.

mireianievas:
* Plots in the order that @QRemy suggested.

* Units inside [ ]

* Colorbar in log-scale for exposure map.

* Figure size increased a little bit (and dpi option removed)

* the 3 functions that were mentioned to be moved somewhere else are gone. Two were not needed with the default cmap, then adjust_color_limits code (just calculating reasonable vmin,vmax) is directly done in peek()

Two examples:

image

image

gammapy/datasets/map.py
```diff
@@ -2342,47 +2344,255 @@ def to_image(self, name=None):
...
 
         Parameters
         ----------
         figsize : tuple
-            Size of the figure. Default is (12, 10).
+            Size of the figure. Default is (13.5, 7).
 
         """
 
+        def plot_counts(ax, counts_data, cmap, vmin, vmax):
```
registerrier, line 2363: I think you don't need two functions for `plot_counts` and `plot_npred`. You could do all at once, passing the map as well as the title explicitly.

gammapy/datasets/map.py
```diff
@@ -2342,47 +2344,255 @@ def to_image(self, name=None):
...
+            Plot the counts map on a given axis.
+
+            Parameters
+            ----------
+            ax : `matplotlib.axes.Axes`
+                The matplotlib Axes to plot on.
+            counts_data : `Map`
+                Gammapy Map containing the counts data.
+            cmap : Colormap
+                The colormap to use for plotting.
```
registerrier, line 2374: For an inline function, parameter description is not useful. If the title is clear enough, even the docstring is not necessary

gammapy/datasets/map.py
```diff
@@ -2342,47 +2344,255 @@ def to_image(self, name=None):
...
+        central_spectrum_dataset = self.to_spectrum_dataset(central_pixel)
+
+        # Determine plotting limits
+        vmin = npredmapdata.data.min()
+        vmax = npredmapdata.data.max()
+        # Fallback if the map is entirely zero
+        if vmin == 0.0:
+            vmin = np.max([countsmapdata.data.max() * 0.02, countsmapdata.data.min()])
+        if vmax == 0.0:
+            vmax = countsmapdata.data.max()
```
registerrier, line 2552: In general, one could have the min and max of the npred equal to that of the counts. No?

registerrier (commented):
Thanks @mireianievas . This is nice. A few comments:

- the inline functions do not need docstrings. You can remove them.
- the plot_counts and plot_npred function are basically identical. You can reuse them.

gammapy/datasets/map.py
mireianievas: fixed

gammapy/datasets/map.py
mireianievas: done

gammapy/datasets/map.py
mireianievas: done

gammapy/datasets/map.py
```diff
@@ -2311,8 +2313,14 @@ def to_image(self, name=None):
...
-    def peek(self, figsize=(12, 8)):
-        """Quick-look summary plots.
+    def peek(self, figsize=(11, 6.5)):
+        """Quick-look summary plots for a given MapDataset:
+        - Counts map
+        - Npred map
+        - Energy dispersion matrix
+        - Exposure map
+        - Exposure profile
+        - PSF containment radius
```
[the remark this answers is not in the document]
mireianievas: done

gammapy/datasets/map.py
```diff
@@ -2342,47 +2344,154 @@ def to_image(self, name=None):
...
+        central_spectrum_dataset = self.to_spectrum_dataset(central_pixel)
+
+        # Determine plotting limits
+        vmin = npredmapdata.data.min()
+        vmax = npredmapdata.data.max()
+        # Fallback if the map is entirely zero
+        if vmin == 0.0:
+            vmin = np.max([countsmapdata.data.max() * 0.02, countsmapdata.data.min()])
+        if vmax == 0.0:
+            vmax = countsmapdata.data.max()
```
QRemy, line 2451: maybe:
```suggestion
        vmin = np.minimum(npredmapdata.data.min(), countsmapdata.data.min())
        vmax = np.maximum(npredmapdata.data.max(), countsmapdata.data.max())
```

mireianievas: Actually maybe not ok. If one of them is zero, not sure if it will work with the log-scaled mapping?

mireianievas:
@QRemy I tested the more compact np.minimum / np.maximum and found some cases for which it fails with 'invalid vmin/vmax' while plotting. So rolling back to the previous version.

gammapy/datasets/map.py
```diff
@@ -5,8 +5,10 @@
 import astropy.units as u
 from astropy.io import fits
 from astropy.table import Table
-from regions import CircleSkyRegion
+from regions import CircleSkyRegion, RectangleSkyRegion
 import matplotlib.pyplot as plt
+from matplotlib import colormaps
+from matplotlib.colors import LogNorm
```
QRemy, line 11:
```suggestion
from matplotlib import colormaps
from matplotlib.colors import LogNorm, Colormap
```

gammapy/datasets/map.py
```diff
@@ -2342,47 +2344,154 @@ def to_image(self, name=None):
...
+        vmin = npredmapdata.data.min()
+        vmax = npredmapdata.data.max()
+        # Fallback if the map is entirely zero
+        if vmin == 0.0:
+            vmin = np.max([countsmapdata.data.max() * 0.02, countsmapdata.data.min()])
+        if vmax == 0.0:
+            vmax = countsmapdata.data.max()
+
+        # Create custom colormaps
+        cmapcustom = colormaps.get_cmap("afmhot")
```
QRemy, line 2454:
```suggestion
        cmapcustom = Colormap("afmhot")
```

gammapy/datasets/map.py
```diff
@@ -2340,47 +2341,154 @@ def to_image(self, name=None):
...
+
+        def plot_exposure_map(ax, exposure_map, cmap):
+            index = int(exposure_map.geom.axes["energy_true"].nbin / 2)
+
+            # Dynamically scale the exposure by powers of 10 for improved readability
+            exp_data = exposure_map.get_image_by_idx([index])
+            vmin = exp_data.data[exp_data.data > 0].min()
+            vmax = exp_data.data[exp_data.data > 0].max()
+
+            energy_center = exposure_map.geom.axes["energy_true"].center[index]
```
QRemy, line 2386:
```suggestion
            energy_center = exposure_map.geom.axes[0].center[index]
```

mireianievas:
Any other request @QRemy @k-egg @registerrier ? @bkhelifi ?

QRemy:
There is some issues in the tests related to the way colormap is defined, I'm trying different solutions.

gammapy/datasets/map.py
```diff
@@ -5,8 +5,9 @@
 import astropy.units as u
 from astropy.io import fits
 from astropy.table import Table
-from regions import CircleSkyRegion
+from regions import CircleSkyRegion, RectangleSkyRegion
 import matplotlib.pyplot as plt
+from matplotlib.colors import LogNorm, Colormap
```
QRemy, line 10:
```suggestion
from matplotlib.colors import LogNorm
```

gammapy/datasets/map.py
```diff
@@ -2340,47 +2341,154 @@ def to_image(self, name=None):
...
+        vmin = npredmapdata.data.min()
+        vmax = npredmapdata.data.max()
+        # Fallback if the map is entirely zero
+        if vmin == 0.0:
+            vmin = np.max([countsmapdata.data.max() * 0.02, countsmapdata.data.min()])
+        if vmax == 0.0:
+            vmax = countsmapdata.data.max()
+
+        # Create custom colormaps
+        cmapcustom = Colormap("afmhot")
```
QRemy, line 2451:
```suggestion
        cmapcustom = plt.get_cmap("afmhot")
```

QRemy (approved):
Thanks @mireianievas, seems good now, will merging if no objection.

Issue #2103, opened 2019-04-01
Fix remaining test warnings


adonath:
Currently when running the tests with `pytest gammapy -s` or `python -c 'import gammapy; gammapy.test()'` a lot of warnings are raised.  This is a reminder to fix those remaining test warnings for v0.12.

Here is a gist with current warnings output:

https://gist.github.com/cdeil/4a3563c5d19ed231f0b822aa15d39cc7

adonath:
Test improvements will be part of the roadmap for v2.0

registerrier:
In the preparation of v2.0, we should try to improve the situation w.r.t. tests warnings. This could be part of the effort of the next coding sprint.
First thing is to make a census of warnings and create issues for each subpackage.

Astro-Kirsty:
I started to write some notes for this issue, starting with our `xfails` we have implemented. An overview:

> - analysis, astro, datasets, scripts, visualisation -- no `xfail`
> - catalog --  1 `xfail` (create issue)
> - data --  3 `xfail`s (create issue)
> - estimators -- 2 `xfail`s (fluxmaps implement ref model? points core unsure)
> - irf, makers, modeling -- `xfail`s adapted see PR #6120
> - maps -- 1 `xfail`

**More in depth for discussion etc:**

- [ ] `xfail` in `gammapy/catalog/tests/test_fermi.py` -- should be removed along with the function
- if we really want to support [light curve functionality in the PC catalogs](https://github.com/gammapy/gammapy/blob/04a818ed732c45b50f88ef0a8b5bfa402d8b9169/gammapy/catalog/tests/test_fermi.py#L807) we should make a proper issue for that 

- [ ] `xfail` in `gammapy/data` -- we should also create issues for the 'checkers' on the data store i.e. test_observations

- [ ] `xfail` in `gammapy/estimators` -- we have a [FluxMaps test](https://github.com/gammapy/gammapy/blob/04a818ed732c45b50f88ef0a8b5bfa402d8b9169/gammapy/estimators/map/tests/test_core.py#L400) which doesn't include a reference model, but we don't support that. Do we want to implement that or not?
[...]

PR #5768, opened 2025-04-08
Correct pydantic deprecation warning

registerrier:
**Description**

This pull request correct pydantic deprecation warning due to the `models_base`being accessible only from  the model class and no longer from the instance.

**Dear reviewer**


registerrier:
This is part of the warning cleanup effort #2103 .

QRemy (approved):
Thanks @registerrier

PR #5772, opened 2025-04-08
Fix to_region_nd_map with masks

### LINKED ISSUE #5751: `to_region_nd_map()` weighted averaging errors
**Gammapy version**
1.3

**Bug description**
Using `to_region_nd_map()` with averaging functions, e.g. `np.mean()` and weights gives wrong values.

**Expected behavior**
A region map containing the weighted average of the input map.

**To Reproduce**
As the `weights` parameter is often used to input masks:
```python
m = WcsNDMap.create(npix=(200, 200))
m.data = np.ones_like(m.data)

weights = WcsNDMap.create(npix=(200, 200))
weights.data = np.zeros_like(weights.data,dtype=bool)
weights.data[:,100:]=True

m.to_region_nd_map(weights=weights,func=np.mean).data
```
Output:
```
array([[0.5]], dtype=float32)
```

A similar problem occurs with non-boolean weights:

```python
weights = WcsNDMap.create(npix=(200, 200))
weights.data = np.ones_like(weights.data)
weights.data[:,100:]=0.1
weights.data[:,:100]=0.2

m.to_region_nd_map(weights=weights,func=np.mean).data
```
Output:
```
array([[0.15000002]], dtype=float32)
```

**Other information**
I believe the issue lies in the multiplication of the weights with the input data, which is done regardless of the input function. I have looked into it, but I believe finding a general solution suitable for all functions may be a challenge.

k-egg:
**Description**

This PR is to resolve #5751. If a mask is give as `weights`, the mask is applied. If not, the weights are multiplied as normal.
A test that was relying on the old behaviour was also fixed.

This pull request is still a draft.


gammapy/maps/wcs/ndmap.py
```diff
@@ -617,7 +617,9 @@ def to_region_nd_map(
             Function to reduce the data. Default is np.nansum.
             For boolean Map, use np.any or np.all.
         weights : `WcsNDMap`, optional
-            Array to be used as weights. The geometry must be equivalent.
+            Array to be used as weights.
+            Weights are multiplied with the input data and should be defined accordingly, in case a weighted average is desired.
+            If a mask is given the mask is applied to the input map. The geometry must be equivalent.
```
bkhelifi, line 622:
```suggestion
            If a mask is given, the mask is applied to the input map. The geometry must be equivalent.
```

QRemy:
Requires #5719 first that invalidates the failing tests in `test_map_dataset_on_off_to_spectrum_dataset_weights`

gammapy/maps/wcs/ndmap.py
```diff
@@ -650,16 +652,22 @@ def to_region_nd_map(
...
-            idx_y, idx_x = np.where(mask)
-            data = func(cutout.data[..., idx_y, idx_x], axis=-1)
+                if weights_cutout.is_mask:
+                    mask.data = np.logical_and(mask.data, weights_cutout.data)
+                    cutout.data *= weights_cutout.data
+                else:
+                    cutout.data *= weights_cutout.data
+
+            data = func(cutout.data, where=mask.data, axis=(-2, -1)).astype(
+                self.data.dtype
```
QRemy, line 669: could you replace this by the lines that were used before ?

```
idx_y, idx_x = np.where(mask)
 data = func(cutout.data[..., idx_y, idx_x], axis=-1)
```

k-egg: Not directly at least, since the mask is now n-dimensional.

bkhelifi (approved):
Great! Thanks a lot

gammapy/maps/wcs/ndmap.py
```diff
@@ -650,16 +652,22 @@ def to_region_nd_map(
             # Casting needed as interp_by_coord transforms boolean
             data = data.astype(self.data.dtype)
         else:
-            cutout, mask = self.cutout_and_mask_region(region=region)
+            cutout, cutout_mask = self.cutout_and_mask_region(region=region)
+            mask = cutout_mask.interp_to_geom(cutout.geom)
```
QRemy, line 656: does this return a boolean ?
for mask I think we should use `method="nearest"`.

registerrier (approved):
Thanks @k-egg . Looks good to me! No specific comments.

PR #5773, opened 2025-04-08
Add scheduled ci test for rst and examples

### LINKED ISSUE #5480: No test on rst testcode
We don't test rst testcode which produces error. 

In the user-guide/estimators.rst, the last output (from the print) of the last testcode has changed since numpy 2.0, which produces compilation and rendering errors of the docs. 

This has gone under the radar because we don't automatically test the rst testcode. We should setup an automated test from time to time on this.

Astro-Kirsty:
This PR is to resolve #5480. It adds a scheduled run of the doctests both for the rst code and the examples in the docstrings.


registerrier (approved):

This could be called `docstests.yml` directly.

morcuended (approved):
Looks good, thanks @Astro-Kirsty. It will be very helpful indeed.

registerrier (approved):
Thanks @Astro-Kirsty . Looks good to me!

PR #5778, opened 2025-04-09
Fix plotting of bow-ties in spectral plot_error(). Fixes #5738

### LINKED ISSUE #5738: Plotting bow-ties in log-log space
**Gammapy version**

In log-log plots, like in SEDs, if you plot a pure the power-law component of a composite spectral_model that contains an absorber (e.g. when fitting something with EBL)  I would expect to recover a perfect bow-tie shape. Instead, in I get something like this:

Image

I think that is a buggy representation of that model. I guess this arises from converting from linear-space to log-spaced, but it probably should be handled and 'simetrized' on the fly by gammapy.

For the coding-sprint ?

mireianievas:
Fix bow tie plotting.  Fixes #5738 

Main differences: 
- new variable (islog) to be passed, which triggers the calculation of log-flux and error log-flux instead of flux, err flux, in which case asymmetric flux errors are calculated and plotted. 

Some examples

image

Left ones have errors of 1/15  in amplitude and index, just to illustrate that in the limit of small errors they basically yield the same output. Right plots have a larger 1/5 error in amplitude and index, and they start to give weird shapes of the error bands for the log-plot. With a PWL instead a LP you get something similar, so it does not really depend on the spectral shape, but rather on the error propagation itself.

This commit needed to modify several parts of that spectral.py module, so better if someone checks it throughfully. Probably some refactoring of code would make it more elegant too.

### DIFF
--- a/gammapy/modeling/models/spectral.py
+++ b/gammapy/modeling/models/spectral.py
@@ -167,7 +167,7 @@ class SpectralModel(ModelBase):
     def __rsub__(self, model):
         return self.__sub__(model)
 
-    def _propagate_error(self, epsilon, fct, **kwargs):
+    def _propagate_error(self, epsilon, fct, islog=False, **kwargs):
         """Evaluate error for a given function with uncertainty propagation.
 
         Parameters
@@ -177,6 +177,9 @@ class SpectralModel(ModelBase):
         epsilon : float
             Step size of the gradient evaluation. Given as a
             fraction of the parameter error.
+        islog : bool, optional
+            Whether the error propagation should be done on the log of
+            the function instead of the function itself.
         **kwargs : dict
             Keyword arguments.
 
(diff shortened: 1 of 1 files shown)


mireianievas:
So a bit more on this. 

This 'fix' is more a cosmetical thing after all. Cosmetic and artificial. The fitting is happening in linear-space because what is linear is the relation counts - flux, not counts - log(flux). So in log(flux) the errors are bound to be asymmetric. 

Probably the only correct solution is going to sampling, which is fine if your model is easy but probably not if your model is very complex (e.g. Fermi 3D model in a busy sky region).

Opinions?

registerrier:
Thanks @mireianievas ! This is good.

The uncertainty propagation is based on a linearization of the function of the parameters. In the case of powerlaw-like spectra, the log of the flux should indeed provide a description valid on a broader range than the flux itself. So the error band in log flux is what you should plot as you do here.

mireianievas:
More tests. 

Snippet with sampling (68 % coverage) using scipy's multivariate module:

```
plt.figure(figsize=(5,5),dpi=100)
PWL.plot_error(energy_bounds=[100*u.GeV,10*u.TeV],sed_type='e2dnde',islog=False,
               facecolor=(0.1,0.6,0.8,0.1),
               edgecolor=(0.1,0.6,0.8,0.5),linewidth=1,label='linear-errprop')
PWL.plot_error(energy_bounds=[100*u.GeV,10*u.TeV],sed_type='e2dnde',islog=True,
               facecolor=(0.8,0.1,0.1,0.1),
               edgecolor=(0.8,0.1,0.1,0.5),linewidth=1,label='log-errprop')
fp.plot(sed_type='e2dnde',color='black')
plt.ylim(1e-15,2e-11)
plt.xlim(60*u.GeV,20*u.TeV)


N=1000
energies = np.geomspace(100*u.GeV,10*u.TeV)
parameters = src.spectral_model.parameters
covariance = src.spectral_model.covariance.get_subcovariance(parameters)
for k in range(5):
    rnd = multivariate_normal.rvs(mean=parameters.value, cov=covariance.data,size=N)
    par_sample = [[rnd[i,j]*parameters[j].unit for j in range(len(parameters))] for i in range(len(rnd))]
    flx = [src.spectral_model.evaluate(energies,*p) for p in par_sample]
    
    # Compute 68% confidence envelope (1 sigma)
    lower = np.percentile(flx, 16, axis=0)*flx[0][0].unit
    upper = np.percentile(flx, 84, axis=0)*flx[0][0].unit
    median = np.percentile(flx, 50, axis=0)*flx[0][0].unit
    
    plt.fill_between(energies,lower*energies*energies,upper*energies*energies,
                     color=(0.3,0.3,0,0.05),
                     edgecolor=(0.3,0.3,0,0.3),linewidth=1,linestyle='dashed',label=f'Sampling N={N}' if k==0 else '')

plt.legend()
```

image
image
image

This would be for 1 source (at N=100 it is unreliable, N=1k samples we still get some changes from realisation to realisation, for N=10k samples it is still reasonably fast to evaluate [~ 4s] and converges well with all 5 realisations yielding similar bands). 

For N sources we need to see, probably as discussed today we have to still do the computation --even if not correct-- ignoring the full covariance matrix and computing this with just the reduced covariance matrix. 

PS. These plots are with some DL4 from real data.

mireianievas:
I think this is superseded by #5786 . We can close this.

PR #5779, opened 2025-04-09
Add ObservationMetadata lazy loading

registerrier:
**Description**

This pull request introduces a mechanism similar to the one introduced for `pointing` to create the `ObservationMetaData`.

**Dear reviewer**

### DIFF
--- a/gammapy/data/observations.py
+++ b/gammapy/data/observations.py
@@ -374,8 +374,6 @@ class Observation:
     @property
     def pointing(self):
         """Get the pointing for the observation as a `~gammapy.data.FixedPointingInfo` object."""
-        if self._pointing is None:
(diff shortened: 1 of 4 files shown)


gammapy/data/observations.py
```diff
@@ -375,8 +374,6 @@ def observation_dead_time_fraction(self):
     @property
     def pointing(self):
         """Get the pointing for the observation as a `~gammapy.data.FixedPointingInfo` object."""
-        if self._pointing is None:
-            self._pointing = FixedPointingInfo.from_fits_header(self.events.table.meta)
```
QRemy, line 379: what happens if there is no pointing hdu ?

QRemy (commented):
Thanks @registerrier, why is there some changes related to the pointing ?

QRemy (approved):
Thanks for the clarification, ok to merge.

Issue #5775, opened 2025-04-09
Add conditions for multiple selections in `DataStore.obs_table`


fabiopintore:
**Is your feature request related to a problem? Please describe.**
`Datastore.obs_table.select_observations` allows to handle multiple selection criteria. They are all applied with an `AND` condition. However, it would be needed to take into account an `OR` condition between two or more selection criteria.

**Describe the solution you'd like**
The observation selection should allow for multiple conditions.

**Describe alternatives you've considered**
It might be possible to extend the keys of the selection dictionary to account for the boolean condition and, when is `OR`, to which other selection it applies. For now, we could keep it simple allowing only one `OR` condition.

PR #5780, opened 2025-04-09
Allow for multiple conditions in `obs_table.select_observations`

fabiopintore:
Fix to #5775 .
In draft.

### DIFF
--- a/gammapy/data/obs_table.py
+++ b/gammapy/data/obs_table.py
@@ -1,6 +1,6 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 from collections import namedtuple
-import itertools
+from itertools import groupby, chain
 import operator
 import numpy as np
 from astropy.coordinates import Angle, SkyCoord
@@ -249,6 +249,8 @@ class ObservationTable(Table):
         **inverted** flag, in which case, the selection is applied to keep all
         elements outside the selected range.
 
+        Multiple conditions can be also used.
+
         A few examples of selection criteria are given below.
 
         Parameters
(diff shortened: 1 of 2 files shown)


gammapy/data/obs_table.py
```diff
@@ -286,12 +288,49 @@ def select_observations(self, selections=None):
         >>> selection = dict(type='par_box', variable='N_TELS', value_range=[4, 4])
         >>> selected_obs_table = obs_table.select_observations(selection)
         """
+        if isinstance(selections, list):
+            for selection in selections:
+                if isinstance(selection, dict):
+                    if "condition" in selection:
+                        continue
+                    else:
+                        selection["condition"] = "AND"
```
bkhelifi, line 297: One would need a nice example in the docstring to show the usage;)

QRemy:
There is a lot of overlap with what is implemented in #5743 which implement the selection at the datastore level.
The idea is that the selection should be done at the datastore to be less dependent on the GADF obs-table format in future.

fabiopintore:
thanks @bkhelifi . The code is bit twisted, but in practice it only manipulates the dictionary(ies) passed as selection filter. It can be simplified for sure. I've added some examples in the documentation and expanded the tests.

@QRemy I agree, the selection should be made at the DataStore creation. However, do you see any issue if the two functionalities co-exist for now?

bkhelifi:
After discussion, it seems much better to have only one place for the selection. And it seems that we converge towards the obs_table...
I leave the last word to the LDs;)

fabiopintore:
Okay, fantastic. Thanks to have discussed it :)

gammapy/data/obs_table.py
```diff
@@ -285,13 +289,93 @@ def select_observations(self, selections=None):
 
         >>> selection = dict(type='par_box', variable='N_TELS', value_range=[4, 4])
         >>> selected_obs_table = obs_table.select_observations(selection)
+
+        Examples for multiple conditions:
+        >>> selection_tels = dict(type='par_box', variable='N_TELS', value_range=[4, 4], operator="AND")
+        >>> selection_alt = dict(type='par_box', variable='ALT_PNT', value_range=value_range, operator="AND")
+        >>> selection_time_box = dict(type='time_box', time_range=time_range, operator="AND")
```
QRemy, line 296: I find this whole logic with the operator overcomplicated for something we can do with simple boolean array indexing :
obs_table[(obs_table["N_TELS"]==4)&(obs_table["ALT_PNT"]>min_value)&(obs_table["ALT_PNT"]<max_value)]

Moreover `selection_event_type = dict(type='par_box', variable='EVENT_TYPE', value_range=["TYPE1", "TYPE1"]`, operator="AND") won't work because type here is a `str`
while `obs_table["EVENT_TYPE"]=="TYPE1"` just work and is simpler.

In my opinion the selection methods `par_box` and `time_box`  (as tstart, tstop can be added to the obs-table) are not helping and could be removed. Instead we should show user how to use boolean array indexing which every numpy user should know anyway (so there is no gammapy specific learning curve increase)

Similarly `obs_table.select_sky_circle(center, radius, inverted=False)`
can be simply written as 
`obs_table[obs_table.pointing_galactic.separation(center) < radius]`
separation should be know by astropy user and gammpy users just have to know that the attribute `pointing_galactic` exits (we could even remove `pointing_galactic` and `pointing_radec` to have only `pointing`). 

So in short I would suggest to remove all the special selection method and expose more on numpy/astropy functionalities and add an API tutorial to specifically explain how to select observations from the obs_table.
If you really want to keep the overcomplicated selection methods I would make them return boolean array so we simply can combine the different selections with the basic operators in one line, and Datastore.get_observation could also accept a boolean mask.

QRemy (commented):
Thanks @fabiopintore , here is a long comment it's not specifically about this PR but also on the observation selection methods in general:

gammapy/data/obs_table.py
bkhelifi: I share your view about the functions associated with `par_box` and `time_box`. A more generic approach could be found.

fabiopintore: Thanks for the comment @QRemy . 
Yes, the proposed approach is quite overcomplicated. I did that to preserve the selection functions and to maintain the selection scheme which is currently implemented. 
However, I agree with you that the `par_box`, `time_box` and `select_sky_circle` can be removed in favour of simple boolean indexing of the table. As far as a tutorial is provided, I do not think the users will complain. In fact, I would say that the boolean indexing is certainly easier to learn than the current selection functions.

registerrier (commented):
Thanks @fabiopintore .

I think the proposed changes don't bring additional functionality as `and` and `not` allow to perform all boolean operations. So the existing scheme which adds sequentially filters with `inverted` option should already allow to perform what you want to support.

 Now whether we should remove the existing selection methods is another question. In general, if observation selection has to be performed in a configurable manner, such functions are necessary at some point.

registerrier:
PR superseded by #5855

PR #5788, opened 2025-04-10
Fix broken icons and add help buttons to documentation

Astro-Kirsty:
This PR is to fix the broken icons. This PR is again trying to limit the css and focuses on utilising the defaults from the PyData Sphinx theme

To do this I decided we can utilised icons from [octicon](https://primer.style/octicons/) (could also use others i.e. 'fontawesome' or 'material' icons). 
Feel free to suggest other icons if you do not like the ones Maxime and I chose! 

I've also added an extra row on the front page that allows the users to seek help i.e. through joining the slack channel or creating an issue. 

Here are what the rendered versions look like: [Home page](https://github.com/user-attachments/assets/46342725-77c0-4be2-bb45-b0d6e6166b28) and  [User guide](https://github.com/user-attachments/assets/bcb5ac80-94af-4bfb-a6b7-87e47561d008)


bkhelifi (approved):
This is very nice and it looks better than before. Thank you!

registerrier (commented):
Thanks @Astro-Kirsty . This new look is good.

Putting the get in touch in the main page is an option. One issue with the slack is that if the address in not part of the list of emails it is not possible to register. We don't have an easy solution for that.

morcuended:
Thanks, it looks great! I also like reducing the dependence on css.

>Don’t have an email address from one of those domains?
Contact the workspace administrator at gammapy for an invitation.

This is what it's stated at the end of the sign-up Slack page. Is that enough or needs another heads up in by the link to Slack?

PR #5799, opened 2025-04-11
Order of spectral model parameters not consistent between definition, docstring and evaluate.

mireianievas:
Issue discovered on the EBL evaluation. Order of parameters was wrong in <= 1.3 . evaluate() gives wrong results. 

redshift, alpha_norm but EBL evaluate() expects alpha_norm, redshift

### DIFF
--- a/gammapy/datasets/tests/test_map.py
+++ b/gammapy/datasets/tests/test_map.py
@@ -796,12 +796,12 @@ def test_map_fit(sky_model, geom, geom_etrue):
     assert_allclose(pars["amplitude"].error, 4.216e-13, rtol=1e-2)
 
     # background norm 1
-    assert_allclose(pars[8].value, 0.5, rtol=1e-2)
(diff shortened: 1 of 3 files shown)


Astro-Kirsty (commented):
Thanks @mireianievas, this was well spotted. 
I would suggest we try to keep a common order i.e. always `index`, `amplitude`, `reference`. Can you adjust it accordingly?

mireianievas:
Hi @Astro-Kirsty 

I did the changes that were minimally invasive. Unfortunately some functions define and make calls to amplitude first. I could modify that instead of the docstring, but the risk of breaking something is higher.

registerrier (approved):
Thanks @mireianievas! This is cleaner this way and should avoid future issues.

QRemy (approved):
Thanks @mireianievas, looks good, merging now.

PR #5804, opened 2025-04-11
New test to evaluate spectral models (and spot issues like PR-5799)

mireianievas:
The order of the parameters as defined in some spectral models (EBLAbsorption, PowerLawNormSpectralModel) is not consistent, as the ''evaluate'' calls expect the parameters to be passed in a different order. Adding a test to spot such cases.


bkhelifi (approved):
Waouh! Well done ;)

QRemy (approved):
Thanks @mireianievas
- There is an issue with the evaluate of PiecewiseNormSpectralModel that do not work like the others (it does not accept passing parameters but use alwyas the ones on self).
Uniformizing this could be done in a separate PR
- Similar tests can be added for temporal and spatial models in a separate PR.
Merging this now.

PR #5805, opened 2025-04-11
Add further message for checking docstring

Astro-Kirsty:
As per the title


registerrier (approved):
Thanks @Astro-Kirsty ! Looks good.

PR #5807, opened 2025-04-13
datasets write() docstring typo

### LINKED ISSUE #5806: Datasets.write() docstring typo?
Image

by default it seems set write_covariance=True, however the docstring says it is set to False.

mireianievas:
covariance is written by default

Fixes #5806


registerrier (approved):
Thanks @mireianievas . Well spotted. Looks good.

PR #5812, opened 2025-04-15
Switch default SpectrumDataset write/read format to gadf (better preservation of masks)

mireianievas:
OGIP does not serialize the fit_mask, only the safe_mask. Since our OGIP format is anyway custom, I believe it is better to serialize by default as GADF (same as with the MapDataset). 

For the reader part, it looks like if you try to read as GADF something that is in fact an OGIP file, it will not fail but return a dataset with counts, background, acceptance, edisp and exposure as None. Therefore I think we can safely try to read as GADF by default, and if all those are None, try with OGIP to support old datasets written with previous gammapy versions.


mireianievas:
Tests adapted to always set the format. Adding a couple of tests to check if the dataset is recovered with the default format (i.e. no format specified in write and read) and another one to check the fallback from 'saving as ogip', reading as gadf (fallbacks into ogip reader as soon as the resulting counts, background, edisp and exposure objects are None).

gammapy/datasets/spectrum.py
```diff
@@ -326,14 +326,23 @@ def read(cls, filename, format="ogip", checksum=False, name=None, **kwargs):
...
-            return super().read(
+            mapd = super().read(
                 filename, format="gadf", checksum=checksum, name=name, **kwargs
             )
+            if (
+                mapd.counts is None
+                and mapd.background is None
+                and mapd.exposure is None
+                and mapd.edisp is None
+            ):
```
QRemy, line 337: in which case does it happen ? is there a way to fix it ?

QRemy (approved):
Thanks @mireianievas, I just have one question :

gammapy/datasets/spectrum.py
mireianievas: Well, it is a perhaps very dirty way of spotting that this dataset is indeed saved as ogip, and if you try to force-read it as gadf it will return a dataset with everything set to None. Probably doing all those ands is a bit of an overkill.

QRemy: Maybe the default format should be`None` and it would be guessed based on this such as we use `ogip` if the output of `gadf` is all `None`. 
But if possible the format should be guessed via some mandatory keys existing in the gadf or ogip headers.

mireianievas: Both done. Defaults to None (guess format), and guess the format based on extensions 

```if "SPECTRUM" in hdulist and "OGIP" in hdulist[1].header["HDUCLASS"]``` ---> OGIP
```elif ("COUNTS" in hdulist and "EXPOSURE" in hdulist and "EDISP" in hdulist):``` ---> GADF

raises an error now if it is a MapDataset rather than Spectrum dataset ('IMAGE' in hdulist[1].header["XTENSION"])

**Review by QRemy — APPROVED** [review]
Thanks @mireianievas, looks good, no further comment.

File: gammapy/datasets/spectrum.py [hunk 2]
```diff
@@ -314,8 +316,8 @@ def read(cls, filename, format="ogip", checksum=False, name=None, **kwargs):
         ----------
         filename : `~pathlib.Path` or str
             OGIP PHA file to read.
-        format : {"ogip", "ogip-sherpa", "gadf"}
-            Format to use. Default is "ogip".
+        format : {"ogip", "ogip-sherpa", "gadf", None}, optional
```
registerrier, line 319:
```suggestion
        format : {"ogip", "gadf"}, optional
```
technically `ogip-sherpa` (for energy in keV) is not an option on read

gammapy/datasets/spectrum.py
```diff
@@ -314,8 +316,8 @@ def read(cls, filename, format="ogip", checksum=False, name=None, **kwargs):
...
+            Format to use. Default is None, which guesses the format (OGIP or GADF).
```
registerrier, line 320:
```suggestion
            Format to use. If None, try to guess the format. Default is None
```

gammapy/datasets/spectrum.py
```diff
@@ -325,15 +327,28 @@ def read(cls, filename, format="ogip", checksum=False, name=None, **kwargs):
...
 
-        if format == "gadf":
-            return super().read(
-                filename, format="gadf", checksum=checksum, name=name, **kwargs
-            )
+        if format is None:
+            with fits.open(filename) as hdulist:
+                # Check if the file is empty
+                if len(hdulist) == 0:
+                    raise ValueError(f"File {filename} is empty")
```
registerrier, line 334: Why not move this at the end with rather cannot determine format

gammapy/datasets/spectrum.py
```diff
@@ -325,15 +327,28 @@ def read(cls, filename, format="ogip", checksum=False, name=None, **kwargs):
...
+                # Check for extensions in OGIP format
+                if "SPECTRUM" in hdulist and "OGIP" in hdulist[1].header["HDUCLASS"]:
```
registerrier, line 336:
```suggestion
                if "SPECTRUM" in hdulist and "OGIP" in hdulist["SPECTRUM"].header["HDUCLASS"]:
```
It seems safer to test this specific hdu

gammapy/datasets/spectrum.py
```diff
@@ -314,8 +316,8 @@ def read(cls, filename, format="ogip", checksum=False, name=None, **kwargs):
         ----------
         filename : `~pathlib.Path` or str
             OGIP PHA file to read.
```
registerrier, line 318:
```suggestion
            Name of the file to read.
```

gammapy/datasets/spectrum.py
```diff
@@ -325,15 +327,28 @@ def read(cls, filename, format="ogip", checksum=False, name=None, **kwargs):
...
+            with fits.open(filename) as hdulist:
+                # Check if the file is empty
+                if len(hdulist) == 0:
+                    raise ValueError(f"File {filename} is empty")
+                # Check for extensions in OGIP format
+                if "SPECTRUM" in hdulist and "OGIP" in hdulist[1].header["HDUCLASS"]:
+                    format = "ogip"
+                # Check for extensions in GADF format
+                elif ("COUNTS" in hdulist and "EXPOSURE" in hdulist and "EDISP" in hdulist):
+                    format = "gadf"
```
registerrier, line 340: given the limited amount of info currently serialized this will be difficult to identify in the future. e.g. so far we have not serialized the `HDUCLASS` in the so-called gadf format...
So you could check that there is a `COUNTS_BAND` extension.

gammapy/datasets/spectrum.py
```diff
@@ -325,15 +327,28 @@ def read(cls, filename, format="ogip", checksum=False, name=None, **kwargs):
...
+                    if 'IMAGE' in hdulist[1].header["XTENSION"]:
+                        raise ValueError(f"File {filename} is not a GADF spectrum, but a GADF map")
```
registerrier, line 342: Maybe check that the `REGION` HDU is there instead?

gammapy/datasets/tests/test_io.py
```diff
@@ -124,16 +125,16 @@ def test_spectrum_datasets_to_io(tmp_path):
 @requires_data()
 def test_ogip_writer(tmp_path):
     dataset = SpectrumDatasetOnOff.read(
-        "$GAMMAPY_DATA/joint-crab/spectra/hess/pha_obs23523.fits"
+        "$GAMMAPY_DATA/joint-crab/spectra/hess/pha_obs23523.fits",
+        format="ogip",
     )
-    dataset.counts_off = None
+    dataset.counts_off.data = np.zeros(dataset.counts_off.data.shape)
```
registerrier, line 131: Why changing this specific test?

gammapy/datasets/tests/test_spectrum.py
```diff
@@ -629,6 +633,20 @@ def test_spectrum_dataset_onoff_fits_io(self, tmp_path):
...
+        assert d1.exposure == self.dataset.exposure
+        assert_allclose(d1.counts_off.data, self.dataset.counts_off.data)
+
+    def test_spectrum_dataset_onoff_fallback_from_gadf(self, tmp_path):
+        self.dataset.write(tmp_path / "test.fits", format="ogip")
+        d1 = SpectrumDatasetOnOff.read(tmp_path / "test.fits")
+        assert isinstance(d1.counts.geom, RegionGeom)
+        assert d1.exposure == self.dataset.exposure
+        assert_allclose(d1.counts_off.data, self.dataset.counts_off.data)
+
```
registerrier, line 649: This misses a test of the format identification with the proper exception raise.

registerrier (changes requested):
Thanks @mireianievas . Introducing a format identification on read seems a good idea. It needs to be well tested though. In particular, the exception raising should be well checked.

Changing the default format on `write` is less clear to me as the format is even more gammapy internal and will likely change. 
I understand that this change is made because `Datasets.write()` cannot set the format for each dataset type. So why not. But `Datasets.write` should be improved.

A minima `write` needs better documentation to expose that the `ogip` format won't serialize the `mask_fit`.

gammapy/datasets/spectrum.py
```diff
@@ -325,15 +327,28 @@ def read(cls, filename, format="ogip", checksum=False, name=None, **kwargs):
         """
         from .io import OGIPDatasetReader
 
-        if format == "gadf":
-            return super().read(
-                filename, format="gadf", checksum=checksum, name=name, **kwargs
-            )
+        if format is None:
+            with fits.open(filename) as hdulist:
```
registerrier, line 331: Note that this won't work in general (e.g. env variable). You should use `path = make_path(filename)`

This is why tests fail.

gammapy/datasets/tests/test_io.py
mireianievas: see https://github.com/gammapy/gammapy/pull/5812#issuecomment-2807431211

mireianievas:
> Thanks @mireianievas . Introducing a format identification on read seems a good idea. It needs to be well tested though. In particular, the exception raising should be well checked.
> 
> Changing the default format on `write` is less clear to me as the format is even more gammapy internal and will likely change. I understand that this change is made because `Datasets.write()` cannot set the format for each dataset type. So why not. But `Datasets.write` should be improved.
> 
> A minima `write` needs better documentation to expose that the `ogip` format won't serialize the `mask_fit`.

True that the current GADF is some sort of placeholder for a future definite format for gamma-ray data, but "our" custom OGIP is not any better. It is also custom in a way, and even incomplete for our purposes (no mask_fit as you say). I agree Datasets.write can be improve to expose the format to write the individual Dataset in a given format, but probably that could be done in a separate PR ?

gammapy/datasets/tests/test_spectrum.py
mireianievas: I believe it is implemented now 

> This misses a test of the format identification with the proper exception raise.

gammapy/datasets/spectrum.py
mireianievas: if REGION is there it should be a SpectrumDatasetOnOff right? If so, this is implemented now.

gammapy/datasets/spectrum.py
mireianievas: I believe this is fixed now

QRemy:
@mireianievas what's the status here ?
I'll move it to 2.1 for now.

QRemy (approved):
Thanks @mireianievas, ready to merge.

QRemy:
The issue reported by SONAR  is not new, I would just ignore it and merge this PR.

PR #5813, opened 2025-04-15
Missing negative sign in PLEC4 `evaluate`

MRegeard:
According to both the equation in gammapy model gallery, as well as, in Eq. 3 of [4FLG paper](https://arxiv.org/pdf/2201.11184), the `evalutate` method of `SuperExpCutoffPowerLaw4FGLDR3SpectralModel` is missing a negative sign.


bkhelifi (approved):
An important fix! Thanks

QRemy (approved):
Thanks @MRegeard, no comment from my side.

Issue #5769, opened 2025-04-08
User feedback on documentation


Astro-Kirsty:
We (+ @bkhelifi) outline here the feedback from the beginners in the gammapy coding sprint and the CTAO SDC.

**Simple fixes**
- [x] fix the link in HowTo for `make an on-axis equivalent livetime map` - PR #5784 
- [x] add format types we support to the Analysis Workflow page - PR #5785 
- [x] add `important` banner at the beginning of the tutorials for the `notice`
- [x] add navigation depths to the API reference so they can be drop down on the side - PR #5810
- [x] add navigation depths to the tutorials so that you can see the headers - PR  #5810
- [x] add further navigation depths for the gammapy analysis workflow and package structure - PR #5810
- [x] some pages are hidden in the IRF: 
	- derivation of the wstat and cash are hidden - PR #5852 
	- IRF functions are only available in the main page (if you manage to read the entire thing!) - PR #5851 
- [x] notice -- tutorials are generic for any instrument, we can provide the needed specificities for each data release for each experiment
- [x] change the wording [here](https://docs.gammapy.org/dev/getting-started/index.html#recommended-setup) to say that the analysis workflow is best to start with - PR #5787 
- [x] change title `Analysis workflow and package structure` to `Fundamental Concepts`. Could have the old one as a subtitle. - PR #5785
- [x] add how to catch deprecation warnings in the tests [here](https://docs.gammapy.org/dev/development/dev_howto.html#removing-an-attribute) - PR #5876 
- [x]  Improve the docstrings of `SamplerResult` and `Sampler` - PR #5853 
- [x] Improve the documentation of the reflectedRegionMaker - see issue #4492   - PR #5925 

**Tutorials**
- [x] Re-ordering of tutorials  - see issue #5796  - PR #5949 
- [x] API fitting tutorial update -- this is frequentist but for bayesian see.. - PR #5803 
- [x] Add note for beginners about the default values of the function, chosen so that beginners do not need to change them
- [x] Add a tutorial on UL computation (issue #5771) - PR #5933 
  - Including flux UL, a parameter UL computation (extension, energy cutoff)
- [x] Add a dedicated theta-square tutorial - PR #5802 

**Other**
- [ ] create a function from the `DataStore` that allows you to easily obtain the livetime at your source position. Yes we have `make_observation_time_map` but this involves more steps - PR #5815 
- [x] expose model gallery better, 2 ideas:   
[...]

PR #5815, opened 2025-04-15
Creation of a util function `get_effective_livetime`

bkhelifi:
**Description**
As mentioned in the issue #5769 , the user feedback of internal CTAO SDC leads to wish the feature to compute quickly and without coding the livetime for a given position.
This pull request offers this feature, based on the code in the Maker tutorials.

**Dear reviewer**
I am still wondering whether one should either keep the function `make_observation_time_map` or introduce a new `DataStore` function associated to it or upgrade the function implemented here by introducing a "mode" calling `make_observation_time_map` .

Once accepted and merged, one should into an other PR
- clean-up the maker tutorial
- create a new "example tutorial" with this function and the nice plotting (still using MSH 15-52)
- update the HowTo

### DIFF
--- a/docs/development/doc_howto.rst
+++ b/docs/development/doc_howto.rst
@@ -284,12 +284,32 @@ Links to Gammapy API are handled via normal Sphinx syntax in comments:
 
 .. code-block:: python
 
-   # Create an `~gammapy.analysis.AnalysisConfig` object and edit it to
-   # define the analysis configuration:
+   # Create a `~gammapy.modeling.models.PowerLawSpectralModel` object:
 
-   config = AnalysisConfig()
+   pwl = PowerLawSpectralModel()
 
-This will resolve to a link to the ``AnalysisConfig`` class in the API documentation.
+The ``~`` prefix hides the full module path in the rendered documentation, so that link will appear
+as ``PowerLawSpectralModel`` which points to the API documentation.
+
+General guidelines
+^^^^^^^^^^^^^^^^^^
+
+- Use the ``~`` prefix to hide long paths unless showing the full path provide clarity
+- Use the full path when the shortened name is unclear or ambiguous
+- For example, ``~gammapy.datasets.FluxPointsDataset.residuals`` will render as ``residuals``, which may
+  not be informative. Instead use the full path ``gammapy.datasets.FluxPointsDataset.residuals``, so that
+  one knows the explicit class name where this belongs.
+
+
+Referencing APIs in the same module
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+When referencing classes or functions defined in the same module, you can omit full path. For example,
+as ``PowerLawNormSpectralModel`` is in the same file as ``PowerLawSpectralModel``, you can simply write
+``PowerLawSpectralModel``, instead of the full path and without the ``~``.
+
+
+
 
 .. _dev-check_html_links:
 
@@ -325,7 +345,7 @@ If you wish to add an image to a tutorial you can utilise the following:
     #
     #   Gammapy Maps Illustration
 
-Where `:alt` is used to create a caption for the image. 
+Where `:alt` is used to create a caption for the image.
 More information on the image directive can be found `here <http://www.sphinx-doc.org/en/stable/rest.html#images>`__.
 
 Documentation guidelines


registerrier (commented):
Thanks @bkhelifi. I must say I don't understand the logic of creating something like this on the `DataStore`. It is not an analysis tool and should not try to compute quantities that are not shipped with the data.

bkhelifi:
> Thanks @bkhelifi. I must say I don't understand the logic of creating something like this on the `DataStore`. It is not an analysis tool and should not try to compute quantities that are not shipped with the data.

The logic is the following
- this is a feature collected during the SDC user feedback
- Gammapy is used as SAT for CTAO
  - I guess that one does not want to maintain 2 libraries...
- Attaching it to the datastore follows this use case:
  - Before learning how to use Gammapy for my analysis, is there data on my favorite source?
  - So, attaching it to the datastore permits to have in 2 lines the answer;)
- The other use case is the proposal preparations in general:
  - is there past data on my favorite source
 
 I do not see the issue to help users...  
 
 Concerning your sentance "should not try to compute quantities that are not shipped with the data.", I am a bit puzzled... because if this quantity was not in the data, one would not be able to compute the stack value;)

bkhelifi:
Hi Regis,
The function has been removed from the datastore... into makers.utils

gammapy/makers/utils.py
```diff
@@ -671,6 +672,80 @@ def make_effective_livetime_map(observations, geom, offset_max=None):
...
+    Returns
+    -------
+    livetime_maps : `~gammapy.maps.Map` or list of `~gammapy.maps.Map`
+        Effective livetime map(s).
+    selected_ids : list of `str`
+        List of selected observation IDs.
+
+    # Examples
+    # --------
+    # ToDo: add the tutorial link after
```
registerrier, line 708: I would avoid comments and TODOs in the docstring

gammapy/makers/utils.py
```diff
@@ -671,6 +672,80 @@ def make_effective_livetime_map(observations, geom, offset_max=None):
...
+        If None, a `border` of 6 deg is used. Default is None.
+    en_edges : list of `~astropy.units.Quantity`
+        Edges of true energy bins. Default is None.
+
+    Returns
+    -------
+    livetime_maps : `~gammapy.maps.Map` or list of `~gammapy.maps.Map`
+        Effective livetime map(s).
+    selected_ids : list of `str`
+        List of selected observation IDs.
```
registerrier, line 704: Same here. You could think that passing the list of obs_id to the function could also be perfectly acceptable. No?

gammapy/makers/utils.py
```diff
@@ -671,6 +672,80 @@ def make_effective_livetime_map(observations, geom, offset_max=None):
...
+    border = Angle(6, "deg") if border is None else border
+    selection = dict(
+        type="sky_circle",
+        frame="icrs",
+        lon=position.ra,
+        lat=position.dec,
+        radius=Angle(0.1, "deg") if radius is None else radius,
+        border=border,
+    )
+    selected_obs_table = datastore.obs_table.select_observations(selection)
```
registerrier, line 720: what about `required_irf`? This won't work for a pointlike data distribution.

gammapy/makers/utils.py
```diff
@@ -671,6 +672,80 @@ def make_effective_livetime_map(observations, geom, offset_max=None):
...
+
+    # Define the geom
+    edges = [1 * u.GeV, 1 * u.PeV] if en_edges is None else en_edges
+    energy_axis_true = MapAxis.from_energy_edges(edges, name="energy_true")
+    geom = WcsGeom.create(
+        skydir=position,
+        binsz=0.02,
+        width=(border.value * 2, border.value * 2),
+        frame="icrs",
+        proj="CAR",
```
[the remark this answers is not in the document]
bkhelifi: Any other suggestion?

gammapy/makers/utils.py
```diff
@@ -671,6 +672,80 @@ def make_effective_livetime_map(observations, geom, offset_max=None):
...
+
+    livetime_maps = make_effective_livetime_map(observations, geom, offset_max=border)
+
+    effective_times_src = livetime_maps.get_by_coord(
+        (position, energy_axis_true.center)
+    )
+    log.info(f"Time spent on position {position}:")
+    for idx, _ in enumerate(edges[:-1]):
+        log.info(
+            f"Effective Livetime in [{edges[idx]}, {edges[idx + 1]}] : {effective_times_src[idx]:.2f} {livetime_maps.unit}"
```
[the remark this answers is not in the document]
bkhelifi: The edge in energy, not in position....

gammapy/makers/utils.py
```diff
@@ -671,6 +672,80 @@ def make_effective_livetime_map(observations, geom, offset_max=None):
...
+    border : `~astropy.coordinates.Angle`, optional.
+        Additional edge summed to the region of interest used for the selection of the observations.The maximum offset FoV. Default is None.
+        If None, a `border` of 6 deg is used. Default is None.
+    en_edges : list of `~astropy.units.Quantity`
+        Edges of true energy bins. Default is None.
+
+    Returns
+    -------
+    livetime_maps : `~gammapy.maps.Map` or list of `~gammapy.maps.Map`
+        Effective livetime map(s).
```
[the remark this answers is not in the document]
bkhelifi: a list as a function of energy... This is not so weird.
I would have made a cube in the return function... Here, I made a list

gammapy/makers/utils.py
bkhelifi: I do not understand your point. selected_ids is the list of obs_ids....

bkhelifi:
@registerrier : the function computes the livetime in a geometry centered on the user position. So, it gives numbers at the position of the users.

gammapy/makers/utils.py
bkhelifi: Good point... we need to check whether the previously coded functions works also (which was not anticipated in the past).

gammapy/makers/utils.py
```diff
@@ -671,6 +672,80 @@ def make_effective_livetime_map(observations, geom, offset_max=None):
...
+    )
+
+    livetime_maps = make_effective_livetime_map(observations, geom, offset_max=border)
+
+    effective_times_src = livetime_maps.get_by_coord(
+        (position, energy_axis_true.center)
+    )
+    log.info(f"Time spent on position {position}:")
+    for idx, _ in enumerate(edges[:-1]):
+        log.info(
```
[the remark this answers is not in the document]
bkhelifi: Both are provided in fact...

gammapy/makers/utils.py
bkhelifi: The core Gammapy function `make_effective_livetime_map` is indeed even not tested for point-like IRFs...
Should I make a separate issue?
For safety, I change the way to get the list of obs_id...

PR #5818, opened 2025-04-16
Fix links in docs after link check

Astro-Kirsty:
There are a number of links that are broken when running link check. This aims to fix as many of those as possible.


bkhelifi (approved):
Many many thanks

QRemy (approved):

PR #5819, opened 2025-04-16
Remove unneeded click argument for Python 3

Astro-Kirsty:
We should be fine to not use
'click.disable_unicode_literals_warning = True'

unicode literals does not exist in Python 3.
Click 8.0 and above doesn’t emit this warning anymore, because Python 2 support was dropped.


QRemy (approved):
Thanks @Astro-Kirsty, merging now.

PR #5802, opened 2025-04-11
Add a small tutorial on the theta-square plot

bkhelifi:
This is a small example tutorial.

Should we then clean-up the HESS tutorial?

PR #5823, opened 2025-04-18
Cleaning of the HESS tutorial

bkhelifi:
**Description**

After the creation of the theta2 tutorial (#5802), I propose to clean up the HESS tutorial by removing the theta-square plot


QRemy (approved):
Thanks @bkhelifi, ok for me.

Issue #5827, opened 2025-04-21
GTI "intersect" and as a possible argument for dataset selection?


mireianievas:
Use case:

I want to select the simultaneous data belonging to two Datasets objects (e.g. from two instruments)

As of now, a _union_ method is implemented in GTI (https://docs.gammapy.org/1.3/api/gammapy.data.GTI.html#gammapy.data.GTI.union) , but there is to my knowledge no _intersect()_ method available.
 
Such method would be handy to select the common GTI from two or more datasets and to potentially select those data that are simultaneously acquired.

Opinions?

registerrier:
Hi @mireianievas . 
If I rephrase the use case you propose, you want to build datasets on common time intervals., hence the necessity to determine the intersection of their GTIs. Correct?

What I don't understand is why is an intersection of the GTI would be necessary on `Datasets` as the time range of the datasets cannot be changed afterwards. The motivation of #5836 is therefore not very clear to me.

PR #5833, opened 2025-04-23
Implement GTI intersect method and its test.

mireianievas:
Partially addresses #5827 . Calculating the intersection of two GTI becomes possible. 

A second PR will take care of implementing the selection of data based on GTI (not only time intervals)


gammapy/data/gti.py
```diff
@@ -438,6 +438,56 @@ def union(self, overlap_ok=True, merge_equal=True):
...
+    def intersect(self, other, padding=0.0):
+        """Intersection of overlapping time intervals between two GTIs.
+
+        Returns a new `~gammapy.data.GTI` object.
+
+        Parameters
+        ----------
+        other : `~gammapy.data.GTI`
+            Another GTI object to intersect with.
+        padding : float, optional
```
registerrier, line 450: Why padding is necessary? On GTIs you would not want to change an interval right?

gammapy/data/gti.py
```diff
@@ -438,6 +438,56 @@ def union(self, overlap_ok=True, merge_equal=True):
...
+        for start1, stop1 in zip(self.time_start, self.time_stop):
+            for start2, stop2 in zip(other.time_start, other.time_stop):
+                # Compute the overlap
+                start = max(start1, start2)
+                stop = min(stop1, stop2)
+                # If there is a positive-duration overlap, keep it
+                if stop > start:
+                    padded_start = start - padding_delta
+                    padded_stop = stop + padding_delta
+                    intervals.append({"START": padded_start, "STOP": padded_stop})
```
QRemy, line 473: might be more efficient to do `intersections(self.time_intervales, other.time_intervals)`
with the function from https://stackoverflow.com/a/40368603
and having `time_intervals` as a `lazyproperty`

gammapy/data/gti.py
```diff
@@ -438,6 +438,56 @@ def union(self, overlap_ok=True, merge_equal=True):
...
+
+        Returns
+        -------
+        intersected_gti : `~gammapy.data.GTI`
+            A GTI object whose intervals are the intersections of the two GTIs.
+        """
+        from astropy.time import TimeDelta
+
+        intervals = []
+        padding_delta = TimeDelta(padding, format="sec")
```
QRemy, line 461: why a padding is this needed ?
don't you have to ensure that the padding do not add overlap between the output intervals ?

gammapy/data/gti.py
```diff
@@ -438,6 +438,56 @@ def union(self, overlap_ok=True, merge_equal=True):
         merged = Table(rows=merged, names=["START", "STOP"], meta=self.table.meta)
         return self.__class__(merged, reference_time=self.time_ref)
 
+    def intersect(self, other, padding=0.0):
```
QRemy, line 441: It might be confusing that union applies only to self while intersection requires an other, could `other` be optional for both union and intersection ?

QRemy (commented):
Thanks @mireianievas, I have few questions :

gammapy/data/gti.py
```diff
@@ -438,6 +438,56 @@ def union(self, overlap_ok=True, merge_equal=True):
...
+        if not intervals:
+            # No overlapping GTIs
+            empty_table = Table(names=["START", "STOP"], dtype=["f8", "f8"])
+            return self.__class__(empty_table, reference_time=self.time_ref)
+
+        # Build an Astropy Table with START/STOP columns
+        intersected_table = Table(rows=intervals, names=["START", "STOP"], meta=self.table.meta)
+
+        # Create a GTI from that table (using the same reference time as self)
+        intersected_gti = self.__class__(intersected_table, reference_time=self.time_ref)
```
registerrier, line 484: Maybe use instead `from_time_intervals`?

gammapy/data/gti.py
```diff
@@ -438,6 +438,56 @@ def union(self, overlap_ok=True, merge_equal=True):
...
+
+        # Optionally, merge any touching intervals
+        intersected_gti = intersected_gti.union(overlap_ok=True, merge_equal=True)
```
registerrier, line 487: Shouldn't this depend on an option?

gammapy/data/gti.py
```diff
@@ -438,6 +438,56 @@ def union(self, overlap_ok=True, merge_equal=True):
...
+                # If there is a positive-duration overlap, keep it
+                if stop > start:
+                    padded_start = start - padding_delta
+                    padded_stop = stop + padding_delta
+                    intervals.append({"START": padded_start, "STOP": padded_stop})
+
+        if not intervals:
+            # No overlapping GTIs
+            empty_table = Table(names=["START", "STOP"], dtype=["f8", "f8"])
+            return self.__class__(empty_table, reference_time=self.time_ref)
```
registerrier, line 478: Maybe rely on `create` or `from_time_intervals` classmethods

gammapy/data/gti.py
```diff
@@ -438,6 +438,56 @@ def union(self, overlap_ok=True, merge_equal=True):
...
+
+        # Build an Astropy Table with START/STOP columns
+        intersected_table = Table(rows=intervals, names=["START", "STOP"], meta=self.table.meta)
```
registerrier, line 481: We shouldn't be using `table.meta` as `GTI.meta` is the proper metadata container.

registerrier (commented):
Thanks @mireianievas . This can indeed be a useful addition. I have some inline comments.

The nitpick question might be whether the result of `intersect` is a GTI or a list of time intervals to be passed to `Observation.select_time`

The difference with `GTI.union` which is only there to merge overlapping intervals in a given GTI is also a bit confusing. 

These issues might be due to the fuzzy scope of GTI at the moment.

QRemy:
Is there still a need for that and would you have time ot wrok on it @mireianievas ?
If not I will close this PR (if necessary you could re-open it whenever you have time).

PR #5835, opened 2025-04-23
Correct Datasets.stat_sum() to apply prior only once

registerrier:
**Description**

This pull request corrects a bug in `Datasets.stat_sum()`. 

Currently priors are computed on each dataset and added to the dataset `stat_sum`. This is incorrect. Prior on a parameter should be added only once to the total likelihood.

This PR modes prior_stat_sum computation to `Datasets` to avoid this issue. A test is added that ensures that the change to the stat_sum is correct when the same model is applied to several datasets.

**Dear reviewer**

### DIFF
--- a/gammapy/modeling/tests/test_fit.py
+++ b/gammapy/modeling/tests/test_fit.py
@@ -31,7 +31,9 @@ class MyDataset(Dataset):
 
     def __init__(self, name="test"):
         self._name = name
-        self._models = Models([MyModel(x=1.99, y=2.99e3, z=3.99e-2)])
+        model = MyModel(x=1.99, y=2.99e3, z=3.99e-2)
+        model.name = name
+        self._models = Models([model])
         self.data_shape = (1,)
         self.meta_table = Table()
 
--- a/gammapy/datasets/core.py
+++ b/gammapy/datasets/core.py
@@ -81,11 +81,7 @@ class Dataset(abc.ABC):
 
     def stat_sum(self):
         """Total statistic given the current model parameters and priors."""
-        prior_stat_sum = 0.0
-        if self.models is not None:
-            prior_stat_sum = self.models.parameters.prior_stat_sum()
-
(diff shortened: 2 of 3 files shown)


registerrier:
A change was necessary on the class `MyDataset` used in test_datasets.py. This is because the class does not ensure models have different names. 
This was not a problem as long as `self.models` was not called in `Datasets.stat_sum`. The change corrects this.

QRemy (commented):
Thanks @registerrier, does the fit and estimators always use a `Datasets` internally ?

registerrier:
> Thanks @registerrier, does the fit and estimators always use a `Datasets` internally ?

The input of all functions such as `Fit.run()`, `Fit.optimize()` etc is sent to this function: https://github.com/gammapy/gammapy/blob/main/gammapy/modeling/utils.py#L4

So in general, `Fit` (and `Sampler`) should see only `Datasets` objects.

QRemy (approved):

gammapy/modeling/tests/test_fit.py
```diff
@@ -31,7 +31,9 @@ class MyDataset(Dataset):
 
     def __init__(self, name="test"):
         self._name = name
-        self._models = Models([MyModel(x=1.99, y=2.99e3, z=3.99e-2)])
+        model = MyModel(x=1.99, y=2.99e3, z=3.99e-2)
+        model.name = name
```
bkhelifi, line 35: the dataset name is the same than the model name? Is it what you wish?

bkhelifi (approved):
Thank you for this fix! Should we add a test with 2 models, as this is the purpose of this fix?

gammapy/modeling/tests/test_fit.py
registerrier: yes. the point is only to make sure model names are different as this is a constraint imposed on `Datasets.models`

PR #5836, opened 2025-04-23
Add function to select datasets given an input GTI.

mireianievas:
Linked to PR #5833 and issue #5827

* Add function to select datasets given an input GTI.
* Add the corresponding unit tests for select_time and the new select_from_gti

### DIFF
--- a/gammapy/datasets/core.py
+++ b/gammapy/datasets/core.py
@@ -288,6 +288,42 @@ class Datasets(collections.abc.MutableSequence):
                 datasets.append(dataset)
 
         return self.__class__(datasets)
+    
+    def select_from_gti(self, gti, atol="1e-6 s", allow_partial=False):
+        """Select datasets corresponding to each interval in a GTI object.
+
+        Parameters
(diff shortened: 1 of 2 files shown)


QRemy (changes requested):
Thanks @mireianievas, I think this could be simplified :

registerrier (commented):
Thanks @mireianievas .

I think the GTI selection makes less sense at the `Datasets` level. Why would one use a `GTI` rather than several time intervals?

mireianievas:
> Thanks @mireianievas .
> 
> I think the GTI selection makes less sense at the `Datasets` level. Why would one use a `GTI` rather than several time intervals?

I guess this is a fermi bias in a way, since some data selection  can be made in gtselect & gtmktime. But it is true that one could pass a list of time intervals.

QRemy:
Is there still a need for that, and would you have time to work on it @mireianievas ?
If not I will close this PR (if necessary you can re-open it whenever you have time).

PR #5837, opened 2025-04-23
Adapt parameter conf_min/conf_max (more stable default for ul and sensitivity)

QRemy:
Resolves the issue found in https://github.com/gammapy/gammapy-benchmarks/pull/158
The recent changes of con_min/conf_max are not sufficient for ul computation  if the fitted norm/error tends to zero.
I don't see a way out other than setting the min/max to some large values by default.
 (it also solves nan in sensitivity computation).

### DIFF
--- a/gammapy/estimators/flux.py
+++ b/gammapy/estimators/flux.py
@@ -76,7 +76,8 @@ class FluxEstimator(ParameterEstimator):
     ):
         self.source = source
 
-        self.norm = _get_default_norm(norm, scan_n_sigma=n_sigma_ul + 1, interp="log")
+        scan_n_sigma = np.maximum(n_sigma_ul, n_sigma_sensitivity) + 1
+        self.norm = _get_default_norm(norm, scan_n_sigma=scan_n_sigma, interp="log")
 
         super().__init__(
             null_value=0,
--- a/gammapy/estimators/flux.py
+++ b/gammapy/estimators/flux.py
@@ -54,7 +54,7 @@ class FluxEstimator(ParameterEstimator):
     norm : `~gammapy.modeling.Parameter` or dict
         Norm parameter used for the fit.
         Default is None and a new parameter is created automatically,
-        with value=1, name="norm", min=-1e3, max=1e3, scan_min=0.2, scan_max=5, and scan_n_values = 11.
+        with value=1, name="norm", scan_min=0.2, scan_max=5, and scan_n_values = 11.
         By default, the min and max are not set and derived from the source model,
         unless the source model does not have one and only one norm parameter.
         If a dict is given the entries should be a subset of
--- a/gammapy/estimators/flux.py
+++ b/gammapy/estimators/flux.py
@@ -54,7 +54,7 @@ class FluxEstimator(ParameterEstimator):
     norm : `~gammapy.modeling.Parameter` or dict
         Norm parameter used for the fit.
         Default is None and a new parameter is created automatically,
-        with value=1, name="norm", scan_min=0.2, scan_max=5, and scan_n_values = 11.
+        with value=1, name="norm", min=-1e3, max=1e3, scan_min=0.2, scan_max=5, and scan_n_values = 11.
         By default, the min and max are not set and derived from the source model,
         unless the source model does not have one and only one norm parameter.
         If a dict is given the entries should be a subset of
--- a/gammapy/estimators/points/lightcurve.py
+++ b/gammapy/estimators/points/lightcurve.py
@@ -77,7 +77,7 @@ class LightCurveEstimator(FluxPointsEstimator):
     norm : ~gammapy.modeling.Parameter` or dict
         Norm parameter used for the fit
         Default is None and a new parameter is created automatically,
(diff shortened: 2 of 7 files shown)


QRemy:
> I don't see a way out other than setting the min/max to some large values by default.

ok that does not work  trying other solutions.

gammapy/estimators/flux.py
```diff
@@ -54,7 +54,7 @@ class FluxEstimator(ParameterEstimator):
     norm : `~gammapy.modeling.Parameter` or dict
         Norm parameter used for the fit.
         Default is None and a new parameter is created automatically,
-        with value=1, name="norm", scan_min=0.2, scan_max=5, and scan_n_values = 11.
+        with value=1, name="norm", min=-1e3, max=1e3, scan_min=0.2, scan_max=5, and scan_n_values = 11.
```
QRemy, line 57:
```suggestion
        with value=1, name="norm", scan_min=0.2, scan_max=5, and scan_n_values = 11.
```

gammapy/estimators/points/lightcurve.py
```diff
@@ -77,7 +77,7 @@ class LightCurveEstimator(FluxPointsEstimator):
     norm : ~gammapy.modeling.Parameter` or dict
         Norm parameter used for the fit
         Default is None and a new parameter is created automatically,
-        with value=1, name="norm", scan_min=0.2, scan_max=5, and scan_n_values = 11.
+        with value=1, name="norm", min=-1e3, max=1e3, scan_min=0.2, scan_max=5, and scan_n_values = 11.
```
QRemy, line 80:
```suggestion
        with value=1, name="norm", scan_min=0.2, scan_max=5, and scan_n_values = 11.
```

gammapy/estimators/points/sed.py
```diff
@@ -77,7 +77,7 @@ class FluxPointsEstimator(FluxEstimator, parallel.ParallelMixin):
     norm : `~gammapy.modeling.Parameter` or dict
         Norm parameter used for the fit
         Default is None and a new parameter is created automatically,
-        with value=1, name="norm", scan_min=0.2, scan_max=5, and scan_n_values = 11.
+        with value=1, name="norm", min=-1e3, max=1e3, scan_min=0.2, scan_max=5, and scan_n_values = 11.
```
QRemy, line 80:
```suggestion
        with value=1, name="norm", scan_min=0.2, scan_max=5, and scan_n_values = 11.
```

QRemy:
It hard to find a default that works for every case, this should work until someone find a case where it breaks.

gammapy/modeling/parameter.py
```diff
@@ -357,26 +357,27 @@ def _step(self):
...
+            min_ = np.minimum(min_, -large_step * 1e5)
+            return min_
 
     # TODO: possibly allow to set this independently
     @property
     def conf_max(self):
-        """Confidence maximum value as a `float`.
+        """Confidence maximum value as a `float`."""
 
-        Return parameter maximum if defined, otherwise return the scan_max.
```
bkhelifi, line 374: Why removing this line? This might be an error... One can keep it, reflecting the docstring of `conf_min`

bkhelifi (changes requested):
Thanks @QRemy. Looks good, only a small comment on a docstring;)

I guess that you have made many tests to be sure that this changes improve the current behaviour... A tiny summary will be nice to have in order to share the knowledge.

Once merge, I propose that we work together on the UL tutorial.
Thank you again

gammapy/modeling/parameter.py
```diff
@@ -357,26 +357,27 @@ def _step(self):
...
+            min_ = self.value - self._step * self.scan_n_sigma
+            large_step = np.maximum(self._step, np.abs(self.value))
+            min_ = np.minimum(min_, -large_step * 1e5)
+            return min_
 
     # TODO: possibly allow to set this independently
     @property
     def conf_max(self):
-        """Confidence maximum value as a `float`.
+        """Confidence maximum value as a `float`."""
```
QRemy, line 372:
```suggestion
        """Confidence maximum value as a `float`.
        Return parameter maximum if defined, otherwise  a default is estimated from value and error.
        """
```

gammapy/modeling/parameter.py
```diff
@@ -357,26 +357,27 @@ def _step(self):
     # TODO: possibly allow to set this independently
     @property
     def conf_min(self):
-        """Confidence minimum value as a `float`.
-
-        Return parameter minimum if defined, otherwise return the scan_min.
-        """
+        """Confidence minimum value as a `float`."""
```
QRemy, line 360:
```suggestion
        """Confidence minimum value as a `float`.
        Return parameter minimum if defined, otherwise  a default is estimated from value and error.
        """
```

gammapy/modeling/parameter.py
QRemy: added a line describing the new behaviour.

bkhelifi (approved):
Thanks

PR #5842, opened 2025-04-28
Replace np.dot by np.matmul

registerrier:
**Description**

This pull request replaces `np.dot` by `np.matmul` in `apply_edisp` function. This improves performances of `aply_edisp` on typical geometries by a factor of a few. Edisp convolution is not the bottleneck in the calculation so the overall gain on benchmarks is marginal. But this might provide more substantial gains in some cases.

**Dear reviewer**
On the following example the change is 4ms to 0.8ms.

```
import astropy.units as u
from gammapy.irf import EDispKernel, PSFKernel
from gammapy.datasets.utils import apply_edisp
import numpy as np

axis = MapAxis.from_energy_bounds("1 TeV", "10 TeV", nbin=30, name="energy_true")
m = Map.create(
        skydir=(0.8, 0.8),
     width=(2, 2),
     binsz=0.02,
     axes=[axis],
    frame="galactic"
)
size = np.prod(m.data.shape)
m.data = np.arange(size).reshape(m.data.shape)
e_true = m.geom.axes[0]
e_reco = MapAxis.from_energy_bounds("1 TeV", "10 TeV", nbin=20)
edisp = EDispKernel.from_diagonal_response(energy_axis_true=e_true, energy_axis=e_reco)

map_edisp = apply_edisp(m, edisp)
```

### DIFF
--- a/gammapy/datasets/utils.py
+++ b/gammapy/datasets/utils.py
@@ -65,7 +65,7 @@ def apply_edisp(input_map, edisp):
     if edisp is not None:
         loc = input_map.geom.axes.index("energy_true")
         data = np.rollaxis(input_map.data, loc, len(input_map.data.shape))
-        data = np.dot(data, edisp.pdf_matrix)
+        data = np.matmul(data, edisp.pdf_matrix)
         data = np.rollaxis(data, -1, loc)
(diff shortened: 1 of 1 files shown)


QRemy:
Is there still a performance gain for very large maps ?

PR #5844, opened 2025-04-29
Correct small bug in PSFMap

registerrier:
**Description**

This pull request corrects a small bug in PSFMap.

**Dear reviewer**


bkhelifi (approved):
Well seen! Thanks

QRemy (approved):
Thanks @registerrier, from an user perpective there was no bug because the default  is always used in the private method, so it looks more like a cleanup of the internal machinery.

PR #5855, opened 2025-05-08
Utility to filter a table parsing string of  logical operations

QRemy:
The pull request Gammapy PR #5780 introduces a new feature that allows users to filter rows in an astropy.table.Table using logical expressions. The goal is to provide a flexible and expressive way to select subsets of data based on conditions involving table columns.

This PR introduces a function to simplify this operation.
Here's what it does:
- Takes an Astropy Table and a logical expression string.
- Evaluates the expression per row using Python's native syntax
- Returns a filtered table based on the evaluated condition.

It uses AST-based evaluation to parse and evaluate only safe, supported Python expressions.
Supports:
- Logical operators: and, or
- Comparison operators: <, >, ==, !=, in, etc.
- Lists and constants
- Any column in the table

The code, tests, and most of this text are AI generated.

### DIFF
--- a/gammapy/utils/scripts.py
+++ b/gammapy/utils/scripts.py
@@ -291,7 +291,7 @@ def logic_parser(table, expression):
 
     Parameters
     ----------
-    table : astropy.table.Table
+    table : `~astropy.table.Table`
         The input table to filter.
     expression : str
         The logical expression to evaluate on each row. The expression can reference
@@ -299,7 +299,7 @@ def logic_parser(table, expression):
 
     Returns
     -------
-    astropy.table.Table
+    table : `~astropy.table.Table`
         A new table containing only the rows that satisfy the expression. If no rows
         match the condition, an empty table with the same column names and data types
         as the input table is returned.
--- a/gammapy/utils/scripts.py
+++ b/gammapy/utils/scripts.py
@@ -1,6 +1,8 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 """Utilities to create scripts and command-line tools."""
 
+
+import ast
 import codecs
 import os.path
 import functools
@@ -9,6 +11,7 @@ import warnings
 from base64 import urlsafe_b64encode
 from pathlib import Path
 from uuid import uuid4
+from astropy.table import Table
 import yaml
 from gammapy.utils.check import add_checksum, verify_checksum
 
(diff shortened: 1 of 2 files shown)


gammapy/utils/scripts.py
```diff
@@ -214,3 +217,108 @@ def recursive_merge_dicts(a, b):
...
+            elif isinstance(expr.ops[0], ast.GtE):
+                return left >= comparators[0]
+        elif isinstance(expr, ast.Name):
+            return row[expr.id]
+        elif isinstance(expr, ast.Constant):
+            return expr.value
+        elif isinstance(expr, ast.List):
+            return [eval_expr(elt, row) for elt in expr.elts]
+        else:
+            raise ValueError(f"Unsupported expression type: {type(expr)}")
```
bkhelifi, line 310: A small test is missing here...

gammapy/utils/tests/test_scripts.py
```diff
@@ -54,3 +56,86 @@ def test_read_write_yaml_checksum(tmp_path):
...
+    # Create a sample table
+    data = {"OBS_ID": [1, 2, 3, 4], "EVENT_TYPE": ["1", "3", "4", "2"]}
+    table = Table(data)
+
+    # Define an expression with a non-existent key
+    expression = "(NON_EXISTENT_KEY < 3) and (OBS_ID > 1)"
+
+    # Check if KeyError is raised
+    with pytest.raises(KeyError):
+        logic_parser(table, expression)
```
bkhelifi, line 141: Can you add a test on `ValueError(f"Unsupported expression type")`?

bkhelifi (changes requested):
This is a great generic function that will help a lot in the future... and maybe also for the HLI.
Looks OK for me, with the add of 1 test case...

gammapy/utils/tests/test_scripts.py
QRemy: AI : Sure! I've updated the unit tests to use pytest.raises for checking the KeyError and ValueError.

registerrier:
This needs a rebase @QRemy

gammapy/utils/scripts.py
```diff
@@ -214,3 +217,108 @@ def recursive_merge_dicts(a, b):
...
+    parsed_expr = ast.parse(expression, mode="eval").body
+
+    # Filter rows based on the evaluated expression
+    filtered_rows = [row for row in table if eval_expr(parsed_expr, row)]
+
+    # Return a new table with the filtered rows
+    if filtered_rows:
+        return Table(rows=filtered_rows, names=table.colnames)
+    else:
+        return Table(
```
registerrier: Here I would suggest to use `table.__class__` to allow getting back an `ObservationTable` or an `EventList`.

gammapy/utils/scripts.py
```diff
@@ -214,3 +217,108 @@ def recursive_merge_dicts(a, b):
...
+        elif isinstance(expr, ast.List):
+            return [eval_expr(elt, row) for elt in expr.elts]
+        else:
+            raise ValueError(f"Unsupported expression type: {type(expr)}")
+
+    # Parse the expression into an AST
+    parsed_expr = ast.parse(expression, mode="eval").body
+
+    # Filter rows based on the evaluated expression
+    filtered_rows = [row for row in table if eval_expr(parsed_expr, row)]
```
registerrier: does it have to be done on rows? Why not directly on columns? That would be more efficient no?

gammapy/utils/scripts.py
```diff
@@ -214,3 +217,108 @@ def recursive_merge_dicts(a, b):
...
+
+    This function evaluates a logical expression on each row of the input table
+    and returns a new table containing only the rows that satisfy the expression.
+    The expression can reference any column in the table by name and supports
+    logical operators (`and`, `or`), comparison operators (`<`, `>`, `==`, `!=`, `in`),
+    lists, and constants.
+
+    Parameters
+    ----------
+    table : astropy.table.Table
```
registerrier:
```suggestion
    table : `~astropy.table.Table`
```

gammapy/utils/scripts.py
```diff
@@ -214,3 +217,108 @@ def recursive_merge_dicts(a, b):
...
+        The input table to filter.
+    expression : str
+        The logical expression to evaluate on each row. The expression can reference
+        any column in the table by name.
+
+    Returns
+    -------
+    astropy.table.Table
```
registerrier:
```suggestion
    table : `~astropy.table.Table`
```

gammapy/utils/scripts.py
QRemy: now it returns a view

gammapy/utils/scripts.py
QRemy: done

bkhelifi (approved):
All good for me. This is a very useful tool.

Issue #5041, opened 2024-01-24
Create dedicated file for coordinate handling


AtreyeeS:
As discussed in https://github.com/gammapy/gammapy/pull/5015#discussion_r1450056683_ the function `get_fov_coords` should be moved to to either `gammapy.data.pointing`, or to a new module `gammapy.coordinate` which uses astropy Frames to implement the different definitions of FoV Cooridinates

registerrier:
This is solved by #5860  and by #5911 . Closing.

PR #5860, opened 2025-05-12
Introduce a FoVFrame in gammapy.utils.coordinates

registerrier:
**Description**

This pull request tries to solve #5041 by introducing a `FoVFrame` coordinate system based on astropy's `BaseCoordinateFrame`. It follows the `NominalFrame` logic implemented in ctapipe except for the longitude reversal which is done with an additional matrix multiplication (not a rotation matrix here).

I have added some tests.

```
geom = WcsGeom.create(width=1, binsz=0.02)
coords = geom.get_coord().skycoord

now = Time.now() 
location = observatory_locations.get("ctao_south")
origin = AltAz(alt=55*u.deg, az=10.5*u.deg)

frame = FoVFrame(obstime=now, location=location, origin=origin)

# transform to FoV frame
new = coords.transform_to(frame)

# transform back to ICRS
back = new.transform_to("icrs")
```
This yields the same result as the current `sky_to_fov`

**Dear reviewer**

### DIFF
--- a/gammapy/utils/coordinates/fov.py
+++ b/gammapy/utils/coordinates/fov.py
@@ -1,7 +1,91 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
-from astropy.coordinates import SkyCoord, SkyOffsetFrame
+import astropy.units as u
+import numpy as np
+from astropy.coordinates import (
+    AltAz,
+    Angle,
+    BaseCoordinateFrame,
+    CoordinateAttribute,
+    DynamicMatrixTransform,
+    EarthLocationAttribute,
+    FunctionTransform,
+    RepresentationMapping,
+    SkyCoord,
+    SkyOffsetFrame,
+    TimeAttribute,
+    UnitSphericalRepresentation,
+    frame_transform_graph,
+)
+from astropy.coordinates.matrix_utilities import matrix_transpose, rotation_matrix
 
-__all__ = ["fov_to_sky", "sky_to_fov"]
+__all__ = ["FoVFrame", "fov_to_sky", "sky_to_fov"]
+
+reflect_lon_matrix = np.array([[1, 0, 0], [0, -1, 0], [0, 0, 1]])
+
+
+class FoVFrame(BaseCoordinateFrame):
+    """
+    FoV coordinate frame. Centered on `origin` and aligned on AltAz frame at `location` and `obstime`.
+
+    Longitudes are reversed.
+
+    Attributes
+    ----------
+    origin: astropy.coordinates.SkyCoord[AltAz]
+        Origin of this frame as a HorizonCoordinate
+    obstime: astropy.time.Time
+        Observation time
+    location: astropy.coordinates.EarthLocation
+        Location of the telescope
+    """
+
+    frame_specific_representation_info = {
+        UnitSphericalRepresentation: [
+            RepresentationMapping("lon", "fov_lon"),
+            RepresentationMapping("lat", "fov_lat"),
+        ]
+    }
+    default_representation = UnitSphericalRepresentation
+
+    origin = CoordinateAttribute(default=None, frame=AltAz)
+
+    obstime = TimeAttribute(default=None)
+    location = EarthLocationAttribute(default=None)
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        # make sure telescope coordinate is in range [-180°, 180°]
+        if isinstance(self._data, UnitSphericalRepresentation):
+            self._data.lon.wrap_angle = Angle(180, unit=u.deg)
+
+
+@frame_transform_graph.transform(FunctionTransform, FoVFrame, FoVFrame)
+def fov_to_fov(from_fov_coord, to_fov_frame):
+    """Transform between two skyoffset frames."""
+    intermediate_from = from_fov_coord.transform_to(from_fov_coord.origin)
+    intermediate_to = intermediate_from.transform_to(to_fov_frame.origin)
+    return intermediate_to.transform_to(to_fov_frame)
+
+
+@frame_transform_graph.transform(DynamicMatrixTransform, AltAz, FoVFrame)
+def altaz_to_fov(altaz_coord, fov_frame):
+    """Convert a reference coordinate to a sky offset frame."""
+    # Define rotation matrices along the position angle vector, and
+    # relative to the origin.
+    origin = fov_frame.origin.represent_as(UnitSphericalRepresentation)
+    mat1 = rotation_matrix(-origin.lat, "y")
+    mat2 = rotation_matrix(origin.lon, "z")
+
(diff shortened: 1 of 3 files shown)


gammapy/utils/coordinates/fov.py
```diff
@@ -1,7 +1,91 @@
...
+    """
+    FoV coordinate frame. Centered on `origin` and aligned on AltAz frame at `location` and `obstime`.
+
+    Longitudes are reversed.
+
+    Attributes
+    ----------
+    origin: astropy.coordinates.SkyCoord[AltAz]
+        Origin of this frame as a HorizonCoordinate
+    obstime: astropy.time.Time
```
bkhelifi, line 36:
```suggestion
    obstime: `~astropy.time.Time`
```

gammapy/utils/coordinates/fov.py
```diff
@@ -1,7 +1,91 @@
...
+        Observation time
+    location: astropy.coordinates.EarthLocation
```
bkhelifi, line 38:
```suggestion
    location: `~astropy.coordinates.EarthLocation`
```

gammapy/utils/coordinates/fov.py
```diff
@@ -1,7 +1,91 @@
...
+        Location of the telescope
```
bkhelifi, line 39:
```suggestion
        Location of the telescope/instrument/observatory
```

gammapy/utils/coordinates/fov.py
```diff
@@ -1,7 +1,91 @@
...
+        super().__init__(*args, **kwargs)
+
+        # make sure telescope coordinate is in range [-180°, 180°]
+        if isinstance(self._data, UnitSphericalRepresentation):
+            self._data.lon.wrap_angle = Angle(180, unit=u.deg)
+
+
+@frame_transform_graph.transform(FunctionTransform, FoVFrame, FoVFrame)
+def fov_to_fov(from_fov_coord, to_fov_frame):
+    """Transform between two skyoffset frames."""
```
bkhelifi, line 65:
```suggestion
    """Transform between two `~astropy.coordinates.SkyOffsetFrame` object."""
```

gammapy/utils/coordinates/fov.py
```diff
@@ -1,7 +1,91 @@
...
+    intermediate_from = from_fov_coord.transform_to(from_fov_coord.origin)
+    intermediate_to = intermediate_from.transform_to(to_fov_frame.origin)
+    return intermediate_to.transform_to(to_fov_frame)
+
+
+@frame_transform_graph.transform(DynamicMatrixTransform, AltAz, FoVFrame)
+def altaz_to_fov(altaz_coord, fov_frame):
```
bkhelifi, line 72: I do not understand this function... Beyond the fact that there is no docstring and I do not know what it returns, the argument `altaz_coord` is never used!

gammapy/utils/coordinates/fov.py
```diff
@@ -1,7 +1,91 @@
...
+from astropy.coordinates.matrix_utilities import matrix_transpose, rotation_matrix
 
-__all__ = ["fov_to_sky", "sky_to_fov"]
+__all__ = ["FoVFrame", "fov_to_sky", "sky_to_fov"]
+
+reflect_lon_matrix = np.array([[1, 0, 0], [0, -1, 0], [0, 0, 1]])
+
+
+class FoVFrame(BaseCoordinateFrame):
+    """
```
bkhelifi, line 27: In the final version, it would be very nice to have  in the documentation a picture of the alignment of the different frames ;)

gammapy/utils/coordinates/fov.py
maxnoe: See https://docs.astropy.org/en/stable/coordinates/frames.html#defining-a-new-frame

gammapy/utils/coordinates/fov.py
```diff
@@ -1,7 +1,91 @@
...
+    mat2 = rotation_matrix(origin.lon, "z")
+
+    return reflect_lon_matrix @ mat1 @ mat2
+
+
+@frame_transform_graph.transform(DynamicMatrixTransform, FoVFrame, AltAz)
+def fov_to_altaz(fov_coord, altaz_frame):
+    """Convert an sky offset frame coordinate to the reference frame"""
+    # use the forward transform, but just invert it
+    mat = altaz_to_fov(altaz_frame, fov_coord)
```
bkhelifi, line 87: the first argument that is passed in a frame, where as in the function arguments the first is a coordinate. It is very confusing

maxnoe: The comment above explains it. We use the inverse direction to compute the matrix, than invert the matrix.

gammapy/utils/coordinates/fov.py
maxnoe: Please read the above documentation. All functions in the transformation graph have the signature:

```
def transform(from_coord, to_frame):
   ...
```

maxnoe: This function will be called internally when, in this case:

```
altaz_coord = SkyCoord(..., frame=AltAz(...))
fov_coord = altaz_coord.transform_to(FoVFrame(...))
```

and it will always pass the two arguments.

gammapy/utils/coordinates/fov.py
```diff
@@ -1,7 +1,91 @@
...
+)
+from astropy.coordinates.matrix_utilities import matrix_transpose, rotation_matrix
 
-__all__ = ["fov_to_sky", "sky_to_fov"]
+__all__ = ["FoVFrame", "fov_to_sky", "sky_to_fov"]
+
+reflect_lon_matrix = np.array([[1, 0, 0], [0, -1, 0], [0, 0, 1]])
+
+
+class FoVFrame(BaseCoordinateFrame):
```
maxnoe: Shouldn't this be called `FoVAltAzFrame`? And we should also have one aligned with ICRS?

gammapy/utils/coordinates/fov.py
registerrier: I propose to leave this for another PR once this frame is used in the code base

gammapy/utils/coordinates/fov.py
registerrier: Indeed @maxnoe ,  an additional frame would be needed to ICRS aligned FoV coordinates. `FoVAltAzFrame` and `FoVICRSFrame` might be clear enough names even if hard to read.

Are there other suggestions for simpler names?

registerrier:
If there are no additional comments I will merge this PR later today.

AtreyeeS (dismissed):
Thanks @registerrier ! I agree that it would be nice to rename `FoVFrame` to `FoVAltAzFrame` before merging.
Once this is in, we can cleanup the 3D irf handling

registerrier:
Merging

Issue #5540, opened 2024-11-05
Class attributes names should be checked at setting time


HealthyPear:
**Is your feature request related to a problem? Please describe.**

After creating a SpectrumDataset instance if one does by mistake 

`spectrum_dataset.model = bla `

instead of 

`spectrum_dataset.models = bla` 

_gammapy_ does not issue any error, instead lets the user go on with the analysis and obtain incorrect results.

(yes, there is the documentation, but in this case the absence of an error message does not help the user pinpoint _where_ to go in the docs to check what's wrong)

**Describe the solution you'd like**

There should be a check (throughout all _gammapy_, not only this example case) where if a user tries to set or get a non-existent class attribute a validation error should be raised.

I see that _gammapy_ uses _pydantic_ here and there, so I guess this could be the easiest solution, but I never used it myself.

adonath:
Thanks @HealthyPear, I am closing this as a duplicate of https://github.com/gammapy/gammapy/issues/3225. Please comment there if you have additional thoughts.

Issue #3225, opened 2021-02-05
Forbid setting of new attributes on Dataset


registerrier:
I have noticed several cases where users get confused between `mask_fit` and `fit_mask` and when they set `dataset.fit_mask = some_mask` nothing is there to tell us that they are adding a new attribute on their object that will never be taken into account.

 Since this error is relatively easy to make, should we consider forbidding setting any attributes not present in the `__init__`?
Or at least specifically the `fit_mask` and possibly `safe_mask` attributes.

adonath:
I can see the motivation for this, however I think it's not a very common pattern in Python and technically not that simple to achieve (see https://stackoverflow.com/questions/3603502/prevent-creating-new-attributes-outside-init), the consequence would be to add a something `__frozen = True` flag to all out `__init__` methods in Gammapy.

Any main dependency we have (Astropy, Numpy, Scipy) can be broken like this:
```python3
from astropy.coordinates import SkyCoord

coord = SkyCoord("0d", "0d")
coord.lon = "1 deg"
```
And my answer would be, that this is just how Python works and we have to live with it.

I'm definitely against adding checks for specific attributes. If we introduce a check for `fit_mask` or `safe_mask`, why not `off_counts` or `off_acceptance`? Once we start doing something like this doesn't end...

I think the only reasonable option here is find a technical solution, that prevents setting new attributes after init on a  meta class, that all our classes can inherit from.

adonath:
One more thought: in general we have a solution for this with the pydantic `BaseModel`, where attributes are not declared in `__init__` but using class level decorators. See e.g.:
```python3
from gammapy.analysis.config import SkyCoordConfig

a = SkyCoordConfig()
a.xyz = "xyz"
```
Which raises a `ValueError`. I think in general this could go into the right direction for configurable classes such as `Maker` and `Estimator` and possibly the `Dataset` classes as well.

adonath:
Pydantic the also offers type validation and serialisation...

adonath:
I think the best candidate for a solution here is the pydantic approach, however this is not in scope for v1.0 will relabel now...

registerrier:
In issue #5540, @HealthyPear  is discussing the case of `dataset.model` instead of `dataset.models` and using pydantic to check it.

While I am not sure we can make most Gammapy objects pydantic `BaseModel`, we might try to achieve this with the `Dataset`. Any prototype for this would be interesting to see.

registerrier:
[...]

PR #5865, opened 2025-05-12
Restrict attribute settings on `Dataset` classes

registerrier:
**Description**

This pull request is a proposal for solving #3225 and #5540 for the specific case of `Dataset` classes. It restricts the list of possible attributes sets on a given `Dataset` class.
For instance: `dataset.model = sky_model` or `dataset.fit_mask = mask` or `dataset.whatever=3` will yield an `AttributeError`. But not `dataset._whatever=3`.

To do so a class attribute listing valid attributes is defined for each class and the `__setattr__` method is overloaded to check in the MRO the valid attributes. This introduces probably a performance penalty, but this is likely not an issue here.

This solution does not rely on pydantic as proposed in #3225. It is a bit overkill since the objective is not input data validation. I also think it is not adapted to the complex nature of the `Dataset` classes with its class level descriptors such as `LazyFitsData`. Adapting this would require large refactoring. Additionally documentation produced for pydantic BaseModel derived classes is hard to read and I would avoid this for something like `Dataset`. 

Note that an approach based on restricting `__slots__` does not work because of the class level descriptors.

**Dear reviewer**

Opinions @adonath  @QRemy @HealthyPear @AtreyeeS @bkhelifi ? 

Is this behaviour desirable ? Does this approach seem reasonable?

### DIFF
--- a/docs/development/dev_howto.rst
+++ b/docs/development/dev_howto.rst
@@ -668,6 +668,23 @@ If you have a alternative attribute to use instead, pass its name in the `altern
     print(some_class(10).old_attribute)
 
 
+How to capture a deprecation
+++++++++++++++++++++++++++++
+
+A deprecation warning raised during CI will be considered an error by `pytest`.
+Nevertheless, it is important to also check that deprecation warnings are correctly
+emitted in the test files. This can be done in the following way:
+
+.. code::
+
+    # Runs with the renamed argument
+    asmooth = ASmoothMapEstimator(scales=scales, spectral_model=PowerLawSpectralModel())
+
+    # Raises deprecation warning
+    with pytest.warns(GammapyDeprecationWarning):
+        ASmoothMapEstimator(scales=scales, spectrum=PowerLawSpectralModel())
+
+
 Others
 ------
 
--- a/docs/development/pigs/index.rst
+++ b/docs/development/pigs/index.rst
@@ -32,6 +32,7 @@ label`_ .
     pig-013
     pig-014
     pig-016
+    pig-017
     pig-018
     pig-019
     pig-020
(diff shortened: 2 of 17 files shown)


AtreyeeS (dismissed):
Thank you @registerrier ! This indeed looks nice, and is simpler than the pydantic approach

Users can still set invalid values, eg: `dataset.models = 3`,no? But I think that's not a problem.

QRemy (approved):
Thanks @registerrier, should this be also applied to other classes like parameters and models ?

registerrier:
The implementation is modified to rely on a decorator and an automated determination of `_allowed_attrs`. 
This simplifies the main code with the simple decorator. But it requires a relatively complex implementation for the decorator.

adonath:
I think Sherpa has something similar: https://github.com/sherpa/sherpa/blob/12441d6fca063d92c42967fab52dd6f57ea302b9/sherpa/utils/__init__.py#L108

Otherwise something like this is very un-common in Python, because of the "We are all consenting adults here" paradigm. We can add this, but it is merely helping, because it can still go wrong in many other places and especially also on 3rd party code (astropy, numpy, matplotlib etc.). But I agree that if there is a place where to add it, then it would be the `Dataset`, which is the class most users interact with the most. 

My tendency would be to say "that's just how Python works" and do nothing. If others think it is a fair benefit to beginners and will safe them a little bit of headache in some scenarios, OK. However we should make sure it does not introduce any unwanted side-effects (but I am confident those would be covered by our tests).

AtreyeeS:
What is the status here @registerrier ? Do we put this in for 2.0?

registerrier:
As the gain of this was not very clear I propose to close for now.

PR #5864, opened 2025-05-12
Remove v1.3 deprecations 

registerrier:
**Description**

This pull request removes deprecations from v1.3 and the associated tests. 
The only things untouched in this PR are deprecations in gammapy.data which require dedicated PRs.

**Dear reviewer**

PR #5786, opened 2025-04-09
More robust  error propagation approach using sampling

QRemy:
More robust error propagation approach using sampling
Fixes #5738

PR #5866, opened 2025-05-13
Add missing import  after #5864

QRemy:
Add missing import  after #5864


registerrier:
This is caused by the successive merging of 
- #5786 which introduced Deprecation of some features in gammapy.modeling.models.spectral.py 
- #5864 which removed the earlier deprecations
Somehow the conflict went unnoticed.

PR #5867, opened 2025-05-13
Update nested_sampling_Crab.py

facero:
fixing broken url for citation (extra white spaces)


QRemy (approved):
Tahnks @facero, looks good, I will merge it directly if you don't want to make other additions.

PR #5873, opened 2025-05-15
Fix link of doc licences in README.rst

morcuended:
There is a missing space between the shown text and the link that makes it not being properly displayed (see current README page).


Astro-Kirsty (approved):
Thanks @morcuended 

Just a note, that the `skip` labels do not work currently (they haven't been set up as far as im aware!) It appears that the CI etc is running twice here instead :'(

bkhelifi (approved):
Thanks a lot for the fix!

PR #5879, opened 2025-05-22
Fix date keywords in DataStore

fabiopintore:
`DataStore` does not read correctly the keywords `DATE-OBS` and `DATE-END`, and `TIME-OBS` and `TIME-END` when loading observations from event files.
This PR fixes this issue.


fabiopintore:
Thanks @Astro-Kirsty .
According to the GADF, the expect keyword is `DATE-OBS`. Should we allow for both options?

bkhelifi:
Well seen, @fabiopintore . The FITS standards is using `-` and not `_`. So, your changes are necessary.
However, as suspected, we did use in HESS for the old productions the `_`. So maybe the two should we added. And I would suggest to add a ToDo to precise that the future I/O scheme should handle that depending on the format version (as a reminder)

fabiopintore:
Thanks for the comment. I've implemented the possibility to take care of both keywords syntax. I've also fixed the test.

QRemy (dismissed):
Thanks @fabiopintore, I modified the test to directly test read_events_info as its name suggests and added a test for the HESS case. Otherwise good to merge.

QRemy (approved):

PR #5880, opened 2025-05-22
Serialization of GTI extension is not correct in the event sampling tutorial

fabiopintore:
The event sampling tutorial shows how to save the events into a fits file. A GTI extension is also included. However, the current tutorial produces a GTI extension which is badly formatted and useless for any temporal analysis of the data. This PR fixes this issue in the tutorial by creating the GTI extension directly from `dataset.gti.to_table_hdu`.


registerrier (approved):
Thanks @fabiopintore . Looks good to me.

PR #5881, opened 2025-05-22
A class for vectorized evaluation of flux uncertainties based on samples

registerrier:
**Description**

This pull request introduces `FluxPredictionBand` which is designed to perform vectorized evaluation of flux uncertainties based on samples. This provides a significantly faster evaluation than the current one (2 to 3 order of magnitude). 
It should support both samples generated from covariance matrix as well as samples generated via nested sampling.

This is still a draft.

**Dear reviewer**

### DIFF
--- a/gammapy/modeling/models/spectral.py
+++ b/gammapy/modeling/models/spectral.py
@@ -100,11 +100,13 @@ def scale_plot_flux(flux, energy_power=0):
 
 
 def integrate_spectrum(
-    func, energy_min, energy_max, ndecade=100, parameter_samples=None
+    func, energy_min, energy_max, ndecade=100, parameter_samples=None, eflux=False
 ):
     """Integrate one-dimensional function using the log-log trapezoidal rule.
 
     Internally an oversampling of the energy bins to "ndecade" is used.
+
+    To compute the integral func(E)*E (e.g. to get an energy flux) one can set eflux to True.
 
     Parameters
     ----------
@@ -121,6 +123,9 @@ def integrate_spectrum(
         Dictionary of parameter quantities to pass to `func.evaluate` if it exists.
         This provides vectorized integral evaluation for parameter samples.
         If None, evaluation is performed with a simple call to `func`. Default is None.
+    eflux : bool, optional
+        If True, the function will integrate func(E)*E, to compute the energy flux.
+        Default is False.
     """
     # Here we impose to duplicate the number
     num = np.maximum(np.max(ndecade * np.log10(energy_max / energy_min)), 2)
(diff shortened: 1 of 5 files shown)


gammapy/modeling/models/spectral.py
```diff
@@ -99,27 +99,62 @@ def scale_plot_flux(flux, energy_power=0):
...
 
-def integrate_spectrum(func, energy_min, energy_max, ndecade=100):
+def integrate_spectrum(
+    func, energy_min, energy_max, ndecade=100, parameter_samples=None, eflux=False
+):
     """Integrate one-dimensional function using the log-log trapezoidal rule.
 
     Internally an oversampling of the energy bins to "ndecade" is used.
 
+    To compute the integral func(E)*E (e.g. to get an energy flux) one can set eflux to True.
```
AtreyeeS, line 109: For `SpectralModel.plot()` we have `sed_type` and `energy_power`. Maybe it is good to be consistent instead  of having `eflux=True` option here

gammapy/modeling/models/spectral.py
```diff
@@ -99,27 +99,62 @@ def scale_plot_flux(flux, energy_power=0):
...
     ndecade : int, optional
         Number of grid points per decade used for the integration.
         Default is 100.
+    parameter_samples : dict, optional
+        Dictionary of parameter quantities to pass to `func.evaluate` if it exists.
+        This provides vectorized integral evaluation for parameter samples.
+        If None, evaluation is performed with a simple call to `func`. Default is None.
+    eflux : bool, optional
+        If True, the function will integrate func(E)*E, to compute the energy flux.
+        Default is False.
```
AtreyeeS, line 128: Can you add an example in the docstring? It is difficult to imagine how to use the `parameter_samples`

gammapy/utils/integrate.py
```diff
@@ -27,6 +27,7 @@ def trapz_loglog(y, x, axis=-1):
     from gammapy.modeling.models import PowerLawSpectralModel as pl
 
     # see https://stackoverflow.com/a/56840428
+    # move integration axis from last position to first
```
AtreyeeS, line 30: What is the change here?

AtreyeeS (dismissed):
Thank you @registerrier ! Since the samples are used only for the error calculation, I agree it is a bit odd to store `samples` on. the `models` object. I left some small comments inline, but it is okay from my side now.
Do you want to expose in the public API this for 2.0 ?

facero:
Note:
when this is merged I will update the NS tutorial so that we can show the enveloppe of the spectral model.

bkhelifi:
If it goes into the 2.0, one should need to expose its usage into a tutorial.

QRemy:
The vectorized evaluation of flux uncertainties is now implemented by #6038 without refactoring. For 2.0 the performance  improvement is the priority, the refactoring could be rediscussed later so I will relabel this one as prototype and move the milestone to wishlist.

QRemy:
Closing this for now as it would require quite a lot refactoring and it's not a priority.

PR #5884, opened 2025-05-27
Rm setting log levels in modules

PreisTo:
**Description**

This pull request removes two statements setting the logging level to INFO in gammapy.data.data_store as well as gammapy.data.ivoa.

**Dear reviewer**
Stumbled upon this as it overwrote my logging configuration in a ipynb.

I see no reason why the logging level should be set to INFO when importing those modules, if there is I am happy to close the PR :)

### DIFF
--- a/gammapy/modeling/models/tests/test_io.py
+++ b/gammapy/modeling/models/tests/test_io.py
@@ -26,8 +26,9 @@ from gammapy.modeling.models import (
     TemplateNPredModel,
 )
 from gammapy.utils.scripts import make_path, read_yaml, to_yaml, write_yaml
-from gammapy.utils.testing import requires_data, requires_dependency
+from gammapy.utils.testing import requires_data
 import os
+import importlib
 
 
 @pytest.fixture(scope="session")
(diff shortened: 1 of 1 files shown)


bkhelifi:
Can you explain why there is no reason to set it at INFO? Because it is by default, or because the given information are not necessary?

PreisTo:
> Can you explain why there is no reason to set it at INFO? Because it is by default, or because the given information are not necessary?

@bkhelifi  thanks for your quick reply!
My point is more that the displayed log level should be up to the user and manually setting it when loading a module may tamper with the user's configuration. With setting it to INFO also the default of WARNING gets overwritten.

These are the only two loggers in gammapy for which the logging level gets set explicitly, which seems a little bit inconsistent in my opinion. This also disables the [level propagation](https://docs.python.org/3/howto/logging.html#loggers) from parent loggers, i.e. by setting the root logging level to e.g.

When running 
```python
import logging
from gammapy.data.data_store import log as log_data
from gammapy.data.event_list import log as log_event

logging.basicConfig(level=logging.DEBUG)
logger = logging.getLogger()

logger.debug("Debug from root")
log_data.debug("Debug from data store")
log_event.debug("Debug from event list")
```
you get the following output:
```bash
DEBUG:root:Debug from root
DEBUG:gammapy.data.event_list:Debug from event list
```

adonath:
Thanks @PreisTo, indeed logging level is user configuration. There can be a single root level default in Gammapy, but I agree that module level is not the correct thing to do.

morcuended (approved):
Agreed, thanks @PreisTo

registerrier (approved):
Thanks @PreisTo . Looks good to me.

PR #5888, opened 2025-05-30
Fix gammapy.catalog to support astropy v7.1

QRemy:
In astropy v7.1 empty field appears as MaskedConstant where previously it would be an empty string.
As we often used .strip on the output of the table, I introduced a  GammapyMaskedConstant class that inherit from MaskedConstant and simply add a.strip() method returning str(self).

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -37,7 +37,7 @@ setup_requires = setuptools_scm
 install_requires =
     numpy>=1.21
     scipy>=1.5,!=1.10
-    astropy>=5.0, <7.1
+    astropy>=5.0
     regions>=0.5.0
     pyyaml>=5.3
     click>=7.0


QRemy:
I had to make more modifications to uniformise the empty entry printout and have this passing all CIs, but it's finally working.

gammapy/utils/table.py
```diff
@@ -13,6 +15,11 @@
 ]
 
 
+class GammapyMaskedConstant(MaskedConstant):
```
registerrier: Add a docstring to explain why this is needed.

gammapy/utils/table.py
```diff
@@ -86,6 +93,9 @@ def table_row_to_dict(row, make_quantity=True):
     for name, col in row.columns.items():
         val = row[name]
 
+        if isinstance(val, np.ma.core.MaskedConstant):
+            val = GammapyMaskedConstant()
```
registerrier: Isn't it simpler to write:
`val = str(MaskedConstant())`
instead of using the overloaded class?

registerrier (commented):
Thanks @QRemy . I have left a few suggestions inline. 

- Is the new class necessary?
- why are there changes in the output format of the catalog string output? e.g. units ordering?

gammapy/utils/table.py
QRemy: done (required also a little change in the hgps catalog object to work)

gammapy/utils/table.py
QRemy: removed

QRemy:
> * Is the new class necessary?

no, removed

> * why are there changes in the output format of the catalog string output? e.g. units ordering?

-empty fields becomes "--"
-for the units seems to be like this since astropy 5.3 and the tests files were generated before. 
Screenshot 2025-06-05 at 15 07 11

QRemy:
> > * Is the new class necessary?
> 
> no, removed
> 
> > * why are there changes in the output format of the catalog string output? e.g. units ordering?
> 
> -in empty field becomes "--" -for the units seems to be like this since astropy 5.3 and the test files were generated before

gammapy/catalog/fermi.py
```diff
@@ -42,6 +42,18 @@
...
 
 
+def get_nonentry_keys(d, keys):
+    vals = [str(d[_]).strip() for _ in keys]
+    return ", ".join([_ for _ in vals if _ not in ["", "--"]])
+
+
+def get_nonentry_key(key):
+    if key.strip() == "":
+        return "--"
```
AtreyeeS, line 52: Why are you returning `--` and not `""` directly?

QRemy: that's what the MaskedConstant returns when type is not str

PR #5889, opened 2025-06-02
Adapt intructions for static figures

### LINKED ISSUE #5811: Docs: static figures in GAMMAPY-data and `gp-image::`  ?
Opening this issue as reminder for the docs.

In trying to add a static image to the tutorial I'm preparing I found this information in the dev doc which says that the image can be stored in GAMMAPY-DATA  and one can use the `gp-image::`  directive.
https://docs.gammapy.org/dev/development/doc_howto.html#include-png-files-as-images

However, besides the doc example I found no document using this directive in the docs. So it seems it has never been used ?
The simplest way seems to just use `image::`  and put the images in docs/_static ?

If agreed then the corresponding paragraph about `gp-image::`  could be in the doc could be removed.

Then it remains to be seen where they are still used and what to do with the images in gammapy-data : 
https://github.com/gammapy/gammapy-data/tree/main/figures

Astro-Kirsty:
This PR is to resolve #5811

I made a comment there regarding what to do with the unused figures in `gammapy-data`

### DIFF
--- a/docs/development/doc_howto.rst
+++ b/docs/development/doc_howto.rst
@@ -42,13 +42,13 @@ Check Python code
 Code in RST files
 +++++++++++++++++
 
-Most of the documentation of Gammapy is present in RST files that are converted into HTML pages using
-Sphinx during the build documentation process. You may include snippets of Python code in these RST files
-within blocks labelled with ``.. code-block:: python`` Sphinx directive. However, this code could not be
-tested, and it will not be possible to know if it fails in following versions of Gammapy. That's why we
-recommend using the ``.. testcode::`` directive to enclose code that will be tested against the results
-present in a block labelled with ``.. testoutput::`` directive. If not ``.. testoutput::`` directive is provided,
-only execution tests will be performed.
+Most of the documentation of Gammapy is present in `restructured text (RST)`_ files that are converted
+into HTML pages using Sphinx during the build documentation process. You may include snippets of Python
+code in these RST files within blocks labelled with ``.. code-block:: python`` Sphinx directive. However,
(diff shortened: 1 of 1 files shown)


docs/development/doc_howto.rst
```diff
@@ -293,24 +293,32 @@ Include png files as images
 In the RST files
 ++++++++++++++++
 
-Gammapy has a ``gp-image`` directive to include an image from ``$GAMMAPY_DATA/figures/``,
-use the ``gp-image`` directive instead of the usual Sphinx ``image`` directive like this:
+To include an image is your rst file first add it to the `docs/_static` folder.
```
bkhelifi:
```suggestion
To include an image in your rst file, first add it to the `docs/_static` folder.
```

bkhelifi (approved):
Thank you, ok for me. I have just a suggestion about a typo

bkhelifi (approved):
Thank you @Astro-Kirsty

facero:
Thanks Kirsty !

Should we explain the difference between `image:: ` and `figure:: ` ?
or maybe explain only one of the two whichever is more relevant ?

Astro-Kirsty:
> Should we explain the difference between `image:: ` and `figure:: ` ? or maybe explain only one of the two whichever is more relevant ?

I decided to leave the explanation minimal as the link gives further in depth information. In fact, for our simple use cases it doesn't really make a different to use `image` or `figure`. `figure` just allows a caption

facero:
Maybe you can just say that `figure` allows for caption using the `:alt: ` directive for example.

PR #5895, opened 2025-06-03
CI fix : no yield in tests

QRemy:
Fix CI error :
`ERROR ../../.tox/py311-test/lib/python3.11/site-packages/gammapy/modeling/models/tests/test_io.py - Failed: 'yield' keyword is allowed in fixtures, but not in tests (test_naima_model)`

### DIFF
--- a/gammapy/modeling/models/tests/test_io.py
+++ b/gammapy/modeling/models/tests/test_io.py
@@ -320,9 +320,7 @@ def test_naima_model():
         amplitude=2e33 / u.eV, e_0=10 * u.TeV, alpha=2.5
     )
     radiative_model = naima.radiative.PionDecay(particle_distribution, nh=1 * u.cm**-3)
-    yield Model.create(
-        "NaimaSpectralModel", "spectral", radiative_model=radiative_model
-    )
+    Model.create("NaimaSpectralModel", "spectral", radiative_model=radiative_model)
 
 
 def make_all_models():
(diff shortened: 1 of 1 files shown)


gammapy/modeling/models/tests/test_io.py
```diff
@@ -320,9 +320,7 @@ def test_naima_model():
         amplitude=2e33 / u.eV, e_0=10 * u.TeV, alpha=2.5
     )
     radiative_model = naima.radiative.PionDecay(particle_distribution, nh=1 * u.cm**-3)
-    yield Model.create(
-        "NaimaSpectralModel", "spectral", radiative_model=radiative_model
-    )
+    Model.create("NaimaSpectralModel", "spectral", radiative_model=radiative_model)
```
morcuended: maybe also add an assertion statement here?

morcuended (commented):
Why was this triggered only now? Wasn't it giving troubles before? I see also yield is being used in the next function test

registerrier:
This used to be a warning in the tests and the test was skipped. 
```
.tox\py311-test-alldeps_noray\Lib\site-packages\gammapy\modeling\models\tests\test_io.py:316: PytestCollectionWarning: yield tests were removed in pytest 4.0 - test_naima_model will be ignored
```
See e.g. https://github.com/gammapy/gammapy/actions/runs/14858528026/job/41717430565?pr=5833#step:8:187

It now fails with pytest 8.4
https://docs.pytest.org/en/stable/changelog.html#removals-and-backward-incompatible-breaking-changes

morcuended:
Shouldn't this `make_all_models` be tagged as a fixture then?

QRemy:
Going to merge if no objections.

PR #5897, opened 2025-06-03
Examples for estimator utils

Astro-Kirsty:
As per title

### DIFF
--- a/gammapy/estimators/points/sed.py
+++ b/gammapy/estimators/points/sed.py
@@ -85,6 +85,29 @@ class FluxPointsEstimator(FluxEstimator, parallel.ParallelMixin):
     Notes
     -----
     For further explanation, see :ref:`estimators`.
+
+    Examples
+    --------
+    .. testcode::
+
+        from astropy import units as u
+        from gammapy.datasets import SpectrumDatasetOnOff
+        from gammapy.estimators import FluxPointsEstimator
+        from gammapy.modeling.models import PowerLawSpectralModel, SkyModel
+
+        path = "$GAMMAPY_DATA/joint-crab/spectra/hess/"
+        dataset = SpectrumDatasetOnOff.read(path + "pha_obs23523.fits")
+
+        pwl = PowerLawSpectralModel(index=2, amplitude='1e-12  cm-2 s-1 TeV-1')
+
+        dataset.models = SkyModel(spectral_model=pwl, name="crab")
+
+        estimator = FluxPointsEstimator(
+            source="crab",
+            energy_edges=[0.1, 0.3, 1, 3, 10, 30, 100] * u.TeV,
+        )
+
+        fp = estimator.run(dataset)
     """
 
     tag = "FluxPointsEstimator"
--- a/gammapy/estimators/utils.py
+++ b/gammapy/estimators/utils.py
@@ -262,6 +262,26 @@ def estimate_exposure_reco_energy(dataset, spectral_model=None, normalize=True):
     -------
     exposure : `~gammapy.maps.Map`
         Exposure map in reconstructed energy.
+
+    Examples
+    --------
+    .. testcode::
+
+        from gammapy.datasets import MapDataset
+        from gammapy.estimators import ExcessMapEstimator
+        from gammapy.estimators.utils import estimate_exposure_reco_energy
+        from gammapy.modeling.models import PowerLawSpectralModel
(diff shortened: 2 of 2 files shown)


bkhelifi (approved):
Thanks a million

gammapy/estimators/utils.py
```diff
@@ -262,6 +262,26 @@ def estimate_exposure_reco_energy(dataset, spectral_model=None, normalize=True):
...
+        from gammapy.estimators.utils import estimate_exposure_reco_energy
+        from gammapy.modeling.models import PowerLawSpectralModel
+        dataset = MapDataset.read("$GAMMAPY_DATA/cta-1dc-gc/cta-1dc-gc.fits.gz")
+        spectral_model = PowerLawSpectralModel(index=15)
+        estimator = ExcessMapEstimator(
+            correlate_off=True,
+            spectral_model=spectral_model,
+        )
+        result = estimator.run(dataset)
+        reco_exposure = estimate_exposure_reco_energy(
```
registerrier, line 275: Why do you expose this?

gammapy/estimators/points/sed.py
```diff
@@ -85,6 +85,29 @@ class FluxPointsEstimator(FluxEstimator, parallel.ParallelMixin):
...
+        from gammapy.datasets import SpectrumDatasetOnOff
+        from gammapy.estimators import FluxPointsEstimator
+        from gammapy.modeling.models import PowerLawSpectralModel, SkyModel
+
+        path = "$GAMMAPY_DATA/joint-crab/spectra/hess/"
+        dataset = SpectrumDatasetOnOff.read(path + "pha_obs23523.fits")
+
+        pwl = PowerLawSpectralModel(index=2, amplitude='1e-12  cm-2 s-1 TeV-1')
+
+        dataset.models = SkyModel(spectral_model=pwl, name="crab")
```
registerrier, line 104: I have an issue with this example because it uses a spectrum that is very different from the best fit one on this dataset. 
While it might be working fine here, I don't think it will always be the case.

Astro-Kirsty: Ahh okay, I just copied what was in the test. I can adapt it

gammapy/estimators/utils.py
Astro-Kirsty: Is there a reason we don't want to expose it?

registerrier: The `ExcessMapEstimator` already computes the reco energy exposure. I don't really see the point of the example shown this way.

gammapy/estimators/utils.py
```diff
@@ -262,6 +262,26 @@ def estimate_exposure_reco_energy(dataset, spectral_model=None, normalize=True):
...
+    Examples
+    --------
+    .. testcode::
+
+        from gammapy.datasets import MapDataset
+        from gammapy.estimators import ExcessMapEstimator
+        from gammapy.estimators.utils import estimate_exposure_reco_energy
+        from gammapy.modeling.models import PowerLawSpectralModel
+        dataset = MapDataset.read("$GAMMAPY_DATA/cta-1dc-gc/cta-1dc-gc.fits.gz")
+        spectral_model = PowerLawSpectralModel(index=15)
```
registerrier: index 15?

gammapy/estimators/utils.py
registerrier: The question is whether this is an implementation detail or a real part of the user API.

If we want to expose it then remove the call to the ExcessMapEstimator. It is useless here and probably confusing because it suggest you need to run it to get the reco energy exposure.

registerrier (commented):
Thanks @Astro-Kirsty . I left some replies/comments inline

gammapy/estimators/utils.py
Astro-Kirsty: Oops, that was a mistake

AtreyeeS (dismissed):
Thanks @Astro-Kirsty ! Usually, I find it nice if  some expectation from the final output can also be present, eg: `fp.plot` or `print(fp)` etc

Astro-Kirsty:
> Thanks @Astro-Kirsty ! Usually, I find it nice if some expectation from the final output can also be present, eg: `fp.plot` or `print(fp)` etc

I added a few, but we decided a while ago to not add too many of these in case formatting etc changes.

registerrier (approved):
Thanks @Astro-Kirsty . Looks good to me. No further comment.

PR #5898, opened 2025-06-03
Add stacking functionality to `LightCurveEstimator`

### LINKED ISSUE #5840: LightCurveEstimator ... stack fitting within a bin instead of joint fitting datasets?
Even though probably a joint fitting is more accurate (IRFs may change from run to run), when you have super long datasets the number of datasets explodes and the current implementation of LightCurveEstimator (which seem to select datasets contained in a bin, and then join fits them) struggles with memory, cpu time, etc.

Probably it would be good to have a switch in the Estimator to indicate that datasets/runs belonging to a time-bin should be pre-stacked, and then the light curve calculated over these intermediate stacks. 

Opinions?

Astro-Kirsty:
This PR is an idea for to resolve #5840

### DIFF
--- a/.github/workflows/linkcheck.yml
+++ b/.github/workflows/linkcheck.yml
@@ -16,6 +16,23 @@ jobs:
       PYTEST_ADDOPTS: --color=yes -n auto --dist=loadscope
       GAMMAPY_DATA: /home/runner/work/gammapy/gammapy/gammapy-datasets/dev
     steps:
+      - name: Check out repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: Set up Python 3.11
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.11'
+      - name: Install base dependencies
+        run: |
+          python -m pip install --upgrade pip
+          python -m pip install tox
+      - name: download datasets
+        run: |
+          python -m pip install tqdm requests
+          python -m pip install -e .
+          gammapy download datasets
       - name: check links
         continue-on-error: true
         run: |
--- a/docs/development/dev_howto.rst
+++ b/docs/development/dev_howto.rst
@@ -9,10 +9,13 @@ Developer How To
 General conventions
 -------------------
 
-Python version support
-++++++++++++++++++++++
+Python and NumPy version support
+++++++++++++++++++++++++++++++++
 
-In Gammapy we currently support Python 3.9 or later.
+Gammapy follows the `NEP 29 guidelines <https://numpy.org/neps/nep-0029-deprecation_policy.html>`__
+for version support, which implements a time-based deprecation policy for
+Python and NumPy versions. This ensures compatibility with actively supported
+versions while maintaining stability.
 
 Coordinate and axis names
 +++++++++++++++++++++++++
--- a/docs/sphinxext.py
+++ b/docs/sphinxext.py
@@ -34,6 +34,7 @@ TUTORIAL_SORT_DICT = {
         "fitting.py": 11,
         "nested_sampling_Crab.py": 12,
         "estimators.py": 13,
+        "parameter_ul.py": 14,
     #  data analysis
     #  1d
         "cta_sensitivity.py": 0,
--- a/docs/user-guide/astro/population/index.rst
+++ b/docs/user-guide/astro/population/index.rst
@@ -48,7 +48,7 @@ surface density of pulsars and related objects used in literature:
 
 .. plot:: user-guide/astro/population/plot_radial_distributions.py
 
-TODO: add illustration of Galactocentric z-distribution model and combined (r,
+.. TODO: add illustration of Galactocentric z-distribution model and combined (r,
 z) distribution for the Besancon model.
 
 Spiral arm models
--- a/docs/user-guide/astro/source/index.rst
+++ b/docs/user-guide/astro/source/index.rst
@@ -10,10 +10,7 @@ Introduction
 The `gammapy.astro.source` module contains classes of source models, which can
 be used for population synthesis of galactic gamma-ray sources.
 
-Getting started
-===============
-
-TODO: add basic example.
+.. TODO: add basic example.
 
 Using gammapy.astro.source
 ==========================
--- a/docs/user-guide/utils.rst
+++ b/docs/user-guide/utils.rst
@@ -100,9 +100,9 @@ For now, we use the `gammapy.utils.time.time_ref_from_dict`,
 to convert MET floats to `~astropy.time.Time` objects via the reference times
 stored in FITS headers.
 
-Time differences
-++++++++++++++++
+.. Time differences
+.. ++++++++++++++++
 
-TODO: discuss when to use `~astropy.time.TimeDelta` or `~astropy.units.Quantity`
-or :term:`MET` floats and where one needs to convert between those and what to watch
-out for.
+.. TODO: discuss when to use `~astropy.time.TimeDelta` or `~astropy.units.Quantity`
+.. or :term:`MET` floats and where one needs to convert between those and what to watch
+.. out for.
--- a/examples/tutorials/astrophysics/astro_dark_matter.py
+++ b/examples/tutorials/astrophysics/astro_dark_matter.py
@@ -43,13 +43,6 @@ from gammapy.astro.darkmatter import (
 )
 from gammapy.maps import WcsGeom, WcsNDMap
 
-######################################################################
-# Check setup
-# -----------
-from gammapy.utils.check import check_tutorials_setup
-
-check_tutorials_setup()
-
 
 ######################################################################
 # Profiles
--- a/examples/tutorials/data/cta.py
+++ b/examples/tutorials/data/cta.py
@@ -306,47 +306,76 @@ plt.show()
 # Source models
 # -------------
 #
-# The 1DC sky model is distributed as a set of XML files, which in turn
-# link to a ton of other FITS and text files. Gammapy doesn’t support this
-# XML model file format. We are currently developing a YAML based format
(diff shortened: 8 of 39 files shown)


gammapy/estimators/points/lightcurve.py
```diff
@@ -145,6 +157,20 @@ def run(self, datasets):
 
             valid_intervals.append([t_min, t_max])
 
+            if self.stack_over_time_interval:
+                name = f"timebin_{t_min.mjd:.0f}_{t_max.mjd:.0f}"
+                dataset_reduced = datasets_to_fit.stack_reduce(name=name)
+                datasets_to_fit = Datasets([dataset_reduced])
+                datasets_to_fit.models = datasets.models
```
registerrier: What about the background model if it exists here?

gammapy/estimators/points/lightcurve.py
```diff
@@ -145,6 +157,20 @@ def run(self, datasets):
...
+                if self.n_jobs == 1:
+                    fp = self.estimate_time_bin_flux(
+                        datasets_to_fit, datasets_to_fit.names
+                    )
+                    rows.append(fp)
+                else:
+                    parallel_datasets.append(datasets_to_fit)
+                continue
```
registerrier: Why is it needed here? Can't you proceed with the new `datasets_to_fit`?

gammapy/estimators/points/lightcurve.py
```diff
@@ -60,25 +60,30 @@ class LightCurveEstimator(FluxPointsEstimator):
...
+    reoptimize : bool, optional
         If True the free parameters of the other models are fitted in each bin independently,
         together with the norm of the source of interest
         (but the other parameters of the source of interest are kept frozen).
         If False only the norm of the source of interest if fitted,
         and all other parameters are frozen at their current values.
-    n_jobs : int
+        Default is False.
+    stack_over_time_interval : bool, optional
+        Whether to stacked over the time intervals or fit the norm on the
```
AtreyeeS:
```suggestion
        Whether to stacked datasets within each time interval
```

gammapy/estimators/points/lightcurve.py
```diff
@@ -60,25 +60,30 @@ class LightCurveEstimator(FluxPointsEstimator):
...
+        individual times. Default is None.
```
AtreyeeS: Default is False

gammapy/estimators/points/lightcurve.py
AtreyeeS: This can also get complex if `datasets_names` were specified on the models. Maybe using `Models.reassign` can be better

gammapy/estimators/points/lightcurve.py
Astro-Kirsty: I did it like this because the names where changed so I encountered an error

Astro-Kirsty:
I've adapted the PR to utilise `Models.reassign`, it is ready for re-review

gammapy/estimators/points/lightcurve.py
QRemy: If the datasets had background models they  are already applied by stack reduce so you cannot reassign them to the reduced dataset you would need to drop the existing  background models and create a new one. Regarding the models that applied to a specific dataset without being a background model I think that's fine.

gammapy/estimators/points/lightcurve.py
```diff
@@ -151,6 +162,24 @@ def run(self, datasets):
...
+                datasets_to_fit = Datasets([dataset_reduced])
+
+                if self.n_jobs == 1:
+                    fp = self.estimate_time_bin_flux(
+                        datasets_to_fit, datasets_to_fit.names
+                    )
+                    rows.append(fp)
+                else:
+                    parallel_datasets.append(datasets_to_fit)
+                continue
```
QRemy: Could you avoid redefining the if/else statement and just redefine `dataset_names` ?

gammapy/estimators/points/lightcurve.py
```diff
@@ -151,6 +162,24 @@ def run(self, datasets):
 
             valid_intervals.append([t_min, t_max])
 
+            if self.stack_over_time_interval:
+                name = f"timebin_{t_min.mjd:.0f}_{t_max.mjd:.0f}"
+                dataset_reduced = datasets_to_fit.stack_reduce(name=name)
+                models = datasets.models.copy()
+                for dataset in datasets:
+                    models.reassign(dataset.name, dataset_reduced.name)
```
QRemy: If the datasets had background models they are already applied by stack reduce so you cannot reassign them to the reduced dataset you would need to drop the existing background models and create a new one. Regarding the models that applied to a specific dataset without being a background model I think that's fine.

QRemy (changes requested):
Thanks @Astro-Kirsty,  just two comments:

gammapy/estimators/points/lightcurve.py
```diff
@@ -151,6 +162,16 @@ def run(self, datasets):
 
             valid_intervals.append([t_min, t_max])
 
+            if self.stack_over_time_interval:
+                name = f"timebin_{t_min.mjd:.0f}_{t_max.mjd:.0f}"
```
AtreyeeS: Will this name be ok for very short time intervals? Won't we get into duplicated names? Maybe just timebin_0, etc might be safer?

gammapy/estimators/points/lightcurve.py
```diff
@@ -161,13 +162,17 @@ def run(self, datasets):
...
+                name = f"timebin_{idx}"
                 dataset_reduced = datasets_to_fit.stack_reduce(name=name)
-                models = datasets.models.copy()
+                models = Models(datasets.models.copy())
+                # Remove background models already applied in stack_reduce
+                for model_name in models.background_models.values():
+                    models.remove(model_name)
                 for dataset in datasets:
                     models.reassign(dataset.name, dataset_reduced.name)
+                bkg_model = FoVBackgroundModel(dataset_name=dataset_reduced.name)
```
AtreyeeS: This is risky. If the user started off with `SpectrumDatasetOnOff` with releected regions, then one should not add a FoVBackgroundModel.
I dont know the safest behaviour here. Should we copy over the spectral/spatial correction models as well? Which parameters should be kept free? opinion @QRemy ?

gammapy/estimators/points/lightcurve.py
```diff
@@ -150,6 +162,20 @@ def run(self, datasets):
                 continue
 
             valid_intervals.append([t_min, t_max])
+            if self.stack_over_time_interval:
+                name = f"timebin_{idx}"
+                dataset_reduced = datasets_to_fit.stack_reduce(name=name)
```
QRemy: This will also stack obervations or event type with different irfs, Is it the intended behavior ?
It should be written explicitly in the docstring.

gammapy/estimators/points/lightcurve.py
QRemy: Maybe don't had a background model and make stack_over_time_interval possible  only if reoptimize=False

gammapy/estimators/points/lightcurve.py
AtreyeeS: If there are different event types, the user should not put `stack=True`.

Astro-Kirsty: Yes I agree, I can specify that in the docstring.

gammapy/estimators/points/lightcurve.py
```diff
@@ -63,25 +64,29 @@ class LightCurveEstimator(FluxPointsEstimator):
...
+    reoptimize : bool, optional
         If True the free parameters of the other models are fitted in each bin independently,
         together with the norm of the source of interest
         (but the other parameters of the source of interest are kept frozen).
         If False only the norm of the source of interest if fitted,
         and all other parameters are frozen at their current values.
-    n_jobs : int
+        Default is False.
+    stack_over_time_interval : bool, optional
+        Whether to stack datasets within each time interval. Default is False.
```
AtreyeeS, line 78:
```suggestion
        Whether to stack datasets within each time interval. Default is False.
        The predicted background counts from all datasets in the given interval 
        will be stacked together, and the final background model (if any) will not 
        have any free parameters. Available only if `reoptimize ` is False.
```

Astro-Kirsty:
The test is failing and I am a bit perplexed why. It was working previously before the name adaptation and removal of bkg model!

AtreyeeS (dismissed):

gammapy/estimators/points/lightcurve.py
```diff
@@ -63,25 +64,32 @@ class LightCurveEstimator(FluxPointsEstimator):
...
+        The predicted background counts from all datasets in the given interval 
+        will be stacked together, and the final background model (if any) will not 
+        have any free parameters. Available only if `reoptimize ` is False.
```
Astro-Kirsty:
```suggestion
        have any free parameters. Available only if ``reoptimize`` is False.
```

gammapy/estimators/points/lightcurve.py
```diff
@@ -150,6 +165,18 @@ def run(self, datasets):
                 continue
 
             valid_intervals.append([t_min, t_max])
+            if self.stack_over_time_interval and not self.reoptimize:
```
AtreyeeS, line 173: We can raise an error if `reoptimize=True` and `stack_overtime_interval` is True

Astro-Kirsty: Okay will do! 

We still have an error in the tests though

AtreyeeS: I will look into this

Astro-Kirsty:
Thanks for the fix @QRemy! Looks good now

PR #5900, opened 2025-06-04
Support for workflow step products 

QRemy:
Introduce the concepts of `Products`  and `Products` and adapt the `WorkflowStepBase` to be able to use them.
This do not change the way existing steps are implemented they still rely on storing their products directly as attribute of the workflow class (which could go away in future to keep everything contained in products).

### DIFF
--- a/gammapy/workflow/steps.py
+++ b/gammapy/workflow/steps.py
@@ -128,7 +128,7 @@ class WorkflowStepBase(abc.ABC):
         if parallel is None:
             parallel = self.parallel
         if parallel and ray is not None:
-            run_parrallel = ray.remote(self.__class__._run)
+            run_parrallel = ray.remote(self.__class__._get_and_run)
             ref = run_parrallel.remote(self)
             for p in self.outputs:
                 p.data = ref
@@ -140,9 +140,12 @@ class WorkflowStepBase(abc.ABC):
                 setattr(self.inputs, product_name, self.outputs[product_name].data)
         return self.outputs
 
+    def _get_and_run(self):
+        self.inputs.get()  # wait other remote and set results on self.inputs
+        return self._run()
+
     @abc.abstractmethod
     def _run(self):
-        self.inputs.get()  # wait other remote and set results on self.inputs
         pass
 
 
--- a/gammapy/workflow/tests/test_products.py
+++ b/gammapy/workflow/tests/test_products.py
@@ -21,8 +21,6 @@ class SumStep(WorkflowStepBase):
         return np.sum(self.inputs.data)
 
     def _run(self):
-        self.inputs.get()  # wait other remote and set results on self.data
-
         time.sleep(1)
         self.outputs[0].data = self._sum()
         self.outputs[1].data = "done"


QRemy:
Noticed that the tests are failing for astropy version 7.1 specifically:
Screenshot 2026-01-26 at 06 44 57

bkhelifi:
astropy 7.1 is not within our CI. I have not seen this error... But you can make an issue about the upgrade of pydantic from v2 to v3 and astropy,,,
However, the code coverage test is failing. Can you improve that?

gammapy/workflow/products.py
```diff
@@ -0,0 +1,246 @@
...
+    """Represents a product generated during a processing step in a workflow."""
+
+    def __init__(self, name=None, step_name=None, data=None):
+        """
+        Initialize a Product instance.
+
+        Parameters
+        ----------
+        name : str, optional
+            A label or identifier for the product.
```
bkhelifi, line 24:
```suggestion
            A label or identifier for the product. Default is None.
```

QRemy:
Implemted SonarCube suggestions, expect [this one](https://sonar-ctao.zeuthen.desy.de/project/issues?pullRequest=5900&issueStatuses=OPEN%2CCONFIRMED&open=50c00a1e-76ac-4d69-949d-01e8eade0678&id=gammapy_gammapy_8c1bf7fc-018a-4afd-b216-015191d282fa).

None of our index method support these arguments, so if we choose to implement them it should be done consistently for all our sequence derived container in a separate PR.

gammapy/workflow/products.py
```diff
@@ -0,0 +1,248 @@
...
```
bkhelifi, line 24: Why not precising the default value, as usual? @Astro-Kirsty, what do you recommend?

gammapy/workflow/products.py
```diff
@@ -0,0 +1,248 @@
...
+    """
+    Container for managing a list of `Product` instances generated by the workflow.
+
+    This class provides list-like behavior and additional utilities for
+    selecting, copying, and updating products, including support for
+    Ray-based asynchronous data resolution.
+
+    Parameters
+    ----------
+    products : list of Product, optional
```
bkhelifi:
```suggestion
    products : list of `Product`, optional
```

gammapy/workflow/products.py
```diff
@@ -0,0 +1,248 @@
...
+        names = []
+        for p in self._products:
+            components = [p.step_name, p.name, p.pid]
+            name = ".".join(components)
+            names.append(name)
+        return names
+
+    @property
+    def data(self):
+        """List of product data."""
```
bkhelifi:
```suggestion
        """List of `Product` data."""
```

gammapy/workflow/products.py
```diff
@@ -0,0 +1,248 @@
...
+            selection.extend(self.select(**req))
+        return selection
+
+    def select(self, name=None, step_name=None, pid=None):
+        """
+        Select products matching the given criteria.
+
+        Parameters
+        ----------
+        tag : str, optional
```
bkhelifi:
```suggestion
        name : str, optional
```

gammapy/workflow/products.py
```diff
@@ -0,0 +1,248 @@
...
+            Name to match.
```
bkhelifi:
```suggestion
            Name to match. Default is None.
```

gammapy/workflow/products.py
bkhelifi: What is the behaviour of `select` when no name is passed. This is not intuitive at all;) To be precised

gammapy/workflow/products.py
```diff
@@ -0,0 +1,248 @@
...
+        step_name : str, optional
+            Step name to match.
```
bkhelifi, line 192: When there is no `step_name`, what is the behaviour? This is not clear...

gammapy/workflow/products.py
```diff
@@ -0,0 +1,248 @@
...
+        Products
+            A new `Products` instance with matching products.
+        """
+
+        mask = self.selection_mask(name=name, step_name=step_name, pid=pid)
+        return self[mask]
+
+    def selection_mask(self, name=None, step_name=None, pid=None):
+        """
+        Create a boolean mask for selecting products.
```
bkhelifi:
```suggestion
        Create a boolean mask for selected products.
```

gammapy/workflow/products.py
```diff
@@ -0,0 +1,248 @@
...
+
+        Parameters
+        ----------
+        name : str, optional
```
bkhelifi, line 213: One must define the behaviour when no selection criteria are applied

gammapy/workflow/steps.py
```diff
@@ -26,37 +26,123 @@
...
+    parallel = False
+    _products_as_workfllow_attributes = False
 
     def __init__(self, config, log=None, name=None, overwrite=True):
         self.config = config
         self.overwrite = overwrite
-        self._name = make_name(name)
-
+        self.name = make_name(name)
+        self._data = Products()
```
bkhelifi: It is not `Products().data`???

gammapy/workflow/tests/test_products.py
```diff
@@ -0,0 +1,99 @@
...
+from gammapy.utils.testing import requires_dependency
+from gammapy.workflow.steps import WorkflowStepBase
+from gammapy.workflow.products import Products, Product
+
+
+class SumStep(WorkflowStepBase):
+    tag = "sum"
+    parallel = True
+    required_data = [{"name": "value"}]
+    products_names = ["value", "log"]  # could be update by the config in init
```
bkhelifi:
```suggestion
    products_names = ["value", "log"]  # could be updated by the config in init
```

bkhelifi:
If I understand correctly, it is not aimed to be exposed now. What is your (test) plan up to the insertion into the documentation and with a tutorial (is possible)?

gammapy/workflow/steps.py
QRemy: Well no, that's might a bit confusing but `.data` is a `Products` listing the input while `.products` is  a `Products` listing the output. That's because data can be a list of `Product` from previous steps that are not yet computed (and in that case Product.data is only a reference and the update mecanism is on the products class).

QRemy: maybe could rename `self.data` as `self.input` and `self.products` as `self.output` ?

QRemy:
> If I understand correctly, it is not aimed to be exposed now. What is your (test) plan up to the insertion into the documentation and with a tutorial (is possible)?

The whole workflow subpackage is yet not exposed, I think it should be complete  and stable before we show it.
At the end it allows to make custom workflow that can run steps in parallel with ray automatically managing the execution order. This could be shown in a dedicated tutorial.

For now this PR simply change the internal machinery, and introduce the products concept.
However current steps implemented do not fully exploit the products system as for now they store directly their output in self.data, I will have another PR  to change that.

gammapy/workflow/products.py
```diff
@@ -0,0 +1,257 @@
...
+            step_name = ""
+        self.step_name = step_name
+        self.data = data
+        if name is None:
+            name = make_name()
+        self.name = name
+        self.pid = make_name()  # unique id generated for each product
+
+    @property
+    @requires_module("ray")
```
registerrier, line 43: Why should `Product` depend on ray? This should limited at the `Workflow` level. Distributing the dependency down to basic elements does seem a good design.

gammapy/workflow/products.py
```diff
@@ -0,0 +1,257 @@
...
+        for idx, p in enumerate(self._products):
+            if name is not None:
+                mask[idx] &= p.name == name
+            if step_name is not None:
+                mask[idx] &= p.step_name == step_name
+            if pid is not None:
+                mask[idx] &= p.pid == pid
+        return mask
+
+    @requires_module("ray")
```
registerrier, line 239: Same question as above.

gammapy/workflow/steps.py
```diff
@@ -26,37 +26,123 @@
 from gammapy.maps import Map, MapAxis, RegionGeom, WcsGeom
 from gammapy.utils.pbar import progress_bar
 from gammapy.utils.scripts import make_name, make_path
+from gammapy.workflow.products import Products, Product
+
+try:
```
registerrier, line 31: Why importing ray here.

gammapy/workflow/steps.py
registerrier: Yes clarify input and output seems best

gammapy/workflow/products.py
```diff
@@ -0,0 +1,257 @@
...
+        """List of product names."""
+        return [p.name for p in self._products]
+
+    @property
+    def pids(self):
+        """List of product names."""
+        return [p.pid for p in self._products]
+
+    @property
+    def unique_names(self):
```
registerrier, line 100: Why should there be products with identical names? Shouldn't it be prevented instead?

gammapy/workflow/products.py
```diff
@@ -0,0 +1,257 @@
...
+        Returns
+        -------
+        Product
+            A new `Product` instance with the same `tag`, `step_name`, and `data`,
+            but a newly generated `name`.
+        """
+        return self.__class__(name=self.name, step_name=self.step_name, data=self.data)
+
+
+class Products(collections.abc.MutableSequence):
```
registerrier, line 70: In general, wouldn't a dict based collection be more useful. Can we really rely on products in a given order? Why not naming them?

gammapy/workflow/products.py
```diff
@@ -0,0 +1,257 @@
...
+        Parameters
+        ----------
+        name : str, optional
+            A label or identifier for the product.
+            Default is None, and the name is randomly generated.
+        step_name : str, optional
+            The name of the processing step that produced this product.
+            Default is None.
+        data : object, optional
+            The data associated with the product. Can be any object, including a Ray ObjectRef.
```
registerrier: The ObjectRef here is a bit confusing

gammapy/workflow/steps.py
```diff
@@ -26,37 +26,123 @@
...
+        Name identifier for the step. If not provided, a unique name is generated.
+    overwrite : bool, optional
+        Whether to overwrite existing outputs. Default is True.
+    """
+
     tag = "workflow-step"
+    required_data = []
+    products_names = []
+    parallel = False
+    _products_as_workfllow_attributes = False
```
registerrier, line 56: what is this for?

registerrier (commented):
Thanks @QRemy. I have some general comments.

- The tight coupling of products and steps with ray is worrisome. Containers should not have to deal with this
- Could there be some type specifications for Product? Shouldn't there be some input validation?
- Why not using a dict like structure for `Products`? They should all have different names no?

gammapy/workflow/steps.py
QRemy: it is used in the `.run` method to determine if the step can be evaluated in parallel. In that case the product data is a reference to the `._run` method.

gammapy/workflow/products.py
QRemy: Product can work without ray but if ray is available their data can be set as a reference to the '_run' method that compute them and be fetch only when necessary.

gammapy/workflow/products.py
QRemy: When ray is available product data can be only references and this method check if they need to be computed.
I the example below the step sum1 and sum2 are independent, if they were not yet executed their `.products` are only reference,  they will run in parallel once `get` is called.
```
    analysis_products = Products([*sum1.products, *sum2.products])
    analysis_products.get()
```

gammapy/workflow/steps.py
QRemy: This is for compatibility with the current behaviour of the hli.
The output of the existing step store are also stored as attributes on the analysis class:
analysis.observations
analysis.datasets
...

gammapy/workflow/products.py
QRemy: Removed so it appears only in the methods that uses this mecanism, and improved the docstring there.

gammapy/workflow/products.py
QRemy: right now name is used more like a type as the current step products names are  "observations", "datasets", "flux_points"...

gammapy/workflow/steps.py
QRemy: renamed step.data as step.inputs and step.products as step.outputs

QRemy:
Added a test without ray.  It works exactly in the same way so parallelisation will make no difference for the user.

registerrier:
As decided during dev call, let's merge this for now. This is still non public API and will require more development.

PR #5901, opened 2025-06-04
Documentation improvement for the case of NaN UL

bkhelifi:
**Description**
We have frequently questions about the case of failure of UL computation by the `LightCurveEstimator` or the `FluxPointsEstimator`. 

This PR adds a reference to the User Guide HowTo into the docstring of these estimators, hoping that it helps...

### DIFF
--- a/docs/user-guide/howto.rst
+++ b/docs/user-guide/howto.rst
@@ -275,10 +275,11 @@ Please give feedback and suggest additions to this page!
 .. dropdown:: Avoid NaN results in Flux Point estimation
 
     Sometimes, upper limit values may show as ``nan`` while running a `~gammapy.estimators.FluxPointsEstimator`
-    or a `~gammapy.estimators.LightCurveEstimator`. This often arises because the range of the norm parameter
-    being scanned over is not sufficient. Increasing this range usually solves the problem (e.g.
-    `fpe.norm.scan_min=1.e-5`, `fpe.norm.scan_max=1.e4`, `fpe.norm.scan_n_values=30`). In some cases,
-    you can also consider configuring the estimator with a different `~gammapy.modeling.Fit` backend.
+    or a `~gammapy.estimators.LightCurveEstimator`. This likely arises because the min/max range of the `norm` parameter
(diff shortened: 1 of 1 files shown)


docs/user-guide/howto.rst
```diff
@@ -276,7 +276,8 @@ Please give feedback and suggest additions to this page!
 
     Sometimes, upper limit values may show as ``nan`` while running a `~gammapy.estimators.FluxPointsEstimator`
     or a `~gammapy.estimators.LightCurveEstimator`. This often arises because the range of the norm parameter
-    being scanned over is not sufficient. Increasing this range usually solves the problem. In some cases,
+    being scanned over is not sufficient. Increasing this range usually solves the problem (e.g.
+    `fpe.norm.scan_min=1.e-5`, `fpe.norm.scan_max=1.e4`, `fpe.norm.scan_n_values=30`). In some cases,
```
QRemy: fpe is not define so it's not very clear and it's min/max not scan_min/scan_max for the flux points ul (scan_min/scan_max is used only for the scan now)

docs/user-guide/howto.rst
```diff
@@ -276,7 +276,8 @@ Please give feedback and suggest additions to this page!
 
     Sometimes, upper limit values may show as ``nan`` while running a `~gammapy.estimators.FluxPointsEstimator`
     or a `~gammapy.estimators.LightCurveEstimator`. This often arises because the range of the norm parameter
```
QRemy:
```suggestion
    or a `~gammapy.estimators.LightCurveEstimator`. This likely arises because the min/max range of the norm parameter attribute attached to the estimator
```

docs/user-guide/howto.rst
```diff
@@ -276,7 +276,8 @@ Please give feedback and suggest additions to this page!
...
-    being scanned over is not sufficient. Increasing this range usually solves the problem. In some cases,
+    being scanned over is not sufficient. Increasing this range usually solves the problem (e.g.
```
QRemy:
```suggestion
  is not well-chosen (the automatic default may not be valid for all use cases). Setting these min/max to large values usually solves the problem (e.g.
```

PR #5903, opened 2025-06-05
Add `LogUniformPrior` to registry

Astro-Kirsty:
Fix issue with serialisation, without this is it not possible to read your model back in


bkhelifi (dismissed):
Well seen!

QRemy (changes requested):
Thanks @Astro-Kirsty, could you add a test ?

PR #5904, opened 2025-06-05
Add requires_module decorator

QRemy:
A decorator that conditionally enables a method or property based on the availability of a module.
Used to activate some methods/properties only when ray is available.
Splitted from #5900

### DIFF
--- a/gammapy/utils/scripts.py
+++ b/gammapy/utils/scripts.py
@@ -254,7 +254,9 @@ def requires_module(module_name):
 
                 return wrapper
             else:
-                return obj  # Fallback: return as-is
+                raise TypeError(
+                    "requires_module can only be used on methods or properties."
+                )
 
     return decorator
 
--- a/gammapy/utils/tests/test_scripts.py
+++ b/gammapy/utils/tests/test_scripts.py
@@ -99,3 +99,11 @@ def test_requires_module():
         match="The 'nonexistent_module' module is required to use this property.",
     ):
         _ = result.prop_unavailable
+
+    with pytest.raises(
+        TypeError, match="requires_module can only be used on methods or properties."
+    ):
+
+        @requires_module("nonexistent_module")
+        class InvalidUsage:
+            pass


gammapy/utils/scripts.py
```diff
@@ -213,3 +215,56 @@ def recursive_merge_dicts(a, b):
...
+def requires_module(module_name):
+    """
+    Decorator that conditionally enables a method or property based on the availability of a module.
+
+    If the specified module is available, the decorated method or property is returned as-is.
+    If the module is not available:
+      - For methods: replaces the method with one that raises ImportError when called.
+      - For properties: replaces the property with one that raises ImportError when accessed.
+
+    Args:
```
bkhelifi:
```suggestion
    Parameters
    ----------
```

gammapy/utils/scripts.py
```diff
@@ -213,3 +215,56 @@ def recursive_merge_dicts(a, b):
...
+        module_name (str): The name of the module to check for.
+
+    Returns:
```
bkhelifi:
```suggestion
    Returns
    -------
```

gammapy/utils/scripts.py
```diff
@@ -213,3 +215,56 @@ def recursive_merge_dicts(a, b):
...
+    Decorator that conditionally enables a method or property based on the availability of a module.
+
+    If the specified module is available, the decorated method or property is returned as-is.
+    If the module is not available:
+      - For methods: replaces the method with one that raises ImportError when called.
+      - For properties: replaces the property with one that raises ImportError when accessed.
+
+    Args:
+        module_name (str): The name of the module to check for.
+
```
bkhelifi:
```suggestion
        module_name : str
            The name of the module to check for.

```

gammapy/utils/scripts.py
```diff
@@ -213,3 +215,56 @@ def recursive_merge_dicts(a, b):
...
+    Returns:
+        function or property: The original object if the module is available, otherwise a fallback.
```
bkhelifi:
```suggestion
        function or property
            The original object if the module is available, otherwise a fallback.
```

gammapy/utils/scripts.py
```diff
@@ -213,3 +215,56 @@ def recursive_merge_dicts(a, b):
...
+                return obj  # Fallback: return as-is
+
+    return decorator
+
+
+def raise_import_error(module_name, is_property=False):
+    """
+    Raises an ImportError with a descriptive message about a missing module.
+
+    Args:
```
bkhelifi: to be updated

gammapy/utils/scripts.py
```diff
@@ -213,3 +215,56 @@ def recursive_merge_dicts(a, b):
...
+        module_name (str): The name of the required module.
+        is_property (bool): Whether the error is for a property (affects the error message).
+
+    Raises:
```
bkhelifi: to be updated (this is not a sphinx keyword)

gammapy/utils/scripts.py
```diff
@@ -213,3 +215,56 @@ def recursive_merge_dicts(a, b):
...
+    return decorator
+
+
+def raise_import_error(module_name, is_property=False):
+    """
+    Raises an ImportError with a descriptive message about a missing module.
+
+    Args:
+        module_name (str): The name of the required module.
+        is_property (bool): Whether the error is for a property (affects the error message).
```
bkhelifi: to be adapted

gammapy/utils/scripts.py
QRemy:
```suggestion
    Parameters
    ----------
```

gammapy/utils/scripts.py
```diff
@@ -213,3 +215,56 @@ def recursive_merge_dicts(a, b):
...
+
+    return decorator
+
+
+def raise_import_error(module_name, is_property=False):
+    """
+    Raises an ImportError with a descriptive message about a missing module.
+
+    Args:
+        module_name (str): The name of the required module.
```
QRemy:
```suggestion
        module_name : str
            The name of the required module.
```

gammapy/utils/scripts.py
QRemy:
```suggestion
        is_property : bool
            Whether the error is for a property (affects the error message).
```

gammapy/utils/scripts.py
```diff
@@ -213,3 +215,56 @@ def recursive_merge_dicts(a, b):
...
+        is_property (bool): Whether the error is for a property (affects the error message).
+
+    Raises:
+        ImportError: Always raised with a formatted message.
```
QRemy:
```suggestion
```

gammapy/utils/scripts.py
```diff
@@ -213,3 +215,60 @@ def recursive_merge_dicts(a, b):
...
+                )
+            elif isinstance(obj, (types.FunctionType, types.MethodType)):
+
+                @functools.wraps(obj)
+                def wrapper(*args, **kwargs):
+                    raise_import_error(module_name)
+
+                return wrapper
+            else:
+                return obj  # Fallback: return as-is
```
registerrier: Maybe test this. 
What case could it cover? If the decorator is put on a class?

gammapy/utils/tests/test_scripts.py
```diff
@@ -54,3 +56,46 @@ def test_read_write_yaml_checksum(tmp_path):
...
+            import math
+
+            return math.sqrt(9)
+
+        @requires_module("nonexistent_module")
+        def method_unavailable(self):
+            return "Should not be called"
+
+        @property
+        @requires_module("math")
```
registerrier: Does the order matter?

registerrier (dismissed):
Thanks @QRemy . Looks good. I have left a few minor comments inline.

gammapy/utils/tests/test_scripts.py
QRemy:
```suggestion
        @requires_module("math")
        @property
```

QRemy: `@requires_module("math")` has to be first so the error message says property instead of method, but for the valid case it is the same.

gammapy/utils/scripts.py
QRemy: now it raises an error and added test

PR #5910, opened 2025-06-12
Change argument name in ul

Astro-Kirsty:
We should change the argument name here because it is incorrect. 
It is actually the `stat_scan` minus your `total_stat` i.e. the deltaTS which should have a zero value in it, allowing the minimum to be found.
As you can see the description of the parameter is correct but using `stat_scan` is misleading.


bkhelifi:
Thanks @Astro-Kirsty . Can you also add a Notes in the docstring to insist on the fact that it should be the DeltaTS and not the TS, for the reasons that you give in this PR?
Thanks

registerrier (commented):
Thanks @Astro-Kirsty .

isn't it simpler to modify `f(x)` so that it is `interp - stat_scan[idx] - delta_ts`?

registerrier:
From the tests failure, it seems that it really needs the best fit statistic value, because the scan values will be slightly off. This raises two questions:
- is all usage in the code base correct w.r.t. passing the stat_scan - total_stat ?
- are there cases where we want to use this because the fit failed?

One could also use the minimum of the interpolated function instead of the minimum scan value.

Astro-Kirsty:
Was there any discussion of this in dev calls? How do we want to proceed?

Astro-Kirsty:
In the end `stat_profile_ul_scipy` is only used in the test and `squash_fluxpoints`. The latter is currently not used anywhere in the code and not exposed in the documentation. It is also utilising the `stat_scan` as the correct parameter. So our implementation here will be correct for that now.
I adjusted the tests accordingly and added another test with real data.

gammapy/modeling/tests/test_scipy.py
```diff
@@ -106,5 +107,14 @@ def test_stat_profile_ul_scipy():
     ul = stat_profile_ul_scipy(x, y)
     assert_allclose(ul, 2)
 
-    ul = stat_profile_ul_scipy(x, x, interp_scale="lin")
+    ul = stat_profile_ul_scipy(y, y, interp_scale="lin")
```
bkhelifi: this is not (x, y)?

gammapy/modeling/scipy.py
```diff
@@ -164,12 +162,26 @@ def stat_profile_ul_scipy(
...
+        )
+
+    result = scipy.optimize.minimize_scalar(
+        interp, bounds=(value_scan[0], value_scan[-1]), method="bounded"
+    )
+    if not result.success:
+        raise RuntimeError("Failed to find minimum in interpolated profile.")
+
+    norm_best_fit = result.x
+    stat_best_fit = result.fun
```
QRemy, line 186: Not sure why it fails to find a minimum in the test.
Maybe this would be more stable but it uses a different interpolation method so might not be 100% consistent:
```python
    f = InterpolatedUnivariateSpline(value_scan, stat_scan, k=4)
    roots = f.derivative().roots()
    roots = np.append(roots, (value_scan[0], stat_scan[-1]))
    stat _roots= f(roots)
    idx = np.argmin(_roots)
    stat_best_fit = stat[idx]
    norm_best_fit = roots[idx]
```

Astro-Kirsty:
Thanks for the adaptation @QRemy, all tests are passing now.

PR #5911, opened 2025-06-13
Introducing generic reproject_irf_on_geom functions using FoVFrame classes

registerrier:
**Description**

This pull request continues the re-factoring of makers and observation to handle FoV coordinate transform.

It introduces two functions in makers.utils to perform all type of IRF evaluation or integration. 
- `project_irf_on_geom(geom, irf, fov_frame, use_region_center=True)`
- `integrate_project_irf_on_geom(geom, irf, fov_frame, use_region_center=True)``

The first one evaluates the IRF at geom coordinates, the second integrates over geom coordinates. The latter is used for background, the former for all other IRFs. 

All the code could have been squeezed in a single function but it gets a bit complex. I chose for a clearer separation of tasks at the cost of some code duplication. 

Time integration is only support for background for now. Test values have changed a bit because of the trapezoidal integration used compared to rectangular one used before.

All make functions have been refactored to used these functions. Now they simply create the FoV frame, call the project function and convert the resulting map to the correct unit or normalize the result.

A next step will be not to rely anymore on the `make_xxx` functions in the `MapMaker`. Then we can see whether we deprecate and remove them or not. 

**Dear reviewer**

### DIFF
--- a/gammapy/data/tests/test_observations.py
+++ b/gammapy/data/tests/test_observations.py
@@ -20,6 +20,7 @@ from gammapy.data.pointing import FixedPointingInfo
 from gammapy.data.utils import get_irfs_features
 from gammapy.irf import PSF3D, load_irf_dict_from_file
 from gammapy.utils.cluster import hierarchical_clustering
+from gammapy.utils.coordinates import FoVICRSFrame, FoVAltAzFrame
 from gammapy.utils.fits import HDULocation
 from gammapy.utils.testing import (
     assert_skycoord_allclose,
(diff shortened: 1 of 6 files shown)


gammapy/irf/psf/parametric.py
```diff
@@ -328,6 +328,7 @@ class EnergyDependentMultiGaussPSF(ParametricPSF):
     tag = "psf_3gauss"
     required_axes = ["energy_true", "offset"]
     required_parameters = ["sigma_1", "sigma_2", "sigma_3", "scale", "ampl_2", "ampl_3"]
+    required_arguments = ["rad", "energy_true", "offset"]
```
bkhelifi, line 338: One should remember that this is gadf specific. Should we keep this simple scheme or start to anticipate a more generic approach?

AtreyeeS (approved):
Thanks @registerrier ! Looks good
I do not completely understand though why you introduced the `required_arguments`.

gammapy/data/tests/test_observations.py
```diff
@@ -52,6 +53,26 @@ def test_observation(data_store):
...
 
+@requires_data()
+def test_observation_get_fov_frame(data_store):
+    """Test Observation class"""
+    obs = data_store.obs(23523)
+
+    fov_icrs_frame = obs.get_fov_frame(obs.tmid, "RADEC")
+    fov_altaz_frame = obs.get_fov_frame(obs.tmid, "ALTAZ")
+
+    pnt_icrs = SkyCoord(83.63333129882812, 21.51444435119629, unit="deg", frame="icrs")
```
bkhelifi, line 64: They are many many digits... ;)

Astro-Kirsty:
Should `project_irf_on_geom` and `integrate_project_irf_on_geom` be exposed in the docs?

Issue #5908, opened 2025-06-11
Sampler fails when you have linked parameters


Astro-Kirsty:
**Gammapy version**
dev version

**Bug description**
When using the sampler on a model that has linked parameters, it fails. 
![Image](https://github.com/user-attachments/assets/d1d74fc7-13ee-480d-8e54-6ff6b2b03689)

The issue is that when you link the parameters you still have all the parameters (i.e. twice the linked params): 
- link two parameters together
`spec1.amplitude = spec2.amplitude`
`spec1.index = spec2.index`
- they are free and have priors added to them 
- `parameters.free_parameters.names` will return `amplitude, index, amplitude, index`

**To Reproduce**
Simply link two parameters and try to run the sampler:

```
from gammapy.modeling.models import PowerLawSpectralModel, SkyModel, UniformPrior, Models
pwl1 = PowerLawSpectralModel(index=2.3)
pwl2 = PowerLawSpectralModel()
pwl2.index = pwl1.index
models = Models([SkyModel(pwl1, name="source1"), SkyModel(pwl2, name="source2")])
models[0].spectral_model.index.prior = UniformPrior(min=1, max=5)
datasets.models = models
result_joint = sampler.run(datasets)
```

facero:
Thanks for noticing that. I had imagined that this could cause issue when we started this.
I'm surprised it even runs as the same issue will happen when running the sampler : 
https://github.com/gammapy/gammapy/blob/79809b32b9867b6c82a37c88c810f911f37d0c4b/gammapy/modeling/sampler.py#L176

UltraNest is then probably sampling the two index because it was given two index to start with.

Is there a function that would replace `parameters.free_parameters ` by something like  `parameters.free_NOT_TIED_parameters` ?

QRemy:
It seems to samples only 3 parameters as expected but there is two issues :
- the models copy (used in sampler.run) break the link
this could be solved by reusing the `_set_link` code from `Models.from_dict` in `Models.copy`
```python
def _set_models_links(models):
    shared_register = {}
    for model in models:
        if isinstance(model, SkyModel):
            submodels = [
                model.spectral_model,
                model.spatial_model,
                model.temporal_model,
            ]
            for submodel in submodels:
                if submodel is not None:
                    shared_register = _set_link(shared_register, submodel)
        else:
            shared_register = _set_link(shared_register, model)

_set_models_links(models_copy)
```

[...]

PR #5913, opened 2025-06-13
Add a function to retrieve free and unique parameters 

facero:
Currently there is no easy way to retrieve the list of parameters that are not frozen AND that are not linked.

This problem arised in issue #5908  in the sampler module when one is linking some parameters to other.
What was happening was that the sampler was received all free parameters including the link parameter twice.

This PR introduces a `parameters.free_unique_parameters` to retrieve free AND not linked parameters by simply combining the existing `unique_parameters` and `free_parameters`.

Also adds a test for free_parameters which wasn't tested before.

This partially solves  issue #5908. The other issue is that copying a model forgets about the link between param. This would need another PR.

### DIFF
--- a/gammapy/modeling/tests/test_parameter.py
+++ b/gammapy/modeling/tests/test_parameter.py
@@ -194,6 +194,32 @@ def test_unique_parameters():
     assert parameters_unique.names == ["a", "b", "c"]
 
 
+def test_free_parameters():
+    a = Parameter("a", 1, frozen=False)
+    b = Parameter("b", 2, frozen=True)
+    c = Parameter("c", 2, frozen=False)
+
+    parameters = Parameters([a, b, c])
+    free = parameters.free_parameters
+    frozen = parameters.frozen_parameters
+    assert free.names == ["a", "c"]
+    assert frozen.names == ["b"]
+    assert all(par.frozen for par in frozen)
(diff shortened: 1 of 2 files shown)


gammapy/modeling/tests/test_parameter.py
```diff
@@ -194,6 +194,32 @@ def test_unique_parameters():
...
 
 
+def test_free_parameters():
+    a = Parameter("a", 1, frozen=False)
+    b = Parameter("b", 2, frozen=True)
+    c = Parameter("c", 2, frozen=False)
+
+    parameters = Parameters([a, b, c])
+    free = parameters.free_parameters
+    frozen = parameters.frozen_parameters
```
QRemy:
```suggestion
```

gammapy/modeling/tests/test_parameter.py
```diff
@@ -194,6 +194,32 @@ def test_unique_parameters():
...
+    assert free.names == ["a", "c"]
+    assert frozen.names == ["b"]
+    assert all(par.frozen for par in frozen)
```
QRemy:
```suggestion
```

Astro-Kirsty (approved):
Thanks for this @facero

QRemy (approved):
Thanks @facero, merging now.

PR #5914, opened 2025-06-13
Modify sampler to handle linked params in models

facero:
This PR is the 2nd step to solve issue https://github.com/gammapy/gammapy/issues/5908 
 which occurs when some parameters of the models are linked together.

The sampler and the `update_model` will be done only on free & unique parameters by replacing  `free_parameters` by `free_unique_parameters` introduced in PR #5913. 

This can be merged after PR #5913.


codemeta.json
```diff
@@ -359,10 +359,8 @@
         "active"
     ],
     "email": "GAMMAPY-COORDINATION-L@IN2P3.FR",
-
-    "dateModified": "2024-11-29",
-
-  "softwareRequirements": [
+    "dateModified": "2025-06-12",
+    "softwareRequirements": [
```
QRemy: this change should not appear

QRemy:
First step to solve the issue done in #5915 so it should go before.

codemeta.json
facero: I have no idea why this showed up. Is it my gitignore that is not ignoring these files ?

codemeta.json
```diff
@@ -373,4 +373,4 @@
         "iminuit>=2.8.0",
         "matplotlib>=3.4, <3.10"
     ]
-}
\ No newline at end of file
+}
```
QRemy, line 376:
```suggestion
}

```

facero:
All the other PRs that were needed have been merged and this one is the last one.
All checks have passed

codemeta.json
```diff
@@ -373,4 +373,5 @@
...
```
registerrier, line 376: Can you revert the commit of condmeta.json

gammapy/modeling/tests/test_sampler.py
```diff
@@ -19,11 +21,17 @@ def test_run(backend="ultranest"):
         )
         datasets.append(dataset)
 
-    datasets.models = [SkyModel.create(spectral_model="pl")]
-    datasets.models.parameters["index"].prior = UniformPrior(min=2, max=3)
-    datasets.models.parameters["amplitude"].prior = LogUniformPrior(
-        min=1e-12, max=1e-10
-    )
+    # test with linked parameters
```
registerrier, line 88: Ideally the linked parameter test should be differnt from the main test

registerrier (commented):
Thanks @facero . Looks good to me.

Please separate the test for linked parameters from the main functionality check, so that it is not removed in some re-factoring.

codemeta.json
facero: I hate this condmeta file ! 
I don't why it keeps changing.
Should I add to the gitignore ?

gammapy/modeling/tests/test_sampler.py
facero: should we do one simple test without linked param and one with linked param then ?

codemeta.json
registerrier: well here in fact you manually removed the changes so its's fine. Sorry

gammapy/modeling/tests/test_sampler.py
QRemy: yes (I only modified the existing test because I was concerned about the time taken by the sampler but actually it is still reasonable)

Astro-Kirsty: Yes, I think this is a good idea

gammapy/modeling/tests/test_sampler.py
```diff
@@ -19,12 +21,66 @@ def test_run(backend="ultranest"):
         )
         datasets.append(dataset)
 
-    datasets.models = [SkyModel.create(spectral_model="pl")]
-    datasets.models.parameters["index"].prior = UniformPrior(min=2, max=3)
-    datasets.models.parameters["amplitude"].prior = LogUniformPrior(
-        min=1e-12, max=1e-10
+    # test with linked parameters
```
Astro-Kirsty:
```suggestion
```

gammapy/modeling/tests/test_sampler.py
```diff
@@ -19,12 +21,66 @@ def test_run(backend="ultranest"):
...
+    assert_allclose(result.models.parameters["amplitude"].value, 4.0e-11, rtol=0.2)
+
+    assert result.models.parameters["index"].error > 0
+    assert result.models.parameters["amplitude"].error > 0
+    assert result.models._covariance is None
+
+
+@requires_dependency("ultranest")
+@requires_data()
+def test_run_linked(backend="ultranest"):
```
Astro-Kirsty:
```suggestion
def test_run_linked_params(backend="ultranest"):
```

gammapy/modeling/tests/test_sampler.py
facero: each tests takes only about 10sec so I added a 2nd one

facero:
All checks are now passing

Astro-Kirsty (dismissed):

registerrier (dismissed):
Thanks @facero . Looks good to me.

I don't remember why the relative tolerance is so large. Do the results vary as much as 20%?

facero:
> Thanks @facero . Looks good to me.
> 
> I don't remember why the relative tolerance is so large. Do the results vary as much as 20%?

Done updated it to 10%. The variation from one run to another are much smaller than that but if we want to compare different backends in the future 5 or 10% might be good.
I also added a test on error values.

PR #5915, opened 2025-06-13
Fix parameters links after models copy

QRemy:
This PR is the 1st step to solve issue #5908
Fix parameters links after models copy.


facero:
Thx !
No more comments from my side.

PR #5916, opened 2025-06-13
 Fix link label display in parameters.to_table

QRemy:
Fix link label display in parameters.to_table


facero:
Great !
This should fix it

Astro-Kirsty (dismissed):
Thanks so much @QRemy! Looks great

PR #5922, opened 2025-06-16
Clean up of `irf/background.py`

Astro-Kirsty:
**Description**

This pull request ...

**Dear reviewer**


gammapy/irf/background.py
```diff
@@ -25,13 +25,22 @@ class BackgroundIRF(IRF):
...
+        True for point-like IRFs, False for full-containment.
+        Default is False.
+    fov_alignment : `~gammapy.irf.FoVAlignment`, optional
+        The orientation of the field of view coordinate system.
+        Default is FoVAlignment.RADEC.
     meta : dict
         Metadata dictionary.
-
+    interp_kwargs : dict, optional
+        Interpolation kwargs to extrapolate.
```
Tobychev: This is confusing; following the code the interpolator ends up being a wrapper around `scipy.interpolate.RegularGridInterpolator` (via `IRF._interpolate` and `utils.ScaledRegularGridInterpolator`) which, _by default_, errors if you try to request values outside the interpolation grid (ie by default it doesn't actually extrapolate, only interpolate). 

However, if you don't pass any keyword arguments gammapy tells the `RegularGridInterpolator` to _allow_ extrapolation. So if you pass a `interp_kwargs` there's a good chance you will disable extrapolation by mistake...

Worse, the `interp_kwargs` is also used to pass an argument to entirely unrelated interpolation code in `IRF.interp_missing_data` that uses numpy with hardcoded settings... 

In summary, the suggested docstring here is probably not sufficient.

Astro-Kirsty: Okay yes that is fair enough. Maybe that is why the `interp_kwargs` was not previously defined. 

We should therefore copy the input for the `ScaledRegularGridInterpolator` which is  "Keyword arguments passed to `RegularGridInterpolator`." 
Does that sound okay? Or do you have a better suggestion?

Tobychev: That's fine, but so long as an a class inherits from `IRF` there should probably be a mention that this is where you specify the kwarg `values_scale` that is used by `interp_missing_data`.

So something like "Keyword arguments passed to RegularGridInterpolator. Also used to pass `values_scale` to `interp_missing_data`"

Astro-Kirsty: Although it is true that its passed to `interp_missing_data`, it still relates to interpolating, and I think it would feel odd to mention that function specifically here.

Tobychev: Yes, but I think it's important because `interp_missing_data` uses a completely different interpolation method and I'm not sure the `ScaledRegularGridInterpolator` will survive seeing `value_scales` among its arguments.

Ultimately I'd it is a problem that `interp_kwargs` is used for communicating with two different things, but that can't be solved by the doc string :'(

Astro-Kirsty: I don't think this is used much so I suggest we proceed with the more simple option for the docstring. 

[This PR](https://github.com/gammapy/gammapy/pull/3219) relates to adding in this function. And [here too](#3420).
Maybe @adonath , @QRemy  can shed some light on this?

Astro-Kirsty: From the dev call it was decided the docstring should mention passing to the `ScaledRegularGridInterpolator`. As the function `interp_missing_data` is also using the parameters from `ScaledRegularGridInterpolator`. 
We should add further information in the docstring of `interp_missing_data` but that is for another PR.

registerrier (approved):

PR #5925, opened 2025-06-16
Adapt docstring for `ReflectedRegionsFinder`

### LINKED ISSUE #4492: ReflectedRegionsBackgroundMaker not handling the distances correctly
**Gammapy version**
1.0.1

**Bug description**
Looks like the distances are not handled correctly in ReflectedRegionsBackgroundMaker.

For example,

```
ReflectedRegionsBackgroundMaker(
                min_distance = 15 * u.deg,
                max_region_number=n_off_regions,
                min_distance_input = 25*u.deg)
```

I would expect this to result in no OFF regions generated, since the distance between them is huge. However, it results in

image

If I put arcmins there, that ring of background circles should not be continuous since the distance is not that small, and it should be interrupted near the ON region. 

```
ReflectedRegionsBackgroundMaker(
                min_distance = 15 * u.arcmin,
                max_region_number=n_off_regions,
                min_distance_input = 25*u.arcmin)
```

but it results in

image

**Expected behavior**
I would expect the first plot (or something alike) when running the code with the distances of the second example. 

**To Reproduce**

Running ReflectedRegionsBackgroundMaker specifying min_distance and min_distance_input to some value.

Astro-Kirsty:
This is a very simple solution to resolve #4492

In the end it was a case of documentation. Maybe something further should be added?

Opinions @adonath, @mireianievas , @AtreyeeS

### DIFF
--- a/gammapy/makers/background/reflected.py
+++ b/gammapy/makers/background/reflected.py
@@ -285,14 +285,20 @@ class ReflectedRegionsFinder(RegionsFinder):
     ----------
     angle_increment : `~astropy.coordinates.Angle`, optional
         Rotation angle applied when a region falls in an excluded region.
+        Default is '0.1 rad'.
     min_distance : `~astropy.coordinates.Angle`, optional
-        Minimum rotation angle between two consecutive reflected regions.
+        Minimum rotation angle between two consecutive reflected regions, with respect to the
+        rotation center (typically the pointing position).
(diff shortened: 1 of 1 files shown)


bkhelifi (approved):
Well seen, thanks

registerrier (approved):
Thanks @Astro-Kirsty . LGTM

PR #5927, opened 2025-06-16
Add mask plot for spectral

### LINKED ISSUE #5830: No mask in the figure `MapDataset.plot_residuals`

The function `MapDataset.plot_residuals` does not represent the `safe_mask` or the `fit_mask`, This feature exists for the classes `SpectrumDataset` and `SpectrumDatasetOnOff` (see `SpectrumDataset*.plot_fit()`), using the common function `plot_mask`.

Astro-Kirsty:
This PR is to resolve #5830

Currently the mask is already visible on the `plot_residuals_spatial` however depending on the method you use it is difficult to see it:
image

Screenshot from 2025-06-16 18-25-52

To account for that a simple adjustment of the code is needed on the `mask_fit`
image

### DIFF
--- a/gammapy/datasets/map.py
+++ b/gammapy/datasets/map.py
@@ -1316,7 +1316,15 @@ class MapDataset(Dataset):
         ax = residuals.plot(ax, **kwargs)
         return ax
 
-    def plot_residuals_spectral(self, ax=None, method="diff", region=None, **kwargs):
+    def plot_residuals_spectral(
+        self,
+        ax=None,
+        method="diff",
+        region=None,
+        kwargs_fit=None,
+        kwargs_safe=None,
+        **kwargs,
+    ):
         """Plot spectral residuals.
 
         The residuals are extracted from the provided region, and the normalization
@@ -1392,6 +1400,21 @@ class MapDataset(Dataset):
         ymin = 1.05 * np.nanmin(residuals.data - yerr)
         ymax = 1.05 * np.nanmax(residuals.data + yerr)
         ax.set_ylim(ymin, ymax)
+
+        kwargs_fit = kwargs_fit or {}
+        kwargs_safe = kwargs_safe or {}
+
+        kwargs_fit.setdefault("label", "Mask fit")
+        kwargs_fit.setdefault("color", "tab:green")
+        kwargs_safe.setdefault("label", "Mask safe")
+        kwargs_safe.setdefault("color", "black")
+
+        if self.mask_fit:
+            self.mask_fit.to_region_nd_map().plot_mask(ax=ax, **kwargs_fit)
+
+        if self.mask_safe:
+            self.mask_safe.to_region_nd_map().plot_mask(ax=ax, **kwargs_safe)
+        ax.legend()
         return ax
 
     def plot_residuals(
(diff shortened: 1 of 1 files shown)


Tobychev (dismissed):

AtreyeeS (dismissed):
Thanks @Astro-Kirsty ! 

I think the mask handling section in `Dataset.plot_residuals` keeps coming up repeatedly. It woudl be good to document what exactly is done and why (ie, which masks are applied and how). Can be a separate PR

gammapy/datasets/map.py
```diff
@@ -1309,14 +1309,26 @@ def plot_residuals_spatial(
             mask = self.mask_safe.reduce_over_axes(func=np.logical_or, keepdims=True)
             residuals.data[~mask.data] = np.nan
 
+        if self.mask_fit is not None:
+            mask = self.mask_fit.reduce_over_axes(func=np.logical_or, keepdims=True)
+            residuals.data[~mask.data] = np.nan
```
QRemy: could replace `mask_fit` by `mask` and remove the block about `mask_safe` above

gammapy/datasets/map.py
```diff
@@ -1316,7 +1316,15 @@ def plot_residuals_spatial(
...
         return ax
 
-    def plot_residuals_spectral(self, ax=None, method="diff", region=None, **kwargs):
+    def plot_residuals_spectral(
+        self,
+        ax=None,
+        method="diff",
+        region=None,
+        kwargs_fit=None,
+        kwargs_safe=None,
```
bkhelifi, line 1325: The docstring should be adapted. And the default in lines 1407-10 should be precised.

bkhelifi (changes requested):
Thank you a lot. I have left some comments

AtreyeeS (dismissed):

gammapy/datasets/map.py
```diff
@@ -1316,22 +1316,39 @@ def plot_residuals_spatial(
...
 
+        Both the mask fit and mask safe are taken into account.
+
         The error bars are computed using the uncertainty on the excess with a symmetric assumption.
 
         Parameters
         ----------
         ax : `~matplotlib.axes.Axes`, optional
             Axes to plot on. Default is None.
         method : {"diff", "diff/sqrt(model)"}
```
bkhelifi:
```suggestion
        method : {"diff", "diff/sqrt(model)"}, optional
```

gammapy/datasets/map.py
```diff
@@ -1316,22 +1316,39 @@ def plot_residuals_spatial(
...
-            Normalization used to compute the residuals, see `SpectrumDataset.residuals`. Default is "diff".
+            Normalization used to compute the residuals, see `SpectrumDataset.residuals`.
+            Default is "diff".
         region : `~regions.SkyRegion` (required)
```
bkhelifi: This is strange that this is required and that a default value exists... This is not consistant

bkhelifi (changes requested):
Thanks. This is good for me, except 2 comments on the docstring

gammapy/datasets/map.py
Astro-Kirsty: Yes good pick up, I will investigate that

registerrier (approved):

Issue #5906, opened 2025-06-06
FluxPoints doesn't properly handle a negative spectral model amplitude


samanthalwong:
**Gammapy version**
1.3/2.0-dev

**Bug description**
When a dataset has a negative excess, the spectral model's amplitude will be negative. When passed to FluxPointsEstimator, since the upper limit is calculated on norm, a very negative flux upper limit is calculated.

**Expected behavior**
A negative norm should be returned instead, to give a positive upper limit (when multiplied by the negative limit value, which is expected in the case of a negative excess). Alternatively, an error should be thrown to warn the user to pass a positive spectral model. 

**To Reproduce**
See attached file negativeexcesul.txt. The Segue I files in the VERITAS development dataset may be useful for diagnosing the issue with real data (see: https://zenodo.org/records/14902744).  

[negativeexcesul.txt](https://github.com/user-attachments/files/20628913/negativeexcesul.txt)

**Other information**
This occurs even when you are passing a model with a fixed index and just a variable amplitude, so even users who think that are doing upper limits (by not taking the non-physical index that is generated when fitting to a null signal) will still currently get wrong answers if they are using the Fit to get close to the minimum so the the norm parameter doesn't need to scan a large area. The norm scan behaviour is improved on the dev branch so this might occur less frequently. If the default user behaviour should be passing a very generic positive model with multiple orders of magnitude in norm_scan this should be clarified in a potential tutorial.

AtreyeeS:
Documentation for handling negative amplitudes added in #5933. For now, I propose that we prevent the estimator from running for negative amplitudes, and then investigate the current behaviour for 2.1.

samanthalwong:
That works for me. I think the best solution I've come up with so far is to re-run the flux points estimator with an un-fit spectral model in the case of a negative amplitude, rather than trying to extract the limits from a much larger norm scan.

bkhelifi:
@AtreyeeS, can you give a status here?

Issue #5771, opened 2025-04-08
Tutorial on upper limits


Astro-Kirsty:
A tutorial should be added for the calculation of upper limits, both for flux and parameters.
We should show:
- SED with only UL
- different methods
- expose the stat_scan

For the dataset the two options are:
- use off runs of HESS data release
- create a test dataset from an empty region

bkhelifi:
I can send you some starting draft, @AtreyeeS ... Thanks a lot

Issue #5816, opened 2025-04-16
Removal of check-setup in tutorials


Astro-Kirsty:
In almost every tutorial (if not every one), we use the lines for [check-setup](https://docs.gammapy.org/dev/tutorials/starting/overview.html#check-setup). I am not sure this seems necessary and adds more complexity for users. Maybe instead we could have this written elsewhere (i.e. along with installation?) if people want to check this

bkhelifi:
**The pros:**
From @adonath (dev call 2025/06/20):
- what used to check the setup when people were not used of the virtual env
- useful when building is failing

From @registerrier : useful when there is an issue

**The cons:**
- not needed to have it on each tutorial
- when someone creates an issue, one could require to systematically add its output

PR #5933, opened 2025-06-18
Add tutorial for computing parameter upper limits

AtreyeeS:
This addresses #5771 
This also partly addresses #5906 with documentation. Whether `FluxPointsEstimator` can (should) work with negative amplitudes needs to be understood.

Will need https://github.com/gammapy/gammapy-data/pull/72 to run


examples/tutorials/analysis-3d/non-detection.py
```diff
@@ -0,0 +1,346 @@
+"""
```
bkhelifi: Add the standard line about the Gammapy licence

examples/tutorials/analysis-3d/non-detection.py
```diff
@@ -0,0 +1,346 @@
...
+Event sampling
+==============
+
+Explore how to deal with upper limits on parameters or flux upper limits
+for a non-detected source.
```
bkhelifi: In case of exponential cut-off, the source is detected.

examples/tutorials/analysis-3d/non-detection.py
```diff
@@ -0,0 +1,346 @@
...
+framework before proceeding with this notebook.
+
+Context
+-------
+
+In the study of VHE sources, one often encounters no significant
+detection even after long exposures. In that case, it may be useful to
+compute flux upper limits. Or, given significant detection of a source,
+it may still be difficult to contrain some model parameters, and we can
+only provide some confidence intervals.
```
bkhelifi: This is statistically correct... But it might be confusing for some people. A confidence interval of [-alpha/2, alpha/2] or [-inf, alpha] or [-alpha, +inf] has obviously a different physics meaning.
PS: sometimes, the expression 'confidence limits' is used, but it is not so common

examples/tutorials/analysis-3d/non-detection.py
```diff
@@ -0,0 +1,346 @@
...
+
+
+######################################################################
+# Check setup
+# -----------
+#
+
+from gammapy.utils.check import check_tutorials_setup
+
+check_tutorials_setup()
```
bkhelifi: Is it really needed, here?

examples/tutorials/analysis-3d/non-detection.py
```diff
@@ -0,0 +1,346 @@
...
+
+
+######################################################################
+# When we don’t detect a source
+# -----------------------------
+#
+# As usual, let’s start with some general imports…
```
bkhelifi: There is not import here. You are running the ExcessMapEstimator

examples/tutorials/analysis-3d/non-detection.py
```diff
@@ -0,0 +1,346 @@
...
+# public H.E.S.S. data release and stored it in `GAMMAPY_DATA`
+#
+
+dataset = MapDataset.read("$GAMMAPY_DATA/empty_dl4/empty-dl4.fits.gz")
+dataset.peek()
+plt.show()
+
+
+######################################################################
+# We will first use the `ExcessMapEstimator` for a quick check to see if
```
bkhelifi:
```suggestion
# We will first use the `~gammapy.estimators.ExcessMapEstimator` for a quick check to see if
```

examples/tutorials/analysis-3d/non-detection.py
```diff
@@ -0,0 +1,346 @@
...
+# -----------------------------
+#
+# As usual, let’s start with some general imports…
+#
+
+
+######################################################################
+# Now, load and inspect the dataset. For computational purposes, we have
+# already created a `MapDataset` from observation id 20275 from the
+# public H.E.S.S. data release and stored it in `GAMMAPY_DATA`
```
bkhelifi:
```suggestion
# public H.E.S.S. data release and stored it in `$GAMMAPY_DATA`
```

examples/tutorials/analysis-3d/non-detection.py
```diff
@@ -0,0 +1,346 @@
...
+# When we don’t detect a source
+# -----------------------------
+#
+# As usual, let’s start with some general imports…
+#
+
+
+######################################################################
+# Now, load and inspect the dataset. For computational purposes, we have
+# already created a `MapDataset` from observation id 20275 from the
```
bkhelifi:
```suggestion
# already created a `~gammapy.datasets.MapDataset` from observation id `20275` from the
```

examples/tutorials/analysis-3d/non-detection.py
AtreyeeS: I think there was some discussion on this recently that I didnt follow. @Astro-Kirsty ?

examples/tutorials/analysis-3d/non-detection.py
```diff
@@ -0,0 +1,346 @@
...
+# -----------
+#
+
+from gammapy.utils.check import check_tutorials_setup
+
+check_tutorials_setup()
+
+
+######################################################################
+# When we don’t detect a source
```
bkhelifi: This subsection could be a section about differential and integral flux upper limits.

examples/tutorials/analysis-3d/non-detection.py
```diff
@@ -0,0 +1,346 @@
...
+
+dataset = MapDataset.read("$GAMMAPY_DATA/empty_dl4/empty-dl4.fits.gz")
+dataset.peek()
+plt.show()
+
+
+######################################################################
+# We will first use the `ExcessMapEstimator` for a quick check to see if
+# there are any hotspots in the field. You may also use the
+# `TSMapEstimator`.
```
bkhelifi:
```suggestion
# `~gammapy.estimators.TSMapEstimator`.
```

examples/tutorials/analysis-3d/non-detection.py
```diff
@@ -0,0 +1,346 @@
...
+
+spectral = PowerLawSpectralModel()
+spatial = PointSpatialModel(frame="icrs")
+spatial.lon_0.value = 187.0
+spatial.lat_0.value = 2.6
+spatial.lat_0.min = 1.0
+spatial.lat_0.max = 4.0
+spatial.lon_0.min = 185
+spatial.lon_0.max = 187
+
```
bkhelifi: Maybe you can precise that this is quite important to constrain the range of the position when there is no signal. Otherwise, iminuit might not converge at all

examples/tutorials/analysis-3d/non-detection.py
```diff
@@ -0,0 +1,346 @@
...
+skymodel = SkyModel(spatial_model=spatial, spectral_model=spectral, name="test")
+
+dataset.models = skymodel
+fit = Fit()
+res = fit.run(dataset)
+print(res.models)
+
+
+######################################################################
+# We can see that there is a slight negative excess in the centre, and
```
bkhelifi: To say that, it would propose to plot the excess map before.

examples/tutorials/analysis-3d/non-detection.py
```diff
@@ -0,0 +1,346 @@
...
+# thus, the fitted model has a negative amplitude. We can use the
+# `select_nested_models` function to perform a likelihood ratio test to
```
bkhelifi:
```suggestion
# `~gammapy.modeling.select_nested_models` function to perform a likelihood ratio test to
```

examples/tutorials/analysis-3d/non-detection.py
AtreyeeS: Where should this license lie? I dont see it in our current tutorials, eg: https://github.com/gammapy/gammapy/blob/main/examples/tutorials/analysis-time/light_curve_simulation.py

examples/tutorials/analysis-3d/non-detection.py
```diff
@@ -0,0 +1,346 @@
+"""
+Event sampling
+==============
+
+Explore how to deal with upper limits on parameters or flux upper limits
```
AtreyeeS:
```suggestion
Explore how to deal with upper limits on parameters for a significant source,  or flux upper limits
```

examples/tutorials/analysis-3d/non-detection.py
```diff
@@ -0,0 +1,346 @@
...
+fit = Fit()
+res = fit.run(dataset)
+print(res.models)
+
+
+######################################################################
+# We can see that there is a slight negative excess in the centre, and
+# thus, the fitted model has a negative amplitude. We can use the
+# `select_nested_models` function to perform a likelihood ratio test to
+# see if this is significant.
```
bkhelifi: What is 'this'? the negative excess or the determine the 'source significance' (one can here refer to corresponding HowTo section. This add will remind to the users that the HowTo is useful)

examples/tutorials/analysis-3d/non-detection.py
AtreyeeS:
```suggestion
# 
```

examples/tutorials/analysis-3d/non-detection.py
```diff
@@ -0,0 +1,346 @@
...
+
+LLR = select_nested_models(
+    Datasets(dataset), parameters=[skymodel.parameters["amplitude"]], null_values=[0]
+)
+print(LLR)
+
+
+######################################################################
+# You can see that the `ts ~ 4.7`, thus suggesting that the observed
+# fluctuations are not significant over the background. Now, we will
```
bkhelifi: Maybe on can precise that there is here only 1 dof, so sqrt(ts)*sign(ts) gives directly the significance

examples/tutorials/analysis-3d/non-detection.py
```diff
@@ -0,0 +1,346 @@
...
+res1 = estimator.run(dataset)
+res1.sqrt_ts.plot(add_cbar=True)
+plt.show()
+
+
+######################################################################
+# The significance map looks rather flat! Suppose we were expecting a
+# source at the centre of our map. Lets try see if we can fit a point
+# source there.
+#
```
AtreyeeS:
```suggestion
# Note that it is necessary to constrain the range of the position, otherwise the fit might not converge. 
```

examples/tutorials/analysis-3d/non-detection.py
```diff
@@ -0,0 +1,346 @@
...
+LLR = select_nested_models(
+    Datasets(dataset), parameters=[skymodel.parameters["amplitude"]], null_values=[0]
+)
+print(LLR)
+
+
+######################################################################
+# You can see that the `ts ~ 4.7`, thus suggesting that the observed
+# fluctuations are not significant over the background. Now, we will
+# estimate the differential upper limits of the source.
```
bkhelifi: I propose to add the computation of the integral UL. When we have no detection, this is common to give this number

examples/tutorials/analysis-3d/non-detection.py
```diff
@@ -0,0 +1,346 @@
...
+#
+# Here, it is important to **set a reasonable model** on the dataset
+# before proceeding with the estimator. This model can come from
+# measurements from other instruments, be an extrapolation of the flux
+# observed at other wavelengths, come from theoretical estimations, etc. A
+# model with a negative amplitude as obtained above should should not be
```
bkhelifi:
```suggestion
# model with a negative amplitude as obtained above should not be
```

examples/tutorials/analysis-3d/non-detection.py
```diff
@@ -0,0 +1,346 @@
...
+# used.
+#
+# Note that the computed upper limits can depend on the assumed model. The
```
bkhelifi:
```suggestion
# Note that the **computed upper limits depend on the assumed model**, in general on the spectral shape and the spatial model. The
```

examples/tutorials/analysis-3d/non-detection.py
```diff
@@ -0,0 +1,346 @@
...
+# values of the amplitude should not matter as long as the values are not
```
bkhelifi: I do not understand this sentance

AtreyeeS: What I meant is that the computed UL can differ depending on the source model, say a power law vs an exponential cutoff. Since the value of the normalisation is adjusted in each bin, the ULs generally dont change with the amplitude value, unless of course one puts a value very far away from reality. I will try to rephrase

examples/tutorials/analysis-3d/non-detection.py
```diff
@@ -0,0 +1,346 @@
...
+# too absurd. Here, we compute the 3-simga upper limits for assuming a
```
bkhelifi:
```suggestion
# too absurd. Here, we compute the 3-sigma upper limits for assuming a
```

examples/tutorials/analysis-3d/non-detection.py
```diff
@@ -0,0 +1,346 @@
...
+# spectral index of 2.0
+#
```
bkhelifi: I have an open question... people always forget that UL corresponds to the interval [-inf, alpha], ie alpha=P(X<UL).
The link between probability and sigma is then different from the usual confidence interval, ie alpha=P(-Err<X<Err). I think that this is important to have a function providing the link between the estimator's `n_sigma_ul`  and the probability. Then, one can also show how to compute a 90% UL.
With @registerrier , we are currently thinking about the correct way to do it (in a separate PR).

examples/tutorials/analysis-3d/non-detection.py
```diff
@@ -0,0 +1,346 @@
...
+
+dataset.models = model1
+fp1 = fp_est.run(dataset)
+
+
+fp1.plot(sed_type="dnde")
+plt.show()
+
+
+######################################################################
```
bkhelifi: This could be a sub(sub)section... the sensitivity estimation.
It is important also to precise that here no systematics or conditions are used for the sensitivity estimation.

examples/tutorials/analysis-3d/non-detection.py
```diff
@@ -0,0 +1,346 @@
...
+# We can then ask, would I have seen my source given this irf/ exposure
+# time? The `FluxPointsEstimator` can be used to obtain the sensitivity,
```
bkhelifi:
```suggestion
# time? The `~gammapy.estimators.FluxPointsEstimator` can be used to obtain the sensitivity,
```

examples/tutorials/analysis-3d/non-detection.py
```diff
@@ -0,0 +1,346 @@
...
+
+
+######################################################################
+# Constraining model upper limits
+# -------------------------------
+#
+# We will now use a precomputed blazar dataset to see how to contrain
+# model parameters. Detailed modeling of this dataset may be found in the
+# :doc:`/tutorials/utorials/analysis-1d/ebl` notebook. We will try to
+# constrain the spectral cutoff in the source. To see
```
bkhelifi: typo: `To see` ;)

examples/tutorials/analysis-3d/non-detection.py
```diff
@@ -0,0 +1,346 @@
...
+
+res_pks = fit.run(dataset_onoff)
+print(res_pks.models)
+
+
+######################################################################
+# We see that the parameter `lambda_` is not well constrained. In this
+# case, it is helpful to look at the likelihood profile of the parameter.
+# We will use the `fit.stat_profile` function, and refit the other free
+# parameters in the process by setting `reoptimize=True`
```
bkhelifi:
```suggestion
# parameters in the process by setting `reoptimize=True`.
```

examples/tutorials/analysis-3d/non-detection.py
```diff
@@ -0,0 +1,346 @@
...
+secay = ax.secondary_yaxis("right", functions=(ts_to_sigma, sigma_to_ts))
+secay.set_ylabel("Significance [$\sigma$]")
+plt.title("Cutoff likelihood profile", fontsize=20, y=1.05)
+plt.show()
+
+
+######################################################################
+# Thus, this dataset does not yield a significant spectral cutoff in
+# PKS2155-304. In particular, we cannot constrain the lower limit from
+# this profile. We can compute the limits of the cutoff from the pdf.
```
bkhelifi:
```suggestion
# this profile. We can compute the limits of the cutoff from the Probability Density Function (PDF).
```

examples/tutorials/analysis-3d/non-detection.py
```diff
@@ -0,0 +1,346 @@
...
+
+ax.axhline(9, ls="dashed", color="black")
+ax.axvline(1.0 / UL, ls="dotted", color="brown")
+plt.title("Cutoff likelihood profile", fontsize=20, y=1.05)
+plt.xscale("log")
+
+
+######################################################################
+# This logic can be extended to any spectral or spatial feature. As an
+# exercise, try to compute the 95% spatial extent on the MSH 1552 dataset
```
bkhelifi:
```suggestion
# exercise, try to compute the 95% spatial extent on the MSH 15-52 dataset
```

examples/tutorials/analysis-3d/non-detection.py
bkhelifi: I know, an historical oversight;) All Gammapy files should contain at the beginning:
`# Licensed under a 3-clause BSD style license - see LICENSE.rst`

examples/tutorials/analysis-3d/non-detection.py
bkhelifi: Sorry, my question was not clear... I was speaking about:
`The values of the amplitude should not matter as long as the values are not too absurd.`
I think that the sentence is correct, but not well placed... Maybe
1. should your spectral shape with an index
2. then you can place your sentence about the normalisation

examples/tutorials/analysis-3d/non-detection.py
Astro-Kirsty: I'm not sure that we really made a decision (#5816), we can discuss at the dev call

examples/tutorials/analysis-3d/non-detection.py
```diff
@@ -0,0 +1,370 @@
...
+if LL > parameter.value:
+    print("lower limit is not constrained")
+if UL < parameter.value:
+    print("upper limit is not constrained")
+
+unit = parameter.unit**-1  # Give in units of energy since lambda_ = 1/cutoff energy
+print(
+    f"{conf_level*100:.2f}% lower limit of energy cutoff: {1/UL:.4f} ({unit.to_string()})"
+)
+# print(f"{conf_level*100:.2f}% upper limit: {1/LL:.4f} ({unit.to_string()})")
```
AtreyeeS:
```suggestion
```

examples/tutorials/analysis-3d/non-detection.py
```diff
@@ -0,0 +1,369 @@
...
+# Sensitivity estimation
+# ------------------
+#
+# We can then ask, would I have seen my source given this irf/ exposure
+# time? The `~gammapy.estimators.FluxPointsEstimator` can be used to obtain the sensitivity,
+# which can be compared to the expected flux. We have the 5-sigma
+# sensitivity here, which can be configured using `n_sigma_sensitivity`
+# on init. Lets see if we would have seen if a Crab-like source was
+# present in the center.
+# Note that this computed sensitivity does not take into account the into factors like the minimum number of gamma-ray, etc (see :doc:`/tutorials/analysis-1d/cta_sensitivity.py`) and is dependent on the analysis configuration.
```
AtreyeeS:
```suggestion
# Note that this computed sensitivity does not take into account the into factors like the minimum number of gamma-ray, etc
# (see :doc:`/tutorials/analysis-1d/cta_sensitivity.py`) and is dependent on the analysis configuration.
```

examples/tutorials/analysis-3d/non-detection.py
```diff
@@ -0,0 +1,370 @@
...
+#
+
+LLR = select_nested_models(
+    Datasets(dataset), parameters=[skymodel.parameters["amplitude"]], null_values=[0]
+)
+print(LLR)
+
+
+######################################################################
+# You can see that the `ts ~ 4.7`, thus suggesting that the observed
```
AtreyeeS:
```suggestion
# You can see that the `ts ~ 4.7`. Since we have only 1 free parameter here, the significance is directly the 
# sqrt(ts) ~ 2.16, thus suggesting that the observed
```

examples/tutorials/analysis-3d/non-detection.py
bkhelifi: @registerrier: maybe a function based on the following could help to get the n_sigma_ul from a probability:
```
from scipy import stats
X = stats.Normal()
X.icdf([0.68, 0.9, 0.95])

```

examples/tutorials/analysis-3d/non-detection.py
```diff
@@ -0,0 +1,371 @@
...
+#
+
+from gammapy.utils.check import check_tutorials_setup
+
+check_tutorials_setup()
+
+
+######################################################################
+# Flux upper limits for non-detected sources
+# -----------------------------
```
bkhelifi:
```suggestion
# Flux upper limits for non-detected sources
# --------------------------------------------------------------
```

examples/tutorials/analysis-3d/non-detection.py
```diff
@@ -0,0 +1,371 @@
...
+fp1 = fp_est.run(dataset)
+
+
+fp1.plot(sed_type="dnde")
+plt.show()
+
+
+######################################################################
+# Sensitivity estimation
+# ------------------
```
bkhelifi:
```suggestion
# Sensitivity estimation
# -------------------------------
```

examples/tutorials/analysis-3d/non-detection.py
```diff
@@ -0,0 +1,371 @@
...
+print(
+    f"{conf_level*100:.2f}% lower limit of energy cutoff: {1/UL:.4f} ({unit.to_string()})"
+)
+
+
+######################################################################
+# Since our dataset actually starts from `200 GeV`, all that we did from
+# this analysis is rule out any cut-off features!
+#
+# We could also have used `stat_profile_ul_scipy` for computing the UL, which uses rootfing to obtain the n-sigma limits, but this can fail if the limits are not constrained as we have here.
```
AtreyeeS:
```suggestion
# We could also have used `stat_profile_ul_scipy` for computing the UL, which uses rootfinding to obtain the n-sigma limits, but this can fail if the limits are not constrained as we have here.
```

examples/tutorials/analysis-3d/non-detection.py
```diff
@@ -0,0 +1,371 @@
...
+#
+# We could also have used `stat_profile_ul_scipy` for computing the UL, which uses rootfinding to obtain the n-sigma limits, but this can fail if the limits are not constrained as we have here.
+#
+
+ax = plt.gca()
+ax.plot(1.0 / values, loglike - np.min(loglike))
+ax.set_xlabel("Energy cutoff (TeV)")
+ax.set_ylabel(r"$\Delta$TS")
+
+ax.axhline(9, ls="dashed", color="black")
```
AtreyeeS:
```suggestion
```

examples/tutorials/analysis-3d/non-detection.py
```diff
@@ -0,0 +1,370 @@
...
+
+ax = plt.gca()
+ax.plot(1.0 / values, loglike - np.min(loglike))
+ax.set_xlabel("Energy cutoff (TeV)")
+ax.set_ylabel(r"$\Delta$TS")
+
+ax.axvline(1.0 / UL, ls="dotted", color="brown")
+plt.title("Cutoff likelihood profile", fontsize=20, y=1.05)
+plt.xscale("log")
+
```
AtreyeeS:
```suggestion
plt.show()
```

AtreyeeS:
I split this into two tutorials as discussed in the dev call.
It might have been better to have this as two separate PRs, but I will be travelling for the next two weeks and wont have any time now

examples/tutorials/analysis-1d/parameter_ul.py
```diff
@@ -0,0 +1,214 @@
...
+"""
+Constraining parameter limits
+================
+
+Explore how to deal with upper limits on parameters
+
+Prerequisites
+-------------
+It is advisable to understand the general Gammapy modelling and fitting framework before proceeding
+with this notebook.
```
Astro-Kirsty: Could add some references to other parts of the documentation here

examples/tutorials/analysis-1d/parameter_ul.py
```diff
@@ -0,0 +1,214 @@
...
+
+Context
+-------
+Given significant detection of a source, it may still be difficult to constrain some model parameters,
+and we can only provide some confidence intervals.
```
Astro-Kirsty:
```suggestion
Even with significant detection of a source, constraining specific model parameters may remain difficult,
allowing only for the calculation of confidence intervals.
```

examples/tutorials/analysis-1d/parameter_ul.py
```diff
@@ -0,0 +1,214 @@
...
+
+Proposed approach
+-----------------
+
+In this section, we will use 6 observations of the blazars PKS 2155-304 taken in 2008 by
+  H.E.S.S to constrain the curvature in the spectrum
```
Astro-Kirsty:
```suggestion
In this section, we will use 6 observations of the blazar PKS 2155-304, taken in 2008 by
H.E.S.S, to constrain the curvature in the spectrum
```

examples/tutorials/analysis-1d/parameter_ul.py
```diff
@@ -0,0 +1,214 @@
...
+check_tutorials_setup()
+
+######################################################################
+# Load observation
+# -----------
+#
+# We will now use a precomputed blazar dataset to see how to contrain
+# model parameters. Detailed modeling of this dataset may be found in the
+# :doc:`/tutorials/utorials/analysis-1d/ebl` notebook. We will try to
+# constrain the spectral cutoff in the source. To see
```
Astro-Kirsty:
```suggestion
# :doc:`/tutorials/utorials/analysis-1d/ebl` notebook. Here, we will try to
# constrain the spectral cutoff in the source. 
```

examples/tutorials/analysis-1d/parameter_ul.py
```diff
@@ -0,0 +1,214 @@
...
+)
+dataset_onoff.peek()
+plt.show()
+
+
+######################################################################
+# Fit spectrum
+# -----------
+#
+# Fit a spectral model with cutoff
```
Astro-Kirsty:
```suggestion
# Fit a spectral model with a cutoff
```

examples/tutorials/analysis-1d/parameter_ul.py
```diff
@@ -0,0 +1,214 @@
...
+model_pks = SkyModel(spectral1, name="model_pks")
+dataset_onoff.models = model_pks
+
+fit = Fit()
+res_pks = fit.run(dataset_onoff)
+print(res_pks.models)
+
+
+######################################################################
+# We see that the parameter `lambda_` is not well constrained. In this
```
Astro-Kirsty: Comment on what you mean by that

examples/tutorials/analysis-1d/parameter_ul.py
```diff
@@ -0,0 +1,214 @@
...
+# case, it is helpful to look at the likelihood profile of the parameter.
```
Astro-Kirsty: Again, why is that useful? We should be very explicit because people will comes to this tutorial to understand not only how to do this, but why do

examples/tutorials/analysis-1d/parameter_ul.py
```diff
@@ -0,0 +1,214 @@
...
+# We will use the `fit.stat_profile` function, and refit the other free
```
Astro-Kirsty:
```suggestion
# We will use the `~gammapy.modeling.Fit.stat_profile` function, and refit the other free
```

examples/tutorials/analysis-1d/parameter_ul.py
```diff
@@ -0,0 +1,214 @@
...
+parameter.scan_n_values = 25
+parameter.scan_min = 0.05
+parameter.scan_max = 5
+parameter.interp = "log"
+profile = fit.stat_profile(datasets=dataset_onoff, parameter=parameter, reoptimize=True)
+
+
+######################################################################
+# `profile` is a dict storing the values of `lambda_` and the
+# likelihood value. Lets try to visualise it.
```
Astro-Kirsty:
```suggestion
# `profile` is a dictionary that stores the values of `lambda_` and the
# corresponding likelihood values. Lets try to visualise it.
```

examples/tutorials/analysis-1d/parameter_ul.py
```diff
@@ -0,0 +1,214 @@
...
+secay.set_ylabel("Significance [$\sigma$]")
+plt.title("Cutoff likelihood profile", fontsize=20, y=1.05)
+plt.show()
+
+
+######################################################################
+# Constrain the limits on the parameter
+# ---------------
+# Thus, this dataset does not yield a significant spectral cutoff in
+# PKS2155-304. In particular, we cannot constrain the lower limit from
```
Astro-Kirsty: Why? Because the shape of the profile or?

examples/tutorials/analysis-1d/parameter_ul.py
```diff
@@ -0,0 +1,214 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+
+"""
+Constraining parameter limits
+================
```
Astro-Kirsty:
```suggestion
Constraining parameter limits
=============================
```

Needs to be the same length as the title

examples/tutorials/analysis-1d/parameter_ul.py
```diff
@@ -0,0 +1,214 @@
...
+# -----------
+#
+
+from gammapy.utils.check import check_tutorials_setup
+
+check_tutorials_setup()
+
+######################################################################
+# Load observation
+# -----------
```
Astro-Kirsty:
```suggestion
# Load observation
# ----------------
```

examples/tutorials/analysis-1d/parameter_ul.py
```diff
@@ -0,0 +1,214 @@
...
+dataset_onoff = SpectrumDatasetOnOff.read(
+    "$GAMMAPY_DATA/PKS2155-steady/pks2155-304_steady.fits.gz"
+)
+dataset_onoff.peek()
+plt.show()
+
+
+######################################################################
+# Fit spectrum
+# -----------
```
Astro-Kirsty:
```suggestion
# Fit spectrum
# ------------
```

examples/tutorials/analysis-1d/parameter_ul.py
```diff
@@ -0,0 +1,214 @@
...
+ax.set_ylabel(r"$\Delta$TS")
+secay = ax.secondary_yaxis("right", functions=(ts_to_sigma, sigma_to_ts))
+secay.set_ylabel("Significance [$\sigma$]")
+plt.title("Cutoff likelihood profile", fontsize=20, y=1.05)
+plt.show()
+
+
+######################################################################
+# Constrain the limits on the parameter
+# ---------------
```
Astro-Kirsty:
```suggestion
# Constrain the limits on the parameter
# -------------------------------------
```

examples/tutorials/analysis-1d/parameter_ul.py
```diff
@@ -0,0 +1,214 @@
...
+
+unit = parameter.unit**-1  # Give in units of energy since lambda_ = 1/cutoff energy
+print(
+    f"{conf_level*100:.2f}% lower limit of energy cutoff: {1/UL:.4f} ({unit.to_string()})"
+)
+
+
+######################################################################
+# Since our dataset actually starts from `200 GeV`, all that we did from
+# this analysis is rule out any cut-off features!
```
Astro-Kirsty:
```suggestion
# Since our dataset actually starts from 200 GeV, all this analysis has 
# done is rule out any cut-off features!
```

examples/tutorials/analysis-1d/parameter_ul.py
```diff
@@ -0,0 +1,214 @@
...
+#
+# We could also have used `stat_profile_ul_scipy` for computing the UL, which uses rootfing to obtain the n-sigma limits,
+# but this can fail if the limits are not constrained.
```
Astro-Kirsty: Do we still want to mention how that would be called though?

examples/tutorials/analysis-1d/parameter_ul.py
```diff
@@ -0,0 +1,214 @@
...
+print(
+    f"{conf_level*100:.2f}% lower limit of energy cutoff: {1/UL:.4f} ({unit.to_string()})"
+)
+
+
+######################################################################
+# Since our dataset actually starts from `200 GeV`, all that we did from
+# this analysis is rule out any cut-off features!
+#
+# We could also have used `stat_profile_ul_scipy` for computing the UL, which uses rootfing to obtain the n-sigma limits,
```
Astro-Kirsty:
```suggestion
# We could also have used `~gammapy.modeling.scipy.stat_profile_ul_scipy` for computing the UL, 
which uses root finding to obtain the n-sigma limits,
```

AtreyeeS:
Ok, adapted! Now this has only the notebook for paramter ul

examples/tutorials/analysis-1d/parameter_ul.py
```diff
@@ -0,0 +1,218 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+
+"""
+Constraining parameter limits
+=============================
+
+Explore how to deal with upper limits on parameters
```
bkhelifi:
```suggestion
Explore how to deal with upper limits on parameters.
```

examples/tutorials/analysis-1d/parameter_ul.py
```diff
@@ -0,0 +1,218 @@
...
+
+Prerequisites
+-------------
+It is advisable to understand the general Gammapy modelling and fitting framework before proceeding
+with this notebook, eg see :doc:`docs/user-guide/modeling`
```
bkhelifi:
```suggestion
with this notebook, eg see :doc:`/docs/user-guide/modeling`.
```

examples/tutorials/analysis-1d/parameter_ul.py
```diff
@@ -0,0 +1,218 @@
...
+Context
+-------
+Even with significant detection of a source, constraining specific model parameters may remain difficult,
+allowing only for the calculation of confidence intervals.
+
+Proposed approach
+-----------------
+
+In this section, we will use 6 observations of the blazar PKS 2155-304, taken in 2008 by
+H.E.S.S, to constrain the curvature in the spectrum
```
bkhelifi:
```suggestion
H.E.S.S, to constrain the curvature in the spectrum.
```

examples/tutorials/analysis-1d/parameter_ul.py
```diff
@@ -0,0 +1,218 @@
...
+check_tutorials_setup()
+
+######################################################################
+# Load observation
+# ----------------
+#
+# We will now use a precomputed blazar dataset to see how to contrain
+# model parameters. Detailed modeling of this dataset may be found in the
+# :doc:`/tutorials/analysis-1d/ebl` notebook. We will try to
+# constrain the spectral cutoff in the source. To see
```
bkhelifi: "To see" is not a remaining of an old version?

examples/tutorials/analysis-1d/parameter_ul.py
bkhelifi: And you can remind what is lambda?

examples/tutorials/analysis-1d/parameter_ul.py
```diff
@@ -0,0 +1,218 @@
...
+
+unit = parameter.unit**-1  # Give in units of energy since lambda_ = 1/cutoff energy
+print(
+    f"{conf_level*100:.2f}% lower limit of energy cutoff: {1/UL:.4f} ({unit.to_string()})"
+)
+
+
+######################################################################
+# Since our dataset actually starts from 200 GeV, all this analysis has
+# done is rule out any cut-off features at 68% confidence!
```
bkhelifi: is there some english phrasing issue, here?

examples/tutorials/analysis-1d/parameter_ul.py
```diff
@@ -0,0 +1,218 @@
...
+ax.set_xlabel("Energy cutoff (TeV)")
+ax.set_ylabel(r"$\Delta$TS")
+ax.axvline(1.0 / UL, ls="dotted", color="brown")
+plt.title("Cutoff likelihood profile", fontsize=20, y=1.05)
+plt.xscale("log")
+
+
+######################################################################
+# This logic can be extended to any spectral or spatial feature. As an
+# exercise, try to compute the 95% spatial extent on the MSH 15 52 dataset
```
bkhelifi:
```suggestion
# exercise, try to compute the 95% spatial extent on the MSH 15-52 dataset
```

examples/tutorials/analysis-1d/parameter_ul.py
```diff
@@ -0,0 +1,218 @@
...
+# We could also have used `stat_profile_ul_scipy` for computing the UL, which uses
+# rootfinding to obtain the n-sigma limits,
+# but note that this can fail if the limits are not constrained.
+#
+# We now plot, in energy units, the likelihood profile of the cutoff with a brown dotted
+# line showing the lower cutoff.
+
+ax = plt.gca()
+ax.plot(1.0 / values, loglike - np.min(loglike))
+ax.set_xlabel("Energy cutoff (TeV)")
```
bkhelifi:
```suggestion
ax.set_xlabel("Energy cutoff [TeV]")
```

examples/tutorials/analysis-1d/parameter_ul.py
```diff
@@ -0,0 +1,218 @@
...
+# `profile` is a dictionary that stores the values of `lambda_` and the
+# corresponding likelihood values. Let's try to visualise it.
+#
+
+values = profile["model_pks.spectral.lambda__scan"]
+loglike = profile["stat_scan"]
+
+ax = plt.gca()
+ax.plot(values, loglike - np.min(loglike))
+ax.set_xlabel("Cutoff value (1/TeV)")
```
bkhelifi:
```suggestion
ax.set_xlabel("Cutoff value [1/TeV]")
```

examples/tutorials/analysis-1d/parameter_ul.py
```diff
@@ -0,0 +1,218 @@
...
+    """
+    cdf = cumulative_trapezoid(pdf, theta_values, initial=0)
+    cdf_interp = CubicSpline(cdf, theta_values)
+    lower_limit = cdf_interp(1 - confidence_level)
+    return float(lower_limit)
+
+
+pdf = likelihood_to_pdf(values, loglike)
+plt.plot(values, pdf)
+plt.xlabel("Cutoff value (1/TeV)")
```
bkhelifi:
```suggestion
plt.xlabel("Cutoff value [1/TeV]")
```

examples/tutorials/analysis-1d/parameter_ul.py
```diff
@@ -0,0 +1,218 @@
...
+plt.ylabel("pdf")
```
bkhelifi:
```suggestion
plt.ylabel("PDF")
```

examples/tutorials/analysis-1d/parameter_ul.py
```diff
@@ -0,0 +1,218 @@
...
+plt.show()
+
+conf_level = 0.68
+# Compute one-sided limits
+UL = compute_upper_limit(values, pdf, confidence_level=conf_level)
+LL = compute_lower_limit(values, pdf, confidence_level=conf_level)
+print("lower limit, upper limit: ", LL, UL)
```
bkhelifi:
```suggestion
print(f"lower limit, upper limit: {LL:.3f}, {UL:.3f}")
```

examples/tutorials/analysis-1d/parameter_ul.py
```diff
@@ -0,0 +1,218 @@
...
+import matplotlib.pyplot as plt
+
+from gammapy.datasets import SpectrumDatasetOnOff
+from gammapy.modeling import Fit
+from gammapy.modeling.models import (
+    SkyModel,
+    ExpCutoffPowerLawSpectralModel,
+)
+from gammapy.stats.utils import ts_to_sigma, sigma_to_ts
+import numpy as np
```
registerrier: Maybe move above.

examples/tutorials/analysis-1d/parameter_ul.py
```diff
@@ -0,0 +1,218 @@
...
+
+
+######################################################################
+# Check setup
+# -----------
+#
+
+from gammapy.utils.check import check_tutorials_setup
+
+check_tutorials_setup()
```
registerrier: This we have discussed to remove in most cases except for a few tutorials. I would keep it out here.

examples/tutorials/analysis-1d/parameter_ul.py
```diff
@@ -0,0 +1,218 @@
...
+
+######################################################################
+# Load observation
+# ----------------
+#
+# We will now use a precomputed blazar dataset to see how to contrain
```
registerrier: spectrum dataset?

registerrier:
```suggestion
# We will now use a precomputed blazar dataset to see how to constrain
```

examples/tutorials/analysis-1d/parameter_ul.py
```diff
@@ -0,0 +1,218 @@
...
+
+values = profile["model_pks.spectral.lambda__scan"]
+loglike = profile["stat_scan"]
+
+ax = plt.gca()
+ax.plot(values, loglike - np.min(loglike))
+ax.set_xlabel("Cutoff value (1/TeV)")
+ax.set_ylabel(r"$\Delta$TS")
+secay = ax.secondary_yaxis("right", functions=(ts_to_sigma, sigma_to_ts))
+secay.set_ylabel("Significance [$\sigma$]")
```
registerrier: while this is a nice way to show the significance, it implicitly assumes that there is one dof here. In a difference situation one would have to create a `lambda` function to change the default `df` value of `ts_to_sigma`.

This should be better explained. Conversely you could also simply have one axis with `sqrt(TS)`.

examples/tutorials/analysis-1d/parameter_ul.py
```diff
@@ -0,0 +1,218 @@
...
+# corresponding likelihood values. Let's try to visualise it.
+#
+
+values = profile["model_pks.spectral.lambda__scan"]
+loglike = profile["stat_scan"]
+
+ax = plt.gca()
+ax.plot(values, loglike - np.min(loglike))
+ax.set_xlabel("Cutoff value (1/TeV)")
+ax.set_ylabel(r"$\Delta$TS")
```
registerrier: Here we need to be careful to how we write things.

In general, in Gammapy, we have tried to use the test statistic TS as delta(-2 log(L)). (see e.g. CountsStatistic)

We should try to be as homogeneous as possible. To me:
- fit statistic (total stat or stat_sum)
- TS : difference of fit statistics between H1 and H0

examples/tutorials/analysis-1d/parameter_ul.py
```diff
@@ -0,0 +1,218 @@
...
+secay = ax.secondary_yaxis("right", functions=(ts_to_sigma, sigma_to_ts))
+secay.set_ylabel("Significance [$\sigma$]")
+plt.title("Cutoff likelihood profile", fontsize=20, y=1.05)
+plt.show()
+
+
+######################################################################
+# Constrain the limits on the parameter
```
registerrier:
```suggestion
# Estimate the limits on the parameter
```

examples/tutorials/analysis-1d/parameter_ul.py
registerrier: OK my bad, this is really a delta TS that you show. 
But then it is really strange to show the significance on the right y axis. Because this is not giving a significance or in a twisted way (i.e. how significant in the best value compared to this specific value as null hypothesis.)

examples/tutorials/analysis-1d/parameter_ul.py
```diff
@@ -0,0 +1,218 @@
...
+######################################################################
+# We see that the parameter `lambda_` is not well constrained. In this
+# case, it can be helpful to look at the likelihood profile of the parameter.
+# We will use the `~gammapy.modeling.Fit.stat_profile` function, and refit the other free
+# parameters in the process by setting `reoptimize=True`
+#
+
+parameter = model_pks.parameters["lambda_"]
+parameter.scan_n_values = 25
+parameter.scan_min = 0.05
```
registerrier: why not go to zero ? or below?

examples/tutorials/analysis-1d/parameter_ul.py
registerrier: Agreed, the profile does not tell you directly if it is significant, or a minima you have to explain that it shows that the best fit improves the fit statistic by a certain amount compared to the `lambda_` parameter value in the null hypothesis.
You might add a vertical line to show the position of the null hypothesis

examples/tutorials/analysis-1d/parameter_ul.py
```diff
@@ -0,0 +1,218 @@
...
+)
+
+
+######################################################################
+# Since our dataset actually starts from 200 GeV, all this analysis has
+# done is rule out any cut-off features at 68% confidence!
+#
+# We could also have used `stat_profile_ul_scipy` for computing the UL, which uses
+# rootfinding to obtain the n-sigma limits,
+# but note that this can fail if the limits are not constrained.
```
registerrier: I would not suggest additional approaches...

examples/tutorials/analysis-1d/parameter_ul.py
```diff
@@ -0,0 +1,218 @@
...
+
+
+######################################################################
+# Constrain the limits on the parameter
+# -------------------------------------
+# Thus, this dataset does not yield a significant spectral cutoff in
+# PKS2155-304. In particular, we cannot constrain the lower limit from
+# this profile. We can compute the limits of the cutoff from the Probability Density Function (PDF).
+# For this, we will first convert the likelihood profile into a normalised PDF with `likelihood_to_pdf`
+# and then use it to compute the intervals from the Cumulative Distribution Function (CDF).
```
registerrier: This whole section is not correct. The exponential of the likelihood profile does not yield a pdf, because likelihoods are not probabilities.
Technically, the profile likelihood is not even a likelihood but usually behaves like one (hence the validity of measuring confidence intervals the way we do). See:
- https://myweb.uiowa.edu/pbreheny/7110/f20/notes/11-04.pdf
- https://arxiv.org/pdf/1903.00162

In general, we should remember that profile likelihoods can be biased. They don't necessarily give the correct minimum when there are many nuisance parameters and small statistics. So finding the best fit value from the profile likelihood is not the first we should show.

registerrier (changes requested):
Thanks @AtreyeeS . I would suggest to re-organize to avoid confusion with statistics.

-perform a direct best fit comparison of H0 and H1, possibly with nested_hypothesis or manually to properly show how to quantify parameter significance wrt to null value. 
- since not significant compute lower limit on `lambda_` with the `Fit.confidence`. Note that here we must be careful of what n_sigma means once converted to p-value
- show the resulting UL on the likelihood profile

One possible question is the case of a parameter with a null value on the boundary of the allowed range. This is not the simplest case. You could show instead the beta parameter of a log parabola.

examples/tutorials/analysis-1d/parameter_ul.py
```diff
@@ -0,0 +1,218 @@
...
+)
+
+
+######################################################################
+# Since our dataset actually starts from 200 GeV, all this analysis has
+# done is rule out any cut-off features at 68% confidence!
+#
+# We could also have used `stat_profile_ul_scipy` for computing the UL, which uses
+# rootfinding to obtain the n-sigma limits,
+# but note that this can fail if the limits are not constrained.
```
QRemy:
```suggestion
```

examples/tutorials/analysis-1d/parameter_ul.py
```diff
@@ -0,0 +1,218 @@
...
+
+unit = parameter.unit**-1  # Give in units of energy since lambda_ = 1/cutoff energy
+print(
+    f"{conf_level*100:.2f}% lower limit of energy cutoff: {1/UL:.4f} ({unit.to_string()})"
+)
+
+
+######################################################################
+# Since our dataset actually starts from 200 GeV, all this analysis has
+# done is rule out any cut-off features at 68% confidence!
```
QRemy:
```suggestion
# Since our dataset actually starts from 200 GeV, this analysis
#  rule out a cut-off at 68% confidence in the energy range considered.
```

examples/tutorials/analysis-1d/parameter_ul.py
```diff
@@ -0,0 +1,215 @@
...
+check_tutorials_setup()
+
+######################################################################
+# Load observation
+# ----------------
+#
+# We will now use a precomputed blazar dataset to see how to constrain
+# model parameters. Detailed modeling of this dataset may be found in the
+# :doc:`/tutorials/analysis-1d/ebl` notebook. We will try to
+# constrain the spectral cutoff in the source. To see
```
QRemy:
```suggestion
# constrain the spectral cutoff in the source. 
```

examples/tutorials/analysis-1d/parameter_ul.py
```diff
@@ -0,0 +1,215 @@
...
+
+
+######################################################################
+# Check setup
+# -----------
+#
+
+from gammapy.utils.check import check_tutorials_setup
+
+check_tutorials_setup()
```
QRemy:
```suggestion
```

examples/tutorials/analysis-1d/parameter_ul.py
```diff
@@ -0,0 +1,215 @@
...
+model_pks = SkyModel(spectral1, name="model_pks")
+dataset_onoff.models = model_pks
+
+fit = Fit()
+res_pks = fit.run(dataset_onoff)
+print(res_pks.models)
+
+
+######################################################################
+# We see that the parameter `lambda_` is not well constrained. In this
```
QRemy:
```suggestion
# We see that the parameter ``lambda_`` (inverse of the cut-off energy) is not well constrained as the errors are very large. In this
```

examples/tutorials/analysis-1d/parameter_ul.py
QRemy: I guess it should be something that match the inverse of the reco energy range of the dataset

examples/tutorials/analysis-1d/parameter_ul.py
```diff
@@ -0,0 +1,215 @@
...
+
+values = profile["model_pks.spectral.lambda__scan"]
+loglike = profile["stat_scan"]
+
+ax = plt.gca()
+ax.plot(values, loglike - np.min(loglike))
+ax.set_xlabel("Cutoff value [1/TeV]")
+ax.set_ylabel(r"$\Delta$TS")
+secay = ax.secondary_yaxis("right", functions=(ts_to_sigma, sigma_to_ts))
+secay.set_ylabel("Significance [$\sigma$]")
```
QRemy:
```suggestion
```

QRemy: let's remove the second axis for simplicity

examples/tutorials/analysis-1d/parameter_ul.py
```diff
@@ -0,0 +1,215 @@
...
+# `profile` is a dictionary that stores the values of `lambda_` and the
+# corresponding likelihood values. Let's try to visualise it.
+#
+
+values = profile["model_pks.spectral.lambda__scan"]
+loglike = profile["stat_scan"]
+
+ax = plt.gca()
+ax.plot(values, loglike - np.min(loglike))
+ax.set_xlabel("Cutoff value [1/TeV]")
```
QRemy: maybe plot the Cutoff value to be consistent with the label 
```suggestion
ax.plot(1/values, loglike - np.min(loglike))
ax.set_xlabel("Cutoff value [TeV]")
```

examples/tutorials/analysis-1d/parameter_ul.py
```diff
@@ -0,0 +1,215 @@
...
+ax.set_ylabel(r"$\Delta$TS")
+secay = ax.secondary_yaxis("right", functions=(ts_to_sigma, sigma_to_ts))
+secay.set_ylabel("Significance [$\sigma$]")
+plt.title("Cutoff likelihood profile", fontsize=20, y=1.05)
+plt.show()
+
+
+######################################################################
+# Estimate the limits on the parameter
```
QRemy: This whole section puzzle me, why not using the `ParameterEstimator` to compute the `ul` on `lambda` ?

examples/tutorials/analysis-1d/parameter_ul.py
```diff
@@ -0,0 +1,215 @@
...
+
+unit = parameter.unit**-1  # Give in units of energy since lambda_ = 1/cutoff energy
+print(
+    f"{conf_level*100:.2f}% lower limit of energy cutoff: {1/UL:.4f} ({unit.to_string()})"
+)
+
+
+######################################################################
+# Since our dataset actually starts from 200 GeV, this analysis
+#  rule out a cut-off at 68% confidence in the energy range considered.
```
QRemy: do we rule-out a cut-off or we can't detect it significantly ?

examples/tutorials/analysis-1d/parameter_ul.py
registerrier: We have never really exposed the `ParameterEstimator` and have not really considered doing so. 

This is a general question that we can consider but it might be more pedagogical to explicitly compute the UL from the profile likelihood using `Fit.confidence`.

QRemy: Could also use the sampler and report the 95% of the posterior distribution. That would be interesting to compare to what we get from the confidence.

AtreyeeS:
Thanks for all your comments. Indeed, it was incorrect of me to compute the ul from the pdf
I have now updated the notebook, should be ready for review

examples/tutorials/details/parameter_ul.py
```diff
@@ -0,0 +1,243 @@
...
+print(errors1sig)
+print(errors2sig)
+
+
+######################################################################
+# Using the likelihood profile, we can set upper limits
+# on the parameter.
+
+ul_2sigma = stat_profile_ul_scipy(values, loglike, delta_ts=4)
+print(f"2-sigma upper limit on beta is {ul_2sigma:.2f}")
```
QRemy: you don't need that, you get the ul from the confidence fit with sigma =2 : `ul = errors2sig["errp"] + parameter.value`
I wouldn't expose `stat_profile_ul_scipy` because it can be very unstable if the profile is not well sampled which is more likely when you have only upper limits.

examples/tutorials/details/parameter_ul.py
```diff
@@ -0,0 +1,243 @@
...
+
+######################################################################
+# To compute higher sigma upper limits, you will need to increase the range
+# of the likelihood scan such that $\Delta$TS value is at least n_sigma**2.
```
QRemy: you can also remove that as by default we don't use the profile is not used to compute the ul

examples/tutorials/details/parameter_ul.py
```diff
@@ -0,0 +1,243 @@
...
+# not improve the fit. Thus, the function has internally updated
+# the best fit model to the one corresponding to the hull hypothesis
+
+print(dataset_onoff.models)
+
+######################################################################
+# Get flux points
+# ---------------
+# In can be useful to compute the flux points and visualise the difference
+# in the two spectral models
```
QRemy: This seems to discuss the influence of the choice of spectral model on flux points, it is not directly related to "upper limits on parameters". It might be more confusing than helpful to have this here especially as there is no guideline on which one should be used.

examples/tutorials/details/parameter_ul.py
```diff
@@ -0,0 +1,243 @@
...
+)
+print(LLR)
+
+
+######################################################################
+# We can see that the change in improvement in test statistic after adding the
+# curvature is only ~0.3, which corresponds to a significance of only 0.5.
+# Thus, we can safely conclude that the addition of the curvature parameter does
+# not improve the fit. Thus, the function has internally updated
+# the best fit model to the one corresponding to the hull hypothesis
```
QRemy:
```suggestion
# the best fit model to the one corresponding to the null hypothesis
```

examples/tutorials/details/parameter_ul.py
```diff
@@ -0,0 +1,243 @@
...
+ax.plot(values, loglike - np.min(loglike))
+ax.set_xlabel("Beta")
+ax.set_ylabel(r"$\Delta$TS")
+ax.set_title(r"$\beta$-parameter likelihood profile")
+plt.show()
+
+######################################################################
+# We can see that the likelihood profile is highly non-symmetric, and thus,
+# the error on the parameter quoted from the covariance matrix is not sufficient.
+# We can compute the asymmetric errors on the parameter
```
QRemy:
```suggestion
# We can compute the asymmetric errors and upper limits on the parameter
```

examples/tutorials/details/parameter_ul.py
```diff
@@ -0,0 +1,243 @@
...
+ax = plt.gca()
+ax.plot(values, loglike - np.min(loglike))
+ax.set_xlabel("Beta")
+ax.set_ylabel(r"$\Delta$TS")
+ax.set_title(r"$\beta$-parameter likelihood profile")
+plt.show()
+
+######################################################################
+# We can see that the likelihood profile is highly non-symmetric, and thus,
+# the error on the parameter quoted from the covariance matrix is not sufficient.
```
QRemy: remove

examples/tutorials/details/parameter_ul.py
```diff
@@ -0,0 +1,243 @@
...
+    sed_type="e2dnde",
+    facecolor="blue",
+    alpha=0.2,
+)
+plt.legend()
+plt.show()
+
+
+######################################################################
+# Compute parameter limits
```
QRemy:
```suggestion
# Compute parameter asymmetric erros and upper limits
```

examples/tutorials/details/parameter_ul.py
```diff
@@ -0,0 +1,244 @@
...
+# the allowed range of the non-significant parameter (eg: to rule
+# out parameter values, to compare from theoretical predications, etc).
+#
+# First, we reset the alternative model on the dataset
+
+dataset_onoff.models = LLR["fit_results"].models
+parameter = dataset_onoff.models.parameters["beta"]
+
+######################################################################
+# then we can compute the asymmetric errors and upper limits on the parameter
```
Astro-Kirsty: Should you mention then what the asymmetric errors are? And what we are seeing in the print statement

examples/tutorials/details/parameter_ul.py
```diff
@@ -0,0 +1,244 @@
...
+ul_2sigma = res_2sig["errp"] + parameter.value
+
+print(f"2-sigma lower limit on beta is {ul_2sigma:.2f}")
+print(f"2-sigma upper limit on beta is {ul_2sigma:.2f}")
+
+######################################################################
+# Likekihood profile
+# ------------------
+# We can also compute the likelihood profile on the parameter.
+# First we define the scan range such as it emcompasses the 1 and 2-sigma parameter limits.
```
Astro-Kirsty: I would write more, because it is still unclear why -1 and 10 are chosen when the 1 sigma is 0.68 and 1.09 and the 2 sigma is 0.95 and 3.71.
10 seems very high

examples/tutorials/details/parameter_ul.py
```diff
@@ -0,0 +1,244 @@
...
+
+res_1sig = fit.confidence(datasets=dataset_onoff, parameter=parameter, sigma=1)
+print(res_1sig)
+
+######################################################################
+# and the upper limits on the parameter.
+
+res_2sig = fit.confidence(datasets=dataset_onoff, parameter=parameter, sigma=2)
+ll_2sigma = res_2sig["errn"] + parameter.value
+ul_2sigma = res_2sig["errp"] + parameter.value
```
Astro-Kirsty: Should you do the same for the 1sigma?

examples/tutorials/details/parameter_ul.py
```diff
@@ -0,0 +1,244 @@
...
+# So quote preferably upper limits from the model which is the most supported by the data.
+
+energies = dataset_onoff.geoms["geom"].axes["energy"].edges
+
+# null hypothesis
+dataset_onoff.models = LLR["fit_results_null"].models
+fpe = FluxPointsEstimator(energy_edges=energies, n_jobs=4, selection_optional=["ul"])
+fp = fpe.run(dataset_onoff)
+ax = fp.plot(sed_type="e2dnde", color="blue")
+LLR["fit_results_null"].models[0].spectral_model.plot(
```
Astro-Kirsty: I would split out the running of the fpe and the plotting., same for the alt. hypothesis.

Astro-Kirsty (commented):
Thank you @AtreyeeS, this is looking much nicer. 

I added some comments and pushed a commit to fix language and formatting

AtreyeeS:
Thanks a lot @QRemy for the restructuring. Looks much better!
@Astro-Kirsty I have addressed your remaining comments. 

This should be good to merge now

QRemy (approved):
Thanks @AtreyeeS, looks good, there was already a lot of comments and improvements so let's merge it. Anything more could be done in a separate PR.

PR #5934, opened 2025-06-19
Estimator documentation: update the n_sigma description

bkhelifi:
**Description**
In the estimators docstring, the parameters `n_sigma`, `n_sigma_ul` and `n_sigma_sensitivity` were not properly described. They are all optional, float and positive.

This pull request fixes that.


AtreyeeS (approved):
Thanks @bkhelifi ! Indeed, no reason why we had `int`

Astro-Kirsty (changes requested):
Cheers @bkhelifi.

Instead of `should be positive` we could instead just say `Positive sigma value to use...`, and apply that logic to each.

ExcessMapEtimator --> need to add optional to `correlate_off`, `spectral_model`, `sum_over_energy_axes`. There are also other missing in TSMapEstimator,

gammapy/estimators/flux.py
```diff
@@ -27,12 +27,12 @@ class FluxEstimator(ParameterEstimator):
...
     source : str or int
         For which source in the model to compute the flux.
-    n_sigma : int, optional
-        Sigma to use for asymmetric error computation. Default is 1.
-    n_sigma_ul : int, optional
-        Sigma to use for upper limit computation. Default is 2.
-    n_sigma_sensitivity : int, optional
-        Sigma to use for sensitivity computation. Default is 5.
+    n_sigma : float, optional
+        Sigma to use for asymmetric error computation. Should be positive. Default is 1.
```
Tobychev: "Should" implies that negative values are ok, if only positive values are allowed using the word "Must" is more clear.

gammapy/estimators/parameter.py
```diff
@@ -413,7 +413,7 @@ class ParameterSensitivityEstimator:
        Parameter to test
     null_value : float or `~gammapy.modeling.Parameter`
         Value of the parameter for the null hypothesis.
-    n_sigma : int, optional
+    n_sigma : float, optional
```
Tobychev, line 419: No statement like "Must be positive." here, is that intentional?

bkhelifi:
> Maybe you can also add here in the "source" description that a best fit source model should be used. Models with negative amplitudes should be avoided.

@AtreyeeS, for which estimator?  Not the excessmap one, as no model is used...

gammapy/estimators/utils.py
```diff
@@ -1330,9 +1330,9 @@ def get_flux_map_from_profile(
     ----------
     flux_maps : `~gammapy.estimators.FluxMaps` or dict of `~gammapy.maps.WcsNDMap`
         Flux map or dict containing  a `stat_scan` entry
-    n_sigma : int
+    n_sigma : float, optional
```
Tobychev, line 1333: No statement like "Must be positive." here, is that intentional?

gammapy/estimators/flux.py
```diff
@@ -44,7 +47,7 @@ class FluxEstimator(ParameterEstimator):
         Default is None so the optional steps are not executed.
     fit : `Fit`, optional
```
Astro-Kirsty:
```suggestion
    fit : `~gammapy.modeling.Fit`, optional
```

gammapy/estimators/points/lightcurve.py
```diff
@@ -38,14 +38,17 @@ class LightCurveEstimator(FluxPointsEstimator):
     ----------
     time_intervals : list of `astropy.time.Time`
         Start and stop time for each interval to compute the LC.
-    source : str or int
-        For which source in the model to compute the flux points. Default is 0.
+    source : str or int, optional
+        For which source in the model to compute the flux points.
+        Default is 0, ie the first source of the models.
```
Astro-Kirsty:
```suggestion
        Default is 0, i.e. the first source of the models.
```

Astro-Kirsty (changes requested):
Thanks @bkhelifi, some comments. 

Also can you please edit the order of the parameter list such that it matches the order of how they are called?

gammapy/estimators/parameter.py
bkhelifi: Well seen, thanks

Astro-Kirsty (approved):

PR #5937, opened 2025-06-20
Remove binlike from testing

Astro-Kirsty:
This is part of the cleanup over on https://github.com/gammapy/gammapy-data/issues/41

### DIFF
--- a/gammapy/estimators/points/core.py
+++ b/gammapy/estimators/points/core.py
@@ -813,13 +813,13 @@ class FluxPoints(FluxMaps):
         Examples
         --------
         >>> from gammapy.estimators import FluxPoints
-        >>> filename = '$GAMMAPY_DATA/tests/spectrum/flux_points/binlike.fits'
+        >>> filename = "$GAMMAPY_DATA/estimators/crab_hess_fp/crab_hess_fp.fits"
         >>> flux_points = FluxPoints.read(filename)
         >>> flux_points_recomputed = flux_points.recompute_ul(n_sigma_ul=3)
-        >>> print(flux_points.meta["n_sigma_ul"], flux_points.flux_ul.data[0])
-        2.0 [[3.95451985e-09]]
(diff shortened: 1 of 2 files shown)


adonath:
> The stat_scan values in tests/spectrum/flux_points/binlike.fits are not compatible with the results. Remove?

While this is true, it still might be the format that is used by fermipy. Not sure we have a separate test for this, would be good.

gammapy/estimators/points/core.py
```diff
@@ -813,13 +813,13 @@ def recompute_ul(self, n_sigma_ul=2, **kwargs):
...
         >>> flux_points = FluxPoints.read(filename)
         >>> flux_points_recomputed = flux_points.recompute_ul(n_sigma_ul=3)
-        >>> print(flux_points.meta["n_sigma_ul"], flux_points.flux_ul.data[0])
-        2.0 [[3.95451985e-09]]
-        >>> print(flux_points_recomputed.meta["n_sigma_ul"], flux_points_recomputed.flux_ul.data[0])
-        3 [[6.22245374e-09]]
+        >>> print(flux_points.meta["n_sigma_ul"], flux_points.flux_ul.data[1])
+        3.0 [[3.99250033e-11]]
+        >>> print(flux_points_recomputed.meta["n_sigma_ul"], flux_points_recomputed.flux_ul.data[1])
+        3 [[3.99323829e-11]]
```
AtreyeeS: I think the `recomputed` can have a different value from the original, otherwise the example is not clear. Maybe with `n_sigma_ul=2` or 4?

Astro-Kirsty:
Good idea, I've created an issue https://github.com/gammapy/gammapy-data/issues/74 for that. For this PR we can keep it as is

AtreyeeS (approved):
Thanks @Astro-Kirsty!
Looks good to me.

bkhelifi (approved):
Seems ok for me, thanks

PR #5938, opened 2025-06-20
ci: allow to skip ci step using PR labels

### LINKED ISSUE #5551: Use labels to skip tests - docs build - CI
I have created three new labels (see on this issue). We should use these labels to deactivate CI jobs like tests, docs build or the whole CI. 
This is especially useful when a PR for sure does not change anything in the gammapy code. 
For instance, no need to run tests when you only change documentation, no need to run the CI if you are changing the README.rst, etc...
This is also saving some computation cost: time, money and carbon footprint !

natthan-pigoux:
**Description**

Resolves #5551 , I added a condition to skip CI docs and tests steps using PR labels.

**Dear reviewer**

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,7 +5,6 @@ jobs:
   lint:
     name: Run pre-commit hooks
     runs-on: ubuntu-latest
-    if: contains(github.event.pull_request.labels.*.name, 'skip-lint') == false
 
     steps:
       - name: Check out repository


natthan-pigoux:
I would personally rename the labels `skip docs` -> `skip-docs` and `skip tests` -> `skip-tests` to keep same logic with other labels. But doesn't change anything.

Tobychev:
I remember that this was incredibly fiddly to get working properly for [ctapipe](https://github.com/cta-observatory/ctapipe/pull/2227/files), in the end the blessed string that worked for us there was this:

```
contains(github.event.pull_request.labels.*.name, 'lable-name') == false
```

I suggest you use that, the current PR seems to fail with a syntax error.

natthan-pigoux:
> I remember that this was incredibly fiddly to get working properly for [ctapipe](https://github.com/cta-observatory/ctapipe/pull/2227/files), in the end the blessed string that worked for us there was this:
> 
> ```
> contains(github.event.pull_request.labels.*.name, 'lable-name') == false
> ```
> 
> I suggest you use that, the current PR seems to fail with a syntax error.

thanks a lot! Yes I was trying a few syntaxes, and indeed there were not all working as expected or not working at all.

Tobychev:
A meta comment is it would be good if there was a way to skip just the data-tests, since they are by far the heaviest and not all changes require data files to be verified via unit tests.

Tobychev:
I think it's better to just use multiple tags on the issue if you want to skip everything: the syntax to check one single tag is bad enough, adding an OR clause will make it even worse and honestly I'm not sure how easy it is to get it working, it is _very_ fiddly.

natthan-pigoux:
> Thanks @natthan-pigoux. I agree with the naming convention. Can we also add one that skips the entire CI?

I would also agree with the above comment that we can have eventually one label for each CI jobs to skip it.

natthan-pigoux:
> A meta comment is it would be good if there was a way to skip just the data-tests, since they are by far the heaviest and not all changes require data files to be verified via unit tests.

Do you mean the tests where data sets are downloaded? e.g. `gammapy_data: true`

natthan-pigoux:
@Astro-Kirsty , I added a condition to also skip the linting step. So by combining the 3 skip labels in the PR no CI jobs should run.
Moreover, do you have the rights to change the labels name?

Tobychev:
> > A meta comment is it would be good if there was a way to skip just the data-tests, since they are by far the heaviest and not all changes require data files to be verified via unit tests.
> 
> Do you mean the tests where data sets are downloaded? e.g. `gammapy_data: true`

Yes, those look to run up to 24 minutes, while the ones without data take about 5. However, looking at the CI file it seems like it could be messy to implement, so best leave it for a separate PR.

.github/workflows/ci.yml
```diff
@@ -5,6 +5,8 @@ jobs:
   lint:
     name: Run pre-commit hooks
     runs-on: ubuntu-latest
+    if: contains(github.event.pull_request.labels.*.name, 'skip-lint') == false
```
registerrier: I would not skip linting. This can be problematic if code or docs are merged without proper formatting.

registerrier (approved):
Thanks @natthan-pigoux . Looks good to me.

PR #5940, opened 2025-06-20
Update sherpa message and add gammapy info to docs

Astro-Kirsty:
This PR is to remove one of the sherpa messages which was incorrect. 
It also adds 'gammapy info' instructions to the setup after the dev call 20/06/25


AtreyeeS (approved):

PR #5941, opened 2025-06-20
Tutorial setup command line option

Astro-Kirsty:
There was a discussion in the dev call (20/06/25) regarding https://github.com/gammapy/gammapy/issues/5816

We might not need to include the `check setup` in every tutorial, but rather somewhere else in the docs. I changed the intro of the tutorials to include this new cli

If we are happy with something like this we could remove `check setup` from each tutorial


Astro-Kirsty:
After the dev call (27/06/2025) it was decided that this is a good way forward, and that we should remove the `check setup` from most tutorials. We can leave it in the first few, so that users see how it works

PR #2344, opened 2019-09-06
Change return type of `WcsGeom.get_coord()` to quantities

adonath:
This PR addresses #1792. It includes the following changes:

- Change `MapAxis.pix_to_coord()` to return a `Quantity`
- Change `WcsGeom.pix_to_coord()` to return `Quantities`
- Allow `lon` and `lat` to be a `Quantity`in `MapCoord.__init__`
- Adapt tests and notebooks

Shall we declare `lon` and `lat` as `Longitude` and `Latitude` objects instead of `Quantity`?

Currently the adaptation of `HpxGeom.coord_to_pix()` is missing, but if I remember correctly there were problems with the adaptation, when I already tried it ~2 years ago, so I'd prefer to do this in a second PR.

Issue #1792, opened 2018-09-13
Support of units in MapCoord


lmohrmann:
I have noticed that the `MapCoord` class apparently does not support units and have wondered whether there is a particular reason for this? The docs at least don't mention a reason, as far as I can see.

adonath:
The basic answer here is, that we would like to have it, but it's not supported completely and uniformly yet. I started to work on this in #1685 and #1757. The current status is, that you can pass a `Quantity` to the `MapCoord` object for the non-spatial axes and it will be handled correctly in the `.get_by_coord()`, `set_by_coord()` and `.interp_by_coord()` methods. But for example `Geom.get_coord()` does not return a `Quantity` yet. 

To fully support quantities in the maps API (for spatial axes and all methods with `coord` in the name, as well as I/O) we still have to put in some work that we also have to coordinate with FermiPy (which does not use quantities but uses `gammapy.maps` as well).  I have not estimated the development effort for this yet, but my guess is, that it should be a milestone for Gammapy 1.0.

cdeil:
I also recently noticed this issue that coords doesn't offer quantity when reviewing #2313 which introduced code like [here](https://github.com/luca-giunti/gammapy/blob/0b1f295d82896fe3bfffef16b49e468929a65795/gammapy/image/models/tests/test_core.py#L83-L85) to get coords with unit that work for model evaluation:
```
    coords = geom.get_coord()
    lon = coords.lon * u.deg
    lat = coords.lat * u.deg
```

PyCharm shows a warning that this is copy & paste code that we have multiple times in the Gammapy project. It's also error-prone, to have the "deg" unit hard-coded, and probably inefficient, using "rad" in model eval should be faster (whereas WCS uses deg, but that's only used in the setup call to create grids, not in the likelihood loop).

I think reviewing the maps API and changing it to use Quantity consistently is a very large task, but changing MapCoord a bit to offer quantity coords should be easy (finding "bad coord code" like that is still hard, but could be deferred to a general Gammapy code review later).

adonath:
The remaining action item here is to do the same change as in #2344 for `HpxGeom`.

adonath:
Could be a address as part of `gammapy.maps` refactoring for v2.0

lmohrmann:
@AtreyeeS  "not planned" as in "never planned"...?! What is the rationale?

PR #5944, opened 2025-06-23
Add unit support on HpxGeom

registerrier:
**Description**

This pull request is the (late) followup of #2344 and should solve #1792 

Note that there remains a number of aspects which are not unit compliants. For instance commands like `set_by_xxx` or `fill_by_xxx` don't check the unit consistency nor do they apply proper conversion. This could be done in a future PR. 

**Dear reviewer**


registerrier:
Most of the changes consist in removing the units in the test, because many tests rely on using the coord to set the content of the map.

registerrier:
Merging now.

PR #5918, opened 2025-06-13
Adding VERITAS tutorial

samanthalwong:
**Description**

Adding VERITAS tutorial to go with this PR to add open VERITAS Crab dataset https://github.com/gammapy/gammapy-data/pull/71

I appreciate any comments on the content and scope of the notebook! We were aiming for something fairly detailed, that mimics a full VERITAS analysis using our proprietary packages but are open to re-scoping for the gammapy audience.

This should be tested with the Crab dataset mentioned above.

PR #5945, opened 2025-06-23
Create exclusion.py

samanthalwong:
Exclusion region function for excluding bright stars and VHE sources to be used in #5918. By default, this includes a file of star values to be added in https://github.com/gammapy/gammapy-data/pull/71. 

I appreciate any comments or suggestions on the implementation - it's possible this could be expanded to a broader class with more flexibility, but for now just mirrors what was in the original notebook in #5918.

### DIFF
--- a/gammapy/utils/exclusion.py
+++ b/gammapy/utils/exclusion.py
@@ -1,0 +1,64 @@
+import astropy.units as u
+import numpy as np
+from astropy.table import Table
+from regions import *
+from astropy.coordinates import SkyCoord
+import os
+
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+"""Exclusion regions for source position, stars, and other gamma-ray sources"""
+
+def make_exclusion_mask(source_position,geom,rad=3*u.deg,max_star_mag=6,other_exc=[]):
+    """
+    Create an exclusion mask for the source position, any stars in the field of view,
+    and any user-provided regions.
+
+    Parameters
+    ----------
+    source_position : `~regions.SkyRegion`
+        Source exclusion region coordinates (can be any geometry).
+    geom: `~gammapy.maps.WcsGeom`
+        Map geometry for the exclusion mask. 
+    rad : `~astropy.units.Quantity`
+        Radius around `source_position` to search for stellar exclusion regions.
+    max_star_mag : 
+        The maximum stellar magnitude (Johnson-Cousins B-band) to exclude. Maximum value is 8th magnitude.  
+    other_exc : list of `~regions.SkyRegion`
+        Other regions to exclude (can be any geometry). 
+
+    Returns
+    -------
+    exclusion_mask : `~gammapy.maps.WcsNDMap`
+        2D exclusion mask 
+    """
+    exc = [source_position]
+
+    for src in other_exc:
+        exc.append(src)
+    star_data = np.loadtxt(os.environ.get("GAMMAPY_DATA")+"/veritas/crab-point-like-ED/Hipparcos_MAG8_1997.dat",usecols=(0, 1, 2, 3))
+    star_cat = Table(
+    {
+        "ra": star_data[:, 0],
+        "dec": star_data[:, 1],
+        "id": star_data[:, 2],
+        "mag": star_data[:, 3],
+    }
+    )
+    star_mask = (
+    np.sqrt(
+        (star_cat["ra"] - source_position.center.ra.deg) ** 2
+        + (star_cat["dec"] - source_position.center.dec.deg) ** 2
+    ) < rad.value
+    )
+
+    for src in star_cat[(star_mask) & (star_cat["mag"] < max_star_mag)]:
+        exc.append(
+            CircleSkyRegion(
+                center=SkyCoord(src["ra"], src["dec"], unit="deg", frame="icrs"),
+                radius=0.3 * u.deg,
+            )
+        )
+
+    exclusion_mask = ~geom.region_mask(exc)
+
+    return exclusion_mask


gammapy/utils/exclusion.py
```diff
@@ -0,0 +1,64 @@
+import astropy.units as u
+import numpy as np
+from astropy.table import Table
+from regions import *
+from astropy.coordinates import SkyCoord
+import os
+
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
```
bkhelifi: Could you move this line at the very top of the file?

gammapy/utils/exclusion.py
```diff
@@ -0,0 +1,64 @@
...
+        + (star_cat["dec"] - source_position.center.dec.deg) ** 2
+    ) < rad.value
+    )
+
+    for src in star_cat[(star_mask) & (star_cat["mag"] < max_star_mag)]:
+        exc.append(
+            CircleSkyRegion(
+                center=SkyCoord(src["ra"], src["dec"], unit="deg", frame="icrs"),
+                radius=0.3 * u.deg,
+            )
```
bkhelifi: This angle is not instrument/analysis dependant? I think it should be configurable...
And maybe energy-dependant? For an other PR maybe... add a ToDO as reminder

gammapy/utils/exclusion.py
```diff
@@ -0,0 +1,64 @@
...
+    Returns
+    -------
+    exclusion_mask : `~gammapy.maps.WcsNDMap`
+        2D exclusion mask 
+    """
+    exc = [source_position]
+
+    for src in other_exc:
+        exc.append(src)
+    star_data = np.loadtxt(os.environ.get("GAMMAPY_DATA")+"/veritas/crab-point-like-ED/Hipparcos_MAG8_1997.dat",usecols=(0, 1, 2, 3))
```
bkhelifi: A naive question... this catalogue is not in the VO?

bkhelifi:
This is good to start to create such functions, thanks. I have some comments:
- this function is about star_exclusion_region, it does not aimed to create some from any catalog. I just propose to rename the function.
- is there no way to get the catalog from the VO, using pyvo?

gammapy/utils/exclusion.py
samanthalwong: Yes - this could be replaced with a line such as:
`vo.dal.conesearch('https://gaia.ari.uni-heidelberg.de/gaiasky/files/catalogs/hip/hipparcos.vot',pos=crab)`, but I thought we didn't want to query external catalogs so that connection errors won't cause tests to fail?

gammapy/utils/exclusion.py
samanthalwong: I can make it configurable - wasn't sure if this would make the inputs more confusing than necessary, but I agree that this will be strongly dependent on the PSF of each instrument/observation

samanthalwong:
Thanks for the quick review, I appreciate it!

> This is good to start to create such functions, thanks. I have some comments:
> 
> * this function is about star_exclusion_region, it does not aimed to create some from any catalog. I just propose to rename the function.

> It excludes gamma-ray sources as well (the observation target is required) and optional custom regions. This could potentially query gamma-cat, etc. to be more automated and query catalogs for both, but isn't just for stars as-is

. 
> * is there no way to get the catalog from the VO, using pyvo?

> See above.

gammapy/utils/exclusion.py
```diff
@@ -0,0 +1,72 @@
...
+            "id": star_data[:, 2],
+            "mag": star_data[:, 3],
+        }
+    )
+    star_mask = (
+        np.sqrt(
+            (star_cat["ra"] - source_position.center.ra.deg) ** 2
+            + (star_cat["dec"] - source_position.center.dec.deg) ** 2
+        )
+        < rad.value
```
registerrier: There you should use `SkyCoord.separation`

gammapy/utils/exclusion.py
```diff
@@ -0,0 +1,72 @@
...
+    """
+    exc = [source_position]
+
+    for src in other_exc:
+        exc.append(src)
+    star_data = np.loadtxt(
+        os.environ.get("GAMMAPY_DATA")
+        + "/veritas/crab-point-like-ED/Hipparcos_MAG8_1997.dat",
+        usecols=(0, 1, 2, 3),
+    )
```
registerrier: How about using `Table.read` directly?

gammapy/utils/exclusion.py
registerrier: Or probably `geom.contains`. Then you don't need the `rad` argument.

gammapy/utils/exclusion.py
```diff
@@ -0,0 +1,72 @@
...
+def make_exclusion_mask(
+    source_position, geom, rad=3 * u.deg, max_star_mag=6, other_exc=[]
+):
+    """
+    Create an exclusion mask for the source position, any stars in the field of view,
+    and any user-provided regions.
+
+    Parameters
+    ----------
+    source_position : `~regions.SkyRegion`
```
registerrier: The variable name does not suggest this is a `region` object.

registerrier (commented):
Thanks @samanthalwong . I don't know what is the best strategy here. Such functions can become very complex in terms on configuration as there are numerous ways to build the list of exclusion regions (e.g. one could use known TeV sources from `gammapy.catalog` as well). The table of stars is likely very specific here. One solution is to have it as part of the Veritas tutorial as function.

Otherwise, if we want to add such utility functions, I think its role could be simply to build the regions list based on a `Table` and a `Geom`.  Creating the mask is then straightforward. 

I have left a number of comments inline.

samanthalwong:
> Thanks @samanthalwong . I don't know what is the best strategy here. Such functions can become very complex in terms on configuration as there are numerous ways to build the list of exclusion regions (e.g. one could use known TeV sources from `gammapy.catalog` as well). The table of stars is likely very specific here. One solution is to have it as part of the Veritas tutorial as function.
> 
> Otherwise, if we want to add such utility functions, I think its role could be simply to build the regions list based on a `Table` and a `Geom`. Creating the mask is then straightforward.
> 
> I have left a number of comments inline.

Thanks @registerrier for the review! The star table here covers the whole sky up to 8th magnitude - if we don't want to query an online database (such as VO, SIMBAD, etc.), I think it should be broadly fine for all instruments, but I'm open to any suggestions! I will make the changes you suggested to use a Table approach instead, and in future releases I think it should be very reasonable to implement sources from `gammapy.catalog` but can leave in the custom exclusion regions for now, if that works.

AtreyeeS:
I have a very naive question here. Can bright stars be masked at the DL4 level like this? I would imagine this needs to be done at a much lower level, ie that the pixels illuminated by bright stars have to be switched off during data taking or removed during the data cleaning process?
Do you actually see these pixels brighter at the DL4 level as well?

samanthalwong:
I believe the mask is needed for the opposite reason, i.e., that due to pixels being switched off due to bright stars, there will be dips in acceptance at the star locations in the FoV. This cause "cold regions" or holes to appear in our sky maps.

AtreyeeS:
> I believe the mask is needed for the opposite reason, i.e., that due to pixels being switched off due to bright stars, there will be dips in acceptance at the star locations in the FoV. This cause "cold regions" or holes to appear in our sky maps.

Thanks @samanthalwong ! But even if the pixels are switched off, the final arrival direction of the primary particle may still be reconstructed to that pointing position, no?
Also, if what you mention is happening, then one would need to mask any "swicthed off" pixel, not necessarily coming from stars, but dues to any low level issues. It can be complex to correctly transmit this info to the DL3 level

samanthalwong:
> > I believe the mask is needed for the opposite reason, i.e., that due to pixels being switched off due to bright stars, there will be dips in acceptance at the star locations in the FoV. This cause "cold regions" or holes to appear in our sky maps.
> 
> Thanks @samanthalwong ! But even if the pixels are switched off, the final arrival direction of the primary particle may still be reconstructed to that pointing position, no? Also, if what you mention is happening, then one would need to mask any "swicthed off" pixel, not necessarily coming from stars, but dues to any low level issues. It can be complex to correctly transmit this info to the DL3 level

I've looked into this a little more and it seems like it's actually quite a complicated effect. It seems that the best way to mitigate it is by masking, since cleaning cannot account for all the effects. I am not sure if other instruments' cameras are able to handle this better/differently, but I've been advised that masking stars is absolutely necessary for VERITAS analysis.

Here are some notes from previous VERITAS work: 
- The result of a pixel being suppressed is that the ellipse parameter calculations for that particular image treats that off pixel as having zero charge. This alteration to the image has other effects that propagate through the software and the analysis.
- When the pixels near stars were shut down, there were fewer gamma-like events per unit time (relative event rate) around the star’s location, for a radius of around 0.2 deg in the skymap
- These holes in the significance map do not have a large effect on the final significance of the gamma-ray point source being studied, as this is calculated via the ring or reflected region methods.

gammapy/utils/regions.py
```diff
@@ -422,3 +424,46 @@ def region_circle_to_ellipse(region):
...
+
+def extract_bright_star_regions(geom, max_star_mag=6, star_rad=0.3 * u.deg):
+    """
+    Find bright star exclusion regions contained in a given `~gammapy.maps.Geom` and returns a 2D exclusion mask to remove these sources from bacgkround estimation.
+
+    Parameters
+    ----------
+    geom: `~gammapy.maps.WcsGeom`
+        Map geometry for the exclusion mask.
+    max_star_mag :
```
bkhelifi:
```suggestion
    max_star_mag : float, optional
```

gammapy/utils/regions.py
```diff
@@ -422,3 +424,46 @@ def region_circle_to_ellipse(region):
...
+        The maximum stellar magnitude (Johnson-Cousins B-band) to exclude. Maximum value is 8th magnitude.
```
bkhelifi:
```suggestion
        The maximum stellar magnitude (Johnson-Cousins B-band) to exclude. Maximum value is 8th magnitude.
        Default is 6.
```

gammapy/utils/regions.py
```diff
@@ -422,3 +424,46 @@ def region_circle_to_ellipse(region):
...
     return region_new
+
+
+def extract_bright_star_regions(geom, max_star_mag=6, star_rad=0.3 * u.deg):
+    """
+    Find bright star exclusion regions contained in a given `~gammapy.maps.Geom` and returns a 2D exclusion mask to remove these sources from bacgkround estimation.
+
+    Parameters
+    ----------
+    geom: `~gammapy.maps.WcsGeom`
```
bkhelifi:
```suggestion
    geom : `~gammapy.maps.WcsGeom`
```

gammapy/utils/regions.py
```diff
@@ -422,3 +424,46 @@ def region_circle_to_ellipse(region):
...
+        Map geometry for the exclusion mask.
+    max_star_mag :
+        The maximum stellar magnitude (Johnson-Cousins B-band) to exclude. Maximum value is 8th magnitude.
+    star_rad : `~astropy.units.Quantity`
```
bkhelifi:
```suggestion
    star_rad : `~astropy.units.Quantity`, optional
```

gammapy/utils/regions.py
```diff
@@ -422,3 +424,46 @@ def region_circle_to_ellipse(region):
...
+        Radius to exclude for each star.
```
bkhelifi:
```suggestion
        Radius to exclude around each star. Default is 0.3 degrees.
```

gammapy/utils/regions.py
```diff
@@ -422,3 +424,46 @@ def region_circle_to_ellipse(region):
         center=region.center, width=region.radius, height=region.radius
     )
     return region_new
+
+
+def extract_bright_star_regions(geom, max_star_mag=6, star_rad=0.3 * u.deg):
+    """
+    Find bright star exclusion regions contained in a given `~gammapy.maps.Geom` and returns a 2D exclusion mask to remove these sources from bacgkround estimation.
+
```
bkhelifi, line 435: You can precise in the docstring which star catalogue is used, and adds a reference towards it.

bkhelifi:
Thank you @samanthalwong . It looks good for me!
Now, one needs to create tests for this new function:
- no region found associated to a large magnitude (l460)
- a geom with no star (l455)
- a standard field which know starts and you test the number of regions

gammapy/utils/regions.py
```diff
@@ -422,3 +424,48 @@ def region_circle_to_ellipse(region):
         center=region.center, width=region.radius, height=region.radius
     )
     return region_new
+
+
+def extract_bright_star_regions(geom, max_star_mag=6, star_rad=0.3 * u.deg):
+    """
+    Find bright star exclusion regions from the Hipparcos Catalog (Perryman et al., 1997)
+    contained in a given `~gammapy.maps.Geom` and returns a 2D exclusion mask to remove these sources from bacgkround estimation.
```
AtreyeeS: You can also  returning a list of regions, so edit the docstring accordingly

gammapy/utils/regions.py
```diff
@@ -422,3 +424,48 @@ def region_circle_to_ellipse(region):
...
+
+    Parameters
+    ----------
+    geom : `~gammapy.maps.WcsGeom`
+        Map geometry for the exclusion mask.
+    max_star_mag : float, optional
+        The maximum stellar magnitude (Johnson-Cousins B-band) to exclude. Maximum value is 8th magnitude.
```
AtreyeeS: Why is 8 the maximum ?

gammapy/utils/regions.py
```diff
@@ -422,3 +424,48 @@ def region_circle_to_ellipse(region):
         center=region.center, width=region.radius, height=region.radius
     )
     return region_new
+
+
+def extract_bright_star_regions(geom, max_star_mag=6, star_rad=0.3 * u.deg):
```
AtreyeeS: You can have a `table=None` so that this function can be used by some other catalog having the same format as your Hipparcos Catalog. If `table` is None, (the default option) it accesses the files in the Veritas folder in Gammapy-data

AtreyeeS (changes requested):
Thanks @samanthalwong !
Please dont delete the `docs.sphinxest.py` from the gammapy repo! :) 

This PR needs unit tests

samanthalwong:
> Thanks @samanthalwong !
> 
> Please dont delete the `docs.sphinxest.py` from the gammapy repo! :) 
> 
> 
> 
> This PR needs unit tests

Ah sorry - I thought I was just removing it from this PR. I can put it back. Do the unit tests need to be added before the release? I won't have internet access until Saturday so won't be able to add them this week.

AtreyeeS:
@samanthalwong We will need the unit tests before merging this PR. Next week should be OK

gammapy/utils/regions.py
samanthalwong: From empirical tests, it seems that only stars < 6th mag have an effect on the VERITAS analysis. I believe that this extends down to 8 just in case analyzers want to be extra cautious or are analyzing fields with very bright star fields, etc.

AtreyeeS: There there is no hard limit from the script? I would suggest you remove the line `Maximum value is 8th magnitude.` and just put the default value

samanthalwong: The limit of the table in the file that's included is 8th magnitude, which is why I wrote that line in the docstring. If it makes sense to remove it, then users can probably infer the lowest magnitude from the file itself if they really want to go lower?

gammapy/utils/regions.py
samanthalwong: Since the file has fairly unique column names (e.g., `_RAJ2000`), would it make sense to rename the columns in the file or just warn the user that the file needs to have specific column names?

AtreyeeS: I think you can mention this in the docstring

gammapy/utils/regions.py
```diff
@@ -422,3 +424,56 @@ def region_circle_to_ellipse(region):
...
+    ----------
+    geom : `~gammapy.maps.WcsGeom`
+        Map geometry for the exclusion mask.
+    max_star_mag : float, optional
+        The maximum stellar magnitude (Johnson-Cousins B-band) to exclude. Default is 6th magnitude.
+    star_rad : `~astropy.units.Quantity`, optional
+        Radius to exclude around each star. Default is 0.3 degrees.
+    star_table : `~astropy.table.Table`, optional
+        Table containing the list of stars to search for exclusion regions. If ``None``, the Hipparcos Catalog (Perryman et al., 1997) contained in ``$GAMMAPY_DATA//veritas/crab-point-like-ED``. The table's column names should match those in ``$GAMMAPY_DATA//veritas/crab-point-like-ED``.
+
```
bkhelifi:
```suggestion
        Table containing the list of stars to search for exclusion regions. If None, the Hipparcos Catalog (Perryman et al., 1997) contained in ``$GAMMAPY_DATA//veritas/crab-point-like-ED``. The table's column names should match those in ``$GAMMAPY_DATA//veritas/crab-point-like-ED``.
        Default is None.

```

bkhelifi (approved):
Perfect. Good for me. Thanks @samanthalwong

docs/sphinxext.py
```diff
@@ -18,6 +18,7 @@
         "cta.py": 1,
         "fermi_lat.py": 2,
         "hawc.py": 3,
+        "veritas.py": 4,
```
AtreyeeS: This change is not part of this PR.

gammapy/utils/regions.py
```diff
@@ -422,3 +424,57 @@ def region_circle_to_ellipse(region):
...
+    Parameters
+    ----------
+    geom : `~gammapy.maps.WcsGeom`
+        Map geometry for the exclusion mask.
+    max_star_mag : float, optional
+        The maximum stellar magnitude (Johnson-Cousins B-band) to exclude. Default is 6th magnitude.
+    star_rad : `~astropy.units.Quantity`, optional
+        Radius to exclude around each star. Default is 0.3 degrees.
+    star_table : `~astropy.table.Table`, optional
+        Table containing the list of stars to search for exclusion regions. If None, the Hipparcos Catalog (Perryman et al., 1997) contained in ``$GAMMAPY_DATA//veritas/crab-point-like-ED``. The table's column names should match those in ``$GAMMAPY_DATA//veritas/crab-point-like-ED``.
```
AtreyeeS:
```suggestion
        Table containing the list of stars to search for exclusion regions. If None, the Hipparcos Catalog (Perryman et al., 1997) contained in ``$GAMMAPY_DATA//veritas/crab-point-like-ED`` is used. The table's column names should match those in ``$GAMMAPY_DATA//veritas/crab-point-like-ED``.
```

gammapy/utils/regions.py
```diff
@@ -422,3 +424,57 @@ def region_circle_to_ellipse(region):
...
+    geom, max_star_mag=6, star_rad=0.3 * u.deg, star_table=None
+):
+    """
+    Find bright star exclusion regions from a catalog `~astropy.table.Table`
+    contained in a given `~gammapy.maps.Geom` and returns a list of `~regions.CircleSkyRegion`.
+
+    Parameters
+    ----------
+    geom : `~gammapy.maps.WcsGeom`
+        Map geometry for the exclusion mask.
```
AtreyeeS:
```suggestion
        WcsGeom within which to search for bright stars
```

gammapy/utils/regions.py
```diff
@@ -422,3 +424,57 @@ def region_circle_to_ellipse(region):
...
+    max_star_mag : float, optional
+        The maximum stellar magnitude (Johnson-Cousins B-band) to exclude. Default is 6th magnitude.
+    star_rad : `~astropy.units.Quantity`, optional
+        Radius to exclude around each star. Default is 0.3 degrees.
+    star_table : `~astropy.table.Table`, optional
```
AtreyeeS:
```suggestion
    star_table : path to file containing a Table
```

AtreyeeS (changes requested):
Thanks @samanthalwong !
Left some small comments inline. 

Once you add the tests, this should be good to go

samanthalwong:
@AtreyeeS @bkhelifi I've added the unit tests requested, using the Mrk 501 field as a star-less region and Crab field as a known region (2 stars should be found within 4 deg). Let me know if any adjustments need to be made to these tests or if I misunderstood what you were looking for!

AtreyeeS (approved):
Thank you @samanthalwong !
No further comments on this.

We have to wait for https://github.com/gammapy/gammapy-data/pull/71 before merging

gammapy/utils/tests/test_regions.py
```diff
@@ -184,3 +187,40 @@ def test_polygon_points_pixel_region_rotate():
     expected_vertices = PixCoord([0, 1], [0, 1])
     assert_allclose(rotated_region.vertices.x, expected_vertices.x, atol=1e-7)
     assert_allclose(rotated_region.vertices.y, expected_vertices.y, atol=1e-7)
+
+
+def test_star_exclusion_no_region_bright():
```
registerrier:
```suggestion

@requires_data()
def test_star_exclusion_no_region_bright():
```

registerrier: This tests needs GAMMAPY_DATA available to be run.

gammapy/utils/regions.py
```diff
@@ -422,3 +424,57 @@ def region_circle_to_ellipse(region):
...
+        Star exclusion regions.
+    """
+    regions = []
+
+    if star_table is None:
+        star_cat = Table.read(
+            os.environ.get("GAMMAPY_DATA")
+            + "/veritas/crab-point-like-ED/Hipparcos_MAG8_1997.dat",
+            format="ascii.commented_header",
+        )
```
registerrier:
```suggestion
        filename = make_path("$GAMMAPY_DATA/veritas/crab-point-like-ED/Hipparcos_MAG8_1997.dat")
        star_cat = Table.read(filename, format="ascii.commented_header")
```

gammapy/utils/regions.py
```diff
@@ -14,10 +14,12 @@
 """
 
 import operator
+import os
 import numpy as np
 from scipy.optimize import Bounds, minimize
 from astropy import units as u
 from astropy.coordinates import SkyCoord
+from astropy.table import Table
```
registerrier, line 21:
```suggestion
from astropy.table import Table
from gammapy.utils.scripts import make_path
```
this function can deal with paths and env variables.

gammapy/utils/tests/test_regions.py
```diff
@@ -184,3 +187,40 @@ def test_polygon_points_pixel_region_rotate():
...
+        binsz=0.02,
+        width=(4, 4),
+        frame="icrs",
+        proj="CAR",
+    )
+    regions = extract_bright_star_regions(geom, max_star_mag=max_star_mag)
+    assert len(regions) == 0
+
+
+def test_star_exclusion_no_region_no_stars():
```
registerrier, line 208:
```suggestion
@requires_data()
def test_star_exclusion_no_region_no_stars():
```

gammapy/utils/tests/test_regions.py
```diff
@@ -184,3 +187,40 @@ def test_polygon_points_pixel_region_rotate():
...
+        binsz=0.02,
+        width=(4, 4),
+        frame="icrs",
+        proj="CAR",
+    )
+    regions = extract_bright_star_regions(geom)
+    assert len(regions) == 0
+
+
+def test_star_exclusion_known_field():
```
registerrier, line 221:
```suggestion
@requires_data()
def test_star_exclusion_known_field():
```

gammapy/utils/tests/test_regions.py
```diff
@@ -31,8 +31,11 @@
     get_centroid,
     PolygonPointsSkyRegion,
     PolygonPointsPixelRegion,
+    extract_bright_star_regions,
 )
 
+from gammapy.maps import WcsGeom
```
registerrier, line 37:
```suggestion
from gammapy.maps import WcsGeom
from gammapy.utils.testing import requires_data
```

registerrier (commented):
Thanks @samanthalwong . Looks good.  Nearly there!

To run the tests one needs to check if the GAMMAPY_DATA are available. A decorator `requires_data` on the test function can check this.
Also I propose to use the helper function `make_path` which allows easier handling of paths.

gammapy/utils/regions.py
```diff
@@ -422,3 +425,54 @@ def region_circle_to_ellipse(region):
...
+
+    Parameters
+    ----------
+    geom : `~gammapy.maps.WcsGeom`
+        WcsGeom within which to search for bright stars
+    max_star_mag : float, optional
+        The maximum stellar magnitude (Johnson-Cousins B-band) to exclude. Default is 6th magnitude.
+    star_rad : `~astropy.units.Quantity`, optional
+        Radius to exclude around each star. Default is 0.3 degrees.
+    star_table : path to file containing a Table
```
bkhelifi:
```suggestion
    star_table : str, optional
```

gammapy/utils/regions.py
```diff
@@ -422,3 +425,54 @@ def region_circle_to_ellipse(region):
...
+        Table containing the list of stars to search for exclusion regions. If None, the Hipparcos Catalog (Perryman et al., 1997) contained in ``$GAMMAPY_DATA//veritas/crab-point-like-ED`` is used. The table's column names should match those in ``$GAMMAPY_DATA//veritas/crab-point-like-ED``.
```
bkhelifi:
```suggestion
        Full name of the file containing a table containing the list of stars to search for exclusion regions. If None, the Hipparcos Catalog (Perryman et al., 1997) contained in ``$GAMMAPY_DATA//veritas/crab-point-like-ED`` is used. The table's column names should match those in ``$GAMMAPY_DATA//veritas/crab-point-like-ED``.
```

registerrier (approved):
Thanks @samanthalwong !

I have applied formatting rules directly on your branch.

PR #5947, opened 2025-06-25
[PIG] Systematics Modifier API

StFroese:
This pull request proposes the handling of systematic effects on the IRFs.

### DIFF
--- a/docs/development/pigs/pig-030.rst
+++ b/docs/development/pigs/pig-030.rst
@@ -1,0 +1,447 @@
+.. include:: ../../references.txt
+
+.. _pig-027:
+
+**************************************
+PIG 30 - Systematics Modifier API
+**************************************
+
+* Author: Stefan Fröse, Axel Donath, Tim Unbehaun
+* Created: ``2025-06-14``
+* Status: Draft
+* Discussion: `#XXXX`_
+
+Abstract
+========
+
+This PIG proposes an API for modeling IRF systematics in Gammapy using a ``Modifier`` system, similar to the implementation found in
+`evermore <https://evermore.readthedocs.io/en/latest/>`_.
+It is centered around the ``NPredSystematicsModel`` class
+which wraps an existing ``SkyModel`` and updates the ``MapEvaluator`` with ``IRFs``
+by user-defined ``Modifiers``.
+Each ``Modifier`` applies an ``Effect`` (i.e., a transformation) to an ``IRF``,
+based on one or more ``Parameters``, enabling flexible, parameter-driven changes to ``IRF`` data or ``IRF`` coordinates.
+
+
+Motivation
+==========
+
+Gammapy currently lacks a general mechanism to incorporate instrumental systematics into likelihood fitting.
+While the ``IRFs`` can technically be modified before fitting,
+no standard interface exists for linking such modifications to fit parameters and priors.
+This limits reproducibility, composability, and clarity of systematic treatments.
+
+The ``NPredSystematicsModel`` enables this by encapsulating parameter-dependent transformations applied to the ``IRFs`` with ``Modifiers`` and ``Effects``,
+while leaving datasets untouched.
+This mirrors the approach taken in particle physics (e.g. modifiers + nuisance parameters) and provides a modern, modular system for handling systematics.
+
+Use cases
+=========
+
+1. Exposure scaling
+
+.. code-block:: python
+
+    exp_parameter = Parameter('exp_scaling', 1)
+    exp_parameter.prior = GaussianPrior(mu=1, sigma=0.2)
+
+    scale_effect = ScalingEffect()
+
+    exposure_modifier = Modifier(exp_parameter, scale_effect)
+
+    model = crab_point_source_model()
+
+    model_npred = NPredSystematicsModel.from_dataset(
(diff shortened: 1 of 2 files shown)


StFroese:
@AtreyeeS @cosimoNigro @LauraOlivera @QRemy @registerrier you all took part in the discussion :) react to this message if you want to be added as co-author :)

LauraOlivera:
I don't need to be a co-author (as I have done nothing!) but I definitely support this - if that is what "authorship" means in this context then feel free to add me

registerrier:
Thanks for the draft @StFroese . A few general questions:
- isn't it more natural to consider IRFs as model themselves. For instance, the `FoVBackgroundModel` is already a an IRF     
  that is modified by modifier models and their nuisance  parameters. (either `spectral_model` or `spatial_model`). The 
  issue of IRF modifications is not conceptually different. In particular, one can already represent a modified exposure with a  `TemplateSpatialModel` associated to a spectral model.
- Is scaling and shift the only two possible transforms? For instance, if you want to take into account statistical uncertainties in the construction of aeff, you could represent this with a collection of scaling at various nodes (e.g. a`PiecewiseNormSpectralModel`)
- Would the concept of model composition help (see e.g. how it is done in astropy.model)?

docs/development/pigs/pig-030.rst
```diff
@@ -0,0 +1,447 @@
...
+
+.. code-block:: python
+
+   class LinearScalingEffect(Effect):
+        def __call__(self, data, parameters, apply_to_axes=None):
+            coords = data.get_coord()
+            scaling = 1
+            for i, ax in enumerate(self.depends_on_axes):
+                alpha = parameters[2 * i].value
+                beta = parameters[2 * i + 1].value
```
registerrier, line 230: In your example, this represents both shift and scaling, no? It is already the composition of the two effects.

docs/development/pigs/pig-030.rst
```diff
@@ -0,0 +1,447 @@
...
+            coords_scaled = MapCoord.create(coords, frame=irf.geom.frame)
+            new_data = irf.interp_by_coord(coords_scaled)
+            return irf.copy(data=new_data)
+
+Effects
+-------
+
+``Effects`` define the mathematical transformation applied to the ``IRF`` *data* or *coordiantes*/*axes*.
+Moreover, ``Effects`` can depend on one or multiple axes, which leads to parameterized transformations, e.g. ``f(enery, offset, ...)``.
+We distinguish between ``ScalingEffects`` and ``ShiftingEffects``.
```
registerrier, line 174: Note that for spatial uncertainty there could be a rotation effect as well e.g. to account for astrometry errors

docs/development/pigs/pig-030.rst
```diff
@@ -0,0 +1,447 @@
...
+--------
+
+A ``Modifier`` links a single instance or list of ``Parameter`` instances to an ``Effect``.
+The modifier is responsible for calling the effect with the current parameter values and applying it to
+either the IRF *data* or its *axis* coordinates (e.g. shifting the ``energy`` axis).
+In the latter case the transformation defined by the ``Effect`` is applied by interpolating the ``IRF`` to new transformed coordinates.
+
+.. code-block:: python
+
+    class Modifier:
```
registerrier, line 152: Shouldn't it be a `Model`? 
In that case, internally, it is a model that take a `MapCoord` and yields a `MapCoord` i.e. from R^3 to R^3.

You could then represent it as `IRFmodel(irf_map, spectral_modifier, spatial_modifier, coord_effect)`
Then when evaluating the IRf model you have:
- new_coords = coord_effect(coords)
- return irf_map(new_coords)*spectral_modifier(new_coords[energy])*spatial_modifier(new_coords[space])

QRemy:
Why the IRFs modification apply to a specific source via `NPredSystematicsModel`, it should apply to all the sources.
Moreover recreating the `MapEvaluator` for each evaluation of every source will create a huge drop in performance and break the caching.
I agree with @registerrier, the modified irfs should appear as models like the `FoVBackgroundModel`
coulb d be PSFModel or AeffModel, or EdispModel... 
The effects would be simply new classes of norm models (like the PiecewiseNormSpatialModel... but potentially applicable to other axes)  attached to the IRFS models.
The `MapEvaluator` should be adapted to recompte the maps only if the parameters of theses models changed.
 
This way we can fit everything in our modelling framework  and we don't have to  introduce new concepts that could confuse user.

StFroese:
@registerrier @QRemy hi 😊 thanks for all review and helpful comments! 
I agree with you that your idea fits better into the gammapy framework without increasing the computational effort so largely. 
I will answer to the comments in more detail soon. 
I would propose that try rework this pig a bit to make it more aligned with the comments and will add a minimal working example notebook in a gist

registerrier (commented):
Systematics handling is definitely something gammapy needs to better support. The adequacy of this proposal with the current code structure is not so clear and in the absence of updates in the last year, we propose to withdraw it and keep it for reference discussion.

registerrier (approved):
Thanks @StFroese . PIG is withdrawn and will be included as is in the doc. The discussion here will serve also as a future reference. 
Merging this.

PR #5949, opened 2025-06-27
Adapt tutorial order

### LINKED ISSUE #5796: Proposal for re-ordering the tutorials
Following a discussion at the codingsprint, we realize that some basic concepts of the analysis steps could benefit from more attention in the tutorial page.

In order to tackle this issue, we propose to re-order the tutorials in the following way :  

Just  between the `Data analysis` and the `1D Spectral` create a new section that would be called `Key analysis steps `.
It would contain a good fraction of the notebooks that are currently in the API section but re-order so to follow the steps of an analysis workflow. Namely  : 

- IRFS
- Observational clustering 
- Data reduction
- Datasets 
- Maps 
- Mask
- Catalogs
- Models
- Priors
- Modeling 
- Fitting
- Bayesian sampling
- Estimators 

This would imply that the dark matter tutorial in the API would be left orphan but it could be added in a new section together with EBL and energy dependent morphology ? (which are advanced astro use case ).

Comments are welcome !

Astro-Kirsty:
This PR reorders the tutorials to resolve #5796

It also refactors the related code


AtreyeeS:
This is a major change. Is this necessary for 2.0?

bkhelifi:
> This is a major change. Is this necessary for 2.0?

If the effort is already made, why not! And an LTS is the perfect moment for improvement...
But, conflicts should be solved first to generate the new documentation by the CI, which is very important to look at here.

docs/conf.py
```diff
@@ -320,12 +320,13 @@ def setup(app):
             "../examples/models/spectral",
             "../examples/models/temporal",
             "../examples/tutorials/starting",
+            "../examples/tutorials/model-gallery",
             "../examples/tutorials/data",
+            "../examples/tutorials/key-steps",
```
AtreyeeS: Did we decided on calling it `key-steps` ? It sounds that one *must* go though these notebooks before doing any analysis, whereas some of the ideas presented here are rather advanced (eg: nested sampling, asymmetric IRFs, priors, etc)

AtreyeeS (commented):
Thanks fir the effort @Astro-Kirsty ! There are merge conflicts that need to be resolved

docs/conf.py
Astro-Kirsty: We had a small discussion in the dev call to call it this, but we could find another name

AtreyeeS:
The proposal here is to create an Astrophysics section with EBL, Energy dep morphology., dark matter and pulsars.

The others can go a "Detailed explanation" section

QRemy:
> The others can go a "Detailed explanation" section

Having one line description explaining that their order follows an analysis workflow could help.

examples/tutorials/details/README.rst
```diff
@@ -0,0 +1,5 @@
+Detailed explanation
+====================
+
+The following tutorials demonstrate different dimensions of the Gammapy API.
+The order follows the typical analysis workflow.
```
QRemy:
```suggestion
From Data to Discovery with Gammapy
===================================

This series of tutorials demonstrates how to use Gammapy tools across the full analysis chain, from data loading to model comparison.
```

QRemy (dismissed):
Thanks @Astro-Kirsty,
Here is a AI suggestion  to make the title of this section more catchy (and somehow in line with CTAO website):

examples/tutorials/details/README.rst
Astro-Kirsty: I don't feel that this title really matches the section, we are showing how to use gammapy. In this section we will not discover anything!

Astro-Kirsty:
The docs build for this PR can be found here:
https://astro-kirsty.github.io/gammapy-docs-preview/tutorials/index.html

examples/tutorials/details/README.rst
QRemy: Yes the title is a bit too much, but the description is not bad.
I think we should avoid using the term API again because almost no one knows what it is.

morcuended:
Thanks @Astro-Kirsty. It's looking good. Some comments and suggestions:
- I would remove the model gallery from tutorials section. It is actually shown in the user guide, where I think it rather belongs to.
- Renaming of some sections:
  - Time -> Time series analysis
  - Data analysis -> 1D Data analysis
  - Scripts -> Scripting or Analysis Scripting
- Ordering of the sections:
  - Introduction
  - Data exploration
  - 1D Data analysis
  - 2D Image
  - 3D Cube
  - Time
  - Detailed explanation
  - Astrophysics use cases
  - Scripting

Astro-Kirsty:
Thanks @morcuended. I agree with the title changes. For the 1D analysis, it is correctly named but due to the number of subheadings it cannot show properly. See [here](https://astro-kirsty.github.io/gammapy-docs-preview/tutorials/analysis-1d/index.html)

As for the `model gallery` this was added as per our user feedback that we had. So we should leave it, but rather move the section to be after the Data exploration.

We decided to make sure this new implemented section was up higher, hence its current placement. We have some key concepts in this section that should be understood by the user, and by having it so far down most users were not even looking at it.

examples/tutorials/details/README.rst
morcuended: What about naming it "In depth analysis" or simply "In depth"? Although I agree also with "Detailed explanation"

Astro-Kirsty: Its a hard title to come up with, because I think `in depth analysis` is not really correct. As most of them are not really `analysis`?

QRemy: "Detailed explanation" is fine

registerrier (approved):
Thanks @Astro-Kirsty . This is nice. 

Unless there is a strong objection on the re-ordering, I propose that we merge as is and revisit later after the new structure is published in the dev docs.

PR #5954, opened 2025-06-27
Separate internal Eventlist data model from gadf format

registerrier:
**Description**

This pull request introduces a gammapy specific model for `EventList` table. 
The required columns are `RA`, `DEC`, `ENERGY` and `TIME`. The three first are `Quantity` objects, the latter a `Time` object.

A validation is performed on __init__. It requires adapting numerous tests which define `TIME` as a time like quantity.

Conversion methods are added to adapt format for and to gadf. They should go in the `gammapy.data.io.EventListReader` and `Writer`.
 
Significant changes are necessary in `gammapy.datasets.simulate` since the `MapDatasetEventSampler` was building the `EventList` (in the gadf format) progressively. This cannot work now because of format validation. The sampler is modified to build a table that is converted to `EventList` at the end.

This PR is part of the effort discussed in  #3767  and specifically solves #4239 . The only remaining thing is to rely on the proper `ObservationMetadata` instead of elements stored in the `table.meta`. 

**Dear reviewer**

### DIFF
--- a/gammapy/data/event_list.py
+++ b/gammapy/data/event_list.py
@@ -75,14 +75,16 @@ class EventList:
 
     """
 
-    REQUIRED_COLUMNS = {"radec": SkyCoord, "energy": u.Quantity, "time": Time}
+    REQUIRED_COLUMNS = {
+        "RA": u.Unit("deg"),
+        "DEC": u.Unit("deg"),
+        "ENERGY": u.Unit("TeV"),
+        "TIME": Time,
+    }
 
     def __init__(self, table, meta=None):
-        if "RA" in table.colnames:
-            table = self._from_gadf_table(table)
         self.table = self._validate_table(table)
         self.meta = meta
-        print(self.meta)
 
     @staticmethod
     def _validate_table(table):
@@ -92,17 +94,25 @@ class EventList:
             raise TypeError(
                 f"EventList expects astropy Table, got {type(table)} instead."
             )
-        #
-        #     missing_columns = set(cls.REQUIRED_COLUMNS.keys()).difference(table.colnames)
-        #     if len(missing_columns) > 0:
-        #         raise ValueError(
-        #             f"EventList table invalid: columns {missing_columns} are missing."
-        #         )
-        #
-        #     for name, check in REQUIRED_COLUMNS.items():
-        #         if not isinstance(table[name], check):
-        #             raise TypeError(f"Column {name} is not a {check} object.")
-        #
+
+        missing_columns = set(EventList.REQUIRED_COLUMNS.keys()).difference(
+            table.colnames
+        )
+        if len(missing_columns) > 0:
+            raise ValueError(
+                f"EventList table invalid: columns {missing_columns} are missing."
+            )
+
+        for name, check in EventList.REQUIRED_COLUMNS.items():
+            if not isinstance(check, u.Unit):
+                if not isinstance(table[name], check):
+                    raise TypeError(f"Column {name} is not a {check} object.")
+            else:
+                if not check.is_equivalent(table[name].unit):
+                    raise u.UnitConversionError(
+                        f"Column {name} is not in {check} unit."
+                    )
+
         return table
 
     def _repr_html_(self):
@@ -114,18 +124,23 @@ class EventList:
     @staticmethod
     def _from_gadf_table(table):
         """Temporary gadf table converter."""
+        if not isinstance(table, Table):
+            raise TypeError(
+                f"_form_fgadf_table expects astropy Table, got {type(table)} instead."
+            )
 
         met = u.Quantity(table["TIME"].astype("float64"), "second")
         time = time_ref_from_dict(table.meta) + met
 
         energy = table["ENERGY"].quantity
 
-        radec = SkyCoord(table["RA"], table["DEC"], unit="deg", frame="icrs")
+        ra = table["RA"].quantity
+        dec = table["DEC"].quantity
 
         removed_colnames = ["RA", "DEC", "GLON", "GLAT", "TIME", "ENERGY"]
 
         new_table = Table(
-            {"TIME": time, "ENERGY": energy, "RADEC": radec}, meta=table.meta
+            {"TIME": time, "ENERGY": energy, "RA": ra, "DEC": dec}, meta=table.meta
         )
         for name in table.colnames:
             if name not in removed_colnames:
@@ -261,7 +276,7 @@ class EventList:
     @property
     def radec(self):
         """Event RA / DEC sky coordinates as a `~astropy.coordinates.SkyCoord` object."""
-        return self.table["RADEC"]
+        return SkyCoord(self.table["RA"], self.table["DEC"], unit="deg", frame="icrs")
 
     @property
     def galactic(self):
--- a/gammapy/data/event_list.py
+++ b/gammapy/data/event_list.py
@@ -99,7 +99,7 @@ class EventList:
             table.colnames
         )
         if len(missing_columns) > 0:
-            raise ValueError(
+            raise KeyError(
                 f"EventList table invalid: columns {missing_columns} are missing."
             )
 
--- a/gammapy/data/event_list.py
+++ b/gammapy/data/event_list.py
@@ -83,13 +83,25 @@ class EventList:
     }
 
     def __init__(self, table, meta=None):
+        if table is None:
+            table = self._create_empty_list()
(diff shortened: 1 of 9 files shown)


gammapy/data/event_list.py
```diff
@@ -28,30 +30,13 @@ class EventList:
...
 
-    The most important reconstructed event parameters
-    are available as the following columns:
+    The required columns are:
 
-    - ``TIME`` - Mission elapsed time (sec)
-    - ``RA``, ``DEC`` - ICRS system position (deg)
-    - ``ENERGY`` - Energy (usually MeV for Fermi and TeV for IACTs)
+    - ``RA`` - ICRS system reconstructed right ascension
+    - ``DEC`` - ICRS system reconstructed declination
```
fabiopintore: you might want to add the units for Ra and Dec as well.

gammapy/data/event_list.py
```diff
@@ -87,16 +72,100 @@ class EventList:
...
+
     def __init__(self, table, meta=None):
-        self.table = table
+        if table is None:
+            table = self._create_empty_list()
+        self.table = self._validate_table(table)
         self.meta = meta or EventListMetaData()
 
+    @staticmethod
+    def _create_empty_list():
```
Tobychev, line 82: I think a cleaner solution is to just specify the columns explicitly directly in the function
```
columns = [
    Column(name="RA", unit=u.deg, description="Event right ascension"),
    Column(name="DEC", unit=u.deg, description="Event declination"),
    Column(name="ENERGY", unit=u.TeV, description="Event energy"),
]
table = Table(columns)
table["TIME"] = Time((), format="mjd")
```

gammapy/data/event_list.py
```diff
@@ -87,16 +72,100 @@ class EventList:
...
+        "RA": u.Unit("deg"),
+        "DEC": u.Unit("deg"),
+        "ENERGY": u.Unit("TeV"),
+        "TIME": Time,
+    }
+
     def __init__(self, table, meta=None):
-        self.table = table
+        if table is None:
+            table = self._create_empty_list()
```
Tobychev, line 77: If you `vstack` the incoming table with the one created by `create_empty_list()` using the arguments `join_type='exact', metadata_conflicts='error'` you can let astropy do the validation. 

The error messages aren't so nice in case of failure though, so it could make sense to catch try-except the vstack and let the `_validate_table` function provide a nicer error message.

gammapy/data/event_list.py
```diff
@@ -87,16 +72,100 @@ class EventList:
...
+        self.table = self._validate_table(table)
         self.meta = meta or EventListMetaData()
 
+    @staticmethod
+    def _create_empty_list():
+        """Create empty event list table."""
+        table = Table()
+        for name, unit in EventList.REQUIRED_COLUMNS.items():
+            if name == "TIME":
+                table[name] = Time((), format="mjd")
```
fabiopintore: I would also add the `scale` for the `Time` object

fabiopintore (commented):
thanks @registerrier . It looks very good to me. I've left some inline comments.

registerrier:
This is failing and I really don't understand what this is supposed to test:
https://github.com/gammapy/gammapy/blob/bcd499e38369a9d2488c0ec779e3ae30c79d397c/gammapy/makers/tests/test_map.py#L612-L616
Any  idea @QRemy ?

QRemy:
> This is failing and I really don't understand what this is supposed to test:
> 
> https://github.com/gammapy/gammapy/blob/bcd499e38369a9d2488c0ec779e3ae30c79d397c/gammapy/makers/tests/test_map.py#L612-L616
> 

It test that the gti cannot be redefined if it exits and the events exits
```
        if self.gti is not None and self.events is not None:
            raise ValueError("GTI cannot be redefined if events is not None.")
```

registerrier:
OK. So the problem is that here `obs.events` is `None` whereas it wasn't before.

In fact, the error is subtle.

In the SWGO test data there is an event table but it does not contain any column. It therefore does not follow the gadf standard. The reader should fail with a `KeyError` when trying to convert to the internal data format.

But what happens is that the `LazyFitsData.__get__()` is doing a `try except` and catches the `KeyError` and silently does set any `obs.events`:
https://github.com/gammapy/gammapy/blob/91e3d8357f398d3135599bf531247dcd197d64ab/gammapy/utils/fits.py#L133-L149

There are therefore two issues:
1. the try/except clause in `LazyFitsData` prevents validation error on read to raise a KeyError
  - option 1: we try to raise a different error in case the format does not comply
  - option 2: we explicitly check whether the key is in the `instance.__dict__` instead of using the try/except clause.
  - opinion @adonath @QRemy ?
2. the event list from the SWGO test data does not comply with GADF. Accessing the `events` attribute should raise some kind of exception, no? Or should it go with a simple warning a create an empty event list?

QRemy:
> 2. the event list from the SWGO test data does not comply with GADF. Accessing the `events` attribute should raise some kind of exception, no? Or should it go with a simple warning a create an empty event list?

The event list is there only because we need the meta data that are found in the events header according to GADF. AS long as these info are set correctly it's fine if the event list returned is None or an empty event list.

You could just modify the test as
```
       t = Table()
       t["EVENT_ID"] = np.array([1, 5], dtype=np.uint16)
       t["RA"] = [5, 11] * u.deg
       t["DEC"] = [0, 0] * u.deg
       t["ENERGY"] = [10, 12] * u.TeV
       t["TIME"] = Time("2025-01-01") + [3, 4] * u.s
       obs.events = EventList(t)
       obs.gti = gti
``` 
and move this at the end of the test to be sure that the maker works correctly if the events are not set.

QRemy:
> 1. the try/except clause in `LazyFitsData` prevents validation error on read to raise a KeyError
> 
> * option 1: we try to raise a different error in case the format does not comply
> * option 2: we explicitly check whether the key is in the `instance.__dict__` instead of using the try/except clause.
> * opinion @adonath @QRemy ?
> 

No strong opinion on that as long as don't raise for the hdus that are not in `required_hdus`.

gammapy/data/event_list.py
registerrier: done

gammapy/data/event_list.py
registerrier: added "TT" but do we want to impose event times in a specific scale?

gammapy/data/event_list.py
```diff
@@ -28,30 +30,13 @@ class EventList:
...
-    are available as the following columns:
+    The required columns are:
 
-    - ``TIME`` - Mission elapsed time (sec)
-    - ``RA``, ``DEC`` - ICRS system position (deg)
-    - ``ENERGY`` - Energy (usually MeV for Fermi and TeV for IACTs)
+    - ``RA`` - ICRS system reconstructed right ascension (deg)
+    - ``DEC`` - ICRS system reconstructed declination (deg)
+    - ``TIME`` - event time as an `~astropy.time.Time` object
+    - ``energy`` - Reconstructed energy (usually MeV for Fermi and TeV for IACTs)
```
QRemy: Not ENERGY in capital as the other ?

QRemy:
```suggestion
    - ``ENERGY`` - Reconstructed energy (usually MeV for Fermi and TeV for IACTs)
```

gammapy/datasets/simulate.py
```diff
@@ -342,39 +340,39 @@ def sample_background(self, dataset):
                 table.meta["MID{:05d}".format(0)] = 0
                 table.meta["MMN{:05d}".format(0)] = dataset.background_model.name
 
-        return EventList(table)
+        return table  # EventList(table)
```
QRemy:
```suggestion
        return table
```
and why it doesn't return an EventList now ?

gammapy/data/event_list.py
AtreyeeS:
```suggestion
    - ``energy`` - Reconstructed energy (usually MeV for Fermi and TeV for IACTs/WCDs)
```

gammapy/datasets/simulate.py
registerrier: Because the rest of the sampler cannot return an eventlist until `sample_edisp` is applied (it needs reco energy). For simplicity I decide to create only tables and instantiate the `EventList` after stacking source photon and background tables.

gammapy/data/event_list.py
registerrier: done

PR #5957, opened 2025-06-29
Fix  #5951

### LINKED ISSUE #5951: Repetitive wording of "Exercises" in Spectral Analysis tutorial
Exercises state : "Fit a different spectral model to the data. You could try [ExpCutoffPowerLawSpectralModel] or LogParabolaSpectralModel"

However, in the tutorial it's already an ExpCutoffPowerLawSpectralModel that is used.

AtreyeeS:
I removed the entire `Exercises` section as whatever was proposed is already showing in the notebook.

Fixes  #5951


Tobychev (commented):
I don't think entirely removing the exercise section is a proper solution, especially as the suggested log parabola fit isn't performed.

bkhelifi (approved):
Thanks a lot

PR #5959, opened 2025-06-30
Adapt SafeMaskMaker docstring

Astro-Kirsty:
The `SafeMaskMaker` can be a difficult concept to grasp.
We don't specify what each of the methods do, I try to fix that here

### DIFF
--- a/gammapy/makers/safe.py
+++ b/gammapy/makers/safe.py
@@ -30,25 +30,37 @@ class SafeMaskMaker(Maker):
     methods : {"aeff-default", "aeff-max", "edisp-bias", "offset-max", "bkg-peak"}
         Method to use for the safe energy range. Can be a
         list with a combination of those. Resulting masks
-        are combined with logical `and`. "aeff-default"
-        uses the energy ranged specified in the DL3 data
-        files, if available.
-    aeff_percent : float
+        are combined with logical `and`.
+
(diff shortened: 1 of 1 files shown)


gammapy/makers/safe.py
```diff
@@ -30,25 +30,37 @@ class SafeMaskMaker(Maker):
...
+        are combined with logical `and`.
+
+            * "aeff-default" : uses the energy ranged specified in the DL3 data
+              files, if available
+            * "aeff-max": the lower energy threshold is determined such as the effective
+              area is above a given percentage of its maximum
+            * "edisp-bias" : the lower energy threshold is determined such as the energy
+              bias is below a given percentage
+            * "offset-max" : the data beyond a given offset radius from the observation center are excluded
+            * "bkg-peak" : the energy threshold is defined as the upper edge of the energy bin
```
AtreyeeS: Do we use the upper edge or the lower edge?

Astro-Kirsty: I think we use the lower edge but as I understood from the paper it is the upper edge. I'm not certain now

registerrier: We use the lower edge. To avoid always removing the first bin.

registerrier:
```suggestion
            * "bkg-peak" : the energy threshold is defined as the lower edge of the energy bin
```

registerrier (approved):

PR #5960, opened 2025-06-30
Energy edges instead of center for `aeff-max` in `SafeMaskMaker`

### LINKED ISSUE #5834: aeff-max option in SafeMaskMaker always removes the first energy bin at DL4 level
**Gammapy version**
1.3
Same code in main

**Bug description**
When applying a SafeMaskMaker using the aeff-max option on DL4 data the first energy bin is always masked, independently of the effective area.

**Expected behavior**
Only bins for which the aeff drops below the input percentage at the edge should be removed (this seems to be the case in the of DL3, which is handled differently, see `gammapy.makers.safe.py` l205 `if self.irfs == "DL3"`). DL4 is handled by the else of the same `if`.

**To Reproduce**
Create a DL4 dataset (with an aeff not dropping below the requested percentage)
Initiate a safe make maker : safe_mask_masker = SafeMaskMaker(methods=["aeff-max"], aeff_percent=0.001)
Run it : safe_mask_masker.run(dataset)

**Other information**
Looking in the code, the issue is that the selection of the min_energy is done with the energy bin center, meaning that the minimum min_energy applied for the mask is the lower bin center. Which removes the first bin in the following line:

`gammapy.makers.safe.py` l 242 `gammapy.makers.safe.py`

Astro-Kirsty:
Fixes #5834

### DIFF
--- a/gammapy/makers/safe.py
+++ b/gammapy/makers/safe.py
@@ -227,7 +227,7 @@ class SafeMaskMaker(Maker):
 
             model = TemplateSpectralModel.from_region_map(aeff)
 
-            energy_true = model.energy
+            energy_true = dataset._geom.axes["energy"].edges
             energy_min = energy_true[np.where(model.values > 0)[0][0]]
             energy_max = energy_true[-1]
 


bkhelifi:
Maybe one could add a specific test linked to the issue

AtreyeeS:
I am a bit unsure here. By definition, the IRFs are defined in true energies on the bin centers, so maybe the present behaviour is correct. Ideally, one should bin the true energy axis finely enough that this should not matter, no? @bkhelifi ?

bkhelifi:
In reality, when reading IRFs written in GADF, the user cab not know whether the values are really the value at the bin centre or the one from the integrated value within the bin (divided by DeltaE). From my knowledge of HESS, VERITAS, MAGIC and CTAO usage, most of the time the `aeff` is computed by counting the number of counts within a bin. The value is then not fully differential (exact value at a given fixed energy). So, this is why I think that the proposed behaviour in this PR is in line with the most common IRF production

AtreyeeS (approved):
Thanks @Astro-Kirsty ! No further comments then

gammapy/makers/safe.py
```diff
@@ -239,7 +239,7 @@ def make_mask_energy_aeff_max(self, dataset, observation=None):
 
             model = TemplateSpectralModel.from_region_map(aeff)
 
-            energy_true = model.energy
+            energy_true = dataset._geom.axes["energy"].edges
```
registerrier, line 242: Beware `_geom` will give you counts geometry

registerrier:
Note from the dev call: this is not working the way it is implemented. edges are in true energy and one cannot guarantee they will match the recon ones. 
Simple option is to explain that you can simply not use this safe mask option in this situation.
More complex one is to modify the logic to compute threshold in reco energy taking into account the edisp (if available).

AtreyeeS:
As discussed in the last dev call, this does not address the problem. We can think of a different algorithm taking into account the edsip in future

Issue #5943, opened 2025-06-23
DataStore.get_observations duplicates


registerrier:
**Gammapy version**
1.3 dev 

**Bug description**
`DataStore.get_observations(obs_id)` will return duplicates of observation if their obs_id are duplicated.

Since it is clear now that `OBSID` must be unique, we could fail if one of them is duplicated.

**Expected behavior**
Raise an error if list of obs_id is not equal to list of unique obsid

**To Reproduce**
```
datastore = DataStore.from_dir("$GAMMAPY_DATA/hess-dl3-dr1")
datastore.get_observations(obs_id=[23523, 23523])
```

**Other information**
Any other information you think will be useful for us to fix the issue can go here.

AtreyeeS:
Note that changing this will mean adapting all our benchmarks, where we rely on this to read N observations. eg

https://github.com/gammapy/gammapy-benchmarks/blob/66143e05073ed392c39cf7cbd23fc8d8055a3e5b/benchmarks/analysis_3d.py#L25

LeanderSchlegel:
Thank you for assigning me to the issue! As @registerrier suggested to me, there could be three possible solutions:

1) Initial check in `DataStore.get_observations(obs_id)` if `obs_id` is unique - otherwise fail.
2) Initial check in `DataStore.get_observations(obs_id)`if  `obs_id` is unique - otherwise raise a UserWarning.
3) Add keyword-argument in `DataStore.get_observations()` like "accept_duplicates" with default on False, to create the option to accept this case of not unique `obs_id`.

As suggested to me by @registerrier, a good solution could maybe be the middle-way in 2) of raising a warning to the user.
I will try to prepare the fix!

PR #5964, opened 2025-07-02
Fix for Issue #5943

LeanderSchlegel:
Dear all,

this PR is intended to resolve the bug described in #5943. 

Thank you @registerrier for introducing me to the issue and the workflow hereby. As you suggested to me, to solve the bug I implemented a UserWarning in `get_observations(obs_id)` that is raised if `obs_id` is not unique but contains e.g. duplicates, and added a corresponding test (analogue to the existing ones) that checks, if this warning is raised from now on.

@registerrier Would this be the solution you had in mind?

Thank you a lot,

Leander

### DIFF
--- a/codemeta.json
+++ b/codemeta.json
@@ -359,8 +359,10 @@
         "active"
     ],
     "email": "GAMMAPY-COORDINATION-L@IN2P3.FR",
-    "dateModified": "2025-07-01",
-    "softwareRequirements": [
+
+    "dateModified": "2024-11-29",
+
+  "softwareRequirements": [
         "numpy>=1.21",
         "scipy>=1.5,!=1.10",
         "astropy>=5.0",
@@ -372,3 +374,4 @@
         "matplotlib>=3.4, <3.10"
     ]
 }
+
--- a/codemeta.json
+++ b/codemeta.json
@@ -359,10 +359,8 @@
         "active"
     ],
     "email": "GAMMAPY-COORDINATION-L@IN2P3.FR",
-
-    "dateModified": "2024-11-29",
-
-  "softwareRequirements": [
+    "dateModified": "2025-07-01",
+    "softwareRequirements": [
         "numpy>=1.21",
         "scipy>=1.5,!=1.10",
         "astropy>=5.0",
@@ -374,4 +372,3 @@
         "matplotlib>=3.4, <3.10"
     ]
 }
-
--- a/gammapy/data/data_store.py
+++ b/gammapy/data/data_store.py
@@ -396,7 +396,7 @@ class DataStore:
             obs_id = self.obs_ids
 
         if len(np.unique(obs_id)) != len(obs_id):
-            raise UserWarning(f"List of obs_id is not unique!")
+            raise UserWarning("List of obs_id is not unique!")
 
         obs_list = []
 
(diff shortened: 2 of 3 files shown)


codemeta.json
Tobychev, line 1: Pretty sure you shouldn't modify this file, I think the modification date refers to when the author or dependency list was last changed.

registerrier: This is automatically modified on users forks when they run the CI. 
One option is to prevent workflows to be run on forks.
Another is to change the cffconvert workflow.
Here you can revert the automated commit.

LeanderSchlegel: Thank you very much for your feedback and the suggestion @Tobychev, @registerrier. I reverted the commit that changed the codemeta.json file.

Tobychev (approved):

gammapy/data/data_store.py
```diff
@@ -395,6 +395,9 @@ def get_observations(
         if obs_id is None:
             obs_id = self.obs_ids
 
+        if len(np.unique(obs_id)) != len(obs_id):
+            raise UserWarning("List of obs_id is not unique!")
```
AtreyeeS: Maybe it would be better to show in the warning the specific obs_ids that appear multiple times

gammapy/data/tests/test_data_store.py
```diff
@@ -257,6 +257,16 @@ def test_data_store_bad_name():
...
 
 
+@requires_data()
+def test_data_store_doubled_id():
+    """Test that doubled id in obs_id causes UserWarning"""
+
+    # Using example provided in Issue #5943:
+    datastore = DataStore.from_dir("$GAMMAPY_DATA/hess-dl3-dr1")
+    with pytest.raises(UserWarning):
+        datastore.get_observations(obs_id=[23523, 23523])
```
AtreyeeS: It can be useful to check the exact warning message, see https://docs.gammapy.org/dev/development/dev_howto.html#caplog-fixture

gammapy/data/data_store.py
AtreyeeS: We normally use `log.warning("Warning message")`

LeanderSchlegel: Thank you very much for your helpful comments and suggestions @AtreyeeS! 
I just implemented to use log.warning like you suggested and to check for the exact message using the caplog by using the code-snippet provided in the docs.

LeanderSchlegel: In the second commit I implemented your suggestion to include the specific obs_id appearing multiple times into the warning message. Is there maybe a better way to find the multiples, or would this be an okay solution?

gammapy/data/data_store.py
```diff
@@ -395,6 +395,12 @@ def get_observations(
         if obs_id is None:
             obs_id = self.obs_ids
 
+        if len(np.unique(obs_id)) != len(obs_id):
+            multiples = np.array(
+                np.unique(obs_id)[(np.unique(obs_id, return_counts=True)[1] > 1)]
```
AtreyeeS: Maybe have the call to `np.unique` only once, and store it in a variable

LeanderSchlegel: Thank you a lot for your review and your suggestion, @AtreyeeS! I implemented your suggestion and it should be more efficient and I hope also better readable now :)

registerrier (approved):
Thanks @LeanderSchlegel . This looks good.

In general, we do use `log.warning` messages. We have not defined a clear definition of what type of warnings should be raised and what type should be logged. So this is fine like this.

Issue #5970, opened 2025-07-04
SpectrumDataset with chi2 statistics


mireianievas:
Right now we have SpectrumDataset which use cstat/wstat, and FluxPointDatasets which use chi2 statistics. For IXPE (X-ray polarization) the high-level ogip files fitted by xspec are actually SpectrumDataset-like objects, but instead of having counts and background, they have excess and sigma columns, and chi2 statistics are used instead of Poisson. 

It would be nice to support this in gammapy. I will create a PR to start the discussion.

PR #5971, opened 2025-07-04
SpectrumDatasetChi2 (SpectrumDataset with chi2 statistics)

mireianievas:
Implements the changes needed by #5970 for the SpectrumDatasetChi2 type dataset.

### DIFF
--- a/gammapy/datasets/spectrum.py
+++ b/gammapy/datasets/spectrum.py
@@ -1,7 +1,9 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 import logging
+import numpy as np
 import matplotlib.pyplot as plt
 from matplotlib.gridspec import GridSpec
+from gammapy.maps import RegionNDMap
 from gammapy.utils.scripts import make_path
 from .map import MapDataset, MapDatasetOnOff
 from .utils import get_axes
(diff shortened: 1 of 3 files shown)


mireianievas:
Potential unit tests pending.

gammapy/datasets/spectrum.py
```diff
@@ -429,3 +431,163 @@ def to_spectrum_dataset(self, name=None):
...
+            On counts.
+        sigma : `~gammapy.maps.RegionNDMap`, optional
+            Uncertainties on counts. Must have same geometry as counts.
+            If None, will default to sqrt(counts).
+        kwargs : dict
+            Other keyword arguments passed to `SpectrumDataset`.
+        """
+        background = counts.copy(data=np.zeros_like(counts.data))
+
+        super().__init__(counts=counts, background=background, **kwargs)
```
AtreyeeS, line 451: Maybe we should avoid calling it `counts`

gammapy/datasets/spectrum.py
```diff
@@ -429,3 +431,163 @@ def to_spectrum_dataset(self, name=None):
...
+        ----------
+        counts : `~gammapy.maps.RegionNDMap`
+            On counts.
+        sigma : `~gammapy.maps.RegionNDMap`, optional
+            Uncertainties on counts. Must have same geometry as counts.
+            If None, will default to sqrt(counts).
+        kwargs : dict
+            Other keyword arguments passed to `SpectrumDataset`.
+        """
+        background = counts.copy(data=np.zeros_like(counts.data))
```
AtreyeeS, line 449: Do we need to set a background forcibly?

AtreyeeS (commented):
Thanks @mireianievas for starting this.

To understand better, the IXPE will provide IRFs like other X-ray instruments, which will be convolved with the model to produce the number of excess counts? There is no concept of background? Will our inbiult models models work here? 

I guess it is not nice to add the complications inside the `Chi2Statistic` class. Maybe one option can be that it always computes on `dataset._data` and `dataset._data` is internally set to `events` or `flux` etc.

All the plotting code should move to `gammapy.visualisation` in 2.1 hopefully.

gammapy/datasets/spectrum.py
```diff
@@ -429,3 +431,163 @@ def to_spectrum_dataset(self, name=None):
...
+
+
+class SpectrumDatasetChi2(SpectrumDataset):
+    def __init__(self, *, counts, sigma, **kwargs):
+        """
+        Parameters
+        ----------
+        counts : `~gammapy.maps.RegionNDMap`
+            On counts.
+        sigma : `~gammapy.maps.RegionNDMap`, optional
```
registerrier, line 443: Maybe `counts_errors` or `counts_stat_errors` is more explicit.

registerrier (commented):
Thanks @mireianievas . A few comments/questions:

- Is the new dataset useful beyond X-ray astronomy? If not, one could consider it rather belongs to gammapy-mwl. 
- Can there be an additional uncertainty term than the Poisson noise in the uncertainty map? 
- The default behavior in the absence of an error term could also be to use `sqrt(npred)` depending on the type on chi square you apply.

gammapy/datasets/spectrum.py
mireianievas: * IXPE IRFs are provided in a similar format than other X-ray telescopes. For example, there is a RMF file to provide the redisribution mattrix, and an ARF file that gives us the exposure/effective area for the I component. What is new is that now there are MRF files (Modulation Response Function) for the Q-U components, but they have exactly the same format as the ARF files and are applied the same way, just for the Q-U components. More info here: https://ixpeobssim.readthedocs.io/en/latest/irf.html . Statistics seem to be chi2 and the files seem to contain just background-subtracted (excess) counts.

* Chi2Statistic class. Right, but right now the class is in my opinion too restrictive. We could make a new one maybe, but honestly I think it would be better to have it generalized to cover SpectrumDatasets (even maybe MapDatasets) and treat the FluxPoints as a rather specific case. 

* Right about the plotting code, that could indeed be moved. My intention with this PR was to create a rather self-contained object that introduced minimal changes without breaking any functionality.

mireianievas: * Calling it counts or something else (excess) I have no strong opinion. But if we call it something else we probably have to reimplement quite a few things (fitting for instance, as counts is used for the likelihood, also probably plotting, residual calculation, etc).

gammapy/datasets/spectrum.py
mireianievas: If background is not set to zero (or some other value), the to_html() function breaks. This is used to display the summary with the quantities in the dataset. Something similar happened with Fermi-LAT in MapDatasets in the past, I am not sure if that one got fixed or not.

gammapy/datasets/spectrum.py
mireianievas: * I am not sure if it is of used, right now, beyond X-ray/optical astronomy. 
* you mean like a systematic term as in UVOT? I am not sure IXPE provides it. But sure, one could think of adding an additional component. But I thought we agreed that systematics in gammapy are better handled within the model using priors. 
* Agree, sqrt(npred) can be a good alternative.

bkhelifi: Is it not possible to handle the case of having error_plus and error_minus?

mireianievas: @registerrier maybe I should stick to stat_err since that is actually stored in those PHA from IXPE (even though it refers to the stat_err of the rate, not the counts)?

@bkhelifi Yes, that should be possible, but I am not aware of any instrument putting together a dataset like that, so it would be inventing a format for a potential future usage. Any concrete proposal on how to handle i) assymetric errors, ii) systematics & stat errors vs & stat errors only?

bkhelifi: The OGIP format allows that, by memory... But maybe one could do that in the second step

mireianievas:
@AtreyeeS @bkhelifi @registerrier what do we do with this? Do I try to simplify that SpectrumDatasetChi2 and move all the plotting part elsewhere to integrate it? Do we close this and keep it in gammapy-mwl?

QRemy:
Development will continue on gammapy-mwl, closing here.

PR #5973, opened 2025-07-08
Warn users when using negative amplitude models in FluxEstimator

AtreyeeS:
This will address #5906 for 2.0 release.
In future, we can explore the correct statistical handling for such cases

### DIFF
--- a/gammapy/estimators/flux.py
+++ b/gammapy/estimators/flux.py
@@ -154,6 +154,10 @@ class FluxEstimator(ParameterEstimator):
 
         energy_min, energy_max = datasets.energy_ranges
         energy_axis = MapAxis.from_energy_edges([energy_min.min(), energy_max.max()])
+        if model.evaluate(energy_max.max(), norm=self.norm.quantity).value < 0.0:
+            log.warning(
+                "Source model predicts negative counts. Results of estimator should be interpreted with caution"
+            )
 
         with np.errstate(invalid="ignore", divide="ignore"):
             result = model.reference_fluxes(energy_axis=energy_axis)
--- a/gammapy/estimators/points/sensitivity.py
+++ b/gammapy/estimators/points/sensitivity.py
@@ -1,11 +1,15 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 import numpy as np
+import logging
+
 from astropy.table import Column, Table
 from gammapy.maps import Map
 from gammapy.modeling.models import PowerLawSpectralModel, SkyModel
 from gammapy.stats import WStatCountsStatistic
 from ..core import Estimator
 from ..utils import apply_threshold_sensitivity
+
+log = logging.getLogger(__name__)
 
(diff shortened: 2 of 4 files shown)


bkhelifi (approved):
Thanks you very much, @AtreyeeS

gammapy/estimators/points/sensitivity.py
```diff
@@ -49,6 +53,10 @@ def __init__(
             spectral_model = PowerLawSpectralModel(
                 index=2, amplitude="1 cm-2 s-1 TeV-1"
             )
+        if spectral_model.integral(1 * u.TeV, 10 * u.TeV).value < 0.0:
```
QRemy: I would rather move that in the run method and evaluate at the energies relevant for the dataset

gammapy/estimators/flux.py
```diff
@@ -154,6 +154,10 @@ def run(self, datasets):
 
         energy_min, energy_max = datasets.energy_ranges
         energy_axis = MapAxis.from_energy_edges([energy_min.min(), energy_max.max()])
+        if model.evaluate(energy_max.max(), norm=self.norm.quantity).value < 0.0:
```
QRemy: maybe evaluate at the energy axis edges

gammapy/estimators/flux.py
```diff
@@ -154,6 +154,10 @@ def run(self, datasets):
...
+            log.warning(
+                "Source model predicts negative counts. Results of estimator should be interpreted with caution"
```
QRemy:
```suggestion
                "Reference source model predicts negative counts. Results of estimator should be interpreted with caution"
```

QRemy (commented):
Thanks @AtreyeeS, just some suggestions:

gammapy/estimators/points/sensitivity.py
AtreyeeS: I see your point. That was my initial idea, but  avoided that because the `run` often gives many other warnings, and this important warning can get hidden. That is why I thought of putting it in the `init`.

In this case, the user will unnecessarily waste time running the estimator. The other option would be to raise an error, but we decided against that

gammapy/estimators/points/tests/test_sed.py
```diff
@@ -865,3 +865,16 @@ def test_resample_axis():
...
+
+def test_warning_for_bad_model(caplog):
+    ds, fpe = create_fpe(PowerLawSpectralModel())
+    spectral_model = PowerLawSpectralModel()
+    spectral_model.amplitude.value = -3e-14
+    ds[0].models = SkyModel(spectral_model, name="source")
+    fpe.run(ds)
+    assert "WARNING" in [_.levelname for _ in caplog.records]
+    assert (
+        "Source model predicts negative counts. Results of estimator should be interpreted with caution"
```
AtreyeeS:
```suggestion
        "Reference source model predicts negative counts. Results of estimator should be interpreted with caution"
```

PR #5976, opened 2025-07-09
Add option to convert PointSkyRegion to CircleSkyRegion while creating masks

AtreyeeS:
It is often useful to create exclusion masks from known sources in the FoV, eg

```
regions = models.to_regions()
mask = dataset._geom.region_mask(models.to_regions(), inside=False)
```

However, when there is Point source in the FoV, the region mask does not get applied for that. This is annoying.

Here, I introduce the option to convert the PointSkyRegion to a CircleSkyRegion with a radius specified by the user and use that for the mask.

### DIFF
--- a/gammapy/maps/wcs/geom.py
+++ b/gammapy/maps/wcs/geom.py
@@ -15,7 +15,7 @@ from astropy.wcs.utils import (
     proj_plane_pixel_scales,
     wcs_to_celestial_frame,
 )
-from regions import RectangleSkyRegion, PointSkyRegion, CircleSkyRegion
+from regions import RectangleSkyRegion, PointSkyRegion, CircleSkyRegion, Regions
 from gammapy.utils.array import round_up_to_even, round_up_to_odd
 from gammapy.utils.compat import COPY_IF_NEEDED
 from ..axes import MapAxes
@@ -1007,6 +1007,9 @@ class WcsGeom(Geom):
 
         if not self.is_regular:
             raise ValueError("Multi-resolution maps not supported yet")
+
+        if not isinstance(regions, list) or isinstance(regions, Regions):
+            regions = [regions]
 
         regions1 = []
         for reg in regions:
--- a/gammapy/maps/wcs/geom.py
+++ b/gammapy/maps/wcs/geom.py
@@ -15,7 +15,7 @@ from astropy.wcs.utils import (
     proj_plane_pixel_scales,
     wcs_to_celestial_frame,
 )
-from regions import RectangleSkyRegion, PointSkyRegion, CircleSkyRegion, Regions
+from regions import RectangleSkyRegion
 from gammapy.utils.array import round_up_to_even, round_up_to_odd
 from gammapy.utils.compat import COPY_IF_NEEDED
 from ..axes import MapAxes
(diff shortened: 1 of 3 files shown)


gammapy/maps/wcs/geom.py
```diff
@@ -970,6 +970,10 @@ def region_mask(self, regions, inside=True):
             For ``inside=True``, set pixels in the region to True.
             For ``inside=False``, set pixels in the region to False.
             Default is True.
+        point_to_radius : `~astropy.units.Quantity`, optional
+            If specified a `~regions.PointSkyRegion` is converted
```
bkhelifi:
```suggestion
            If specified, a `~regions.PointSkyRegion` is converted
```

gammapy/maps/wcs/geom.py
```diff
@@ -970,6 +970,10 @@ def region_mask(self, regions, inside=True):
...
+            to a `~regions.CircleSkyRegion` of radius `point_to_radius`
+            while creating the mask.
```
bkhelifi, line 976:
```suggestion
            while creating the mask. Default is None.
```

bkhelifi (approved):
This is a very usefull small feature. Thanks. Good for me

registerrier (commented):
Thanks @AtreyeeS . 

A general comment here, what is the good behavior regarding circle regions in this case?  
You could expect that all radii are at least a given minimal value or that you take something like sqrt(radius**2+min_radius**2)

I would suggest to have the function as a utility function in `gammapy.utils.regions` as I find cleaner to explicit build a clean exclusion region list before creating a mask.

Tobychev:
The code change itself looks fine, but your motivating example to me strongly implies the need for a helper function like `make_sky_region_mask` intended to live at the `dataset` level, either as a separate helper or as a method on the dataset itself. 

One concrete advantage of working at the dataset level is that it allows the function to optionally select the radius using the psf directly.

Tobychev (approved):

gammapy/maps/wcs/tests/test_geom.py
```diff
@@ -441,6 +442,9 @@ def test_region_mask():
     mask = geom.region_mask(regions, inside=False)
     assert np.sum(mask.data) == 8
 
+    mask = geom.region_mask(regions, inside=False, point_to_radius=0.5 * u.deg)
```
QRemy, line 445: In my opinion we don't need the `point_to_radius` option but `region_mask` should handle `PointSkyRegion` correctly by putting a True pixel at the position of the source instead of nothing right now (which can be seen as a bug). Then the addition of the exclusion radius can be performed as for the other regions with a binary dilate on the mask:
```
exclusion_radius = np.median(obs.psf.containment_radius(fraction=0.99,
      energy_true=obs0.psf.axes["energy_true"].center,
      offset=0*u.deg))
mask = mask.binary_dilate(width=exclusion_radius)      
```

AtreyeeS: Thanks @QRemy ! This is a much cleaner solution.
What shall we do if the position lies on the edge of a pixel? Assign it to both the pixels?

QRemy: Maybe evaluate the corresponding point spatial model on the geom and keep the 4 pixels where it has values ?
https://docs.gammapy.org/dev/user-guide/model-gallery/spatial/plot_point.html#point-spatial-model

QRemy:
I'm going to close this. A separate PR should be open as a bugfix see:
https://github.com/gammapy/gammapy/pull/5976#discussion_r2220009271

Issue #5691, opened 2025-02-03
Upper limit computation for cases with negative fluctuations in flux (or significance)


juanjq:
The upper limit computation is not differentiating when the computed flux is negative or when is positive. Therefore this results in negative upper limits. The feature is mentioned in the function description but the method is not applying any action when flux is negative [link](https://github.com/gammapy/gammapy/blob/b878233e2719dbeeaefba6a45644a55b48a78634/gammapy/stats/counts_statistic.py#L163C18-L163C21).

As mentioned in the function (compute_upper_limit) help text:

> Searches the signal value for which the test statistics is n_sigma**2 away from the maximum or from 0 if the measured excess is negative.

The last part is not being applied so in my understanding, 
`min_range = np.where(self.n_sig < 0.0, 0.0, self.n_sig)`
Should do the work, as in these cases, the reference point will be 0, instead of the negative significance point.

AtreyeeS:
Hello @juanjq ! Which version of Gammapy are you using? What is the best fit amplitude of your spectral model? Did you constrain it to be above 0?
When doing the FluxPoint extraction, did you configure the `norm` parameter of the estimator to set `norm.min = 0.0`?

AtreyeeS:
Thanks! Do you have a `MapDataset` or a `MapDatasetOnOff`?

juanjq:
I have a MapDataset, of stacked observations (6 runs in total). I don't have OFF region so I only have MapDataset.

AtreyeeS:
What should we do here @registerrier @QRemy ?

registerrier:
This is a documentation issue that was introduced after #3519 which changed the default behavior to use the actual number of excess (possibly negative) as the reference value to compute the UL from.

The first thing is to adapt the docstring to make it consistent with what is done. 

We could introduce an option to only consider positive upper limits. But we should see how to do this in a clean manner.

bkhelifi:
@registerrier, @AtreyeeS and @QRemy : should we close it?

AtreyeeS:
The docs and code is consistent now. It was kept open as a reminder issue for upper limit calculations. Its a question if we should impose positive flux upper limits in the ExcessMapEstimator.

PR #5977, opened 2025-07-09
Update docs of counts_statistic.py

AtreyeeS:
Update docs to clarify #5691

### DIFF
--- a/gammapy/stats/counts_statistic.py
+++ b/gammapy/stats/counts_statistic.py
@@ -160,9 +160,8 @@ class CountsStatistic(abc.ABC):
     def compute_upper_limit(self, n_sigma=3):
         """Compute upper limit on the signal.
 
-        Searches the signal value for which the test statistics is n_sigma**2 away from the maximum
-        or from 0 if the measured excess is negative.
-
+        Searches the signal value for which the test statistics is n_sigma**2 away from the maximum.
+        
         Parameters
         ----------
         n_sigma : float


gammapy/stats/counts_statistic.py
```diff
@@ -160,9 +160,8 @@ def compute_errp(self, n_sigma=1):
     def compute_upper_limit(self, n_sigma=3):
         """Compute upper limit on the signal.
 
-        Searches the signal value for which the test statistics is n_sigma**2 away from the maximum
-        or from 0 if the measured excess is negative.
-
+        Searches the signal value for which the test statistics is n_sigma**2 away from the maximum.
```
AtreyeeS:
```suggestion
        Searches the signal value for which the test statistic is n_sigma**2 
        away from the maximum.
```

bkhelifi:
Why is it a clarification? You have removed the case where we have a negative excess

AtreyeeS:
> Why is it a clarification? You have removed the case where we have a negative excess

@bkhelifi Please see the discussion in #5691 
By design choice do not do a different handling for negative excess

bkhelifi:
Ah yes, indeed... the key line is `self._stat_fcn(min_range[it.multi_index], 0.0, it.multi_index)`... mu=0 in the stat function, and whatever the sign of the excess, DeltaTs is computed.
Fine with me.

registerrier:
Note that here changing the code directly from the interface causes issues with the formatter.  
In this case, it is not sure this would have been solved as the pre-commit skips this file. We should check the correct approach.

gammapy/stats/counts_statistic.py
```diff
@@ -160,9 +160,9 @@ def compute_errp(self, n_sigma=1):
     def compute_upper_limit(self, n_sigma=3):
         """Compute upper limit on the signal.
 
-        Searches the signal value for which the test statistics is n_sigma**2 away from the maximum
-        or from 0 if the measured excess is negative.
-
+        Searches the signal value for which the test statistic is n_sigma**2 
+        away from the maximum.
+
```
QRemy:
```suggestion


```

gammapy/stats/counts_statistic.py
```diff
@@ -160,8 +160,9 @@ def compute_errp(self, n_sigma=1):
     def compute_upper_limit(self, n_sigma=3):
         """Compute upper limit on the signal.
 
-        Searches the signal value for which the test statistics is n_sigma**2 away from the maximum
-        or from 0 if the measured excess is negative.
+        Searches the signal value for which the test statistic is n_sigma**2 
+        away from the maximum.
+
```
QRemy:
```suggestion
        Searches the signal value for which the test statistic is n_sigma**2
        away from the maximum.

```

QRemy (approved):
Thanks @AtreyeeS, fixed codestyle merging now.

PR #5979, opened 2025-07-11
Fix memory leak related to lru_cache

### LINKED ISSUE #5706: Possible memory leak in FoVBackgroundMaker

QRemy:
Resolves #5706 using the code from https://github.com/python/cpython/issues/102618#issuecomment-2839489762

### DIFF
--- a/gammapy/utils/cache.py
+++ b/gammapy/utils/cache.py
@@ -1,0 +1,96 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+"""Utilities for caching"""
+
+# from https://github.com/python/cpython/issues/102618#issuecomment-2839489762.
+import functools as ft
+import inspect
+import weakref
+from collections.abc import Callable
+from typing import Concatenate, ParamSpec, TypeVar
+
+_Self = TypeVar("_Self")
+_Params = ParamSpec("_Params")
+_Return = TypeVar("_Return")
+
+
+# Like `weakref.WeakKeyDictionary`, but uses identity-based hashing and equality.
+class _WeakIdDict:
+    __slots__ = ("_dict",)
+
+    def __init__(self):
+        self._dict = {}
+
+    def __getitem__(self, key):
+        item, _ = self._dict[id(key)]
+        return item
+
+    def __setitem__(self, key, value):
+        id_key = id(key)
+        ref = weakref.ref(key, lambda _: self._dict.pop(id_key))
+        self._dict[id_key] = (value, ref)
+
+
+# Used to compute an object's hash just once.
+class _CacheKey:
+    __slots__ = ("hashvalue", "value")
+
+    def __init__(self, value):
+        self.hashvalue = hash(value)
+        self.value = value
+
+    def __hash__(self) -> int:
+        return self.hashvalue
+
+    def __eq__(self, other) -> bool:
+        # Assume `type(other) is _Key`
+        return self.value == other.value
+
+
+def cachemethod(
+    fn: Callable[Concatenate[_Self, _Params], _Return],
+) -> Callable[Concatenate[_Self, _Params], _Return]:
+    """Like `functools.cache`, except that it only holds a weak reference to its first argument.
+
+    Note that `functools.cached_property` (which also uses a weak reference) can often be used for similar purposes.
+    The differences are that (a) `cached_property` will be pickled whilst `cachemethod` will not, and (b) `cachemethod`
+    can be used on functions with additional arguments, and (c) `cachemethod` requires brackets to call, helping to
+    visually emphasise that computationl work may be being performed.
+    """
+    cache1 = _WeakIdDict()
+    sig = inspect.signature(fn)
+    parameters = list(sig.parameters.values())
+    if len(parameters) == 0:
+        raise ValueError(
+            "Cannot use `cachemethod` on a function without a `self` argument."
+        )
+    if parameters[0].kind not in {
+        inspect.Parameter.POSITIONAL_ONLY,
+        inspect.Parameter.POSITIONAL_OR_KEYWORD,
+    }:
+        raise ValueError(
+            "Cannot use `cachemethod` on a function without a positional argument."
+        )
+    parameters = parameters[1:]
+    sig = sig.replace(parameters=parameters)
+
+    @ft.wraps(fn)
+    def fn_wrapped(
+        self: _Self, *args: _Params.args, **kwargs: _Params.kwargs
+    ) -> _Return:
+        # Standardise arguments to a single form to encourage cache hits.
+        # Not binding `self` (and instead doing the signature-adjustment above) in order to avoid keeping a strong
+        # reference to `self` via `argkey`.
+        bound = sig.bind(*args, **kwargs)
+        del args, kwargs
+        argkey = _CacheKey((bound.args, tuple(bound.kwargs.items())))
+        try:
+            out = cache1[self][argkey]
+        except KeyError:
+            try:
+                cache2 = cache1[self]
+            except KeyError:
+                cache2 = cache1[self] = {}
+            out = cache2[argkey] = fn(self, *bound.args, **bound.kwargs)
+        return out
+
+    return fn_wrapped
--- a/gammapy/utils/cache.py
+++ b/gammapy/utils/cache.py
@@ -1,19 +1,53 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 """Utilities for caching"""
 
-# from https://github.com/python/cpython/issues/102618#issuecomment-2839489762.
 import functools
 import inspect
+import hashlib
 import weakref
-from typing import ParamSpec, TypeVar
+from gammapy.utils.parallel import is_ray_available
 
-_Self = TypeVar("_Self")
-_Params = ParamSpec("_Params")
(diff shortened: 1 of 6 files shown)


QRemy:
Here is an example

```
import astropy.units as u
import numpy as np
import gc
import pickle

dataset = MapDataset.read("/Users/qremy/Downloads/test1.fits")


N=10
for i in range(N):
    print(i)
    ds = MapDataset.read("/Users/qremy/Downloads/test1.fits")  
    #ds.mask_fit = dataset._geom.region_mask([reg], inside=False) 
    #gc.collect()
    
np.savez("/Users/qremy/Downloads/testpk.npz", ds=ds)

ds_loaded = np.load("/Users/qremy/Downloads/testpk.npz", allow_pickle=True)["ds"].item()
ds_loaded._geom.get_coord()

```

Main without gc.collect()

main

Main with gc.collect()
main_with_gccollect

Fix without gc.collect()
fix

AtreyeeS (dismissed):
Thank you very much @QRemy ! This does seem to solve the memory leak, but I dont understand why the memory usage is high to begin with. For example, testing on

```
import numpy as np
import astropy.units as u
dataset = MapDataset.read(
    "test1.fits.gz"
)
reg = CircleSkyRegion(center=dataset._geom.center_skydir, radius=1.0*u.deg)
N = 50

for i in range(N):
    if i%10 == 0:
        print('Running for ',i, '/',N)
     
    ds = MapDataset.read("test1.fits.gz")  
```

On main:
plot_main


On this branch:
plot_qr


I dont see any loss of performance in the analysis from the tests that I did, so I propose we merge this now and see if there is any effect in the benchmarks.

adonath:
Thanks @QRemy, this looks good to me! From what I know about the issue, the proposed code is what is needed to make it work with a universal decorator. 

The alternative solution would be to just change to a "stupid" caching pattern where the cache is part of the instance:

```python
class MyClass:
    def __init__(self, ...):
        self._cache = {}

    def method(self, *args):
        value = self._cache.get(args, None)
        if value is not None:
            return value
        value = ... # compute value
        self._cache[args] = value
        return value
```
This avoids the back-reference introduced by self. Would this works as well?

QRemy:
> The alternative solution would be to just change to a "stupid" caching pattern where the cache is part of the instance:

The order of the arguments might be a problem.

I wonder as simple alternative solution if we could just give up an this caching with args and cache the coord of the full geom directly. Then we could also propagate the cache to the cutout as simple views and that will save even more memory in the MapEvaluator.

adonath:
> The order of the arguments might be a problem.

Yes, this makes the book-keeping complex. 

> I wonder as simple alternative solution if we could just give up an this caching with args and cache the coord of the full geom directly. Then we could also propagate the cache to the cutout as simple views and that will save even more memory in the MapEvaluator.

Yes, this would be nice to have. But again the different args make the book-keeping complex.

adonath:
From the dev meeting today: the proposed solution is complex but it solves the problem and is general (can be used in other parts of Gammapy) and I think there is a high chance we never have to touch that code again. So let's keep the solution.

gammapy/utils/cache.py
```diff
@@ -0,0 +1,94 @@
...
+        inspect.Parameter.POSITIONAL_ONLY,
+        inspect.Parameter.POSITIONAL_OR_KEYWORD,
+    }:
+        raise ValueError(
+            "Cannot use `cachemethod` on a function without a positional argument."
+        )
+    parameters = parameters[1:]
+    sig = sig.replace(parameters=parameters)
+
+    @ft.wraps(fn)
```
registerrier: for clarity, I would explicitly call `@functools.wraps(fn)`

gammapy/utils/cache.py
```diff
@@ -0,0 +1,94 @@
...
+    def fn_wrapped(self, *args, **kwargs):
+        # Standardise arguments to a single form to encourage cache hits.
+        # Not binding `self` (and instead doing the signature-adjustment above) in order to avoid keeping a strong
+        # reference to `self` via `argkey`.
+        bound = sig.bind(*args, **kwargs)
+        del args, kwargs
+        argkey = _CacheKey((bound.args, tuple(bound.kwargs.items())))
```
registerrier: As far as I understand this will try create to a weak reference based on the __hash__ of the argument. But what if we pass a numpy array or a list? They are not hashable, right?

Some of the methods this should cache have arguments that can be passed as list or arrays (e.g. `idx` arguments).

gammapy/utils/cache.py
```diff
@@ -0,0 +1,94 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+"""Utilities for caching"""
+
+# from https://github.com/python/cpython/issues/102618#issuecomment-2839489762.
+import functools as ft
```
registerrier:
```suggestion
import functools
```
clearer this way

registerrier (dismissed):

On question though, this requires that arguments are all hashable, which means that certain types of arguments cannot be used e.g. lists or numpy arrays. Some of the functions decorated can take such arguments (e.g. `idx` for `get_coord`) can this be a problem?
Now, it might well be that this is already a problem with the regular `lru_cache` itself.
With the current version, `geom.get_coord(idx=(1,))` works, while `geom.get_coord(idx=[1,])` fails

gammapy/utils/cache.py
QRemy:
```suggestion
    @ functools.wraps(fn)
```

gammapy/utils/cache.py
QRemy, line 5:
```suggestion
import functools
```

gammapy/utils/cache.py
QRemy: yes, idx should be a tuple, it's already the case with the lru_cache.

QRemy: an alternative could be to use ray.cloudpickle to make the cache key as in #5982 :
```python
def make_key(sig, *args, **kwargs):
    bound = sig.bind_partial(None, *args, **kwargs)
    bound.apply_defaults()
    normalized_args = dict(sorted(bound.arguments.items()))
    normalized_args.pop("self", None)
    try:
        return hash(normalized_args)
    except TypeError:
        data = ray.cloudpickle.dumps(normalized_args)
        return hashlib.sha256(data).hexdigest()

class _WeakIdDict:
    """Like `weakref.WeakKeyDictionary`, but uses identity-based hashing and equality."""

    __slots__ = ("_dict",)

    def __init__(self):
        self._dict = {}

    def __getitem__(self, key):
        item, _ = self._dict[id(key)]
        return item

    def __setitem__(self, key, value):
        id_key = id(key)
        ref = weakref.ref(key, lambda _: self._dict.pop(id_key))
        self._dict[id_key] = (value, ref)
        
def cachemethod(fn):
    
    cache1 = _WeakIdDict()
    sig = inspect.signature(fn)

    @functools.wraps(fn)
    def wrapper(self, *args, **kwargs):
        argkey = make_key(sig, *args, **kwargs)
        try:
            out = cache1[self][argkey]
        except KeyError:
            try:
                cache2 = cache1[self]
            except KeyError:
                cache2 = cache1[self] = {}
            out = cache2[argkey] = fn(self, *args, **kwargs)
        return out
    return wrapper
```

QRemy: I implemented this version

AtreyeeS (approved):
This works, so good for me

PR #5982, opened 2025-07-14
Add class instance caching for WcsGeom and HpxGeom

QRemy:
Add class instance caching for WcsGeom and HpxGeom.
This avoid in memory duplication of geoms and their associated coordinates caching.
Requires #5979.

### DIFF
--- a/gammapy/utils/cache.py
+++ b/gammapy/utils/cache.py
@@ -1,0 +1,30 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+"""Utilities for caching"""
+
+import weakref
+
+
+class CacheEquivalentMixin:
+    """Cache class instance using equality condition.
+    Simpler for classes with complex arguements that are not hashable.
+    """
+
+    _instances = []
+
+    def __new__(cls, *args, **kwargs):
+        # Clean up dead references
+        cls._instances = [ref for ref in cls._instances if ref() is not None]
+
+        # Create a temporary instance to compare
+        temp = super().__new__(cls)
+        temp.__init__(*args, **kwargs)
+
+        # Search for an equivalent instance
+        for ref in cls._instances:
+            instance = ref()
+            if instance is not None and instance == temp:
+                return instance
+
+        # Store a weak reference to the new instance
+        cls._instances.append(weakref.ref(temp))
+        return temp
--- a/gammapy/maps/hpx/geom.py
+++ b/gammapy/maps/hpx/geom.py
@@ -8,6 +8,7 @@ from astropy.coordinates import SkyCoord
 from astropy.io import fits
 from astropy.units import Quantity
 from gammapy.utils.array import is_power2
+from gammapy.utils.cache import CacheEquivalentMixin
 from ..axes import MapAxes
 from ..coord import MapCoord, skycoord_to_lonlat
 from ..geom import Geom, pix_tuple_to_idx
@@ -31,7 +32,7 @@ from .utils import (
 __all__ = ["HpxGeom"]
 
 
-class HpxGeom(Geom):
+class HpxGeom(Geom, CacheEquivalentMixin):
     """Geometry class for HEALPix maps.
 
     This class performs mapping between partial-sky indices (pixel
--- a/gammapy/maps/wcs/geom.py
+++ b/gammapy/maps/wcs/geom.py
@@ -1,5 +1,6 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 import copy
+from gammapy.utils.cache import CacheEquivalentMixin
 from functools import lru_cache
 import numpy as np
 import astropy.units as u
@@ -60,7 +61,7 @@ def get_resampled_wcs(wcs, factor, downsampled):
     return wcs
 
 
-class WcsGeom(Geom):
+class WcsGeom(Geom, CacheEquivalentMixin):
     """Geometry class for WCS maps.
 
     This class encapsulates both the WCS transformation object and
--- a/gammapy/maps/wcs/tests/test_geom.py
+++ b/gammapy/maps/wcs/tests/test_geom.py
@@ -687,3 +687,19 @@ def test_wcs_geom_with_timeaxis():
         [[4.000e-01, 2.000e-01, 0.000e00, 3.598e02, 3.596e02]],
         rtol=1e-5,
     )
+
+
+def test_instance_cache():
+    kwargs = dict(skydir=(83.63, 22.01), width=5, binsz=0.02)
+
+    geom1 = WcsGeom.create(**kwargs)
+    geom2 = WcsGeom.create(**kwargs)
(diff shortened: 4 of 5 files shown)


QRemy:
Example:

```
import astropy.units as u
import gc

dataset = MapDataset.read("/Users/qremy/Downloads/test1.fits")


N=10
for i in range(N):
    print(i)
    ds = MapDataset.read("/Users/qremy/Downloads/test1.fits")  
    #ds.mask_fit = dataset._geom.region_mask([reg], inside=False) 
    gc.collect()
 
```

Main with gc.collect
main_with_gccollect_

This PR
branch

QRemy:
The caching is ignored during  unpickle, but I think it ok as it's used only for a parallelisation with multiprocessing.

adonath:
> The caching is ignored during unpickle, but I think it ok as it's used only for a parallelisation with multiprocessing.

Yes, this is fine. Coords can be re-computed in the new process.

gammapy/utils/cache.py
```diff
@@ -0,0 +1,33 @@
...
+
+import weakref
+
+
+class CacheEquivalentMixin:
+    """Cache class instance using equality condition.
+    Simpler for classes with complex arguements that are not hashable.
+    """
+
+    _instances = []
```
adonath: You can use a set here, there is no meaning to the order of the instances.

adonath:
Thanks @QRemy! Just to understand: this PR introduces the behavior to return an already existing geom, when one of the same geometry (defined by `__eq__`) is found.

I think this behavior should be fine, as we never rely on the geom being a unique object. However we should double check, that we never have any in-place modifications of geom objects.

Relying on a hash would be more efficient I think, but would require the effort to define a meaningful one for geometries, which is not the case I think: https://github.com/gammapy/gammapy/blob/9edee9ec0a03121c7d8148a3d79abdbda6f3c5d3/gammapy/maps/wcs/geom.py#L1223

adonath:
However this use case would be a fair justification to implement a meaningful `__hash__`, because
> The only required property is that objects which compare equal have the same hash value;

Taken from: https://docs.python.org/3.5/reference/datamodel.html#object.__hash__

Which is not the case currently for us.

QRemy:
> Relying on a hash would be more efficient I think

I tried before but turning all the non-hashable arguments of the init into something hashable was quite complicated.
At the end I opted for this solution which is simpler even if less efficient.
But I don't think it's a problem because the creation of the geom is fast and light in memory so it's not really the bottleneck

QRemy:
> we should double check, that we never have any in-place modifications of geom objects.

could we make them immutable ?

adonath:
From the dev meeting today: in principle `Geom` objects are designed to not be modified in place, but because it is Python we cannot really prevent this. So this proposed change carries the risk of breaking at some point, because a geom object is modified in place. The performance improvement is currently too small to justify this yet.

Also: the proper way for the caching would be to implement a meaningful `__hash__` and use a dict. However we currently do not have a good idea of what makes two geom objects the same. We a proposal have implemented but this relies on numerical precision, which might no even be high enough for people in x-ray astronomy (as Regis noted).

So let's put this on hold and think a bit more about the `__hash__` first and find the use cases, where the geom caching makes the largest difference.

QRemy:
One of the use case is that the de-serialization does not preserve memory usage as each dataset will create new geoms. For  joint analyses with many datasets sharing the same geom this can be a problem. For example with event types there will be several datasets repeating the same cache (typically 10 with HAWC/SWGO).

QRemy:
The new version uses a dict and a hashing of the init arguments.

new_version

gammapy/utils/cache.py
```diff
@@ -0,0 +1,39 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+"""Utilities for caching"""
+
+import cloudpickle
```
QRemy: could use `ray.cloudpickle` instead to avoid extra dependancy

gammapy/utils/cache.py
```diff
@@ -0,0 +1,39 @@
...
+
+    # Serialize and hash compatible with pickle
+    data = cloudpickle.dumps(normalized_args)
+    return hashlib.sha256(data).hexdigest()
+
+
+class CacheEquivalentMixin:
+    """Cache class instance"""
+
+    def __new__(cls, *args, **kwargs):
```
QRemy, line 128: could activate with global property and if ray is available
 
```suggestion
    def __new__(cls, *args, **kwargs):
    
    if not USE_GEOM_CACHE and not is_ray_available():
            return super().__new__(cls)
```

QRemy:
Now it shares the `make_key` mecanism with `cachemethod` implemented in #5979,
so the addition of  the `CacheInstanceMixin` is very minimal.
It can be activate via a global variable `USE_INSTANCE_CACHE`, for no now it is disabled by default or if ray is not available (it uses ray implementation of cloudpickle to avoid extra dependancy).

QRemy:
I removed it from `HpxGeom` because we don't have caching for `get_coord` and `get_idx` there (not sure why).

AtreyeeS:
I dont understand this PR fully, do you want this go in for 2.0 ?
Otherwise we can try a few benchmarks to ensure that this actually improves the performance

QRemy:
Yes I think this should go in for 2.0, by default the new caching behaviour is deactivated so that won't cause unexpected issues, but we will be able to activate it to improve performance of large scale joint analyses.

In the simple benchmark above you can see it already reduce memory usage by 20%.

QRemy:
This PR can be summarise as : Sometimes you need to cache more to cache less.

bkhelifi (approved):
Thank you @QRemy . Let see later when it will be useful

bkhelifi:
@QRemy  and @AtreyeeS : is our benchmark fine enough to track in the future mem issue? I am wondering whether one could have in the future the same type of problem without noticing it.
@QRemy : can the code you give here be used for the benchmarking?

QRemy:
@QRemy : can the code you give here be used for the benchmarking?

https://github.com/gammapy/gammapy-benchmarks/blob/main/benchmarks/profilers/run_memray.py

QRemy:
> @QRemy : can the code you give here be used for the benchmarking?

https://github.com/gammapy/gammapy-benchmarks/blob/main/benchmarks/profilers/run_memray.py

Just adapted to run the little example given at the beginning.

AtreyeeS:
> is our benchmark fine enough to track in the future mem issue? I am wondering whether one could have in the future the same type of problem without noticing it.

@bkhelifi : no your benchmarks are insufficient. We have to create better benchmarks. There is an open issue https://github.com/gammapy/gammapy-benchmarks/issues/171

QRemy:
By default the new caching behaviour is deactivated so that won't cause unexpected issues.
So I'm going to merge this for now and we will see later if we need to activate it by default.

PR #5984, opened 2025-07-16
Coherent checksum behavior in the IO function of `Datasets`

bkhelifi:
**Description**

This pull request fixes the default of the checksum argument in the `write` function to be in line with the `read` function.


QRemy (approved):
Thanks @bkhelifi, no comment merging now.

Issue #5991, opened 2025-07-18
Fix aspect ratio of IRF peek plot


morcuended:
The aspect ratio of this plot looks weird.

<img width="668" height="450" alt="image" src="https://github.com/user-attachments/assets/a340b8a9-5ee1-4f6a-9747-2744d1bf9cdf" />

_Originally posted by @morcuended in https://github.com/gammapy/gammapy/pull/5918#discussion_r2213577299_

bkhelifi:
I am preparing a PR for that... almost ready;)

samanthalwong:
Thanks for helping with this so quickly!

AtreyeeS:
Fixed in #5992

PR #5992, opened 2025-07-18
Update the peek function

bkhelifi:
**Description**

This pull request is associated to the issue #5991. 
The `observtion.peek()` is updated to forced to have square plots. For that, it was needed to update the `utils.add_colorbar()` function, to have a height equal to the plot one, excluding the title height.

**Dear reviewer**
The outputs of the change are here:
Figure_1
for the VERITAS dataset,

and 
Figure_2

for the usual output for a standard data production

### DIFF
--- a/examples/tutorials/api/irfs.py
+++ b/examples/tutorials/api/irfs.py
@@ -226,7 +226,8 @@ print(res)
 # sphinx_gallery_thumbnail_number = 1
 aeff_eval = aeff_3d.evaluate(energy_true=[1.0] * u.TeV)
 
-ax = plt.subplots(1, 1, figsize=(9, 9))
+plt.figure(figsize=(9, 9))
+ax = plt.gca()
 with quantity_support():
     caxes = ax.pcolormesh(
         fov_lat_axis.edges, fov_lon_axis.edges, aeff_eval.value.squeeze()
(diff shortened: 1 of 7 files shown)


examples/tutorials/starting/analysis_2.py
```diff
@@ -138,6 +138,13 @@
 
 observations = data_store.get_observations(selected_obs_table["OBS_ID"])
 
+######################################################################
+# We can have a quick look on the content of each observation with
```
samanthalwong:
```suggestion
# We can have a quick look at the content of each observation with
```

samanthalwong (approved):
Looks a lot better for the VERITAS plots, thanks for this!

examples/tutorials/api/irfs.py
```diff
@@ -226,7 +226,8 @@ class EffectiveArea3D(IRF):
 # sphinx_gallery_thumbnail_number = 1
```
morcuended, line 226: There is a plot that does not place properly the color bar:

image

QRemy (approved):
Thanks @bkhelifi, no comment from my side.

examples/tutorials/api/irfs.py
AtreyeeS: oops! looks like we merged too quickly!

bkhelifi: I wil check that!!!
Thanks @morcuended

PR #5993, opened 2025-07-18
Fix tutorial page LHS dropdown

Astro-Kirsty:
This fixes the dropdown on the tutorial page LHS.

Thanks to @morcuended for noticing that


registerrier (approved):

PR #5995, opened 2025-07-19
Update Fermi-LAT tutorial

### LINKED ISSUE #5980: Improve the Fermi-Lat notebook
We should improve the Fermi-Lat notebook before the 2.0 release, showing in particular changes introduced in #5649, and references therein. Some effort was started in #5675, but was postponed for later.

The tutorial can show the 

- [ ]`FermipyDatasetsReader` (with and without a fermipy config file)
- [ ] Setting isotropic diffuse model
- [ ] Creating a yaml models list from the Fermi catalog
- [ ] One can also explain the Gammapy does not allow to have an exposure map on a larger region than the counts map, and how this can be handled with masking
- [ ] How to enable weighted likelihood

QRemy:
Update Fermi-LAT tutorial to show the new datasets reader and a workflow well-suited for Fermi-LAT analyses.

Requires https://github.com/gammapy/gammapy-data/pull/76

Resolves #5980

### DIFF
--- a/examples/tutorials/data/fermi_lat.py
+++ b/examples/tutorials/data/fermi_lat.py
@@ -540,32 +540,22 @@ ts_estimator = TSMapEstimator(
 
 ts_results = ts_estimator.run(datasets)
 
-import matplotlib as mpl
 
+image = ts_results["sqrt_ts"]
+image = image.cutout(
+    image.geom.center_skydir, width=np.max(image.geom.width) - 2 * margin
+)
 
-def plot_significance_map(estimator_results, source_models=None, margin=None):
-    fig = plt.figure(figsize=(7, 5))
-    image = estimator_results["sqrt_ts"]
-    if margin is not None:
-        image = image.cutout(
-            image.geom.center_skydir, width=np.max(image.geom.width) - 2 * margin
-        )
-    ax = image.plot(clim=[-8, 8], cmap=plt.cm.RdBu_r)
-    im = [obj for obj in ax.get_children() if isinstance(obj, mpl.image.AxesImage)][0]
-    if source_models:
-        source_models.plot_regions(
-            ax=ax, edgecolor="g", linestyle="-", kwargs_point=dict(marker=".")
-        )
-    plt.title(r"$\sqrt{TS}$", loc="left", fontsize=12, pad=10)
-    cb_ax = fig.add_axes([0.5, 0.91, 0.15, 0.04])
-    cb = fig.colorbar(im, cax=cb_ax, orientation="horizontal")
-    cb.ax.xaxis.set_ticks_position("top")
-    cb.ax.xaxis.set_label_position("top")
-    cb.ax.tick_params(axis="both", which="major", labelsize=12)
-    cb.ax.set_xlabel(r"[$\sigma$]", fontsize=12, labelpad=-28, x=-0.1)
-
-
-plot_significance_map(ts_results, source_models=sources_inside_roi, margin=margin)
+fig = plt.figure(figsize=(7, 5))
+ax = image.plot(
+    clim=[-8, 8],
+    cmap=plt.cm.RdBu_r,
+    add_cbar=True,
+    kwargs_colorbar={"label": r"$\sqrt{TS}$ [$\sigma$]"},
+)
+sources_inside_roi.plot_regions(
+    ax=ax, edgecolor="g", linestyle="-", kwargs_point=dict(marker=".")
+)
 plt.show()
 
 
--- a/examples/tutorials/data/fermi_lat.py
+++ b/examples/tutorials/data/fermi_lat.py
@@ -59,7 +59,7 @@ from gammapy.utils.scripts import make_path
 
 ######################################################################
 # Check setup
-# ^^^^^^^^^^^
+# ~~~~~~~~~~~
 #
 # We check the setup in this tutorial, as we require specific files to be
 # downloaded to continue.
--- a/examples/tutorials/data/fermi_lat.py
+++ b/examples/tutorials/data/fermi_lat.py
@@ -154,7 +154,7 @@ check_tutorials_setup()
 # - If you want to use the `$FERMI_DIR` variable to read the background models
 #   it must also be defined in your Gammapy environment,
 #   otherwise you have to define your own paths.
-# - For this tutorial we copied the iso files in $GAMMAPY_DATA/fermi-gc` and
+# - For this tutorial we copied the iso files in `$GAMMAPY_DATA/fermi-gc` and
 #   edited the paths in the yaml file for simplicity.
 #
 # More generally in order to select a good binning it is important to know
--- a/examples/tutorials/data/fermi_lat.py
+++ b/examples/tutorials/data/fermi_lat.py
@@ -9,9 +9,9 @@ Introduction
 
 Gammapy fully supports Fermi-LAT data analysis from DL4 level (binned
 maps). In order to perform data reduction from the events list and
-spacecrast files to binned counts and IRFs maps we recommend to use
-`Fermipy <http://fermipy.readthedocs.io/>`__\ *, which is based on
-the*\ `Fermi Science
+spacecraft files to binned counts and IRFs maps we recommend to use
+`Fermipy <http://fermipy.readthedocs.io/>`__, which is based on
+the `Fermi Science
 Tools <https://fermi.gsfc.nasa.gov/ssc/data/analysis/software/>`__
 (Fermi ST).
 
@@ -59,25 +59,26 @@ from gammapy.utils.scripts import make_path
 
 ######################################################################
 # Check setup
-# -----------
+# ^^^^^^^^^^^
 #
+# We check the setup in this tutorial, as we require specific files to be
+# downloaded to continue.
 
-# %%capture
 from gammapy.utils.check import check_tutorials_setup
 
 check_tutorials_setup()
 
 
 ######################################################################
-# Fermipy congifuration file
+# Fermipy configuration file
 # --------------------------
 #
-# Gammapy can re-use the same configuration file than Fermipy to convert
-# the Fermipy maps to Gammapy datasets. For details on these files
-# struture see the `Fermipy configuration
+# Gammapy can utilise the same configuration file as Fermipy to convert
+# the Fermipy-generated maps into Gammapy datasets. For more information on the
+# structure of these files, refer to the `Fermipy configuration
 # page <https://fermipy.readthedocs.io/en/latest/config.html>`__. In this
-# tutorial we will analyse Galactic center data generated with Fermipy version 1.3 and the
-# configuration given in
+# tutorial, we will analyse Galactic center data generated with Fermipy version 1.3 and
+# the configuration given in
 # `$GAMMAPY_DATA/fermi-gc/config_fermipy_gc_example.yaml`:
 #
 
@@ -142,44 +143,44 @@ check_tutorials_setup()
 
 
 ######################################################################
-# | The most important points for Gammapy users are:
-# - `emin`/`emax`in this file should be considered as the energy true range.
-#   It should be larger that the recosntructed energy range.
-# - `edisp_bins : 0` is strongly recommended at this stage otherwise you
-#   might face inconsitancies in the energy axes of the different IRFs created by Fermipy.
-# | The `edisp_bins` value will be redefined later on by Gammapy as a positive value
+# The most important points for Gammapy users are:
+#
+# - ``emin`` and ``emax`` in this file should be considered as the energy true range.
+#   It should be larger that the reconstructed energy range.
+# - ``edisp_bins : 0`` is strongly recommended at this stage otherwise you
+#   might face inconsistencies in the energy axes of the different IRFs created by Fermipy.
+# - The ``edisp_bins`` value will be redefined later on by Gammapy as a positive value
 #   in order to create the reconstructed energy axis properly.
 # - If you want to use the `$FERMI_DIR` variable to read the background models
-#   it must also be defined in your Gammapy environemnt,
+#   it must also be defined in your Gammapy environment,
 #   otherwise you have to define your own paths.
-# | For this tutorial we copied the iso files in
-#   `$GAMMAPY_DATA/fermi-gc` and edited the paths in the yaml file for
-#   simplicity.
+# - For this tutorial we copied the iso files in $GAMMAPY_DATA/fermi-gc` and
+#   edited the paths in the yaml file for simplicity.
 #
 # More generally in order to select a good binning it is important to know
 # the instrument resolution, for that you can have a quick look at the
-# irfs in the `Fermi-LAT performamce
+# IRFs in the `Fermi-LAT performance
 # page <https://www.slac.stanford.edu/exp/glast/groups/canda/lat_Performance.htm>`__.
 #
 # Since the energy resolution varies with energy, it is important to
 # choose an energy binning that is fine enough to capture this energy
 # dependence. That is why we recommend a binning with 8 to 10 bins per
 # decade. The energy axes will be created such as it is linear in log space
-# so it's better to define `emin`/`emax` such as they align with a log binning.
-# Here we have as true energy range log(emin) = 0.6 ~ 4 GeV to $log(emax) = 3.4 ~ 2500 GeV.
+# so it's better to define ``emin`` and ``emax`` such as they align with a log binning.
+# Here we have as true energy range :math:`log(emin) = 0.6 ~ 4`GeV to
+# :math:`log(emax) = 3.4 ~ 2500`GeV.
 # While the reconstructed energy range of our analysis will be 10 GeV to 1000 GeV.
-
 #
 # The spatial binning should be of the same order of the PSF 68%
 # containment radius which is in average 0.1 degree above 10 GeV and
 # rapidly increases at lower energy. Ideally it should remain within a
 # factor of 2 or 3 of the PSF radius at most. In order to properly take
-# into account for the sources outside of the region of interest that
-# contribute inside due to the PSF we have to define a wider `roiwidth`
+# into account for the sources outside the region of interest that
+# contribute inside due to the PSF we have to define a wider ``roiwidth``
 # than our actual region of interest. Typically, we need a margin equal to the
 # 99% containment of the PSF on each side. Above 10 GeV considering only
 # PSF2&3 the 99% PSF containment radius is about 1 degree. Thus, if we
-# want to study a 3 degree radius around the GC we have to take a `roiwidth` of 8
+# want to study a 3 degree radius around the GC we have to take a ``roiwidth`` of 8
 # deg. (If considering lower energies or including PSF0 and PSF1, it should be much
 # larger).
 #
@@ -197,11 +198,12 @@ check_tutorials_setup()
 #    gta = GTAnalysis('config_fermipy_gc_example.yaml',logging={'verbosity' : 3})
 #    gta.setup()
 #
-#    gta.compute_psf(overwrite=True) # this create the psf kernel
-#    gta.compute_drm(edisp_bins=0, overwrite=True) # this create the energy dispersion matrix
+#    gta.compute_psf(overwrite=True) # this creates the psf kernel
+#    gta.compute_drm(edisp_bins=0, overwrite=True) # this creates the energy dispersion matrix
 #    # DO NOT CHANGE edisp_bins here, it will be redefined by Gammapy later on
 #
 # This will produce a number of files including:
+#
 # - “ccube_00.fits” (counts)
 # - “bexpmap_00.fits” (exposure)
 # - “psf_00.fits” (psf)
@@ -211,7 +213,7 @@ check_tutorials_setup()
 
 
 ######################################################################
-# Then is your Gammapy environment you can create the datasets using the
+# In your Gammapy environment you can create the datasets using the
 # same configuration file.
 #
 
@@ -224,39 +226,38 @@ print(datasets)
 
 
 ######################################################################
-# Note that the `edisp_bins` is set again here as a positive number so
-# Gammapy can create its reconstructed enerygy axis properly. The energy
+# Note that the ``edisp_bins`` is set again here as a positive number so
+# Gammapy can create its reconstructed energy axis properly. The energy
 # dispersion correction implemented Gammapy is closer to the version
 # implemented in Fermitools >1.2.0, which take into account the interplay
 # between the energy dispersion and PSF.
 #
-# For most of the Fermi energy range, the level of migration in log10(E)
-# due to energy dispersion is within 0.2 (when going below 100 MeV it
-# increases up to about 0.4). As a consequence we recommend that the
-# product of \|edisp_bins\| times the width of the log10(E) bins is at
-# least equal to 0.2. With a binning of 8 to 10 bins per decade, it
-# corresponds to \|edisp_bins\| ≥ 2. For more details see
+# Across most of the Fermi energy range, the level of migration in log10(E)
+# remains within 0.2, increasing up to 0.4 below 100 MeV,
+# due to energy dispersion. Therefore, we recommend that the
+# product of \|edisp_bins\| and the width of the log10(E) bins be at
+# least equal to 0.2. For a binning of 8 to 10 bins per decade, this
+# corresponds to \|edisp_bins\| ≥ 2. For further information, see
 # `Pass8_edisp_usage <https://fermi.gsfc.nasa.gov/ssc/data/analysis/documentation/Pass8_edisp_usage.html>`__.
-# In our case the we have 10 bins per decade and true energy axis starts
-# at about 4 GeV, so with `edisp_bins=4` the reconstructed enegry axis
+# In our case, we have 10 bins per decade and true energy axis starts
+# at about 4 GeV, so with ``edisp_bins=4`` the reconstructed energy axis
 # starts at 10 GeV:
 #
 
 print(datasets[0].exposure.geom.axes["energy_true"])
 print(datasets[0].counts.geom.axes["energy"])
 
-
 ######################################################################
-# Note that choosing instead `edisp_bins=2` means that the reconstructed energy
-# of the counts geom would have start  at $10^0.8$ ~ 6.3 GeV.
-# In that case if we want to start the analysis at 10 GeV we would have to
-# update the `mask_fit` to ignore the 2 first reconstructed energy bins.
-# Considering more `edisp_bins` is in general safer but consumes more memory
-# and will increase computation time.
+# Note that selecting ``edisp_bins=2`` means the reconstructed energy
+# of the counts geometry will start at :math:`10^{0.8} ~ 6.3`GeV.
+# If we want to start the analysis at 10 GeV in this case, we need to
+# update the ``mask_fit`` to exclude the first 2 reconstructed energy bins.
+# Considering more ``edisp_bins`` is generally safer but requires more memory
+# and increases computation time.
 #
-# Alternatively if you created the counts and irfs files from the
+# Alternatively, if you created the counts and IRF files from the
 # Fermi-LAT science tools without Fermipy you can use the
-# `create_dataset` method. Note that in this case we cannot guarantee
+# ``create_dataset``  method. Note that in this case we cannot guarantee
 # that your maps have the correct axes dimensions to be properly converted
 # into Gammapy datasets.
 #
@@ -284,13 +285,13 @@ dataset1 = reader.create_dataset(
 datasets_fromST = Datasets([dataset0, dataset1])
 
 
-# the above was an alternative reading method we don't need those after
+# The above was an alternative reading method we don't need those after
 del dataset0, dataset1, datasets_fromST
 
 
 ######################################################################
-# Fermi-LAT irfs properties
-# -------------------------
+# Fermi-LAT IRF properties
+# ------------------------
 #
 
 
@@ -302,7 +303,7 @@ del dataset0, dataset1, datasets_fromST
 
 ######################################################################
 # Exposure is almost constant across the field of view, with less than 5%
-# variations at a given energy.
+# variation at a given energy.
 #
 
 datasets[0].exposure.plot_interactive(add_cbar=True)
@@ -313,9 +314,9 @@ plt.show()
 # PSF
 # ~~~
 #
-# | For Fermi-LAT, the PSF only varies little within a given regions of
-#   the sky, especially at high energies like what we have here.
-# | So we have only one PSF kernel.
+# For Fermi-LAT, the PSF only varies little within a given regions of
+# the sky, especially at high energies like what we have here.
+# So we have only one PSF kernel.
 #
 
 datasets[0].psf.plot_containment_radius_vs_energy(fraction=(0.68, 0.95, 0.99))
@@ -323,13 +324,13 @@ plt.show()
 
 
 ######################################################################
-# Region of interest and Mask definition
+# Region of interest and mask definition
 # --------------------------------------
 #
-# As mention previously the width of dataset is larger that our actual
+# As mentioned previously, the width of dataset is larger that our actual
 # region of interest in order to properly take into account for the
 # sources outside that contributes inside due to the PSF. So we define the
-# valid RoI for fitting by creating a `mask_fit`.
+# valid RoI for fitting by creating a ``mask_fit``.
 #
 
 margin = (
@@ -357,7 +358,7 @@ for d in datasets:
 # Isotropic diffuse background
 # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 #
-# The `FermipyDatasetsReader` also created one isotropic diffuse model
+# The `~gammapy.datasets.FermipyDatasetsReader` also created one isotropic diffuse model
 # for each dataset:
 #
 
@@ -374,11 +375,11 @@ print(models_iso)
 ######################################################################
 # The Fermi-LAT collaboration provides a galactic diffuse emission model,
 # that can be used as a background model for Fermi-LAT source analysis.
-# These files are called usually IEM for intersellar emission model, the
+# These files are called usually IEM for interstellar emission model, the
 # latest is
 # `gll_iem_v07.fits <https://fermi.gsfc.nasa.gov/ssc/data/analysis/software/aux/4fgl/gll_iem_v07.fits>`__.
 # For details see the `BackgroundModels
-# page <https://fermi.gsfc.nasa.gov/ssc/data/access/lat/BackgroundModels.html>`__
+# page <https://fermi.gsfc.nasa.gov/ssc/data/access/lat/BackgroundModels.html>`__.
 # If you have Fermipy installed it can also be found in
 # `$FERMI_DIR/refdata/fermi/galdiffuse/gll_iem_v07.fits`
 #
@@ -407,6 +408,8 @@ model_iem = SkyModel(
 
 template_iem.map.plot_interactive(add_cbar=True)
 plt.show()
+# sphinx_gallery_thumbnail_number = 4
+
 
 models_diffuse = models_iso + model_iem
 
@@ -415,16 +418,16 @@ models_diffuse = models_iso + model_iem
 # Sources
 # ~~~~~~~
 #
-# Sources models can be loaded from the 4FGL catalog directly available in
-# $GAMMAPY_DATA. For details see the `Fermi-LAT catalog
-# page <https://fermi.gsfc.nasa.gov/ssc/data/access/lat/14yr_catalog/>`__
+# Source models can be loaded from the 4FGL catalog directly available in
+# `$GAMMAPY_DATA`. For details see the `Fermi-LAT catalog
+# page <https://fermi.gsfc.nasa.gov/ssc/data/access/lat/14yr_catalog/>`__.
 #
 
 catalog_4fgl = CATALOG_REGISTRY.get_cls("4fgl")()  # load 4FGL catalog
 
 
 ######################################################################
-# we want to select only the sources inside the dataset spatial geometry:
+# We want to select only the sources inside the dataset spatial geometry:
 #
 
 in_geom = geom.to_image().contains(catalog_4fgl.positions)
@@ -439,8 +442,8 @@ print("Number of source models", len(models_4fgl_gc))
 
 
 ######################################################################
-# In order to improve performaces we can store all the sources outside of
-# the `mask_fit` region into a single template (the same could be done
+# In order to improve performances we can store all the sources outside
+# the ``mask_fit`` region into a single template (the same could be done
 # for all the sources we want to keep frozen).
 #
 
@@ -469,9 +472,9 @@ print("Number of source models", len(models_sources))
 # ---
 #
 # Now, the big finale: let’s do a 3D of the brightest sources and IEM
-# models
+# models.
 #
-# First we attach the models to the datasets
+# First we attach the models to the datasets.
 #
 
 models = models_sources + models_diffuse
@@ -486,10 +489,10 @@ print("Number of models", len(models))
 #
 
 n_brightest = 3
-intergrated_flux = u.Quantity(
+integrated_flux = u.Quantity(
     [m.spectral_model.integral(10 * u.GeV, 1 * u.TeV) for m in sources_inside_roi]
 )
-order = np.argsort(intergrated_flux)
+order = np.argsort(integrated_flux)
 selected_sources = Models([sources_inside_roi[int(ii)] for ii in order[:n_brightest]])
 
 print(selected_sources.names)
@@ -498,20 +501,18 @@ free_models = selected_sources + model_iem
 
 
 ######################################################################
-# we keep only their normalisation free for simplicity:
+# We keep only their normalisation free for simplicity:
 #
 
 models.freeze()  # freeze all parameters
 
-# and unfreeze only the ampltitude or norm of the selected models
+# and unfreeze only the amplitude or norm of the selected models
 for p in free_models.parameters:
     if p.name in ["amplitude", "norm"]:
         p.frozen = False
         p.min = 0
 
 print("Number of free parameters", len(models.parameters.free_parameters))
-
-# %%time
 
 fit = Fit()
 result = fit.run(datasets=datasets)
@@ -523,7 +524,7 @@ print(result)
 # Residual TS map
 # ---------------
 #
-# Now we can look at the residual TS map to check there is no signifiant
+# Now we can look at the residual TS map to check there is no significant
 # excess left:
 #
 
@@ -546,7 +547,7 @@ import matplotlib as mpl
 
 
 def plot_significance_map(estimator_results, source_models=None, margin=None):
-    fig = plt.figure(figsize=(11, 4), dpi=130)
+    fig = plt.figure(figsize=(7, 5))
     image = estimator_results["sqrt_ts"]
     if margin is not None:
         image = image.cutout(
@@ -594,9 +595,9 @@ datasets_read = Datasets.read(
 # ---------
 #
 # - Fit the position and spectrum of the source `SNR
-#   G0.9+0.1 <http://gamma-sky.net/#/cat/tev/110>`__\ \_.
+#   G0.9+0.1 <http://gamma-sky.net/#/cat/tev/110>`__.
 # - Make maps and fit the position and spectrum of the `Crab
-#   nebula <http://gamma-sky.net/#/cat/tev/25>`__\ \_.
+#   nebula <http://gamma-sky.net/#/cat/tev/25>`__.
 #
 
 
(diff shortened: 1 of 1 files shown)


examples/tutorials/data/fermi_lat.py
```diff
@@ -26,196 +20,326 @@
...
 -----
 
-**IMPORTANT**: For this notebook you have to get the prepared `3fhl`
-dataset provided in your `$GAMMAPY_DATA`.
+**IMPORTANT**: For this notebook you have to get the prepared
+`fermi-gc` data provided in your $GAMMAPY_DATA.
 
-Note that the `3fhl` dataset is high-energy only, ranging from 10 GeV
-to 2 TeV.
+Note that the this dataset is high-energy only, ranging from 4 GeV to 2
```
facero:
```suggestion
Note that this dataset focuses on the high-energy only, ranging from 4 GeV to 2
```

examples/tutorials/data/fermi_lat.py
```diff
@@ -26,196 +20,326 @@
...
+#      roiwidth   : 8.0
+#      binsz      : 0.1
+#      binsperdec   : 10
+#      coordsys : GAL
+#      proj: CAR
+#      projtype: WCS
+#
+#    selection :
+#    #gtselect parameters
+#      emin : 3981.0717055349733 #ENERGY TRUE for gammapy
```
facero: why these stranges non integer values. Is this the energy of the bin edges or something ?
Why not 4 GeV and 2 TeV ?

examples/tutorials/data/fermi_lat.py
```diff
@@ -26,196 +20,326 @@
...
+# least equal to 0.2. With a binning of 8 to 10 bins per decade, it
+# corresponds to \|edisp_bins\| ≥ 2. For more details see
+# `Pass8_edisp_usage <https://fermi.gsfc.nasa.gov/ssc/data/analysis/documentation/Pass8_edisp_usage.html>`__.
 #
-# Let us start to prepare things for an 3D map analysis of the Galactic
-# center region with Gammapy. The first thing we do is to define the map
-# geometry. We chose a TAN projection centered on position
-# `(glon, glat) = (0, 0)` with pixel size 0.1 deg, and four energy bins.
+# In our case the we have 10 bins per decade and true energy axis starts
+# at about 4 GeV, so with edisp_bins=4 the reconstructed enegry axis
```
facero: Does the user has to compute which edisp_bins would be needed to start at 10 GeV ? If I had set edisp_bins=2 my analysis would have started at something like 8 GeV ?

examples/tutorials/data/fermi_lat.py
```diff
@@ -26,196 +20,326 @@
...
+# In your fermipy environment you have to run the following commands
+#
+# .. code:: python
+#
+#    from fermipy.gtanalysis import GTAnalysis
+#    gta = GTAnalysis('config_fermipy_gc_example.yaml',logging={'verbosity' : 3})
+#    gta.setup()
+#
+#    gta.compute_psf(overwrite=True) # this create the psf kernel
+#    gta.compute_drm(edisp_bins=0, overwrite=True) # this create the energy dispersion matrix
```
facero: Maybe add again the warning that edisp_bins=0 should not be modified here and edisp will be handled by gp later.

examples/tutorials/data/fermi_lat.py
```diff
@@ -26,196 +20,326 @@
...
-energy_axis = MapAxis.from_edges(
-    [1e4, 3e4, 1e5, 3e5, 2e6], name="energy", unit="MeV", interp="log"
+print(datasets[0].exposure.geom.axes["energy_true"])
+print(datasets[0].counts.geom.axes["energy"])
+
+
+######################################################################
+# Alternatively if you created the counts and irfs files from the
+# Fermi-LAT science tools without fermipy you can use the
+# `create_dataset` method. Note that in this case we cannot garantee
```
facero:
```suggestion
# `create_dataset` method. Note that in this case we cannot guarantee
```

examples/tutorials/data/fermi_lat.py
```diff
@@ -26,196 +20,326 @@
...
-cube (using `gtexpcube2` and PSF (using `gtpsf`). You can then use
-`~gammapy.data.EventList`, `~gammapy.maps` and the
-`~gammapy.irf.PSFMap` to read the Fermi-LAT maps and PSF, i.e. support
-for these high level analysis products from the Fermi ST is built in. To
-do a 3D map analysis, you can use Fit for Fermi-LAT data in the same way
-that it’s use for IACT data. This is illustrated in this notebook. A 1D
-region-based spectral analysis is also possible, this will be
-illustrated in a future tutorial.
+This tutorial will show you how to convert Fermi-LAT data into a DL4
+format that can be used by Gammapy and perform a 3D analysis. As an
```
AtreyeeS:
```suggestion
format  that can be used by Gammapy (`~gammapy.datasets.MapDataset`) and perform a 3D analysis. As an
```

examples/tutorials/data/fermi_lat.py
```diff
@@ -26,196 +20,326 @@
...
+example, we will look at the Galactic center.
 
 Setup
 -----
 
-**IMPORTANT**: For this notebook you have to get the prepared `3fhl`
-dataset provided in your `$GAMMAPY_DATA`.
+**IMPORTANT**: For this notebook you have to get the prepared
+`fermi-gc` data provided in your $GAMMAPY_DATA.
```
AtreyeeS: You can maybe mention how this data was obtained, maybe link to the script in Gammapy data ?

examples/tutorials/data/fermi_lat.py
```diff
@@ -26,196 +20,326 @@
...
-# We put this call into the same Jupyter cell as the Map.create
-# because otherwise we could accidentally fill the counts
-# multiple times when executing the `fill_by_coord` multiple times.
-counts.fill_events(events)
 
-print(counts.geom.axes[0])
+datasets_fromST = Datasets([dataset0, dataset1])
+
+
+del dataset0, dataset1, datasets_fromST  # we don't need those
```
facero:
```suggestion
del dataset0, dataset1, datasets_fromST  # the above was an alternative reading method we don't need those after
```

examples/tutorials/data/fermi_lat.py
AtreyeeS:
```suggestion
Note that this dataset contains only high energy photons, ranging from 4 GeV to 2
```

examples/tutorials/data/fermi_lat.py
```diff
@@ -26,196 +20,326 @@
...
-# would be better, but it doesn’t matter much here, because the energy
-# binning is fine. Finally, we just copy the counts map geometry, which
-# contains an energy axis with ``node_type="edges"``. This is non-ideal
-# for exposure cubes, but again, acceptable because exposure doesn’t vary
-# much from bin to bin, so the exact way interpolation occurs in later use
-# of that exposure cube doesn’t matter a lot. Of course you could define
-# any energy axis for your exposure cube that you like.
+# | For Fermi-LAT, the PSF only varies little within a given regions of
+#   the sky, especially at high energies like what we have here.
+# | So we have only one PSF kerne.
```
facero:
```suggestion
# | So we have only one PSF kernel.
```

examples/tutorials/data/fermi_lat.py
AtreyeeS: Ok you already do this later on. Ignore

examples/tutorials/data/fermi_lat.py
```diff
@@ -236,205 +367,211 @@
...
 ######################################################################
-# Isotropic diffuse background
-# ----------------------------
+# Sources
+# -------
 #
-# To load the isotropic diffuse model with Gammapy, use the
-# `~gammapy.modeling.models.TemplateSpectralModel`. We are using
-# `'extrapolate': True` to extrapolate the model above 500 GeV:
+# Sources models can ce loaded from the 4FGL catalog directly available in
```
facero:
```suggestion
# Sources models can be loaded from the 4FGL catalog directly available in
```

examples/tutorials/data/fermi_lat.py
```diff
@@ -236,205 +367,211 @@
...
+models_sources = sources_inside_roi + sources_outside_roi
+
 
 ######################################################################
-# In addition we can check how the actual shape of the PSF varies with
-# energy and compare it against the mean PSF between 50 GeV and 2000 GeV:
+# Fit
+# ===
+#
+# Now, the big finale: let’s do a 3D of the brightes sources and IEM
```
facero:
```suggestion
# Now, the big finale: let’s do a 3D of the brightest sources and IEM
```

examples/tutorials/data/fermi_lat.py
```diff
@@ -236,205 +367,211 @@
...
+print("Number of models", len(models))
 
-plt.xlim(1e-3, 0.3)
-plt.ylim(1e3, 1e6)
-plt.legend()
-plt.show()
 
 ######################################################################
-# This is what the corresponding PSF kernel looks like:
+# Let’s find the 3 brightess sources:
```
facero:
```suggestion
# Let’s find the 3 brightest sources:
```

examples/tutorials/data/fermi_lat.py
```diff
@@ -236,205 +367,211 @@
...
-
-
-######################################################################
-# Energy Dispersion
-# ~~~~~~~~~~~~~~~~~
-#
-# For simplicity we assume a diagonal energy dispersion:
-#
+order = np.argsort(intergrated_flux)
+selected_sources = Models([sources_inside_roi[int(ii)] for ii in order[:n_brightess]])
```
facero:
```suggestion
n_brightest = 3
integrated_flux = u.Quantity(
    [m.spectral_model.integral(10 * u.GeV, 1 * u.TeV) for m in sources_inside_roi]
)
order = np.argsort(integrated_flux)
selected_sources = Models([sources_inside_roi[int(ii)] for ii in order[:n_brightest]])

```

examples/tutorials/data/fermi_lat.py
```diff
@@ -26,196 +20,326 @@
...
-print(events.table.colnames)
 
-display(events.table[:5][["ENERGY", "RA", "DEC"]])
+######################################################################
+# | The most important points for gammapy users are: \* `emin`/`emax`
+#   in this file should be considered as the energy true range. \*
+#   `edisp_bins : 0` is strongly recommended at this stage otherwise you
+#   might face inconsitancies in the energies axes of the different IRFs.
+# | The reconstructed energy axis will be configurated with gammapy later
+#   on. \* if you want to use the `$FERMI_DIR` variable it must also be
```
AtreyeeS: You can mention that `FERMI_DIR` contains the path to fermi calibration files

examples/tutorials/data/fermi_lat.py
```diff
@@ -26,196 +20,326 @@
...
-# As a short analysis example we will count the number of events above a
-# certain minimum energy:
+# Then is your gammapy environment you can create the datasets using the
+# same configuration file.
 #
 
-for e_min in [10, 100, 1000] * u.GeV:
-    n = (events.energy > e_min).sum()
-    print(f"Events above {e_min:4.0f}: {n:5.0f}")
+from gammapy.datasets import FermipyDatasetsReader
```
AtreyeeS: We usually put all imports at the top, no?

facero (commented):
Thanks for tis very nice tutorial !

We had already identified this issue previously but it seems the main difficulty for the user relies on this edisp handling and whether Emin, Emax have to be set by hand and have exact energy edge values like in the yaml example you provided ?
As I suggested in the review, maybe add a warning at every line where edisp_bins is shown to say whether it should be modified by the user or not.

examples/tutorials/data/fermi_lat.py
```diff
@@ -26,196 +20,326 @@
...
+    "$GAMMAPY_DATA/fermi-gc/config_fermipy_gc_example.yaml", edisp_bins=4
+)
+datasets = reader.read()
+print(datasets)
 
 
 ######################################################################
-# Counts
-# ------
+# Note that the `edisp_bins` is set again here as a positive number so
```
AtreyeeS: This is a bit confusing. In line 145, you say edisp bins should be 0, now you put 4.
Can it be made the default and the user normally does not need to adapt it?

examples/tutorials/data/fermi_lat.py
```diff
@@ -26,196 +20,326 @@
...
+# gammapy can create its reconstructed enerygy axis properly. The energy
+# dispersion correction implemented gammapy is closer to the version
+# implemented in Fermitools >1.2.0, which take into account the interplay
```
AtreyeeS, line 232: This Fermitools version info can be given at the beginning

examples/tutorials/data/fermi_lat.py
QRemy: it's linear in log space the edges are  `10**0.6` and `10**3.4 GeV`.

examples/tutorials/data/fermi_lat.py
AtreyeeS: The reconstruction energy axis is  from 10GeV as you explain later. So, should this number also no be 10 GeV ?

examples/tutorials/data/fermi_lat.py
AtreyeeS:
```suggestion
# `create_dataset` method. Note that in this case we cannot guarantee
```

AtreyeeS: Maybe phrase it as  "Note that in this case it is the users responsibility to ensure that the counts and irf maps have the proper binning to be convertible to a gammapy `MapDataset`

examples/tutorials/data/fermi_lat.py
QRemy: yes with edisp_bins=2 the counts geom would have started at 10**0.8 =6.3 GeV but the user can stil determine the analysis range by updating the mask_fit to start a 10 GeV. Should I have this precision ?

examples/tutorials/data/fermi_lat.py
AtreyeeS: You can display the psf kernel as well

AtreyeeS:
```suggestion
# | So we have only one PSF kernel.
```

examples/tutorials/data/fermi_lat.py
```diff
@@ -236,205 +367,211 @@
...
 
-diffuse_iso = create_fermi_isotropic_diffuse_model(
-    filename=filename, interp_kwargs={"extrapolate": True}
-)
+catalog_4fgl = CATALOG_REGISTRY.get_cls("4fgl")()  # load 4FGL catalog
 
 
 ######################################################################
-# We can plot the model in the energy range between 50 GeV and 2000 GeV:
+# we want to select only the sources inside the dataset:
```
AtreyeeS:
```suggestion
# we want to select only the sources inside the dataset spatial geometry:
```

examples/tutorials/data/fermi_lat.py
```diff
@@ -236,205 +367,211 @@
...
 #
-energy_bounds = [50, 2000] * u.GeV
-diffuse_iso.spectral_model.plot(energy_bounds, yunits=u.Unit("1 / (cm2 MeV s)"))
-plt.show()
+
+in_geom = geom.to_image().contains(catalog_4fgl.positions)
+catalog_4fgl_gc = catalog_4fgl[in_geom]
+
+models_4fgl_gc = catalog_4fgl_gc.to_models()
```
AtreyeeS, line 433: Maybe add a `len(models_4fgl_gc)` or `models_4fgl_gc.names`

examples/tutorials/data/fermi_lat.py
AtreyeeS:
```suggestion
# Now, the big finale: let’s do a 3D of the brightest sources and IEM
```

examples/tutorials/data/fermi_lat.py
```diff
@@ -236,205 +367,211 @@
...
+spectral_model = PowerLawSpectralModel(index=2)
+model = SkyModel(spatial_model=spatial_model, spectral_model=spectral_model)
+
+ts_estimator = TSMapEstimator(
+    model,
+    kernel_width="1 deg",  # this set close to the 95-99% containment radius of the PSF
+    selection_optional=[],
+    sum_over_energy_groups=True,
+    energy_edges=[10, 1000] * u.GeV,
+    n_jobs=1,
```
AtreyeeS: Why only 1?

AtreyeeS (commented):
Thank you very much @QRemy for this excellent tutorial. Very very useful.

Maybe you can add in the beginning a list of different things this tutorial is exposing

examples/tutorials/data/fermi_lat.py
```diff
@@ -26,196 +20,326 @@
...
 #
 
-gc_pos = SkyCoord(0, 0, unit="deg", frame="galactic")
-energy_axis = MapAxis.from_edges(
-    [1e4, 3e4, 1e5, 3e5, 2e6], name="energy", unit="MeV", interp="log"
+print(datasets[0].exposure.geom.axes["energy_true"])
+print(datasets[0].counts.geom.axes["energy"])
+
+
+######################################################################
```
QRemy, line 250:
```suggestion
######################################################################
# Note that choosing instead `edisp_bins=2` means that the reconstructed energy of the counts geom
# would have start  at $10^0.8$ ~ 6.3 GeV$. In that case to start the analysis at 10 GeV we would have to
# update the `mask_fit` to ignore the 2 first reconstructed energy bins.
# Considering more `edisp_bins` is in general safer but consumes more memory and will increase computation time.
```

examples/tutorials/data/fermi_lat.py
QRemy: it doesn't help here the TS map computation is already fast compared to the multiprocessing overhead, I'll remove that line.

examples/tutorials/data/fermi_lat.py
QRemy:
```suggestion
#    gta.compute_drm(edisp_bins=0, overwrite=True) # this create the energy dispersion matrix 
#    # DO NOT CHANGE edisp_bins, it will be handled by gammapy
```

examples/tutorials/data/fermi_lat.py
QRemy: yes but for new features it's a bit annoying to have scroll up to find in which module their are and the code it not complete if we want to copy paste the cell. Opinions ?

examples/tutorials/data/fermi_lat.py
QRemy: Not really It has to be different between fermipy and gammapy,
It could also be 0 in gammpy but then we would have to always define a mask_fit to ignore the first and last reco bins which just move the problem.

I modified the text above to be more clear that gammapy set the value later.

examples/tutorials/data/fermi_lat.py
```diff
@@ -26,196 +20,326 @@
...
 #
 
-print(events.table.colnames)
 
-display(events.table[:5][["ENERGY", "RA", "DEC"]])
+######################################################################
+# | The most important points for gammapy users are: \* `emin`/`emax`
+#   in this file should be considered as the energy true range. \*
+#   `edisp_bins : 0` is strongly recommended at this stage otherwise you
+#   might face inconsitancies in the energies axes of the different IRFs.
```
QRemy:
```suggestion
#   might face inconsitancies in the energies axes of the different IRFs created by fermipy.
```

examples/tutorials/data/fermi_lat.py
```diff
@@ -26,196 +20,326 @@
...
+# | The reconstructed energy axis will be configurated with gammapy later
```
QRemy:
```suggestion
# | The `edisp_bins` value will be redefined later on as a positive value by gammapy in order to create the reconstructed energy axis properly.
```

examples/tutorials/data/fermi_lat.py
QRemy:
```suggestion
```

examples/tutorials/data/fermi_lat.py
```diff
@@ -26,196 +20,331 @@
...
+# Fermipy Congifuration file
+# --------------------------
 #
-# To load up the Fermi-LAT event list, use the `~gammapy.data.EventList`
-# class:
+# Gammapy can re-use the same configuration file than fermipy to convert
+# the fermipy maps to gammapy datasets. For details on these files
+# struture see the `fermipy configuration
+# page <https://fermipy.readthedocs.io/en/latest/config.html>`__. In this
+# tutorial we will analyse Galactic center data generated with the
```
QRemy:
```suggestion
# tutorial we will analyse Galactic center data generated with fermipy version 1.3 and the
```

examples/tutorials/data/fermi_lat.py
QRemy: added just before the config file

examples/tutorials/data/fermi_lat.py
QRemy: added :
```
"The energy axes will be created such as it is linear in log space 
# so it's better to define `emin`/`emax` such as they align with a log binning
# here we have $emin = 10^0.6$ ~ 4 GeV and $emax = 10^3.4$ ~ 2500 GeV`.
#"
```

examples/tutorials/data/fermi_lat.py
QRemy: added below

examples/tutorials/data/fermi_lat.py
QRemy: added "if you want to use the `$FERMI_DIR` variable to read the background models"

examples/tutorials/data/fermi_lat.py
QRemy: that doesn't bring much here

examples/tutorials/data/fermi_lat.py
AtreyeeS: I agree about the annoying part, but since we dont make this exception in other tutorials, it might be good to be consistent. Copy paste anyways does not work for most cells independently.

facero:
thanks for the quick  implementation of the comments !
Can we try ot have https://github.com/gammapy/gammapy-data/pull/76 in so that we can see the Sphinx  rendering of the tutorial ?

examples/tutorials/data/fermi_lat.py
```diff
@@ -2,220 +2,358 @@
...
 -----
 
-**IMPORTANT**: For this notebook you have to get the prepared `3fhl`
-dataset provided in your `$GAMMAPY_DATA`.
+**IMPORTANT**: For this notebook you have to get the prepared
+`fermi-gc` data provided in your $GAMMAPY_DATA.
 
-Note that the `3fhl` dataset is high-energy only, ranging from 10 GeV
-to 2 TeV.
+Note that this dataset contains only high energy photons, ranging from 4 GeV to 2
```
AtreyeeS: This 4 GeV I am still confused about. 
What is the minimum energy of the counts cube?

examples/tutorials/data/fermi_lat.py
```diff
@@ -2,220 +2,358 @@
...
+# Since the energy resolution varies with energy, it is important to
+# choose an energy binning that is fine enough to capture this energy
+# dependence. That is why we recommend a binning with 8 to 10 bins per
+# decade. The energy axes will be created such as it is linear in log space
+# so it's better to define `emin`/`emax` such as they align with a log binning
+# here we have $emin = 10^0.6$ ~ 4 GeV and $emax = 10^3.4$ ~ 2500 GeV`.
+#
+# The spatial binning should be of the same order of the PSF 68%
+# containment radius which is in average 0.1 degree above 10 GeV and
+# rapidly increases at lower energy. Ideally it should remain whitin of
```
AtreyeeS:
```suggestion
# rapidly increases at lower energy. Ideally it should remain within a
```

examples/tutorials/data/fermi_lat.py
```diff
@@ -2,220 +2,358 @@
...
+# factor of 2 or 3 of the PSF radius at most. in order to properly take
```
AtreyeeS:
```suggestion
# factor of 2 or 3 of the PSF radius at most. In order to properly take
```

examples/tutorials/data/fermi_lat.py
```diff
@@ -2,220 +2,358 @@
...
+# into account for the sources outside of the region of interest that
+# contributes inside due to the PSF we have to define a wider `roiwidth`
```
AtreyeeS:
```suggestion
# contribute inside due to the PSF we have to define a wider `roiwidth`
```

examples/tutorials/data/fermi_lat.py
```diff
@@ -2,220 +2,358 @@
...
+# than our actual region of interest typicaly we need a margin egal to the
```
AtreyeeS:
```suggestion
# than our actual region of interest. Typically, we need a margin equal to the
```

examples/tutorials/data/fermi_lat.py
```diff
@@ -2,220 +2,358 @@
...
+# 99% containment of the PSF on each side. Above 10 GeV considering only
+# PSF2&3 the 99% PSF containment radius is about 1 degree, assuming we
```
AtreyeeS:
```suggestion
# PSF2&3 the 99% PSF containment radius is about 1 degree. Thus, if we
```

examples/tutorials/data/fermi_lat.py
```diff
@@ -2,220 +2,358 @@
...
+# want to study 3 degree around the GC we have to take a `roiwidth` of 8
+# deg. (At 1 GeV and 100 MeV the margin should rather be of about 3 deg
```
AtreyeeS: This is confusing.
At 10 GeV: 8 degrees 
1 GeV: 3 degrees 
100 MeV: 10 degrees ?

examples/tutorials/data/fermi_lat.py
```diff
@@ -2,220 +2,358 @@
...
+# and 10 deg respectively, considering PSF0 and PSF1 it should be even
```
AtreyeeS:
```suggestion
# and 10 deg respectively. If including PSF0 and PSF1, it should be even
```

examples/tutorials/data/fermi_lat.py
```diff
@@ -2,220 +2,358 @@
...
+#
+# .. code:: python
+#
+#    from fermipy.gtanalysis import GTAnalysis
+#    gta = GTAnalysis('config_fermipy_gc_example.yaml',logging={'verbosity' : 3})
+#    gta.setup()
+#
+#    gta.compute_psf(overwrite=True) # this create the psf kernel
+#    gta.compute_drm(edisp_bins=0, overwrite=True) # this create the energy dispersion matrix
+#    # DO NOT CHANGE edisp_bins will be redefined by gammapy later on
```
AtreyeeS:
```suggestion
#    # DO NOT CHANGE edisp_bins here, it will be redefined by gammapy later on
```

examples/tutorials/data/fermi_lat.py
```diff
@@ -2,220 +2,358 @@
...
-energy_axis = MapAxis.from_edges(
-    [1e4, 3e4, 1e5, 3e5, 2e6], name="energy", unit="MeV", interp="log"
+print(datasets[0].exposure.geom.axes["energy_true"])
+print(datasets[0].counts.geom.axes["energy"])
+
+
+######################################################################
+# Note that choosing instead `edisp_bins=2` means that the reconstructed energy
+# of the counts geom would have start  at $10^0.8$ ~ 6.3 GeV$.
+# In that case to start the analysis at 10 GeV we would have to
```
AtreyeeS: But what is holy about starting analysis at 10 GeV?

examples/tutorials/data/fermi_lat.py
```diff
@@ -2,220 +2,358 @@
...
-# would be better, but it doesn’t matter much here, because the energy
-# binning is fine. Finally, we just copy the counts map geometry, which
-# contains an energy axis with ``node_type="edges"``. This is non-ideal
-# for exposure cubes, but again, acceptable because exposure doesn’t vary
-# much from bin to bin, so the exact way interpolation occurs in later use
-# of that exposure cube doesn’t matter a lot. Of course you could define
-# any energy axis for your exposure cube that you like.
+# | For Fermi-LAT, the PSF only varies little within a given regions of
+#   the sky, especially at high energies like what we have here.
+# | So we have only one PSF kernel.
```
AtreyeeS: If we have a section on the IRFs. it might be good to show what the said irf looks like. Looking at the kernel, one can see the things you have mentioned before about the energy dependence of the containment radius. Otherwise, avoid the subsection on the psf

examples/tutorials/data/fermi_lat.py
```diff
@@ -236,205 +381,212 @@
...
 
-source = SkyModel(
-    spectral_model=spectral_model,
-    spatial_model=spatial_model,
-    name="source-gc",
-)
-
-models = Models([source, diffuse_iem, diffuse_iso])
+# and unfreeze only the ampltitude or norm of the selected models
+for p in free_models.parameters:
```
AtreyeeS, line 507: Not relevant to this PR, I dont remember why we decided to do away with the `is_norm` attribute. We can consider adding some convenience functionality in models as freezing/un-freezing norms is a common use case

AtreyeeS (approved):
Thanks a lot! Left some small inline comments.
The edisp handling still is complicated to understand, but i think thats ok

examples/tutorials/data/fermi_lat.py
facero: you need to have an extra margin for the edisp. So you create a cube that is larger in energy and then you will use that for edisp. The final cube that you use for fitting in a smaller subset of the loaded dataset. (or at least this is how i understand it).
But it's true that this might not be easy to catch for users.
Maybe introduce this specificity strainght from the start in the header of the tutorial ?

examples/tutorials/data/fermi_lat.py
QRemy: Because the implementation was overcomplicated especially when multiple norm exits on a model, poorly defined for some models, and not fundamentally necessary to make the flux points work which was the original motivation to introduce it. If we want to define parameters types more generally we would need a better solution.

examples/tutorials/data/fermi_lat.py
```diff
@@ -2,220 +2,358 @@
...
-energy_axis = MapAxis.from_edges(
-    [1e4, 3e4, 1e5, 3e5, 2e6], name="energy", unit="MeV", interp="log"
+print(datasets[0].exposure.geom.axes["energy_true"])
+print(datasets[0].counts.geom.axes["energy"])
+
+
+######################################################################
+# Note that choosing instead `edisp_bins=2` means that the reconstructed energy
+# of the counts geom would have start  at $10^0.8$ ~ 6.3 GeV$.
+# In that case to start the analysis at 10 GeV we would have to
```
QRemy:
```suggestion
# In that case if we want to start the analysis at 10 GeV we would have to
```

examples/tutorials/data/fermi_lat.py
QRemy: well that was the minimal energy of the previous tutorial

examples/tutorials/data/fermi_lat.py
```diff
@@ -2,220 +2,358 @@
...
 
-**IMPORTANT**: For this notebook you have to get the prepared `3fhl`
-dataset provided in your `$GAMMAPY_DATA`.
+**IMPORTANT**: For this notebook you have to get the prepared
+`fermi-gc` data provided in your $GAMMAPY_DATA.
 
-Note that the `3fhl` dataset is high-energy only, ranging from 10 GeV
-to 2 TeV.
+Note that this dataset contains only high energy photons, ranging from 4 GeV to 2
+TeV.
```
QRemy:
```suggestion
We are going to analyses high-energy data from 10 GeV from 1 TeV (in reconstructed energy).
```

examples/tutorials/data/fermi_lat.py
```diff
@@ -2,220 +2,357 @@
 Fermi-LAT with Gammapy
 ======================
 
-Data inspection and preliminary analysis with Fermi-LAT data.
```
Astro-Kirsty: We should still have a small sentence here, in line with other tutorials

examples/tutorials/data/fermi_lat.py
```diff
@@ -2,220 +2,357 @@
...
-
-We note that support for Fermi-LAT data analysis in Gammapy is very
-limited. For most tasks, we recommend you use
-`Fermipy <http://fermipy.readthedocs.io/>`__, which is based on the
-`Fermi Science
+Gammapy fully supports Fermi-LAT data analysis from DL4 level (binned
+maps). In order to perform data reduction from the events list and
+spacecrast files to binned counts and IRFs maps we recommend to use
+`Fermipy <http://fermipy.readthedocs.io/>`__\ *, which is based on
+the*\ `Fermi Science
```
Astro-Kirsty:
```suggestion
the *\ `Fermi Science
```

examples/tutorials/data/fermi_lat.py
```diff
@@ -2,220 +2,357 @@
...
+format  that can be used by Gammapy (`~gammapy.datasets.MapDataset`) and perform a 3D analysis. As an
+example, we will look at the Galactic center.
 
 Setup
 -----
 
-**IMPORTANT**: For this notebook you have to get the prepared `3fhl`
-dataset provided in your `$GAMMAPY_DATA`.
+**IMPORTANT**: For this notebook you have to get the prepared
+`fermi-gc` data provided in your $GAMMAPY_DATA.
```
Astro-Kirsty:
```suggestion
`fermi-gc` data provided in your `$GAMMAPY_DATA`.
```

examples/tutorials/data/fermi_lat.py
```diff
@@ -2,220 +2,357 @@
...
 
-Note that the `3fhl` dataset is high-energy only, ranging from 10 GeV
-to 2 TeV.
+We are going to analyses high-energy data from 10 GeV from 1 TeV (in reconstructed energy).
 
 """
 
+# %matplotlib inline
+
```
Astro-Kirsty:
```suggestion
```

Astro-Kirsty: It is already on line 43

examples/tutorials/data/fermi_lat.py
```diff
@@ -2,220 +2,357 @@
...
 
 ######################################################################
 # Check setup
 # -----------
+#
+
+# %%capture
 from gammapy.utils.check import check_tutorials_setup
 
 check_tutorials_setup()
```
Astro-Kirsty: You can remove this now

examples/tutorials/data/fermi_lat.py
```diff
@@ -2,220 +2,357 @@
...
 
+
 ######################################################################
-# Events
-# ------
+# Fermipy Congifuration file
```
Astro-Kirsty:
```suggestion
# Fermipy configuration file
```

examples/tutorials/data/fermi_lat.py
```diff
@@ -2,220 +2,357 @@
...
+# --------------------------
 #
-# To load up the Fermi-LAT event list, use the `~gammapy.data.EventList`
-# class:
+# Gammapy can re-use the same configuration file than fermipy to convert
+# the fermipy maps to gammapy datasets. For details on these files
+# struture see the `fermipy configuration
+# page <https://fermipy.readthedocs.io/en/latest/config.html>`__. In this
+# tutorial we will analyse Galactic center data generated with fermipy version 1.3 and the
```
Astro-Kirsty:
```suggestion
# Gammapy can utilise the same configuration file as Fermipy to convert
# the Fermipy-generated maps into Gammapy datasets. For more information on the 
# structure of these files, refer to the `Fermipy configuration
# page <https://fermipy.readthedocs.io/en/latest/config.html>`__. In this
# tutorial, we will analyse Galactic center data generated with Fermipy version 1.3 and the
```

examples/tutorials/data/fermi_lat.py
```diff
@@ -2,220 +2,357 @@
...
+#
+#    model:
+#      src_roiwidth : 10.0 # This is used by Fermipy to compute the PSF RADMAX, even if no models are set
 #
 
-print(events.table.colnames)
 
-display(events.table[:5][["ENERGY", "RA", "DEC"]])
+######################################################################
+# | The most important points for gammapy users are: \* `emin`/`emax`
```
Astro-Kirsty: Unsure of the formatting you are trying to achieve, but I would make these bullet points. 
Remove the `|` at the start of each sentence and then put the `*` on new lines

examples/tutorials/data/fermi_lat.py
```diff
@@ -2,220 +2,357 @@
...
+# More generally in order to select a good binning it is important to know
+# the instrument resolution, for that you can have a quick look at the
+# irfs in the `Fermi-LAT performamce
+# page <https://www.slac.stanford.edu/exp/glast/groups/canda/lat_Performance.htm>`__.
+#
+# Since the energy resolution varies with energy, it is important to
+# choose an energy binning that is fine enough to capture this energy
+# dependence. That is why we recommend a binning with 8 to 10 bins per
+# decade. The energy axes will be created such as it is linear in log space
+# so it's better to define `emin`/`emax` such as they align with a log binning
```
Astro-Kirsty: As a general rule, when you reference a variable you should use double `  i.e. "``emax``"

examples/tutorials/data/fermi_lat.py
```diff
@@ -2,220 +2,357 @@
...
+# here we have $emin = 10^0.6$ ~ 4 GeV and $emax = 10^3.4$ ~ 2500 GeV`.
```
Astro-Kirsty: Math mode should be used

examples/tutorials/data/fermi_lat.py
```diff
@@ -2,220 +2,357 @@
...
+#    gta = GTAnalysis('config_fermipy_gc_example.yaml',logging={'verbosity' : 3})
+#    gta.setup()
+#
+#    gta.compute_psf(overwrite=True) # this create the psf kernel
+#    gta.compute_drm(edisp_bins=0, overwrite=True) # this create the energy dispersion matrix
+#    # DO NOT CHANGE edisp_bins here, it will be redefined by gammapy later on
+#
+# This will produce a number of files including: \* “ccube_00.fits”
+# (counts) \* “bexpmap_00.fits” (exposure) \* “psf_00.fits” (psf) \*
+# “drm_00.fits” (edisp)
```
Astro-Kirsty: Make proper bullet points again

examples/tutorials/data/fermi_lat.py
```diff
@@ -2,220 +2,357 @@
...
+#
 
 
 ######################################################################
-# As a short analysis example we will count the number of events above a
-# certain minimum energy:
+# Then is your gammapy environment you can create the datasets using the
+# same configuration file.
```
Astro-Kirsty:
```suggestion
# In your Gammapy environment you can create the datasets using the
# same configuration file.
```

examples/tutorials/data/fermi_lat.py
QRemy: At 10 GeV 1deg margin so 8 deg roiwidth
I removed the details for the margin at other energies to avoid confusion.

Astro-Kirsty (changes requested):
Thanks @qremy. I agree with the outstanding comments of Atreyee. 

Other than that I added some comments, and in general:
- you should use the proper math mode not `$`
- Fermipy and Gammapy should be capatilised
- you should use double ` for variables i.e. ``time``
- the only title with `====` should be the first one, others should be lower than that

I will re-review once these have been implemented and look at the docs build

examples/tutorials/data/fermi_lat.py
QRemy: ok I put it back

examples/tutorials/data/fermi_lat.py
```diff
@@ -2,220 +2,357 @@
 Fermi-LAT with Gammapy
 ======================
```
QRemy:
```suggestion

Data inspection and preliminary analysis with Fermi-LAT data.

```

examples/tutorials/data/fermi_lat.py
QRemy: I think we should keep it that's safer to download the data in case their are missing.
But we could add an option to hide the printout by default like `verbose=False`.

examples/tutorials/data/fermi_lat.py
QRemy: added

examples/tutorials/data/fermi_lat.py
QRemy: Oh you mean the capture, sorry

Astro-Kirsty: I did mean the `check_tutorials_setup()` as #5941 add the extra option. 
But if you think it is important for this tutorial, then yes keep it

facero:
Maybe my last point would be to add a comment in of where the user should input the weights file if they have produced one.
Certainly for this tutorial we don't want to include weights but could be good to just explain where to inject them (without saying how to produce them as this is advanced user stuff anyway).

Astro-Kirsty:
For correct bullet points see https://github.com/gammapy/gammapy/blob/0b2c8b52e807afadaf51d700aef5a481acaad362/examples/tutorials/starting/overview.py#L94-L97

QRemy:
> Maybe my last point would be to add a comment in of where the user should input the weights file if they have produced one. Certainly for this tutorial we don't want to include weights but could be good to just explain where to inject them (without saying how to produce them as this is advanced user stuff anyway).

For now the reader does nothing about the weight maps. So if the user want to add them they have to open them as a map and set them as mask_fit but I'm not sure if they have to be interpolated before or not. Are they in reco or true energy ?
I would keep that for later, first check how to create the weigths with fermipy and then update the reader.

facero:
Good point.
Yeah let's keep it for later then or for advanced user who want to explore

Le lun. 21 juil. 2025, 18:40, Quentin Remy ***@***.***> a
écrit :

examples/tutorials/data/fermi_lat.py
```diff
@@ -236,205 +390,214 @@
...
+    plt.title(r"$\sqrt{TS}$", loc="left", fontsize=12, pad=10)
+    cb_ax = fig.add_axes([0.5, 0.91, 0.15, 0.04])
+    cb = fig.colorbar(im, cax=cb_ax, orientation="horizontal")
+    cb.ax.xaxis.set_ticks_position("top")
+    cb.ax.xaxis.set_label_position("top")
+    cb.ax.tick_params(axis="both", which="major", labelsize=12)
+    cb.ax.set_xlabel(r"[$\sigma$]", fontsize=12, labelpad=-28, x=-0.1)
+
+
+plot_significance_map(ts_results, source_models=sources_inside_roi, margin=margin)
```
Astro-Kirsty: This is over complexity. I would suggest to stick with our plotting functionalites:

```
image = ts_results["sqrt_ts"]
image = image.cutout(
    image.geom.center_skydir, width=np.max(image.geom.width) - 2 * margin
)

fig = plt.figure(figsize=(7, 5))
ax = image.plot(
    clim=[-8, 8],
    cmap=plt.cm.RdBu_r,
    add_cbar=True,
    kwargs_colorbar={"label": "Significance ($\sigma$)"},
)
sources_inside_roi.plot_regions(
    ax=ax, edgecolor="g", linestyle="-", kwargs_point=dict(marker=".")
)
```

QRemy: done

QRemy:
Thanks all for reviewing. 
All comments are implemented and the final formatting problems were fixed by @Astro-Kirsty. 
I'm going to merge this once the doc build pass.

bkhelifi:
Let's discuss about this PR during a dev call. We have remaining questions...

examples/tutorials/data/fermi_lat.py
```diff
@@ -7,226 +7,378 @@
...
+# - ``emin`` and ``emax`` in this file should be considered as the energy true range.
+#   It should be larger that the reconstructed energy range.
+# - ``edisp_bins : 0`` is strongly recommended at this stage otherwise you
+#   might face inconsistencies in the energy axes of the different IRFs created by Fermipy.
+# - The ``edisp_bins`` value will be redefined later on by Gammapy as a positive value
+#   in order to create the reconstructed energy axis properly.
+# - If you want to use the `$FERMI_DIR` variable to read the background models
+#   it must also be defined in your Gammapy environment,
+#   otherwise you have to define your own paths.
+# - For this tutorial we copied the iso files in $GAMMAPY_DATA/fermi-gc` and
```
Astro-Kirsty:
```suggestion
# - For this tutorial we copied the iso files in `$GAMMAPY_DATA/fermi-gc` and
```

Final comment

bkhelifi:
@QRemy, could you solve the conflicts?
PS: when the doc will be built, one could check the final web rendering of the tutorial (except if there are other methods, @Astro-Kirsty?) 

Otherwise, it looks great and good for me. I left some very minor comments. Thank you @QRemy for the effort.

PR #5996, opened 2025-07-21
Cleanup docs

Astro-Kirsty:
This is to clean up the docs a bit. 

I also ran `pytest --doctest-glob="*.rst" docs/` and fixed some of those errors. 
We still have an error with the code [here](https://docs.gammapy.org/dev/user-guide/dl3.html#writing-event-lists-and-gtis-to-file). It is unclear how we should proceed to fix that, as we are trying to show saving the EventList and GTI individually, but then if you try to save the `obs` you don't have the pointing, so it will not write (leading to an error).


docs/getting-started/usage.rst
```diff
@@ -50,7 +50,7 @@ Gammapy is a Python package, so you can of course import and use it from Python:
     Type "help", "copyright", "credits" or "license" for more information.
     >>> from gammapy.stats import CashCountsStatistic
     >>> CashCountsStatistic(n_on=10, mu_bkg=4.2).sqrt_ts
-    np.float64(2.397918129147546)
+    2.397918129147546
```
AtreyeeS, line 53: is it good to test on a string like this? Some values might change because of the tolerance, no?

AtreyeeS (dismissed):
Thanks a lot @Astro-Kirsty ! We have been having this broken docs issues for long now.
I think in the broken example you point to, you can add a pointing info because it is indeed needed for serialisation in gadf format

docs/development/pigs/pig-027.rst
```diff
@@ -44,19 +44,20 @@ The following PIG focuses on the implementation of the NS with the UltraNest pac
 However, UltraNest is claimed to be more robust, provide safer uncertainties, and better parallelization (see `UltraNest doc <https://johannesbuchner.github.io/UltraNest/issues.html#what-is-the-difference-between-ultranest-and-multinest-dynesty>`_)
 
 Reference :
+
 - Ultranest : https://johannesbuchner.github.io/UltraNest
 - Literature : `Buchner 2023, <http://arxiv.org/abs/2101.09675>`_ , `Ashton 2022, <https://ui.adsabs.harvard.edu/abs/2022NRvMP...2...39A>`_
 
 
-**Potential limitations of MCMC in a BayesianInference class ?**
+**Potential limitations of MCMC in a BayesianInference class?**
```
bkhelifi, line 52: A naive question: is it wise to modify an approved PIG?

Astro-Kirsty: Given that I am only fixing some issues with the docs build and spelling, I think it is fine

docs/development/doc_howto.rst
```diff
@@ -110,8 +110,9 @@ following syntax.
         >>> from gammapy.data import EventList
         >>> event_list = EventList.read('events.fits') # doctest: +SKIP
 
-In the case above, we could check the execution of the first two lines importing the ``Quantity`` and ``EventList``
-modules, whilst the third line will be skipped. On the contrary, in the example below we could check the execution of
+In the case above, we could check the execution of the first two lines importing the
+`~astropy.units.Quantity` and '~gammapy.data.EventList` modules, whilst the third line will be skipped.
```
bkhelifi:
```suggestion
`~astropy.units.Quantity` and `~gammapy.data.EventList` modules, whilst the third line will be skipped.
```

docs/user-guide/makers/reflected.rst
```diff
@@ -87,7 +87,7 @@ The reflected region finder
 
 The following example illustrates how to create reflected regions for a given
 circular on region and exclusion mask using the
-`~gammapy.makers.background.ReflectedRegionsFinder`. In particular, it shows how to
+`~gammapy.makers.ReflectedRegionsFinder`. In particular, it shows how to
```
AtreyeeS, line 90: So the path should be the one where one imports from? 
Like `~gammapy.modeling.models.PowerLawSpectralModel` but `~gammapy.makers.FovBackgroundMaker`?

Astro-Kirsty: Yes exactly

docs/user-guide/dl3.rst
```diff
@@ -111,7 +116,8 @@ of the sky.
 Combining event lists and GTIs
 ------------------------------
 
-Both event lists and GTIs can be stacked into a larger one. An `~gammapy.data.EventList` can also be created `~gammapy.data.EventList.from_stack`, that is,
+Both event lists and GTIs can be stacked into a larger one. An `~gammapy.data.EventList` can
+also be created `~gammapy.data.EventList.from_stack`, that is,
```
bkhelifi, line 120:
```suggestion
also be created `~gammapy.data.EventList.from_stack()`, that is,
```

bkhelifi (approved):
Very very nice enhancement. Thanks

docs/user-guide/dl3.rst
Astro-Kirsty: You don't actually need the brackets (see the docs build!)

PR #5998, opened 2025-07-21
Remove tutorial setup

### LINKED ISSUE #5816: Removal of check-setup in tutorials

Astro-Kirsty:
This PR is to be merged after #5941. Resolves #5816

A command line option is added to check the tutorial setup and then this PR is to remove most of these actions from the tutorials. I leave it in the first three, as a demonstration.

I haven't adapted the api tutorials yet as we should wait for a decision on  #5949 first (otherwise there will be major conflicts)


AtreyeeS (approved):
Thanks @Astro-Kirsty

morcuended (approved):
Thanks @Astro-Kirsty. Looks good to me

Issue #5549, opened 2024-11-07
TemplateSpatialModel serialisation is not  backward compatible


QRemy:
**Gammapy version**
1.2 & dev
**Bug description**
yaml serialisation of TemplateSpatialModel  is not  backward compatible due to the addition of lon/lat parameters that change the expected shape of the covariance matrix.

**Expected behavior**
I would propose add  a new model StaticTemplateSpatialModel to support the legagy TemplateSpatialModel as defined in v1.1 and add a  mecanism that creates an instance of this model if the lon/lat parameters are not given to TemplateSpatialModel.from_dict().

registerrier:
Adding a new class to allow for backward compatibility seems a bit complex to me. Can't we handle this with the `from_dict` method only?

bkhelifi:
When serialising models and datasets, do we store the gammapy version? This is the minimal to have to permit extended feaures...

registerrier:
Pooposal here is to write the workaround of not reading the covariance matrix in the `Models.read()` in the release notes and work on a solution for 2.0.
Postponing.

AtreyeeS:
Should we postpone this again? Or do we remove support for 1.x files from 2.x? @registerrier @bkhelifi @QRemy @adonath ?

Astro-Kirsty:
I think we can/should fix this. We should hopefully be able to adjust the `to_dict` functionalities

AtreyeeS:
@Astro-Kirsty I think the problem is in reading previously serialised files, so we have to adapt `from_dict`

AtreyeeS:
Is this a problem for the SDC @fabiopintore ? Do you have Template Spatial models written with with version 1.1 and before?

fabiopintore:
For the SDC, we are currently collecting new models that I expect to be created with the latest Gammapy versions. 
However, some of the models will not be replaced and this might introduce some troubles when reading them. 
The easiest solution could be, as @registerrier Regis proposed, to avoid reading the covariance matrix.

bkhelifi:
And other possibility is:
- when there is no stored metadata (containing the gammapy version), we do not read the cov matrix
- when we have metadata with the gammapy version (feature that we have now), we read the cov matrix...
Because we are saving such metadata, one can deal with special cases

AtreyeeS:
Addressed by #5999

PR #5999, opened 2025-07-21
Handling non-backward compatible format of the covariance matrix

bkhelifi:
**Description**

This pull request is associated to the issue #5549. It correctly catches the problem.

**Dear reviewer**
For the 2.1, one should read correctly the Gammapy version in the metadata of the correlation matrix, and treat correctly non-backward compatible formats.
We have no available version of an old model of the TemplateSpatialModel with a such correlation matrix. So, this is hard to make a test...

### DIFF
--- a/gammapy/modeling/models/core.py
+++ b/gammapy/modeling/models/core.py
@@ -573,11 +573,11 @@ class DatasetModels(collections.abc.Sequence, CovarianceMixin):
                 path, filename = split(filename)
             try:
                 models.read_covariance(path, filename, format="ascii.fixed_width")
-            except ValueError:
+            except ValueError as exception:
                 log.warning(
-                    "Impossible to read correctly the covariance! \n"
-                    f"Might be due to bad filename [{filename}] or "
-                    "a bad formatting (not backward compatible)... "
+                    "Impossible to read the covariance correctly: \n"
+                    f"{exception} \n "
+                    "Covariance will be created from errors and non-diagonal terms ignored."
                 )
 
         _set_models_link(models)
--- a/gammapy/modeling/models/core.py
+++ b/gammapy/modeling/models/core.py
@@ -571,7 +571,12 @@ class DatasetModels(collections.abc.Sequence, CovarianceMixin):
             filename = data["covariance"]
             if not (path / filename).exists():
                 path, filename = split(filename)
-            models.read_covariance(path, filename, format="ascii.fixed_width")
+            try:
+                models.read_covariance(path, filename, format="ascii.fixed_width")
+            except ValueError:
+                log.warning(
+                    f"Impossible to read correctly the covariance! \n Might be due to bad filename [{filename}] or a bad formatting (not backward compatible)... "
+                )
 
         _set_models_link(models)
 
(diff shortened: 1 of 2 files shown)


QRemy:
> We have no available version of an old model of the TemplateSpatialModel with a such correlation matrix. So, this is hard to make a test...

We just need to save a valid model, hack the covariance to make it invalid, and read back to catch the message, so I just asked an AI to implement that:
```
import numpy as np
from pathlib import Path
from gammapy.modeling.models import SkyModel, PowerLawSpectralModel, Models

def test_invalid_covariance_matrix(tmp_path):
    # Step 1: Create a simple spectral model
    spectral_model = PowerLawSpectralModel()
    model = SkyModel(spectral_model=spectral_model, name="test-model")
    models = Models([model])

    # Step 2: Assign a valid 3x3 identity covariance matrix
    models.covariance = np.identity(3)

    # Step 3: Write models.yaml and covariance.dat
    models_file = tmp_path / "models.yaml"
    models.write(models_file, write_covariance=True, overwrite=True)

    # Step 4: Overwrite covariance.dat with an invalid 1x1 matrix
    covariance_file = tmp_path / "models_covariance.dat"
    with open(covariance_file, "w") as f:
        f.write("|                    Parameters |   0 |\n")
        f.write("|     test-model.spectral.index | 1.0 |\n")

    # Step 5: Attempt to read the model and catch the expected error
    try:
        Models.read(models_file)
    except Exception as e:
        print("Captured error message:", str(e))
        assert "covariance" in str(e).lower() or "shape" in str(e).lower()
```

gammapy/modeling/models/core.py
```diff
@@ -571,7 +571,14 @@ def from_dict(cls, data, path=""):
...
                 path, filename = split(filename)
-            models.read_covariance(path, filename, format="ascii.fixed_width")
+            try:
+                models.read_covariance(path, filename, format="ascii.fixed_width")
+            except ValueError:
+                log.warning(
+                    "Impossible to read correctly the covariance! \n"
+                    f"Might be due to bad filename [{filename}] or "
+                    "a bad formatting (not backward compatible)... "
+                )
```
QRemy:
```suggestion
            except ValueError as exception:
                log.warning(
                    "Impossible to read the covariance correctly: \n"
                    f"{exception} \n "
                    "Covariance will be created from errors ignoring non-diagonal terms."
                )
```

AtreyeeS (dismissed):
Thanks @bkhelifi !
I wonder if we should make the error message more explicit, like "Mismatch in number of parameters between model and covariance matrix. Check your Gammapy version.
Covariance will be created from errors and non-diagonal terms ignored."

bkhelifi:
One might have other type of errors that is catch by this function. So the message is generic. With the addon of @QRemy, the exception is printed in the warning, allowing the user to understand the nature of the issue

gammapy/modeling/models/tests/test_core.py
```diff
@@ -360,3 +361,30 @@ def test_two_fov_bkg_models_single_dataset():
...
+    covariance_file = tmp_path / "models_covariance.dat"
+    with open(covariance_file, "w") as f:
+        f.write("|                    Parameters |   0 |\n")
+        f.write("|     test-model.spectral.index | 1.0 |\n")
+
+    # Step 5: Attempt to read the model and catch the expected error
+    with caplog.at_level(logging.WARNING):
+        Models.read(models_file)
+        assert (
+            "Impossible to read correctly the covariance" in caplog.records[0].message
```
QRemy:
```suggestion
            "Impossible to read the covariance correctly" in caplog.records[0].message
```

PR #6000, opened 2025-07-21
Tutorials run as notebooks

### LINKED ISSUE #4696: Check that tutorials run as notebooks

It seems the tutorials are not tested as notebooks anymore since the conversion to jupytext.

Executing the event_sampling tutorial I got an error "Did not find line magic function %%time", 
which seems to come from a missing cell boundary (%%time in the middle of a cell instead of at the beginning as first line).

Check that tutorials work when run as notebook.

Astro-Kirsty:
This PR resolves #4696

Removing the %%time element from the tutorials as they are not even rendered by sphinx-gallery and to add a workflow for creating the tutorials

### DIFF
--- a/docs/development/doc_howto.rst
+++ b/docs/development/doc_howto.rst
@@ -110,8 +110,9 @@ following syntax.
         >>> from gammapy.data import EventList
         >>> event_list = EventList.read('events.fits') # doctest: +SKIP
 
-In the case above, we could check the execution of the first two lines importing the ``Quantity`` and ``EventList``
-modules, whilst the third line will be skipped. On the contrary, in the example below we could check the execution of
+In the case above, we could check the execution of the first two lines importing the
+`~astropy.units.Quantity` and `~gammapy.data.EventList` modules, whilst the third line will be skipped.
+On the contrary, in the example below we could check the execution of
 the code as well as the output value produced.
 
 .. code-block:: text
--- a/docs/development/intro.rst
+++ b/docs/development/intro.rst
@@ -119,7 +119,7 @@ using git branches.
 
 So how large should one pull request be?
 
-Our experience in Gammapy (and others confirm, see e.g. `here
+Our experience in Gammapy (and others confirm, see e.g. `this link
 <https://alexgaynor.net/2015/dec/29/shrinking-code-review/>`__) is that smaller
 is better. Working on a pull request for an hour or maximum a day, and having a
 diff of a few to maximum a few 100 lines to review and discuss is pleasant.
@@ -154,8 +154,8 @@ Firstly, fork the `gammapy repository <https://github.com/gammapy/gammapy>`_ to 
 personal GitHub. If you have not already done so, copy your public SSH key (normally under
 ``$HOME/.ssh/*.pub``) to your GitHub profile. This is done by going into your GitHub settings,
 selecting **SSH and GPG keys** then **New SSH key** and then choosing **Authentication Key**
-as the key type. If you do not have a SSH key, you may create it following the instructions on
-`this page <https://docs.github.com/en/authentication/connecting-to-github-with-ssh/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent>`_.
+as the key type. If you do not have a SSH key, you may create it following
+`these instructions <https://docs.github.com/en/authentication/connecting-to-github-with-ssh/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent>`_.
 
 Go to the location on your machine where you want to clone the gammapy
 repository from your GitHub:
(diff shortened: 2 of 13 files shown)


Astro-Kirsty:
Currently the following have failures when running `jupytext --execute examples/tutorials/tutorial-name.py`

- `analysis-1d/spectral_analysis.py` [#L484](https://github.com/gammapy/gammapy/blob/67821b7177daecdd24b8f32f18895c1d2f7672ee/examples/tutorials/analysis-1d/spectral_analysis.py#L484) --> AttributeError: FillBetweenPolyCollection.set() got an unexpected keyword argument 'yunits'
- `analysis-1d/spectral_analysis_rad_max.py` [#L343-347](https://github.com/gammapy/gammapy/blob/67821b7177daecdd24b8f32f18895c1d2f7672ee/examples/tutorials/analysis-1d/spectral_analysis_rad_max.py#L343) --> AttributeError: FillBetweenPolyCollection.set() got an unexpected keyword argument 'yunits'
- `api/catalogs.py` [#L304](https://github.com/gammapy/gammapy/blob/67821b7177daecdd24b8f32f18895c1d2f7672ee/examples/tutorials/api/catalog.py#L304) --> AttributeError: FillBetweenPolyCollection.set() got an unexpected keyword argument 'yunits'


--> these errors were fixed by removing the `yunits` which is not supported

.github/workflows/tutorialjupytexttests.yml
```diff
@@ -0,0 +1,27 @@
...
+        uses: actions/checkout@v4
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.11'
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
```
morcuended, line 24: Is this installing the dependencies or just upgrading pip?

examples/tutorials/api/catalog.py
```diff
@@ -299,7 +299,9 @@
 
 # %%
 energy_bounds = (100 * u.MeV, 100 * u.GeV)
-opts = dict(sed_type="e2dnde", yunits=u.Unit("TeV cm-2 s-1"))
+opts = dict(sed_type="e2dnde")
+fig, ax = plt.subplots()
+ax.yaxis.set_units(u.Unit("TeV cm-2 s-1"))
```
morcuended, line 304: Checking at the rendered docs, the plot is not actually showing the units in y-axis, although it does show up in the notebook, so it must be cropped because of the size. It was like this before the changes in this PR.

morcuended (commented):
Thanks @Astro-Kirsty. Very much needed. I left some comments.

QRemy (commented):
The %time can give a valuable insight on the performace of gammapy for new users so I would replace them by a time_elapsed decorator has show [here](https://stackoverflow.com/questions/30433910/decorator-to-time-specific-lines-of-the-code-instead-of-whole-method)

Similarly %capture that hide the print out could be replace by a decorator [see](https://stackoverflow.com/questions/8391411/how-to-block-calls-to-print)
so we could keep the check_tutorials_setup() but hide the versions output
(so gammapy data is still downloaded if missing).

Astro-Kirsty:
`%%time` does not show in the render of the docs build, hence why Regis suggested to remove it all together. 
 I will see if the decorator works, but we for sure cannot use `%%time` as it causes errors when trying to run. 
 
 As for the `check_tutorials_setup()` this was discussed a few times in dev calls. The decision was to add [this PR ](#5941) stating how to do this at the start of tutorials section and leave the check setup in the first set of tutorials (as done in here #5998)

AtreyeeS:
The time_elapsed decorator as suggested by @QRemy is vey nice, but I am worried it might get complex for beginners.
%%time is simpler to understand

QRemy:
> The time_elapsed decorator as suggested by @QRemy is vey nice, but I am worried it might get complex for beginners. %%time is simpler to understand

Is it really more complicated to understand ?
```
%% time
do_something
```
```
with estimate_time():
    do_something
```

Astro-Kirsty:
I think @AtreyeeS's point was more that, adding this decorator to so many tutorials could be confusing because in the end it is another function that the user would be curious to see what is it actually doing.

.github/workflows/tutorialjupytexttests.yml
Astro-Kirsty: Good question, I was copying from another workflow!

Astro-Kirsty: It should do as we expect

morcuended:
I agree with removing the %%time command, it does only show up if you download the tutorial as a python script and I think it is overkill to show it in all the tutorials.

adonath:
@QRemy VSCode, Google Collab etc report the execution time per cell anyway.

AtreyeeS:
Thanks @Astro-Kirsty ! Merging now

PR #6002, opened 2025-07-22
Update of plot functions for the colour bar

bkhelifi:
**Description**

A last minute issue was found just after the merging of the PR #5992. This PR fixes the colour bar printing for both simple plots and figures with sub-plots.

**Dear reviewer**
I have checked:
- the plot functions of : events, aeff, psf, edisp, bkg
- the peek functions of : events, aeff, psf, edisp, bkg, observation (for HESS and VERITAS)
One should check after the building of the documentation that there is no remaining issue in the tutorials. Could you check that during your review? Thank you

### DIFF
--- a/gammapy/maps/tests/test_core.py
+++ b/gammapy/maps/tests/test_core.py
@@ -517,6 +517,8 @@ def test_map_plot_mask():
 
     with mpl_plot_check():
         mask.plot_mask()
+    with mpl_plot_check():
+        mask.plot_mask(add_cbar=True)
 
 
 def test_reproject_2d():
--- a/gammapy/maps/wcs/tests/test_ndmap.py
+++ b/gammapy/maps/wcs/tests/test_ndmap.py
@@ -703,6 +703,10 @@ def test_plot():
     m = WcsNDMap.create(binsz=0.1 * u.deg, width=1 * u.deg, axes=[axis])
     with mpl_plot_check():
         m.plot(add_cbar=True)
+    with mpl_plot_check():
+        m.plot()
+    with mpl_plot_check():
+        m.plot(squared=True)
 
 
 def test_plot_grid():
@@ -1097,6 +1101,13 @@ def test_plot_mask():
 
     with mpl_plot_check():
         mask.plot_grid()
+
+    summed_mask = mask.sum_over_axes()
+    summed_mask.data = np.array(summed_mask.data, dtype=bool)
+    with mpl_plot_check():
+        summed_mask.plot_mask()
+    with mpl_plot_check():
+        summed_mask.plot_mask(add_cbar=True)
 
 
 def test_histogram_center_value():


morcuended (commented):
Thanks @bkhelifi. I had a quick look to the tutorials and I only saw a peek plot with problems in the cbar location in tutorials/api/datasets (`Dataset.peek()`):

image

bkhelifi:
> Thanks @bkhelifi. I had a quick look to the tutorials and I only saw a peek plot with problems in the cbar location in tutorials/api/datasets (`Dataset.peek()`):
> 
Well catch! Thanks a lot for the detailed check. Fixed

Figure_1

gammapy/datasets/map.py
```diff
@@ -2391,7 +2391,7 @@ def peek(self, figsize=(13.0, 7)):
 
         Parameters
         ----------
-        figsize : tuple
+        figsize : tuple, optional
             Size of the figure. Default is (13.5, 7).
```
morcuended: Be consistent here with the actual default value
```suggestion
            Size of the figure. Default is (13.0, 7).
```

gammapy/irf/psf/core.py
```diff
@@ -222,12 +230,20 @@ def plot_containment_radius(
 
         kwargs_colorbar = kwargs_colorbar or {}
 
+        squared = False
+        if "squared" in kwargs:
+            squared = kwargs["squared"]
+            kwargs.pop("squared")
+
```
morcuended, line 237: I guess these "squared" settings that are all over the place could be generalized at some point later.

gammapy/data/observations.py
```diff
@@ -480,25 +480,23 @@ def peek(self, figsize=(15, 10)):
...
             if name == "bkg":
                 bkg = self.bkg
                 if not bkg.has_offset_axis:
                     bkg = bkg.to_2d()
-                bkg.plot(ax=ax)
+                bkg.plot(ax=ax, squared=True)
                 ax.set_title("Background rate")
 
             if name == "psf":
                 self.psf.plot_containment_radius_vs_energy(ax=ax)
```
morcuended: To be consistent with the squared layout, I think here should be set as well. See, e.g., observations peek plot in `tutorials/starting/analysis_2`

morcuended (commented):
Thanks @bkhelifi! I think it is looking much better now. Just realized in some places that the axis labels/ticks of the color bars when there are multiple panels are overplotted onto other panels. See e.g. `tutorials/api/datasets`

image

Astro-Kirsty:
Recall the functionality implemented here https://github.com/gammapy/gammapy/pull/5055

That `axes_loc` allows us to easily adjust the colorbar positioning. Could it be of use??
Something like this:

```
kwargs_colorbar = {'label':'Colorbar label'}
axes_loc = {"position": "right", "size": "2%", "pad": "10%"}
fig = plt.figure(figsize=(6, 3))
ax = fig.add_subplot(111)
map_.sum_over_axes().plot(ax=ax, add_cbar=True, axes_loc=axes_loc,
                          kwargs_colorbar=kwargs_colorbar)
```

bkhelifi:
> Recall the functionality implemented here #5055
> 
> That `axes_loc` allows us to easily adjust the colorbar positioning. Could it be of use?? Something like this:
> 
This does not help, unfortunately. This is the reason of the proposed solution.

Astro-Kirsty:
Does this also have something to do with the warnings we get?
Example, below [this plot](https://docs.gammapy.org/dev/tutorials/data/hawc.html#check-the-dataset-we-produced)

_/home/runner/work/gammapy-docs/gammapy-docs/gammapy/.tox/build_docs/lib/python3.11/site-packages/gammapy/datasets/map.py:2530: UserWarning: This figure includes Axes that are not compatible with tight_layout, so results might be incorrect.
  plt.tight_layout(w_pad=0)_

bkhelifi:
> Does this also have something to do with the warnings we get? Example, below [this plot](https://docs.gammapy.org/dev/tutorials/data/hawc.html#check-the-dataset-we-produced)
> 
> _/home/runner/work/gammapy-docs/gammapy-docs/gammapy/.tox/build_docs/lib/python3.11/site-packages/gammapy/datasets/map.py:2530: UserWarning: This figure includes Axes that are not compatible with tight_layout, so results might be incorrect. plt.tight_layout(w_pad=0)_

`plt.tight_layout` is not used anymore in datasets and maps.

morcuended:
In the dev call, it was decided to revert the changes made in #5992 to adjust the peek plot in VERITAS tutorial and just tweak the figsize of that plot. The changes introduced in #5922 were having other unintended side effects in other plots, and it was overcomplicated for just setting the proper size and aspect for the figure. Closing this.

PR #6005, opened 2025-07-22
Fixing `FluxPoints` docstring

### LINKED ISSUE #6003: FluxPoints documentation showing extra attribute
Dear gammapy team,

The documentation page for the `FluxPoints` class reports that the class has a `table` attribute, however this does not seem to be the case. Something like this returns an attribute error:

```
fp = FluxPoints.from_table(Table.read("data.ecsv"))
fp.table
```

lgreaux:
Dear gammapy team,

This PR fixes  #6003 

I copied the parameters in the docstring of `FluxMaps` to the docstring of `FluxPoints`

### DIFF
--- a/codemeta.json
+++ b/codemeta.json
@@ -359,11 +359,9 @@
         "active"
     ],
     "email": "GAMMAPY-COORDINATION-L@IN2P3.FR",
-
-    "dateModified": "2024-11-29",
-
-  "softwareRequirements": [
-        "numpy>=1.21",
+    "dateModified": "2025-07-21",
+    "softwareRequirements": [
+        "numpy>=1.21,<2.2",
         "scipy>=1.5,!=1.10",
         "astropy>=5.0",
         "regions>=0.5.0",
@@ -374,4 +372,3 @@
         "matplotlib>=3.4, <3.10"
     ]
 }
-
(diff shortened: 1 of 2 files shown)


bkhelifi:
@AtreyeeS: I am not sure that the FluxPoints take as input maps... ;) but rather flux values

AtreyeeS (commented):
Thanks @bkhelifi !
I will check

Astro-Kirsty:
Agree with @bkhelifi on the wording. 
We should also add a specific test for `FluxPoints` being initiated in that manner i.e. `FluxPoints(data)`

QRemy:
> @AtreyeeS: I am not sure that the FluxPoints take as input maps... ;) but rather flux values

It really takes maps as input, the current description is wrong it cannot take table on init only via the `from_table` classmethod and looking at  `from_table` you will see it create a `Maps` dict composed of `RegionNDMap`.

QRemy (approved):
Thanks @lgreaux, the docstring was indeed outdated.
I added a minimal test for the init, and just changed the data description to specify it's a dict of `~gammapy.maps.RegionNDMap` instead of `~gammapy.maps.Map`.

Merging now.

PR #6006, opened 2025-07-22
Add utility function to create a grid of rectangular regions

### LINKED ISSUE #5732: Adding a forward folding estimator for estimating spectral profiles
The current `FluxProfileEstimator` uses backward folding to estimate the flux in different regions.
An alternative way would be to have a forward folding approach - assume a constant spatial model in the whole region, and fit the spectra model to obtain the flux in each region bin.

This has some distinct advantages, eg: allowing for the contributions of additional source models, correctly accounting for the PSF leakage, or creating a spectral index map.

A prototype implementation by @registerrier maybe found here: https://gist.github.com/registerrier/e5bc8a2939807521b1b59dff9f666b89

It can be interesting to expose this to users. We can either

1. Add a new estimator. We already have the `TSMapEstimator` and the `ExcessMapEstimator` implementing forward folding vs unfolding resp.). One issue here might be to to have the correct format for returning the results. Variations in index (or other parameters like the cutoff energy, etc) cannot be accomodated into our existing region based sky maps format. This is similar to the issue we faced for having an estimator for Time Resolved Spectroscopy.
2. Otherwise, we can simply add a notebook exposing this, like we have for the timing case.

Preferences @adonath @QRemy @bkhelifi ?

AtreyeeS:
This PR adds a simple utility function to create a grid of regions similar to `make_orthogonal_sky_regions` and `make_annular_sky_regions`


gammapy/utils/regions.py
```diff
@@ -349,6 +350,47 @@ def make_orthogonal_rectangle_sky_regions(start_pos, end_pos, wcs, height, nbin=
     return regions
 
 
+def make_grid_rectangle_sky_regions(geom, nbinx, nbiny):
+    """Utility function to create a list of grid of rectangular
```
bkhelifi: It is hard to understand the difference with the function `make_orthogonal_rectangle_sky_regions`. Better docstring (and maybe for both) might be needed. One is along an axis and yours in really a grid

bkhelifi (dismissed):
It looks good for me. Maybe a comment on the doctring (optional)

QRemy:
I suggest to delay this new feature to 2.1 so we can finish bugfix and doc PRs.

Astro-Kirsty:
Whilst this is a nice feature and something we should implement ( past 2.0), I am not certain about having to call a `geom` to do it. Another implementation could be better.

AtreyeeS:
@Astro-Kirsty Sure, we can pass a wcs. But usually we always have a geom during the analysis, so this seems the simplest

registerrier (commented):
Thanks @AtreyeeS . It is an interesting additional functionality. I have a few general comments:
- passing a `geom` seems a bit peculiar. Isn't creating the geom if e.g. there is a rotation to be applied already complex? You could also define the rectangle with an upper left, upper right and lower left edges for instance. In which case this would be very close to the longitudinal box function.
- Another possible issue is the possible size distortion from region to region. Does it work for all `WcsGeom`? Or can there be projections that will cause problem?
- With the center, widths nbins and rotation angle, you could create a WCS which pixel centers are directly the center of the boxes. Maybe you would have to assume some projection.

Astro-Kirsty:
> @Astro-Kirsty Sure, we can pass a wcs. But usually we always have a geom during the analysis, so this seems the simplest

I think we should pass the `wcs` as this is in line with the other regions util function `make_orthogonal_rectangle_sky_regions`

registerrier:
What should we do here @AtreyeeS ?

registerrier:
Where do we stand here @AtreyeeS ? Can this be finished soon?

Astro-Kirsty:
We still needed to make a decision here if we want to input the `geom` or `wcs`

AtreyeeS:
Finally addressed comments!
The implementation is a bit complex with the wcs instead of the geom but it works better, specially with the angle.

gammapy/utils/regions.py
```diff
@@ -402,6 +403,71 @@ def make_orthogonal_rectangle_sky_regions(start_pos, end_pos, wcs, height, nbin=
...
+
+    pix_dx = dx.to_pixel(wcs)
+    pix_dy = dy.to_pixel(wcs)
+
+    ux = np.array([pix_dx[0] - pix_center[0], pix_dx[1] - pix_center[1]])
+    uy = np.array([pix_dy[0] - pix_center[0], pix_dy[1] - pix_center[1]])
+    xe = np.linspace(-1, 1, nbinx + 1)
+    ye = np.linspace(-1, 1, nbiny + 1)
+    xc = 0.5 * (xe[:-1] + xe[1:])
+    yc = 0.5 * (ye[:-1] + ye[1:])
```
Astro-Kirsty: I  would be a bit more informative with the names i.e.
```suggestion
    x_edges = np.linspace(-1, 1, nbinx + 1)
    y_edges = np.linspace(-1, 1, nbiny + 1)
    x_centers = 0.5 * (x_edges[:-1] + x_edges[1:])
    y_centers = 0.5 * (y_edges[:-1] + y_edges[1:])
```

Not sure what ux and uy should be?

gammapy/utils/regions.py
```diff
@@ -402,6 +403,71 @@ def make_orthogonal_rectangle_sky_regions(start_pos, end_pos, wcs, height, nbin=
...
+
+            sky_center = SkyCoord.from_pixel(pix_x, pix_y, wcs)
+            reg = RectangleSkyRegion(
+                center=sky_center,
+                width=width / nbinx,
+                height=height / nbiny,
+                angle=angle,
+            )
+            regions.append(reg)
+    return Regions(regions)
```
Astro-Kirsty, line 472: We should just return 'regions' which is inline with the other functionalities

gammapy/utils/regions.py
```diff
@@ -402,6 +403,71 @@ def make_orthogonal_rectangle_sky_regions(start_pos, end_pos, wcs, height, nbin=
...
 
 
+def make_grid_rectangle_sky_regions(
+    center, width, height, wcs, nbinx=1, nbiny=1, angle=0 * u.deg
+):
+    """Utility function to creates a 2D grid of
+    `~regions.RectangleSkyRegion` tiles centered on
+    a given sky coordinate. This is similar to
+    `make_orthogonal_rectangle_sky_regions` but covers a 2-D field
+    instead of a 1-D profile
```
Astro-Kirsty:
```suggestion
    """Utility to return a list of 2D grid of rectangular projections.
    
    This creates a 2D grid of  `~regions.RectangleSkyRegion` tiles centered on
    a given sky coordinate. This is similar to  
    `make_orthogonal_rectangle_sky_regions` but covers a 2D field
    instead of a 1D profile.
```

gammapy/utils/regions.py
```diff
@@ -402,6 +403,71 @@ def make_orthogonal_rectangle_sky_regions(start_pos, end_pos, wcs, height, nbin=
...
+    nbiny : int
+        Number of boxes along y-axis (Dec/latitude)
+    angle : `~astropy.units.Quantity`, optional
+        Rotation angle (in deg) of the grid, anti-clockwise.
+        Default is 0 deg.
+
+    Returns
+    -------
+    regions : list of `~regions.RectangleSkyRegion`
+
```
Astro-Kirsty, line 441:
```suggestion
```

gammapy/utils/regions.py
```diff
@@ -402,6 +403,71 @@ def make_orthogonal_rectangle_sky_regions(start_pos, end_pos, wcs, height, nbin=
...
+        Total angular width of the grid (longitude direction).
+    height : `~astropy.units.Quantity`
+        Total angular height of the grid (latitude direction).
+    wcs : `~astropy.wcs.WCS`
+        WCS projection object used to convert between sky and pixel
+        coordinates.
+    nbinx : int
+        Number of boxes along x-axis (RA/longitude)
+    nbiny : int
+        Number of boxes along y-axis (Dec/latitude)
```
Astro-Kirsty:
```suggestion
    nbinx : int, optional
        Number of boxes along x-axis (RA/longitude). Default is 1.
    nbiny : int, optional
        Number of boxes along y-axis (Dec/latitude). Default is 1.
```

Astro-Kirsty (commented):
Thanks @AtreyeeS. I think this implementation with the wcs is better!
Some minor comments really, plus there are test failures?

gammapy/utils/regions.py
AtreyeeS: having a `Regions` allows, eg `regions.center` directly and keeps all the functionality of having a list of regions

gammapy/utils/regions.py
AtreyeeS: ok

gammapy/utils/regions.py
Astro-Kirsty: Then shouldn't we also have that for the other functions?

gammapy/utils/regions.py
AtreyeeS: ux and uy are unit vectors in the two directions. Adapted now

gammapy/utils/regions.py
AtreyeeS: `compound_region_to_regions`: already a `Regions`.Adapted the others

gammapy/utils/regions.py
```diff
@@ -402,6 +403,72 @@ def make_orthogonal_rectangle_sky_regions(start_pos, end_pos, wcs, height, nbin=
     return regions
 
 
+def make_grid_rectangle_sky_regions(
+    center, width, height, wcs, nbinx=1, nbiny=1, angle=0 * u.deg
+):
+    """Utility function to creates a 2D grid of
+    `~regions.RectangleSkyRegion` tiles centered on
+    a given sky coordinate. This is similar to
```
registerrier: Maybe mention that the tiling is made on the projection plane defined by the WCS given in input.

registerrier (dismissed):
Thanks @AtreyeeS . This looks good. I have just added a small inline comment for the docstring

gammapy/utils/regions.py
Astro-Kirsty: This comments appears not to be implemented

gammapy/utils/regions.py
Astro-Kirsty: This comment has also not been resolved yet

gammapy/utils/regions.py
```diff
@@ -420,8 +488,8 @@ def make_concentric_annulus_sky_regions(
 
     Returns
     -------
-    regions : list of `~regions.CircleAnnulusSkyRegion`
-        Regions in which the profiles are made.
+    regions : `~regions.Regions`
+        list of `~regions.CircleAnnulusSkyRegion` in which the profiles are made.
```
Astro-Kirsty:
```suggestion
        List of `~regions.CircleAnnulusSkyRegion` in which the profiles are made.
```

registerrier (approved):
Thanks @AtreyeeS . Looks good!

PR #6007, opened 2025-07-23
Fix doc uniformprior

### LINKED ISSUE #5988: Non-coherent documentation for UniformPrior class
Dear gammapy team,

**gammapy 1.3, gammapy dev**
The documentation for the UniformPrior class is not coherent with the behavior of the class: it is said to return 1 if the parameter value is between min and max, while the opposite behavior is implemented.

lgreaux:
Dear gammapy team,

This PR fixes #5988

I exchanged 0 and 1 in the docstring of `UniformPrior` to match the values currently returned by `evaluate`

### DIFF
--- a/gammapy/modeling/models/prior.py
+++ b/gammapy/modeling/models/prior.py
@@ -171,8 +171,8 @@ class GaussianPrior(Prior):
 class UniformPrior(Prior):
     """Uniform Prior.
 
-    Returns 1 if the parameter value is in (min, max).
-    0, if otherwise.
+    Returns 0 if the parameter value is in (min, max).
+    1, if otherwise.
 
     Parameters
     ----------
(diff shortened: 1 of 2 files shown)


gammapy/modeling/models/prior.py
```diff
@@ -171,8 +171,8 @@ def _inverse_cdf(self, value):
 class UniformPrior(Prior):
     """Uniform Prior.
 
-    Returns 1 if the parameter value is in (min, max).
-    0, if otherwise.
+    Returns 0 if the parameter value is in (min, max).
```
bkhelifi:
```suggestion
    Returns 0 if the parameter value is in ]min, max[.
```

gammapy/modeling/models/prior.py
```diff
@@ -171,8 +171,8 @@ def _inverse_cdf(self, value):
...
+    1, if otherwise.
```
Astro-Kirsty:
```suggestion
    Returns 1, if otherwise.
```

bkhelifi (approved):
Good for me. Thanks @lgreaux

Issue #6011, opened 2025-07-25
Dependencies not properly installed in jupytext workflow


morcuended:
See https://github.com/gammapy/gammapy/actions/runs/16513023222/job/46698402243

It misses the installation of gammapy and jupytext

PR #6012, opened 2025-07-25
Install gammapy and jupytext in check notebooks action

morcuended:
Solves #6011


Astro-Kirsty:
Thanks @morcuended. But can you open another PR as this one has 157 commits :o

PR #6014, opened 2025-07-25
Fix discrepancy in exposure maps unit  attached to edsip and psf from FermipyDatasetsReader

QRemy:
Fix discrepancy in exposure maps units  attached to edsip and psf from the FermipyDatasetsReader.
This should ensure that stacking is performed correctly as discussed on slack.

### DIFF
--- a/gammapy/datasets/tests/test_io.py
+++ b/gammapy/datasets/tests/test_io.py
@@ -229,7 +229,7 @@ def test_fermipy_datasets_reader():
         datasets[0]._psf_kernel.psf_kernel_map.data.sum(axis=(1, 2)), 1, rtol=1e-5
     )
     assert_allclose(
-        datasets[0].edisp.exposure_map.data, datasets[0].psf.exposure_map.data
+        datasets[0].edisp.exposure_map.quantity, datasets[0].psf.exposure_map.quantity
     )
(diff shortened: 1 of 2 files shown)


gammapy/datasets/tests/test_io.py
```diff
@@ -229,7 +229,10 @@ def test_fermipy_datasets_reader():
         datasets[0]._psf_kernel.psf_kernel_map.data.sum(axis=(1, 2)), 1, rtol=1e-5
     )
     assert_allclose(
-        datasets[0].edisp.exposure_map.data, datasets[0].psf.exposure_map.data
+        datasets[0].edisp.exposure_map.quantity, datasets[0].psf.exposure_map.quantity
     )
+    assert datasets[0].edisp.exposure_map.unit == datasets[0].psf.exposure_map.unit
```
AtreyeeS, line 234: If you test on the quantity, that should include the units, no?

AtreyeeS (approved):
Thanks @QRemy ! Good from my side

gammapy/datasets/tests/test_io.py
QRemy: We can have equivalent quantity with different units, I wanted to check that the unit of all the maps are the same that's not mandatory just for consistency.

PR #6015, opened 2025-07-25
Update minimum python version supported mentioned in howto dev documentation

morcuended:


registerrier:
Thanks @morcuended . 
I think it would be more informative to write that we follow the NEP 29 guidelines of a “time window-based” policy for support of Python and NumPy versions.

PR #6016, opened 2025-07-25
Fix for reco PSF in create_map_dataset_from_dl4

QRemy:
Fix for reco PSF in create_map_dataset_from_dl4

### DIFF
--- a/gammapy/datasets/utils.py
+++ b/gammapy/datasets/utils.py
@@ -253,7 +253,13 @@ def create_map_dataset_from_dl4(data, geom=None, energy_axis_true=None, name=Non
 
     # ensure that DL4 IRFs have the axes
     rad_axis = data.psf.psf_map.geom.axes["rad"]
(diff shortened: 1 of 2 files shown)


gammapy/datasets/utils.py
```diff
@@ -253,7 +253,13 @@ def create_map_dataset_from_dl4(data, geom=None, energy_axis_true=None, name=Non
 
     # ensure that DL4 IRFs have the axes
     rad_axis = data.psf.psf_map.geom.axes["rad"]
-    geom_psf = data.psf.psf_map.geom.to_image().to_cube([rad_axis, energy_axis_true])
+    if data.psf.energy_name == "energy":
+        geom_psf = data.psf.psf_map.geom.to_image().to_cube([rad_axis, energy_axis])
```
AtreyeeS, line 257: So the HAWC files are with reco energy in psf?
This will work correctly for our MapEvaluator?

QRemy: yes

AtreyeeS (approved):
Thanks @QRemy

PR #6022, opened 2025-07-29
Add random_variable  property on prior

QRemy:
Add `random_variable`  property on Priors.
It's  a small cleanup but il also allows to compute all the prior properties directly for example `random_variable.entropy()` .

### DIFF
--- a/gammapy/modeling/models/prior.py
+++ b/gammapy/modeling/models/prior.py
@@ -79,7 +79,7 @@ class Prior(ModelBase):
 
     def _inverse_cdf(self, value):
         """Return inverse CDF for prior."""
-        return self.random_variable.ppf(value)
+        return self._random_variable.ppf(value)
 
     @property
     def weight(self):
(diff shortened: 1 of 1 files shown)


gammapy/modeling/models/prior.py
```diff
@@ -162,10 +166,16 @@ def evaluate(value, mu, sigma):
...
+    @property
+    def random_variable(self):
+        """Return random variable object for prior."""
+        return norm(self.mu.value, self.sigma.value)
+
+
+@property
+def loguniform_rv(self):
+    """Return random variable object for prior."""
+    return loguniform(self.min.value, self.max.value)
```
QRemy:
```suggestion
```

QRemy:
> Thanks @QRemy . Looks nice. I have a few small comments.
> 
> * I would make random_variable private.

Done

> * One possibly problematic issue is the mutability of `PriorParameter` if `_random_variable` is a property. In general, I think `PriorParameter` should be immutable contrary to the regular `Parameter` objects. In general, most of the functionality coming from the inheritance is not useful in this case. It might be worth simplifying at some point.

I think the motivation for the inheritance was partly to re-use the serialisation of Parameters

PR #6023, opened 2025-07-29
Restructure gammapy.modeling.selection

QRemy:
Restructure gammapy.modeling.selection to be a subfolder and selection.py is renamed nested.py. So future PRs could include also the Bayesian model selection in this folder.

### DIFF
--- a/gammapy/estimators/energydependentmorphology.py
+++ b/gammapy/estimators/energydependentmorphology.py
@@ -6,7 +6,7 @@ from gammapy.datasets import Datasets
 from gammapy.modeling import Fit
 from gammapy.modeling.models import FoVBackgroundModel, Models
 from gammapy.datasets.actors import DatasetsActor
-from gammapy.modeling.selection import TestStatisticNested
+from gammapy.modeling.selection import NestedModelSelection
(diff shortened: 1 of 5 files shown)


bkhelifi:
It seems fair by anticipation... As one should rename a Class, should we not have a depreciation warning, or something like that?

facero:
Hi
I think `TestStatisticNested` was not used widely so few users might have an issue with the renaming. 
But indeed a warning would be nice

facero:
Btw was there a implementation of this `TestStatisticNested` in a tutorial ? (I think not)
Myabe for 2.1 it would be good to showcase an example of Model Selection with sampler like the things Quentin is developing. But this would need some time and explanation probably

QRemy:
`TestStatisticNested` is not part of the public api so in principle it can be renamed without issue. Only the `select_nested_models ` is shown and it's still the case with this PR.

registerrier:
> It seems fair by anticipation... As one should rename a Class, should we not have a depreciation warning, or something like that?

This class was never made public, just the `select_nested_model` function. So technically it is an implemntation detail so far.

PR #6026, opened 2025-07-29
Fix the call to tutorials/api

Astro-Kirsty:
Oops, missed a number of changes for the `api` to `details` conversion


morcuended:
Side note, I think the linkcheck action is not working currently, even if it appears green. See https://github.com/gammapy/gammapy/actions/runs/16496936503/job/46644859507

Astro-Kirsty:
> Side note, I think the linkcheck action is not working currently, even if it appears green. See https://github.com/gammapy/gammapy/actions/runs/16496936503/job/46644859507

Hmm okay, I can try to fix that
(I was running the link checks locally!)

PR #6029, opened 2025-07-29
Standardized implementation prior_stat_sum

### LINKED ISSUE #5989: Non standardized implementation of the Prior classes
Dear gammapy team,

The three currently implemented Prior classes are not being evaluated in compatible ways: 

- `LogUniformPrior`: evaluate returns `-2 * log(pdf(value))`
- `GaussianPrior`: evaluate returns `-2 * log(pdf(value)) - 2 * log(sigma * sqrt(2 pi))`
- `UniformPrior`: evaluate returns neither something linked with `pdf(value)` nor with `log(pdf(value))`

As log probabilities are equivalent up to an additive constant, the behavior of the `GaussianPrior` and the `LogUniformPrior` can be seen as technically compatible, but not conceptually equivalent. However the `UniformPrior` is currently deviating from both. I understand that such an implementation can be useful for penalized likelihoods, however I do believe that it can be very misleading for users. I can two different ways forward, depending on which philosophy fits gammapy best:

1. Consider the `Prior` class as a representation of the actual probability distribution corresponding to the prior. This would mean evaluating the pdf of the distribution, and the computation of the likelihood would need to take the log of the computed values. The `prior_stat_sum` method of the `Parameter` class could simply return `-2 * log(prior)`, keeping the coherence with the log likelihoods that are used.

2. Consider the `Prior` class as computation tools for -2 times the logarithm of the prior. The `GaussianPrior` and `LogUniformPrior` could then be kept as is (defined up to an additive constant), but the `UniformPrior` class would need to return `0` inside the bounds, and ` infinity` outside.

In both cases, I believe that the penalized likelihood usage of the `UniformPrior` should not be the default behavior. I would suggest either implementing a new class inheriting from `Prior` to do represent penalized likelihoods, or at the very least making it an optional behavior for the `UniformPrior` class, that experienced users can choose to use if they want to. 

I would be very happy to get your feedback on this, and would gladly help with the implementation of these changes.

QRemy:
Resolves #5989
-`UniformPrior` and `GaussianPrior` now return values in -2log(prior)
-A `CompoundUniformPrior` is introduce to support the penalisation use case.

### DIFF
--- a/examples/tutorials/details/priors.py
+++ b/examples/tutorials/details/priors.py
@@ -40,9 +40,9 @@ results.
 
 The tutorial addresses three examples:
 
-1. Including prior information about the sources index
-2. Encouraging positive amplitude values
-3. How to add a custom prior class?
+-  Including prior information about the sources index
+-  How to add a custom prior class?
+-  Encouraging positive amplitude values
 
 In the first example, the Gaussian prior is used. It is shown how to set
 a prior on a model parameter and how it modifies the fit statistics. A


QRemy:
-removed the newly introduced prior `CompoundUniformPrior`  as it was too complex to explain
-added `UniformPenalty` to reproduce the old behaviour of the  `UniformPrior ` but including explicitly a penalty parameter instead of using the weights and raise an error if trying to access the `random_variable`.
-Adapted the notebook to include the terms of Maximum a posteriori and penalized likelihood.

QRemy:
- removed `UniformPenalty ` from the code base and show it only at the end of the notebook.

adonath (approved):
Thanks a lot @QRemy, looks good to me. The example with the uniform penalty is illustrative and keeps the consistency!

PR #6031, opened 2025-07-30
Exposing `AnalysisConfig` better

### LINKED ISSUE #5181: Lack of documentation about AnalysisConfig option identifiers
I was following the tutorial here
https://docs.gammapy.org/1.2/tutorials/starting/analysis_1.html

and I had to go into the source code to know e.g. which are the allowed string identifiers for the
`background` subsection.

For example in this case, going backwards from the class I was using, I traced back the answer to my doubt:

`AnalysisConfig` --> `BackgroundConfig` --> `BackgroundMethodEnum`

which allowed me to understand the possible choices are `reflected`, `fov_background` or `ring`.

Apologies if I have overlooked a clearer source of information, but if I am right, then at least the API reference should be updated with more classes (to start, all those used by `AnalysisConfig` in order to make clear which are all the supported options to use when filling an analysis configuration file).

Astro-Kirsty:
This resolves #5181.

I've added some changes to the conf.py such that the pydantic related code looks nicer. 
You can see an obvious difference already between the previous output and this one

As we don't have many of these functionalities exposed, it is unclear how to proceed. 

We could add more of them to the documentation? Thoughts?

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -86,7 +86,7 @@ source_suffix = '.rst'
 master_doc = 'index'
 
 # Allow to add the canonical html flag on all the pages
-# This permits to precise to web search engine that the pages for the stable version are privileged
+# This allows you to indicate to the web search engine that the pages for the stable version are privileged
 html_baseurl = 'https://docs.gammapy.org/stable/'
 
 # The reST default role (used for this markup: `text`) to use for all
@@ -155,6 +155,10 @@ automodapi_toctreedirnm = 'api'
 
 # Suppress member summaries
 numpydoc_show_class_members = False
+
+# autodoc options to make the AnalysisConfig look nice
+autodoc_class_signature = "separated"
+autodoc_typehints = "description"
 
 # Ensures that when users click the "Copy" button, only the actual code is copied,
 # excluding interactive prompts and indentation markers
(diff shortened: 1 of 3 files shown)


QRemy (commented):
Thanks @Astro-Kirsty. 
I wouldn't  write the name of the classes it configures here so people don't click on that link instead of the corresponding config class.
I would rather add the detail of which estimator it configures in the description of the specific configs.  Do you plan to add them in this PR or separate ones ?

Astro-Kirsty:
Okay thanks @QRemy. 
Happy to add details to each of the specific configs but they are not exposed in the documentation, so this would still not be helpful to users. We should make the decision to expose them first

Astro-Kirsty:
The decision from the dev call is to expose the other `config`s. This will be done in this PR

Astro-Kirsty:
Should I add examples to each of the functionalities exposed?

QRemy:
> Should I add examples to each of the functionalities exposed?

I would say no, we have a full example config files for 1d and 3d analysis exposed somewhere right ?

Astro-Kirsty:
We do, but at least for the 1d maybe it is a bit hidden!

QRemy (approved):
Thanks @Astro-Kirsty, looks great !

PR #6039, opened 2025-07-31
Adapt CTAO tutorial with xmltodict

### LINKED ISSUE #6032: Should we remove or keep extern?
We removed the `extern` part of gammapy in #6018, as per the discussion in the dev call

We actually have a call to that in this tutorial:
https://docs.gammapy.org/dev/tutorials/data/cta.html#source-models

Should we remove that from the tutorial? Or was it not correct to remove it all together in the first place?

Astro-Kirsty:
This closes #6032

I change the import to be independent of gammapy, then I extend the example a little to include the entire sky model as is written in the description, thereby removing the TODO

### DIFF
--- a/examples/tutorials/data/cta.py
+++ b/examples/tutorials/data/cta.py
@@ -321,10 +321,14 @@ plt.show()
 # for a given source, and convert it into the sky model format Gammapy
 # understands.
 #
-
 # Uncomment this if you have installed `xmltodict`.
+
+
 # This is what the XML file looks like
 # !tail -n 20 $CTADATA/models/models_gps.xml
+
+######################################################################
+# Here is an example on how to create a gammapy source model from this
 
 # import xmltodict
 # from astropy import units as u
@@ -363,8 +367,8 @@ plt.show()
 # # Create the spatial model
 # spatial_params = spatial0["parameter"]
 # spatial_model = PointSpatialModel(
-#     lon=get_param(spatial_params, "RA") * u.deg,
-#     lat=get_param(spatial_params, "DEC") * u.deg,
+#     lon_0=get_param(spatial_params, "RA") * u.deg,
+#     lat_0=get_param(spatial_params, "DEC") * u.deg,
 #     frame="icrs"
 # )
 # print(spatial_model)
--- a/examples/tutorials/data/cta.py
+++ b/examples/tutorials/data/cta.py
@@ -321,10 +321,10 @@ plt.show()
 # for a given source, and convert it into the sky model format Gammapy
 # understands.
 #
-# Uncomment this if you have installed `xmltodict`.
-
-
-# This is what the XML file looks like
+# With the following command we can see what the XML file looks like
+# (uncomment it if you have installed `xmltodict`).
+ 
+ 
 # !tail -n 20 $CTADATA/models/models_gps.xml
 
 ######################################################################
--- a/examples/tutorials/data/cta.py
+++ b/examples/tutorials/data/cta.py
@@ -306,47 +306,72 @@ plt.show()
 # Source models
 # -------------
 #
-# The 1DC sky model is distributed as a set of XML files, which in turn
-# link to a ton of other FITS and text files. Gammapy doesn’t support this
-# XML model file format. We are currently developing a YAML based format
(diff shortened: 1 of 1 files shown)


examples/tutorials/data/cta.py
```diff
@@ -306,47 +306,72 @@
...
+#     index=par_to_val(spectrum0[1]),
+#     reference=par_to_val(spectrum0[2]) * u.Unit("MeV"),
+# )
+# print(spectral_model)
+
+# # Create the spatial model
+# spatial_params = spatial0["parameter"]
+# spatial_model = PointSpatialModel(
+#     lon=get_param(spatial_params, "RA") * u.deg,
+#     lat=get_param(spatial_params, "DEC") * u.deg,
```
QRemy:
```suggestion
#     lon_0=get_param(spatial_params, "RA") * u.deg,
#     lat_0=get_param(spatial_params, "DEC") * u.deg,
```

examples/tutorials/data/cta.py
```diff
@@ -306,47 +306,72 @@
...
-# `xmltodict <https://github.com/martinblech/xmltodict>`__ if you
+# `xmltodict <https://github.com/martinblech/xmltodict>`__ installed through
 # `pip install xmltodict`. Here’s an example how to load the information
-# for a given source, and to convert it into the sky model format Gammapy
+# for a given source, and convert it into the sky model format Gammapy
 # understands.
 #
 
+# Uncomment this if you have installed `xmltodict`.
 # This is what the XML file looks like
```
QRemy: this should be a markdown block

examples/tutorials/data/cta.py
```diff
@@ -306,47 +306,72 @@
...
 # !tail -n 20 $CTADATA/models/models_gps.xml
 
-# TODO: write this example!
+# import xmltodict
```
QRemy: you should have another markdown block before this code block saying: "Here is an example on how to create a gammapy source model from this"

examples/tutorials/data/cta.py
Astro-Kirsty: I'm not sure what you mean? 
This is a click copy block like [here](https://docs.gammapy.org/dev/tutorials/details/nested_sampling_Crab.html#comparing-the-posterior-distribution-of-all-runs)

examples/tutorials/data/cta.py
```diff
@@ -306,47 +306,72 @@
...
 # from the Python standard library, or
-# `xmltodict <https://github.com/martinblech/xmltodict>`__ if you
+# `xmltodict <https://github.com/martinblech/xmltodict>`__ installed through
 # `pip install xmltodict`. Here’s an example how to load the information
-# for a given source, and to convert it into the sky model format Gammapy
+# for a given source, and convert it into the sky model format Gammapy
 # understands.
 #
 
+# Uncomment this if you have installed `xmltodict`.
```
Astro-Kirsty:
```suggestion
# Uncomment this if you have installed `xmltodict`.


```

examples/tutorials/data/cta.py
Astro-Kirsty, line 333:
```suggestion
######################################################################
# Here is an example on how to create a gammapy source model from this

# import xmltodict
```

examples/tutorials/data/cta.py
QRemy: the line below also

examples/tutorials/data/cta.py
```diff
@@ -306,47 +306,76 @@
...
+# `xmltodict <https://github.com/martinblech/xmltodict>`__ installed through
 # `pip install xmltodict`. Here’s an example how to load the information
-# for a given source, and to convert it into the sky model format Gammapy
+# for a given source, and convert it into the sky model format Gammapy
 # understands.
 #
+# Uncomment this if you have installed `xmltodict`.
+
 
 # This is what the XML file looks like
```
QRemy:
```suggestion
# With the following command we can see what the XML file looks like
# (uncomment it if you have installed `xmltodict`).
 
 
```

QRemy (approved):
Thanks @Astro-Kirsty, looks good, merging now.

PR #6043, opened 2025-08-01
Fix docstrings

Astro-Kirsty:
The initial purpose of this PR was to add the `datasets_names` to the docstring of `create_fermi_isotropic_diffuse_model`, but then I also noticed some other docstring faults

Note: the `read_hermes_cube` is included twice in the API list as it is already there from including `utils`, therefore we don't need to specify it in the main file


gammapy/modeling/models/__init__.py
```diff
@@ -66,7 +66,6 @@
     TemplatePhaseCurveTemporalModel,
     TemporalModel,
 )
-from .utils import read_hermes_cube
```
QRemy: this will prevent to load it as `from gammapy.modeling.models import read_hermes_cube`

I think we should rather remove utils from modeling.rst, we don't have it for the other subpacakges

```
.. automodapi:: gammapy.modeling.models.utils
    :no-inheritance-diagram:
    :include-all-objects:

```

Astro-Kirsty: Given that we only expose one function through this way I will remove it from `modeling.rst`.
But we do use this method for some other groups

AtreyeeS (approved):

gammapy/modeling/models/spectral_crab.py
```diff
@@ -53,6 +53,11 @@ def create_crab_spectral_model(reference="meyer"):
     reference : {'meyer', 'hegra', 'hess_pl', 'hess_ecpl', 'magic_lp', 'magic_ecpl'}, optional
         Which reference to use for the spectral model. Default is 'meyer'.
 
+    Returns
+    -------
+    `~gammapy.modeling.models.SpectralModel`
+        Spectral model (for all-sky cosmic ray flux).
```
morcuended: > (for all-sky cosmic ray flux)

This should be: `Crab Nebula spectral model.` or something like that

Astro-Kirsty:
```suggestion
        Spectral model (for the crab).
```

morcuended (approved):
Thanks @Astro-Kirsty

PR #6044, opened 2025-08-01
Mention NEP29 guideline for python and numpy support in dev docs

morcuended:
As suggested by @registerrier here https://github.com/gammapy/gammapy/pull/6015#issuecomment-3127734692, I mention NEP 29 guidelines for Python and NumPy support, as described in https://numpy.org/neps/nep-0029-deprecation_policy.html


registerrier (approved):
Thanks @morcuended . Looks good to me!

PR #6053, opened 2025-08-05
Add mask for PointSkyRegion in geom.region_mask

AtreyeeS:
This gives a cleaner solution to #5976 as suggested by @QRemy

### DIFF
--- a/gammapy/maps/wcs/geom.py
+++ b/gammapy/maps/wcs/geom.py
@@ -993,10 +993,12 @@ class WcsGeom(Geom, CacheInstanceMixin):
         """
         from gammapy.maps import Map, RegionGeom
         from gammapy.modeling.models import PointSpatialModel
+        from ..region.geom import _parse_regions
 
         if not self.is_regular:
             raise ValueError("Multi-resolution maps not supported yet")
 
+        regions = _parse_regions(regions)
         geom = RegionGeom.from_regions(regions, wcs=self.wcs)
         idx = self.get_idx()
         mask = geom.contains_wcs_pix(idx)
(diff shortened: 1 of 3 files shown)


gammapy/maps/wcs/geom.py
```diff
@@ -991,13 +992,19 @@ def region_mask(self, regions, inside=True):
         since this method expects a list of regions.
         """
         from gammapy.maps import Map, RegionGeom
+        from gammapy.modeling.models import PointSpatialModel
 
         if not self.is_regular:
             raise ValueError("Multi-resolution maps not supported yet")
```
QRemy, line 1000:
```
you will need part of the code that is in `RegionGeom.from_regions` before, could be moved in a helper function an re-used here:

def _parse_regions(regions):
        if isinstance(regions, str):
            regions = Regions.parse(data=regions, format="ds9")
        elif isinstance(regions, SkyRegion):
            regions = [regions]
        elif isinstance(regions, SkyCoord):
            regions = [PointSkyRegion(center=regions)]
        return regions

File: gammapy/maps/wcs/tests/test_geom.py [hunk 2]
```diff
@@ -424,14 +424,15 @@ def test_region_mask():
 
     r1 = CircleSkyRegion(SkyCoord(0, 0, unit="deg"), 1 * u.deg)
     r2 = CircleSkyRegion(SkyCoord(20, 20, unit="deg"), 1 * u.deg)
-    regions = [r1, r2]
+    r3 = PointSkyRegion(SkyCoord(2, -2, unit="deg"))
+    regions = [r1, r2, r3]
```
QRemy: better keep the previous regions as they were 
but another regions list below with only one point source and check that mask.data.sum==4

AtreyeeS: Good catch! The `contains_wcs_pix` always returns True if there is only `PointSkyRegion`.
So for a single point source, the whole mask is set to True :-/
Adapted  `contains_wcs_pix` function accordingly

QRemy: I pushed a modified version that doesn't require to adapt `contains_wcs_pix`

PR #6054, opened 2025-08-05
Correct errors in docs build

### LINKED ISSUE #6052: Change deprecated `pkg_resources` call in docs/conf.py
When building the documentation, there is this deprecation message:

```python
gammapy/docs/conf.py:24: UserWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html. The pkg_resources package is slated for removal as early as 2025-11-30. Refrain from using this package or pin to Setuptools<81.
  from pkg_resources import get_distribution
```

I think the same info can be got from `gammapy.__version__`

Astro-Kirsty:
This PR fixes #6052, as well as fixing a number of other warnings that occur in the documentation build

### DIFF
--- a/docs/getting-started/index.rst
+++ b/docs/getting-started/index.rst
@@ -62,10 +62,7 @@ Then, there are two methods to quickly install Gammapy.
             Learn more
 
 
-.. toctree::
-   :maxdepth: 1
-
-   quickstart
+.. include:: quickstart.rst
 
 
 Tutorials Overview


morcuended (approved):
Thanks @Astro-Kirsty. Looks good to me

AtreyeeS (approved):
Thanks @Astro-Kirsty

PR #6056, opened 2025-08-05
Add `make_counts_off_rad_max` to `__all__`

aaguasca:
**Description**

This pull request adds `make_counts_off_rad_max` to `__all__` in makers/utils.py. 

**Dear reviewer**

The PR is ready for review.

### DIFF
--- a/gammapy/makers/utils.py
+++ b/gammapy/makers/utils.py
@@ -16,6 +16,7 @@ from gammapy.utils.coordinates import FoVICRSFrame, FoVAltAzFrame
 from gammapy.utils.regions import compound_region_to_regions
 
 __all__ = [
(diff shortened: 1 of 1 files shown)


Astro-Kirsty:
@registerrier, @adonath  is there a reason this is not currently exposed in our API?

morcuended (approved):
> @registerrier, @adonath is there a reason this is not currently exposed in our API?

I see no reason not to. Thanks @aaguasca

QRemy (approved):
Thanks @aaguasca, for consistency we can also expose this.
At some point we discussed to not expose these make functions as in principle the users can interact only with the Maker classes but I think we kept them so the core of the code is visible from the documentation.

PR #6058, opened 2025-08-06
Adapt the docstring for PSF containment

### LINKED ISSUE #4299: Exposing base classes to make source code available in gammapy doc?
**Gammapy version**
1.0 and dev

**Bug description**
I wanted to get the value of the containment radius of the PSF for a given energy. So I explored the doc of gammapy to find the right methods to do that. I found that `containment_radius`gives me exactly what I need. However, since `containment_radius` is coded in `gammapy.irf.psf.core.PSF`I cannot access the source code from the gammapy doc.
Of course in my case I went to the GitHub rep to see it. But I think it can be an issue because some of these useful methods are only coded there. 
Moreover, for this method, the handling of `**kwargs`is not really explicit, which can make it difficult for the user to know what can be passed through. I think that this method is not the only one in this case.

I also suggest that we put an example for what I wanted to do in the `containment_radius` methods since it would seems really common to want to get the value of the PSF at given coordinates.

Astro-Kirsty:
This fixes #4299

In the end, we do not want to expose the `PSF` base class in the documentation, so it won't be possible to see the source code for this function. Instead I add a sentence which says how to know the options for **kwargs as well as adding a simple example.

### DIFF
--- a/gammapy/irf/psf/core.py
+++ b/gammapy/irf/psf/core.py
@@ -26,6 +26,7 @@ class PSF(IRF):
             Rad value.
         **kwargs : dict
             Other coordinates.
+            You can view the available options through "psf.axes.names".
 
         Returns
         -------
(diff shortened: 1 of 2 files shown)


QRemy (approved):
Thanks @Astro-Kirsty looks good.

Issue #6042, opened 2025-07-31
Implement scheduled nightly build of wheels


morcuended:
Implement a daily/weekly workflow that build Gammapy wheels using openastronomy workflow. See https://github.com/gammapy/gammapy/pull/5708, without the publishing part.

For reference, Gammapy could also follow this https://scientific-python.org/specs/spec-0004/

PR #6061, opened 2025-08-06
Workflow to check the building of wheels nightly

morcuended:
As proposed in #5708 and #6042 here is a workflow to build the wheels nightly and catch any potential issue in the process early.

**Description**

This workflow makes use of `OpenAstronomy/github-actions-workflows/.github/workflows/publish.yml` but removing the publishing part to PyPI. Wheels are available as artifacts. See e.g. https://github.com/morcuended/gammapy/actions/runs/16780075849

Later, this could be extented with https://scientific-python.org/specs/spec-0004/

### DIFF
--- a/.github/workflows/nightly_builds.yml
+++ b/.github/workflows/nightly_builds.yml
@@ -1,0 +1,16 @@
+name: Nightly wheel building
+
+on:
+  workflow_dispatch:
+  schedule:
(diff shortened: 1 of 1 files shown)


registerrier (approved):
Thnaks @morcuended . Looks good to me.

Issue #3844, opened 2022-03-09
Better use of fixtures in gammapy tests 


luca-giunti:
Reminder to make a better use of fixtures, which allow to shorten and speed up the tests. Here are a few ideas.

1.In many tests (e.g. [here](https://github.com/gammapy/gammapy/blob/25f8aa20e263c7665d6a848c481e28b4b6b245f9/gammapy/estimators/points/tests/test_sed.py#L25), [here](https://github.com/gammapy/gammapy/blob/25f8aa20e263c7665d6a848c481e28b4b6b245f9/gammapy/estimators/points/tests/test_sensitivity.py#L10) and [here](https://github.com/gammapy/gammapy/blob/25f8aa20e263c7665d6a848c481e28b4b6b245f9/gammapy/estimators/points/tests/test_profile.py#L20)) the `spectrum_dataset` fixture defined in `gammapy/conftest.py` could be used instead of simulating/creating new datasets. Perhaps this could work also for onoff datasets using `SpectrumDatasetOnOff.from_spectrum_dataset(spectrum_dataset)`
2. Also for `MapDataset` and other frequently used objects, we could define a fixture under `gammapy/conftest.py` and always reuse it
3. We could speed up the tests by changing the scope of the dataset fixtures to `"session"`, so that they are created only once (see related [TODO](https://github.com/gammapy/gammapy/blob/25f8aa20e263c7665d6a848c481e28b4b6b245f9/gammapy/conftest.py#L66)). But for this to work all tests should be adapted in such a way that they don't modify the datasets in place. I suspect this would be a painful PR

If you agree with some of these ideas, I could work on them.

adonath:
Thanks @luca-giunti, I agree this would be very nice clean up work for the tests! In terms of code organisation we could maybe introduce a dedicated `fixture.py` file per sub-package and then import all into the global `conftest.py` namespace.

luca-giunti:
Good idea! When I find the time I can perhaps work on this

adonath:
@registerrier I think the structure of the tests is not a priority for v1.0. However it should be after v1.0, so we will establish the API and feature completeness for v1.0 and then we improve the tests against it, before we move on with adding significant new features...

Issue #94, opened 2014-03-03
Increase gammapy test coverage


cdeil:
Currently there are many gammapy functions / methods that are not covered by the existing unit tests.
The current status can also be seen here:
https://coveralls.io/r/gammapy/gammapy

Any and all help in adding tests (and probably fixing bugs and improving docstrings along the way) would be much appreciated!

You can create a gammapy coverage report and view it locally:

```
cd gammapy
python setup.py test -V --coverage
open htmlcov/index.html
```

I'll keep this issue open for a while until the situation has improved significantly.

cdeil:
I'm closing this old issue from 2014.
Cleanup in Gammapy and improving code and tests is ongoing and will be for a while.
I just don't think this generic reminder issue will help.

cdeil:
I'm re-opening this general reminder issue, to increase test coverage in Gammapy.

Things are much better now, but we still have a few 100 lines / a few 10 methods that are completely untested. Many of them contain bugs, and generally untested code slows us down when refactoring in Gammapy.

The goal isn't 100%, but to spend a few days to get to v1.0 quality.
@adonath @registerrier - How about one day in the next 1-2 weeks, where we do this together in a Google hangout / co-working hour, and invite anyone that wants to help to join?
Doesn't just have to be next tests, generally: spend one day to improve Gammapy tests.

cdeil:
I'm adding "high-priority" here. 

Note that we now have a nice overview for coverage: https://codecov.io/gh/gammapy/gammapy

We should add a test, or remove, code like [here](https://codecov.io/gh/gammapy/gammapy/src/d4c7c9224c0e03023d7f2281ed20c55ab8872862/gammapy/cube/fit.py#L556...570) or [here](https://codecov.io/gh/gammapy/gammapy/src/d4c7c9224c0e03023d7f2281ed20c55ab8872862/gammapy/cube/background.py#L105...117) or in 10 - 20 other places.

Ideally the guilty parties would spend a day and improve coverage for code they added. But adding a test to improve coverage is also a good task for new people, and pair coding, to teach at the coding sprint next week.

adonath:
Just discussed at the dev meeting:
- Specifically we need parameterized testing of flux estimators with multiple models, to avoid issues like #3690 
- Improve test coverage for fitting with shared / fixed / linked parameters
- Improve serialization test for shared model components (SkyModel objects that share a spectral or spatial component)

PR #6069, opened 2025-08-08
Remove TODO

Astro-Kirsty:
This removes a couple TODO that are not necessary.

- we can add more required columns as we see fit, this is a TODO from 4 years ag0
- `[gammapy/irf/background` not a TODO just a statement
- the function that it mentions no longer exists

I also saw this one:
https://github.com/gammapy/gammapy/blob/445fd4f59de8a98a1b9db4fdb05443653995c100/gammapy/estimators/points/tests/test_sed.py#L46

Should I create a test dataset for this in gammapy-data and remove this section of code ?

### DIFF
--- a/gammapy/estimators/map/core.py
+++ b/gammapy/estimators/map/core.py
@@ -41,7 +41,6 @@ REQUIRED_COLUMNS = {
     "e2dnde": ["e_ref", "e2dnde"],
     "flux": ["e_min", "e_max", "flux"],
     "eflux": ["e_min", "e_max", "eflux"],
-    # TODO: extend required columns
     "likelihood": [
         "e_min",
         "e_max",
--- a/gammapy/irf/background.py
+++ b/gammapy/irf/background.py
@@ -65,9 +65,6 @@ class BackgroundIRF(IRF):
         bkg : `Background2D` or `Background2D`
             Background IRF class.
         """
-        # TODO: some of the existing background files have missing HDUCLAS keywords
-        #  which are required to define the correct Gammapy axis names
-
         if "HDUCLAS2" not in table.meta:
             log.warning("Missing 'HDUCLAS2' keyword assuming 'BKG'")
             table = table.copy()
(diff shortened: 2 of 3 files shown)


gammapy/irf/background.py
```diff
@@ -65,9 +65,6 @@ def from_table(cls, table, format="gadf-dl3"):
         bkg : `Background2D` or `Background2D`
             Background IRF class.
         """
-        # TODO: some of the existing background files have missing HDUCLAS keywords
```
Tobychev, line 68: Agree that it's unclear what, precisely, the author thinks should be done, maybe the intention was to have someone fix the test files that lacked the HDUCLAS keyword?

Alternatively it is actually just a comment explaining why Gammapy carries on reading the file even as a a keyword is missing; the code could equally well just not accept files without this keyword and throw an error instead of a warning.

gammapy/estimators/map/core.py
```diff
@@ -41,7 +41,6 @@
     "e2dnde": ["e_ref", "e2dnde"],
     "flux": ["e_min", "e_max", "flux"],
     "eflux": ["e_min", "e_max", "eflux"],
-    # TODO: extend required columns
```
Tobychev, line 44: Indeed a very mysterious comment, presumably reminding someone who forgot that they added required columns somewhere else and should add them here too?

But with no way of knowing which columns should be added the comment is useless.

AtreyeeS: I think this is from the time @adonath introduced the skymaps format (https://github.com/open-gamma-ray-astro/gamma-astro-data-formats/pull/170) and we were not sure of the required columns
We can remove the TODO now

AtreyeeS (approved):
Thanks @Astro-Kirsty !

AtreyeeS:
>Should I create a test dataset for this in gammapy-data and remove this section of code ?

We have been planning a cleanup of the tests for sometime now.
Maybe, after the 2.0 release, you can propose a full cleanup.

AtreyeeS:
We have two open issues about improving the tests: #94 and #3844

PR #6071, opened 2025-08-11
Allow passing of  `kwargs` to `EventListReader.read`

LeanderSchlegel:
Dear all, 

as discussed in the last developer call, I wanted to make a PR for this tiny change, that makes sure, that the keyword arguments `**kwargs` in `EventList.read(cls, filename, hdu="EVENTS", checksum=False, **kwargs)` are passed through to the `EventListReader.read(self, filename, format="gadf")`, that did not receive the `**kwargs`, like e.g. the `format.` Thank you @QRemy for looking into it together!
I added a small test in `test_event_list.py` that tests now the successful passing of the argument, by checking that the ValueError for an unknown `format`-string is raised, when such a `format` is given upon calling `EventList.read`.

Thank you a lot,
Leander

### DIFF
--- a/gammapy/data/tests/test_event_list.py
+++ b/gammapy/data/tests/test_event_list.py
@@ -136,10 +136,7 @@ class TestEventListBase:
         assert "CREATED" in hdu.header
         assert hdu.header["CREATOR"] == "SASH FITS::EventListWriter"
 
-    def test_eventlist_format_keyword_is_passed_to_EventListReader(self):
-        # Read eventlist with unknown format specification, to test that the keyword argument is correctly
-        # passed through to the read-method in the EventListReader class. There, this is expected, to
-        # cause a ValueError due to the received unknown format, testing the passing of the keyword argument.
+    def test_eventlist_read_kwargs(self):
         with pytest.raises(ValueError):
             EventList.read(
                 "$GAMMAPY_DATA/hess-dl3-dr1/data/hess_dl3_dr1_obs_id_020136.fits.gz",


QRemy (dismissed):
Thanks @LeanderSchlegel, looks good.

gammapy/data/tests/test_event_list.py
```diff
@@ -136,6 +136,16 @@ def test_eventlist_hdu_creation_metadata(self):
         assert "CREATED" in hdu.header
         assert hdu.header["CREATOR"] == "SASH FITS::EventListWriter"
 
+    def test_eventlist_format_keyword_is_passed_to_EventListReader(self):
+        # Read eventlist with unknown format specification, to test that the keyword argument is correctly
+        # passed through to the read-method in the EventListReader class. There, this is expected, to
+        # cause a ValueError due to the received unknown format, testing the passing of the keyword argument.
```
Astro-Kirsty: You don't need to include this information. We can see from the test what it does!

gammapy/data/tests/test_event_list.py
```diff
@@ -136,6 +136,16 @@ def test_eventlist_hdu_creation_metadata(self):
         assert "CREATED" in hdu.header
         assert hdu.header["CREATOR"] == "SASH FITS::EventListWriter"
 
+    def test_eventlist_format_keyword_is_passed_to_EventListReader(self):
```
Astro-Kirsty:
```suggestion
    def test_eventlist_read_kwargs(self):
```

Are there other tests you can include here with other `**kwargs` that work?

gammapy/data/tests/test_event_list.py
LeanderSchlegel: Thank you @Astro-Kirsty , I completely agree and removed it!

gammapy/data/tests/test_event_list.py
LeanderSchlegel: Thank you for your suggestions @Astro-Kirsty, I changed the name and just checked, the `EventListReader.read(self, filename, format="gadf")` in io.py takes no further kwargs that would have to be passed through from the `EventList.read`.

LeanderSchlegel:
Thank you @QRemy as well for your feedback and for adding the milestone and labels!

QRemy (approved):

gammapy/data/event_list.py
```diff
@@ -142,7 +142,7 @@ def read(cls, filename, hdu="EVENTS", checksum=False, **kwargs):
         """
         from gammapy.data.io import EventListReader
 
-        return EventListReader(hdu, checksum).read(filename)
+        return EventListReader(hdu, checksum).read(filename, **kwargs)
```
registerrier, line 145: Thanks @LeanderSchlegel . Here I would simply have removed the `**kwargs` argument altogether.

And possibly add an explicit argument `format` for future usage.

LeanderSchlegel: Thank you for your review @registerrier, I see and completely agree, would you think it is okay to leave it like this, or would it be better to quickly change the **kwargs to an explicit format keyword?

PR #6072, opened 2025-08-12
Adapt ObservationsEventsSampler.run docstring to observations

### LINKED ISSUE #6063: Should it be `observation` or `observations` in `ObservationsEventsSampler.run`
I noticed that the `run` function [here](https://docs.gammapy.org/dev/api/gammapy.data.ObservationsEventsSampler.html#gammapy.data.ObservationsEventsSampler.run)  has the parameter `observations` but in the docstring it is `observation`. 

As I understand from the docstring it should not be plural, but then this is an API change. 
Lets decide what we want to do before 2.0, because if its an API change it is best to adapt asap.

Astro-Kirsty:
Resolves #6063

I also adjusted the location of the tests for this function


gammapy/data/simulate.py
```diff
@@ -13,14 +13,14 @@ class ObservationsEventsSampler(parallel.ParallelMixin):
...
-        Overwrite the output files or not
+    outdir : str, optional
+        Path of the output files created. Default is "./simulated_data/".
+        If None, a list of `~gammapy.data.Observation` is returned.
+    overwrite : bool, optional
+        Overwrite existing file. Default is True.
     n_jobs : int, optional
         Number of processes to run in parallel.
-        Default is one, unless `~gammapy.utils.parallel.N_JOBS_DEFAULT` was modified.
+        Default is None, unless `~gammapy.utils.parallel.N_JOBS_DEFAULT` was modified.
```
QRemy: The default is one  as it is the default for N_JOBS_DEFAULT. Maybe you can find a better phrasing, but it should still mention one.

QRemy (commented):
Thanks @Astro-Kirsty, just a comment on `n_jobs`should :

AtreyeeS (approved):

gammapy/data/tests/test_simulate.py
```diff
@@ -0,0 +1,180 @@
...
+    PowerLawSpectralModel,
+    SkyModel,
+)
+from gammapy.utils.testing import requires_data
+
+LOCATION = EarthLocation(lon="-70d18m58.84s", lat="-24d41m0.34s", height="2000m")
+
+
+@pytest.fixture()
+def models(signal_model):
```
AtreyeeS: I think `signal_model` is not defined. 
Do you need this `models` fixture? The model is anyways reset in `energy_dependent_temportal_sky_model`.

Maybe just a `signal_model` fixture can suffice?

Astro-Kirsty: I adapted how the code worked! It is not possible to simply import `signal_model` so I added the code directly

registerrier (commented):

I think it would be clearer to have a specific `sampler` sub-package. `ObservationsEventSampler` is an algorithm, not a data container. Internally the import of `ObservationEventSampler` from `gammapy.datasets` shows that the logic should be simplified.

gammapy/data/tests/test_simulate.py
```diff
@@ -0,0 +1,207 @@
...
+        spectral_model=spectral_model,
+        spatial_model=spatial_model,
+        temporal_model=temporal_model,
+    )
+    return model
+
+
+@requires_data()
+def test_observation_event_sampler(signal_model):
+    from gammapy.datasets.simulate import ObservationEventSampler
```
QRemy: As `ObservationEventSampler` is defined in `gammapy.datasets.simulate` maybe the etest should stay at his original location and move only the one for `ObservationsEventsSampler`

PR #6074, opened 2025-08-12
Fix warnings on deprecated license in classifiers and universal bdist wheel options

morcuended:
**Description**

This pull request remove two deprecated options in setup.cfg causing warnings when building the wheels

```
   running bdist_wheel
  /private/var/folders/y6/nj790rtn62lfktb1sh__79hc0000gn/T/build-env-52qz50vf/lib/python3.14t/site-packages/setuptools/_distutils/cmd.py:135: SetuptoolsDeprecationWarning: bdist_wheel.universal is deprecated
  !!
  
          ********************************************************************************
          With Python 2.7 end-of-life, support for building universal wheels
          (i.e., wheels that support both Python 2 and Python 3)
          is being obviated.
          Please discontinue using this option, or if you still need it,
          file an issue with pypa/setuptools describing your use case.
  
          By 2025-Aug-30, you need to update your project and remove deprecated calls
          or your builds will no longer be supported.
          ********************************************************************************
```

and regarding the licence specified in the classifiers section:

```
  * Building wheel...
  /private/var/folders/y6/nj790rtn62lfktb1sh__79hc0000gn/T/build-env-52qz50vf/lib/python3.14t/site-packages/setuptools/dist.py:759: SetuptoolsDeprecationWarning: License classifiers are deprecated.
  !!
  
          ********************************************************************************
          Please consider removing the following classifiers in favor of a SPDX license expression:
  
          License :: OSI Approved :: BSD License
  
          See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license for details.
          ********************************************************************************
  
```

**Dear reviewer**

You can see the warnings here in the nightly builds:

https://github.com/morcuended/gammapy/actions/runs/16899754510

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -3,7 +3,7 @@ name = gammapy
 description = A Python package for gamma-ray astronomy
 author = The Gammapy developers
 author_email = gammapy-coordination-l@in2p3.fr
-license = BSD 3-Clause
+license = BSD-3-Clause
 license_files = LICENSE.rst
 url = https://gammapy.org
 url_docs = https://docs.gammapy.org/dev/
@@ -15,7 +15,6 @@ url_raw_github = https://raw.githubusercontent.com/gammapy/gammapy/master/
 platforms = any
 classifiers =
     Intended Audience :: Science/Research
-    License :: OSI Approved :: BSD License
     Operating System :: OS Independent
     Programming Language :: C
     Programming Language :: Cython
@@ -103,9 +102,6 @@ gammapy.workflow = config/*
 gammapy.irf.psf.tests = data/*
 gammapy.modeling.models.tests = data/*
 gammapy.catalog.tests = data/*
-
-[bdist_wheel]
-universal = true
 
 [tool:pytest]
 minversion = 3.0


setup.cfg
```diff
@@ -104,9 +103,6 @@ gammapy.irf.psf.tests = data/*
 gammapy.modeling.models.tests = data/*
 gammapy.catalog.tests = data/*
 
-[bdist_wheel]
```
Astro-Kirsty, line 107: Do we really want to remove this all together, or is it just the that `universal = true` command is the deprecated part?
How else are the wheels distributed correctly?

Thanks @morcuended

morcuended:
In https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#wheels there is this note:

> If your project also supports Python 2 and contains no C extensions, then you should create what’s called a Universal Wheel by adding the following to your setup.cfg file:
> 
> [bdist_wheel]
> universal=1
> 
> Only use this setting if your project does not have any C extensions and supports Python 2 and 3.

which I think it does not fulfil none of the two conditions. I do not see any other use of it. It seems something quite old already. Wheels are build and distributed regardless of that universal option (https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#platform-wheels)

PR #6077, opened 2025-08-12
Fix evaluate_error for PiecewiseNormSpectralModel

QRemy:
Fix evaluate_error for PiecewiseNormSpectralModel and added missing test for that model.

### DIFF
--- a/gammapy/modeling/models/spectral.py
+++ b/gammapy/modeling/models/spectral.py
@@ -1522,7 +1522,7 @@ class PiecewiseNormSpectralModel(SpectralModel):
         e_nodes = scale(self.energy.to(energy.unit).value)
         v_nodes = scale(norms)
         log_interp = scale.inverse(np.interp(e_eval, e_nodes, v_nodes))
-        return log_interp
+        return u.Quantity(log_interp)
 
     def __call__(self, energy):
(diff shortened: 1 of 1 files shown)


gammapy/modeling/models/spectral.py
```diff
@@ -1516,13 +1516,16 @@ def norms(self):
         """Norm values."""
         return u.Quantity([p.value for p in self.parameters])
 
-    def evaluate(self, energy, **norms):
+    def evaluate(self, energy, *norms):
         scale = interpolation_scale(scale=self._interp)
         e_eval = scale(np.atleast_1d(energy.value))
         e_nodes = scale(self.energy.to(energy.unit).value)
-        v_nodes = scale(self.norms)
+        v_nodes = scale(norms)
```
AtreyeeS, line 1523: Do we need to keep an option to use `self.norms` if norms is None?

AtreyeeS: Because this is in out public API, so this change would be not be backwards compatible, no? Though I doubt if people are really using the `evaluate` functions from outside.

QRemy: no using self.norms was likely a bug because then the norms given on input are not used.

Issue #6091, opened 2025-08-18
Post 2.0rc1 update and remaining PRs


bkhelifi:
In this issue are listed the remaining PRs to do into the 2.0rc1 for the LTS v2.0:

- [ ] Update a file ``requirement.txt`` for pip
- [x] Remove the first note into https://docs.gammapy.org/dev/getting-started/index.html#recommended-setup about sherpa (that is no longer required)
- [x] Fix warning about `config.copy` originating from deprecated pydantic warning  https://docs.gammapy.org/dev/tutorials/analysis-3d/analysis_3d.html#configuration-for-stacked-and-joint-analysis
- AoB?

AtreyeeS:
Where should the `requirements.txt` file be located?

AtreyeeS:
Is this a good requirements.txt file? Shall this be located alongside the setup.cfg?

```
python>3.12
ipython==8.27
jupyter
jupyterlab==4.2
cython
numpy==2.1
scipy==1.16
pandas==2.2
matplotlib==3.9
astropy==7.1
regions==0.10           
iminuit==2.31
healpy>=1.18             
naima ==0.10                         
pyyaml==6.0
click==8.2
pydantic==2.11
corner==2.2
requests
tqdm
ray[default]==2.48
gammapy==2.0rc2
```

@QRemy @maxnoe

registerrier:
> Is this a good requirements.txt file? Shall this be located alongside the setup.cfg?
> 

You could use approximate versions to not impose a specific bugfix e.g. `numpy~=2.1.0` or I suppose `numpy==2.1.*`

I would simply have the file in the install folder of the gammapy page as the conda env file. It does not need to be shipped alongside gammapy

maxnoe:
For what do you need a requirements file?

maxnoe:
What does this file offer over `pip install gammapy==2.1.0` or `pip install gammapy[all]==2.1.0`?

It seems to me it's just one more thing to maintain, keep in sync and potentially confuse users.

AtreyeeS:
@maxnoe For conda, we impose specific version in the `environment.yml` file. 
How can we do that for `pip` ? The `setup.cfg` provides only a minimal version, right?

This is mainly for people not using conda, for example as on the MPIK cluster by @QRemy

QRemy:
> How can we do that for `pip` ? The `setup.cfg` provides only a minimal version, right?

`pip install -r environment.txt`

and `environment.txt` is just something like 
[...]

PR #6093, opened 2025-08-18
Suppress pydantic warnings

### LINKED ISSUE #2221: Review and improve inheritance and docstrings
In Gammapy we have use [astropy.utils.misc.InheritDocstrings](http://docs.astropy.org/en/stable/api/astropy.utils.misc.InheritDocstrings.html) here:
```
gammapy/maps/geom.py:1094:class MapGeomMeta(InheritDocstrings, abc.ABCMeta):
gammapy/maps/base.py:18:class MapMeta(InheritDocstrings, abc.ABCMeta):
gammapy/utils/fitting/fit.py:16:class FitMeta(InheritDocstrings, abc.ABCMeta):
```
And we currently have ~ 100 classes where we use inheritance, but not `InheritDocstrings`:
https://gist.github.com/cdeil/d3573e477e12e94853cd406a509c93b8

I noticed that for some classes, like [gammapy.astro.population.CaseBattacharya1998](https://docs.gammapy.org/0.12/api/gammapy.astro.population.CaseBattacharya1998.html) a ton of useless API HTML docs are generated, see e.g. [here](https://docs.gammapy.org/0.12/api/gammapy.astro.population.CaseBattacharya1998.html#gammapy.astro.population.CaseBattacharya1998.bounding_box). This makes the API docs almost useless (hard to find relevant methods), and also makes the Sphinx docs build slow.

There are other cases like [gammapy.spectrum.models.LogParabola](https://docs.gammapy.org/0.12/api/gammapy.spectrum.models.LogParabola.html) where IMO it's useful to show the methods from the super class in the API docs.

This is a reminder issue to:
- review and possibly reduce the use of inheritance in Gammapy
- review the API docs and decide where we want super docstrings and not, and where we want to use InheritDocstrings or not

The scope should probably be limited to just the question of API docstring configuration and inheritance, but a general discussion on inheritance might not be avoidable - e.g. if the solution we want for `CaseBattacharya1998` eventually is "don't sub-class the Astropy model, and either use the Gammapy model or no superclass at all", the question / task to change the API docs generation to avoid showing the superclass methods becomes mute.

@adonath or anyone - thoughts?

Astro-Kirsty:
This suppresses >100 warnings in the docs build which are caused by pydantic and not to do with our documentation. 

We should take our time to review the docs build to ensure the implementation is correct

Also resolves #2221 -- we don't inherit from classes outside gammapy

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -40,15 +40,6 @@ def setup(app):
     app.add_directive("substitution-code-block", SubstitutionCodeBlock)
     app.add_post_transform(DynamicPRLinkTransform)
 
-    # Skip documenting Pydantic to suppress warnings
-    def filter_pydantic_docstrings(app, what, name, obj, options, lines):
-        module = getattr(obj, "__module__", "")
-        if module and module.startswith("pydantic"):
-            # wipe docstring to silence parsing warnings
-            lines[:] = []
-
-    app.connect("autodoc-process-docstring", filter_pydantic_docstrings)
-
 
 conf = ConfigParser()
 conf.read([os.path.join(os.path.dirname(__file__), "..", "setup.cfg")])
@@ -105,8 +96,7 @@ default_role = 'obj'
 # Add any Sphinx extension module names here, as strings.
 extensions = [
     # Order for sphinx_automodapi is important
-    "sphinx.ext.autosummary",
-    "sphinx_automodapi.automodapi", # This should come after autosummary
+    "sphinx_automodapi.automodapi",
     "sphinx_automodapi.smart_resolver",
     "sphinx_click.ext",
     'sphinx_copybutton',
@@ -148,13 +138,7 @@ intersphinx_mapping = {
 	"pandas": ("https://pandas.pydata.org/pandas-docs/stable/", None),
 	}
 
-# -- Options for autosummary/autodoc output ------------------------------------
-# Enable generation of stub files
-autosummary_generate = True
-
-# Document inherited members
-automodsumm_inherited_members = True
-
+# -- Options for autodoc output ------------------------------------
 # Include class and __init__ docstrings
 autoclass_content = "both"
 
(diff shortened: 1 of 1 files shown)


registerrier:
@Astro-Kirsty Could you paste some of the warnings?

Astro-Kirsty:
image

Here is an example of the warnings

docs/conf.py
```diff
@@ -280,21 +288,18 @@ def setup(app):
 
 github_issues_url = "https://github.com/gammapy/gammapy/issues/"
 
-# In `about.rst` and `references.rst` we are giving lists of citations
-# (e.g. papers using Gammapy) that partly aren't referenced from anywhere
-# in the Gammapy docs. This is normal, but Sphinx emits a warning.
-# The following config option suppresses the warning.
-# http://www.sphinx-doc.org/en/stable/rest.html#citations
-# http://www.sphinx-doc.org/en/stable/config.html#confval-suppress_warnings
```
Tobychev, line 288: The link to the sphinx doc about the option seems useful to keep in the comment, why do you want to delete it?

AtreyeeS:
Don't we need this for 2.0 ?

adonath:
The issue comes from the fact that we inherit from Pydantic's `BaseModel` and re-expose the inherited method docs, see https://docs.gammapy.org/dev/api/gammapy.analysis.AnalysisConfig.html#gammapy.analysis.AnalysisConfig. I think we should not do this, it clutters the docs unnecessarily.

Maybe try:
```
# In conf.py
autodoc_default_options = {
    'members': True,
    'inherited-members': False,  # or just omit this line
}
```

Astro-Kirsty:
Thanks for the suggestion @adonath but this creates many more warnings (4431 warnings!!! )about other things:
Screenshot from 2025-08-19 17-37-32

docs/conf.py
Astro-Kirsty: I removed this because it can be easily found through a search of the sphinx docs, which we should not repeat unnecessarily. 
We do not have links to all the other commands in the file either

Astro-Kirsty:
> Don't we need this for 2.0 ?

Not necessary because it is only warnings for the build (not errors). 
I would rather we leave for 2.1 so we can ensure that this method is the correct one and that it doesn't break any formatting

AtreyeeS:
Does this address #6091 third point?

Astro-Kirsty:
> Does this address #6091 third point?

No, it is unrelated

Astro-Kirsty:
A much simpler fix has now been implemented which also helps with the inheriting from outside classes (resolves #2221)

Important to note that `automodapi` has its own fallback path for if `autosummary` isn’t there, it just generates API docs directly with `autodoc`. So by enabling `autosummary` we might have been causing extra stub files to be generated that we don't actually want.

AtreyeeS (approved):
Thanks @Astro-Kirsty ! Good from my side

PR #6027, opened 2025-07-29
Correct setup in tutorialjupytexttests.yml

morcuended:
**Description**

This pull request adds the setup needed for running jupytext. The notebooks require gammapy datasets, and some additional dependencies jupytext, jupyter-client, ipykernel and others used in the notebooks (tqdm, requests, etc). This is why I decided to directly use the `environment-dev.yml` file.

**Dear reviewer**

Current installation is still not sufficient. See https://github.com/gammapy/gammapy/actions/runs/16594581310/job/46938043061
Is there a way to test the workflow before merging?

PR #6095, opened 2025-08-19
jupytext requires plt.show

Astro-Kirsty:
This PR is linked to the jupytext execution of the tutorials (workflow being create here #6027).

We must ensure that each plot we create has an associated `plt.show`. There were some missing from the docs, this PR fixes that. As well as removing unnecessary `plt.show` that were found


Astro-Kirsty:
The lines in the CTAO tutorials for terminal are also run i.e. https://docs.gammapy.org/dev/tutorials/data/cta.html#latest-ctao-performance-files
Currently trying to find a way such that this doesn't happen

AtreyeeS (approved):
Thanks @Astro-Kirsty !
We can have this in for 2.0

bkhelifi (approved):
Looks good for me. Thanks

PR #6097, opened 2025-08-21
Update release.rst after 2.0 release

AtreyeeS:


docs/development/release.rst
```diff
@@ -25,12 +25,8 @@ Feature Freeze and Branching
 
 #. Fill the changelog ``docs/release-notes/<version>.rst`` for the version you are about to release.
 
-   * To generate the list of pull requests and issues run ``python dev/github_summary.py create_pull_request_table``.
-     This will create a list of all closed pull requests and save it to ``table_pr.ecsv``.
-     Next, use the option ``merged_PR`` to extract a relevant list corresponding to the release.
-     You can then manually delete any entries which correspond to small improvements or bug fixes.
+   * To generate the list of pull requests and issues, and list of authors run ``python dev/github_summary.py create_pull_request_table``. Note that you will need to use your github token here.
```
Astro-Kirsty: Is there a reason why you delete all the extra information too?

docs/development/release.rst
```diff
@@ -40,32 +36,28 @@ Feature Freeze and Branching
 
 #. Add a new milestone to the `GitHub issue tracker <https://github.com/gammapy/gammapy/milestones>`__ for the version
    ``v<version>.x`` (this will also be used for the next bugfix release). Also create a ``backport-v<version>.x``
-   `label <https://github.com/gammapy/gammapy/labels>`__.
+   `label <https://github.com/gammapy/gammapy/labels>`__. Put the description as ``on merge: backport to v<version.x`` for the automatic backport triggering.
```
Astro-Kirsty:
```suggestion
   `label <https://github.com/gammapy/gammapy/labels>`__., which has the description ``on merge: backport to v<version.x`` to allow for automatic backport triggering.
```

docs/development/release.rst
```diff
@@ -40,32 +36,28 @@ Feature Freeze and Branching
...
     git fetch upstream --tags --prune
     git checkout -B main upstream/main
 
-#. Create a new branch with the name of the version::
-
-    git branch v<version>.x
-
-#. Stay on the ``main`` branch and make a copy and update the ``docs/release-notes/<version>.rst``
-#. Commit the changes and push to GitHub ``main``.
+#. From the github online interface, create a new branch v<version>.x
```
Astro-Kirsty, line 50: This is the same as the instructions we already have right? It doesn't matter if you do those steps directly on the repo or on the terminal?

Also why do you remove the note about making a copy of the release-notes?

docs/development/release.rst
```diff
@@ -40,32 +36,28 @@ Feature Freeze and Branching
...
 #. Update the entry for the feature freeze in the
    `Gammapy release calendar <https://github.com/gammapy/gammapy/wiki/Release-Calendar>`__.
 
 
 Releasing the first major release candidate
 -------------------------------------------
 
+#. First, make sure you have `gpg` keys generated for your github account (https://docs.github.com/en/authentication/managing-commit-signature-verification/generating-a-new-gpg-key).
```
Astro-Kirsty:
```suggestion
#. First, make sure you have a [`gpg key`](https://docs.github.com/en/authentication/managing-commit-signature-verification/generating-a-new-gpg-key) generated for your GitHub account.
```

docs/development/release.rst
```diff
@@ -40,32 +36,28 @@ Feature Freeze and Branching
...
 #. In the `gammapy-webpage repo <https://github.com/gammapy/gammapy-webpage>`__:
 
    * Add an entry for the release candidate like ``v1.0rc1`` or ``v1.1rc1`` in the ``download/index.json`` file,
      by copying the entry for ``dev`` tag. As we do not handle release candidates nor bug fix releases for data,
-     this still allows to fix bugs in the data during the release candidate testing.
+     this still allows to fix bugs in the data during the release candidate testing. Update the link to the conda environment.
    * In the ``download/install`` folder, copy a previous environment file file as ``gammapy-1.0rc1-environment.yml``.
```
Astro-Kirsty:
```suggestion
   * In the ``download/install`` folder, copy a previous environment file as ``gammapy-1.0rc1-environment.yml``.
```

docs/development/release.rst
```diff
@@ -40,32 +36,28 @@ Feature Freeze and Branching
...
-   * Adapt the dependency conda env name and versions as required in this file.
+   * Adapt the dependency conda env name and versions as required in this file. Normally, it should be the latest versions of the packages. Note that for the release candidates, `gammapy` must be included under pip.
```
Astro-Kirsty:
```suggestion
   * Adapt the dependency conda env name and versions as required in this file. Normally, it should be the latest versions of the packages. 
     Note that for the release candidates, `gammapy` must be included under pip.   
```

docs/development/release.rst
```diff
@@ -40,32 +36,28 @@ Feature Freeze and Branching
...
 Releasing the first major release candidate
 -------------------------------------------
 
+#. First, make sure you have `gpg` keys generated for your github account (https://docs.github.com/en/authentication/managing-commit-signature-verification/generating-a-new-gpg-key).
 #. In the `gammapy-webpage repo <https://github.com/gammapy/gammapy-webpage>`__:
 
    * Add an entry for the release candidate like ``v1.0rc1`` or ``v1.1rc1`` in the ``download/index.json`` file,
      by copying the entry for ``dev`` tag. As we do not handle release candidates nor bug fix releases for data,
-     this still allows to fix bugs in the data during the release candidate testing.
+     this still allows to fix bugs in the data during the release candidate testing. Update the link to the conda environment.
```
Astro-Kirsty: Is it really necessary to add that, because we mention to add an entry for the release candidate.  So one can see from previous releases that this needs to be adapted.
If so, I would reword it to have it at the beginning of these sentences

docs/development/release.rst
```diff
@@ -89,6 +81,7 @@ Releasing the first major release candidate
    `gammapy-docs <https://github.com/gammapy/gammapy-docs>`__  to check that the necessary actions have started.
 #. Once the docs build is successful find the ``tutorials_jupyter.zip`` file for the release candidate in the
    `gammapy-docs repo <https://github.com/gammapy/gammapy-docs>`__ and adapt the ``download/index.json`` to point to it.
+#. Edit `docs/stable/index.html` so that the url points to the last stable version, eg: `url=../1.3` instead of `url=../${release}`.
```
Astro-Kirsty:
```suggestion
#. Edit `docs/stable/index.html` so that the URL points to the last stable version, e.g.: `url=../1.3` instead of `url=../${release}`.
```


Maybe you can be explicit that this is also on the gammapy-doc repo?

docs/development/release.rst
AtreyeeS: From v2.0, we are not putting the list of PRs in the release notes. This was started as a way to give credit to contributors, but since we have the zenodo releases, we decided to remove this.
Now, we only need the list of authors

docs/development/release.rst
AtreyeeS: Quentin and I faced some issues doing it from the terminal, and found that the web interface is much smoother.
We don't need a copy of the release notes as the notes are not merged at this stage

docs/development/release.rst
Astro-Kirsty: Okay so maybe mentioning that we only utilised the list of authors?
Also should we then create a separate py file which only creates the list of authors. Because if this is how we want to proceed for future versions, it does not make sense to run all of this

docs/development/release.rst
AtreyeeS: For the release candidates, we dont want the gammapy-doc link to change

Astro-Kirsty: So you mean the sentence shouldn't be here? I don't understand

My suggestion was simply to add something like:
```suggestion
#. Edit `docs/stable/index.html` in the `gammapy-docs repo <https://github.com/gammapy/gammapy-docs>`__ so that the url points to the last stable version, eg: `url=../1.3` instead of `url=../${release}`.
```

AtreyeeS: oh okay i see now. I misunderstood your comment

docs/development/release.rst
```diff
@@ -80,7 +73,7 @@ Releasing the first major release candidate
 #. Locally create a new release candidate tag on the ``v1.0.x``, like ``v1.0rc1`` for Gammapy and push::
 
     git tag -s v1.0rc1 -m "Tagging v1.0rc1"
-    git push upstream v1.0.x
+    git push upstream tag v1.0.x
```
registerrier:
```suggestion
    git push upstream v1.0rc1
```
Here it is the tag that must be pushed not the branch

docs/development/release.rst
```diff
@@ -111,6 +108,13 @@ Releasing the final version of the major release
...
      notebook entry, typically the link extensions are the same between versions.
 
+#. In the gammapy repo, switch to the correct branch and update the ``CITATION.cff`` date and version by running the
+   ``dev/prepare-release.py`` script::
+
+    git checkout v1.0.x
+    python ./dev/prepare-release.py --release v1.0rc1
+    git push 
+
 #. Locally create a new release tag like ``v1.0`` for Gammapy and push::
```
registerrier:
```suggestion
#. Locally create a new release tag like ``v1.0`` for Gammapy and push the tag to upstream::
```

registerrier (changes requested):
Thanks @AtreyeeS . 

I think we need to specify that the tag must be pushed. Not the branch. We should do `git push --tags` as this pushes all tags and not only the one that was created.
So one needs to update the `git push` to use the tag name, not the branch name.

I think this was incorrectly modified (maybe by myself...) during the last cleanup of the instructions.

docs/development/release.rst
```diff
@@ -150,34 +160,32 @@ Steps for the day to announce the release:
...
+       git add postBuild requirements.txt runtime.txt start
+       git commit -s -m "Update binder configuration for new release"
+       git push origin master
 
-        git tag -s v1.3 -m "Tagging v1.3"
-        git push origin v1.3
+   * Tag the new version and push to the upstream repository::
 
+       git tag -s v1.3 -m "Tagging v1.3"
+       git push origin v1.3
```
AtreyeeS:
```suggestion
       git push --tags
```

docs/development/release.rst
AtreyeeS: This we can decide later. Can `towncrier` update this info?

Astro-Kirsty:
I will make a commit to fix the errors in the docs build first

PR #6098, opened 2025-08-21
Clearer appearance for the click scipts in the documentation

Astro-Kirsty:
As per the title.
The current build looks a bit ugly:
https://docs.gammapy.org/2.0rc1/api-reference/scripts.html

Here is the updated version:
Screenshot from 2025-08-21 09-03-40


AtreyeeS (approved):
Thanks @Astro-Kirsty

PR #6102, opened 2025-08-21
jfactory integral calculation description

### LINKED ISSUE #5296: Update docs for `JFactory` integral calculation
We should add an explanation in the docs for this integral and why its is 2 * [rmin, rmax] + [rmax, 4*rmax]

https://github.com/gammapy/gammapy/blob/2176b9588a7d047fcc814a6aa42d9fc9c78d8a51/gammapy/astro/darkmatter/utils.py#L57-L75

Explanation can be found here:
https://gist.github.com/katrinstreil/eeee9f2969dbc037888f876d5f4e5b9d

Astro-Kirsty:
This fixes #5296

The explanation which I utilised to create this comes from:
 https://gist.github.com/katrinstreil/eeee9f2969dbc037888f876d5f4e5b9d

**The current implementation has a lot of information:**

Screenshot from 2025-08-21 15-17-29

**Otherwise we could go for a more minimalist approach:**

Screenshot from 2025-08-21 15-17-35

I also added the parameters and returns to the docstring and adapted the ZhaoProfile because it was causing a warning in the docs build

### DIFF
--- a/gammapy/astro/darkmatter/profiles.py
+++ b/gammapy/astro/darkmatter/profiles.py
@@ -138,7 +138,7 @@ class ZhaoProfile(DMProfile):
     ----------
     .. [1] `Zhao (1996), "Analytical models for galactic nuclei"
       <https://ui.adsabs.harvard.edu/abs/1996MNRAS.278..488Z>`_
-    * `Marco et al. (2011), "PPPC 4 DM ID: a poor particle physicist cookbook for dark matter indirect detection"
+    .. [2] `Marco et al. (2011), "PPPC 4 DM ID: a poor particle physicist cookbook for dark matter indirect detection"
       <https://ui.adsabs.harvard.edu/abs/2011JCAP...03..051C>`_
     """
 
--- a/gammapy/astro/darkmatter/utils.py
+++ b/gammapy/astro/darkmatter/utils.py
@@ -1,5 +1,6 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 """Utilities to compute J-factor maps."""
+
 import html
 import numpy as np
 import astropy.units as u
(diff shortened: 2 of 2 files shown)


aritra-gupta:
Hi,

I was trying to understand the origin of r_min and r_max. For example, a geometrical interpretation of r_min and r_max would make things clearer. Specifically, I am curious to know the origin of the expression for $dl$ using $r_{min}$.

As far as my previous understanding, we intend to compute $$I = \int_{0}^{s_{max}}  \rho^2 (r)  ds$$. r is the distance of the point of dark matter annihilation from the galactic center. s is the distance of the point of occurrence of dark matter annihilation to the observer (line of sight). If the distance from us to the galactic center is R, then from cosine law: $r^2 = R^2 + s^2 - 2 R s \cos\theta$. This provides us with $I$ as a function of $\theta$ and R as required after performing the $s$ integral. In general $s$ varies from $0$ to $\infty$, but numerically it converges much before that. Here $\theta$ is the aperture angle and is related to galactic latitude and longitude by: $\cos\theta = \cos b \cos l$.

aritra-gupta:
After doing some more checks, I found that although the results match with PPPC approximately for small angular deviations, they were off by a large amount when the angular separation increases. A possible reason may lie in the following line of the code
https://github.com/gammapy/gammapy/blob/da31a77ea4d204010bdccc3d9534c051a6e918e2/gammapy/astro/darkmatter/utils.py#L54 we have used $\tan \theta = r / r_{sun}$. This is not always true in the general case since the point of dark matter interaction may lie anywhere in the sky as explained in the following illustration:
Screenshot from 2025-09-19 15-23-35   
The point $B$ does not always necessarily coincide with $D$. If it does, then the above line of code is exact and correct. Other places where this idea was used was:https://github.com/gammapy/gammapy/blob/da31a77ea4d204010bdccc3d9534c051a6e918e2/gammapy/astro/darkmatter/profiles.py#L52 A correct approach should be to use the cosine rule as was outlined in my preceding comment.

bkhelifi:
What is the status here?

QRemy:
Improving the doc and having less approximations are two different issues so I would merge this PR with only doc changes and ask for one of the DM people to open a separate PR if they want to change the code itself.

AtreyeeS (approved):
Thanks! I agree we can adapt the docs for now. Maybe we can have a reminder issue (or a TODO) about the small angle approximation?

PR #6106, opened 2025-08-22
Remove deprecated pydantic copy in tutorial

registerrier:
**Description**

This pull request adapts the call to `AnalysisConfig.copy()` which inherits from pydantic `BaseModel.copy()`. The method is deprecated and one should use `BaseModel.model_copy()` instead. 

Another option is to create an `GammapyBaseConfig.copy()` method and encapsulate the pydantic call instead. Since this is the only place where we use `copy`. I suppose it is simpler to adapt the tutorial with the proper pydantic call.

**Dear reviewer**

### DIFF
--- a/examples/tutorials/analysis-3d/analysis_3d.py
+++ b/examples/tutorials/analysis-3d/analysis_3d.py
@@ -91,10 +91,10 @@ print(config)
 # future reference.
 #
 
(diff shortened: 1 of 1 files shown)


AtreyeeS (approved):
Thanks @registerrier !

Issue #6114, opened 2025-08-27
When to keep kwargs


Astro-Kirsty:
While looking at the following methods:
[HpxGeom.cutout](https://docs.gammapy.org/dev/api/gammapy.maps.HpxGeom.html#gammapy.maps.HpxGeom.cutout)
[HpxNDMap.cutout](https://docs.gammapy.org/dev/api/gammapy.maps.HpxNDMap.html#gammapy.maps.HpxNDMap.cutout)

I noticed that they accept `*args`/`**kwargs`, but these are not used in the implementation.
This raises a few questions about API consistency and design:

- Why were these arguments added in the first place?
- Should they be removed, since they don’t currently do anything?
- Or should they be implemented properly (e.g. for consistency with [WcsGeom.cutout](https://docs.gammapy.org/dev/api/gammapy.maps.WcsGeom.html#gammapy.maps.WcsGeom.cutout))

Some considerations:
- `HpxGeom` inherits from `Geom`, which has no `cutout` method
- `HpxNDMap` inherits from `HpxMap` → `Map`, again none of these have a `cutout` method
- `WcsGeom` also inherits from `Geom`, but it does have a `cutout` with `**kwargs`

So the current `*args`/`**kwargs` in `HpxGeom`/`HpxNDMap` are not related to parent classes.

**Possible options:**
- Remove `*args`/`**kwargs` from these methods to avoid confusion
- Keep them for API consistency (e.g. in case code calls geom.cutout() without knowing the specific type), but add a note in the docstring that extra kwargs are ignored or raise a warning if kwargs are passed.

We should decide on some consistent idea as this might occur in other places through out the code base.

PR #6115, opened 2025-08-27
Kwargs updates

### LINKED ISSUE #3457: Keyword forwarding and documentation

Astro-Kirsty:
This is final PR to fix #3457

I browsed through the code and we have implemented the **kwargs nicely. I opened a new issue which came up when doing this PR #6114

### DIFF
--- a/gammapy/estimators/profile.py
+++ b/gammapy/estimators/profile.py
@@ -1,5 +1,6 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 """Tools to create profiles (i.e. 1D "slices" from 2D images)."""
+
 import numpy as np
 import scipy.ndimage
 from astropy import units as u
@@ -225,7 +226,7 @@ class ImageProfile:
             Smoothing width given as quantity or float. If a float is given it
             is interpreted as smoothing width in pixels. If an (angular) quantity
             is given it is converted to pixels using `xref[1] - x_ref[0]`. Default is "0.1 deg".
-        kwargs : dict, optional
+        **kwargs : dict, optional
             Keyword arguments passed to `~scipy.ndimage.uniform_filter`
             ('box') and `~scipy.ndimage.gaussian_filter` ('gauss').
 
--- a/gammapy/maps/hpx/geom.py
+++ b/gammapy/maps/hpx/geom.py
@@ -219,7 +219,7 @@ class HpxGeom(Geom):
         else:
             return ravel_hpx_index(idx_local, self.npix)
 
-    def cutout(self, position, width, **kwargs):
+    def cutout(self, position, width):
         """Create a cutout around a given position.
 
         Parameters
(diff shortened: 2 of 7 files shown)


QRemy (approved):
Thanks @Astro-Kirsty no comment from my side.

registerrier:

PR #6119, opened 2025-09-01
Remove `plot_interactive` in 3D analysis tutorial

Astro-Kirsty:
This is related to #6027

The jupytext script will not run with the `plot_interactive` functions in `analysis_3d.py`. 
This type of functionality does not show in our documentation anyway as an interactive plot, so instead I propose here to exclude using this function and just use a plot. I have made a note about the possibility of using `plot_interactive`.
The jupytext for this now takes <20seconds to run.

The only other usage of `plot_interactive` is in the Fermi-LAT tute. We do not have issues with this a jupytext for some reason. Do we want to just keep this functionality as is or also remove it here?


AtreyeeS (approved):
Thanks @Astro-Kirsty ! It will still be good to expose `plot_interactive`. If it creates serious problems with jupytext, maybe it can be mentioned as an option in the text (something like "You can also use `plot_grid` or `plot_interactive` to see visualize the counts in different energy slices")

Astro-Kirsty:
> Thanks @Astro-Kirsty ! It will still be good to expose `plot_interactive`. If it creates serious problems with jupytext, maybe it can be mentioned as an option in the text (something like "You can also use `plot_grid` or `plot_interactive` to see visualize the counts in different energy slices")

I already mention it in the text, would you like me to have this specific wording though?

morcuended:
Seeing now @AtreyeeS comment, I would reword it as suggested. I think it will be more clear that way

examples/tutorials/analysis-3d/analysis_3d.py
```diff
@@ -135,17 +135,18 @@
 print(dataset_stacked)
 
 ######################################################################
-# To plot a smooth counts map
+# To visualise a smooth counts map, you can use the `~gammapy.maps.WcsNDMap.plot_interactive`
+# functionality, or create a plot of the counts summed over the energy axis:
```
morcuended:
```suggestion
# To visualise a smooth counts map in different energy slices, you can use the
# `~gammapy.maps.WcsNDMap.plot_interactive` or `~gammapy.maps.WcsNDMap.plot_grid`
# functionalities. Alternatively, you can create a plot of the counts summed over the energy axis:
```

morcuended (commented):
Suggestion

morcuended (approved):
Thanks @Astro-Kirsty, it looks good to me know

examples/tutorials/analysis-3d/analysis_3d.py
AtreyeeS: I would remove the word `smooth` from here. smoothing has to be done explicitly

QRemy:

PR #6120, opened 2025-09-01
Remove some unnecessary `xfail` commands in tests

Astro-Kirsty:
This is to remove some unnecessary `xfail`.
Related to the fixing of #2103 

The `modeling/tests/test_sherpa.py` was failing because of the `pars['y'].value`. In this case the min was 301000, but the value was 301999.999624371. This is probably because of the numerical optimiser being user. The different of 1000 doesn't seem that large at this scale i.e. 1e3 difference but number is 1e5. Not sure if we should use the change I suggest or we really expect something better/else. Let me know

### DIFF
--- a/examples/tutorials/analysis-3d/analysis_3d.py
+++ b/examples/tutorials/analysis-3d/analysis_3d.py
@@ -135,17 +135,19 @@ dataset_stacked = analysis_stacked.datasets["stacked"]
 print(dataset_stacked)
 
 ######################################################################
-# To plot a smooth counts map
+# To visualise a counts map in different energy slices, you can use the
+# `~gammapy.maps.WcsNDMap.plot_grid` or `~gammapy.maps.WcsNDMap.plot_interactive`
+# functionalities, or create a plot of the counts summed over the energy axis:
 #
 
-dataset_stacked.counts.smooth(0.02 * u.deg).plot_interactive(add_cbar=True)
+dataset_stacked.counts.sum_over_axes().smooth(0.02 * u.deg).plot(add_cbar=True)
 plt.show()
 
 ######################################################################
-# And the background map
+# Similarly with the background map:
 #
 
-dataset_stacked.background.plot_interactive(add_cbar=True)
+dataset_stacked.background.sum_over_axes().plot(add_cbar=True)
 plt.show()
 
 ######################################################################


gammapy/utils/tests/test_scripts.py
```diff
@@ -16,7 +16,6 @@
 )
 
 
-@pytest.mark.xfail
 def test_get_images_paths():
     assert any("images" in str(p) for p in get_images_paths())
```
Astro-Kirsty, line 20: This is failing in the test, but it ran locally.

Looking into this now

Tobychev:
I'm guessing the test was checking that the fit pushed the y-value all the way to the edge of the allowed range, if you look at the function being optimised `fcn`in `MyDataset` you can see that the optimal value for y should be around 3e5. 

The fact the sherpa fit doesn't step all the way to the edge is a bit worrying I think, even considering that the error on that parameter is quite large.

gammapy/data/tests/test_observations.py
```diff
@@ -367,12 +367,12 @@ def test_observation_read_single_file_fixed_rad_max():
     assert u.allclose(obs.rad_max.quantity, 0.1414213 * u.deg)
 
 
-@pytest.mark.xfail
 @requires_data()
 class TestObservationChecker:
     def setup_method(self):
         self.data_store = DataStore.from_dir("$GAMMAPY_DATA/cta-1dc/index/gps")
 
+    @pytest.mark.xfail
```
AtreyeeS, line 375: So does this mean we have a problem with `observation.check` or specifically with the cta-1dc ?

gammapy/irf/edisp/tests/test_kernel.py
```diff
@@ -129,8 +126,5 @@ def test_io_ogip_checksum(tmp_path):
     path = tmp_path / "test.fits"
     edisp.write(path, checksum=True)
 
-    # TODO: why does this fail?
-    # MATRIX hdu checksum is changed.
-    with warnings.catch_warnings():
-        warnings.simplefilter("error")
-        fits.open(path, checksum=True)
+    edisp_2 = fits.open(path, checksum=True)
+    assert edisp_2[1].header["HDUCLASS"], "OGIP"
```
AtreyeeS, line 130: Do we need this? This test is only for the checksum.
Also, the checksum problem has resolved?

gammapy/modeling/tests/test_sherpa.py
```diff
@@ -73,6 +72,12 @@ def test_sherpa_limits(pars):
 
     assert info["success"]
 
+    # Check that optimiser respects bound
+    assert pars["y"].value >= pars["y"].min
+
     # Check the result in parameters is OK
-    assert_allclose(pars["x"].value, 2, rtol=1e-2)
-    assert_allclose(pars["y"].value, 301000, rtol=1e-3)
```
AtreyeeS, line 78: It seems to me this check was flawed. If we have a constrained fit where we do not allow the parameter to reach the "true" value, it is not necessary that the minima will be at the edge.

AtreyeeS (approved):
Thanks @Astro-Kirsty ! Looks mostly good to me. Do you understand what is happing with the observation.check?

gammapy/data/tests/test_observations.py
registerrier: We have a problem with `ObservationChecker`, and probably we should have deprecated `Observation.check`.
The modification of `EventList` made the internal model not compliant with what `EventListChecker` was checking. There is a validation mechanism in place and technically `EventListChecker` is no longer use. We should adapt or completely remove the checkers.

bkhelifi: I will be nice to update this checker, there are very useful. They might/will be extended to check formats of any products... (FAIRisation of Gammapy).
 But one could move them in some utils files

Astro-Kirsty: I suggest we create a new issue regarding this then

gammapy/modeling/tests/test_sherpa.py
Astro-Kirsty: Yes exactly!

gammapy/data/tests/test_observations.py
AtreyeeS: For people creating IRFs/event lists, the checkers can be helpful if implemented correctly.

registerrier: Yes they are useful but mostly as format validation in writers.

Issue #4238, opened 2022-12-06
ObservationTable definition / API


maxnoe:
**Is your feature request related to a problem? Please describe.**

Part of the ongoing work to support multiple data formats / versions and #3767 

* [ ] Define columns of ObservationTable
* [ ] remove reliance on meta dict in properties
* [ ] add methods to convert to / from fits / GADF

LeanderSchlegel:
Dear @maxnoe, Dear all,

@registerrier and @bkhelifi introduced me to this project and I'm starting to look into it. I want to share some first thoughts on it:


**Define columns of ObservationTable**
As a very first step, I tried to find a minimal set of arguments required, by comparing:

- the GADF-specification for the observation index table (https://gamma-astro-data-formats.readthedocs.io/en/v0.3/data_storage/obs_index/index.html?highlight=obs%20table) which requires 8 mandatory arguments `(OBS_ID, OBS_MODE, RA_PNT, DEC_PNT, ALT_PNT, AZ_PNT, TSTART, TSTOP)` and foresees 33 optional arguments that can be present,

- and the selection methods of the class ObservationTable(Table) in gammapy/data/obs_table.py, which require in total 5 mandatory arguments `(OBS_ID, RA_PNT, DEC_PNT, TSTART, TSTOP),` which are also checked to be given by class ObservationTableChecker(Checker), also in gammapy/data/obs_table.py.

@bkhelifi made me aware, that the name of the observatory, e.g. in case of the two sides for CTAO, is also an important information, I think also treated in #3797 for the `location` in Observation.py, and that it should be added to the list of the mandatory arguments for the ObservationTable. Based on this, I would propose to have the following columns as mandatory for the internal gammapy ObservationTable read from a file of a certain format, in order to decouple it from GADF:


`OBSY_NAME, OBS_ID, RA_PNT, DEC_PNT, TSTART, TSTOP`.

Would you agree, that this could be the minimal set to go for? @bkhelifi further suggested to me, that all optional / complementary columns could then be read automatically and non-hardcoded into memory, if present in a file.

maxnoe:
Hi @LeanderSchlegel 

I think we cannot require RA_PNT / DEC_PNT but rather we should do the same as we do in the observations themselves:
require OBS_MODE and one pair of RA_PNT / DEC_PNT  or ALT_PNT / AZ_PNT depending on OBS_MODE.

Otherwise the drift scan instruments will be unhappy :)

bkhelifi:
[...]

PR #6128, opened 2025-09-02
Prototype for a new ObservationTable class

LeanderSchlegel:
Dear all,

@bkhelifi, @registerrier and me are working on a new design for the observation table in gammapy, based on the issue https://github.com/gammapy/gammapy/issues/4238 with the goal, to separate the observation table from the gadf format and support more file formats in the future.

This PR is intended as a draft containing a prototype and still **work in progress**. 

The codedesign is strongly based on the design of the class EventList and I took code from there (files: event_list.py and io.py) and adapted it for some functionalities.
- It specifies an **internal table model in the form of a reference table**,
- contains a **validator** to check that class instances follow this table format
-  and has a **reader** functionality, that can estimate the file format on disk and has the potential to call different
-  **converter functions** from a certain data format into the internal format.

The following changes are introduced in the class `ObservationTable:`

- A constructor is added, that instanciates the observation table from an astropy-table with a well defined internal format, specified in a reference table:
```
OBS_ID	RA_PNT	DEC_PNT	TSTART	TSTOP
          deg	  deg		
int64	float64	float64	Time	Time

```

- A validator checks on init, if the table to construct the observation table object from, is following the internal format. 

- The read method calls a `read()`-function in the file _io.py_ in a new class `ObservationTableReader,` based on `EventListReader`, and is described below.

A new class `ObservationTableReader:` is added in io.py:

- Offers a read function that can infer the file format from the HDU-header if not specified and calls a converter function between this file format and the internal format.

- Has a converter `from_gadf02_table`, that converts a gadf table into the internal format. It is non-strict on read and only the `OBS_ID` - column is required. If the other columns that are specified in the internal model are found, they are converted into the form of the internal model (units, types and TSTART and TSTOP are now astropy.time-objects). All other columns found in the file are added as additional columns without any changes, like before.

**Adaptions in data_store.py** were made, to adjust for the new API. These should be probably uncritical, but it would be great if the change in l.84 could be checked.

**A new typecasting function** is added in utils/types.py, that is based on numpys astype and can cast a value into a specific type.

The existing **test** in `test_obs_table.py` was designed for a gadf table format and can be still used nearly unchanged through using the converter internally. New tests checking the properties of the internal data model were added in the same file and to test the reader and converter, a test was added in `test_io.py`.

In total of gammapy, 8 tests are failing for now, due to the API change. This is because a writer into the new format is missing so far and because some data tables do not contain the now mandatory time-keywords (MJDREFI, ...), that are needed to construct the time-objects in the current approach. These have still to be fixed.

**Dear reviewer**
Any feedback on the design and implementation is very welcome. In the attached jupyter-notebook the class can be tested.
[Test_ObservationTable.ipynb.txt](https://github.com/user-attachments/files/22097175/Test_ObservationTable.ipynb.txt)

Thank you a lot,
Leander, Régis, Bruno

### DIFF
--- a/gammapy/data/__init__.py
+++ b/gammapy/data/__init__.py
@@ -9,7 +9,6 @@ from .gti import GTI
 from .hdu_index_table import HDUIndexTable
 from .metadata import EventListMetaData, ObservationMetaData, GTIMetaData
 from .obs_table import ObservationTable
-from .io import ObservationTableReader
 from .observations import Observation, Observations
 from .pointing import FixedPointingInfo, PointingInfo, PointingMode
 from .simulate import ObservationsEventsSampler
@@ -29,7 +28,6 @@ __all__ = [
     "Observations",
     "ObservationsEventsSampler",
     "ObservationTable",
-    "ObservationTableReader",
     "observatory_locations",
     "PointingInfo",
     "PointingMode",
--- a/gammapy/data/__init__.py
+++ b/gammapy/data/__init__.py
@@ -8,7 +8,7 @@ from .filters import ObservationFilter
 from .gti import GTI
 from .hdu_index_table import HDUIndexTable
 from .metadata import EventListMetaData, ObservationMetaData, GTIMetaData
-from .obs_table import ObservationTable, ObservationTableChecker
+from .obs_table import ObservationTable
 from .io import ObservationTableReader
 from .observations import Observation, Observations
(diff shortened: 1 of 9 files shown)


bkhelifi:
could you create a new PR? it would be simpler and will resolve the conflicts...
Thanks

registerrier:
Thanks @LeanderSchlegel ! Now that #6170 is merged. We can close this one.

Issue #5560, opened 2024-11-12
Use of `towncrier` for the change log


bkhelifi:
Based on the CTAO s/w programming conventions (see http://cta-computing.gitlab-pages.cta-observatory.org/documentation/developer-documentation/languages/python/index.html#), one should use  `towncrier` for the change log.

registerrier:
This is not a requirement per-se, rather a tool suggestion in the CTAO dev doc. We already comply with our changelog. 

Moving to towncrier can be a way to simplify the release process, so we should look into it. 
Changing the label.

PR #6010, opened 2025-07-24
Changelog for 2.0

AtreyeeS:
I started filling in the change log for 2.0, this is still in progress.
I know a lot of work was done for the infrastructure and documentation, but I missed a lot of it and will need some help in filling up the *Infrastructure* heading. Can you add a few bullet points @morcuended @Astro-Kirsty 

Additionally, if I am missing any other important features, please comment

PR #6130, opened 2025-09-05
Add towncrier

Astro-Kirsty:
This PR is to add towncrier.

It is a draft as I am still trying to figure out how everything works. My summary so far is:
- with towncrier we would need to add appropriate 'fragments' to each PR
   -  for example for a PR to add a file that relates to that

I managed to create a template similar to what we build now. I can run `towncrier build --version 2.0` locally and it will at least build the CHANGELOG.rst with the correct headings
Note: Jinja doesn’t have a built-in filter for ordinals. So we cannot have 26"th" easily. So I had to add a strange handling.

Closes https://github.com/gammapy/gammapy/issues/5560

### DIFF
--- a/.github/workflows/cffconvert.yml
+++ b/.github/workflows/cffconvert.yml
@@ -39,7 +39,7 @@ jobs:
           eossr-metadata-validator codemeta.json 
       
       - name: commit changes
-        uses: stefanzweifel/git-auto-commit-action@v6.0.1
+        uses: stefanzweifel/git-auto-commit-action@v7.0.0
         if: success()
         with:
           commit_author: GitHub Actions <actions@github.com>
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -157,7 +157,7 @@ jobs:
         run: |
           tox -e build_docs
       - name: Upload HTML output
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@v5
         with:
           name: gammapy-doc-html
           path: docs/_build/html/
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -14,6 +14,7 @@ repos:
     rev: v2.4.1
     hooks:
     - id: codespell
+      args: ["--ignore-words", "dev/codespell/ignore-words.txt"]
       additional_dependencies:
         - tomli
 
--- a/dev/authors.py
+++ b/dev/authors.py
@@ -150,7 +150,7 @@ def check_author_lists(since_last_lts):
 
     print(message + "\n  ".join(authors_annotated) + "\n")
 
-    print("(CC) = Coordination Committe, (AA) = Additional Authors, (OO) = Opted out\n")
+    print("(CC) = Coordination Committee, (AA) = Additional Authors, (OO) = Opted out\n")
 
 
 if __name__ == "__main__":
--- a/dev/codespell/ignore-words.txt
+++ b/dev/codespell/ignore-words.txt
@@ -1,8 +1,15 @@
 dne
+egal
 esy
 hist
+indx
 livetime
+livetimes
 nd
+ned
 merrors
+NotIn
+OptIn
+slac
 unparseable
 compiletime
--- a/dev/github_summary.py
+++ b/dev/github_summary.py
@@ -70,7 +70,7 @@ class GitHubInfoExtractor:
         return result
 
     def _extract_pull_request_info(self, pull_request):
-        """Builds a dictionary containing a list of summary informations.
+        """Builds a dictionary containing a list of summary information.
 
         Parameters
         ----------
--- a/docs/development/pigs/pig-009.rst
+++ b/docs/development/pigs/pig-009.rst
@@ -201,7 +201,7 @@ So far Gammapy only supports binned likelihood analysis and technically most
 use- cases for the event sampling could be solved with binned simulations. A
 binned simulation can be basically achieved by a call to
 ``numpy.random.poisson()`` based on the predicted number of counts map. This is
-conceptionally simpler as well as computationally more efficient than a sampling
+conceptually simpler as well as computationally more efficient than a sampling
 of event lists. In ``Gammapy`` a similar dataset simulation is already
 implemented in ``Dataset.fake()``, although this has a limited number of use
 cases than an event sampler. However, to support the full data access and data
--- a/docs/development/pigs/pig-017.rst
+++ b/docs/development/pigs/pig-017.rst
@@ -82,7 +82,7 @@ for the standard logging, but with a different severity level. The choice of the
 Data Model allows for interoperability with provenance information captured from *ctapipe* as well
 as with other astronomical archive services or software providing provenance information in the Virtual
 Observatory framework and with the W3C provenance models. In this optic, the development will be done on
-top of `ctapipe.core.provenance`_, re-using its present capabilities when possible and extending it. The
+top of `ctapipe.core.provenance`_, reusing its present capabilities when possible and extending it. The
 information captured will relate to the execution environment, the overall high-level interface configuration,
 the workflow execution process (i.e the relationships among datasets and processes and the parameter values
 used in each process step), and the serialisations of the intermediate and final datasets or objects produced.
--- a/docs/development/pigs/pig-020.rst
+++ b/docs/development/pigs/pig-020.rst
@@ -36,7 +36,7 @@ works as long as all objects reside in the same memory space.
 
 If datasets are distributed to different processes in future, it is technically
 complex and probably in-efficient to share model states between all sub-processes.
-It is conceptionally simpler if processes communicate with a single server process
+It is conceptually simpler if processes communicate with a single server process
 that contains the single global model object.
 
 The fundamental important difference to the current design is, that the model objects
--- a/docs/development/pigs/pig-024.rst
+++ b/docs/development/pigs/pig-024.rst
@@ -100,7 +100,7 @@ Releases
 
 Each Gammapy release should be associated with an updated list of authors that will be public on repository (Zenodo and
 HAL/SWH). This section is about feature releases. Bug releases will have as list of authors the one of its associated
-feature release whose rules are described here, plus the potentiel additional contributors.
+feature release whose rules are described here, plus the potential additional contributors.
 
 The list of authors is composed of people who contributed to the current release code
 by accepting the DCO. The history of merged pull requests will be used as starting point.
--- a/docs/development/pigs/pig-028.rst
+++ b/docs/development/pigs/pig-028.rst
@@ -37,7 +37,7 @@ Event type handling
 +++++++++++++++++++
 
 Starting from a datastore containing events lists and IRFs with types and classes of events,
-the user produces of a list of datasets per event type or class. Meta informations of
+the user produces of a list of datasets per event type or class. Meta information of
 the datasets allow for complex types handling at modeling/fitting step (e.g. joint fit
 of A & B type, stack all type together etc).
 
@@ -154,7 +154,7 @@ We discuss here some aspects that should be explored.
   ``MapAxes`` and an interpolator. In addition, ``Maps`` use the ``Geom`` object to
   represent the spherical coordinates.
 
-  - Having a common data structure could help make maps fully re-usable for IRFs.
+  - Having a common data structure could help make maps fully reusable for IRFs.
     This might be a common use case with pyirf.
   - One could allow ``Maps`` and ``MapCoord`` objects without spatial axes.
     Introducing specialized spatial axes such as `WcsMapAxis`, `RegionMapAxis` or
@@ -180,7 +180,7 @@ This is problematic for the following reasons:
 - This prevents the support of multiple formats, since the internal data structure
 is tied to one specific format.
 - Data is not in the optimal in-memory representation. For instance, times should be
-stored as astropy.time.Time instances, and coodinates as `SkyCoord`.
+stored as astropy.time.Time instances, and coordinates as `SkyCoord`.
 - Data is not validated on input. Errors can happen deep into the code for something
 that could have been caught on input file reading or object creation.
 - Writing data out is harder
@@ -245,8 +245,8 @@ Main documentation
 - More detailed
 - Use type hints in Gammapy everywhere?
 
-Gammapy-recipes and additional ressources
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Gammapy-recipes and additional resources
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 The Gammapy-recipes gallery offers a nice additional source of tutorials for advanced or non
 standard use cases.
@@ -265,8 +265,8 @@ Infrastructure
 
 - Improve test coverage and quality.
 - Improve our tools helping to the creation of releases
-- Creation of Docker images with an automatized tool
-- Re-use docker image for Binder, this config here: https://github.com/gammapy/gammapy-webpage/tree/v1.0rc1 already creates docker image in Binder.
+- Creation of Docker images with an automated tool
+- Reuse docker image for Binder, this config here: https://github.com/gammapy/gammapy-webpage/tree/v1.0rc1 already creates docker image in Binder.
 
 Distributed Computing and Performance
 +++++++++++++++++++++++++++++++++++++
--- a/examples/models/spatial/plot_piecewise_norm_spatial.py
+++ b/examples/models/spatial/plot_piecewise_norm_spatial.py
@@ -6,7 +6,7 @@ Piecewise norm spatial model
 
 This model parametrises a piecewise spatial correction
 with a free norm parameter at each fixed node in longitude, latitude
-and optionaly energy.
+and optionally energy.
 """
 
 # %%
--- a/examples/tutorials/analysis-3d/flux_profiles.py
+++ b/examples/tutorials/analysis-3d/flux_profiles.py
@@ -265,7 +265,7 @@ flux_profile_estimator = FluxProfileEstimator(
     selection_optional=["ul", "scan"],
 )
 ######################################################################
-# The configuration of the fit statistic profile is done throught the norm parameter:
+# The configuration of the fit statistic profile is done through the norm parameter:
 flux_profile_estimator.norm.scan_values = np.linspace(-1, 5, 11)
 
 ######################################################################
--- a/examples/tutorials/details/maps.py
+++ b/examples/tutorials/details/maps.py
@@ -304,7 +304,7 @@ map_4d = Map.create(
 print(map_4d.geom)
 
 ######################################################################
-# It is possible to utilise the `~gammapy.maps.TimeMapAxis.slice` attrribute
+# It is possible to utilise the `~gammapy.maps.TimeMapAxis.slice` attribute
 # to create new a `~gammapy.maps.TimeMapAxis`. Here we are slicing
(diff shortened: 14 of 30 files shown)


morcuended:
Thanks @Astro-Kirsty! This will be very helpful

Some points that should also be done:
- Add howto towncrier in dev docs
- Add `no-changelog-needed` label to make the towncrier check be skipped when needed for minor changes

docs/conf.py
```diff
@@ -103,6 +103,7 @@ def setup(app):
     'sphinx_copybutton',
     "sphinx_design",
     "sphinx_gallery.gen_gallery",
+    "sphinx_changelog",
```
registerrier, line 105: Maybe you need to add `sphinx-changelog` to setup.cfg?

docs/development/dev_howto.rst
```diff
@@ -569,12 +569,21 @@ As explained in the
 of the Astropy docs, there are (at least) two approaches for adding to the releases,
 each with pros and cons.
 
-We've had some pain due to merge conflicts in the releases notes and having to wait
-until the contributor rebases (and having to explain git rebase to new contributors).
+We recently chose to utilise `towncrier <https://towncrier.readthedocs.io/en/stable/>`__ for our release notes.
+
+- For each PR, a related 'fragment' file should be created in the ``docs/release-notes`` folder.
+- The naming convention of the file should be ``<PULL REQUEST NUMBER>.<TYPE>.rst``, where the available
+  types are ``infrastructure``, ``docs``, ``feature`` or ``bugfix``.
```
registerrier, line 581: Can't we use a structure following the sub-package structure with e.g.:
 `data`, `irf`, `makers`, `datasets`, `modeling`, `estimators`, `maps`, `others`, `docs`.

docs/development/dev_howto.rst
```diff
@@ -569,12 +569,21 @@ As explained in the
...
+- The file should contain a suitable message for the PR, for example "Add a new feature for the estimators.".
```
registerrier: Well this one is too vague. Maybe a one or two lines of synthetic description of the change introduced.

docs/release-notes/template.rst
```diff
@@ -0,0 +1,67 @@
...
+
+- Released {{ day }}{{ suffix }} {{ month_name }}, {{ year }}
+- contributors
+- pull requests since v (not all listed below)
+- closed issues
+
+Summary
+-------
+
+
```
registerrier: If the summary template is completely defined from the beginning, is there a way to add something like a chapeau over the whole list of changes? e.g. what's new?

docs/development/dev_howto.rst
```diff
@@ -569,12 +569,21 @@ As explained in the
...
+
+To build the release notes you can simply utilise the following line:
+
+.. code-block:: bash
+
+    towncrier build --version=<VERSION NUMBER>
+
```
registerrier, line 589: I suppose this will require an additional step during the release process as well.

registerrier (commented):
Thanks @Astro-Kirsty . This looks nice. An important design choice we have to make is the categorization of entries in the changelog.
Is it possible to have a section on top that is not pre-filled in the template, such as a what's new?
Is it possible to impose the entry in the changelog with a `require-changelog` label rather than the opposite?

I have left some comments inline.

docs/development/dev_howto.rst
Astro-Kirsty: We definitely could. We should discuss further how we want the release notes page to look.
For now, my goal was to be similar to what we currently have.

Also from looking at other packages the 'norm' is to have them by the type of issue they are

AtreyeeS: This file will be automatically created, or it has to be done by the contributor?

docs/release-notes/template.rst
Astro-Kirsty: I'm not sure I completely understand, but with `towncrier` we would still have to change parts of the `CHANGELOG.rst` once we build it. 
i.e. to include our own `Summary` like we did in the past

Astro-Kirsty:
So everyone can have an idea of how this renders. I added a number of fragments and then ran:
`towncrier build --version=example --keep`

Note that the default is to `delete` which removes all the fragments. 
The resulting changelog is this:
[CHANGELOG.zip](https://github.com/user-attachments/files/22425501/CHANGELOG.zip)

docs/development/dev_howto.rst
Astro-Kirsty: It needs to be included by the contributor

Astro-Kirsty:
From the dev call 19/09/25:

- add label 'changelog-needed'
    - require this label for feature, bugfix and api
- add workflow to CI such that it checks for this

Example of fragment files:
https://github.com/pytest-dev/pytest/tree/main/changelog

Astro-Kirsty:
Update:
- added a label `changelog-required`
- added a work flow to check if the label is there that a fragment is required. It also checks if the labels `api-change`, `bug` or `feature` are present, and if the `changelog-required` is not present, the CI will fail. We could also have the possibility to add the label automatically (https://github.com/actions-ecosystem/action-add-labels) but I think it is better to require the PR maker to do that.
- this workflow runs with this PR as we don't have the label and neither the others

maxnoe:
You should update the docs to use `sphinx-changelog` to have the changelog rendered on main builds with the changes received so far.

This also tests that the changelog fragments are actually valid rst when rendered.

See: https://github.com/astropy/astropy/blob/main/docs/changelog.rst

docs/development/dev_howto.rst
```diff
@@ -569,12 +569,23 @@ As explained in the
 of the Astropy docs, there are (at least) two approaches for adding to the releases,
 each with pros and cons.
 
-We've had some pain due to merge conflicts in the releases notes and having to wait
-until the contributor rebases (and having to explain git rebase to new contributors).
+We recently chose to utilise `towncrier <https://towncrier.readthedocs.io/en/stable/>`__ for our release notes.
+
+- For each PR, a related 'fragment' file should be created in the ``docs/release-notes`` folder.
+- The naming convention of the file should be ``<PULL REQUEST NUMBER>.<TYPE>.rst``, where the available
```
AtreyeeS, line 580: is the `type` necessary in the name? Can't it be interfered from the GitHub label?

docs/release-notes/template.rst
```diff
@@ -0,0 +1,67 @@
...
+{% set parts = versiondata.date.split('-') %}
+{% set year = parts[0] %}
+{% set month = parts[1]|int %}
+{% set day = parts[2]|int %}
+{% set month_name = ["January","February","March","April","May","June","July","August","September","October","November","December"][month-1] %}
+
+{% set suffix = "th" %}
+{% if day in [1,21,31] %}
+  {% set suffix = "st" %}
```
AtreyeeS, line 13: We can put the data as MM-DD-YYYY to avoid these lines...

AtreyeeS (approved):
Thanks @Astro-Kirsty ! I propose we merge this and test how it goes.

The release notes will be created before the release, or we can also get notes for the dev version?

Another question is, for a bug fix of say 2.0.1 , will that PR also go in 2.1 changelog?

maxnoe:
> The release notes will be created before the release, or we can also get notes for the dev version?

That's what I commented above, using `sphinx-changelog` will take the currently existing snippets and render the changelog for the dev version.

bkhelifi (dismissed):
So great and technical! Thanks @Astro-Kirsty

Astro-Kirsty:
> You should update the docs to use `sphinx-changelog` to have the changelog rendered on main builds with the changes received so far.
> 
> This also tests that the changelog fragments are actually valid rst when rendered.
> 
> See: https://github.com/astropy/astropy/blob/main/docs/changelog.rst

`sphinx-changelog` is already included here

docs/development/dev_howto.rst
Astro-Kirsty: Yes we need to have the type. That is the important part.

The label is not as important, but we thought it would be nice that if there is a specific label, that we should expect to have a fragment

docs/release-notes/template.rst
Astro-Kirsty: We could indeed, but personally I like the format being in words

maxnoe:
In the dependencies yes, but I don't see it used in the docs. Maybe I missed it

docs/development/dev_howto.rst
```diff
@@ -569,12 +569,23 @@ As explained in the
...
+  types are ``infrastructure``, ``docs``, ``feature`` or ``bugfix``.
+- The file should contain a suitable message for the PR, for example "A new function `select_nested_models`
+  has been introduced to perform nested model fits and compute the resulting test statistic (TS) between
+  two nested hypotheses.".
+
+To build the release notes you can simply utilise the following line:
```
maxnoe: I think this should go into the release procedure, not into the developer docs, this page here:

https://docs.gammapy.org/2.0/development/release.html

Astro-Kirsty: Yes that is a good idea

docs/development/release.rst
```diff
@@ -48,6 +42,12 @@ Feature Freeze and Branching
     git checkout -B main upstream/main
 
 #. From the github online interface, create a new branch v<version>.x
+
+#. Stay on the ``main`` branch to build the changelog for the version you are about to release.
+   This is done through the use of towncrier with the following line::
+
+    towncrier build --version=<version>
```
Astro-Kirsty: I think we can just stay with creating the change log at the Feature Freeze and Branching Phase. Or do we want to include it instead only at the major release?

maxnoe: The usual workflow is to open a PR with the rendered changelog just before doing a release.

Astro-Kirsty: Okay so probably to open the PR at the release candidate stage

docs/release-notes/template.rst
maxnoe: @registerrier We almost never add the towncrier changelog unedited. The developer preparing the release renders the changelog using towncrier and uses that as a starting point. We usually add a summary of the highlights at the top and e.g. remove empty sections and also re-order items for priority.

Astro-Kirsty:
> Another question is, for a bug fix of say 2.0.1 , will that PR also go in 2.1 changelog?

If we want to include the bug fix items from 2.0.1 in 2.1, we simply need to create the changelog using the `keep` keyword. As I have now described under the bug fix release instructions

maxnoe:
> Another question is, for a bug fix of say 2.0.1 , will that PR also go in 2.1 changelog?
> If we want to include the bug fix items from 2.0.1 in 2.1, we simply need to create the changelog using the keep keyword. 
> As I have now described under the bug fix release instructions

I don't think this is how it is supposed to work. Pull requests add snippets of changes. You render the changelog for the current release line on main. For bugfix releases you render the changelog on the maintenance branch for that release line. 
No need to `keep` snippets, as they will not be removed on the maintenance branch when you render the changelog on the main. That's actually the main advantage of towncrier, the changelog snippets get backported along with the changes.

Astro-Kirsty:
> > Another question is, for a bug fix of say 2.0.1 , will that PR also go in 2.1 changelog?
> > If we want to include the bug fix items from 2.0.1 in 2.1, we simply need to create the changelog using the keep keyword.
> > As I have now described under the bug fix release instructions
> 
> I don't think this is how it is supposed to work. Pull requests add snippets of changes. You render the changelog for the current release line on main. For bugfix releases you render the changelog on the maintenance branch for that release line. No need to `keep` snippets, as they will not be removed on the maintenance branch when you render the changelog on the main. That's actually the main advantage of towncrier, the changelog snippets get backported along with the changes.

Ahh yes, thanks @maxnoe. That is very true!

docs/development/dev_howto.rst
```diff
@@ -569,12 +570,17 @@ As explained in the
 of the Astropy docs, there are (at least) two approaches for adding to the releases,
 each with pros and cons.
 
-We've had some pain due to merge conflicts in the releases notes and having to wait
-until the contributor rebases (and having to explain git rebase to new contributors).
+We recently chose to utilise `towncrier <https://towncrier.readthedocs.io/en/stable/>`__ for our release notes.
```
Tobychev: Don't forget to set a calendar reminder to change this "recently" once we reach 2026 ;)

docs/development/dev_howto.rst
```diff
@@ -569,12 +570,17 @@ As explained in the
...
+
+- For each PR, a related 'fragment' file should be created in the ``docs/release-notes`` folder.
+- The naming convention of the file should be ``<PULL REQUEST NUMBER>.<TYPE>.rst``, where the available
```
Tobychev, line 580: It would be good to add a note along the line of "Yes, you will need to create the pull request before you can properly name the changelog file."

Astro-Kirsty: I don't think that is necessary because the PR numbers are sequential

docs/development/release.rst
```diff
@@ -23,18 +23,19 @@ Feature Freeze and Branching
...
-   * You can use the helper script ``dev/authors.py`` for this.
-#. Open a PR including both changes and mark it with the ``backport-v<version>.x`` label.
-   Gather feedback from the Gammapy user and dev community and finally merge and backport to the
-   ``v<version>.x`` branch.
+#. On the ``main`` branch, build the changelog for the version you are about to release.
+   This is done through the use of towncrier with the following line::
+
+    towncrier build --version=<version>
+
+#. Open two separate PRs for each of these changes and mark each with the ``backport-v<version>.x`` label.
```
maxnoe, line 43: I think this is misleading now. You do not want to backport the changelog.

docs/development/release.rst
```diff
@@ -23,18 +23,19 @@ Feature Freeze and Branching
...
-     need to use your github token here.
+    * You can use the helper script ``dev/authors.py`` for this.
 
-#. Update the author list manually in the  ``CITATION.cff``.
 
-   * You can use the helper script ``dev/authors.py`` for this.
-#. Open a PR including both changes and mark it with the ``backport-v<version>.x`` label.
-   Gather feedback from the Gammapy user and dev community and finally merge and backport to the
-   ``v<version>.x`` branch.
+#. On the ``main`` branch, build the changelog for the version you are about to release.
```
maxnoe, line 31: This should be "on the release branch", right? I.e. main for current, and the maintenance branch for LTS

maxnoe:
I think the inclusion of the dev changelog by `sphinx-changelog` is still missing.

you need a page like this in docs and link it in the toctree: https://github.com/astropy/astropy/blob/main/docs/changelog.rst

docs/development/release.rst
```diff
@@ -23,18 +23,19 @@ Feature Freeze and Branching
 
 **A few days before the feature freeze:**
 
-#. Fill the changelog ``docs/release-notes/<version>.rst`` for the version you are about to release.
+#. Update the author list manually in the  ``CITATION.cff``.
 
-   * To generate the list of pull requests and issues, and list of authors run
```
AtreyeeS, line 28: So you are removing the list of authors from the changelog?

I think it might be still good to keep the names of authors who contributed to a specific release (it is of course available in Zenodo, but a quick look is nice), and also mention the number of new contributors. Maybe we can still mention the GitHub_summary.py file, but update it in a separate PR to be specific to our purposes

docs/development/release.rst
Astro-Kirsty: In the past we have done it like this.
i.e. [see here](https://github.com/gammapy/gammapy/pull/6010)

docs/development/release.rst
Astro-Kirsty: It will be 'main' at this points if we keep the instructions as they are. Otherwise we should move this entire description to releasing the first major candidate. 
It would make sense to me to open this PR early  (like we did in the past) so that we can make the relative changes

docs/development/release.rst
maxnoe: I find that PR confusing. Why do you need to backport the changelog of 2.0.0 to the 2.0.x maintenance branch?

The 2.0.x branch should only be created after 2.0.0 has been tagged on main, with the first commit on it being the 2.0.0 release tag. Then the changelog is already included.

docs/development/release.rst
Astro-Kirsty: Yes we would still want to include the list of authors, but I imagine we do not want to run that entire piece of code. Rather that we reduce the code to only include the contributors part. 

Should we include that in this PR or?

docs/development/release.rst
maxnoe: The order I would expect is:

1. Render changelog for 2.0.0
2. Tag 2.0.0
3. Create maintenance branch `v2.0.x` starting from the tag: `git switch -c v2.0.x origin/v2.0.0`

Then you create new commits on main, that include changelog snippets.

Some of these commits are backported to the maintenance branch, which includes the snippets.

For a bugfix release, you render the changelog on the 2.0.x branch which will collect the relevant snippets there.

For a feature release, you render the changelog on the main branch, which will also collect the relevant snippets there.

At no point do you need to backport an already rendered changelog.

registerrier: Technically, we create the 2.0.x branch at the feature freeze and the 2.0.0 tag is made on this specific branch. It follows the astropy procedure I think (except maybe that we don't tag the feature freeze).

maxnoe: This makes the whole thing a bit more complicated, as the commit you tag would never be on the main branch.

It would require actually backporting (or rather front-porting) the changelog from the release branch to main.

maxnoe: Or rendering on main in the moment you branch, backporting to the release branch.

maxnoe: Maybe the procedure is simplest if you render the changelog in the moment of the feature freeze and use the commit of the rendered changelog to branch off the release branch?

Astro-Kirsty:
As per the dev call we decided to proceed with the current implementation and adapt when necessary (possibly at bug release or next release). 

I created an issue related to needing the list of contributors

docs/development/dev_howto.rst
morcuended: I propose to simply say here "We utilise towncrier..."

docs/development/dev_howto.rst
morcuended, line 580: Maybe say that this will not be done for all PRs, but the lead developers will mark certain PRs to need the changelog fragment entries.

bkhelifi (approved):
Thanks a lot, @Astro-Kirsty . This will render our change log very clear for users

.github/workflows/changelog.yml
```diff
@@ -0,0 +1,40 @@
...
+
+      - name: Verify changelog label and fragment
+        run: |
+          set -euo pipefail
+
+          # Obtain all labels set to the PR
+          labels="${{ toJson(github.event.pull_request.labels.*.name) }}"
+
+          # If the labels contains feature, bug or api-change then automatically require the 'changelog-required' label
+          if echo "$labels" | grep -Eq '"(feature|bug|api-change)"'; then
```
natthan-pigoux, line 21: Should you include the `infrastructure` label as well here? I see that it is one of the fragment types mentionned in the docs:
>   types are ``infrastructure``, ``docs``, ``feature`` or ``bugfix``.

Astro-Kirsty: We decided not to because not all infrastructure is important to be added to the changelog (not always useful for users -- same as the docs)

natthan-pigoux: Alright, sorry if missed these discussions...

Astro-Kirsty: No worries, it was during the dev call and not properly documented here

registerrier (approved):
Let's merge now and see how this impacts the PR workflow.

PR #6136, opened 2025-09-09
Add source to FPE in upper limits tutorial

Astro-Kirsty:
We should specify the source of interest in this tutorial so that the user knows this is needed. 

I also remove the 'copy' of the model. It is unclear to me, why we did that?

Another important point: shouldn't we include a bkg model too?


examples/tutorials/analysis-3d/non_detected_source.py
```diff
@@ -199,18 +199,18 @@
 # from wandering off to different regions in the FoV.
 #
 
-
-model1 = sky_model.copy(name="model1")
```
AtreyeeS, line 203: It can be useful to keep a copy of the best fit model, because we are putting in something completely different for the flux point estimation

Astro-Kirsty: Yes but then we reset the values, so I'm not sure it makes sense to have the unnecessary copy

QRemy (approved):
Thanks @Astro-Kirsty, looks good. I agree the copy seems unnecessary.

PR #5975, opened 2025-07-09
Allow configuration of a size_factor in Models.to_regions

AtreyeeS:
The `to_region` functions for Gaussian Spatial models allow one to configure the containment fraction. However, the same argument cannot be passed while going through `Models.to_regions`

My implementation might be a bit unclean with adding the parameter to the dict and then removing it again, but I could not find a better way to do it without breaking the API

PR #6137, opened 2025-09-09
Adapt size_factor value for error

Astro-Kirsty:
We missed adapting the `size_factor` to be consistent with the errors models in PR #5975


gammapy/modeling/models/spatial.py
```diff
@@ -417,11 +422,15 @@ def plot_error(
                 * "position": plot the position error of the spatial model
                 * "extension": plot the extension error of the spatial model
 
+        size_factor : float
+            Number of :math:`\sigma`.
+            Default is :math:`1.0\sigma` which corresponds to about 39%
+            containment for a 2D symmetric Gaussian.
```
QRemy: this definition is valid only for a Gaussian while this method applies to all models

Astro-Kirsty: True, but then how can we describe it globally?

AtreyeeS: The wording here is also confusing. One may confuse it as the width of the error bar.
Maybe say "Number of size_factor around which the 1-sigma error is to be plotted"?

Astro-Kirsty: It feels odd to describe the parameter using the parameter maybe:
"Multiplicative factor for the 1-sigma error band. "

gammapy/modeling/models/spatial.py
```diff
@@ -417,11 +422,14 @@ def plot_error(
                 * "position": plot the position error of the spatial model
                 * "extension": plot the extension error of the spatial model
 
+        size_factor : float, optional
+            Multiplicative factor for the 1:math:`\sigma` error band.
+            Default is :math:`1.0\sigma`
```
AtreyeeS:
```suggestion
            Default is :math:`1.0`
```

AtreyeeS: Maybe mention "should be the same as used for `plot`"

gammapy/modeling/models/spatial.py
```diff
@@ -417,11 +422,14 @@ def plot_error(
                 * "position": plot the position error of the spatial model
                 * "extension": plot the extension error of the spatial model
 
+        size_factor : float, optional
+            Multiplicative factor for the 1:math:`\sigma` error band.
```
AtreyeeS:
```suggestion
            Extension around which the 1:math:`\sigma` error band is to be plotted. 
```

gammapy/modeling/models/spatial.py
```diff
@@ -665,10 +673,12 @@ def to_region(self, size_factor=1.0, **kwargs):
 
         Parameters
         ----------
-        size_factor : float
-            Number of :math:`\sigma
+        size_factor : float, optional
+            Number of :math:`\sigma`.
```
AtreyeeS, line 688: Adapt as for plot_error

gammapy/modeling/models/spatial.py
```diff
@@ -665,10 +673,12 @@ def to_region(self, size_factor=1.0, **kwargs):
...
             Default is :math:`1.0\sigma` which corresponds to about 39%
```
AtreyeeS, line 689: Not required

Astro-Kirsty: This is attached to the `GaussianSpatialModel`, so I think we should keep it

gammapy/modeling/models/spatial.py
Astro-Kirsty: I would not mention the other method as one might not always use both. As the default is the same, the user will see very quickly if they change it and its not the same

gammapy/modeling/models/spatial.py
Astro-Kirsty: This is specific to `GaussianSpatialModel` so I would not adapt it to be the same as `plot_error` which is generic.
This is also for the `to_region` functionality, so it has nothing to do with the error

gammapy/modeling/models/spatial.py
AtreyeeS: My only point is that this should not be confused as the 1-sigma error width

Astro-Kirsty: Yes but this is only for the `to_region` functionality. Nothing to do with any error
I did not adjust anything here just added the term optional.

gammapy/modeling/models/spatial.py
```diff
@@ -441,8 +449,13 @@ def plot_error(
...
             self.plot_position_error(ax, **kwargs_position)
-
-            region = self._to_region_error()
+            try:
+                region = self._to_region_error(size_factor=size_factor)
+            except TypeError as e:
+                if "unexpected keyword argument 'size_factor'" in str(e):
+                    region = self._to_region_error()
+                else:
+                    raise
```
Astro-Kirsty, line 458: I had to add this because we have some functions that don't have `size_factor`, for example the `PointSpatialModel` and the `DiskSpatialModel`.
The other option is to do a `inspect.signature`

```
    signature = inspect.signature(self._to_region_error)
    if "size_factor" in signature.parameters:
        return self._to_region_error(size_factor=size_factor)
    else:
        return self._to_region_error()
```

gammapy/modeling/models/spatial.py
```diff
@@ -950,13 +973,13 @@ def evaluate(lon, lat, lon_0, lat_0, r_0, e, phi, edge_width):
         )
         return u.Quantity(norm * in_ellipse, "sr-1", copy=COPY_IF_NEEDED)
 
-    def to_region(self, size_factor=1.0, **kwargs):
+    def to_region(self, **kwargs):
```
Astro-Kirsty: We never supported the `x_sigma` value previously, I'm not sure why it was added and if it really should?

gammapy/modeling/models/spatial.py
```diff
@@ -417,11 +422,14 @@ def plot_error(
                 * "position": plot the position error of the spatial model
                 * "extension": plot the extension error of the spatial model
 
+        size_factor : float, optional
+            Extension around which the 1:math:`\sigma` error band is plotted.
```
registerrier, line 426: 1 sigma is meaningful only for gaussian-like models.

registerrier: Does the factor also applies for position errors?

gammapy/modeling/models/spatial.py
QRemy: Might be for API consistancy : all the `to_region` methods should have a `size_factor` if added to one.

Astro-Kirsty:
The original purpose of this PR was to solve the fact that the `_to_region_error` did not have the size factor value changed (https://github.com/gammapy/gammapy/pull/6137/files#diff-a8a011395c385be1ef1d20aee8c2fe43c0c13c8b09517e298d1fe9209b12e764R714) resulting in plots like this: 
spatial_model_plotting_v1 3
spatial_model_plotting_v2 0

For this code snippet:
```
ax.add_artist(model_gauss_1.to_region(x_sigma=1.5).to_pixel(ax.wcs).as_artist(color=color))
model_gauss_1.plot_error(ax=ax, which="extension", kwargs_extension={"facecolor": color, "edgecolor":color})
```

And the `plot_error` functionality does not have an option for `size_factor`.

AtreyeeS:
Thanks @Astro-Kirsty ! This example shows that we need to have a `size_factor` option on the `plot_regions_error` to configure where the error band is to be plotted.

I also find it a API confusing that 
 - `SpatialModel.plot` plots a colour image of the full spatial model
 - `SpatialModel.plot_error` plots the error region of the spatial model
 - `Models.plot_regions`  plots the regions of the spatial models

  We can introduce a `SpatialModel.plot_region` for simplicity, but I think this motivates the introduction of the plotters pig.

Astro-Kirsty:
Yes exactly, but to do this we need the `size_factor` option  on `_to_region_error` for each model. The question is whether we want to have that on every model for API consistency, or just the ones that actually require it.

AtreyeeS:
> Yes exactly, but to do this we need the `size_factor` option on `_to_region_error` for each model. The question is whether we want to have that on every model for API consistency, or just the ones that actually require it.

Since we have `size_factor` option on `_to_region` for all models, I would say we should have the same for `to_region_error` as well

AtreyeeS:
A side question on the `to_region_error` function
This only takes into the consideration the error on the "size" parameter, ie, the `sigma` or `r_0`
I suppose it should also consider the error on the position, and other free parameters if present, eg ellipticities...

Astro-Kirsty:
> A side question on the `to_region_error` function This only takes into the consideration the error on the "size" parameter, ie, the `sigma` or `r_0` I suppose it should also consider the error on the position, and other free parameters if present, eg ellipticities...

I agree that for the other parameters it would be nice, but then this is really a extra feature. So I would suggest that for another PR. In this PR would should really deal with the 'bug' like behaviour so that it can be included as soon as possible for v2.0.1

Astro-Kirsty:
> > Yes exactly, but to do this we need the `size_factor` option on `_to_region_error` for each model. The question is whether we want to have that on every model for API consistency, or just the ones that actually require it.
> 
> Since we have `size_factor` option on `_to_region` for all models, I would say we should have the same for `to_region_error` as well

I have adjusted the PR for API consistency.

Astro-Kirsty:
CI fail is unrelated (also seen in other PRs). This is ready to be reviewed

gammapy/modeling/models/spatial.py
bkhelifi: So, this is a kwargs associated to our spatial model functions `_to_region_error` that are not exposed. So sad fot the user.

gammapy/modeling/models/spatial.py
bkhelifi, line 426: for me, it is not so clear! Is it a fraction of the sigma value that is?

bkhelifi:
I think that this size factor is weird, in depth...
For a Generalised Gaussian, I understand the concept, and then the kwarg name is `n_sigma`. For other models, what is it???
In the case of a Generalised Gaussian, if `size_factor`=`n_sigma` is 2, the error seems to be multiplied by 2. But is it really correct? Does it not depend on the correlation matrix?

Astro-Kirsty:
> I think that this size factor is weird, in depth... For a Generalised Gaussian, I understand the concept, and then the kwarg name is `n_sigma`. For other models, what is it??? In the case of a Generalised Gaussian, if `size_factor`=`n_sigma` is 2, the error seems to be multiplied by 2. But is it really correct? Does it not depend on the correlation matrix?

Yes I agree with you, it is not clear exactly what it is and if we should adjust that, but I would suggest to leave that for another PR as that would be more of a feature. This is simply a bug fix because we cannot have the two different values everywhere as it does not plot well

gammapy/modeling/models/spatial.py
Astro-Kirsty: It used to be "Multiplicative factor" but as you can see from the other reviews, it was changed

bkhelifi: I start to be confused...
`_to_region_error` says that size_factor is the number of sigma! This is not what it is explained here.

Astro-Kirsty: Yes but in this section `plot_error` is the generic function for all spatial models, where as in `_to_region_error` i.e. line 714 it is specifically for the `GaussianSpatialModel`

bkhelifi: Which comes back to the question of an associated PR to create rationale between the whole...

Astro-Kirsty: No, we only have it on the extension

Astro-Kirsty: > 1 sigma is meaningful only for gaussian-like models.

Yes true! Any suggestion for the wording? It is difficult for it to encompass all

PR #6141, opened 2025-09-15
Add sensitivity for other quantities of the flux maps

### LINKED ISSUE #6140: Addition of sensitivity for other quantities
We currently have access to the `flux_sensitivity` but do not support the other quantities having a sensitivity i.e. `dnde_sensitivity`. 

We should add those into the options for the `FluxMaps`.

This is a simple addition. It would also help to solve the 'issue' with the flux upper points tutorial, where we utilise `flux` [here](https://docs.gammapy.org/dev/tutorials/analysis-3d/non_detected_source.html#sensitivity-estimation) but we are not really comparing the correct values. 
We should rather be comparing the `dnde_sensitivity`

Astro-Kirsty:
This resolves #6140

### DIFF
--- a/gammapy/estimators/map/core.py
+++ b/gammapy/estimators/map/core.py
@@ -684,7 +684,7 @@ class FluxMaps:
 
     @property
     def dnde_sensitivity(self):
-        """Sensitivity given as the dnde for which the significance is ``self.meta["n_sigma_sensitivity]``."""
+        """Sensitivity given as dnde for which the significance is ``self.meta["n_sigma_sensitivity]``."""
         return self.norm_sensitivity * self.dnde_ref
 
     @property
@@ -714,7 +714,7 @@ class FluxMaps:
 
     @property
     def e2dnde_sensitivity(self):
-        """Sensitivity given as the e2dnde for which the significance is ``self.meta["n_sigma_sensitivity]``."""
+        """Sensitivity given as e2dnde for which the significance is ``self.meta["n_sigma_sensitivity]``."""
         return self.norm_sensitivity * self.e2dnde_ref
 
     @property
--- a/gammapy/estimators/map/core.py
+++ b/gammapy/estimators/map/core.py
@@ -56,10 +56,16 @@ REQUIRED_COLUMNS = {
 REQUIRED_QUANTITIES_SCAN = ["stat_scan", "stat"]
 
 OPTIONAL_QUANTITIES = {
-    "dnde": ["dnde_err", "dnde_errp", "dnde_errn", "dnde_ul"],
-    "e2dnde": ["e2dnde_err", "e2dnde_errp", "e2dnde_errn", "e2dnde_ul"],
+    "dnde": ["dnde_err", "dnde_errp", "dnde_errn", "dnde_ul", "dnde_sensitivity"],
+    "e2dnde": [
+        "e2dnde_err",
+        "e2dnde_errp",
+        "e2dnde_errn",
+        "e2dnde_ul",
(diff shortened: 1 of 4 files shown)


Astro-Kirsty:
I was thinking about this more and I'm not sure it makes sense to support all quantities, maybe just the `dnde`?

QRemy:
> I was thinking about this more and I'm not sure it makes sense to support all quantities, maybe just the `dnde`?

I think it's ok if we have everything for all the optional quantities.

QRemy (dismissed):
Thanks @Astro-Kirsty, looks good.

gammapy/estimators/map/tests/test_excess.py
```diff
@@ -161,15 +161,23 @@ def test_significance_map_estimator_map_dataset(simple_dataset):
 
     assert_allclose(result["norm_sensitivity"].data[0, 10, 10], 48.997699, atol=1e-3)
     assert_allclose(result["flux_sensitivity"].data[0, 10, 10], 4.850772e-10, rtol=1e-4)
+    assert_allclose(
+        result["eflux_sensitivity"].data[0, 10, 10], 2.256427e-10, rtol=1e-4
+    )
+    assert_allclose(result["dnde_sensitivity"].data[0, 10, 10], 4.89977e-11, atol=1e-3)
+    assert_allclose(
+        result["e2dnde_sensitivity"].data[0, 10, 10], 4.89977e-11, atol=1e-3
```
AtreyeeS, line 169: is it a coincidence this value is exactly equal to the `dude_sensitivity`?

Astro-Kirsty: No not at all, its because the reference energy is 1TeV

gammapy/estimators/map/core.py
```diff
@@ -676,6 +682,11 @@ def dnde_ul(self):
         """Return differential flux (dnde) SED upper limit."""
         return self._use_center_as_labels(self.norm_ul * self.dnde_ref)
 
+    @property
+    def dnde_sensitivity(self):
+        """Sensitivity given as dnde for which the significance is ``self.meta["n_sigma_sensitivity]``."""
```
registerrier: why `"n_sigma_sensitivity`?

gammapy/estimators/map/core.py
```diff
@@ -701,6 +712,11 @@ def e2dnde_ul(self):
         """Return differential energy flux (e2dnde) SED upper limit."""
         return self._use_center_as_labels(self.norm_ul * self.e2dnde_ref)
 
+    @property
+    def e2dnde_sensitivity(self):
+        """Sensitivity given as e2dnde for which the significance is ``self.meta["n_sigma_sensitivity]``."""
```
registerrier: same here

gammapy/estimators/map/core.py
```diff
@@ -756,6 +772,11 @@ def eflux_ul(self):
         """Return energy flux (eflux) SED upper limits."""
         return self.norm_ul * self.eflux_ref
 
+    @property
+    def eflux_sensitivity(self):
+        """Sensitivity given as the eflux for which the significance is ``self.meta["n_sigma_sensitivity]``."""
```
registerrier: same

gammapy/estimators/map/tests/test_excess.py
registerrier: Then you might check units

QRemy (approved):

QRemy:
All comments have been implement merging now.

PR #6142, opened 2025-09-17
Add further information about fitting to the Low Level Analysis tutorial

### LINKED ISSUE #5986: Spatial model not frozen in tutorial
**gammapy dev**
In the Low level API tutorial, the spatial model parameters are not frozen. The tutorial still works fine for the Crab, but breaks down when switching to a different source, such as PKS 2155-304.

Astro-Kirsty:
This resolves #5986.
Whilst we make this example as generic as possible, its true that if the user changes source the fit might not converge.
The idea here is to simply add further information to other parts of the documentation, as we have the information listed elsewhere.

### DIFF
--- a/examples/tutorials/starting/analysis_2.py
+++ b/examples/tutorials/starting/analysis_2.py
@@ -304,7 +304,7 @@ stacked.models = [sky_model, bkg_model]
 # overview tutorial or :ref:`this section <modifying-model-parameters>`
 # about modifying your parameters.
 #
-# The fit constructor takes a list of datasets as argument.
+# The fit constructor takes dictionaries that define global options for the optimizer.
 #
 
 fit = Fit(optimize_opts={"print_level": 1})
--- a/examples/tutorials/starting/analysis_2.py
+++ b/examples/tutorials/starting/analysis_2.py
@@ -297,7 +297,14 @@ stacked.models = [sky_model, bkg_model]
 # the ``stats`` method of the dataset to the minimizer. By default, it
 # uses ``iminuit``.
 #
-# Its constructor takes a list of dataset as argument.
+# It is possible that the fit will try to converge on an unrelated position
+# therefore you may choose to constrain the model parameter ranges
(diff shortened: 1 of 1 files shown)


examples/tutorials/starting/analysis_2.py
```diff
@@ -297,7 +297,14 @@
 # the ``stats`` method of the dataset to the minimizer. By default, it
 # uses ``iminuit``.
 #
-# Its constructor takes a list of dataset as argument.
+# It is possible that the fit will try to converge on an unrelated position
```
QRemy, line 300:
```suggestion
# It is possible that the fit will try to test positions far from your source of interest and fail to converge
```

QRemy (commented):
Thanks @Astro-Kirsty, maybe this would be more clear:

examples/tutorials/starting/analysis_2.py
```diff
@@ -297,7 +297,14 @@
...
 #
-# Its constructor takes a list of dataset as argument.
+# It is possible that the fit will try to test positions far from your source of interest and fail to converge
+# therefore you may choose to constrain the model parameter ranges
+# to stay within your expected region (see :ref:`here <dropdown-improve-fit>`).
+# For further information on fitting you can see the :doc:`/tutorials/details/fitting`
+# overview tutorial or :ref:`this section <modifying-model-parameters>`
+# about modifying your parameters.
+#
+# The fit constructor takes a list of datasets as argument.
```
registerrier: Well this is not true since v0.20!!! We should have changed that a while ago...

registerrier: write instead that it takes dictionaries defining global options for the optimizer

registerrier (changes requested):
Thanks @Astro-Kirsty . Looks good, but there is an old line that is no longer true since a long time

QRemy (approved):
Thanks @Astro-Kirsty no further comment.

PR #6147, opened 2025-09-19
Add Logparabola2Spectral model

QRemy:
Add a new spectral log parabola model defined such as the energy scale of the exponent and the reference energy can be different. This way the reference energy can be set to the pivot energy for each source without modifying the escale of the exponent, so the alpha and beta parameters are on the same scale for all sources. This is default model used in the 2HGPS.


gammapy/modeling/models/spectral.py
```diff
@@ -1952,6 +1952,76 @@ def e_peak(self):
...
+    ----------
+    amplitude : `~astropy.units.Quantity`
+        :math:`\phi_0`. Default is 1e-12 cm-2 s-1 TeV-1.
+    reference : `~astropy.units.Quantity`
+        :math:`E_0`. Default is 10 TeV.
+    alpha : `~astropy.units.Quantity`
+        :math:`\alpha`. Default is 2.
+    beta : `~astropy.units.Quantity`
+        :math:`\beta`. Default is 1.
+    reference : `~astropy.units.Quantity`
```
AtreyeeS: Both should not be called `reference`, no?

AtreyeeS (commented):
Thanks @QRemy !
Can you please adapt the doctoring to say `escale` instead of `reference`?

Do we need to maintain two different log parabola models? 
Can we just have this one and let the normal log parabola inherit from this to have `reference` and `escale` as tied parameters?

QRemy:
> Can we just have this one and let the normal log parabola inherit from this to have `reference` and `escale` as tied parameters?

No this would change the number of parameters and the shape of the covariance matrix creating a backward incompatibility.

AtreyeeS (approved):
> No this would change the number of parameters and the shape of the covariance matrix creating a backward incompatibility.

True! The tied parameters won't work. Then maybe can we deprecate the `LogParabola` in future?

examples/models/spectral/plot_logparabola2.py
```diff
@@ -0,0 +1,55 @@
+r"""
+.. _logparabola2-spectral-model:
+
+Log parabola spectral model
+===========================
+
+This model parametrises a log parabola spectrum.
```
AtreyeeS:
```suggestion

This model parametrises a log parabola spectrum, where the energy scale of the exponent and the reference energy can be different. See (add link to normal logpar) for the standard log parabola model
```

gammapy/modeling/models/spectral.py
```diff
@@ -1952,6 +1952,76 @@ def e_peak(self):
...
+    and the reference energy can be different.
+
+    For more information see :ref:`logparabola2-spectral-model`.
+
+    Parameters
+    ----------
+    amplitude : `~astropy.units.Quantity`
+        :math:`\phi_0`. Default is 1e-12 cm-2 s-1 TeV-1.
+    reference : `~astropy.units.Quantity`
+        :math:`E_0`. Default is 10 TeV.
```
AtreyeeS, line 1966: We usually have `reference` energy defaults at 1 TeV for all models. Maybe keep the same?

QRemy: it's 10 TeV for the Log-Parabola : https://docs.gammapy.org/dev/api/gammapy.modeling.models.LogParabolaSpectralModel.html

examples/models/spectral/plot_logparabola2.py
```diff
@@ -0,0 +1,55 @@
...
+        \phi(E) = \phi_0 \left( \frac{E}{E_0} \right) ^ {
+          - \alpha - \beta \log{ \left( \frac{E}{E_s} \right) }
+        }
+
+Note that :math:`log` refers to the natural logarithm. This is consistent
+with the `Fermi Science Tools
+<https://fermi.gsfc.nasa.gov/ssc/data/analysis/scitools/source_models.html>`_
+and `ctools
+<http://cta.irap.omp.eu/ctools-devel/users/user_manual/getting_started/models.html#log-parabola>`_.
+The `Sherpa <http://cxc.harvard.edu/sherpa/ahelp/logparabola.html>`_ package, however, uses :math:`log_{10}`. If you have
```
registerrier: Maybe this is not necessary as the implementation here is different from the logparabola from fermi tools?

examples/models/spectral/plot_logparabola2.py
```diff
@@ -0,0 +1,55 @@
...
+parametrization based on :math:`log_{10}` you can use the
+:func:`~gammapy.modeling.models.LogParabolaSpectralModel.from_log10` method.
```
QRemy:
```suggestion
Note that :math:`log` refers to the natural logarithm. 
If you have parametrization based on :math:`log_{10}` you can use the
:func:`~gammapy.modeling.models.LogParabola2SpectralModel.from_log10` method.
```

examples/models/spectral/plot_logparabola2.py
QRemy: removed

Issue #6121, opened 2025-09-01
Use SonarQube for quality assurance


morcuended:
**Is your feature request related to a problem? Please describe.**
CTAO computing projects use Sonarqube for code quality checks, defining some common rules and conditions that the software should fulfill. Gammapy as core part of CTAO SAT should also be checked by this system.

**Describe the solution you'd like**
Decide whether Gammapy should also integrate SonarQube quality checks and if so, integrate it.

**Describe alternatives you've considered**
Keep using codacy and get the quality reports from it.

**Additional context**
Some examples for this usage outside CTAO GitLab are ctapipe and simtools packages.

**Actionable steps**
- [ ] Make a first integration of SonarQube with Gammapy repository. With a workflow that runs daily and that can be triggered manually for the checks.
- [ ] Compare output with Codacy checks.
- [ ] Report back in this issue.

registerrier:
Thanks @morcuended . How would you see it? Should it:
- replace competely codecov and be run on every PR/push
- be run e.g. on a daily basis with a cron job

bkhelifi:
What are the differences between the SonarQube checks and criteria and the ones that we have? Are they complementary (ie new criteria) or similar? 

Should it be installed on the GitLab for the SAT or the Github of Gammapy?

morcuended:
> Thanks @morcuended . How would you see it? Should it:
>   replace competely codecov and be run on every PR/push
>   be run e.g. on a daily basis with a cron job

The idea is to help in the development process avoiding code smells, catching errors, etc., so indeed to have it in every PR would be ideal. But I do not want to overload the CI, right now. I would say we can start with a daily scheduled check, see what's the status of the code and compare it with codacy quality gates so we can decide which one to keep. For consistency with the rest of the project, I would stick with CTAO SonarQube quality gate and comply with those requirements (which I think already does).


> What are the differences between the SonarQube checks and criteria and the ones that we have? Are they complementary (ie new criteria) or similar?

I think the differences are minimal. The quality conditions, checks are probably the same.

> Should it be installed on the GitLab for the SAT or the Github of Gammapy?

It is already running for SAT in GitLab, but the question was whether to also trigger directly the checks on Gammapy itself.

PR #6150, opened 2025-09-19
Sonarqube integration

morcuended:
**Description**

Addresses #6121

It needs SONAR_TOKEN to be defined in the project.

**Dear reviewer**

This workflow is triggered by the completion of our CI workflow.
It then checks out the pull request repository / branch, runs the sonar scanner, downloads the coverage report and uploads the report to the sonarqube server. This is necessary as forks don't have access to secrets and SONAR_TOKEN is required to upload reports.

It might need some back and forth tries until we see the sonar report.

### DIFF
--- a/.github/workflows/sonar.yml
+++ b/.github/workflows/sonar.yml
@@ -1,0 +1,100 @@
+# This workflow is triggered by the completion of our CI workflow
+# It then checks out the pull request repository / branch, runs the
+# sonar scanner, downloads the coverage report and uploads the report
+# to the sonarqube server. This is necessary as forks don't have access
+# to secrets and SONAR_TOKEN is required to upload reports.
+#
+# Adapted from https://github.com/medplum/medplum/
+
+name: Sonar
+
+on:
+  workflow_run:
+    workflows: [CI]
+    types: [completed]
+
+jobs:
+  sonar:
+    name: Sonar
+    runs-on: ubuntu-latest
+    if: github.event.workflow_run.conclusion == 'success'
+    steps:
+      - name: 'Download code coverage'
+        uses: actions/github-script@v8
+        with:
+          script: |
+            let allArtifacts = await github.rest.actions.listWorkflowRunArtifacts({
+               owner: context.repo.owner,
+               repo: context.repo.repo,
+               run_id: context.payload.workflow_run.id,
+            });
+            let matchArtifact = allArtifacts.data.artifacts.filter((artifact) => {
+              return artifact.name == "gammapy-coverage-report"
+            })[0];
+            let download = await github.rest.actions.downloadArtifact({
+               owner: context.repo.owner,
+               repo: context.repo.repo,
+               artifact_id: matchArtifact.id,
+               archive_format: 'zip',
+            });
+            let fs = require('fs');
+            fs.writeFileSync(`${process.env.GITHUB_WORKSPACE}/gammapy-coverage-report.zip`, Buffer.from(download.data));
+
+      - name: 'Unzip code coverage'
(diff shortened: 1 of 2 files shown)


morcuended:
I did a test with my fork and looks like this: https://sonar-ctao.zeuthen.desy.de/dashboard?id=gammapy_gammapy_8c1bf7fc-018a-4afd-b216-015191d282fa&codeScope=overall

To make it fully integrated (adding the coverage report), the new step in the CI workflow uploading the info for sonarqube needs to run first from main. I think we can give it a try in the main repository.

morcuended:
To do:
 - [x] Check what files to exclude from the SQ analysis (tests, documentation, examples, etc.?). In a separate PR.
 - [x] Coverage report is ingested but paths inside the report are not fetched by SQ

.github/workflows/sonar.yml
```diff
@@ -0,0 +1,104 @@
...
+# to secrets and SONAR_TOKEN is required to upload reports.
+#
+# Adapted from https://github.com/medplum/medplum/
+
+name: SonarQube quality gate
+
+on:
+  workflow_run:
+    workflows: [CI]
+    types: [completed]
```
natthan-pigoux, line 14: I guess this workflow will never be triggered if one the `CI` jobs is skipped (e.g. the doc is skipped). Is that wanted?

Since you actually only depend on the `coverage` step, we would need to add a specific condition on the coverage job. What do you think?

Astro-Kirsty:
We should also add a fragment file here

.github/workflows/sonar.yml
morcuended: In the current configuration, the sonarqube step ran with the `skip-docs` label. But indeed we should make sure to require coverage step

morcuended:
There is an error in SonarQube when tries to check coverage analysis:

> The following error(s) occurred while trying to import coverage report:
> Cannot resolve the file path '/home/runner/work/gammapy/gammapy/.tox/py310-test-alldeps-cov/lib/python3.10/site-packages/gammapy/__init__.py' of the coverage report, the file does not exist in all 'source'.
> Cannot resolve 262 file paths, ignoring coverage measures for those files

I try to map the tox path relative to gammapy src code with the last commit so that these kinds of paths:
```
/home/runner/work/gammapy/gammapy/.tox/py310-test-alldeps-cov/lib/python3.10/site-packages/gammapy/__init__.py
```
are mapped to:
```
gammapy/__init__.py
```
where SonarQube is looking for the files.

morcuended:
Changelog check it failing despite having included the fragment. There is a problem with the label parsing syntax, that I tried to fix in #6224

morcuended:
@natthan-pigoux @registerrier I would like to test it in main repository. The thing still missing is the coverage report analysis by SonarQube, I still find problems with the paths of files that SonarQube is trying to find in the source code. But it is not critical and can be fixed later. From my side, this is good to go.

setup.cfg
```diff
@@ -139,6 +139,25 @@ omit =
     */gammapy/extern/*
      */gammapy/version*
 
+# Make filenames relative to the repository root
```
registerrier, line 142: So we merge the pyproject.toml first and adapt after?

registerrier (approved):
Thanks @morcuended . This looks good to me. 

Should it go first and we adapt #6176. Or the opposite? Opinions @morcuended @natthan-pigoux ?

setup.cfg
morcuended: Either way is fine. This part will go to `.coveragerc`

registerrier:
In the dev call, we decided to merge this one first. 
Merging now.

PR #6155, opened 2025-09-19
Fix blank plot in astro dark matter tutorial

Astro-Kirsty:
As per title


AtreyeeS (commented):
There are two blank plots in this tutorial. I think you need to remove the comment line around 154 as well

morcuended (approved):
Thx @Astro-Kirsty, I check the rendered version and it looks good.

PR #6156, opened 2025-09-19
Remove extra mask plotting

### LINKED ISSUE #6117: Mask safe plotted twice in residual panel of count distribution plot
**Gammapy version**
Gammapy 2.0 (see docs).

**Bug description**
It seems that the mask is plotted twice.

**Expected behavior**
Only be plotted once.

**To Reproduce**
For example, see plot in https://docs.gammapy.org/2.0/tutorials/analysis-1d/extended_source_spectral_analysis.html#explore-the-fit-results

Astro-Kirsty:
Fixes #6117

### DIFF
--- a/gammapy/datasets/spectrum.py
+++ b/gammapy/datasets/spectrum.py
@@ -81,7 +81,6 @@ class PlotMixin:
         ax_residuals.set_ylabel(f"Residuals\n{label}")
         plt.setp(ax_spectrum.get_xticklabels(), visible=bool_visible_xticklabel)
         self.plot_masks(ax=ax_spectrum)
-        self.plot_masks(ax=ax_residuals)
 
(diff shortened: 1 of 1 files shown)


gammapy/datasets/spectrum.py
```diff
@@ -81,7 +81,6 @@ def plot_fit(
         ax_residuals.set_ylabel(f"Residuals\n{label}")
         plt.setp(ax_spectrum.get_xticklabels(), visible=bool_visible_xticklabel)
         self.plot_masks(ax=ax_spectrum)
-        self.plot_masks(ax=ax_residuals)
```
AtreyeeS, line 84: Ok. `plot_residuals_spectral` already includes the mask

PR #6158, opened 2025-09-22
Add further test for sensitivity

Astro-Kirsty:
Per the title

I decided to backport as we did not test this previously. Not sure if you agree it is correct?


registerrier (approved):

PR #6159, opened 2025-09-22
Fix typo in edisp plotting

Astro-Kirsty:
@LeanderSchlegel  noticed that in the docs [here](https://docs.gammapy.org/2.0/user-guide/irf/edisp.html#as-a-function-of-true-energy-gadf-ogip-rmf) the x axis label was not correctly showing the units. 

This is a simple fix where we were accidentally calling the yaxis, instead of xaxis


morcuended (approved):
Well spotted

AtreyeeS (approved):
Thanks @Astro-Kirsty ! Well spotted

PR #6170, opened 2025-09-26
Introduce ObservationTableReader and Converter between GADF and internal format

LeanderSchlegel:
Dear all,

on behalf of @registerrier, @bkhelifi and me, this PR introduces a new `class ObservationTableReader` based on the `class EventListReader`. It is intended to prepare the separation of the internal data format of the observation table from the gadf format discussed in #4238 and is based on the prototype in PR #6128.

In this first step, only the two files `io.py` and `test_io.py` are modified, by adding a new class `ObservationTableReader` that contains a `read`-function to read observation tables from disk and a `from_gadf_table`-converter function, that converts the table from the gadf-format into an internal format. The internal format is only requiring the column "OBS_ID" to be present and defines otherwise only types and units for the entries "RA_PNT", "DEC_PNT", "ALT_PNT", "AZ_PNT", "TSTART", "TSTOP", while the latter are astropy.time objects. The converter ensures the correct types, units and conversion of times into time-objects. In the future, further converters for more file formats can directly be added.

For now, the new reader and converter is not used in Gammapy and only the test_io.py calls it to test its functionalities.
 In the moment, the test of the converter fails, as it relies already on a new ObservationTable constructor, that will be introduced in a separate PR. Two tests for the reader are still missing, testing the case of assuming the gadf format if it can not be inferred from disk and a test if the version is unknown.

The codedesign is strongly based on the design of the class EventList and I took code from there (files: event_list.py and io.py) and adapted it for some functionalities.

**Dear reviewer**
Any review is very welcome! 
Two questions I had are: Can I optimize the code of the converter and shall we take out the many try-excepts for catching different exceptions, maybe only: "Format unknown", "Type conversion failed", "Unit conversion failed", "Time object could not be created"?
Is the `version`-parameter in the reader for now irrelevant and shall we take it out completely?

Thank you very much,
Leander, Régis, Bruno

### DIFF
--- a/.github/workflows/cffconvert.yml
+++ b/.github/workflows/cffconvert.yml
@@ -39,7 +39,7 @@ jobs:
           eossr-metadata-validator codemeta.json 
       
       - name: commit changes
-        uses: stefanzweifel/git-auto-commit-action@v6.0.1
+        uses: stefanzweifel/git-auto-commit-action@v7.0.0
         if: success()
         with:
           commit_author: GitHub Actions <actions@github.com>
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -1,0 +1,40 @@
+name: Check changelog fragments
+
+on: [pull_request]
+
+jobs:
+  check-changelog:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v5
+
+      - name: Verify changelog label and fragment
+        run: |
+          set -euo pipefail
+
+          # Obtain all labels set to the PR
+          labels="${{ toJson(github.event.pull_request.labels.*.name) }}"
+
+          # If the labels contains feature, bug or api-change then automatically require the 'changelog-required' label
+          if echo "$labels" | grep -Eq '"(feature|bug|api-change)"'; then
+            if ! echo "$labels" | grep -q '"changelog-required"'; then
+              echo "PR has 'feature', 'bug', or 'api-change' label but is missing 'changelog-required'."
+              exit 1
+            fi
+          fi
+
+          # Now we look for 'changelog-required' label
+          if echo "$labels" | grep -q "changelog-required"; then
+            echo "Label 'changelog-required' is set. Checking for fragments in docs/release-notes/*.rst..."
+
+            if ! git diff --name-only origin/${{ github.base_ref }}... | grep -q '^docs/release-notes/.*\.rst$'; then
+              echo "No changelog fragment found in docs/release-notes/"
+              exit 1
+            fi
+
+            echo "Changelog fragment found."
+          else
+            echo "No 'changelog-required' label, skipping this check."
+          fi
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,7 @@ jobs:
       - name: Check out repository
         uses: actions/checkout@v5
       - name: Set up Python
-        uses: actions/setup-python@v5
+        uses: actions/setup-python@v6
         with:
           python-version: '3.11'
       - name: Install pre-commit
@@ -90,7 +90,7 @@ jobs:
         with:
           fetch-depth: 0
       - name: Set up Python ${{ matrix.python }}
-        uses: actions/setup-python@v5
+        uses: actions/setup-python@v6
         with:
           python-version: ${{ matrix.python }}
       - name: Install base dependencies
@@ -141,7 +141,7 @@ jobs:
         with:
           fetch-depth: 0
       - name: Set up Python 3.11
-        uses: actions/setup-python@v5
+        uses: actions/setup-python@v6
         with:
           python-version: '3.11'
       - name: Install base dependencies
@@ -157,7 +157,7 @@ jobs:
         run: |
           tox -e build_docs
       - name: Upload HTML output
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@v5
         with:
           name: gammapy-doc-html
           path: docs/_build/html/
--- a/.github/workflows/dispatch_benchmark.yml
+++ b/.github/workflows/dispatch_benchmark.yml
@@ -9,7 +9,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Dispatch Gammapy Benchmarks
-        uses: peter-evans/repository-dispatch@v3
+        uses: peter-evans/repository-dispatch@v4
         with:
           token: ${{ secrets.REMOTE_DISPATCH }}
           repository: gammapy/gammapy-benchmarks
--- a/.github/workflows/doctests.yml
+++ b/.github/workflows/doctests.yml
@@ -16,7 +16,7 @@ jobs:
         uses: actions/checkout@v5
 
       - name: Set up Python
-        uses: actions/setup-python@v5
+        uses: actions/setup-python@v6
         with:
           python-version: '3.11'
 
--- a/.github/workflows/linkcheck.yml
+++ b/.github/workflows/linkcheck.yml
@@ -21,7 +21,7 @@ jobs:
         with:
           fetch-depth: 0
       - name: Set up Python 3.11
-        uses: actions/setup-python@v5
+        uses: actions/setup-python@v6
         with:
           python-version: '3.11'
       - name: Install base dependencies
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -36,7 +36,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Dispatch Gammapy Docs
-        uses: peter-evans/repository-dispatch@v3
+        uses: peter-evans/repository-dispatch@v4
         with:
           token: ${{ secrets.REMOTE_DISPATCH }}
           repository: gammapy/gammapy-docs
@@ -49,7 +49,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Dispatch Gammapy Webpage
-        uses: peter-evans/repository-dispatch@v3
+        uses: peter-evans/repository-dispatch@v4
         with:
           token: ${{ secrets.REMOTE_DISPATCH }}
           repository: gammapy/gammapy-webpage
--- a/.github/workflows/tutorialjupytexttests.yml
+++ b/.github/workflows/tutorialjupytexttests.yml
@@ -19,7 +19,7 @@ jobs:
         with:
           fetch-depth: 0
 
-      - uses: actions/setup-python@v5
+      - uses: actions/setup-python@v6
         with:
           python-version: '3.13' 
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,7 +1,25 @@
-
 repos:
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v6.0.0
+    hooks:
+      - id: trailing-whitespace
+      - id: end-of-file-fixer
+      - id: check-yaml
+        args: ["--unsafe"]
+      - id: check-added-large-files
+      - id: check-case-conflict
+      - id: check-merge-conflict
+
+  - repo: https://github.com/codespell-project/codespell
+    rev: v2.4.1
+    hooks:
+    - id: codespell
+      args: ["--ignore-words", "dev/codespell/ignore-words.txt", "--skip", "./.git"]
+      additional_dependencies:
+        - tomli
+
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.7.3
+    rev: v0.14.1
     hooks:
       - id: ruff
         name: ruff linting
--- a/dev/authors.py
+++ b/dev/authors.py
@@ -150,7 +150,7 @@ def check_author_lists(since_last_lts):
 
     print(message + "\n  ".join(authors_annotated) + "\n")
 
-    print("(CC) = Coordination Committe, (AA) = Additional Authors, (OO) = Opted out\n")
+    print("(CC) = Coordination Committee, (AA) = Additional Authors, (OO) = Opted out\n")
 
 
 if __name__ == "__main__":
--- a/dev/codespell/ignore-words.txt
+++ b/dev/codespell/ignore-words.txt
@@ -1,8 +1,16 @@
 dne
+egal
 esy
 hist
+indx
 livetime
+livetimes
 nd
+ned
 merrors
+NotIn
+OptIn
+slac
 unparseable
 compiletime
+THIRDPARTY
--- a/dev/github_summary.py
+++ b/dev/github_summary.py
@@ -70,7 +70,7 @@ class GitHubInfoExtractor:
         return result
 
     def _extract_pull_request_info(self, pull_request):
-        """Builds a dictionary containing a list of summary informations.
+        """Builds a dictionary containing a list of summary information.
 
         Parameters
         ----------
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -102,6 +102,7 @@ extensions = [
     'sphinx_copybutton',
     "sphinx_design",
     "sphinx_gallery.gen_gallery",
+    "sphinx_changelog",
     "sphinx.ext.doctest",
     "sphinx.ext.mathjax",
     'sphinx.ext.viewcode',
--- a/docs/development/dev_howto.rst
+++ b/docs/development/dev_howto.rst
@@ -6,6 +6,8 @@
 Developer How To
 ****************
 
+.. _general-conventions:
+
 General conventions
 -------------------
 
@@ -516,13 +518,15 @@ repository. It is recommended that developers have `$GAMMAPY_DATA` environment v
 where they have fetched the `gammapy-data <https://github.com/gammapy/gammapy-data>`__ GitHub repository,
 so they can push and pull eventual modification of its content.
 
+.. _skip-actions-pr:
+
 Making a pull request which skips GitHub Actions
 ++++++++++++++++++++++++++++++++++++++++++++++++
 
 For minor PRs (eg: correcting typos in doc-strings) we can skip GitHub Actions.
 Adding ``[ci skip]`` in a specific commit message will skip CI for that specific commit which can be useful for draft or incomplete PR.
 For details, `see here. <https://github.blog/changelog/2021-02-08-github-actions-skip-pull-request-and-push-workflows-with-skip-ci/>`__
-See also section: Skip GitHub Actions on local fork, below.
+See also the section on skipping GitHub Actions on a local fork, :ref:`here <skip-actions-local-fork>`.
 
 Fix non-Unix line endings
 +++++++++++++++++++++++++
@@ -557,6 +561,7 @@ one of which involves the `meeseeksmachine <https://github.com/meeseeksmachine>`
 
 
 
+.. _release-notes:
 
 Release notes
 +++++++++++++
@@ -569,12 +574,37 @@ As explained in the
 of the Astropy docs, there are (at least) two approaches for adding to the releases,
 each with pros and cons.
 
-We've had some pain due to merge conflicts in the releases notes and having to wait
-until the contributor rebases (and having to explain git rebase to new contributors).
+We utilise `towncrier <https://towncrier.readthedocs.io/en/stable/>`__ for our release notes.
 
-So our recommendation is that releases entries are not added in pull requests,
-but that the core developer adds a releases notes entry after right after having
-merged a pull request (you can add ``[skip ci]`` on this commit).
+- For each PR, a related 'fragment' file should be created in the ``docs/release-notes`` folder.
+- The naming convention of the file should be ``<PULL REQUEST NUMBER>.<TYPE>.rst``, where the available
+  types are ``infrastructure``, ``docs``, ``feature`` or ``bugfix``.
+- The file should contain a suitable message for the PR, for example "A new function `select_nested_models`
+  has been introduced to perform nested model fits and compute the resulting test statistic (TS) between
+  two nested hypotheses.".
+- Note: not all PRs need to be included in the changelog. For example, fixing a typo in the documentation
+  does not need a changelog fragment entry. If you are unsure if you need a fragment, the lead developers can
+  help you decide.
+
+
+
+
+How to review a pull request
+----------------------------
+
+The checklist below outlines key steps to follow when reviewing a PR.
+
+1. Ensure that a milestone and labels are correctly set for the PR.
+2. Check that the PR links to associated issue(s), if any.
+3. Check that the code does what it is supposed to do.
+4. Check that the CI workflow passes without failure.
+5. Check that the general conventions are fulfilled, :ref:`see here <general-conventions>`.
+6. Check that the code can be understood by reading it.
+7. Confirm all docstrings are correct and formatted correctly, :ref:`see here <docstring-formatting>`.
+8. Check that all commits are signed.
+9. Check that a fragment was added, if necessary, for :ref:`towncrier <release-notes>`.
+10. Where relevant, add examples for the users, :ref:`see here <docstring-code-py-file>`.
+11. When needed, check that the documentation build looks correct by checking the artifacts of the CI, :ref:`see here <access-doc-on-pr>`.
 
 How to handle API breaking changes?
 -----------------------------------
@@ -947,9 +977,11 @@ script can be used as follows::
 If the path of the output file is not provided, the script will be written in the same folder as the notebook and with
 the same name.
 
+.. _skip-actions-local-fork:
+
 Skip GitHub Actions on local fork
 +++++++++++++++++++++++++++++++++
 If not explicitly needed, it can be convenient to skip the GitHub Actions for pushes onto local forks and only perform them on the upstream, once a pull request is opened.
 This way computation power is saved and automatic commits by the Actions are avoided in the forks commit-history. An easy way to achieve this, is to deactivate the
 GitHub Actions completely for the fork, following the GitHub documentation, `see here. <https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#managing-github-actions-permissions-for-your-repository>`__
-See also section: Making a pull request which skips GitHub Actions, above.
+See also the section on making a pull request which skips GitHub Actions, :ref:`here <skip-actions-pr>`.
--- a/docs/development/doc_howto.rst
+++ b/docs/development/doc_howto.rst
@@ -36,6 +36,7 @@ This should work::
 
 You need a bunch or LaTeX stuff, specifically ``texlive-fonts-extra`` is needed.
 
+.. _access-doc-on-pr:
 
 Accessing documentation build on a Pull Request
 +++++++++++++++++++++++++++++++++++++++++++++++
@@ -94,6 +95,8 @@ In order to perform tests of these snippets of code present in RST files, you ma
 .. code-block:: bash
 
     pytest --doctest-glob="*.rst" docs/
+
+.. _docstring-code-py-file:
 
 Code in docstrings in Python files
 ++++++++++++++++++++++++++++++++++
@@ -186,6 +189,7 @@ Another option is to create a general list of references, as follows:
         * `Author et al. (2023), "Title" <link_to_nasaads>`_
         * `Author2 et al. (2022), "Title2" <link_to_nasaads>`_
 
+.. _docstring-formatting:
 
 Docstring formatting
 ^^^^^^^^^^^^^^^^^^^^
--- a/docs/development/pigs/pig-009.rst
+++ b/docs/development/pigs/pig-009.rst
@@ -201,7 +201,7 @@ So far Gammapy only supports binned likelihood analysis and technically most
 use- cases for the event sampling could be solved with binned simulations. A
 binned simulation can be basically achieved by a call to
 ``numpy.random.poisson()`` based on the predicted number of counts map. This is
-conceptionally simpler as well as computationally more efficient than a sampling
+conceptually simpler as well as computationally more efficient than a sampling
 of event lists. In ``Gammapy`` a similar dataset simulation is already
 implemented in ``Dataset.fake()``, although this has a limited number of use
 cases than an event sampler. However, to support the full data access and data
--- a/docs/development/pigs/pig-017.rst
+++ b/docs/development/pigs/pig-017.rst
@@ -82,7 +82,7 @@ for the standard logging, but with a different severity level. The choice of the
 Data Model allows for interoperability with provenance information captured from *ctapipe* as well
 as with other astronomical archive services or software providing provenance information in the Virtual
 Observatory framework and with the W3C provenance models. In this optic, the development will be done on
-top of `ctapipe.core.provenance`_, re-using its present capabilities when possible and extending it. The
+top of `ctapipe.core.provenance`_, reusing its present capabilities when possible and extending it. The
 information captured will relate to the execution environment, the overall high-level interface configuration,
 the workflow execution process (i.e the relationships among datasets and processes and the parameter values
 used in each process step), and the serialisations of the intermediate and final datasets or objects produced.
--- a/docs/development/pigs/pig-020.rst
+++ b/docs/development/pigs/pig-020.rst
@@ -36,7 +36,7 @@ works as long as all objects reside in the same memory space.
 
 If datasets are distributed to different processes in future, it is technically
 complex and probably in-efficient to share model states between all sub-processes.
-It is conceptionally simpler if processes communicate with a single server process
+It is conceptually simpler if processes communicate with a single server process
 that contains the single global model object.
 
 The fundamental important difference to the current design is, that the model objects
--- a/docs/development/pigs/pig-024.rst
+++ b/docs/development/pigs/pig-024.rst
@@ -100,7 +100,7 @@ Releases
 
 Each Gammapy release should be associated with an updated list of authors that will be public on repository (Zenodo and
 HAL/SWH). This section is about feature releases. Bug releases will have as list of authors the one of its associated
-feature release whose rules are described here, plus the potentiel additional contributors.
+feature release whose rules are described here, plus the potential additional contributors.
 
 The list of authors is composed of people who contributed to the current release code
 by accepting the DCO. The history of merged pull requests will be used as starting point.
--- a/docs/development/pigs/pig-028.rst
+++ b/docs/development/pigs/pig-028.rst
@@ -37,7 +37,7 @@ Event type handling
 +++++++++++++++++++
 
 Starting from a datastore containing events lists and IRFs with types and classes of events,
-the user produces of a list of datasets per event type or class. Meta informations of
+the user produces of a list of datasets per event type or class. Meta information of
 the datasets allow for complex types handling at modeling/fitting step (e.g. joint fit
 of A & B type, stack all type together etc).
 
@@ -154,7 +154,7 @@ We discuss here some aspects that should be explored.
   ``MapAxes`` and an interpolator. In addition, ``Maps`` use the ``Geom`` object to
   represent the spherical coordinates.
 
-  - Having a common data structure could help make maps fully re-usable for IRFs.
+  - Having a common data structure could help make maps fully reusable for IRFs.
     This might be a common use case with pyirf.
   - One could allow ``Maps`` and ``MapCoord`` objects without spatial axes.
     Introducing specialized spatial axes such as `WcsMapAxis`, `RegionMapAxis` or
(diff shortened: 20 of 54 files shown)


gammapy/data/tests/test_io.py
```diff
@@ -50,3 +56,117 @@ def test_eventlist_reader_empty_gadf_table():
...
+    hess_obs_table = "$GAMMAPY_DATA/hess-dl3-dr1/obs-index.fits.gz"
+
+    with pytest.raises(AstropyDeprecationWarning):
+        obs_table = ObservationTableReader().read(
+            hess_obs_table, hdu="unknown_extension"
+        )
+        assert len(obs_table) == 105
+
+
+def test_observationtable_reader_unknown_format(tmpdir):
```
registerrier:
```suggestion
@requires_data()
def test_observationtable_reader_unknown_format(tmpdir):
```

gammapy/data/io.py
```diff
@@ -141,3 +143,137 @@ def to_hdu(self, event_list, format="gadf"):
...
+
+    @staticmethod
+    def from_gadf_table(table_gadf):
+        """Convert gadf observation table into internal data format.
+
+        Parameters
+        ----------
+        table_gadf : `~astropy.Table.table`
+            Table in gadf 0.2/0.3 format.
+            https://gamma-astro-data-formats.readthedocs.io/en/v0.2/data_storage/obs_index/index.html
```
registerrier: I think this is not the correct way to introduce a link to an external page.

gammapy/data/io.py
```diff
@@ -141,3 +143,137 @@ def to_hdu(self, event_list, format="gadf"):
...
+        ObservationTable : `~gammapy.data.ObservationTable`
+            ObservationTable in internal data format.
+
+            Mandatory Columns: OBS_ID
+
+            Defined Units and Types for Columns as follows:
+
+            OBS_ID	RA_PNT	DEC_PNT	ALT_PNT	AZ_PNT	TSTART	TSTOP
+                        deg	    deg	    deg	   deg
+            int64   float64	float64	float64 float64	 Time	 Time
```
registerrier: This is not the place to present the format of `ObservationTable`

gammapy/data/io.py
```diff
@@ -141,3 +143,137 @@ def to_hdu(self, event_list, format="gadf"):
...
+        if format == "GADF" or format == "OGIP":
+            if version == "0.2" or version == "0.3":
+                return self.from_gadf_table(table_disk)
+            else:
+                raise ValueError(f"Unknown {format} version :{version}")
+        else:
+            raise ValueError(f"Unknown fileformat :{format}")
+
+    @staticmethod
+    def from_gadf_table(table_gadf):
```
registerrier: you could make this a private function as this is an implementation detail

gammapy/data/io.py
```diff
@@ -141,3 +143,137 @@ def to_hdu(self, event_list, format="gadf"):
...
+        """Convert gadf observation table into internal data format.
```
registerrier:
```suggestion
        """Convert gadf observation table into`~gammapy.data.ObservationTable`.
```

gammapy/data/io.py
```diff
@@ -141,3 +143,137 @@ def to_hdu(self, event_list, format="gadf"):
             raise ValueError(f"Only the 'gadf' format supported, got {format}")
 
         return self._to_gadf_table_hdu(event_list)
+
+
+class ObservationTableReader:
+    """Reader class for ObservationTable"""
+
+    def read(self, filename, hdu=None):
+        """Read ObservationTable from file.
```
registerrier:
```suggestion
        """Read `~gammapy.data.ObservationTable` from file.
```

gammapy/data/io.py
```diff
@@ -141,3 +143,137 @@ def to_hdu(self, event_list, format="gadf"):
...
+
+        Returns
+        -------
+        ObservationTable : `~gammapy.data.ObservationTable`
+            ObservationTable from disk converted into internal data format.
+        """
+        filename = make_path(filename)
+
+        table_disk = Table.read(filename, format="fits", hdu=hdu)
+        table_disk_meta = table_disk.meta
```
registerrier: is it needed?

gammapy/data/io.py
```diff
@@ -141,3 +143,137 @@ def to_hdu(self, event_list, format="gadf"):
...
+
+        format = table_disk_meta.get("HDUCLASS", "unknown")
+        version = table_disk_meta.get("HDUVERS", "unknown")
```
registerrier:
```suggestion
        format = table_disk_meta.get("HDUCLASS")
        version = table_disk_meta.get("HDUVERS")
```
Using `None` seems simpler

gammapy/data/io.py
```diff
@@ -141,3 +143,137 @@ def to_hdu(self, event_list, format="gadf"):
...
+
+        if (format == "unknown") or (format is None):
+            format = "GADF"
+            version = "0.3"
+            warnings.warn(
```
registerrier: You might log a warning

gammapy/data/io.py
```diff
@@ -141,3 +143,137 @@ def to_hdu(self, event_list, format="gadf"):
...
+                UserWarning(
+                    f"Could not infer fileformat from metadata in {filename}, assuming GADF."
+                )
+            )
+
+        if format == "GADF" or format == "OGIP":
```
registerrier:
```suggestion
        if format.upper() in ["GADF", "OGIP"]:
```

gammapy/data/io.py
```diff
@@ -141,3 +143,137 @@ def to_hdu(self, event_list, format="gadf"):
...
+            if version == "0.2" or version == "0.3":
+                return self.from_gadf_table(table_disk)
+            else:
+                raise ValueError(f"Unknown {format} version :{version}")
```
registerrier: As there is no difference between versions here, I don't think you should complexify the code.

gammapy/data/io.py
```diff
@@ -141,3 +143,137 @@ def to_hdu(self, event_list, format="gadf"):
...
+        removed_names = []
+
+        try:
+            obs_id = table_gadf["OBS_ID"].astype("int")
+        except ValueError:
+            raise RuntimeError(
+                "Could not convert OBS_ID to int. Can not create table without OBS_ID."
+            )
+        new_table = Table({"OBS_ID": obs_id}, meta=meta_gadf)
+        new_table = table.unique(new_table, keys="OBS_ID")
```
registerrier: Is it required here?

gammapy/data/io.py
```diff
@@ -141,3 +143,137 @@ def to_hdu(self, event_list, format="gadf"):
...
+        removed_names.append("OBS_ID")
+
+        for colname in ["RA_PNT", "DEC_PNT", "ALT_PNT", "AZ_PNT"]:
+            if colname in names_gadf:
+                try:
+                    new_table[colname] = Quantity(
+                        table_gadf[colname].astype("float64"),
+                        u.Unit(table_gadf[colname].unit),
+                    ).to(u.deg)
```
registerrier:
```suggestion
                    new_table[colname] = table_gadf[colname].quantity.to("deg")
```
you don't really need to adapt the type

gammapy/data/io.py
```diff
@@ -141,3 +143,137 @@ def to_hdu(self, event_list, format="gadf"):
...
+
+        removed_names = []
+
+        try:
+            obs_id = table_gadf["OBS_ID"].astype("int")
+        except ValueError:
+            raise RuntimeError(
+                "Could not convert OBS_ID to int. Can not create table without OBS_ID."
+            )
+        new_table = Table({"OBS_ID": obs_id}, meta=meta_gadf)
```
registerrier: Do you need to create a new table? You could simply replace the time columns by the correct ones.

gammapy/data/io.py
```diff
@@ -141,3 +143,137 @@ def to_hdu(self, event_list, format="gadf"):
...
+        new_table = table.unique(new_table, keys="OBS_ID")
+        removed_names.append("OBS_ID")
+
+        for colname in ["RA_PNT", "DEC_PNT", "ALT_PNT", "AZ_PNT"]:
```
registerrier: This part could be left to the validation in the `ObservationTable` constructor. It does not bring much to do it here.

gammapy/data/io.py
```diff
@@ -141,3 +143,137 @@ def to_hdu(self, event_list, format="gadf"):
...
+                        u.Unit(table_gadf[colname].unit),
+                    ).to(u.deg)
+                except TypeError:
+                    warnings.warn(f"Could not convert unit for column {colname}.")
+                removed_names.append(colname)
+
+        for colname in ["TSTART", "TSTOP"]:
+            if colname in names_gadf:
+                try:
+                    time_ref = time_ref_from_dict(meta_gadf)
```
registerrier: this should be outside the for loop. It is the same for both columns.

gammapy/data/io.py
```diff
@@ -141,3 +143,137 @@ def to_hdu(self, event_list, format="gadf"):
...
+                try:
+                    new_table[colname] = Quantity(
+                        table_gadf[colname].astype("float64"),
+                        u.Unit(table_gadf[colname].unit),
+                    ).to(u.deg)
+                except TypeError:
+                    warnings.warn(f"Could not convert unit for column {colname}.")
+                removed_names.append(colname)
+
+        for colname in ["TSTART", "TSTOP"]:
```
registerrier: to avoid the additional tests you could try:
`for colname in set(["TSTART", TSTOP]).intersect(set(names_gadf)):`

for clarity you could do :
```
time_columns = set(["TSTART", TSTOP]).intersect(set(names_gadf))
for colname in time_columns:
```

registerrier (commented):
Thanks @LeanderSchlegel . This is nice. This is a partial review, there are quite a few comments. This is normal :)

gammapy/data/io.py
LeanderSchlegel: Thank you very much for the optimization, I applied your suggestion in cebd507.

LeanderSchlegel:
Thank you very much for your review and for all your helpful suggestions, @registerrier! I applied your suggested commits already and am going through your suggested changes :)

gammapy/data/io.py
LeanderSchlegel: Thank you, I fully agree, I took the version parameter out in commit fd8ff17.

gammapy/data/io.py
LeanderSchlegel: I have to admit, I am not sure with the make_path, I took that from the EventListReader read function, could it be helpful or can it be removed?
Is there a better way to infer the format, not requiring opening the table, that I could use?
Would it be a good idea to pass the `table_disk_meta` to the converter or would it be better to take this line out?

gammapy/data/io.py
LeanderSchlegel: Thank you, I changed it to a log-warning according to the developer guide, in commit e14bb56. I took the UserWarning out, was that correct?

gammapy/data/io.py
LeanderSchlegel: Thank you, I fully agree, I underscored the function name here and in the tests, with commit dd31fdc.

gammapy/data/io.py
LeanderSchlegel: Thank you, I completely agree! I took the part out with commit 0fa986b and would place it in the docstring for the observation table class.

gammapy/data/io.py
LeanderSchlegel: Thank you very much for the suggestion, I agree and removed the new_table and just replace the time columns if present by the new ones in commit 3e08dbe.

gammapy/data/io.py
LeanderSchlegel: Thank you very much, I took it out in commit 3e08dbe, to leave it to the validator to check the units. Could it be a problem, that the unit conversion is not done anymore?

gammapy/data/io.py
LeanderSchlegel: Thank you, I agree it is not necessary here and took it out with commit 3e08dbe when removing new_table.

gammapy/data/io.py
LeanderSchlegel: Thank you, I agree, I took it out the for loop with commit 3e08dbe.

gammapy/data/tests/test_io.py
```diff
@@ -50,3 +54,105 @@ def test_eventlist_reader_empty_gadf_table():
...
+        {"OBS_ID": ["1"], "TSTART": [100]},
+        meta={
+            "MJDREFI": 50000,
+            "MJDREFF": 100.0,
+            "TIMEUNIT": "s",
+        },
+    )
+    obs_table = ObservationTableReader._from_gadf_table(t)
+    assert obs_table.keys() == ["OBS_ID", "TSTART"]
+    assert isinstance(obs_table["TSTART"], Time) == True
```
QRemy:
```suggestion
    assert isinstance(obs_table["TSTART"], Time)
```

gammapy/data/tests/test_io.py
```diff
@@ -50,3 +54,105 @@ def test_eventlist_reader_empty_gadf_table():
...
+        {"OBS_ID": ["1"], "TSTART": [100]},
+        meta={
+            "MJDREFI": 50000,
+            "MJDREFF": 100.0,
+            "TIMEUNIT": "s",
+        },
+    )
+    obs_table = ObservationTableReader._from_gadf_table(t)
+    assert obs_table.keys() == ["OBS_ID", "TSTART"]
+    assert isinstance(obs_table["TSTART"], Time) is True
```
QRemy:
```suggestion
    assert isinstance(obs_table["TSTART"], Time)
```

gammapy/data/tests/test_io.py
```diff
@@ -50,3 +54,105 @@ def test_eventlist_reader_empty_gadf_table():
...
+
+    # If TSTART or TSTOP in table and header keywords present
+    # but conversion fails for any reason (wrong time unit, wrong types)
+    # warning is raised and time-columns are dropped.
```
QRemy: Why not an error instead ?

gammapy/data/io.py
```diff
@@ -141,3 +145,100 @@ def to_hdu(self, event_list, format="gadf"):
...
+                        warnings.warn(f"Invalid unit for column {colname}.")
+                        table_gadf.remove_column(colname)
+                    except TypeError:
+                        warnings.warn(f"Could not convert type for column {colname}.")
+                        table_gadf.remove_column(colname)
+            except KeyError:
+                warnings.warn(
+                    "Found column TSTART or TSTOP in gadf table, but can not create columns in internal format (MixinColumn Time) due to missing header keywords in file."
+                )
+                table_gadf.remove_columns(time_columns)
```
QRemy, line 232: Maybe this could be moved in a separate in helper function if the time conversion validation is needed elsewhere.
And same comment as above : why not raising an error instead of the warning ?

gammapy/data/tests/test_io.py
LeanderSchlegel: An error would definitely also be an option, we thought dropping the time columns and raising a warning would be less strict, so that the other columns could still be loaded.

gammapy/data/io.py
LeanderSchlegel: Thank you very much for the suggestion, could it be an idea to place the validation of the time conversion into gammapy.utils.time time_ref_from_dict, to capture the different possibilities it could fail?

LeanderSchlegel:
Thank you very much for your comments @QRemy! I'm very sorry for not writing, that I was still changing something and committed a larger change to the time conversion and the tests, to make sure, the tests actually cover the different warning-cases. Would you think the different warning cases are not needed and a single one, e.g. "Failed to build time object" would be sufficient?

gammapy/data/io.py
LeanderSchlegel: Here we also thought, if a warning is less strict and the other columns could still be loaded in case the time-conversion fails.

LeanderSchlegel:
I cleaned up the test file by removing the commentary of the tests and ensuring consistent names for each test. I would be through with my changes for now and wait for your feedback, thank you a lot for your review so far!

gammapy/data/io.py
```diff
@@ -141,3 +145,106 @@ def to_hdu(self, event_list, format="gadf"):
             raise ValueError(f"Only the 'gadf' format supported, got {format}")
 
         return self._to_gadf_table_hdu(event_list)
+
+
+class ObservationTableReader:
+    """Reader class for ObservationTable"""
+
+    def read(self, filename, hdu=None):
+        """Read `~gammapy.data.ObservationTable` from file.
```
AtreyeeS: Maybe you can be explicit here, like "Reads a Observation index table in GADF format and converts to internal `~gammapy.data.ObservationTable` "

gammapy/data/io.py
```diff
@@ -141,3 +145,106 @@ def to_hdu(self, event_list, format="gadf"):
...
+            Table in gadf 0.2/0.3 format, see [1], [2].
+
+        Returns
+        -------
+        ObservationTable : `~gammapy.data.ObservationTable`
+            ObservationTable in internal data format.
+
+        References
+        ----------
+        [1] Gamma-ray astronomy community, 2018, 'Data formats for gamma-ray astronomy v0.2 <https://gamma-astro-data-formats.readthedocs.io/en/v0.2/data_storage/obs_index/index.html>'_
```
AtreyeeS: Do you want to put both refs?

gammapy/data/io.py
```diff
@@ -141,3 +145,106 @@ def to_hdu(self, event_list, format="gadf"):
...
+        [2] Gamma-ray astronomy community, 2018, 'Data formats for gamma-ray astronomy v0.3 <https://gamma-astro-data-formats.readthedocs.io/en/v0.3/data_storage/obs_index/index.html>'_
+        """
+
+        names_gadf = table_gadf.colnames
+        meta_gadf = table_gadf.meta
+
+        required_names_gadf = [
+            "OBS_ID",
```
AtreyeeS, line 213: I am confused. GADF has more required names, no?

AtreyeeS: Ok I understand now. Maybe remove the `gadf` from the name

gammapy/data/io.py
```diff
@@ -141,3 +145,106 @@ def to_hdu(self, event_list, format="gadf"):
...
+
+        time_columns = set(["TSTART", "TSTOP"]).intersection(set(names_gadf))
+
+        if len(time_columns) != 0:
+            try:
+                time_ref = time_ref_from_dict(meta_gadf)
+                time_unit = meta_gadf["TIMEUNIT"]
+            except KeyError:
+                log.warning(
+                    "Found column TSTART or TSTOP in gadf table, but can not create columns in internal format (MixinColumn Time) due to missing header keywords in file."
```
AtreyeeS: The user probably does not care about (MixinColumn Time) internal formats. Rather, print what header keyword is missing

gammapy/data/io.py
```diff
@@ -141,3 +145,106 @@ def to_hdu(self, event_list, format="gadf"):
...
+                    log.warning(f"Could not convert type for column {colname}.")
+                    table_gadf.remove_column(colname)
+                    break
+                try:
+                    q = Quantity(col_type_converted, time_unit)
+                except ValueError:
+                    log.warning(f"Invalid unit for column {colname}.")
+                    table_gadf.remove_column(colname)
+                    break
+                table_gadf[colname] = time_ref + q
```
AtreyeeS: Why dont you save as time object?

gammapy/data/tests/test_io.py
```diff
@@ -50,3 +54,107 @@ def test_eventlist_reader_empty_gadf_table():
     assert "TIME" in str(err.value)
     assert "RA" in str(err.value)
     assert "DEC" in str(err.value)
+
+
+@requires_data()
+def test_observationtable_reader_unknown_hdu_extension():
+    hess_obs_table = make_path("$GAMMAPY_DATA/hess-dl3-dr1/obs-index.fits.gz")
+
+    with pytest.raises(AstropyDeprecationWarning):
```
AtreyeeS: Why are you introducing a deprecated feature?

gammapy/data/io.py
```diff
@@ -141,3 +145,106 @@ def to_hdu(self, event_list, format="gadf"):
...
+        ----------
+        filename : `pathlib.Path`, str
+            Filename
+        hdu : str, optional
+            Name of observation table HDU. Default is None.
+
+        Returns
+        -------
+        ObservationTable : `~gammapy.data.ObservationTable`
+            ObservationTable from disk converted into internal data format.
```
Astro-Kirsty: This line should describes what it returns i.e. "Observation table." 
And rather the information that you mention here can be mentioned in the first part of the docstring, as Atreyee suggests

gammapy/data/io.py
```diff
@@ -141,3 +145,106 @@ def to_hdu(self, event_list, format="gadf"):
...
+        """
+        filename = make_path(filename)
+
+        table_disk = Table.read(filename, format="fits", hdu=hdu)
```
Astro-Kirsty: I would replace `table_disk` with just `table`

gammapy/data/io.py
```diff
@@ -141,3 +145,106 @@ def to_hdu(self, event_list, format="gadf"):
...
+
+        format = table_disk.meta.get("HDUCLASS")
+
+        if format is None:
+            format = "GADF"
+            log.warning(
+                f"Could not infer fileformat from metadata in {filename}, assuming GADF."
```
Astro-Kirsty:
```suggestion
                f"Could not infer the file format from metadata in {filename}, assuming GADF."
```

gammapy/data/io.py
```diff
@@ -141,3 +145,106 @@ def to_hdu(self, event_list, format="gadf"):
...
+            )
+
+        if format.upper() in ["GADF", "OGIP"]:
+            return self._from_gadf_table(table_disk)
+        else:
+            raise ValueError(f"Unknown fileformat :{format}")
```
Astro-Kirsty:
```suggestion
            raise ValueError(f"Unknown file format :{format}")
```

As a side note, do we only have "GADF", "OGIP"? Because if those are the only 'accepted' then we could expand on this message a little more to say

`f"Expected one of: {valid_formats}."`

gammapy/data/io.py
```diff
@@ -141,3 +145,106 @@ def to_hdu(self, event_list, format="gadf"):
...
+
+    @staticmethod
+    def _from_gadf_table(table_gadf):
+        """Convert gadf observation table into`~gammapy.data.ObservationTable`.
```
Astro-Kirsty:
```suggestion
        """Convert GADF observation table into `~gammapy.data.ObservationTable`.
```

gammapy/data/io.py
```diff
@@ -141,3 +145,106 @@ def to_hdu(self, event_list, format="gadf"):
...
+
+        Parameters
+        ----------
+        table_gadf : `~astropy.Table.table`
+            Table in gadf 0.2/0.3 format, see [1], [2].
+
+        Returns
+        -------
+        ObservationTable : `~gammapy.data.ObservationTable`
+            ObservationTable in internal data format.
```
Astro-Kirsty: Same as above

gammapy/data/io.py
```diff
@@ -141,3 +145,106 @@ def to_hdu(self, event_list, format="gadf"):
...
+            raise ValueError(f"Unknown fileformat :{format}")
+
+    @staticmethod
+    def _from_gadf_table(table_gadf):
+        """Convert gadf observation table into`~gammapy.data.ObservationTable`.
+
+        Parameters
+        ----------
+        table_gadf : `~astropy.Table.table`
+            Table in gadf 0.2/0.3 format, see [1], [2].
```
Astro-Kirsty:
```suggestion
            Table in GADF format (version 0.2 or 0.3). 
            See [1]_, [2]_ for further information. 
```

gammapy/data/io.py
```diff
@@ -141,3 +145,106 @@ def to_hdu(self, event_list, format="gadf"):
...
+
+        Returns
+        -------
+        ObservationTable : `~gammapy.data.ObservationTable`
+            ObservationTable in internal data format.
+
+        References
+        ----------
+        [1] Gamma-ray astronomy community, 2018, 'Data formats for gamma-ray astronomy v0.2 <https://gamma-astro-data-formats.readthedocs.io/en/v0.2/data_storage/obs_index/index.html>'_
+        [2] Gamma-ray astronomy community, 2018, 'Data formats for gamma-ray astronomy v0.3 <https://gamma-astro-data-formats.readthedocs.io/en/v0.3/data_storage/obs_index/index.html>'_
```
Astro-Kirsty: The correct format for references can be found [here](https://docs.gammapy.org/dev/development/doc_howto.html#code-in-docstrings-in-python-files).

gammapy/data/io.py
Astro-Kirsty: I am a confused! Can you explain what is happening?

gammapy/data/io.py
```diff
@@ -141,3 +145,106 @@ def to_hdu(self, event_list, format="gadf"):
...
+
+        names_gadf = table_gadf.colnames
+        meta_gadf = table_gadf.meta
+
+        required_names_gadf = [
+            "OBS_ID",
+        ]
+
+        missing_names = set(required_names_gadf).difference(names_gadf)
+        if len(missing_names) != 0:
```
Astro-Kirsty:
```suggestion
        if missing_names:
```

I think this should work

gammapy/data/io.py
```diff
@@ -141,3 +145,106 @@ def to_hdu(self, event_list, format="gadf"):
...
+            raise RuntimeError(
+                f"Not all columns required to read from GADF were found in file. Missing: {missing_names}"
```
Astro-Kirsty: Maybe instead something like:

"The input table is missing required GADF columns. The missing columns are: {missing_names}."

gammapy/data/io.py
```diff
@@ -141,3 +145,106 @@ def to_hdu(self, event_list, format="gadf"):
...
+                table_gadf.remove_columns(time_columns)
+                return ObservationTable(
+                    data=table_gadf, meta=meta_gadf
+                )  # TODO: Replace "data" by "table" once new constructor is added in class ObservationTable.
+
+            for colname in time_columns:
+                try:
+                    col_type_converted = table_gadf[colname].astype("float64")
+                except ValueError:
+                    log.warning(f"Could not convert type for column {colname}.")
```
Astro-Kirsty: I would make this warning a little more description, so that the user understands what they should/shouldn't do following it.

gammapy/data/io.py
```diff
@@ -141,3 +145,106 @@ def to_hdu(self, event_list, format="gadf"):
...
+                    table_gadf.remove_column(colname)
+                    break
+                try:
+                    q = Quantity(col_type_converted, time_unit)
+                except ValueError:
+                    log.warning(f"Invalid unit for column {colname}.")
```
Astro-Kirsty: Same as above

gammapy/data/io.py
LeanderSchlegel: Thank you very much for your suggestion, I applied it in 4eb29eb!

gammapy/data/io.py
LeanderSchlegel: Thank you very much for the point, I took the v.0.2 out in 4eb29eb, I think the link to the latest version is probably enough

gammapy/data/io.py
LeanderSchlegel: Thank you, I agree completely and removed it!

gammapy/data/io.py
LeanderSchlegel: Thank you, I completely agree and took "(MixinColumn Time)" out in 695334e and will try to print the missing header keywords. Is there a way I can get them out of the KeyError? Or would I have to check for the keywords myself? It would not be much code, I just hoped I could save it, by relying on the `time_ref_from_dict(meta)`.

gammapy/data/tests/test_io.py
LeanderSchlegel: When testing the code, I encountered the problem, that some data files do not contain the HDU-name "OBS_INDEX" for the observation index table and I had to guess the location of the obs-index HDU with extra code in those cases. This feature is also implemented in astropys table reader and I tried to use it there. When passing the hdu-keyword and it is not recognized, like in this test, it guesses the location of the HDU, but this feature will be deprecated in astropy. I will think if there is a better way how to solve this.

LeanderSchlegel:
Thank you a lot for your review and all your comments, @AtreyeeS and @Astro-Kirsty! I will go through them!

gammapy/data/io.py
LeanderSchlegel: Thank you, I already changed to uppercase and applied the space with ea07db1!

gammapy/data/io.py
LeanderSchlegel: Thank you, that is better and clearer! Applied it with a468269!

gammapy/data/tests/test_io.py
Astro-Kirsty: Could you provide an example where it failed? 

Could it be because it contains "OBS-INDEX" instead?

gammapy/data/io.py
```diff
@@ -141,3 +145,121 @@ def to_hdu(self, event_list, format="gadf"):
...
+                time_ref = time_ref_from_dict(meta_gadf)
+                time_unit = meta_gadf["TIMEUNIT"]
+            except KeyError:
+                log.warning(
+                    "Found column TSTART or TSTOP in GADF table, but can not create columns in internal format due to missing header keywords in file."
+                )
+                table_gadf.remove_columns(time_columns)
+                return ObservationTable(
+                    data=table_gadf, meta=meta_gadf
+                )  # TODO: Replace "data" by "table" once new constructor is added in class ObservationTable.
```
Astro-Kirsty: For the #TODO are they to be fixed in another PR? 
I prefer that we would create an issue for that, to make sure we do it.

gammapy/data/io.py
```diff
@@ -141,3 +145,121 @@ def to_hdu(self, event_list, format="gadf"):
...
+                    table_gadf[colname] = time_ref + q
+                except TypeError:
+                    log.warning(
+                        f"Unit for column {colname} is in wrong dimension. Unit with dimension of time expected."
+                    )
+                    table_gadf.remove_column(colname)
+
+        return ObservationTable(
+            data=table_gadf, meta=meta_gadf
+        )  # TODO: Replace "data" by "table" once new constructor is added in class ObservationTable.
```
Astro-Kirsty: This can be included in the same issue as the above TODO

gammapy/data/io.py
LeanderSchlegel: Thank you very much, when thinking about it, I noticed that there was an uncatched error. So far I tested that a unit that can not be understood like "-" fails, but not a wrong dimension like "m" instead of "s" is captured by this same failure. I added another try-except for this case and a corresponding test. I'm not sure, if I catch this way every possible case, could it maybe work this way?

LeanderSchlegel: I implemented the new try except and test in 9b37938

gammapy/data/io.py
LeanderSchlegel: Thank you, I added "to float64" to let the user know, that it is a problem of the type-conversion into float

gammapy/data/io.py
LeanderSchlegel: Good point, I will do an issue! Yes they are fixed in a second PR we plan to update the observation table class to use the new reader.

gammapy/data/tests/test_io.py
LeanderSchlegel: I have to admit, I am still trying to understand the issue, I think the deprecation warning only appears, if the user specifies the hdu-name and the astropy.table-reader can not find it, telling that it is reading the first available table, that is the first hdu-extension (hdu=1) in the list.
If it is not specified, I think the reader tries also to read hdu=1 as standard but without warning in this case and the reading works. I will check this further.

I tested four obs-index tables and in the first two the keyword "EXTNAME" is present, while in the third and fourth, it is not found. The deprecation warning is only raised, if a user specifies "OBS_INDEX" in these two cases, where it is not found. The default is, to not specify it, the default value in the code is `hdu=None`.
If the user does not specify it, the reader does not fail in all four cases. Thank you for the idea with the spelling, I think it is not the cause, I tested to print the whole header and did not saw it differently spelled.


```
from astropy.io import fits
from gammapy.data.io import ObservationTableReader

path= "../gammapy-datasets/dev/hess-dl3-dr1/obs-index.fits.gz"
hdul = fits.open(path)
header =hdul[1].header
if "EXTNAME" in header.keys():
    print("HESS",header["EXTNAME"])

#reader = ObservationTableReader
#reader.read(reader,filename=path,hdu="OBS_INDEX")

path="../gammapy-datasets/dev/magic/rad_max/data/obs-index.fits.gz"
hdul = fits.open(path)
header =hdul[1].header
if "EXTNAME" in header.keys():
    print("MAGIC",header["EXTNAME"])

#reader = ObservationTableReader
#reader.read(reader,filename=path,hdu="OBS_INDEX")

path="../gammapy-datasets/dev/tests/format/swgo/obs-index.fits.gz"
hdul = fits.open(path)
header =hdul[1].header
if "EXTNAME" in header.keys():
    print("SWGO",header["EXTNAME"])

#reader = ObservationTableReader
#reader.read(reader,filename=path,hdu="OBS_INDEX")

path="../gammapy-datasets/dev/hawc/crab_events_pass4/obs-index-table-GP-no-events.fits.gz"
hdul = fits.open(path)
header =hdul[1].header
if "EXTNAME" in header.keys():
    print("HAWC",header["EXTNAME"])

reader = ObservationTableReader
reader.read(reader,filename=path,hdu="OBS_INDEX")
```

LeanderSchlegel: [TestHDU.ipynb.txt](https://github.com/user-attachments/files/22641847/TestHDU.ipynb.txt)

gammapy/data/io.py
Astro-Kirsty: Okay but then I would remove the TODO here and for the issue simply write that for the `_from_gadf_table` in the `ObservationTableReader` we want to return an ObservationTable were we replace "data" by "table" once new constructor is added in class ObservationTable.

LeanderSchlegel: Thank you, I agree and will remove the TODO and adapt the issue!

gammapy/data/io.py
LeanderSchlegel: Do you mean why I use the MixinColumn with `table_gadf[colname] = time_ref + q`? I based this on the code in obs_table.py, where the absolute time is calculated from the reference time and the time-columns entries, the code is  in `time_start` and `time_stop`: 
```
    @property
    def time_start(self):
    """Observation start time as a `~astropy.time.Time` object."""
    return self.time_ref + Quantity(self["TSTART"], "second")`
```
If I see it correctly, the sum is still a time-object and we thought of directly overwriting the existing column with the MixinColum /Time object to replace the column with the proper time in the internal format while keeping the other columns.

gammapy/data/tests/test_io.py
LeanderSchlegel: I tried to look a bit more into how the reader works in case the hdu is not specified and while I'm not completely sure, I think if no hdu is specified, it tries to read the first available hdu extension at index 1 of the hdulist. If this is the only extension available in the file (at index 0 is the primary hdu and at index 1 the only extension), it gives no warning and thereby the reading does not fail, even without specifying the extension by name. In case of multiple hdus and no specification, it warns, but this feature is not deprecated:

```
from astropy.io import fits
from gammapy.data.io import ObservationTableReader

path= "../gammapy-datasets/dev/hess-dl3-dr1/obs-index.fits.gz" #Path to original obs-index table file
path2= "../gammapy-datasets/dev/hess-dl3-dr1/obs-index-test.fits.gz" #Path to test file to be written.

with fits.open(path) as hdul:
    print(hdul.info()) #The hdulist shows two hdus in the file, PRIMARY at index 0 and OBS_INDEX at index 1.

    #Output:
    #Filename: ../gammapy-datasets/dev/hess-dl3-dr1/obs-index.fits.gz
    #No.    Name      Ver    Type      Cards   Dimensions   Format
    #0  PRIMARY       1 PrimaryHDU       4   ()      
    #1  OBS_INDEX     1 BinTableHDU    104   105R x 31C   [K, E, E, E, E, E, E, E, 18A, E, E, E, E, E, E, D, D, 10A, 12A, 10A, 12A, K, 7A, K, E, K, 21A, 14A, E, E, E]   
    #None
    
    ext2 =hdul[1] #Testwise copy first hdu extension.
    hdul.append(ext2) #and append it to the list, to create a second hdu at index 2.
    hdul.writeto(path2,overwrite=True) #Write into test file, that has now two table-extensions.
    
reader = ObservationTableReader
reader.read(reader,filename=path2) #The reader shows a warning, that multiple tables are present and the first is taken.

#Output:
#WARNING: hdu= was not specified but multiple tables are present, reading in first available table (hdu=1) [astropy.io.fits.connect]
```

In case of a non-found hdu, that was specified by name, it warns and this leads also to the deprecation warning by astropy. I think the case of the deprecation warning does not have to occur often for the user, as the default argument is hdu=None. In case this feature is removed from astropy in the future, our thought was, that the test fails and we are informed to change the code.

A possibility could also be to avoid the deprecated feature, by opening the fits-file like in the example-code or the EventListReader class and checking if the hdu extension given by the user is available, to raise an own warning. A problem that @registerrier made me aware of is, that in this case, the table is opened twice, as the `Table.read` will open the table as well and it is less optimized to have the code doubled, but we could add the code to avoid the deprecation warning.

LeanderSchlegel: Thank you very much for our discussion in the dev call today, I took the test out in 62fa1c3

LeanderSchlegel:
Thank you very much for discussing in the dev call today! As decided, I took the test for the unknown hdu extension out and would be through with my changes for now and am very happy for your re-review :)

PR #6176, opened 2025-10-01
infra: drop setup cfg for pyproject

### LINKED ISSUE #6129: Move project setup and metadata to `pyproject.toml`
Move all the package configuration options from `setup.cfg` to `pyproject.toml` to bring Gammapy inline with [PEP621](https://peps.python.org/pep-0621/). This also makes possible the use of uv and pixi seamlessly.

Get rid of setup.cfg and use instead only `pyproject.toml` for project configuration and metadata as recommended by  [PEP621](https://peps.python.org/pep-0621/)

**Additional context**
See configuration of astropy, numpy, matplotlib, etc.

natthan-pigoux:
**Description**

- drop `setup.cfg` for `pyproject.toml` completely
- remove most of setup.cfg references
- keep `setup.py` since pyproject do not yet handle properly cython extension: https://github.com/pypa/setuptools/discussions/4154
- Add a `.coveragerc` to configure pytest coverage

Closes  #6129

**Dear reviewer**

Some setup.cfg references are still there in the [Makefile](https://github.com/gammapy/gammapy/blob/main/Makefile#L32) and [codemeta.py](https://github.com/gammapy/gammapy/blob/main/dev/codemeta.py). 

Shloud we remove `codemeta.py`?


natthan-pigoux:
Quote: https://github.com/gammapy/gammapy/issues/6129#issuecomment-3319303315 

@bkhelifi how can I check that?

natthan-pigoux:
Any idea why is this failing? doesn't seems to be an issue for me locally...

registerrier:
> Any idea why is this failing? doesn't seems to be an issue for me locally...

I think it means the test does not find the CITATION file. 

See https://github.com/gammapy/gammapy/blob/34e803c7d7ab634bc63b80892bdc8a524d7c37a9/gammapy/__init__.py#L83-L92

registerrier:
Now thinking about 

> > Any idea why is this failing? doesn't seems to be an issue for me locally...
> 
> I think it means the test does not find the CITATION file.
> 

In fact, in any installation, the file cannot be found this way (it is actually not present...). In v2.0, it does not work. 
```
In [1]: import gammapy

In [2]: gammapy.__maincitation__
Out[2]: ''
``` 
It used to work in the dev environment, but not in this approach (I don't know why). My proposal here is to `xfail` the failing tests in test_init.py and open an issue about this.

natthan-pigoux:
Thanks @registerrier for having a look, I will open the issue.

registerrier:
> Thanks @registerrier for having a look, I will open the issue.

It is opened : #6181 .

pyproject.toml
```diff
@@ -1,19 +1,126 @@
+[project]
+name = "gammapy"
+description = "A Python package for gamma-ray astronomy"
+readme = "LONG_DESCRIPTION.rst"
+license = { text = "BSD-3-Clause" }
```
morcuended:
```suggestion
license = "BSD-3-Clause"
```
See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license

.flake8
```diff
@@ -0,0 +1,11 @@
+[codespell]
+skip = [".git"]
+ignore-words = "dev/codespell/ignore-words.txt"
+exclude-file = "dev/codespell/exclude-file.txt"
+
+[flake8]
```
morcuended: I think all the flake8 setup leftovers should be removed. We should be using ruff now.

morcuended (commented):
Thanks @natthan-pigoux. I left some comments

.flake8
natthan-pigoux: Ok, that what I was thinking. But then I will remove the "codestyle" env from tox.

pyproject.toml
```diff
@@ -1,19 +1,128 @@
...
 [tool.pytest.ini_options]
 filterwarnings = [
     "error::astropy.utils.exceptions.AstropyDeprecationWarning",
     "error::gammapy.utils.deprecation.GammapyDeprecationWarning",
     "error::matplotlib.MatplotlibDeprecationWarning",
 ]
 
+[tool.setuptools]
+zip-safe = false
+
```
morcuended: I think this can be dropped:

> Obsolete - only relevant for `pkg_resources`, `easy_install` and `setup.py` install in the context of [eggs](https://setuptools.pypa.io/en/latest/deprecated/python_eggs.html) (deprecated).

docs/development/intro.rst
```diff
@@ -316,15 +316,6 @@ To build the documentation locally you can use:
...
 
-The codestyle can be checked using the command:
-
-.. code-block:: bash
-
-    tox -e codestyle
-
-Which will run the tool `flake8` to check for code style issues.
-
-
```
morcuended, line 327: Maybe here instead of removing it completely, say that the codestyle will be enforced by the pre-commit which includes ruff for linting and formatting (running `pre-commit run`)

morcuended (dismissed):
Thanks @natthan-pigoux. It looks good to me. I just left some comments. I think more cleanup of pylint, black, isort, etc. settings will be needed later.

natthan-pigoux:
> Thanks @natthan-pigoux. It looks good to me. I just left some comments. I think more cleanup of pylint, black, isort, etc. settings will be needed later.

Thank you for the comments. Yes, I basically just copied without doing more cleanup.

docs/development/intro.rst
natthan-pigoux: I added some infos here: https://github.com/gammapy/gammapy/pull/6176/files#diff-12226b9fdd8ddf96d059f1baa3d2705d5fb8a0248f21f9be1b0cf7a01f9d1b53R267-R275

Is that enough?

morcuended: Yes, I did not see it before

morcuended (dismissed):
Thanks! It looks good to me now

pyproject.toml
```diff
@@ -105,7 +105,7 @@ docs = [
 include = ["gammapy*"]
 
 [tool.setuptools.package-data]
-"gammapy" = ["../CITATION", "../CITATION.cff"]
+"gammapy" = ["../CITATION", "../CITATION.cff", "../codemeta.json", "../LICENSE", "../README"]
```
morcuended: I think the proper way of including these files is through `MANIFEST.in` instead of through package data, which I think it is though to be used when those resources are included in the source code itself.

See https://setuptools.pypa.io/en/latest/userguide/miscellaneous.html

morcuended (dismissed):
Thanks @natthan-pigoux, from my side this is ready. I also added a fragment for the changelog. The only thing is that the check for the changelog fragment is not working. It needs https://github.com/gammapy/gammapy/pull/6217. But it shouldn't be a reason to block the merge IMO

bkhelifi:
I had a careful look on `pyproject.toml`
I have noticed that we lose the 2 following metadata:

> author = The Gammapy developers
> author_email = gammapy-coordination-l@in2p3.fr

Also, the version of toml is not fiven as comment or metadata.

Can we handle these 3 points?

morcuended:
> I had a careful look on `pyproject.toml` I have noticed that we lose the 2 following metadata:
> 
> > author = The Gammapy developers
> > author_email = [gammapy-coordination-l@in2p3.fr](mailto:gammapy-coordination-l@in2p3.fr)

This is now in:
```TOML
authors = [
    { name = "The Gammapy developers", email = "gammapy-coordination-l@in2p3.fr" }
]
```

> Also, the version of toml is not fiven as comment or metadata.

I've never seen this being defined in any `pyproject.toml` of any project, I think. Can you give and example?

registerrier (commented):
Thanks @natthan-pigoux and @morcuended . Looks good.

One question: (when) do we backport? A priori, yes. Maybe wait for nightly wheels build to make sure we can merge on 2.0.x

bkhelifi:
I have no example for toml. But other file formats are able to precise the format, e.g.

```
# %ECSV 1.0
# ---
```
Do you see my point?

morcuended:
> I have no example for toml. But other file formats are able to precise the format, e.g.
> 
> ```
> # %ECSV 1.0
> # ---
> ```
> 
> Do you see my point?

I understand now, thanks. I think this is not used for `pyproject.toml` or I've never seen it in any project.

bkhelifi:
> > I have no example for toml. But other file formats are able to precise the format, e.g.
> > ```
> > # %ECSV 1.0
> > # ---
> > ```
> I understand now, thanks. I think this is not used for `pyproject.toml` or I've never seen it in any project.

I do not know the FAIR practices within the TOML community. And, for the case of TOML, there is only 1 version of such format! So, I am not surprised that there is no example.
This metadata (as comment) is so basic and can anticipate the future;) But this is not mandatory and it is not a NoGo for me. 
-> As you prefer;)

morcuended:
After #6150 is merged, we have to make some changes here in the `.coveragerc` that are currently in `setup.cfg`

natthan-pigoux:
> After #6150 is merged, we have to make some changes here in the `.coveragerc` that are currently in `setup.cfg`

I think everything concerning the coverage in `setup.cfg` should already be in the `.coveragerc`.

.coveragerc
```diff
@@ -0,0 +1,54 @@
...
+    */gammapy/tests/*
+    */gammapy/*/tests/*
+    */gammapy/extern/*
+    */gammapy/version*
+
+# Make filenames relative to the repository root
+# for the coverage report to match source paths in sonarqube
+relative_files = True
+source =
+    gammapy
```
bkhelifi:
```suggestion
source = gammapy
```

natthan-pigoux: Thanks @bkhelifi for noticing the typo.

natthan-pigoux:
I think the codecov difference has to do with changes in the `coverage.rc` file. But not totally sure.

registerrier (approved):
Thanks @natthan-pigoux ! This looks good.
This is a very important change.

Issue #6138, opened 2025-09-14
Problems with temporal model.


mashachernyakova:
**Gammapy version**
Gammapy 2.0 (and same for Gammapy 1.3)

**Bug description**
Temporal model with scale factor in a phase curve model equal to 1 for all phase bins  significantly reduces the simulated counts, by more than an order of magnitude.

**Expected behavior**
A constant lightcurve on the output at the requested flux level, as well as the spectrum with the parameters specified in the input yaml file,  similar to what is produced by gammapy 1.2

**To Reproduce**

**Other information**

AtreyeeS:
Thanks @mashachernyakova.
Indeed, the `TemplatePhaseCurveTemporalModel` was adapted after the 1.2 release, because we realized there were issues with simulations of pulsed emission. The assumption made in v1.3 (and 2.0) is that the spectral model on your SkyModel is the phase averaged spectrum, and thus, **the normalization of the `TemplatePhaseCurveTemporalModel` should integrate to `1`.** This normalization is done internally

The scaling of the spectral model amplitude will have to be adjusted accordingly.

mashachernyakova:
Dear @AtreyeeS , but in the model I tried (normalisation is equal to 1 for all orbital phases from 0 to1), the integral should be equal to 1, right?

AtreyeeS:
Can you provide a minimal working example to show the difference in the two versions?

Cormed-Mack:
Hi, I can provide some examples, give me some time though.

Cormed-Mack:
Hi, I've attached a file containing examples, inside is the model used, (I've elected to use a very simple constant phasecurve this time to show the issue) and two folders labelled "1.2" and "1.3" they each contain the respective counts maps, ts maps, lightcurves and spectra that were generated using the specified gammapy versions. Let us know what you think!!
Kind Regards,
C. Dignam

[Examples.tar.gz](https://github.com/user-attachments/files/22337076/Examples.tar.gz)

AtreyeeS:
Thank you @Cormed-Mack ! The example you posted is difficult to execute without all the links.
Can you give a simple example, ie, just the full definition of your `model_simu` without linking to the irfs or your `models` list

Cormed-Mack:
I've cut out most of the code and it now just prints the model values, is this what you meant? You'll need to include all attached files in the same directory to run it. You will need to unpack the phasecurve, for some reason github isn't allowing me to attach .FITS files directly.

[...]

PR #6182, opened 2025-10-02
Correct PhaseCurveTemporalModel integration bug

registerrier:
**Description**

This pull request solves issue #6138  by performing global normalization by the duration of the integration window as done in all temporal models.

Temporal models in gammapy do not carry the per unit time. They are only dimensionless normalization factor. Hence a source model will be:
$$\Phi(E, t) = \Phi_0(E)\ \psi(t) = \psi(t)\ \frac{dN}{dEdt}$$

Npred evaluation will rely on a integral of the model over time. Be default, a steady source will have $\psi(t)=1$.

In general, temporal model integration should yield:
$$I(t_1, t_2) =\frac{1}{t_2-t_1} \int_{t_1}^{t_2} \psi(t)\ dt $$

The phase model will work a bit differently. 

We can define the phase as a function of time:
$$\phi(t) = \phi_0 + f_0(t-t_0) + \frac{1}{2}f_1(t-t_0)^2 + \frac{1}{6}f_2(t-t_0)^3 $$
Our time model is defined as a function of phase : $\eta(\phi)$ and we have $\psi(t) = \eta(\phi(t))$.

If the time interval covers multiple periods defined by their starting times $\tau_i$  (i.e. $\tau_{i+1}=\tau_i + T_i$ with $1 \leq i \leq n$), we rewrite $I$ as
$$I(t_1, t_2) = \frac{1}{t_2-t_1} \left( \int_{t_1}^{\tau_1} \psi(t)\ dt + \sum_1^n \int_{\tau_i}^{\tau_{i+1}} \psi(t)\  dt + \int_{\tau_n}^{t_2} \psi(t)\ dt \right)$$

Then: 
$$I(t_1, t_2) = \frac{1}{t_2-t_1} \left( \int_{t_1}^{\tau_1} \psi(t)\ dt + \sum_1^n T_i\int_0^1 \eta(\phi)\  d\phi + \int_{\tau_n}^{t_2} \psi(t)\ dt \right)$$

The corrected implementation will perform this. The main assumption made is that the frequency does not significantly change during the integration time. This might prove incorrect for long integration times.

**Dear reviewer**

### DIFF
--- a/gammapy/modeling/models/tests/test_temporal.py
+++ b/gammapy/modeling/models/tests/test_temporal.py
@@ -33,13 +33,6 @@ def light_curve():
     return LightCurveTemplateTemporalModel.read(path)
 
 
-@pytest.fixture()
-def phase_curve_table():
-    phase = np.linspace(0.0, 1, 101)
-    norm = phase * (phase < 0.5) + (1 - phase) * (phase >= 0.5)
-    return Table(data={"PHASE": phase, "NORM": norm})
-
-
 @requires_data()
 def test_light_curve_str(light_curve):
     ss = str(light_curve)


QRemy (approved):
Thanks @registerrier looks good.

Astro-Kirsty:
Thanks @registerrier.
I suggest we also add extra documentation to explain this reasoning

AtreyeeS (approved):
Thanks @registerrier ! 

@fabiopintore can you please check if it resolves the integration issue for binaries?

MRegeard (approved):
@registerrier Thanks, looks good to me !

MRegeard:
> The main assumption made is that the frequency does not significantly change during the integration time. This might prove incorrect for long integration times.

Speaking for pulsar, I don't think that there is a use case for this model other than simulation. If you want to compute pulsar phase for real data, you **need** to use codes like PINT of Tempo2. So this assumption will always be correct in this case.

fabiopintore:
Thanks @registerrier . It looks good to me as well (although I haven't directly tested yet if it solves the issue with the binaries).

Issue #5985, opened 2025-07-16
`RADECSYS' is deprecated


bkhelifi:
**Bug description**
`RADECSYS' is deprecated, one should use now `RADESYSa`. The code and the FITS files in gammapy data should be updated.

LeanderSchlegel:
Thank you very much for assigning and discussing this issue with me, @bkhelifi! 
As we discussed, we could maybe need two steps:

**1) Update of deprecated header-keyword "RADECSYS" to "RADESYSa" for gammapy-own datasets.**

 I used the Astropy-documentation (https://docs.astropy.org/en/latest/io/fits/index.html and https://docs.astropy.org/en/latest/io/fits/api/headers.html) to write:

```
with fits.open(path) as hdul:
    header =hdul[1].header  # Load Header of "EVENTS"-extension of HDU.
    header.rename_keyword("RADECSYS","RADESYSa") # Rename deprecated keyword in this header.
    hdul.writeto(path,overwrite=True) #Overwrite existing file.
```

and tested that it updates the keyword correctly. 
I still check, if it does any other change to the file that could be a problem (trailing maybe, but should be harmless).
Probably, this solution is not the fastest, as the whole file is overwritten, in a brief check the `.flush()`-method seemed not to work, but most likely I can optimize the code still. 
A remaining problem could be, that it is all upper case, as FITS standard. Would this be a problem?

**2) Update the gammapy code, to be able to read files using "RADESYSa" and still be able to read old data using the deprecated keyword.**

LeanderSchlegel:
I started looking into **2) Update the gammapy code** and tried to identify the files that need to be changed.

I unfortunately have to disclaim, that I'm not sure if I correctly identified every file and it would be super kind, if maybe someone could check, that I did not overlooked files:

I tried to find the files containing the deprecated keyword on GitHub using the query: `org:gammapy RADECSYS`, which yields 96 files. I do not understand the search capacities and limitations completely yet (see https://docs.github.com/en/search-github/github-code-search/about-github-code-search) and it seems, that the query does not actually show all files containing the keyword (for the repo _gammapy/gammapy-docs-archive_, only 4 are shown contributing to the total of 96, which are for the dev-branch only, while for _gammapy/gammapy-docs_ 85 are shown, including version-copies from different branches). In any case, the number contains version-copies of individual files.

If the search-query finds all repositories and only the count is not accurate, 5 repositories remain.

[...]

PR #6189, opened 2025-10-08
Replace deprecated keyword by RADESYSa

LeanderSchlegel:
Dear all,

this small PR addresses Issue #5985 as a first step, by updating the deprecated keyword for the coordinate system (see https://docs.astropy.org/en/latest/api/astropy.wcs.Wcsprm.html) with its recommended update "RADESYSa" in the three identified code-files of gammypy.

In a separate future PR for the gammapy-data repository, the datafiles can be updated accordingly.

**Dear reviewer**
In a first try in #5985, I could only update the keyword update in the FITS capitalized as "RADESYSA", but most likely this can be still changed, shall we stay with the writing "RADESYSa"?
Do you know if I missed maybe any occurrence of the old keyword, that still needs to be updated?

Thank you a lot,
Leander

### DIFF
--- a/gammapy/data/event_list.py
+++ b/gammapy/data/event_list.py
@@ -992,7 +992,7 @@ class EventListChecker(Checker):
         # They are currently listed as required in the spec,
         # but I think we should just require ICRS and those
         # are irrelevant, should not be used.
-        # 'RADECSYS',
+        # 'RADESYSa',
         # 'EQUINOX',
         "ORIGIN",
         "TELESCOP",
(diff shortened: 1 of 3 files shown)


registerrier:
Regarding the occurence as a comment, I would say that a properly re-written `EventListWriter` should re-use part of the `EventListChecker` that is currently no-longer used.

LeanderSchlegel:
Thank you very much for your review and discussing, @registerrier!
Then I would keep the updated keyword in the occurrence as a comment to be correct when re-used and a PR for the gammapy-benchmarks repository is not needed.

PR #6193, opened 2025-10-13
Unbinned 1d analysis using reflected background

Tobychev:
A draft unbinned  1d analysis using reflected background


QRemy:
Thanks @Tobychev,  closing this prototype until activity on unbinned analyses restart.

PR #6194, opened 2025-10-15
Add fix to skip making cutouts if observation is far away from target…

Tobychev:
Using the high level interface to perform a "quick look" analysis of data via the sciance portal, it is not impossible to create a runlist containing runs from two different hemispheres. Currently this will lead to a crash because the HLI will attempt to make a cutout for each observation, but the region functionality this depends on throws an error if the regions used for cutting do not overlap.

As a temporary fix I added a check of the distance between the observation pointing at t_mid and the dataset sky position, skipping doing the cutout if the distance is larger than twice the cutout width. I think if the distance is large enough the proper thing is to not process the observation further and just return an empty dataset, but because I'm not entirely sure my distance check is entirely safe, I skip only the cutout.

A more proper solution would be to create a utility function that can filter out these useless very distant observations from the runlist before the `Analysis` class does any work at all on them, but that's for later.

### DIFF
--- a/docs/release-notes/6194.bug.rst
+++ b/docs/release-notes/6194.bug.rst
@@ -1,0 +1,1 @@
+Ignore observations without spatial overlap with the target dataset geom in DatasetsMaker instead of raising an error.


gammapy/makers/tests/test_reduce.py
```diff
@@ -316,6 +334,29 @@ def test_datasets_maker_map_2_steps(observations_cta, map_dataset):
...
+    observations_hess_far_apart, makers_map, crab_map_dataset
+):
+    makers = DatasetsMaker(
+        makers_map,
+        stack_datasets=True,
+        cutout_mode="partial",
+        cutout_width="5 deg",
+        n_jobs=1,
+    )
+    datasets = makers.run(crab_map_dataset, observations_hess_far_apart)
```
bkhelifi, line 348: you can capture the log.warning here

Astro-Kirsty:
I had a look into this and do not get an error when a safe mask is applied (which is expected). 
Please see the attached working code on dev
[PR6194.py](https://github.com/user-attachments/files/22971051/PR6194.py)

gammapy/makers/reduce.py
```diff
@@ -98,15 +99,24 @@ def make_dataset(self, dataset, observation):
...
+            try:
+                dataset_obs = dataset.cutout(
+                    **cutouts_kwargs,
+                )
+            except NoOverlapError:
+                log.warning(
+                    f"No cutout for {observation.obs_id} because"
+                    " it is outside the FOV of the reference geom."
+                )
+                dataset_obs = dataset.copy()
```
registerrier: Is it the correct thing to do here? In general this will copy a large dataset for no good reason. What happens if the mask of the `dataset` in question not False?

gammapy/makers/tests/test_reduce.py
```diff
@@ -316,6 +334,29 @@ def test_datasets_maker_map_2_steps(observations_cta, map_dataset):
     assert_allclose(exposure.data.mean(), 1.350841e09, rtol=3e-3)
 
 
+@requires_data()
+def test_datasets_maker_map_far_apart(
+    observations_hess_far_apart, makers_map, crab_map_dataset
+):
+    makers = DatasetsMaker(
+        makers_map,
```
registerrier, line 342: Is there an impact of the order in which makers are passed?

QRemy (commented):
Thanks @Tobychev I made a modification to ignore the invalid datasets instead of having a copy of the reference dataset.

Issue #6196, opened 2025-10-17
`FluxPoints.plot_ts_profiles` does not plot a lightcurve when there is only one time bin


fabiopintore:
**Is your feature request related to a problem? Please describe.**
When plotting the `FluxPoints` as a lightcurve, the `plot_ts_profiles` function does not show the `time` in the x-axis but the `energy`. This is because the `energy` and `time` axes have the same shape and, by construction, the `FluxPoints.geom.axes.primary_axis` is set as the `energy` axis. 

**Describe the solution you'd like**
A possibility would be to add an argument to select the axis to plot (as for `FluxPoints.plot(axis_name=...)`).

**Describe alternatives you've considered**
Alternatively, it could be possible to introduce a setter for `FluxPoints.geom.axes.primary_axis` so that the users can change it according to their needs

bkhelifi:
Maybe, this should be expose in the documentation or into an HowTo....

PR #6197, opened 2025-10-17
add axis_name argument and test

fabiopintore:
The pull request fixes the issue #6196 for the `FluxPoints.plot_ts_profiles`. It introduces an additional argument that allows to select the axis to plot.


QRemy (approved):
Thanks @fabiopintore, looks good.
The fail on devdeps CI seems unreleated.

registerrier:
I think this also suggests that a LightCurve object might be useful.

PR #6198, opened 2025-10-17
Update integrate_project_irf_on_geom to also support EffectiveAreas

Tobychev:
This is a very minimal update to `integrate_project_irf_on_geom` that allows it to also work with Effective Areas, I have not testes other IRFs yet.

### DIFF
--- a/gammapy/estimators/map/tests/test_ts.py
+++ b/gammapy/estimators/map/tests/test_ts.py
@@ -398,6 +398,7 @@ def test_compute_ts_map_downsampled(input_dataset):
     assert np.isnan(result["ts"].data[0, 30, 40])
 
 
+@pytest.mark.xfail
 def test_ts_map_stat_scan(fake_dataset):
     kernel_model = fake_dataset.models["source"]
 


gammapy/makers/tests/test_utils.py
```diff
@@ -151,18 +153,33 @@ def bkg_2d():
 def bkg_3d_custom(symmetry="constant", fov_align="RADEC"):
     if symmetry == "constant":
         data = np.ones((2, 3, 3))
+        sky_edges = [-3, -1, 1, 3] * u.deg
     elif symmetry == "symmetric":
         data = np.ones((2, 3, 3))
         data[:, 1, 1] *= 2
+        sky_edges = [-3, -1, 1, 3] * u.deg
+    elif symmetry == "hirez_symmetric":
```
bkhelifi, line 164: It seems that this symmetry is never used in the tests. Do I mistaken?

gammapy/makers/tests/test_utils.py
```diff
@@ -151,18 +153,33 @@ def bkg_2d():
...
+        data[1, 1] *= 2
+        data = (  # Upscale 3x3 to 9x9 by simply repeating
+            data.repeat(3, axis=0).repeat(3, axis=1)[np.newaxis, ...].repeat(2, axis=0)
+        )
+        sky_edges = np.linspace(-3, 3, 10) * u.deg
     elif symmetry == "asymmetric":
         data = np.indices((3, 3))[1] + 1
         data = np.stack(2 * [data])
+        sky_edges = [-3, -1, 1, 3] * u.deg
+    elif symmetry == "hirez_asymmetric":
```
bkhelifi: The same here...

gammapy/makers/tests/test_utils.py
Tobychev: I added a test that uses it now.

gammapy/makers/tests/test_utils.py
Tobychev: Didn't need it in the end, took it out.

bkhelifi:
What is the status here? So many tests are failing

Tobychev:
> What is the status here? So many tests are failing

I have no idea, I don't get these results running locally (instead I get a lot of `AttributeError: 'TableLoc' object has no attribute 'with_index'`) and the functions I see failing here (like `test_resample_axis`) don't use anything I've modified in this PR.

gammapy/makers/utils.py
```diff
@@ -659,9 +660,18 @@ def _get_fov_coord(
 ):
     """Return coord dict in fov_coord."""
     coords = {}
-
+    if isinstance(fov_frame, FoVICRSFrame) or (
+        fov_frame.obstime is fov_frame.origin.obstime
```
bkhelifi: I have not understood this second condition. Is it tested, by the way?

gammapy/makers/utils.py
```diff
@@ -710,24 +735,32 @@ def project_irf_on_geom(geom, irf, fov_frame, use_region_center=True):
...
+
     if not use_region_center:
         data = np.average(data, axis=-1, weights=weights, keepdims=True)
 
     return Map.from_geom(geom=geom, data=data.value, unit=data.unit)
 
 
 def integrate_project_irf_on_geom(geom, irf, fov_frame, use_region_center=True):
-    """Integrate and project an IRF on a given `~gammapy.maps.Geom` object according to a given FoV Frame.
+    """Integrate and project a BackgroundIRF on a given `~gammapy.maps.Geom` object according to a given FoV Frame.
```
bkhelifi:
```suggestion
    """Integrate and project a `~gammapy.irf.BackgroundIRF` on a given `~gammapy.maps.Geom` object according to a given FoV Frame.
```

bkhelifi (changes requested):
Thanks @Tobychev. I have some minor comments to ask you

gammapy/makers/utils.py
Tobychev: This function is for translating from an fov_frame to sky coordinates, and since the sky moves over time you have to keep track of at which time things happened. 

`fov_frame` can be either a single object, or an array valued object which has different type I don't remember at the moment. In the case of an array valued object there are a series of different obs times and since the sky moves over time we should not blindly use the sky coordinates of the origin that defined the frame without checking that the user is indeed requesting the sky coordinates for the fov_origin for the same instant in time at which the fov_origing was defined.

Not sure if this makes it clearer, its unfortunately a bit hard to explain. I include two pictures illustrating that the same FoV coordinates translate to different sky position at different times.
SkyRotDiag_0
SkyRotDiag_1

PR #6203, opened 2025-10-24
Add __repr__ method to Geom

Tobychev:
The geom classes don't have a `__repr__` method, which makes them annoying to use in the terminal. 

This PR adds an implementation that simply calls `self.__str__`


registerrier (approved):
Thanks @Tobychev . That's fine. It could have a test associated. But that's minor.

PR #6211, opened 2025-11-07
Multiple OFF regions in `makers.make_theta_squared_table`

fabiopintore:
This PR includes a new feature in `~gammapy.makers.utils.make_theta_squared_table`. It will allow the users to choose the number of OFF regions, currently limited to only one (the mirror position). The implementation gives more flexibility to analysis of the theta2 plots.
Tests are also included in the PR.

The following example shows the new feature. For a given set of observations and target positions:

```
rad_axis = MapAxis.from_bounds(0, 0.5, nbin=25, unit="deg2", interp="lin")

off_regions_number = 5
theta2_dist = make_theta_squared_table(observations=observations, 
                                                                       theta_squared_axis=rad_axis, 
                                                                       position=target_position, 
                                                                       position_off=None,
                                                                       off_regions_number = off_regions_number,
                                       )
```

The OFF regions will be five, evenly spaced around the pointing position as in figure:

Screenshot 2025-11-11 alle 08 28 20

and the theta2 plot for the OFF will take into account that `alpha` is now 0.2.

The `ReflectedRegionsFinder` approach cannot be adopted here as it limits the ON/OFF region size to the maximum angular separation to the pointing direction. However, a theta2 plot should be more flexible allowing for an unlimited exploration of the parameter space.


maxnoe:
Li&Ma does not hold if the counts are correlated, which is the case if the regions overlap. We should either prohibit this case or not provide a significance for distances where the regions start to overlap.

maxnoe:
Also, I think you've re-implemented the `WobbleRegionsFinder` here, you could just use it: https://docs.gammapy.org/2.0/api/gammapy.makers.WobbleRegionsFinder.html#gammapy.makers.WobbleRegionsFinder

fabiopintore:
thanks @maxnoe . I wasn't useing `WobbleRegionsFinder` to avoid the limitations on the overlapping region. However, it's true that Li&Ma cannot be used for correlated counts. A possibility could be to mask overlapping regions before computing the theta2 plot, but I think this is a too complex approach. Therefore, I've implemented the `WobbleRegionsFinder`.

p.s.: sorry, by mistake I removed you from the reviewer and I've realised only now :(

AtreyeeS:
@fabiopintore We discussed this a bit during the last dev call.
We need to understand better the usage and limitations of the theta_sq plot.
Double counting should be avoided, both for correlated of counts, and between on and off counts

fabiopintore:
Thanks @AtreyeeS . 
I agree that correlated counts should be definitely avoided. 

In the case of 1 ON and 1 OFF regions, actually they can overlap and Li&Ma can be adopted as far as you consider it at theta2 bin level. Instead, if you integrate the counts in the full ON and OFF regions, Li&Ma does not hold anymore! So, this is certainly something to be taken into account, but I suppose this is an issue for the user which is not adopting the function in the proper way, rather than a Gammapy issue.

It is instead a different argument when allowing for multiple OFF regions. In such cases, if regions overlaps, Li&Ma couldn't be used not even at the level of the theta2 bins. A possibility would be to mask all the overlapping regions and take this to correct the alpha parameter. However, this might be a bit tricky to handle and probably the best approach is simply to adopt the reflected regions.

AtreyeeS:
Whats the status here @fabiopintore ?

fabiopintore:
@AtreyeeS The PR is in stand-by at the moment. I was waiting for feedbacks after my last comment. The options in the theta2 plot calculation were  either 1) to allow only for the reflected regions or 2) to allow for overlapping OFF regions, but taking care properly on masking them. 
My first guess was that the reflected regions approach was easier to handle.

bkhelifi:
> In the case of 1 ON and 1 OFF regions, actually they can overlap and Li&Ma can be adopted as far as you consider it at theta2 bin level. Instead, if you integrate the counts in the full ON and OFF regions, Li&Ma does not hold anymore! So, this is certainly something to be taken into account, but I suppose this is an issue for the user which is not adopting the function in the proper way, rather than a Gammapy issue.

Pragmatically, I think that this does not occur for point sources or smally extended ones. Let's take an exemple with 0.5deg offset. The OFF position is the mirror position. In order to have overlapping ON-OFF regions and double-counting, one need theta2>0.25 deg2. Which is not what people are doing! Do you get my point or do I make a mistake? PS: the current doc could be update to precise the limitation or to compute it automatically!

> It is instead a different argument when allowing for multiple OFF regions. In such cases, if regions overlaps, Li&Ma couldn't be used not even at the level of the theta2 bins. A possibility would be to mask all the overlapping regions and take this to correct the alpha parameter. However, this might be a bit tricky to handle and probably the best approach is simply to adopt the reflected regions.

Adding a correlation term in the Li&Ma stat is a nightmare, and I guess that it is not permitted. So, one should find a way to avoid any double-counting! The masking has you propose, or the trick of the WobbleRegionsFinder that automatically limits the number OFF region given the offset and the theta^2 max value.

QRemy:
What's the status here ? do we just close this ?

fabiopintore:
I'm currently working on the solution we proposed. I would like to do a new push by today, if possible.
So, please do not close it.

QRemy:
Replaced by #6348.

PR #6213, opened 2025-11-12
Add a link towards the `npred` computation into the stats user-guide page

bkhelifi:
**Description**

All is in the title...

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -1,0 +1,42 @@
+name: Check changelog fragments
+
+on: [pull_request]
+
+jobs:
+  check-changelog:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v5
+        with:
+          fetch-depth: 0
+
+      - name: Verify changelog label and fragment
+        run: |
+          set -euo pipefail
+
+          # Obtain all labels set to the PR
+          labels="${{ toJson(github.event.pull_request.labels.*.name) }}"
+
+          # If the labels contains feature, bug or api-change then automatically require the 'changelog-required' label
+          if echo "$labels" | grep -Eq '"(feature|bug|api-change)"'; then
+            if ! echo "$labels" | grep -q '"changelog-required"'; then
+              echo "PR has 'feature', 'bug', or 'api-change' label but is missing 'changelog-required'."
+              exit 1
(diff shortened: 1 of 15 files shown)


bkhelifi:
If someone has an idea of the reason of the CI failure....

registerrier:
> If someone has an idea of the reason of the CI failure....

a new version of pytest I suppose: https://docs.pytest.org/en/stable/changelog.html

docs/user-guide/stats/index.rst
```diff
@@ -87,7 +87,12 @@ They are used also to estimate the excess counts significance, i.e. the probabil
 a given number of measured events :math:`n_{\mathrm{on}}` actually contains some signal events :math:`n_{\mathrm{sig}}`,
 as well as the errors associated to this estimated number of signal counts.
 
-.. _ts:
+.. _npred:
```
Astro-Kirsty: You should not be calling these the same name. I suggest to leave this as 'ts'

docs/user-guide/stats/index.rst
```diff
@@ -87,7 +87,12 @@ They are used also to estimate the excess counts significance, i.e. the probabil
 a given number of measured events :math:`n_{\mathrm{on}}` actually contains some signal events :math:`n_{\mathrm{sig}}`,
 as well as the errors associated to this estimated number of signal counts.
 
-.. _ts:
```
AtreyeeS, line 90: I think this line should remain, just move to line 96

docs/user-guide/stats/index.rst
```diff
@@ -87,7 +87,12 @@ They are used also to estimate the excess counts significance, i.e. the probabil
...
+.. _npred:
+
+Estimation of the predicted number of counts
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The formula are be found in this section: :ref:`npred`.
```
Astro-Kirsty, line 94:
```suggestion

.. _ts:
```

AtreyeeS (approved):
Thanks @bkhelifi ! No further comments.

@Astro-Kirsty Do we need a fragment here for the town crier?

Astro-Kirsty:
> @Astro-Kirsty Do we need a fragment here for the town crier?

Given that it is such a small PR, I would not add one

PR #6215, opened 2025-11-12
Solve CI fails caused by new iminuit and pytest version

registerrier:
**Description**

This pull request solves the issues introduced by pytest version 9.0 and the latest iminuit. 

Fixtures with additional decorators such as `require_data` are no longer working. They were ignored with a warning before. The additional decorators have been removed. 

**Dear reviewer**

### DIFF
--- a/dev/codespell/ignore-words.txt
+++ b/dev/codespell/ignore-words.txt
@@ -13,3 +13,4 @@ OptIn
 slac
 unparseable
 compiletime
+THIRDPARTY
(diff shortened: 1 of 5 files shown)


registerrier:
Merging now.

morcuended:
I still see more occurrences of `require_data` across the code. Are they supposed to be removed as well?

registerrier:
I think the issue is the association of `require_data` with a `pytest.fixture`. There remain 2 cases in the code, where a test class use the `require_data` decorator while it declares a `fixture` among its members. This does not seem to break pytest.

PR #6184, opened 2025-10-03
fix: run pre-commit on all files

natthan-pigoux:
fixes: #6183 

I changed the pre-commit ci job so that it runs `pre-commit run --all-files`.

I also run it locally and added the changes here. Otherwise next pre-commit ci jobs will failed.

PR #6221, opened 2025-11-14
Add monthly workflow to run pre-commit over all files with fixes

### LINKED ISSUE #6219: Monthly pre-commit run over all files
Monthly workflow to run `pre-commit run --all` and open a pull request with the fixes. Also allow manual trigger

morcuended:
Closes #6219 

Run `pre-commit run --all-files` then creates a PR applying the fixes.

It needs the updated configuration from #6184

### DIFF
--- a/.github/workflows/monthly_precommit.yml
+++ b/.github/workflows/monthly_precommit.yml
@@ -61,7 +61,8 @@ jobs:
       - name: Push branch
         if: steps.changes.outputs.changes == 'true'
         run: |
-          git push origin "$FIX_BRANCH"
+          echo "Pushing branch: ${{ steps.create_branch.outputs.branch }}"
+          git push origin "${{ steps.create_branch.outputs.branch }}"
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 


morcuended:
Trying it out in my fork: https://github.com/morcuended/gammapy/actions/runs/19571271283/job/56045148783

morcuended:
I think it is working: https://github.com/morcuended/gammapy/actions/runs/19629712820/job/56206318930

But this will not be able to push changes on `.github/workflows`. After all these tests, I think all this complexity is not worth and I would just run `pre-commit run --all` on every push and PR. I think after the first big linting, the rest of the times won't take much effort, even might be able to cache the pre-commit runs between jobs.

morcuended:
Needs the exclusion rules set in #6184

morcuended:
> Thanks @morcuended . This looks good. What is your conclusion? Does it work or should we have the linting on all files in each PR?

I think it works as is right now. But this automated check and push of changes does not work for `.github/workflows` folder. So it is currently blocked. All in all, having this monthly check feels a bit overkill looking at the alternatives listed below.

Steps:
- Need to merge https://github.com/gammapy/gammapy/pull/6184
- Someone needs to run pre-commit over the  `.github/workflows` and merge it.
- Only then this monthly will work.

Alternatives
- Integrate with https://pre-commit.ci/ bot (e.g. astropy)
- Do `pre-commit,run --all` in every PR/push (using ruff, so quite fast)

I would go for one of the alternatives.

natthan-pigoux:
Thanks @morcuended . I would also got for one of the alternatives. They seem to be the simplest options at the end.

morcuended:
Need to take a decision here on adopting one of the alternative proposals instead of the more complex monthly workflow.
 - Integrate with https://pre-commit.ci/ (just needs `.pre-commit-config.yaml` and sign-in with gammapy repository)
 - Run `pre-commit run --all` on every push, which runs over all the files (minimal change wrt current pre-commit stage in CI)
 
 I would go for the first option.

morcuended:
I would go for one of the simpler options in the previous comment. Closing this.

PR #6227, opened 2025-11-19
Corrected required model unit if exposure not applied

### LINKED ISSUE #6226: Wrong model unit required if exposure not applied
**Gammapy version**
2.0 and 1.3

**Bug description**
The required unit of a spectral model, for which the exposure is not applied, is at the moment "", i.e. unitless. This, however, leads to an error during the fit, as the units do not match up.

**Expected behavior**
There should not be a unit error during the fit.

**To Reproduce**
```python
dataset = SpectrumDatasetOnOff.read("$GAMMAPY_DATA/joint-crab/spectra/hess/pha_obs23523.fits")

template = TemplateSpectralModel(energy=[0.5,1,2]*u.keV, values=[1,2,3])

model = SkyModel(
    spectral_model=template*PowerLawNormSpectralModel(),
    name="source1",
    apply_irf={"exposure": False, "edisp": True}
)

dataset.models=[model]

fit = Fit()
result = fit.run(datasets=[dataset])
```

**Other information**
The required unit should be "MeV-1". A PR on this is to follow.

k-egg:
**Description**

This pull request fixes #6226. The required model unit for a model where exposure is not applied is changed to "MeV-1". This resolves the error during the fit.

**Dear reviewer**

This PR is ready for review!

### DIFF
--- a/gammapy/modeling/models/tests/test_cube.py
+++ b/gammapy/modeling/models/tests/test_cube.py
@@ -169,20 +169,15 @@ def test_sky_model_init():
         energy=[0.5, 1, 2] * u.TeV, values=[1, 2, 3] * u.dimensionless_unscaled
     )
 
-    template_exp = TemplateSpectralModel(
-        energy=[0.5, 1, 2] * u.TeV, values=[1, 2, 3] * (1 / (u.TeV * u.s * u.cm**2))
-    )
-
     template_noexp = TemplateSpectralModel(
         energy=[0.5, 1, 2] * u.TeV, values=[1, 2, 3] * (1 / u.TeV)
     )
 
-    SkyModel(spectral_model=template_exp)
-
-    SkyModel(
+    model_noexp = SkyModel(
         spectral_model=template_noexp,
         apply_irf={"exposure": False},
     )
+    assert model_noexp(None, None, 1 * u.TeV).unit == (1 / u.TeV)
 
     with pytest.raises(ValueError):
         SkyModel(spectral_model=template_dimensionless)


AtreyeeS:
Thanks @k-egg ! The `check_units` function should fail before the fit, at the model creation stage. I am confused why it is not failing on in it? 
Are we missing a test for this case of model creation with no exposure?

k-egg:
Hi @AtreyeeS! `check_units` does fail in 2.0/1.3 if you give "MeV-1" or something equivalent. If you do adhere to the unit requirement, you can create the model, but there will be an error during the fit.

gammapy/modeling/models/tests/test_cube.py
```diff
@@ -163,6 +164,29 @@ def test_sky_model_init():
...
 
+    # test unit checks with exposure applied
+    with pytest.raises(ValueError):
+        template = TemplateSpectralModel(energy=[0.5, 1, 2] * u.TeV, values=[1, 2, 3])
+        SkyModel(spectral_model=template * PowerLawNormSpectralModel())
+
+    template = TemplateSpectralModel(
+        energy=[0.5, 1, 2] * u.TeV, values=[1, 2, 3] * (1 / (u.TeV * u.s * u.cm**2))
+    )
+    SkyModel(spectral_model=template * PowerLawNormSpectralModel())
```
AtreyeeS: Why do you need the norm model here?

AtreyeeS (commented):
I see! So the `check_units` had wrong hard coded unit for this case.
This can go in as a bug-fix.

For the long term, I think `check_units` can be modified to evaluate the model once (on a predefined geom) and the final units tested. That can also help for more complex models...

gammapy/modeling/models/tests/test_cube.py
k-egg: Ah you're right, it's not necessary here. I'll change it.

gammapy/modeling/models/tests/test_cube.py
```diff
@@ -163,6 +164,33 @@ def test_sky_model_init():
     with pytest.raises(TypeError):
         SkyModel(spectral_model=PowerLawSpectralModel(), spatial_model=1234)
 
+    # test unit checks with exposure applied
+    with pytest.raises(ValueError):
+        template = TemplateSpectralModel(
+            energy=[0.5, 1, 2] * u.TeV, values=[1, 2, 3] * u.dimensionless_unscaled
+        )
```
registerrier: In general to structure the tests (Arrange-Act-Assert) and to limit code duplication, you could define first the two `TemplateSpectralModel` (template and template_dimensionless).
Then create the `SkyModel` objects with the exception capture.

registerrier (approved):
Thanks @k-egg . This looks good. I have left a small inline comment to structure the unit test.

gammapy/modeling/models/tests/test_cube.py
k-egg: Thanks for the comment! I've restructured the test to limit code duplication.

gammapy/modeling/models/tests/test_cube.py
```diff
@@ -163,6 +164,32 @@ def test_sky_model_init():
...
         SkyModel(spectral_model=PowerLawSpectralModel(), spatial_model=1234)
 
+    # test unit checks
+    template_dimensionless = TemplateSpectralModel(
+        energy=[0.5, 1, 2] * u.TeV, values=[1, 2, 3] * u.dimensionless_unscaled
+    )
+
+    template_exp = TemplateSpectralModel(
+        energy=[0.5, 1, 2] * u.TeV, values=[1, 2, 3] * (1 / (u.TeV * u.s * u.cm**2))
+    )
```
registerrier: This one is probably not needed here as it is the regular case, sorry if I suggested the opposite.

gammapy/modeling/models/tests/test_cube.py
```diff
@@ -163,6 +164,32 @@ def test_sky_model_init():
...
+
+    template_noexp = TemplateSpectralModel(
+        energy=[0.5, 1, 2] * u.TeV, values=[1, 2, 3] * (1 / u.TeV)
+    )
+
+    SkyModel(spectral_model=template_exp)
```
registerrier: Not needed here as it is standard and tested elsewhere.

gammapy/modeling/models/tests/test_cube.py
```diff
@@ -163,6 +164,32 @@ def test_sky_model_init():
...
+
+    SkyModel(
+        spectral_model=template_noexp,
+        apply_irf={"exposure": False},
+    )
```
registerrier: Here I would add an assert for instance on the evaluation.

registerrier (approved):
Thanks @k-egg . A last little comment :) 
The CI fail is unrelated, it is connected to a new astropy version which deprecates some table usage.

gammapy/modeling/models/tests/test_cube.py
k-egg: Hi, thanks for the comments. What would you say should be asserted here? The model unit upon evaluation or something else?

registerrier: > Hi, thanks for the comments. What would you say should be asserted here? The model unit upon evaluation or something else?

yes. The point is to make sure that an assert is there to avoid that some removes an apparently unused command in the future.

gammapy/modeling/models/tests/test_cube.py
k-egg: Ah I see, I removed it.

gammapy/modeling/models/tests/test_cube.py
k-egg: I removed this as well.

gammapy/modeling/models/tests/test_cube.py
k-egg: I added an assert of the model unit.

registerrier (approved):
Thanks @k-egg . No more comments on my side.

PR #6232, opened 2025-11-25
Dark Matter module: Include the option of using either CosmiXs or PPPC4 as a source for creating the DM spectra

alexcervino:
Hello, my name is Alexander Cerviño, and I am a PhD student at the High Energies Group (GAE) at the Complutense University of Madrid, supervised by Daniel Nieto. This is my first contribution to Gammapy.

In this contribution regarding the Dark Matter block, I have implemented the possibility of using CosmiXs instead of PPPC4 as a source for creating the DM spectra. The benefit of using CosmiXs is that the results they provide are more accurate since they obtain them while considering more robust approaches and more precise algorithms and physical processes.

For the implementation , I have set an option in the spectra.py file, so the user can choose between any of the sources, with PPPC4 as the default option. I have also made a match between the available channels in both projects and have found and indicated in the code that there are some discrepancies.

One thing to note is that I include the CosmiXs production file as an external source (path: gammapy/external/CosmiXs/Data/AtProduction-Gamma.dat), and I would like to know if it is possible to include it in the datasets module, just to be consistent with the procedure followed with PPPC4.

Additionally, after conducting some tests, I have checked that the differences in the results obtained by using the current Gammapy interpolator and the one provided by CosmiXs are negligible . Therefore , for simplicity, I have considered it correct to keep using the current one.

Thanks!

### DIFF
--- a/docs/release-notes/6232.feature.rst
+++ b/docs/release-notes/6232.feature.rst
@@ -1,0 +1,1 @@
+For the Dark Matter module: Implementation of the possibility of using CosmiXs or PPPC4 as a source for creating Dark Matter spectra.
--- a/docs/release-notes/6232.feature.rst
+++ b/docs/release-notes/6232.feature.rst
@@ -1,1 +1,1 @@
-Implementation of the option of using CosmiXs or PPPC4 as a source models for creating Dark Matter spectra.
+Implementation of the option of using ``CosmiXs`` or ``PPPC4`` as a source models for creating dark matter spectra.
--- a/docs/release-notes/6232.feature.rst
+++ b/docs/release-notes/6232.feature.rst
@@ -1,1 +1,1 @@
-For the Dark Matter module: Implementation of the possibility of using CosmiXs or PPPC4 as a source for creating Dark Matter spectra.
+Implementation of the option of using CosmiXs or PPPC4 as a source models for creating Dark Matter spectra.
--- a/gammapy/astro/darkmatter/spectra.py
+++ b/gammapy/astro/darkmatter/spectra.py
@@ -9,6 +9,7 @@ from gammapy.modeling import Parameter
 from gammapy.modeling.models import SpectralModel, TemplateNDSpectralModel
 from gammapy.utils.scripts import make_path
 from gammapy.utils.table import table_map_columns
+import warnings
 
 __all__ = ["PrimaryFlux", "DarkMatterAnnihilationSpectralModel"]
 
@@ -80,22 +81,60 @@ class PrimaryFlux(TemplateNDSpectralModel):
         "s": "s"
     }
 
+    mapping_dict_PPPC4_to_CosmiXs = {
+        "DM": "mDM",
+        "Log10[x]": "Log[10,x]", 
+        "dNdLog10x[eL]": "eL",
+        "dNdLog10x[eR]": "eR",
+        "dNdLog10x[e]": "e",
+        "dNdLog10x[muL]": "\\[Mu]L",
+        "dNdLog10x[muR]": "\\[Mu]R",
+        "dNdLog10x[mu]": "\\[Mu]",
+        "dNdLog10x[tauL]": "\\[Tau]L",
+        "dNdLog10x[tauR]": "\\[Tau]R",
+        "dNdLog10x[tau]": "\\[Tau]",
+        "dNdLog10x[nue]": "\\[Nu]e",
+        "dNdLog10x[numu]": "\\[Nu]\\[Mu]",
+        "dNdLog10x[nutau]": "\\[Nu]\\[Tau]",
+        "dNdLog10x[u]": "u", # Does not exist explicitly on PPPC4, but it is equivalent to q
+        "dNdLog10x[d]": "d", # Does not exist explicitly on PPPC4, but it is equivalent to q
+        "dNdLog10x[s]": "s", # Does not exist explicitly on PPPC4, but it is equivalent to q
+        "dNdLog10x[c]": "c", 
+        "dNdLog10x[b]": "b",
+        "dNdLog10x[t]": "t",
+        "dNdLog10x[a]": "\\[Gamma]",
+        "dNdLog10x[g]": "g", 
+        "dNdLog10x[W]": "W",
+        "dNdLog10x[WL]": "WL",
+        "dNdLog10x[WT]": "WT",
+        "dNdLog10x[Z]": "Z",
+        "dNdLog10x[ZL]": "ZL",
+        "dNdLog10x[ZT]": "ZT",
+        "dNdLog10x[H]": "h",
+        "dNdLog10x[aZ]": None,  # Does not exist  on PPPC4
+        "dNdLog10x[HZ]": None # Does not exist  on PPPC4
+    }
+
     tag = ["PrimaryFlux", "dm-pf"]
 
     def __init__(self, mDM, channel, source='pppc4'):
 
         if source is None:
             source='pppc4'
+            warnings.warn(
+                f"\nSince no spectra source has been chosen, PPPC4 will be used by default.\n",
+                UserWarning
+            )
 
-        if source.lower() == "pppc4" or not source:
+        if source.lower() == "pppc4":
             table_filename = "$GAMMAPY_DATA/dark_matter_spectra/AtProduction_gammas.dat"
             if channel in ('aZ','HZ'):
                 raise ValueError(
-                    f"\n\nThe channel "+channel+" is not available in PPPC4, please choose another channel or use CosmiXs (cosmixs) as source\n"
+                    f"\n\nThe channel {channel} is not available in PPPC4, please choose another channel or use CosmiXs (cosmixs) as source\n"
                 )
             elif channel in ('d','u', 's'):
                 raise ValueError(
-                    f"\n\nThe channel "+channel+" is not available in PPPC4, please choose the equivalent channel q or use CosmiXs (cosmixs) as source\n"
+                    f"\n\nThe channel {channel} is not available in PPPC4, please choose the equivalent channel q or use CosmiXs (cosmixs) as source\n"
                 )
 
         elif source == 'cosmixs':
@@ -103,7 +142,7 @@ class PrimaryFlux(TemplateNDSpectralModel):
 
             if channel in ("V->e", "V->mu", "V->tau"):
                 raise ValueError(
-                    f"\n\nThe channel "+channel+" is not available in CosmiXs, please choose another channel or use PPPC4 as source\n"
+                    f"\n\nThe channel {channel} is not available in CosmiXs, please choose another channel or use PPPC4 as source\n"
                 )
             elif channel =='q':
                 raise ValueError(
@@ -130,41 +169,7 @@ class PrimaryFlux(TemplateNDSpectralModel):
                 delimiter=" ",
             )
             if source == 'cosmixs':
-
-                mapping_dict = {
-                    "DM": "mDM",
-                    "Log10[x]": "Log[10,x]", 
-                    "dNdLog10x[eL]": "eL",
-                    "dNdLog10x[eR]": "eR",
-                    "dNdLog10x[e]": "e",
-                    "dNdLog10x[muL]": "\\[Mu]L",
-                    "dNdLog10x[muR]": "\\[Mu]R",
-                    "dNdLog10x[mu]": "\\[Mu]",
-                    "dNdLog10x[tauL]": "\\[Tau]L",
-                    "dNdLog10x[tauR]": "\\[Tau]R",
-                    "dNdLog10x[tau]": "\\[Tau]",
-                    "dNdLog10x[nue]": "\\[Nu]e",
-                    "dNdLog10x[numu]": "\\[Nu]\\[Mu]",
-                    "dNdLog10x[nutau]": "\\[Nu]\\[Tau]",
-                    "dNdLog10x[u]": "u", # Does not exist explicitly on PPPC4, but it is equivalent to q
-                    "dNdLog10x[d]": "d", # Does not exist explicitly on PPPC4, but it is equivalent to q
-                    "dNdLog10x[s]": "s", # Does not exist explicitly on PPPC4, but it is equivalent to q
-                    "dNdLog10x[c]": "c", 
-                    "dNdLog10x[b]": "b",
-                    "dNdLog10x[t]": "t",
-                    "dNdLog10x[a]": "\\[Gamma]",
-                    "dNdLog10x[g]": "g", 
-                    "dNdLog10x[W]": "W",
-                    "dNdLog10x[WL]": "WL",
-                    "dNdLog10x[WT]": "WT",
-                    "dNdLog10x[Z]": "Z",
-                    "dNdLog10x[ZL]": "ZL",
-                    "dNdLog10x[ZT]": "ZT",
-                    "dNdLog10x[H]": "h",
-                    "dNdLog10x[aZ]": None,  # Does not exist  on PPPC4
-                    "dNdLog10x[HZ]": None # Does not exist  on PPPC4
-                }
-                self.table = table_map_columns(self.table, mapping_dict)
+                self.table = table_map_columns(self.table, self.mapping_dict_PPPC4_to_CosmiXs)
 
         self.channel = channel
 
--- a/gammapy/astro/darkmatter/spectra.py
+++ b/gammapy/astro/darkmatter/spectra.py
@@ -264,7 +264,7 @@ class DarkMatterAnnihilationSpectralModel(SpectralModel):
         Type of dark matter particle (k:2 Majorana, k:4 Dirac).
 
     Examples
-    ----------
+    --------
     This is how to instantiate a `DarkMatterAnnihilationSpectralModel` model::
 
         >>> import astropy.units as u
--- a/gammapy/astro/darkmatter/spectra.py
+++ b/gammapy/astro/darkmatter/spectra.py
@@ -40,10 +40,10 @@ class PrimaryFlux(TemplateNDSpectralModel):
 
     References
     ----------
-    .. [1] `Marco et al. (2011), "PPPC 4 DM ID: a poor particle physicist cookbook for dark matter indirect detection", <https://ui.adsabs.harvard.edu/abs/2011JCAP...03..051C>`_
-    .. [2] `Cirelli et al. (2016), "PPPC 4 DM ID: A Poor Particle Physicist Cookbook for Dark Matter Indirect Detection", <http://www.marcocirelli.net/PPPC4DMID.html>`_
-    .. [3] `Arina et al. (2024), "CosmiXs: Cosmic messenger spectra for indirect dark matter searches", <https://arxiv.org/abs/2312.01153>`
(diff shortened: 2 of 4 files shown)


gammapy/astro/darkmatter/spectra.py
```diff
@@ -66,27 +73,95 @@ class PrimaryFlux(TemplateNDSpectralModel):
...
+        "s": "s"
     }
 
-    table_filename = "$GAMMAPY_DATA/dark_matter_spectra/AtProduction_gammas.dat"
-
     tag = ["PrimaryFlux", "dm-pf"]
 
-    def __init__(self, mDM, channel):
-        self.table_path = make_path(self.table_filename)
+    def __init__(self, mDM, channel, source='pppc4'):
```
AtreyeeS: Would it make sense here to optionally pass a `filename` or a `table` for future use cases where people want to include their own source models? We can provide built in support for PPPC4 and CosmiXs, but its not good to force users to go through this entire process for such a simple change.

AtreyeeS (commented):
Thanks @rcervinoucm ! The implementation looks clean to me.

This feature will need to be included in our CI testing, so please add a test in gammapy/astro/darkmatter/tests/test_spectra.py

The main question here is where the CosmiXs file should lie. It can be shipped with GAMMAPY_DATA as we do for the PPPC4 models, but it is a general question of how much we can bloat up GAMMAPY_DATA. This file is ~4.5MB, so probably it's OK to put it there for now. Opinions @registerrier ?

gammapy/astro/darkmatter/spectra.py
alexcervino: Thanks for the review. In my opinion, this would be a good option but just in the case that the files or table follow the same structure/columns as PPPC4, for example; if not, the user should implement also a dictionary with the equivalences.

AtreyeeS: We can show an example of how to do this in the tutorials

StFroese:
I am currently on my phone and can't check but maybe we need to check if the authors of CosmiXs use the same interpolation as we do. 

https://github.com/ajueid/CosmiXs/blob/main/Interpolate.py


Also, they provide uncertainties for the spectra, can we incorporate them as well?

gammapy/astro/darkmatter/spectra.py
alexcervino: Yes, good idea! Anyway I have a question, the users would like to include their own production files? I think PPC4 and CosmiXs are already very standard and stablished in the community, I mean, maybe they don't need to use another in order to make comparisons.

alexcervino:
Hi, the interpolator is not exactly the same, but I have perform some tests and the difference between methods are negligible.

StFroese:
@rcervinoucm hi, sry, should have read to the end of pr message :D

bkhelifi:
> The main question here is where the CosmiXs file should lie. It can be shipped with GAMMAPY_DATA as we do for the PPPC4 models, but it is a general question of how much we can bloat up GAMMAPY_DATA. This file is ~4.5MB, so probably it's OK to put it there for now. Opinions @registerrier ?

Yes, this is not an issue;)
When one put data on our gammapy-data, one should add a readme with all the provenance and licence information. Please have a look on the other data

alexcervino:
I have included the test for this part on gammapy/astro/darkmatter/tests/test_spectra.py.

gammapy/astro/darkmatter/spectra.py
```diff
@@ -66,27 +73,98 @@ class PrimaryFlux(TemplateNDSpectralModel):
...
+
+        if source is None:
+            source='pppc4'
```
StFroese: I think this is not necessary. Or at least there are two options:
1. Overwrite if user inputs `None` but raise a warning to let them know what they are using.
2. Do not allow None and just raise an error

gammapy/astro/darkmatter/spectra.py
```diff
@@ -66,27 +73,98 @@ class PrimaryFlux(TemplateNDSpectralModel):
...
+
+        if source.lower() == "pppc4" or not source:
```
StFroese: `or not source:`

Same as above. Maybe not necessary if you already overwrite the None case.

gammapy/astro/darkmatter/spectra.py
```diff
@@ -66,27 +73,98 @@ class PrimaryFlux(TemplateNDSpectralModel):
...
+            table_filename = "$GAMMAPY_DATA/dark_matter_spectra/AtProduction_gammas.dat"
+            if channel in ('aZ','HZ'):
+                raise ValueError(
+                    f"\n\nThe channel "+channel+" is not available in PPPC4, please choose another channel or use CosmiXs (cosmixs) as source\n"
+                )
+            elif channel in ('d','u', 's'):
+                raise ValueError(
+                    f"\n\nThe channel "+channel+" is not available in PPPC4, please choose the equivalent channel q or use CosmiXs (cosmixs) as source\n"
+                )
```
StFroese: I think this can be one raise for all particles available in cosmics but not in pppc4.
Also:
```diff python
- f"\n\nThe channel "+channel+" is not available in PPPC4, please choose another channel or use CosmiXs (cosmixs) as source\n"
+ f"\n\nThe channel {channel} is not available in PPPC4, please choose another channel or use CosmiXs (cosmixs) as source\n"
```

gammapy/astro/darkmatter/spectra.py
```diff
@@ -66,27 +73,98 @@ class PrimaryFlux(TemplateNDSpectralModel):
...
+            table_filename = 'gammapy/external/CosmiXs/Data/AtProduction-Gamma.dat'
+
+            if channel in ("V->e", "V->mu", "V->tau"):
+                raise ValueError(
+                    f"\n\nThe channel "+channel+" is not available in CosmiXs, please choose another channel or use PPPC4 as source\n"
+                )
+            elif channel =='q':
+                raise ValueError(
+                    f"\n\nThe channel q is not available in cosmixs, please choose an equivalent channel such as d, u or s or use PPPC4 as source\n"
+                )
```
StFroese: Maybe it's worth just printing the available particles of either model in the error. 
This requires splitting the `channel_registry` into to dicts.

gammapy/astro/darkmatter/spectra.py
```diff
@@ -66,27 +73,98 @@ class PrimaryFlux(TemplateNDSpectralModel):
...
             self.table = Table.read(
                 str(self.table_path),
-                format="ascii.fast_basic",
+                format=ascii_format,
                 guess=False,
                 delimiter=" ",
             )
+            if source == 'cosmixs':
+
+                mapping_dict = {
```
StFroese: this could be global at top of file similar to `channel_registry`

gammapy/astro/darkmatter/tests/test_spectra.py
```diff
@@ -23,11 +23,11 @@ def test_primary_flux():
 
 
 @pytest.mark.parametrize(
-    "mass, expected_flux", [(1.6, 0.00025037), (11, 0.00502079), (75, 0.02028309)]
+    "mass, expected_flux, source", [(1.6, 0.00025037,'pppc4'), (11, 0.00549445,'cosmixs'), (75, 0.02028309,None)]
```
StFroese: `None` defaults to pppc4, see comment in spectra.py. Maybe one rather wants to test a `source="nonexistend"` with xfail

StFroese:
Just some minor comments :) Nice PR!

gammapy/astro/darkmatter/spectra.py
alexcervino: This channel validation is already donde in 
    @channel.setter
    def channel(self, channel):
        if channel not in self.allowed_channels:
            raise ValueError(
                f"Invalid channel: {channel}\nAvailable: {self.allowed_channels}\n"
            )
        else:
            self._channel = channel

gammapy/astro/darkmatter/spectra.py
alexcervino: Done! I have implemented the first option.

gammapy/astro/darkmatter/spectra.py
alexcervino: Done!

gammapy/astro/darkmatter/spectra.py
alexcervino: I think maybe is correct to leave the raise for the two datsets, since there are channels that are in PPPC4 and not in CosmiXs and viceversa, and maybe some user want to use them.

gammapy/astro/darkmatter/spectra.py
alexcervino: Done!

gammapy/astro/darkmatter/tests/test_spectra.py
alexcervino: Done!

alexcervino:
Thanks for the comments! I have made the corrections :)

registerrier (commented):
Thanks @rcervinoucm . This is really nice. 

First I have a general set of comments:
- We need to understand where/how to store the CosmiXs table. 
  - Is it really fine to simply copy the table from the CosmiXs GH repository? As far as I can see there is no license there. We should check that this option suits the authors. If not there might be other options.
  - If it is OK, it could be stored in `gammapy_data` with proper license and credit.
- Wouldn't it be simpler to have specialized `PrimaryFlux` classes for the PPPC4 and CosmiXs tables?
- The `to_dict` methods should be updated with the required input. But is the serialization of `DarkMatterAnnihilationSpectralModel` working? 
- Last, it would be useful that we have some validation for physical models that are used in gammapy. This is important to make sure we stay consistent with the proper references (see e.g. https://github.com/gammapy/gammapy-benchmarks/issues/174 for PCCC4). At some point it would be good to have one for CosmiXs

alexcervino:
Thanks for the comments!

- I am not sure what of you mean by specialising the PrimaryFlux class, but my intention was to make the less changes as possible in the classes structure, since the changes were mainly about the correlation between both datasets classes.
- I have updated the to_dict() methods, thanks for noticing. In principle the serialization is working, the tests have passed.

Thanks!

alexcervino:
I have modified the code to get the ComiXs file from the new path. I have opened a PR in gammapy-data with the file inclusion https://github.com/gammapy/gammapy-data/pull/80. 
Thanks!

gammapy/astro/darkmatter/spectra.py
StFroese: I see, I think maybe you should move this additional check for channel validity to this setter, otherwise it is only checked on init and not when the user switches the channel

alexcervino: Very good point, I have made the changes. Thanks for the review

AtreyeeS:
It will be good to merge this PR soon. Can you please fix the merge conflict @rcervinoucm ?

docs/release-notes/6232.feature.rst
```diff
@@ -0,0 +1 @@
+For the Dark Matter module: Implementation of the possibility of using CosmiXs or PPPC4 as a source for creating Dark Matter spectra.
```
AtreyeeS:
```suggestion
Implementation of the option of using CosmiXs or PPPC4 as a source models for creating Dark Matter spectra.
```

AtreyeeS (dismissed):
Thanks @rcervinoucm ! No more comments

AtreyeeS (dismissed):

docs/release-notes/6232.feature.rst
```diff
@@ -0,0 +1 @@
+Implementation of the option of using CosmiXs or PPPC4 as a source models for creating Dark Matter spectra.
```
Astro-Kirsty:
```suggestion
Implementation of the option of using ``CosmiXs`` or ``PPPC4`` as a source models for creating dark matter spectra.
```

gammapy/astro/darkmatter/spectra.py
```diff
@@ -8,33 +8,46 @@
...
 class PrimaryFlux(TemplateNDSpectralModel):
     """DM-annihilation gamma-ray spectra.
 
-    Based on the precomputed models by Cirelli et al. (2016). All available
-    annihilation channels can be found there. The dark matter mass will be set
-    to the nearest available value. The spectra will be available as
-    `~gammapy.modeling.models.TemplateNDSpectralModel` for a chosen dark matter mass and
-    annihilation channel. Using a `~gammapy.modeling.models.TemplateNDSpectralModel`
-    allows the interpolation between different dark matter masses.
+        Based on the precomputed models of PPPC4 DM ID by Cirelli et al. (2016), CosmiXs by Arina et al. (2024); Di Mauro et al. (2025). All available
```
Astro-Kirsty: These lines do not have the correct indentation. 
You have the ability to use numbers for the references, for usage see the documentation:
https://docs.gammapy.org/2.0.1/development/doc_howto.html#code-in-docstrings-in-python-files

gammapy/astro/darkmatter/spectra.py
```diff
@@ -8,33 +8,46 @@
...
-    channel: str
-        Annihilation channel. List available channels with `~gammapy.spectrum.PrimaryFlux.allowed_channels`.
-
-    References
-    ----------
-    * `Marco et al. (2011), "PPPC 4 DM ID: a poor particle physicist cookbook for dark matter indirect detection"
-      <https://ui.adsabs.harvard.edu/abs/2011JCAP...03..051C>`_
-    * `Cirelli et al. (2016), "PPPC 4 DM ID: A Poor Particle Physicist Cookbook for Dark Matter Indirect Detection"
-      <http://www.marcocirelli.net/PPPC4DMID.html>`_
+        Parameters
```
Astro-Kirsty: Incorrect indentation again

gammapy/astro/darkmatter/spectra.py
```diff
@@ -190,6 +289,9 @@ class DarkMatterAnnihilationSpectralModel(SpectralModel):
         Redshift value.
     k: int
         Type of dark matter particle (k:2 Majorana, k:4 Dirac).
+    source: str
```
Astro-Kirsty:
```suggestion
    source: str, optional
```

gammapy/astro/darkmatter/spectra.py
```diff
@@ -298,6 +404,9 @@ class DarkMatterDecaySpectralModel(SpectralModel):
         is used.
     z: float
         Redshift value.
+    source: str
```
Astro-Kirsty:
```suggestion
    source: str, optional
```

We should add the options here, like we have else where i.e. `methods` here
https://docs.gammapy.org/2.0.1/api/gammapy.makers.SafeMaskMaker.html#gammapy.makers.SafeMaskMaker

gammapy/astro/darkmatter/spectra.py
Astro-Kirsty: See comment below also

gammapy/astro/darkmatter/spectra.py
```diff
@@ -190,6 +289,9 @@ class DarkMatterAnnihilationSpectralModel(SpectralModel):
         Redshift value.
     k: int
         Type of dark matter particle (k:2 Majorana, k:4 Dirac).
+    source: str
+        Data source for the spectra. Choose between 'pppc4' and 'cosmixs'.
+        Delfault is 'pppc4'.
```
Astro-Kirsty:
```suggestion
        Default is 'pppc4'.
```

gammapy/astro/darkmatter/spectra.py
```diff
@@ -298,6 +404,9 @@ class DarkMatterDecaySpectralModel(SpectralModel):
         is used.
     z: float
         Redshift value.
+    source: str
+        Data source for the spectra. Choose between 'pppc4' and 'cosmixs'.
+        Delfault is 'pppc4'.
```
Astro-Kirsty:
```suggestion
        Default is 'pppc4'.
```

gammapy/astro/darkmatter/spectra.py
```diff
@@ -298,6 +404,9 @@ class DarkMatterDecaySpectralModel(SpectralModel):
...
 
     Examples
```
Astro-Kirsty, line 404: Should this section be updated now?

Astro-Kirsty (changes requested):
Minor comments on the documentation, thanks @rcervinoucm!

micheledoro:
@rcervinoucm thanks for this.
1. We checked the cosmixs table in https://linkinghub.elsevier.com/retrieve/pii/S2212686425003474. We found (besides some differences in the columns, there is actually one more column in cosmixs than Cirelli) that cosmixs does not compute the yield _at_ the DM mass while PPPC does it. This is bad because the last bin is 'empty' on cosmixs and for some computations creates a difference. Did you check that?
2. I believe PPPC and cosmiXs files are just two of the possible. They should be readable online, or the user should be invited to download them by asking explicitly a path, they should not be given with `GAMMAPY_DATA`.
3. Was this pushed? Can this issue be closed?

alexcervino:
Hello @micheledoro 

(1) No, indeed we had no clue about this, but we should discuss and analyse it. We have created this issue for this purpose #6725 
(2) This should be discuss with the Gammapy main developers, since it was already implemented to get the data from GAMMAPY_DATA and I don't know if it is a requirement or not. @AtreyeeS @bkhelifi can you please clarify this point?
(3) Yes, this issue can be closed, it is already merged and released

AtreyeeS:
@micheledoro @rcervinoucm You raise an important point regarding GAMMAPY_DATA
So far, for whatever "external" files we support, we have been shipping them with GAMMAPY_DATA - EBL models, source catalogs or DM models. This helps us with the versioning and the tests, and ensures we dont break compatibility at any point

This is becoming an issue in general - GAMMAPY_DATA is getting bulky, maintaining the latest models, etc 

As you say, PPPC and cosmiXs files are just two of the possible - it can get difficult (and probably pointless) to ship everything together.

At this point, I would suggest we have a tutorial/some documentation about how to add other models in Gammapy (what are the points to consider, what should be the file format, how to choose specific channels, etc)

micheledoro:
Hi @AtreyeeS I support
* removing altogether PPPC from `GAMMAPY-DATA`
* having tutorial/some documentation about how to add models in Gammapy. Being external inputs, I think it's also less prone to automatisms/biases.

alexcervino:
Hello, if we remove the data, the other option is that the user must provide the PPPC4/CosmiXs  source file everytime from their own file system or that the code must be able to retrieve from the official source?

micheledoro:
In both cases pppc and cosmixs have stable online accessible pages

alexcervino:
I know, but the constraint I see there is that we would depend on them (if they do any maintenance, they decide to remove the file, etc) for the code execution, for now since the data is in GAMMAPY-DATA we don't have that dependance and the code will execute successfully in any case.

micheledoro:
If they do maintenance we can always link the file. I think this is more correct than having a single option available  so that it becomes a standard and people use it 'without check'. I mean unless as consortium we decide we prefer this to another model. I hope I explained myself.

alexcervino:
Understood, I see your point, thanks a lot for explaining. Anyway , I think this is more a Gammapy management decision to approve, since they are very careful about external dependance and so. 

I will include something about this in the tutorial we are working on.

Issue #6246, opened 2025-11-30
Error in creation of all-sky template model with TemplateSpatialModel based on HpxNDMap in SkyModel


dima137:
**Gammapy version**
2.0 (currently the latest released version)

**Bug description**
When I try to define an all-sky template model based on a healpix map, I get an error while executing SkyModel(spectral_model=spectrum, spatial_model=template)
There is no error for a WCS spatial template.

**To Reproduce**
Snippet of the problem code:

from gammapy.maps import HpxGeom, HpxNDMap, WcsGeom, WcsNDMap
from gammapy.modeling.models import TemplateSpatialModel, PowerLawSpectralModel, SkyModel

nside_test = 32
geom_test = HpxGeom(nside_test, frame='galactic', nest=False)
ndmap = HpxNDMap(geom_test, 1., unit="")
template = TemplateSpatialModel(ndmap, normalize=True, filename='')
spectrum = PowerLawSpectralModel(index=2.7, amplitude="3.e-6 cm-2 s-1 GeV-1", reference="1 GeV")
mdl = SkyModel(spectral_model=spectrum, spatial_model=template)

The last command causes the error ending with:
File .../anaconda3/envs/gammapy-2.0/lib/python3.12/site-packages/gammapy/modeling/models/spatial.py:1484, in TemplateSpatialModel.evaluate(self, lon, lat, energy, lon_0, lat_0)
   1481 if energy is not None:
   1482     coord["energy_true"] = energy
-> 1484 val = self.map.interp_by_coord(coord, **self._interp_kwargs)
   1485 val = np.clip(val, 0, a_max=None)
   1486 return u.Quantity(val, self.map.unit, copy=COPY_IF_NEEDED)

TypeError: HpxNDMap.interp_by_coord() got an unexpected keyword argument 'values_scale'

The following WCS version works fine:

geom = WcsGeom.create(skydir=ndmap.geom.center_skydir, npix=40, binsz=0.5, frame='galactic', axes=[])
wcsndmap = WcsNDMap(geom, 1., unit='')
template_wcs = TemplateSpatialModel(wcsndmap, normalize=True, filename='')
mdl = SkyModel(spectral_model=spectrum, spatial_model=template_wcs)

The only difference is that the healpix template is now replaced with a WCS template.

PR #6248, opened 2025-12-01
Add check for map type in TemplateSpatialModel

registerrier:
**Description**

This pull request solves issue #6246 . It changes default interpolation arguments in the case of `HpxNDMap`.

Possible associated questions are what to do with `TemplateNDSpatialModel` and what if the map has an energy axis?

**Dear reviewer**

### DIFF
--- a/gammapy/modeling/models/spatial.py
+++ b/gammapy/modeling/models/spatial.py
@@ -1313,7 +1313,7 @@ class TemplateSpatialModel(SpatialModel):
     interp_kwargs : dict
         Interpolation keyword arguments passed to `gammapy.maps.Map.interp_by_coord`.
         Default arguments are {'method': 'linear', 'fill_value': 0, "values_scale": "log"}.
-        For a `~gammapy.maps.HpxNDMap`, default arguments are {'method': 'linear', 'fill_value': 0}.
+        For a `~gammapy.maps.HpxNDMap`, default arguments are {'method': 'linear'}.
     filename : str
         Name of the map file.
     copy_data : bool


gammapy/modeling/models/spatial.py
```diff
@@ -1388,7 +1388,9 @@ def __init__(
         interp_kwargs = {} if interp_kwargs is None else interp_kwargs
         interp_kwargs.setdefault("method", "linear")
         interp_kwargs.setdefault("fill_value", 0)
-        interp_kwargs.setdefault("values_scale", "log")
+
+        if isinstance(self.map, WcsNDMap):
+            interp_kwargs.setdefault("values_scale", "log")
```
bkhelifi, line 1394: The docstring must be updated

gammapy/modeling/models/spatial.py
```diff
@@ -1313,6 +1313,7 @@ class TemplateSpatialModel(SpatialModel):
     interp_kwargs : dict
         Interpolation keyword arguments passed to `gammapy.maps.Map.interp_by_coord`.
         Default arguments are {'method': 'linear', 'fill_value': 0, "values_scale": "log"}.
+        For a `~gammapy.maps.HpxNDMap`, default arguments are {'method': 'linear', 'fill_value': 0}.
```
QRemy: could remove  `'fill_value': 0` because it is not used in `interp_by_coord` even if defined

registerrier:
> Thanks @registerrier, maybe the solution would be to implement the `values_scales` and `fill_values` options because for now `values_scales` is not present in `interp_by_coord` arguments, while `fill_values` is present but not used.
> 
> Alternatively a simple fix could be to add the `values_scales` argument on `interp_by_coord` even if not yet used (as `fill_values`), this way you don't need to modify the TemplateSpatialModel and the API is more consistent.

Indeed, we should ideally harmonize the interpolation scheme between `HpxNDMap` and `WcsNDMap`.  But can we completely do it? Currently even `method` is not defined on `HpxNDMap`. It supports only `"linear"`. The interp_kwargs are valid only for the `WcsNDMap`.

gammapy/modeling/models/spatial.py
registerrier:
```suggestion
        For a `~gammapy.maps.HpxNDMap`, default arguments are {'method': 'linear'}.
```

QRemy (approved):
Thanks @registerrier, no further comment.

PR #6251, opened 2025-12-01
Resolves issues associated to plot_error

### LINKED ISSUE #6195: Error from CompoundSpectralModel.evaluate() args for TemplateNDSpectralModel.
**Gammapy version 2.0**

**Bug description**
CompoundSpectralModel.evaluate() turns parameter value into array of values when evaluating model2, causes errors for model2 that is a TemplateNDSpectralModel. In example, errors with plot_error.

**Expected behavior**
Expect to be able to plot_error for CompoundSpectralModel. With gammapy 1.3 CompoundSpectralModel.evaluate() gives args2 as a single parameter for the below example, which is expected and works, but with gammapy 2.0 args2 is given as an array of values. If the array of values change is intentional TemplateNDSpectralModel needs to be updated to accept that.

**To Reproduce**

```python
import astropy.units as u
import numpy as np
from gammapy.modeling import Parameter
from gammapy.modeling.models import (
    Models,
    SkyModel,
    TemplateNDSpectralModel,
    create_crab_spectral_model,
)
from gammapy.maps import MapAxis, RegionNDMap

edges = np.array([0.030,  0.100, 0.284, 0.400, 0.532, 0.707, 0.867, 1.303, 1.840, 2.471, 3.210, 4.038, 7.111, 8.331, 10, 11])
c2=np.array([0.030,  0.100, 0.284, 0.400, 0.532, 0.707, 0.867, 1.303, 1.840, 2.471, 3.210, 4.038, 7.111, 8.331, 10, 11])
energy_axis = MapAxis(nodes=10**edges,node_type='center',interp='log',name='energy_true',unit='keV')
nh_axis     = MapAxis(nodes=10**c2*1e22,node_type='center',interp='log',name='nH',unit='cm-2')
region_ndmap = RegionNDMap.create(
        region = None,
        axes = [energy_axis,nh_axis],
    )

temp = TemplateNDSpectralModel(
        map = region_ndmap, 
        interp_kwargs={'method': 'linear', 'fill_value': 1e-5}
    )

param = Parameter('nH', value=2.0699999999999997e+21, unit="1 / cm2")
temp.nH.quantity = param.value * param.unit

source_SkyModel = SkyModel(name='crab', spectral_model=create_crab_spectral_model(reference='magic_lp'))
models = Models([source_SkyModel])

models['crab'].spectral_model *= temp

fig = plt.figure(figsize=(5, 5),dpi=100)

ax = models['crab'].spectral_model.plot_error(
    [energy_axis.edges[0],energy_axis.edges[-1]],
    sed_type='e2dnde')
```

**Other information**

### LINKED ISSUE #6204: plot_error fails with TemplateNDSpectralModel
**Gammapy version**
2.0 and dev

**Bug description**
`TemplateNDSpectralModel.plot_error()` fails. This is because `plot_error` will call `spectral_model.evaluate(energy[..., np.newaxis], *args)` while `TemplateNDSpectralModel.evaluate` expects `energy` and `**kwargs`. 

**Expected behavior**
`plot_error` should work.

**To Reproduce**
```
import astropy.units as u
from gammapy.maps import MapAxis, RegionNDMap
from gammapy.modeling.models import TemplateNDSpectralModel
energy_axis = MapAxis.from_bounds(
    1.0, 100, 10, interp="log", name="energy_true", unit="GeV"
)
norm = MapAxis.from_bounds(0, 10, 10, interp="lin", name="norm", unit="")
tilt = MapAxis.from_bounds(-1.0, 1, 5, interp="lin", name="tilt", unit="")
region_map = RegionNDMap.create(
    region="icrs;point(83.63, 22.01)", axes=[energy_axis, norm, tilt], unit="TeV-1cm-2s-1"
)
region_map.data[:, :, :5, 0, 0] = 1
region_map.data[:, :, 5:, 0, 0] = 2

template = TemplateNDSpectralModel(region_map)

template.plot([1., 100.]*u.GeV) # works

template.tilt.error=0.1
template.norm.error=0.1
template.plot_error([1., 100.]*u.GeV) # fails
```

QRemy:
Resolves issues associated to plot_error with the new sampling-based implementation.
Resolves #6204 BrokenPowerLawSpectralModel, TemplateNDSpectralModel

Resolves #6195 TemplateNDSpectralModel
This second issue mention a problem with unit. Now the template ND model can be plotted if is map has no unit because it will be considered as a norm model. Otherwise it will fails wait an astropy unit conversion error if the map is not in an sed compatible unit.

### DIFF
--- a/gammapy/modeling/models/tests/test_spectral.py
+++ b/gammapy/modeling/models/tests/test_spectral.py
@@ -1311,7 +1311,7 @@ def test_template_ND(tmpdir, caplog):
     assert template.parameters["tilt"].value == 0
     assert_allclose(template([1, 100, 1000] * u.GeV), [1.0, 2.0, 2.0])
 
-    assert template.is_norm_spectral_model == True
+    assert template.is_norm_spectral_model
 
     template.parameters["norm"].value = 1
     template.filename = str(tmpdir / "template_ND.fits")


gammapy/modeling/models/spectral.py
```diff
@@ -756,6 +756,11 @@ def plot_error(
         """
         from gammapy.estimators.map.core import DEFAULT_UNIT
 
+        if len(self.parameters) == 0:
+            raise NotImplementedError(
+                "plot_error is not defined for models without paraemters."
```
bkhelifi:
```suggestion
                "plot_error is not defined for models without parameter."
```

bkhelifi:
The issue #6195 mentions in addition "Expect to be able to plot_error for CompoundSpectralModel". However, I do not see an associated new test for that. Is it expected? It seems that the previous tests did not catch this issue

QRemy:
> The issue #6195 mentions in addition "Expect to be able to plot_error for CompoundSpectralModel". However, I do not see an associated new test for that. Is it expected? It seems that the previous tests did not catch this issue

The CompoundSpectralModel case is covered by  `test_plot_error` the error in the issue appears to be associated to TemplateNDSpectralModel (there is also something weird with units in the example given in the issue but it is unrelated).

gammapy/astro/darkmatter/spectra.py
```diff
@@ -150,14 +150,15 @@ def channel(self, channel):
         else:
             self._channel = channel
 
-    def evaluate(self, energy, **kwargs):
+    def evaluate(self, energy, *args):
```
bkhelifi, line 153: A docstring might be great, no?

gammapy/modeling/models/spectral.py
```diff
@@ -756,6 +756,11 @@ def plot_error(
         """
         from gammapy.estimators.map.core import DEFAULT_UNIT
 
+        if len(self.parameters) == 0:
+            raise NotImplementedError(
+                "plot_error is not defined for models without paraemter."
```
bkhelifi:
```suggestion
                "plot_error is not defined for models without parameter."
```

bkhelifi:
Can you try to improve the code coverage? >95%

gammapy/astro/darkmatter/spectra.py
QRemy: I would say another PR

registerrier: In general, the change we might need to clarify is that `evaluate` function arguments have to be passed in the order of the parameter definition, for instance for custom models.

gammapy/astro/darkmatter/spectra.py
```diff
@@ -150,14 +150,15 @@ def channel(self, channel):
...
         """Evaluate the primary flux."""
-        mass = {"mass": self.mDM}
-        kwargs.update(mass)
+
+        args = list(args)
```
registerrier, line 156: Should we check that there is only one element in the list?

gammapy/astro/darkmatter/spectra.py
```diff
@@ -150,14 +150,15 @@ def channel(self, channel):
...
+        args.append(self.mDM)
```
registerrier, line 157: Should we check that there is only one element in the list?

gammapy/modeling/models/spectral.py
```diff
@@ -2247,12 +2261,18 @@ def __init__(self, map, interp_kwargs=None, meta=None, filename=None):
         self._interp_kwargs = interp_kwargs
         super().__init__()
 
+    @property
```
registerrier, line 2264: Why the change needed?

gammapy/modeling/models/spectral.py
```diff
@@ -2247,12 +2261,18 @@ def __init__(self, map, interp_kwargs=None, meta=None, filename=None):
...
+        return self._map.unit == u.Unit("")
+
     @property
     def map(self):
         """Template map as a `~gammapy.maps.RegionNDMap`."""
         return self._map
 
-    def evaluate(self, energy, **kwargs):
+    def evaluate(self, energy, *args):
+        kwargs = {name: q for name, q in zip(self.default_parameters.names, args)}
```
registerrier, line 2274: What will happen if `args` is not of the correct size?

Astro-Kirsty:
Thanks @QRemy.
I tested the example code [here](https://github.com/gammapy/gammapy/issues/6204)  and [here](https://github.com/gammapy/gammapy/issues/6195), and although it no longer fails, the plot is blank. Should that really be expected?
image

gammapy/modeling/models/spectral.py
QRemy: The default is:
```
    @classproperty
    def is_norm_spectral_model(cls):
        """Whether model is a norm spectral model."""
        return "Norm" in cls.__name__
```
This does not work for the TemplateNDSpectralModel even if in some case it could be used as a norm model.

gammapy/modeling/models/spectral.py
QRemy: the standard message x args provided instead of n args expected.

gammapy/astro/darkmatter/spectra.py
QRemy: here I use list to convert from a tuple but the number of arguments is correct otherwise it would fail directly passing the arguments

QRemy:
> Thanks @QRemy. I tested the example code [here](https://github.com/gammapy/gammapy/issues/6204) and [here](https://github.com/gammapy/gammapy/issues/6195), and although it no longer fails, the plot is blank. Should that really be expected? 

In the first example It's because the data are constant along a given parameter so the variance is 0 
if you define `region_map.data = np.random.rand(*region_map.data.shape)` the error region will be visible.

In the second example the covariance is empty.

PR #6252, opened 2025-12-01
Fix plot_error for models evaluated with not finite values

### LINKED ISSUE #6228: `plot_error` method of `SpectralModel` can fail by sampling specific parameter values
Trying to "plot_error" this model fails:

```python
ecpl = ExpCutoffPowerLawSpectralModel(
    index=2.0,
    reference=1*u.TeV,
    amplitude="2.2e-13 TeV-1 s-1 cm-2",
    lambda_ = 0.09/u.TeV,
    alpha=5.1,
)
ecpl.index.error=0.35
ecpl.amplitude.error=0.7e-13
ecpl.alpha.error=3
ecpl.lambda_.error=0.045
```
This is because in the new propagation of error scheme, samples are drawn from the model parameter and covariance matrix. In the specific model above, some samples have negative `lambda_` value and non-integer `alpha` value which result in NaN (see #6229). Drawing negative `lambda_` values is not suprising as it is only ~2 std from the mean value, and it is natural to have non-integer `alpha` if it is left free during the fitting procedure.

Plotting errors in an analysis, one would want to use the samples drawn by e.g. UltraNest or the covariance matrix after a fit, which would take into account potential prior set on parameters. But it is not unusual to take model parameters and associated errors from say a paper (thus without covariance matrix) to compare.

QRemy:
Fix plot_error for models evaluated with not finite values
Resolves #6228

### DIFF
--- a/gammapy/modeling/models/tests/test_spectral.py
+++ b/gammapy/modeling/models/tests/test_spectral.py
@@ -1467,8 +1467,8 @@ def test_plot_error_invalid():
         ecpl.plot_error([1, 100] * u.TeV, energy_power=2)
         plt.ylim(1e-15, 1e-11)
         plt.show()
-        
-        
+
+
 def test_bpl_evalaute_array():
     model = BrokenPowerLawSpectralModel(
         index1=1.5 * u.Unit(""),


Astro-Kirsty:
Thanks @QRemy. Could you post an image which shows this one versus what we currently have implemented?

QRemy:
The new test gives this:
Screenshot 2025-12-02 at 09 45 48
while it returns an error in the current version.

gammapy/modeling/models/spectral.py
```diff
@@ -268,11 +268,12 @@ def _get_errors(self, samples, n_sigma=1):
             Median, negative, and positive errors
 
         """
-        cdf = stats.norm.cdf
 
-        median = np.percentile(samples, 50, axis=-1)
-        errn = median - np.percentile(samples, 100 * cdf(-n_sigma), axis=-1)
-        errp = np.percentile(samples, 100 * cdf(n_sigma), axis=-1) - median
+        samples[~np.isfinite(samples)] = np.nan
```
AtreyeeS, line 272: we can keep only the finite samples instead of using the np.nanxx ?

registerrier (dismissed):
Thanks @QRemy . Looks good.

PR #6253, opened 2025-12-02
Fix Models and DatasetModels addition for any order 

QRemy:
Fix the addition between Models and DatasetModels such as it works regardless of the order they are passed.


bkhelifi (approved):
Well seen, thanks @QRemy

docs/release-notes/6253.bug.rst
```diff
@@ -0,0 +1 @@
+Fix the addition between Models and DatasetModels such as it works regardless of the order they are given.
```
Astro-Kirsty:
```suggestion
Fix the addition logic so that `gammapy.modeling.models.Models` and `gammapy.modeling.models.DatasetModels`  can be added together regardless of the order they are given.
```

registerrier (approved):
Thanks @QRemy !

PR #6254, opened 2025-12-02
Fix map stacking for non float type

QRemy:
Fix map stacking if data type is not float.
Change a line in WcsNDMap  and HpxNDMap to match what is done in RegionNDMap.

### DIFF
--- a/gammapy/maps/hpx/tests/test_ndmap.py
+++ b/gammapy/maps/hpx/tests/test_ndmap.py
@@ -634,3 +634,4 @@ def test_stack_int():
     map2 = map1.copy()
 
     map1.stack(map2)
+    assert map1.data.dtype == int
--- a/gammapy/maps/wcs/tests/test_ndmap.py
+++ b/gammapy/maps/wcs/tests/test_ndmap.py
@@ -1132,3 +1132,4 @@ def test_stack_int():
     map2 = map1.copy()
 
     map1.stack(map2)
+    assert map1.data.dtype == int


gammapy/maps/hpx/tests/test_ndmap.py
```diff
@@ -625,3 +625,12 @@ def test_hpx_map_sampling():
...
         hpxmap.sample_coord(2)
+
+
+def test_stack_int():
+    geom = HpxGeom.create(nside=16)
+
+    map1 = Map.from_geom(geom=geom, dtype=int)
+    map2 = map1.copy()
+
+    map1.stack(map2)
```
registerrier, line 636: I would add an assert here

gammapy/maps/wcs/tests/test_ndmap.py
```diff
@@ -1120,3 +1120,15 @@ def test_histogram_center_value():
...
+def test_stack_int():
+    geom = WcsGeom.create(
+        binsz=0.02,
+        width=(2, 2),
+        frame="icrs",
+    )
+    map1 = Map.from_geom(geom=geom, dtype=int)
+    map2 = map1.copy()
+
+    map1.stack(map2)
```
registerrier, line 1134: Add an assert here.

gammapy/maps/wcs/ndmap.py
```diff
@@ -1108,7 +1108,7 @@ def stack(self, other, weights=None, nan_to_num=True):
                 "Can only stack equivalent maps or cutout of the same map."
             )
 
-        data = other.quantity[cutout_slices].to_value(self.unit)
+        data = other.quantity[cutout_slices].to_value(self.unit).astype(self.data.dtype)
```
registerrier, line 1111: what happens if you stack a float map into an int map?

QRemy: it will truncate float to integer. Should we round instead ? or just send a warning ?

gammapy/maps/hpx/tests/test_ndmap.py
QRemy, line 636:
```suggestion
    map1.stack(map2)
    assert map1.data.dtype == int
```

gammapy/maps/wcs/tests/test_ndmap.py
QRemy, line 1134:
```suggestion
    map1.stack(map2)
    assert map1.data.dtype == int
```

registerrier (approved):
Thanks @QRemy. Looks good to me!

PR #6255, opened 2025-12-03
Add sample_parameters_from_covariance method on models.

QRemy:
Add sample_parameters_from_covariance method on models.
Allows  create parameters samples from covariance using multivariate normal distribution. Cleanup spectral error propagation code to re-use this method.

### DIFF
--- a/gammapy/utils/scripts.py
+++ b/gammapy/utils/scripts.py
@@ -369,7 +369,8 @@ def logic_parser(table, expression):
 
 
 def method_wrapper(func):
-    """Wrap a function for use as a method while preserving its metadata.
+    """
+    Wrap a function for use as a method while preserving its metadata.
 
     This utility wraps a function so it can be assigned as a method on other
     classes (or instances) and still calls the original function with the
@@ -378,6 +379,7 @@ def method_wrapper(func):
     which makes the wrapped method appear
     in introspection and documentation like the original.
     """
+
     def wrapper(self, *args, **kwargs):
         return func(self, *args, **kwargs)
 


gammapy/modeling/covariance.py
```diff
@@ -243,3 +244,35 @@ def covariance(self, covariance):
...
+
+        Parameters
+        ----------
+        n_samples : int, optional
+            Number of samples to generate. Default is 1000.
+        random_state : {int, 'random-seed', 'global-rng', `~numpy.random.RandomState`}, optional
+            Defines random number generator initialisation.
+            Passed to `~gammapy.utils.random.get_random_state`. Default is 42.
+        free_only : bool, optional
+            If True sample only free parameters (default).
```
bkhelifi:
```suggestion
            If True, sample only free parameters (default).
```

gammapy/utils/scripts.py
```diff
@@ -365,3 +366,11 @@ def eval_node(node):
     expr_ast = ast.parse(expression, mode="eval")
     mask = eval_node(expr_ast.body)
     return table[mask]
+
+
+def method_wrapper(func):
```
bkhelifi, line 371: Can you add a docstring to explain the purpose, the input and output?

gammapy/modeling/models/core.py
```diff
@@ -12,7 +12,13 @@
 from gammapy.maps import Map, RegionGeom
 from gammapy.modeling import Covariance, Parameter, Parameters
 from gammapy.modeling.covariance import CovarianceMixin
-from gammapy.utils.scripts import from_yaml, make_path, to_yaml, write_yaml
+from gammapy.utils.scripts import (
+    from_yaml,
+    make_path,
+    to_yaml,
+    write_yaml,
```
bkhelifi, line 19: Why are they necessary?

gammapy/modeling/models/core.py
```diff
@@ -287,6 +293,10 @@ def covariance(self, covariance):
             variance = self._covariance.get_subcovariance(pars).data
             par.error = np.sqrt(variance[0][0])
 
+    sample_parameters_from_covariance = method_wrapper(
```
bkhelifi, line 296: Why a wrapper is needed? I have not catched this subtlety...

gammapy/modeling/covariance.py
```diff
@@ -243,3 +244,35 @@ def covariance(self, covariance):
...
+
+        Returns
+        -------
+        sed_samples : np.array
```
facero: why is it named sed_samples ? 
If I understand correctly these are samples of model parameters ? 
I see that name is only in the docstring but could be named `param_samples`

facero:
```suggestion
        param_samples : np.array
```

facero:
Thanks !
This looks nice as it is. Just a small comment noted above and a general question.
Have you tested that this works equally well with naima models.
It should but it would be interesting to test it

gammapy/modeling/models/core.py
QRemy: The wrapper allows to reuse the method from another class importing also the docstring.
Alternatively I could have used a Mixin, but the CovarianceMixin would conflict with the some of the Model,  and I don't want to introduce another Mixin class for a single method that's why I wrapped it instead.

gammapy/modeling/models/core.py
QRemy: just a changed of formatting  by black because the line limit was reached adding the method_wrapper

gammapy/utils/scripts.py
QRemy, line 371:
```suggestion
def method_wrapper(func):
    """Wrap a function for use as a method while preserving its metadata.

    This utility wraps a function so it can be assigned as a method on other
    classes (or instances) and still calls the original function with the
    receiving object as the first argument. The wrapper copies the original
    function's metadata (e.g. docstring, name, module, annotations),
    which makes the wrapped method appear
    in introspection and documentation like the original.
    """
```

gammapy/modeling/covariance.py
QRemy: yes these are samples of model parameters,  too much copy paste

QRemy:
Comments are implemented, CI failures are unrelated, merging now.

PR #6256, opened 2025-12-03
Adds `Datasets` addition functionality

QRemy:
Enables the addition of `Datasets` objects with each other and with individual datasets.

### DIFF
--- a/gammapy/datasets/tests/test_datasets.py
+++ b/gammapy/datasets/tests/test_datasets.py
@@ -229,3 +229,10 @@ def test_add_datasets(datasets):
     new_datasets = datasets[0] + datasets[1]
     assert len(new_datasets) == len(datasets.models)
     assert len(new_datasets) == 2
+
+    tmp = np.ones(3)
+    with pytest.raises(TypeError):
+        datasets[0] + tmp
+
+    with pytest.raises(TypeError):
+        datasets + tmp


docs/release-notes/6256.feature.rst
```diff
@@ -0,0 +1 @@
+Enables the addition of Datasets with each other and with individual dataset to create a new Datasets
```
Astro-Kirsty:
```suggestion
Enable the addition of `~gammapy.datasets.Datasets` and individual `~gammapy.datasets.Dataset` objects to create a new `~gammapy.datasets.Datasets` instance.
```

Astro-Kirsty (changes requested):
Thank you @QRemy

gammapy/datasets/core.py
```diff
@@ -126,6 +126,14 @@ def _compute_residuals(data, model, method="diff"):
                 )
         return residuals
 
+    def __add__(self, other):
```
registerrier, line 129: While the other `__add__` is perfectly fine, I am less convinced by this one. In particular the addition of `Dataset` objects is ambiguous.
If you want to support `dataset + datasets` you might add a `Datasets.__radd__`

gammapy/datasets/core.py
```diff
@@ -630,3 +638,11 @@ def index(self, key):
 
     def __len__(self):
         return len(self._datasets)
+
+    def __add__(self, other):
```
registerrier, line 642: How about `__iadd__` as well?

gammapy/datasets/core.py
QRemy: I mirrored what we support for Models, the implementation here is almost identical.

gammapy/datasets/core.py
QRemy: same reason as above

QRemy:
> In general, we should make clear in the documentation that addition of Datasets does not conserve models or covariance.

It does conserve the models and the covariance.

bkhelifi (dismissed):
THanks @QRemy

QRemy:
Comments are implemented, CI failures are unrelated merging now.

Issue #6258, opened 2025-12-04
`SpectrumDataset.create()` creates PSFMap that can not be applied if RegionGeom without `region` attribute


k-egg:
**Gammapy version**
2.0 and 1.3

**Bug description**
`MapDataset.create()` (called by `SpectrumDataset.create()`) always creates a PSFMap. If the RegionGeom does not have a defined region (`region` attribute), however, an error will occur in the fit if the SkyModel has a spatial_model defined.
Manually removing the PSFMap resolves this error.

**Expected behavior**
A RegionGeom PSFMap should only be created in the empty SpectrumDataset if the region is defined (i.e. the `region` attribute of the RegionGeom exists), since it can otherwise not be parsed in the fit.

**To Reproduce**
```python
energy=MapAxis.from_edges([0.1,1,2,5,10]*u.TeV,name="energy")

geom = RegionGeom(region=None,axes=[energy])
dataset = SpectrumDataset.create(geom)

model = SkyModel(
    spatial_model = GaussianSpatialModel(),
    spectral_model=PowerLawSpectralModel(),
    name="source1"
)

dataset.models=[model]

fit = Fit()
result = fit.run(datasets=[dataset])
```

**Other information**
A PR on this is to follow.

k-egg:
Looking further into this issue, I see now that the `SpectrumDatasetMaker` does not allow the creation of a PSF at all. Most users will thus not run into this issue.
Meanwhile `SpectrumDataset.create()`, however, always creates a PSF. Should that perhaps be changed in general?

PR #6259, opened 2025-12-04
Don't create PSFMap for `SpectrumDataset`

k-egg:
**Description**

This PR solves #6258. When using `SpectrumDataset.create()` a PSFMap is now only created if the RegionGeom has a defined `region` attribute. This resolves the error during the fit.

A test for this behaviour was added.

**Dear reviewer**

This PR is ready for review, let me know what you think!

### DIFF
--- a/gammapy/datasets/tests/test_spectrum.py
+++ b/gammapy/datasets/tests/test_spectrum.py
@@ -208,7 +208,6 @@ def test_spectrum_dataset_create():
     # test that model evaluation works:
     model = SkyModel.create("pl", "gauss", name="test")
     empty_spectrum_dataset.models = [model]
-    empty_spectrum_dataset.npred()
     assert empty_spectrum_dataset.npred().data.sum() == 0
 
     assert empty_spectrum_dataset.name == "test"


gammapy/datasets/tests/test_map.py
```diff
@@ -1118,6 +1119,21 @@ def test_create_high_dimension():
...
+    )
+    e_true = MapAxis.from_edges(
+        np.logspace(-1.0, 1.0, 4), name="energy_true", unit=u.TeV, interp="log"
+    )
+
+    geom = RegionGeom.create(region=None, axes=[e_reco])
+    empty_dataset = SpectrumDataset.create(geom=geom, energy_axis_true=e_true)
+
+    assert empty_dataset.psf is None
+
```
bkhelifi: Can you add also a test directly associated to the issue #6258? with the spatial model into the SkyModel?

k-egg: Hi Bruno, I can add this.
I think the main question (as mentioned in #6258) is if we want `SpectrumDataset.create()` to create a PSFMap at all, given that the SpectrumDatasetMaker does not have that option.

bkhelifi: For sure! It just that it is always good in general to add a test when we see an issue

k-egg:
It was decided in the dev meeting to not create a PSFMap for SpectrumDatasets at all and to implement this via removing the PSFMap in the SpectrumDataset.create() function. I will ammend the PR accordingly.

k-egg:
I noticed there is no check that the geom given to `SpectrumDataset.create()` is a RegionGeom, so I added that and included it in the test.

docs/release-notes/6259.bug.rst
```diff
@@ -0,0 +1 @@
+Changed `SpectrumDataset.create()` to not create a PSFMap anymore and moved function into the `SpectrumDataset` class. Added check for `RegionGeom` of given geometry.
```
bkhelifi: When looking at the files' changes, it is not a change of the create function but the creation of the creation function!

gammapy/datasets/tests/test_spectrum.py
```diff
@@ -199,6 +199,15 @@ def test_spectrum_dataset_create():
         geom, energy_axis_true=e_true, name="test"
     )
 
+    with pytest.raises(TypeError):
+        geom2 = WcsGeom.create()
+        SpectrumDataset.create(geom2, energy_axis_true=e_true)
+
```
bkhelifi, line 207: Maybe one should add a test with a succeded creation of a spectrum dataset and check whether the psfmap is not present

gammapy/datasets/spectrum.py
```diff
@@ -274,6 +274,74 @@ class SpectrumDataset(PlotMixin, MapDataset):
...
+            energy_axis_true=energy_axis_true,
+            migra_axis=migra_axis,
+            reference_time=reference_time,
+            name=name,
+            meta_table=meta_table,
+            **kwargs,
+        )
+
+        # remove PSF
+        dataset.psf = None
```
bkhelifi, line 341: @QRemy: should we not make a `del` of the PSF map to be sure that the memory is correctly clean up? If not, should we add a memory test to be really sure (garbage collector)?

gammapy/datasets/tests/test_spectrum.py
```diff
@@ -199,6 +199,15 @@ def test_spectrum_dataset_create():
         geom, energy_axis_true=e_true, name="test"
     )
 
+    with pytest.raises(TypeError):
```
AtreyeeS:
```suggestion
    with pytest.raises(TypeError, match="`SpectrumDataset` is only supported for `RegionGeom`"):
```

gammapy/datasets/tests/test_spectrum.py
```diff
@@ -199,6 +199,15 @@ def test_spectrum_dataset_create():
...
+        geom2 = WcsGeom.create()
+        SpectrumDataset.create(geom2, energy_axis_true=e_true)
+
+    # test that model evaluation works:
+    model = SkyModel.create("pl", "gauss", name="test")
+    empty_spectrum_dataset.models = [model]
+    empty_spectrum_dataset.npred()
```
AtreyeeS: can you please assert on the `npred.data.sum` to see that it is not filled with reasonable values

docs/release-notes/6259.bug.rst
AtreyeeS: This line is a bit confusing for the reader. Maybe simply say
"Adapted `SpectrumDataset` to avoid creating PSFMap during dataset creation"

AtreyeeS (approved):
Thanks @k-egg ! Looks good to me. Can you please assert that the `npred` creation works ok

gammapy/datasets/tests/test_spectrum.py
k-egg: Hi Bruno, thanks for the comments! This check should already be happening here:

https://github.com/gammapy/gammapy/blob/716e7e3a836fcff968f3c22162df9a6537e08422/gammapy/datasets/tests/test_spectrum.py#L222

docs/release-notes/6259.bug.rst
k-egg: Thanks for the suggestions, I changed the line!

gammapy/datasets/tests/test_spectrum.py
k-egg: I added this now, thanks!

gammapy/datasets/spectrum.py
k-egg: I tried this and ran into issues, I would need to look further into this.

gammapy/datasets/tests/test_spectrum.py
```diff
@@ -199,6 +199,18 @@ def test_spectrum_dataset_create():
...
+    with pytest.raises(
+        TypeError, match="`SpectrumDataset` is only supported for `RegionGeom`"
+    ):
+        geom2 = WcsGeom.create()
+        SpectrumDataset.create(geom2, energy_axis_true=e_true)
+
+    # test that model evaluation works:
+    model = SkyModel.create("pl", "gauss", name="test")
+    empty_spectrum_dataset.models = [model]
+    empty_spectrum_dataset.npred()
```
AtreyeeS:
```suggestion
```

PR #6265, opened 2025-12-09
Adapt estimator tutorial and add warning for lin interp

### LINKED ISSUE #6261: Correct estimators tutorial for `plot_ts_profiles`
In the estimator tutorial we showcase the `plot_ts_profiles` as seen below [here](https://docs.gammapy.org/2.0/tutorials/details/estimators.html#accessing-and-visualising-the-results), however the implementation is causing a number of the blue shaded regions to be ABOVE their orange flux points. 
This is likely because we don't have a wide enough scan for the stat or enough stat scan evaluations. 

This should be fixed and explained in the tutorial. My suggest would be to leave the tutorial as is and explain in detail why we would obtain a figure like this and then do the correct in another figure just below. For example:

```
fp_estimator.norm.scan_min = 0.01
fp_estimator.norm.scan_max = 20
fp_estimator.norm.scan_n_values = 20
```

It is also important to note that this has been incorrect since we introduced it in v1.3

Astro-Kirsty:
This resolves #6261

The issue in the end relates to using a linear interpolation for the the `norm` parameter. After some investigate it is clear that `log` should be utilised. 

This fix is simple to allow the adaption to be included in v2.0.1, however, we should discuss this in further detail and potentially change the implementation at a later stage

### DIFF
--- a/gammapy/estimators/tests/test_flux.py
+++ b/gammapy/estimators/tests/test_flux.py
@@ -160,7 +160,7 @@ def test_flux_estimator_norm_dict(caplog):
     assert estimator.norm.interp == "log"
 
     norm = dict(value=1, name="norm", min=1e-3, max=1e2, interp="lin")
-    estimator = FluxEstimator(source="test", norm=norm)
+    FluxEstimator(source="test", norm=norm)
     assert "WARNING" in [_.levelname for _ in caplog.records]
     assert (
         "Linear interpolation should be used with care on the 'norm' parameter. "


gammapy/estimators/tests/test_flux.py
```diff
@@ -159,6 +159,15 @@ def test_flux_estimator_norm_dict():
     assert_allclose(estimator.norm.max, 1e2)
     assert estimator.norm.interp == "log"
 
+    norm = dict(value=1, name="norm", min=1e-3, max=1e2, interp="lin")
+    estimator = FluxEstimator(source="test", norm=norm)
+    assert "WARNING" in [_.levelname for _ in caplog.records]
+    assert (
+        "Linear interpolation should be used with care on the 'norm' parameter. "
```
AtreyeeS, line 166: Is this warning necessary?

AtreyeeS:
Can you please add the plot described in the issue after this fix?

bkhelifi (approved):
Looks good. I think that the warning is informative for the user today. They need to be used somehow and trained

gammapy/estimators/tests/test_flux.py
Astro-Kirsty: Yes because otherwise the user could try with 'lin' when it is really not advised

Astro-Kirsty:
The updated figure in the tutorial

image

PR #6270, opened 2025-12-11
Add optional normalization on phasecurve models

registerrier:
**Description**

This pull request adds an optional `normalize` argument on `TemplatePhaseCurveTemporalModel`. It is set by default to `True` to keep the previous behaviour.
This is serialised with `model.to_dict()` and read back in `from_dict()`. To keep compatibility with previous version of the dictionary, if the keyword is absent, fall back to default.

It is not clear to me if this can go in a bugfix as it will change the `Models` serialization. Do we need this as part of 2.0.1 and is this fine on 2.1 @fabiopintore , @mashachernyakova and @Cormed-Mack?

This should solves #6138 

**Dear reviewer**

### DIFF
--- a/docs/release-notes/6270.bug.rst
+++ b/docs/release-notes/6270.bug.rst
@@ -1,2 +1,1 @@
-Add an optional normalize argument on `~gammapy.modeling.models.TemplatePhaseCurveTemporalModel`. It is set by default to True,
-for backward compatibility with 2.0.
+Add an optional normalize argument on `~gammapy.modeling.models.TemplatePhaseCurveTemporalModel`. It is set by default to True, for backward compatibility with 2.0.


fabiopintore:
Thanks @registerrier . It looks good to me. 
If this should be considered as a bugfix or not, I'm not sure. The code was working properly, apart for a specific case. So I would consider it rather as new functionality.
My guess is that this code will be included in 2.0.1 as well, we could already perform tests and fine-tuning on the models for the SDC.

docs/release-notes/6270.bug.rst
```diff
@@ -0,0 +1,2 @@
+Add an optional normalize argument on TemplatePhaseCurveTemporalModel. It is set by default to True,
```
Astro-Kirsty:
```suggestion
Add an optional normalize argument on `~gammapy.modeling.models.TemplatePhaseCurveTemporalModel`. It is set by default to True,
```

Astro-Kirsty: I'm not sure of the behaviour if this is split over 2 lines. I would keep it on one

gammapy/modeling/models/temporal.py
```diff
@@ -965,6 +965,8 @@ class TemplatePhaseCurveTemporalModel(TemporalModel):
         A table with 'PHASE' vs 'NORM'.
     filename : str
         The name of the file containing the phase curve.
+    normalize : bool
```
Astro-Kirsty:
```suggestion
    normalize : bool, optional
```

gammapy/modeling/models/temporal.py
```diff
@@ -1040,12 +1052,15 @@ def read(
         ----------
         path : str or `~pathlib.Path`
             Filename with path.
+        normalize : bool
```
Astro-Kirsty:
```suggestion
        normalize : bool, optional
```

Issue #6143, opened 2025-09-18
RegionGeom.from_regions throw errors with some list of regions containing regions distant from each other


mdebony:
**Gammapy version**
The bug exist on version 1.3 and 2.0 (I didn't test older version).

**Bug description**
Sometimes RegionGeom.from_regions throw 'ValueError: cannot convert float NaN to integer' errors. Seems to be correlated when the list of regions contain regions far from each others but I didn't find a clean distance cut (sometime work for regions further away than some creating the issue).

**Expected behavior**
A new RegionGeom object is created without any errors.

**To Reproduce**
A small snippet of code to reproduce the error :
```python
from gammapy.maps import RegionGeom
from regions import CircleSkyRegion
from astropy.coordinates import SkyCoord
import astropy.units as u

list_region = [CircleSkyRegion(SkyCoord(326*u.deg, -13*u.deg), radius=0.4*u.deg), CircleSkyRegion(SkyCoord(165*u.deg, -23*u.deg), radius=0.4*u.deg)]
RegionGeom.from_regions(list_region)
```

**Other information**
The test with gammapy 2.0 was done with regions v0.10.0 and astropy v7.1.0

AtreyeeS:
For the `RegionGeom`, we are fixing the projection to `TAN` (https://github.com/gammapy/gammapy/blob/3a87e9e6108a934cc25efe8d7273e98dac5a57e0/gammapy/maps/region/geom.py#L92), which leads to points that are far from the projection center getting mapped to infinity.

This example works if we adapt the projection or pass a `wcs` , eg:

```
wcs = WcsGeom.create().wcs
RegionGeom.from_regions(list_region, wcs=wcs)
```

Can we adapt the projection here @registerrier ?

PR #6272, opened 2025-12-12
Add docs on RegionGeom.from_regions

AtreyeeS:
Addresses #6143

### DIFF
--- a/docs/release-notes/6272.docs.rst
+++ b/docs/release-notes/6272.docs.rst
@@ -1,1 +1,1 @@
-Clarify usage of `RegionGeom.from_regions` regarding underlying projection effects.
+Clarify usage of `~gammapy.maps.RegionGeom.from_regions` regarding underlying projection effects.


gammapy/maps/region/geom.py
```diff
@@ -735,6 +735,12 @@ def from_regions(cls, regions, **kwargs):
 
         The regions are combined with union to a compound region.
 
+        Note that this function is intended to be used combining close by
+        regions into one RegionGeom. Since it uses an underlying tangential
+        projection undefined behaviour can happen when list of regions
+        contain regions far from each other. If required, one can explicitly
+        pass an appropriate wcs through kwargs.
```
Astro-Kirsty:
```suggestion
        Note that this function is intended for combining close by
        regions into one `~gammapy.maps.RegionGeom`. Since it uses an underlying tangential
        projection, undefined behaviour may occur if the list of input regions
        contains regions far from each other. If required, one can explicitly
        pass an appropriate WCS through ``kwargs``.
```

gammapy/maps/region/geom.py
```diff
@@ -746,6 +752,19 @@ def from_regions(cls, regions, **kwargs):
...
+        --------
+        from gammapy.maps import RegionGeom, WcsGeom
+        from regions import CircleSkyRegion
+        from astropy.coordinates import SkyCoord
+        import astropy.units as u
+
+        list_region = [CircleSkyRegion(SkyCoord(326*u.deg, -13*u.deg), radius=0.4*u.deg),
+        CircleSkyRegion(SkyCoord(325*u.deg, -14*u.deg), radius=0.3*u.deg)]
+        wcs = WcsGeom.create().wcs
+        RegionGeom.from_regions(list_region, wcs=wcs)
```
Astro-Kirsty: Did this run correctly? It prints something so I imagine it would fail:
https://docs.gammapy.org/2.0/development/doc_howto.html#code-in-docstrings-in-python-files

docs/release-notes/6272.bug.rst
```diff
@@ -0,0 +1 @@
+Clarify usage of RegionGeom.from_regions.
```
Astro-Kirsty: This should be a `docs` rst i.e. `6272.docs.rst` not a bug.

Also something a little more description for the sentence is good, like what it actually does

Astro-Kirsty (changes requested):
Thanks @AtreyeeS. I attached some comments

docs/release-notes/6272.docs.rst
```diff
@@ -0,0 +1 @@
+Clarify usage of `RegionGeom.from_regions` regarding underlying projection effects.
```
Astro-Kirsty:
```suggestion
Clarify usage of `~gammapy.maps.RegionGeom.from_regions` regarding underlying projection effects.
```

PR #6274, opened 2025-12-12
Correct documentation before bug release

Astro-Kirsty:
Minor PR. Lets wait til the last moment, so we can use this to correct any other errors that we see


AtreyeeS (approved):
Thanks @Astro-Kirsty 
Merging now

PR #6280, opened 2025-12-15
Simplify drift pointing mode with built-in broadcasting in astropy>=6

morcuended:
**Description**

This pull request does the simplification foreseen in the TODO comment as now Gammapy requires Astropy >= 6.0 (built-in broadcasting is supported).

```
# see https://github.com/astropy/astropy/issues/12965
# TODO: should be solved in astropy v6.0. Simplify once the astropy dependency imposes >=6.0
```

**Dear reviewer**

### DIFF
--- a/gammapy/data/pointing.py
+++ b/gammapy/data/pointing.py
@@ -18,7 +18,6 @@ from astropy.io import fits
 from astropy.table import Table
 from astropy.units import Quantity
 from astropy.utils import lazyproperty
-from gammapy.utils.compat import COPY_IF_NEEDED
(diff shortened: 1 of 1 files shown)


registerrier (approved):
Thanks @morcuended . Looks good!

Are there other places in the code where we have this logic?

morcuended:
> Thanks @morcuended . Looks good!
> 
> Are there other places in the code where we have this logic?

I think yes, there are other occurrences of this `COPY_IF_NEEDED` related also to minimum allowed numpy version.

Issue #6060, opened 2025-08-06
Use pixi and uv for packaging, installation and CI setup


morcuended:
As suggested by @adonath in Slack:

> Hi Gammapy devs, at Scipy 2025 I had the chance of attending a tutorial on pixi (the “successor” to conda) (https://pixi.sh/latest/). I would highly recomment you check it out. It has a much faster dependency resolution, supports environment management (like tox) and tasks, which allow you to build simple pipelines with task dependencies (a bit like make). Here is the tutorial material https://matthewfeickert-talks.github.io/reproducible-ml-for-scientists-with-pixi-scipy-2025/

Also, `uv` tool should be used in the CI/CD workflows, as it is faster setting up environments.

This will require to clean up and unify all the current `setup.py`, `setup.cfg` and `pyproject.toml` files.

natthan-pigoux:
If anyone started working on setting up pixi, I can assign myself to this task.

PR #6281, opened 2025-12-15
infra: setup pixi configuration

natthan-pigoux:
From https://github.com/gammapy/gammapy/issues/6060

Here I setup the [pixi](https://prefix.dev/) environment for gammapy. But it does not replace the tox environment used in the CI.

the pixi confinguration is done via the `pixi.toml` file. But we could also add it in the `pyproject.toml` file. Any preferences?

Since the configuration is quiet long, having a `pixi.toml` seems good to me.

To test it, after installing pixi:
```
pixi install
pixi info
# to start a shell environment
pixi shell -e <env_name>
# to run a task (e.g. command)
pixi run <task_name>
# to run a task or a command in a specific env
pixi run -e <env> <command>
```

The `pixi.lock` is [supposed to be versionned](https://pixi.sh/latest/workspace/lockfile/#when-is-a-lock-file-generated). 

But there is a recommended action to update it: https://pixi.sh/latest/integration/ci/updates_github_actions/ 

It is not done in this PR, but if you think it is nice to have it I can add it here.

Once `pixi` is set up, the `environment-dev.yml` is not needed anymore.

/cc @morcuended @registerrier


natthan-pigoux:
I would need the reviewers to test locally the pixi envs installation and the pixi tasks (I tested only on linux).

natthan-pigoux:
This PR is ready for review.

docs/development/doc_howto.rst
```diff
@@ -321,6 +325,12 @@ To check for broken external links you can use ``tox``:
 
    tox -e linkcheck
 
+Or using ``pixi``::
+
+.. code-block:: bash
+
```
Astro-Kirsty: These lines can be removed as the `::` above already makes it the correct format
```suggestion
```

docs/development/intro.rst
```diff
@@ -189,24 +189,13 @@ repository from your GitHub:
     cd [your-working-path]
     git clone git@github.com:[your-github-username]/gammapy.git
 
-Create the conda environment and activate it:
```
Astro-Kirsty, line 192: Do we really want to remove conda all together and only use pixi?

In the installation guide we still use both: https://docs.gammapy.org/dev/getting-started/install.html

docs/development/intro.rst
```diff
@@ -189,24 +189,13 @@ repository from your GitHub:
...
+The development environment is manage by `pixi <https://pixi.prefix.dev/latest/>`__ . Once **pixi** is installed, one can display the list of available pixi environments and tasks using:
```
Astro-Kirsty:
```suggestion
The development environment is managed by `pixi <https://pixi.prefix.dev/latest/>`__ . 
Once **pixi** is installed, one can display the list of available pixi environments and tasks using:
```

docs/development/intro.rst
```diff
@@ -352,3 +347,31 @@ Finally to make a completely clean folder for build the documentation you can ru
...
+**Using pixi**
+
+Once `pixi <https://pixi.prefix.dev/latest/>`__ is installed, information on environments and tasks can be displayed using:
+
+.. code-block:: bash
+
+    pixi info
+
+To run tests:
+.. code-block:: bash
```
Astro-Kirsty, line 383:
```suggestion
To run tests:

.. code-block:: bash
```

docs/development/intro.rst
```diff
@@ -352,3 +347,31 @@ Finally to make a completely clean folder for build the documentation you can ru
 .. code-block:: bash
 
     make clean-repo
+
+
+**Using pixi**
```
Astro-Kirsty: I would add this as a subheading:
```suggestion
Using pixi
----------
```

docs/development/intro.rst
```diff
@@ -352,3 +347,31 @@ Finally to make a completely clean folder for build the documentation you can ru
...
+
+Once `pixi <https://pixi.prefix.dev/latest/>`__ is installed, information on environments and tasks can be displayed using:
```
Astro-Kirsty: Make sure all lines of code do not exceed 88 characters :)

natthan-pigoux:
@Astro-Kirsty thanks for the review. In general I tried to add some infos about pixi in the docs but I'm not completely sure if I put them at the correct place. Please tell me if some infos must go else where or if some are missing.

Astro-Kirsty (changes requested):
Thank you @natthan-pigoux. In general it looks good, but I wonder if we should support both the conda env and pixi to start with and then move to only pixi once people have used it more?

docs/development/intro.rst
natthan-pigoux: I removed it knowing that it would trigger the discussion :) 
I agree that we should probably keep both for some time. But as we have seen in the past,  `environment-dev.yml` is error prone and needs to be kept up to date. If we setup pixi we should not have to use this file anymore I think.

morcuended:
> Thank you @natthan-pigoux. In general it looks good, but I wonder if we should support both the conda env and pixi to start with and then move to only pixi once people have used it more?

I agree with keeping conda env

natthan-pigoux:
I tested running gammapy tests using pixi here: https://github.com/gammapy/gammapy/pull/6731

PR #5838, opened 2025-04-25
Addition of FitStatistic penalties

registerrier:
**Description**

This pull request introduces `FitStatisticPenalty` classes. 

- *Motivation* : support constraints on multiple parameters e.g. regularity constraints, multiparametric priors

- *Use cases* : 
  - regularized flux points (in true energy, see #5625)
  - background model corrections
  - systematic uncertainty models 
  - etc

- *Proposed approach*:
  - Rather than creating a `MultiNormalPrior` with a logic similar to individual parameter `Prior` objects as in #5468 , the 
     idea here is to define a `FitStatisticPenalty` object. 
  - It takes on init the `Parameters` on which it should apply and implements a `stat_sum()` (could be `evaluate` for clarity) method.  
  - It is set on the `Models` object.
  - It introduces for now a unique `GaussianPriorPenalty` but constructors are provided to build the various useful cases (e.g. 
     diagonal covariance matrix, smoothness constraints). Sub-classing for those cases might be useful though for book- 
     keeping and provenance.
 

**Dear reviewer**

PR #5625, opened 2024-11-29
A regularized flux points estimator

registerrier:
**Description**

This pull request is a proposal draft for a regularized flux points estimator. It would require multi parameter priorss to be fully functional.

**Dear reviewer**

PR #6296, opened 2025-12-19
[Prototype] Multi-source regularized flux points

QRemy:
Combines the idea of #5625, #5838, and #2877 to allow to compute (regularized) flux points in true energy for multiple sources conjointly.


QRemy:
Commits are now pushed on their respective PRs #5625, #5838, #6322. Closing the prototype.

PR #6298, opened 2025-12-23
Fix physics error: Use mass/2 for DM decay spectrum lookup

yemrebahar:
**Description**
This PR fixes a physics bug in the `DarkMatterDecaySpectralModel`.

**The Issue**
The model was initializing `PrimaryFlux` using the full Dark Matter mass (`mass`). However, `PrimaryFlux` relies on PPPC4DMID tables which are computed for annihilation (Total Energy = 2 * Mass). This caused decay spectra to have the wrong shape and an incorrect kinematic endpoint at M_dm instead of M_dm/2.

**The Fix**
Updated the initialization to `PrimaryFlux(mass / 2, ...)` to correctly map the annihilation tables to the decay process, as per Cirelli et al. (2011).


AtreyeeS (commented):
Thanks for the fix @yemrebahar !
Can you please adapt the test in `gammapy/astro/darkmatter/tests/test_spectra.py` ?
If there are any reference values from published papers, it will be good to compare against that.

We will also need a town crier fragment

yemrebahar:
I have updated the test values in test_spectra.py to match the corrected physics (checked against the predicted decay signal presented in the CTAO Perseus Galaxy Cluster paper, Abe et al. 2024). I also added the towncrier newsfragment. Ready for review!

yemrebahar:
Realized that test_utils.py also needed to be adapted for the decay fix. I have just pushed that update as well. Now it should be fully ready.

QRemy (changes requested):
Thanks @yemrebahar, just a minor comment, the fragment should be in gammapy/docs/release-notes/

PR #6299, opened 2025-12-23
Make _get_fov_coord support frame observation time, add unit test

Tobychev:
**Description**
Currently `makers.utils._get_fov_coord` ignores the `obstime` of a `FoVAltAzFrame` being passed when calculating offsets and only uses the `.origin` property which has its own, fixed `origin.obstime`.  
This PR instead uses the centre position of the FoV frame at the specified obstime to calculate offsets. 

It also adds support for passing several obstimes at once, as well as explicit unit tests.

It is ready to review.

### DIFF
--- a/gammapy/makers/utils.py
+++ b/gammapy/makers/utils.py
@@ -2,8 +2,8 @@
 import logging
 import numpy as np
 import astropy.units as u
+from astropy.coordinates import SkyCoord
 from astropy.coordinates.erfa_astrom import erfa_astrom, ErfaAstromInterpolator
-from astropy.coordinates.representation import UnitSphericalRepresentation
 from astropy.table import Table
 from astropy.time import Time
 from gammapy.data import FixedPointingInfo, PointingMode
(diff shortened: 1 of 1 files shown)


gammapy/makers/tests/test_utils.py
```diff
@@ -665,6 +667,65 @@ def test_make_effective_livetime_map():
...
+    sky_coord = sky_geom.to_image().get_coord().skycoord
+    coords = _get_fov_coord(sky_coord, fov_frame, use_offset=True)["offset"]
+    assert_allclose(center_sep, coords[0, 1].value, rtol=3e-5)
+    assert_allclose(center_sep, coords[1, 0].value, rtol=3e-5)
+
+    # Check that pixel centers end up at correct offsets
+    # after letting the sky drift away from the starting position
+    fov_frame = FoVAltAzFrame(
+        origin=fov_origin, location=location, obstime=time_start + 4.303333 * u.minute
+    )
```
registerrier, line 839: I think this is where the issue lies. 
It should not be possible to assign inconsistent `origin` and `obstime`/`location` to a `FoVAltAzFrame`.  The current design is unclear and error-prone because of this. 

The workaround you propose here allows to compute the correct offset when the `origin` and `obstime` are not consistent but it would still compute incorrect `fov_lon` and `fov_lat.`

registerrier:
See related #6456

registerrier:
Removed the merge conflicts and simplified a bit the code. Waiting for the CI to run before merging.

PR #6303, opened 2026-01-04
Pin numpy<2.4

registerrier:
**Description**

This pull request pin numpy dependency <2.4. This should be removed once issues with 1D arrays - scalar conversions are solved.

**Dear reviewer**


registerrier:

PR #6312, opened 2026-01-14
Minor improvement from SonarCube report

bkhelifi:
**Description**
Fron the SonarCube report, one should use `!=` instead of 'not xx == yy'

### DIFF
--- a/gammapy/maps/core.py
+++ b/gammapy/maps/core.py
@@ -1712,11 +1712,11 @@ class Map(abc.ABC):
         maps : list of `Map` objects
             List of maps.
         axis : `MapAxis`, optional
-            If a `MapAxis` is provided the maps are stacked along the last data
+            If a `MapAxis` is provided, the maps are stacked along the last data
             axis and the new axis is introduced. Default is None.
         axis_name : str, optional
-            If an axis name is as string the given the maps are stacked along
-            the given axis name.
+            If an axis name is given, the maps are stacked along
+            the given axis name. Default is None.
 
         Returns
         -------
@@ -1725,12 +1725,11 @@ class Map(abc.ABC):
         """
         geom = maps[0].geom
 
-        if geom.is_image:
-            raise ValueError(
-                "Map.from_stack requires that maps have at least one non-spatial axis"
-            )
-
         if axis_name is None and axis is None:
+            if geom.is_image:
+                raise ValueError(
+                    "Map.from_stack requires that maps have at least one non-spatial axis"
+                )
             axis_name = geom.axes.names[-1]
 
         if axis_name:


gammapy/maps/tests/test_core.py
```diff
@@ -1089,3 +1093,29 @@ def test_make_mask_geom():
         mask_energy.sum_over_axes(["phase", "freq"]).data.sum(),
         48,
     )
+
+
+def test_stack():
```
QRemy, line 1099: Is this related ?

bkhelifi: Well seen;)
It is not related, you're right. But the SonarCube test fails because of a bad coverage. But I have just tried to pass all the tests;)

gammapy/maps/core.py
```diff
@@ -1726,7 +1726,10 @@ def from_stack(cls, maps, axis=None, axis_name=None):
         geom = maps[0].geom
 
         if axis_name is None and axis is None:
-            axis_name = geom.axes.names[-1]
+            try:
+                axis_name = geom.axes.names[-1]
+            except Exception:
+                raise ValueError("The first map should habe a non-spatial axis")
```
AtreyeeS:
```suggestion
                raise ValueError("Non-spatial axis is required for stacking")
```

AtreyeeS (approved):
Thanks @bkhelifi ! Hard work correcting all these small issues :)

gammapy/maps/core.py
```diff
@@ -1726,7 +1726,10 @@ def from_stack(cls, maps, axis=None, axis_name=None):
         geom = maps[0].geom
```
registerrier, line 1726: I think that in this function all maps must share the same non-spatial axes names. And as docstring states all axes must be aligned except for the one used for stacking. Maybe the docstring could be made clearer insisting that image geometries can be flat (i.e. 1 bin non-spatial axis) but not be images (pure 2D).

Instead of a `try except` clause, I would rather explicitly test for the proper behavior. e.g. 
```
 if geom.is_image():
    raise ValueError("Map.from_stack requires that maps have at least one non-spatial axis")
```

gammapy/maps/tests/test_core.py
```diff
@@ -1089,3 +1093,29 @@ def test_make_mask_geom():
...
+    with pytest.raises(ValueError):
+        _ = Map.from_stack(maps=[m1, m2])
+
+    energy3 = MapAxis.from_energy_bounds("1 TeV", "10 TeV", nbin=1)
+    geom3 = WcsGeom.create(binsz=1.0, width=10.0, axes=[energy3])
+    m3 = Map.from_geom(geom3)
+    energy4 = MapAxis.from_energy_bounds("10 TeV", "20 TeV", nbin=1)
+    geom4 = WcsGeom.create(binsz=2.0, width=11.0, axes=[energy4])
+    m4 = Map.from_geom(geom4)
+    with pytest.raises(ValueError):
```
registerrier: The error message here is different (non-aligned 2D geometries). You might test the error message returned.

gammapy/maps/tests/test_core.py
```diff
@@ -1089,3 +1093,29 @@ def test_make_mask_geom():
...
     )
+
+
+def test_stack():
+    geom1 = WcsGeom.create(binsz=1.0, width=10.0)
+    m1 = Map.from_geom(geom1)
+    geom2 = HpxGeom.create(binsz=1.0, width=10.0)
+    m2 = Map.from_geom(geom2)
+
+    with pytest.raises(ValueError):
```
registerrier: Maybe check your error message.

gammapy/maps/tests/test_core.py
```diff
@@ -1089,3 +1093,29 @@ def test_make_mask_geom():
...
+        _ = Map.from_stack(maps=[m1, m2])
```
registerrier, line 1106: Note that `_ = Map.from_stack(maps=[m1, m2], axis=energy3)` should fail too. But it would be simpler it failed with the same error: maps are pure images.
It is better to put the test of the dimensionality early in the function.

registerrier:
Note that those clean-up PRs touch many files. If we want to keep the ability to port future changes easily to v2.0.x, we should backport these PRs as well as long as they touch code that is in v2.0. They change neither the API nor the behavior.

QRemy (approved):
Thanks @bkhelifi ok to merge. The coverage is still below threshold just because these PR touch files where the coverage is already below threshold but the change here do not reduce coverage. So I think it should be addressed in separate PRs, at least one per submodule.

PR #6313, opened 2026-01-14
Minor improvement from SonarCube report (2)

bkhelifi:
**Description**
From the SonarCube report, one should use `_` for the unused variables

### DIFF
--- a/docs/release-notes/6313.cleanup.rst
+++ b/docs/release-notes/6313.cleanup.rst
@@ -1,1 +1,0 @@
-Apply cleaning suggestions from SonarCube (use of _ for unused variables)


QRemy (dismissed):
Thanks @bkhelifi, ok to merge.

registerrier (dismissed):
Thanks @bkhelifi . Looks good. But this type of change don't have a fragment. There is no cleanup entry. Can you remove the file?

registerrier:
> Thanks @bkhelifi ! If you want this to appear in the changelog, you can add a towncrier fragmnet

Hi @AtreyeeS I think we have to decide clearly what changes should go explicitly in the Changelog. Currently the dev doc states that all PRs should have a fragment. I don't think this is true as this would become un-readable. The main categories are bug, feature, doc and infrastructure. But I would say that everything that is a "simple" clean-up (i.e. no change in the API or behaviour) can be omitted not to clutter the release notes.  Adding a general statement about numerous code quality improvements in the introduction can be a good summary for these PRs.
We might update the dev doc accordingly, @Astro-Kirsty ?

registerrier (approved):
Thanks @bkhelifi !

registerrier:
Backporting to v2.0.x  with 2.0.2 milestone with the same reasoning as in #6312

PR #6314, opened 2026-01-14
Minor improvement from SonarCube report (3)

bkhelifi:
**Description**
From the SonarCube report, one should use the python:S117 standard, by renaming local variables to match the regular expression ^[_a-z][a-z0-9_]*$.

### DIFF
--- a/gammapy/catalog/fermi.py
+++ b/gammapy/catalog/fermi.py
@@ -612,10 +612,10 @@ class SourceCatalogObject4FGL(SourceCatalogObjectFermiBase):
         table["flux_errn"] = np.abs(flux_err[:, 0])
         table["flux_errp"] = flux_err[:, 1]
 
-        nu_fnu = self._get_flux_values("nuFnu_Band", "erg cm-2 s-1")
-        table["e2dnde"] = nu_fnu
-        table["e2dnde_errn"] = np.abs(nu_fnu * flux_err[:, 0] / flux)
-        table["e2dnde_errp"] = nu_fnu * flux_err[:, 1] / flux
+        nuFnu = self._get_flux_values("nuFnu_Band", "erg cm-2 s-1")
+        table["e2dnde"] = nuFnu
+        table["e2dnde_errn"] = np.abs(nuFnu * flux_err[:, 0] / flux)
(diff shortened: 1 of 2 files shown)


morcuended:
> Thanks @bkhelifi . Apparently there was an issue with the branching. This contain also changes from the first PR. Can you correct? Also no fragment needed here. I must say that the python rule is arguable in a number of cases. Imposing `dnde_10gev` instead of `dnde_10GeV` can be seen as less readable as the upper letters refer to a unit. Same with `nuFnu`, up to a point. Opinion @morcuended ?

I agree, there are some cases where I think it is justified to deviate to the convention like units, magnitudes. So we can disable the rule from sonar analysis (I think it is not easy to configure it to allow for this kind of regex expressions). We can however set some rules at ruff level to allow for this kind of variable names, but avoid in general non-conventional variable naming in normal cases.

bkhelifi:
> Thanks @bkhelifi ! I think you need to remove the files addressed in #6312 
> We discussed keeping variable names with capital letters for readability, and @morcuended is looking into disabling this complain of sonarcube

So, one should then probably close this PR w/o merging. Agree?

morcuended:
Looking at the occurrences of this issue, I would not disable the rule completely from SonarQube, but rather mark the cases that you want to keep with mixed case style as "Accept: Won't fix" in SonarQube. I suggest to be consistent in the naming, sometimes I see `dnde`, `mDM`, `dN_dE`, `massDM`, which I think it should be revised and unified. So far I've marked these cases as accepted:

- nuFnu
- test_4FGL_DR4
- dnde_10GeV
- test_SNRTrueloveMcKee_radius
- test_SNR_radius_inner
- test_PWN_radius
- n_ISM
- rho_ISM
- eta_B

morcuended:
> > Thanks @bkhelifi ! I think you need to remove the files addressed in #6312
> > We discussed keeping variable names with capital letters for readability, and @morcuended is looking into disabling this complain of sonarcube
> 
> So, one should then probably close this PR w/o merging. Agree?

Agreed. Let's close this and follow-up with an issue where we check for consistency in the naming of these variables. Here #6341

QRemy:
@morcuended  I'll reopen and revert the change for variable you have accepted, but other like cases like `psfMap` could be kept.

PR #6318, opened 2026-01-20
Fix issues for the test ubuntu-latest, devdeps

bkhelifi:
**Description**
For this test using python 3.12 and its dependencies, the CI fails. This PR aims to fix it.

### DIFF
--- a/gammapy/utils/regions.py
+++ b/gammapy/utils/regions.py
@@ -146,7 +146,7 @@ def get_centroid(vertices):
     """Compute centroid of a polygon. Implicitly assumes a flat
     cartesian projection, will probably break for very large polygons.
 
-    Code comes from:
+    Code adapted from:
     https://stackoverflow.com/questions/75699024/finding-the-centroid-of-a-polygon-in-python
 
     Parameters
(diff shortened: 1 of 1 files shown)


gammapy/utils/regions.py
```diff
@@ -161,7 +161,7 @@ def get_centroid(vertices):
     """
     polygon = []
     for vertex in vertices:
-        polygon.append((vertex.ra.degree, vertex.dec.degree))
+        polygon.append((vertex.ra.degree, vertex.dec.degree, 0))
```
registerrier, line 164: why do you need the extra axis here?

gammapy/modeling/parameter.py
```diff
@@ -875,7 +875,7 @@ def select(
                 if frozen:
                     selection[idx] &= par.frozen
                 else:
-                    selection[idx] &= ~par.frozen
+                    selection[idx] &= not par.frozen
```
registerrier, line 878: Is it the only place where we use the bitwise inversion of a boolean variable?

QRemy: couldn't find another place where there is "= ~" with  a boolean variable

gammapy/utils/regions.py
bkhelifi: the main cause the failure is because np.cross does not work anymore in 2D. So, one should pass in 3D, explaining all the tricks

gammapy/modeling/parameter.py
bkhelifi: I found it by chance in this file and spontaneously correct it. But I have not checked systematically in other file: so thanks quentin.
This is still a warning, but a deprecation warning...

PR #6320, opened 2026-01-21
Adapt test to include pre-computed simulated datasets - Part 1

Astro-Kirsty:
We required https://github.com/gammapy/gammapy-data/pull/82 to be merged before this

This is start on updating to use the pre-computed models https://github.com/gammapy/gammapy-data/pull/79

This is a draft as some tests are failing locally. 
In particular, `test_run_pwl_parameter_range` as this dataset was originally created using a `PowerLawSpectralModel` with `amplitude="1e-16 cm-2s-1TeV-1"`. The question is whether we need a simulated dataset with such a low amplitude? Or can we simply use the pre-computed dataset and adapt the test values?

### DIFF
--- a/gammapy/estimators/points/tests/test_lightcurve.py
+++ b/gammapy/estimators/points/tests/test_lightcurve.py
@@ -7,17 +7,18 @@ from astropy.table import Column, Table
 from astropy.time import Time
 from astropy.timeseries import BinnedTimeSeries, BoxLeastSquares
 from gammapy.data import GTI
-from gammapy.datasets import Datasets, FluxPointsDataset
+from gammapy.datasets import (
+    MapDataset,
+    SpectrumDatasetOnOff,
+    Datasets,
+    FluxPointsDataset,
+)
 from gammapy.datasets.actors import DatasetsActor
 from gammapy.estimators import FluxPoints, LightCurveEstimator
 from gammapy.estimators.points.lightcurve import parallel as parallel
-from gammapy.estimators.points.tests.test_sed import (
-    simulate_map_dataset,
-    simulate_spectrum_dataset,
-)
 from gammapy.estimators.map.core import DEFAULT_UNIT
 from gammapy.modeling import Fit
-from gammapy.modeling.models import FoVBackgroundModel, PowerLawSpectralModel, SkyModel
+from gammapy.modeling.models import FoVBackgroundModel, PowerLawSpectralModel, Models
 from gammapy.utils.testing import (
     assert_time_allclose,
     mpl_plot_check,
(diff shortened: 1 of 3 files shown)


Astro-Kirsty:
An update here:
- if we want this to work completely we also need to save another lot of datasets for a different `random_state` see i.e. [here](https://github.com/gammapy/gammapy/blob/8750a134cf948cbdc5efe9b8999da560f04c8997/gammapy/estimators/points/tests/test_lightcurve.py#L164)

AtreyeeS:
> An update here:
> 
> * if we want this to work completely we also need to save another lot of datasets for a different `random_state` see i.e. [here](https://github.com/gammapy/gammapy/blob/8750a134cf948cbdc5efe9b8999da560f04c8997/gammapy/estimators/points/tests/test_lightcurve.py#L164)

These we don't really need to save, do we? We are not really using those two datasets anywhere else, and spectrum simulation is quite fast, so it's ok to compute them on the fly

Astro-Kirsty:
Although it is computationally quick, we are still recomputing every time, so we added the datasets to avoid that. 

We could utilise these datasets in new tests also. Open to suggestions on how we want to proceed?

gammapy/estimators/points/tests/test_lightcurve.py
```diff
@@ -161,14 +162,22 @@ def test_lightcurve_to_time_series():
...
+    model = Models.read(
+        "$GAMMAPY_DATA/datasets/simulations/simulated_spectrum_dataset_PL_model.yaml"
+    )
     dataset_1._name = "dataset_1"
+    dataset_1.models = model
     gti1 = GTI.create("0h", "1h", Time("2010-01-01T00:00:00").tt)
     dataset_1.gti = gti1
 
-    dataset_2 = simulate_spectrum_dataset(model=model, random_state=1)
+    # dataset_2 = simulate_spectrum_dataset(model=model, random_state=1)
```
QRemy:
```suggestion
```

gammapy/estimators/points/tests/test_lightcurve.py
```diff
@@ -573,12 +582,21 @@ def test_lightcurve_estimator_spectrum_datasets_gti_not_include_in_time_interval
...
+    )
+    models = Models.read(
+        "$GAMMAPY_DATA/datasets/simulations/simulated_map_dataset_model.yaml"
+    )
+    dataset_1.models = models
     gti1 = GTI.create("0 h", "1 h", "2010-01-01T00:00:00")
     dataset_1.gti = gti1
 
-    dataset_2 = simulate_map_dataset(random_state=1, name="dataset_2")
+    # dataset_2 = simulate_map_dataset(random_state=1, name="dataset_2")
```
QRemy:
```suggestion
```

gammapy/estimators/points/tests/test_lightcurve.py
```diff
@@ -161,14 +162,22 @@ def test_lightcurve_to_time_series():
...
+    )
     dataset_1._name = "dataset_1"
+    dataset_1.models = model
     gti1 = GTI.create("0h", "1h", Time("2010-01-01T00:00:00").tt)
     dataset_1.gti = gti1
 
-    dataset_2 = simulate_spectrum_dataset(model=model, random_state=1)
+    # dataset_2 = simulate_spectrum_dataset(model=model, random_state=1)
+    dataset_2 = dataset_1.copy()
     dataset_2._name = "dataset_2"
```
QRemy:
```suggestion
    dataset_2 = dataset_1.copy(name="dataset_2")
```

QRemy (dismissed):
Thanks @Astro-Kirsty, looks good, just few minor suggestions.

QRemy:
A number of tests still need to be adapted, can you look into that @Astro-Kirsty ?

Astro-Kirsty:
@QRemy, we were not sure whether to implement this or not, as we will also need additional simulated datasets that are different from the one created. We can discuss further in an upcoming dev call

PR #6322, opened 2026-01-22
Addition of `FluxCollectionEstimator`

QRemy:
Flux points implementation fitting multiple source simultaneously as discussed in #2877.

### DIFF
--- a/gammapy/estimators/points/sed.py
+++ b/gammapy/estimators/points/sed.py
@@ -326,7 +326,7 @@ class FluxCollectionEstimator:
     n_sigma_ul : float, optional
         Number of sigma to use for upper limit computation. Must be a positive value.
         Default is 2.
-    norm : `~gammapy.modeling.Parameter` or dict, optional
+    norm : `~gammapy.modeling.Parameter`, optional
         Norm parameter used for the fit.
         Default is None and a new parameter is created with value=1, name="norm".
         If the `solver` is a sampler the default prior is uniform between [-10, 10].


bkhelifi:
While comparing with the FluxPointEstimation, the API options are quite different.
In `FluxPointEstimation`, we do have :
- source
- n_sigma
- n_sigma_ul
- n_sigma_sensitivity
- selection_optional # which is quite important
- energy_edges
- fit
- reoptimize
- sum_over_energy_groups
- n_jobs
- parallel_backend
- norm

Whereas here:
- energy_edges # in common
-  models  # in common but with a different name, one should use source instead, no?
- dataset  # in common
- n_sigma_ul  # in common
- zero_min # with the standard estimator, we are using either priors or min value of the flux parameter to achieve that
- fit  # in common

There are then missing parameters here, and one extra one that one should not have in order to have an unify API.

QRemy:
> While comparing with the FluxPointEstimation, the API options are quite different. In `FluxPointEstimation`, we do have :
> 
> * source

It should be model/models as a component is not necessarily a source

> * n_sigma
> * n_sigma_ul
> * n_sigma_sensitivity
> * selection_optional

The behaviour here is to compute ul sytematically but only if below n_sigma_ul (which is faster)
for the other options it's unclear if they are relevant/feasible in this case where we want to compute them simultaneously for all the models.

> * reoptimize

not applicable, by definition here the models given are re-optimized the other are frozen

> * sum_over_energy_groups

not relevant, the fit is done on a masked dataset not sliced,
sum_over_energy_groups would be unnecessary and overcomplicated

> * n_jobs
> * parallel_backend

from what I have tested there was no gain from parallelisation as this implementation is better optimised, 
so I removed it but could be added

> * norm

could be used instead of zero_min

I would prefers to keep it a separate light class that solve a specific problem rather than merging it with the FluxPointsEstimator.  The FluxPointsEstimator  would become an even more complex and we would have to explain why some option are not available depending of the number of models you request, which will be for sure utterly confusing for users.

bkhelifi:
I understand, agree.
One still needs: n_sigma, n_sigma_sensitivity, 
Having selection_optional aims to have the same API and to same a bit of computing time for users. Having norm permits here again to have the same API.

Because this Estimator is computing the same quantity than the FluxPointEstimator, I think that this is great to really have the same arguments when possible.

gammapy/estimators/points/sed.py
```diff
@@ -282,3 +290,374 @@ def _nan_result(self, datasets, model, energy_min, energy_max):
...
+        Number of sigma to use for upper limit computation. Must be a positive value.
+        Default is 2.
+    norm : `~gammapy.modeling.Parameter` or dict, optional
+        Norm parameter used for the fit.
+        Default is None and a new parameter is created with value=1, name="norm".
+        If the `solver` is a sampler the default prior is uniform between [-10, 10].
+    solver : `Fit` or `Sampler`
+        Fit or Sampler instance specifying the backend and options.
+        Default is a Sampler with options live_points=300, frac_remain=0.3.
+    reoptimize : bool
```
facero, line 336: Just checked that this default reoptimize is the same as `FluxPointEstimator`.
the default value in fermipy SED is also False for the bkg model.
If I want to let free only the background (iso+iem in Fermi or hadrons in IACTs) one would have give a subset of the SkyModel to models ? or how would that work ?

QRemy:
requires numpy>=2
same as https://github.com/gammapy/gammapy/pull/6389#issuecomment-3884240265

QRemy:
Few examples (see https://github.com/gammapy/gammapy/pull/6389#issuecomment-3896723716 for the equivalent violin plots):
Screenshot 2026-02-13 at 12 34 22
Screenshot 2026-02-13 at 12 34 29

QRemy:

Astro-Kirsty:
As there is only one review, we should wait!

docs/release-notes/6322.feature.rst
```diff
@@ -0,0 +1 @@
+Add a new `FluxCollectionEstimator` that allow to compute flux points or flux samples for a set of sources conjointly.
```
AtreyeeS:
```suggestion
Add a new `FluxCollectionEstimator` that enables the computation of flux points or flux samples jointly for a group of sources.
```

gammapy/estimators/points/sed.py
```diff
@@ -282,3 +290,369 @@ def _nan_result(self, datasets, model, energy_min, energy_max):
...
+        Number of sigma to use for asymmetric error computation. Must be a positive value.
+        Default is 1.
+    n_sigma_ul : float, optional
+        Number of sigma to use for upper limit computation. Must be a positive value.
+        Default is 2.
+    norm : `~gammapy.modeling.Parameter` or dict, optional
+        Norm parameter used for the fit.
+        Default is None and a new parameter is created with value=1, name="norm".
+        If the `solver` is a sampler the default prior is uniform between [-10, 10].
+    solver : `Fit` or `Sampler`
```
AtreyeeS: `~gammapy.modeling.Fit` or `~gammapy.modeling.Sampler`

gammapy/estimators/points/sed.py
```diff
@@ -282,3 +290,369 @@ def _nan_result(self, datasets, model, energy_min, energy_max):
...
+                            TemplateNPredModel(
+                                npred,
+                                name=name + "_" + fp_dataset.name,
+                                spectral_model=spectral_models[name],
+                                datasets_names=[fp_dataset.name],
+                            )
+                        )
+                    else:
+                        npred_frozen.stack(npred)
+            bkg_frozen = TemplateNPredModel(
```
AtreyeeS: The bkg_frozen is then only the irf background, ie, `dataset.background`?

gammapy/estimators/points/sed.py
QRemy: SkyModels (including iso+iem) that have to be reoptimized should be given in models, this reoptimize option  applies only to dataset.background_model which is supposed to be only the CR bkg.

gammapy/estimators/points/sed.py
QRemy: no `dataset.background_model` is `bkg_model`
while `bkg_frozen` is a precomputed `npred` including all the frozen SkyModels not listed in `models`

QRemy:
> The test seems to be quite complicated. 

True but theses tests are designed as real unit tests for the FluxPointsCollection class only without actually  fitting and sampling. Tests real data analysis would be quite time consuming.
I simplified them a bit removing some duplication.

docs/release-notes/6322.feature.rst
```diff
@@ -0,0 +1 @@
+Add a new `FluxCollectionEstimator` that enables the computation of flux points or flux samples jointly for a group of sources.
```
bkhelifi: or `~gammapy.estimators.FluxCollectionEstimator`?

gammapy/estimators/points/sed.py
```diff
@@ -282,3 +290,369 @@ def _nan_result(self, datasets, model, energy_min, energy_max):
             result.update({"norm_sensitivity": np.nan})
 
         return result
+
+
+class FluxCollectionEstimator:
+    """Estimate the flux points from a collection of sources simultaneously.
```
bkhelifi, line 302: A global question: do we have decided to not expose this new class? (because the __init__.py does not contain it)

gammapy/estimators/points/sed.py
```diff
@@ -282,3 +290,369 @@ def _nan_result(self, datasets, model, energy_min, energy_max):
...
+        Fit or Sampler instance specifying the backend and options.
+        Default is a Sampler with options live_points=300, frac_remain=0.3.
+    reoptimize : bool
+        Whether to reoptimize the background models. Default is False.
+        Only SkyModel given in `models` will be fitted the others remain frozen,
+        regardless of this option.
+    selection_optional : list of str, optional
+        Which additional quantities to estimate. Available options are:
+            * "errn-errp": estimate asymmetric errors on flux.
+        Fit solver computes upper limits if sqrt(TS) < n_sigma_ul.
```
bkhelifi: This is problematic! `n_sigma_ul` is the number of sigma to use for upper limit computation, and not the number of sigma used to compute upper limits. Mixing both concepts is a bad idea.

docs/release-notes/6322.feature.rst
QRemy:
```suggestion
Add a new `~gammapy.estimators.FluxCollectionEstimator` that enables the computation of flux points or flux samples jointly for a group of sources.
```

gammapy/estimators/points/sed.py
QRemy:
```suggestion
    solver : `~gammapy.modeling.Fit` or `~gammapy.modeling.Sampler`
```

gammapy/estimators/points/sed.py
```diff
@@ -282,3 +290,369 @@ def _nan_result(self, datasets, model, energy_min, energy_max):
...
+        Default is 2.
+    norm : `~gammapy.modeling.Parameter` or dict, optional
+        Norm parameter used for the fit.
+        Default is None and a new parameter is created with value=1, name="norm".
+        If the `solver` is a sampler the default prior is uniform between [-10, 10].
+    solver : `Fit` or `Sampler`
+        Fit or Sampler instance specifying the backend and options.
+        Default is a Sampler with options live_points=300, frac_remain=0.3.
+    reoptimize : bool
+        Whether to reoptimize the background models. Default is False.
```
QRemy:
```suggestion
        Whether to reoptimize each dataset.background_model. Default is False.
```

gammapy/estimators/points/sed.py
```diff
@@ -287,3 +296,381 @@ def _nan_result(self, datasets, model, energy_min, energy_max):
...
+        Energy edges of the flux point bins.
+    models : str or int
+        Source models for which the flux points are computed (others are frozen).
+    n_sigma : float, optional
+        Number of sigma to use for asymmetric error computation. Must be a positive value.
+        Default is 1.
+    n_sigma_ul : float, optional
+        Number of sigma to use for upper limit computation. Must be a positive value.
+        Default is 2.
+    norm : `~gammapy.modeling.Parameter` or dict, optional
```
AtreyeeS: Do you want to support the `dict` option here? Its not clear what keys the dict should contain

gammapy/estimators/points/sed.py
```diff
@@ -287,3 +296,381 @@ def _nan_result(self, datasets, model, energy_min, energy_max):
...
+
+        return fp_result
+
+    def run(self, datasets):
+        """Compute flux point in each energy band
+
+        Parameters
+        ----------
+        datasets : `Datatets`
+            Datasets used to compute the flus points.
```
AtreyeeS:
```suggestion
            Datasets used to compute the flux points.
```

gammapy/estimators/points/sed.py
```diff
@@ -287,3 +296,381 @@ def _nan_result(self, datasets, model, energy_min, energy_max):
...
+            Datasets must share the same geometry.
+
+        Returns
+        -------
+        result : dict
+            Dict with results
```
AtreyeeS, line 587: It can be good to explain better what this returns, since this is not a FluxPoints object.
At least, how to obtain the flux of each source. More details can go in a tutorial later

gammapy/estimators/points/sed.py
```diff
@@ -287,3 +296,381 @@ def _nan_result(self, datasets, model, energy_min, energy_max):
...
+                prior = UniformPrior(min=-norm_max, max=norm_max)
+            else:
+                prior = None
+            self.norm = Parameter(name="norm", value=1, unit="", prior=prior)
+
+        # define energy edges
+        self.energy_edges = energy_edges
+        self.ne = len(self.energy_edges)
+        self.energy_centers = (energy_edges[:-1] * energy_edges[1:]) ** 0.5
+        self.energy_unit = "TeV"
```
registerrier: Why do you impose TeV here? Why not use `energy_edges.unit`?

gammapy/estimators/points/sed.py
```diff
@@ -287,3 +296,381 @@ def _nan_result(self, datasets, model, energy_min, energy_max):
...
+            fp_result["ts"][km] = ts_null
+
+        return fp_result
+
+    def run(self, datasets):
+        """Compute flux point in each energy band
+
+        Parameters
+        ----------
+        datasets : `Datatets`
```
registerrier:
```suggestion
        datasets : `~gammapy.datasets.Datasets`
```

gammapy/estimators/points/sed.py
```diff
@@ -287,3 +296,381 @@ def _nan_result(self, datasets, model, energy_min, energy_max):
...
+        self.energy_unit = "TeV"
+
+        self.dnde_unit = u.Unit("cm-2 s-1 TeV-1")
+
+    def _prepare_datasets(self):
+        """define datasets with cached npred models to be renormalized"""
+
+        spectral_models = {}
+        for m in self.models:
+            spectral_models[m.name] = PowerLawNormSpectralModel(norm=self.norm.copy())
```
registerrier: Why not use `ScaleSpectralModel` ?

gammapy/estimators/points/sed.py
```diff
@@ -287,3 +296,381 @@ def _nan_result(self, datasets, model, energy_min, energy_max):
...
+            spectral_models[m.name].tilt.frozen = True
+
+        fp_datasets = []
+        for d in self.datasets:
+            fp_dataset = d.copy(name=d.name)
```
registerrier: This is a deepcopy. Is it needed?

gammapy/estimators/points/sed.py
```diff
@@ -287,3 +296,381 @@ def _nan_result(self, datasets, model, energy_min, energy_max):
...
+    (``~gammapy.modeling.Fit``) or with a sampler
+    (``~gammapy.modeling.Sampler``), which can be used to derive asymmetric
+    errors and upper limits.
+
+
+    Parameters
+    ----------
+    energy_edges : `~astropy.units.Quantity`
+        Energy edges of the flux point bins.
+    models : str or int
```
registerrier:
```suggestion
    models : list of str or int
```
Note that this is a bit confusing as in general `models` refers to a `Models` object.

gammapy/estimators/points/sed.py
```diff
@@ -287,3 +296,381 @@ def _nan_result(self, datasets, model, energy_min, energy_max):
...
+        Source models for which the flux points are computed (others are frozen).
```
registerrier, line 322:
```suggestion
        Source models names or indices for which the flux points are computed (others are frozen).
```

gammapy/estimators/points/sed.py
```diff
@@ -287,3 +296,381 @@ def _nan_result(self, datasets, model, energy_min, energy_max):
...
+    def _compute_ts(datasets, param):
+        """Test statistic against no source as null hypothesis"""
+        cash = datasets._stat_sum_likelihood()
+        with Parameters([param]).restore_status():
+            param.value = 0
+            cash0 = datasets._stat_sum_likelihood()
+        return cash0 - cash
+
+    def _run_fit(self, fp_datasets, spectral_models):
+        """compute npred, dnde, TS, errn, errp, and ul (if necessasy)"""
```
registerrier:
```suggestion
        """Compute flux estimates, uncertainties and UL using log-likelihood profile (i.e. using `~gammapy.modeling.Fit`)."""
```

gammapy/estimators/points/sed.py
```diff
@@ -287,3 +296,381 @@ def _nan_result(self, datasets, model, energy_min, energy_max):
...
+        fp_result = dict(
+            npred=np.zeros(self.ns),
+            norm=np.zeros(self.ns),
+            norm_err=np.zeros(self.ns),
+            norm_errn=np.zeros(self.ns),
+            norm_errp=np.zeros(self.ns),
+            norm_ul=np.zeros(self.ns),
+            ts=np.zeros(self.ns),
+            solver_results=fit_results,
+        )
```
registerrier: This is duplicated code. It could go in a helper function.

gammapy/estimators/points/sed.py
```diff
@@ -287,3 +296,381 @@ def _nan_result(self, datasets, model, energy_min, energy_max):
...
+        -------
+        result : dict
+            Dict with results
+        """
+
+        for kd, d in enumerate(datasets):
+            if kd == 0:
+                self.geom = d.counts.geom
+            elif d.counts.geom != self.geom:
+                raise ValueError("Inconstistant geometries between datasets")
```
registerrier: It would be cleaner to make this validation a standalone function, why not  directly `Datasets.is_all_same_geom()`?

gammapy/estimators/points/sed.py
```diff
@@ -287,3 +296,381 @@ def _nan_result(self, datasets, model, energy_min, energy_max):
...
+
+        for d in datasets:
+            d.npred()  # precompute npred
+        self.datasets = datasets
```
registerrier: Estimators should remain stateless. Is it really needed?

gammapy/estimators/points/sed.py
```diff
@@ -287,3 +296,381 @@ def _nan_result(self, datasets, model, energy_min, energy_max):
...
+        fp_results = dict(
+            npred=np.zeros((self.ne - 1, self.ns)),
+            npred_err=np.zeros((self.ne - 1, self.ns)),
+            norm=np.zeros((self.ne - 1, self.ns)),
+            norm_err=np.zeros((self.ne - 1, self.ns)),
+            norm_errn=np.zeros((self.ne - 1, self.ns)),
+            norm_errp=np.zeros((self.ne - 1, self.ns)),
+            norm_ul=np.zeros((self.ne - 1, self.ns)),
+            ts=np.zeros((self.ne - 1, self.ns)),
+            solver_results=np.empty(self.ne - 1, dtype=object),
```
registerrier: Is an array better than a plain list here?

gammapy/estimators/points/sed.py
```diff
@@ -287,3 +296,381 @@ def _nan_result(self, datasets, model, energy_min, energy_max):
...
+        )
+
+        for ke in range(self.ne - 1):
+            with set_and_restore_mask_fit(
+                fp_datasets,
+                energy_min=self.energy_edges[ke],
+                energy_max=self.energy_edges[ke + 1 - self.ne],
+            ):
```
registerrier:
```suggestion
        for ke, emin, emax in enumerate(zip(self.energy_edges[:-1], self.energy_edges[1:])):
            with set_and_restore_mask_fit(fp_datasets, energy_min=emin, energy_max=emax):
```
This could be even cleaner if `energy_edges` is stored as a `MapAxis` directly.

gammapy/estimators/points/sed.py
```diff
@@ -287,3 +296,381 @@ def _nan_result(self, datasets, model, energy_min, energy_max):
...
+                self.geom = d.counts.geom
+            elif d.counts.geom != self.geom:
+                raise ValueError("Inconstistant geometries between datasets")
+
+        for d in datasets:
+            d.npred()  # precompute npred
+        self.datasets = datasets
+        fp_datasets, spectral_models = self._prepare_datasets()
+
+        fp_results = dict(
```
registerrier: A list of `fp_result` would be simpler

gammapy/estimators/points/sed.py
```diff
@@ -287,3 +296,381 @@ def _nan_result(self, datasets, model, energy_min, energy_max):
...
+                    fp_result = self._run_fit(*args)
+
+            fp_results["npred"][ke, :] = fp_result["npred"]
+            fp_results["norm"][ke, :] = fp_result["norm"]
+            fp_results["norm_err"][ke, :] = fp_result["norm_err"]
+            fp_results["norm_errn"][ke, :] = fp_result["norm_errn"]
+            fp_results["norm_errp"][ke, :] = fp_result["norm_errp"]
+            fp_results["norm_ul"][ke, :] = fp_result["norm_ul"]
+            fp_results["ts"][ke, :] = fp_result["ts"]
+            fp_results["solver_results"][ke] = fp_result["solver_results"]
```
registerrier: fp_result["emin"]=emin
fp_result["emax"]=emax
fp_results.append(fp_result)

gammapy/estimators/points/sed.py
```diff
@@ -287,3 +296,381 @@ def _nan_result(self, datasets, model, energy_min, energy_max):
...
+
+        return self._get_flux_points_dict(fp_results)
+
+    def _get_flux_points_dict(self, fp_results):
+        """get flux points for each mddel
```
registerrier:
```suggestion
        """Extract flux points for each model from list of results. 
```

gammapy/estimators/points/sed.py
```diff
@@ -287,3 +296,381 @@ def _nan_result(self, datasets, model, energy_min, energy_max):
...
+
+        Parameters
+        ----------
+        fp_results : dict
+            dict used to generate the flus points table.
```
registerrier:
```suggestion
            dict used to generate the flux points table.
```

gammapy/estimators/points/sed.py
```diff
@@ -287,3 +296,381 @@ def _nan_result(self, datasets, model, energy_min, energy_max):
...
+        )
+        if isinstance(self.solver, Sampler):
+            fp_dict["samples"] = {"dnde": {}}
+
+        for km, m in enumerate(self.models):
+            model = _get_reference_model(m, self.energy_edges)
+            table = Table()
+            table["e_min"] = self.energy_edges[:-1].to(self.energy_unit)
+            table["e_max"] = self.energy_edges[1:].to(self.energy_unit)
+            table["e_ref"] = self.energy_centers.to(self.energy_unit)
```
registerrier: Why is the energy conversion needed?

gammapy/estimators/points/sed.py
```diff
@@ -287,3 +296,381 @@ def _nan_result(self, datasets, model, energy_min, energy_max):
...
+            fp_results["norm_err"][ke, :] = fp_result["norm_err"]
+            fp_results["norm_errn"][ke, :] = fp_result["norm_errn"]
+            fp_results["norm_errp"][ke, :] = fp_result["norm_errp"]
+            fp_results["norm_ul"][ke, :] = fp_result["norm_ul"]
+            fp_results["ts"][ke, :] = fp_result["ts"]
+            fp_results["solver_results"][ke] = fp_result["solver_results"]
+
+        return self._get_flux_points_dict(fp_results)
+
+    def _get_flux_points_dict(self, fp_results):
```
registerrier, line 616: You could specialize the function for regular `Fit` or `Sampler`

registerrier:
I can make a number of changes as a PR on this branch @QRemy . Is that OK ?

QRemy:
> I can make a number of changes as a PR on this branch @QRemy . Is that OK ?

Let me change a few things first

registerrier:
Hi @QRemy . See https://github.com/QRemy/gammapy/pull/2

gammapy/estimators/points/sed.py
```diff
@@ -287,3 +296,376 @@ def _nan_result(self, datasets, model, energy_min, energy_max):
...
+        Energy edges of the flux point bins.
+    models : ~gammapy.modeling.Models` or list
+        Source models for which the flux points are computed (others are frozen).
+    n_sigma : float, optional
+        Number of sigma to use for asymmetric error computation. Must be a positive value.
+        Default is 1.
+    n_sigma_ul : float, optional
+        Number of sigma to use for upper limit computation. Must be a positive value.
+        Default is 2.
+    norm : `~gammapy.modeling.Parameter` or dict, optional
```
QRemy:
```suggestion
    norm : `~gammapy.modeling.Parameter`, optional
```

PR #6187, opened 2025-10-07
Unbinned likelihood analysis 

guillaumegrolleron:
**Description**

This pull request aims to implement unbinned likelihood analysis in gammapy. I have been working to minimize changes in the core code, and have based my implementation on recommendations made in https://github.com/gammapy/gammapy/pull/4253. 

For now, I only support 1D ON-OFF analysis for point-source models. I have tested my implementation and cross-checked with standard binned ON-OFF analysis for both simulations and real Crab data analysis. These codes are not in the PR, but could be added for tests. 

This PR is now a draft to first get feedback. 

Although I tried to minimize changes in the code, there are still a lot of things, and I'm ready to discuss splitting it into multiple PR. 

The documentation is partially implemented, and unit tests are still missing. But I think it should be easier to do it in a separate PR.

PR #6323, opened 2026-01-22
Unbinned Analysis  : EventDataset (1/7)

guillaumegrolleron:
**Description**
 
 

This PR is the first part of the implementation of the unbinned analysis that I propose. It concerns the EventDataset. 
There are two classes: EventDataset and EventDatasetOnOff. The first is a base class ,and the second one is used for reflected background data reduction with unbinned analysis. 

The implementation comes from PR #6187  which has been split into mulitple PR


QRemy:

PR #6324, opened 2026-01-22
 Unbinned Analysis : Unbinned geometry / Region and Map (2/7)

guillaumegrolleron:
**Description**

This PR follows PR #6323. Here, we are focusing on the implementation of the unbinned geometries (an maps or regions) used by the EventDataset to perform the unbinned analysis.

### DIFF
--- a/gammapy/maps/axes.py
+++ b/gammapy/maps/axes.py
@@ -188,8 +188,7 @@ class MapAxis:
         """
         if self.name != required_name:
             raise ValueError(
-                "Unexpected axis name,"
-                f' expected "{required_name}", got: "{self.name}"'
+                f'Unexpected axis name, expected "{required_name}", got: "{self.name}"'
             )
 
     def is_aligned(self, other, atol=2e-2):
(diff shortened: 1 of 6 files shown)


gammapy/maps/axes.py
Tobychev, line 1: These changes should go in their own PR, and you should probably use your `_format_plot_axis` function for both `format_xaxis` and `format_yaxis`

gammapy/maps/geom.py
```diff
@@ -663,19 +663,44 @@ def create_mask(self, axis_name, edge_min=None, edge_max=None, round_to_edge=Fal
...
+    def is_unbinned(self):
+        """Whether the geom is unbinned (i.e. contains a `LabelMapAxis`)."""
+        if self.axes is None:
+            return False
+        for axis in self.axes:
+            if axis.__class__.__name__ in [
+                "LabelMapAxis",
+            ]:
+                return True
+        return False
```
QRemy, line 706:
```suggestion
```

gammapy/maps/geom.py
```diff
@@ -663,19 +663,44 @@ def create_mask(self, axis_name, edge_min=None, edge_max=None, round_to_edge=Fal
...
-        mask = np.expand_dims(
-            mask, axis=tuple(np.where(np.array(axes_names) != axis_name)[0])
-        )
-        data = np.broadcast_to(mask, shape=self.data_shape)
+            axes_names.reverse()
+            mask = (axis_edges[:-1] >= edge_min) & (axis_edges[1:] <= edge_max)
+            mask = np.expand_dims(
+                mask, axis=tuple(np.where(np.array(axes_names) != axis_name)[0])
+            )
+            data = np.broadcast_to(mask, shape=self.data_shape)
```
QRemy, line 693: remove this change an redefine `create_mask` for the unbinned geom class

gammapy/maps/core.py
```diff
@@ -1693,14 +1756,30 @@ def normalize(self, axis_name=None):
...
+            Along which axis to compute the norm.
+
+        Returns
+        -------
+        norm : `~astropy.units.Quantity`
+            Norm array.
+        """
+        cumsum = self.cumsum(axis_name=axis_name).quantity
+        axis = self.geom.axes.index_data(axis_name=axis_name)
+        return cumsum.max(axis=axis, keepdims=True)
```
QRemy, line 1781: why do we need norm to be defined ? is it used somewhere else ?
if not I would remove this change, and if it is needed I would still keep it as a private method for now.

gammapy/maps/core.py
```diff
@@ -1685,6 +1683,71 @@ def integral(self, axis_name, coords, **kwargs):
...
+            geom=upsample_geom,
+            data=gradient_data,
+            unit=cumsum.unit,
+        )
+        return u.Quantity(
+            gradient.interp_by_coord(coords, **kwargs),
+            gradient.unit,
+            copy=COPY_IF_NEEDED,
+        )
+
```
QRemy, line 1750: The use case for this seems too specific to be defined of the generic map class and to be public methods.
evaluate seems to do something quite different of our other evaluate method so a more explicit name should probably used.

gammapy/maps/geom.py
QRemy, line 706: remove this

gammapy/maps/geom.py
QRemy, line 693: remove this change and rewrite a specific create_mask method on the unbinned geom

gammapy/maps/core.py
```diff
@@ -1693,14 +1756,30 @@ def normalize(self, axis_name=None):
...
+        Returns
+        -------
+        norm : `~astropy.units.Quantity`
+            Norm array.
+        """
+        cumsum = self.cumsum(axis_name=axis_name).quantity
+        axis = self.geom.axes.index_data(axis_name=axis_name)
+        return cumsum.max(axis=axis, keepdims=True)
+
```
QRemy, line 1782: This norm. method  seems to be used only here. so I would remove it and keep  the code as it is.
If necessary elsewhere you can keep it but make it a private method.

QRemy (commented):
Thanks @guillaumegrolleron 

No sure why UnbinnedRegionGeom is needed but there is some simplification possible.

QRemy:

PR #6325, opened 2026-01-22
 Unbinned Analysis : Unbinned IRF (3/7)

guillaumegrolleron:
**Description**
This PR follows #6324. Here, there is an implementation of the computation of unbinned IRF, based on the unbinned geometries defined in #6324.

PR #6326, opened 2026-01-22
Unbinned Analysis : Unbinned Evaluator (4/7)

guillaumegrolleron:
**Description**
Following PR #6325, this PR addresses the implementation of an unbinned evaluator. I is completely inspired by the standard evaluator. It has been adapted to the unbinned analysis, where the evaluation of the signal and background is specific to the unbinned method. 

The main change is the need for a proper normalization of the PDF of the signal and the background. The critical point is that for unbinned analysis, we do not need a n_pred, but we need a continuous probability for events to be signal or background

### DIFF
--- a/gammapy/datasets/unbinned_evaluator.py
+++ b/gammapy/datasets/unbinned_evaluator.py
@@ -1,0 +1,627 @@
+import html
+import logging
+import numpy as np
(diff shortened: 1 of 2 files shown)


QRemy:

PR #6327, opened 2026-01-22
Unbinned Analysis : Spectral Background interpolated model (5/7)

guillaumegrolleron:
**Description**
This Pr follows #6326. The Background model is needed by the method to estimate the PDF of the background. Currently, it is a simple interpolation of the OFF counts distribution.

PR #6328, opened 2026-01-22
Unbinned Analysis : Base Makers + reflected background update (6/7)

guillaumegrolleron:
**Description**
It follows PR #6327 
This PR adress the implementation of a base EventDatasetMaker. There is also the update of the SafeMaskMaker to be used with EventDataset. 
I also adapted the reflected background maker to perform the analysis directly on events (applied on a EventDataset instead of the SpectrumDataset)


QRemy:

PR #6339, opened 2026-01-28
Add multiple telescope option in `FluxPointsEstimator`

### LINKED ISSUE #6135: Should we allow FluxPoint computation from different instruments
Currently, Gammapy does not allow computation of flux points from different telescopes, 

https://github.com/gammapy/gammapy/blob/ef433bda5a7730b027aec7f6b7b10195e53569fd/gammapy/estimators/points/sed.py#L172-L178

It is a long standing issue which we never resolved, and we should take a decision now @registerrier @bkhelifi @QRemy @adonath @Astro-Kirsty @LauraOlivera 

One issue is that we plot flux points in reco-energy, and the spectral models in true energies on the same plot, and including instruments with different edisp can be confusing. However, similar issue is valid for different event types as well.

We had once decided that we ask users to compute fluxes for each instrument separately.
However, have repeated requests for computing joint flux points.

Astro-Kirsty:
Resolves #6135

A fragment is potentially needed but unsure under what category

### DIFF
--- a/docs/release-notes/6339.feature.rst
+++ b/docs/release-notes/6339.feature.rst
@@ -1,0 +1,1 @@
+Add an boolean argument `allow_multiple_telescopes` which allows the user to compute `~gammapy.estimators.FluxPointsEstimator` for multiple telescopes.


Astro-Kirsty:
Although it was decided in a dev call to change this from an exception to a warning, we now have an issue with proceeding in some cases. 
For example in the test `test_fpe_non_uniform_datasets`, normally it would not make it to the `rows = parallel.run_multiprocessing`, but with only a warning it is allowed to proceed and we have an issue with this. 

In the end, the issue arises from `estimate_flux_point`. After splitting the dataset there are no models. We might need to add another `if` in that case, something like:

```
if datasets_sliced.models is None or len(datasets_sliced.models) == 0:
        log.warning(
            f"No valid models available in range {energy_min}-{energy_max}"
        )
        model = ??
```

bkhelifi:
I still do not understand the logic: either we accept 2 different telescopes and then no warning should be even raised, or we do you accept that and we raise an error.
The proposed intermediated state is weird for me: have a warning when 2 different telescopes are used t4elleing that we can not! (warning message: "All datasets **must** use the same value of the 'TELESCOP' meta keyword")

Astro-Kirsty:
> I still do not understand the logic: either we accept 2 different telescopes and then no warning should be even raised, or we do you accept that and we raise an error. The proposed intermediated state is weird for me: have a warning when 2 different telescopes are used t4elleing that we can not! (warning message: "All datasets **must** use the same value of the 'TELESCOP' meta keyword")

Yes I am also not sure. I prefer the original method of the exception, but it was discussed in a dev call some time ago and decided to be a warning instead:
https://github.com/gammapy/gammapy/issues/6135

Lets discuss again

Astro-Kirsty:
After discussion in the dev call 30/01 we decided that we should keep the current behaviour but have an argument which allows the user to compute FP for multiple telescopes. This is left up to the user as it can only be used in very specific cases and if they know what they are doing. 


**Improvement is need on the docstring still**

gammapy/estimators/points/sed.py
```diff
@@ -84,7 +84,10 @@ class FluxPointsEstimator(FluxEstimator, parallel.ParallelMixin):
         scan_min=0.2, scan_max=5, and scan_n_values = 11. By default, the min and max are not set
         (consider setting them if errors or upper limits computation fails). If a dict is given,
         the entries should be a subset of `~gammapy.modeling.Parameter` arguments.
-
+    allow_multiple_telescopes : bool, optional
+        Whether to allow the computation for different telescopes.
+        This is currently an experimental feature.
```
bkhelifi:
```suggestion
         **WARNING :** This is currently an experimental feature.
```

AtreyeeS:
I think we should at least test that the code executes for multiple telescopes, validity of the results notwithstanding.

gammapy/estimators/points/tests/test_sed.py
```diff
@@ -773,6 +773,27 @@ def test_fpe_non_uniform_datasets():
...
+    datasets = Datasets([dataset_1, dataset_2])
+    datasets.models = SkyModel(spectral_model=PowerLawSpectralModel(), name="test")
+    fpe = FluxPointsEstimator(
+        energy_edges=[1, 2, 4, 10] * u.TeV,
+        source="test",
+        allow_multiple_telescopes=True,
+    )
+    fp = fpe.run(datasets)
+    assert_allclose(fp.counts.data[0], ([[[0]], [[0]]]), rtol=1e-3)
+
```
registerrier, line 795: You might add a test to make sure that `allow_multiple_telescope=False` fails.

Astro-Kirsty: That is already done in the previous test `test_fpe_non_uniform_datasets`, as this was the old behaviour

Issue #6239, opened 2025-11-27
SonarQube Reliability Issues


registerrier:
# SonarQube Reliability Issues

This is a list of SonarQube reliability issues found by CTAO SonarQube.

## BLOCKER

- [x] Remove this unexpected named argument 'size_factor'. See #6353 
  - Occurrences:
    - `gammapy/modeling/models/spatial.py:453` — [View in SonarQube](https://sonar-ctao.zeuthen.desy.de/project/issues?id=gammapy_gammapy_8c1bf7fc-018a-4afd-b216-015191d282fa&issues=f4a30d4c-7be7-49a3-bc56-6501a16a6dc3&open=f4a30d4c-7be7-49a3-bc56-6501a16a6dc3)
    - `gammapy/modeling/models/spatial.py:468` — [View in SonarQube](https://sonar-ctao.zeuthen.desy.de/project/issues?id=gammapy_gammapy_8c1bf7fc-018a-4afd-b216-015191d282fa&issues=a5b651e3-e86e-4eef-a203-846a2fe8ec1b&open=a5b651e3-e86e-4eef-a203-846a2fe8ec1b)

- [x] Remove this "is not" check; it will always be True. See #6354. **This will be removed from the Sonarqube checks**
  - Occurrences:
    - `gammapy/utils/tests/test_cache.py:43` — [View in SonarQube](https://sonar-ctao.zeuthen.desy.de/project/issues?id=gammapy_gammapy_8c1bf7fc-018a-4afd-b216-015191d282fa&issues=807938a6-78e5-4ade-b38c-27e9055d45ca&open=807938a6-78e5-4ade-b38c-27e9055d45ca)
    - `gammapy/utils/tests/test_cache.py:60` — [View in SonarQube](https://sonar-ctao.zeuthen.desy.de/project/issues?id=gammapy_gammapy_8c1bf7fc-018a-4afd-b216-015191d282fa&issues=989628e2-de22-4258-99ff-3c23cc01f954&open=989628e2-de22-4258-99ff-3c23cc01f954)
    - `gammapy/utils/tests/test_cache.py:40` — [View in SonarQube](https://sonar-ctao.zeuthen.desy.de/project/issues?id=gammapy_gammapy_8c1bf7fc-018a-4afd-b216-015191d282fa&issues=bda6df9a-108c-43cb-9412-d5b7d951ee40&open=bda6df9a-108c-43cb-9412-d5b7d951ee40)

## CRITICAL

[...]

PR #6353, opened 2026-01-30
Propagate the argument `size_fraction` into `SpatialModel._to_region_error()`

bkhelifi:
**Description**

This pull request is associated to a "blocker flag" by the SonarCube reliability diagnostic (see the issue #6239).

### DIFF
--- a/gammapy/modeling/models/spatial.py
+++ b/gammapy/modeling/models/spatial.py
@@ -450,23 +450,19 @@ class SpatialModel(ModelBase):
 
         if "all" in which:
             self.plot_position_error(ax, **kwargs_position)
-            try:
-                region = self._to_region_error(size_factor=size_factor)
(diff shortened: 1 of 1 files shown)


QRemy (commented):
Thanks @bkhelifi, does that mean we no longer need the two try/except here : 
https://github.com/gammapy/gammapy/blob/8750a134cf948cbdc5efe9b8999da560f04c8997/gammapy/modeling/models/spatial.py#L450-L473

bkhelifi:
@QRemy: I think that one should keep it. There several spatial models with no `_region_error` function.

QRemy:
> @QRemy: I think that one should keep it. There several spatial models with no `_region_error` function.

It's in the base class so they have it, it will just return None.
We should remove these try/except.

PR #6354, opened 2026-01-30
Fix `is not` statement for SonarQube

Astro-Kirsty:

### DIFF
--- a/gammapy/utils/tests/test_cache.py
+++ b/gammapy/utils/tests/test_cache.py
@@ -36,13 +36,13 @@ def test_dummy_cache():
     v = Dummy2(1, np.array(3))
     w = Dummy2(1, np.array(3))
 
-    assert x == y
-    assert z != y
+    assert x is y  # NOSONAR
+    assert z is not y  # NOSONAR
 
-    assert u != x
-    assert u != v
-    assert v != z
-    assert v == w
+    assert u is not x  # NOSONAR
+    assert u is not v  # NOSONAR
+    assert v is not z  # NOSONAR
+    assert v is w  # NOSONAR
 
     # the cache is not propagated through pickle
     # (implementing that would be more complex)


bkhelifi (approved):
Thanks @Astro-Kirsty

registerrier (commented):

@QRemy can you confirm that this is the expected behaviour? Or does the test really want to make sure x and y are the same object?

QRemy:
`is` and `is not` are the correct behaviour 

assert x is y => we want to check that the objects are really identical in memory which is the expected if the caching works.
assert z is not => we want to check that different subclasses of the mixin are cached separately
 
 You can close this PR, and make SonarCube ignore this test.

Astro-Kirsty:
Thanks @QRemy.

But is this not just the same?
`is` same as `==`
`is not` same as `!=`

QRemy:
> But is this not just the same? `is` same as `==` `is not` same as `!=`

`is` check that the variables point to the same object in memory not their equivalence.
Screenshot 2026-02-02 at 10 31 24

morcuended:
> `is` and `is not` are the correct behaviour
> 
> assert x is y => we want to check that the objects are really identical in memory which is the expected if the caching works. assert z is not => we want to check that different subclasses of the mixin are cached separately
> 
> You can close this PR, and make SonarCube ignore this test.

Thanks for checking @QRemy. I just deactivated the rule in SQ. You can close this then. This will not appear as an issue in the next checks.

registerrier:
Thanks for the feedback @QRemy . 
In general the rule is probably good except in this specific case. Is there a way to ignore this specific set of tests only @morcuended ?

QRemy:
> Thanks for checking @QRemy. I just deactivated the rule in SQ

I don't think it should be deactivated in general but we could add `noqa` or a `NOSONAR` to ignore the lines we don't want to test:
see https://community.sonarsource.com/t/python-issue-suppression-improvements-nosonar-and-new-rules/145017

`assert x is y # NOSONAR(S5655) ignored to test the caching `

registerrier:
Another solution is to test `id(x) == id(y)` instead of `x is y`

morcuended:
It can be flagged with  `# NOSONAR` comment as @QRemy says

QRemy:
I'll open a PR.

Astro-Kirsty:
I can change this PR to be in line, do we want #nosonar solution or testing the id?

gammapy/utils/tests/test_cache.py
QRemy:
```suggestion
    assert z is not y  # NOSONAR(S3403) check caching is not shared across subclasses
```

QRemy:
Resolved by #6366.
Only lines 40, 43, and 60 had to be changed because they are testing that different classes result in different objects in memory, which is always true in principle. But that in that case the caching might break this behaviour because it is done using the same parent class and we want to check that’s not the case.

PR #6355, opened 2026-01-30
Use "np.nonzero" for SonarQube - Part 1

Astro-Kirsty:

It expects to use`np.nonzero` instead of `np.where`, as these are boolean quantities this works well


registerrier (approved):

Issue #6356, opened 2026-01-30
Using `Enum` for option-like arguments ?


MRegeard:
Hi all, 

I was thinking, what if we were to use [`Enum`](https://docs.python.org/3.14/library/enum.html) for our option-like parameters. I am thinking for instance at the `selection_optional` of `FluxPointEstimator`. 

Currently, we are using string for such parameters. This has the drawback that you have to know which string to use beforehand, and therefore would probably have to look at the documentation. And even if you know which "option" you want to use, you have to know it to the character: `lowercase vs uppercase`, `- vs _`, etc. 

Instead, a very common pattern is to use an `Enum`. This has the strong advantage that if the user is using a code editor with an LSP, it can give it the field of the enum as code completion. I also personally find that it usually makes for cleaner code than carrying string around. 

Another common code pattern is to use `match` (or `switch`) statement on `Enum` fields, which usually also makes clean code. Since we now require python `3.10` or above, we can fully support `match` statement. 

I made a small "test" PR for such implementation on `MapDatasetMaker` and `SpectrumDatasetMaker` that can be found here #6357

If we ever were to adopt this proposal, this would imply strong breaking changes. However, as showcase in the example PR, the transition from string to enum would be rather smooth by using `StrEnum` which contain a `str` and can therefore be initialised by their string value:

```python
class Color(StrEnum):
    RED = "red"
    BLUE = "blue"
    GREEN = "green"

color = Color("red") # <- is valid 
```
So we can easily transtion by trying the str on the enum class.

MRegeard:
**List of where this would apply** (WIP)

- [ ] `FluxPointEstimator`

...

PR #6357, opened 2026-01-30
Example PR: Using `Enum` for option-like arguments ?

MRegeard:
This PR is an example implementation of #6356. 

The usage then becomes:

```python
from gammapy.makers import SpectrumDatasetMaker

maker = SpectrumDatasetMaker() # <- Same as before, none -> all selections
maker = SpectrumDatasetMaker(selection=[SpectrumDatasetMaker.selection_enum.COUNTS, SpectrumDatasetMaker.selection_enum.EDISP, SpectrumDatasetMaker.selection_enum.EXPOSURE])
```
I transformed the `selection_available` to `selection_enum` and attached it the `Enum` class so that one can pull the enum field directly. This avoid having to import the `Enum` class. 

At first view it might seems more time consuming to use this that just typing the string, but with code completion it actually faster. This example is also a bit special as a list of "selection" has to be provided (though most of the time, users would leave it to `none` to select every selection possible), but in a case like `selection_optional` of `FluxPointEstimator`, one would only use one option and therefore one enum field.

### DIFF
--- a/gammapy/makers/map.py
+++ b/gammapy/makers/map.py
@@ -9,6 +9,7 @@ from gammapy.irf import EDispKernelMap, PSFMap
 from gammapy.data import Observation
 from gammapy.maps import Map
 from .core import Maker
+from enum import StrEnum
 from .utils import (
     make_counts_rad_max,
     make_edisp_kernel_map,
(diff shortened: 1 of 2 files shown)


QRemy:
If we add this here, it should probably done in other places also for consistency.
Listing all potential cases could be done in an issue and then eventually a PIG if there is many of them.
Closing this PR for now.

PR #6359, opened 2026-01-30
Use "np.nonzero" for SonarQube - Part 3

Astro-Kirsty:

It expects to use`np.nonzero` instead of `np.where`, as these are boolean quantities this works well. 
Part 3 of this

### DIFF
--- a/gammapy/irf/core.py
+++ b/gammapy/irf/core.py
@@ -160,7 +160,7 @@ class IRF(metaclass=abc.ABCMeta):
 
         if np.shape(value) != required_shape:
             raise ValueError(
-                f"data shape {value.shape} does not matchaxes shape {required_shape}"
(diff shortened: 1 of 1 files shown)


gammapy/irf/core.py
```diff
@@ -160,8 +160,7 @@ def data(self, value):
 
         if np.shape(value) != required_shape:
             raise ValueError(
-                f"data shape {value.shape} does not match"
-                f"axes shape {required_shape}"
+                f"data shape {value.shape} does not matchaxes shape {required_shape}"
```
registerrier:
```suggestion
                f"data shape {value.shape} does not match axes shape {required_shape}"
```

registerrier (approved):
Thanks @Astro-Kirsty . No further comment.

PR #6360, opened 2026-01-30
Do not perform equality checks with floating point values

bkhelifi:
**Description**

This pull request is associated to the SonarCube reliability issue: #6239

### DIFF
--- a/gammapy/modeling/models/spectral.py
+++ b/gammapy/modeling/models/spectral.py
@@ -2,6 +2,7 @@
 """Spectral models for Gammapy."""
 
 import logging
+import warnings
 import operator
 import os
 from pathlib import Path


gammapy/datasets/io.py
```diff
@@ -578,7 +578,7 @@ def create_dataset(
             position=counts.geom.center_skydir,
         )
         # check that pdf is well defined (fails if edisp_bins>0 in fermipy)
-        if np.any(psf_r68s.value) == 0.0:
+        if np.isclose(np.any(psf_r68s.value), 0.0):
```
QRemy: even the original line seems wrong and it should rather be :
```suggestion
        if np.any(np.isclose(psf_r68s.value, 0.0)):
```

gammapy/datasets/map.py
```diff
@@ -2974,7 +2974,12 @@ def stack(self, other, nan_to_num=True):
         # For the bins where the stacked OFF counts equal 0, the alpha value is
         # performed by weighting on the total OFF counts of each run
         is_zero = total_off.data == 0
-        acceptance_off.data[is_zero] = total_acceptance.data[is_zero] / average_alpha
+        if average_alpha > 0.0:
+            acceptance_off.data[is_zero] = (
+                total_acceptance.data[is_zero] / average_alpha
+            )
+        else:
+            acceptance_off.data[is_zero] = 0.0
```
QRemy: why this change, it's more than cleanup ?
why the if/else ?
and why you still have a check against 0 `is_zero = total_off.data == 0` in the line above

gammapy/makers/tests/test_map.py
```diff
@@ -368,8 +368,10 @@ def test_interpolate_map_dataset():
         geom=geom_target, energy_axis_true=energy_true, rad_axis=rad_axis, name="test"
     )
     dataset = maker.run(dataset, obs)
-    assert dataset.psf.exposure_map and np.all(dataset.psf.exposure_map.data == 1.0)
-    assert dataset.edisp.exposure_map and np.all(dataset.edisp.exposure_map.data == 0.0)
+    assert dataset.psf.exposure_map
+    assert_array_almost_equal(dataset.psf.exposure_map.data, 1.0)
+    assert dataset.edisp.exposure_map
+    assert_array_almost_equal(dataset.edisp.exposure_map.data, 0.0)
```
QRemy: is this better than assert_allclose ?

gammapy/stats/fit_statistics_jit.py
```diff
@@ -92,7 +92,7 @@ def f_cash_root_jit(x, counts, background, model):
         if model[i] > 0.0:
             if counts[i] > 0.0:
                 denom = x * model[i] + background[i]
-                if denom != 0.0:
+                if not np.isclose(denom, 0.0):
```
QRemy: maybe this should remain in pure python

gammapy/stats/fit_statistics_jit.py
```diff
@@ -135,7 +135,7 @@ def norm_bounds_jit(counts, background, model):
             sn = background[i] / model[i]
             if sn < sn_min_total:
                 sn_min_total = sn
-    if s_model == 0.0:
+    if np.isclose(s_model, 0.0):
```
QRemy: maybe this should remain in pure python

gammapy/modeling/models/spectral.py
```diff
@@ -1630,7 +1630,7 @@ def e_peak(self):
         lambda_ = self.lambda_.quantity
         alpha = self.alpha.quantity
 
-        if index >= 2 or lambda_ == 0.0 or alpha == 0.0:
+        if index >= 2 or np.isclose(lambda_, 0.0) or np.isclose(alpha, 0.0):
```
QRemy: is the default tolerance ok for lamba_ ?

QRemy:
In the code (not tests) for cases where the tolerance might be an issue when testing against 0 because very small values might be possible, should we test np.isfinite(np.log(value)) instead ?

gammapy/modeling/models/spectral.py
bkhelifi: The default is `atol=1e-08`. Do we need more?

gammapy/stats/fit_statistics_jit.py
bkhelifi: I get your point... should we have a manual test on the spirit of that:

```
 def isclose(a, b, rel_tol=1e-09, abs_tol=0.0):
    return abs(a-b) <= max(rel_tol * max(abs(a), abs(b)), abs_tol)
```

if something like:
`if abs(s_model - 0.0) <= 1e-09 * s_model:
`

gammapy/makers/tests/test_map.py
bkhelifi: the data is an array, is not it?

gammapy/datasets/map.py
bkhelifi: Agree. This aims to avoid a division per 0 from `average_alpha`. Is it not wise?

gammapy/stats/fit_statistics_jit.py
QRemy:
```suggestion
                if abs(demon) > 0:
```

gammapy/stats/fit_statistics_jit.py
QRemy:
```suggestion
    if not (abs(s_model) > 0):
```

gammapy/datasets/map.py
QRemy: This change should go in a separate bugfix PR with a regression test to cover it.

gammapy/makers/tests/test_map.py
QRemy: from https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_array_almost_equal.html :
- It is recommended to use one of [assert_allclose](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_allclose.html#numpy.testing.assert_allclose), [assert_array_almost_equal_nulp](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_array_almost_equal_nulp.html#numpy.testing.assert_array_almost_equal_nulp) or [assert_array_max_ulp](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_array_max_ulp.html#numpy.testing.assert_array_max_ulp) instead of this function for more consistent floating point comparisons.
- **The test verifies identical shapes** and that the elements of actual and desired satisfy:
abs(desired-actual) < 1.5 * 10**(-decimal)

so we should keep `assert_allclose` here

gammapy/modeling/models/spectral.py
QRemy: The problem is lambda_ can take any values, if the user decide to to have the unit in /eV then you get very small values.

maybe a cleaner way would be to remove the `if` test, do the computation catching the `ZeroDivisionError` and `RuntimeWarning: overflow encountered in power`, and return nan if any of these appear.

gammapy/stats/fit_statistics_jit.py
```diff
@@ -92,7 +92,7 @@ def f_cash_root_jit(x, counts, background, model):
         if model[i] > 0.0:
             if counts[i] > 0.0:
                 denom = x * model[i] + background[i]
-                if denom != 0.0:
+                if abs(demon) > 0:
```
QRemy:
```suggestion
                if abs(denom) > 0:
```

gammapy/makers/tests/test_map.py
```diff
@@ -1,7 +1,7 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 import pytest
 import numpy as np
-from numpy.testing import assert_allclose
+from numpy.testing import assert_allclose, assert_array_almost_equal
```
QRemy:
```suggestion
from numpy.testing import assert_allclose
```

gammapy/makers/tests/test_map.py
```diff
@@ -368,8 +368,10 @@ def test_interpolate_map_dataset():
         geom=geom_target, energy_axis_true=energy_true, rad_axis=rad_axis, name="test"
     )
     dataset = maker.run(dataset, obs)
-    assert dataset.psf.exposure_map and np.all(dataset.psf.exposure_map.data == 1.0)
-    assert dataset.edisp.exposure_map and np.all(dataset.edisp.exposure_map.data == 0.0)
+    assert dataset.psf.exposure_map
+    assert_array_almost_equal(dataset.psf.exposure_map.data, 1.0)
```
QRemy:
```suggestion
    assert_allclose(dataset.psf.exposure_map.data, 1.0)
```

gammapy/makers/tests/test_map.py
QRemy:
```suggestion
    assert_allclose(dataset.edisp.exposure_map.data, 0.0)
```

gammapy/modeling/models/spectral.py
```diff
@@ -1630,10 +1630,19 @@ def e_peak(self):
...
+        except (ZeroDivisionError, RuntimeWarning, OverflowError, ValueError):
             return np.nan * reference.unit
-        else:
-            return np.power((2 - index) / alpha, 1 / alpha) / lambda_
+
+        try:
+            en = np.power((2 - index) / alpha, 1 / alpha) / lambda_
+        except ZeroDivisionError:
+            en = np.nan * reference.unit
+        return en
```
QRemy: duplicated ?

gammapy/irf/core.py
```diff
@@ -176,7 +176,7 @@ def interp_missing_data(self, axis_name):
         scale = interpolation_scale(values_scale)
 
         axis = self.axes.index(axis_name)
-        mask = ~np.isfinite(data) | (data == 0.0)
+        mask = ~np.isfinite(data) | np.isclose(data, 0.0)
```
Astro-Kirsty: This will check that is it the 'same' within some tolerance. Is that really what we want?
Or should it just be 0?
i.e. small non-zero values will be silently masked

gammapy/catalog/hawc.py
```diff
@@ -252,7 +252,7 @@ def _info_spectrum_one(self, idx):
 
     @property
     def is_pointlike(self):
-        return self.data["spec0_radius"] == 0.0
+        return np.isclose(self.data["spec0_radius"], 0.0)
```
Astro-Kirsty: Do we want that to be a float? i.e. the 'same' within tolerance or just '0'

gammapy/modeling/models/spectral.py
```diff
@@ -1630,10 +1630,16 @@ def e_peak(self):
         lambda_ = self.lambda_.quantity
         alpha = self.alpha.quantity
 
-        if index >= 2 or lambda_ == 0.0 or alpha == 0.0:
+        try:
```
Astro-Kirsty, line 1585: Can you explain a bit why all this extra code is needed?

gammapy/modeling/scipy.py
```diff
@@ -168,7 +168,7 @@ def stat_profile_ul_scipy(
         )
 
     mask_valid = np.isfinite(stat_scan) & np.isfinite(value_scan)
-    if mask_valid.sum() == 0.0:
+    if np.isclose(mask_valid.sum(), 0.0):
```
registerrier:
```suggestion
    if not np.any(mask_valid):
```
Here this is a boolean array, we should not use floating point comparisons.

gammapy/stats/fit_statistics_jit.py
```diff
@@ -92,7 +92,7 @@ def f_cash_root_jit(x, counts, background, model):
         if model[i] > 0.0:
             if counts[i] > 0.0:
                 denom = x * model[i] + background[i]
-                if denom != 0.0:
+                if abs(denom) > 0:
```
registerrier, line 95: does jitted code need to be adapted too?

gammapy/modeling/models/spectral.py
```diff
@@ -1630,10 +1630,16 @@ def e_peak(self):
         lambda_ = self.lambda_.quantity
         alpha = self.alpha.quantity
 
-        if index >= 2 or lambda_ == 0.0 or alpha == 0.0:
```
registerrier, line 1584: In this cases where we are just safe guarding against division by zero, I think `==0` is acceptable and probably the correct thing to do.

gammapy/modeling/models/tests/test_spectral.py
```diff
@@ -796,7 +796,7 @@ def test_template_spectral_model_evaluate_tiny():
         values[mask] / values.max(), result[mask].value / values.max()
     )
     mask = abs(result.value) - tiny <= tiny
-    assert np.all(result[mask] == 0.0)
+    assert np.allclose(result[mask], 0.0)
```
registerrier: should there be a low tolerance here as the test is checking for very low values, e.g. absolute tolerance =tiny?

gammapy/maps/tests/test_core.py
```diff
@@ -601,7 +601,7 @@ def test_resample_weights():
     )
 
     map2 = map1.resample(geom2, weights=np.zeros(npix1), preserve_counts=False)
-    assert np.sum(map2.data) == 0.0
+    assert_allclose(np.sum(map2.data), 0.0)
```
registerrier: In this case, should we test `assert_allclose(map2.data, 0.0)`

gammapy/irf/core.py
registerrier: Yes here too, 0.0 is the exact value that we want to test as this is considered a missing value. So I think we have to keep as is with a possible `#NOSONAR : here we explicitly reject NaN, Inf or 0.0`

gammapy/datasets/utils.py
```diff
@@ -293,7 +293,9 @@ def create_map_dataset_from_dl4(data, geom=None, energy_axis_true=None, name=Non
     if not dataset.background:
         dataset.background = Map.from_geom(geom, data=0.0)
 
-    if dataset.edisp.exposure_map and np.all(dataset.edisp.exposure_map.data) == 0.0:
+    if dataset.edisp.exposure_map and np.all(
+        np.isclose(dataset.edisp.exposure_map.data, 0.0)
```
registerrier: Here I am not sure this should be changed as we are not testing if the exposure is small but if there is any exposure.

gammapy/catalog/hawc.py
registerrier: Yes, here we should keep as is and add  a `#NOSONAR `

gammapy/irf/core.py
bkhelifi: Yes, but because of floating point approximation 0 might not be exactly 0. Do you wish then to have a tiny tolerance, e.g. 1e-16?

gammapy/modeling/models/spectral.py
bkhelifi: Again, lambda_ == 0.0 or alpha == 0.0 is not the proper test because 0 can be 10-32 because of the floating point error. So, Quentin suggest this algorithm instead of adding the np.isclose in the if.

gammapy/stats/fit_statistics_jit.py
bkhelifi: float point.... the condition might be always true

gammapy/irf/core.py
QRemy: In that case we could assume that the zeros we want to remove are not the results of an operation but exactly representable zeros (as the ones given by float(0) or np.zeros) because the IRFs where just filled with 0 where no valid values could be computed.  
Then we could do as @registerrier suggested : 
`mask = ~np.isfinite(data) | (data == 0.0) #NOSONAR (S1244): explicit check for exactly representable zeros`

gammapy/catalog/hawc.py
QRemy: same as above

gammapy/datasets/utils.py
QRemy: here also we want check that the map is empty as created so it should be exact zeros.

QRemy:
Coverage is low only because jit code is not analysed, that's not new.
Comments are implemented so I'm going to merge this now to avoid more merge conflict (just had to rebase again).

PR #6365, opened 2026-02-02
SonarQube remove self-assignment and clean up for docs build

Astro-Kirsty:
This pull request is associated to a "MAJOR" flag by the SonarCube reliability diagnostic (see the issue https://github.com/gammapy/gammapy/issues/6239).

I also took the chance to fix some docs errors in the build

### DIFF
--- a/examples/models/spectral/plot_logparabola2.py
+++ b/examples/models/spectral/plot_logparabola2.py
@@ -13,7 +13,7 @@ It is defined by the following equation:
           - \alpha - \beta \log{ \left( \frac{E}{E_s} \right) }
         }
 
-Note that :math:`log` refers to the natural logarithm.
+Note that :math:`log` refers to the natural logarithm. 
 If you have parametrization based on :math:`log_{10}` you can use the
 :func:`~gammapy.modeling.models.LogParabola2SpectralModel.from_log10` method.
 


gammapy/stats/fit_statistics.py
```diff
@@ -470,6 +470,7 @@ class GaussianPriorPenalty(FitStatisticPenalty):
     r"""Penalty based on a multivariate Gaussian prior.
 
     This implements a quadratic penalty of the form:
+
```
QRemy: This is not on 2.0.x can't be backported.

gammapy/utils/docs.py
```diff
@@ -169,7 +169,6 @@ def run(self):
         """Replace placeholders with given variables."""
         app = self.state.document.settings.env.app
         new_content = []
-        self.content = self.content
```
QRemy, line 172: Should this be backported ?

Astro-Kirsty:
I'm not convinced we should backport any of it. Please tell me if otherwise

registerrier:
I think we should backport corrections of sonarqube issues. If you could split in two it would be simpler.

examples/models/spectral/plot_logparabola2.py
```diff
@@ -13,7 +13,7 @@
           - \alpha - \beta \log{ \left( \frac{E}{E_s} \right) }
         }
 
-Note that :math:`log` refers to the natural logarithm. 
+Note that :math:`log` refers to the natural logarithm.
```
Astro-Kirsty:
```suggestion
Note that :math:`log` refers to the natural logarithm. 
```

Astro-Kirsty:
It should be backportable now!

QRemy (approved):
Thanks @Astro-Kirsty, merging now.

PR #6371, opened 2026-02-02
MAJOR fixes for Sonar Qube

Astro-Kirsty:
A number of the tests were missing the assert or had too many '='. Please advise if I made an error in how things were intended to be tested.

Associated to the clean up for #6239

### DIFF
--- a/gammapy/makers/tests/test_safe.py
+++ b/gammapy/makers/tests/test_safe.py
@@ -203,6 +203,7 @@ def test_safe_mask_maker_edisp_bias(dataset, observation_cta_1dc):
     assert_allclose(mask_edisp_bias.data.sum(), 1815)
 
 
+@requires_data()
 def test_safe_mask_maker_errors(observation_cta_1dc):
     with pytest.raises(
         ValueError,


gammapy/datasets/tests/test_simulate.py
```diff
@@ -344,7 +344,7 @@ def test_sample_coord_time_energy_random_seed(
 def test_sample_coord_time_energy_unit(dataset, energy_dependent_temporal_sky_model):
     sampler = MapDatasetEventSampler(random_state=1)
 
-    energy_dependent_temporal_sky_model.temporal_model.map._unit == "cm-2 s-1 TeV-1"
+    energy_dependent_temporal_sky_model.temporal_model.map._unit = "cm-2 s-1 TeV-1"
```
QRemy, line 346: Any idea why these tests works with or without these changes ?
@fabiopintore @AtreyeeS you worked on that no ?

Astro-Kirsty: Well the above with `==` is not a test, it is simply a statement that would print True. If we want that as a test we need to add an assert option. 
It was not clear whether the point was to test using assert or to just assign the unit

AtreyeeS: I think that the unit is correctly set on these models already (in line 125 of this file), so that assignment was useless (probably remnant of old tests using SDC simulations) I think we should just remove this line as this function should test the sampler, and not the model created beforehand

gammapy/datasets/tests/test_simulate.py
```diff
@@ -321,7 +321,7 @@ def test_sample_coord_time_energy_random_seed(
 ):
     sampler = MapDatasetEventSampler(random_state=2)
 
-    energy_dependent_temporal_sky_model.temporal_model.map._unit == ""
+    energy_dependent_temporal_sky_model.temporal_model.map._unit = ""
```
AtreyeeS:
```suggestion
```

gammapy/makers/safe.py
```diff
@@ -102,7 +102,7 @@ def __init__(
             )
 
         if irfs not in ["DL3", "DL4"]:
-            ValueError(
+            raise ValueError(
```
AtreyeeS, line 105: Can you test that this value error is actually being raised?

gammapy/maps/wcs/tests/test_geom.py
```diff
@@ -579,7 +579,7 @@ def test_irregular_geom_equality():
     geom1 = WcsGeom.create(skydir=(0, 0), npix=10, binsz=binsizes, axes=[axis])
 
     with pytest.raises(NotImplementedError):
-        geom0 == geom1
+        geom0.__eq__(geom1)
```
AtreyeeS, line 582: I guess the point here was to check for the `==` operation, eg as in line 571
This has not yet been implemented for irregular geoms

gammapy/modeling/models/tests/test_cube.py
```diff
@@ -272,7 +272,6 @@ def test_background_slice(background):
     bkg1 = TemplateNPredModel(background)
     e_edges = background.geom.axes[0].edges
     bkg1_slice = bkg1.slice_by_energy(e_edges[0], e_edges[1])  # 1 bin slice
-    assert bkg1_slice.name == bkg1_slice.name
```
AtreyeeS, line 275: assert bkg1_slice.name == bkg1.name

Astro-Kirsty: This will not work because the name of `bkg1` will not be the same as the slices  (unless we set that)

bkhelifi:
The code coverage might be improved with this PR ;)

gammapy/makers/safe.py
Astro-Kirsty: Yes, done!

gammapy/maps/wcs/tests/test_geom.py
Astro-Kirsty: But this is for raising a warning so the above logic on line 571 does not apply

gammapy/modeling/models/tests/test_cube.py
AtreyeeS: Ah yes it makes a random string. Then fine to remove

gammapy/maps/wcs/tests/test_geom.py
AtreyeeS: Sorry I don't understand. This should raise an error, not a warning

gammapy/makers/tests/test_safe.py
```diff
@@ -203,6 +203,20 @@ def test_safe_mask_maker_edisp_bias(dataset, observation_cta_1dc):
     assert_allclose(mask_edisp_bias.data.sum(), 1815)
 
 
+def test_safe_mask_maker_errors(observation_cta_1dc):
```
QRemy, line 207:
```suggestion
@requires_data()
def test_safe_mask_maker_errors(observation_cta_1dc):
```

PR #6378, opened 2026-02-04
add GTI to FermipyDataset

ToraTherese:
**Description**

Motivation: To get some time information for FermiDatasets.
I've done this by getting the GTI for FermiDataset from ft1_0{str(file_id)}.fits files already generated by fermipy gtmktime.
The gti_file can here be set manually for FermipyDatasetsReader.create_dataset, with defualt not setting it.
For FermipyDatasetsReader.read, it is set to read in the gti_file that should already be automatically created.

This pull request ...

**Dear reviewer**

Would need to add the ft1_0{str(file_id)}.fits files that should already be created together with the current fermipy files in gammapy-data for the tutorials to still run smoothly.

I have tested it with the "Fermi-LAT with Gammapy" tutorial, and it works fine when including the ft1_0{str(file_id)}.fits files corresponding to the tutorial data provided by @QRemy .

### DIFF
--- a/gammapy/datasets/tests/test_simulate.py
+++ b/gammapy/datasets/tests/test_simulate.py
@@ -321,7 +321,6 @@ def test_sample_coord_time_energy_random_seed(
 ):
     sampler = MapDatasetEventSampler(random_state=2)
 
-    energy_dependent_temporal_sky_model.temporal_model.map._unit == ""
     dataset.models = energy_dependent_temporal_sky_model
     evaluator = dataset.evaluators["test-source"]
 
@@ -344,7 +343,7 @@ def test_sample_coord_time_energy_random_seed(
 def test_sample_coord_time_energy_unit(dataset, energy_dependent_temporal_sky_model):
     sampler = MapDatasetEventSampler(random_state=1)
 
-    energy_dependent_temporal_sky_model.temporal_model.map._unit == "cm-2 s-1 TeV-1"
+    energy_dependent_temporal_sky_model.temporal_model.map._unit = "cm-2 s-1 TeV-1"
     energy_dependent_temporal_sky_model.spectral_model.parameters[0].unit = ""
     dataset.models = energy_dependent_temporal_sky_model
     evaluator = dataset.evaluators["test-source"]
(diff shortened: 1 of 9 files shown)


QRemy:
Thanks @ToraTherese, this requires https://github.com/gammapy/gammapy-data/pull/83

gammapy/datasets/io.py
```diff
@@ -522,6 +522,7 @@ def create_dataset(
         edisp_file,
         isotropic_file=None,
         edisp_bins=0,
+        gti_file=None,
```
QRemy: New arguments have to be added in last position for backward compatibility.

AtreyeeS (approved):
Thanks @ToraTherese !

QRemy:
Could you add some assert in the `test_fermipy_datasets_reader` function of  `gammapy/gammapy/datasets/tests/test_io.py` to check that gtis are not None when using the `read` method and `None` if not given in `create_dataset` 
Could you also add assert to check that the time values are the ones expected ?

gammapy/datasets/io.py
```diff
@@ -545,6 +546,8 @@ def create_dataset(
             and  a value edisp_bins>0 should be set here in order to apply the energy dispersion correctly.
             With a binning of 8 to 10 bins per decade, it is recommended to use edisp_bins ≥ 2
             (See https://fermi.gsfc.nasa.gov/ssc/data/analysis/documentation/Pass8_edisp_usage.html)
+        gti_file : str
+            GTI file path. Default is None
         name : str, optional
             Dataset name. The default is None, and the name is randomly generated.
```
QRemy:
```suggestion
       name : str, optional
           Dataset name. The default is None, and the name is randomly generated.
       gti_file : str, optional
            GTI file path. Default is None

```

bkhelifi:
There is an easy SonarCube issue to be fixed. Otherwise, OK

ToraTherese:
> Can be good to add a test ensuring that the GTI is set

@AtreyeeS  I'm sorry, I'm not sure I understand what you're suggesting. Do you mean something else than the assert tests added to test_fermipy_datasets_reader?

PR #6381, opened 2026-02-07
Evaluator: add GPU acceleration to apply_psf with benchmark

yaochengchen:
Update `apply_psf` in `gammapy/datasets/evaluator.py` to use GPU when available
to speed up PSF application. When GPU is not available, it naturally falls back
to the original CPU path.

Add a benchmark test,
`gammapy/datasets/tests/bench_evaluator_apply_psf.py`, to evaluate the
performance improvement. The GPU path achieves several-times speedup compared
to the CPU implementation.

**Description**

This pull request adds optional GPU acceleration to the
`Evaluator.apply_psf` workflow. The existing CPU implementation is preserved
and used automatically when GPU support is not available, ensuring full
backwards compatibility.

A new benchmark test is included to quantify the performance impact of the
change and to allow comparison between the CPU and GPU implementations.

**Dear reviewer**

This PR is ready for review.

- The CPU path is fully covered by the existing test suite and CI.
- The GPU path has been tested on a GPU-enabled machine.
- The benchmark is intended for performance evaluation only and does not
  affect CI execution.

Feedback on the implementation approach and benchmark setup is very welcome.


yaochengchen:
As a side note, I’ll be joining CTAO as a postdoc and have been working extensively on GPU acceleration. I’d be happy to gradually contribute more GPU support to Gammapy if this direction is aligned with the project roadmap.

Longer term, this could also include native CUDA kernel implementations, beyond relying on higher-level frameworks, but that would of course be discussed separately.

gammapy/datasets/evaluator.py
```diff
@@ -352,8 +363,103 @@ def apply_exposure(self, flux):
         return Map.from_geom(self.geom, data=npred, unit="")
 
     def apply_psf(self, npred):
-        """Convolve npred cube with PSF."""
-        return npred.convolve(self.psf)
+        # return npred.convolve(self.psf)
+        """Convolve npred cube with PSF (GPU optimized with grouped conv2d)."""
```
QRemy, line 358: the gpu convolution could be moved into a dedicated function and go to gammapy.utils.parallel

gammapy/datasets/tests/bench_evaluator_apply_psf.py
```diff
@@ -0,0 +1,201 @@
+#!/usr/bin/env python
+"""
+Benchmark MapEvaluator.apply_psf in two fixed modes (one run prints both):
```
QRemy: This could go in [gammapy-benchmark](https://github.com/gammapy/gammapy-benchmarks/tree/main/benchmarks), while in gammapy one of the existing test for the MapEvaluator should also be run with the gpu alternative to check it gives equivalent results.

gammapy/datasets/evaluator.py
```diff
@@ -93,6 +93,17 @@ def __init__(
...
+        self.gpu_device = None
+
+        try:
+            import torch
+
+            if torch.cuda.is_available():
+                self.gpu_device = torch.device("cuda")
+
+        except ImportError:
+            pass
```
QRemy: I would move that to gammapy.utils.parallel as two function `is_cuda_available()`  and `get_gpu_device()` similar to what we have for multiprocessing and ray.

QRemy (commented):
Thanks @yaochengchen, looks interesting, could you share the results of your benchmark in a comment ?

Few years a ago we had a coding spring discussing a lot about using JAX https://github.com/gammapy/gammapy-meetings/blob/master/coding-sprints/2024-06-Heidelberg/slides/gammapy-jax.pdf
https://github.com/gammapy/gammapy-meetings/tree/master/coding-sprints/2024-06-Heidelberg/notes
Any comments on why we should go for torch ?

Regarding this PR, for now I have only few comments about the location of the code :

bkhelifi:
In addition to the computation time, one should check also the RAM. Can we monitor it?

gammapy/datasets/evaluator.py
yaochengchen: Thanks for the suggestion!
I’ve moved the GPU convolution into a dedicated helper function in gammapy.utils.parallel and updated Evaluator.apply_psf accordingly.

gammapy/datasets/tests/bench_evaluator_apply_psf.py
yaochengchen: Thanks for the suggestion.
I’ve added a test in test_evaluator.py to check that the GPU and CPU implementations give equivalent results, as suggested.
I’ve kept bench_evaluator_apply_psf.py in this PR for now, as I plan to extend GPU support to additional functions shortly. In a follow-up PR, I’ll move the benchmarks to gammapy-benchmark as recommended.

gammapy/datasets/evaluator.py
yaochengchen: Done — the CUDA checks have been moved to gammapy.utils.parallel as helper functions, following the existing parallel utilities.

yaochengchen:
I’ve run the benchmark locally on an NVIDIA H100 GPU to assess both performance and memory usage.
Summary of results:
GPU implementation shows a consistent ~4–6× speedup compared to the CPU path, depending on kernel size and data layout.
CPU/GPU outputs are numerically consistent (allclose=True with rtol=1e-5, atol=1e-6).
GPU memory usage peaks at ~0.9 GB VRAM for the tested configuration (32 energy bins, 1000×1000 spatial grid).
CPU RSS peak reaches ~1702.3 MiB during convolution.

(gammapy-dev) [cyc@pu1 gammapy]$ python gammapy/datasets/tests/bench_evaluator_apply_psf.py
=== Fixed benchmark settings ===
RUNS=50, WARMUP=5
WIDTH_DEG=10.0, BINSZ_DEG=0.01
N_E=32, E_K=32, PSF_SIGMA_DEG=0.05

=== Compare settings ===
COMPARE_DTYPE=float64
RTOL=1e-05, ATOL=1e-06
REL_THRESHOLD=0.0001

=== Kernel sweep settings ===
KERNEL_SIZES=[11, 21, 31, 41, 51, 61, 71, 81, 91, 101, 111]
RUNS_SWEEP=10, WARMUP_SWEEP=2

Torch: 2.10.0+cu128
CUDA available: True
CUDA device: NVIDIA H100 PCIe

=== [Mode A] cube npred (E,Y,X) ===
  npred.shape : (32, 1000, 1000)
  kernel.shape: (32, 51, 51)

  [CPU]
    mean  = 382.531 ms
    stdev = 1.492 ms
    out.shape=(32, 1000, 1000) dtype=float32
    RSS (CPU)    : 1123.0 -> 1259.0 MiB (delta=+136.0 MiB)

  [GPU]
    mean  = 74.233 ms
    stdev = 0.112 ms
    out.shape=(32, 1000, 1000) dtype=float32
    VRAM peaks   : allocated=931.2 MiB, reserved=942.0 MiB

  [CPU vs GPU output diff]
    compare_dtype  = float64
    max_abs        = 4.235165e-21
    rms_abs        = 2.093291e-23
    max_rel_masked = 0.000000e+00 (mask: |CPU|>0.0001, fraction=0.000000)
    allclose       = True (rtol=1e-05, atol=1e-06)
    global_scale   = 1.403704e-14, eps_used=1.403704e-26

=== [Mode B] image_broadcast npred (Y,X) + kernel (Ek,Ky,Kx) ===
  npred.shape : (1000, 1000)
  kernel.shape: (32, 51, 51)

  [CPU]
    mean  = 375.961 ms
    stdev = 1.412 ms
    out.shape=(32, 1000, 1000) dtype=float32
    RSS (CPU)    : 1564.2 -> 1702.3 MiB (delta=+138.2 MiB)

  [GPU]
    mean  = 65.772 ms
    stdev = 0.192 ms
    out.shape=(32, 1000, 1000) dtype=float32
    VRAM peaks   : allocated=931.2 MiB, reserved=942.0 MiB

  [CPU vs GPU output diff]
    compare_dtype  = float64
    max_abs        = 5.364418e-07
    rms_abs        = 5.752013e-08
    max_rel_masked = 2.405416e-03 (mask: |CPU|>0.0001, fraction=0.822293)
    allclose       = True (rtol=1e-05, atol=1e-06)
    global_scale   = 9.982443e-01, eps_used=9.982443e-13


=== Kernel sweep: Mode A (cube) (CPU vs GPU speedup) ===
kernel_req  kernel_act      CPU ms      GPU ms     speedup
     11x11       11x11     307.962      74.266        4.15
     21x21       21x21     308.953      77.911        3.97
     31x31       31x31     367.084      84.288        4.36
     41x41       41x41     368.634      74.239        4.97
     51x51       51x51     382.660      74.069        5.17
     61x61       61x61     384.421      76.665        5.01
     71x71       71x71     382.624      77.885        4.91
     81x81       81x81     386.593      74.628        5.18
     91x91       91x91     427.384      77.987        5.48
   101x101     101x101     428.593      75.178        5.70
   111x111     111x111     433.623      75.237        5.76

=== Kernel sweep: Mode B (image_broadcast) (CPU vs GPU speedup) ===
kernel_req  kernel_act      CPU ms      GPU ms     speedup
     11x11       11x11     314.917      65.750        4.79
     21x21       21x21     317.409      69.914        4.54
     31x31       31x31     381.326      75.881        5.03
     41x41       41x41     382.683      66.363        5.77
     51x51       51x51     384.120      66.124        5.81
     61x61       61x61     383.482      69.359        5.53
     71x71       71x71     384.355      69.613        5.52
     81x81       81x81     383.148      66.817        5.73
     91x91       91x91     434.111      70.096        6.19
   101x101     101x101     421.780      66.889        6.31
   111x111     111x111     435.225      67.296        6.47

yaochengchen:
> Thanks @yaochengchen, looks interesting, could you share the results of your benchmark in a comment ?
> 
> Few years a ago we had a coding spring discussing a lot about using JAX https://github.com/gammapy/gammapy-meetings/blob/master/coding-sprints/2024-06-Heidelberg/slides/gammapy-jax.pdf https://github.com/gammapy/gammapy-meetings/tree/master/coding-sprints/2024-06-Heidelberg/notes Any comments on why we should go for torch ?
> 
> Regarding this PR, for now I have only few comments about the location of the code :

Thanks for the question. I’ve shared the benchmark results in the previous comment.
Since Gammapy is currently largely CPU-based, this PR accelerates only a single step on the GPU. As a result, the observed speedup is limited by CPU↔GPU data transfers, and we see a moderate ~4–6× improvement for this stage.
From a performance perspective, the difference between JAX and PyTorch for this type of workload is relatively small compared to the overall overheads. In my experience, switching from PyTorch to JAX alone would not change the order of magnitude of the speedup observed here.
If significantly larger gains are required, they would likely come from lower-level GPU approaches, such as custom CUDA kernels (e.g. kernel fusion, shared memory usage) or higher-level CUDA-based frameworks like cuTile, which allow relatively easy programming from Python while achieving substantially higher performance than framework-level abstractions.

yaochengchen:
> In addition to the computation time, one should check also the RAM. Can we monitor it?

Thanks for the suggestion.
I’ve included memory usage measurements in the benchmark results shared in my previous comment.
For the tested configuration (32 energy bins, 1000×1000 spatial grid):
GPU memory usage peaks at ~0.9 GB VRAM.
CPU RSS peak reaches ~1702.3 MiB during the convolution.

Astro-Kirsty:
I only had a quick look but for now I would suggest that we move the bench mark test to the correct repo, as Quentin suggests. Most of the comments can also be removed.

As you are a new contributor to gammapy I suggest you read the relevant documentation carefully:
https://docs.gammapy.org/2.0.1/development/intro.html

Thanks @yaochengchen!

gammapy/datasets/tests/test_evaluator.py
```diff
@@ -277,3 +277,87 @@ def test_norm_only_changed():
     spectral_model.amplitude.value *= 2
     spectral_model.index.value *= 2
     assert not evaluator.parameter_norm_only_changed
+
+
```
QRemy, line 281:
```suggestion
@requires_dependency("torch")
```

check if there is `from gammapy.utils.testing import requires_dependency` at the begining

gammapy/datasets/tests/test_evaluator.py
```diff
@@ -277,3 +277,87 @@ def test_norm_only_changed():
...
+def test_apply_psf_cpu_gpu_equivalent_mode_b():
+    """Check GPU path gives equivalent results to CPU path for Mode B (image_broadcast).
+
+    - npred is (Y, X)
+    - PSF kernel is (Ek, Ky, Kx)
+    - Skip automatically if torch/CUDA not available.
+    """
+    torch = pytest.importorskip("torch")
+    if not torch.cuda.is_available():
+        pytest.skip("CUDA not available -> skip CPU/GPU equivalence test")
```
QRemy:
```suggestion
    """
```

gammapy/datasets/tests/test_evaluator.py
QRemy: you will also have to add `torch` to `gammapy/pyproject.toml` in `all`, `all_no_ray`, and `cov` lists (like numba).

yaochengchen:
> I only had a quick look but for now I would suggest that we move the bench mark test to the correct repo, as Quentin suggests. Most of the comments can also be removed.
> 
> As you are a new contributor to gammapy I suggest you read the relevant documentation carefully: https://docs.gammapy.org/2.0.1/development/intro.html
> 
> Thanks @yaochengchen!

Ok, Thanks for the suggestion, I have moved it to gammapy-benchmarks.

gammapy/datasets/tests/test_evaluator.py
yaochengchen: Thanks! I have updated it according to your suggestion.

gammapy/datasets/tests/test_evaluator.py
yaochengchen: Thanks! I have updated it according to your suggestion.

QRemy:
Before merging any gpu code we have to setup a specific workflow to test it (ideally without paying).
@bkhelifi @registerrier could it be hosted at APC or in Lyon ?
https://docs.github.com/en/enterprise-cloud@latest/actions/concepts/runners/larger-runners
https://betatim.github.io/posts/github-action-with-gpu/
https://docs.github.com/fr/actions/concepts/runners/self-hosted-runners

bkhelifi:
> Before merging any gpu code we have to setup a specific workflow to test it (ideally without paying). @bkhelifi @registerrier could it be hosted at APC or in Lyon ? https://docs.github.com/en/enterprise-cloud@latest/actions/concepts/runners/larger-runners https://betatim.github.io/posts/github-action-with-gpu/ https://docs.github.com/fr/actions/concepts/runners/self-hosted-runners

@natthan-pigoux: do you know whether one can test that in Lyon? IF yes, can you help us to make such installation in a short time?
@jlenain has already installed Gammapy in Lyon. And he might be very helpful.

natthan-pigoux:
Hi @bkhelifi , sure I can try to help you test that in Lyon, but I'm on vacation this week... I'll be available next week if it's not too late for you

facero:
A more general question is whether this can in general be applied to all convolution ?
PSF is the 2D convolution but could it be applied also spectral 1d convultion (the Energy dispersion) and would it also provide a speedup in 1D ?

I'm asking as we've been using gammapy also for some X-ray analysis where the Edisp can be 10^3 x 10^3 (or even 10^4 x 10^4) where the convolution step can be slow.

yaochengchen:
> A more general question is whether this can in general be applied to all convolution ? PSF is the 2D convolution but could it be applied also spectral 1d convultion (the Energy dispersion) and would it also provide a speedup in 1D ?
> 
> I'm asking as we've been using gammapy also for some X-ray analysis where the Edisp can be 10^3 x 10^3 (or even 10^4 x 10^4) where the convolution step can be slow.

你可以这样更简短、直接一些：

> Yes, this task sounds very suitable for GPU acceleration. I would be happy to contribute to Gammapy by replacing suitable CPU-based functions with GPU implementations to speed up the calculations.

natthan-pigoux:
After discussing with @registerrier and @bkhelifi , this can be done in two steps:

1. run this test "by hand" with a SLURM job on a GPU at CC-IN2P3 (or elsewhere?)-> I can help setup the job or actually do it.

2. if more of these GPU implementations and benchmarks need to be done in the future, we shall probably think of a more automatized way of running these. Organizing a discussion on the subject with CC people is I think a good start, starting by opening a ticket on the user support.

facero:
Just for reference in case you want to test perf on your Mac.

Torch on Mac Silicon (M-series) can use the GPU in a transparent manner for the user using the Metal Performance Shaders ([doc here](https://developer.apple.com/documentation/metalperformanceshaders)).
You just need to install torch (in an arm64 conda python env)  and then you can test whether the GPU is recognized doing :

```
import torch
device = torch.device("mps" if torch.backends.mps.is_available()
                      else "cpu")
print(f"PyTorch running on: {device}")
```

PR #6383, opened 2026-02-08
Correct off-by-one error in `WcsGeom.pix_to_idx` clipping

yaochengchen:
Description This pull request fixes an off-by-one error in WcsGeom.pix_to_idx when clip=True is used.

Currently, the implementation uses np.clip(idx, 0, npix_). In a 0-indexed system, a dimension with N pixels (npix_) has a valid index range of [0,N−1]. The previous code allowed an index of N to be returned for coordinates exactly at or slightly beyond the map boundary, which leads to IndexError when accessing the underlying data array or potential memory issues in GPU kernels.

Changes:

Updated the upper bound of np.clip from npix_ to npix_ - 1 within the WcsGeom.pix_to_idx method.

This ensures that edge pixels are correctly mapped to the last valid bin.


yaochengchen:
in gammapy/maps/axes.py, def pix_to_idx, it is: idx = np.clip(pix, 0, self.nbin - 1), so this one should also be npix_ - 1

bkhelifi:
If it is an error, do you have a small code that shows it?
It will serve also to make a dedicated test function

yaochengchen:
> If it is an error, do you have a small code that shows it? It will serve also to make a dedicated test function

Thanks for the question — yes, I added a minimal reproducible test triggered by real WCS edge coordinates, not by manually constructed pixel indices.
The test uses get_coord(mode="edges") to take a sky coordinate exactly on the footprint boundary, then converts it back with coord_to_pix(). Due to WCS projection and floating-point rounding, the resulting pixel coordinate can be close to N - 0.5, which is then rounded to N by pix_tuple_to_idx.
With the original implementation (np.clip(idx, 0, npix_)), this allows an out-of-bounds index to pass through. The failure below is observed with the current code and is fixed by changing the upper bound to npix_ - 1:

_________________________________________ test_wcsgeom_pix_to_idx_clip_edge_from_wcs_edges __________________________________________

    def test_wcsgeom_pix_to_idx_clip_edge_from_wcs_edges():
        geom = WcsGeom.create(
            npix=(10, 10),
            binsz=1.0,
            proj="TAN",
            frame="icrs",
            skydir=(0, 0),
        )
    
        sky_edges = geom.to_image().get_coord(mode="edges").skycoord
        ny, nx = sky_edges.shape
        sky_on_right_edge = sky_edges[ny // 2, -1]
    
        pix = geom.coord_to_pix(sky_on_right_edge)
        idx = geom.pix_to_idx(pix, clip=True)
    
        # After clipping, the index should be on the edge of the map
>       assert idx[0] == geom.npix[0] - 1
E       assert array([10]) == array([9])
E         
E         Use -v to get more diff

gammapy/maps/wcs/tests/test_geom.py:761: AssertionError

PR #6384, opened 2026-02-08
Cleanup`TimeMapAxis.pix_to_coord` validity mask

yaochengchen:
- Fix a bug where np.logical_and(a, b, c) mistakenly treated the 3rd argument
  as the ufunc 'out' parameter instead of a third condition, potentially
  skipping the isfinite(idx) check.
- Replace with a correct boolean mask combination:
  (cond1) & (cond2) & np.isfinite(idx).
- Add a regression test that makes np.isfinite() return a read-only array to
  ensure the old implementation fails (writing to out) while the fixed one
  works correctly.

**Description**

This pull request fixes a logic error in `TimeMapAxis.pix_to_coord()`.
The original implementation used `np.logical_and(a, b, c)`, assuming the
third argument was an additional condition. However, in NumPy ufunc
semantics the third positional argument is the `out` parameter, used for
storing the result rather than for logical evaluation.

As a consequence, the `np.isfinite(idx)` check was not actually applied as
a condition and could be silently ignored, allowing NaN or invalid indices
to pass the validity mask.

The code is corrected by explicitly combining all conditions using boolean
operations, ensuring `isfinite(idx)` is properly enforced.

A regression test is added that exposes the bug by making
`np.isfinite()` return a read-only array: the old implementation attempts
to write into it (via `out`) and fails, while the fixed implementation
works as expected.

**Dear reviewer**

This PR is ready for review.

The change is minimal and localized, does not affect the public API, and
adds a targeted regression test to guard against future regressions of the
same kind. No documentation update is required, as this is an internal bug
fix.

Feedback is welcome, but the PR should be ready to merge as-is.

### DIFF
--- a/docs/release-notes/6384.bug.rst
+++ b/docs/release-notes/6384.bug.rst
@@ -1,1 +1,0 @@
-Fix a bug in `gammapy.maps.axes.pix_to_coord` where np.logical_and(a, b, c) mistakenly treated the 3rd argument as the ufunc 'out' parameter instead of a third condition, potentially skipping the isfinite(idx) check.


QRemy (commented):
Thanks @yaochengchen 
Can you find a test case in which these are not equivalent without the read-only trick ?
or can you justify why we would need to defend against the read only case, what would produce this scenario ?

For current usage I think both are equivalent given that 0 and nbin are finite values the np.isfinite test is not even necessary:
Image
so I would relabel that as cleanup and remove the new test unless you can justify the read-only case.

yaochengchen:
> Thanks @yaochengchen Can you find a test case in which these are not equivalent without the read-only trick ? or can you justify why we would need to defend against the read only case, what would produce this scenario ?
> 
> For current usage I think both are equivalent given that 0 and nbin are finite values the np.isfinite test is not even necessary: Image so I would relabel that as cleanup and remove the new test unless you can justify the read-only case.

Thanks @QRemy , If idx is produced from floating-point operations (e.g. modf, division, normalization, log), NaN/Inf values are not uncommon. In that case np.clip does not clip NaN into [0, nbin-1] — NaN remains NaN — so np.isfinite(idx) is still a necessary guard, especially if idx is later cast to int or used for indexing, where NaN/Inf can turn into subtle bugs or hard errors.
Regarding the read-only aspect: this is not about defending against read-only inputs per se. It is a way to make the bug deterministic. In NumPy, np.logical_and(a, b, c) treats the third argument as out=, i.e. it attempts to write the result of a & b into c. By monkeypatching np.isfinite(idx) to return a read-only array, the test reliably exposes that the old implementation was writing into the third argument. The fixed implementation no longer does this, so the test passes.
If the read-only test feels too indirect, I’ve also added a more “semantic” test that includes NaN/Inf in pix/idx and verifies that isfinite correctly masks those entries.

gammapy/maps/axes.py
```diff
@@ -2672,7 +2672,8 @@ def pix_to_coord(self, pix):
         coords = np.zeros_like(pix)
         frac, idx = np.modf(pix)
         idx1 = idx.astype(int)
-        valid = np.logical_and(idx >= 0, idx < self.nbin, np.isfinite(idx))
+
+        valid = (idx >= 0) & (idx < self.nbin) & np.isfinite(idx)
```
QRemy:
```suggestion
        valid = (idx >= 0) & (idx < self.nbin)
```

This also pass both tests, as said in my previous comment, given that 0 and nbin are finite values the np.isfinite test is not even necessary because (idx >= 0) & (idx < self.nbin) will already return False.
So you can relabel this as cleanup and remove the fragment and tests.

yaochengchen: Thanks @QRemy, and sorry about my previous reply — I wasn’t thinking clearly last night. I’m about to head back to China for my holiday and was rushing to finish those PRs, since I won’t be able to access the server from China.

QRemy (dismissed):
Thanks @yaochengchen, no further comment from my side.

gammapy/maps/axes.py
```diff
@@ -2672,7 +2672,8 @@ def pix_to_coord(self, pix):
         coords = np.zeros_like(pix)
         frac, idx = np.modf(pix)
         idx1 = idx.astype(int)
-        valid = np.logical_and(idx >= 0, idx < self.nbin, np.isfinite(idx))
+
+        valid = (idx >= 0) & (idx < self.nbin)
```
Astro-Kirsty, line 2676: There is a long explanation of how this can happen in another comment. I suggest we add a test then too

A change log should also be added

registerrier (approved):
Thanks @yaochengchen . Looks good. No further comments.

PR #6385, opened 2026-02-08
Fix: Raise IndexError for negative indices in Map.get_image_by_idx

yaochengchen:
Prevents silent data corruption when interacting with external tools (like Astropy WCS) where `-1` is used as a sentinel value for invalid coordinates.

Previously, passing `-1` would trigger Numpy negative indexing (returning the last bin), leading to correct data shapes but wrong physical values. This change enforces strict index validation to avoid hard-to-debug scientific errors.

Description

This pull request introduces a safeguard in Map.get_image_by_idx to explicitly disallow negative indices.

Motivation While internal Gammapy calls typically ensure indices are valid (via clip=True or internal logic), interactions with external tools (e.g., Numpy, Scipy, Astropy WCS) present a risk. In astronomical data processing, -1 is frequently used as a sentinel value for "invalid" or "out-of-bounds" pixels (e.g., often resulting from astropy.wcs.world_to_array_index boundary handling).

The Problem Previously, if a user or external script inadvertently passed -1 to get_image_by_idx, Numpy would interpret it as the last element (negative indexing). This resulted in a "silent failure": the returned Map has the correct shape but contains scientifically incorrect data (e.g., returning the data from the last energy bin instead of indicating an error).

The Solution This change enforces strict index validation. It explicitly checks for negative indices and raises an IndexError, preventing hard-to-debug scientific errors downstream.

Dear reviewer

This is ready for review.

Scope: This is a low-risk, defensive change. I confirmed that existing internal tests pass, as internal Gammapy logic correctly handles indices. This is primarily to protect against user-side scripts or external WCS conversions that might leak -1 as a sentinel value.

Tests: I have included a test case to verify that passing negative indices now correctly raises an IndexError instead of silently returning data.

### DIFF
--- a/gammapy/maps/tests/test_core.py
+++ b/gammapy/maps/tests/test_core.py
@@ -120,7 +120,6 @@ def test_map_get_image_by_pix(binsz, width, map_type, skydir, axes, unit):
         binsz=binsz, width=width, map_type=map_type, skydir=skydir, axes=axes, unit=unit
     )
     pix = (1.2345, 0.1234)[: len(m.geom.axes)]
-    # idx = m.geom.pix_to_idx(pix)
     # print("idx:", idx)
     if map_type == "hpx":
         # Reason: In HPX mode, `get_image_by_pix` incorrectly parses non-spatial coordinates
@@ -1140,23 +1139,12 @@ def test_quantity():
 def test_get_image_by_idx_error_handling():
     """Test that get_image_by_idx raises IndexError for negative indices derived from coords."""
 
-    # 1. Create a Map with an energy axis (range: 1 TeV - 10 TeV)
     axis = MapAxis.from_edges([1, 10], unit="TeV", name="energy", interp="log")
     m = Map.create(npix=(5, 5), axes=[axis])
 
-    # 2. Define a coordinate that is clearly out of bounds (0.1 TeV < 1 TeV)
-    # This simulates an out-of-range value passed by a user or external tool
     energy_out_of_bounds = 0.1 * u.TeV
-
-    # 3. Calculate the index via transformation
-    # Key point: use clip=False. Normally this returns -1 (indicating out of bounds)
-    # Without the fix, this idx (-1) would be silently accepted as the last bin
     idx = axis.coord_to_idx(energy_out_of_bounds, clip=False)
-
-    # 4. Verify that the calculated index is indeed -1 (ensure test logic is correct)
     assert idx == -1
 
-    # 5. Pass the calculated index to the function and verify it raises an error
-    # match="Negative" ensures we catch the specific error added in the fix
     with pytest.raises(IndexError, match="Negative"):
         m.get_image_by_idx((idx,))


bkhelifi:
An associated test function is needed to check the error and catch the raise

yaochengchen:
> An associated test function is needed to check the error and catch the raise

Thanks for the review! I have added the requested test case.

I added test_get_image_by_idx_error_handling which explicitly checks that an IndexError is raised when an out-of-bounds coordinate (resulting in -1) is passed.

Note on test_map_get_image_by_pix: I also had to update the existing test_map_get_image_by_pix. It turns out this test was previously passing for HPX maps only because of the silent failure bug I just fixed.

For HPX maps, m.geom.pix_to_idx(pix) currently returns -1 when provided with non-spatial coordinates. Previously, get_image_by_idx silently wrapped this to the last bin. Now, it correctly raises an error. I updated the test to catch this IndexError for HPX maps. The behavior of pix_to_idx for HPX might need a separate look in the future, but the safety check is now working as intended.

yaochengchen:
As a follow-up, while working on this change I noticed an inconsistency between HPX and WCS map behavior. I fixed the corresponding HPX-specific issue as well, and included the changes in PR #6388.

QRemy:
> The Problem Previously, if a user or external script inadvertently passed -1 to get_image_by_idx, Numpy would interpret it as the last element (negative indexing). This resulted in a "silent failure": the returned Map has the correct shape but contains scientifically incorrect data (e.g., returning the data from the last energy bin instead of indicating an error).

Returning the data from the last energy bin seems to be the expected behaviour to me.
If the problem appears only for HPX maps as suggested by the test, I would suggest to directly go for a fix of `HpxGeom.pix_to_idx` and close this PR.

registerrier:
@yaochengchen This PR is superseded by #6388 . Correct? If yes, we should close this one.

yaochengchen:
> @yaochengchen This PR is superseded by #6388 . Correct? If yes, we should close this one.

@registerrier Thanks for checking. I don’t think this PR is strictly superseded by #6388.

This PR adds a defensive check in Map.get_image_by_idx against negative indices that can come from external tools (e.g. Astropy WCS, where -1 is often a sentinel for invalid coordinates). I only noticed the HPX-specific issue fixed in #6388 after this change when running the tests.

While -1 is valid under NumPy indexing, silently mapping it to the last energy bin in this context can be confusing and potentially lead to unintended scientific results

AtreyeeS:
Thanks @yaochengchen ! Do you have a specific use case where this created a problem? I believe we have been using numpy indexing rules so far, and this change might break many user scripts.

yaochengchen:
> Thanks @yaochengchen ! Do you have a specific use case where this created a problem? I believe we have been using numpy indexing rules so far, and this change might break many user scripts.

@AtreyeeS Thanks for raising this.

A minimal example (attached)
[energy_scan.py](https://github.com/user-attachments/files/25269911/energy_scan.py)
 that shows the issue.

I didn’t encounter it in a specific analysis — I noticed it while looking into potential GPU acceleration paths and reading through the `coord_to_idx` → `get_image_by_idx` flow.

What caught my attention is that in Gammapy, `MapAxis.coord_to_idx(..., clip=False)` explicitly defines `-1` as an out-of-range sentinel. So in Gammapy semantics, `-1` means “invalid”, not “last bin”.

In automated workflows (e.g. energy grids from `logspace`, catalog energies, floating-point edge effects), it’s quite easy for one energy to fall slightly outside the axis. In that case, users might expect an error or invalid value, but instead they get the highest energy plane with no indication anything went wrong. That silent reinterpretation of an invalid request is what concerns me.

That said, I’m relatively new in this field, so it’s very possible that my concern is misplaced or that this behavior is intentional and well understood by experienced users. I’m very happy to follow the maintainers’ judgement here and align with the project’s established design choices.

bkhelifi:
@yaochengchen, a side question: do you sign your commits as explained here https://docs.gammapy.org/dev/development/intro.html#acceptation-of-the-developer-certificate-of-origin-dco ?

yaochengchen:
> @yaochengchen, a side question: do you sign your commits as explained here https://docs.gammapy.org/dev/development/intro.html#acceptation-of-the-developer-certificate-of-origin-dco ?

@bkhelifi Yes, I have signed-off my commits.

AtreyeeS:
> A minimal example (attached) [energy_scan.py](https://github.com/user-attachments/files/25269911/energy_scan.py) that shows the issue.

Thanks a lot @yaochengchen  for this code snippet. The issue is same for hpx and wcs maps.

Indeed, we do have an inconsistency between `coord_to_idx` having -ve idx as invalid, and `get_image_by_idx` where we use numpy slicing for the idx. 
In fact, we use -1 as INVALID_INDEX as defined here https://github.com/gammapy/gammapy/blob/4a0d85ad4387b3dbd0ceec37679b5731e8d69d3b/gammapy/maps/utils.py#L85, but use slicing rules for `get_by_idx`, `set_by_idx` etc, but not `fill_by_idx`.

We need to look into this carefully and implement a consistent scheme.
I am closing this PR and opening a reminder issue

PR #6386, opened 2026-02-08
fix(maps): align TimeMapAxis pixel convention with MapAxis; add round…

yaochengchen:
fix(maps): align TimeMapAxis pixel convention and add round-trip test

TimeMapAxis used an inconsistent pixel convention compared to MapAxis and the
geometry helpers: coord_to_pix applied a -0.5 shift (pix=0 treated as bin
center), while pix_to_coord interpreted pix=0 as the lower edge using np.modf.
This mismatch caused a systematic -0.5 pixel offset in pix->coord->pix
round-trips and led to ambiguous center/edge semantics.

This change makes TimeMapAxis follow the standard Gammapy convention:
integer pix corresponds to the bin center, and bin edges lie at half-integer
pixel positions.

Changes
- Update TimeMapAxis.pix_to_coord() to apply the same pixel-offset convention as
  MapAxis (q = pix - self._pix_offset with _pix_offset = -0.5), and compute
  idx = floor(q) and frac = q - idx for linear interpolation within bins.
- Update TimeMapAxis.coord_to_pix() to return pix + self._pix_offset so that
  coord_to_pix() and pix_to_coord() use the exact same offset convention.
- Replace the stub TimeMapAxis.pix_to_idx() (previously a staticmethod returning
  pix) with an instance method consistent with MapAxis, supporting clipping and
  mapping out-of-range / non-finite values to -1.

Tests
- Add a new regression test to ensure pix->coord->pix round-trip consistency for
  representative pixel positions and to prevent reintroducing the systematic
  -0.5 pixel shift.

Notes
- coord_to_idx boundary semantics are unchanged by this commit; the fix targets
  only the pixel offset convention and associated conversions.

Description
-----------
This pull request fixes an inconsistency in the TimeMapAxis pixel convention that
caused pixel coordinates not to be preserved under a pix -> coord -> pix
transformation.

As demonstrated by the newly added regression test, the original implementation
produced a systematic shift after a round trip: converting from pixel to time
coordinate and back resulted in a -0.5 pixel offset. The root cause was a mismatch
between pix_to_coord (interpreting pix=0 as a lower bin edge) and coord_to_pix
(interpreting pix=0 as a bin center).

The updated implementation aligns TimeMapAxis with the pixel conventions used by
MapAxis and the geometry layer across Gammapy. After this change, pix_to_coord
and coord_to_pix are mutually consistent and preserve pixel coordinates for
valid inputs.

Dear reviewer
-------------
This PR is ready for review.

It includes a focused regression test that reproduces the original bug and
verifies that the updated implementation preserves pixel coordinates under a
pix -> coord -> pix round-trip.

The changes are limited to internal coordinate transformations; no public API
is altered, and existing coord_to_idx boundary semantics are intentionally left
unchanged.

Feedback on the implementation details or test coverage is very welcome.

### DIFF
--- a/docs/release-notes/6386.bug.rst
+++ b/docs/release-notes/6386.bug.rst
@@ -1,0 +1,1 @@
+Align TimeMapAxis pixel convention. Update pix_to_coord to apply the same pixel-offset convention as MapAxis.


QRemy:
@AtreyeeS @fabiopintore, I think you opened PRs about shift in time bin previously, should the convention really be the same in that case ?

gammapy/maps/tests/test_axes.py
```diff
@@ -1127,3 +1127,54 @@ def test_periodic_map_axis():
         axis1 = MapAxis.from_bounds(
             -0.5, 0.5, 5, boundary_type="periodic", interp="log"
         )
+
+
+def test_time_map_axis_pix_coord_roundtrip():
+    """
```
Astro-Kirsty: Just a note: we typically do not comment on the test code as much. Your test name and argument names should be informative enough. 

Adding a changelog fragment will also help with that.

Thanks!

yaochengchen:
> Thanks @yaochengchen, I didn't read closely the code but from a quick look, it seems largely AI generated, nothing against it, but you should add some instructions in your prompts such as "write the docstring in the style of gammapy" "write the tests using pytest in the style of gammapy", so we will waste less time in review to point these kind of discrepancies.
> 
> Also the comments about the corrections can be useful in the regression test but in the core code that's not really necessary.

yaochengchen:
> Thanks @yaochengchen, I didn't read closely the code but from a quick look, it seems largely AI generated, nothing against it, but you should add some instructions in your prompts such as "write the docstring in the style of gammapy" "write the tests using pytest in the style of gammapy", so we will waste less time in review to point these kind of discrepancies.
> 
> Also the comments about the corrections can be useful in the regression test but in the core code that's not really necessary.

Thanks @QRemy , I'll be more careful with the docstring and test styles next time. Sorry for the trouble.

gammapy/maps/tests/test_axes.py
yaochengchen: @Astro-Kirsty Thanks for the comments, sorry about that, I'll be more careful with the docstring and test styles next time.

Astro-Kirsty: No worries at all, it is normal for your first time contributing to gammapy :)

You can check out other tests and pieces of code to see how things are done, in addition there are some steps to follow seen here:
https://docs.gammapy.org/2.0.1/development/intro.html

AtreyeeS:
@QRemy I have to think more, but I remember some different handling was introduced because `TimeMapAxis` can contain non-contiguous values, so we decided to forbid negative pix values. Lets discuss this during the dev call before merging

gammapy/maps/tests/test_axes.py
```diff
@@ -1127,3 +1127,37 @@ def test_periodic_map_axis():
...
+def test_time_map_axis_pix_to_idx():
+    ref_time = Time("2024-01-01T00:00:00")
+    edges_min = np.array([0, 10, 20]) * u.s
+    edges_max = np.array([10, 20, 30]) * u.s
+    axis = TimeMapAxis(edges_min, edges_max, reference_time=ref_time)
+
+    pix = np.array([-0.6, -0.5, 0.5, 1.5, 2.5, np.nan])
+    idx = axis.pix_to_idx(pix, clip=False)
+
+    assert_equal(idx, [-1, 0, 1, 2, -1, -1])
```
AtreyeeS, line 1160: TimeMapAxis is between two edges, so idx 0 corresponds to pix 0 to 1 
Pix -0.5 falls outside this range

yaochengchen: Hi @AtreyeeS, @QRemy 

Thanks for the clarification. I realise my understanding of the TimeMapAxis pixel convention might not be fully aligned with the intended design.

At this point I’m not confident enough about the overall convention to continue modifying this PR without risking further inconsistencies. I would prefer to pause here until I have a better understanding of the full design. If needed, I’m also fine with closing this PR.

AtreyeeS:
Thanks @yaochengchen ! I am closing this PR now. If you face any issues with the TimeMapAxis in future, please open an issue, and we will continue there.

PR #6387, opened 2026-02-08
Make LogScale clipping threshold adaptive to input dtype

yaochengchen:
**Description**

This pull request makes the clipping threshold used by `LogScale`
adaptive to the input dtype.

Previously, `LogScale` always clipped values to the float32 machine
minimum (`~1e-38`), even when the input data were explicitly provided
as `float64`. This caused silent precision loss for double-precision
inputs and could bias intermediate results.

While extremely small physical flux values are unlikely in typical
gamma-ray analyses, Gammapy frequently operates on intermediate
quantities where very small numbers are both expected and meaningful:

- Exponentially cut-off spectral models, where high-energy tails can
  extend far below `1e-38`.
- Probability density functions and likelihood evaluations, especially
  in the far tails of distributions.
- Products of multiple small factors (e.g. IRFs or convolutions), where
  individually moderate values can combine into very small numbers.
- Data consistency: `float64` inputs should not be implicitly degraded
  by internal `float32` assumptions, as this can lead to misleading
  results on logarithmic scales.

To address this, the clipping threshold is now derived from
`np.finfo(values.dtype).tiny`, ensuring that:
- `float32` inputs retain the previous behavior and numerical stability.
- `float64` inputs are allowed to reach much smaller values, avoiding
  silent precision loss.

The existing inverse-transform semantics of `LogScale` are intentionally
preserved.

New tests were added to verify the adaptive behavior. The previous
implementation fails these tests, while the updated implementation
passes.

**Dear reviewer**

This pull request is ready for review.

I would appreciate feedback on the API and numerical behavior, in
particular on preserving the inverse-transform semantics while making
the clipping threshold dtype-aware. Tests covering the new behavior are
included; no additional documentation changes are expected to be
necessary.

### DIFF
--- a/docs/release-notes/6387.feature.rst
+++ b/docs/release-notes/6387.feature.rst
@@ -1,1 +1,1 @@
-Make LogScale clipping threshold adaptive to input dtype.
+Make LogScale clipping threshold adaptive to input dtype. This allows interpolation on float64 arrays without hard clipping at the float32 threshold (~1e-38).


gammapy/utils/interpolation.py
```diff
@@ -196,16 +196,29 @@ def _inverse_deriv(self, values):
 class LogScale(InterpolationScale):
     """Logarithmic scaling."""
 
+    # Keep this attribute for backward compatibility / default behavior,
+    # but do not use it as the sole clipping threshold.
```
QRemy:
```suggestion
```

gammapy/utils/interpolation.py
```diff
@@ -196,16 +196,29 @@ def _inverse_deriv(self, values):
...
     tiny = np.finfo(np.float32).tiny
 
+    @staticmethod
+    def _tiny_for(arr):
+        arr = np.asanyarray(arr)
+        # If the input is not a floating type (e.g. int), log/exp semantics
+        # still require floating point; use float64 tiny as a fallback.
```
QRemy:
```suggestion
```

gammapy/utils/interpolation.py
```diff
@@ -196,16 +196,29 @@ def _inverse_deriv(self, values):
...
+        if not np.issubdtype(arr.dtype, np.floating):
+            return np.finfo(np.float64).tiny
```
QRemy:
```suggestion
            return tiny
```

gammapy/utils/interpolation.py
```diff
@@ -196,16 +196,29 @@ def _inverse_deriv(self, values):
 class LogScale(InterpolationScale):
     """Logarithmic scaling."""
 
+    # Keep this attribute for backward compatibility / default behavior,
+    # but do not use it as the sole clipping threshold.
     tiny = np.finfo(np.float32).tiny
 
+    @staticmethod
+    def _tiny_for(arr):
```
QRemy:
```suggestion
    def _tiny_for(arr, tiny):
```

QRemy:
Old discussions related #2374 #2405

yaochengchen:
Thanks @QRemy , I think the values 5.7314076 in "gammapy/makers/tests.test_spectrum_dataset_maker_hess_dl3" and 5.731624 in "test_region_center_spectrum_dataset_maker_hess_dl3" need to be updated so that this commit can pass CI.

gammapy/utils/tests/test_interpolation.py
```diff
@@ -4,11 +4,41 @@
 from gammapy.utils.testing import assert_allclose
 
 
-def test_LogScale_inverse():
-    tiny = np.finfo(np.float32).tiny
+def test_LogScale_adaptive_behavior():
```
QRemy:
```suggestion
def test_logscale_adaptive_behavior():
```

QRemy:
> Old discussions related #2374 #2405

The previous behaviour of float32 was a design choice motivated by the fact that float64 could increase memory usage, but the results of the memray profiler for the 3d joint analysis don't look dramatically different. 

Main:
analysis_3d_joint_main

This PR:
analysis_3d_joint_pr6387

QRemy (approved):
Thanks @yaochengchen, looks good, ok to merge from my side.

gammapy/estimators/points/tests/test_sensitivity.py
```diff
@@ -112,7 +112,7 @@ def test_parameter_sensitivity_estimator(spectrum_dataset):
     estimator = ParameterSensitivityEstimator(spectral_model.amplitude, 0, rtol=1e-2)
 
     value = estimator.run(datasets)
-    assert_allclose(value, 4.67553e-12, rtol=1e-2)
+    assert_allclose(value, 4.451851e-12, rtol=1e-2)
```
registerrier, line 115: Any idea why the impact is so large here @QRemy @yaochengchen ?

gammapy/modeling/models/tests/test_spectral.py
```diff
@@ -789,7 +789,7 @@ def test_template_spectral_model_evaluate_tiny():
         energy=energy, values=values * u.Unit("MeV-1 s-1 sr-1")
     )
     result = model(energy)
-    tiny = np.finfo(np.float32).tiny
+    tiny = np.finfo(values.dtype).tiny
```
registerrier, line 792: Shouldn't you impose `values`'s `dtype` to `np.float32` instead? Or test both cases, float32 and float64 template arrays?

docs/release-notes/6387.feature.rst
```diff
@@ -0,0 +1 @@
+Make LogScale clipping threshold adaptive to input dtype.
```
registerrier: You might add a sentence to specify this allows to use interpolation on float64 arrays without clipping at 1e-38

gammapy/modeling/models/tests/test_spectral.py
yaochengchen: The intention here is to keep the test dtype-agnostic and consistent with the input precision, rather than hard-coding float32 assumptions. Using np.finfo(values.dtype) makes the test more robust if the input dtype or defaults change.

This change wasn’t introduced by me originally, but that’s my understanding of the motivation.

gammapy/estimators/points/tests/test_sensitivity.py
yaochengchen: Sorry, I have no idea.

docs/release-notes/6387.feature.rst
yaochengchen: Thanks, updated accordingly.

gammapy/estimators/points/tests/test_sensitivity.py
QRemy: I guess internally brentq was dealing with very small numbers and that's not super stable.

PR #6388, opened 2026-02-09
Fix HPX pix_to_idx handling of non-spatial pix and add regression test

yaochengchen:
Description
This pull request fixes an inconsistency in the HPX geometry handling of pixel coordinates and adds a regression test to guard against it.
Previously, HpxGeom.pix_to_idx() always interpreted the first element of pix as a global HEALPix spatial index. This caused incorrect -1 indices (or downstream errors) when get_image_by_pix() passed only non-spatial pixel coordinates, which is the documented and expected usage of that method. The issue primarily affected partial-sky HPX geometries, where such global indices are almost never valid.
The fix updates HpxGeom.pix_to_idx() to explicitly distinguish between two valid input cases:
pix contains only non-spatial axes (as used by get_image_by_pix)
pix contains full spatial + non-spatial coordinates
For the non-spatial-only case, the method now delegates to axes.pix_to_idx() instead of interpreting the first element as a HEALPix spatial index. This preserves the existing behavior for full pixel inputs while making the HPX and WCS implementations consistent for non-spatial access.
A regression test is added to ensure that get_image_by_pix() works correctly for HPX maps when only non-spatial pixel coordinates are provided.
Dear reviewer
This PR is ready for review.
The change is intentionally limited to HPX-specific code and does not modify shared Map or get_image_by_pix() logic.
Existing behavior for full (spatial + non-spatial) pixel inputs is preserved.
A regression test is included and fails with the previous implementation while passing with this fix.
No documentation changes are included, as the behavior now matches the existing docstring and WCS behavior.
Feedback on the approach or edge cases is very welcome.

### DIFF
--- a/gammapy/maps/tests/test_core.py
+++ b/gammapy/maps/tests/test_core.py
@@ -119,6 +119,7 @@ def test_map_get_image_by_pix(binsz, width, map_type, skydir, axes, unit):
     m = Map.create(
         binsz=binsz, width=width, map_type=map_type, skydir=skydir, axes=axes, unit=unit
     )
+    m.data = np.arange(m.data.size, dtype=float).reshape(m.data.shape)
     pix = (1.2345, 0.1234)[: len(m.geom.axes)]
     m_image = m.get_image_by_pix(pix)
 
@@ -1124,28 +1125,3 @@ def test_quantity():
     m = Map.from_geom(geom)
     with pytest.raises(TypeError):
         m.data = 0 * u.deg
-
-
-@pytest.mark.parametrize("axes", [map_axes[:1], map_axes])
-def test_hpx_get_image_by_pix_non_spatial_only_regression(axes):
-    pytest.importorskip("healpy")
-
-    m = Map.create(
-        binsz=0.1,
-        width=10.0,
-        map_type="hpx",
-        skydir=SkyCoord(0.0, 30.0, unit="deg"),
-        axes=axes,
-        unit="",
-    )
-    m.data = np.arange(m.data.size, dtype=float).reshape(m.data.shape)
-
-    pix = (1.2345, 0.1234)[: len(m.geom.axes)]
-
-    m_image = m.get_image_by_pix(pix)
-
-    im_geom = m.geom.to_image()
-    idx = im_geom.get_idx()
-    m_vals = m.get_by_pix(idx + pix)
-
-    assert_equal(m_image.data, m_vals)


gammapy/maps/tests/test_core.py
```diff
@@ -1124,3 +1124,30 @@ def test_quantity():
...
+        axes=axes,
+        unit="",
+    )
+    m.data = np.arange(m.data.size, dtype=float).reshape(m.data.shape)
+
+    # get_image_by_pix contract: pix contains ONLY non-spatial axes coordinates
+    pix = (1.2345, 0.1234)[: len(m.geom.axes)]
+
+    # old buggy HPXGeom.pix_to_idx mis-parses pix[0] as ipix -> -1 / IndexError
+    m_image = m.get_image_by_pix(pix)  # should NOT raise after the fix
```
QRemy:
```suggestion
    m_image = m.get_image_by_pix(pix)
```

gammapy/maps/tests/test_core.py
```diff
@@ -1124,3 +1124,30 @@ def test_quantity():
...
+        binsz=0.1,
+        width=10.0,
+        map_type="hpx",
+        skydir=SkyCoord(0.0, 30.0, unit="deg"),
+        axes=axes,
+        unit="",
+    )
+    m.data = np.arange(m.data.size, dtype=float).reshape(m.data.shape)
+
+    # get_image_by_pix contract: pix contains ONLY non-spatial axes coordinates
```
QRemy:
```suggestion
```

yaochengchen: Thanks @QRemy , Sorry for the messy coding, I will follow the coding format next time.

QRemy (approved):
Thanks @yaochengchen, no further comment from my side.

gammapy/maps/tests/test_core.py
```diff
@@ -1124,3 +1124,28 @@ def test_quantity():
     m = Map.from_geom(geom)
     with pytest.raises(TypeError):
         m.data = 0 * u.deg
+
+
+@pytest.mark.parametrize("axes", [map_axes[:1], map_axes])
+def test_hpx_get_image_by_pix_non_spatial_only_regression(axes):
+    pytest.importorskip("healpy")
```
registerrier:
```suggestion
```
the importorskip is applied to the whole file.

gammapy/maps/hpx/geom.py
```diff
@@ -314,6 +314,13 @@ def pix_to_coord(self, pix):
 
     def pix_to_idx(self, pix, clip=False):
         # FIXME: Look for better method to clip HPX indices
+
+        # If pix only contains non-spatial coordinates (e.g. Map.get_image_by_pix),
+        # do NOT interpret pix[0] as HEALPix ipix.
+        if len(pix) == len(self.axes):
+            pix = pix_tuple_to_idx(pix)
```
registerrier, line 321: Why is this step necessary? Why don't you pass `pix` directly. You are transforming it to idx twice.

registerrier: I suppose this might be because `MapAxis.pix_to_idx` is in fact not really returning an `idx`...

gammapy/maps/tests/test_core.py
```diff
@@ -1124,3 +1124,28 @@ def test_quantity():
     m = Map.from_geom(geom)
     with pytest.raises(TypeError):
         m.data = 0 * u.deg
+
+
+@pytest.mark.parametrize("axes", [map_axes[:1], map_axes])
+def test_hpx_get_image_by_pix_non_spatial_only_regression(axes):
```
registerrier: I ma not sure you need to reimplement a test. You could already adapt existing ones as this test is re-using most of `test_map_get_image_by_pix`.

gammapy/maps/hpx/geom.py
yaochengchen: The extra pix_tuple_to_idx() step is intentional.
In this branch pix is a tuple of non-spatial pixel coordinates, while MapAxis.pix_to_idx() does not operate on tuple-style pix input but expects array-like indices for clipping/masking.
So this is not a double pix→idx conversion, but a tuple→array normalization followed by the axis-level index handling.
I agree the naming of pix_to_idx is a bit misleading here, which is likely the source of the confusion.

gammapy/maps/tests/test_core.py
yaochengchen: Thanks for pointing this out, I will modify this.

gammapy/maps/tests/test_core.py
yaochengchen: Thanks @registerrier, updated as suggested. Reusing test_map_get_image_by_pix with non-trivial data makes the coverage more straightforward.

yaochengchen:
@registerrier @QRemy Sorry I accidently push re-review.

QRemy:
The behaviour of `MapAxis.pix_to_idx()` could be clarified in a separate PR, merging this now.

PR #6389, opened 2026-02-11
Add plot_flux_violin

QRemy:
Add plot_flux_violin. 
Matplotlib’s violinplot cannot be used because it always builds violins in linear space, without support for log‑space KDE, per‑sample weights, energy‑dependent flux scaling, or containment-based tail clipping. As a result, the shapes and quantiles would be incorrect on log–log axes and for weighted posterior samples.

### DIFF
--- a/docs/release-notes/6389.feature.rst
+++ b/docs/release-notes/6389.feature.rst
@@ -1,1 +1,1 @@
-Add `~gammapy.visualizzation.plot_samples_violin_vs_energy` to create a violin plot from samples of a given quantity as a function of energy.`
+Add `~gammapy.visualization.plot_samples_violin_vs_energy` to create a violin plot from samples of a given quantity as a function of energy.`


QRemy:
`np.percentile(weight=...)` requires numpy>=2.0
We should drop numpy=1.26 as it is no longer supported: https://scientific-python.org/specs/spec-0000/#support-window

facero:
A general approach question: 
do you need to use the weights at all here ?
In the sampler outputs, you have access to samples that have been reweighted already which are the one used for the corner plot I think.

QRemy:
> A general approach question: do you need to use the weights at all here ? In the sampler outputs, you have access to samples that have been reweighted already which are the one used for the corner plot I think.

It differs slightly

Screenshot 2026-02-11 at 23 21 47
Screenshot 2026-02-11 at 23 22 00

I guess that's because the samples are drawn among the weighted samples with the probability given by the weight
so you can have a more peaked distribution because the most probable sample appears many times
Screenshot 2026-02-11 at 23 26 54

QRemy:
> A general approach question: do you need to use the weights at all here ? In the sampler outputs, you have access to samples that have been reweighted already which are the one used for the corner plot I think.

It's a bit similar to what I shown in https://github.com/QRemy/gammapy/blob/bayesian_model_selectio_test_nbn/examples/tutorials/bayesian_model_selection.ipynb
the distributions from the samples seems a bit noisy because the same value is often picked when the size of the sample is the same as the size of the weighted sample : 
Screenshot 2026-02-11 at 23 32 11
but if we resample the posterior to have less samples than weighed samples there is less repetition and it looks more smooth
Screenshot 2026-02-11 at 23 32 26

gammapy/visualization/violin.py
```diff
@@ -0,0 +1,318 @@
...
+    bw_method="scott",
+    grid_size=200,
+    errorbar_kwargs=None,
+    errorbar_ul_kwargs=None,
+    violin_kwargs=None,
+    violin_clip=None,
+    y_label="dN/dE",
+):
+    """
+    Plot flux-sample violin distributions per energy bin on log–log axes.
```
bkhelifi: Only in log-log? the log-lin is not possible? If X-ray analysis in made, the energy axis can be in lin... Do you confirm? 

The second question is about the used statistics. In principle, even a sample of "frequentist" likelihoods could be used. Here, it seems that only posteterios (ie fully Bayesian or maximum a posteriori estimation). Do you confirm? Can we extend that?

facero: most of the time the X-ray spectral plot are also displayed in log-log.
But it would be nice in general to be able to control this behavior

QRemy: improvement could be made in a follow-up PR, I'm not even sure our others plotting methods support lin scale properly.

JohannesBuchner:
Hi, @facero asked me to look here. The method implemented in ultranest is here: https://github.com/JohannesBuchner/UltraNest/blob/master/ultranest/utils.py#L149-L215 , see the docstring for the reference.

If you can go to the weighted samples, great, it's more accurate. For most people it is probably a bit cumbersome to carry the weights around.

UltraNest can also give you the uncertainty of the posterior distribution, see the "Quantifying the distribution tails" subsection in the line fitting tutorial at https://johannesbuchner.github.io/UltraNest/example-line.html

QRemy:
Few examples :
Screenshot 2026-02-13 at 12 34 39
Screenshot 2026-02-13 at 12 34 48

JohannesBuchner:
Why do you want violin plots though in the first place? Looks like error bars would be enough of a summary, or alternatively you could do a boxplot. 

Here is a rant on violinplots: https://www.youtube.com/watch?v=_0QMKFzW9fw

QRemy:
> Why do you want violin plots though in the first place? Looks like error bars would be enough of a summary, or alternatively you could do a boxplot.
> 
> Here is a rant on violinplots: https://www.youtube.com/watch?v=_0QMKFzW9fw

Should I make a 40 min video to answer this ?
No I'm joking, I don't having any opinion on violin plot, @registerrier just asked for it.

Looking at the figures above they still contains more information than the flux points.
Typically in the first figure looking only at the flux points people would tell you that the two points at 0.5-1TeV are incompatible. In the second plot they would display only upper limits because TS<4 (like in the figure below) and then refit a physical model on the flux points to conclude that any model is possible...

Screenshot 2026-02-13 at 12 34 29

So in short we were looking for an alternative representation because the problem is that people don't use flux points as a "good summary" but as data points to fit (sometimes they even fit the upper limits as regular points which will give dramatically different result in the previous example).

facero:
I think one of the nice thing about displaying the full posterior distribution for every energy bin is that you get rid of the arbitrary distinction between a point and an UL.

In traditional SED you require two arbitrary values ( UL at XX% if TS< YY).
This can be confusing when you compare to the model as illustrated by Quentin's SED example .

If there are alternative options to draw the posterior distributions rather than violin I'm all in.

gammapy/visualization/violin.py
facero:
```suggestion
    Plot flux-sample violin distributions per energy bin on log–log axes.
    Unlike standard error bars which only shows summary statistics, a violin plot shows the full probability density for each energy bin.
```

gammapy/visualization/violin.py
```diff
@@ -0,0 +1,318 @@
...
+      mirroring the normalized log-density around the bin center in
+      log-energy.
+    - Violin clipping is applied only to the KDE grid, not to the
+      quantile-based error bars.
+    - Negative flux samples are replaced by a small positive surrogate
+      (10% of the smallest positive sample) to allow log-space evaluation.
+    - The plotted error bars correspond to the 16%, 50%, and 84%
+      weighted percentiles of the *unclipped* samples, unless the median
+      is non-positive, in which case an upper limit marker is drawn
+      (corresponding to the 2σ one-sided percentile).
```
facero: Can this be an optional argument in percentile or sigma ?
for the distribution you'll need to specify it in percentile but it would be good to be consistent with the approach of FluxPoint for UL sigma definition

QRemy:
```suggestion
    - The plotted error bars correspond to the 16%, 50%, and 84%
      weighted percentiles of the *unclipped* samples, unless the median
      is non-positive, in which case an upper limit marker is drawn
      at the 98% percentile. For a gaussian distribution these values corresponds to 
      1σ errors and 2σ upper limit plotted by default by the `FluxPoints.plot` method.
```

QRemy: I tried to make the docstring more clear. 
Noe that `FluxPoints.plot` doesn't have a` n_sigma` option because everything is defined on the `FluxPoints` class so to mirror the behaviour we might need a specific class for samples rather than adding options on the plot method. This can be discussed and added in a future PR.

QRemy:

gammapy/visualization/tests/test_violin.py
```diff
@@ -0,0 +1,190 @@
...
+import matplotlib.pyplot as plt
+from matplotlib.artist import Artist
+from matplotlib.container import ErrorbarContainer
+import astropy.units as u
+
+from gammapy.visualization.violin import plot_flux_violin
+
+
+def test_plot_flux_violin_invalid():
+    with pytest.raises(ValueError):
```
AtreyeeS: Please check with `match` that the correct error is raised

gammapy/visualization/tests/test_violin.py
```diff
@@ -0,0 +1,190 @@
...
+        ]
+
+        _, ax = plt.subplots()
+        plot_flux_violin(
+            ax=ax,
+            energy_edges=energy_edges,
+            samples_per_band=samples_per_band,
+        )
+
+    with pytest.raises(ValueError):
```
AtreyeeS: Same

gammapy/visualization/tests/test_violin.py
```diff
@@ -0,0 +1,190 @@
...
+
+        _, ax = plt.subplots()
+        plot_flux_violin(
+            ax=ax,
+            energy_edges=energy_edges,
+            samples_per_band=samples_per_band,
+            weights_per_band=weights_per_band,
+        )
+
+    with pytest.raises(ValueError):
```
AtreyeeS: Same

gammapy/visualization/tests/test_violin.py
```diff
@@ -0,0 +1,190 @@
...
+        ]
+
+        _, ax = plt.subplots()
+        plot_flux_violin(
+            ax=ax,
+            energy_edges=energy_edges,
+            samples_per_band=samples_per_band,
+        )
+
+    with pytest.raises(ValueError):
```
AtreyeeS: ...

gammapy/visualization/tests/test_violin.py
```diff
@@ -0,0 +1,190 @@
...
+
+    energy_edges = np.array([1, 3]) * u.TeV
+    samples_per_band = [np.array([1.0]) * u.Unit("cm-2 s-1 TeV-1")]
+    weights_per_band = [np.array([0.0])]  # zero-sum → skipped internally
+
+    fig, ax = plt.subplots()
+
+    artists = plot_flux_violin(ax, energy_edges, samples_per_band, weights_per_band)
+
+    assert isinstance(artists, list)
```
AtreyeeS: How does this ensure the correct behaviour?

gammapy/visualization/violin.py
```diff
@@ -0,0 +1,320 @@
...
+    weights_per_band=None,
+    energy_power=None,
+    color="C0",
+    bw_method="scott",
+    grid_size=200,
+    errorbar_kwargs=None,
+    errorbar_ul_kwargs=None,
+    violin_kwargs=None,
+    violin_clip=None,
+    y_label="dN/dE",
```
AtreyeeS, line 21: `sed_type` is not necessary?

gammapy/visualization/violin.py
```diff
@@ -0,0 +1,320 @@
...
+    errorbar_ul_kwargs : dict, optional
+        Keyword arguments forwarded to ``ax.errorbar`` for the ul bars.
+    violin_kwargs : dict, optional
+        Keyword arguments forwarded to ``ax.fill`` for violin plot.
+    violin_clip : (float, float), optional
+        Lower and upper containment fractions (in [0, 1]) used to clip
+        the violin tails in log-space. If omitted, defaults to
+        ``(norm.cdf(-4), norm.cdf(4))``.
+    y_label : str, optional
+        Base label for the y-axis, before unit and scaling prefixes.
```
AtreyeeS, line 65: You are missing `color` here

gammapy/visualization/violin.py
```diff
@@ -0,0 +1,320 @@
...
+      1σ errors and 2σ upper limit plotted by default by the `FluxPoints.plot` method.
+    """
+
+    # Axis scaling
+    ax.set_xscale("log")
+    ax.set_yscale("log")
+
+    # Energy edges
+    energy_edges = u.Quantity(energy_edges)
+    unit_x = f"{energy_edges.unit}"
```
AtreyeeS, line 98: Will this work if users mix units `energy_edges=[100*u.GeV, 1*u.TeV, 10*u.TeV]`

gammapy/visualization/violin.py
```diff
@@ -0,0 +1,320 @@
...
+            marker="o",
+            ms=4.5,
+            mec="black",
+            mfc="white",
+            color="black",
+            capsize=2.5,
+            elinewidth=1.2,
+            lw=1.2,
+        )
+    if errorbar_ul_kwargs is None:
```
AtreyeeS: I think you can define the default and users can overwrite the arguments, like in our other plot methods. One might just want to change the marker shape keeping the rest to default

gammapy/visualization/violin.py
```diff
@@ -0,0 +1,320 @@
...
+from scipy.stats import gaussian_kde, norm
+
+
+def plot_flux_violin(
+    ax,
+    energy_edges,
+    samples_per_band,
+    weights_per_band=None,
+    energy_power=None,
+    color="C0",
```
AtreyeeS: I would not keep the colour here but inside the kwargs

QRemy: it's not actually used I'll remove it

gammapy/visualization/violin.py
QRemy: yes u.Quantity will uniformise the unit to the one of the first entry

gammapy/visualization/violin.py
QRemy: Maybe this can be added in a follow up PR.
Right now `samples_per_band` can be anything and `y_label` setting is let to the user so this can be used for any parameter (even radius for example).

gammapy/visualization/tests/test_violin.py
QRemy: added  ` assert len(artists) > 0`

gammapy/visualization/violin.py
QRemy: done

gammapy/visualization/violin.py
QRemy: name changed as `plot_samples_violin_vs_energy`

gammapy/visualization/violin.py
QRemy: remove from init

docs/release-notes/6389.feature.rst
```diff
@@ -0,0 +1 @@
+Add `~gammapy.visualizzation.plot_flux_violin.Datasets` to create a violin plot from flux samples given by the `FluxCollectionEstimator`
```
AtreyeeS: Adapt the name here,
Do you need to mention `FluxCollectionEstimator` here? This function can have other usages as we discussed last dev call

gammapy/visualization/violin.py
```diff
@@ -0,0 +1,333 @@
...
+        Axes object on which to draw the violins and error bars.
+    energy_edges : array-like or `~astropy.units.Quantity`
+        Monotonically increasing energy bin edges of length ``nbins + 1``.
+        Must be positive and finite for log scaling.
+    samples_per_band : list of `~astropy.units.Quantity`
+        List of sample arrays, one per energy bin. Each array contains
+        draws from the flux posterior (or other distribution) in that bin.
+    weights_per_band : list of array-like, optional
+        Per-sample weights for each bin, same length as ``samples_per_band``.
+        If omitted, uniform weights are assumed. Requires NumPy ≥ 2.0 for
```
AtreyeeS: Do we still need to support numpy < 2.0 ?

gammapy/visualization/violin.py
```diff
@@ -0,0 +1,333 @@
...
+    errorbar_ul_kwargs : dict, optional
+        Keyword arguments forwarded to ``ax.errorbar`` for the ul bars.
+    violin_kwargs : dict, optional
+        Keyword arguments forwarded to ``ax.fill`` for violin plot.
+    violin_clip : (float, float), optional
+        Lower and upper containment fractions (in [0, 1]) used to clip
+        the violin tails in log-space. If omitted, defaults to
+        ``(norm.cdf(-4), norm.cdf(4))``.
+    y_label : str, optional
+        Base label for the y-axis, before unit and scaling prefixes.
```
AtreyeeS, line 65: We usually give the defaults also

gammapy/visualization/violin.py
```diff
@@ -0,0 +1,333 @@
...
+
+    Returns
+    -------
+    artists : list
+        A list of Matplotlib artists corresponding to the violin polygons
+        and error-bar elements added to the axes.
+
+    Notes
```
AtreyeeS, line 73: For the correct rendering, we used "note" this this 
https://github.com/gammapy/gammapy/blob/2e3382820f080ab4bc72b5406cee61411c29e34a/gammapy/estimators/map/excess.py#L129

docs/release-notes/6389.feature.rst
Astro-Kirsty: Make sure that these are the proper links too

gammapy/visualization/violin.py
```diff
@@ -0,0 +1,333 @@
...
+    bw_method="scott",
+    grid_size=200,
+    errorbar_kwargs=None,
+    errorbar_ul_kwargs=None,
+    violin_kwargs=None,
+    violin_clip=None,
+    y_label="dN/dE",
+):
+    """
+    Plot flux-sample violin distributions per energy bin on log–log axes.
```
Astro-Kirsty:
```suggestion
    """Plot sample violin distributions per energy bin on log–log axes.
```

gammapy/visualization/violin.py
Astro-Kirsty: Wouldn't it be better to use the `sed_type`?

gammapy/visualization/violin.py
Astro-Kirsty: For this case, using the `Notes` is correct

gammapy/visualization/violin.py
```diff
@@ -0,0 +1,333 @@
...
+    1σ-equivalent quantiles are overlaid using Gammapy-style flux-point
+    error bars. Users can optionally apply an ``E_center**p`` scaling to the
+    flux values and clip violin tails to a specified containment interval.
+    Samples of other energy-dependent parameters can also be plotted,
+    in that case the `y_label` should be changed accordingly.
+
+    Parameters
+    ----------
+    ax : `matplotlib.axes.Axes`
+        Axes object on which to draw the violins and error bars.
```
Astro-Kirsty: Our normal gammapy way is to make this optional, i.e. `ax=None` in the function:
```suggestion
        ax : `~matplotlib.axes.Axes`, optional
            Matplotlib axes to draw the violins and error bars. Default is None.
```

Then inside the function:
`ax = plt.gca() if ax is None else ax`

gammapy/visualization/violin.py
```diff
@@ -0,0 +1,333 @@
...
+    energy_edges : array-like or `~astropy.units.Quantity`
```
Astro-Kirsty:
```suggestion
    energy_edges : `~numpy.ndarray` or `~astropy.units.Quantity`
```

gammapy/visualization/violin.py
```diff
@@ -0,0 +1,333 @@
...
+        Monotonically increasing energy bin edges of length ``nbins + 1``.
+        Must be positive and finite for log scaling.
+    samples_per_band : list of `~astropy.units.Quantity`
+        List of sample arrays, one per energy bin. Each array contains
+        draws from the flux posterior (or other distribution) in that bin.
+    weights_per_band : list of array-like, optional
```
Astro-Kirsty:
```suggestion
    weights_per_band : list of `~numpy.ndarray`, optional
```

Also why does this have to be a list of array?

Astro-Kirsty: If optional, you should name the default values

gammapy/visualization/violin.py
```diff
@@ -0,0 +1,333 @@
...
+        Per-sample weights for each bin, same length as ``samples_per_band``.
+        If omitted, uniform weights are assumed. Requires NumPy ≥ 2.0 for
+        percentile computation with weights.
+    energy_power : float, optional
+        Power-law scaling applied as ``E_center**energy_power * flux``.
+        Default is None.
```
Astro-Kirsty: We should define this in the same way that we do for [SpectralModel.plot ](https://docs.gammapy.org/dev/api/gammapy.modeling.models.SpectralModel.html#gammapy.modeling.models.SpectralModel.plot)for consistency, both in the docstring and inside the code itself

gammapy/visualization/violin.py
```diff
@@ -0,0 +1,333 @@
...
+    bw_method : str or float, optional
+        Bandwidth selection passed to ``scipy.stats.gaussian_kde``.
```
Astro-Kirsty:
```suggestion
        Bandwidth selection passed to `~scipy.stats.gaussian_kde`.
```

gammapy/visualization/violin.py
```diff
@@ -0,0 +1,333 @@
...
+    grid_size : int, optional
+        Number of evaluation points in log-flux space for the KDE grid.
+    errorbar_kwargs : dict, optional
+        Keyword arguments forwarded to ``ax.errorbar`` for the quantile bars.
```
Astro-Kirsty:
```suggestion
        Keyword arguments forwarded to `~matplotlib.pyplot.errorbar` for the quantile bars.
```

gammapy/visualization/violin.py
```diff
@@ -0,0 +1,333 @@
...
+    errorbar_ul_kwargs : dict, optional
+        Keyword arguments forwarded to ``ax.errorbar`` for the ul bars.
```
Astro-Kirsty:
```suggestion
        Keyword arguments forwarded to `~matplotlib.pyplot.errorbar` for the upper limit bars.
```

gammapy/visualization/violin.py
```diff
@@ -0,0 +1,333 @@
...
+    violin_kwargs : dict, optional
+        Keyword arguments forwarded to ``ax.fill`` for violin plot.
```
Astro-Kirsty:
```suggestion
        Keyword arguments forwarded to `~matplotlib.axes.Axes.fill` for violin plot.
```

gammapy/visualization/violin.py
```diff
@@ -0,0 +1,333 @@
...
+    violin_clip : (float, float), optional
+        Lower and upper containment fractions (in [0, 1]) used to clip
+        the violin tails in log-space. If omitted, defaults to
+        ``(norm.cdf(-4), norm.cdf(4))``.
+    y_label : str, optional
+        Base label for the y-axis, before unit and scaling prefixes.
+
+    Returns
+    -------
+    artists : list
```
Astro-Kirsty: For a plot function I would return the `ax` not the artists. Otherwise that is something else.

The user can then simply do the usual:
`ax = plot_samples_violin_vs_energy(....)`
and adjust anything they see fit

gammapy/visualization/violin.py
```diff
@@ -0,0 +1,333 @@
...
+      log-energy.
+    - Violin clipping is applied only to the KDE grid, not to the
+      quantile-based error bars.
+    - Negative flux samples are replaced by a small positive surrogate
+      (10% of the smallest positive sample) to allow log-space evaluation.
+    - The plotted error bars correspond to the 16%, 50%, and 84%
+      weighted percentiles of the *unclipped* samples, unless the median
+      is non-positive, in which case an upper limit marker is drawn
+      at the 98% percentile. For a gaussian distribution these values corresponds to
+      1σ errors and 2σ upper limit plotted by default by the `FluxPoints.plot` method.
```
Astro-Kirsty: Link here

gammapy/visualization/violin.py
```diff
@@ -0,0 +1,333 @@
...
+
+    # clip defaults
+    if violin_clip is None:
+        violin_clip = (norm.cdf(-4), norm.cdf(4))
+    lc, uc = violin_clip
+    if not (0 <= lc < uc <= 1):
+        raise ValueError("violin_clip must satisfy 0 <= low < high <= 1.")
+
+    # Errorbar styles
+    if errorbar_kwargs is None:
```
Astro-Kirsty: The usual way for setting kwargs defaults is like this:
```
interp_kwargs = interp_kwargs or {}
interp_kwargs.setdefault("values_scale", "log")
```

gammapy/visualization/violin.py
AtreyeeS: @Astro-Kirsty The `sed_type` makes sense only for flux-type quantities. This function can be useful for any (energy dependent) quantities, so the function was renamed

docs/release-notes/6389.feature.rst
QRemy:
```suggestion
Add `~gammapy.visualizzation.plot_samples_violin_vs_energy` to create a violin plot from samples of a given quantity as a function of energy.`
```

gammapy/visualization/violin.py
```diff
@@ -0,0 +1,333 @@
...
+    energy_edges : array-like or `~astropy.units.Quantity`
+        Monotonically increasing energy bin edges of length ``nbins + 1``.
+        Must be positive and finite for log scaling.
+    samples_per_band : list of `~astropy.units.Quantity`
+        List of sample arrays, one per energy bin. Each array contains
+        draws from the flux posterior (or other distribution) in that bin.
+    weights_per_band : list of array-like, optional
+        Per-sample weights for each bin, same length as ``samples_per_band``.
+        If omitted, uniform weights are assumed. Requires NumPy ≥ 2.0 for
+        percentile computation with weights.
```
QRemy:
```suggestion
        If omitted, uniform weights are assumed.
```

gammapy/visualization/violin.py
```diff
@@ -0,0 +1,333 @@
...
+    # clip defaults
+    if violin_clip is None:
+        violin_clip = (norm.cdf(-4), norm.cdf(4))
+    lc, uc = violin_clip
+    if not (0 <= lc < uc <= 1):
+        raise ValueError("violin_clip must satisfy 0 <= low < high <= 1.")
+
+    # Errorbar styles
+    if errorbar_kwargs is None:
+        errorbar_kwargs = dict()
```
QRemy:
```suggestion
    errorbar_kwargs = errorbar_kwargs or {}
```

gammapy/visualization/violin.py
```diff
@@ -0,0 +1,333 @@
...
+        color="black",
+        capsize=2.5,
+        elinewidth=1.2,
+        lw=1.2,
+    )
+    for key in errorbar_kwargs_defaults.keys():
+        errorbar_kwargs.setdefault(key, errorbar_kwargs_defaults[key])
+
+    if errorbar_ul_kwargs is None:
+        errorbar_ul_kwargs = dict()
```
QRemy:
```suggestion
    errorbar_ul_kwargs = errorbar_ul_kwargs or {}
```

docs/release-notes/6389.feature.rst
```diff
@@ -0,0 +1 @@
+Add `~gammapy.visualizzation.plot_samples_violin_vs_energy` to create a violin plot from samples of a given quantity as a function of energy.`
```
QRemy:
```suggestion
Add `~gammapy.visualization.plot_samples_violin_vs_energy` to create a violin plot from samples of a given quantity as a function of energy.`
```

Issue #5909, opened 2025-06-11
Warnings when using EffectiveArea2D.peek


maxnoe:
**Gammapy version**

```
❯ gammapy info

System:

        python_executable      : /home/maxnoe/.local/conda/envs/ctao-school25/bin/python
        python_version         : 3.12.10
        machine                : x86_64
        system                 : Linux


Gammapy package:

        version                : 1.3
        path                   : /home/maxnoe/.local/conda/envs/ctao-school25/lib/python3.12/site-packages/gammapy

INFO:numexpr.utils:NumExpr defaulting to 14 threads.

Other packages:

        numpy                  : 2.2.6
        scipy                  : 1.15.2
        astropy                : 6.1.7
        regions                : 0.10
        click                  : 8.2.1
        yaml                   : 6.0.2
        IPython                : 9.2.0
        jupyterlab             : 4.4.3
        matplotlib             : 3.10.3
        pandas                 : 2.2.3
        healpy                 : not installed
        iminuit                : 2.31.1
        sherpa                 : not installed
        naima                  : not installed
        emcee                  : not installed
        corner                 : not installed
        ray                    : not installed


Gammapy environment variables:

        GAMMAPY_DATA           : /home/maxnoe/Projects/gammapy/gammapy-datasets/dev
```

**Bug description**
A short description what the bug is (usually 1-2 sentences)

I get these two warnings when using the `peek` method of an `EffectiveArea2D` object:

```
irfs["aeff"].peek()

/home/maxnoe/.local/conda/envs/ctao-school25/lib/python3.12/site-packages/gammapy/irf/effective_area.py:105: UserWarning: This axis already has a converter set and is updating to a potentially incompatible converter
  ax.plot(energy_axis.center, area, label=label, **kwargs)
/home/maxnoe/.local/conda/envs/ctao-school25/lib/python3.12/site-packages/gammapy/irf/effective_area.py:147: UserWarning: This axis already has a converter set and is updating to a potentially incompatible converter
  ax.plot(offset_axis.center, area, label=label, **kwargs)
```

**Expected behavior**
No warnings.

**To Reproduce**

```
from gammapy.data import Observation
obs = Observation.read("$GAMMAPY_DATA/hess-dl3-dr1/data/hess_dl3_dr1_obs_id_020136.fits.gz")
obs.aeff.peek()
```

morcuended:
[...]

PR #5659, opened 2025-01-05
Avoid matplotlib 3.10 in CI

QRemy:
Fix CI sphinx all-deps : fix the following error in doc build :
```
Traceback (most recent call last):
  File "/home/runner/work/gammapy/gammapy/.tox/build_docs/lib/python3.10/site-packages/sphinx/config.py", line 529, in eval_config_file
    exec(code, namespace)  # NoQA: S102
  File "/home/runner/work/gammapy/gammapy/docs/conf.py", line 43, in <module>
    from gammapy.utils.docs import SubstitutionCodeBlock, gammapy_sphinx_ext_activate
  File "/home/runner/work/gammapy/gammapy/.tox/build_docs/lib/python3.10/site-packages/gammapy/utils/docs.py", line 28, in <module>
    from gammapy.analysis import AnalysisConfig
  File "/home/runner/work/gammapy/gammapy/.tox/build_docs/lib/python3.10/site-packages/gammapy/analysis/__init__.py", line 3, in <module>
    from .config import AnalysisConfig
  File "/home/runner/work/gammapy/gammapy/.tox/build_docs/lib/python3.10/site-packages/gammapy/analysis/config.py", line 12, in <module>
    from gammapy.makers import MapDatasetMaker
  File "/home/runner/work/gammapy/gammapy/.tox/build_docs/lib/python3.10/site-packages/gammapy/makers/__init__.py", line 3, in <module>
    from .background import (
  File "/home/runner/work/gammapy/gammapy/.tox/build_docs/lib/python3.10/site-packages/gammapy/makers/background/__init__.py", line 2, in <module>
    from .fov import FoVBackgroundMaker
  File "/home/runner/work/gammapy/gammapy/.tox/build_docs/lib/python3.10/site-packages/gammapy/makers/background/fov.py", line 5, in <module>
    from gammapy.maps import Map, RegionGeom
  File "/home/runner/work/gammapy/gammapy/.tox/build_docs/lib/python3.10/site-packages/gammapy/maps/__init__.py", line 10, in <module>
    from .region import RegionGeom, RegionNDMap
  File "/home/runner/work/gammapy/gammapy/.tox/build_docs/lib/python3.10/site-packages/gammapy/maps/region/__init__.py", line 2, in <module>
    from .geom import RegionGeom
  File "/home/runner/work/gammapy/gammapy/.tox/build_docs/lib/python3.10/site-packages/gammapy/maps/region/geom.py", line 11, in <module>
[...]

PR #6398, opened 2026-02-16
Address UserWarning for setting units converter several times, lift matplotlib version upper bound, require astropy>=6.1.5

morcuended:
**Description**
This pull request tries to address #5909.

Remove the use of `quantity_support` and use `.values` for plotting quantities using the units from `MapAxis` to avoid `UserWarning` for registering unit converters multiple times.

**Dear reviewer**
To avoid all `UserWarning` occurrences, similar changes need to be done in various places. Here I just touch the one for  `EffectiveArea`

All occurrences can be checked by adding `"error:This axis already has a converter set:UserWarning"` in pyproject.toml
```
[tool.pytest.ini_options]
filterwarnings = [...]
```

I also set astropy>=6.1.5 in requirements to avoid AnchoredEllipse deprecation


morcuended:
Not the only issue with matplotlib 3.10, see #5659

`AnchoredEllipse` deprecation was fixed for astropy >= 6.1.5
https://github.com/astropy/astropy/pull/17144

bkhelifi:
Thanks @morcuended, there is still a failing test... Otherwise, do not forget to sign your commits (see https://docs.gammapy.org/dev/development/intro.html#acceptation-of-the-developer-certificate-of-origin-dco)

QRemy:
Is the fail  on devdeps related ? 
it seems to pass on the other PRs that run recently.

morcuended:
There is something off indeed, related to matplotlib

```
FAILED ../../.tox/devdeps/lib/python3.12/site-packages/gammapy/visualization/tests/test_datasets.py::test_plot_spectrum_datasets_off_regions - AssertionError: assert (0.17254901960784313, 0.6274509803921569, 0.17254901960784313) == '#2ca02c'
 +  where (0.17254901960784313, 0.6274509803921569, 0.17254901960784313) = get_color()
 +    where get_color = <matplotlib.lines.Line2D object at 0x7f584aa808f0>.get_color
```
Better to understand it before merging

registerrier (commented):
Thanks @morcuended . Can you clarify the issue here. The multiple converter settings are caused by the successive use of `quantity_support` and `format_plot_axis`?
I am not sure to understand why you have to change the logic here.

morcuended:
No, the warning comes from subsequent uses of `quantity_support` on the same axes within `plot_energy_dependence`, `plot_offset_dependence` and therefore also `peek`. Each time, `quantity_support` register a new converter instance on the same axes object. Using the context manager does not clean up the converters after them.

I had to change `format_plot_axis` because after removing `quantity_support`, `ax.xaxis.units` is then None. The change was needed because instead of reading the unit label from `ax.xaxis.units`, only defined if `quantity_support` is used, now it reads it directly from the MapAxis units.

gammapy/irf/effective_area.py
```diff
@@ -99,12 +98,11 @@ def plot_energy_dependence(self, ax=None, offset=None, **kwargs):
...
             area = self.evaluate(offset=off, energy_true=energy_axis.center)
             label = kwargs.pop("label", f"offset = {off:.1f}")
-            with quantity_support():
-                ax.plot(energy_axis.center, area, label=label, **kwargs)
+            ax.plot(energy_axis.center, area, label=label, **kwargs)
 
         energy_axis.format_plot_xaxis(ax=ax)
         ax.set_ylabel(
-            f"Effective Area [{ax.yaxis.units.to_string(UNIT_STRING_FORMAT)}]"
+            f"Effective Area [{energy_axis.unit.to_string(UNIT_STRING_FORMAT)}]"
```
k-egg, line 105: I don't think an energy unit is the correct unit for effective area.

k-egg: It should be `self.unit.to_string(UNIT_STRING_FORMAT)` instead, shouldn't it?

morcuended: Indeed, good catch, thanks

morcuended:
As discussed during the dev call, this fix of removing quantity_support and relying on MapAxis units should be done in multiple places, which would require quite some effort and make sure that it does not have side effects. Instead, the suggestion was to catch and silence the UserWarning.

morcuended:
@registerrier, another posibility could be to move the loops inside the same `quantity_support` context:

Do:
```python
        with quantity_support():
            for off in offset:
                area = self.evaluate(offset=off, energy_true=energy_axis.center)
                label = kwargs.pop("label", f"offset = {off:.1f}")
                ax.plot(energy_axis.center, area, label=label, **kwargs)
```
instead of:
```python
        for off in offset:
            area = self.evaluate(offset=off, energy_true=energy_axis.center)
            label = kwargs.pop("label", f"offset = {off:.1f}")
            with quantity_support():
                ax.plot(energy_axis.center, area, label=label, **kwargs)
```

PR #6407, opened 2026-02-17
Modify info script to use importlib.metadata and fix deprecation warning

### LINKED ISSUE #6404: Modify gammapy info script to use importlib.metadata
`gammapy info` prints the version of all installed dependencies of gammapy. It relies on `get_info_dependencies()` in gammapy.scripts.info.py. It looks for a `__version__` attribute. This results in the following deprecation warning:
```
 /Users/terrier/Code/gammapy-dev/gammapy/gammapy/scripts/info.py:108: DeprecationWarning: The '__version__' attribute is deprecated and will be removed in Click 9.1. Use feature detection or 'importlib.metadata.version("click")' instead.
    module_version = getattr(module, "__version__", "no version info found")
 ```

We should rely on `importlib.metadata.version(dependency)` instead.

Basmala92:
Fixes #6404
This PR replaces the  `__version__` attribute access with `importlib.metadata.version` in the `info` script.

### DIFF
--- a/gammapy/maps/tests/test_core.py
+++ b/gammapy/maps/tests/test_core.py
@@ -356,7 +356,8 @@ def test_map_arithmetics(map_type):
     gt_m2 = m2 > 15000 * u.cm**2
     assert_allclose(gt_m2, False)
 
-    ge_m2 = m2 >= m2
+    m2_copy = m2.copy()
+    ge_m2 = m2 >= m2_copy
     assert_allclose(ge_m2, True)
 
     eq_m2 = m2 == 500 * u.cm**2
@@ -382,7 +383,8 @@ def test_boolean_arithmetics():
     m_not = ~m_2
     assert np.all(m_not.data)
 
-    m_xor = m_1 ^ m_1
+    m_1_copy = m_1.copy()
+    m_xor = m_1 ^ m_1_copy
     assert not np.any(m_xor.data)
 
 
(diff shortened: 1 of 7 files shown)


registerrier:
Thanks @Basmala92 !
Could you sign one of your commits (e.g. with git commit -s ) as we need this to extract the author list when releasing the software:
https://docs.gammapy.org/2.0.1/development/intro.html#acceptation-of-the-developer-certificate-of-origin-dco

Basmala92:
@registerrier Done!
I have amended the commit with the sign-off , Thanks for the guidance!

gammapy/scripts/info.py
registerrier: You might uncomment this.

registerrier: Does this work? The package name is pyyaml

registerrier: You might use a similar logic here

registerrier: Same here

Basmala92: yes the package is actually PyYAML not yaml , I will fix it and do the other recommendations.

Basmala92:
@registerrier For the logic in `get_info_version` I applied single `try-except` that handles both version and path.
Currently, if either fails, both will default to `"unknown"`. Does this approach work for you, or would it be better if  I handle them independently to allow displaying one even if the other fails?

registerrier:
Thanks @Basmala92 . Looks good. Merging now.

PR #6411, opened 2026-02-18
Fix SonarQube relability for `Priors`

Astro-Kirsty:
This fixes a number of minor issues. Firstly, we did not have the `GeneralizedGaussianPrior` included in the registry therefore the user cannot write the model correctly. 

This became noticeable as we were only testing one prior in the `test_serialisation`. By fixing this, it also fixes the SQ error.

The `test_uniform_prior_weight` also only needs to test one option. 

Finally, the kwargs for `from_dict` were not defined correctly. Another approach would be to set `kwargs.setdefault("weight", data["weight"])` but this means the user can over write which I don't think is the intended meaning.

Associated to the clean up for https://github.com/gammapy/gammapy/issues/6239

### DIFF
--- a/gammapy/modeling/models/prior.py
+++ b/gammapy/modeling/models/prior.py
@@ -132,8 +132,6 @@ class Prior(ModelBase):
     @classmethod
     def from_dict(cls, data, **kwargs):
         """Get prior parameters from dictionary."""
-        kwargs = {}
-
         key0 = next(iter(data))
         if key0 in ["prior"]:
             data = data[key0]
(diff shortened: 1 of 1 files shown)


gammapy/modeling/models/prior.py
```diff
@@ -145,8 +143,9 @@ def from_dict(cls, data, **kwargs):
         priorparameters = _build_priorparameters_from_dict(
             data["parameters"], cls.default_parameters
         )
-        kwargs["weight"] = data["weight"]
-        return cls.from_parameters(priorparameters, **kwargs)
+        init_kwargs = dict(kwargs)
```
registerrier, line 146: I am not sure constructing a new dict really necessary. But fine.

PR #6412, opened 2026-02-18
Fix SonarQube reliability issue with reassigned variables

MSCarrasco:
This PR fixes MINOR issues from #6239 about the reassignment of a variable given in argument

1. `gammapy/maps/wcs/core.py:114` the `format` is identified in the method so it was removed from the arguments
2. `gammapy/modeling/models/core.py:1355` and `gammapy/modeling/parameter.py:924` the `value` is obtained in the method so it was removed from the arguments. Also `type` and `traceback` were unused so they were removed.

### DIFF
--- a/gammapy/maps/wcs/core.py
+++ b/gammapy/maps/wcs/core.py
@@ -111,7 +111,7 @@ class WcsMap(Map):
             raise ValueError(f"Invalid map type: {map_type!r}")
 
     @classmethod
-    def from_hdulist(cls, hdu_list, hdu=None, hdu_bands=None):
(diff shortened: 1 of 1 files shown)


MSCarrasco:
The variables removed from `gammapy/modeling/models/core.py:1355` and `gammapy/modeling/parameter.py:924` were in fact decorators, I put them back with more explicit names.

The `format` argument in `gammapy/maps/wcs/core.py:114` is in fact needed but was set to None as default instead of "gadf". Now the format is identified from the header only if format is None. If the header is None, then default is "gadf".

registerrier (approved):
Thanks @MSCarrasco . That's a nice first PR (albeit more tricky than it looks) on gammapy.

PR #6415, opened 2026-02-18
Fix SonarQube reliability issue with dict key access

MSCarrasco:
This PR fixes a critical issue from #6239 about the access on a dictionary key that may be missing.

The `MapDatasetOnOff.to_spectrum_dataset()` method now declares 'counts_off' and 'acceptance_off' set to None in the kwargs dictionary.
If the `dataset.counts_off is not None` the dictionary is updated. 
I added a condition so that if `dataset.counts_off is None` the `acceptance_off` is not computed since it needs `counts_off`. 

I added a test function to assert that the method outputs a spectrum_dataset where `counts_off` and `acceptance_off` have None values.


gammapy/datasets/map.py
```diff
@@ -3223,7 +3223,7 @@ def to_spectrum_dataset(self, on_region, containment_correction=False, name=None
             name=name,
         )
 
-        kwargs = {"name": name}
+        kwargs = {"name": name, "counts_off": None, "acceptance_off": None}
```
Astro-Kirsty, line 3226: Is it necessary to add these here?

Maybe the logic is better with using `self.counts_off` as in the lines on 3228.

i.e. calling:
```
if self.acceptance is not None:
     <<CODE>>
     if self.counts_off is not None:
```

Astro-Kirsty (changes requested):
Thank you @MSCarrasco.

Just a small suggestion

gammapy/datasets/map.py
MSCarrasco: I implemented your suggestion and removed the "counts_off" key, but the "acceptance_off" key was left because the test function showed it is passed as a required positional argument when counts_off is None.

MSCarrasco: SQ complained again about the "counts_off" key that can be missing, so I reverted back to the kwargs initialization.
I kept the condition logic you suggested

gammapy/datasets/tests/test_map.py
```diff
@@ -1462,6 +1462,18 @@ def get_map_dataset_onoff(images, **kwargs):
...
+@requires_data()
+def test_mapdataset_on_off_to_spectrum_dataset_with_no_counts_off(images):
+    dataset = get_map_dataset_onoff(images)
+    dataset.counts_off = None
+    on_region = CircleSkyRegion(
+        center=dataset.counts.geom.center_skydir, radius=0.1 * u.deg
+    )
+    spectrum_dataset = dataset.to_spectrum_dataset(on_region)
+    assert spectrum_dataset.counts_off is None
+    assert spectrum_dataset.acceptance_off is None
```
registerrier, line 1475: Here you might add an assert on `spectrum_dataset.acceptance`  which should exist.

registerrier (commented):
Thanks @MSCarrasco . Looks good, I have left a minor comment.

gammapy/datasets/tests/test_map.py
MSCarrasco: Thank you, it is done

PR #6420, opened 2026-02-19
refactor: remove linting dependencies

### LINKED ISSUE #6059: Get rid of linting/formatting settings usings flake8, isort, black, pydocstyle
Since we now use ruff for the linting and formatting (which is enforced by a pre-commit hook), I think we should remove all leftovers from lake8, isort, black, pydocstyle in the project config, Makefile, etc

natthan-pigoux:
**Description**
 Remove linting dependencies handled by `ruff` and related Makefile and docs references.

**Dear reviewer**

I removed the dependencies from the `environment-dev.yml` assuming that this file was built by hand.

One question is what to do with this: https://github.com/gammapy/gammapy/blob/main/docs/development/doc_howto.rst#docstring-formatting

Closes #6059

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,9 +21,6 @@ help:
 	@echo '     docs-show          Open local HTML docs in browser'
 	@echo ''
 	@echo '     trailing-spaces    Remove trailing spaces at the end of lines in *.py files'
-	@echo '     black              Run black code formatter'
-	@echo '     isort              Run isort code formatter to sort imports'
-	@echo '     polish             Run trailing-spaces, black and isort'
 	@echo '     pylint             Run pylint static code analysis'
 	@echo ''
 	@echo ' To get info about your Gammapy installation and setup run this command'
@@ -80,14 +77,6 @@ docs-show:
 trailing-spaces:
 	find $(PROJECT) examples docs -name "*.py" -exec perl -pi -e 's/[ \t]*$$//' {} \;
 
-black:
-	black $(PROJECT)
-
-isort:
-	isort -rc gammapy examples docs -s docs/conf.py
-
-polish: black isort trailing-spaces;
-
 # TODO: once the errors are fixed, remove the -E option and tackle the warnings
 # Note: pylint is very thorough, but slow, and has false positives or nitpicky stuff
 pylint:
@@ -103,11 +92,5 @@ pylint:
 # D400: First line should end with a period (not ')')
 # D401: First line should be in imperative mood; try rephrasing (found 'Function')
 # D403: First word of the first line should be properly capitalized ('Add', not 'add')
-pydocstyle:
-	pydocstyle $(PROJECT) \
-	--convention=numpy \
-	--match-dir='^(?!extern).*' \
-	--match='(?!test_).*\.py' \
-	--add-ignore=D100,D102,D103,D104,D105,D200,D202,D205,D400,D401,D403,D410
 
 # TODO: add test and code quality checks for `examples`
--- a/docs/development/doc_howto.rst
+++ b/docs/development/doc_howto.rst
@@ -189,22 +189,6 @@ Another option is to create a general list of references, as follows:
         * `Author et al. (2023), "Title" <link_to_nasaads>`_
         * `Author2 et al. (2022), "Title2" <link_to_nasaads>`_
 
-.. _docstring-formatting:
-
-Docstring formatting
-^^^^^^^^^^^^^^^^^^^^
-
-It is also important to check that you have correctly formatted your docstring.
-An easy way to check this is by utilising `pydocstyle`. `pydocstyle` utilises the
-PEP257 convention by default, which means a number of errors are automatically ignored.
-In gammapy we chose to opt for the `numpy` convention therefore the flag must be added
-to correctly check the docstrings. To check the docstring for your specific file, i.e.:
-
-.. code-block:: bash
-
-    pydocstyle --convention=numpy gammapy/data/event_list.py
-
-
 
 Sphinx gallery extension
 ------------------------
--- a/docs/development/doc_howto.rst
+++ b/docs/development/doc_howto.rst
@@ -189,6 +189,13 @@ Another option is to create a general list of references, as follows:
         * `Author et al. (2023), "Title" <link_to_nasaads>`_
         * `Author2 et al. (2022), "Title2" <link_to_nasaads>`_
 
+.. _docstring-formatting:
+
+Docstring formatting
+^^^^^^^^^^^^^^^^^^^^
+
+The docstring formatting is done via `ruff <https://docs.astral.sh/ruff/>`__ tool, which runs as a ``pre-commit`` hook and follow the `numpy docstring style convention <https://numpydoc.readthedocs.io/en/latest/format.html>`__.
+``ruff`` include most of `pydocstyle rules <https://docs.astral.sh/ruff/rules/#pydocstyle-d>`__ except those explicitly excluded in ``pyproject.toml``.
 
 Sphinx gallery extension
 ------------------------
--- a/docs/development/doc_howto.rst
+++ b/docs/development/doc_howto.rst
@@ -194,8 +194,11 @@ Another option is to create a general list of references, as follows:
 Docstring formatting
 ^^^^^^^^^^^^^^^^^^^^
 
-The docstring formatting is done via `ruff <https://docs.astral.sh/ruff/>`__ tool, which runs as a ``pre-commit`` hook and follow the `numpy docstring style convention <https://numpydoc.readthedocs.io/en/latest/format.html>`__.
-``ruff`` include most of `pydocstyle rules <https://docs.astral.sh/ruff/rules/#pydocstyle-d>`__ except those explicitly excluded in ``pyproject.toml``.
+The docstring formatting is handled via the `ruff <https://docs.astral.sh/ruff/>`__
+tool, which runs as a ``pre-commit`` hook and follows the
+`numpy docstring style convention <https://numpydoc.readthedocs.io/en/latest/format.html>`__.
+``ruff`` includes most of the `pydocstyle rules <https://docs.astral.sh/ruff/rules/#pydocstyle-d>`__
+except those explicitly excluded in ``pyproject.toml``.
 
 Sphinx gallery extension
 ------------------------
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -186,6 +186,9 @@ skip-magic-trailing-comma = false
 # Like Black, automatically detect the appropriate line ending.
 line-ending = "auto"
 
+[tool.ruff.lint.isort]
+skip = ["docs/conf.py"]
+
 # Including towncrier to create the changelog
 [tool.towncrier]
 package = "gammapy"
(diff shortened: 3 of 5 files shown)


Makefile
```diff
@@ -103,11 +92,5 @@ pylint:
 # D400: First line should end with a period (not ')')
 # D401: First line should be in imperative mood; try rephrasing (found 'Function')
 # D403: First word of the first line should be properly capitalized ('Add', not 'add')
-pydocstyle:
-	pydocstyle $(PROJECT) \
-	--convention=numpy \
-	--match-dir='^(?!extern).*' \
-	--match='(?!test_).*\.py' \
-	--add-ignore=D100,D102,D103,D104,D105,D200,D202,D205,D400,D401,D403,D410
```
morcuended, line 111: Make sure that these settings (convention, ignores) match ruff rules set for the repository.

natthan-pigoux: Right forgot about these, thanks ! Should be added in ruff config now.

pyproject.toml
```diff
@@ -136,14 +136,23 @@ except Exception:
...
 
 [tool.ruff.lint]
-ignore = ["E741"]
-extend-ignore = ["E203","E712"]
+select = ["D", "E"]
+ignore = [
+    "D100", "D102", "D103",
+    "D104", "D105", "D200",
+    "D202", "D205", "D400",
+    "D401", "D403", "D410",
```
natthan-pigoux: These rules were copy paste from the `setup.cfg` originally.
Here are the references:
* [E rules](https://docs.astral.sh/ruff/rules/#error-e)
* [D rules](https://docs.astral.sh/ruff/rules/#pydocstyle-d)

Shall we keep them as they are? or should we have a look and limit the number of ignored?

Astro-Kirsty: I am going through them now and will write a summary here soon

Astro-Kirsty: So we have a number we need to keep and some we can remove now. The others I made a separate issue for #6434

- D102: Keep. We don't always want a docstring, as some functions are inherited from the parent class
- D104: Keep, typically comes from `__init__.py`
- D105: Keep, comes from magic methods i.e. `__str__`
- D200: Remove (no refactor needed)
- E203: Remove (no refactor needed)
- E712: Remove (no refactor needed)
- E741: Remove (no refactor needed)

natthan-pigoux: Thanks a lot for going through them! I'll update them here then.

natthan-pigoux:
I've forgot to add most of the linting tools in ruff config. I added almost all of them. Please confirm that we want to use them.

Note that I tried to run pre-commit on a few files and the most recurrent error is `E501 Line too long (92 > 88)`  which we can be tuned here:
```
 [tool.ruff]
 exclude = ["docs", "dev", "extern*", "*test_*.py"]
 # Like black
 line-length = 88
```

Astro-Kirsty:
If we want to utilise the PEP-8 standards, then the max line length is 79.

We adhere to the PEP8 coding style --> [from here](https://docs.gammapy.org/dev/development/intro.html#get-set-up)

pyproject.toml
```diff
@@ -138,14 +138,33 @@ except Exception:
 """
 
 [tool.ruff]
-exclude = ["docs", "dev"]
+exclude = ["docs", "dev", "extern*", "*test_*.py"]
 # Like black
-line-length = 88
+line-length = 79
```
registerrier: will this require reformatting all files?

pyproject.toml
```diff
@@ -138,14 +138,33 @@ except Exception:
 """
 
 [tool.ruff]
-exclude = ["docs", "dev"]
+exclude = ["docs", "dev", "extern*", "*test_*.py"]
```
registerrier: Is there a reason to exclude tests?

registerrier (commented):
Thanks @natthan-pigoux . See few minor comments.

pyproject.toml
registerrier: We have been using black for a very long time with the 88 line length limit. Even if this is not fully consistent with the dev doc I think we should keep this value.

Astro-Kirsty: I suggested this as the 88 with black is a random choice by them and not with any kind of standards. I know it will probably touch lots of files, happy to do as you see fit

pyproject.toml
natthan-pigoux: No, I think this is a remnant from the old config indeed. I'll remove it.

natthan-pigoux: Shall I even remove `docs` and `dev`?

pyproject.toml
natthan-pigoux: Just to notice that in DPPS we usually use 88. I think it is up to us to decide if 79 is enough or if it splits some code lines we don't want to be split for readability.

morcuended: I agree with keeping the 88 limit

pyproject.toml
morcuended: I wouldn't exclude any of those first and see what ruff checks say. In general all code could benefit from it.

pyproject.toml
```diff
@@ -138,14 +138,33 @@ except Exception:
...
+  # "UP",   # pyUpgrade
+  "FLY", # flynt
+  "DTZ", # flake8-datetimez
+  "S",   # flake8-bandit
+  "N",   # pep8-naming
+]
+ignore = [
+    "D100", "D102", "D104", "D105",
+    "D400", "D401", "D403",
+    "S101",
```
morcuended: Why ignoring S101? And what about all the rules that appear in Makefile command?

Makefile
```diff
@@ -80,14 +77,6 @@ docs-show:
 trailing-spaces:
 	find $(PROJECT) examples docs -name "*.py" -exec perl -pi -e 's/[ \t]*$$//' {} \;
 
-black:
-	black $(PROJECT)
-
-isort:
-	isort -rc gammapy examples docs -s docs/conf.py
```
morcuended, line 87: This needs then something like:

```
[tool.ruff.lint.isort]
skip = ["docs/conf.py"]
```

Makefile
```diff
@@ -103,11 +92,5 @@ pylint:
 # D400: First line should end with a period (not ')')
 # D401: First line should be in imperative mood; try rephrasing (found 'Function')
 # D403: First word of the first line should be properly capitalized ('Add', not 'add')
-pydocstyle:
-	pydocstyle $(PROJECT) \
-	--convention=numpy \
-	--match-dir='^(?!extern).*' \
-	--match='(?!test_).*\.py' \
```
morcuended, line 110: this should translate into ruff settings:

```
[tool.ruff.lint.per-file-ignores]
"test_*.py" = ["D"]
```

morcuended (changes requested):
Thanks @natthan-pigoux. I left some comments.

pyproject.toml
natthan-pigoux: For "S101", no particular reason, that's one about "assert" syntax, but indeed in gammapy there is no assertion which do not respect this. I will re-enable it. 

> And what about all the rules that appear in Makefile command?

Which ones?

registerrier:
Because this is not key ingredient for v2.1, I would propose to wait for the v2.1.x branching before merging this. Is it fine @natthan-pigoux @morcuended ?
Hence changing milestone from 2.1 to 2.2

pyproject.toml
```diff
@@ -143,17 +143,35 @@ except Exception:
...
+  # "UP",   # pyUpgrade
+  "FLY", # flynt
+  "DTZ", # flake8-datetimez
+  "S",   # flake8-bandit
+  "N",   # pep8-naming
+]
+ignore = [
+    "D100", "D102", "D104", "D105",
+    "D400", "D401", "D403",
+]
```
Astro-Kirsty, line 167: We should change these once they are decided in #6434

Astro-Kirsty: The decision is to keep the ones in this PR currently

Astro-Kirsty (dismissed):
Looks good to me, thanks @natthan-pigoux!

pyproject.toml
Astro-Kirsty: Our convention for `pydocstyle` is numpy, is that some how implemented with this?

morcuended: see below:

```
[tool.ruff.lint.pydocstyle]
convention = "numpy"
```

Astro-Kirsty: Sorry, I didnt look far enough

registerrier (dismissed):
Thanks @natthan-pigoux . This looks good to me. No further comment.

Astro-Kirsty (commented):
Thanks @natthan-pigoux. Can you make sure we do not reference to `docstring-formatting` elsewhere in the docs?

natthan-pigoux:
> Can you make sure we do not reference to docstring-formatting elsewhere in the docs?

Sure, I'll check that.

docs/development/doc_howto.rst
```diff
@@ -189,22 +189,6 @@ Another option is to create a general list of references, as follows:
         * `Author et al. (2023), "Title" <link_to_nasaads>`_
         * `Author2 et al. (2022), "Title2" <link_to_nasaads>`_
 
-.. _docstring-formatting:
-
-Docstring formatting
-^^^^^^^^^^^^^^^^^^^^
-
-It is also important to check that you have correctly formatted your docstring.
```
bkhelifi, line 197: This is sad that at the end no instruction is given about the docstring formatting, when in reality we have many rules. I would suggest to describe what we are using and where to find the rules

bkhelifi (changes requested):
Thanks a lot @natthan-pigoux. I have one comment regarding to the instructions for the developpers

pyproject.toml
```diff
@@ -143,17 +143,35 @@ except Exception:
...
+  "PLE", # pylint errors
+  "D",   # pydocstyle
+  # "UP",   # pyUpgrade
+  "FLY", # flynt
+  "DTZ", # flake8-datetimez
+  "S",   # flake8-bandit
+  "N",   # pep8-naming
+]
+ignore = [
+    "D100", "D102", "D104", "D105",
```
Astro-Kirsty, line 165: After investigation on my PR, I think we also need to add `D205`  which is related to `D400` : "1 blank line required between summary line and description (found 0)"

docs/development/doc_howto.rst
natthan-pigoux: You are right that there is no more infos about it. I can keep this section and link it with the ruff rules used.

natthan-pigoux: I rewrite this part of the doc. Is it ok?

pyproject.toml
natthan-pigoux: Thanks for the suggestion.

docs/development/doc_howto.rst
natthan-pigoux: @Astro-Kirsty could you check the english?

registerrier (approved):
Thanks @natthan-pigoux . Looks good!

morcuended (approved):
Thanks @natthan-pigoux, no further comments from my side

registerrier (approved):
Thanks @natthan-pigoux . Looks good

PR #6422, opened 2026-02-20
Drop support for numpy=1.26 and astropy=6.0

QRemy:
Drop support for numpy=1.26 following
https://scientific-python.org/specs/spec-0000/#support-window


tox.ini
morcuended: remove

morcuended (changes requested):
Thanks @QRemy. Update `ci.yml` still referencing numpy126 (`tox -e py310-test-alldeps_noray-astropy60-numpy126`) also update `softwareRequirements` in `codemeta.json`

PR #6425, opened 2026-02-20
Avoid using multiple times `quantity_support` inside loops

morcuended:
**Description**

Alternative solution to #5909

`quantity_support` is used sometimes multiple times within the same function inside loops. Here, I define only one context manager for `quantity_support` per function.

In some places it was already used like this.

**Dear reviewer**

I set `"error:This axis already has a converter set:UserWarning",` to be raised as error by pytest to identify all the occurrences.

### DIFF
--- a/docs/release-notes/6425.bug.rst
+++ b/docs/release-notes/6425.bug.rst
@@ -1,1 +1,0 @@
-Avoid multiple use of ``quantity_support`` within one loop.
--- a/docs/release-notes/6425.bug.rst
+++ b/docs/release-notes/6425.bug.rst
@@ -1,1 +1,1 @@
-Avoid multiple use of ``quantity_support`` in one functionality.
+Avoid multiple use of ``quantity_support`` within one loop.
--- a/docs/release-notes/6425.bug.rst
+++ b/docs/release-notes/6425.bug.rst
@@ -1,0 +1,1 @@
+Avoid multiple use of ``quantity_support`` in one functionality.
--- a/gammapy/maps/region/ndmap.py
+++ b/gammapy/maps/region/ndmap.py
@@ -111,46 +111,46 @@ class RegionNDMap(Map):
         ax : `~matplotlib.pyplot.Axis`
             Axis used for plotting.
         """
-        ax = ax or plt.gca()
+        with quantity_support():
+            ax = ax or plt.gca()
 
-        if axis_name is None:
-            if self.geom.axes.is_unidimensional:
-                axis_name = self.geom.axes.primary_axis.name
-            else:
-                raise ValueError(
-                    "Plotting a region map with multiple extra axes requires "
-                    "specifying the 'axis_name' keyword."
-                )
+            if axis_name is None:
+                if self.geom.axes.is_unidimensional:
+                    axis_name = self.geom.axes.primary_axis.name
+                else:
+                    raise ValueError(
+                        "Plotting a region map with multiple extra axes requires "
+                        "specifying the 'axis_name' keyword."
+                    )
 
-        axis = self.geom.axes[axis_name]
+            axis = self.geom.axes[axis_name]
 
-        kwargs.setdefault("marker", "o")
-        kwargs.setdefault("markersize", 4)
-        kwargs.setdefault("ls", "None")
-        kwargs.setdefault("xerr", axis.as_plot_xerr)
+            kwargs.setdefault("marker", "o")
+            kwargs.setdefault("markersize", 4)
+            kwargs.setdefault("ls", "None")
+            kwargs.setdefault("xerr", axis.as_plot_xerr)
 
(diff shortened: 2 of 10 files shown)


AtreyeeS:
Thanks @morcuended ! I remember we used this in the past as well. Can you resolve the failing tests?

morcuended:
Failing test comes from `plot` method of `RegionNDMap` and `SpectralModel` despite having moved the `quantity_support` context manager outside the loop to be called only once. Still, the problem persists.

In the dev call, it was suggested to silence the warning until the proper fix with the new Plotter class is in place or proper fix is done (as not relying on `quantity_support` but passing directly `.values`). For now, I will simply silence the warning, leaving a FIXME. Let me know if this is the wanted way to proceed.

However, I would not silence the warnings, but leave them to be raised explicitly so you are really aware that there is something fishy here. The failing tests can be made disappear by just removing the warning from the list of errors in pytest configuration.

morcuended:
I leave this here. Please someone take over this issue. Feel free to modify or close it.

AtreyeeS:
Thanks @morcuended ! 
There is a test fail in the devdeps. Apart from that, is there anything else remaining to do here? 
We can make further improvements once the plotters PIG is accepted.

gammapy/modeling/models/spectral.py
```diff
@@ -646,9 +646,16 @@ def plot(
 
         flux = self._get_plot_flux(sed_type=sed_type, energy=energy)
         flux = scale_plot_flux(flux, energy_power=energy_power)
-
-        with quantity_support():
-            ax.plot(energy.center, flux.quantity[:, 0, 0], **kwargs)
+        # FIXME: Get rid of this warning properly. One option is
+        # to explicitly pass values instead of quantities.
+        with warnings.catch_warnings():
```
Astro-Kirsty: I think we should not catch the warning, but rather let it be. 

We can figure out how to solve it in another PR

pyproject.toml
```diff
@@ -122,6 +122,7 @@ filterwarnings = [
     "error::astropy.utils.exceptions.AstropyDeprecationWarning",
     "error::gammapy.utils.deprecation.GammapyDeprecationWarning",
     "error::matplotlib.MatplotlibDeprecationWarning",
+    "error:This axis already has a converter set:UserWarning",
```
Astro-Kirsty: Again, lets not filter the warning

morcuended: This was just added to make this specific warning to appear as error and fail for pytest, so I could spot the occurrences. It can be removed from here.

gammapy/modeling/models/spectral.py
morcuended: Agreed

docs/release-notes/6425.bug.rst
```diff
@@ -0,0 +1 @@
+Avoid multiple use of ``quantity_support`` within one loop.
```
morcuended:
```suggestion
Avoid multiple use of ``quantity_support`` within loops, which was causing matplotlib UserWarning because unit converter were set several times in plot axes.
```

gammapy/maps/region/ndmap.py
```diff
@@ -111,60 +112,74 @@ def plot(self, ax=None, axis_name=None, **kwargs):
...
-                    "Plotting a region map with multiple extra axes requires "
-                    "specifying the 'axis_name' keyword."
+        # FIXME: Get rid of this warning properly. One option is
+        # to explicitly pass values instead of quantities.
+        with warnings.catch_warnings():
+            warnings.filterwarnings(
+                "ignore",
+                message="This axis already has a converter set",
+                category=UserWarning,
+            )
```
morcuended: Remove warning filter and leave the rest within the `quantity_support` context manager.

gammapy/maps/region/ndmap.py
```diff
@@ -8,6 +8,7 @@
 from astropy.table import Table
 from astropy.visualization import quantity_support
 import matplotlib.pyplot as plt
+import warnings
```
morcuended: remove

morcuended (commented):
Thanks @Astro-Kirsty, I think this is the way to go here. Remove the other warning filter and I suggest updating the fragment (maybe no fragment needed at all? It's more like a cleanup?)

docs/release-notes/6425.bug.rst
Astro-Kirsty: I wasn't sure about the fragment, because we currently label it `bug` so I had to add it to get it to run

QRemy: could remove the fragment and label as cleanup

Astro-Kirsty: Whatever you think is best

QRemy (approved):
Thanks @morcuended and @Astro-Kirsty seems ready to merge.

QRemy:

PR #6428, opened 2026-02-21
Replace `coadd` in `WcsNDMap.crop` with direct tensor slicing

kushagarwal2910-lang:
**Description**
This PR resolves a hidden `# FIXME` located in `gammapy/maps/wcs/ndmap.py` within the `WcsNDMap.crop` method. 

Previously, when cropping maps with irregular non-spatial geometries, the code fell back to allocating an empty map and calling `.coadd()`, which evaluates WCS coordinates, interpolates, and iterates, causing a significant and unnecessary performance bottleneck.

I have replaced this fallback with a pure NumPy dynamic slicing algorithm. It calculates the N-dimensional bounding box intersections by extracting the true pixel boundaries for each specific image plane, and dynamically generates the exact `slice()` tuples. 

This allows us to crop irregular geometries directly in memory, bringing the performance in line with the regular geometry slice logic ($O(1)$ memory view operation vs $O(N)$ math operation).

**Dear reviewer**
* Is this ready for review? Yes.
* All 200+ tests in `test_ndmap.py` still pass successfully.

### DIFF
--- a/gammapy/maps/wcs/ndmap.py
+++ b/gammapy/maps/wcs/ndmap.py
@@ -355,10 +355,10 @@ class WcsNDMap(WcsMap):
             else:
                 crop_x = (crop_width[0] / scales[0]).to_value("")
                 crop_y = (crop_width[1] / scales[1]).to_value("")
-            
+
             # Explicitly cast to native Python int for SonarQube
             crop_width = (int(np.round(crop_x)), int(np.round(crop_y)))
-            
+
         elif np.isscalar(crop_width):
             crop_width = (int(crop_width), int(crop_width))
         else:


kushagarwal2910-lang:
Hi @AtreyeeS, thanks for the review! 

1. I stripped out the overly detailed AI comments to keep the code clean and readable.
2.  Added support for this (scaling via _**pixel_scales**_) and pushed a unit test for it.
3. The speedup is massive for multi-resolution cubes. Cropping a 100x100 map down to 50x50 across 50 energy bins takes ~1.45s with the old .coadd() fallback, but only ~0.003s with direct slicing (~500x faster).

I also ran _**pre-commit**_ to fix the CI formatting errors. Let me know how it looks now!

bkhelifi:
Hello,
Have you read and followed the DCO instructions?
https://docs.gammapy.org/dev/development/intro.html#acceptation-of-the-developer-certificate-of-origin-dco
Are you commits signed?

kushagarwal2910-lang:
Hey @bkhelifi, thanks for bearing with me!

I'm still getting the hang of the strict open-source git workflows and totally missed the DCO sign-off when syncing with main earlier. I've just rebased and force-pushed the branch, so all the commits should be properly signed now.

While the git mechanics are a bit new to me, I'm really enjoying digging deep into Gammapy's core architecture and WCS slicing logic. Let me know if everything looks green on your end now, or if there's anything else I need to tweak!

gammapy/modeling/parameter.py
bkhelifi, line 248: This is not associated with the title of the PR, is not it?

gammapy/maps/wcs/ndmap.py
```diff
@@ -317,10 +331,27 @@ def crop(self, crop_width):
             data = self.data[tuple(slices)]
             map_out = self._init_copy(geom=geom, data=data)
         else:
-            # FIXME: This could be done more efficiently by
-            # constructing the appropriate slices for each image plane
             map_out = self._init_copy(geom=geom, data=None)
-            map_out.coadd(self)
+            non_spatial_shape = self.geom.data_shape[:-2]
+
+            for idx in np.ndindex(non_spatial_shape):
```
bkhelifi: A for loop is necessary? What is the impact on performance?

bkhelifi:
The SonarCube tests are failing. Additional tests are required

gammapy/modeling/parameter.py
kushagarwal2910-lang: You're absolutely right, my apologies. That was an accidental inclusion from some local experiments I was running on prior bounds. I've just pushed a commit to revert parameter.py entirely to keep this PR strictly focused on the WcsNDMap performance fix and the relevant tests

gammapy/maps/wcs/ndmap.py
kushagarwal2910-lang: Yes, the loop is necessary here because of how gammapy handles irregular geometries. In these cases, the spatial bounding box is not always the same size or in the same position for every energy bin. Because the crop area shifts from one plane to the next, we can't use a single, uniform slice for the whole 3D cube.

Regarding performance: the impact is actually very low. We are only looping over the bins (usually 10 to 100), not the pixels (which are in the millions). Even with the loop, this is still hundreds of times faster than the old .coadd() method because we are doing a simple memory-view slice instead of expensive math and interpolation for every pixel.

on top of that we can not use broadcasting methods as we have irregular geometries.

kushagarwal2910-lang:
I've just pushed the updated test suite. I've added a dedicated **test_wcsndmap_crop_irregular** to **test_ndmap.py** which specifically constructs a multi-resolution geometry to trigger the new slicing path. It verifies both integer and Quantity inputs, and confirms that the spatial bounds are correctly updated for each plane. This should resolve the coverage failure and clear the SonarQube gate

kushagarwal2910-lang:
"Hi @bkhelifi, I've **updated** the PR with the requested tests and fixed the formatting issues.

I also want to say thank you for the **guidance** today! Handling the CI failures and the irregular geometry logic was a great challenge. I actually learned a lot about strict **Git** **workflows** and pre-commit hooks through this process—topics I hadn't mastered before.

The new test **test_wcsndmap_crop_irregular** now passes locally and should resolve the coverage and assertion issues. Ready for another look whenever you have the time!"

kushagarwal2910-lang:
sorry!

I got my git branches tangled locally while trying to rebase and sign my commits.

I just did a hard reset and force-pushed a single, clean, DCO-signed commit.

gammapy/maps/wcs/ndmap.py
```diff
@@ -304,10 +304,24 @@ def _pad_coadd(self, geom, pad_width, mode, cval, method):
         return map_out
 
     def crop(self, crop_width):
-        if np.isscalar(crop_width):
-            crop_width = (crop_width, crop_width)
+        if isinstance(crop_width, (u.Quantity, str)):
```
AtreyeeS, line 349: You may add a docstring here

gammapy/maps/wcs/tests/test_ndmap.py
```diff
@@ -1139,3 +1139,26 @@ def test_stack_int():
...
+
+    m_crop = m.crop(crop_width=1)
+    assert_allclose(m_crop.geom.npix[0][0], 8)
+    assert_allclose(m_crop.geom.npix[0][1], 6)
+    assert m_crop.data.shape == m_crop.geom.data_shape
+    
+    assert_allclose(m_crop.data[0, 0:8, 0:8], 1.0)
+    assert_allclose(m_crop.data[1, 0:6, 0:6], 1.0)
+
+    m_crop_qty = m.crop(crop_width=1.0 * u.deg)
```
AtreyeeS, line 1156: test that this also works when `crop_width` is a tuple

gammapy/maps/wcs/ndmap.py
```diff
@@ -317,10 +331,27 @@ def crop(self, crop_width):
             data = self.data[tuple(slices)]
             map_out = self._init_copy(geom=geom, data=data)
         else:
-            # FIXME: This could be done more efficiently by
-            # constructing the appropriate slices for each image plane
             map_out = self._init_copy(geom=geom, data=None)
-            map_out.coadd(self)
+            non_spatial_shape = self.geom.data_shape[:-2]
+
```
AtreyeeS: Maybe this entire block can be moved to a private function and reused in https://github.com/kushagarwal2910-lang/gammapy/blob/b47f3b2bd462b34fa5ab20dc0fff0548fefd7083/gammapy/maps/wcs/ndmap.py#L288
instead of the coadd there?

kushagarwal2910-lang:
Hi @AtreyeeS ,
Done! I have added the towncrier fragment in docs/release-notes/6428.feature.rst as asked.

gammapy/maps/wcs/ndmap.py
```diff
@@ -304,23 +331,57 @@ def _pad_coadd(self, geom, pad_width, mode, cval, method):
...
+            cw_x, cw_y = crop_width
+            nx = int(self.geom.npix[0][0])
+            ny = int(self.geom.npix[1][0])
+
             slices = [slice(None)] * len(self.geom.axes)
             slices += [
-                slice(crop_width[1], int(self.geom.npix[1][0] - crop_width[1])),
-                slice(crop_width[0], int(self.geom.npix[0][0] - crop_width[0])),
+                slice(cw_y, ny - cw_y),
+                slice(cw_x, nx - cw_x),
```
QRemy: this change seems unnecessary

gammapy/maps/wcs/ndmap.py
```diff
@@ -317,13 +375,11 @@ def crop(self, crop_width):
...
         else:
-            # FIXME: This could be done more efficiently by
-            # constructing the appropriate slices for each image plane
             map_out = self._init_copy(geom=geom, data=None)
-            map_out.coadd(self)
+            self._copy_data_irregular(map_out, crop_width, is_crop=True)
 
         return map_out
-
+
```
QRemy:
```suggestion

```

kushagarwal2910-lang:
Thanks @QRemy

PR #6429, opened 2026-02-21
Fix consistency of `TimeMapAxis.edges` to return `Time` objects

### LINKED ISSUE #6118: Inconsisten edge unit/type for MapTimeAxis
**Gammapy version**
In `2.0.dev1438+gaf5b9f002` 

**Bug description**
Accessing the `.edges` property of a `TimeMapAxis` object returns an array of type `Quantity`, with the values being the elapsed time since the first edge of the axis. 

Accessing the edges via the `.iter_by_edges` will instead return tupes of `astropy.time.Time` objects which are date-times! 

**Expected behavior**
I expect that both ways of accessing something called the "edges" would return the same type of object.

**To Reproduce**

```
from astropy.time import Time
from gammapy.maps import TimeMapAxis

t = Time(['2000-01-01', '2025-01-01'], scale='utc')
ta = TimeMapAxis.from_time_bounds(*t, nbin=5)

for (span_start, span_end), edg in zip(ta.iter_by_edges, ta.edges):
    print(type(span_start), type(edg) , span_start == edg)
```

Basmala92:
Fixes #6118
This PR changes the return type of  `.edges` in `TimeMapAxis` to return the same type of object as `.iter_by_edges` by adding the `reference_time` .
@registerrier

### DIFF
--- a/docs/release-notes/6450.bug.rst
+++ b/docs/release-notes/6450.bug.rst
@@ -1,0 +1,1 @@
+Fix a ``RuntimeWarning`` due to division by zero in `~gammapy.irf.EffectiveAreaTable2D.plot_offset_dependence` when the maximum effective area is zero.
--- a/gammapy/estimators/points/sensitivity.py
+++ b/gammapy/estimators/points/sensitivity.py
@@ -117,7 +117,7 @@ class SensitivityEstimator(Estimator):
     def _get_criterion(self, excess, bkg):
         is_gamma_limited = excess == self.gamma_min
         is_bkg_syst_limited = excess == bkg * self.bkg_syst_fraction
-        criterion = np.char.chararray(excess.shape, itemsize=12)
+        criterion = np.empty(excess.shape, dtype="U12")
         criterion[is_gamma_limited] = "gamma"
         criterion[is_bkg_syst_limited] = "bkg"
         criterion[~np.logical_or(is_gamma_limited, is_bkg_syst_limited)] = (
--- a/gammapy/irf/effective_area.py
+++ b/gammapy/irf/effective_area.py
@@ -137,7 +137,10 @@ class EffectiveAreaTable2D(IRF):
 
         for ee in energy:
             area = self.evaluate(offset=offset_axis.center, energy_true=ee)
-            area /= np.nanmax(area)
+            max_area = np.nanmax(area)
+            if max_area == 0:
+                continue
+            area /= max_area
             if np.isnan(area).all():
                 continue
             label = f"energy = {ee:.1f}"


Basmala92:
@AtreyeeS I've deprecated `.edges` and used the alternative function as `.time_edges` and also added the test for it,
any changes on this should be made?

gammapy/maps/axes.py
```diff
@@ -2418,6 +2420,7 @@ def edges_max(self):
         return self._edges_max
 
     @property
+    @deprecated("2.1", alternative="time_edges")
```
Astro-Kirsty: Do we really want to remove edges all together?
It is more intuitive to use `.edges` as it is utilised elsewhere in the code. 

Thoughts @AtreyeeS and @registerrier?

gammapy/maps/axes.py
```diff
@@ -2459,7 +2462,10 @@ def time_mid(self):
     @property
     def time_edges(self):
         """Time edges as a `~astropy.time.Time` object."""
-        return self.reference_time + self.edges
+        if not self.is_contiguous:
+            raise ValueError("Time axis is not contiguous")
```
Astro-Kirsty: We should make the error more informative. i.e.
```suggestion
            raise ValueError("Time axis is not contiguous, therefore cannot compute bin edges.")
```

Astro-Kirsty: Same could be adapted for the other error message

gammapy/maps/axes.py
Basmala92: I agree @Astro-Kirsty !  maybe we could use `.edges` as it is and issue a warning that it is currently returning `Quantity` and will switch to returning ` Time` in the next version ?

Basmala92:
@Astro-Kirsty @AtreyeeS I have made the changes based on the last review.

Astro-Kirsty:
We need to have further discussion on this issue to decide what to do. 
We had a small discussion and it is not yet clear if we want `edges` to return a `Time` object or if we should rather include an `iter_by_time_edges`. We will discuss in the dev call and comment back here soon!

gammapy/maps/axes.py
AtreyeeS: I think its a design choice we are struggling with.
We decided to return a quantity object to be consistent with the dtype of the normal MapAxis.

I just check, and operations with time objects are ~10 slower. This might have a significant effect on fitting lightcurves, for example.
This might also break some of our plotting scripts.

Do we really gain anything from this change?

Astro-Kirsty:
The discussion from the dev call is that we should:
- leave `edges` and `times_edges` as is
- add an `iter_by_time_edges` which will return a `Time` object
- change the `iter_by_edges` return a `Quantity` object
- adjust the code to replace `iter_by_edges` with `iter_by_time_edges`

Basmala92:
Hello @Astro-Kirsty @AtreyeeS !
I have made the changes  as discussed in the call for `iter_by_edges` to return `Quantity` 
,created  ` iter_by_time_edges` to return `Time` and adjusted the code based on this.

also made a change on the error messege in `edges` based on the earlier review from @Astro-Kirsty. 
any changes?

docs/release-notes/6429.bug.rst
```diff
@@ -0,0 +1 @@
+Fixed inconsistent return type of ``TimeMapAxis.iter_by_edges`` to consistently return ``Quantity`` objects, and added a new ``iter_by_time_edges`` method that returns ``Time`` objects.
```
Astro-Kirsty: This is simply a clean up and not a bug, so I would not include a fragment, thanks!

gammapy/maps/axes.py
```diff
@@ -20,6 +20,7 @@
 
 __all__ = ["MapAxes", "MapAxis", "TimeMapAxis", "LabelMapAxis"]
 
+
```
Astro-Kirsty: Was this just added by the pre-commit hook?

gammapy/maps/axes.py
```diff
@@ -2619,9 +2622,15 @@ def is_aligned(self, other, atol=2e-2):
 
     @property
     def iter_by_edges(self):
-        """Iterate by intervals defined by the edges."""
-        for time_min, time_max in zip(self.time_min, self.time_max):
-            yield (time_min, time_max)
+        """Iterate by intervals defined by the edges , returning Quantity objects."""
```
Astro-Kirsty:
```suggestion
        """Iterate by time intervals defined by the edges as an `~astropy.units.Quantity`."""
```

gammapy/maps/axes.py
```diff
@@ -2619,9 +2622,15 @@ def is_aligned(self, other, atol=2e-2):
...
+        for t_min, t_max in zip(self._edges_min, self._edges_max):
+            yield (t_min, t_max)
+
+    @property
+    def iter_by_time_edges(self):
+        """Iterate by intervals defined by the edges , returning Time objects."""
```
Astro-Kirsty:
```suggestion
        """Iterate by time intervals defined by the edges as a `~astropy.time.Time` object."""
```

Astro-Kirsty (changes requested):
Thank you @Basmala92. Just some minor comments

gammapy/maps/axes.py
Basmala92: Sorry, this was an extra blank line added unintentionally ,now fixed!

AtreyeeS (approved):
Thanks @Basmala92 ! Looks good to me

QRemy (approved):
Thanks @Basmala92 looks good merging now

Issue #6393, opened 2026-02-13
Implement Highest Density Region (HDR) for sample results ?


facero:
This is a note from a zoom meeting fo future reference.

When using the `Sampler`, if the posterior distribution of samples is multimodal, the standard quantile approach does not return the multiple modes but only an average of it.

An alternative approach is the Highest Density Region for which a description is given here : 

<img width="668" height="413" alt="Image" src="https://github.com/user-attachments/assets/b36da890-3aca-48f7-971a-ab71b2fbe3f7" />

Ref: https://stats.stackexchange.com/questions/148439/what-is-a-highest-density-region-hdr

Maybe this could be added to the SamplerResult class in the future.

registerrier:
Note that this is fully covered by `arviz.hdi`. If you take the `Sampler` result and execute 
```
arviz.hdi(result.samples, 0.68)
``` 
you obtain the desired result.

As `arviz` is already a dependency (via ultranest?), we could simply rely on it.

facero:
Yes ultranest relies on cornerplot which relies on arviz.
This is an optional dependency.

Should we add this example somewhere in the tutorials ?
However the best show case is a multimodal distribution which we don't have
in the tutorials.

Le jeu. 19 févr. 2026, 16:32, Régis Terrier ***@***.***> a
écrit :

QRemy:
Resolved by #6433

PR #6024, opened 2025-07-29
Add utilities to create arviz.InferenceData from  SamplerResult

QRemy:
Add utilities to create arviz.InferenceData from SamplerResult
None of these utilities are meant to be exposed, other classes will use them in a separate PR introducing the BayesianModelSelection class.
Notes that arviz is an optional dependency.

PR #6433, opened 2026-02-24
Add the high density interval (HDI) to the tutorial

facero:
In response to #6393 this adds a small example in the Crab nested sampling tutorial using `arviz.hdi`.
Arviz is already an optional dependency (via ultranest).
A description of what the HDI means is also added in the tutorial ans in the glossary.

### DIFF
--- a/examples/tutorials/details/nested_sampling_Crab.py
+++ b/examples/tutorials/details/nested_sampling_Crab.py
@@ -468,7 +468,7 @@ result_2 = sampler.run(datasets[2])
 
 ######################################################################
 # Highest density intervals
-# ----------------------
+# -------------------------
 #
 # Given the samples, one can also compute the highest density interval (HDI)
 # which is also known as the smallest credible interval (SCI).


docs/user-guide/references.rst
```diff
@@ -79,6 +79,10 @@ Glossary
     FoV Background
       Background estimation method typically used for the 3D analysis. See :ref:`makers`.
 
+    HDI
+      Short for "highest density interval": it is a region of the posterior distribution that contains a specified probability mass.
+      It is used to compute credible intervals based on samples.
+
```
bkhelifi, line 85: You could make here a link towards the tutorial, no?

facero:
For some reason the docs do not compile as it complains that the arviz package cannot be imported.
This is strange as I can see that ultranest is installed, the Sampler can be run and fails at the end when importing arviz.

```
          File "/home/runner/work/gammapy/gammapy/examples/tutorials/details/nested_sampling_Crab.py", line 486, in <module>
            from arviz import hdi
        ModuleNotFoundError: No module named 'arviz'
```

I have checked in a new env on my machine and I can indeed confirm that arviz is installed along ultranest.
Any sugggestions on how to see which packages are installed in the docs compilation ?
Capture d’écran 2026-02-24 à 12 07 59

docs/user-guide/references.rst
facero: I've added this althoug I'm not sure how to make a local reference to another tutorial (I want to avoid putting the absolute path). ..

facero:
Apparently the docs cannot find `arviz` as it is a dependency of `ultranest` and that the env for the docs is created via ( found in the logs) : 
```
python -I -m pip install --force-reinstall --no-deps /home/runner/work/gammapy/gammapy/.tox/.tmp/package/1/gammapy-2.0.dev2918+g60c1f41ce.tar.gz
```
The `--no-deps` here will not install `arviz`.

The solution that works is to add `arviz` to the tom file in the `docs` section.

docs/user-guide/references.rst
facero: what I tried did not work. Do you know how to make internal links to another page ?

docs/user-guide/references.rst
```diff
@@ -79,6 +79,10 @@ Glossary
     FoV Background
       Background estimation method typically used for the 3D analysis. See :ref:`makers`.
 
+    HDI
+      Short for "highest density interval": it is a region of the posterior distribution that contains a specified probability mass.
+      It is used to compute credible intervals based on samples as shown in the :ref:`tutorial on Nested Sampling <../tutorials/details/nested_sampling_Crab.html>`.
```
Astro-Kirsty: The correct format can be found [here](https://docs.gammapy.org/dev/development/doc_howto.html#links-to-other-tutorials)
```suggestion
      It is used to compute credible intervals based on samples as shown in the :doc:`tutorial on Nested  </tutorials/details/nested_sampling_Crab>`.
```

examples/tutorials/details/nested_sampling_Crab.py
```diff
@@ -465,3 +465,36 @@ def get_samples_from_posterior(spectral_model, results):
...
 # amplitude (possibly due to calibrations or/and atmospheric issues).
 #
+
+######################################################################
+# High density intervals
+# ----------------------
+#
+# Given the samples, one can also compute the high density interval (HDI)
+# which is also known as the smallest credible interval (SCI).
+# See more details here : https://en.wikipedia.org/wiki/Credible_interval
```
Astro-Kirsty:
```suggestion
# See more details `here <https://en.wikipedia.org/wiki/Credible_interval>`__.
```

examples/tutorials/details/nested_sampling_Crab.py
```diff
@@ -465,3 +465,36 @@ def get_samples_from_posterior(spectral_model, results):
...
+# This is the interval in which a given probability (e.g. 68%) is contained.
+#
+# For unimodal distributions, the HDI is a single continuous interval containing
+# the mode whereas for multimodal distributions, the HDI can be a set of disconnected intervals.
+# The HDI can be particularly helpful with multimodal distributions as opposed
+# to the mean and quantiles approaches which will not report the important information.
+# Here we showcase the HDI using the Arviz package.
```
Astro-Kirsty:
```suggestion
# Here, we showcase the HDI using the Arviz package.
```

examples/tutorials/details/nested_sampling_Crab.py
```diff
@@ -465,3 +465,36 @@ def get_samples_from_posterior(spectral_model, results):
 # One can note as well that one of the run has a notably different
 # amplitude (possibly due to calibrations or/and atmospheric issues).
 #
+
+######################################################################
+# High density intervals
```
registerrier:
```suggestion
# Highest density intervals
```

examples/tutorials/details/nested_sampling_Crab.py
```diff
@@ -465,3 +465,36 @@ def get_samples_from_posterior(spectral_model, results):
...
+# ----------------------
+#
+# Given the samples, one can also compute the high density interval (HDI)
```
registerrier:
```suggestion
# Given the samples, one can also compute the highest density interval (HDI)
```

examples/tutorials/details/nested_sampling_Crab.py
```diff
@@ -465,3 +465,36 @@ def get_samples_from_posterior(spectral_model, results):
...
+# which is also known as the smallest credible interval (SCI).
+# See more details `here <https://en.wikipedia.org/wiki/Credible_interval>`__.
+# This is the interval in which a given probability (e.g. 68%) is contained.
```
registerrier:
```suggestion
# This is the smallest interval in which a given probability (e.g. 68%) is contained.
```

examples/tutorials/details/nested_sampling_Crab.py
```diff
@@ -465,3 +465,36 @@ def get_samples_from_posterior(spectral_model, results):
...
+
+from arviz import hdi
+
+amplitude_hdi = []
+for s in [result_0.samples, result_1.samples, result_2.samples]:
+    amplitude_hdi.append(hdi(s[:, 1], hdi_prob=0.95, multimodal=False))
+
+fix, ax = plt.subplots(1, 1, figsize=(7, 5))
+for i, amp in enumerate(amplitude_hdi):
+    ax.errorbar(amp.mean(), i, xerr=(amp[1] - amp[0]) / 2, lw=15, label=f"run {i}")
```
registerrier: Is this completely meaning full to show the HDI as as error bar? The center of the HDI might not be specifically meaningful.

You might use plt.hlines to have a bar extending over the full interval. 

If one wants to add a point value, I would assume the mode to be the better information to represent.

pyproject.toml
```diff
@@ -101,6 +101,7 @@ docs = [
     "pydata-sphinx-theme",
     "nbformat",
     "docutils",
+    "arviz",
```
registerrier: This is a bit strange to have it here. Why not in `'all'`?

facero: good question
I tested several things and possible options were to have it in `all`, `docs`, `all`+`docs`.
Arviz is already a dependency so in theory not need to add it at all except that docs are built using the `--no-deps` so it needs to be specified by hand.

Does CI carries out the test with the `--no-deps` option ?
Adding it to `all` should cover most cases so it might be safer.

examples/tutorials/details/nested_sampling_Crab.py
facero: using error bar here was just the first thing that came to my mind for plotting.
`plt.hline` is indeed simpler. I'll fix that.

pyproject.toml
morcuended: We should not rely on the fact that something is a dependency of another gammapy dependency. If used, it should be explicitly included. +1 for including it in `all`

facero: done

facero:
All done
Should be ready to merge once  #6024  is merged (to have arviz is in all)

QRemy:
Thanks @facero I made some changes to show a multi-model HDI in comparison to other intervals

image

examples/tutorials/details/nested_sampling_Crab.py
```diff
@@ -465,3 +465,99 @@ def get_samples_from_posterior(spectral_model, results):
 # One can note as well that one of the run has a notably different
 # amplitude (possibly due to calibrations or/and atmospheric issues).
 #
+
+######################################################################
+# Highest density intervals
+# ----------------------
```
Astro-Kirsty:
```suggestion
# Highest density intervals
# -------------------------
```

PR #6435, opened 2026-02-24
Fix typos.

yaochengchen:
**Description**

Fix some typos.

### DIFF
--- a/docs/development/dev_howto.rst
+++ b/docs/development/dev_howto.rst
@@ -581,7 +581,7 @@ We utilise `towncrier <https://towncrier.readthedocs.io/en/stable/>`__ for our r
   does not need a changelog fragment entry. If you are unsure if you need a fragment, the lead developers can
   help you decide.
 - The naming convention of the file should be ``<PULL REQUEST NUMBER>.<TYPE>.rst``, where the available
-  types are ``infrastructure``, ``docs``, ``feature`` or ``bugfix``.
+  types are ``infrastructure``, ``docs``, ``feature`` or ``bug``.
 - The file should contain a suitable message for the PR, for example "A new function `select_nested_models`
   has been introduced to perform nested model fits and compute the resulting test statistic (TS) between
   two nested hypotheses.".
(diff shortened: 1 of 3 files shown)


gammapy/modeling/models/spatial.py
```diff
@@ -141,7 +141,11 @@ def position_error(self):
         sub_covar[0, 0] *= cos_lat**2.0
         sub_covar[0, 1] *= cos_lat
         sub_covar[1, 0] *= cos_lat
+
+        print(sub_covar.dtype)
```
bkhelifi: Why these prints? And this is not a typo fix... otherwise, the PR is good

yaochengchen: Hi @bkhelifi,

The `print` statements were added for debugging because the CI job **“CI / ubuntu-latest, devdeps (pull_request)”** was consistently failing, while I could not reproduce the issue locally.

Locally (NumPy 2.4.2), everything works fine.
However, in the CI environment (NumPy 2.5.0.dev0), I observed that:

* `sub_covar.dtype` is `float64`
* but `np.linalg.eig(sub_covar)` returns `eig_vecs.dtype == complex128`
  (even though the values are `1.+0.j` and `0.+0.j`)
* therefore `y_vec = eig_vecs[:, 0]` becomes complex
* and `np.arctan2(complex, complex)` raises the `TypeError` seen in CI

So the failure is triggered by a change in behavior in the newer NumPy dev version.

Since `sub_covar` is a real symmetric covariance matrix, it would be more appropriate (and numerically safer) to use:

```python
eig_vals, eig_vecs = np.linalg.eigh(sub_covar)
```

instead of `np.linalg.eig`, which should also avoid the complex dtype issue entirely.

I will remove the debug prints and open a separate PR specifically addressing this fix, as it is indeed more than a typo change.

Let me know if you agree with this approach.

Astro-Kirsty: It would be good if you can rather open an issue on what the problem is here, so we can discuss it there. 
We should not include print statements in our code base :)

Astro-Kirsty: Otherwise, the small typo fixes are okay for me

yaochengchen: Done as requested.

PR #6436, opened 2026-02-24
Ci fail numpy complex

yaochengchen:
**Description**

This PR replaces `np.linalg.eig` with `np.linalg.eigh` when diagonalizing the position covariance sub-matrix.

The CI job (`ubuntu-latest, devdeps`) was failing with NumPy 2.5.0.dev0 because `np.linalg.eig` returned eigenvectors with `complex128` dtype (even though the matrix is real symmetric and the imaginary parts were zero). This caused `np.arctan2` to raise a `TypeError`, as it does not support complex inputs.

Since the covariance sub-matrix is real and symmetric, `np.linalg.eigh` is the numerically appropriate choice and guarantees real-valued eigenvalues and eigenvectors. This follows standard numerical linear algebra best practices for symmetric matrices.

This change restores CI stability while also making the implementation more robust and mathematically consistent.

### DIFF
--- a/docs/development/dev_howto.rst
+++ b/docs/development/dev_howto.rst
@@ -581,7 +581,7 @@ We utilise `towncrier <https://towncrier.readthedocs.io/en/stable/>`__ for our r
   does not need a changelog fragment entry. If you are unsure if you need a fragment, the lead developers can
   help you decide.
 - The naming convention of the file should be ``<PULL REQUEST NUMBER>.<TYPE>.rst``, where the available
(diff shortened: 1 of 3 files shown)


yaochengchen:
> Thanks @yaochengchen . This is well spotted.
> 
> I think that technically this is not a bugfix as the error has not appeared yet, but a cleanup. A towncrier fragment is therefore not expected. Could you remove it?

@registerrier Thanks for the review, updated as requested.

Issue #6401, opened 2026-02-17
SonarQube reliability: attribute access on a value that can be 'None'.


registerrier:
This is a list of SonarQube reliability issues found by CTAO SonarQube.

This specific issue connects to the error `Fix this attribute access on a value that can be 'None'`:

Occurrences:

gammapy/modeling/models/core.py:36 — [View in SonarQube](https://sonar-ctao.zeuthen.desy.de/project/issues?id=gammapy_gammapy_8c1bf7fc-018a-4afd-b216-015191d282fa&issues=18053c11-f96b-458f-ad55-4043d05a78f5&open=18053c11-f96b-458f-ad55-4043d05a78f5)
gammapy/workflow/steps.py:125 — [View in SonarQube](https://sonar-ctao.zeuthen.desy.de/project/issues?id=gammapy_gammapy_8c1bf7fc-018a-4afd-b216-015191d282fa&issues=74754509-151b-41ed-95d6-e672598db1b2&open=74754509-151b-41ed-95d6-e672598db1b2)
gammapy/workflow/steps.py:465 — [View in SonarQube](https://sonar-ctao.zeuthen.desy.de/project/issues?id=gammapy_gammapy_8c1bf7fc-018a-4afd-b216-015191d282fa&issues=85a8bde8-e9be-4690-9f10-66eba56c191a&open=85a8bde8-e9be-4690-9f10-66eba56c191a)
gammapy/workflow/steps.py:490 — [View in SonarQube](https://sonar-ctao.zeuthen.desy.de/project/issues?id=gammapy_gammapy_8c1bf7fc-018a-4afd-b216-015191d282fa&issues=5b4bc074-662c-4661-aca0-01b17007ea0c&open=5b4bc074-662c-4661-aca0-01b17007ea0c)
gammapy/modeling/models/core.py:176 — [View in SonarQube](https://sonar-ctao.zeuthen.desy.de/project/issues?id=gammapy_gammapy_8c1bf7fc-018a-4afd-b216-015191d282fa&issues=07a88dcb-5cc2-4e8c-a140-2bfeb63bb11a&open=07a88dcb-5cc2-4e8c-a140-2bfeb63bb11a)
[...]

PR #6441, opened 2026-02-25
Handle None value for SonarQube - Part 1

MSCarrasco:
This pull request is associated to the SonarCube reliability issue #6401 for attribute access on a value that can be 'None'.

The first 10 subissues are fixed.

Two of them came from missing parameters in configuration to run the light curve and the flux points.
I added a value error.

The others came from the case when make_path returns a None value.
If it happens in a write or read function, I added an error value when the provided file path is None, since it is needed to perform the writing or reading.
If a None value is acceptable, then it is handled with the adequate boolean.


bkhelifi:
Thanks a lot @MSCarrasco . The SonarCube analysis is still failing, because the new checks have no associated test...

MSCarrasco:
> Thanks a lot @MSCarrasco . The SonarCube analysis is still failing, because the new checks have no associated test...

Thanks @bkhelifi, yes i ended up understanding this, while also realizing some potentially missed issues. I turned the PR into draft to take the time to check all of this.

gammapy/datasets/core.py
```diff
@@ -493,6 +493,8 @@ def write(
             When True adds both DATASUM and CHECKSUM cards to the headers written to the FITS files.
             Default is True.
         """
+        if filename is None:
+            raise ValueError("Filename cannot be None")
```
Astro-Kirsty: Maybe we could change the error message to:
`The filename is not defined.`

As this is used in some other places in gammapy

gammapy/modeling/models/core.py
```diff
@@ -176,6 +178,8 @@ def _write_models(
 ):
     """Write models to YAML file with additional information using an `extra_dict`."""
 
+    if path is None:
+        raise ValueError("Path cannot be None")
```
Astro-Kirsty: Again something like:
`The path is not defined.`

Astro-Kirsty (changes requested):
Thanks @MSCarrasco. 
Just a few comments. You should also add tests to catch all the new error messages


(i realise you changed to a draft now but I already made these before!)

gammapy/workflow/steps.py
```diff
@@ -235,6 +235,8 @@ def _get_data_store(self):
         elif path.is_dir():
             self.log.debug(f"Setting datastore from directory: {path}")
             return DataStore.from_dir(path)
+        elif path is None:
+            raise ValueError("No datastore file in configuration")
```
MSCarrasco: I am unsure this should be implemented, since ObservationsConfig requires a Path or string value
When I try to assign None to config.observations.datastore, I get a validation error
So I don't see how to write a test for this case.

gammapy/workflow/steps.py
```diff
@@ -580,7 +582,14 @@ def _run(self):
...
 
-        energy_edges = make_energy_axis(fp_settings.energy).edges
+        energy_edges = make_energy_axis(fp_settings.energy)
+        if energy_edges is not None:
+            energy_edges = energy_edges.edges
+        else:
+            raise ValueError(
+                "Missing or incomplete energy axis parameters in flux points configuration."
+            )
+
```
MSCarrasco, line 592: Same as previous comment, I am unsure this should be implemented
Here EnergyAxisConfig requires Quantities
When I try to assign a None value, I get a validation error.
When I try to assign it in the configuration dict, I get TypeError: Cannot parse "None" as a Quantity. It does not start with a number.
I was unable to make a test function for this change.

gammapy/workflow/steps.py
```diff
@@ -608,7 +617,13 @@ def _run(self):
...
         lc_settings = self.config.light_curve
         self.log.info("Computing light curve.")
-        energy_edges = make_energy_axis(lc_settings.energy_edges).edges
+        energy_edges = make_energy_axis(lc_settings.energy_edges)
+        if energy_edges is not None:
+            energy_edges = energy_edges.edges
+        else:
+            raise ValueError(
+                "Missing or incomplete energy axis parameters in light curve configuration."
+            )
```
MSCarrasco, line 627: Same as previous comment, I am unsure this should be implemented
Here EnergyAxisConfig requires Quantities
When I try to assign a None value, I get a validation error.
When I try to assign it in the configuration dict, I get TypeError: Cannot parse "None" as a Quantity. It does not start with a number.
I was unable to make a test function for this change.

gammapy/datasets/core.py
MSCarrasco: Thanks, I changed all the occurences

gammapy/modeling/models/core.py
MSCarrasco: Thanks, I changed it

MSCarrasco:
I implemented test functions for all the changes except for the issues related to configuration classes.
It seems impossible for the user to set the considered value to None, but I may just lack knowledge of the code, so any advice is welcome.
The tests are failing for some reason unrelated to the files I changed.

gammapy/datasets/tests/test_datasets.py
```diff
@@ -232,6 +232,9 @@ def test_datasets_write(map_datasets, tmp_path):
             overwrite=False,
         )
 
+    with pytest.raises(ValueError):
```
Astro-Kirsty: I would also add the message for each of these i.e. 

`with pytest.raises(ValueError, match="message"):`

Astro-Kirsty (changes requested):
Thanks @MSCarrasco.
I left one minor comment for adaption and read through your other comments.

I agree that it appears in all cases there is not a situation where the `filename` can be wrong, therefore I suggest we add these to ignore for SonarQube. This can be done with `# NOSONAR` on the line you want SQ to ignore and then followed by the error key, as this is what we have been doing. As an example:

```
path = make_path(self.config.observations.datastore) # NOSONAR
# (S2259): attribute cannot be None
```

gammapy/datasets/tests/test_datasets.py
MSCarrasco: thank you, I added the matching messages

MSCarrasco:
> Thanks @MSCarrasco. I left one minor comment for adaption and read through your other comments.
> 
> I agree that it appears in all cases there is not a situation where the `filename` can be wrong, therefore I suggest we add these to ignore for SonarQube. This can be done with `# NOSONAR` on the line you want SQ to ignore and then followed by the error key, as this is what we have been doing. As an example:
> 
> ```
> path = make_path(self.config.observations.datastore) # NOSONAR
> # (S2259): attribute cannot be None
> ```

Thanks for the help, I did the changes

MSCarrasco:
> ## [Quality Gate failed](https://sonar-ctao.zeuthen.desy.de/dashboard?id=gammapy_gammapy_8c1bf7fc-018a-4afd-b216-015191d282fa&pullRequest=6441) **Quality Gate failed**
> 
> Failed conditions [](https://sonar-ctao.zeuthen.desy.de/project/issues?id=gammapy_gammapy_8c1bf7fc-018a-4afd-b216-015191d282fa&pullRequest=6441&resolved=false) [5 New issues](https://sonar-ctao.zeuthen.desy.de/project/issues?id=gammapy_gammapy_8c1bf7fc-018a-4afd-b216-015191d282fa&pullRequest=6441&resolved=false)
> 
> [See analysis details on SonarQube](https://sonar-ctao.zeuthen.desy.de/dashboard?id=gammapy_gammapy_8c1bf7fc-018a-4afd-b216-015191d282fa&pullRequest=6441)
> 
>  Catch issues before they fail your Quality Gate with our IDE extension [SonarQube for IDE](https://www.sonarsource.com/products/sonarlint/features/connected-mode/?referrer=pull-request) [SonarQube for IDE](https://www.sonarsource.com/products/sonarlint/features/connected-mode/?referrer=pull-request)

The new issues are not related to this PR

Astro-Kirsty (dismissed):
Thanks @MSCarrasco. 

We discussed in the dev call and decided it was ready!

MSCarrasco:
Hi @Astro-Kirsty, @morcuended made me realize in the PR part 2 that there was in fact a way for the energy axis to be declared with None values. I put back the ValueError and was able to produce a test.
The SQ errors are again unrelated to the PR

Astro-Kirsty:
Okay great to hear @MSCarrasco. Can you confirm the affect lines of code to be adjusted here?

Astro-Kirsty (approved):
Thanks @MSCarrasco. Great that you were able to find a way to test for the missing values.

gammapy/workflow/steps.py
MSCarrasco: It was in fact possible to set it with None values using an empty dict, or using the default config objects.
I put it back, with the corresponding testing

gammapy/workflow/steps.py
MSCarrasco: It was in fact possible to set it with None values using an empty dict, or using the default config objects.
I put it back, with the corresponding testing

bkhelifi (approved):
Thanks a lot @MSCarrasco

QRemy (approved):
Thanks @MSCarrasco looks good, merging now.

PR #6444, opened 2026-02-26
Render log as math operator in docstrings

aaguasca:
**Description**

This pull request fixes the render of log as a math operator in docstrings.

**Dear reviewer**

The PR is ready for review.

### DIFF
--- a/examples/models/spectral/plot_logparabola.py
+++ b/examples/models/spectral/plot_logparabola.py
@@ -13,13 +13,13 @@ It is defined by the following equation:
           - \alpha - \beta \log{ \left( \frac{E}{E_0} \right) }
         }
 
-Note that :math:`log` refers to the natural logarithm. This is consistent
+Note that :math:`\log` refers to the natural logarithm. This is consistent
 with the `Fermi Science Tools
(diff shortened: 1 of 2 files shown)


Astro-Kirsty (changes requested):
Thank you @aaguasca.

This is something we would typically backport however, `plot_logparabola2.py` cannot be backported. Therefore, I suggest we split this into to PRs

bkhelifi (dismissed):
Thanks!

AtreyeeS (dismissed):
Thanks @aaguasca !

I agree with @Astro-Kirsty about the backport to 2.0

aaguasca:
Okay! I will create a separate PR for `plot_logparabola2.py`.

Could you please clarify how I can know if a file can be backported or not? Thank you!

Astro-Kirsty:
A backport should only be performed if the change was introduced before the previous release. In this case, we have not introduced `plot_logparabola2.py` yet (i.e. it will be in the future 2.1 release), therefore it did not exist at 2.0, so we cannot backport it
It is difficult to know unless you are contributing all the time, which is why we are here to help :)

aaguasca:
Thank you very much for the explanation! Now it is clear to me:)

QRemy (approved):
Thanks @aaguasca, merging now.

Issue #5556, opened 2024-11-08
Creation of Docker images for each release?


bkhelifi:
CTAO will use the SAT in automatic pipelines and there is a deployment requirement:

**Requirement B-SUSS-1530**
The SUSS science tools must be readily deployable on all widely used unix-based platforms within 1 hour.

A docker image of the SAT versions is then highly envisioned.

Even if this requirement is for the SAT, one should discuss whether one should make Docker image for each Gammapy releases. This will permit to test the following features
- add a docker image builder in the release process
- render this creation automatic through an action
- add a test to run a simple script into such Docker image

HealthyPear:
Not an expert of these requirements, but _gammapy_ is already a Python package and thus already comes with a cross-platform installation procedure (_conda_ or _venv_) - why the need for Docker?

bkhelifi:
The idea is quite simple...
For each job or process, calling conda to make the full installation at each time takes a lot of time. Whereas, having a docker already running permits to avoid this and to immediately run scripts.

HealthyPear:
then if you assume that gammapy does not need to be reinstalled each time you can just create the environment once for the specific version and let any job access that, this should be equivalent no?

of course I do not know the details of the platform you want to use I might be missing details

bkhelifi:
The limitation is the following: all machines will not have Gammapy env installed per default, ie the EGI computing elements. Virtual Machines will be deployed on nodes and monitored by kebernetes.
It allows also to not depend on the conda forge, that we know that it can be slow, but on the local CTAO forge of Docker.

registerrier:
I think we already comply with the requirement as written. I would remove the SAT req. label.

We can discuss Docker images. I remind that there has been an attempt to create a Dockerfile for the dev version here: #3339

bkhelifi:
Let's converge on the Docker aspect before closing it. This issue is a reminder

morcuended:
Maybe as starting point, we can try the Dockerfile suggested here https://gitlab.cta-observatory.org/cta-computing/documentation/python-project-template

guillaumegrolleron:
Dear all, 
Is this feature still ongoing? 
I asked the question because I'm currently facing several difficulties performing analysis with old gammapy versions. For example, it is challenging to recreate a clean install of version 0.19, because dependencies in the environment.yml are not pinned. Docker would be great in this case.
I already performed such a deployment in CI., I could have a look if you want

morcuended:
[...]

PR #6446, opened 2026-02-26
Build docker image

morcuended:
**Description**

This pull request adds a Dockerfile and a monthly scheduled CI workflow that build the image.

Later it can be extended e.g. to publish it to DockerHub or any other registry.

See #5556 #3339 `

**Dear reviewer**

### DIFF
--- a/.github/workflows/docker_build.yml
+++ b/.github/workflows/docker_build.yml
@@ -1,0 +1,16 @@
+on:
+  workflow_dispatch:
+
+jobs:
+  docker:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+
+      - name: Build image
+        run: docker build -t gammapy:test .
+
+      - name: Test image
+        run: docker run --rm gammapy:test python -c "import gammapy; print(gammapy.__version__)"
--- a/.github/workflows/docker_build.yml
+++ b/.github/workflows/docker_build.yml
@@ -1,8 +1,6 @@
-on:
-  workflow_dispatch:
-  push:
-    branches:
-      - main
+name: Docker image build
+
+on: [push]
 
 jobs:
   docker:
--- a/.github/workflows/docker_build.yml
+++ b/.github/workflows/docker_build.yml
@@ -1,4 +1,4 @@
-name: Docker image build
+name: Build docker image
 
 on: [push]
 
--- a/.github/workflows/docker_build.yml
+++ b/.github/workflows/docker_build.yml
@@ -1,5 +1,8 @@
 on:
   workflow_dispatch:
+  push:
+    branches:
+      - main
 
 jobs:
   docker:
--- a/.github/workflows/docker_build.yml
+++ b/.github/workflows/docker_build.yml
@@ -1,6 +1,8 @@
 name: Build docker image
 
-on: [push]
+on:
+  schedule:
+    - cron: '0 0 1 * *'  # Run monthly, at 12:00 AM on day 1 of the month
 
 jobs:
   docker:
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@ FROM python:3.12 AS builder
 # Add necessary sources, including .git for version info
 COPY gammapy/ /repo/gammapy/
 COPY .git/ /repo/.git/
-COPY pyproject.toml setup.py codemeta.json CITATION.cff MANIFEST.in README.rst LICENSE.rst CHANGES.rst /repo/
+COPY pyproject.toml setup.py codemeta.json CITATION.cff MANIFEST.in README.rst LICENSE.rst /repo/
 
 WORKDIR /repo
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,23 @@
+# Two-stage build, first build the python wheel
+FROM python:3.12 AS builder
+
+# Add necessary sources, including .git for version info
+COPY gammapy/ /repo/gammapy/
+COPY .git/ /repo/.git/
+COPY pyproject.toml setup.py codemeta.json CITATION.cff MANIFEST.in README.md LICENSE.rst CHANGES.rst /repo/
+
+WORKDIR /repo
+
+# Build the Gammapy wheel
+RUN python -m pip install --no-cache-dir build \
+    && python -m build --wheel
+
+# Second stage, copy and install wheel in the slim python image to reduce image size.
+FROM python:3.12-slim
+COPY --from=builder /repo/dist /tmp/dist
+
+RUN python -m pip install --no-cache-dir /tmp/dist/* \
+    && rm -r /tmp/dist
+
+RUN useradd --create-home --system --user-group gammapy
+USER gammapy
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@ FROM python:3.12 AS builder
 # Add necessary sources, including .git for version info
 COPY gammapy/ /repo/gammapy/
 COPY .git/ /repo/.git/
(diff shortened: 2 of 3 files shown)


morcuended:
Build the image and dummy test work, see: https://github.com/morcuended/gammapy/actions/runs/22447431810/job/65005795435?pr=11

Dockerfile
```diff
@@ -0,0 +1,23 @@
+# Two-stage build, first build the python wheel
+FROM python:3.12 AS builder
```
bkhelifi, line 4: The pyhton version is hardcoded here. Can we more sustainable?

bkhelifi:
An associated question:
Pypi is our forge for our code. Should we use DockerHub for our images?
There is a https://hub.docker.com/r/gammapy/gammapy: I do not know who has the permission rights... Should we use it? If yes, do we need to improve it (with metadata, descriptions, associated EU projects) like for our Zenodo domain>

morcuended:
> An associated question: Pypi is our forge for our code. Should we use DockerHub for our images? There is a https://hub.docker.com/r/gammapy/gammapy: I do not know who has the permission rights... Should we use it? If yes, do we need to improve it (with metadata, descriptions, associated EU projects) like for our Zenodo domain>

Yes, I think that could be the place. For pushing the image you can use https://github.com/marketplace/actions/build-and-push-to-dockerhub I guess you need to set up some tokens for it. For the metadata I do not know

Dockerfile
natthan-pigoux: In general, its a good practice to specify all versions in the Dockerfile (base image, dependencies,...)

Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+
+# Second stage, copy and install wheel in the slim python image to reduce image size.
+FROM python:3.12-slim
+COPY --from=builder /repo/dist /tmp/dist
+
+RUN python -m pip install --no-cache-dir /tmp/dist/* \
+    && rm -r /tmp/dist
+
+RUN useradd --create-home --system --user-group gammapy
+USER gammapy
```
natthan-pigoux, line 28: I think you should run as `gammapy` user before running all the other commands, e.g. copying the repo and installing gammapy, it should avoid at least later permissions issues.

morcuended: I haven't tried, but I do not know if I can copy and run pip install without being non-root (unless using `pip install --user ...`). Also, I think the first stage can be done with root user, then switch to non-root in the second one to run whatever that does not need privileged user.

natthan-pigoux: You are right here, since you are building in two steps it won't work indeed. Fine for me.

natthan-pigoux (approved):

bkhelifi:
Thanks a lot, @morcuended !!!

I think that one should add:
- at the beginning of the files, the line about the Gammapy licence
- improve the user documentation about how to use a docker image
- improve the user documentation about how to use a docker image
- update our official dockerhub repo (https://hub.docker.com/r/gammapy/gammapy) with a better description of Gammapy (e.g. "Gammapy is an open-source Python package for gamma-ray astronomy built on [Numpy](https://numpy.org/), [Scipy](https://scipy.org/) and [Astropy](https://www.astropy.org/).
It is used as core library for the Science Analysis tools of the [Cherenkov Telescope Array Observatory (CTAO)](https://www.ctao.org/), recommended by the [H.E.S.S.](https://hess-experiment.eu/) collaboration to be used for Science publications, and is already widely used in the analysis of existing gamma-ray instruments, such as [MAGIC](https://magic.mpp.mpg.de/), [VERITAS](https://veritas.sao.arizona.edu/) and [HAWC](https://www.hawc-observatory.org/).") and tagged versions of our latest releases

.github/workflows/docker_build.yml
```diff
@@ -0,0 +1,17 @@
+name: Build docker image
+
+on: [push]
```
registerrier: Should it be on any push? 
It would make more sense to have it as a cron job no?

Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+
+# Build the Gammapy wheel
+RUN python -m pip install --no-cache-dir build \
+    && python -m build --wheel
+
+# Second stage, copy and install wheel in the slim python image to reduce image size.
+FROM python:3.12-slim
+COPY --from=builder /repo/dist /tmp/dist
+
+RUN python -m pip install --no-cache-dir /tmp/dist/* \
```
registerrier: Should it use only the core dependencies or a larger set of dependencies?

registerrier (commented):
Thanks @morcuended . This is nice. I have left two inline questions.

.github/workflows/docker_build.yml
morcuended: Could be scheduled job indeed

Dockerfile
```diff
@@ -0,0 +1,23 @@
+# Two-stage build, first build the python wheel
```
morcuended, line 3:
```suggestion
# Licensed under a 3-clause BSD style license - see LICENSE.rst

# Two-stage build, first build the python wheel
```

.github/workflows/docker_build.yml
morcuended:
```suggestion
on:
  schedule:
    - cron: '0 0 1 * *'  # Run monthly, at 12:00 AM on day 1 of the month
```

Dockerfile
morcuended: I can add `all` optional dependencies

Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+# Build the Gammapy wheel
+RUN python -m pip install --no-cache-dir build \
+    && python -m build --wheel
+
+# Second stage, copy and install wheel in the slim python image to reduce image size.
+FROM python:3.12-slim
+COPY --from=builder /repo/dist /tmp/dist
+
+RUN python -m pip install --no-cache-dir /tmp/dist/* \
+    && rm -r /tmp/dist
```
morcuended:
```suggestion

RUN python -m pip install --no-cache-dir /tmp/dist/*[all] \
    && rm -r /tmp/dist
```

morcuended (commented):
Thanks for the comments

morcuended:
> at the beginning of the files, the line about the Gammapy licence

Done

> improve the user documentation about how to use a docker image

I can do it here, but this needs that gammapy image is available in dockerhub. So this might need to be done in a second PR.

Along these lines to be included in https://docs.gammapy.org/dev/getting-started/install.html#installation

### Using Docker Image
You can run Gammapy in a fully pre-configured environment using Docker.

#### Pull the image:
```
docker pull gammapy:v2.1
```
#### Run an analysis script
From your working directory:
```
docker run --rm \
  -v $(pwd):/work \
  -w /work \
  gammapy:latest \
  python my-analysis-script.py
```
This mounts your current directory into the container and executes your script with all Gammapy dependencies available.

> update our official dockerhub repo (https://hub.docker.com/r/gammapy/gammapy) with a better description of Gammapy (e.g. "Gammapy is an open-source Python package for gamma-ray astronomy built on [Numpy](https://numpy.org/), [Scipy](https://scipy.org/) and [Astropy](https://www.astropy.org/).
> It is used as core library for the Science Analysis tools of the [Cherenkov Telescope Array Observatory (CTAO)](https://www.ctao.org/), recommended by the [H.E.S.S.](https://hess-experiment.eu/) collaboration to be used for Science publications, and is already widely used in the analysis of existing gamma-ray instruments, such as [MAGIC](https://magic.mpp.mpg.de/), [VERITAS](https://veritas.sao.arizona.edu/) and [HAWC](https://www.hawc-observatory.org/).") and tagged versions of our latest releases

This has to be done by someone with access to that account

Astro-Kirsty:
Hello @morcuended, what is the status here?

morcuended:
There is a suggested action item that must be done by Gammapy admins (interaction with DockerHub). That can be made in a subsequent PR.
Updating the documentation on how to use it as well can be done in a different PR.

PR #6456, opened 2026-02-27
FoV frame initialization.

registerrier:
**Description**

This pull request proposes a change to enforce that for the `FovAltAzFrame` `origin` is consistent with `obstime` and `location`.  It is currently possible to assigne a different obstime than that of the `AltAz` defining the origin. This can cause weird side effects and incorrect results.

**Dear reviewer**

### DIFF
--- a/gammapy/utils/coordinates/fov.py
+++ b/gammapy/utils/coordinates/fov.py
@@ -73,7 +73,51 @@ class FoVAltAzFrame(BaseCoordinateFrame):
     obstime = TimeAttribute(default=None)
     location = EarthLocationAttribute(default=None)
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, *args, origin=None, obstime=None, location=None, **kwargs):
+        if not isinstance(origin, (AltAz, SkyCoord)):
+            raise TypeError(
+                f"origin must be AltAz SkyCoord , got {type(origin).__name__} instead."
+            )
+        if isinstance(origin, SkyCoord) and not isinstance(origin.frame, AltAz):
+            raise TypeError("origin must be AltAz.")
+
+        if obstime is None and origin.obstime is not None:
+            obstime = origin.obstime
+        if location is None and origin.location is not None:
+            location = origin.location
+
+        kwargs["origin"] = origin
+        kwargs["location"] = location
+        kwargs["obstime"] = obstime
+
+        # Validate obstime consistency
+        if obstime is not None and origin.obstime is not None:
+            if obstime.shape != origin.obstime.shape:
+                raise ValueError(
+                    f"obstime shape mismatch: frame obstime has shape "
+                    f"{obstime.shape} but origin.obstime has shape "
+                    f"{origin.obstime.shape}."
+                )
+            if not np.all(obstime == origin.obstime):
+                raise ValueError(
+                    "obstime mismatch: frame obstime and origin.obstime must be equal."
+                )
+
+        # Validate location consistency
+        if location is not None and origin.location is not None:
+            if not np.all(location == origin.location):
+                raise ValueError(
+                    "location mismatch: frame location and origin.location must be equal."
+                )
+
+        # Validate shape consistency between origin and obstime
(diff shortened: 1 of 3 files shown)


gammapy/utils/coordinates/fov.py
```diff
@@ -24,6 +24,37 @@
...
+    if obstime is None:
+        return origin.obstime
+    if origin.obstime is None:
+        return obstime
+    if obstime.shape != origin.obstime.shape:
+        raise ValueError(
+            f"origin and obstime have inconsistent shapes: "
+            f"{origin.obstime.shape} vs {obstime.shape}."
+        )
+    if not np.all(obstime == origin.obstime):
```
AtreyeeS, line 38: We dont need to consider a tolerance here?

gammapy/utils/coordinates/fov.py
```diff
@@ -24,6 +24,37 @@
...
+        )
+    return obstime
+
+
+def _validate_location_consistency(location, origin):
+    """Check that location is consistent with origin.location and return location."""
+    if location is None:
+        return origin.location
+
+    if origin.location is not None and not np.all(location == origin.location):
```
AtreyeeS, line 50: same about the tolerance. What if `origin.location` is defined to 5 decimal points?

AtreyeeS (commented):
Thanks @registerrier ! I am slightly confused. One might want the frame to be fixed for a few minutes?

maxnoe:
I think the `__init__` is the wrong way to check this. You can have frames with and without data.

If you want to forbid the seemingly incorrect results that come from mismatches, you could validate in the transformation functions that the location and time attributes of the `to_frame / from_frame` are matching, i.e. forbid transforms between different reference frames.

> seemingly incorrect results that come from mismatches,

I say "seemingly", because the result is probably "correct" in the sense that it correctly converts to the FoV coordinates for the new location and time by converting to ICRS using the current values and then back to FoV using the new values, i.e. assuming the fixed sky position of these coordinates.

registerrier:
> I think the `__init__` is the wrong way to check this. You can have frames with and without data.
> 
> If you want to forbid the seemingly incorrect results that come from mismatches, you could validate in the transformation functions that the location and time attributes of the `to_frame / from_frame` are matching, i.e. forbid transforms between different reference frames.
> 
> > seemingly incorrect results that come from mismatches,
> 
> I say "seemingly", because the result is probably "correct" in the sense that it correctly converts to the FoV coordinates for the new location and time by converting to ICRS using the current values and then back to FoV using the new values, i.e. assuming the fixed sky position of these coordinates.

Thanks @maxnoe . 
Part of the problem comes from the fact that we are using `origin` to compute offset `theta` values with `origin.separation`. In that case, it becomes necessary to update the position of `origin` to match the different `obstime`.  See e.g. #6299 .

Imposing that `origin` and `obstime` match alleviates the issue but might not be the better approach.
Of course the issue is that `origin` does not need to be the center of the FoV at a given time.

The better approach is probably to compute `theta` as the separation with `SkyCoord(fov_lon=0, fov_lat=0, unit="deg" , frame=fov_frame)` .

An even cleaner solution could be to use a different `Representation` to access `theta` and `phi` directly. Unfortunately `PhysicsSphericalRepresentation` does not match and one would have to design a new one. 
Right?

maxnoe:
> The better approach is probably to compute theta as the separation with SkyCoord(fov_lon=0, fov_lat=0, unit="deg" , frame=fov_frame) .

Yes, that should always work and avoid going through potentially expensive transformations as `self.frame is other.frame`

maxnoe:
> An even cleaner solution could be to use a different Representation to access theta and phi directly. Unfortunately PhysicsSphericalRepresentation does not match and one would have to design a new one.
Right?

Yes, this is an open issue I tried to tackle a long time ago, didn't manage to finish the implementation and than abandoned, sorry...

https://github.com/astropy/astropy/issues/10783

https://github.com/astropy/astropy/pull/12845

maybe something you could revive?

registerrier:
This needs a better, cleaner approach. Let's postpone to 2.2.

PR #6458, opened 2026-02-27
Remove `FixedPointingInfo` deprecation warnings

### LINKED ISSUE #6445: Remove deprecation for `FixedPointingInfo`
We have two deprecation's that we have kept in the code base for a long time now (~ 2years). 
We should remove those and ensure that the code base does not utilise the older versions. 

https://github.com/gammapy/gammapy/blob/72c7a4b53d4cd5903de0d7119efffeee0c8af768/gammapy/data/observations.py#L302-L307

https://github.com/gammapy/gammapy/blob/72c7a4b53d4cd5903de0d7119efffeee0c8af768/gammapy/data/pointing.py#L133-L137

Manimaran-tech:
Closes #6445 
## Summary

Remove `GammapyDeprecationWarning` usages related to `FixedPointingInfo` and enforce strict type checking.

### Core
- **`observations.py`**: Replace deprecation warning + implicit `SkyCoord -> FixedPointingInfo` conversion with a `TypeError`
- - **`pointing.py`**: Replace deprecation warning for NaN coordinates with a `ValueError`
- - **`makers/utils.py`**: Fix `get_icrs()` calls in `make_map_exposure_true_energy`, `make_psf_map`, and `make_edisp_map` (removed invalid `pointing.obstime` attribute access)
### Tests & Examples
- Updated `test_map.py`, `test_sed.py`, `test_utils.py`, `plot_stack.py`, and `irfs.py` to use `FixedPointingInfo` instead of `SkyCoord` for pointing arguments
## Testing

Full test suite: **1577 passed**, 833 skipped, 9 xfailed, **0 failures**


examples/tutorials/details/irfs.py
```diff
@@ -278,12 +279,13 @@ class EffectiveArea3D(IRF):
 
 axis = MapAxis.from_energy_bounds(0.1 * u.TeV, 10 * u.TeV, 6, name="energy_true")
 pointing = SkyCoord(2, 1, unit="deg")
+pointing_info = FixedPointingInfo(fixed_icrs=pointing.icrs)
 geom = WcsGeom.create(npix=(4, 3), binsz=2, axes=[axis], skydir=pointing)
 
 print(geom)
 
 exposure_map = make_map_exposure_true_energy(
-    pointing=pointing, livetime="42 h", aeff=aeff_3d, geom=geom
```
AtreyeeS, line 286: Ah too bad that we were exposing a very deprecated feature in our tutorial

AtreyeeS (commented):
Thanks @Manimaran-tech 
Looks mostly good to me.

examples/tutorials/details/irfs.py
Astro-Kirsty: I actually think this is supposed to just be a `pointing` and does not need to be a `FixedPointingInfo`. You can see from the tutorial output [here](https://docs.gammapy.org/2.0.1/tutorials/details/irfs.html#create-exposure-map-dl4-product), that if it was deprecated we would receive the warning.

Astro-Kirsty: Maybe we can also adjust this in the future but for now `make_map_exposure_true_energy` accepts `SkyCoord`

gammapy/data/observations.py
Astro-Kirsty: We should adapt this now to only include `FixedPointingInfo`

gammapy/data/observations.py
```diff
@@ -300,11 +299,7 @@ def create(
         )
 
         if not isinstance(pointing, FixedPointingInfo):
-            warnings.warn(
-                "Pointing will be required to be provided as FixedPointingInfo",
-                GammapyDeprecationWarning,
-            )
-            pointing = FixedPointingInfo.from_fits_header(obs_info)
+            raise TypeError("pointing must be a FixedPointingInfo instance")
```
Astro-Kirsty: I would also remove this error, and in this manner we have deprecated the functionality completely. The message has already been here for +2 years

gammapy/data/pointing.py
```diff
@@ -127,10 +126,7 @@ def __init__(self, fixed_icrs=None, fixed_altaz=None):
             _check_coord_frame(fixed_icrs, ICRS, "fixed_icrs")
 
             if np.isnan(fixed_icrs.ra.value) or np.isnan(fixed_icrs.dec.value):
-                warnings.warn(
-                    "In future, fixed_icrs must have non-nan values",
-                    GammapyDeprecationWarning,
-                )
+                raise ValueError("fixed_icrs must have non-nan values")
```
Astro-Kirsty, line 128: Again, remove this error.

gammapy/datasets/tests/test_map.py
```diff
@@ -149,7 +149,7 @@ def get_exposure(geom_etrue):
     aeff = EffectiveAreaTable2D.read(filename, hdu="EFFECTIVE AREA")
 
     exposure_map = make_map_exposure_true_energy(
-        pointing=SkyCoord(1, 0.5, unit="deg", frame="galactic"),
+        pointing=FixedPointingInfo(fixed_icrs=SkyCoord(1, 0.5, unit="deg", frame="galactic").icrs),
```
Astro-Kirsty: For now `make_map_exposure_true_energy` accepts a `SkyCoord`

gammapy/datasets/tests/test_map.py
```diff
@@ -173,7 +173,7 @@ def get_psf():
 
     return make_psf_map(
         psf=psf,
-        pointing=SkyCoord(0, 0.5, unit="deg", frame="galactic"),
+        pointing=FixedPointingInfo(fixed_icrs=SkyCoord(0, 0.5, unit="deg", frame="galactic").icrs),
```
Astro-Kirsty: For now `make_psf_map` accepts a `SkyCoord`

gammapy/estimators/points/tests/test_sed.py
```diff
@@ -72,7 +72,7 @@ def simulate_spectrum_dataset(model, random_state=0):
         region="icrs;circle(0, 0, 0.1)", axes=[energy_axis_true]
     )
     exposure = make_map_exposure_true_energy(
-        pointing=SkyCoord("0d", "0d"), aeff=aeff, livetime=100 * u.h, geom=geom_true
+        pointing=FixedPointingInfo(fixed_icrs=SkyCoord("0d", "0d").icrs), aeff=aeff, livetime=100 * u.h, geom=geom_true
```
Astro-Kirsty: As above

gammapy/makers/tests/test_utils.py
```diff
@@ -93,7 +93,7 @@ def geom(map_type, ebounds):
 )
 def test_make_map_exposure_true_energy(aeff, pars):
     m = make_map_exposure_true_energy(
-        pointing=SkyCoord(2, 1, unit="deg"),
+        pointing=FixedPointingInfo(fixed_icrs=SkyCoord(2, 1, unit="deg").icrs),
```
Astro-Kirsty: As above

Astro-Kirsty (changes requested):
Thanks @Manimaran-tech.

For now I think we can safely remove all messages for the deprecation, but there are some functions here which are still allowing a `SkyCoord` and are different to the `Observation.create`. I would leave those as is, and we should think about how we want to proceed in the future. 
Thoughts @AtreyeeS, @registerrier ?

Manimaran-tech:
Thanks @Astro-Kirsty and @AtreyeeS for the detailed review!

I've addressed all comments 
now it should work ig!!

docs/user-guide/datasets/plot_stack.py
```diff
@@ -31,7 +31,7 @@
 
 observation = Observation.create(
     obs_id=0,
-    pointing=SkyCoord("0d", "0d", frame="icrs"),
+    pointing=FixedPointingInfo(fixed_icrs=SkyCoord("0d", "0d", frame="icrs").icrs),
```
AtreyeeS: Why do you need the `.icrs` here?
`fixed_icrs=SkyCoord("0d", "0d", frame="icrs")` should work, no?

gammapy/makers/utils.py
```diff
@@ -109,7 +109,7 @@ def make_map_exposure_true_energy(
         Exposure map.
     """
     if isinstance(pointing, FixedPointingInfo):
-        origin = pointing.get_icrs(pointing.obstime)
+        origin = pointing.get_icrs()
```
AtreyeeS, line 116: Why did you remove the `obstime`?

gammapy/makers/utils.py
AtreyeeS: adapt docstring to mention this can be a `FixedPointingInfo`

AtreyeeS: adapt docstring as before

AtreyeeS: adapt docstring

AtreyeeS: apapt docstring

gammapy/estimators/points/tests/test_sed.py
```diff
@@ -6,7 +6,6 @@
 from astropy.coordinates import EarthLocation, SkyCoord
 from astropy.table import Table
 from gammapy.data import Observation
-from gammapy.data.pointing import FixedPointingInfo
```
AtreyeeS, line 9: Why are you removing this import? This is used in line 123

gammapy/datasets/tests/test_map.py
```diff
@@ -13,7 +13,7 @@
 from regions import CircleSkyRegion
 import gammapy.irf.psf.map as psf_map_module
 from gammapy.catalog import SourceCatalog3FHL
-from gammapy.data import GTI, DataStore, Observation, FixedPointingInfo
```
AtreyeeS, line 16: `FixedPointingInfo` is used in line 2503
Why are you removing this import?

AtreyeeS (changes requested):
I left some comments inline.
For now, our irf maker functions accept both `SkyCoord` and `FixedPointingInfo` and I would keep it like that.
Please update the docstrings accordingly.

This PR will also need a towncrier fragment.
Please follow the instructions as mentioned on our developer guide

gammapy/makers/utils.py
Manimaran-tech: FixedPointingInfo file does not have an obstime attribute. The original code pointing.get_icrs(pointing.obstime) would fail with an AttributeError. For PointingMode.POINTING, 
get_icrs() works without obstime , it simply returns the fixed ICRS coordinates. obstime is only required for DRIFT mode (to transform alt-az to ICRS)

i hope i made it clear !!

docs/user-guide/datasets/plot_stack.py
Manimaran-tech: You're right! .icrs is unnecessary since the SkyCoord is already created with frame="icrs". Removed it. Thanks for catching that!

gammapy/makers/utils.py
Manimaran-tech: Done! Updated the docstrings for all 4 functions to mention both SkyCoord and FixedPointingInfo.

gammapy/estimators/points/tests/test_sed.py
Manimaran-tech: Sorry, my mistake! I incorrectly removed this import. Restored it now — FixedPointingInfo is indeed used in the simulate_map_dataset function.

gammapy/datasets/tests/test_map.py
Manimaran-tech: Apologies for the oversight! Restored the FixedPointingInfo import — it's used in test_map_dataset_hpx_evaluation_with_model.

Manimaran-tech:
Added a towncrier fragment at docs/release-notes/api/6445.api.rst.

gammapy/data/observations.py
```diff
@@ -276,15 +274,6 @@ def create(
             tstop = tstart + Quantity(livetime)
 
         gti = GTI.create(tstart, tstop, reference_time=reference_time)
-        obs_info = cls._get_obs_info(
```
AtreyeeS, line 279: why did you remove this?

docs/release-notes/feature/6445.feature.rst
```diff
@@ -0,0 +1 @@
+Remove ``GammapyDeprecationWarning`` from ``FixedPointingInfo`` and ``Observation.create``, enforcing ``FixedPointingInfo`` as the required type for the ``pointing`` parameter.
```
AtreyeeS: file name `.api` is not supported. available types are `infrastructure`, `docs`, `feature` or `bugfix`.
@Astro-Kirsty What do you suggest here? `feature`?

gammapy/makers/utils.py
AtreyeeS: Can you add a test in `gammapy/makers/tests/test_utils` that this works correctly for pointings as FixedPointingInfo

AtreyeeS (changes requested):
Some futher comments :)

gammapy/data/observations.py
Manimaran-tech: The obs_info was previously only used by the deprecation fallback code: FixedPointingInfo.from_fits_header(obs_info). Since we removed that deprecation warning and now enforce FixedPointingInfo as required, obs_info became unused — ruff flagged it as F841 (local variable assigned but never used), so I removed it.

docs/release-notes/feature/6445.feature.rst
Manimaran-tech: Thanks for pointing that out! I've moved the fragment to docs/release-notes/feature/6445.feature.rst. Let me know if a different type (like bugfix) is more appropriate.

Manimaran-tech:
@AtreyeeS  Done! Added test_make_map_exposure_true_energy_fixed_pointing that tests make_map_exposure_true_energy with FixedPointingInfo instead of SkyCoord, verifying it produces the same results.

docs/release-notes/feature/6445.feature.rst
Astro-Kirsty: I would not actually add a fragment for this type of change, as it is simply cleanup to remove something we deprecated years ago

gammapy/makers/tests/test_utils.py
```diff
@@ -104,6 +104,36 @@ def test_make_map_exposure_true_energy(aeff, pars):
...
+            "sum": 2.387916e08,
+        },
+    ],
+)
+def test_make_map_exposure_true_energy_fixed_pointing(aeff, pars):
+    pointing = FixedPointingInfo(fixed_icrs=SkyCoord(2, 1, unit="deg"))
+    m = make_map_exposure_true_energy(
+        pointing=pointing,
+        livetime="42 s",
+        aeff=aeff,
```
Astro-Kirsty, line 123: I think you can rather use the pre defined aeff and geom at the start of this file

docs/release-notes/feature/6445.feature.rst
Manimaran-tech: Thanks @Astro-Kirsty! Removed the fragment file.

gammapy/makers/tests/test_utils.py
Manimaran-tech: Thanks @Astro-Kirsty! Simplified the test to use the predefined aeff fixture and geom() helper directly, removing the duplicated parametrize block.

Manimaran-tech:
Hi @Astro-Kirsty @AtreyeeS, I've addressed all the review comments:

Removed the fragment file
Simplified the test to use the predefined aeff fixture and  geom() helper

Please let me know if there's anything else that needs to be changed, otherwise I believe this is ready for merge. Thanks!

gammapy/makers/tests/test_utils.py
```diff
@@ -104,6 +107,99 @@ def test_make_map_exposure_true_energy(aeff, pars):
...
+
+    psfmap = make_psf_map(psf, pointing, test_geom)
+
+    assert psfmap.psf_map.geom.axes[0] == geom_rad_axis
+    assert psfmap.psf_map.geom.axes[1] == energy_axis
+    assert psfmap.psf_map.unit == "deg-2"
+    assert psfmap.psf_map.data.shape == (4, 50, 25, 25)
+
+
+def test_make_edisp_map_fixed_pointing():
```
Astro-Kirsty: As above I would rename to `test_make_edisp_map` and test both options here.

gammapy/makers/tests/test_utils.py
```diff
@@ -104,6 +107,99 @@ def test_make_map_exposure_true_energy(aeff, pars):
...
+        aeff=aeff,
+        geom=test_geom,
+    )
+
+    assert m.data.shape == (2, 3, 4)
+    assert m.unit == "m2 s"
+    assert_allclose(m.data.sum(), 8.103974e08, rtol=1e-5)
+
+
+def test_make_psf_map_fixed_pointing():
```
Astro-Kirsty: I would rename to `test_make_psf_map()` and then we can test both options.

A more simple solution is to use our psf from `gammapy-data`:

```
from gammapy.irf import EnergyDependentMultiGaussPSF
from gammapy.makers.utils import make_psf_map
from gammapy.maps import WcsGeom, Map
from gammapy.datasets.map import RAD_AXIS_DEFAULT

filename = (
    "$GAMMAPY_DATA/cta-1dc/caldb/data/cta/1dc/bcf/South_z20_50h/irf_file.fits"
)
psf = EnergyDependentMultiGaussPSF.read(filename, hdu="POINT SPREAD FUNCTION")

geom = WcsGeom.create(
    skydir=(0, 0),
    frame="galactic",
    binsz=2,
    width=(2, 2),
    axes=[RAD_AXIS_DEFAULT, psf.axes["energy_true"]],
)

psfmap = make_psf_map(
    psf=psf,
    pointing=SkyCoord(0, 0.5, unit="deg", frame="galactic"),
    geom=geom,
    exposure_map=Map.from_geom(geom.squash("rad"), unit="cm2 s"),
)
```

Then you also create a psfmap2 that uses the fixed pointing

gammapy/makers/tests/test_utils.py
```diff
@@ -104,6 +107,99 @@ def test_make_map_exposure_true_energy(aeff, pars):
     assert_allclose(m.data.sum(), pars["sum"], rtol=1e-5)
 
 
+@requires_data()
+def test_make_map_exposure_true_energy_fixed_pointing(aeff):
```
Astro-Kirsty: Again, I would just add this to the `test_make_map_exposure_true_energy` test and add the other option

gammapy/makers/tests/test_utils.py
```diff
@@ -104,6 +107,99 @@ def test_make_map_exposure_true_energy(aeff, pars):
...
+    pointing_coord = SkyCoord(0, 0, frame="icrs", unit="deg")
+    pointing = FixedPointingInfo(fixed_icrs=pointing_coord)
+    test_geom = WcsGeom.create(10, binsz=0.5, axes=[migra, etrue])
+    edispmap = make_edisp_map(edisp, pointing, test_geom)
+
+    assert edispmap.edisp_map.data.shape == (6, 49, 10, 10)
+    assert edispmap.edisp_map.unit == ""
+
+
+def test_make_edisp_kernel_map_fixed_pointing():
```
Astro-Kirsty: I would not create another function but rather add it to `test_make_edisp_kernel_map`, and simply create a second `edispmap` i.e.:
```
pointing2 = FixedPointingInfo(fixed_icrs=pointing)
edispmap2 = make_edisp_kernel_map(edisp, pointing, geom)
```

Astro-Kirsty (changes requested):
Thanks! Just some comments relating to the tests

gammapy/makers/utils.py
maxnoe: The time should stay, if this results in an error, the underlying cause should be fixed.

The API was designed to make it explicit that pointing is always a time dependent quantity.

maxnoe: > The original code `pointing.get_icrs(pointing.obstime)`,

the error is not passing the time to `get_icrs`, it is looking up the time on the `pointing` object.

gammapy/data/pointing.py
```diff
@@ -126,12 +124,6 @@ def __init__(self, fixed_icrs=None, fixed_altaz=None):
         if fixed_icrs is not None:
             _check_coord_frame(fixed_icrs, ICRS, "fixed_icrs")
 
-            if np.isnan(fixed_icrs.ra.value) or np.isnan(fixed_icrs.dec.value):
```
maxnoe, line 129: Instead of removing this check, it needs to stay with the warning converted to a error

gammapy/data/pointing.py
maxnoe: Why? The deprecation is for passing in nan values in the `SkyCoord`, so the correct change is turning the warning into an error.

gammapy/makers/utils.py
AtreyeeS: @maxnoe  We had an incorrect code here (untested!!!). There is no `FixedPointingInfo.obstime` attribute. Since this function currently works for pointing observations only, the time is not relevant

maxnoe: > There is no FixedPointingInfo.obstime attribute.

That was the point of my comment. You need to get the observation time either from the observation or just provide it as parameter.

Manimaran-tech: Thanks @maxnoe and @AtreyeeS for the discussion!

Just to clarify the current situation — in make_map_exposure_true_energy, when the pointing is a FixedPointingInfo in POINTING mode, calling get_icrs() works fine without passing obstime, since the coordinates are already fixed in ICRS and no transformation is needed.

The issue with the original code (pointing.get_icrs(pointing.obstime)) was that FixedPointingInfo doesn’t have an .obstime attribute, which caused the failure.

If you'd prefer that obstime is passed explicitly, I’m happy to adjust things. I can either:

Add an optional obstime parameter to make_map_exposure_true_energy (and the related maker functions) and forward it to get_icrs() when provided, or

Handle it in a different way — whatever you think fits best.

Let me know your preference, and I’ll implement it right away.

gammapy/makers/utils.py
```diff
@@ -283,7 +283,7 @@ def make_psf_map(psf, pointing, geom, exposure_map=None):
         The resulting PSF map.
     """
     if isinstance(pointing, FixedPointingInfo):
-        origin = pointing.get_icrs(pointing.obstime)
+        origin = pointing.get_icrs()
```
bkhelifi, line 290: In the FixedPointingInfo is in Alt/Az, then an obstime is mandatory.
Do we have test functions for a pointing fixed in ICRS and in Alt/Az for this function? and the same for the others?

gammapy/makers/utils.py
AtreyeeS: I think as a depreciation warning cleanup, this is fine.
Further PRs can adapt passing `time` to the maker functions, which is beyond the scope of this PR

Manimaran-tech:
Hi @Astro-Kirsty  I've just pushed the changes you suggested:

Merged all the separate _fixed_pointing tests into the existing test functions, so each test now verifies both SkyCoord and FixedPointingInfo pointing options
Rewrote test_make_psf_map using EnergyDependentMultiGaussPSF from gammapy-data as you recommended
Cleaned up the unused PSF3D import
Let me know if everything looks good!

gammapy/makers/utils.py
bkhelifi: Again, let's wait the merge of the PR of @registerrier and the add of fixed pointîg in Alt/Az.
I suspect that the current change will break our usage of data from SWGO and Km3Net.
@QRemy and @Tobychev, can you also make a review?

bkhelifi (changes requested):

gammapy/makers/utils.py
AtreyeeS: @bkhelifi  We were never correctly handling this actually. The above line of code is broken (there is no `pointing.obstime` attribute), and we didn't have tests to catch it.

Manimaran-tech:
Hi @bkhelifi, I saw that you requested changes on this PR, but I don't see any comments or details attached to the review. Could you please let me know what needs to be updated? Thanks!

gammapy/data/data_store.py
```diff
@@ -571,13 +570,15 @@ def copy_obs(self, obs_id, outdir, hdu_class=None, verbose=False, overwrite=Fals
...
-            subprocess.run(cmd)
+            import shutil
+
+            source_path = Path(loc.path())
+            dest_path = targetdir / source_path.name
+
+            if overwrite or not dest_path.exists():
+                shutil.copy2(source_path, targetdir)
+                if verbose:
+                    log.info(f"Copied {source_path} to {targetdir}")
```
Astro-Kirsty: Changing these lines are outside of this PR, I would leave them as is

gammapy/makers/utils.py
Astro-Kirsty: What is the current idea here? Do we continue with the merge of this PR, as we are just trying to resolve the deprecation?

Astro-Kirsty (commented):
Thank you @Manimaran-tech. The tests are passing. 
Just a minor comment.

Otherwise, I ask for further review on how to proceed here. 
@AtreyeeS, @registerrier, @QRemy ?

gammapy/data/data_store.py
```diff
@@ -571,13 +570,15 @@ def copy_obs(self, obs_id, outdir, hdu_class=None, verbose=False, overwrite=Fals
             loc = subhdutable.location_info(idx)
             targetdir = outdir / loc.file_dir
             targetdir.mkdir(exist_ok=True, parents=True)
-            cmd = ["cp"]
-            if verbose:
-                cmd += ["-v"]
-            if not overwrite:
-                cmd += ["-n"]
-            cmd += [str(loc.path()), str(targetdir)]
```
AtreyeeS, line 579: Why did you change these lines of code?

gammapy/data/pointing.py
AtreyeeS: Please add a test to ensure `ValueError` is raised with correct message

gammapy/makers/tests/test_utils.py
```diff
@@ -103,6 +105,74 @@ def test_make_map_exposure_true_energy(aeff, pars):
...
+    )
+
+    assert m2.data.shape == pars["shape"]
+    assert m2.unit == "m2 s"
+    assert_allclose(m2.data.sum(), pars["sum"], rtol=1e-5)
+
+
+@requires_data()
+def test_make_psf_map():
+    from gammapy.irf import EnergyDependentMultiGaussPSF
```
AtreyeeS: Usually we put all imports at the start unless they lead to circular imports

gammapy/makers/tests/test_utils.py
```diff
@@ -103,6 +105,74 @@ def test_make_map_exposure_true_energy(aeff, pars):
...
+    )
+    psf = EnergyDependentMultiGaussPSF.read(filename, hdu="POINT SPREAD FUNCTION")
+
+    pointing_coord = SkyCoord(0, 0.5, unit="deg", frame="galactic")
+    test_geom = WcsGeom.create(
+        skydir=(0, 0),
+        frame="galactic",
+        binsz=2,
+        width=(2, 2),
+        axes=[RAD_AXIS_DEFAULT, psf.axes["energy_true"]],
```
AtreyeeS: decrease number of bins in RAD and energy_true

gammapy/makers/tests/test_utils.py
```diff
@@ -103,6 +105,74 @@ def test_make_map_exposure_true_energy(aeff, pars):
...
+
+@requires_data()
+def test_make_psf_map():
+    from gammapy.irf import EnergyDependentMultiGaussPSF
+    from gammapy.datasets.map import RAD_AXIS_DEFAULT
+
+    filename = (
+        "$GAMMAPY_DATA/cta-1dc/caldb/data/cta/1dc/bcf/South_z20_50h/irf_file.fits"
+    )
+    psf = EnergyDependentMultiGaussPSF.read(filename, hdu="POINT SPREAD FUNCTION")
```
AtreyeeS: I will put the psf also as a fixture

AtreyeeS (changes requested):
Thanks @Manimaran-tech !

Left some comments inline

AtreyeeS:
@Astro-Kirsty I think we can merge this PR when ready and the makes to work for AltAz pointings in a different feature PR

Manimaran-tech:
Hi @Astro-Kirsty  and @AtreyeeS, thank your for the feedbacks! 

I have pushed the requested changes:
* **Reverted** the `cmd = ["cp"]` change in [gammapy/data/data_store.py](cci:7://file:///c:/Users/K/Desktop/my%20python%20programs/Contribution/gammapy/gammapy/data/data_store.py:0:0-0:0) to keep this PR strictly focused on the deprecation warning.
* **Moved** the `EnergyDependentMultiGaussPSF` import to the top of [test_utils.py](cci:7://file:///c:/Users/K/Desktop/my%20python%20programs/Contribution/gammapy/gammapy/makers/tests/test_utils.py:0:0-0:0) as requested. 
* **Decreased** the number of bins in `RAD` and `energy_true` in [test_make_psf_map](cci:1://file:///c:/Users/K/Desktop/my%20python%20programs/Contribution/gammapy/gammapy/makers/tests/test_utils.py:130:0-154:66) to improve test speed.
* **Extracted** the PSF setup into a modular `@pytest.fixture` to clean up the test function.

registerrier:
Dev call decision: postpone to 2.2 because we need a proper handling of coordinates transofrm on Observation. This requires changing the API of `make_exposure_map`

Astro-Kirsty:
@QRemy, @AtreyeeS, @registerrier we need to make a decision here on how we want to proceed

Astro-Kirsty:
This requires further discussion before 2.2 and the exact solution is currently not clear. Closing for now.

PR #6461, opened 2026-02-28
Adapt spectral simulation notebook to account for containment correction

### LINKED ISSUE #6245: Containment correction in spectral simulation
In our current 1D spectral simulation, we have an `on_region`, but the containment correction is not applied, see https://github.com/gammapy/gammapy/blob/0733504a881a753d42a5f6ad0d74555a2591a18d/examples/tutorials/analysis-1d/spectrum_simulation.py#L145

In fact, the PSF is never taken into account at all.

There is no bias in the fit as we are using the same IRFs for the simulation and the fit, but I am a bit confused if we are overestimating the effective area.

I see this is the behaviour since v0.15, wondering if this is correct. Opinions @registerrier @bkhelifi @adonath @QRemy  ?

AtreyeeS:
Fixes #6245

### DIFF
--- a/examples/tutorials/analysis-1d/spectrum_simulation.py
+++ b/examples/tutorials/analysis-1d/spectrum_simulation.py
@@ -135,6 +135,7 @@ print(obs)
 ######################################################################
 # Simulate a spectrum
 #
+#
 # Note that here we are using full containment IRFs and thus set
 # ``containment_correction=True`` in the `~gammapy.makers.SpectrumDatasetMaker` and use
 # a circular on region. If you have pointlike IRFs, please set


examples/tutorials/analysis-1d/spectrum_simulation.py
```diff
@@ -133,7 +133,12 @@
 print(obs)
 
 ######################################################################
-# Simulate a spectra
+# Simulate a spectra:
```
morcuended:
```suggestion
# Simulate a spectrum
```
Also note that this is render in the docs in the same line even if here it is broken into two lines, which is fine, but just saying.

> Simulate a spectra: Note that here we are using ...

examples/tutorials/analysis-1d/spectrum_simulation.py
```diff
@@ -133,7 +133,12 @@
...
+# Note that here we are using full containment IRFs and thus set
+# `containment_correction=True` in the `SpectrumDatasetMaker` and use
+# a circular on region. If you have pointlike IRFs, please set
+# `containment_correction=False` and use a `~regions.PointSkyRegion`.
+# See :doc:`/tutorials/data/magic` tutorial for details
 #
```
Astro-Kirsty, line 144:
```suggestion
# Note that here we are using full containment IRFs and thus set
# ``containment_correction=True`` in the `~gammapy.makers.SpectrumDatasetMaker` and use
# a circular on region. If you have pointlike IRFs, please set
# ``containment_correction=False`` and use a `~regions.PointSkyRegion`.
# See :doc:`/tutorials/data/magic` tutorial for details.
#
```

examples/tutorials/analysis-1d/spectrum_simulation.py
Astro-Kirsty: I would event leave this out and just keep it from 'Note'.

examples/tutorials/analysis-1d/spectrum_simulation.py
```diff
@@ -133,7 +133,12 @@
 print(obs)
 
 ######################################################################
-# Simulate a spectra
+# Simulate a spectrum
```
Astro-Kirsty, line 136: I would event leave this out and just keep it from 'Note'.

QRemy, line 136:
```suggestion
# Simulate a spectrum
#
```

bkhelifi:
@AtreyeeS, can you check whether the fit results change with and without the containment correction?
I think that it was also a source of concern...

AtreyeeS:
@bkhelifi The best fit values dont change - we were not considering the psf in the simulations and the fit, so consistent there. 

The error bars increase on accounting for the psf. 

I tested with fake psf of various sizes, and the number of source counts varies accordingly now, so it seems we are ok

PR #6464, opened 2026-03-01
Add a function to rescale a dataset for a new livetime

AtreyeeS:
This can be useful is someone has a DL4 dataset and wants to see if certain parameters could be constrained with additional data.
I kept the option to pass `old_livetime` as sometimes the exposure does not contain a meta information, eg in case of Fermi-LAT.

### DIFF
--- a/gammapy/datasets/map.py
+++ b/gammapy/datasets/map.py
@@ -3514,9 +3514,10 @@ class MapDatasetOnOff(MapDataset):
         dataset : `MapDataset` or `SpectrumDataset`
             Dataset assuming a new livetime
         """
-        dataset = super().reset_livetime(new_livetime, old_livetime)
-
+        if old_livetime is None:
+            old_livetime = self.exposure.meta.get("livetime")
         ratio = (new_livetime / old_livetime).to("")
+        dataset = super().reset_livetime(new_livetime, old_livetime)
 
         counts_off = None
         if self.counts_off is not None:


examples/tutorials/data/magic.py
```diff
@@ -381,3 +381,22 @@
...
+
+dataset = datasets.stack_reduce()
+print(f"Current livetime: {dataset.exposure.meta['livetime'].to(u.h)}")
+dataset_new = dataset_simulated.reset_livetime(new_livetime=6 * u.h)
+
+########################################################################
+# We have a new dataset with 6h of livetime. You can now simulate spectra
+# with it as described in the previous section. Note that this will give you a
+# rough idea only, as we are just scaling the off counts. If you have additional
+# off measurements, you can use them instead.
```
QRemy, line 401: doesn't sound like good practice if we have to give this caveat

gammapy/datasets/map.py
```diff
@@ -2376,6 +2377,72 @@ def to_image(self, name=None):
         energy_axis = self._geom.axes["energy"].squash()
         return self.resample_energy_axis(energy_axis=energy_axis, name=name)
 
+    def reset_livetime(self, new_livetime, old_livetime=None):
```
QRemy, line 2380: would be simple have only a `livetime_factor` as parameter

AtreyeeS:
@QRemy I simulated a Gaussian and with an overlapping fainter point source and tried to compute the TS of the extension with livetime.

- For an Asimov dataset, the TS scales linearly with the livetime, but the slope is less than one depending on the level of confusion
- For a fake dataset, there is more scatter (even when I use the same random number for different livetimes). I suppose this is expected

Since with one observation, we cant know the "slope" for the scaling, I guess having the option for scaling the livetime might be still useful.

image

AtreyeeS:
Also, when I simulate two overlapping point sources, with enough exposure, the Asimov dataset is able to separate the two to (apparently) arbitrarily small distances, but this does not hold true after doing the fake.

QRemy (commented):
Thanks @AtreyeeS I'm not convinced it should be a separate method, adding an option as `livetime_factor=1` on the `.fake()` and .`_to_asimov_dataset()` method would be simpler. Also showing it for an OnOff dataset in the doc is maybe not the best.

gammapy/datasets/map.py
AtreyeeS: Ok

bkhelifi:
The need of this feature is clear. But, it looks to me that it is not evonomic in term of memory, just for a float update!
I would rather prefer that the user decides herself/himself to make a new dataset or not, having a simple function to update the needed elements with a new livetime.

The fonction mane is not intuitive, it is not a reset but a modification. 

The tests fail because of a reduced code coverage.

PR #6466, opened 2026-03-01
Fix TSmapEstimator flux estimate

### LINKED ISSUE #6448: TSMapEstimator unexpected behaviour
Using Gammapy 2.0.1 the example below does not work as reported:
https://gammapy.github.io/gammapy-recipes/_build/html/notebooks/ts-vs-excess-map-estimator/ts-vs-exess-map-estimator.html

TSMapEstimator with a disk model is expected to give similar results to ExcessMapEstimator, as shown in the link above:
Image

But with 2.0.1, what I get is pretty different:

Image

**To Reproduce**
Just run the gammapy example notebook in the link above.

Actually I came to this after finding an unexpected (at least for me) behaviour of TSMapEstimator when **not** using a disk model, but the actual PSF (same used to create a fake dataset). For a FOV with a single  point-like source, the resulting flux map  was essentially flat up to a large offset angle (basically the region where the psf was defined). I mention it in case anyone has come across similar behaviour.

QRemy:
Resolves #6448

Try out:
[ts-vs-exess-map-estimator.py](https://github.com/user-attachments/files/25667246/ts-vs-exess-map-estimator.py)

### DIFF
--- a/docs/release-notes/6466.bug.rst
+++ b/docs/release-notes/6466.bug.rst
@@ -1,0 +1,1 @@
+Fix flux estimates in `gammapy.estimators.TSMapEstimator`
--- a/docs/release-notes/6466.bug.rst
+++ b/docs/release-notes/6466.bug.rst
@@ -1,1 +1,1 @@
-Fix flux estimates in `gammapy.estimators.TSMapEstimator`
+Fix flux estimates in `gammapy.estimators.TSMapEstimator` (now the kernel is properly normalized in each energy bin rather than over the total energy range).
--- a/gammapy/estimators/map/tests/test_ts.py
+++ b/gammapy/estimators/map/tests/test_ts.py
@@ -858,7 +858,9 @@ def test_excess_map_compatibility():
         rtol=1e-2,
     )
     assert_allclose(
-        np.median(result["flux"].data[mask] / result_ts_sum["flux"].data[mask]), 0.97
+        np.median(result["flux"].data[mask] / result_ts_sum["flux"].data[mask]),
+        0.97,
+        rtol=1e-2,
     )
 
     mask = np.isfinite(result["flux"].data) & np.isfinite(result_ts["flux"].data)
--- a/gammapy/estimators/map/tests/test_ts.py
+++ b/gammapy/estimators/map/tests/test_ts.py
@@ -873,9 +873,9 @@ def test_excess_map_compatibility():
         spatial_model=DiskSpatialModel(r_0=ndeg),
     )
 
-    ts_est_sum = TSMapEstimator(model=model, sum_over_energy_groups=True)
+    ts_est_sum = TSMapEstimator(kernel_model=model, sum_over_energy_groups=True)
 
-    ts_est = TSMapEstimator(model=model, sum_over_energy_groups=False)
+    ts_est = TSMapEstimator(kernel_model=model, sum_over_energy_groups=False)
 
     result = est.run(dataset=dataset)
 
--- a/gammapy/estimators/map/tests/test_ts.py
+++ b/gammapy/estimators/map/tests/test_ts.py
@@ -857,6 +857,9 @@ def test_excess_map_compatibility():
         0.96,
         rtol=1e-2,
     )
+    assert_allclose(
+        np.median(result["flux"].data[mask] / result_ts_sum["flux"].data[mask]), 0.97
+    )
 
     mask = np.isfinite(result["flux"].data) & np.isfinite(result_ts["flux"].data)
     assert_allclose(
--- a/gammapy/estimators/map/tests/test_ts.py
+++ b/gammapy/estimators/map/tests/test_ts.py
@@ -5,8 +5,9 @@ import numpy as np
 from numpy.testing import assert_allclose
 import astropy.units as u
 from astropy.coordinates import Angle, SkyCoord
+from scipy.stats import pearsonr
 from gammapy.datasets import Datasets, MapDataset, MapDatasetOnOff
-from gammapy.estimators import TSMapEstimator
+from gammapy.estimators import TSMapEstimator, ExcessMapEstimator
 from gammapy.estimators.utils import (
     approximate_profile_map,
     combine_flux_maps,
@@ -18,6 +19,7 @@ from gammapy.irf import EDispKernelMap, PSFMap
 from gammapy.maps import Map, MapAxis, WcsGeom
 from gammapy.modeling.models import (
     ConstantSpatialModel,
+    DiskSpatialModel,
     GaussianSpatialModel,
     PointSpatialModel,
     PowerLawSpectralModel,
@@ -819,3 +821,46 @@ def test_tsmap_extended_source():
             result["sqrt_ts"].data[0, 12, 16],
             rtol=5e-2,
         )
+
+
+@requires_data()
+def test_excess_map_compatibility():
+    dataset = MapDataset.read(
+        "$GAMMAPY_DATA/cta-1dc-gc/cta-1dc-gc.fits.gz", name="cta_dataset"
+    )
+
+    dataset = dataset.downsample(factor=2)
+    dataset.psf = None
+
+    ndeg = "0.1 deg"
+
+    est = ExcessMapEstimator(correlation_radius=ndeg)
+
+    model = SkyModel(
+        spectral_model=PowerLawSpectralModel(),
+        spatial_model=DiskSpatialModel(r_0=ndeg),
+    )
+
+    ts_est_sum = TSMapEstimator(model=model, sum_over_energy_groups=True)
+
+    ts_est = TSMapEstimator(model=model, sum_over_energy_groups=False)
+
+    result = est.run(dataset=dataset)
+
+    result_ts = ts_est.run(dataset)
+
+    result_ts_sum = ts_est_sum.run(dataset)
+
+    mask = np.isfinite(result["flux"].data) & np.isfinite(result_ts_sum["flux"].data)
+    assert_allclose(
+        pearsonr(result["flux"].data[mask], result_ts_sum["flux"].data[mask]).statistic,
+        0.96,
+        rtol=1e-2,
+    )
+
+    mask = np.isfinite(result["flux"].data) & np.isfinite(result_ts["flux"].data)
+    assert_allclose(
+        pearsonr(result["flux"].data[mask], result_ts["flux"].data[mask]).statistic,
+        0.85,
+        rtol=1e-2,
+    )
--- a/gammapy/estimators/map/tests/test_ts.py
+++ b/gammapy/estimators/map/tests/test_ts.py
@@ -535,7 +535,11 @@ def test_ts_map_asimov(fake_dataset):
     )
     maps = estimator.run(asimov_dataset)
 
-    assert_allclose(maps["dnde"].data[:, 25, 25], 1e-10, rtol=5e-2)
+    expected_dnde = model.spectral_model.amplitude.value
+    expected_flux = model.spectral_model.integral(0.1 * u.TeV, 10 * u.TeV).value
+
+    assert_allclose(maps["dnde"].data[:, 25, 25], expected_dnde, rtol=5e-2)
+    assert_allclose(maps["flux"].data[:, 25, 25], expected_flux, rtol=5e-2)
 
     estimator = TSMapEstimator(
         model,
@@ -545,7 +549,8 @@ def test_ts_map_asimov(fake_dataset):
     )
     maps = estimator.run(asimov_dataset)
 
-    assert_allclose(maps["dnde"].data[:, 25, 25], 1e-10, rtol=5e-2)
+    assert_allclose(maps["dnde"].data[:, 25, 25], expected_dnde, rtol=5e-2)
+    assert_allclose(maps["flux"].data[:, 25, 25], expected_flux, rtol=5e-2)
 
 
 def test_ts_map_with_model(fake_dataset):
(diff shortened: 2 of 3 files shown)


gammapy/estimators/map/tests/test_ts.py
```diff
@@ -796,3 +796,51 @@ def test_tsmap_extended_source():
...
+    result = est.run(dataset=dataset)
+
+    result_ts = ts_est.run(dataset)
+
+    result_ts_sum = ts_est_sum.run(dataset)
+
+    mask = np.isfinite(result["flux"].data) & np.isfinite(result_ts_sum["flux"].data)
+    assert_allclose(
+        pearsonr(result["flux"].data[mask], result_ts_sum["flux"].data[mask]).statistic,
+        0.96,
```
AtreyeeS, line 889: Are these test values statistically motivated? You are saying it can't be higher than 0.98?

AtreyeeS (dismissed):
Thanks @QRemy ! Great catch!

docs/release-notes/6466.bug.rst
```diff
@@ -0,0 +1 @@
+Fix flux estimates in `gammapy.estimators.TSMapEstimator`
```
morcuended: Maybe explain a bit more in this case: cause, how was affecting, what do you expect now, etc.

gammapy/estimators/map/tests/test_ts.py
QRemy: not with that precision it but it should be close to one

docs/release-notes/6466.bug.rst
QRemy: should I write in parentheses ?
`sum_over_energy_groups=True` case was wrong from version 1.3
`sum_over_energy_groups=False` case has always been wrong

moralejo:
Thanks @QRemy, I confirm that the other strange results I was getting with TSMapEstimator are also gone with this fix. Basically, for a simulated point-like source, the flux sky map obtained using a point-like model ~was flat on a very wide region around the source location (about the size of the kernel_width) - and~ the flux did not match the simulated one. Now the correct flux is reconstructed. 

**Edited**: the issue about "a very broad flux map" was not related to the fixed bug. This  just happens (both for old and new version) when computing a flux map from a simulated dataset of a bright point-like source on top of a negligible background. I do not think it is a bug, just the result of how the likelihood behaves in such an extreme case.

docs/release-notes/6466.bug.rst
QRemy:
```suggestion
Fix flux estimates in `gammapy.estimators.TSMapEstimator` (now the kernel is properly normalized in each energy bin rather than over the total energy range).
```

gammapy/estimators/map/tests/test_ts.py
```diff
@@ -249,13 +251,13 @@ def test_compute_ts_map_psf(fermi_dataset):
     )
     result = estimator.run(fermi_dataset)
 
-    assert_allclose(result["ts"].data[0, 29, 29], 830.97957, rtol=2e-3)
+    assert_allclose(result["ts"].data[0, 29, 29], 762.890201, rtol=2e-3)
```
registerrier, line 260: why does it change `ts` here? In the recipe example, the significances seemed very similar before and after 1.2.

gammapy/estimators/map/tests/test_ts.py
```diff
@@ -681,21 +683,21 @@ def test_joint_ts_map_hawc():
...
-    assert_allclose(result["stat_scan"].data[0, 0, 59, 59], 5193.588657, rtol=1e-3)
+    assert_allclose(result["stat_scan"].data[0, 0, 59, 59], 5093.431946, rtol=1e-3)
 
     estimator = TSMapEstimator(
         kernel_width=2 * u.deg, sum_over_energy_groups=True, n_jobs=4
     )
     result = estimator.run(datasets)
-    assert_allclose(result["flux"].data[0, 59, 59], 1.99452e-13, rtol=1e-3)
-    assert_allclose(result["sqrt_ts"].data[0, 59, 59], 11.997135, rtol=1e-3)
+    assert_allclose(result["flux"].data[0, 59, 59], 4.174005e-12, rtol=1e-3)
```
registerrier, line 756: Can't we have an idea of the expected value?

registerrier (commented):
Thanks @QRemy . This is an important correction.
- Could you detail the incorrect behavior and why this approach is correct? Why did the incorrect behaviour appeared in 1.3?
- It is good to have introduced a unit test to cover the recipe. 
- It is a bit annoying to have flux test values changed without knowing which is correct. Could we have a test where we fake a simple source with known flux?

registerrier (commented):
Thanks @QRemy . This is an important correction.

Could you detail the incorrect behavior and why this approach is correct? Why did the incorrect behaviour appeared in 1.3?

It is good to have introduced a unit test to cover the recipe. 

It is a bit annoying to have flux test values changed without knowing which is correct. Could we have a test where we fake a simple source with known flux?

registerrier (commented):
Thanks @QRemy . This is an important correction.

Could you detail the incorrect behavior and why this approach is correct? Why did the incorrect behaviour appeared in 1.3?

It is good to have introduced a unit test to cover the recipe. 

It is a bit annoying to have flux test values changed without knowing which is correct. Could we have a test where we fake a simple source with known flux?

QRemy:
> Could you detail the incorrect behavior and why this approach is correct? Why did the incorrect behaviour appeared in 1.3?

My undertanding of the problem is that the kernel has always been normalized over the full energy range instead of per nenrgy bin. So the case where `sum_over_energy_groups=False` has always been wrong. The case `sum_over_energy_groups=True` was ok up to v1.2 where the kernel was computed in 2d in that case. For v1.3 i introduced a fix to support the mask handling correctly which implied to compute the kernel in 3d even for the `sum_over_energy_groups=True` case and then weight by the mask so it end up to have the same problem as for the  `sum_over_energy_groups=False` case. Another problem was that `exposure_npred` was computed with `normalize=True` option that divide by the flux in each energy bin and then multiplied by the total  flux creating an inconsistency, computing directly `exposure_npred` with `normalize=False` prevent this.

QRemy:
> It is a bit annoying to have flux test values changed without knowing which is correct. Could we have a test where we fake a simple source with known flux?

I added a test on an asimov dataset but somehow it also pass on main that. The test against the excess map should be enough as regression test.

gammapy/estimators/map/tests/test_ts.py
QRemy: I added a test on an asimov dataset but somehow it also pass on main that. The test against the excess map should be enough as regression test.

gammapy/estimators/map/tests/test_ts.py
QRemy: in significance the difference will be relatively small

QRemy:

Astro-Kirsty:
I will have a look at it today!

Astro-Kirsty:
After having a brief look it is clear we need to understand this a bit further as we are getting different results from v1.1 which should not be expected if the change was introduced in v1.3.
I give here just two examples that are changing with the source detection, where we have missing/additional sources: [TSMapEstimator.pdf](https://github.com/user-attachments/files/25870860/TSMapEstimator.pdf)

QRemy:
The mask handling changed between 1.1 and 1.3 introducing `mask_invalid = (counts_cutout == 0) & (background_cutout == 0) & (model == 0)` and changes  in `estimate_mask_default`. The usage of `estimate_mask_default` also changed previously it was applied to the exposure while now the dataset mask is applied instead.

QRemy:
Also changed the default stopping criterion for brentq https://github.com/gammapy/gammapy/pull/5380 but it didn't affect the tests.

QRemy:
For the example https://docs.gammapy.org/1.1/tutorials/analysis-3d/cta_data_analysis.html?highlight=source%20detection#source-detection we have `sum_over_energy_groups=False` so we do expect differences as in this case the kernel is incorrectly normalised since introduced in v0.18

Astro-Kirsty:
Could you explain further why `sum_over_energy_groups=False` has always been wrong?

It does not seem to explain why this tutorial is different though:
https://docs.gammapy.org/dev/tutorials/analysis-2d/detect.html#source-candidates

Here we have it set to True

gammapy/estimators/map/ts.py
```diff
@@ -324,7 +326,10 @@ def estimate_kernel(self, dataset):
         )
 
         kernel = evaluator.compute_npred()
-        kernel.data /= kernel.data.sum()
+
+        with np.errstate(invalid="ignore", divide="ignore"):
+            kernel.data /= kernel.data.sum(axis=(1, 2))[:, None, None]
+            kernel.data[~np.isfinite(kernel.data)] = 0
```
QRemy, line 341: The normalisation should be applied per energy so previously sum_over_energy_group=False was wrong (while  sum_over_energy_group=True because the dataset was reduced to a 2d dataset before)

gammapy/estimators/map/ts.py
```diff
@@ -430,6 +435,9 @@ def estimate_fit_input_maps(self, dataset):
         # First create 2D map arrays
 
         exposure = estimate_exposure_reco_energy(dataset, self.model.spectral_model)
```
QRemy: here normalize=True so we have in **each energy bin** exposure = exposure_npred/flux

gammapy/estimators/map/ts.py
```diff
@@ -441,15 +449,14 @@ def estimate_fit_input_maps(self, dataset):
...
-        exposure *= mask
+        exposure_npred *= mask
 
         energy_axis = counts.geom.axes["energy"]
 
         flux_ref = self.model.spectral_model.integral(
             energy_axis.edges[0], energy_axis.edges[-1]
         )
 
-        exposure_npred = (exposure * flux_ref).to_unit("")
```
QRemy, line 465: but here
 exposure_npred_per_energy_bin = exposure * flux_ref
 = exposure_npred_per_energy_bin/**flux_per_energy_bin * flux_ref_total**
  which is wrong

gammapy/estimators/map/ts.py
Astro-Kirsty: Okay, so then I imagine we also have issues with `ASmoothMapEstimator`?

QRemy: no because in ASmoothMapEstimator we really need exposure and not exposure_npred
this line is wrong only used in combination with the line in the comment below

gammapy/estimators/map/ts.py
QRemy, line 756: This can explain the changes for the sum_over_axes case=True case:
Previously in order to get the kernel we were doing `dataset.to_image` before computing npred, so the psf and edisp were averaged before, while now we do the npred computation in 3d with the exact irfs and then compute the 2d kernel as a mask weighted average of the 3d kernel.

AtreyeeS (dismissed):

gammapy/estimators/map/ts.py
registerrier: Here I would add the `normalize=True` argument to be sure.

gammapy/estimators/map/ts.py
registerrier: Shouldn't the normalization scheme depend on the value of `sum_over_axis`?

gammapy/estimators/map/tests/test_ts.py
```diff
@@ -519,6 +521,38 @@ def test_ts_map_stat_scan_different_energy(fake_dataset):
...
+    estimator = TSMapEstimator(
+        model,
+        kernel_width=None,
+        energy_edges=[0.1, 10] * u.TeV,
+        sum_over_energy_groups=True,
+    )
+    maps = estimator.run(asimov_dataset)
+
+    assert_allclose(maps["dnde"].data[:, 25, 25], expected_dnde, rtol=5e-2)
+    assert_allclose(maps["flux"].data[:, 25, 25], expected_flux, rtol=5e-2)
```
registerrier, line 553: Thanks, the test is nice.
What is the typical relative errors that you have in this case? 
I would have imagined that in a situation like this, the performance would be better than 5%

gammapy/estimators/map/tests/test_ts.py
```diff
@@ -696,8 +730,8 @@ def test_joint_ts_map_hawc():
         kernel_width=2 * u.deg, sum_over_energy_groups=False, n_jobs=4
     )
     result = estimator.run(datasets)
-    assert_allclose(result["flux"].data[0, 59, 59], 1.909396e-13, rtol=1e-3)
-    assert_allclose(result["sqrt_ts"].data[0, 59, 59], 10.878956, rtol=1e-3)
+    assert_allclose(result["flux"].data[0, 59, 59], 4.034048e-12, rtol=1e-3)
```
registerrier, line 733: We had an error as large as a factor 20?

registerrier (dismissed):

Maybe you could just clarify why the normalization is not different in the case of `sum_over_energy_group`?

QRemy:
> Maybe you could just clarify why the normalization is not different in the case of `sum_over_energy_group`?

The kernel is also computed in 3d in that case and then in the loop over each pixel reduced to 2d taking into account the mask at this location in the weighting.

gammapy/estimators/map/ts.py
```diff
@@ -443,6 +448,9 @@ def estimate_fit_input_maps(self, dataset):
         exposure = estimate_exposure_reco_energy(
             dataset, self.kernel_model.spectral_model
```
QRemy:
```suggestion
            dataset, self.kernel_model.spectral_model, normalize=True
```

registerrier:
Merging now.

PR #6467, opened 2026-03-02
Support chained comparisons parsing in logic_parser

yaochengchen:
**Description**

This PR fixes a bug in `logic_parser` where chained comparisons (e.g., `1 < x < 3`) returned incorrect boolean arrays due to overriding the left comparator with a boolean result. 

The fix evaluates the AST comparators pairwise, merges them using `np.logical_and`, and includes a regression test.

**Dear reviewer**

### DIFF
--- a/gammapy/utils/scripts.py
+++ b/gammapy/utils/scripts.py
@@ -24,6 +24,7 @@ __all__ = [
     "recursive_merge_dicts",
     "to_yaml",
     "write_yaml",
+    "logic_parser",
 ]
 
 PATH_DOCS = Path(__file__).resolve().parent / ".." / ".." / "docs"


docs/release-notes/6467.bug.rst
```diff
@@ -0,0 +1 @@
+Fixed a bug in ``gammapy.utils.scripts.logic_parser`` where chained comparisons (e.g., ``1 < x < 3``) were evaluated incorrectly.
```
QRemy:
```suggestion
Support chained comparisons in ``gammapy.utils.scripts.logic_parser`` (e.g., ``1 < x < 3``).
```

gammapy/utils/tests/test_scripts.py
```diff
@@ -196,3 +196,24 @@ def test_logic_parser_error():
         expression = "unsupported_expression()"
         logic_parser(table, expression)
     assert "Unsupported expression type" in str(excinfo.value)
+
+
+def test_logic_parser_chained_comparisons():
+    data = {"OBS_ID": [0, 1, 2, 3, 4, 5]}
+    table = Table(data)
+
+    result = logic_parser(table, "1 < OBS_ID < 4")
```
QRemy, line 209: could you add a test where this is combined with other conditions ?

yaochengchen:
> Thanks @yaochengchen, looks good. I think this can be relabelled as feature because the docsting was not mentioning this as a possibility and the example is explicitly showing `expression = '(OBS_ID < 3) and (OBS_ID > 1) '` instead of `'1 < OBS_ID < 3'`. Please extend the docstring and add an example to show this is possible now. The fragment name should also be changed to feature.
> 
> By curiosity how did you come across this ?

@QRemy Thanks for the review. I’ve updated the PR as requested; regarding how I came across this: I’m currently training an ML model and learning Gammapy since I’m new to this field, and while waiting for training jobs I used AI tools to analyze the codebase, which suggested this improvement, so I looked into it and opened a PR.

QRemy (dismissed):
Thanks @yaochengchen, just on last comment : could you add the `logic_parser` to `__all__` so it shows up in the doc.

yaochengchen:
> Thanks @yaochengchen, just on last comment : could you add the `logic_parser` to `__all__` so it shows up in the doc.

Done as you requested.

PR #6472, opened 2026-03-02
Change UL limit to 5-sigma is non-detection tutorial

AtreyeeS:
It is confusing to have different limits for the ul and the sensitivity. Minor change to adapt that.

### DIFF
--- a/examples/tutorials/analysis-3d/non_detected_source.py
+++ b/examples/tutorials/analysis-3d/non_detected_source.py
@@ -247,8 +247,8 @@ print(
 # which can be compared to the flux prediction for a given (hypothetical) source. We have the 5-sigma
 # sensitivity here, which can be configured using ``n_sigma_sensitivity``
 # parameter of this estimator. It is good to compare the computed upper limits with the sensitivity.
-# An upper limit below the sensitivity level of the measurement is likely to be incorrect. 
-# Ensure that you are comparing **UL and sensitivity computed for the same significance level.** 
+# An upper limit below the sensitivity level of the measurement is likely to be incorrect.
+# Please ensure that you are comparing **UL and sensitivity computed for the same significance level.**
 #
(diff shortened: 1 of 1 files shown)


examples/tutorials/analysis-3d/non_detected_source.py
```diff
@@ -246,7 +246,10 @@
 # The `~gammapy.estimators.FluxPointsEstimator` can be used to obtain the sensitivity,
 # which can be compared to the flux prediction for a given (hypothetical) source. We have the 5-sigma
 # sensitivity here, which can be configured using ``n_sigma_sensitivity``
-# parameter of this estimator. Let us see what we would have seen if a Crab-like source was
+# parameter of this estimator. It is good to compare the computed upper limits with the sensitivity.
+# The upper limits should not be lower than the sensitivity limit.
```
bkhelifi: Are we sure that this is true? Is it really the case all the time???

AtreyeeS: I wondered if its ok to say it. I can rephrase it to say "You should carefully interpret the results if your ULs are below the sensitivity limit ".

@facero thoughts?

registerrier: I would say that an upper limit below the sensitivity level of the measurement is likely to be incorrect.

registerrier: But then insist that it is for the same significance level

AtreyeeS:
```suggestion
# An upper limit below the sensitivity level of the measurement is likely to be incorrect. 
# Ensure that you are comparing **UL and sensitivity computed for the same significance level.** 
```

bkhelifi (approved):
Thanks @AtreyeeS

Issue #6403, opened 2026-02-17
Fix warning messages in tutorials


Astro-Kirsty:
There are a number of tutorials which contain some form of error message.

i.e. [H.E.S.S.](https://docs.gammapy.org/dev/tutorials/data/hess.html#dl3-dr1) for the lines
```
[obs.aeff.peek](https://docs.gammapy.org/dev/api/gammapy.irf.EffectiveAreaTable2D.html#gammapy.irf.EffectiveAreaTable2D.peek)()
[plt.show](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.show.html#matplotlib.pyplot.show)()
````

The warning:
```
/home/runner/work/gammapy-docs/gammapy-docs/gammapy/.tox/build_docs/lib/python3.11/site-packages/astropy/units/quantity.py:648: RuntimeWarning: invalid value encountered in divide
  result = super().__array_ufunc__(function, method, *arrays, **kwargs)
```

We should check if these can be mitigated

The tutorials are:

- [ ] [H.E.S.S.](https://docs.gammapy.org/dev/tutorials/data/hess.html#dl3-dr1) 
     - [x] #6450
     - [ ] "UserWarning: This axis already has a converter set and is updating to a potentially incompatible converter"
- [ ]  [CTAO](https://docs.gammapy.org/dev/tutorials/data/cta.html)
- [ ] [MAGIC](https://docs.gammapy.org/dev/tutorials/data/magic.html)
- [ ] [VERITAS](https://docs.gammapy.org/dev/tutorials/data/veritas.html)
- [ ] [IRFs](https://docs.gammapy.org/dev/tutorials/details/irfs.html)
- [ ] [Models](https://docs.gammapy.org/dev/tutorials/details/models.html)

There are plenty more, but we should decide first what we want to do (if anything?)

Basmala92:
@Astro-Kirsty I've been looking into this and reproduced the warnings , in `hess` the warning was caused  by a division by zero in `plot_offset_dependence`  called by `.peek` , I added a condition to check the divison value which removes the warning. 

shall  i open a PR and solve the warnings mentioned in this issue as a start ?

Astro-Kirsty:
Hi @Basmala92, thanks for the interest. It seems that this may already be covered by [#6425](https://github.com/gammapy/gammapy/pull/6425). 
It would be nice for you to open a draft PR for the issue you mention above, to solve just one warning. 
We can go from there

[...]

PR #6475, opened 2026-03-02
Fix UserWarning by resolving color conflict in RegionGeom.plot_region

anshitad58:
This fix handles the `color` and `c` keywords in `RegionGeom.plot_region` to prevent conflicts with the default `edgecolor`. 

Verified locally that this removes the `UserWarning` and allows custom colors to render correctly.

Related to #6403

### DIFF
--- a/gammapy/maps/region/geom.py
+++ b/gammapy/maps/region/geom.py
@@ -892,7 +892,6 @@ class RegionGeom(Geom):
             if user_color is not None:
                 kwargs["edgecolor"] = user_color
 
-
             kwargs_point.setdefault("marker", "*")
 
             for key, value in kwargs.items():
@@ -904,7 +903,6 @@ class RegionGeom(Geom):
                 kwargs_point.setdefault("color", user_color)
             if "color" in kwargs_point:
                 kwargs_point.setdefault("markeredgecolor", kwargs_point["color"])
-
 
             for region in compound_region_to_regions(self.region):
                 region_pix = region.to_pixel(wcs=ax.wcs)


gammapy/maps/region/geom.py
QRemy: because of the change above these lines will be never be reached, this will break the current test
in gammapy/visualization/tests/test_datasets.py/test_plot_regions_color_point

gammapy/maps/region/geom.py
```diff
@@ -887,16 +887,20 @@ def plot_region(self, ax=None, kwargs_point=None, path_effect=None, **kwargs):
                     ax = m.plot(add_cbar=False, vmin=-1, vmax=0)
 
             kwargs.setdefault("facecolor", "None")
-            kwargs.setdefault("edgecolor", "tab:blue")
+            user_color = kwargs.pop("color", kwargs.pop("c", None))
+            kwargs.setdefault("edgecolor", kwargs.pop("ec", "C0"))
```
AtreyeeS, line 891: I prefer having a color like "blue" insteads of "C0"

gammapy/maps/region/geom.py
```diff
@@ -887,16 +887,20 @@ def plot_region(self, ax=None, kwargs_point=None, path_effect=None, **kwargs):
...
             for key, value in kwargs.items():
                 key_point = ARTIST_TO_LINE_PROPERTIES.get(key, None)
                 if key_point and key_point not in kwargs_point:
                     kwargs_point[key_point] = value
 
-            if "color" in kwargs:
-                kwargs_point.setdefault("color", kwargs["color"])
+            if user_color is not None:
+                kwargs_point.setdefault("color", user_color)
             if "color" in kwargs_point:
```
AtreyeeS, line 904: This `if` statement can be removed. `color` on kwargs point will be set

gammapy/maps/region/geom.py
```diff
@@ -887,16 +887,20 @@ def plot_region(self, ax=None, kwargs_point=None, path_effect=None, **kwargs):
...
                 kwargs_point["markeredgecolor"] = kwargs_point["color"]
```
AtreyeeS: This will be problematic. 
With this new logic, if user passes `color=blue` in kwargs and `markeredgecolor=pink` in `kwargs_point`, the latter will be reset to blue.

gammapy/maps/region/geom.py
anshitad58: Thank you for pointing it out, @QRemy. I'll remove the redundant logic and ensure kwargs_point correctly uses the user_color variable. I'll push an update shortly.

AtreyeeS (changes requested):
Thanks @anshitad58 ! 
I think there is a slight problem with this logic, see inline.

This warning mainly comes from matplotlib with overriding preferences between the different color options. Do we really need to fix it?

It was rather annoying to get it to work correctly, see https://github.com/gammapy/gammapy/pull/5697
and I am scared of breaking it

gammapy/maps/region/geom.py
QRemy: "C0" is the default color regardless of the style.

Astro-Kirsty:
> Thanks @anshitad58 ! I think there is a slight problem with this logic, see inline.
> 
> This warning mainly comes from matplotlib with overriding preferences between the different color options. Do we really need to fix it?
> 
> It was rather annoying to get it to work correctly, see #5697 and I am scared of breaking it

I understand about not wanting to break it, but currently there is a warning, so I thought we could try to fix it. It is easily solvable for some of the others, but maybe here is it not

gammapy/maps/region/geom.py
QRemy: no otherwise the None case will overwrite markeredgecolor with the default color

gammapy/maps/region/geom.py
QRemy: color has priority over edgecolor this is the expected behaviour of regions: 
"Setting the 'color' property will override the edgecolor or facecolor properties."

gammapy/maps/region/geom.py
```diff
@@ -887,16 +886,13 @@ def plot_region(self, ax=None, kwargs_point=None, path_effect=None, **kwargs):
                     ax = m.plot(add_cbar=False, vmin=-1, vmax=0)
 
             kwargs.setdefault("facecolor", "None")
-            kwargs.setdefault("edgecolor", "tab:blue")
-            kwargs_point.setdefault("marker", "*")
-
-            for key, value in kwargs.items():
-                key_point = ARTIST_TO_LINE_PROPERTIES.get(key, None)
-                if key_point and key_point not in kwargs_point:
-                    kwargs_point[key_point] = value
```
QRemy, line 896: why removing this ?

anshitad58: @QRemy I have restored the property mapping loop in the latest push.

I originally removed it while refactoring to address the SonarQube complexity failure, but I realize it is necessary for correctly mapping artist properties. But now i have:

Restored the loop: Re-implemented the mapping of kwargs to kwargs_point.

Fixed Linting (Ruff): Locally defined ARTIST_TO_LINE_PROPERTIES within the method to resolve an Undefined name (F821) error.

Verified Locally: Ran pytest gammapy/maps/region/tests/test_geom.py and all 31 tests passed (100%).

gammapy/maps/region/geom.py
```diff
@@ -28,7 +28,6 @@
     compound_region_to_regions,
     regions_to_compound_region,
 )
-from gammapy.visualization.utils import ARTIST_TO_LINE_PROPERTIES
```
QRemy, line 31: keep that and it won't be necessary to redefine after

anshitad58: Thanks @QRemy. I've updated the PR to import ARTIST_TO_LINE_PROPERTIES from gammapy.visualization.utils and removed the local definition.

gammapy/maps/region/geom.py
```diff
@@ -887,16 +887,18 @@ def plot_region(self, ax=None, kwargs_point=None, path_effect=None, **kwargs):
                     ax = m.plot(add_cbar=False, vmin=-1, vmax=0)
 
             kwargs.setdefault("facecolor", "None")
-            kwargs.setdefault("edgecolor", "tab:blue")
+            user_color = kwargs.pop("color", kwargs.pop("c", None))
+            current_ec = kwargs.get("edgecolor", kwargs.get("ec", "C0"))
+            kwargs["edgecolor"] = user_color if user_color is not None else current_ec
+            kwargs.pop("ec", None)
```
QRemy: here ec overwrites edgecolor while usually it's the opposite
using setdefault preserve the priority for edgecolor
so the order of priority is
color>c>edgecolor>ec>C0
```suggestion
            kwargs.setdefault("edgecolor", kwargs.pop("ec", "C0"))
            if user_color is not None:
                kwargs["edgecolor"] = user_color

```

gammapy/maps/region/geom.py
```diff
@@ -887,16 +887,18 @@ def plot_region(self, ax=None, kwargs_point=None, path_effect=None, **kwargs):
...
+
             kwargs_point.setdefault("marker", "*")
+            kwargs_point.setdefault("color", kwargs["edgecolor"])
```
QRemy: this will force edgecolor to overwrite markeredgecolor however only color should have priority over the other color types
```suggestion
```

gammapy/maps/region/geom.py
```diff
@@ -887,16 +887,18 @@ def plot_region(self, ax=None, kwargs_point=None, path_effect=None, **kwargs):
...
 
             for key, value in kwargs.items():
                 key_point = ARTIST_TO_LINE_PROPERTIES.get(key, None)
                 if key_point and key_point not in kwargs_point:
                     kwargs_point[key_point] = value
-
-            if "color" in kwargs:
-                kwargs_point.setdefault("color", kwargs["color"])
             if "color" in kwargs_point:
                 kwargs_point["markeredgecolor"] = kwargs_point["color"]
```
QRemy: and this way points inherit the main color
```suggestion

            if user_color is not None:
                kwargs_point.setdefault("color", user_color)
            if "color" in kwargs_point:
                kwargs_point.setdefault("markeredgecolor", kwargs_point["color"])

```

anshitad58:
Thank you for applying the changes and the force push, @QRemy. I pulled the changes and the tests passed locally on my end.

